PR #211, opened 2014-02-27
Generalizing all_tri to handle 3D, adding support for PRISMs

jwpeterson:
Roy's comments:
```
In theory this branch adds the ability to split prisms into tets (and makes it easier to add
a future capability to split pyramids and hexes into tets).

In practice the refactoring here may have even broken the ability to split quads into tris, 
and the whole thing needs more testing before merge.
```

### DIFF
--- a/src/mesh/mesh_modification.C
+++ b/src/mesh/mesh_modification.C
@@ -40,6 +40,22 @@
 #include "libmesh/string_to_enum.h"
 #include "libmesh/unstructured_mesh.h"
 
+namespace
+{
+  bool split_first_diagonal(const Elem* elem,
+                            unsigned int diag_1_node_1,
+                            unsigned int diag_1_node_2,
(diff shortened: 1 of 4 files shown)


canesin:
Just a kick question, why is this not merged ?? Is this a preparation work for XFEM support ?

jwpeterson:
It requires additional testing.  It extends the capability of all_tri(), which may be useful for XFEM.

jwpeterson:
Ping @roystgnr.  I've added unit tests and I'm satisfied this is at least good enough to merge now.  There were rebase conflicts with your recent unique_ids changes which I *think* I incorporated correctly, but if you could take a look that would be good.  I'd like to get this merged, since it's a pain to rebase every year...

roystgnr:
Sure you don't want to wait for the 2nd anniversary?

Looks sane still to me.

PR #333, opened 2014-08-25
DMlibMesh

dkarpeyev:
This removes the need for PetscDMNonlinearSolver when using DMlibMesh.
In particular, this removes the need for PetscDMNonlinearSolver when solving VIs.
miscellaneous_ex7 has been updated to reflect that as well as the changing PETSc names for VI solvers.
The present treatment of DMlibMesh also makes it easier to extend this DM, 
DMMOOSE of www.mooseframework.org can, for example, move much of its generic
DMCrateXXXDecomposition() functionality into DMlibMesh and only extend the pieces 
necessary for the support of MOOSE-specific contact-based splits.
Similarly, the geometric multigrid support, when added to DMlibMesh, will be usable
from derived DMs.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -422,11 +422,11 @@ am__libmesh_dbg_la_SOURCES_DIST = src/base/dof_map.C \
 	src/solvers/nonlinear_solver.C \
 	src/solvers/petsc_auto_fieldsplit.C \
 	src/solvers/petsc_diff_solver.C \
-	src/solvers/petsc_dm_nonlinear_solver.C \
 	src/solvers/petsc_linear_solver.C \
 	src/solvers/petsc_nonlinear_solver.C \
-	src/solvers/petscdmlibmesh.C src/solvers/slepc_eigen_solver.C \
-	src/solvers/steady_solver.C src/solvers/time_solver.C \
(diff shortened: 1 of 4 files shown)


lindsayad:
It seems like there was never a migration of significant parts of the MOOSE DM upstream AFAICT which is unfortunate. It also is unfortunate that the newer geometric multigrid work of @pbauman and @vikramvgarg did not add-to/use/reuse this code. I do not think there is place where we've done so much re-implementation in libMesh and MOOSE as with the DM code

roystgnr:
IIRC it was @bboutkov who did the new code, not Vikram.  IIRC both Boris and Paul looked at the code here and found reasons to just start from scratch, but I'm not sure if they ever looked at the MOOSE DM code.

I sure would love to see it all consolidated, though.

pbauman:
So it's been a min... but my recollection is that 1. Jed and Matt suggested just starting from scratch because the existing DM bits were targeted at variational inequalities and they were not being actively developed anymore (I think the developer took a different job?), 2. It appeared when I looked at it many a year ago, the existing DM bits were trying to be a proper DM "subclass" via PETSc's DM framework whereas we used the DMShell interface, 3. I think the MOOSE DM bits were actively being developed around the same time and there was not a lot of interest in pushing those into libMesh at that time.

Obviously, I agree consolidating the code would be a good thing.

PR #357, opened 2014-09-23
Update libMesh's AutoPtr class to a modern unique_ptr implementation

jwpeterson:
(This is a follow-on to #335, it does not need to be merged before 1.0 is tagged.)

This branch defines a new `UniquePtr` type, and gracefully deprecates the old AutoPtr class using a new default configure option, `--disable-unique-ptr`.  When we are eventually ready to break backwards compatibility, we just need to change the default over to `--enable-unique-ptr`.

I've tested this branch with MOOSE, which still uses `AutoPtr`, on OSX, and it works fine.


jwpeterson:
This branch is now ready to merge. Since it touches a lot of files and has been under development for quite a while, I'd like to get it merged ASAP.

Summary:
* The branch adds the `--{enable,disable}-unique-ptr` configure flag, which is disabled by default.
* It also introduces the new `UniquePtr` type in LibMesh, which is simply `AutoPtr` if `--disable-unique-ptr` is set.
* If `--enable-unique-ptr` is specified, then `UniquePtr` is one of:
  * `std::unique_ptr`, if your compiler supports it.
  * Howard Hinnant's boost-based C++03-compatible `unique_ptr`, again if your compiler supports it.
  * `LibMesh::AutoPtr` as a last resort.
* Apps which currently use LibMesh's `AutoPtr` should still work just fine with the default configure flags.
* Apps can phase in the `UniquePtr` name slowly over time, or switch everything all at once, at their discretion, as long as libmesh is configured with `--disable-unique-ptr`.

In 90% of cases, all that is required to update one's app code to work with `--enable-unique-ptr` is a text replacement of `AutoPtr` with `UniquePtr`.  The two cases where slightly more work has to be done are:

1) Avoiding `UniquePtr` copies.  The following construct:
```
AutoPtr<X> ap(new X);
return ap;
```
is replaced by
```
return UniquePtr<X>(new X);
```

2) It is ambiguous to construct a `std::unique_ptr` (in a constructor initialization list) with an explicit (NULL) argument. This is apparently because there is a `nullptr` constructor for this class.  Luckily, using the default constructor works just fine, so a constructor initialization list which looks like:
```
Foo () : member(UniquePtr<MemberType>(NULL)) {}
```
is replaced by
```
Foo () : member(UniquePtr<MemberType>()) {}
```

roystgnr:
Is the C++03-compatible unique_ptr usable in STL containers?

jwpeterson:
> On Mar 17, 2015, at 7:09 AM, roystgnr <notifications@github.com> wrote:
> 
> Is the C++03-compatible unique_ptr usable in STL containers?
> 
With C++03 containers, I'd say no, since they aren't copy-able...  Would they work in move-enabled C++11 containers?  Maybe, but I haven't investigated this case, since it implies you'd be using std::unique_ptr anyway...

roystgnr:
That's kind of dangerous, then, isn't it?  I could easily see myself starting to write ```std::vector<UniquePtr<foo> >``` code, it'll work fine for me since I only compile with C++11 on, but it'll break as soon as someone tries libMesh with something ancient.

jwpeterson:
> On Mar 17, 2015, at 8:05 AM, roystgnr <notifications@github.com> wrote:
> 
> That's kind of dangerous, then, isn't it? I could easily see myself starting to write std::vector<UniquePtr<foo> > code, it'll work fine for me since I only compile with C++11 on, but it'll break as soon as someone tries libMesh with something ancient.
> 
I don't really know what to do about that other than say "don't do that" and hopefully the examples and/or unit tests which run with --disable-cxx11 can cover this.

This PR can't simultaneously support C++03 compilers, work with existing libmesh codes, and prevent all possible instances of C++11-code-breaks-with-C++03-compilers code.   The goal is to at least make it *possible* to use a non-deprecated smart pointer type within libmesh and begin the process of updating applications.

roystgnr:
Yeah, I'm happy if we merge this as-is; I'm just trying to figure out any "traps" it could lead us into.  I'm a little paranoid right now because my attempt to fix our BuildBot install didn't work, so I've got no continuous integration with old compilers.

roystgnr:
I was going to suggest that we put in an AutoPtr(UniquePtr) constructor for more backwards compatibility, but even that wouldn't be a total fix (attempting to subclass UniquePtr-based FunctionBase with an AutoPtr-based subclass is still an error).  The --disable-unique-ptr default-for-now will be good enough.

This is working for me.  I'd like to merge this afternoon unless anyone else has concerns.

jwpeterson:
> This is working for me. I'd like to merge this afternoon unless anyone else has concerns.

:+1: Thanks for reviewing.

Issue #397, opened 2014-10-29
Rewrite the FParser automatic differentiation


dschwen:
I'm rewriting the automatic differentiation to make use of the tree representation of expressions generated by the fpoptimizer. This will allow for higher flexibility, such as being able to differentiate conditionals (see #387).

This does not necessarily mean that the FPoptimizer has to be enabled (although I'll have to tweak the Makefile to include the necessary code parts of fpoptimizer in that case)

PR #399, opened 2014-10-30
Rewrite of FParser automatic differentiation

dschwen:
Addressing #397 this new AutoDiff implementation does more with less (370 lines fewer) code and is a lot more readable. 

I've added a test program that compares the derivatives of a bunch of expressions to finite differences.


jwpeterson:
Latest branch (with --disable-fpoptimizer disabled) still doesn't work with some of the MOOSE modules tests (hang):

```
richards:gravity_head_2.gh02...................................................................... RUNNING...
richards:gravity_head_2.gh_lumped_08.............................................................. RUNNING...
richards:gravity_head_2.gh_lumped_17.............................................................. RUNNING...
richards:gravity_head_2.gh_lumped_18.............................................................. RUNNING...
richards:gravity_head_2.gh_bounded_17............................................................. RUNNING...
richards:gravity_head_2.gh_fu_06.................................................................. RUNNING...
richards:gravity_head_2.gh_fu_02.................................................................. RUNNING...
richards:gravity_head_2.gh06...................................................................... RUNNING...
richards:gravity_head_2.gh08...................................................................... RUNNING...
richards:gravity_head_2.gh17...................................................................... RUNNING...
richards:gravity_head_2.gh18...................................................................... RUNNING...
richards:gravity_head_2.gh_fu_17.................................................................. RUNNING...
```

I'm recompiling now in debug mode to see if we can get more error information.

dschwen:
These tests hang in ```SynthesizeByteCode(..)```, an FParser method that transforms a ```CodeTree``` object into byte code. I did not touch this so I'm assuming that I somehow manage to build an invalid ```CodeTree``` object in the automatic differentiation. Probably one rule is broken. I thought I covered all opcodes in my tests... :-/

dschwen:
All hanging tests (and only the hanging tests). Contain the expression ```abs((-b*log(-(gdens0*xval+(-b*exp(-p0/b)))/b)-p1)/p1)``` (the derivative of which then causes the SynthesizeByteCode method to fail). 
At fist glance it does not look as if the CodeTree of that derivative is invalid (it prints fine and a CodeTree is really hard to screw up). FParser uses a rather weak hash to hash subtrees. So this could be a "fluke". However it would lessen my trust in FParser a bit...

jwpeterson:
Replaced by #425.

PR #420, opened 2014-12-09
New candidate for example: systems_of_equations_ ex7

dknez:
This provides a nonlinear elasticity example (St. Venant-Kirchoff material) using NonlinearImplicitSystem.


ehsanakrami:
Hi,
I feel there is a fundamental issue in this problem. Since the grid is not deformed, geometrical parameters such as the element's jacobian (JxW) are calculated regarding the initial geometry, which is not correct. These parameters need to be updated according to the deformed geometry. Would you please let me know if there is a treatment to address this issue inside the algorithm,

dknez:
The approach in the example is correct, e.g. see the "Total Lagrangian" formulation from nonlinear finite element textbooks. This is the most common approach for geometrically nonlinear problems.

PR #429, opened 2015-01-12
Instantiate fparser AD std::complex options

roystgnr:
I was getting build failures with --enable-complex otherwise


benkirk:
I see a compiler error with gcc-4.6, both on my development CentOS 6 machines and vanilla Ubuntu-LTS with gcc-4.6 as the  system compiler.  Also looks like moose-build seems the same issue.

I know internal compiler errors strictly are not our fault, but this one we'll need to be able to work around somehow.

```pre
CXX      libdbg_la-fpoptimizer.lo
fpoptimizer/transformations.cc: In instantiation of 'std::complex<double> {anonymous}::RootPowerTable<std::complex<double> >::RootPowers [20]':
fpoptimizer/transformations.cc:201:21:   instantiated from 'static {anonymous}::PowiResolver::PowiResult {anonymous}::PowiResolver::CreatePowiResult(Value_t) [with Value_t = std::complex<double>]'
fpoptimizer/transformations.cc:755:85:   instantiated from 'bool FPoptimizer_CodeTree::CodeTree<Value_t>::RecreateInversionsAndNegations(bool) [with Value_t = std::complex<double>]'
../../../contrib/fparser/fpoptimizer/instantiate_for_ad.hh:10:5:   instantiated from here
fpoptimizer/transformations.cc:90:19: internal compiler error: Segmentation fault
Please submit a full bug report,
```

dschwen:
I guess that it's why I put that comment there @roystgnr...

roystgnr:
Damn.  Sorry; I only tested with gcc-4.8 and gcc-4.9.

I don't suppose "here's a nickle, go buy yourself a real compiler" counts as a workaround?  ;-)

Looks like gcc-4.6 is still available on Ubuntu 14.10; let me try it and see if I can replicate.

benkirk:
> On Jan 14, 2015, at 8:34 AM, roystgnr <notifications@github.com> wrote:
> 
> I don't suppose "here's a nickle, go buy yourself a real compiler" counts as a workaround? ;-)

I wish, but not really with anything newer than gcc-4.4!

roystgnr:
I think gcc-4.8.2 is where I stopped seeing internal compiler errors from correct C++1y code.

But back on topic: I can replicate the gcc-4.6 bug, it seems to occur in the same place on my machine, and I've found a half-assed workaround: instead of converting to Value_t before the division in that initialization, do the division in long double and convert to Value_t afterwards.

This will cause errors if anyone wants to do extended precision tricks, but it should get us compiling again with --enable-complex and gcc-4.6.  Any objections or suggestions?

jwpeterson:
You're talking about the array initialization around line 90 in transformations.cc?  Simply changing these lines from:
```
Value_t(1) / Value_t(3),
```
to:
```
Value_t(1./3)
```
fixes the ICE?

jwpeterson:
#434

PR #430, opened 2015-01-12
Added System::reinit_constraints().

dknez:
This allows us to update constraints with less overhead than by calling EquationSystems::reinit().
This new method can be used in the case that the mesh hasn't changed.


roystgnr:
Looks good to me.  I'll merge now, and we'll revisit if I run into any problems when I use it from GRINS.

PR #434, opened 2015-01-14
Workaround for GCC 4.6 internal compiler error

roystgnr:
Trying to instantiate this array as std::complex<double> was causing
g++ 4.6 to segfault.

Doing these divisions in long double would lead to errors in cases
where Value_t is using higher (e.g. variable) precision arithmetic;
fortunately we don't instantiate any such cases.

### DIFF
--- a/contrib/fparser/fpoptimizer.cc
+++ b/contrib/fparser/fpoptimizer.cc
@@ -11771,30 +11771,30 @@ namespace
     template<typename Value_t>
     const Value_t RootPowerTable<Value_t>::RootPowers[(1+4)*(1+3)] =
     {
-        // (sqrt^n(x))
+        // (sqrt^n(x)) // Workaround for gcc-4.6 bug by RHS
         Value_t(1),
-        Value_t(1) / Value_t(2),
-        Value_t(1) / Value_t(2*2),
-        Value_t(1) / Value_t(2*2*2),
-        Value_t(1) / Value_t(2*2*2*2),
+        Value_t(1.L/2.L),
+        Value_t(1.L/ (2.L*2.L)),
+        Value_t(1.L/ (2.L*2.L*2.L)),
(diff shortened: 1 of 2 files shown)


jwpeterson:
> Pardon my ignorance, but aren't all these divisions just returning 0 now?!

Huh?  L == long double

dschwen:
> Huh?  L == long double

Nope, not without a decimal point. Those are ```long int```! There is no automatic int->double promotion.

```
#include <iostream>

int main() {
  std::cout << (1L/(2L*2L)) << '\n';
  return 0;
}
```

Outputs ```0```

jwpeterson:
Wow, I didn't know that.  Good catch.

jwpeterson:
Interesting that all the moose tests pass with all the entries in this table set = 0...

dschwen:
Indeed. Do the fparser tests pass? I could dig a little to check where this table is actually used and improve test coverage.

jwpeterson:
> Do the fparser tests pass?

Are they part of "make check"?  If so, then yes.

dschwen:
> Are they part of "make check"? 

I'm pretty sure they are not. Building the fparser test_bed doesn't even work for me right now.

jwpeterson:
> On Jan 14, 2015, at 10:28 AM, Daniel Schwen <notifications@github.com> wrote:
> 
> Are they part of "make check"?
> 
> I'm pretty sure they are not. Building the fparser test_bed doesn't even work for me right now.
> 

What is the issue?

Let's try to get a cppunit test for fparser checked in if we can...

roystgnr:
Wow, I am an *idiot* today.  Thanks for the catch, @dschwen

roystgnr:
From the point of view of historical accuracy I still have some distaste for git rebase and push --force, but from the point of view of historical whitewashing they're awesome!

Issue #437, opened 2015-01-14
Allow addition/removal of new systems after init() has been called


permcody:
The title is self-explanatory. In terms of MOOSE priorities we need the former but not the latter at this time.

permcody:
@andrsd  - Do you have a test case for this feature or better direction of what you may need?

andrsd:
Our projection transfer is waiting for this to be implemented. Then it will either work right out of the box or will need to be finished (I think I did not finish the work because of this, so I lean towards the latter case). So, starting with `tests/transfers/mutliapp_projection_transfer/tosub_master.i` would be a good... start ;-)

roystgnr:
Actually, I appear to have been foolishly running "make check" in a different checkout; it looks like my own unit test might have been enough to trigger the failure after all.  I'll start with that.

roystgnr:
We're passing unit tests now, so I'll give tosub_master.i a shot.  I just need to remove the "deleted" line corresponding to it in tests/transfers/multiapp_projection_transfer/tests before running "make test"?

permcody:
Right, alternatively you can just run moose_test manually on that input
file.

Thanks

jwpeterson:
I think this is done, but feel free to reopen if not.

PR #438, opened 2015-01-14
New post-init systems

roystgnr:
This PR is approximately 10% "make EquationSystems::add_system() work better after EquationSystems::init() has already been called" (as in #437) and approximately 90% "broken stuff I found in the process".

We're passing the most basic unit tests, but I'm guessing this isn't enough to handle whatever problems MOOSE ran into; we'll probably need a more sophisticated test to trigger that.

### DIFF
--- a/include/systems/equation_systems.h
+++ b/include/systems/equation_systems.h
@@ -477,6 +477,12 @@ protected:
 private:
 
   /**
+   * \p true when additional systems do not require immediate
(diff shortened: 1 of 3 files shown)


roystgnr:
That was still expected to fail; I just wanted to rebase on master to get the version of these changes that went in via #445.

roystgnr:
Obsoleted by #478

PR #439, opened 2015-01-15
rbOOmit change: Added inner_product_solver

dknez:
Add a second solver to RBConstruction which can optionally be used
to solve all problems that involve inner_product_matrix. This leads to efficiencies
e.g. we need not recompute the preconditioner for inner_product_solver.


dknez:
This only affects the reduced basis code, so I'll go ahead and merge.

PR #440, opened 2015-01-15
MeshBase::subdomain_ids(), MeshBase::recalculate_n_partitions() should use local iterators

pbauman:
Not Earth-shattering, but noticed these slight inefficiencies (active_elements go over the whole mesh whereas active_local_elements are only on the current processor, right?) and thought I'd throw in a quick fix. `make check` ran correctly on my workstation.

I set the PR to master, but I think it would be appropriate to merge to 0.9.4 for that release.

### DIFF
--- a/src/mesh/mesh_base.C
+++ b/src/mesh/mesh_base.C
@@ -211,8 +211,8 @@ void MeshBase::subdomain_ids (std::set<subdomain_id_type> &ids) const
 
   ids.clear();
 
-  const_element_iterator       el  = this->active_elements_begin();
-  const const_element_iterator end = this->active_elements_end();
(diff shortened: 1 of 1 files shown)


jwpeterson:
> I set the PR to master, but I think it would be appropriate to merge to 0.9.4 for that release.

Since this isn't technically a bugfix, I'd prefer not to merge onto 0.9.4, but would be willing to have the tie broken either way by @roystgnr.  As far as the commit itself, it looks OK to me.

roystgnr:
I'd also prefer not to merge to 0.9.4; wouldn't be upset either way.

Definitely good to merge to master.

PR #441, opened 2015-01-15
Output the PerfLog when we terminate

roystgnr:
This is nice when you can't figure out why a process is taking
forever.  "Ctrl-C" doesn't trigger a terminate, but "kill $ITS_PID"
does.


roystgnr:
Bug in MooseBuild?

milljm:
Sorry about that! I made a change to the way some of our Doxyfiles operate... as such, I missed a location no longer valid... Fixed that just now. But, it will be a while before that change effects libMesh builds...

PR #442, opened 2015-01-16
Slightly Better Support for Mixed Dimension Meshes

pbauman:
This is the first PR to help better support mixed dimensional meshes. The second will update FEMContext/FEMSystem. (FYI, my main focus is mixed 1D/2D meshes.)

I'm going to push a test case to this branch later, but wanted people to be able to start laying eyes on this sooner rather than later as it changes `MeshBase`. Before, it was assumed that the mesh had only one dimension (`MeshBase::_dim`). The issue is that we will need to be able to query the different dimensions of elements present in the mesh without doing a loop over elements each time (e.g. for deciding which FE objects need to be created). So, the main goal of this PR is to add that capability. `mesh_dimension()` still returns the largest dimension present in the mesh. Internally, the dimensions are stored in `std::set` instead of a single `unsigned char`. We can also now query this `std::set` to grab which element dimensions are present in the mesh.

make check passes for me. Again, I'll push a test case here later (hopefully today) - the test case is a mixed 1D and 2D element mesh where 1 EDGE2 element overlaps the edge where 2 QUAD4 elements meet. Then, I do a uniform refinement and check that the node numbers and dof numbers also overlap, i.e. the 1D nodes shared by the 2D element nodes are the same node and similarly for the dofs. I've manually checked so far in the test, now  just need to automate that part and put it in the tests (somewhere... any suggestions?).

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -165,6 +165,12 @@ public:
   { _dim = d; }
 
   /**
+   * @returns set of dimensions of elements present in the mesh.
+   */
+  const std::set<unsigned char>& elem_dimensions() const
+  { return _elem_dims; }
+
+  /**
    * Returns the spatial dimension of the mesh.  Note that this is
    * defined at compile time in the header \p libmesh_common.h.
    */
@@ -880,6 +886,12 @@ protected:
   { return _n_parts; }
 
   /**
+   * Search the mesh and cache the different dimenions of the elements
+   * present in the mesh.
+   */
+  void cache_elem_dims();
+
+  /**
    * The number of partitions the mesh has.  This is set by
    * the partitioners, and may not be changed directly by
    * the user.
@@ -943,6 +955,12 @@ protected:
    */
   std::map<subdomain_id_type, std::string> _block_id_to_name;
 
+  /**
+   * We cache the dimension of the elements present in the mesh.
+   * So, if we have a mesh with 1D and 2D elements, this structure
+   * will contain 1 and 2.
+   */
+  std::set<unsigned char> _elem_dims;
   /**
    * The partitioner class is a friend so that it can set
    * the number of partitions.
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -156,13 +156,19 @@ public:
    * then this will return the largest such dimension.
    */
   unsigned int mesh_dimension () const
-  { return cast_int<unsigned int>(_dim); }
+  { libmesh_assert(!_elem_dims.empty());
+    typedef std::set<unsigned char>::const_iterator it_type;
+    return cast_int<unsigned int>(*(std::max_element<it_type>(_elem_dims.begin(),_elem_dims.end()))); }
 
   /**
-   * Resets the logical dimension of the mesh.
+   * Resets the logical dimension of the mesh. If the mesh has
+   * elements of multiple dimensions, this should be set to the largest
+   * dimension. E.g. if the mesh has 1D and 2D elements, this should
+   * be set to 2. If the mesh has 2D and 3D elements, this should be
+   * set to 3.
    */
   void set_mesh_dimension (unsigned char d)
-  { _dim = d; }
+  { _elem_dims.clear(); _elem_dims.insert(d); }
 
   /**
    * @returns set of dimensions of elements present in the mesh.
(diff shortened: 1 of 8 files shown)


pbauman:
Good point about `std::rbegin()`, thanks. Will make those changes.

pbauman:
So just realized that cppunit is not shipped with libMesh. Any objection to me creating this test within `tests`, but not using cppunit so that this test will be run with `make check` regardless of whether cppunit is present or not?

roystgnr:
That seems like a slippery slope.  What makes this one more important than the rest of the cppunit tests?  Should we move all of them to separate tests?

jwpeterson:
> That seems like a slippery slope.

Agree.  Out of curiosity and presuming your question implies you don't have cppunit installed on your system, what does 'make check' actually do in this scenario?

pbauman:
>  What makes this one more important than the rest of the cppunit tests?

I'm writing it. :P

> Should we move all of them to separate tests?

Personally, I think so. The more test coverage w/o needing 3rd party libraries, the better IMHO.

> Out of curiosity and presuming your question implies you don't have cppunit installed on your system, what does 'make check' actually do in this scenario?

It runs the tests in `contrib`, then goes to `examples` and skips `tests` entirely.

Are both BuildBot and MooseBuild both using cppunit?

pbauman:
> skips tests entirely.

I should've said *silently* skips tests entirely.

roystgnr:
BuildBot is back on life support at the moment, but when it's running, yes, we've got cppunit installed.  As should anyone who has "yum", "apt-get", or "the ability to find a better OS".  ;-)

pbauman:
Yeah, well, you can imagine how hard it is to get `sudo` when someone else is admin. It's not a huge deal if you guys like cppunit, I can do it. I might patch it though so it will at least print that it's skipping those tests if cppunit is not found.

As an aside, one thing I would like to see is the examples actually becoming tests (in addition to examples) - right now, the majority of them pass if they run to completion without checking if the solution is correct/changed from a "gold standard".

roystgnr:
Printing a warning if cppunit isn't installed would be great.  Using a "gold standard" solution in the applicable examples would be extra-great.

friedmud:
Why not add cppunit to contrib? We ship it in the MOOSE package now... but
it would be nice if we didn't have to...

roystgnr:
"Why not add cppunit to contrib?" - just for the same reason we don't add PETSc and SLEPc etc. to contrib - because it's extra megabytes that most people don't need us to hand hold them for.  I could probably have made a *better* case for putting PETSc in contrib/, due to PETSc's more frequent API changes and feature improvements, its greater utility for regular libMesh users, and the more out-of-date distro package alternatives.

That being said... I've now changed my mind about cppunit.  Looking at the intermediate results in #445, it seems that with some versions of cppunit the only way to safely get dbg-mode compatibility is to compile it in dbg-mode as well.  The easiest way for us to do that is add cppunit to contrib/.  If Paul or someone wants to do that, I no longer object.

And on the gripping hand, I would still like to find a solution for contrib/ bloat.  It's now ten times the size of src/!  Plus its very existence encourages a bit of laziness w.r.t. the ability to link to external builds of packages like netcdf that are commonly available (and potential version conflicts) on operating system distributions.

What would be ideal for users is if we did what PETSc does with its third party stuff - don't package useful dependencies with our own source code, but do provide shims that will auto-download and build those dependencies first upon request.  Seems like a lot of work to set up, though, and I don't think we could just drop a copy of their python build infrastructure into autotools.

permcody:
That sounds like a potential summer student project for a good C.S. student.

friedmud:
If any of these contribs are in Git somewhere then we can add them as submodules.  Makes it easy to update them to a new revision too.

Submodules were really made for this sort of thing (an external dependency that you don't _work_ in).  We're using them quite a bit now (for instance libMesh is a submodule in MOOSE).

That would be a really good approach for dealing with PETSc for instance....

pbauman:
Rebased to add new test using @roystgnr CPPUnit addition.

roystgnr:
Forgot to mention, while I was cheerleading cppunit: the actual contents of this patch look good.  Feel free to commit after adding the test(s).

pbauman:
Test added. Once MooseBuild confirms all is well, I'll merge this later tonight.

Issue #432, opened 2015-01-14
PointLocator robustness


permcody:
There are issues with PointLocator failing to find points consistently in certain situations, especially in parallel near element processor borders:

Cross-referencing https://github.com/idaholab/moose/issues/4528

roystgnr:
We've gone a month with no more screaming, so for now I'm going to guess we've fixed all the PointLocator issues.  Go ahead and reopen if/when something crops up.

PR #443, opened 2015-01-16
Tree builds now assert completeness

roystgnr:
I don't think this explains the problems in #432, but it's an additional opportunity for internal verification.  I'll be very interested to see if MooseBuild likes this in dbg mode or not.


roystgnr:
Yeah, I didn't think this was going to catch anything.  I'd still rather keep it in as a protection against future bitrot though.

PR #444, opened 2015-01-19
Revert "Fix JIT compilation when $CXX has spaces in it"

dschwen:
This reverts commit 9f2ac686e6103adf4c3df0accf8655755cb2c26c, which is borked. The code literally inserts the string ```FPARSER_JIT_COMPILER``` as the compiler name. I'm not sure there was a problem to be fixed here in the first place. Compiler names with spaces in them would fail either way. Im my version explicit double quotes are added. The call would simply fail on execution. Such a command (with spaces) would not work to build libmesh in the first place (unless the space is quited, but then my version should work as well).

### DIFF
--- a/contrib/fparser/Makefile.am
+++ b/contrib/fparser/Makefile.am
@@ -101,7 +101,7 @@ EXTRA_DIST += $(FPOPTIMIZER_CC_FILES) fpoptimizer/fpoptimizer_header.txt fpoptim
 pkg_sources = fparser.cc fparser_ad.cc
 if FPARSER_SUPPORT_JIT
   pkg_sources += lib/sha1.cpp
-  pkg_cppflags += -DFPARSER_JIT_COMPILER="$(CXX)"
+  pkg_cppflags += -DFPARSER_JIT_COMPILER="\"$(CXX)\""
 endif
 
 pkg_cppflags += $(FEATURE_FLAGS)
--- a/contrib/fparser/Makefile.in
+++ b/contrib/fparser/Makefile.in
@@ -60,7 +60,7 @@ target_triplet = @target@
 @HAVE_CXX11_MOVE_TRUE@am__append_2 = -DFP_SUPPORT_CXX11_MOVE
 noinst_PROGRAMS = $(am__EXEEXT_1)
 @FPARSER_SUPPORT_JIT_TRUE@am__append_3 = lib/sha1.cpp
-@FPARSER_SUPPORT_JIT_TRUE@am__append_4 = -DFPARSER_JIT_COMPILER="$(CXX)"
+@FPARSER_SUPPORT_JIT_TRUE@am__append_4 = -DFPARSER_JIT_COMPILER="\"$(CXX)\""
 @FPARSER_DEVEL_TRUE@am__append_5 = fpoptimizer/grammar_data.cc
 @FPARSER_DEVEL_TRUE@am__append_6 = util/tree_grammar_parser \
 @FPARSER_DEVEL_TRUE@	util/bytecoderules_parser \
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -77,8 +77,6 @@ private:
   } RefuseToTakeCrazyDerivativeException;
 };
 
-#define FPARSER_MAKE_STR(s) #s
-
 template<typename Value_t>
 FunctionParserADBase<Value_t>::FunctionParserADBase() :
     FunctionParserBase<Value_t>(),
@@ -799,9 +797,9 @@ bool FunctionParserADBase<Value_t>::JITCompileHelper(const std::string & Value_t
   // run compiler
 #if defined(__GNUC__) && defined(__APPLE__) && !defined(__INTEL_COMPILER)
   // gcc on OSX does neither need nor accept the  -rdynamic switch
-  std::string command = FPARSER_MAKE_STR(FPARSER_JIT_COMPILER) " -O2 -shared -fPIC ";
+  std::string command = FPARSER_JIT_COMPILER" -O2 -shared -fPIC ";
 #else
-  std::string command = FPARSER_MAKE_STR(FPARSER_JIT_COMPILER) " -O2 -shared -rdynamic -fPIC ";
+  std::string command = FPARSER_JIT_COMPILER" -O2 -shared -rdynamic -fPIC ";
 #endif
   command += ccname_cc + " -o " + object;
   status = system(command.c_str());


jwpeterson:
FWIW, here is the error message @dschwen says he gets:
```
sh: FPARSER_JIT_COMPILER: command not found
```

Unless different compilers behave differently as far as initializing `std::strings` with the result of macro expansions (is this required to work only for const char *?) I don't really understand how that error message could have happened...

It's also possible I applied @roystgnr's original patch incorrectly --- I had to rebase it on top of the autodiff2 branch and may have left out something important...

roystgnr:
I promise, I only dig deep into fparser guts when I discover that something is broken, not for funsies.

In this case, when I compile your PR, what breaks is:

```
sshcc /opt/apps/ossw/applications/gcc/gcc-4.8/sl6/bin/g++ -DHAVE_CONFIG_H -I. -I../../../contrib/fparser -I../../include  -I../../../contrib/fparser/fpoptimizer -DFPARSER_JIT_COMPILER="\"sshcc /opt/apps/ossw/applications/gcc/gcc-4.8/sl6/bin/g++\"" -DFP_ENABLE_EVAL -DFP_SUPPORT_FLOAT_TYPE -DFP_SUPPORT_LONG_DOUBLE_TYPE -DFP_SUPPORT_LONG_INT_TYPE -DFP_SUPPORT_COMPLEX_DOUBLE_TYPE -DFP_SUPPORT_COMPLEX_FLOAT_TYPE -DFP_SUPPORT_COMPLEX_LONG_DOUBLE_TYPE  -DFP_SUPPORT_CXX11_MOVE    -std=gnu++11 -MT util/util_bytecoderules_parser-bytecoderules_parser.o -MD -MP -MF util/.deps/util_bytecoderules_parser-bytecoderules_parser.Tpo -c -o util/util_bytecoderules_parser-bytecoderules_parser.o `test -f 'util/bytecoderules_parser.cc' || echo '../../../contrib/fparser/'`util/bytecoderules_parser.cc
g++: error: /opt/apps/ossw/applications/gcc/gcc-4.8/sl6/bin/g++": No such file or directory
```

We don't even make it to compilation, because the compiler seems to be interpreting the command line in an unintended way.

dschwen:
Is there really only one double quote in that error line?

roystgnr:
Only one, yes.

dschwen:
Ok, I've configured my libmesh with ```--with-cxx="env mpicxx"``` (just to cook up a compiler command with spaces) and it builds just fine. I verified that it gets used as the JIT command. So I'm having trouble reproducing this issue. How to you configure your compiler? Is your ```CXX``` == ```sshcc /opt/apps/ossw/applications/gcc/gcc-4.8/sl6/bin/g++```?

roystgnr:
Right; I'm just leaving $CXX set, not setting a configure argument.

dschwen:
@roystgnr , help me out here. I now managed to break libmesh compilation with a compile command that contains double quotes, setting ```CXX``` to

```
env testvar="hallo"  mpicxx
```

But it breaks with this error
```
../src/utils/perf_log.C:165:358: error: expected ')'
      std::string libmesh_configure_info("../configure  '--with-methods=opt oprof dbg' '--prefix=/Users/schwd/Programs/libmesh/installed' '--enable-silent-rules' '--enable-unique-id' '--disable-warnings' '--disable-cxx11' '--enable-openmp' '--enable-fparser-jit' '--enable-fparser-debugging' '--enable-sanitize=dbg' '--disable-timestamps' 'CXX=env testvar="hallo"  mpicxx' 'CC=mpicc' 'FC=mpif90' 'F77=mpif77' 'METHODS=opt oprof dbg' 'PETSC_DIR=/opt/moose/petsc/openmpi_petsc-3.5.1/clang-opt'");
                                                                                                                                                                                                                                                                                                                                                                     ^
../src/utils/perf_log.C:165:41: note: to match this '('
      std::string libmesh_configure_info("../configure  '--with-methods=opt oprof dbg' '--prefix=/Users/schwd/Programs/libmesh/installed' '--enable-silent-rules' '--enable-unique-id' '--disable-warnings' '--disable-cxx11' '--enable-openmp' '--enable-fparser-jit' '--enable-fparser-debugging' '--enable-sanitize=dbg' '--disable-timestamps' 'CXX=env testvar="hallo"  mpicxx' 'CC=mpicc' 'FC=mpif90' 'F77=mpif77' 'METHODS=opt oprof dbg' 'PETSC_DIR=/opt/moose/petsc/openmpi_petsc-3.5.1/clang-opt'");
                                        ^
1 error generated.
```

Which seems to be unrelated to my fparser jit stuff.

Can you please give me the contents of your ```CXX``` variable? I don't see how I can get anywhere without it.

roystgnr:
Sorry, thought it was implied by the compiler command line.  I'm using:

```
$ echo $CXX
sshcc /opt/apps/ossw/applications/gcc/gcc-4.8/sl6/bin/g++
```

No odd punctuation other than the space.

dschwen:
Ok, sorry, you are right, I guess I could have inferred this info from your error message above!

I just entered the command line in ```libmesh/build/contrib/fparser```, replacing the first occurrence of your compiler with ```g++``` (and leaving the rest intact).

It compiled without an error. I'm starting to think that the problem lies in your sshcc. could it be possible that it does not retain quotes when passing the commandline to the underlying compiler? That would explain the  error (if I remove the quotes I get the same error)

does not work:
```
-DFPARSER_JIT_COMPILER=\"sshcc /opt/apps/ossw/applications/gcc/gcc-4.8/sl6/bin/g++\"
```

works:
```
-DFPARSER_JIT_COMPILER="\"sshcc /opt/apps/ossw/applications/gcc/gcc-4.8/sl6/bin/g++\""
```

roystgnr:
Gah!  It looks like you're right.  There are a few code paths in sshcc, and not all of them preserve spaces in arguments properly.  Give me 15 minutes to double-check that everything is working now, and I'll merge this.  Please accept my apologies for the screwup.

Only one thing still concerns me - if the stringification trick I put in wasn't working, how have we still been passing MooseBuild?  Is it running on a system without dlopen or with some other quirk that disables JIT fparser?

dschwen:
We may need to improve our moose test coverage for JIT. I wasn't testing for JIT, because it was allowed to fail gracefully (FParser then just continues to use the interpreted version). So JIT failures will not cause moose tests to fail hard.

roystgnr:
Hmm... should we be disabling graceful failure in dbg (or even devel) modes?  I can definitely imagine not wanting to screw up a real run because some compiler couldn't grab a floating license, but when NDEBUG isn't defined (or at least when DEBUG is defined) we can probably afford to be more paranoid.

dschwen:
I already have runtime options for my fparser evaluation to fail hard in case of eval errors (sometimes it is convenient to gloss over a local eval error, like a c<=0 at a single qp in a PF simulation).

I could add a runtime option for this that could be utilized in a moose test. But then we'd require JIT to reliably work on all our test boxes. I'll look into that.

jwpeterson:
> if the stringification trick I put in wasn't working

Yeah, the stringification trick definitely doesn't work for me.  Here's a small example:

```
#include <iostream>
#include <string>

#define STRINGIFY(s) #s

int main()
{
  std::string command = STRINGIFY(CLI_DEFINE) " -O2 -shared -fPIC ";
  std::cout << command << std::endl;
  return 0;
}
```

Compile with:
```
clang++ -DCLI_DEFINE="$(CXX)" stringification.C -o stringification
```

The output for me is:
```
CLI_DEFINE -O2 -shared -fPIC 
```

jwpeterson:
Using additional macro [black magic](http://stackoverflow.com/questions/5256313/c-c-macro-string-concatenation) does work, however:
```
#include <iostream>
#include <string>

// Turn A into a string literal without expanding macro definitions
// (however, if invoked from a macro, macro arguments are expanded).
#define STRINGIZE_NX(A) #A

// Turn A into a string literal after macro-expanding it.
#define STRINGIZE(A) STRINGIZE_NX(A)

int main()
{
    std::string command = STRINGIZE(CLI_DEFINE) " -O2 -shared -fPIC ";
    std::cout << command << std::endl;

  return 0;
}
```

dschwen:
Ok, but what is the problem that this solves?

jwpeterson:
> Ok, but what is the problem that this solves?

It fills a gap in all of our C language knowledge.

dschwen:
Hm, well, there is still the separate issue of compiler commands that contain double quotes, which breaks compilation in other places as well (see above).

jwpeterson:
> compiler commands that contain double quotes

I think I am OK with not supporting this.

PR #446, opened 2015-01-22
Rename protected FEMContext members

pbauman:
This falls squarely in OCD minutiae, but this was annoying enough when reading the code that I went ahead and changed it (actually separated this from a forthcoming PR). Doesn't touch the interface or public members. Compiles and fem_system examples run. Unless I offend @roystgnr or @jwpeterson sensibilities, I'll merge later today. (I did them separately, but I can squash these into 1 if y'all would rather.)

### DIFF
--- a/include/systems/diff_context.h
+++ b/include/systems/diff_context.h
@@ -111,13 +111,13 @@ public:
    * Accessor for element solution.
    */
   const DenseVector<Number>& get_elem_solution() const
-  { return elem_solution; }
+  { return _elem_solution; }
 
   /**
    * Non-const accessor for element solution.
    */
   DenseVector<Number>& get_elem_solution()
-  { return elem_solution; }
+  { return _elem_solution; }
 
   /**
    * Accessor for element solution of a particular variable corresponding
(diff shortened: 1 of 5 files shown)


roystgnr:
As long as we're going OCD on this code, should we have the internals use the accessors too?  That'll mean fewer changes down the road if we ever need to make the accessors more than shims.  I think member-variables-with-accessors are probably better off private than protected, too.

If the intermediate commits compile and run then I wouldn't squash them.

pbauman:
Good idea. I'll make a pass and push to this branch.

pbauman:
OK, I'm done. FEMSystem examples and GRINS `make check` are both happy. @dknez you might want to take a quick look at the (small) changes in DGFEMContext since I don't have any tests for it (did I miss a libMesh example that has it?). Otherwise, I'd like to merge this today.

dknez:
The changes to dg_fem_context.C look fine to me. (There is no libMesh example that uses DGFEMContext, unfortunately. I added it for making a Reduced Basis version of misc_ex9, but I never made that Reduced Basis version into an example.)

PR #447, opened 2015-01-22
Composite functions

roystgnr:
I've given these a lot of use in https://github.com/grinsfem/grins , and could use them in some separate utilities too now.


roystgnr:
Damn, just screwed up a rebase.

PR #448, opened 2015-01-22
Composite functions

roystgnr:
I've given these a lot of use in https://github.com/grinsfem/grins , and could use them in some separate utilities too now.

Rebased so that all three commits should pass "make check" independently.


roystgnr:
There's no changes here to existing code paths, and MooseBuild seems to think I haven't screwed up the build system changes, so I'm merging.

PR #450, opened 2015-01-23
FEMSystem SCALAR assembly

pbauman:
Should be merged after #446.

Skip assembly of SCALAR variables in FEMSystem::assembly is there's no SCALAR variables. Only a small optimization, but is more forward looking to the mixed dimensional support. GRINS `make check` passed, which should hit this code path (in particular the cavity benchmark has a SCALAR variable).

### DIFF
--- a/src/systems/fem_system.C
+++ b/src/systems/fem_system.C
@@ -871,9 +871,20 @@ void FEMSystem::assembly (bool get_residual, bool get_jacobian)
                                          mesh.active_local_elements_end()),
                         AssemblyContributions(*this, get_residual, get_jacobian));
 
+  // Check and see if we have SCALAR variables
+  bool have_scalar = false;
+  for(unsigned int i=0; i != this->n_vars(); ++i)
+    {
+      if( this->variable_type(i).family == SCALAR )
+        {
+          have_scalar = true;
+          break;
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


pbauman:
Try this again after rebasing... not sure what was up with 10888.

src/systems/fem_system.C
```diff
@@ -871,9 +871,20 @@ void FEMSystem::assembly (bool get_residual, bool get_jacobian)
                                          mesh.active_local_elements_end()),
                         AssemblyContributions(*this, get_residual, get_jacobian));
 
+  // Check and see if we have SCALAR variables
+  bool have_scalar = false;
+  for(unsigned int i=0; i != this->n_vars(); ++i)
```
roystgnr: Let's loop over variable groups rather than over variables here?  That should be a little faster in extreme cases, infinitesimally faster in other cases.

roystgnr:
:+1: either as-is or after my suggested change.

src/systems/fem_system.C
pbauman: Good call, thanks for the point that out. I'll push that update in a moment (after running GRINS tests).

pbauman:
Rebased, then going to push the fix that @roystgnr pointed out. Then will merge.

PR #451, opened 2015-01-27
Point locator tree fixes

roystgnr:
This definitely isn't a complete set of fixes (I've still got a mesh where we appear to be failing to find points in the tree) but it does seem to be a correct set of fixes, and it does fix the tree for most of the meshes I've tried it on, so let's see what MooseBuild thinks of it.

For anyone who has use cases related to #432 or https://github.com/idaholab/moose/issues/4528 then you might want to play with them here, or you might just want to wait until I've exhausted all the failure cases I can come up with myself.

### DIFF
--- a/src/utils/tree_node.C
+++ b/src/utils/tree_node.C
@@ -500,40 +500,26 @@ const Elem* TreeNode<N>::find_element_in_children
 {
   libmesh_assert (!this->active());
 
-  unsigned int excluded_child = libMesh::invalid_uint;
+  std::vector<bool> searched_child(children.size(), false);
 
   // First only look in the children whose bounding box
-  // contain the point p.  Note that only one child will
-  // bound the point since the bounding boxes are not
-  // overlapping
+  // contain the point p.
   for (unsigned int c=0; c<children.size(); c++)
     if (children[c]->bounds_point(p, relative_tol))
       {
-        if (children[c]->active())
-          {
-            const Elem* e =
-              children[c]->find_element(p,allowed_subdomains,
-                                        relative_tol);
+        const Elem* e =
+          children[c]->find_element(p,allowed_subdomains,
+                                    relative_tol);
 
-            if (e != NULL)
-              return e;
-          }
-        else
-          {
-            const Elem* e =
-              children[c]->find_element_in_children(p,allowed_subdomains,
-                                                    relative_tol);
+        if (e != NULL)
+          return e;
 
-            if (e != NULL)
-              return e;
-          }
-
-        // If we get here than the child that bounds the
+        // If we get here then a child that bounds the
         // point does not have any elements that contain
         // the point.  So, we will search all our children.
         // However, we have already searched child c so there
         // is no use searching her again.
-        excluded_child = c;
+        searched_child[c] = true;
       }
 
 
@@ -542,26 +528,14 @@ const Elem* TreeNode<N>::find_element_in_children
   // the point p.  So, let's look at the other children
   // but exclude the one we have already searched.
   for (unsigned int c=0; c<children.size(); c++)
-    if (c != excluded_child)
+    if (!searched_child[c])
       {
-        if (children[c]->active())
-          {
-            const Elem* e =
-              children[c]->find_element(p,allowed_subdomains,
-                                        relative_tol);
+        const Elem* e =
+          children[c]->find_element(p,allowed_subdomains,
+                                    relative_tol);
 
-            if (e != NULL)
-              return e;
-          }
-        else
-          {
-            const Elem* e =
-              children[c]->find_element_in_children(p,allowed_subdomains,
-                                                    relative_tol);
-
-            if (e != NULL)
-              return e;
-          }
+        if (e != NULL)
+          return e;
       }
 
   // If we get here we have searched all our children.
--- a/src/utils/tree_node.C
+++ b/src/utils/tree_node.C
@@ -196,7 +196,7 @@ bool TreeNode<N>::bounds_point (const Point& p,
   const Point& min = bounding_box.first;
   const Point& max = bounding_box.second;
 
-  const Real tol = (max - min).size();
+  const Real tol = (max - min).size() * relative_tol;
 
   if ((p(0) >= min(0) - tol)
       && (p(0) <= max(0) + tol)
--- a/src/utils/tree_node.C
+++ b/src/utils/tree_node.C
@@ -180,30 +180,33 @@ void TreeNode<N>::set_bounding_box (const std::pair<Point, Point>& bbox)
 
 
 template <unsigned int N>
-bool TreeNode<N>::bounds_node (const Node* nd) const
+bool TreeNode<N>::bounds_node (const Node* nd,
+                               Real relative_tol) const
 {
   libmesh_assert(nd);
-  return bounds_point(*nd);
+  return bounds_point(*nd, relative_tol);
 }
 
 
 
 template <unsigned int N>
-bool TreeNode<N>::bounds_point (const Point& p) const
+bool TreeNode<N>::bounds_point (const Point& p,
+                                Real relative_tol) const
 {
   const Point& min = bounding_box.first;
   const Point& max = bounding_box.second;
 
+  const Real tol = (max - min).size();
 
-  if ((p(0) >= min(0))
-      && (p(0) <= max(0))
+  if ((p(0) >= min(0) - tol)
+      && (p(0) <= max(0) + tol)
 #if LIBMESH_DIM > 1
-      && (p(1) >= min(1))
-      && (p(1) <= max(1))
+      && (p(1) >= min(1) - tol)
+      && (p(1) <= max(1) + tol)
 #endif
 #if LIBMESH_DIM > 2
-      && (p(2) >= min(2))
-      && (p(2) <= max(2))
+      && (p(2) >= min(2) - tol)
+      && (p(2) <= max(2) + tol)
 #endif
       )
     return true;
@@ -457,25 +460,30 @@ unsigned int TreeNode<N>::n_active_bins() const
 
 
 template <unsigned int N>
-const Elem* TreeNode<N>::find_element(const Point& p, const std::set<subdomain_id_type> *allowed_subdomains) const
+const Elem*
+TreeNode<N>::find_element
+(const Point& p,
+ const std::set<subdomain_id_type> *allowed_subdomains,
+ Real relative_tol) const
 {
   if (this->active())
     {
       // Only check our children if the point is in our bounding box
       // or if the node contains infinite elements
-      if (this->bounds_point(p) || this->contains_ifems)
+      if (this->bounds_point(p, relative_tol) || this->contains_ifems)
         // Search the active elements in the active TreeNode.
         for (std::vector<const Elem*>::const_iterator pos=elements.begin();
              pos != elements.end(); ++pos)
           if (!allowed_subdomains || allowed_subdomains->count((*pos)->subdomain_id()))
-            if ((*pos)->active() && (*pos)->contains_point(p))
+            if ((*pos)->active() && (*pos)->contains_point(p, relative_tol))
               return *pos;
 
       // The point was not found in any element
       return NULL;
     }
   else
-    return this->find_element_in_children(p,allowed_subdomains);
+    return this->find_element_in_children(p,allowed_subdomains,
+                                          relative_tol);
 
   libmesh_error_msg("We'll never get here!");
   return NULL;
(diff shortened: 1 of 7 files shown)


friedmud:
At first glance this looks MUCH more robust. I'll look through it a little
more in depth in a bit.

Thanks Roy!

src/utils/tree_node.C
```diff
@@ -485,40 +493,33 @@ const Elem* TreeNode<N>::find_element(const Point& p, const std::set<subdomain_i
...
-const Elem* TreeNode<N>::find_element_in_children(const Point& p, const std::set<subdomain_id_type> *allowed_subdomains) const
+const Elem* TreeNode<N>::find_element_in_children
+(const Point& p,
+ const std::set<subdomain_id_type> *allowed_subdomains,
+ Real relative_tol) const
 {
   libmesh_assert (!this->active());
 
-  unsigned int excluded_child = libMesh::invalid_uint;
+  std::vector<bool> searched_child(children.size(), false);
```
jwpeterson: Can we avoid `vector<bool>` here?  This isn't threaded code but I'd rather not introduce [potential future data races](http://www.cplusplus.com/reference/vector/vector-bool/) if possible...

src/utils/tree_node.C
```diff
@@ -457,25 +460,30 @@ unsigned int TreeNode<N>::n_active_bins() const
 
 
 template <unsigned int N>
-const Elem* TreeNode<N>::find_element(const Point& p, const std::set<subdomain_id_type> *allowed_subdomains) const
+const Elem*
+TreeNode<N>::find_element
+(const Point& p,
+ const std::set<subdomain_id_type> *allowed_subdomains,
+ Real relative_tol) const
```
jwpeterson: `TreeNode<N>::find_element` is using "non-standard" indentation... It will probably be re-indented the next time we run `make beautify`.

jwpeterson:
> Failed on: linux-gnu

Only a timeout failure in debug mode.  It doesn't seem like you added a huge amount of asserts... The timeout may also be unrelated to your changes, but I've never seen this particular test (outputs/oversample.test_first_order) fail with timeout before.

src/utils/tree_node.C
roystgnr: If we want to avoid potential future threading problems, that precludes using `vector<anything>`!  Sure, `vector<bool>` is the worst case, since "write on one thread, read nearby element on another" can fail; but in any std::vector "append on one thread, read any element on another" can fail.  For that matter, "foo = foo + 1" can fail even when foo is a simple int.  AFAIK the only guaranteed way to avoid threading problems before-the-fact is by writing even the simplest code as nothing but "pure functions" - pass-by-value, no side effects, no non-const variables.

Now, *here*, we could change things to `vector<char>`.  I think "bool" is more self-documenting, but 7 bytes and an extra comment wouldn't be a big deal; let me know if you still want and I'll change it.  Let's just not go nuts with getting rid of vector<bool> elsewhere in the library until we need to; there's a few serial code paths where making an O(n_elem) vector 8 times bigger could hurt.

src/utils/tree_node.C
```diff
@@ -180,30 +180,33 @@ void TreeNode<N>::set_bounding_box (const std::pair<Point, Point>& bbox)
...
 
 template <unsigned int N>
-bool TreeNode<N>::bounds_point (const Point& p) const
+bool TreeNode<N>::bounds_point (const Point& p,
+                                Real relative_tol) const
 {
   const Point& min = bounding_box.first;
   const Point& max = bounding_box.second;
 
+  const Real tol = (max - min).size();
```
friedmud: Am I missing something?  You're passing in `relative_tol` here but not using it... this will currently lead to checking a lot more TreeNodes and hence elements (you're basically making the bounding boxes ~3x larger than they actually are).

src/utils/tree_node.C
roystgnr: Hell, I forgot we had "make beautify"; thanks for the reminder!  Don't even remember the last time I ran that.  Let's set up a separate PR for it the next time we don't have so many existing outstanding pull requests that it might cause conflicts in?

roystgnr:
It might not be an assert causing the timeout.  It might be a combination of "dbg mode is slow" with "elements now fall into more bins and searches might be slower", or it might be an outright bug.  I'll see if I can replicate, but if anyone else has time to look into it too that would be nice.

src/utils/tree_node.C
roystgnr: You're missing nothing; I'm a fool.  Thanks.

src/utils/tree_node.C
jwpeterson: Yeah, but I'm not even talking about read/write race conditions.  There can be a data race in *read-only* threaded code that uses vector<bool>, as I understand it, and we have in fact encountered it before in MOOSE.  

Not to mention that there is apparently [a general consensus among the C++ Standard Committee and the Library Working Group that vector<bool> should be deprecated](http://en.wikipedia.org/wiki/Sequence_container_(C%2B%2B)#Specialization_for_bool)

src/utils/tree_node.C
```diff
@@ -457,25 +460,30 @@ unsigned int TreeNode<N>::n_active_bins() const
...
+(const Point& p,
+ const std::set<subdomain_id_type> *allowed_subdomains,
+ Real relative_tol) const
 {
   if (this->active())
     {
       // Only check our children if the point is in our bounding box
       // or if the node contains infinite elements
-      if (this->bounds_point(p) || this->contains_ifems)
+      if (this->bounds_point(p, relative_tol) || this->contains_ifems)
```
friedmud: This check can't be here.

It's a bit complicated to explain (I tried to explain in my email)....

`find_element_in_children` is expecting to be able to _actually_ search a bounding box if it calls `find_element`... regardless of what its bounding box says.

This call right here: https://github.com/libMesh/libmesh/pull/451/files?diff=split#diff-31ecb7de88a89c206400bd14cf60de8fR534

If this condition is left here then that call doesn't work properly (it will never find anything... because the only way we were looping over those children is if they _failed_ the bounding box check in the first place)

src/utils/tree_node.C
```diff
@@ -527,22 +528,14 @@ const Elem* TreeNode<N>::find_element_in_children(const Point& p, const std::set
...
   // but exclude the one we have already searched.
   for (unsigned int c=0; c<children.size(); c++)
-    if (c != excluded_child)
+    if (!searched_child[c])
       {
-        if (children[c]->active())
-          {
-            const Elem* e = children[c]->find_element(p,allowed_subdomains);
+        const Elem* e =
+          children[c]->find_element(p,allowed_subdomains,
```
friedmud: This is the call that requires that https://github.com/libMesh/libmesh/pull/451/files?diff=split#diff-31ecb7de88a89c206400bd14cf60de8fR473 NOT be there.

The only way we got here is if the child failed the bounding box check... so if we check it inside `find_element()` then we're not doing ANYTHING at all here...

friedmud: This is the call that requires that https://github.com/libMesh/libmesh/pull/451/files?diff=split#diff-31ecb7de88a89c206400bd14cf60de8fR473 NOT be there.

The only way we got here is if the child failed the bounding box check... so if we check it inside find_element() then we're not doing ANYTHING at all here...

src/utils/tree_node.C
roystgnr: The C++ standard specifies otherwise.  Concurrent const operations on C++ containers are required to be thread-safe.

http://stackoverflow.com/questions/10833512/safe-parallel-read-only-access-to-a-stl-container?rq=1

The exception for vector<bool> only refers to non-const modifications.

Do you have any more details on what you encountered in MOOSE?  It wouldn't be the first time we'd had to work around a common compiler bug.

Or for that matter it might not be a bug, technically.  All the above requirements are from the C++11 standard.  The C++98 standard specifies "What are threads?  You're on your own."

The reason for wanting to deprecate vector<bool> (and replace it with some non-STL-looking bitset class, not with "ha-ha, you have to use 8 times more RAM now!") dates back to the weird proxy references it needs, not multithreading concerns.  Right now you can write something as simple as `T& = v[0]` and it will compile for every T *except* bool.

friedmud: It was the "write on one thread, read nearby element on another" case like you were saying.

All write's _and_ reads to `std::vector<bool>` must be blocking in threaded code... even if two processors are working on completely disjoint sets of entries...

All of that said: this is a _local_ variable... so there is no chance of it having threading issues.  Not only that but we need it to be fairly small and quickly searchable because it is in a critical code path.  I would vote for keeping it as `std::vector<bool>`

src/utils/tree_node.C
roystgnr: This does make the backup find_element_in_children search into a null operation...

But I'm not yet convinced, once we get the other issues straightened out, that that fallback should exist.  If we've searched all child nodes that contain the point, but we didn't find the element that contains the point, then that seems to be a failure in the tree construction, and I'd like to first try to fix the construction rather than paper over the failure with more-expensive fallback searches.

On the other hand, fixing the flaw in general would require being able to find a true bounding box for higher-order geometric elements.  If we can't figure out how to do that, then the expensive fallback here would still be preferable to the vastly-more-expensive linear search fallback, and the right thing to do would be to delete that bounding box check.

src/utils/tree_node.C
jwpeterson: > Do you have any more details on what you encountered in MOOSE?  It wouldn't be the first time we'd had to work around a common compiler bug.

Looks like the [thing I was remembering](https://github.com/idaholab/moose/commit/579e54fd64cfa58b14ffe96c5ba61a828c593205) was doing non-const modifications...

Still, I don't see the point of introducing *even the possibility* of hard-to-diagnose future threading bugs by continuing to use `vector<bool>` in any capacity.  Then again, since everyone knows about the issues, maybe it isn't that hard to diagnose...

friedmud:
The bounding box thing relative tolerance thing might have been the reason for the timeout earlier.  That test HEAVILY utilizes the PointLocator so any slowdown would be magnified.

That said - we should really do some timing comparisons with a scaled up version of that test to see how these changes are impacting things.

src/utils/tree_node.C
friedmud: As long as we're talking about removing that piece of `find_element_in_children`... I would vote for removing find_element_in_children!  I think it makes things less clear.  I would prefer to just have one recursive function that does what it's supposed to do...

roystgnr: Ha!  Agreed.  But I'd like to get things completely working first, then get them factored more nicely second.

Dropping bounding box tolerances back to sanity didn't change any of my results... which means that right now I can only trigger failures on a multi-million-element mesh, which makes ongoing debugging tricky.

roystgnr:
Timing comparisons for a not-previously-failing case would be good.

It turns out that the still-failing cases I had were a false positive, broken use case: the point locator was being asked for points well outside of the domain.  Which means:

1. I'd like to pull this as soon as everyone's comfortable.  @nicholasmalaya has some GRINS usage cases that now take minutes to complete, but were churning overnight with little progress when mesh-to-mesh projection was O(N^2) rather than O(N log N).

2. I'll try re-running the line_value_sampler test loop myself, but @friedmud was reporting an intermittent bug there, so if I don't see it I'm not sure if I can safely conclude "it's fixed" and rule out "it doesn't manifest on my system".  If anyone else who's had point locator issues has time to see if they're still replicable then I'd appreciate it.

3.  Now's the time to look at and to either drop or fix that in-tree fallback code.  My inclination is to drop it: figuring out a correct bounding box for quadratic elements is something I ought to do right now *anyway* (note that the MeshTools::bounding_box() code currently fails to do so!  Hope nobody has a convex curved domain without a node at each maxima!), and once that's done then we shouldn't need the fallback in any case.

permcody:
I had to remove this test from the test suite because it was failing several times a day: https://github.com/idaholab/moose/commit/80e6a1a3f6cb47f5073c7649cb9cc1dc9859c28f

Even so, it still might run 100 times on your system before it fails. @friedmud put it in a loop and claimed it ran hundreds of times before failing. Nevertheless it does fail:
https://www.moosebuild.org/view_result/10982

roystgnr:
I ran the loop from @friedmud 2500 times with no failures.  I'm going to try again with a libmesh master build; if I see failures there then I'll assume I fixed them rather than got lucky with them and I'll merge this.

roystgnr:
Can't replicate the failure with libMesh master either.  So can't yet be sure if this resolves #432 - but since it doesn't seem to make things *worse*, we'll merge this, re-enable the previously-intermittently-failing Moose test, and see if the failures ever recur.

PR #453, opened 2015-02-02
Less draconian get_id_by_name()

jwpeterson:
Instead of throwing an error, return the proper type of "invalid" id when a named sideset/nodeset/subdomain is not found.  That way the calling application can decide what to do about it.  As requested by @YaqiWang.

### DIFF
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -468,8 +468,8 @@ public:
   std::string& nodeset_name(boundary_id_type id);
 
   /**
-   * Returns a the id of the requested boundary by name.  Throws an error
-   * if a sideset or nodeset by name is not found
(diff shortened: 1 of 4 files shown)


roystgnr:
:+1: Looks like an improvement to me.  But for future reference: in these types of cases often a quick application-level workaround is to just wrap the questionable method in a try/catch block.  Letting applications decide when an error is recoverable was the original motivation for making libmesh_error() work via exception throwing instead of just aborting.

jwpeterson:
> in these types of cases often a quick application-level workaround is to just wrap the questionable method in a try/catch block

That's true.  In this case I feel like returning invalid_id is more in line with the longer-established behavior of the `BoundaryInfo::boundary_id()` function.  The MeshBase one is more off a toss-up.

PR #454, opened 2015-02-02
Avoid std::unordered container types on Clang/OSX until they are fixed

jwpeterson:
See also this [issue](https://github.com/idaholab/moose/issues/4624#issuecomment-72278831).

The slowness affects `unordered_map`, `unordered_set`, and `unordered_multimap`. This patch makes the BEST_UNORDERED_XYZ types fall back to `map`, `set` and `multimap`, respectively, on Clang/OSX.

### DIFF
--- a/configure
+++ b/configure
@@ -26159,6 +26159,22 @@ else
 fi
 
 
+  # After a lengthy investigation in Jan. 2015, it was determined that
+  # there is a quality of implementation issue with clang's
+  # unordered container types on OSX, which does not appear to affect
+  # Linux.  Until this gets resolved, we will fall back to the "ordered"
+  # containers (map/set/multimap) for this compiler/architecture combination.
+  # https://github.com/idaholab/moose/issues/4624#issuecomment-72278831
+  if test "x$is_clang" != "x" ; then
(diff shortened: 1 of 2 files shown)


roystgnr:
Let's test for clang version <= 3.5.1 before disabling, and submit a bug report since you can replicate with your own LLVM builds?  Just because this doesn't work now doesn't mean we shouldn't prepare for them and help them to fix it the future.

jwpeterson:
> Let's test for clang version <= 3.5.1 before disabling,

I think I'd rather wait for a known working version and start disabling unordered containers prior to that version.  Otherwise we are probably going to silently start using broken unordered containers again without knowing it when 3.5.2 comes out...

> and submit a bug report since you can replicate with your own LLVM builds?

Yeah, I was planning to figure out how to do that at some point.

jwpeterson:
> Yeah, I was planning to figure out how to do that at some point.

Looks like there already is [one](http://llvm.org/bugs/show_bug.cgi?id=21275).  It's for `unordered_map::insert` but it must have the same root issue...

roystgnr:
Okay, we can wait for the fix.

PR #455, opened 2015-02-04
MeshRefinement not sticking

pbauman:
DO NOT MERGE.

I created a PR because 1. I can see if MooseBuild sees this failure too and 2. Easier for everyone else to see the code. This is a dumbed down test case I was working up for #442.

This test fails for me. Somehow the uniformly_refine() call doesn't stick when testRefinedMesh() is called. The order of the test calls should fixed by the macro placement (right?); the print_info() output indicates this. Am I missing something obvious @roystgnr @jwpeterson @benkirk ?

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -15,6 +15,7 @@ unit_tests_sources = \
 	geom/node_test.C \
 	geom/point_test.C \
 	geom/point_test.h \
+        mesh/mesh_refine_wtf.C \
 	numerics/composite_function_test.C \
 	numerics/distributed_vector_test.C \
 	numerics/laspack_vector_test.C \
--- a/tests/Makefile.in
+++ b/tests/Makefile.in
@@ -119,7 +119,8 @@ CONFIG_CLEAN_VPATH_FILES =
 @LIBMESH_OPROF_MODE_TRUE@am__EXEEXT_5 = unit_tests-oprof$(EXEEXT)
 am__unit_tests_dbg_SOURCES_DIST = driver.C test_comm.h \
 	base/dof_object_test.h geom/node_test.C geom/point_test.C \
-	geom/point_test.h numerics/composite_function_test.C \
+	geom/point_test.h mesh/mesh_refine_wtf.C \
+	numerics/composite_function_test.C \
 	numerics/distributed_vector_test.C \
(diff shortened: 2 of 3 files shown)


pbauman:
Here's the two print_info outputs that I see:
```
.................................... Mesh Information:
  mesh_dimension()=2
  spatial_dimension()=3
  n_nodes()=9
    n_local_nodes()=9
  n_elem()=5
    n_local_elem()=5
    n_active_elem()=4
  n_subdomains()=1
  n_partitions()=1
  n_processors()=1
  n_threads()=1
  processor_id()=0

. Mesh Information:
  mesh_dimension()=2
  spatial_dimension()=3
  n_nodes()=4
    n_local_nodes()=4
  n_elem()=1
    n_local_elem()=1
    n_active_elem()=1
  n_subdomains()=1
  n_partitions()=1
  n_processors()=1
  n_threads()=1
  processor_id()=0

F
```

Here's the failure output from CppUnit that I see:
```
!!!FAILURES!!!
Test Results:
Run:  233   Failures: 1   Errors: 0


1) test: MeshRefineTest::testRefinedMesh (F) line: 93 ../../../pbauman/tests/mesh/mesh_refine_wtf.C
equality assertion failed
- Expected: 5
- Actual  : 1
```

pbauman:
OK, so MooseBuild is seeing the same. I feel like this must be something foolish on my part.

jwpeterson:
On Wed, Feb 4, 2015 at 9:46 AM, Paul T. Bauman <notifications@github.com>
wrote:

> OK, so MooseBuild is seeing the same. I feel like this must be something
> foolish on my part.
>

I'd guess that uniformlyRefine() and void testRefinedMesh() are being
called on completely different MeshRefineTest objects.  In other words,
with CPPUnit you can't guarantee that the same object is used to run all
the unit tests.

-- 
John

pbauman:
Ah, OK.

*Reads TFM*

Appears you're right. The macro name SUITE indicated to me this class would run as is. But looks like the addTest hidden in the macro is passing a function pointer to the member function, so it's entirely conceivable that it's using a fresh object each time.

pbauman:
Thanks @jwpeterson and sorry for the noise. This is sorted now. Moral of the story: Don't assume state stays the same between each test method because it doesn't.

PR #456, opened 2015-02-05
Mixed Dimension Mesh Support for FEMContext

pbauman:
Updated `FEMContext` to support cases of mixed dimension meshes. This is accomplished by caching `FEAbstract*` etc. now for all dimensions present in the mesh as well as `FEType`. It is fully backward compatible with dimension dependent calls defaulting to element dimension if it's guaranteed (or at least supposed to be) available or to mesh dimension if it's not. `FEMSystem` examples ran for me and I have a working GRINS test case that I'll create a PR for in a bit, but at least wanted a chance for eyes to be put on this.

### DIFF
--- a/include/systems/fem_context.h
+++ b/include/systems/fem_context.h
@@ -218,29 +218,71 @@ public:
 
 #endif // LIBMESH_ENABLE_SECOND_DERIVATIVES
 
-
-
   /**
-   * Accessor for interior finite element object for variable var.
+   * Accessor for interior finite element object for variable var for
+   * the largest dimension in the mesh. If you have lower dimensional elements
+   * in the mesh and need to query for those FE objects, use the alternative
+   * get_element_fe method.
    */
   template<typename OutputShape>
-  void get_element_fe( unsigned int var, FEGenericBase<OutputShape> *& fe ) const;
+  void get_element_fe( unsigned int var, FEGenericBase<OutputShape> *& fe ) const
+  { this->get_element_fe<OutputShape>(var,fe,this->get_dim()); }
 
   /**
-   * Accessor for interior finite element object for scalar-valued variable var.
+   * Accessor for interior finite element object for scalar-valued variable var
+   * for the largest dimension in the mesh. If you have lower dimensional elements
+   * in the mesh and need to query for those FE objects, use the alternative
+   * get_element_fe method.
    */
-  FEBase* get_element_fe( unsigned int var ) const;
+  FEBase* get_element_fe( unsigned int var ) const
+  { return this->get_element_fe(var,this->get_dim()); }
 
   /**
-   * Accessor for edge/face (2D/3D) finite element object for variable var.
+   * Accessor for interior finite element object for variable var for
+   * dimension dim.
    */
   template<typename OutputShape>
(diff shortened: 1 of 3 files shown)


include/systems/fem_context.h
```diff
@@ -776,18 +842,22 @@ class FEMContext : public DiffContext
...
   /**
    * Pointers to the same finite element objects, but indexed
-   * by variable number
+   * by variable number. We store FE objects for each element dimension
+   * present in the mesh, except for edge_fe_var which only applies
+   * for 3D elements.
    */
-  std::vector<FEAbstract*> _element_fe_var;
-  std::vector<FEAbstract*> _side_fe_var;
+  std::map<unsigned char, std::vector<FEAbstract*> > _element_fe_var;
```
roystgnr: Why a map here?  The extra overhead from a single map node is probably as much as the extra overhead of a size 3 vector would be.

include/systems/fem_context.h
```diff
@@ -776,18 +842,22 @@ class FEMContext : public DiffContext
...
+  std::map<unsigned char, std::vector<FEAbstract*> > _side_fe_var;
   std::vector<FEAbstract*> _edge_fe_var;
```
roystgnr: No dimensionality here because "edge" only makes sense if you're on a 3D element?

pbauman: Correct. I thought I had a comment to that effect... I'll add one.

include/systems/fem_context.h
pbauman: Laziness. Made asserts easier to code later. I'll update to vector.

src/systems/fem_context.C
```diff
@@ -64,78 +66,98 @@ FEMContext::FEMContext (const System &sys)
...
-  for (std::map<FEType, FEAbstract *>::iterator i = _side_fe.begin();
-       i != _side_fe.end(); ++i)
-    delete i->second;
-  _side_fe.clear();
+  for (std::map<unsigned char, std::map<FEType, FEAbstract *> >::iterator d = _element_fe.begin();
+       d != _element_fe.end(); ++d)
+    for (std::map<FEType, FEAbstract *>::iterator i = (d->second).begin();
+         i != (d->second).end(); ++i)
+      delete i->second;
+
```
roystgnr: This looks correct, but is scary.  Could you churn a GRINS run through valgrind (after switching map to vector if you take my advice about that) and make sure we're not leaking or mis-deleting anything here?

pbauman: Will do.

roystgnr:
Nothing obviously wrong here; I like it.

include/systems/fem_context.h
pbauman: OK, there actually is a comment there, it's just not showing up on the diff because the diff is ugly.

pbauman:
OK, updated to using vector data structure instead. GRINS make check runs and my test case runs. Will run through valgrind tomorrow (I don't trust valgrind on the Mac).

pbauman:
Ran through valgrind this morning and all is well. Will merge later today if no further suggestions and/or complaints.

roystgnr:
:+1:

PR #457, opened 2015-02-10
Always repartition mesh nodes

roystgnr:
This fixes potential errors resulting from adaptive mesh coarsening:

Imagine a node N connected to element E on partition 1 and B on
partition 2.  By current libMesh rules, N.processor_id() should be 1.

Now imagine that E is adaptively coarsened away, and N is not also a node of
E->parent().  Now, N.processor_id() should be 2.  But if
skip_partitioning is true, N.processor_id() will remain 1.  When
processor 1 notices that it owns a node but no elements connected to
that node, then assertions (and algorithms, later) fail.

This fixed the case @permcody set up in https://github.com/permcody/libmesh/tree/adaptivity_ex2_assertion for me.  Hopefully it also fixes https://github.com/idaholab/moose/pull/4638 more generally.

Even if it does: I'd like to know before we merge if others consider this to be an adequate fix.  The comment documenting MeshBase::skip_partitioning() suggests that it's an attempt at performance optimization, and certainly tweaking a few node processor ids should still be cheaper than setting up and calling *Metis.  We could just clarify that documentation further.  But if skip_partitioning() is also required to guarantee unchanged processor_id() values, then any bug fix is going to have to be much deeper: IIRC there are places scattered throughout libMesh code where we assume that a loop over local elements wrapping a loop over elements' nodes will touch all local nodes.

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -47,6 +47,8 @@ class Node;
 class Point;
 class MeshData;
 
+template <class MT>
+class MeshInput;
 
 
 /**
@@ -960,11 +962,18 @@ protected:
    * will contain 1 and 2.
    */
   std::set<unsigned char> _elem_dims;
+
   /**
    * The partitioner class is a friend so that it can set
    * the number of partitions.
    */
   friend class Partitioner;
+
+  /**
+   * The MeshInput classes are friends so that they can set the number
+   * of partitions.
+   */
+  friend class MeshInput<MeshBase>;
 
   /**
    * Make the \p BoundaryInfo class a friend so that
--- a/include/mesh/mesh_input.h
+++ b/include/mesh/mesh_input.h
@@ -87,6 +87,14 @@ protected:
   MT& mesh ();
 
   /**
+   * Sets the number of partitions in the mesh.  Typically this gets
+   * done by the partitioner, but some parallel file formats begin
+   * "pre-partitioned".
+   */
+  void set_n_partitions (unsigned int n_parts)
+    { this->mesh().set_n_partitions() = n_parts; }
+
+  /**
    * A vector of bools describing what dimension elements
    * have been encountered when reading a mesh.
    */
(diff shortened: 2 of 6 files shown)


friedmud:
Hmmm... as far as our stuff is concerned... as long as element processor_id's never change and children elements are always owned by the same processor_id as their parents... Then we should be good.

That is a pretty crazy corner case we found there!

Let's see what MooseBuild has to say about it...

roystgnr:
Let me guess: the stateful_prop tests are associating properties with nodes, and aren't expecting nodes to be migrated between processors?

I've got no good guess about what might have killed the plain nemesis test, though...

friedmud:
Nope, all of those failures look like actual logic issues with your new code.

The only reason the Stateful tests are showing up are because they're the only code path that calls `skip_partitioning(true)`

roystgnr:
*Fewer* failing tests now, at least.  I'll see if I can replicate the nemesis failure locally.

permcody:
The call to skip_partitioning is not there for performance optimization. It's there due to a MOOSE limitation. We currently don't have a way of serializing and moving our stateful material properties from processor to processor so we just turn that important load-balancing feature off when we run with adaptivity and stateful materials. We definitely want to fix that limitation. Now that we have unique_ids at our disposal we have the pieces we need to make this a reality.

roystgnr:
Are the stateful material properties associated with nodes, too, or only with elements?

permcody:
Elements

roystgnr:
I think that's got it, then.  The remaining bug was simply that parallel readers were creating partitioned meshes but not setting n_partitions().

I'll merge this afternoon unless anyone screams.

permcody:
Looks good and thanks! I just noticed that we don't run the parallel tests
for MOOSE in the libMesh PR. I probably should add that for future
reference since this issue wouldn't show up in a normal serial run. I'll do
that an invalidate the current build to pick that testing configuration up.

roystgnr:
Thanks!  I'll wait for the additional tests to finish, then.

permcody:
Everything passed in parallel - I'd say merge this when ready

PR #458, opened 2015-02-10
More flexible ParameterVector

roystgnr:
Previously our ImplicitSystem only supported sensitivity solves for derivatives w.r.t. sets of parameters that could be each handed to us via a pointer-to-Number.

This new ParameterVector API still supports the old usage (modulo some places where a backwards-compatible constness change might be required), but now allows ParameterAccessor subclasses to be created with arbitrary getter/setter functions, for supporting consistent perturbations of more complicated parameters: values copied to multiple places in memory, values used to precompute cached data, etc.


roystgnr:
@vikramvgarg has been testing this on the app side, and it looks like our previous apps are all still giving correct sensitivities, so we ought to be good to merge at any time.

We're not yet getting good sensitivity results from the newest code, though, and I'm only 90% sure that's not a library bug, so I'm going to let this PR sit until that's figured out; that way others can take a look at it too.

@pbauman, this is only 2/3rds of what we need to do sensitivities to the simplest GRINS parameters; the last bits are a ParameterMultiPointer class, to register all the different places where something like mu gets copied, and whatever ParameterManager infrastructure we put into GRINS to actually do the registration.  My current thought was that a ParameterManager would look through the input file for parameters to differentiate against, then for each would ask each Physics (referring to parameters by GetPot name, to be canonical) for any copies they keep of that parameter.  Each Physics would ask each submodel in turn, etc.

Further ParameterMultiPointer discussion should probably go here; further ParameterManager discussion into a new GRINS ticket.

roystgnr:
We've figured out what was messing up our new app, and it is going to require library-level fixes (or new features, depending on how you look at it), but it's an issue in master, not in the API changes here.

pbauman:
Thanks for this. I like the addition of the getter/setter.

permcody:
I'm curious about this interface. It's the first I've seen of it but looks like it might be useful to us as well. So the intent is to vary parameters during the simulation? Maybe I haven't given it enough thought yet but why not just hang onto a reference to an individual parameter over time assuming you also have access to the original InputParameters object?

I'm curious because we are currently working on a new general purpose system in MOOSE to tweak live parameters over time too. Not just for sensitivity analysis but for a wide range of "control" based simulations.

pbauman:
@roystgnr might have a better answer, but for me it makes it clearer what code paths would be changing the parameters in libMesh (e.g. if you're doing FD or something) and also provides a clean way of interacting with parameters. E.g. if I'm doing parameter inference where the forward model is a PDE using libMesh, something outside of libMesh (QUESO) will be changing the parameter value and this gives a clean API to interact between MCMC sampler (or whatever) and libMesh-based app to update the parameters.

But maybe I'm missing the point of your question?

permcody:
Not at all, that's a very good reason. Our const parameters won't allow (at least without doing const_casts) users to change parameters where we don't want them to, but I can see the value of using the interface instead. Something to think about.

roystgnr:
I'm not sure whether "I'm curious about this interface" refers to the ParameterVector API in general or to the changes from this PR in particular.  So let's answer both.  :-)

The rationale behind the ParameterVector API in general is simple: we wanted to enable parameter sensitivity solves, including forward and adjoint and hessians and hessian-vector products, and I didn't want to ever have to do all that *again*, so it needed to go into libMesh.  (as it turned out, implicit_system.C specifically)  ParameterVector was just a way of letting the library code change your parameters via pointers, since you probably already had those parameters being owned by your app code and probably didn't want to change that.

The rationale behind the changes in this PR is twofold: 
1) when looking at complex existing apps (as opposed to simpler created-specifically-for-sensitivity-studies apps), I'm finding that parameters not only get owned by the app, they get owned in *multiple* places.  GRINS ends up with a separate copy of viscosity mu in each of several physics objects, for example.  So we need some kind of "multi-pointer" structure to keep track of and change each copy of a parameter.  Worse, in some codes we end up parsing important parameters at initialization time, then using those parameters to precalculate and cache intermediate results.  We only want to redo those precalculations when the parameter changes, not on every residual evaluation.  So even "change each copy of the parameter" isn't always enough; we need a way for the app to be able to say "when you want me to change this, use this callback function so I can update everything that depends on it".  But we're using C++, so abstract base classes and functors are a better alternative to callback functions, and I decided to go with an abstract base class.  (because the coin landed heads, I guess - it'll be trivial to write a functor-accepting subclass later if someone wants that interface instead)
2) I didn't want to break existing library or app code, so I had to make the common case *look* like we were still playing with simple pointer-to-Number data.  It's not possible to *perfectly* preserve the old API, but I wanted to make sure any necessary app updates were both trivial and backward-compatible.  Using a ParameterAccessor defaulting to ParameterPointer gives us compatibility with most pointer resetting operations, and with operator* returning a ParameterProxy (which itself has overloaded everything from "operator Number" to "operator +=", and which is compatible with any ParameterAccessor subclass) we get compatibility with most value operations.

PR #459, opened 2015-02-11
Heterogeneous constraints assembly option

roystgnr:
This isn't ready to merge yet, but I'll put it up ahead of time for public perusal while @vikramvgarg tries some apps on it and while I write up some documentation for it.

Summary: libMesh with FEMSystem or NonlinearSystem currently fails to take parameter sensitivities w.r.t. constraint equation right hand sides, because when we assemble residuals, we're setting up a homogeneously constrained system (the better for solving for delta_u in a Newton loop) and we don't end up having any direct influence from heterogeneous constraints (which get directly applied to u before the nonlinear solve and corrected after each step).  This PR adds an option to request heterogeneous constraint application in a forward problem assembly, modifies FEMSystem to implement that option, and modifies the ImplicitSystem sensitivity code to use that option.

### DIFF
--- a/include/systems/diff_system.h
+++ b/include/systems/diff_system.h
@@ -124,7 +124,8 @@ public:
    * Assembles a residual in \p rhs and/or a jacobian in \p matrix,
    * as requested.
    */
-  virtual void assembly (bool get_residual, bool get_jacobian) = 0;
+  virtual void assembly (bool get_residual, bool get_jacobian,
+                         bool apply_heterogeneous_constraints = false) = 0;
 
   /**
    * Invokes the solver associated with the system.  For steady state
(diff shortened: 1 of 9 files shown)


pbauman:
OK I never hit this because my problem only had homogeneous constraints.

roystgnr:
Actually, sensitivity problems *with* heterogeneous constraints work fine in master; what wasn't working was sensitivity problems *to* heterogeneous constraints.  Like, you've got an inflow profile that's specified as a Dirichlet constraint, and you want to see what happens to your QoI when it's perturbed.

roystgnr:
This feature seems to be working, now that we've run down a red herring of a bug that we triggered with it.  Once we're sure it doesn't regress other sensitivity calculations then we'll merge.

PR #460, opened 2015-02-12
New candidate example: systems_of_equations_ex8

dknez:
This code implements linear elasticity with contact using a penalty approach. I think this demonstrates some interesting capabilities which may be of broad interest, so I thought it'd make a good libMesh example. Note: I haven't tried to optimize this for efficiency at all at this stage; I just wanted to get something working.

The example basically works, e.g. see the attached screenshots that compare the "without contact penalty" vs the "with contact penalty" cases.

without_contact_penalty

with_contact_penalty

However, there are a few problems with this code as it is now, so I was hoping to get a bit of feedback via a PR. The problems are as follows:

* (Minor issue) Apparently I haven't set up the Makefiles correctly yet since when I do "make" I get this error:

> Making all in systems_of_equations/systems_of_equations_ex8
> /bin/bash: line 21: cd: systems_of_equations/systems_of_equations_ex8: No such file or directory
> make[1]: *** [all-recursive] Error 1
> make[1]: Leaving directory `/home/dknez/software/libmesh_build/opt_real/examples'

I'm sure this is easy to fix, can someone point me in the right direction on this? By the way, are the Makefile changes required for adding a new example documented anywhere?

* This example requires the sparsity pattern to be augmented, since the contact force depends on the solution on either side of the contact surface. I added code to augment the sparsity pattern (commented out code around line 447 of linear_elasticity_with_contact.C), but it causes the jacobian matrix to be in an invalid state which leads to a crash. I looked into PetscNonlinearSolver, but it wasn't clear to me what was going wrong. For now, as a temporary hack, I just turned off the PETSc error on allocating new non-zeros (line 455 of linear_elasticity_with_contact.C), but allocating extra non-zeros is slow, so I'd like to get this issue fixed.

* The nonlinear solver appears to converge to reasonable solutions, e.g. as per the screenshots above. But the residual stagnates, and the nonlinear solve terminates due to "DIVERGED_LINE_SEARCH". However, I have a feeling this might be the expected behavior in this case since a contact penalty is non-smooth and hence causes problems for nonlinear solvers.

* I haven't tested this in parallel yet. That will probably require augmenting the send_list as well. I wanted to resolve the augment_sparsity issue before I look into that.


jwpeterson:
> Making all in systems_of_equations/systems_of_equations_ex8
> /bin/bash: line 21: cd: systems_of_equations/systems_of_equations_ex8: No such file or directory
> make[1]: *** [all-recursive] Error 1
> make[1]: Leaving directory `/home/dknez/software/libmesh_build/opt_real/examples'

I get a similar error with your branch on my system:
/bin/sh: line 0: cd: systems_of_equations/systems_of_equations_ex8: No such file or directory

I tried adding that directory by hand in the build directory, but then there was another error:
```
Making all in systems_of_equations/systems_of_equations_ex8
make[2]: *** No rule to make target `all'.  Stop.
```

> I'm sure this is easy to fix, can someone point me in the right direction on this? By the way, are the Makefile changes required for adding a new example documented anywhere?

As far as I know, you updated the Makefile.am in the examples directory correctly, so I'm not sure why it's not working.  Maybe look back at what you did for systems_of_equations_ex7?  The process is not formally documented (that I know of) so please feel add something on the wiki...

jwpeterson:
Ah, you have to also update configure.ac

pbauman:
You need to add the appropriate line to configure.ac. That will generate the Makefile.in, you shouldn't have to edit that, only Makefile.am.

pbauman:
Err, appropriate line being in the AC_CONFIG_FILES part.

jwpeterson:
In before @pbauman's answer!

I'm going to make a wiki page on this really quick...

pbauman:
Arg, ninja'ed!

jwpeterson:
@dknez, [here's](https://github.com/libMesh/libmesh/wiki/Adding-a-new-example) an initial attempt at documenting this process, but since you've created examples most recently it would be great if you could add details/fix errors.  Thanks!

dknez:
Cool, thanks for the help with the Makefile.am and configure.ac stuff. @jwpeterson good idea to document that on the wiki, that was very helpful.

The example installs and runs for me now, so you should be able to try it out.

dknez:
MOOSE doesn't seem to like the mesh I included. I copied the mesh over from miscellaneous_ex9, so there shouldn't be any problem there...

Also, I've been using a direct solver for my testing so far, with "-ksp_type cg -pc_type bjacobi" it doesn't seem to converge properly yet, so I'd need to look into other iterative solver options.

pbauman:
I'm pulling this now to have a look.

pbauman:
Applying the following patch got this to work for me. Basically, need to add symlink the .exo file for out-of-source (VPATH) builds (in addition to the input file).

```
diff --git a/examples/systems_of_equations/systems_of_equations_ex8/Makefile.am b/examples/systems_of_equations/systems_of_equations_ex8/Makefile.am
index 4a3d4c6..fb0b8ef 100644
--- a/examples/systems_of_equations/systems_of_equations_ex8/Makefile.am
+++ b/examples/systems_of_equations/systems_of_equations_ex8/Makefile.am
@@ -11,9 +11,10 @@ if LIBMESH_VPATH_BUILD
   BUILT_SOURCES = .linkstamp
 .linkstamp:
        -rm -f systems_of_equations_ex8.in && $(LN_S) -f $(srcdir)/systems_of_equations_ex8.in .
+ -rm -f systems_of_equations_ex8.exo && $(LN_S) -f $(srcdir)/systems_of_equations_ex8.exo .
        $(AM_V_GEN)touch .linkstamp
 
-  CLEANFILES += systems_of_equations_ex8.in .linkstamp
+  CLEANFILES += systems_of_equations_ex8.in systems_of_equations_ex8.exo .linkstamp
 endif
 
 ##############################################
```

dknez:
@pbauman I made that change, thanks.

pbauman:
I'd run bootstrap again for it to update the Makefile.in

pbauman:
Not sure what MooseBuild's problem is. This runs for me from a fresh build.

dknez:
OK it, passes on MooseBuild now, good.

dknez:
Here's another example screenshot from turning off the load and setting a non-zero Dirichlet BC on the top surface.

nonzero_bc

examples/systems_of_equations/systems_of_equations_ex8/augment_sparsity_on_contact.C
```diff
@@ -0,0 +1,140 @@
...
+void AugmentSparsityOnContact::clear_contact_element_map()
+{
+  _contact_element_map.clear();
+}
+
+void AugmentSparsityOnContact::add_contact_element(
+  dof_id_type element_id,
+  dof_id_type other_element_id)
+{
+  std::map<dof_id_type, std::set<dof_id_type> >::iterator it =
```
roystgnr: Do we want std::map or LIBMESH_BEST_UNORDERED_MAP here?

std::set is probably a decent choice.

examples/systems_of_equations/systems_of_equations_ex8/augment_sparsity_on_contact.C
```diff
@@ -0,0 +1,140 @@
...
+  std::vector<dof_id_type> & n_nz,
+  std::vector<dof_id_type> & n_oz)
+{
+  // get a constant reference to the mesh object
+  const MeshBase& mesh = _sys.get_mesh();
+
+  const DofMap& dof_map = _sys.get_dof_map();
+
+  // Create a set that will have the DoF numbers we need to augment the sparsity pattern
+  std::set<dof_id_type> local_coupled_dofs, remote_coupled_dofs;
```
roystgnr: Here we might consider LIBMESH_BEST_UNORDERED_SET

roystgnr:
Code looks good, that second screenshot looks very good, having augment_sparsity_pattern coverage looks very very good.  :+1:

jwpeterson:
> I copied the mesh over from miscellaneous_ex9

If this uses the exact same mesh as misc_ex9, can you just link it directly instead of adding a second copy to the repository?

After all the small changes get ironed out, I'd also like to squash this branch down to 1 or 2 commits before it gets merged.  That way we can point people to the last commit on this branch for a clean demonstration of how to add a new libmesh example.

dknez:
Thanks for the comments.

@roystgnr I'll make those changes to LIBMESH_BEST_UNORDERED_MAP and LIBMESH_BEST_UNORDERED_SET where you suggested.

Also, regarding this comment:

> having augment_sparsity_pattern coverage looks very very good

I agree! However, the augment_sparsity_pattern is not working at the moment. Some help debugging it would be appreciated :) I can't seem to re-initialize the jacobian matrix inside residual_and_jacobian (see my comment at the top of this thread). The relevant part of the code is around line 447 of linear_elasticity_with_contact.C.

@jwpeterson I'll be happy to squash this down into 1 commit once everything is in order. Also, I can change the systems_of_equations_ex8.exo mesh to a symlink if you prefer?

jwpeterson:
> I'll be happy to squash this down into 1 commit once everything is in order. 

I think at least 2 commits, one for human-added stuff and one for bootstrap-generated stuff.  I can take care of it.  Don't worry, you will still be the author of all commits :)

> Also, I can change the systems_of_equations_ex8.exo mesh to a symlink if you prefer?

Yes, please.

dknez:
I made those small changes: the mesh is now a symlink and I changed to BEST_UNORDERED_SET/MAP.

@jwpeterson if you can take care of commit-squashing at the appropriate time, that'd be great :)

dknez:
I'm having a hard time with this sparsity pattern issue.

My plan was to clear and reinitialize the sparsity pattern inside LinearElasticityWithContact::residual_and_jacobian (see commented code on line 447 of linear_elasticity_with_contact.C). I've tried lots of different variations on this, but it always seems to give a petsc error related to the matrix state.

I think the problem is due to how PetscNonlinearSolver uses the jacobian matrix, e.g. it reuses the same matrix for jac and pc. I can clear and init a matrix without any problems in cases unrelated to PetscNonlinearSolver.

I also tried changing the code to use FEMSystem+PetscDiffSolver, and the same issues arise in that case. (It does run with FEMSystem+NewtonSolver, though, but the linesearch that code uses doesn't seem to work as well for this particular problem.)

dknez:
To follow up on my last comment, I think the better approach will be to provide a wrapper for MatMPIAIJSetPreallocation so that we can update the matrix's sparsity pattern without clearing and re-initializing the matrix. I'll look into that.

dknez:
I think this example is in good shape now, except that I'm hitting an error in parallel with the call to petsc_jacobian->update_preallocation_and_zero();

The error I get with mpiexec -np 2 ./example-dbg is:
nnz cannot be greater than row length: local row 168 value 24 rowlength 0

It seems that local row 168 initially has no "off diagonal" entries (i.e. n_oz[168] = 0 on the first nonlinear iteration), and then on the second iteration we get an error... This makes me wonder if it's invalid to call MatMPIAIJSetPreallocation on an already-initialized PETSc matrix.

dknez:
Thanks to a patch from Dmitry Karpeyev, I think this is in decent shape now. It works in parallel now (thanks Dmitry!).

There are still some convergence problems (i.e. the nonlinear solver terminates with DIVERGED_LINE_SEARCH), but I believe that is due to the fact that the contact penalty is non-smooth and hence is nasty for the nonlinear solver. But the example gives reasonable results, so it seems OK to me.

Let me know if you guys think this is OK to be cleaned up (i.e. squash the commits) and merged?

jwpeterson:
I'd like to get this PR merged so that David can continue developing it a bit more easily. @karpeev, it looks like your branch in PETSc, karpeev/mat-reset, hasn't yet been merged anywhere (`git branch -r --contains cfb96d2` returns only karpeev/mat-reset). So I think let's just commit the petsc-private hacks for now, and eventually add an alternate code path once mat-reset has made it into a PETSc release.

A relevant PETSc commit would seem to be:

```
commit b45d2f2cb7e031d9c0de5873eca80614ca7b863b
Author: Jed Brown <jed@59A2.org>
Date:   Fri Mar 23 20:18:49 2012 -0500

    Rename include/private/ to include/petsc-private/
    
    Hg-commit: 9a1730720b0d9756437b40462617551504641ade
```

which means it probably made it into the Petsc 3.3 release of Jun 5, 2012.  So my plan is to guard this the new update_preallocation_and_zero() code and the new example with 

```
#if !PETSC_VERSION_LESS_THAN(3,3,0)
#endif
```

and update keep an eye on the PETSc development.  I've rebased and cleaned up this branch in upstream at systems_of_equations_ex8_rebased and will merge if moosebuild is still happy with it...

jwpeterson:
Replaced with #510.

PR #461, opened 2015-02-12
Fix -Wundefined-bool-conversion warning in FEMContext.

jwpeterson:
Clang (I believe correctly) warns that because FEMContext::get_elem()
returns a reference, there is no point in checking &(get_elem())
against NULL.

This does raise the question of whether get_elem() itself should check
for NULL before dereferencing _elem...

The full warning text is:
```
../src/systems/fem_context.C:1765:29: warning: reference cannot be bound to dereferenced null pointer
in well-defined C++ code; pointer may be assumed to always convert to true [-Wundefined-bool-conversion]
  unsigned int elem_dim = &(this->get_elem()) ? this->get_elem().dim() : 0;
               ~~~~~~~~    ~^~~~~~~~~~~~~~~~~

./include/libmesh/fem_context.h:709:15: note: 'get_elem' returns a reference
  const Elem& get_elem() const
              ^
```

### DIFF
--- a/src/systems/fem_context.C
+++ b/src/systems/fem_context.C
@@ -1762,16 +1762,14 @@ AutoPtr<FEGenericBase<OutputShape> > FEMContext::build_new_fe( const FEGenericBa
   // everywhere.
   libmesh_assert(&(this->get_elem()) || fe_type.family == SCALAR);
 
-  unsigned int elem_dim = &(this->get_elem()) ? this->get_elem().dim() : 0;
+  unsigned int elem_dim = this->get_elem().dim();
 
   AutoPtr<FEGenericBase<OutputShape> >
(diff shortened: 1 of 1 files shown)


roystgnr:
Hmm... clang is correct; however this looks like an actual regression, not a warning we should squelch.  We need to be testing has_elem() and only using get_elem() if appropriate, and you're right that we ought to be asserting a valid pointer in get_elem().

If you agree, go ahead and close this PR.  I'll try and prepare the more tedious fixes myself now.  Thanks for catching this.

pbauman:
Damn, this is probably my fault (from the FEMContext pedantic updating I did a week or two ago). Let me know if you want me to fix this.

PR #464, opened 2015-02-12
Register dependencies between variables for automatic differentiation

### LINKED ISSUE #463: Add custom derivatives to FParser automatic differentiation
It would be nice if custom rules could be passed into the AD algorithm. As a first step it would be nice if variable dependencies could be resolved that way. I.e. let's say we have a function 

```
F(x,y,t)
```

and ```x``` happens to be ```x(t)``` with ```dx/dt=y```. A use case for this would be coupling in DerivativeMaterial properties into another DerivativeMaterial. 

I'm thinking of a table where tuples ```(a,b,c)``` can be registered by the user such that ```da/db =c```.

```
void
FParserAD::registerDerivative(const std::string & a, const std::string & b, const std::string & c) 
{ ... }
```

In the AD algorithm we'd add a look up to that table for leaf node derivatives (which currently just return either 0 or 1).

dschwen:
Check ```FParserAutodiffTest::registerDerivativeTest()``` for an example of this.

Closes #463.

### DIFF
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -38,7 +38,7 @@ public:
       parser(_parser),
       UnsupportedOpcodeException(),
       RefuseToTakeCrazyDerivativeException() {}
-  int AutoDiff(const std::string& var_name);
+  int AutoDiff(unsigned int);
 
 private:
   /**
@@ -84,6 +84,7 @@ FunctionParserADBase<Value_t>::FunctionParserADBase() :
     compiledFunction(NULL),
     mSilenceErrors(false),
     mFPlog(mData->mFuncPtrs.size()),
+    mRegisteredDerivatives(),
     ad(new ADImplementation<Value_t>(this))
 {
   this->AddFunction("plog", fp_plog, 2);
@@ -96,6 +97,7 @@ FunctionParserADBase<Value_t>::FunctionParserADBase(const FunctionParserADBase& 
     compiledFunction(cpy.compiledFunction),
     mSilenceErrors(cpy.mSilenceErrors),
     mFPlog(cpy.mFPlog),
+    mRegisteredDerivatives(cpy.mRegisteredDerivatives),
     ad(new ADImplementation<Value_t>(this))
 {
 }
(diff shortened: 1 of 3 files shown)


contrib/fparser/fparser_ad.cc
```diff
@@ -345,11 +357,39 @@ int FunctionParserADBase<Value_t>::AutoDiff(const std::string& var_name)
...
+FunctionParserADBase<Value_t>::LookUpVarOpcode(const std::string & var_name)
+{
+  // get c string and length of var argument
+  const unsigned len = unsigned(var_name.size());
+  const char* name = var_name.c_str();
+
+  typename FUNCTIONPARSERTYPES::NamePtrsMap<Value_t> & NamePtrs = mData->mNamePtrs;
+  typename FUNCTIONPARSERTYPES::NamePtrsMap<Value_t>::iterator vi;
+  for (vi = NamePtrs.begin(); vi != NamePtrs.end(); ++vi)
+    if (len == vi->first.nameLength && std::memcmp(name, vi->first.name, len) == 0)
```
jwpeterson: std::memcmp?

tests/fparser/autodiff.C
```diff
@@ -119,6 +116,42 @@ public:
 
     CPPUNIT_ASSERT_EQUAL (passed, ntests);
   }
+
+  void registerDerivativeTest()
```
jwpeterson: This gets run by 'make check', right?

contrib/fparser/fparser_ad.cc
jwpeterson: OK, so it is basically `strncmp`.  Depending on how many places this gets used in the code, it might be worth adding another `operator==` overload to `NamePtr`, which already has this one:

```
         inline bool operator==(const NamePtr& rhs) const
        {
            return nameLength == rhs.nameLength
                && std::memcmp(name, rhs.name, nameLength) == 0;
        }
```

tests/fparser/autodiff.C
dschwen: Yep.

contrib/fparser/fparser_ad.cc
dschwen: I'm not excited about ```std::memcmp```, but that's what _FParser_ is using internally for these comparisons. I don't know if we should modify the FParser core...

jwpeterson: I thought you totally rewrote the FParser core...

I understand, `memcmp` just threw me off as a way of comparing strings.

dschwen: I'm trying to be _minimally invasive_ and try to keep all my contributions in separate derived classes to maintain a chance of upgrading to a newer fparser upstream release if it ever came out. I have contributed my core changes (bug fixes) to upstream.

PR #465, opened 2015-02-13
configure --with-gdb-command="/your/own/gdb"

roystgnr:
This allows us to choose a different gdb path for backtraces (or to eschew gdb entirely, falling back on glibc backtrace if available) at configure time.

### DIFF
--- a/configure
+++ b/configure
@@ -1079,6 +1079,7 @@ enable_exceptions
 enable_timestamps
 enable_unordered_containers
 enable_openmp
+with_gdb_command
 enable_warnings
 enable_blocked_storage
 enable_default_comm_world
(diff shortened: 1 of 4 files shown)


pbauman:
Bless you sir.

jwpeterson:
Nice.  Forgot to mention that the other way I will definitely use this is
```
--with-gdb-command=lldb
```
so this patch definitely looks if you wanted to make the option name more generic e.g. `--with-debugger-command=` I wouldn't complain, but I also wouldn't complain if you merged it as-is.

roystgnr:
How command-line-argument compatible is lldb?  I'd hate to say "You can pick a debugger" explicitly but then in the fine print say "It needs to support -p and -batch and -ex bt and whatever other gdb arguments we might want to use later"; at this point calling what we're asking for a "gdb command" just seems like fair warning.

jwpeterson:
You're right, lldb stupidly still doesn't support -batch so it won't even work as-is...

jwpeterson:
For future reference, the following [hack](http://stackoverflow.com/questions/26267289/how-can-i-exit-lldb-after-running-commands-with-o) works for me with lldb, but you have to have python installed on your system in order to coax lldb into exiting.
```
lldb -p 87187 -o "bt" -o "script import os; os._exit(1)"
```

PR #462, opened 2015-02-12
Fix -Wundefined-bool-conversion warning in FEMContext

roystgnr:
Or not.  I don't actually have a clang-built libMesh handy to test this quickly.

In theory the warnings @jwpeterson found in #461 will be fixed by this PR.  If you could verify or disconfirm that, John, I'd appreciate it.

PR #466, opened 2015-02-16
Turn off asserts in FEMContext::get_elem()

pbauman:
When we have SCALAR variables present, it is assumed that elem=NULL. While I'd like to keep these asserts in principle, this really snowballs on us, granted mostly in GRINS. There's two places in FEMContext that I had to change before I tripped asserts from GRINS calling get_elem() with SCALAR variables.

These asserts were not there before #462, so I don't feel like it's a huge step backwards. Thoughts @roystgnr?

### DIFF
--- a/include/systems/fem_context.h
+++ b/include/systems/fem_context.h
@@ -707,14 +707,14 @@ public:
    * Accessor for current Elem object
    */
   const Elem& get_elem() const
-  { libmesh_assert(this->_elem);
+  { //libmesh_assert(this->_elem); PB: SCALAR variables have elem=NULL
     return *(this->_elem); }
 
   /**
    * Accessor for current Elem object
    */
   Elem& get_elem()
-  { libmesh_assert(this->_elem);
+  { //libmesh_assert(this->_elem); PB: SCALAR variables have elem=NULL
     return *(const_cast<Elem*>(this->_elem)); }
 
   /**


pbauman:
I should've noted that with these changes, GRINS master `make check` runs successfully again.

jwpeterson:
Can you describe "really snowballs on us"?  Asserting that a pointer is non-NULL before dereferencing it is always a solid programming principle... so it seems like the application should be fixed here.

Also, if you are going to remove them, delete them, don't comment them out.

roystgnr:
An assert here means that app code has to only use get_elem() if has_elem() returned true, or we break in an easy to diagnose way in dbg mode.

A lack of assert here means that app code has to only use get_elem() if has_elem() returned true, or we're dereferencing a NULL pointer, which is undefined behavior in C++, which may work perfectly on even numbered days and break in horribly unexpected ways on odd numbered days.

See http://blog.llvm.org/2011/05/what-every-c-programmer-should-know_14.html to get some sense of what "horribly unexpected ways on odd numbered days" can mean.

pbauman:
Yeah, I understand all the dereferencing NULL pointer business.

"Snowballs on us" means there's a whole bunch of places in the app code where have to now add checks for `has_elem()` and, in some cases do a lot more work since things like subdomain_id and so on are tied to the `Elem`.

But I'll close this out and do the refactoring.

PR #467, opened 2015-02-16
Only get_elem() if FEMContext has_elem().

pbauman:
Need to guard calls to get_elem() with has_elem() calls. App code is still broken, so I don't know if I've hit all the places yet so don't merge this yet.

### DIFF
--- a/src/systems/fem_context.C
+++ b/src/systems/fem_context.C
@@ -1628,7 +1628,12 @@ void FEMContext::pre_fe_reinit(const System &sys, const Elem *e)
   this->set_elem(e);
 
   // Initialize the per-element data for elem.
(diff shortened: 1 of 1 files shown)


roystgnr:
Looks great so far.  Sorry I didn't run the GRINS suite and catch these; I forgot we don't have any FEMContext+Scalar coverage in the libMesh suite yet.

pbauman:
OK, those are the last two I hit. When MooseBuild and others are happy, this should be good to merge.

roystgnr:
No libmesh_PR_test_dbg recipe is running?

roystgnr:
Never mind; my browser hadn't updated yet.

roystgnr:
Looks good to merge after that finishes.

PR #468, opened 2015-02-17
Use NaN for NaN, not zero

roystgnr:
Get NaN from Real, where it should exist, not from Number, which may be std::complex, in which case quiet_NaN() silently returns zero, rather than sanely returning NaN or throwing an exception or sending Stroustrop hate mail.

This seems to fix the adjoints_ex3 --enable-complex regression for me.  I'm going to run some more complex tests and some GRINS tests before merging.

Assuming this works, we ought to be able to cherry-pick it onto 0.9.4 branch and do a final release of that - the --enable-complex regression there was the last thing that was holding us up.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -262,7 +262,7 @@ private:
         if (c)
           return f_fem->component(*c, i, p, time);
         else
-          return std::numeric_limits<Number>::quiet_NaN();
+          return std::numeric_limits<Real>::quiet_NaN();
       }
     return f->component(i, p, time);
   }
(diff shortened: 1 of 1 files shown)


jwpeterson:
According to [this](http://en.cppreference.com/w/cpp/types/numeric_limits)

"Non-arithmetic standard types, such as std::complex<T> ...  do not have [numeric_limits] specializations"?  

If that's the case, I'm not sure why `std::numeric_limits<Number>` even compiles?  Are we working around double-bad (it both compiles something it shouldn't and returns a bad value at runtime) implementations on a specific compiler?

roystgnr:
There's no *specialization* but there is a *default*.

PR #470, opened 2015-02-19
Bug fix for PointLocatorList 

pbauman:
Before, the cached distance used the distance from the first element
in the list. But, the loop didn't include the first element so if
the first element was the closest one, we didn't detect it and the
accompanying unit test failed. I put the unit test in the mixed
dimension unit test because this bug was found as part of some
refactoring I'm doing and I went ahead and made this bug fix standalone.

### DIFF
--- a/src/utils/point_locator_list.C
+++ b/src/utils/point_locator_list.C
@@ -161,11 +161,11 @@ const Elem* PointLocatorList::operator() (const Point& p, const std::set<subdoma
   {
     std::vector<std::pair<Point, const Elem *> >& my_list = *(this->_list);
 
-    Real              last_distance_sq = Point(my_list[0].first -p).size_sq();
+    Real              last_distance_sq = std::numeric_limits<Real>::max();
     const Elem *      last_elem        = NULL;
     const std::size_t max_index        = my_list.size();
 
-    for (std::size_t n=1; n<max_index; n++)
+    for (std::size_t n=0; n<max_index; n++)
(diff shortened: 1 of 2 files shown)


jwpeterson:
Wow.  Definitely seems like unsafe code, you never know when the `my_list` vector you get back might be empty; it's undefined behavior to access the 0th element in that scenario.

I went back really far (SVN r378, July 2003) to find where the "list" and "tree" PointLocator types originally split from the original PointLocator class, and it hasn't changed substantially since that time.  I think most people use the Tree type, so thanks for adding the associated regression test.

:+1:

jwpeterson:
BTW, here's Daniel Dreyer's comment from the time, in case it is relevant for you.
```
reworked the PointLocator structure.  PointLocator<T> disappeared,
didn't make sense.  Instead, PointLocatorTree and PointLocatorList
are there.
But use PointLocatorList with CAUTION!, since this is primarily intended
for locating the closest element, where the point may live in 3D, and the
element, too, but the element is actually 2D (surface).  When applied to
volumetric meshes (which is not truly intended), it may happen that
the centroid of an adjacent element is closer due to extremely distorted
elements
```

pbauman:
Yeah, I don't I'd use this but I'm having to work on the PointLocator API's and went ahead and put the unit tests in in anticipation and tripped this. So I figured I'd drop this fix in. I'll let MooseBuild finish before merging.

Issue #469, opened 2015-02-18
Regression in (or triggered by?) Gmsh changes


roystgnr:
Building with METHODS=dbg --enable-complex I find that reduced_basis_ex7 succeeds when using cf8fe174f3dfccc2dda54007aa6ba6076366c507 but fails when using 5260d07dc8bd02935f5f4ad43c13994522bece42

This is a "handle mixed-dimension meshes" commit from @jwpeterson, and so there's a small chance this is relevant to #433.  The actual failure is in MeshCommunication::broadcast(), and so there's a large chance that the actual error is something in broadcast_packed_range() which merely was triggered by John's new code.

If anyone is familiar with GMsh/.geo/.msh (is this your example, @dknez?), I'd love help scaling down horn.msh to ease debugging.

jwpeterson:
Can you post the stack trace to this issue?  Also, I forget, does one need a complex-enabled PETSc to run this example successfully in --enable-complex mode?

It's definitely possible there are bugs in the commit you mentioned, but I don't see how they would be related to Real/complex... the patch doesn't really do anything significant with Reals or Numbers.

jwpeterson:
Also -- the referenced commit doesn't have anything to do with mixed-dimension *libmesh* meshes, only with handling BCs specified on lower-dimensional elements in gmsh files.  The lower-dimensional elements themselves don't make it into the final Mesh.

roystgnr:
The --enable-complex is required to run the whole example, but I can strip out the relevant ifdefs and get the same failure with little more than ```mesh.read("horn.msh");```

The assertion failure is the ```verify(mesh.n_elem())``` at mesh_communication.C:716, and the rest of the stack is just unstructured_mesh.C:769 and reduced_basis_ex7.C.

roystgnr:
Running "gmsh -2 horn.geo" gives me a much smaller horn.msh file (albeit one with roughly the same number of elements?) but the new file exactly reproduces the original error so I'm not too concerned about being unable to exactly reproduce the original mesh.

jwpeterson:
> The assertion failure is the verify(mesh.n_elem())

Hmm.. are you running in parallel?  How many procs?

roystgnr:
[...]

PR #471, opened 2015-02-19
Update n_elem/n_nodes caches more frequently

roystgnr:
This seems to fix the regression with Gmsh->ParallelMesh reads


roystgnr:
This fixes #469 for me.

roystgnr:
Exactly.  I never worried much about those "intermediate" values, since in any case it's impossible to ensure their perfect accuracy at all times (what do you do if a remote processor deletes an element you don't have a copy of without telling you?) and since in prepare_for_use() they eventually get hit by updates to fix their accuracy... but in the case we're looking at, where the mesh is getting built on just one processor, there's no good excuse for not keeping n_elem() accurate even before the final mesh preparation.

PR #472, opened 2015-02-19
Mixed Dimension Support for Trees and PointLocators

pbauman:
Ultimate goal is to have ExactSolution do the right thing with mixed dimension meshes. This is the first stage. Second stage will be MeshFunction updates. Final stage will be ExactSolution.

I've updated the Tree* and PointLocator* APIs to now accept the desired dimension of the element in the search. I've retained the former API which defaults to the mesh_dimension which should be completely backward compatible. I've also updated the MixedDimensionMeshTest to test out the PointLocators (which actually uncovered #470).

In doing this, I preferred to avoid include the mesh_base.h in the header file Tree* and PointLocator* headers; this is why the 1 liner backward compatible function implementations are in the .C files.

I also made the detection of missing elements of dimension elem_dim a runtime check instead of an assert.

### DIFF
--- a/include/utils/tree.h
+++ b/include/utils/tree.h
@@ -76,13 +76,15 @@ public:
 
   /**
    * @returns a pointer to the element containing point p,
+   * for elements of dimension elem_dim,
    * optionally restricted to a set of allowed subdomains,
    * optionally using a non-zero relative tolerance for searches.
    */
-  const Elem* find_element(const Point& p,
-                           const std::set<subdomain_id_type>
-                           *allowed_subdomains = NULL,
-                           Real relative_tol = TOLERANCE) const;
+  virtual const Elem* find_element(const Point& p,
+                                   unsigned int elem_dim,
+                                   const std::set<subdomain_id_type>
+                                   *allowed_subdomains = NULL,
+                                   Real relative_tol = TOLERANCE) const;
 
   /**
    * @returns a pointer to the element containing point p,
--- a/include/utils/tree.h
+++ b/include/utils/tree.h
@@ -88,6 +88,7 @@ public:
 
   /**
    * @returns a pointer to the element containing point p,
+   * for elements with dimension mesh.mesh_dimension(),
    * optionally restricted to a set of allowed subdomains,
    * optionally using a non-zero relative tolerance for searches.
    */
--- a/include/utils/tree_base.h
+++ b/include/utils/tree_base.h
@@ -90,13 +90,24 @@ public:
 
   /**
    * @returns a pointer to the element containing point p,
+   * for elements of dimension mesh->mesh_dimension(),
    * optionally restricted to a set of allowed subdomains,
    * optionally using a non-zero relative tolerance for searches.
    */
-  virtual const Elem* find_element(const Point& p,
-                                   const std::set<subdomain_id_type>
-                                   *allowed_subdomains = NULL,
-                                   Real relative_tol = TOLERANCE) const = 0;
+   const Elem* find_element(const Point& p,
+                            const std::set<subdomain_id_type> *allowed_subdomains = NULL,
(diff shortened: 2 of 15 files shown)


include/utils/tree.h
```diff
@@ -76,16 +76,19 @@ class Tree : public TreeBase
...
   /**
    * @returns a pointer to the element containing point p,
+   * for elements of dimension elem_dim,
    * optionally restricted to a set of allowed subdomains,
    * optionally using a non-zero relative tolerance for searches.
    */
-  const Elem* find_element(const Point& p,
-                           const std::set<subdomain_id_type>
-                           *allowed_subdomains = NULL,
-                           Real relative_tol = TOLERANCE) const;
```
roystgnr: We're removing the original signature of the function?

This is a public function; although I doubt anyone's directly using it I'd rather not break the API just in case.  Could you add the elem_dim version as a separate overload again?

pbauman: Oops, thanks for catching that. I meant to leave all the APIs intact.

include/utils/tree.h
```diff
@@ -76,16 +76,19 @@ class Tree : public TreeBase
...
+  virtual const Elem* find_element(const Point& p,
+                                   unsigned int elem_dim,
+                                   const std::set<subdomain_id_type>
+                                   *allowed_subdomains = NULL,
+                                   Real relative_tol = TOLERANCE) const;
```
pbauman: Oh, wait, no, it's still there, just down a little ways.

include/utils/tree_base.h
```diff
@@ -90,13 +90,24 @@ class TreeBase : public ReferenceCountedObject<TreeBase>
...
    * optionally using a non-zero relative tolerance for searches.
    */
-  virtual const Elem* find_element(const Point& p,
-                                   const std::set<subdomain_id_type>
-                                   *allowed_subdomains = NULL,
-                                   Real relative_tol = TOLERANCE) const = 0;
+   const Elem* find_element(const Point& p,
+                            const std::set<subdomain_id_type> *allowed_subdomains = NULL,
+                            Real relative_tol = TOLERANCE) const;
+
```
pbauman: Oh, wait, no, it's still there, just down a little ways.

roystgnr:
Everything else looks good to me.  I'd say merge as soon as you've restored that API and MooseBuild has given it another once-over.

include/utils/tree.h
pbauman: For crying out loud, even reading the diff I'm getting confused by all the find_element functions. I'll fix this up in a minute.

pbauman: Oh wait, never mind. The it's still there, but as a non virtual method in TreeBase. Internally, it call find_element with mesh_dimension so it will dispatch to the derived class.

src/utils/tree_base.C
```diff
@@ -0,0 +1,30 @@
...
+#include "libmesh/tree_base.h"
+#include "libmesh/mesh_base.h"
+
+namespace libMesh
+{
+  const Elem* TreeBase::find_element(const Point& p,
+                                     const std::set<subdomain_id_type> *allowed_subdomains,
+                                     Real relative_tol) const
+  {
+    return this->find_element(p,this->mesh.mesh_dimension(),allowed_subdomains,relative_tol);
```
pbauman: Here's the base class version dispatching.

pbauman:
OK, I should've pointed this out in the original PR statement. The original signature was pure virtual in TreeBase. What I did was make that version non virtual and have it dispatch to the new version, which is pure virtual in TreeBase and implemented now in the subclasses.

I'll way for @roystgnr to confirm he's OK with it before merging.

roystgnr:
Sorry about the confusion; thanks!

PR #473, opened 2015-02-20
Mixed Dimension Support for MeshFunction

pbauman:
Updated MeshFunction to support Point queries with element dimension restriction. Previous methods default to mesh_dimension(). I ran GRINS `make check` with this and all is well (ExactSolution uses this implicitly). Note I have not yet exercised the non mesh_dimension() case, that will be forthcoming.

This should be merged after #472.

### DIFF
--- a/include/mesh/mesh_function.h
+++ b/include/mesh/mesh_function.h
@@ -219,6 +219,10 @@ public:
 
 protected:
 
+  /**
+   * Helper function to reduce code duplication
+   */
+  const Elem* find_element( const Point& p ) const;
 
   /**
    * The equation systems handler, from which
--- a/include/mesh/mesh_function.h
+++ b/include/mesh/mesh_function.h
@@ -145,16 +145,24 @@ public:
 #endif
 
   /**
-   * Computes values at coordinate \p p and for time \p time, which
-   * defaults to zero.
+   * Computes values at coordinate \p p, that lies in element with
+   * dimemnsion mesh_dimension() (for mixed dimension meshes), and for
+   * time \p time, which defaults to zero.
    */
   void operator() (const Point& p,
                    const Real time,
                    DenseVector<Number>& output);
 
   /**
-   * Computes gradients at coordinate \p p and for time \p time, which
-   * defaults to zero.
+   * Computes values at coordinate \p p, that lies in element with
+   * dimemnsion elem_dim (for mixed dimension meshes), and for
+   * time \p time, which defaults to zero.
+   */
+  void operator() (const Point& p,
+                   unsigned int elem_dim,
+                   const Real time,
+                   DenseVector<Number>& output);
+
    */
   void gradient (const Point& p,
                  const Real time,
--- a/include/mesh/mesh_function.h
+++ b/include/mesh/mesh_function.h
@@ -183,10 +183,21 @@ public:
                  std::vector<Gradient>& output);
 
   /**
-   * Computes gradients at coordinate \p p and for time \p time, which
-   * defaults to zero.
+   * Computes gradients at coordinate \p p, that lies in element with
+   * dimemnsion mesh_dimension() (for mixed dimension meshes), and for
+   * time \p time, which defaults to zero.
    */
   void hessian (const Point& p,
+                const Real time,
+                std::vector<Tensor>& output);
+
+  /**
+   * Computes gradients at coordinate \p p, that lies in element with
+   * dimemnsion elem_dim (for mixed dimension meshes), and for
+   * time \p time, which defaults to zero.
+   */
+  void hessian (const Point& p,
+                unsigned int elem_dim,
                 const Real time,
                 std::vector<Tensor>& output);
 
--- a/include/mesh/mesh_function.h
+++ b/include/mesh/mesh_function.h
@@ -163,8 +163,22 @@ public:
                    const Real time,
                    DenseVector<Number>& output);
 
+  /**
+   * Computes gradients at coordinate \p p, that lies in element with
+   * dimemnsion mesh_dimension() (for mixed dimension meshes), and for
+   * time \p time, which defaults to zero.
    */
   void gradient (const Point& p,
+                 const Real time,
+                 std::vector<Gradient>& output);
+
+  /**
+   * Computes gradients at coordinate \p p, that lies in element with
+   * dimemnsion elem_dim (for mixed dimension meshes), and for
+   * time \p time, which defaults to zero.
+   */
+  void gradient (const Point& p,
+                 unsigned int elem_dim,
                  const Real time,
                  std::vector<Gradient>& output);
 
--- a/include/mesh/mesh_function.h
+++ b/include/mesh/mesh_function.h
@@ -222,7 +222,7 @@ protected:
   /**
    * Helper function to reduce code duplication
    */
-  const Elem* find_element( const Point& p ) const;
+  const Elem* find_element( const Point& p, unsigned int elem_dim ) const;
 
   /**
    * The equation systems handler, from which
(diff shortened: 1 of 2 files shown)


roystgnr:
I've got a concern with this design.

What happens when we try to, say, run our projection app on a pair of mixed dimension meshes?

On a single-dimension mesh, we build the MeshFunction, we have a couple functions that can query it for point values and gradients, project_solution uses those to get data at vertices and local quadrature points, we get the result we want.

On a mixed-dimension mesh, we try to do all the same... but now project_solution fails at any point where a lower-dimensional element exists but a higher-dimensional element does not, because the FunctionBase-derived methods all default to querying higher-dimensional elements only.

Having a projection fail isn't the end of the world (we have the same problem on domains with "slits", where you might accidentally query a node or element on the wrong side of the slit, and I don't see a good way to fix that).  But we could probably make it a little more robust by changing the default behavior - in cases where "query the highest-dimensional element we find" fails to find any elements of mesh_dimension(), could we fall back on "try the next-highest-dimensional elements" etc. until we truly fail to find an element of any dimension containing the point?

pbauman:
First, you're right, projection is not going to work as is, but it also wouldn't work before this PR (because internally, the methods explicitly used mesh_dimension). So, in that sense, I'm not breaking backward compatibility.

I'm trying to proceed in small steps, adding functionality as I need it. This particular PR targets the next one which is support for mixed dimension meshes for ExactSolution (ExactSolution uses MeshFunction to query the attached reference EquationSystem). In that vein, I'm trying to add the new functionality needed (calling operator(), gradient, and hessian with element dimension) while preserving the current API.

That said, I see what you're saying, but I think that the element dimension should not be the concern of the FunctionBase classes. For example, in ExactSolution, we need to construct finite element objects and query the solution based on the element dimension. So, *those calling* functions need to worry about the dimensionality of the element and then use these new APIs to supply it. Indeed, in the case where you have more than 1 element containing the point (such as my case), you *need* to supply which one you want. 

So, my plan moving forward was adding these new APIs where we can query based on the dimension of the element and the calling methods worry about supplying that dimension. This is exactly my plan for ExactSolution (I've implemented it, I'm just testing and preparing the PR) and I think projection should work in the same way. However, as you've pointed out, this PR was not targeted at supporting projection operations so they won't be supported yet. In fact, I avoided updating the FunctionBase APIs here for exactly that reason, keeping this PR self contained. So this provides a small new chunk of functionality, but it's not complete functionality for mixed dimension meshes, which will be ongoing.

But maybe you don't agree with this plan, so I'm open to suggestions.

pbauman:
Rebased on #472

friedmud:
I just wanted to throw out there that we do have fairly extensive testing
of our uses of MeshFunction in MOOSE.  So you can be assured that if
MooseBuild is telling you it's ok... then you're not breaking our stuff.

BTW: This is all positive... I'm really happy to see all of these changes
going in surrounding mixed dimensional use cases....

roystgnr:
I'm happy merging this as-is, then; "small steps" sounds great to me.  I don't think my proposed fix would be very intrusive, though.

I'm not sure I'm happy with the idea of in the future forcing FunctionBase users to worry about elem_dim.  FunctionBase was supposed to be an ABC representing F(x,y,z,t) - do we really want to change that to F(x,y,z,t,elem_dim)?  At least the default behavior ought to be dimension agnostic.

I can see how adding an elem_dim option would simplify some use cases, where elements of one dimension overlap elements of a different dimension with different values.  But there's probably lots of possibilities for expanding the function domain that would each make a couple different use cases simpler.  A normal_vector, to handle the slit case?  A subdomain_ids set, to handle overlapping elements of the same dimension?

...

Huh.  Actually, that subdomain_ids alternative would be able to handle overlapping same-dimension elements, and the slit case, and the overlapping different-dimension elements case that elem_dim covers.  Maybe that's worth considering?  We've already got the underlying support in the PointLocator hierarchy, too.

pbauman:
> I don't think my proposed fix would be very intrusive, though.

You're right and I think it should be added. I've been trying to think about the appropriate place to put it. `PointLocator`? Or in `MeshFunction`?

> I'm not sure I'm happy with the idea of in the future forcing FunctionBase users to worry about elem_dim. FunctionBase was supposed to be an ABC representing F(x,y,z,t) - do we really want to change that to F(x,y,z,t,elem_dim)? At least the default behavior ought to be dimension agnostic.

I agree with this. For sure the the default behavior is agnostic, but I'm not sure what the best solution will be long term for the API. Another reason to proceed in small steps, to try things out. I agree having elem_dim violates the spirit of the class. But we'll need a way to pass along which elements information we want. Just FYI, for my present case, the solution is continuous so operator() is not so much a worry, but gradient and hessian are (in fact what got me started down this path).

> I can see how adding an elem_dim option would simplify some use cases, where elements of one dimension overlap elements of a different dimension with different values. But there's probably lots of possibilities for expanding the function domain that would each make a couple different use cases simpler. A normal_vector, to handle the slit case? A subdomain_ids set, to handle overlapping elements of the same dimension?

This is a very legitimate point, maybe I'm moving in the wrong direction.

> Huh. Actually, that subdomain_ids alternative would be able to handle overlapping same-dimension elements, and the slit case, and the overlapping different-dimension elements case that elem_dim covers. Maybe that's worth considering?

For sure it would cover my cases - I'm using subdomain ids to restrict which Physics we use (ElasticCable for 1D and ElasticMembrane for 2D). What are you thinking exactly? We pull the subdomain id from the current element in the element loop and then use that to restrict the subsequent computations? So for MeshFunction, we'd need to add an API to pass an std::set of subdomain ids... I think that could work. In fact, if we are going to assume different dimension elements have different subdomain ids, we could probably roll back #472 because we could pass that set down to the PointLocator, as you mentioned.

Hmm... What do you think @roystgnr? I think this might be a better solution. We'd be adding a restriction that the user must separate the subdomain ids between different dimension elements, which is probably a reasonable restriction (and I'm already doing that anyways)...

pbauman:
@friedmud @permcody , would such a restriction be acceptable for how you guys handle mixed-dimension meshes? I.e. different subdomain ids for different dimensioned elements?

friedmud:
Yes - that's totally fine.

Exodus doesn't support mixed element types in the same Block... let alone
mixed dimension!  So even if someone wanted to do it... how would they make
the mesh and read it in?  :-)

pbauman:
OK, I'll try this out, without #472, and see if I can make it work.

I'm still not sure how we'd handle FunctionBase without adding a way to pass down the subdomain id with the `Point`. I'm all ears for suggestions there.

roystgnr:
This is so grossly general that my "FunctionBase must be kept pure!" sensibilities don't seem to be triggering.  So I'd actually be happiest simply adding an additional optional default-NULL allowed_subdomains argument to the FunctionBase APIs, same semantics as PointLocator.

roystgnr:
The question of what to do in the library projection routines is more interesting, though... my inclination would be to, at each point calculation, initially query the FunctionBase with only the current subdomain id allowed (thereby doing the right thing in cases of overlap), but test the result for something like the MeshFunction::_out_of_mesh_value, and if that's what we get then we try a query with NULL allowed_subdomains as a fallback.  That way we do the right thing in cases where there is overlap, but we don't break any existing code that tries to project between domains with non-matching subdomain decompositions.

friedmud:
My only thought is that you could supply it in the constructor to a MeshFunction... so each MeshFunction would only be for a certain subdomain and any calls to `operator()` would automatically do the right thing from there...

friedmud:
I can't imagine any scenarios when you would want to "blindly" allow it to
select between higher and lower dimensional meshes.  That is just asking
for trouble.

But: I won't be perturbed if that's the default behavior either.... I just
won't use it ;-)

pbauman:
Closing this in favor of #475.

PR #475, opened 2015-02-24
MeshFunction subdomain ids support

pbauman:
Effectively replaces #473 based on the discussion there. In particular, we're going to assume that elements with different dimension will have different subdomain ids and that the user/other library code will construct subdomain id sets accordingly. Note I haven't updated the `FunctionBase` API yet, I was saving that for updating the projections, which will be more substantial work. In particular, I've left the old `operator()` API intact and it internally calls the new one with NULL subdomain ids pointer. I haven't put a `libmesh_deprecated` in yet --- will do that as part of the projections work. 

I've tested this with `ExactSolution` (forthcoming PR) as well as with GRINS and all is well.

Note that I did this without the changes #472 so I'll also be creating a PR to roll that one back.

### DIFF
--- a/include/mesh/mesh_function.h
+++ b/include/mesh/mesh_function.h
@@ -154,11 +154,14 @@ public:
 
   /**
    * Computes gradients at coordinate \p p and for time \p time, which
-   * defaults to zero.
+   * defaults to zero, optionally restricting the point to the passed
+   * subdomain_ids. This is useful in cases where there are multiple
+   * dimensioned elements, for example.
    */
   void gradient (const Point& p,
                  const Real time,
-                 std::vector<Gradient>& output);
+                 std::vector<Gradient>& output,
+                 const std::set<subdomain_id_type>* subdomain_ids = NULL);
 
   /**
    * Computes gradients at coordinate \p p and for time \p time, which
--- a/include/mesh/mesh_function.h
+++ b/include/mesh/mesh_function.h
@@ -165,11 +165,14 @@ public:
 
   /**
    * Computes gradients at coordinate \p p and for time \p time, which
-   * defaults to zero.
+   * defaults to zero, optionally restricting the point to the passed
+   * subdomain_ids. This is useful in cases where there are multiple
+   * dimensioned elements, for example.
    */
   void hessian (const Point& p,
                 const Real time,
-                std::vector<Tensor>& output);
+                std::vector<Tensor>& output,
+                const std::set<subdomain_id_type>* subdomain_ids = NULL);
 
   /**
    * Returns the current \p PointLocator object, for you might want to
--- a/include/mesh/mesh_function.h
+++ b/include/mesh/mesh_function.h
@@ -219,6 +219,10 @@ public:
 
 protected:
 
+  /**
+   * Helper function to reduce code duplication
+   */
+  const Elem* find_element( const Point& p ) const;
 
   /**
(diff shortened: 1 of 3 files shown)


include/mesh/mesh_function.h
```diff
@@ -219,6 +237,11 @@ class MeshFunction : public FunctionBase<Number>,
 
 protected:
 
+  /**
+   * Helper function to reduce code duplication
+   */
+  const Elem* find_element( const Point& p,
+                            const std::set<subdomain_id_type>* subdomain_ids = NULL ) const;
```
roystgnr: Thank you for this.

roystgnr:
:+1:

include/mesh/mesh_function.h
```diff
@@ -146,27 +146,45 @@ class MeshFunction : public FunctionBase<Number>,
...
+   * restricting the point to the passed subdomain_ids. This is useful in
+   * cases where there are multiple dimensioned elements, for example.
+   */
+  void operator() (const Point& p,
+                   const Real time,
+                   DenseVector<Number>& output,
+                   const std::set<subdomain_id_type>* subdomain_ids);
+
+  /**
    * Computes gradients at coordinate \p p and for time \p time, which
```
roystgnr: Why two function overloads here?  Why not just a default argument value subdomain_ids=NULL?

pbauman: The former is a virtual function overriding the `FunctionBase` (pure virtual) definition. I was trying to avoid redoing the FunctionBase here because that could spiral on us a bit. However, I debated which way to go and went the lazy route. If you'd prefer me not introducing the new overload, I'll go back and update `FunctionBase` as well. My plan there would be the same as here: Add the new overload and keep the old one with a `libmesh_deprecated()` inside and have it call the new one with NULL for the default. Otherwise, we'll break the API on everyone subclassing FunctionBase. But maybe there's another way to go about it?

roystgnr: No, I'm an idiot.  We don't want to force people subclassing FunctionBase to change their code, so we can't change the method signature, so we have to add a new overload.

I don't think we should deprecate the old method, though.  Just because people are now *allowed* to worry about which subdomain they're querying, that doesn't mean they should be *forced* to worry about it.  Ideally, most Functions should still be functions.  ;-)

pbauman: > No, I'm an idiot. We don't want to force people subclassing FunctionBase to change their code, so we can't change the method signature, so we have to add a new overload.

You want me to go ahead and do that here?

> I don't think we should deprecate the old method, though. Just because people are now allowed to worry about which subdomain they're querying, that doesn't mean they should be forced to worry about it. Ideally, most Functions should still be functions. ;-)

Agreed, I was just thinking so that we could, in the future, have one method were we default to NULL for the subdomain ids instead of persisting with two.

roystgnr: The FunctionBase changes can go in a separate PR.  Let's merge this now.

I'm ambivalent about whether moving to one method is worth the hassle.  There's nothing *wrong* about the original API, it's just less flexible.  If you feel strongly (or if someone else does and wants to pipe up) then we could deprecate it, though.

pbauman: I don't feel strongly about it, but if someone else does, then I'll take care of it.

PR #476, opened 2015-02-24
ExactSolution subdomain ids support

pbauman:
Do not merge before #475.

This builds off of #475 and refactors `MeshFunction::_compute_error` a bit to support the mixed dimension mesh case. We do this by using the current elements subdomain id to feed to `MeshFunction`. As far as I've tested, this is backward compatible, but we'll see if MooseBuild hits anything. We are making an assumption though.

The assumption we are making here is that the two systems we are comparing have compatible subdomain ids. For example, this will not work if one system has 1D elements labeled as subdomain id 1 and the comparing system has 1D elements with subdomain id 2.

With these changes, GRINS `make check` passes, including my new (grinsfem/grins#220) mixed dimension regression test case.

### DIFF
--- a/src/error_estimation/exact_solution.C
+++ b/src/error_estimation/exact_solution.C
@@ -609,40 +609,6 @@ void ExactSolution::_compute_error(const std::string& sys_name,
   // Attach quadrature rule to FE object
   fe->attach_quadrature_rule (qrule.get());
 
-  // The Jacobian*weight at the quadrature points.
-  const std::vector<Real>& JxW                               = fe->get_JxW();
-
-  // The value of the shape functions at the quadrature points
-  // i.e. phi(i) = phi_values[i][qp]
-  const std::vector<std::vector<OutputShape> >&  phi_values         = fe->get_phi();
-
-  // The value of the shape function gradients at the quadrature points
-  const std::vector<std::vector<typename FEGenericBase<OutputShape>::OutputGradient> >&
-    dphi_values = fe->get_dphi();
-
-  // The value of the shape function curls at the quadrature points
-  // Only computed for vector-valued elements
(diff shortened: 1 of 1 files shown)


friedmud:
This makes sense to me. Let's see what the build system says about it...

jwpeterson:
> Do not merge before #475.

If you have PRs which depend on others, you should branch off from the other PR (instead of master) when making them... that way if somebody merges something in the "wrong" order, github will detect that and automatically close the other PR too.

pbauman:
Hmmm, I thought that's exactly what I did here. In fact, I'd thought that's why the #475 changes were in the diff on this PR because #475 had not yet been merged to master, but it was in the history since I branched from that branch.

jwpeterson:
> Hmmm, I thought that's exactly what I did here. In fact, I'd thought that's why the #475 changes were in the diff on this PR because #475 had not yet been merged to master, but it was in the history since I branched from that branch.

OK, sounds good.  It's not super important that we merge #475 first then, since this one will merge both.

pbauman:
Went ahead and rebased now that #475 is merged to make the diff a little easier to read.

roystgnr:
Nothing looks unreasonable to me.

pbauman:
Fixed the typos in the log. Thanks @roystgnr

pbauman:
> OK, sounds good. It's not super important that we merge #475 first then, since this one will merge both. 

Thanks @jwpeterson 

In the future, I'll use language along the lines of "Build off of PR...".

pbauman:
I'll leave this up for `max(MooseBuild finishes, pbauman-heads-home)` before merging, unless there are any further complaints/points-of-discussion/etc.

PR #477, opened 2015-02-24
Rollback PR #472

pbauman:
In hindsight, #472 was the wrong way to handle the mixed dimension cases (this is why @roystgnr is the chief software architect and I'm the "code-talking guy" ;-) ), so rolled back that PR (details of how I did it are in the commit log). Then, updated the unit test to use the subdomain id to test restricting the PointLocator search.


pbauman:
Huh, I guess not. I'd thought git revert would rewrite history, which I was trying to avoid, but upon reading [this](https://github.com/git/git/blob/master/Documentation/howto/revert-a-faulty-merge.txt) I guess it wouldn't have.

jwpeterson:
> I guess it wouldn't have.

Right, it just makes a new commit that magically gets rid of all the bad codes :)

pbauman:
I'll leave this up for a few more hours and merge before I head home, unless there are any complaints/point-of-discussion/etc.

PR #478, opened 2015-02-25
Post-init add_system()

roystgnr:
This obsoletes #438 - it's been rebased for compatibility, I've tried a saner approach, and I've fixed a few overzealous assertions.

### DIFF
--- a/include/numerics/petsc_vector.h
+++ b/include/numerics/petsc_vector.h
@@ -928,13 +928,15 @@ void PetscVector<T>::init (const NumericVector<T>& other,
 
   this->_type = v._type;
 
-  if (v.size() != 0)
-    {
+  // We want to have a valid Vec, even if it's initially of size zero
+
+//  if (v.size() != 0)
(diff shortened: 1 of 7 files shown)


roystgnr:
This actually has a chance of resolving #437, and even if it doesn't it's a step in the right direction.  I'll merge later today once people have had a little time to glance it over.

roystgnr:
Alternatively, I'll add another assertion first.  MOOSE's tests/transfers/multiapp_projection_transfer/tosub_master.i seemed to be *almost* working, but I'm seeing an error, not due to the now-supported post-EquationSystems::init() system addition, but in a still-unsupported post-System::init variable addition.  I'd like to see whether MooseBuild finds that happening anywhere else or not.  If it's an unintentional artifact of that one specific test then we can try to fix the Moose code triggering it; if enough other code paths are doing the same thing then we should probably change the behavior from "unsupported but you might get lucky" to "supported".

PR #480, opened 2015-02-26
Allow adding additional variables to existing FParser objects

dschwen:
This allows users to add variables to existing and parsed FParser objects (for example copies of fully build parsers). The main use is to add variables to use ```RegisterDerivative``` from #463 to make them targets of custom derivatives on demand (i.e. at different levels in an automatic differentiation chain to get a high order derivative).

I'm also using ```friend``` to allow the ```FunctionParserADBase``` class access to private members of ```FunctionParserBase```. These classes are closely tied together and the only alternative would be straight up modifying ```FunctionParserBase``` to put in the AD functionality.

### DIFF
--- a/contrib/fparser/fparser.hh
+++ b/contrib/fparser/fparser.hh
@@ -32,6 +32,9 @@
 #pragma warning(disable : 4661)
 #endif
 
+template<typename Value_t>
+class FunctionParserADBase;
+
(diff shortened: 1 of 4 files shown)


jwpeterson:
If this works with the MOOSE apps, :+1:

PR #481, opened 2015-02-27
GetPot Unit Tests + GetPot::have_section()

pbauman:
Added CppUnit tests for GetPot. Actual source file has to be generated by configure since CppUnit doesn't provide access to argv (annoying) and GetPot only takes file names in the constructor, no streams (so couldn't do stringstream). Motivation for the unit tests was adding a new API to GetPot - have_section(). It's useful to be able to test whether a given section is in the input file. To facilitate a binary search instead of a linear search, I added a sort() to the section_list at the end of parsing.

### DIFF
--- a/include/base/getpot.h
+++ b/include/base/getpot.h
@@ -201,6 +201,9 @@ public:
   inline bool have_variable(const char* VarName) const;
   inline bool have_variable(const std::string& VarName) const;
 
+  bool have_section(const char* section_name) const;
+  bool have_section(const std::string& section_name) const;
+
   /**
    * scalar values
    */
@@ -2141,7 +2144,20 @@ GetPot::have_variable(const std::string& VarName) const
   return have_variable(VarName.c_str());
 }
 
+inline bool
+GetPot::have_section(const char* section_name) const
+{
+  std::string s = std::string(section_name);
+  return this->have_section(s);
+}
 
+inline bool
+GetPot::have_section(const std::string& section_name) const
+{
+  /* We sorted the section_list after we were done parsing
+     so we can use a binary search instead of a linear search. */
+  return std::binary_search( section_list.begin(), section_list.end(), section_name );
+}
 
 template <typename T>
 inline T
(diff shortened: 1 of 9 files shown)


jwpeterson:
FYI, looks like the failure was in the new unit test...
```

1) test: GetPotTest::testSections (F) line: 73 base/getpot_test.C
assertion failed
- Expression: input.have_section("Section2/Subsection2/Subsection3")
```

pbauman:
Thanks. Getting back from dinner and am going to take a look. Might've forgotten a last commit in my hurry to get out.

pbauman:
OK, weird. I could've sworn those were passing before I left the office. The network is conveniently down so I can't see what happened. Nevertheless, fixed and squashed that commit and added documentation that when querying the section, it *must* end in a "/". e.g. sections names are "/Section1/Section2/", etc.

pbauman:
Would it be better to have the method look at your string and append the "/" if it's missing? I was trying to play it safe by not mucking with what was coming in, but I'm not thinking of any cases that it could bite you. Could save a lot of headaches...

pbauman:
Wat?!

I don't see how I'd be hosing y'all. Looked like a few Exodus diff problems. Are these new functions somehow colliding? Perhaps something because I bootstrapped? Other?

jwpeterson:
Hmm... looks like you are now sorting the entries in the GetPot object?  I'm not sure, but I think `materials/derivativematerialinterface.property_names` is a new test that relies on input file ordering.  So that might be the issue...

pbauman:
Yeah, I sorted the section_list, but at parsing time so that it was sorted before anyone could call 'GetPot::get_section_names()'. OK, I'll revert the sorting and use a linear search in have_section().

include/base/getpot.h
```diff
@@ -202,6 +202,24 @@ class GetPot
   inline bool have_variable(const std::string& VarName) const;
 
   /**
+   * Check for a section name. When querying, the section_name
+   * MUST end in a forward slash. e.g.
+   * /Section1/
```
roystgnr: Kind of fragile, no?  Can we test for a trailing forward slash and add it internally if absent?

roystgnr:
That's my only nitpick.  :+1:

pbauman:
Addressed @roystgnr's concern and MooseBuild is passing. If a different strategy for handling out-of-source builds + CppUnit's unwillingness to give you argc/argv, I'll address @jwpeterson's concern. Merging.

Issue #436, opened 2015-01-14
Reduce geometric-based checks for nodes with adaptivity


permcody:
libMesh tends to collapse nodes that are geometrically close which interferes with XFEM work and or very small elements. We should attempt to use connectivity information in the decision process of determining when nodes are truly duplicates or not. 

Generation of nodes on parallel processors may still require geometric checks.

roystgnr:
@dknez just reminded me that we've got a third section of the code where we assume unique node geometry: the libHilbert ordering.

Hopefully @benkirk will correct me, but I believe the *_global_indices() methods will fail in the case where two nodes occupy the same point.

I don't see an easy fix.  My "bracketing nodes" idea in TopologyMap depends on the fact that we can use neighboring nodes for disambiguation... but neighboring node ids aren't canonical enough to give us a partition-independent ordering, and neighboring node Hilbert indices may *themselves* be duplicates in meshes with slits, and it's hard to define "neighboring" in the case of non-child elements in the first place.

We could get the set of active elements which own a node, take the average of their centroids, take the Hilbert key of that, then use *pairs* of Hilbert keys for uniqueness (and dictionary ordering for sorting).  It would be nice to have a solution that doesn't double the amount of communication required, though.

dschwen:
The Hilbert index is a key according to which the nodes are sorted before being partitioned. Having two nodes with the same HI should not necessarily result in a failure. I looked at the code a while ago and my memory is a bit foggy. The sorting of nodes with the same HI could be made unambiguous by including the node id in the comparison operation.

roystgnr:
For partitioning, the sorting just needs to be "good enough" to hand off to METIS etc.  A slight change here or there makes no difference.  I'm pretty sure having a duplicate Hilbert key currently leads to bugs (that was the proximate cause of the bug that I fixed in #316 IIRC), but we could just fix up the sorting algorithms to handle duplicate keys.

For partition_agnostic restarts, the node id is too ambiguous, though: if we write a mesh on N processors, then read it in on M, the id assigned to a node in the read won't necessarily match the id it had during the write.  We need a completely id-independent ordering to get away with that... and to avoid breaking this functionality in old restart files, it should be a *compatible* id-dependent ordering; i.e. we need to be sorting by (hilbert_key, something_else).

dschwen:
Oh, wait, duplicate HI may be a problem because only a list of HI sorted. What follows is a lookup of the nodes by their HI values. With duplicate HI this lookup may fail.

[...]

PR #482, opened 2015-02-27
Sync node by elem

roystgnr:
This isn't ready to pull, but I'd like to get MooseBuild started on testing.  This should take us much of the way towards solving #436 once it's working.

### DIFF
--- a/include/mesh/mesh_communication.h
+++ b/include/mesh/mesh_communication.h
@@ -35,7 +35,6 @@ namespace libMesh
 class MeshBase;
 class ParallelMesh;
 class BoundaryInfo;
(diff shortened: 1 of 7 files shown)


roystgnr:
Huh.  I wasn't actually expecting that to pass MooseBuild.  Usually ripping out and replacing critical libMesh guts *doesn't* work on the very first try.  I still want to do a lot more local testing before merging, though.

roystgnr:
This is passing all my tests now.  If anyone wants to give it a few runs (ParallelMesh AMR cases would be ideal) then let me know; otherwise I'll merge this afternoon.

PR #484, opened 2015-03-05
Default comm self

### LINKED ISSUE #483: --disable-mpi fails with triangle support
When configuring with --disable-strict-lgpl (for triangle support) and --disable-mpi, compilation fails with src/geom/elem_cutter.C:43:49: error: 'MPI_COMM_SELF' was not declared in this scope. It seems that this part of the code is not properly #ifdef'd.

roystgnr:
Default-construct Communicator objects with MPI_COMM_SELF, so they're at least good for *something*, instead of MPI_COMM_NULL.

Use that Communicator rather than explicitly referring to MPI (which may not be enabled!) in ElemCutter.

Hopefully this fixes #483

### DIFF
--- a/include/geom/elem_cutter.h
+++ b/include/geom/elem_cutter.h
@@ -157,7 +157,7 @@ protected:
   AutoPtr<SerialMesh> _inside_mesh_3D;
   AutoPtr<SerialMesh> _outside_mesh_3D;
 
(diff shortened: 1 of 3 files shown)


roystgnr:
I'll merge this if/when @rvetter reports his compilation problems fixed.

rvetter:
Works for me, thanks.

PR #487, opened 2015-03-05
Check vector size before getting first element

### LINKED ISSUE #485: FParser debug mode container subscript check fail
It turns out that 

```
return (*compiledFunction)(Vars, &(this->mData->mImmed[0]), Epsilon<Value_t>::value);
```

is a pretty dumb way of getting a pointer to data in a vector. If the vector is empty subscripting it with ```[0]``` (first element) will trigger an error if debug containers are used. 

I need to fix this.

dschwen:
Confirmed to fix failing moose PR https://github.com/idaholab/moose/pull/4777

Closes #485.


jwpeterson:
Merged in 79359dc.

PR #490, opened 2015-03-05
Re-run JITCompile after Optimize

### LINKED ISSUE #486: JIT needs to be rerun after optimize
I'll have to overwrite the ```Optimize``` method in FParserAD to call ```JITCompile``` again if the Parser object had been JIT compiled, as optimization can change the layout of immediate variables in the Parser, which the compiled code relies on.

dschwen:
Wrap calls to Optimize and re-run JITCompile if necessary. This makes FParser objects more robust w.r.t. order of Optimize and JITCompile calls.

Makes #487 obsolete.
Closes #486.


jwpeterson:
Merged in 79359dc

PR #491, opened 2015-03-07
Build unit_tests-dbg as noinst at least

roystgnr:
Even if some systems can't run it correctly due to cppunit ABI
incompatibility, we can always at least build it, and thereby have it
available for manual testing.  We just can't safely make it part of
"make check".

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -38,6 +38,7 @@ if LIBMESH_ENABLE_FPARSER
 endif
 
 check_PROGRAMS = # empty, append below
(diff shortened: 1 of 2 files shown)


roystgnr:
One guess as to whether this is an "out of the goodness of Roy's heart" pull request or a "failing unit test in another branch" pull request...

roystgnr:
IIRC MooseBuild is using a dbg-incompatible cppunit ABI, so that's working.  I've verified that the noinst build was done locally.  This doesn't touch anything besides the unit tests build, so should be safe to merge.

Issue #433, opened 2015-01-14
Adaptivity on lower dim elems in higher dim space


permcody:
libMesh has a bug that prevents us from using adaptivity with lower dimensional elements in a mesh with a higher dimension number.

Currently we us a workaround where we set the mesh dim == elem dim even if the elements aren't limited to the lower dimensional space. The issue stems from inspecting the wrong dim value in the adaptivity routines.

Cross-referencing: https://github.com/idaholab/moose/issues/4530

pbauman:
@permcody - could you elaborate a bit? @cahaynes and myself will care about this very much soon. In particular, were asserts tripped or was it different failure modes? I scanned MeshRefinement, but nothing stuck out immediately (I didn't try very hard though) - any pointers? @cahaynes and I can chip in on this.

friedmud:
Asserts get tripped. It's in the EquationSystems::reinit() path...

pbauman:
Thanks for the pointer @friedmud. Do you guys have a running test that catches this? I'm curious if #442 maybe helped - there was [two spots](https://github.com/pbauman/libmesh/commit/ef72864f5feebd2fe79265c3c5f2c88314836865) where a `_dim` got changed to `elem->dim()` because it was the right thing.

permcody:
@andrsd was the one that ran into this. I'll work with him to get a test case set up.

pbauman:
No rush. If we hit it before y'all have a chance, I'll let you know.

roystgnr:
So here's a question for everyone: when you're mixing lower and higher dimensional elements, is it ever possible for the centroids of two different elements to overlap?  If the answer is anything other than an unqualified "no" then there's some LocationMap-based internals that I'm going to need to rewrite...

jwpeterson:
On Mon, Mar 2, 2015 at 2:30 PM, roystgnr <notifications@github.com> wrote:

> So here's a question for everyone: when you're mixing lower and higher
> dimensional elements, is it ever possible for the centroids of two
> different elements to overlap? If the answer is anything other than an
> unqualified "no" then there's some LocationMap-based internals that I'm
> going to need to rewrite...
>

I can imagine a layer of Quads cutting perfectly through the middle of a
layer of Hexes... so not an unqualified "no", but I wonder if this a
realistic use case.

-- 
John

pbauman:
[...]

PR #492, opened 2015-03-09
Use topology map instead of location map

roystgnr:
This isn't ready to merge yet: there's still one test failing with ParallelMesh for me, I haven't hit it with everything I've got yet, and even after all tests pass I'll want third parties to have a chance to hammer it too.  This stabs through the libMesh adaptivity guts deeply enough that we'll need to be paranoid about whether or not I've hit any vital organs.

Basically, this PR rips out our last LocationMap-based node lookups.  When creating a new Node during refinement, we still check for any existing Node in the proper place, but we now define "proper place" via element topology, not via node location.

Hopefully this is the last fix we need to close out #436.

I'm trying to take care that we haven't regressed functionality in the #433 cases where lower- and higher-dimensional elements share nodes.  If your node sharing and refinement flagging are compatible, e.g. if you have an Edge3 sharing three (collinear on the master element) nodes on a Hex27 and you refine both, then our policy (implemented via this PR) will be that the child edges' nodes will all be shared with the child hexes.  If your refinement flags are not compatible, e.g. you want constraint equations for elements of different dimension which share nodes, our policy will be that you need to add those constraints manually.  If your parent nodes are not shared in a compatible way, e.g. an Edge3 whose nodes aren't collinear on the master hex, our policy will be that I hate you.


roystgnr:
I'm now getting good results with SerialMesh, and if I'm correctly starting to understand the remaining ParallelMesh bug, it really is ParallelMesh-interaction-specific, not just a corner case that happened to be triggered by ParallelMesh.

I'd appreciate it if other SerialMesh AMR users would start hammering this PR with their apps now.

permcody:
Acknowledging the request. We are all working on a sprint problem and then
heading to SIAM next week. It may be a few weeks until we can do any
testing on this. This will be a very nice capability so we are looking
forward to trying it out.

Cody

roystgnr:
Looks like it's not just a ParallelMesh-only bug, it also requires --enable-nodeconstraint (or --enable-everything) to trigger.  So people with ParallelMesh-compatible AMR codes should throw those at it too.

We've got some time to play.  I'd like to merge this once I've fixed the ParallelMesh+NodeConstraints bug and everybody feels confident, but we won't be able to make full use of the duplicate node support until I've also figured out what to do with libHilbert duplicate keys.

roystgnr:
This is working for me now.  Tentative plan is to merge it a week from Monday unless I hear screams first.

roystgnr:
I'm leaving some unused code in the commit history, by the way.  I've rebased away (almost?) all the cases of "wrote code, then deleted it when it wasn't working", but there's a large case of "wrote complicated special case code, wrote general code for other cases, decided that special case code was less trustworthy even though it's probably working" that I'm leaving in just in case I change my mind again later.

PR #494, opened 2015-03-11
Free malloced digest buffer

dschwen:
There is a small memory leak in my FParser code. The sha1 digest buffer is allocated via ```malloc``` in the sha1 library. It was up to me to free it. Done now.


jwpeterson:
I love C.

roystgnr:
Little known fact: in early drafts of "Terminator", it was explained that Skynet went haywire because of a bug in some low-level C code.  This was removed from the script when it was discovered that half of the audience wouldn't understand and the other half would reach the same conclusion without prompting.  :+1:

PR #495, opened 2015-03-11
Make target_bin_size a member variable of PointLocatorTree, instead of using magic numbers.

dknez:


roystgnr:
:+1:, presuming MooseBuild is happy.

PR #496, opened 2015-03-12
Need to increase bin size further for point_locator_tree

dknez:
I have another mesh (this one is quite reasonable) which hits the same infinite loop as yesterday with a bin size of 250.

As a result, I propose we increase the default bin size to 300 (and work towards a more reliable fix for this issue).

### DIFF
--- a/src/utils/point_locator_tree.C
+++ b/src/utils/point_locator_tree.C
@@ -40,7 +40,7 @@ PointLocatorTree::PointLocatorTree (const MeshBase& mesh,
   _tree            (NULL),
   _element         (NULL),
   _out_of_mesh_mode(false),
-  _target_bin_size (250),
+  _target_bin_size (300),
   _build_type(Trees::NODES)
 {
   this->init(_build_type);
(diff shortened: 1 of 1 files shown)


jwpeterson:
> As a result, I propose we increase the default bin size to 300 (and work towards a more reliable fix for this issue).

What are the drawbacks to making the bin size larger?  Longer linear searches or something?

dknez:
Yep. It relates to how often a TreeNode gets refined into N children. The description from tree_node.h is: 

  /**
   * The maximum number of things we should store before
   * refining ourself.
   */
  const unsigned int tgt_bin_size;

So a smaller bin size gives a deeper tree, and hence a shorter search in TreeNode<N>::find_element().

But I've found two cases where refine() triggers an infinite loop, so setting target_bin_size to 300 to avoid that seems reasonable to me for now.

dknez:
Mind if I merge this?

jwpeterson:
> But I've found two cases where refine() triggers an infinite loop, so setting target_bin_size to 300 to avoid that seems reasonable to me for now.

OK.  I can't download files from dropbox at work, so I didn't get a chance to check out the previous mesh you sent, but I'll take a look tonight (or if you can google drive them, that would be great).  I'm wondering if there's some simple metric one could compute on the mesh that would give an estimate of the target bin size... something like number of elements touching a point?

roystgnr:
I could set up a failing mesh with no more than 4 elements touching the same point, with that "diagonal anisotropy" trick.  We could do some on-the-fly tests, though: refine a tree node only if all of the child nodes have at least X% fewer elements than the parent?

PR #499, opened 2015-03-16
Fixed bug in Tree code.

dknez:
There was a bug in TreeNode in the case that we have 2D elements that aren't in the xy-plane. Should use LIBMESH_DIM instead of elem->dim() to get the number of non-zero Point coordinates.


jwpeterson:
Looks good... thanks for finding this bug!

dknez:
@jwpeterson you're happy if I go ahead and merge then? Should we squash into one commit?

jwpeterson:
Oh yeah, please squash, didn't notice it was 2 commits.

PR #500, opened 2015-03-16
mesh_dimension() -> elem_dimensions() fixes

roystgnr:
There's still a lot of libMesh code that assumes we don't mix elements of different dimension, but this should fix the worst offenders.


pbauman:
Looks like a nice improvement to me. Thanks.

PR #503, opened 2015-03-19
Avoid throwing exceptions in backtrace

roystgnr:
I just had an application go into an infinite loop when a backtrace was
called before the command line had been set up.


jwpeterson:
:+1:

PR #506, opened 2015-03-24
Make Metis datatypes consistent with libmesh's

jwpeterson:
This branch ensures that the sizes of METIS's `idx_t` and `real_t` are consistent with libmesh's `dof_id_type` and `Real`, respectively.  It also applies some patches from [PETSc's parmetis distro](http://ftp.mcs.anl.gov/pub/petsc/externalpackages/parmetis-4.0.3-p1.tar.gz) pointed out by @jedbrown.


jwpeterson:
I've also tested this against a PETSc configured with
```
--download-parmetis=1
--download-metis=1
--with-64-bit-indices
```

and libmesh configured with:
```
--with-metis=PETSc
--with-dof-id-bytes=8
```

and it seems to work.

include/partitioning/parmetis_partitioner.h
```diff
@@ -111,23 +114,10 @@ class ParmetisPartitioner : public Partitioner
   vectormap<dof_id_type, dof_id_type> _global_index_by_pid_map;
 
   /**
-   * Data structures used by ParMETIS to describe the connectivity graph
-   * of the mesh.  Consult the ParMETIS documentation.
+   * Pointer to the Parmetis-specific data structures.  Lets us avoid
+   * including parmetis.h here.
    */
```
roystgnr: Appreciate this

roystgnr:
I reserve the right to scream when this causes some strange breakage for me.

But as far as I can tell from the diffs this should do nothing but avoid strange breakage.  :+1: to merge when you're ready.

jwpeterson:
> I reserve the right to scream when this causes some strange breakage for me.

OK, thanks for reviewing.

PR #507, opened 2015-03-26
Adding MeshBase::bid_nodes_{begin,end}

jwpeterson:
This allows for iteration over nodes with a user-selected boundary id.

@andrsd

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -857,6 +857,13 @@ public:
   virtual const_node_iterator bid_nodes_begin    (const boundary_id_type bndry_id) const = 0;
   virtual const_node_iterator bid_nodes_end      (const boundary_id_type bndry_id) const = 0;
 
+  /**
+   * Iterate over nodes for which BoundaryInfo::n_boundary_ids(node) > 0.
+   */
+  virtual node_iterator bnd_nodes_begin () = 0;
+  virtual node_iterator bnd_nodes_end () = 0;
+  virtual const_node_iterator bnd_nodes_begin () const = 0;
+  virtual const_node_iterator bnd_nodes_end () const = 0;
 
   /**
    * Return a writeable reference to the whole subdomain name map
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -829,33 +829,44 @@ public:
 
 
   /**
-   * non-const Node iterator accessor functions.
+   * Iterate over all the nodes in the Mesh.
    */
-  virtual node_iterator nodes_begin        () = 0;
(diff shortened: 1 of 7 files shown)


permcody:
Cool, Are there any places where we might need to take advantage of this in
MooseMesh or BoundaryRestrictable?

andrsd:
@jwpeterson Nice!
@permcody My intended use was for transfers. Nearest-node transfer and projection transfers will benefit from this (maybe others as well)

jwpeterson:
On Thu, Mar 26, 2015 at 10:40 AM, Cody Permann <notifications@github.com>
wrote:

> Cool, Are there any places where we might need to take advantage of this in
> MooseMesh or BoundaryRestrictable?
>

Looks like right now we just use the BoundaryInfo object
in MooseMesh::buildNodeList() to create our vector of BndNode* and BndElem*
objects ahead of time.  I don't know about BoundaryRestrictable, this was
mostly for David's transfer thing as he mentioned.

-- 
John

permcody:
Yes, I knew what it was for. Just wanted to see if there was potential for
clean up anywhere else as a result

aeslaughter:
I don't think this will be needed in the restrictable interfaces, since
there is no looping. But, I think some of the MeshModifiers could benefit
from this.

jwpeterson:
It should be relatively easy to add one more type of iterator that loops over *all* boundary nodes rather than just the ones with a specified ID, so I'm probably going to tack that onto this PR as well...

include/mesh/mesh_base.h
```diff
@@ -852,6 +854,8 @@ class MeshBase : public ParallelObject
   virtual const_node_iterator local_nodes_end    () const = 0;
   virtual const_node_iterator pid_nodes_begin    (const processor_id_type proc_id) const = 0;
   virtual const_node_iterator pid_nodes_end      (const processor_id_type proc_id) const = 0;
+  virtual const_node_iterator bid_nodes_begin    (const boundary_id_type bndry_id) const = 0;
```
roystgnr: This is probably a good place to break our long-standing "don't properly document iterator range methods" tradition.  I'd have assumed that the behavior here was "iterate through nodes that are on sides with the given boundary id", but that only happens if the nodes themselves also share that id, right?  We should probably clarify that with a comment.

jwpeterson: It only iterates through nodes for which 

I agree that all these iterator methods should probably be documented.  I was thinking about reorganizing them so the const and non-const versions are together and we can put one 
```
/**
*
*/
```
section that covers all 4.

roystgnr:
Looks good, thanks!

PR #508, opened 2015-03-27
Refactor projection code

roystgnr:
This refactoring is only about a third complete, but the first phase of the work is passing all my tests, and I think (assuming I didn't screw up the conflict resolutions when synching up with master this morning) it's in a state that's ready for merge.

The final goal is to replace *all* the cut-and-pasted-and-modified projection code in libMesh with a single algorithm that gets specialized via a few functor objects, and then to add *more* projection codes that use their own functor objects to allow capabilities like projection matrix assembly.

The current state is that only a couple of the projection methods have been replaced, although even at this state I've managed to remove more code than I've added.

The changes *should* all be under the hood at this point (although I'll eventually make the functor classes public so they can be used by https://github.com/idaholab/moose/blob/devel/framework/src/ics/InitialCondition.C#L84 and others) but even now it's a sweeping enough change that it deserves careful testing.


dknez:
Sounds great. I'll try it out when I get a chance.

jwpeterson:
> Conflicts:
>	include/libmesh/Makefile.am
>	src/systems/system_projection.C

There's that one super-long line of header file names that always conflicts for me.  I can't remember if I've posted it before, but I use the command:

```
GIT_PAGER='' git diff --word-diff :1:include/libmesh/Makefile.am :3:include/libmesh/Makefile.am
```

during conflict resolution to isolate the part of the long line that is actually different, and then incorporate the change manually into the "HEAD" version.  I'm curious, what conflicted in system_projection.C?

roystgnr:
I'd branched from a pre-UniquePtr version of master, so your "make system_projection.C use UniquePtr" changes conflicted with my "delete large swaths of system_projection.C" changes.  Hmm... looks like I missed a few UniquePtr changes when resolving, too.  I'll try and fix that now.

jwpeterson:
If your branch is that old, you might consider just rebasing it instead of the "merge before you merge" approach.

roystgnr:
I considered that.  But I figured that if I screwed up the merge conflicts, then I've at least got tested commits to fall back on while figuring out what went wrong.  If I'd screw up rebasing conflicts (which there were; that was the first thing I tried) then I'd have written a new git history with nothing but broken commits.

jwpeterson:
> then I'd have written a new git history with nothing but broken commits.

Well, yeah keep the pre-rebased brach around in case of failures by all means.

roystgnr:
Keep it around for how long?  We've had a couple nasty cases (*cough*, libHilbert, *cough*) where breakage was so subtly restricted to corner cases as to go undiscovered for several months.  There's definitely some risk of that here; one bugfix (which ironically is now invisible due to an interactive rebase...) was only prompted when a GRINS test hit a code path that libMesh examples (and IIRC MOOSE and FIN-S and my private apps) don't touch.

jwpeterson:
> Keep it around for how long?

Just as long as it takes to make sure you didn't screw up conflict resolution.

> We've had a couple nasty cases (cough, libHilbert, cough) where breakage was so subtly restricted to corner cases as to go undiscovered for several months. There's definitely some risk of that here; one bugfix (which ironically is now invisible due to an interactive rebase...)

Not asking you to squash anything here (sorry if you squashed an important bugfix that should have remained independent, that is pretty annoying).  The main reason I prefer to rebase branches from ages ago is that it helps with the output of `git log --graph` when you don't have to scroll down 3 screens where nothing has happened...

roystgnr:
Oh, no, it wasn't an important bugfix that should have remained independent; I'd have put something like that in a separate PR, not just a separate commit.  (although I admit I'm sometimes lazy enough to put multiple *features* in separate commits in one PR - e.g. #509 got pulled out of an unrelated branch, but 0f6a2cd could have also stood on its own yet I left it lumped in here).

The original commit record was "buggy code A, more code B, more code C, bugfix D, more code E...", so I just reordered and fixup-ed to give "fixed code A+D, more code B, more code C, more code E...".  This wasn't supposed to be an example of the horrors of rebasing (I love reorder+fixup in private branches, although IMHO in a perfect world we could choose to retain the original history while adding a "fixed up history" as metadata or vice-versa).  It was rather another example of the difficulty of safely assuming that I haven't introduced a bug just because all the tests I've thrown at it so far are passing.

jwpeterson:
> We can’t automatically merge this pull request.

Will you now merge again before you merge?  I think I draw the line at one merge from master onto a topic branch :-)

roystgnr:
Yeah, two merges from master would be excessive.  I'll get rid of the existing merge and try my hand at the rebase again, but on a new branch.

roystgnr:
Obsoleted by #511

PR #509, opened 2015-03-27
Reduce Makefile.am conflicts

roystgnr:
Ironically, this PR will conflict with the Makefile.am changes in #508 and possibly others.

But hopefully in the future we will have fewer PRs with that problem.  By building include/libmesh/Makefile.am with headers separated via escaped-newlines instead of spaces we can keep more header-adding diffs small enough to merge without conflict.

I *didn't* forget to run bootstrap here; automake reformats variables in Makefile.in anyway, and since I kept the headers in the same order they ended up identical in Makefile.in.

Makefile.in conflicts from newly-added headers are usually unavoidable due to that reformatting; the only complete solution would be to stop distributing it and require our non-release users to have autotools installed.


jwpeterson:
Nice!

include/libmesh/rebuild_makefile.sh
```diff
@@ -30,16 +30,22 @@ cat <<EOF > Makefile.am
...
 EOF
 
+echo -n "BUILT_SOURCES =" >> Makefile.am
+for built_src in $built_sources ; do
+    echo " \\" >> Makefile.am
+    echo -n "        "$built_src >> Makefile.am
+done
+
+echo >> Makefile.am
+echo >> Makefile.am
```
jwpeterson: Does empty shell echo just make a newline?

roystgnr: Yup.  Seems to be the shortest completely-portable way to do it.  There's a bash/zsh/ksh idiom for simply embedding \n in a string, which even seems to work in dash (/bin/sh on Debian) but it's not POSIX compliant and I'd hate to have our stuff break as soon as someone decides to make their sh even stricter.

PR #511, opened 2015-03-27
Refactor projection code

roystgnr:
Rebased and slightly edited version of #508.  I'll probably merge this mid next week or whenever @jwpeterson finishes reviewing, whichever comes last.

---

This refactoring is only about a third complete, but the first phase of the work is passing all my tests, and I think (assuming I didn't screw up the conflict resolutions when synching up with master this morning) it's in a state that's ready for merge.

The final goal is to replace all the cut-and-pasted-and-modified projection code in libMesh with a single algorithm that gets specialized via a few functor objects, and then to add more projection codes that use their own functor objects to allow capabilities like projection matrix assembly.

The current state is that only a couple of the projection methods have been replaced, although even at this state I've managed to remove more code than I've added.

The changes should all be under the hood at this point (although I'll eventually make the functor classes public so they can be used by https://github.com/idaholab/moose/blob/devel/framework/src/ics/InitialCondition.C#L84 and others) but even now it's a sweeping enough change that it deserves careful testing.


jwpeterson:
:+1:

Issue #515, opened 2015-03-30
std::hash is C++11 only


jwpeterson:
This new code in topology_map.h doesn't compile on GCC 4.6.3 with `--disable-cxx11` however it does seem to work with the same config flags on clang 3.5.0.  I'll add a test for this, I'm envisioning something like `LIBMESH_BEST_HASH` in the vein of `LIBMESH_BEST_UNORDERED_MAP`...

```
struct myhash {
public:
  template <typename T1, typename T2>
  std::size_t operator()(const std::pair<T1, T2> &x) const
    {
      // recommendation from
      // http://stackoverflow.com/questions/5889238/why-is-xor-the-default-way-to-combine-hashes
      return 3 * std::hash<T1>()(x.first) + std::hash<T2>()(x.second);
    }
};
```

jwpeterson:
We already do this test in unordered.m4.  Just need to use it.

PR #517, opened 2015-03-30
Call ACX_BEST_HASH, use LIBMESH_BEST_HASH in code

jwpeterson:
I've tested this with both C++11 (std::hash) and C++03 (std::tr1::hash) compilers, and it seems to work as expected.  If it passes moosebuild, I will merge it.

#515

### DIFF
--- a/configure
+++ b/configure
@@ -27362,6 +27362,205 @@ fi
 
   fi
 
+# Determine which of std::hash, std::tr1::hash, or __gnu_cxx::hash is available
+
+{ $as_echo "$as_me:${as_lineno-$LINENO}: checking whether the compiler supports std::hash" >&5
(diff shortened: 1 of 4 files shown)


roystgnr:
Did you test any gcc version using GCC hash_map ?  In that case I believe __gnu_cxx::hash is what we need to override.

(or did we ever support ext/hash_map.h in the first place?...)

jwpeterson:
> Did you test any gcc version using GCC hash_map ? In that case I believe __gnu_cxx::hash is what we need to override.

I don't have a compiler that requires __gnu_cxx::hash, but the tests in unordered.m4 do cover that case.

(Sorry if you are seeing this response twice, I replied to the email but it didn't end up here on the PR for some reason.)

PR #518, opened 2015-03-31
Another ACX_MPI fallback

roystgnr:
We currently end up in an inconsistent state when configuring against the repo libpetsc3.4.2-dev on Ubuntu 14.10 - PETSc is there, and configured with MPI, but doesn't define MPIEXEC, so we assume it's configured in serial, and we don't try to detect MPI.

This PR fixes the configure for me, by trying ACX_MPI in that case.  It's possible that this PR will *break* cases in which the user has an autodetectable MPI but wants to configure against a mpiuni PETSc, but in those cases "--with-mpi=false" ought to be an adequate workaround.

### DIFF
--- a/configure
+++ b/configure
@@ -809,17 +809,17 @@ SLEPC_DIR
 LIBMESH_ENABLE_PETSC_FALSE
 LIBMESH_ENABLE_PETSC_TRUE
 enablepetsc
+PETSC_FC_INCLUDES
+PETSC_CC_INCLUDES
+PETSCINCLUDEDIRS
+PETSCLINKLIBS
 MPI_INCLUDES_PATHS
 MPI_INCLUDES_PATH
 MPI_LIBS_PATHS
 MPI_LIBS_PATH
 MPI_LIBS
-MPI
 MPI_IMPL
-PETSC_FC_INCLUDES
-PETSC_CC_INCLUDES
-PETSCINCLUDEDIRS
-PETSCLINKLIBS
+MPI
 petscmajorminor
 petscmajor
 petscversion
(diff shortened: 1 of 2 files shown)


roystgnr:
Forgot to mention: thanks go to Zhenyu (more id omitted since I don't know his github username and I want to spare his email address from spammers) for finding this problem.  I'm running on Ubuntu 14.10 myself but haven't been using the default repo PETSc for a long time.

jwpeterson:
So are we wrong to be grepping for `MPIEXEC` in petscvariables as a way to snoop MPI?  It looks like we've been doing it that way since time immemorial...

I can definitely see why Ubuntu can't simply tie its PETSc dpkg to a single mpiexec since it has no idea what a user might install...

jwpeterson:
I don't suppose Ubuntu's petscvariables file provides `MPI_INCLUDE`, `MPICXX_SHOW`, or `MPICC_SHOW` variables either?

roystgnr:
Nothing in petscvariables, or in the rest of /usr/lib/petsc/conf/* as far as grep can tell.

roystgnr:
I can't imagine that merely omitting config variables produces an MPI-independent PETSc, though.  IIRC the MPI standard doesn't specify enough of an ABI to make that possible.  I guess I'll try swapping between openmpi and mpich2 and find out for myself, though.

roystgnr:
Trying to use Ubuntu PETSc with mpich2 ends up adding mpich2 compile/link flags via ACX_MPI autodetection *and* openmpi compile/link flags via PETSCINCLUDEDIRS/PETSCLINKLIBS.  I'm not going to bother testing *that* case.  This works for the Ubuntu PETSc+openmpi case, though, so ready to merge.

PR #519, opened 2015-03-31
Fix petsc version check in update_preallocation_and_zero()

roystgnr:
This fixes compilation errors for me with PETSc 3.4.2.

It's built on top of the Ubuntu-PETSc configure error fixes in #518


roystgnr:
Huh.  That was a lot of wasted minutes on bitbucket in lieu of spending 5 seconds remembering to "git pull".

PR #521, opened 2015-04-01
fix compilation with --enable-unique-ptr

capitalaslash:
some leftovers after the latest merges.
this fixes compilation for me using `--enable-unique-ptr`.


roystgnr:
Thanks!  :+1:

PR #523, opened 2015-04-01
Shape function second derivatives on non-affine elements

jwpeterson:
This is a very old branch that I created a long time ago with @roystgnr and @pbauman but never quite got merged into master.  Here's a basic [description](https://drive.google.com/file/d/0B9BK7pg8se_iWm9XenJSd0p5Z0k/view?usp=sharing) of what was implemented back then.

I have now tested it fairly extensively using this [code](https://gist.github.com/jwpeterson/a085afd592feca73f81f) and am pretty confident it's working.  The main idea of the test code is to verify the second derivatives on distorted elements by central differencing the first derivatives and ensuring that the result converges at second order in the differencing parameter, epsilon.  An example of the convergence results for the Pyramid14 element is available [here](https://drive.google.com/file/d/0B9BK7pg8se_iREN2djg5blNpVmc/view?usp=sharing).  Similar convergence results hold for the other geometric element types.

### DIFF
--- a/include/fe/fe_map.h
+++ b/include/fe/fe_map.h
@@ -783,6 +783,13 @@ protected:
    * Jacobian*Weight values at quadrature points
    */
   std::vector<Real> JxW;
+
+private:
+  /**
+   * A helper function used by FEMap::compute_single_point_map() to
+   * second derivatives of the inverse map.
+   */
+  void compute_inverse_map_second_derivs(unsigned p);
 };
 
 }
--- a/include/fe/fe_map.h
+++ b/include/fe/fe_map.h
@@ -266,6 +266,26 @@ public:
   const std::vector<Real>& get_dzetadz() const
   { return dzetadz_map; }
 
+#ifdef LIBMESH_ENABLE_SECOND_DERIVATIVES
+  /**
+   * Second derivatives of "xi" reference coordinate wrt physical coordinates.
+   */
+  const std::vector<std::vector<Real> >& get_d2xidxyz2() const
+  { return d2xidxyz2_map; }
+
+  /**
+   * Second derivatives of "eta" reference coordinate wrt physical coordinates.
+   */
+  const std::vector<std::vector<Real> >& get_d2etadxyz2() const
+  { return d2etadxyz2_map; }
+
+  /**
+   * Second derivatives of "zeta" reference coordinate wrt physical coordinates.
+   */
+  const std::vector<std::vector<Real> >& get_d2zetadxyz2() const
+  { return d2zetadxyz2_map; }
+#endif
+
   /**
    * @returns the reference to physical map for the side/edge
    */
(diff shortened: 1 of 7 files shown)


roystgnr:
Wow, thanks, I'd forgotten about this!  :+1:

Would it be worth testing has_affine_map() to see if we can skip the more expensive parts of the calculation in those cases?

pbauman:
> Would it be worth testing has_affine_map() to see if we can skip the more expensive parts of the calculation in those cases?

Was thinking the same thing.

Also, thanks @jwpeterson! I'll try and give this a whirl on some stabilized problems tomorrow, see if any magical convergence improvements happen.

jwpeterson:
> On Apr 1, 2015, at 9:19 PM, roystgnr <notifications@github.com> wrote:
> 
> Wow, thanks, I'd forgotten about this!  
> 
> Would it be worth testing has_affine_map() to see if we can skip the more expensive parts of the calculation in those cases?
> 
Yes, this should be doable.  It will be a little while until I can get to it though...

jwpeterson:
> Would it be worth testing has_affine_map() to see if we can skip the more expensive parts of the calculation in those cases?

I think Roy from the deep past has already handled this case for us.  If `elem->has_affine_map()`, we call `FEMap::compute_affine_map()` which takes care of sizing and zero'ing all the appropriate vectors.  Then `FEMap::compute_single_point_map()` (and hence the expensive second derivative calculations) is never called at all.

Now, you might want me to also optimize the bits of the `H1FETransformation<OutputShape>::map_d2phi()` function which are just multiplying/adding zeros on affine elements but I'd argue it's cleaner the way it currently is (does not need to be passed an `Elem*`), might not get all that much faster, and I'd be likely to introduce bugs if I tried to split up the terms into affine/not affine, especially in the 3D case...

That being said, I did clean up a few little things that I saw while reviewing the code and pushed that, as far as I'm concerned this is ready to merge.

pbauman:
> I think Roy from the deep past has already handled this case for us. If elem->has_affine_map(), we call FEMap::compute_affine_map() which takes care of sizing and zero'ing all the appropriate vectors. Then FEMap::compute_single_point_map() (and hence the expensive second derivative calculations) is never called at all.

OK, cool. For some reason, glancing at the diffs I thought we'd still walk this code path for affine elements.

> Now, you might want me to also optimize the bits of the H1FETransformation<OutputShape>::map_d2phi() function which are just multiplying/adding zeros on affine elements but I'd argue it's cleaner the way it currently is (does not need to be passed an Elem*), might not get all that much faster, and I'd be likely to introduce bugs if I tried to split up the terms into affine/not affine, especially in the 3D case...

OK, I'm good with this. There are certainly lower hanging fruit for optimization (looks at shape function index ordering... hopefully one of my Ph.D. students will do this for the HPC class project in the fall) and if I ever suspect this to be an issue, I can always hack on it and test and we can update then. 

> as far as I'm concerned this is ready to merge

Thanks for your work on this. I don't suppose we could get your test code lumped into the CPPUnit suite? Otherwise, I've been using this branch the past few days with some stabilization schemes (where I actually need the second derivatives) and everything seems fine.

roystgnr:
Roy-from-the-deep-past is smarter than me, so if he took care of the optimization then I won't worry about it.  :-)

As far as the remaining low hanging fruit for optimization goes; I'd put my money on vectorization of ```FE<*>::shape*()```.

PR #527, opened 2015-04-03
Cleaned up some of the reduced basis code.

dknez:
- Added RBConstruction::solve_for_matrix_and_rhs so that we can more easily mix and match solvers
and matrices. This resulted in a simplification to how we call the solvers, and how we set
reuse_preconditioner
- Removed a bunch of old unused code related to constraints (e.g. matrix that imposes incompressibility).
This code wasn't well tested anymore. If someone wants to add it back in, better to put it in a subclass
of RBConstruction.

### DIFF
--- a/examples/reduced_basis/reduced_basis_ex5/run.sh
+++ b/examples/reduced_basis/reduced_basis_ex5/run.sh
@@ -8,7 +8,7 @@ example_name=reduced_basis_ex5
 
 example_dir=examples/reduced_basis/$example_name
 
-options="-online_mode 0 "
+options="-online_mode 0 -ksp_type cg"
 run_example "$example_name" "$options"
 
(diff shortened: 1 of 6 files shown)


dknez:
This only touches the reduced basis code, not sure why MOOSE tests failed?

pbauman:
> This only touches the reduced basis code, not sure why MOOSE tests failed?

FYI, looks like reduced_basis_ex5 failed.

dknez:
@pbauman thanks for pointing that out. I added a solver convergence assertion that failed for reduced_basis_ex5 with the default solver (it works fine with a direct solver, for example). I just turned off the assertion, so it should pass now.

dknez:
Actually, no, there is something wrong with that example... I'll need to look at it more carefully.

Issue #525, opened 2015-04-03
UniquePtr and libMesh namespace


francesco-ballarin:
Dear libmesh developers,
I saw the recent merge of the UniquePtr branch (PR #357) and now I would like to update my code with the new UniquePtr instead of AutoPtr.
Unfortunately, in my code I always use libMesh::AutoPtr, therefore I think a simple replace AutoPtr -> UniquePtr would not work, because of the current definition of UniquePtr as a macro.
For instance, in the disable-unique-ptr case, I guess libMesh::UniquePtr would be expanded to libMesh::libMesh::AutoPtr.

Even though I can of course replace in my code all libMesh::UniquePtr with UniquePtr, this would kind of defy the purpose of having the libMesh:: qualification on all the other libMesh-related objects of my code.

I created a fork with a patch, starting from the comments in the commits, which may provide a possible fix.
You can find it at https://github.com/francesco-ballarin/libmesh/commits/unique_ptr_namespace
I provide here a small summary of the patch, if you are willing to consider it at least for a discussion.

The patched auto_ptr.h is basically:

```
#if have c++11
use template aliases to define UniquePtr in the libMesh namespace
#elif have c++03
just copy the boost::unique_ptr implementation with the libMesh::UniquePtr name
#else // for instance, if disable-unique-ptr is on
just copy the old AutoPtr implementation with the new UniquePtr name
and provide a
#define AutoPtr UniquePtr
for backward compatibility
#endif
```

If this is helpful, I can (try to) create a P.R. However:
0. please check if this is the backward compatibility that you have in mind (there is no AutoPtr class in the enable-unique-ptr case)
1. in the #if we should also test if template aliases are supported by the compiler. I do not know how to do that.
2. contrib/unique_ptr/unique_ptr.hpp has been modified replacing the boost namespace with the libMesh one. After this modification, I wonder if it would be better to just move the content of this file inside auto_ptr.h (or move the file unique_ptr.h in the libMesh includes, rather then in the contrib folder).

Best regards,
Francesco

roystgnr:
John, why use a macro rather than a typedef?

roystgnr:
Oh, because template typedefs didn't make it into the language until C++11, of course...

jwpeterson:
> On Apr 3, 2015, at 7:43 AM, roystgnr <notifications@github.com> wrote:
> 
> Oh, because template typedefs didn't make it into the language until C++11, of course...
> 

[...]

PR #529, opened 2015-04-05
UniquePtr and libMesh namespace

francesco-ballarin:
As suggested by Cody Permann in the discussion of issue #525, I am opening a pull request with a (possible, still untested) unique_ptr refactoring that is compatible with the libMesh namespace.
Please see the issue #525, items 0., 1. and 2. in my original post, and the discussion in the following comments.

### DIFF
--- a/contrib/unique_ptr/unique_ptr.hpp
+++ b/contrib/unique_ptr/unique_ptr.hpp
@@ -1,5 +1,5 @@
 ///////////////////////////////////////////////////////////////////////////////
-// unique_ptr.hpp header file
+// UniquePtr.hpp header file
 //
 // Copyright 2009 Howard Hinnant, Ion Gazta&ntilde;aga.
 // Distributed under the Boost Software License, Version 1.0. (See
@@ -13,18 +13,18 @@
 //   reference http://www.open-std.org/jtc1/sc22/wg21/docs/lwg-active.html
 //   for any pending issues against this specification.
 
-#ifndef UNIQUE_PTR_HPP
-#define UNIQUE_PTR_HPP
+#ifndef HINNANT_UNIQUE_PTR_HPP
+#define HINNANT_UNIQUE_PTR_HPP
 
 #include <boost/utility/enable_if.hpp>
 #include <boost/type_traits.hpp>
 #include <boost/static_assert.hpp>
 #include <boost/mpl/if.hpp>
 
-namespace boost
+namespace libMesh
 {
 
-namespace detail_unique_ptr
+namespace detail_UniquePtr
 {
 
 typedef char one;
@@ -73,13 +73,13 @@ struct identity
     typedef T type;
 };
 
-}  // detail_unique_ptr
+}  // detail_UniquePtr
 
 template <class T>
 inline
 typename enable_if_c
 <
-    !detail_unique_ptr::is_convertible<T, detail_unique_ptr::rv<T> >::value,
+    !detail_UniquePtr::is_convertible<T, detail_UniquePtr::rv<T> >::value,
     T&
 >::type
 move(T& t)
@@ -91,7 +91,7 @@ template <class T>
 inline
 typename enable_if_c
 <
-    !detail_unique_ptr::is_convertible<T, detail_unique_ptr::rv<T> >::value,
+    !detail_UniquePtr::is_convertible<T, detail_UniquePtr::rv<T> >::value,
     const T&
 >::type
 move(const T& t)
@@ -103,12 +103,12 @@ template <class T>
 inline
 typename enable_if_c
 <
-    detail_unique_ptr::is_convertible<T, detail_unique_ptr::rv<T> >::value,
+    detail_UniquePtr::is_convertible<T, detail_UniquePtr::rv<T> >::value,
     T
 >::type
 move(T& t)
 {
-    return T(detail_unique_ptr::rv<T>(t));
+    return T(detail_UniquePtr::rv<T>(t));
 }
 
 template <class T>
@@ -118,7 +118,7 @@ typename enable_if_c
     is_reference<T>::value,
     T
 >::type
-forward(typename detail_unique_ptr::identity<T>::type t)
+forward(typename detail_UniquePtr::identity<T>::type t)
 {
     return t;
 }
@@ -130,7 +130,7 @@ typename enable_if_c
     !is_reference<T>::value,
     T
 >::type
-forward(typename detail_unique_ptr::identity<T>::type& t)
+forward(typename detail_UniquePtr::identity<T>::type& t)
 {
     return move(t);
 }
@@ -142,16 +142,16 @@ typename enable_if_c
     !is_reference<T>::value,
     T
 >::type
-forward(const typename detail_unique_ptr::identity<T>::type& t)
+forward(const typename detail_UniquePtr::identity<T>::type& t)
 {
     return move(const_cast<T&>(t));
 }
 
-namespace detail_unique_ptr {
+namespace detail_UniquePtr {
 
 // A move-aware but stripped-down compressed_pair which only optimizes storage for T2
 template <class T1, class T2, bool = is_empty<T2>::value>
-class unique_ptr_storage
+class UniquePtr_storage
 {
     T1 t1_;
     T2 t2_;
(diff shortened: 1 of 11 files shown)


francesco-ballarin:
There are a couple of "using libMesh::UniquePtr" missing in empirical interpolation headers in reduced basis examples, I forgot them. I will try to fix this tomorrow.

Please also note that the AutoPtr deprecation message is messed up by my search & replace. I can also fix this.

jwpeterson:
One big issue with this PR, aside from my conceptual issues with renaming `boost::unique_ptr` to `libMesh::UniquePtr`, is that the renamed class no longer compiles, probably due to a bad search-and-replace.  If you dig down into the details of the CI, you'll see that configure reports
```
checking if Howard Hinnant's C++03 unique_ptr implementation can be compiled... no
```
Unfortunately the config.log file for this build has since been over-written so I can't tell you exactly what the error was.

francesco-ballarin:
m4/hinnant_unique_ptr.m4

is still testing for boost::unique_ptr and not for its new name. 

Anyway, I see that you are working on this, thanks (#536). Should I close this PR (since you do not agree on the rename bost->libMesh)? I am willing to help you on a new one, if you need.

jwpeterson:
> is still testing for boost::unique_ptr and not for its new name.

I see.

> Should I close this PR (since you do not agree on the rename bost->libMesh)?

Yeah... I don't think this PR will be merged as-is but there's no harm in keeping it open if you wish to continue the discussion. 

> I am willing to help you on a new one, if you need.

At the moment, I'm a bit stuck on how to best create the shim class that @roystgnr mentioned to handle the cases where C++11 alias declarations are not available.  My first attempt was to simply try deriving from boost::unique_ptr, see test code below.

```
#include <iostream>
#include "libmesh/unique_ptr.hpp"

// Dummy class which prints its address when created or destroyed
struct Foo
{
  Foo() {std::cout << "Foo [" << this << "] constructed\n";}
  virtual ~Foo() {std::cout << "Foo [" << this << "] destructed\n";}
};

template <class T>
class UniquePtrShim : public boost::unique_ptr<T>
{
public:
  // Constructor which simply calls the boost::unique_ptr constructor
  explicit UniquePtrShim(typename boost::unique_ptr<T>::pointer p) : boost::unique_ptr<T>(p) {}
  // explicit UniquePtrShim(T* t) : boost::unique_ptr<T>(t) {}

private:
  // Shim class copy constructor and assignment operator are private and unimplemented.
  UniquePtrShim(UniquePtrShim &);
  UniquePtrShim & operator=(UniquePtrShim &);
};

// Simulate returning a UniquePtrShim object, which we commonly do in libmesh...
UniquePtrShim<Foo> source()
{
  return UniquePtrShim<Foo>(new Foo);
}

{
  // This works fine
  {
    UniquePtrShim<Foo> ups(new Foo);
    ups.reset();
  }

  // This doesn't compile
  UniquePtrShim<Foo> ups = source();

  return 0;
}
```

but this doesn't quite work.  We need to be able to return UniquePtr objects from `source()` functions as in the test code, but this doesn't compile.  The error message is:
```
test_unique_ptr_shim.cc:28:10: error: no matching constructor for initialization of 'UniquePtrShim<Foo>'
  return UniquePtrShim<Foo>(new Foo);
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
test_unique_ptr_shim.cc:21:3: note: candidate constructor not viable: expects an l-value for 1st argument
  UniquePtrShim(UniquePtrShim &);
  ^
test_unique_ptr_shim.cc:40:22: error: no matching constructor for initialization of 'UniquePtrShim<Foo>'
  UniquePtrShim<Foo> ups = source();
                     ^     ~~~~~~~~
test_unique_ptr_shim.cc:21:3: note: candidate constructor not viable: expects an l-value for 1st argument
  UniquePtrShim(UniquePtrShim &);
  ^
2 errors generated.
```

Note that the same type of code works just fine when using `boost::unique_ptr` directly. Hopefully I'm missing something simple...

francesco-ballarin:
Maybe some of the comments here http://stackoverflow.com/questions/2910157/cant-inherit-from-auto-ptr-without-problems help?

Avoiding to have to inherit from boost::unique_ptr was the reason why a renamed boost::unique_ptr to libmesh::UniquePtr

jwpeterson:
Apologies if this gets emailed twice, it seems like github ate the first one.

> Maybe some of the comments here http://stackoverflow.com/questions/2910157/cant-inherit-from-auto-ptr-without-problems help?

I think the problem is definitely related to the various `unique_ptr` constructors, and agree that deriving from it is probably not possible in general.
 
> Avoiding to have to inherit from boost::unique_ptr was the reason why a renamed boost::unique_ptr to libmesh::UniquePtr

Right, I see why you went that route now.  We have that solution available if we really need it, but I still want to try a couple other things first...

By the way, the C++03-compatible solution to this problem is apparently to declare a (named) typedef inside a templated class as in the code snippet below: 
```
#include <iostream>
#include "libmesh/unique_ptr.hpp"

// Dummy class which prints its address when created or destroyed
struct Foo
{
  Foo() {std::cout << "Foo [" << this << "] constructed\n";}
  virtual ~Foo() {std::cout << "Foo [" << this << "] destructed\n";}
};

// We can "alias" a type using a class, but it's ugly.  See also:
// http://stackoverflow.com/questions/14144122/how-to-generate-alias-for-class-templates
template <class T>
class UniquePtrShim
{
public:
  typedef boost::unique_ptr<T> type;
};


// Simulate returning a UniquePtrShim object, which we commonly do in libmesh...
UniquePtrShim<Foo>::type source()
{
  return UniquePtrShim<Foo>::type(new Foo);
}

{
  // This works fine
  {
    UniquePtrShim<Foo>::type ups(new Foo);
    ups.reset();
  }

  // This works
  UniquePtrShim<Foo>::type ups = source();

  return 0;
}
```
but this only works if you are willing to litter your code with `::type`, which is pretty terrible.

francesco-ballarin:
Yes, I agree that this is the C++03 equivalente of template alias. I also agree that it is pretty terrible ;)

jwpeterson:
OK, I pushed a [branch](https://github.com/libMesh/libmesh/commits/unique_ptr_namespace) which uses namespacing and macros instead of renaming the entire `boost::unique_ptr` class 257db394dbacb351ecb43c7b9b35430f9af5117f.  I think I prefer this approach to completely renaming `boost::unique_ptr`.

But... it turns out that master is not currently working with clang 3.5.0 and `--enable-unique-ptr --disable-cxx11` anyway.  I'm pretty sure this configuration used to work, but I don't test it often myself so there may have been a regression recently.  The error seems to be related to the fact that there is a `move()` somewhere in clang's `std::` namespace even without `-std=c++11` specified.

```
./include/libmesh/unique_ptr.hpp:197:15: error: call to 'move' is ambiguous
        : t1_(move(t1)) {}
              ^~~~
./include/libmesh/unique_ptr.hpp:330:11: note: in instantiation of member function 'boost::detail_unique_ptr::unique_ptr_storage<std::__1::map<const libMesh::System *, libMesh::SystemNorm, std::__1::less<const libMesh::System *>, std::__1::allocator<std::__1::pair<const libMesh::System *const, libMesh::SystemNorm> > > *, boost::default_delete<std::__1::map<const libMesh::System *, libMesh::SystemNorm,
      std::__1::less<const libMesh::System *>, std::__1::allocator<std::__1::pair<const libMesh::System *const, libMesh::SystemNorm> > > >, true>::unique_ptr_storage' requested here
        : ptr_(p)
          ^
../src/error_estimation/uniform_refinement_estimator.C:98:5: note: in instantiation of member function 'boost::unique_ptr<std::__1::map<const libMesh::System *, libMesh::SystemNorm, std::__1::less<const libMesh::System *>, std::__1::allocator<std::__1::pair<const libMesh::System *const, libMesh::SystemNorm> > >, boost::default_delete<std::__1::map<const libMesh::System *, libMesh::SystemNorm, std::__1::less<const
      libMesh::System *>, std::__1::allocator<std::__1::pair<const libMesh::System *const, libMesh::SystemNorm> > > > >::unique_ptr' requested here
    UniquePtr<std::map<const System*, SystemNorm > >
    ^
./include/libmesh/auto_ptr.h:43:21: note: expanded from macro 'UniquePtr'
#  define UniquePtr boost::unique_ptr
                    ^
/opt/moose/clang_3.5.0/bin/../include/c++/v1/type_traits:1632:1: note: candidate function [with _Tp = std::__1::map<const libMesh::System *, libMesh::SystemNorm, std::__1::less<const libMesh::System *>, std::__1::allocator<std::__1::pair<const libMesh::System *const, libMesh::SystemNorm> > > *]
move(_Tp& __t)
^
./include/libmesh/unique_ptr.hpp:85:1: note: candidate function [with T = std::__1::map<const libMesh::System *, libMesh::SystemNorm, std::__1::less<const libMesh::System *>, std::__1::allocator<std::__1::pair<const libMesh::System *const, libMesh::SystemNorm> > > *]
move(T& t)
^
/opt/moose/clang_3.5.0/bin/../include/c++/v1/type_traits:1640:1: note: candidate function [with _Tp = std::__1::map<const libMesh::System *, libMesh::SystemNorm, std::__1::less<const libMesh::System *>, std::__1::allocator<std::__1::pair<const libMesh::System *const, libMesh::SystemNorm> > > *]
move(const _Tp& __t)
^
./include/libmesh/unique_ptr.hpp:97:1: note: candidate function [with T = std::__1::map<const libMesh::System *, libMesh::SystemNorm, std::__1::less<const libMesh::System *>, std::__1::allocator<std::__1::pair<const libMesh::System *const, libMesh::SystemNorm> > > *]
move(const T& t)
^
```

francesco-ballarin:
Thanks John, I will try to install a clang compiler to test it.

I still have to look at your branch more carefully. A possible suggestion is related to change line 300 of

unique_ptr_namespace/contrib/unique_ptr/unique_ptr.hpp

I am quite sure we do not want to pollute the libMesh namespace with all boost classes.  Is it possible to be more selective? May we need to use only boost::detail_unique_ptr ?

And, related to this, on line 135 there is an occurence of boost::move. I guess that this is the ambiguous move method.

Best,
Francesco

jwpeterson:
> I am quite sure we do not want to pollute the libMesh namespace with all boost classes. Is it possible to be more selective? May we need to use only boost::detail_unique_ptr ?

Oops, yes, that should be possible.  I will try to fix that now.

jwpeterson:
> Oops, yes, that should be possible. I will try to fix that now.

Should be fixed in 9711950d2fd3dc4762167fda0fcfe02353026bf3.

roystgnr:
Superceded by #537

Thanks again for helping us with this.

PR #531, opened 2015-04-06
AdjointRefinement::estimate_error, Check if user already solved adjoint problem

pbauman:
When computing the AdjointRefinement::estimate_error, we solve
the adjoint problem on the coarse space first. But, the user may
have already computed this, so we should check.

I got identical results between adjoints/ex4 before and after this change, but that's the only thing I've checked.


roystgnr:
Good catch.

PR #532, opened 2015-04-07
ErrorEstimatorType

pbauman:
It is useful for app codes to be able to query the error estimator type from the base class pointer. This PR adds the enum `ErrorEstimatorType` and an interface in the `ErrorEstimator` classes that returns the appropriate `ErrorEstimatorType`.


roystgnr:
This is good as-is.  Want to code up an enum_to_string specialization for it while you're at it?

PR #536, opened 2015-04-14
Add test for cxx11 alias declarations

jwpeterson:
I think these are sometimes conflated with template typedefs, e.g.
```
typedef check<int> check_type;
```
but the syntax is different and they are different features as far as I can tell...

### DIFF
--- a/configure
+++ b/configure
@@ -6699,6 +6699,10 @@ else
     // at compile time.
     constexpr int multiply (int x, int y) { return x * y; }
 
+    // A minimally-conforming C++11 compiler must support alias declarations
+    template <typename T>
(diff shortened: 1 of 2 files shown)


roystgnr:
Looks good to me.  Should we sync up with the autoconf version while we're at it?

http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=history;f=m4/ax_cxx_compile_stdcxx_11.m4

shows several changes we lack, albeit nothing critical.

jwpeterson:
Thanks for that link.  I'll see about merging the upstream changes into our version at some point.

PR #537, opened 2015-04-18
Allow UniquePtr to be used in properly-namespaced code

jwpeterson:
This is similar to @francesco-ballarin's solution in #529 except that instead of text-replacing `boost::unique_ptr` to UniquePtr, we add `unique_ptr` to the libMesh namespace.  This solution is a bit less invasive, and the original unique_ptr.hpp header is recovered if you don't `#define LIBMESH_IS_COMPILING_HINNANT_UNIQUE_PTR` when including it.  Also fixed a few ambiguous uses of `boost::move` and `boost::forward` triggered by the clang 3.5.0 compiler.

### DIFF
--- a/contrib/unique_ptr/unique_ptr.hpp
+++ b/contrib/unique_ptr/unique_ptr.hpp
@@ -132,7 +132,7 @@ typename enable_if_c
 >::type
 forward(typename detail_unique_ptr::identity<T>::type& t)
 {
-    return move(t);
+    return boost::move(t);
 }
 
 template <class T>
@@ -144,7 +144,7 @@ typename enable_if_c
 >::type
 forward(const typename detail_unique_ptr::identity<T>::type& t)
 {
-    return move(const_cast<T&>(t));
+    return boost::move(const_cast<T&>(t));
 }
 
 namespace detail_unique_ptr {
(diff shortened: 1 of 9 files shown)


contrib/unique_ptr/unique_ptr.hpp
```diff
@@ -462,18 +471,27 @@ class unique_ptr<T[], D>
...
+               typename boost::enable_if_c<boost::detail_unique_ptr::is_convertible<U, pointer>::value>::type* = 0);
 };
 
+#ifdef LIBMESH_IS_COMPILING_HINNANT_UNIQUE_PTR
+} // namespace libMesh
+
+
+namespace boost {
+
+using libMesh::unique_ptr;
```
roystgnr: Why declare it in libMesh and then using-declare it in boost, rather than the other way around?

roystgnr:
I like this fix.  If @francesco-ballarin is happy with it, or if @jwpeterson has tested all the necessary zillion configuration combinations himself, then let's go ahead and merge.

francesco-ballarin:
Thanks for the effort, I am definitely happy with it

contrib/unique_ptr/unique_ptr.hpp
jwpeterson: You mean, why not say
```
using boost::unique_ptr;
```
in libmesh and then put non-namespace-qualified `unique_ptr` declarations throughout the library? I don't think that allows us to support old app code that still has `AutoPtr` declarations, does it?

PR #539, opened 2015-04-22
Attempted nonfullorder fix

roystgnr:
As discovered in https://github.com/idaholab/moose/issues/4966 - https://github.com/libMesh/libmesh/pull/492 causes regressions when refining meshes with non-full-order second-order elements.

This "fallback" code path in the Elem::*bracketing_nodes() methods should hopefully fix that.  I'm not nearly done with my own testing yet, but I'm setting up the PR right away to get MooseBuild looking at it in parallel with me.

Regardless of what MooseBuild says, this shouldn't be merged until I've had time to hammer it with at least QUAD8 and a few of the 3D non-full-order options.

### DIFF
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -2284,18 +2284,27 @@ ElemType Elem::second_order_equivalent_type (const ElemType et,
   switch (et)
     {
     case EDGE2:
+    case EDGE3:
       {
         // full_ordered not relevant
         return EDGE3;
       }
 
+    case EDGE4:
+      {
+        // full_ordered not relevant
+        return EDGE4;
+      }
+
     case TRI3:
+    case TRI6:
       {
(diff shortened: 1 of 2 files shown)


jwpeterson:
This works for my modified simple_diffusion test that uses QUAD8's and refinement!

jwpeterson:
> This works for my modified simple_diffusion test that uses QUAD8's and refinement!

It also passes all the other MOOSE and modules tests for me, as well as the app where the problem was originally triggered.

roystgnr:
Sounds like this is at least a solid enough improvement to be worth merging immediately, but I don't think we're out of the woods yet.  I also see good results with QUAD8, but I get an assertion failure with HEX20.  I'll try to get that fixed too before someone runs into it for real.

Issue #538, opened 2015-04-22
LIBMESH_DEFINE_HASH_POINTERS does not work with Intel 15


jwpeterson:
On that compiler, this macro (which is used in serial_mesh.C) expands to
```
namespace __gnu_cxx 
{
template<typename T>
struct hash<T*>
{
  size_t operator()(const T* p) const 
  {
    return hash<size_t>()(reinterpret_cast<size_t>(p));
  }
};
}
```
The error message is:
```
../src/mesh/serial_mesh.C(30): error: a template argument list is not allowed in a declaration of a primary template
```
I actually think the `struct` declaration is wrong here...

jwpeterson:
Actually the syntax is correct, the error is caused because the naked LIBMESH_DEFINE_HASH_POINTERS declaration is trying to specialize the `__gnu_cxx::hash` template, but we haven't included a header (like ext/hash_set or ext/hash_map) that defines `__gnu_cxx::hash`, hence it can't be specialized.  Adding `#include <ext/hash_set>` to the macro should fix the problem.

PR #540, opened 2015-04-22
Fix compilation error on Intel-15 compilers with GCC 4.3 STL.

jwpeterson:
This was a fairly weird issue.  It turns out that this particular
compiler can compile std::tr1::unordered_{map,set} but not
std::tr1::hash because... things.  But this is OK because actually
using std::tr1::unordered_{map,set} does not require std::tr1::hash.

We should only require hash function specializations for pointer types
when LIBMESH_BEST_UNORDERED_MAP is actually __gnu_cxx::hash_map, and
ensuring that we include LIBMESH_INCLUDE_HASH should handle this case.

Addresses #538.


roystgnr:
:+1:

I can't believe you fixed that labyrinth of a compatibility issue with a one-line patch.

jwpeterson:
> I can't believe you fixed that labyrinth of a compatibility issue with a one-line patch.

I deleted all evidence of my first 2 attempts ;)

PR #541, opened 2015-04-23
Topology fix take 2

roystgnr:
This fixes HEX20 without regressing QUAD8 for me...

### DIFF
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -63,6 +63,7 @@
 #include "libmesh/remote_elem.h"
 #include "libmesh/reference_elem.h"
 #include "libmesh/string_to_enum.h"
+#include "libmesh/threads.h"
 
 #ifdef LIBMESH_ENABLE_PERIODIC
 #include "libmesh/mesh.h"
@@ -72,6 +73,9 @@
 
 namespace libMesh
 {
+
+Threads::spin_mutex parent_indices_mutex;
+Threads::spin_mutex parent_bracketing_nodes_mutex;
 
 const subdomain_id_type Elem::invalid_subdomain_id = std::numeric_limits<subdomain_id_type>::max();
 
@@ -1727,6 +1731,11 @@ unsigned int Elem::as_parent_node (unsigned int child,
     cached_parent_indices = this->_get_parent_indices_cache();
 
   unsigned int em_vers = this->embedding_matrix_version();
+
+  // We may be updating the cache on one thread, and while that
+  // happens we can't safely access the cache from other threads.
+  Threads::spin_mutex::scoped_lock lock(parent_indices_mutex);
+
   if (em_vers >= cached_parent_indices.size())
     cached_parent_indices.resize(em_vers+1);
 
@@ -1788,6 +1797,10 @@ Elem::parent_bracketing_nodes(unsigned int child,
     cached_bracketing_nodes = this->_get_bracketing_node_cache();
 
   const unsigned int em_vers = this->embedding_matrix_version();
+
+  // We may be updating the cache on one thread, and while that
+  // happens we can't safely access the cache from other threads.
+  Threads::spin_mutex::scoped_lock lock(parent_bracketing_nodes_mutex);
 
   if (cached_bracketing_nodes.size() <= em_vers)
     cached_bracketing_nodes.resize(em_vers+1);
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -2036,13 +2036,26 @@ Elem::bracketing_nodes(unsigned int child,
                   }
               }
 
-          // We should always find all bracketing nodes by the time
+          // We should *usually* find all bracketing nodes by the time
           // we query them (again, because of the child & node add
           // order)
-          libmesh_assert_not_equal_to (pt1, DofObject::invalid_id);
-          libmesh_assert_not_equal_to (pt2, DofObject::invalid_id);
+          //
+          // The exception is if we're a HEX20, in which case we will
+          // find pairs of vertex nodes and edge nodes bracketing the
+          // new central node but we *won't* find the pairs of face
+          // nodes which we would have had on a HEX27.  In that case
+          // we'll still have enough bracketing nodes for a
+          // topological lookup, but we won't be able to make the
+          // following assertions.
+          if (this->type() != HEX20)
+            {
+              libmesh_assert_not_equal_to (pt1, DofObject::invalid_id);
+              libmesh_assert_not_equal_to (pt2, DofObject::invalid_id);
+            }
 
-          returnval.push_back(std::make_pair(pt1, pt2));
+          if (pt1 != DofObject::invalid_id &&
+              pt2 != DofObject::invalid_id)
+            returnval.push_back(std::make_pair(pt1, pt2));
         }
     }
 
--- a/src/utils/topology_map.C
+++ b/src/utils/topology_map.C
@@ -58,6 +58,8 @@ void TopologyMap::add_node(const Node& mid_node,
 {
   const dof_id_type mid_node_id = mid_node.id();
 
+  libmesh_assert_not_equal_to(mid_node_id, DofObject::invalid_id);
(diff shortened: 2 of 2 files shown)


roystgnr:
Oh, context for future reference: this is an extension of the changes in #539 attempting to fix the class of bugs which were discovered in idaholab/moose#4966 and introduced in #492

roystgnr:
Refinement of Prism15 elements is still broken, but in *that* case it's a pre-existing problem, not anything TopologyMap related: Prism15 doesn't have any embedding_matrix defined.

That actually wouldn't be too hard a problem to fix, but honestly "We should be able to refine prisms even if they're not first order or full order" probably goes on the TODO list way below "we should be able to refine pyramids".

roystgnr:
Wow, *there* was a lucky catch.  I'm building up static caches on first use in objects that may be called from multiple threads, but I didn't think to add a lock.  I'll do that now.

jwpeterson:
Hang on, I don't think that failure is real.

permcody:
Oh boy...  that threads verification test is still problematic... I doubt this is real, how about I invalidate once and see. That test fails about 1% of the time.

roystgnr:
If we've got threading race conditions (which we do, looking at Elem::as_parent_node() and Elem::parent_bracketing_nodes()) then you'd *expect* failure to be very infrequent and non-deterministic.

Even if that test has some sort of app-level false positives, it did alert me to a real bug in the TopologyMap code.

roystgnr:
Well, not in the TopologyMap code, but in the relatively-new Elem code it depends on.

roystgnr:
Seriously?  Running with *one thread* we manage to hang on a scoped_lock constructor??

permcody:
@roystgnr - where are you seeing that?

If you are looking at results run through the MOOSE TestHarness they may not always be serial even for a serial run. There are a few tests that don't make sense in serial (like partitioning tests for instance or tests with threads). When we create a test specification we can specify (min|max)_parallel and/or (min|max)_threads to force a test to run at minimums or maximums respectively. When that happens you'll see the cyan "caveat" printed next to the result saying if it ran on more or less threads/procs than you specified.

roystgnr:
Just running the unit tests executable myself I see the same hang.  I'm pretty sure *that* isn't threaded, and I certainly don't see any threaded loops in the stack trace:

```
(gdb) where
#0  0x00007ffff41b0737 in sched_yield () at ../sysdeps/unix/syscall-template.S:81
#1  0x000000000049b89a in tbb::internal::atomic_backoff::pause (
    this=0x7fffffffcb50) at /usr/include/tbb/tbb_machine.h:378
#2  0x000000000049b8fa in __TBB_LockByte (flag=@0x810c00: 1 '\001')
    at /usr/include/tbb/tbb_machine.h:910
#3  0x000000000049b985 in tbb::spin_mutex::scoped_lock::scoped_lock (
    this=0x7fffffffcbd0, m=...) at /usr/include/tbb/spin_mutex.h:96
#4  0x00007ffff68c087c in libMesh::Elem::as_parent_node (this=0xaaa380, child=0, 
    child_node=0) at ../src/geom/elem.C:1734
#5  0x00007ffff68c17fc in libMesh::Elem::parent_bracketing_nodes (this=0xaaa380, 
    child=0, child_node=1) at ../src/geom/elem.C:1925
#6  0x00007ffff68c1fd3 in libMesh::Elem::bracketing_nodes (this=0xaaa380, child=0, 
    child_node=1) at ../src/geom/elem.C:1988
#7  0x00007ffff6a4dc31 in libMesh::MeshRefinement::add_node (this=0x7fffffffd130, 
    parent=..., child=0, node=1, proc_id=0) at ../src/mesh/mesh_refinement.C:117
#8  0x00007ffff68ddac1 in libMesh::Elem::refine (this=0xaaa380, 
    mesh_refinement=...) at ../src/geom/elem_refinement.C:65
#9  0x00007ffff6a55fac in libMesh::MeshRefinement::_refine_elements (
    this=0x7fffffffd130) at ../src/mesh/mesh_refinement.C:1700
#10 0x00007ffff6a5665e in libMesh::MeshRefinement::uniformly_refine (
    this=0x7fffffffd130, n=1) at ../src/mesh/mesh_refinement.C:1790
#11 0x00000000004e8a49 in MixedDimensionRefinedMeshTest::setUp (this=0xa9df40)
    at ../../tests/mesh/mixed_dim_mesh_test.C:265
#12 0x00000000004f5479 in CppUnit::TestCaller<MixedDimensionRefinedMeshTest>::setUp
    (this=0xa9df70) at /usr/include/cppunit/TestCaller.h:177
#13 0x00007ffff47bc4d2 in CppUnit::TestCaseMethodFunctor::operator()() const ()
   from /usr/lib/x86_64-linux-gnu/libcppunit-1.13.so.0
#14 0x00007ffff47b2a90 in CppUnit::DefaultProtector::protect(CppUnit::Functor const&, CppUnit::ProtectorContext const&) ()
   from /usr/lib/x86_64-linux-gnu/libcppunit-1.13.so.0
#15 0x00007ffff47b97dd in CppUnit::ProtectorChain::protect(CppUnit::Functor const&, CppUnit::ProtectorContext const&) ()
   from /usr/lib/x86_64-linux-gnu/libcppunit-1.13.so.0
#16 0x00007ffff47c21e9 in CppUnit::TestResult::protect(CppUnit::Functor const&, CppUnit::Test*, std::string const&) ()
   from /usr/lib/x86_64-linux-gnu/libcppunit-1.13.so.0
#17 0x00007ffff47bc18e in CppUnit::TestCase::run(CppUnit::TestResult*) ()
   from /usr/lib/x86_64-linux-gnu/libcppunit-1.13.so.0
#18 0x00007ffff47bc853 in CppUnit::TestComposite::doRunChildTests(CppUnit::TestResult*) () from /usr/lib/x86_64-linux-gnu/libcppunit-1.13.so.0
#19 0x00007ffff47bc76e in CppUnit::TestComposite::run(CppUnit::TestResult*) ()
---Type <return> to continue, or q <return> to quit---
   from /usr/lib/x86_64-linux-gnu/libcppunit-1.13.so.0
#20 0x00007ffff47bc853 in CppUnit::TestComposite::doRunChildTests(CppUnit::TestResult*) () from /usr/lib/x86_64-linux-gnu/libcppunit-1.13.so.0
#21 0x00007ffff47bc76e in CppUnit::TestComposite::run(CppUnit::TestResult*) ()
   from /usr/lib/x86_64-linux-gnu/libcppunit-1.13.so.0
#22 0x00007ffff47c1dc2 in CppUnit::TestResult::runTest(CppUnit::Test*) ()
   from /usr/lib/x86_64-linux-gnu/libcppunit-1.13.so.0
#23 0x00007ffff47c46fe in CppUnit::TestRunner::run(CppUnit::TestResult&, std::string const&) () from /usr/lib/x86_64-linux-gnu/libcppunit-1.13.so.0
#24 0x00007ffff47c65e0 in CppUnit::TextTestRunner::run(std::string, bool, bool, bool) () from /usr/lib/x86_64-linux-gnu/libcppunit-1.13.so.0
#25 0x00000000004934fb in main (argc=1, argv=0x7fffffffd868)
    at ../../tests/driver.C:21
```

Running some non-threaded GRINS regression tests, I get tons of successes, plus some (the ones that do refinement, I'm guessing) that just hang.

Has it been too long since I wrote threaded code??  Is there something *complicated* about constructing a scoped_lock?

roystgnr:
Nothing complicated, I'm just an idiot.  I'm using the same mutex in both methods despite the fact that the one calls the other.  Of course the inner method can't acquire a mutex that the outer method has locked.  I'll fix it now.

roystgnr:
Ironically, even though this code wasn't thread-safe, it was also never called from within a threaded loop.  So whatever triggered the threaded regression test failure earlier must have been a problem elsewhere.

roystgnr:
Going to fixup+rebase so as to not leave a "make check hangs indefinitely" commit in our history.

PR #542, opened 2015-04-24
Fix an issue with MeshBase::elem_dimensions().

dknez:
If all elements of dimension d are deleted from a mesh, then MeshBase::elem_dimensions() still returned
d as one of the dimensions in the mesh. The fix was just to clear _elem_dims at the start of MeshBase::cache_elem_dims().

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -444,12 +444,18 @@ public:
    * To add an element locally, set e->processor_id() before adding it.
    * To ensure a specific element id, call e->set_id() before adding it;
    * only do this in parallel if you are manually keeping ids consistent.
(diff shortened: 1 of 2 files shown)


roystgnr:
Thanks very much for the fix!

I don't think the new comments are very clear... maybe just comment that users should prepare_for_use() after they finish manually adding/deleting elements?

dknez:
I changed the comments, hopefully that's clearer now.

roystgnr:
Yeah, looks great.

jwpeterson:
Hehe, that last failure was because you had already deleted the branch I guess.

dknez:
ok, all good then

PR #543, opened 2015-04-24
Only perform second derivative calculations when !elem_is_affine.

jwpeterson:
This addresses idaholab/moose#4980 and should have been in the original branch, my bad.


roystgnr:
:+1:

PR #544, opened 2015-04-24
Fix mutex self-deadlock in the non-full_order case

roystgnr:
Thanks to @jwpeterson for catching something *I* should have tested.


jwpeterson:
This works in my new Hex20 refinement test, btw.

PR #547, opened 2015-04-27
Don't call dof_number() if there is no DOF to get.

jwpeterson:
This fixes part of an issue reported by @micahjohnson150 where a `Node` has no DOF for the system/variable in question, for example when trying to read in a LINEAR, LAGRANGE element defined on a mesh of QUAD9s.


roystgnr:
Good catch.  :+1:

PR #548, opened 2015-04-27
Don't compute second derivative-specific values when second derivatives ...

jwpeterson:
...aren't required.

This handles the case where an element is non-affine but the user has
not requested second derivatives to be computed.

### DIFF
--- a/include/fe/fe_map.h
+++ b/include/fe/fe_map.h
@@ -52,15 +52,14 @@ public:
    * single point with index p.  Takes the integration weights as
    * input, along with a pointer to the element and a list of points
    * that contribute to the element.  Also takes a boolean flag
-   * telling whether the element has an affine map to enable various
-   * optimizations.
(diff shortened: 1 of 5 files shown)


roystgnr:
:+1+

Were there still empirical performance regressions a la https://github.com/idaholab/moose/issues/4980 even after #543 was merged, or is this just a belt-and-suspenders optimization?

jwpeterson:
> On Apr 27, 2015, at 8:47 PM, roystgnr <notifications@github.com> wrote:
> 
> :+1+
> 
> Were there still empirical performance regressions a la idaholab/moose#4980 even after #543 was merged, or is this just a belt-and-suspenders optimization?
> 
Not in our regression tests, anyway.  I think they mostly all use affine elements.  This was the logic I had intended, but had a brain damage moment with the LIBMESH_ENABLE_SECOND_DERIVATIVES ifedfs...

PR #549, opened 2015-04-28
Disable tetgen when configured with --enable-strict-lgpl

jwpeterson:
Tetgen does not have an LGPL2-compatible license.

It has an MIT license encumbered with the following non-commerical clause:

"Distribution of this code for  any  commercial purpose  is permissible
 ONLY BY DIRECT ARRANGEMENT WITH THE COPYRIGHT OWNER."

In case anyone is interested, the PETSc team forked and rewrote Tetgen
(renaming it ctetgen) a little while back, but they don't distribute
it in the main PETSc repo because of the original license.  The
ctetgen source is available from https://bitbucket.org/petsc/ctetgen.

### DIFF
--- a/configure
+++ b/configure
@@ -34105,8 +34105,13 @@ fi
 
 
 # -------------------------------------------------------------
-# TetGen -- enabled by default
+# TetGen -- enabled unless --enable-strict-lgpl is specified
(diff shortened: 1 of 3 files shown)


roystgnr:
Nice catch.

PR #553, opened 2015-04-29
Be more permissive with the derivatives of conditionals

dschwen:
I'm already allowing derivatives of ```if``` statements, but the optimizer may turn a statement like 

```
if(c<0, 1, 0) 
```

into 

```
c<0
```

While the former statement is differentiable the later is not. This prevents users form optimizing a function before taking its derivatives (which will be required in the new version of the MOOSE ```DerivativeParsedMaterial```). We can easily be more permissive here and _not_ throw an exception in the second case.

### DIFF
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -333,13 +333,17 @@ typename ADImplementation<Value_t>::CodeTreeAD ADImplementation<Value_t>::D(cons
     case cFloor:
     case cCeil:
     case cTrunc:
+      throw RefuseToTakeCrazyDerivativeException;
+
+    // we gloss over the single undifferentiable point here
+
     case cEqual:
     case cNEqual:
     case cLess:
     case cLessOrEq:
     case cGreater:
     case cGreaterOrEq:
-      throw RefuseToTakeCrazyDerivativeException;
+      return CodeTreeAD(CodeTreeImmed(Value_t(0)));
 
     // these opcodes will never appear in the tree when building with keep_powi == false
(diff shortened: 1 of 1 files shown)


contrib/fparser/fparser_ad.cc
```diff
@@ -333,13 +333,17 @@ typename ADImplementation<Value_t>::CodeTreeAD ADImplementation<Value_t>::D(cons
     case cFloor:
     case cCeil:
     case cTrunc:
+      throw RefuseToTakeCrazyDerivativeException;
```
roystgnr: I don't know what cTrunc does, but why not allow returning 0 derivatives of floor() and ceil() ?  They've got the same "zero everywhere except at a jump" structure as the comparison operators do.

dschwen: Trunc rounds towards zero. Yeah, I could do that. I guess I was trying to
retain a shred of mathematical integrity ;-). Those functions have an
infinite number of undefined points in their derivatives, while the ones
I'm allowing now only have one single one.

jwpeterson:
Talked to @dschwen, he's not planning to change the PR to allow derivatives of floor() and ceil(), and this seems like a reasonable change on its own, so I'm merging it...

roystgnr:
Fair enough; it's definitely an improvement.  floor() and ceil() would still be good to change, though.  It's an infinite number of undefined points in those cases, but it's still a set of measure zero.

dschwen:
Hmm, I guess I could permit those. I'll think about how to handle this. I don't like completely ignoring this issue :-/.

PR #555, opened 2015-04-30
More _msg versions of macros

roystgnr:
Also saving a few lines by refactoring similar versions.


jwpeterson:
"Now with extra MSG"

:+1:

PR #556, opened 2015-05-05
Print more information when we hit an element with a negative Jacobian.

jwpeterson:
Because why not?

### DIFF
--- a/src/fe/fe_map.C
+++ b/src/fe/fe_map.C
@@ -382,7 +382,10 @@ void FEMap::compute_single_point_map(const unsigned int dim,
         jac[p] = dxyzdxi_map[p].size();
 
         if (jac[p] <= 0.)
-          libmesh_error_msg("ERROR: negative Jacobian: " << jac[p] << " in element " << elem->id());
+          {
+            elem->print_info();
+            libmesh_error_msg("ERROR: negative Jacobian: " << jac[p] << " in element " << elem->id());
+          }
 
         // The inverse Jacobian entries also come from the
         // generalized inverse of T (see also the 2D element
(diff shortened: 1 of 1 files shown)


roystgnr:
:+1:, and thanks!

dknez:
I'd like to merge this so I can use it in my testing. I assume that's OK with everyone?

roystgnr:
Yup.  Although "git checkout origin/better_error" wouldn't have been much of a hassle.

roystgnr:
It's slightly more tricky to test with a PR when it's coming from some_fork/some_branch and you haven't yet added a remote for some_fork.

permcody:
If there's a PR, there's one more shortcut that works for the "don't have a
remote for that fork" case:
git fetch origin pull/ID/head: BRANCHNAME

I personally like this approach quite a bit as I don't have to manage a
bunch of remotes to peoples forks.

Cody

jwpeterson:
On Tue, May 5, 2015 at 11:25 AM, Cody Permann <notifications@github.com>
wrote:

> If there's a PR, there's one more shortcut that works for the "don't have a
> remote for that fork" case:
> git fetch origin pull/ID/head: BRANCHNAME
>
> I personally like this approach quite a bit as I don't have to manage a
> bunch of remotes to peoples forks.
>

IMO it isn't that bad having a bunch of remotes, it's basically
add-and-forget.

Just remember *not* to "git fetch bob", since then you will get all of
bob's branches.

Instead,

git fetch bob branch_name
... do work...
git branch -d -r bob/branch_name

AFAIK the "git fetch origin pull/ID/head: BRANCHNAME" method is
github-specific.

-- 
John

PR #557, opened 2015-05-06
Fix the better error message fix

jwpeterson:
This fixes the infinite recursion bug @dknez found for elements with negative Jacobians.  It also adds printing of the physical-space point where the negative Jacobian is found.

### DIFF
--- a/src/fe/fe_map.C
+++ b/src/fe/fe_map.C
@@ -384,7 +384,7 @@ void FEMap::compute_single_point_map(const unsigned int dim,
         if (jac[p] <= 0.)
           {
             elem->print_info();
-            libmesh_error_msg("ERROR: negative Jacobian: " << jac[p] << " in element " << elem->id());
+            libmesh_error_msg("ERROR: negative Jacobian " << jac[p] << " at point " << xyz[p] << " in element " << elem->id());
           }
 
         // The inverse Jacobian entries also come from the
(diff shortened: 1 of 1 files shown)


dknez:
I'll try this on my mesh with the negative jacobian.

src/fe/fe_map.C
```diff
@@ -383,8 +383,23 @@ void FEMap::compute_single_point_map(const unsigned int dim,
...
-            libmesh_error_msg("ERROR: negative Jacobian: " << jac[p] << " in element " << elem->id());
+            // Don't call print_info() recursively if we're already
+            // failing.  print_info() calls Elem::volume() which may
+            // call FE::reinit() and trigger the same failure again.
+            static bool failing = false;
+            if (!failing)
+              {
+                failing = true;
+                elem->print_info();
+                libmesh_error_msg("ERROR: negative Jacobian " << jac[p] << " at point " << xyz[p] << " in element " << elem->id());
```
dknez: How about 80 columns per line? :)

roystgnr:
:+1: once @dknez is done testing it.

src/fe/fe_map.C
jwpeterson: OK, but only if we also follow [these](http://web.stanford.edu/class/me200c/tutorial_77/03_basics.html) guidelines...
```
Col. 1    : Blank, or a "c" or "*" for comments
Col. 1-5  : Statement label (optional)
Col. 6    : Continuation of previous line (optional)
Col. 7-72 : Statements
Col. 73-80: Sequence number (optional, rarely used today)
```

dknez:
Works for me, thanks!

src/fe/fe_map.C
dknez: hehe

jwpeterson:
> once @dknez is done testing it.

Let me amend the commit with some line continuation characters on the long lines.  And send @dknez a 30" monitor...

dmcdougall:
>Let me amend the commit with some line continuation characters on the long lines. And send @dknez a 30" monitor...

@jwpeterson I'd suggest doing one or the other.  Doing both just means @dknez will have plenty of unused screen real estate.

PR #558, opened 2015-05-10
An interface for optimization solvers

dknez:
Added the following:
- A generic interface for optimization solvers and specific interfaces for TAO and NLopt.
- Support for unconstrained and constrained optimization. Supports equality, inequality and bound constraints.
- Two new examples that illustrate unconstrained (optimization_ex1) and constrained (optimization_ex2) optimization.

@jwpeterson and I have worked on this and tested the examples. It all seems to be working well to us, but comments/feedback would be appreciated!


jwpeterson:
:+1:

dknez:
This PR is quite independent of the rest of libMesh since it adds new functionality without really touching any of the existing stuff. As a result, I don't see any harm in merging it. Does anyone have any objections to that?

roystgnr:
Give me a couple hours to finish looking over diffs?  If you don't hear back from me and I haven't already pushed the button by lunch time then go ahead and merge.

dknez:
No worries. There's no rush to merge, so take your time.

roystgnr:
Shouldn't take too long even if I don't rush.  I nope'd out the first time I saw "119 files changed", but I didn't realize that was just the new m4 file getting sucked into every Makefile.in.

jwpeterson:
One comment: we are aware that the new optimization_ex2 example doesn't converge with "vanilla" PETSc iterative solvers/preconditioners, but works fine with a direct solver, e.g.

```
./example-opt -tao_monitor -tao_view -tao_type ipm -ksp_type preonly -pc_type lu -pc_factor_mat_solver_package superlu_dist
```

(or `-pc_factor_mat_solver_package mumps -mat_mumps_icntl_14 50`, or with the nlopt solvers).  The issue is that the only preconditioner we could get working in the iterative case was Jacobi, others die with a "matrix is in wrong state" type of error.  The Tao support built-in to PETSc is relatively new (PETSc >= 3.5.0) so it's likely that it will improve over time.

roystgnr:
I hadn't caught that; thanks for the comment.  Let's overwrite LIBMESH_RUN and LIBMESH_OPTIONS in run.sh, then, to make sure we only run in serial and we don't override the necessary solver options?

jwpeterson:
> Let's overwrite LIBMESH_RUN and LIBMESH_OPTIONS in run.sh, then, to make sure we only run in serial and we don't override the necessary solver options?

Is 
```
LIBMESH_RUN=
LIBMESH_OPTIONS=
```
before calling `run_example` sufficient to do this?  I didn't see an example of this elsewhere in the examples...

roystgnr:
That should do it, yeah.

roystgnr:
:+1:

PR #560, opened 2015-05-12
Add dlopen configure test

jwpeterson:
We have been testing for `<dlfcn.h>` and searching for libraries which provide dlopen for a while. This test formalizes the process, tests for dlsym() and dlclose(), and sets the LIBMESH_HAVE_DLOPEN preprocessor macro.  The fparser JIT test, which used to check for dlopen, now simply uses the results of this test.

dlopen()/dlsym are used in MOOSE (see, e.g. idaholab/moose#5051), with this PR we can guard that code and fail gracefully on systems that don't support dlopen.


roystgnr:
It seems like we should have the DLOPEN test called from inside the fparser test, then?  There's some autotools magic to make sure the previous result is cached, but even duplicating the test or *only* calling the test from where it's used would feel more robust than "fparser JIT will be incorrectly disabled if you don't call a different macro first".

jwpeterson:
Since we control the order in which autoconf macros are called, I didn't think it was a big issue, but I suppose I could add an `AC_CACHE_CHECK`-based call to `AX_CXX_DLOPEN` in fparser.m4.

I disagree with *only* testing for dlopen from within the fparser test, however.  The entire point of this PR is to still get a valid result for LIBMESH_HAVE_DLOPEN independently of whether someone configures with --disable-fparser, so that apps can guard use of dlopen appropriately.

roystgnr:
Very good point; of course you're right that we want the separate check in configure.ac

roystgnr:
I'd still prefer a backup call in fparser.m4, but I'm happy merging as-is.

jwpeterson:
> I'd still prefer a backup call in fparser.m4, but I'm happy merging as-is.

Working on it.  I'm a little worried about the documentation of [AC_CACHE_CHECK](http://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Caching-Results.html), which says that "The commands-to-set-it must have no side effects".  Currently I'm calling `AC_DEFINE` there, which I believe counts as a side-effect, so I'm experimenting with moving it outside.

jwpeterson:
Configure now prints
```
checking for dlopen support... (cached) yes
```
in the fparser test, so it seems like the caching is working.  I did not test the case where dlopen is not found, but it seems like that would work the same way.

PR #561, opened 2015-05-13
Don't use cpp_compress in fparser

roystgnr:
It was already removed from the optimizer code generation, but we
missed it in this opcode code generation.

This fixes a compilation bug with clang 3.5 for me: the definition of
a macro named "q1" here (why not "#define i" while we're at it?!)
conflicted with an argument name in the emmintrin.h header.

### DIFF
--- a/contrib/fparser/Makefile.am
+++ b/contrib/fparser/Makefile.am
@@ -137,7 +137,6 @@ util_tree_grammar_parser_SOURCES   = \
                                  fpoptimizer/rangeestimation.hh \
                                  fpoptimizer/logic_boolgroups.hh \
                                  fpoptimizer/logic_powoperations.hh \
-                                 util/cpp_compress.hh \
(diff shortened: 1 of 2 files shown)


jwpeterson:
Should we update `contrib/fparser/extrasrc/fp_opcode_add.inc.release`?  Actually, is anything using that?

roystgnr:
We distribute the release version but nothing uses it.  I'm thinking just delete it?

jwpeterson:
@dschwen you added the contrib/fparser/extrasrc/fp_opcode_add.inc.release file, do you recall why?

dschwen:
I'll have to double check, by it may be the upstream version from vanilla
fparser.

roystgnr:
I believe you're right.  I'll merge this as-is and put in a second PR removing the extraneous file; that's an independent (and less concerning, since it doesn't break my clang install...) problem.

PR #562, opened 2015-05-13
Bugfix for create_submesh() with ParallelMesh

roystgnr:
Thanks to Harshad Sahasrabudhe for catching this!


roystgnr:
Let's hold off on merging for now; there are other bugs with this configuration so they might as well be added to this PR.

roystgnr:
Canceling this for now.  It's not the right fix in general, and until a few related bugs are fixed it won't be a working fix even in the particular "input mesh is serialized" case.

roystgnr:
This got fixed for real in #566

PR #564, opened 2015-05-14
Refactor around FEMContext::some_foo

roystgnr:
This saves a little code, but more importantly it adds implementations
of a few obscure methods like side_rate().

This passes the GRINS test suite, which ought to exercise this code fairly well.

### DIFF
--- a/src/systems/fem_context.C
+++ b/src/systems/fem_context.C
@@ -185,15 +185,10 @@ std::vector<boundary_id_type> FEMContext::side_boundary_ids() const
 
 
 template<typename OutputType,
-         void (FEMContext::*fe_getter)
-           (unsigned int,
-            FEGenericBase<typename TensorTools::MakeReal<OutputType>::type> *&)
-           const,
+         typename FEMContext::FENeeded<OutputType>::value_getter fe_getter,
          FEMContext::diff_subsolution_getter subsolution_getter>
 void FEMContext::some_value(unsigned int var, unsigned int qp, OutputType& u) const
 {
-  typedef typename TensorTools::MakeReal<OutputType>::type OutputShape;
-
   // Get local-to-global dof index lookup
   libmesh_assert_greater (this->get_dof_indices().size(), var);
   const unsigned int n_dofs = cast_int<unsigned int>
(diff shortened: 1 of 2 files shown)


pbauman:
Dat pointer-to-member function. :+1:

roystgnr:
Pointer-to-member function overloaded by a templated argument.  It took me three tries and one round of "let's see if clang gives more comprehensible compiler warnings" to get that to compile.

I'll merge this today unless anyone screams.  I'm fairly confident our tests would have caught any bugs, though.  Aside from FEMSystem and GRINS stuff, the refactored System projection code uses FEMContext now, so there's code coverage from all directions.

permcody:
That has got to be the craziest non-obfuscated C++ code I have ever looked
at! Pretty cool when it's all said and done!

jwpeterson:
> non-obfuscated C++ code
```
<OutputType>::type>::type>::type>
```

src/systems/fem_context.C
```diff
@@ -250,8 +259,16 @@ void FEMContext::some_interior_gradient(unsigned int var, unsigned int qp, Outpu
 
 #ifdef LIBMESH_ENABLE_SECOND_DERIVATIVES
 template<typename OutputType,
+         void (FEMContext::*fe_getter)
+           (unsigned int,
+            FEGenericBase
+              <typename TensorTools::MakeReal
+                <typename TensorTools::DecrementRank
+                  <typename TensorTools::DecrementRank
+                    <OutputType>::type>::type>::type> *&)
```
jwpeterson: Is it possible to use a `typedef` for the function pointer type?  Seems like you have repeated the same 7 lines of code used from line 843 of the header file...

roystgnr: It's possible with a C++11 using statement, but not in C++03.  OutputType is a free argument and appears as a template argument to DecrementRank here, and "template typedef" doesn't exist.  Best we could do is to add a nested templated class to the header and use three typedefs inside that class.  I'll bet that ends up being just as many lines of code, but I'll try it out if you like.

jwpeterson: Hmm... maybe a templated helper class (basically a c++03-compatible template typedef) would be good here?

I (tried)[https://github.com/jwpeterson/libmesh/commit/42fc9791d287f697c5c10e217968c07e01b67a40] to add one, but it doesn't compile.  Not sure if it's something fundamental or just a syntax error.

jwpeterson: Pushed one more commit to the [branch](https://github.com/jwpeterson/libmesh/tree/fem_context_refactor) which makes it compile.  The net lines are the same, but I think this follows the DRY principle a bit better.

roystgnr:
What do you think of this version?  Sorry, didn't mean to step on your toes; I'd already half-finished it when I saw you were working on the same.

jwpeterson:
:+1:

roystgnr:
Cool.  I'll merge once the MooseBuild and GRINS tests all finish.

PR #565, opened 2015-05-15
Remove fp_opcode_add.inc.release

roystgnr:
We generate this file at build time; there's no need to keep the
pregenerated code around.


roystgnr:
Ah, never mind.  We use this file in the if FPARSER_RELEASE case.  I'll leave it.

PR #566, opened 2015-05-20
ParallelMesh::create_submesh() fixes

roystgnr:
A second attempt to fix the bug that Harshad Sahasrabudhe caught.  This now passes simple tests for me with both ParallelMesh and SerialMesh.


roystgnr:
...and Harshad reports "This fix is working for me" too.  I'll merge once MooseBuild is happy.

PR #568, opened 2015-05-21
Implement eigen_get_converged_reason()

jwpeterson:
This trivial change allows us to run several of the reduced basis examples with `--use-eigen` on the command line.  I picked a somewhat arbitrary mapping between the Eigen::ComputationInfo enum and our LinearConvergenceReason enum in 11af914, but it's better than nothing.


roystgnr:
Nice; thanks!  :+1:

PR #570, opened 2015-05-22
Fixes to find_nodal_neighbors() for 1D elements

jwpeterson:
The edge-based algorithm we were using didn't work with 1D elements, because they don't have edges.

### DIFF
--- a/include/mesh/mesh_tools.h
+++ b/include/mesh/mesh_tools.h
@@ -313,7 +313,7 @@ unsigned int max_level (const MeshBase &mesh);
  * every node directly attached to the given one.
  */
 void find_nodal_neighbors(const MeshBase &mesh, const Node &n,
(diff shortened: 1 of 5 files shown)


roystgnr:
Huh.  So how *were* 1D elements working before?

jwpeterson:
They weren't... it was just returning zero neighbors for all the nodes.  MeshTools::find_nodal_neighbors() isn't called that much in 1D I guess.

roystgnr:
If beggars can be choosers, I'd say add a unit test.  :+1: regardless.

jwpeterson:
> On May 22, 2015, at 11:26 AM, roystgnr <notifications@github.com> wrote:
> 
> If beggars can be choosers, I'd say add a unit test.   regardless.
> 

I'll add the test case I used to develop it...
.

PR #571, opened 2015-05-22
Solution derivative access in ParsedFEMFunction

roystgnr:
If we have scalar variables "u", "Temperature", etc. then this lets us also compute parsed functions using e.g. "grad_x_u", "hess_xy_Temperature", etc.

We now also only compute input FEM data for a given variable and derivative level if the expressions we've been asked to evaluate include that data.

### DIFF
--- a/include/numerics/parsed_fem_function.h
+++ b/include/numerics/parsed_fem_function.h
@@ -345,7 +345,14 @@ public:
       {
         FEBase* elem_fe;
         c.get_element_fe(v, elem_fe);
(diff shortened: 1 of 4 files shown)


roystgnr:
This doesn't break the (sparse) ParsedFEMFunction usage in the GRINS regression tests.  But during the course of development I hit a scary point of "I know this code is broken but it's not breaking any tests", so now we'll have a bunch of ParsedFEMFunction unit tests in libMesh too.

PR #574, opened 2015-05-27
Silence clang 3.5.1 warnings.

jwpeterson:
```
/src/systems/fem_context.C:531:16: warning: reference cannot be bound to dereferenced
null pointer in well-defined C++ code; pointer may be assumed to always convert to
true [-Wundefined-bool-conversion]
```

The compiler seems to be in the right here.  Interestingly, clang 3.5.0 doesn't seem to mind them.

### DIFF
--- a/src/systems/fem_context.C
+++ b/src/systems/fem_context.C
@@ -493,7 +493,6 @@ void FEMContext::interior_curl(unsigned int var, unsigned int qp,
 
   // Get current local coefficients
   libmesh_assert_greater (this->_elem_subsolutions.size(), var);
-  libmesh_assert(&(this->get_elem_solution(var)));
   const DenseSubVector<Number> &coef = this->get_elem_solution(var);
 
   // Get finite element object
@@ -528,7 +527,6 @@ void FEMContext::interior_div(unsigned int var, unsigned int qp,
 
   // Get current local coefficients
   libmesh_assert_greater (this->_elem_subsolutions.size(), var);
-  libmesh_assert(&(this->get_elem_solution(var)));
   const DenseSubVector<Number> &coef = this->get_elem_solution(var);
 
   // Get finite element object
@@ -634,7 +632,6 @@ void FEMContext::side_gradient(unsigned int var, unsigned int qp,
 
   // Get current local coefficients
   libmesh_assert_greater (this->_elem_subsolutions.size(), var);
-  libmesh_assert(&(this->get_elem_solution(var)));
   const DenseSubVector<Number> &coef = this->get_elem_solution(var);
 
   // Get finite element object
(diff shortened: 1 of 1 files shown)


roystgnr:
Rather than remove the assertions entirely, could we e.g. libmesh_assert(_elem_subsolutions[i]) without going through the accessor?

jwpeterson:
The `get_elem_solution( unsigned int )` accessor already asserts, there doesn't seem to be anything in the `get_elem_solution()` accessor?

jwpeterson:
The libmesh_PR_test_dbg test that seems to hang forever might be a real issue -- we dug into the logs and there is a segfault during the unit testing.  I invalidated it once, and the same thing happened.  I'm taking a closer look...

jwpeterson:
The segfault here is down in CPPUnit, and is probably caused at least in part by our debugging flags.  I think we've seen this before, but I can't remember how we fixed it... @dschwen do you recall?
```
#0  0x00007ffff50a8360 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x00007ffff50a84ec in __gnu_debug::_Safe_sequence_base::_M_detach_all() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x000000000050948c in __gnu_debug::_Safe_sequence_base::~_Safe_sequence_base (this=0x7fffffffd6d8, __in_chrg=<optimized out>) at /usr/include/c++/4.6/debug/safe_base.h:199
#3  0x000000000052ffc6 in __gnu_debug::_Safe_sequence<std::__debug::deque<std::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::basic_string<char, std::char_traits<char>, std::allocator<char> > > > >::~_Safe_sequence (this=0x7fffffffd6d8, __in_chrg=<optimized out>) at /usr/include/c++/4.6/debug/safe_sequence.h:112
#4  0x000000000052ffe5 in std::__debug::deque<std::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::basic_string<char, std::char_traits<char>, std::allocator<char> > > >::~deque (this=0x7fffffffd688, __in_chrg=<optimized out>) at /usr/include/c++/4.6/debug/deque:116
#5  0x0000000000529d81 in CppUnit::Message::~Message (this=0x7fffffffd680, __in_chrg=<optimized out>) at /usr/include/cppunit/Message.h:38
```

jwpeterson:
Merging this, will look into debugging the cppunit/dbg stuff separately.

roystgnr:
How are we running anything with CppUnit and the std::__debug containers?  tests/Makefile.am puts...

Ah, I'm an idiot.  I'll fix this.

jwpeterson:
>  I'll fix this.

:+1:

PR #576, opened 2015-05-28
ParameterMultiAccessor

roystgnr:
This is a bit more overhead and a lot more flexible than ParameterMultiPointer.


jwpeterson:
The debug test has probably already segfaulted/hung forever so if you are happy, merge any time.

PR #577, opened 2015-05-28
Don't run unit_tests-dbg with GLIBCXX_DEBUG flags

roystgnr:
That breaks on some cppunit installations.  We had this fixed before, but then I regressed it again when adding the LIBMESH_RUN support for unit tests.


roystgnr:
Hopefully this fixes the red herrings John was tracking down in #574.

jwpeterson:
Yay, already made it past the debug unit testing bit.

PR #580, opened 2015-05-29
Removing unnecessary headers suggested by test_src.sh

jwpeterson:
This is a little tricky because of different configuration options which are possible... but a lot of these seem redundant.

### DIFF
--- a/src/mesh/abaqus_io.C
+++ b/src/mesh/abaqus_io.C
@@ -23,11 +23,8 @@
 
 // Local includes
 #include "libmesh/abaqus_io.h"
-#include "libmesh/point.h"
 #include "libmesh/elem.h"
 #include "libmesh/string_to_enum.h"
-#include "libmesh/boundary_info.h"
-#include "libmesh/utility.h"
 
 // Anonymous namespace to hold mapping Data for Abaqus/libMesh element types
 namespace
--- a/src/mesh/boundary_info.C
+++ b/src/mesh/boundary_info.C
@@ -26,8 +26,6 @@
 #include "libmesh/elem.h"
 #include "libmesh/mesh_data.h"
 #include "libmesh/mesh_serializer.h"
-#include "libmesh/parallel.h"
-#include "libmesh/partitioner.h"
 #include "libmesh/unstructured_mesh.h"
 
 namespace libMesh
--- a/src/mesh/checkpoint_io.C
+++ b/src/mesh/checkpoint_io.C
@@ -27,18 +27,8 @@
 #include <sstream> // for ostringstream
 
 // Local includes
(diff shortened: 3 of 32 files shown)


roystgnr:
This seems too fragile a methodology to me.  test_src.sh might be a good way to avoid false negatives but IMHO it comes up with a lot of false positives.  Take a look at the first thing on the list, abaqus_io.C: we do construct a Point, we do use get_boundary_info(), and we do use a Utility:: method.  So the only reason we can compile if we remove those three headers is that they're being indirectly included by the others.  That means that removing them isn't actually saving much (the preprocessor won't take long to skim from a header guard to its close) and will backfire if we later get more aggressive about header removal or pImpl or forward declaration in the intermediary headers.

jwpeterson:
> the only reason we can compile if we remove those three headers is that they're being indirectly included by the others.

Right, but where should we draw the line?  If you use both `Node` and `Point` in a source file, should you include both "point.h" and "node.h" on the off chance that some day node.h will no longer #include point.h?

> if we later get more aggressive about header removal or pImpl or forward declaration in the intermediary headers.

If we refactor the header files, wouldn't we just fix up the affected source files at the same time?

Anyway, I do understand your point, so I won't merge this as-is, but I do think some header cleanup would be good.

roystgnr:
Silly thought - could we check the output of "gcc -M" in test_src.sh?  That still won't be perfect, but I think the only category of false positives left would be conditional compilation.

jwpeterson:
On Fri, Jun 5, 2015 at 9:24 AM, roystgnr <notifications@github.com> wrote:

> Silly thought - could we check the output of "gcc -M" in test_src.sh? That
> still won't be perfect, but I think the only category of false positives
> left would be conditional compilation.
>
Not exactly sure what you mean... if we have redundant includes in the
source file, they will still be in the output of "gcc -M" right?

If we remove a header from foo.C and it still shows up in the output of
"gcc -M" it means the inclusion in the C file was redundant, but you said
you wanted to keep redundant includes around...

-- 
John

roystgnr:
Right.  But if we remove a header and it *doesn't* show up in gcc -M, then (assuming we're sure it isn't conditionally compiled in some ENABLE_FOO case) we can definitely safely remove it.

jwpeterson:
On Fri, Jun 5, 2015 at 9:44 AM, roystgnr <notifications@github.com> wrote:

> Right. But if we remove a header and it *doesn't* show up in gcc -M, then
> (assuming we're sure it isn't conditionally compiled in some ENABLE_FOO
> case) we can definitely safely remove it.
>

I see.  Yeah, let me take a crack at that this weekend some time...  I'm
thinking it won't find all that much though.

-- 
John

PR #581, opened 2015-06-01
PETSc 3.6 compatibility fixes

jwpeterson:
* The conf directory is now in lib/petsc/conf
* Add an AC_COMPILE_IFELSE test to be sure that the PETSc installation actually works
* PETSc private headers are now in petsc/private rather than petsc-private
* Try to support both prefix-installed and uninstalled PETSc builds (@karpeev can you try this branch?) 
* The temporary Makefile generation code now uses print rather than cat, so we don't have to have explicit tab characters in the m4.

### DIFF
--- a/m4/petsc.m4
+++ b/m4/petsc.m4
@@ -105,76 +105,113 @@ AC_DEFUN([CONFIGURE_PETSC],
         # Check for snoopable MPI
         if (test -r $PETSC_DIR/bmake/$PETSC_ARCH/petscconf) ; then           # 2.3.x
           PETSC_MPI=`grep MPIEXEC $PETSC_DIR/bmake/$PETSC_ARCH/petscconf | grep -v mpiexec.uni`
+
         elif (test -r $PETSC_DIR/$PETSC_ARCH/conf/petscvariables) ; then # 3.0.x
           PETSC_MPI=`grep MPIEXEC $PETSC_DIR/$PETSC_ARCH/conf/petscvariables | grep -v mpiexec.uni`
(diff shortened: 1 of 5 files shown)


m4/petsc.m4
```diff
@@ -101,82 +101,117 @@ AC_DEFUN([CONFIGURE_PETSC],
...
         fi
+
+        # If we couldn't snoop MPI from PETSc, fall back on ACX_MPI.
         if test "x$PETSC_MPI" != x ; then
-          AC_DEFINE(HAVE_MPI, 1,
-    	        [Flag indicating whether or not MPI is available])
+          AC_DEFINE(HAVE_MPI, 1, [Flag indicating whether or not MPI is available])
           MPI_IMPL="petsc_snooped"
-          AC_MSG_RESULT(<<< Configuring library with MPI from PETSC config >>>)
+          AC_MSG_RESULT(<<< Attempting to configure library with MPI from PETSC config... >>>)
```
roystgnr: humility.m4

jwpeterson: Well, it is confusing to say "we are doing this" when it became possible to fail the AC_COMPILE_IFELSE test below...

roystgnr:
Diffs look good.  Let me double-check a system or two before we merge.

roystgnr:
This works on the Ubuntu default petsc 3.4 install, my own current 3.5 build, and a couple old 3.3 and 3.5 modules we have lying around.  Good enough for me.

jwpeterson:
Thanks!

PR #582, opened 2015-06-02
NewmarkSolver

pbauman:
This PR adds `NewmarkSolver` as a `TimeSolver` subclass for use within the `DifferentiableSystem` framework. This required adding acceleration data structures and accessors to `DiffContext` and `FEMContext` as well as a `damping_residual` method, in analogy with `mass_residual`. I've reused `mass_residual` for second order systems. I think this should be OK, but there may be some unforeseen complications in, say, FSI problems that may bite later.

On first pass, it wasn't clear that fem_system_ex2 would be easy to update to use this. It would probably only take an hour or so to port over a special case of the elastic membrane stuff I'm doing, but not sure when I'd get to it.

I have a working GRINS example (PR coming soon for that) so this is tested (and will be continually tested once the GRINS PR goes in); however, I haven't done MMS on this yet so second order convergence is still yet to be determined. But I do converge to the steady state solution at the very least.


include/physics/diff_physics.h
```diff
@@ -303,6 +303,53 @@ class DifferentiablePhysics
   virtual bool nonlocal_mass_residual (bool request_jacobian,
                                        DiffContext &c);
 
+  /**
+   * Subtracts a damping vector contribution on \p elem from
+   * elem_residual.
+   *
```
roystgnr: Could we document this better?  Likewise for the definition of mass_residual in second order problems?

src/systems/fem_context.C
```diff
@@ -1252,6 +1252,28 @@ void FEMContext::side_rate(unsigned int var, unsigned int qp,
...
 
+template<typename OutputType>
+void FEMContext::interior_accel(unsigned int var, unsigned int qp,
+                                OutputType& u) const
+{
+  this->some_value<OutputType,
+                   &FEMContext::get_element_fe
+                   <typename TensorTools::MakeReal<OutputType>::type>,
+                   &DiffContext::get_elem_solution_accel>(var, qp, u);
+}
```
roystgnr: Looks like I refactored just in time?

src/systems/fem_context.C
```diff
@@ -1532,6 +1554,7 @@ void FEMContext::pre_fe_reinit(const System &sys, const Elem *e)
   if (sys.use_fixed_solution)
     this->get_elem_fixed_solution().resize(n_dofs);
   this->get_elem_solution_rate().resize(n_dofs);
+  this->get_elem_solution_accel().resize(n_dofs);
```
roystgnr: I'd like to only initialize this in second-order problems (and only initialize rate in first-order problems) but I don't see how the FEMContext can *determine* what order time integration we need...

roystgnr:
Document the non-self-documenting bits before merging?  Code looks good at first glance.  :+1:

jwpeterson:
Branch looks good... I'd prefer uniformly C++-style comments (e.g. 23254bde0c) but that's obviously minor.  Adding an example/unit test for this much new functionality is really important, though, hopefully you can get to this soon?

src/systems/fem_context.C
pbauman: Indeed. :)

src/systems/fem_context.C
pbauman: Agreed. Only thing that comes to mind is to force `FEMSystem` in the constructor and extract the time_solver.

pbauman:
Rebased with #585 merged and did a documentation pass. @roystgnr, you OK with what I added in 4df5144? Anything else I could add?

@jwpeterson, curious why you prefer only C++-style comments? Also, I might just copy then refactor into `FEMSystem` (fem_system_ex3?) the linear elasticity example in systems_of_equations_ex5. Would that be acceptable?

jwpeterson:
> @jwpeterson, curious why you prefer only C++-style comments? 

Mostly just consistency of style, the 
```
/**
 *
 */
```
blocks are Doxygen-specific class and method documentation, and inline parameter comments like `func(/*foo=*/4, /*bar=*/2);` are fine, everything else should use C++ style comments.  There are also some [gotchas](http://debian.fmi.uni-sofia.bg/~mrpaff/Effective%20C++/EC/EI4_FR.HTM) with C-style comments in general...

> Also, I might just copy then refactor into FEMSystem (fem_system_ex3?) the linear elasticity example in systems_of_equations_ex5. Would that be acceptable?

systems_of_equations_ex5 is the linear elastic cantilever with constraint problem?  That's fine with me, but does that example benefit from Newmark time integration?

pbauman:
> That's fine with me, but does that example benefit from Newmark time integration?

The example as-is is static, so, in that sense, yes, we'll be adding dynamics to the example. In general, unless you're doing stress formulation of elasticity or rewriting as a first order system, then yes, Newmark is the only second order scheme we have at the moment. But maybe I'm missing the point of your question?

dknez:
systems_of_equations_ex4 or ex6 might be a better candidate? ex5 has a lagrange multiplier constraint in it, so not sure you want to have that in your new example?

pbauman:
> systems_of_equations_ex4 or ex6 might be a better candidate? ex5 has a lagrange multiplier constraint in it, so not sure you want to have that in your new example?

I was actually thinking that would be a good test of the constraint part to make sure I didn't muck it up; the problems for which I'm using Newmark don't have constraints, so that part is untested in Newmark ATM.

pbauman:
That said, if you think the other examples are better, I'll go that route. Frankly, I'm looking for the path-of-least-resistance.

dknez:
That makes sense and I think ex5 is fine, I just wanted to make sure you wanted to have the SCALAR in your example (if not, then the other examples are better).

jwpeterson:
> But maybe I'm missing the point of your question?

Nope, adding dynamics to that example sounds great.

src/systems/fem_context.C
roystgnr: Definitely not an option.  We use FEMContext all the time with non-FEMSystem code.

roystgnr:
Otherwise comments look decent.  You might replace instances of e.g. (u,dot(u)) with just (u).  Technically all of M, C, and F are allowed to depend on dot(u) (and ddot(u) in the second-order case) now; only G(u) really ought to be independent of rate+accel.

pbauman:
> Technically mass_residual computes -M now, not M, right?

Well, since you redid the rate stuff, not necessarily, right? My understanding is that, if the user is overriding all of the DiffPhysics residual methods, in particular mass_residual, it kind of doesn't matter as long as their own sign convention is consistent with the underlying PDE since we just call their methods for accumulating the residuals in EulerSolver; we just have a default implementation for first order mass_residual that assumes negative so it makes sense to say we're computing -M to be consistent with that for first order problems. Or am I missing something? Regardless, it is consistent with the way first order problems are usually written, i.e. M(u)\dot{u} = F(u) so that the residual would be F(u) - M(u)\dot{u} = 0.

Second order problems (for solids/structures anyway) are usually written as M(u)\ddot{u} + C(u)\dot{u} + F(u) = 0 (actually they usually separate internal and external loading so there would be another F on the r.h.s.). And if the user is doing a second order problem, they *have* to override mass_residual for unsteady problems, so I figured we're free to say what convention we want independent of first order problems, so I just chose the convention that is consistent with that (but again, as long as the user is slef-consistent, it doesn't really matter, right?).

I did update the class documentation for `DiffPhysics` to reflect both of the above conventions.

src/systems/fem_context.C
pbauman: Are `elem_solution_rate` and (now) `elem_solution_accel` used outside of FEMSystem code?

If not, what about using `dynamic_cast`? If it returns NULL, then it's not FEMSystem and we can safely assume solution_rate etc. aren't being used. If we cast to FEMSystem, then we can grab the time solver and check.

roystgnr: Huh.  No, they aren't, and yes, that sounds like a good idea.  We'll eventually need to add a just-build-everything codepath for the !LIBMESH_HAVE_RTTI case, but that's not an important configuration and I don't think it's currently working either.

pbauman: Cool, I'll do that in a separate PR.

pbauman:
OK, example added, comments updated. I think this is good to go. Not sure why MooseBuild isn't firing up again though...

pbauman:
OK, I get an internal server error page when I try to go to the page of the last build in this PR.

jwpeterson:
> OK, I get an internal server error page when I try to go to the page of the last build in this PR.

Hardware failure on our moosebuild server... may be a little while before it's back online.

roystgnr:
I'll try out the branch on a couple systems.  The only change post-dating the last MooseBuild test is the new example, right?  So it's almost surely safe to merge.

permcody:
Moosebuild is back up.

roystgnr:
Moosebuild tests of that last commit set are running now.

pbauman:
Thanks. I want to hold off merging for just a bit --- I'm trying some cases out and I'm seeing a lot of numerical dissipation. I want to make sure it's not a bug.

pbauman:
There was a sign error in 1 term. Will be pushing fix shortly.

pbauman:
Reference I was using had a sign error in one of the terms :@ Should've just done the math myself from the beginning. That is fixed, I'm getting zero numerical dissipation in fem_system_ex3 (no damping there) as expected, so I feel good about that. Also updated example with a ghetto rig to output the velocity and acceleration fields as well. Feel free to take another once over, but at this point I'm happy with this and think it's ready to be merged once MooseBuild is good.

PR #584, opened 2015-06-03
Mixed dim refinement

roystgnr:
This needs more tests added before merging, but it should be ready for MooseBuild to look at.  There are a swath of edits here designed to support mixed-dimension meshes with linked interior and boundary elements.


pbauman:
@cahaynes, you need to send @roystgnr some nice whiskey for this one.

roystgnr:
Test coverage is getting better, a full application test is going to have to wait for after #587 and some additional fixes, and we're not interfering with any other test cases, so this is probably ready to merge as-is once anyone interested has had time to look at it.

include/geom/cell_hex.h
```diff
@@ -180,6 +180,9 @@ inline
 Hex::Hex(const unsigned int nn, Elem* p, Node** nodelinkdata) :
   Cell(nn, Hex::n_sides(), p, _elemlinks_data, nodelinkdata)
 {
+  // Make sure the interior parent isn't undefined
+  if (LIBMESH_DIM > 3)
```
pbauman: Is `LIBMESH_DIM` ever `> 3`?

roystgnr: Never, but I figure there's no reason to preclude it.  Twenty years from now when we're solving on tesseracts in space-time meshes you'll thank me.

roystgnr: Oh, and to elaborate, "no reason to preclude it" includes "anything past -O0 will strip out the contents of an if(false) block", since we don't have any weird "the compiler has to instantiate the function even if it knows it won't reach the call" situations like we did in Antioch.

include/geom/elem.h
```diff
@@ -322,36 +322,66 @@ class Elem : public ReferenceCountedObject<Elem>,
...
-   * This function finds all elements (including this one) which
-   * touch the current active element at the specified point, which
-   * should be a point in the current element.
+   * This function finds all active elements (including this one)
+   * which are in the same manifold as this element and which touch
+   * the current active element at the specified point, which should
+   * be a point in the current element.
+   *
+   * Elements which are not "in the same manifold" (e.g. the
+   * interior_parent of a boundary element) will not be found with
```
pbauman: I'm not quite sure I understand this. So, let's say I have a 2D manifold (in 3D space) with a 1D manifold overlapping on the boundary of the 2D manifold. Will this then not find the 1D element overlapping?

roystgnr: That's correct.  And that's always *been* correct; all this PR does is document it a little better.

Currently, interior_parent() is a one-way link, so there's no way to get from higher-dimensional to lower-dimensional manifolds.

pbauman: Ah, OK. I was thinking the other way around. Makes sense now.

pbauman:
I did a pass through this and nothing sticks out as glaring.

PR #585, opened 2015-06-03
Pass elem_dim to FEMContext::get_element/side_fe

pbauman:
Internally in FEMContext, when evaluating quantities, pass the
element dimension to make sure we get the right element in mixed
dimension meshes. We can do this because the element will have
already been set in the FEMContext before these methods are called.

I didn't hit this before because I wasn't actually calling these methods due to the nature of the solid mechanics weak forms I was using (derivatives are w.r.t. reference element coordinates, not spatial coordinates). I hit this in the mixed-dimensional dynamic case in the mass_residual/interior_accel (#582) case. This fixes that.

### DIFF
--- a/include/systems/fem_context.h
+++ b/include/systems/fem_context.h
@@ -820,17 +820,17 @@ protected:
     // Typedefs for "Value getter" function pointer
     typedef typename TensorTools::MakeReal<OutputType>::type value_shape;
     typedef FEGenericBase<value_shape> value_base;
-    typedef void (FEMContext::*value_getter) (unsigned int, value_base *&) const;
+    typedef void (FEMContext::*value_getter) (unsigned int, value_base *&, unsigned char) const;
 
     // Typedefs for "Grad getter" function pointer
     typedef typename TensorTools::MakeReal<Rank1Decrement>::type grad_shape;
     typedef FEGenericBase<grad_shape> grad_base;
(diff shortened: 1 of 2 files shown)


roystgnr:
Remind me again, why do we pass in get_dim() instead of get_elem_dim() in the default case?

pbauman:
Well... What I remember is that `get_element_fe` may be called before the element dimension is actually set in the FEMContext. Commit log says as much.

> all dimension dependent calls have a
default version that defaults preferably to the element dimension,
if guaranteed available, otherwise to the mesh dimension; for single
dimension meshes, there is no difference.

But since I can't actually point exactly where this happens at this second, clearly I should've documented more...

OK, there it is. It's for backwards compatibility - we can call `get_element_fe` in `init_context` without the dimension argument, but the element dimension has not yet been set yet in the context, so we default to the mesh dimension.

roystgnr:
Ah, that makes sense.  And going with a more complicated "use elem dim if elem is set, use mesh dim if not" would just make the API more confusing.  :+1:

pbauman:
Ok, great. I'll a (trimmed down...) version of my comment above to the documentation so I don't forget again.

pbauman:
Alright comments updated. I'll merge once MooseBuild is happy.

pbauman:
Arg. Squashed an update into the comments mentioning `init_context` specifically.

PR #587, opened 2015-06-03
JumpErrorEstimator refactor

roystgnr:
This is necessary to support use of the subclass estimators on
elements of mixed dimension.

To keep the code length from exploding, switch to FEMContext for the
FE evaluations and bookkeeping.

### DIFF
--- a/include/error_estimation/discontinuity_measure.h
+++ b/include/error_estimation/discontinuity_measure.h
@@ -81,9 +81,7 @@ protected:
    * An initialization function, for requesting specific data from the FE
    * objects
    */
-  virtual void initialize(const System& system,
-                          ErrorVector& error_per_cell,
-                          bool estimate_parent_error);
+  virtual void init_context(FEMContext &c);
 
   /**
    * The function which calculates a normal derivative jump based error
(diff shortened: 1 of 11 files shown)


jwpeterson:
FYI, summary of MOOSE test failures:
```
mesh/adapt.test............................................................................. FAILED (EXODIFF)
mesh/adapt.adapt_cycles_test................................................................ FAILED (EXODIFF)
mesh/adapt.test_time........................................................................ FAILED (EXODIFF)
preconditioners/pbp.pbp_adapt_test.......................................................... FAILED (EXODIFF)
```

roystgnr:
Oh, yeah; I'm replicating them locally.  Looks like, in problems with variables u+v, we're only getting the v contribution to error.  Trying to debug now.

roystgnr:
This should be safe to merge and should be an improvement, but I'm going to hold off until we know whether it actually solves the problem it's supposed to solve.

Diffs should be ready for review if anyone else wants to get eyes on them.

jwpeterson:
:+1: Good idea using `FEMContext` for this

jwpeterson:
> A default construction, e.g. fine_context(), should do the right thing, right?

Yes, I wasn't asking you to use std::nullptr, just to default initialize them like in the old code.

pbauman:
+1 on the use of FEMContext. Refactoring looks cleaner than the original (IMHO). :+1:

PR #588, opened 2015-06-03
Merge femparameters code

roystgnr:
We had five slightly-different forks of femparameters.{C,h} in our examples.  Rather than add a 6th for the new test I'm working on, I'd like to combine them all (along with the version I use in some private test codes).

This PR does half of the consolidation.  Getting a .C file via a BUILT_SOURCES symlink was straightforward, but I couldn't get it working for headers.  At least our five copies of femparameters.h are now identical...


roystgnr:
Merging right away; this only affects the examples/adjoints directory so there should be no way to mess up anyone else's stuff.

Issue #586, opened 2015-06-03
SparsityPattern::Build::operator() optimization


permcody:
While profiling Rattlesnake, we found a rather expensive operation in `SparsityPattern::Build::operator()`

When running a simulation with ~1300 variables the "if" statement in the following loop was taking about 50% of the time in this routine which itself was taking around 30% of the total setup time (several minutes). The issue is with using a dense matrix that is mostly zero for this operation. It should be replaced with a different data structure.

The problem in question can be found here:
`rattlesnake/examples/C5G7/diffusion_c5g7.i`

```
            for (unsigned int vj=0; vj<n_var; vj++)
              if ((*dof_coupling)(vi,vj)) // If vi couples to vj
                {
                  // Find element dofs for variable vj, note that
                  // if vi==vj we already have the dofs.
                  if (vi != vj)
```

roystgnr:
@benkirk I could swear you used the CouplingMatrix to optimize FIN-S, but I can't find it in the code.  Am I just misremembering?

It looks like we'll be able to be backwards-compatible the existing API with a sparse matrix structure.  Actually *using* the existing API won't be efficient, but it'll be an O(N_vars^2) inefficiency, not an O(N_vars^2*N_elem) inefficiency.  Even so we'll want a new API (appending sparse indices to a row, not directly modifying i,j) for newer code.

benkirk:
I don't, all the variables are coupled there. But I do rely on the variable groups, so that might be the confusion.

-Ben

On Jun 3, 2015, at 3:35 PM, roystgnr <notifications@github.com<mailto:notifications@github.com>> wrote:


@benkirk<https://github.com/benkirk> I could swear you used the CouplingMatrix to optimize FIN-S, but I can't find it in the code. Am I just misremembering?

It looks like we'll be able to be backwards-compatible the existing API with a sparse matrix structure. Actually using the existing API won't be efficient, but it'll be an O(N_vars^2) inefficiency, not an O(N_vars^2*N_elem) inefficiency. Even so we'll want a new API (appending sparse indices to a row, not directly modifying i,j) for newer code.

roystgnr:
@dknez, I see some CouplingMatrix usage in the Reduced Basis code.  Is there any in your private app codes?

[...]

PR #589, opened 2015-06-04
Sparse CouplingMatrix

roystgnr:
My attempt to optimize the hotspot found in #586.  This should be much faster in the many-sparsely-coupled-variables case, and should use much less memory in all but the most perverse cases.

This API change changes the return type of operator(), but the proxy class it uses should be backwards compatible for essentially everyone.

### DIFF
--- a/include/numerics/coupling_matrix.h
+++ b/include/numerics/coupling_matrix.h
@@ -24,13 +24,17 @@
 #include "libmesh/libmesh_common.h"
 
 // C++ includes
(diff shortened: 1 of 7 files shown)


roystgnr:
@permcody assures me the failing test wasn't libMesh related, it was just that threaded-PRNG bug they haven't tracked down.

roystgnr:
And this gives us an order-of-magnitude or two speedup in #586.

Issue #275, opened 2014-06-17
Better Error Message for getting non-existent Parameters


friedmud:
If you call `parameters.get<Real>("something")` when "something" doesn't exist you get an error message like:

```
Stack frames: 15
0: 0   libmesh_oprof.0.dylib               0x000000010b941ca9 libMesh::print_trace(std::ostream&) + 57
1: 1   libmesh_oprof.0.dylib               0x000000010b9402f0 libMesh::MacroFunctions::report_error(char const*, int, char const*, char const*) + 48
2: 2   librelap-7-oprof.0.dylib            0x000000010a2d1734 double const& libMesh::Parameters::get<double>(std::string const&) const + 372
3: 3   librelap-7-oprof.0.dylib            0x000000010a5365da Pipe::Pipe(std::string const&, InputParameters) + 1914
4: 4   librelap-7-oprof.0.dylib            0x000000010a3ba984 MooseObject* buildObject<Pipe>(std::string const&, InputParameters) + 68
5: 5   libmoose-oprof.0.dylib              0x000000010ae358ad Factory::create(std::string const&, std::string const&, InputParameters) + 173
6: 6   librelap-7-oprof.0.dylib            0x000000010a3afa67 Simulation::addComponent(std::string const&, std::string const&, InputParameters) + 151
7: 7   librelap-7-oprof.0.dylib            0x000000010a2bde82 AddComponentAction::act() + 242
8: 8   libmoose-oprof.0.dylib              0x000000010ad51110 ActionWarehouse::executeActionsWithAction(std::string const&) + 272
9: 9   libmoose-oprof.0.dylib              0x000000010ad50fa1 ActionWarehouse::executeAllActions() + 225
10: 10  librelap-7-oprof.0.dylib            0x000000010a375276 Relap7App::runInputFile() + 246
11: 11  libmoose-oprof.0.dylib              0x000000010ae95af8 MooseApp::run() + 24
12: 12  relap-7-oprof                       0x000000010a073e9a main + 106
13: 13  relap-7-oprof                       0x000000010a073e24 start + 52
14: 14  ???                                 0x0000000000000003 0x0 + 3
[0] /Users/gastdr/projects/libmesh/include/libmesh/parameters.h, line 448, compiled Jun 17 2014 at 17:16:46
ERROR: no double parameter named "Tw_transferred":
Name	 Type	 Value
---------------------
 A	 double	 0.44934
 A_func	 std::string	 
 Dh	 double	 0.002972
[...]

PR #591, opened 2015-06-05
Better error messages

roystgnr:
My crack at @friedmud's wish in #275

This should be an improvement in error reporting for every libmesh_foo_msg() macro, not just that particular Parameters case.


jwpeterson:
:+1:

PR #592, opened 2015-06-05
Guards in NewmarkSolver

pbauman:
The primary purpose of this PR is to add two guards in NewmarkSolver that were discussed in #582, namely only build the elem_solution_rate/accel (in DiffContext) if we need it and to force the user to set the initial acceleration either by directly projecting it or computing it using the helper method.

The changes for the latter guard are straight forward and are in 53b38bd.

Changes for the former ended up being a bit more involved because I can't help myself. The underlying idea is to add a `time_order()` method that we can query to figure out if we need rates and, possibly, acceleration. It occurred to me that I could've made the second order solver stuff more extensible. So, I ended up making intermediate classes for first order and second order (in time) solvers. The first order one just holds the `time_order()`. I could see the argument for just having `UnsteadySolver` default to 1 and override in the `SecondOrderUnsteadySolver`, but I think have the intermediate is more self-documenting. I moved common functionality out of `NewmarkSolver` into `SecondOrderUnsteadySolver`. After that preliminary work, `time_order()` was made pure virtual in `UnsteadySolver` and we can query it from there (see log for f2bb932 on why I chose this). All that was the first 7 commits in this PR. 

After that is the only-build-rate-structures-if-they're-needed (dd287bb). I welcome suggestions - those (a little) lengthy checks were repeated 3 times for different operations on those data structures in 3 different places; felt dirty and wrong.

### DIFF
--- a/include/solvers/adaptive_time_solver.h
+++ b/include/solvers/adaptive_time_solver.h
@@ -22,7 +22,7 @@
 
 // Local includes
 #include "libmesh/system_norm.h"
-#include "libmesh/unsteady_solver.h"
+#include "libmesh/first_order_unsteady_solver.h"
 
 // C++ includes
 
@@ -48,13 +48,13 @@ class System;
 
 // ------------------------------------------------------------
 // Solver class definition
-class AdaptiveTimeSolver : public UnsteadySolver
+class AdaptiveTimeSolver : public FirstOrderUnsteadySolver
 {
 public:
   /**
    * The parent class
    */
-  typedef UnsteadySolver Parent;
+  typedef FirstOrderUnsteadySolver Parent;
 
   /**
    * Constructor. Requires a reference to the system
--- a/include/solvers/unsteady_solver.h
+++ b/include/solvers/unsteady_solver.h
@@ -119,6 +119,13 @@ public:
   virtual Real error_order () const = 0;
 
   /**
+   * Returns the maximum order of time derivatives for which the
+   * UnsteadySolver subclass is capable of handling. E.g. EulerSolver
+   * will have time_order = 1 and NewmarkSolver will have time_order = 2
+   */
+  virtual Real time_order () const = 0;
+
+  /**
    * @returns the old nonlinear solution for the specified global
    * DOF.
    */
--- a/include/solvers/unsteady_solver.h
+++ b/include/solvers/unsteady_solver.h
@@ -123,7 +123,7 @@ public:
    * UnsteadySolver subclass is capable of handling. E.g. EulerSolver
    * will have time_order = 1 and NewmarkSolver will have time_order = 2
    */
-  virtual Real time_order () const = 0;
+  virtual unsigned int time_order () const = 0;
 
   /**
    * @returns the old nonlinear solution for the specified global
--- a/src/systems/diff_context.C
+++ b/src/systems/diff_context.C
@@ -18,6 +18,8 @@
 
 #include "libmesh/diff_context.h"
 #include "libmesh/diff_system.h"
+#include "libmesh/fem_system.h"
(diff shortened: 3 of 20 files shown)


include/solvers/adaptive_time_solver.h
```diff
@@ -22,7 +22,7 @@
 
 // Local includes
 #include "libmesh/system_norm.h"
-#include "libmesh/unsteady_solver.h"
```
roystgnr: Why can't this still generalize to either UnsteadySolver type?

include/solvers/unsteady_solver.h
```diff
@@ -119,6 +119,13 @@ class UnsteadySolver : public TimeSolver
   virtual Real error_order () const = 0;
 
   /**
+   * Returns the maximum order of time derivatives for which the
+   * UnsteadySolver subclass is capable of handling. E.g. EulerSolver
+   * will have time_order = 1 and NewmarkSolver will have time_order = 2
+   */
+  virtual Real time_order () const = 0;
```
roystgnr: Shouldn't this be returning unsigned int?

pbauman: Oops, just copy/pasted error_order and didn't pay attention. Thanks.

src/systems/diff_context.C
```diff
@@ -18,6 +18,8 @@
 
 #include "libmesh/diff_context.h"
 #include "libmesh/diff_system.h"
+#include "libmesh/fem_system.h"
```
roystgnr: You don't need this.  Cast to DiffSystem and you can still get the TimeSolver out.  We still might implement FDMSystem one of these days!

src/systems/fem_context.C
```diff
@@ -27,6 +27,7 @@
 #include "libmesh/mesh_base.h"
 #include "libmesh/quadrature.h"
 #include "libmesh/system.h"
+#include "libmesh/fem_system.h"
```
roystgnr: Likewise use DiffSystem here.

include/solvers/adaptive_time_solver.h
pbauman: It might be able to, I was just being lazy.

I don't understand why `AdaptiveTimeSolver::advance_timestep()` doesn't just call `core_time_solver->advance_timestep`. This looks like the only place I would have to touch to have `AdaptiveTimeSolver` just subclass `UnsteadySolver`.

src/systems/diff_context.C
pbauman: Good call, thanks.

include/solvers/adaptive_time_solver.h
roystgnr: You know?  I don't understand that either.  I probably mistakenly added some unfortunate redundancy when factoring out TwostepTimeSolver, and the correct solution is probably to just hand off to the core_time_solver.

On the other hand, "I don't understand that either" means Chesterton's fence applies here.  I've got a bunch of private apps that use this class, but we *don't* yet have any regression test coverage of it.  So let's leave it the way you've got it here, and we can reconsider refactoring it later.

roystgnr:
Okay; the fem_system.h->diff_system.h changes and the Real->uint change are the only issues I can see.  Everything else looks good.

include/solvers/adaptive_time_solver.h
pbauman: OK, sounds like a plan. I'll open an issue summarizing this once I finish up the other changes.

pbauman:
OK, I fixed up those bits and rebased on master. Once MooseBuild is happy, I'll merge this.

roystgnr:
:+1:

PR #593, opened 2015-06-05
Metis graph improved for mixed-dimensional meshes

roystgnr:
We now add links between any boundary element and its interior neighbors.

### DIFF
--- a/src/partitioning/parmetis_partitioner.C
+++ b/src/partitioning/parmetis_partitioner.C
@@ -46,6 +46,12 @@ extern "C" {
 #endif
 
 
+// Hash maps for interior->boundary element lookups
+#include LIBMESH_INCLUDE_UNORDERED_MAP
+#include LIBMESH_INCLUDE_HASH
+LIBMESH_DEFINE_HASH_POINTERS
+
+
 namespace libMesh
 {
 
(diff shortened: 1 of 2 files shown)


src/partitioning/parmetis_partitioner.C
```diff
@@ -410,6 +416,52 @@ void ParmetisPartitioner::build_graph (const MeshBase& mesh)
...
+  // the connectivity between them and interior elements.  We can find
+  // interior elements from boundary elements, but we need to build up
+  // a lookup map to do the reverse.
+
+  typedef LIBMESH_BEST_UNORDERED_MAP<const Elem *, const Elem *>
+    map_type;
+  map_type interior_to_boundary_map;
+
+  {
+    MeshBase::const_element_iterator       elem_it  = mesh.active_elements_begin();
```
pbauman: Should these be local for ParMetis?

roystgnr: I don't think so.  We might have non-local boundary elements that link to local interior elements.

pbauman: Ah, OK, good point.

roystgnr:
This seems to be working; I'm going to run some more tests on private apps and then merge this afternoon unless there are objections first.

Issue #550, opened 2015-04-29
hashtable weirdness in src/utils/topology_map.C


benkirk:
I've got my buildbot machine that runs a bunch of old compilers on a CentOS 6.6 host, and it's choking on some new code in topology_map.C. I'm sorry this buildbot is not publicly visible, but here's the bit:
@roystgnr, any thoughts?  Intel 12.1+ and GCC 4.4+ are good...

GCC 3.4:

```pre
/usr/lib/gcc/x86_64-redhat-linux/3.4.6/../../../../include/c++/3.4.6/ext/hashtable.h: In member function `size_t __gnu_cxx::hashtable<_Val, _Key, _HashFcn, _ExtractKey, _EqualKey, _Alloc>::_M_bkt_num_key(const _Key&, size_t) const [with _Val = std::pair<const std::pair<libMesh::dof_id_type, libMesh::dof_id_type>, libMesh::dof_id_type>, _Key = std::pair<libMesh::dof_id_type, libMesh::dof_id_type>, _HashFcn = __gnu_cxx::hash<std::pair<libMesh::dof_id_type, libMesh::dof_id_type> >, _ExtractKey = std::_Select1st<std::pair<const std::pair<libMesh::dof_id_type, libMesh::dof_id_type>, libMesh::dof_id_type> >, _EqualKey = std::equal_to<std::pair<libMesh::dof_id_type, libMesh::dof_id_type> >, _Alloc = std::allocator<unsigned int>]':
[...]

PR #597, opened 2015-06-05
gcc-3.4 fixes

roystgnr:
This fixes the compilation error Ben noticed in #550 for me.

It also fixes a compilation error in the new coupling_matrix.h - it looks like the previous code was correct C++11 but C++98 was more restrictive about nested class access to private things accessible in the enclosing class?

A few warning fixes are thrown in too.  We'd have run into these with newer compilers too eventually once we cranked up the warning level again.

### DIFF
--- a/include/numerics/coupling_matrix.h
+++ b/include/numerics/coupling_matrix.h
@@ -38,6 +38,8 @@ class CouplingAccessor;
 
 class ConstCouplingRow;
 
+class ConstCouplingRowConstIterator;
+
 /**
  * This class defines a coupling matrix.  A coupling
  * matrix is simply a matrix of ones and zeros describing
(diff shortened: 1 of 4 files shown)


roystgnr:
Any ideas why that build is hanging?

permcody:
Just slow. Do you believe there is a problem we need to look into? It's
possible the server was a bit overloaded but we usually keep the loads at a
reasonable level. It's always possible that your change has introduced some
bad behavior though. Let me know if you want me to investigate or just
invalidate to see how long it takes a second time.

roystgnr:
Naah, wasn't expecting any possibility of a problem with the patch; just wondering if MooseBuild was having hardware problems again or just overloaded by last-minute PRs.

PR #598, opened 2015-06-08
BoundaryInfo::add_elements()

roystgnr:
Since we should now support fun tricks with mixes of "interior" and "boundary" elements on the same mesh, let's add an easy way to put them into the same mesh.

This passes the FIN-S test suite, which was at least exhaustive enough of BoundaryInfo::sync() to catch bugs in my initial refactoring attempts.

### DIFF
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -91,12 +91,11 @@ public:
   void clear ();
 
   /**
-   * Close the data structures and prepare for use.
-   * Synchronizes the \p boundary_mesh
-   * data structures with the \p mesh data structures.
-   * Allows the \p boundary_mesh to be used like any other mesh.
-   * Before this is called the \p boundary_mesh data structure is
-   * empty.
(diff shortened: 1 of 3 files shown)


roystgnr:
No, wait; I can still get this to trigger a regression.

roystgnr:
Looks like that was actually a bug introduced in #584

Everything seems to be working again; I'll merge tonight absent objections.

jwpeterson:
IMO these three patches should be squashed together.  The first is a placeholder, the second adds implementation with bugs, the third fixes those bugs.

roystgnr:
I'll squash the first two together.  But I'm afraid the third doesn't fix bugs from the first two, it fixes a bug I missed from a previous PR...

jwpeterson:
> I'll squash the first two together. But I'm afraid the third doesn't fix bugs from the first two, it fixes a bug I missed from a previous PR...

I see, sorry, I missed your final comment there, somehow. Sounds good.

PR #603, opened 2015-06-10
Boundary elements fixes, tests

roystgnr:
This should be safe to merge as-is if MooseBuild agrees, but I intend to do some more private app testing and add an examples/ test case before merging.


jwpeterson:
There are a lot of failing builds on Moosebuild at the moment... we are in the middle of upgrading everything to PETSc 3.6.0 that was just released.  In theory, the libmesh PR tests should pass, though.

roystgnr:
I assume misc/exception.parallel_exception_jacobian_transient failure is a false positive?

jwpeterson:
.

> On Jun 10, 2015, at 10:43 PM, roystgnr <notifications@github.com> wrote:
> 
> I assume misc/exception.parallel_exception_jacobian_transient failure is a false positive?
> 

Yeah, it's one of the new tests enabled by petsc 3.6, but I thought I had fixed it.  Taking a closer look today..
.

jwpeterson:
> Taking a closer look today...

Hmmm, so your PR tested against idaholab/moose@1fc6632, which was before my fix was merged.  I invalidated the failed build, so it will run again.

jwpeterson:
> Passed on: linux-gnu

Looks like you are all good.

roystgnr:
This is matching analytical results for me and should be good to go as soon as MooseBuild agrees.

roystgnr:
Okay, that was a real bug.  Never caught it because it only affects the SerialMesh case.

roystgnr:
Hopefully that caught everything.

roystgnr:
Okay, this is also still passing with ParallelMesh and with FIN-S.  I think we're good.

PR #606, opened 2015-06-17
Reduced Basis change: Use Cap'n Proto for writing/reading

dknez:
Updated Reduced Basis (RB) code to use the Cap'n Proto (CP) serialization library. This is a nicer approach than the "legacy" writing format that involved writing out a bunch of small Xdr files.

Note that this PR only touches the RB code (the large number of files affected is because of all the Makefile.in's that are modified due to the new configure options related to Cap'n Proto).

I've tested the PR and it works well for me (e.g. all the RB examples work).

I think the only remaining question is what to do about the schema file, rb_data.capnp, and the source files that are generated from it. As I mentioned on the mailing list, I think it's a bit problematic to commit the generated source files to the repo because CP will throw an error if the generated files don't match the version of the library that is linked to. As a result, perhaps it would be feasible to do something like this instead:
- If we configure with CP, then run "capnp compile -oc++ rb_data.capnp" during configuration. This generates rb_data.capnp.h and rb_data.capnp.c++
- In rb_data.capnp.c++, change #include "rb_data.capnp.h" to #include "libmesh/rb_data.capnp.h"
- Wrap rb_data.capnp.c++ and rb_data.capnp.h in #if defined(LIBMESH_HAVE_CAPNPROTO)
- Rename rb_data.capnp.c++ to rb_data.capnp.C and move it to src/reduced_basis

This could all be done with a simple script, could the script be run at configure time?

Also, note that I wrote capnproto.m4 by copying and modifying other m4 files. I'm not experienced with m4 though, so it'd be good to get a quick review of that. One thing that confused me was that I had to use AC_CHECK_FILE instead of AC_CHECK_HEADER to check if a particular header file was present... AC_CHECK_HEADER always returned false, even if the header was present, not sure what happened there.


dknez:
Oops, need some more #ifdefs, I'll fix that

dknez:
@jwpeterson any thoughts about the CP configure issues that we discussed on the mailing list? I'd be happy to commit the generated code to the repo if you think that'd be easier.

jwpeterson:
> One thing that confused me was that I had to use AC_CHECK_FILE instead of AC_CHECK_HEADER

`AC_CHECK_FILE` is good to avoid anyway, since it doesn't work when cross-compiling.  I think it's better to just use
```
if (test -r /path/to/file)
```
to check for the existence of a file in general.

> This could all be done with a simple script, could the script be run at configure time?

Hmm... I don't know of any cases where we currently run a separate script during configure.  It seems like more of a Makefile thing to me.  Using the AM_CONDITIONALs, we could possibly put something in contrib/capnproto that would do what you describe above to generate the schema file at build time if capnproto is enabled.  I'll look at it a bit more...

friedmud:
I guess I'll have to vote :-1: on this.

The extra burden and complication of all of this preprocessing is not worth it.  Every thing like this that gets added to the libMesh build makes it more brittle.

jwpeterson:
Replaced by #613

PR #607, opened 2015-06-17
Fix multipointer/multiaccessor heisenbug

roystgnr:
Thanks to @vikramvgarg for catching this.

### DIFF
--- a/include/systems/parameter_multiaccessor.h
+++ b/include/systems/parameter_multiaccessor.h
@@ -75,7 +75,8 @@ public:
   {
     libmesh_assert(!_accessors.empty());
 #ifndef NDEBUG
-    T& val = _accessors[0]->get();
+    // Compare other values to the last one we'll change
+    T& val = _accessors[_accessors.size()-1]->get();
 #endif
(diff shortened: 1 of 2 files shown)


include/systems/parameter_multiaccessor.h
```diff
@@ -75,7 +75,8 @@ class ParameterMultiAccessor : public ParameterAccessor<T>
   {
     libmesh_assert(!_accessors.empty());
 #ifndef NDEBUG
-    T& val = _accessors[0]->get();
+    // Compare other values to the last one we'll change
+    T& val = _accessors[_accessors.size()-1]->get();
```
jwpeterson: `_accessors` is a vector, so `.back()`?  I haven't analyzed the rest of the code, but this seems dangerous if there's any chance that `_accessors.size()==0`...

roystgnr: If accessors.size() == 0, the assertion above will fail, and .back() won't be defined anyway.  It certainly would have been more self-documenting, though...

PR #608, opened 2015-06-18
Obey CouplingMatrix with implicit_neighbor_dofs

roystgnr:
It looked like we were collecting *all* neighbor dofs before, even in the CouplingMatrix code path.

Hopefully this will do something about the overallocation and memory use issues Sebastian Schunert discovered in the DG+CouplingMatrix case.

This is really a one-liner change but I'm trying to keep the variable name self-documenting.

I haven't tested this myself yet; relying on MooseBuild and @snschune to let me know if I'm an idiot.

### DIFF
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -2632,7 +2632,7 @@ void SparsityPattern::Build::operator()(const ConstElemRange &range)
       std::vector<dof_id_type>
         element_dofs_i,
         element_dofs_j,
-        neighbor_dofs,
+        neighbor_dofs_j,
         dofs_to_add;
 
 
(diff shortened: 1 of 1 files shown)


dknez:
I use get_dof_map().set_implicit_neighbor_dofs(false); for discontinuous basis functions with CouplingMatrix in order to reduce memory usage (e.g. see DofMap::use_coupled_neighbor_dofs). Is that related to the  memory use issues discovered in the DG+CouplingMatrix case that you referred to?

roystgnr:
You should be unaffected by the problem and by this change, @dknez.  The troublesome case here is when implicit_neighbor_dofs remains true (and when it needs to remain true; obviously if you want it false then you've got the right solution) and when you have a custom CouplingMatrix.  It looks to me like the former code was correctly restricting which variables coupled within an element but was *failing* to restrict which variables couples between a DG element and its neighbor.

roystgnr:
This fixes the overallocation problems!

dknez:
@roystgnr thanks for your comment, that makes sense.

PR #612, opened 2015-07-02
Make subdivision elements compatible with DirichletBoundary and RB

svallaghe:
Subdivision elements are C1 but their data structure is similar to C0 elements.
In order to use the DirichletBoundary with subdivision elements (not compatible until now), I added a few flags in dof_map_constraints.C so that they are treated as C0.

Also, when performing assembly on subdivision elements, "ghosts" need to be discarded and a particular quadrature rule needs to be initialized. I modified rb_construction.C so that it can handle subdivision elements.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -351,7 +351,7 @@ private:
 
     const FEContinuity cont = fe->get_continuity();
 
-    if (cont == C_ONE)
+    if ( (cont == C_ONE) && (fe_type.family != SUBDIVISION) )
       {
         // We'll need gradient data for a C1 projection
         libmesh_assert(g || g_fem);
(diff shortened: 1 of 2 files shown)


dknez:
This looks fine to me. Shall we merge it?

jwpeterson:
> On Jul 2, 2015, at 5:31 PM, David Knezevic <notifications@github.com> wrote:
> 
> This looks fine to me. Shall we merge it?
> 
Sure, thanks for taking a look.

pbauman:
I know nothing about subdivision. Could someone briefly explain why treating these as C0 constraints will still yield C1 continuity?

svallaghe:
Hi Paul,

Subdivision elements are defined on 2D triangular meshes (embedded in 3D),
and they have only one dof per node (or three dofs per node if considering
elasticity).
The C1 continuity is achieved thanks to the enlarged support of the
subdivision basis functions: each basis function is a spline defined on the
1-ring of the triangle (triangle plus its edge-neighboring elements), with
its value equal to zero on the support boundaries.

Hence the BC are not enforced using a function plus its gradient value, but
by considering a layer of "ghost" elements outside the normal boundaries,
and adding dof constraints on the four nodes directly connected with the
(real)boundary edges.

If you want more info on that topic, you can take a look at the seminal
paper by Cirak et. al.

Sylvain

pbauman:
Perfect, thanks @svallaghe

PR #613, opened 2015-07-12
Reduced Basis change: Use Cap'n Proto for writing/reading

jwpeterson:
This PR replaces #606.

Original description from @dknez:
```
Updated Reduced Basis (RB) code to use the Cap'n Proto (CP) serialization library.
This is a nicer approach than the "legacy" writing format that involved writing out a
bunch of small Xdr files.

Note that this PR only touches the RB code (the large number of files affected is
because of all the Makefile.in's that are modified due to the new configure options 
related to Cap'n Proto).

I've tested the PR and it works well for me (e.g. all the RB examples work).
```

The new PR rebases/reorganizes the previous branch and adds Autotools stuff for detecting CP installations and building the CP source from the schema file on the fly.


dknez:
:+1:

PR #618, opened 2015-07-31
Read additional data *iff available*

roystgnr:
Currently if a code tries to read a solution file while requesting additional vectors (as we do in e.g. projection.C to preserve as much data as possible), but the solution file was written without those additional vectors (as we do in e.g. many transient checkpoints when our time integrators don't need old solution vectors to restart), that code trips an assert and dies.

Now it instead leaves the additional vectors unchanged.


roystgnr:
Huh!  Let me see if I can replicate that.

Issue #617, opened 2015-07-28
Quadruple precision support is broken


roystgnr:
I added --enable-quadrupleprecision after noticing that PETSc had dropped support for long double and added support for __float128, but I've never tried it until today.

It doesn't work.  Including cmath only declares e.g. std::sqrt() for float, double, and long double, not for __float128.  Anything trying to do sqrt(someLibmeshReal) fails to compile.

The only obvious fix I see is to create a libmesh_cmath.h shim which includes quadmath.h and defines all the overloads e.g. `std::sqrt(__float128 foo) { return sqrtq(foo); }`  We then replace inclusion of cmath with inclusion of libmesh_cmath.h throughout the library.

That's a bit invasive.  Anyone have any better ideas?

jwpeterson:
[Boost](http://www.boost.org/doc/libs/1_56_0/libs/multiprecision/doc/html/boost_multiprecision/tut/floats/float128.html) has some stuff for float128, and it also claims to work with Intel.  Would that help?

roystgnr:
That's no less invasive but it's definitely more featureful.  Thanks!

I'm probably going to make only the minimal changes to get stuff compiling on gcc for now, but refactoring into libmesh_cmath.h should make the switch to boost relatively trivial.  (and as a thank-you for refactoring, we'll now get to read about how "PGI and Sun Studio CC suck" in only one file rather than 30)

roystgnr:
"Minimal changes to get stuff compiling on gcc" turns out to be massive, and my initial "application giving wonky behavior that might be rounding-related" problem was solved via investigation using Eigen with long double, so I'm putting this back on the shelf for now.  Whenever we do get back to it, Boost float128 is definitely the way to go.

PR #620, opened 2015-08-06
Non-double precision fixes

roystgnr:
This is nowhere near quadruple-precision support a la #617 but it does fix a few issues that cropped up in the single- and triple-precision cases.


roystgnr:
Going to pull this right away, since MooseBuild is happy, I don't see any way the one deep change (mixed scalar vector constructor arguments) could backfire, and some of the other changes ought to be critical to some subsequent stuff I'm testing.

PR #624, opened 2015-08-12
Fixes for --enable-complex

dknez:


src/base/dof_map_constraints.C
```diff
@@ -702,7 +702,7 @@ private:
               for (unsigned int qp=0; qp<n_qp; qp++)
                 {
                   // solution at the quadrature point
-                  OutputNumber fineval = 0;
+                  OutputNumber fineval(0);
```
roystgnr: If anyone understands why *this* part of the fix worked, please speak up!  ;-)

roystgnr:
:+1: as soon as MooseBuild is happy

src/base/dof_map_constraints.C
jwpeterson: Are you guys both using the same compiler?  Which compiler?  Is it calling different constructors in the two cases?

dknez: >> gcc --version
gcc (Ubuntu 4.8.4-2ubuntu1~14.04) 4.8.4

@roystgnr were you able to build with --enable-complex without my fixes?

roystgnr: No; I had the same errors.  I was using gcc 4.8.0, though, so there may be something compiler-specific going on.

roystgnr: Hmm... It's got to do with the Scalar1/Scalar2/Scalar3 stuff I put in recently to handle initialization from mixes of e.g. double literals and long double Reals.  The compiler probably can't infer what Scalar2 and Scalar3 are in the defaults case, so if it can't directly cast from int to T it needs its hand held.

PR #625, opened 2015-08-12
Enable ParallelMesh repartitioning

roystgnr:
And add bugfixes for everything I've found so far that gets triggered by ParallelMesh repartitioning.

This probably isn't working yet, but it's passing enough of my tests that I'd like MooseBuild to hammer at it too.

### DIFF
--- a/include/mesh/mesh_tools.h
+++ b/include/mesh/mesh_tools.h
@@ -433,6 +433,8 @@ void libmesh_assert_valid_refinement_tree (const MeshBase &mesh);
 /**
  * A function for verifying that neighbor connectivity is correct (each
  * element is a neighbor of or descendant of a neighbor of its neighbors)
+ * and consistent (each neighbor link goes to either the same neighbor
+ * or to a RemoteElem on each processor)
  */
 void libmesh_assert_valid_neighbors (const MeshBase &mesh);
 #endif
--- a/include/parallel/parallel_implementation.h
+++ b/include/parallel/parallel_implementation.h
@@ -34,6 +34,22 @@ namespace Parallel {
 
 #ifdef LIBMESH_HAVE_MPI
 
+#ifndef NDEBUG
+#define libmesh_assert_mpi_success(error_code) \
+  do { \
+    if (error_code != MPI_SUCCESS) \
+      { \
+        char error_string[MPI_MAX_ERROR_STRING+1]; \
+        int resultlen; \
+        MPI_Error_string(error_code, error_string, &resultlen); \
+        libmesh_assert_equal_to_msg(error_code, MPI_SUCCESS, error_string); \
+      } \
+    } \
+  while(0)
+#else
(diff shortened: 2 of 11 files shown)


roystgnr:
Not passing all my tests yet after all.

friedmud:
Just 1... but it's definitely related!  ;-)

roystgnr:
That was a nasty bug I fixed in f036430.  If this PR isn't in working shape soon I'll put together a separate PR to get that into master ASAP.  I suspect it's the cause of our past "why are packed ranges breaking with some mpich2 installs!?" frustration.

roystgnr:
That last exodiff failure actually looks like a false positive to me.  Of course if we turn on repartitioning and it happens then we're going to get different element counts on each partition.

I'll put an issue in for Moose to add skip_repartitioning to that test.  If nothing else fails (including on my side; I'm definitely hitting this code much harder than MooseBuild) then I'd like to merge.

roystgnr:
GRINS and libMesh tests all pass for me with ParallelMesh on 4 and 12 processors.

roystgnr:
MOOSE tests pass for me once I disable skip_partitioning as in https://github.com/idaholab/moose/pull/5565 - I'd like to merge this tomorrow if there are no objections.

jwpeterson:
I can merge idaholab/moose#5565 and invalidate your test results so we can try again?  Just double-check that this PR is green before we merge?

roystgnr:
Hmm... on the one hand, it looks like that ought to be safe.  5565 is green even without the libmesh module updated to include this PR.

On the other hand, that *shouldn't* be safe!  5565 should be failing right now!  The nemesis_repartitioning_test is comparing against a gold standard that was generated with repartitioning on, but since I haven't updated the libmesh submodule MooseBuild is running the tests with repartitioning implicitly off.  I'm not sure how that exodiff isn't screaming bloody murder.

permcody:
It's definitely showing that it ran and completed successfully with [MIN_CPUS=4] so everything looks normal there. Let me see if I can replicate on my desktop.

roystgnr:
Wait a minute.  Looking more closely, we would *expect* to see a successful run there.  I assumed that the nemesis_repartitioning_test would do some repartitioning (AMR, whatever) because repartitioning is what had thrown the original test off.  But I *also* removed some unnecessary repartitioning in https://github.com/roystgnr/moose/commit/24190ccd753855d355957181997c6699709d0b80 and it looks like that was the *only* repartitioning that was being done in the Nemesis tests.  So with no repartitioning being done, there's no difference between the repartitioning-enabled and -disabled runs.

roystgnr:
Let me add an AMR step or something to the Moose test so we actually get some coverage here.

permcody:
OK sounds good, FWIW - I just confirmed the same result on my desktop so at least that's sane.

roystgnr:
Okay, now the new exodiff that should be failing is failing, and nothing else is.

PR #626, opened 2015-08-12
Added MeshCommunication::check_for_duplicate_global_indices. 

dknez:
This just calls some of the debugging code that was commented out in MeshCommunication::assign_global_indices.

### DIFF
--- a/include/mesh/mesh_communication.h
+++ b/include/mesh/mesh_communication.h
@@ -155,6 +155,11 @@ public:
    */
   void assign_global_indices (MeshBase &) const;
 
+  /**
+   * Throw an error if we have any index clashes in the numbering used by
+   * assign_global_indices.
+   */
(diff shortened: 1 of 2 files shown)


dknez:
This provides the Hilbert index clash check in a separate function call, like @jwpeterson suggested.

jwpeterson:
Looks good.  So this isn't automatically called (say in DEBUG mode or something) since it's pretty slow?

dknez:
Yeah, it's too slow to call automatically, since it's O(N^2).

roystgnr:
Doing a sort would get that down from O(N^2) to O(N log N), cheap enough for ```#ifdef DEBUG```

To be really useful we'd want to check more than just local against local indices, though, and IIRC our parallel sorting doesn't work when duplicates exist.

PR #627, opened 2015-08-12
Fix a few more higher-precision issues

jwpeterson:
Following up @roystgnr's #623 with a few more fixes.

69f3a8ccbb should only be merged if the experimental re-enabling of 6th-8th order 3D monomial quadrature rules pass Roy's triple-precision regression tests, otherwise it should be dropped.


roystgnr:
Thanks for catching these!  I'll probably revert 002bc6c whenever I get serious about #617, though.  The answer to "sqrt(Real) doesn't work" is "make it work", not "try to avoid it".  If gcc hasn't gotten around to proper C++ float128 support by that time then we'll just get those overloads via boost.

PR #628, opened 2015-08-14
Make slepc.m4 detect SLEPc 3.6.0

jwpeterson:
The build system won't test this, but it works for me so if @dknez can confirm it works I think it can be merged.


dknez:
This works for me too, thanks! I suggest we go ahead and merge.

PR #629, opened 2015-08-14
Miscellaneous m4 file updates

jwpeterson:
This branch deletes tabs from all m4 files, and adds to the previous slepc.m4 by trying to compile a test code that uses SLEPc.

### DIFF
--- a/configure
+++ b/configure
@@ -2014,7 +2014,7 @@ Optional Packages:
                           Path to X11 header files. E.g. /usr/include but
                           _not_ /usr/include/X11
   --with-metis=<internal,PETSc>
-                          metis to use. interal: build from contrib. PETSc:
+                          metis to use. interal: build from contrib, PETSc:
                           rely on PETSc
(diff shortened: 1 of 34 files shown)


roystgnr:
Should we delete tabs from files that have a third-party source?  Seems like that would just make it harder to merge to/from upstream later.

jwpeterson:
> Should we delete tabs from files that have a third-party source? Seems like that would just make it harder to merge to/from upstream later.

Hmm... I can revert the tab deletions in: 

* ax_boost_base.m4
* ax_tls.m4
* blas.m4
* coverage.m4

but I'd prefer to try and get the changes into upstream instead.  These all seem to be in the autoconf-archive except for coverage.m4, which seems to be from ICES.  Do you accept pull requests?

roystgnr:
Ha!  We're using svn, so if by "pull requests" you mean "emails to Roy", then yes.  Don't bother reverting that one.

PR #633, opened 2015-08-19
Reorder parens in assert

dschwen:
I tried to squash a compile warning here and this ends up being a logic change. I believe the parenthesis was set wrong here. My guess is that the ```&&``` is a guard for both terms combined with ```||```.


roystgnr:
Shoot, yes, your version is correct.  Thanks!

friedmud:
Hah - I saw this same warning today when compiling in debug mode using
Clang... it's a warning I've never seen before...

PR #630, opened 2015-08-16
Make libmesh work with debug PETSc

jwpeterson:
This branch has now been reorganized and rebased, and I'm pretty confident it's working.  @roystgnr and @pbauman, if you want to try your own apps with this branch (PETSc debugging enabled or not) I'd appreciate it.

MOOSE also needs to be updated to work with this version of libMesh (see idaholab/moose#5569).

PR #635, opened 2015-08-20
Fix "can't plot error on super coarse grids" bug; improve verification

roystgnr:
This fixes the corner-case bug that #630 introduced, as well as some bugs and lacunae in our debugging code that I caught along the way.

I heard you like debugging so I put a static assertion in our runtime assertion so we can debug while we debug.


permcody:
I believe the proper verbiage is "Yo Dawg ..."

PR #641, opened 2015-08-26
Fix broken MPI behavior

### LINKED ISSUE #631: Issue with communication of vector<pair<Real, Point> >
Discovered in idaholab/moose#5513.

Run the test code with `mpiexec -np 2 ./allgather-opt --keep-cout --redirect-stdout`

Here's the output on my system:

Output on processor 0:
```
Before:
0, (x,y,z)=(       0,        0,        0)
1, (x,y,z)=(       1,        1,        1)
2, (x,y,z)=(       2,        2,        2)
After:
0, (x,y,z)=(       0,        0,        0)
1, (x,y,z)=(       1,        1,        1)
2, (x,y,z)=(       0,        0,        0)
0, (x,y,z)=(       2,        2,        2)
2, (x,y,z)=(       4,        4,        0)
0, (x,y,z)=(       0,        0,        0)
```

Output on processor 1:
```
Before:
0, (x,y,z)=(       0,        0,        0)
2, (x,y,z)=(       2,        2,        2)
4, (x,y,z)=(       4,        4,        4)
After:
0, (x,y,z)=(       0,        0,        0)
1, (x,y,z)=(       1,        1,        1)
2, (x,y,z)=(       2,        0,        0)
0, (x,y,z)=(       2,        2,        2)
2, (x,y,z)=(       4,        4,        0)
0, (x,y,z)=(       0,        0,        0)
```

The "Before" output is as I would expect: the Real value and the Point components should all have the same values.  The "After" output on both processors is seriously messed up.

Test code:
```
#include "libmesh/libmesh.h"
#include "libmesh/point.h"
#include "libmesh/parallel.h"
#include "libmesh/parallel_algebra.h"

using namespace libMesh;

{
  LibMeshInit init(argc, argv);

  // allgather a std::vector<std::pair<Real, Point> > onto all processors
  // This is a test code for allgathering a std::vector<std::pair<Real, Point> >
  // See MOOSE #5513.
  {
    unsigned rank = init.comm().rank();
    std::vector<std::pair<Real, Point> > vec(3);
    for (unsigned i=0; i<vec.size(); ++i)
      {
        // The Real entry and the Point entry will all consist of the same value for a given i.
        unsigned val = i*(rank+1);
        vec[i].first = val;
        vec[i].second = Point(val, val, val);
      }

    // Print before allgather
    std::cout << "Before:" << std::endl;
    for (unsigned i=0; i<vec.size(); ++i)
      libMesh::out << vec[i].first << ", " << vec[i].second << std::endl;
    
    init.comm().allgather(vec);

    // This will only be printed on processor 0 unless you --keep-cout.
    std::cout << "After:" << std::endl;
    for (unsigned i=0; i<vec.size(); ++i)
      libMesh::out << vec[i].first << ", " << vec[i].second << std::endl;
  }

  return 0;
}
```

roystgnr:
The third commit here fixes #631 on my OpenMPI 1.6 install, my OpenMPI 1.7 install, and (for the first time ever!) my MPICH2 3.1.3 install.  Wider testing before we merge would be greatly appreciated.

### DIFF
--- a/include/parallel/parallel.h
+++ b/include/parallel/parallel.h
@@ -294,6 +294,8 @@ public:
 #ifdef LIBMESH_HAVE_MPI
   DataType (const DataType &other, unsigned int count)
   {
+    // FIXME - if we nest an inner type here will we run into bug
+    // https://github.com/libMesh/libmesh/issues/631 again?
     MPI_Type_contiguous(count, other._datatype, &_datatype);
     this->commit();
   }
--- a/include/parallel/parallel.h
+++ b/include/parallel/parallel.h
@@ -105,6 +105,52 @@ class Request;
 class Status;
 
 #ifdef LIBMESH_HAVE_MPI
+
+/*
+ * Macros to test MPI return values
+ */
+
+#ifndef NDEBUG
+#define libmesh_assert_mpi_success(error_code) \
+  do \
+    { \
+      if (error_code != MPI_SUCCESS) \
+        { \
+          char libmesh_mpi_error_string[MPI_MAX_ERROR_STRING+1]; \
+          int libmesh_mpi_error_string_len; \
+          MPI_Error_string(error_code, libmesh_mpi_error_string, \
+                           &libmesh_mpi_error_string_len); \
+          libmesh_assert_equal_to_msg(error_code, MPI_SUCCESS, \
+                                      libmesh_mpi_error_string); \
+        } \
+    } \
+  while(0)
+
+// Only catch MPI return values when asserts are active.
+#define libmesh_call_mpi(mpi_call) \
+  do \
+    { \
+      unsigned int libmesh_mpi_error_code = \
+        mpi_call; \
+      libmesh_assert_mpi_success (libmesh_mpi_error_code); \
+    } \
+  while(0)
+
+#else
+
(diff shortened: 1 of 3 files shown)


jwpeterson:
For what it's worth, your branch also works for my original test code on slightly older versions of both MPIs (mpich-3.1.2, openmpi-1.6.5).

BTW, would you ever consider reorganizing parallel.h into multiple headers? I found it very difficult to find things in there while debugging, mainly because it defines so many classes (Communicator, DataType, StandardType, Request, Status) and is over 1300 lines long...

roystgnr:
I was kind of hoping that this bug was the same as the breaks-intro_ex4-with-mpich2 bug, but no, that turns out to be completely distinct.  On the bright side, delving into the seedy side of MPI for so long changed my attitude from "I wonder why that's breaking, but it works with OpenMPI so whatever" to "I WILL END ALL THE ERRORS OR DIE TRYING", which motivation seems to have been enough to let me hunt it down successfully.  I've added a fix which works for me here.

The fact that parallel.h is not organized into multiple headers is an insane artifact of its "let's add another useful bit... ooh, let's add another useful bit" history.  It should definitely be refactored, although we should leave parallel.h around as an "include all the others" header for backwards compatibility.

PR #642, opened 2015-08-26
Add new parameter to MeshExtrusion for user-defined sub ids

permcody:
This PR adds the capability for users to define how subdomain ids are assigned to new elements created during mesh extrusion.

I have an example in MOOSE ready to go but I'd be happy to create one in libMesh as well. Do we have anything else using extrusion in the examples right now?


roystgnr:
Nothing in the examples currently uses build_extrusion.  If you'd add one that would be great, thanks!

jwpeterson:
Just add a cppunit test if you want to add something...

roystgnr:
Agreed, no need for a full-on-example here.

jwpeterson:
I'm editing this branch to remove the multi_block.e mesh file which you don't appear to be using any more.  Will merge by hand.

permcody:
Thanks. Forgot to rip that out

jwpeterson:
Merged in 0dd5ae7.

PR #643, opened 2015-08-27
Add more options for disabling GDB backtrace

jwpeterson:
We have found that GDB backtraces cause issues with MOOSE's `run_tests` script on some OS's (see idaholab/moose#4201) so this branch provides both a compile time and runtime method for disabling them.  If GDB backtraces are enabled at compile time (the default) they can be turned off by passing `--no-gdb-backtrace` on the command line.

### DIFF
--- a/configure
+++ b/configure
@@ -1984,7 +1984,8 @@ Optional Packages:
   --with-sysroot=DIR Search for dependent libraries within DIR
                         (or the compiler's sysroot if not specified).
   --with-gdb-command=commandname
-                          command to invoke gdb
+                          Specify command to invoke gdb. Use
+                          --without-gdb-command to disable GDB backtraces.
   --with-boundary-id-bytes=<1|2|4|8>
                           bytes used per boundary side per boundary_id [2]
   --with-dof-id-bytes=<1|2|4|8>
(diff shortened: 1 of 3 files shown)


roystgnr:
Can't we already disable gdb at compile time by configuring with --without-gdb-command?

roystgnr:
There's also already a run time "--gdb=a_different_command" option, but I don't think that works to disable it entirely, just to select a different version.

jwpeterson:
> There's also already a run time "--gdb=a_different_command" option, but I don't think that works to disable it entirely, just to select a different version.

I saw that, but that still incurs a system call and didn't strike me as immediately obvious to users as the "official" way to disable gdb backtraces...

roystgnr:
Yeah, I definitely like "--no-gdb-backtrace" at runtime better than "I wonder if --gdb=false would work?"  The configure option still feels redundant, though.  Could we just add some better documentation in configure --help of the "without" version?

jwpeterson:
As it stands now, configuring with `--without-gdb-command` sets GDB_COMMAND to "gdb"!  Or am I misunderstanding how `AC_ARG_WITH` works?  I don't see a `--gdb=foo` option in our configure scripts.

jwpeterson:
> Or am I misunderstanding how AC_ARG_WITH works?

Yes, I am misunderstanding... `--without-gdb-command` sets `GDB_COMMAND` to "no".  So we could use this something like:
```
#if GDB_COMMAND != "no"
  if (!libMesh::on_command_line("--no-gdb-backtrace"))
    gdb_worked = gdb_backtrace(out_stream);
#endif
```
Except that you can't do string comparisons with the postprocessor...

jwpeterson:
Should now be fixed to not introduce redundant configure flags.

roystgnr:
Thanks!  :+1:

PR #645, opened 2015-08-31
Remove custom BLAS and LAPACK autoconf tests.

jwpeterson:
These were out of date from the upstream autoconf-archive repository,
and did not seem to be used internally by other m4 scripts.  I think
we are opening up an unnecessary can of worms by giving people the
ability to specify their own BLAS/LAPACK independently of the ones the
linear solver package is using...  Also note that, if you don't have
PETSc, then none of our DenseMatrix routines that use BLAS/LAPACK are
turned on anyway and I don't believe we have any other direct calls to
BLAS/LAPACK in the library.

I also checked the history of these files all the way back through
SVN, and it appears they were originally added to our catch-all
aclocal.m4 file in a big commit by @benkirk way back in 2003, but were
not specifically used at that time, and haven't been used since.  As a
final nail in the coffin, note that removing these files and running
bootstrap results in no changes to any autoconf-controlled files.


roystgnr:
Ha!  I think I've noticed these every now and then, but assumed that they were support for your DenseMatrix backing options, and then not given them a second glance.

PR #646, opened 2015-09-03
Handle numeric jacobians in adjoint dirichlet case

roystgnr:
Previously if the element_residual had reported a failure to compute
analytic jacobians we'd just ignored it and moved on

Bugfix (and bug discovery) from @vikramvgarg


pbauman:
:+1:

PR #647, opened 2015-09-03
Make perturbations relative to parameter size

roystgnr:
This works better for @vikramvgarg when taking sensitivities to small
turbulence model parameters.  But this is kind of a stopgap until we
find time to add support for configurable perturbation sizes and
analytic sensitivities.


pbauman:
:+1:

PR #649, opened 2015-09-10
Add ghost_elements_{begin,end} iterators

jwpeterson:
This branch also refactors all the SerialMesh and ParallelMesh iterator accessor instantiations to use macros, which removes about 2500 lines of code and makes it easier to add new iterators in the future.


roystgnr:
Sorry, was a little busy yesterday; I meant to rave about how much I liked the refactoring before you merged.

jwpeterson:
Do you know of any compilers that don't properly support `__VA_ARGS__`?  I can't imagine such a compiler would be good for much... but the new macros use that while nothing else in the library did before.

jwpeterson:
Hmm... looks like officially it's C99 and C++11 https://en.wikipedia.org/wiki/Variadic_macro, but I can't imagine a C++03 compiler that is capable of compiling C99-compliant code wouldn't work with it...

roystgnr:
Hmmm... that's dangerous reasoning.  There are a lot of cases of "this code looks like it should work, so it was accepted by early C++ compilers, but it's technically not allowed by the C++ standard, so it got taken out of more recent C++ compilers".

I don't *think* __VA_ARGS__ was ever (or ever will be) such a case, though: the double leading underscore means it's a reserved-for-implementations identifier, so a C++03 compiler can support it without technically violating the C++03 standard.

permcody:
ghost1
ghost2

Looks nice!

pbauman:
Neat pictures!

roystgnr:
No kidding.  Is that a Moose feature?  I want "plot_highlighted_range(filename, iter_begin, iter_end)" in libMesh!

permcody:
It's not automatic in MOOSE, but it could be! I was just thinking that I wanted to add an ElementHighlightRange AuxKernel after creating this code in an existing class. We already have several other AuxKernels that make fields like processor regions and so forth. Why not add this one as a debugging tool too?

This particular field though takes one extra step since the regions colored here are only the ghost elements which _aren't_ owned by the processor that would normally highlight that value. I got around that by adding all ghost elements to a set and then doing a global set union. Voila!

Issue #648, opened 2015-09-10
m4 for Trilinos 12.X and DTK 2 needs to be updated


rppawlo:
The latest version of Trilinos/Tpetra and DTK both changed to use subpackages. This breaks the m4 file detection for the generated tpetra and dtk *_config.h files.  The m4 files for trilinos need to be updated.  Will submit a pull request with fixes later today.

PR #650, opened 2015-09-10
issue #648: fix trilinos and dtk library checks for new versions.

rppawlo:
Trilinos 12.0 and DTK 2.0 changed to subpackages.  This moved the place where the config files are generated that autotools checks.  This commit adds alternate paths so that the libmesh configure passes for the latest versions.


jwpeterson:
Thanks!  We will have to take your word for it, I don't think the CI system has Trilinos installed.

rppawlo:
Thanks!

PR #651, opened 2015-09-11
Factor out ParsedFEMFunction::eval()

roystgnr:
We can simplify the code a bit and enable devel-mode checks.

This is based on the old refactoring in ParsedFunction

We're still passing the ParsedFEMFunction tests in GRINS


jwpeterson:
I'd like to merge this since we may be doing a libmesh update soon.  Any objections?

PR #653, opened 2015-09-16
Add missing chain rule derivative to plog, modify tests 

### LINKED ISSUE #652: Automatic differentiation bug for plog
The derivative of the plog function is missing the chain rule term. This impacts phase field simulations using plog in a big way (wrong residuals, wrong Jacobians). Dang!

dschwen:
Closes #652


jwpeterson:
:+1: Will merge as soon as tests pass.

dschwen:
It will require re-golding two MARMOT tests that internally use plog (the CALPHAD material system). @tonkmr.

PR #654, opened 2015-09-18
Initialize context in FEMSystem corner case

roystgnr:
This seems to be enough to fix the bugs in
https://github.com/grinsfem/grins/pull/275


roystgnr:
Okay, *now* I'm getting correct results from the GRINS cases.  I'll merge this tomorrow assuming MooseBuild is happy.

PR #656, opened 2015-09-22
Changed LIBMESH_CHKERRABORT so that it throws an exception instead of aborting.

dknez:
This allows us to recover gracefully if the solver fails for whatever reason.

TODO:
- [x] Call PetscErrorMessage to get a more informative error message string out of PETSc (I think that's what that function is supposed to do...). I'll probably inherit from SolverException to do this PETSc-specific additional checking.
- [x] Fall back on the original CHKERRABORT-based implementation if !LIBMESH_ENABLE_EXCEPTIONS.
- [x] Change the name of the macro to LIBMESH_CHKERR or something similar to make clear that we're still hooking into the PETSc error handling macro stuff, but not necessarily aborting.  Retain backwards-compatibility with the original macro for a while.


jwpeterson:
This looks good, but I'd also like to: 

* Call [PetscErrorMessage](http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Sys/PetscErrorMessage.html) to get a more informative error message string out of PETSc (I think that's what that function is supposed to do...).  I'll probably inherit from `SolverException` to do this PETSc-specific additional checking.  
* Fall back on the original `CHKERRABORT`-based implementation if `!LIBMESH_ENABLE_EXCEPTIONS`.
* Change the name of the macro to `LIBMESH_CHKERR` or something similar to make clear that we're still hooking into the PETSc error handling macro stuff, but not necessarily aborting.

I have a few other things I have to work on first but I'd like to get to this today...

dknez:
Sounds good to me.

jwpeterson:
@dknez I still need to clean up this branch a bit, but we are now calling `PetscErrorMessage` to get more information from PETSc in our exception when a non-zero error code is returned.  If you turn on the demonstration code in f064750, which hardcodes ierr=77, that example prints:
```
Caught exception during solve! e.what() = Petsc has generated inconsistent data
```
which matches the error string in PETSc.

One change is that I decided to derive from `std::exception` instead of `std::runtime_error`, as that seemed to give us more opportunity to craft an appropriate `what()` message.  Can you let me know if this is still going to work for your application?

I'm still planning to change the name of `LIBMESH_CHKERRABORT` in an upcoming commit.

dknez:
This all looks good to me, thanks for improving the PR! I'm happy if you go ahead and merge.

PR #657, opened 2015-09-22
Added SolverConfiguration class

dknez:
The idea is that this can be used to programmatically configure the solvers by making a subclass of SolverConfiguration that overrides SolverConfiguration::configure_solver(). Note that we call configure_solver in petsc_linear_solver.C before we do the solve, and the same thing can be done in other solver types if desired.

This is related to the "solver exception" PR, since we sometimes need to be able to change solver options in order to redo a solve successfully.

The SolverConfiguration class is quite empty, where the expectation is that the solver specific data is implemented in sub-classes.

### DIFF
--- a/examples/systems_of_equations/systems_of_equations_ex6/run.sh
+++ b/examples/systems_of_equations/systems_of_equations_ex6/run.sh
@@ -6,6 +6,6 @@ source $LIBMESH_DIR/examples/run_common.sh
 
 example_name=systems_of_equations_ex6
 
-options="-ksp_type cg -pc_type bjacobi"
+options=""
 
 run_example "$example_name" "$options"
--- a/examples/systems_of_equations/systems_of_equations_ex6/systems_of_equations_ex6.C
+++ b/examples/systems_of_equations/systems_of_equations_ex6/systems_of_equations_ex6.C
@@ -38,6 +38,7 @@
 #include <math.h>
 
 // libMesh includes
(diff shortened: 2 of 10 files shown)


jwpeterson:
Are you planning to add `SolverConfiguration` use to one of the examples?

dknez:
@jwpeterson I added SolverConfiguration to systems_of_equations_ex6. We have options="-ksp_type cg -pc_type bjacobi" in run.sh, but now I set those options in the code (if LIBMESH_HAVE_PETSC is defined).

Note that this SolverConfiguration stuff is similar to things like PetscLinearSolver::set_petsc_solver_type, except I think SolverConfiguration is better since it allows the user more freedom. It's not practical for libMesh to wrap all of PETSc's solver and preconditioner options in an interface like set_petsc_solver_type, but with SolverConfiguration, you can just override configure_solver and do what you need to do.

jwpeterson:
Looks like run.sh lost its execute bit:

 100755 → 100644

```
/bin/bash: line 9: ../../../../examples/systems_of_equations/systems_of_equations_ex6/run.sh: Permission denied
FAIL: run.sh
```

dknez:
Sorry about that, should be fixed now.

jwpeterson:
Looks like our CI machine ran out of space on its hard drive.  I'll try invalidating this...

jwpeterson:
I squashed together the last two commits, will merge once the testing finishes.

PR #659, opened 2015-09-22
Tidy up a couple variadic macro warnings that popped up in GCC 4.6.3 

jwpeterson:
This compiler is fairly old now, but we still have Ubuntu 12 systems so it's good to keep them happy.  Note that I also tried the: 

```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wno-variadic-macros"
...
#pragma GCC diagnostic pop
```

approach, but that did not suppress the warning, I think due to a bug in the compiler's implementation of pragmas.


roystgnr:
Where was the variadic macro?

jwpeterson:
I added some in the new mesh iterators refactoring.

PR #660, opened 2015-09-23
Better handling of corners when adding boundary ghost elements to tri…

svallaghe:
…angular meshes for subdivision elements

### DIFF
--- a/src/mesh/mesh_subdivision_support.C
+++ b/src/mesh/mesh_subdivision_support.C
@@ -249,6 +249,88 @@ void MeshTools::Subdivision::add_boundary_ghosts(MeshBase& mesh)
       Elem* elem = mesh.elem(eid);
       libmesh_assert_equal_to(elem->type(), TRI3SUBDIVISION);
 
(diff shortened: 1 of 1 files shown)


jwpeterson:
Hmm.. since I don't really understand the algorithm just by looking at it, would it be possible for you to add a unit test that verifies it works for a simple test case?

svallaghe:
The example miscellaneous11 is using that piece of code.
It's definitely not a unit test, but would it still be sufficient ?
Otherwise I'm not sure I can provide a unit test in a timely fashion.

jwpeterson:
OK, I'll take a look at misc. ex11 before/after the patch.  As long as we have some code checked in that exercises the new algorithm that's fine.

svallaghe:
Works for me! Thanks!

jwpeterson:
misc ex11 produces the outputs below before and after the patch, respectively.  I take it this is doing the right thing?

Before patch:
before_patch

After patch:
after_patch

PR #667, opened 2015-09-29
Parsed functions editing

roystgnr:
Adds new unit tests, setters, and getters to ParsedFunction and ParsedFEMFunction, as well as ParameterAccessor subclasses for interfacing with them.  This allows us to differentiate with respect to inline variables in parsed functions.

### DIFF
--- a/include/numerics/parsed_fem_function.h
+++ b/include/numerics/parsed_fem_function.h
@@ -111,7 +111,7 @@ public:
    * and if the inline variable takes the same value within any
    * subexpressions where it appears.
    */
-  Output get_inline_value(const std::string& inline_var_name);
+  Output get_inline_value(const std::string& inline_var_name) const;
 
 protected:
   // Helper function for (minor) changes to expression
@@ -119,7 +119,7 @@ protected:
 
   // Helper function for parsing out variable names
   std::size_t find_name (const std::string & varname,
-                         const std::string & expr);
+                         const std::string & expr) const;
 
   // Helper function for evaluating function arguments
   void eval_args(const FEMContext& c,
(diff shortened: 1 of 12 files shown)


roystgnr:
If https://github.com/idaholab/moose/pull/5761 passes the app tests and if nobody screams, then I plan to merge this without further delay.

There should be no changes in existing behavior here, just a bunch of refactoring and new features, but the sheer number of changes and our limited ParsedFunction test coverage has me feeling paranoid.

include/numerics/parsed_fem_function.h
```diff
@@ -48,238 +52,184 @@ namespace libMesh
...
 #ifdef LIBMESH_ENABLE_SECOND_DERIVATIVES
-    , _need_var_hess(_n_vars*LIBMESH_DIM*LIBMESH_DIM, false)
+      _need_var_hess(_n_vars*LIBMESH_DIM*LIBMESH_DIM, false),
 #endif // LIBMESH_ENABLE_SECOND_DERIVATIVES
+      _additional_vars (additional_vars ? *additional_vars :
+                        std::vector<std::string>()),
+      _initial_vals (initial_vals ? *initial_vals :
+                     std::vector<Output>())
+  {
+    this->reparse(expression);
```
jwpeterson: Looks like `reparse()` is a non-virtual function, so this works fine, but I was working on some code the other day where I got bit by trying to call a virtual `init()` function from a constructor, and learned that the vtable stuff is not active in the constructor body, and with good reason...

jwpeterson:
> If idaholab/moose#5761 passes the app tests and if nobody screams, then I plan to merge this without further delay.

:+1:

PR #668, opened 2015-10-01
Relative delta p in forward sensitivity solves

roystgnr:
Also some cleanup for readability in adjoint solves.

Can't believe we missed this when doing #647

### DIFF
--- a/src/systems/implicit_system.C
+++ b/src/systems/implicit_system.C
@@ -703,9 +703,6 @@ void ImplicitSystem::adjoint_qoi_parameter_sensitivity
  const ParameterVector& parameters_in,
  SensitivityData&       sensitivities)
 {
-  // We currently get partial derivatives via central differencing
-  const Real delta_p = TOLERANCE;
-
   ParameterVector& parameters =
     const_cast<ParameterVector&>(parameters_in);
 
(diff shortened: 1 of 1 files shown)


src/systems/implicit_system.C
```diff
@@ -675,21 +673,25 @@ void ImplicitSystem::assemble_residual_derivatives(const ParameterVector& parame
       // (R(p-dp) - R(p+dp)) / (2*dp)
 
       Number old_parameter = *parameters[p];
-      *parameters[p] -= deltap;
+
+      const Real delta_p =
+        TOLERANCE * std::max(std::abs(old_parameter), 1e-3);
```
pbauman: Just to make sure I'm clear, the `std::max` is there just to prevent `delta_p` from getting absurdly small? I guess if someone has an insanely small parameter, they need to change units?

roystgnr: Yeah, it's an ugly hack.  If someone has a parameter whose default value is 0, in the extreme case, the best we can do is "hope TOLERANCE*1e-3 works fine", because the alternative is "divide-by-zero error".

"Let users override delta_p on a per-parameter basis" would be better, and "Let users supply analytic derivatives" better still, but this works for us for now.

pbauman:
:+1:

PR #669, opened 2015-10-02
Per-variable numerical jacobian h

roystgnr:
Tested and working via GRINS


roystgnr:
Didn't fix our adjoint-solve-of-adjoint-qoi problems, but it works and it can help Newton convergence a bit.

PR #670, opened 2015-10-05
Add rpath flags for TBB to link line.

jwpeterson:
This was apparently an oversight, we were already doing it for pretty much every other package we link to.  This issue is apparently going be more important in OSX El Capitan, as they are changing the way DYLD_LIBRARY_PATH is handled there.


jwpeterson:
Thanks @andrsd for testing.  Looks like this combined with using TBB or newer is working with El Capitan.  The binary TBB 4.2 distribution (not sure about 4.3?) actually relied on `DYLD_LIBRARY_PATH` in order to work, but they have fixed that in 4.4 by properly using `-install_name` when linking their stuff on OSX.

PR #671, opened 2015-10-06
Update API in GetPot

permcody:
TL;DR;
There appears to be no way to set a variable externally (that may already exist)
without having it marked as overridden. This API change allows the user to
choose which variables appear overridden and which do not.

Full Details
I would like to support the ability for users to supply variables right in their input files that may be replaced by user-defined logic (See https://github.com/idaholab/moose/issues/4840 and https://github.com/idaholab/moose/issues/5649). This would look like the following in a normal MOOSE input file:

```
[Outputs]
  file_base = 'heat_conduction_$DATE$'
  ...
[]
```

The Parser would then give the user the opportunity to replace these variables with a callback supplying various other pieces of information from the input file or system. Some of these variables would be system defined like `$DATE$` while others would be user-defined like `$TRANSPORT_METHOD$`. To avoid the explosion of new template functions, my implementation works by peaking at each raw variable value as a string type and allowing the user to replace the value as another string type. The normal templated GetPot extraction methods would then be used to convert the input as normal. The only hitch to this whole plan is the fact that the current implementation of the private method `_set_variable()` in GetPot automatically marks all existing parameters as overridden which I am already making use of for error checking in GetPot files, specifically for these types of cases:

```
[Outputs]
   file_base = foobar.e
[]

...

[Outputs]
  file_base = junk.e  # Perfectly legal in GetPot but causes confusion
[]
```

I'd like these "replacements" to happen without triggering this override behavior. I propose either this PR which simply piggybacks on the existing `sets()` methods or adding a new group of methods called `replace()` that will do the same thing.  Any objects, comments?

### DIFF
--- a/include/base/getpot.h
+++ b/include/base/getpot.h
@@ -278,13 +278,13 @@ public:
    *   ii) from inside, use '_set_variable()' below
    */
   template<typename T>
-  inline void set(const char* VarName, const T& Value, const bool Requested = true);
+  inline void set(const char* VarName, const T& Value, const bool Requested = true, const bool Overridden = true);
 
   template<typename T>
-  inline void set(const std::string& VarName, const T& Value, const bool Requested = true);
+  inline void set(const std::string& VarName, const T& Value, const bool Requested = true, const bool Overridden = true);
 
-  inline void set(const char* VarName, const char* Value, const bool Requested = true);
-  inline void set(const std::string& VarName, const char* Value, const bool Requested = true);
+  inline void set(const char* VarName, const char* Value, const bool Requested = true, const bool Overridden = true);
+  inline void set(const std::string& VarName, const char* Value, const bool Requested = true, const bool Overridden = true);
 
   inline unsigned vector_variable_size(const char* VarName) const;
   inline unsigned vector_variable_size(const std::string& VarName) const;
@@ -539,7 +539,8 @@ private:
    */
   inline void _set_variable(const std::string& VarName,
                             const std::string& Value,
-                            const bool Requested);
+                            const bool Requested,
+                            const bool Overridden);
 
   /**
    * produce three basic data vectors:
@@ -1113,7 +1114,7 @@ GetPot::_parse_argument_vector(const STRING_VECTOR& ARGV)
           // => arg (from start to '=') = Name of variable
           //        (from '=' to end)   = value of variable
           _set_variable(arg.substr(0,equals_pos),
-                        arg.substr(equals_pos+1), false);
+                        arg.substr(equals_pos+1), false, true);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


roystgnr:
GetPot already has DBEs for allowing variables to incorporate other variables.  Why add an additional variable-identifier syntax?  If the user sets

file_base = 'heat_conduction_${DATE}'

and DATE has already been set (whether via API or include directive) then it will be substituted there.

permcody:
Oh... I claim ignorance on that one.  Let me take a look at the existing API and see if I can get it to work. I'll report back if there are any shortcomings in the existing API that I can't get around.

roystgnr:
Only shortcoming in the existing API is the lack of namespaces, which is understandable since pre-libMesh GetPot didn't have include directives.  My own include-directive-happy input decks haven't been complicated enough or collaborative enough to need namespaces, but I could definitely see that changing on projects of larger scope.

permcody:
Well on initial inspection it appears that all expansion occurs autonomously when the input file (or CLI) is initially read. This doesn't lend itself very well to user defined callbacks...  While I can definitely populate my pre-defined list ahead of time, it may not be very easy to know which new variables exist in the file until it's read. I may still want to provide a way for the GetPot object to callback to a user-defined routine.

roystgnr:
Giving the user (or middleware) a string to replace has its own problems, though:
```
radius = ${/ ${diameter}, 2} # works

radius = ${/ $diameter$, 2} # fails to convert '$diameter$' to double
```

On the other hand, I wouldn't swear you *can* populate your pre-defined list ahead of time.  DATE and TRANSPORT_METHOD are easy enough, but there might be *some* value I haven't thought of which you need for phase 2 of your simulation but which you don't know until phase 1 is done.

On the gripping hand, it's *really* nice to do all your parsing right up front at the beginning of your simulation, so you can use GetPot UFO detection before you so much as allocate a matrix.  I think I've refactored three major libMesh codes to do all the parsing up front, and although the original impetus was to get string parsing out of inner loops, each time it's been a huge win for ease of use because it also allowed us to call GetPot::unidentified_arguments() afterward.  If you hold off on calling GetPot::operator() until the middle of your simulation then you're missing a big opportunity to error-check your input files.

permcody:
Right, MOOSE does all parsing ahead of time, uses UFOs detection and overridden variables to help with error checking. I'm just looking to allow users to take it to the next level.

roystgnr:
In case I'm sounding like a nitpicky ass right now: I'm not actually complaining about this PR, just trying to help with your future direction.  In particular, I don't test for overridden variables in my own codes, and I'm not sure I can without work (possibly starting with this PR) that could distinguish between "variables overridden because someone overrode the default value from an included file, which is a common idiom" and "variables overridden within the same file, which must be a mistake".

jwpeterson:
> In case I'm sounding like a nitpicky ass right now

nitpickyasshub.com was the original name of github

PR #672, opened 2015-10-07
Fix for DirichletBoundary on non-vertex nodesets.

dknez:
We added support for DirichletBoundary conditions based on nodesets a while ago. However, I just found out that this doesn't work in cases where the nodesets are on "non-vertex nodes" (e.g. on a mid-side node on a second order mesh).

This PR fixes it for me.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -431,6 +431,9 @@ private:
         std::vector<bool> is_boundary_node(elem->n_nodes(), false),
           is_boundary_edge(elem->n_edges(), false),
           is_boundary_side(elem->n_sides(), false);
+
+        // We also maintain a separate list of nodeset-based boundary nodes
+        std::vector<bool> is_boundary_nodeset(elem->n_nodes(), false);
         for (unsigned char s=0; s != elem->n_sides(); ++s)
           {
             // First see if this side has been requested
@@ -466,7 +469,10 @@ private:
 
             for (std::size_t i=0; i != bc_ids.size(); ++i)
               if (b.count(bc_ids[i]))
+              {
                 is_boundary_node[n] = true;
+                is_boundary_nodeset[n] = true;
+              }
           }
 
         // We can also impose Dirichlet boundary conditions on edges, so we should
@@ -508,7 +514,10 @@ private:
         // hold those fixed and project boundary edges, then hold
         // those fixed and project boundary faces,
 
-        // Interpolate node values first
+        // Interpolate node values first. Note that we have a special 
+        // case for nodes that have a boundary nodeset, since we do
+        // need to interpolate them directly, even if they're non-vertex
+        // nodes.
         unsigned int current_dof = 0;
         for (unsigned int n=0; n!= n_nodes; ++n)
           {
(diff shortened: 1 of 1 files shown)


src/base/dof_map_constraints.C
```diff
@@ -431,6 +431,9 @@ private:
         std::vector<bool> is_boundary_node(elem->n_nodes(), false),
           is_boundary_edge(elem->n_edges(), false),
           is_boundary_side(elem->n_sides(), false);
+
+        // We also maintain a separate list of nodeset-based boundary nodes
+        std::vector<bool> is_boundary_nodeset(elem->n_nodes(), false);
```
jwpeterson: Even if there's no chance this code will run threaded (is there?) I'd still like to lodge my usual formal complaint against `std::vector<bool>`.

See section on [data races](http://www.cplusplus.com/reference/vector/vector-bool/) at the bottom.

dknez: OK, but I just made it consistent with the preceding lines. Do you propose to change all four of them?

dknez: P.S. If you do want to change all four lines, perhaps you could do that in a separate commit (outside of this PR)? It seems to me that is not germane to this PR.

jwpeterson: > OK, but I just made it consistent with the preceding lines.

I know, just giving @roystgnr a hard time.

> P.S. If you do want to change all four lines, perhaps you could do that in a separate commit (outside of this PR)?

Agreed, sorry didn't mean to make it sound like you were supposed to do that.

roystgnr: Hey, you know what else suffers from race conditions during multithreaded access?  "i++" on an integer.  Let me know when you've gone through the code and fixed all those.  :-P

roystgnr:
Looks good to me.

roystgnr:
:+1:

src/base/dof_map_constraints.C
jwpeterson: > "i++" on an integer.

The difference being that `i++` is a write operation, while *read-only* access into `vector<bool>` is thread-unsafe...

roystgnr: I think read-only is still supposed to be safe (in C++11, anyway; the C++03 standard on data races says "what's a thread?"); what the standard allows vector<bool> to screw up is concurrent writes to different elements.  https://stackoverflow.com/questions/16130494/are-stdmap-and-stdvector-thread-safe

jwpeterson: Thanks, didn't know about the distinction on this issue between the different standards.  I wonder if it means that operator[] access is slower for `vector<bool>` in C++11 because they have to do some locking to enforce 23.2.2.

roystgnr: I don't think any locking is needed.  Basically the way that const ```vector<bool>::operator[]``` works is to bitwise-and the appropriate char against a mask.  You can do that as many times as you want concurrently and not break anything.

With non-const ```vector<bool>::operator[]```, by contrast, under the hood ```=``` is similar to ```++``` for integers: it has to first read data (that char), then modify that data (with a bitwise-or for true or bitwise-and for false), then write the data back.  If someone else tries to do the same thing at the same time, then you risk them both getting the same initial data and then only one of the two modifications gets written back.

Issue #655, opened 2015-09-18
Add libmesh_cxx11 macros and functions


roystgnr:
In cases where std::move(x) would leave more efficient code but where (x) would work, we should have a libmesh_move(x) inline function that compiles to the former when available and the latter otherwise.  Likewise for std::forward().

The "final" and "override" keywords are useful for efficiency and compile-time error checking, respectively, but they can be safely omitted too; we should have libmesh_final and libmesh_override macros that disappear when using C++98 but give us the C++11 features when they're supported.

We should eventually do the same with noexcept/throw().  Maybe with constexpr (we can't safely use that in C++98, but we can declare with it so that C++11 user code can take advantage).  Possibly with nullptr (or is that too risky due to differing overload selection?).

Any opportunities I'm missing?  I don't see any way to use lambdas, auto, or the other really interesting features while maintaining backward compatibility.

Edit to add checklist:
- [x] Add `libmesh_make_unique` wrapping `std::make_unique`
- [ ] Add `libmesh_move` wrapping `std::move`
- [x] Add `libmesh_final` wrapping `final` specifier
- [x] Add `libmesh_override` wrapping `override` specifier
- [x] Add `libmesh_delete` wrapping `delete` specifier for deleted functions
- [ ] Use `constexpr` where appropriate.  Not possible to [emulate in C++03](http://programmers.stackexchange.com/questions/142836/achieving-forward-compatibility-with-c11).
- [ ] Add `libmesh_noexcept` which expands to `noexcept` for C++11 compilers and `throw()` for older compilers.  Use this in place of `throw()` whereever it is currently in the library.
- [x] `nullptr`: There is a C++03 compatible `nullptr_t` example in [Effective C++, 2nd ed.](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/nullptr#Solution_and_Sample_Code).
- [ ] Add `libmesh_default` wrapping the `=default` function specifier.

friedmud:
This all sounds awesome @roystgnr !  Great idea!

roystgnr:
After talking to the Dakota people recently I'm resigned to supporting C++98 (well, C++03 - if your compiler can't clear() an empty container then it can go to hell) for years to come, at least for our core features.

[...]

PR #679, opened 2015-10-15
Add support for C++11 override keyword

jwpeterson:
The `override` [keyword](http://en.cppreference.com/w/cpp/language/override) can be used to mark functions that, as the name suggests, override virtual functions from a base class.  Thus not all virtual functions are overrides, but all override functions must be virtual.

This keyword imposes no additional runtime overhead, but provides compile-time checks to ensure that you are actually overriding a virtual and haven't accidentally defined a new function. They also provide extra documentation to the reader of a class: they can now distinguish whether a given virtual function in class X overrides a function from a base of X or is defining a new virtual interface for X and its subclasses.

Refs #655.


include/fe/fe.h
```diff
@@ -686,7 +690,9 @@ class FEHierarchic : public FE<Dim,HIERARCHIC>
    * to be used in dimension \p Dim.
    */
   explicit
-  FEHierarchic(const FEType& fet);
+  FEHierarchic(const FEType& fet) :
+    FE<Dim,HIERARCHIC> (fet)
+  {}
```
roystgnr: How'd you end up catching these?

roystgnr: Wait, never mind, that's not a catch, just a reorganization.  At some point we should nail down style guidelines about when inline functions should go inside the class definition and when outside.

jwpeterson: Yeah, these changes should be in a separate patch, so they can be reverted if you want, but I highly prefer this version.  It's easier to read and about 75% less lines for one-liner constructors.

roystgnr: As long as they're in a separate commit, they're fine mixed in this PR.

jwpeterson: > As long as they're in a separate commit, they're fine mixed in this PR.

Yeah, they're in 9274862

roystgnr:
I made it up to "Sorry, we could not display the changes to this file because there were too many other changes to display."  You broke GitHub!

No problems in what I saw, though.  If the whole rest of the diff is just more add-libmesh_override-everywhere then go ahead and merge.

jwpeterson:
> No problems in what I saw, though. If the whole rest of the diff is just more add-libmesh_override-everywhere then go ahead and merge.

Thanks for the review.

PR #680, opened 2015-10-15
Amr projection refactor

roystgnr:
This needs more testing before merge.

The good news is it reduces code size.

The bad news is it probably makes mesh->mesh projections less efficient (but hopefully not noticeably).

The really good news is that it should enable mesh->mesh projections on meshes with elements of mixed dimension.

The hopefully upcoming news is that we're now about 2/3rds of the way through the refactoring that will give us arbitrary projection operators, thereby enabling geometric multigrid.


roystgnr:
I'm going to have to retract "reduces code size" before this passes testing, aren't I?

At least some of the new code might be useful for other reasons too...

dschwen:
@roystgnr will this refactor make it possible to add AMR projections that conserve the integral over the non-linear variables? We noticed that coarsening concentration fields in phase field simulations violates conservation of mass (depending on the local curvature).

roystgnr:
Make it possible?  Not yet.  Make it easier?  Probably.  We're finally getting towards the point where there's one copy of the projection algorithm instead of half a dozen, which means that if we want to allow users to select or supply different projection algorithms we can make them do so once instead of a half dozen times.

I don't even think mass conservation violations are the worst risk with coarsening in phase fields.  Conservation violations upon coarsening happen with any problem, but phase fields are special.  :-P  If you have a PDE with a limited domain of definition (e.g. Flory-Huggins free energy doesn't do so hot with sub-0% or above-100% concentrations), and you compute on a set of C1 elements a solution which is (whether entirely, or just at quadrature points) within that domain, coarsening can still leave you with a new solution which goes outside the valid range of the PDE.  I ended up solving the "mathematician's version" (i.e. the wrong version) of Cahn-Hilliard in all my transient AMR cases because you couldn't go wrong with a dumb defined-on-all-of-R quartic.

So there's definitely something worth fixing here!

The trouble is choosing the different projection algorithms.  Consider the simplest case: coarsening a pair of EDGE2 elements.  All the DoFs are nodal, so if you want continuity and locality then there's only one projection operator possible, but it violates conservation (except in the trivial case) for C0 elements, and it can violate range restrictions for C1 elements.

What did you have in mind instead?  For guaranteeing conservation all that comes to mind is "build a new system, add lagrange multiplier variables, solve global projection system".  But that can actually *break* range constraints in the previously-nearly-immune C0 case, I think.  Consider an approximation of the Heaviside function, then coarsen the elements around the discontinuity, then project (whether in L2 or H1) with a conservation constraint.  Gibbs-type effects could then push the surrounding solution past its previous maximum or minimum, depending on whether the removed node was just after or just before the discontinuity.

roystgnr:
Have I mentioned lately how much I love the Moose test coverage?  I was aware that the new algorithm still isn't ready yet (it underintegrates compared to the previous code) but I'll feel a lot better about fixing that knowing there's an exodiff that catches the problem.

roystgnr:
Damn it.  Take a look at the mesh in adaptivity/cycles_per_step/gold/cycles_per_step_out.e-s005, near the bottom left.

That little nub of refinement looks like what you get when you tell a symmetric problem to flag cells in ways that can't be satisfied symmetrically.  I'm not yet sure precisely what's going on (with cycles_per_step = 2 I'm only seeing half the grids) or how (this is the sort of problem that I'd expect from refinement by elem fraction, but you're using error fraction for this test), but the end result is that my solution from this PR looks identical to the gold standard except that the mesh has a 1-element-nub instead of a 3-element-nub.

My s004 files match a June result precisely (despite the even more asymmetric grid at that step).  I'm tempted to declare surrender and just put in a Moose PR which changes the regression test to compare that instead of s005.  There's no way I can get this refactor to match a result which might well be order-of-operations dependent.

friedmud:
Yeah - shit happens sometimes. I wouldn't worry about it ;-)

roystgnr:
That "how" is still worrying to me.  The elem_fraction option might as well be the make_shit_happen option in this case, but... hmm... I suppose error_fraction could be expected to fail symmetry at the top and bottom edges if you're using the option to scale by number of faces, and once you've got an asymmetric mesh you'll get an asymmetric solution, and any ringing in that solution might break symmetry on the interior.  Still not certain, though.

roystgnr:
The remaining test failures are timeouts in tests I shouldn't have affected.  I'm guessing a consequence of the MooseBuild->Civet upgrade shakedown?  Even compiling seemed to be going slowly.

I'll do a quick look at before-and-after PerfLog look while others have the chance to look this patch over, and I might add some AMR to fem_system_ex4 to exercise that case, but unless someone finds a problem this should be ready to merge.

permcody:
Right, Brian was trying to increase the number of active clients per machine. Clearly he went too far. Ignore that failure, and any time issue on those perf_logs.

roystgnr:
> The bad news is it probably makes mesh->mesh projections less efficient (but hopefully not noticeably).

Roy says, in the understatement of the year, after making the LAGRANGE code path ten times slower.  I'll have a fix shortly.

PR #683, opened 2015-10-19
Add missing overrides.

jwpeterson:
Clang 3.7.0 apparently has -Winconsistent-missing-override turned on
by default, and warned me about a bunch of overrides I missed.  Note
also that you can use the `override` keyword when virtual functions are
overridden to become pure virtual.

Refs #655.


roystgnr:
Thanks!  :+1:

PR #684, opened 2015-10-19
Fix getpot

jwpeterson:
Clang 3.7.0 warns:
```
./include/libmesh/getpot.h:2300:1: warning: all paths through this function will call itself [-Winfinite-recursion]
```
I think this is the right fix, but if @roystgnr and @pbauman can confirm that it also doesn't break their apps, that would be great.

BTW, this commit is based off bb34874 in #683, so it appears here as well but is not really part of this PR.  The fix is one line in c0fe8b0.


roystgnr:
I'm pretty sure nobody is *using* that code path in our apps; recursing until we run out of stack space would raise some eyebrows.

Fix looks correct, though.

PR #686, opened 2015-10-21
We should copy over node boundary data when we do BoundaryInfo::sync

dknez:


jwpeterson:
Seems reasonable.  Please squash these two commits together though.

dknez:
I squashed them.

roystgnr:
If we add this behavior we should document it in the comments.

Should we also hunt for edge boundary data to copy over?  Should edge data in 3D turn into side data in the 2D boundary?

dknez:
I'll add extra comments.

Your questions are certainly valid... I'm not sure what the best answer is, what do you think?

I needed the nodeset data on a BoundaryMesh for an application code I'm working on, which is what motivated this change.  Personally I don't need edge data on the BoundaryMesh right now, so I'd be happy to proceed with the current PR and add the edge data later if/when someone needs it.

roystgnr:
My whim would be "yes" and "yes", but if you want to leave that unimplemented (and documented as such) until someone needs it, that's fine.

dknez:
Comments added. But I think I will leave the edges unimplemented for now. I indicated this in the updated comments.

dknez:
Did this really fail??

permcody:
I highly doubt that's you. I'll look at it in the morning.

permcody:
Yeah, we occasionally see one or two of our threaded tests fail that way (Error 139) and we've never been able to repeat during debugging... I invalidated that target, it should pass just fine.

PR #688, opened 2015-10-22
Generalize VTK m4 test

jwpeterson:
@capitalaslash If you have access to a Linux distro that names their VTK libs in a non-standard way, would you mind trying out this branch and letting me know if it works?


floli:
I can confim that this PR configures and compiles on a recent Arch Linux (which used this non-standard naming scheme)

capitalaslash:
thank you for looking into this!
it works for me too.

jwpeterson:
Thanks both of you for taking a look.

PR #689, opened 2015-10-23
Assert no repeated calls to System::init()

roystgnr:
This isn't hard to do by accident (e.g. first manually calling
EquationSystems::init() and later calling EquationSystems::read() with
read_headers on), and even in dbg mode it may result in no saner error
than "suddenly PETSc LU is finding zero pivots on Poisson matrices".

Apologies to @vikramvgarg for the aforementioned insane error.

### DIFF
--- a/src/systems/system.C
+++ b/src/systems/system.C
@@ -227,6 +227,10 @@ void System::clear ()
 
 void System::init ()
 {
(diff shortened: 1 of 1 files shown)


roystgnr:
Those timeouts in dbg mode are annoying.  No assertion failures, though.

PR #690, opened 2015-10-27
Use full precision on *all* ASCII XDA output

roystgnr:
We've been using the default precision on critical paths for an
embarrassingly long time.

### DIFF
--- a/src/utils/xdr_cxx.C
+++ b/src/utils/xdr_cxx.C
@@ -728,6 +728,12 @@ void Xdr::data (T& a, const char* comment_in)
         libmesh_assert(out.get());
         libmesh_assert (out->good());
 
+        // We will use scientific notation with a precision of 16
+        // digits in the following output.  The desired precision and
+        // format will automatically determine the width.
+        *out << std::scientific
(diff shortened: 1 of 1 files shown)


roystgnr:
Pretty sure linker failures here aren't my fault.  ;-)  How could they have possibly triggered in opt but not debug?

roystgnr:
Triggered in both opt and debug.  At least that's *slightly* less insane.

jwpeterson:
Yeah, there's no way your PR caused this failure.  We'll look into it.

jwpeterson:
I invalidated the failed builds and they are rerunning on different machines... I think it's likely they will pass this time, and the failure was due to a specific build host being upgraded, but I'll keep an eye on it.

permcody:
However, you will still see the timeout failures on the two MOOSE tests
doing a lot of geometric search until that projection logic is cleaned up.

jwpeterson:
The only remaining failure seems to be the timeout in `geomsearch/3d_penetration_locator.test` in debug mode, so I'm fine with merging this any time...

PR #694, opened 2015-10-28
Mixed dimension refinement fixes/more testing

roystgnr:
My attempt at fixing the remaining bugs in #691 and #693.  This finally passes with fem_system_ex4 for me now, and it doesn't seem to break the tests @cahaynes added either.

PR #695, opened 2015-10-28
Add an operator bool() to AutoPtr.

jwpeterson:
I don't think this changes the inner workings of AutoPtr in any way,
and allows us to be more compatible with newer smart pointer types.

This fixes the compile error we encountered in #694, so presumably once this is merged we can invalidate that one and it will make it a bit further...

### DIFF
--- a/include/base/auto_ptr.h
+++ b/include/base/auto_ptr.h
@@ -376,7 +376,11 @@ public:
    * Forward compatibility with newer smart pointer types.  This
    * allows code like if (!foo) to work with AutoPtr.
    */
-  explicit operator bool() const
+#ifdef LIBMESH_HAVE_CXX11
+  // Conversion operators can only be marked explicit in C++11.
+  explicit
+#endif
+  operator bool() const
   { return (this->get() != NULL); }
 
   /**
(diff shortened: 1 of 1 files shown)


roystgnr:
:+1: and :100: only because I can't find a :+infinity: emoticon. Thanks for fixing this so quickly.

include/base/auto_ptr.h
```diff
@@ -373,6 +373,13 @@ class AutoPtr
   }
 
   /**
+   * Forward compatibility with newer smart pointer types.  This
+   * allows code like if (!foo) to work with AutoPtr.
+   */
+  explicit operator bool() const
+  { return (this->get() != NULL); }
```
roystgnr: You guys and your "!= NULL" pointer comparisons.  How about we return "(this->get() != NULL) == true" for good measure?  ;-)

jwpeterson:
This actually produces a warning in C++03 code, so I'll guard the use of explicit appropriately before merging...

include/base/auto_ptr.h
jwpeterson:
```
if (!this->get())
  that->get()
```

jwpeterson:
The only failures were the tests that time out, so I'm going to merge.

PR #696, opened 2015-10-28
Un-pessimize projections

roystgnr:
I can't yet call this "optimization", because on my system the projection in MOOSE 3d_penetration_locator is still taking ten times as long as it should.  But since #680 caused that projection to take two hundred times as long as it should, this is at least a good start at a fix.

### DIFF
--- a/src/systems/system_projection.C
+++ b/src/systems/system_projection.C
@@ -56,23 +56,42 @@ class GenericProjector
 private:
   const System &system;
 
-  FFunctor & f;
-  GFunctor * g;  // Needed for C1 type elements only
-  ProjectionAction & action;
+  // For TBB compatibility and thread safety we'll copy these in
+  // operator()
+  const FFunctor & master_f;
+  const GFunctor * master_g;  // Needed for C1 type elements only
+  bool g_was_copied;
+  const ProjectionAction & master_action;
   const std::vector<unsigned int>& variables;
 
 public:
   GenericProjector (const System &system_in,
-                    FFunctor & f_in,
-                    GFunctor * g_in,
-                    ProjectionAction & act_in,
+                    const FFunctor & f_in,
+                    const GFunctor * g_in,
+                    const ProjectionAction & act_in,
                     const std::vector<unsigned int>& variables_in) :
     system(system_in),
-    f(f_in),
-    g(g_in),
-    action(act_in),
+    master_f(f_in),
+    master_g(g_in),
+    g_was_copied(false),
+    master_action(act_in),
     variables(variables_in)
   {}
+
+  GenericProjector (const GenericProjector& in) :
+    system(in.system),
+    master_f(in.master_f),
+    master_g(in.master_g ? new GFunctor(*in.master_g) : NULL),
+    g_was_copied(in.master_g),
+    master_action(in.master_action),
+    variables(in.variables)
+  {}
+
+  ~GenericProjector()
+  {
+    if (g_was_copied)
+      delete master_g;
+  }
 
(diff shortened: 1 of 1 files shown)


jwpeterson:
Ah, the conflict is going to be in auto_ptr.h where I changed operator bool in the other PR.

roystgnr:
No problem.  I ought to eliminate those fixup commits anyway, and I can remove the copies of your old stuff and rebase on master while I'm at it.

roystgnr:
Well, damn it.  I can't actually reproduce this on my machine.  @permcody thinks it's a threading bug, though, and I was already suspicious about how a bit of this would behave when threaded, so I'll try fixing that and see if civet is happy afterwards.

roystgnr:
Not a threading bug.  The optimized code tries to copy dofs whenever it sees a node with an old_dof_object, but fails to synchronize dofs when that node only exists on other processors' old elements not on its own.  If a hanging node with fine elements on a remote processor gets turned into a non-hanging node, the newly refined elements then can't find the dofs they're looking for.

roystgnr:
Okay, not *just* a threading bug.  But I think the remaining bugs have been fixed - if the tests here still pass and all my tests pass then I'll merge this afternoon.

jwpeterson:
> if the tests here still pass and all my tests pass then I'll merge this afternoon.

Sounds good, the big history merge is in as well.

roystgnr:
Just utils/random.threads_verification_uo failing on opt.  I can't seem to reproduce it.

roystgnr:
And apparently civet can't reproduce it either, if changing a commit message gets us to pass.  That might just be one of the long standing threading issues @permcody was telling me about; I believe I've thoroughly scoured *this* PR for possible threading problems and fixed them, so even if I missed something elsewhere in the code, merging now shouldn't make things any worse.

All my other tests are passing, so merging now.

friedmud:
Don't worry about that test. It fails randomly (hah!).- I'm not sure why
@permcody hasn't disabled it yet.

If you see it fail... Just invalidate your tests and let them rerun..

roystgnr:
I've actually been begging @permcody for the opposite - it's obvious that we have a hard-to-trigger multithreading bug *somewhere* in MOOSE or libMesh, and IMHO the first thing we ought to do with it is try to figure out how to make it easier to trigger.

friedmud:
Nah - it's just the random number generator in that test. It's testing the
random number generator with threads... and it's never been reliable. There
is something wrong with it... and it does need to be fixed, but it's not a
systemic problem with threading in libMesh/MOOSE.

Note that we run all of our other tests very successfully and reliably with
threads. Only that one ever fails...

dschwen:
https://github.com/idaholab/moose/issues/5567 ?

permcody:
Actually Derek we see more than the random number test failing. Nearly all
of the threaded tests fail from time to time which means there may be a
bigger issue. Also Jason Hales reported quite awhile back that he has even
seen a slight difference in the result when running with or without threads
which is even worse! I've never been able to trigger these bugs when I try.
It may be about time to dig deeper.

friedmud:
You have to admit that the random number test fails an order of magnitude more often than any other threaded test.  There _is_ something specifically wrong with the random number generator and threading.

As for Hales... I wonder how much changes... their stuff depends so much on floating point stuff going one way or the other (in particular solid mechanical contact) that I wonder if even just different round-off is enough to lead down slightly different paths.

I'm not completely disregarding the idea that something could be up with threading... but no one has been able to come up with something reproducible...

PR #697, opened 2015-10-28
Fix to ignore "unused typedef"

permcody:
present in clang 3.7.0 when not using C++11 with METHOD=dbg.


jwpeterson:
:+1: I might try and figure out which ones are actually unused and delete them some day...

PR #698, opened 2015-10-28
Add the safe_bool helper class, use in AutoPtr

jwpeterson:
More information on why this is useful in C++03 code [here](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Safe_bool).  The branch also contains some simple unit tests verifying that it works.  I also plan to add a configure test to verify that `if (foo==bar)` *does not* compile when both `foo` and `bar` are `AutoPtrs`, however I have already checked this is indeed the case...


roystgnr:
:+1: after you tweak that comment; feel free to disagree with me on the warning-vs-error question.

PR #699, opened 2015-10-29
Move system libraries to end of link line in strip_dup_libs.pl

jwpeterson:
I tested this while configuring with `--enable-curl`, which lives in /usr/lib for me, and it seems to do the right thing, which is move `-L/usr/lib` to the end of the link line.


jwpeterson:
This passed the CI but obviously this was not one of the systems that needed it... we'll just have to keep an eye out for emails like [this one](https://sourceforge.net/p/libmesh/mailman/message/34573190/) to see if this actually fixes the problem and/or introduces new issues.

pbauman:
Thanks!

PR #701, opened 2015-11-02
Implement fixes suggested by cppclean

jwpeterson:
This is primarily changing several #includes to forward declarations as suggested by [cppclean](https://github.com/myint/cppclean).  This tool also produces quite a few false positives and crashes on some files, but I think the changes I've made in this PR are valid.  

I've already tested MOOSE, and it doesn't seem to have any issues, but several of the libmesh examples actually had to have #includes added in order to compile.  I've tested with most of the optional stuff enabled, major exceptions being complex number and GLPK support, but it would be great if some other app developers could try this branch as well and make sure it doesn't cause them too much heartburn.  I *don't* plan to merge this into the 0.9.5/1.0.0 branch, since it can potentially break previously-compiling code.

### DIFF
--- a/include/numerics/parsed_function_parameter.h
+++ b/include/numerics/parsed_function_parameter.h
@@ -24,11 +24,10 @@
 // Local Includes -----------------------------------
 #include "libmesh/libmesh_common.h"
 #include "libmesh/parameter_accessor.h"
+#include "libmesh/parsed_function.h"
 
 namespace libMesh
 {
-
-template <typename T> class ParsedFunction;
 
 /**
  * Accessor object allowing reading and modification of the
--- a/include/numerics/parsed_function_parameter.h
+++ b/include/numerics/parsed_function_parameter.h
@@ -24,10 +24,11 @@
 // Local Includes -----------------------------------
 #include "libmesh/libmesh_common.h"
 #include "libmesh/parameter_accessor.h"
-#include "libmesh/parsed_function.h"
 
 namespace libMesh
 {
+
+template <typename T> class ParsedFunction;
 
 /**
  * Accessor object allowing reading and modification of the
--- a/examples/eigenproblems/eigenproblems_ex3/eigenproblems_ex3.C
+++ b/examples/eigenproblems/eigenproblems_ex3/eigenproblems_ex3.C
@@ -64,6 +64,7 @@
 #include "libmesh/dof_map.h"
 #include "libmesh/fe_interface.h"
 #include "libmesh/getpot.h"
+#include "libmesh/elem.h"
 
 // Bring in everything from the libMesh namespace
 using namespace libMesh;
--- a/examples/fem_system/fem_system_ex1/naviersystem.C
+++ b/examples/fem_system/fem_system_ex1/naviersystem.C
@@ -28,6 +28,7 @@
 #include "libmesh/quadrature.h"
 #include "libmesh/string_to_enum.h"
 #include "libmesh/zero_function.h"
+#include "libmesh/elem.h"
 
 // Bring in everything from the libMesh namespace
 using namespace libMesh;
--- a/examples/fem_system/fem_system_ex2/solid_system.C
+++ b/examples/fem_system/fem_system_ex2/solid_system.C
@@ -34,6 +34,8 @@
 #include "libmesh/sparse_matrix.h"
 #include "libmesh/steady_solver.h"
 #include "libmesh/transient_system.h"
+#include "libmesh/node.h"
+#include "libmesh/elem.h"
 
 #include "nonlinear_neohooke_cc.h"
 #include "solid_system.h"
--- a/examples/fem_system/fem_system_ex3/fem_system_ex3.C
+++ b/examples/fem_system/fem_system_ex3/fem_system_ex3.C
@@ -43,6 +43,7 @@
 #include "libmesh/diff_solver.h"
 #include "libmesh/newmark_solver.h"
 #include "libmesh/steady_solver.h"
+#include "libmesh/elem.h"
 
 #define x_scaling 1.3
 
--- a/examples/fem_system/fem_system_ex4/heatsystem.C
+++ b/examples/fem_system/fem_system_ex4/heatsystem.C
@@ -10,6 +10,7 @@
 #include "libmesh/quadrature.h"
 #include "libmesh/string_to_enum.h"
 #include "libmesh/zero_function.h"
+#include "libmesh/elem.h"
 
 using namespace libMesh;
 
--- a/examples/miscellaneous/miscellaneous_ex7/biharmonic.h
+++ b/examples/miscellaneous/miscellaneous_ex7/biharmonic.h
@@ -4,6 +4,7 @@
 #include "libmesh/equation_systems.h"
 #include "libmesh/serial_mesh.h"
 #include "libmesh/exodusII_io.h"
+#include "libmesh/mesh_refinement.h"
 
 // Bring in bits from the libMesh namespace.
 // Just the bits we're using, since this is a header.
--- a/examples/miscellaneous/miscellaneous_ex7/biharmonic_jr.C
+++ b/examples/miscellaneous/miscellaneous_ex7/biharmonic_jr.C
@@ -9,6 +9,7 @@
 #include "libmesh/numeric_vector.h"
 #include "libmesh/periodic_boundaries.h"
 #include "libmesh/periodic_boundary.h"
+#include "libmesh/elem.h"
 
 // Example includes
 #include "biharmonic_jr.h"
--- a/examples/miscellaneous/miscellaneous_ex8/miscellaneous_ex8.C
+++ b/examples/miscellaneous/miscellaneous_ex8/miscellaneous_ex8.C
@@ -33,6 +33,7 @@
 #include "libmesh/numeric_vector.h"
 #include "libmesh/tecplot_io.h"
 #include "libmesh/threads.h"
+#include "libmesh/node.h"
 #include "meshless_interpolation_function.h"
 
 // C++ includes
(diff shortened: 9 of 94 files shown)


jwpeterson:
Update: added a couple compilation fixes for GLPK and complex-enabled builds.

jwpeterson:
Oops, didn't mean to push to this branch again so I cancelled the CI.  The extra commit was just the one fixing some unit tests in complex mode.

jwpeterson:
@pbaumann or @roystgnr have you had a chance to try out this branch with GRINS yet?  I'd like to get it merged since it touches quite a few files to try and prevent conflicts...  @dknez any comments on your apps?

pbauman:
Sorry, hadn't had a chance yet. Will do so now, thanks for the prod.

pbauman:
Hmm. I get the following just compiling libMesh:

<pre>
  CXX      src/base/libmesh_dbg_la-sparsity_pattern.lo
In file included from /fry1/data/shared/software/apps/gcc/4.8.4/include/c++/4.8.4/bits/stl_algobase.h:64:0,
                 from /fry1/data/shared/software/apps/gcc/4.8.4/include/c++/4.8.4/bits/char_traits.h:39,
                 from /fry1/data/shared/software/apps/gcc/4.8.4/include/c++/4.8.4/ios:40,
                 from /fry1/data/shared/software/apps/gcc/4.8.4/include/c++/4.8.4/istream:38,
                 from /fry1/data/shared/software/apps/gcc/4.8.4/include/c++/4.8.4/sstream:38,
                 from /fry1/data/shared/software/apps/gcc/4.8.4/include/c++/4.8.4/complex:45,
                 from ./include/libmesh/libmesh_common.h:46,
                 from ./include/libmesh/dof_map.h:24,
                 from ../../pbauman/src/base/sparsity_pattern.C:21:
/fry1/data/shared/software/apps/gcc/4.8.4/include/c++/4.8.4/bits/stl_pair.h: In instantiation of 'struct std::pair<std::__debug::map<const libMesh::Node*, double, std::less<const libMesh::Node*>, libMesh::Threads::scalable_allocator<std::pair<const libMesh::Node* const, double> > >, libMesh::Point>':
/fry1/data/shared/software/apps/gcc/4.8.4/include/c++/4.8.4/bits/stl_pair.h:102:11:   required from 'struct std::pair<const libMesh::Node* const, std::pair<std::__debug::map<const libMesh::Node*, double, std::less<const libMesh::Node*>, libMesh::Threads::scalable_allocator<std::pair<const libMesh::Node* const, double> > >, libMesh::Point> >'
/fry1/data/shared/software/apps/gcc/4.8.4/include/c++/4.8.4/debug/safe_iterator.h:181:2:   required from '__gnu_debug::_Safe_iterator<_Iterator, _Sequence>::_Safe_iterator(__gnu_debug::_Safe_iterator<_Iterator, _Sequence>&&) [with _Iterator = std::_Rb_tree_const_iterator<std::pair<const libMesh::Node* const, std::pair<std::__debug::map<const libMesh::Node*, double, std::less<const libMesh::Node*>, libMesh::Threads::scalable_allocator<std::pair<const libMesh::Node* const, double> > >, libMesh::Point> > >; _Sequence = std::__debug::map<const libMesh::Node*, std::pair<std::__debug::map<const libMesh::Node*, double, std::less<const libMesh::Node*>, libMesh::Threads::scalable_allocator<std::pair<const libMesh::Node* const, double> > >, libMesh::Point>, std::less<const libMesh::Node*>, libMesh::Threads::scalable_allocator<std::pair<const libMesh::Node* const, std::pair<std::__debug::map<const libMesh::Node*, double, std::less<const libMesh::Node*>, libMesh::Threads::scalable_allocator<std::pair<const libMesh::Node* const, double> > >, libMesh::Point> > > >]'
./include/libmesh/dof_map.h:692:36:   required from here
/fry1/data/shared/software/apps/gcc/4.8.4/include/c++/4.8.4/bits/stl_pair.h:102:11: error: 'std::pair<_T1, _T2>::second' has incomplete type
       _T2 second;                /// @c second is a copy of the second object
           ^
In file included from ./include/libmesh/mesh_base.h:33:0,
                 from ./include/libmesh/elem_range.h:24,
                 from ./include/libmesh/dof_map.h:32,
                 from ../../pbauman/src/base/sparsity_pattern.C:21:
./include/libmesh/point_locator_base.h:41:7: error: forward declaration of 'class libMesh::Point'
 class Point;
       ^
make[2]: *** [src/base/libmesh_dbg_la-sparsity_pattern.lo] Error 1
make[2]: Leaving directory `/fry1/data/users/pbauman/software/libs/libmesh/build/pr701'
make[1]: *** [../libmesh_dbg.la] Error 2
make[1]: Leaving directory `/fry1/data/users/pbauman/software/libs/libmesh/build/pr701/tests'
make: *** [install-recursive] Error 1
[05:53:26][pbauman@fry:/fry1/data/users/pbauman/software/libs/libmesh/build/pr701]$
</pre>

My environment:

<pre>
Currently Loaded Modules:
  1) emacs/24.4      3) git/2.3.0         5) paraview/4.3.1   7) texlive/2014      9) gcc/4.8.4    11) openblas/0.2.13    13) tbb/4.2    15) boost/1.57.0
  2) autotools/1.0   4) valgrind/3.10.1   6) cmake/3.1.3      8) doxygen/1.8.9.1  10) mpich/3.1.3  12) petsc/3.5.3-debug  14) vtk/6.1.0  16) hdf5/1.8.14
</pre>

jwpeterson:
Hmm... I didn't actually change point_locator_base.h in my branch but I guess some other file affected it?

I wonder if this has to do with the GLIBCXX_DEBUG stuff needing complete types to be defined.  I'll try on a linux box.

jwpeterson:
Compiles OK for me on Ubuntu 12.04 with GCC 4.6.3.

jwpeterson:
This also builds for me on Ubuntu 14.04.3, using GCC 4.8.4, so I'm not sure what the difference between our systems is.

Anyway, the problem seems to be triggered by line 692 of dof_map.h:

```
NodeConstraints::const_iterator node_constraint_rows_begin() const
{ return _node_constraints.begin(); }
```

where `_node_constraints` is a 

```
class NodeConstraints : public std::map<const Node *,
                                        std::pair<NodeConstraintRow,Point>,
                                        std::less<const Node *>,
                                        Threads::scalable_allocator<std::pair<const Node * const, std::pair<NodeConstraintRow,Point> > > >
```

So, in order to return an iterator into this, you have to be able to construct a `std::pair<X,Point>`, which requires `Point` to be a complete type...  can you try adding #include "libmesh/point.h" to dof_map.h and see if that fixes your build?

This is really complicated, and makes me wonder how any of this header file rejiggering is supposed to work...

pbauman:
OK, confirmed two things:

1. This is not just a `dbg` mode thing. Building only with `devel` and `opt` also triggers the failure.
2. Adding the `point.h` header to `dof_map.h` gets libMesh to build for me.

I'm on CentOS 6.7, but I have no idea why that should matter. I built the compiler from scratch, it uses its own STL library and C preprocessor, so that shouldn't depend on the system, just the compiler.

That is worrisome.

I'll go ahead with my local fix and at least test GRINS, but this is strange.

include/numerics/parsed_function_parameter.h
```diff
@@ -24,11 +24,12 @@
 // Local Includes -----------------------------------
 #include "libmesh/libmesh_common.h"
 #include "libmesh/parameter_accessor.h"
-#include "libmesh/parsed_function.h"
 
 namespace libMesh
 {
 
+template <typename T> class ParsedFunction;
```
pbauman: This forward declaration is incorrect. `ParsedFunction` has two template parameters (maybe cppclean missed it because we're always using the default second argument within this class?). I think we need to keep the header here, for now, and remove this forward declaration. @roystgnr, do we always want the default second template argument for `ParsedFunction` in `ParsedFunctionParameter`?

pbauman:
Modulo the comment on `parsed_function_parameter.h`, GRINS only needed a handful of `#include libmesh/elem.h` statements added and `make check` passed. To get to that point, I dropped the forward declaration in `parsed_function_parameter.h` (plus the previously discussed dof_map.h addition).

include/numerics/parsed_function_parameter.h
jwpeterson: > On Nov 4, 2015, at 5:56 PM, Paul T. Bauman <notifications@github.com> wrote:
> 
> In include/numerics/parsed_function_parameter.h:
> 
> >  
> >  namespace libMesh
> >  {
> >  
> > +template <typename T> class ParsedFunction;
> This forward declaration is incorrect. ParsedFunction has two template parameters (maybe cppclean missed it because we're always using the default second argument within this class?). I think we need to keep the header here, for now, and remove this forward declaration. @roystgnr, do we always want the default second template argument for ParsedFunction in ParsedFunctionParameter?
> 

Hmm, good catch, no idea why that compiled...

jwpeterson:
> On Nov 4, 2015, at 6:07 PM, Paul T. Bauman <notifications@github.com> wrote:
> 
> Modulo the comment on parsed_function_parameter.h, GRINS only needed a handful of #include libmesh/elem.h statements added and make check passed. To get to that point, I dropped the forward declaration in parsed_function_parameter.h (plus the previously discussed dof_map.h addition).
> 

Cool, thanks for taking a look.  I'll push your fixes to the branch before merge.

include/numerics/parsed_function_parameter.h
pbauman: Because it's not being used directly in the library. We use it in GRINS, which is why I caught it --- it didn't compile. :)

roystgnr:
map<K,T>::begin() shouldn't require a full definition of T, but including a map<K,T> member variable *should*.  I'm not sure why some compilers might be letting us get away with it, but they shouldn't; in the ENABLE_NODE_CONSTRAINTS case, dof_map.h should require point.h

PR #708, opened 2015-11-05
Updated BoundaryInfo::sync

dknez:
I wanted to be able to use the node and side id maps that are used internally when doing BoundaryInfo::sync(). This small PR gives us access to this data.

### DIFF
--- a/src/mesh/boundary_info.C
+++ b/src/mesh/boundary_info.C
@@ -140,16 +140,49 @@ void BoundaryInfo::sync (UnstructuredMesh& boundary_mesh,
   if (!_mesh.is_serial())
     this->comm().set_union(request_boundary_ids);
 
-  this->sync(request_boundary_ids, boundary_mesh,
-             boundary_mesh_data, this_mesh_data);
+  this->sync_helper(request_boundary_ids,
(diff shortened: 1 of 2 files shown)


src/mesh/boundary_info.C
```diff
@@ -259,6 +292,15 @@ void BoundaryInfo::sync (const std::set<boundary_id_type> &requested_boundary_id
   // and finally distribute element partitioning to the nodes
   Partitioner::set_node_processor_ids(boundary_mesh);
 
+  if(node_id_map_ptr)
+  {
+    *node_id_map_ptr = node_id_map;
+  }
```
roystgnr: Pretty inefficient to make a second copy right before destroying the first, no?  Do a swap() instead?

roystgnr:
:+1: once the efficiency issue is fixed.  Assuming I've correctly diagnosed it; on first glance I read the code wrong and I almost ranted at you about returning pointers to local variables, so it's possible I misunderstood something on second glance as well.

dknez:
Good point about using swap, I've fixed that now. I'll merge once the tests pass.

PR #709, opened 2015-11-05
Auto set interior parent/update mixed dim tests

cahaynes:
Added an algorithm to auto set interior parents.  Placed it in find_neighbors since it handles setting the interior parent of child elements when the parent has an interior parent.  Also, removed the manual setting of the interior parent in the mixed dim refinement tests since it is handled by find_neighbors.

### DIFF
--- a/src/mesh/mesh_base.C
+++ b/src/mesh/mesh_base.C
@@ -35,6 +35,8 @@
 #include "libmesh/point_locator_base.h"
 #include "libmesh/threads.h"
 
+#include LIBMESH_INCLUDE_UNORDERED_MAP
+
 namespace libMesh
 {
 
@@ -166,6 +168,14 @@ void MeshBase::prepare_for_use (const bool skip_renumber_nodes_and_elements, con
   if(!skip_find_neighbors)
     this->find_neighbors();
 
+  // Search the mesh for all the dimensions of the elements
+  // and cache them.
+  this->cache_elem_dims();
+
+  // Search the mesh for elements that have a neighboring element
+  // of dim+1 and set that element as the interior parent
+  this->detect_interior_parents();
+
   // Partition the mesh.
   this->partition();
 
@@ -179,10 +189,6 @@ void MeshBase::prepare_for_use (const bool skip_renumber_nodes_and_elements, con
 
   if(!_skip_renumber_nodes_and_elements)
     this->renumber_nodes_and_elements();
-
-  // Search the mesh for all the dimensions of the elements
-  // and cache them.
-  this->cache_elem_dims();
 
   // Reset our PointLocator.  This needs to happen any time the elements
   // in the underlying elements in the mesh have changed, so we do it here.
@@ -505,14 +511,116 @@ void MeshBase::cache_elem_dims()
   // particular dimension have been deleted.
   _elem_dims.clear();
 
-  const_element_iterator el  = this->active_local_elements_begin();
-  const_element_iterator end = this->active_local_elements_end();
+  const_element_iterator el  = this->active_elements_begin();
+  const_element_iterator end = this->active_elements_end();
 
   for (; el!=end; ++el)
     _elem_dims.insert((*el)->dim());
 
   // Some different dimension elements may only live on other processors
   this->comm().set_union(_elem_dims);
+}
+
+void MeshBase::detect_interior_parents()
+{
+  // This requires an inspection on every processor
+  parallel_object_only();
+
+  // Check if the mesh contains mixed dimensions. If so, then set interior parents, otherwise return.
+  if( this->elem_dimensions().size() == 1 )
+    return;
+
+  //This map will be used to set interior parents
+  LIBMESH_BEST_UNORDERED_MAP<dof_id_type, std::vector<dof_id_type> > node_to_elem;
+
+  const_element_iterator el  = this->active_elements_begin();
+  const_element_iterator end = this->active_elements_end();
+
+  for (; el!=end; ++el)
+    {
+      //Populating the node_to_elem map, same as MeshTools::build_nodes_to_elem_map
+      for (unsigned int n=0; n<(*el)->n_vertices(); n++)
+        {
+          libmesh_assert_less ((*el)->id(), this->max_elem_id());
+
+          node_to_elem[(*el)->node(n)].push_back((*el)->id());
+        }
+    }
+
+  // Automatically set interior parents
+  el = this->elements_begin();
+  for( ; el!=end; ++el )
+    {
+      Elem * element = *el;
+
+      //Ignore an 3D element or an element that already has an interior parent
+      if( element->dim()>=LIBMESH_DIM || element->interior_parent() )
+        continue;
+
+      // Start by generating a SET of elements that are dim+1 to the current
+      // element at each vertex of the current element, thus ignoring interior nodes.
+      // If one of the SET of elements is empty, then we will not have an interior parent
+      // since an interior parent must be connected to all vertices of the current element
+      std::vector< std::set<dof_id_type> > neighbors( element->n_vertices() );
+
+      bool found_interior_parents = false;
+
+      for( dof_id_type n=0; n < element->n_vertices(); n++ )
+        {
+          std::vector<dof_id_type> &element_ids = node_to_elem[element->node(n)];
+          for( std::vector<dof_id_type>::iterator e_it = element_ids.begin();
+               e_it != element_ids.end(); e_it++)
+            {
+              dof_id_type eid = *e_it;
+              if( this->elem(eid)->dim() == element->dim()+1 )
+                neighbors[n].insert(eid);
+            }
+          if( neighbors[n].size()>0 )
+            {
+              found_interior_parents = true;
+            }
+          else
+            {
+              // We have found an empty set, no reason to continue
+              // Ensure we set this flag to false before the break since it could have
+              // been set to true for previous vertex
+              found_interior_parents = false;
+              break;
+            }
+        }
+
+      // If we have successfully generated a set of elements for each vertex, we will compare
+      // the set for vertex 0 will the sets for the vertices until we find a id that exists in
+      // all sets.  If found, this is our an interior parent id.  The interior parent id found
+      // will be the lowest element id if there is potential for multiple interior parents.
+      if(found_interior_parents)
+        {
+          std::set<dof_id_type> &neighbors_0 = neighbors[0];
+          for(std::set<dof_id_type>::iterator e_it = neighbors_0.begin();
+              e_it != neighbors_0.end(); e_it++)
+            {
+              found_interior_parents=false;
+              dof_id_type interior_parent_id = *e_it;
+              for(dof_id_type n=1; n < element->n_vertices(); n++)
+                {
+                  if(neighbors[n].find(interior_parent_id)!=neighbors[n].end())
+                    {
+                      found_interior_parents=true;
+                    }
+                  else
+                    {
+                      found_interior_parents=false;
+                      break;
+                    }
+                }
+              if(found_interior_parents)
+              {
+                element->set_interior_parent( this->elem(interior_parent_id) );
+                break;
+              }
+            }
+        }
+    }
 }
 
 } // namespace libMesh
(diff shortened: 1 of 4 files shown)


pbauman:
Here's the failure:

<pre>
mesh/nemesis.nemesis_repartitioning_test: Assertion `(*el)->id() < this->n_elem()' failed.
mesh/nemesis.nemesis_repartitioning_test: (*el)->id() = 4671
mesh/nemesis.nemesis_repartitioning_test: this->n_elem() = 4454
mesh/nemesis.nemesis_repartitioning_test: 
mesh/nemesis.nemesis_repartitioning_test: 
mesh/nemesis.nemesis_repartitioning_test: [0] ../src/mesh/unstructured_mesh.C, line 275, compiled Nov  4 2015 at 23:56:01
mesh/nemesis.nemesis_repartitioning_test: application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0
</pre>

Looks like this is happening during a Nemesis output? I'm guessing based on the test name. @jwpeterson @permcody, anyway we can get a stacktrace? Is this a `SerialMesh` or `ParallelMesh` test case?

jwpeterson:
> Looks like this is happening during a Nemesis output? I'm guessing based on the test name. @jwpeterson @permcody, anyway we can get a stacktrace? Is this a SerialMesh or ParallelMesh test case?

This test runs in parallel on 4 (and only 4) procs and uses `ParallelMesh`.  Our testing system does not write trace files (in parallel, you have to configure with --enable-trace-files to see them), but I don't think the stack trace will be very long: `find_neighbors` is probably called after the first repartitioning step happens.  I can pull this branch and try it myself, but it will be later today since I'm working on some other PRs at the moment.

jwpeterson:
Sorry, that configure flag is actually `--enable-tracefiles`.

permcody:
This is probably a good idea for a test systems anyway. Perhaps we should
always enable trace-files.

roystgnr:
I can replicate that failure in libMesh alone, running adaptivity_ex3 on 2 or more processors after configuring with --enable-parmesh

roystgnr:
Wait, WTF?  Make that 1 or more processors.

pbauman:
OK, thanks. I'm building this branch now and will take a look, too.

roystgnr:
Here you go:

```            libmesh_assert_less ((*el)->id(), this->n_elem());

            node_to_elem[(*el)->node(n)].push_back((*el)->id());
```

With ParallelMesh, we *aren't* guaranteed to have contiguously numbered elements.  If you need an array that can hold all element ids, then you don't want to use n_elem(), you want to use max_elem_id() IIRC.

pbauman:
Huh. Just to check my understanding, `n_elem()` is still the total number of elements in the mesh, but there's no guarantee in the element numbering that `max_elem_id()` <= `n_elem()`?

pbauman:
The following patch gets adaptivity_ex3 to work for me with `ParallelMesh`

<pre>
diff --git a/src/mesh/unstructured_mesh.C b/src/mesh/unstructured_mesh.C
index a69af59..ace1f10 100644
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -232,8 +232,7 @@ void UnstructuredMesh::find_neighbors (const bool reset_remote_elements,
   START_LOG("find_neighbors()", "Mesh");
 
   //This map will be used to set interior parents
-  std::vector< std::vector<dof_id_type> > node_to_elem;
-  node_to_elem.resize(this->n_nodes());
+  std::map< dof_id_type, std::vector<dof_id_type> > node_to_elem;
 
   const element_iterator el_end = this->elements_end();
 
@@ -271,8 +270,7 @@ void UnstructuredMesh::find_neighbors (const bool reset_remote_elements,
         //Populating the node_to_elem map, same as MeshTools::build_nodes_to_elem_map
         for (unsigned int n=0; n<(*el)->n_vertices(); n++)
           {
-            libmesh_assert_less ((*el)->node(n), node_to_elem.size());
-            libmesh_assert_less ((*el)->id(), this->n_elem());
+            libmesh_assert_less ((*el)->id(), this->max_elem_id());
 
             node_to_elem[(*el)->node(n)].push_back((*el)->id());
           }
</pre>

pbauman:
Should we try to provide the user a way to opt out of doing the check for interior parents?

pbauman:
Actually, we ought to be able to just check `MeshBase::elem_dimensions()` and if that only has 1 element, then we shouldn't need to check for interior neighbors.

roystgnr:
That patch ought to be noticeably less efficient on large SerialMesh, but since it scales on ParallelMesh where a simple vector(max_node_id) wouldn't, I think I like it better anyway.  Best would be LIBMESH_BEST_UNORDERED_MAP, which in cases where you don't need the ordering ought to be a little faster than plain map on small meshes and asymptotically O(log(N)) faster on large meshes.

I'm pretty worried by this PR now, though.  If nobody's been thinking about ParallelMesh, is this going to handle the local/ghosted/remote element distinction okay?  I swear about 20% of the bugs I ran into when creating ParallelMesh ended up boiling down to corner cases in find_neighbors()/remote_elem interaction.

pbauman:
> Best would be LIBMESH_BEST_UNORDERED_MAP,

Derp. Of course, thanks.

> I'm pretty worried by this PR now, though. If nobody's been thinking about ParallelMesh, is this going to handle the local/ghosted/remote element distinction okay? I swear about 20% of the bugs I ran into when creating ParallelMesh ended up boiling down to corner cases in find_neighbors()/remote_elem interaction.

Totally fair. @cahaynes, get everything addressed here and run the GRINS example we'd been playing with through some adaptive refinements with `mesh_type = parallel` and ramp up the processor count to try and this.

pbauman:
Also, what's missing here is the adding-interior-parent-of-interior-parent to the neighbor list, which I'd asked @cahaynes to push onto the this after some scrutiny of this first part.

pbauman:
@cahaynes, meant to say too, be sure to add checks for `mesh.elem_dimensions().size() > 1` around the bits you've added so we don't build the node_to_elem_map and don't do the interior parent checking/setting when we don't have to. Thanks!

cahaynes:
I made the changes requested by @jwpeterson. To handle ParallelMesh more appropriately (using the patch by @pbauman and suggestions from @roystgnr) I used LIBMESH_BEST_UNORDERED_MAP and assert when an element id is greater than the maximum element id rather than the number of elements.  

Also, I skip setting the interior parents when the mesh has only one dimension, as suggested by @pbauman.  In order to use `mesh.elem_dimensions().size()` as the check, I moved the code to a separate method in mesh_base.C.  This way `mesh.cache_elem_dimensions()` can be called prior to setting the interior parents and thus `mesh.elem_dimensions()` will be available.

Also, the methods were moved prior to the partitioning in prepare_for_use in order to avoid the complexities of one processor containing the interior parent of an element on another processor.  Possibly a lack of foresight helped this decision, but I felt this was the simplest solution.

One side effect of moving to a separate method is that I introduced an additional iteration over active elements to to populate the node_to_elem map.  The hope is that one less iteration over elements for meshes that don't need interior parents outweigh the extra iteration for meshes that do need interior parents.

src/mesh/mesh_base.C
```diff
@@ -515,4 +521,105 @@ void MeshBase::cache_elem_dims()
   this->comm().set_union(_elem_dims);
 }
 
+void MeshBase::detect_interior_parents()
+{
+  // This requires an inspection on every processor
+  parallel_object_only();
+
+  if( this->elem_dimensions().size() == 1 )
```
pbauman: Let's add a comment here indicating why we do this, i.e. we only need to do this check if the mesh has elements of multiple different dimensions.

pbauman:
BTW, @cahaynes has hit this with `ParallelMesh` and adaptivity on a GRINS example with increasing mesh size and processor count, so there is some warm and fuzzy feeling about this. But only 1-D/2-D manifolds, not 2-D/3-D or 1-D/3-D.

src/mesh/mesh_base.C
cahaynes: Okay, should I also squash this new commit?

pbauman:
Yeah, go ahead and squash into 028d240 (the last commit currently there).

roystgnr:
Do we want to skip the dim=1 case?  AFAIK it is currently possible to create dim=0 NodeElem boundaries on 1-D meshes, and if it isn't then that's a bug we should fix, not exacerbate.

I admit I can't imagine someone wanting to write a code which mixes 1-D elements and 0-D elements, but I *can* imagine someone first wanting to write a code which mixes N-D and (N-1)-D elements and then wanting to test that code in the N=1 case.

src/mesh/mesh_base.C
```diff
@@ -505,8 +511,8 @@ void MeshBase::cache_elem_dims()
   // particular dimension have been deleted.
   _elem_dims.clear();
 
-  const_element_iterator el  = this->active_local_elements_begin();
-  const_element_iterator end = this->active_local_elements_end();
+  const_element_iterator el  = this->active_elements_begin();
+  const_element_iterator end = this->active_elements_end();
```
roystgnr: We really ought to add an iterator which traverses both active and unpartitioned elements...

roystgnr:
:+1: to go into master once we finish arguing about the 1-D case.  I can't decide whether we should fork off 0.9.5-rc1 before or after the merge, though.  Major edits to core functionality right before release time scare me a little, but this looks like a well-encapsulated feature addition which shouldn't break interior_parent apps and definitely shouldn't break any other apps.

pbauman:
> Do we want to skip the dim=1 case?

@roystgnr, I'm not sure I'm understanding you here. Are you talking about [line 529 of mesh_base.C](https://github.com/cahaynes/libmesh/blob/auto_set_interior_parent/src/mesh/mesh_base.C#L529)? If so, what we're checking is if there is actually a mixed dimension mesh; if there's not, then we don't need to do all that work (which is why we sacrificed putting an extra element loop for building the node-to-elem map, so we could do that check). And, if that is what you're talking about, it makes @cahaynes adding comments there all the more prudent. So, we should be able to detect meshes with dim=0 and dim=1 (i.e. elem_dimensions() should return 2). But maybe I misunderstood you.

If that's not what you're talking about, could you point us to a line?

roystgnr:
My fault - I misread "if the mesh has only one dimension" as "if the mesh is one dimensional".

pbauman:
Great. @cahaynes let's get that comment pushed here then we can merge this.

@roystgnr @jwpeterson IMHO, I think we can plug this in 0.9.5 since this new functionality only gets triggered for mixed-dimensional meshes, so the only potential breakage is there. But I defer to y'alls judgement.

jwpeterson:
> On Nov 25, 2015, at 8:41 PM, Paul T. Bauman <notifications@github.com> wrote:
> 
> Great. @cahaynes let's get that comment pushed here then we can merge this.
> 
> @roystgnr @jwpeterson IMHO, I think we can plug this in 0.9.5 since this new functionality only gets triggered for mixed-dimensional meshes, so the only potential breakage is there. But I defer to y'alls judgement.
> 

I'd prefer not, 0.9.5 is for bugfixes only, however I'll let @roystgnr be the tiebreaker.

roystgnr:
"0.9.5 is for bugfixes only" is cryptic.  Have we started a 0.9.5 branch yet?

jwpeterson:
> On Nov 25, 2015, at 9:26 PM, roystgnr <notifications@github.com> wrote:
> 
> "0.9.5 is for bugfixes only" is cryptic.  Have we started a 0.9.5 branch yet?
> 

Yea, it's been going for several weeks now.  You should 'git fetch upstream' some time...

roystgnr:
If I was only several weeks behind on my todo list I'd be a happy person.  Okay, let's merge this to master and keep it out of 0.9.5.

jwpeterson:
I'd like to take one last look at this and get it merged today.

pbauman:
Great. Looks like @cahaynes pushed (and squashed) the comment update, so let us know if there's anything else. Otherwise, this looks kosher to me.

jwpeterson:
Merged in 778c90e.  I noticed that you had originally added code directly to `find_neighbors()` and then moved it.  I just squashed all those commits into one.

PR #707, opened 2015-11-04
Prevent duplicate entries in BoundaryInfo

jwpeterson:
This fixes the bug reported in #706, plus refactors a few things for readability.

PR #711, opened 2015-11-06
Update and deprecate several variants of BoundaryInfo::boundary_ids()

jwpeterson:
This is a continuation of the work in #707 where the BoundaryInfo::add_{side,node,edge} interfaces that took std::vectors were upgraded to take std::sets instead.

I believe the new versions of boundary_ids() are potentially more efficient, especially when used in loops, because they don't create brand new containers to return to the user, they just fill up a passed-in container with IDs.

I've also converted library and examples code over to using these new interfaces.  These last patches could all be squashed down together if someone wants, but I opted to keep them separate for bisecting, just in case a bug was introduced during the refactoring.

### DIFF
--- a/examples/reduced_basis/reduced_basis_ex6/assembly.h
+++ b/examples/reduced_basis/reduced_basis_ex6/assembly.h
@@ -94,9 +94,9 @@ struct AssemblyA0 : ElemAssemblyWithConstruction
 {
   virtual void boundary_assembly(FEMContext &c)
   {
-    const std::vector<boundary_id_type> bc_ids =
-      rb_con->get_mesh().get_boundary_info().boundary_ids (&c.get_elem(),c.side);
-    for (std::vector<boundary_id_type>::const_iterator b =
+    std::set<boundary_id_type> bc_ids;
+    rb_con->get_mesh().get_boundary_info().boundary_ids (&c.get_elem(), c.side, bc_ids);
+    for (std::set<boundary_id_type>::const_iterator b =
            bc_ids.begin(); b != bc_ids.end(); ++b)
       if( *b == 1 || *b == 2 || *b == 3 || *b == 4 )
         {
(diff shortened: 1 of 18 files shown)


friedmud:
Hmmm... filling a set is actually pretty expensive (each addition has to be checked against all others).  Did you do any testing on the timing of this?  (Maybe setup a situation where you call it a million times in a tight loop and compare the new and old versions?)  It gets called a LOT and it would be good to know...

friedmud:
BTW: the test should contain quite a few boundary ids .  Maybe see how many a typical BISON solve used (on the order of 30-40 I think, but possibly hundreds).

jwpeterson:
I'll try it, but note that the original code paths *allocated and deallocated* an entire std::vector at each iteration of said loop while the new code path just clears/reuses existing memory.  I'll be very disappointed if reusing std::set is slower than that, but I'll do some timings to make sure...

jwpeterson:
I created a little test case [here](https://gist.github.com/jwpeterson/ef68cf5c4bee0e9b2096) that measures how long it takes to call `boundary_ids()` on a mesh of 1000 HEX8's where every side of every element has  `n` boundary ids, both before and after the recent refactoring.  The old method which created a vector at every invocation was faster across the entire range of different values of `n`:

| n | set | vector |
|--------|--------|-----|
| 10  | .0110826  | .0050758 |
| 20  | .0220908  | .0080900 |
| 40  | .0406006  | .0125184 |
| 80  | .0837440  | .0241320 |
| 160 | .1733694  | .0444768 |

I guess I forgot the good old rule of thumb that "std::vector beats all other containers"...

I think I'm going to reverse course on the entire refactoring, but not quite a full revert since I think there were also some good changes there.  I'll have `BoundaryInfo::add_side()` simply reject duplicate entries in its input vector without throwing an error, and we should be good...

friedmud:
Dang.  I thought that might be the case.  Thanks for checking into it!  Did you happen to run any cases below 10?  I'd be interested to see if set is ever faster...

Also: did you try anything other than std::set?  An `unordered_set` would work just as well here and may be faster...

friedmud:
Oh - can we get the best of both worlds by passing in a `std::vector` to fill?  That we don't have the reallocation issue... and we have fast inserting?

jwpeterson:
Yep, that's the plan for reverting the std::set version.

PR #712, opened 2015-11-06
NewmarkSolver::set_initial_accel_avail

pbauman:
This adds the function `NewmarkSolver::set_initial_accel_avail` allowing the user to indicate the acceleration vector is already loaded and thus, don't need to either compute it or project onto it. The intended use for this function is for the case when the user is restarting, having used  `libMesh::EquationSystems::WRITE_ADDITIONAL_DATA` so that the acceleration vector (among others) was written to the restart file, and shouldn't recompute the initial acceleration.

I'm not really thrilled with this solution, but I can't come up with a different way of doing this other than checking the norm of the acceleration vector. But then we have to introduce a tolerance, and then a knob for that tolerance. I feel like the user controls when they're restarting so they should be able to discern when to compute the initial acceleration vs. knowing it was read from a restart. But I'm all ears on this.

### DIFF
--- a/include/solvers/newmark_solver.h
+++ b/include/solvers/newmark_solver.h
@@ -93,6 +93,14 @@ public:
   void project_initial_accel( FunctionBase<Number> *f, FunctionBase<Gradient> *g = NULL );
 
   /**
+   * Allow the user to (re)set whether the initial acceleration is available.
+   * This is not needed if either compute_initial_accel() or project_initial_accel()
+   * are called. This is useful is the user is restarting a calculation and the acceleration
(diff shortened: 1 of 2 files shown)


jwpeterson:
:+1:

roystgnr:
Even if you used a tolerance of 0 (recompute for |a|=0, don't otherwise) then you'd be wasteful for the (probably not uncommon) a(0)=0 case.  This flag looks a bit more annoying but much safer than trying to autodetect.  :+1:

pbauman:
Thanks for the comments.

Issue #716, opened 2015-11-10
automake refresh


benkirk:
I'm working a automake-1.15 refresh.

For now, I'm specifying the 'serial-tests' to get the old 'make check' behavior, namely print the example output to stdout, e.g.

```pre
============================================================================
Testsuite summary for libmesh 1.0.0-pre
============================================================================
# TOTAL: 1
# PASS:  1
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
Making check in introduction/introduction_ex3
  CXX      example_dbg-introduction_ex3.o
  CXX      example_dbg-exact_solution.o
  CXX      example_devel-introduction_ex3.o
  CXX      example_devel-exact_solution.o
  CXX      example_opt-introduction_ex3.o
  CXX      example_opt-exact_solution.o
  CXXLD    example-opt
  CXXLD    example-dbg
  CXXLD    example-devel
PASS: run.sh
============================================================================
Testsuite summary for libmesh 1.0.0-pre
============================================================================
# TOTAL: 1
# PASS:  1
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
Making check in introduction/introduction_ex4
  GEN      exact_solution.C
  CXX      example_dbg-introduction_ex4.o
  CXX      example_dbg-exact_solution.o
  CXX      example_devel-introduction_ex4.o
  CXX      example_devel-exact_solution.o
  CXX      example_opt-introduction_ex4.o
  CXX      example_opt-exact_solution.o
  CXXLD    example-opt
  CXXLD    example-dbg
  CXXLD    example-devel
PASS: run.sh
============================================================================
Testsuite summary for libmesh 1.0.0-pre
============================================================================
# TOTAL: 1
# PASS:  1
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
```

The issue is using the default parallel-tests, the output of each example is captured to a log file.  and not displayed.  

This makes perfect sense when make check is a bunch of tiny codes you'd like to run in parallel, but in our case when each check target itself can be a parallel application can be problematic.

A couple options I see could be:

1. Stick with the 'serial-tests' as long at it is supported,
2. Add a custom 'make verbose-check' target or something that basically runs 'make check' with the parallel suite and then cats each log file, along the lines of

[...]

PR #717, opened 2015-11-10
Automake1.15  refresh

benkirk:
Implements automake refresh #716


benkirk:
Hmm, the moose PR test doesn't look like it runs bootsrtrap, which I expect would fix this.

I'll hold off on the merge for a bit.

jwpeterson:
Hmmm... does this mean we need to check in more temporary build files in the build-aux directory or something?

jwpeterson:
The error message, for reference:
```
/home/moosetest/civet/client_root_1/libmesh/build-aux/missing: line 81: aclocal-1.15: command not found
WARNING: 'aclocal-1.15' is missing on your system.
         You should only need it if you modified 'acinclude.m4' or
         'configure.ac' or m4 files included by 'configure.ac'.
         The 'aclocal' program is part of the GNU Automake package:
         <http://www.gnu.org/software/automake>
         It also requires GNU Autoconf, GNU m4 and Perl in order to run:
         <http://www.gnu.org/software/autoconf>
         <http://www.gnu.org/software/m4/>
         <http://www.perl.org/>
make: *** [../aclocal.m4] Error 127
```

benkirk:
It means that 'make' triggered a rebuild of the autotools geenrated files, why I'm not sure (date stamps perhaps...).  But it wanted to be very pick about the version of  aclocal and mandated the 1.15 version.

./bootstrap should look for automake-1.15 on the system and install it in ./contrib/autotools if it's not there.  But I don't think that gets run as part of the PR recipe?

That said, we usually don't need to run bootstrap when pulling from git since we check in the built sources, so I'll try and see why this is getting triggered at all.

jwpeterson:
> ./bootstrap should look for automake-1.15 on the system and install it in ./contrib/autotools if it's not there. But I don't think that gets run as part of the PR recipe?

Right, I still think we want to be able to build libmesh without running bootstrap, and this is what the CI tests.

> That said, we usually don't need to run bootstrap when pulling from git since we check in the built sources, so I'll try and see why this is getting triggered at all.

OK, sounds good.

benkirk:
It may be bootstrap is required once when the automake version gets updated, but not on every ./configure. Would that break you?

jwpeterson:
> It may be bootstrap is required once when the automake version gets updated, but not on every ./configure. Would that break you?

It's not ideal.  It would break every MOOSE user that relies on our scripts to update their copy of libmesh inside of MOOSE.  These are users who have no idea what configure even is...  I'd rather not update automake if it's not possible to check in enough build files for "configure; make; make install" to just work.

benkirk:
I think disabling maintainer-mode is the fix, and will try that tomorrow...  

http://www.gnu.org/software/automake/manual/html_node/maintainer_002dmode.html

benkirk:
For reference on this thread:

https://autotools.io/automake/maintainer.html

I'm thinking a trivial change in e.g. configure.ac and a resubmit might fix it. As it stands the new missing and configure should have the same time stamp, I think.

benkirk:
On Nov 10, 2015, at 4:28 PM, John W. Peterson <notifications@github.com<mailto:notifications@github.com>> wrote:


It's not ideal. It would break every MOOSE user that relies on our scripts to update their copy of libmesh inside of MOOSE. These are users who have no idea what configure even is... I'd rather not update automake if it's not possible to check in enough build files for "configure; make; make install" to just work.

how about ./configure … --disable-maintainer-mode on your end?

That disables the logic which rebuilds e.g Makefile.in from Makefile.am, or configure from configure.ac.  I wouldn’t want that to be the default behavior when you pull from git and work on development, but would be fine for your users I’d think...

jwpeterson:
> how about ./configure … --disable-maintainer-mode on your end?
> 
> That disables the logic which rebuilds e.g Makefile.in from Makefile.am, or configure from configure.ac.  I wouldn’t want that to be the default behavior when you pull from git and work on development, but would be fine for your users I’d think...

I think I'd rather make `--enable-maintainer-mode` an opt-in for developers if that's possible. I still think the majority of people that download libmesh are just using it and not developing it and would like to avoid the extra hoop of requiring bootstrap or `./configure --disable-maintainer-mode` just for simplicity's sake...

friedmud:
I agree with @jwpeterson

There are exactly 6 people in this world that want to use maintainer mode
and everyone else just wants it to work ;-)

benkirk:
> On Nov 11, 2015, at 10:46 AM, Derek Gaston <notifications@github.com> wrote:
> 
> I agree with @jwpeterson
> 
> There are exactly 6 people in this world that want to use maintainer mode
> and everyone else just wants it to work ;-)

Ha!  But not quite - this is only an issue sporadically (once in three years at this point) with putting straight from git; and never from a release tarball. That's the whole point of 'make dist.'

Still, given there's no rush to 1.15, I don't care that much. If/when 2.0 arrives we can recycle this tired old debate! ;-)

jwpeterson:
So I added `--disable-maintainer-mode` to our CI tests for both MOOSE and libMesh and invalidated/restarted this PR.  It looks like its going to work...(?)  This should actually be fine as far as MOOSE users are concerned: they are trained to use the `update_and_rebuild_libmesh.sh` script and will just continue to do so.

friedmud:
Did you update all of the instructions on mooseframework.org for this? I
still think TD a bad default... and, yes, I'm a little grumpy that my own
personal build scripts have have yet another option added to them...

(Cue flashback)

I remember a time in the not so distant past (like only 3 years ago) when
all you needed to do to build libMesh was "configure; make". Even all of
our users were able to do it without problems....

jwpeterson:
> Did you update all of the instructions on mooseframework.org for this? I

I assume you're referring to [this](http://mooseframework.org/getting-started)? There's nothing to update, right, you just run `update_and_rebuild_libmesh.sh`.

> still think it's a bad default... and, yes, I'm a little grumpy that my own
> personal build scripts have have yet another option added to them...

I don't think you necessarily need to update your personal build scripts, you just have to run bootstrap once.

PR #719, opened 2015-11-11
Hilbert + unique_id sorting

roystgnr:
If this works, it should enable slit meshes, overset meshes, and all the other fun meshes where nodes or centroids share the same location and therefore the same Hilbert indices.

This is starting to pass my own tests but is not ready to merge; even if it actually passes MooseBuild already I've still got more testing lined up.

### DIFF
--- a/examples/fem_system/fem_system_ex4/fem_system_ex4.C
+++ b/examples/fem_system/fem_system_ex4/fem_system_ex4.C
@@ -30,6 +30,7 @@
 #include <iomanip>
 
 // Basic include files
+#include "libmesh/elem.h"
 #include "libmesh/equation_systems.h"
 #include "libmesh/error_vector.h"
 #include "libmesh/exact_solution.h"
@@ -131,6 +132,21 @@ int main (int argc, char** argv)
     bcids.insert(bcid);
     mesh.get_boundary_info().add_elements(bcids, mesh);
     mesh.prepare_for_use();
+  }
+
+  // To work around ExodusII file format limitations, we need elements
+  // of different dimensionality to belong to different subdomains.
+  // Our interior elements defaulted to subdomain id 0, so we'll set
+  // boundary elements to subdomain 1.
+  {
+    const MeshBase::element_iterator end_el = mesh.elements_end();
+    for (MeshBase::element_iterator el = mesh.elements_begin();
+         el != end_el; ++el)
+      {
+        Elem * elem = *el;
+        if (elem->dim() < dim)
+          elem->subdomain_id() = 1;
+      }
   }
 
   mesh_refinement.uniformly_refine(coarserefinements);
@@ -274,9 +290,8 @@ int main (int argc, char** argv)
 #ifdef LIBMESH_HAVE_EXODUS_API
   // We want to write the file in the ExodusII format, but we don't
   // yet support mixed-dimension meshes there?
-/*  ExodusII_IO(mesh).write_equation_systems
+  ExodusII_IO(mesh).write_equation_systems
     ("out.e", equation_systems);
-*/
 #endif // #ifdef LIBMESH_HAVE_EXODUS_API
 
 #ifdef LIBMESH_HAVE_GMV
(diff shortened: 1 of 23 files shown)


dknez:
Very cool. I can test this on some slit meshes soon, when I get some time. Maybe I'll wait until you've finished your first round of testing though, if that's OK.

roystgnr:
Not just OK, but a very good strategy, in hindsight.  ;-)

I'm pretty baffled by those results, though.  The MOOSE Nemesis tests are the only things failing in debug mode, but there's dozens of crashes in opt mode?

I'm planning to get a slit mesh unit test (and maybe an overset mesh test) in there, but running on big complicated slit meshes would be better.

dknez:
OK, no problem.

I can send you a 3D slit mesh now to test on, if you like. Otherwise, I can do that test later, once these moose tests are green.

roystgnr:
We need to tell MooseBuild to run the dbg mode tests in parallel too.  Took me forever to realize that the reason I couldn't replicate failures was nothing more than the fact that I was neglecting to run with "-p 2".

jwpeterson:
> We need to tell MooseBuild to run the dbg mode tests in parallel too.

Hmm... yeah I don't see why not.  I'll open an issue about that.

roystgnr:
Okay, this passes all the libMesh, MOOSE, and GRINS tests I can throw at it.  And it's a good patch - it brings unique_id() support to a hundred corner cases instead of "maybe it'll work if you hold it really steady and don't breathe too hard".

But the *sorting* isn't actually useful yet, because we don't write out nodal unique_id and so we can't use it to disambiguate colocated nodes.  I'll start a separate branch and PR for that bit, and I won't merge this until that's ready to go, in case anyone wants to take time to review it first.

permcody:
Well thanks for working on this.

examples/fem_system/fem_system_ex4/fem_system_ex4.C
```diff
@@ -274,9 +290,8 @@ int main (int argc, char** argv)
 #ifdef LIBMESH_HAVE_EXODUS_API
   // We want to write the file in the ExodusII format, but we don't
```
jwpeterson: Looks like you can delete the comment on lines 291-292 now.

jwpeterson:
I'm pretty sure all the moosebuild targets configure with `--enable-unique-id`.  I don't suppose you tested this branch with `--disable-unique-id`.

include/mesh/mesh_communication.h
```diff
@@ -171,13 +171,20 @@ class MeshCommunication
    */
   void make_elems_parallel_consistent (MeshBase &);
 
-  /**
+   /**
```
jwpeterson: Looks like an extra space got inserted here.

roystgnr:
What, do you think I'm an idiot?

(/me sheepishly goes to test with --disable-unique-id)

jwpeterson:
Looks good, did you want to add any unit tests for the new functionality?  I guess they would need to run in parallel to really test it...

roystgnr:
Not sure unit tests for these specific methods would help much; all the bugs which cropped up along the way felt more like interactions between the way different methods exptected things to work.

What we ought to have, though, is unit tests for all the mesh I/O and MeshGeneration options...

roystgnr:
We ought to merge this soon (assuming I didn't screw up any of the conflict resolutions and it still passes tests).  The additional tests I'm working on adding can all go in #765.  With the parallel failure discovered in #766, we should now have a clear case of "previously-broken case which this PR fixes", which is good enough for me.

dknez:
@roystgnr I sent you a test case a while ago, did you get a chance to try this PR on that? If not, let me know and I'll try it.

roystgnr:
Not yet.  If it doesn't require a restart you could try it now; if it does then we'll need to wait until these tests pass and I rebase #765 and then we can try it on that.

dknez:
It doesn't require a restart. It just loads up a 3D mesh with overlapping nodes and checks that there's no ID clash.

roystgnr:
Yeah, that should work with this.  Let me know?  Thanks!

dknez:
I just checked that test case and it worked fine. Thanks for this PR!

roystgnr:
Okay; definitely planning to merge once the debug tests are complete, then.  And I'll put up the rebased version of #765 ASAP.  After this PR is merged, the state of libMesh will be "meshes with collocated dof objects work, until you do a restart with them, at which point dof values on collocated nodes may be silently randomly swapped"... which in hindsight is actually kind of scarier than the simple "meshes with collocated dof objects don't work"

Issue #408, opened 2014-11-12
Assertion Failure in ExodusIO when reading TRI3 mesh


pbauman:
Email me for the file if you want it - 6 elements.

I'm just meshing a circle in CUBIT (Trelis, whatever) using triangles (yes, I can use quads, yes I know I can generate a circle other ways - this is for testing/debugging before moving to more complex manifolds that may be possibly meshed with triangles).

Assertion failure is from get_side_map (it's getting called with 4 and side_map_size is 3). I'm going to try and dig into this, but was hoping someone else might've hit something similar to help accelerate things.

jwpeterson:
I think this turned out to be a Cubit issue, but if it turns out there's something we can fix in libmesh's Exodus reader, please re-open.

pbauman:
Sorry, I got derailed off this discussion with the csimsoft guys, but the current TL;DR from them is: There's no such thing as a 3D TRI3 - TRI3 is 2D and if 3D it's a shell. Which is entirely inconsistent with how they treat quads (or at least Cubit). So I still have to follow up with them. If that's the case, we might need to update the Exodus reader.

PR #722, opened 2015-11-12
Fix issue with reading 3D TRI3 meshes generated by Cubit 14, 15

jwpeterson:
There was a mailing list [discussion](http://sourceforge.net/p/libmesh/mailman/libmesh-devel/thread/CAJ-bnT7g-SOXXAnmhELhcQPGJLmmwqvro7jM0_J3fEBsVkn3uw%40mail.gmail.com/#msg32673256) and an issue (#408) about this a while back involving @dknez and @pbauman.

It recently affected one of our MOOSE users, so I took a closer look at how it might be fixed.  The basic idea is to detect the version of Cubit used to generate the Exodus file by looking at the "QA" records, and then use a workaround side numbering map for TRI3s if version 14+ and a three-dimensional mesh are detected.

I've confirmed the issue affects Cubit 14 and 15, but don't have Cubit 13 and below executables laying around to test.  If anyone wants to check an older version of Cubit, the following simple journal file can be used to generate a two-element mesh that demonstrates the issue (it will produce one sideset referring to side "4" of a TRI3 if the Cubit bug is present).

```
# Create 2D rectangular surface
create vertex 0 0 0 
create vertex 1 0 0 
create vertex 1 1 0 
create vertex 0 1 0 
create surface vertex 1,2,3,4 

# Sidesets
sideset 1 curve 1 
sideset 2 curve 2
sideset 3 curve 3
sideset 4 curve 4

# Set intervals for this surface manually
surface 1 interval 1.0

# Do meshing (with an interval of 1.0, it will simply split
# the unit square along one of the diagonals.
surface 1 scheme trimesh
mesh surf 1

# Set up block info
block 1 surf 1
block 1 element type TRI3

# Export
#
# Note - in Cubit 14 if you export with "dimension 3", you can get a
# mesh which shows the incorrect sideset IDs. That is, you should see
# the following for sideset 4:
# elem_ss4 = 2;
# side_ss4 = 4;
# On the other hand, if you export with "dimension 2" the sidest
# numbering is fine.
export mesh "/Users/petejw/Desktop/unitsquare.e" dimension 3 block 1 overwrite

# Local Variables:
# mode: shell-script
# End:
```


pbauman:
Thanks for this. I'd tried following up with them on this issue, but their "ExodusII guy" claimed that there's no such thing as a 3-D TRI3 that it should be a SHELL and it's not a bug. :-/

friedmud:
Haha! Do you remember who that was?

PR #726, opened 2015-11-16
Add Elem::key(), refactor Gmsh reader to use it

jwpeterson:
This PR adds the ability to generate a "key" for an element, which is based on its global node ids, in a similar manner to the way you could get a key for an element's sides.  This is then used to refactor and simplify the GmshIO class, which uses these keys to set boundary information in libmesh.

### DIFF
--- a/include/geom/cell_tet4.h
+++ b/include/geom/cell_tet4.h
@@ -176,6 +176,11 @@ public:
   std::pair<Real, Real> min_and_max_angle() const;
 
   /**
+   * Don't hide Tet::key(side) defined in the base class.
+   */
+  using Tet::key;
+
+  /**
    * @returns an id associated with the global node ids of this
    * element.  The id is not necessariy unique, but should be
    * close.
--- a/include/geom/cell_tet4.h
+++ b/include/geom/cell_tet4.h
@@ -175,6 +175,13 @@ public:
    */
   std::pair<Real, Real> min_and_max_angle() const;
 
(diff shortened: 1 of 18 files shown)


pbauman:
FYI, I'm getting lots of `-Woverloaded-virtual` warnings with GCC 4.8 on this method. Having looked any deeper.

jwpeterson:
Hmm... the last commit, ef1f0ce, got rid of all warnings for me.  I'll try it out on GCC4.8 locally and see if I'm still getting warnings.

pbauman:
To be more precise, I saw them with GCC 4.8.4 on my CentOS 6 workstation and GCC 4.9.3 on my OS X Yosemite laptop.

jwpeterson:
OK, yeah, I'm seeing them too now.  clang must not be as good at warning about this (or we don't have the right flags turned on) as GCC is.  I'll fix it momentarily and get you to try HEAD when you have a chance.

jwpeterson:
OK, commit f6e8730 fixes all the warnings for me, let me know if you see any others.

pbauman:
[make.txt](https://github.com/libMesh/libmesh/files/40648/make.txt)

HEY! We can attached text files now! I attached my make.log. Looks like infinite elements are the culprit (I configure with --enable-everything.)

jwpeterson:
Oops, forgot about those little guys, will clean them up in another commit.

PR #728, opened 2015-11-18
(not quite) fix std::string non-blocking send/receive

friedmud:
It looks like no one has actually tried to use a non-blocking receive of a `std::string` yet... because it doesn't compile.

This change gets it to compile: BUT IT DOES NOT WORK!

I cannot at all see why the hell not though.  I can see that the `PostWaitCopyBuffer` is getting called after `Request::wait()` is called by my code... but the buffer to copy is simply of size 0.  So, the code doesn't crash or hang or anything... it's just that nothing comes through.

If I change line 2411 there to a _blocking_ receive then this code works: the right data comes through.  (similarly, just calling the blocking form of `receive()` directly with a `std::string` works fine).

@roystgnr could I get a little help here?

### DIFF
--- a/include/parallel/parallel_implementation.h
+++ b/include/parallel/parallel_implementation.h
@@ -2402,13 +2402,13 @@ inline void Communicator::receive (const unsigned int src_processor_id,
   req.add_post_wait_work
     (new Parallel::PostWaitCopyBuffer<std::vector<T>,
      std::back_insert_iterator<std::basic_string<T> > >
-     (tempbuf, std::back_inserter(buf)));
+     (*tempbuf, std::back_inserter(buf)));
 
   // Make the Request::wait() then handle deleting the buffer
   req.add_post_wait_work
     (new Parallel::PostWaitDeleteBuffer<std::vector<T> >(tempbuf));
 
-  this->receive(src_processor_id, tempbuf, req, tag);
+  this->receive(src_processor_id, *tempbuf, req, tag);
 }
 
 
--- a/tests/parallel/parallel_test.C
+++ b/tests/parallel/parallel_test.C
@@ -25,6 +25,7 @@ public:
   CPPUNIT_TEST( testIsendRecv );
   CPPUNIT_TEST( testIrecvSend );
   CPPUNIT_TEST( testSemiVerify );
+  CPPUNIT_TEST( testDerek );
 
   CPPUNIT_TEST_SUITE_END();
 
@@ -225,7 +226,7 @@ public:
     if (TestCommWorld->size() > 1)
       {
         // Default communication
-        TestCommWorld->send_mode(Parallel::Communicator::DEFAULT);
+//        TestCommWorld->send_mode(Parallel::Communicator::DEFAULT);
 
         TestCommWorld->receive (procdown,
                                 recv_val,
@@ -242,7 +243,7 @@ public:
           CPPUNIT_ASSERT_EQUAL( src_val[i] , recv_val[i] );
 
         // Synchronous communication
-        TestCommWorld->send_mode(Parallel::Communicator::SYNCHRONOUS);
+//       TestCommWorld->send_mode(Parallel::Communicator::SYNCHRONOUS);
         std::fill (recv_val.begin(), recv_val.end(), 0);
 
 
@@ -261,11 +262,36 @@ public:
           CPPUNIT_ASSERT_EQUAL( src_val[i] , recv_val[i] );
 
         // Restore default communication
-        TestCommWorld->send_mode(Parallel::Communicator::DEFAULT);
+//        TestCommWorld->send_mode(Parallel::Communicator::DEFAULT);
       }
   }
 
 
+  void testDerek ()
+  {
+    if (TestCommWorld->rank() == 0)
+    {
+      std::vector<int> vector_data;
+      vector_data.push_back(4);
+
+      Parallel::Request request;
+      TestCommWorld->send(1, vector_data, request);
+      request.wait();
+    }
+
+    if (TestCommWorld->rank() == 1)
+    {
+      Parallel::Request request;
+      std::vector<int> vector_data;
+      TestCommWorld->receive(0, vector_data, request);
+      request.wait();
+
+      std::cerr<<"vector_data.size(): "<<vector_data.size()<<std::endl;
+
+      CPPUNIT_ASSERT_EQUAL((int)vector_data.size(), 1);
+      CPPUNIT_ASSERT_EQUAL(vector_data[0], 4);
+    }
+  }
 
   void testSemiVerify ()
   {


roystgnr:
Could you add a unit test to this PR that triggers the problem?

friedmud:
Actually... I don't even think a non-blocking receive works for `std::vector`.  I just tried the following code:


```C++
  if (processor_id() == 0)
  {
    std::vector<int> vector_data;
    vector_data.push_back(4);

    Parallel::Request request;
    _communicator.send(1, vector_data, request);
    request.wait();
  }

  if (processor_id() == 1)
  {
    Parallel::Request request;
    std::vector<int> vector_data;
    _communicator.receive(0, vector_data, request);
    request.wait();

    for (auto value : vector_data)
      std::cerr<<"Val: "<<value<<std::endl;
  }
```

And nothing comes through.

It works fine if the receive is a blocking receive.  Do you see anything I'm doing wrong... or are non-blocking receives borked even for `std::vector`?

friedmud:
Sorry: I don't know anything about unit testing in libMesh.  I just went into the `tests` directory and did a `make` and it says there is `Nothing to be done`... what's the proper way to build the unit tests?

roystgnr:
"make check" in tests should do it

friedmud:
It looks like there is a unit test that is passing a vector using non-blocking receives and it's passing.  Let me try to reproduce this issue there...

friedmud:
Ok - I've added a unit test.  Make sure you run it with 2 processors ;-)

friedmud:
I can't see why my newly added test fails... when the existing test succeeds.

friedmud:
Ok - here's the problem: the non-blocking versions of `receive()` for a `std::vector` _require_ the receiving buffer to be resized _before_ calling receive!

The blocking versions of receive, on the other hand, DO correctly resize the vector!

That means that the code that was in here for std::string won't work at all (and I suspect the code for std::set won't work either - since it's similar).

So, the question is, how do we fix it?  I actually need to send unknown sized things in a non-blocking way.  So, somehow the non-blocking receive for `std::vector` needs to be able to figure out the size and do the resize...

friedmud:
It appears as if there is no way to do this currently with MPI.  Non-blocking receives MUST have properly allocated receive buffers beforehand.

That means a few things:

  1.  I need to rethink my algorithm (not that hard, I'll just use non-blocking probes to check for messages and then do blocking receives.  BTW: How do we do a non-blocking probe with libMesh's parallel stuff?  I only see blocking probes).
  2.  The current implementations of non-blocking receive for `std::basic_string` and `std::set` are INCREDIBLY BROKEN and will never work.  I don't even see a way to make them work at all... so we should probably remove them - or just make them error.

pbauman:
Looks like we need a wrapper to [MPI_Iprobe](http://www.mpich.org/static/docs/v3.1/www3/MPI_Iprobe.html)?

roystgnr:
I believe @pbauman is correct.  MPI_Iprobe seems to have been defined since MPI-1, and seems to be exactly what we need.  For a non-blocking receive, we could give the Request object to the non-blocking probe, then add_post_wait_work with a functor that does the resizing, the actual receive, and if necessary the copy.

permcody:
Derek, FWIW I created the implementation for the string broadcast routines
and those do work. They aren't necessary the most optimal because they do
two communications, one to size things, then one to send things. Since they
are broadcasts, they are blocking anyway so I didn't run into the problem
you did. It sounds like Paul and Roy are already on the right track towards
a solution. There might even be an opportunity to improve the broadcast
implementations.

jwpeterson:
> Looks like we need a wrapper to MPI_Iprobe?

iProbe 6: it's our most advanced probe ever.

friedmud:
I'm not quite sure how you could do this automatically with iProbe. iProbe
is just non-blocking... you have to keep checking it periodically with your
thread and then, once you see that something is available, you can fire up
a receive.

So, the issue is that you need a thread of execution that is "spinning" on
iProbe. You're not really going to be able to do that within a call to
receive().

roystgnr:
Gah, you're right.  MPI_Iprobe doesn't even take an MPI_Request, does it?  We still ought to add a

Parallel::nonblocking_probe() // or whatever name

wrapper for it, but I don't see a way to do real non-blocking receives of containers with non-pre-sized or with intermediate temporary buffers.

Even if we switched back to the old "send a size_t first" method instead of relying on probes, we'd have the same problem, wouldn't we?  You can't start the real MPI_IRECV until you've got the buffer sized, you can't resize the buffer until the size information has really been received, and you can't be sure the size information has really been received until you're in a wait() unless you spin in a manual loop.

roystgnr:
MPI2 "generalized requests" use callback functions, but those don't seem to do anything that our post-wait-work idiom couldn't?  The docs suggest they only get activated by user code calling query/free/cancel methods, not by the actual request completion.

roystgnr:
Looks like even if you give up and use busy looping it's not safe to rely on MPI_Iprobe, only MPI_Test; some stacks won't progress asynchronous messages without a wait or a test call being made?

permcody:
I want to revisit this thread after chatting with Derek prior to the
Thanksgiving break. I have a much better understanding about his problem
and recognize that we both have the same needs. Essentially (in a sense) he
is (we are) trying to communicate serialized data structures. I went about
it with the packed range utilities and Derek was trying to communicate it
via strings (char[] buffers which are also variable sized). If we can get
the packed range routines going then his problem is solved. My issue is
that I still haven't figured out a satisfactory solution to those
specializations just yet. I did manage to get those routines working but it
seems like at least one of the template specializations shouldn't be
necessary but I just haven't been able to figure out how to get rid of it.

My hacked up branch (that does indeed compile) is here:
https://github.com/permcody/moose/commit/eb8014d84b118bcbbd481f0be19e5cdfaa7a92c0#diff-cbaec2e9d7f1cc328f62c7d0ba830225L24

I don't see why I should have to specialize the "OutputIter" used in the
unpack routine. In your packed elem examples you specialize it so you can
just add new Nodes and Elems to the mesh data structure but if I just want
to shove unpacked data into a std::container, it seems like I shouldn't
have to specialize this type.

I have another branch started where I'm just "all gathering" raw strings
but I'm once again starting to repeat code which is already in the
templated all_gather routines (i.e. sending a vector of sizes to calculate
offsets).

I guess I'm just looking for guidance on how you'd go about communicating
the most basic variable sized data (a variable sized buffer of chars or
something) using the packed range routines. It would be an invaluable
example for others. I've done it myself doing all my own book-keeping but
ready to move beyond that.

Cody

permcody:
UPDATE:

I've figured most of this out on my own. A little RTFM helped me find out
about libMesh::null_output_iterator which got me mostly on track. I've
created a public gist that's pretty close except for the extreme memory
leak (already time to define my own OutputIter ..... ).  Still this was a
lot to figure out and took me several hours. I'll clean this all up when
I'm done and make it a pure libMesh example with comments so others don't
have to figure this all out from scratch later.

In progress gist:
https://gist.github.com/permcody/999bc9e3ef5531a174ca

roystgnr:
To answer your earlier questions about output iterators, there's two problems with using built-in C++ iterators with libMesh packed ranges.

One problem is an API choice on my part: since our output iterators have to take ownership of heap objects, I signal this by requiring their operator= to take pointers-to-objects, not objects themselves.  Even if I changed this to take references, you'd still have an enormous memory leak when your ```ostream_iterator<std::string>``` didn't ```delete &each_string_passed;```  I could change the semantics to have the unpacking routine do the delete and thereby require output iterators to make copies of anything they wanted to keep around, but that means the common use case would be doing two or three times as much memory fiddling.

I should probably fix null_output_iterator to do those deletes.  Like you I was just debugging with it.

The other problem is an API choice in the C++ standard:
https://stackoverflow.com/questions/2619551/output-iterators-value-type
In the unpacking code I thought I could rely on ```iterator_traits<OutputIter>``` to determine what value_type to construct, but C++ standard output iterators declare a value_type of *void*.  I'm not sure where else the algorithms could or should expect to find the right value_type, though.

permcody:
I finished out the gist for anyone who cares:
https://gist.github.com/permcody/999bc9e3ef5531a174ca

I can certainly make this work with what Derek's trying to do. It certainly
simplifies the serialization/deserialization of packing data for
communication. However, building the inserter is more work than I wanted
for my case as you can see in the gist.

On Wed, Dec 2, 2015 at 6:08 PM roystgnr <notifications@github.com> wrote:

> To answer your earlier questions about output iterators, there's two
> problems with using built-in C++ iterators with libMesh packed ranges.
>
> One problem is an API choice on my part: since our output iterators have
> to take ownership of heap objects, I signal this by requiring their
> operator= to take pointers-to-objects, not objects themselves. Even if I
> changed this to take references, you'd still have an enormous memory leak
> when your ostream_iterator<std::string> didn't delete &each_string_passed;
> I could change the semantics to have the unpacking routine do the delete
> and thereby require output iterators to make copies of anything they wanted
> to keep around, but that means the common use case would be doing two or
> three times as much memory fiddling.
>
> I should probably fix null_output_iterator to do those deletes. Like you I
> was just debugging with it.
>
I guess if plan to leave it in there for debugging or not, that'd be a good
idea.

> The other problem is an API choice in the C++ standard:
> https://stackoverflow.com/questions/2619551/output-iterators-value-type
> In the unpacking code I thought I could rely on
> iterator_traits<OutputIter> to determine what value_type to construct,
> but C++ standard output iterators declare a value_type of *void*. I'm not
> sure where else the algorithms could or should expect to find the right
> value_type, though.
>
Since we've chosen to depart from using std::inserters anyway, our extended
implementation could just as well "patch" that shortcoming since we already
require the user to supply additional method implementations anyway. It
seems you've sort of done it already in your null_output_iterator where you
populate the value_type template in iterator traits:
https://github.com/libMesh/libmesh/blob/master/include/utils/null_output_iterator.h#L35
It should be possible to make an extended iterator that simply inserts into
a generic container with the appropriate value type while still supporting
the interface you want for mesh_insert_iterator.

Right now there's not a good way to pass a generic container to one of
these extended inserters that I can see since the template argument is the
value_type itself not the container type.

Thanks for the help,
Cody

roystgnr:
Building the inserter is too much work and apparently too error prone.  Your operator= appears to be a memory leak - it inserts a copy of the new object then loses the original.  I'm not sure how your gist compiles, actually... that delete loop is trying to delete std::string which isn't a pointer type.

While I'm fixing up null_output_iterator want me to try making an adapter class and possibly a named constructor for it?  Something along the lines of libMesh::make_copy_inserter<T>(standard_iterator) would return a class which for its operator= passes a copy of the dereferenced argument to the standard iterator and then delete the original.

permcody:
Wow... Yeah this program shouldn't work at all. Not only does it compile
and but it seems to run just fine and gives me the correct result for
several different numbers of processors! I'll take a closer look.

I'd be happy to try an adaptor class out.

Cody

permcody:
OK, I just fixed up the gist - should make sense now:
https://gist.github.com/permcody/999bc9e3ef5531a174ca

On Wed, Dec 2, 2015 at 11:11 PM roystgnr <notifications@github.com> wrote:

> Building the inserter is too much work and apparently too error prone.
> Your operator= appears to be a memory leak - it inserts a copy of the new
> object then loses the original.
>
The operator= method is just taking the pointer and saving it to the list.
It shouldn't need to delete anything right? I just compiled this code in
debug mode (figured out I needed a non-const iterator version of the
packed_size routine which caused me some confusion for a few minutes) and
ran it through Valgrind in parallel. I don't see any non-MPI, non-openMP
errors as it stands now.

> I'm not sure how your gist compiles, actually... that delete loop is
> trying to delete std::string which isn't a pointer type.
>
> While I'm fixing up null_output_iterator want me to try making an adapter
> class and possibly a named constructor for it? Something along the lines of
> libMesh::make_copy_inserter(standard_iterator) would return a class which
> for its operator= passes a copy of the dereferenced argument to the
> standard iterator and then delete the original.
>
Since there are at least 5 related methods that should all be templated
from the same type maybe this whole design should be part of a abstract
templated class. It's definitely a little tricky to get them all right
together and there's not much point in having a subset of them defined.

Thanks for the guidance,
Cody

roystgnr:
If you're inserting the provided pointer into a list of pointers, then operator= shouldn't delete anything.

Originally you were saving a copy of the value pointed to into a list of values, and to avoid a memory leak in that case you would have needed to delete the original.

Which semantics would you like for the adapter?  A container of values is easier to play with than a container of pointers to heap-allocated values, but requires extra memory fiddling to fill.

permcody:
That's a good question so let me explain our use case: 

We need to ship state data that lives on elements (stateful material properties) to neighboring processors due to repartitioning (with adaptivity). As you may be aware, these properties are user-defined and can be any type so we serialize these properties into a raw byte buffer. The actual buffer itself likely lives on the heap depending on the implementation (currently ostringstream) but we probably have to copy that value to the stack anyway as we retrieve and prepare the buffer for shipment. This is why Derek was interested in sending std::strings in the first place. As you can see by my gist, the current solution is already doing excess copying as we have to also promote the chars in the buffer (std::string -> std::vector<long long int>). I think no matter what we do we'll end up copying at least once even if we avoid the type promotion as we pack and encode the buffer lengths.

BTW - My use case is the same. I'm using the same templated MOOSE "dataLoad/dataStore" routines to serialize/deserialize my structures so I end up with an equivalent raw buffer as well.

I don't think I answered your question, but I'm still thinking about how to clean this up some more.

roystgnr:
How crazy would it be if I deprecated or outright trashed the templated packed_size/packable_size/pack/unpack function API and instead used static member functions of a templated class?

I thought I would be clever by making an adapter iterator test code support ```std::basic_string<T>``` instead of plain ```std::string```, but now I'm discovering that the inability to partially specialize function templates is a seemingly insurmountable pain in the neck.  I think the [Dimov/Abrahams example](http://www.gotw.ca/publications/mill17.htm) has completely eclipsed my previous candidate for [the most annoying part of C++](https://github.com/grinsfem/grins/commit/6e1e8d12e300fc17567756aaa5660a88777eb470).

roystgnr:
Wait, you just said "maybe this whole design should be part of a abstract templated class".  I dismissed the "abstract" part (dispatch needs to be compile-time not run-time here) and then foolishly ignored the core of the advice...

permcody:
I have my doubts that anyone is exploiting the existing API. Maybe I
overestimate my ability but the issues with the inserter drove me crazy so
I assume most other coders would have have at least queried the list before
implementing some new packed type. Therefore, trash away :)

BTW - I managed to switch the underlying type in my string gist to "char"
instead of "long long int" saving a fair amount of storage and copying work
when populating the communication buffer. I'm starting to use it now to
communicate my arbitrary data structures. I'll contribute the string
example back after you've finished fiddling with the API and I've adapted
it to my existing code.

roystgnr:
While we're at it I'll break receive_packed_range() - if we add a pointer-to-T argument (unused, NULL acceptable) then we won't have to infer T from iterator_traits<OutputIter> and then we won't have to have our own special snowflake definition of how an output iterator should behave.

permcody:
Yes!

roystgnr:
send_receive_packed_range() will need the extra argument too; I don't think we've ever *used* the ability to send objects of one type while receiving objects of another type (whether plain or packed), but we might as well not deliberately break it.

jwpeterson:
> How crazy would it be if I deprecated or outright trashed the templated packed_size/packable_size/pack/unpack function API and instead used static member functions of a templated class?

Apparently just crazy enough to work?  @roystgnr @permcody now that we merged #802 is this PR still relevant?

roystgnr:
I'd wait for @friedmud to chime in, but IMHO half of what he wanted is fixed and the other half unfixable.

friedmud:
@permcody is after the same thing I was... So if it works for him then I'm happy

permcody:
Right, and we now have the ability to communicate strings although not with this non-blocking probe that Derek was after. My needs are slightly different from his but we can probably close this ticket for now as we have what we need for the immediate future.

PR #729, opened 2015-11-19
Avoid clear() in DiffSystem class destructors

roystgnr:
The previous code doesn't do what it might be expected to do.  Since
the methods called by clear are virtual, and any subclasses have been
partially-destructed down to the base class at this point, and the
base class implementations of physics and qoi clearing are no-ops,
calling those functions does nothing.  If clearing of subclass data
needs to be done then it needs to be done in the subclasses'
destructor.

On Intel 16 in opt mode, those virtual function calls give me a
segfault, which I'm pretty sure is a compiler bug, but since fixing my
stupid code also avoids the segfault I'm not complaining too hard.

### DIFF
--- a/include/systems/fem_system.h
+++ b/include/systems/fem_system.h
@@ -78,12 +78,6 @@ public:
   typedef DifferentiableSystem Parent;
 
   /**
-   * Clear all the data structures associated with
-   * the system.
-   */
-  virtual void clear () libmesh_override;
-
-  /**
    * Prepares \p matrix or \p rhs for matrix assembly.
    * Users may reimplement this to add pre- or post-assembly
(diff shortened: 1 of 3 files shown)


pbauman:
:+1:

roystgnr:
@jwpeterson, could we cherry-pick this into 0.9.5?  As soon as I saw the segfault stack trace with 0.9.5-rc2 I realized, "this looks awfully familiar..."

jwpeterson:
Sure.  So this fixes the segfault for you, eh?

roystgnr:
It did on master.  Haven't tried applying it to 0.9.5-rc2 yet.

PR #730, opened 2015-11-19
Add configure tests for unordered_multiset variants

jwpeterson:
I initially thought I would use these in some of the IO classes (and therefore added some useful stuff in the new elem_hash.h header), but it turned out that `unordered_multimap` was faster for what I was doing.  I think `unordered_multiset<Elem*>` could be useful for other things in the future, though, so I left it in.

### DIFF
--- a/configure
+++ b/configure
@@ -27993,6 +27993,336 @@ fi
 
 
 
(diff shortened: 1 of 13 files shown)


roystgnr:
:+1: once it passes, thanks!

PR #731, opened 2015-11-20
Refactor Elem::key(), side() implementations

jwpeterson:
We had a lot of redundant code in these functions that should just have been using the appropriate `side_nodes_map` arrays.


pbauman:
> 61 additions and 432 deletions.

Win.

jwpeterson:
cloc_libmesh

If we all work together, maybe we can actually reverse the trends on this graph :smile:

roystgnr:
:+1:

PR #732, opened 2015-11-21
Refactor Elem::build_side() routines

jwpeterson:
Similar to the work in #731, this just clears out a bunch of redundant code from the non-proxy branches of the various `build_side()` routines.


roystgnr:
I feel like we need to submit a bug report to git-fame, now that it's become obvious that project contributions should be ranked by number of lines removed, not number of lines added.

permcody:
+1

PR #733, opened 2015-11-27
Make Elem::key() implementations consistent with Elem::key(side)

jwpeterson:
The key computed for the quadratic side of an element should be consistent with the key computed for the lower-dimensional element itself.  Also, a few of the infinite element types did not follow the convention of using mid-edge/face nodes for computing keys, so fixed those.

### DIFF
--- a/include/geom/cell_hex.h
+++ b/include/geom/cell_hex.h
@@ -120,7 +120,7 @@ public:
 
   /**
    * @returns an id associated with the \p s side of this element.
-   * The id is not necessariy unique, but should be close.  This is
+   * The id is not necessarily unique, but should be close.  This is
    * particularly useful in the \p MeshBase::find_neighbors() routine.
    */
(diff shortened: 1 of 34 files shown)


roystgnr:
Looks reasonable, and makes the behavior more intuitive.  :+1:

Is there a use case where you'd be using side keys and elem keys in the same map?

jwpeterson:
> On Nov 29, 2015, at 7:17 PM, roystgnr <notifications@github.com> wrote:
> 
> Is there a use case where you'd be using side keys and elem keys in the same map?
> 

Not currently, but it's certainly possible.  Did you have something particular in mind or see a possible issue there?

roystgnr:
I've been skimming the code trying to figure out what this could possibly break or fix in the multi-dimensional mesh case, but I don't see any issues.  We make maps with elem keys, we make maps with side keys, but we don't seem to mix them.

PR #734, opened 2015-11-27
Fix DenseMatrix::svd(sigma, U, VT);

jwpeterson:
Lapack is actually computing the SVD of A^T, so we have to swap the
meaning of U and V before we hand them back to the user.

This fixes the issue reported by Jonas Ballani on the mailing list.  Once it passes the CI testing, I'll also cherry-pick this bugfix onto 0.9.5.

### DIFF
--- a/include/numerics/dense_matrix_base.h
+++ b/include/numerics/dense_matrix_base.h
@@ -121,7 +121,7 @@ public:
    * Prints the matrix entries with more decimal places in
    * scientific notation.
    */
-  void print_scientific(std::ostream& os) const;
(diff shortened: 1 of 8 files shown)


roystgnr:
:+1: happily if you'll add a unit test.

:+1: even if you don't add a unit test, because who am I to talk?

jwpeterson:
> On Nov 29, 2015, at 7:18 PM, roystgnr <notifications@github.com> wrote:
> 
>  happily if you'll add a unit test.
> 
>  even if you don't add a unit test, because who am I to talk?
> 

I'll add one based on the test code we got on the mailing list.

PR #735, opened 2015-12-01
Fix PETSc m4 test for petscconf.h

jwpeterson:
We need to be a little more careful while grepping in the petscconf.h file.  It may be in a different location depending on how PETSc was built, or the grep command could simply fail for any reason at all, in which cases the user will see something like:
```
../configure: line 36562: test: -gt: unary operator expected
```

This patch should handle both issues.

### DIFF
--- a/configure
+++ b/configure
@@ -31742,9 +31742,14 @@ $as_echo "<<< PETSc 2.x detected and \"\$PETSC_ARCH\" not set.  PETSc disabled. 
 
       # Determine if PETSc has been built with debugging enabled.
       # Note that this token will appear as LIBMESH_PETSC_USE_DEBUG in
-      # our header, so it won't collide with PETSc's.
-      petsc_use_debug=`grep -c PETSC_USE_DEBUG $PETSC_DIR/include/petscconf.h`
-      if (test $petsc_use_debug -gt 0); then
+      # our header, so it won't collide with PETSc's.  We look for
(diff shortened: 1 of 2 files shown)


roystgnr:
Yikes - that *is* bash specific!  There goes a few dozen characters I was proud of cleverly saving.

configure uses /bin/sh, which on my Ubuntu system is /bin/dash, which I believe is pretty plain-Bourne-shell compliant and which definitely doesn't support simple brace expansion.

jwpeterson:
> Yikes - that is bash specific!

It definitely doesn't work with `/bin/dash`, so the rest of this is a moot point, but I wouldn't say it's bash-specific.  The following line:

```
touch testing_file{1,2}.txt
```

works fine for me on all the following shells on OSX:

/bin/csh -> tcsh 6.18.01 
/bin/zsh -> zsh 5.0.8 
/bin/bash -> GNU bash, version 3.2.57(1)-release
/bin/ksh -> (AT&T Research) 93u+ 2012-08-01 
/bin/tcsh -> tcsh 6.18.01 (Astron) 2012-02-14

roystgnr:
Sorry, I meant in the bash-vs-POSIX-sh context.  StackExchange thinks that brace expansion started in csh in the 70s and made it to practically everything not-set-in-ISO-stone by 2005.  https://unix.stackexchange.com/questions/92819/why-brace-expansion-not-supported

PR #739, opened 2015-12-02
DenseMatrixBase::multiply() helper can be static

roystgnr:
Just something I ran across while working on other code.

### DIFF
--- a/include/numerics/dense_matrix_base.h
+++ b/include/numerics/dense_matrix_base.h
@@ -137,14 +137,15 @@ public:
 protected:
 
   /**
+   * Helper function -
    * Performs the computation M1 = M2 * M3 where:
    * M1 = (m x n)
    * M2 = (m x p)
    * M3 = (p x n)
    */
-  void multiply (DenseMatrixBase<T>& M1,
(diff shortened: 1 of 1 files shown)


jwpeterson:
There are a couple of places in DenseMatrix<T> where we explicitly call:
```
this->multiply(*this, M2, M3);
```
Clearly the compiler is fine with this, but I wonder if that would look better/make more sense to change it to:
```
DenseMatrixBase<T>::multiply(*this, M2, M3);
```
?

roystgnr:
I was actually *planning* to change it, then happily surprised that it compiled without the change.

I do admit that the less terse syntax seems slightly less ambiguous to me.  I'd still rather not change it, but I'll so so if you prefer.  I suspect my resistance is about 50% "don't make things less terse needlessly" and 50% "now I can't make fun of @pbauman for ending all his void functions with a redundant return;"

jwpeterson:
Fine with me if you don't want to change it, looks like it is [officially allowed by the standard](http://stackoverflow.com/questions/325555/c-static-member-method-call-on-class-instance) anyway.

As far as making fun of @pbauman, I think that should continue.

PR #741, opened 2015-12-03
Check for mismatch between dof_id_type and PetscInt at configure time

jwpeterson:
This can lead to weird runtime errors if it happens.  We have a `static_assert` that also checks for this in petsc_vector.h, but that is only for C++11 compilers, and you still waste your time compiling about half of libmesh before you realize that you need to start over.

### DIFF
--- a/configure
+++ b/configure
@@ -31758,7 +31758,20 @@ $as_echo "#define PETSC_USE_DEBUG 1" >>confdefs.h
         enablepetsc=no
     fi
 
+    # If PETSc is using 64-bit indices, make sure that
+    # $dof_bytes==8, or else print an informative message and
+    # disable PETSc.
+    if (test $enablepetsc != no) ; then
+      petsc_use_64bit_indices=`cat ${PETSC_DIR}/include/petscconf.h ${PETSC_DIR}/${PETSC_ARCH}/include/petscconf.h 2>/dev/null | grep -c PETSC_USE_64BIT_INDICES`
 
+      if (test $petsc_use_64bit_indices -gt 0) ; then
+        if (test $dof_bytes != 8) ; then
+          { $as_echo "$as_me:${as_lineno-$LINENO}: result: <<< PETSc is using 64-bit indices, you must configure libmesh with --with-dof-id-bytes=8. >>>" >&5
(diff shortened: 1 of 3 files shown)


roystgnr:
Sorry to be a PITA, but could we move this test to a separate m4 file?  I've got other projects that use libMesh petsc.m4 unchanged (which is a life saver every time they discover that PETSc has again played hide-and-seek with its config files), but those don't define $dof_bytes.

A separate CHECK_LIBMESH_TYPE_CONSISTENCY macro might be more self-documenting, too - as it is, we could someday innocently swap the order of id type selection and PETSc detection and then configuration would break for no initially obvious reason.

jwpeterson:
I thought you might say that...  I am also going to change it to an `AC_MSG_ERROR` when there's a mismatch.  Since no one (including me) actually checks the configure summary output, they are not likely to notice that PETSc has been silently disabled on them until they've built everything anyway.

roystgnr:
Thanks!

Totally agree re: AC_MSG_ERROR - I'd assumed you'd do that and hadn't even thought to double-check!

While I'm looking more carefully and you're having to make changes anyway, two other issues:

You're catching the "libMesh configured as 32-bit, PETSc as 64" case; while we're at it we should scream-and-die in the "64-bit libMesh, 32-bit PETSc" case too.

Should we also check for consistency between libMesh --enable-complex and PETSc --with-scalar-type=complex?

roystgnr:
I'm also tempted to say "let's do the PETSc detection before the id type selection, then we can use PETSc values as our defaults"... but that might be going too far.

pbauman:
http://imgur.com/gallery/t0XHtgJ

jwpeterson:
http://imgur.com/gallery/t0XHtgJ

+1

jwpeterson:
> You're catching the "libMesh configured as 32-bit, PETSc as 64" case; while we're at it we should scream-and-die in the "64-bit libMesh, 32-bit PETSc" case too.
> 
> Should we also check for consistency between libMesh --enable-complex and PETSc --with-scalar-type=complex?

Yes, I'll check for these situations as well.

PR #743, opened 2015-12-04
Updated systems_of_equations_ex8.

dknez:
The previous version of systems_of_equations_ex8 was unnecessarily complicated for the sake of an example, and didn't converge very well. In this updated version, we restrict to "small sliding" contact,
and we use the augmented Lagrangian method to achieve high accuracy in the contact solve.

One issue I haven't figured out yet is the sparsity pattern augmentation. The augment_sparsity_pattern
code in this example seems to have no effect. Any insights into this would be appreciated. Does
sparsity pattern augmentation not work for NonlinearImplicitSystems, perhaps?

In the meantime, I've turned off PETSc's sparsity pattern check (see
LinearElasticityWithContact::residual_and_jacobian).

### DIFF
--- a/examples/systems_of_equations/systems_of_equations_ex8/augment_sparsity_on_contact.C
+++ b/examples/systems_of_equations/systems_of_equations_ex8/augment_sparsity_on_contact.C
@@ -14,32 +14,16 @@ AugmentSparsityOnContact::AugmentSparsityOnContact(
   _sys(sys)
 {}
 
-void AugmentSparsityOnContact::clear_contact_element_map()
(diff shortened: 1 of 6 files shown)


dknez:
This is in good shape now. I realized that the issue with the sparsity pattern was that PETSc condenses any unused entries from the sparsity pattern after the first matrix assembly. To fix that I now add zeros to the jacobian as placeholders for the contact force terms, which fixes the issue.

This change only affects one example, and I think it's an improvement over the previous version of this example, so I'll go ahead and merge once the tests pass.

PR #744, opened 2015-12-04
If starting the nonlinear solve with a really good initial guess, we …

vikramvgarg:
…dont want to set an absurd linear tolerance

### DIFF
--- a/src/solvers/newton_solver.C
+++ b/src/solvers/newton_solver.C
@@ -371,6 +371,9 @@ unsigned int NewtonSolver::solve()
           current_linear_tolerance = minimum_linear_tolerance;
         }
 
+      // If starting the nonlinear solve with a really good initial guess, we dont want to set an absurd linear tolerance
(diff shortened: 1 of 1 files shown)


vikramvgarg:
All tests passed on make check.

jwpeterson:
@vikramvgarg I think we are waiting on you to confirm that this indeed works for the specific situation you coded it to address, this line of code is most likely not exercised by any of the existing examples in libMesh...

vikramvgarg:
Yeah, still testing that.

jwpeterson:
OK, thanks.

roystgnr:
It works correctly.  Not quite as *well* as I'd hoped (it requires the user to set a sane absolute_residual_tolerance even if they were previously relying on relative tolerances, and the safety factor is definitely necessary but still leads to some oversolving) but it's better than we had before on Vikram's solved-restart case.

PR #746, opened 2015-12-08
Add DenseMatrix::svd_solve().

jwpeterson:
This function fills a missing requirement in the DenseMatrix classes,
allowing us to solve non-square systems of equations in a
least-squares sense.  The user can pass a tolerance to svd_solve()
which determines the cutoff for small singular values. svd_solve() is
a const member function: we make a copy internally instead of allowing
Lapack to modify A.

Note that Eigen also has the capability to solve non-square systems of
equations, but it is relatively slow, as discussed in this thread:
https://forum.kde.org/viewtopic.php?f=74&t=102088, so having our own
Lapack-based implementation is worthwhile.

### DIFF
--- a/include/numerics/dense_matrix.h
+++ b/include/numerics/dense_matrix.h
@@ -395,6 +395,26 @@ public:
    */
   void svd(DenseVector<T>& sigma, DenseMatrix<T>& U, DenseMatrix<T>& VT);
 
+  /**
+   * Solve the system of equations A*x = rhs for x in the
+   * least-squares sense. $A$ may be non-square and/or rank-deficient.
+   * You can control which singular values are treated as zero by
+   * changing the "rcond" parameter.  Singular values S(i) for which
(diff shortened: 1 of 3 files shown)


jwpeterson:
libmesh_vs_eigen

The Lapack-based implementation is much faster than Eigen's.  As mentioned above, this seems to be a [known issue](https://forum.kde.org/viewtopic.php?f=74&t=102088).

roystgnr:
:+1:

When we were talking earlier you seemed to think Eigen might still have a shot at winning in extremely-rectangular cases.  I take it that didn't pan out?

jwpeterson:
It did not.  I finally realized that "MatrixXf" was doing single-precision calculations... Changing it to double-precision made it uncompetitive with Lapack dgelss in all situations I tested.  dgelss is even faster than dgesvd if all you want is a least squares solution.

PR #747, opened 2015-12-08
Cache derivatives and remove a byte code round trip for FPOptimization

### LINKED ISSUE #745: Automatic Differentiation is slow for large expressions
The problem is actually _not_ the AD algorithm, but the byte code synthesizer in FParser that transforms the tree structure into a stack-based VM byte code. Initial profiling indicates that a specific optimization (looking for common subexpressions) takes a lot of time.

It seems there is not much I can do to optimize this, so I propose to mitigate this issue by providing aggressive caching for the bytecode synthesis (or the derivative construction as a whole). I've already implemented bytecode hashing to avoid repeated JIT compilation. This could be exploited for the automatic differentiation as well. I would need to implement (un)serialization of FParser objects to make this happen (should be fairly easy).

dschwen:
This adds a caching mechanism for derivatives that provides a _substantial_ speed up for long function expressions (such as free energies from thermodynamic databases where it can cut _minutes_ of byte code synthesis down to sub second cache loads).

I've also added an optional immediate FPOptimizer call that operates on the freshly generated derivative tree, thus eliminating a roundtrip `-> tree -> byte code`. And I added a feature flag system to avoid having an API where the user needs to pass in multiple bools into several functions (which makes default arguments difficult).

<s>This will need a MOOSE update!</s> (actually MOOSE builds fine with those changes)

Closes #745

### DIFF
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -12,6 +12,14 @@ using namespace FPoptimizer_CodeTree;
 
 #include <iostream>
 
+#include <sys/time.h>
+
+struct timeval tp;
+long int timestamp() {
+  gettimeofday(&tp, NULL);
+  return tp.tv_sec * 1000 + tp.tv_usec / 1000;
+}
+
 #if LIBMESH_HAVE_FPARSER_JIT
 #  include <fstream>
 #  include <cstdio>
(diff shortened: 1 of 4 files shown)


dschwen:
Crap I'm running into clang/g++ differences again. This compiles fine on my machine.

jwpeterson:
Looks like a bunch of linking failures on `std::complex` template specializations?  This seems familiar, but I can't remember how you fixed it before.

contrib/fparser/fparser_ad.cc
```diff
@@ -386,15 +387,85 @@ template<typename Value_t>
...
 {
   this->CopyOnWrite();
+  const bool cached = mADFlags & ADCacheDerivatives;
 
   try
   {
-    return ad->AutoDiff(LookUpVarOpcode(var_name), this->mData);
+    unsigned int var_number = LookUpVarOpcode(var_name);
+
+    // should and cen we load a cached derivative?
```
jwpeterson: typo "cen"

contrib/fparser/fparser_ad.cc
```diff
@@ -386,15 +387,85 @@ template<typename Value_t>
...
+    if (cached)
+    {
+      // generate a sha1 hash of the Value type size, byte code, and immediate list
+      SHA1 *sha1 = new SHA1();
+      char result[41]; // 40 sha1 chars plus null
+      size_t value_t_size = sizeof(Value_t);
+      sha1->addBytes((char*) &value_t_size, sizeof(value_t_size));
+      sha1->addBytes((char*) &this->mData->mByteCode[0], this->mData->mByteCode.size() * sizeof(unsigned));
+      if (!this->mData->mImmed.empty())
+        sha1->addBytes((char*) &this->mData->mImmed[0], this->mData->mImmed.size() * sizeof(Value_t));
```
jwpeterson: IMO it's better to spell out `reinterpret_cast` rather than C-style casting here, but this is your deal...

contrib/fparser/fparser_ad.hh
```diff
@@ -38,10 +38,41 @@ public:
...
 
+  // feature flags for this parser
+  enum ADFlags {
+    /**
+     * In certain applications derivatives are built proactively and may never be used.
+     * We silence all AutoDiff exceptions in that case to avoid confusing the user.
+     */
+    ADSilenceErrors = 1,
+    /**
+     * Immediately apply the optimizwer grammars to the derivative tree structure.
```
jwpeterson: typo "optimizwer"

contrib/fparser/fparser_ad.cc
dschwen: Yeah, I'should probably change this. I was sticking to the (ugly) FParser code conventions. But why should I?

PR #748, opened 2015-12-09
Proposed packed_range refactor

roystgnr:
Make functions into class static methods, so we can do partial template specialization on the class.

pack() can take an output iterator for future flexibility (although for Elem and Node we'll only instantiate the methods which take a back_inserter)

Use references to T rather than pointers to T.  This still lets us use references-to-pointers-to-heap-objects in the Elem* and Node* cases, but lets us manipulate objects on the stack if we prefer in other specializations, and (given the proper Packing<Foo> and Packing<Foo*> specializations) should even let us mix and match between the two options when sending and receiving.

Use a return value rather than a return parameter in unpack

Remove the now-redundant function specialization argument from packed_size

Add a function specialization argument to receive_packed_range and send_receive_packed_range, so we can now use standard output iterators whose underlying type can't be inferred from traits.

Use the new Packing API with Node* and Elem* types

Add unit tests with std::string

This is based on the discussion and advice from @permcody in #728 and emails.

This PR changes the packed_range code from "a neat refactoring of internal DofObject communications" to "intended for public specialization and general use", so now is the time for complaints, suggestions, etc. about the new API.

### DIFF
--- a/include/parallel/parallel.h
+++ b/include/parallel/parallel.h
@@ -526,80 +526,55 @@ inline void wait (std::vector<Request> &r)
 
 
 /**
- * Define the data type to be used for data arrays whne encoding
- * a potentially-variable-size object of type T.
+ * Define data types and (un)serialization functions for use when
+ * encoding a potentially-variable-size object of type T.
+ *
+ * Users will need to specialize this class for their particular data
+ * types.
  */
 template <typename T>
-struct BufferType;
+class Packing {
+public:
+  // Should be an MPI sendable type in specializations, e.g.
+  // typedef char buffer_type;
+  // typedef unsigned int buffer_type;
 
-/**
- * Encoding non-const T* always uses the same buffer type as
- * encoding const T*, so other classes only have to specialize the
- * latter.
- */
-template <typename T>
-struct BufferType<T*> {
-  typedef typename BufferType<const T*>::type type;
+  // Should copy an encoding of the provided object into the provided
+  // output iterator (which is of type buffer_type)
+  template <typename OutputIter, typename Context>
+  static void pack(const T& object,
+                   OutputIter data_out,
+                   const Context* context);
+
+  // Should return the number of array entries (of type buffer_type)
+  // required to encode the provided object
+  template <typename Context>
+  static unsigned int packable_size(const T& object,
+                                    const Context* context);
+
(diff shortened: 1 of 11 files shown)


permcody:
I think the new design is much improved. I like that all the methods are specialized together in a single template class rather than disparate separate methods. 
:+1: - No need to build a specialized inserter
:+1: - Ability to use reference to types instead of pointer types

I've been using the packed range stuff for moving around my raw buffer strings so I'll snag a copy of this branch once it's passing tests and start using it. I don't have any gripes about the interface so far. I'll let you know if I see something.

roystgnr:
This isn't getting properly tested, because the API change breaks MooseMesh.C.

Suggestions?  I can run the test suite myself, then when we update the Moose libMesh submodule we'll make the API change concurrently.

roystgnr:
But we'd need to update ASAP, because our regression tests would be near-useless until we did.

jwpeterson:
The main thing holding up a libmesh update in MOOSE are the performance regressions and crashes caused by the changes to system_projection.  I just opened a PR to test my `remove_id()` stuff over at idaholab/moose#6075, you can see what's still failing.

jwpeterson:
But in order to test this with Moosebuild, you can push your packed_range_refactor branch into libmesh upstream, make whatever fixes are necessary in MooseMesh.C, and then open a MOOSE PR which updates the libmesh submodule to the upstream hash of packed_range_refactor...

permcody:
Now that MOOSE is "closer" to libMesh head, (Updated 01/12/2016). I'd like to see this PR revived. As John suggested, your PR should just include the appropriate hash to update to (preferably one on this side branch that doesn't have the projection logic) and the corresponding fix necessary for MooseMesh. Both @dschwen and I have immediate use for this refactor once it's in.

jwpeterson:
I'm going to take a crack at rebasing this branch tomorrow.  (There are quite a few conflicts due to the operator spacing commits.) 

@permcody would like it to be merged ASAP, and apparently it's *almost* passing all the MOOSE tests already.  If necessary, we can even merge it before it's actually passing all the MOOSE tests in the CI, as long as we can prepare a libmesh update for MOOSE in advance...

roystgnr:
Obsoleted by #799

PR #749, opened 2015-12-11
PETSc 3.6.3 still has old TaoSetTolerances options

pbauman:
With this change, I can build libMesh with the PETSc 3.6.3 release.


jwpeterson:
Right, I'm thinking it will need to be 3.7.0 eventually...

PR #750, opened 2015-12-14
Widen NewtonSolver immediate-convergence test

roystgnr:
Not only should we converge on our initial iterate if it has a zero
residual, we should converge if it has a residual under our
absolute_residual_tolerance.

This should fix the cases where we try to re-solve an oversolved
solution and we tell PETSc to use a linear tolerance above 1.0

I haven't finished compiling and testing this myself yet; just wanted to get it out there for @pbauman to try ASAP.

### DIFF
--- a/src/solvers/newton_solver.C
+++ b/src/solvers/newton_solver.C
@@ -324,24 +324,29 @@ unsigned int NewtonSolver::solve()
           continue;
         }
 
-      if (current_residual == 0)
+      if (current_residual <= absolute_residual_tolerance)
         {
           if (verbose)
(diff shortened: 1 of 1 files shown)


roystgnr:
This is in response to regression cases introduced by #744 - sorry we left a known bug lying around for days.

pbauman:
This fixed the reference regression cases for me. Thanks for the quick turnaround!

roystgnr:
I'd like to merge this ASAP, but where did @moosebuild go?

jwpeterson:
> I'd like to merge this ASAP, but where did @moosebuild go?

Not sure, we're looking into it...

PR #754, opened 2015-12-16
Add BoundaryInfo::remove_id()

jwpeterson:
It is already possible to remove sides/edges/nodes from the BoundaryInfo, but if, for example, you remove *all* of the geometric entities with a given ID, that ID may still appear in the following `BoundaryInfo` data structures:
```
_ss_id_to_name
_ns_id_to_name
_boundary_ids
_side_boundary_ids
_edge_boundary_ids
_node_boundary_ids
```

This can lead to slightly confusing behavior in calls to `n_boundary_ids()` and other functions.  This fixes a minor issue over in idaholab/moose#3554, where a boundary id is updated due to Mesh extrusion.

The branch also adds a unit test of the new capability.

### DIFF
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -115,7 +115,7 @@ public:
    * pass a pointer to both the boundary_mesh's MeshData object,
    * and the MeshData object used for this mesh.
    */
-  void sync (const std::set<boundary_id_type> &requested_boundary_ids,
+  void sync (const std::set<boundary_id_type>& requested_boundary_ids,
              UnstructuredMesh& boundary_mesh,
              MeshData* boundary_mesh_data=NULL,
              MeshData* this_mesh_data=NULL);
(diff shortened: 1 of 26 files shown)


include/mesh/boundary_info.h
```diff
@@ -115,7 +115,7 @@ class BoundaryInfo : public ParallelObject
    * pass a pointer to both the boundary_mesh's MeshData object,
    * and the MeshData object used for this mesh.
    */
-  void sync (const std::set<boundary_id_type> &requested_boundary_ids,
+  void sync (const std::set<boundary_id_type>& requested_boundary_ids,
```
roystgnr: Are we getting OCD about ```&``` and ```*``` placement now?  Can I put in a vote for "separate it by spaces on both sides in declarations"?  I was surprised at how much more readable that style made MOOSE code.

roystgnr:
:+1:

include/mesh/boundary_info.h
jwpeterson: > Are we getting OCD about & and * placement now? Can I put in a vote for "separate it by spaces on both sides in declarations"? I was surprised at how much more readable that style made MOOSE code.

Absolutely, "space on both sides" is now my preference as well.  I changed these ones because the original libmesh convention was always "space on the right" and the rest of the file used that convention except for a few lines...

PR #756, opened 2015-12-18
Prevent user-supplied mesh dimension from being ignored

bwspenc:
The mesh dimension supplied in the constructors to MeshBase currently gets put into _elem_dims, but when prepare_for_use() gets called, that gets cleared, and it gets lost.  As a result, if the user asked for a mesh with dimension 3, but it only contains elements of dimension 1 or 2, the mesh dimension is reported as the lower number.  This fixes that issue.

My interest in doing this is that I'm trying to use a 3d or 2d truss mesh composed only of line elements.  The elements themselves are 1d, but the nodal coordinates are 2d or 3d.  With this fix, I can pass a 3 in to the constructor, and then the mesh dimension is reported as 3, and my MOOSE job runs happily.  I should note that this patch breaks many MOOSE tests because it currently passes 3 in to the MeshBase constructor, which then gets ignored. Passing in 1 instead by default from the MOOSE side fixes that.

Ideally, I'd like to not have to tell it that it's a 3d mesh and have that be detected. I see that there is a spatial dimension and a mesh dimension.  I might have the terminology wrong, but I think what I have here is a mesh that is spatially 3D, but topologically 1D.  If that's correct, I think what I really need is a way to get the spatial dimension of my mesh. The spatial dimension is hardcoded to LIBMESH_DIM, so that isn't useful for the application to figure out what the actual dimension of the current mesh is.

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -1052,6 +1052,11 @@ protected:
   std::map<subdomain_id_type, std::string> _block_id_to_name;
 
   /**
+   * User-supplied mesh dimension.
+   */
+  unsigned char _user_dim;
+
+  /**
    * We cache the dimension of the elements present in the mesh.
    * So, if we have a mesh with 1D and 2D elements, this structure
    * will contain 1 and 2.
--- a/src/mesh/mesh_base.C
+++ b/src/mesh/mesh_base.C
@@ -56,9 +56,10 @@ MeshBase::MeshBase (const Parallel::Communicator &comm_in,
   _next_unique_id(DofObject::invalid_unique_id),
 #endif
   _skip_partitioning(libMesh::on_command_line("--skip-partitioning")),
-  _skip_renumber_nodes_and_elements(false)
+  _skip_renumber_nodes_and_elements(false),
+  _user_dim(d)
 {
-  _elem_dims.insert(d);
+  _elem_dims.insert(_user_dim);
   libmesh_assert_less_equal (LIBMESH_DIM, 3);
   libmesh_assert_greater_equal (LIBMESH_DIM, d);
   libmesh_assert (libMesh::initialized());
@@ -77,9 +78,10 @@ MeshBase::MeshBase (unsigned char d) :
   _next_unique_id(DofObject::invalid_unique_id),
 #endif
   _skip_partitioning(libMesh::on_command_line("--skip-partitioning")),
-  _skip_renumber_nodes_and_elements(false)
+  _skip_renumber_nodes_and_elements(false),
+  _user_dim(d)
 {
-  _elem_dims.insert(d);
+  _elem_dims.insert(_user_dim);
   libmesh_assert_less_equal (LIBMESH_DIM, 3);
   libmesh_assert_greater_equal (LIBMESH_DIM, d);
   libmesh_assert (libMesh::initialized());
@@ -100,6 +102,7 @@ MeshBase::MeshBase (const MeshBase& other_mesh) :
 #endif
   _skip_partitioning(libMesh::on_command_line("--skip-partitioning")),
   _skip_renumber_nodes_and_elements(false),
+  _user_dim(other_mesh._user_dim),
   _elem_dims(other_mesh._elem_dims)
 {
   if(other_mesh._partitioner.get())
@@ -510,6 +513,9 @@ void MeshBase::cache_elem_dims()
   // Need to clear _elem_dims first in case all elements of a
   // particular dimension have been deleted.
   _elem_dims.clear();
+
+  //Insert the user-supplied mesh dimension
+  _elem_dims.insert(_user_dim);
 
   const_element_iterator el  = this->active_elements_begin();
   const_element_iterator end = this->active_elements_end();


roystgnr:
I fear this will break others' apps; ISTR some which rely on the current behavior of "reading a mesh in from file overrides the previously supplied mesh_dimension".  It definitely breaks the API as defined, where mesh_dimension() returns the largest dimension of any element in the mesh.

If we wanted to make an API more clever (and we do; you've clearly got a reasonable use case), then spatial_dimension() would be the one to use IMHO.  Its current implementation is obviously redundant, but its documented definition seems to be what you want: it's to tell users whether a mesh is wholly contained in the x axis or the xy plane or neither.  Rather than trying to hang on to a user-defined dimension, we could test for non-zero y and z nodal coordinates in prepare_for_use().

I'd say wait for @jwpeterson or @pbauman or others to chime in before deciding on anything.

jwpeterson:
> On Dec 17, 2015, at 6:56 PM, roystgnr <notifications@github.com> wrote:
> 
> I fear this will break others' apps; ISTR some which rely on the current behavior of "reading a mesh in from file overrides the previously supplied mesh_dimension". It definitely breaks the API as defined, where mesh_dimension() returns the largest dimension of any element in the mesh.
> 
> If we wanted to make an API more clever (and we do; you've clearly got a reasonable use case), then spatial_dimension() would be the one to use IMHO. Its current implementation is obviously redundant, but its documented definition seems to be what you want: it's to tell users whether a mesh is wholly contained in the x axis or the xy plane or neither. Rather than trying to hang on to a user-defined dimension, we could test for non-zero y and z nodal coordinates in prepare_for_use().
> 
> I'd say wait for @jwpeterson or @pbauman or others to chime in before deciding on anything.
> 

I agree with Roy, the mesh's dimension() has been defined as "the largest element dimension contained in the mesh during prepare_for_use()" for quite a while, and was recently generalized with the elem_dimensions container.

Mesh::spatial_dimension() is pretty pointless at the moment... it should definitely return something mesh-specific instead of a compile time constant that is already available.

roystgnr:
If we use spatial_dimension() I think the only question I have is whether we want spatial_dimension()==2 to imply "your mesh is in the xy plane" or just "your mesh is coplanar"; likewise for "in the x axis" vs "collinear".  My vote would be "xy plane"/"x axis" since that's an easy optimization opportunity, but "coplanar"/"collinear" does feel like a more natural definition.

roystgnr:
Maybe I should harass @dknez too; IIRC he was the one who caught some bug we had with 2D meshes in the xz or yz plane...

dknez:
Yeah, there was a bug with the point locator for a mesh in the XZ plane. IIRC the fix was simple once we saw what was triggering it, though.

In this case, I don't have a strong opinion about spatial_dimension should return, as long as its documented.

Though I think spatial_dimension is a bit of a messy function because it requires a tolerance to decide whether elements are really co-planar or not... I would question whether we need this function at all?

roystgnr:
The "coplanar"/"collinear" definition would require a tolerance.  The "xz"/"x" definition wouldn't.

I would like to know exactly how @bwspenc wants the dimension result to be used, I admit.  If you have a dome truss and a user changes it to a flat truss, for example, how would the change in dimension from 3 to 2 change what you wanted your code to do?

pbauman:
Just my two cents, but I agree with what's been discussed here. This would break things in unexpected ways. It really sounds like @bwspenc wants spatial_dimension(), but I too am curious what he's trying to do.

friedmud:
@bwspenc I'm also interested in what you're up to.  Is it something where we can keep track of the dimension in MOOSE and not involve changing this in libMesh?

bwspenc:
I agree that what I want is the spatial dimension.  That could be determined by looking at nodal coordinates, probably with a tolerance. In the case of Exodus meshes, the dimension is supplied in the mesh metadata, and we should probably use that.

I still don't like the fact that the constructor for the mesh object takes a dimension argument that seems to just get ignored. It should either be used (this patch is one way to accomplish that), or removed. I don't care either way.

@friedmud we have multiple uses for 2d or 3d meshes that are composed entirely of line (truss) elements. We can use that for traditional truss or beam structural models, and we have also gotten crude implementations of the discrete element method and peridynamics working in MOOSE using such meshes.  To work around the issue of the dimension being incorrectly reported, we have been putting a dummy hex element in the mesh to make it get treated as a 3d mesh, but I'd like to make it so we don't have to do that. If you want to see what I'm talking about, there are a few regression tests in solid_mechanics/tests/solid_mech_truss that have dummy hex elements, and I have versions of those models without the hexes that break that we can use for testing a solution to this.

I'm sure this is something we could handle on the MOOSE side, but since libmesh owns the mesh, it makes sense to me that there should be some function on the libmesh side to tell us the spatial dimension of the mesh.

bwspenc:
Just to clarify, I only know of the following two reasons why this matters to me:
1) MOOSE has a check to make sure that the size of the displacement variable vector matches the mesh dimension, and it is calling mesh_dimension() to do that check. Maybe that should just be changed.
2) This also affects the dimension set in the metadata for output Exodus meshes. I don't think it's a huge deal if this is set wrong, but some postprocessors behave differently depending on how that is set. Paraview has a 2D mode, for instance. The output mesh should have the same dimension as the input mesh.

pbauman:
> we have multiple uses for 2d or 3d meshes that are composed entirely of line (truss) elements. 

I too have use cases for one and two-dimensional manifolds, but what I don't understand is where you're using the spatial dimension in the first place. All the nodal information is tracked in libMesh and the Jacobian of the element map is also taken care of automatically, so I would think you would be agnostic to the spatial dimension. For instance, in my cable and membrane elements, we never query the spatial_dimension() (other than an assert to make sure libMesh was built with LIBMESH_DIM > 1).

friedmud:
@bwspenc gotcha.  I agree that the `spatial_dimension()` is the right thing to modify here.

jwpeterson:
> I still don't like the fact that the constructor for the mesh object takes a dimension argument that seems to just get ignored. It should either be used (this patch is one way to accomplish that), or removed. I don't care either way.

@bwspenc and I discussed this a bit just now.  I think it's inaccurate to say that the dimension is "ignored", but rather that it can and will be updated based on the dimension of the elements in the mesh during `prepare_or_use()`.

This is partly a documentation issue (you have to read the documentation fairly carefully to understand the differences between the various "dimensions" that a Mesh can have) but it is also a backwards-compatibility issue: in the earliest versions of libMesh you *had* to specify the "dimension" of the Mesh at construction time, and I don't think it changed after that point.  So, we have this legacy still hanging around in the constructors, and maybe it would be a good idea to get rid of it now.  We already changed the constructors once during the `LIBMESH_DISABLE_COMMWORLD` stuff,  why not change them again?

roystgnr:
IIRC when we started auto-setting the mesh_dimension there was still one or two file format readers that couldn't do so, so you still had to set a consistent dimension in the constructor before reading one of those.  That might no longer be the case now that we support mixed dimensions and mesh_dimension is just a historical subset of the elem_dimensions data, but we should check to be sure.

jwpeterson:
> This also affects the dimension set in the metadata for output Exodus meshes. I don't think it's a huge deal if this is set wrong, but some postprocessors behave differently depending on how that is set. Paraview has a 2D mode, for instance. The output mesh should have the same dimension as the input mesh.

@bwspenc by default in MOOSE, meshes with `mesh_dimension()` == 2 or 3 produce Exodus files with `num_dim==Mesh::mesh_dimension()`.

You can force the "spatial" dimension to be used instead by setting `use_problem_dimension=false` in the Exodus block.  The flag name is a bit confusing... 

The following lines in Exodus.C describe this:
```
  if (isParamValid("use_problem_dimension"))
    _exodus_io_ptr->use_mesh_dimension_instead_of_spatial_dimension(getParam<bool>("use_problem_dimension"));
  else
  {
    // Utilize the spatial dimensions
    if (_es_ptr->get_mesh().mesh_dimension() != 1)
      _exodus_io_ptr->use_mesh_dimension_instead_of_spatial_dimension(true);
  }
```

jwpeterson:
One other design thing @bwspenc and I discussed before I forget: there may be some value to letting `spatial_dimension()` (or some other, differently-name interface) be something that is *specifiable by the user* rather than something automatically calculated by libMesh.

This would be used as a way for the application developer to tell libMesh e.g. "I know this initially planar 2D mesh is going to deform into the third dimension during my simulation".  One benefit would be that code which performs other initialization (setting up the number of displacement variables on the System, etc.) could rely on the value of `Mesh::spatial_dimension()` before the simulation has actually begun.

roystgnr:
Hmm... maybe initialize spatial_dimension() to 0 for an empty mesh, automatically calculate it in prepare_for_use(), but only *increase* the dimension, never decrease?  That way the user can bump it up manually but not worry about that getting reverted if the mesh is temporarily flattened.

friedmud:
Seems reasonable.

bwspenc:
@roystgnr I like that too.  I think there are cases when the user would want to bump it up, but never down.

jwpeterson:
Yeah, I guess there's a very weird scenario where your 3D mesh collapses down to *exactly* 2D for some reason and you accidentally write out an Exodus file with num_dim==3 when you actually wanted it to realize that num_dim==2, but that's not even really a bug, just potentially a little less convenient to visualize in Paraview.

roystgnr:
I could imagine a few corner cases.  Someone's doing adaptive modeling with something like shallow-water+navier-stokes, they run a case where the model coarsening strips the last 3D element from the mesh, and now it would be nice to optimize for all-2D.  But the user could handle anything like that manually, and since we've been fixed at LIBMESH_DIM all this time nobody's code will be surprised or broken when their now-2D mesh still reports that it's 3D.

Definitely need to document this, though.  I think the "prepare_for_use() can increment spatial_dimension()" behavior is the right way to go (so users can rely on spatial_dimension() without having to micromanage it) but we'll want to spell it out in mesh_base.h

Now we just need a consensus on "2D means coplanar" vs "2D means xy or yz or xz" vs "2D means xy".  My vote is still "xy".

jwpeterson:
> Now we just need a consensus on "2D means coplanar" vs "2D means xy or yz or xz" vs "2D means xy". My vote is still "xy".

Sounds fine to me.  And you are going to check this by looking at the z-coords of every Node and seeing if they are exactly-all-bytes-0-no-tolerance-check-no-really-i-mean-exactly-0?

bwspenc:
My vote is that it's 2d only if it's in the xy plane, and with no tolerance on z.

There's a store in the mountains here that sells "Square Ice Cream". Maybe they could shed some light on how to collapse things down from a higher-dimensional space.

roystgnr:
Yeah, no tolerance for z.

roystgnr:
And yeah, the only way to do things is a loop over every Node.  Should be cheap in the common 3D case, though, where you'll hit a non-xy node before you've passed N^2/3 entries and you can break out of the loop as soon as you do.

jwpeterson:
Replaced by #761.

PR #759, opened 2015-12-21
Fix issues with writing Exodus files after deleting nodes.

### LINKED ISSUE #753: Issue with node deletion, allow_renumbering(), and prepare_for_use()
While debugging something yesterday, @permcody came up with a simple test case that surprised me by not working.  The test code below produces the wrong mesh connectivity (see image).  We found three different ways of "fixing" the problem (lines 1, 2, and 3 in the comments below) but it seems like the code as written should "just work" :tm:.

Method 3 *may* cause it to work since it introduces a call to `renumber_nodes_and_elements()` that is otherwise avoided in `prepare_for_use()`... I haven't dug into it closely enough yet to know for sure.  One easy fix might be to always call `contract()` in `prepare_for_use()`.  It should be a fairly inexpensive operation if there are no NULL Elems in the Mesh.

```
#include "libmesh/mesh.h"
#include "libmesh/mesh_generation.h"
#include "libmesh/elem.h"
#include "libmesh/getpot.h"

using namespace libMesh;

// Test behavior of prepare_for_use() when elements are deleted.
{
  LibMeshInit init(argc, argv);

  // Create a GetPot object to parse the command line
  GetPot command_line (argc, argv);

  // Read the number of elements on a side from the command line.
  int N = 2;
  if (command_line.search(1, "-N"))
    N = command_line.next(N);

  Mesh mesh(init.comm());
  MeshTools::Generation::build_square (mesh,
                                       /*nx=*/N, /*ny=*/N,
                                       /*xmin=*/0, /*xmax=*/1.,
                                       /*ymin=*/0, /*ymax=*/1.,
                                       QUAD4);

  // Delete element 0 and node 0, disallow numbering, and call
  // prepare_for_use().  This results in the incorrect connectivity in
  // the resulting file.
  mesh.delete_elem(mesh.elem(0));
  mesh.delete_node(mesh.node_ptr(0)); // 1.) Comment out this line and it works fine!
  mesh.allow_renumbering(false);      // 2.) Set this to true and it works fine!
  // mesh.contract();                 // 3.) Uncomment this line and it works fine!
  mesh.prepare_for_use();
  mesh.write("test_contract.e");

  return 0;
}
```

connectivity

jwpeterson:
* mesh.n_nodes() does not return a reliable result when Nodes have been
  deleted but mesh.contract() has not yet been called.
* Must use `node_num_map` (or its `std::map` equivalent,
  "libmesh_node_num_to_exodus") to write element connectivity arrays
  when there are holes in the node numbering.

Closes #753.

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -1116,7 +1116,12 @@ void ExodusII_IO_Helper::initialize(std::string str_title, const MeshBase & mesh
   num_elem = mesh.n_elem();
 
   if (!use_discontinuous)
-    num_nodes = mesh.n_nodes();
+    {
+      // Don't rely on mesh.n_nodes() here.  If nodes have been
+      // deleted recently, it will be incorrect.
+      num_nodes = cast_int<int>(std::distance(mesh.nodes_begin(),
+                                              mesh.nodes_end()));
+    }
   else
     {
       MeshBase::const_element_iterator       it  = mesh.active_elements_begin();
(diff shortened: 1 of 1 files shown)


jwpeterson:
Another random failure on `utils/random.threads_verification_uo`

PR #761, opened 2015-12-21
Change MeshBase::spatial_dimension()

jwpeterson:
Initial attempt as per the discussion in #756.  I suspect this will need a couple iterations before it is exactly right, but this basically does what I think it is supposed to do.  Would be great if @roystgnr could comment, especially regarding the tests for x-aligned and planar x-y meshes.  I'll close #756 and put a pointer to this PR there.

### DIFF
--- a/src/mesh/nemesis_io_helper.C
+++ b/src/mesh/nemesis_io_helper.C
@@ -917,7 +917,7 @@ void Nemesis_IO_Helper::write_exodus_initialization_info(const MeshBase & pmesh,
 {
   // This follows the convention of Exodus: we always write out the mesh as LIBMESH_DIM-dimensional,
   // even if it is 2D...
-  this->num_dim = pmesh.spatial_dimension();
+  this->num_dim = LIBMESH_DIM;
 
   this->num_elem = static_cast<unsigned int>(std::distance (pmesh.active_local_elements_begin(),
                                                             pmesh.active_local_elements_end()));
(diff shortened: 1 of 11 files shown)


src/mesh/nemesis_io_helper.C
```diff
@@ -917,7 +917,7 @@ void Nemesis_IO_Helper::write_exodus_initialization_info(const MeshBase& pmesh,
 {
   // This follows the convention of Exodus: we always write out the mesh as LIBMESH_DIM-dimensional,
   // even if it is 2D...
-  this->num_dim = pmesh.spatial_dimension();
+  this->num_dim = LIBMESH_DIM;
```
roystgnr: Is this what we want in the long run, or just an attempt to keep exodiff tests from failing?

roystgnr:
I'd have said :+1: if MooseBuild hadn't disagreed...

src/mesh/nemesis_io_helper.C
jwpeterson: Definitely a short-term fix to maintain the original behavior.  Longer term, the right thing to do is let the user choose which dimension to use when writing the mesh, similarly to ExodusII_IO.

jwpeterson:
> I'd have said :+1: if MooseBuild hadn't disagreed...

Yeah, I thought this might happen.  The exodiffs are all of the "Dimension doesn't agree" variety, and I think they are all on 1D problems that are now being written with `num_dim=1` where before they were being written with `LIBMESH_DIM`.  I should be able to fix this with a MOOSE patch that doesn't require a libmesh update...

jwpeterson:
Closing this for now as I don't think I'll be able to merge it before first fixing something in MOOSE.  Will re-open when that is more squared away.

PR #762, opened 2015-12-31
Getpot stream-based constructor

pbauman:
This adds a new GetPot constructor using an `std::istream`. Idea is to be able to use a `std::stringstream` to supply the input to GetPot. This facilitates unit testing in applications as well as our own GetPot unit test by removing the need for the autoconf business to point to an input file path.


jwpeterson:
> On Dec 31, 2015, at 3:04 PM, Paul T. Bauman <notifications@github.com> wrote:
> 
> This adds a new GetPot constructor using an std::istream. Idea is to be able to use a std::stringstream to supply the input to GetPot. This facilitates unit testing in applications as well as our own GetPot unit test by removing the need for the autoconf business to point to an input file path.
> 

Haven't looked at the code, but +1 to the idea.

roystgnr:
Could we shave off a few lines by making parse_input_file create an ifstream and hand off to parse_input_stream?

pbauman:
> Could we shave off a few lines by making parse_input_file create an ifstream and hand off to parse_input_stream?

Yup. Went ahead and did that in 96d98f6.

roystgnr:
Interesting.  This killed "make check" for me (in pre-existing out-of-source build directories) until I manually removed the old getpot_test.C that configure had generated in the previous system - the getpot_test.C in the build dir took precedence over the one in the source dir, apparently.

I don't see any way such a problem could have been avoided...

Issue #763, opened 2016-01-02
Compilation error with --disable-second


salazardetroya:
In the [uniform_refinement_estimator.C](https://github.com/libMesh/libmesh/blob/master/src/error_estimation/uniform_refinement_estimator.C#L624) the variable ```H2seminormsq``` is declared inside the ```LIBMESH_ENABLE_SECOND_DERIVATIVES``` compilation flags and makes the compilation fail if we pass --disable-second. Should that ```H2``` summation enclosed with the same compilation flags or the declaration of ```H2seminormsq``` declared outside the flags?

roystgnr:
On Sat, 2 Jan 2016, salazardetroya wrote:

> In the uniform_refinement_estimatorC the variable H2seminormsq is declared inside the LIBMESH_ENABLE_SECOND_DERIVATIVES compilation flags and
> makes the compilation fail if we pass --disable-second Should that H2summation enclosed with the same compilation flags or the declaration
> ofH2seminormsq` declared outside the flags?

The summation should be enclosed.  Looks like there's a couple other
errors that have crept in since the last time we built with
--disable-second, though; I'll try to prepare a patch fixing them all.
---
Roy

jwpeterson:
> On Jan 2, 2016, at 4:39 PM, roystgnr <notifications@github.com> wrote:
> 
> 
> On Sat, 2 Jan 2016, salazardetroya wrote:
> 
> > In the uniform_refinement_estimatorC the variable H2seminormsq is declared inside the LIBMESH_ENABLE_SECOND_DERIVATIVES compilation flags and
> > makes the compilation fail if we pass --disable-second Should that H2summation enclosed with the same compilation flags or the declaration
> > ofH2seminormsq` declared outside the flags?
> 
> The summation should be enclosed. Looks like there's a couple other
> errors that have crept in since the last time we built with
> --disable-second, though; I'll try to prepare a patch fixing them all.

Thanks.  Also sounds like a good candidate for the 0.9.5 branch.

roystgnr:
@salazardetroya, let me know if #764 works for you?

salazardetroya:
It works. Thank you.

PR #764, opened 2016-01-03
--disable-second fixes

roystgnr:
This should fix the compilation problem in #763, and one other I noticed while replicating the first.  Also give a sane error message rather than a wrong result if anyone tries to use that code path with --disable-second anyway.


roystgnr:
@jwpeterson, you want to cherry-pick these onto 0.9.5?

jwpeterson:
> @jwpeterson, you want to cherry-pick these onto 0.9.5?

Sure.

PR #765, opened 2016-01-05
Write node unique ids to xda/xdr files

roystgnr:
This is just one commit on top of #719 - it's actually logically independent of the former, but I'd like them tested together and I'd like to merge that first regardless.

I'll be adding subsequent commits to test out some in-theory-might-work-now cases of overlapping nodes; once that's done then everything ought to be safe to merge.


roystgnr:
Obsoleted by #768

PR #766, opened 2016-01-05
Add space around pointer/reference declarations

jwpeterson:
AKA: what I did during Christmas vacation.

I don't really expect anyone to review this branch since it's way too large, but it unifies the various styles of declaring pointer and reference variables we had in the library in accordance with the [libmesh coding conventions](https://github.com/libMesh/libmesh/wiki/Source-code-formatting-conventions) @roystgnr and I recently discussed and have now added to the Wiki.


jwpeterson:
FYI, the non-whitespace changes on this branch can be obtained by checking it out and running:

```
git diff --word-diff=porcelain --word-diff-regex=[^[:space:]] upstream/master..HEAD | grep "^-\|^+" | grep -v "^---\|^+++"
```

The results of this command are below.  The changes are all pretty innocuous.  The +v, -V changes are due to something weird with the way `tags-query-replace` works in Emacs, but they are only in dummy function parameter names in header files and don't break anything.  The other diffs are either deleting comments, reindenting, fixing typos, or splitting multiple variable definitions onto separate lines.

```
-*
+pointer
-p
+* \date 2009
-*
+s
-V
+v
-V
+v
-V
+v
-V
+v
-V
+v
-V
+v
-V
+v
-V
+v
-V
+v
-V
+v
-V
+v
-V
+v
-V
+v
-V
+v
-V
+v
-V
+v
-V
+v
-V
+v
-//
-//
-void
-void
+*
-// No good for Solaris C++! - BSK
-  //  Parameters::Parameter<T>
-  //    *copy = new Parameters::Parameter<T>;
+!
-          //const Elem* elem = *el;
-,
+;
+  std::vector<Real>
-,
+;
+  DenseMatrix<T>
-,
+;
+  int
-    //     active_elem_iterator       elem_it (mesh.elements_begin());
-    //     const active_elem_iterator elem_end(mesh.elements_end());
-(
+=
-)
-(
+=
-)
-,
+;
+  char
-,
+;
+  char
-,
+;
+  struct token
-,
+;
+  struct token
-,
+;
+    struct token
-,
+;
+  PetscInt
-,
+;
+  PetscInt
-template UniquePtr<FEGenericBase<RealGradient> > FEMContext::build_new_fe
-  ( const FEGenericBase<RealGradient>*, const Point &, const Real ) const;
+template UniquePtr<FEGenericBase<RealGradient> >
+FEMContext::build_new_fe(const FEGenericBase<RealGradient>*,
+                         const Point &,
+                         const Real) const;
-// Optimize for the common case, where this node was part of the
-    // old solution.
-    //
-    // Be sure to handle cases where the variable wasn't defined on
-    // this node (due to changing subdomain support) or where the
-    // variable has no components on this node (due to Elem order
-    // exceeding FE order)
-    if (n.old_dof_object &&
-        n.old_dof_object->n_vars(sys.number()) &&
-        n.old_dof_object->n_comp(sys.number(), i))
-      {
-        const dof_id_type old_id =
-          n.old_dof_object->dof_number(sys.number(), i, 0);
-        STOP_LOG ("Number eval_at_node()", "OldSolutionValue");
-        return old_solution(old_id);
-      }
+// Optimize for the common case, where this node was part of the
+  // old solution.
+  //
+  // Be sure to handle cases where the variable wasn't defined on
+  // this node (due to changing subdomain support) or where the
+  // variable has no components on this node (due to Elem order
+  // exceeding FE order)
+  if (n.old_dof_object &&
+      n.old_dof_object->n_vars(sys.number()) &&
+      n.old_dof_object->n_comp(sys.number(), i))
+    {
+      const dof_id_type old_id =
+        n.old_dof_object->dof_number(sys.number(), i, 0);
+return old_solution(old_id);
+    }
+STOP_LOG ("Number eval_at_node()", "OldSolutionValue");
-// Optimize for the common case, where this node was part of the
-    // old solution.
-    //
-    // Be sure to handle cases where the variable wasn't defined on
-    // this node (due to changing subdomain support) or where the
-    // variable has no components on this node (due to Elem order
-    // exceeding FE order)
-    if (n.old_dof_object &&
-        n.old_dof_object->n_vars(sys.number()) &&
-        n.old_dof_object->n_comp(sys.number(), i))
-      {
-        Gradient g;
-        for (unsigned int d = 0; d != LIBMESH_DIM; ++d)
-          {
-            const dof_id_type old_id =
-              n.old_dof_object->dof_number(sys.number(), i, d+1);
-            g(d) = old_solution(old_id);
-          }
-        STOP_LOG ("Gradient eval_at_node()", "OldSolutionValue");
-        return g;
-      }
+// Optimize for the common case, where this node was part of the
+  // old solution.
+  //
+  // Be sure to handle cases where the variable wasn't defined on
+  // this node (due to changing subdomain support) or where the
+  // variable has no components on this node (due to Elem order
+  // exceeding FE order)
+  if (n.old_dof_object &&
+      n.old_dof_object->n_vars(sys.number()) &&
+      n.old_dof_object->n_comp(sys.number(), i))
+    {
+      Gradient g;
+      for (unsigned int d = 0; d != LIBMESH_DIM; ++d)
+        {
+          const dof_id_type old_id =
+            n.old_dof_object->dof_number(sys.number(), i, d+1);
+          g(d) = old_solution(old_id);
+        }
+return g;
+    }
+STOP_LOG ("Gradient eval_at_node()", "OldSolutionValue");
-void
```

jwpeterson:
Hmm... this error does not seem to be related to this branch, as I can get the same thing (consistently) with libmesh HEAD.

The error is an assertion in the METIS partitioner:
```
misc/no_exodiff_map.test: Assertion `csr_graph.vals.size() == std::max(graph_size,std::size_t(1))' failed.
misc/no_exodiff_map.test: csr_graph.vals.size() = 360
misc/no_exodiff_map.test: std::max(graph_size,std::size_t(1)) = 720
```
but I haven't dug into it further than that.

This error *doesn't* show up in #765, which is based on 70eca2c, so something between that commit and the current HEAD must be the culprit.

jwpeterson:
Hmm... that MOOSE test,
```
mpiexec -np 2 ../../../moose_test-dbg -i no_exodiff_map.i
```
doesn't work with 70eca2c either.

We also figured out we didn't see this failure on on the original PR because we don't run the tests in parallel in debug mode, so it's almost certain that it's never worked.  I've opened an issue for it over at idaholab/moose#6123.

@roystgnr, is it possible that #765 actually *fixes* some issue with the METIS partitioner in cases where there are overlapping nodes, and that's why your PR passed?  Maybe related to interior parent stuff?  This particular test has two 10x10 perfectly overlapping 2D grids ...

roystgnr:
Hmm... the unique commit in #765 shouldn't have affected partitioning, just XdrIO.  But #765 includes #719, and half of the point of #719 (originally the whole point, before I discovered how fubar'ed the unique_id code was) was to fix Hilbert sorting of collocated nodes and elements, and the two places that could show up are in restarts and in parallel runs.  Is it possible that we do a Hilbert pre-sort before generating the graph to hand METIS/ParMETIS?

roystgnr:
Yup.  There's a find_global_indices() right there in MetisPartitioner::_do_partition.  So while I can't be sure that #719 fixed that case, I can tell you that #719 was definitely *intended* to fix that case.  :-)

jwpeterson:
OK, that's great actually!  It also means that I can merge this unrelated PR.  In looking at the diffs on #719 and #765 I can definitely see they are going to at least trivially conflict with this one; I volunteer to fix any conflicts during the merge or rebase them for you when you're ready to merge...

PR #768, opened 2016-01-06
Write nodal unique_id values to xda/xdr files

roystgnr:
Increment the file version number to 0.9.6+ to signify the format change.

This rebases (and replaces) #765 

I need to add more test coverage before we merge.

This probably shouldn't be left in limbo too long, though; now that #719 has fixed slit mesh (and other collocated mesh) uses, we need to make sure there aren't any lingering bugs w.r.t. restarts with such meshes.

### DIFF
--- a/src/mesh/xdr_io.C
+++ b/src/mesh/xdr_io.C
@@ -128,8 +128,8 @@ XdrIO::XdrIO (MeshBase & mesh, const bool binary_in) :
 #else
   _write_unique_id    (false),
 #endif
-  _field_width        (4),   // In 0.7.0, all fields are 4 bytes, in 0.9.2 they can vary
-  _version            ("libMesh-0.9.2+"),
+  _field_width        (4),   // In 0.7.0, all fields are 4 bytes, in 0.9.2+ they can vary
+  _version            ("libMesh-0.9.6+"),
   _bc_file_name       ("n/a"),
   _partition_map_file ("n/a"),
   _subdomain_map_file ("n/a"),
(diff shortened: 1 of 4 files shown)


roystgnr:
So I guess the slit mesh restart case passes with SerialMesh?  ParallelMesh is still failing for me in that new unit test, though, so I'm leaving this marked Do Not Merge until I have time to track that down.

roystgnr:
The remaining bug with ParallelMesh is not a recent regression, but a deep flaw which has existed for years (but which was coincidentally identified on another mesh just weeks ago by Miguel A. Salazar de Troya!).  I've added a workaround so the test here will hit the new capability but won't exercise a known-broken use case.

roystgnr:
Link to the flaw discussion:

http://sourceforge.net/p/libmesh/mailman/libmesh-users/thread/D2BD6AB0.D2E%25salazardetro1%40llnl.gov/#msg34767594

http://sourceforge.net/p/libmesh/mailman/message/34779672/

dknez:
Does this mean we can do restarts with slit meshes now?

roystgnr:
In theory, yes: as long as you're configured with unique_id() support on (for both the write and the read), you can do restarts with slit meshes.

In practice: you can do restarts with the 1.5 slit meshes in the unit test.  After all the nasty corner cases I've fixed in the last week and a half, I'm unwilling to swear to any more than that.  Anyone who wants to add more regression tests would have my gratitude.

dknez:
OK, thanks, I'll test this out when I get some time.

PR #771, opened 2016-01-08
Change operator<<(pair<Hilbert,unique_id>) namespace

roystgnr:
This still works with my gcc version; hopefully this will fix the
problem Xujun Zhao encountered with XCode.

### DIFF
--- a/include/parallel/parallel_hilbert.h
+++ b/include/parallel/parallel_hilbert.h
@@ -75,9 +75,14 @@ public:
 } // namespace Parallel
 
 
+} // namespace libMesh
 
-// This has to be in libMesh namespace for Koenig lookup to work
+
+namespace Hilbert {
+
+// This has to be in the Hilbert namespace for Koenig lookup to work?
 // g++ doesn't find it if it's in the global namespace.
+// XCode didn't find it in the libMesh namespace.
 #ifdef LIBMESH_ENABLE_UNIQUE_ID
 inline
 std::ostream&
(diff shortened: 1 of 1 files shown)


jwpeterson:
I can confirm that this fix works for us under Xcode 7.2, and you don't even need to be using the compiler that comes with Xcode, the previous code fails even with our hand-built compiler somehow.  So I'm going to merge this.

jwpeterson:
Actually this fails on clang on Linux as well.  If you dig down into the output of the "Build libmesh" target [here](https://www.moosebuild.org/job/6703/) you should see:

```
In file included from ../src/mesh/mesh_communication_global_indices.C:24:
In file included from ./include/libmesh/libmesh_common.h:65:
./include/libmesh/ostream_proxy.h:116:16: error: call to function 'operator<<' that is neither visible in the template definition nor found by argument-dependent lookup
    (*_target) << in; return *this;
               ^
../src/mesh/mesh_communication_global_indices.C:617:66: note: in instantiation of function template specialization 'libMesh::BasicOStreamProxy<char, std::char_traits<char> >::operator<<<std::pair<Hilbert::HilbertIndices, unsigned long> >' requested here
                    *(Point *)(*nodej) << " has HilbertIndices " <<
                                                                 ^
./include/libmesh/parallel_hilbert.h:84:1: note: 'operator<<' should be declared prior to the call site or in namespace 'Hilbert'
operator <<
^
```

roystgnr:
"fails", or "failed"?

roystgnr:
I.e. is that an error message from before this merge or after?

jwpeterson:
Before.  This patch fixed the error for us so I merged it.

roystgnr:
Excellent; thanks!

PR #772, opened 2016-01-08
Skip writing NODEELEMs in ExodusII writer

dknez:
Simple PR to skip writing out NODEELEMs in Exodus output.

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -1357,7 +1357,15 @@ void ExodusII_IO_Helper::write_elements(const MeshBase & mesh, bool use_disconti
       //Use the first element in this block to get representative information.
       //Note that Exodus assumes all elements in a block are of the same type!
       //We are using that same assumption here!
-      const ExodusII_IO_Helper::Conversion conv = em.assign_conversion(mesh.elem(tmp_vec[0])->type());
+      ElemType elem_type = mesh.elem(tmp_vec[0])->type();
+      if(elem_type == NODEELEM)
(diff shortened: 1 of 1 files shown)


roystgnr:
I guess there's nothing we can slip into the element_equivalence_map that's appropriate for these?  So better to write out a mesh than to spit out an error.

However, won't this just get you right back to the error you were having earlier?  Instead of "Try to add an unattached node to the mesh, and watch it get auto-deleted" you'll have "Try to add an unattached node to the mesh, write it out to file, and watch it get auto-deleted if you ever read that file back in to a new mesh"

dknez:
Yes, you're right. The idea of this is that I would read in a mesh, then add NodeElems and solve, then write out the solution. This works fine. However, I do lose the NodeElems when the mesh or solution data is written out.

It would be better if we would write out NodeElems as part of the mesh. However, I'm not sure if that is supported by Exodus or any of the other mesh formats?

dknez:
Looks like Exodus has an element type SPHERE which is just a node? If that's right then I think it'd be good to add support for SPHEREs to our Exodus reader/writer, which would read/write NodeElems. That seems like a better fix than this PR. What do you think?

roystgnr:
Hmm... and give it some tiny arbitrary radius?  That sounds reasonable.

Perhaps use CIRCLE instead to try and avoid weirdness when outputting 2D meshes?

dknez:
Yeah, that sounds good to me.

dknez:
I've closed this since #773 is a better approach.

PR #773, opened 2016-01-09
Added support for reading and writing NodeElems in ExodusII

dknez:
We use the ExodusII SPHERE element to represent NodeElems.

### DIFF
--- a/include/mesh/exodusII_io_helper.h
+++ b/include/mesh/exodusII_io_helper.h
@@ -788,6 +788,16 @@ private:
 public:
 
   /**
+   * 0D node maps.  These define mappings from ExodusII-formatted
+   * element numberings.
+   */
+
+  /**
+   * The NodeElem node map.
+   */
+  static const int nodeelem_node_map[1];
+
+  /**
    * 1D node maps.  These define mappings from ExodusII-formatted
    * element numberings.
    */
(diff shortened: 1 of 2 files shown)


dknez:
This works as I hoped. We can add NodeElems to a mesh and they're preserved on writing to and reading from Exodus files. Also, paraview handles Exodus SPHERE data with no problems, so that's convenient too.

roystgnr:
:+1:, thanks!

Should we add read support too?  Exodus/Nemesis are lousy formats for restarts but it's nice to support what we can.

dknez:
What change do you have in mind regarding read support?

I did the following test:
- I did a solve, and wrote out the mesh and solution data via equation_systems.write("solution_data.exo");
- I then made a new Mesh and EquationSystems by reading in the mesh exo file and solution_data.exo file.

This all worked fine, and it retained the NodeElem in the solution after reading in solution_data.exo. So it seems to me that this already does everything that is necessary, no?

dknez:
Oh, oops, I thought equation_systems.write("solution_data.exo") would write an exodus file, but it writes an Xda file.

I'm not sure what code to change to make exodus restarts work?

roystgnr:
I don't think anything should need to be changed to make the equation_systems part of a restart work; IIRC Exodus reads assume isoparametric Lagrange data.

But we probably ought to fix mesh.read("mesh.exo") in the case where mesh.exo has SPHERE elements in it.

dknez:
> But we probably ought to fix mesh.read("mesh.exo") in the case where mesh.exo has SPHERE elements in it.

That already works, I've confirmed that with a test case.

roystgnr:
Ready to merge, then?

dknez:
Yep. I'll do it.

PR #775, opened 2016-01-12
Skip lower-dimensional elements in System::calculate_norm()

dknez:
System::calculate_norm() currently assumes that all elements have the same dimension, and hence fails on mixed-dimension meshes. A simple fix is to just ignore the lower-dimensional elements, since they have zero measure on a dim-dimensional mesh.

### DIFF
--- a/src/systems/system.C
+++ b/src/systems/system.C
@@ -1557,6 +1557,13 @@ Real System::calculate_norm(const NumericVector<Number> & v,
         {
           const Elem * elem = *el;
 
+          if(elem->dim() < dim)
+          {
+            // Skip lower-dimensional elements since they do
+            // not contribute to the norm
+            continue;
+          }
+
           fe->reinit (elem);
 
           this->get_dof_map().dof_indices (elem, dof_indices, var);


pbauman:
What's the code path for this? I thought I'd hit all these during testing way back when.

I don't think I agree with ignoring the lower dimensional elements. My solution lives on both manifolds (2-D and 1-D in my case) and the solution on both should contribute to the norm. For example, if I compute the norm of the residual and we ignore the 1-D manifold, I could have a zero residual norm and not satisfy the equations. But maybe I'm misunderstanding the use case?

 Perhaps something akin to what do we in [ExactSolution::_computer_error](https://github.com/libMesh/libmesh/blob/master/src/error_estimation/exact_solution.C#L607) would make sense?

Regardless, sorry for trouble.

dknez:
Thanks for checking, I wasn't sure this was the right change...

Specifically, I hit an error (due to failed quadrature) in a 3D mesh with that had a NodeElem in it, hence the mesh is mixed dimesion with dimensions 0 and 3.

I agree with your comments about not ignoring the lower-dimensional elements.

What fix would you recommend that deals with the NodeElem issue I ran into?

pbauman:
> Specifically, I hit an error (due to failed quadrature) in a 3D mesh that had a NodeElem in it, hence the mesh is mixed dimesion with dimensions 0 and 3.

Ah, OK, thanks.

> What fix would you recommend that deals with the NodeElem issue I ran into?

I would think we could follow what we did in the `compute_error` case, namely, build up  FE and quadrature objects for each of the dimensions present in the mesh (before the element loop), check the dimension of the current element inside the element loop and then use the appropriate FE and quadrature object. 

That said, in your particular case, you would actually *not* want that NodeElem as part of the norm, right? IIRC, it was just to ease the BC specification? I'm not sure what to do about that ATM.

pbauman:
Maybe add an optional argument to `calculate_norm` to take a `std::set*` of dims to ignore during the norm calcuation, with the default being NULL and therefore all dims included?

dknez:
>I would think we could follow what we did in the compute_error case, namely, build up FE and quadrature objects for each of the dimensions present in the mesh (before the element loop), check the dimension of the current element inside the element loop and then use the appropriate FE and quadrature object.

OK, that sounds good to me. I'll look into that.

>That said, in your particular case, you would actually not want that NodeElem as part of the norm, right? IIRC, it was just to ease the BC specification? I'm not sure what to do about that ATM.

Actually, it's fine if the NodeElem is part of the norm, so there's no problem there.

PR #776, opened 2016-01-12
System::calculate_norm() now works with mixed-dimension meshes

dknez:
This follows the approach used in ExactSolution::compute_error, where we initialize FE objects for each dimension and then pick out the right one on each element. We can also now skip specific dimensions by setting the skip_dimensions set (by default we calculate contributions from all dimensions in the mesh).

### DIFF
--- a/include/systems/system.h
+++ b/include/systems/system.h
@@ -1204,14 +1204,16 @@ public:
    */
   Real calculate_norm(const NumericVector<Number> & v,
                       unsigned int var,
-                      FEMNormType norm_type) const;
+                      FEMNormType norm_type,
+                      std::set<unsigned int> * skip_dimensions=NULL) const;
 
   /**
    * @returns a norm of the vector \p v, using \p component_norm and \p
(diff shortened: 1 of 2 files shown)


dknez:
@pbauman can you have a look at this? I guess this is what you had in mind?

include/systems/system.h
```diff
@@ -1204,14 +1204,16 @@ class System : public ReferenceCountedObject<System>,
    */
   Real calculate_norm(const NumericVector<Number> & v,
                       unsigned int var,
-                      FEMNormType norm_type) const;
+                      FEMNormType norm_type,
+                      std::set<unsigned int>* skip_dimensions=NULL) const;
```
jwpeterson: I'll just add a space around these pointer declarations before merging.

pbauman:
:+1: Thanks!

dknez:
@jwpeterson I fixed the whitespace around the pointers. I'll merge now.

PR #777, opened 2016-01-12
GetPot fixes when libmesh is absent

dmcdougall:
I fix two bugs here:

1.  The declaration of the mutex when libmesh is not present (`#undef USE_LIBMESH`)
2.  Using the libmesh-independent `getpot_file_error` macro instead of the `libmesh_file_error` macro

Feedback welcome.

### DIFF
--- a/include/base/getpot.h
+++ b/include/base/getpot.h
@@ -96,6 +96,7 @@ extern "C" {
 #else // !USE_LIBMESH
 
 // Currently threaded GetPot use is only supported via libMesh Threads
+#define GETPOT_DISABLE_MUTEX
 #define SCOPED_MUTEX
 #define GETPOT_MUTEX_DECLARE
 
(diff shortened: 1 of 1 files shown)


roystgnr:
Let's add a #define GETPOT_DISABLE_MUTEX in the non-libMesh case while we're at it?  No need to make user code responsible for that.

dmcdougall:
Yep.  Makes total sense.  Done.

roystgnr:
Thanks!  :+1: when it passes tests and others have had time to glance at it.

jwpeterson:
Seems fine to me...

jwpeterson:
@roystgnr This should probably be in 0.9.5 too?

roystgnr:
While we're at it, sure.  But these are bugs that only activate when someone strips the file out and undefines USE_LIBMESH, so not the typical use case.  :-)

dmcdougall:
> when someone strips the file out and undefines USE_LIBMESH...

¯\_(ツ)_/¯

I call it "code reuse"  :)

PR #783, opened 2016-01-14
Avoid using double leading underscore preprocessor macro names.

jwpeterson:
This wasn't actually hurting anything, but technically those names are
reserved for the implementation.  For more information, see:

http://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html

### DIFF
--- a/include/base/auto_ptr.h
+++ b/include/base/auto_ptr.h
@@ -268,7 +268,7 @@ public:
       {
         libMesh::warned_about_auto_ptr = true;
         libMesh::out << "*** Warning, AutoPtr is deprecated and will be removed in a future library version! "
-                     << __FILE__ << ", line " << __LINE__ << ", compiled " << __LIBMESH_DATE__ << " at " << __LIBMESH_TIME__ << " ***" << std::endl;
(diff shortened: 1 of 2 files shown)


include/base/auto_ptr.h
```diff
@@ -268,7 +268,7 @@ class AutoPtr : public safe_bool<AutoPtr<Tp> >
       {
         libMesh::warned_about_auto_ptr = true;
         libMesh::out << "*** Warning, AutoPtr is deprecated and will be removed in a future library version! "
-                     << __FILE__ << ", line " << __LINE__ << ", compiled " << __LIBMESH_DATE__ << " at " << __LIBMESH_TIME__ << " ***" << std::endl;
```
roystgnr: Hey, wait a minute; why aren't we just doing a plain libmesh_deprecated() here?

roystgnr:
Thanks!  Just because something doesn't hurt with the compilers we use doesn't mean we should be risking [nasal demons](http://www.catb.org/jargon/html/N/nasal-demons.html) when we don't have to.

include/base/auto_ptr.h
jwpeterson: Hmm... maybe because the message pre-dates the macro?  I'll change that...

PR #784, opened 2016-01-14
Make meshplot solution file optional

roystgnr:
And remove a couple overzealous asserts that got in the way.

This is only tested with ExodusII; if it breaks with other formats that'll be a separate bug to fix.


jwpeterson:
This seems good to me, although I might change the `libmesh_error` into a `libmesh_error_msg` with the associated print statement?

PR #787, opened 2016-01-18
Fix memory leak in vector_fe_ex{2,3,4}

jwpeterson:
`ExactSolution::attach_exact_{values,derivs}` can also take a constant reference to a vector, since it copies the objects internally.


pbauman:
Sorry for the sloppiness and thanks for fixing that.

Issue #790, opened 2016-01-19
mesh_dimension in build_sphere


BalticPinguin:
[mesh_gen.patch.txt](https://github.com/libMesh/libmesh/files/95857/mesh_gen.patch.txt)
I think there is a bug in build_sphere() concerning the dimension. 
Starting from miscellaneous_ex6 and changeing
 MeshTools::Generation::build_cube(...)
to
const libMesh::Real radius = 2;
MeshTools::Generation::build_sphere(cube_mesh, radius, 2, HEX8, 2, true);
I always get the error "Unknown dimension 0".
I think it is due to the mesh.clear() in that function where the dimension is removed as well.
At least for me, the patch in supplement made it working for me. 
Best,
Hubert

jwpeterson:
Thanks for sending the patch.  I'll take a closer look/merge it today hopefully.

PR #791, opened 2016-01-19
Fix dimension==0 issue in build_sphere after clearing mesh.

jwpeterson:
Also, handle the case where the user did not provide a valid
mesh_dimension at Mesh construction time (in which case it defaults to
1).  In that case, let the ElemType, which can be explicitly requested
by the user, determine the type of geometric entity (line, circle,
ball, sphere) which gets meshed.

Refs #790.  Thanks @BalticPinguin for reporting the issue.


roystgnr:
:+1:

PR #792, opened 2016-01-19
Fix obscure issue with errno_t not being defined.

jwpeterson:
This seems to currently only affect the following very specific configuration:
.) OSX with command line tools installed.
.) C++11 disabled.
.) GCC == 5.2.0.

The exact error message was:
In file included from /opt/moose/gcc_5.2.0/include/c++/5.2.0/cstring:42:0,
                 from /opt/moose/tbb/include/tbb/tbb_allocator.h:36,
                 from /opt/moose/tbb/include/tbb/tbb_exception.h:123,
                 from /opt/moose/tbb/include/tbb/parallel_for.h:28,
                 from ./include/libmesh/threads.h:31,
                 from ./include/libmesh/reference_counter.h:25,
                 from ./include/libmesh/reference_counted_object.h:24,
                 from ./include/libmesh/dof_object.h:28,
                 from ./include/libmesh/elem.h:25,
                 from ../src/mesh/exodusII_io_helper.C:31:
/usr/include/string.h:145:1: error: ‘errno_t’ does not name a type
 errno_t memset_s(void *, rsize_t, int, rsize_t) __OSX_AVAILABLE_STARTING(__MAC_10_9, __IPHONE_7_0);

### DIFF
--- a/include/parallel/threads.h
+++ b/include/parallel/threads.h
@@ -23,6 +23,16 @@
 #include "libmesh/libmesh_config.h"
 #include "libmesh/libmesh_common.h"  // for libmesh_assert
 
+// We ran across a very strange issue with hand-built GCC 5.2.0 on OSX
+// with C++11 disabled, in which the "errno_t" type was not defined.
+// This is an attempt to fix that issue without unduly affecting other
+// compilers and build configurations.
+#if !defined(LIBMESH_HAVE_CXX11) && defined(__APPLE__) && defined(__GNUC__)
+# if (__GNUC__ == 5 && __GNUC_MINOR__ >= 2 && __GNUC_PATCHLEVEL__ >= 0)
+typedef int errno_t;
+# endif
+#endif
+
 // Threading building blocks includes
 #ifdef LIBMESH_HAVE_TBB_API
 #  include "libmesh/libmesh_logging.h" // only mess with the perflog if we are really multithreaded


roystgnr:
It looks like what we actually want to do is s/errno_t/int

The C++2003 standard says errno is a macro defined in a header "the same as the Standard C library header \<errno.h\>"; and the C99 standard says errno "expands to a modifiable lvalue that has type int".

errno_t appears to be a Microsoft thing that got adopted elsewhere because "sigh, let's just make that code work".

roystgnr:
The latest C++ draft I have handy now refers to POSIX for the errno.h definition, but the POSIX definition still says "errno evaluates to an int lvalue" and doesn't say anything about errno_t.

permcody:
As much as we'd like to eradicate errno_t, it still appears in system header files and probably several compiler distributed includes as well. I don't think we have much choice but to handle these oddball cases as they arise.

jwpeterson:
> It looks like what we actually want to do is s/errno_t/int

Wait, do you mean in my patch, or in string.h like @permcody suggested?

roystgnr:
Wait, I didn't read this slowly enough; I just saw a definition of errno_t and tried to figure out whether we should be using it or not.  But we're *not* using it, it's just that the *system* *headers* are buggy?

Okay, at that point there is no good solution.  An autoconf test would be better than a compiler version check, though.  Imagine if Apple starts using "#define errno_t int"; then "typedef int errno_t;" becomes a syntax error.

jwpeterson:
> Wait, I didn't read this slowly enough; I just saw a definition of errno_t and tried to figure out whether we should be using it or not. But we're not using it, it's just that the system headers are buggy?

Yeah, I think it's a consequence of GCC picking up a (possibly old, or non-standard) header from Xcode in /usr/include on OSX.  Clang doesn't fail, so it must either be using different headers or always wrapping the inclusion of string.h, or ignoring --disable-cxx11, or some combination of the above.

>  An autoconf test would be better than a compiler version check, though. 

You're right.  I think I'm going to still use this patch in the meantime to prepare a libmesh update in MOOSE, and I'll work on a more general configure test later today once I get the GCC stack up and running on my machine (I initially tested this patch on a different machine).

jwpeterson:
Through trial and error, I've narrowed the error down to the following minimal program:

```
namespace Foo {
#include <errno.h>
}

#include <cstring>

int main()
{
  return 0;
}
```

We actually have this sequence of includes in exodusII_io_helper.C, which is why that file originally triggered the error.  The error goes away if:

1. You remove the namespace.
1. You remove either (or both, obviously) includes.

In the real code, the Foo namespace is actually libMesh, and <errno.h> is included by netcdf.h via exodusII.h.  I wonder if it we can just move this inclusion outside the libMesh namespace and have it work...

PR #793, opened 2016-01-20
Make FEMSystem numerical jacobians more robust

roystgnr:
We were previously failing in cases where the user's element_residual
failed to provide analytical jacobians, their side_residual provided
them, and libMesh was not compiled in dbg mode.

This fixes the bug for me, and the GRINS test suite should hit this code path pretty hard, but I'd also like to wait for @vikramvgarg to double-check and @pbauman to take a look before merging.

### DIFF
--- a/src/solvers/newmark_solver.C
+++ b/src/solvers/newmark_solver.C
@@ -211,8 +211,6 @@ bool NewmarkSolver::_general_residual (bool request_jacobian,
   // We might need to save the old jacobian in case one of our physics
   // terms later is unable to update it analytically.
   DenseMatrix<Number> old_elem_jacobian(n_dofs, n_dofs);
-  if (request_jacobian)
-    old_elem_jacobian.swap(context.get_elem_jacobian());
 
   // Local velocity at old time step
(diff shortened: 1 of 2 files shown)


pbauman:
Holy hell. Thanks for fixing that. I can't believe it took us thing long to trip it...

roystgnr:
"Fixing"?  Apparently not yet.  One step forward, two steps back.

roystgnr:
That's the right fix for libMesh (our bug was disguising a different bug!).  There's either one regression or one red herring cropping up in a GRINS test; if I can fix the former or be sure it's the latter then I'll commit right away.

@vikramvgarg figured out why we didn't trip this earlier: the bug only cropped up in cases where your element_residual returned false (indicating a need for numerical jacobians) but your side_residual returned true (indicating that it had added its jacobian contribution analytically).  That's always supposed to have been a supported use case, but in practice we never tended to use it; if we used numerical jacobians we used them everywhere.  Switching from penalty to Dirichlet boundary conditions was the trigger in his code - once he outright removed his side_foo() overrides they defaulted to returning (trivially 0) analytic jacobians.

PR #794, opened 2016-01-20
Add configure test for errno_t issue on OSX with GCC 5.2.0

jwpeterson:
This fixes the compilation issues we were experiencing yesterday on OSX, as well as a few GCC warnings (mostly about "overlength strings") that I came across while testing.


roystgnr:
:+1:, thanks!

jwpeterson:
Hmm... so I think this failed because the spatial_dimension branch needs to be merged first.  I forgot that MOOSE obviously pulls the latest libmesh when doing libmesh PR tests...

PR #795, opened 2016-01-20
Change MeshBase::spatial_dimension()

jwpeterson:
This should (hopefully!) pass now, since MOOSE has already been updated to a version of libmesh (on a branch) which includes these changes.

(See also discussion in #761.)


pbauman:
No complaints from me. :+1:

jwpeterson:
Urgh, we both added new unit tests.  I'll fix that by hand.

PR #796, opened 2016-01-20
Set SLEPc solver options for eigenproblems_ex3

roystgnr:
Without these, I get a "No support for this operation with this object
type" error when running in parallel with PETSc 3.6


jwpeterson:
This failed for the same reason #794 did, I can invalidate it after the spatial_dimension branch is merged, or you can just merge it as-is, since all the libmesh tests passed.

roystgnr:
I'll merge as-is; this doesn't touch anything outside of examples/ dir.

PR #799, opened 2016-01-21
Packed range refactor, take 2

jwpeterson:
Rebased version of #748.

Should work with MOOSE (as of the version I just merged to devel) according to @permcody.


jwpeterson:
Looks like I jumped the gun.  Civet pulls from MOOSE master, but @permcody's commit is not there quite yet.  I'll invalidate and try again in a bit.

roystgnr:
Great; thanks!

PR #800, opened 2016-01-21
Rereoptimize projections

roystgnr:
Hopefully this will fix the last of the performance regressions well enough; it adds special cases for the rest of the LAGRANGE dofs, which ought to be a decent help in 3D.

This needs to get more testing before merge.

### DIFF
--- a/src/systems/system_projection.C
+++ b/src/systems/system_projection.C
@@ -1337,8 +1337,8 @@ void GenericProjector<FFunctor, GFunctor, FValue, ProjectionAction>::operator()
           START_LOG ("project_edges","GenericProjector");
 
           // In 3D, handle any edge values next.  Optimize (at a
-          // slight cost of accuracy) the LAGRANGE case by using
-          // interpolation instead of L2 projection.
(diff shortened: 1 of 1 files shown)


roystgnr:
This is working with libMesh, MOOSE, and GRINS now, at least.

jwpeterson:
Let me try a libmesh update PR over in MOOSE, let this run through the various apps?

roystgnr:
Very good; I won't merge until we see the results of that.

jwpeterson:
Testing now at idaholab/moose#6236.

roystgnr:
For reference, this extends the changes in #696 to try to re-optimize the refactoring from #680.

roystgnr:
There's a nasty failure in idaholab/moose#6236, but I can replicate it with libMesh versions predating this patch, so I think we're clear here.

roystgnr:
Note for the future: I've thought of one more way to speed up the LAGRANGE code path, but that can go on the back burner indefinitely.  I don't expect it to be a significant improvement, but I do expect to eventually be intolerably tempted to create a "Rerereoptimize projections" PR.

PR #802, opened 2016-01-21
packed_range output_type fixes

roystgnr:
Without MPI enabled, there was a discrepancy between T* and const T*
in an output_type declaration.

Since output_type is solely there to allow template selection, it
should be const everywhere.


jwpeterson:
:+1:

PR #803, opened 2016-01-21
Slit mesh restarts may fail without unique_id

roystgnr:
And *do* fail on this particular mesh, according to @jwpeterson

So let's only enable this test when we know it should succeed.

### DIFF
--- a/tests/mesh/slit_mesh_test.C
+++ b/tests/mesh/slit_mesh_test.C
@@ -26,7 +26,7 @@ public:
 
   ~SlitFunc () {}
 
-  virtual void init_context (const FEMContext &) {}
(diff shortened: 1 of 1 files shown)


jwpeterson:
I'll try this on my machine; moosebuild is always testing with unique ids enabled.

jwpeterson:
Would you mind also fixing this warning:
```
../../tests/mesh/slit_mesh_test.C:29:16: warning: 'init_context' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
  virtual void init_context (const FEMContext &) {}
```

jwpeterson:
Other than that, this gets the unit tests passing on my system again.

PR #805, opened 2016-01-26
Add libmesh_nullptr macro, tests

jwpeterson:
I guess this won't get tested for a few days, but I wanted to get it out there nonetheless.  This adds backwards-compatible support for the C++11 [nullptr](http://stackoverflow.com/questions/13816385/what-are-the-advantages-of-using-nullptr) keyword.

Refs #655


roystgnr:
Looks good, but I'm not sure it gives us much benefit.

The nice thing about nullptr is that it's an actual pointer type, not a freaking int, so it behaves the way you'd expect in overloaded functions.  But we haven't been dumb enough to overload any functions with both pointer and int versions, have we?  And even if we have, and we try to pass it libmesh_nullptr, we'd just end up getting the right overload with C++11 enabled and the wrong overload without.

roystgnr:
:+1: anyway for encouraging good practice

jwpeterson:
> Looks good, but I'm not sure it gives us much benefit.

Right, I'd say the main benefit is that it looks like we actually know what we are doing with regard to C++11.

> The nice thing about nullptr is that it's an actual pointer type, not a freaking int, so it behaves the way you'd expect in overloaded functions. But we haven't been dumb enough to overload any functions with both pointer and int versions, have we? 

Nope.

> And even if we have, and we try to pass it libmesh_nullptr, we'd just end up getting the right overload with C++11 enabled and the wrong overload without.

Actually the C++03 workaround is designed to make this work as well.  You'd write your overloaded function to take a "nullptr_type", which would correspond to `std::nullptr_t` on C++11 compilers and `libmesh_nullptr_t` on C++03 compilers.  The current branch doesn't define "nullptr_type" since we don't have any overloads of this sort, but it could in the future.

PR #806, opened 2016-01-26
Removing pthread_unique_id() from Threads

### LINKED ISSUE #801: pthreads may be broken in libMesh
OK, I've spent a few hours trying to nail down some bad threading behavior and I still don't understand enough to solve the problem but I'll report what I think I've learned. I started running "helgrind" on MOOSE and it repeatedly prints out an issue with a recursive mutex on one that doesn't support it here:
https://github.com/libMesh/libmesh/blob/master/src/parallel/threads.C#L40

I started digging to see why it may be getting called and it looks like it's "first" locked here:
https://github.com/libMesh/libmesh/blob/master/include/parallel/threads.h#L552-L554

There are several odd things happening here, I may as well start with the first glaring issue. According to the POSIX standard. "Thread identifiers should be considered opaque: any attempt to use a thread ID other than in pthreads calls is nonportable and can lead to unspecified results."
http://man7.org/linux/man-pages/man3/pthread_self.3.html

So even though the type pthread_t is defined on our platforms, using it in a map probably isn't the best idea (but we do it anyway):
https://github.com/libMesh/libmesh/blob/master/src/parallel/threads.C#L32

Assuming that it works for a given platform (maybe), we move onto what's happening with this odd "recursive mutex" error. Again, according to the POSIX standard: "Unless real-time scheduling policies are being employed, after a call to pthread_create(), it is indeterminate which thread, the caller or the new thread will next execute."
http://man7.org/linux/man-pages/man3/pthread_create.3.html

The order in which things happen are this:
 1. lock the `_pthread_unique_id_mutex`
 2. create the thread
   1. thread: might ask for its id which takes us back to `Threads.C`
   2. thread: lock the `_pthread_unique_id_mutex` again  (possibly the source of the helgrind error)
   3. thread: return the thread id from the map which of course violates the POSIX recommendation (also, the bracket operators are being used here so we may also have an accidental insertion into the map if the id doesn't exist for some reason, or worse yet if the comparison operator is unreliable?).
 3. record the thread's unique id by using the buffer space as a key in the map (yikes)

This at least helps me to understand why this code was written the way it was. If a user desires a thread id for thread identification purposes and possibly indexing (i.e. everything in MOOSE) then we'd need to make sure we have a lock on the map before inserting into it it. Also, we'd need to make sure nobody else had a lock before accessing the thread id too. Logically, these two locks seem to make sense to me...

There are too many problems to fully understand the cause of the helgrind error. I could use another pair of eyes on this. I think the best way to resolve this would be to engineer a way to pass the thread id as an argument when we create the thread. In that world we wouldn't need any locks at all and we wouldn't have all this crazy logic to create a thread id through this weird mapping scheme. Since we have all these layers of abstraction anyway, it should be possible to reserve the first location in the threads args vector to hold the integer id. We'd then peel that off before passing control to the user thread routine.

permcody:
This method is buggy and unused by libMesh
it should simply be removed. I have a patch in MOOSE (https://github.com/idaholab/moose/pull/6246) that will remove the need to call this method.

Additionally, I believe we should cleanup these thread defines. We have at least four that I can see but there are already a few places where we have "else" clauses floating around without testing very many of these defines. For now, I thought I'd be more explicit about what case we are talking about. In the future we should consider a more robust model that allows us to turn on several configurations at the same time and choose which configuration to use at runtime!

closes #801

### DIFF
--- a/include/parallel/threads.h
+++ b/include/parallel/threads.h
@@ -433,15 +433,6 @@ private:
   pthread_mutexattr_t attr;
 };
 
-extern std::map<pthread_t, unsigned int> _pthread_unique_ids;
-extern spin_mutex _pthread_unique_id_mutex;
-
-/**
- * When called by a thread this will return a unique number from 0 to num_pthreads-1
- * Very useful for creating long-lived thread local storage
- */
-unsigned int pthread_unique_id();
-
 template <typename Range>
 unsigned int num_pthreads(Range & range)
 {
@@ -548,14 +539,12 @@ void parallel_for (const Range & range, const Body & body)
     {
 #if LIBMESH_HAVE_OPENMP
       run_body<Range, Body>((void *)&range_bodies[i]);
-#else // Just use Pthreads
-      spin_mutex::scoped_lock lock(_pthread_unique_id_mutex);
+#elif LIBMESH_HAVE_PTHREAD
       pthread_create(&threads[i], NULL, &run_body<Range, Body>, (void *)&range_bodies[i]);
-      _pthread_unique_ids[threads[i]] = i;
 #endif
     }
 
-#if !LIBMESH_HAVE_OPENMP
+#if LIBMESH_HAVE_PTHREAD
   // Wait for them to finish
 
   // The use of 'int' instead of unsigned for the iteration variable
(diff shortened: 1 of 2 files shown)


roystgnr:
:+1: as far as I'm concerned.

Also, runtime threading model selection would be *awesome*.  The same sort of factory pattern as we use for linear algebra would be nice, but we'd want to keep the old APIs around (albeit libmesh_deprecated()) for quite a while.

friedmud:
While Runtime selection sounds awesome... it's going to be tough. We use
compile time logic to completely switch out the TYPES of objects like
locks, etc... because sometimes they are TBB and sometimes not, etc.

I don't really see a way to do it currently.

PR #807, opened 2016-01-27
Fixed memory leak introduced in PR #773

dknez:


roystgnr:
Introduced in PR #776, you mean?

Anyway, the checks here are redundant.  Calling delete on a null pointer is guaranteed to have no effect.

dknez:
Oh yeah, #776, sorry.

PR #808, opened 2016-01-27
We were storing coarse vectors (including the adjoint) before solving…

vikramvgarg:
…, now solving before saving.

Passes adjoints ex4, which is the only test this should affect.


roystgnr:
:+1: and I'm planning to merge today.  If anyone wants to be paranoid and wait for the full regression suite scream now, but IIRC we don't have any ARefEE test coverage in MOOSE or GRINS.

pbauman:
> but IIRC we don't have any ARefEE test coverage in MOOSE or GRINS

That's correct, but I'll be dropping in a new test today that should help alleviate this problem.

PR #809, opened 2016-01-28
(Reduced Basis change) Optimized the EIM construction code

dknez:
Before we did multiple L2 projections for all variables in one big system. Now we do a separate L2 projection for each variable. This gives the same results as before but uses less memory.


dknez:
This only affects the reduced basis code, so I'll go ahead and merge.

PR #811, opened 2016-01-28
Add API to Detect Time-Dependence in FunctionBase Objects

pbauman:
The use case I have in mind here is to be able to query the `FunctionBase` in a `DirichletBoundary` object to see if we need to `reinit_constraints` or not. `FunctionBase` is assumed to be `f(x,y,z,t)` so there already is a notion of time. I've added functionality for `ParsedFunction` (which can just look for variable "t" in the expression), `CompositeFunction` (look at the subfunctions and see if any are time-dependent), and `ConstFunction` (never time-dependent).

I took a crack at `AnalyticFunction`, but I'm not sure how `init()` and `clear()` are typically used. I at least put a crack at the changes as the last commit to show what I was thinking. Can update it accordingly based on discussion. And `WrappedFunction` seems to not even assume time-dependence? I wasn't really sure what the "right thing" was to do there.

Labeling as do not merge until there's some discussion/consensus. I realize there may not even be interest in having this, but it was quick to do, so I just did it to throw it out there for thoughts.

### DIFF
--- a/include/numerics/composite_function.h
+++ b/include/numerics/composite_function.h
@@ -75,6 +75,20 @@ public:
         reverse_index_map[index_map[j]] =
           std::make_pair(subfunction_index, j);
       }
+
+    // Now check for time dependence
+    // We only check the function we just added instead of researching all subfunctions
+    // If this is the first subfunction, then that determines the time-dependence.
+    if( subfunctions.size() == 1 )
+      this->_is_time_dependent = subfunctions[0]->is_time_dependent();
+
+    // Otherwise, we have more than 1 function already.
+    // If _is_time_dependent is true, then one of the previous
+    // subfunctions is time-dependent and thus this CompositeFunction
+    // time-dependent. If _is_time_dependent is false, then the subfunction
+    // just added determines the time-dependence.
+    else if( !this->_is_time_dependent )
+      this->_is_time_dependent = (subfunctions.back())->is_time_dependent();
   }
 
   virtual Output operator() (const Point & p,
--- a/include/numerics/const_function.h
+++ b/include/numerics/const_function.h
@@ -34,7 +34,8 @@ class ConstFunction : public FunctionBase<Output>
 {
 public:
   explicit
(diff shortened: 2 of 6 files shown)


jwpeterson:
Seems reasonable to me.  Do you eventually intend to use this interface in the library for determining when functions need to be re-evaluated?  Or did I just miss that in the current set of diffs?

pbauman:
> Do you eventually intend to use this interface in the library for determining when functions need to be re-evaluated? 

That is definitely possible, but I have not searched the library for such opportunities yet. I could see this being part of the transient AMR+DirichletBoundary fix.

> Or did I just miss that in the current set of diffs?

There's nothing about that here in this PR. If there are things that spring to your mind, I'm happy to drop them in.

roystgnr:
Looks nice.  I presume we don't need to bother adding such an API to FEMFunctionBase, since even if your functional doesn't change with t your solution will?

pbauman:
>  I presume we don't need to bother adding such an API to FEMFunctionBase, since even if your functional doesn't change with t your solution will?

This was my thinking, but it would be easy enough to add if others have a use for it.

Anyone have thoughts on the `AnalyticFunction` and `WrappedFunction` cases?

roystgnr:
You could add a setter to FunctionBase and then get rid of the changes to AnalyticFunction; people who wanted to specify time-independence could then do so right after construction.  Alternatively you could go the AnalyticFunction route with WrappedFunction too: add an optional constructor argument which defaults to time-dependent.

pbauman:
> You could add a setter to FunctionBase and then get rid of the changes to AnalyticFunction; people who wanted to specify time-independence could then do so right after construction.

I'd guessed that solution would be hated. Since you suggested, that's my preferred route because it's much smaller changes. :) I'll drop cda7027 and then squash in a commit to 0defc72 that adds the setter and document that it is meant to only be used by subclasses that can't natively determine time-dependence.

pbauman:
Done. This is ready to go as far as I'm concerned once checks are happy. Thanks!

PR #812, opened 2016-01-28
Correct adjoint rhs values in process_constraints

roystgnr:
In theory we could have had bad adjoint solution values at hanging nodes next to adjoint dirichlet boundaries.

Pinging @vikramvgarg to see whether this might have ever come up in practice.


roystgnr:
Vikram's been helping me test this; it seems broken.

pbauman:
Check the test case I just added in grinsfem/grins#348. I'm pretty sure this manifested there.

roystgnr:
The problems I saw here were finally fixed in #2504

PR #815, opened 2016-02-01
Add -Wno-unknown-pragmas to GCC flags.

jwpeterson:
Otherwise, every file emits the following two warnings from threads.h:
```
../include/libmesh/threads.h:539:0: warning: ignoring #pragma omp parallel [-Wunknown-pragmas]
../include/libmesh/threads.h:640:0: warning: ignoring #pragma omp parallel [-Wunknown-pragmas]
```

I also tried disabling the warnings "in place" with yet more pragmas:
```
#pragma GCC diagnostic ignored "-Wunknown-pragmas"
...
#pragma GCC diagnostic warning "-Wunknown-pragmas"
```
but this seemed to have no effect, only adding the compiler flag (and it has to appear *after* `-Wall` in the list of flags) made the warning go away.

### DIFF
--- a/configure
+++ b/configure
@@ -7438,11 +7438,17 @@ fi
 
     # GCC 4.6.3 warns about variadic macros but supports them just
     # fine, so let's turn off that warning.
+    # GCC also warns about an "unknown pragma" on every file that
+    # includes threads.h when you configure with --disable-openmp.
+    # This is really annoying if you know what you are doing, and
+    # really don't want openmp enabled.  It's also a little unexpected
+    # for a pragma, presumably the compiler silently ignores unknown
+    # ones all the time...
     case "$GXX_VERSION" in
       gcc4.6 | gcc5)
-        CXXFLAGS_OPT="$CXXFLAGS_OPT -Wno-variadic-macros"
-        CXXFLAGS_DEVEL="$CXXFLAGS_DEVEL -Wno-variadic-macros"
-        CXXFLAGS_DBG="$CXXFLAGS_DBG -Wno-variadic-macros"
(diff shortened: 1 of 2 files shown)


pbauman:
Even with `-fopenmp` that happens? Otherwise, I'd suggest we `#ifdef LIBMESH_HAVE_OPENMP` (or whatever) around those. However, I've never poked my head in `threads.h` so I could be way off base.

permcody:
I agree with Paul, why not just put our existing #ifdefs around those
statements?

jwpeterson:
The warnings only pop up *without* -fopenmp.  I suppose we could ifdef the #pragmas, but that just feels... redundant somehow.  Also in case any more omp pragmas were added, you'd also have to remember to #ifdef those, etc.

pbauman:
> The warnings only pop up without -fopenmp. I suppose we could ifdef the #pragmas, but that just feels... redundant somehow. Also in case any more omp pragmas were added, you'd also have to remember to #ifdef those, etc.

It may be redundant, but I'd rather `#ifdef` around it and still be able to get warnings than the other way around. Just my two cents, maybe everyone else disagrees.

jwpeterson:
Are you using other #pragmas?  There aren't any in the library except the omp ones and the ones that disable warnings...

jwpeterson:
Replaced by #816.

pbauman:
> Are you using other #pragmas? There aren't any in the library except the omp ones and the ones that disable warnings...

I was thinking that if downstream users were sucking up libMesh flags, we'd be disabling that warning for them. Sorry for being a PITA.

PR #818, opened 2016-02-03
Absolute values for error_per_cell in the elementwise adjoint refinement error estimator

vikramvgarg:
We werent putting absolute values on the error per cells from the adjoint refinement estimator earlier. Compiles and fixes the problem I was having in GRINS because of this.


pbauman:
:+1:

roystgnr:
Fixes https://github.com/grinsfem/grins/issues/346 ?  Looks great to me; thanks!

PR #820, opened 2016-02-03
Consolidate LAGRANGE projection code

roystgnr:
This doesn't seem to fix the regression case we've found, but it does
make the code a little faster and a lot simpler.


jwpeterson:
We killed the optimized mode PR test manually, looks like it hung "make check".

roystgnr:
Thanks!  Looking into the debug failure now.

jwpeterson:
Also, :+1: for branch name alone.

PR #821, opened 2016-02-04
Bugfix for DofMap::constrain_element_matrix

rblake:
DofMap::constrain_element_matrix (the 4 argument asymmetric version) was
returning incorrect element matrices when the row_dofs have a constraint and
the col_dofs do not.

The initial if condition in the routine was too stringent and missed the XOR case.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -1755,18 +1755,26 @@ void DofMap::constrain_element_matrix (DenseMatrix<Number> & matrix,
   row_dofs = orig_row_dofs;
   col_dofs = orig_col_dofs;
 
+  // K_constrained = R^T K C
+
+  if ((R.m() == matrix.m()) &&
+      (R.n() == row_dofs.size()))
+    {
+      matrix.left_multiply_transpose  (R);
+    }
+
+  if ((C.m() == matrix.n()) &&
+      (C.n() == col_dofs.size()))
+    {
+      matrix.right_multiply (C);
+    }
 
   // It is possible that the matrix is not constrained at all.
   if ((R.m() == matrix.m()) &&
       (R.n() == row_dofs.size()) &&
       (C.m() == matrix.n()) &&
-      (C.n() == col_dofs.size())) // If the matrix is constrained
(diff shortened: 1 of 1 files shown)


rblake:
I might need someone with more experience to look at this:

https://github.com/rblake/libmesh/commit/89a01a5a6a04b41f2480525b4970ca7391b39784

To my naive reading of the source, it seems this version would do better when R!=NULL and C==NULL.  This version fails the tests though.  Is it a problem with the tests?

In any case, the current version is bugged.  I discovered this while trying to introduce incompressibility through Lagrangian multipliers for hydrostatic pressure.  Things broke when trying to insert the (pressure/displacement) sub matrices wherever I had a dirichlet BC for displacement.

rblake:
Posted on the mailing list:

http://sourceforge.net/p/libmesh/mailman/message/34823135/

jwpeterson:
I'll wait for @roystgnr to comment since the constraints code is his, but I'll just mention that this breaks all of the periodic BCs tests in MOOSE as well as some of the adaptivity tests, which makes sense since they all use the constraint system.

Can you maybe give some more details of your application or even post the source code?  We have an example (systems_of_equations_ex3) that uses a Lagrange multiplier to constrain the average value of the pressure in an incompressible flow solver that might be related...

jwpeterson:
Oh, just saw your email over on the mailing list, nevermind that second paragraph...

roystgnr:
Actually, IIRC this part of the constraints code is Ben's...

I apologize to @rblake for being incommunicado for so long; I was getting beaten down by other work.

This certainly looks like a real bug.  Unfortunately the test failures *also* look real, not just "we generated a gold standard with a buggy version" false positives.  The first part of this patch looks correct, but I'm not certain about the second.  I'll check now to see if I can reproduce the test failure with this PR, and then if I can I'll try one change and see if my hunch fixes it.

roystgnr:
I can't seem to actually replicate the failures here...

roystgnr:
Not sure what I did differently, but I can replicate the failures on a different system.

roystgnr:
Fixed by #835

PR #824, opened 2016-02-05
Added OrderWrapper shim class to enable higher FEType order

dknez:
As per the discussion with @roystgnr and @jwpeterson on the libmesh-devel list.

### DIFF
--- a/include/base/variable.h
+++ b/include/base/variable.h
@@ -118,7 +118,7 @@ public:
    * The number of components of this variable.
    */
   unsigned int n_components() const
-  { return type().family == SCALAR ? _type.order : 1; }
+  { return type().family == SCALAR ? _type.order.get_order() : 1; }
 
   /**
    * \p returns \p true if this variable is active on subdomain \p sid,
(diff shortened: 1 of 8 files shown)


dknez:
hmm, this builds for me. Might be because I'm using a C++11-compatible compiler.

dknez:
Oh, the reason it compiled for me before is that I configure with infinite elements on, and I forgot to modify the infinite elements code path...

roystgnr:
Get rid of the first default argument in the old FEType constructor

dknez:
Yep, doing that now.

dknez:
hmm, any suggestions on what to do about errors like this:
`ambiguous overload for ‘operator==’ (operand types are ‘libMesh::OrderWrapper’ and ‘libMesh::Order’`

I tried adding an operator== for OrderWrapper and Order, but that gave ambiguous overloads related to Order and int.

dknez:
This doesn't seem to be as straightforward as I'd hoped. We can just drop this PR if you like. Let me know what you think?

roystgnr:
Go ahead and leave it; I won't have time to play with it for a while but I think we're almost there.

roystgnr:
Superceded by #866

PR #825, opened 2016-02-07
Add triple_product()

jwpeterson:
This is useful in some Elem::volume() formulas, and might be in other places as well.

### DIFF
--- a/include/numerics/type_vector.h
+++ b/include/numerics/type_vector.h
@@ -977,6 +977,31 @@ bool TypeVector<T>::operator != (const TypeVector<T> & rhs) const
   return (!(*this == rhs));
 }
 
+
+//------------------------------------------------------
+// Non-member functions on TypeVectors
+
+// Compute a * (b.cross(c)) without creating a temporary
+// for the cross product.  Equivalent to the determinant
+// of the 3x3 tensor:
+// [a0, a1, a2]
+// [b0, b1, b2]
+// [c0, c1, c2]
+template <typename T>
+inline
+T triple_product(const TypeVector<T> & a,
+                 const TypeVector<T> & b,
+                 const TypeVector<T> & c)
+{
+  // We only support cross products when LIBMESH_DIM==3, same goes for this.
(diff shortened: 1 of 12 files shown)


roystgnr:
:+1: - what's the motivation, though?  Trying to improve readability or trying to optimize?  Compiler ought to be able to do the latter.

jwpeterson:
> what's the motivation, though? Trying to improve readability or trying to optimize? Compiler ought to be able to do the latter.

The main motivation was to optimize, but I think it may also be more readable.

I generated assembly (clang -O2 -S) for the following test code with one of the lines defining `tp` uncommented:
```
#include "libmesh/point.h"
using namespace libMesh;
int main()
{
  Point a(random(),1,1), b(random(),2,2), c(random(),3,3);
  Real tp = a * b.cross(c); // 1
  // Real tp = triple_product(a, b, c); // 2
  printf("%f\n", tp);
  return 0;
}
```

`diff test1.s test2.s` produces the following:
```
6a7,8
> LCPI0_1:
> 	.quad	-4611686018427387904    ## double -2
43,48c45
< 	cvtsi2sdq	%rax, %xmm0
< 	mulsd	LCPI0_0(%rip), %xmm2
< 	addsd	%xmm0, %xmm0
< 	movapd	%xmm0, %xmm3
< 	subsd	%xmm2, %xmm3
< 	subsd	%xmm0, %xmm2
---
> 	cvtsi2sdq	%rax, %xmm3
50a48,51
> 	mulsd	LCPI0_0(%rip), %xmm2
> 	mulsd	LCPI0_1(%rip), %xmm3
> 	addsd	%xmm2, %xmm3
> 	subsd	%xmm3, %xmm0
52d52
< 	addsd	%xmm2, %xmm0
```

I definitely don't know much about assembly, but they are at least different, and version 1 has "movapd", which [google](https://en.wikipedia.org/wiki/MOVAPD) tells me involves copying numbers to temporary space in the processor for use in other computations...

PR #833, opened 2016-02-18
Try copying old dofs whenever possible

### LINKED ISSUE #831: Fix new projection code, if possible
I've finally figured out why the bug here is so hard to trigger a regression failure with.  (instructions for triggering one via MOOSE are in https://github.com/idaholab/moose/pull/6236 )

The MOOSE failure simply manages to magnify the problem: a constant monomial field, written out to the restart file with a value of 2 on each element, is getting read back in with a value of 2-epsilon (1.9999999999999996) on a handful of those elements.  Obviously we're not going to notice a 1e-15 scale error on any numerics tests, but these particular values are being used as *refinement flags*, and are being cast to an enum, and in C++ casting a floating point number to an integer (or a thinly disguised integer) truncates, not rounds.  Dropping the pre-cast value by 1e-15 drops the post-cast value by 1, which changes the flag from "refine" to "do_nothing".

Honestly, though, I'm not sure how this worked reliably before.  We've always done projection for non-lagrange elements, and that's always got a risk of floating point error.  I'll submit a PR to make MOOSE more robust once I'm done seeing if I can also get that last epsilon back on the libMesh side.

roystgnr:
"Whenever possible" should be "quite a lot" whenever most mesh
elements haven't been flagged, and before we were only copying old
dofs for vertices and Lagrange nodes even in those cases.

This will hopefully fix #831 

I'm out of time for the day, so this hasn't yet been tested.  Or compiled.  Sorry for abusing the CI.

### DIFF
--- a/src/systems/system_projection.C
+++ b/src/systems/system_projection.C
@@ -240,8 +240,11 @@ public:
                        unsigned int var,
                        unsigned int dof_index)
   {
+    if (!this->check_old_context(c, c.get_elem().point(0)))
+      libmesh_error();
+
     const std::vector<dof_id_type> & old_dof_indices =
-      c.get_old_dof_indices(var);
+      old_context.get_dof_indices(var);
 
     libmesh_assert_greater(old_dof_indices.size(), dof_index);
 
--- a/src/systems/system_projection.C
+++ b/src/systems/system_projection.C
@@ -163,6 +163,13 @@ public:
                         const Real time)
   { return _f->component(c, i, n, time); }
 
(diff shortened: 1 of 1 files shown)


jwpeterson:
> Sorry for abusing the CI.

Pssh, only one failed push.  That's not abuse, try getting on @dschwen's level...

roystgnr:
This seems to fix the restart error for me.  We might be able to get MOOSE back onto mainline libMesh now...

jwpeterson:
:+1: This would make me very happy!

PR #835, opened 2016-02-22
Nonsquare constraint bugfix

roystgnr:
This attempts to fix both the problem that @rblake solved in #821 and the regressions that a bug in his solution created.

MOOSE tests all pass for me now.  Hopefully Civet agrees; otherwise we'll need a bugfix for the bug in the bugfix for the bug in the bugfix.


jwpeterson:
Thanks!

rblake:
Thanks Roy! It seems obvious in retrospect.

PR #836, opened 2016-02-23
adding MITC4 example

svallaghe:
Since libMesh was missing some "traditional" shell elements, here is an example of an implementation of the MITC4 element that is one of the most popular shell elements in commercial finite element software.


jwpeterson:
On Tue, Feb 23, 2016 at 4:03 AM, svallaghe <notifications@github.com> wrote:

> Since libMesh was missing some "traditional" shell elements, here is an
> example of an implementation of the MITC4 element that is one of the most
> popular shell elements in commercial finite element software.
>
Thanks!  I'll take a closer, I'm curious how you implemented a new element
entirely in an example...

-- 
John

examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
```diff
@@ -0,0 +1,751 @@
...
+  UniquePtr<FEBase> fe (FEBase::build(dim, fe_type));
+  QGauss qrule (dim, fe_type.default_quadrature_order());
+  fe->attach_quadrature_rule (&qrule);
+
+  // The element Jacobian * quadrature weight at each integration point.
+  const std::vector<Real> & JxW = fe->get_JxW();
+
+  // The element shape function and its derivatives evaluated at the
+  // quadrature points.
+  const auto & dxyzdxi = fe->get_dxyzdxi();
```
jwpeterson: Sorry, still no C++11 code allowed in examples for the time being.  (I will fix any small stuff like this, no need to make extra commits.)

svallaghe:
MITC4 is mostly just a QUAD4 first order Lagrange element (bilinear shape functions).
Only the shear part of the strain is interpolated differently (constant on the edges and linear between opposite edges), but it is simple enough that you can write it explicitly.

examples/miscellaneous/miscellaneous_ex12/run.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+#set -x
+
+source $LIBMESH_DIR/examples/run_common.sh
+
+example_name=miscellaneous_ex12
+
+options="-pc_type lu -pc_factor_mat_solver_package mumps"
```
jwpeterson: Do you know if this example works without a direct solver and/or without PETSc?  We'll have to programmatically restrict this example to PETSc-enabled builds otherwise.

svallaghe: In my experience shell problems are quite sensitive and a direct solver is almost mandatory.
Any direct solver should work though.

jwpeterson: OK, good to know.  By the way, it seems like you are fairly experienced in using Eigen; I wonder if they have any direct solvers available that we might add interfaces for in the `EigenSparseLinearSolver`.  Not necessarily for this PR, more of a note to self for future work.

svallaghe: According to their doc they do: http://eigen.tuxfamily.org/dox-devel/group__TopicSparseSystems.html.
In this example Eigen classes were helpful for doing lots of small matrix-vector arithmetic.

jwpeterson:
Replaced by #837.

PR #837, opened 2016-02-25
Add MITC4 example from @svallaghe to build system.

jwpeterson:
I found that, while this example did not work with EigenSparseLinearSolver (`--use-eigen`) for whatever reason, it worked just fine with Jacobi+CG, so there's no reason to require special solver packages or restrict the example to running in serial because of the need for direct solvers.

Also removed some C++11 only code, and verified that it still compiled with `--disable-cxx11`.

Replaces #836.


examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
```diff
@@ -0,0 +1,750 @@
...
+#include "libmesh/dense_vector.h"
+#include "libmesh/dense_subvector.h"
+#include "libmesh/sparse_matrix.h"
+#include "libmesh/numeric_vector.h"
+#include "libmesh/exodusII_io.h"
+#include "libmesh/dirichlet_boundaries.h"
+#include "libmesh/zero_function.h"
+
+//Eigen includes
+#include <Eigen/Dense>
```
roystgnr: This (and hence the whole example) needs to be wrapped in ifdef LIBMESH_HAVE_EIGEN

jwpeterson:
Good point, will add a commit for that now.

jwpeterson:
Tested c5a969f and it works (compiles, runs, and prints the "Eigen required" message) for me with Eigen disabled.  Since this configuration is not tested by the CI, I'm just going to merge it.

Issue #838, opened 2016-02-25
Investigate logic issues in threads.h


jwpeterson:
This is tricky since, as @permcody pointed out, we can have multiple threading models enabled simultaneously.  Imagine a case where both `LIBMESH_HAVE_OPENMP` and `LIBMESH_HAVE_PTHREAD` are defined.  Then, the following code from threads.h will call `pthread_join()` while never having called `pthread_create()`...

```
#ifdef LIBMESH_HAVE_OPENMP
#pragma omp parallel for schedule (static)
#endif
  for(unsigned int i=0; i<n_threads; i++)
    {
#if LIBMESH_HAVE_OPENMP
      run_body<Range, Body>((void *)&range_bodies[i]);
#elif LIBMESH_HAVE_PTHREAD
      pthread_create(&threads[i], libmesh_nullptr, &run_body<Range, Body>, (void *)&range_bodies[i]);
#endif
    }

#if LIBMESH_HAVE_PTHREAD
    for (int i=0; i<static_cast<int>(n_threads); i++)
      pthread_join(threads[i], libmesh_nullptr);
#endif
```

I suggest that we only allow a single threading model to be active at a time.  If the user requests a specific one, we give them that (if available) otherwise there is a specific hierarchy that we use to choose *the* threading model that is active, and disable the others.

friedmud:
Yep.  @permcody broke it.

Right here:  https://github.com/libMesh/libmesh/commit/bca60c6aa668b84b47003106f31823375a3206b2#diff-da1d293dd8ccb6bcfd350add22c3e363L558

friedmud:
Oh - but I do agree that this is really a symptom of the underlying issue that we allow for multiple models to be turned on... but there is an implicit preference hierarchy that's built into the code tbb>openmp>pthreads.  That hierarchy should really be moved to configure time and enforced.

friedmud:
While we're at it we should make sure the "pure serial" case works (i.e. without ANY threading model!).

jwpeterson:
Agreed.

permcody:
Well it was already incorrect. I was just maintaining status quo. ;)

friedmud:
How was it incorrect before?  It had `#if !LIBMESH_HAVE_OPENMP` which is exactly right.

[...]

PR #839, opened 2016-02-26
Fix in Threads::parallel_for().

jwpeterson:
This reverts a regression introduced by bca60c6aa, but is not the
complete overhaul of threads.h that is discussed in #838. It fixes
a definite issue with parallel_for() when both OpenMP and pthreads are
active at the same time.

### DIFF
--- a/include/parallel/threads.h
+++ b/include/parallel/threads.h
@@ -549,7 +549,7 @@ void parallel_for (const Range & range, const Body & body)
 #endif
     }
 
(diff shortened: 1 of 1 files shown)


friedmud:
@jwpeterson You missed one!

https://github.com/libMesh/libmesh/blob/5f783cbcb79e6ddc8870b068608aacee88889fe8/include/parallel/threads.h#L664

permcody:
Ah, there it is

PR #840, opened 2016-02-26
Add SPARSELU SolverType for Eigen.

jwpeterson:
This sparse direct solver works with the new misc_ex12, and lets us
enable that example when PETSc is not available.  It is also faster
than using the Eigen CG solver, or PETSc JACOBI+CG (see table of solve
times for misc_ex12 below).  I also tested Eigen's SparseQR
implementation, but it was much slower than SparseLU, and doesn't
really buy us much flexibility over that method.

```
Solver                             | solve() time
--------------------------------------------------
PETSc -pc_type lu                  | 0.0450
Eigen SparseLU                     | 0.0774
PETSc -ksp_type cg -pc_type jacobi | 0.3371
Eigen CG                           | 0.5971
Eigen SparseQR                     | 6.2501
```


roystgnr:
Nice!  :+1:

PR #841, opened 2016-02-26
add locking to PetscVector::_get_array

friedmud:
Uses a doubly-checked scoped lock to efficiently protect against a race condition on setting the `_values` pointer.

fixes idaholab/moose#6436

### DIFF
--- a/src/numerics/petsc_vector.C
+++ b/src/numerics/petsc_vector.C
@@ -34,7 +34,8 @@
 namespace libMesh
 {
 
-
+// Mutex for Petsc vector thread safety
+Threads::spin_mutex petsc_vector_mutex;
 
 //-----------------------------------------------------------------------
 // PetscVector members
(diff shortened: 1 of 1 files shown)


roystgnr:
Is there a higher-level race condition here?  Or can we trust/require that "we won't have one thread trying to get the array while a different thread tries to restore it" will always be the case in user code?

Should we make that mutex a member variable rather than a global?  Right now we typically work on one vector at a time, but if in the future we go more asynchronous there's no reason to lock one vector because of work being done on another.

friedmud:
I don't think we really need to worry about restore() and definitely not
about threads calling restore and get() simultaneously (but, isn't that
covered by using the same mutex in both places like I did? I can't see how
it would cause a problem the way it is.)

I can see making it a member variable. Let me know of you want it that way
and I'll make the changes.

jwpeterson:
I think this is probably OK as-is, and it's a serious enough bug that I'd like to make a "libmesh update" PR over in MOOSE that uses this branch and see if it makes the TBB target any more stable.  We can invalidate & re-run it a couple times to make sure it's working...

PR #842, opened 2016-02-28
Complete the threading fix

friedmud:
Complete the threading fix started in #839 caused by #838

### DIFF
--- a/include/parallel/threads.h
+++ b/include/parallel/threads.h
@@ -661,7 +661,7 @@ void parallel_reduce (const Range & range, Body & body)
 #endif
     }
 
-#if LIBMESH_HAVE_PTHREAD
+#if !LIBMESH_HAVE_OPENMP
   // Wait for them to finish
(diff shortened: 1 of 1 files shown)


dschwen:
This looks weird. Why would you not ifdef a pthreads function with a pthreads ifdef? Just curious.

friedmud:
Look further up, it's already in and ifdef pthreads block.  Note: this is actually just reverting to what was there before @permcody fixed something else and decided to change these at the same time (most likely because it didn't "look right" to him either :-), which broke this.

friedmud:
@dschwen a little more: OpenMP and Pthreads are comingled.  The current implementation actually has the OpenMP capability depending on PThreads for some things.  Because of this, they are wound together in the code inside `ifdef pthread` blocks... with the pthread bits separated off by `if !openmp`

dschwen:
Ok, this probably has to do with the precedence "logic" :-)

PR #843, opened 2016-02-29
Don't build FEMContext quadrature based on SCALAR order or family

roystgnr:
This should fix the bug that @jwpeterson caught, as well as an as-yet-untriggered bug I noticed while investigating.


jwpeterson:
:+1:

PR #844, opened 2016-02-29
Make --disable-maintainer-mode the default

jwpeterson:
As discussed [here](https://sourceforge.net/p/libmesh/mailman/message/34877760/) we came to the consensus that we will try things this way for a little bit.  This means that developers will have to explicitly configure with `--enable-maintainer-mode` to ensure that changes to the build system correctly trigger updates to the auto-generated build files, but "normal" users hopefully won't get weird configure errors as their first experience using the library.


jwpeterson:
BTW, I've tested this on a fresh clone, and everything works without first running bootstrap or passing `--disable-maintainer-mode` to configure, so that aspect at least is working...

PR #846, opened 2016-03-01
Add print_helper for double vector (#845)

snschune:
@permcody @jwpeterson  
This relates to https://github.com/idaholab/moose/pull/6459

### DIFF
--- a/include/utils/parameters.h
+++ b/include/utils/parameters.h
@@ -40,13 +40,16 @@
 namespace libMesh
 {
 /**
- * Helper functions for printing scalar and vector types.  Called from Parameters::Parameter<T>::print(...).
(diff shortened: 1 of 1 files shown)


include/utils/parameters.h
```diff
@@ -40,7 +40,7 @@
 namespace libMesh
 {
 /**
- * Helper functions for printing scalar and vector types.  Called from Parameters::Parameter<T>::print(...).
+ * Helper functions for printing scalar, vector and double vector types.  Called from Parameters::Parameter<T>::print(...).
```
jwpeterson: I got confused by the "double" terminology in this comment and the comment below, until I realized you meant "2D array".  Other than that, the code looks fine.

snschune: True enough, I'll rename it it vector<vector> types.

PR #847, opened 2016-03-02
Overhaul of threads.h, add support for --with-thread-model

jwpeterson:
This PR is designed to hopefully make it easier to understand threads.h by moving all the interleaved #ifdefs into separate files, and forcing only one threading "model" to be active at a time.  We currently support two real threading models and one fallback model:

* TBB
* pthreads
* none

OpenMP is also still supported, but is not a full model in the same way that the others are.  The current implementation is the same as the old one: OpenMP pragmas are used in the `parallel_for()` and `parallel_reduce()` implementations of pthreads, and are not used in the TBB model at all.  I've tested this locally, and it works with MOOSE in all three modes, but it is a pretty big change so it would be good if others could take a look/comment as well...

Refs #838.

### DIFF
--- a/configure
+++ b/configure
@@ -36108,6 +36108,13 @@ $as_echo "$ac_cv_tls" >&6; }
       libmesh_optional_LIBS="$TBB_LIBRARY $libmesh_optional_LIBS"
       found_thread_model=tbb
     fi
+
+    # If TBB was not enabled, but the user requested it, we treat that as an error:
+    # we want to alert the user as soon as possible that their requested thread model
+    # could not be configured correctly.
+    if (test $enabletbb = no -a "x$requested_thread_model" = "xtbb"); then
+      as_fn_error $? "requested threading model, TBB, could not be found." "$LINENO" 5
+    fi
   fi
 
   # If TBB wasn't selected, try pthreads as long as the user requested it (or auto)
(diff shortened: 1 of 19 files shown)


jwpeterson:
One comment I have gotten here: if a user asks for a specific thread model, e.g. `--with-thread-model=tbb`, and it's *not* found (or the configure test fails for some reason), configure currently continues on with thread model == "none".

We could instead have this situation trigger an `AC_MSG_ERROR`, the idea being to prevent the user from getting much farther down the line before they realize they've screwed up their build.  The default, `--with-thread-model=auto` would not trigger such an error.

friedmud:
I know this is just a (good) first step but I would really like to see OpenMP separated from Pthreads... even if there is currently just duplication.

Ultimately, all of the OpenMP structures need to get rewritten using OpenMP directives/structures (like OpenMP locks and atomics, etc.).  I was just being lazy when I first did OpenMP and decided that I could just use the existing stuff I had done for Pthreads... but it's not really right.

jwpeterson:
> On Mar 4, 2016, at 10:00 AM, Derek Gaston <notifications@github.com> wrote:
> 
> I know this is just a (good) first step but I would really like to see OpenMP separated from Pthreads... even if there is currently just duplication.
> 
> Ultimately, all of the OpenMP structures need to get rewritten using OpenMP directives/structures (like OpenMP locks and atomics, etc.). I was just being lazy when I first did OpenMP and decided that I could just use the existing stuff I had done for Pthreads... but it's not really right.
> 

I didn't realize there were any types in OpenMP at all... I thought it was all just pragmas wrapping "normal" code.

friedmud:
There are some types (like `omp_lock_t`)... but what I mean is that we can
use OpenMP statements to implement our `Threads` types.

Like `omp atomic` ( https://software.intel.com/en-us/node/432628 ) for
`Threads::atomic`

and `omp_set_lock()` and `omp_unset_lock()` for `Threads::spin_mutex` and
`scoped_lock`

and `omp_set_nest_lock()` and `omp_unset_nest_lock()` for
`Threads::recursive_mutex`

jwpeterson:
I'm planning to eventually make OpenMP a first-class threading model like the others, but this PR is already big enough as-is, and I'd like to get it merged in time for the next libmesh update in MOOSE.

PR #851, opened 2016-03-04
P projection fix, HP refinement on ParallelMesh fixes

roystgnr:
This gets "make check" running adaptivity_ex3 in p and hp refinement modes, and fixes some bugs (a project_vector regression and some ParallelMesh incompatibilities) revealed thereby.


roystgnr:
Here's a type of C++11 double checked locking that wasn't in my first link and that includes a non-singleton example.  Looks like manually specifying memory ordering might be more efficient than just using std::atomic?

https://www.justsoftwaresolutions.co.uk/threading/multithreading-in-c++0x-part-6-double-checked-locking.html

PR #853, opened 2016-03-06
Add a bit of delay to PetscVector::_get_array

friedmud:
Ok - look... I'm not crazy... and this DOES work.  I ran hundreds of thousands of tests to prove it.

The deal is that we need to delay setting `_array_is_present = true` until the values of `_first` and `_last` have had time to propagate to the other processors.  I brainstormed some ideas with people around here and this is the best we could come up with... making `_array_is_present` depend on the values of complex functions called with `_first` and `_last` as arguments will ensure that `_array_is_present` could never get set to `true` before there has been plenty of time for the memory subsystem to propagate those values.

This looks like it would slow things down... but actually, this only happens on the first access out of thousands/millions... so it's a fairly rare event and we're only adding a few flops to that one line.  It should definitely be faster than doing any locking or most likely even memory fences.

Let's have a healthy discussion about this.  I know it looks like a hack... but it works.

Later, when we switch to C++11 we can investigate adding memory fences here as a permanent/true fix.  Without C++11 the only alternatives would be INCREDIBLY slow.

refs idaholab/moose#6436

### DIFF
--- a/src/numerics/petsc_vector.C
+++ b/src/numerics/petsc_vector.C
@@ -1368,7 +1368,11 @@ void PetscVector<T>::_get_array() const
             _last = static_cast<numeric_index_type>(petsc_last);
           }
 
-          _array_is_present = true;
+	  // Ok, I know this is funky.  What we're trying to do is delay
+	  // setting _array_is_present until the values of _first and _last
+	  // have been proagated to the other processors when doing threading.
+	  // See the discussion here: https://github.com/idaholab/moose/issues/6436
+          _array_is_present = std::cos(_first) + std::sin(_last) + 10;
         }
     }
 }


roystgnr:
:-1: unless I'm misunderstanding this.  The solution to a race condition isn't to hobble one of the racers.  That may fix us completely on one architecture and then break us horribly on the next generation of the same architecture, or it may turn a one-in-a-million bug that we still have a hope of fixing into a one-in-a-billion-bug that will never be triggered except in large-scale production runs.

I like your line of thinking in #854, assuming we give the user a bit of libmesh_deprecated() time before forcing the manual cache control on them.

friedmud:
@roystgnr This erases the race condition.  The race was about setting the value of `_last` and `_first` vs setting the value of `_array_is_present`.  By making `_array_is_present` depend on the value of the other two it removes the possibility of a race condition.

friedmud:
BTW: we have to do something here.  Locking the whole thing is not an option.  I've done a lot of reading on this issue and this is the best I can come up with (and I've verified that it works).

I am open to fresh ideas... but we can't wait for #854 to fix this...

jwpeterson:
> Locking the whole thing is not an option

Bummer, but is that because you profiled it and it's too slow, or because it doesn't actually fix the race?

friedmud:
Well... you can see some timing I've done here: https://docs.google.com/spreadsheets/d/1fBvWxXTPTkYcdbfonPS_lDuB3zd5uV6545pCcqtOM9k/edit?usp=sharing

You can definitely see thread contention take over with `_get_array` being fully locked.

That said - it doesn't seem to effect the single threaded run all that much.  I'll run some more tests... simple_diffusion is... well... simple ;-)

friedmud:
Added another test... but similar results

roystgnr:
"By making _array_is_present depend on the value of the other two it removes the possibility of a race condition."

Is this true on platforms with weak hardware ordering?  Just because a is set before b on one thread doesn't necessarily mean it will be on another.

What about on platforms where the compiler is sneaky enough to first optimize bool(sin(a)+cos(b)+10) away to true before deciding what instructions it can reorder?

From "C++ and the Perils of Double-Checked Locking", Scott Meyers and Andrei Alexandrescu,
> In essence, you’ve just fired the opening salvo in a war of optimization. Your
compiler wants to optimize. You don’t want it to, at least not here. But this is
not a battle you want to get into. Your foe is wiley and sophisticated, imbued
with strategems dreamed up over decades by people who do nothing but think
about this kind of thing all day long, day after day, year after year. Unless you
write optimizing compilers yourself, they are way ahead of you.

They even specifically cover the "let's call a function" case:

> So you try to disable inlining by defining a helper function in a different file,
thus forcing the compiler to assume that the constructor might throw an exception
and therefore delay the assignment to pInstance. Nice try, but some build
environments perform link-time inlining followed by more code optimizations
[5, 11, 4]. GAME OVER. YOU LOSE.

The trouble with finding a fence-free alternative is that all the clever little solutions floating around the net (using thread-local-storage to only lock once per thread is the best I've seen) seem designed for a Singleton that persists through the multithreaded life of the program.  For a "singleton" that may be repeatedly created and destroyed nothing seems applicable except memory fences.

Which raises the question again: could we just use a memory fence?  Maybe pthread_spin_trylock()?  Or we could use a full heavyweight lock in C++03 but use a fence in C++11.  I want to allow people to compile our stuff with C++03 still, but I'm okay with it not being as fast until they get a decent compiler.

friedmud:
What Meyers and Alexandrescu were talking about was a different problem.  They were trying to create an object and set a pointer and they had no way to enforce a dependency between the two.  Here, we can enforce a direct dependency between the values.

BTW: there is no way a compiler is going to know the possible _range_ of values of `sin()` and `cos()` and know that adding 10 to that will always be true.  Show me that compiler and I'll never use another!

But anyway: I don't want to argue too vehemently.  It is definitely a hack... and you have the right to turn it down.

I would like to add some perspective here though.  Up until last weekend `_get_array` has had a  completely unprotected race condition (like, a real one, not because of memory subsystem oddities or compiler optimization) for the last ~10 years (just a guess... history on Github only goes back to 2012 ;-) and we've only very rarely seen failures (even in my dedicated testing it only failed about 1 in 10,000 runs).  We've been successfully doing massive threaded runs (like >100,000 threads on BG-Q) for many years with this "severe problem".

Now there is a fix that I cannot make fail... even in hundreds of thousands of dedicated runs (looped the tests for hours) where I was trying to do just that.  Even if there is some 1 in a billion chance of failure... I would call that better than locking `operator()`.

Maybe the old... don't let "perfect" get in the way of "good enough"?

But: It's your call!

However: I can completely agree about C++11.  Tomorrow I will implement the C++11 fix so we can get a look at what it will be.  Unfortunately @permcody completely refuses to allow MOOSE to depend on C++11 :stuck_out_tongue: :wink: (I'll put Martineau's disclaimer here: I used emoji so that means I can say anything I want because I was joking!)

friedmud:
Oh - I didn't address `pthread_spin_trylock()`:

It turns out that there is no amount of locking you can do to fix this because there is no way in C++03 to enforce that the setting of `_array_is_present` happens after the lock.  For instance... if you tried to do this:

```C++
if(_array_is_present)
{
  lock();
  if(_array_is_present)
  {
    lock2();
    _first = stuff;
    _last = stuff;
    unlock2();
    _array_is_present = true;
  }
  unlock();
}
```

You might think that the synchronization stuff in pthreads would help at `unlock2()`... but it doesn't.  The compiler (or hardware!) can still choose to move `_array_is_present = true` wherever it wants (it's a completely independent instruction).  It can even move it _inside_ the section protected by `lock2/unlock2`.

I saw that covered on a different website about the difficulties of DCLP....

roystgnr:
> Show me that compiler and I'll never use another!

I'm not sure I can yet, but I'll bet I can soon.  Half the point of constexpr was to make it approved and expected for compilers to pull tricks like that.

> The compiler (or hardware!) can still choose to move _array_is_present = true wherever it wants (it's a completely independent instruction). It can even move it inside the section protected by lock2/unlock2.

Shit.  You may be right...  Although I don't see how even mutexes can be guaranteed to work in that case; what prevents a C++03-only compiler from deciding to move an instruction outside a synchronized block?

My respect for the C++11 threads people has been going through the roof during this research, BTW.  I used to think the problem was little more than "add a stamp of approval on some methods and data types that everybody was using with C++03 anyway".

Okay, maybe this is actually impossible to solve with C++03.  My vote is still "lock the whole damn thing with C++03, fix it properly with C++11", but if you want to merge this as a workaround in the meantime then :+1:  I'm okay with replacing "slightly broken" with "even less broken" as long as we don't just let it drop there.

permcody:
> BTW: there is no way a compiler is going to know the possible range of values of sin() and cos() and know that adding 10 to that will always be true. Show me that compiler and I'll never use another!

Actually, speaking of that one in a billion bug, your sneaky line of code could be it! There is a solution to this equation such that `std::cos(_first) + std::sin(_last) + 10 == 0`. While that may not cause a seg fault it certainly could cause odd behavior the day somebody hits it. If on the other hand you do change it so that it's always true, then it does become subject to being optimized out by some future compiler.

> What Meyers and Alexandrescu were talking about was a different problem. They were trying to create an object and set a pointer and they had no way to enforce a dependency between the two. Here, we can enforce a direct dependency between the values.

They are related. Creating a dependency isn't enough. This is the whole crux of the problem. Even if you do create a dependency on the values inside of the double lock, the other threads not in that section don't see that dependency at all. How can you guarantee that the hardware won't reorder the cache coherency so that other threads won't see those values in a different order even with your dependency?

> Okay, maybe this is actually impossible to solve with C++03. My vote is still "lock the whole damn thing with C++03, fix it properly with C++11", but if you want to merge this as a workaround in the meantime then :+1: I'm okay with replacing "slightly broken" with "even less broken" as long as we don't just let it drop there.

:+1: for this solution. This is also something @friedmud should like as it puts a real fire under us moving towards C++11.

> Unfortunately @permcody completely refuses to allow MOOSE to depend on C++11 :stuck_out_tongue: :wink:

This comment isn't accurate. What I said is that we need to understand the impact on our actual customers that unfortunately do tend to run archaic machines with broken compilers. We need to make sure our primary customers can run as we move forward. If they can, I'm all for moving forward.

jwpeterson:
> There is a solution to this equation such that std::cos(_first) + std::sin(_last) + 10 == 0

Wut

roystgnr:
http://preshing.com/20130930/double-checked-locking-is-fixed-in-cpp11/ seems to be the best list I can find of the many C++11-ish ways to create a guaranteed-correct DCLP.

permcody:
Isn't that what @friedmud implied? 

> BTW: there is no way a compiler is going to know the possible range of values of sin() and cos() and know that adding 10 to that will always be true. Show me that compiler and I'll never use another!

I'll also invoke smiley so I can say whatever I want

friedmud:
> Isn't that what @friedmud implied?

>  BTW: there is no way a compiler is going to know the possible range of values of sin() and cos() and know that adding 10 to that will always be true. Show me that compiler and I'll never use another!

The equation can definitely never be zero :smile: 

I was saying that the compiler has no way to figure that out at compile time and optimize that statement to `_array_is_present = true`.

@roystgnr countered with `constexpr` which I hadn't heard of.  I'm still not quite sure that even `constexpr` could lead to optimizing that whole RHS to `true`... but I've only been reading about it this morning.

At any rate... there seems to be a consensus that people just want to lock the whole thing.  That makes me throw up in my mouth a little bit... but I can understand that this PR does the same for others :wink: 

It basically means that without C++11 libMesh won't scale with threads.  Maybe we should just not allow threading without C++11?

dschwen:
The fparser optimizer makes such optimizations based on the ranges of
functions

jwpeterson:
> It basically means that without C++11 libMesh won't scale with threads. Maybe we should just not allow threading without C++11?

Would we also need to fully lock `_restore_array()`?  That seems to be called quite a bit more often than `_get_array()`... and would probably further skew your spreadsheet results.

permcody:
Libmesh may not scale without threads but most of our users don't use
threads today even on big runs because we can't trust them. We get
different answers and I hope we can figure out why with this current
effort. The impact of over locking is minimal for users.

roystgnr:
_restore_array() would need to be locked (or "tricked") too.

I'm okay with requiring C++11 for threading, or defaulting to --disable-tbb and --disable-pthreads when we don't have C++11 and requiring the user to specify them manually, or screaming "libmesh_experimental()" plus some explanation when someone uses threads from C++03... as long as it's well documented.

friedmud:
@permcody That's just more of a reason to disable threads without C++11.  If we don't think our users are using it anyway... then let's just turn off threads unless their compiler can do it properly.

@jwpeterson To be completely correct we would need to fully lock both.  That said, I don't think it's required for `_restore_array()` as we shouldn't be calling it within any threaded regions.  But I'm open to suggestions.

Ok: coming back around to the C++11 option we already have a design issue... we would need to make `_array_is_present` a `std::atomic`.  I'm not quite sure how to do that in a generic way.  We have `Threads::atomic`... but we currently don't have a C++11 specialization of it.  Not only that, but presumably we would have some _other_ threading model (TBB, OpenMP, Pthreads) enabled at the same time.

Should I just put a lot of ifdef code around it to make it a direct `std::atomic` (i.e. bypass `Threads::atomic`) when using C++11?

jwpeterson:
> Ok: coming back around to the C++11 option we already have a design issue... we would need to make _array_is_present a std::atomic. I'm not quite sure how to do that in a generic way. We have Threads::atomic... but we currently don't have a C++11 specialization of it. Not only that, but presumably we would have some other threading model (TBB, OpenMP, Pthreads) enabled at the same time.

Does it need to be generic if you are inside `#ifdef LIBMESH_HAVE_CXX11_THREAD` blocks?  I'd just use std::atomic regardless of what thread model is currently active in that case...

I do agree that we should have "full" support for C++11 thread model at some point, i.e. with support for `parallel_for`, `parallel_reduce`, `spin_mutex`, etc. etc.

friedmud:
@jwpeterson Thanks - that's what I was thinking as well.

roystgnr:
Here's a type of C++11 double checked locking that wasn't in my first link and that includes a non-singleton example. Looks like manually specifying memory ordering might be more efficient than just using std::atomic?

https://www.justsoftwaresolutions.co.uk/threading/multithreading-in-c++0x-part-6-double-checked-locking.html

friedmud:
@roystgnr I'm not so sure that's the right thing for us.  We're not allocating a chunk of data... and we don't have any `shared_ptr`s here.

My plan was to make `_array_is_present` be `std::atomic`... which I think is going to work out just fine.  Let me finish this up and you can take a look at it...

roystgnr:
Oh, the shared_ptr bit is definitely *wrong* for us.  I was just referring to the pattern where they specify memory_order_acquire/release/relaxed.

Start with std::atomic, though, and if that scales well enough then we don't need to go more esoteric to try and get it to scale better.

friedmud:
Haha... this stuff is SO simple... from http://en.cppreference.com/w/cpp/atomic/atomic_thread_fence

```
Establishes memory synchronization ordering of non-atomic and relaxed atomic accesses, as instructed by order, without an associated atomic operation. For example, all non-atomic and relaxed atomic stores that happen before a std::memory_order_release fence in thread A will be synchronized with non-atomic and relaxed atomic loads from the same locations made in thread B after an std::memory_order_acquire fence if there exist atomic operations X and Y, both operating on some atomic object M, such that the fence in thread A is sequenced before X, X modifies M, Y is sequenced before the fence in thread B, and Y reads the value written by X or a value written by any side effect in the hypothetical release sequence X would head if it were a release operation..
```

It starts off ok... but quickly gets out of hand... :smile:

friedmud:
@roystgnr I believe we need _both_ fencing _and_ atomic on the guard.

friedmud:
I've got it implemented... going to look at performance... and make sure it's working...

jwpeterson:
Closing this since we merged the fix in #856 instead.

Issue #854, opened 2016-03-06
Make `PetscVector::operator()` actually const


friedmud:
After doing some more thinking about https://github.com/libMesh/libmesh/pull/853 the problem is really that we shouldn't be doing _any_ of the stuff that is happening in `PetscVector::_get_array()`.  `NumericVector::operator()` is `const` for a reason... and we shouldn't be violating that by using `mutable` (like here: https://github.com/libMesh/libmesh/blob/master/include/numerics/petsc_vector.h#L510 ).  That's what's gotten us into this ( https://github.com/idaholab/moose/issues/6436 ) mess.

I propose this:

1.  We remove all `mutable` out of `PetscVector`.
2.  We add a new `virtual void NumericVector::cacheValues()` method that developers can explicitly call in order to allow `NumericVectors` to cache local data (i.e. what `PetscVector::_get_array()` is doing).  Developers can call this function to speed up subsequent `operator()` calls.  This will allow developers the chance to do this explicitly before threaded sections.
3.  Add a new method: `virtual void NumericVector::restoreCachedValues()`
4.  All of the current methods that call `PetscVector::_restore()` should be made to error out in the case that `cacheValues()` has been called... and the data is still cached (i.e. that `restoreCachedValues()` has not been called).

Essentially: I think that we should get much more explicit about the memory movement and caching inside `NumericVector`s.

friedmud:
For a while this can be a compile time switch that we turn on.

`--enable-thread-safe-numeric-vectors`

?

jwpeterson:
:+1: for making `operator()` actually `const`.  I have gone from regarding the `mutable` keyword as "interesting hack" to "actively evil".

PR #856, opened 2016-03-07
Implement thread safe locking in PetscVector.

friedmud:
Decided to start a new PR since this is a different idea... but most of the discussion that led to this is in #853.

This causes threads to not be scalable unless using C++11. refs idaholab/moose#6436 #853

You can see some timing of it over here: https://docs.google.com/spreadsheets/d/1fBvWxXTPTkYcdbfonPS_lDuB3zd5uV6545pCcqtOM9k/edit#gid=0

I am currently running infinite loops of MOOSE-test... but so far the C++11 implementation looks fast and correct.

### DIFF
--- a/src/numerics/petsc_vector.C
+++ b/src/numerics/petsc_vector.C
@@ -34,9 +34,6 @@
 namespace libMesh
 {
 
-// Mutex for Petsc vector thread safety
-Threads::spin_mutex petsc_vector_mutex;
-
 //-----------------------------------------------------------------------
 // PetscVector members
 
@@ -1318,13 +1315,21 @@ void PetscVector<T>::create_subvector(NumericVector<T> & subvector,
 template <typename T>
 void PetscVector<T>::_get_array() const
 {
+#ifndef LIBMESH_HAVE_CXX11_THREAD
+  Threads::spin_mutex::scoped_lock lock(_petsc_vector_mutex);
+#endif
+
   libmesh_assert (this->initialized());
   if(!_array_is_present)
     {
-      Threads::spin_mutex::scoped_lock lock(petsc_vector_mutex);
-
+#ifdef LIBMESH_HAVE_CXX11_THREAD
+      std::atomic_thread_fence(std::memory_order_acquire);
+#endif
       if (!_array_is_present)
         {
+#ifdef LIBMESH_HAVE_CXX11_THREAD
+	  std::lock_guard<std::mutex> lock(_petsc_vector_mutex);
+#endif
           PetscErrorCode ierr=0;
           if(this->type() != GHOSTED)
             {
@@ -1367,8 +1372,12 @@ void PetscVector<T>::_get_array() const
             _first = static_cast<numeric_index_type>(petsc_first);
             _last = static_cast<numeric_index_type>(petsc_last);
           }
-
+#ifdef LIBMESH_HAVE_CXX11_THREAD
+          std::atomic_thread_fence(std::memory_order_release);
+          _array_is_present.store(true, std::memory_order_relaxed);
+#else
           _array_is_present = true;
+#endif
         }
     }
 }
@@ -1378,12 +1387,22 @@ void PetscVector<T>::_get_array() const
 template <typename T>
 void PetscVector<T>::_restore_array() const
 {
+#ifndef LIBMESH_HAVE_CXX11_THREAD
+  Threads::spin_mutex::scoped_lock lock(_petsc_vector_mutex);
+#endif
+
   libmesh_assert (this->initialized());
   if(_array_is_present)
     {
-      Threads::spin_mutex::scoped_lock lock(petsc_vector_mutex);
+#ifdef LIBMESH_HAVE_CXX11_THREAD
+      std::atomic_thread_fence(std::memory_order_acquire);
+#endif
+
       if(_array_is_present)
         {
+#ifdef LIBMESH_HAVE_CXX11_THREAD
+          std::lock_guard<std::mutex> lock(_petsc_vector_mutex);
+#endif
           PetscErrorCode ierr=0;
           if(this->type() != GHOSTED)
             {
@@ -1418,7 +1437,12 @@ void PetscVector<T>::_restore_array() const
               _local_size = 0;
 #endif
             }
+#ifdef LIBMESH_HAVE_CXX11_THREAD
+          std::atomic_thread_fence(std::memory_order_release);
+          _array_is_present.store(false, std::memory_order_relaxed);
+#else
           _array_is_present = false;
+#endif
         }
     }
 }
--- a/include/numerics/petsc_vector.h
+++ b/include/numerics/petsc_vector.h
@@ -41,6 +41,11 @@
 #include <cstring>
 #include <vector>
 
+#ifdef LIBMESH_HAVE_CXX11_THREAD
+#include <atomic>
+#include <mutex>
+#endif
+
 namespace libMesh
 {
 
@@ -507,7 +512,12 @@ private:
    * currently accessible.  That means that the members \p _local_form
    * and \p _values are valid.
    */
+#ifdef LIBMESH_HAVE_CXX11_THREAD
+  // Note: we can't use std::atomic_flag here because we need load and store operations
+  mutable std::atomic<bool> _array_is_present;
+#else
   mutable bool _array_is_present;
+#endif
 
   /**
    * First local index.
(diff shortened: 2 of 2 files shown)


jwpeterson:
Nice, the C++11 numbers look pretty good!

If I understand correctly, we still have the doubly-checked lock structure for the C++03 code branch, even though we're locking the whole damn function... so we just end up checking the pointer twice in that case?

roystgnr:
:+1: and many many thanks.

permcody:
It looks like the scalability isn't too bad to 4 threads for C++03. This is
plenty good for now.

friedmud:
@jwpeterson yeah... just check the bool twice.  I don't think it's an issue.  It will only happen for the first thread through... on the first time you start reading from the vector.  Every other time it's just going to be `true` and the whole thing will be skipped.

@permcody I agree... it looks fine for moderate numbers of threads.  One thing I did here is I moved the C++03 mutex _inside_ the PetscVector... so at least locking is only done locally to that one vector (so if multiple threads are accessing multiple PetscVectors there is no slow down).  You can see that it's a bit faster than my previous implementation... even _with_ fully locking `_restore()` now as well.

friedmud:
Oh - One more thing... I've been looping the tests for about an hour with C++11... no failures yet so I think the memory fences are doing their job.

friedmud:
I would like to point out that this _is_ a 20% slowdown in single threaded solves with C++03!  So we didn't get this for free.

I highly recommend we move to allow some portions of C++11 be on in MOOSE as soon as possible.

pbauman:
> I would like to point out that this is a 20% slowdown in single threaded solves with C++03! So we didn't get this for free.

Does that go away if we disable all threading at configure time?

friedmud:
@pbauman currently there's actually no way to do that!

@jwpeterson has a fix for that over here: #847  I don't yet know how that might impact it.

I suppose I can try to merge his branch into mine and see what happens...

pbauman:
Oh, right, I forgot that had not been merged yet. I'm bueno as long as #847 makes the slow down go away (I can always just change my build script to turn off threading for production runs).

Also, thanks much for this. This kind of testing is worse than pulling teeth.

jwpeterson:
I'd say the impact would be minimal on the `--with-thread-model=none` case.  The extra cost of constructing the empty `spin_mutex` class... it might even get optimized away since it's empty.

friedmud:
@jwpeterson do you think it would be possible to define `spin_mutex` as a macro that does nothing?  It would be great if it would go away _completely_

(I guess we should be having this conversation on your PR!)

jwpeterson:
Oh, but I guess we'll have to change the ifdef checks slightly to:
```
#ifdef LIBMESH_HAVE_CXX11_THREAD && LIBMESH_USING_THREADS
```
?

src/numerics/petsc_vector.C
```diff
@@ -1367,8 +1372,12 @@ void PetscVector<T>::_get_array() const
             _first = static_cast<numeric_index_type>(petsc_first);
             _last = static_cast<numeric_index_type>(petsc_last);
           }
-
+#ifdef LIBMESH_HAVE_CXX11_THREAD
+	  std::atomic_thread_fence(std::memory_order_release);
```
jwpeterson: I think this line needs to be indented one more space... it's not lined up quite right.

src/numerics/petsc_vector.C
```diff
@@ -1378,12 +1387,22 @@ void PetscVector<T>::_get_array() const
...
     {
-      Threads::spin_mutex::scoped_lock lock(petsc_vector_mutex);
+#ifdef LIBMESH_HAVE_CXX11_THREAD
+      std::atomic_thread_fence(std::memory_order_acquire);
+#endif
+
       if(_array_is_present)
         {
+#ifdef LIBMESH_HAVE_CXX11_THREAD
+	  std::lock_guard<std::mutex> lock(_petsc_vector_mutex);
```
jwpeterson: Possibly indent one more space.

src/numerics/petsc_vector.C
```diff
@@ -1418,7 +1437,12 @@ void PetscVector<T>::_restore_array() const
               _local_size = 0;
 #endif
             }
+#ifdef LIBMESH_HAVE_CXX11_THREAD
+	  std::atomic_thread_fence(std::memory_order_release);
```
jwpeterson: indent one space.

friedmud:
Hmmm... check out the new timing test I did: https://docs.google.com/spreadsheets/d/1fBvWxXTPTkYcdbfonPS_lDuB3zd5uV6545pCcqtOM9k/edit#gid=0

These new methods are definitely slower than what's in #853

The C++11 option is about 10% slower in serial.  The C++03 option is 20% slower in serial.

permcody:
Can we just have a different branch that avoids the locking altogether if
n_threads == 1?

friedmud:
@permcody possibly.  Let me check it out

friedmud:
@permcody I tried it... it *maybe* helps a little bit in single threaded mode... but it also has some really weird characteristics as the number of threads go up.  Overall... I'm not impressed.

Ok guys - I gotta bow out on this.  I've put more time than I really had into this at this point.  I'm going to fix up the issues @jwpeterson pointed out and anything more is going to have to be investigated later.

I honestly think that #854 is a better idea in the long run... so we shouldn't put too much more time in optimizing this.

permcody:
Well I think I'd rather have correct code at the expense of speed given the
choice. Thanks for working on this I think this was time well spent.

friedmud:
@permcody No problem!  I was motivated... I wanted to _really_ fix it.  It's at least properly _fixed_ at this point.  We can do better optimization later.

jwpeterson:
> We can do better optimization later.

I agree, and I think #854 is a good approach for this.  I don't think it will affect apps too much... we'll just have to try it and see.  I'm merging this now, so we can possibly do a libmesh update today, but I'd like to work on #854 either later this week or next to see how it goes.

YaqiWang:
I am encounter an issue with one of Yak tests with multithreading. MOOSE and libMesh are both up to date. I use lldb and locate the line using libMesh::PetscVector::operator() causing trouble. I am using c++03. 

```
* thread #1: tid = 0x6dbc1, 0x0000000104c2b717 libmesh_dbg.0.dylib`libMesh::PetscVector<double>::operator(this=0x000000010e816500, i=87)(unsigned int) const + 375 at petsc_vector.h:1226, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x2b8)
  * frame #0: 0x0000000104c2b717 libmesh_dbg.0.dylib`libMesh::PetscVector<double>::operator(this=0x000000010e816500, i=87)(unsigned int) const + 375 at petsc_vector.h:1226
    frame #1: 0x0000000100a4835a libyak-dbg.0.dylib`YakFaceVariable::getLocalVector(this=0x000000010f804400, local_vector=0x00007fff5fbfcef8, pt=87, dof=3) const + 458 at YakFaceVariable.C:160
```
If I add lock in `YakFaceVariable::getLocalVector`, it seems the issue is fixed. But I want to know why this is happening. Any thoughts?

permcody:
Yuck, I hate to say this Yaqi, but I hope that you're doing something wrong for our sakes!

Up to this point we've been verifying that this works by running tests in a while loop all night and seeing the evidence that it works just fine on our test regression tool. It dawned on me just now that we may be able to force the bug immediately on any test by actually forcing the condition that all of these locks is trying to prevent. During testing why don't we stick in a "yield", "sleep" or whatever it the right operation is that we need to get the bad behavior right in the middle of this locking code. These locks are here to prevent the case where a thread is interrupted right in the middle of this critcal section while another thread plows through. If we make a specific thread sleep (i.e. 0) right in the middle of the critical section it should expose this bug if it still exists right? Unless of course Yaqi has managed to find yet another case.

@YaqiWang - before we start another bug hunt, I'm going to suggest that you update to the head of libMesh. Derek discovered that he messed up the patch after #858 was merged and John fixed it later here: https://github.com/libMesh/libmesh/commit/30ef8c099f82b59f70d50fc53e45a28485cbcd32
The later commit is still not in MOOSE's submodule so it may help? I was under the impression that John's fix only affected C++11 which seems correct since our build boxes aren't showing this behavior.

After that make triple sure that you have both patches actually compiled libMesh. Finally, if you have a repeatable test, you might try building with C++11 to see if it goes away. (--enable-cxx11 to the libMesh build script). The C++03 code is even more conservative than C++11 so I don't know what to tell you at the moment.

jwpeterson:
> I was under the impression that John's fix only affected C++11 which seems correct since our build boxes aren't showing this behavior.

That's right, the C++03 code path should already have been correct and foolproof -- it simply locks at the beginning of the function -- the later patch only affects C++11.

PR #858, opened 2016-03-08
Enhance configure test of C++11 threads.

jwpeterson:
With the new code introduced by 01072ba, we need to be sure the
compiler supports several things, including:

* std::atomic
* std::mutex
* std::atomic::store()
* std::atomic::exchange()
* std::atomic_thread_fence
* std::memory_order_acquire
* std::memory_order_release
* std::memory_order_relaxed
* std::lock_guard<T>

Refs #856.


friedmud:
Thanks for this @jwpeterson !  :+1:

Issue #832, opened 2016-02-18
FParser eval is not threadsafe


dschwen:
Calls to `Eval` on FParser objects are not thread-safe, despite the claims implied by `#define FP_USE_THREAD_SAFE_EVAL` as reported by @jwpeterson at #810

I'll investigate this.

friedmud:
That's seriously bad news.  We might need to detect this in MOOSE and
thrown an error if you're trying to use FParser and threads until a fix is
made

permcody:
John made a fix. He added a mutex in the appropriate place to protect
against concurrent access which was why several of our periodic boundary
tests were failing (with pthreads) but no longer are. Our PThreads target
is pretty clean and our TBB target is "cleaner".

dschwen:
I think I know what is going on. It just dawned on me that my JIT compiled functions probably share the same stack each across threads! I can fix this.

dschwen:
Ok, new theory. `libMesh::ParsedFunction::eval` uses a buffer `_spacetime` to pass the arguments to FParser. This buffer is a member (so threads can clobber it). We should probably make that buffer a local variable in function scope.

jwpeterson:
@dschwen did #861 fix the thread safety issue with FParser?

dschwen:
FParser IS thread safe. This can be closed. The issue is a different one.

PR #859, opened 2016-03-08
Make JIT Eval thread safe

dschwen:
This adds a local stack to the JIT compiled FParser functions to respect `FP_USE_THREAD_SAFE_EVAL`. Fixes a pretty big threading screw-up.

Refs #832

### DIFF
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -1,3 +1,9 @@
+#include "libmesh_config.h"
+// Communicate to fparser that threads are being utilized
+#ifdef LIBMESH_USING_THREADS
+#  define FP_USE_THREAD_SAFE_EVAL
+#endif
+
 #include "fparser_ad.hh"
 #include "extrasrc/fpaux.hh"
 #include "extrasrc/fptypes.hh"
@@ -19,6 +25,12 @@ using namespace FPoptimizer_CodeTree;
 #  include <dlfcn.h>
 #  include <errno.h>
 #  include <sys/stat.h>
+#endif
+
+#ifdef FP_USE_THREAD_SAFE_EVAL_WITH_ALLOCA
+#ifndef FP_USE_THREAD_SAFE_EVAL
+#define FP_USE_THREAD_SAFE_EVAL
+#endif
 #endif
 
 #include "lib/sha1.h"
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -576,7 +576,24 @@ Value_t FunctionParserADBase<Value_t>::Eval(const Value_t* Vars)
   if (compiledFunction == NULL)
     return FunctionParserBase<Value_t>::Eval(Vars);
   else
-    return (*compiledFunction)(Vars, pImmed, Epsilon<Value_t>::value);
+  {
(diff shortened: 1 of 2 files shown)


contrib/fparser/fparser_ad.cc
```diff
@@ -682,8 +703,8 @@ bool FunctionParserADBase<Value_t>::JITCompileHelper(const std::string & Value_t
   ccfile << "#define _USE_MATH_DEFINES\n";
   ccfile << "#include <cmath>\n";
   ccfile << "extern \"C\" " << Value_t_name << ' '
-         << fnname << "(const " << Value_t_name << " *params, const " << Value_t_name << " *immed, const " << Value_t_name << " eps) {\n";
-  ccfile << Value_t_name << " r, s[" << this->mData->mStackSize << "];\n";
+         << fnname << "(const " << Value_t_name << " *params, const " << Value_t_name << " *immed, const " << Value_t_name << " eps, " << Value_t_name << " s) {\n";
+  //ccfile << Value_t_name << " r;\n";
```
dschwen: I can probably remove this line entirely...

contrib/fparser/fparser_ad.hh
```diff
@@ -122,7 +122,7 @@ private:
   bool JITCompileHelper(const std::string &);
 
   /// JIT function pointer
-  Value_t (*compiledFunction)(const Value_t *, const Value_t *, const Value_t);
+  Value_t (*compiledFunction)(const Value_t *, const Value_t *, const Value_t, Value_t *);
```
dschwen: The compiled functions now take a pointer to a stack rather than using a local array.

contrib/fparser/fparser_ad.cc
```diff
@@ -601,7 +618,11 @@ bool FunctionParserADBase<Value_t>::JITCompileHelper(const std::string & Value_t
   // generate a sha1 hash of the current program and the Value type name
   SHA1 *sha1 = new SHA1();
   char result[41];
+  // 1. version string
+  sha1->addBytes("2.0", 3);
```
dschwen: I'm adding a version string here to force a rebuild of previously cached JIT objects.

contrib/fparser/fparser_ad.cc
```diff
@@ -576,7 +576,24 @@ Value_t FunctionParserADBase<Value_t>::Eval(const Value_t* Vars)
   if (compiledFunction == NULL)
     return FunctionParserBase<Value_t>::Eval(Vars);
   else
-    return (*compiledFunction)(Vars, pImmed, Epsilon<Value_t>::value);
+  {
+    // Use a local stack (alloca or new/delete), or the global stack for this parser object
```
dschwen: This block is more or less copied from the FParser Eval routine (@roystgnr, did you add this?)

dschwen:
Ok, I need to add some ifdefs copied from fparser.cc to the top of this file

contrib/fparser/fparser_ad.cc
roystgnr: IIRC this is straight out of upstream fparser

dschwen:
Argh, this **is** thread safe. And `libMesh::ParsedFunction` doesn't even use JIT compilation. Good news: I'm innocent, bad news: no idea what cause the eval to be not threadsafe.

PR #861, opened 2016-03-09
Make spacetime local 

dschwen:
libMesh::ParsedFunction is not threadsafe due to the `_spacetime` buffer. This PR is one suggested option to fix this. The other option is to create per-thread copies of the functions at the application level.

I have confirmed that this fixes the wedge test issue in MOOSE, @jwpeterson.

Refs #832

### DIFF
--- a/include/numerics/parsed_function.h
+++ b/include/numerics/parsed_function.h
@@ -121,11 +121,13 @@ protected:
 
 private:
   // Set the _spacetime argument vector
-  void set_spacetime(const Point & p,
+  void set_spacetime(std::vector<Output> & spacetime,
+                     const Point & p,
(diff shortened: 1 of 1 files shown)


roystgnr:
The proper fix is to create per-thread copies of the functions.  That's why in FunctionBase we have

```
  /**
   * Returns a new copy of the function.  The new copy should be as
   * ``deep'' as necessary to allow independent destruction and
   * simultaneous evaluations of the copies in different threads.
   */
  virtual UniquePtr<FunctionBase<Output> > clone () const = 0;
```

dschwen:
Ok, makes sense. I'll try to track this down on the MOOSE side then.

PR #866, opened 2016-03-14
OrderWrapper shim class to enable arbitrary FE orders

roystgnr:
Based 99% on the work @dknez did.

Obsoletes #824

This seems to get everything compiling and running sanely for me.

I'd like to get this merged soon if everyone is happy; I just discovered that we might have a use for specifying *negative* FE order as well; I need to be able to specify temporary p-coarsening via some less hackish code in order to fix a race condition in threaded constraint evaluation on hp-refined meshes.

### DIFF
--- a/include/base/variable.h
+++ b/include/base/variable.h
@@ -118,7 +118,7 @@ public:
    * The number of components of this variable.
    */
   unsigned int n_components() const
-  { return type().family == SCALAR ? _type.order : 1; }
+  { return type().family == SCALAR ? _type.order.get_order() : 1; }
 
   /**
    * \p returns \p true if this variable is active on subdomain \p sid,
--- a/include/fe/fe_type.h
+++ b/include/fe/fe_type.h
@@ -73,15 +73,6 @@ public:
   }
 
   /**
-   * Operator that enables implicit conversion to
-   * an int.
-   */
-  operator int() const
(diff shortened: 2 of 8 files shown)


dknez:
Nice job getting that disambiguation in good shape, thanks! :)

roystgnr:
I feel like there has to be a better standard way to do that in C++.

If not then I'm not happy with the standard.  In general it would be nice if instead of "Error: two overloads have equal precedence for this function call, so I don't know what to do", the compiler would at most say "Warning: two overloads have equal precedence for this function call, so I assume either will work just fine".  If I write a function to handle ```foo(Bar b, T t)```, and another to handle ```foo(T t, Qux q)```, then I promise, it's okay to call either one to handle ```foo(b,q)```!  If what I'd really written with that first overload was implicitly ```foo(Bar b, TButNotQux nq)```, then I'd have used SFINAE and made the type explicitly TButNotQux.

roystgnr:
Ugh.  Not sure whether I missed that because I tested with a slightly older MOOSE build or because I only built in opt.

roystgnr:
I'll merge this tomorrow unless there's objections.

jwpeterson:
> I'll merge this tomorrow unless there's objections.

:+1:

PR #868, opened 2016-03-15
Fix p_level calculation when elem==NULL

roystgnr:
We clearly need better test coverage if this doesn't make libMesh or
MOOSE scream.

We should also get GRINS tests running with CI as soon as @pbauman has his new build hardware.

### DIFF
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -1911,7 +1911,7 @@ void DofMap::dof_indices (const Elem * const elem,
 
   // Use the default p refinement level?
   if (p_level == -12345)
-    p_level = elem->p_level();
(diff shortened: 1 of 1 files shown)


jwpeterson:
Yes... we don't do any kind of adaptive p-refinement in our test suite...

roystgnr:
This actually doesn't require adaptive p refinement to trigger, just attempts to get dof indices without an element.  It's what I added to handle coupling between PDEs and scalar ODEs.

roystgnr:
I didn't think anyone was doing that in MOOSE yet, and it'd be contrived to make a full fledged libMesh example, but I should have added a unit test.

PR #870, opened 2016-03-16
Add C++11 configure test for range-based for loops.

jwpeterson:
Refs idaholab/moose#6581.


friedmud:
Do you know if this passes on Ubuntu 120.04 and 14.04?

jwpeterson:
> Do you know if this passes on Ubuntu 120.04 and 14.04?

Just got back from lunch, will check now.

jwpeterson:
Yes, it works on both of those.

PR #871, opened 2016-03-16
More GenericProjector reoptimization

roystgnr:
This still isn't back to where we should be but it gives a big speedup on constant monomial variables and a decent speedup on lagrange variables, which are the two biggest MOOSE use cases.

### DIFF
--- a/src/systems/fem_context.C
+++ b/src/systems/fem_context.C
@@ -1632,14 +1632,17 @@ void FEMContext::pre_fe_reinit(const System & sys, const Elem * e)
             }
         }
 
-      // These resize calls also zero out the residual and jacobian
-      this->get_elem_residual().resize(n_dofs);
-      this->get_elem_jacobian().resize(n_dofs, n_dofs);
+      if (algebraic_type() != OLD)
+        {
+          // These resize calls also zero out the residual and jacobian
+          this->get_elem_residual().resize(n_dofs);
+          this->get_elem_jacobian().resize(n_dofs, n_dofs);
 
-      this->get_qoi_derivatives().resize(n_qoi);
-      this->_elem_qoi_subderivatives.resize(n_qoi);
-      for (std::size_t q=0; q != n_qoi; ++q)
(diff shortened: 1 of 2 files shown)


roystgnr:
Well, it gave a huge speedup on the problems where it didn't crash, at least.  Let me see if I can replicate the failures.

roystgnr:
That's far fewer failures, at least...

roystgnr:
Damn.  This might actually expose a MOOSE bug.  Checkout f8c2c5e9 and then cherry-pick 026673ad.  The result definitely gives *more* correct results than the libMesh head, which, instead of copying monomial values on refinement, does our default projection problem solve and thereby introduces epsilon rounding error.

But getting precisely correct values is enough to trigger failure in MOOSE adaptivity/cycles_per_step.test

Normally I'd say this might be the old "AMR can go in different directions depending on whether one element's error indicator value is epsilon higher or epsilon lower than another" problem, but there's no monomial solution values in this case, just monomials being used to hold metadata, and metadata which isn't getting rounded robustly IIRC when being converted back to int.

Time for a MOOSE patch fixing that first, I guess.

roystgnr:
See https://github.com/libMesh/libmesh/issues/831 for a description of the problem, although in that case it really was a regression rather than an anti-regression.

friedmud:
This sounds dubious to me.  I'm interested to see what you dig up....

roystgnr:
The only remaining regression failure here is the anti-regression (progression?) which should be cleared up by https://github.com/idaholab/moose/pull/6622

jwpeterson:
Invalidated now that idaholab/moose#6622 is merged, we'll see what happens.

jwpeterson:
Whoops.  Civet tests against MOOSE master, and your idaholab/moose#6622 had not made it there at the time I invalidated.

roystgnr:
Thanks for the update!  I was going nuts trying to figure out why this could be working for me and failing for the CI.

PR #872, opened 2016-03-17
CondensedEigenSystem: condense constrained dofs

roystgnr:
Doing so automatically should make the user's experience much easier;
it's hard to write condensed dof detection that doesn't have subtle
bugs in parallel, but most users aren't going to want to bother
condensing anything except constrained dofs.


dknez:
:thumbsup:

dknez:
I checked this with Torquil's example for 1 up to 8 MPI processes, and it works properly in all cases. As a result, I'm going to go ahead and merge.

PR #873, opened 2016-03-18
First crack at fixing nodal unique_id output

roystgnr:
Not even sure if this compiles yet, but let's abuse Civet; I'm running out of work time before a vacation and I don't want this to wait until Monday.

Thanks to @friedmud for the test case that pinned this down for us.

### DIFF
--- a/src/mesh/xdr_io.C
+++ b/src/mesh/xdr_io.C
@@ -860,23 +860,140 @@ void XdrIO::write_serialized_nodes (Xdr & io, const dof_id_type n_nodes) const
 
           io.data_stream (coords.empty() ? libmesh_nullptr : &coords[0],
                           cast_int<unsigned int>(coords.size()), 3);
-
-#ifdef LIBMESH_ENABLE_UNIQUE_ID
-          // XDR unsigned char doesn't work as anticipated
-          unsigned short write_unique_ids = 1;
-
-          io.data (write_unique_ids, "# presence of unique ids");
-          io.data_stream (unique_ids.empty() ? libmesh_nullptr : &unique_ids[0],
-                          cast_int<unsigned int>(unique_ids.size()), 1);
-#else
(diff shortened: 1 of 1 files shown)


roystgnr:
Feel free to merge this without me once @friedmud and/or @nicholasmalaya reports success with it; I'm on vacation until Monday.

We need to make sure nothing needs a backport of this fix.  The nodal unique_id writing didn't make it into 0.9.5, right?  And we haven't branched for 1.0.0 yet?

jwpeterson:
> On Mar 18, 2016, at 8:39 AM, roystgnr <notifications@github.com> wrote:
> 
> Feel free to merge this without me once @friedmud and/or @nicholasmalaya reports success with it; I'm on vacation until Monday.
> 
> We need to make sure nothing needs a backport of this fix. The nodal unique_id writing didn't make it into 0.9.5, right? And we haven't branched for 1.0.0 yet?
> 

I don't think that was in 0.9.5, and no haven't started 1.0.0 but will soon.

nicholasmalaya:
I'm now able to successfully restart with both a small working example I had previously failing as well as large scale runs performed at TACC's LS5. I'm in favor of merging this into libmesh:master. 

Thanks @roystgnr.

roystgnr:
Thanks!  @friedmud, let me know if you see any further problems.

dknez:
Can you provide any further info about what situations you would have expected this bug to arise? Was it only an issue with AMR?

friedmud:
Thanks for tracking this down and fixing it @roystgnr !  The fix is MUCH more involved than I would have wanted to take on myself... so I'm really grateful that you were able to dig into this one!

roystgnr:
It was triggered by AMR only because nobody was using unrefined meshes with enough nodes to break.  Anything over the io_blksize (128,000) nodes would have been affected.

PR #876, opened 2016-03-24
Allow threading directly over std::vector ranges

friedmud:
As with any good PR from me... I'm sure this will be controversial... so let's get into it :wink: 

Here's the deal: I've always hated that with the threading interfaces in libMesh we have to make _copy_ of a `std::vector` (by creating a `StoredRange`) before we can do a `parallel_for` over it.  This is extremely wasteful because `std::vector` has random access iterators... and in fact a `std::vector` is _exactly_ what underlies `StoredRange` in the first place!  Making a copy wastes time (and space if the original `std::vector` is also needed!)

You can see where we do this in MOOSE:
https://github.com/idaholab/moose/blob/devel/framework/src/mesh/MooseMesh.C#L445
https://github.com/idaholab/moose/blob/devel/framework/src/mesh/MooseMesh.C#L696
https://github.com/idaholab/moose/blob/devel/framework/src/geomsearch/NearestNodeLocator.C#L140
https://github.com/idaholab/moose/blob/devel/framework/src/geomsearch/NearestNodeLocator.C#L155

In the current libMesh based code I'm working on I needed to run threads on _millions_ of medium sized `std::vector`s and I just can't pay the penalty of making a copy of each one... thus this PR (I'm already using this code... so if it doesn't make it through review I'll just copy this into my own app ;-)

This PR makes it so that a `StoredRange` can be initialized with the pointer of an existing `std::vector`.  That way no copy is necessary.  This makes `StoredRange` effectively a "range wrapper" that can very efficiently allow threaded iteration over any `std::vector`... even on the fly in the middle of a C++11 call to `Parallel::parallel_for` that uses a lambda.  Very cool :-)

My one gripe with my current implementation is that I have to set a `_should_release` to know whether `StoredRange` created the `std::vector` or if it was passed in.  If this were done with `std::shared_ptr` that wouldn't be necessary.  However, we currently don't have a libMesh wrapper for `std::shared_ptr`!

Ok - fire away ;-)

### DIFF
--- a/include/geom/stored_range.h
+++ b/include/geom/stored_range.h
@@ -67,7 +67,8 @@ public:
     _last(),
     _first(),
     _grainsize(new_grainsize),
-    _objs()
+    _objs(new std::vector<object_type>()),
+    _should_release(true)
   {}
 
   /**
@@ -84,9 +85,33 @@ public:
     _last(),
     _first(),
     _grainsize(new_grainsize),
-    _objs()
+    _objs(new std::vector<object_type>()),
+    _should_release(true)
   {
     this->reset(first, last);
+  }
+
+  /**
+   * Constructor.  Takes a std::vector of objects.
+   * Optionally takes the \p grainsize parameter, which is the
+   * smallest chunk the range may be broken into for parallel
+   * execution.
+   *
+   * Note: The std::vector passed in here MUST live for the
+   * lifetime of this StoredRange!
+   *
+   * TODO: This should be a std::shared_ptr in the future!
+   */
+  StoredRange (std::vector<object_type> * objs,
+               const unsigned int new_grainsize = 1000) :
+    _end(objs->end()),
+    _begin(objs->begin()),
+    _last(objs->size()),
+    _first(0),
+    _grainsize(new_grainsize),
+    _objs(objs),
+    _should_release(false)
+  {
   }
 
   /**
@@ -108,7 +133,8 @@ public:
     _last(er._last),
     _first(er._first),
     _grainsize(er._grainsize),
-    _objs()
+    _objs(NULL),
+    _should_release(false)
   {
     // specifically, do *not* copy the vector
   }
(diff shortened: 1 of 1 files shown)


roystgnr:
:+1: but give it a little time in case anyone sees a problem I missed.

I'm actually not sure why you'd expect this to be controversial.  The new API is imperfect, but it's a uniform improvement over the old API, and I don't see any way to do better without breaking backwards compatibility and/or requiring C++11.

In fact, if you really want to get crazy with the zero-copy threading, there's no reason why we need Elem and Node ranges to store copies of vectors under the hood.  Store pairs of iterators, cache their distance to return when range size() is called, and then instead of "iterate to build vector then iterate over vector in functor" we'd just be doing "iterate in functor".

The only catch would be the range splitting code.  Our node and element iterators are forward iterators but our current threads code assumes random access iterators.  Looks like that's just our own attempt at simplicity, though, not anything forced on us by the lower level APIs.  Even if we just replace + and - by std::advance and std::distance(), it looks like we'd still have an O(N) algorithm there.

permcody:
:+1: - Looks fine to me too.

Perhaps it's time to move MOOSE's C++03 compatible shared_pointer wrappers down to libMesh. We've cleaned up a lot of our library code since adding them and I think libMesh could benefit from them as well as this PR illustrates.

Our wrappers are nothing more than typedefs
https://github.com/idaholab/moose/blob/devel/framework/include/utils/MooseTypes.h#L31-L39

roystgnr:
Moving shared pointer wrappers to libMesh would be great.

Relying on them in library code... actually, yes, I think I'd be okay with that too, since they're already in our contrib subset of Boost.  I just don't want to add any more third party dependencies that require a separate download.

friedmud:
Ok - cool.  The reason I thought it would be controversial is the memory "ownership" thing.  I think it's good the way it is now, and it can get better with a `SharedPtr` On the future.

As for zero copies for element and node ranges... it's definitely something I've thought of... but we do it SO much I actually think it's worth having cached copies of those around where the predicates have already been evaluated.

At any rate: no time to look into that now, but definitely something to think about for the future...

PR #877, opened 2016-03-24
RB Change: Updated relative error in greedy algorithm.

dknez:
Previously we divided by the norm of the solution in order to provide a relative error measurement.
A better way to get relative error is to compare the error at each iteration to the initial error,
and hence ensure that the error has decreased a certain amount. This is the way that relative tolerances
are generally implemented in iterative solvers, for example.


dknez:
Actually, this isn't quite what I want... closing this PR.

PR #879, opened 2016-03-24
Nonblocking send/receive of packed ranges

friedmud:
Since my last PR went smoothly... how about this one? :smile: 

I _really_ needed to do non-blocking receives of packed ranges... so I implemented it.

The basic design is like this:
 1.  I gave the new functions new names because they have a different communication patterns from the others.  You REALLY MUST use these together!
 1.  `nonblocking_send_packed_range()` can _only_ send one message.  This is critical to being able to deal with the messages asynchronously on the receiving end.
 1.  To make that possible I upped the max message size in `pack_range()` (there was an arbitrary limit there... which I kind of understand... but is entirely too low for my purposes)
 1.  I added `Communicator::packed_range_probe()` to make it easy to check for packed_range messages.
 1.  It returns a `Status` that must be fed into `nonblocking_receive_packed_range()` to get the buffer sizing correct.  I actually think that _all_ nonblocking receives in `Parallel` should require a `Status` in the same fashion!  Right now they are all unsafe as they require that you resize your buffer exactly before calling the function!
 1.  `nonblocking_receive_packed_range()` is actually pretty simple:
  - Resizes the buffer according to the `Status`
  - Sets up a non-blocking receive
  - Sets up some `PostWaitWork` to actually do the unpacking and delete the temporary buffer.

So... the sending side is pretty simple... just call `nonblocking_send_packed_range()`.

On the receiving side you need to do these things:
 1.  Wait for a message using `Communicator::packed_range_probe()` and catch the `Status` that's returned from it.
 1.  Call `nonblocking_receive_packed_range()` passing in the `Status` you got from `packed_range_probe()`
 1.  Use `Request::test()` to see when the data has been fully  received
 1.  Use `Request::wait()` to finish up the request by unpacking the data and deleting the temporary buffer

I've checking in a full working implementation of an example code over here: https://github.com/friedmud/simple_libmesh_app/blob/non_blocking_packed_range/src/main.C

Note that that code is on the branch `non_blocking_packed_range` in that repository (in case you try to clone that repo).

Let me know what you think!


friedmud:
Sorry - fixing the small issue right now.  Will update the commit in a second...

friedmud:
BTW: I was able to use this yesterday in my current research application to achieve near-perfect scaling out to 4800 processors.  Before I was using non-blocking sends with _blocking_ receives... and it was killing scalability... I was only getting about 50% (or worse) at 4800 processors.

I plan to push it to 9600 procs today... and hopefully 15000 or so next week.  We'll see what happens...

So... at least for me... this stuff is working out well!

roystgnr:
For what purposes is a million-entry buffer too small?  I thought I'd exceeded the point at which MPI bandwidth typically plateaus, by nearly an order of magnitude.

roystgnr:
Ah, I think I get it.  It's not about optimization; you just need to fit everything into a single buffer so that packed_range_probe() will be able to tell you that the entire send is ready to receive?

But now what I still don't understand is the advantage of the non-blocking-receive-that-only-works-after-probe.  If you've done a probe, and it's verified that the message is ready to receive, then won't a blocking receive also not end up blocking?

friedmud:
A probe does not actually pull the data down into the buffer (on good MPI implementations anyway!  I have heard of some bad ones that do!).  Just because the probe can tell you that data is available does not mean the data has been transferred.  The receive itself does that... and the actual receive can happen asynchronously... either using a background thread (on some MPI libraries) or offloading that whole operation to the hardware (if available... which it is on infiniband which is what I'm using).

That allows your application to continue to do computation while the data is actually received.  You then come back to it later and `test()` to see if the data has been completely received and then you can use that buffer (here we still use `wait()` after `test()` tells us the receive is finished in order to trigger `PostWaitWork`... you don't actually need to `wait()` at this point)

friedmud:
Check the docs for `MPI_Iprobe`: https://www.open-mpi.org/doc/v1.8/man3/MPI_Iprobe.3.php

friedmud:
I just realized that I'm in the perfect position to actually show that non-blocking receives with non-blocking probes is better than _blocking_ receives with non-blocking probes.  So I just grabbed 2400 processors and fired up a couple of runs.

First: this is the result of my algorithm (which is a domain-decomposed, unstructured mesh, ray-tracing algorithm I'm working on for Supercomputing 2016) that uses _this_ PR (i.e. non-blocking receives prompted by non-blocking probes):

```
Duration: 32.4508
Average Processor Crossings Per Ray: 33.419
Max Processor Crossings: 153
Average number of rays in a received ray buffer: 5641.73

ns/intersection: 0.0337192
Levelized ns/intersection: 80.926

Execution time (s): 76754.1 (100%) 
Generation time (s): 350.218 (0.456286%) 
Tracing time (s): 45395.8 (59.1445%) 
Buffer time (s): 5927.24 (7.72238%) 
Force Buffer time (s): 700.848 (0.913108%) 
Wait  time (s): 1608.41 (2.09553%) 
Receive time (s): 18214.9 (23.7315%) 
```

`Duration` is the actual run-time (in seconds)... while `Execution time` is the sum of the run-time over all of the processors (in seconds).  The things to note here are that:

 1. A "perfect scaling" `Levelized ns/intersection would be about 75 ns/intersection.  So I'm seeing about 93% scalability (which is pretty damn good for 2400 MPI processes!)
 2. `Receive time` is measuring all of the time it takes to receive packed ranges... in this run it was about 24%... which is quite a large chunk of the runtime - but manageable.

Ok.  Now... let's switch to doing _blocking_ receives.  To do this I just changed the `receive()` call in `nonblocking_receive_packed_range()` to a _blocking_ receive.  So... I'm doing `packed_range_probe()` to do non-blocking checks for incoming messages then doing _blocking_ receives immediately after a message is detected.  Same problem, same processors.  Here are the results:

```
Duration: 44.833
Average Processor Crossings Per Ray: 33.419
Max Processor Crossings: 153
Average number of rays in a received ray buffer: 3992.3

ns/intersection: 0.0465854
Levelized ns/intersection: 111.805

Execution time (s): 106505 (100%) 
Generation time (s): 279.887 (0.262792%) 
Tracing time (s): 44642.5 (41.9158%) 
Buffer time (s): 4473.96 (4.20069%) 
Force Buffer time (s): 988.937 (0.928534%) 
Wait  time (s): 2879.12 (2.70327%) 
Receive time (s): 46341.4 (43.5109%) 
```

Note that:
 1. Duration went from 32.45 seconds to 44.833... yikes!
 2. Levelized ns/intersection went to 111.8... meaning only a 67% scalability.  Damn!
 3. `Receive time` is now taking up 44% of the runtime!  Damn!

SO.  I'm not just pissing in the wind here.  Non-blocking probes followed by non-blocking receives is kicking ass here.  I can get nearly perfect scalability (93% on this problem) using this scheme.

friedmud:
Ooooh - also just noticed that `Receive time` went from 18214.9 seconds to 46341.4... so non-blocking receives are effectively 2.5 times faster for this problem!

include/parallel/parallel_implementation.h
```diff
@@ -2052,6 +2049,29 @@ inline status Communicator::probe (const unsigned int src_processor_id,
...
+                              this->get(),
+                              &int_flag,
+                              stat.get()));
+
+  flag = int_flag;
+
+  STOP_LOG("packed_range_probe()", "Parallel");
+
+  return stat;
+}
```
roystgnr: Could this be merged with the existing Communicator::probe() somehow?  The flag seems like something that ought to be encapsulated in a Status somehow and queryable after any probe.  The use of T to specify a datatype here seems unnecessarily premature; it ends up being used in Status::size() later, but at that point the data type is also specified and it would be possible to call size(data_type).

friedmud: No way to merge it with the existing probe... that one is blocking and doesn't set a flag.

You are right about the datatype.  Either way with the datatype I just wanted to make sure it was inferred from the `Packing` buffer type.  This was the first way I thought to do it - but I like your idea of doing it in `nonblocking_send_packed_range()` better.  Don't know why I didn't think of that ;-)

I could make it look just like the other call - except add in the flag.

That said, I kind of think the existing `probe()` is a bit inconsistent.  Shouldn't we make `Communicator::probe()` return our own datatype (`Status`)?  Right now it returns a plain `MPI_Status` (do we even know that it's ok to return one of those by copy?).

Personally I would really like the new, non-blocking `probe()` to return a `bool` that is governed by the `flag` (and it wouldn't take a `flag` argument at all).

That way code could look like:

```
if (comm.nonblocking_probe(0, my_tag, my_status))
  comm.nonblocking_receive_packed_range(my_context, &my_type, my_status);
```

Then `nonblocking_receive_packed_range` will use `status.source()` and `status.tag()` implicitly.  The nice thing is... it will tie these two functions together even more closely... you definitely won't be able to use `nonblocking_receive_packed_range()` without first getting a status.

What do you think?

BTW: thanks for the suggestions.  I definitely want to get this right.  The MPI stuff in `Parallel` is really awesome and I want this to fit in the right way!

friedmud: Oh - one more thing.  I see that in a lot of the existing functions we're passing a pointer through just for the purpose of defining the template type.  Is there a reason we do that instead of just requiring template arguments on the function?  For instance, the above code could look like:

```C++
if (comm.nonblocking_probe(0, my_tag, my_status))
  comm.nonblocking_receive_packed_range<MyType>(my_context, my_status);
```

roystgnr:
You're right that blocking-vs-nonblocking probe is something we can't paper over, so we need two different versions...  I'd like them to be as similar as possible in other respects, but I guess there's not much room for improvement; "returns a flag" vs "doesn't return until a message is found" is inherent to the semantics.

I do like the idea of returning bool directly and Status by output argument in the non-blocking case.  Maybe enough to just make the blocking case return Status by output argument and deprecate the old version.

We should pass around our own Status rather than the MPI status.  This will still entail a MPI_Status copy under the hood, but that ought to be safe; the standard says it's not an opaque type (although there are setters and getters since it's implementation-defined), encourages creating arrays of it, and doesn't have any constructors (except for converting between fortran and C status objects) or destructors for it.

I thought I had a reason for passing pointers rather than using explicit type specification...  I'm afraid I don't recall what that reason was.  Consistency with other methods where we were passing in pointers for more substantial reasons?  Problems with the explicit specification on an old compiler?  This is definitely dubious enough to be worth revisiting and possibly changing.

Pulling source and tag from status... if we have to require a successful probe before receive, then I love making the API help enforce that.  Is there no way to avoid requiring a successful probe before receive, though?  I'd hate to hard code that and then have it turn into an atavism.

Sorry about the nitpicking.  This is fantastic and should go in soon; I just want to make sure the API is rock-solid before we set it in stone, to misuse metaphors.

I think the only thing I still don't like is the inability to do block chunking, and the ensuing inability to interoperate with the other send_ and receive_packed_range methods.  I'd envisioned more of a streaming process, and less of a "copy up to INT_MAX*sizeof(T) then send one giant hunk" process.  I'm not sure how to fix that, though.

If we have to wait for a successful non-blocking probe anyway, could we have a probe_packed_range that only returns success when multiple buffer chunks are all ready?  That doesn't fix the "one giant hunk" problem (except for being more tolerant of fragmented memory, I guess?), since work still couldn't begin on the first chunk until the last chunk was received, but it would allow interoperability.

friedmud:
@roystgnr I've been using this for a few months now on multiple machines / architectures and it's working well.

Do you still have the same objections... or would it be possible to merge this?

friedmud:
I even have lots of benchmarks showing that this is WAY better than doing blocking sends/receives of packed ranges...

roystgnr:
Those are more questions than objections.  And they're only merge-blocking questions insofar as I don't want to commit to an API that we might need to break later.  If you'll stick a libmesh_experimental() in there, and if you promise not to scream too badly if/when I announce "I figured out how to get interoperability; you'll just need to change all your calls in user code" some day, then we can merge first and ask questions later.  ;-)

friedmud:
That's all fine with me... I'll push a new patch

friedmud:
@roystgnr should I wrap that `libmesh_experimental()` in `#ifdef DEBUG`?  I don't want it slowing down calls to these function in `opt`

jwpeterson:
libmesh_experimental is already wrapped in libmesh_do_once

roystgnr:
One static local variable and if test shouldn't matter compared to hitting the network stack, should it?

friedmud:
@jwpeterson I know it's wrapped in `libmesh_do_once` I was hoping to avoid even _that_ `if` test.  I call this millions (billions?) of times.

But yeah: you're right... it's probably over optimization...

friedmud:
Ok - the `libmesh_experimental()`s are in there.  This is good to go on my end.

roystgnr:
Sorry for the delay.

Would you throw in a unit test at some point?  It'll be a month before I get a chance to consider any changes to this code, and if I do I'd like to know first-hand just how badly I might be breaking yours.  ;-)

friedmud:
Will do... thanks for merging!

PR #883, opened 2016-03-28
mark TypeTensor row() const

### LINKED ISSUE #880: Mark TypeTensor<T>::row const
I don't see why `TypeTensor<T>::row(const unsigned int r)` is not marked `const`. We've run into fetching rows of const tensor type material properties in MOOSE and need to make copies just to pull out a row.

dschwen:
Closes #880


jwpeterson:
:+1: seems like a weird omission... just checked the rest of the class and it appears const-correct.

Issue #882, opened 2016-03-28
VTK Output in parallel has some problems.


aeslaughter:
In MOOSE run the simple diffusion test with two processors. The following *.vtu files are produced, the second has all zeros:

simple_diffusion_out_001_0.vtu
```
<?xml version="1.0"?>
<VTKFile type="UnstructuredGrid" version="0.1" byte_order="LittleEndian" header_type="UInt32" compressor="vtkZLibDataCompressor">
  <UnstructuredGrid>
    <Piece NumberOfPoints="67" NumberOfCells="50">
      <PointData>
        <DataArray type="Float64" Name="u" format="ascii" RangeMin="0" RangeMax="1">
          0 0.1 0.1 0 0.2 0.2
          0.3 0.3 0.4 0.4 0.5 0.5
          0.6 0.6 0.7 0.7 0.8 0.8
          0.9 0.9 1 1 0.099999999999 0
          0.2 0.3 0.4 0.5 0.6 0.7
          0.8 0.9 1 0.1 0 0.2
          0.3 0.4 0.5 0.6 0.7 0.8
          0.9 1 0.1 0 0.2 0.3
          0.4 0.5 0.6 0.7 0.8 0.9
          1 0.1 0 0.2 0.3 0.4
          0.5 0.6 0.7 0.8 0.1 0
          0.2
        </DataArray>
      </PointData>
      <CellData>
```

simple_diffusion_out_001_1.vtu
```
<?xml version="1.0"?>
<VTKFile type="UnstructuredGrid" version="0.1" byte_order="LittleEndian" header_type="UInt32" compressor="vtkZLibDataCompressor">
  <UnstructuredGrid>
    <Piece NumberOfPoints="67" NumberOfCells="50">
      <PointData>
        <DataArray type="Float64" Name="u" format="ascii" RangeMin="0" RangeMax="0">
          0 0 0 0 0 0
          0 0 0 0 0 0
          0 0 0 0 0 0
          0 0 0 0 0 0
          0 0 0 0 0 0
          0 0 0 0 0 0
          0 0 0 0 0 0
          0 0 0 0 0 0
          0 0 0 0 0 0
          0 0 0 0 0 0
          0 0 0 0 0 0
          0
        </DataArray>
      </PointData>
      <CellData>
```

jwpeterson:
This is not related to your reproduction test, but it's also not good news for people using AMR with this Output class:
```
/*
 * FIXME: This is known to write nonsense on AMR meshes
[...]

PR #886, opened 2016-03-30
Fix issue with writing VTKIO files in parallel.

jwpeterson:
This branch fixes an issue with the VTKIO writer in which it would not write solution values in parallel.  I also refactored some stuff to make the code a bit easier to read and to save a few lines.

It won't pass the MOOSE testing initially.  I have to regold some files over there first, and I need to think about how to actually merge this, it should probably stay as a branch in upstream which we do a MOOSE update to, eventually merging it into libmesh master at a later time....

### DIFF
--- a/include/mesh/vtk_io.h
+++ b/include/mesh/vtk_io.h
@@ -104,13 +104,6 @@ public:
   void set_compression(bool b);
 
 private:
-#ifdef LIBMESH_HAVE_VTK
-  /**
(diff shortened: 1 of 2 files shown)


jwpeterson:
This fixes the issue raised in #882.

jwpeterson:
Huh, I guess it passed because `VTK!=true` on the test box it ran on, or possibly for all the libmesh PR tests.

PR #887, opened 2016-03-31
Fix issues with C++11 thread locks in PetscVector.

jwpeterson:
In _get_array() and _restore_array():
* Move first thread fence outside the first if-test.
* Move lock to immediately after first if-test.

@friedmud please take a look.  I did not run millions of times threaded to check, but the fix was pretty simple so hopefully it's OK.

### DIFF
--- a/src/numerics/petsc_vector.C
+++ b/src/numerics/petsc_vector.C
@@ -1315,21 +1315,21 @@ void PetscVector<T>::create_subvector(NumericVector<T> & subvector,
 template <typename T>
 void PetscVector<T>::_get_array() const
 {
-#ifndef LIBMESH_HAVE_CXX11_THREAD
+  libmesh_assert (this->initialized());
+
+#ifdef LIBMESH_HAVE_CXX11_THREAD
+  std::atomic_thread_fence(std::memory_order_acquire);
+#else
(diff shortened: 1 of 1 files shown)


friedmud:
The reason you don't see a fix from me is that I actually did this last night... but couldn't get a clean `run_tests` with it.

I couldn't figure it out... But it may have just been a local issue (maybe something else wasn't up to date).  Let's see what the testing system says in your libMesh update PR.

permcody:
Testing system only builds C++03 so that's not useful.

jwpeterson:
On Thu, Mar 31, 2016 at 9:38 AM, Cody Permann <notifications@github.com>
wrote:

> Testing system only builds C++03 so that's not useful.
>

I'll try it on my workstation.

-- 
John

jwpeterson:
I tested this PR on 200x200 simple diffusion with 1-8 threads using the command:
```
while :; do ../../../moose_test-opt --n-threads=$(( ( RANDOM % 8 )  + 1 )) -i simple_diffusion.i; if [ $? -ne 0 ]; then break; fi; done
```
and it ran all night without incident.  So I'm pretty convinced that it's working, and I'm going to merge it.

friedmud:
Cool - thanks for cleaning up my mess!

PR #888, opened 2016-04-01
Delete leading whitespace from libMesh.

jwpeterson:
The natural next step now that we have successfully cleaned up all the trailing whitespace.  Another big benefit: no more pesky tab indentation characters sneaking in!  I'll also post some updated emacs/vim/atom configuration files which will help enforce this new convention momentarily...


pbauman:
FWIW, I think having everything lined up in the first column is gross. There's a reason I let emacs auto-indent things --- it makes the code much easier to read.

dschwen:
The first column should be reserved for comment characters. I suggest you reindent everything exactly sixspaces.

jwpeterson:
>  I suggest you reindent everything exactly six spaces.

To be F77 compatible, maybe we should start in column 7.

jwpeterson:
> FWIW, I think having everything lined up in the first column is gross.

Sounds like @dschwen agrees.  How do you feel about column 7?

dschwen:
Not that you mention it, these benefits of fixed form never occurred to me. Fortran does use one based indexing, do six spaces mean starting in column seven! However I suggest that we stick to zero based indexing for now (or at least make it a separate PR). So let's start in column 5.

jwpeterson:
> So let's start in column 5.

Makes sense.  It looks like the code is already in row-major order, though, so we should be good there.

> or at least make it a separate PR

Agreed, I'd like to keep this in the history.

Issue #890, opened 2016-04-03
FE Optimization


friedmud:
So... I have a VERY time sensitive application.  It is not actually doing FE... but it _does_ need to evaluate FE variables at arbitrary points (actually, currently just _one_ point) within each element.  Here is the basic loop it's doing:

1.  Determine the point within the element
2.  Inverse map that point into the element's reference space
3.  FE::reinit() with that reference point
4.  Compute the value of variables using the shape functions

(This is using MOOSE... so it's using the FEProblem::reinitAtPhysical() code path.  There is some overhead in MOOSE as well which I'm already addressing here: https://github.com/idaholab/moose/pull/6676 )

Also... my application needs to obtain surface normal vectors.  It doesn't need any other information on the surface other than a normal vector.  This is where the problems start.

The FE objects are supposed to be "smart" and only compute what they're asked for.  The reality falls far short of that.  Especially in FEMap... it does LOTS with second derivatives (including a LOT of resizing and memory allocation and zeroing) even if they're not being asked for.

In the case where all I need is a normal vector... I also have to call `FE::get_phi()` or the `FE` object will happily compute everything (I understand why... but there should maybe be an override).  Currently I am calling `get_phi()` even for my face FE just to keep it from computing more stuff (I don't actually need `phi`!).  In that same vein... I don't need JxW at all because I'm not doing any integration...

Also: in general, there is _too much_ allocation of temporary objects happening in FE and FEMap.  This shows up directly in timing (even of normal MOOSE applications) and hinders threaded scalability (memory allocations and deallocations are mutexed).

This is absolutely critical to me right now.  My application is running about 5x slower than I need it to... and the majority of the time is going into the FE objects.  This is all for some huge runs I need to do for a conference that is a month away (where I have an hour long plenary talk).  I've barely slept in the last two weeks as I've been working on this... and I don't plan on sleeping much for the next month.  ANY help you guys can give me on optimizing this I would GREATLY appreciate!

Ok.  So, in my normal way I've made a little test application to help figure out where the time is going and allow us to make optimizations and see their effect.  That application can be found here: https://github.com/friedmud/simple_libmesh_app/tree/fe_optimization (note that it's the `fe_optimization` branch in that repo).

Right now it's just a simple little application that runs `FE::reinit()` on every element.  You can run the application like this:

```
./simple-opt --n-elem 10 --n-sweeps 10 --fe-reinit-qrule --fe-reinit-custom
```

And you will see this:

```
Reinit with qrule...
[...]

PR #892, opened 2016-04-05
Optimize FEMap

roystgnr:
My first crack at addressing some of the issues discussed in https://github.com/libMesh/libmesh/issues/890

On my system it cuts the init_reference_to_physical_map() runtime in half for point evaluations.  Nice, but that's for the most extreme case; I was hoping for more.

This needs serious testing.

### DIFF
--- a/examples/adjoints/adjoints_ex1/L-shaped.C
+++ b/examples/adjoints/adjoints_ex1/L-shaped.C
@@ -40,6 +40,7 @@ void LaplaceSystem::init_context(DiffContext & context)
   elem_fe->get_JxW();
   elem_fe->get_phi();
   elem_fe->get_dphi();
+  elem_fe->get_xyz();
 
   FEBase * side_fe = libmesh_nullptr;
   c.get_side_fe(0, side_fe);
--- a/examples/adjoints/adjoints_ex2/L-qoi.C
+++ b/examples/adjoints/adjoints_ex2/L-qoi.C
@@ -8,6 +8,20 @@ void LaplaceQoI::init_qoi(std::vector<Number> & sys_qoi)
   sys_qoi.resize(1);
 }
 
+void LaplaceQoI::init_context(DiffContext & context)
+{
+  FEMContext & c = cast_ref<FEMContext &>(context);
+
+  // Now make sure we have requested all the data
+  // we need to build the linear system.
+  FEBase * elem_fe = libmesh_nullptr;
+  c.get_element_fe(0, elem_fe);
+  elem_fe->get_JxW();
+  elem_fe->get_phi();
+  elem_fe->get_xyz();
+}
+
+
 // We only have one QoI, so we don't bother checking the qois argument
 // to see if it was requested from us
 void LaplaceQoI::element_qoi (DiffContext & context,
--- a/examples/adjoints/adjoints_ex2/L-qoi.h
+++ b/examples/adjoints/adjoints_ex2/L-qoi.h
@@ -21,6 +21,9 @@ public:
 
   virtual void init_qoi(std::vector<Number> & sys_qoi);
 
+  // Context initialization
+  virtual void init_context (DiffContext & context);
+
   virtual void postprocess() {}
 
   virtual void element_qoi_derivative(DiffContext & context, const QoISet & qois);
--- a/examples/adjoints/adjoints_ex2/L-shaped.C
+++ b/examples/adjoints/adjoints_ex2/L-shaped.C
@@ -40,6 +40,7 @@ void LaplaceSystem::init_context(DiffContext & context)
   elem_fe->get_JxW();
   elem_fe->get_phi();
   elem_fe->get_dphi();
+  elem_fe->get_xyz();
 
   FEBase * side_fe = libmesh_nullptr;
   c.get_side_fe(0, side_fe);
@@ -47,6 +48,7 @@ void LaplaceSystem::init_context(DiffContext & context)
   side_fe->get_JxW();
   side_fe->get_phi();
   side_fe->get_dphi();
+  side_fe->get_xyz();
 }
 
 #define optassert(X) {if (!(X)) libmesh_error_msg("Assertion " #X " failed.");}
--- a/examples/adjoints/adjoints_ex2/L-shaped.C
+++ b/examples/adjoints/adjoints_ex2/L-shaped.C
@@ -38,16 +38,12 @@ void LaplaceSystem::init_context(DiffContext & context)
   FEBase * elem_fe = libmesh_nullptr;
   c.get_element_fe(0, elem_fe);
   elem_fe->get_JxW();
-  elem_fe->get_phi();
   elem_fe->get_dphi();
-  elem_fe->get_xyz();
 
   FEBase * side_fe = libmesh_nullptr;
   c.get_side_fe(0, side_fe);
-
   side_fe->get_JxW();
   side_fe->get_phi();
-  side_fe->get_dphi();
   side_fe->get_xyz();
 }
 
--- a/examples/adjoints/adjoints_ex4/L-shaped.C
+++ b/examples/adjoints/adjoints_ex4/L-shaped.C
@@ -40,6 +40,7 @@ void LaplaceSystem::init_context(DiffContext & context)
   elem_fe->get_JxW();
   elem_fe->get_phi();
   elem_fe->get_dphi();
+  elem_fe->get_xyz();
 
   FEBase * side_fe = libmesh_nullptr;
   c.get_side_fe(0, side_fe);
@@ -47,6 +48,7 @@ void LaplaceSystem::init_context(DiffContext & context)
   side_fe->get_JxW();
   side_fe->get_phi();
   side_fe->get_dphi();
+  side_fe->get_xyz();
 }
 
 #define optassert(X) {if (!(X)) libmesh_error_msg("Assertion " #X " failed.");}
--- a/examples/reduced_basis/reduced_basis_ex6/rb_classes.h
+++ b/examples/reduced_basis/reduced_basis_ex6/rb_classes.h
@@ -141,6 +141,7 @@ public:
     elem_fe->get_JxW();
     elem_fe->get_phi();
     elem_fe->get_dphi();
+    elem_fe->get_xyz();
   }
 
   /**
--- a/examples/vector_fe/vector_fe_ex3/curl_curl_system.C
+++ b/examples/vector_fe/vector_fe_ex3/curl_curl_system.C
@@ -96,6 +96,7 @@ void CurlCurlSystem::init_context(DiffContext & context)
 
   side_fe->get_JxW();
   side_fe->get_phi();
+  side_fe->get_xyz();
 }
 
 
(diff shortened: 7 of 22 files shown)


friedmud:
Holy crap!  Thanks so much Roy!  I'm going to try it out RIGHT NOW!

permcody:
Nice - I'll try to pull this branch as well and see if I can benchmark it against some of our heavier tests once the tests pass. Thanks!

friedmud:
@permcody I don't expect it to help MOOSE much... yet

We currently request _most_ things from the FE object... so most stuff will get recomputed even with this patch.  We may need to break up our FE objects... have some for "special occasions"... especially one that ONLY computes shape functions for point value evaluation (when we need nothing else).  There are many cases in MOOSE when that is true.

That said - this may still help _some_

friedmud:
I'm seeing some good speedup with this in the test application:

```
> ./simple-opt --n-elem 500 --n-sweeps 50 --fe-reinit-qrule --fe-reinit-custom

Reinit with qrule...
Total Time: 2.50062
ns/element: 200.049
Levelized ns/element: 200.049

Reinit with custom point...
Total Time: 5.14279
ns/element: 411.423
Levelized ns/element: 411.423
```

Nearly 2x... which is awesome!  Definitely a good first step!

I'll see if I can get MOOSE going with it...

roystgnr:
FINALLY!

roystgnr:
So this definitely introduces the possibility of userspace breakage:

1.  Before, the physical point locations were calculated even if you didn't prerequest get_xyz().  After, if you try to use get_xyz() after reinit() you'll get an assertion failure in debug/devel modes or an OOB or segfault in opt/prof modes.

2.  Before, if you prerequested only get_xyz(), it would basically count as no prerequest at all, which would put us on the backwards compatibility "just compute everything" code path, which would let you then use everything later.  After, if you prerequest only get_xyz() from an FE, requesting anything else after reinit will be another assert/OOB failure.

3.  For side and edge reinit, we need to do FEMap prerequests from within the library itself, which probably breaks anyone who was trying to use our "just compute everything" code path with those.

I'll add a unit test to verify my suspicions on (3), then fix if necessary.  I'll also add a libmesh_deprecated() warning to that code path; anybody still not using prerequests at this point should be informed that their code should be better.  For (1) and (2) I don't think there's a fix which won't hurt efficiency, and technically affected user code is wrong, so I'll leave it to users to fix.

But anyway, even before I fix (3), this should now be in shape for people to play with and test with.

friedmud:
Awesome @roystgnr !  I'm crazy busy today - but I'll try this out tonight for sure!

Thanks!

jwpeterson:
Interesting that we had this inconsistency with `get_xyz()` vs. the other getters all along, I never knew that.

jwpeterson:
For whenever you get around to cleaning up the branch before merge:

```
$ git diff --check a59cce1..775439d
src/fe/fe_map.C:196: tab in indent.
+                     (mapping_elem_type, mapping_order, i, qp[0]);
```

friedmud:
Wait - what happened here?  This is what I see from the simple timing application now:

```
./simple-opt --n-elem 500 --n-sweeps 50 --fe-reinit-qrule --fe-reinit-custom

Reinit with qrule...
Total Time: 26.0809
ns/element: 2086.47
Levelized ns/element: 2086.47

Reinit with custom point...
Total Time: 30.9024
ns/element: 2472.19
Levelized ns/element: 2472.19
```

It got 10x slower!

friedmud:
This is what I see in a profiling run now:

slow

friedmud:
After a bit of manual bisecting (should have used _actual_ bisecting but I was lazy :-) it appears as if fa46135 is the culprit.  10x slowdown from that commit on.

I haven't looked into why yet...

friedmud:
I've double checked it - it REALLY is that commit - but I can't see why...

friedmud:
Yep: If I take this branch exactly as it is... and all I do is revert fa46135 it fixes it!  The only thing I can see in that commit is that it added some extra `this->_fe_trans->init_map_phi(*this);` to `determin_calculations()`.

I'm not sure if that's what's causing the slowdown... but it does feel like a bit of a design problem anyway.  `determine_calculations()` shouldn't really be "initializing" data-structures like that... (that's just IMO...)

roystgnr:
Take a look at what init_map_phi() actually does - it initializes necessary prerequests, it doesn't initialize any data structures.

roystgnr:
What's strange is that point queries should have *sped up* after that commit - before, get_xyz() got called even if you didn't need it; after, it didn't.

roystgnr:
Thanks for catching this; there's definitely a regression hiding there.  I'm on it.

roystgnr:
Oh, hell, I think I get it!  Now that we no longer calculate xyz without a prerequest, calculating phi alone asks *nothing* from the map object.  And what do we get when we prerequest nothing?  We fall back on the "backwards compatibility" code path and calculate *everything*.

That should be easy to fix.  We don't actually need a backwards compatibility code path in FEMap since nobody uses that directly, so when the map is asked for nothing it should actually just compute nothing.

friedmud:
Hmmm... it's better, but still slower than it was...

```
./simple-opt --n-elem 500 --n-sweeps 50 --fe-reinit-qrule --fe-reinit-custom

Reinit with qrule...
Total Time: 3.65856
ns/element: 292.685
Levelized ns/element: 292.685

Reinit with custom point...
Total Time: 8.37675
ns/element: 670.14
Levelized ns/element: 670.14
```

roystgnr:
I'm actually seeing ~5% faster (than the original optimization, not just than before the regression).  But that's clearly trumped by ~50% slower.  I'll keep looking into it.

roystgnr:
Hmm...  there may be some compiler issues getting in my way, too?  I'm testing on an E5-2620, a chip that should be 50% as fast as the top of the line per-core, but I'm getting 25% of your speed.  Should I be rebuilding with icpc, -O3, or anything else fancy?

friedmud:
Can you post your numbers?  I'm having a hard time parsing what you're
seeing :-)

roystgnr:
```
$ ./simple-opt --n-elem 500 --n-sweeps 50 --fe-reinit-qrule --fe-reinit-custom 2>&1 | tee derek-after3.txt  
Reinit with qrule...  
Total Time: 15.2313  
ns/element: 1218.5  
Levelized ns/element: 1218.5  
  
Reinit with custom point...  
Total Time: 33.1688  
ns/element: 2653.51  
Levelized ns/element: 2653.51  
```

roystgnr:
Hell, even our newest machine (E5-2670v3) is giving me less than half the single-core speed of yours.

friedmud:
Interesting - this is just on my laptop.  I'm not doing anything interesting.  My compiler is `clang 3.7.0`...

roystgnr:
I can try clang.  What's your CPU?

friedmud:
According to what I can find it's: `22 nm "Haswell/Crystalwell" 2.6 GHz Intel "Core i7" processor (4960HQ)`

All of these numbers are for my 2013 Macbook Pro.

pbauman:
Yeah, I haven't been able to find memory bandwidth numbers, etc. for the Macbook's. My experience (on my late 2013/early 2014 model) is that it is a fair bit slower per core than my Ivy Bridge workstation (E5-2650v2), based on just coarse timings of benchmark problems in GRINS where the software stack is identical between the two machines (starting at the compiler level). So, I'd expect @roystgnr to be able to match your performance. But performance portability is a PITA.

friedmud:
We're talking a 4-6x change in performance here though.  Something is definitely up with that!

friedmud:
Oh - I _do_ have C++11 turned on too... but other than that it's a pretty typical libMesh build.

pbauman:
@friedmud are you using libstdc++ or libc++ (libcxx... whatever LLVM's std library replacement is)? I've never clocked the timing difference, but I wonder if that could be part of this?

friedmud:
I have no idea.  Maybe @milljm can answer that.

You can actually perfectly replicate my environment: http://mooseframework.org/getting-started/

Just download the OS X package and install it.  It all installs in /opt/moose and won't mess with your machine in any way.  There is just one line you put in your bash_profile to pull in the MOOSE environment... leaving it out gets you back to your normal environment.

milljm:
@pbauman Well I do not know much about it myself... Just enough to answer your question (I think).

-stdlib=libc++

I grabbed that info from an old set of instructions on how to build openMP (when LLVM-Clang did not support it) Here is the example so as to proved a bit more than just taking my word for it :)

```bash
cmake -DCMAKE_CXX_COMPILER=$PACKAGES_DIR/clang_3.5.1/bin/clang++ \
-DCMAKE_C_COMPILER=$PACKAGES_DIR/clang_3.5.1/bin/clang \
-DLLVM_ENABLE_LIBCXX=on \
-DCMAKE_BUILD_TYPE=Release \
-DCMAKE_INSTALL_PREFIX=$PACKAGES_DIR/clang_3.5.1  \
-DLLVM_CONFIG=$PACKAGES_DIR/clang_3.5.1/bin/llvm-config \
-Wno-dev -DCMAKE_CXX_FLAGS=-stdlib=libc++ \
-DCMAKE_C_FLAGS=-stdlib=libc++ \
-DCMAKE_EXE_LINKER_FLAGS=-stdlib=libc++ ..
````

roystgnr:
It's definitely not clang - I get slower results there than with gcc.

So I can't replicate the regression any further.  I'm also not sure I could repair it if I did.  The switch from direct _fe_map manipulation to _fe_trans calls was necessary to fix Nedelec elements.  IIRC the refactoring in compute_shape_functions was necessary to fix a regression triggered by MOOSE.

roystgnr:
Going to merge this soon unless there's objections.  It works fine with MOOSE and GRINS.

roystgnr:
Wait, almost forgot John's request for cleanup.  I'll do that first.

dknez:
Can you confirm the implications for applications? I haven't been following this thread closely, but I gather that applications that did not pre-request data properly will now hit an assertion failure in dbg mode, or segfault in opt mode? Might be worth sending an email to libmesh-users about that, after you merge.

roystgnr:
That's close to correct, and will-do.  Technically, "if you prerequest improperly you'll get an assert or a segfault" has been libMesh behavior for years, but the definition of "improperly" gets stricter here.

roystgnr:
That still leaves us with two commits in the history which I'd expect to fail.  I think I can get it down to 0, but after so much rebasing I figured I might as well push and make sure I haven't screwed anything up with the final product in the process.

roystgnr:
In theory that should leave every commit passing CI, as well as eliminate the "we need this method in this class, no its abstract base class, no just its superclass" reversions.  In practice I'm not going to test any of the intermediate commits, and I'll merge if/when this last commit passes.

jwpeterson:
:+1:

friedmud:
@roystgnr sorry I was incommunicado.  I was traveling... and the place I was at doesn't allow cell-phones or laptops :-)

Thanks for putting this in!

I'm still puzzled as to why my laptop is SO much faster than what you guys are seeing.  I'll try to do some digging on that soon by running on a few different machines.

I've ended up coming up with a workaround where I don't use the libMesh FE infrastructure for my current problem anymore.  That said: I really DO want to in the future for this problem so I do want to get this fixed.  It's just not quite as urgent any more.

In addition, after all of the digging I've done on this problem I really think there can be some pretty huge gains in performance by removing a lot of the temporaries from the FE computations.

I'll come back around to that... but right now I'm focused on this talk which is about 2.5 weeks away (and still no results!)

Thanks again for taking the time to work on this right now.  I think these changes are really valuable and will speed up a lot of libMesh based applications!

PR #895, opened 2016-04-13
TAO API change isn't in PETSc 3.6.4.

pbauman:
So let's guess 3.7.0 now.

Should've updated after @jwpeterson suggested it in #749. This gets us building with PETSc 3.6.4.


jwpeterson:
:+1:

pbauman:
FYI, I just built with the 3.7.0 release of PETSc and all is good.

PR #897, opened 2016-04-14
fix libmesh_nullptr usage

brianmoose:
libmesh_nullptr is used outside LIBMESH_HAVE_EXODUS_API but included inside.


jwpeterson:
:+1: we never noticed this since we don't build with Exodus disabled very often.  Not bad if this was the only regression though.

PR #900, opened 2016-04-15
Initialize pImmed pointer in FParser copy constructor

### LINKED ISSUE #899: FParser copy constructor omits a pointer necessary for JIT evaluation
The `pImmed` pointer, which is passed to the JIT compiled functions and contains the constants used in the expression. Remains uninitialized in a copy-constructed FParser. This leads to crashes when performing evals on copy-constructed JIT compiled FParser objects that use constants (or number literals). 

Yikes.

I can fix this by rebuilding the pointer in the copy constructor.

dschwen:
Closes #899


dschwen:
:star2: :boom: 900 :fireworks: :rocket:

jwpeterson:
Quite the momentous commit for #900.

Issue #199, opened 2014-02-10
RAII for perf_log


benkirk:
OK, START_LOG(.,.) / STOP_LOG(.,.) pairs are annoying.

@roystgnr, we've talked about this a few times, and I think I have a good solution I'd like to run by you. The basic problem is one of scope & unique names.

What we'd like is something like

``` c++
PerfItem foo(.,.,perf_log);
...
```

where the constructor begins the logging, and the destructor stops the logging.

But requiring unique names seems like a hassle and a pain to go back and drop in existing code.

What if we used the compiler **LINE** macro to create something automatic along the lines of

http://stackoverflow.com/questions/1597007/creating-c-macro-with-and-line-token-concatenation-with-positioning-macr

So then behind the scenes

``` c++
START_LOG("action A", "Group");
START_LOG("action B", "Group");
```

would expand to 

``` c++
PerfItem perf_item_1("action A", "Group",perf_log);
PerfItem perf_item_2("action B", "Group",perf_log);
```

and then STOP_LOG(.,.) would be a no-op, and implemented in the PerfItem (or better class name...) destructor?

@permcody - thoughts?

jwpeterson:
Not closing this yet, since we still have not implemented:

- [ ] static PerfLookup objects
- [ ] The ability to specify a PerfLog other than libMesh::PerfLog when creating `LOG_SCOPE` objects.

PR #906, opened 2016-04-21
RAII for PerfLog

jwpeterson:
This gets us about 95% of the way to the RAII PerfLog feature discussed in #199.  It uses the macro trick to generate a unique variable name based on `__LINE__`, but instead of outright replacing the existing `START_LOG`/`STOP_LOG` macros, I've introduced a new one called `LOG_SCOPE` that creates a simple logging object that does the `push()`/`pop()` stuff in the constructor/destructor.

There are situations where a scope-based logging macro just doesn't make sense, but they are pretty rare. Almost all `START_LOG`/`STOP_LOG` pairs in the library can be replaced with the new `LOG_SCOPE` macro, and the ones that can't easily be replaced can just remain unchanged.

Another nice aspect of this approach not mentioned in #199 is when logged code throws an exception.  Previously, if the exception was caught and dealt with, the PerfLog would likely have been in a bad state (`push()` with no corresponding `pop()`).  Now the destructor for the new class is guaranteed to be called, hopefully leaving the PerfLog in a valid state for continuing after dealing with the exception.

### DIFF
--- a/include/base/libmesh_logging.h
+++ b/include/base/libmesh_logging.h
@@ -26,6 +26,11 @@
 
 #include "libmesh/perf_log.h"
 
+// Two-level macro substitution trick, used to construct a unique
+// variable name for a given line.
+#define TOKENPASTE(x, y) x ## y
+#define TOKENPASTE2(x, y) TOKENPASTE(x, y)
+
 namespace libMesh
 {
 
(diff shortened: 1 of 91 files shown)


roystgnr:
Awesome!  But what's the missing 5% you allude to?

jwpeterson:
> Awesome! But what's the missing 5% you allude to?

Well in #199 you were talking about a macro that would create two objects instead of 1 that would allow us to do fewer string tests.  I didn't really understand how that would work, and I'd rather not introduce hidden static variables... but I suppose that could still be done if desired.

roystgnr:
Yeah, that was a wishlist item, though.  I'd consider it a separate potential upgrade, not an important part of this upgrade.

include/base/libmesh_logging.h
```diff
@@ -45,6 +50,44 @@ class PerfLog;
...
+{
+  PerfItem(const char * label,
+           const char * header,
+           bool enabled=true) :
+    _label(label),
+    _header(header),
+    _enabled(enabled)
+  {
+    if (_enabled)
+      libMesh::perflog.push(label, header);
```
roystgnr: Should we hardcode libMesh::perflog here?  That's obviously what we want for the LOG_SCOPE macro, but it might be nice to let users instantiate this class with their own PerfLog object instead.

roystgnr:
:+1: whether or not you agree with my "don't hardcode libMesh::perflog" quibble.  This looks great!

include/base/libmesh_logging.h
jwpeterson: Yeah, I was thinking about implementing that as another constructor argument that defaults to `libMesh::perflog`, but in the end decided it wasn't worth the extra macro name, or introducing variadic macros for.  This could definitely be added in the future, though.  Some of the examples create their own PerfLogs and could make use of it, but none of them have the nasty multiple-returns-requiring-matching-pops-for-correctness case.

Issue #903, opened 2016-04-19
Configure should fail if "make" or "find" utilities not present


jwpeterson:
We definitely require both, "make" obviously, and the "find" command is used directly in Makefile.am. configure checks for make,
```
configure:4208: checking whether make sets $(MAKE)
configure:4230: result: yes
configure:4259: checking whether make supports nested variables
configure:4276: result: yes
```
but it's not currently a fatal error if it's not found... we discovered this while @milljm was testing some very stripped down system images on Docker.  I don't think configure checks for "find" at all, currently.

roystgnr:
Do we want a missing "find" to be a fatal error?  We only use that in the install-data-hook, to fix up pkgconfig and libmesh-config scripts when someone does make install prefix=somethingelse.  That's such an uncommon use case that I'm not sure I want everything else to depend on it working.

jwpeterson:
> On Apr 20, 2016, at 7:04 AM, roystgnr <notifications@github.com> wrote:
> 
> 
> Do we want a missing "find" to be a fatal error? We only use that in the install-data-hook, to fix up pkgconfig and libmesh-config scripts when someone does make install prefix=somethingelse. That's such an uncommon use case that I'm not sure I want everything else to depend on it working.

I'm not sure I want to depend on "find" either, but the fact remains that you can't install libmesh without it.

roystgnr:
Sure, but we can fix that.  Wrap the relevant chunk of the install hook in "if type find"?

jwpeterson:
I'm looking at the `install-data-hook` target.  If you skip the whole `for genfile in` loop because you don't have `find` installed, you're going to get a (silently!) broken installation.  I also don't think a `find` requirement is an onerous one, almost all systems that we support either have it or can get it easily enough.  Letting configure fail early is the fastest way to alert someone that they need to install it.

PR #907, opened 2016-04-21
Replace find command in install-data-hook with shell wildcards.

jwpeterson:
This partially addresses concerns raised in #903, and should allow us
to 'make install prefix=/somewhere/else' on systems where find is not
installed.  We could still check for find in configure, but it should
not be a fatal error if it's not found.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -89,7 +89,7 @@ install-data-hook:
 	@if (test "x$(prefix)" != "x@prefix@") ; then \
           oldprefix="@prefix@" ; \
           newprefix="$(prefix)" ; \
-          for genfile in `find $(DESTDIR)$(prefix) -name "*.pc" -o -name Make.common -o -name libmesh-config -type f` ; do \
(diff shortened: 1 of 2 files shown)


roystgnr:
Ha!  I didn't even consider that we could rip it out entirely!  :+1:

Morbid curiosity has finally gotten the best of me, though: what system doesn't have *find*?  That was in the 1992 POSIX standard, it was in both System V and BSD Unix before that... ah, here we go.  https://en.wikipedia.org/wiki/Research_Unix came up with find in its 5th edition, released in 1974.

permcody:
It's not about not having find, it's that it doesn't work the same way
across systems. I blame OS X. We can't reliably use -prune, -depth, or
other slightly advanced flags and have them work the same way.

jwpeterson:
> Morbid curiosity has finally gotten the best of me, though: what system doesn't have find?

We were playing aroudn with installing MOOSE using Docker, and picked a very minimal Ubuntu distro.  It didn't come with 'find' or make installed, but of course it was trivial to install them...

> It's not about not having find, it's that it doesn't work the same way across systems.

This is true, but it isn't the case for the super simple usage pattern we had in Makefile.am.

PR #908, opened 2016-04-21
If PETSc has its own METIS, libMesh must use it!

jwpeterson:
This has become a bit more important since PETSc recently [patched](https://bitbucket.org/petsc/pkg-metis/commits/c27a7c30921dc587ae07286774ea2735ad21bb80) thread-local storage out of their METIS fork for all architectures, as it was broken by Xcode 7.3 (not METIS' fault, as far as I can tell).

We haven't done this to our METIS fork yet (I'd prefer to wait and see if Apple fixes their stuff first), so our GKlib/gk_externs.h header is now sufficiently different from the one distributed by PETSc, that we can't get away with mixing them (we had somehow been getting away with this in MOOSE for awhile, not sure exactly how, but it no longer worked for us as of 3.6.4).


dknez:
I configure with "--with-metis=PETSc". Will that be redundant after this change?

jwpeterson:
> Will that be redundant after this change?

Yeah, but it won't hurt anything either.

jwpeterson:
This failed due to an unrelated timeout during debug testing in parallel.  We are going to dial back the load and try again...

jwpeterson:
Ugh, looks like there was another timeout.  I'm just going to merge this since the failures are related solely to the new build boxes.

PR #911, opened 2016-04-25
fixed syntax-error with 'disable-second'

BalticPinguin:
when configured with the option '  --disable-second ',  I get an error message with the newest version saying that 'get_d2xidxyz2' does not exist. 
With the changes here, it works for me. 
Maybe it is better to remove the whole function init_map_d2phi() in that case but I am not able to foresee the consequences of it completely.
I hope, this helps,
Best


jwpeterson:
This fix looks correct to me.  It shouldn't hurt anything to still call `init_map_d2phi()` when second derivatives are disabled, as it's just pre-requesting something that is most likely already pre-requested by something else.  I'll let @roystgnr merge since he recently added these new functions in #892.

roystgnr:
This is correct; thanks!

PR #912, opened 2016-04-26
Handle case where PETSc has no METIS, yet user requests to use it

jwpeterson:
This is a follow-up to #908 that handles the converse case.

### DIFF
--- a/configure
+++ b/configure
@@ -38478,6 +38478,19 @@ fi
     build_metis=no
   fi
 
+  # Conversely, if:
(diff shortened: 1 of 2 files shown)


jwpeterson:
@brianmoose another timeout (bcs/periodic.testlevel1) on moose test parallel in dbg mode on new build boxes...

brianmoose:
Ugh, that is with -j3. We might need to wait until Cody's fix to the test harness gets in.

Issue #910, opened 2016-04-25
HDF5 Linking Error


friedmud:
Something isn't quite right with the linking of HDF5 if you build the HDF5 C++ interface.

I get this:

```
[...]

PR #913, opened 2016-04-26
Various fixes for HDF5 configure test

jwpeterson:
Misc. fixes and libMesh will now link against the HDF5 C++ library if it is found, as well as setting the `LIBMESH_HAVE_HDF5_CXX` preprocessor macro so you can guard code that uses the HDF5 C++ interfaces appropriately.

Refs #910.

### DIFF
--- a/configure
+++ b/configure
@@ -40495,17 +40495,23 @@ HAVE_HDF5=0
 
 # Check whether --with-hdf5 was given.
 if test "${with_hdf5+set}" = set; then :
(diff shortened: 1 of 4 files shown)


jwpeterson:
Another failure on a new build box.  We've confirmed that even reducing the load to 6 on these boxes has no effect on the timeouts -- they still pop up.  We really need to figure out what is going on with those...

PR #915, opened 2016-05-02
Don't use Metis from unusable PETSc

roystgnr:
If we detect METIS via PETSc, but we can't or won't use that PETSc, then we shouldn't try to use that METIS either.

### DIFF
--- a/configure
+++ b/configure
@@ -38480,12 +38480,15 @@ fi
 
   # Conversely, if:
   # .) METIS is enabled in libmesh,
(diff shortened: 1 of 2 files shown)


jwpeterson:
Thanks, I didn't consider this case!

roystgnr:
I wouldn't have considered it either; I just ran into it.  ;-)  I'm rebuilding my PETSc too, of course, but I figure I might as well make us more robust while I'm at it.

PR #917, opened 2016-05-03
Bounding box fixes for ParallelMesh

roystgnr:
This should give us correct results rather than incorrect results and segfaults when using subdomain_bounding_box() or processor_bounding_box() with ParallelMesh.


jwpeterson:
:+1:

roystgnr:
Rebased, conflicts fixed.  I originally caught these bugs as part of #918 work but I extracted them rather than slip them deep into the middle of a conceptually unrelated PR.

Issue #889, opened 2016-04-02
"get_node()" API Consistency


friedmud:
This has driven me crazy for years... can we fix this?  Obviously, it's API breakage... so I know it won't get fixed soon... but I just want to "get it on the docket" for fixing at some point.

Look at these accessors in `Mesh`:

```C++
virtual const Point & 	point (const dof_id_type i) const libmesh_override 
virtual const Node & 	node (const dof_id_type i) const libmesh_override
virtual Node & 	node (const dof_id_type i) libmesh_override
virtual const Node * 	node_ptr (const dof_id_type i) const libmesh_override
virtual Node * 	node_ptr (const dof_id_type i) libmesh_override
```

Now look at them in `Elem`:

```C++
dof_id_type 	node (const unsigned int i) const
Node * 	get_node (const unsigned int i) const
```

Note a few things:
  - `node()` in `Mesh` returns a `Node &` where it returns a `dof_id_type` in `Elem`!  Ah!!!! It drives me crazy... I can never remember which one to freaking call on which object.  I actually just had a bug that I lost about 15 minutes to because I called the wrong one and through an automatic cast of `dof_id_type` to `TypeVector` it was assigning the NODE ID as the x coordinate of a `Node`!  AAAAAAHHHHHHHH!!!!!!
  - For getting a pointer to a node `Mesh` uses `node_ptr()` where `Elem` uses `get_node()`.  Again, I can never remember which one is which!!

Personally: I think `Elem` is wrong.  The ones on `Mesh` make more sense to me.  Can we change `Elem` so that `Elem::node()` returns a `Node &` and rename `Elem::get_node()` to `Elem::node_ptr()` so we can have some consistency!

Ok - just had to vent... but wanted that venting to get captured so that one day we have some hope of fixing this :smile:

permcody:
I always figured the difference was because `Elem` doesn't actually store the `Node` object. Yes, you could create a method that calls through to the Mesh to get the object but the Node just gives you what it has (i.e. the dofs).

Were you using `auto` by chance :wink:?

friedmud:
Ass!  lol!

I was NOT using auto.  Try this:

```C++
mesh.add_point(elem->node(0));
```

It will happily assign a node id as the x position of that new node in this other mesh.  Handy! :imp: 

Your justification doesn't make sense... because `Elem` _does_ have `get_node()` which does return a `Node` already.

While we're at it... what's up the the weird inconsistency with getting `Elems` from `Mesh` also:

[...]

PR #918, opened 2016-05-04
API renaming

roystgnr:
This is the first phase or two of the changes discussed in https://github.com/libMesh/libmesh/issues/889

We should probably cherry-pick the first couple commits here onto the next release; the zillion "get rid of old API usage" changes can wait and get more shakedown time.

### DIFF
--- a/examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
+++ b/examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
@@ -392,7 +392,7 @@ void assemble_shell (EquationSystems & es,
       // First compute element data at the nodes
       std::vector<Point> nodes;
       for (unsigned int i=0; i<elem->n_nodes(); ++i)
-        nodes.push_back(*(elem->node_ptr(i)));
+        nodes.push_back(elem->node_ref(i));
       fe->reinit (elem, &nodes);
 
       // Convenient notation for the element node positions
--- a/examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
+++ b/examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
@@ -392,7 +392,7 @@ void assemble_shell (EquationSystems & es,
       // First compute element data at the nodes
       std::vector<Point> nodes;
       for (unsigned int i=0; i<elem->n_nodes(); ++i)
-        nodes.push_back(*(elem->get_node(i)));
+        nodes.push_back(*(elem->node_ptr(i)));
       fe->reinit (elem, &nodes);
 
       // Convenient notation for the element node positions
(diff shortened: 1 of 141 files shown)


jwpeterson:
While testing this, did you try temporarily removing the old interfaces and letting the compiler tell you if you'd missed any deprecated calls?  That's what I did for the TypeVector::size() -> norm() change and it worked pretty well for something that was hard to text replace.

jwpeterson:
> We should probably cherry-pick the first couple commits here onto the next release; the zillion "get rid of old API usage" changes can wait and get more shakedown time.

@roystgnr already mailed me privately on this, there is no 1.0.0 release branch yet, but I plan to start one today or tomorrow.  This entire PR should just go into master, though in my opinion.

examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
```diff
@@ -392,7 +392,7 @@ void assemble_shell (EquationSystems & es,
       // First compute element data at the nodes
       std::vector<Point> nodes;
       for (unsigned int i=0; i<elem->n_nodes(); ++i)
-        nodes.push_back(*(elem->node_ptr(i)));
+        nodes.push_back(elem->node_ref(i));
```
jwpeterson: I think this is a perfect 2-line illustration of the shortcomings of the old interface.

src/mesh/parallel_mesh.C
```diff
@@ -290,10 +290,7 @@ dof_id_type ParallelMesh::parallel_max_node_id() const
 
 const Point & ParallelMesh::point (const dof_id_type i) const
```
jwpeterson: Should `Mesh::point()` actually be `libmesh_deprecated()` in your opinion?  You can catch the return value of `Mesh::node_ref()` as a Point if you really want to...

roystgnr:
I'll put it all into master, then.

And yes, I did comment out the old interfaces and recompile.  That last "Rename more calls" commit was the result.  It's so huge; I was clearly missing a lot from my first few searches.  I tried different regexps based on the stuff I'd missed to see if I could catch the same patterns in cpp-disabled code, but I wouldn't swear that there still aren't a few more old API calls I missed in --disable-something-weird or --enable-something-weirder code paths.

src/mesh/parallel_mesh.C
roystgnr: No, IMHO it shouldn't.

I actually thought about this!  Although I probably should have put my thoughts into a comment here rather than a git log message.

We don't want to deprecate point() for the same reason we don't want to put the shim for it into MeshBase: because some day in the distant future we may want to add mesh subclasses like CartesianMesh which generate elements and nodes on the fly, and in that case it will probably be cheaper to generate just the Point when requested, not the full Node.

We probably do want to move point() from serial_/parallel_mesh.C into unstructured_mesh.h; I just didn't do that in this PR because I was trying to keep the 3rd through Nth commits as straightforward a translation as practical.

PR #919, opened 2016-05-04
Work around PetscOptions API changes in PETSc 3.7

roystgnr:
We don't seem to have any affected library code, but one example used the old API.


jwpeterson:
:+1: and thanks for pointing this out.  We call this function in several places in MOOSE.

jwpeterson:
Squash these together before merge though.

roystgnr:
Done.  I won't bother verifying that it still passes tests after the git rebase fixup.

PR #920, opened 2016-05-04
Add support for actual parallel I/O of nodal solutions in Nemesis

jwpeterson:
Previously, we were broadcasting the serialized node-major solution vector to all processors in order to support parallel I/O.  This is obviously a big waste of memory for large problems, and has been replaced by an algorithm that does additional parallel communication to get just the values needed.  VTKIO, another parallel I/O class, currently falls back on the existing broadcast implementation, but could also be updated in the future.

A new kind of `NumericVector::localize()` interface (with implementations for PETSc, Trilinos, and the serial NumericVector types) was added to support this capability.  The basic idea is described in the numeric_vector.h documentation:

```
  /**
   * Fill in the local std::vector "v_local" with the global indices
   * given in "indices".  Note that indices can be different on every
   * processor, and the same index can be localized to more than one
   * processor.  The resulting v_local can be shorter than the
   * original, and the entries will be in the order specified by
   * indices.
   *
   * Example:
   *   On 4 procs *this = {a, b, c, d, e, f, g, h, i} is a parallel vector.
   *   On each proc, the indices arrays are set up as:
   *   proc0, indices = {1,2,4,5}
   *   proc1, indices = {2,5,6,8}
   *   proc2, indices = {2,3,6,7}
   *   proc3, indices = {0,1,2,3}
   *
   *   After calling this version of localize, the v_local vectors are:
   *   proc0, v_local = {b,c,e,f}
   *   proc1, v_local = {c,f,g,i}
   *   proc2, v_local = {c,d,g,h}
   *   proc3, v_local = {a,b,c,d}
   *
   * This function is useful in parallel I/O routines, when you have a
   * parallel vector of solution values which you want to write a
   * subset of.
   */
```

### DIFF
--- a/include/mesh/mesh_output.h
+++ b/include/mesh/mesh_output.h
@@ -106,8 +106,7 @@ public:
    * (u0,v0,w0, u1,v1,w1, u2,v2,w2, u3,v3,w3, ...)
    * and contains n_nodes*n_vars total entries.  Then, it is up to the
    * individual I/O class to extract the required solution values from
-   * this vector, perhaps using NumericVector::create_subvector(), in
-   * order to write chunks of the nodal solution on each processor.
+   * this vector and write them in parallel.
    *
    * If not implemented, localizes the parallel vector into a std::vector
    * and calls the other version of this function.
--- a/include/mesh/mesh_output.h
+++ b/include/mesh/mesh_output.h
@@ -40,6 +40,7 @@ namespace libMesh
 class EquationSystems;
 template <typename T> class NumericVector;
 
+
 /**
  * This class defines an abstract interface for \p Mesh output.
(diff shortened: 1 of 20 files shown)


friedmud:
Wow!  This is awesome!

We may be able to finally recommend that people with large problems use Nemesis now!

Thanks so much for working on this!

jwpeterson:
> Thanks so much for working on this!

Sure, I guess you didn't get a chance to try it out on the old branch for your application then?

roystgnr:
I still think we should put the DistributedVector implementation into NumericVector, since it should work as a default.  On the other hand, we don't *need* a default any time soon, since you resisted the temptation to be lazy w.r.t. EpetraVector.  :+1:

PR #925, opened 2016-05-10
Fixes for --disable-amr builds

roystgnr:
This was much more straightforward than the MPI-free case.  No conflicts between the patches, even.

### DIFF
--- a/examples/adaptivity/adaptivity_ex2/adaptivity_ex2.C
+++ b/examples/adaptivity/adaptivity_ex2/adaptivity_ex2.C
@@ -78,8 +78,10 @@ using namespace libMesh;
 // In subsequent examples we will employ adaptive mesh refinement,
 // and with a changing mesh it will be necessary to rebuild the
 // system matrix.
+#ifdef LIBMESH_ENABLE_AMR
 void assemble_cd (EquationSystems & es,
                   const std::string & system_name);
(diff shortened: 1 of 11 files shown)


jwpeterson:
Thanks.  I'm undecided whether to cherry-pick all dozen or so of the `--disable-amr` and `--disable-mpi` commits onto 1.0.0 or squash them down into two commits (one for AMR, one for MPI) before cherry-picking.  The latter has the benefit of making the history of the release branch a bit easier to understand logically, and it would still merge cleanly back into master if we decided to go that route...

roystgnr:
Damn, I didn't think about that.  I prefer separate commits in case we have to bisect anything, but to make the history easier to understand they really ought to be cherry-picked onto 1.0.0-based branches and then merged onto 1.0.0, and that sounds like a big pain in the neck.

jwpeterson:
> they really ought to be cherry-picked onto 1.0.0-based branches

Yeah, initially I thought I'd like to avoid merges on the release branch itself, but that would be the right thing to do if you want to maintain this same history on the release branch.  But, are you really planning to ever bisect on 1.0.0?

roystgnr:
Good point.  Never planning to bisect on the 1.0.0 branch.  Two merged commits works for me.

Issue #927, opened 2016-05-11
--disable-cxx11 should actually disable C++11


roystgnr:
Not just skip searching for it.

With libstdc++ from GCC 5.1 and up, -D_GLIBCXX_USE_CXX11_ABI=0 needs to be explicitly specified to use the pre-C++11-compatible ABI.

With g++ 6.1 and up, -std=c++03 needs to be explicitly specified to use the pre-C++11 standard.

This is an issue rather than a patch because I have no idea what needs to be done for non-gcc compilers.

jwpeterson:
This would be nice to have... 

I'm not sure clang will support doing this at all.  They have enabled random C++11 features (creating iostreams from std::strings rings a bell) in the last several versions of the compiler, even when *not* compiling with -std=c++11.

roystgnr:
Closing this issue, because it's time to just demand C++11 and stop supporting --disable-cxx11.

PR #928, opened 2016-05-11
Cppunit link test

roystgnr:
This inspired #927 but does not depend on or resolve it.  This PR just gives us a non-blocking failure at configure time rather than a failed "make check" at run time.

### DIFF
--- a/configure
+++ b/configure
@@ -41807,6 +41807,91 @@ $as_echo "$cppunit_major_version.$cppunit_minor_version.$cppunit_micro_version" 
 $as_echo "no" >&6; }
       no_cppunit=yes
     fi
+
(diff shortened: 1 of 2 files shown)


roystgnr:
There's no way that a cppunit configure change caused a timeout in mesh/patterned_mesh.patterned_generation, so I'm going to merge anyway.

jwpeterson:
> There's no way that a cppunit configure change caused a timeout in mesh/patterned_mesh.patterned_generation

That's a relatively new test.  It might be a bit large for a regression test...

jwpeterson:
I'll cherry-pick these over onto master as well.

PR #929, opened 2016-05-12
Rename SerialMesh->ReplicatedMesh, ParallelMesh->DistributedMesh

roystgnr:
So, @jwpeterson, when I told you "I don't have any upcoming PRs that need to get into 1.0", it turns out that what I actually meant was "I need to rejigger one of our most fundamental classes before 1.0 comes out".  My bad.

This implements the changes settled on a few months ago:
https://sourceforge.net/p/libmesh/mailman/libmesh-devel/thread/CAFfxPjp%2B0%3Dx_w-7eACBh9nTn5WR%2BrPjJVvytvb9iX_%2BnzaNDCw%40mail.gmail.com/
in order to make our mesh class names less confusing.

Shim subclasses should keep us backwards-compatible with older application code from before the change.  I don't intend to remove the old names for years; just want to make sure that the new names are available to everyone from 1.0.0 onward.

GRINS, my private apps, and MOOSE still compile and pass test suites for me.

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -742,7 +742,7 @@ public:
    * Prints (from processor 0) all DoF and Node constraints.  If \p
    * print_nonlocal is true, then each constraint is printed once for
    * each processor that knows about it, which may be useful for \p
-   * ParallelMesh debugging.
+   * DistributedMesh debugging.
    */
   void print_dof_constraints(std::ostream & os=libMesh::out,
                              bool print_nonlocal=false) const;
--- a/include/geom/elem_cutter.h
+++ b/include/geom/elem_cutter.h
@@ -40,9 +40,13 @@ namespace libMesh
 
 // Forward declarations
 class Elem;
-class SerialMesh;
+class ReplicatedMesh;
 class TriangleInterface;
 class TetGenMeshInterface;
+
+// This is for backwards compatibility, but if your code relies on
+// forward declarations in our headers then fix it.
+class SerialMesh;
 
 /**
  * This class implements cutting a single element into a collection
--- a/include/geom/elem_cutter.h
+++ b/include/geom/elem_cutter.h
@@ -59,7 +59,7 @@ public:
 
   /**
    * Constructor.  Initializes pointer data
-   * without requiring a full \p SerialMesh in this header file.
+   * without requiring a full \p ReplicatedMesh in this header file.
    */
   ElemCutter();
 
--- a/include/geom/elem_cutter.h
+++ b/include/geom/elem_cutter.h
@@ -154,10 +154,10 @@ protected:
   std::vector<Elem const *> _inside_elem;
   std::vector<Elem const *> _outside_elem;
 
-  UniquePtr<SerialMesh> _inside_mesh_2D;
-  UniquePtr<SerialMesh> _outside_mesh_2D;
-  UniquePtr<SerialMesh> _inside_mesh_3D;
-  UniquePtr<SerialMesh> _outside_mesh_3D;
+  UniquePtr<ReplicatedMesh> _inside_mesh_2D;
+  UniquePtr<ReplicatedMesh> _outside_mesh_2D;
+  UniquePtr<ReplicatedMesh> _inside_mesh_3D;
+  UniquePtr<ReplicatedMesh> _outside_mesh_3D;
 
   Parallel::Communicator _comm_self; // defaults to MPI_COMM_SELF
 
--- a/include/mesh/mesh.h
+++ b/include/mesh/mesh.h
@@ -24,12 +24,12 @@
 #ifdef LIBMESH_ENABLE_PARMESH
 #include "libmesh/parallel_mesh.h"
 namespace libMesh {
-typedef ParallelMesh DefaultMesh;
+typedef DistributedMesh DefaultMesh;
 }
 #else
 #include "libmesh/serial_mesh.h"
 namespace libMesh {
-typedef SerialMesh DefaultMesh;
+typedef ReplicatedMesh DefaultMesh;
 }
 #endif
 
--- a/include/mesh/mesh.h
+++ b/include/mesh/mesh.h
@@ -37,11 +37,11 @@ namespace libMesh
 {
 
 // Forward declarations don't like typedefs...
-// typedef SerialMesh Mesh;
+// typedef ReplicatedMesh Mesh;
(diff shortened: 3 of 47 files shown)


jwpeterson:
Does this really need to be in 1.0.0?

pbauman:
FWIW, I'd vote for it to be in 1.0.0. Would be nice to have that name change in if we're going to stamp 1.0. Just my two cents.

jwpeterson:
> FWIW, I'd vote for it to be in 1.0.0. Would be nice to have that name change in if we're going to stamp 1.0. Just my two cents.

Why? It's a huge change that we haven't had time to test in the wild, while the 1.0.0 branch is supposed be seconds away from being "final".  I have no problem cherry-picking bugfixes onto it but don't see why a huge API change needs to go in at the last minute.

jwpeterson:
> a huge API change

And yes, I realize it's not an API *breaking* change, but it's a change nonetheless.

pbauman:
One of the things I was hoping we could do after the 1.0 stamp is actually adopt a consistent release numbering scheme, e.g. X.Y.Z: X is API changes, Y is new features+bugfixes (no previous API changes), Z is bugfixes only. (Can be something else, but *something* other than "what feels right".) That doesn't carry a lot of weight if one of the first things we do after 1.0 is change the API of one the core classes.

Also, I think a sed+shim class is not the huge sweeping change that you're painting.

roystgnr:
My immediate motivation was simple.  I'd started tossing notes together for the libMesh 1.0 release paper, and I realized we had a few ugly options:

a) Describe SerialMesh and ParallelMesh by their old names, which would make the paper's terminology an atavism probably by the time it passed review.

b) Describe ReplicatedMesh and DistributedMesh by their new names, but leave those names out of the 1.0 release, which would leave the paper describing classes that didn't exist yet.

c) Describe ReplicatedMesh and DistributedMesh by their new names, and get those names into 1.0.

The third seemed like the lesser of three evils.

As a secondary motivation (which Paul seems to have beaten me to), if we get the name change into 1.0.0 then we can call it the official 1.x API.  If we don't, then we leave new users forced to decide whether they want to write a 1.0-incompatible code using the 1.1 API or a deprecated-at-birth code using the 1.0 API.

jwpeterson:
> adopt a consistent release numbering scheme, e.g. X.Y.Z: X is API changes, Y is new features+bugfixes (no previous API changes), Z is bugfixes only.

I would like to do this as well, but I'd rather follow the PETSc model where they can make backwards-incompatible changes in "Y" level updates.  Otherwise I think the major version number just gets silly huge...

> That doesn't carry a lot of weight if one of the first things we do after 1.0 is change the API of one the core classes.

Yes, but as we both just agreed, this isn't an API change (in the usual API-breaking sense) and would not warrant a major or minor version bump anyway.

jwpeterson:
> The third seemed like the lesser of three evils.

Which language to use in upcoming papers is a good point.  I agree that if a paper is about 1.0.0, it needs to use the terminology in 1.0.0, but I'm OK with that still being the old terminology.

> they want to write a 1.0-incompatible code using the 1.1 API or a deprecated-at-birth code using the 1.0 API.

Are the old names already officially `libmesh_deprecated()`?  I didn't look closely at the diff yet, but if it's technically possible to hold off on doing that, I think I'd prefer it...

roystgnr:
Oh, hell no on the libmesh_deprecated().  Like with the changes in #918 (see, I remembered *one* of our crazy plans before the last minute...) I want to have at least one minor release (maybe two) with silently-working dual APIs before I start hassling people with deprecated messages.

roystgnr:
Damn.  Forget what I said in the commit log re: MeshOutput; there actually is another place where we're templating around mesh type, isn't there?  I'll fix that now.

roystgnr:
Any idea what could be going on there?  I can't reproduce it.  And I can't figure out why anything would be calling MPI_Abort without a decent error message first.

jwpeterson:
We have a `dynamic_cast<SerialMesh &>` in `TiledMesh` without a check that it actually worked.  Think that could be an issue?  Are you still building with --enable-parmesh in your test?

permcody:
Make sure you are using the same CLI options we use. That particular test
uses the "--mesh-only" flag for instance. You can see the exact command
executed at the top of the failed test output.

roystgnr:
Running with the same CLI options doesn't trigger a failure for me.  I am running on a --enable-parmesh build, but I'd have expected that to *create* failures, not solve them...

permcody:
Well if you are unable to get to the bottom of it. I'll log in and take a
closer look on the build machine.

roystgnr:
I've managed to replicate it.  Looking into it now.

jwpeterson:
> I've managed to replicate it. Looking into it now.

Don't leave me in suspense, what did you have to change to replicate the build box failure?

roystgnr:
I just had to stop being an idiot.  I'd inadvertently linked against a master libMesh install rather than against the PR install.

And the problem definitely is the dynamic_cast - a ReplicatedMesh is failing a cast to SerialMesh.  That's actually pretty nasty.  I'll have to look into why there's a ReplicatedMesh rather than a SerialMesh floating around in non-updated code to begin with, but if there's no way to fix that, then this transition might require actual API breakage.

roystgnr:
Oh of freaking course.  I didn't override clone() in the shim classes.

roystgnr:
Need to run out early but I'll be back to check on this tonight.

roystgnr:
Okay, that last failure is just the new takes-too-long-in-debug-mode test.

Any objections to getting this into master ASAP, then?  We can wait longer on the getting-this-into-1.0 question to debate and/or to get more opinions.

jwpeterson:
I think we fixed the timeout problem with that one test, so I invalidated the build just to be sure.

If all is well, I'd squash the fixup commit 9f42973 into 6f79d51, or wherever the shim classes were first added.  I'd be in favor of merging this to master, but I still think (and I think the MOOSE regression test results showed) it's too much not-well-enough-tested code to squeeze into the 1.0.0 release.

roystgnr:
Okay.  Let me know if you want to squash b836363 too; if not I'll merge.  As-is, every commit here will pass "make check", but 48bf98b won't be compatible with all 3rd party packed_range code, including MOOSE.

jwpeterson:
> squash b836363 too

Up to you.

PR #932, opened 2016-05-17
Fixes for assorted odd build configurations

roystgnr:
PECOS finally has BuildBot up and running again!

This means that we can finally resume automatic testing of all the little fiddly configuration options that the main developers don't commonly use.  In particular I'm looking forward to explaining that the answer to "does --enable-complex work" is "yes, at least as of last night", rather than "yes, at least if you use a git hash from shortly after the last time someone complained about --enable-complex not working".

This PR is still a work in progress; I'll take off the "do not merge" tag once everything in our current set of test configurations builds cleanly again.

There are a lot of warnings to be fixed as well; I'll leave those in a separate PR.  My hope is that the fixes for even corner-case errors can get merged down to 1.0, but IMHO that's not worth it for fixes for warnings.

### DIFF
--- a/examples/adjoints/adjoints_ex1/femparameters.C
+++ b/examples/adjoints/adjoints_ex1/femparameters.C
@@ -58,7 +58,9 @@ FEMParameters::FEMParameters(const Parallel::Communicator & comm_in) :
 
   system_types(0),
 
+#ifdef LIBMESH_ENABLE_PERIODIC
   periodic_boundaries(0),
+#endif
 
   run_simulation(true), run_postprocess(false),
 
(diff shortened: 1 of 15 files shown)


roystgnr:
That should be it for today.  I'd like to merge this afternoon so BuildBot can take a crack again overnight.

jwpeterson:
Does buildbot have any GitHub integration?  We could have it test PRs like Civet...

roystgnr:
It does, but we don't have any examples of it in our old config files (which may have predated it...) so I'd rather not start testing it on libMesh.  I'll pick one of our other GitHub hosted projects and give it a try.

jwpeterson:
Looks like everything here can go on 1.0.0?  I'll do that now...

roystgnr:
Definitely.  There'll be another swath of fixes to come later, though.

PR #933, opened 2016-05-17
Assorted warnings fixes

roystgnr:
This will probably end up being mostly fixes to -Wshadow, that being the warning option which we use in BuildBot but which I always neglect to enable on my personal builds.

### DIFF
--- a/include/parallel/parallel_implementation.h
+++ b/include/parallel/parallel_implementation.h
@@ -233,19 +233,19 @@ namespace {
 // Internal helper function to create vector<something_useable> from
 // vector<bool> for compatibility with MPI bitwise operations
 template <typename T>
-inline void pack_vector_bool(const std::vector<bool> & in,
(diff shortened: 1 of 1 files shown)


roystgnr:
I think we're past all the potentially-conflicting bugfixes in need of backporting, and I don't think I'm going to find time to add more warnings fixes any time soon, so this should probably be merged as-is whenever everyone's comfortable with that.

jwpeterson:
:+1: Is it still really difficult to pick out cases of "legitimate" shadowing from "you named a local variable the same name as a function that is also in this class" shadowing?

roystgnr:
It is!  I'd still call that shadowing and want to get rid of it, though I acknowledge that shadowing is much less concerning in cases where strong typing would make it impossible for an actual shadowing-induced-bug to slip by the compiler.

PR #934, opened 2016-05-17
Make Serial/ParallelMesh shims classes.

jwpeterson:
There's no reason not to, and it's more likely to be compatible with
people's existing forward declarations this way.

Refs #929.


roystgnr:
Thanks!  :+1:

PR #935, opened 2016-05-17
Make default unique_id size in help string match actual default.

jwpeterson:
The default is 8 bytes, but the help message said 4.


roystgnr:
:+1:

I'm not eagerly pushing for libMesh to be user-*friendly*, but I see no need for us to be user-*hostile*.

PR #936, opened 2016-05-18
Add sanity check for DirichletBoundaries

pbauman:
Check that the boundary ids provided to the DirichletBoundary are actually present in the mesh. If not, throw an error. Prior to these checks, the Dirichlet constraints were silently ignored.

I haven't run this through the libMesh examples (I'm abusing Civet), but did test on application code that error is thrown (or not) when expected. I dropped this in `create_dof_constraints` since that is called everytime `reinit_constraints` is called. That seemed appropriate to me since it's conceivable (although I'd guess unlikely) the mesh boundary ids changed between the calls to `reinit_constraints`.

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -1409,6 +1409,12 @@ private:
 
 #ifdef LIBMESH_ENABLE_DIRICHLET
   /**
+   * Check that all the ids in dirichlet_bcids are actually present in the mesh.
+   * If not, this will throw an error.
+   */
+  void check_dirichlet_bcid_consistency (const MeshBase & mesh,
+                                         const DirichletBoundary & boundary) const;
+  /**
    * Data structure containing Dirichlet functions.  The ith
    * entry is the constraint matrix row for boundaryid i.
    */
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -1158,6 +1158,10 @@ void DofMap::create_dof_constraints(const MeshBase & mesh, Real time)
          i = _dirichlet_boundaries->begin();
        i != _dirichlet_boundaries->end(); ++i, range.reset())
     {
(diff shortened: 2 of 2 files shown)


roystgnr:
Damn.  I thought this was overkill until I saw what it caught in GRINS.  :+1: if it passes MOOSE or if it uncovers any problems in tests there.

This shouldn't be backported to 1.0.0 though, right?  It's basically a new feature, not a bug fix, even if it's a great way to bug hunt in user codes.

pbauman:
> I thought this was overkill until I saw what it caught in GRINS.

Never underestimate the power of the universe to create idiots like me. :)

> This shouldn't be backported to 1.0.0 though, right? It's basically a new feature, not a bug fix, even if it's a great way to bug hunt in user codes.

Yeah, that's probably the right thing to do (although I admit I'm tempted to push to get this extra bug catching in).

pbauman:
Looks like timeout failure?

roystgnr:
Huh.  Timeout failure, but not in the same new test that was giving us trouble recently.  There's no way an extra get_boundary_ids() call is slowing things down noticeably, is there?  Even in a "multiapp" test where there might be some redundant reinitialization going on?

permcody:
That looks like a fairly heavy test but it is one that we don't typically
see failing. I'll invalidate the target and we'll keep an eye on it.

pbauman:
Cool, I guess it was fluke then?

pbauman:
@permcody, you agree it was a fluke? If not, I'd be OK with throwing a `libmesh_do_once()` around those calls since it's very unlikely the boundary ids will change on the fly.

permcody:
A fluke. I don't see why the pattern didn't match one of the two errors. If
it fails again, I'll work on trying to replicate the problem.

pbauman:
Cool, thanks. I'll go ahead and merge this then.

PR #942, opened 2016-05-20
suggestion to fix #875

### LINKED ISSUE #875: infinite elements: no initialisation of shape functions
Recently, I reallised that the infinite-element-part still seems to be erroneous: Adding the lines
```
if (elem->infinite()){
    out<<phi[i][qp]<<std::endl;
}
```
in the part, where the matrices Me and Ce are set up reveals, that the shape functions are 0 always.

I can try to fix it but I have not yet an idea what actually needs to be changed (except of course that I somewhere need to assign values to the shape functions; so far it seems they are just initialised).

BalticPinguin:
As pointed out in #875  ,the element shape functions are not computed for infinite elements.
With this pull request, I hope to fix this problem finally.
(more over, I have updated the website of tetgen, the old one is not available anymore)


jwpeterson:
Your fixes look reasonable.  If you don't mind, I will pull your branch and rebase it to get rid of the merge conflict, squash down the last 2 commits into 1, then merge by hand.  By the way, I assume this actually fixes the application where you were having issues (your message above says "hope to fix")?

jwpeterson:
Merged to master in a300214.

@roystgnr this seems like a bugfix/candidate for back-porting to 1.0.0, what do you think?

roystgnr:
Definitely!

BalticPinguin:
great, thank you for merging.
I wrote 'hope to fix' just because I thought already multiple times I would fix it, so I became pre-cautious with that :-)

PR #943, opened 2016-05-21
Global dirichlet id check

roystgnr:
The second commit here fixes a regression in DistributedMesh introduced by #936 and triggered by fem_system_ex3.  I'm confident this is the right fix.

The first commit here tries to fix the misleading documentation which led to the regression.  I'm less confident this is the right phrasing.

The third commit tries to make sure that, even if users are still thoroughly confused at this point about the supposed-to-be-local BCID sets our APIs hand them, we sanely catch any inconsistencies in the supposed-to-be-global BCID sets that they hand us.

### DIFF
--- a/examples/eigenproblems/eigenproblems_ex3/eigenproblems_ex3.C
+++ b/examples/eigenproblems/eigenproblems_ex3/eigenproblems_ex3.C
@@ -160,8 +160,10 @@ int main (int argc, char ** argv)
 
   {
     // Add boundary IDs to this mesh so that we can use DirichletBoundary
-    MeshBase::const_element_iterator       el     = mesh.active_local_elements_begin();
-    const MeshBase::const_element_iterator end_el = mesh.active_local_elements_end();
+  // Each processor should know about each boundary condition it can
+  // see, so we loop over all elements, not just local elements.
+    MeshBase::const_element_iterator       el     = mesh.elements_begin();
+    const MeshBase::const_element_iterator end_el = mesh.elements_end();
     for ( ; el != end_el; ++el)
       {
         const Elem * elem = *el;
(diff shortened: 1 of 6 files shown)


roystgnr:
And now the last commit makes sure that we're not failing to generate consistent BCIDs in our own user code.  I'll see if I can add a test for inconsistencies in the library in debug mode, and we should probably be checking our own apps too.  That type of bug was getting copied into example after example, and it *usually* doesn't trigger an error, but it could theoretically end up silently screwing up your solution even with SerialMesh depending on your repartitioning and on what you use BCIDs for.

roystgnr:
That test for inconsistencies will be a separate PR.  This one should be ready to merge as-is.  First and fourth commits should probably be cherry-picked onto 1.0.0

pbauman:
Sorry for sending you down this rabbit hole. Thanks for the fixes.

jwpeterson:
> First and fourth commits should probably be cherry-picked onto 1.0.0

Combined (and DistributedMesh comment edited to say ParallelMesh) in d671515.

roystgnr:
Thanks!

Although I originally had hopes that Ben and Derek would join the "get DistributedMesh into 1.0.0" bandwagon, so far I have yet to even see [the lurkers support me in email](http://www.jowaltonbooks.com/poetry/whimsy/the-lurkers-support-me-in-email/), so it looks like "Roy and Paul" is outvoted by "John and Civet and standard-practice".

boyceg:
> Although I originally had hopes that Ben and Derek would join the "get DistributedMesh into 1.0.0" bandwagon, so far I have yet to even see the lurkers support me in email <http://www.jowaltonbooks.com/poetry/whimsy/the-lurkers-support-me-in-email/>, so it looks like "Roy and Paul" is outvoted by "John and Civet and standard-practice".


I would like to see DistributedMesh in 1.0.0. :-)

PR #944, opened 2016-05-22
Added support for ExodusII's TRISHELL3 and SHELL4

dknez:
This involved the following changes:
- Added new element types TRISHELL3 and QUADSHELL4 to the element type enums
- Added "shell face" data to BoundaryInfo
- Updated ExodusII_IO to read in and write out data for these elements types

Note that I originally intended to not add new element types, and just use the existing TRI3 and QUAD4 types. However, it turned out that the new types were needed since we need to be able to distinguish a TRI3 from a TRISHELL3 when we're writing out or reading in the mesh.


dknez:
@jwpeterson, @pbauman does this PR look OK to you? It worked for me on some simple test cases. If we go ahead and merge this, then I'll ask @svallaghe if he can update the miscellaneous_ex11 to use QUADSHELL4 elements to provide some coverage of this in the examples.

jwpeterson:
> On May 22, 2016, at 5:11 PM, David Knezevic <notifications@github.com> wrote:
> 
> @jwpeterson, @pbauman does this PR look OK to you? It worked for me on some simple test cases. If we go ahead and merge this, then I'll ask @svallaghe if he can update the miscellaneous_ex11 to use QUADSHELL4 elements to provide some coverage of this in the examples.
> 

I'd like to take a closer look before merging but don't have time just this second... I agree it would be great if  @svallaghe could use the new element in misc_ex11. He can pull your branch and work on that before we merge this, if he wants.

jwpeterson:
Looks like @roystgnr's merging of #943 may have brought this into conflict?  Also, the MOOSE test that failed was only a timeout in debug mode, so probably nothing related to this branch.

include/geom/face_quad4.h
```diff
@@ -159,6 +159,27 @@ class Quad4 : public Quad
 
 };
 
+class QuadShell4 : public Quad4
```
jwpeterson: I think this new class should be in its own file.  Also it should have doxygen style documentation, e.g.
```
/**
 *
 */
```
at the top of the class.

include/geom/face_tri3.h
```diff
@@ -175,6 +175,27 @@ class Tri3 : public Tri
 
 };
 
+class TriShell3 : public Tri3
```
jwpeterson: Same comment about separate C/h file for new class, plus Doxygen-style documentation.

include/mesh/boundary_info.h
```diff
@@ -368,6 +410,32 @@ class BoundaryInfo : public ParallelObject
...
+   * Returns the list of boundary ids associated with the specified shell face
+   * of element \p elem. This is only relevant for shell elements.
+   */
+  void shellface_boundary_ids (const Elem * const elem,
+                               const unsigned short int shellface,
+                               std::vector<boundary_id_type> & vec_to_fill) const;
+
+  /**
+   * Returns the list of raw boundary ids associated with the specified shell face
+   * of element \p elem.  These ids are ``raw'' because they
```
jwpeterson: Are LaTeX-style quotes actually parsed by Doxygen?

src/fe/fe_lagrange_shape_2D.C
```diff
@@ -44,6 +44,7 @@ Real FE<2,LAGRANGE>::shape(const ElemType type,
         switch (type)
           {
           case QUAD4:
+          case QUADSHELL4:
```
jwpeterson: I suppose we should also technically support the new geometric element types for all the FEs (hierarchic, monomial, etc.) that currently support TRI3 and QUAD4, since they are basically the same thing?

jwpeterson:
This looks good overall.  If you don't want to mess with adding new C/h files to the build system, I can probably take care of that before merging...

include/geom/face_quad4.h
pbauman: > I think this new class should be in its own file.

I feel like that's overkill for a class with < 20 lines of code. It's overriding one function of the parent - I think that's related enough to justify keeping it in the same file and not having to monkey with the build system. Just my two cents.

dknez:
OK, thanks for looking through.

If you're happy to add the C/h files for the new elements types (and do any other cleanup that you want to see) and then merge, that'd be great. Otherwise, let me know and I'll do it.

Regarding the discussion of whether we should to add new files for the shell elements: I don't have a strong opinion either way on that.

@svallaghe is busy with some other stuff right now, so I think we should just go ahead with the merge and he can modify the example afterwards.

include/mesh/exodusII_io_helper.h
```diff
@@ -918,6 +994,15 @@ class ExodusII_IO_Helper::ElementMaps
   static const int quad_edge_map[4];
 
   /**
+   * Maps the Exodus edge numbering for "shell quads". In this case
+   * we have "6 sides", where the the first two sides correspond to
+   * the triangle faces and are mapped to "shell face" boundary conditions.
```
pbauman: triangle->quad

include/mesh/exodusII_io_helper.h
```diff
@@ -918,6 +994,15 @@ class ExodusII_IO_Helper::ElementMaps
...
+   * The remaining three sides are mapped to edge boundary conditions.
```
pbauman: three ->four

include/geom/face_quad4.h
jwpeterson: It's more an organizational thing than a number of new lines of code per new file thing.  There is one header per geometric element type at the moment, I'd like to keep it that way...

roystgnr: Agreed with @jwpeterson on this one.

Although "number of new lines of code" is good to keep an eye on too.  In hindsight there are a whole slew of methods that I should have put into an ElemTools:: namespace rather than the Elem class...

dknez:
@jwpeterson can you confirm whether you're planning to go ahead with the merge (and to add the h/C files)? There's no major rush on this, but I just wanted to check if you were planning to do it, or if I should look into it. Thanks.

jwpeterson:
Yes, planning to do it.

dknez:
OK, thanks.

jwpeterson:
Replaced by #948.

PR #945, opened 2016-05-23
get_eigen*() should be indexed by dof_id_type

roystgnr:
Not by unsigned int.

I don't actually know whether SLEPc supports this or will truncate
anything above 2^32.

I also don't know whether there are any algorithms which would make
querying the five billionth eigenvalue a computationally sane thing to
do.

I also don't know whether there are any use cases which would make
querying the five billionth eigenvalue a reasonable scientific or
engineering question.

But I do know that there was a libmesh_override error when we built
with a 64-bit-enabled SLEPc, and this should fix it in the most
aspirationally appropriate way.

### DIFF
--- a/include/solvers/eigen_solver.h
+++ b/include/solvers/eigen_solver.h
@@ -199,14 +199,14 @@ public:
    * Returns the \p ith eigenvalue (real and imaginary part),
    * and copies the \ ith eigen vector to the solution vector.
    */
-  virtual std::pair<Real, Real> get_eigenpair (unsigned int i,
+  virtual std::pair<Real, Real> get_eigenpair (dof_id_type i,
                                                NumericVector<T> & solution) = 0;
 
   /**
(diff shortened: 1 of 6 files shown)


jwpeterson:
Thanks, I don't think I have ever tested 64-bit indices with SLEPc turned on before.  Do you have a buildbot target that does this?

roystgnr:
On Mon, 23 May 2016, John W. Peterson wrote:

> Thanks, I don't think I have ever tested 64-bit indices with SLEPc turned on before. Do you have a buildbot target that does this?

Yup.  Continuous integration with 64-bit indices and SLEPc is now a thing!

Although, it was an accidental thing.  I had to rejigger module loading configurations to comport with the setup on the new build server, I didn't think to omit slepc from the lists, and when builds started failing I realized it would be just as fast to build the missing slepc modules as to remove the load commands.

PR #946, opened 2016-05-23
Assert valid boundary IDs

roystgnr:
This should be a more reliable way to catch some of the inconsistency bugs that I fixed in the examples in #943

### DIFF
--- a/include/mesh/mesh_tools.h
+++ b/include/mesh/mesh_tools.h
@@ -404,6 +404,12 @@ void libmesh_assert_valid_amr_interior_parents (const MeshBase & mesh);
 void libmesh_assert_connected_nodes (const MeshBase & mesh);
 
 /**
+ * A function for verifying that boundary condition ids match
+ * across processors.
+ */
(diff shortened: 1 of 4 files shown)


roystgnr:
IMHO we don't need to bother backporting this; it's not a fix in our code but rather a new feature to help users fix their codes.

jwpeterson:
Looks like it may have uncovered a bug in MOOSE?
```
materials/stateful_prop.test_older_mpi_threads:
Assertion `mesh.comm().semiverify (elem ? &bcids : nullptr)' failed.
[0] ../src/mesh/mesh_tools.C, line 1314, compiled May 23 2016 at 15:12:46
```
Let me know if there's a more or less "obvious" fix, otherwise we can dig down into it further...

roystgnr:
I can replicate this, at least.

friedmud:
I suspect that the fact that it's a "stateful_prop" test has nothing to do
with it... I bet a bunch of tests would have failed if it would have made
it to the `parallel` set of tests...

PR #947, opened 2016-05-23
Copy edge boundary conditions in all_first_order()

roystgnr:
Previously we were only preserving node and side conditions.

This actually turned out to be a red herring w.r.t. the bug I was
hunting, but it's nice to fix it too.

### DIFF
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -504,6 +504,15 @@ public:
                          const unsigned short int side,
                          std::vector<boundary_id_type> & vec_to_fill) const;
 
+  /*
+   * Copy boundary ids associated with old_elem (but not its nodes)
+   * from old_boundary_info (which may be this) into this boundary
+   * info, associating them with new_elem.
+   */
+  void copy_boundary_ids (const Elem * const new_elem,
+                          const Elem * const old_elem,
+                          const BoundaryInfo &old_boundary_info);
+
   /**
    * Returns a side of element \p elem whose associated boundary id is
    * \p boundary_id if such a side exists.
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -509,9 +509,9 @@ public:
    * from old_boundary_info (which may be this) into this boundary
    * info, associating them with new_elem.
    */
-  void copy_boundary_ids (const Elem * const new_elem,
+  void copy_boundary_ids (const BoundaryInfo &old_boundary_info,
                           const Elem * const old_elem,
-                          const BoundaryInfo &old_boundary_info);
+                          const Elem * const new_elem);
 
   /**
    * Returns a side of element \p elem whose associated boundary id is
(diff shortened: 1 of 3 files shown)


jwpeterson:
:+1:

dknez:
So now we need to update this PR to copy shellfaces too, right?

roystgnr:
Right, but honestly I'm wondering if we need to rethink how we handle boundary conditions.  Those "raw" boundary condition calls were a hack and now it's starting to show.  Something like a BoundaryInfo::copy_boundary_conditions(old_elem, new_elem) would be a better fit for here, for example, and then when adding new boundary condition types we'd only have to alter BoundaryInfo rather than finding every BoundaryInfo::raw* user.

Perhaps a serialization and unserialization function as well?  One thing I missed when looking over #948 was parallel_elem.C - there's a good chance that DistributedMesh and even ReplicatedMesh will lose shellface BCs because we're not copying them with the others.

dknez:
> Right, but honestly I'm wondering if we need to rethink how we handle boundary conditions.  Those "raw" boundary condition calls were a hack and now it's starting to show.  Something like a BoundaryInfo::copy_boundary_conditions(old_elem, new_elem) would be a better fit for here, for example, and then when adding new boundary condition types we'd only have to alter BoundaryInfo rather than finding every BoundaryInfo::raw* user.

Yeah, I agree. The current implementation could use some cleaning up. BoundaryInfo::copy_boundary_conditions sounds like a good idea to me.

roystgnr:
Rebased and refactored.  Hopefully this is good to merge now.  I'll put the parallel_elem.C fix into another PR.

dknez:
Looks good to me.

Beyond this PR, are there any other updates/changes to BoundaryInfo that you have in mind?

roystgnr:
Depends; do you have any other changes in mind?  ;-)

If BoundaryInfo is going to remain stable for a couple years then I'll just add the shellface (un)serialization directly to ParallelElem; if not then I'll want to factor that out too.

dknez:
I don't have any other changes in mind :) The shellface stuff was all I needed.

include/mesh/boundary_info.h
```diff
@@ -504,6 +504,15 @@ class BoundaryInfo : public ParallelObject
                          const unsigned short int side,
                          std::vector<boundary_id_type> & vec_to_fill) const;
 
+  /*
+   * Copy boundary ids associated with old_elem (but not its nodes)
+   * from old_boundary_info (which may be this) into this boundary
+   * info, associating them with new_elem.
+   */
+  void copy_boundary_ids (const BoundaryInfo &old_boundary_info,
```
jwpeterson: Need space on both sides of ampersand.

src/mesh/boundary_info.C
```diff
@@ -1285,6 +1285,37 @@ void BoundaryInfo::raw_boundary_ids (const Elem * const elem,
 
 
 
+void BoundaryInfo::copy_boundary_ids
+  (const BoundaryInfo &old_boundary_info,
```
jwpeterson: Space on both sides of ampersand, also this will be reindented by the reindent script.

PR #948, opened 2016-05-23
Shell elements (rebased and with PR comments addressed)

jwpeterson:
Replaces #944.


dknez:
Looks good, thanks for cleaning the PR up! It's ready to merge as far as I'm concerned.

dknez:
I'll go ahead and merge this, unless there's any objection?

roystgnr:
First concern - #947 fixes a bug in master, which is almost certainly reintroduced in a slightly different form here.  When we add a new boundary condition type we need to make sure it gets copied whereever necessary.

On the other hand, looking over the situation again, it's clear that #947 isn't a complete fix (because I missed the same bug elsewhere in the same file) or the right fix (because if I'd anticipated the possibility of adding more boundary condition types, then I'd have refactored to make that addition less bug-prone).  So we ought to commit this as-is, and then I'll fix #947 after rebasing.

Second concern - what mesh formats are we writing these new boundary conditions out to?  Just ExodusII?  Possibly Nemesis??  It appears to currently be fairly trivial to create a libMesh mesh which cannot be correctly saved to *any* file format - first use h refinement or p > 3 to lose ExodusII_IO support, then add an edge boundary condition or shellface boundary condition to lose XdrIO support.

That shouldn't block merging this PR either (after all, it's already a flaw in the edge BC case) but it feels like we're just digging the hole deeper.

pbauman:
> use h refinement or p > 3 to lose ExodusII_IO support

Uh... I see the `p > 3`, but all my AMR images were made writing out using ExodusII and plotting in ParaView... is there something I should know about? Or did you mean that the mesh hierarchy is lost?

roystgnr:
I meant that the mesh hierarchy was lost.  Nick wasted days (weeks?) at one point before discovering that the reason his restarts were turning to garbage was that he was trying to use .exo restart files on an adapted mesh.

dknez:
> On the other hand, looking over the situation again, it's clear that #947 isn't a complete fix (because I missed the same bug elsewhere in the same file) or the right fix (because if I'd anticipated the possibility of adding more boundary condition types, then I'd have refactored to make that addition less bug-prone). So we ought to commit this as-is, and then I'll fix #947 after rebasing.

Yeah, I think we should commit as-is and then fix #947.

> Second concern - what mesh formats are we writing these new boundary conditions out to? Just ExodusII? Possibly Nemesis??

I've only tested ExodusII. I'm not sure about Nemesis.

> It appears to currently be fairly trivial to create a libMesh mesh which cannot be correctly saved to any file format - first use h refinement or p > 3 to lose ExodusII_IO support, then add an edge boundary condition or shellface boundary condition to lose XdrIO support.

These are valid concerns. But I see this PR as a step in the right direction, in that it improves our handling of ExodusII shell meshes (before we had to have a "hack" to deal with shell elements). My preference would be to merge the PR as-is and then try to improve on these other issues over time.

roystgnr:
Agreed.

roystgnr:
Could we assert that shellface < 2 when adding shellface BCs?

dknez:
Yes, I think that's appropriate.

PR #950, opened 2016-05-24
adjoints_ex3 seems to need PETSc to converge

roystgnr:
We really ought to look into this, but for now let's just require it.


jwpeterson:
I wonder if it ever worked with the humble old Laspack solvers...

PR #951, opened 2016-05-24
Add --enable-cxx11-required configure option.

jwpeterson:
If enabled, this option throws an error at configure time if the compiler does not support a particular subset of C++11. In an ideal world I would have named this option differently, but autoconf only supports user-defined command line options that start with `--{enable,disable}` and `--{with,without}`.

I tested this on compilers which both support and don't support C++11 and it seems to work fine.  Our goal is to start using this in the libmesh build script for MOOSE, so that users are notified ASAP if their compiler does not support C++11.

### DIFF
--- a/configure
+++ b/configure
@@ -1116,6 +1116,7 @@ enable_fortran
 with_fc
 with_f77
 enable_cxx11
+enable_cxx11_required
 enable_sanitize
 enable_glibcxx_debugging
 enable_glibcxx_debugging_cppunit
@@ -1901,6 +1902,7 @@ Optional Features:
                           speeds up one-time build
   --disable-fortran       build without Fortran language support
   --disable-cxx11         build without C++11 support
+  --enable-cxx11-required Error if compiler does not support C++11
   --enable-sanitize="opt dbg devel prof oprof"
                           turn on sanitizer flags for the given methods
   --disable-glibcxx-debugging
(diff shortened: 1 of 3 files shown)


friedmud:
Any chance we could have an optional "error string" that we pass in that would be printed if this check fails?

I would really like to let MOOSE users know the minimum compilers we support if this check fails.

jwpeterson:
I think the right approach there is to parse the output of configure (or config.log) in `update_and_rebuild_libmesh.sh` and if it says
```
configure: error: *** A compiler with support for C++11 language features is required.
```
Then print something about our minimum compiler versions...  IMO it doesn't make sense to do something like --enable-cxx11-required="A custom error string if that fails."

permcody:
We don't need the minimum compilers here. However we do need them on our site! How about I put together a "minimum system requirements" section on the Getting Started tab?

Also, sorry for highjacking a libmesh thread...

friedmud:
@permcody I didn't want to hardcode the minimum compilers... I was just thinking that we could pass an error string in when specifying `--enable-cxx11-required`... but I agree that it would be weird.

@jwpeterson 's suggestion works fine too.  Or is there a way to return a specific return code (integer) from configure that can signify what failed?  That would make it much easier.

jwpeterson:
> Or is there a way to return a specific return code (integer) from configure that can signify what failed? That would make it much easier.

Yes, looks like `AC_MSG_ERROR` can specify an exit status code.  I'll pick a random number (9999?) and add that to this PR.

roystgnr:
:+1:

PR #953, opened 2016-05-25
XDA writer should not write trailing whitespace

jwpeterson:
Also fixed some confusingly formatted code in XdrIO.


roystgnr:
:+1:

I assume this is to prevent the trailing whitespace from annoying you in diffs of xda files, not because you're needing to save a few precious bytes?

jwpeterson:
> I assume this is to prevent the trailing whitespace from annoying you in diffs of xda files,

Yes.  In a personal repo I have a code that writes xda files which happen to be checked in, and since my trailing whitespace removing git pre-commit deletes those, they would always show up as a diff...

PR #954, opened 2016-05-25
Fix example compilation with --enable-complex

roystgnr:
This looks like the last barrier in between us and a successful "make check" there.


jwpeterson:
:+1: will backport this one too.

PR #955, opened 2016-05-25
reduced_basis_ex7 fixes

roystgnr:
Before, complex-valued PETSc-enabled libMesh builds would die on reduced_basis_ex7 unless you manually set LIBMESH_OPTIONS to a direct solver.

Now, complex-valued PETSc-enabled libMesh builds will die on reduced_basis_ex7 unless you configured PETSc with MUMPS.

This should be enough to get my BuildBot libMesh --enable-complex stage working, which is what I care about, but suggestions for further improvement would be welcome.

### DIFF
--- a/examples/reduced_basis/reduced_basis_ex7/reduced_basis_ex7.C
+++ b/examples/reduced_basis/reduced_basis_ex7/reduced_basis_ex7.C
@@ -159,7 +159,7 @@ int main (int argc, char** argv)
                        << "Try running with -ksp_type preonly -pc_type lu instead.\n"
                        << "********************************************************************************"
                        << std::endl;
-          return 1;
+          return 77;
         }
 #endif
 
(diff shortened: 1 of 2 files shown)


jwpeterson:
I take it this does it not work with just straight `-pc_type lu`?  Otherwise I don't see the need to introduce a MUMPS dependency in addition to the PETSc dependency?

jwpeterson:
Also I am ready to backport this to 1.0.0 if you think that's a good idea.

jwpeterson:
> I take it this does it not work with just straight -pc_type lu?

Ah, I see now that you were more worried about it working in parallel... 

Maybe a `run_example_serial()` function in run_common.sh?  This would ignore the user's $LIBMESH_RUN settings but otherwise be the same as the existing `run_example()`.

roystgnr:
The trouble with ignoring LIBMESH_RUN is that we don't just use that for MPI.  It'd be annoying to be debugging and go through the "LIBMESH_RUN='libtool --mode=execute gdb --args'" rigamarole only to have the example outright ignore you.  You can use LIBMESH_RUN to do a fair "make check" on systems where the admins don't like you running jobs on the head/compile nodes, too.

Oh, and there used to be some MPI implementations that would crash on you if you didn't use "mpirun -np 1" for serial jobs.  I don't think that's ever the case anymore, but who knows.

roystgnr:
I wouldn't bother backporting this one.  There was already a crummy-but-tolerable workaround (do "make check" in serial with LIBMESH_OPTIONS specifying LU, or with LIBMESH_OPTIONS specifying MUMPS LU), and it's only a flaw in the example setup, not in library code.

PR #956, opened 2016-05-25
Pack shellface bcids

roystgnr:
The shellface stuff isn't covered by anything in our unit tests or examples, is it?  So I'm pretty sure I'm fixing bugs here, but who knows?

PR #957, opened 2016-05-26
using mesh with SHELL4 instead of QUAD4 elements for example misc12 +…

svallaghe:
… minor fixes


examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
```diff
@@ -203,77 +211,81 @@ int main (int argc, char ** argv)
...
+          if (w_dof >= system.get_dof_map().first_dof() &&
+              w_dof < system.get_dof_map().end_dof())
+            w = system.current_solution(w_dof);
+        }
+      system.comm().sum(w);
 
 
-  Number w_C_bar = -E*h*w/q;
-  const Real w_C_bar_analytic = 164.24;
+      Real w_C_bar = -E*h*w/q;
```
roystgnr: Nope.  This is just going to get us breaking with --enable-complex again.

When you fix conflicts, it's important to integrate changes, not revert them.  ;-)

examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
```diff
@@ -756,28 +777,50 @@ void assemble_shell (EquationSystems & es,
 
         } // end of the quadrature point qp-loop
 
+      if (distributed_load)
+      {
+        //Loop on shell faces
+        for (unsigned int shellface=0; shellface<2; shellface++)
+          {
+            std::vector<boundary_id_type> bids;
+            mesh.get_boundary_info().shellface_boundary_ids(elem,shellface,bids);
```
roystgnr: This is exactly what I was hoping for - test coverage!  I'll hold off on merging #956 until I can rebase off these changes and see what happens.

examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
svallaghe: I didn't mean to revert anything, I guess I did something wrong when incorporating my changes.
Thanks for spotting.

src/geom/reference_elem.C
```diff
@@ -247,11 +247,20 @@ const Elem & get (const ElemType Type)
...
+    BaseType = QUAD4 ;
+
   init_ref_elem_table();
 
-  libmesh_assert (ref_elem_map[Type] != libmesh_nullptr);
+  libmesh_assert (ref_elem_map[BaseType] != libmesh_nullptr);
 
-  return *ref_elem_map[Type];
+  return *ref_elem_map[BaseType];
 }
```
roystgnr: Is this safe?  Users might reasonably expect to be able to ```dynamic_cast<TriShell3 &>(ReferenceElem::get(TRISHELL3))``` but attempting to do so would fail.

src/reduced_basis/rb_parametrized.C
```diff
@@ -100,7 +100,7 @@ void RBParametrized::initialize_parameters(const RBParameters & mu_min_in,
         Real min_val = *std::min_element(it->second.begin(), it->second.end());
         Real max_val = *std::max_element(it->second.begin(), it->second.end());
 
-        libmesh_assert(min_val > max_val);
+        libmesh_assert(min_val <= max_val);
```
roystgnr: As long as we have to fix this anyways let's use libmesh_assert_less_equal()?

examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
```diff
@@ -89,6 +90,12 @@ int main (int argc, char ** argv)
...
 #endif
 
+  // Read the "distributed_load" flag from the command line
+  GetPot command_line (argc, argv);
+  int distributed_load = 0;
+  if (command_line.search(1, "-distributed_load"))
+    distributed_load = command_line.next(distributed_load);
+
   // Create a mesh distributed across the default MPI communicator.
   Mesh mesh (init.comm(), 3);
```
roystgnr: The changes to cylinder.exo are merely to add shell boundary ids?

roystgnr:
The only critical problem is the Number->Real regression, but if you could check out the rest while you're having to change something anyway I'd appreciate it.

Thanks!

examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
jwpeterson: Should we update all the examples to use GNU style `--long-option-names`?  I suppose it's a matter of personal preference, but at least in the library I think we do this.

src/geom/reference_elem.C
jwpeterson: This is probably OK, other than the spurious spaces before semi-colons.

I think the long term fix here is to generate new entries in the reference_elem.data file for the new element types.  These entries include boundary conditions, so they would be slightly different than the existing QUAD4 and TRI3 types.  However, I think these are generated from the files in the reference_elements/{2D,3D} directories, and we don't yet support writing the new shell types to XDA files... so that would need to be fixed first.

roystgnr: Gah, good point.  This isn't a big enough problem to be worth that much effort to fix right away.

examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
```diff
@@ -397,14 +415,14 @@ void assemble_shell (EquationSystems & es,
       // First compute element data at the nodes
       std::vector<Point> nodes;
       for (unsigned int i=0; i<elem->n_nodes(); ++i)
-        nodes.push_back(elem->node_ref(i));
+        nodes.push_back(elem->reference_elem()->node_ref(i));
```
jwpeterson: So, is this the line of code that prompted you to update `reference_elem()`?  Does `elem->node_ref()` just not work for the shell element types?

examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
svallaghe: yes the mesh cylinder.exo was modified:
- it now has SHELL4 elements instead of QUAD4 elements
- it has two new sidesets for the top and bottom of the surface (using shellface ids)

examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
svallaghe: I realized that you have to pass points on the reference element to fe->reinit().
I'm a bit puzzled regarding why it was giving a correct solution before.

svallaghe:
I'll update this PR tomorrow morning (CET)

dknez:
Looks good to me, I'll merge soon, unless there's any objection.

PR #959, opened 2016-05-26
Add all_first_order option to meshtool

### LINKED ISSUE #958: Add all_first_order option to meshtool
The `meshtool` executable built with libmesh already has `-2` and `-3` options to convert elements to second (or highest possible order). I'd like to add a `-1` option to run `all_first_order` on the mesh.

dschwen:
Adds the `-1` option to `meshtool`. Probably works.
JK, I tested it on a QUAD8 mesh and it produced a QUAD4 mesh.

Closes #958


jwpeterson:
:+1: I'm fairly confident this works for all elements that `mesh.all_first_order()` works for.  Unfortunately I don't know if it works for all of them.

PR #960, opened 2016-05-27
Update to MeshFunction::clone()

dknez:
Call clone->init() inside the clone method if the original object was initialized. This is important when you want to use a MeshFunction with System::project(), because project makes a clone of the MeshFunction and we need to make sure that the clone is initialized.

### DIFF
--- a/src/mesh/mesh_function.C
+++ b/src/mesh/mesh_function.C
@@ -166,9 +166,13 @@ MeshFunction::clear ()
 
 UniquePtr<FunctionBase<Number> > MeshFunction::clone () const
 {
-  return UniquePtr<FunctionBase<Number> >
-    (new MeshFunction
-     (_eqn_systems, _vector, _dof_map, _system_vars, this));
+  FunctionBase<Number> * mf_clone =
+    new MeshFunction(_eqn_systems, _vector, _dof_map, _system_vars, this);
+
+  if(this->initialized())
+    mf_clone->init();
+
+  return UniquePtr< FunctionBase<Number> >(mf_clone);
 }
 
 
--- a/tests/systems/systems_test.C
+++ b/tests/systems/systems_test.C
@@ -8,6 +8,8 @@
 #include <libmesh/mesh.h>
 #include <libmesh/mesh_generation.h>
 #include <libmesh/serial_mesh.h>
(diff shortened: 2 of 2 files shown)


dknez:
Hmm, this compiles for me. Anyone see what the issue is?

roystgnr:
```
./include/libmesh/unique_ptr.hpp: In member function ‘virtual boost::unique_ptr<libMesh::FunctionBase<double>, boost::default_delete<libMesh::FunctionBase<double> > > libMesh::MeshFunction::clone() const’:
./include/libmesh/unique_ptr.hpp:321:5: error: ‘boost::unique_ptr<T, D>::unique_ptr(boost::unique_ptr<T, D>&) [with T = libMesh::FunctionBase<double>; D = boost::default_delete<libMesh::FunctionBase<double> >]’ is private
     unique_ptr(unique_ptr&);
     ^
../src/mesh/mesh_function.C:176:10: error: within this context
   return mf_clone;
```

Looks like a "UniquePtr is auto_ptr" vs "UniquePtr is really unique_ptr" issue?  That's actually a bit tricky, if I'm understanding it correctly.  IIRC the proper fix in C++11 is to make the transfer explicit with std::move?  But that's not a C++03 function.

roystgnr:
I guess you could just create it and init it as a raw pointer and only turn it into UniquePtr in the return call?

jwpeterson:
> On May 26, 2016, at 9:51 PM, roystgnr <notifications@github.com> wrote:
> 
> I guess you could just create it and init it as a raw pointer and only turn it into UniquePtr in the return call?
> 

Yea that's the current workaround.  At worst you can release() it into a newly created UniquePtr in the return statement.

dknez:
This looks OK to merge now to me.

roystgnr:
We should also backport at least the mesh_function.C changes to 1.0.0

jwpeterson:
> On May 27, 2016, at 8:16 AM, roystgnr <notifications@github.com> wrote:
> 
> We should also backport at least the mesh_function.C changes to 1.0.0
> 

OK, I will take care of it.

jwpeterson:
MeshFunction changes backported in c672bc1.

PR #961, opened 2016-05-27
Speed up make check

roystgnr:
Alhough "make check" runs at a decent clip with PETSc on 4 cores, it can be interminable when BuildBot forces it to use weaker solvers and run in serial.  Tweaking example parameters (and disabling one when we're using Eigen's sparse solver) should fix that.


roystgnr:
This shouldn't be backported; the changes are all just for runtime speedup, not failure fixes.

PR #963, opened 2016-05-27
Remove overzealous assert from set_p_level()

roystgnr:
This obviously fails when p coarsening back down to minimal p.

Thanks to @vikramvgarg for catching it.


roystgnr:
This'll be another one that requires backporting.  @jwpeterson, let me know when you want to take a break from that.  The workload seems to have passed "I owe you a beer" a while back.

jwpeterson:
> This'll be another one that requires backporting.

Done!

PR #967, opened 2016-05-28
Speed up stitch meshes by avoiding unnecessary PointLocator calls.

dknez:


dknez:
This change is only a minor change since it only affects how we set "hmin" in ReplicatedMesh::stitch_meshes, and that value is only used to set a tolerance. As a result, I'll go ahead and merge.

PR #972, opened 2016-05-30
Remove overzealous asserts in FEMContext

pbauman:
When we have the case of subdomain-restricted variables, the dof_indices
only get populated with indices from the variables currently active
on the element. Thus, if the variable var for which we are querying
was inserted in the System after variables that are not active on
the current element, then this assert will fail. The more appropriate
assert actually happens in the get_dof_indices(var) call, so we
just remove these asserts entirely.

### DIFF
--- a/src/systems/fem_context.C
+++ b/src/systems/fem_context.C
@@ -209,7 +209,6 @@ template<typename OutputType,
 void FEMContext::some_value(unsigned int var, unsigned int qp, OutputType & u) const
 {
   // Get local-to-global dof index lookup
-  libmesh_assert_greater (this->get_dof_indices().size(), var);
   const unsigned int n_dofs = cast_int<unsigned int>
     (this->get_dof_indices(var).size());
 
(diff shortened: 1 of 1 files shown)


pbauman:
FWIW, I think this would be appropriate to back port to 1.0.0.

roystgnr:
Damn.  Those are the wrong asserts.  And they've clearly been wrong since at least our move to git in 2012.  And the bug even survived the translation from "dof_indices.size()" to "get_dof_indices().size()", presumably thanks to the nice perl scripts we had for automating that.

And you're right that they're also (now) redundant in any case, and that they ought to be expected to break things in the subdomain restrictions case.

This should definitely be back ported.

jwpeterson:
Cherry-picked as fdb5dba.

Speaking of which, let me know when we can tag 1.0.0-rc2.  Maybe some time this week?  We've back-ported quite a few fixes since the last RC.

roystgnr:
Sorry about the delays on 1.0.0-rc2.  Our buildbot server went down last weekend and it's just getting back up to speed.  But the good news is that we're already passing tests like the complex cases that we'd been failing as of last Wednesday, and looking through the rest of that last build the only clearly unsolved issue I see is a problem in the reduced basis code interaction with 64-bit indices.  Let me take a swipe at that and then we can put out rc2.

PR #973, opened 2016-05-30
Updated XdrIO so that it supports edge and shell-face boundary IDs

dknez:


examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
```diff
@@ -96,10 +96,18 @@ int main (int argc, char ** argv)
...
+  {
+    Mesh mesh (init.comm(), 3);
+
+    // To confirm that both ExodusII and Xdr formats work for shell
+    // meshes, we read in cylinder.exo, then write out cylinder.xdr,
+    // then read in cylinder.exo again below and use that for the rest
+    // of the example.
+    mesh.read("cylinder.exo");
+    mesh.write("cylinder.xdr");
+  }
```
roystgnr: I like this.

include/mesh/xdr_io.h
```diff
@@ -161,6 +161,23 @@ class XdrIO : public MeshInput<MeshBase>,
   const std::string & polynomial_level_file_name() const { return _p_level_file; }
   std::string &       polynomial_level_file_name()       { return _p_level_file; }
 
+  /**
+   * @return true if the current file has an XDR/XDA version that
+   * matches or exceeds 0.9.2.
+   */
+  bool exceeds_version_0_9_2() const;
+
```
roystgnr: Hmm... now that we're putting my stupid variable name into an API, perhaps we should make it less stupid?  "meets_or_exceeds_version_" is quite a mouthful, but accuracy is more important than brevity.  "version_at_least_"?

roystgnr:
Very nicely factored.  If you decide on a better name for the helper method, switch it and merge, if not then it's still fine to merge as-is.

dknez:
OK, I'll push an update shortly with:
- a new unit test added to the mesh/boundary_info.C test which makes sure that we can add edge IDs, write an xdr file, and then read back in and preserve the edge IDs
- the function name update that you proposed: version_at_least_

tests/mesh/boundary_info.C
```diff
@@ -20,41 +21,35 @@ public:
...
-                                        2, 2,
-                                        0., 1.,
-                                        0., 1.,
-                                        QUAD4);
-  }
-
 public:
   void setUp()
   {
-    this->build_mesh();
```
jwpeterson: @roystgnr Is there some reason that we `new`/`delete` meshes in these unit tests?  It looks like an idiom that has been copy/pasted around for a while, but I can't remember if there was a reason for it...

roystgnr: Was it to make it easier for us to build the same type of mesh again and again to use it for several tests?  I don't recall.

dknez: I changed it because I wanted a different mesh for the new test that I added...

pbauman: Recall too that everything is wiped between tests, so if you want a mesh for a test, you need it in `setUp()` (and the corresponding cleanup in `tearDown()`.

jwpeterson: Ah yes, `mixed_dim_mesh_test.C` factored out `build_mesh()` so it could be called multiple times, that must be it.

Issue #974, opened 2016-05-31
Iterating over semi local elements fails on refined meshes


dschwen:
Incrementing the iterator obtained via `.semilocal_elements_begin()` can trigger an `Assertion \`start_elem->active()' failed. `

dschwen:
@jwpeterson has proposed a fix for this: adding

```
this->_predicates.push_back(new active<T>)
```

to the `SemiLocal` multi-predicate. This seems to fix teh issue I observe.

jwpeterson:
Will make PR momentarily...

roystgnr:
Semilocal elements are not necessary active elements, and we may already have other code in which that assumption is important.  The fix here should probably be two-fold: create a semilocal_active_elements_foo() pair of iterators if that's what you need, and fix the iteration in the semilocal_elements_foo() iterators too.

roystgnr:
Huh.  We do rely on find_point_neighbors(), which does assert active().  This seems like the underlying bug...

jwpeterson:
OK, so you think that `find_point_neighbors()` shouldn't assert this?

roystgnr:
No, find_point_neighbors() was definitely written to work for active elements only.  Perhaps what we need to be doing is, for inactive elements, looping over the children and testing is_semilocal() on each of them?

roystgnr:
Yeah, that sounds like the right fix - iff any descendant of mine is semilocal, then so am I.  Could be slow as hell to iterate through the low level ancestors... but I suppose the number of remote ancestors decreases exponentially with level.  And I don't think there's any other safe way to handle it.

jwpeterson:
OK, is there any reason to have #974 as a temporary quick fix until someone implements this loop over children checking for semi-locality?  Otherwise I will just close that...

PR #975, opened 2016-05-31
We can't check whether an element is_semilocal() unless it is active().

jwpeterson:
This is because Elem::is_semilocal() calls Elem::find_point_neighbors(),
which asserts that the Elem is active.  Note that we had a check for
"not_subactive" but this is apparently not the same thing as being
"active".

Refs #974.


jwpeterson:
@roystgnr any yea/nay vote on this as a temporary fix until `Elem::is_semilocal` can be refactored?

roystgnr:
What's the use case?  If there's something breaking in the library, then yeah, merge this until we can do something better.  If you need semilocal active iterators at the app level, I think what would be better would be to add semilocal_active_ iterators and use those so that you don't have to worry about the behavior changing under the hood at a later date.

jwpeterson:
Nothing breaking in the library, no, but fixes an inconsistency for anyone who today writes a loop over semi_local iterators on a refined mesh.  They'll get either a segfault or assert as the code stands now.  I think @dschwen fixed his issue by looping over active elements and checking whether each is_semilocal, but not before he came to our cube complaining about libmesh bugs.

dschwen:
Mister complainsy pants.

roystgnr:
Okay, it looks like we're not using this in the library at all?  And anyone using it outside of the library is flirting with disaster?  Then yeah, let's change the semilocal_elements iterators into active_semilocal_elements iterators, let's drop the semilocal_elements iterators entirely for now, and let's reintroduce semilocal_elements iterators once we've fixed is_semilocal() and can accommodate them.

jwpeterson:
> Then yeah, let's change the semilocal_elements iterators into active_semilocal_elements iterators, let's drop the semilocal_elements iterators entirely for now,

`semilocal_elements_begin()`/`end()` are currently being used in MOOSE so we can't drop them outright without also patching MOOSE.  How about for now we just make those existing functions return the same thing as the new `active_semilocal_elements_begin()`/`end()` until they can be patched?

roystgnr:
Sounds good to me.  Thanks!

PR #978, opened 2016-05-31
Trilinos fails to give the right values for adjoints_ex5

roystgnr:
For now we'll skip this test so we can still use "make check" to keep
an eye out for regressions elsewhere.

Long-term we clearly need to figure out WTF is wrong with our Trilinos
adjoint solves.


jwpeterson:
:+1:

jwpeterson:
Backported to 1.0.0 in 579b508.

PR #980, opened 2016-05-31
Add configure/unit tests and macro for std::make_unique

jwpeterson:
`std::make_unique` is C++14 only, but some compilers might (?) support it with `-std=c++11` and eventually will support it with no flags at all.  This PR tests for those cases, as well as a C++11-compatible workaround, and adds the `libmesh_make_unique` macro which expands to the appropriate version.  Also added a unit test to ensure that the macro was actually doing the right thing.

Note: we will eventually want to add a `--enable-cxx14` configure flag as well as a test for setting it, but that is a PR for another time.  

Refs #655.


include/base/auto_ptr.h
```diff
@@ -65,6 +65,37 @@ using boost::unique_ptr;
...
+template<typename T, typename... Args>
+std::unique_ptr<T> make_unique(Args&&... args)
+{
+  return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+}
+}
+#define libmesh_make_unique make_unique
+
+#else
+#define libmesh_make_unique ; Error: You cannot use libmesh_make_unique with this compiler!
```
friedmud: Is this a good error?  Do you mean to use `#error` here?  Won't this just result in a weird compilation error as it tries to compile `Error: You cannot use libmesh_make_unique with this compiler!` ?

include/base/auto_ptr.h
```diff
@@ -65,6 +65,37 @@ using boost::unique_ptr;
...
+#ifdef LIBMESH_HAVE_CXX14_MAKE_UNIQUE
+#define libmesh_make_unique std::make_unique
+
+#elif LIBMESH_HAVE_CXX11_MAKE_UNIQUE_WORKAROUND
+namespace libMesh
+{
+template<typename T, typename... Args>
+std::unique_ptr<T> make_unique(Args&&... args)
+{
+  return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
```
friedmud: Really cool!

include/base/auto_ptr.h
roystgnr: Definitely don't want to use #error there - the code should error out only when that macro is used, not when it's defined.

jwpeterson: Yeah, @roystgnr and I chatted about ways to do this error message.  On the screen you will see something like this:
```
unique_ptr.C:64:24: note: expanded from macro 'my_make_unique'
#define my_make_unique ; Error: You cannot use libmesh_make_unique with this compiler!
                                          ^
```
if you use the new macro with a compiler that does not support C++11.  You can do something better in C++11, but this branch is for when you don't have C++11 😞

friedmud: Hmmm... I would say use `static_assert()` ( http://en.cppreference.com/w/cpp/keyword/static_assert ) but the whole point is that their compiler doesn't support C++11!

So is this the best we can do... just give a weird ass compile error?

We could use `libmesh_error()` but that would make it a runtime error...

friedmud: BTW: I think this will generate seriously weird compile errors if you try to use it as the argument to a function... or inside an `if()` statement.

But whatevs... I don't care that much... people with old compilers simply need to update!  :-)

friedmud: Sorry @jwpeterson ... my GitHub page hadn't updated to see your comment.

I think this is fine ;-)

roystgnr:
Looks good!  :+1:

PR #983, opened 2016-06-01
mesh stitcher needs to copy shellface_boundary_ids

svallaghe:


jwpeterson:
Fix looks good to me but I'll let @dknez merge.

Should we add an `Elem::is_shellface()` accessor to avoid hardcoding the types in if-statements like this?  There is some precedent for this with `Elem::infinite()` already...

src/mesh/serial_mesh.C
```diff
@@ -1271,6 +1271,18 @@ void ReplicatedMesh::stitching_helper (ReplicatedMesh * other_mesh,
               other_mesh->get_boundary_info().boundary_ids(other_elem, s, bc_ids);
               this->get_boundary_info().add_side(this_elem, s, bc_ids);
             }
+
+          // Copy shellface boundary ids
+          // This is only relevant for shell elements
+          if (other_elem->type()==TRISHELL3 || other_elem->type()==QUADSHELL4)
```
dknez: I would get rid of this "if" statement. Just copy any shellface ids that exist in the original mesh, there's no harm in that, and it's easier to maintain since otherwise if we add new shell types we'd have to remember to update this if-statement.

dknez:
> Should we add an Elem::is_shellface() accessor to avoid hardcoding the types in if-statements like this? There is some precedent for this with Elem::infinite() already...

That makes sense, but for now I suggest getting rid of the if-statement in this PR.

src/mesh/serial_mesh.C
roystgnr: Agreed.  Even if we add is_shellface() and we only want to allow shellface boundary ids there, the place to enforce that is in the BoundaryInfo methods which add shellface ids, not in every code that calls those methods.

dknez:
This was replaced by #984

PR #985, opened 2016-06-02
Handle NULL elements in EquationSystems::get_solution().

jwpeterson:
Previously, we were using:
```
mesh.n_elem()
```
to determine the global size of the parallel_soln vector, and
```
std::distance(_mesh.local_elements_begin(), _mesh.local_elements_end())
```
to determine its local size on each processor.  These two sets of
numbers are not consistent when there are NULL Elems in the Mesh
because the former counts the NULL entries while the latter does not.
This would happen if e.g. get_solution() was called prior to calling
MeshBase::contract().  The fix is to just compute the local sizes
given the global size, without relying on Mesh iterators.  Values for
contracted elements simply show up as zeros in the solution vector
which gets built, it is up to the individual I/O formats to handle
this correctly (Exodus should be able to).

### DIFF
--- a/src/systems/equation_systems.C
+++ b/src/systems/equation_systems.C
@@ -900,11 +900,6 @@ void EquationSystems::get_solution (std::vector<Number> & soln,
 
   libmesh_assert_equal_to (ne, _mesh.max_elem_id());
 
-  // Get the number of local elements
-  dof_id_type n_local_elems = cast_int<dof_id_type>
-    (std::distance(_mesh.local_elements_begin(),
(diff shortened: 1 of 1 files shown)


jwpeterson:
Hmm... this doesn't pass the MOOSE tests in parallel for me locally, we'll see if Civet agrees...might need to be more careful about only setting locally-owned values in `parallel_soln` or something.

jwpeterson:
OK, I fixed one new MOOSE test that was not robust in parallel for me and this now passes on my system, plus Civet agrees.

Issue #990, opened 2016-06-03
Add contains_point specializations for simple elements


dschwen:
Elements that are guaranteed to have flat sides, such as `TRI3` and `TET4` should have trivial geometric expressions to determine if a point is contained in them or not. Overloading `contains_point` for those should speed up this check (which is used heavily by `PointLocator`).

I'll push a draft for `TET3` that delivers a 2x speedup over the old method, which does BBOX + inverse map.

jwpeterson:
*Ass*?

jwpeterson:
I assume you mean `TRI3`... I was hoping it would be more than 2x.

friedmud:
As long as you're going down this path... I would LOVE to have `volume()`
specializations for these "simple" elements as well... it's crazy just how
much time we spend calling `volume()`

jwpeterson:
> As long as you're going down this path... I would LOVE to have `volume()`
> specializations for these "simple" elements as well... it's crazy just how
> much time we spend calling `volume()`

This has already been done.

friedmud:
Really?  I still see `volume()` showing up a lot (like 6-7%) in my timing of my new code.  Maybe I'll take a closer look at it next time...

jwpeterson:
> Really? I still see volume() showing up a lot (like 6-7%) in my timing of my new code. Maybe I'll take a closer look at it next time...

What type of element, and how often are you calling `volume()`?  Specializations of `volume()` for all the element types have been implemented, resulting in speedups of over 100x for some of the more expensive types.

friedmud:
Quad4 and Hex8... I am calling it quite a bit.  I'll pay more attention next time I'm doing profiling.

Anyway - I'll quit polluting this Issue.  This particular issue gets a 👍 from me!  Should really help out a few things I'm working on right now!  (Just today I added a massive set of `contains_point()` searches to my code...)

jwpeterson:
> Quad4 and Hex8

Those two are some of the most highly-optimized `volume()` comptations there are outside of the really simple ones like `TRI3` and `TET4`.  If they are still showing up in profiling, you might be better off caching values if possible... I don't think the actual calculations could be speeded up much more.

friedmud:
Thanks @jwpeterson .  I've considered caching them and I might just need to do that...

dschwen:
Looks more like a factor of 4 now.

jwpeterson:
[...]

PR #993, opened 2016-06-03
contains_point optimization for TRI3

dschwen:
This `contains_point` specialization on `Tri3` is about 6 times faster according to my tests. I'm putting this up for discussion because it changes the meaning of `tol` quite a bit. I'm also not sure how to appropriately handle the degenerate triangle cases. My out-of-plane test projects the point on a cross product. If that is zero I cannot test for proximity anymore (so I'm just returning false). This is not ideal, because for that test the meaning of `tol` is now the **opposite** (i.e. almost degenerate triangles reject all points instead of permitting _more_ points within `tol` of trianglke sliver).

Refs #990

### DIFF
--- a/src/geom/face_tri3.C
+++ b/src/geom/face_tri3.C
@@ -217,8 +217,8 @@ bool Tri3::contains_point (const Point & p, Real tol) const
   Point oop ( v1.cross(v2) );
 
   // project moved test point onto out of plane component and bail
-  // if it is farther than tol (TODO: degenerate triangle!)
-  if ( std::abs(p0 * oop) > tol )
+  // if it is farther out of plane than tol.
+  if ( std::abs(p0 * oop) > tol || oop.norm_sq() < tol * tol)
     return false;
 #endif
 
@@ -228,16 +228,25 @@ bool Tri3::contains_point (const Point & p, Real tol) const
   const Real d12 = v1 * v2;
   const Real d22 = v2.norm_sq();
 
+#if LIBMESH_DIM == 3
+  // the denominator can only be 0 if v1 = v2, but then the cross product would be 0
+  // and we'd have bailed out already
+  const Real d = 1.0 / (d11 * d22 - d12 * d12);
+#else
   Real d = d11 * d22 - d12 * d12;
   if (d < tol)
     return false;
-
   d = 1.0 / d;
+#endif
 
+  // the in plane check implements the barycentric technique from
+  // http://www.blackpawn.com/texts/pointinpoly/
   const Real s1 = (d02 * d11 - d01 * d12) * d;
-  const Real s2 = (d01 * d22 - d02 * d12) * d;
+  if (s1 <= -tol)
+    return false;
 
-  return s1 > -tol && s2 > -tol && s1 + s2 < 1.0 + tol;
+  const Real s2 = (d01 * d22 - d02 * d12) * d;
+  return s2 > -tol && s1 + s2 < 1.0 + tol;
 }
 
 } // namespace libMesh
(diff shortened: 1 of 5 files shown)


dschwen:
This passes `moose/test` and `moose/modules/combined` on my machine.

jwpeterson:
Would you mind adding some unit tests?  I wouldn't worry too much about degenerate triangle cases.  They are not going to be useful for finite element calculations anyway, right?

jwpeterson:
:+1:

src/geom/face_tri3.C
```diff
@@ -203,4 +203,50 @@ std::pair<Real, Real> Tri3::min_and_max_angle() const
...
+  Point v1 ( this->point(1) - this->point(0) );
+  Point v2 ( this->point(2) - this->point(0) );
+
+#if LIBMESH_DIM == 3
+  // define out of plane vector
+  Point oop ( v1.cross(v2) );
+
+  // project moved test point onto out of plane component and bail
+  // if it is farther out of plane than tol.
+  if ( std::abs(p0 * oop) > tol || oop.norm_sq() < tol * tol)
```
benkirk: @dschwen, @jwpeterson, I think this line is triggering a failure for me in small triangles.

I have a pretty small triangle, but it's nearly equilateral:
```c++
 p=(x,y,z)=(   0.001,    0.005,        0)
 e=571
 node 0,   Node id()=275, processor_id()=0, Point=(x,y,z)=(0.000808805, 0.0047284,        0)
 node 1,   Node id()=376, processor_id()=0, Point=(x,y,z)=(0.0011453, 0.00472831,        0)
 node 2,   Node id()=371, processor_id()=0, Point=(x,y,z)=(0.000982249, 0.00508037,        0)
```
I'm pretty sure the test fails simply because v1, v2 are really small, oop is really small.  This isn't a sign of a bad triangle though.  What case are you trying to protect for? 

I think maybe we might need to define a local coordinate system with p0, p1, p2 scaled by the maxiumum bounding box dimension or something to put these in the x,y in [0,1] space or something.

Note a false failure can be triggered by adding

```c++
    a /= 5000.;
    b /= 5000.;
    c /= 5000.;
```

to the unit test tests/mesh/contains_point.C.  Under these conditions the element will fail to find its centroid.

Thoughts?

roystgnr: That tolerance argument is supposed to be a *relative* tolerance (implicitly, because we applied it on the master element), but with ```p0 * oop``` we're using it as an *absolute* tolerance... and I don't even understand the point of the ```oop.norm_sq()``` subcondition...

jwpeterson: It seems like all we should need is the "barycentric technique" described at `http://blackpawn.com/texts/pointinpoly`... there should not be any cross products, etc. required. I'll try and take a look tonight unless you are already working on that.

benkirk:
Thanks guys.  Looks like 

```c++
  // Out of plane check
  if (std::abs(triple_product(v2, v0, v1)) / std::max(dot00, dot11) > tol)
    return false;
```

Nicely handles the out of plane business in a relative way that still avoids square root / unit vector construction, which I assume was part of the motivation in speeding up this method to begin with.

jwpeterson:
> Nicely handles the out of plane business in a relative way that still avoids square root / unit vector construction, 

This is just off the top of my head, it might still have hidden issues. I did try some approaches involving `Point::unit()` but then you have to guard against zero vectors since we assert in that case. Better to avoid if possible.

> which I assume was part of the motivation in speeding up this method to begin with.

The base class `contains_point()` implementation has a full blown `FEInterface::inverse_map` map call in it, so I think the main motivation was just avoiding that...

PR #995, opened 2016-06-06
A few Elem::volume() optimizations

jwpeterson:
Avoid making a few unnecessary `Point` copies and add `TypeVector::cross_norm(a,b)` which is slightly more efficient than doing `a.cross(b).norm()`.


friedmud:
Nice! I can use cross_norm() in some pf the stuff I'm doing as well...

PR #996, opened 2016-06-06
Add optimized version of Tet4::contains_point()

jwpeterson:
Following on the heels of #990, this is a related optimization for Tet4.  I also added some unit tests.  

@dschwen If the volume of the Tet is identically zero, the call to `inverse()` divides by zero and either asserts in dbg mode, or produces NaNs. In the latter case, the function always returns false, which I think is reasonable for a degenerate element (no points are ever contained in such an element).

### DIFF
--- a/include/geom/cell_tet4.h
+++ b/include/geom/cell_tet4.h
@@ -187,6 +187,12 @@ public:
    */
   virtual dof_id_type key () const libmesh_override;
 
+  /**
+   * Uses simple geometric tests to determine if the point p is inside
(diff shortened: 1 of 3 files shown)


jwpeterson:
This cuts the time required to [use a PointLocator to find 1M random points](https://gist.github.com/jwpeterson/3150fe0d7d0aa8be9991d06f7cfdc55e) by a factor of 1/3 (from about 45s to 15s) on my workstation.

PR #999, opened 2016-06-08
Dirichlet BC for shellface boundaries

svallaghe:


dknez:
This looks good to me.

dknez:
I'll merge this soon, unless there are any objections.

roystgnr:
Any chance we could get test coverage on this soon?  I can't actually picture a use case (doesn't this end up constraining away every dof on the element?) and I'd be afraid of screwing it up in future refactoring.

dknez:
Yes, it constrains every dof on a shell element. This is useful if you want to impose constraints on a surface in a structural model. The alternative would be to delete the constrained elements and impose constraints on some 1D sides, but we don't want to have to modify the mesh in order to impose constraints.

Also, if the model has multiple variables then you may want to only constrain a subset of them (e.g. impose a constraint in the vertical direction in an elasticity model). So in that case we do not constrain all the dofs.

Regarding a test case: Yes, that's of course a good idea. I'd like to add a unit test for this, but I don't see an easy way to set up a mesh with SHELL4's in a unit test. Any suggestions on that?

pbauman:
Is it possible to just construct a simple mesh "by hand", like we did in our [mixed-dimensional mesh testing](https://github.com/libMesh/libmesh/blob/master/tests/mesh/mixed_dim_mesh_test.C#L43)?

PR #1000, opened 2016-06-08
Add missing rehash call in Automatic Differentiation

### LINKED ISSUE #998: Potential FPOptimizer issue
This needs further investigation, but I ran into a fairly simple expression (`A*sin(x*pi)*sin(x*pi)`) for which the first derivative was FPOptimized away to `0`. Gulp. Disabling FPOptimizer produces the expected results. I'll double check if this is related to the AD, otherwise I'm in for some FPOptimizer bug hunting fun... :-(

dschwen:
Ok, rehash issue was right on the money. Still took me only six more hours to pin it down and fix it :-(
This is actually quite a critical bug, that is exposed by a recent change that causes AD results to be immediately optimized without an expensive round-trip via the bytecode representation. 

The immediate optimize speeds up the AD process considerably, but leaves my with the responsibility to build all hashes for the code tree of the derived expression. Stale hashes can fool the optimizer into thinking that terms are the same that are not at all the same, leading to possible eliminations of expressions that are not really zero.

The crucial change is on line 252, where the subexpressions in a multiplication were _not_ rehashed, the rest are simplifications.

Closes #998

### DIFF
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -397,7 +397,7 @@ int FunctionParserADBase<Value_t>::AutoDiff(const std::string& var_name)
 
     // should and can we load a cached derivative?
     std::string cache_file;
-    const std::string jitdir = ".jitdir";
+    const std::string jitdir = ".jitcache";
     if (cached)
     {
       // generate a sha1 hash of the Value type size, byte code, and immediate list
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -188,14 +188,14 @@ typename ADImplementation<Value_t>::CodeTreeAD ADImplementation<Value_t>::D(cons
 {
   // derivative of a constant number is 0
   if (func.IsImmed())
-    return CodeTreeAD(CodeTreeImmed(Value_t(0)));
+    return CodeTreeAD(0);
 
   // derivative of a variable is 1 for the variable we are diffing w.r.t. and 0 otherwise
   if (func.IsVar())
   {
     if (func.GetVar() == var) {
       // dx/dx = 1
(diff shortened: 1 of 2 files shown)


dschwen:
🎆 🍰 🍬 💯 
Totally worth the 10 hours I put into debugging this.

dschwen:
I have confirmed that this fixes the error in MOOSE. The whole debugging process was muddied by the fact that I cache derivatives, so my tests still returned the wrong derivatives from the cache after the fix was applied. 

I will add another commit that will invalidate the caches and will consolidate `.jitcache` and `.jitdir` into a single directory.

dschwen:
4c4137e changes the cache directory for the derivative bytecodes. This change will avoid old simulations reusing broken derivatives and it makes it that only one `.jitcache` directory is created (and not another `.jitdir` directory as well).

contrib/fparser/fparser_ad.cc
```diff
@@ -188,14 +188,14 @@ typename ADImplementation<Value_t>::CodeTreeAD ADImplementation<Value_t>::D(cons
...
-    return CodeTreeAD(CodeTreeImmed(Value_t(0)));
+    return CodeTreeAD(0);
 
   // derivative of a variable is 1 for the variable we are diffing w.r.t. and 0 otherwise
   if (func.IsVar())
   {
     if (func.GetVar() == var) {
       // dx/dx = 1
-      return CodeTreeAD(CodeTreeImmed(Value_t(1)));
+      return CodeTreeAD(1);
```
jwpeterson: What's the significance of not calling `CodeTreeImmed(Value_t(x))` on these constants?

dschwen: I had specifically implemented a constructor for `CodeTreeAD` that takes a number to allow for this more compact syntax. I just forgot to use it in a few places.

PR #1002, opened 2016-06-10
Fix new unit tests

roystgnr:
Technically I'm on vacation right now, but... we were finally passing all the crazy configurations I've thrown at BuildBot, and you people are *not* taking that away from me!  ;-)

### DIFF
--- a/tests/mesh/contains_point.C
+++ b/tests/mesh/contains_point.C
@@ -51,11 +51,11 @@ public:
     elem->set_node(2) = &c;
 
     // midpoint
-    Point mid = 1.0/3.0 * (a + b + c);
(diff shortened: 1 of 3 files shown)


tests/mesh/contains_point.C
```diff
@@ -51,11 +51,11 @@ public:
     elem->set_node(2) = &c;
 
     // midpoint
-    Point mid = 1.0/3.0 * (a + b + c);
+    Point mid = (a + b + c) / 3;
     CPPUNIT_ASSERT (elem->contains_point(mid));
```
jwpeterson: This should just be changed to one line of code, `elem->contains_point(elem->centroid())`.  I'll make that small change...

PR #1007, opened 2016-06-22
Fix documentation for TypeVector::operator==

### LINKED ISSUE #1004: TypeVector<T>::operator== Documentation is lying
<pre>
/**
   * @returns \p true iff two vectors occupy approximately the same
   * physical location in space, to within an absolute tolerance of \p TOLERANCE.
   */
  bool operator == (const TypeVector<T> & rhs) const;
</pre>

<pre>
bool TypeVector<T>::operator == (const TypeVector<T> & rhs) const
{
#if LIBMESH_DIM == 1
  return (_coords[0] == rhs._coords[0]);
#endif

#if LIBMESH_DIM == 2
  return (_coords[0] == rhs._coords[0] &&
          _coords[1] == rhs._coords[1]);
#endif

#if LIBMESH_DIM == 3
  return (_coords[0] == rhs._coords[0] &&
          _coords[1] == rhs._coords[1] &&
          _coords[2] == rhs._coords[2]);
#endif
}
</pre>

Should we internally call `absolute_fuzzy_equals`? Or change the documentation?

pbauman:
Closes #1004.

### DIFF
--- a/include/numerics/type_vector.h
+++ b/include/numerics/type_vector.h
@@ -311,8 +311,9 @@ public:
                              Real tol = TOLERANCE) const;
 
   /**
-   * @returns \p true iff two vectors occupy approximately the same
-   * physical location in space, to within an absolute tolerance of \p TOLERANCE.
+   * @returns \p true iff two vectors occupy the same physical location in space,
+   * to within tolerance of the type T. For floating point types, the function
+   * absolute_fuzzy_equals may be a more appropriate choice.
    */
   bool operator == (const TypeVector<T> & rhs) const;
 


include/numerics/type_vector.h
```diff
@@ -311,8 +311,9 @@ class TypeVector
                              Real tol = TOLERANCE) const;
 
   /**
-   * @returns \p true iff two vectors occupy approximately the same
-   * physical location in space, to within an absolute tolerance of \p TOLERANCE.
+   * @returns \p true iff two vectors occupy the same physical location in space,
+   * to within tolerance of the type T. For floating point types, the function
```
jwpeterson: I maybe would have said "calls operator== on every component" of the Point, but this is definitely an improvement over what was there previously.

pbauman:
I should've mentioned: we should probably backport this to 1.0.0.

jwpeterson:
Done.  And I ended up changing it slightly to explicitly say that it calls operator== on every component of the vectors.

PR #1010, opened 2016-06-24
fix detection of vtk 6.0

andreasbuhr:
vtk in version 6.0 was improperly detected: the test
designed for vtk 6.2 was run and failed. This fix changes
the code for detecting vtk 6.1 to also match on vtk 6.0


jwpeterson:
Thanks for testing this out.  I will backport this bugfix to the 1.0.0 branch as well.

PR #1012, opened 2016-06-27
Allow mesh function to return multiple values when point is near face

snschune:
1. Allow `PoinLocator` to return multiple elements containing a point for covering cases where the
point is on (aka close) to a face or node. This is necessary if mesh function is discontinuous.

2. Modify `MeshFunction`  to allow multiple values for a single point.

### DIFF
--- a/include/mesh/mesh_function.h
+++ b/include/mesh/mesh_function.h
@@ -123,6 +123,15 @@ public:
                      const Real time=0.) libmesh_override;
 
   /**
+   * @returns a map of values of variable 0 at point
+   * \p p and for \p time.
+   * map is from element to Number and accounts for double defined
+   * values on faces if discontinuous variables are used
+   */
+  std::map<const Elem *, Number> discontinuous_value (const Point & p,
+                                                      const Real time=0.);
+
+  /**
    * @returns the first derivatives of variable 0 at point
    * \p p and for \p time, which defaults to zero.
    */
@@ -157,6 +166,25 @@ public:
                    const Real time,
                    DenseVector<Number> & output,
                    const std::set<subdomain_id_type> * subdomain_ids);
+
+  /**
+   * Similar to operator() with the same parameter list, but with the difference
+   * that multiple values on faces are explicitly permitted. This is useful for
+   * discontinuous shape functions that are evaluated on faces.
+   */
+  void discontinuous_value (const Point & p,
+                            const Real time,
+                            std::map<const Elem *, DenseVector<Number> > & output);
+
+  /**
+   * Similar to operator() with the same parameter list, but with the difference
+   * that multiple values on faces are explicitly permitted. This is useful for
+   * discontinuous shape functions that are evaluated on faces.
+   */
+  void discontinuous_value (const Point & p,
+                            const Real time,
+                            std::map<const Elem *, DenseVector<Number> > & output,
+                            const std::set<subdomain_id_type> * subdomain_ids);
 
   /**
    * Computes gradients at coordinate \p p and for time \p time, which
@@ -236,6 +264,13 @@ protected:
    */
   const Elem * find_element(const Point & p,
                             const std::set<subdomain_id_type> * subdomain_ids = libmesh_nullptr) const;
+
+  /**
+   * Similar to find_element but returns all elements that are close to a point
+   * to cover cases where p is on the boundary
+   */
+  std::set<const Elem *> find_elements(const Point & p,
+                                       const std::set<subdomain_id_type> * subdomain_ids = libmesh_nullptr) const;
 
   /**
    * The equation systems handler, from which
--- a/include/mesh/mesh_function.h
+++ b/include/mesh/mesh_function.h
@@ -138,6 +138,15 @@ public:
   Gradient gradient (const Point & p,
                      const Real time=0.);
 
+  /**
+   * @returns a map of first derivatives (gradients) of variable 0 at point
+   * \p p and for \p time.
+   * map is from element to Gradient and accounts for double defined
+   * values on faces if the gradient is discontinuous
+   */
+  std::map<const Elem *, Gradient> discontinuous_gradient (const Point & p,
+                                                           const Real time=0.);
+
 #ifdef LIBMESH_ENABLE_SECOND_DERIVATIVES
   /**
    * @returns the second derivatives of variable 0 at point
@@ -196,6 +205,25 @@ public:
                  const Real time,
                  std::vector<Gradient> & output,
                  const std::set<subdomain_id_type> * subdomain_ids = libmesh_nullptr);
+
+  /**
+   * Similar to gradient, but with the difference
+   * that multiple values on faces are explicitly permitted. This is useful for
+   * evaluating gradients on faces where the values to the left and right are different.
+   */
+  void discontinuous_gradient (const Point & p,
+                               const Real time,
+                               std::map<const Elem *, std::vector<Gradient> > & output);
+
+  /**
+   * Similar to gradient, but with the difference
+   * that multiple values on faces are explicitly permitted. This is useful for
+   * evaluating gradients on faces where the values to the left and right are different.
+   */
+  void discontinuous_gradient (const Point & p,
+                               const Real time,
+                               std::map<const Elem *, std::vector<Gradient> > & output,
+                               const std::set<subdomain_id_type> * subdomain_ids);
 
   /**
    * Computes gradients at coordinate \p p and for time \p time, which
--- a/include/utils/point_locator_base.h
+++ b/include/utils/point_locator_base.h
@@ -99,6 +99,14 @@ public:
                                    const std::set<subdomain_id_type> * allowed_subdomains = libmesh_nullptr) const = 0;
 
   /**
+   * Locates a set of elements in proximity to the point with global coordinates
+   * \p p  Pure virtual. Optionally allows the user to restrict the subdomains searched.
+   */
+  virtual void operator() (const Point & p,
+                           std::set<const Elem *> & candidate_elements,
+                           const std::set<subdomain_id_type> * allowed_subdomains = libmesh_nullptr) const = 0;
+
+  /**
    * @returns \p true when this object is properly initialized
    * and ready for use, \p false otherwise.
    */
--- a/include/utils/point_locator_list.h
+++ b/include/utils/point_locator_list.h
@@ -103,6 +103,14 @@ public:
                                    const std::set<subdomain_id_type> * allowed_subdomains = libmesh_nullptr) const libmesh_override;
 
   /**
+   * Locates a set of elements in proximity to the point with global coordinates
+   * \p p  Pure virtual. Optionally allows the user to restrict the subdomains searched.
+   */
+  virtual void operator() (const Point & p,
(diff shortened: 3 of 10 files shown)


include/mesh/mesh_function.h
```diff
@@ -123,6 +123,15 @@ class MeshFunction : public FunctionBase<Number>,
                      const Real time=0.) libmesh_override;
 
   /**
+   * @returns a map of values of variable 0 at point
+   * \p p and for \p time, which defaults to zero.
+   * map is from element on Number and accounts for double defined
+   * values on faces when discontinuous variables are used
+   */
+  std::map<const Elem *, Number> discontinuousValue (const Point & p,
```
jwpeterson: The libmesh function naming convention is to use all lower-case letters and underscores.

include/mesh/mesh_function.h
```diff
@@ -159,6 +168,25 @@ class MeshFunction : public FunctionBase<Number>,
...
+   * dicontinuos shape functions that are evaluated on faces.
+   */
+  void discontinuousValue (const Point & p,
+                           const Real time,
+                           std::map<const Elem *, DenseVector<Number> > & output);
+
+  /**
+   * Similar to operator() with the same parameter list, but with the difference
+   * that multiple values on faces are explicitly permitted. This is useful for
+   * dicontinuos shape functions that are evaluated on faces.
```
jwpeterson: typo: "dicontinuos" seems to have been copy/pasted from somewhere else a couple of times.

include/mesh/mesh_function.h
```diff
@@ -238,6 +266,13 @@ class MeshFunction : public FunctionBase<Number>,
                             const std::set<subdomain_id_type> * subdomain_ids = libmesh_nullptr) const;
 
   /**
+   * Similar to find_element but returns all elements that are close to a point
+   * to cover cases where p is on the boundary
+   */
+  std::set<const Elem *> find_multi_element(const Point & p,
```
jwpeterson: I'd prefer the name "find_elements" (with an s) for the new interface.

include/utils/point_locator_base.h
```diff
@@ -99,6 +99,14 @@ class PointLocatorBase : public ReferenceCountedObject<PointLocatorBase>
                                    const std::set<subdomain_id_type> * allowed_subdomains = libmesh_nullptr) const = 0;
 
   /**
+   * Locates a set of element in proximity to the point with global coordinates
```
jwpeterson: typo: element -> elements.

include/utils/point_locator_list.h
```diff
@@ -103,6 +103,14 @@ class PointLocatorList : public PointLocatorBase
                                    const std::set<subdomain_id_type> * allowed_subdomains = libmesh_nullptr) const libmesh_override;
 
   /**
+  * Locates a set of element in proximity to the point with global coordinates
```
jwpeterson: element -> elements

include/utils/point_locator_tree.h
```diff
@@ -108,6 +108,14 @@ class PointLocatorTree : public PointLocatorBase
                                    const std::set<subdomain_id_type> * allowed_subdomains = libmesh_nullptr) const libmesh_override;
 
   /**
+  * Locates a set of element in proximity to the point with global coordinates
```
jwpeterson: element -> elements

src/mesh/mesh_function.C
```diff
@@ -187,7 +187,20 @@ Number MeshFunction::operator() (const Point & p,
...
+std::map<const Elem *, Number> MeshFunction::discontinuousValue (const Point & p,
+                                                                 const Real time)
+{
+  libmesh_assert (this->initialized());
 
+  std::map<const Elem *, DenseVector<Number> > buffer;
+  this->discontinuousValue (p, time, buffer);
+  std::map<const Elem *, Number> return_value;
+  for (std::map<const Elem *, DenseVector<Number> >::const_iterator it = buffer.begin(); it != buffer.end(); ++it)
+    return_value[(*it).first] = (*it).second(0);
```
jwpeterson: `(*it).first` should just be `it->first`, similarly for `second`.

src/mesh/mesh_function.C
```diff
@@ -300,6 +313,96 @@ void MeshFunction::operator() (const Point & p,
...
+
+          // next variable
+        }
+
+      // Insert temp_output into output
+      output[element] = temp_output;
+    }
+
+  // all done
+  return;
```
jwpeterson: Delete redundant return

include/utils/point_locator_list.h
jwpeterson: Also I think the indentation of the comment block is off here.

include/mesh/mesh_function.h
jwpeterson: Also, the comment no longer makes sense I think, as the default value is not zero but rather an empty map in this case?

src/utils/point_locator_list.C
```diff
@@ -192,6 +192,35 @@ const Elem * PointLocatorList::operator() (const Point & p,
 }
 
 
+void PointLocatorList::operator() (const Point & p,
```
jwpeterson: PointLocatorList is officially deprecated, so I wouldn't spend too much time on this... or simply leave it as `libmesh_unimplemented()`.

include/mesh/mesh_function.h
snschune: I think the `which defaults to zero.` refers to `time=0`. The return value default (`out_of_mesh_value`) can be set.

snschune:
@jwpeterson how should the new methods be tested?

jwpeterson:
> how should the new methods be tested?

This is a good candidate for unit testing.  You can copy/paste one of the existing examples in tests/mesh, or create your own .C file which is just for PointLocator stuff.  If you do add a new file, you have to manually add it to the list of files in tests/Makefile.am and run bootstrap for it to be picked up by the build system.

snschune:
Ok, thanks @jwpeterson, I'll do.

include/mesh/mesh_function.h
```diff
@@ -159,6 +168,25 @@ class MeshFunction : public FunctionBase<Number>,
                    const std::set<subdomain_id_type> * subdomain_ids);
 
   /**
+   * Similar to operator() with the same parameter list, but with the difference
+   * that multiple values on faces are explicitly permitted. This is useful for
+   * discontinuous shape functions that are evaluated on faces.
+   */
+  void discontinuous_value (const Point & p,
+                            const Real time,
+                            std::map<const Elem *, DenseVector<Number> > & output);
```
friedmud: What's the reasoning behind using `DenseVector` here?  I would prefer just using `std::vector`.

roystgnr: DenseVector is what we've always done for FunctionBase etc. which return gradients.  You can use DenseVector::get_values() if you need to pass the output as a std::vector but don't want to copy it.

friedmud: Ok.  Just checking :-)

tests/mesh/mesh_function_dfem.C
```diff
@@ -0,0 +1,219 @@
...
+      elem_bottom->set_node(3) = _mesh->node_ptr(0);
+    }
+
+    // libMesh will renumber, but we numbered according to its scheme
+    // anyway. We do this because when we call uniformly_refine subsequenly,
+    // it's going use skip_renumber=false.
+    _mesh->prepare_for_use(false /*skip_renumber*/);
+
+    // get a point locator
+    _point_locator = _mesh->sub_point_locator().release();
```
jwpeterson: `release()` doesn't seem right, since I think it "steals" the `_point_locator` held internally by the Mesh.

Why not just let the Mesh manage the PointLocator lifetime, and get a `UniquePtr` to it when you use it in `test_point_locator_dfem()`?

tests/mesh/mesh_function_dfem.C
```diff
@@ -0,0 +1,218 @@
...
+      CPPUNIT_ASSERT ((*it)->id() == 1);
+
+    // test interior point
+    std::set<const Elem *> face_cand;
+    (*_point_locator)(face, face_cand, libmesh_nullptr);
+    CPPUNIT_ASSERT (face_cand.size() == 2);
+    unsigned int count = 0;
+    for (std::set<const Elem *>::iterator it = face_cand.begin(); it != face_cand.end(); ++it)
+    {
+      // FIXME: this tests occasionally fails. Why?
```
jwpeterson: Since the std::set is sorted on pointers, you have no guarantee that Element 0's address is going to come before Element 1's address in memory.  So I'd change the check here to something else... maybe just
```
int array[2] = {0,0};
for (each face_cand)
  array[(*it)->id()] = 1;
// Check that the array is all 1's
```

snschune: `for (each face_cand)`
is that c++11?

jwpeterson: Haha, no, just pseudo code.

snschune: `for(auto j : face_cand)` ... not so far off.

friedmud: @snschune we usually recommend `for (auto & j : face_cand)` 😉

snschune: thanks, I gotta get on that c++11 train.

snschune:
I will implement a discontinuousGradient as well before merging as it will most likely be more useful for CFEM.

jwpeterson:
The PR seems to be in good shape, I am just worried that you will not be happy with the resulting performance of your new version of `PointLocatorTree::operator()` which does a linear search through all the elements... 

Have you already started using this in your app, and if so, did you try it on a fairly large mesh and get some idea of the timings?

snschune:
I was concerned about that as well. I have not looked at timing at all. Do you think that your suggestion to search a patch around the first hit would be a viable alternative? Or should I just check performance first before looking into optimization?

jwpeterson:
> Do you think that your suggestion to search a patch around the first hit would be a viable alternative?

In cases where there are no lower-dimensional manifolds present, I think it would probably work OK, provided that you could guarantee that elements only sharing a single node (i.e. node neighbors) were always included in the patch you built.

Aside from that, I don't have any other ideas really.  I need to look at the details of how we build the binary/quad/oct-trees, but it seems to me that if they are the "region" type of trees (each "node" in the tree represents a region of 1D/2D/3D space) then linearly searching all elements in the parent node's region (or maybe the grandparent's node's region) would be a possible optimization on linearly-searching the entire Mesh.

> Or should I just check performance first before looking into optimization?

I would definitely see how bad it is first, yes.

snschune:
I am working on the Moose PR right now. The application is in Rattlesnake where we want to use 
an existing solution in boundary conditions for a smaller model. I'd say I'll see how bad it is for our application and come back if it's significant. 

My additions do not affect existing code since I implemented separate methods. I'd like to have this
in the next libmesh update if you agree.

jwpeterson:
> I'd like to have this in the next libmesh update if you agree.

Seems fine to me.  If we come up with something more efficient later, we probably won't have to change the API added in this PR.

PR #1013, opened 2016-06-28
Remove deprecated classes

jwpeterson:
This PR removes the deprecated I/O classes:

* XdrMGF, XdrMESH, XdrSOLN
* XdrHEAD, XdrMHEAD, XdrSHEAD
* LegacyXdrIO

These have only been "officially" deprecated since 7a279f4e (Apr. 2016) but I don't think they have actually been used by anyone for many years.


roystgnr:
Nobody's been using them on purpose for many years, but how sure are we that nobody's been using them by accident?  IIRC when we read a legacy format file we don't scream about it we just silently succeed.  And if that's true then we should probably have a "scream about it but succeed" phase before we progress to "scream about it and die".

jwpeterson:
> Nobody's been using them on purpose for many years, but how sure are we that nobody's been using them by accident?

You wouldn't be using LegacyXdrIO by accident to *write* files, but yes, you may have been using it to unknowingly read them. But that would mean you are still reading an XDA/R file that was written before 536f495 (Feb 20, 2008).  I'd rather volunteer to help anyone who's in this situation port their files over to the new format (using say 1.0.0) than keep `LegacyXdrIO` around for some indeterminate length of time while people continue to ignore any screaming we might do...

> we should probably have a "scream about it but succeed" phase

Scream more than the current `libmesh_deprecated()` message?  What do you suggest?

Also, what about the non-`LegacyXdrIO` classes?  OK with removing them?

roystgnr:
Ah!  I didn't notice the libmesh_deprecated() in the constructors; I was just looking at the silence higher on the stack.

That's only been there since April... but it made it into 1.0.0, which is good enough for me.  :+1: to merge this as-is.

jwpeterson:
Thanks!  I was not looking forward to reverting selected chunks of the existing commits ;-)

pbauman:
https://imgflip.com/i/16n199

roystgnr:
Ha!!  Regression in GRINS, Spalart-Allmaras turbulent channel test.  @vikramvgarg must have written a script to convert some data into that grid, and used our old format?  At least it's xda so it'll be easy to upgrade.

PR #1016, opened 2016-06-30
Better default-values for system-parameters and error-handling in misc. 12

BalticPinguin:
Hi,
I have a small correction for the example miscellaneous-12: Here, second derivatives are used in the assemble-function but there is no exception in case of compile-option --disable-second.

The second issue concerns the compute_data() function for infinite elements: If the parameters speed and frequency are not set, they default to 0 and hence the wavelength=speed/frequency becomes 0 or NaN respectively. I would therefore suggest to default both values to 1 instead and tell the user to set them according to his needs.

(btw: concerning the InfFE::compute_data() function I think there is an error at the moment, I will write a report soon but need to study it a bit more).

Sorry for mixing two different topics into one request, but since they are only few lines...

### DIFF
--- a/examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
+++ b/examples/miscellaneous/miscellaneous_ex12/miscellaneous_ex12.C
@@ -84,6 +84,10 @@ int main (int argc, char ** argv)
   libmesh_example_requires (false, "ExodusII support");
 #endif
 
+#ifndef LIBMESH_ENABLE_SECOND_DERIVATIVES
+  libmesh_example_requires (false, "second derivatives enabled");
+#endif
+
   // This example does a bunch of linear algebra during assembly, and
   // therefore requires Eigen.
 #ifndef LIBMESH_HAVE_EIGEN
--- a/src/fe/fe_compute_data.C
+++ b/src/fe/fe_compute_data.C
@@ -28,12 +28,13 @@ void FEComputeData::clear ()
   this->shape.clear();
 #if defined(LIBMESH_ENABLE_INFINITE_ELEMENTS) && !defined(LIBMESH_USE_COMPLEX_NUMBERS)
   this->phase = 0.;
-  this->speed = 0.;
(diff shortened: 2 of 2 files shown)


jwpeterson:
Here's the error message:
```
../src/fe/fe_compute_data.C: In member function ‘void libMesh::FEComputeData::init()’:
../src/fe/fe_compute_data.C:67:21: error: ‘class libMesh::FEComputeData’ has no member named ‘frequency’
    libmesh_assert_not_equal_to(this->frequency,0);
                     ^
../src/fe/fe_compute_data.C:67:147: error: ‘class libMesh::FEComputeData’ has no member named ‘frequency’
../src/fe/fe_compute_data.C:68:21: error: ‘class libMesh::FEComputeData’ has no member named ‘speed’
    libmesh_assert_not_equal_to(this->speed,0 );
                     ^
../src/fe/fe_compute_data.C:68:135: error: ‘class libMesh::FEComputeData’ has no member named ‘speed’
make[2]: *** [src/fe/libmesh_dbg_la-fe_compute_data.lo] Error 1
```

Looks like you might need to wrap some code in `#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS`?

BalticPinguin:
Thanks. I didn't see the reason for this error.

jwpeterson:
> I didn't see the reason for this error.

If you click the "Details" link to the side of the continuous integration testing links, does it take you to a moosebuild.org website where you can see the status of the tests?  If not, we can try and figure out why...

jwpeterson:
In looking at fe_compute_data.h, it seems that `frequency` is only available when both 
```
defined LIBMESH_ENABLE_INFINITE_ELEMENTS && defined(LIBMESH_USE_COMPLEX_NUMBERS)
```
while `speed` only requires infinite elements to be defined.  I can fix up this last detail in the merge, though, no need to worry about it.

BalticPinguin:
hmm... you are right. Sorry, I didn't check for which variables are available on condition only.

Issue #1017, opened 2016-06-30
Possibly rethink System::point_value() API


jwpeterson:
We recently had an email support issue where the user had done the following:
```
Point p(1.5, 0.2, 0);
const Elem *e = (*locator)(p);

float A[3];
A[0] = asys->point_value(Ax_var, p, e);
A[1] = asys->point_value(Ay_var, p, e);
A[2] = asys->point_value(Az_var, p, e);
```

This incorrectly calls the following version of the `point_value()` function:
```
  Number point_value(unsigned int var, const Point & p,
                     const bool insist_on_success = true) const;
```
when the intent was to call this version:
```
  Number point_value(unsigned int var, const Point & p, const Elem & e) const;
```

The compiler does not warn about the automatic conversion from non-`nullptr` pointer to `true` (I wouldn't expect it to), and the less efficient version of the function is called.

One idea might be to deprecate the bool interface and add one that takes an enum instead?  I don't think this would have the same automatic conversion issue, but I haven't tested it.  @roystgnr

roystgnr:
Hmm... could we add an overload of the API that takes an Elem pointer?  Then arguments of type Elem* or its subclasses will resolve to that version, but we don't have to break any code which relies on the bool version.

jwpeterson:
That seems like a good solution, I'll try it with my test code.

roystgnr:
This should be resolved by #1018, yes?

PR #1018, opened 2016-06-30
Add versions of System::point_{value,gradient,hessian}() taking an Elem*

jwpeterson:
Refs #1017.


roystgnr:
Thanks!

Shame I can't figure out how to get a unit test to ensure the correct binding doesn't get regressed...

jwpeterson:
The only failure was a timeout on a debug mode test that we've seen happen before, so I'm going ahead with the merge.

PR #1019, opened 2016-07-01
Reduced Basis change: Removed "extra" interpolation data from EIM.

dknez:
This extra interpolation data is not necessary and it complicates the code.
Also, it can cause errors during the Greedy algorithm if the "extra" sample
point is linearly dependent with the other points. Therefore it's better
to get rid of it.


dknez:
The test is showing an error for reduced_basis_ex4, but that works fine for me... not sure what's going on here.

dknez:
Oh, it's probably because I'm using capnproto. I'll have to check the non-capnproto case.

PR #1021, opened 2016-07-08
Got DenseMatrix SVD working with --enable-complex and updated unit test.

dknez:


jwpeterson:
I think `std::real` is a C++11 only thing.

dknez:
I switch it to libmesh_real. Hopefully that fixes it.

PR #1030, opened 2016-07-19
TypeTensor::solve,inverse() should throw an exception on zero determinant

jwpeterson:
I don't expect this to be a very common situation, but the way we were handling it before wasn't great: in debug mode we would assert, while in opt mode we would continue on and divide by zero.  This branch throws an exception (regardless of METHOD) in case of a zero determinant, and handles the exception at the call sites in an appropriate way.  I also came across some new warnings while compiling with GCC (usually I use clang) and a few deprecated function calls, so those are fixed as well.


roystgnr:
:+1:, but now you don't get to talk smack at me if *I* ever stuff three unrelated changesets into the same PR.

PR #1031, opened 2016-07-20
Stop GCC 6.1 screaming like a banshee

roystgnr:
Apparently "-pedantic" isn't a warning so it can't be ignored;
"-Wpedantic" is the best we can do.


jwpeterson:
Sounds like this compiler is being pretty... 😎 ... pedantic.

PR #1032, opened 2016-07-20
Document BoundaryInfo behavior on DistributedMesh

roystgnr:
Technically @pbauman broke GRINS on DistributedMesh, but after looking at the documentation for the functions he was using I have to say the mistakes were clearly my fault.

This should be more understandable for future users.  I'm going to put in the DistributedMesh-compatible behavior at the GRINS level rather than in here, though, because it should be roughly a one-line fix per use and "get_local_stuff(); merge_it();" is better than adding a "get_global_stuff();" API for every type of stuff.

### DIFF
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -526,43 +526,63 @@ public:
 
   /**
    * Builds the list of unique node boundary ids.
+   *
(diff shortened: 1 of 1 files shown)


pbauman:
Pretty clear to me, thanks! 👍

PR #1033, opened 2016-07-20
Remove previously-deprecated PointLocatorList

jwpeterson:
This has been deprecated since well before 1.0.0, plus it has been deprecated in @roystgnr's mind since 025cc9671c7852e347476d5bde19dc2f3069919a (2007).

```
// This code will only work if your mesh is the Voronoi mesh of its
// own elements' centroids.  If your mesh is that regular you might
// as well hand-code an O(1) algorithm for locating points within
// it. - RHS
```


roystgnr:
:+1:

I do wish we did have a decent slow-but-ultra-reliable PointLocator option, though.  If we ever futz with anisotropic or non-rectilinear tree options it would be nice to have a fallback.

jwpeterson:
If `_out_of_mesh_mode == false`, `PointLocatorTree` falls back on a linear search of the elements... is that good enough?

roystgnr:
It's certainly good enough for now.  And I guess we can always fork the class if we want to create a more complicated version later.

PR #1036, opened 2016-07-21
Fix Hypre check in petsc.m4.

jwpeterson:
I never noticed this because we don't actually use the #define
for anything, but we haven't properly been setting LIBMESH_HAVE_PETSC_HYPRE
since the location of the petscvariables file moved in 3.6.x.


roystgnr:
:+1:

PR #1040, opened 2016-07-26
Two patches from Junchao Zhang

jwpeterson:


roystgnr:
Thanks for this, both of you!  Hold off on merging until I can check it with the GRINS test suite too, though?  I can easily imagine that second fix screwing with any regression tests against gold results which were generated with the buggy behavior.

pbauman:
> Thanks for this, both of you! Hold off on merging until I can check it with the GRINS test suite too, though?

Thanks for pointing that out @roystgnr. I'll take care of running against this and will report in here.

jwpeterson:
@roystgnr I'll let you merge whenever you're ready, I'm in no hurry.

pbauman:
OK, I am getting a failure in one of the GRINS tests with this PR. Going to double check against current libMesh master to make sure nothing else crept in.

pbauman:
GRINS passes tests against current libMesh master so it's definitely this PR.

pbauman:
Confirmed that I can generate a new "gold" for the test that works with this PR. Also confirmed that the new gold fails with current libMesh master.

jwpeterson:
Is the regolded file one you want to keep?

Doesn't look like you guys have a libmesh submodule in GRINS... do you want this PR to be merged before you update the gold file in GRINS?  After?

roystgnr:
Go ahead and merge now.  I hate when this makes it harder for us to bisect (perhaps we ought to do a git commit date check and XFAIL if we know a test is not synched with a known bug?) but we've coped with it before.

PR #1042, opened 2016-07-29
Remove DenseMatrix::svd()

dknez:
I added svd to DenseMatrix a few years ago. On reflection, I prefer to remove it
since I've found that the current implementation fails in some cases. The LAPACK
interface that is used under the hood is complex and has many options, and rather
than try to handle all those options correctly in all cases, I prefer to just
remove this method. If users want an SVD of a DenseMatrix, it's better if they
call LAPACK themselves (which they can do via the PETSc interface to LAPACK).


jwpeterson:
I'd rather not remove it.  I spent time fixing it for the non-square case.  also its faster than Eigen IIRC.

roystgnr:
Agreed with @jwpeterson.  Documentation of the failure cases would be nice, though.

dknez:
OK... I'll close this PR then.

It's failing for me in some complex-valued cases. It's not obvious to me what's going wrong, so my fix was to copy the implementation in petsc mat example 116, which works better for me. That has a bunch of PETSc boilerplate in it though, so I didn't want to put that into DenseMatrix.

PR #1045, opened 2016-08-01
Fix warnings

jwpeterson:
I inadvertently broke SLEPc-enabled builds while trying to fix warnings.  That should be fixed by this branch as well as ignoring other warnings from the SLEPc and Trilinos headers.


roystgnr:
I was just about to fix e94cb02 myself; thanks.

PR #1049, opened 2016-08-03
Add some tests for manual indexing of dof indices

friedmud:
Just a few tests related to manual indexing of dof_indices.  I'm starting to write code that depends on this... so @roystgnr thought we should have some tests for it.


roystgnr:
Looks good; thanks!
:+1:

PR #1051, opened 2016-08-03
Updates to systems_of_equations_ex8.

dknez:
We now add EDGE2 elements into the mesh to connect contact nodes. This provides a good test for multi-dimensional meshes. Also, it is a better approach in general for ensuring consistent data transfer in parallel for the contact dofs.


roystgnr:
Fine by me.  Leave misc_ex9 alone, though, so we can use it to test the upcoming ghosting APIs?

dknez:
Yep. I won't change misc_ex9.

The reason for the change is that I use this approach of adding in contact elements in app codes, so I wanted to get some test coverage for it.

Issue #1028, opened 2016-07-18
Proper Ghosted Elements


friedmud:
This is a follow-on from the discussion in #977 .  I'm going to repost most of my comment from there:

We need the ability to just flag elements (any element) as needing to be "ghosted" locally... and libMesh should just treat that element as if it's semilocal.  It should never get removed by `ParallelMesh` and its dofs should always be ghosted in all relevant ghosted vectors.

Basically: I want the ability to arbitrarily mark elements as needing to be treated just like the "one layer" of halo elements we always have.

This issue is not just relevant to periodic boundaries... it happens anytime you have a "gap" in the mesh and you want to keep elements on both sides of the gap for whatever reason.

For instance, in MOOSE this happens because of contact.  Because of that we have quite a bit of code to try to force libMesh into not deleting elements and ghosting those extra degrees of freedom.  To deal with this we have a few things:
1.  `ghosted_boundaries`.  This is an input file parameter that a user can supply that will cause any elements connected to those sidesets to be ghosted to _all_ processors (definitely non-optimal and heavy handed but it works for periodic and for contact boundaries).  You can see the code that does the ghosting here:
   
   https://github.com/idaholab/moose/blob/devel/framework/src/mesh/MooseMesh.C#L2116
2.  A couple of years ago I added `libMesh::DistributedMesh::add_extra_ghost_elem()` that at least tries to mark elements so that they won't be removed.  That function is here:
   https://github.com/libMesh/libmesh/blob/master/src/mesh/parallel_mesh.C#L1415
   
   And you can see the extra code _not_ remove the elems flagged this way here:
   
   https://github.com/libMesh/libmesh/blob/master/src/mesh/mesh_communication.C#L1349
3.  We do a LOT of work in MOOSE to try to make sure the dofs on these "extra ghost elems" end up in ghosted vectors.  It starts by keeping track of them here:
   https://github.com/idaholab/moose/blob/devel/framework/src/base/FEProblem.C#L913
   
   And then adding any dofs connected to those elements to the `send_list` here:
   
   https://github.com/idaholab/moose/blob/devel/framework/src/base/SystemBase.C#L377
4.  We have to do several `EquationSystem::reinit()` calls because of ghosted elements... we have to do one for adaptivity... and then another one after we re-adjust the ghosted elements (to get the vector ghosting right).  You can see that here:
   https://github.com/idaholab/moose/blob/devel/framework/src/base/FEProblem.C#L3761
[...]

PR #1052, opened 2016-08-03
Reduce overghosting

roystgnr:
I'll be adding the new GhostingFunctor interface as discussed in #1028 eventually, but while getting that ready I noticed that we were grossly over-ghosting elements on AMR meshes, by getting point neighbors of not just our active local elements but also of their ancestors.

This PR eliminates some overzealous assertions which were being enabled by overghosting, eliminates the overghosting, and does a little refactoring because redistribute() and delete_remote_elements() should really be using the same ghosting criteria.

### DIFF
--- a/src/mesh/mesh_tools.C
+++ b/src/mesh/mesh_tools.C
@@ -1128,8 +1128,13 @@ void MeshTools::libmesh_assert_valid_remote_elems(const MeshBase & mesh)
     {
       const Elem * elem = *el;
       libmesh_assert (elem);
-      for (unsigned int n=0; n != elem->n_neighbors(); ++n)
-        libmesh_assert_not_equal_to (elem->neighbor_ptr(n), remote_elem);
+
+      // We currently don't allow active_local_elements to have
+      // remote_elem neighbors
+      if (elem->active())
+        for (unsigned int n=0; n != elem->n_neighbors(); ++n)
+          libmesh_assert_not_equal_to (elem->neighbor_ptr(n), remote_elem);
+
 #ifdef LIBMESH_ENABLE_AMR
       const Elem * parent = elem->parent();
       if (parent)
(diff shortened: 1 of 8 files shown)


src/mesh/mesh_tools.C
```diff
@@ -1128,16 +1128,23 @@ void MeshTools::libmesh_assert_valid_remote_elems(const MeshBase & mesh)
...
       const Elem * elem = *el;
       libmesh_assert (elem);
-      for (unsigned int n=0; n != elem->n_neighbors(); ++n)
-        libmesh_assert_not_equal_to (elem->neighbor_ptr(n), remote_elem);
+
+      // We currently don't allow active_local_elements to have
+      // remote_elem neighbors
+      if (elem->active())
+        for (unsigned int n=0; n != elem->n_neighbors(); ++n)
+          libmesh_assert_not_equal_to (elem->neighbor_ptr(n), remote_elem);
```
pbauman: I should know the answer to this, but the `for` loop will get optimized away in opt mode, yes? Otherwise, we need to `#ifdef` around these.

jwpeterson: Yeah, I think any compiler worth its salt should figure out the loop body is empty and optimize it out.

roystgnr: Shit, no.  n_neighbors() is virtual.  I'll wrap in ifndef NDEBUG.

roystgnr: Wait.  That whole function is in our "crazy expensive verification tests" MeshTools suite, which isn't even defined unless ifdef DEBUG.

Still, good lesson for future optimization.  For loops aren't always as cheap as they look or as they should be.

roystgnr:
This is passing for me with GRINS too.

PR #1054, opened 2016-08-04
Fixup dof indices test

friedmud:


jwpeterson:
This fixed the test for me and we know it won't be tested by Civet (also I'm buying beer via idaholab/moose#410) so I went ahead and merged.

PR #1056, opened 2016-08-04
Add get_array() capability to PetscVector 

### LINKED ISSUE #1050: API additions for PetscVector
Somewhat related to #854 ...

I'm planning on adding a few new members to `PetscVector`:

`get_array()`: Get a read/write raw PETSc data array for the vector
`get_array_read()`: Get a read only raw PETSc data array for the vector

And the corresponding `restore_array()` (Note a `restore_array_read()` is not needed because we can just do the right thing internally depending on which thing was called.

Internally this is going to call `_get_array()` and set the internal `_values` member and return it.  `_array_is_present` will get set to `true` and a flag indicating which kind of array (read/write or just read) will be set.

From this point on - all other functions on `PetscVector` will just use the existing `_values` instead of doing their own `VecGetArray()` or `VecRestoreArray()`.

friedmud:
Ads a new interface `PetscVector` to allow getting the raw array.

There are two ways to retrieve the array: `get_array()` and `get_array_read()` associated with whether or not you want read/write or just read only.  There is also `restore_array()` for when you're finished.

One important thing is that the state of that array is now stored... and error checking is done to make sure the new API interoperates with the existing API.  In particular, if you use `get_array()` and then attempt to use one of the other API functions that needs to restore the array... it will produce an error (i.e. a manual "getting" of the array requires a manual "restore").

closes #1050

### DIFF
--- a/include/numerics/petsc_vector.h
+++ b/include/numerics/petsc_vector.h
@@ -294,6 +294,30 @@ public:
                    T * values) const libmesh_override;
 
   /**
+   * Get read/write access to the raw PETSc Vector data array
+   *
+   * Note: This is an extremely advanced interface.
+   * In general you should avoid using it!
+   */
+  PetscScalar * get_array();
+
+  /**
+   * Get read only access to the raw PETSc Vector data array
+   *
+   * Note: This is an extremely advanced interface.
+   * In general you should avoid using it!
+   */
+  const PetscScalar * get_array_read() const;
+
+  /**
+   * Restore the data array.
+   *
+   * MUST be called after get_array() or get_array_read() and before using
+   * any other interface functions on PetscVector.
+   */
+  void restore_array();
+
+  /**
    * Addition operator.
    * Fast equivalent to \p U.add(1, V).
    */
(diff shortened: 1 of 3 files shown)


roystgnr:
:+1:

I still wish we had a better way to benchmark performance changes with edits to critical code.  But "time make check" is worthless unless you have a dedicated benchmarking server.  "LIBMESH_RUN='valgrind --foo'" is immune to server load but is ridiculously slow.  Even perf/oprofile/PAPI probably aren't trustworthy enough unless you're running on a mostly-quiescent computer; we could get reliable numbers for operation counts but not for cache misses.

permcody:
We do have a dedicated benchmarking server. I suppose we could start with
the libMesh suite before developing the MOOSE testing suite. Our regression
test suite has too many tests and runs too fast for good timing
information. I'll see about getting something up and running soon. We could
go back a few commits to pick some of these changes up for our first
baseline.

jwpeterson:
@friedmud are you OK with this being merged?  @snschune wants to do a libmesh update soonish, and I assume you'd like this in upstream before that happens?

friedmud:
Yep - go ahead. I'm already using it pretty successfully. I'm sure there
will be some tweaks to it going forward though...

PR #1057, opened 2016-08-04
Ghosting functors

roystgnr:
This abstracts out the concept of geometric ghosting (the "G" discussed in #1028), moves our default ghosting behavior into a functor, and allows user code to add its own ghosting functors to extend that behavior.

This only covers geometric ghosting, not yet algebraic ghosting or coupling, so two thirds of the work in #1028 is still to come.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -208,11 +208,12 @@ am__DEPENDENCIES_2 = $(am__DEPENDENCIES_1)
 @LIBMESH_DBG_MODE_TRUE@	$(am__DEPENDENCIES_2)
 am__libmesh_dbg_la_SOURCES_DIST = src/base/dof_map.C \
 	src/base/dof_map_constraints.C src/base/dof_object.C \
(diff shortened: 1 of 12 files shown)


roystgnr:
I'm still running tests on this myself; I don't intend to merge until those are done.

I'd like to merge this despite being at the "one-third done" state, because this is a complete (albeit not yet widely useful) feature as-is, and because my newly added files are affecting the build system and so are likely to lead to annoying merge conflicts if I leave things branched too long.

roystgnr:
This passes GRINS (with --enable-parmesh of course) too.

roystgnr:
Oh, damn, I forgot some callbacks in delete_remote_elements.  Those are a no-op right now, but they belong in this PR.

roystgnr:
I wish the CI wouldn't put an X beside commits that didn't pass tests solely because the commit was superceded before Civet was done testing and so the test was canceled.  I worked hard on that, and it was correct, even if you didn't have time to prove it!

PR #1060, opened 2016-08-09
Set _initialized flag to false on clear()

dschwen:
PointLocatorTree sets `_initialized` to true in `init()` but it never ever sets it to false again, which I think it should do in `clear()`  where the tree is deleted (and that's a public method).


jwpeterson:
I attached the MOOSE App tests to this PR via Civet as well.

Issue #1059, opened 2016-08-09
Make interface of DenseMatrix::evd similar as svd


YaqiWang:
We'd like to get eigenvalues of a small dense matrix instead of singular values, which are non-negative. Current interface does not return the eigenvectors which we also want. It will be great we can have similar interface like `svd` for `evd`. Thanks!

jwpeterson:
This can probably be done, but note that `DenseMatrix::evd()` is implemented in terms of `DGEEVX`, which we assume is available through SLEPc (to avoid worrying about different ways of wrapping the Fortran function call) so you will need to have SLEPc available in order to use it...

dknez:
A patch for this change would be welcome! It should be pretty straightforward to provide a second implementation of evd that returns the eigenvectors.

jwpeterson:
After looking at it in a bit more detail, it looks like we can get what we need from [dgeev](http://www.netlib.org/lapack/explore-html/d9/d8e/group__double_g_eeigen_ga8ec1625302675b981eb34ed024b27a47.html#ga8ec1625302675b981eb34ed024b27a47), which is available via PETSc, rather than relying on the more complicated/sophisticated [dgeevx](http://www.netlib.org/lapack/explore-html/d9/d8e/group__double_g_eeigen_gaf816b85f7a6fcc4ae22ab0fc17768ce6.html#gaf816b85f7a6fcc4ae22ab0fc17768ce6) as is done now... so the first part of the fix for this Issue should be using the simpler interface, and the second part would be making the eigenvectors available.  @YaqiWang do you need left eigenvalues or right eigenvalues or both?

YaqiWang:
In our case the matrix is symmetric, so left and right are the same. But in general we need both. Another thing: do we always have such a decomposition like svd?

jwpeterson:
> Another thing: do we always have such a decomposition like svd?

The eigen-decomposition A = VDV^{-1} exists if the eigen-vectors are linearly independent (hence V^{-1} makes sense), but this is not guaranteed for general matrices...  It should be guaranteed for symmetric positive semi-definite matrices though, which is why SVD is guaranteed...

jwpeterson:
Implemented in #1063.

PR #1061, opened 2016-08-10
Re-implement DenseMatrix::evd() in terms of dgeev()

jwpeterson:
This is an important first step towards implementing the eigenvector computation requested in #1059.  Since `DenseMatrix::evd()` is now available to anyone using PETSc (whereas before it required SLEPc) also added a couple unit tests to make sure the eigenvalue calculation is working.


dknez:
:thumbsup:

Issue #1058, opened 2016-08-08
MPI_Scatter in parallel.h?


permcody:
In doing some preliminary runs of the optimized grain tracker I made crude observations of the memory usage and am simply blown away by how much my data structures appear to occupy. Running 6000 grains on 180^3 mesh seemed to increase the memory usage right after the `allgather` operation (cause of the non-issue in #1039) by over a GB per processor! This is the optimized version mind you!  The back of the envelope calculations don't seem to add up to this amount of memory but I do have a lot of data in sets which takes often more than 4x the amount of memory for POD types AND that data structure exists on every processor, so maybe...

I'm thinking about really digging into this algorithm and changing it so that only the root process maintains the big data structure and the remaining ranks maintain only local data and local to global mappings. I don't think this will be too difficult but I could really use `MPI_Scatter` to get this done. Can we add this to parallel.h?

TL;DR;
I'd love to have MPI_Scatter in parallel.h. Any objections?

roystgnr:
Sounds good to me!

permcody:
I'm starting to flesh out the `scatter` operations. Before I finish this up I want to make sure that everyone (@roystgnr, @jwpeterson, @friedmud, etc) like the API I'm setting up. Here are a few options:

```
/**
 * Takes a vector equal to the communicator size and sends the ith item to the ith processor
 * stores it in recv.
 */
template <typename T>
inline void Communicator::scatter (std::vector<T> & data, T & recv, const unsigned int root_id=0) const;

/**
  * Take a "stacked" vector and distributes equal sized chunks to each processor and 
  * stores in recv. The recv vectors do not have to be sized for this operation.
  */
template <typename T>
inline void Communicator::scatter (std::vector<T> & data, std::vector<T> & recv, const unsigned int root_id=0) const;

/**
  * Takes a "stacked" data vector and a vector of sizes for distributing non-uniform 
  * sized chunks to all processors.
  * The recv vectors do not have to be sized for this operation.
  */
template <typename T>
inline void Communicator::scatter (std::vector<T> & data, std::vector<std::size_t> counts, std::vector<T> & recv, const unsigned int root_id=0) const;
```

jwpeterson:
This seems fine to me...

roystgnr:
This all looks good exactly as-is.

[...]

PR #1062, opened 2016-08-11
Implementing wrappers for MPI_Scatter operations

permcody:
First set of wrappers. Hopefully I've put in enough libMesh asserts and other code to make @roystgnr happy.

### DIFF
--- a/include/parallel/parallel.h
+++ b/include/parallel/parallel.h
@@ -1163,6 +1163,39 @@ public:
 
   //-------------------------------------------------------------------
   /**
+   * Take a vector of local variables and scatter the ith item to the ith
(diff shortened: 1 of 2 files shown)


permcody:
@roystgnr I saw your comments on #1058 after I had already created the patch. I went with `int` instead of `std::size_t` for buffer lengths for consistency with existing MPI and other methods throughout `parallel.h`. I would vote that we keep it that way for now. It'll be easy to change them all at once if/when MPI updates their API.

permcody:
Hoping to get this merged so I can use it in my referenced PR!

roystgnr:
Looks fine to me, unless others are more hardass about size_t than I am.

PR #1064, opened 2016-08-15
Algebraic ghosting functors

roystgnr:
DO NOT MERGE - I'm just noticing that the complexity of this is growing, some API changes are going to delay it, and I'd like Civet to let me know if I've already screwed anything up.

When this branch is done it should complete the work for #1028

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -214,7 +214,8 @@ am__libmesh_dbg_la_SOURCES_DIST = src/base/default_coupling.C \
 	src/base/libmesh_version.C src/base/periodic_boundaries.C \
 	src/base/periodic_boundary.C src/base/periodic_boundary_base.C \
 	src/base/print_trace.C src/base/reference_counted_object.C \
-	src/base/reference_counter.C src/base/sparsity_pattern.C \
+	src/base/reference_counter.C src/base/single_predicates.C \
+	src/base/sparsity_pattern.C \
 	src/error_estimation/adjoint_refinement_estimator.C \
 	src/error_estimation/adjoint_residual_error_estimator.C \
 	src/error_estimation/discontinuity_measure.C \
@@ -503,6 +504,7 @@ am__objects_1 = src/base/libmesh_dbg_la-default_coupling.lo \
 	src/base/libmesh_dbg_la-print_trace.lo \
 	src/base/libmesh_dbg_la-reference_counted_object.lo \
 	src/base/libmesh_dbg_la-reference_counter.lo \
+	src/base/libmesh_dbg_la-single_predicates.lo \
 	src/base/libmesh_dbg_la-sparsity_pattern.lo \
 	src/error_estimation/libmesh_dbg_la-adjoint_refinement_estimator.lo \
 	src/error_estimation/libmesh_dbg_la-adjoint_residual_error_estimator.lo \
@@ -927,7 +929,8 @@ am__libmesh_devel_la_SOURCES_DIST = src/base/default_coupling.C \
 	src/base/libmesh_version.C src/base/periodic_boundaries.C \
 	src/base/periodic_boundary.C src/base/periodic_boundary_base.C \
 	src/base/print_trace.C src/base/reference_counted_object.C \
-	src/base/reference_counter.C src/base/sparsity_pattern.C \
+	src/base/reference_counter.C src/base/single_predicates.C \
+	src/base/sparsity_pattern.C \
 	src/error_estimation/adjoint_refinement_estimator.C \
 	src/error_estimation/adjoint_residual_error_estimator.C \
 	src/error_estimation/discontinuity_measure.C \
@@ -1215,6 +1218,7 @@ am__objects_2 = src/base/libmesh_devel_la-default_coupling.lo \
 	src/base/libmesh_devel_la-print_trace.lo \
 	src/base/libmesh_devel_la-reference_counted_object.lo \
 	src/base/libmesh_devel_la-reference_counter.lo \
+	src/base/libmesh_devel_la-single_predicates.lo \
 	src/base/libmesh_devel_la-sparsity_pattern.lo \
 	src/error_estimation/libmesh_devel_la-adjoint_refinement_estimator.lo \
 	src/error_estimation/libmesh_devel_la-adjoint_residual_error_estimator.lo \
@@ -1636,7 +1640,8 @@ am__libmesh_oprof_la_SOURCES_DIST = src/base/default_coupling.C \
 	src/base/libmesh_version.C src/base/periodic_boundaries.C \
 	src/base/periodic_boundary.C src/base/periodic_boundary_base.C \
 	src/base/print_trace.C src/base/reference_counted_object.C \
-	src/base/reference_counter.C src/base/sparsity_pattern.C \
+	src/base/reference_counter.C src/base/single_predicates.C \
+	src/base/sparsity_pattern.C \
 	src/error_estimation/adjoint_refinement_estimator.C \
 	src/error_estimation/adjoint_residual_error_estimator.C \
 	src/error_estimation/discontinuity_measure.C \
@@ -1924,6 +1929,7 @@ am__objects_3 = src/base/libmesh_oprof_la-default_coupling.lo \
 	src/base/libmesh_oprof_la-print_trace.lo \
 	src/base/libmesh_oprof_la-reference_counted_object.lo \
 	src/base/libmesh_oprof_la-reference_counter.lo \
+	src/base/libmesh_oprof_la-single_predicates.lo \
 	src/base/libmesh_oprof_la-sparsity_pattern.lo \
 	src/error_estimation/libmesh_oprof_la-adjoint_refinement_estimator.lo \
 	src/error_estimation/libmesh_oprof_la-adjoint_residual_error_estimator.lo \
@@ -2345,7 +2351,8 @@ am__libmesh_opt_la_SOURCES_DIST = src/base/default_coupling.C \
 	src/base/libmesh_version.C src/base/periodic_boundaries.C \
 	src/base/periodic_boundary.C src/base/periodic_boundary_base.C \
 	src/base/print_trace.C src/base/reference_counted_object.C \
-	src/base/reference_counter.C src/base/sparsity_pattern.C \
+	src/base/reference_counter.C src/base/single_predicates.C \
+	src/base/sparsity_pattern.C \
 	src/error_estimation/adjoint_refinement_estimator.C \
 	src/error_estimation/adjoint_residual_error_estimator.C \
 	src/error_estimation/discontinuity_measure.C \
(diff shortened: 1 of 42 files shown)


pbauman:
http://ci.memecdn.com/888/3505888.jpg

roystgnr:
Looks like a performance regression; I'll get to a good break point and commit, then back up and fix the performance issues shortly.

roystgnr:
Well *that* definitely isn't a simple performance regression.  I can't imagine what could be killing all the MOOSE tests but passing all the libMesh tests, but I'll see if I can replicate.

roystgnr:
Ah; it's because libMesh doesn't have enough coverage of CouplingMatrix use, just a few unit tests.  We ought to fix the missing coverage eventually, but for now MOOSE is making the problem easy to replicate so I'll just fix the CouplingMatrix codepath.

roystgnr:
Looks like the problem wasn't in CouplingMatrix, but in my attempt to dereference an end() iterator in a not-as-clever-as-I-thought-it-was dual-purpose loop.  Easy to factor out and fix.

There's a separate problem triggered by MOOSE, though: ActionWarehouse destroys the FEProblem before destroying the MooseMesh, which doesn't work as soon as we have problem components (like the DofMap) whose destructors try to do cleanup of objects registered with a MeshBase.  I'll put in a PR for MOOSE with the tiny fix once I'm done testing it and making sure nothing else crops up.

roystgnr:
The new ghosting functor APIs should all be complete and working at this point, and both MOOSE and GRINS seem happy, so we can merge this at any time if anybody else wants to play with it right away, or if anyone has extensive MeshCommunication or DofMap-affecting changes upcoming that might conflict.

If neither of those are the case, then I'd prefer waiting until the new iterator APIs, the test and example codes, and the extra utility classes are done too.

friedmud:
@roystgnr : looking at the `DefaultCoupling` object... it looks pretty hardcore.  Any chance you could genericize it a bit for the common case of wanting more than one level of ghosting (that's what both @permcody and myself are after).  I'm interested to see how that would work with your API...

jwpeterson:
As long as you don't add any more lines...

roystgnr:
That timeout might be a fluke or might be the cost of an additional DefaultCoupling invocation.  I'll see if I can replicate and if I can optimize.

roystgnr:
Sure looks like a fluke to me.  DefaultCoupling::operator() is 0.30% of runtime; the rest of the modified add_neighbors_to_send_list is an additional 0.20%.

roystgnr:
We're now at the point where we can do contact problems via the new system, including with refinement and repartitioning, including on DistributedMesh.  Time to rebase, though; there are some recent fixes which needed to go directly into master but which need to be applied to my tests here too.

roystgnr:
Damn, @jwpeterson, did you say "don't add anymore lines"?

:sweat_smile:

In my defense, the evaluable element iterators weren't in yet, which was half the point of the exercise.  And while I was at it I noticed that we were getting really close to getting PeriodicBoundaries problems working seamlessly on DistributedMesh too.  That's almost done, and then I don't think @friedmud's request will be much work, and after that there'll be nothing left to write except unit tests, which I probably won't go overboard on now that we've got test coverage of non-default functors in both the contact and the HP refinement examples.

jwpeterson:
> On Sep 2, 2016, at 4:16 PM, roystgnr <notifications@github.com> wrote:
> 
> Damn, @jwpeterson, did you say "don't add anymore lines"?
> 

This was just a reference to the number +1337... as in, you are an 1337 haXor.

roystgnr:
Rebased, with conflicts fixed.

This is getting big enough that I may merge tomorrow or this weekend, before more conflicts have the opportunity to arise.  (assuming I haven't regressed anything in those last dozen commits)  I'm gradually getting closer to fixing the remaining DistributedMesh corner case bugs, but those were bugs which exist in master too and so aren't worth holding up this PR, whose only fault is that it gave me the opportunity to trigger them.

roystgnr:
That last batch of commits adds and tests the capability @friedmud was asking for.  If there are no objections or regressions I'll merge tonight.

friedmud:
I definitely like the `n_levels` capability!

I suppose I should pull this down and try it with my new parallel checkpoint stuff...

Let me do that...

friedmud:
@roystgnr One question about `n_levels`: If I'm creating a parallel mesh file (like my new parallel Checkpoint stuff) do I need to put all of the levels I would want ghosted into the mesh file?  Or, will this new capability automatically ghost over more levels?

Right now, I'm putting just one level of elements around the local elements into the mesh file itself...

roystgnr:
Hmm... right now you do need to start from a mesh that's in an "everything which needs to be topologically ghosted is" state, with the bare exception that we've got some code in MeshCommunication for gathering one layer of ghosted cells around an initially local-elements-only mesh.

friedmud:
Ok.  So I may want to add the ability to my `splitter` code to output arbitrary levels of ghosting...

I suppose that code will look just like what you've already written here... hmmm.

roystgnr:
In general I don't think there's any getting around such a requirement...  If processor A has element A1 and processor B has element B1 and your ghosting algorithm can't compute true or false for "A1 and B1 are connected" unless it's got both at once to look at, then there's a chicken-and-egg problem w.r.t. getting A1 and B1 on the same processor.

In specific cases, we may be able to work up some all-to-all code that handles "elements within distance d of each other might be ghosted" or "elements within N neighbor levels of each other might be ghosted" and then anyone whose ghosting rules are a subset of one or the other or the union of those would be okay.

friedmud:
Sure - not a problem.

Do you see any way to cleanly extract your algorithm from DefaultGhosting so I can use it for the splitter?  Maybe I just run the DefaultGhosting functor and disect the results?

friedmud:
Actually - that should work perfectly!

The way you've made `operator()` I should be able to pass in `pid_elements_begin()` for whatever PID I'm currently trying to output a mesh for... and then the `map` that comes back should have exactly what I want in it: the extra halo elements!

Awesome!

friedmud:
@roystgnr I just rebased my stuff on yours... and I can confirm that not only does my new stuff work (the parallel splitting and reading of CheckpointIO) but my application still works well in parallel as well.

Just one datapoint... but I thought you might want to hear back that nothing seems broken.

I'm now going to change my splitter code to use the new functor to find the halo elements... I'll report back about success (or not) on that :-)

friedmud:
Hmmm.... well.  I ran into something when trying to use DefaultCoupling for my stuff: 1 level is not enough.

I had coded CheckpointIO to use `point_neighbors()` to get all of the neighbors for all of the local elements... it looks like DefaultCoupling isn't doing that (it's only using true neighbors).

My application actually depends on those point neighbors (with ray tracing... sometimes a ray will go right through a vertex and end up in a point neighbor).

I was able to "fix" it by using `set_n_levels(2)`... but I'm wondering if that was really the intended behavior.  I thought I remembered that libMesh always ghosted point neighbors before (but I could be misremembering).

This might actually explain why I've had so much trouble trying to use ParallelMesh with my application...

At any rate: using 2 levels does get the job done... but is a bit wasteful.  I'll always have to use one more level of ghosting than I really need/want (depending on your definition of ghosting anyway).

One bonus though: the new code is quite a bit faster (`point_neighbors()` is slow!).

friedmud:
Actually - even 2 levels might not be enough, right?  With triangles and tets it's easy to end up with many elements sharing a vertex.  You may need a fairly high number of levels to guarantee that you capture all of the point neighbors...

permcody:
Interesting, I hadn't considered that case. 

I'm going to need the same thing for my halos too. My case is slightly different than Derek's in that I only need geometric information, not FE or solution related information on those halos. That means that everything already works just fine for the ReplicatedMesh case for me, but soon I'd like to get my algorithms running on DistributedMesh, so this is of interest to me too.

friedmud:
Maybe `DefaultCoupling` needs an option to use `point_neighbors()`?  And / or maybe `DefaultCoupling` needs to expose some virtual functions we can override to modify its behavior a bit (i.e. `virtual void get_neighbors()`).

permcody:
Point neighbors is uber expensive on it's own. Hopefully we can call
something else that gives us an equivalent set of elements.

friedmud:
It's not that bad in this case.  It will only get called at the beginning of the simulation or once per timestep.  In my timing neighbors is about 2x faster than point_neighbors.  So it's not huge... but it is maybe worth having an option for.

One thing though: in MOOSE, how will we detect which one needs to be used?

permcody:
I think we'll have to think about each case a little bit to figure out how to make this clean in MOOSE. I imagine the decision will be a little different for Contact versus rDG. 

Daniel removed the point neighbors call out of the FeatureFloodCount object opting instead for using the node to elem map. It was a lot more in 2x savings in our case. However I was calling that routine on every element every time step.

friedmud:
Good point about rDG... it will definitely need to use `point_neighbors`.

I swear that `point_neighbors()` is what libMesh has always done before now.  We'll find out when @roystgnr responds :-)

friedmud:
Oh - BTW: the situation is worse in 3D.  With Hexes it would take 3 levels using neighbors to pick up all of the elements around the vertex of a 3D cube partition.

roystgnr:
What libMesh has always done is point neighbors of all elements, which in AMR is ridiculous overkill, but which actually made synchronization operations much simpler - reducing to neighbors of active elements (and their ancestors) is responsible for new hard-to-pin-down DistributedMesh bugs.

roystgnr:
I think point neighbors of active elements only ought to be a reasonable option, but I'll implement it as a new class not as a DefaultCoupling option.

friedmud:
Cool - thanks @roystgnr .

I'm actually going to make `CheckpointIO` take a `GhostingFunctor` to use to decide what gets written out to a file!  That way a user can store as much (or as little) as they want into the parallel decomposed mesh file.  Will make doing restart for `DistributedMesh` contact problems easier.

`CheckpointIO`, by default, will build a `DefaultCoupling` functor if no other one is specified.

roystgnr:
For CheckpointIO writes: shouldn't the default be to use the ghosting functors already attached to the Mesh?

The important thing will be CheckpointIO reads, though: in the long run I do want to have some facility for "sync ghosted elements via all to all communications using assumptions X and Y", and we'll want to have a way for the user to override that and say "this restart already has all the elements we need properly ghosted, so skip the expensive communication."

roystgnr:
The most consistent upgrade path would be "CheckpointIO/Nemesis/etc reads don't do any sync, but the user can call it manually", but that way leads to users who don't understand why their contact problem doesn't work reliably, so I'd rather go to "CheckpointIO/Nemesis/etc reads sync, and the user can disable it manually" instead.

friedmud:
I agree: the default should be to use the one attached to the Mesh.

I also agree that it would be nice to have some sort of "sync" capability.  I think that, for now, we can get away with just "over ghosting" during the mesh splitting phase... and then throwing away elements we don't need when the problem actually starts...

jwpeterson:
On Mon, Sep 26, 2016 at 7:26 AM, roystgnr <notifications@github.com> wrote:

> For CheckpointIO writes: shouldn't the default be to use the ghosting
> functors already attached to the Mesh?
>
> The important thing will be CheckpointIO reads, though: in the long run I
> do want to have some facility for "sync ghosted elements via all to all
> communications using assumptions X and Y", and we'll want to have a way for
> the user to override that and say "this restart already has all the
> elements we need properly ghosted, so skip the expensive communication."
>

You might want to double-check your editor settings, this merge introduced
several whitespace errors, mostly tabs:

git -c core.whitespace=trailing-space,space-before-tab,tab-in-indent diff
--check 29b7ab4..upstream/master -- '*.[Ch]'

-- 
John

roystgnr:
Damn, thanks!  I'll fix it now.

roystgnr:
And yeah, screwed up editor settings.  I was coding on a different system than usual and hadn't copied over my .vimrc.

PR #1066, opened 2016-08-16
Update bundled Eigen from 3.2.5 to 3.2.9.

jwpeterson:
This updates us from the 16 Jun 2015 version of Eigen to the latest
stable version as of 18 Jul 2016.

This compiles and runs the examples on my machine with
```
LIBMESH_OPTIONS=--use-eigen
```
set.


friedmud:
Nice!  I can definitely use this...

friedmud:
BTW: I just tested this.  It works with my new code that directly uses Eigen... and it's about 5% faster.  Not much... but at least it didn't go the other way ;-)

friedmud:
Thanks for doing the update!

PR #1067, opened 2016-08-19
Add vector of vector scatter implementation

permcody:
Slightly more handy interface.

### DIFF
--- a/include/parallel/parallel_implementation.h
+++ b/include/parallel/parallel_implementation.h
@@ -3078,19 +3078,26 @@ inline void Communicator::allgather(std::vector<T> & r,
                      &displacements[0], send_type, this->get()));
 }
 
+
+
 template <typename T>
 void Communicator::scatter(const std::vector<T> & data,
                            T & recv,
                            const unsigned int root_id) const
 {
+  libmesh_assert_less (root_id, this->size());
+
+  // Do not allow the root_id to scatter a NULL vector.
+  // That would leave recv in an indeterminate state.
+  libmesh_assert (this->rank() != root_id || this->size() == data.size());
+
   if (this->size() == 1)
     {
       libmesh_assert (!this->rank());
       libmesh_assert (!root_id);
+      recv = data[0];
       return;
     }
-
-  libmesh_assert_less (root_id, this->size());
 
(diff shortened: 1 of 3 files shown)


include/parallel/parallel_implementation.h
```diff
@@ -3191,6 +3191,44 @@ void Communicator::scatter(const std::vector<T> & data,
...
+void Communicator::scatter(const std::vector<std::vector<T> > & data,
+                           std::vector<T> & recv,
+                           const unsigned int root_id,
+                           const bool identical_buffer_sizes) const
+{
+  if (this->size() == 1)
+    {
+      libmesh_assert (!this->rank());
+      libmesh_assert (!root_id);
+      return;
```
roystgnr: This case should be std::copy from data[0] to recv, no?

include/parallel/parallel_implementation.h
```diff
@@ -3191,6 +3191,44 @@ void Communicator::scatter(const std::vector<T> & data,
...
+      for (unsigned int i=0; i < data.size(); ++i)
+        {
+          if (!identical_buffer_sizes)
+            counts[i] = data[i].size();
+          std::copy(data[i].begin(), data[i].end(), std::back_inserter(stacked_data));
+        }
+    }
+
+  if (identical_buffer_sizes)
+    this->scatter(stacked_data, recv, root_id);
```
roystgnr: We should probably have a loop (wrapped in ifndef NDEBUG) which asserts that the user wasn't lying about the identical buffer sizes.

include/parallel/parallel_implementation.h
permcody: Yes...  I missed that on the other implementations too. Too much copy and paste.

include/parallel/parallel_implementation.h
permcody: Will do

roystgnr:
That's all I saw; good to merge once fixed.

A unit test would be nice, but if we're going to start throwing around blame at people who write code without writing enough test coverage then I'm going to have to flee the country.

roystgnr:
Thanks!  In the future I'd say put the unit test in a separate commit.

PR #1068, opened 2016-08-24
TimeSolver Unit Tests

pbauman:
Referencing [this discussion](https://sourceforge.net/p/libmesh/mailman/libmesh-devel/thread/CAKiVBtt3HmhsAT6QK49-yj%3DKtFyV5eQ-aaJ3%2BBe8s-ZkziZfYw%40mail.gmail.com/#msg35128047), I'll be monkeying with the `FEMSystem` `TimeSolvers` shortly, so I wanted to drop in some unit tests.

There are some fixes coming in a subsequent PR for problems that I uncovered, but this PR establishes the base for the tests and tests for the things that do work. There's also an issue with using `SCALAR` variables here that I'll discuss in another issue.

PR #1069, opened 2016-08-24
FEMSystem TimeSolver Fixes

pbauman:
Builds on #1068.

There are several fixes here:

1. Only call `elem_fe_reinit` if there's mesh motion (3f3a757) (`elem_fe_reinit` wasn't inside the if-block).
2. Add requisite `reinit_func` calls in `Euler2Solver` (eed32a9) and `NewmarkSolver` (f28459). Also added comments clarifying that we're also setting the time in the context.
3. Add setters needed in `NewmarkSolver` (fde77ae)

Additional unit tests were added to cover testing the cases that were previously failing before the fixes.

### DIFF
--- a/src/systems/fem_context.C
+++ b/src/systems/fem_context.C
@@ -1305,18 +1305,20 @@ void FEMContext::elem_reinit(Real theta)
 
   // Handle a moving element if necessary.
   if (_mesh_sys)
-    // We assume that the ``default'' state
-    // of the mesh is its final, theta=1.0
-    // position, so we don't bother with
-    // mesh motion in that case.
-    if (theta != 1.0)
-      {
-        // FIXME - ALE is not threadsafe yet!
-        libmesh_assert_equal_to (libMesh::n_threads(), 1);
+    {
+      // We assume that the ``default'' state
+      // of the mesh is its final, theta=1.0
+      // position, so we don't bother with
+      // mesh motion in that case.
+      if (theta != 1.0)
+        {
+          // FIXME - ALE is not threadsafe yet!
+          libmesh_assert_equal_to (libMesh::n_threads(), 1);
 
-        elem_position_set(theta);
-      }
-  elem_fe_reinit();
+          elem_position_set(theta);
+        }
+      elem_fe_reinit();
+   }
 }
 
 
--- a/include/solvers/newmark_solver.h
+++ b/include/solvers/newmark_solver.h
@@ -192,7 +192,8 @@ protected:
                                   ResFuncType mass,
                                   ResFuncType damping,
                                   ResFuncType time_deriv,
-                                  ResFuncType constraint);
+                                  ResFuncType constraint,
+                                  ReinitFuncType reinit);
 };
 
 } // namespace libMesh
--- a/include/solvers/newmark_solver.h
+++ b/include/solvers/newmark_solver.h
@@ -141,6 +141,19 @@ public:
                                   DiffContext &) libmesh_override;
 
 
+  /**
+   * Setter for \f$ \beta \f$
+   */
+  void set_beta ( Real beta )
+  { _beta = beta; }
+
+  /**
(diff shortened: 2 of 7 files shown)


pbauman:
Da fuq? These changes should not have had *any* effect on MOOSE. The debug failure a problem on the MOOSE testing side?

roystgnr:
"Max time exceeded", so probably just an extra-long-test that happened to be scheduled on an already-loaded node?

pbauman:
I should've mentioned too that GRINS `make check` passes using the tip of this branch. (Looks for Civet on the idaholab page...)

permcody:
Roy is right however, this really shouldn't happen. We try to keep those
boxes within load specs but some of these tests are just right at the
tipping point. I'll invalidate and see if it happens again.

jwpeterson:
@permcody I don't know if you invalidated this already or not... if so it failed again, if not, well, now it's invalidated.

jwpeterson:
I have no issues with this.  @roystgnr?

Also, @pbauman do you mind if I rebase this before merging since your first PR was merged successfully?  No reason to have those same commits in the history twice.

pbauman:
I'll rebase and push here.

pbauman:
Done. But it wouldn't have put the same commits in the history twice would it? They were the same hashes (before rebasing), I'd just created this branch from the tip of the previous one and I thought GitHub was just stupid. Am I the stupid one?

jwpeterson:
You're right, it would have since those hashes didn't change.

src/systems/fem_context.C
```diff
@@ -1305,18 +1305,20 @@ void FEMContext::elem_reinit(Real theta)
...
+          libmesh_assert_equal_to (libMesh::n_threads(), 1);
 
-        elem_position_set(theta);
-      }
-  elem_fe_reinit();
+          elem_position_set(theta);
+        }
+      elem_fe_reinit();
+   }
 }
```
roystgnr: This is good, but the doxygen comment for elem_reinit should probably be made clearer.

pbauman: You mean adding the note about setting the time?

roystgnr: Nah, I mean making it clearer that we're not reinitializing FE objects unless a moving mesh requires it.

Originally elem_reinit was *supposed* to reinitialize FE objects; and if you look at the documentation in DiffSystem that's clearly implied.  It wasn't until later optimization that I broke out elem_fe_reinit() as a separate method.

roystgnr:
Everything's fine by me except for the now-atavistic elem_reinit() documentation.  @pbauman can take a crack at rewriting that or I can try myself late tonight.

fe32f8b is technically a big API change, but I suspect the changing API only ever gets used in the TimeSolver subclasses that @pbauman is currently in the process of fixing.

src/systems/fem_context.C
pbauman: > Nah, I mean making it clearer that we're not reinitializing FE objects unless a moving mesh requires it.

Ah,OK.

> Originally elem_reinit was supposed to reinitialize FE objects; and if you look at the documentation in DiffSystem that's clearly implied. It wasn't until later optimization that I broke out elem_fe_reinit() as a separate method.

Thanks for the clarification. I don't remember a time elem_fe_reinit wasn't called in the AssemblyContributions functor. But my memory has never been the same since my kids were born.

pbauman:
> Everything's fine by me except for the now-atavistic elem_reinit() documentation. @pbauman can take a crack at rewriting that or I can try myself late tonight.

I'll drop a PR in shortly.

> fe32f8b is technically a big API change

Huh, that's a good point. I'd assumed it was a bug since elem_fe_reinit was called elsewhere and the <side,edge>_reinit versions had the <side,edge>_fe_reinit in the if-block.

> but I suspect the changing API only ever gets used in the TimeSolver subclasses

I'd be shocked if this wasn't true.

PR #1071, opened 2016-08-24
Don't assume interior_parents() are always part of the same Mesh.

jwpeterson:
The Metis partitioner code assumed this, but it's not true for e.g. BoundaryMeshes that have `interior_parent()` in another mesh entirely.

### DIFF
--- a/src/partitioning/metis_partitioner.C
+++ b/src/partitioning/metis_partitioner.C
@@ -399,20 +399,31 @@ void MetisPartitioner::_do_partition (MeshBase & mesh,
                 std::set<const Elem *>::iterator n_it = neighbor_set.begin();
                 for (; n_it != neighbor_set.end(); ++n_it)
                   {
-                    // FIXME - non-const versions of the Elem set methods
-                    // would be nice
-                    Elem * neighbor = const_cast<Elem *>(*n_it);
+                    const Elem * neighbor = *n_it;
 
                     // Not all interior neighbors are necessarily in
-                    // the global_index_map.  This will be the case
-                    // when partitioning a BoundaryMesh, whose
-                    // elements all have interior_parents() that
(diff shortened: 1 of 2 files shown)


jwpeterson:
@xikaij

src/partitioning/metis_partitioner.C
```diff
@@ -403,8 +403,16 @@ void MetisPartitioner::_do_partition (MeshBase & mesh,
...
-                    csr_graph(elem_global_index, connection++) =
-                      global_index_map[neighbor->id()];
+                    // Not all interior neighbors are necessarily in
+                    // the global_index_map.  This will be the case
+                    // when partitioning a BoundaryMesh, whose
+                    // elements all have interior_parents() that
+                    // belong to some other Mesh.
+                    vectormap<dof_id_type, dof_id_type>::iterator global_index_map_it =
+                      global_index_map.find(neighbor->id());
+
```
roystgnr: This isn't a sufficient fix, though, is it?  The find will succeed if the current mesh has an element whose ID is the same as the interior_parent, whether or not that element *is* the interior_parent.  To actually be sure we've got the interior_parent on the same mesh we'd have to query the mesh for the element having that ID and make sure the pointers match.

jwpeterson: In my case, the elements of the `BoundaryMesh` were numbered in a non-overlapping way from their `interior_parents()`, but I suppose that won't be the case in general, especially if it later gets renumbered.  I'll make a more thorough fix...

roystgnr:
Looks good now, thanks.  :+1: from me once it passes MOOSE tests, but please wait for @pbauman in case he wants to be paranoid and run it by some multi-dimensional GRINS tests too.

pbauman:
GRINS is happy. Thanks!

PR #1072, opened 2016-08-25
Remove memory churn in compute_affine_map

friedmud:
I've been using Open SpeedShop (OSS: https://openspeedshop.org if you don't know about it you should!) to do some threading optimization and `free()` kept showing up as I went to more and more threads (showing, once again, that memory churn is the death of threading).

I was able to get OSS to show me the "hot path" to `free()` that was generating the churn... and it was in `FEMap::compute_affine_map()`.  A quick inspection later and I spotted the `std::vector<const Node*>` getting created and destroyed...

To fix it I've just added it as a persistent member variable.  Works perfectly!

Here are the timings before and after this fix for 20 "residual evaluations" (in a toy code) using a 2000x2000 generated mesh:

| threads | original (s) | optimized (s) |
|---------|----------|-----------|
| 1       | 55.96    | 52.50     |
| 2       | 36.58    | 29.96     |
| 4       | 18.32    | 19.33     |
| 8       | 21.76    | 16.61     |
| 12      | 18.89    | 9.96      |
| 16      | 17.53    | 7.81      |
| 24      | 18.86    | 6.10      |

And a plot (why not?):

affine_map_optimization

Here is what I saw in Open SpeedShop on 12 threads for this problem:

affine_map_origin

And here's what the same run looks like on 12 procs now:

affine_map_optimized

Note: `OldVar` is the name of the testing object I was using

### DIFF
--- a/include/fe/fe_map.h
+++ b/include/fe/fe_map.h
@@ -884,6 +884,11 @@ private:
    * compute second derivatives of the inverse map.
    */
   void compute_inverse_map_second_derivs(unsigned p);
+
+  /**
+   * Work vector for compute_affine_map()
+   */
+  std::vector<const Node *> elem_nodes;
 };
 
 }
--- a/src/fe/fe_map.C
+++ b/src/fe/fe_map.C
@@ -1235,8 +1235,9 @@ void FEMap::compute_affine_map(const unsigned int dim,
   this->resize_quadrature_map_vectors(dim, n_qp);
 
   // Determine the nodes contributing to element elem
-  std::vector<const Node *> elem_nodes(elem->n_nodes(), libmesh_nullptr);
-  for (unsigned int i=0; i<elem->n_nodes(); i++)
+  unsigned int n_nodes = elem->n_nodes();
+  elem_nodes.resize(elem->n_nodes());
+  for (unsigned int i=0; i<n_nodes; i++)
     elem_nodes[i] = elem->node_ptr(i);
 
   // Compute map at quadrature point 0


jwpeterson:
Seems like a good fix to me, this function should never be called by multiple threads on the same object, so there should be no issues with making that a class variable...

roystgnr:
Awesome catch, but can I ask some stupid questions?

Does the same problem apply when considering vector declarations inside vs outside loops, or is the compiler smart enough to optimize away the implied redundant free-then-malloc-again-right-away?  My coding style is to declare variables in the innermost scope possible, but if that's going to be slowing us down by a factor of 3 on many-core CPUs then it's time to rethink that style.

Could we have worked around the problem by instead using a non-default allocator or a sufficiently large array or std::array?  Something that uses ~450 bytes of memory on the stack would end up never calling heap management at all, right?  But then we'd either need our helper functions to only take vectors specialized with that allocator or we'd need them to take pointer instead of vector arguments.

Are the standard malloc and free sufficiently thread-aware?  I'd have expected an order of magnitude less slowdown from memory management as simple as "chop up virtual memory into thread-specific spaces, don't edit anyone else's space during malloc, save the slow code paths for when one thread is asked to free another thread's malloc output".

permcody:
On Fri, Aug 26, 2016 at 8:35 AM roystgnr <notifications@github.com> wrote:

> Awesome catch, but can I ask some stupid questions?
>
> Does the same problem apply when considering vector declarations inside vs
> outside loops, or is the compiler smart enough to optimize away the implied
> redundant free-then-malloc-again-right-away? My coding style is to declare
> variables in the innermost scope possible, but if that's going to be
> slowing us down by a factor of 3 on many-core CPUs then it's time to
> rethink that style.
>
> Could we have worked around the problem by instead using a non-default
> allocator or a sufficiently large array or std::array? Something that uses
> ~450 bytes of memory on the stack would end up never calling heap
> management at all, right? But then we'd either need our helper functions to
> only take vectors specialized with that allocator or we'd need them to take
> pointer instead of vector arguments.
>

I'm sure that any of those options would work but now you are writing
slightly more complex and slightly less elegant code. Any trip to the
memory manager in threaded or non-threaded regions is always slow,
sometimes due to the necessary locking to make it thread safe. As a first
take on any optimization in a C++ algorithm we should probably strive to
reduce trips to the allocator whenever possible even if it means wasting a
little space. Also, I like the idea of creating more stack variables and
using the heap a little less again, that's got to help a ton in threaded
code. My guess is, it'll remove some of the "False Sharing" crap Derek is
seeing now.


> Are the standard malloc and free sufficiently thread-aware? I'd have
> expected an order of magnitude less slowdown from memory management as
> simple as "chop up virtual memory into thread-specific spaces, don't edit
> anyone else's space during malloc, save the slow code paths for when one
> thread is asked to free another thread's malloc output".
>

I'm believe that "new" and "delete" are typically thread-safe but I'm
pretty sure that standard definitions of malloc and free are not since they
are ANSI-C.

friedmud:
@permcody got this right... whenever possible we need to remove ANY amount of allocation / deallocation in compute loops (regardless of threading).

Memory churn slows things down even in single-threaded applications but with threading it hammers you (as you can see here).

>  is the compiler smart enough to optimize away the implied redundant free-then-malloc-again-right-away

As far as I know: no... not for a std::vector

> Could we have worked around the problem by instead using a non-default allocator or a sufficiently large array or std::array? Something that uses ~450 bytes of memory on the stack would end up never calling heap management at all, right? But then we'd either need our helper functions to only take vectors specialized with that allocator or we'd need them to take pointer instead of vector arguments.

I believe so (for the large stack array): as long as it's not too large.  But I agree with @permcody that this is inelegant and overly complex.

As for allocators... "scalable allocators" _are_ a thing.  But they are complex, and as far as I know there are none built in.  TBB comes with some you can use: https://www.threadingbuildingblocks.org/tutorial-intel-tbb-scalable-memory-allocator

But, again, that's overly complex for just a simple little vector like this.  Use of these should be relegated to the "only if you really have to" bin.

(Note: That page also talks about "False Sharing" a bit... which is ANOTHER threading problem I've been fighting lately...)

> I'm believe that "new" and "delete" are typically thread-safe but I'm
pretty sure that standard definitions of malloc and free are not since they
are ANSI-C.

They are all thread safe (even `malloc`) (if they weren't we would be a SHIT TON of trouble!)... and is one of the major reasons why they're slow.  Essentially they all use some form of locking to maintain thread safety... that is a global lock... so every call to `new` is contending for the same lock.  If a couple of calls to `new` are in compute sensitive areas then this can effectively serialize your code!

PR #1075, opened 2016-08-25
Add support for --drop-cerr command line argument.

jwpeterson:
There was also a libmesh-devel thread discussing this: http://tinyurl.com/hseqsm4

### DIFF
--- a/src/base/libmesh.C
+++ b/src/base/libmesh.C
@@ -595,6 +595,14 @@ LibMeshInit::LibMeshInit (int argc, const char * const * argv,
     if (!libMesh::on_command_line ("--keep-cout"))
       libMesh::out.rdbuf (libmesh_nullptr);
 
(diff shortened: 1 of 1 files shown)


jwpeterson:
@permcody

permcody:
Hmm, I don't think I like the name. Maybe `--drop-parallel-cerr`, `--drop-other-cerr` `--drop-slave-cerr`, etc

permcody:
`--keep-root-cerr`?

jwpeterson:
It goes along with the existing `--keep-cout` option.  It's like the converse of that.

permcody:
Fine you win...
image

permcody:
We probably really should add something to the the libMesh abort handler (or whatever that is) that says to rerun without --drop-cerr.

PR #1077, opened 2016-08-25
Better documentation for FEMContext::<elem,side,edge>_reinit

pbauman:
Here's my stab at the doc update mentioned in #1069.


roystgnr:
Works for me, thanks.

PR #1078, opened 2016-08-26
Add BoundaryVolumeSolutionTransfer class.

jwpeterson:
This class can be used for transferring solutions between the surface
of a volume mesh and the BoundaryMesh associated with that surface.
This is joint work with with Xikai Jiang from Argonne National
Laboratory (@xikaij) who is the original author of the transfer.

See also: X. Jiang et al., "An O(N) and parallel approach to integral
problems by a kernel-independent fast multipole method: Application to
polarization and magnetization of interacting particles," The Journal
of Chemical Physics vol. 145, 064307, http://dx.doi.org/10.1063/1.4960436.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -425,6 +425,7 @@ am__libmesh_dbg_la_SOURCES_DIST = src/base/dof_map.C \
 	src/reduced_basis/transient_rb_construction.C \
 	src/reduced_basis/transient_rb_evaluation.C \
 	src/reduced_basis/transient_rb_theta_expansion.C \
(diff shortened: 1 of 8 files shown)


jwpeterson:
This has been tested and works well in the [application code](https://bitbucket.org/mesoscience/infinitepoisson) so I'm going to go ahead and merge, will add a unit test/example momentarily, as that may take a bit longer to gin up.

xikaij:
Thanks! I'm OK with it.

mangerij:
Woo!

PR #1080, opened 2016-08-31
Don't override user refinement settings

roystgnr:
Thanks to Harriet Li for catching this: our internal MeshRefinement
(to enable the first-coarsen-then-refine memory optimization) was
getting default level restrictions set, which would then override any
relaxing of those restrictions which users had done.

### DIFF
--- a/src/systems/equation_systems.C
+++ b/src/systems/equation_systems.C
@@ -230,9 +230,11 @@ void EquationSystems::reinit ()
     dof_constraints_created = true;
   }
 
-  // FIXME: Where should the user set maintain_level_one now??
-  // Don't override previous settings, for now
+  // Don't override any user refinement settings
   MeshRefinement mesh_refine(_mesh);
+  mesh_refine.face_level_mismatch_limit() = 0; // unlimited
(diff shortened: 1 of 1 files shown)


pbauman:
Sorry to hijack this, but I was looking at this code a couple of weeks ago and I don't understand the refine and coarsen calls inside EquationSystems::reinit(). All the adaptivity examples show calling the refinement before EquationSystems::reinit() - why call them again?

roystgnr:
Ben had the great idea long long ago that we could save a lot of memory (in a few cases; at the time we were working on a few different kinds of transient moving front problems?) by doing coarsening first, then refinement.  In libMesh, that can be accomplished by flagging elements but then just calling EquationSystems::reinit() and letting it do all the refinement and coarsening.

In hindsight I think this was only technically backwards-compatible (if the user did the AMR/C already then you're just burning CPU needlessly) and probably pointless (the memory wasted when doing AMR/C all at once is probably less than the memory which can be freed by deallocating old sparse matrices) idea, and making things sane eventually is buried on my todo list somewhere.

PR #1082, opened 2016-09-02
Fix tests

roystgnr:
@permcody and @pbauman heard I like testing for bugs so they put bugs in our tests for bugs so I can test while I test.

This PR gets the new tests from #1067 and #1068 working in parallel.

I plan to merge this ASAP so I can rebase #1064 again; I really need to get that finished up and start working nearer to master again.

### DIFF
--- a/tests/parallel/parallel_test.C
+++ b/tests/parallel/parallel_test.C
@@ -145,7 +145,7 @@ public:
 
         for (unsigned int i=0; i<src.size(); i++)
           src[i] = i;
-        for (unsigned int i=0; i<src.size(); i++)
+        for (unsigned int i=0; i<TestCommWorld->size(); i++)
           counts[i] = static_cast<int>(i+1);
       }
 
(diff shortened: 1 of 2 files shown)


tests/parallel/parallel_test.C
```diff
@@ -145,7 +145,7 @@ public:
 
         for (unsigned int i=0; i<src.size(); i++)
           src[i] = i;
-        for (unsigned int i=0; i<src.size(); i++)
+        for (unsigned int i=0; i<TestCommWorld->size(); i++)
```
permcody: Yikes! Sorry about that! Buggy tests are better than no tests? 😄

roystgnr: Buggy tests are definitely better than no tests.  We probably ought to have Civet running all our tests in parallel, though; that's a common enough use case to be worth catching at the source.

roystgnr:
Even the debug mode checks made it past the unit tests, so I'm going to assume that changes to unit tests won't break anything else and merge right away.

PR #1083, opened 2016-09-06
FEInterface::map() did fail for infinite elements

BalticPinguin:
I figured out that FEInterface::map() does not work for infinite elements. Supposing this is not a feature, I added the respective functions. Therefore, I needed to change InfFE::map() from protected to public, but I don't see problems with it. Are there?

Best,
Hubert

### DIFF
--- a/include/fe/fe_interface.h
+++ b/include/fe/fe_interface.h
@@ -394,6 +394,11 @@ private:
                               const std::vector<Number> & elem_soln,
                               std::vector<Number> & nodal_soln);
 
+  static Point ifem_map (const unsigned int dim, 
+                         const FEType& fe_t,
+                         const Elem * elem,
+                         const Point & p);
+
   static Point ifem_inverse_map (const unsigned int dim,
                                  const FEType & fe_t,
                                  const Elem * elem,
--- a/src/fe/fe_interface.C
+++ b/src/fe/fe_interface.C
@@ -539,6 +539,11 @@ Point FEInterface::map(unsigned int dim,
                        const Elem * elem,
                        const Point & p)
 {
+#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+
+  if (is_InfFE_elem(elem->type()))
+     return ifem_map(dim, fe_t, elem, p); //new function now.
+#endif
   fe_with_vec_switch(map(elem, p));
 
   libmesh_error_msg("We'll never get here!");
--- a/src/fe/fe_interface_inf_fe.C
+++ b/src/fe/fe_interface_inf_fe.C
@@ -425,6 +425,33 @@ void FEInterface::ifem_nodal_soln(const unsigned int dim,
 
 
 
+Point FEInterface::ifem_map (const unsigned int dim, 
+                             const FEType& fe_t,
+                             const Elem * elem,
+                             const Point & p)
+{
+   switch (fe_t.inf_map)
+   {
+      case CARTESIAN:{
+         switch (dim)
+         {
+         case 1:
+            return InfFE<1,JACOBI_20_00,CARTESIAN>::map(elem, p);
+         case 2:
+            return InfFE<2,JACOBI_20_00,CARTESIAN>::map(elem, p);
+         case 3:
+            return InfFE<3,JACOBI_20_00,CARTESIAN>::map(elem, p);
+         default:
+            libmesh_error_msg("Invalid dim = " << dim);
+         }
+      }
+      case SPHERICAL:
+      case ELLIPSOIDAL:
+         libmesh_not_implemented_msg("ERROR: Spherical and Ellipsoidal IFEMs not (yet) implemented.");
+      default:
+         libmesh_error_msg("Invalid map = " << fe_t.inf_map);
+   }
+}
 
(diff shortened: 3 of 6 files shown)


src/fe/inf_fe_map.C
```diff
@@ -486,17 +486,18 @@ Point InfFE<Dim,T_radial,T_map>::inverse_map (const Elem * inf_elem,
...
-          //  }
+           if (p(2)+1 < tolerance)
+            {
+           libmesh_here();
+           libMesh::err << "WARNING: radial distance p(2) is "
+                << p(2)
+                << std::endl;
+            }
+          */
 #endif
```
roystgnr: Why the commenting changes?

roystgnr:
I'd like more eyes on it before we merge, but looks reasonable to me, thanks.

BalticPinguin:
I uncommented it for my personal use and than inserted them in the wrong way, so you can ignore them, sorry for that.

By the way: If you want, I can write an extra report on that but I get there errors when inverse_mapping and mapping again of up to  1.13449 (or 0.2/0.3 ) at some points. For me this seems to be an unreasonable large error, isn't it?

src/fe/inf_fe_map.C
jwpeterson: Please either undo the commenting change or simply delete the code altogether.  There's no reason for this to be part of the diff...

src/fe/inf_fe_map.C
```diff
@@ -592,18 +593,16 @@ Point InfFE<Dim,T_radial,T_map>::inverse_map (const Elem * inf_elem,
...
-    }
-  */
+    //if (diff.size() > tolerance)
+    //{
+    //   libmesh_here();
+    //   libMesh::err << "WARNING:  diff is "
+    //   << diff.size()
+    //   << std::endl;
+    //}
 #endif
```
jwpeterson: Same goes for this commenting change.

src/fe/fe_interface_inf_fe.C
```diff
@@ -425,6 +425,30 @@ void FEInterface::ifem_nodal_soln(const unsigned int dim,
 
 
 
+Point FEInterface::ifem_map (const unsigned int dim, 
+                             const FEType& fe_t,
+                             const Elem * elem,
+                             const Point & p){
+   switch (fe_t.inf_map){
```
jwpeterson: Curly braces in this entire function need to go on their own line to match our coding conventions.

src/fe/fe_interface.C
```diff
@@ -539,6 +539,11 @@ Point FEInterface::map(unsigned int dim,
                        const Elem * elem,
                        const Point & p)
 {
+#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+
+  if (is_InfFE_elem(elem->type()))
```
jwpeterson: I hadn't noticed this before, but this function's name does not match our usual naming scheme... that should probably be fixed, but not in this PR.

include/fe/fe_interface.h
```diff
@@ -394,6 +394,11 @@ class FEInterface
                               const std::vector<Number> & elem_soln,
                               std::vector<Number> & nodal_soln);
 
+  static Point ifem_map (const unsigned int dim, 
+                                      const FEType& fe_t,
```
jwpeterson: Indentation is not correctly aligned for the new function.

jwpeterson:
> when inverse_mapping and mapping again of up to 1.13449 (or 0.2/0.3 ) at some points. For me this seems to be an unreasonable large error, isn't it?

Yes, this sounds like a bug.  Can you get this result prior to making your changes, or is it something introduced by this patch?  A minimal working example posted as a gist would be helpful, but I'm not sure any of us would really be able to debug the math any time soon...

BalticPinguin:
Ok, I will do the respective changes in the patch. 
The error in the mapping seems should not come from my code since I don't touch the calculations itself; I will investigate it more and than write a report/patch for it separately.

jwpeterson:
> Ok, I will do the respective changes in the patch.

Thanks.  Just do `git commit --amend` after making the formatting changes, and then `git push -f` to update the PR.



> The error in the mapping seems should not come from my code since I don't touch the calculations itself; I will investigate it more and than write a report/patch for it separately.

OK, it's possible that it is a long-standing bug.  If you can get it resolved, we should also add some basic unit testing for the infinite elements code path...

BalticPinguin:
sorry for the many commits; I missed the format of the comments.

jwpeterson:
> sorry for the many commits; I missed the format of the comments.

No problem, just use [interactive rebase](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) to squash them down into a single commit. 

Also, don't forget to fix the opening curly braces in fe_interface_inf_fe.C.  They should all be on a separate line.

BalticPinguin:
nice! Thanks for that hint.
Now it looks fine, if I didn't miss something.

jwpeterson:
Thanks for making the formatting changes.  I fixed a couple others and merged, but keep us up to date about the issues you were having with `ifem_map`.

Issue #1084, opened 2016-09-07
large errors in InfFE::inverse_map()


BalticPinguin:
Hi all,
as discussed in my PR yesterday, here is a description of the error in the inverse_map()/map() function.
As code example I use miscellaneous_ex1.C and added a function map_remap() that runs over the quadrature points and prints them together with the map(inverse_map()) of them.
In addition, I have the output I get for it in two versions: First (Interpol.out) as you will get it from the current libMesh version and in the second one I enforced interpolated=false in InfFE::inverse_map() (src/fe/inf_fe_map.C:108).
Code and output in the gist:
https://gist.github.com/BalticPinguin/09e4b3090a87d4c5d8af15255b1fd8dc

BalticPinguin:
So as I see it, the default scheme used to obtain the 3-rd component gives large errors, while the scheme with Newton iteration works quite well. 
As I see it (without having had a closer look at the math in there), there are several things that should be changed:
1. First, I would uncomment the warnings (l. 415ff and 597ff in src/fe/inf_fe_map.C) such that the user at least has a chance to see that there is an error.
2. At least through the FEInterface-class, one can not govern the parameter interpolated so one is forced to use the default. Hence, I would either change the defaults or the FEInterface interface.
3. If I find time in the next days, I can try to dig into the maths of the inverse_map() with interpolated==true and see why the errors come up (numerically instable or what else).

What do you think about it?
Best,
Hubert

jwpeterson:
> 1 -0.77621  -0.77621  -1.0349   -0.75981  -0.75981  -1.0131

I'd say the error is unacceptably high in every component (I am just looking at the first three floats vs. the last three on each line, not sure exactly what you meant by "3-rd component") for the current default algorithm, while the Newton algorithm looks fine.  Not to mention that the current default algorithm also asserts on one of the points:

> 1 -0.25874  -0.25874  -1.0349   Assertion `-1.-1.e-5 <= v && v < 1.' failed.

I'd definitely change the default algorithm to not use interpolation, as it is just wrong... Do you know whether the bug is restricted to `map` or `inverse_map`, or if both are wrong?

jwpeterson:
> Do you know whether the bug is restricted to map or inverse_map, or if both are wrong?

[...]

PR #1085, opened 2016-09-09
InfFE::inverse_map() - removed interpolated==true

BalticPinguin:
Hi, 
with this commit, issue #1084 should be resolved.
I removed the parameter interpolated (which was to my knowledge not used in any function anyway) and the part of the code where the interpolated scheme was used.
Should one still provide a function with the old interface for compatibility?
Best

### DIFF
--- a/include/fe/inf_fe.h
+++ b/include/fe/inf_fe.h
@@ -378,8 +378,7 @@ public:
   static Point inverse_map (const Elem * elem,
                             const Point & p,
                             const Real tolerance = TOLERANCE,
-                            const bool secure = true,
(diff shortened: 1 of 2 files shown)


jwpeterson:
:+1: I assume you have tested this and it works, our CI is not configured to test with infinite elements enabled so "passing" all the tests doesn't mean much...

I don't think we need to maintain the old, incorrect interface in this case.

BalticPinguin:
I did test it on the map_remap()-function so far. 
So all I can tell is that it runs and the inverse_map() seems to give correct results. So if you consider this testing :-)

PR #1086, opened 2016-09-11
NemesisIO Arbitrary Processor Writing

friedmud:
Adds a capability to NemesisIO to specify the number of processors and the current processor_id.

This allows a mesh to be read in using just one processor... partitioned for `N` processors and then written out to `N` Nemesis files... giving us the ability to "pre-split" meshes for parallel mesh capability.

### DIFF
--- a/include/mesh/nemesis_io.h
+++ b/include/mesh/nemesis_io.h
@@ -67,6 +67,24 @@ public:
   virtual ~Nemesis_IO ();
 
   /**
+   * Used to set the processor_id to use when reading writing.
+   *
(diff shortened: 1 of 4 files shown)


friedmud:
Hold up on this... still verifying that it works properly

jwpeterson:
@friedmud should we still merge this or would you like to drop it now that the CheckpointIO stuff is working for you?

friedmud:
Let's drop this.  This is kind of cool - and maybe useful... but there are still pieces of parallel communication in there (so it doesn't work QUITE right) and the `CheckpointIO` is awesome.

Issue #1087, opened 2016-09-12
Nemesis writing/reading Issue


friedmud:
It seems as if either our Nemesis writing or our Nemesis reading is buggered.  Using NemesisIO to write out a Mesh and then reading it directly back in results in errors.

Here is a minimal libMesh code to demonstrate the issue and I'll attach a mesh to go with it.  For the attached mesh this application works on 2 processors but fails on 4.

@jwpeterson Can I get some help here?

Mesh: [pin_cell_fine.e.zip](https://github.com/libMesh/libmesh/files/467530/pin_cell_fine.e.zip)

```C++
#include "libmesh/libmesh.h"
#include "libmesh/mesh.h"
#include "libmesh/nemesis_io.h"
#include "libmesh/metis_partitioner.h"
#include "libmesh/parallel_mesh.h"

using namespace libMesh;

int main (int argc, char ** argv)
{
  LibMeshInit init (argc, argv);

  if (libMesh::on_command_line("--help"))
  {
    std::cout<<"Example: ./simple-opt --mesh filename.e"
             <<"--mesh             Full name of the mesh file to read in. \n"
             <<std::endl;

    return 0;
  }

  std::string filename = libMesh::command_line_next("--mesh", std::string());
  unsigned int n_procs = libMesh::command_line_next("--n-procs", 1);

  Parallel::Communicator & comm = init.comm();

  Mesh mesh(init.comm());

  mesh.read(filename);

  MetisPartitioner partitioner;

  partitioner.partition(mesh);

  {
    Nemesis_IO nemesis(mesh);
    nemesis.verbose(true);
    nemesis.write(filename);
  }

  {
    DistributedMesh reader_mesh(init.comm());
    Nemesis_IO reader(reader_mesh);
    reader.verbose(true);
    reader.read(filename);
  }

  return 0;
}
```

Here is the error message:

```
Exodus Library Error: [ex_get_node_cmap]
	Error: failed to get variable "n_comm_nids" from file ID 65536
	[-40] Index exceeds dimension bound
Error reading node cmap node and processor ids!
[0] src/mesh/nemesis_io_helper.C, line 393, compiled nodate at notime
application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0
```

jwpeterson:
On Mon, Sep 12, 2016 at 8:33 AM, Derek Gaston <notifications@github.com>
wrote:

> It seems as if either our Nemesis writing or our Nemesis reading is
[...]

PR #1090, opened 2016-09-12
Two fixes while debugging Nemesis reading/writing issues

jwpeterson:
1. Can't call `ex_update()` on a file opened read-only.
2. Fix bug where `MPI_LONG_LONG_INT` was used to wrap `unsigned long long`, causing signed comparison issues.

Refs #1087.

### DIFF
--- a/include/parallel/parallel_implementation.h
+++ b/include/parallel/parallel_implementation.h
@@ -104,7 +104,15 @@ INT_TYPE(unsigned int,MPI_UNSIGNED);
 INT_TYPE(long,MPI_LONG);
 INT_TYPE(long long,MPI_LONG_LONG_INT);
 INT_TYPE(unsigned long,MPI_UNSIGNED_LONG);
-INT_TYPE(unsigned long long,MPI_LONG_LONG_INT);
+#if MPI_VERSION > 1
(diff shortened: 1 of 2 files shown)


include/parallel/parallel_implementation.h
```diff
@@ -104,7 +104,15 @@ INT_TYPE(unsigned int,MPI_UNSIGNED);
...
 INT_TYPE(unsigned long,MPI_UNSIGNED_LONG);
-INT_TYPE(unsigned long long,MPI_LONG_LONG_INT);
+#if MPI_VERSION > 1
+INT_TYPE(unsigned long long,MPI_UNSIGNED_LONG_LONG);
+#else
+// MPI 1.0 did not have an unsigned long long type, so we have to use
+// MPI_UNSIGNED_LONG in this case.  If "unsigned long" and "unsigned
+// long long" are different sizes on your system, we detect this and
+// throw an error rather than communicating values incorrectly.
+INT_TYPE(unsigned long long,MPI_UNSIGNED_LONG);
```
roystgnr: Well, we don't detect this *yet*, but we will soon.  :-)

PR #1091, opened 2016-09-13
pack_range should stop just *after* buffer_size

roystgnr:
If we stop just *before*, then we break if we ever try to pack a
single object that won't fit into the buffer.

In theory this should fix the bug @dschwen pinned down in https://github.com/idaholab/moose/pull/7729

In practice, touching parallel*.h requires rebuilding everything and I haven't even made sure this finishes compiling yet.

### DIFF
--- a/include/parallel/parallel.h
+++ b/include/parallel/parallel.h
@@ -579,13 +579,17 @@ inline void unpack_range (const typename std::vector<buffertype> & buffer,
 /**
  * Encode a range of potentially-variable-size objects to a data
  * array.
+ *
+ * The data will be buffered in vectors with lengths that do not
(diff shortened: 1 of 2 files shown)


dschwen:
Yeah, this looks good. I don't understand why we need to do the packing at all when we are communicating just string buffers. I guess this routine is meant to be more general. A specialization might be a good idea.

roystgnr:
After an embarrassing bugfix-to-the-bugfix, this seems to be working for me.  On my own machine, I also tried cranking down the default approx_buffer_size from 1000000 to 1 just to make sure the "objects bigger than the buffer size" case was *really* tested.  ;-)

PR #1092, opened 2016-09-13
Add allgather string specialization

dschwen:
Specialization to allgather strings into a vector of strings. Will be used for communicating serialized data in MOOSE. This will replace the current use of `allgather_packed_range` and skips the _packing_ aspect as we already serialize the data in binary form on the MOOSE side.

### DIFF
--- a/include/parallel/parallel.h
+++ b/include/parallel/parallel.h
@@ -1098,6 +1098,14 @@ public:
                      T send,
                      std::vector<T> & recv) const;
 
+   /**
+    * Gather overload for string types
+    */
+   template <typename T>
+   inline void gather(const unsigned int root_id,
+                      std::basic_string<T> send,
+                      std::vector<std::basic_string<T> > & recv) const;
+
   /**
    * Take a vector of local variables and expand it on processor root_id
    * to include values from all processors
@@ -1131,6 +1139,13 @@ public:
   template <typename T>
   inline void allgather(T send,
                         std::vector<T> & recv) const;
+
+  /**
+  * AllGather overload for string types
+   */
+  template <typename T>
+  inline void allgather(std::basic_string<T> send,
+                        std::vector<std::basic_string<T> > & recv) const;
 
 
   /**
--- a/include/parallel/parallel.h
+++ b/include/parallel/parallel.h
@@ -1103,7 +1103,7 @@ public:
     */
    template <typename T>
    inline void gather(const unsigned int root_id,
-                      std::basic_string<T> send,
+                      const std::basic_string<T> & send,
                       std::vector<std::basic_string<T> > & recv,
                       const bool identical_buffer_sizes=false) const;
 
@@ -1145,7 +1145,7 @@ public:
   * AllGather overload for string types
    */
   template <typename T>
-  inline void allgather(std::basic_string<T> send,
+  inline void allgather(const std::basic_string<T> & send,
                         std::vector<std::basic_string<T> > & recv,
                         const bool identical_buffer_sizes=false) const;
 
--- a/include/parallel/parallel.h
+++ b/include/parallel/parallel.h
@@ -1104,7 +1104,8 @@ public:
    template <typename T>
    inline void gather(const unsigned int root_id,
                       std::basic_string<T> send,
-                      std::vector<std::basic_string<T> > & recv) const;
+                      std::vector<std::basic_string<T> > & recv,
+                      const bool identical_buffer_sizes=false) const;
 
   /**
    * Take a vector of local variables and expand it on processor root_id
@@ -1145,7 +1146,8 @@ public:
    */
   template <typename T>
   inline void allgather(std::basic_string<T> send,
-                        std::vector<std::basic_string<T> > & recv) const;
+                        std::vector<std::basic_string<T> > & recv,
+                        const bool identical_buffer_sizes=false) const;
 
 
   /**
--- a/include/parallel/parallel_implementation.h
+++ b/include/parallel/parallel_implementation.h
@@ -2950,7 +2950,7 @@ inline void Communicator::gather(const unsigned int root_id,
 
 template <typename T>
 inline void Communicator::gather(const unsigned int root_id,
-                                 std::basic_string<T> sendval,
+                                 const std::basic_string<T> & sendval,
                                  std::vector<std::basic_string<T> > & recv,
                                  const bool identical_buffer_sizes) const
 {
@@ -3084,7 +3084,7 @@ inline void Communicator::allgather(T sendval,
 
 
 template <typename T>
-inline void Communicator::allgather(std::basic_string<T> sendval,
+inline void Communicator::allgather(const std::basic_string<T> & sendval,
                                     std::vector<std::basic_string<T> > & recv,
                                     const bool identical_buffer_sizes) const
 {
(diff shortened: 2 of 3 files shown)


include/parallel/parallel_implementation.h
```diff
@@ -3024,6 +3024,52 @@ inline void Communicator::allgather(T sendval,
...
+    displacements(this->size(), 0);
+
+  // first comm step to determine buffer sizes from all processors
+  const int mysize = static_cast<int>(sendval.size());
+  this->allgather(mysize, sendlengths);
+
+  // Find the total size of the final array and
+  // set up the displacement offsets for each processor
+  unsigned int globalsize = 0;
+  for (unsigned int i = 0; i != this->size(); ++i)
```
permcody: libMesh has different spacing rules. Make this loop look like the others in this file.

include/parallel/parallel_implementation.h
```diff
@@ -3024,6 +3024,52 @@ inline void Communicator::allgather(T sendval,
 
 
 
+template <>
+inline void Communicator::allgather(std::string sendval,
+                                    std::vector<std::string> & recv) const
+{
```
permcody: Spacing for braces is different for libMesh

permcody:
You might also consider a special case for when the communicator is length == 1

permcody:
Oh and it would be cool if you did `gather` while you are at it. I could use that immediately 😄

include/parallel/parallel_implementation.h
dschwen: That particular brace is correct though, right?

permcody: No libmesh has a style like this:
```
<type> foo(params...)
  {
    statements;
  }
```

dschwen: ?!... no

permcody: NVM

dschwen:
Ok, both `gather` and `allgather` are implemented. Any comments @roystgnr? Is that file the right place for these specializations? Did I do something stupid? 

I guess I should add a unit test for these two methods...

roystgnr:
We want a declaration in parallel.h too; in theory parallel_implementation is supposed to only be for definitions and for specializations that you can infer exist by examining parallel.h

roystgnr:
You might as well template around basic_string<T>; it's no additional work.  I'm not expecting anyone to want to allgather unicode but then I wasn't expecting anyone to want to allgather ASCII either.  :-D

roystgnr:
I just realized that nearly all our Parallel:: operations on string and basic_string, including this one, are in theory undefined behavior in C++03, where strings are allowed to use discontiguous implementations like https://en.wikipedia.org/wiki/Rope_(data_structure) under the hood.  In practice I can't find anyone who knows of anyone who knows of a C++ library with a non-contiguous std::string, and C++11 mandates contiguous strings, so I won't stress about it.

roystgnr:
Would it make sense to add the identical_buffer_sizes code path, as we have in allgather<vector>?

roystgnr:
It's weird that even the C version of MPI wants a pointer plus an array of displacements, rather than just an array of pointers.  It would be nice to be able to receive directly into the vector-of-strings rather than slicing up the huge temporary concatenated receive buffer.  But displacements are int and differences between pointers aren't guaranteed to fit in an int, so I don't see any way around it.

roystgnr:
Wait, never mind about the parallel.h declaration.  This fits the ```T -> vector<T>``` pattern so the existing declaration applies.

dschwen:
Yeah, but I'll change `std::string` to `std::basic_string<T>` and add some size magic

roystgnr:
Hmm... adding identical_buffer_sizes would break the ```T -> vector<T>``` pattern.  It could wait until someone really wants that little optimization, if you don't want the hassle.

Do make the switch to ```basic_string<T>```, just because that should be trivial.

dschwen:
aaaand the unit test I just wrote hangs. is there an `Under Development` label here? :-)

dschwen:
@roystgnr, I do need to add a declaration in `parallel.h` since the `std::basic_string<T>` variant is not a specialization, but an _overload_! This would allow me to add a identical size flag as well.

dschwen:
I'm done here. And: 

> current ISO C++ (03) does require &str[0] to cough up a pointer to contiguous string data

http://stackoverflow.com/questions/1986966/does-s0-point-to-contiguous-characters-in-a-stdstring

I know the story of the missing requirement of contiguous storage. But there are other requirements which pretty much demand this.

roystgnr:
Hey, you're right!  data() returns a pointer, so data()[pos] must be an offset from that pointer, so 21.3.4.1 means that the only value which isn't forced to be contiguous is the implicit trailing NULL, which we never let MPI see.

In C++2003 21.3.4.1, at least.  Now I'm tempted to dig up a C++98 standard and see if the same requirement was already in there...  I could swear I recall reading that the C++ standards people had been trying to leave the door open for rope implementations at one point.

include/parallel/parallel.h
```diff
@@ -1098,6 +1098,14 @@ class Communicator
                      T send,
                      std::vector<T> & recv) const;
 
+   /**
+    * Gather overload for string types
+    */
+   template <typename T>
+   inline void gather(const unsigned int root_id,
+                      std::basic_string<T> send,
```
roystgnr: Should we be adding an overload so we can pass send via const reference?  I suppose if we're about to communicate N_proc*O(N) data then a single in-memory copy of O(N) data isn't so bad...

dschwen:
Nevermind. I looked at the 03 standard and it has gems like 

> _References, pointers, and iterators referring to the elements of a basic_string sequence may be invalidated by [...] calling `data()` and `c_str()` member functions_. 

So while some contiguous access is guaranteed, it is only _read_ access. I.e. a string class could store a non-contiguous list of blocks that is rendered into a contiguous buffer when calling `data()` or `c_str()`, and...

> _Nor shall the program treat the returned value as a valid pointer value after any subsequent call to a non-const member function of the class `basic_string` that designates the same object as `this`_ 

(also quoted from the standard). So technically that contiguous view can just vanish again.

include/parallel/parallel.h
dschwen: I tried that, but MPI needs a pointer to a non const object :-/

include/parallel/parallel_implementation.h
```diff
@@ -2966,9 +2967,13 @@ inline void Communicator::gather(const unsigned int root_id,
...
 
-      // first comm step to determine buffer sizes from all processors
       const int mysize = static_cast<int>(sendval.size());
-      this->gather(root_id, mysize, sendlengths);
+
+      if (identical_buffer_sizes)
+        sendlengths.assign(this->size(), mysize);
+      else
+        // first comm step to determine buffer sizes from all processors
+        this->gather(root_id, mysize, sendlengths);
```
roystgnr: We could also use MPI_Gather instead of MPI_Gatherv in the identical_buffer_sizes case, but I suspect just avoiding the latency of the preliminary gather is 95% of the possible optimization here.

include/parallel/parallel.h
roystgnr: We already use const_cast to handle insanity like MPI1 MPI_Send taking a non-const void*.  If there's other took-them-a-while-to-notice-not-Fortran bits in the standard then we can use the same workaround again.  I'd say just check the MPI3 standard first; if things still aren't const-correct there then there might be an actual reason.

roystgnr:
:+1:

roystgnr:
Thanks!

PR #1093, opened 2016-09-13
Don't call ne_get_node_cmap() if there is nothing to get.

jwpeterson:
I thought Nemesis handled this correctly when you passed NULL pointers
for the data vectors, but apparently it throws an error.

Refs #1087.


friedmud:
Awesome!  I'll try it out!

friedmud:
@jwpeterson I can confirm that this _does_ fix it... I was even able to actually write out Nemesis files and read them back into my MOOSE-based simulation and run my code with them!

SO... you think this same error is in some other places?  Should they get fixed now... or is this good to go?

jwpeterson:
I'll fix the other ones today some time...

PR #1097, opened 2016-09-14
Remove deprecated MeshRefinement APIs

### LINKED ISSUE #1095: Fix deprecation messages in adaptivity examples
Currently all of the examples which do adaptivity print two deprecation warnings:
```
*** Warning, This code is deprecated, and likely to be removed in future library versions!
../src/mesh/mesh_refinement.C, line 880, compiled Sep 14 2016 at 14:24:19 ***
../src/mesh/mesh_refinement.C, line 1225, compiled Sep 14 2016 at 14:24:19 ***
```
which stem from this code in `EquationSystems::reinit()`:
```
// Don't override any user refinement settings
MeshRefinement mesh_refine(_mesh);
mesh_refine.face_level_mismatch_limit() = 0; // unlimited
mesh_refine.overrefined_boundary_limit() = -1; // unlimited
mesh_refine.underrefined_boundary_limit() = -1; // unlimited
```
that changes the value `face_level_mismatch_limit` to 0 instead of its default value of 1.  These lines were added in 5b98cbd0 by @roystgnr to avoid *relaxing* restrictions set by the user, but would it not be better to deprecate the current `EquationSystems::reinit()` interface in favor of one which can be passed a `MeshRefinement` object for internal use?  

If that's not a good idea, then I'd like to come up with some other way to avoid all of our adaptivity examples throwing deprecation warnings...

roystgnr:
We were accidentally using this argument internally, creating a bunch of spurious warnings, but it's an API that's been deprecated for years so let's remove it altogether.

This should resolve #1095

### DIFF
--- a/include/mesh/mesh_refinement.h
+++ b/include/mesh/mesh_refinement.h
@@ -685,14 +685,20 @@ private:
   /**
    * Take user-specified coarsening flags and augment them
    * so that level-one dependency is satisfied.
+   *
+   * This function used to take an argument, \p maintain_level_one -
+   * new code should use face_level_mismatch_limit() instead.
    */
-  bool make_coarsening_compatible (const bool);
+  bool make_coarsening_compatible ();
 
   /**
    * Take user-specified refinement flags and augment them
    * so that level-one dependency is satisfied.
+   *
+   * This function used to take an argument, \p maintain_level_one -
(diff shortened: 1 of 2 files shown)


jwpeterson:
:+1:

PR #1098, opened 2016-09-14
assertion for the MPI1-with-32-bit-long case

roystgnr:
Not sure if it's safe to put a static_assert here too, so I decided
not to.  In theory anyone in this use case can turn all their id types
down to 32-bit and still have a working libMesh, and we shouldn't
risk breaking that.

If you're using MPI 1.0, Windows, and only 32-bit integers, I don't
want to risk making your life any more difficult than it already is.

This makes good on the "throw an error" promise in #1090.

Does anyone actually have an MPI 1 module around somewhere to test this against?  I can only find 2.0 and 3.0 on my own systems.


jwpeterson:
:+1:

PR #1099, opened 2016-09-14
Avoid deprecated code in adjoints_ex2.

jwpeterson:
This initially didn't work on the no-C++11 tests because of a slight
type issue, but it should be working now.


jwpeterson:
Also trying out the new valgrind add-on test with this PR... I don't really expect it to pass since it's the first time we've tried it, so I'm not going to let it hold up the merge.

jwpeterson:
Test is green and `grep "definitely lost" ~/Desktop/results_42654_Valgrind/05_Check` reports 0 bytes for all tests, so I think we're doing good on the examples.  Thanks again @brianmoose for adding the optional Valgrind target.

PR #1101, opened 2016-09-15
Miscellaneous fixes/updates in Nemesis_IO_Helper.

jwpeterson:
* Only call ne_get_eb_info_global() if there is something to get.
* Only call ne_get_elem_cmap() if there is something to get.
* Avoid multiple accidental insertion possibilities, throw errors
  instead.  We were not hitting any of these in my test code, but it's
  good to get rid of them nonetheless.
* Remove unnecessary ternary operators.
  These calls were already guarded by if statements, so there was no
  way they could have been called with empty vectors.
* Delete commented out code.
* Fix spacing in if-statements.
* Declare loop iterators outside of declaration, use simpler iterator
  syntax.

I will also make sure that the commit in #1086 rebases cleanly on this, since it may have touched some of the same lines...

### DIFF
--- a/src/mesh/nemesis_io_helper.C
+++ b/src/mesh/nemesis_io_helper.C
@@ -117,9 +117,9 @@ void Nemesis_IO_Helper::get_ss_param_global()
 
       nemesis_err_flag =
         Nemesis::ne_get_ss_param_global(ex_id,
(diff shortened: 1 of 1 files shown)


jwpeterson:
Rebased the splitter branch commit on this one at 7baf07c9673d2 for @friedmud.

friedmud:
Awesome!

I'love try this out in a little while.

Thanks!

PR #1102, opened 2016-09-15
Write named entities to Nemesis files

### LINKED ISSUE #1094: Nemesis Named Mesh Entities
Named mesh entities (blocks, sidesets, etc.) don't seem to be getting written to Nemesis files...

jwpeterson:
Whenever we added named entity writing support to the `ExodusII_IO_Helper`, we never added it to the `Nemesis_IO_Helper`, which has to specialize those functions.

Fixes #1094.

### DIFF
--- a/src/mesh/nemesis_io.C
+++ b/src/mesh/nemesis_io.C
@@ -949,6 +949,17 @@ void Nemesis_IO::read (const std::string & base_filename)
 
       libMesh::out << "[" << this->processor_id() << "] "
                    << "nemhelper->num_elem_all_sidesets = " << nemhelper->num_elem_all_sidesets << std::endl;
+
+      if (nemhelper->num_side_sets > 0)
+        {
+          libMesh::out << "Sideset names are: ";
+          std::map<int, std::string>::iterator
+            it = nemhelper->id_to_ss_names.begin(),
+            end = nemhelper->id_to_ss_names.end();
+          for (; it != end; ++it)
+            libMesh::out << "(" << it->first << "," << it->second << ") ";
+          libMesh::out << std::endl;
+        }
     }
 
 #ifdef DEBUG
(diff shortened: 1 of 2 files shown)


jwpeterson:
This also conflicted with the patch in #1086, so I rebased that again here: c9d6f9f.

friedmud:
Coolness! Thanks!

friedmud:
How in the heck do I checkout c9d6f9f2 ?  I fetched libMesh/libmesh... but can't find c9d6f9f2 anywhere... doesn't appear to be on any branch...

friedmud:
Ok - got it - I don't know why... but GitHub was showing the commit in `libMesh/libmesh`... but actually it's on your fork.  Weird.

friedmud:
This is kind of working (in simple test codes and checking the resulting Nemesis files with Paraview): but MOOSE is failing when trying to use these Nemesis files:

```
*** ERROR ***
The object 'left' contains the following boundary ids that do no exist on the mesh: 7

[0] /Users/gastdr/projects/moose/framework/src/base/BoundaryRestrictable.C, line 130, compiled nodate at notime
application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0
```

Do you think it's a problem with MOOSE... or is something off with _reading_ named entities in Nemesis?

jwpeterson:
Hmm... maybe a non-local node id made it into the BoundaryInfo object and then was written out to file?  This happens with the same test mesh when you try to run a problem on it?

friedmud:
Yep

Should be easily reproducible for you... just try to use that mesh with
2d_diffusion_test on 4 processors.

The thing is: I think MooseMesh does an allreduce on the block/boundary
names / IDs specifically to deal with the face that MOOSE objects may be
referencing mesh entities that are not on the local processor.

But something may be messed up there.

PR #1103, opened 2016-09-19
Checkpoint splitter

friedmud:
Forget Nemesis reading.  It's too damn complicated. 

We can do this ourselves.  DistributedMesh is actually really awesome at reinitializing itself... it barely needs anything.  Just the local elements and the nodes they're connected to... along with BC info.  Done.

That said... there is still a small issue with this PR that must be resolved before it can be merged.  It currently throws an `assert()` in devel/dbg mode if you leave mesh partitioning on when you prepare the newly read-in mesh for use.

You can see the way I'm needing to do the reading over here:

https://github.com/friedmud/moose/commit/6e08eef9d7c18fa9c6675fdde46251a575489d8b

@jwpeterson @roystgnr could I get just the tiniest bit of help tracking that one down?

To easily create meshes that are split using this new capability use this branch of my `simple_libmesh_app`:

https://github.com/friedmud/simple_libmesh_app/tree/splitter

refs https://github.com/idaholab/moose/issues/7752 , https://github.com/idaholab/moose/issues/7744 , https://github.com/idaholab/moose/pull/7745 , #1087 , #1086

### DIFF
--- a/include/mesh/checkpoint_io.h
+++ b/include/mesh/checkpoint_io.h
@@ -99,6 +99,11 @@ public:
   bool   binary() const { return _binary; }
   bool & binary()       { return _binary; }
 
+  /**
+   * Get/Set the flag indicating if we should read/write binary.
+   */
+  bool   parallel() const { return _parallel; }
+  bool & parallel()       { return _parallel; }
 
   /**
    * Get/Set the version string.
@@ -109,6 +114,17 @@ public:
 private:
   //---------------------------------------------------------------------------
   // Write Implementation
+
+  /**
+   * Build up the elem list
+   */
+  void build_elem_list();
+
+  /**
+   * Build up the node list
+   */
+  void build_node_list();
+
   /**
    * Write subdomain name information - NEW in 0.9.2 format
    */
@@ -181,7 +197,11 @@ private:
   unsigned int n_active_levels_on_processor(const MeshBase & mesh) const;
 
   bool _binary;
+  bool _parallel;
   std::string _version;
+  unsigned int _mesh_dimension;
+  std::set<largest_id_type> _local_elements;
+  std::set<largest_id_type> _nodes_connected_to_local_elements;
 };
 
 
--- a/include/mesh/checkpoint_io.h
+++ b/include/mesh/checkpoint_io.h
@@ -111,6 +111,18 @@ public:
   const std::string & version () const { return _version; }
   std::string &       version ()       { return _version; }
 
+  /**
+   * Get/Set the processor_id to use
+   */
+  const processor_id_type & current_processor_id() const { return _my_processor_id; }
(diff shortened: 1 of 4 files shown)


include/mesh/checkpoint_io.h
```diff
@@ -99,16 +99,44 @@ class CheckpointIO : public MeshInput<MeshBase>,
...
+   */
+  const processor_id_type & current_processor_id() const { return _my_processor_id; }
+  processor_id_type & current_processor_id() { return _my_processor_id; }
+
+  /**
+   * Get/Set the n_processors to use
+   */
+  const processor_id_type & current_n_processors() const { return _my_n_processors; }
+  processor_id_type & current_n_processors() { return _my_n_processors; }
+
```
roystgnr: These aren't redundant with the ParallelObject APIs because during reading they come from the file rather than the CheckpointIO object?

If I'm understanding that correctly, could you make it clearer in the comments?

friedmud: Actually - those are for `m->n` splitting.  Like run with `m` MPI processes... but create output files for `n` MPI processes.

To do that a code can loop over and set `current_n_processors()` and `current_processor_id()` and call `write()`... and the file that comes out will be for _that_ situation.

I'll update the comment

friedmud: See here: https://github.com/friedmud/simple_libmesh_app/blob/9062a3b0fb1d9ae7e75958e6270c15c0edb6b3f8/src/main.C#L77

permcody:
Hmm - that same test timed out yesterday as well. I don't recall it ever timing out before...

friedmud:
@permcody It looks like it passed this time.  It definitely doesn't have anything to do with this PR... so I don't know what's up.

src/mesh/parallel_mesh.C
```diff
@@ -1349,14 +1349,14 @@ void DistributedMesh::delete_remote_elements()
 {
 #ifdef DEBUG
   // Make sure our neighbor links are all fine
-  MeshTools::libmesh_assert_valid_neighbors(*this);
+  //MeshTools::libmesh_assert_valid_neighbors(*this);
```
roystgnr: I definitely don't want to delete them entirely.  "This assert would pass here" is very useful information to leave in the code even if we don't compile it.

Personally, I thought DEBUG *was* the extra_paranoid_slow flag.  We've already got -O0 and GLIBCXX_DEBUG* active in dbg mode too; it's basically as unoptimized as it gets.

friedmud: Sorry guys: I didn't mean to leave these changes in here.  Those just slipped through.  I'll revert those and force-push this.

BTW: As you can see with my newest commit... I'm still working on this a bit.  One of my problems still won't run using this capability.  Still hunting down why.

friedmud:
Closing this in favor of #1106

PR #1105, opened 2016-09-26
Fix unit test

jwpeterson:
This unit test was never actually added by the branch merged in 4813964.  Once I re-ran bootstrap and enabled it, it would not even compile.  Those issues are both now fixed and the test does pass, but it makes the unit tests take a lot longer to run, about 36 seconds on my system in debug mode, so we may want to revisit the size of the test. @roystgnr


roystgnr:
:+1:

I should have known that when the unit tests worked the first time that meant they weren't even being compiled.

Let's crank down the cube size from 15x15x15 to 10x10x10 and see if that's a bit faster?  The only other obvious tweak is to go down to 2 levels of ghosting but I think the test coverage for the former might be better.

jwpeterson:
> Let's crank down the cube size from 15x15x15 to 10x10x10 and see if that's a bit faster?

Will do.

jwpeterson:
Is 5x5x5 OK?  That gets it down to about 5.5 seconds in debug mode:
```
15x15x15 36s
10x10x10 13.5s
8x8x8    9s
5x5x5    5.5s
```

roystgnr:
Hmm... yeah, 5^d is good enough.  Smaller grids might actually even be a better unit test, since on smaller processor counts that'll force the ghost layer to cross multiple processor boundaries rather than just testing one-on-one interactions.  Plus that way I won't feel too guilty if I later go back and force a repartitioning or something which makes the test even longer.  :+1: when Civet is happy, thanks!

PR #1106, opened 2016-09-27
Parallel CheckpointIO and Splitter

friedmud:
This PR adds the ability to do parallel, decomposed `CheckpointIO`... and a new app called `splitter` that can generate these files by partitioning a serial mesh into `n` partitions.

As a bonus: you can run splitter with MPI... and any number of MPI processes can output any number of partitions.  For instance:

`mpiexec -n 24 ./splitter-opt --mesh=mesh.e --n-procs='9 24 48 112'`

Will use 24 processors to read in `mesh.e` and split it into four different partitionings suitable for running with `DistributedMesh` on 9, 24, 48 and 112 processors.

### DIFF
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -2442,11 +2442,18 @@ std::string Elem::get_info () const
 
   oss << "  Elem Information"                                      << '\n'
       << "   id()=";
-
   if (this->valid_id())
     oss << this->id();
   else
     oss << "invalid";
+
+#ifdef LIBMESH_ENABLE_UNIQUE_ID
+  oss << ", unique_id()=";
+  if (this->valid_unique_id())
+    oss << this->unique_id();
+  else
+    oss << "invalid";
+#endif
 
   oss << ", processor_id()=" << this->processor_id()               << '\n';
 
--- a/src/mesh/checkpoint_io.C
+++ b/src/mesh/checkpoint_io.C
@@ -52,7 +52,9 @@ CheckpointIO::CheckpointIO (MeshBase & mesh, const bool binary_in) :
   ParallelObject      (mesh),
   _binary             (binary_in),
   _parallel           (false),
-  _version            ("checkpoint-1.1")
+  _version            ("checkpoint-1.1"),
+  _my_processor_id    (processor_id()),
+  _my_n_processors    (n_processors())
 {
 }
 
@@ -61,8 +63,10 @@ CheckpointIO::CheckpointIO (MeshBase & mesh, const bool binary_in) :
 CheckpointIO::CheckpointIO (const MeshBase & mesh, const bool binary_in) :
   MeshOutput<MeshBase>(mesh,/* is_parallel_format = */ true),
   ParallelObject      (mesh),
-  _binary (binary_in),
-  _parallel (false)
+  _binary             (binary_in),
+  _parallel           (false),
+  _my_processor_id    (processor_id()),
+  _my_n_processors    (n_processors())
 {
 }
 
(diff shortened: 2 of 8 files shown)


src/geom/elem.C
```diff
@@ -2442,12 +2442,19 @@ std::string Elem::get_info () const
...
     oss << "invalid";
 
+#ifdef LIBMESH_ENABLE_UNIQUE_ID
+  oss << ", unique_id()=";
+  if (this->valid_unique_id())
+    oss << this->unique_id();
+  else
+    oss << "invalid";
+#endif
+
```
roystgnr: This ought to be in its own commit

friedmud: Is it really worth it?  It's just an informational message... hardly seems like it's worth the trouble in rebasing and splitting that commit...

roystgnr:
This still just outputs local elements and point neighbors to each file; it doesn't yet query the mesh ghosting functors?

friedmud:
@roystgnr Right.  Until the ability exists to do point-neighbors with the ghosting functors I can't use that capability yet.

As soon as it's possible I'll update CheckpointIO to use the ghosting functors.

roystgnr:
Needs unit tests.  Otherwise looks good to me.  :+1:

friedmud:
@roystgnr I actually have a branch that _does_ use ghosting functors... so I know how to do it and can switch to it quickly when it's possible.

friedmud:
@roystgnr How do you feel-about the partitioning "reset": https://github.com/libMesh/libmesh/pull/1106/files#diff-811e61f3bbb6a2947e99750e61cdca05R82

Is there a better way to do that?

roystgnr:
Sorry I was AFK while you were having that discussion earlier.

On the on hand, it looks like a bug to me - we ought to be able to repartition from any N to any M without clearing a thing in between.

On the other hand, even after we fix the bug, I think an app like splitter will want to do this "reset", because some of our partitioners are initial-state-dependent and it would be a good idea to keep that initial state as consistent as possible.

friedmud:
Ok - sounds good to me.

friedmud:
@jwpeterson could I get some help on getting this merged?  I'm swamped right now... and this only needs a bit more to go in... plus it would be good to have another set of eyes on it... and this capability needs to get added to MOOSE soon as well so it would be good to get another MOOSE dev involved.

All this needs is a little more work in getting the ghosting functor stuff right (it should use the ghosting functor from the Mesh) a test or two and some conflicts with automake generated files resolved...

jwpeterson:
> @roystgnr I actually have a branch that does use ghosting functors... so I know how to do it and can switch to it quickly when it's possible.

Should I be starting from the branch you mention above instead?

friedmud:
@jwpeterson good point!

Here is that code: https://github.com/friedmud/libmesh/tree/checkpoint_splitter_functor

You won't want to build directly on it... the code here in this PR is newer and has other fixes, etc.  But you can see how I was using the `GhostingFunctor` to pull in the correct elements.

It turned out not to work perfectly for me though... because it wasn't pulling in point neighbors.  But now, @roystgnr added a GhostingFunctor that will pull in the point neighbors so it's possible to switch back to using the GhostingFunctor again.

In addition - I was just building a GhostingFunctor myself... when really it should be pulling it from the Mesh.

At any rate: you can see how I was trying to use it there.

friedmud:
Thanks for helping with this!  It's been on my todo list for a month... and I don't see that changing anytime soon...

jwpeterson:
Replaced by #1160.  That PR implements @roystgnr's suggestion to split the elem.C content into a separate commit.

src/mesh/checkpoint_io.C
```diff
@@ -129,17 +147,59 @@ void CheckpointIO::write (const std::string & name)
...
-      // pause all processes until the writing ends -- this will
-      // protect for the pathological case where a write is
-      // followed immediately by a read.  The write must be
-      // guaranteed to complete first.
+      // close it up
       io.close();
     }
 
-  this->comm().barrier();
+  // this->comm().barrier();
```
lindsayad, line 154: Was removal of the barrier here the right thing to do? I agree that the comment was not well placed since I don't think the `Xdr` `io close()` has any idea about the parallel process writes, but otherwise the comment seems to do a good job motivating the `barrier()` call

roystgnr: IMHO it was.  "Read immediately after writing" is a super common use case in test codes, but those get their own `barrier()` calls in between, and in real codes there's usually work to do after writing and we don't want to postpone any of the quickest writers to wait on the slowest writer.

lindsayad: So you're saying that people doing write and then read should be doing the barrier themselves. If so, I feel like we should consider documenting that somewhere

PR #1107, opened 2016-09-28
Point neighbor coupling

roystgnr:
This won't be ready to merge until I add more tests, by which I mean "at least one test", but there's a slim chance it might be in working order, so I figured I'd at least upload it in case @friedmud wants to play.


friedmud:
Cool - thanks!

I'll try it out.

Still don't know yet how we're going to expose this in MOOSE....

roystgnr:
The number of possible combinations of stuff you can do now does seem daunting even before you try to figure out how to encapsulate them as input file options.

On the other hand, the input file syntax could be much simpler than the C++ API for basic behaviors, and could lead to optimized code under the hood.  For instance, if you just have a "n_neighbor_levels" and "n_point_neighbor_levels" for the user to specify, MOOSE could be smart enough to not create the GhostingFunctor for a n_levels=0 case, to remove the neighbors-only functor in cases where a point-neighbors functor is being requested with the same or more levels, etc.

permcody:
@roystgnr - This is what I was thinking. In MOOSE we have several stages of setup on the Mesh. First we create it (depending on the type, that might mean reading or generating it). We then call `prepare_for_use()` the first time. Next MOOSE allows you to run any number of modifiers optionally allowing you to force additional `prepare_for_use()` calls in between modifiers. Generally we've found that we don't need this but there are a few cases where it is necessary. There's a final `Action` that fires to make sure the Mesh is ready to go before moving on to setup the rest of the systems. `prepare_for_use()` is called here if it hasn't been or there have been additionally modifications made to the mesh.

Now that I've dumped all this out, perhaps we can think about where this extra ghosting method can be triggered. It seems like we need another "task" to handle the ghosting after all modifiers are complete. As you said, most users will just use pre-canned operators like n_neighbor_levels or n_neighbor_point_levels (the former being *exactly* what I need for my current work).

It would be easy to slot in another "task" somewhere in that workflow to do this, but I have a few questions. 

1. Does it make sense to ever use multiple ghosting modifiers in a single simulation? Does that theoretically work or does one "overwrite" the work of the other. Maybe I have a contact problem that automatically sets up a specific ghosting algorithm but I combine that later with rDG and need to turn on extra neighbor ghosting too. If this can/might work, then our existing MeshModifier system might work perfectly since you can already set up several modifiers. We'll just need to "sort" these special modifier (ghoster) types to the back of the modifier list which is easy to accomplish. 

2. Do we need to be careful about calling prepare_for_use() with these ghost functors? As you can see we sometimes call that method more than once.

@friedmud - Am I on the right track here?

roystgnr:
1. Multiple ghosting functors are fine, and in some cases are necessary.  In fact, by default we have multiple ghosting functors!  The default algebraic ghosting functor includes neighbors, but the default coupling functor doesn't, to match the old libMesh behavior: by default we expect that users want to be able to evaluate all dofs on neighbor elements (for flux jump or patch error indicators) but don't want to couple those dofs across boundaries (unless they're doing DG).

Suppose you have one ghosting functor that, called on element e, says to couple elements a, b, and c, with CouplingMatrix M1, and another functor that says to couple elements c, and d, with CouplingMatrix M2.  Then a and b will be coupled according to M1, d will be coupled according to M2, and c will be coupled according to max(M1,M2).  (note that a NULL CouplingMatrix is, as it was in the DofMap, shorthand for "an all ones matrix, couple everything to everything else")

The only reason not to use multiple ghosting functors is for efficiency, as I said above; if you've got one functor which returns a subset of what the rest of your functors return, then it's pure wasted CPU cycles to not remove it.

2. We don't have to be careful about calling prepare_for_use() more than once; you should be able to repartition and refine to your heart's content, even.  But we have to be *VERY CAREFUL* that, when prepare_for_use() is first called on a DistributedMesh, the ghost functors associated with that mesh are *already* configured to return a (not necessarily strict) superset of what they'll be returning in the long term.  Take a look at 03ada0b8 for an example: if we read a DistributedMesh first, and then set up PeriodicBoundaries later, then the mesh has *already* stopped ghosting its periodic neighbors by the time we let it know that they exist, and by this time it's too late to fix that.  We have to create the System so that the DofMap's default functors will exist, then create the PeriodicBoundary objects so those functors will know what to do, before we can read or generate the Mesh.

2b.  I say "should be able to" above, but DistributedMesh currently has a broken corner case with adaptive refinement.  It's hard to trigger, and I'm working on the fix, but that'll take a while, so if you need to do distributed AMR/C soon let me know and I can put in an inefficient workaround instead.

friedmud:
@permcody We don't call the ghosting functors at all.  You just attach the one you want to use to the Mesh/EquationSystem and it gets called automatically at the right time.  So, in MOOSE, once we decide which one is needed... we need to attach that one as soon as possible.

@roystgnr is right in that we're going to need the ghosting functors to be conservative with `DistributedMesh` early on.  If you wipe out local elements early... it's hard to get them back.  In MOOSE we do already have the ability to forcibly re-"ghost" (copy) elements along boundaries to bring them back from other processors if they get removed (it's the `ghosted_boundaries` option in the `[Mesh]` block).  So we may be able to use that to pull elements back (or create them for the first time if you read in a Distributed CheckpointIO that didn't include all of the elements you need ghosted) to the local processors so that the ghosting functors can do their job (which will ultimately lead to some elements getting re-deleted while others are tagged as ghosted and stay permanently).

@permcody : About multiple ghosting functors: @roystgnr already handled this.  Look here:

https://github.com/libMesh/libmesh/blob/master/include/mesh/mesh_base.h#L706

You can add any number of ghosting functors... and if an element is required by _any_ of them it will be ghosted.

PR #1108, opened 2016-10-03
Better System::point_*() methods

roystgnr:
This fixes a bug in the new DefaultCoupling unit tests on my laptop, and makes System::point_value() and kin more useful.  I'm not sure why I didn't manage to trigger the bug sooner.


jwpeterson:
👍

PR #1109, opened 2016-10-03
New DirichletBoundary APIs

roystgnr:
We have to leave the old constructor API options, both for backwards compatibility and because they're useful in benchmark problems where an exact solution functor exists, but this PR adds new constructor API options which ought to be much easier to work with for most use cases, as well as more documentation of how DirichletBoundary variable indexing works with either option set.


examples/adjoints/adjoints_ex5/heatsystem.C
```diff
@@ -62,8 +62,10 @@ void HeatSystem::init_data ()
 
   ZeroFunction<Number> zero;
 
+  // Most DirichletBoundary users should use this style of
+  // constructor, which takes a "locally indexed" functor
   this->get_dof_map().add_dirichlet_boundary
-    (DirichletBoundary (all_bdys, T_only, &zero));
+    (DirichletBoundary (zero, all_bdys, T_only));
```
jwpeterson: So, from this diff it seems that the new constructors take the same arguments, just in a different order and as references rather than pointers?  This is probably going to forever confuse me/have me looking at the docs to see which order is which... would it not be possible to have the constructor take an `enum LOCALLY_INDEXED` or something so that it's clear from reading the code what indexing you are using?

roystgnr: No, it's even more confusing than that.  :-D  The old APIs can also accept references instead of pointers.

Your enum idea sounds like a big improvement over either this confusing disambiguation or my named constructor alternative.

roystgnr:
Going to resubmit this after I try it with @jwpeterson's enum suggestion instead.

jwpeterson:
Looks much better, looks like you just missed updating one or two of the examples.

PR #1111, opened 2016-10-05
Use smart pointer in test.

### LINKED ISSUE #1110: Fix memory leak in numeric_vector_test.h
Only a unit test, so not a big deal, but still there's no reason to have [this](https://github.com/libMesh/libmesh/blob/master/tests/numerics/numeric_vector_test.h#L55).

jwpeterson:
People use the unit tests as exapmles for getting started, we don't
want to propagate the manual memory allocation/deallocation idiom
unnecessarily.

Closes #1110.


roystgnr:
:+1:

PR #1112, opened 2016-10-05
Prefer UniquePtr to manual delete

roystgnr:
As with the unit test @jwpeterson caught, we should probably be
teaching new users to use smart pointers rather than manual
deallocation.

Let's not merge this until I can test it manually - reduced_basis_ex7 isn't even compiled unless we build with --enable-complex.


roystgnr:
The reduced_basis_ex7 change *didn't* work at first crack, but does now.

PR #1113, opened 2016-10-05
Fix recent unit tests with --enable-complex

roystgnr:
Apparently it's time for me to give up on getting BuildBot SingleBranchScheduler to work properly with github, and just reenable our Periodic scheduler for all the non-default configuration tests.

### DIFF
--- a/tests/numerics/petsc_vector_test.C
+++ b/tests/numerics/petsc_vector_test.C
@@ -55,13 +55,13 @@ public:
 
     // Check the values through the interface
     for (unsigned int i=0; i<local_size; i++)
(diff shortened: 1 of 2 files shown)


jwpeterson:
Can we do a complex-enabled build sans PETSc?  It would be pretty easy for us to set up a complex target for moosebuild if we didn't also have to worry about having a complex-enabled PETSc in our binary package that no one would actually use.

roystgnr:
Hmm...  We couldn't originally.  I'll bet Eigen supports complex Numbers, though.  I'll try it out.

roystgnr:
My "make check" hasn't finished yet, but it's churning away successfully so far; apparently Eigen supports complex Numbers.

jwpeterson:
With no changes on the libmesh side?  Nice...

PR #1114, opened 2016-10-05
scatter, specialized gather !HAVE_MPI cases

roystgnr:
We added these methods recently but didn't add their fallback
implementations for !LIBMESH_HAVE_MPI configurations; this gave us
undefined reference errors at link time for any code using these
methods.


permcody:
Noted for next time.

dschwen:
Yikes. I saw the need for the no MPI case at one point but completely forgot about it! :-(

PR #1116, opened 2016-10-05
Fix systems_of_equations_ex6 with --enable-complex

roystgnr:


jwpeterson:
:+1: This also passed the new "enable complex" target on moosebuild!  Thanks @brianmoose for adding that.

PR #1117, opened 2016-10-06
Options fixes

roystgnr:
I can't even turn BuildBot off for a month without bit rot setting in.

### DIFF
--- a/tests/base/dof_object_test.h
+++ b/tests/base/dof_object_test.h
@@ -153,15 +153,15 @@ public:
 
     // Check that we can manually index dofs of variables based on the first dof in a variable group
     // Using: id = base + var_in_vg*ncomp + comp
-    CPPUNIT_ASSERT_EQUAL(aobject.vg_dof_base(0, 0) + 1*1 + 0, aobject.dof_number(0, 1, 0));
+    CPPUNIT_ASSERT_EQUAL((dof_id_type)(aobject.vg_dof_base(0, 0) + 1*1 + 0), aobject.dof_number(0, 1, 0));
 
     // Another Check that we can manually index dofs of variables based on the first dof in a variable group
     // Using: id = base + var_in_vg*ncomp + comp
-    CPPUNIT_ASSERT_EQUAL(aobject.vg_dof_base(0, 1) + 2*3 + 2, aobject.dof_number(0, 4, 2));
+    CPPUNIT_ASSERT_EQUAL((dof_id_type)(aobject.vg_dof_base(0, 1) + 2*3 + 2), aobject.dof_number(0, 4, 2));
 
     // One More Check that we can manually index dofs of variables based on the first dof in a variable group
     // Using: id = base + var_in_vg*ncomp + comp
-    CPPUNIT_ASSERT_EQUAL(aobject.vg_dof_base(1, 1) + 0*3 + 0, aobject.dof_number(1, 2, 0));
+    CPPUNIT_ASSERT_EQUAL((dof_id_type)(aobject.vg_dof_base(1, 1) + 0*3 + 0), aobject.dof_number(1, 2, 0));
   }
 
   void testJensEftangBug()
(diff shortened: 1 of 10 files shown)


permcody:
Hmm, Is that something you want to do? Perhaps we should expand some of this testing on our side? Actually, hopefully (and I mean hopefully) CIVET will be open sourced and you'll be able to run additional clients at your site using our recipes which will at least keep the configuration and testing down to a minimum. In any case I think we can do something sensible to consolidate testing systems.

tests/base/dof_object_test.h
```diff
@@ -153,15 +153,15 @@ class DofObjectTest {
 
     // Check that we can manually index dofs of variables based on the first dof in a variable group
     // Using: id = base + var_in_vg*ncomp + comp
-    CPPUNIT_ASSERT_EQUAL(aobject.vg_dof_base(0, 0) + 1*1 + 0, aobject.dof_number(0, 1, 0));
+    CPPUNIT_ASSERT_EQUAL((dof_id_type)(aobject.vg_dof_base(0, 0) + 1*1 + 0), aobject.dof_number(0, 1, 0));
```
jwpeterson: Let's use `static_cast` here and below?

jwpeterson:
The build failure could be caused because we are using `--disable-petsc`, but MPI is still being detected... and not linked against correctly?  This is another uncommon configuration so I could see it regressing, but this target passed on a different machine already once...

```
undefined reference to `MPI_Init_thread'
./.libs/libmesh_opt.so: undefined reference to `MPI_Bcast'
./.libs/libmesh_opt.so: undefined reference to `MPI_Ssend'
./.libs/libmesh_opt.so: undefined reference to `MPI_Op_create'
./.libs/libmesh_opt.so: undefined reference to `MPI_Pack_size'
./.libs/libmesh_opt.so: undefined reference to `MPI_Probe'
./.libs/libmesh_opt.so: undefined reference to `MPI_Send'
...
```

jwpeterson:
So for the complex target, as @brianmoose pointed out, configure picks up `ccache g++` from the environment as the compiler rather than the `mpicxx` which should also be in the environment.
```
C++ compiler type.................. : gcc5
C++ compiler....................... : ccache g++
C compiler......................... : ccache gcc
```
This means that it doesn't have the right libs for linking against MPI...
```
libmesh_optional_LIBS............ : -lz -Wl,-rpath,/opt/moose/tbb/lib -L/opt/moose/tbb/lib -ltbb -ltbbmalloc
```

roystgnr:
How's it managing to get through mpi.m4 without either setting a proper MPI_LIBS or unsetting enablempi?

permcody:
So I can't answer all your questions but I know that I've gotten away with using the clang++ compiler directly to compile the PETSc/libMesh/MOOSE stack for a long time. Yeah, that's probably not a good idea, but it generally works.

jwpeterson:
> How's it managing to get through mpi.m4 without either setting a proper MPI_LIBS or unsetting enablempi?

That's a good question.  Our configure line has some contradictory options in it:
```
./configure --with-methods=opt --prefix=/home/moosetest/civet/client_root_0/libmesh/installed --enable-silent-rules --enable-unique-id --disable-warnings --enable-openmp --disable-maintainer-mode --with-metis=PETSc --enable-complex --disable-petsc
```
i.e. we've both enabled PETSc (by setting `--with-metis=PETSc`) and explicitly disabled it.  This happened because we are just appending flags to the "default" configure flags.  I wonder if this prevents `ACX_MPI` from being called somehow.

jwpeterson:
We fixed the module which was causing `$CXX` to be set wrong, so I invalidated the `--enable-complex` target and set it to re-run on the same client.  That should make the PR pass, but I'm not sure if I want to debug the petsc-enabled-but-disabled-and-non-mpi-compiler-in-CXX issue in our configure scripts right at this moment...

jwpeterson:
Also note that this has only been fixed on one build box, so any other PRs submitted in the next couple of days which run the `--enable-complex` target on a different build box (i.e. almost all of them) will still fail at the moment.  Need to wait until the MOOSE package manager guy gets back in town to figure out the best long term solution.

PR #1119, opened 2016-10-11
Fixed bug in InfFE::inverse_map() concerning v being out of range.

BalticPinguin:
Hi,
I found two further bugs in the inverse_map() function:
First, the distance of the intersection should be taken to the origin of infinite elements, not to the origin of the coordinate system.

Moreover, the function is called from the PointLocator where the physical point might be far away from the element. In this case I had the problem that the InfFE::eval() function failed due to v<-1-1e-5.
To circumvent this, I added the respective check and just leave the loop, if v<-1.

Finally,  the parameter v can be assigned analytically in my point of view, at least in the CARTESIAN mapping.
Probably you can have a look at this part to see whether I am right or if there is some configuration currently available, where this does not hold. However, one can use this knowledge for a better guess anyway.

### DIFF
--- a/src/fe/inf_fe_map.C
+++ b/src/fe/inf_fe_map.C
@@ -351,11 +351,15 @@ Point InfFE<Dim,T_radial,T_map>::inverse_map (const Elem * inf_elem,
 
   // the distance from the intersection on the
   // base to the origin
-  const Real a_dist = intersection.norm();
+  const Real a_dist = Point(o-intersection).norm();
 
   // element coordinate in radial direction
   // here our first guess is 0.
   Real v = 0.;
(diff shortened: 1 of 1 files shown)


jwpeterson:
These changes seem reasonable to me.  Did they change the results of misc_ex1 at all?  We don't test this with the CI, so just curious...

BalticPinguin:
no, since in misc_ex1 the origin is in 0 and the mapping is only used to map points that are known to be in the mesh. Therefore I didn't find these issues before...

jwpeterson:
Thanks for the info.  I merged your original commit after fixing a few indentation things.

PR #1122, opened 2016-10-12
Add an interface for passing the transpose nullspace to PETSc

snschune:


snschune:
I think this is all I need to do, but I gotta test it with Moose.

jwpeterson:
:+1: although I'd prefer the underscored name `transpose_nullspace` and changing the near nullspace stuff `near_nullspace`, but the latter is already part of the public interface of `NonlinearSolver`...

snschune:
>  although I'd prefer the underscored name transpose_nullspace and changing the near nullspace stuff to near_nullspace, but the latter is already part of the public interface of NonlinearSolver...

@jwpeterson  do you want this change?

jwpeterson:
> @jwpeterson do you want this change?

I'd like to at least name the new one with an underscore.  If you make that change, I'll merge this, then it can be tested in your other MOOSE PR.

snschune:
@jwpeterson does this look ok to you?

jwpeterson:
Thanks!

snschune:
Thanks.

anilkunwar:
Hi All,
Is the libmesh compilation affected by this change as in this?
https://groups.google.com/forum/#!topic/moose-users/lV-fOw1i06I

Yours Sincerely,
Anil Kunwar

jwpeterson:
@anilkunwar Yes, please don't start a second thread here, the one you already started on the moose users mailing list is sufficient.

PR #1123, opened 2016-10-14
Parallel tweaks

### LINKED ISSUE #1118: Add const_casts to new "scatter" implementations in parallel_implementation.h
We need a few const casts around the raw buffers to support MPI-1 standards which apparently is still in use on the version of OpenMPI that you receive with Ubuntu.

roystgnr:
The first commit here should resolve #1118, but I don't have an MPI-1 or even broken-MPI-2 system handy to test with, so I'll hold off on merging until @permcody reports that this works or that he can't test it easily either.

While I was doing that I noticed that many of *our* APIs were being too restrictive about requiring non-const containers as input arguments; now I feel bad about mocking the MPI-1 designers' mistakes.

While I was fixing *that* I noticed that some of our never-used std::set overloads were buggy or unusable, so I fixed those and added test cases.

### DIFF
--- a/include/parallel/parallel.h
+++ b/include/parallel/parallel.h
@@ -839,7 +839,7 @@ public:
    */
   template <typename T>
   void send (const unsigned int dest_processor_id,
-             T & buf,
+             const T & buf,
              const MessageTag & tag=no_tag) const;
 
   /**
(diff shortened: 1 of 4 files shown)


permcody:
Thanks for taking care of this. It's been on my radar too but I've been swamped. I don't have one a good test box for this right now either. I'll see if I can find a guinea pig from our mailing list.

roystgnr:
Okay.  I'll merge as-is, since there's definitely *some* fixes here even if I may have missed others.  And I'll start building an openmpi-1.1.5 module tree, so we can get some MPI-1 support in buildbot to test with.

PR #1125, opened 2016-10-18
More fixes from optional configuration tests

roystgnr:


roystgnr:
Pretty sure that timeout was a fluke - MeshCommunication::query_ghosting_functors shouldn't even be hit except on a DistributedMesh.  I'll wait for INL folks to comment before merging, though.

PR #1126, opened 2016-10-20
FE tests and fixes

roystgnr:
Hundreds more unit tests (well, 3 more unit tests in a hundred odd configurations each), plus assorted bug fixes for corner cases nobody currently uses, plus bug fixes for the case @manavbhatia caught a few days ago, plus bug fixes for equally embarrassing lesser-used cases.

There's at least one bug still outstanding here: XYZ is giving me errors on Tets.  There's also a few cases (e.g. p > 3 HERMITE elements) that I could swear worked at one point but that need more than trivial fixes to reenable.


jwpeterson:
Hurray for the new complex-valued CI testing!
```
from ../../tests/fe/fe_monomial_test.C:2:
../../tests/fe/fe_test.h: In member function ‘void FETest<order, family, elem_type>::testU()’:
../../tests/fe/fe_test.h:153:13: error: cannot convert ‘libMesh::Number {aka std::complex<double>}’ to ‘double’ for argument ‘1’ to ‘void CppUnit::assertDoubleEquals(double, double, double, CppUnit::SourceLine, const string&)’
```

jwpeterson:
Are the serendipity (`QUAD8`, `HEX20`, `PRISM15`) elements deliberately excluded from the tests?

tests/fe/fe_test.h
```diff
@@ -0,0 +1,232 @@
...
+class FETest : public CppUnit::TestCase {
+
+private:
+  unsigned int _dim, _nx, _ny, _nz;
+  Elem *_elem;
+  std::vector<dof_id_type> _dof_indices;
+  FEBase * _fe;
+  Mesh* _mesh;
+  System* _sys;
+  EquationSystems* _es;
```
jwpeterson: Three different kinds of spacing for pointer declarations used in these 7 lines.

src/base/dof_map.C
```diff
@@ -620,17 +620,20 @@ void DofMap::reinit(MeshBase & mesh)
...
+                   (base_fe_type.order.get_order()),
+                 FEInterface::max_order(base_fe_type,type),
+                 "ERROR: Finite element "
+                  << Utility::enum_to_string(base_fe_type.family)
+                  << " on geometric element "
+                  << Utility::enum_to_string(type)
+                  << "\nonly supports FEInterface::max_order = "
+                  << FEInterface::max_order(base_fe_type,type)
+                  << ", not fe_type.order = "
+                  << base_fe_type.order);
```
roystgnr: Only when we #define macros, not when we call them, AFAIK

manavbhatia:
Super!  Thanks Roy!

-Manav

PR #1127, opened 2016-10-21
XYZ tests and fix

roystgnr:
It turns out that FE caching via static variables is broken even when we're not using threads; it just breaks in very contrived, rare, hard-to-replicate circumstances.

The cache in XYZ doesn't seem to be saving us very much, so let's get rid of it entirely.

The cache in CLOUGH is actually pretty heavy-duty, so for now all I'm doing is adding dbg-mode tests so that people who suspect cache corruption can at least hope to catch it.


roystgnr:
Okay, that's enough of a swath of new tests.  I'll reorder and combine a couple commits to get rid of make-check-failing intermediate states and then I'll merge this.

PR #1129, opened 2016-10-24
switch for tetgen-options added

BalticPinguin:
Hi,
I took Roy's suggestion now for implementing the additional switch.
I hope, this meets your expectations and I didn't make any mistake in this part.
Best,
Hubert


include/mesh/mesh_tetgen_interface.h
```diff
@@ -155,6 +159,12 @@ class TetGenMeshInterface
   std::vector<unsigned> _sequential_to_libmesh_node_map;
 
   /**
+   * Parameter controling the behaviour of tetgen.
+   * By default quiet.
+   */
+  std::string _switches = "Q";
```
jwpeterson: I believe this (inline initialization of non-const class members) is a C++11 only feature, although it was not caught by our "no C++11" test for some reason.

jwpeterson: Oh, probably because we don't test with `--disable-strict-lgpl`, so this new code is not even being tested.

jwpeterson: I'll add a quick fix for this, otherwise it looks fine.

BalticPinguin: Ok; I changed this line now.

src/mesh/mesh_tetgen_interface.C
```diff
@@ -39,8 +39,22 @@ TetGenMeshInterface::TetGenMeshInterface (UnstructuredMesh & mesh) :
   _mesh(mesh),
   _serializer(_mesh)
 {
+  _switches="Q";
```
jwpeterson: Thanks for fixing this, but it should just be initialized in the initialization list (two lines up) like the other members.

jwpeterson:
:+1: thanks!

PR #1132, opened 2016-10-24
fix petsc_vector operator= for vecs with nonlocal state

rwcarlsen:


roystgnr:
local_size() is supposed to equal "index_stop - index_start" according to the numeric_vector and petsc_vector headers.  If this is true, then why would changing from _local_size to "first-last" change things?  If this isn't true, then don't we have a deeper bug somewhere?

A bug in the header comments, at least?  I just noticed incidentally that we're using last_local_index() as the horribly-chosen name of a function which we claim refers to "the last vector element actually stored on this processor", but PETSc docs say they're returning "one more than the last local element".  The comment in numeric_vector.h seems correct; we just need to update the one in petsc_vector.h?

roystgnr:
If this fixes a bug, then we ought to be adding a unit test for that bug (and a PR description of it!) alongside the fix.  Otherwise the fix is likely to be reverted as soon a s

jwpeterson:
> If this is true, then why would changing from _local_size to "first-last" change things? If this isn't true, then don't we have a deeper bug somewhere?

We found that, for ghosted vectors, `_local_size` could be something like 80 even when `last_local_index - first_local_index` was ~60.  @rwcarlsen can post the actual numbers... and yes there may be a bigger bug somewhere, but I figured it was due to `operator=(std::vector)` not being a commonly-used function...

roystgnr:
Knepley says VecGetLocalSize "will have the local size without ghosts", in which case our _local_size should be correct.
http://lists.mcs.anl.gov/pipermail/petsc-users/2009-January/000016.html

Ah, but wait.  Our _local_size is not the same as our local_size().  The latter is really local, the former is local+ghosted.  So the bug makes sense and this is the correct fix.

We really ought to make that naming less insane.  last_local_index() is a public API with some immunity, but IMHO _local_size should be renamed _semilocal_size ASAP.

jwpeterson:
> _local_size should be renamed _semilocal_size ASAP.

I agree it should be renamed, but I'd rather not propagate the "semi" naming scheme into NumericVectors, considering that most people, primarily myself, still don't understand exactly what that means in the context of `Elem::is_semilocal` elements, the `semiverify` macro, etc...

jwpeterson:
How about `_local_plus_ghosted_size`.

roystgnr:
Wouldn't increased use of the definition "semilocal == local plus ghosted" promote, rather than reduce, understanding of that jargon?

I sympathize with the problem, but it seems that the two consistent alternatives here are "better document the term semilocal" or "find every single comment and API which includes the term semilocal and replace it with e.g. local_plus_ghosted".  I'd prefer the first alternative, but if it's unacceptable then the sooner we bite the bullet on the second, the better.  Even restricting ourselves to the context of vectors, if talking about semilocal values is to be discouraged, then DofMap::all_semilocal_indices() has been a problem for 6 years.

roystgnr:
Oh, but regardless, let's merge this now?  We can fix comments and variable names later; I shouldn't be holding up a bug fix to debate wording.

jwpeterson:
I doubt.  This change will only affect code that calls `PetscVector::operator=(std::vector)` which isn't very common...

PR #1134, opened 2016-10-24
Treat empty _dof_coupling as full _dof_coupling

roystgnr:
Hopefully this will be sufficient to fix the regression in support that Yak triggered in https://github.com/idaholab/moose/pull/7912

### DIFF
--- a/include/base/default_coupling.h
+++ b/include/base/default_coupling.h
@@ -55,8 +55,7 @@ public:
  {}
 
   // Change coupling matrix after construction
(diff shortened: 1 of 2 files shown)


roystgnr:
I just force pushed a fix to the fix.

This turns out to have absolutely nothing to do with the Yak regression, but it does restore (libmesh_deprecated(), at least) support for an old libMesh behavior that we shouldn't break without warning.

IMHO this should be backported to 1.1.0 once we're sure it's working - I don't know if anyone depends on it but I'd be happier never finding out.

PR #1135, opened 2016-10-24
Use UniquePtr to manage helper class lifetimes in ExodusII_IO and Nemesis_IO.

jwpeterson:
We no longer have to explicitly call `delete` in the destructors to avoid a memory leak.

### DIFF
--- a/include/mesh/nemesis_io.h
+++ b/include/mesh/nemesis_io.h
@@ -125,10 +125,18 @@ public:
 
 private:
 #if defined(LIBMESH_HAVE_EXODUS_API) && defined(LIBMESH_HAVE_NEMESIS_API)
-  Nemesis_IO_Helper * nemhelper;
-  int _timestep;
+  UniquePtr<Nemesis_IO_Helper> nemhelper;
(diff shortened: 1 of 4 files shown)


include/mesh/nemesis_io.h
```diff
@@ -125,10 +125,18 @@ class Nemesis_IO : public MeshInput<MeshBase>,
...
-  int _timestep;
+  UniquePtr<Nemesis_IO_Helper> nemhelper;
 #endif
 
+  /**
+   * Keeps track of the current timestep index being written. Used
+   * when calling write_nodal_data() and other functions.
+   */
+  int _timestep;
+
```
roystgnr: Why move this out of the ifdef?

roystgnr:
:+1:

include/mesh/nemesis_io.h
jwpeterson: It seems inconsistent to have it inside, as none of the other members of the class are.  It is also consistent with the Exodus helper this way.

PR #1136, opened 2016-10-25
Optionally skip find_neighbors in MeshBase::read

dknez:


roystgnr:
Oh come on, give us more than 2 after-work hours to review an API change to a core libMesh method!

For starters: What's the use case here?  When do we need a mesh but not need elem->neighbor(s) to return a valid result?

dknez:
Eek, I assumed it was uncontroversial since it only an optional argument... happy to revert it if you want.

The use case is that I sometimes need info about a mesh (e.g. element order info) but don't need to do a solve, hence the overhead of find_neighbors is unnecessary in that case.

roystgnr:
Okay, that makes sense.  And this looks like the right way to enable that.  Fine by me unless anyone else objects.

dknez:
OK. I was too hasty with the merge though, sorry about that! Bad etiquette on my part :)

Issue #1130, opened 2016-10-24
Add initial guess for eigenproblems


juan42:
As discussed on the libmesh-users mailinglist, there is a demand for setting the initial guess in the EigenSystem / EigenSolver interface. This has to be done via a SLEPc call to EPSSetInitialSpace.

My questions for the enhancement are:

- Should the option be available in EigenSystem or EigenSolver interface? I don't use the EigenSystem class at all, because i have no use for _just_ solving an eigenproblem.

- Should the option, if set, assume that the last solution is always the next initial space for further solves? (Could be irritating, but beneficial if explicitly documented)

jwpeterson:
> Should the option be available in EigenSystem or EigenSolver interface? I don't use the EigenSystem class at all, because i have no use for just solving an eigenproblem.

I think it would make more sense on the `EigenSystem`.  The `System` base class already provides the following API for setting initial conditions that should probably be used for this as well:
```
  /**
   * Register a user function to use in initializing the system.
   */
  void attach_init_function (void fptr(EquationSystems & es,
                                       const std::string & name));

  /**
   * Register a user class to use to initialize the system.
   * Note this is exclusive with the \p attach_init_function.
   */
  void attach_init_object (Initialization & init);
```

If you are currently just using `EigenSolver` objects directly, it shouldn't be too hard to use it from within an `EigenSystem` object, which also gives you convenient access to things like the `Mesh`, variables, vector projection APIs, etc.



> Should the option, if set, assume that the last solution is always the next initial space for further solves? (Could be irritating, but beneficial if explicitly documented)

We do this for time-dependent problems (the solution from the previous timestep is the initial guess for the current timestep) so there is certainly precedent for it.  I'd say go ahead unless you think there is genuine potential for confusion/bad results.

roystgnr:
If we were designing from scratch, I'd say it was a mistake to use system.solution for eigenfunctions in the first place.  Eigensolves of nonlinear systems are a very useful thing, and you don't want to overwrite the current state of the system just to examine an eigenfunction.

It's too late to change that decision now, but I wonder if using the initial solution as an eigensolve initial guess would make things worse.  0 is a pretty decent initial guess, and the nonlinear system's solution may not be, so this would be a potential regression for some users.

I don't there's one correct answer here, but my vote would be to put the option into EigenSolver.

PR #1137, opened 2016-10-25
Add option to add an initial guess for the SLEPc eigensolver

juan42:
Pull request for issue #1130 

I decided to just implement the EigenSolver interface for now and just used eigenproblems_ex2 for the demonstration to keep eigenproblems_ex1 as simple as possible.

EDIT: As i'm thinking about it afterwards, a new example with two solves and checking for the number of iterations would make sene.

### DIFF
--- a/src/solvers/slepc_eigen_solver.C
+++ b/src/solvers/slepc_eigen_solver.C
@@ -752,8 +752,7 @@ void SlepcEigenSolver<T>::set_initial_space(NumericVector<T> & initial_space_in)
   
   PetscErrorCode ierr = 0;
   Vec initial_vector = (cast_ptr<PetscVector<T> *>(&initial_space_in))->vec();
-  Vec * initial_space = &initial_vector;
-  ierr = EPSSetInitialSpace(_eps, 1, initial_space);
+  ierr = EPSSetInitialSpace(_eps, 1, &initial_vector);
   LIBMESH_CHKERR(ierr);
 #endif
 }
--- a/src/solvers/slepc_eigen_solver.C
+++ b/src/solvers/slepc_eigen_solver.C
@@ -745,6 +745,9 @@ void SlepcEigenSolver<T>::attach_deflation_space(NumericVector<T> & deflation_ve
 template <typename T>
 void SlepcEigenSolver<T>::set_initial_space(NumericVector<T> & initial_space_in)
 {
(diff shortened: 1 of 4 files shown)


jwpeterson:
`EPSSetInitialVector` has been in SLEPc since v3.1, aka 2010 time frame:

```
commit 5f7b509317e2c0b9c87c93529abfe48dd64f4899
Author: Jose E. Roman <jroman@dsic.upv.es>
Date:   Thu Apr 1 13:46:42 2010 +0000

    Replaced EPSSetInitialVector by the new function EPSSetInitialSpace; still more work required for left initial vectors
    
    Subversion-commit: r1933
```

So we'll need to guard this function call, i.e.
```
#if SLEPC_VERSION_LESS_THAN(3,1,0)
  libmesh_error_msg("SLEPc 3.1 is required to call EigenSolver::set_initial_space() ");
#else
// current PR code.
#endif
```

src/solvers/slepc_eigen_solver.C
```diff
@@ -743,6 +743,18 @@ void SlepcEigenSolver<T>::attach_deflation_space(NumericVector<T> & deflation_ve
 }
 
 template <typename T>
+void SlepcEigenSolver<T>::set_initial_space(NumericVector<T> & initial_space_in)
+{
+  this->init();
+  
+  PetscErrorCode ierr = 0;
+  Vec initial_vector = (cast_ptr<PetscVector<T> *>(&initial_space_in))->vec();
```
jwpeterson: (Just a reminder to self, this is not something for you to necessarily fix.)

I see that this idiom is used in several places in the `SlepcEigenSolver` code, but in optimized mode, `cast_ptr` simply does a `static_cast`, which could cause this line of code to segfault somewhat mysteriously if a non-PETSc `NumericVector` is passed in.  I'd like to change these to be done in two lines with a dynamic cast, checking the result.

jwpeterson:
Thanks for making the changes.  I think I will squash these commits together before merging.  Then @BalticPinguin can perhaps rebase his related changes on master and try his PR again...

BalticPinguin:
So I removed the initiial_guess related part. No rebase needed from my side.

jwpeterson:
Merged to master in b64cb69.  I also fixed some whitespace issues that were introduced by the patches during merge...

PR #1138, opened 2016-10-25
position_of_spectrum value "target" added

BalticPinguin:
Hi,
In contrast to my expectations I did have some time to implement the interface for setting the initial space by some guess.
Also, I added options to set the position of spectrum. (Sorry for this being in the same PR but I worked on it at the same time, making it hard to keep the changes separated.)

However, I hope the changes are ok this way.

### DIFF
--- a/src/solvers/eigen_solver.C
+++ b/src/solvers/eigen_solver.C
@@ -60,12 +60,29 @@ void EigenSolver<T>::set_solver_configuration(SolverConfiguration & solver_confi
   _solver_configuration = &solver_configuration;
 }
 
+template <typename T>
+void EigenSolver<T>::set_position_of_spectrum (Real pos)
+{
+   if (pos>=0)
+     _position_of_spectrum=TARGET_MAGNITUDE;
+  else
+     _position_of_spectrum= TARGET_REAL;
+  _target_val=pos;
+}
+
+template <typename T>
+void EigenSolver<T>::set_position_of_spectrum (Real pos, PositionOfSpectrum target)
+{
+    _position_of_spectrum=target;
+    _target_val=pos;
+}
+
+
 
 //------------------------------------------------------------------
 // Explicit instantiations
 template class EigenSolver<Number>;
 
 } // namespace libMesh
-
 
 #endif // LIBMESH_HAVE_SLEPC
--- a/src/solvers/slepc_eigen_solver.C
+++ b/src/solvers/slepc_eigen_solver.C
@@ -633,15 +633,29 @@ void SlepcEigenSolver<T>:: set_slepc_position_of_spectrum()
       ierr = EPSSetWhichEigenpairs (_eps, EPS_LARGEST_MAGNITUDE);  LIBMESH_CHKERR(ierr); return;
     case SMALLEST_MAGNITUDE:
       ierr = EPSSetWhichEigenpairs (_eps, EPS_SMALLEST_MAGNITUDE); LIBMESH_CHKERR(ierr); return;
+    case TARGET_MAGNITUDE:
+      {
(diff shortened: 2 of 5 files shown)


juan42:
Now we have two implementations for the initial guess haha :+1:

BalticPinguin:
oh! I didn't look at the current progress though :-)

BalticPinguin:
Nice, at least there is progress :-)
So I would suggest to take your implementation (they are not very different anyway and I shorten this PR to keep only the position_of_spectrum() part...

include/solvers/slepc_eigen_solver.h
```diff
@@ -267,6 +267,13 @@ class SlepcEigenSolver : public EigenSolver<T>
    */
   EPS _eps;
 
+  /**
+   * Store a set of \p _initial_space_n initial
+   * guesses in \p _initial_space_vect.
+   */
+  int _initial_space_n;
```
jwpeterson: This chunk should probably no longer be part of the diff since you are going to use @juan42's interface for that instead?

src/solvers/slepc_eigen_solver.C
```diff
@@ -633,15 +633,29 @@ void SlepcEigenSolver<T>:: set_slepc_position_of_spectrum()
       ierr = EPSSetWhichEigenpairs (_eps, EPS_LARGEST_MAGNITUDE);  LIBMESH_CHKERR(ierr); return;
     case SMALLEST_MAGNITUDE:
       ierr = EPSSetWhichEigenpairs (_eps, EPS_SMALLEST_MAGNITUDE); LIBMESH_CHKERR(ierr); return;
+    case TARGET_MAGNITUDE:
+      {
+      ierr = EPSSetTarget(_eps, this->_target_val); LIBMESH_CHKERR(ierr);
+      ierr = EPSSetWhichEigenpairs (_eps, EPS_TARGET_IMAGINARY); LIBMESH_CHKERR(ierr); return;
```
jwpeterson: Is this right?  You are in `TARGET_MAGNITUDE` case, but you're passing `EPS_TARGET_IMAGINARY` to SLEPc...

src/solvers/eigen_solver.C
```diff
@@ -53,19 +53,74 @@ EigenSolver<T>::build(const Parallel::Communicator & comm,
...
+     }
+  else
+    {
+       _position_of_spectrum= TARGET_REAL;
+        _target_val=pos;
+    }
+}
+
+template <typename T>
+void EigenSolver<T>::set_position_of_spectrum (Real pos, std::string ImagReal)
```
jwpeterson: IMO this function should just take a `PositionOfSpectrum` enum rather than a string.  `PositionOfSpectrum` is already part of the public interface of `EigenSolver`, and it would save you all this code which converts from string to enum...

src/solvers/eigen_solver.C
```diff
@@ -53,19 +53,74 @@ EigenSolver<T>::build(const Parallel::Communicator & comm,
...
+     {
+       _position_of_spectrum=TARGET_IMAGINARY;
+       _target_val=pos;
+     }
+   else if( ImagReal == "r")
+     {
+       _position_of_spectrum=TARGET_REAL;
+       _target_val=pos;
+     }
+   else if( ImagReal == "r")
```
jwpeterson: This `else if` block will never be hit because the condition is identical to the one before it...

src/solvers/slepc_eigen_solver.C
BalticPinguin: That is true!

include/solvers/slepc_eigen_solver.h
BalticPinguin: yes, I forgot it.

jwpeterson:
Thanks for making those fixes!

PR #1139, opened 2016-10-26
Leave _default_evaluating->_dof_coupling == NULL

roystgnr:
This should restore the old default libMesh behavior and should
hopefully fix https://github.com/idaholab/moose/pull/7912

### DIFF
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -1546,9 +1546,29 @@ void DofMap::merge_ghost_functor_outputs
                         }
                       const_cast<CouplingMatrix&>(*existing_it->second) &= *metg_it->second;
                     }
+                  else
+                    {
+                      // Any existing_it matrix merged with a full
+                      // matrix (symbolized as nullptr) gives another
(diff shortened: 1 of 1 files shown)


jwpeterson:
:+1: I will merge this tomorrow and update the MOOSE PR with the new hash to make sure it works.

roystgnr:
Yeah, I see another bug.  Shouldn't be hard to fix.

I am loving this Yak test, and kicking myself for not putting anything equally advanced into the libMesh unit tests yet.

roystgnr:
Yak works for me now.  I still need to work up some unit tests for the CouplingMatrix merging (there's an even more complicated use case that Yak isn't exercising; I'd say 50% chance it works 50% it's a ticking time bomb) but that'll be a later PR; we should merge these fixes to master and 1.1.0 ASAP.

jwpeterson:
> Yak works for me now.

Cool.  I will temporarily push these commits into upstream so that the libmesh update PR in MOOSE  can test against them as well.  Feel free to merge whenever is convenient.

PR #1140, opened 2016-10-26
Fix binary GMV file writing

jwpeterson:
Paraview can read the ASCII and binary GMV files libmesh writes, the binary file output was not working correctly before.  Also removed some C-style string handling and manual `new`/`delete` memory management code.


roystgnr:
:+1:

PR #1141, opened 2016-10-26
Use UniquePtr to manage lifetimes of more stuff.

jwpeterson:
This prevents us from needing to call delete manually in destructors.


roystgnr:
:+1:

PR #1145, opened 2016-10-27
Deprecate ImplicitSystem::get_linear_solver().

jwpeterson:
Add libmesh_error_msg() that will eventually be turned on after the
function has been deprecated for a while.

### DIFF
--- a/include/systems/implicit_system.h
+++ b/include/systems/implicit_system.h
@@ -109,9 +109,15 @@ public:
    * Returns a pointer to a linear solver appropriate for use in
    * adjoint and/or sensitivity solves
    *
-   * In \p ImplicitSystem this just gives us a default LinearSolver;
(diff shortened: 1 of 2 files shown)


jwpeterson:
@roystgnr I think this works with all of our examples and with MOOSE so it means no one has been calling it directly.  Actually I think it's unlikely that anyone has ever used it, since it would give unexpected results if it was called in the way that e.g. `misc_ex12` calls it to set parameters:
```
system.get_linear_solver()->set_solver_type(SPARSELU);
```

My plan is to deprecate the base class functionality for now, turn it into a `libmesh_error_msg()` later.

roystgnr:
Yeah, this was never intended to be used except via derived classes.  Not sure why I didn't have a libmesh_error() in the base class implementation in the first place.

PR #1146, opened 2016-10-28
Added functions to set a spectral transformation to SLEPc

BalticPinguin:
Hi,
since the problem I study is a bit unstable due to close-lying eigenvalues inserted a function to set the ST type in SLEPc.
Maybe it is interesting for other users as well?
Is the way I wrote it the 'correct' one or would you suggest an other interface?
Best,
Hubert


dknez:
This looks good, but my concern with this type of thing is that it may be a bit of a "fool's errand" to try to wrap PETSc and SLEPc solver options in detail. There are so many options in those libraries, and our wrappers in libMesh inevitably get out of date or don't support the full range of options.

My preferred approach these days is to just to write PETSc or SLEPc code where needed (or use the PETSc/SLEPc command line options) when more detailed configuration options are needed. Just my 2 cents.

BalticPinguin:
That is a good point actually.
So basically the question would be whether on could write a more general function in which the user can pass 'free format' options to slepc so the user is responsible to keep this up to date?
The other variant I would see is to keep the interface minimalistic, so everybody who wants to do some more advanced things need to write it on his own...
Or would you suggest some other variant?

dknez:
> So basically the question would be whether on could write a more general function in which the user can pass 'free format' options to slepc so the user is responsible to keep this up to date?

Yeah, this is my preferred approach, personally. I do this approach via the SolverConfiguration class (SlepcEigenSolver is already aware of this class). See systems_of_equations_ex6, for example, for a PETSc usage of it. I find this approach more flexible, since you can get access to the KSP, PC, EPS objects yourself and do whatever you want with them.

BalticPinguin:
Ah, ok; now I start understanding this class...
I thought it would contain just a set of 'standard' options that can be set.
Ok than this PR is probably really not needed.
Thank you for your help and advises.

jwpeterson:
> Ok than this PR is probably really not needed.

Please feel free to close this if you decide it's not needed, I'm fine with it either way.

PR #1148, opened 2016-10-28
Add extra geometric element cases for Szabab FE

jwpeterson:
Thanks @manavbhatia for pointing this out.


roystgnr:
That first commit clearly looks like the right bug fix, but I'm not sure how we weren't already covering it in the testGradU unit test.

The second commit doesn't make sense to me - we can't support second order C0 FE types on first order elements in general.  The FE either needs to be discontinous (so dofs live on the element) or C1 (so more than one dof can live on each vertex).

jwpeterson:
> The second commit doesn't make sense to me

Sorry, I didn't look to see what `INSTANTIATE_FETEST` actually did...

Issue #1149, opened 2016-10-31
Mesh refinement with Infinite elements fails


BalticPinguin:
Hi,
It is me again with some issue with infinite elements: I wanted to try to refine my mesh that has infinite elements but failed with the message 
'ERROR: Bad ElemType = INFPRISM6 for FIRST order approximation!'
which is thrown when trying to reinit the sides; more precisely, the backtrace is 
`
0: libMesh::print_trace(std::ostream&) 

1: libMesh::MacroFunctions::report_error(char const*, int, char const*, char const*)
2: /home/tm162/bin/libmesh/my_libmesh/lib/libmesh_dbg.so.0(+0x12c9b50) [0x7ffff5ab7b50]
3: libMesh::FE<3u, (libMesh::FEFamily)0>::n_dofs(libMesh::ElemType, libMesh::Order)
4: libMesh::FE<3u, (libMesh::FEFamily)0>::n_shape_functions(libMesh::ElemType, libMesh::Order)
5: void libMesh::FEMap::init_reference_to_physical_map<3u>(std::__debug::vector<libMesh::Point, std::allocator<libMesh::Point> > const&, libMesh::Elem const*)
6: libMesh::FE<3u, (libMesh::FEFamily)0>::reinit(libMesh::Elem const*, std::__debug::vector<libMesh::Point, std::allocator<libMesh::Point> > const*, std::__debug::vector<double, std::allocator<double> > const*)
7: libMesh::JumpErrorEstimator::reinit_sides()
8: libMesh::JumpErrorEstimator::estimate_error(libMesh::System const&, libMesh::ErrorVector&, libMesh::NumericVector<std::complex<double> > const*, bool)
`
As I see it, JumpErrorEstimator::reinit_sides () calles the fe->reinit() directly which of course does not know the infinite element types.

My question now is mainly: Is mesh refinement supposed to work with infinite elements or do I need to expect there more errors with that? 
Best,
Hubert

roystgnr:
On Mon, 31 Oct 2016, Hubert Weißmann wrote:

> My question now is mainly: Is mesh refinement supposed to work with infinite elements or do I need to expect there more errors with that?

That's two questions; the answer to both is "yes".

It's been years and years since we had a major developer doing work
with infinite elements, which means that the only thing between the
InfFE code and bit-rot has been our test coverage, of which we don't
have enough.  PRs adding unit tests would be greatly appreciated.

[...]

PR #1150, opened 2016-11-02
InfFE evaluation at non-quadrature points

roystgnr:
This PR was *intended* to add mesh refinement capability for infinite elements as requested in #1149.  I think I got about 50% of the way there by at least fixing the bugs with evaluation of InfFE solutions at arbitrary points.


roystgnr:
I haven't found time to put in any more regression testing, so I'm going to rely on https://github.com/libMesh/libmesh/issues/1149#issuecomment-258435302 and merge this now; it's definitely providing some improvement and probably not breaking anything.

PR #1154, opened 2016-11-08
Converting OSSpinLock to pthread_mutex

andrsd:
OSSpinLock has been deprecated on macOS X Sierra. People suggest to
covert it to mutex, which is what this does.

### DIFF
--- a/include/parallel/threads_pthread.h
+++ b/include/parallel/threads_pthread.h
@@ -37,7 +37,11 @@
 #include <vector>
 
 #ifdef __APPLE__
-#include <libkern/OSAtomic.h>
+#  ifdef __MAC_10_12
+#    include <os/lock.h>
+#else
+#    include <libkern/OSAtomic.h>
+#  endif
 #endif
 
 // Thread-Local-Storage macros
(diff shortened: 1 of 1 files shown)


andrsd:
This should not break any backwards compatibility.

Also, the Apple SDK recommends to replace it with `os_unfair_lock_t`, but I had no luck with. It can be that I was doing something wrong.  I kept the `#ifdef __APPLE__` part, but obviously `pthread_mutex` is available everywhere, so one might be able to merge the parts into a single implementation if that is preferred.

Lastly [this page](https://idea.popcount.org/2012-09-12-reinventing-spinlocks/) reinvents spinlocks and claims to provide a drop in functions for `pthread_spin`. I tried this, but I ran into weird behavior (sometimes the code would run, sometimes it would not). So I abandoned that route...

friedmud:
`os_unfair_lock` looks like the right thing here.  We really want this to be as fast and light as possible.

I'm not 100% against `pthread_mutex` but it can be a little heavy in places where we would want to use a quick spin lock...

There is a reason that we had this separate branch for `OSSpinLock` in the first place...

friedmud:
Here are some timings showing that `os_unfair_lock` is ~12 times faster than `pthread_mutex`: https://gist.github.com/steipete/36350a8a60693d440954b95ea6cbbafc

jwpeterson:
Marking as "do not merge" for the moment, @andrsd is still looking into things...

friedmud:
Here's some C code that seems to use the new stuff: https://git.enlightenment.org/core/efl.git/commit/?id=519207c7eb228b6fd3dc4e5626b1c4cd374991d4

It's definitely hard to find good examples of `os_unfair_lock`

andrsd:
I think me and @jwpeterson found the reason why my original code did not work, I did not call `OS_UNFAIR_LOCK_INIT` - found in the `os/lock.h`. So, let me try this and let's see of that will fix the problem.

I agree that `os_unfair_lock` would be a better option here.

andrsd:
Updated the PR to use `os_unfair_lock`. Both libmesh tests and moose tests finished successfully on my machine.

jwpeterson:
> Updated the PR to use os_unfair_lock. Both libmesh tests and moose tests finished successfully on my machine.

OK, I will try your branch now on El Capitan...

jwpeterson:
This works for me on El Capitan.  By the way, the test 
```
#ifdef __MAC_10_12
...
#endif
```
seems to be the correct way to test for OSX >= 10.12.  On my El Capitan system, the following test code prints:
```
__MAC_10_11=101100
__MAC_10_10=101000
__MAC_10_9=1090
```

```
#include <iostream>

// Test whether certain #defines are set on OSX.
int main()
{
#ifdef __MAC_10_12
  std::cout << "__MAC_10_12=" << __MAC_10_12 << std::endl;
#endif
#ifdef __MAC_10_11
  std::cout << "__MAC_10_11=" << __MAC_10_11 << std::endl;
#endif
#ifdef __MAC_10_10
  std::cout << "__MAC_10_10=" << __MAC_10_10 << std::endl;
#endif
#ifdef __MAC_10_9
  std::cout << "__MAC_10_9=" << __MAC_10_9 << std::endl;
#endif
  return 0;
}
```

so all the old defines are still set in the new versions.  In case this threading stuff ever changes again, we can just add another `#ifdef` for the newer OS version earlier in the list.

jwpeterson:
👍

PR #1155, opened 2016-11-09
Allow virtual methods for edge_fe_reinit() and elem_fe_reinit()

balborian:
This way I can derive a class from fem_context() for specific cases when I don't need to recalculate the shape functions for each element. For instance, a linear QUAD4 in a uniform grid with all elements having the same shape. This saves a lot of time.

Edit: I don't understand why it fails on the check stage. I didn't make other changes. I can run the check on my machine though.

### DIFF
--- a/include/systems/fem_context.h
+++ b/include/systems/fem_context.h
@@ -679,7 +679,7 @@ public:
   /**
    * Reinitializes interior FE objects on the current geometric element
    */
-  void elem_fe_reinit(const std::vector<Point> * const pts = libmesh_nullptr);
+  virtual void elem_fe_reinit(const std::vector<Point> * const pts = libmesh_nullptr);
 
   /**
    * Reinitializes side FE objects on the current geometric element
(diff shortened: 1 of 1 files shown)


roystgnr:
I'm pretty baffled by that "make check" failure too.  It almost looks as if someone made an ABI-incompatible libstdc++ upgrade under the Civet libMesh build without a "make clean" afterward, and it definitely doesn't look like anything that could be caused by your patch.

I'm a little disappointed that your override helps much.  Years ago I put some caching code in FE::reinit() that was a big optimization for Hermite elements on Hexes... but I guess for really low order elements merely testing the cache is relatively time consuming.

@jwpeterson, can you kick that test and make sure I'm not wrong about it being a false positive.  :+1: to merge when that's done.

jwpeterson:
> @jwpeterson, can you kick that test and make sure I'm not wrong about it being a false positive.

@roystgnr You should also have the ability to invalidate tests if you wish.  Just click the "Details" link and then hit the Invalidate button on the page that takes you to.

roystgnr:
Ah; thanks!

roystgnr:
Going to merge this; if we're seeing the same problem on another PR then I'm satisfied that it's Civet box weirdness and not anything real.

PR #1156, opened 2016-11-10
NonlinearImplicitSystem::ComputeVectorSubspace as an abstract class

balborian:
Declaring void operator()(std::vector<NumericVector<Number> *> & sp, sys_type & s) as a pure virtual function to make the class abstract, otherwise it needs an implementation.

### DIFF
--- a/include/systems/nonlinear_implicit_system.h
+++ b/include/systems/nonlinear_implicit_system.h
@@ -145,7 +145,7 @@ public:
      * It must be implemented by the user in a derived class.
      */
     virtual void operator()(std::vector<NumericVector<Number> *> & sp,
-                            sys_type & s);
+                            sys_type & s) = 0;
   };
 
   /**


jwpeterson:
:+1:  Was this causing a linking error for you? Just wondering how you encountered the issue...

balborian:
Yes, I could not subclass  NonlinearImplicitSystem::ComputeVectorSubspace because of this. I found this answer of a similar issue: http://stackoverflow.com/a/16460763/2839757

jwpeterson:
@roystgnr in case you noticed the build failure on this PR...there appears to be an issue with `cppunit` on some of our build boxes being linked against a different GCC's libstdc++.so than libmesh is linked against.  We're trying to straighten it out, but in the meantime the workaround is just to invalidate the jobs and run them on hosts without "q" in the name.

PR #1157, opened 2016-11-14
Fix "maybe uninitialized" warning from GCC 5.3.0.

jwpeterson:
The logic in this function ensures that it asserts in debug mode
rather than using an uninitialized variable.  In optimized mode,
old_var_value could be returned without being initialized, at best
leading to non-repeatable results.

This type of initialization should work for Real, TypeVector, and
TypeTensor.  Not sure if we really need to worry about anything
else...

```
../include/libmesh/parsed_fem_function.h: In member function 'Output libMesh::ParsedFEMFunction<Output>::get_inline_value(const string&) const [with Output = double; std::__cxx11::string = std::__cxx11::basic_string<char>]':
../include/libmesh/parsed_fem_function.h:514:10: warning: 'old_var_value' may be used uninitialized in this function [-Wmaybe-uninitialized]
   return old_var_value;
```

### DIFF
--- a/include/numerics/parsed_function.h
+++ b/include/numerics/parsed_function.h
@@ -319,7 +319,7 @@ ParsedFunction<Output,OutputGradient>::get_inline_value (const std::string & inl
 #ifndef NDEBUG
   bool found_var_name = false;
 #endif
-  Output old_var_value;
+  Output old_var_value(0.);
 
   for (unsigned int s=0; s != _subexpressions.size(); ++s)
     {


roystgnr:
Looks okay to me.  I wonder if opt mode can still optimize away the redundant assignment?  At some point in the distant future I'm going to want to put big operator-overloaded-vector types through here, and the malloc-vs-calloc difference can't be hidden, but for the foreseeable future I bet even TypeVector old_var_value ought to end up entirely in registers and zero-initialization ought to be free.

jwpeterson:
Thanks for taking a look!

PR #1161, opened 2016-11-18
Removed overzealous assert

dknez:
I hit this assert when I called ReplicatedMesh::stitch_meshes to connect a 1D mesh and 2D mesh. It looks like it's an overzealous assert, now that we support multi-dimensional meshes.


dknez:
This should be an innocuous change, just removed an apparently out-of-date assert. Any objections to merging it?

roystgnr:
Looked fine; sorry for the slow response.

Issue #1159, opened 2016-11-16
optional filename argument to --redirect-stdout


permcody:
In an attempt to make MOOSE RunException testing more robust, we need the ability to write output from a parallel run that cannot be interleaved. To accomplish this, we plan to use the `--redirect-stdout` flag when running these types of test through the TestHarness. We'll concatenate the resulting files written to the filesystem scanning for our error message. 

In order to do this efficiently though, we need the ability to control the names of these files so that we can run multiple instances of test programs in the same directory at the same time without them stomping on each other.

Tag @jwpeterson, @milljm

jwpeterson:
While we're at it, it would probably be good to deprecate the `--redirect-stdout` flag, since it actually redirects both stdout and stderr now.  We could add a new flag
```
--redirect-output filebase
```
that takes an optional argument for specifying the base filename to which libmesh would append the processor id when writing.

jwpeterson:
Added in #1163.

PR #1163, opened 2016-11-18
Add --redirect-ouput [filename] option

jwpeterson:
This turned out to be a bit more involved than I thought it was going to be, and I ended up fixing two other bugs in the process.

1. The first was that we were calling `command_line->parse_command_line()` without first clearing the existing `command_line` object, so the `GetPot` object actually had two copies of `argv`.  No telling what else this might affect, but hopefully nothing.
1. Fixing the first bug led me to uncover an off-by-one bug in `GetPot` which had been covered up by us having two copies of all the command line args all this time.

Refs #1159.

### DIFF
--- a/include/base/getpot.h
+++ b/include/base/getpot.h
@@ -1765,7 +1765,7 @@ GetPot::search(const char* Option)
     return false;
 
   // (*) second loop from 0 to old cursor position
-  for (unsigned c = 1; c < OldCursor; c++)
+  for (unsigned c = 1; c <= OldCursor; c++)
     {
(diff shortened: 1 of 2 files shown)


include/base/getpot.h
```diff
@@ -1765,7 +1765,7 @@ GetPot::search(const char* Option)
     return false;
 
   // (*) second loop from 0 to old cursor position
-  for (unsigned c = 1; c < OldCursor; c++)
+  for (unsigned c = 1; c <= OldCursor; c++)
```
permcody: image

PR #1164, opened 2016-11-21
MeshBase option to stop delete_remote_elements()

roystgnr:
This ought to be useful in cases where users don't want to add
GhostingFunctors and limit remote element deletion until after the
initial Mesh preparation.

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -646,6 +646,10 @@ public:
    *
    * The argument to skip renumbering is now deprecated - to prevent a
    * mesh from being renumbered, set allow_renumbering(false).
+   *
+   * If this is a distributed mesh, local copies of remote elements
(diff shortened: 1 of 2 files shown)


roystgnr:
@permcody - this is what I hashed out with @friedmud, but I'd appreciate it if you take a look before we merge.

permcody:
I was off yesterday but saw the discussion. This looks good. So this means that we'd simply call the separate deletion method later. As long as that works, I'd say merge this and we'll get it worked into MOOSE.

roystgnr:
Call the separate deletion, but also turn automatic deletion back on after you've got the functors added; otherwise you could get a bunch of irrelevant elements "left behind" whenever you do much repartitioning.

friedmud:
Awesome - thanks @roystgnr !

PR #1168, opened 2016-11-22
Use ReplicatedMesh/DistributedMesh everywhere

roystgnr:
This gets rid of all SerialMesh/ParallelMesh use in the library, tests, and examples, except for a couple template specializations that might be necessary for backwards compatibility.

The SerialMesh and ParallelMesh shim classes are still available in serial_mesh.h and parallel_mesh.h, but the new ReplicatedMesh and DistributedMesh are now also available in replicated_mesh.h and distributed_mesh.h

We ought to backport this to 1.1 before the release.  In hindsight I ought to have done this months ago.


roystgnr:
As per IM discussion with @jwpeterson, I'm merging this into master.  I'll cherry-pick only the first 3 commits onto 1.1.0 and I'll re-bootstrap again there.

jwpeterson:
:+1:

PR #1172, opened 2016-11-28
Getpot::get_subsection_names()

roystgnr:
@pbauman, the ideal method to submit feature requests is via the issue tracker rather than via random comments in source code, but I eventually noticed this anyway:

https://github.com/grinsfem/grins/blob/f4c70889c59cd869d536b022a3c37cfeabea9178/src/common/src/builder_helper.C#L60

And at about the same time I hit a use case for the same feature, so here's my crack at it.

### DIFF
--- a/include/base/getpot.h
+++ b/include/base/getpot.h
@@ -323,8 +323,29 @@ public:
 
   inline unsigned vector_variable_size(const char* VarName) const;
   inline unsigned vector_variable_size(const std::string& VarName) const;
+
+  /*
+   * Return a list of all variables set by the current input
+   */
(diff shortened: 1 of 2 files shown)


pbauman:
Yeah, sorry for being lazy. Thanks for implementing it.

roystgnr:
Thank me when I implement it *correctly*.  Still testing.

roystgnr:
Wow; I didn't realize std::string::back() was C++11 only.  Good catch Civet.

roystgnr:
In addition to passing unit tests this works as a drop-in replacement for the GRINS code.

PR #1173, opened 2016-11-29
Fix for VTK 7.1 compatibility

jwpeterson:
This should also go into the libmesh update PR over in idaholab/moose#8132.

### DIFF
--- a/include/utils/ignore_warnings.h
+++ b/include/utils/ignore_warnings.h
@@ -32,6 +32,7 @@
 #pragma clang diagnostic ignored "-Wnested-anon-types"
 #pragma clang diagnostic ignored "-Wsign-compare"
 #pragma clang diagnostic ignored "-Wunused-private-field"
+#pragma clang diagnostic ignored "-Winconsistent-missing-override"
 #endif
 
 #if defined(__GNUC__) && !defined(__INTEL_COMPILER) && !defined(__clang__)
--- a/src/mesh/vtk_io.C
+++ b/src/mesh/vtk_io.C
@@ -400,7 +400,11 @@ void VTKIO::nodes_to_vtk()
       _local_node_map[node.id()] = local_node_counter;
 
       // add point
+#if VTK_VERSION_LESS_THAN(7,1,0)
       pcoords->InsertNextTupleValue(pnt);
+#else
+      pcoords->InsertNextTuple(pnt);
+#endif
     }
 
   // add coordinates to points


jwpeterson:
This gets libmesh compiling with VTK-7.1 for me, but there is still a strange error at runtime in examples/introduction_ex3:
```
ERROR: In /Users/petejw/software/vtk/7.1/VTK-7.1.0/IO/XML/vtkXMLWriter.cxx, line 731
vtkXMLPUnstructuredGridWriter (0x7fe4bb509dd0): Writer called with no FileName set.

ERROR: In /Users/petejw/software/vtk/7.1/VTK-7.1.0/Common/ExecutionModel/vtkExecutive.cxx, line 784
vtkCompositeDataPipeline (0x7fe4bb50b980): Algorithm vtkXMLPUnstructuredGridWriter(0x7fe4bb509dd0) returned failure for request: vtkInformation (0x7fe4bb50d5c0)
  Debug: Off
  Modified Time: 205
  Reference Count: 1
  Registered Events: (none)
  Request: REQUEST_DATA
  ALGORITHM_AFTER_FORWARD: 1
  FORWARD_DIRECTION: 0
  FROM_OUTPUT_PORT: -1
```

I've verified that we definitely *do* set the `FileName` before calling `writer->Write();`, which is where the error message above originates from.  It would be good if someone else could actually confirm this behavior happens for them on VTK-7.1, and even better if someone could actually get to the bottom of the issue.

Tagging @capitalaslash as I believe he has done some VTK work in the past...

capitalaslash:
@jwpeterson I cannot reproduce your error using v7.1, introduction_ex3 works fine and generates working output files.

p.s.: I had to change the order of the libraries pushing `libvtkCommonCore` to the end in order to compile with the new version

jwpeterson:
> @jwpeterson I cannot reproduce your error using v7.1, introduction_ex3 works fine and generates working output files.

OK, thank you for checking, it must be that I have a bad VTK build somehow.  I thought this might be the case, as the error really made no sense to me.

> p.s.: I had to change the order of the libraries pushing libvtkCommonCore to the end in order to compile with the new version

Interesting, I did not see any linking issues... Was this on Linux or Mac?  Also if you built VTK from source would you mind sharing your cmake flags?

capitalaslash:
> Interesting, I did not see any linking issues... Was this on Linux or Mac? Also if you built VTK from source would you mind sharing your cmake flags?

the linking issue baffled me too... it does not happen with v7.0. I am on archlinux, I just adapted this

https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/vtk

to v7.1 (patches are no longer required).

jwpeterson:
I've traced this issue down to the following line in vtk_io.C:
```
#define VTK_LEGACY_REMOVE
```

This line is present in libmesh, and when I put it in my test code I see the same error as I see in libmesh.  We originally added this back in 2013 to avoid some warnings, so we can probably get rid of it safely enough... I haven't dug into it in much detail, but we may just be using it incorrectly and have been getting away with it for some time.  The fix is in PR #1177.

PR #1176, opened 2016-12-01
Fix Trilinos ifdefs.

jwpeterson:


src/base/libmesh.C
```diff
@@ -952,7 +952,7 @@ SolverPackage default_solver_package ()
         libMeshPrivateData::_solver_package = PETSC_SOLVERS;
 #endif
 
-#ifdef LIBMESH_HAVE_TRILINOS
+#ifdef LIBMESH_TRILINOS_HAVE_AZTECOO
```
roystgnr: We need both Aztec and Epetra here, don't we?

jwpeterson: Yes, but I discovered that having AztecOO implies having Epetra (Trilinos won't build AztecOO without it) so there's no need to check both.

PR #1178, opened 2016-12-05
PointLocatorBase::locate_node()

roystgnr:
I found myself writing this in an app code, then realizing how easy it was to get wrong due to AMR and tolerance corner cases, then moving it into libMesh because if I ever have to write it again I don't want to have to make and fix the same mistakes again.


friedmud:
Thanks!  I know that I've probably (incorrectly) written this a few times myself...

roystgnr:
@friedmud - if you ever go through and replace your code with a call to this, let me know so we can revisit the design choices.  I put in the most straightforward implementation, not the most efficient, since in my app it wasn't going to be called anywhere near an inner loop.

PR #1180, opened 2016-12-06
is_evaluable(Node)

roystgnr:
Hopefully this is roughly what @permcody was asking for on libmesh-devel last week.

You can now get dof_indices() on a Node, test is_evaluable() on a Node, and create mesh iterator ranges from evaluable_nodes_begin() to end()


permcody:
Yes, this looks good! Thanks!

Issue #1179, opened 2016-12-06
New behavior of vtkXMLPUnstructuredGridWriter in VTK 7.1


jwpeterson:
After getting some help on the [vtkusers list](http://public.kitware.com/pipermail/vtkusers/2016-December/097359.html) I figured out that we've been using the `vtkXMLPUnstructuredGridWriter` incorrectly (but getting away with it) for quite a while.  The issue was originally caught by a MOOSE test, `outputs/vtk.diff_serial_mesh_parallel`, otherwise we probably would not have realized it for quite some time...

The issue is present in VTK-7.1 (and presumably later versions), since they added a "feature" that prevents empty files from being listed as "Piece Source" entries in the master ".pvtu" file.  Unfortunately, this new feature only works if you have MPI enabled in your VTK build, otherwise VTK decides that all processors other than processor 0 have empty files and omits them all from the ".pvtu" master file (despite the various piece files still actually being written and non-empty!).  

The fix suggested on the VTK mailing list is to set up a `vtkMPIController` object, e.g.
```
#include "vtkMPIController.h"
...
vtkMPIController* controller = vtkMPIController::New();
controller->Initialize(&argc, &argv, 1);
controller->SetGlobalController(controller);
...
controller->Finalize(1);
```
so that VTK can correctly do a communication step in order to determine which files are empty.  You have to configure VTK with `-DVTK_Group_MPI:BOOL=ON` to enable this feature, and link against two additional libraries, `-lvtkParallelMPI` and `-lvtkParallelCore` to use the `vtkMPIController` object.

I've tested this in a standalone test code and it works fine, so the question is: do we want to continue to support writing parallel VTK files?  I don't really see it as a key feature, but maybe others do.  It's a bit annoying to have to do VTK initialize/finalize stuff in libmesh.C, but we already do it for other MPI-aware libraries so maybe it's not too bad...?  We'd also have to detect whether a user's VTK has MPI enabled during configure, and disable it in libmesh when it doesn't.

PR #1181, opened 2016-12-06
VTK-7.1 fixes involving vtkMPIController

jwpeterson:
This PR gets our use of `vtkXMLPUnstructuredGridWriter` into a more conformant state, and fixes the MOOSE test broken by VTK-7.1 described in #1179.

~~I'm pretty sure this will work for all VTK-6.x and 7.x versions, have not had a chance to test 5.x yet, but hopefully not many people are still using that old version.  The worst that will happen is that VTK support will not be enabled for 5.x~~.

This also works in VTK-5.10.1.  Have not tested VTK-6.x, but it should be the same as 7.x.

### DIFF
--- a/configure
+++ b/configure
@@ -40070,7 +40070,7 @@ $as_echo "VTK header files not found!" >&6; }
          fi
 
                   if (test $vtkmajor -eq 5); then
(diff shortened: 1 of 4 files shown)


aeslaughter:
Thanks for figuring all this out.

PR #1183, opened 2016-12-07
Actually overwrite constraints when users request

roystgnr:
If a new constraint differs from an old constraint we want to use the
new constraint, not the old.  This is useful for e.g. allowing user
constraints to override just one node of a DirichletBoundary; more
importantly it causes our behavior to match its documentation.

Today I learned, the hard way, about the C++17 method std::map::insert_or_assign(), which behaves the way I *expected* std::map::insert() to behave; the *actual* standard for std::map::insert() might have instead been named std::map::insert_or_give_up().


pbauman:
And I just posted to the wrong PR. Gonna be a good day...

Issue #770, opened 2016-01-07
The node order of PRISM15 does not match with gmsh format.


lw4992:
In gmsh_io.C, PRISM15 was defined as following, 

```
em.add_def(ElementDefinition(PRISM15, 13, 3, 15)); // TODO: what should be done with this on writing?
```

Maybe not correct. Let me fix it.

jwpeterson:
Hi @lw4992, Any luck with this issue?

jwpeterson:
Fixed by #1192.

PR #1192, opened 2016-12-15
Fix gmsh numbering for PRISM15. (Resolves #1191)

### LINKED ISSUE #1191: gmsh numbering for 15-node second order prism incorrect
Both PRISM15 and PRISM18 are currently being assigned to gmsh number 13. PRISM18 should be 13 and PRISM15 should be 18.

lindsayad:
Was formerly given gmsh_number 13 which actually clashes with the (correct) numbering for PRISM18.

### DIFF
--- a/src/mesh/gmsh_io.C
+++ b/src/mesh/gmsh_io.C
@@ -58,7 +58,6 @@ GmshIO::ElementMaps GmshIO::build_element_maps()
   em.add_def(ElementDefinition(HEX8, 5, 3, 8));
   em.add_def(ElementDefinition(TET4, 4, 3, 4));
   em.add_def(ElementDefinition(PRISM6, 6, 3, 6));
-  em.add_def(ElementDefinition(PRISM15, 13, 3, 15)); // TODO: what should be done with this on writing?
   em.add_def(ElementDefinition(PYRAMID5, 7, 3, 5));
 
   // Add elements with non-trivial node mappings
@@ -84,6 +83,14 @@ GmshIO::ElementMaps GmshIO::build_element_maps()
   {
     ElementDefinition eledef(TET10, 11, 3, 10);
     const unsigned int nodes[] = {0,1,2,3,4,5,6,7,9,8};
+    std::vector<unsigned int>(nodes, nodes+eledef.nnodes).swap(eledef.nodes); // swap trick
+    em.add_def(eledef);
+  }
+
+  // PRISM15
+  {
+    ElementDefinition eledef(PRISM15, 18, 3, 15);
+    const unsigned int nodes[] = {0,1,2,3,4,5,6,8,9,7,10,11,12,14,13};
     std::vector<unsigned int>(nodes, nodes+eledef.nnodes).swap(eledef.nodes); // swap trick
     em.add_def(eledef);
   }


jwpeterson:
Thanks for this fix, l'll test whether PRISM15 actually works as well, hopefully later today/tomorrow...

lindsayad:
Cool. I'm looking through now, but is there a place in the libMesh documentation where the node ordering convention is given?

jwpeterson:
There is ascii art in the cell_prism15.h header file, or scroll down to the Detailed Description on this [page](http://libmesh.github.io/doxygen/classlibMesh_1_1Prism15.html)

lindsayad:
Yea, you were definitely right. The node ordering was wrong. Fixed now. Follows the PRISM18 node ordering convention exactly.

jwpeterson:
Great, thanks for figuring this out.

lindsayad:
@jwpeterson Is there an example anywhere of just using libMesh to convert a `.msh` file to exodusII format or something close enough that I can easily take it the rest of the way? I don't think that either the PRISM15 ([this issue](https://github.com/libMesh/libmesh/issues/770)) or PRISM18 node mappings are correct.

Here's first order prism mesh after rendering to exodusII:

first_order_wireframe

Second order mesh (PRISM18) after rendering to exodusII:

second_order_wireframe

jwpeterson:
I'm pretty sure writing a valid PRISM15 from libmesh to Exodus works (see test code below) so the problem is almost certainly on the Gmsh side.  You should be able to modify this code relatively easily to test with Gmsh.
```
#include "libmesh/libmesh.h"
#include "libmesh/mesh.h"

using namespace libMesh;

{
  LibMeshInit init(argc, argv);

  // Create a mesh of the dimension specified on the command line
  Mesh mesh(init.comm());

  // Location of the reference_elements meshes.  This only works if
  // you have LIBMESH_DIR set in your environment...
  std::string libmesh_dir = std::getenv("LIBMESH_DIR");

  // Test that we can correctly compute the volume of the reference element.
  mesh.read(libmesh_dir+std::string("/share/reference_elements/3D/one_prism15.xda"));

  // Write to Exodus file
  mesh.write("prism15_exodus.e");

  return 0;
}
```


To compile the test code, you will need to set `$LIBMESH_DIR` in your environment to the location of a libmesh installation, and use this Makefile:
```
# This Makefile assumes you have libmesh built and installed in $LIBMESH_DIR.

# If the user has no environment variable
# called METHOD, he gets optimized mode.
ifeq (x$(METHOD),x)
  METHOD := opt
endif

# installed libmesh location of libmesh-config script
libmesh_config := $(LIBMESH_DIR)/bin/libmesh-config

# Use the libmesh-config script to determine the usual make variables.
# Note: passing $METHOD along to the libmesh-config script handles the
# case where METHOD is not set and the user does
# make METHOD=dbg foo
# since in this case, METHOD is never set in the environment and thus
# never passed to libmesh-config correctly.
libmesh_CXX      := $(shell METHOD=$(METHOD) $(libmesh_config) --cxx)
libmesh_INCLUDE  := $(shell METHOD=$(METHOD) $(libmesh_config) --include)
libmesh_CPPFLAGS := $(shell METHOD=$(METHOD) $(libmesh_config) --cppflags)
libmesh_CXXFLAGS := $(shell METHOD=$(METHOD) $(libmesh_config) --cxxflags)
libmesh_LIBS     := $(shell METHOD=$(METHOD) $(libmesh_config) --libs)

# File management variables.
srcfiles 	:= $(wildcard *.cc)
opt_executables := $(patsubst %.cc, %-opt, $(srcfiles))

.PHONY: clean

# How to build executables. If you have a source file called foo.cc,
# type 'make foo' to build an executable from it.
%: %.cc
	@echo "Compiling" $<
	@$(libmesh_CXX) $(libmesh_INCLUDE) $(libmesh_CPPFLAGS) $(libmesh_CXXFLAGS) $< -o $@-$(METHOD) $(libmesh_LIBS) $(libmesh_LDFLAGS) $(EXTERNAL_FLAGS)

# Generate assembly
%.s: %.cc
	@echo "Generating assembly" $<
	@$(libmesh_CXX) $(libmesh_INCLUDE) $(libmesh_CPPFLAGS) $(libmesh_CXXFLAGS) -S $< -o $@

# File management rules.
clean:
	@rm -f *~ $(opt_executables)
```

jwpeterson:
> I'm pretty sure writing a valid PRISM15 from libmesh to Exodus works

I don't think Paraview properly displays the PRISM18 elements we write, unfortunately.  I'm not sure exactly what it's doing, but it seems to only recognize the first 6 vertices.  In the image below, there is a PRISM15 on the left and a PRISM18 on the right with the nodes labeled.  The PRISM18 is only displaying 6 vertex nodes.

prism_combined

So it means that you can't use Paraview (I assume your screenshots above came from that?) to visually verify if you got the Prism18 numbering from Gmsh correct.  But you could e.g. compute the volume of the element you read in from Gmsh, and make sure it's what you expect?

lindsayad:
On 12/19/2016 05:02 PM, John W. Peterson wrote:
>
>     I'm pretty sure writing a valid PRISM15 from libmesh to Exodus works
>
> I don't think Paraview properly displays the PRISM18 elements we 
> write, unfortunately. I'm not sure exactly what it's doing, but it 
> seems to only recognize the first 6 vertices. In the image below, 
> there is a PRISM15 on the left and a PRISM18 on the right with the 
> nodes labeled. The PRISM18 is only displaying 6 vertex nodes.
>
> prism_combined 
> <https://cloud.githubusercontent.com/assets/1775907/21332100/843c6122-c603-11e6-8164-0a41d30ee7e9.png>
>
> So it means that you can't use Paraview (I assume your screenshots 
> above came from that?) to visually verify if you got the Prism18 
> numbering from Gmsh correct. But you could e.g. compute the volume of 
> the element you read in from Gmsh, and make sure it's what you expect?
>
That's annoying. Yes screenshots came from Paraview. Good plan.

jwpeterson:
Here's another idea.  The native VTK format supports viewing 18-node prisms.  So the following slightly modified version of the original test code should work:

```
#include "libmesh/libmesh.h"
#include "libmesh/mesh.h"
#include "libmesh/vtk_io.h"

using namespace libMesh;

{
  LibMeshInit init(argc, argv);

  // Create a mesh of the dimension specified on the command line
  Mesh mesh(init.comm());

  // Location of the reference_elements meshes.  This only works if
  // you have LIBMESH_DIR set in your environment...
  std::string libmesh_dir = std::getenv("LIBMESH_DIR");

  // Choose between "prism15" and "prism18" here.
  const std::string elem_name = "prism18";

  // Test that we can correctly compute the volume of the reference element.
  mesh.read(libmesh_dir+std::string("/share/reference_elements/3D/one_" + elem_name + ".xda"));

  // Write to Exodus file
  mesh.write(elem_name + ".e");
  VTKIO(mesh).write(elem_name + ".pvtu");

  return 0;
}
```

You have to have VTK support enabled in libmesh for this to work.  The current El Capitan and Ubuntu 14 packages should have a working VTK, but some recent changes to VTK-7.1 means that older packages' VTK might not work with libmesh HEAD at the moment.

jwpeterson:
Here's a screenshot of the PRISM18 node numbering (which happens to match libmesh's) in Paraview.

prism18_vtk

lindsayad:
Alright, using the volume calculation routines, I think I've convinced myself that this PR is good as it stands. On a side note, I'm sad that exodus doesn't allow multiple element_types on a given subdomain. I might also have to look into the Paraview source...like here I'm hoping a fix might be as easy as changing a few node orderings in their exodus reader.

lindsayad:
With Paraview though this boils down to the exodus format... And there's not really a standard for PRISM18 or WEDGE18 elements in exodus. Neither the exodus documentation or the PATRAN documentation to which the former refers describes/supports PRISM18. The latter skips from 15 and 16 node elements to 20 node elements. So without a standard no one can really be wrong (or right).

lindsayad:
Attached screenshot shows mesh after conversion from `.msh` format to `.vtu` format using libMesh. Looks great! Also realized that gmsh supports direct output as vtk, which is nice.

vtu_second_order_converted_from_gmsh_in_libmesh

PR #1193, opened 2016-12-15
AdjointRefinement ghosted adjoint compatibility

roystgnr:
The init() overload we were using earlier was only appropriate for
serial or parallel adjoint vectors.

Thanks to @vikramvgarg for catching this; we should wait for him to checkout the proposed fix before merging.


vikramvgarg:
Compiled and checked this out with my 1d advection example (system.add_vector("adjoint_solution0", false, GHOSTED); ) and it ran fine.

PR #1196, opened 2016-12-20
Update to TreeNode to avoid infinite regression

dknez:
The motivation here is to avoid an infinite regression that can occur if you have many elements connected to one node. In this case it can happen that refining the TreeNode does not reduce the number of elements in the child TreeNode, and hence we never get below the target_bin_size threshold.

The fix proposed here is to increase the target bin size when we reach a specified refinement level.

### DIFF
--- a/include/utils/tree_node.h
+++ b/include/utils/tree_node.h
@@ -209,6 +209,15 @@ private:
   const unsigned int tgt_bin_size;
 
   /**
+   * This specifies the refinement level beyond which we will
+   * scale up the target bin size in child TreeNodes. We set
+   * the default to be 10, which should be large enough such
+   * that in most cases the target bin size does not need to
+   * be increased.
+   */
+  unsigned int target_bin_size_increase_level;
+
+  /**
    * Does this node contain any infinite elements.
    */
   bool contains_ifems;
@@ -228,6 +237,7 @@ TreeNode<N>::TreeNode (const MeshBase & m,
   mesh           (m),
   parent         (p),
   tgt_bin_size   (tbs),
+  target_bin_size_increase_level(10),
   contains_ifems (false)
 {
   // libmesh_assert our children are empty, thus we are active.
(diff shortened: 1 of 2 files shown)


include/utils/tree_node.h
```diff
@@ -209,6 +218,17 @@ class TreeNode
...
 
   /**
+   * The current refinement level of this TreeNode.
+   */
+  unsigned int refinement_level;
+
+  /**
+   * This specifies the refinement level beyond which we will
+   * scale up the target bin size in child TreeNodes.
+   */
```
roystgnr: Comment on the default here?  It is a (deserved!) behavior change.  We should also wait and see whether others prefer more or less than 10 levels, though that certainly sounds reasonable to me.

include/utils/tree_node.h
```diff
@@ -154,6 +154,15 @@ class TreeNode
...
 
+  /**
+   * Get the refinement level of this TreeNode.
+   */
+  unsigned int get_refinement_level() const;
+
+  /**
+   * Set the refinement level of this TreeNode.
+   */
+  void set_refinement_level(unsigned int refinement_level);
```
dknez: Wait a sec, it looks like we already have TreeNode::level that works in the way you propose! Oops. I can update it to use that instead, if you agree.

roystgnr: Yeah, let's do that.  If PointLocatorTree construction suddenly starts taking too long for anyone then we can reconsider whether O(N log N) is good enough.

include/utils/tree_node.h
dknez: I guess you mean to add something like "We set the default to be 10 levels." at the end of this comment?

roystgnr:
Looks good after the change to use level().

dknez:
OK, I made those changes, so it should be in good shape now I think.

If anyone has any comments re the default choice of 10 levels, let me know. I think that should be large enough such that we only trigger this new behavior (enlarging the target_bin_size) in "pathological cases", like the case that I hit.

dknez:
I think this should be OK to merge now. I'll do the merge today, unless anyone has any further comments.

Issue #1197, opened 2016-12-21
Connection of finite and infinite element region


BalticPinguin:
Hi,
I have a quite unspecific error, but I have the impression that it still might be related to some error or unconsistency in libmesh:
I try to compute some Coulomb functions (mainly some spherical oscillations which may have a few angular nodes)
In the figures below, the finite elements are in a spherical region (exactly there, where the cuts are) where the box size and order of the radial polynomial in infinite region are varied.
The figures are obtained from computing the solution on a uniform grid for each point as
```
    Point map_point=FEInterface::inverse_map(3, fe_type, elem, q_point, TOLERANCE, true);
    FEComputeData data(es, map_point);
    FEInterface::compute_data(3, fe_type, elem, data);

     unsigned int n_sf= cfe->n_shape_functions();
     for (unsigned int i=0; i<n_sf; i++){-                         
              soln+=(*solution_vect)(dof_indices[i])*data.shape[i];  
      }
```

Here the interface shows strong irregularity
![snapshot](https://cloud.githubusercontent.com/assets/2708704/21401972/b5b768aa-c7b6-11e6-8990-be0d42530ebc.png)

Besides the the fact that the solution is almost 0 in the finite region, the oscillations in the infinite part (which are also present in the absolute value, so its not a change in phase) are far too stiff: The edge is 8 units and the wavelength should be ~6 units...
![re0](https://cloud.githubusercontent.com/assets/2708704/21401919/6bdbf4e4-c7b6-11e6-9b8f-53992728855b.png)

So long explanation, now my real question: Could it be that the DOFs of finite and infinite region are more or less independent? 
Do you have some other idea what to check?
Thank you very much,
Hubert

BalticPinguin:
Ok, after some more investigation, this behaviour shows up only in a certain region for parameters so it doesn't seem to be a bug in libmesh, probably.
I'll close this issue respectively.

jwpeterson:
OK, sorry we could not be more helpful...

PR #1198, opened 2016-12-22
Resolved issue with linear dependence in inverse_map().

BalticPinguin:
I used the inverse_map() scheme quite intensely in the last weeks for infinite elements and there seem to be some instabilities left.
The main problem I had is fixed with this patch.
I hope, it looks reasonable.
Best,
Hubert

### DIFF
--- a/src/fe/inf_fe_map.C
+++ b/src/fe/inf_fe_map.C
@@ -29,6 +29,18 @@
 namespace libMesh
 {
 
+bool libmesh_isfinite(Real value)
+{
+  // check if \p value is infinite
+  if( value > std::numeric_limits<Real>::max()
+         || value <  std::numeric_limits<Real>::min() )
+     return false;
+  // check if \p value is nan
+  if( value != value)
+     return false;
+  // \p value is finite
+  return true;
+}
 
 
 // ------------------------------------------------------------
@@ -187,6 +199,13 @@ Point InfFE<Dim,T_radial,T_map>::inverse_map (const Elem * inf_elem,
            +o(2)*p1(0)*p0(1)+o(0)*p0(2)*p1(1));
 
 
+        if( !libmesh_isfinite(c_factor)){
+           // The above system is badly posed.
+           // It should only happen when \p physical_point is not
+           // in \p inf_elem. In this case, any point that is not in the
+           // element is a valid answer.
+           return o;
+        }
         // Compute the intersection with
         // {intersection} = {fp} + c*({fp}-{o}).
         intersection.add_scaled(fp,1.+c_factor);


jwpeterson:
@brianmoose what's the right way to add external developers to the "automatically activated testing" list?

src/fe/inf_fe_map.C
```diff
@@ -187,6 +187,13 @@ Point InfFE<Dim,T_radial,T_map>::inverse_map (const Elem * inf_elem,
            +o(2)*p1(0)*p0(1)+o(0)*p0(2)*p1(1));
 
 
+        if(! std::isfinite(c_factor)){
```
jwpeterson: `std::isfinite` is C++11-only code, unfortunately.  In C99 there is an `isinf(x)` macro which might be possible to use instead.  I think we already require C99 for other things...

roystgnr: We use C99 for isnan() if the user has a C99 C compiler, but even in that case we fall back on the IEEE "a != a" isnan-equivalent if the user doesn't have a C99 C compiler.

We definitely *shouldn't* be requiring a C99-cum-C++98 C++ compiler.  Such frankencompilers did exist, but they're technically non-conforming to *either* standard.  IIRC we added the libmesh_isnan.c circumlocutions after a compiler upgrade took *away* C99 isnan().

The right thing to do here is probably another shim function, but without anything so ridiculous as a C file this time.  libmesh_isfinite() could use std::isfinite() if C++11 is supported, or test ```a > numeric_limits<T>::max() || a < -numeric_limits<T>::max()``` otherwise?

jwpeterson:
@BalticPinguin does this fix or have anything to do with the issue you raised in #1197?

BalticPinguin:
No, this is an independent issue.
Concerning the  libmesh_isfinite() function: I can write one, it shouldn't be a thing. But can you tell me where to place it at best?

jwpeterson:
> Concerning the libmesh_isfinite() function: I can write one, it shouldn't be a thing. But can you tell me where to place it at best?

For this PR you can just put it in `inf_fe_map.C` since that is the only place it's needed.  I'm not 100% sure about @roystgnr's implementation based on `numeric_limits::max` (for instance subnormal numbers should still be finite but they might be less than -max?) but if it works for your test cases it's fine with me.

brianmoose:
@jwpeterson The person initiating the PR needs to be a collaborator on the repository for automatic testing. Add them on the `Repo -> Settings -> Collaborators & teams` page.

jwpeterson:
> The person initiating the PR needs to be a collaborator on the repository

Thanks, invite sent to @BalticPinguin.

permcody:
@brianmoose - Do we still have the capability to initiate testing manually for people who are not in the collaborators group?

jwpeterson:
On Tue, Jan 3, 2017 at 10:43 AM, Cody Permann <notifications@github.com>
wrote:

> @brianmoose <https://github.com/brianmoose> - Do we still have the
> capability to initiate testing manually for people who are not in the
> collaborators group?
>

Yes, I just used it today for Roger's PR.

-- 
John

BalticPinguin:
Finally pushed the updated version. I hope, this is portable now!?

src/fe/inf_fe_map.C
```diff
@@ -29,6 +29,18 @@
 namespace libMesh
 {
 
+bool libmesh_isfinite(Real value)
+{
+  // check if \p value is infinite
+  if( value > std::numeric_limits<Real>::max()
+         || value <  std::numeric_limits<Real>::min() )
```
jwpeterson: `std::numeric_limits<Real>::min()` is a positive quantity.  For example, on my system:
```
min()=2.225074e-308
max()=1.797693e+308
epsilon()=2.220446e-16
```
so even if you want to require `c_factor` to be positive, I don't think this is a valid implementation in general.  Let me take a look at what the best way to do this is...

jwpeterson:
@roystgnr what do you think about implementing `libmesh_isinf()` in terms of `libmesh_isnan()`, like [this](http://stackoverflow.com/a/2249173/659433)?

src/fe/inf_fe_map.C
BalticPinguin: ah! of course.
But this 
'''if( std::abs(value) > std::numeric_limits<Real>::max())'''
should work, doesn't it?

jwpeterson: I'm not sure if comparisons against `max` are really the right way to go.  At least in this stack overflow [thread](http://stackoverflow.com/questions/2249110/how-do-i-make-a-portable-isnan-isinf-function/2249173) no one is taking that approach, and it seems like those people are knowledgable.  I have a commit based on your 26a697d which I am testing now, will post a PR if it seems to work.

BalticPinguin: Maybe you are right that the comparison is not very smart; and the suggestion in your link seems to be much more elegant to me.
But if you are working on this, I am fine. Thank you!

Issue #1201, opened 2017-01-03
Fix path search for inline Trilinos build support


rppawlo:
The CASL project builds libmesh inline with Trilinos. This means that we don't install a Trilinos build prior to configuring libmesh. Recent changes to libmesh have broken the detection process for ifpack, epetra and epetraext during casl builds. An extra search path needs to be added to the trilinos.m4 to allow for casl builds. Will submit a pull request with the changes shortly.

jwpeterson:
Hmm.. I recently also added a test for the `Epetra_CombineMode.h` header file.  If the `Ifpack_config.h` test is not working for you, this one probably is not either...

rppawlo:
yes. had to change that slightly. epetra, ifpack and epetraext stopped working in casl builds. pull request #1202 fixes this for us. I did change the file checked in epetra to one that is pregenerated in the build directory tree as opposed to the source directory tree. For installed trilinos all files are in the same place but before installation, the generated headers and source headers are in separate locations. Hope this is ok.

jwpeterson:
Fixed by #1202.

PR #1202, opened 2017-01-03
Add extra trilinos package search paths

rppawlo:
issue #1201

### DIFF
--- a/m4/trilinos.m4
+++ b/m4/trilinos.m4
@@ -168,7 +168,9 @@ AC_DEFUN([CONFIGURE_TRILINOS_10],
        dnl ------------------------------------------------------
        AC_CHECK_HEADER([$withtrilinosdir/include/Ifpack_config.h],
                        [enableifpack=yes],
-                       [enableifpack=no])
+                       [AC_CHECK_HEADER([$withtrilinosdir/packages/ifpack/src/Ifpack_config.h],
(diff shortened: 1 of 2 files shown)


m4/trilinos.m4
```diff
@@ -197,9 +201,11 @@ AC_DEFUN([CONFIGURE_TRILINOS_10],
        dnl header files whose absence indicates Epetra is not available,
        dnl we just choose one here that libmesh actually includes.
        dnl ------------------------------------------------------
-       AC_CHECK_HEADER([$withtrilinosdir/include/Epetra_CombineMode.h],
```
jwpeterson: This needs to remain a test for `Epetra_CombineMode.h`

jwpeterson: Unless the existence of Epetra_config.h definitely guarantees the other is present... since libmesh directly includes the other.

rppawlo: Yes - the Epetra_config.h should guarantee the existence of Epetra_CombineMode.h. The Epetra_config.h is generated during configuration so it is actually a more strict test since it makes sure that trilinos has been configured correctly as well.

jwpeterson: 👍

rppawlo: Thanks!

PR #1204, opened 2017-01-05
UniquePtr test

roystgnr:
We use UniquePtr rather than AutoPtr now; our test coverage should probably reflect that.

Noticed this due to the AutoPtr libmesh_deprecated() warning getting called from "make check".


jwpeterson:
This change should be fine.  I originally wrote this thinking that we only needed to test that the old, deprecated AutoPtr `safe_bool` stuff was working, but that test is not necessary if libmesh is configured to use a more modern `unique_ptr` type.

PR #1205, opened 2017-01-05
Loosen tolerance on straight-line check in Edge3::volume().

jwpeterson:
A user [reported](https://groups.google.com/d/topic/moose-users/-8eC6TVwT5g/discussion) a failure in Edge3::volume() for the otherwise normal element:

x0 = (0.0041755731555024704, 0.02103, 0.)
x1 = (0.0042782212723039047, 0.02103, 0.)
x2 = (0.0042268963965901322, 0.02103, 0.)

Loosening the straight-line tolerance check is sufficient to fix this
particular case, but the failure was actually caused by passing a nan
to the std::log function.  In debug mode, we now verify that the
argument is valid before calling std::log.

### DIFF
--- a/src/geom/edge_edge3.C
+++ b/src/geom/edge_edge3.C
@@ -179,7 +179,7 @@ Real Edge3::volume () const
   const Real c = B.norm_sq();
 
   // Degenerate straight line case
-  if (a < TOLERANCE*TOLERANCE*TOLERANCE)
+  if (a < TOLERANCE*TOLERANCE)
     return (this->point(1) - this->point(0)).norm();
 
   const Real ba=b/a;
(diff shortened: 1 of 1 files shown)


jwpeterson:
@jasondhales this fixes the issue you raised on the MOOSE mailing list.

jwpeterson:
This test actually seems to be failing consistently now, but I don't know what's causing it...
```
Working Directory: /home/moosetest/civet/client_root_0/moose/test/tests/interfacekernels/1d_interface
Running command: mpiexec -n 2 /home/moosetest/civet/client_root_0/moose/test/moose_test-dbg --n-threads=1 -i coupled_value_coupled_flux.i --error --no-gdb-backtrace

Assertion `mesh.comm().semiverify (elem ? &bcids : nullptr)' failed.Assertion `mesh.comm().semiverify (elem ? &bcids : nullptr)' failed.

[1] ../src/mesh/mesh_tools.C, line 1329, compiled Jan  9 2017 at 08:53:46
[0] ../src/mesh/mesh_tools.C, line 1329, compiled Jan  9 2017 at 08:53:46
```
This assert wasn't added recently or anything.  I'll see if I can reproduce it on my system, but it should not hold up this PR, which has nothing to do with it.

PR #1207, opened 2017-01-06
CheckpointIO test barrier between read and write

roystgnr:
It looks as if my failure here was due to a race condition.  It's no longer repeatable after adding the barrier(), anyway.


roystgnr:
Thanks!

PR #1210, opened 2017-01-10
Reorder paths starting with /lib too

jwpeterson:
Linux and OSX don't really use `/lib` for much (or at all), but other UNIX systems might?  At least one user on ANL's "Blues" machine seemed to have HDF5 installed in /lib.  So it shouldn't hurt anything to reorder these paths to the end of the list of libraries as well.

In the process of testing the script, I also figured out that some of our regexes were too greedy, that is now fixed as well.

### DIFF
--- a/contrib/bin/strip_dup_libs.pl
+++ b/contrib/bin/strip_dup_libs.pl
@@ -73,6 +73,18 @@ foreach( @cleaned_up_libs_third ) {
     {
         push @system_lib_paths, $_;
     }
+    # Discard library paths starting with /lib. This directory is not
+    # even present on OSX and it is typically fairly empty on Linux,
+    # but it may be present on other UNIX systems.
+    elsif( ($_=~/-L\/lib/) )
+    {
+        push @system_lib_paths, $_;
+    }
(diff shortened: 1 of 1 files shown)


friedmud:
This is cool... does it fix the issues with having a "system" (package manager) installed MPI on Ubuntu and trying to use the MOOSE package?

jwpeterson:
The Ubuntu issues should have already been fixed (?), since their stuff lives in `/usr/lib` and we were already re-ordering those paths.  This new patch just reorders `/lib` paths to try and help out with an issue on ANL Blues.

jwpeterson:
It doesn't fix all possible issues with having multiple MPIs installed on a system (as we have seen from the moose-users mailing list, people mess this up all the time) but it does prevent accidentally linking against a libmpi.so in /usr/lib instead of the MOOSE one...

friedmud:
Gotcha - this has always been a bit of a pain... I'm glad there has been good progress on it.

I still always just tell people to uninstall their system MPI (why would you want it when our package stuff is so good ;-)

PR #1211, opened 2017-01-13
Change to TreeNode element counting

dknez:
Optionally skip lower dimensional element in element counting in TreeNode.

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -871,6 +871,18 @@ public:
   void clear_point_locator ();
 
   /**
+   * In the point locator, do we count lower dimensional elements
+   * when we refine point locator regions? This is relevant in
+   * tree-based point locators, for example.
+   */
+  void set_count_lower_dim_elems_in_point_locator(bool count_lower_dim_elems);
+
+  /**
+   * Get the current value of _count_lower_dim_elems_in_point_locator.
+   */
+  bool get_count_lower_dim_elems_in_point_locator() const;
+
+  /**
    * Verify id and processor_id consistency of our elements and
    * nodes containers.
    * Calls libmesh_assert() on each possible failure.
@@ -1247,6 +1259,12 @@ protected:
    * and it operates on a constant reference to the mesh, this is OK.
    */
   mutable UniquePtr<PointLocatorBase> _point_locator;
+
+  /**
+   * Do we count lower dimensional elements in point locator refinement?
+   * This is relevant in tree-based point locators, for example.
+   */
+  bool _count_lower_dim_elems_in_point_locator;
 
   /**
    * A partitioner to use at each prepare_for_use().
--- a/src/mesh/mesh_base.C
+++ b/src/mesh/mesh_base.C
@@ -54,6 +54,7 @@ MeshBase::MeshBase (const Parallel::Communicator & comm_in,
   _n_parts       (1),
   _is_prepared   (false),
   _point_locator (),
+  _count_lower_dim_elems_in_point_locator(true),
   _partitioner   (),
 #ifdef LIBMESH_ENABLE_UNIQUE_ID
   _next_unique_id(DofObject::invalid_unique_id),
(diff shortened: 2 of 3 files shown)


dknez:
This was motivated by the same use-case that required #1196: A 1D/3D mesh where the 1D elements corresponded to a "spider node" where many elements connect at a single node. I ran into a case that had a very dense spider node, and even with the fix from #1196 in place the PointLocator crashed (in this case it used a huge amount of RAM as the tree kept being refined).

The fix here is to by default ignore "lower dimensional elements" when we calculate the current bin size. This seems reasonable to me, but I'd be interested in input from others, e.g. @roystgnr

roystgnr:
It seems like this just introduces a converse failure case.  Imagine you're solving shallow water or Hele-Shaw or some other 2D flow equations, but then you use model refinement to switch to 3D Navier-Stokes in part of the domain: as soon as you add a handful of 3D cells, the rest of your domain no longer gets a refined tree in your point locator?

On the other hand, I can't think of a better idea.  Anisotropic trees would work well in more circumstances, I suppose; but I still can't think of an *easy* better idea.

Let count_lower_dim_elems default to true?  If the best we can do is add crazy heuristics then at least we should try to stay as backwards-compatible as possible with them.

dknez:
I agree with your converse failure case, but in that case we should set count_lower_dim_elems to true. I'll change it to default to true, good idea.

However, it occurs to me that there's no easy way to change the value of count_lower_dim_elems for a Tree, or for a PointLocator. The problem is that everything gets initialized when the PointLocator is constructed. The best idea I can come up with is to add a boolean count_lower_dim_elems to MeshBase, since then the TreeNodes can access that directly. But that might be considered ugly? Let me know what you think.

dknez:
I added a setter/getter in MeshBase, as suggested in my previous comment. This works well for me. Let me know if it's OK with you?

roystgnr:
:+1: as far as I'm concerned; if nobody else screams let's merge when it passes tests.

jwpeterson:
I don't have a better solution, but this certainly feels like we're applying band-aids on top of band-aids at this point.  It seems like the current tree-based `PointLocator` is singularly ill-equipped to deal with the multi-dimensional-element meshes (even manifold meshes should be somewhat sketchy I'd imagine?) which are of interest to you, so perhaps a new type of PointLocator should be developed at some point...

dknez:
I don't think the current tree-based point-locator is ill-equipped for multi-dimensional meshes, actually. The bounding box refinement approach that it uses works fine for multi-dimensional meshes. It's just that the case that I ran into is kind of pathological in the sense that there's no way to satisfy the "bin size" requirement because so many 1D elements connect at a single point. But I think that is not a typical situation with multi-dimensional meshes, and I think the change in this PR is useful in order to handle the pathological case that I ran into.

In any case, as you said, there may be other ideas that we could consider in the future, e.g. initialize the point locator based on a list of subdomains rather than on the entire mesh.

dknez:
I don't remember how long the tests take to run, but it looks to me like they may have stalled? I would have expect the non-debug tests to be finished by now?

pbauman:
If you go to the CIVET page, they all say they've passed, but, apparently, the results didn't get reported back. Not sure if CIVET or GitHub problem...

roystgnr:
Probably a GitHub problem, based on this morning.  Thanks for checking, @pbauman

roystgnr:
@jwpeterson, I still like my anisotropic tree idea, which would make manifold meshes more efficient, but @dknez is probably right that there's no way to efficiently handle the case where we have a zillion elements meeting at a point.  That point (and its epsilon-neighborhood, assuming we still allow slightly fuzzy searches) must fall in *some* bin, and whatever bin it falls in has to contain a zillion elements.  So the best we can do is let the locator constructor know to step back and admit that it's got to have an insanely oversized bin.

jwpeterson:
> So the best we can do is let the locator constructor know to step back and admit that it's got to have an insanely oversized bin.

OK, understood.

PR #1212, opened 2017-01-18
More robust caching with Clough-Tocher elements

roystgnr:
This fixes corner case errors when an FE is reused through mesh
renumbering such that its next use is on an element with the same id
but a different geometry.


jwpeterson:
👍  Belt and suspenders.

PR #1214, opened 2017-01-22
Only serialize the mesh to processor 0 for Exodus

friedmud:
This changes mesh serialization behavior from an `allgather` to a `gather_to_zero` for Exodus.

This is akin to the fix I did a while ago in #190... we only need stuff serialized to _processor 0_ to output an Exodus file.

The memory savings here will be ENORMOUS for larger meshes.  At INL we now have nodes in our cluster that have 36 processors.  So, if you serialize the mesh to all processors you get 36 copies of the mesh on every node.  If that mesh is 3GB you just destroyed yourself.  With this fix only the head node will need 3GB of room to store the mesh.

In addition, I've also slightly modified the behavior so that when Exodus is using this new capability the mesh will be serialized to processor 0 exactly _once_... and then left there.  Yes, it will hold all of that extra memory on processor zero forever... but it will also be _lightning_ fast to output each subsequent Exodus file.  I think it's a good tradeoff, but I'm definitely willing to have a conversation about it.

This capability will come in handy when you have _medium_ sized problems.  Problems where you want to use `DistributedMesh` but where the extra hassle (and lack of features) of Nemesis output can be avoided by writing Exodus (I happen to be solving a problem like that right now).

### DIFF
--- a/src/mesh/mesh_communication.C
+++ b/src/mesh/mesh_communication.C
@@ -1001,9 +1001,6 @@ void MeshCommunication::gather (const processor_id_type, DistributedMesh &) cons
 // ------------------------------------------------------------
 void MeshCommunication::gather (const processor_id_type root_id, DistributedMesh & mesh) const
 {
-  // The mesh should know it's about to be serialized
-  libmesh_assert (mesh.is_serial());
-
   // Check for quick return
   if (mesh.n_processors() == 1)
     return;
@@ -1061,9 +1058,7 @@ void MeshCommunication::gather (const processor_id_type root_id, DistributedMesh
   // Inform new elements of their neighbors,
   // while resetting all remote_elem links on
   // the appropriate ranks.
-  if ((root_id == DofObject::invalid_processor_id) ||
-      (mesh.comm().rank() == root_id))
-    mesh.find_neighbors(true);
+  mesh.find_neighbors(true);
 
   // All done!
 }
--- a/include/mesh/distributed_mesh.h
+++ b/include/mesh/distributed_mesh.h
@@ -166,6 +166,12 @@ public:
   virtual void allgather() libmesh_override;
 
   /**
+   * Gathers all elements and nodes of the mesh onto
+   * processor zero
+   */
+  virtual void gather_to_zero() libmesh_override;
+
+  /**
    * Deletes all nonlocal elements of the mesh
    * except for "ghosts" which touch a local element, and deletes
    * all nodes which are not part of a local or ghost element
@@ -479,6 +485,11 @@ protected:
    * A boolean remembering whether we're serialized or not
    */
   bool _is_serial;
+
+  /**
+   * A boolean remembering whether we're serialized to proc 0 or not
+   */
+  bool _is_serial_on_proc_0;
 
   /**
    * Cached data from the last renumber_nodes_and_elements call
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -142,6 +142,12 @@ public:
   virtual void allgather () {}
 
   /**
+   * Gathers all elements and nodes of the mesh onto
+   * processor zero
+   */
+  virtual void gather_to_zero() {}
+
+  /**
    * When supported, deletes all nonlocal elements of the mesh
    * except for "ghosts" which touch a local element, and deletes
    * all nodes which are not part of a local or ghost element
--- a/include/mesh/mesh_output.h
+++ b/include/mesh/mesh_output.h
@@ -59,14 +59,14 @@ protected:
    * rendering this object useless.
    */
   explicit
-  MeshOutput (const bool is_parallel_format = false);
+  MeshOutput (const bool is_parallel_format = false, const bool serial_only_needed_on_proc_0 = false);
 
   /**
    * Constructor.  Takes a reference to a constant object.
    * This constructor will only allow us to write the object.
    */
   explicit
-  MeshOutput (const MT &, const bool is_parallel_format = false);
+  MeshOutput (const MT &, const bool is_parallel_format = false, const bool serial_only_needed_on_proc_0 = false);
 
 
 public:
@@ -140,6 +140,14 @@ protected:
    */
   const bool _is_parallel_format;
 
+  /**
+   * Flag specifying whether this format can be written by only
+   * serializing the mesh to processor zero
+   *
+   * If this is false (default) the mesh will be serialized to
+   * all processors
+   */
+  const bool _serial_only_needed_on_proc_0;
 
 private:
 
@@ -165,8 +173,9 @@ private:
 // MeshOutput inline members
 template <class MT>
 inline
-MeshOutput<MT>::MeshOutput (const bool is_parallel_format) :
+MeshOutput<MT>::MeshOutput (const bool is_parallel_format, const bool serial_only_needed_on_proc_0) :
   _is_parallel_format(is_parallel_format),
+  _serial_only_needed_on_proc_0(serial_only_needed_on_proc_0),
   _obj(libmesh_nullptr),
   _ascii_precision (std::numeric_limits<Real>::digits10 + 2)
 {}
@@ -175,8 +184,9 @@ MeshOutput<MT>::MeshOutput (const bool is_parallel_format) :
 
 template <class MT>
 inline
-MeshOutput<MT>::MeshOutput (const MT & obj, const bool is_parallel_format) :
+MeshOutput<MT>::MeshOutput (const MT & obj, const bool is_parallel_format, const bool serial_only_needed_on_proc_0) :
   _is_parallel_format(is_parallel_format),
+  _serial_only_needed_on_proc_0(serial_only_needed_on_proc_0),
   _obj (&obj),
   _ascii_precision (std::numeric_limits<Real>::digits10 + 2)
 {
--- a/include/mesh/mesh_serializer.h
+++ b/include/mesh/mesh_serializer.h
@@ -42,7 +42,7 @@ class MeshBase;
 class MeshSerializer
 {
 public:
-  MeshSerializer(MeshBase & mesh, bool need_serial = true);
+  MeshSerializer(MeshBase & mesh, bool need_serial = true, bool serial_only_needed_on_proc_0 = false);
 
   ~MeshSerializer();
 
--- a/src/mesh/distributed_mesh.C
+++ b/src/mesh/distributed_mesh.C
@@ -1367,6 +1367,8 @@ void DistributedMesh::delete_remote_elements()
 #endif
 
   _is_serial = false;
+  _is_serial_on_proc_0 = false;
+
   MeshCommunication().delete_remote_elements(*this, _extra_ghost_elems);
 
   libmesh_assert_equal_to (this->max_elem_id(), this->parallel_max_elem_id());
@@ -1451,6 +1453,15 @@ void DistributedMesh::allgather()
   this->libmesh_assert_valid_parallel_ids();
   this->libmesh_assert_valid_parallel_flags();
 #endif
+}
+
+void DistributedMesh::gather_to_zero()
+{
+  if (_is_serial_on_proc_0)
+    return;
+
+  _is_serial_on_proc_0 = true;
+  MeshCommunication().gather(0, *this);
 }
 
 
--- a/src/mesh/exodusII_io.C
+++ b/src/mesh/exodusII_io.C
@@ -34,6 +34,8 @@
 #include "libmesh/numeric_vector.h"
 #include "libmesh/exodusII_io_helper.h"
 #include "libmesh/string_to_enum.h"
+#include "libmesh/mesh_communication.h"
+#include "libmesh/parallel_mesh.h"
 
 namespace libMesh
 {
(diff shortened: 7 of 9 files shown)


friedmud:
BTW: This works perfectly.  I'm using it on a 10M element problem that I'm running on 240 processors.  The mesh is already pre-split using the `splitter`.

All of the processes start up and go up to about 170MB of RAM each.  Then, during the first write, process 0 goes up ~6.5GB of RAM.  That first write takes a while (while the mesh is serialized).  After that process 0 _stays_ at 6.5GB of RAM... and each subsequent write is actually pretty _fast_ (considering that each step write out about 300MB... it takes a moment - but overall it's pretty fast).

I think this capability will go a long way toward making our applications more stable.  It removes the last reason for a memory spike that user's might not be aware of.

jwpeterson:
Seems fine to me...  I'd like to see if @roystgnr has any comments before merging.

permcody:
This sounds good but have you thought about adaptivity? I know you aren't
currently using it in your runs so I think it's important to ask the
question. Some of your other commits to exodus and the mesh utilities have
had issues when we've tested them against MOOSE due to oversights in the
adaptivity branches.

mangerij:
This sounds great. When it merges into libMesh, and tested with MOOSE, will
we get some idea of how much it will speed up the tests? Or is the speed up
only going to be seen in medium size problems?

I typically run these medium size problems with Ferret so this should be
super useful to us.

friedmud:
There doesn't appear to be a problem with adaptivity.  What happens with adaptivity is that the remote elements will get re-deleted from processor 0... and then re-synced the next time there is output.

BTW: Our testing system is a mess with `./run_tests -p 2 --distributed-mesh`.  There are ~25 failures or so... I thought we were testing that regularly now.

This PR did add a few more failures... but they were in things related to the ones that are already failing without this PR... so I don't think they are really "new" failures from this PR.  There were ~30 failures with this PR.

All of the failures are in fringe things that shouldn't be done with distributed mesh: like MeshModifiers, etc.

friedmud:
BTW: I have now used this capability on a run with 20M elements.  I think that's about the limit of its usefulness.  It simply takes way too long to do the first sync down to processor 0... and each output step takes too long as well.

For 20M+ you probably really do want to switch to Nemesis output... which still needs a little work.

roystgnr:
gather_to_zero() is clearly a hack.  But it's a wonderful, beautiful hack, it's much much much less of a hack than our current allgather() based hacks, so I'm strongly in favor.

```
if (_is_serial_on_proc_0)
  return;
```

on the other hand, is a horrible hack which will cause people to output invalid data as soon as they try to do mesh refinement, element addition, element deletion, p refinement, etc. in between two serial_only_needed_on_proc_0-output-format writes.  I would honestly be surprised if I didn't already have multiple application input files lying around which would be broken by those two lines.

If you've got a mesh that doesn't change between two writes, then the way to handle efficient I/O is to recognize that at the application level and then not write the mesh the second time, right?  With ExodusII in particular, don't we have a write_timestep() method to handle that case?

roystgnr:
Hmm... actually, most of those cases are avoided by you setting the cache invalid in delete_remote_elements().  This might not break most code.  Let me think for a bit about whether that's sufficient, though.

friedmud:
@roystgnr why do you think this would break anything?  On processor zero it won't be any different than if you were using `ReplicatedMesh`... or if you had EXTREME ghosting... which doesn't break anything.  On all other processors `DistributedMesh` is still doing its thing.

As for the other part: `MeshSerializer` always does `delete_remote_elements()` currently.  Every time you call `write_timestep()` it creates a `MeshSerializer` that copies the mesh everywhere and then deletes it all again.  What do you propose?

friedmud:
BTW: I can't say enough great things about `DistributedMesh`!  Currently running a problem with ~0.5B DoFs on 20M elements on 3840 processors.

Each process is only ~100MB in size!  That is so freaking cool.

roystgnr:
If you're using ReplicatedMesh, then all your mesh operations are done in serial, the same on every processor.  If you remove element 911, then every processor did so and now every processor is missing element 911.

On a DistributedMesh, that isn't the case without syncing... but wow, you're right, this would work like an "extreme ghosting" case, huh?  Everything we need to be synced is already being synced.  Processor 0 would indeed query the entire remainder of the mesh and update its elements accordingly.

Ok, :+1: in theory, but would you give me a day or two to run this by a couple codes and see if theory matches basic practice?

permcody:
Derek I don't think we've had parallel distributed mesh testing for a long
time. I noticed that our distributed mesh tests were running serially quite
awhile ago but it just hasn't been a high priority to get them working.
Yeah I realize that serial testing is almost worthless.

roystgnr:
Even parallel testing is tricky.  #1212 fixed a corner case bug that showed up for me on 24 processors but not on 4.

For years I've been dreaming of hooking up BuildBot to a "make random_check" test suite that does "if (!last_test_failed) increment_seed(); run_tests();", where "run_tests()" uses the PRNG seed to decide everything from "should I start from a 4x4x5 grid or 5x4x4" to "should I use mpirun -np 12 or 13".  At this rate it'll probably be another decade before I implement it.

roystgnr:
Speaking of tricky parallel testing on 24 processors: with this branch I get a failure in adaptivity_ex3 that way:

```
Assertion `(this->comm()).verify(std::string("../src/mesh/boundary_info.C").size())' failed.
Assertion `(mesh.comm()).verify(std::string("../src/mesh/mesh_serializer.C").size())' failed.
```

I'll see if I can dig into it further tomorrow.

friedmud:
Bah - I can look into it too. Does it happen in less than 24 procs?

roystgnr:
Breaks even on 2 for me.

src/mesh/mesh_communication.C
```diff
@@ -1001,9 +1001,6 @@ void MeshCommunication::gather (const processor_id_type, DistributedMesh &) cons
 // ------------------------------------------------------------
 void MeshCommunication::gather (const processor_id_type root_id, DistributedMesh & mesh) const
 {
-  // The mesh should know it's about to be serialized
-  libmesh_assert (mesh.is_serial());
```
permcody: I guess this patch caused this assert to fail? Perhaps we need to look into this part more?

friedmud: It caused that assert to fail because the assert is just wrong.  This `gather()` function can take a processor ID to gather to.  In that case the mesh is NOT serialized (at least not everywhere).

roystgnr: This isn't related to the failure I'm seeing.

That said: this assert is wrong in N_proc out of N_proc+1 cases, but the fix isn't to delete it entirely, the fix is to make it more specific, e.g.  ```libmesh_assert(root_id != DofObject::invalid_processor_id || mesh.is_serial())```

friedmud: Ok  - I'll do it.  I kind of just thought it was overzealous.  Even in the case where you really are doing a full serialization you may not want to actually tell the mesh it's been serialized...

friedmud:
How do you run the examples in parallel using the Makefile?

I tried `mpiexec -n 2 ./example-dbg dimension=2 refinement_type=h` in `adaptivity_ex3` and it's working for me... but I suspect that it must be one of the other sets of parameters...

jwpeterson:
I think you can use the `LIBMESH_RUN` environment variable for this, e.g.
```
LIBMESH_RUN="mpiexec -np 2 libtool --mode=execute" make run
```

Or you can always run the executable using specific options directly with:
```
mpiexec -np 2 libtool --mode=execute ./example-opt --foo --bar
```

roystgnr:
@jwpeterson beat me to it.

When I want to hammer a particular example on many different core counts, I do e.g.

```
(for num in $(seq 1 24); do echo $num; LIBMESH_RUN="mpiexec -np $num" make -C examples/adaptivity/adaptivity_ex3 check || break; done; echo $num)
```

friedmud:
Damn - well... I can't get adaptivity_ex3 to fail for me.

roystgnr:
Looks like I can replicate the problem on my laptop too, so this shouldn't be *too* hard to debug...

roystgnr:
That was easy enough to fix, but it was just hiding a more serious bug in MeshCommunication::gather().  The more serious bug turned out to be a horribly-nasty symptom masking another one-line fix.  I've pushed those fixes to https://github.com/roystgnr/libmesh/tree/mesh_to_zero if anyone wants to look or if @friedmud wants to cherry-pick them here.

Those two bits were enough to get adaptivity_ex3 passing (with --enable-parmesh; sorry I forgot to specify that earlier) on 4 processors for me, but I want to hammer more processor counts and more test suites before we merge.

friedmud:
@roystgnr Ah!  That makes sense - did you see that I had already modified that same line of code:

https://github.com/roystgnr/libmesh/commit/0c296b1cabf615ed19776187c7f60832a0aeaf89#diff-e0f97e4fb343bcbc0b0b937880fd17a6L1064

That `if()` was definitely wrong because `find_neighbors()` is `parallel_only`!  Your fix is the correct fix.

You can create a PR to my branch that I can merge and then everything will auto-update...

roystgnr:
This PR just changed a major years-old libMesh bug from "triggers for me only on one system, only every few weeks at most, can never seem to be replicated, maybe it's something wonky with my NFS setup and not actually a bug" to "wow, introduction_ex5 exits with failure pretty frequently on high processor counts now".  Even on just two processes I can run introduction_ex5 in a loop and it'll die within a couple dozen runs.

I *thought* I figured out a race condition that might cause the problem: ExodusII I/O doesn't wait for all processors to finish before returning from a write(), and the LibMeshInit destructor charges right ahead with cleanup without any kind of parallel synchronization, so maybe the latter sometimes interferes with the former somehow?  But tossing ```this->comm().barrier();``` at the beginning of the destructor doesn't seem to help at all.

friedmud:
Exodus shouldn't be writing from all processors though... only processor 0
writes...

permcody:
I think that's the point. The other processors finish instantly so the
others get to the end very quickly. As he pointed out though. A barrier
didn't fix it.

roystgnr:
And *nothing* fixes it, but it only shows up on mpich2, and it's completely incomprehensible (all processes definitely make it past the LibMeshInit destructor, and run.sh generally dies in the middle of the message_done_running() script, between one "echo" command and the next!!!), so I'm giving up on fixing it.  I can run a hundred times in a row with openmpi with no problems, so I'll be happy merging this if everything is clean there.

permcody:
Ugh, These are the worst kind of issues. We swap back and forth between
MPICH2 and OpenMPI all the time. We recently switched back to OpenMPI
because it's at least 10x faster in our test scenario (startup overhead
dominates). This means this new "old" bug may have a significant impact on
us.

jwpeterson:
@friedmud @roystgnr let's decide whether we're cherry-picking Roy's branch here or opening a new PR on Roy's branch?  I can help out with the latter but not the former....

roystgnr:
We can open a new PR with my branch if you don't mind it being a bit conceptually cluttered.  There are commits I stuffed there not because they're really related to mesh_to_zero but because I hit red herrings while debugging.

roystgnr:
libMesh tests themselves all work fine for me now, BTW; I should be able to approve this commit tonight after some app tests finish.

jwpeterson:
> We can open a new PR with my branch if you don't mind it being a bit conceptually cluttered.

This is fine with me, although I only see the 2 extra commits on your `mesh_to_zero` branch at the moment.

friedmud:
I can cherry-pick - just let me know which ones to get.

jwpeterson:
> I can cherry-pick - just let me know which ones to get.

Let's start with just pulling @roystgnr's `mesh_to_zero` branch and pushing to yours to get this PR testing again.  It just adds two commits on top of yours should it should be a fast-forward if you want to avoid the merge commit.

roystgnr:
Okay; :+1: after you cherry-pick the fixes, which are 123c840 and 4303b73.

I'll split out the fixes-to-red-herrings into a couple separate PRs.

roystgnr:
Sorry, @jwpeterson, messages crossed in passing.  We probably *do* want to cherry-pick here, not pull my whole fork of the branch, which has some irrelevant stuff and which I had to rebase on master so the bug fixed by #1212 wouldn't trip me up on certain processor counts.

jwpeterson:
> We probably do want to cherry-pick here,

Yeah, I just fetched and noticed that you had rebased.

lindsayad:
I'm getting pretty tired of dealing with the `gather_to_zero` and leave it there behavior. I've maintained the consequences of it at the MOOSE level for a few years now and I've just about had it. MOOSE caches a bunch of additional data related to the mesh such as a node to element map. As soon as you gather to zero and keep it there, if downstream user code iterates over `node_ptr_range` and queries that node to element map, you better well have made sure to update that node to element map after the `gather_to_zero`. And then if you clear that map before rebuilding it, then you better well remember to re-add "quadrature nodes" to that map and then reinit your geometric search. And then you may do all that and then someone calls `delete_remote_elements` and now your MOOSE mesh cached data is invalid all over again on proc 0. You know who calls`delete_remote_elements`? Other mesh output classes using `MeshSerializer` who aren't just gathering to 0!!! I am down to 1 failing test on https://github.com/idaholab/moose/pull/32745 after weeks of grinding at this exodus driven debacle. And the 1 failing test is failing because I have exodus output followed by tecplot output

roystgnr:
It's an utter hack, yeah.  We ought to be doing work in parallel in ExodusII_IO, not serializing and then doing everything on proc 0.  The trouble is that we have (okay, 60% "I have") let the ExodusII_IO design bloat and bloat over the years, and it's going to be a lot more work to get it parallel-compatible now than it would have been in the first place.

PR #1215, opened 2017-01-23
Avoid printing unprintable parameters

brianmoose:
For char and unsigned char parameters, convert to int before printing to avoid displaying unprintable characters.
This is to fix idaholab/moose#7806

### DIFF
--- a/include/utils/parameters.h
+++ b/include/utils/parameters.h
@@ -541,6 +541,22 @@ void print_helper(std::ostream & os, const P * param)
   os << *param;
 }
 
+template<>
+inline
+void print_helper(std::ostream & os, const char * param)
+{
+  // Specialization so that we don't print out unprintable characters
+  os << (int)(*param);
+}
+
+template<>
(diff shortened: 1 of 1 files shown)


include/utils/parameters.h
```diff
@@ -301,6 +301,22 @@ void Parameters::Parameter<T>::print (std::ostream & os) const
   print_helper(os, static_cast<const T *>(&_value));
 }
 
+template <>
+inline
+void Parameters::Parameter<char>::print (std::ostream & os) const
+{
+  // Specialization so that we don't print out unprintable characters
+  os << (int)_value;
```
jwpeterson: Hmm... not `print_helper(os, static_cast<const int *>(&_value)`.  I think we should try and keep it consistent, just in case `print_helper` is ever updated to do something else some day...

jwpeterson:
👍  Looks good, I will probably just change it to use C++ style casts instead during merge.

PR #1217, opened 2017-01-25
Add '-Wunused -Wunused-parameter' for clang in all modes.

jwpeterson:
These need to be enabled in optimized mode to catch parameters which are
only used for debugging.  Such parameters need to be wrapped in the
libmesh_dbg_var() macro.

Note: previously we were using -Wunused in dbg and devel modes, but
that apparently does not imply -Wunused-parameter, at least in the
version of clang I'm using.  There is a fairly comprehensive list of
warnings at http://fuckingclangwarnings.com, but -Wunused is not
listed there. -Wunused *does* imply -Wunused-variable, however.

Refs idaholab/moose#8388.

### DIFF
--- a/configure
+++ b/configure
@@ -8133,14 +8133,14 @@ $as_echo "Unknown Intel complier" >&6; }
           ;;
 
       clang)
-          CXXFLAGS_OPT="$CXXFLAGS_OPT -O2 -felide-constructors -Qunused-arguments"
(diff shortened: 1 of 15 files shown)


brianmoose:
https://clang.llvm.org/docs/DiagnosticsReference.html
has `-Wunused` and what it turns on.

jwpeterson:
> https://clang.llvm.org/docs/DiagnosticsReference.html
> has -Wunused and what it turns on.

Thanks, that is a handy reference!

PR #1219, opened 2017-01-26
LibMeshInit tweaks

roystgnr:
Adding a barrier before the LibMeshInit() destructor may prevent console output from being scrambled in some use cases.

Asserting MPI return values in debug/devel modes is probably a good idea.

### DIFF
--- a/src/base/libmesh.C
+++ b/src/base/libmesh.C
@@ -687,6 +687,16 @@ LibMeshInit::LibMeshInit (int argc, const char * const * argv,
 
 LibMeshInit::~LibMeshInit()
 {
+  // Every processor had better be ready to exit at the same time.
+  // This would be a libmesh_parallel_only() function, except that
+  // libmesh_parallel_only() uses libmesh_assert() which throws an
+  // exception() which causes compilers to scream about exceptions
+  // inside destructors.
+
+  // Even if we're not doing parallel_only debugging, we don't want
+  // one processor to try to exit until all others are done working.
+  this->comm().barrier();
+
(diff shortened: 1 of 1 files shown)


src/base/libmesh.C
```diff
@@ -798,7 +816,20 @@ LibMeshInit::~LibMeshInit()
 #endif
 
       if (libmesh_initialized_mpi)
-        MPI_Finalize();
+        {
+          // We can't just libmesh_assert here because destructor,
+          // but we ought to report any errors
+          unsigned int error_code = MPI_Finalize();
```
jwpeterson: Interestingly, it looks like [some MPI implementations might even have a Barrier in Finalize](http://www.mcs.anl.gov/research/projects/mpi/mpi-standard/mpi-report-2.0/node32.htm) if they are following this implementation advice.
```
An implementation may need to delay the return from MPI_FINALIZE until all potential future message
cancellations have been processed. One possible solution is to place a barrier inside MPI_FINALIZE
```
I wonder if the different behavior you observed stemmed from something like this...

roystgnr: I honestly can't understand how *anything* sane could explain the different behavior I observed.  I mean, I guess I could write an MPI implementation that forks and nohups a trojan which then kills its parent shell a fraction of a second after exit, but how does the parent shell die in between two echo commands by *accident*?

MPI_Finalize() is only valid to call after all non-blocking communication has been resolved, so I can see how we might have a bug earlier in a program which would only trigger at the final cleanup... but what I'm seeing gets *past* the final cleanup and even past program exit.

Issue #1216, opened 2017-01-24
ReplicatedMesh::renumber_nodes_and_elements() always renumbers


jwpeterson:
nodes even when `_skip_renumber_nodes_and_elements==true`.

(This is the same Issue as #938, but this also provides a simple test case.)

The renumbering happens on [line 732 of replicated_mesh.C](https://github.com/libMesh/libmesh/blob/master/src/mesh/replicated_mesh.C#L732) which is inside a 
```
if(_skip_renumber_nodes_and_elements)
```
block... 

Here's a [mesh](https://drive.google.com/file/d/0B9BK7pg8se_iV0F5YlFlS3B4U2s/view?usp=sharing) and simple test code which triggers the issue.  It is contiguously numbered, but the node numbering does not start from 0, so when `renumber_nodes_and_elements()` gets called, the nodes get renumbered.
```
#include "libmesh/mesh.h"

using namespace libMesh;

int main (int argc, char ** argv)
{
  LibMeshInit init(argc, argv);

  Mesh mesh(init.comm());
  mesh.allow_renumbering(false);
  mesh.read("split_blocks.e");

  // Calling renumber_nodes_and_elements() will renumber nodes,
  // regardless of what the allow_renumbering() flag is set to.
  mesh.renumber_nodes_and_elements();

  mesh.write("renumbered.e");

  return 0;
}
```

I'm not sure what the right fix is.  As mentioned in #938, if we do truly skip renumbering then there are issues in other parts of the code.  On the other hand, I don't think the `renumber_nodes_and_elements()` function itself can remain as-is, since it currently produces this unexpected behavior...

lindsayad:
I think one can reasonably assert that if a user directly calls `renumber_nodes_and_elements` then we should renumber nodes and elements, or alternatively perhaps we could error if there's been a previous call to `allow_renumbering(false)`. However, probably what we can't debate is that other APIs like `contract` should *not* call `renumber_nodes_and_elements` if `_skip_renumber_nodes_and_elements` is `true` which was the case brought up in #938

roystgnr:
> error if there's been a previous call to allow_renumbering(false).

I like this idea.  Anyone who's *really* sure they want to override that setting can turn it off first.

Issue #938, opened 2016-05-19
Mesh is getting renumbered even if we asked it not to be


bwspenc:
In `UnstructuredMesh::contract()`, a call is made to `renumber_nodes_and_elements()`, regardless of whether `_skip_renumber_nodes_and_elements` is set.  That's causing the NULL elements to be removed from the element list and the elements to be renumbered.

I'm a little confused about how this is supposed to work, because in places like `MeshBase::prepare_for_use()`, there are checks to prevent `renumber_nodes_and_elements()` from getting called if the mesh is not to be renumbered. However, if you look at `SerialMesh::renumber_nodes_and_elements()`, there are some checks in there for it to behave differently if `_skip_renumber_nodes_and_elements` is set, although it still does end up renumbering the elements.

Is the design that you are just supposed to never call `renumber_nodes_and_elements()` if renumbering to be skipped? If so, the solution is to simply put a check in `UnstructuredMesh::contract()` so that the renumbering function is not called. If the intent is that `renumber_nodes_and_elements()` shouldn't actually renumber anything, then some changes need to be made to that function.

This issue is preventing us from modifying the mesh (with renumbering off) in conjunction with contact in MOOSE.  (See MOOSE pull requests [#6801](https://github.com/idaholab/moose/pull/6801) and [#6817](https://github.com/idaholab/moose/pull/6817).  Since we have two meshes (displaced and undisplaced), they get out of sync during the equation system reinit because one of them gets renumbered. That messes things up with the contact search that happens during that phase, because it depends on the two meshes being in sync.

jwpeterson:
I'll have to look into what `contract()` does in more detail.  Is it possible to do what it does without renumbering?  If not, then we would need to ensure that the `_skip_renumber...` also implies that `contract()` is skipped too.  But I get the impression that you need to `contract()` in order for XFEM stuff to work?

bwspenc:
For the XFEM case, I don't think we need anything that is done in `contract()`.  I tried skipping the call to `renumber_nodes_and_elements()` within `contract()`, and the model seems to run OK, although it crashes on a call to `EquationSystems::get_solution()` within `ExodusII_IO::write_element_data()` I'm guessing that has something to do with the NULLs in the element vector.  If I don't output anything, it runs to completion without hitting any of the issues I was having before.  Since I don't have the output, I guess I'm not really sure whether it ran correctly, though...

roystgnr:
[...]

PR #1220, opened 2017-01-26
Actually prevent renumbering when requested

jwpeterson:
This PR fixes some long-standing issues with the `MeshBase::allow_renumbering(false)` flag not actually preventing renumbering.  This state of affairs was inconsistent and annoying to analysts who tried to track the values of a single Node/Elem throughout a solution and found it changing on them.

Note that disallowing renumbering means *allowing* gaps in the mesh and element numbering, which libmesh already seemed to do fairly well.  There were just a few issues surrounding output that I fixed in this PR, but I will also keep a close eye on anything else that might be affected. 

@gardnerru, @bwspenc are interested in this.

Refs #938, #1216.

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -679,6 +679,17 @@ public:
    * when being prepared for use.  This may slightly adversely affect
    * performance during subsequent element access, particulary when
    * using a distributed mesh.
+   *
+   * Important! When allow_renumbering(false) is set,
+   * ReplicatedMesh::n_elem() and ReplicatedMesh::n_nodes() will
+   * return *wrong* values whenever adaptive refinement is followed by
+   * adaptive coarsening. (Uniform refinement followed by uniform
+   * coarsening is OK.) This is due to the fact that n_elem() and
+   * n_nodes() are currently O(1) functions that just return the size
+   * of the respective underlying vectors, and this size is wrong when
+   * the numbering includes "gaps" from nodes and elements that have
+   * been deleted. We plan to implement a caching mechanism in the
+   * near future that will fix this incorrect behavior.
    */
   void allow_renumbering(bool allow) { _skip_renumber_nodes_and_elements = !allow; }
   bool allow_renumbering() const { return !_skip_renumber_nodes_and_elements; }
--- a/src/mesh/exodusII_io.C
+++ b/src/mesh/exodusII_io.C
@@ -671,27 +671,52 @@ void ExodusII_IO::write_nodal_data (const std::string & fname,
       unsigned int variable_name_position =
         cast_int<unsigned int>(pos - output_names.begin());
 
-#ifdef LIBMESH_USE_COMPLEX_NUMBERS
-      std::vector<Real> real_parts(num_nodes);
-      std::vector<Real> imag_parts(num_nodes);
-      std::vector<Real> magnitudes(num_nodes);
+      // Set up temporary vectors to be passed to Exodus to write the
+      // nodal values for a single variable at a time.
+#ifdef LIBMESH_USE_REAL_NUMBERS
+      std::vector<Number> cur_soln;
 
-      for (unsigned int i=0; i<num_nodes; ++i)
-        {
-          real_parts[i] = soln[i*num_vars + c].real();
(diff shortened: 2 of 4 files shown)


roystgnr:
So 8edd915 fixes ReplicatedMesh::n_elem() in the uniform-coarsening-of-a-uniform-mesh case...  What happens in the adaptive-coarsening or uniform-coarsening-of-an-adaptively-refined-mesh cases?

It looks like we need to reimplement n_elem() and n_nodes() (in the ReplicatedMesh case) with something more robust?

jwpeterson:
> It looks like we need to reimplement n_elem() and n_nodes() (in the ReplicatedMesh case) with something more robust?

I thought about this, and it seems to me the only truly safe implementations of n_elem() and n_nodes() would require changing the functions' algorithmic complexity from O(1) to O(N).  Maybe in debug mode we could go with the O(N) implementation, and try to remove/avoid as many instances of `for(...; e<mesh.n_elem(); ...)` as possible?

roystgnr:
Removing n_elem() calls... it wouldn't be the first time.  What we do in DistributedMesh to keep n_elem() from being parallel_only() is we cache those values, which is a huge PITA when you can have another processor adding or removing elements without your knowledge.  So there are significant stretches in the internal AMR code where the code can't even trust that n_elem() is correct!

But I think making users remove n_elem() and n_nodes() calls too (if they want to get decent performance) is a step too far.  Can we just cache the value in ReplicatedMesh too?  Keeping the cache accurate at all times ought to be easy in the replicated case.

jwpeterson:
> Can we just cache the value in ReplicatedMesh too? Keeping the cache accurate at all times ought to be easy in the replicated case.

This sounds like a good approach to me.  Mind if I make that the subject of another PR, though?  This one fixes enough existing issues that it should be useful as-is, and there's been enough recent changes that it's time to do another libmesh update in MOOSE. I'd like this PR to be included in that update.

roystgnr:
Hmm...  I agree that the new bug is still an improvement over the old bug, I just don't want this to bite people.

At the very least we need to mention the "no adaptivity with coarsening or stuff may break in scary ways" limitation in the allow_renumbering() doxygen comment.

Is it possible to put an assert *somewhere* that will catch the problem before it hits the "n_elem() silently returns the wrong number" phase?  I'm guessing the answer is "no" since you must have had a non-contiguous element vector reach ```EquationSystems::build_parallel_solution_vector()``` to prompt those changes and you presumably don't want whatever code that was to assert-and-die instead; but it would be nice if I was wrong.

permcody:
It's like exchanging a horse fly for a deer fly. Yes, and "improved bug". :)

jwpeterson:
> At the very least we need to mention the "no adaptivity with coarsening or stuff may break in scary ways" limitation in the allow_renumbering() doxygen comment.

OK, I will push a commit to this effect.

> Is it possible to put an assert somewhere that will catch the problem before it hits the "n_elem() silently returns the wrong number" phase? I'm guessing the answer is "no" since you must have had a non-contiguous element vector reach EquationSystems::build_parallel_solution_vector() to prompt those changes and you presumably don't want whatever code that was to assert-and-die instead; but it would be nice if I was wrong.

We could definitely add an O(N) debug mode only assert to the `n_elem()` and `n_nodes()` calls themselves as a temporary fix until the caching is in place.  

I don't know how many apps this is really going to break though...  MOOSE and MOOSE apps seem to be the primary users of `MeshBase::allow_renumbering(false);` and I actually discovered that we are actually *allowing* renumbering in most cases at the moment due to the logic [on this line of code](https://github.com/idaholab/moose/blob/devel/framework/src/mesh/FileMesh.C#L88).

roystgnr:
:+1: as is - changing n_elem() to O(N), even in debug mode only, is a bit much for me.

jwpeterson:
Thanks, I'm just investigating one test failure in Marmot over on idaholab/moose#8418 that seems like it could be related to this change.  Once I get that straightened out I'll merge this.

jwpeterson:
Allowing non-contiguous node and elem ranges breaks the XDA writer currently... and this will take some extra thought to fix.  Temporarily closing the PR while I work on that.

PR #1222, opened 2017-01-30
fix n_levels for multi-processor, serial, multi-level adaptivity cases

rwcarlsen:
Previously, ``MooseUtils::n_levels`` assumed that all processors had their own local elements for calculating the number of mesh adaptivity levels.  This assumption is violated in the following scenario (i.e. when all the following are true):

* performing mesh initialization
* using replicated mesh
* running on multiple processors
* initial mesh has multiple adaptivity levels already (e.g. via ``CheckpointIO``).
* initial mesh has elements partitioned onto multiple processor already

CheckpointIO, for example, skips initialization of the mesh from a file for all processors except processor 0.  This previously resulted in the n_levels function only looping over elements local to a processor to determine the level and unpartitioned elements.  The rendezvous ``max`` call at the end of ``n_levels`` doesn't help because the non processor 0 processes all have zero element meshes.  This was discovered when investigating idaholab/moose#8410

### DIFF
--- a/src/mesh/mesh_tools.C
+++ b/src/mesh/mesh_tools.C
@@ -638,8 +638,14 @@ unsigned int MeshTools::n_levels(const MeshBase & mesh)
 
   MeshBase::const_element_iterator el =
     mesh.unpartitioned_elements_begin();
-  const MeshBase::const_element_iterator end_el =
+  MeshBase::const_element_iterator end_el =
     mesh.unpartitioned_elements_end();
+
+  if (mesh.is_serial())
+  {
+    el = mesh.elements_begin();
+    end_el = mesh.elements_end();
+  }
 
   for( ; el != end_el; ++el)
     nl = std::max((*el)->level() + 1, nl);


roystgnr:
Is this the right fix?

It replaces the current assumption of "all processors have all their local elements", which makes intuitive sense, with a new assumption of "all processors have all their local elements, except when the mesh is flagged as serial", which is intuitive nonsense, since serial is supposed to mean every processor has *more* elements than it may need, never *less*.

This also turns n_levels() from O(N/N_proc) into plain O(N) on ReplicatedMesh, even though the O(N/N_proc) algorithm should work except in a few degenerate cases.

But there's got to be some fix... our above assumption is currently violated, and it looks like there's a chicken-and-egg obstacle to fixing that violation, isn't there.  We use n_levels() in MeshCommunication::broadcast() and until CheckpointIO uses broadcast() to serialize a newly read ReplicatedMesh, the ReplicatedMesh isn't truly replicated, and the fast n_levels() implementation won't work.

What I've done in similar cases, where a usually-correct fast method would fail for some internal uses, is just create a different method.  See MeshBase::n_elem() vs MeshBase::parallel_n_elem().  We probably ought to do the same here.  MeshTools::n_levels() with the current algorithm for most uses, MeshTools::i_cant_think_of_a_good_name() for users like MeshCommunication that can't assume a properly broadcast mesh?

rwcarlsen:
Yes - it's not a particularly "clean" fix.  I just wasn't sure what sort of approach would be preferred - so I picked the easiest one to start.  Another solution is to have CheckpointIO read in the mesh file on all processors for replicated mesh - then the assumption that each processor has its own elements/data will be true in the call to n_levels.  However, you would need to be confident that there were no other callers of n_levels that violate that assumption.

permcody:
Tagging in @friedmud here too.

roystgnr:
I've got a better fix ready, I think; will open a new PR in a minute

rwcarlsen:
Replaced by #1223.

PR #1223, opened 2017-01-31
paranoid_n_levels()

roystgnr:
This should replace #1222 

We retain the old n_levels() algorithm, but in dbg mode (where we're already throwing away even asymptotic complexity guarantees) we test the result against the slower paranoid_n_levels() before returning.

I checked through everything in libMesh and MOOSE, and MeshCommunication::broadcast() appears to be the only place where we should ever expect to need paranoid_n_levels().

### DIFF
--- a/include/mesh/mesh_tools.h
+++ b/include/mesh/mesh_tools.h
@@ -242,8 +242,9 @@ dof_id_type n_non_subactive_elem_of_type_at_level(const MeshBase & mesh,
 
 /**
  * Return the number of levels of refinement in the mesh.
- * Implemented by looping over all the local elements and finding the
- * maximum level, then summing in parallel.
+ * Implemented by looping over all the local elements and
+ * unpartitioned elements and finding the maximum level, then summing
+ * in parallel.
  */
 unsigned int n_levels(const MeshBase & mesh);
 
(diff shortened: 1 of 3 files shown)


jwpeterson:
```
../src/mesh/mesh_tools.C: In function ‘unsigned int libMesh::MeshTools::paranoid_n_levels(const libMesh::MeshBase&)’:
../src/mesh/mesh_tools.C:671:5: error: ‘nl’ was not declared in this scope
     nl = std::max((*el)->level() + 1, nl);
     ^
../src/mesh/mesh_tools.C:673:19: error: ‘nl’ was not declared in this scope
   mesh.comm().max(nl);
```

roystgnr:
See, that's what you get for making me turn off my loud-fan-cooled recompile.

jwpeterson:
:rofl:

rwcarlsen:
LGTM. Not that you were waiting for my approval :)

roystgnr:
Are those test failures on master that we're just now adding to CI, or new test failures?

permcody:
I enabled it "on demand" we can turn on additional targets that we want to test for PR. We test recover in MOOSE for every PR but not for libMesh. Since this PR arose from problems with recover, I thought I'd turn that target on here. I don't know if these are definitely related to this PR or not without diving in.

permcody:
Ugh... I just looked and I don't see that we do recover(dbg) with MOOSE PRs. We do debug and recover but not together. Is that correct @brianmoose? That means these are likely existing failures.

brianmoose:
That is correct. This has come up before and I was told to not enable recover in debug mode until the tests were fixed.
I can disable it on the libmesh "Test Moose Recover" recipe if you want.

permcody:
That's OK, Just don't want Roy to think that these are his issues.

roystgnr:
Oh, these are probably Roy's issues.  They're just not PR 1223 issues.  :-P

jessecarterMOOSE:
Just poking my head in here. I don't see the specific assert that I was running into, so that appears to be fixed. I believe the rest of those errors have to do with parallel meshes.

rwcarlsen:
I plan on going through the parallel recovery issues in the coming days and will make as much progress as I can.  I may find more issues with libMesh - but I anticipate at least as many in MOOSE.

jessecarterMOOSE:
Progress! :+1:

roystgnr:
It looks like the Test Moose Recover issues are unrelated and this is a fix for at least one issue, so merging now.  I'll see if I can hack away a few of those in a later PR.

PR #1224, opened 2017-01-31
Send coarse ghosts

roystgnr:
This fixes the last remaining bug I know about with DistributedMesh AMR/C: when coarsening, it was possible for a newly-activated coarse element to not have all the ghost elements it needed, and nothing would communicate those ghost elements to it.  This was causing even one of the libMesh examples to fail on 24 processors for me, when a coarsened element happened to straddle a partition boundary.


roystgnr:
It turns out that this isn't quite the last bug, but it does appear to be one of the last bugs, and it is fixed by this patch.

PR #1225, opened 2017-01-31
MeshTools::n*levels was broken on empty meshes

roystgnr:
Some methods would claim an empty mesh had 0 levels; others would
claim 1 level; n_levels() itself would scream and die in debug mode
when it noticed the inconsistency.

This fixes these methods and makes the return value sane: an empty
mesh has 0 levels.


roystgnr:
Going to merge this right away since we have some empty meshes in the unit tests which hit that debug mode assertion.

PR #1226, opened 2017-02-01
Avoid vtkSmartPointer include

roystgnr:
My old VTK version uses the deprecated strstream header, and
ignore_warnings.h isn't enough to stop gcc from complaining about
that.

We can't use UniquePtr instead, because VTK objects have protected
destructors, but we can do cleanup manually.

This doesn't completely get rid of the gcc screaming, but it limits it to libmesh.C rather than every_single_source_file.C

I'm following the VTK examples for manual cleanup, but I'd appreciate @jwpeterson taking a look to make sure I haven't just added a memory leak or somehow regressed #1179

### DIFF
--- a/include/base/libmesh.h
+++ b/include/base/libmesh.h
@@ -33,9 +33,6 @@
 
 // For dealing with MPI stuff in VTK.
 #if defined(LIBMESH_HAVE_MPI) && defined(LIBMESH_HAVE_VTK)
-# include "libmesh/ignore_warnings.h"
-# include "vtkSmartPointer.h"
-# include "libmesh/restore_warnings.h"
 class vtkMPIController;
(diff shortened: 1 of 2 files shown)


jwpeterson:
The fix for this *used* to be to `#define VTK_LEGACY_REMOVE` but this had to be removed in 4636ea583c8ffd39490c165eee1097e5f7ec6741 since it broke something else in VTK-7.1.  I suspect we may have been using that define in an incorrect manner and that there might be a "right" way to do it.

Your manual memory management solution is better anyway, in hindsight avoiding includes in libmesh.h is definitely the right approach.  I'm not going to pull this until after the Tiger Team, though, so I don't have to recompile the universe 😉

PR #1227, opened 2017-02-02
DistributedMesh fixes

roystgnr:
We now handle cases where processors need multiple communication sweeps to get node ids correct, cases where coarsening away thin partitions was failing to correctly communicate newly ghosted elements to the other partitions on each side, and cases where I can't correctly write a receive_packed_range call without noticing that I've cut-and-pasted blatantly incorrect code.

This is passing basic parallel tests for me, and should be ready to merge.

I can no longer find any outstanding DistributedMesh bugs, but I'm still sweeping through test suites on varying processor counts to search.

### DIFF
--- a/include/parallel/parallel_ghost_sync.h
+++ b/include/parallel/parallel_ghost_sync.h
@@ -124,8 +124,9 @@ void sync_element_data_by_parent_id(MeshBase &       mesh,
  *                  std::vector<sync::datum> & data),
  * by resizing and setting the values of the data vector.
  * Respond to fulfillment with
- * sync.act_on_data(const std::vector<unsigned int> & ids,
- *                  std::vector<sync::datum> & data)
+ * bool sync.act_on_data(const std::vector<unsigned int> & ids,
+ *                       std::vector<sync::datum> & data)
+ * and return true iff the response changed any data.
  * The user must define Parallel::StandardType<sync::datum> if
  * sync::datum isn't a built-in type.
  */
@@ -135,8 +136,8 @@ template <typename ElemCheckFunctor,
 void sync_node_data_by_element_id(MeshBase & mesh,
                                   const MeshBase::const_element_iterator & range_begin,
                                   const MeshBase::const_element_iterator & range_end,
-                                  ElemCheckFunctor & elem_check,
-                                  NodeCheckFunctor & node_check,
+                                  const ElemCheckFunctor & elem_check,
+                                  const NodeCheckFunctor & node_check,
                                   SyncFunctor & sync);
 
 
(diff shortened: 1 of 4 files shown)


pbauman:
> I can no longer find any outstanding DistributedMesh bugs

Epic.

roystgnr:
Well, DistributedMesh was totally solid at this time last year, too, so there wasn't too much to fix.  The current swaths of bug fixes have all been to fix DistributedMesh-without-ghosted-point-neighbors bugs, which weren't introduced until last summer when I realized just how much ridiculous over-ghosting we get if we try to retain all point neighbors on hierarchically refined grids.

pbauman:
For some reason, I thought there was some longstanding bug you'd just squashed. Still happy  `DistributedMesh` is in good shape again, though.

roystgnr:
I think "since August" is pretty long-standing for a bug that I could replicate as simply as "make check" on 24 processors.

In hindsight maybe I should have left point neighbors ghosting turned on until we had things working with it off.

On the other hand, it was very motivating to have @friedmud raving to everyone (with interactive "look at the results in top!" terminal window even) about how DistributedMesh has been cutting his RAM usage by 95%, and how everyone needs to switch to it ASAP...  If at that point the DistributedMesh landmines were buried by workarounds then I might have been too lazy to dig in and defuse them.

permcody:
In other news, people searching the list for years to come will enjoy
reading remarks from Derek and unfortunately myself that you both "never"
need to use ParallelMesh and that you should simultaneously should "always"
use it for large runs...  Sigh...

jwpeterson:
On Thu, Feb 2, 2017 at 8:00 AM, Cody Permann <notifications@github.com>
wrote:

> In other news, people searching the list for years to come will enjoy
> reading remarks from Derek and unfortunately myself that you both "never"
> need to use ParallelMesh and that you should simultaneously should "always"
> use it for large runs... Sigh...


We've always been at war with DistributedMesh.

-- 
John

roystgnr:
It'll take a while until my ```for num in $(seq 1 24); do LIBMESH_RUN="mpirun -j $num" ...``` tests are done, but we're past the point where I was able to trigger errors previously, so I'm tempted to merge this ASAP.

I'll run it through GRINS on 12 processors too and if nobody's screamed by the time that finishes, I'll merge.

roystgnr:
I can still trigger an assertion failure (inconsistent refinement flags) when using AdjointRefinementErrorEstimator on 7 or more processors.  It shouldn't affect opt mode runs, though (we always sync refinement flags just in case the user flagged things inconsistently; we just scream about it in devel/debug modes), and I'm not yet sure whether the problem is a bug we can fix or whether this is just an overzealous assertion we need to remove.

Every test I've thrown passes in opt mode, and the inconsistent flagging assertion is the only thing I've managed to trip in debug mode.

PR #1229, opened 2017-02-06
Use std::size_t consistently

jwpeterson:
We initially used `unsigned int` when looping to `std::vector::size()` but technically the type of `size()` is `std::size_t`.  Newer code seems to have used this convention in a few places already, this just gets most of the remaining places...

### DIFF
--- a/examples/adaptivity/adaptivity_ex1/adaptivity_ex1.C
+++ b/examples/adaptivity/adaptivity_ex1/adaptivity_ex1.C
@@ -274,11 +274,11 @@ void assemble_1D(EquationSystems & es,
         {
           // Now build the element matrix and right-hand-side using loops to
           // integrate the test functions (i) against the trial functions (j).
-          for (unsigned int i=0; i<phi.size(); i++)
+          for (std::size_t i=0; i<phi.size(); i++)
             {
               Fe(i) += JxW[qp]*phi[i][qp];
 
-              for (unsigned int j=0; j<phi.size(); j++)
+              for (std::size_t j=0; j<phi.size(); j++)
                 {
                   Ke(i,j) += JxW[qp]*(1.e-3*dphi[i][qp]*dphi[j][qp] +
                                       phi[i][qp]*phi[j][qp]);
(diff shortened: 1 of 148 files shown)


roystgnr:
Looks good, and I see no obvious mistakes, but I reserve the right to back off to std::ptrdiff_t in the future, when I get around to optimizations, if it turns out that the "unsigned integers don't disallow overflow and so the compiler has to disable some loop optimizations and vectorization" advice I've read in a few places is correct.

9-sextillion-entry vectors ought to be enough for anybody.

I think that's as far as I'd want to go though.  It's tempting to go through and say "We'll never have more than 2^15 systems, or more than 2^31 shape functions on an element" etc. and then use int_fast16_t etc. as appropriate, but as far as I can tell none of the architectures we use would benefit and so the int_fastxx_t types would be pointless error-prone hassle.

jwpeterson:
> Looks good, and I see no obvious mistakes, but I reserve the right to back off to std::ptrdiff_t in the future, when I get around to optimizations, 

std::ptrdiff_t is a signed integer type, so wouldn't you get a warning about comparing with the unsigned `vector::size_type` returned by `size()`?

> if it turns out that the "unsigned integers don't disallow overflow and so the compiler has to disable some loop optimizations and vectorization" advice I've read in a few places is correct

Oh, I guess you mean you would cast `size()` to `std::ptrdiff_t` to coerce the compiler into optimizing?  I thought the  issue preventing optimization was related to the mismatch in types in the for loop termination condition (unsigned comparison with size_t).

roystgnr:
Right; we'd have a temporary size variable.

But yeah, using unsigned integers isn't the only way to prevent the compiler from proving that your for loop isn't *really* sometimes supposed to be an infinite loop.  Comparing a signed integer to a signed end size would also do it.

Range based for loops would also fix everything, but we still have downstream users who believe they really want C++03 compliance.  I'll try again to convince them otherwise soon.  I'm tired of hearing about custom high performance compilers; at this point gcc is almost certainly a better optimizer than anything 5 years old so if you have something 5 years old you need to use it to compile gcc and then delete it.

PR #1233, opened 2017-02-08
EquationSystems::reinit preserve refinement flags

tradowsk:
I mentioned a few months ago on `libmesh-devel` that if an app (such as `GRINS`) calls `refine_and_coarsen_elements()` and then `es->reinit()`, the `JUST_REFINED` and `JUST_COARSENED` element flags get cleared out by the calls to `coarsen_elements()` and `refine_elements()` within `EquationSystems::reinit()`. My `GRINS::RayfireMesh` class requires that these flags be set properly post-reinit.

This PR adds a flag to `EquationSystems` that, when set to `false`, skips over these refine/coarsen calls, preserving the refinement flags. By default in the ES constructor, this flag is set to `true`, maintaining backward compatibility.

A `CPPUNIT` test is included to verify that the refinement flags for a simple 2-element mesh are correct post-reinit. Also tested this with a `GRINS` QoI AMR run that originally exposed this issue.

`make check` passes for both libMesh and `GRINS`. I welcome any comments on the variable/function naming or any other suggestions. In particular, I am unsure if the call to mesh `contract` (line 262) should be inside or outside the guarding `if` statement (line 234).

### DIFF
--- a/include/systems/equation_systems.h
+++ b/include/systems/equation_systems.h
@@ -454,6 +454,24 @@ public:
    **/
   void allgather ();
 
+  /**
+   * Calls to reinit() will also do two-step coarsen-then-refine
+   **/
+  void enable_refine_in_reinit()
+    { this->_refine_in_reinit = true; }
+
+  /**
+   * Calls to reinit() will not try to coarsen or refine the mesh
+   **/
+  void disable_refine_in_reinit()
+    { this->_refine_in_reinit = false; }
+
+  /**
+   * @returns whether or not calls to reinit() will try to coarsen/refine the mesh
+   **/
+  bool refine_in_reinit_flag()
+    { return this->_refine_in_reinit; }
+
 
   /**
    * Data structure holding arbitrary parameters.
@@ -489,6 +507,12 @@ protected:
    * Typedef for constatnt system iterators
    */
   typedef std::map<std::string, System *>::const_iterator const_system_iterator;
+
+  /**
+   * Flag for whether to call coarsen/refine in reinit().
+   * Default value: true
+   */
+  bool _refine_in_reinit;
 
 private:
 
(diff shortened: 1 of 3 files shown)


include/systems/equation_systems.h
```diff
@@ -454,6 +454,24 @@ class EquationSystems : public ReferenceCountedObject<EquationSystems>,
    **/
   void allgather ();
 
+  /**
+   * Calls to reinit() will also do two-step coarsen-then-refine
+   **/
+  void enable_refine_in_reinit()
```
pbauman: @roystgnr and @jwpeterson will have the final word, but I would've preferred just a traditional `set/get_refine_in_reinit_flag()` accessor pattern myself.

pbauman:
@jwpeterson, @tradowsk is one of my PhD students, so these tests should be safe to authorize.

jwpeterson:
> these tests should be safe to authorize.

Whoops, forgot we had turned that check back on again, thanks for the ping.

include/systems/equation_systems.h
roystgnr: Yeah; it might be useful to temporarily disable; with just an enable() method once you've enabled you're committed.

jwpeterson:
I also sent @tradowsk an invite to join a GitHub group that gets their PRs tested automatically.

pbauman:
> I also sent @tradowsk an invite to join a GitHub group that gets their PRs tested automatically.

While you're at it, could you send one to @borisboutkov who is another one of my PhD students. He'll have some libMesh PRs in the (hopefully) not too distant future.

jwpeterson:
> While you're at it, could you send one to @borisboutkov

Done!

include/systems/equation_systems.h
roystgnr: Oh, I'm an idiot.  I was looking at the main page, not the whole diff, and didn't see the disable_ function.

Okay, IMHO this is fine as is.

roystgnr:
In fact, this whole PR is shorter than I'd feared and looks like a reasonable workaround.  :+1: when tests are done unless anyone else objects.

pbauman:
Since @roystgnr signed off on this, I'm going to go ahead and merge.

roystgnr:
Gah, I should have looked at the unit test more closely.  This breaks with --enable-parmesh on 3 processors.  I'll have a fix shortly.

PR #1234, opened 2017-02-08
BoundaryInfo::operator= fix

roystgnr:
This fixes a couple errors I managed to trigger while working on https://github.com/idaholab/moose/issues/1500


roystgnr:
Going to merge this right away, since it's a clear bugfix in some cases and it changes the behavior of BoundaryInfo::operator= to be more intuitive in other cases.

It's theoretically possible that some apps are actually *depending* on the current behavior, but if we discover any such case then let's put it in an appropriately named function: "merge" or "append".

PR #1238, opened 2017-02-13
Small optimization in TreeNode

friedmud:
This little optimization gives ~10% speedup in heavy Tree searching activity.

And, yes, the `if` statement is actually necessary!  Without it... just the `resize()` call takes ~2% of runtime.  It really shouldn't since most of the time each TreeNode is going to have the same number of children so it should be a no-op.

My only guess is that once again `std::vector<bool>` is rearing its ugly head... maybe it doesn't have an optimized version of `resize()` for the case where the size is staying the same since it's an odd duck.  (pinging @permcody so he can weigh in on that 😎 )

This falls in the category of: Found while idly running Instruments while waiting on something to solve 😄

### DIFF
--- a/src/utils/tree_node.C
+++ b/src/utils/tree_node.C
@@ -528,7 +528,10 @@ const Elem * TreeNode<N>::find_element_in_children (const Point & p,
 {
   libmesh_assert (!this->active());
 
-  std::vector<bool> searched_child(children.size(), false);
+  if (searched_child.size() != children.size())
+    searched_child.resize(children.size(), false);
+
+  std::fill(searched_child.begin(), searched_child.end(), false);
 
   // First only look in the children whose bounding box
   // contain the point p.
--- a/include/utils/tree_node.h
+++ b/include/utils/tree_node.h
@@ -221,6 +221,11 @@ private:
    * Does this node contain any infinite elements.
    */
   bool contains_ifems;
+
+  /**
+   * Used in find_element_in_children
+   */
+  mutable std::vector<bool> searched_child;
 };
 
 


src/utils/tree_node.C
```diff
@@ -528,7 +528,10 @@ const Elem * TreeNode<N>::find_element_in_children (const Point & p,
 {
   libmesh_assert (!this->active());
 
-  std::vector<bool> searched_child(children.size(), false);
+  if (searched_child.size() != children.size())
+    searched_child.resize(children.size(), false);
+
+  std::fill(searched_child.begin(), searched_child.end(), false);
```
jwpeterson: Looks fine to me, although I think I will just change this bit to 1 line:
```
searched_child.assign(children.size(), false);
```

permcody:
Hmm, it's hard to say, the resize implementations are close but they call different methods for inserting which I would hope would be about the same for the no-op case.

Here's the implementation in the GCC 6.2 includes:
std::vector<>
```
      void
      resize(size_type __new_size)
      {
	if (__new_size > size())
	  _M_default_append(__new_size - size());
	else if (__new_size < size())
	  _M_erase_at_end(this->_M_impl._M_start + __new_size);
      }
```

std::vector<bool>
```
    void
    resize(size_type __new_size, bool __x = bool())
    {
      if (__new_size < size())
        _M_erase_at_end(begin() + difference_type(__new_size));
      else
        insert(end(), __new_size - size(), __x);
    }
```

roystgnr:
@friedmud, would you try just switching to a local std::vector<unsigned char> instead and see how the performance compares?  I really dislike mutable members; in this case in particular I wonder if you may have just added a race condition to threaded tree searches.

We should never have been using vector<bool> there; I'm not as anti-bool-shit as everyone else, but I'm at least willing to admit that it's only a good idea in cases where you *really* want to avoid memory use (or occasionally cache misses) at the cost of CPU bit fiddling, and that means tens of millions of entries, not 8 or less.

Damn, that "8 or less" also makes me wonder: would it make sense to have some kind of static libMesh::max_n_children=8 and use that to size an array rather than a vector here?  I assume your optimization works because it allows the compiler to stop hitting the heap for every node we pass through in a search, and an array would do that just as well.

roystgnr:
Wait, this is descending through tree children, not element children.  So our "max_n_children" here is just the template parameter N?

permcody:
If that's the case, you could use a local array here which gets rid of the
mutable member. Once libmesh goes to C++11, that can then be switched to a
std::array<N, ...> which is stack allocated.

friedmud:
I dislike mutable too... let me try it with the stack allocated local array.

PR #1235, opened 2017-02-09
Ignore Eigen warnings

roystgnr:
There are multiple warning sources in our contrib Eigen headers with
gcc 6.1, and dense_vector.h now imports those headers practically
*everywhere*.

PR #1239, opened 2017-02-13
Ignore -Wmisleading-indentation on GCC 6+ only

roystgnr:
GCC 5 and below heard we like ignoring warnings so they warn us about
an unrecognized warning and we had to ignore a warning in the
ignore_warnings header.

Hopefully this PR fixes the gcc 4.9 spew @BalticPinguin discovered after #1235


jwpeterson:
😆  👍

jwpeterson:
This fixes the issue for me on GCC 5.2 so I'm merging it.

PR #1240, opened 2017-02-14
Only use out_of_mesh_value in out_of_mesh_mode

roystgnr:
If we weren't expecting to be searching outside the mesh then if we
can't find a mesh element it's an error.

This makes some errors in MeshFunction use much easier to debug.

However, this assert actually triggers for me in reduced_basis_ex4 for me when I run in parallel!  We've already manually set that example to use ReplicatedMesh, and it works on one processor, so I'm not sure what could be going wrong.

### DIFF
--- a/src/reduced_basis/rb_eim_construction.C
+++ b/src/reduced_basis/rb_eim_construction.C
@@ -217,6 +217,12 @@ Real RBEIMConstruction::train_reduced_basis(const bool resize_rb_eval_data)
 Number RBEIMConstruction::evaluate_mesh_function(unsigned int var_number,
                                                  Point p)
 {
+  // Set default values to be an empty vector so that we can use it
+  // below
+  // to check if this processor returned valid data.
+  DenseVector<Number> default_values;
+  _mesh_function->enable_out_of_mesh_mode(default_values);
+
   _mesh_function->set_point_locator_tolerance( get_point_locator_tol() );
 
   DenseVector<Number> values;
--- a/src/mesh/mesh_function.C
+++ b/src/mesh/mesh_function.C
@@ -257,6 +257,9 @@ void MeshFunction::operator() (const Point & p,
 
   if (!element)
     {
+      // We'd better be in out_of_mesh_mode if we couldn't find an
+      // element in the mesh
+      libmesh_assert (_out_of_mesh_mode);
       output = _out_of_mesh_value;
     }
   else
--- a/src/utils/point_locator_tree.C
+++ b/src/utils/point_locator_tree.C
@@ -233,11 +233,7 @@ const Elem * PointLocatorTree::operator() (const Point & p,
           // out-of-mesh mode this is sometimes expected, and we can
           // just return NULL without searching further.  Out of
           // out-of-mesh mode, something must have gone wrong.
-
-          // We'll avoid making this assertion just yet, though,
-          // because some users are leaving out_of_mesh_mode disabled
-          // as a workaround for old PointLocatorTree behavior.
-          // libmesh_assert_equal_to (_out_of_mesh_mode, true);
+          libmesh_assert_equal_to (_out_of_mesh_mode, true);
 
           return this->_element;
         }


roystgnr:
And apparently there's a MOOSE test that I've also just converted from "silently and unexpectedly uses out_of_mesh_value" to "throws an assertion".  And this one in serial!  I'm pretty sure the new assertion isn't overzealous, so there's definitely something to look into here.

permcody:
No surprise there! We have plenty of bad tests

roystgnr:
You might have some bad tests, and it's theoretically possible that this is one of them (trying to project between two non-overlapping grids and depending on out_of_mesh_mode without enabling it, perhaps?), but so far the "investigate MOOSE+DistributedMesh" work has led to 5 libMesh bug fixes, 4 MOOSE framework bug fixes, and 0 MOOSE test bug fixes, so I'm not putting my faith in the latter.

permcody:
I just looked into this, and it's a test that's actually verifying that we
get some sort of error message if we evaluate out of mesh. In particular,
it's hitting this branch:
https://github.com/idaholab/moose/blob/devel/framework/src/userobject/SolutionUserObject.C#L923-L930

dknez:
Good catch regarding reduced_basis_ex4. The issue here is in RBEIMConstruction::evaluate_mesh_function. There we expect the mesh function on all but one processor  to fail, and we handle the broadcast ourselves.

One way to fix this is by inserting the code below at the top of RBEIMConstruction::evaluate_mesh_function:

```
  // Set default values to be an empty vector so that we can use it below
  // to check if this processor returned valid data.
  DenseVector<Number> default_values;
  _mesh_function->enable_out_of_mesh_mode(default_values);
```

friedmud:
Since we're talking about `out_of_mesh_mode`... let me bring up something.

I _really_ need to use `out_of_mesh_mode`... and it works fine in `opt`... but if I try to use it in `dbg` it tells me that it's only meant to be used for affine elements.

I can't really see why that's true... at the end of the day both ways of searching (tree and linear) use the same `contains_point()` call.  Is it just because of the bounding boxes used in the tree mode?  I think I fixed it a few years ago so those bounding boxes are "fuzzy"... certainly fuzzy enough for most non-affine elements.

Anything else you can think of?

roystgnr:
Yeah, the bounding box is the catch.  When we do a contains_point() then we're examining a location in master element space, so any nonlinear geometry mapping has already been inverted and our results are reliable... but this is slow so we don't want to do it if we can avoid it.

With the tree, we avoid that inverse_map because we don't even bother to check elements if our desired point is not in the node's bounding box.  TreeNode::insert(const Elem*) is the offender here.  We don't actually have an Elem::bounding_box() function, so we just check to see whether any of the Elem's nodes fall within the TreeNode's box.  This is enough for affine elements (and for first order elements, I believe - that libmesh_error_msg appears to be overzealous!) but for quadratic non-affine elements it's not too hard to cook up curved side examples where the "bulge" extends outside the bounding box of the nodes.

If we're not in out_of_mesh_mode then we know the user expects to get an element back, so we must be in one of those crazy "bulge" situations, so we cry and we do a linear search until we find an element.  If we are in out_of_mesh mode then the user probably expects to have lots of points that don't fall within elements, so doing a linear search for each one would be a big waste of CPU time, so that's not an option, but we still don't actually want to return a wrong answer, so we error out if we're in a situation where a wrong answer might be possible.

I'm not sure how we can improve the situation, unless someone knows of a reliable bounding box calculation for higher order elements.  I couldn't find anything the last time I looked.  I suppose it's just a constrained optimization problem so we could solve it with that tool

roystgnr:
We're going to need to throw this whole branch away if there's no solution to the "out of mesh mode can't work efficiently on quadratics" problem.

In 1D we don't have to worry about quadratic elements, because if the element extends past either of it's vertices then it's not invertable and we don't support it anyway.

On flat 2D elements we can use the same reasoning to conclude that any bounding box which contains all the element's edges must contain the whole element.  For quadratic potentially-curved edges I can show that

x_max <= x_mid + max(abs(x_left - x_mid), abs(x_right - x_mid))

and so on, which makes loose bounding boxes (which collapse to tight bounding boxes in the linear case!) cheap to compute.

In 3D you can conclude that any bounding box which contains all the element's faces must contain the whole element, but since we can't assume the faces are flat we can't reuse the flat-2D result.  The generalized result *seems* to still hold: fiddling around, the value of a biquadratic on [-1,1]^2 never exceeds the center value plus the difference between the center value and another node's value.  I can't seem to prove this, though!

friedmud:
@roystgnr I would be happy with just "fuzzy" bounding boxes for now... with a user specified "fuzz" amount.  It's rare that we generate curvilinear elements which are extremely far away from their linear counterparts.

roystgnr:
If it doesn't make anyone else cringe, I'm just going to implement using the ideas above, which won't add any fuzz (which makes the tree search less efficient) for first order elements and which seems to be a decently safe overestimate of the worst case scenarios for higher order elements.

If it *does* make anyone cringe, then I'll see if I can prove that bound for 2D quadratics and biquadratics.  Doing so doesn't seem more difficult than the 1D quadratic case, just much more tedious.

friedmud:
Sounds good to me... I only need fast first-order currently.

Derek

roystgnr:
This ought to speed up the "we want to use out of mesh mode but we can't say so because we have non-affine elements" case quite a bit.  Right now every time you actually *hit* out of mesh, you end up trigging a whole linear search just in case there was a curved element extending into a point locator tree node that didn't see it.

Do you guys use the close_to_point_tol() option much?  We could probably eliminate the linear search from that use case too, but if you don't need it or have test coverage of it then I don't want to bother with or risk futzing with it.

friedmud:
I don't think we use it... but that doesn't mean that we shouldn't!

Derek

roystgnr:
Oh, I know *someone* is using it; @dknez at least, I believe.  Hence my reluctance to futz with it and risk breaking it until we have test coverage.

roystgnr:
And done.  In theory this will still trigger the MOOSE test error because of MOOSE relying on but not turning on out_of_mesh_mode.  Right now I'm just abusing Civet to hit this series of edits with more testing to make sure nothing *else* has regressed.  If it hasn't, then we can merge the "fix out_of_mesh_mode with non-affine elements" stuff first (which I'll put in as a separate PR), then I'll make a PR to MOOSE to enable out_of_mesh_mode (and update libMesh to the required version) there, then we can merge the assertion here last.

friedmud:
Sounds good - thanks Roy!

Derek

dknez:
I do use close_to_point_tol() a lot. That only applies when the search for an element fails, since we then "fall back" on a linear search that uses the close-to-point tolerance.

But it looks like this PR is a uniform improvement, and we still fall back on the linear search with close-to-point-tol when requested. So this looks good to me.

dknez:
This might pass CIVET now, after your recent PRs, right @roystgnr ?

src/reduced_basis/rb_eim_construction.C
```diff
@@ -217,6 +217,12 @@ Real RBEIMConstruction::train_reduced_basis(const bool resize_rb_eval_data)
 Number RBEIMConstruction::evaluate_mesh_function(unsigned int var_number,
                                                  Point p)
 {
+  // Set default values to be an empty vector so that we can use it
+  // below
```
dknez: Formatting of this comment is a little weird, might want to move "here" to the previous line

roystgnr:
It ought to pass only after https://github.com/idaholab/moose/pull/8626 is merged into MOOSE, but I have no idea *what* is going on there.  Segmentation faults that occur only in debug mode?  A swath of timeouts?

PR #1241, opened 2017-02-14
Serialized DistributedMesh fixes

roystgnr:
We now have the capability to keep a DistributedMesh serialized, and exercising that capability revealed a few bugs.

In particular, we can't use "is_serial()" to tell whether a Mesh is a DistributedMesh, and I don't like the RTTI solution, so I added an is_replicated() method.

### DIFF
--- a/include/mesh/distributed_mesh.h
+++ b/include/mesh/distributed_mesh.h
@@ -119,6 +119,13 @@ public:
   { return _is_serial; }
 
   /**
+   * @returns \p true if new elements and nodes can and should be
+   * created in synchronization on all processors, \p false otherwise
+   */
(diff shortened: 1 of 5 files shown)


friedmud:
Still no RTTI, eh?  After all these years... ;-)

roystgnr:
Ha!  That is only about 79.99% because I dislike dynamic_cast, 0.01% because I imagine someone's going to need -fno-rtti working soon.  It's 20% because I still fantasize about introducing cartesian, block-based, and other MeshBase subclasses someday, and I imagine that there might be some options (e.g. the block-based) which would be a distributed mesh without being a DistributedMesh.

PR #1242, opened 2017-02-14
Allow users to control if the matrix and RHS should be zeroed out

### LINKED ISSUE #1231: Add flag in ImplicitSystem to avoid zeroing matrix/rhs during assemble
`ImplicitSystem::assemble()` currently zeros the matrix and rhs every time it's called:
```
void ImplicitSystem::assemble ()
{
  libmesh_assert(matrix);
  libmesh_assert (matrix->initialized());
  libmesh_assert(rhs);
  libmesh_assert (rhs->initialized());

  // The user assembly gets to expect to accumulate on an initially
  // empty system
  matrix->zero ();
  rhs->zero ();

  // Call the base class assemble function
  Parent::assemble ();
}
```
but there are situations (linear problems) where you might want to keep the matrix constant and just reassemble the rhs.  I was thinking we could do this by adding a flag at the `ImplicitSystem` level, but there might already be a better or existing way of achieving this... 

@andrsd asked about this.

andrsd:
By default, ImplicitSystem::assemble() zeroes out the matrix and
right-hand side when the method is called.  In several cases, the matrix
can be constant and the user code can choose to not form it over and
over. For this purpose we are introducing a flag that will allow us to
do so.

Closes #1231

### DIFF
--- a/include/systems/implicit_system.h
+++ b/include/systems/implicit_system.h
@@ -338,6 +338,12 @@ public:
    */
   SparseMatrix<Number> * matrix;
 
+  /**
+   * By default, the system will zero out the matrix and the right hand side.
+   * If this flag is false, it is the responsibility of the client code
+   * to take care of setting these to zero before assembly begins
+   */
+  bool zero_out_matrix_and_rhs;
 
 
 protected:
--- a/src/systems/implicit_system.C
+++ b/src/systems/implicit_system.C
@@ -44,6 +44,7 @@ ImplicitSystem::ImplicitSystem (EquationSystems & es,
 
   Parent            (es, name_in, number_in),
   matrix            (libmesh_nullptr),
+  zero_out_matrix_and_rhs(true),
   _can_add_matrices (true)
 {
 }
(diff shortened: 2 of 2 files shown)


roystgnr:
Gah.  I guess this is the best immediate solution to the problem... but in hindsight we really should have deprecated assemble() *years* ago, and gotten everything supporting and everybody moved to assembly() (the version with the get_residual+get_jacobian arguments) instead.

jwpeterson:
Directly accessing the flag (no getter/setter) matches the way `System::assemble_before_solve` is currently implemented, but maybe we shouldn't follow that precedent?  I don't feel too strongly about it either way.

@roystgnr I don't mind eventually deprecating `ImplicitSystem::assemble()` but currently it's not even mentioned in the documentation that one should prefer the `assembly()` interface.  Personally I thought it was something `FEMSystem`-specific...

andrsd:
I admit that I did not dig a lot about how to use the API. I am using `LinearImplicitSystem` and inside `assembly()`, it calls `assemble()`, which zeroes out the matrix and RHS, so I have to form the matrix every time.

PR #1243, opened 2017-02-15
AbaqusIO: Support generated elsets and nsets.

jwpeterson:
This is a specific feature of the Abaqus input file format which looks like e.g.:

*Elset, elset=Set-1, generate
 38,  54,   2

where the three numbers are (start, end, stride). The "generate"
keyword is what differentiates these sets from standard elsets and
nodesets.

### DIFF
--- a/include/mesh/abaqus_io.h
+++ b/include/mesh/abaqus_io.h
@@ -99,7 +99,16 @@ private:
    * The input to the function in this case would be foo, the
    * output would be bar
    */
-  std::string parse_label(std::string line, std::string label_name);
+  std::string parse_label(std::string line, std::string label_name) const;
+
+  /**
+   * Returns true if the input string is a generated elset or nset,
+   * false otherwise. The input string is assumed to already be in all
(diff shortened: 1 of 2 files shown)


jwpeterson:
Error message is:
```
In file included from /usr/include/errno.h:36:0,
                 from ./include/libmesh/netcdf.h:15,
                 from ./include/libmesh/exodusII.h:46,
                 from ./include/libmesh/exodusII_io_helper.h:65,
                 from ../src/mesh/exodusII_io_helper.C:19:
../contrib/eigen/eigen/Eigen/src/Core/util/Memory.h: In function ‘void* Eigen::internal::generic_aligned_realloc(void*, std::size_t, std::size_t)’:
../contrib/eigen/eigen/Eigen/src/Core/util/Memory.h:167:5: error: ‘__errno_location’ was not declared in this scope
     errno = ENOMEM; // according to the standard
     ^
```
We have had issues with errno.h not being include-able from within a namespace, but I don't know why it only happens sometimes?

Issue #1046, opened 2016-08-02
Optimize DenseVector


friedmud:
DenseVector is ancient and it shows!  One of the main things is that none of the functions are vectorizable (at least by Clang).  The current reason why is because of the use of `unsigned int` for loop counters while using `_vals.size()` for the ending.  The type mismatch causes issues.

However, that isn't the only thing that needs to be fixed.  There is a lot of using `operator()` instead of just accessing `_vals`, etc.

So: this is just a general ticket to go through and optimize `DenseVector` as much as possible.

jwpeterson:
The type mismatch issue was mostly addressed in #1220.
Getting optimzations by using Eigen was addressed in 8d74337750e4f2f2c99f21ce8e1d9d896fbf9f58.

friedmud:
Cool - thanks for working on this @jwpeterson !

PR #1244, opened 2017-02-15
Add Eigen-based optimization for DenseVector<T>::dot().

jwpeterson:
Also, improve and shorten the l1_norm(), l2_norm(), and linfty_norm()
Eigen-based optimizations done previously.  We are now not restricted
to using the Eigen optimizations for double: they should work with any
underlying type which Eigen itself supports.

C++ bonus meme: we get to employ the obscure forms of the 'typename'
and 'template' keywords in a single line of code.

Refs #1046.

### DIFF
--- a/include/numerics/dense_vector.h
+++ b/include/numerics/dense_vector.h
@@ -258,24 +258,6 @@ private:
    * The actual data values, stored as a 1D array.
    */
   std::vector<T> _val;
-
-  /**
-   * The original inline l1_norm implementation, we fall back on this
-   * when Eigen is not available.
-   */
-  Real l1_norm_fallback () const;
-
-  /**
-   * The original inline l2_norm implementation, we fall back on this
-   * when Eigen is not available.
-   */
-  Real l2_norm_fallback () const;
-
-  /**
-   * The original inline linfty_norm implementation, we fall back on this
-   * when Eigen is not available.
-   */
-  Real linfty_norm_fallback () const;
 };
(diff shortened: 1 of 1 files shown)


jwpeterson:
Haven't seen this particular one before.
```
---- Basis dimension: 0 ----
Performing RB solves on training set
/opt/moose/gcc-6.2.0/include/c++/6.2.0/debug/vector:422:
Error: attempt to subscript container with out-of-bounds index 0, but 
container only holds 0 elements.

Objects involved in the operation:
    sequence "this" @ 0x0x7fff9dfec238 {
      type = std::__debug::vector<double, std::allocator<double> >;
    }
```

jwpeterson:
> Error makes no sense, this patch does not touch the RB code...

Actually it did make sense, needed to check for zero length vectors before accessing `&v[0]`.

friedmud:
Did you do any timing studies on this?  I'm interested because my current application REALLY relies on `DenseVector::dot()`... like REALLY relies on it.

jwpeterson:
Yes, it is 3-5x faster for large vectors (like, very large, 1M+ elements).  For smaller vectors I couldn't really tell any difference with PerfLog timings... you might have more luck with Instruments or some other profiler...

friedmud:
Ok - well... I have really detailed timings on my code (everything is
measured in nanoseconds) so I'll let you know if anything changes.

Derek

PR #1246, opened 2017-02-16
Don't delete_remote_elements() inappropriately

roystgnr:
When the user has a DistributedMesh that they're allowing to
distribute, then it will already be distributed by this point and
we're being redundant.

When the user has a DistributedMesh that they're not allowing to
distribute, distributing it anyway is a bug.

### DIFF
--- a/src/systems/equation_systems.C
+++ b/src/systems/equation_systems.C
@@ -98,10 +98,6 @@ void EquationSystems::init ()
 
   libmesh_assert_not_equal_to (n_sys, 0);
 
-  // Distribute the mesh if possible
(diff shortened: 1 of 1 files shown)


roystgnr:
Merging since this was an obvious bug.  Obvious once I saw it, anyway; I don't want to admit how many hours I wasted finding it.

friedmud:
Yikes!

@permcody This may explain why you were never able to get #8100 to do anything different when using the ghosting functors...

friedmud:
Doh: https://github.com/idaholab/moose/pull/8100

PR #1249, opened 2017-02-17
Use Elem::loose_bounding_box() in TreeNode searches

roystgnr:
This is the prerequisite part of the work in #1240: getting PointLocatorTree working correctly and efficiently with non-affine elements.


roystgnr:
Want to hit this with my own testing too before we trust it, even if Civet turns out happy.

roystgnr:
MOOSE seems happy now that I removed that last premature assert, and my own tests are working too.

PR #1251, opened 2017-02-18
Switch to signed integer loops in DenseVector.

jwpeterson:
Overflow of a signed integer results in undefined behavior, therefore
the compiler may assume a program will never generate signed overflow.
This fact may in turn introduce opportunities for optimization which
would not be possible using unsigned integers for iteration.

See also: http://www.airs.com/blog/archives/120

Refs #1046.

### DIFF
--- a/include/numerics/dense_vector.h
+++ b/include/numerics/dense_vector.h
@@ -283,10 +283,10 @@ DenseVector<T>::DenseVector (const DenseVector<T2> & other_vector) :
 
   _val.clear();
 
-  const std::size_t N = other_vals.size();
+  const int N = cast_int<int>(other_vals.size());
   _val.reserve(N);
 
-  for (std::size_t i=0; i<N; i++)
+  for (int i=0; i<N; i++)
     _val.push_back(other_vals[i]);
 }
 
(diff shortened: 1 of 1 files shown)


jwpeterson:
With this patch and compiling a simple code that calls `DenseVector::scale()` with clang and `-Rpass=loop-vectorize` (see [here](http://llvm.org/docs/Vectorizers.html)), I get:
```
/Users/petejw/projects/libmesh_git/installed/include/libmesh/dense_vector.h:398:3:
remark: vectorized loop (vectorization width: 2, interleaved count: 2) [-Rpass=loop-vectorize]
  for (int i=0; i<N; i++)
  ^
```
Note that this is for a vector which is sized at runtime, so this seems promising...

friedmud:
Nice!

jwpeterson:
After some further testing,  the following `DenseVector` functions either vectorize or can be coerced into vectorizing using pragmas:
* `scale()` and `operator*=`
* `dot()`
* `add()`
* `operator+=`
* `operator-=`
* `l2_norm()`

While the following functions did not vectorize:
* `max()`
* `min()`
* `l1_norm()`
* `linfty_norm()`

The `max()`, `min()`, and `linfty_norm()` functions don't vectorize because of the following reason:
```
loop not vectorized: value that could not be identified as reduction is used outside the loop
```
While `l1_norm()` seems to vectorize for `Real` but not `Complex`, not really sure why.  During testing, I also found that the performance of our vectorized, hand-coded loops are comparable to Eigen's performance, so that's good to know.

jwpeterson:
After a bit more digging, the min/max computation loops that did not vectorize will do so if you compile with `-ffast-math`.  This basically tells the compiler that floating point numbers behave like real numbers and the order of certain operations can be interchanged.  This comment is just for informational purposes, I'm not advocating compiling with `-ffast-math`...

Issue #965, opened 2016-05-27
Boundary id caches may not be accurate after BoundaryInfo::remove()


roystgnr:
We remove ids from the maps, but not from the lists of IDs that are maintained alongside those maps.

jwpeterson:
@roystgnr I think you've made some fixes in the vein of this Issue recently?  Can you be more specific about which lists of IDs are getting out of date?

roystgnr:
The caches _boundary_ids, _side_boundary_ids, _edge_boundary_ids, _node_boundary_ids, and _shellface_boundary_ids are not updated when remove*() is called.

If the removed entry was the *last* entry with a particular id, then that id no longer exists in the mesh, and the corresponding 2 caches are now out of date.

I don't see how this is fixable without turning remove*() from O(1) into O(N) operations.  The only fix may be to document the failing and create some sort of rebuild_caches() method that a user can call after they're done with all their remove calls.

roystgnr:
I've put in a BoundaryInfo fix or two but nothing related to this ticket.

roystgnr:
Okay, now I've got a fix for this ticket.

jwpeterson:
Pretty sure this was fixed by #1255.

PR #1255, opened 2017-02-22
BoundaryInfo::regenerate_id_sets()

roystgnr:
This allows application code to fix issue #965 by regenerating cached sets of ids after performing operations which may remove all entries of a particular id.

This isn't as user friendly as the alternative (maintain maps with use counts for every id) but should be more efficient.

I've also added regenerate_id_sets() calls to everywhere in the library I can find that might invalidate the old cache.

This could potentially trigger errors in DistributedMesh-using application code, since it makes us much more strict about not storing id information for remote elements and nodes.

### DIFF
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -88,6 +88,16 @@ public:
   void clear ();
 
   /**
+   * Clears and regenerates the cached sets of ids.
+   * This is necessary after use of remove_*() functions, which remove
+   * individual id associations (an O(1) process) without checking to
+   * see whether that is the last association with the id (an O(N)
+   * process.
+   */
+  void regenerate_id_sets ();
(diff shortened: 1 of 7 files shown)


jwpeterson:
:+1:

roystgnr:
That timeout might not be a fluke.  ReplicatedMesh::stitch_meshes is one of the places where we need to recompute id caches, and using std::set means that that recomputation gets asymptotically worse in debug mode with the -D_GLIBCXX_SET_BECOMES_STUPIDLY_SLOW option on.

roystgnr:
Going to merge this anyway; I need it to avoid triggering all kinds of DistributedMesh errors in MOOSE runs.  If the timeout recurs then we'll need to disable that test in dbg mode.

PR #1257, opened 2017-02-28
Use *parallel* max_elem_id in assert code

roystgnr:
Otherwise it's not safe to call before prepare_for_use() is called,
but that's actually one of the most useful places to be able to test
neighbor consistency!

Thanks to Barna Becsek for catching this bug.


jwpeterson:
:+1: for debugging the debugging code.

Issue #1256, opened 2017-02-24
build_node_list_from_side_list can fail on distributed meshes


roystgnr:
If a node is shared by two element's sides, one of which has a side boundary id and the other of which does not, and if the node is semilocal on a certain processor but the element-with-side-bcid is not, then that processor will not correctly assign the node its new boundary id.

This appears to be caught in at least one MOOSE test, actions/meta_action.test_meta_action, when running with 3 or 4 processors and --distributed-mesh

This blocks https://github.com/idaholab/moose/issues/1500

PR #1258, opened 2017-02-28
Fix DistributedMesh build_node_list_from_side_list

roystgnr:
It's entirely possible for processor A to have a ghost node which gets
a new BCID from a processor which doesn't own the node and doesn't
even neighbor A.  The new communication here handles that case.

I need to finish some private tests before we merge this, even if Civet likes it.

### DIFF
--- a/src/mesh/boundary_info.C
+++ b/src/mesh/boundary_info.C
@@ -1965,6 +1965,8 @@ BoundaryInfo::build_node_list_from_side_list()
     }
 
   // Process all incoming requests
+  std::vector<vec_type> responses(n_proc-1);
+
   for (processor_id_type p = 1; p != n_proc; ++p)
     {
       std::vector<dof_id_type> requested_nodes;
@@ -1980,8 +1982,6 @@ BoundaryInfo::build_node_list_from_side_list()
       Parallel::Request &request =
         node_response_requests[p-1];
 
-      vec_type response;
-
       std::vector<boundary_id_type> bcids;
 
       for (std::vector<dof_id_type>::const_iterator
@@ -1993,12 +1993,12 @@ BoundaryInfo::build_node_list_from_side_list()
           for (std::size_t i=0; i != bcids.size(); ++i)
             {
               const boundary_id_type b = bcids[i];
-              response.push_back(std::make_pair(*it, b));
+              responses[p-1].push_back(std::make_pair(*it, b));
             }
         }
 
       this->comm().send
-        (source_pid, response, request, node_responses_tag);
+        (source_pid, responses[p-1], request, node_responses_tag);
     }
 
   // Process all incoming responses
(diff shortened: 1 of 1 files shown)


roystgnr:
This doesn't fix all the BC sync issues I see but it fixes the ones that it was supposed to; ready to merge after review or after I get impatient enough.  This is another one of the many, many fixes caught by https://github.com/idaholab/moose/issues/1500

roystgnr:
Oh, yeah, and this will resolve #1256

src/mesh/boundary_info.C
```diff
@@ -1865,9 +1877,154 @@ BoundaryInfo::build_node_list_from_side_list()
...
+
+  // At this point we should know all the BCs for our own nodes; now
+  // we need BCs for ghost nodes.
+  //
+  // FIXME - parallel_ghost_sync.h doesn't work here because it
+  // assumes a fixed size datum on each node.
+  std::vector<std::vector<dof_id_type> > node_ids_requested(n_proc);
+
+  // Determine what nodes we need to request
+  for (MeshBase::const_node_iterator it = _mesh.nodes_begin(),
```
jwpeterson: I would avoid putting the `nodes_end()` call in the for-loop clause like this.  That still seems to introduce an unnecessary and noticeable performance penalty over declaring it outside, unfortunately...

src/mesh/boundary_info.C
```diff
@@ -1865,9 +1877,154 @@ BoundaryInfo::build_node_list_from_side_list()
...
+    }
+
+  // If we're on a serial mesh then we're done.
+  if (mesh_is_serial)
+    return;
+
+  // Otherwise we need to push ghost node bcids to their owners, then
+  // pull ghost node bcids from their owners.
+  Parallel::MessageTag
+    node_pushes_tag = this->comm().get_unique_tag(31337),
```
jwpeterson: Not `80085`?

roystgnr: We'll see.  I'm using "867", "5309" and "42", "6*9" in an upcoming MOOSE PR, so I'm already nearing the bottom of the cutesy jokes barrel.

src/mesh/boundary_info.C
roystgnr: How?  I understand that using nodes_end() in the condition imposes a penalty, but the initialization is only called once.  The C++ standard literally states that
```
for (init; cond; iter) { foo; }
```
should produce code equivalent to
```
{
init;
while (cond) { foo; iter; }
}
```

jwpeterson: Right, the `cond` part involves calling `_mesh.nodes_end()` repeatedly, and this is slower than hosting it outside the loop IIRC.  I can try to make a test code and verify that's still the case...

jwpeterson: Oh, sorry I see what you did now...

roystgnr:
This seems to be handling whatever I throw at it too; merging now.

PR #1259, opened 2017-02-28
Extend FirstOrderUnsteadySolvers to Support Second Order Variables

pbauman:
This PR extends the capabilities of `FirstOrderUnsteadySolver`s to support second variables as well. This is done by adding new variables to the `FEMSystem` that correspond to the "velocities" of the second order variables. That is, instead of `M(u)\ddot{u} + C(u)\dot{u} + F(u) = 0`, the user can treat the system as `M(u)\dot{v} + C(u)v + F(u)` and the `FirstOrderUnsteadySolver` will deal with the equations `\dot{u} = v`. Ultimately, this was not as seamless to existing `FEMSystem` subclasses as I was hoping, but we at least still achieve an API such that the user can support either  `FirstOrderUnsteadySolver` or `SecondOrderUnsteadySolver` without changing their code once they adhere to the paradigm I elaborate on below. I'll preface that `fem_system_ex3` was updated as part of this PR to illustrate this paradigm (see e380d4b).

First changes require the user to tell us whether their `time_evolving` variable is first order or second order. I've added a new `time_evolving` override and deprecated the old version (which defaults to 1 for backward compatibility). 

Internal to the  `FirstOrderUnsteadySolver` and subclasses, I've added functionality to deal with assembling the `\dot{u} =v` equations and the underlying infrastructure, including adding the `v` equations to the `FEMSystem`, setting the boundary conditions for `v`, etc. The resulting changes to `EulerSolver` and `Euler2Solver` were pretty minimal: we need to prepare `accel` data in the `FEMContext` so the user can use the `*_accel` methods in their code to preserve compability between ` FirstOrderUnsteadySolver` classes and  `SecondOrderUnsteadySolver` classes and then call `damping` functions and assemble the `\dot{u} = v` equations. (See commit 	2040d79 for the specific changes to those classes, supporting infrastructure was in previous commits in this PR.)

From the user perspective in the `FEMSystem` subclass, if they want their second order in time problem to be solved as a first order in time problem, they'll need to make the following changes:

1. Use the new `time_evolving` API to specify second order variables
2. For the steady parts of the residual, they'll need to test if the `TimeSolver` is steady or not and, if not, they'll need to use the `UnsteadySolver::get_second_order_dot_var` method to extract the "velocity" variable index (v in the notation above) corresponding to their second order variable (u in the notation above). This is required also in the unsteady parts of the residual, e.g. `FEMSystem::mass_residual`, but obviously the user doesn't need to check if they have a steady time solver or not. Note that if the `TimeSolver` is actually second order, the velocity index returned is just the second order variable index (u) - this is how we get compatibility between the two `TimeSolver` types.
3. Assemble into the velocity part of the residual corresponding to the velocity index the user got in step 2. Similarly, assemble into the appropriate blocks of the Jacobian.

There should be no other changes required from the user perspective in the `FEMSystem` class itself. Existing second order `FEMSystem` programs would also need to adjust projecting the initial velocity, if they are specifying it, based on the `TimeSolver` type (i.e. instead of calling `SecondOrderUnsteadySolver::project_initial_rate`, they'll just project onto the "dot" variables in the projection functor); also wouldn't need to compute initial acceleration.

Again, I was hoping the user wouldn't have to change their element assembly routines to gain this capability, but I don't see any other way. If they don't desire these features and only want to use `SecondOrderUnsteadySolver` types, then they don't need to change anything. But if anyone (like me) wants to FSI with `FEMSystem` and use existing `FirstOrderUnsteadySolver` types, they'll need to do this.

Feedback particularly welcome on the documentation. I've tried to be verbose in the code and the example, but obviously if there are still confusing points, we should make it clearer in the documentation.


include/physics/diff_physics.h
```diff
@@ -240,22 +240,45 @@ class DifferentiablePhysics
...
    * behave like du/dt = F(u), and should not call time_evolving()
    * for any variables which behave like 0 = G(u).
    */
   bool is_time_evolving (unsigned int var) const {
-    return _time_evolving[var];
+    libmesh_assert_less(var,_time_evolving.size());
+    libmesh_assert( _time_evolving[var] == 0 ||
+                    _time_evolving[var] == 1 ||
+                    _time_evolving[var] == 2 );
+    return !(_time_evolving[var] == 0);
```
roystgnr: return _time_evolving[var];
would have done the same thing here, and IMHO is just as easy to understand.

roystgnr: For that matter, why not just return the unsigned int?  That will be API compatible with 99% of uses in theory and will probably be more useful in practice.

include/solvers/unsteady_solver.h
```diff
@@ -150,6 +150,38 @@ class UnsteadySolver : public TimeSolver
    */
   virtual bool is_steady() const libmesh_override { return false; }
 
+  /**
+   * Indicate to the UnsteadySolver which variables are first order in time.
+   * Subclasses may use this information in different ways.
+   */
+  void add_first_order_var(unsigned int var)
+  { _first_order_vars.insert(var); }
+
```
roystgnr: This whole commit feels redundant; if we extend its return value to be unsigned int instead of bool, can't we query TimeSolver::system()::is_time_evolving()?  https://en.wikipedia.org/wiki/Don't_repeat_yourself isn't a hard rule but there should usually be a good reason for breaking it and I'm not sure what that is here.

roystgnr: Second-guessing myself as I read through subsequent commits: you do hit this stuff often enough that "only 2nd order variables" might in fact be worth caching so you don't end up with a bunch of "loop over all variables but skip 1st order" code.

examples/fem_system/fem_system_ex3/fem_system_ex3.C
```diff
@@ -57,15 +57,14 @@ int main (int argc, char ** argv)
   LibMeshInit init (argc, argv);
 
   // Parse the input file
-  GetPot infile("fem_system_ex3.in");
+  GetPot cl(argc,argv);
```
roystgnr: Must this be either-or?  I've been moving example after example to a "read input file, then override from command line" model, which is really easy with GetPot and seems to offer the best of both worlds.

src/solvers/first_order_unsteady_solver.C
```diff
@@ -131,4 +137,92 @@ bool FirstOrderUnsteadySolver::compute_second_order_eqns(bool compute_jacobian,
   return compute_jacobian;
 }
 
+
+void FirstOrderUnsteadySolver::add_dot_var_dirichlet_bcs( unsigned int var_idx,
+                                                          unsigned int dot_var_idx )
```
roystgnr: Okay, *this* one is actually broken.  We can't set u_dot = f on a boundary where u = f, except in the f = 0 case.  In general we need u_dot = df/dt... and I'm not sure what the right way to get that is.

roystgnr:
Documenting offline conversation:

@pbauman's immediate idea for BCs is "check if BC is time evolving, if not then set ZeroFunction for the corresponding velocity BC, if so then libmesh_error()", which would be good enough for me for now; after that's done and the --enable-complex issues are sorted, we should merge.

My other suggestions are debatable and not critical.

pbauman:
Capturing discussion with @roystgnr from earlier: The major flaw with the var order business was not housing that in the `DiffSystem`. That information really belongs there and not in the `TimeSolver`. The `TimeSolver` has access to the `DiffSystem`, so that will be fine and it should actually clean up the usage a bit in the actual use cases. I will push a fix for that and then I'll need to build a complex-valued stack to figure out what the deal is in the complex case.

jwpeterson:
Just a note that the "complex" build might be a red herring here.  The real bug is probably to do with `fem_system_ex3` and the Eigen sparse linear solvers, which CIVET uses for `--enable-complex` builds so that we do not have to build/maintain a complex-enabled PETSc in the MOOSE package.

```
Solving time step 0, time = 0
Assembling the System
Nonlinear Residual: 0.00597199
Linear solve starting, tolerance 5.97199e-06
#iterations: 5316
estimated error: -nan
Linear solve finished, step 5316, residual -nan
Trying full Newton step
  Current Residual: -nan
  Shrinking Newton step to 0.5
  Current Residual: -nan
  Shrinking Newton step to 0.25
  Current Residual: -nan
  Shrinking Newton step to 0.125
  Current Residual: -nan
  Shrinking Newton step to 0.0625
  Current Residual: -nan
  Shrinking Newton step to 0.03125
  Current Residual: -nan
  Shrinking Newton step to 0.015625
  Current Residual: -nan
  Shrinking Newton step to 0.0078125
  Current Residual: -nan
  Shrinking Newton step to 0.00390625
  Current Residual: -nan
  Shrinking Newton step to 0.00195312
  Current Residual: -nan
  Shrinking Newton step to 0.000976562
  Current Residual: -nan
  Shrinking Newton step to 0.000488281
  Current Residual: -nan
  Shrinking Newton step to 0.000244141
  Current Residual: -nan
  Shrinking Newton step to 0.00012207
  Current Residual: -nan
  Shrinking Newton step to 6.10352e-05
  Current Residual: -nan
  Shrinking Newton step to 3.05176e-05
  Current Residual: -nan
  Shrinking Newton step to 1.52588e-05
  Current Residual: -nan
Inexact Newton step FAILED at step 0
```

pbauman:
Pushed design fixes and fix for "dot" variable boundary conditions. So those parts should be good. @jwpeterson, could you elaborate on what I need to do for the example? Should I just exclude it in complex builds? Obviously that needs to be fixed before we merge, but I need to use this capability immediately, so I'll come back to this PR when I've got the results I need.

jwpeterson:
> @jwpeterson, could you elaborate on what I need to do for the example?

Well, you just need to figure out why your PR causes Eigen to produce a NaN in its residual calculation...  could be something simple like a `Real` used where `Number` should have been?  I don't think we should restrict this example to not running in complex mode.

pbauman:
@jwpeterson Ah, OK, thanks for the clarification. I thought you were possibly indicating I should disable it for complex Eigen solves, which I didn't know how to do. Will track down the problem in the near future.

BTW, I've tested this branch against GRINS master and all is good, so backward compatibility should be preserved.

pbauman:
OK, so the issue seem to be that Eigen's BiCGStab implementation chokes on the problem (don't know why). I get reasonable results using `SPARSELU`.

One note though: is the performance known to be spectacularly bad for `EigenSparseMatrix`? In particular the assembly part? Just did a little profiling and 98% of the time was spent inserting values into the Eigen matrix (inside which almost all the time was resizing crap internally in Eigen). Can we not preallocate Eigen data structures?

I'll push the fix after someone else can say something about Eigen (or our usage of it) in this regard.

jwpeterson:
> One note though: is the performance known to be spectacularly bad for EigenSparseMatrix? In particular the assembly part? Just did a little profiling and 98% of the time was spent inserting values into the Eigen matrix (inside which almost all the time was resizing crap internally in Eigen). Can we not preallocate Eigen data structures?

It's not known to be bad, but I guess I would not be surprised... I don't think anyone has spent time trying to optimize it.  Maybe open a separate issue to investigate this?  No reason for it to hold up this PR.

pbauman:
It's only the case with the `EulerSolver` and `Euler2Solver`. I mean, the others still show bottle necks at the same spots (`apply_heterogeneous_constraints` with ultimately goes down into the `add` call in eigen_sparse_matrix.C where we add each matrix component individually instead of in a block), but the `Euler(2)Solver` case is running at like 800+ seconds vs. 40 seconds for the Newmark solver case (with the same number of timesteps), with 98% of the time spent putting values into the `EigenSparseMatrix`. The number of dofs is about double in the `Euler(2)Solver` case vs. the Newmark case. I was dubious of tacking on another 20 mins. to the "Test enable complex" case.

pbauman:
OK, I pushed the fixes that got this running for me locally. This required adding f1bb6f6 to provide accessors to the `NewtonSolver`'s `LinearSolver`. I also threw in 4b56f94 for source consistency. I still need to address @roystgnr's point above about just overriding with command line options and keeping the input file before this is merged.

I'll guess we'll see how slow this ends up being on CIVET and can decide whether we want to preserve those cases or not (maybe only 1?).

pbauman:
> I don't think this is quite right: one could be using the Eigen sparse linear solvers even when complex numbers are not enabled... in which case this would still use BICGSTAB and NaN I think (?).

That's a good point. I'd assumed it was just the complex numbers bit throwing a wrench in everything, but maybe not. I'll check locally and if BICGSTAB barfs even the real-valued case, I'll squash in a fix to the example.

> Also - your commit message has a typo, "GMRES doesn't work with Eigen" maybe?

Thanks. Will fix either together with the above, or, if the above is not needed, I'll push a reword of that commit message.

pbauman:
Huh, well, it looks like the total time is dominated by Test debug anyway and since INL folks have multiple clients going so Test debug and Test enable complex can run simultaneously, I guess my vote would be for leaving the runs in there.

I'll finish up checking real vs. complex BICGSTAB and the input file stuff and commit log update hopefully today.

jwpeterson:
Once this PR is merged, I was thinking we should start a v1.2.0 branch?  @roystgnr can continue to make DistributedMesh bug fixes as long as necessary -- we'll just cherry-pick them over as usual. Thoughts?

roystgnr:
There's something wrong with the new point locator bounding box stuff, and we definitely need that sorted out before a release, but the fix ought to be localized to a few files and easy to cherry-pick once I've figured it out, so we can branch now if you like.

jwpeterson:
@pbauman I can grab this and add a small commit fixing fd16f5f, then get it merged?  I think it is all working except for the minor #ifdef COMPLEX issue...

pbauman:
Sorry, I got slammed. I'll finish this out now (I have an interruption at 1). in addition to testing the complex/ifdef thing, I need to address @roystgnr's comment about keeping the input file around and overriding via the command line. Shouldn't take too long.

And yes, I would love for this to go into v1.2.

jwpeterson:
OK, I will let you continue...

pbauman:
Confirmed that Eigen BICGSTAB dies even in the real-valued case. I'll squash in removing the ifdef.

pbauman:
OK, closing this and reopening a new one since I branched to do the rebase squashing in fixes here and there.

pbauman:
Rebased version in #1276

PR #1263, opened 2017-03-07
Updating for VTK piece/parallel XML writers for changes in functionality

acbauer:
In parallel the vtkXMLPUnstructuredGridWriter needs to have the
vtkMPIController set so that it can communicate properly
between MPI processes to figure out the information in the .pvtu
meta-file.

This compiled on my machine but I did not run any tests on this change since I did not know how to do that. I am more than happy to run tests on the change if someone can tell me how to do that.

### DIFF
--- a/include/mesh/vtk_io.h
+++ b/include/mesh/vtk_io.h
@@ -34,6 +34,9 @@
 
 // Forward declarations
 class vtkUnstructuredGrid;
+#ifdef LIBMESH_HAVE_MPI
+class vtkMPIController;
+#endif
 
 namespace libMesh
 {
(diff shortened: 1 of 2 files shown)


acbauer:
@pbauman Please guide me through getting this change into libMesh.

jwpeterson:
Hi @acbauer, thanks for your contribution.

I've invited you to join the "Associates" group for libmesh so your PRs will be tested automatically in the future.

jwpeterson:
Note that we already have the `vtkMPIController` setup and teardown code in libmesh.h and libmesh.C.  Is that not sufficient?

acbauer:
Didn't notice that -- yes that should work just fine. Probably better than my implementation since I didn't know where the libMesh global initialization and finalization stuff was done.

jwpeterson:
OK.  As you can probably tell, the `VTKIO` class was written by non-VTK-experts so it would be great if you gave it a once over...

For example, on line 172 we have an `Update()` call commented out, but I'm not sure if we should, etc.
```
// _vtk_grid->Update(); // FIXME: Necessary?
```

PR #1264, opened 2017-03-07
Cleaning up the VTK parts

acbauer:
There was a likely VTK memory leak and/or access to deleted
memory (the leak check can be done at run-time by enabling the
VTK_DEBUG_LEAKS CMake option in the build). Other changes were
replacing non-thread safe methods with thread safe methods.
These are the getter methods that pass in the data to fill
as opposed to returning a value to the data.

### DIFF
--- a/include/mesh/vtk_io.h
+++ b/include/mesh/vtk_io.h
@@ -26,6 +26,7 @@
 
 #ifdef LIBMESH_HAVE_VTK
 #include "vtkType.h"
+#include "vtkSmartPointer.h"
 #endif
 
 // C++ includes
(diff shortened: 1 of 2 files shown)


jwpeterson:
:+1: looks like a definite improvement.

acbauer:
To  get a bit more comfortable with the code I ran the `fem_system/fem_system_ex2` example and it seems to properly write out the vtu files. No VTK memory leaks were detected both before and after the change. I didn't find an example where libMesh reads in a VTK file. Is there one to test that?

Also, is there anything else that I need to do in order to have this PR merged into master? It looks like I don't have write permission (rightfully so).

pbauman:
Agree with @jwpeterson, this looks like a nice improvement.

> I didn't find an example where libMesh reads in a VTK file. Is there one to test that?

I'm not sure, actually. Most of us use ExodusII. Certainly that shouldn't hold up this PR.

> Also, is there anything else that I need to do in order to have this PR merged into master?

This looks good to me, but I'll let @jwpeterson or @roystgnr make the final call.

Thanks a bunch @acbauer!

jwpeterson:
> Also, is there anything else that I need to do in order to have this PR merged into master?

Nope, you are good.  I was just waiting for the CI to completely finish, which didn't happen before I left work for the day yesterday. Thanks!

PR #1266, opened 2017-03-07
Add SubdomainPartitioner

jwpeterson:
This has been requested to support a specific MOOSE application, but it seems like it could also be generally useful in certain situations.

Refs idaholab/moose#8672

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -367,9 +367,11 @@ am__libmesh_dbg_la_SOURCES_DIST = src/base/default_coupling.C \
 	src/partitioning/parmetis_partitioner.C \
 	src/partitioning/partitioner.C \
 	src/partitioning/partitioner_factory.C \
-	src/partitioning/sfc_partitioner.C src/physics/diff_physics.C \
-	src/physics/diff_qoi.C src/physics/fem_physics.C \
-	src/quadrature/quadrature.C src/quadrature/quadrature_build.C \
+	src/partitioning/sfc_partitioner.C \
+	src/partitioning/subdomain_partitioner.C \
+	src/physics/diff_physics.C src/physics/diff_qoi.C \
+	src/physics/fem_physics.C src/quadrature/quadrature.C \
+	src/quadrature/quadrature_build.C \
 	src/quadrature/quadrature_clough.C \
 	src/quadrature/quadrature_clough_1D.C \
 	src/quadrature/quadrature_clough_2D.C \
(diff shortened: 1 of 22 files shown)


YaqiWang:
Looks like this partitioner simply assign processor id specified by user to elements, am I right? What I want is to partition subdomains independently into np parts and all elements of all subdomains with the same processor id form the subdomain for the processor with that ID.

jwpeterson:
> Looks like this partitioner simply assign processor id specified by user to elements, am I right?

Yes, my reading of idaholab/moose#8672 led me to believe this is what you wanted.

> What I want is to partition subdomains independently into np parts and all elements of all subdomains with the same processor id form the subdomain for the processor with that ID.

I've read this sentence four times now and can't quite parse it.  You should probably come by and draw a picture of what you want... generally speaking, subdomain ids and processor ids are completely independent concepts, and it sounds like you might be mixing up their meanings...

jwpeterson:
In any case, I still think that there is a general usefulness in this PR as it stands so I'd still like to merge it.  

One theoretical use case would be that you have a Cubit-generated mesh with different subdomains which actually correspond to different physics and intend to employ subdomain-restricted variables.  Then you'd like to partition certain subdomains together on a single processor or set of processors so that a particular `-sub_pc_type` could be employed... not sure if this is actually possible.

roystgnr:
What we really need for the domain-dependent physics case is an ability to "weight" different elements differently; in particular we probably ought to be *already* automatically taking the number of variables (or the number squared?) on an element into account before handing it to Parmetis, when deciding how to break up an SFC, etc.

I can see how this code could be useful regardless, but I'm sure it'll bitrot if it's not being used from something.  So if it's not going to be the backend of a MOOSE feature right away, could you add a unit test before merging?  I know we only run the unit tests on one processor in Civet right now, but I typically run them on 4, 12, or 24 so we'll at least be able to catch regressions within a few days.

jwpeterson:
> So if it's not going to be the backend of a MOOSE feature right away, could you add a unit test before merging?

Sure, I wrote a test code while developing this, so I'll just modify that into a unit test.

roystgnr:
Thanks!  When you're done run the unit test manually on a few different processor counts and then :+1: to merge; the code looked fine.

YaqiWang:
Weighted domain partitioning does not help in my case because MOOSE takes evaluation of aux kernels on linear and kernel evaluation as two separated parts. In between there is a solution vector close which is an MPI barrier. Because aux kernel evaluation could be severely imbalanced, lots of processors have to wait at the beginning of the kernel evaluations. We could remove the solution close, because it is unnecessary in my case, but it may not apply in general.

jwpeterson:
After discussing with @YaqiWang I have a better idea of what's needed.  Basically the idea is to partition specified sets of subdomains separately (using e.g. Metis) onto N procs and then combine the individual partitionings to get the final partitioning.  This will generally result in non-optimal communication patterns (see cartoon) but for certain applications will improve the load balancing.  E.g. If you assume the "cost" of computing residuals, etc. for subdomains 1..N is much higher than for N+1..M, you want to ensure that each processor gets a fair share of the work being done on 1..N.

subdomain_partitioner

Issue #1271, opened 2017-03-09
Infinite loop for mesh in the xz-plane


dknez:
I recently updated to the git HEAD, and it is triggering an infinite loop for me related to the point locator. In particular, I have a simple 2D mesh in the xz-plane, and calling

```UniquePtr<PointLocatorBase> point_locator = mesh.sub_point_locator();```

gives an infinite loop.

I'm happy to send the mesh to anyone who wants to test this themselves...

dknez:
My guess is that the new call to loose_bounding_box in TreeNode<N>::insert is the cause of the issue. I'm looking into this...

dknez:
@roystgnr the cause of the issue is Tri6::loose_bounding_box. If I comment that method out and fall back on Tri3::loose_bounding_box then this works fine.

The issue seems to be that Tri6::loose_bounding_box generates *very* loose bounding boxes so that bounding_box.intersect(bbox) in TreeNode<N>::insert ends up being always true, which leads to the infinite loop.

Note that the TRI6 elements in the mesh I'm using all have straight sides.

dknez:
It looks like the bug is that the index for the two for-loops in Tri6::loose_bounding_box should start at 1, not 0. That fixes the issue for me. I'll make a PR for this.

roystgnr:
Closed via #1272

PR #1272, opened 2017-03-09
Bug fix for Tri6::loose_bounding_box ()

dknez:
The for-loop indices should start at 1.

This refers to #1271.


roystgnr:
Well, that's embarrassing.  Thanks for catching it so quickly!

pbauman:
Was there an easy test case (a couple of elements?) that triggered this so we could drop it in the unit test suite to help prevent future regressions?

dknez:
We could make a test case pretty easily. I think a test that asserted something about the bounding boxes for some example elements would be a good idea. A TRI6 test case like that would have probably caught this error.

PR #1274, opened 2017-03-10
Elem::loose_bounding_box() fixes

roystgnr:
I clearly should not be allowed to write code without writing unit tests along with it.

This adds Elem::loose_bounding_box() unit tests and adds two fixes to the flaws they made clear.

This should fix the PointLocatorTree (and hence PeriodicBoundary) regression caught by https://github.com/idaholab/moose/pull/8626


dknez:
:+1:

PR #1276, opened 2017-03-10
#1259 Rebase

pbauman:
Redux of #1259 with several fixes squashed in. This should address all comments from the #1259.

### DIFF
--- a/src/systems/diff_system.C
+++ b/src/systems/diff_system.C
@@ -117,7 +117,17 @@ void DifferentiableSystem::init_data ()
   time_solver->init_data();
 }
 
+void DifferentiableSystem::time_evolving (unsigned int var)
+{
+  libmesh_deprecated();
 
+  DifferentiablePhysics::time_evolving(var);
+}
+
+void DifferentiableSystem::time_evolving (unsigned int var, unsigned int order)
+{
+  DifferentiablePhysics::time_evolving(var,order);
+}
 
 UniquePtr<DiffContext> DifferentiableSystem::build_context ()
 {
--- a/src/systems/diff_system.C
+++ b/src/systems/diff_system.C
@@ -19,6 +19,7 @@
 #include "libmesh/diff_solver.h"
 #include "libmesh/diff_system.h"
 #include "libmesh/time_solver.h"
+#include "libmesh/unsteady_solver.h"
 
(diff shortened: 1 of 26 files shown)


pbauman:
Rebased on master while I was at it.

pbauman:
990c73d in particular has the squashed in changes to override via command line and keeps around the input file. 2148606 has squashed in fixes for always using SPARSELU with any Eigen solver, not just in complex mode.

This should be good to merge, assuming CIVET is still happy.

src/systems/diff_system.C
```diff
@@ -209,4 +192,174 @@ void DifferentiableSystem::release_linear_solver(LinearSolver<Number> *) const
...
+                                                    dbc.g_fem.get());
+
+                  else
+                    new_dbc = new DirichletBoundary(dbc.b,
+                                                    vars_to_add,
+                                                    *this,
+                                                    dbc.f_fem.get());
+                }
+              else
+                libmesh_error_msg("Could not find valid boundary function!");
```
roystgnr: Didn't we decide that this was wrong?  The best we can do for now is add ZeroFunction DBCs for the dot vars, or scream-and-die if the primal var DBCs are FEM-dependent or time-dependent?

pbauman: Yes, and it was fixed in 9edd8ba. Should I squash that fix commit in?

roystgnr: No, that's fine; I just missed it.

roystgnr:
:+1: when Civet's done.

jwpeterson:
After merging this, I get a compile error:
```
../src/systems/diff_system.C: In member function ‘bool libMesh::DifferentiableSystem::have_first_order_scalar_vars() const’:
../src/systems/diff_system.C:338:56: error: ‘FEFamily’ is not a class or namespace
           if( this->variable(*var_it).type().family == FEFamily::SCALAR )
                                                        ^
```
The CI was happy... is my compiler just dumb?

jwpeterson:
Maybe it's a C++11 thing that our no-C++11 test doesn't catch.  I happen to be configured with `--disable-cxx11` on GCC 5.2.0 in this particular case...

pbauman:
Thanks!

PR #1277, opened 2017-03-10
Added pyramid_face_map info to exodusII_io_helper.C

dknez:
Pyramid elements are not specified in the ExodusII document. The face map given here matches what is generated for pyramid elements in CUBIT.

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -238,7 +238,10 @@ const int ExodusII_IO_Helper::ElementMaps::hex_face_map[6]     = {1, 2, 3, 4, 0,
 const int ExodusII_IO_Helper::ElementMaps::hex27_face_map[6]   = {1, 2, 3, 4, 0, 5};
 //const int ExodusII_IO_Helper::ElementMaps::hex27_face_map[6]   = {1, 0, 3, 5, 4, 2};
 const int ExodusII_IO_Helper::ElementMaps::prism_face_map[5]   = {1, 2, 3, 0, 4};
-const int ExodusII_IO_Helper::ElementMaps::pyramid_face_map[5] = {-1,-1,-1,-1,-1}; // Not Implemented!
+
+// Pyramids are not included in the ExodusII specification. The ordering below matches
+// the sideset ordering that CUBIT generates.
+const int ExodusII_IO_Helper::ElementMaps::pyramid_face_map[5] = {0, 1, 2, 3, 4};
 
 //These take a libMesh ID and turn it into an Exodus ID
 const int ExodusII_IO_Helper::ElementMaps::tet_inverse_face_map[4]     = {4, 1, 2, 3};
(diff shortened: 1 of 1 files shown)


jwpeterson:
Thanks!  I thought I had tested Pyramid elements in Exodus, but probably I did not have any sidesets or didn't check it...

dknez:
I hit an assert when opening an exo file with a pyramid with a sideset, since it tried to use the dummy "-1" values that are currently in the map. This fixes that issue.

dknez:
OK to merge?

roystgnr:
Good to merge right away if necessary, but a unit test with a CUBIT-generated mesh wouldn't hurt if you have time.

dknez:
I have a CUBIT-generated exodus file which could be read in for this test. However, I'd have to save the exo file in the repository, which didn't seem ideal for a unit test...

roystgnr:
It depends on how large the file is.  A dozen elements would be fine; a million not so much.

jwpeterson:
Is it really that straightforward? I seem to recall @pbauman trying to add a unit test that would read a file, but running into issues with the automake build system?  If this is a solved problem, it would open up a lot more possibilities for unit testing mesh things.

pbauman:
> I seem to recall @pbauman trying to add a unit test that would read a file, but running into issues with the automake build system? If this is a solved problem, it would open up a lot more possibilities for unit testing mesh things.

It wasn't an issue per se, but more of an annoyance (for the solution I was using then). However, I've been using a different solution since then that's been working pretty well and it's pretty easy to add new files and parse them.

My solution has been to add testing_files_path.h.in (or whatever name) that will be passed through configure which will set the abs_src_path for where the files live (in this case mesh files, but could be restarts or GetPot files). The following is what I do in GRINS: see [here](https://github.com/grinsfem/grins/blob/master/test/common/grins_test_paths.h.in) for the header, [here](https://github.com/grinsfem/grins/blob/master/configure.ac#L422) for the complementary requirement in configure.ac, and [here](https://github.com/grinsfem/grins/blob/master/test/unit/variables.C#L74) is an example using it. That testing_files_path.h.in will also need to be added to EXTRA_DIST in the Makefile.am to make sure it gets distributed with tar balls. 

Once that is setup, you can just drop new files in the source tree, include that header in your test, and then just read the file. The file doesn't need to be symlinked or otherwise copied to the build tree, so this just works with VPATH builds. This has worked great for me for needing to read files from within a CppUnit test.

roystgnr:
IIRC the way we used to handle files with out-of-source-build-compatible tests in automake was a giant pain in the neck: we'd use configure to create symlinks, or to put path names in a script, or some such fragile thing.  What we do now is just std::getenv("srcdir") to find where our files are and we're off and running.

The remaining catch is that $srcdir is set by the Makefile, which means we'd lose the ability to just run by hand "./unit_tests-devel", etc., which would cramp debugging a bit anytime someone forgot they needed set e.g. "srcdir=../../tests" manually.  I think spitting out a warning/reminder if we find srcdir unset would be adequate though.

pbauman:
> IIRC the way we used to handle files with out-of-source-build-compatible tests in automake was a giant pain in the neck: we'd use configure to create symlinks, or to put path names in a script, or some such fragile thing

Yeah, it was pretty gross. But that was mostly for doing tests where we wanted to run a shell script that fed path-dependent things to the testing program, not for CppUnit things.

> What we do now is just std::getenv("srcdir") to find where our files are and we're off and running.

Basically, yeah. When I was trying to hash out ideas a while back about how to improve the situation in GRINS, I stumbled onto the documentation for the automake variable `AM_TESTS_ENVIRONMENT` that will set whatever variables you define into the shell environment when it runs your tests with `make check`, e.g. [here](https://github.com/grinsfem/grins/blob/master/test/Makefile.am#L2) in GRINS. So we could use those variables in our bash script that ran the tests without having to generate a bunch of extra crap. (We couldn't avoid all symlinks though when we need to parse something from a path specified in the input file, but that hasn't been a head ache and is really only relevant when running tests using a bash script and the test uses a GetPot input file and the GetPot input file needs a full path to a file somewhere else).

> The remaining catch is that $srcdir is set by the Makefile, which means we'd lose the ability to just run by hand "./unit_tests-devel", etc., which would cramp debugging a bit anytime someone forgot they needed set e.g. "srcdir=../../tests" manually.

This is true.

> I think spitting out a warning/reminder if we find srcdir unset would be adequate though.

Since something is going to have to be set anyway by configure (either in the Makefile or elsewhere), why not just drop it in a header and just use that header for the paths? That way, we could still run the tests manually and is no more fragile than having to have that path set in the Makefile.

PR #1278, opened 2017-03-12
Add new subdomain-based Partitioner

jwpeterson:
The design/requirements for this new Partitioner are described in idaholab/moose#8672 by @YaqiWang.

In the first image below, the existing `MetisPartitioner` was used to partition the domain into 6 parts, while in the second image, the new `SubdomainPartitioner` was used to partition the central (Quads) and outer regions (Tris) separately using the `MetisPartitioner` internally. In the latter case, the elements in the central region are more evenly divided among the available processors, so the work load might be better distributed.  On the other hand, several of the partitions in the second case are non-contiguous, and the overall communication pattern is probably much worse.  Any of the other libmesh partitioners can be used as the "internal" partitioner within `SubdomainPartitioner`, so different results are of course possible.
 
No unit test added yet, however I am planning add one based on the Mesh in these images, as soon as I figure out how to add Mesh files to the unit test suite (discussed over on #1277).

metis_only_6proc_viridis
metis_6proc_2chunks_viridis_c

### DIFF
--- a/include/geom/bounding_box.h
+++ b/include/geom/bounding_box.h
@@ -95,7 +95,7 @@ public:
   bool intersects (const BoundingBox &) const;
 
   bool intersect (const BoundingBox & b) const
-    { libmesh_deprecated(); return this->intersect(b); }
+    { libmesh_deprecated(); return this->intersects(b); }
 
   /*
    * Returns true iff the bounding box contains the given point.
--- a/Makefile.in
+++ b/Makefile.in
@@ -363,6 +363,7 @@ am__libmesh_dbg_la_SOURCES_DIST = src/base/default_coupling.C \
 	src/parallel/parallel_sort.C src/parallel/threads.C \
 	src/partitioning/centroid_partitioner.C \
 	src/partitioning/linear_partitioner.C \
+	src/partitioning/mapped_subdomain_partitioner.C \
 	src/partitioning/metis_partitioner.C \
 	src/partitioning/parmetis_partitioner.C \
 	src/partitioning/partitioner.C \
@@ -760,6 +761,7 @@ am__objects_1 = src/base/libmesh_dbg_la-default_coupling.lo \
 	src/parallel/libmesh_dbg_la-threads.lo \
 	src/partitioning/libmesh_dbg_la-centroid_partitioner.lo \
 	src/partitioning/libmesh_dbg_la-linear_partitioner.lo \
+	src/partitioning/libmesh_dbg_la-mapped_subdomain_partitioner.lo \
 	src/partitioning/libmesh_dbg_la-metis_partitioner.lo \
 	src/partitioning/libmesh_dbg_la-parmetis_partitioner.lo \
 	src/partitioning/libmesh_dbg_la-partitioner.lo \
@@ -1085,6 +1087,7 @@ am__libmesh_devel_la_SOURCES_DIST = src/base/default_coupling.C \
 	src/parallel/parallel_sort.C src/parallel/threads.C \
 	src/partitioning/centroid_partitioner.C \
 	src/partitioning/linear_partitioner.C \
+	src/partitioning/mapped_subdomain_partitioner.C \
 	src/partitioning/metis_partitioner.C \
 	src/partitioning/parmetis_partitioner.C \
 	src/partitioning/partitioner.C \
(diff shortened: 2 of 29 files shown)


include/geom/bounding_box.h
```diff
@@ -95,7 +95,7 @@ class BoundingBox : public std::pair<Point, Point>
   bool intersects (const BoundingBox &) const;
 
   bool intersect (const BoundingBox & b) const
-    { libmesh_deprecated(); return this->intersect(b); }
+    { libmesh_deprecated(); return this->intersects(b); }
```
jwpeterson: @roystgnr please double-check this.  clang was warning about infinite recursion for every translation unit that included bounding_box.h...

YaqiWang:
Why is yellow in the triangular region is separated into three small parts? Similarly as the green in the square region. This is very close to what I want. If we can switch few colors in the central square region to reduce the communication, it will be idea.

jwpeterson:
> Why is yellow in the triangular region is separated into three small parts?

I don't know, and we can't really control that since we use Metis as a black box more or less.  There is no guarantee that Metis will give you contiguous partitions, for example @permcody showed the following two drastically different results when requesting 15 vs. 16 partitions of a simple square geometry!

metis_15
metis_16

include/geom/bounding_box.h
roystgnr: Yes, this is clearly the correct fix.  Sorry about the bad shim.

jwpeterson:
Hmm, I just thought of an issue with my implementation... when we look at the neighbors of an element for the purpose of generating the connectivity graph, we need to make sure the potential neighbors are *also* from the current chunk of subdomains being partitioned.  This could change the results, I'll report back here if it does.

include/geom/bounding_box.h
jwpeterson: Thanks, I will cherry-pick this commit over to master so it's fixed ASAP.

roystgnr: Ha!  I was just about to ask you if you would mind if I cherry-picked it over to master to fix it ASAP.

jwpeterson:
The last two patches improve the result on 6 processors a little, but e.g. the red partition is still very suspect:

metis_6proc_2chunk

The quality of the result is also highly dependent on the number of partitions used.  For example, on 5 procs it does about what you would want actually:

metis_5proc_2chunk

I'm also going to test whether we get the same result in the 6 processor case when the center region is missing entirely, just to be sure the `SubdomainPartitioner` didn't introduce the non-contiguous mappings and it is really a Metis thing...

jwpeterson:
So there's a subtle issue with `vectormap::find()` (include/utils/vectormap.h) that was biting me.  I am used to checking the result of `map::find()` against `end()` but vectormap does not follow this convention: it returns `std::lower_bound()` for the key!  This seems rather confusing and error prone, so I'd like to fix it, but for this PR it will suffice to use `vectormap::count()` instead.  The updated partitioning on 6 procs is below:

metis_6proc_2chunk_fixed

YaqiWang:
This is good! @fdkong You may want to try this our with our test to see if parallel efficiency is improved or not.

jwpeterson:
> This is good! @fdkong You may want to try this our with our test to see if parallel efficiency is improved or not.

Is that a test in MOOSE?  I still have to make a MOOSE PR providing an interface to this Partitioner... but it shouldn't be too hard.

YaqiWang:
Nope, it is the test we targeted during our last tiger team meeting.

fdkong:
> Is that a test in MOOSE? I still have to make a MOOSE PR providing an interface to this Partitioner... but it shouldn't be too hard.

It is a Rattlesnake test for TREAT.

jwpeterson:
> It is a Rattlesnake test for TREAT.

Right, I should have said MOOSE-based app. I will still need to make a MOOSE interface for using this new Partitioner, but we can talk about it offline.

friedmud:
This whole thing is still odd to me.  We should really be specifying the
_weights_ to METIS instead of trying to do a double-partitioning.  It's
pretty easy to specify the weights... you can do it per element or per node.

Derek

YaqiWang:
@friedmud The reason is that we have a MPI barrier between linear aux kernel evaluation and the kernel evaluation and the imbalance happens in the linear aux kernel evaluation, so the weighed partitioning does not help. Although in my case, the barrier is not necessary, removing it requires some non-trivial changes in MOOSE. We will try out this partitioner very soon. I think it is not a bad idea overall.

PR #1283, opened 2017-03-15
Fix 64-bit indices regression.

jwpeterson:
This fixes one 64-bit indices regression that crept in, I'll see if we can use CIVET to test for others...


jwpeterson:
One timeout on patterned mesh generation in debug mode.  It seems that I'm seeing this test timeout more... perhaps new asserts have made it slower?  If it keeps popping up, I'll look into making the test smaller.

PR #1284, opened 2017-03-15
Run contrib/bin/reindent.sh

jwpeterson:
Want to get all the formatting right in preparation for tagging v1.2.0.


roystgnr:
:+1: when Civet is happy, and thanks for the bugfix.

jwpeterson:
> thanks for the bugfix.

No problem.  I don't think we would ever have noticed this issue since we never build in that configuration, but it definitely confused the reindentation script.

PR #1287, opened 2017-03-20
Revert new METIS neighbor code until we can fix it

roystgnr:
I don't understand why this code could be leading to an occasional
segfault within METIS, but commenting it out fixes the problem for me
and so might be a useful temporary fix for others.

Refs https://github.com/idaholab/moose/issues/8759


permcody:
@roystgnr - the consensus is that we merge this and take our time to figure this out. We don't want random seg faults in MOOSE for the next month. Care if I merge? or you can. Thanks

PR #1288, opened 2017-03-28
Ignore deprecated strstream warnings in VTK

roystgnr:
Hopefully newer VTK versions have fixed this, but VTK 6.3.0 causes GCC
6.1 to spew deprecated header warnings, and we don't need to expose
our users to that.


jwpeterson:
I don't see these warnings... so if they are fixed for you I'm calling it good!

PR #1289, opened 2017-03-28
Maintain subactive neighbor links

roystgnr:
We've been mostly ignoring these in the past, but doing so has led to assertion failures in heavy adaptive coarsening with DistributedMesh.  (Thanks again to the MOOSE test suite and https://github.com/idaholab/moose/issues/1500)

Rather than weaken the assertions let's just maintain the neighbor links; the cost is trivial and it might prove useful if our geometric multigrid code ends up relying on subactive elements heavily.

### DIFF
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -1135,6 +1135,14 @@ public:
                                 const bool reset=true) const;
 
   /**
+   * Same as the \p family_tree_by_neighbor() member, but also adds
+   * any subactive descendants.
+   */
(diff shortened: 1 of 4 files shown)


roystgnr:
This fixes several MOOSE tests and it doesn't seem to regress any libMesh/GRINS/MOOSE tests, so merging right away.

roystgnr:
Where by "right away" I mean after I spent 12 hours making sure that the failures fixed by https://github.com/idaholab/moose/pull/8830 were really independent failures and not some two-steps-forward, one-step-back effect from this patch.

PR #1291, opened 2017-03-30
CheckpointIO DistributedMesh fixes

roystgnr:
With luck this may be all we need to fix https://github.com/idaholab/moose/issues/8410 - at the very least it fixes the recovery tests I've tried so far via https://github.com/idaholab/moose/issues/1500

This isn't ready to merge yet because it needs:

1. A lot more testing on my part, because after getting those recovery tests working I rebased the hell out of the code and it's possible I regressed something again

2. A bit of review from @friedmud, because I tore out half his code.  I believe that my reasons for doing so were both necessary for robustness and consistent with the philosophy he described to me last month, but still, you don't rip down Chesterton's fence all by yourself if you have a chance to chat first with the guy who put it up.

### DIFF
--- a/include/geom/compare_elems_by_level.h
+++ b/include/geom/compare_elems_by_level.h
@@ -1,0 +1,61 @@
+// The libMesh Finite Element Library.
+// Copyright (C) 2002-2017 Benjamin S. Kirk, John W. Peterson, Roy H. Stogner
+
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+
+
+#ifndef LIBMESH_COMPARE_ELEM_IDS_BY_LEVEL_H
+#define LIBMESH_COMPARE_ELEM_IDS_BY_LEVEL_H
+
+
+// Local Includes
+#include "libmesh/elem.h"
+
+// C++ Includes
+#include <numeric>
+#include <set>
+
+
+
+
+namespace libMesh {
+
+/**
+ * Specific weak ordering for Elem *'s to be used in a set.
+ * We use the id, but first sort by level.  This guarantees
+ * when traversing the set from beginning to end the lower
+ * level (parent) elements are encountered first.
+ */
+struct CompareElemIdsByLevel
+{
+  bool operator()(const Elem * a,
+                  const Elem * b) const
+  {
+    libmesh_assert (a);
+    libmesh_assert (b);
+    const unsigned int
+      al = a->level(), bl = b->level();
+    const dof_id_type
+      aid = a->id(),   bid = b->id();
+
+    return (al == bl) ? aid < bid : al < bl;
+  }
+};
+
+
+} // namespace libMesh
+
+#endif // LIBMESH_COMPARE_ELEM_IDS_BY_LEVEL_H
--- a/include/geom/compare_elems_by_level.h
+++ b/include/geom/compare_elems_by_level.h
@@ -17,8 +17,8 @@
 
 
 
-#ifndef LIBMESH_COMPARE_ELEM_IDS_BY_LEVEL_H
-#define LIBMESH_COMPARE_ELEM_IDS_BY_LEVEL_H
+#ifndef LIBMESH_COMPARE_ELEMS_BY_LEVEL_H
+#define LIBMESH_COMPARE_ELEMS_BY_LEVEL_H
 
 
 // Local Includes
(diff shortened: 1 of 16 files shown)


roystgnr:
Oh, huh.  I should have tested with ReplicatedMesh too at some point.

roystgnr:
Yeah, I *definitely* need to chat with @friedmud about this one.

I stripped out the local elements and nodes sets because, in the distributed mesh case, they were redundant with GhostingFunctor output at best and inconsistent with it at worst.  But one of the use cases of CheckPointIO is to write out distributed files from replicated meshes, so I'll want to do that still, just using the GhostingFunctors this time rather than hard-coding nodal neighbors.

But how was the older code supposed to *read* distributed files into replicated meshes?  That's clearly a feature we intended to support, since it's in the unit test, and it's a feature we *can* support with a simple allgather... but the older code didn't have an allgather!  If you tried to read a parallel CheckpointIO file into a ReplicatedMesh would it just leave the mesh in a totally inconsistent state?  If not then how was the state made consistent?  If so then how was that passing unit tests?

roystgnr:
No, wait, we can't.  Mesh allgather() is only a DistributedMesh feature.

roystgnr:
4327698 wasn't a test failure, BTW; I just didn't let the tests complete because I found a bug in --distributed-mesh behavior on my own.  I think this may be in good shape now.

roystgnr:
Alternatively, I can break DistributedMesh with every fix for ReplicatedMesh and break ReplicatedMesh with every fix for DistributedMesh.

I'm now testing with both at the same time, and I'll make the current unit test more comprehensive afterward so it won't depend so heavily on our --enable-parmesh setting.

roystgnr:
I think this has it.  It's passing all my tests too, and fixing MOOSE recovery tests with --distributed-mesh.  I'll merge tomorrow unless anyone has objections or wants time to look it over.

include/geom/compare_elems_by_level.h
```diff
@@ -0,0 +1,61 @@
...
+// Lesser General Public License for more details.
+
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+
+
+#ifndef LIBMESH_COMPARE_ELEM_IDS_BY_LEVEL_H
+#define LIBMESH_COMPARE_ELEM_IDS_BY_LEVEL_H
```
jwpeterson: This include guard doesn't exactly match the header filename.  This is an unofficial coding convention but one that we follow pretty closely everywhere else...

include/geom/compare_elems_by_level.h
```diff
@@ -0,0 +1,61 @@
...
+
+namespace libMesh {
+
+/**
+ * Specific weak ordering for Elem *'s to be used in a set.
+ * We use the id, but first sort by level.  This guarantees
+ * when traversing the set from beginning to end the lower
+ * level (parent) elements are encountered first.
+ */
+struct CompareElemIdsByLevel
```
jwpeterson: FYI The elem_hash.h file already has some Elem comparison helper functors in this vein... but since you've already done all the build system changes to add a new header, this is probably fine.

tests/mesh/checkpoint.C
```diff
@@ -47,18 +54,24 @@ public:
...
+  template <typename MeshA, typename MeshB>
+  void testSplitter(bool binary, bool using_distmesh)
   {
-    // In this test, we partition the mesh into n_procs parts.
-    const unsigned int n_procs = 2;
+    // In this test, we partition the mesh into n_procs parts.  Don't
+    // try to partition a DistributedMesh into more parts than we have
+    // processors, though.
+    const unsigned int n_procs =
+      using_distmesh ? std::min((unsigned int)2, TestCommWorld->size()) : 2;
```
jwpeterson: Prefer a C++ style `static_cast` here.

src/mesh/checkpoint_io.C
```diff
@@ -523,82 +510,129 @@ void CheckpointIO::read (const std::string & name)
...
-      std::ostringstream file_name_stream;
-
-      file_name_stream << name;
-
-      if (_parallel)
-        file_name_stream << "-" << _my_n_processors << "-" << _my_processor_id;
-
       {
-        std::ifstream in (file_name_stream.str().c_str());
+        std::ifstream in (name);
```
jwpeterson: Is `name` a `std::string`? The `ifstream` constructor is one of those insidious things that passes some (most?) C++03 compilers but is technically C++11 only code. On the other hand, we are going to start putting C++11 code in master, now, right?

roystgnr:
This *is* the repeatedly cleaned up and rebased version, I'm afraid.  I don't know if I could do much better without some incredibly time-consuming rebase surgery or without merging a lot of logically-independent commits, so I figured that if we do need to bisect I at least want to see *some* level of commit-by-commit history.

include/geom/compare_elems_by_level.h
roystgnr: Gah; this was part of a much more widespread mistake that I fixup'ed, but I missed this bit.  I'll clean it up now.

include/geom/compare_elems_by_level.h
roystgnr: I actually like this better, personally; digging through MOOSE is made much, much easier by how consistent you guys are about keeping one class per file, with the class name and filename matching.

tests/mesh/checkpoint.C
roystgnr: Will do; thanks.

src/mesh/checkpoint_io.C
roystgnr: Hmm...  this is too big a change set to backport to the release branch, huh?  But when we do add C++11 code it ought to be because we need it, and we ought to add a configure-time warning about it.  I'll make this C++03-safe.

PR #1293, opened 2017-04-03
Added support to configure for Intel 2017 compilers.

garvct:
Would like to add Intel 2017 compiler support to the libmesh configure file.


jwpeterson:
@garvct it looks like you changed the `configure` script directly? This script gets regenerated automatically when we run `bootstrap`, so your changes would be lost the next time that happens. The correct file to change is probably `m4/compiler.m4`, lines 361 and 824. I can make these changes though, thanks for the heads up.

garvct:
Ooops, I modified m4/compiler.m4 with the change to support Intel 2017 compilers.

jwpeterson:
Thanks. I will pull this and make sure bootstrap is happy before merging.

jwpeterson:
Looks good. I also invited you to the Collaborators team so your PRs will be tested automatically in the future.  Thanks again!

Issue #1286, opened 2017-03-17
v1.2.0 branch started


jwpeterson:
Tag this issue in the commit message of any commits you'd like cherry-picked over.  

As usual, let's try to restrict the cherry-picked commits to bugfixes as much as possible and fight the urge to sneak in new features at the last minute.

jwpeterson:
Release v1.2.0 is done.

PR #1295, opened 2017-04-04
bug fix for exodus write_timestep with single precision

svallaghe:
Enabling single precision when writing exodus time series was giving a wrong visualization under paraview.
This can be easily checked on the example transient_ex1 for instance, by adding the flag "true" when creating ExodusII_IO objects:
"ExodusII_IO exio(mesh,true)"

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -332,10 +332,20 @@ void ExodusII_IO_Helper::open(const char * filename, bool read_only)
   // Version of Exodus you are using
   float ex_version = 0.;
 
-  // Word size in bytes of the floating point variables used in the
-  // application program.  Exodus only supports 4-byte and 8-byte
-  // floats.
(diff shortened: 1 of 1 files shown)


roystgnr:
The second half of the patch looks good.

I'm afraid I don't understand the first half... Ah!  We can request a single-precision write from the constructor even if we're configured for higher precision, and we were doing the wrong thing in that case?

:+1: - I'll merge this afternoon unless someone objects.

svallaghe:
Yes, there's this option to write in floats to save memory/disk space, even if the computations was done using doubles.
This feature was added a few years back but without treating the time dependent case.
This fix is just for that.

jwpeterson:
Thanks for the fix, just fixed a small formatting issue before merging.

jwpeterson:
FYI this bugfix was also cherry-picked onto the v1.2.0 branch. See also #1286.

PR #1297, opened 2017-04-04
fix slow exodus file writing for large # blocks and/or boundaries

### LINKED ISSUE #968: Writing an ExodusII file with many subdomains and boundary_ids is very slow
I used mesh.write("mesh_name.exo") to write out a mesh with 547,532 elements, and with about 1,000 subdomains and boundary_ids. This took about 250 seconds. When I set all elements to be in the same subdomain, and cleared all boundary_ids, the writing took 2 seconds.

I'd be interested to see if we can optimize this "many subdomains/many boundary IDs" case.

rwcarlsen:
Switch to using batch-outputting exodus functions for block and boundary info.  Previously, the ``ex_put_elem_block`` and ``ex_put_side_set_param`` functions were being called for one block/side-set at a time in loops.  This was causing the exodus file (on disk) to be modified to handle resizing of the arrays storing this data over and over instead of just sizing them up front once.

The pathological example code provided in https://github.com/libmesh/libmesh/issues/968#issuecomment-291220239 runs (on my machine) in ~50 seconds without this PR and in ~2 seconds with this PR.

fixes #968.

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -1348,26 +1348,73 @@ void ExodusII_IO_Helper::write_elements(const MeshBase & mesh, bool use_disconti
   elem_num_map.resize(n_active_elem);
   std::vector<int>::iterator curr_elem_map_end = elem_num_map.begin();
 
+  std::vector<int> blkid;
+  std::vector<int> nelems;
+  std::vector<int> nnodes;
+  std::vector<int> nedges;
+  std::vector<int> nfaces;
+  std::vector<int> nattr;
+  NamesData types_table(num_elem_blk, MAX_STR_LENGTH);
+
   // Note: It appears that there is a bug in exodusII::ex_put_name where
   // the index returned from the ex_id_lkup is erronously used.  For now
   // the work around is to use the alternative function ex_put_names, but
   // this function requires a char ** datastructure.
   NamesData names_table(num_elem_blk, MAX_STR_LENGTH);
 
-  // This counter is used to fill up the libmesh_elem_num_to_exodus map in the loop below.
-  unsigned libmesh_elem_num_to_exodus_counter = 0;
-
   // counter indexes into the block_ids vector
   unsigned int counter = 0;
-
-  // node counter for discontinuous plotting
-  unsigned int node_counter = 0;
-
   for (subdomain_map_type::iterator it=subdomain_map.begin(); it!=subdomain_map.end(); ++it)
     {
       block_ids[counter] = (*it).first;
       names_table.push_back_entry(mesh.subdomain_name((*it).first));
 
+      // Get a reference to a vector of element IDs for this subdomain.
+      subdomain_map_type::mapped_type & tmp_vec = (*it).second;
+
+      ExodusII_IO_Helper::ElementMaps em;
+#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+      // Skip infinite element-blocks; they can not be viewed in most visualization software
+      // as paraview.
+      if (mesh.elem_ref(tmp_vec[0]).infinite())
+        continue;
+#endif
+      //Use the first element in this block to get representative information.
+      //Note that Exodus assumes all elements in a block are of the same type!
+      //We are using that same assumption here!
+      const ExodusII_IO_Helper::Conversion conv =
+        em.assign_conversion(mesh.elem_ref(tmp_vec[0]).type());
+      num_nodes_per_elem = mesh.elem_ref(tmp_vec[0]).n_nodes();
+
+      blkid.push_back((*it).first);
+      types_table.push_back_entry(conv.exodus_elem_type().c_str());
+      nelems.push_back(tmp_vec.size());
+      nnodes.push_back(num_nodes_per_elem);
+      nedges.push_back(0);
+      nfaces.push_back(0);
+      nattr.push_back(0);
+
+      counter++;
+    }
+
+  exII::ex_block_params blocks = {0};
+  blocks.elem_blk_id = &blkid[0];
+  blocks.elem_type = types_table.get_char_star_star();
+  blocks.num_elem_this_blk = &nelems[0];
+  blocks.num_nodes_per_elem = &nnodes[0];
+  blocks.num_edges_per_elem = &nedges[0];
+  blocks.num_faces_per_elem = &nfaces[0];
+  blocks.num_attr_elem = &nattr[0];
+  ex_err = exII::ex_put_concat_all_blocks(ex_id, &blocks);
+  EX_CHECK_ERR(ex_err, "Error writing element blocks.");
+
+  // This counter is used to fill up the libmesh_elem_num_to_exodus map in the loop below.
+  unsigned libmesh_elem_num_to_exodus_counter = 0;
+
+  // node counter for discontinuous plotting
+  unsigned int node_counter = 0;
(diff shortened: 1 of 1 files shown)


src/mesh/exodusII_io_helper.C
```diff
@@ -1343,6 +1343,46 @@ void ExodusII_IO_Helper::write_elements(const MeshBase & mesh, bool use_disconti
   elem_num_map.resize(n_active_elem);
   std::vector<int>::iterator curr_elem_map_end = elem_num_map.begin();
 
+  exII::ex_block_params blocks = {};
```
jwpeterson: Not sure if the tests will catch this one, but I think it's C++11 only.

src/mesh/exodusII_io_helper.C
```diff
@@ -1588,20 +1618,28 @@ void ExodusII_IO_Helper::write_sidesets(const MeshBase & mesh)
...
 
+      struct exII::ex_set sets[side_boundary_ids.size()];
+
       for (std::size_t i=0; i<side_boundary_ids.size(); i++)
         {
           boundary_id_type ss_id = side_boundary_ids[i];
           int actual_id = ss_id;
 
-          names_table.push_back_entry(mesh.get_boundary_info().get_sideset_name(ss_id));
+          sets[i] = {};
```
jwpeterson: And same here.

src/mesh/exodusII_io_helper.C
```diff
@@ -1588,20 +1618,28 @@ void ExodusII_IO_Helper::write_sidesets(const MeshBase & mesh)
...
 
-          ex_err = exII::ex_put_side_set_param(ex_id, actual_id, elem[ss_id].size(), 0);
-          EX_CHECK_ERR(ex_err, "Error writing sideset parameters");
-
-          ex_err = exII::ex_put_side_set(ex_id, actual_id, &elem[ss_id][0], &side[ss_id][0]);
-          EX_CHECK_ERR(ex_err, "Error writing sidesets");
+          names_table.push_back_entry(mesh.get_boundary_info().get_sideset_name(ss_id));
         }
 
+      ex_put_sets(ex_id, side_boundary_ids.size(), sets);
```
jwpeterson: You don't actually catch the returned error code here.  Also, not sure why you don't need to prefix it with `exII::` but please do so for consistency.

src/mesh/exodusII_io_helper.C
```diff
@@ -1343,6 +1343,46 @@ void ExodusII_IO_Helper::write_elements(const MeshBase & mesh, bool use_disconti
...
+
+      ((int*)blocks.elem_blk_id)[i] = (*it).first;
+      blocks.elem_type[i] = (char*)conv.exodus_elem_type().c_str();
+      blocks.num_elem_this_blk[i] = tmp_vec.size();
+      blocks.num_nodes_per_elem[i] = num_nodes_per_elem;
+      blocks.num_edges_per_elem[i] = 0;
+      blocks.num_faces_per_elem[i] = 0;
+      blocks.num_attr_elem[i] = 0;
+    }
+  exII::ex_put_concat_all_blocks(ex_id, &blocks);
```
jwpeterson: Need to catch the return error code.

src/mesh/exodusII_io_helper.C
```diff
@@ -1588,20 +1618,28 @@ void ExodusII_IO_Helper::write_sidesets(const MeshBase & mesh)
     {
       NamesData names_table(side_boundary_ids.size(), MAX_STR_LENGTH);
 
+      struct exII::ex_set sets[side_boundary_ids.size()];
```
jwpeterson: `struct` keyword should not be necessary.

src/mesh/exodusII_io_helper.C
```diff
@@ -1343,6 +1343,46 @@ void ExodusII_IO_Helper::write_elements(const MeshBase & mesh, bool use_disconti
...
   std::vector<int>::iterator curr_elem_map_end = elem_num_map.begin();
 
+  exII::ex_block_params blocks = {};
+  blocks.elem_blk_id = new int[num_elem_blk];
+  blocks.elem_type = new char*[num_elem_blk];
+  blocks.num_elem_this_blk = new int[num_elem_blk];
+  blocks.num_nodes_per_elem = new int[num_elem_blk];
+  blocks.num_edges_per_elem = new int[num_elem_blk];
+  blocks.num_faces_per_elem = new int[num_elem_blk];
+  blocks.num_attr_elem = new int[num_elem_blk];
```
jwpeterson: All the memory allocated for the blocks struct to use seems to be at best leaked, or at worst, incorrectly free'd by the Exodus API? Maybe just create some local `vector<int>`s here and have the `blocks` struct use that storage.

jwpeterson:
The diffs over in the MOOSE tests do seem to be related to this change. For example,
```
mortar/3d.non-conforming: exodiff: INFO .. Map option is not needed.
mortar/3d.non-conforming: Coordinates will be compared .. tol:    1e-06 (absolute), floor:        0
mortar/3d.non-conforming: exodiff: ERROR .. Block 1: element types don't agree (HEX8 != QUAD4).
mortar/3d.non-conforming: exodiff: ERROR .. Block 2: element types don't agree (HEX8 != QUAD4).
mortar/3d.non-conforming: exodiff: ERROR .. Differences found in mesh metadata.  Aborting...
```

and
```
mortar/3d.non-conforming-2nd-order: exodiff: INFO .. Map option is not needed.
mortar/3d.non-conforming-2nd-order: Coordinates will be compared .. tol:    1e-06 (absolute), floor:        0
mortar/3d.non-conforming-2nd-order: exodiff: ERROR .. Block 1: element types don't agree (HEX27 != QUAD9).
mortar/3d.non-conforming-2nd-order: exodiff: ERROR .. Block 2: element types don't agree (HEX27 != QUAD9).
mortar/3d.non-conforming-2nd-order: exodiff: ERROR .. Differences found in mesh metadata.  Aborting...
```

so it looks like the blocks are being written out in a different order in mixed geometric dimension cases.

jwpeterson:
Just some notes for possible future refactorings (does not have to be fixed for current PR):
* The ExodusII manual describes the `ex_put_concat_elem_block()` API but not the `ex_put_concat_all_blocks()` used in the PR.  The former might be easier to use because you don't have to mess with initializing the struct, you just pass a bunch of arrays.
* Similarly, the `ex_put_sets()` API used here is not described in the ExodusII manual, but the related functions `ex_put_concat_node_sets()` and `ex_put_concat_side_sets()` are. The main difference is again the passing of a struct vs. several of arrays, and needing to explicitly set the type when calling `ex_put_sets()`.

jwpeterson:
Also, I feel a little silly that we haven't figured this out before now, as it explicitly states in Section 5.2.13 that the way we do things will be slow :nauseated_face: 

```
If element blocks are written using ex_put_elem_block(or EXPELB for Fortran), significant inefficiencies can result, because the netcdf file is completely re-written after each call. For large files with many element blocks this can be prohibitively slow. To avoid this problem, the function ex_put_concat_elem_block (or EXPCLB for Fortran) may be used to write the block parameters for all element blocks in a single call.
```

dknez:
This sounds like a great update, thanks!!

jwpeterson:
Thanks! Just going to fix 1 or 2 formatting issues before merge...

dknez:
Awesome, thanks! I'll test this out.

PR #1298, opened 2017-04-04
Hopefully fix Metis bug originally introduced with SubdomainPartitioner

jwpeterson:
This fix is very relatively simple, and I have not tested it with all the cases that were segfaulting previously, but it does reliably pass adaptivity_ex2 on 4 processors which was one of the cases @roystgnr had reported not working.

### DIFF
--- a/examples/adaptivity/adaptivity_ex2/adaptivity_ex2.C
+++ b/examples/adaptivity/adaptivity_ex2/adaptivity_ex2.C
@@ -133,9 +133,9 @@ int main (int argc, char ** argv)
   // Use commandline parameter to specify if we are to
   // read in an initial solution or generate it ourself
   libMesh::out << "Usage:\n"
-               <<"\t " << argv[0] << " -init_timestep 0\n"
+               <<"\t " << argv[0] << " -init_timestep 0 -n_timesteps 25 [-n_refinements 5]\n"
                << "OR\n"
-               <<"\t " << argv[0] << " -read_solution -init_timestep 26\n"
+               <<"\t " << argv[0] << " -read_solution -init_timestep 26 -n_timesteps 25\n"
                << std::endl;
 
   libMesh::out << "Running: " << argv[0];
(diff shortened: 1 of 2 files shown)


pbauman:
Oh, cool, I'll try this out late this afternoon. I was getting segfaults on an AMR case I was running before I left for SIAM NC. Hopefully this will fix it.

roystgnr:
Ah!  That looks like it could be it.  That's both reassuring and disappointing.  I was kind of hoping we had exposed some pre-existing METIS API usage bug, so fixing it would also fix the crummy results METIS gives us on occasion.

@pbauman, the bug existed for a little over a week before we reverted, and had a high probability of crashing during AMR repartitioning, so yeah, if you pulled at the wrong time, that could have been it.

roystgnr:
:+1: when you're ready

roystgnr:
Oh, wait

roystgnr:
Could you reorder those last two commits, or merge them?  We might as well not have a gap in between uncommenting the buggy code and fixing it.

jwpeterson:
> Could you reorder those last two commits, or merge them?

I will squash them, I was initially testing with 8b2fd9b to confirm that I could reproduce the bug, but it does not need to be in the final history.

jwpeterson:
This patch also seems to work for Cody's AMR grain growth example:
```
cd $MOOSE_DIR/modules/phase_field/examples/grain_growth/grain_growth_2D_graintracker.i
mpiexec -n 12 ../../phase_field-opt -i grain_growth_2D_graintracker.i
```
that was crashing on previously, so I'm fairly optimistic it's the right fix.

jwpeterson:
By the way @roystgnr, I cherry-picked your initial fix (439f48e) onto the v1.2.0 branch so the release would not have the original bug, but if this new fix works, I'd like to also cherry-pick it onto v1.2.0. On the other hand, I could just leave it alone depending on your preference...

roystgnr:
No, let's definitely cherry-pick this on too.

PR #1299, opened 2017-04-05
Fix ExodusII_IO::copy_* on distributed meshes

roystgnr:
Using Exodus with a distributed mesh is probably a bad idea, but if
a user does it, then we should support it, not just throw a false
positive exception or die with a segfault.

This fixes the MOOSE restart_diffusion tests with --distributed-mesh
for me.


permcody:
Well if this is all it takes to fix that, great. If this becomes a burden to support, then I'm all for dropping this support since it won't scale.  Thanks!

roystgnr:
Nah, we should definitely support this.  There are legitimate use cases, like starting up a fine scale distributed simulation from an initial condition or initial guess in a coarse mesh in an Exodus file.

PR #1300, opened 2017-04-05
Fix for upcoming MatGetSubMatrix name change in PETSc 3.8.

jwpeterson:
The MatGetSubMatrix function was renamed to MatCreateSubMatrix in
petsc/petsc@7dae84e. Note: using the "RELEASE" version of this macro
supports people currently using petsc-dev since the change has already
been merged there.

@fdkong if you keep up with petsc-dev, would you mind pulling this branch and seeing if it works for you?


fdkong:
OK, let me take a try.

fdkong:
There are also a few `MatGetSubMatrix`s in `petsc_linear_solver.C`. You may want to update there as well.

jwpeterson:
> You may want to update there as well.

Whoops, should have done a text replace.  Thanks for checking on that, will push an updated fix.

jwpeterson:
Thanks @fdkong for confirming that this makes libmesh compile with petsc-dev.

We are also running `make check` now and will open a separate PRs if we find any other issues...

PR #1302, opened 2017-04-05
More complete deprecation of MeshTools::BoundingBox

jwpeterson:
MeshTools::BoundingBox is itself deprecated, so to be consistent we should also deprecate interfaces that return them.


roystgnr:
Thanks!

PR #1303, opened 2017-04-05
Assert consistency in processor_bounding_* calls

roystgnr:
There appear to be some serious bugs in MOOSE use of this API;
hopefully this will smoke them out.

### DIFF
--- a/src/mesh/mesh_tools.C
+++ b/src/mesh/mesh_tools.C
@@ -430,6 +430,10 @@ libMesh::BoundingBox
 MeshTools::create_processor_bounding_box (const MeshBase & mesh,
                                           const processor_id_type pid)
 {
+  // This can only be run in parallel, with consistent arguments.
+  libmesh_parallel_only(mesh.comm());
+  libmesh_assert(mesh.comm().verify(pid));
+
   libmesh_assert_less (pid, mesh.n_processors());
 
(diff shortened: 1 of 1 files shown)


jwpeterson:
Looks like this has a good chance of conflicting with #1302.

roystgnr:
Let's merge #1302 first; I'll redo this later.  I just wanted to get it into CI right away and confirm or disprove my suspicions.

jwpeterson:
OK, either way it should be a pretty simple conflict resolution.

roystgnr:
What's up with the "Test debug:linux-gnu" Civet target?  The "Details" link just leads to the "Test debug" page, but invalidating that only restarts the "Test debug" target.

permcody:
@brianmoose - any ideas here?

brianmoose:
This PR is rather old. Test names reported to GitHub used to have the target platform, ie "linux-gnu" in the name. This could get rather long so I removed them (unless there was ambiguity).
So when you invalidated "Test debug:linux-gnu" it restarted the job but now CIVET is reporting that same job as "Test debug".

roystgnr:
Okay, that makes sense, thanks.

In that case, I'll merge this if/when "Test debug" passes and I won't worry that there might be some separate test suite I've failed to restart.

PR #1304, opened 2017-04-06
Fix PETSc lock error from #1294

### LINKED ISSUE #1294: Lock error in PetscDiffSolver
When changing the solver on the fem_system_ex1 from DiffSolver to a PetscDiffSolver there is a problem with a lock on PETSc vec.

The error is copied bellow and attached is the diff from master code and modified.

0]PETSC ERROR: --------------------- Error Message --------------------------------------------------------------
[0]PETSC ERROR:  Vec is locked read only, argument # 1
[0]PETSC ERROR: See http://www.mcs.anl.gov/petsc/documentation/faq.html for trouble shooting.
[0]PETSC ERROR: Petsc Release Version 3.7.5, Jan, 01, 2017 
[0]PETSC ERROR: ./test on a linux-gnu-dbg named office by fcanesin Fri Mar 31 02:03:41 2017
[0]PETSC ERROR: Configure options PETSC_ARCH=linux-gnu-dbg --with-cc=clang --with-cxx=clang++ --with-fc=0 --prefix=/home/fcanesin/libs/petsc --download-metis --download-parmetis --download-openmpi --download-f2cblaslapack --with-debugging=1
[0]PETSC ERROR: #1 VecGetArray() line 1647 in /home/fcanesin/libs/petsc-clang/src/vec/vec/interface/rvector.c
[0]PETSC ERROR: #2 VecSetValues_MPI() line 876 in /home/fcanesin/libs/petsc-clang/src/vec/vec/impls/mpi/pdvec.c
[0]PETSC ERROR: #3 VecSetValues() line 892 in /home/fcanesin/libs/petsc-clang/src/vec/vec/interface/rvector.c

Example with diff:
[fem_system_ex1.zip](https://github.com/libMesh/libmesh/files/892091/fem_system_ex1.zip)

canesin:
This just replace .enforce_constraints_exactly(sys) with .enforce_constraints_exactly(sys, sys.current_local_solution.get()) on the residual and Jacobian computations.

In the solve there is already a:
#ifdef LIBMESH_ENABLE_CONSTRAINTS
  _system.get_dof_map().enforce_constraints_exactly(_system);
#endif

I'm in doubt if this shouldn't be moved after SNESSolve and have the the ifdef removed, closes #1294 .

### DIFF
--- a/src/solvers/petsc_diff_solver.C
+++ b/src/solvers/petsc_diff_solver.C
@@ -110,7 +110,7 @@ extern "C"
     R_input.swap(R_system);
 
     // We may need to correct a non-conforming solution
-    sys.get_dof_map().enforce_constraints_exactly(sys);
+    sys.get_dof_map().enforce_constraints_exactly(sys, sys.current_local_solution.get());
 
     // We may need to localize a parallel solution
     sys.update();
@@ -177,7 +177,7 @@ extern "C"
     J_input.swap(J_system);
 
     // We may need to correct a non-conforming solution
-    sys.get_dof_map().enforce_constraints_exactly(sys);
+    sys.get_dof_map().enforce_constraints_exactly(sys, sys.current_local_solution.get());
 
     // We may need to localize a parallel solution
     sys.update();


jwpeterson:
I'll invite you to our "Collaborators" group who get their PRs tested automatically.

jwpeterson:
@roystgnr and @pbauman can comment further, but it looks like we only have test coverage of `PetscDiffSolver` in adjoints_ex3.C and adjoints_ex5.C, and they don't do gold file comparisons, so if there are any subtle changes to the solutions due to your PR we might not catch them. There's definitely no coverage of `PetscDiffSolver` in MOOSE tests... maybe there is something in GRINS.

Also, you didn't say so, but I assume this PR fixes the issues in your app, which if I understand correctly only arise when PETSc is built with `--with-debugging=1`?

canesin:
Yes, this fixes my issue.
I was under the impression that the test system would hammer it, maybe I can add a test latter today.
I haven't tested with no debugging in PETSc (can I run GRINS test suit in this PR ?).

pbauman:
> There's definitely no coverage of PetscDiffSolver in MOOSE tests... maybe there is something in GRINS.

There's not, but there needs to be. This should be changing in the next 2 week or so (before the GRINS workshop at [Scientific Software Days](http://scisoftdays.org).

> I was under the impression that the test system would hammer it, maybe I can add a test latter today.

That would be amazing.

> I haven't tested with no debugging in PETSc (can I run GRINS test suit in this PR ?).

Yes, by building GRINS against this branch. But since GRINS doesn't have testing of this (yet), you would've had to have changed GRINS to use `PetscDiffSolver`.

And thanks for taking this on @canesin and sorry I haven't been able to be much help the past 6 weeks. I'm absolutely drowning right now.

roystgnr:
:+1: either after testing with GRINS or after determining that quickly testing with GRINS would be too difficult; this does look like the right fix to me.

jwpeterson:
This should also be cherry-picked to v1.2.0 when it's merged. I'll do a manual exodiff test with adjoints_ex3 and adjoints_ex5 and if that passes will probably merge...

canesin:
Thanks all, I tested in my application with all snes solvers. @jwpeterson sorry for not pushing the test. I cant find any exodiff based test in libMesh repository, if you guys could include in the examples a "reference test" the one could use as base for contributing would be awesome.

jwpeterson:
> I cant find any exodiff based test in libMesh repository

Right, this is the problem, we don't really have any. If this works to your satisfaction in your application that's a good step though. Does it use adaptivity?

jwpeterson:
I tested this by manually writing Exodus files from `adjoints_ex3` both before and after applying the patch and there were no diffs, so I'm satisfied. I should have a small PR momentarily to add optional Exodus output to adjoints_ex3 as well, since it might come in handy in the future...

PR #1305, opened 2017-04-06
Parallel::OpFunction, BoundingBox refactor

roystgnr:
Actual bounding box bugfixes (for relatively little-used cases, I admit) will be coming shortly, but the first step here turned out to be a rabbit hole that needs testing.  We can now define OpFunction<T> specializations which wrap MPI_Op_create to perform default reductions for user-defined types.


roystgnr:
I'll merge this tomorrow morning unless there are objections.

PR #1306, opened 2017-04-07
Bounding box accuracy for curved elements

roystgnr:
This PR fixes the MeshTools::create*bounding_box behavior for problems with curved elements, and adds a new create_local_bounding_box() method that is useful for refactoring and to fix some MOOSE bugs.

### DIFF
--- a/include/mesh/mesh_tools.h
+++ b/include/mesh/mesh_tools.h
@@ -152,6 +152,16 @@ bounding_sphere (const MeshBase & mesh);
 
 /**
  * @returns two points defining a cartesian box that bounds the
+ * nodes of the mesh.
+ *
+ * In the case of curved elements, this box might *not* bound the
+ * elements of the mesh.
+ */
+libMesh::BoundingBox
+create_nodal_bounding_box (const MeshBase & mesh);
+
+/**
+ * @returns two points defining a cartesian box that bounds the
  * elements belonging to the local processor.
  *
  * Unlike the other bounding box creation functions, this does *not*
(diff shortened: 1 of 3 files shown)


roystgnr:
This worked in MOOSE tests for me, but there may be a regression that GRINS caught; investigating now.

roystgnr:
Definitely a backwards-compatibility regression GRINS caught.  I'll see if I can work around it.

pbauman:
Dammit, I really need to get the CIVET recipe set up for testing against GRINS against libMesh PRs so @roystgnr doesn't have to keep doing it manually.

roystgnr:
I'd really appreciate it.  I was quaking here wondering whether it was really this PR that broke things or whether I'd just been failing to test the last half dozen PRs thoroughly enough...

On the other hand, NICE test coverage!  MOOSE has like 20 times as many tests as GRINS still, but about 10% of your tests caught the bug here and 0% of theirs did.

roystgnr:
Wow, and by "your tests" I mean *your* tests.  None of the tests I've added to GRINS or libMesh died either.

roystgnr:
Okay, I think I've fixed the problem; at least there's no more regression test failures and the fix makes sense.  I reordered commits so there won't be any broken states mid-PR.

I'm working a half day today before going on vacation for the weekend, so if this passes Civet I'd like to merge right away, then see if I can finish up a PR on MOOSE which depends on it.

roystgnr:
Wait, I just retested against the wrong build.  Not sure whether this fixed all the regressions or not.

roystgnr:
My tests are passing.  Unless something crops up I'll merge shortly.

Issue #1301, opened 2017-04-05
Assert tripped in adjoint_refinement_estimator


tradowsk:
I've discovered an issue in the adjoint refinement error estimator where [this assert](https://github.com/libMesh/libmesh/blob/master/src/error_estimation/adjoint_refinement_estimator.C#L455) gets tripped.

I have a test case on a GRINS branch that demonstrates this: https://github.com/tradowsk/grins/tree/adjoint-refinement-bug

In that test, I have a mesh with a single QUAD4 elem. In the error estimator, after the refine and then the coarsen, I get `mesh.n_elem() == 5` instead of 1.

I believe this results from my previous PR #1233. The guarding `if` statement I added there keeps `mesh.contract()` from being called in `ES::reinit()` when the `_refine_in_reinit` is flag is set to `false`, meaning the child elements stick around when they shouldn't.

I can fix this, but it is a matter of design choice. Do I remove the `if` guard around `mesh.contract()` in `ES::reinit()`, thereby making it be called every time? Or should we expect `contract()` to be called outside of `ES::reinit()`, in which case `adjoint_refinement_estimator` (and possibly more classes) would need to be updated? Or is there another option?

roystgnr:
In this case ARefEE internals were pretty much expected to be the "smart" side of the interaction, examining the system state and cleaning up after itself as necessary...

But I'm not actually sure what the "smart" thing to do here is, at least among achievable options.  The original goal was for ARefEE to restore the mesh to its original state after it was done, but that's not really possible when the mesh might have had some arbitrary subactive elements.  We can't save a whole copy of the subactive mesh state easily, but our only other alternatives are "contract(), and thereby wipe out any subactive elements the user wanted to keep" or "don't contract(), and thereby leave the mesh with more subactive elements than the user wanted, and only assert consistency of n_active_elem()".

I'd lean toward the latter when _refine_in_reinit is false, I think.

tradowsk:
Okay, I'll update the assert to only check the number of active elements then. Should have a PR tomorrow

PR #1308, opened 2017-04-07
Assert fix for AdjointRefinementErrorEstimator

tradowsk:
Addresses the problem raised in #1301 by only checking for the same number of active elements in ARefEE.

Passes my `GRINS` test case, as well as the larger run that originally identified this issue.


roystgnr:
Thanks for catching this!

PR #1309, opened 2017-04-07
Add optional ExodusII output to adjoints examples

jwpeterson:
This is a good first step if we ever want to convert these examples to exodiff-based regression tests, which I think we do want to do at some point for all the examples it makes sense for.


jwpeterson:
The original impetus for this was making sure the results were unchanged after the changes in #1304.

roystgnr:
Thanks!

PR #1310, opened 2017-04-10
-Wunused-parameter does not make sense as a C flag.

jwpeterson:
This reverts part of d58277d84e which turned out to be a bit overzealous.

### DIFF
--- a/configure
+++ b/configure
@@ -8143,9 +8143,9 @@ $as_echo "Unknown Intel complier" >&6; }
           CXXFLAGS_DBG="$CXXFLAGS_DBG -O0 -felide-constructors -g -pedantic -W -Wall -Wextra -Wno-long-long -Wunused-parameter -Wunused -Wpointer-arith -Wformat -Wparentheses -Qunused-arguments -Woverloaded-virtual -fno-limit-debug-info"
           NODEPRECATEDFLAG="-Wno-deprecated"
 
-          CFLAGS_OPT="-O2 -Qunused-arguments -Wunused-parameter -Wunused"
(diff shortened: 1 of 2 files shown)


roystgnr:
You can't have unnamed parameters in C, but if for some reason (compatibility with function pointer type?) you need to have unused parameters, you can still avoid the compiler warnings, most portably just by doing ```(void) unused_parameter_name;``` in the function body.

https://stackoverflow.com/questions/3599160/unused-parameter-warnings-in-c-code?rq=1

roystgnr:
I'd prefer suggesting users do that rather than reverting the warning flag, but I'm happy either way.

jwpeterson:
Hm... my idea was when bringing in a "contrib" package written in C to libmesh, we don't really want to see a bunch of unused parameter warnings *or* have to `(void) foo;` a bunch of variables to avoid them, so we might as well just not warn about them in the first place. 

Since people aren't regularly committing C code to libmesh itself, I don't think we need to be as vigilant about C warnings...

roystgnr:
For contrib packages whose headers we include, I think ignore_warnings.h is the way to go.  For people linking a little new C code to libMesh and swiping our compiler flags (a use case I've been in) I think having more warnings added is useful.  But for people linking a *lot* of old C code to libMesh and swiping our compiler flags I can see how it might be annoying.

Hmm... for compiling the contrib packages themselves, though... we don't want to either fork or pester upstream about every single warning, and it's a hassle to even edit our Makefiles to selectively disable warnings, so maybe taking particularly-overzealous warning flags out of our CFLAGS entirely really is the way to go.

PR #1311, opened 2017-04-12
Disallow remote element removal during Mesh copy

roystgnr:
Whether the copied mesh allows remote element removal or not, we
should make an exact copy first; we can then copy the copied mesh's
setting for later use.


jwpeterson:
This is somewhat unrelated, but I had to look at your diff an extra second or two to pick out the `prepare_for_use()` call there in the middle of all the flag setting/resetting. I don't know, maybe a newline or a comment would be helpful. Otherwise I don't have any constructive comments :)

jwpeterson:
👍

PR #1313, opened 2017-04-17
Update to error reporting in ReplicatedMesh::stitch_meshes.

dknez:
Throw runtime_error exceptions instead of calling libmesh_error_msg. This makes it easier to cleanly report the error when calling mesh stitching from external code, since it allows us to easily get the error message via e.what() when we catch the exception, whereas libmesh_error_msg prints the error msg to std::err.

### DIFF
--- a/src/mesh/replicated_mesh.C
+++ b/src/mesh/replicated_mesh.C
@@ -1068,7 +1068,11 @@ void ReplicatedMesh::stitching_helper (ReplicatedMesh * other_mesh,
                     Point other_point = other_iter->first;
 
                     if (!this_point.absolute_fuzzy_equals(other_point, tol*h_min))
-                      libmesh_error_msg("Error: mismatched points: " << this_point << " and " << other_point);
+                      {
+                        std::stringstream msg;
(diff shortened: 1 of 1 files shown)


dknez:
I agree. That sounds like a better fix to me.

dknez:
P.S. That would require us to add a msg argument to libMesh::LogicError. Are you OK with that?

Do you plan to look into this yourself, or would you prefer it if I make a new PR based on your suggestion?

dknez:
P.P.S. I tried making the change and it causes a bunch of compilation errors due to code in the library like this:

```
libmesh_error_msg("Non-blocking packed range sends cannot exceed " << std::numeric_limits<int>::max() << "in size");
```

roystgnr:
That kind of code is definitely supposed to be supported.  Rather than assume the argument is a string, we'd need to create a local strstream and feed the argument to it.

We want to continue to support the existing LogicError constructor, but a new constructor with a msg argument would be great.

dknez:
See #1314.

dknez:
Closing this PR. Use #1314 instead.

PR #1314, opened 2017-04-17
Store the msg passed to libmesh_error_msg inside the exception object.

dknez:


roystgnr:
:+1: as far as I'm concerned; let's give others a day or so to look at it before merging.

jwpeterson:
:+1:

dknez:
Sounds like we're in agreement on this one. I'll go ahead and merge soon.

PR #1315, opened 2017-04-18
Remove nullptr checks from BoundaryInfo::sync().

jwpeterson:
This makes it possible to build BoundaryMeshes from internal sidesets,
a feature which did not work previously.  The decision of whether to
add a side to the BoundaryMesh should not be governed by whether an
Elem is on a *geometric* boundary, only by the information in the
_boundary_side_id container.

The relevant changes are in the helper functions:
* BoundaryInfo::_find_id_maps().
* BoundaryInfo::add_elements().

We are also able to hoist the multimap lookup outside of the loop over
sides, since we can get all the boundary information for a given
element's sides in a single call to equal_range(). Otherwise, the
majority of the diff is just the reindentation which accompanied the
removal of an outer if-statement.

### DIFF
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -121,6 +121,20 @@ public:
              UnstructuredMesh & boundary_mesh);
 
   /**
+   * Like the other sync() implementations, but specifically intended
+   * for building "boundary" meshes from internal sidesets. In the
+   * case of an internal sideset, each side may belong to 2
+   * higher-dimensional parent elements, and typically we do not want
+   * to add the same side to the boundary mesh twice. The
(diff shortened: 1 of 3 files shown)


jwpeterson:
Hmm... I guess there is at least a unit test in libmesh relying on the original behavior. I will take a look at this.

jwpeterson:
Since we still need to support the case where no boundary ids are requested, i.e.
```
vol_mesh.get_boundary_info().sync(surf_mesh);
```
there still needs to be a check for a `NULL` neighbor in that case. This should be fixed by 73526cd.

jwpeterson:
Also testing this against MOOSE apps just in case there is some non-standard use of `BoundaryInfo::sync()` there that I'm not aware of.

Issue #1316, opened 2017-04-19
Memory leak detected when there appears to be no leak?


dknez:
I ran a job which printed the reference count info below. As you can see, it reports a memory leak, but all the Creation/Destruction counts match up. I added a print statement
```
std::cout << "n_objects: " << ReferenceCounter::n_objects() << std::endl;
```
and that printed 4294967264, as you can see below.

```
 ---------------------------------------------------------------------------- 
| Reference count information                                                |
 ---------------------------------------------------------------------------- 
| 14RBRefComponent reference count information:
|  Creations:    2
|  Destructions: 2
| 20RBComponentDofObject reference count information:
|  Creations:    3
|  Destructions: 3
| 23RBComponentConnectivity reference count information:
|  Creations:    1
|  Destructions: 1
| N7libMesh10FEAbstractE reference count information:
|  Creations:    290
|  Destructions: 290
| N7libMesh10Parameters5ValueE reference count information:
|  Creations:    2
|  Destructions: 2
| N7libMesh12LinearSolverIdEE reference count information:
|  Creations:    1
|  Destructions: 1
| N7libMesh12SparseMatrixIdEE reference count information:
|  Creations:    2
|  Destructions: 2
| N7libMesh13NumericVectorIdEE reference count information:
|  Creations:    8
|  Destructions: 8
| N7libMesh15EquationSystemsE reference count information:
|  Creations:    1
|  Destructions: 1
| N7libMesh15GhostingFunctorE reference count information:
|  Creations:    5
|  Destructions: 5
| N7libMesh19SolverConfigurationE reference count information:
|  Creations:    1
|  Destructions: 1
| N7libMesh4ElemE reference count information:
|  Creations:    6932
|  Destructions: 6932
| N7libMesh4NodeE reference count information:
|  Creations:    1587
|  Destructions: 1587
| N7libMesh5QBaseE reference count information:
|  Creations:    279
|  Destructions: 279
| N7libMesh6DofMapE reference count information:
|  Creations:    1
|  Destructions: 1
| N7libMesh6SystemE reference count information:
|  Creations:    1
|  Destructions: 1
| N7libMesh9DofObjectE reference count information:
|  Creations:    8519
|  Destructions: 8519
 ---------------------------------------------------------------------------- 
n_objects: 4294967264
Memory leak detected!
```

permcody:
Well that's some sort of overflow. That number is unsigned max!

jwpeterson:
[...]

PR #1317, opened 2017-04-20
Add incrementing copy constructor to ReferenceCounter class.

jwpeterson:
Refs #1316.
Reported-by: David Knezevic <david.knezevic@akselos.com>

### DIFF
--- a/include/base/reference_counter.h
+++ b/include/base/reference_counter.h
@@ -46,11 +46,12 @@ class ReferenceCounter
 protected:
 
   /**
-   * Constructor. Protected so that you cannont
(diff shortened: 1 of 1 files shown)


jwpeterson:
This fixes the problem demonstrated by the test code in #1316 for me. 

I would definitely like to cherry-pick this commit onto v1.2.0 as well. It could even be backported to previous releases, but I think most people don't encounter it because we actually disable copying for most of the reference-counted classes (`Elem`, `NumericVector`, etc.).

dknez:
:+1:

PR #1319, opened 2017-04-21
Make sure reference counting works with move constructors 

jwpeterson:
While the bug fix in 45d1390 was still fresh on my mind, I wanted to be sure that we didn't run into the same problem when we started adding move constructors. This PR adds a configure test for the C++11 features required to declare move constructors and a `LIBMESH_HAVE_CXX11_MOVE_CONSTRUCTORS` macro, so everything is still C++03-compatible. I tested that the reference counts are still correct even when move constructing objects with the relatively simple `RBTheta` class.


permcody:
Seems like you won't really need that reference count base class after
C++11, right?

jwpeterson:
Why not?

permcody:
I guess I'm thinking forward to the day where all objects that might leak
(i.e. objects built by the library and returned to the user) are all
managed through C++11 smart pointers.

jwpeterson:
Ah, I see what you mean. The `ReferenceCountedObject` stuff (which should really be named `ConstructorCallCountedObject` FWIW) is intended to be a bit more low level than the true reference counting provided by the std library. For instance, it can catch the situations where someone purposefully circumvents the smart pointer protections, e.g.
```
NumericVector<Number> * foo = NumericVector<Number>::build(_communicator).release();
```
and then later leaks `foo`, since the number of constructions/destructions won't agree in that case.

roystgnr:
This seems to be causing failures for me, in operator= for an Epetra matrix, because the ReferenceCountedObject base class is no longer copy-assignable.  The automatic copy assignment operator is only generated if no move constructor and no move assignment operator are defined.

jwpeterson:
Huh, sorry for the unintended consequence. I did not know that C++11 rule...

PR #1321, opened 2017-04-25
Fix Tri3::contains_point().

jwpeterson:
The "barycentric coordinates" part of the algorithm is copied from
http://blackpawn.com/texts/pointinpoly. This part assumes that the
point in question already lies in the plane of the three points
defining the triangle, and returns false positives when it does not,
so we have augmented it with a normalized out-of-plane test. The new
code seems to pass the unit tests suggested by @benkirk in #993, but I
have little doubt that it could still be much better.

The right approach is probably to use so-called robust geometric
predicates with adaptive-precision floating-point arithmetic, but that
invovles quite a bit more code complexity...

### DIFF
--- a/src/geom/face_tri3.C
+++ b/src/geom/face_tri3.C
@@ -230,7 +230,7 @@ bool Tri3::contains_point (const Point & p, Real tol) const
   Real v = (dot00 * dot12 - dot01 * dot02) * invDenom;
 
   // Check if point is in triangle
-  return (u >= 0) && (v >= 0) && (u + v <= 1);
+  return (u > -tol) && (v > -tol) && (u + v < 1 + tol);
 }
 
 } // namespace libMesh
--- a/src/geom/face_tri3.C
+++ b/src/geom/face_tri3.C
@@ -205,48 +205,32 @@ std::pair<Real, Real> Tri3::min_and_max_angle() const
 
 bool Tri3::contains_point (const Point & p, Real tol) const
 {
-  // move test point relative to node 0
-  const Point p0 ( p - this->point(0) );
+  // See "Barycentric Technique" section at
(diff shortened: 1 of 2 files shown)


jwpeterson:
I am still planning to update the unit tests with the small triangle tests suggested by @benkirk, but I figured I would at least let the CI determine if anything else got broken in the process...

src/geom/face_tri3.C
```diff
@@ -205,48 +205,32 @@ std::pair<Real, Real> Tri3::min_and_max_angle() const
...
+  // Compute barycentric coordinates
+  Real invDenom = 1 / (dot00 * dot11 - dot01 * dot01);
+  Real u = (dot11 * dot02 - dot01 * dot12) * invDenom;
+  Real v = (dot00 * dot12 - dot01 * dot02) * invDenom;
 
-  const Real s2 = (d01 * d22 - d02 * d12) * d;
-  return s2 > -tol && s1 + s2 < 1.0 + tol;
+  // Check if point is in triangle
+  return (u >= 0) && (v >= 0) && (u + v <= 1);
 }
```
roystgnr: Shouldn't we be using tol at this point?

jwpeterson: Yeah, just pushed a new commit changing that. I realized it after the first push, but wanted to see if any tests actually depended on it... and they did apparently.

benkirk:
how about we cherry pick this for 1.2 as well?


On Apr 25, 2017, at 2:15 PM, John W. Peterson <notifications@github.com<mailto:notifications@github.com>> wrote:


Merged #1321<https://github.com/libMesh/libmesh/pull/1321>.

roystgnr:
Seconded.

jwpeterson:
OK

PR #1322, opened 2017-04-25
Print out filename when we die with a file error

roystgnr:
This makes the debugging process a bit shorter


jwpeterson:
New internet acronym: DWAFE

permcody:
Still not as good as DRYP.

PR #1324, opened 2017-04-25
Use geometry, not ids, in BoundaryMesh test

roystgnr:
This fixes unit test failures on distributed meshes for me.


jwpeterson:
Thanks for fixing this.

Issue #1199, opened 2016-12-27
HDF5 >1.8.12 support for bundled netCDF library


barnabecsek:
Hello,

I was wondering whether it would be possible to update the netcdf version that comes with libmesh. For HDF5 libraries newer than 1.8.12 the function 'H5Pset_fapl_mpiposix' has been removed and consequently I get a linker error. I know this issue has been resolved in newer netcdf libraries:

https://github.com/Unidata/netcdf-c/commit/435d8a03ed28bb5ad63aff12cbc6ab91531b6bc8

I am working on a project that uses libmesh and unfortunately I do not have access to older HDF5 libraries than 1.8.16 on the system that I am using.
Thank you,

Barna

jwpeterson:
We are definitely planning to do this.  Just hasn't been a high priority since not many of us use hdf5, and i'm not really sure exactly how to do the netcdf upgrade since I have not done one before.

jwpeterson:
This should be fixed as of #1326.

PR #1326, opened 2017-04-26
NetCDF 4.4.1.1 refresh

benkirk:
Updates our bundled NetCDF to 4.4.1.1.  Should address #1199 

capturing @roystgnr comment on this thread for considerartion

> If you really want a challenge: I'd prefer being able to link to external NetCDF over just another refresh of our own.  It's nice to have a fallback version in contrib, sure, but it's no longer unheard of to see a system-installed NetCDF, in which case IIRC our only current solution is "pray it's either ABI-compatible with our version or it's not in the same library path as something else we need".


jwpeterson:
350k lines, wow. LibMesh src + include is around 185k.

benkirk:
yes, this is a beast.  Roy's suggestion is a good one, although we have to change 2 preprocessor values 
https://github.com/libMesh/libmesh/pull/1326/commits/b7142a6b62ae80312ea2e0526ab437775a896812

Otherwise using an external install should be pretty easy.

jwpeterson:
> we have to change 2 preprocessor values

Huh, doesn't this pretty much rule out the possibility of us using vanilla NetCDF installs?!

jwpeterson:
Since this is a low-level change, I attached some additional MOOSE based application tests to the PR. If all those pass, I'd say merge whenever you are ready.

benkirk:
So does this mean the additional tests have passed now, or are those run elsewhere?

jwpeterson:
> So does this mean the additional tests have passed now, or are those run elsewhere?

They passed (they are the ones called "App tests" and "External App tests" in the list of checks.

mboisson:
It does not look like this was merged in the latest release (1.2.0). When will the next release be ?

jwpeterson:
We are on a fairly loose schedule of two releases/year, so the next one will probably be some time toward the end of 2017. In the meantime you could try master.  There have not been many major changes (other than the NetCDF update) since the 1.2.0 release so it would most likely work for you.

mboisson:
Mmmm, I'm installing it for a cluster. Installing a "master" (i.e. un-numbered) version will be frowned-upon...

roystgnr:
Typically we put out two releases a year.  "Sooner rather than later" might be warranted for an update that's blocking compatibility with some systems, though, huh?  I've got some intrusive changes that I'd planned to put in next month, and I'd thought we'd wait to shake the bugs off those before forking a 1.3.0 branch (which would then end up coming out around November), but perhaps we should fork beforehand instead?

roystgnr:
If you're setting up an installation now, I'd call your installed version "libmesh-20070724" and build master, I'm afraid.  Even if we forked off a new branch *right now*, it'd be in testing for a month before we'd feel comfortable not calling it "1.3.0-pre".

mboisson:
1.3.0-pre / 1.3.0-alpha or something similar would work better. A version 20170724 would confuse the version detection of our module system (i.e. it would not know which of 1.3.0 or 20170724 is the most recent).

I guess I could label it internally as 1.3.0-pre, or 1.2.1 ?

roystgnr:
We definitely don't want to call our current version 1.2.1 (which suggests that it's merely bug fixes backported to 1.2.0), or 1.3.0-pre (which will be the "official" prerelease later)... 1.3.0-alpha sounds reasonable.

benkirk:
So actually John and I traded some emails on this for the OpenHPC stuff I was doing.

This change was back ported to 1.2 on John’s jwpeterson/1.2.1.

I actually propose we release a 1.2.1 with this change; there’s been no issues building the new netcdf I have encountered.

-Ben



On Jul 24, 2017, at 3:37 PM, roystgnr <notifications@github.com<mailto:notifications@github.com>> wrote:


We definitely don't want to call our current version 1.2.1 (which suggests that it's merely bug fixes backported to 1.2.0), or 1.3.0-pre (which will be the "official" prerelease later)... 1.3.0-alpha sounds reasonable.

roystgnr:
Huh.  An official 1.2.1 does sound like a good idea.  Is there anything else that ought to be back ported while we're at it?  We can probably skip my slew of DistributedMesh corner case fixes; they're important but they're too complicated for a subminor version increment IMHO.

jwpeterson:
> This change was back ported to 1.2 on John’s jwpeterson/1.2.1.

jwpeterson/1.2.1 was created by a merge of the NetCDF update branch, jwpeterson/1.2.1-cherrypick was created by simply cherry-picking commits from the NetCDF update branch. I'd recommend we use the cherrypick one since it did not generate any conflicts, while the merge one did (and brought in some extraneous commits as well).

benkirk:
Thanks for the reminder & clarification. There’s nothing else I saw that needs to get back ported into  1.2.1, although browsing the logs there are some small changes for things like 64-bit fixes (#1361) that would be easy.

jwpeterson:
> Is there anything else that ought to be back ported while we're at it?

Definitely, it just depends on how much time someone wants to spend making sure all the cherry-picked stuff works nicely together. Some candidates:

3df15f5 Add empty LIBMESH_PARALLEL_FLOAT_OPS macro for --disable-mpi case.
aee0b3e 64-bit build fix
25154f3 solve PETSc lock
fe9b839 Fix for upcoming MatGetSubMatrix name change in PETSc 3.8.
4e0bb43 Add support for Intel 2017 compilers.
7194f06 Added support to configure for Intel 2017 compilers.

@roystgnr I don't know exactly which of your commits you think should be cherry-picked, but if you fetch my branch and run:
```
git log --graph --oneline --decorate jwpeterson/1.2.1-cherrypick..upstream/master
```
you can get a summary of all the commits that have been made in the meantime...

jwpeterson:
@benkirk and @roystgnr, PR #1384 is now testing the set of commits cherry-picked onto 1.2.0. If everything passes to our satisfaction, I will tag the HEAD of the 1.2.1 branch as "v1.2.0" and make the release, tarballs, etc...

jwpeterson:
> Mmmm, I'm installing it for a cluster. Installing a "master" (i.e. un-numbered) version will be frowned-upon...

@mboisson Can you try the "v1.2.1" tag and see if it works for your cluster installation?

mboisson:
Will do tomorrow, thanks.
  

  
  

  
  
>

mboisson:
I launched a build tonight since it is quite long. Will see later what gives.

mboisson:
I confirm that the internal netcdf builds fine with our version of HDF5.

PR #1328, opened 2017-04-27
Uncommented constrain_element_matrix in eigenproblems_ex3.

dknez:


roystgnr:
Thanks!

I'm kind of surprised we hadn't done this already, though.  I seem to recall a frustrating period in 2007 trying to debug a friend's inaccurate adaptive libMesh flow code, only to discover that they'd started it by modifying one of our non-AMR examples and they'd added AMR without adding constraint calls; at that point we went through all the non-adaptive examples and added bits like:

```
      // If this assembly program were to be used on an adaptive mesh,
      // we would have to apply any hanging node constraint equations
      dof_map.constrain_element_matrix_and_vector (Ke, Fe, dof_indices);
```

on the theory that fractions of a second of wasted CPU time were cheaper than man-weeks of wasted researcher time.

dknez:
Is this a real failure? Seems unlikely...

jwpeterson:
No, just debug mode timeouts. And we aren't actually testing the change you made since there are no SLEPc targets on libmesh PRs...

PR #1332, opened 2017-05-03
Fix warnings/errors in non-standard configurations

jwpeterson:
I have now added optional "No AMR", "No Exceptions", "No Exodus", "Infinite Elements", "No MPI/No PETSc" and "No Optional" to CIVET and attached them to this PR.

~~It should still fail the No MPI test for the time being, since I haven't looked into fixing that yet.~~

### DIFF
--- a/CITATION
+++ b/CITATION
@@ -1,15 +1,14 @@
 Please use the following reference to cite libMesh:
 
 @Article{libMeshPaper,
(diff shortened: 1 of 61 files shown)


jwpeterson:
Looks like it failed (therefore it worked).

Issue #1333, opened 2017-05-04
Would like erf function in contrib/fparser


lindsayad:
Going to try implementing this myself...

jwpeterson:
I will generalize the existing `LIBMESH_TEST_CXX11_INVERSE_HYPERBOLIC_FUNCS` configure test in cxx11.m4 to test for `std::erf`. You can just assume that `std::erf` is available while you are doing the Fparser development.

pbauman:
Thanks @jwpeterson you read my mind.

lindsayad:
Great. Thanks!

lindsayad:
Closed by #1336

PR #1334, opened 2017-05-04
Add configure test for std::erf

jwpeterson:
This will eventually be used by #1333.


jwpeterson:
Good old timeout failure in `patterened_mesh.patterned_generation` test. This PR should be good to go.

PR #1336, opened 2017-05-05
Add error function to FunctionParser

lindsayad:
### Current issue

Just in time compilation of this file fails
```cpp
#define _USE_MATH_DEFINES
#include <cmath>
extern "C" double f_f69e440ab47bf78583454b09daa59af1d393097f(const double *params, const double *immed, const double eps) {
double r, s[1];
s[0] = params[0];
s[0] = std::erf(s[0]);
return s[0]; }
```
with the error message `no member named 'erf' in namespace 'std'; did you mean simply 'erf'`. I don't understand why it's failing...`cmath` is included. If I compile independently with `g++ -O2 -shared -rdynamic -fPIC test_compile.cpp -o test_compile`, there's no error.

### Type (float, double, long double) conventions

In my template specializations of `fp_erf` for complex numbers I'll defer to suggestions on how I should do the typing of constants.

### Bigger picture

I'll revert the `autodiff.cc` test before this ever gets merged, but I am seeing the correct function and derivative output in the `.dat` files.

Pinging @dschwen perhaps for review


dschwen:
@lindsayad, this looks good so far. I have to admit that I haven't given much thought on how to get complex support into the JIT code, and I'm wondering why the compilation does not work. You could look at the `FPARSER_JIT_COMPILER` define in the libmesh config.h file. Also look at `FP_SUPPORT_CPLUSPLUS11_MATH_FUNCS`; I don't recall why I needed to put that in there though...

permcody:
@dschwen - with these edits going straight into contrib source, do we have a plan on how we might upgrade this package? I'd think we'd at least like to keep the new methods that have been added in separate source files. Just posing the question.

dschwen:
`fparser_ad.(hh|cc)` is all our stuff, it does not exist in upstream fparser. Development of fparser has been quite slow anyways. Upgrading will not be super trivial (I _did_ make some changes in the upstream code).

jwpeterson:
> with these edits going straight into contrib source, do we have a plan on how we might upgrade this package?

autodiff.cc and fparser_ad.cc are @dschwen's new files, so they most likely would not conflict with an update.

We have offered our changes to upstream and had them rejected/ignored, so the only way to update would be to investigate the diffs line by line... not fun.

jwpeterson:
Is there any way to see what command the JIT compiler is using? Is it definitely calling a C++ compiler?

dschwen:
> Is there any way to see what command the JIT compiler is using?

Yes, read my comment above

permcody:
ECHO, echo, echo...

BTW - The JIT fallback stuff may not be working quite right. Stop by and talk to @milljm when you have some time.

lindsayad:
Ok, here's what the JIT compiler ran:

`ccache clang++ -O2 -shared -rdynamic -fPIC test_compile.cpp -o test_compile`

That produces the `no member named 'erf' in namespace 'std'`.  `g++` works fine. So looks like I have a system issue...

For the record, if I run
```bash
grep --include=\*.{C,h,c,cpp,hpp,hh,cc} -rn FPARSER_JIT_COMPILER`
```
from the libmesh root directory, my output is:
```
contrib/fparser/fparser_ad.cc:1045:  std::string command = FPARSER_JIT_COMPILER" -O2 -shared -fPIC ";
contrib/fparser/fparser_ad.cc:1047:  std::string command = FPARSER_JIT_COMPILER" -O2 -shared -rdynamic -fPIC ";
```
So `libmesh_config.h` didn't give me any information...I had to go into the debugger to figure out it was using `clang++`

jwpeterson:
It probably needs to pass `-std=c++11`? What happens if you run that command by hand in the shell?

lindsayad:
Ah yes, passing `-std=c++11` makes it work at least by hand in the shell. I'll try with the `autodiff.cc` example...

dschwen:
Ugh, I think I might be setting that define in a makefile somehow. I'm in a meeting right now. But I can investigate later.

lindsayad:
Yep, compiling with
```cpp
  std::string command = FPARSER_JIT_COMPILER" -O2 -shared -rdynamic -fPIC -std=c++11 ";
```
in `fparser_ad.cc` makes it work. Should I commit that change or should it be built into `FPARSER_JIT_COMPILER`?

dschwen:
Does libmesh require C++11 support? Does that switch work with all supported compilers?

jwpeterson:
> Does libmesh require C++11 support? 

Not yet, no. Ideally you should be getting the compiler and flags from the libmesh build system, and guarding this particular function with `LIBMESH_HAVE_CXX11_ERF`.

> Does that switch work with all supported compilers?

That I know of, yes.

lindsayad:
@dschwen @jwpeterson I think this is ready for more review.

lindsayad:
Here's the output of the erf autodiff example in case anyone's curious:
```
Input Expression:
erf(x)
Size of stack: 1
0000: push Var0     = [0]x
0001: f:erf (1)     = [0]f:erf(x)

Optimized Input Expression:
erf(x)
Size of stack: 1
0000: push Var0     = [0]x
0001: f:erf (1)     = [0]f:erf(x)

AutoDiff returned -1
Unsimplified derivative:
Size of stack: 3
0000: push 1.1283792= [0]1.1283792
0001: push Var0     = [1]x
0002: sqr           = [1]x^2
0003: neg           = [1](-(x^2))
0004: exp           = [1]exp((-(x^2)))
0005: mul           = [0]1.1283792*exp((-(x^2)))

Simplified derivative:
Size of stack: 3
0000: push 1.1283792= [0]1.1283792
0001: push Var0     = [1]x
0002: sqr           = [1]x^2
0003: neg           = [1](-(x^2))
0004: exp           = [1]exp((-(x^2)))
0005: mul           = [0]1.1283792*exp((-(x^2)))

AutoDiff 2nd time returned -1
Simplified second derivative:
Size of stack: 3
0000: push -2.2567583= [0]-2.2567583
0001: push Var0     = [1]x
0002: mul           = [0]-2.2567583*x
0003: push Var0     = [1]x
0004: sqr           = [1]x^2
0005: neg           = [1](-(x^2))
0006: exp           = [1]exp((-(x^2)))
0007: mul           = [0]-2.2567583*x*exp((-(x^2)))
```

dschwen:
What is in `CXXFLAGS`? Looking at my config.log it only contains the `-std=c++11` flag. All method specific stuff is in `CXXFLAGS_OPT` etc. 
Or is that folded in at this point in the make process?

lindsayad:
No I believe you're spot on, for  me `CXXFLAGS`  only provides `-std=c++11`. I thought about providing a method and thus more flags,  but we're already providing optimization flags etc. in the source file for `FPARSER_JIT_COMPILER`.

dschwen:
This looks good to me, but why don't you go ahead revert the example and instead put in a unit test in https://github.com/libMesh/libmesh/blob/master/tests/fparser/autodiff.C

jwpeterson:
@lindsayad I started the CI manually and sent you an invite to join the libmesh group whose PRs get tested automatically.

contrib/fparser/fparser_ad.cc
```diff
@@ -111,6 +114,95 @@ Value_t FunctionParserADBase<Value_t>::fp_plog(const Value_t * params)
   return x < a ? fp_log(a) + (x-a)/a - (x-a)*(x-a)/(Value_t(2)*a*a) + (x-a)*(x-a)*(x-a)/(Value_t(3)*a*a*a) : fp_log(x);
 }
 
+template <typename Value_t>
+Value_t FunctionParserADBase<Value_t>::fp_erf(const Value_t * params)
+{
+  const Value_t x = params[0];
+  return std::erf(x);
```
jwpeterson: Probably I am missing something very obvious, but where are the C++11 ifdefs guarding this call to `std::erf`?

contrib/fparser/fparser_ad.cc
```diff
@@ -111,6 +114,95 @@ Value_t FunctionParserADBase<Value_t>::fp_plog(const Value_t * params)
...
+std::complex<double>
+FunctionParserADBase<std::complex<double>>::fp_erf(const std::complex<double> * params)
+{
+  const std::complex<double> z = params[0];
+  double x = z.real(), y = z.imag();
+  double fx = std::erf(x) + std::exp(-(x*x)) / (2. * M_PI) * (1. - std::cos(2. * x * y));
+  double fy = std::exp(-(x*x)) / (2. * M_PI) * std::sin(2. * x * y);
+
+  // Set this to as high as you want for greater accuracy
+  unsigned n = 3;
```
jwpeterson: According to the website you linked, this part of the formula should actually be an infinite series! 

What are the ramifications of truncating the series after an arbitrary n=3 terms, i.e. how quickly does the series converge and what is the error as a function of `n`? The algorithm in the linked question states that the number of terms needed to reach a desired tolerance is "fairly sensitive to" the argument, z.

I think at the very least there should be a tolerance (either hard-coded into the function or something specified by the user) that controls the number of iterations, possibly with a `max_its` to avoid infinite looping... there is even an algorithm in the code you linked.

contrib/fparser/fparser_ad.cc
```diff
@@ -111,6 +114,95 @@ Value_t FunctionParserADBase<Value_t>::fp_plog(const Value_t * params)
...
+    fx += mult * fxk;
+    fy += mult * fyk;
+  }
+
+  std::complex<float> f (fx, fy);
+  return f;
+}
+
+template<>
+std::complex<long double>
```
jwpeterson: The copy/pasted `float` and `long double` versions of the algorithm seem to unnecessarily duplicate code. Can you not achieve the same thing with explicit instantiations?

jwpeterson:
You may be able to just use [Steven Johnson's implementation](http://ab-initio.mit.edu/wiki/index.php/Faddeeva_Package) directly. His API takes a user-defined relative error tolerance, but I have not looked at the implementation in detail. It is open source and (permissive) MIT-licensed...

contrib/fparser/fparser_ad.cc
lindsayad: Sorry, I don't fully understand explicit instantiation. Could you demonstrate what this would look like here?

jwpeterson: Take a look at the `fp_plog` function, for instance.  There is only one (generic) version of it defined, and it gets instantiated for the various complex types at the bottom of the file by these macro calls:
```
#ifdef FP_SUPPORT_COMPLEX_DOUBLE_TYPE
FUNCTIONPARSERAD_INSTANTIATE_CLASS(std::complex<double>)
#endif

#ifdef FP_SUPPORT_COMPLEX_FLOAT_TYPE
FUNCTIONPARSERAD_INSTANTIATE_CLASS(std::complex<float>)
#endif

#ifdef FP_SUPPORT_COMPLEX_LONG_DOUBLE_TYPE
FUNCTIONPARSERAD_INSTANTIATE_CLASS(std::complex<long double>)
#endif
```
etc, i.e. it's not repeated for each of the different types the way you've done your PR. I think your case is a bit different since you need different versions for real vs. complex scalars, but all of the `complex<T>` should be the same.

lindsayad: It was those instantiations that led me to code the complex method specializations in the first place. (Out of curiosity, do we actually have people using function parser with complex numbers?). If I comment out the `long double` and `float` specializations, then I get the error:
```
../../../contrib/fparser/fparser_ad.cc:123:10: error: no matching function for call to 'erf'
  return std::erf(x);
         ^~~~~~~~
../../../contrib/fparser/fparser_ad.cc:1203:16: note: in instantiation of member function 'FunctionParserADBase<std::complex<long double> >::fp_erf' requested here
template class FunctionParserADBase< std::complex<long double> >; template class ADImplementation< std::complex<long double> >;
               ^
/usr/include/x86_64-linux-gnu/bits/mathcalls.h:259:15: note: candidate function not viable: no known conversion from 'const std::complex<long double>' to 'double' for 1st argument
extern double erf (double) throw (); extern double __erf (double) throw ();
              ^
/opt/moose/gcc-6.2.0/lib/gcc/x86_64-pc-linux-gnu/6.2.0/../../../../include/c++/6.2.0/cmath:1306:3: note: candidate function not viable: no known conversion from 'const std::complex<long double>' to 'float' for 1st argument
  erf(float __x)
  ^
/opt/moose/gcc-6.2.0/lib/gcc/x86_64-pc-linux-gnu/6.2.0/../../../../include/c++/6.2.0/cmath:1310:3: note: candidate function not viable: no known conversion from 'const std::complex<long double>' to 'long double' for 1st argument
  erf(long double __x)
  ^
/opt/moose/gcc-6.2.0/lib/gcc/x86_64-pc-linux-gnu/6.2.0/../../../../include/c++/6.2.0/cmath:1317:5: note: candidate template ignored: substitution failure [with _Tp = std::complex<long double>]: no type named '__type' in '__gnu_cxx::__enable_if<false, double>'
    erf(_Tp __x)
```

jwpeterson:
> It was those instantiations that led me to code the complex method specializations in the first place.

OK, if you don't actually need erf(z) for complex numbers, we can skip improving that implementation for now... maybe just have it throw an error if you want.



Something like this
```
template<>
std::complex<T>
FunctionParserADBase<std::complex<T>>::fp_erf(const std::complex<T> * params)
{
/* complex code */
}
```
should be possible I think, but off the top of my head I'm not sure. I will see if I can cook up an example.

lindsayad:
@jwpeterson:

So I went ahead and used the Faddeeva methods from Johnson. The `complex<float>` and `complex<long double>` code is still kind of ugly, but I don't know how to fix it without template work in the `Faddeeva` source where it only takes `double` type. It looks to me like the include guards are pretty good in `Faddeeva` but then again I'm only an intermediate c++ programmer. i would say that if you don't like this newest commit, I'll revert to the last commit and then just throw errors for complex numbers like you suggested.

jwpeterson:
Thanks for digging into the Faddeeva stuff. If we don't need the `erfc`, `erfi`, and `Dawson` functions, we might be able to avoid adding quite so much new code. Also, the various `complex<T>` implementations are definitely improved, but I think I can streamline them just a bit more... I'll send you a PR for this branch if I come up with anything useful.

jwpeterson:
@lindsayad if you merge and push my [PR](https://github.com/lindsayad/libmesh/pull/1) to your branch this should start testing again automatically...

contrib/fparser/fparser_ad.cc
```diff
@@ -843,6 +877,17 @@ bool FunctionParserADBase<Value_t>::JITCompileHelper(const std::string & Value_t
           // --sp; ccfile << "s[" << sp << "] = s[" << sp << "] < s[" << (sp+1) << "] ? std::log(s[" << (sp+1) << "]) - 1.5 + 2.0/s[" << (sp+1) << "] * s[" << sp << "] - 0.5/(s[" << (sp+1) << "]*s[" << (sp+1) << "]) * s[" << sp << "]*s[" << sp << "] : std::log(s[" << sp << "]);\n";
           --sp; ccfile << "s[" << sp << "] = s[" << sp << "] < s[" << (sp+1) << "] ? std::log(s[" << (sp+1) << "])  +  (s[" << sp << "]-s[" << (sp+1) << "])/s[" << (sp+1) << "]  -  std::pow((s[" << sp << "]-s[" << (sp+1) << "])/s[" << (sp+1) << "],2.0)/2.0  +  std::pow((s[" << sp << "]-s[" << (sp+1) << "])/s[" << (sp+1) << "],3.0)/3.0 : std::log(s[" << sp << "]);\n";
         }
+        else if (function == mFErf)
+        {
+#if LIBMESH_HAVE_CXX11_ERF
```
jwpeterson: Crap, we should drop this guard now since our `erf` does not require C++11. Would you mind adding one more commit for this?

lindsayad: This is the JIT compiled stuff though, and it *is* currently using `std::erf` (per line below). In order to use the Faddeeva methods in the JIT compiled file we would need to add the include of the Faddeeva header and make sure the JIT compiler also compiles the Faddeeva source, which I suppose is not hard.

contrib/fparser/fparser_ad.cc
lindsayad: This isn't really relevant now that we're using the Faddeeva stuff, but I'm pretty sure I figured out how I could create a single templated method for real calculations and another single templated method for complex calculations. The source would be like this I think:
```cpp
template<typename T>
template<typename U>
std::complex<U> FParserStuff<T>::fp_erf(const std::complex<U> x)
{
  /* Complex methods */
}

template<typename T>
template<typename U>
U FParserStuff<T>::fp_erf(const U x)
{
  /* Real methods */
}

template class FParserStuff<float>;
template class FParserStuff<std::complex<float>>;
template float FParserStuff<float>::fp_erf(float);
template std::complex<float> FParserStuff<std::complex<float>>::fp_erf(std::complex<float>);
```
I was concerned that the latter defined method would get called for `std::complex<float>` arguments but I checked in a debugger and the former method does get called (regardless of the ordering of the definitions). I guess the compiler chooses the more specific method?...

contrib/fparser/fparser_ad.cc
jwpeterson: Oh, I see now. Yeah, this should stay as-is then. Since we are not installing Faddeeva.hh, JIT-compiled codes can't #include it...

contrib/fparser/fparser_ad.cc
jwpeterson: Th double template parameter looks like a reasonable approach. I played around with this a bit the other day but did not hit on this particular idea. We might need it for the next function that gets added to fparser, though.

PR #1342, opened 2017-05-11
Fem system assembly with no constraints

vikramvgarg:
Assemble an FEMSystem with an option to not apply any constraints.

### DIFF
--- a/include/systems/diff_system.h
+++ b/include/systems/diff_system.h
@@ -123,7 +123,8 @@ public:
    */
   virtual void assembly (bool get_residual,
                          bool get_jacobian,
-                         bool apply_heterogeneous_constraints = false) libmesh_override = 0;
+                         bool apply_heterogeneous_constraints = false,
+                         bool apply_no_constraints = false) libmesh_override = 0;
 
   /**
    * Invokes the solver associated with the system.  For steady state
--- a/src/systems/fem_system.C
+++ b/src/systems/fem_system.C
@@ -74,7 +74,7 @@ void assemble_unconstrained_element_system(const FEMSystem & _sys,
 
   // We need jacobians to do heterogeneous residual constraints
   const bool need_jacobian =
-    _get_jacobian || _constrain_heterogeneously;
+    (_get_jacobian || _constrain_heterogeneously) ;
 
   bool jacobian_computed =
     _sys.time_solver->element_residual(need_jacobian, _femcontext);
@@ -251,6 +251,7 @@ void add_element_system(const FEMSystem & _sys,
                         const bool _get_residual,
                         const bool _get_jacobian,
                         const bool _constrain_heterogeneously,
+                        const bool _no_constraints,
                         FEMContext & _femcontext)
 {
 #ifdef LIBMESH_ENABLE_CONSTRAINTS
@@ -280,6 +281,7 @@ void add_element_system(const FEMSystem & _sys,
           (_femcontext.get_elem_jacobian(),
            _femcontext.get_elem_residual(),
            _femcontext.get_dof_indices(), false);
+      // Do nothing if(_no_constraints)
     }
   else if (_get_residual)
     {
@@ -291,13 +293,16 @@ void add_element_system(const FEMSystem & _sys,
       else
         _sys.get_dof_map().constrain_element_vector
           (_femcontext.get_elem_residual(), _femcontext.get_dof_indices(), false);
+      // Do nothing if(_no_constraints)
     }
   else if (_get_jacobian)
     {
       // Heterogeneous and homogeneous constraints are the same on the
       // matrix
-      _sys.get_dof_map().constrain_element_matrix
-        (_femcontext.get_elem_jacobian(), _femcontext.get_dof_indices(), false);
+      // Only get these contribs if we are applying some constraints
+      if(!_no_constraints)
+        _sys.get_dof_map().constrain_element_matrix
+          (_femcontext.get_elem_jacobian(), _femcontext.get_dof_indices(), false);
     }
 #endif // #ifdef LIBMESH_ENABLE_CONSTRAINTS
 
@@ -348,11 +353,13 @@ public:
   AssemblyContributions(FEMSystem & sys,
                         bool get_residual,
                         bool get_jacobian,
-                        bool constrain_heterogeneously) :
+                        bool constrain_heterogeneously,
+bool no_constraints) :
     _sys(sys),
     _get_residual(get_residual),
     _get_jacobian(get_jacobian),
-    _constrain_heterogeneously(constrain_heterogeneously) {}
+    _constrain_heterogeneously(constrain_heterogeneously),
+    _no_constraints(no_constraints){}
 
   /**
    * operator() for use with Threads::parallel_for().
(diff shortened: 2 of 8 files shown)


roystgnr:
Those last two commits are just bugfixes and ought to be squashed.

More seriously, though... we should be changing the *base* class functions to match here, shouldn't we?!?  I'll be very confused if this compiles with C++11; libmesh_override ought to catch that brand of error in that context.

roystgnr:
Ah, those changes are in fb91928, I get it.  Yeah, we probably just need to squash those two then before merging.

include/systems/diff_system.h
```diff
@@ -123,7 +123,8 @@ class DifferentiableSystem : public ImplicitSystem,
    */
   virtual void assembly (bool get_residual,
                          bool get_jacobian,
-                         bool apply_heterogeneous_constraints = false) libmesh_override = 0;
+                         bool apply_heterogeneous_constraints = false,
+			 bool apply_no_constraints = false) libmesh_override = 0;
```
dknez: Small point: The whitespace isn't aligned here (and in several other cases below).

jwpeterson:
Squashing into 1 commit should also fix your commit metadata. The last two commits were made by 
```
From: Vikram Garg <vikram@scott.ices.utexas.edu>
```
since your .gitconfig wasn't set up.

src/systems/fem_system.C
```diff
@@ -288,16 +294,27 @@ void add_element_system(const FEMSystem & _sys,
...
     }
   else if (_get_jacobian)
     {
-      // Heterogeneous and homogeneous constraints are the same on the
-      // matrix
-      _sys.get_dof_map().constrain_element_matrix
-        (_femcontext.get_elem_jacobian(), _femcontext.get_dof_indices(), false);
+      if (_no_constraints)
+	{
+	  // Do noting
```
jwpeterson: This is a typo and also it's just not necessary. `if (!_no_constraints)` should suffice...

vikramvgarg: Not sure I understand, should I just get rid of that entire if(_no_constraints) block ?

jwpeterson: Yes, if the point of it was documentation, just add a comment that there is nothing to do if there are `_no_constraints`.

vikramvgarg: Sounds good.

include/systems/diff_system.h
jwpeterson: You should be able run contrib/bin/reindent.sh on the files you changed. What editor are you using that doesn't automatically align stuff like this?

vikramvgarg: I am using emacs. Those misalignments happened because of tabbing. Have fixed them manually for now. Will run reindent.sh as well.

jwpeterson: Aha, add
```
(setq-default indent-tabs-mode nil)
```
to your .emacs file, as we don't allow mixes of tabs and spaces for indentation.

vikramvgarg: Cool, thanks :+1:

vikramvgarg: I am now getting strange netcdf errors soon after compilation starts. Did a bootstrap + config, but the error persists.

/bin/sh: ../../../../.././contrib/netcdf/v4/libsrc/../../../../.././contrib/netcdf/v4/libsrc/ncx.c: No such file or directory
  CC       libnetcdf3_la-nc_hashmap.lo
make[4]: *** [../../../../.././contrib/netcdf/v4/libsrc/ncx.c] Error 1
make[4]: *** Waiting for unfinished jobs....
  CC       libnetcdf3_la-lookup3.lo
make[4]: Leaving directory `/workspace/vikram/libmesh/FEMSystem_assembly_with_no_constraints/contrib/netcdf/v4/libsrc'
make[3]: *** [all-recursive] Error 1
make[3]: Leaving directory `/workspace/vikram/libmesh/FEMSystem_assembly_with_no_constraints/contrib/netcdf/v4'
make[2]: *** [all] Error 2
make[2]: Leaving directory `/workspace/vikram/libmesh/FEMSystem_assembly_with_no_constraints/contrib/netcdf/v4'
make[1]: Leaving directory `/workspace/vikram/libmesh/FEMSystem_assembly_with_no_constraints/contrib'

src/systems/fem_system.C
```diff
@@ -74,7 +75,7 @@ void assemble_unconstrained_element_system(const FEMSystem & _sys,
 
   // We need jacobians to do heterogeneous residual constraints
   const bool need_jacobian =
-    _get_jacobian || _constrain_heterogeneously;
+    (_get_jacobian || _constrain_heterogeneously) && (!_no_constraints);
```
roystgnr: This can't be right, can it?  If the user wants an unconstrained Jacobian, then _get_jacobian is true, _no_constraints is true, and need_jacobian will be *false*?

vikramvgarg: Yes, in fact, the _no_constraints option isnt needed at all in this function. Fixing this.

vikramvgarg:
@jwpeterson @dknez does the pull request look ok now ?

jwpeterson:
No, this should all be in 1 commit as @roystgnr mentioned. Google "git squash" if you're not sure how to do this... Once everything is squashed you'll need to force push to the same branch to restart the tests.

jwpeterson:
Also, it looks like you still have tabs in a few places. There is another script, `contrib/bin/detab.sh` you can use to get rid of those...

vikramvgarg:
Have made all the changes, let me know if there are any more issues.

roystgnr:
Let me throw this at GRINS before we merge.

roystgnr:
That Civet test is just an unrelated timeout, and GRINS seems happy.  Will merge tonight unless anyone objects.

pbauman:
Noticed while setting up `femputer` that this has generated lots of these warnings:

<pre>
/femputer/pbauman/civet_build_testing_root/libmesh-devel-install/include/libmesh/fem_system.h:85:16: warning: ‘virtual void libMesh::FEMSystem::assembly(bool, bool, bool, bool)’ was hidden [-Woverloaded-virtual]
   virtual void assembly (bool get_residual,
</pre>

pbauman:
I might see if I can get a `-Werror` build going to try and catch these.

roystgnr:
Damn, nice catch.  I'm sorry, but this will cause some confusing breakage in GRINS::MultiphysicsSystem until that gets updated to match the new FEMSystem API: stuff will *mostly* work, but GRINS::Physics objects with preassembly() methods won't see those called.  Is that acceptable?  Vikram's patch really did add a pretty blatantly missing new capability, and it would be a PITA to get it working in a partially backwards-compatible way; impossible to get it working in a fully backwards-compatible way.

I'd suggest going through GRINS and tacking on override keywords everywhere that's appropriate, to catch any future breakage of this nature at compile time *without* having to use -Werror or watch your warning messages closely.

However, a -Werror build is a brilliant idea... except that I'd then be tempted to start throwing my paranoid warning options into libMesh defaults, as soon as I could get a build to pass with each, as a way of enforcing that we'd never get code triggering that warning again.

pbauman:
Sorry, didn't mean to imply that we shouldn't have merge Vikram's patch, but rather I just wasn't aware it was going to break the API. I'll fix downstream, but it'll be good to catch such things here instead of later.

Also, yes I want to use C++11 keywords. Was trying to crank out remaining input file parsing bits before a new release and then hello C++11 requirement and then `override` all the things (in addition to `std::move`, `std::unique_ptr`, `std::shared_ptr`, `auto`, etc.)

PR #1343, opened 2017-05-15
fem_system_ex1 update API usage and fix mass_residual

canesin:
Fixed deprecation messages on API usage.
On mass_residual from my understanding of fem_system default mass_residual the contributions should actually be -M(u)u_dot, so fixed signals.

Regards,
F


examples/fem_system/fem_system_ex1/naviersystem.C
```diff
@@ -267,13 +267,16 @@ bool NavierSystem::element_time_derivative (bool request_jacobian,
...
-      Gradient grad_u = c.interior_gradient(u_var, qp),
-        grad_v = c.interior_gradient(v_var, qp),
-        grad_w = c.interior_gradient(w_var, qp);
+      Number p, u, v, w;
+      c.interior_value(p_var, qp, p),
+      c.interior_value(u_var, qp, u),
+      c.interior_value(v_var, qp, v),
+      c.interior_value(w_var, qp, w);
+
+      Gradient grad_u, grad_v, grad_w;
```
jwpeterson: Minor point, but I think the idea with this API is that the `grad_u`, `grad_v`, and `grad_w` variable declarations would be hoisted outside of the qp-loop to avoid repeatedly creating/destroying them...

examples/fem_system/fem_system_ex1/naviersystem.C
```diff
@@ -549,29 +555,30 @@ bool NavierSystem::mass_residual (bool request_jacobian,
...
 
       for (unsigned int i = 0; i != n_u_dofs; ++i)
         {
-          Fu(i) += JxWxRexU * phi[i][qp];
-          Fv(i) += JxWxRexV * phi[i][qp];
+          Fu(i) -= JxWxRexU * phi[i][qp];
+          Fv(i) -= JxWxRexV * phi[i][qp];
           if (dim == 3)
-            Fw(i) += JxWxRexW * phi[i][qp];
+            Fw(i) -= JxWxRexW * phi[i][qp];
```
jwpeterson: Do you have some screenshots of this example's solutions from before/after the sign change? It seems like the results should have been very wrong before...

examples/fem_system/fem_system_ex1/naviersystem.C
canesin: Out of curiosity, did you guys have profiled it ? I know being a object can kill this but I usually declare inside loops so that the compiler knows it is limited to that scope and can do register allocation.

jwpeterson: > Out of curiosity, did you guys have profiled it ? 

Definitely didn't profile it, and doubt that it makes any performance difference in practice, but IMO the old API should be preferred if you want to declare the variables inside the loop since it combines creation and initialization into a single statement. The current version is the worst of both worlds since it both opens up the possibility to forget to set the values *and* does separate creation and initialization steps.

examples/fem_system/fem_system_ex1/naviersystem.C
canesin: Not very, that's a major reason for being confusing (and why having a review on this PR is extra helpful, thanks).

New:
fem_system_ex1_new

Old:
fem_system_ex1_old

PR #1344, opened 2017-05-16
Update FEM System ex1

canesin:
(forgot to force push on previous PR)

Fixed deprecation messages on API usage.
From my understanding of fem_system default mass_residual the contributions should actually be the negative -M(u)u_dot, so fixed signals. Visual changes are very minor, which is a major reason for being confusing.

New
fem_system_ex1_new

Old
fem_system_ex1_old

### DIFF
--- a/examples/fem_system/fem_system_ex1/naviersystem.C
+++ b/examples/fem_system/fem_system_ex1/naviersystem.C
@@ -111,10 +111,10 @@ void NavierSystem::init_data ()
 
   // Tell the system to march velocity forward in time, but
   // leave p as a constraint only
-  this->time_evolving(u_var);
-  this->time_evolving(v_var);
+  this->time_evolving(u_var, 1);
+  this->time_evolving(v_var, 1);
   if (dim == 3)
-    this->time_evolving(w_var);
+    this->time_evolving(w_var, 1);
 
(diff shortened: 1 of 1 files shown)


jwpeterson:
I suspect you are looking at a (nearly) steady state solution here. If so, the sign on the transient term doesn't make any difference. @roystgnr would you mind weighing in on this change??

roystgnr:
I would greatly mind weighing in on this change, because as best as I can tell, it highlights a bug I wrote, in an important example code, which has gone unfixed for nearly a decade, and the shame I'm experiencing right now is torturous.

My only concern right now is, was this the *only* bug?  I'd like to check the other terms' signs before we merge.

roystgnr:
Convection is negative, as it should be since we pull it to the RHS.

Pressure is positive and diffusion negative, thanks to the integration by parts.

Forcing is positive by definition.

Pressure constraint is positive for symmetry.

Yeah, :+1: to merge.

jwpeterson:
Thanks, I just invalidated the one CI test that failed, so we can wait for that to finish or not, there's no way this example can affect the running of MOOSE.

jwpeterson:
Thanks @canesin!

Issue #1351, opened 2017-05-19
Assertion in dbg mode in AdjointRefinementEstimator


pbauman:
This is reproducible only in `dbg` mode (don't need `GLIBCXX_DEBUG`) on GRINS test `regression/extra_quadrature_order_laplace_arefee_amr.sh`. I haven't tried to use a libMesh-only case yet. It's for sure between hash's 4b96823 (current GRINS libMesh-minimum hash) and 47c3dfa (before #1342). Will try and bisect more tonight. This is the failure that's showing up on the GRINS-dbg CIVET testing.

The trace below was generated with only 1 processor.

<pre>
==========================================================
Estimating error
==========================================================
Assertion `neigh_has_remote_children' failed.

[Thread debugging using libthread_db enabled]                                                                                                                                                                                                 
0x0000003321cac82e in waitpid () from /lib64/libc.so.6
To enable execution of this file add
        add-auto-load-safe-path /fry1/data/shared/software/apps/gcc/6.3.0/lib64/libstdc++.so.6.0.22-gdb.py
line to your configuration file "/eng/home/pbauman/.gdbinit".
To completely disable this security protection add
        set auto-load safe-path /
line to your configuration file "/eng/home/pbauman/.gdbinit".
For more information about this security protection see the
"Auto-loading safe path" section in the GDB manual.  E.g., run from the shell:
        info "(gdb)Auto-loading safe path"
#0  0x0000003321cac82e in waitpid () from /lib64/libc.so.6
#1  0x0000003321c3e479 in do_system () from /lib64/libc.so.6
#2  0x00007fd97cebe6c4 in (anonymous namespace)::gdb_backtrace(std::basic_ostream<char, std::char_traits<char> >&) () at ../../libMesh/src/base/print_trace.C:145
#3  0x00007fd97cebea51 in libMesh::print_trace(std::basic_ostream<char, std::char_traits<char> >&) () at ../../libMesh/src/base/print_trace.C:192
#4  0x00007fd97ceb92fd in libMesh::MacroFunctions::report_error(char const*, int, char const*, char const*) () at ../../libMesh/src/base/libmesh_common.C:82
#5  0x00007fd97d602148 in libMesh::UnstructuredMesh::find_neighbors(bool, bool) () at ../../libMesh/src/mesh/unstructured_mesh.C:449
[...]

PR #1352, opened 2017-05-23
Find subactive tricky neighbors

roystgnr:
The first commit here should fix the bug Paul Bauman caught in #1351; the second commit should fix potential DistributedMesh bugs in this case and similar cases.


pbauman:
Thanks much @roystgnr. Sorry I didn't have these `femputer` tests up sooner. Hopefully I'll be able to get `DistributedMesh` tests going in the near term.

PR #1354, opened 2017-05-25
AdjointRefinementEstimator::_residual_evaluation_physics

roystgnr:
@vikramvgarg and I are getting much-improved error estimator results from stabilized formulations when we enable the user to supply an unstabilized residual to evaluate.

This PR includes fixes to a long-standing DiffSystem bug (incomplete feature?) where we weren't respecting attached physics pointers, a fix to the no_constraints option, and adds the user-residual option to ARefEE.

### DIFF
--- a/include/error_estimation/adjoint_refinement_estimator.h
+++ b/include/error_estimation/adjoint_refinement_estimator.h
@@ -111,11 +111,6 @@ public:
     return computed_global_QoI_errors[qoi_index];
   }
 
-  /**
-   * Swap pointer of current physics object with that of another called swap_physics
-   */
-  void swap (FEMPhysics* swap_physics_1, FEMPhysics* swap_physics_2);
-
   virtual ErrorEstimatorType type() const
   { return ADJOINT_REFINEMENT;}
 
@@ -135,6 +130,12 @@ public:
    */
   FEMPhysics * get_residual_evaluation_physics()
   { return this->_residual_evaluation_physics; }
+
+  /**
+   * Set the _residual_evaluation_physics member to argument
+   */
+  void set_residual_evaluation_physics(FEMPhysics* set_physics)
+  { this->_residual_evaluation_physics = set_physics; }
 
 protected:
 
--- a/include/error_estimation/adjoint_refinement_estimator.h
+++ b/include/error_estimation/adjoint_refinement_estimator.h
@@ -24,11 +24,13 @@
 #include "libmesh/error_estimator.h"
 #include "libmesh/libmesh.h"
 #include "libmesh/qoi_set.h"
-#include "libmesh/diff_physics.h"
 
 // C++ includes
 #include <cstddef>
 #include <vector>
+
+// Forward declarations
+class DifferentiablePhysics;
 
 #ifdef LIBMESH_ENABLE_AMR
 
--- a/include/error_estimation/adjoint_refinement_estimator.h
+++ b/include/error_estimation/adjoint_refinement_estimator.h
@@ -56,8 +56,8 @@ public:
     ErrorEstimator(),
     number_h_refinements(1),
     number_p_refinements(0),
-      _qoi_set(QoISet()),
-      _residual_evaluation_physics(libmesh_nullptr)
+    _qoi_set(QoISet()),
+    _residual_evaluation_physics(libmesh_nullptr)
   {
     // We're not actually going to use error_norm; our norms are
     // absolute values of QoI error.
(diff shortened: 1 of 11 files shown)


roystgnr:
That actually might be a real error, and thanks so much to @pbauman for getting test coverage on it.  I'll see if I can replicate.

roystgnr:
Yeah, I can replicate this.  Hopefully in the morning I'll be able to explain it too...

roystgnr:
It was a real bug; I've improved the documentation as well as adding the fix.  IMHO this should be okay to merge, though I can try and clean up that hideous commit history first if @jwpeterson balks.

jwpeterson:
> if @jwpeterson balks

It's up to you how much you want to combine/reorder, but I'd at least fix the commit metadata for the first dozen or so commits which have "Vikram Garg <vikram@scott.ices.utexas.edu>" as the author. I'd also reformat the commit messages slightly on those ones so they more or less follow [this format](https://chris.beams.io/posts/git-commit/). Consecutive cleanup commits 890aaeb and a613558 could definitely be squashed together.

pbauman:
Glad I got `femputer` going here! I'm sorry I don't have time right now to look more carefully at this, but I'm anxious to because updating error indicators in the adjoint refinement case is something I'm very keen on playing with (preferably with a student helping, but I'm not holding my breath on that at the moment).

roystgnr:
Unrelated timeout this time, but I want to do some cleanup so I'll let tests rerun after I do.

roystgnr:
This is the best I can do without really dissecting some intermediate commits, so (presuming that last rebase didn't screw anything up) I'll merge this weekend.

vikramvgarg:
Folks,
          I guess the non intuitive commits are present due to the fact
that I work on a work and home desktop. Therefore I push at the end of the
day, even if its not a logical end point.

I'll subbranch feature branches and merge back to maintain a clean commit
history in the future.

Thanks.

dschwen:
Or just amend your end of day commits the next day...

jwpeterson:
> Or just amend your end of day commits the next day

Yes, just to expand slightly on this point, unless a feature branch is very simple, *no one* gets a clean commit history on the first attempt, and IMO you shouldn't even worry about trying to do so when you're developing. 

Git makes it very easy to clean up/reorganize/combine/etc. commits, and you should definitely do this before making a PR, which represents a more polished ~~turd~~ version of your work that you are ready for others to review. 

In my own workflow, I also maintain the original (messy) development branch for historical/reference purposes (under a different branch name, of course) for a couple weeks after the cleaned-up version has been merged, but there's nothing preventing you from keeping development branches around forever if you want.

roystgnr:
Check point commits are a fine way to backup work.  I usually put "Don't put this in a PR or John will kill me" as my commit log messages for those, though.  (with names other than John for other projects)  The catch is that the first thing you do after making a checkpoint isn't figure out what to put in the next commit, it's back out of the checkpoint, finish the work, and then re-commit in chunks.

The big problem that made more cleanup intractable here was the "kitchen sink" commits.  Take a look at the first commits above: "Added a swap function" also added the first use of it; "Moved swap; minor bug fixes" should obviously have been two separate commits; likewise for "printing out pointers and some errors fixed".  Cleaning that kind of thing up then requires going through line by line and figuring out which change should go with which changeset.  Use "git add --patch" and sort all those things out at first commit time, when the distinctions are still fresh in your mind.

pbauman:
`git add -p` is the killer feature here IMHO, allowing the avoidance of "kitchen sink" commits. I, also, work in fits and starts and using `git add -p` allows me to make micro commits that I can come back later and fixup/squash in appropriate places or reorder commits etc. It takes 30 extra seconds when committing your work, but saves potentially hours down the line (or even the commit history as in this case where there's not enough developer time to clean up the mess).

Don't mistake this work as just a hurdle you have to get over to get your stuff in or think you can rely on us for it. It's essential that you be able to do this yourself and it's essential that it be done so we can understand what you do and we can revisit back in the future and understand what was done, e.g. if we have to `git bisect` (like I did last week(?)).

vikramvgarg:
Paul, git add --patch will definitely solve a lot of problems for me. 

I guess the slightly more philosophical point here is that 'code always has bugs', and if the current suite of regression tests doesnt detect one, a future addition (or alternate suite like in this case) could detect one. And library commits/merges should be done in a way to make debugging then as easy as possible.

I'll get a chance to do a better job soon, as we need to add capability to solve the adjoint with a user specified physics, which should help us move towards a automated model adaptivity framework.

Issue #293, opened 2014-07-14
adaptivity_ex3 generates garbled output


jwpeterson:
The bug must be in `error.plot_error( error_output, mesh );` but I'm not sure exactly what's going on beyond that.

[garbled 1](https://drive.google.com/file/d/0B9BK7pg8se_ibWtNVUxOVVZVejg/edit?usp=sharing)
[garbled 2](https://drive.google.com/file/d/0B9BK7pg8se_iT25HOEJlMzFSYkk/edit?usp=sharing)
[garbled 3](https://drive.google.com/file/d/0B9BK7pg8se_iRlp1UlotbDY2Unc/edit?usp=sharing)

PR #1356, opened 2017-06-01
Disallow ErrorVector::plot_error() mesh renumbering

roystgnr:
Our original reason for allowing renumbering was probably a workaround for when disabling renumbering was iffy; perhaps disallowing nodal renumbering in all_first_order() would leave a bunch of NULL Node* gaps in the SerialMesh data structures that the SerialMesh code wasn't prepared for?  @jwpeterson might remember better.

My reason for disabling renumbering is that in corner cases it can cause *element* renumbering in the mesh copy here, which trashes the plotting output of error vectors which we keep indexed by element id.  If there still is any remaining problem with disabling renumbering, this isn't the right workaround.

### DIFF
--- a/src/mesh/mesh_modification.C
+++ b/src/mesh/mesh_modification.C
@@ -439,8 +439,8 @@ void UnstructuredMesh::all_first_order ()
   // might now have an invalid nodal processor_id()
   Partitioner::set_node_processor_ids(*this);
 
-  // delete or renumber nodes, etc
-  this->prepare_for_use(/*skip_renumber =*/ false);
+  // delete or renumber nodes if desired
+  this->prepare_for_use();
 }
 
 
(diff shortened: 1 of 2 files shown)


jwpeterson:
For context, see discussion on the corresponding [PR](https://github.com/libMesh/libmesh/pull/298) and [Issue](https://github.com/libMesh/libmesh/issues/293).

I'd say definitely double check the output of adaptivity_ex3 and make sure you haven't re-broken it...

roystgnr:
I missed that; thanks!

roystgnr:
Okay; I *think* I understand what was going on: up until 405fe519, our Exodus output didn't handle discontiguous node numberings, so if we deleted a bunch of nodes without allowing renumbering to make the survivors contiguous again, trying to make error.e from the result gave us trash.

adaptivity_ex3 (with h refinement, to match the 2014 behavior) doesn't currently look scrambled to me, and I think I understand why it would be fixed now without the workaround, so this ought to be fine to merge.

PR #1362, opened 2017-06-12
64-bit build fix

### LINKED ISSUE #1361: Minor 64-bit build regression
Looks like the AMR app has picked up a 64-bit dof regression. Patching coming...

permcody:
closes #1361

@roystgnr - this needs to go into the current MOOSE update since we test against 64-bit builds on master. I tested this locally, it's the only regression so you can merge before testing completes.


roystgnr:
Sounds good; thanks!

Looks like amr.C was *never* 64-bit compatible; we just didn't notice until now because @jwpeterson just added it to our Makefile a few weeks ago?

PR #1366, opened 2017-06-16
Update BOOST m4 file to include proper flag

aeslaughter:
Thanks @jwpeterson

### DIFF
--- a/configure
+++ b/configure
@@ -33448,7 +33448,7 @@ $as_echo "$as_me: Your boost libraries seems to old (version $_version)." >&6;}
 
 
 
-$as_echo "#define HAVE_BOOST /**/" >>confdefs.h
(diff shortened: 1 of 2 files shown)


jwpeterson:
:+1:

roystgnr:
Has this been submitted upstream?  I don't think having to use ```#ifdef``` is so awful that it's worth forking boost.m4.

jwpeterson:
Well, we've already forked it (13ef034177d5618d9b340770a7fc01e170e9cbb4, 0d8ea46682b082a0bb8f638af31f23273e0515e0) and those changes are not in upstream either...

jwpeterson:
Not to say that two wrongs make a right, but I don't think diverging from upstream autoconf-archive in order to support libmesh-specific changes is all that big of a deal...

Issue #1364, opened 2017-06-15
Drop redundant documentation from libmesh_override functions


jwpeterson:
We have a bunch of copy-pasted documentation in e.g. the `NumericVector` and `SparseMatrix` class hierarchies which is sometimes also inaccurate. I propose just removing the documentation for functions in derived classes that are marked libmesh_override unless there is some special behavior that needs to be discussed.

Issue #1371, opened 2017-06-20
SparseMatrix::close() is not logically const


jwpeterson:
So it shouldn't be tagged as `const`. You can see how we worked around the interface using `const_cast`s in `LaspackMatrix` and `EigenSparseMatrix`, but that really shouldn't be necessary... 

I suspect making this non-const will lead to a chain reaction of other const-related issues, but I just wanted to keep this on the radar.

jwpeterson:
Fixed by #1373.

PR #1372, opened 2017-06-20
Delete copy/pasted documentation of overridden virtual functions.

jwpeterson:
Copy/pasted documentation is bad because it either requires extra
maintenance when the base class functionality changes, or it gets out
of date. This commit only touches classes in the numerics directory,
but the same concept could be applied across the entire library.

Refs #1364.

### DIFF
--- a/include/numerics/analytic_function.h
+++ b/include/numerics/analytic_function.h
@@ -34,9 +34,10 @@ template <typename T>
 class DenseVector;
 
 /**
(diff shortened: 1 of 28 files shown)


jwpeterson:
Hopefully this isn't too controversial. I have an upcoming PR for #1371 which I've branched from this for, so I'm going to go ahead and merge.

PR #1375, opened 2017-06-28
Fix move constructor test

jwpeterson:
We came across a version of the [Intel 2015 compiler](https://groups.google.com/d/msg/moose-users/MqjruY_SWs0/kwmdkbErBwAJ) that supported all parts of the C++11 move constructor syntax apart from the `noexcept` keyword, and then I noticed that I forgot to test for it in the configure test. This branch fixes the test.


jwpeterson:
The timeout in dbg/parallel mode isn't related to this m4 file change...

PR #1378, opened 2017-07-06
Clean up system vector and matrix initialization

### LINKED ISSUE #1376: TransientSystem::init_data() is not necessary?
If the template class is a `System`, `Base::init_data()` should have initialized the old and older vectors. Thus https://github.com/libMesh/libmesh/blob/master/src/systems/transient_system.C#L106 is really not needed, right?

Similarly, why is `ExplicitSystem::add_system_rhs()` not called in constructor? If so, its `init_data` will also not be needed. I see a commented line in constructor https://github.com/libMesh/libmesh/blob/master/src/systems/explicit_system.C#L39 maybe somebody have thought about this?

YaqiWang:
This also enables the construction of vectors and matrices outside of the systems. Close #1376.

### DIFF
--- a/include/systems/implicit_system.h
+++ b/include/systems/implicit_system.h
@@ -293,6 +293,11 @@ public:
   SparseMatrix<Number> & add_matrix (const std::string & mat_name);
 
   /**
+   * Removes the additional matrix \p mat_name from this system
+   */
+  void remove_matrix(const std::string & mat_name);
+
+  /**
    * \returns \p true if this \p System has a matrix associated with the
    * given name, \p false otherwise.
    */
@@ -351,7 +356,6 @@ public:
    */
   bool zero_out_matrix_and_rhs;
 
-
 protected:
 
   /**
@@ -364,8 +368,6 @@ protected:
    * Initializes the matrices associated with this system.
    */
   virtual void init_matrices ();
-
-
 
 private:
 
--- a/src/systems/explicit_system.C
+++ b/src/systems/explicit_system.C
@@ -42,16 +42,6 @@ ExplicitSystem::ExplicitSystem (EquationSystems & es,
 
 
 
-ExplicitSystem::~ExplicitSystem ()
-{
-  // clear data
-  this->clear();
-
-  remove_vector("RHS Vector");
-}
-
-
-
 void ExplicitSystem::clear ()
 {
   // Clear the parent data
--- a/src/systems/explicit_system.C
+++ b/src/systems/explicit_system.C
@@ -36,7 +36,8 @@ ExplicitSystem::ExplicitSystem (EquationSystems & es,
   rhs(libmesh_nullptr)
 
 {
-  //rhs = &(this->add_vector ("RHS Vector"));
+  // Add the system RHS.
+  this->add_system_rhs ();
 }
 
 
@@ -45,6 +46,8 @@ ExplicitSystem::~ExplicitSystem ()
 {
   // clear data
   this->clear();
+
+  remove_vector("RHS Vector");
 }
 
 
@@ -54,22 +57,8 @@ void ExplicitSystem::clear ()
   // Clear the parent data
   Parent::clear();
 
-  // NULL-out the vector.  Note that
-  // System::clear() actually deleted it.
-  rhs = libmesh_nullptr;
-}
-
-
-
-void ExplicitSystem::init_data ()
-{
-  // Add the system RHS.
-  // (We must do this before initializing the System data,
-  //  then we lose the opportunity to add vectors).
+  // Restore us to a "basic" state
   this->add_system_rhs ();
-
-  // initialize parent data
-  Parent::init_data();
 }
 
 
--- a/src/systems/implicit_system.C
+++ b/src/systems/implicit_system.C
@@ -47,6 +47,8 @@ ImplicitSystem::ImplicitSystem (EquationSystems & es,
   zero_out_matrix_and_rhs(true),
   _can_add_matrices (true)
 {
+  // Add the system matrix.
+  this->add_system_matrix ();
 }
 
 
@@ -55,6 +57,8 @@ ImplicitSystem::~ImplicitSystem ()
 {
   // Clear data
   this->clear();
+
+  remove_matrix("System Matrix");
 }
 
 
@@ -78,8 +82,8 @@ void ImplicitSystem::clear ()
     _can_add_matrices = true;
   }
 
-  // NULL the matrix.
-  matrix = libmesh_nullptr;
+  // Restore us to a "basic" state
+  this->add_system_matrix ();
 }
 
 
@@ -93,9 +97,6 @@ void ImplicitSystem::init_data ()
   for (matrices_iterator pos = _matrices.begin();
        pos != _matrices.end(); ++pos)
     pos->second->clear();
-
-  // Add the system matrix.
-  this->add_system_matrix ();
 
   // Initialize the matrices for the system
   this->init_matrices ();
@@ -222,6 +223,20 @@ SparseMatrix<Number> & ImplicitSystem::add_matrix (const std::string & mat_name)
   _matrices.insert (std::make_pair (mat_name, buf));
 
   return *buf;
+}
+
+
+void ImplicitSystem::remove_matrix (const std::string & mat_name)
+{
+  matrices_iterator pos = _matrices.find (mat_name);
+
+  //Return if the matrix does not exist
+  if (pos == _matrices.end())
+    return;
+
+  delete pos->second;
+
+  _matrices.erase(pos);
 }
 
 
--- a/src/systems/system.C
+++ b/src/systems/system.C
@@ -713,15 +713,16 @@ NumericVector<Number> & System::add_vector (const std::string & vec_name,
 
 void System::remove_vector (const std::string & vec_name)
 {
+  vectors_iterator pos = _vectors.find(vec_name);
+
   //Return if the vector does not exist
-  if (!(this->have_vector(vec_name)))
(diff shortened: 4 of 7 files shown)


YaqiWang:
I will not touch the part in @roystgnr 's #1377.

roystgnr:
Eliminating TransientSystem::init_data once it's unused was a nice touch.  We can apply that on top of my refactoring.

jwpeterson:
@YaqiWang please accept my invitation to join the libmesh "Associates" team so your PRs will be tested automatically. Every time you push I need to activate them all by hand...

YaqiWang:
@jwpeterson Thanks for the invitation. I have accepted.

jwpeterson:
Hm... sounds like you guys were not expecting this to conflict with #1377, but it does.

YaqiWang:
yes, i am resolving it.

src/systems/implicit_system.C
```diff
@@ -113,6 +107,11 @@ void ImplicitSystem::init_matrices ()
   if (matrix->initialized())
     return;
 
+  // Clear any existing matrices
```
YaqiWang: I am not sure if I should move this code inside here. Let me try move it back. I will need to run MOOSE test to make sure. BTW, how do I run libmesh test locally?

jwpeterson: > how do I run libmesh test locally?

`make check` in the top level of the build directory will run all the libmesh unit tests and examples. You can also run `make check` within a single example directory to test it alone.

YaqiWang:
I saw a `Memory leak detected!` in one of the examples, possibly due to system matrix management. So do not merge it.

YaqiWang:
I think this is good to go. Thanks.

include/systems/implicit_system.h
```diff
@@ -352,6 +357,11 @@ class ImplicitSystem : public ExplicitSystem
   bool zero_out_matrix_and_rhs;
 
 
+  /**
+   * Initializes the matrices associated with this system.
+   */
+  virtual void init_matrices ();
```
jwpeterson: Maybe I didn't fully understand the new design that was discussed. Did `init_matrices()` need to become public?

src/systems/explicit_system.C
```diff
@@ -54,22 +57,8 @@ void ExplicitSystem::clear ()
...
-
-
-
-void ExplicitSystem::init_data ()
-{
-  // Add the system RHS.
-  // (We must do this before initializing the System data,
-  //  then we lose the opportunity to add vectors).
+  // Restore us to a "basic" state
   this->add_system_rhs ();
```
jwpeterson: Were we calling `add_system_rhs()` in `ExplicitSystem::clear()` before? If not, I find it confusing for a `clear()` function to be adding vectors.

src/systems/implicit_system.C
```diff
@@ -225,6 +226,18 @@ SparseMatrix<Number> & ImplicitSystem::add_matrix (const std::string & mat_name)
...
 
+void ImplicitSystem::remove_matrix (const std::string & mat_name)
+{
+  //Return if the matrix does not exist
+  if (!(this->have_matrix(mat_name)))
+    return;
+
+  _matrices[mat_name]->clear();
+  delete _matrices[mat_name];
+  _matrices[mat_name] = libmesh_nullptr;
```
jwpeterson: This leaves a NULL entry in the `_matrices` container that will cause `ImplicitSystem::have_matrix()` to return confusing values.

jwpeterson: It also does 3 map lookups (via operator[]) which is unnecessary.

include/systems/implicit_system.h
YaqiWang: I thought I would need this to be public. But after working on moose after this, turns out I only need to do
```
  // do not assemble system matrix for JFNK solve
  if (solverParams()._type == Moose::ST_JFNK)
    _nl->system().set_basic_system_only();
  _nl->nonlinearSolver()->jacobian = NULL;
```
in MOOSE `FEProblemBase::init`. So this does not have to be public to me. However, it may be desired in general.

src/systems/explicit_system.C
YaqiWang: Me too, but this appears necessary for restoring. It also follows @roystgnr 's coding in `TransientSystem::clear`.

src/systems/implicit_system.C
YaqiWang: good catch. I forgot to erase the entry in `_matrices`.

include/systems/implicit_system.h
jwpeterson: I never desire to make things public without very compelling reasons....

src/systems/implicit_system.C
YaqiWang: I know that is not ideal, I basically copied some codes in `remove_vector`... Want me change that as well?

include/systems/implicit_system.h
YaqiWang: ok, will make it back.

src/systems/implicit_system.C
jwpeterson: > Want me change that as well?

No, not for this PR. I'll fix it some other time...

YaqiWang: Yeah, because it is in system.C, I fixed it here already. It belongs to cleanup.

src/systems/implicit_system.C
```diff
@@ -225,6 +226,21 @@ SparseMatrix<Number> & ImplicitSystem::add_matrix (const std::string & mat_name)
...
+  if (!(this->have_matrix(mat_name)))
+    return;
+
+  matrices_iterator pos = _matrices.find (mat_name);
+
+  pos->second->clear();
+  delete pos->second;
+
+  _matrices.erase(mat_name);
+}
```
jwpeterson: But this still does 3 lookups: 1 for `have_matrix()`, one for `find()` and one for `erase(key)`. Also I don't think it's necessary to call `clear()` followed by delete because all the `SparseMatrix` destructors call `clear()`. So, the following should work:
```
matrices_iterator pos = _matrices.find (mat_name);
if (pos==end) return;
delete pos->second;
_matrices.erase(pos);
```

YaqiWang: You are right, will update.

src/systems/implicit_system.C
```diff
@@ -225,6 +226,20 @@ SparseMatrix<Number> & ImplicitSystem::add_matrix (const std::string & mat_name)
 }
 
 
+void ImplicitSystem::remove_matrix (const std::string & mat_name)
+{
+  matrices_iterator pos = _matrices.find (mat_name);
+
+  //Return if the matrix does not exist
+  if (pos == _matrices.end())
+    return
```
jwpeterson: No semi-colon here! I'm not sure how this even compiles?!

YaqiWang: yes, strange.

src/systems/explicit_system.C
```diff
@@ -45,6 +46,8 @@ ExplicitSystem::~ExplicitSystem ()
 {
   // clear data
   this->clear();
+
+  remove_vector("RHS Vector");
```
roystgnr: Why do we need this here?  The additional vectors are all about to be destroyed by System::~System anyways.

src/systems/implicit_system.C
```diff
@@ -55,6 +57,8 @@ ImplicitSystem::~ImplicitSystem ()
 {
   // Clear data
   this->clear();
+
+  remove_matrix("System Matrix");
```
roystgnr: This, on the other hand, looks like the ideal workaround for what otherwise would have been a memory leak introduced by the improved ImplicitSystem::clear() behavior.

src/systems/system.C
```diff
@@ -713,15 +713,16 @@ NumericVector<Number> & System::add_vector (const std::string & vec_name,
 
 void System::remove_vector (const std::string & vec_name)
 {
+  vectors_iterator pos = _vectors.find(vec_name);
+
```
roystgnr: Thanks for the optimization

src/systems/explicit_system.C
YaqiWang: Not sure, because System::~System calls this->clear(), which is ExplicitSystem::clear(), right? If so, rhs vector is not cleared.

roystgnr: Although, to be fair, that's not obvious.  System::~System() calls this->clear(), and C++ rules say that that means System::clear(), *not* ExplicitSystem::clear(), which in this case is exactly what we want and won't leave a re-added RHS vector, but which is such obscure behavior that CERT recommends [never doing it.](https://www.securecoding.cert.org/confluence/display/cplusplus/OOP50-CPP.+Do+not+invoke+virtual+functions+from+constructors+or+destructors)

roystgnr:
Verify that that remove_vector() call is redundant, and take it out (a separate commit, no need to rebase, since this isn't a breakage-inducing bug), and this should be good to merge.

roystgnr:
And if you want to be super nice you could add a comment to System::~System explaining what's going on so that it doesn't confuse anyone else in the future.

roystgnr:
Or you could even just change this->clear() to an explicit System::clear() there, if you feel that's clearer.

(no, I couldn't resist the pun)

src/systems/explicit_system.C
YaqiWang: I see. Thanks. We then can completely remove the ExplicitSystem destructor, right?

YaqiWang:
@roystgnr added one commit to address your comment.

roystgnr:
This looks good to me, once tests are done.

PR #1380, opened 2017-07-13
Actually write the version info to reduced basis Xdr headers

dknez:
Previously the version was not written to the header files.

### DIFF
--- a/src/reduced_basis/rb_evaluation.C
+++ b/src/reduced_basis/rb_evaluation.C
@@ -932,7 +932,14 @@ void RBEvaluation::write_out_vectors(System & sys,
   file_name << directory_name << "/" << data_name << "_header" << basis_function_suffix;
   Xdr header_data(file_name.str(),
                   write_binary_vectors ? ENCODE : WRITE);
-  sys.write_header(header_data, get_io_version_string(), /*write_additional_data=*/false);
+
+  std::string version("libMesh-" + libMesh::get_io_compatibility_version());
+#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+  version += " with infinite elements";
+#endif
+  header_data.data(version ,"# File Format Identifier");
+
+  sys.write_header(header_data, /*(unused arg)*/ version, /*write_additional_data=*/false);
 
   // Following EquationSystemsIO::write, we use a temporary numbering (node major)
   // before writing out the data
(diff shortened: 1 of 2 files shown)


src/reduced_basis/rb_evaluation.C
```diff
@@ -932,7 +932,14 @@ void RBEvaluation::write_out_vectors(System & sys,
...
                   write_binary_vectors ? ENCODE : WRITE);
-  sys.write_header(header_data, get_io_version_string(), /*write_additional_data=*/false);
+
+  std::string version("libMesh-" + libMesh::get_io_compatibility_version());
+#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+  version += " with infinite elements";
+#endif
+  header_data.data(version ,"# File Format Identifier");
+
+  sys.write_header(header_data, /*(unused arg)*/ version, /*write_additional_data=*/false);
```
jwpeterson: The `version` arg is now actually used, though, right?

jwpeterson: Oh, wait, different `version`. Never mind.

src/reduced_basis/rb_evaluation.C
```diff
@@ -1054,19 +1061,19 @@ void RBEvaluation::read_in_vectors_from_multiple_files(System & sys,
...
 
-  // set the version number in header_data from io_version_string
-  // (same code as in EquationSystemsIO::_read_impl)
-  std::string io_version_string = get_io_version_string();
-  std::string::size_type lm_pos = io_version_string.find("libMesh");
-  std::istringstream iss(io_version_string.substr(lm_pos + 8));
+  // Set up the header info (based on EquationSystems::_read_impl).
+  std::string version;
+  header_data.data(version);
+  std::string::size_type lm_pos = version.find("libMesh");
```
jwpeterson: I know it's not really part of the patch, but the `string::find()` call without a corresponding `npos` check and the hard-coded 8 seems a little clunky to me.
```
const std::string label = "libMesh-";
std::string::size_type lm_pos = version.find(label);
// check for lm_pos==std::string::npos
std::istringstream iss(version.substr(lm_pos + label.size()));
```

dknez: Good point. I'll take care of that before merging.

dknez: Should I also fix this in equation_systems_io.C? (I must have copied that code from equation_systems_io.C.)

dknez: Actually, I suggest a separate patch where we fix both of those at the same time, and not include it in this PR. What do you think?

jwpeterson: > What do you think?

👍

dknez:
I'll merge this once tests pass, unless anyone has any objection?

jwpeterson:
> I'll merge this once tests pass, unless anyone has any objection?

Sounds good.

PR #1382, opened 2017-07-14
Further clean-up of vector writing/reading in RBEvaluation.

dknez:
Previously we separated the Xdr header and body into two files, which was unnecessary and ugly.
Since we cleaned up the header writing and reading in a recent change, now is also a good time
to fix this two-separate-files issue as well.


dknez:
I've tested this and it works for me. It only affects the RB code, so I assume it's OK to merge unless anyone indicates otherwise (I'll do the merge tomorrow, assuming the tests pass).

PR #1384, opened 2017-07-25
Test cherry-picked commits for 1.2.1 release

jwpeterson:
This ~~may or may~~ will not pass all the MOOSE (or GRINS) testing, it is primarily intended as a version of the 1.2.0 release that also includes the recent NetCDF update.

This is also a test of whether or not CIVET is happy testing PRs into non-master branches. ✅ 

This PR should *not* be merged. If it passes to everyone's satisfaction I will simply tag it with "v1.2.1". The PR is only to trigger the CI testing...


jwpeterson:
@pbauman I tried to check why GRINS failed but femputer is timing out for me... it probably isn't anything that would prevent a 1.2.1 bugfix release from going ahead, but it would be nice to know for sure. Is there any way you can check on the failed build?

jwpeterson:
The MOOSE builds fail because MOOSE has already been updated to use the new `libMesh::BoundingBox` which didn't exist at the time 1.2.0 was created.

```
MOOSE Compiling C++ (in opt mode) /home/moosetest/civet/client_root_0/moose/framework/src/meshmodifiers/ElementDeleterBase.C...
../framework/src/mesh/MooseMesh.C: In member function 'libMesh::MeshTools::BoundingBox MooseMesh::getInflatedProcessorBoundingBox(libMesh::Real) const':
../framework/src/mesh/MooseMesh.C:2382:22: error: 'create_local_bounding_box' is not a member of 'libMesh::MeshTools'
   BoundingBox bbox = MeshTools::create_local_bounding_box(getMesh());
```

pbauman:
In a similar vein, GRINS isn't building because we adopted the new `FEMSystem::assembly` API after 1.2.0

<pre>
/femputer/pbauman/civet_build_testing_root/grins-libmesh/src/physics/src/multiphysics_sys.C:283:54: error: no matching function for call to ‘GRINS::MultiphysicsSystem::assembly(bool&, bool&, bool&, bool&)’
</pre>

jwpeterson:
I think this is probably good enough as it stands. There are a few more commits that could possibly be cherry-picked, but they were mostly for compiling in non-standard configurations and the more I try to cherry pick, the more potential there is for conflicts/breaking things. I'll post here again when the release and tarballs are created....

jwpeterson:
> I'll post here again when the release and tarballs are created....

New release is [available now](https://github.com/libMesh/libmesh/releases/tag/v1.2.1).

PR #1387, opened 2017-07-27
Close file descriptors returned by mkstemp

### LINKED ISSUE #1386: Unclosed filehandle build-up in FParser JIT compile
I'm using `mkstemp` to create temporary files. However that function actually opens the newly created fiIe. I don't store the returned handle, nor do I close the file (instead I open it again with `std::ofstream`).
This can cause failure if the maximum number of open file handles allowed by the OS is exhausted.

See https://github.com/idaholab/moose/issues/9549

dschwen:
I'm closing the file descriptors returned by `mkstemp` right away now. Two are later opened using a C++ stream and one by the compiler (passed in using `-o`). I also renamed variables to improve consistency.

Closes #1386

### DIFF
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -430,7 +430,7 @@ int FunctionParserADBase<Value_t>::AutoDiff(const std::string& var_name)
     unsigned int var_number = LookUpVarOpcode(var_name);
 
     // should and can we load a cached derivative?
-    std::string cache_file;
+    std::string cachename;
     const std::string jitdir = ".jitcache";
     if (cached)
     {
@@ -454,14 +454,14 @@ int FunctionParserADBase<Value_t>::AutoDiff(const std::string& var_name)
       sprintf(var_number_string, "%x", var_number - VarBegin);
 
       // function name
-      cache_file = jitdir + "/d_";
-      cache_file += result;
-      cache_file += "_";
-      cache_file += var_number_string;
+      cachename = jitdir + "/d_";
+      cachename += result;
+      cachename += "_";
+      cachename += var_number_string;
 
       // try to open cache file
       std::ifstream istr;
-      istr.open(cache_file.c_str(), std::ios::in | std::ios::binary);
+      istr.open(cachename.c_str(), std::ios::in | std::ios::binary);
       if (istr)
       {
         Unserialize(istr);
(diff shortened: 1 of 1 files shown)


contrib/fparser/fparser_ad.cc
```diff
@@ -677,18 +679,23 @@ bool FunctionParserADBase<Value_t>::JITCompileHelper(const std::string & Value_t
...
+  char objectname[] = "./tmp_jit_XXXXXX";
+  int objectfile = mkstemp(objectname);
+  if (objectfile == -1)
   {
-    std::cerr << "Error creating JIT tmp file " << object << ".\n";
+    std::cerr << "Error creating JIT tmp file " << objectname << ".\n";
     std::remove(ccname);
     return false;
   }
+  close(objectfile); // close file. Will be reopened by the compiler.
```
jwpeterson: It's too bad there's not a iostreams equivalent of mkstemp, or a way to construct an ofstream from an existing file descriptor, or some RAII thingy that would avoid this manual closing and reopening. Surely this impacts the performance of the JIT somehow???

dschwen: There are non portable ways to avoid a close/reopen. I don't think this will impact the performance significantly, though. This is not affecting the code path, where the JIT stuff is already cached at all.

dschwen: And this particular line cannot be avoided anyways.

jwpeterson: I guess we should probably just be doing `fprintf()`/C-based I/O for the files if we find there is a performance hit....

dschwen: I could do some profiling if you think open/close is an issue. But I guess this will be OS dependent...

dschwen: On a mac 1,000,000 open/close call pairs take 0.3s

dschwen: Ugh, on falcon (home FS) that test code takes 2.5 minutes. I'll do a proper test now.

dschwen:
Unrelated timeout in the `misc/line_source.test` test.

jwpeterson:
> Ugh, on falcon (home FS) that test code takes 2.5 minutes. I'll do a proper test now.

LOL.

dschwen:
There seems to be no significant difference. I performed three mesurements each. The scatter is due to fluctuating login node load (I monitored it, it was fairly constant, my request for an interactive shell on an empty node is still in the queue...)

* x axis is the number of JIT compiles (derivative caches are one less then that number)
* y axis is total execution time (I did the linear regression to back out the JIT time, but setup and teardown is negligable anyways)

image

PR #1389, opened 2017-08-03
Improved and documented building on Windows.

hrittich:
- Created Windows fix symlink script.
- Documented building on Windows.
- Fixed HTML in documentation.
- Fixed build of qhull on Windows.
  - Removed src/libquhull from the include path, because it shadows the
    windows system header io.h.
  - Included string.h in the exodusII IO module hack that is needed to
    include the exodus library into a separate namespace.

### DIFF
--- a/contrib/bin/fix_windows_symlinks.sh
+++ b/contrib/bin/fix_windows_symlinks.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+
+LIBMESH_ROOT=$(shell git rev-parse --show-toplevel)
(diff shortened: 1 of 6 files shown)


jwpeterson:
The new script has DOS line endings. If that isn't strictly necessary for having the script work on Windows, I'd like to change them to "Unix" style before merge.

hrittich:
I will try out, if the line endings matter.

hrittich:
Unix line endings are fine.

PR #1390, opened 2017-08-07
Distributed checkpoint fixes

roystgnr:
This fixes --distributed-mesh recovery on https://github.com/idaholab/moose/issues/8890 for me.

It needs more testing (and possibly more test cases to be added) before we merge, to make sure I haven't regressed checkpoints elsewhere.

As is (currently) S.O.P. with CheckpointIO, this patch changes the file format without caring in the slightest about backwards compatibility with old files.  The ensuing mixed feelings of horror and liberation are hard to describe.

### DIFF
--- a/include/mesh/checkpoint_io.h
+++ b/include/mesh/checkpoint_io.h
@@ -153,7 +153,7 @@ private:
                            const std::set<const Elem *, CompareElemIdsByLevel> & elements) const;
 
   /**
-   * Write the remote_elem neighbor links for part of a mesh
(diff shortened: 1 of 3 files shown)


roystgnr:
Those dbg failures - three timeouts, all look unrelated to CheckpointIO?

permcody:
Yeah - We have cleaned up a lot of dbg timeouts but apparently not all. I know that those tests have failed in the past. Feel free to invalidate.

PR #1393, opened 2017-08-11
Fixed building with AMR disabled.

hrittich:
Building with --disable-amr failed. This should fix it.

### DIFF
--- a/src/mesh/checkpoint_io.C
+++ b/src/mesh/checkpoint_io.C
@@ -322,12 +322,14 @@ void CheckpointIO::write_connectivity (Xdr & io,
       elem_data[2] = elem.processor_id();
       elem_data[3] = elem.subdomain_id();
 
+#ifdef LIBMESH_ENABLE_AMR
       if (elem.parent() != libmesh_nullptr)
         {
           elem_data[4] = elem.parent()->id();
           elem_data[5] = elem.parent()->which_child_am_i(&elem);
         }
       else
+#endif
         {
           elem_data[4] = DofObject::invalid_processor_id;
           elem_data[5] = DofObject::invalid_processor_id;
@@ -390,6 +392,7 @@ void CheckpointIO::write_remote_elem (Xdr & io,
             }
         }
 
+#ifdef LIBMESH_ENABLE_AMR
       if (elem.has_children())
         for (unsigned int c = 0; c != elem.n_children(); ++c)
         {
@@ -401,6 +404,7 @@ void CheckpointIO::write_remote_elem (Xdr & io,
               child_numbers.push_back(c);
             }
         }
+#endif
     }
 
   io.data(elem_ids, "# remote neighbor elem_ids");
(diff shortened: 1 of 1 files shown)


contrib/netcdf/4.4.1.1/include/netcdf_meta.h
```diff
@@ -38,10 +38,10 @@
 #define NC_VERSION       "4.4.1.1"
 
 #define NC_HAS_NC2       1 /*!< API version 2 support. */
-#define NC_HAS_NC4       1 /*!< API version 4 support. */
+#define NC_HAS_NC4       0 /*!< API version 4 support. */
 #define NC_HAS_HDF4      0 /*!< hdf4 support. */
-#define NC_HAS_HDF5      1 /*!< hdf5 support. */
-#define NC_HAS_SZIP      1 /*!< szip support (hdf5 only) */
```
roystgnr: How are these three changes necessary?

include/numerics/dense_vector.h
```diff
@@ -28,7 +28,7 @@
 
 #ifdef LIBMESH_HAVE_EIGEN
 #include "libmesh/ignore_warnings.h"
-#include <Eigen/Dense>
+#include <Eigen/Core>
 #include "libmesh/restore_warnings.h"
```
roystgnr: This looks like a correct change, but probably belongs split into a separate PR to make bisecting easier if people encounter problems - surely including too much from Eigen wasn't causing --disable-amr problems?

include/numerics/eigen_core_support.h
```diff
@@ -45,7 +45,7 @@
 #include "libmesh/ignore_warnings.h"
 
 // Eigen includes
-#include <Eigen/Dense>
```
roystgnr: This, on the other hand, I'm not so sure about.  radial_basis_interpolation.C uses Eigen dense solvers, which aren't in Core, are they?

src/solution_transfer/radial_basis_interpolation.C
```diff
@@ -27,6 +27,9 @@
 #include "libmesh/libmesh_logging.h"
 #include "libmesh/eigen_core_support.h"
 
+#ifdef LIBMESH_HAVE_EIGEN
+#include <Eigen/Dense>
+#endif
```
roystgnr: Ah, now I see.  Disregard what I said about the eigen_core_support.h change above.

include/numerics/eigen_core_support.h
roystgnr: Never mind, just saw the corresponding change there.

roystgnr:
The CheckpointIO fixes look good; sorry for messing that up for you!

The Eigen changes look good, but should probably be split into a separate PR, unless they really do fix something specific to --disable-amr.

The NetCDF changes I don't get.

hrittich:
Sorry. I messed up the commit.

Those should have been two separate things...

hrittich:
This should be it, to fix the --disable-amr.

roystgnr:
:+1: once the tests pass.. and probably even if they fail; there's no way these changes should trigger anything that isn't a false positive.

brianmoose:
FYI, there is already an optional CIVET test for this, which I added to this PR.

jwpeterson:
Oh, I see, this was just recently broken in 314e09c473. Thanks @brianmoose for attaching the `--disable-amr` test.

PR #1394, opened 2017-08-11
Changed includes from Eigen/Dense to Eigen/Core.

hrittich:
This should improve the build times, because only necessary files are
included.

Eigen is a rather huge library. Including things like LU decomposition everywhere should not be necessary.


roystgnr:
:+1: unless anyone else has objections.

Issue #1395, opened 2017-08-14
ExodusII gives error when including infinite elements


BalticPinguin:
Hi everyone,
after some time being concerned with different things, I am back to my FEM/InFEM code and I realised that the exodus-interface seems to have changed in the meantime. 
Before (I think my last version was from February), exodus just ignored the infinite elements block and wrote only the FEM part. But now, the program fails with the message
```
WARNING: nodal_soln(...) does _not_ work for infinite elements.
 Will return an empty nodal solution.  Use 
 InfFE<Dim,T_radial,T_map>::compute_data(..) instead!
Exodus Library Error: [ex_put_concat_all_blocks]
	Error: failed to complete element block definition in file id 65536
	[-62] One or more variable sizes violate format constraints
Error writing element blocks.
Stack frames: 9
0: libMesh::print_trace(std::ostream&)
1: libMesh::MacroFunctions::report_error(char const*, int, char const*, char const*)
2: libMesh::ExodusII_IO_Helper::write_elements(libMesh::MeshBase const&, bool)
3: libMesh::ExodusII_IO::write_nodal_data_common(std::string, std::__debug::vector<std::string, std::allocator<std::string> > const&, bool)
4: libMesh::ExodusII_IO::write_nodal_data(std::string const&, std::__debug::vector<std::complex<double>, std::allocator<std::complex<double> > > const&, std::__debug::vector<std::string, std::allocator<std::string> > const&)
5: libMesh::MeshOutput<libMesh::MeshBase>::write_equation_systems(std::string const&, libMesh::EquationSystems const&, std::__debug::set<std::string, std::less<std::string>, std::allocator<std::string> > const*)
6: /scratch/tm162/bin/FreeWilly/FrWll-dbg() [0x4a9b98]
7: __libc_start_main
8: /scratch/tm162/bin/FreeWilly/FrWll-dbg() [0x4a5f19]
[0] src/mesh/exodusII_io_helper.C, line 1420, compiled Aug 13 2017 at 17:54:52
application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0
```
which indicates that it tries to also write the infinite elements.
Having a brief look into the changes of respective code, it looks to me like there were some major changes in the code-structure.
Can someone of those that are into this function fix it again?
If you don't have time to fix it, I can do it but I didn't touch this part yet, so it might take some time for me to understand how the work-flow of this is. In this case I appreciate any advice.
Best,
Hubert

[...]

PR #1398, opened 2017-08-14
Move counter increment to for-loop clause.

jwpeterson:
This should hopefully fix issue #1395 that was discovered while
writing meshes with infinite elements.


jwpeterson:
@BalticPinguin would you try pulling this branch and see if it fixes the issue you raised in #1395? It passes our internal infinite element testing, but that was not catching the problem before so I'm not sure how good its coverage is...

jwpeterson:
I'm fairly confident this is the right fix, so I'm going to go ahead and merge.

PR #1399, opened 2017-08-14
New FEMContext constructor

dknez:
Added a new FEMContext constructor which allows user to set the quadrature order for each dimension.

### DIFF
--- a/include/systems/fem_context.h
+++ b/include/systems/fem_context.h
@@ -69,6 +69,13 @@ public:
   FEMContext (const System & sys);
 
   /**
+   * Constructor.  Specify the extra quadrature order instead
+   * of getting it from \p sys.
+   */
+  explicit
+  FEMContext (const System & sys, int extra_quadrature_order);
+
+  /**
    * Destructor.
    */
   virtual ~FEMContext ();
@@ -1136,7 +1143,17 @@ protected:
    */
   UniquePtr<QBase> _edge_qrule;
 
+  /**
+   * The extra quadrature order for this context.
+   */
+  int _extra_quadrature_order;
+
 private:
+  /**
+   * Helper function used in constructors to set up internal data.
+   */
+  void init_internal_data(const System & sys);
+
   /**
    * Uses the coordinate data specified by mesh_*_position configuration
    * to set the geometry of \p elem to the value it would take after a fraction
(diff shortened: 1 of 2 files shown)


dknez:
@roystgnr this is what I had in mind. Let me know if it looks OK to you?

roystgnr:
Is there a use case for independent orders for each dimension, or did you put that feature in just because it was so easy to implement since we already had independent objects for each dimension?

roystgnr:
Gah.  I can't see the femputer failure details from behind the INL firewall.

dknez:
Regarding this:

> Is there a use case for independent orders for each dimension, or did you put that feature in just because it was so easy to implement since we already had independent objects for each dimension?

I don't need that right now, but I can imagine it being useful in the future, e.g. if you want reduced integration for 1D beam elements, but not for 3D elements. It was easy to add, so I figured I might as well do it that way.

However, I'd be OK to just use one Order for all dims, if you prefer.

roystgnr:
Nah, this way is fine, just curious.

dknez:
@roystgnr I'll leave it up to you regarding merging this. From my end this is ready to merge.

dknez:
P.S. I made another update in which I added a third constructor takes extra_quadrature_order as an argument. I think is helpful since it would allow us to, for example, make multiple FEMContexts for the same System with different values of extra_quadrature_order without changing the system object.

This is useful because in some reduced integration formulations some terms are integrated with full quadrature order and other terms integrated with reduced quadrature order, and creating multiple contexts with different values extra_quadrature_order is a convenient way to handle this.

dknez:
P.P.S. I've gone full circle here and returned to your original suggestion. After doing some more implementation/testing I agree that it is better to stick to relative quadrature after all! As a result, this PR is quite minor now: I only add one extra constructor that allows us to specify extra_quad_order instead of getting it from sys, with the motivation for this explained in my previous comment. There is no more "absolute quad order" constructor, since in practice I do not think that is useful after all (sorry for the misleading back-and-forth about that).

roystgnr:
No worries.  If this works for you, it looks good to merge to me.

(although, if you want to shave off a few lines of code, remember that we're allowing C++11 in the library now: you could use delegating constructors and be the author of The First Modern C++ In LibMesh)

jwpeterson:
> remember that we're allowing C++11 in the library now

Let's flip the default on `--enable-cxx11-required` before we commit C++11 code, though.

dknez:
> (although, if you want to shave off a few lines of code, remember that we're allowing C++11 in the library now: you could use delegating constructors and be the author of The First Modern C++ In LibMesh)

haha that's too much pressure! I think I'll just merge as-is. Though good to know that we now allow C++11, that'll be very nice going forward.

BalticPinguin:
Sorry for the late reply; but yes, it works now :-)

Issue #1392, opened 2017-08-11
Compile CFD code using libmesh (source embeded) failed on ALCF-Mira


felixit:
In fact, this issue is my original issue, which is related with #1391 . I am trying to compile my own CFD code on ALCF-Mira (IBM Blue Gene/Q machines), but always get some errors related with libMesh. My code uses some source files from libMesh. My code is written in C/C++. My code has been compiled on many Linux clusters successfully, but maybe this IBM machines have some special things to be noted. Here is the output of the 'make':

`[ 94%] Built target hpMusic_core
[ 95%] Linking CXX executable ../../../bin/hpMusic_debug
/home/fljia/hpMusic/../hpMesh_gui/debug/lib/libhpMesh_core_debug.a(xdr_cxx.C.o): In function `libMesh::Xdr::open(std::string const&)':
/home/fljia/hpMesh_gui/src/core/mesh/xdr_cxx.C:239: undefined reference to `xdrstdio_create'
/home/fljia/hpMusic/../hpMesh_gui/debug/lib/libhpMesh_core_debug.a(xdr_cxx.C.o): In function `xdr_translate<std::basic_string<char> >':
/home/fljia/hpMesh_gui/src/core/mesh/xdr_cxx.C:482: undefined reference to `xdr_string'
/home/fljia/hpMusic/../hpMesh_gui/debug/lib/libhpMesh_core_debug.a(xdr_cxx.C.o): In function `void libMesh::Xdr::data_stream<double>(double*, unsigned int, unsigned int)':
/home/fljia/hpMesh_gui/src/core/mesh/xdr_cxx.C:837: undefined reference to `xdr_vector'
/home/fljia/hpMusic/../hpMesh_gui/debug/lib/libhpMesh_core_debug.a(xdr_cxx.C.o): In function `void libMesh::Xdr::data_stream<float>(float*, unsigned int, unsigned int)':
/home/fljia/hpMesh_gui/src/core/mesh/xdr_cxx.C:923: undefined reference to `xdr_vector'
/home/fljia/hpMusic/../hpMesh_gui/debug/lib/libhpMesh_core_debug.a(xdr_cxx.C.o): In function `void libMesh::Xdr::data_stream<__float128>(__float128*, unsigned int, unsigned int)':
/home/fljia/hpMesh_gui/src/core/mesh/xdr_cxx.C:1027: undefined reference to `xdr_vector'
[...]

PR #1400, opened 2017-08-14
Fix a few small issues with --disable-xdr builds

jwpeterson:
I did not experience any of the XDR-related errors posted in #1392, but there were a few tests and examples that weren't correctly guarded when `!LIBMESH_HAVE_XDR`. Those should now should be fixed.


jwpeterson:
Unrelated timeout is unrelated.

Issue #1391, opened 2017-08-09
Can not compile on IBM Blue Gene (ALCF-Mira)


felixit:
I am trying to install the libmesh on the IBM Blue Gene machine on ALCF-Mira. I am using the gcc 4.8.4. The ./configure works fine, but there are several errors during the make process. I attach both of the configure log and the make log files.

**Below is the configure log file:**

[config.log.txt](https://github.com/libMesh/libmesh/files/1213007/config.log.txt)

**Below is the make log file:**
`Making all in include
make[1]: Entering directory `/gpfs/mira-home/fljia/library/libmesh/include'
make  all-recursive
make[2]: Entering directory `/gpfs/mira-home/fljia/library/libmesh/include'
Making all in libmesh
make[3]: Entering directory `/gpfs/mira-home/fljia/library/libmesh/include/libmesh'
make  all-am
make[4]: Entering directory `/gpfs/mira-home/fljia/library/libmesh/include/libmesh'
make[4]: Nothing to be done for `all-am'.
make[4]: Leaving directory `/gpfs/mira-home/fljia/library/libmesh/include/libmesh'
make[3]: Leaving directory `/gpfs/mira-home/fljia/library/libmesh/include/libmesh'
make[3]: Entering directory `/gpfs/mira-home/fljia/library/libmesh/include'
make[3]: Leaving directory `/gpfs/mira-home/fljia/library/libmesh/include'
make[2]: Leaving directory `/gpfs/mira-home/fljia/library/libmesh/include'
make[1]: Leaving directory `/gpfs/mira-home/fljia/library/libmesh/include'
Making all in contrib
make[1]: Entering directory `/gpfs/mira-home/fljia/library/libmesh/contrib'
Making all in boost/include
make[2]: Entering directory `/gpfs/mira-home/fljia/library/libmesh/contrib/boost/include'
make[2]: Nothing to be done for `all'.
make[2]: Leaving directory `/gpfs/mira-home/fljia/library/libmesh/contrib/boost/include'
Making all in eigen/eigen
make[2]: Entering directory `/gpfs/mira-home/fljia/library/libmesh/contrib/eigen/3.2.9'
make[2]: Nothing to be done for `all'.
make[2]: Leaving directory `/gpfs/mira-home/fljia/library/libmesh/contrib/eigen/3.2.9'
Making all in gmv
make[2]: Entering directory `/gpfs/mira-home/fljia/library/libmesh/contrib/gmv'
[...]

PR #1403, opened 2017-08-15
Miscellaneous fixes for --enable-all-static builds

jwpeterson:
This should address most of the issues raised in #1391.


roystgnr:
:+1: and thanks.

PR #1405, opened 2017-08-15
Add DofMap::dof_owner() search

roystgnr:
This can be done without the current API, but I'd rather have a
short-to-invoke O(log N_procs) method available in libMesh, not force
users to write a longer O(N_procs) linear search or an even longer
(because we don't give raw access to the _end_dfs array) binary
search.

This is going to be necessary for MOOSE to use to fix one of the last bugs in https://github.com/idaholab/moose/issues/1500

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -578,6 +578,16 @@ public:
   dof_id_type end_dof() const
   { return this->end_dof(this->processor_id()); }
 
+  /**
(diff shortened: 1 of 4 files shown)


roystgnr:
Going to merge right away - the only test coverage on the new feature is in our unit tests and has already passed, and I've got two more PRs that will depend on this.

Issue #1396, opened 2017-08-14
Libmesh linking fails with address sanitizer enabled


dschwen:
Build fails with

```
Undefined symbols for architecture x86_64: "___asan_after_dynamic_init", referenced from:
...
```

with

```
--enable-sanitize=devel
```

when using the latest Xcode (8.3.3). It worked with at least 8.1

jwpeterson:
Can you come up with a simple test code that fails? We test the sanitize compiler flags with a super simple main program in m4/sanitize.m4, but presumably this passes on your system, while using sanitizer flags on "real" codes fails to link for some reason.

dschwen:
Spent 30mins on it. No success. Hard to figure out what exactly triggers it, as it fails on linking, not on compile.

dschwen:
Ok, got it:

Makefile:
```
test: test.o
	clang++ test.o -o test -L/opt/moose/llvm-3.9.0/lib/clang/3.9.0/lib/darwin -lclang_rt.osx

test.o: test.cc
	clang++ -std=gnu++11 -march=native -fsanitize=address test.cc -c -o test.o
```

test.cc:
```
#include <vector>
#include <iostream>

typedef unsigned long dof_id_type;
typedef dof_id_type index_t;
typedef std::vector<index_t> index_buffer_t;
index_buffer_t _idx_buf;

int main()
{
	index_buffer_t().swap(_idx_buf);
	return _idx_buf.empty() ? 0 : 1;
}
```

dschwen:
The `-lclang_rt.osx` is key

jwpeterson:
What if you take out that`-march=native` flag?

I think you are the only person using this... maybe it has a bad interaction with the address sanitizer.

jwpeterson:
I ask because I don't think we have control over `-lclang_rt.osx`, it must be something coming from libtool?

dschwen:
Same deal ` -march=native` does not matter

jwpeterson:
I believe I have traced this issue to needing to pass the `-fsanitize=address` flag to the linker as well as the compiler when compilation is done in separate stages. Doing so fixes the example posted above, anyway. I think the fix will be to ensure that the sanitize flag gets added to the `LDFLAGS` variable during configure...

jwpeterson:
Fixed by #1406.

PR #1406, opened 2017-08-16
Fix --enable-sanitize builds

jwpeterson:
We were not passing the `-fsanitize=address` flag to the compiler doing the linking since it was (apparently) never required before. This fixes that issue by using the libtool `-Wc,` [syntax](https://www.gnu.org/software/libtool/manual/html_node/Link-mode.html#Link-mode) for passing flags to the compiler during linking. (I checked, and simply adding the flag to libmesh_LDFLAGS did not work).

Refs #1396.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -13,7 +13,7 @@ AM_LDFLAGS       = $(libmesh_LDFLAGS)
 # with libtool, because Windows requires all shared
 # libraries to include all used symbols. This flag
 # should do no harm on other platforms.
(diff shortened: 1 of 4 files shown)


jwpeterson:
@roystgnr Can we include this branch with your libmesh update PR in MOOSE? @dschwen apparently compiles with the sanitizer flags fairly regularly, and this should fix that for him.

roystgnr:
Got it.

PR #1407, opened 2017-08-16
Remove 'old' Makefiles from contrib and unit test directories.

jwpeterson:
These date back to before we changed over to the automake-based build
system and are no longer used for anything. If we ever need to refer
back to them, they are in the git history.


contrib/gzstream/Makefile.old
```diff
@@ -1,107 +0,0 @@
-# ============================================================================
-# gzstream, C++ iostream classes wrapping the zlib compression library.
-# Copyright (C) 2001  Deepak Bandyopadhyay, Lutz Kettner
-#
```
roystgnr: Is their license + copyright info elsewhere in the package too?  We don't want to strip that out entirely.

jwpeterson: Yeah, the same stuff is in gzstream.{h,C}.

roystgnr:
:+1:, thanks

PR #1408, opened 2017-08-17
Put sideset/nodeset names in CheckpointIO header

roystgnr:
At present we need these names to be present on all processors if we
want name maps to be replicated everywhere even in early phases of
N->M restarts where M > N.

At some point in the future we might want to distribute these maps,
but for the moment it's very very unlikely that users will have so
many boundary names that scaling will be a problem.

This breaks backwards compatibility with old cpr/cpa files, and I'm
*just* about to check some cpr/cpa files into MOOSE, so after this
passes testing I'd like to sit on it until MOOSE is ready for their
next libMesh update.

### DIFF
--- a/src/mesh/checkpoint_io.C
+++ b/src/mesh/checkpoint_io.C
@@ -124,6 +124,11 @@ void CheckpointIO::write (const std::string & name)
 
       // write subdomain names
       this->write_subdomain_names(io);
+
+      // write boundary id names
+      const BoundaryInfo & boundary_info = mesh.get_boundary_info();
+      write_bc_names(io, boundary_info, true);  // sideset names
+      write_bc_names(io, boundary_info, false); // nodeset names
     }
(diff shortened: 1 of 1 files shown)


roystgnr:
I think I see how to instead support replicated boundary names in a way that *doesn't* break backwards compatibility, at the cost of a tiny bit of efficiency... but I now also realize that I'm about to have to break backwards compatibility *anyway*, if we want to reenable the MOOSE tests we had to turn off in https://github.com/idaholab/moose/pull/9727, so we might as well merge this now.  With those tests disabled there shouldn't be any MOOSE test breakage anymore, and anyway I hope to have all the N->M and Linux->Mac and 64bit->32bit issues squared away by the time MOOSE is ready for their next libMesh update, in which case we can reenable their mesh/checkpoint tests at the same time.

PR #1409, opened 2017-08-17
Checkpoint N->M restart fixes

roystgnr:
Previously we could throw overzealous asserts in N->M restarts, or
even end up with corrupted child pointers in N->M AMR restarts.

This is an API and format backwards-compatible change.

### DIFF
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -1065,6 +1065,12 @@ public:
 
   /**
    * \returns A constant pointer to the \f$ i^{th} \f$ child for this element.
+   * For internal use only - skips assertions about null pointers.
+   */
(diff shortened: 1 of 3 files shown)


permcody:
@jwpeterson - I noticed all of these skipped tests in MOOSE today while doing training. I looks like we've done an update. Can these be un-skipped now?

jwpeterson:
The CheckpointIO file format has all changed now, so the skipped tests don't run at all; for example I get:
```
*** ERROR ***
ERROR: cannot locate header file for input '/Users/foo/bar/moose/test/tests/mesh/checkpoint/checkpoint_split_in.2.cpr'
```
for the first one. I will bring in @rwcarlsen for help on this.

PR #1410, opened 2017-08-23
Small update to RBEIMConstruction

dknez:
Use the base class version of greedy_termination_test in RBEIMConstruction.


roystgnr:
This is your bailiwick, but I don't see any obvious problems.  This already gets test coverage via examples/reduced_basis, right?

dknez:
Yes, I will merge this. It's a very minor change, and only affects the reduced basis code, and has coverage in the examples.

Issue #1412, opened 2017-08-30
Let TetGenIO read in subdomain IDs


YaqiWang:
Can we add an integer argument in `TetGenIO::read` to indicate that TetGen ele file contains an attribute which can be used to set the subdomain ID? `-1` means no subdomain ID, otherwise, we used it to access `element_attributes` and assign element subdomain ID. If you agree, I can send in a small PR on this.

YaqiWang:
Oh, one more thing, although looks like we can use *.node to indicate a tetgen file based on line https://github.com/libMesh/libmesh/blob/master/src/mesh/namebased_io.C#L219, but actually the code will treat it as a nemsis file.

jwpeterson:
> Oh, one more thing, although looks like we can use *.node to indicate a tetgen file based on line https://github.com/libMesh/libmesh/blob/master/src/mesh/namebased_io.C#L219, but actually the code will treat it as a nemsis file in previous lines

Are you talking about this logic?
```
if (name.rfind(".nem") + 4 == name.size() ||
      name.rfind(".n") + 2 == name.size())
```
I don't think either of these if-statements will trigger for a filename ending in ".node"

jwpeterson:
NM, I found it on line 150. Yeah, that looks like a bug.

YaqiWang:
Probably!

YaqiWang:
I am thinking that the reader will need one more integer provided by us, which is not desirable. Let me read tetgen webpage more carefully to see if there is a sign in the file to indicate the existence of subdomain IDs.

jwpeterson:
> NM, I found it on line 150. Yeah, that looks like a bug.

Nevermind again, I think the real bug may be in the `is_parallel_file_format()` function. The `name.rfind(".n")` check will cause the function to incorrectly return `true`.

```
inline bool
NameBasedIO::is_parallel_file_format (const std::string & name)
{
  return ((name.rfind(".xda") < name.size()) ||
          (name.rfind(".xdr") < name.size()) ||
          (name.rfind(".nem") < name.size()) ||
          (name.rfind(".n") < name.size())   ||
          (name.rfind(".cp") < name.size())
          );
}
```

jwpeterson:
> Can we add an integer argument in TetGenIO::read to indicate that TetGen ele file contains an attribute which can be used to set the subdomain ID? -1 means no subdomain ID, otherwise, we used it to access element_attributes and assign element subdomain ID. If you agree, I can send in a small PR on this.

[...]

PR #1413, opened 2017-08-31
Check for exact file extensions in NameBasedIO::is_parallel_file_format()

jwpeterson:
As pointed out by @YaqiWang in #1412, this function previously
returned true for ".node" files since they incorrectly matched the
".n" check, causing them to later be treated as Nemesis files. To be
consistent with a similar check in namebased_io.C, I also updated the
".nem" check.

There may be other places in the code where these rfind() checks could
be updated, but they can't all just be changed, because (I think) we
still want to support reading meshes from files with names like
out.xdr.0000, etc. where some number or other designation appears
after the "normal" file extension.

Nemesis (and parallel exodus) files are a special case in this regard,
since the user specifies a "base" filename while reading and then
libmesh automatically appends the number of processors and the
processor id (see Nemesis_IO_Helper::construct_nemesis_filename())
so those filenames cannot have extra characters after the file extension.

### DIFF
--- a/include/mesh/namebased_io.h
+++ b/include/mesh/namebased_io.h
@@ -125,8 +125,8 @@ NameBasedIO::is_parallel_file_format (const std::string & name)
 {
   return ((name.rfind(".xda") < name.size()) ||
           (name.rfind(".xdr") < name.size()) ||
-          (name.rfind(".nem") < name.size()) ||
-          (name.rfind(".n") < name.size())   ||
+          (name.rfind(".nem") + 4 == name.size()) ||
+          (name.rfind(".n") + 2 == name.size()) ||
           (name.rfind(".cp") < name.size())
           );
 }


jwpeterson:
I don't see how this change could have affected these tests, as they don't use filenames with the extension in question. Several of the other tests in this directory are skipped because they require a libmesh change, maybe these need to be skipped as well...
```
mesh/checkpoint.test_2a....................................................................... FAILED (CRASH)
mesh/checkpoint.test_4a....................................................................... FAILED (CRASH)
mesh/checkpoint.test_8a....................................................................... FAILED (CRASH)
```

permcody:
We may want to do another libMesh update sooner than later. There are
already a few nice features that can help us get this checkpoint format
support finished up which is pretty high priority right now. Maybe next
week?

jwpeterson:
The problem is that these tests are failing with the current libmesh master, but not with the libmesh version in MOOSE. If we do a libmesh update now they will start failing 😢  @roystgnr could these new failures have snuck in somehow with libMesh/libmesh#1409 somehow? I'm seeing the same failures on my Mac.

roystgnr:
Wait, didn't you disable those tests?  Those failures are *expected* as of #1408, which broke format compatibility.

Shoot, you only disabled the non-ASCII tests, didn't you?  Because those were the only ones affected by the integer-type-selection issues?  But then how did #1408 ever pass?

roystgnr:
Anyway, there's one more change (again format-breaking, sorry) that I need to make to CheckpointIO, to fix the integer-type-selection issues in the same way that our XDR support fixes them, and after that we should be good to do a libMesh update to MOOSE IMHO, which will include re-generated checkpoint files for that test.

jwpeterson:
> Shoot, you only disabled the non-ASCII tests, didn't you? Because those were the only ones affected by the integer-type-selection issues? 

Right.

> But then how did #1408 ever pass?

Yeah, that's the real mystery here. I'll add a MOOSE PR temporarily disabling the other tests in that directory as well...

jwpeterson:
> Yeah, that's the real mystery here.

I just searched through the lists of passed tests in [The PR for #1408](https://www.moosebuild.org/job/109848/) and I don't see the checkpoint.test_2a, 4a, 8a test names in the list at all... So they were never run, but I'm not sure how this happened.

jwpeterson:
Just created idaholab/moose#9761 to skip the remaining checkpoint tests.

jwpeterson:
Oops... libmesh pulls MOOSE master and the commit for skipping tests hasn't made it there yet. Hence it failed again.

PR #1415, opened 2017-08-31
allgather(vector<string>) overload

roystgnr:
It looks like I need essentially this code in order to work on idaholab/moose#9700 and it's generic enough that it ought to be useful for other libMesh users as well; it's just a new overload on our existing allgather(vector).

There's unit test coverage, but I intend to wait to merge until after I'm confident that this is working correctly in its final use case too.  After that, if MOOSE really can [do another libMesh update sooner than later](https://github.com/libMesh/libmesh/pull/1413#issuecomment-326351328) then I'll follow it up with the MOOSE PR for that issue.

### DIFF
--- a/include/parallel/parallel_implementation.h
+++ b/include/parallel/parallel_implementation.h
@@ -3302,6 +3302,93 @@ inline void Communicator::allgather(std::vector<T> & r,
 
 
 template <typename T>
+inline void Communicator::allgather(std::vector<std::basic_string<T> > & r,
+                                    const bool identical_buffer_sizes) const
+{
+  if (this->size() < 2)
+    return;
+
+  LOG_SCOPE("allgather()", "Parallel");
+
+  if (identical_buffer_sizes)
+    {
+      libmesh_assert(this->verify(r.size()));
+
+      // identical_buffer_sizes doesn't buy us much since we have to
+      // communicate the lengths of strings within each buffer anyway
+      if (r.empty())
+        return;
+    }
+
+  // Concatenate the input buffer into a send buffer, and keep track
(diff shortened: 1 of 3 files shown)


roystgnr:
Github now shows me a little squiggly line when I've written commit logs or PR descriptions which will inadvertently auto-close issues.  That's nice.  I like finding out I'm an idiot in subtle easily-reversable ways.

include/parallel/parallel_implementation.h
```diff
@@ -3302,6 +3302,90 @@ inline void Communicator::allgather(std::vector<T> & r,
...
+    {
+      displacements[i] = globalsize;
+      globalsize += concat_sizes[i];
+    }
+
+  // Check for quick return
+  if (globalsize == 0)
+    return;
+
+  StandardType<T> send_type(&concat_src[0]);
```
jwpeterson: Have you tested the stupid case where the vector of strings is initially empty on all processors? I think `concat_src` might be empty then (?) and indexing into the zeroth entry could fail.

roystgnr: Damn, you're right.  The stupid "empty on all processors" case should only work if the user set identical_buffer_sizes.  The "empty on one processor" case (which is not stupid, I'm pretty sure my use case will actually hit it on not-too-crazy processor counts) won't work at all.  I'll put it in a fixup since I need to interactive-rebase anyway to fix that commit message.

roystgnr: Thanks for catching that!

jwpeterson:
:+1:

PR #1422, opened 2017-09-11
Update README.md

jwpeterson:
The default for new users should be out-of-tree builds and *not*
running ./bootstrap, as this can potentially cause a large and
somewhat confusing diff for new users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,26 +1,16 @@
-libMesh-specific configuration options
-======================================
-
-For generic installation instructions, refer also to the INSTALL file.
-
-Build Process
--------------
-0. ./bootstrap (optional, needed *only* if you add/modify m4 files or configure.ac)
-
-1. ./configure --prefix=/path/to/libmesh/install
-
-2. make
-
-3. make check
-   (optional, runs the example programs and unit tests when possible)
-
-4. make install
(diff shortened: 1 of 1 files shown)


jwpeterson:
I'm going to cancel all the testing since this is just a README file update...

README.md
```diff
@@ -41,51 +31,40 @@ the library are:
 
 * oprof: Optimized code path with compiler flags suitable for use with oprofile.
 
-
 To select a set of methods, you can pass them to configure in one of two ways:
 
-    $ ./configure --with-methods="opt dbg devel"
+    ../configure --with-methods="opt dbg devel"
```
roystgnr: If we remove the leading dollar sign, does that also undo the monospace font?

    $ ./test

versus

    ../test

No, looks like it's just the indentation that markdown cares about.  Ok then.

jwpeterson: Right... the dollar sign was meant to represent a shell prompt, but over the years I've found that it confuses people (do I type a dollar sign?) as much or more than it helps.

jwpeterson: Also, I used GitHub's in-browser editor (with preview) to make the markdown changes, and I think everything is still going to render correctly.

PR #1423, opened 2017-09-11
GCC 7 warning fixes

roystgnr:
There are still fallthrough warnings coming from quadrature .C files, but I can't seem to figure out how to get gcc to ignore those; it's as if the nested switch statements confuse the parser there.

Still, the output from this PR has *far* fewer false positives than it did before, thanks to the header fixes.

### DIFF
--- a/include/base/libmesh_common.h
+++ b/include/base/libmesh_common.h
@@ -600,9 +600,9 @@ inline Tnew libmesh_cast_int (Told oldvar)
 // We should add support for standard C++17 [[fallthrough]]
 // Maybe [[clang::fallthrough]] or [[gcc::fallthough]] too?
 #if defined(__GNUC__) && (__GNUC__ > 6)
-#define libmesh_fallthrough __attribute__((fallthrough))
+#define libmesh_fallthrough() __attribute__((fallthrough))
 #else
-#define libmesh_fallthrough
+#define libmesh_fallthrough() ((void) 0)
 #endif
 
 } // namespace libMesh
--- a/include/base/libmesh_common.h
+++ b/include/base/libmesh_common.h
@@ -596,6 +596,15 @@ inline Tnew libmesh_cast_int (Told oldvar)
 #define libmesh_final
 #endif
 
+// Define backwards-compatible fallthrough attribute.
+// We should add support for standard C++17 [[fallthrough]]
+// Maybe [[clang::fallthrough]] or [[gcc::fallthough]] too?
+#if defined(__GNUC__) && (__GNUC__ > 6)
+#define libmesh_fallthrough __attribute__((fallthrough))
+#else
+#define libmesh_fallthrough
+#endif
+
 } // namespace libMesh
 
 
--- a/include/utils/hashword.h
+++ b/include/utils/hashword.h
@@ -172,12 +172,12 @@ uint32_t hashword(const uint32_t * k, size_t length, uint32_t initval=0)
   switch(length)                     // all the case statements fall through
     {
     case 3 : c+=k[2];
-      libmesh_fallthrough;
+      libmesh_fallthrough();
     case 2 : b+=k[1];
-      libmesh_fallthrough;
+      libmesh_fallthrough();
     case 1 : a+=k[0];
       final(a,b,c);
-      libmesh_fallthrough;
+      libmesh_fallthrough();
     default:     // case 0: nothing left to add
       break;
     }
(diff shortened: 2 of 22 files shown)


include/utils/hashword.h
```diff
@@ -172,9 +172,12 @@ uint32_t hashword(const uint32_t * k, size_t length, uint32_t initval=0)
   switch(length)                     // all the case statements fall through
     {
     case 3 : c+=k[2];
+      /* FALLTHROUGH */
```
jwpeterson: Is the compiler parsing this comment somehow? I'm not sure I'm crazy about adding these all over the place...

src/fe/fe_l2_lagrange_shape_3D.C
```diff
@@ -1248,9 +1248,10 @@ Real FE<3,L2_LAGRANGE>::shape_second_deriv(const ElemType type,
                              << std::endl;
               warning_given_HEX20 = true;
             }
+            /* FALLTHROUGH */
 
-            // triquadraic hexahedral shape funcions
           case HEX27:
+            // triquadraic hexahedral shape funcions
```
jwpeterson: Hmm... GitHub thinks you changed this line (whitespace?) but I don't see it. Might as well fix the 50% of the words that are misspelled too.

include/utils/hashword.h
roystgnr: It is.  Search https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html for "fallthrough".

I'm a little dubious of syntactically-meaningful comments too, but the only standard way to suppress these warnings is C++17-only, and I think the comment option is more readable than the C++11 ```[[gnu::fallthrough]]``` extension.

We could selectively disable that warning flag in our compiler options, but I kind of like it, when it works.  It actually caught a bug or two (albeit not in a codepath we ever use...)

src/fe/fe_l2_lagrange_shape_3D.C
roystgnr: Will do; thanks.

I moved that line to after the case statement, in the process of figuring out how loose I could be with those comments and still have the gcc7 warning engine recognize them.  But "fixing multiple typos" is a much better excuse for the edit than "probing the gcc7 comment parser via trial and error".

include/utils/hashword.h
jwpeterson: > the only standard way to suppress these warnings is C++17-only, and I think the comment option is more readable than the C++11 [[gnu::fallthrough]] extension.

Do you also prefer the syntactic comment over `__attribute__ ((fallthrough));`? I think that should be ignored by other compilers, but if it's not, we could always wrap it in a `libmesh_fallthrough();` macro.

roystgnr: I prefer the comment, but if we can get a ```libmesh_fallthrough();``` macro to work then I'd be happy with that too, if you'd prefer.  ... and it looks like the macro would probably work as suggested; both the GNU ```__attribute__``` and the C++17 attribute are supposed to be followed by a semicolon.  Should I try that?

jwpeterson: > Should I try that?

It would be my preference. Aside from "syntactic comments are weird" issue, we don't know that all compilers will eventually support the same comment regexes as GCC (although I don't see why they wouldn't).

jwpeterson: I'll also write a configure test of C++17 `[[fallthrough]]`. Then the `libmesh_fallthrough();` macro can eventually expand to that on compilers that support the standard attribute specifiers.

roystgnr: Sounds good.  We'll hold off on this until I can finish and test (the gcc7 specific part of) that.

include/base/libmesh_common.h
```diff
@@ -596,6 +596,15 @@ inline Tnew libmesh_cast_int (Told oldvar)
 #define libmesh_final
 #endif
 
+// Define backwards-compatible fallthrough attribute.
+// We should add support for standard C++17 [[fallthrough]]
+// Maybe [[clang::fallthrough]] or [[gcc::fallthough]] too?
+#if defined(__GNUC__) && (__GNUC__ > 6)
+#define libmesh_fallthrough __attribute__((fallthrough))
+#else
+#define libmesh_fallthrough
```
jwpeterson: Two questions: 
1. No parentheses (to match the look of `libmesh_error_msg()` et al)?
1. This will expand to an empty line with a semi-colon on most compilers. Won't that also trigger a warning? Should we use some non-statement like:
       
       #define libmesh_assert_msg(asserted, msg)  ((void) 0)

   instead?

roystgnr: I felt like the lack of parentheses gave it more of a "fake keyword" feel rather than "fake function", but I can switch it.

You're definitely right about the ```((void) 0)``` suggestion; I'll fix that too.

PR #1424, opened 2017-09-12
stitch_meshes() optimizing

roystgnr:
I got tired of dbg-mode timeouts from the MOOSE mesh/patterned_mesh.patterned_generation test killing my test runs.

This PR:

Adds some logging in stitch_meshes(), in the MeshTools::libmesh_assert_foo() methods, and in a couple heavyweight libMesh methods.

Switches to unordered_map for a speedup in UnstructuredMesh::copy_nodes_and_elements()

Switches to more direct BoundaryInfo copying to speed up stitch_meshes()

Collects some MPI communications to *greatly* speed up parallel libmesh_assert_valid_boundary_ids() tests.  This actually might be a noticeable speedup for parallel dbg runs across the board.

### DIFF
--- a/src/mesh/mesh_base.C
+++ b/src/mesh/mesh_base.C
@@ -172,6 +172,8 @@ void MeshBase::set_spatial_dimension(unsigned char d)
 
 void MeshBase::prepare_for_use (const bool skip_renumber_nodes_and_elements, const bool skip_find_neighbors)
 {
+  LOG_SCOPE("prepare_for_use()", "MeshBase");
+
   parallel_object_only();
 
   libmesh_assert(this->comm().verify(this->is_serial()));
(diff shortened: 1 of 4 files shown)


permcody:
We would have also accepted a new reduced size patterned mesh test. Well, nothing like a thorn in the side to spur you into creating better protective clothing.

roystgnr:
I'm writing up the statement of work for next year's budget from INL.  I'm pretty sure Rich doesn't want me to cross out "libMesh and MOOSE optimizations" and write in "make MOOSE solve smaller problems" instead.  :-D

permcody:
Oh yeah, don't do that!

dknez:
Thanks for this optimization!

Another optimization I'd like to do at some point would be to use nanoflann to find nearest nodes on the "other" mesh to stitch. This should give a decent speedup compared to the N^2 search. I normally don't use the "binary search" since it doesn't work as reliably as the N^2 search, at least in cases that I've tested.

roystgnr:
Huh... do we need nanoflann for that?  Wouldn't the PointLocatorTree give us the same transition to O(N log N)?

dknez:
I guess it would, yes. Though I've found nanoflann to be faster and more lightweight for cases like this where all we care about is a point cloud, and element connectivity is not needed. It's already bundled with libMesh so it'd be nice to use in this case, IMO.

PR #1425, opened 2017-09-12
Add MPI-less allgather(vector<string>)

roystgnr:
I forgot to add this as part of #1415

Without it we can't even compile our own unit tests unless MPI is
installed.


jwpeterson:
👍  Manually attached a "No MPI" test in CIVET as well.

roystgnr:
The only failure was a timeout in dbg parallel patterned_mesh.patterned_generation, which means that not only is this PR ready to merge, so is #1424 ...

PR #1426, opened 2017-09-14
Fix typos

jwpeterson:
Inspired by adfd541b9, I found [scspell](https://github.com/myint/scspell) and used it to fix a bunch of typos in our documentation.

### DIFF
--- a/configure
+++ b/configure
@@ -1955,14 +1955,13 @@ Optional Features:
   --enable-legacy-using-namespace
                           add "using namespace libMesh" to libMesh headers
   --enable-everything     enable all non-conflicting options
-  --enable-unique-id      build with unique id suppport
+  --enable-unique-id      build with unique id support
   --enable-tracefiles     write stack trace files on unexpected errors
-  --disable-amr           build without adaptive mesh refinement (AMR)
-                          suppport
-  --disable-vsmoother     build without variational smoother suppport
-  --disable-periodic      build without periodic boundary condition suppport
+  --disable-amr           build without adaptive mesh refinement (AMR) support
+  --disable-vsmoother     build without variational smoother support
+  --disable-periodic      build without periodic boundary condition support
   --disable-dirichlet     build without Dirichlet boundary constraint support
-  --enable-nodeconstraint build with node constraints suppport
+  --enable-nodeconstraint build with node constraints support
   --enable-parmesh        Use distributed ParallelMesh as Mesh
   --disable-ghosted       Use dense instead of sparse/ghosted local vectors
   --disable-node-valence  Do not compute and store node valence values
@@ -1986,7 +1985,7 @@ Optional Features:
   --disable-hinnant-unique-ptr
                           build without Howard Hinnant's unique_ptr
                           implementation
-  --disable-petsc         build without PETSc iterative solver suppport
+  --disable-petsc         build without PETSc iterative solver support
   --enable-petsc-required Error if PETSc is not detected by configure
   --disable-slepc         build without SLEPc eigen solver support
   --disable-trilinos      build without Trilinos support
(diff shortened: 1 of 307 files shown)


configure
```diff
@@ -1955,14 +1955,13 @@ Optional Features:
   --enable-legacy-using-namespace
                           add "using namespace libMesh" to libMesh headers
   --enable-everything     enable all non-conflicting options
-  --enable-unique-id      build with unique id suppport
+  --enable-unique-id      build with unique id support
```
roystgnr: Ha!  We've been copying and pasting that triple p around for a decade, haven't we?

configure
```diff
@@ -2028,7 +2027,7 @@ Optional Features:
   --enable-fparser-debugging
                           Build fparser with bytecode debugging functions
   --disable-cppunit       Build without cppunit C++ unit testing support
-  --disable-nanoflann     build without nanoflann KD-tree suppport
+  --disable-nanoflann     build without nanoflann KD-tree support
```
roystgnr: So how many false positives did you end up with from scspell?  I can't imagine it has "nanoflann", "fparser", and "cppunit" in its dictionary.

include/enums/enum_eigen_solver_type.h
```diff
@@ -71,8 +71,8 @@ enum PositionOfSpectrum {LARGEST_MAGNITUDE=0,
                          LARGEST_IMAGINARY,
                          SMALLEST_IMAGINARY,
                          TARGET_IMAGINARY,
-
-                         INVALID_Postion_of_Spectrum};
+                         INVALID_Postion_of_Spectrum,
+                         INVALID_POSITION_OF_SPECTRUM};
```
roystgnr: Ok, this worries me.  Are there corresponding changes in the code to accept the new enum option and to spit out a deprecated warning if we see the old option?  Ctrl-F isn't finding them for me, even after I expand all the foo_eigen_bar.C diffs.

src/mesh/mesh_communication_global_indices.C
```diff
@@ -309,7 +309,7 @@ void MeshCommunication::assign_global_indices (MeshBase & mesh) const
 
     communicator.allgather (my_max, /* identical_buffer_sizes = */ true);
 
-    // Be cereful here.  The *_upper_bounds will be used to find the processor
```
roystgnr: The irony is delicious.

roystgnr:
:+1: after my questions about the enum are resolved.  This is great!

Did you script up any kind of ```git blame | cut | sort | uniq -c``` on the diff, by any chance?  I recognize an embarrassing number of those typos as my own, but I'm not sure if that's availability bias or if I'd really win the Most Illiterate LibMesh Developer booby prize.  I'm not going to check myself, though, because the only way the results could come out are "mean" or "sad"...

include/enums/enum_eigen_solver_type.h
jwpeterson: The old enum was never used anywhere, so I didn't really know what else to do to support it?

configure
jwpeterson: There were definitely *lots* of false positives. But scspell makes it pretty painless to update ~/.scspell/dictionary.txt, which is a human-readable text file with all your local words. So for common words like "libmesh" etc, you only have to add them to the dictionary once. Most of the false positives were triggered by our variable names in code, and those are much harder to ignore.

jwpeterson:
> Did you script up any kind of git blame | cut | sort | uniq -c on the diff, by any chance?

I did not. `git blame` isn't very useful in this regard, as it only tells you the *last* person to touch a line... `git log --reverse -Ssuppport` shows that this typo dates back to d7a66045d, i.e. the "Initial import of sources" commit from 2003.

include/enums/enum_eigen_solver_type.h
roystgnr: Oh, was this just one of those habits we get into where whenever we make an enum we give it an "INVALID_WHATEVER" option too?  In that case we're good to go.

PR #1427, opened 2017-09-14
Fix for Exodus reader when infinite elems are present.

jwpeterson:
This is a real fix for #1395. The previous fix, in 79ca57d0496b, did
not actually do anything. The problem was we were telling Exodus there
were e.g. two element blocks and then only writing 1, since we skipped
the block with infinite elems. This caused invalid reads during the
subsequent call to exII::ex_put_concat_all_blocks().

I'm not sure why, but misc_ex1 still passed reliably on Linux with
this bug in place, which is probably why the failure never popped up
during infinite element testing on CIVET. It failed reliably for me on
OSX before this patch, however. In the course of fixing this, I also
switched to calling exII::ex_put_concat_elem_block() since the
resulting code is simpler.

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -1166,8 +1166,16 @@ void ExodusII_IO_Helper::initialize(std::string str_title, const MeshBase & mesh
   for (; it!=end; ++it)
     {
       const Elem * elem = *it;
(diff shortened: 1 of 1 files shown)


jwpeterson:
This should also fix any remaining issues @BalticPinguin was having...

Issue #1418, opened 2017-09-06
Support 2^32 nodes in XdrIO


roystgnr:
The fixes @permcody added years ago get us support for 64-bit *dof* indexing in that format, but we still have a 32-bit n_elem and 32-bit n_node written to the Xdr header.  This will be an easy but also a backwards-compatibility-breaking fix.

roystgnr:
No, scratch that - a forwards-compatibility-breaking fix.  We'll do the usual "increment the version string and handle older strings" and handle backwards compatibility just fine.

roystgnr:
And now we might as well combine this with the fixes to https://github.com/idaholab/moose/pull/9829#issuecomment-328913288 and take care of it soon, because *actually* writing out 8-byte integers on Linux is a pretty critical feature and it turns out we're going to have to break backwards compatibility to do so.

roystgnr:
There doesn't appear to be any way to perfectly recover files from the "8-byte integers from Linux are screwed up" era, either.  The best we can do is going to be "old files with 8-byte integers written by Mac will read properly on Mac", and "old files with 8-byte integers under 2^32 written by Linux will read properly on Linux", because we *thought* this was a cross-platform write call and so there's nothing in the files distinguishing what platform we were writing from, only what byte sizes we *thought* we were writing.

roystgnr:
The fix to the fix seems to have fixed this.

PR #1428, opened 2017-09-14
XDR_IO 1.3.0 format

roystgnr:
This should fix the xdr_long bug (which was breaking binary integer output above 2^32 on at least some LP64 systems like Linux, and was also breaking binary file compatibility between those systems and LLP64 systems Mac OSX), and while we're at it upgrade our "header integer" size to 64 bits too (which was previously blocking binary output for problems with more than 2^32 nodes).

This is going to need some manual testing before we merge.

### DIFF
--- a/include/mesh/xdr_io.h
+++ b/include/mesh/xdr_io.h
@@ -164,20 +164,39 @@ public:
   /**
    * \returns \p true if the current file has an XDR/XDA version that
    * matches or exceeds 0.9.2.
+   *
+   * As of this version we encode integer field widths, nodesets,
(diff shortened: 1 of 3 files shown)


roystgnr:
In theory this should fix all the issues discussed in #1418.
...
Except for the ones that were effectively unfixable - if you have an old 64-bit .xdr file that was written on one of OSX or Linux, you still can't read it on the other system type; you'll need to translate it to .xda or to a 1.3.0 binary file first.

roystgnr:
Gah, I accidentally left in a cherry-picked commit that I'd been experimenting with again.  Let me rebase and take that out...

PR #1429, opened 2017-09-14
Add configure test for C++17 [[fallthrough]] attribute.

jwpeterson:
This goes along with #1423.


jwpeterson:
@roystgnr if you have a chance, give this branch a try with your GCC7 installation? I'm building my own here, but it looks like it won't be done by the time I leave for the day...

roystgnr:
Huh; doesn't work for me, not with "-std=c++1z" on GCC 7.2.

roystgnr:
```
configure:11492: g++ -c -std=c++1z -Werror -Wall -Wextra -pedantic  conftest.cpp >&5
conftest.cpp: In function 'int main()':
conftest.cpp:51:20: error: variable 'j' set but not used [-Werror=unused-but-set-variable]
           int i=1, j;
                    ^
```

roystgnr:
Sticking a ```(void)j; // Avoid unused variable warning``` in the AC_COMPILE_IFELSE fixes it for me, and now we detect fallthrough attribute support with C++1z.

roystgnr:
And it looks like GCC 7 is respecting the C++17 fallthrough attribute, too.

jwpeterson:
Thanks for checking.

Also, very interesting that you can force-push to branches in my fork? I don't remember setting up write permissions for you in my fork, but it's possible that I did...

jwpeterson:
BTW, there are 2 test programs in the m4 file, so you will probably want to do your little trick (or just initialize j to something) in both of them.

roystgnr:
There's a "let developers push to this branch" checkbox somewhere when you submit a PR, which IIRC is on by default, but which feature I don't recall any of us ever *making use of* before.  But I figured it'd be safest for me to add the new commits since I was set up to test them, except creating a new branch and a new PR and canceling this PR as per our usual workflow sounded like a huge hassle for one tiny change and a different bootstrap, so I figured I'd see whether that permission checkbox actually works the way it claims...

jwpeterson:
Aha, "allow edits from maintainers". Good to know.

Issue #1430, opened 2017-09-15
"Error writing nodeset parameters" with HEAD version


dknez:
I just updated to fc1536d4200a43dfa56270caeb107619e2827cea and now I hit the "Error writing nodeset parameters" error in ExodusII_IO_Helper for one of my test models. I'll look into this in more detail, but also wanted to check if anyone has any thoughts on what may have happened?

jwpeterson:
I merged 7e83cfcdc yesterday but it passed all the MOOSE testing and should really have only affected infinite elements builds...

dknez:
Running in dbg mode gives this error message:

Exodus Library Error: [ex_put_sets]
	Error: failed to define number of entries in node set 130004 in file id 65536
	[-41] netcdf constraint NC_MAX_DIMS exceeded
Error writing nodeset parameters

jwpeterson:
OK... and if you revert 7e83cfcdc it goes away? I really don't see how my change could have broken nodeset writing. 

Has it been a while since you updated? We also upgraded the version of NetCDF that we bundle with libmesh somewhat recently (April 2017). If so, maybe try a clean build.

dknez:
I updated from 96532f4d070713c5c452fe150a589c0af36d9554 from August 23, so it hasn't been very long. I'll try reverting 7e83cfc and I'll also try a clean build and see what happens.

Also, I see there is some discussion about this issue, and it's relation to pre-installed packages:
https://github.com/Unidata/netcdf-c/issues/148

dknez:
Reverting 7e83cfc didn't help, and neither did doing a clean build... I'll have to look further. I guess it's something to do with the NetCDF version I'm linking to, or something. Not sure what changed for me to trigger this issue, though.

jwpeterson:
Yeah if I recall correctly, Ben had to change the default values of some constants in NetCDF when he vendored in the new version. It's likely that you can't use a vanilla NetCDF with libmesh because of this...

dknez:
Well I still haven't figured out what's going wrong. As far as I can tell based on config log I'm using contrib/netcdf, and I reverted to the libMesh version that I was using before, and I still get the same error.

It's almost certainly a problem with my configuration somehow, not a libMesh issue, so I will close this issue. But @jwpeterson if you do have any other suggestions on what I could try in order to get to the bottom of this, let me know :)

jwpeterson:
[...]

PR #1432, opened 2017-09-18
libmesh-config: Move libmesh library PATH to beginning of --libs.

jwpeterson:
This should hopefully fix the issue reported by @dknez in #1430 where
he was getting a libnetcdf.so from /usr/lib in his linker line before
the one in $LIBMESH_INSTALLED_DIR/lib.


dknez:
This works for me, thanks :thumbsup:

PR #1433, opened 2017-09-19
Small update to nodal assembly in reduced basis code.

dknez:
Allow nodal assembly contributions to the matrix as well as the rhs.

### DIFF
--- a/include/systems/elem_assembly.h
+++ b/include/systems/elem_assembly.h
@@ -19,6 +19,7 @@
 #define LIBMESH_ELEM_ASSEMBLY_H
 
 #include "libmesh/reference_counted_object.h"
+#include "libmesh/dense_matrix.h"
 
 namespace libMesh
 {
@@ -42,7 +43,10 @@ public:
    * Constructor.  Initializes required
    * data structures.
    */
-  ElemAssembly () {}
+  ElemAssembly ()
+  :
+  is_nodal_rhs_values_overriden(true)
+  {}
 
   /**
    * Destructor.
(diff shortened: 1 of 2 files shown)


include/systems/elem_assembly.h
```diff
@@ -60,16 +61,17 @@ class ElemAssembly : public ReferenceCountedObject<ElemAssembly>
...
+  get_nodal_values(std::vector<dof_id_type> ,
+                   DenseMatrix<Number>& ,
+                   DenseVector<Number>& ,
+                   const System & ,
+                   const Node &)
   {
-    // By default, just clear the values map
-    values.clear();
+    // Do nothing by default
   }
```
roystgnr: This is API-breaking for a class that's intended to be subclassed, isn't it?  Any way to keep some (libmesh_deprecated()) backwards compatibility here?

dknez:
hmm, yes, you're right. Though this feature has never been used before by anyone (as far as I know) including me, so I figured it'd be OK to be non-backwards compatible in this case.

Nevertheless, if you prefer that I add the old version back in, I can do that.

dknez:
There's an error here anyway, according to the tests... I'll have a look.

dknez:
The bug should be fixed now, and I also added the old get_nodal_values method back in, with libmesh_deprecated in it. @roystgnr let me know if you think that's better now?

roystgnr:
No, actually I think listening to me probably made things worse, sorry.  :-P

The trouble here is not that we might have old code *calling* get_nodal_rhs_values() and not finding it in a new library build.  The problem is that we might have old code *providing* get_nodal_rhs_values() and not having that respected in a new library build.  Instead the default implementation of get_nodal_values() would be called, which would be 0, which would give an erroneous solution, which is the worst type of error possible.

Could we get rid of get_nodal_rhs_values() after all, but also just make get_nodal_values() pure virtual?  Then if there are any users who subclassed the old design, they'll get a compile-time error with an obvious fix, rather than a run-time incorrect result which might not even be obvious to detect.

dknez:
Yes, your concern is right. However, I think making get_nodal_values() pure virtual would be a bit of a pain since we subclass ElemAssembly in lots of places (e.g. in all of the reduced_basis examples, and essentially all reduced basis app codes), and we'd have to add an empty get_nodal_values() implementation in all of those cases. Those are just compiler errors so easy to fix, but it would be a pain...

A workaround would be to keep the get_nodal_rhs_values call inside RBConstruction as it was before, and to eventually deprecate that along with ElemAssembly::get_nodal_rhs_values after some reasonable amount of time. What do you think of that approach?

roystgnr:
Hmm... could you also have the new default get_nodal_rhs_values set some _i_wasnt_overridden flag, so if you call it and that flag *isn't* set then you know you're dealing with old code and you call libmesh_deprecated()?  Then we get a release with *that* in it out, and we can strip the whole thing out right afterward.

dknez:
@roystgnr let me know what you think about the updated version? We only call get_nodal_rhs_values in RBConstruction if there is a non-default implementation of that method, and we call libmesh_deprecated in that case to tell users to update to the new version. Similarly, the base version of get_nodal_rhs_values also calls libmesh_deprecated. That looks OK to me, but let me know what you think.

Thanks!

dknez:
I'm going to go ahead and merge since we went through several iterations, and this seems OK to me now.

PR #1441, opened 2017-09-25
Added a API flush in sparse matrix

fdkong:
It is useful when we switch between  `ADD_VALUES` and `INSERT_VALUES` for a petsc matrix.
`Flush` won't shrink the memory. Users still have to call `close` before doing any actual calculation.

### DIFF
--- a/include/numerics/sparse_matrix.h
+++ b/include/numerics/sparse_matrix.h
@@ -171,6 +171,13 @@ public:
   virtual void close () = 0;
 
   /**
+   *  For PETSc matrix , this function is similar to close but without shrinking memory.
+   *  This is useful when we want to switch between ADD_VALUES and INSERT_VALUES.
+   *  close should be called before using the matrix.
+   */
+  virtual void flush () { close(); }
+
+  /**
    * \returns The row-dimension of the matrix.
    */
   virtual numeric_index_type m () const = 0;
--- a/src/numerics/petsc_matrix.C
+++ b/src/numerics/petsc_matrix.C
@@ -913,6 +913,19 @@ void PetscMatrix<T>::close ()
   LIBMESH_CHKERR(ierr);
 }
 
+template <typename T>
+void PetscMatrix<T>::flush ()
+{
+  semiparallel_only();
+
+  PetscErrorCode ierr=0;
+
+  ierr = MatAssemblyBegin (_mat, MAT_FLUSH_ASSEMBLY);
+  LIBMESH_CHKERR(ierr);
+  ierr = MatAssemblyEnd   (_mat, MAT_FLUSH_ASSEMBLY);
+  LIBMESH_CHKERR(ierr);
+}
+
 
 
 template <typename T>
--- a/include/numerics/petsc_matrix.h
+++ b/include/numerics/petsc_matrix.h
@@ -150,6 +150,8 @@ public:
 
   virtual void close () libmesh_override;
 
+  virtual void flush () libmesh_override;
+
   virtual numeric_index_type m () const libmesh_override;
 
   virtual numeric_index_type n () const libmesh_override;


include/numerics/sparse_matrix.h
```diff
@@ -171,6 +171,13 @@ class SparseMatrix : public ReferenceCountedObject<SparseMatrix<T> >,
   virtual void close () = 0;
 
   /**
+   *  For PETSc matrix , this function is similar to close but without shrinking memory.
+   *  This is useful when we want to switch between ADD_VALUES and INSERT_VALUES.
+   *  close should be called before using the matrix.
+   */
+  virtual void flush () = 0;
```
jwpeterson: Let's make the default implementation call `close()` instead of being pure virtual. Then you can drop the changes from the Trilinos and Eigen specializations. Also, this PR misses the Laspack implementation (because it's not compiled) but adding a default base class behavior fixes this as well.

src/numerics/petsc_matrix.C
```diff
@@ -913,6 +913,19 @@ void PetscMatrix<T>::close ()
...
 }
 
+template <typename T>
+void PetscMatrix<T>::flush ()
+{
+  semiparallel_only();
+
+  PetscErrorCode ierr=0;
+
+  ierr = MatAssemblyBegin (_mat, MAT_FLUSH_ASSEMBLY);
```
jwpeterson: Looks like this dates back to at least 1996 petsc/petsc@6d4a8577250, so we should be good introducing this without version guards.

jwpeterson:
@fdkong I sent you an invite to join the group of people who get their libmesh PRs tested automatically.

src/numerics/petsc_matrix.C
fdkong: I was in primary school at that time.

permcody: lol

include/numerics/sparse_matrix.h
```diff
@@ -171,6 +171,13 @@ class SparseMatrix : public ReferenceCountedObject<SparseMatrix<T> >,
   virtual void close () = 0;
 
   /**
+   *  For PETSc matrix , this function is similar to close but without shrinking memory.
+   *  This is useful when we want to switch between ADD_VALUES and INSERT_VALUES.
+   *  close should be called before using the matrix.
+   */
+  void flush () { close(); }
```
jwpeterson: This needs to be marked `virtual`

roystgnr: Huh.  How did the libmesh_override in the subclass not catch that?

roystgnr: Oh, it *did* catch that, just not in the No-C++11 build.

src/numerics/petsc_matrix.C
pbauman: Did... did @fdkong just call us old? I think he just called us old.

roystgnr: You need him to speak up, @pbauman?  Sad, but they do say the hearing is the first thing to go...

roystgnr: I should probably replace my gravatar one of these days... the baby in that picture starts after-school robotics club next week.

roystgnr:
Looks good, thanks.  Maybe add a unit test in future PRs, though?  I want to say "this is so simple that it couldn't possibly need special testing" but it feels like every time I say that it turns around and bites me...

src/numerics/petsc_matrix.C
jwpeterson: > Did... did @fdkong just call us old? I think he just called us old.

He's not wrong.

pbauman: Whatever, speak for yourself. :P Excuse me while I go get those kids off my lawn...

Issue #1442, opened 2017-10-01
Poor Memory Usage Patterns During Projection


friedmud:
During some adaptivity runs that I'm doing... I'm seeing a big stall just after the mesh gets adapted.  I ran a quick profiler on it and found that it's spending at least 20% (actually more like 30% once I collapsed some of the small stuff) of it's time just allocating and deleting an FEBase object in FEMContext::point_value() (see stack below).

Does anyone see a good way to fix this up?

<img width="1126" alt="projection_time_stack" src="https://user-images.githubusercontent.com/870705/31059594-d5cfccf8-a6d2-11e7-8190-f8a5b7a620db.png">

roystgnr:
On Sun, 1 Oct 2017, Derek Gaston wrote:

> During some adaptivity runs that I'm doing... I'm seeing a big stall
> just after the mesh gets adapted. I ran a quick profiler on it and
> found that it's spending at least 20% (actually more like 30% once I
> collapsed some of the small stuff) of it's time just allocating and
> deleting an FEBase object in FEMContext::point_value() (see stack
> below).
> 
> Does anyone see a good way to fix this up?

Not immediately, but that's a private method with just the one use
case, so there ought to be a ton of room to refactor it for
performance.

I don't think I can safely reuse the existing FEBase here, since
FEMContext users might reasonably want to mix point_value() calls into
standard quadrature operations, but I can surely save any
point_value-generated objects and reuse them for successive operations
with the same OutputShape.  Then the only memory operations left are
the internal resize() calls, which ought not to touch malloc at all when
they're not changing the size.

Could you set up a similar test case (with just libMesh or at most
MOOSE/framework) I can use for benchmarking?  If not then I'll just
crank up the mesh size on one of our adaptivity or adjoints AMR
examples.

Thanks,
---
Roy

PR #1443, opened 2017-10-02
Cache FE in FEMContext::point_value()

roystgnr:
@friedmud, would you give this a try?  It's a minor design tweak rather than a good permanent solution but I'm hoping it'll be noticeably cheaper for you than the current code.

### DIFF
--- a/include/fe/fe_abstract.h
+++ b/include/fe/fe_abstract.h
@@ -218,6 +218,12 @@ public:
 #endif // LIBMESH_ENABLE_PERIODIC
 
   /**
+   * \returns the dimension of this FE
+   */
+  unsigned int get_dim() const
+  { return dim; }
+
+  /**
    * \returns The \p xyz spatial locations of the quadrature
    * points on the element.
    */
--- a/include/systems/fem_context.h
+++ b/include/systems/fem_context.h
@@ -995,9 +995,21 @@ protected:
    * Helper function to reduce some code duplication in the *_point_* methods.
    */
   template<typename OutputShape>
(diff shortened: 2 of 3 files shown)


roystgnr:
Refs #1442

friedmud:
Will do

roystgnr:
@jessecarterMOOSE reports a significant benefit:

> for one particular case in 2D using 8 cpus, after applying the patch, the time spent in reinit() after the patch was about 20%-25% less! (depending on the initial mesh size)

And IMHO even if he's on a corner case, that's worth an extra 80 lines of code complication, so unless someone screams I'll probably merge this to master this weekend.

jessecarterMOOSE:
I'd like to test if that figure applies more generally for other cases, but I'm pleased with the speedups I've seen so far.

However, does this fix @friedmud 's original issue regarding memory usage? 

Or maybe the problems are one in the same. You know what they say - "time is memory", or something like that.

roystgnr:
> However, does this fix @friedmud 's original issue regarding memory usage?

It's at least a step in the right direction, unless I misunderstood: the issue wasn't memory usage, it was memory usage *patterns*.  We weren't leaking memory, or using too much memory; what we were doing was hitting the heap memory allocator and deallocator far too frequently, thereby killing CPU performance.

roystgnr:
I'm going to wait until after #1459 goes in before merging, though: I don't *expect* any conflicts between it and the other ready PRs, but a critical-feature addition is more urgent than performance fixes or convenience-feature additions, so best to get that one in first.

jessecarterMOOSE:
Makes sense to me :+1:

PR #1444, opened 2017-10-04
Replace Side -> Edge in Elem::edge_nodes_map documentation.

jwpeterson:
This is a documentation-only change but PR because [hacktoberfest T-shirt](https://hacktoberfest.digitalocean.com/).


roystgnr:
I was going to make a crack about "only 4 PRs in a whole month!?" but looking through my own history I see a two week gap in June with nothing merged.  That was a week of meetings followed immediately by a week of vacation, but still, you'd think I could have gotten *something* accomplished...

PR #1445, opened 2017-10-04
Update DifferentiableQoI to handle more complex derivatives

tradowsk:
This adds `DifferentiableQoI::finalize_derivative()` which is analogous to `DifferentiableQoI::parallel_op()`. I'm working with an exponential QoI function in GRINS for AMR (see [here](https://github.com/grinsfem/grins/blob/master/src/qoi/include/grins/absorption_coeff.h#L51)). Because of the exponential, in addition to summing the `element_qoi_derivatives`, I also need to apply the chain rule to the derivative values before they are fed to the adjoint solver.

By default, this function does nothing so there are no backward compatibility issues. If you have a better name, I'd be happy to change it.

### DIFF
--- a/include/physics/diff_qoi.h
+++ b/include/physics/diff_qoi.h
@@ -23,6 +23,7 @@
 // Local Includes
 #include "libmesh/auto_ptr.h"
 #include "libmesh/parallel.h"
+#include "libmesh/diff_context.h"
 
 // C++ includes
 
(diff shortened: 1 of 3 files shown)


roystgnr:
Seems reasonable.  Originally we'd planned to just override FEMSystem::assemble_qoi_derivative() (calling the base class function to do some of the grunt work) in situations like these, but I suppose that's not really a good option in the newly preferred idiom of swappable/composable/etc diff_qoi functors.

tradowsk:
Originally I did override `assemble_qoi_derivative` but I quickly found out that I basically just copied the code with one extra step at the end, so this seemed like a cleaner solution (at least for my purposes)

PR #1446, opened 2017-10-05
Refinement early return

roystgnr:
Add a test to MeshRefinement::refine_and_coarsen_elements() to make sure we actually have something to refine or coarsen; if not then we can return early without first testing for level mismatches and testing for unrefined islands and so forth.

Also remove the probably-redundant call to make_flags_parallel_compatible(), except as a dbg-mode test; if you need it then you can call it manually, and if you don't then we shouldn't be wasting your time with it.

See https://github.com/idaholab/moose/pull/10025 for discussion.

### DIFF
--- a/src/error_estimation/hp_coarsentest.C
+++ b/src/error_estimation/hp_coarsentest.C
@@ -31,6 +31,7 @@
 #include "libmesh/elem.h"
 #include "libmesh/error_vector.h"
 #include "libmesh/mesh_base.h"
+#include "libmesh/mesh_refinement.h"
 #include "libmesh/quadrature.h"
 #include "libmesh/system.h"
 #include "libmesh/tensor_value.h"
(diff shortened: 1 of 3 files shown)


jessecarterMOOSE:
This speeds things up quite well. Thanks!

jessecarterMOOSE:
I went ahead and closed idaholab/moose#10025 in favor of this PR. Your way is much cleaner.

roystgnr:
Oh, of course I'm an idiot.

Even if we're not h-refining, we can't exit early unless we know we're not p-refining either.

We also need one sync between processors to handle distributed mesh cases where one processor sees refinement flags but another does not.

jessecarterMOOSE:
Ah, that didn't register because I'm not using p-adaptivity nor distributed mesh for my problem.

roystgnr:
Okay, I've got fixes for every problem I can find (none of which ought to noticeably affect performance), but I'm going to do a bunch of rebasing and force push so that we don't end up with any intermediate broken commits.

roystgnr:
Random side note: I honestly suspect we're going to see a small but noticeable speedup after moving to C++11 ranged-for loops.  There are a ton of places where we put a virtual function call into a loop stopping condition, not because we expect "n_children()" or"n_neighbors()" to change from iteration to iteration, but because that's just the most terse way to write those loops in C++98...

jessecarterMOOSE:
👍 thanks!

PR #1447, opened 2017-10-09
Require (and use) C++11

roystgnr:
The first two commits in this PR make C++11 support a mandatory, rather than an optional, libMesh dependency at configure time.  I no longer have any users or users-of-middleware-users demanding C++98 support, and I think I was the last major libMesh developer in that predicament, so IMHO it's time to bite the bullet and let our developers write C++11 code whenever they please.

As a first internal-C++11-usage test case, I then decided to add a couple Elem methods returning iterator ranges (one iterating over pointers-to-neighbors, the other over references-to-children), and then use one of those methods (```child_ref_range()```) within the library for range-based for loops replacing the previous C-style for loops.  In theory iterating over these ranges ought to be *faster* than the old code, since we only make a single virtual function call to ```n_children()``` rather than making a call at each iteration of the for loop.

In places where we can't use the new ranges, because we need indices for e.g. ```is_child_on_side()``` calls, I now cache the output of ```n_children()``` manually.  In theory we could have been doing that from the outset, but it takes extra code (especially compared to the range loop, which also hides the dereferencing operation).  Hopefully now that the fastest loop is also the most terse we'll use it by default in new code.

This is obviously a big change, so let's not merge until everyone's had a chance to have their say.  Even if everyone agrees on the substance of the change, now is the time to decide on the naming convention for ```child_ref_range()``` and ```neighbor_ptr_range()``` and such, before they ever end up in user code.

Right now ```neighbor_ptr_range()``` is unused, because this PR was already 29 commits long (I kept all the n_children() changes split into commits by file, to make bisections easier if I screwed any up).  We can strip it out of this PR and into another PR with use cases added, if anyone wants, or I can tack the use cases onto the end of this PR if nobody minds the end result being 60 commits long.


jwpeterson:
I'll put in a CIVET PR to drop the "No C++11" integration testing (although it looks like that now "worked" since it failed correctly). Haven't looked at why everything else failed.

IMO there don't need to be 17 separate "Avoid n_childeren() ..." commits but I leave that to your discretion.

jwpeterson:
Did you maybe add a new header?
```
No rule to make target `../../../include/utils/intrange.h', needed by `intrange.h'.  Stop.
```
There are some [convoluted instructions](https://github.com/libMesh/libmesh/wiki/Adding-or-removing-source-files) you have to follow to add files...

jwpeterson:
This PR introduces the use of range-based for-loops, so that configure test at least (`LIBMESH_TEST_CXX11_RANGEFOR`) needs to become a hard fail.

jwpeterson:
Can we choose a better name for `class PPIter`? How about `PointerToPointerIter` or something more descriptive?

roystgnr:
Ah, hell.  I added the intrange.h header for another use case, changed my mind and stripped it out, but must have forgotten to strip out the build instructions for it too.  I'll fix that now.

roystgnr:
Yeah, I don't like PPIter, I just couldn't think of anything better.  I don't like PointerToPointerIter either, but I guess "too verbose" is much better than "not descriptive enough".  I'll change it.

jwpeterson:
This isn't super relevant, but since I noticed it in the diffs, I see a bunch of 
```
if (child == remote_elem)
``` 
tests. Could/should those be changed to `child->is_remote()` calls instead? We also have some where we go through the gymnastics of taking the address to do the pointer comparison...
```
if (&mychild == remote_elem)
```

include/geom/elem.h
```diff
@@ -1573,6 +1609,45 @@ class Elem : public ReferenceCountedObject<Elem>,
...
+public:
+  ConstChildRefIter (const Elem * const * childpp) : PPIter<const Elem>(childpp) {}
+};
+
+
+inline
+SimpleRange<Elem::ChildRefIter> Elem::child_ref_range()
+{
+  libmesh_assert(_children);
+  return {_children, _children + this->n_children()};
```
jwpeterson: I forget what construction syntax is called, maybe "uniform initilization"? 

I don't think we explicitly test for this in *any* of the configure tests, although I sure hope we don't come across a compiler that doesn't support this yet supports everything else we do test for....

roystgnr:
I'd actually expect ```if (child == remote_elem)```, with or without the ```&```, to be more efficient.  ```is_remote()``` is a virtual function call, which means a vtable fetch and indexing and jump and stack push and jump back and stack pop, whereas the pointer comparison (which is equivalent for the foreseeable future, since the remote_elem singleton is still the only remote Elem object to ever be instantiated) is just a fetch and compare.

Although, now that I think about it, perhaps I should just make Elem::is_remote() non-virtual and do the address comparison in that?  We'd get the same speed as doing the comparison manually, the code would be more readable, and we'd be able to upgrade the implementation seamlessly if it ever does turn out that the word doing all the work in "for the foreseeable future" was "foreseeable".

I'd like to put that in a separate PR, though.  There are a million other "if (some_elem == remote_elem)" calls throughout the code.

roystgnr:
I'll put a hard fail on the RANGEFOR test.  I actually considered doing that but I convinced myself that the reason not to was "we're now going to let developers let whatever C++11 they want, so it would be mean to make them turn on a new hard configure test each time they employ another feature".  Upon further introspection, I'm pretty sure my actual reason was "I'm lazy and I hate touching autoconf stuff".

include/utils/ppiter.h
```diff
@@ -0,0 +1,68 @@
...
+
+template <typename T>
+class PPIter
+{
+public:
+  PPIter (T * const * it) : _it(it) {}
+
+  T & operator* () const { return **_it; }
+
+  const PPIter & operator++ () {
```
jwpeterson: Formatting: opening curly brace on separate line.

include/geom/elem.h
roystgnr: Uniform initialization.  I didn't want to put any explicit intermediaries here because I was wary that doing so incorrectly might force an actual temporary construction, make return value elision harder, or in some other way pessimize the code.

You may be right to be concerned, though?  [It looks like](http://en.cppreference.com/w/cpp/compiler_support) initializer lists were implemented *after* ranged for loops by some compilers; in particular users of clang 3.0, MSVC++ 17.0, and possibly Intel 13.0 could be bitten at compile time rather than configure time.

include/utils/ppiter.h
roystgnr: Will fix.

Would you mind putting comments in the catch-all thread rather than inline with code?  IIRC Github can drop the latter into the ether after a force push changes commit hashes, and I'm going to need to force push at least the intrange.h fix eventually to avoid having broken intermediate commits.

jwpeterson:
👍  Other than the minor stuff I mentioned, this looks good to me.

Once this gets merged, I'll go and tag the last commit on master prior to this merge as "cpp03_final" or something along those lines.

jwpeterson:
BTW, the "No C++11" should now be gone from future libmesh PRs, but it might stick around for this one (since the PR was originally created with it) so just ignore it.

pbauman:
Sorry femputer was down for this, it is back up now. I'll check this PR against GRINS manually today.

roystgnr:
I think we're ready to go, but I'll leave this unmerged for another day or two in case anyone else wants to chime in or in case @jwpeterson wants to double-check my responses to his reviews.

jwpeterson:
Ah, so we did have a test for initializer lists already... and lucky for you, it looks like your PR got tested just before the last of the build boxes went offline for the day. The rest looks OK to me as well.

jwpeterson:
Tagged 613c152 as the official last C++03-compatible revision.

PR #1448, opened 2017-10-09
Catch exception during side assembly in RBConstruction.

dknez:
Exceptions can be thrown in side_fe_reinit, e.g. due to a zero or negative Jacobian on an element side. This is sometimes intentional, e.g. some people use "degenerate HEX" elements when generating meshes. In order to support this case we just print a message and skip the side assembly on this element if an exception occurs during side_fe_reinit.


dknez:
@jwpeterson thanks, I changed it to a reference.

jwpeterson:
FYI our build boxes are going to be offline this week, so you may not get accurate (or any) continuous integration testing for your PRs...

dknez:
OK. I've tested it and it works fine for me, and it's a change that only affects the reduced basis code (also I just updated the PR so that it's a smaller change than before). So I'll just go ahead and merge then, if that's OK with you?

PR #1449, opened 2017-10-10
Adding support for SHELL8 elements

svallaghe:
Similar to SHELL4 added last year, but second order.

### DIFF
--- a/src/geom/face_quad8_shell.C
+++ b/src/geom/face_quad8_shell.C
@@ -1,0 +1,18 @@
+// The libMesh Finite Element Library.
+// Copyright (C) 2002-2017 Benjamin S. Kirk, John W. Peterson, Roy H. Stogner
+
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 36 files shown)


src/geom/face_quad8_shell.C
```diff
@@ -0,0 +1,18 @@
...
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+// The QuadShell8 functions are all declined inline.
```
jwpeterson: typo: "declared inline". Probably this was copy-pasted from somewhere else, might as well fix that too?

svallaghe:
I will try to add an example tomorrow, basically the second order version of miscellaneous_ex12.

PR #1452, opened 2017-10-12
Deprecated code use can optionally be an error, not just a warning

roystgnr:
While working on other matters I noticed a long-deprecated function being called, and wondered just how many of those are still hiding somewhere in our codes.

This option makes it much easier to find out.

```configure --disable-deprecated``` turns off as much support for deprecated code as possible: completely deprecated APIs are ifdef'ed out in that case, and in deprecated code paths the ```libmesh_deprecated()``` calls turn from warnings into errors.

I've bundled in many fixes for just-discovered deprecated code usage within the library.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -4639,6 +4639,7 @@ datarootdir = @datarootdir@
 docdir = @docdir@
 dvidir = @dvidir@
 enabledefaultcommworld = @enabledefaultcommworld@
+enabledeprecated = @enabledeprecated@
 enablelegacyincludepaths = @enablelegacyincludepaths@
 enablepetsc = @enablepetsc@
 enableuniqueptr = @enableuniqueptr@
--- a/configure
+++ b/configure
@@ -855,6 +855,7 @@ LIBMESH_ENABLE_INFINITE_ELEMENTS_FALSE
 LIBMESH_ENABLE_INFINITE_ELEMENTS_TRUE
 enablelegacyincludepaths
 enabledefaultcommworld
+enabledeprecated
 enablewarnings
 enableuniqueptr
 PWD
@@ -1155,6 +1156,7 @@ enable_unordered_containers
 with_gdb_command
 enable_unique_ptr
 enable_warnings
+enable_deprecated
 enable_blocked_storage
 enable_default_comm_world
 enable_legacy_include_paths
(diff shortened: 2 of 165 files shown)


jwpeterson:
I like this idea. If we can get it totally clean, we could also add a CIVET test with this configuration to make sure none sneak back in.

jwpeterson:
I'd like to wait for CIVET on this one too, though, if you don't mind.

roystgnr:
Let's *definitely* wait for Civet.  I'd say let's even add the new Civet --disable-deprecated test first to be safe.  I've fixed everything deprecated that got caught by the compile-time errors, but as we speak I'm fixing an adjoints_ex5 run-time error.

roystgnr:
When are the Civet boxes expected to be back online by?  Monday?

jwpeterson:
> When are the Civet boxes expected to be back online by? Monday?

I don't think we have an ETA other than the outage as a whole is scheduled to end by 10/18.

roystgnr:
Okay, I think that CondensedEigenSystem issue was the last runtime nastiness I could find.

And hey - those are the Civet tests up and running!  We ought to be good to merge this when they're done, then, although if the new Civet test will be ready soon then I'd prefer to wait for that.

jwpeterson:
> if the new Civet test will be ready soon then I'd prefer to wait for that.

Let me see if I can merge PRs in the civet_recipes repo. We are off today so Brian is not around... The new flag is just `--disable-deprecated`, right?

roystgnr:
Right.

jwpeterson:
Did you want the new CI test to build all of MOOSE with --disable-deprecated turned on as well. That will probably uncover several places in MOOSE where we are still using deprecated interfaces.

jwpeterson:
OK, new recipe merged in CIVET and attached to this PR. At the moment it's optional, but if we decide we can like it, we can run it on every PR automatically.

roystgnr:
Okay, good, we're not just passing with --disable-deprecated on my machine.

Whether to make Test No Deprecated permanent/automatic... I'd say not yet, so long as we remember to test it out before releases at least.  I'll also try it out by hand on my PRs from time to time.  Making it mandatory on every PR might be more worthwhile after middleware and user codes have had time to catch up, at which point our advanced users might get into the habit of using that option by default and we'll want to be careful not to regress it even in git master.

Issue #1451, opened 2017-10-12
DofObject copy constructor has bugs, also is misused, also exists.


roystgnr:
Not sure whether this is really zero issues, one, two, or three.

In terms of "proper" C++ semantics we have at least one issue: The DofObject copy constructor doesn't properly handle non-null old_dof_object entries, so a copy is not equivalent to the original, which is philosophically incorrect unless that's fixed.

But it's probably two issues: if we fix the DofObject constructor, then when we construct an old_dof_object by copying ```*this```, the copy will be recursive, which will be a waste of time and memory since we never care about "older-than-old" dof object data.  Correct behavior for non-existent we-want-a-real-copy use cases would just be a performance regression in the one existing use case, so we'd need some kind of copy-but-not-entirely method to use there instead.

And in a sense it's three issues: do we really *want* a DofObject copy constructor to exist, considering that our only use case for it is in some sense an abuse of the term "copy"?

I discovered the above bug by accident while experimenting with new code, by writing ```for (auto node : elem.node_ref_range())``` rather than ```for (auto & node : elem.node_ref_range())``` as I should have... and fortunately this gave me a run-time bug, because we catch run-time bugs.  After fixing the copy constructor, the effect of such a mistake will be a performance regression, and we don't yet have any mechanism to catch any but the most serious performance regressions.  The ```child_ref_range()``` method doesn't have this risk, because there's no way to copy an ```Elem&``` to an abstract base class ```Elem```.

I'm tempted to get rid of (or at least ```libmesh_deprecated()```) the Node copy constructor and make the DofObject copy constructor private (which latter change wouldn't *fix* the bad semantics but would at least limit their scope).  Trying that out, nothing in libMesh appears to break except for the Node copy constructor unit test.  It's a compatibility-breaking API change, of course, but I strongly suspect that anyone *else* using that API is likewise using it inadvertently and paying a performance penalty for doing so.  If my app codes have a ```Node n = *node_ptr[i];``` somewhere then I'd be happy to find it at compile time.

roystgnr:
Let's postpone "get rid of" for a couple releases, and just do ```libmesh_deprecated()``` as per our usual policy.  The only reason I was leaning towards "get rid of" was that it makes inadvertent usage much easier to find, but #1452 accomplishes the same feat without immediately *forcing* inadvertent users to change their code.

PR #1454, opened 2017-10-13
More ranged-for loops, deprecate Node copy constructor

roystgnr:
This PR adds Elem::side/edge/node_index_range() methods, which seem to be the most terse way to efficiently loop over ranges of indices from 0 to Elem::n_foo().  Using these ranges replaces a lot of one-virtual-call-per-iteration loops with one-virtual-call loops.  I think the new loops may be easier for compilers to unroll, too.

It also adds Elem::node_ref_range(), which iterates directly over element nodes the way child_ref_range() does over children, and which likewise is a little more terse (and a little more typo safe now, and possibly a little more efficient) than iterating over and evaluating indices.

Finally, we deprecate the Node copy constructor, which makes node_ref_range() (and plain Node referencing) safer to use by avoiding inadvertent copying, fixing #1451.

As with #1447, this is probably broken into far more commits than necessary, but unfortunately these aren't regexp-automated changes, I have made (and found and fixed and rebased) a couple typos, and in case there are still any subtle bugs left I'd like to be able to bisect them precisely.

### DIFF
--- a/examples/adaptivity/adaptivity_ex1/adaptivity_ex1.C
+++ b/examples/adaptivity/adaptivity_ex1/adaptivity_ex1.C
@@ -304,7 +304,7 @@ void assemble_1D(EquationSystems & es,
       // Loop over the sides of this element. For a 1D element, the "sides"
       // are defined as the nodes on each edge of the element, i.e. 1D elements
       // have 2 sides.
-      for (unsigned int s=0; s<elem->n_sides(); s++)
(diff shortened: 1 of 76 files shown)


roystgnr:
I know this is a big swath of changes, but since it includes a lot of (admittedly individually very minor) performance fixes and since I'm looking at performance right now I'd like to get it merged.  I'll probably do so this weekend unless anybody screams first.

PR #1455, opened 2017-10-15
Redo naming scheme for Checkpoint IO

friedmud:
Ok - the "new" naming scheme for parallel checkpoint IO wasn't working for me.  I'm putting this up as an idea (that I'm currently using) to start a discussion.

The thing is: I want to be able to say open `blah.cpr` and I want the framework to figure out that I'm running on `x` processors and automatically find the pre-split set of files that are right for that number of processors.

With the way it is currently implemented when you run `splitter` it generates files like this:

```
split.4.cpr
split.4.cpr-4-0
split.4.cpr-4-1
split.4.cpr-4-2
split.4.cpr-4-3
```

This means that if I want to now run on 6 processors I have to actually change my input file / code to read `split.6.cpr` instead of `split.4.cpr`.  Think of a scaling study where I want to run hundreds of different processor combinations.... So: I want to just be able to tell libMesh to open `split.cpr` and let it figure out the rest.

That's what's in this PR.  I change the naming to:

```
split.cpr-4
split.cpr-4-0
split.cpr-4-1
split.cpr-4-2
split.cpr-4-3
```

And I use the fact that you're trying to load into a replicated / distributed mesh to figure out whether or not to read a serial or distributed mesh.

It's not perfect however... because it does mean that you now can't read a serial cpr file into a distributed mesh (which I believe was supported before)... but that's an edge case I don't care much about and maybe we can make a hardcore parameter that would allow you to force that behavior.

Let me know what you think...

### DIFF
--- a/src/apps/splitter.C
+++ b/src/apps/splitter.C
@@ -153,8 +153,7 @@ int main (int argc, char ** argv)
           cpr.parallel() = true;
           cpr.binary() = binary;
           std::ostringstream outputname;
-          outputname << remove_extension(filename) << '.' << n_procs
-            << (binary ? ".cpr" : ".cpa");
+          outputname << remove_extension(filename) << (binary ? ".cpr" : ".cpa");
           cpr.write(outputname.str());
         }
     }
--- a/src/mesh/checkpoint_io.C
+++ b/src/mesh/checkpoint_io.C
@@ -96,7 +96,11 @@ void CheckpointIO::write (const std::string & name)
   // restarts later:
   if (this->processor_id() == 0)
     {
-      Xdr io (name, this->binary() ? ENCODE : WRITE);
+      std::ostringstream file_name_stream;
+
+      file_name_stream << name << "-" << (_parallel ? _my_n_processors : 1);
+
+      Xdr io (file_name_stream.str(), this->binary() ? ENCODE : WRITE);
 
       // write the version
       io.data(_version, "# version");
@@ -563,14 +567,22 @@ void CheckpointIO::read (const std::string & name)
   // We'll read a header file from processor 0 and broadcast.
   if (this->processor_id() == 0)
     {
+      std::string header_name = name;
+
       {
-        std::ifstream in (name.c_str());
+        std::ostringstream file_name_stream;
+
+        file_name_stream << name << "-" << (!mesh.is_replicated() ? _my_n_processors : 1);
+
+        header_name = file_name_stream.str();
+
+        std::ifstream in (header_name.c_str());
 
         if (!in.good())
-          libmesh_error_msg("ERROR: cannot locate header file:\n\t" << name);
+            libmesh_error_msg("ERROR: cannot locate header file:\n\t" << header_name);
       }
 
-      Xdr io (name, this->binary() ? DECODE : READ);
+      Xdr io (header_name, this->binary() ? DECODE : READ);
 
       // read the version, but don't care about it
       std::string input_version;
@@ -684,7 +696,10 @@ file_id_type CheckpointIO::read_header (const std::string & name)
   // We'll write a header file from processor 0 and broadcast.
   if (this->processor_id() == 0)
     {
-      Xdr io (name, this->binary() ? DECODE : READ);
+      std::ostringstream file_name_stream;
+      file_name_stream << name << "-" << (!mesh.is_replicated() ? _my_n_processors : 1);
+
+      Xdr io (file_name_stream.str(), this->binary() ? DECODE : READ);
 
       // read the version, but don't care about it
       std::string input_version;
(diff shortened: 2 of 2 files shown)


aeslaughter:
What is the difference between the "split.cpr" and "split.cpr-4-0"?

roystgnr:
So this changeset regresses serial-to-distributed reads as well as N-to-M reads?  Pass.

You've definitely got a valid use case, though, and it looks like the new naming scheme may be the best way to handle it.  Maybe if we just make the reading a little more flexible?

My inclination would be, if you try to open "foo.cpr":
1) We try to open a foo.cpr header; if it exists we parse it and use the subfiles associated with the number of processors in the header.  This gets us backwards compatibility with the current behavior, principle-of-least-surprise, and lets people manually precisely set up N-to-M or distributed-to-serial or whatever corner cases they want.
2) We try to open a foo.cpr-$n_procs header; if it exists we parse it and use the associated subfiles.  This solves your use case: you can make a bunch of presplits, open "foo.cpr" with a distributed process matching one of the presplits, and get the desired match.
3) We try to open a foo.cpr-"nearest match" header; if it exists we parse it and use the associated subfiles.  This means e.g. if you only did a presplit for 256 procs and you want to quickly try it on 128 you can do so without any further manual mesh mangling, you get the N-to-M read automatically.

(3) is an "aspirational" behavior, since if we have (1) then ```ln -s foo.cpr-256 foo.cpr``` isn't too too much work.  It wouldn't be too hard to add, though; we already depend on POSIX so opendir() is an option.

roystgnr:
> we already depend on POSIX so opendir() is an option.

Random aside: C++ now officially defines std::filesystem, as of **C++17**.  Wish I'd been a fly on the wall for that meeting.  "Hey guys, has anyone else noticed the C++ standard is nearly 20 years old and it still includes no way to open a directory?"  *crickets chirping*

permcody:
PIng! I think this issue needs to be revisited soon. I'm about ready to start running a few scaling studies and will definitely need pre-split for this to work for the number of cores I'm targeting. I'm going to run into the same issue Derek did here where I have to either write a script to change the name of the input mesh for each run, or hack libMesh to ignore the number so that my one input file works for any number of processors.

Looking back at this I can see that there are a few tricky cases we need to deal with but I'll just weigh in on my requirements.

1. I just need to be able to put a mesh filename in the input file and have MOOSE/libMesh do the right thing for reading in the right pre-splits.

That's really it. I don't really have a lot of need for N-to-M either but if you want to support that, fine. Looking over your options above let me make a few comments:

> 1. We try to open a foo.cpr header; if it exists we parse it and use the subfiles...

Are you suggesting a separate header only file, or are you suggesting reading one of the pre-splits? I'm a little wary of reading anything but a file from the "current" split because the header has the potential to get out of date with other nearby files. I can see somebody creating a listing of splits for several runs and then having a problem with say the biggest run. They end up re-splitting for some reason but the header doesn't get updated? Anyway, chaos ensues.

> 2. We try to open a foo.cpr-$n_procs header...

This seems like a pretty good choice to me.

> 3. Cool, but probably not necessary at this point.

Anyway, if you are good with 2, I'd like to bump priority on it. If you don't have time, either Derek or I can take a swing at it.

friedmud:
I was pretty sure that (2) is what I implemented here - is it not?

While we're changing this anyway - I really think it would be better to put the split out files in a _directory_ called `foo.cpr-#`... so that we can avoid having tens of thousands of files in the same directory when you do multiple splits.  On our filesystems here at INL that can cause them to chug.

roystgnr:
I think my comment got misinterpreted.  Those aren't three options, they're three *steps* in the same code.

I'll see if I can get an implementation (of steps 1+2) for you guys to try out by the end of the week.

permcody:
@roystgnr - just pinging you on this. We might update the submodule in MOOSE today. In fact @jwpeterson has already put in a PR to test out the current head. If this misses today's update, that's ok, but if it's close or ready to go, I'd personally love to see it in this update since I plan to run scaling studies over the Xmas break.

pbauman:
@permcody Would be very interested to see how your scaling runs go. @bboutkov is collecting some data as well and it's looking like #1468 is appearing for us too (just solving Poisson problem); hoping we'll be able to add more detail there, but if you do or do not see it, would be good to know.

permcody:
Sure, I'll be sharing back relevant details. I'm hoping to go up to 16K procs.

roystgnr:
@friedmud - the directory idea is a very good one.  I put the multiple-preset-selection support alone in #1537, since that seems most urgent, but I'll create another PR with split subdirectories when I can.

I'll probably do "write to subdirectories only; read from subdirectories if possible or from current directory if not" initially, for backwards compatibility, but then we can update the MOOSE unit test and get rid of the "current directory" behavior eventually.  The "read header files without a n_processors postfix" behavior I want to keep around forever, but in hindsight I don't see any good long-run reason for the "read split files from the current directory" behavior.

permcody:
Superseded by #1537.

PR #1458, opened 2017-10-19
Correctly output vector and scalar variables

lindsayad:
Before these changes, if the test below was run, the scalar output would be all NaNs. This was because `repeat_count` in the `equation_systems` write was never being incremented for the scalar; instead the increment was being applied to the components of the vector variable. This should be fixed here.


roystgnr:
Great catch; thanks!

Changing ```ZeroFunction<Real>``` to ```ZeroFunction<Number>``` ought to fix the --enable-complex failure.

tests/Makefile.am
```diff
@@ -46,6 +46,8 @@ unit_tests_sources = \
   mesh/spatial_dimension_test.C \
   mesh/mapped_subdomain_partitioner_test.C \
   mesh/mesh_function_dfem.C \
+  mesh/write_vec_and_scalar.C \
+  mesh/curl_curl_system.C \
```
jwpeterson: You added some new headers too, I think they should probably be in this list?

tests/mesh/write_vec_and_scalar.C
```diff
@@ -0,0 +1,131 @@
...
+    solver.max_nonlinear_iterations = 15;
+    solver.relative_step_tolerance = 1.e-3;
+    solver.relative_residual_tolerance = 1.0e-13;
+    solver.absolute_residual_tolerance = 0.0;
+
+    // And the linear solver options
+    solver.max_linear_iterations = 50000;
+    solver.initial_linear_tolerance = 1.e-10;
+
+    system.solve();
```
jwpeterson: Do we really need to do all this work (assembly/solve/etc) to test this? How about you just write some pre-determined values to the system solution vector, then write and Exodus file and read it back in to make sure they match? I'd prefer not to duplicate all this example code in the unit test (and run a solve!) if possible. What if the solve fails on e.g. Trilinos? That failure would have nothing to do with your actual test...

tests/mesh/write_vec_and_scalar.C
```diff
@@ -0,0 +1,131 @@
...
+    for (const auto & var_name : nodal_vars)
+      sys2.add_variable(var_name, SECOND, LAGRANGE);
+
+    es2.init();
+
+    for (const auto & var_name : nodal_vars)
+      exio.copy_nodal_solution(sys2, var_name, var_name, 1);
+
+    std::vector<Real> gold_vector(
+        {-0.020949, -0.020949, 0., 0.009495,  0.183852,  0.,       -0.010570, -0.010570, 0.250000,
```
jwpeterson: If you end up just writing pre-determined values to the solution vector we can shoot for a tighter tolerance on the checks as well.

jwpeterson:
Sorry for all the back and forth, but I think this version is now much better!

roystgnr:
This new unit test fails in parallel.

roystgnr:
And even fixing the obvious "don't try to set dofs you don't own" and "don't try to test dofs you don't own" errors isn't enough to change that.  I guess the gold_vector is assuming a fixed dof ordering, but partitioning and then keeping dofs contiguous within each partition means we can't provide that?

jwpeterson:
That must be the problem. Is there a way to fix it short of disabling this unit test in parallel?

roystgnr:
I'm giving it a try now; I believe we can just loop over nodes rather than over dofs for both setting and testing.

roystgnr:
It'll be fixed by #1475 once I get the --enable-complex path straight.

PR #1459, opened 2017-10-19
added a flag to indicate whether or not to close the matrix before solve.

fdkong:
It is useful for nonlinear eigenvalue solvers embedded in EPS recently. The nonlinear solvers have callbacks so that we do not need to close the matrix before solve. If we close the matrix before solve, the memory will be shrunk and the preallocation is not right any more.

### DIFF
--- a/include/solvers/eigen_solver.h
+++ b/include/solvers/eigen_solver.h
@@ -82,13 +82,24 @@ public:
   bool initialized () const { return _is_initialized; }
 
   /**
-   * \returns \p true if we want to close the matrix before solve,
-   * false otherwise. It is true by default.
+   * \returns \p The value of the flag which controls whether libmesh
+   * closes the eigenproblem matrices before solving. \true by
+   * default.
    */
-  bool & close_matrix_before_solve()
+  bool get_close_matrix_before_solve() const
   {
     libmesh_experimental();
     return _close_matrix_before_solve;
+  }
+
+  /**
+   * Set the flag which controls whether libmesh closes the
+   * eigenproblem matrices before solving.
+   */
+  void set_close_matrix_before_solve(bool val)
+  {
+    libmesh_experimental();
+    _close_matrix_before_solve = val;
   }
 
   /**
--- a/include/solvers/eigen_solver.h
+++ b/include/solvers/eigen_solver.h
@@ -81,6 +81,16 @@ public:
    */
   bool initialized () const { return _is_initialized; }
 
+  /**
+  * \returns \p true if we want to close the matrix before solve,
+  * false otherwise. It is true by default.
+  */
+  bool close_matrix_before_solve() const { return _close_matrix_before_solve; }
+
+  /**
+  * Set if or not to close the matrix before solve
+  */
+  void close_matrix_before_solve(bool close_matrix) { _close_matrix_before_solve = close_matrix; }
 
   /**
    * Release all memory and clear data structures.
@@ -269,6 +279,8 @@ protected:
   SolverConfiguration * _solver_configuration;
 
   Real _target_val;
+
+  bool _close_matrix_before_solve;
 };
 
 
@@ -283,7 +295,8 @@ EigenSolver<T>::EigenSolver (const Parallel::Communicator & comm_in) :
   _eigen_problem_type   (NHEP),
   _position_of_spectrum (LARGEST_MAGNITUDE),
   _is_initialized       (false),
-  _solver_configuration(libmesh_nullptr)
+  _solver_configuration(libmesh_nullptr),
+  _close_matrix_before_solve(true)
 {
 }
 
--- a/include/solvers/eigen_solver.h
+++ b/include/solvers/eigen_solver.h
@@ -82,15 +82,14 @@ public:
   bool initialized () const { return _is_initialized; }
 
   /**
-  * \returns \p true if we want to close the matrix before solve,
-  * false otherwise. It is true by default.
-  */
-  bool close_matrix_before_solve() const { return _close_matrix_before_solve; }
-
-  /**
-  * Set if or not to close the matrix before solve
-  */
-  void close_matrix_before_solve(bool close_matrix) { _close_matrix_before_solve = close_matrix; }
+   * \returns \p true if we want to close the matrix before solve,
+   * false otherwise. It is true by default.
+   */
+  bool & close_matrix_before_solve()
+  {
+    libmesh_experimental();
+    return _close_matrix_before_solve;
+  }
 
   /**
    * Release all memory and clear data structures.
--- a/src/solvers/slepc_eigen_solver.C
+++ b/src/solvers/slepc_eigen_solver.C
@@ -96,7 +96,8 @@ SlepcEigenSolver<T>::solve_standard (SparseMatrix<T> & matrix_A_in,
     libmesh_error_msg("Error: input matrix to solve_standard() must be a PetscMatrix.");
 
   // Close the matrix and vectors in case this wasn't already done.
-  matrix_A->close ();
+  if (this->_close_matrix_before_solve)
+    matrix_A->close ();
 
   return _solve_standard_helper(matrix_A->mat(), nev, ncv, tol, m_its);
 }
@@ -282,8 +283,11 @@ SlepcEigenSolver<T>::solve_generalized (SparseMatrix<T> & matrix_A_in,
     libmesh_error_msg("Error: inputs to solve_generalized() must be of type PetscMatrix.");
 
   // Close the matrix and vectors in case this wasn't already done.
-  matrix_A->close ();
-  matrix_B->close ();
+  if (this->_close_matrix_before_solve)
+  {
+    matrix_A->close ();
+    matrix_B->close ();
+  }
 
   return _solve_generalized_helper (matrix_A->mat(), matrix_B->mat(), nev, ncv, tol, m_its);
 }
@@ -321,7 +325,8 @@ SlepcEigenSolver<T>::solve_generalized (ShellMatrix<T> & shell_matrix_A,
     libmesh_error_msg("Error: inputs to solve_generalized() must be of type PetscMatrix.");
 
   // Close the matrix and vectors in case this wasn't already done.
-  matrix_B->close ();
+  if (this->_close_matrix_before_solve)
+    matrix_B->close ();
 
   ierr = MatShellSetOperation(mat_A,MATOP_MULT,reinterpret_cast<void(*)(void)>(_petsc_shell_matrix_mult));
   LIBMESH_CHKERR(ierr);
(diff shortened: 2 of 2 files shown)


fdkong:
@jwpeterson, could you take a look at this?

include/solvers/eigen_solver.h
```diff
@@ -81,6 +81,16 @@ class EigenSolver : public ReferenceCountedObject<EigenSolver<T> >,
    */
   bool initialized () const { return _is_initialized; }
 
+  /**
+  * \returns \p true if we want to close the matrix before solve,
+  * false otherwise. It is true by default.
+  */
+  bool closeMatrixBeforeSolve() const { return _close_matrix_before_solve; }
```
jwpeterson: Should be named `close_matrix_before_solve()`. libmesh uses lowercase letters and underscores in function names.

roystgnr:
The preallocation shrinking behavior is extremely annoying.  Is there no PETSc way to say "sync up parallel operations, but *don't* assume that you can discard currently-still-zero sparse entries?

jwpeterson:
> The preallocation shrinking behavior is extremely annoying. Is there no PETSc way to say "sync up parallel operations, but don't assume that you can discard currently-still-zero sparse entries?

Yeah, this is what the recently-added `SparseMatrix::flush()` command is for. @fdkong's original idea was to call `flush()` in the `EigenSolver` class, but he decided that might not be the correct default behavior in all applications. It is also not what we generally do in the `LinearSolvers` (we `close()` matrices for users there).

roystgnr:
>  @fdkong's original idea was to call ```flush()``` in the ```EigenSolver``` class, but he decided that might not be the correct default behavior in all applications.

Are we sure about that?  I know we use ```close()``` elsewhere, but I'm not confident that we're not wrong to do so.  If ```close()``` is slightly faster for most users, but horribly slower (or with current PETSc, an outright error) for others, then we might want to consider making ```if (!closed()) flush()``` the default internal behavior, leaving it to users to call ```close()``` if they determine that will benefit them.

fdkong:
> The preallocation shrinking behavior is extremely annoying. Is there no PETSc way to say "sync up parallel operations, but don't assume that you can discard currently-still-zero sparse entries?

`Flush` does that, but you have to call `close` before you actually going to use the matrix for the calculation. 

To have an efficiency computation for the sparse matrix, shrinking the memory is the right thing to do because all data are moved together to avoid doing extra zero operations such as  "0+0, 0*0".

I recently added one API (MatReSetPreallocation) that will give the original sparsity pattern back.

fdkong:
> Are we sure about that? I know we use close() elsewhere, but I'm not confident that we're not wrong to do so. If close() is slightly faster for most users, but horribly slower (or with current PETSc, an outright error) for others, then we might want to consider making if (!closed()) flush() the default internal behavior, leaving it to users to call close() if they determine that will benefit them.

In my option, libMesh does too much on `close`. I would like to remove all unnecessary `close`, and let users handle this. For example, before `solve` we may not call `close`. If users forget to close, they should go back to fix their code.  I just checked PETSc API again, and it looks like `close` is not free.

roystgnr:
> Flush does that, but you have to call close before you actually going to use the matrix for the calculation.

I don't actually see the purpose of flush, then.  In my head the whole reason to *want* "sync up parallel operations" *was* to be able to calculate.  I guess switching between add and set is nice?

> To have an efficiency computation for the sparse matrix, shrinking the memory is the right thing to do because all data are moved together to avoid doing extra zero operations such as "0+0, 0*0".

This is true if and only if you don't expect to fill those zeros later, right?  It seems reasonable to me to want to be able to tell PETSc "this entry is zero now, and I'd like to be able to calculate with it now, but I also want to make it non-zero later without forcing a huge reallocation".  IIRC current PETSc gives you an error rather than a reallocation by default, too, which makes it even more imperative to be able to avoid that.

roystgnr:
> I just checked PETSc API again, and it looks like close is not free.

It's certainly not, and removing redundant close() calls is a very good idea.

> I would like to remove all unnecessary close, and let users handle this. For example, before solve we may not call close. If users forget to close, they should go back to fix their code. 

This, on the other hand, seems backwards.  If we always need a close before solve, why not do that close in the library, where we know we need it?  That means less redundant code (since we have one library and many applications), less difficult code (since users then have one less thing to worry about), and just as much efficiency (since we'd still only be doing a single close call).

fdkong:
> I don't actually see the purpose of flush, then. In my head the whole reason to want "sync up parallel operations" was to be able to calculate. I guess switching between add and set is nice?

The purpose of flush is to switch between `set values` and` add values`.  Normal `close` will destroy your sparse pattern,  you lose the preallocation info and your code will be very inefficient.  `Close` should be called only once just right before we are going use. Most of time we should call `flush`.

roystgnr:
> The purpose of flush is to switch between set values and add values.

In that case, I correct the terminology of my previous lament:

Is there no PETSc way to say "prepare to calculate, but don't assume that you can discard currently-still-zero sparse entries"?

fdkong:
> This, on the other hand, seems backwards. If we always need a close before solve, why not do that close in the library, where we know we need it? That means less redundant code (since we have one library and many applications), less difficult code (since users then have one less thing to worry about), and just as much efficiency (since we'd still only be doing a single close call).

This is not alway true.  Users want to delay this call sometime. For example,  I do not want to close the matrix before `EPSSolve` because my matrix is empty at that point. I will fill up my matrix when SLEPc trigger the callback function in nonlinear eigenvalue solvers.  For some applications, forming the matrix may be delayed.  The matrix is not always ready to use right before `solve`.

fdkong:
> Is there no PETSc way to say "prepare to calculate, but don't assume that you can discard currently-still-zero sparse entries"?

We do not have this option.  But you can explicitly fill zeros according to your sparsity pattern, and then these zeros will be reserved.

roystgnr:
By solve() I meant for a LinearSolver.  We do always want a closed matrix before a linear solve, yes?  And so we have LinearSolver subclasses close it.

Conversely, for e.g. PetscNonlinearSolver::solve(), we don't close the matrix, since we know we don't expect to receive or want a closed matrix there; we close the matrix in the callback functions, where we do expect to need a closed matrix.

I'd have expected SlepcEigenSolver to be more like the former case, though, since we don't have any provision there for callback functions.  How are you managing to trick it into solving a nonlinear eigenproblem anyway?

fdkong:
> This is true if and only if you don't expect to fill those zeros later, right? It seems reasonable to me to want to be able to tell PETSc "this entry is zero now, and I'd like to be able to calculate with it now, but I also want to make it non-zero later without forcing a huge reallocation". IIRC current PETSc gives you an error rather than a reallocation by default, too, which makes it even more imperative to be able to avoid that.

I agree with you on this. It is the reason that we added one API (MatResetPreallocation) into PETSc recently.

fdkong:
> I'd have expected SlepcEigenSolver to be more like the former case, though, since we don't have any provision there for callback functions. How are you managing to trick it into solving a nonlinear eigenproblem anyway?

I recently added two nonlinear eigenvalue solvers into SLEPc EPS.  It is very similar to SNES, and we evaluate function residual and Jacobian via callbacks.

roystgnr:
> It is very similar to SNES, and we evaluate function residual and Jacobian via callbacks.

Okay, but:

What callbacks?

Should we have a NonlinearEigenSolver + SlepcNonlinearEigenSolver for them?  That would seem to clean up everything, wouldn't it?  Users would have a callback interface in NonlinearEigenSolver they could use, the SlepcNonlinearEigenSolver implementation would know to close matrices after a callback returns but not during solve(), and the SlepcEigenSolver would still be able to close matrices during solve().

fdkong:
> Should we have a NonlinearEigenSolver + SlepcNonlinearEigenSolver for them? That would seem to clean up everything, wouldn't it? Users would have a callback interface in NonlinearEigenSolver they could use, the SlepcNonlinearEigenSolver implementation would know to close matrices after a callback returns but not during solve(), and the SlepcEigenSolver would still be able to close matrices during solve().

The algorithms are just experimental at this point because we still do not have a good idea how to solve the nonlinear eigenvalue problems mathematically (where operators depend on the eigenvector).  We therefore do not have a good design in SLPEc level yet.  We could add `SlepcNonlinearEigenSolver` once we have more algorithms and a good design in SLEPc.

roystgnr:
>  jwpeterson approved these changes 18 minutes ago 

I'm sorry, but I'm still not seeing it.  It's a support method for a use case which is not yet usable in master, even after this PR, yes?  If you have a fixed matrix then you might as well close it; if you don't have a fixed matrix and you also don't have any callbacks from which to change it then it doesn't matter whether you close it or not, you will not be solving an eigenvalue problem today.

jwpeterson:
> It's a support method for a use case which is not yet usable in master, even after this PR, yes?

I think it fixes a current, actual problem in MOOSE... I agree that it would be better to eventually have a NonlinearEigenSolver class with it's own API, but this at least lets @fdkong work around a limitation of the current EigenSolver.

roystgnr:
Ok.  Can we tag it with libmesh_experimental() at least?  I'd like to remove it as soon as we get the final class structure worked out.

roystgnr:
I'd also still prefer "fill out the zeros on the app side" over "add a new weird method on the library side" as a temporary workaround, but I'm not insistent on it.

fdkong:
@roystgnr, here are callbacks https://github.com/idaholab/moose/blob/devel/framework/src/utils/SlepcSupport.C

roystgnr:
Thanks!  I don't see where the callbacks get assigned, though.  I assume you're just calling SlepcEigenSolver::eps() and manually setting them via that for now?

fdkong:
We compose functions to matrices (A and B) in `assemble_matrix`. SLEPc will retrieve the composed function and call these composed functions at the right time and in the right place. We form the Jacobian and residuals inside of  the composed functions.  

https://github.com/idaholab/moose/blob/devel/framework/src/base/NonlinearEigenSystem.C

roystgnr:
Yeah, that's definitely not the right way to go long-term, but I can see why it's very attractive.  So my only remaining concern is that we not advertise this as a long-term API.  Could we just stick a ```libmesh_experimental();``` call into the setter?

jwpeterson:
> Could we just stick a libmesh_experimental(); call into the setter?

Sure. I also didn't realize that there's a setter and getter with the same name in this PR (?) so I'll fix that as well and push to this same PR.

roystgnr:
:+1:, and let's merge as soon as Civet is done.  Thanks for putting up with my griping and questions.

As a side note: in the future, I actually prefer *not* to use combined setter/getter functions.  With a separate setter and getter it's really easy to change the underlying data structure or to add associated code.  But with a combined returns-a-reference setter/getter, the actual set doesn't take place until operator=, and inserting code there later becomes either hard (if user code only used the return values as temporaries and/or kept it in auto & variables, so you can replace reference-to-Foo with reference-to-custom-Foo-shim-class) or impossible (if user code actually tried to set a reference variable from your reference).  At that point you might as well just make your private variable public, since you can do the same shim trick with that.

(yes, I'm aware that I've written a ton of combined setter/getter accessors myself; it wasn't until I had to upgrade one and it turned around to bite me that I decided I was making a mistake)

jwpeterson:
> But with a combined returns-a-reference setter/getter, the actual set doesn't take place until operator=, and inserting code there later becomes either hard

This is a valid concern. Another one is the `const` issue, i.e. you can't use the getter aspect of this non-const combo function on a const object.

Shall we define a coding convention of no more combo getter/setter functions and standardized
```
FlagType get_flag_name() const;
void set_flag_name(value);
```
API names?

roystgnr:
> Shall we define a coding convention of no more combo getter/setter functions and standardized
> 
> ```
> FlagType get_flag_name() const;
> void set_flag_name(value);
> ```
> 
> API names?

Sounds perfect to me.

PR #1461, opened 2017-10-20
Updates to make ExodusII_IO::copy_elemental_solution work in parallel.

dknez:


dknez:
@jwpeterson this is based on our discussion on libmesh-users. This change fixes the issue for me.

PR #1462, opened 2017-10-25
PerfLog speedup

roystgnr:
Avoiding std::string allocation and deallocation speeds up PerfLog by a factor of 3 or 4 for me, and merging adjacent gettimeofday() calls speeds it up by about another factor of 2.

PerfLog is still not efficient enough to do very-fine-grained profiling, and if the remaining gettimeofday() calls are all that's left to blame then I may have to just remove the worst Heisen-profiling instances eventually, but at least now the worst-case scenario looks like 15% overhead rather than 115%.

### DIFF
--- a/include/utils/perf_log.h
+++ b/include/utils/perf_log.h
@@ -99,6 +99,7 @@ public:
   void   start ();
   void   restart ();
   double pause ();
+  double pause_for(PerfData & other);
   double stopit ();
 
   int called_recursively;
@@ -426,6 +427,24 @@ double PerfData::stop_or_pause(const bool do_stop)
 
 
 inline
+double PerfData::pause_for(PerfData & other)
+{
+  gettimeofday (&(other.tstart), libmesh_nullptr);
+
+  const double elapsed_time = (static_cast<double>(other.tstart.tv_sec  - this->tstart.tv_sec) +
+                               static_cast<double>(other.tstart.tv_usec - this->tstart.tv_usec)*1.e-6);
+  this->tot_time += elapsed_time;
+
+  other.count++;
+  other.called_recursively++;
+  other.tstart_incl_sub = other.tstart;
+
+  return elapsed_time;
+}
+
+
+
+inline
 double PerfData::stopit ()
 {
   // stopit is just similar to pause except that it decrements the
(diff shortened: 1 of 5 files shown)


include/utils/perf_log.h
```diff
@@ -279,9 +328,13 @@ class PerfLog
 
   /**
    * The actual log.
+   *
+   * An unsorted_map would work fine here and would be asymptotically
+   * faster, but in my tests for our log sizes there was no
+   * improvement.
```
jwpeterson: This was exactly what I was going to ask about, thanks for testing it already!

roystgnr: I almost just unthinkingly switched to unsorted_map after I had the prototype working with map, but I then ran across that "gifs of how different sorting algorithms work" gallery on Reddit, which got me reading about weird little O(N^2) algorithms that are actually still useful for small N, which reminded me that hey, our N here isn't exactly huge...

It's not obviously *slower* either, but it does require us to define our own pair hash function (or factor out the crummy one in the TopologyMap code), so it's definitely less terse.

roystgnr:
https://github.com/dterei/Scraps/tree/master/c/time suggests that we can shave 10% off our overhead and get better precision to boot if we use ```clock_gettime()``` instead of ```gettimeofday()```, but IMHO that may not be worth extra compatibility headaches: ```clock_gettime()``` made it into POSIX in 2001 but didn't make it into OSX until 2016.

jessecarterMOOSE:
@roystgnr Something about this PR doesn't appear to be compatible with MOOSE performance logging. It appears that all the data gets printed out fine, but when trying to get at the data with `get_perf_data()` in [framework/src/postprocessors/PerformaceData.C](https://github.com/idaholab/moose/blob/devel/framework/src/postprocessors/PerformanceData.C), it's just returning zeros.

roystgnr:
Gah, you're right, I broke ```get_perf_data()``` pretty thoroughly.  I'll put together a PR shortly.

roystgnr:
Is there a particular MOOSE regression test set I ought to be trying this with?  'postprocessors/print_perf_data'?

jessecarterMOOSE:
The current MOOSE test suite won't cover this because it only checks for the _existence_ of the output but not the _value_ of the data returned by `get_perf_data()` (if I'm reading the test specs right). I have a PR into MOOSE (idaholab/moose#10197) that adds a value test. Sure enough, it fails with the current libmesh master. You should be able to activate it with `./run_tests --re print_perf_data`

permcody:
Yeah - I've never come up with a good way to test for values in the perf
log. We have multiple tests (a few that use post processors) that report
values but we just do RE pattern matching to make sure there are numbers
there. Zero is a perfectly good number so it's hard to detect problems that
creep in.

PR #1463, opened 2017-10-25
Remove LIBMESH_BEST_UNORDERED_XYZ macros

jwpeterson:
These macros are no longer necessary now that we require a C++11 compliant compiler.

This branch does not implement shim classes for printing deprecation warnings (as suggested by @roystgnr) since I felt the benefit was not worth the extra complexity/hassle of adding 4 shims with multiple template arguments... They could still be added later if desired.


include/utils/topology_map.h
```diff
@@ -68,18 +68,8 @@ struct myhash {
...
-  defined(LIBMESH_HAVE_HASH_MAP)
-#  define MYHASH ,myhash
-#else
-#  define MYHASH
-#endif
-
-  typedef LIBMESH_BEST_UNORDERED_MAP<std::pair<dof_id_type, dof_id_type>,
-                                     dof_id_type MYHASH> map_type;
+  // We need to supply our own hash function.
+  typedef std::unordered_map<std::pair<dof_id_type, dof_id_type>, dof_id_type, myhash> map_type;
```
roystgnr: So much cleaner now.  Thanks!

roystgnr:
Hmm... okay, I agree that a runtime libmesh_deprecated(); would be a PITA.  But could we leave simple macro definitions for one more release, only wrapped in compile-time libmesh deprecation?, e.g.
```
#ifdef LIBMESH_ENABLE_DEPRECATED
#  define LIBMESH_BEST_UNORDERED_MAP unordered_map
// etc
#endif
```
That way we're not *instantly* yanking those macros away from our users.

jwpeterson:
The macros aren’t being yanked away though, they are still there...

roystgnr:
Wait, are they?  I thought they were being provided by AC_DEFINE, and 8106c6f was removing all those...  Oh, never mind; you're removing all *except* the std:: versions.

Okay, that's fine then.  I'd still like to deprecate the macros some day and remove them entirely a release or two later, but that doesn't have to be today.

jwpeterson:
Thanks for merging!

PR #1464, opened 2017-10-26
second order infinite elements

BalticPinguin:
Hi,
I have realised that second order infinite elements seem to be disabled without a reason, so I enabled them.

Closely related to this, (therefore part of this PR) is the fact that the inverse_map() function works quite unstable if one inserts points that are far away from the element tested (which regularly occurs when the PointLocator is used).
Thus, I added in the second commit a further criterium to the contains_point() function.
In my tests it works quite well; I hope, I didn't miss some cases not covered therein.

I hope, it is ok to put both changes into one PR and that the changes are fine so far.

### DIFF
--- a/include/geom/cell_inf_hex.h
+++ b/include/geom/cell_inf_hex.h
@@ -157,6 +157,13 @@ public:
    */
   virtual std::pair<Real, Real> qual_bounds (const ElemQuality q) const libmesh_override;
 
+  /**
+   * \returns \p true when this element contains the point
+   * \p p.  Customized for infinite elements, since knowledge
+   * about the envelope can be helpful.
+   */
+  virtual bool contains_point (const Point & p, Real tol=TOLERANCE) const libmesh_override;
+
 protected:
 
   /**
--- a/include/geom/cell_inf_hex8.h
+++ b/include/geom/cell_inf_hex8.h
@@ -138,13 +138,6 @@ public:
   { return 12; }
 
   /**
-   * \returns \p true when this element contains the point
-   * \p p.  Customized for infinite elements, since knowledge
-   * about the envelope can be helpful.
-   */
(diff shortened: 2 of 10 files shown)


roystgnr:
> I have realised that second order infinite elements seem to be disabled without a reason, so I enabled them.

In general, the allegory of [Chesterton's Fence](https://www.chesterton.org/taking-a-fence-down/) applies to situations "without a reason".

However, in this *specific* situation, there's a clear set of reasons:
1) None of the libMesh developers during some of the FE refactoring were comfortable with infinite elements, so we didn't dig into those code paths as deeply as we should have.
2) Without any test coverage for second-order infinite elements, we might not have even noticed that they'd been broken.

And it looks like you've fixed (1), so :+1: to merge.  I'll get it this afternoon, to allow time for others to look it over first.  Thanks!

If you'd like to fix (2) some time while you're at it, we'd greatly appreciate that, and it would be to your benefit as well: more test cases in CI would probably help keep your own uses of infinite elements safer from future regressions when us non-infinite-element users futz with the code.  IIRC the only InfFE test coverage we have at *all* right now is miscellaneous_ex1!

That almost sounds like a Mafia shakedown, "Nice code here; better give us another PR if youse don't want anything to happen to it", but I swear the request is pleading, not malicious.  I'm trying out some optimization ideas on private branches right now, I've *already* run into a case where I'll have to rework one of my optimizations for InfFE compatibility before putting it into master, and I'd feel like a jerk if I missed other places and bugs got merged because our test coverage wasn't enough to catch it either.

BalticPinguin:
ok, 'without a reason' was a bit clumsy, but now I know Chesterton's fence :+1: 

Yes, I am after fixing (2) actually; I hope, there will be some code soon for it. (and hopefully testing all relevant things.)

jwpeterson:
👍  I will convert the C-style comments in the .C files to C++ comments on merge if that's OK with everyone.

PR #1466, opened 2017-10-26
Remove spaces between right angle brackets.

jwpeterson:
We already test for this in our general C++11 configure test (now
required to pass), so no build system modifications should be
necessary.

* Remove spaces from between right angle brackets. May need to run multiple times.
perl -pli -e 's/>\s+>/>>/g' foo.C

* Remove whitespace following opening left angle brackets.
perl -pli -e 's/([a-z])<\s+/\1</g' foo.C


roystgnr:
Looks reasonable to me.  You only ran it over our stuff, right, not anything in contrib?

I was getting ready to fight about "getpot.h is shared with other projects and we should keep it C++98 compatible until we have a stronger reason not to" but it looks like it didn't have any double angle brackets to worry about in the first place.

:+1:

jwpeterson:
> You only ran it over our stuff, right, not anything in contrib?

Yeah.

> getpot.h

I generally treat this file as a contrib for the purposes of coding convention stuff.

PR #1467, opened 2017-10-26
 Add MeshBase::elements_range() and many range-based for loops

jwpeterson:
Subsequent PRs will add range functions for the other Elem and Node iterator types in MeshBase. 

Large number of commits as requested by @roystgnr.

### DIFF
--- a/include/mesh/distributed_mesh.h
+++ b/include/mesh/distributed_mesh.h
@@ -291,6 +291,10 @@ public:
   virtual element_iterator elements_end () libmesh_override;
   virtual const_element_iterator elements_begin() const libmesh_override;
   virtual const_element_iterator elements_end() const libmesh_override;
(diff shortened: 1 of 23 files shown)


roystgnr:
Hmm... could we call this element_ptr_range() instead?

I'd like eventually to be using an element_ref_range() everywhere, because there's no reason the user should have to worry about our internal layer of redirection, but since a pointer range is so much easier to implement we should start with that, we just shouldn't bless it with a definitive "this is THE way to range over elements" name.

jwpeterson:
OK. Where do you stand on the singular vs. plural aspect of the name? I was originally thinking e.g. active_elementS_begin/end should map to active_elementS_ptr_range but I'd be OK with either.

roystgnr:
I prefer singular over plural, but only weakly; your pick is fine.

PR #1469, opened 2017-10-27
Add MeshBase::node_ptr_range() and many range-based for loops

jwpeterson:
Also change the previous `elements_range()` function to `element_ptr_range()`.

Follow on work to #1467.


roystgnr:
:+1: and I'm loving these "+N, -2N" changesets.

Issue #1465, opened 2017-10-26
change type of "current frequency"


BalticPinguin:
Dear all,
in libmesh there is a parameter called 
`equation_systems.parameters.get<Real>("current frequency")`
which is used in `src/fe/fe_compute_data.C` and `src/systems/frequency_system.C`.
I think, its only purpose is the oscillation of the infinite elements.  Am I right with this or do people use the FrequencySystem for some non-related purposes (i.e. not Helmholtz-like problem)?

So what I would like to have is that the "current frequency" is rather a 'Number', thus the oscillatory behaviour can become an exp. damping (or damped oscillation).
At least in quantum mechanics, this makes sense.
My suggestion would be to either

- change the type from Real to Number or
- add a second parameter \<Number\>("current frequency") and use this if it is set and  \<Real\>("current frequency") in other cases.

Probably the latter is advantageous since existing code doesn't need to be changed, but this leads to a bit more ugly code.
Does someone have any other/better idea, or would you rather like not to touch this at all?
Best,
Hubert

jwpeterson:
The first option sounds OK to me, and isn't it kind of backwards-compatible in a way? Anyone using a real-valued "current frequency" before will still be able to do so after it's changed to `Number`... possibly with some trivial code changes.

I don't think many (or any?) other people currently use the infinite elements, so I'm much more willing to change APIs in those parts of the library than others...

roystgnr:
Definitely with some trivial user code changes required; there's no mechanism by which Parameter<Real> can be converted to Parameter<Number> or vice-versa, and there's no mechanism with which Parameters can hold two Parameter objects of the same name.

My inclination with regards to infinite elements stuff is indeed "Hubert does whatever he wants here, because this is the most love that our infinite elements code has seen in years".  However, this isn't infinite elements code; IIRC we've actually had a couple other FrequencySystem users in recent years, even though none of our flagship apps or frameworks touch it.

The "proper" way to handle this upgrade IMHO would be to have the FrequencySystem code check for ```parameters.have_parameter<Number>("current frequency")```, and use it if that returns true, or use the Real type and report ```libmesh_deprecated()``` otherwise.

[...]

PR #1470, opened 2017-10-27
Changed type of 'current frequency' to Number

BalticPinguin:
this implements the changes discussed in #1465 .
I hope, this is somehow in agreement with your notion. 

Just let me know, if I should change something (e.g. I was not sure which notation I should use for explicit casts).

### DIFF
--- a/examples/miscellaneous/miscellaneous_ex2/miscellaneous_ex2.C
+++ b/examples/miscellaneous/miscellaneous_ex2/miscellaneous_ex2.C
@@ -113,9 +113,9 @@ int main (int argc, char ** argv)
   // This example requires at least 2D support.
   libmesh_example_requires(2 <= LIBMESH_DIM, "2D support");
 
-  // Check for proper usage.
-  if (argc < 3)
-    libmesh_error_msg("Usage: " << argv[0] << " -f [frequency]");
+  // Check for proper usage. frequency has two terms:
+  if ( argc <4 )
+    libmesh_error_msg("Usage: " << argv[0] << " -f [real_frequency imag_frequency]");
 
   if (init.comm().size() > 1)
     {
@@ -140,7 +140,12 @@ int main (int argc, char ** argv)
 
   // Get the frequency from argv[2] as a float, solve for 1/3rd, 2/3rd
   // and 1/1th of the given frequency.
-  const Real frequency_in = atof(argv[2]);
+
+  const Number frequency_in (atof(argv[2]), atof(argv[3]));
+
+  // this must work as well, but is deprecated.
+  // const Real frequency_in = atof(argv[2]);
+
   const unsigned int n_frequencies = 3;
 
   // Create a mesh, with dimension to be overridden later, distributed
@@ -203,7 +208,7 @@ int main (int argc, char ** argv)
   // may directly access them. Must be called before the
   // EquationSystems object is initialized.  Will solve for 1/3,
   // 2/3, and 1 times the given frequency.
-  f_system.set_frequencies_by_steps (frequency_in/n_frequencies,
+  f_system.set_frequencies_by_steps (frequency_in/static_cast<Number>(n_frequencies),
                                      frequency_in,
                                      n_frequencies);
 
@@ -436,7 +441,7 @@ void assemble_helmholtz(EquationSystems & es,
       // If the element has no neighbor on a side then that
       // side MUST live on a boundary of the domain.
       for (auto side : elem->side_index_range())
-        if (elem->neighbor(side) == libmesh_nullptr)
+        if (elem->neighbor_ptr(side) == libmesh_nullptr)
           {
             LOG_SCOPE("damping", "assemble_helmholtz");
 
@@ -519,13 +524,13 @@ void add_M_C_K_helmholtz(EquationSystems & es,
     es.get_system<FrequencySystem> (system_name);
 
   // Get the frequency, fluid density, and sound speed of the current solve.
-  const Real frequency = es.parameters.get<Real> ("current frequency");
+  const Number frequency = es.parameters.get<Number> ("current frequency");
   const Real rho       = es.parameters.get<Real> ("rho");
   const Real speed     = es.parameters.get<Real> ("wave speed");
 
   // Compute angular frequency omega and wave number k
-  const Real omega = 2.0*libMesh::pi*frequency;
-  const Real k     = omega / speed;
+  const Number omega = 2.0*libMesh::pi*frequency;
+  const Number k     = omega / speed;
 
   // Get writable references to the system matrix and vector, where the
   // frequency-dependent system is to be collected.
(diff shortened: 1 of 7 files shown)


roystgnr:
We prefer C++-style casts, e.g. ```static_cast<Number>(frequency_in)```

examples/miscellaneous/miscellaneous_ex2/miscellaneous_ex2.C
```diff
@@ -436,7 +441,7 @@ void assemble_helmholtz(EquationSystems & es,
       // If the element has no neighbor on a side then that
       // side MUST live on a boundary of the domain.
       for (auto side : elem->side_index_range())
-        if (elem->neighbor(side) == libmesh_nullptr)
+        if (elem->neighbor_ptr(side) == libmesh_nullptr)
```
roystgnr: Ha!  I thought I finally got rid of the last of our deprecated code when I built with the new --disable-deprecated, but I never tried --enable-complex --disable-deprecated.  Thanks!

examples/miscellaneous/miscellaneous_ex2/miscellaneous_ex2.C
```diff
@@ -543,9 +548,9 @@ void add_M_C_K_helmholtz(EquationSystems & es,
 
   // Compute the scale values for the different operators.
   const Number scale_stiffness (1., 0.);
-  const Number scale_damping (0., omega);
-  const Number scale_mass (-k*k, 0.);
-  const Number scale_rhs (0., -(rho*omega));
+  const Number scale_damping=sqrt(-(Number)1.)*omega;
```
roystgnr: Why is this preferable to the two-argument constructor?  It'll all compile to the same thing in the end, but I find the former easier to read.

examples/miscellaneous/miscellaneous_ex2/run.sh
```diff
@@ -6,4 +6,4 @@ source $LIBMESH_DIR/examples/run_common.sh
 
 example_name=miscellaneous_ex2
 
-run_example "$example_name" -f .5
+run_example "$example_name" -f .5 0.1 i
```
roystgnr: I guess the "i" is supposed to be sort of like a comment here?  It was more like a red herring to me; I went back to hunt for what exactly the code was doing with argv[4]...

src/fe/fe_compute_data.C
```diff
@@ -58,14 +57,23 @@ void FEComputeData::init ()
     this->speed = this->equation_systems.parameters.get<Real>("speed");
 
   if (equation_systems.parameters.have_parameter<Real>("current frequency"))
-    this->frequency = this->equation_systems.parameters.get<Real>("current frequency");
+    {
+
+    // please use the type Number instead.
+    libmesh_deprecated();
+    this->frequency = static_cast<Number> (this->equation_systems.parameters.get<Number>("current frequency"));
+    }
```
roystgnr: This isn't right.  You need to be using ```get<Real>``` if we're in the ```have_parameter<Real>``` case.

src/fe/inf_fe_static.C
```diff
@@ -304,18 +304,18 @@ void InfFE<Dim,T_radial,T_map>::compute_data(const FEType & fet,
 #ifdef LIBMESH_USE_COMPLEX_NUMBERS
 
   // assumption on time-harmonic behavior
-  const short int sign (-1);
+  Number sign (0,1.);
```
roystgnr: Should we rename this sign_i, if it's now encompassing both parts of ```sign*i``` in the comment below?

src/systems/frequency_system.C
```diff
@@ -208,20 +208,20 @@ void FrequencySystem::set_frequencies_by_steps (const Real base_freq,
...
+                                                const Number max_freq,
                                                 const unsigned int n_freq,
                                                 const bool allocate_solution_duplicates)
 {
   this->_keep_solution_duplicates = allocate_solution_duplicates;
 
-  // sanity checks
-  libmesh_assert_greater (max_freq, min_freq);
+  // sanity checks. Only look at the real part.
+  libmesh_assert_greater (std::real(max_freq), std::real(min_freq));
```
roystgnr: Should this be libmesh_assert_greater_equal now?  Could users want to step through different damping levels at the same frequency?

src/systems/frequency_system.C
```diff
@@ -290,6 +327,7 @@ void FrequencySystem::set_frequencies (const std::vector<Real> & frequencies,
 
 
 
+
```
roystgnr: A bad whitespace change leaked in

examples/miscellaneous/miscellaneous_ex2/run.sh
BalticPinguin: oh yes, it is rather an artefact; initially the i was read as well. But lets remove it.

examples/miscellaneous/miscellaneous_ex2/miscellaneous_ex2.C
BalticPinguin: yes, it would be; but since omega is complex as well, the two-argument constructor doesn't work. 
But I'll add a variable for i; than it looks already better.

BalticPinguin:
ok, thanks for your comments. I think I fixed them now.

examples/miscellaneous/miscellaneous_ex2/miscellaneous_ex2.C
```diff
@@ -543,9 +548,9 @@ void add_M_C_K_helmholtz(EquationSystems & es,
 
   // Compute the scale values for the different operators.
   const Number scale_stiffness (1., 0.);
-  const Number scale_damping (0., omega);
-  const Number scale_mass (-k*k, 0.);
-  const Number scale_rhs (0., -(rho*omega));
+  const Number scale_damping=I*omega; // I is imaginary unit (from complex.h)
```
roystgnr: Not portably, I'm afraid.  I is the imaginary unit from complex.h in C99, which isn't guaranteed to be supported by or compatible with C++ compilers.  In fact, technically if you don't explicitly ```#include "complex.h"```, you just expect it to be included under the hood by ```#include <complex>```, then "I" is just polluting the global namespace in a C++ context, so any compiler/stdlib that *does* support it is violating the standard!

This appears to be why the Civet --enable-complex build barfed.

You can write 1i (after using the complex_literals sub-namespace) in C++14, but we don't demand users have C++14 support yet, and if you want to use i in C++11 then I'm afraid you'll have to define it yourself.

examples/miscellaneous/miscellaneous_ex2/miscellaneous_ex2.C
```diff
@@ -113,9 +113,9 @@ int main (int argc, char ** argv)
   // This example requires at least 2D support.
   libmesh_example_requires(2 <= LIBMESH_DIM, "2D support");
 
-  // Check for proper usage.
-  if (argc < 3)
-    libmesh_error_msg("Usage: " << argv[0] << " -f [frequency]");
+  // Check for proper usage. frequency can have 1 or three terms:
```
roystgnr: If you're forcing argc >= 4 then frequency must always have 2 terms, no?

BalticPinguin: yes; I think this is okay here because the parameter will be of type number anyway.
But you are right; the comment  is misleading.

examples/miscellaneous/miscellaneous_ex2/miscellaneous_ex2.C
BalticPinguin: ok, I will use an own variable. Thanks for the explanation.

roystgnr:
Thanks!

In the future it would probably be best to address comments with additional commits, and save the interactive-rebase+force-push until after the final changeset is otherwise complete.  When you just have a couple lines to fixup it'd be much faster for me to look over just those lines rather than looking over a whole brand-new diff out of fear that some unrelated+unreviewed other changes had crept in too.

(although personally I think this is a missing feature from git - it would be nice if it were somehow possible to look at both the history of how commits were made and the "history" of how commits *should* have been made, without overwriting the former to create the latter)

PR #1471, opened 2017-10-30
Add MeshBase::active_element_ptr_range() and many range-based for loops

jwpeterson:
And goodbye to another ~700 lines of boilerplate iterator initialization.

### DIFF
--- a/examples/eigenproblems/eigenproblems_ex3/eigenproblems_ex3.C
+++ b/examples/eigenproblems/eigenproblems_ex3/eigenproblems_ex3.C
@@ -158,21 +158,13 @@ int main (int argc, char ** argv)
 
   mesh.read(mesh_name_exodus.str());
 
-  {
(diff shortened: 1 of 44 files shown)


roystgnr:
Better a build error than a runtime failure, trust me.

roystgnr:
That was definitely one of my categories of error on the neighbor_ptr_range stuff, though: I'd copy and paste the loop statements from one loop to another, not noticing that the first loop was over "neigh" and the second over "n" or "s" or some such.

jwpeterson:
> Better a build error than a runtime failure, trust me.

Yeah, had one of those too that I found locally with bisection. A loop that was reseating the loop iteration variable which meant that we could not use `auto &`.

The compile failure was because I had my local build configured with infinite elements enabled, and large sections of the file in question were ifdef'd out. Ironically I configure with infinite elements enabled to get *more* test coverage, not less :P

PR #1473, opened 2017-10-31
Add MeshBase::active_local_element_ptr_range() 

jwpeterson:
This removes the majority of the remaining mesh iterator boilerplate code. 

I count about 150 other miscellaneous mesh iterator loops (`flagged_pid_elements_begin()`, etc.) but I don't plan on fixing them since the net number of lines saved won't be as large.

### DIFF
--- a/src/error_estimation/adjoint_refinement_estimator.C
+++ b/src/error_estimation/adjoint_refinement_estimator.C
@@ -370,93 +370,72 @@ void AdjointRefinementEstimator::estimate_error (const System & _system,
   std::unordered_set<dof_id_type> processed_node_ids;
 
   // We will be iterating over all the active elements in the fine mesh that live on
-  // this processor
-  {
-    MeshBase::const_element_iterator elem_it = mesh.active_local_elements_begin();
-    const MeshBase::const_element_iterator elem_end = mesh.active_local_elements_end();
-
-    // Start loop over elems
-    for (; elem_it != elem_end; ++elem_it)
+  // this processor.
+  for (const auto & elem : mesh.active_local_element_ptr_range())
+    for (unsigned int n=0; n != elem->n_nodes(); ++n)
       {
-        // Pointer to this element
-        const Elem * elem = *elem_it;
+        // Get a reference to the current node
+        const Node & node = elem->node_ref(n);
 
-        // Loop over the nodes in the element
-        for (unsigned int n=0; n != elem->n_nodes(); ++n)
+        // Get the id of this node
+        dof_id_type node_id = node.id();
+
+        // If we havent already processed this node, do so now
+        if (processed_node_ids.find(node_id) == processed_node_ids.end())
           {
-            // Get a reference to the current node
-            const Node & node = elem->node_ref(n);
+            // Declare a neighbor_set to be filled by the find_point_neighbors
+            std::set<const Elem *> fine_grid_neighbor_set;
 
-            // Get the id of this node
-            dof_id_type node_id = node.id();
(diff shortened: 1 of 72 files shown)


src/error_estimation/adjoint_refinement_estimator.C
```diff
@@ -370,93 +370,72 @@ void AdjointRefinementEstimator::estimate_error (const System & _system,
...
-  // this processor
-  {
-    MeshBase::const_element_iterator elem_it = mesh.active_local_elements_begin();
-    const MeshBase::const_element_iterator elem_end = mesh.active_local_elements_end();
-
-    // Start loop over elems
-    for (; elem_it != elem_end; ++elem_it)
+  // this processor.
+  for (const auto & elem : mesh.active_local_element_ptr_range())
+    for (unsigned int n=0; n != elem->n_nodes(); ++n)
```
roystgnr: Huh.  Why didn't I change this to a node_ref_range()?

src/error_estimation/adjoint_refinement_estimator.C
```diff
@@ -529,18 +502,12 @@ void AdjointRefinementEstimator::estimate_error (const System & _system,
...
 
   for (std::size_t j=0; j != system.qoi.size(); j++)
-    {
-      if (_qoi_set.has_index(j))
-        {
-          delete coarse_adjoints[j];
-        }
-    }
+    if (_qoi_set.has_index(j))
+      delete coarse_adjoints[j];
```
roystgnr: I'm still thrilled by the number of lines of code this PR strips, but slipping in unrelated whitespace and brace removal doesn't count for your final score.

src/mesh/ensight_io.C
```diff
@@ -198,19 +198,13 @@ void EnsightIO::write_geometry_ascii()
...
-        ensight_parts_map[elem->type()].push_back(elem);
+  for (const auto & elem : the_mesh.active_local_element_ptr_range())
+    {
+      ensight_parts_map[elem->type()].push_back(elem);
 
-        for (unsigned int i = 0; i < elem->n_nodes(); i++)
-          mesh_nodes_map[elem->node_id(i)] = elem->point(i);
-      }
-  }
+      for (unsigned int i = 0; i < elem->n_nodes(); i++)
```
roystgnr: And this I should have turned into a node_index_range() already.  I'll make another pass at those later.

src/error_estimation/adjoint_refinement_estimator.C
jwpeterson: Maybe this one loop didn't match your regex...

src/error_estimation/adjoint_refinement_estimator.C
jwpeterson: The brace removal counts! Since we don't have to have the extra line for dereferencing an iterator, it allowed several extra braces to be collapsed.

src/mesh/ensight_io.C
jwpeterson: 👍

roystgnr:
Awesome, thanks!

I've noticed that with mesh iterators we never seemed to make the mistake of calling foo_end() inside the loop test, though...  So if we want to improve our iteration performance then we *actually* need to improve our iterator performance, there was no simple-to-fix inefficient code the way there was for the element node and element children ranges.

src/error_estimation/adjoint_refinement_estimator.C
roystgnr: Most of the brace removal counts, but this set was an exception.

PR #1475, opened 2017-10-31
Parallelize recent unit test

roystgnr:
This gets our WriteVecAndScalar unit test working consistently in parallel, using the same gold_vector as before, by using a fixed node ordering and querying for ```dof_number()``` values to determine what dof ordering we ended up with.

Could @brianmoose or someone change the "Test"/"Check" and "Test Debug"/"Check" Civet targets to add something like ```LIBMESH_RUN='mpiexec -np 16'``` to the environment?  I'd rather not let something like this slip through unnoticed again.

### DIFF
--- a/tests/mesh/write_vec_and_scalar.C
+++ b/tests/mesh/write_vec_and_scalar.C
@@ -162,15 +162,15 @@ public:
         CPPUNIT_ASSERT_DOUBLES_EQUAL(libmesh_real(sys2_soln(node->dof_number(0,1,0))),
                                      0.0, tol);
         CPPUNIT_ASSERT_DOUBLES_EQUAL(libmesh_real(sys2_soln(node->dof_number(0,2,0))),
-                                     gold_vector[gold_i_ux], tol);
+                                     std::abs(gold_vector[gold_i_ux]), tol);
         CPPUNIT_ASSERT_DOUBLES_EQUAL(libmesh_real(sys2_soln(node->dof_number(0,4,0))),
                                      0.0, tol);
         CPPUNIT_ASSERT_DOUBLES_EQUAL(libmesh_real(sys2_soln(node->dof_number(0,5,0))),
-                                     gold_vector[gold_i_uy], tol);
+                                     std::abs(gold_vector[gold_i_uy]), tol);
         CPPUNIT_ASSERT_DOUBLES_EQUAL(libmesh_real(sys2_soln(node->dof_number(0,7,0))),
                                      0.0, tol);
         CPPUNIT_ASSERT_DOUBLES_EQUAL(libmesh_real(sys2_soln(node->dof_number(0,8,0))),
-                                     gold_vector[gold_i_v], tol);
+                                     std::abs(gold_vector[gold_i_v]), tol);
 #endif
       }
 #endif // #ifdef LIBMESH_HAVE_EXODUS_API
--- a/tests/mesh/write_vec_and_scalar.C
+++ b/tests/mesh/write_vec_and_scalar.C
@@ -86,6 +86,7 @@ public:
         if (mesh.n_processors() == 1)
           CPPUNIT_ASSERT_EQUAL(initial_vector[node->id()], dof_id*Real(2));
       }
+    sys_solution.close();
 
 #ifdef LIBMESH_HAVE_EXODUS_API
 
--- a/tests/mesh/write_vec_and_scalar.C
+++ b/tests/mesh/write_vec_and_scalar.C
@@ -132,15 +132,44 @@ public:
 
     Real tol = 1e-12;
     NumericVector<Number> & sys2_soln(*sys2.solution);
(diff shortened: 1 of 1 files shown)


jwpeterson:
You want to run the examples *and* unit tests in parallel? Would we also run them in serial? This is likely to add quite a bit of time to the testing of each PR.

brianmoose:
Also, 16 might be a bit too much for our build machines, how about 12?

roystgnr:
> You want to run the examples and unit tests in parallel? Would we also run them in serial?

That's a good question... I'm tempted to say no, since I can't recall ever seeing a serial bug that didn't exhibit in parallel too, but I know we do have some ```if (!n_processors) cheaper_path();``` code here and there...

Maybe we'd run "Test Debug"/"Check" in parallel, "Test"/"Check" in serial, as a compromise?  Not perfect but IMHO strictly better than the current tests.

> This is likely to add quite a bit of time to the testing of each PR.

Really?  I guess Civet is smart enough not to dedicate a whole machine to a sub-test that it knows will only use 1 processor, but if we throw a ton of processors at it then the whole machine gets tied up?

> 16 might be a bit too much for our build machines, how about 12?

If John is right, then I'd say go as low as 4 or even 3 procs.  That'll catch 90% of parallel bugs in my experience, and 100% of the obvious bugs.  The non-obvious bugs are usually my fault, but I run my private tests at -np 24 now so (fingers crossed) hopefully I won't need Civet as a backstop...

That last clause reeks of hubris, doesn't it?  Now I'm practically guaranteed to commit something soon that works for me on DistributedMesh but breaks on ReplicatedMesh at high processor counts.  Still, I think testing at 3 or 4 would be a huge step up from 1.

Oh, wait, I see I've already just committed a "fix" which breaks for --enable-complex.  Maybe Nemesis let me off the hook quickly.

brianmoose:
I added `LIBMESH_RUN = 'mpiexec -np 4'` to "Test debug"/Check.
It will be in effect on your next push (or you can invalidate the current one).
I realized after I merged this to civet_recipes that maybe you wanted to wait since other libmesh PRs will likely fail until this PR gets merged...

roystgnr:
Actually, it didn't seem to take effect yet?  #1478 ran Test Debug/Check without LIBMESH_RUN set, anyway.

Still, this fix works for me in parallel and it works for Civet Test enable complex in serial (we should probably run that with at least np 2 as well, huh...) so I'm declaring victory and merging.

brianmoose:
It looks like #1478 is using `mpiexec -np 4`, at least on the examples where it shows the command being run. Not sure if it took affect on the other checks...

roystgnr:
Hmm... okay, I think I understand my confusion:

Civet isn't showing the *whole* environment at the top of it's output, just a few critical bits like LIBMESH_DIR, so we don't see LIBMESH_RUN there.

In the Test Debug runs, we don't configure with --enable-glibcxx-debugging-cppunit, and we're only building dbg mode with glibcxx-debugging still enabled, so the unit tests never get run, so the at-that-point-broken-in-parallel unit test didn't scream and die.

This throws a wrench into my "Test Debug in parallel, Test in serial" suggestion.

Is it too much to hope that the version of cppunit on your build systems works with --enable-glibcxx-debugging-cppunit?  The only other alternative is --disable-glibcxx-debugging, which I'd rather avoid.

brianmoose:
I don't know if we have tried that. I remember moose having problems with cppunit in debug mode but I will try it on one of the build machines and see what happens.

brianmoose:
We are only looking at unit tests right? run_unit_tests.sh explicitly skips running unit tests in dbg mode.
Running unit_tests-dbg manually results in a crash. This is libmesh compiled with --enable-glibcxx-debugging-cppunit.

jwpeterson:
I think glibcxx debugging enabled unit tests will need to wait until one of us finds time to convert our unit tests over to the gtest framework, which is header-only and doesn't rely on an external library the way cppunit does. I suppose it might be theoretically possible to compile cppunit itself with the required glibcxx debugging flags, but I don't think anyone has actually tried that.

roystgnr:
Ok, I was afraid of that.  That leaves no good short-term options, then, since we aren't running any devel checks.  At least parallel test coverage of the example codes is better than nothing.

brianmoose:
So, the only change you want to civet recipes is to add the LIBMESH_RUN to "Check enable complex"/Check?

roystgnr:
Let's keep the np 4 on "Test debug"/Check.

Adding np 2 to either "Test enable complex"/Check or "Test"/Check too would be a good idea - that way we get at least one CI test running the unit tests in parallel, even if only in opt mode.  The buggy unit test fixed here was a segfault in opt mode, IIRC, rather than a nice clear error message in dbg and devel modes, but we don't really need the nice clear error message from CI, we just need it to notice that *something* is wrong.

PR #1478, opened 2017-10-31
Add as_range() helper function.

jwpeterson:
This lets us treat the iterator `std::pairs` returned by `equal_range()` as ranges for the purpose of using range-based for loops. This in turn allows us to remove more bolierplate code.


roystgnr:
We do seem to be getting decreasing marginal utility as we progress from "range stuff that makes code run faster" to "range stuff that makes code much clearer" to "range stuff that makes code somewhat clearer", but it's still progress.

:+1:

PR #1479, opened 2017-11-01
Added miscellaneous_ex14

BalticPinguin:
So this is a suggestion for a new example with infinite elements. It tests
 - second order elements
 - the inverse_map() function, I was working on some time ago
 - the new feature of imaginary 'current frequency'

I hope, the syntax etc. is okay and I didn't insert too many errors.


roystgnr:
We need quite a few ```#ifdef``` and ```libmesh_example_requires()``` in here; lots of people build libMesh without eigensolvers, or with eigensolvers but without infinite elements.

BalticPinguin:
yes, I see. I thought, I would have caught the major parts already, but seemingly not...

BalticPinguin:
I must admit that I don't understand, at which point 'Test' fails. Probably you can help me with this? For me it looks as if it were not related to this pr...

jwpeterson:
> I must admit that I don't understand, at which point 'Test' fails. Probably you can help me with this?

Yeah, I was attempting to add `-std=c++11` to our CIVET testing, but there are issues with doing that so I'm reverting it for the time being...

doc/html/src/examples.html
```diff
@@ -175,6 +175,10 @@
 
 <li><L1><a href="examples/miscellaneous_ex12.html">MITC4 Shell Elements</a></L1></li>
 
+<li><L1><a href="examples/miscellaneous_ex13.html">Quad8 Shell Elements</a></L1></li>
+
+<li><L1><a href="examples/miscellaneous_ex14.html">Infinite Elements With Imaginary Frequency</a></L1></li>
+
```
roystgnr: Wow, we've been forgetting to keep this up to date... Thanks!

examples/miscellaneous/miscellaneous_ex14/miscellaneous_ex14.C
```diff
@@ -0,0 +1,685 @@
...
+        case SINVERT:
+          // this method has been renamed in 3.1
+#if SLEPC_VERSION_LESS_THAN(3,1,0)
+          ierr = STSetType(st, STSINV);
+#else
+          ierr = STSetType(st, STSINVERT);
+#endif
+          break;
+        case CAYLEY:
+#if SLEPC_VERSION_LESS_THAN(2,2,1)
```
roystgnr: Could this case get turned into a libmesh_example_requires() in main?

Thanks for these, by the way; we currently have no test coverage for older SLEPc, but that doesn't mean we have no users stuck with older SLEPc.

roystgnr:
Code review gets :+1: from me; my only remaining gripe is about a problem that won't affect many (any?) current users, so really isn't worth delaying the merge.

Waiting to retry "Test" probably isn't worth it, either, since the failure is clearly an unrelated issue.

However... none of our CI tests are built with SLEPc, are they?  Neither for GRINS nor for MOOSE?  Let me check this out and try it privately before I hit the big button.

If you do have time to move around that outdated-SLEPc-case code, do so in a separate commit, without rebase/force-push, [please](https://github.com/libMesh/libmesh/pull/1470#issuecomment-340825777).  I do like using "fixup" in cases where an intermediate commit won't pass some tests, to make future bisection easier, but nobody's going to be doing bisection in the future using SLEPc from the distant past.

jwpeterson:
> ... none of our CI tests are built with SLEPc, are they?

Not currently, but I think we do now have SLEPc in the MOOSE package, so I am going to add an optional SLEPc CIVET recipe to the list of attachable recipes for libmesh PRs. I was going to do this yesterday until I broke all libmesh PR testing with the -std=c++11 thing.......

examples/miscellaneous/miscellaneous_ex14/miscellaneous_ex14.C
jwpeterson: The oldest ersion of PETSc we still support is 2.3.3, so SLEPc support should not need to extend prior to that either...

jwpeterson:
> Waiting to retry "Test" probably isn't worth it, either, since the failure is clearly an unrelated issue.

I reverted that CIVET recipe, so I was going to invalidate this PR and see if it works again if that's OK with you.

roystgnr:
> so I am going to add an optional SLEPc CIVET recipe to the list of attachable recipes for libmesh PRs.

Silly question: why make it optional?  Could we build one of {Test, Test Debug} with SLEPc and the other without?  If we add SLEPc to Test Debug, the whole setup ought to be nearly as fast to run, we'll get default coverage of both cases.  We could also have optional "Test Debug Without SLEPc" and "Test With SLEPc" recipes, but I don't think we'd ever be paranoid enough to use them unless someone reported a specific bug.

> I was going to invalidate this PR and see if it works again if that's OK with you.

Sure, thanks.  They're your CPU cycles.  ;-)

examples/miscellaneous/miscellaneous_ex14/miscellaneous_ex14.C
BalticPinguin: yes, I can do so but in fact, we use the shift-and-invert (STSINV); with this respect, the doc was wrong so I would change it and keep the version-error where it is, if it is okay for you.

jwpeterson:
> Silly question: why make it optional? Could we build one of {Test, Test Debug} with SLEPc and the other without?

Sure, that would be fine with me.

examples/miscellaneous/miscellaneous_ex14/miscellaneous_ex14.C
roystgnr: @jwpeterson has a point - we can probably get rid of the LESS_THAN(2,2,1) block entirely.

BalticPinguin: sure.

roystgnr:
I get build errors:
```
../../../../examples/miscellaneous/miscellaneous_ex14/miscellaneous_ex14.C:302:15: error: ‘imag’ was not declared in this scope
           if (imag(k)>0)
               ^~~~
```

And likewise for conj.  Explicitly using std::imag and std::conj fixes that for me.

More sneakily:
```
../../../../examples/miscellaneous/miscellaneous_ex14/miscellaneous_ex14.C: In function ‘int main(int, char**)’:                                      
../../../../examples/miscellaneous/miscellaneous_ex14/miscellaneous_ex14.C:288:54: error: expression list treated as compound expression in initialize
r [-fpermissive]                                                                                                                                      
       Number eigenvalue(eigpair.first, eigpair.second);                                                                                              
                                                      ^                                                                                               
../../../../examples/miscellaneous/miscellaneous_ex14/miscellaneous_ex14.C:305:66: error: cannot convert ‘std::complex<double>’ to ‘double’ in assignm
ent                                                                                                                                                   
             eq_sys.parameters.set<Number>("momentum")=std::conj(k);
```

Because naturally I didn't think to build with --enable-complex.  Switching a few variable declarations from Number to Complex fixes that.

8c15c5972f0048cdbdfccaedf0ae6d96497e7e12 in the roystgnr/example_Hydrogen branch fixes compilation entirely for me, so you can just cherry-pick that.

It turns out that I don't have a complex SLEPc module lying around to go along with my complex PETSc module, so it'll take a little while before I can test that configuration.

roystgnr:
Huh.  The automatic github hyperlink to that hash works, even though the URL says libMesh/libmesh but I only pushed to roystgnr/libmesh.  Snazzy.

roystgnr:
This doesn't work with out-of-source builds yet: GetPot fails to read sample_inp (which probably ought to be renamed something.in to match our other examples) when sample_inp is only in the srcdir but we're running from the build dir.

You could do the hokey .linkstamp trick we do in other examples... but IIRC we put that hack together before @pbauman discovered that autotools tests are run with the srcdir variable defined.

Just changing run.sh to use $srcdir/sample_inp gets the example running for me.  The example then *crashes* for me, but that appears to be a result of my --distributed-mesh configuration... and looking into it I believe you've exposed a bug in my code, not written one in your own.  I'll see about fixing that now.

roystgnr:
You renamed to sample.in but changed the run.sh name to sample.inp in e6750a9

BalticPinguin:
argh. Thanks.
I updated my source tree, so I didn't run a test-calculation.

roystgnr:
This appears to work fine for me on a serial DistributedMesh now.  If you'll verify that it works on a parallel ReplicatedMesh too then I think we'll be ready to merge once Civet is.

The parallel DistributedMesh failure appears to be due entirely to DistributedMesh bugs, at least the first of which was easy to fix, so I'll have a separate PR for those this afternoon.

BalticPinguin:
sorry for the stupid question; but how do I test it explicitly?  Do I set Mesh to ReplicatedMesh or use some option?

roystgnr:
Not a stupid question, but fortunately a simple one: unless you've explicitly configured with --enable-parmesh, your Mesh class is just a shim around ReplicatedMesh, so the only explicit thing you need to do for me is make sure you've run the test in parallel, e.g. with:
```
LIBMESH_RUN='mpiexec -n 4' make -C examples/miscellaneous/miscellaneous_ex14 check
```

BalticPinguin:
ok. Since I am running it on my pc, I usually don't use mpi and currently I update the sources; so you'll get an answer by next week .

roystgnr:
Okay, it sounds like you've soundly beaten me on the "how much of a hassle will that be to check" game; I'll test the parallel ReplicatedMesh case myself shortly.

Also: I just figured out why we have so many bugs in the parallel DistributedMesh case: I got lazy years ago, forced miscellaneous_ex1 to use ReplicatedMesh rather than fixing build_inf_elem to handle DistributedMesh, and then never got around to it again.  However, I didn't discover this until *after* multiple fixes, so I guess at this point I might as well try to complete the job.

BalticPinguin:
oh, this was not the intended message :-) But when I try to run some example, the complete library is reconfigured. (I think, my branches ran a bit too far apart and I should rather hold several source-trees.)  Thus I just need to wait this to get finished before I can run something...

roystgnr:
No worries, I just ran a parallel ReplicatedMesh case myself.  It's buggy in the QoI evaluation, where you try to evaluate the solution at indices which may be nonlocal.  Let me see if the fix is easy enough.

roystgnr:
Huh.  Looks like System::point_value() doesn't handle InfFE right now?  Let me try fixing that; there's no reason you should have 20 lines of code here instead of 1.

BalticPinguin:
I really can work further on it as well. 
I anyway wanted to get hands on the parallelisation (which doesn't work on my main project as well), so it might be a nice starting point for me.

roystgnr:
> I really can work further on it as well.
> I anyway wanted to get hands on the parallelisation (which doesn't work on my main project as well), so it might be a nice starting point for me.

Well, then, don't let me look a gift horse in the mouth!

I believe the "right" way to do things would be to go through system.C, find the three places where we use ```FEBase::build()``` in ```System::point_*()```, and add code paths which iff ```elem->infinite()``` use ```FEBase::build_InfFE()``` instead.  Then in the actual example code, you'd just have ```Number soln = system.point_value(0, q_point);```

roystgnr:
Thanks!  Hopefully it wasn't too discouraging that this took you so far down the rabbit hole before we could get it merged.

https://www.reddit.com/r/ProgrammerHumor/comments/54dmni/when_fixing_one_bug_leads_to_another/

BalticPinguin:
Not at all! I take it as education, so I'm fine with it :-) 
Thanks for merging.

PR #1480, opened 2017-11-01
dof_indices, old_dof_indices optimizations

roystgnr:
When I removed the subdivision-surface "pessimization" from dof_indices() a couple years ago, I apparently neglected to do the same in old_dof_indices(); this PR rectifies that.

We can also get a little bit of speedup by reducing the number of times we hit FEInterface switch statements: instead of using FEInterface at every node, we can use it once to get a function pointer, then reuse that function pointer on every node on the same FE/Elem combo.  This gives me about a 15% speedup even in the QUAD4 case, and ought to be even better for higher order and/or 3D elements.

I also get rid of redundant n_nodes() calls, which is probably a nearly-invisible performance improvement, but in functions which get called a billion times I figure every little bit helps.

### DIFF
--- a/include/fe/fe_interface.h
+++ b/include/fe/fe_interface.h
@@ -25,6 +25,7 @@
 #include "libmesh/enum_elem_type.h"
 #include "libmesh/vector_value.h"
 #include "libmesh/enum_fe_family.h"
+#include "libmesh/enum_order.h"
 
 // C++ includes
 #include <map>
@@ -108,6 +109,18 @@ public:
                                      const FEType & fe_t,
                                      const ElemType t,
                                      const unsigned int n);
+
+  typedef unsigned int (*n_dofs_at_node_ptr) (const ElemType,
+                                              const Order,
+                                              const unsigned int);
+
+  /**
+   * \returns A function which evaluates n_dofs_at_node for the
+   * requested FE type and dimension.
+   */
+  static n_dofs_at_node_ptr
+  n_dofs_at_node_function(const unsigned int dim,
+                          const FEType & fe_t);
 
   /**
    * \returns The number of dofs interior to the element,
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -1253,7 +1253,7 @@ private:
    * In DEBUG mode, the tot_size parameter will add up the total
    * number of dof indices that should have been added to di.
    */
-  void _dof_indices (const Elem * const elem,
+  void _dof_indices (const Elem & elem,
                      int p_level,
                      std::vector<dof_id_type> & di,
                      const unsigned int v,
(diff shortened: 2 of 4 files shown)


permcody:
This is awesome! BTW - @rwcarlsen is right on the verge of nailing down a more consistent way to handle our benchmarks. He's been messing around with turning off frequency scaling on modern CPUs to remove the variation that we see when we run the same executable over and over again. It looks like we are approaching a solution that will help us track our improvements in a more concrete way. With any luck we'll have this deployed very soon and we'll be able to look at a graph seeing improvements like this over time with select problems in our test suite.

roystgnr:
I've been using callgrind to get consistency, but that's tedious on small problems and impossible on large.  IIRC the deal.II people use "run three times and take the fastest timing", which makes some sense but is still too expensive to put in CI.  If the true solution turns out to be something as simple as ```echo 1 > /proc/cpu/go_fast``` then I'll be dumbfounded.

You could look at trendlines rather than just runs at a single git hash, to smooth out noise, but that only works with a very stable performance testing suite; every time you add a new test you're comparing apples to oranges.

roystgnr:
I'll merge this tonight unless @jwpeterson objects to the "function returning function pointer" design.  The only alternative I saw would have been "vectorized n_dofs_at_nodes()", but since that would have added more vector allocation/deallocation I'm not sure whether it would have ended up faster or slower in the end.

include/fe/fe_interface.h
```diff
@@ -109,6 +110,18 @@ class FEInterface
                                      const ElemType t,
                                      const unsigned int n);
 
+  typedef unsigned int (*n_dofs_at_node_ptr) (const ElemType,
```
jwpeterson: TODO for later maybe: there's a C++11 `using` syntax that improves the typedef'ing function pointers IMO. In your case it would read:
```
using n_dofs_at_node_ptr = unsigned int (*) (const ElemType, const Order, const unsigned int);
```
I don't think we have a configure test for it yet but it should be simple to add one.

permcody: I've been using that syntax all over the place in MOOSE since we first switched to C++11. It's well supported. You probably don't need a configure test for every little piece of syntax. Yes there are still problems with things like `<regex>` but the syntax seems ok.

roystgnr: Looks no better to me (although I do like the similar ```using foo = decltype(example_function);``` alternative), but we can use using if you prefer.

http://en.cppreference.com/w/cpp/compiler_support and https://gcc.gnu.org/projects/cxx-status.html claim that template aliases (which ought to be a dependant on other using declarations) went into gcc 4.7, which actually is later than any of the other features we specifically test for.  I'm not sure whether they got non-templated using declarations in first... but it looks like the whole thing was defined in the same [N2258](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2007/n2258.pdf) proposal, so I wouldn't bet on it.

On the gripping hand, I'm not sure if we *want* to do extra m4 work to help gcc 4.6 users decide it's finally time to upgrade; maybe they can figure that out on their own?

PR #1481, opened 2017-11-01
Create add_scaled method for TypeNTensor

lindsayad:
This is mostly for just getting some general code in MOOSE to compile...I have no plans to actually use this at present...

### DIFF
--- a/include/numerics/type_n_tensor.h
+++ b/include/numerics/type_n_tensor.h
@@ -27,6 +27,7 @@
 // C++ includes
 #include <cstdlib> // *must* precede <cmath> for proper std:abs() on PGI, Sun Studio CC
 #include <cmath>
+#include <vector>
 
 namespace libMesh
 {
@@ -46,13 +47,13 @@ template <unsigned int N, typename T>
 class TypeNTensor
 {
 public:
-  TypeNTensor () {}
+  TypeNTensor () : _coords(std::vector<T>(int_pow(LIBMESH_DIM, N))) {}
 
-  TypeNTensor (const T &) {}
+  TypeNTensor (const T &) : _coords(std::vector<T>(int_pow(LIBMESH_DIM, N))) {}
 
-  TypeNTensor (const TypeVector<T> &) {}
+  TypeNTensor (const TypeVector<T> &) : _coords(std::vector<T>(int_pow(LIBMESH_DIM, N))) {}
 
-  TypeNTensor (const TypeTensor<T> &) {}
+  TypeNTensor (const TypeTensor<T> &) : _coords(std::vector<T>(int_pow(LIBMESH_DIM, N))) {}
 
   operator TypeVector<T> () const { libmesh_not_implemented(); return 0; }
   operator VectorValue<T> () const { libmesh_not_implemented(); return 0; }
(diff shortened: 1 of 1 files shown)


include/numerics/type_n_tensor.h
```diff
@@ -205,9 +217,37 @@ class TypeNTensor
...
+  template<typename T2>
+  void add_scaled (const TypeNTensor<N, T2> &, const T &);
+
+  /**
+   * The coordinates of the \p TypeNTensor
+   */
+  std::vector<T> _coords;
+
+private:
+  static int int_pow(int b, int e)
```
roystgnr: There's probably some constexpr tricks that we ought to do to make sure that the compiler knows not to actually evaluate this every time it's invoked.  But that can wait until someone actually uses it.

roystgnr:
:+1: when Civet is done.

include/numerics/type_n_tensor.h
```diff
@@ -46,13 +47,25 @@ template <unsigned int N, typename T>
...
+
+  TypeNTensor (const TypeVector<T> &)
+    {
+      _coords.resize(int_pow(LIBMESH_DIM, N));
+    }
+
+  TypeNTensor (const TypeTensor<T> &)
+    {
+      _coords.resize(int_pow(LIBMESH_DIM, N));
+    }
```
jwpeterson: Was this a bug? How were we not resizing these before? Shouldn't this be done in the initializer lists?

lindsayad:
Looks like we have some problems with running complex tests in parallel

include/numerics/type_n_tensor.h
jwpeterson: Oh I see, you just added `_coords`. So I guess this class wasn't used for anything previously?

roystgnr:
IIRC it was a shim added when @pbauman added vector-valued elements.

Maybe for eventual second-derivative support, which never got added?

jwpeterson:
> Looks like we have some problems with running complex tests in parallel

Ah yes, we didn't want to mess with having a complex-valued PETSc available, so our complex builds fall back on Eigen and therefore can't be run in parallel. This CIVET test was originally just created to test that compilation with complex numbers wasn't broken...

brianmoose:
I removed running check in parallel on "Test enable complex" and added a parallel check on Test (in addition to the serial check).

brianmoose:
It looks like running the unit tests in parallel doesn't work? Or is it only a civet issue?
It ran ~12 hours before civet automatically killed it: https://www.moosebuild.org/job/130559/
Should I remove the parallel check in opt mode?
@lindsayad Sorry for making this PR the civet testing ground.

roystgnr:
Running unit tests in parallel works for me (at least, it does ever since I fixed that new unit test in parallel).

Oh...  But it looks like I didn't fix that new unit test in parallel, I left it with a race condition.  Let me try fixing that now.

roystgnr:
Oh, but that thing where we exit with an error but MPI_Abort doesn't actually abort and we end up spinning our wheels for 12 hours until Civet times out?  I don't know how to fix that.  IIRC on my systems it made the MOOSE expected-error tests unusable for me until I switched back from openmpi to mpich2.  I can't understand why it would be affecting Civet, though, since obviously you've got an MPI stack that handles all your own tests' MPI_Aborts correctly.

jwpeterson:
> I can't understand why it would be affecting Civet, though

I'm willing to blame it on cppunit doing something screwy with exceptions, error handlers, etc.

lindsayad:
Tests pass!

PR #1482, opened 2017-11-02
Fix PerfLog::get_perf_data()

roystgnr:
We can no longer directly look up strings in the log.  We'll have to
search for the corresponding character pointers first.

Thanks to @jessecarterMOOSE for finding the regression in #1462.  I haven't tested this for correctness yet, I'm rebuilding MOOSE to do so now, but I just wanted to let him have a chance to do the same ASAP.

This uses a C++11 lambda function; if it turns out to be the right fix then I'll add a second commit (and a third bootstrap commit) to make lambda support mandatory at configure time before we merge.

### DIFF
--- a/src/systems/fem_system.C
+++ b/src/systems/fem_system.C
@@ -854,6 +854,9 @@ void FEMSystem::assembly (bool get_residual, bool get_jacobian,
                           bool apply_no_constraints)
 {
   libmesh_assert(get_residual || get_jacobian);
+
+// Log residual and jacobian and combined performance separately
+#ifdef LIBMESH_ENABLE_PERFORMANCE_LOGGING
   const char * log_name;
   if (get_residual && get_jacobian)
     log_name = "assembly()";
@@ -863,6 +866,7 @@ void FEMSystem::assembly (bool get_residual, bool get_jacobian,
     log_name = "assembly(get_jacobian)";
 
   LOG_SCOPE(log_name, "FEMSystem");
+#endif
 
   const MeshBase & mesh = this->get_mesh();
 
--- a/src/utils/perf_log.C
+++ b/src/utils/perf_log.C
@@ -703,9 +703,11 @@ PerfData PerfLog::get_perf_data(const std::string & label, const std::string & h
       return log[std::make_pair(header_c_str, label_c_str)];
     }
 
+  typedef decltype(*log.begin()) map_pair;
+
   auto iter = std::find_if
     (log.begin(), log.end(),
-     [&label, &header] (const auto & a)
+     [&label, &header] (const map_pair & a)
      { return !std::strcmp(header.c_str(), a.first.first) &&
               !std::strcmp(label.c_str(), a.first.second); });
 
--- a/src/utils/perf_log.C
+++ b/src/utils/perf_log.C
@@ -18,6 +18,7 @@
 #include "libmesh/perf_log.h"
 
 // C++ includes
+#include <algorithm>
 #include <iostream>
 #include <iomanip>
 #include <cstring>
@@ -694,7 +695,23 @@ void PerfLog::print_log() const
 
 PerfData PerfLog::get_perf_data(const std::string & label, const std::string & header)
 {
-  return log[std::make_pair(header.c_str(), label.c_str())];
+  if (non_temporary_strings.count(label) &&
+      non_temporary_strings.count(header))
+    {
+      const char * label_c_str = non_temporary_strings[label];
+      const char * header_c_str = non_temporary_strings[header];
+      return log[std::make_pair(header_c_str, label_c_str)];
+    }
+
+  auto iter = std::find_if
+    (log.begin(), log.end(),
+     [&label, &header] (const auto & a)
+     { return !std::strcmp(header.c_str(), a.first.first) &&
+              !std::strcmp(label.c_str(), a.first.second); });
+
+  libmesh_assert(iter != log.end());
+
+  return iter->second;
 }
 
 void PerfLog::start_event(const std::string & label,


jwpeterson:
> I'll add a second commit (and a third bootstrap commit) to make lambda support mandatory at configure time

Looks like we already require that C++11 compilers support this empty lambda syntax (ax_cxx_compile_stdcxx_11.m4) 
```
auto l = [](){};
```
So I'd say we're good or you can just generalize the existing test slightly.  We also have `LIBMESH_TEST_CXX11_LAMBDA` which tests that a lambda object can be passed to a function expecting a traditional function pointer, can't remember what I originally had in mind for that one.

roystgnr:
Huh, okay; I'll leave configure as is, then.

roystgnr:
This doesn't seem to fix https://github.com/idaholab/moose/pull/10197 for me...

jwpeterson:
:+1: I'd like to merge this today. We haven't had a libmesh update since Sep 8, so it's probably time.

jwpeterson:
Oh, whoops, just saw your message. I will hold off merging then.

jessecarterMOOSE:
> This doesn't seem to fix idaholab/moose#10197 for me...

Maybe it's a bad test. What does the output look like? If it's no longer just a bunch of zeros then your fix probably worked and we need to come up with a better way of checking the perf log postprocessor.

roystgnr:
```
Postprocessor Values:
+----------------+----------------+----------------+------------------+---------------------------+----------------+----------------------------+-------------------------------------+----------------+-------------------------+----------------+
| time           | elapsed_active | elapsed_alive  | jac_average_time | jac_average_time_with_sub | jac_calls      | jac_percent_of_active_time | jac_percent_of_active_time_with_sub | jac_total_time | jac_total_time_with_sub | res_calls      |
+----------------+----------------+----------------+------------------+---------------------------+----------------+----------------------------+-------------------------------------+----------------+-------------------------+----------------+
|   0.000000e+00 |   0.000000e+00 |   0.000000e+00 |     0.000000e+00 |              0.000000e+00 |   0.000000e+00 |               0.000000e+00 |                        0.000000e+00 |   0.000000e+00 |            0.000000e+00 |   0.000000e+00 |
|   1.000000e+00 |   5.524200e-01 |   5.992240e-01 |     0.000000e+00 |              0.000000e+00 |   0.000000e+00 |               0.000000e+00 |                        0.000000e+00 |   0.000000e+00 |            0.000000e+00 |   0.000000e+00 |
+----------------+----------------+----------------+------------------+---------------------------+----------------+----------------------------+-------------------------------------+----------------+-------------------------+----------------+
```

So, not *just* a bunch of zeros, but the important bits are a bunch of zeros, correct?

jessecarterMOOSE:
Right `elapsed_active` and `elapsed_alive` come from `PerfLog::get_active_time()` and `PerfLog::get_elapsed_time()` directly and don't use `get_perf_data()`. The rest do.

permcody:
I suppose that we could check to make sure that those values in this test are never exactly zero. That's probably still a terrible metric but could catch this case and similar errors where the logging just wasn't working.

jessecarterMOOSE:
I'm trying to re-compile libmesh **on falcon** after applying this patch to test it myself and I get first a warning and then an error about the lamda function...
```
../src/systems/fem_system.C: In member function ‘virtual void libMesh::FEMSystem::assembly(bool, bool, bool, bool)’:
../src/systems/fem_system.C:857:16: warning: variable ‘log_name’ set but not used [-Wunused-but-set-variable]
   const char * log_name;
                ^
...

../src/utils/perf_log.C: In member function ‘libMesh::PerfData libMesh::PerfLog::get_perf_data(const string&, const string&)’:
../src/utils/perf_log.C:708:25: warning: use of ‘auto’ in lambda parameter declaration only available with -std=c++1y or -std=gnu++1y
      [&label, &header] (const auto & a)
                         ^
In file included from /apps/local/easybuild/software/GCC/4.9.2/include/c++/4.9.2/bits/stl_algobase.h:71:0,
                 from /apps/local/easybuild/software/GCC/4.9.2/include/c++/4.9.2/bits/char_traits.h:39,
                 from /apps/local/easybuild/software/GCC/4.9.2/include/c++/4.9.2/ios:40,
                 from /apps/local/easybuild/software/GCC/4.9.2/include/c++/4.9.2/istream:38,
                 from /apps/local/easybuild/software/GCC/4.9.2/include/c++/4.9.2/sstream:38,
                 from /apps/local/easybuild/software/GCC/4.9.2/include/c++/4.9.2/complex:45,
                 from ./include/libmesh/libmesh_common.h:46,
                 from ./include/libmesh/perf_log.h:25,
                 from ../src/utils/perf_log.C:18:
/apps/local/easybuild/software/GCC/4.9.2/include/c++/4.9.2/bits/predefined_ops.h: In instantiation of ‘bool __gnu_cxx::__ops::_Iter_pred<_Predicate>::operator()(_Iterator) [with _Iterator = __gnu_debug::_Safe_iterator<std::_Rb_tree_iterator<std::pair<const std::pair<const char*, const char*>, libMesh::PerfData> >, std::__debug::map<std::pair<const char*, const char*>, libMesh::PerfData> >; _Predicate = libMesh::PerfLog::get_perf_data(const string&, const string&)::<lambda(const auto:1&)>]’:
/apps/local/easybuild/software/GCC/4.9.2/include/c++/4.9.2/bits/stl_algo.h:104:50:   required from ‘_InputIterator std::__find_if(_InputIterator, _InputIterator, _Predicate, std::input_iterator_tag) [with _InputIterator = __gnu_debug::_Safe_iterator<std::_Rb_tree_iterator<std::pair<const std::pair<const char*, const char*>, libMesh::PerfData> >, std::__debug::map<std::pair<const char*, const char*>, libMesh::PerfData> >; _Predicate = __gnu_cxx::__ops::_Iter_pred<libMesh::PerfLog::get_perf_data(const string&, const string&)::<lambda(const auto:1&)> >]’
/apps/local/easybuild/software/GCC/4.9.2/include/c++/4.9.2/bits/stl_algo.h:162:43:   required from ‘_Iterator std::__find_if(_Iterator, _Iterator, _Predicate) [with _Iterator = __gnu_debug::_Safe_iterator<std::_Rb_tree_iterator<std::pair<const std::pair<const char*, const char*>, libMesh::PerfData> >, std::__debug::map<std::pair<const char*, const char*>, libMesh::PerfData> >; _Predicate = __gnu_cxx::__ops::_Iter_pred<libMesh::PerfLog::get_perf_data(const string&, const string&)::<lambda(const auto:1&)> >]’
/apps/local/easybuild/software/GCC/4.9.2/include/c++/4.9.2/bits/stl_algo.h:3804:45:   required from ‘_IIter std::find_if(_IIter, _IIter, _Predicate) [with _IIter = __gnu_debug::_Safe_iterator<std::_Rb_tree_iterator<std::pair<const std::pair<const char*, const char*>, libMesh::PerfData> >, std::__debug::map<std::pair<const char*, const char*>, libMesh::PerfData> >; _Predicate = libMesh::PerfLog::get_perf_data(const string&, const string&)::<lambda(const auto:1&)>]’
../src/utils/perf_log.C:710:61:   required from here
/apps/local/easybuild/software/GCC/4.9.2/include/c++/4.9.2/bits/predefined_ops.h:231:30: error: invalid use of ‘auto’
  { return bool(_M_pred(*__it)); }
```

permcody:
Lol - @roystgnr are you ready to go for C++17 already? That didn't take long ;)

roystgnr:
Hmm... that warning would show up without --enable-perflog, wouldn't it?  I'll put ```#ifdef``` around that variable.

And that error would show up on any C++11-only compiler, huh?  Apparently the trouble is that I'm using gcc 7.2, and gcc 7 defaults to --std=c++14, and I don't go out of my way to change that.  Presumably the same thing is happening in CI; we should probably force Civet to add --std=c++11 the way we used to with a --std=c++98 test.  Probably not a separate test this time, though, just tack it on to Test Debug?

permcody:
Oh, wait, that's actually C++14. The joys of several in-flight standards at once.

jwpeterson:
> we should probably force Civet to add --std=c++11 the way we used to with a --std=c++98 test.

`--enable-cxx11` used to do exactly this. I think you changed it so that it either doesn't run or doesn't do set any std flags. The most recent CIVET run reports the following CXXFLAGS in the summary:

```
-O2 -felide-constructors -funroll-loops -fstrict-aliasing -Wdisabled-optimization -fopenmp
```

jwpeterson:
I had forgotten about [generic lambdas](http://en.cppreference.com/w/cpp/language/lambda).

The list of parameters, as in named functions, except that default arguments are not allowed (until C++14). If auto is used as a type of a parameter, the lambda is a generic lambda. (since C++14)

roystgnr:
Hmm.. I don't recall that change but it definitely does sound like the sort of thing I'd do.  I was working with a few libMesh apps that either required C++14 to compile or preferred C++14 for performance, but that also did our usual "swipe CXXFLAGS from libMesh" trick for compatibility, and I would have been dumbfounded and annoyed if I ever turned on --enable-cxx11 and discovered it was a *downgrade*.

We can just pass an arbitrary ```libmesh_CXXFLAGS='-std=whatever'``` to configure, can't we?

jwpeterson:
> We can just pass an arbitrary libmesh_CXXFLAGS='-std=whatever' to configure, can't we?

We can do this on libmesh PRs, yeah. I'm not confident that compilers are good at explicitly disallowing things from later standards (clang at least seems to be lax about this) but it would certainly be better than nothing. I'll look into it.

roystgnr:
Okay, I swear it wasn't working before, but it works now... I may have just forgotten "make install"?

Looking forward to [the MOOSE version upgrade](https://github.com/idaholab/moose/pull/10176/files/8b532b33680fee33c5cb1f035713f337cd3a567d#diff-532a42baadabb3c76dfc0380633f424aR37).

jessecarterMOOSE:
I just want to say that it's a pleasure working with you guys. Thanks for being so responsive!

roystgnr:
> I just want to say that it's a pleasure working with you guys. Thanks for being so responsive!

You're welcome.  But this was a special case, after all.  "Adding feature X would be useful" is easy to be lazy about, but "When you added new feature Y it regressed correctly-written code" is terrifyingly motivating.  At that point the only alternative to "I'll fix it immediately" would be a tearful "I'll revert it and watch all my clever work get undone and wasted".

PR #1484, opened 2017-11-02
Add MeshBase::local_node_ptr_range()

jwpeterson:
Turned out to be a fair number of places where we could replace `mesh.local_nodes_begin/end()` calls with range objects.

### DIFF
--- a/examples/miscellaneous/miscellaneous_ex1/miscellaneous_ex1.C
+++ b/examples/miscellaneous/miscellaneous_ex1/miscellaneous_ex1.C
@@ -467,28 +467,16 @@ void assemble_wave(EquationSystems & es,
 
   // Note that we have not applied any boundary conditions so far.
   // Here we apply a unit load at the node located at (0,0,0).
-  {
-    // Iterate over local nodes
-    MeshBase::const_node_iterator           nd = mesh.local_nodes_begin();
-    const MeshBase::const_node_iterator nd_end = mesh.local_nodes_end();
-
-    for (; nd != nd_end; ++nd)
+  for (const auto & node : mesh.local_node_ptr_range())
+    if (std::abs(node(0)) < TOLERANCE &&
+        std::abs(node(1)) < TOLERANCE &&
+        std::abs(node(2)) < TOLERANCE)
       {
-        // Get a reference to the current node.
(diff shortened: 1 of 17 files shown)


examples/miscellaneous/miscellaneous_ex1/miscellaneous_ex1.C
```diff
@@ -467,28 +467,16 @@ void assemble_wave(EquationSystems & es,
...
-  {
-    // Iterate over local nodes
-    MeshBase::const_node_iterator           nd = mesh.local_nodes_begin();
-    const MeshBase::const_node_iterator nd_end = mesh.local_nodes_end();
-
-    for (; nd != nd_end; ++nd)
+  for (const auto & node : mesh.local_node_ptr_range())
+    if (std::abs(node(0)) < TOLERANCE &&
+        std::abs(node(1)) < TOLERANCE &&
+        std::abs(node(2)) < TOLERANCE)
```
roystgnr: Wait, this can't be right.  You've changed node from a reference to a pointer, so we need to dereference it now before calling ```operator()()``` or ```dof_number()```.

jwpeterson: Why does `node(0)` even compile?

roystgnr: It doesn't.  This is an ENABLE_INFINITE_ELEMENTS-only example.

roystgnr:
No other obvious mistakes; :+1: after the miscellaneous_ex1 code is fixed.

Although, it *did* just become obvious to me that we need to add --enable-ifem to at least one of our regular CI tests.

jwpeterson:
> Although, it did just become obvious to me that we need to add --enable-ifem to at least one of our regular CI tests

There is an optional one which I just attached. We could make it run on every PR if there is going to be more frequent development on ifems.

roystgnr:
We've got current ifem users, so we should definitely have something running on every PR that might hit them, which may include more PRs than not while we're in the middle of swaths of optimization and C++11 style changes.  I assume Test Infinite Elements is relatively cheap to run since it doesn't have to touch MOOSE?

Issue #1472, opened 2017-10-31
continuity of elements


BalticPinguin:
Hi,
I had finally some desire to look, why in my installation the adaptivity_ex3 failed (when running with hp-refinement) and stubled over the issue that I have disabled second derivatives (they are not available for infinite elements) but the  `fe->get_continuity()` (see `src/error_estimation/hp_coarsentest.C`, line 599) tells that the continuity is C_ONE and, even if d2phi is not computed, we try to access it.

Since I am not into the refinement/coarsening nor elements, I don't know how to solve it best. But I think, one should
  - add respective precompiler-flags throughout the AMR-functions or
  - ensure that the continuity is not 1 if no second derivatives are available.
Probably the first one is the more reasonable approach?

What do you think about it?

roystgnr:
Hmm... that's actually an interesting question, but I don't like either of your solutions.  Changing AMR behavior depending on --enable-second-derivatives ought to be a no-no, that's grossly surprising behavior.

We could avoid running adaptivity_ex3 with HERMITE (the only C1 element we try it with) if second derivatives are disabled, I suppose.  That's probably the correct quick fix, if you want to take a crack at it... maybe just grep for "LIBMESH_ENABLE_SECOND_DERIVATIVES 1" in libmesh_config.h, and don't run the HERMITE cases unless the grep succeeded?

But in the long run... IMHO the HPCoarsenTest behavior here is just wrong.  We're trying to minimize H2 error when we see C1 elements, and we're doing so regardless of whether or not we're actually solving an H2 problem!  adaptivity_ex3 is H1 regardless of whether we use a more-continuous-than-necessary basis or not.

Maybe that's just an academic question, since nobody uses HERMITE elements for H1 problems in the real world, and since the HPCoarsenTest isn't a great selector to begin with?  I'd still like us to get it *right*... but I'm not sure what the right way to enable that is.  Default to continuity-based norm, but add a ```HPCoarsenTest::set_system_norm(SystemNorm sn)``` method with which users can override that?  That would let users do multiphysics with mixed spaces or dimensional norms properly too.

Okay, I think that's the right thing to do, but I don't think I want to bother with it unless/until there's real user interest.  Until then disabling the HERMITE tests when second derivatives aren't supported ought to be adequate.

jwpeterson:
> adaptivity_ex3 is H1 regardless of whether we use a more-continuous-than-necessary basis or not.

[...]

PR #1485, opened 2017-11-02
Prevent adaptivity_ex3 from being run in invalid configurations.

jwpeterson:
Refs #1472.

### DIFF
--- a/examples/adaptivity/adaptivity_ex3/adaptivity_ex3.C
+++ b/examples/adaptivity/adaptivity_ex3/adaptivity_ex3.C
@@ -153,6 +153,13 @@ int main(int argc, char ** argv)
   const bool output_intermediate    = input_file("output_intermediate", false);
 #endif
 
+  // If libmesh is configured without second derivative support, we
+  // can't run this example with Hermite elements and will therefore
+  // fail gracefully.
+#if !defined(LIBMESH_ENABLE_SECOND_DERIVATIVES)
+  libmesh_example_requires(approx_type != "HERMITE", "--enable-second");
+#endif
+
   dim = input_file("dimension", 2);
   const std::string indicator_type = input_file("indicator_type", "kelly");
   singularity = input_file("singularity", true);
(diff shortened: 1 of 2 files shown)


jwpeterson:
This PR is being tested with `libmesh_CXXFLAGS = -std=c++11` in the environment.

The failure is during compilation of the unit tests:
```
../../tests/base/unique_ptr_test.C: In member function 'void UniquePtrTest::testMakeUnique()':
../../tests/base/unique_ptr_test.C:99:32: error: 'make_unique' is not a member of 'std'
     std::unique_ptr<int> foo = libmesh_make_unique<int>(42);
                                ^~~
../../tests/base/unique_ptr_test.C:99:49: error: expected primary-expression before 'int'
     std::unique_ptr<int> foo = libmesh_make_unique<int>(42);
                                                 ^~~
```
but I'm not sure why it fails, because the configure test of `std::make_unique` passes. I'll need to check it out on a system similar to the build box to figure out what's going on.

BalticPinguin:
what do you think about adding 
```--- a/src/error_estimation/hp_coarsentest.C
+++ b/src/error_estimation/hp_coarsentest.C
@@ -236,12 +236,6 @@ void HPCoarsenTest::select_refinement (System & system)
           d2phi = &(fe->get_d2phi());
           d2phi_coarse = &(fe_coarse->get_d2phi());
         }
-#else
-      if (cont == C_ONE)
-        {
-           libmesh_error_msg("Minimisation of H2 error without second derivatives impossible.");
-           libmesh_not_implemented();
-        }
 #endif // defined (LIBMESH_ENABLE_SECOND_DERIVATIVES)

       // Iterate over all the active elements in the mesh
                                                                      ` ` `

as well? So we can prevent users from running into this issue with their own code...

roystgnr:
The libmesh_not_implemented() in your suggestion would never get reached.  Otherwise, that does look like a good idea.  It won't prevent anyone from running into the issue but it would make it much easier for them to debug if they do.

BalticPinguin:
oh, you are right :-) Than probably the error message is better to keep...

PR #1488, opened 2017-11-05
Added infinite element support for System::point_value().

BalticPinguin:
In this pr I addressed the improvements @roystgnr suggested in #1479.
It is a separate pr because I think, this could lead to some discussions:
The changes in `System::point_value()` should work and I think, they correspond to what you expected, but I came across some further issues:

- in the FEInterface I am not sure whether there is a bug or if I don't understand how the code works?
- The `System::point_hessian()` is not implemented for infinite elements and currently there is no reason to implement second derivatives for infinite elements.
- The `System::point_gradient()` could in principle be generalised, but would require to add gradients to the `FEInterface` / `FEComputeData` as well.

Probably I will fix the latter at some point because I might want gradients at some point, but currently it is probably not necessary...

### DIFF
--- a/src/fe/fe_interface.C
+++ b/src/fe/fe_interface.C
@@ -678,9 +678,10 @@ void FEInterface::shape<Real>(const unsigned int dim,
 #ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
 
   if (is_InfFE_elem(t))
-    phi = ifem_shape(dim, fe_t, t, i, p);
-
-  // here we should need an 'or'-statement, right? Else, we will run in trouble!?
+    {
+      phi = ifem_shape(dim, fe_t, t, i, p);
+      return;
+    }
 
 #endif
 
@@ -718,8 +719,10 @@ void FEInterface::shape<Real>(const unsigned int dim,
 #ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
 
   if (elem && is_InfFE_elem(elem->type()))
-    phi = ifem_shape(dim, fe_t, elem, i, p);
-  // here we should need an 'or'-statement, right? Else, we will run in trouble!?
+    {
+      phi = ifem_shape(dim, fe_t, elem, i, p);
+      return;
+    }
 #endif
 
   const Order o = fe_t.order;
--- a/src/fe/fe_interface.C
+++ b/src/fe/fe_interface.C
@@ -680,6 +680,8 @@ void FEInterface::shape<Real>(const unsigned int dim,
   if (is_InfFE_elem(t))
     phi = ifem_shape(dim, fe_t, t, i, p);
 
+  // here we should need an 'or'-statement, right? Else, we will run in trouble!?
+
 #endif
 
   const Order o = fe_t.order;
@@ -717,7 +719,7 @@ void FEInterface::shape<Real>(const unsigned int dim,
 
   if (elem && is_InfFE_elem(elem->type()))
     phi = ifem_shape(dim, fe_t, elem, i, p);
-
+  // here we should need an 'or'-statement, right? Else, we will run in trouble!?
 #endif
 
   const Order o = fe_t.order;
@@ -751,6 +753,7 @@ void FEInterface::shape<RealGradient>(const unsigned int dim,
                                       const Point & p,
                                       RealGradient & phi)
 {
+   // This even does not handle infinite elements at all!?
   const Order o = fe_t.order;
 
   switch(dim)
--- a/src/fe/inf_fe_static.C
+++ b/src/fe/inf_fe_static.C
@@ -178,6 +178,8 @@ Real InfFE<Dim,T_radial,T_map>::shape(const FEType & fet,
   compute_shape_indices(fet, inf_elem_type, i, i_base, i_radial);
 
   //TODO:[SP/DD]  exp(ikr) is still missing here!
+  // but is it intended?  It would be probably somehow nice, but than it would be Number, not Real !
+  // --> thus it would destroy the interface...
   if (Dim > 1)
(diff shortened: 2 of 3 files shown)


src/fe/fe_interface.C
```diff
@@ -680,6 +680,8 @@ void FEInterface::shape<Real>(const unsigned int dim,
   if (is_InfFE_elem(t))
     phi = ifem_shape(dim, fe_t, t, i, p);
 
+  // here we should need an 'or'-statement, right? Else, we will run in trouble!?
+
```
roystgnr: Braces and a "return", I'd say.

This function IIRC comes from @pbauman's refactoring for vector-valued elements, which was kind of in the Dark Ages for infinite elements.  I would be ashamed but unsurprised if we failed to get infinite element support correct here and nobody realized it.

src/fe/inf_fe_static.C
```diff
@@ -178,6 +178,8 @@ Real InfFE<Dim,T_radial,T_map>::shape(const FEType & fet,
   compute_shape_indices(fet, inf_elem_type, i, i_base, i_radial);
 
   //TODO:[SP/DD]  exp(ikr) is still missing here!
+  // but is it intended?  It would be probably somehow nice, but than it would be Number, not Real !
+  // --> thus it would destroy the interface...
```
roystgnr: This I couldn't tell you about... but I was under the strong impression that infinite elements were supposed to work even without --enable-complex!

src/systems/system.C
```diff
@@ -1462,6 +1464,15 @@ Real System::calculate_norm(const NumericVector<Number> & v,
       return v_norm;
     }
 
+#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+
+  // One way for implementing this would be to exchange the fe with the FEInterface- class.
+  // However, it needs to be discussed whether integral-norms make sense for infinite elements.
+  // or in which sense they could make sense.
+  libmesh_not_implemented();
```
roystgnr: Oh, absolutely *not*.  IIRC I always enable infinite elements at configure time indirectly via --enable-everything, and that doesn't mean I want to break any code that tries to calculate a norm on a mesh which doesn't include them.

L-infinity norm at least ought to make sense with infinite elements in the mesh, right?  But even if you don't want to implement that, at least move the ```libmesh_not_implemented()``` into the element loop where you can check ```elem->infinite()``` first.

src/systems/system.C
```diff
@@ -2096,25 +2107,20 @@ Number System::point_value(unsigned int var, const Point & p, const Elem & e) co
...
+  // Map the physical co-ordinates to the master co-ordinates using the inverse_map from fe_interface.h.
+  Point coor = FEInterface::inverse_map(e.dim(), fe_type, &e, p);
 
-  // Map the physical co-ordinates to the master co-ordinates using the inverse_map from fe_interface.h
-  // Build a vector of point co-ordinates to send to reinit
-  std::vector<Point> coor(1, FEInterface::inverse_map(e.dim(), fe_type, &e, p));
-
-  // Get the shape function values
-  const std::vector<std::vector<Real>> & phi = fe->get_phi();
+  // create this object to handle the correct
```
roystgnr: I think you might be missing a

src/systems/system.C
```diff
@@ -2096,25 +2107,20 @@ Number System::point_value(unsigned int var, const Point & p, const Elem & e) co
...
+  FEComputeData fe_data(this->get_equation_systems(), coor);
 
-  // Reinitialize the element and compute the shape function values at coor
-  fe->reinit (&e, &coor);
+  FEInterface::compute_data(e.dim(), fe_type, &e, fe_data);
```
roystgnr: I'm about to be AFK, but remind me to check and make sure this works correctly with p refinement.  I had to change a lot of code when I added adaptive p refinement support, and I don't remember changing the FEComputeData stuff.  Maybe it was just so long ago that I've forgotten, but maybe I missed it...

src/fe/inf_fe_static.C
BalticPinguin: ... I will check that. 
But if we use complex numbers, the 'full' shape of infinite elements include a complex oscillation, so it becomes complex.

src/systems/system.C
BalticPinguin: yes, you are absolutely right.

src/systems/system.C
BalticPinguin: ... I think, the important part is removed by markdown?

src/systems/system.C
roystgnr: Okay - it looks like I added adaptive p refinement support to the FE codepath of compute_data(), but didn't feel safe touching the InfFE code path.  So there's definitely still work to be done there at some future date, but this change won't regress any existing p-refinement AMR codes, which is good enough for now.

src/systems/system.C
roystgnr: Sorry, I was trying to make a joke.  I say "I think you might be missing a word", but then I omit a word myself, and everyone laughs at the irony?

Anyway, I'm afraid I didn't understand the comment here: I assumed you were typing "to handle the correct shape functions for both FE and InfFE" or some such, but it got cut off somehow.

BalticPinguin: oh, I did not only miss the joke but also the fact that this comment is indeed senseless.

roystgnr:
I'll merge later this afternoon unless Civet or anyone else screams first.

roystgnr:
You're planning to rebase #1479 on this and then use it for the point evaluations, right?

BalticPinguin:
Yes, I do.
Should I rebase the commits into one than?

roystgnr:
Nah, don't merge any commits, no need to even interactive rebase.  Just rebase off of this branch, then add the last commit with the point_value() use

Issue #1489, opened 2017-11-06
char* optimized PerfLog may not be Valgrind clean


jwpeterson:
See also [here](https://www.moosebuild.org/job/131580/). 

A summary of the Valgrind output from that job is below:
```
4 errors in context 1 of 2:
Mismatched free() / delete / delete []
   at 0x4C2F24B: operator delete(void*) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
   by 0x75C029D: libMesh::PerfLog::~PerfLog() (perf_log.C:72)
   by 0xF47C369: __cxa_finalize (cxa_finalize.c:56)
   by 0x5892432: ??? (in /opt/civet/build_0/moose/framework/libmoose-oprof.so.0.0.0)
   by 0x4010DE6: _dl_fini (dl-fini.c:235)
   by 0xF47BFF7: __run_exit_handlers (exit.c:82)
   by 0xF47C044: exit (exit.c:104)
   by 0xF462836: (below main) (libc-start.c:325)
 Address 0x136513f0 is 0 bytes inside a block of size 23 alloc'd
   at 0x4C2E80F: operator new[](unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
   by 0x75BC6EC: libMesh::PerfLog::push(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (perf_log.C:110)
   by 0x75BCD85: libMesh::PerfLog::push(char const*, char const*) (perf_log.C:135)
   by 0x60516BC: NonlinearSystemBase::init() (NonlinearSystemBase.C:159)
   by 0x61A82C1: FEProblemBase::init() (FEProblemBase.C:3727)
   by 0x5946CD4: ActionWarehouse::executeActionsWithAction(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (ActionWarehouse.C:361)
   by 0x594845C: ActionWarehouse::executeAllActions() (ActionWarehouse.C:336)
   by 0x611DE63: MooseApp::runInputFile() (MooseApp.C:555)
   by 0x611B6A2: MooseApp::run() (MooseApp.C:746)
   by 0x4041C1: main (main.C:35)


26 errors in context 2 of 2:
Mismatched free() / delete / delete []
[...]

PR #1491, opened 2017-11-06
Use array form of delete for strings.

jwpeterson:
Refs #1489.


jwpeterson:
I've attached the Valgrind test here...

roystgnr:
Thanks!  I'm kind of surprised it took Valgrind to ferret that out.  I guess the GLIBCXX_DEBUG type checks are only potentially safe because they can juggle with namespaces, whereas there's no way to make a "slower but self-testing" delete without breaking ABI compatibility in undetectable ways...

Issue #896, opened 2016-04-14
Delay projection when adapting multiple times


friedmud:
It would be nice to be able to not have to reproject the solution if we're going to do multiple "steps" of adaptivity at once (like: refine 4 times... THEN do one projection).  I have no idea how this would work... I'm just putting this out there...

PR #1497, opened 2017-11-09
Break EquationSystems::reinit() into solutions,systems parts

roystgnr:
This may let savvy users call only the former if they don't (yet) need
the latter.

As discussed in https://groups.google.com/d/msg/moose-users/ewSE3PsOkoE/FRIAu3P7CwAJ

The goal here was to avoid unnecessary Jacobian reallocation when doing multiple adaptivity steps per time step.  The benefit seems to be mediocre (2% speed up on the test application @jessecarterMOOSE gave me) and the MOOSE half of the update isn't ready to go yet (and won't be compatible with libMesh pre-this-PR), but even if the optimization never goes into MOOSE this is a sensible refactoring for libMesh, so I'd like to put it in regardless.

### DIFF
--- a/include/systems/equation_systems.h
+++ b/include/systems/equation_systems.h
@@ -109,7 +109,8 @@ public:
   virtual void init ();
 
   /**
-   * Reinitialize all the systems
+   * Handle any mesh changes and reinitialize all the systems on the
+   * updated mesh
    */
   virtual void reinit ();
 
@@ -460,6 +461,18 @@ public:
    **/
   bool refine_in_reinit_flag() { return this->_refine_in_reinit; }
 
+  /**
+   * Handle any mesh changes and project any solutions onto the
+   * updated mesh.
+   *
+   * \returns Whether or not the mesh may have changed.
+   */
+  bool reinit_solutions ();
+
+  /**
+   * Reinitialize all systems on the current mesh.
+   */
+  virtual void reinit_systems ();
 
   /**
    * Data structure holding arbitrary parameters.
--- a/src/systems/equation_systems.C
+++ b/src/systems/equation_systems.C
@@ -118,6 +118,18 @@ void EquationSystems::init ()
 
 void EquationSystems::reinit ()
 {
+  const bool mesh_changed = this->reinit_solutions();
+
+  // If the mesh has changed, systems will need to reinitialize their
+  // own data on the new mesh.
+  if (mesh_changed)
+    this->reinit_systems();
+}
+
+
+
+bool EquationSystems::reinit_solutions ()
+{
   parallel_object_only();
 
   const unsigned int n_sys = this->n_systems();
@@ -249,15 +261,17 @@ void EquationSystems::reinit ()
           // dof_constraints_created = true;
         }
     }
-
-  // If the mesh has changed, systems will need to create new dof
-  // constraints and update their global solution vectors
-  if (mesh_changed)
-    {
-      for (unsigned int i=0; i != this->n_systems(); ++i)
-        this->get_system(i).reinit();
-    }
 #endif // #ifdef LIBMESH_ENABLE_AMR
+
+  return mesh_changed;
+}
+
+
+
+void EquationSystems::reinit_systems()
+{
+  for (unsigned int i=0; i != this->n_systems(); ++i)
+    this->get_system(i).reinit();
 }
 
 


jessecarterMOOSE:
@roystgnr Do you have a sense how this will be handled in MOOSE? There will obviously need to be some way of knowing, when `reinit()` is called, whether or not it is the last time, right? Is the idea to call `EquationSystems::reinit_solutions()` after each cycle, but only `EquationSystems::reinit_systems()` after the last cycle?

>  The benefit seems to be mediocre (2% speed up on the test application @jessecarterMOOSE gave me)

I don't see how, on the MOOSE end of things, there would be a speed up at all if you're just splitting something up.

roystgnr:
> There will obviously need to be some way of knowing, when reinit() is called, whether or not it is the last time, right? Is the idea to call EquationSystems::reinit_solutions() after each cycle, but only EquationSystems::reinit_systems() after the last cycle?

That's what my MOOSE patch does, yes.

> I don't see how, on the MOOSE end of things, there would be a speed up at all if you're just splitting something up.

There isn't any.  2% is the improvement with both this *and* the MOOSE patch to use it.

roystgnr:
I am utterly baffled by the MOOSE side of this.  If I *add* a call to ```_eq.reinit_systems();``` at the end of FEProblemBase::adaptMesh(), even without other changes, that's enough for ```time_integrators/explicit-euler.2d-linear-adapt``` to die on me with a PETSc ILU failure:
```
[0]PETSC ERROR: Matrix is missing diagonal entry 0
```

Now, the ImplicitSystem::reinit() is indeed zeroing out the matrices... but surely that's okay, because the matrix will be re-assembled after adaptMesh() is over?

jessecarterMOOSE:
> 2% is the improvement with both this and the MOOSE patch to use it.

That's a bit of a surprise. Where's all the time going?

> I am utterly baffled by the MOOSE side of this.

It takes a skilled hand to muck around in FEProblemBase. Most of my attempts have failed. Things are very order dependent. Might need some help from @permcody on the MOOSE side.

roystgnr:
> Where's all the time going?

Constructing sparsity patterns, allocating Jacobians with them, then deallocating right afterward.

Which is cheaper than I thought, it turns out.  Are you asking "Where's the *remaining* time going?"  Looks like by far the heaviest expense is the point evaluations in our projection code.  I'm going to hit that next, but before I can do so I need to bring https://github.com/roystgnr/libmesh/tree/projection_matrix up to date and into master so I don't end up inadvertently forking libMesh.  Half the point of the current projection code design was to allow @bboutkov and @pbauman to get it working for geometric multigrid too, but I haven't yet merged everything I set up for them to play with.

> Might need some help from @permcody on the MOOSE side.

I'd appreciate it.  Even if this optimization only gives us 2%, there may be other opportunities I'm missing because I don't yet understand well enough how the FEProblem classes work.

permcody:
Are you only failing the explicit Euler case? If so, that should be a hint here. Using the explicit integrator flips several switches throughout the framework to enable what looks like an update type calculation instead of the normal implicit calculation. We swap out access to current vectors in favor of "old" in key places to make this work. I suspect that the problem is buried down in Assembly or MooseVariable. @andrsd - do you have any insight to this problem before I dive in? It's a pretty small patch.

andrsd:
It can be because `_const_jacobian` is `true`. If `_dt` changes this is set to false, so the jacobian is recomputed in the next time step. 

If mesh changes, we have to force jacobian formation. You can do it via setting `_has_jacobian = false;` in `FEProblemBase::meshChanged()`. `_has_jacobian` and `_const_jacobian` work together.
Try that if that helps.

roystgnr:
Ah!  That's got to have something to do with it, doesn't it?  Is it possible that in the explicit Euler case, because the Jacobian is such a simple standard mass matrix, it gets written by meshChanged() rather than later, in the solve?

andrsd:
Not sure I understand. `meshChanged()` does not mess with the matrix. When you flip the flag, the matrix gets formed in the next `computeJacobian` call.  The flags are just a simple optimization, so that the matrix gets formed just once and is not assembled over and over, while it is "constant".  Note that the way we do nodal BCs can modify the matrix, so it is not really pure mass matrix...

roystgnr:
Well, I must be the one who doesn't understand, then.  I was hoping that the Jacobian computation might be getting snuck in via one of the _notify_if_mesh_changes objects.

andrsd:
Go look at `FEProblemBase::computeJacobian` and it should be clear. Just flipping the flag should be enough to tell MOOSE to assemble the new Jacobian matrix and fix the wrong-state problem you were seeing before.

roystgnr:
Hmm... perhaps I'm getting into a state where the adaptivity loop doesn't change the mesh?  Then the flag would never have been flipped to begin with, and if I stick a reinit() at the end of the loop anyway then I'm zeroing out the matrices without telling MOOSE to regenerate them.

I'll try checking for that possibility first; even if it's not what's hitting me, it might happen and it would be a bug if it ever did.

jessecarterMOOSE:
@roystgnr I think I see where you got 2% speedup from. I didn't actually program the logic into `FEProblemBase` to delay the call to `reinit_systems()` but I did look at the time spent there and it was on the order of 2%-3% for my benchmark problem. Unfortunately, with more CPU's this drops down below 1%. Same thing goes for a 3D problem. The good news is that code is pretty efficient!

This gets me wondering, since each reinit is so expensive, and since there can be so many, is there any precedence for delaying _all_ reinitialization until the end? In my naive view of things, if you have a linear variable on an element, and you split (refine) that element into a bunch of children and grandchildren, you can wait until the end to project your solution, right? (I believe some discussion on this led to #896). 

Additionally, what if, instead of doing "sweeps" across the mesh, you recursively called each child element you created? In my problem, my refinement criteria is solely based on position, so you don't need the solution on the previous mesh to inform your algorithm. Once you have your element centroid and list of "interesting" coordinates (and the distance between them), you have enough to make a decision.

I'm sure there's a lot more going on under the hood than I realize or even can possibly understand. If this is all crazy talk then just let me know.

roystgnr:
Handling the "just kidding, no adaptivity" case fixed the regression for me, so I'll put in a PR for the MOOSE side shortly.

> is there any precedence for delaying all reinitialization until the end? In my naive view of things, if you have a linear variable on an element, and you split (refine) that element into a bunch of children and grandchildren, you can wait until the end to project your solution, right?

In theory, yes.  In practice, not without some major underlying changes.  Right now we don't even have a way to flag an element as "JUST_COARSENED_TWICE", only as "JUST_COARSENED"; likewise for refinement.  We don't have a way to tell reinit "refine the mesh topology but don't renumber DoFs".  We don't have a way to iterate over all coarsened-away descendant elements in coarsening projections, only over all coarsened-away child elements of a single level.

Holding off and doing a single projection is definitely the *right* way to do multi-level AMR/C, but we've never had a reason to do that before, and so we're far short of being able to do it; don't expect it soon.

PR #1502, opened 2017-11-14
Add support for detecting Intel 18

jwpeterson:
This bug recently hit an Intel 18 user whose compiler was actually
detected to be Intel 8 (a compiler from circa 2004!) due to the
overzealous pattern matching used in compiler.m4.

A larger question is whether we should being removing support for
older compilers, especially those that don't support C++11, since we
now require that in the library.

### DIFF
--- a/configure
+++ b/configure
@@ -5767,6 +5767,11 @@ $as_echo "<<< C++ compiler is Intel Itanium ICC 7.0 >>>" >&6; }
           if test "x$is_intel_icc" != "x" ; then
             GXX_VERSION_STRING="`($CXX -V 2>&1) | grep 'Version '`"
             case "$GXX_VERSION_STRING" in
+              *18.*)
+                { $as_echo "$as_me:${as_lineno-$LINENO}: result: <<< C++ compiler is Intel(R) icc 18 >>>" >&5
+$as_echo "<<< C++ compiler is Intel(R) icc 18 >>>" >&6; }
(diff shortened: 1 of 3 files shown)


jwpeterson:
Looks like @benkirk recently fixed some regressions in our `--disable-amr` and `--disable-optional` configurations, so I've tacked those recipes onto this PR as well.

roystgnr:
Thanks!

jwpeterson:
> Thanks!

Sure, what do you think about the idea of a major culling in compiler.m4, removing all the configuration stuff for compilers that we know don't support C++11?

jwpeterson:
This isn't (just) ASCII art, it's also the end of our `DETERMINE_CXX_BRAND` macro in compiler.m4:
```
                        fi
                      fi
                    fi
                  fi
                fi
              fi
            fi
          fi
        fi
      fi
    fi
  fi
```

roystgnr:
We should get rid of that atrocity by merging the command substitutions into the if statements and then replacing the "else if" with elif.

That said, I don't really have any objections to a culling.

benkirk:
Case statement?

On Nov 15, 2017, at 12:43 PM, roystgnr <notifications@github.com<mailto:notifications@github.com>> wrote:


We should get rid of that atrocity by merging the command substitutions into the if statements and then replacing the "else if" with elif.

That said, I don't really have any objections to a culling.

PR #1504, opened 2017-11-15
System reinit tweaks

roystgnr:
Our solution update in System::reinit() has long been redundant; this PR removes it.  If we projected our solution then the solution is already up to date; if not then we only need a new init() (which can be done in the projection codepath for simplicity), and we don't need to actually set any values.

We have a few other reinit() overload which have been vestigial ever since we moved our System subclass's vectors into System control via the add_vector() interface revamp many years ago; this PR removes them.

### DIFF
--- a/include/systems/explicit_system.h
+++ b/include/systems/explicit_system.h
@@ -71,12 +71,6 @@ public:
   virtual void clear () libmesh_override;
 
   /**
-   * Reinitializes the member data fields associated with
-   * the system, so that, e.g., \p assemble() may be used.
-   */
-  virtual void reinit () libmesh_override;
-
-  /**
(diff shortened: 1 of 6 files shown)


roystgnr:
Turning on a stupid level of extra testing, just to be paranoid.

My first iteration of this found a bug that was only uncovered by our adjoint refinement examples, because only they exercised the solution-projection-disabling code which I'd forgotten to consider.  If there's any other corner cases in active use that I've missed then I'd like to find out now rather than at MOOSE libmesh update time.

roystgnr:
That failure looks like nothing more than trying to test a VTK-dependent external app with a non-VTK libMesh build?

Okay, I'm confident enough now.

Issue #1503, opened 2017-11-15
Discontinuous Time Dependent Solution Output


kevindugan:
I would like to add a function in the `ExodusII_IO` class which will output a discontinuous solution given a specific time step. This function would be similar to `write_timestep` except instead of calling `write_equation_systems`, it would call the existing `write_discontinuous_exodusII` function.

This change should have minimal impact on existing users, and this change will facilitate outputting discontinuous solutions in other codes which depend on libMesh

kevindugan:
I'm going to go ahead and close this issue.

kevindugan:
I'm reopening this issue since I found some deficiencies in the previous implementation. The deficiencies were found during a pull request with MOOSE (idaholab/moose#11210).

PR #1505, opened 2017-11-16
Adding discontinuous timestep functionality to exodus output.

kevindugan:
Issue #1503

### DIFF
--- a/include/mesh/exodusII_io.h
+++ b/include/mesh/exodusII_io.h
@@ -140,6 +140,18 @@ public:
                                      const std::set<std::string> * system_names=libmesh_nullptr);
 
   /**
+   * Writes a discontinuous solution at a specific timestep
+   * \param fname Name of the file to be written
+   * \param es EquationSystems object which contains the solution vector
+   * \param timestep The timestep to write out. (should be _1_ indexed)
+   * \param time The current simulation time
+   */
+  void write_timestep_discontinuous (const std::string &fname,
+                                     const EquationSystems &es,
+                                     const int timestep,
+                                     const Real time);
+
+  /**
    * Write out element solution.
    */
   void write_element_data (const EquationSystems & es);
--- a/src/mesh/exodusII_io.C
+++ b/src/mesh/exodusII_io.C
@@ -100,7 +100,19 @@ void ExodusII_IO::write_discontinuous_exodusII(const std::string & name,
 }
 
 
+void ExodusII_IO::write_timestep_discontinuous (const std::string &fname,
+                                                const EquationSystems &es,
+                                                const int timestep,
+                                                const Real time)
+{
+  _timestep = timestep;
+  write_discontinuous_exodusII(fname,es);
 
+  if (MeshOutput<MeshBase>::mesh().processor_id())
+    return;
+
+  exio_helper->write_timestep(timestep, time);
+}
 
 // ------------------------------------------------------------
 // When the Exodus API is present...


jwpeterson:
@kevindugan do you have a test you used while developing this patch? I seem to recall a limitation of `write_discontinuous_exodusII()` on meshes with non-contiguous node numberings, which I doubt has been fixed since no one has recently used it. BTW, I invited you to join the libMesh Associates team so your PRs will begin testing automatically in the future.

kevindugan:
@jwpeterson 
I was using a test to develop this patch, but I wasn't sure where to put it in the libmesh testing framework. Actually, I was testing using the Moose framework; but the test is simple enough that I could extract it for libmesh. I'm going to include an outline of the test I was performing here.

Given a set of elemental data, I transfer that data to the all dofs for an element.
```
void transferObject::setLibmeshFieldData (const std::vector<double> &data) const {

    libMesh::MeshBase::const_element_iterator elements_end = d_libmesh_mesh->local_elements_end();
    unsigned int i = 0;
    for ( libMesh::MeshBase::const_element_iterator el_it = d_libmesh_mesh->local_elements_begin();
          el_it != elements_end; ++el_it){
        for (unsigned int dof = 0; dof < (*el_it)->n_dofs(d_system_id, d_variable_id); dof++){
            libMesh::dof_id_type dof_id =
                (*el_it)->dof_number( d_system_id, d_variable_id, dof );
            // printf("Rank: %4d Dof_id: %4d  Power: %10.4E\n",proteusRank, dof_id, data[i]);
            d_libmesh_system->solution->set( dof_id, data[i] );
        }
        i++;
    }

    d_libmesh_system->solution->close();
    d_libmesh_system->update();
}
``` 

I initialize and write the discontinuous data for each time step using the following
```
void Executioner::initializeExodusOutput(const std::string &filename){
    _exodusFilename = filename;
    libMesh::EquationSystems* sys = &_app.executioner()->feProblem().es();
    _exodus_io_ptr = libmesh_make_unique<ExodusII_IO>(sys->get_mesh());
    _exodus_io_ptr->write_timestep_discontinuous(_exodusFilename, _app.executioner()->feProblem().es(), 1, 0.0);
}

void Executioner::outputDiscontinuousExodus (const int timestep, const Real time) const {
    _exodus_io_ptr->write_timestep_discontinuous(_exodusFilename, _app.executioner()->feProblem().es(), timestep, time);
}
```

So the main test would look something like this
```
int main (){
    initializeExodusOutput("output.e");
    for (unsigned int i = 0; i < nSteps; i++){
        std::vector<double> data = getSomeNodalData();
        transfer.setLibmeshFieldData(data);
        outputDiscontinuousExodus (i+1, current_time + dt);
        current_time += dt;
    }
}
```

I was then comparing a gold file to the one produced from this test using exodiff. Could you point me to any documentation for the libmesh testing framework, and would this test be sufficient?

jwpeterson:
The MOOSE test harness is probably the best (eventual) place for your test, since libmesh does not currently build/distribute exodiff, and therefore we currently focus mainly on unit testing (cppunit stuff in the `tests` directory). I was mainly curious if the Mesh you were testing on was of the `GeneratedMesh` variety or something more complicated.

kevindugan:
The mesh I am using is an exodus mesh generated from Cubit. It's a cylinder inside a cylindrical shell, which is inside a box; each of the objects is its own block. The solution is zero except for within the cylinder.

Attached is a cutaway Image of my solution/mesh.
screen shot 2017-11-16 at 12 07 33

jwpeterson:
> Attached is a cutaway Image of my solution/mesh.

OK, cool, this is definitely working well enough to be merged I'd say. 

Cubit sometimes generates meshes with non-contiguous node numberings, so it's possible that is already working. You can tell by ncdump'ing the Exodus file and looking at the entries of the `node_num_map`...

kevindugan:
So if you open up the file output from libmesh through this new discontinuous solution method, the `elem_num_map` and `connect#` are contiguous. But if you open up the initial mesh that I feed into libmesh, the mesh is not contiguous. It looks like the `write_discontinuous_exodusII` function can handle a non-contiguous input mesh.

roystgnr:
I do remember fixing bugs with ExodusII on discontiguously numbered meshes, years ago, but I don't recall (and can't find via git log) any discontinuous-output-specific fixes.  Glad to hear it's working!

PR #1507, opened 2017-11-29
projection_matrix() unit tests

roystgnr:
Swiping commits from @bboutkov.  He did all the work and deserves all the credit, but this way if there are any more conflicts with --enable-complex, --enable-parmesh, --enable-mesopotamian-mode, etc. then I can add fixes quickly.

### DIFF
--- a/include/systems/system.h
+++ b/include/systems/system.h
@@ -1667,7 +1667,7 @@ public:
    * degrees of freedom on that side will still match the coarse grid
    * approximation of the BC, not the fine grid approximation.
    */
-  void projection_matrix (SparseMatrix<Real> & proj_mat) const;
+  void projection_matrix (SparseMatrix<Number> & proj_mat) const;
 
 protected:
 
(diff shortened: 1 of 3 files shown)


pbauman:
@bboutkov can chime in here. He and I were chatting earlier and he said he was hitting snags with the node number in the DistributedMesh, at which point I suggested he do exactly what @roystgnr did: open a PR for discussion.

bboutkov:
Yeah sorry, communications a bit outta sync but as discussed - the DistributedMesh case fails because node numbering on DistributedMesh wont match the SerialMesh case upon refinement despite the coarse node numbering generated by MeshGeneration matching. As far as I understand we need to do the node->dof mapping even more manually by referencing the physical node coordinates.

roystgnr:
Is this even running on ReplicatedMesh in parallel?  The node numbering may match the serial case, but the dof numbering won't...

roystgnr:
Ah, never mind, I see the node2dof translations you're using.  And this is passing on ReplicatedMesh in parallel for me, so forcing ReplicatedMesh ought to be sufficient for now.

Issue #1509, opened 2017-11-30
MetaPhysicL tests failing in make check


dschwen:
Running `make check` in an up to date libmesh repo yields

```
FAIL: derivs_unit
PASS: compare_types_unit
PASS: instantiations_unit
PASS: main_unit
FAIL: identities_unit
PASS: shadow_sparse_struct_pde_unit
PASS: divgrad_unit
PASS: shadow_sparse_vector_navier_unit
PASS: shadow_sparse_vector_pde_unit
PASS: shadow_sparse_struct_navier_unit
PASS: dynamic_sparse_vector_pde_unit
PASS: shadow_dynamic_sparse_vector_pde_unit
PASS: shadow_vector_pde_unit
PASS: shadow_vector_navier_unit
PASS: sparse_struct_pde_unit
PASS: sparse_struct_navier_unit
PASS: testheaders_unit
PASS: testopt_unit
PASS: sparse_vector_navier_unit
PASS: sparse_vector_pde_unit
FAIL: sparse_derivs_unit
FAIL: sparse_identities_unit
PASS: vector_navier_unit
PASS: vector_pde_unit
PASS: dualnamedarray_unit
PASS: namedindexarray_unit
PASS: physics_unit
============================================================================
Testsuite summary for metaphysicl 0.2.0
============================================================================
# TOTAL: 27
# PASS:  23
# SKIP:  0
# XFAIL: 0
# FAIL:  4
# XPASS: 0
# ERROR: 0
```

roystgnr:
Damn; I get "PASS:  27" and so did Civet.  I hate trying to fix bugs I can't replicate.

Could you share system details (is it failing on Macs?)?  And... let's say the identities_unit.log file, since there's a good chance fixing that will fix sparse_identities_unit and a slight chance it will fix *_derivs_unit.

dschwen:
./contrib/metaphysicl/0.2.0/test/identities_unit.log:
```
Value 1.08399
Error nan
Tol   1.19209e-06
Failed test: random_vec - cos(acos(random_vec))
FAIL identities_unit (exit status: 1)
```

Yes, this is a mac (High Sierra, moose package version 8, clang 5.0) and I'm on 10c606d268da0aab1c9e810b01173c50db27bdde

roystgnr:
Well, *that* was unexpected.  I could have sworn cos and acos were inverse functions...

[...]

PR #1510, opened 2017-11-30
Don't try to test acos() outside its domain

roystgnr:
Thanks to @dschwen for catching this in #1509.

I'm actually kind of disturbed that these worked on Linux.

I double checked and, for the seed 12345, the PRNG apparently only gives me values between 0 and 0.4 times RAND_MAX for its first 10 outputs (and possibly more, but 10 is all I asked for in those unit tests).  I'm not sure what the proper Bonferroni correction factor is here, but I hope it's big, because without correction I'm forced to assume that the Linux rand() implementation sucks with p~=1e-4.  I know rand() sucks in *general* but I thought the Linux implementation was supposed to be half decent.

### DIFF
--- a/contrib/metaphysicl/0.2.0/test/derivs_unit.C
+++ b/contrib/metaphysicl/0.2.0/test/derivs_unit.C
@@ -81,10 +81,10 @@ int vectester (void)
 
   std::srand(12345); // Fixed seed for reproduceability of failures
 
-  // Avoid divide by zero errors later
+  // Avoid divide by zero errors or acos(x>1) NaNs later
   for (unsigned int i=0; i != N; ++i)
     {
-      random_vec[i] = .25 + (static_cast<Scalar>(std::rand())/RAND_MAX);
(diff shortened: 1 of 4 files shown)


dschwen:
`sparse_identities_unit` is still broken for me at 03c82c6

roystgnr:
That looks like an independent error, so I'm going to merge this shortly; 1 failure is still better than 4...

roystgnr:
I still need to figure out how best to propagate this stuff upstream.  I don't suppose git has an automated way to take a set of commits from a completely different repo, strip off leading directory names, and try to apply them?

dschwen:
Yes, there is a way to do this! One sec...

dschwen:
```
~/Programs/libmesh/contrib> git format-patch HEAD^ --relative
```

jwpeterson:
Followed up by `git am 0001-foo-bar.patch` in the other repo...

PR #1513, opened 2017-12-01
More MetaPhysicl fixes and test coverage

roystgnr:
Technically there's only one bugfix here, but the fact that we weren't catching that bug on Linux was sort of a meta-bug, so this PR also has two new API additions designed to make testing and usage easier, plus four commits expanding test coverage.

@dschwen, see if this is a sufficient fix for you?  I actually *don't* think this adds as much testing as the bugfixed method ought to have, but we don't actually use that class within libMesh yet so I'm happy with "make sure 'make check' is robust right away" as a temporary stop on the way to "make sure 'make check' has enough test coverage eventually".

### DIFF
--- a/contrib/metaphysicl/0.2.0/src/numerics/include/metaphysicl/dynamicsparsenumberbase.h
+++ b/contrib/metaphysicl/0.2.0/src/numerics/include/metaphysicl/dynamicsparsenumberbase.h
@@ -151,7 +151,7 @@ template <typename T, typename I, template <typename, typename> class SubType>
 inline
 T&
 DynamicSparseNumberBase<T,I,SubType>::operator[](index_value_type i)
-{ return _data[runtime_index_of(i)]; }
+{ return this->query(i); }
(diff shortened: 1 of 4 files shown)


roystgnr:
I'm guessing this is a 9/80 day at INL?  Going to merge this now since it at least fixes *some* bugs, but we won't touch #1509 until @dschwen can confirm or deny that it fixes all the bugs he uncovered.

jwpeterson:
> I'm guessing this is a 9/80 day at INL?

Nope. Every other Friday is an extra long lunch day for @dschwen though.

dschwen:
I'm at my desk. Sipping a latte.. but at my desk.

dschwen:
Testing it now. Finally learned about the `git fetch upstream pull/1513/head:meta_fix2` magic!

dschwen:
I'm happy to report that all tests pass now! Thanks @roystgnr

PR #1514, opened 2017-12-01
Projection Matrix 3D Tet4 test

bboutkov:
As discussed, here is an additional Tet4 unit test for projection_matrix. I elected to construct the Tet4 manually instead of going through MeshGen build_cube() to avoid the Hex -> Tet conversion that would bloat the test with extra elements.

Please let me know if theres any issues I should address - thanks.


roystgnr:
Reusing the method from the previous unit test means you also piggy-back on the issue we fixed in the previous swath of unit tests, and I don't see any other problems.  Thanks!

PR #1515, opened 2017-12-03
Use std::unique_ptr.

jwpeterson:
Now that we require C++11, there's no reason to keep supporting all the C++03 compatible workarounds. The `UniquePtr` type will of course remain in place for backwards compatibility.


jwpeterson:
Looks like GRINS was relying on "libmesh/libmesh_common.h" being included by "libmesh/auto_ptr.h"? Should be a fairly easy fix.

The failure in MOOSE is because there is still a deprecated `AutoPtr` in there somehow... thought I had removed all those but I guess one snuck back in...
```
In file included from src/base/MooseTestApp.C:192:0:
include/userobjects/RandomHitSolutionModifier.h:61:3: error: 'AutoPtr' does not name a type
   AutoPtr<PointLocatorBase> _point_locator;
```

roystgnr:
> Looks like GRINS was relying on "libmesh/libmesh_common.h" being included by "libmesh/auto_ptr.h"? Should be a fairly easy fix.

Yeah, that's definitely technically @pbauman's bug.

On the other hand, it's the sort of bug that we found about a hundred of in the course of https://github.com/idaholab/moose/pull/8743 and https://github.com/idaholab/moose/pull/8758, so I don't want to throw stones.

Maybe we should have a CI test insisting that all our code just live up to [iwyu](https://github.com/idaholab/moose/pull/8758#issuecomment-290432447) standards eventually...

:+1: once GRINS is ready; I can put in the fix upstream there later this week if @pbauman is busy.

pbauman:
> Maybe we should have a CI test insisting that all our code just live up to iwyu standards eventually...

Holy crap, iwyu looks awesome. Even better than the Python tool I'd seen ... a year ago(?). I'll try and get to the fix today (would be late this afternoon or this evening).

pbauman:
All [jobs](https://femputer.eng.buffalo.edu:8000/event/310/) on f668b7e : invalidated by @pbauman

Invalidated to pull in fixes downstream in GRINS (thanks @jwpeterson).

include/base/factory.h
```diff
@@ -146,7 +146,7 @@ std::unique_ptr<Base> Factory<Base>::build (const std::string & name)
     }
 
   Factory<Base> * f = factory_map()[name];
-  return std::unique_ptr<Base>(f->create());
+  return libmesh_make_unique<Base>(f->create());
```
roystgnr: You'd think this was trivial, but it just broke a branch I'm working on!

```
../../../include/libmesh/auto_ptr.h(47): error: object of abstract class type "libMesh::Partitioner" is not allowed:
            function "libMesh::Partitioner::clone" is a pure virtual function
            function "libMesh::Partitioner::_do_partition" is a pure virtual function
    return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
                                  ^
          detected during:
            instantiation of "std::unique_ptr<T, std::default_delete<T>> libMesh::make_unique<T,Args...>(Args &&...) [with T=libMesh::Partitioner, Args=<std::unique_ptr<libMesh::Partitioner, std::default_delete<libMesh::Partitioner>>>]" at line 149 of "../../../include/libmesh/factory.h"
            instantiation of "std::unique_ptr<Base, std::default_delete<Base>> libMesh::Factory<Base>::build(const std::string &) [with Base=libMesh::Partitioner]" at line 332 of "../../../../examples/adjoints/adjoints_ex1/adjoints_ex1.C"
```

Here ```f->create()``` isn't arguments to a new Base, it is itself a ```unique_ptr<Derived>``` that we're "casting" down to ```unique_ptr<Base>```.

roystgnr: I'll add the fix in that branch.

I was *just* thinking that it was nice that we'd finally be getting some test coverage inside the library of the Factory class!

jwpeterson: Huh, yeah I can see the issue, but I wonder what compiler you are using that caught it when it had already passed all the CI testing?

roystgnr: Not compiler, branch.  I guess we just never instantiated that method in GRINS or MOOSE?!  I know we didn't internally.

IIRC Factory is in use in FIN-S, though, so hopefully I'll finish testing this PR soon and we can get it merged before @benkirk has to step in with one of his quiet "they broke my stuff again" pushes to master.

PR #1518, opened 2017-12-05
Fix Metaphysicl warnings

jwpeterson:
The unit test ones I was seeing because I don't have MASA installed. The others anyone compiling in opt mode should have seen...

### DIFF
--- a/contrib/metaphysicl/0.2.0/test/testopt_unit.C
+++ b/contrib/metaphysicl/0.2.0/test/testopt_unit.C
@@ -30,7 +30,7 @@ int main(void)
   NumberArray<3, float> NA = 0;
   std::cos(NA);
 
-  NumberVector<3, float> NV = 0;
+  // NumberVector<3, float> NV = 0;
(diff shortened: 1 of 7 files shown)


contrib/metaphysicl/0.2.0/test/testopt_unit.C
```diff
@@ -30,7 +30,7 @@ int main(void)
   NumberArray<3, float> NA = 0;
   std::cos(NA);
 
-  NumberVector<3, float> NV = 0;
```
roystgnr: Can we change this to ```(void)NV;``` or something to get rid of the unused variable warning?  Otherwise we're removing test coverage (probably redundant, but I can't swear that) for that constructor.

jwpeterson: Will do, sorry didn't actually realize this was a constructor for some reason.

roystgnr:
Thanks!

PR #1519, opened 2017-12-05
More std::unique_ptr updates

jwpeterson:
A follow-up to the changes in #1515.

This PR includes some additional uses of `libmesh_make_unique` as well as a refactoring of the `ExactErrorEstimator` and `ExactSolution` classes to use containers of `std::unique_ptrs`. (They were always intended to use them, but we couldn't rely on having std containers of smart pointers until now.)

### DIFF
--- a/examples/adaptivity/adaptivity_ex2/adaptivity_ex2.C
+++ b/examples/adaptivity/adaptivity_ex2/adaptivity_ex2.C
@@ -375,7 +375,6 @@ int main (int argc, char ** argv)
               // solution and assigns to each element a positive error value.
               // This value is used for deciding which elements to refine
               // and which to coarsen.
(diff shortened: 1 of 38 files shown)


roystgnr:
Ha!  I'm not the only one lazy enough to abuse Civet before running "make check" myself!

jwpeterson:
> Ha! I'm not the only one lazy enough to abuse Civet before running "make check" myself!

`make check` passed on my system... I need to look at the error carefully, maybe clang is allowing something illegal here.

jwpeterson:
Ah, I think it's the difference between a compiler that supports C++14 std::make_unique and one that doesn't...

jwpeterson:
Hm... there are several more of the same issue which I need to fix and then reopen this PR.

PR #1520, opened 2017-12-05
Don't turn on verbose ExodusII_IO output by default in debug mode.

jwpeterson:
This prevents some unnecessary output during debug unit tests, and is less surprising when comparing the output of optimized and debug runs.

This has been like this forever so we may not want to just change it, but I thought we could at least start a discussion...

### DIFF
--- a/src/mesh/exodusII_io.C
+++ b/src/mesh/exodusII_io.C
@@ -147,10 +147,6 @@ void ExodusII_IO::read (const std::string & fname)
   elems_of_dimension.clear();
   elems_of_dimension.resize(4, false);
 
-#ifdef DEBUG
-  this->verbose(true);
-#endif
-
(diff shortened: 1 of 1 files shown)


pbauman:
Not necessarily opposed to this, but just so I'm clear, this means it will now be app responsibility to trigger verbose output if they want it, i.e. there's no runtime knob on the libMesh side? 

That's a reasonable thing, just means I have to update my app side. 

But whenever I wanted that output, it almost was always to just get info about the mesh that was breaking for which libMesh ships the Exodus tools, so really I'm not missing much since I can just run the Exodus tool to get mesh info.

jwpeterson:
> there's no runtime knob on the libMesh side

Not at this time, no. It would be nice to have a unified/comprehensive system for increasing the debug output of all libmesh objects from the command line, though!

PR #1521, opened 2017-12-05
Linear partitioner example

roystgnr:
The first commit here fixes the case that @bboutkov set up for me in https://github.com/bboutkov/libmesh/tree/distmesh_linearpart_fail - LinearPartitioner now works with DistributedMesh too.

Subsequent commits add test coverage for LinearPartitioner and others to adjoints_ex1, fix the regression in https://github.com/libMesh/libmesh/pull/1515/commits/40ec2d5f1933ec0fae4f49aa1d366d4243ce5247, and make a few tweaks to shared macros and example codes to make the new test coverage easier.


jwpeterson:
Hmm, there are some changes to adjoints_ex6/femparameters.C in #1522 as well, so depending on which PR gets merged first, someone will have conflicts...

roystgnr:
> someone will have conflicts...

Where by "someone" I assume you mean "whoever passes CI last".  *GO!*

roystgnr:
Damn, I didn't actually see which CI finished first.  I'll let the tie go to you, then.

roystgnr:
Damn, don't merge this yet.  Some of the other partitioners enabled by --disable-strict-lgpl aren't yet DistributedMesh-compatible either, so need to be disabled in that case in the new test.

jwpeterson:
> Damn, don't merge this yet.

Want me to attach the "Non-LGPL Contrib" test to this PR?

roystgnr:
Hmm... that's a good idea, even though it's not sufficient to trigger the failure cases I'm seeing.  It ought to actually pass, unless one of those partitioners has regressed so badly as to break with ReplicatedMesh too.

src/partitioning/centroid_partitioner.C
```diff
@@ -38,6 +38,10 @@ void CentroidPartitioner::partition_range(MeshBase & /*mesh*/,
       return;
     }
 
+  // We don't yet support distributed meshes with this Partitioner
+  if (!mesh.is_serial())
```
pbauman: Is `is_serial()` deprecated along with `SerialMesh`? Perhaps favor `is_replicated()`? Not a big deal, just noticed when I glanced to see what partioners don't support `DistributedMesh`

roystgnr: Those aren't just a name change; they are two different methods!

is_serial()
```
   * \returns \p true if all elements and nodes of the mesh
   * exist on the current processor, \p false otherwise
```

whereas is_replicated()
```
   * \returns \p true if new elements and nodes can and should be
   * created in synchronization on all processors, \p false otherwise
```

In other words, DistributedMesh::is_serial() can return true!  And it's the right thing to use here: if e.g. a user wants to keep their DistributedMesh serialized long enough to get a space-filling-curve partitioning of the coarse mesh, but then hand off to ParMETIS for any subsequent repartitioning, there's no reason we can't or shouldn't let them.

pbauman: Thanks for the clarification! 1 more week until I can purge teaching for a month. I need to hide and code my ass off to get back in the swing. Why'd I take this job again?

PR #1524, opened 2017-12-07
JumpErrorEstimator fix for non-zero-norm SCALAR

roystgnr:
I'd been imagining that users would manually set that norm entry to
zero, why should they?  Scalar variables have zero jump, and zero
times anything is still zero.

Thanks to @andrsd for catching this.

I'm shamelessly abusing Civet here; I'm about to be AFK and I haven't actually checked myself yet to see if this even compiles.  I'll also probably rely on David's RELAP-7 test to give us the final word about whether it solves his problem or not.


andrsd:
> I'll also probably rely on David's RELAP-7 test to give us the final word about whether it solves his problem or not.

It fixed the problem. Thanks!

PR #1525, opened 2017-12-08
Replace auto_ptr.h with standard <memory> header 

jwpeterson:
Also simplify some more code that uses unique_ptr.

### DIFF
--- a/examples/adaptivity/adaptivity_ex5/adaptivity_ex5.C
+++ b/examples/adaptivity/adaptivity_ex5/adaptivity_ex5.C
@@ -34,7 +34,7 @@
 #include <cstdlib> // *must* precede <cmath> for proper std:abs() on PGI, Sun Studio CC
 #include <cmath>
 
-// Basic include file needed for the mesh functionality.
+// libMesh includes
 #include "libmesh/libmesh.h"
 #include "libmesh/replicated_mesh.h"
 #include "libmesh/mesh_refinement.h"
@@ -48,13 +48,12 @@
 #include "libmesh/numeric_vector.h"
 #include "libmesh/dense_matrix.h"
 #include "libmesh/dense_vector.h"
-
 #include "libmesh/periodic_boundaries.h"
 #include "libmesh/periodic_boundary.h"
 #include "libmesh/mesh_generation.h"
 #include "libmesh/parsed_function.h"
-
 #include "libmesh/getpot.h"
+#include "libmesh/auto_ptr.h" // libmesh_make_unique
 
 // This example will solve a linear transient system,
 // so we need to include the TransientLinearImplicitSystem definition.
--- a/examples/adjoints/adjoints_ex1/adjoints_ex1.C
+++ b/examples/adjoints/adjoints_ex1/adjoints_ex1.C
@@ -68,6 +68,7 @@
 #include "libmesh/partitioner.h"
 #include "libmesh/steady_solver.h"
 #include "libmesh/system_norm.h"
+#include "libmesh/auto_ptr.h" // libmesh_make_unique
 
 // Error Estimator includes
 #include "libmesh/kelly_error_estimator.h"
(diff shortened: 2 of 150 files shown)


roystgnr:
Removing auto_ptr.h from dof_map.h etc. is a bit iffy - although technically users should include the appropriate header for every single class they use, it's really hard to fault them for not including the header for a class which was required by C++ standards to be included in a header they did include, because that class was returned by a method in that header.

So this definitely feels like API backwards compatibility breakage, and I'd like to leave those auto_ptr.h inclusions until we can deprecate that indirect inclusion somehow... I'm just not sure how.  Put libmesh_deprecated() in UniquePtr::UniquePtr to direct users to switch to std::unique_ptr?

jwpeterson:
I'd be OK with reverting a7bc81f (the commit that moves around headers) for the time being. One goal of putting this PR up here was to see what all would be affected...

> Put libmesh_deprecated() in UniquePtr::UniquePtr to direct users to switch to std::unique_ptr?

Not sure what this means.

jwpeterson:
That being said, you did break all of MOOSE and the apps for like a week once while moving around headers once IIRC?

roystgnr:
Rather than revert a7bc81f, we could split it in two, and only revert the part that removes auto_ptr.h includes from other include/foo/bar.h files.  There are 150 files changed in that commit, and I count less than 50 which could cause problems downstream.

Forget my suggestion re: UniquePtr; making a whole shim class just to stick a libmesh_deprecated() in the constructor is overkill.

But hell, I don't actually see another way to give users a smooth transition here.  I still maintain that there's a non-hypocritical difference between "It's a userspace bug if you rely on indirect inclusion because it happens to work right now" and "We should be nice to users who relied on indirect inclusion as mandated by the C++ language"... but that's a moot point, isn't it, unless there actually *exists* some way for us to be nice?

We want to get rid of auto_ptr.h eventually, and if it's at all practical to give users warning before our internal removals break their code then I'd like to do our usual "they get that warning in an official release or two first" procedure... but I don't actually see any practical way to leave a warning without making a whole annoying shim class.  Can you?  If not then we have to just rip off the band-aid some day, and I guess we ought to just do so right away, so (after GRINS is fixed up) you can merge this as-is.

roystgnr:
Actually, I wonder: how hard would it be to use C++11 variadic templates to create a ```Deprecated<T> : public T``` shim class?  The only code would be a variadic constructor that takes any arguments, passes those arguments to the constructor of T, and then calls ```libmesh_deprecated()``` itself.  I guess we might also need an implicit ```operator T``` to allow "slicing" when passing by value?  Anyway, then we could ```template <typename T> using UniquePtr = Deprecated<std::unique_ptr<T>>``` and be done.  The ```Deprecated``` class would still be non-trivial work but it might be more generally useful in the future.

The other problem in this specific case, though, is that it could break template specializations around unique_ptr/UniquePtr for anyone who (reasonably) assumed that those two were identical.

pbauman:
All [jobs](https://femputer.eng.buffalo.edu:8000/event/336/) on a3509f1 : invalidated by @roystgnr

Trying again with https://github.com/grinsfem/grins/pull/514 merged

jwpeterson:
Hmm... looks like GRINS failed to build again because I don't have some optional dependencies available (Antioch). I will try push another GRINS PR later today hopefully.

jwpeterson:
Rather than get too fancy with deprecation approaches, I partially reverted a7bc81f so that we don't yet drop any header inclusions, and commented on the includes which can eventually be removed. I think we should just drop the includes some time after 1.3.0 is released, but I also won't complain if @roystgnr wants to add some kind of `Deprecated<T>` class in the meantime...

This last commit should get GRINS compiling again, but @pbauman may also want to try compiling GRINS+Antioch+etc. with a3509f1 to see what other GRINS headers would need to include "auto_ptr.h" until they eventually get around to dropping `UniquePtr`.

PR #1527, opened 2017-12-12
Deprecate Preconditioner::build().

jwpeterson:
This method returns dumb pointers to heap-allocated objects and
therefore is prone to creating memory leaks. The new
build_preconditioner() method should be used instead.

### DIFF
--- a/include/numerics/preconditioner.h
+++ b/include/numerics/preconditioner.h
@@ -70,12 +70,25 @@ public:
   virtual ~Preconditioner ();
 
   /**
+   * Builds a \p Preconditioner using the linear solver package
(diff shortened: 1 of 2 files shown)


roystgnr:
Didn't I just add a call to this in adjoints_ex1?  Could we switch that to use the new method now?

jwpeterson:
> Didn't I just add a call to this in adjoints_ex1?  

No, that was `Factory<Partitioner>::build()`, which looks like it uses `libmesh_make_unique` internally.

roystgnr:
Ah, that's right.  Thanks!

PR #1528, opened 2017-12-12
Remove error-then-return code sections.

jwpeterson:
If I recall correctly, we originally used this idiom to prevent
compiler warnings, but it seems like compilers have gotten smarter
about detecting when something is/is not returned from a non-void
function. I think we can now get rid of these...

### DIFF
--- a/include/base/dof_object.h
+++ b/include/base/dof_object.h
@@ -955,8 +955,7 @@ unsigned int DofObject::var_to_vg (const unsigned int s,
       if (var < vg_end) return vg;
     }
 
-  libmesh_error_msg("We'll never get here!");
-  return 0;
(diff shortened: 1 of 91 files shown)


roystgnr:
What compiler versions have you tested on?

jwpeterson:
Clang 3.9.0 and GCC on Ubuntu 16. Also I don’t think we’ve been consistently adding these statements to new code for quite a while... and haven’t seen any warnings about it.

dschwen:
FWIW we've done that for over three and a half years in MOOSE modules and compilers haven't complained: https://github.com/idaholab/moose/commit/e5de842b9e333425285bf3805bec4263f1ba8252#diff-1d4e4cdaf1d83a469c5b427bed4003fcR47

PR #1529, opened 2017-12-12
Drop configure support for ancient compilers

jwpeterson:
libmesh now requires C++11, so  we can drop all the configure tests for older compilers that don't support that standard.


roystgnr:
:+1:

PR #1530, opened 2017-12-12
Work around GCC 4.9.2 issue.

jwpeterson:
This compiler doesn't like the combination of auto-declared variables
and returning `unique_ptr<Derived>` as `unique_ptr<Base>`. Newer compilers
accept this code just fine, so presumably it is OK, this change is
just to work around the older compiler's limitation.

In playing around with this, we discovered that return by move also
fixed the problem, but according to Meyers' "Effective Modern C++"
pg. 173-174, this would actually prevent compilers from applying the
Return Value Optimization (RVO), otherwise known as copy elision,
in this case.

### DIFF
--- a/src/geom/cell_hex.C
+++ b/src/geom/cell_hex.C
@@ -98,7 +98,7 @@ std::unique_ptr<Elem> Hex::side_ptr (const unsigned int i)
 {
   libmesh_assert_less (i, this->n_sides());
 
-  auto face = libmesh_make_unique<Quad4>();
+  std::unique_ptr<Elem> face = libmesh_make_unique<Quad4>();
 
   for (unsigned n=0; n<face->n_nodes(); ++n)
     face->set_node(n) = this->node_ptr(Hex8::side_nodes_map[i][n]);
--- a/src/geom/cell_hex20.C
+++ b/src/geom/cell_hex20.C
@@ -152,7 +152,7 @@ std::unique_ptr<Elem> Hex20::build_side_ptr (const unsigned int i,
 
   else
     {
-      auto face = libmesh_make_unique<Quad8>();
+      std::unique_ptr<Elem> face = libmesh_make_unique<Quad8>();
       face->subdomain_id() = this->subdomain_id();
 
       for (unsigned n=0; n<face->n_nodes(); ++n)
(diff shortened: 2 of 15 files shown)


roystgnr:
Huh... this might actually be the best way to do things regardless.  Is the compiler allowed to do copy elision in the case where the value being returned isn't actually of the declared return value type, but is merely implicitly convertable to such?  http://en.cppreference.com/w/cpp/language/copy_elision seems to suggest otherwise.

jwpeterson:
Phew... standardese is tough to read, but I agree that the compiler is *not required* to perform copy elision in this particular case, at least according to my reading of the **bolded text** below:

---
Under the following circumstances, the compilers are **permitted, but not required** to omit the copy- and move- (since C++11)construction of class objects...

* If a function returns a class type by value, and the return statement's expression is the name of a non-volatile object with automatic storage duration, which isn't a function parameter, or a catch clause parameter, **and which has the same type (ignoring top-level cv-qualification) as the return type of the function**, then copy/move (since C++11) is omitted. When that local object is constructed, it is constructed directly in the storage where the function's return value would otherwise be moved or copied to. This variant of copy elision is known as NRVO, "named return value optimization".
---

So actually being more explicit about the type like this is the most likely way to guarantee the optimization.

roystgnr:
I believe NRVO is not required of the compiler in *any* case.  (So make sure your copy constructors don't suck, kids!  You never know when your users' compiler will be invoking a copy constructor that your test suite's compiler didn't...)

But as far as I can tell, NRVO between different types might not even be permitted!  I'm kind of tempted to set up a test case and see.

permcody:
Fortunately C++11 gives us the ability to delete copy constructors or copy assignments that we never want invoked. I've used that in a few keys places.

jwpeterson:
In this case, the return value can't be copied anyway because it's a `std::unique_ptr`...

roystgnr:
Yeah, but according to the standard the compiler is still allowed not to do NRVO, even if there's no copy constructor to use instead:

> In a return statement or a throw-expression, if the compiler cannot perform copy elision but the conditions for copy elision are met or would be met, except that the source is a function parameter, the compiler will attempt to use the move constructor even if the object is designated by an lvalue

PR #1533, opened 2017-12-14
 ASPECT_RATIO and STRETCH quality metrics for face_quad elements

hchen139:
close #1532

PR #1535, opened 2017-12-14
Add shape and skew metrics for Quad4.

jwpeterson:
For some simple test cases, I think these new metrics work better than the ASPECT_RATIO, DISTORTION/DIAGONAL, and STRETCH metrics we currently have.

Results with existing metrics:

ASPECT_RATIO:
aspect

DISTORTION:
distort

STRETCH:
stretch

Results with new metrics:

SHAPE:
shape

SKEW:
skew

Note that these metrics are in the range [0=bad,1=good] except for ASPECT_RATIO, which is always >= 1, and the larger the value, the worse the element is. I'd actually vote for the current DISTORTION metric to be removed entirely, as it seems to think the nearly triangular element in the center of the domain is OK. SHAPE and SKEW correctly flag both of these nearly triangular elements as being of relatively low quality.

This is follow-up work to #1533.

### DIFF
--- a/src/geom/face_quad.C
+++ b/src/geom/face_quad.C
@@ -218,7 +218,7 @@ Real Quad::quality (const ElemQuality q) const
 
         // Nodal Jacobians. These are 3x2 matrices, hence we represent
         // them by Array6.
-        std::vector<Array6> A(4);
+        std::array<Array6, 4> A;
         for (unsigned int k=0; k<4; ++k)
           {
             unsigned int
@@ -234,7 +234,7 @@ Real Quad::quality (const ElemQuality q) const
 
         // Compute metric tensors, T_k = A_k^T * A_k. These are 2x2
         // square matrices, hence we represent them by Array4.
-        std::vector<Array4> T(4);
+        std::array<Array4, 4> T;
         for (unsigned int k=0; k<4; ++k)
           {
             Real
--- a/src/geom/face_quad.C
+++ b/src/geom/face_quad.C
@@ -15,12 +15,13 @@
 // License along with this library; if not, write to the Free Software
 // Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
-// C++ includes
-
 // Local includes
 #include "libmesh/face_quad.h"
 #include "libmesh/edge_edge2.h"
 #include "libmesh/face_quad4.h"
+
+// C++ includes
+#include <array>
 
 
 namespace libMesh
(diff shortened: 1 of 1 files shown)


jwpeterson:
@bwspenc and @hchen139 may be interested in these as well, the SHAPE metric in particular should be able to detect long, thin rectangular elements.

src/geom/face_quad.C
```diff
@@ -200,14 +201,85 @@ Real Quad::quality (const ElemQuality q) const
...
+            // To initialize std::array we need double curly braces in
+            // C++11 but not C++14 apparently.
+            A[k] = {{x[kp1] - x[k], x[kp3] - x[k],
+                     y[kp1] - y[k], y[kp3] - y[k],
+                     z[kp1] - z[k], z[kp3] - z[k]}};
+          }
+
+        // Compute metric tensors, T_k = A_k^T * A_k. These are 2x2
+        // square matrices, hence we represent them by Array4.
+        std::vector<Array4> T(4);
```
roystgnr: Why std::vector rather than std::array for the outer container here?  We've got a fixed compile-time size.

jwpeterson: I'm using `std::vector` in some places because of the annoying [double curly brace initialization syntax](http://en.cppreference.com/w/cpp/container/array) that is still required in C++11 (and which clang warns about if you forget), but you're right, this one might as well be a `std::array` as well.

hchen139:
@jwpeterson As for long ,thin rectangular elements, aspect_ratio and stretch both should work. But for cases like all edges are four equal but several distorted or degenerated, shape and skew will work better.

jwpeterson:
The "Minimum clang version" recipe failure appears to be caused by a combination of:
1. The version of clang (3.5.0) we're using combined with
1. The version of the GCC (4.8.4) std library headers clang is using, and
1. metaphysicl's use -std=c++14 (is this required for metaphysicl?)

We might be able to fix this simply by configuring libmesh with `--disable-metaphysicl` for those builds... @roystgnr is MOOSE going to need libmesh to be built with metaphysicl in the medium/near future?

```
In file included from ../../../../../contrib/metaphysicl/0.2.0/src/apps/version.C:28:
In file included from ../src/utilities/include/metaphysicl/metaphysicl_version.h:47:
In file included from /opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/iostream:39:
In file included from /opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/ostream:38:
In file included from /opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/ios:42:
In file included from /opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/bits/ios_base.h:41:
In file included from /opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/bits/locale_classes.h:40:
In file included from /opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/string:52:
In file included from /opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/bits/basic_string.h:2815:
In file included from /opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/ext/string_conversions.h:43:
/opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/cstdio:120:11: error: no member named 'gets' in the global namespace
  using ::gets;
        ~~^
```

jwpeterson:
In any case, the failure has nothing to do with this PR so I'm planning to merge it once I change a few more `std::vectors` into `std::arrays`.

roystgnr:
> metaphysicl's use -std=c++14 (is this required for metaphysicl?)

There's part of MetaPhysicL that requires it, but libMesh won't ever be requiring that part.  There may be C++14-dependent-stuff built in a similar vein that *does* end up as a libMesh dependency, but if so it'll be *very* optional.

I'd be happy to force our MetaPhysicL to skip that "upgrade" if only I know how to do that with a nested configure.

> is MOOSE going to need libmesh to be built with metaphysicl in the medium/near future?

The Geometric MultiGrid support is going to require MetaPhysicL.  I don't think anything else will, unless @friedmud ever wants to incorporate [the AD stuff he was playing with](https://github.com/friedmud/meta).

jwpeterson:
OK, for the time being I think we will pass --disable-metaphysicl to our `update_and_rebuild_libmesh.sh` script. Shouldn't be too hard to turn it back on when we're ready to start using the new MG capabilities...

jwpeterson:
Ah, and it looks like @brianmoose already tested out the `--disable-metaphysicl` fix on the min clang version recipe and it works.

bwspenc:
@jwpeterson Thanks for looking into this.  It looks like the current code for ASPECT_RATIO is working correctly, and might be the best option for catching those long, skinny elements we were discussing. Is that right?

jwpeterson:
> Is that right?

I don't personally prefer that one because it's not normalized to (0,1), but yeah as long as you are able to say "we don't like elements with aspect ratios over X" then it should be workable. I also think that nearly degenerate (almost triangular) elements can still have "good" aspect ratios so you may end up missing some otherwise bad elements in this metric. IMO the shape metric takes care of both of these issues, but it's always good to have different options.

bwspenc:
OK, good.  We'll look into the feasibility of using SHAPE as a catch-all quality measure.

PR #1537, opened 2017-12-15
Checkpoint naming

roystgnr:
This implements the first part of my suggestion in #1455 - when opening a CheckpointIO file, we first try to open the requested filename as the header file, but if that isn't found then we next try to open the requested filename with the number of processors appended.

This should allow MOOSE users to easily create multiple pre-splits and have the correct N-to-N option chosen; just rename the header after each creation.

### DIFF
--- a/src/mesh/checkpoint_io.C
+++ b/src/mesh/checkpoint_io.C
@@ -632,17 +632,17 @@ void CheckpointIO::read (const std::string & input_name)
 
       for (processor_id_type proc_id = begin_proc_id; proc_id < input_n_procs; proc_id += stride)
         {
-          std::ostringstream file_name_stream;
+          std::string file_name = input_name;
 
-          file_name_stream << name;
-
-          file_name_stream << "-" << (input_parallel ? input_n_procs : 1) << "-" << proc_id;
+          file_name += "-" +
+            std::to_string(input_parallel ?  input_n_procs : 1) +
+            "-" + std::to_string(proc_id);
 
           {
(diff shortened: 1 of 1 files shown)


src/mesh/checkpoint_io.C
```diff
@@ -618,17 +632,17 @@ void CheckpointIO::read (const std::string & name)
...
         {
-          std::ostringstream file_name_stream;
-
-          file_name_stream << name;
+          std::string file_name = input_name;
 
-          file_name_stream << "-" << (input_parallel ? input_n_procs : 1) << "-" << proc_id;
+          file_name += "-" +
+            std::to_string(input_parallel ?  input_n_procs : 1) +
+            "-" + std::to_string(proc_id);
```
permcody: I have a little more food for thought on the naming scheme for parallel files in general. I think we should consider filling so that all names are fixed width and naturally sort when doing a directory listing, etc. Right now when I do a listing on a big split I see 1, 10, 100, 1000, 2, 20, 200, etc... 

It's minor and doesn't really change anything. The fill could just be based on the width of n_procs so that computable immediately (e.g. You'd know immediately to look for a file named `foo.cpr-1000-0001` or `foo.cpr-800-001`.

roystgnr: I like it; I'll probably wait to get it in with the next PR though, so hopefully this PR can get slipped into the MOOSE upgrade this afternoon.

permcody:
This looks good - would like to test this out with MOOSE and add to the libMesh update.

PR #1545, opened 2017-12-18
Add wrapper in exodus_io to access mapping from the unique_id to the value of an elemental variable

### LINKED ISSUE #1531: Add function to ExodusII_IO to read additional variables from exodus
Rather than going the route described by #1439 whose resolution is still an open debate, we are interested - in the meantime - in saving additional information (e.g. material ID) directly in the exodus mesh and read it during mesh construction. Our app will construct a mapping from the `top_parent()->unique_id` of each element to a structure containing the corresponding value(s).

What we need to add to libMesh is a utility function in `exodusII_io.C` that, given a variable name `elemental_var_name` and a timestep, will construct a map from the `unique_id` to the value `elemental_var_name` takes on in that element.

Merge request to come.

@YaqiWang @jwpeterson

vincentlaboure:
Close #1531 

@YaqiWang @jwpeterson

### DIFF
--- a/include/mesh/exodusII_io.h
+++ b/include/mesh/exodusII_io.h
@@ -133,6 +133,32 @@ public:
                                unsigned int timestep=1);
 
   /**
+   * Given an elemental variable and a time step, returns a mapping from the
+   * elements (top parent) unique IDs to the value of the elemental variable at
+   * the corresponding time step index.
+   * Note that this function MUST only be called before renumbering!
+   * This function is essentially a wrapper for read_elemental_var_values from
+   * the exodus helper (which is not accessible outside this class).
+   * \param elemental_var_name Name of an elemental variable
+   * \param timestep The corresponding time step index
+   * \param unique_id_to_value_map The map to be filled
+   */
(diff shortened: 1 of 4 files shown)


YaqiWang:
Can we add another wrapper function to read global variables in an Exodus II file? Also can we merge this into our coming libMesh update?

jwpeterson:
> Can we add another wrapper function to read global variables in an Exodus II file? 

Yes, that should be fine. We already have `write_global_data()` so it would make sense to be able to read it as well!

The "helper" object design is a very old one that I think eventually should go away so that all the information is available from the `ExodusII_IO` object.

> Also can we merge this into our coming libMesh update?

Yes, I believe we are waiting a few days on that so this should be fine.

YaqiWang:
I like the idea of merging. If it is available, we can remove this stupid unnecessary read here https://github.com/idaholab/moose/blob/devel/framework/src/timesteppers/ExodusTimeSequenceStepper.C#L47.

permcody:
Merging since @jwpeterson's single comment has been addressed. I'll open a PR up in MOOSE to perform the libMesh update.

PR #1548, opened 2018-01-04
Updating/Fixing PetscDiffSolver reinit()

pbauman:
@roystgnr 

This is still not completely correct as it still fails for AMR problems, see grinsfem/grins#528. Therefore, labeling as do not merge for the time being.

#1304 was the initial fix for dealing with the PETSc-debug-mode-lock issue. There were other problems that prevented AMR runs within the FEMSystem framework to work; d9fae0c is the state that I thought should allow AMR runs to proceed correctly (if a bit more inefficiently), but produces the problem described in grinsfem/grins#528.

Note that grinsfem/grins#528 will run correctly at d9fae0c if I run with PETSc opt and remove the change from #1304, that is we enforce constraints on the whole solution vector and not just current_local_solution in the residual and Jacobian functions (of course, PETSc will error in PETSc debug mode).

Commits  6092e29  e41672b 20b9b7e is basically cleanup/efficiency (and what I actually did first before hitting the AMR snag) and can probably be squashed together once we get to the bottom of this, but being overly pedantic now to try and see what's going on.

### DIFF
--- a/examples/adjoints/adjoints_ex5/adjoints_ex5.C
+++ b/examples/adjoints/adjoints_ex5/adjoints_ex5.C
@@ -652,11 +652,12 @@ int main (int argc, char ** argv)
                    << sensitivity_0_0
                    << std::endl;
 
-      // Hard coded assert to ensure that the actual numbers we are
+      // Hard coded test to ensure that the actual numbers we are
       // getting are what they should be
       // The 2e-4 tolerance is chosen to ensure success even with
       // 32-bit floats
-      libmesh_assert_less(std::abs(sensitivity_0_0 - (-5.37173)), 2.e-4);
+      if(std::abs(sensitivity_0_0 - (-5.37173)) >= 2.e-4)
+        libmesh_error_msg("Mismatch in sensitivity gold value!");
 
 #ifdef NDEBUG
     }
--- a/examples/adjoints/adjoints_ex5/heatsystem.C
+++ b/examples/adjoints/adjoints_ex5/heatsystem.C
@@ -164,6 +164,8 @@ void HeatSystem::perturb_accumulate_residuals(ParameterVector & parameters_in)
 {
   const unsigned int Np = parameters_in.size();
 
+  this->update();
+
   for (unsigned int j=0; j != Np; ++j)
     {
       Number old_parameter = *parameters_in[j];
--- a/include/solvers/petsc_diff_solver.h
+++ b/include/solvers/petsc_diff_solver.h
@@ -93,6 +93,13 @@ protected:
    * Nonlinear solver context
    */
   SNES _snes;
+
+private:
+
+  /**
+   * Common helper function to setup PETSc data structures
+   */
+  void setup_petsc_data();
 };
 
 } // namespace libMesh
--- a/include/solvers/petsc_diff_solver.h
+++ b/include/solvers/petsc_diff_solver.h
@@ -69,14 +69,12 @@ public:
   virtual void reinit () libmesh_override;
 
   /**
-   * The initialization function.  solve() calls this to create
-   * a new SNES context.
+   * The initialization function.
    */
   virtual void init () libmesh_override;
 
   /**
-   * The clear function.  solve() calls this to destroy
-   * a used SNES context.
+   * The clear function. Called if we reinit or when we destroy this object.
    */
   void clear ();
 
--- a/include/systems/diff_system.h
+++ b/include/systems/diff_system.h
@@ -94,7 +94,13 @@ public:
 
   /**
    * Prepares \p matrix and \p rhs for matrix assembly.
-   * Users should not reimplement this
+   * Users should not reimplement this.
+   * Note that in some cases only
+   * \link current_local_solution \endlink is used during assembly,
+   * and, therefore, if \link solution \endlink has been altered
+   * without \link update() \endlink being called, then the
+   * user must call \link update() \endlink before calling
+   * this function.
    */
   virtual void assemble () libmesh_override;
 
@@ -121,6 +127,12 @@ public:
   /**
    * Assembles a residual in \p rhs and/or a jacobian in \p matrix,
    * as requested.
+   * Note that in some cases only
+   * \link current_local_solution \endlink is used during assembly,
+   * and, therefore, if \link solution \endlink has been altered
+   * without \link update() \endlink being called, then the
+   * user must call \link update() \endlink before calling
+   * this function.
    */
   virtual void assembly (bool get_residual,
                          bool get_jacobian,
--- a/include/systems/fem_system.h
+++ b/include/systems/fem_system.h
@@ -81,7 +81,13 @@ public:
   /**
    * Prepares \p matrix or \p rhs for matrix assembly.
    * Users may reimplement this to add pre- or post-assembly
-   * code before or after calling FEMSystem::assembly()
+   * code before or after calling FEMSystem::assembly().
+   * Note that in some cases only
+   * \link current_local_solution \endlink is used during assembly,
+   * and, therefore, if \link solution \endlink has been altered
+   * without \link update() \endlink being called, then the
+   * user must call \link update() \endlink before calling
+   * this function.
    */
   virtual void assembly (bool get_residual,
                          bool get_jacobian,
--- a/include/systems/implicit_system.h
+++ b/include/systems/implicit_system.h
@@ -256,6 +256,13 @@ public:
    * H_ij = (d^2 q)/(d p_i d p_j)
    * This Hessian is the output of this method, where for each q_i,
    * H_jk is stored in \p hessian.second_derivative(i,j,k).
+   *
+   * Note that in some cases only
+   * \link current_local_solution \endlink is used during assembly,
+   * and, therefore, if \link solution \endlink has been altered
+   * without \link update() \endlink being called, then the
+   * user must call \link update() \endlink before calling
+   * this function.
    */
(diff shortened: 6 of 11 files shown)


pbauman:
Ugh, the order of the commits displayed here is not what I had (I rebased and reordred them a bit). The order should be 
2b3df77 
561ab02 
d9fae0c 

One should be able to pull down these three and reproduce the problem I'm seeing.

Then 6092e29 etc.

pbauman:
One thought I'd had was that maybe the constraints weren't being applied to the old_nonlinear_solution vector in the TimeSolver, but I added the enforce_constraints_exactly call right after SNESSolve, so the whole solution should've been properly constrained before advance_timestep() is called called later and the copy made.

pbauman:
BTW, friendly reminder if one is going to test changes here against the GRINS PR: do `export INSTALL="/usr/bin/install -C" before configuring libMesh so GRINS doesn't even have to rebuild after libMesh make install if you don't change any libMesh headers.

roystgnr:
> Note that grinsfem/grins#528 will run correctly at d9fae0c if I run with PETSc opt and remove the change from #1304, that is we enforce constraints on the whole solution vector and not just current_local_solution in the residual and Jacobian functions (of course, PETSc will error in PETSc debug mode).

So... with only the current_local_solution constrained, there's effectively nothing preventing the PETSc solution vector from "drifting" at hanging node values.  We leave asymmetric_constraint_rows off in FEMSystem, so the hanging node dof equations will be "x = 0", which means they won't be updated by Newton steps... but we shouldn't *care*, since they're decoupled from the rest of the system and we're only looking at current_local_solution when doing each assembly.  Obviously the update-after-constrain code was broken, but I don't see how 2b3df77 and d9fae0c aren't enough to fix that.

Is there any possibility that we *aren't* looking at current_local_solution during assembly?  Could you save a copy of sys.solution, NaN out the original before assembly, then restore the copy after?  Then if we're really ignoring it the way we expect to be we'll still get the same suboptimal-for-some-other-reason convergence, but if we're using it inadvertently the residual and Jacobian will be full of NaNs.

pbauman:
> Could you save a copy of sys.solution, NaN out the original before assembly, then restore the copy after?

Excellent suggestion. I will give that a whirl and report back.

pbauman:
OK, so we are using System.solution somewhere. I see NANs in residual when I add to the residual function and NANs in Jacobian when I set in the Jacobian function (and not the residual function). Let the hunt begin. Suggestions are welcome. I tried `-fp_trap` with PETSc, but I have to use opt since we're altering PETSc's solution and therefore will error in petsc-debug, so I'm guessing there are no floating point traps turned on in PETSc opt. 

I'm embarassed to say that I know to how turn on floating point traps in a Fortran compiler, but I'm not sure about C++. I see [this](https://linux.die.net/man/3/feenableexcept) page, anyone have any experience trying it out?

pbauman:
> I see this page, anyone have any experience trying it out?

To clarify, I meant the feenableexcept() function all the way at the bottom of the page.

pbauman:
OK, I couldn't get the floating point trap to actually trap. However, I've at least narrowed to fishyness happening `FEMContext::pre_fe_reinit`. I verified that `current_local_solution` is valid after I set `solution` to NAN, and immediately after `pre_fe_reinit`, `print_element_solution` is showing NANs for all element dofs.

pbauman:
OK, I've pinned it down and am testing a fix.

TL;DR: `FEMSystem::assembly()` is calling `System::update()` which is blowing away the constraints we'd set in `current_local_solution` (but not in `solution`).

We'll need to discuss what exactly the fix should be. Will push a commit with my fix to set groundwork for discussion.

pbauman:
OK, I rebased on master, squashed together the helper function commits into adce594, basic `PetscDiffSolver::reinit()` fix in f316adb (this removed the segfault problem, but exposed the incorrect AMR constraint issues). The two fixes that I originally though should've solved the problem are separately in 49e4783 and 39fcbcd. The last commit 4addeaa just cosmetic, removing the stale commented out code.

The main remaining piece is in c8196de. Because of the way we have to approach PETSc's solve, in particular constraining only the current_local_solution, we cannot call `System::update()` in `FEMSystem::assembly()` when using a `PetscDiffSolver`. My solution was to remove the update() call and append where need in `NewtonSolver` (the only missing places I've detected). My rationale is in the commit log which I paste below:

> Nominally we need to do this, particularly in NewtonSolver, but with
PetscDiffSolver, we're not allowed to constrain the entire solution
vector (System.solution), only current_local_solution. Thus, if
we update in assembly, we wipe out the constraints we just applied
in current_local_solution. Thus, we move update() out of assembly().
Therefore, this requires calling update() on the assembly() calls in
NewtonSolver.

> While the does seem to violoate the principle of least surprise, this
also seems like the cleanest solution to me. assembly() is supposed
to be agnostic to the solver, to we don't want to put
checks/setters/whatever into assembly() to decide whether or not to
update. Additionally, it's not entirely unreasonable for assembly()
to require the current_local_solution to be prepared beforehand.
Finally, one of the main points of FEMSystem is to hide a lot of these
details from the user, so while this is a behavior change, it should
be hidden from the user entirely. Indeed, in testing this fix in
grinsfem/grins#528, no surprises were found: all tests passed, including
the new test that exposed this problem.

How do folks feel about this fix? I've hit GRINS with petsc-debug, both serial and parallel up to 16 cores (only ReplicatedMesh though), and everything is passing, so I think I've gotten this right.

pbauman:
Figures. I tested adjoints_ex3 locally, but not adjoints_ex5.

pbauman:
OK, so perhaps first example where my argument failed. assembly() calls in `HeatSystem::perturb_accumulate_residuals()` needed an update() call at the beginning. And this failed with both `NewtonSolver` and `PetscDiffSolver`. This looks like a special case, but I'm open to better suggestions for this fix.

I'll momentarily push the fix that should get this to pass (and a fix for adjoints_ex5, why is calling libmesh_assert on a runtime check??)

pbauman:
OK, for me, this would be good to merge. But want to at least have @roystgnr weigh in. Potentially also affects @vikramvgarg.

roystgnr:
Well, this definitely isn't perfectly backwards-compatible, as the adjoints_ex5 case caught.  The breakage won't touch most users, but we ought to *carefully* document it in diff_system.h, I think.

I can't see any better solution, though, personally.  No other minor fix seems possible - if we can't directly modify the raw PETSc-provided solution, then we either have to force our constraints to be enforced in the system solve (a major change, requiring either asymmetric constraint application or some weird new modification of constrained dof columns) or we have to condense constraints out (another major change) or we have to force user code to ignore that solution in lieu of a fixed copy.

Let me look things over and give Vikram a chance to do the same before we merge.  Thanks for figuring this out.

pbauman:
Actually @canesin was the one who initially fixed `PetscDiffSolver`, so this potentially could affect him as well.

roystgnr:
Is this going to require corresponding manual update() calls in all our ImplicitSolver:: parameter sensitivity methods?

pbauman:
Honestly, I'm not sure. I don't think so because we're not doing sensitivity things until forward solve is done, but I'm certainly not experienced enough with that call stack (it's been... 4+ years since I played with... good lord) to swear to it.

roystgnr:
That's a good point.  We're not perturbing the solution there, we're perturbing parameters.  Similarly in error estimator code... probably.

I think I'm going to want to grep through the whole library looking at each assemble+assembly call before we merge this.  :-P  And even if everything is kosher for our own uses, we might want to add documentation regarding which methods expect solution+current_local_solution to be in sync, for others' benefit.

pbauman:
> I think I'm going to want to grep through the whole library looking at each assemble+assembly call before we merge this. :-P

Oh man. I forgot there's both assemble and assembly.

> And even if everything is kosher for our own uses, we might want to add documentation regarding which methods expect solution+current_local_solution to be in sync, for others' benefit.

Agreed. I'll drop in a documentation commit to get that started.

roystgnr:
> We're not perturbing the solution there, we're perturbing parameters. 

Well, *mostly*.

linear_system.C, assembly calls at lines 1387 and 1326 ought to be preceded by an update() call.  Those are both in qoi_parameter_hessian(), where we use a couple finite differences that perturb the solution using first-derivative sensitivity solutions.

Other than that, I don't see a problem.  Add those and the documentation (just a bit of "current_local_solution should be up to date, using update() if necessary, before this is called" in the docs for the assembly and sensitivity methods, that ought to do it?) and then :+1: to merge.

roystgnr:
Oh, and I chatted with @vikramvgarg, he thinks that manual user calls to assemble/assembly are rare enough that we shouldn't stress too much about the breakage of removing the update() there.

pbauman:
OK, I believe I got all of them. I'll let @roystgnr take one last peek before merging.

roystgnr:
Thanks!  Sorry for the review delay; I've been out sick today.

roystgnr:
Did this mess up fem_system_ex2?  I seem to be getting convergence failures when the aux_system current_local_solution values get wiped.  I'm not sure how that would have slipped by Civet, though.

I'm trying pre-PR now to see if this is just a red herring.

roystgnr:
No, still seeing a failure pre-merge.  Not sure how this isn't triggering CI problems.

pbauman:
I'll see if I can reproduce.

pbauman:
I see the failure as well and am similarly puzzled why `make check` isn't failing on CI, because it's reporting `FAIL` for that particular case for me.

PR #1550, opened 2018-01-06
minor corrections for real-valued numbers

BalticPinguin:
Hi,
I finally made an installation with real numbers instead of complex ones, leading to some (mostly minor) issues, I found:
- the only real issue is in the example eigenproblem_ex3: One cannot run a calculation with the shift-and-invert scheme without setting a **target** "position of spectrum".
- Also, I tested it with the complex-number installation of mine and it worked fine, so maybe one can think of loosening the condition for not running it ?
 - The change `src/fe/fe_map.C` is just to avoid a compiler warning.

The second point is also a general question concerning the preferred policy: Do you rather disable examples in case of any doubt or should these statements be as loose as possible to run a maximum number of tests respectively? 
I think, there are several statements which might be too strict as e.g. 
```
#ifdef LIBMESH_DEFAULT_SINGLE_PRECISION
  // SLEPc currently gives us a nasty crash with Real==float
  libmesh_example_requires(false, "--disable-singleprecision");
#endif
```
from miscellaneous_ex14.C which I actually never checked (I think, I took it from misc_ex1 at some point; but it isn't there any more.)


examples/eigenproblems/eigenproblems_ex3/eigenproblems_ex3.C
```diff
@@ -110,9 +110,9 @@ int main (int argc, char ** argv)
   libmesh_example_requires(false, "--disable-singleprecision");
 #endif
 
-#ifdef LIBMESH_USE_COMPLEX_NUMBERS
-  // SLEPc currently gives us an "inner product not well defined" with
-  // Number==complex
+#ifdef LIBMESH_USE_COMPLEX_NUMBERS && SLEPC_VERSION_LESS_THAN(3,6;2)
```
jwpeterson: That semicolon between 6 and 2 should be a comma!

examples/eigenproblems/eigenproblems_ex3/eigenproblems_ex3.C
```diff
@@ -110,9 +110,9 @@ int main (int argc, char ** argv)
...
+  // SLEPc may give an "inner product not well defined" with
+  // Number==complex; but for me it worked fine with vs 3.6.2
   libmesh_example_requires(false, "--disable-complex");
```
jwpeterson: You may want to update the help string here to "--disable-complex and SLEPc >= 3.6.2"

examples/eigenproblems/eigenproblems_ex3/eigenproblems_ex3.C
jwpeterson: Also isn't the `&&` logic wrong here? You want to trigger the `libmesh_example_requires` statement if complex numbers are enabled *or* SLEPc is too old, yes?

jwpeterson:
We don't yet have an optional SLEPc-enabled test for libmesh PRs but it should be possible to add one... I will look into it.

examples/eigenproblems/eigenproblems_ex3/eigenproblems_ex3.C
BalticPinguin: oh, you are absolutely right; strange that it compiled...

No, the logic should be right; it was stated that it doesn't work with complex numbers; but in my case (I checked with 3.6.2) it works fine. I'll check with 3.8.3 as well, but I assume that it works for newer versions so it should be skipped only for older versions and if complex numbers are used.

examples/eigenproblems/eigenproblems_ex3/eigenproblems_ex3.C
BalticPinguin: sure, this is of course better.

BalticPinguin:
... is there currently no test with SLEPc? (I actually had the impression that it is needed for most cases; but obviously I am biased).

BalticPinguin:
... What do you think, is it reasonable to check further statements of this kind and see, if there are some more that can be removed? If you don't think it is useful, I'll use my time more meaningful.

(changes will be added soon.)

examples/eigenproblems/eigenproblems_ex3/eigenproblems_ex3.C
jwpeterson: OK

jwpeterson:
> ... is there currently no test with SLEPc? (I actually had the impression that it is needed for most cases; but obviously I am biased).

There are tests, but they were previously not being run in any of our continuous integration configurations because SLEPc was not available. We've now added a SLEPc configuration.

jwpeterson:
> What do you think, is it reasonable to check further statements of this kind and see, if there are some more that can be removed? If you don't think it is useful, I'll use my time more meaningful.

You are referring to the `libmesh_example_requires()` statements? I think it would be good to keep them up to date so if you see one that's wrong, feel free to change it.

BalticPinguin:
Sorry for the long delay-time; but finally I managed to rebase.
I hope, it looks fine by now.

PR #1552, opened 2018-01-09
Make Parameters::Value public to allow iterators to use the Value type

### LINKED ISSUE #1551: Parameters::Value should be public
The `Value` class is currently defined as a private class within the Parameters object. However, the public API, namely the `Parameters::iterator`, use `Value *`, which causes difficulty when trying to iterate over the parameters.

aeslaughter:
(closes #1551)

### DIFF
--- a/include/utils/parameters.h
+++ b/include/utils/parameters.h
@@ -160,8 +160,6 @@ public:
    */
   void print (std::ostream & os=libMesh::out) const;
 
-private:
-
   /**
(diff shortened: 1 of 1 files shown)


jwpeterson:
This was an inconsistency in the interface exposed by trying to use the `begin()`/`end()` API of `Parameters` in a range-based for loop. These APIs explicitly depend on the `Parameters::Value` class, so I think this is a reasonable fix.

PR #1553, opened 2018-01-10
Allow users to provide separate functions for fd and mffd evaluations

lindsayad:
Petsc allows different functions to be set with:

- SNESSetFunction
- MatMFFDSetFunction
- MatFDColoringSetFunction

We have a use case in MOOSE with mechanical contact where we may want a contact update to be included in the function passed to `SNESSetFunction` (e.g. non-linear residual evaluations) but not when forming the Jacobian or computing Jacobian-vector products during the linear solve. Thus it would be nice to have a libMesh interface where we can pass different functions depending on the context.

### DIFF
--- a/include/solvers/petsc_nonlinear_solver.h
+++ b/include/solvers/petsc_nonlinear_solver.h
@@ -41,6 +41,7 @@ extern "C"
   PetscErrorCode __libmesh_petsc_snes_monitor (SNES, PetscInt its, PetscReal fnorm, void *);
   PetscErrorCode __libmesh_petsc_snes_residual (SNES, Vec x, Vec r, void * ctx);
   PetscErrorCode __libmesh_petsc_snes_fd_residual (SNES, Vec x, Vec r, void * ctx);
+  PetscErrorCode __libmesh_petsc_snes_mffd_interface (void * ctx, Vec x, Vec r);
   PetscErrorCode __libmesh_petsc_snes_mffd_residual (SNES, Vec x, Vec r, void * ctx);
 #if PETSC_RELEASE_LESS_THAN(3,5,0)
   PetscErrorCode __libmesh_petsc_snes_jacobian (SNES, Vec x, Mat * jac, Mat * pc, MatStructure * msflag, void * ctx);
@@ -155,6 +156,11 @@ public:
    */
   void use_default_monitor(bool state) { _default_monitor = state; }
 
+  /**
+   * Whether the Jacobian is matrix-free
+   */
+  bool mf;
+
 protected:
   /**
    * Nonlinear solver context
@@ -206,6 +212,7 @@ protected:
 private:
   friend PetscErrorCode __libmesh_petsc_snes_residual (SNES snes, Vec x, Vec r, void * ctx);
   friend PetscErrorCode __libmesh_petsc_snes_fd_residual (SNES snes, Vec x, Vec r, void * ctx);
+  friend PetscErrorCode __libmesh_petsc_snes_mffd_interface (void * ctx, Vec x, Vec r);
   friend PetscErrorCode __libmesh_petsc_snes_mffd_residual (SNES snes, Vec x, Vec r, void * ctx);
 #if PETSC_RELEASE_LESS_THAN(3,5,0)
   friend PetscErrorCode __libmesh_petsc_snes_jacobian (SNES snes, Vec x, Mat * jac, Mat * pc, MatStructure * msflag, void * ctx);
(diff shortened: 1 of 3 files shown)


roystgnr:
Hmm... should we have our underlying interface use functor objects (presumably of NonlinearImplicitSystem::ComputeResidual type) rather than raw C function pointers?  C function pointers can be used in a functor interface via a shim, but C++ functors can't be used in a C function pointer interface without a ```void*``` argument, which the PETSc API has but which your fd_residual and mffd_residual declarations lack.

We already made this mistake with residual vs residual_object - if the latter had come first then the former wouldn't have needed to exist.

pbauman:
> Hmm... should we have our underlying interface use functor objects (presumably of NonlinearImplicitSystem::ComputeResidual type) rather than raw C function pointers?

FWIW, this would be my preference.

lindsayad:
K, I'll work on that

jwpeterson:
It feels like there's quite a bit of code duplication in the new `__libmesh_petsc_snes_fd_residual()` and `__libmesh_petsc_snes_mffd_residual()` extern C functions and the original `__libmesh_petsc_snes_residual()`. It would be good if a future PR could combine some of the shared functionality.

I also just remembered that we chose double leading underscore to "fake" namespace these extern C functions, but this was before we realized that all the double leading underscore identifiers are reserved to the implementation. I'm less worried about name collisions than I am about just propagating bad coding practices, so that definitely doesn't have to be addressed by this PR.

lindsayad:
The failing check looks to be unrelated to this PR?
```
/tmp/tmpjnYEUr: line 279: FEMPUTER_N_MAKE_THREADS: undefined
```

pbauman:
Oops, yes, it's unrelated to the PR (typo in my recipe).

src/solvers/petsc_nonlinear_solver.C
```diff
@@ -39,6 +39,68 @@
 
 namespace libMesh
 {
+class ResidualContext
```
jwpeterson: :+1: I like this design, it seems cleaner than duplicating code.

lindsayad:
I think this is ready for further review

src/solvers/petsc_nonlinear_solver.C
```diff
@@ -652,14 +750,26 @@ PetscNonlinearSolver<T>::solve (SparseMatrix<T> &  jac_in,  // System Jacobian M
...
+
+  // Take care of case where the user specifies matrix-free jacobian
+  Mat J;
+  if (this->mf)
+  {
+    ierr = MatCreateSNESMF(_snes, &J);
+    LIBMESH_CHKERR(ierr);
+    ierr = MatMFFDSetFunction(J, __libmesh_petsc_snes_mffd_interface, this);
+    LIBMESH_CHKERR(ierr);
+    ierr = SNESSetJacobian(_snes, J, 0, 0, 0);
```
jwpeterson: Looks like we need one more `LIBMESH_CHKERR` here for consistency.

jwpeterson:
> I think this is ready for further review

:+1: for me, I'll let @roystgnr comment though, since he requested the functor approach in the first place.

include/solvers/petsc_nonlinear_solver.h
```diff
@@ -153,6 +158,18 @@ class PetscNonlinearSolver : public NonlinearSolver<T>
...
   void use_default_monitor(bool state) { _default_monitor = state; }
 
+  /**
+   * Petsc solve type
+   */
+  enum SolveType
+  {
+    MF_OPERATOR, ///< Preconditioned Jacobian-free Newton
+    MF,          ///< Jacobian-free Newton
+    STANDARD     ///< Newton with explicit matrix
```
jwpeterson: Should this be called NEWTON?

include/solvers/petsc_nonlinear_solver.h
```diff
@@ -153,6 +158,18 @@ class PetscNonlinearSolver : public NonlinearSolver<T>
...
+  };
+
+  SolveType solve_type;
```
jwpeterson: I think I'd prefer a protected variable behind a setter/getter interface, to be consistent with the rest of the `PetscNonlinearSolver` interface...

include/solvers/petsc_nonlinear_solver.h
jwpeterson: I see. How about "Matrix Full" Newton, then we can abbreviate it with "MF"... oh wait.

roystgnr: "Newton" is a somewhat fuzzy term; I've even heard some insist that the typical way libMesh apps do solves be referred to only as "Newton-Krylov" or "quasi-Newton", to avoid implying that we're doing a solve to machine precision at every (quasi-)Newton step.

I'm happy either way but I think the tie-breaker would be consistency with MOOSE.

include/solvers/petsc_nonlinear_solver.h
roystgnr: Seconded.

roystgnr:
:+1:

Issue #1486, opened 2017-11-03
eigenproblems_ex3 error with SLEPc 3.8


dknez:
We get this error:
"Must select a target sorting criterion if using shift-and-invert"

This is explained on the PETSc mailing list here:
https://lists.mcs.anl.gov/mailman/htdig/petsc-users/2017-October/033649.html

jwpeterson:
Thanks for the info. SLEPc is suddenly a hot topic in our PRs/Issues this week.

jwpeterson:
By the way, @dknez are you already working on fixing this? 

@fdkong you were on the email thread, can you take a crack at it?

jwpeterson:
I can look at it but I need to get PETSc/SLEPc 3.8 builds going first.

dknez:
I don't have to work on it right now, so if someone else wants to do it first that would be welcome.

I think the fix just requires us to change the call to set_position_of_spectrum(SMALLEST_MAGNITUDE) so that we set a target value of 0 instead. Note that according to the PETSc thread, this fix applies to earlier versions of SLEPc as well, and it is considered a common mistake to use EPS_SMALLEST_MAGNITUDE along with shift-and-invert.

dknez:
Fixed with PR #1555.

PR #1555, opened 2018-01-10
Fix for issue #1486.

dknez:
Update to eigenproblems_ex3 to avoid an error with SLEPc-3.8. This update should still
work fine with earlier versions of SLEPc too.

### DIFF
--- a/examples/eigenproblems/eigenproblems_ex3/eigenproblems_ex3.C
+++ b/examples/eigenproblems/eigenproblems_ex3/eigenproblems_ex3.C
@@ -200,8 +200,8 @@ int main (int argc, char ** argv)
   // a generalized Hermitian problem.
   eigen_system.set_eigenproblem_type(GHEP);
 
-  // Order the eigenvalues "smallest first"
-  eigen_system.eigen_solver->set_position_of_spectrum(SMALLEST_MAGNITUDE);
+  // Set the target eigenvalue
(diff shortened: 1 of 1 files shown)


jwpeterson:
I think this was done over on #1550 as well? It isn't merged yet, though.

dknez:
Yes, good point, I didn't realize that. Should I cancel this PR?

jwpeterson:
> Should I cancel this PR?

I don't know, the patches are slightly different so let's be sure and get the right one...

dknez:
I think they both do the same thing. My version is a little more explicit in that I used the version of set_position_of_spectrum where we set specify TARGET_REAL, so I prefer my version, but it's not a big deal either way.

jwpeterson:
Let's go with yours, and either let @BalticPinguin rebase his or I can fix the conflict during merge...

dknez:
OK. Also, it's a bit easier having the fix isolated in one commit, as it is here.

Should I wait for the tests, or do we assume that it's fine given that @BalticPinguin's already passed the tests?

jwpeterson:
This example isn't run by the CI since we don't have SLEPc available, so yeah, I'm good to merge.

BalticPinguin:
I'm late, but completely fine with this.
I'll rebase my PR (currently have more urgent things, therefore this is a bit hanging).

PR #1556, opened 2018-01-11
Updates to reduced basis code related to recent unique_ptr changes.

dknez:
This is a small change to the reduced basis code. Change some functions (or add new versions of existing functions) to take raw pointers instead of unique_ptrs, for easier compatibility with existing code bases.


dknez:
@jwpeterson let me know if these changes are OK with you, since they're due to your recent unique_ptr updates.

PR #1560, opened 2018-01-12
Added comment to ReplicatedMesh::stitch_meshes

dknez:
The comment specifies how element and node IDs in the stitched mesh correspond
to element and node IDs in the original meshes. This should be considered part
of the specification of this method so that we can rely on it in user code.

### DIFF
--- a/include/mesh/replicated_mesh.h
+++ b/include/mesh/replicated_mesh.h
@@ -204,6 +204,14 @@ public:
    * If \p skip_find_neighbors is true, a faster stitching method is used, where the lists of
    * neighbors for each elements are copied as well and patched, without calling the time-consuming
    * find_neighbors() function.
+   * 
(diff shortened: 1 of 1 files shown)


dknez:
I wanted to add this comment to confirm that I can rely on the specified behavior for element IDs.

Now that I think about it, I think the comment is incorrect for node IDs. There will be a gap in node ID numbering in the stitched mesh due to any nodes that are stitched, and that gap will then be removed when we call prepare_for_use/renumber_nodes_and_elems.

I believe the comment is correct for element IDs, though, and that element IDs will not be changed by calling renumber_nodes_and_elems.

I'll update the comment to remove the part about node IDs.

dknez:
I have started relying on the behavior specified in this comment change, so I'm going to go ahead and merge.

PR #1561, opened 2018-01-12
Update copyright notice for 2018.

jwpeterson:
Testing this on the off chance that my perl script went haywire and did something crazy.


jwpeterson:
FWIW, It appears that we are basically following the [GNU.org guidelines for supplying copyright notices](https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html). In particular, we *could* try to maintain a more accurate list of copyright holders in each file:

>  Each contributor of nontrivial text is a copyright holder. In that case, you should always include a copyright notice in the name of "main" copyright holder of the file. 

Quotes added, the word "main" is not explicitly defined. The description goes on to say:

> You can also include copyright notices for other copyright holders as well, and this is a good idea for those who have contributed a large amount and for those who specifically ask for notices in their names. ... But you don’t have to include a notice for everyone who contributed to the file (which would be rather inconvenient).

PR #1562, opened 2018-01-16
Check for cyclic constraints.

dknez:
A simple example of a cyclic constraint is if dof 1 is constrained in terms of dof 0,
and dof 0 constrained in terms of dof 1. In this case libMesh will not resolve the
constraints correctly and we will get an incorrect solution.

This change asserts that cyclic constraints are not present in debug/devel mode, and
gives us the option to check for cyclic constraints and throw an error in opt mode.

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -433,6 +433,14 @@ public:
   void add_variable_group (const VariableGroup & var_group);
 
   /**
+   * Specify whether or not we perform an extra (opt-mode enabled) check
+   * for cyclic constraints. If a cyclic constraint is present then
+   * the system constraints are not valid, so if \p error_on_cyclic_constraint
+   * is true we will throw an error in this case.
+   */
+  void set_error_on_cyclic_constraint(bool error_on_cyclic_constraint);
+
+  /**
    * \returns The \p VariableGroup description object for group \p g.
    */
   const VariableGroup & variable_group (const unsigned int c) const;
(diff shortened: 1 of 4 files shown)


dknez:
@roystgnr this is a first attempt to check that the debug/devel mode implementation is what you had in mind? Is there anywhere else that we need to add an assertion, e.g. what about the node constraints part of all_gather_recursive_constraints?

Regarding what to do in opt mode: I gather that what you had in mind is to add a new method which is close to a copy/paste of DofMap::process_constraints, except that we only check for the presence of a cyclic constraint without actually modifying anything? I haven't done that for now since I first wanted to check that this first version with you, but I can update it shortly once I get some feedback.

roystgnr:
That's exactly what I had in mind for dbg/devel modes.

But yeah, I'd rather it be a separate method in opt mode, rather than a test of that option in every single constrained/constrainee pair.

Good catch on the node constraints!  They use the same algorithm and they'll have the same risk with user-added constraints.

dknez:
I just pushed an update to separate the opt mode check into a new function, let me know if that looks OK to you?

I have not attempted to add the node constraint check since that code looks a bit different and I'm not 100% sure what to do in that case, so I think it's better if I leave that to you, or to a subsequent PR.

dknez:
Are these real test failures, or not?

jwpeterson:
Nope, just the known fem_system_ex2 thing and some debug mode timeouts in MOOSE which happen from time to time.

dknez:
OK, thanks.

@roystgnr let me know if this is OK with you to merge?

roystgnr:
Looks great; thanks!

Issue #1559, opened 2018-01-12
git bisect fem_system_ex2


pbauman:
I went to a year ago and b36673c is passing. So will start `git bisect` there.

pbauman:
So it's all the way back to 0ef555e so far, without about 4 bisect steps to go. Any running bets on how close my git bisect good is the to offending commit?

pbauman:
For posterity, here's my script that was fed to `git bisect run`

<pre>
#!/bin/bash

MYTEST_LIBMESH_BUILD_DIR=/fry1/data/users/pbauman/research/libmesh/build/test-femsystemex2-fail
MYTEST_LIBMESH_INSTALL_DIR=/fry1/data/users/pbauman/research/libmesh/install/test-femsystemex2-fail

# Go to libMesh build directory, configure, make, make install
mkdir -p $MYTEST_LIBMESH_BUILD_DIR
cd $MYTEST_LIBMESH_BUILD_DIR
module load gcc mpich openblas petsc vtk boost hdf5 cppunit
rm -rf ./*
../../libMesh/configure --enable-everything --with-metis=PETSc --with-methods=devel --prefix=$MYTEST_LIBMESH_INSTALL_DIR
make -j 32
cd examples/fem_system/fem_system_ex2/
make -j 16 check
</pre>

jwpeterson:
This thread finally jogged my memory. I believe the error is coming from a TypeVector::inverse() call on a matrix of all zeros. At some point I changed this to throw an exception, whereas before it would just divide by zero...

jwpeterson:
Sorry for not commenting before. This test's VTK requirement means that it doesn't get run as much as it should... we should probably consider writing multiple output files so the test can be turned on in more configurations...

pbauman:
So.... git bisect said ec5c76f8 was the offending commit... which is exactly one after the one I randomly chose (and thought I tested...). I'm very suspicious, will report back.

jwpeterson:
You might test ae45822a81687b54202f902543058457b74c3a91. That was the commit where we first started throwing exceptions from `TypeVector::inverse()`.

jwpeterson:
> On the other hand, I don't think that explains this bug, because it's dying in the first assembly of the second time step, not in the middle of a solve.

IIRC the error was from this line:

[...]

PR #1564, opened 2018-01-17
Update Nanoflann

### LINKED ISSUE #1563: Update nanoflann.hpp
Please let's try to get a more up to date nanoflann header into libmesh. The one in contrib seems to be five years behind. I'll see if the current one works as a drop in replacement. The API has some subtle but useful changes.

dschwen:
Let's run this through the test system...

Closes #1563


jwpeterson:
FYI the GRINS-devel failures are not going to be related to your PR.

dschwen:
Thanks, I was a bit puzzled by those.

pbauman:
Yes, that is because I started covering the test mentioned #1559 and a fix hasn't been made yet.

pbauman:
If it would make everyone feel warmer and fuzzier and I can change that part of the recipe to allow to fail. I just didn't want it to get lost.

dschwen:
Btw. @jwpeterson this dumps the entire nanoflann distribution into contrib. It adds 130k lines, which seems excessive. Did nanoflann just grow that much or did you leave out stuff previously. Technically we just need the header and a handful of tests.

jwpeterson:
> Did nanoflann just grow that much or did you leave out stuff previously.

The original nanoflann import was ef53f149e6fa. It did not contain all the .dat files you are checking in, so maybe we don't really need those. The testing suite looks like it got a bit more extensive.

jwpeterson:
> If it would make everyone feel warmer and fuzzier and I can change that part of the recipe to allow to fail. I just didn't want it to get lost.

I think we can keep the red X for now, but we run the risk of introducing additional GRINS-devel failures without people noticing, simply because they assume it's always the fem_system_ex2 thing.

dschwen:
I threw out a ton of stuff (160k lines). We are not running the tests in `make check` (just the examples).

dschwen:
There is actually a libmesh check failing (Example fem_system_ex2).

jwpeterson:
> There is actually a libmesh check failing (Example fem_system_ex2).

Yes, this is the GRINS-devel failure. It only runs if you have VTK enabled, which you must have on your system.

pbauman:
If we squash f2d6c08 and 4a5ff6e, will that remove the files from the history so we don't bloat it?

dschwen:
@pbauman, yeah, I'll squash. Just put it as a separate commit while we're still discussing this.

pbauman:
Sounds good, just wanted to make sure.

jwpeterson:
> So is everyone fine with the pruning I did?

I think we have Nanoflann tests in MOOSE, so if this passes those I'd say the update is probably OK.

> Do we need to ad a notice that we removed files from the nanoflann upstream?

IANAL but I think it's OK to *not* include pieces of BSD 2-clause licensed software that you distribute, provided that you maintain the copyright notices, list of conditions, etc. as required by the license.

dschwen:
Ok. MOOSE passed, so I'll go ahead and squash.

PR #1565, opened 2018-01-18
refactor mesh splitter app into a function

rwcarlsen:
We plan on building in some automatic assistance for splitting and using distributed mesh in MOOSE and since the splitter app also has this functionality I figured refactoring it out into a function call that MOOSE and the splitter utility app can use would be worthwhile.

### DIFF
--- a/include/mesh/checkpoint_io.h
+++ b/include/mesh/checkpoint_io.h
@@ -35,6 +35,16 @@ namespace libMesh
 {
 // Forward declarations
 class Xdr;
+class CheckpointIO;
+
+/**
+ * split_mesh takes the given initialized/opened mesh and partitions it into nsplits pieces or
+ * chunks.  It returns a CheckpointIO object that can be used to write the mesh chunks into
+ * individual files (e.g. by calling checkpoint_obj.write(out_file_name)) - the number of files is
+ * equal to the number of chunks.  This function supports MPI parallelism and can be used with
+ * several MPI procs to speed up splitting.
+ */
+std::unique_ptr<CheckpointIO> split_mesh(MeshBase & mesh, unsigned int nsplits);
 
 /**
  * The CheckpointIO class can be used to write simplified restart
--- a/src/mesh/checkpoint_io.C
+++ b/src/mesh/checkpoint_io.C
@@ -38,12 +38,62 @@
 #include "libmesh/node.h"
 #include "libmesh/parallel.h"
 #include "libmesh/partitioner.h"
+#include "libmesh/metis_partitioner.h"
 #include "libmesh/remote_elem.h"
 #include "libmesh/xdr_io.h"
 #include "libmesh/xdr_cxx.h"
 
 namespace libMesh
 {
+
+namespace
+{
+// chunking computes the number of chunks and first-chunk-offset when splitting a mesh
+// into nsplits pieces using size procs for the given MPI rank.  The number of chunks and offset
+// are stored in nchunks and first_chunk respectively.
+void chunking(processor_id_type size, processor_id_type rank, unsigned int nsplits,
+              processor_id_type & nchunks, processor_id_type & first_chunk)
+{
+  if (nsplits % size == 0) // the chunks divide evenly over the processors
+    {
+      nchunks = nsplits / size;
+      first_chunk = nchunks * rank;
+      return;
+    }
+
+  int nextra = nsplits % size;
+  if (rank < nextra) // leftover chunks cause an extra chunk to be added to this processor
+    {
+      nchunks = nsplits / size + 1;
+      first_chunk = nchunks * rank;
+    }
+  else // no extra chunks, but first chunk is offset by extras on earlier ranks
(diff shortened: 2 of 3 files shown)


src/mesh/checkpoint_io.C
```diff
@@ -38,13 +38,62 @@
...
+{
+  MetisPartitioner partitioner;
+  partitioner.partition(mesh, 1);
+  partitioner.partition(mesh, nsplits);
+
+  processor_id_type my_num_chunks = 0;
+  processor_id_type my_first_chunk = 0;
+  chunking(mesh.comm().size(), mesh.comm().rank(), nsplits, &my_num_chunks, &my_first_chunk);
+
+  auto cpr = new CheckpointIO(mesh);
```
permcody: Use `libmesh_make_unique` here instead of using a raw allocation.

jwpeterson: Please use `libmesh_make_unique`. I have tried to remove as many raw `new` commands from the code as possible.

src/mesh/checkpoint_io.C
```diff
@@ -38,13 +38,62 @@
...
 #include "libmesh/xdr_io.h"
 #include "libmesh/xdr_cxx.h"
 
 namespace libMesh
 {
 
+// chunking computes the number of chunks and first-chunk-offset when splitting a mesh
+// into nsplits pieces using size procs for the given MPI rank.  The number of chunks and offset
+// are stored in nchunks and first_chunk respectively.
+void chunking(processor_id_type size, processor_id_type rank, unsigned int nsplits,
```
jwpeterson: `nchunks` and `first_chunk` should be passed by reference.

src/mesh/checkpoint_io.C
```diff
@@ -38,13 +38,62 @@
 #include "libmesh/node.h"
 #include "libmesh/parallel.h"
 #include "libmesh/partitioner.h"
+#include "libmesh/metis_partitioner.h"
 #include "libmesh/remote_elem.h"
 #include "libmesh/xdr_io.h"
 #include "libmesh/xdr_cxx.h"
 
 namespace libMesh
```
jwpeterson: The `chunking` helper function should be in an anonymous namespace.

src/mesh/checkpoint_io.C
```diff
@@ -38,13 +38,62 @@
...
+      *nchunks = nsplits / size;
+      // account for the case where nchunks is zero where we want max int
+      *first_chunk = std::max((int)((*nchunks + 1) * (nsplits % size) + *nchunks * (rank - nsplits % size)),
+                             (1 - (int)*nchunks) * std::numeric_limits<int>::max());
+    }
+}
+
+std::unique_ptr<CheckpointIO> splitmesh(MeshBase & mesh, unsigned int nsplits)
+{
+  MetisPartitioner partitioner;
```
jwpeterson: I think you should just use `mesh.partition(1)` etc. here instead of building your own hard-coded `MetisPartitioner`.

include/mesh/checkpoint_io.h
```diff
@@ -35,6 +35,15 @@ namespace libMesh
 {
 // Forward declarations
 class Xdr;
+class CheckpointIO;
+
+/**
+ * splitmesh takes the given initialized/opened mesh and partitions it into nsplits pieces or
```
jwpeterson: I'd prefer an underscore in the name, i.e. `split_mesh()` but don't feel too strongly about it.

roystgnr: Agreed on both counts.

include/mesh/checkpoint_io.h
```diff
@@ -35,6 +35,15 @@ namespace libMesh
...
+ * chunks.  It returns a CheckpointIO object that can be used to write the mesh chunks into
+ * individual files (e.g. by calling checkpoint_obj.write(out_file_name)).  This function supports
+ * MPI parallelism and can be used with several MPI procs to speed up splitting.
+ */
```
roystgnr: This comment should probably make the "N procs, M files" case support or lack thereof clearer.

src/mesh/checkpoint_io.C
roystgnr: Yeah, this is an artifact of how I did things in the splitter app, which I think was pulled from Derek's code?  It was a perfectly reasonable hack for a private app but it's been getting more strained with each upgrade.

roystgnr:
Do we have any test coverage on the splitter app? I've been using it to *create* test cases but I don't think it's invoked in CI anywhere.

rwcarlsen:
No tests I know of.  I did take the original chunking logic and tested it against the ``chunking`` function for several cases to make sure it produced the same chunks for various combinations of procs, chunks, rank - making sure it was equivalent.

src/mesh/checkpoint_io.C
rwcarlsen: done

Issue #1567, opened 2018-01-22
[Discussion]: Geometric Multigrid via PETSc DM


pbauman:
I'm opening this up as a catch-all thread for the design discussion for this interface. This is work myself, @roystgnr, and @bboutkov have been pecking away at for awhile over in [GRINS](https://github.com/grinsfem/grins) and we're starting to migrate what we've done into libMesh. 

The idea is to interact with the DM infrastructure in PETSc. By setting up this interaction, we'll interact with PETSc from the command line to setup/use solvers in the same way Matt Knepley does with his Plex examples (unstructured grids) or PETSc DMDA (structured grids). In particular, using Geometric Multigrid (GMG) as a preconditioner or solver and potentially only on particular solution blocks. The plan is to use the mesh hierarchy from libMesh and setup the needed data structures in PETSc and then we can just do this from the command line.

I'll reference this issue (and the discussion below) in PRs. I'll try and briefly describe what's needed, where we are, current bottlenecks, etc.

What's Needed:
Step 1. Setup PetscSection and PetscSF interaction. This is essentially DoF info and DoF partitioning. These alone will facilitate recursive fieldsplits at the command line (my understanding is that the current interface we use does not support the recursive fieldsplit). This will be step one of our migration.

Step 2. Setup interpolation (and possibly restriction) matrices between each of the meshes in the hierarchy and then hooks into PETSc to supply these. The first part of this (creating actual matrices from `GenericProjector`) was migrated by @roystgnr, based on @bboutkov's work, in #1506. We still need to provide those matrices (and function hooks) into the PETSc DM infrastructure. This will facilitate "Galerkin" multigrid.

Step 3. Setup interface to facilitate nonlinear multigrid ("Full Approximation Scheme", FAS). This will require actually setting functions for assembling residual (and possibly Jacobian) on each mesh level. We haven't really experimented this with yet.

Status:
Step 1. I'm going to open a PR for this shortly. 

Step 2. This will be the next step of the migration. We've experimented with this and need to clean it up; the clean up will probably be "interactive" within the PR. The support we have so far is explicitly traversing the mesh hierarchy and constructing these matrices on the fly. The code does correctly support `DistributedMesh` for most cases (periodic boundary conditions are the current sticking point), but the scaling sucks. But we get expected solver behavior on Laplace for 1D, 2D, and 3D, for quads, triangles, hexes, and tets.

[...]

PR #1568, opened 2018-01-22
PetscDMWrapper Part 1

pbauman:
This implements Step 1 in the discussion in #1567.

I've marked this as do not merge until I can add testing into both libMesh and GRINS. I did testing way back when I wrote this code, but it was never formalized.

The testing to be added here is 1. The command line options for fieldsplit in this code path (both serial and parallel) and 2. recursive fieldsplit that I believe the current interface doesn't support. I'm thinking the Laplace/LAGRANGE_VEC example (where we have 3 decouple Laplace equations) and a incompressible thermal flow regression in GRINS. Also, I need to check the min PETSc version. Changelogs in PETSc aren't super helpful. I recall starting this work at PETSc 3.6.X, so was thinking I'll try that. Any urgent need to test earlier? Otherwise I'll just restrict this to PETSc 3.6.4 or later.

And dammit I just noticed I forgot documentation on the `init_and_attach_petscdm` function so I need to add that too.

Anyway, want to let folks start commenting on the design while I get tests into build system since I'm pretty confident this code will work correctly (famous last words...).

### DIFF
--- a/src/solvers/petsc_diff_solver.C
+++ b/src/solvers/petsc_diff_solver.C
@@ -230,6 +230,8 @@ void PetscDiffSolver::clear()
 
   int ierr = LibMeshSNESDestroy(&_snes);
   LIBMESH_CHKERR(ierr);
+
+  _dm_wrapper.clear();
 }
 
 
@@ -350,18 +352,29 @@ void PetscDiffSolver::setup_petsc_data()
       LIBMESH_CHKERR(ierr);
     }
 
+  bool use_petsc_dm = libMesh::on_command_line("--use_petsc_dm");
+
+  // This needs to be called before SNESSetFromOptions
+  if (use_petsc_dm)
+    this->_dm_wrapper.init_and_attach_petscdm(_system, _snes);
+
   ierr = SNESSetFromOptions(_snes);
   LIBMESH_CHKERR(ierr);
 
-  KSP my_ksp;
-  ierr = SNESGetKSP(_snes, &my_ksp);
-  LIBMESH_CHKERR(ierr);
+  // If we're not using PETSc DM, let's keep around
+  // the old style for fieldsplit
+  if (!use_petsc_dm)
+    {
+      KSP my_ksp;
+      ierr = SNESGetKSP(_snes, &my_ksp);
+      LIBMESH_CHKERR(ierr);
 
-  PC my_pc;
-  ierr = KSPGetPC(my_ksp, &my_pc);
-  LIBMESH_CHKERR(ierr);
+      PC my_pc;
+      ierr = KSPGetPC(my_ksp, &my_pc);
+      LIBMESH_CHKERR(ierr);
 
-  petsc_auto_fieldsplit(my_pc, _system);
+      petsc_auto_fieldsplit(my_pc, _system);
+    }
 }
 
 } // namespace libMesh
--- a/src/solvers/petsc_dm_wrapper.C
+++ b/src/solvers/petsc_dm_wrapper.C
@@ -32,6 +32,24 @@
 namespace libMesh
 {
 
+PetscDMWrapper::~PetscDMWrapper()
+{
+  this->clear();
+}
+
+void PetscDMWrapper::clear()
+{
+  // This will also Destroy the attached PetscSection and PetscSF as well
+  // Destroy doesn't free the memory, but just resets points internally
+  // in the struct, so we'd still need to wipe out the memory on our side
+  for( auto dm_it = _dms.begin(); dm_it < _dms.end(); ++dm_it )
+    DMDestroy( dm_it->get() );
+
+  _dms.clear();
+  _sections.clear();
+  _star_forests.clear();
+}
+
 void PetscDMWrapper::init_and_attach_petscdm(const System & system, SNES & snes)
 {
   START_LOG ("init_and_attach_petscdm", "PetscDMWrapper");
--- a/src/solvers/petsc_dm_wrapper.C
+++ b/src/solvers/petsc_dm_wrapper.C
@@ -207,6 +207,38 @@ void PetscDMWrapper::check_section_n_dofs_match( const System & system, PetscSec
   libmesh_assert_equal_to(total_n_dofs,(PetscInt)system.n_local_dofs());
 }
 
+PetscInt PetscDMWrapper::find_dof_rank( unsigned int dof, const DofMap& dof_map ) const
+{
+  libmesh_assert_greater_equal( dof, dof_map.first_dof( 0 ) );
+  libmesh_assert_less( dof, dof_map.end_dof(dof_map.comm().size()-1) );
+
+  // dofs are in order on each processor starting from processor 0, so we can
+  // use a binary search
+  unsigned int max_rank = dof_map.comm().size();
+  unsigned int current_rank = max_rank/2;
+  unsigned int min_rank = 0;
+
+  do
+    {
+      if( dof >= dof_map.first_dof(current_rank) )
+        {
+          min_rank = current_rank;
+          current_rank = (max_rank + min_rank)/2;
+        }
+      else
+        {
+          max_rank = current_rank;
+          current_rank = (max_rank + min_rank)/2;
+        }
+    }
+  while( max_rank - min_rank > 1 );
+
+  libmesh_assert_less( dof, dof_map.end_dof(current_rank) );
+  libmesh_assert_greater_equal( dof, dof_map.first_dof(current_rank) );
+
+  return current_rank;
+}
+
 } // end namespace libMesh
 
 #endif // LIBMESH_HAVE_PETSC
--- a/src/solvers/petsc_dm_wrapper.C
+++ b/src/solvers/petsc_dm_wrapper.C
@@ -1,0 +1,22 @@
+// The libMesh Finite Element Library.
+// Copyright (C) 2002-2018 Benjamin S. Kirk, John W. Peterson, Roy H. Stogner
+
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+#include "libmesh/libmesh_common.h"
+
+#ifdef LIBMESH_HAVE_PETSC
+
+#endif // LIBMESH_HAVE_PETSC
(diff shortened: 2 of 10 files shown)


pbauman:
And also need to setup the stub correctly in the first commit (the .C stub is missing and sucked into 
e5e83e3 which will cause 5240870 to not compile).

jwpeterson:
Is there any relation between this new DM stuff and the existing DM stuff ([1](https://github.com/libMesh/libmesh/blob/master/src/solvers/petscdmlibmesh.C), [2](https://github.com/libMesh/libmesh/blob/master/src/solvers/petscdmlibmeshimpl.C))? For example, could we get rid of the old stuff?

pbauman:
I don't interact at all with the old stuff. I don't use it (I'll use the stuff in this PR), but does MOOSE use it? I would suggest it can be removed if not, but we should probably poll libmesh-users before we remove it since maybe the original authors might still be using it in some context.

jwpeterson:
@fdkong may be interested in this.

pbauman:
OK, rebased on master, and fixed the stub commit so all commits in the PR will compile. Still need testing and documentation for `init_and_attach_petscdm`.

src/solvers/petsc_dm_wrapper.C
```diff
@@ -0,0 +1,388 @@
...
+  unsigned int n_levels = 1;
+  this->init_dm_data(n_levels);
+
+  for(unsigned int level = 0; level < n_levels; level++)
+    {
+      DM & dm = this->get_dm(level);
+      PetscSection & section = this->get_section(level);
+      PetscSF & star_forest = this->get_star_forest(level);
+
+      ierr = DMShellCreate(system.comm().get(), &dm);
```
fdkong: I was wondering why we do not implement a native libmesh DM? The native DM will give us more flexibility on the setup of the solver and the preconditioner.  In PETSc, we have native DMs for DMPlex and DMDA. This allows us do more customized stuffs in the future. 

We could clean up the old DM stuff, and combine with this PETScDMWrapper. It should be easy.

I miss something here?

src/solvers/petsc_diff_solver.C
```diff
@@ -350,18 +352,29 @@ void PetscDiffSolver::setup_petsc_data()
...
+    {
+      KSP my_ksp;
+      ierr = SNESGetKSP(_snes, &my_ksp);
+      LIBMESH_CHKERR(ierr);
 
-  PC my_pc;
-  ierr = KSPGetPC(my_ksp, &my_pc);
-  LIBMESH_CHKERR(ierr);
+      PC my_pc;
+      ierr = KSPGetPC(my_ksp, &my_pc);
```
fdkong: It is not introduced by this PR.But  it would be great if we could get ride of this manual setup of the preconditoner.

fdkong:
@pbauman Could you possibly add an example illustrating the usage of FieldSplit based on the PETScWrapper in this PR?  I guess GMG is not working yet in this PR.

src/solvers/petsc_dm_wrapper.C
pbauman: > I was wondering why we do not implement a native libmesh DM instead of shell dm?

Can you elaborate on what you mean by "native" DM? Do you mean to essentially do a full subclassing of DM?

> The native DM will give us more flexibility on the setup of the solver and the preconditioner. In PETSc, we have native DMs for DMPlex and DMDA. This allows us do more customized stuffs in the future.

Can you elaborate on what flexibility and customization is gained vs. the shell? As far as I can tell, we will get the full solver flexibility (once the part 2 comes with GMG). What else would be enabled that not's enabled by the shell?

src/solvers/petsc_diff_solver.C
pbauman: My intent is to eventually (after several libMesh releases) remove the old `petsc_auto_fieldspilt` in favor of this.

pbauman:
> Could you possibly add an example illustrating the usage of FieldSplit based on the PETScWrapper in this PR?

Yes, forthcoming (hopefully tomorrow) as mentioned in the PR.

> I guess GMG is not working yet in this PR.

Correct, this is only step 1. Please see #1567 for a more elaborate/global discussion of where I'm going.

src/solvers/petsc_dm_wrapper.C
pbauman: The other reason I like the shell interface is it minimizes the PETSc API we have to touch since PETSc has no intent so support backward compatibility in API so the more API we touch, the larger maintenance headache it becomes.

fdkong: > Can you elaborate on what you mean by "native" DM? Do you mean to essentially do a full subclassing of DM?

I mean something like  https://github.com/petsc/petsc/blob/65b40d628b4e55462e59718b6e7e25b2e90f85aa/include/petsc/private/dmpleximpl.h#L175

We may have something similar: DM_Libmesh.

> 
> The native DM will give us more flexibility on the setup of the solver and the preconditioner. In PETSc, we have native DMs for DMPlex and DMDA. This allows us do more customized stuffs in the future.
> Can you elaborate on what flexibility and customization is gained vs. the shell? As far as I can tell, we will get the full solver flexibility (once the part 2 comes with GMG). What else would be enabled that not's enabled by the shell?

Personally I like a native DM more than the shell DM because I could add some internal data (or other crazy things definitely at the user-side) in "typedef struct" that may be useful for implementing the solver interfaces. For example, to construct a geometric DDM, we may need to implement the interface: dm->ops->createdomaindecomposition. If we have our own internal data (such mesh info or system) inside of DM, it is straightforward to feedback the indices to PETSc, otherwise, we have to figure out a way (maybe PetscObjectCompose()) to bring in mesh or system info because the DDM interface does not have AppCtx.

Again, I could be wrong. All these stuffs are based on my experiences. I implemented GMG (a pure petsc-based code) for my  PhD thesis using the native DM.

src/solvers/petsc_diff_solver.C
fdkong: In MOOSE, we have our own fieldsplit preconditioner implementation. The preconditioner allows us split the matrix based on not only the physics field but also the geometry. This requires us to implement dm->ops->createfielddecomposition.   

I am guessing you are using the default petsc implementation for this. Then we can not construct  the fieldsplit preconditioner based on the geometry.

src/solvers/petsc_dm_wrapper.C
pbauman: OK, so it looks like the existing DM bits were moving towards a "native" DM.

> Personally I like a native DM more than the shell DM because I could add some internal data (or other crazy things definitely at the user-side) in "typedef struct" that may be useful for implementing the solver interfaces. For example, to construct a geometric DDM, we may need to implement the interface: dm->ops->createdomaindecomposition. If we have our own internal data (such mesh info or system) inside of DM, it is straightforward to feedback the indices to PETSc, otherwise, we have to figure out a way (maybe PetscObjectCompose()) to bring in mesh or system info because the DDM interface does not have AppCtx.

Shell provides access to `createdomaindecomposition` and takes a context (step 2 of this process uses that context hook) so we can feed `EquationSystem` and therefore everything that would be needed. http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/DM/DMShellSetContext.html

That said, I'm not opposed to a native DM, but what's there looks rather cumbersome. Perhaps we could talk off thread about the DM API and what you're looking for. But in the near term, I'd prefer to stick to a shell so we can keep this moving.

src/solvers/petsc_diff_solver.C
pbauman: You are correct, the field split access here is only for physics field split. Again, perhaps we (and whoever is interested) can talk off thread about how you use the domain decomposition so we can integrate that cleanly because, I believe, this would provide all you'd need to for a physics field split and not require anything downstream in MOOSE.

pbauman:
> Yes, forthcoming (hopefully tomorrow) as mentioned in the PR.

I do have a working test case comparing to PETSc ex62, just needs to be added to GRINS and ought to be able to adapt to one of the fem_system examples.

pbauman:
@roystgnr pointed out `dof_owner` should be able to replace my `find_dof_rank` function.

src/solvers/petsc_dm_wrapper.C
```diff
@@ -19,4 +19,68 @@
...
+#include "libmesh/elem.h"
+
+namespace libMesh
+{
+
+void PetscDMWrapper::set_point_range_in_section( const System & system, PetscSection & section)
+{
+  const MeshBase & mesh = system.get_mesh();
+
+  unsigned int pstart = 2^30;
```
pbauman: We need to make this not restricted

src/solvers/petsc_dm_wrapper.C
```diff
@@ -19,4 +19,68 @@
...
+  // If we're on a processor who coarsened the mesh to have no local elements,
+  // we should make an empty PetscSection. An empty PetscSection is specified
+  // by passing [0,0) to the PetscSectionSetChart call.
+  else
+    {
+      pstart = 0;
+      pend = 0;
+    }
+
+  PetscErrorCode ierr = PetscSectionSetChart(section, pstart, pend);
```
pbauman: I need to ask @knepley if the chart really want there to be number of points = pend-pstart since we could easily have pstart = 0, pend=1e8 and have only those 2 elements on the processor.

roystgnr:
:+1: after the dof_owner and unsigned-int-vs-PetscInt stuff is fixed, but before committing put up issues for fixing all the current half-baked bits: the "No DoFs on elements" assumption, the "No more than 1 DoF on each node" assumption.  None of those things should be hard to fix, but the question of what is the best way to fix them will depend on the PetscSectionSetChart issue.

roystgnr:
Oh, and wait a second - what do we need to do about SCALAR variables here?

pbauman:
> Oh, and wait a second - what do we need to do about SCALAR variables here?

Oh, good point. Those ought to be able to be their own "points" and they get 1 for each one. They always live on the processor with the largest rank IIRC? But you're correct, we'll need to account for those as well.

pbauman:
I need to not to forget to put PETSc version guards around this too.

src/solvers/petsc_dm_wrapper.C
fdkong: We usually assume that the number is consecutively increased from pstart to pend.

pbauman: OK, just finished talking to @knepley. He said it does need to be contiguous, but it doesn't need to be global. In fact, he said because the PetscSF stashes the tuple of local id, local proc id, they don't care really about the point number for solvers (which is all we're using this for). So we can manually construct our own local point numbering and it's fine as long as we're consistent when we set the dof ids for the point. I'll take a shot at working that in here before we merge.

roystgnr: Oh, seriously?  That sounds just fine, then.  Thank him for me.

We're probably going to be stuck creating a lookup table, aren't we?  libMesh ids are all global, but (except for dof numbers) aren't necessarily contiguous.

bboutkov:
As per our discussion, we also want init_and_attach_petscdm to pass in a non-const System as we will be walking the mesh hierarchy.

pbauman:
I'm closing this for now because I'm going to working on this and push repeatedly with branch cleanups etc and don't want unnecessarily keep triggering CI. I'll reopen once I'm done.

PR #1570, opened 2018-01-25
fix mesh splitting bug to actually partition the mesh

rwcarlsen:
Bug added in #1565.  Oops :-(


src/mesh/checkpoint_io.C
```diff
@@ -80,6 +80,7 @@ void chunking(processor_id_type size, processor_id_type rank, unsigned int nspli
 std::unique_ptr<CheckpointIO> split_mesh(MeshBase & mesh, unsigned int nsplits)
 {
   mesh.partition(1);
```
jwpeterson: There is a reason for this partiton-to-1 followed by actually partitioning paradigm, so I'll add a comment to that effect during merge.

PR #1571, opened 2018-01-26
Updates to the reduced basis code to handle constraints.

dknez:
If we have constraints, e.g. hanging node constraints, then we should perform matrix-vector multiplications using vectors that already have constraints enforced exactly, and matrices which have no constraints imposed on them. This commit adds support for this when "non-Dirichlet operators" are enabled.


dknez:
This only affects the reduced basis code, and it should have no effect by default (since store_non_dirichlet_operators is false by default), hence I will go ahead and merge.

PR #1572, opened 2018-01-29
Use PETSc API MatSNESMFSetReuseBase

fdkong:
the residual vector from nonlinear solver for MFFD computation.

We may like to wait until  MatSNESMFSetReuseBase is merged at PETSc side. 

https://bitbucket.org/petsc/petsc/pull-requests/845/added-matsnesmfsetreusebase-at-request-of/diff

### DIFF
--- a/src/solvers/petsc_nonlinear_solver.C
+++ b/src/solvers/petsc_nonlinear_solver.C
@@ -713,8 +713,6 @@ PetscNonlinearSolver<T>::solve (SparseMatrix<T> &  pre_in,  // System Preconditi
   {
       ierr = MatCreateSNESMF(_snes, &J);
       LIBMESH_CHKERR(ierr);
-      ierr = MatMFFDSetFunction(J, __libmesh_petsc_snes_mffd_interface, this);
-      LIBMESH_CHKERR(ierr);
       if (_solve_type == MF_OPERATOR)
         ierr = SNESSetJacobian(_snes, J, PETSC_NULL, PETSC_NULL, PETSC_NULL);
       else
@@ -737,7 +735,6 @@ PetscNonlinearSolver<T>::solve (SparseMatrix<T> &  pre_in,  // System Preconditi
             ierr = MatSetNullSpace(J, msp);
             LIBMESH_CHKERR(ierr);
           }
-
           ierr = MatNullSpaceDestroy(&msp);
           LIBMESH_CHKERR(ierr);
         }
@@ -760,7 +757,6 @@ PetscNonlinearSolver<T>::solve (SparseMatrix<T> &  pre_in,  // System Preconditi
             ierr = MatSetTransposeNullSpace(J, msp);
             LIBMESH_CHKERR(ierr);
           }
-
           ierr = MatNullSpaceDestroy(&msp);
           LIBMESH_CHKERR(ierr);
         }
@@ -782,12 +778,17 @@ PetscNonlinearSolver<T>::solve (SparseMatrix<T> &  pre_in,  // System Preconditi
             ierr = MatSetNearNullSpace(J, msp);
             LIBMESH_CHKERR(ierr);
           }
-
           ierr = MatNullSpaceDestroy(&msp);
           LIBMESH_CHKERR(ierr);
         }
     }
 #endif
+
+if (_solve_type != STANDARD)
+  {
+    ierr = LibMeshMatDestroy(&J);
+    LIBMESH_CHKERR(ierr);
+  }
 
   // Have the Krylov subspace method use our good initial guess rather than 0
   KSP ksp;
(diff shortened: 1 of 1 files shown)


fdkong:
@lindsayad, @jwpeterson

src/solvers/petsc_nonlinear_solver.C
```diff
@@ -715,6 +715,11 @@ PetscNonlinearSolver<T>::solve (SparseMatrix<T> &  pre_in,  // System Preconditi
       LIBMESH_CHKERR(ierr);
       ierr = MatMFFDSetFunction(J, __libmesh_petsc_snes_mffd_interface, this);
       LIBMESH_CHKERR(ierr);
+#if !PETSC_VERSION_LESS_THAN(3, 8, 3)
```
jwpeterson: If this was a recent change, it won't be in 3.8.3. 

We'll probably have to test `!PETSC_VERSION_LESS_THAN(3, 8, 4)` if you are sure it will be in the next patch level release...

fdkong: Let us hold this PR until PETSc-side is done.

lindsayad:
I can verify that this does eliminate the extra residual call at the beginning of the linear solve 👍

lindsayad:
Those tests will fail with this PR until the functors are added in MOOSE

src/solvers/petsc_nonlinear_solver.C
```diff
@@ -760,9 +763,10 @@ PetscNonlinearSolver<T>::solve (SparseMatrix<T> &  pre_in,  // System Preconditi
             ierr = MatSetTransposeNullSpace(J, msp);
             LIBMESH_CHKERR(ierr);
           }
-
           ierr = MatNullSpaceDestroy(&msp);
           LIBMESH_CHKERR(ierr);
+
+          if ()
```
lindsayad: ?

src/solvers/petsc_nonlinear_solver.C
```diff
@@ -782,13 +778,22 @@ PetscNonlinearSolver<T>::solve (SparseMatrix<T> &  pre_in,  // System Preconditi
...
         }
     }
 #endif
 
+if (_solve_type != STANDARD)
+  {
+#if PETSC_VERSION_LESS_THAN(3,2,0)
+    ierr = MatDestroy(J);
+#else
+    ierr = MatDestroy(&J);
```
jwpeterson: For this, please use the `LibMeshMatDestroy()` macro which is already set up to handle the API change.

jwpeterson:
Were you guys talking with @brianmoose about adding an optional CIVET test of newer PETScs for libmesh PRs? Seems like that would be handy for this one...

brianmoose:
First I heard of it, but it would be no problem. Just build PETSc master?

permcody:
Several years ago we used to not allow -snes_mf_operator on the command line. It's the whole point of using `solve_type = PJFNK`. The former isn't very user friendly. Try not to put that in your test inputs etc.

lindsayad:
We don't ever put it in test inputs. We do warn users if the specify `-snes_mf_operator` in the input file, but there are no warnings if they specify it on the command line. It might be a rare user who would put it on the command line, but I think we would still want our stuff to work even if that option is passed.

jwpeterson:
> Just build PETSc master?

In this case we are specifically interested in testing PETSc 3.8.4 or newer, but yeah I think master would be fine. And we'd want to compile libmesh, run "make check," and then run the MOOSE tests.

fdkong:
I still want to use -snes_mf_operator

fdkong:
> In this case we are specifically interested in testing PETSc 3.8.4 or newer, but yeah I think master would be fine. And we'd want to compile libmesh, run "make check," and then run the MOOSE tests.

Both PETSc-maint and PETSc-master are still tagged as 3.8.3 at this point.

jwpeterson:
@fdkong just talked about this offline, and I think the right check here is:
```
#if !PETSC_RELEASE_LESS_THAN(3, 8, 4)
```
On PETSc master, `PETSC_VERSION_RELEASE` is always 0. Therefore `PETSC_RELEASE_LESS_THAN(x, y, z)` always returns false on master, regardless of the values of x, y, and z.

jwpeterson:
👍  @fdkong and thanks @brianmoose for adding that optional test! Looks like it takes about 9 minutes to build PETSc master from scratch, so not too bad.

lindsayad:
The three failing tests with MOOSE master are a sign that Fande's petsc PR and this one are doing their jobs. Those tests just count residual evaluations. With this PR the residual counts are less!

jwpeterson:
When we exodiff, the gold value comes first. Is it the same for csvdiff? If so, then the gold had fewer iterations for this test:
```
outputs/iterative.csv: Running CSVDiffer.py
outputs/iterative.csv: In iterative_csv_out.csv: The values in column "iterations" don't match
outputs/iterative.csv: 	rel diff:  204.0 ~ 238.0 = 0.142857142857
outputs/iterative.csv: 
outputs/iterative.csv: ################################################################################
outputs/iterative.csv: Tester failed, reason: CSVDIFF
```

permcody:
That's critical information (which one's first). That needs to be printed
here. I'll open a ticket.

lindsayad:
No it's switched for csvdiff

permcody:
https://github.com/idaholab/moose/issues/10673

jwpeterson:
> No it's switched for csvdiff

Nice.

lindsayad:
So we can merge this?

roystgnr:
@brianmoose, could we do the same "just build the prerequisite on the fly" trick for an optional test with SLEPc too?  SLEPc configuration is trivial and build time is minimal once PETSc's been built first.

jwpeterson:
> On Feb 2, 2018, at 7:09 AM, roystgnr <notifications@github.com> wrote:
> 
> @brianmoose, could we do the same "just build the prerequisite on the fly" trick for an optional test with SLEPc too? SLEPc configuration is trivial and build time is minimal once PETSc's been built first.
> 

We already have an optional SLEPc test for libmesh PRs. Do you want to test SLEPc “master” as well? Do they have a public repo?

roystgnr:
Ha, false alarm.  For some reason I recalled us not having a SLEPc test at all.  No, any recent release ought to be fine.

brianmoose:
The current optional SLEPc test for libmesh just uses the SLEPc that comes with the MOOSE environment package.
We can easily build SLEPc master as well. We do that for all the MOOSE optional recipes that build with PETSc master ( I believe that is what @fdkong wanted).
Their repo: https://bitbucket.org/slepc/slepc

PR #1574, opened 2018-01-31
Add optional system_names arguments to ExodusII write_timestep methods.

dknez:
write_timestep and write_timestep_discontinuous call methods that take an optional system_names argument, so we should also have the optional system names argument for these methods.


dknez:
I'll go ahead and merge this since it doesn't change any existing behavior and adds a bit of extra flexibility.

roystgnr:
This looks good, thanks a lot!

In the future, though, remember that "is the obvious API choice" is also a criterion for quick merges, because today's "doesn't change any existing behavior" depends on yesterday's "doesn't create new behavior that will be hard to avoid changing".

PR #1575, opened 2018-01-31
Cache dof indices when computing sparsity

roystgnr:
This gives me about 85% speedup on the test case in
https://github.com/idaholab/moose/issues/9480

### DIFF
--- a/include/base/sparsity_pattern.h
+++ b/include/base/sparsity_pattern.h
@@ -87,9 +87,12 @@ private:
   const bool implicit_neighbor_dofs;
   const bool need_full_sparsity_pattern;
 
-  void handle_vi_vj(const Elem * partner,
-                    const std::vector<dof_id_type> & element_dofs_i,
-                    unsigned int vj);
+  void handle_vi_vj(const std::vector<dof_id_type> & element_dofs_i,
+                    const std::vector<dof_id_type> & element_dofs_j);
+
+  void sorted_connected_dofs(const Elem * elem,
+                             std::vector<dof_id_type> & dofs_vi,
+                             unsigned int vi);
 
 public:
 
--- a/src/base/sparsity_pattern.C
+++ b/src/base/sparsity_pattern.C
@@ -78,9 +78,23 @@ void _dummy_function(void) {}
 
 
 
-void Build::handle_vi_vj(const Elem * partner,
-                         const std::vector<dof_id_type> & element_dofs_i,
-                         unsigned int vj)
+void Build::sorted_connected_dofs(const Elem * elem,
+                                  std::vector<dof_id_type> & dofs_vi,
+                                  unsigned int vi)
+{
+  dof_map.dof_indices (elem, dofs_vi, vi);
+#ifdef LIBMESH_ENABLE_CONSTRAINTS
+  dof_map.find_connected_dofs (dofs_vi);
+#endif
+  // We can be more efficient if we sort the element DOFs into
+  // increasing order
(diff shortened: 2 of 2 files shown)


dschwen:
Does it also change the big O?

permcody:
image

roystgnr:
I'm afraid I don't yet understand the asymptotic complexity here.  The stuff I optimized was a loop over elements (1 in this case), over elements coupled to each (1), over target variables (N), over source variables (N), over nodes...

Hmm... and maybe there's another O(N) in the Node DofObject operations, post-Ben's-data-compression?  I need to look at that more carefully.

roystgnr:
There's an O(N_variable_groups) var_to_vg conversion in the DofObject operations, but libMesh defaults identify_variable_groups() to true, and MOOSE doesn't futz with that IIRC, and with that setting on, adding a thousand identically typed variables should give you N_variable_groups==1.

roystgnr:
Wait a minute, we're looking at SCALAR variables here, not variables stored on DofObjects.  And in the SCALAR_dof_indices helper the only loop is over variable order, which should be 1 for every variable.  So we're back to "I don't yet understand", except without the sliver of hope previously implied by "yet".

roystgnr:
Profiling says my remaining time is mostly being spent in handle_vi_vj and operator(), not in the dof_indices methods anymore, so maybe I just removed an O(N^2) term with a huge constant but there's an O(N^3) term with a small constant still hiding in one of those places?

roystgnr:
The std::equal_range search is guaranteed to be O(log2(N)), but here the range should be empty.

The std::lower_bound search that empties the range is guaranteed to be O(log2(N)), and we can probably short-circuit that in the common case (including all cases here) with a single O(1) test?  I'll try that and see if it helps for me.

SparsityPattern::sort_row() is a reverse bubble-sort, which looks insane in general (and probably needs to be reconsidered for 3D problems with lots of non-scalar variables...), but which in *this* case (an already sorted row with middle+1==end) should be O(1).

I'm just not seeing anything I'd expect to be cubic.

roystgnr:
No luck with that short-circuiting test, and there's no other obvious improvements to try, so I'd say let's merge this ASAP and I'll try more subtle changes later.

dschwen:
Sounds good. Thanks for working on it. I'm compiling MOOSE right now with this PR to update the plot on the MOOSE PR

roystgnr:
Great; thanks!  I'd very much like to have my current "big O(N^2) term fixed, smaller O(N^3) term untouched" theory confirmed or denied.

dschwen:
out

dschwen:
I don't know. It looks like you are reducing cost by a constant factor (while keeping an O(1) contribution that comes from the way I time it)

roystgnr:
Hm, it certainly does.  I'll keep hunting, then.  I could buy O(N^2 logN) but there's no way this should be O(N^3).

roystgnr:
Wait - what's "nonlinear"?  Not the same as "lagrange"?

jwpeterson:
> Wait - what's "nonlinear"?

I think it's the same as Lagrage. We often use "nonlinear" to mean "not auxiliary" when referring to variables.

dschwen:
Yeah sorry, I saw I messed up the label but I was too lazy to fix the plot.

roystgnr:
Going to merge this now - I've tried a dozen other possible optimizations with no improvement, so although it seems like there *must* be an O(N/log2(N)) improvement still possible somehow, I don't want the continuing search for it to delay this PR's ~85% improvement.

PR #1576, opened 2018-02-01
read/write checkpoint files to/from nprocs-suffixed directories

rwcarlsen:
When writing and reading checkpoint files, put sets of checkpoint files for a single object/split together in a subdirectory named ``[input_name]-split[n]``.  When reading checkpoint files, look for a subdirectory with the ``-split[n]`` suffix for the current/correct number of procs.  This also helps prevent running with a lower number of procs than cpr split files which previously use to run but just with part of the mesh missing.

For non-parallel mesh cases, the single checkpoint file is just written as normal - not into a subdirectory.

Edit:

This now always creates a subdirectory named ``[input_name]`` with subsubdirectories named for the available split configurations for the mesh - e.g. fullpaths of ``[input_name]/[n]``.  Header and split files are ``[input_name]/[n]/header.cpr`` and ``[input_name]/[n]/split-[n]-[id]`` respectively.  When reading - check for nprocs split configuration - if none exists - try to fall back to a serial (nprocs=1) configuration.

### DIFF
--- a/src/mesh/checkpoint_io.C
+++ b/src/mesh/checkpoint_io.C
@@ -42,6 +42,7 @@
 #include "libmesh/remote_elem.h"
 #include "libmesh/xdr_io.h"
 #include "libmesh/xdr_cxx.h"
+#include "libmesh/utility.h"
 
 namespace libMesh
 {
@@ -115,8 +116,6 @@ CheckpointIO::CheckpointIO (MeshBase & mesh, const bool binary_in) :
 {
 }
 
-
-
 CheckpointIO::CheckpointIO (const MeshBase & mesh, const bool binary_in) :
   MeshOutput<MeshBase>(mesh,/* is_parallel_format = */ true),
   ParallelObject      (mesh),
@@ -127,14 +126,17 @@ CheckpointIO::CheckpointIO (const MeshBase & mesh, const bool binary_in) :
 {
 }
 
-
-
 CheckpointIO::~CheckpointIO ()
 {
 }
 
-
-
+std::string extension(const std::string & s)
+{
+  auto pos = s.rfind(".");
+  if (pos == std::string::npos)
+    return "";
+  return s.substr(pos, s.size() - pos);
+}
 
 void CheckpointIO::write (const std::string & name)
 {
@@ -148,11 +150,32 @@ void CheckpointIO::write (const std::string & name)
   // do a gather_to_zero() and support that case too.
   _parallel = _parallel || !mesh.is_serial();
 
+  auto ext = extension(name);
+  std::string dir_name;
+  std::string str_my_n_procs = "1";
+  if (_parallel)
+    str_my_n_procs = std::to_string(_my_n_processors);
+
+  dir_name = name + "/" + str_my_n_procs;
+  std::string header_file_name = dir_name + "/header" + ext;
+
+  auto errcode = Utility::mkdir(name.c_str());
+  // error only if we failed to create dir - don't care if it was already there
+  if (errcode != 0 && errcode != -1)
+    libmesh_error_msg("Failed to create mesh split directory '" << name << "'");
+
+  errcode = Utility::mkdir(dir_name.c_str());
+  if (errcode == -1)
+    libmesh_warning("In CheckpointIO::write, directory "
+                    << dir_name << " already exists, overwriting contents.");
+  else if (errcode != 0)
+    libmesh_error_msg("Failed to create mesh split directory '" << dir_name << "'");
+
   // We'll write a header file from processor 0 to make it easier to do unambiguous
   // restarts later:
   if (this->processor_id() == 0)
     {
-      Xdr io (name, this->binary() ? ENCODE : WRITE);
+      Xdr io (header_file_name, this->binary() ? ENCODE : WRITE);
 
       // write the version
       io.data(_version, "# version");
@@ -220,11 +243,8 @@ void CheckpointIO::write (const std::string & name)
     {
       const processor_id_type my_pid = *id_it;
 
-      std::ostringstream file_name_stream;
-
-      file_name_stream << name << "-" << (_parallel ? _my_n_processors : 1) << "-" << my_pid;
-
-      Xdr io (file_name_stream.str(), this->binary() ? ENCODE : WRITE);
+      auto file_name = dir_name + "/split-" + str_my_n_procs + "-" + std::to_string(my_pid) + ext;
+      Xdr io (file_name, this->binary() ? ENCODE : WRITE);
 
       std::set<const Elem *, CompareElemIdsByLevel> elements;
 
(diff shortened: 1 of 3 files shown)


jwpeterson:
If I understand correctly, this is a behavior change, since anyone previously using `CheckpointIO::write()` in parallel will now have their files written to a `foo-split12/` directory, but not an API change? Sounds OK to me, any issue with this @roystgnr?

rwcarlsen:
Correct.  We can get the behavior we need from the MOOSE side without libmesh changes if there are too many concerns with this, but this seemed more foolproof and will help prevent people from getting their dirs bombed with bajillions of cpr files.

roystgnr:
The obvious issue is that this was on my TODO list, and I'm ashamed someone else got to it first.

I think putting n_procs into the subdirectory name is redundant - either you're writing out different splits with different header file names, in which case they'll already have different subdirectory names too, or you're writing out different splits with the same header file name, in which case the overwritten header file makes the older split files impossible to read correctly.

Other than that (which I'm not too too bothered by, if others disagree with me) this looks good.

Except...

Why are we now having to manually set cpr.current_n_processors() in the test?  Shouldn't that be read in from the header file?

Ah, now I see.  The header file is getting written to the subdirectory too.

So this basically has the same problems I had with the original suggestion #1455, except IMO worse: instead of reader code having to know a priori what n_procs it's reading so it can munge a filename, it has to know a priori what n_procs it's reading so it can set an API that will later munge a directory name.

Could we leave just the header file in the current directory, so that it can be used to determine n_procs when reading?  If we support the same "when asked to read foo.cpr, try to read foo.cpr then try to read foo.cpr-$n_procs" syntax, then you'd still get the N-to-N behavior you'd want without making it harder to use N-to-M too.

rwcarlsen:
I don't disagree with any of your concerns.  However, leaving the header file in the CWD would make it all to easy for the header to get separated from its subdirectory of cprs.  Also it would be a bit pain to move them around - you would always have to remember to move the header together with the subdir.  I don't currently have any good suggestions, just pointing out more problems :-)

roystgnr:
Hmm... could we put the header in the subdir, but make the subdir name be whatever.cpr as specified by the user, unchanged, and do the "try whatever.cpr first, try whatever.cpr-$n_proc" second trick on that?

friedmud:
@roystgnr I think you need to convince us (at least me) why a header file is necessary at all.

What is one single advantage of having the header file?

It is completely redundant and unnecessary information.

But maybe I'm not seeing a use-case you're thinking of.

roystgnr:
You can see how many processors the file was written to without reading an entire directory.

We can store version information once instead of O(N_proc) times.  (Not that we use version information now, but the intent is to stabilize this format eventually)

Likewise for datatype size, mesh dimension, subdomain names, and boundary id name strings.

friedmud:
Is every processor supposed to read that header file?  That could be an issue if you're trying to run on 100k procs or more.  Or does processor 0 read that file and broadcast (still could be a scaling problem).

I really just want each processor to read it's own file and be done.

All the information the processor needs should be in that one file.

roystgnr:
Processor 0 reads the file and broadcasts.

It's a scaling problem, but it's not as bad as additional file I/O would be.  We can probably reduce the problem in the future, by distributing the currently-serialized name strings.

friedmud:
I'm not super convinced - but ok 😄 

Ok - so let's talk about the specific thing here: directory naming.

Here is my new proposal (trying to meet both sides here).  Let's say your mesh is called `mesh.e` and you want to split it into `mesh.cpr` for `8`,`16`,`32` procs.  I'm proposing this:

A _directory_ named `mesh.cpr`

Underneath there, subdirectories named `8`, `16`, `32`

Within each subdirectory would be the header for that split and all of the `mesh.cpr-proc` files (where `proc` is the proc that should be reading that file).

So we would have:

mesh.cpr/8/mesh.cpr
mesh.cpr/8/mesh.cpr-0
mesh.cpr/8/mesh.cpr-1
...
mesh.cpr/8/mesh.cpr-7
...

This is nice and tidy and clean...

roystgnr:
Hey, *I'm* not super convinced either!  In theory reading a whole directory to figure out n_procs is O(N) communication with the filesystem, but reading it once and broadcasting is a single O(log(N)) step, so a priori I say lets do the latter, but if someone sets up a benchmark that says do otherwise for e.g. N==10000 then I'll consider myself outvoted.  The other performance issues, I'm speculating about constants, not asymptotic rates, so I'm even less sure there.

I love your subdirectory idea.

That's *perfect* for the N-to-M case, even - you could presplit a mesh many different ways, and then when running auto-select whichever split you expect to work best on your current n_procs: use $n_procs if it's there, of course, but otherwise pick the smallest split with more than $n_procs pieces if one exists, and otherwise pick the largest split.

rwcarlsen:
How do we handle the case where the filename is "foo/mesh.cpr"? - then you would end up with ``foo/mesh.cpr/8/foo/mesh.cpr``...

rwcarlsen:
How about for filename ``foo/mesh.cpr`` we do:

```
foo/mesh.cpr/8/header.cpr
foo/mesh.cpr/8/split1.cpr
foo/mesh.cpr/8/split2.cpr
...
```

friedmud:
I'm ok with that - but let's do:

```
foo/mesh.cpr/8/header.cpr
foo/mesh.cpr/8/split0.cpr
foo/mesh.cpr/8/split1.cpr
```

😄

rwcarlsen:
I was thinking it might be worthwhile to switch all our indexing to be 1-based.

pbauman:
> I was thinking it might be worthwhile to switch all our indexing to be 1-based.

Is the next suggestion going to be switch to Fortran?

roystgnr:
In case anyone hasn't already read it, [Why numbering should start at zero](http://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html), Edsger W. Dijkstra, UT Austin.  (okay, Burroughs Corporation when that was written, but UT snagged him eventually)

rwcarlsen:
Okay - I think this should be ~ready.  It will fail the moose tests - I have a change ready to fix them that will need to go in simultaneously with the MOOSE libmesh update.

src/mesh/checkpoint_io.C
```diff
@@ -148,11 +150,32 @@ void CheckpointIO::write (const std::string & name)
   // do a gather_to_zero() and support that case too.
   _parallel = _parallel || !mesh.is_serial();
 
+  std::string dir_name;
+  std::string header_file_name = name;
```
jwpeterson: This `header_file_name` initialization seems unnecessary/should just be moved down a few lines where you are setting it to something else.

src/mesh/checkpoint_io.C
```diff
@@ -148,11 +150,32 @@ void CheckpointIO::write (const std::string & name)
...
+  auto ext = extension(name);
+  dir_name = name + "/" + str_my_n_procs;
+  header_file_name = dir_name + "/header" + ext;
+
+  auto errcode = Utility::mkdir(name.c_str());
+  if (errcode != 0 && errcode != -1)
+    libmesh_error_msg("Failed to create mesh split directory '" << name << "'");
+  errcode = Utility::mkdir(dir_name.c_str());
+  if (errcode == -1)
+    libMesh::out << "In CheckpointIO::write, directory "
```
jwpeterson: Consider using the `libmesh_warning()` macro for this, so that users who configure with `--disable-warnings` don't have to see it.

PR #1577, opened 2018-02-05
Fix mis-branching in CouplingMatrix iterator++

roystgnr:
This was causing a *huge* performance penalty in cases where we had
many variables.

Hopefully https://github.com/idaholab/moose/issues/9480 will be fully
resolved by the fix.  I see orders of magnitude speedup in our test
case.

Assuming this doesn't break a hundred test cases (whether by exposing an underlying bug, or by failing to match a regression standard generated with the broken version, although I don't think either is likely to happen), @dschwen would you mind checking the asymptotic behavior again?  There's still some slight optimizations that might be possible here, but I think I've caught the entire O(N^3) mistake.

### DIFF
--- a/include/numerics/coupling_matrix.h
+++ b/include/numerics/coupling_matrix.h
@@ -337,12 +337,14 @@ class ConstCouplingRow
 public:
   ConstCouplingRow(unsigned int row_in,
                    const CouplingMatrix & mat_in) :
-    _row_i(row_in), _mat(mat_in)
+    _row_i(row_in), _mat(mat_in),
+    _end_location(_row_i*_mat.size()+_mat.size()-1)
   {
     libmesh_assert_less(_row_i, _mat.size());
 
-    // Location for i,N
-    _begin_location = _row_i*_mat.size()+_mat.size()-1;
+    // Location for i,N, where we'll start looking for our beginning
+    // range
+    _begin_location = _end_location;
 
     const std::size_t max_size = std::numeric_limits<std::size_t>::max();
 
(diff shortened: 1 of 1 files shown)


roystgnr:
Breaking unit tests is never a good sign.  Trying to replicate.

jwpeterson:
@roystgnr is this still in "do not merge" condition? Looks like it's passing all the tests now... does it also still fix the original performance regression?

roystgnr:
Nope.  I'm back in debugging mode at the moment, testing out Plan C.

roystgnr:
I'm hoping to have it in working order by this afternoon.  I assume you're chomping at the bit to get this included in the next MOOSE libMesh update?

jwpeterson:
> I assume you're chomping at the bit to get this included in the next MOOSE libMesh update?

It would be nice, but we are also mainly waiting for the CheckpointIO changes to be un-reverted (re-reverted?) so you're not holding anything up.

roystgnr:
Ok, only seeing 1 more OOM speedup with the newest version, but I think I've got it correct now.

roystgnr:
Hmmm... I made the mistake of doing those last timings with more than just this PR; there's also another minor optimization in my local checkout (so 1 OOM might be an overestimate) and there's some thick cerr-debugging (so 1OOM might be an underestimate).  Regardless, I think this PR alone fixes the ridiculously awful asymptotic complexity bug and should be merged ASAP if it passes tests.  If my other optimization makes a constant-factor improvement on top of that then I'll put it in a separate PR.

There's a possibility of false positive test failures, if one of the side effects of the previous iterator behavior (overfilling sparsity patterns) happens to have been masking an application bug (declaring a too-sparse CouplingMatrix and then attempting to set entries outside it), so normally I'd want to turn on all the MOOSE app tests right now, but if there's a general MOOSE libMesh update coming soon then we can just wait for that.

roystgnr:
Yup; most of my timing was stderr output.  :-P  I see about 2.5 OOM speedup now.

dschwen:
Pulling the branch now

roystgnr:
Actually, that looks like it might possibly be N^2 log(N)??  That might actually be the best we can do, if it is.  I'll put up my last optimization, though, which may at least shave log(N) off the scalar case.

YaqiWang:
I am lazy to check, has this been merged into moose libmesh submodule?

roystgnr:
Went in via https://github.com/idaholab/moose/pull/10740, yes.

PR #1578, opened 2018-02-08
Short circuit pthread/openmp execution for serial runs

friedmud:
Why are you always preparing - just go!

### DIFF
--- a/include/parallel/threads_pthread.h
+++ b/include/parallel/threads_pthread.h
@@ -271,13 +271,19 @@ void parallel_for (const Range & range, const Body & body)
 {
   Threads::BoolAcquire b(Threads::in_threads);
 
+  // If we're running in serial - just run!
+  if (libMesh::n_threads() == 1)
+  {
+    body(range);
+    return;
+  }
+
 #ifdef LIBMESH_ENABLE_PERFORMANCE_LOGGING
   const bool logging_was_enabled = libMesh::perflog.logging_enabled();
 
   if (libMesh::n_threads() > 1)
     libMesh::perflog.disable_logging();
 #endif
-
   unsigned int n_threads = num_pthreads(range);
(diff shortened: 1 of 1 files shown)


roystgnr:
Could we keep the BoolAcquire(in_threads) for this code path, at least in dbg/devel modes?  I like it when I get caught trying to run non-within-thread-safe code within a threaded section, even when (as is typical) I'm not currently using threads.

Otherwise :+1: once we figure out what the heck is going wrong (failure from the naming update that didn't get caught there somehow??) with all those MOOSE CheckpointIO failures.

friedmud:
Does that `BoolAcquire` take any time?

I circumvented it because this was an optimization anyway - so I was trying to do the most optimal thing.

Yeah - the CheckpointIO stuff is because it's an incompatible change with MOOSE.  MOOSE has to be patched for it... but that can't happen until the libMesh submodule is updated in MOOSE.  We hope to do all of that today.

roystgnr:
The constructor sets a bool to true and stores its address, the destructor sets the bool to false.  Pretty trivial, but it definitely doesn't get completely optimized away, and that's a global bool so it might even be a cache miss if you're hitting threaded sections infrequently.  Even O(100ns) per parallel_for call wouldn't sound too bad to me, *except* that if you're already finding yourself optimizing this then I guess it shows up in profiling somewhere.

We currently only use Threads::in_threads within libMesh and MOOSE in internal assertions... and it doesn't look like FIN-S or GRINS use it at all... I'd be willing to bet that we could make that variable's very existence conditional on ```#ifndef NDEBUG``` without really inconveniencing anybody, if you want.

friedmud:
That sounds light enough to me: I'll change it

friedmud:
I think this is ready...

PR #1579, opened 2018-02-08
Updates to --with-thread-model

jwpeterson:
This PR adds the `--with-thread-model=openmp` option, and removes TBB from the list of threading models that are "auto detected". TBB has declined significantly in popularity over time, and probably isn't as safe of a default as pthreads at this point.

### DIFF
--- a/configure
+++ b/configure
@@ -1210,7 +1210,6 @@ enable_tbb
 with_tbb
 with_tbb_lib
 enable_pthreads
-enable_openmp
 enable_laspack
 enable_sfc
 enable_gzstreams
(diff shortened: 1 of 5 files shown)


jwpeterson:
Since we decided to rethink some stuff on #1576, I think it is appropriate to temporarily revert those changes... I pushed a commit which does that to this branch, but if it works, the revert commit should go in a separate PR (or directly onto master) and this PR should be rebased.

roystgnr:
I'm happy with pushing TBB to the bottom of the autodetection list, but do we really need to remove it from autodetection entirely?

jwpeterson:
> I'm happy with pushing TBB to the bottom of the autodetection list, but do we really need to remove it from autodetection entirely?

You're right, I'll allow it to be auto-detected after pthreads...

PR #1581, opened 2018-02-08
read/write checkpoint files to/from nprocs-suffixed directories

rwcarlsen:
When writing and reading checkpoint files, put sets of checkpoint files
for a single mesh into a subdirectory with the given checkpoint object
input_name.  Within this directory, every different split
configuration of the mesh gets its own subdirectory named "[n]".  Within
those directories, headers are named "header.cpr" and the actual split
files are named "split-[proc]-[n_procs]".

When reading checkpoint files, only the name if the top-level directory
(e.g. "foo.cpr") need be specified.  A split configuration will be
automatically selected based on the current number of procs being run.

Added a ``cleanup`` function for robustly removing files+dirs associated with a particular checkpoint configuration.  This will help users of the checkpoint code to be able to remove/modify checkpoint files without having to write checkpoint file format specific code.

### DIFF
--- a/src/mesh/checkpoint_io.C
+++ b/src/mesh/checkpoint_io.C
@@ -42,6 +42,7 @@
 #include "libmesh/remote_elem.h"
 #include "libmesh/xdr_io.h"
 #include "libmesh/xdr_cxx.h"
+#include "libmesh/utility.h"
 
 namespace libMesh
 {
@@ -115,8 +116,6 @@ CheckpointIO::CheckpointIO (MeshBase & mesh, const bool binary_in) :
 {
 }
 
-
-
 CheckpointIO::CheckpointIO (const MeshBase & mesh, const bool binary_in) :
   MeshOutput<MeshBase>(mesh,/* is_parallel_format = */ true),
   ParallelObject      (mesh),
@@ -127,14 +126,17 @@ CheckpointIO::CheckpointIO (const MeshBase & mesh, const bool binary_in) :
 {
 }
 
-
-
 CheckpointIO::~CheckpointIO ()
 {
 }
 
-
-
+std::string extension(const std::string & s)
+{
+  auto pos = s.rfind(".");
+  if (pos == std::string::npos)
+    return "";
+  return s.substr(pos, s.size() - pos);
+}
 
 void CheckpointIO::write (const std::string & name)
 {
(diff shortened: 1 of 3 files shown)


src/mesh/checkpoint_io.C
```diff
@@ -76,6 +78,53 @@ void chunking(processor_id_type size, processor_id_type rank, unsigned int nspli
                              (1 - (int)nchunks) * std::numeric_limits<int>::max());
     }
 }
+
```
jwpeterson: I just realized that this anonymous namespace is declared *inside* of `namespace libMesh`... I'm not sure how this works, are these functions actually local only to this file, or are they actually in namespace libMesh? Maybe just move the opening `namespace libMesh {` down to avoid this potential confusion.

src/mesh/checkpoint_io.C
```diff
@@ -76,6 +78,53 @@ void chunking(processor_id_type size, processor_id_type rank, unsigned int nspli
...
+  return s.substr(pos, s.size() - pos);
+}
+
+std::string splitDir(const std::string & input_name, processor_id_type n_procs)
+{
+  return input_name + "/" + std::to_string(n_procs);
+}
+
+
+std::string headerFile(const std::string & input_name, processor_id_type n_procs)
```
jwpeterson: Ideally these helper functions should also not be camel-cased for consistency, but I don't care that much since they are just local.

src/mesh/checkpoint_io.C
```diff
@@ -130,12 +179,88 @@ CheckpointIO::~CheckpointIO ()
...
+
+  if (!input_n_procs)
+    input_n_procs = 1;
+  return input_n_procs;
+}
+
+void CheckpointIO::cleanup(const std::string & input_name, processor_id_type n_procs)
+{
+  auto header = headerFile(input_name, n_procs);
+  auto ret = remove(header.c_str());
```
jwpeterson: Please `#include <cstdio>` and use `std::remove()` for this, just so it's really obvious what function you're calling.

jwpeterson:
How confident are we on the MOOSE side changes that need to go along with this? Every libmesh PR is going to fail in the interim, so I'd like to be sure we can actually do a libmesh update in MOOSE relatively quickly once this is merged...

jwpeterson:
On second thought, I think I will simply push Robert's branch into upstream libmesh, but wait on merging it into master. That way, we can create a libmesh update PR over on the MOOSE side (updating to a non-master libmesh hash if necessary) and avoid breaking all libmesh PRs in the meantime.

jwpeterson:
Flagging this PR as "do not merge" for now. If idaholab/moose#10740 ends up getting merged, these commits will subsequently also be merged into master.

jwpeterson:
idaholab/moose#10740 was merged; removing the "do not merge" flag from this PR.

PR #1582, opened 2018-02-08
One more SparsityPattern::Build optimization

roystgnr:
This adds one commit to the tail of #1577, with a different (and independent, and not nearly as impressive) optimization targeting the same problem.

We'll often be adding new j dofs which are already all subsequent to
all existing j dofs in a row; in that case we can avoid searches and
sorts entirely.

This seems to give me another 40% or so speedup, but I'm not sure I trust my now-gratifyingly-short timings anymore on only 1000 variables; let's wait for @dschwen to look at more precise timings and scaling before we decide whether or not to merge this.

### DIFF
--- a/include/numerics/coupling_matrix.h
+++ b/include/numerics/coupling_matrix.h
@@ -337,12 +337,14 @@ class ConstCouplingRow
 public:
   ConstCouplingRow(unsigned int row_in,
                    const CouplingMatrix & mat_in) :
-    _row_i(row_in), _mat(mat_in)
+    _row_i(row_in), _mat(mat_in),
+    _end_location(_row_i*_mat.size()+_mat.size()-1)
   {
     libmesh_assert_less(_row_i, _mat.size());
 
-    // Location for i,N
-    _begin_location = _row_i*_mat.size()+_mat.size()-1;
+    // Location for i,N, where we'll start looking for our beginning
+    // range
+    _begin_location = _end_location;
 
     const std::size_t max_size = std::numeric_limits<std::size_t>::max();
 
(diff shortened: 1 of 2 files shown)


jwpeterson:
Is this a separate PR from #1577 because there's a chance #1577 will be merged but this one won't be?

roystgnr:
Exactly.  #1577 fixes a huge bug, which definitely makes a huge performance improvement for some cases, should be a slight improvement for others, and IMHO has no reasonable alternative; I plan to merge as soon as CI finishes and we should get it into MOOSE right away.  #1582 adds a new short-circuiting code path which probably makes a moderate performance improvement for some cases and which isn't the only reasonable refactoring; we can wait on it until we've looked at it more.

dschwen:
Small constant factor improvement for scalars only.

out

roystgnr:
Let me look at the Lagrange behavior before merging, then.  Still on one 1d element?

roystgnr:
Okay, I think I get it now.  Our default dof numbering has changed from "var_major" to var_GROUP major, with some autodetection of variable groups, so for field variables we generally can no longer say that all the dof indices belonging to a particular variable are ordered after all the dof indices belonging to a previously handled variable, so my short-cut for such cases is only going to be triggered on SCALAR variables.

This also probably explains our remaining poor asymptotic behavior on LAGRANGE - when we do our final sort after each of those N^2 steps, the reverse bubble sort we do isn't going to be O(1), but rather O(n)...

Anyway, I'll kill this PR; a small performance improvement in an uncommon case isn't worth the combination of more complicated code and what looks like a (even smaller) performance regression in a common case.

PR #1583, opened 2018-02-09
Use GCC diagnostic push/pop to ignore warnings.

brianmoose:
This came up due to idaholab/moose#10733
Using `-Werror` doesn't turn all the warnings into errors because the GCC diagnostic was restoring
specific warnings as **warnings**, ignoring that they should be treated as errors.
GCC diagnostic with push/pop was introduced in gcc 4.6. I don't think we need to support anything without push/pop support as libmesh now requires a C++-11 compiler (4.8 or later).

### DIFF
--- a/include/utils/ignore_warnings.h
+++ b/include/utils/ignore_warnings.h
@@ -36,8 +36,9 @@
 #endif
 
 #if defined(__GNUC__) && !defined(__INTEL_COMPILER) && !defined(__clang__)
-// GCC > 4.1 supports diagnostic pragmas
-#if (__GNUC__ > 4) || (__GNUC__ == 4 && __GNUC_MINOR__ > 1)
+// GCC > 4.5 supports diagnostic pragmas with push/pop
+#if (__GNUC__ > 4) || (__GNUC__ == 4 && __GNUC_MINOR__ > 5)
+#pragma GCC diagnostic push
 // These two don't work?
 #pragma GCC diagnostic ignored "-Wpedantic"
 #pragma GCC diagnostic ignored "-Wdeprecated"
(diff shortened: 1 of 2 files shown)


jwpeterson:
👍  Thanks, this is much cleaner now in addition to actually playing nicely with `-Werror`.

roystgnr:
This has been at the bottom of my TODO list forever as a means of cleaning up the code; I didn't realize the old ugly code also had practical problems!  Thanks!

:+1: as soon as Civet is done.

roystgnr:
As an aside, it's also been my dream for years to have CI running with -Werror active more often than not.  If the scrutiny and work done for https://github.com/idaholab/moose/issues/10733 ends up actually getting us able to turn that on then we should do so when we can.

(I'd even like to make it a default libMesh flag, but that will never happen; there are too many people who snarf libMesh compiler flags and aren't crazy-obsessive about warnings in their code)

jwpeterson:
Along the same lines, @brianmoose also pointed out to me that [we only turn on the full complement of warnigns in dbg/devel modes](https://github.com/libMesh/libmesh/blob/47ec79ec1ab1b09f7ded1fa9d8a1292d5e65355f/m4/compiler.m4#L500) for some reason. 

I think the original intent behind this was to prevent the average user from seeing lots of warnings when compiling libmesh for the first time, but if we are planning to just not allow any warnings then this distinction is no longer relevant, and actually leads to warning "surprises" on the CI when it compiles with different METHODS.

roystgnr:
Huh.  I don't actually recall what the original intent was there.  We should definitely turn warnings on in opt/oprof modes too.

PR #1586, opened 2018-02-12
Always detect compiler's OpenMP flags.

jwpeterson:
PR #1579 made enabling OpenMP "support" optional, but after subsequent testing we discovered a few problems with this approach. 

In particular, when `!LIBMESH_HAVE_OPENMP` we don't call `omp_set_num_threads()` in src/base/libmesh.C, and therefore are relying on the value of the `OMP_NUM_THREADS`environment variable to set this. If `OMP_NUM_THREADS` is not set, the default value is ["1 thread per CPU"](https://gcc.gnu.org/onlinedocs/libgomp/OMP_005fNUM_005fTHREADS.html). This default can lead to *dramatic* slowdowns (we observed something like 10x) in MPI codes due to oversubscription of the hardware. It also affects things like the MOOSE TestHarness, where we run dozens of separate tests in serial at the same time, and we obviously don't want each of them potentially spawning e.g. NCPUs=16 threads.

This PR essentially restores our original behavior of always detecting OpenMP during configure while still maintaining `pthreads` as the default thread model and `tbb` as the backup thread model.

### DIFF
--- a/configure
+++ b/configure
@@ -795,13 +795,13 @@ LASPACK_LIB
 LASPACK_INCLUDE
 TBB_INCLUDE
 TBB_LIBRARY
-OPENMP_FFLAGS
-OPENMP_CFLAGS
-OPENMP_CXXFLAGS
 PTHREAD_CFLAGS
 PTHREAD_LIBS
 PTHREAD_CC
 ax_pthread_config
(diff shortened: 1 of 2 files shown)


roystgnr:
:+1:

PR #1587, opened 2018-02-13
Cleanup petsc solve

lindsayad:
#1572 allows us to simplify this class


fdkong:
Nice

jwpeterson:
:+1: My only concern was whether `J` is cleaned up by PETSc, but I'll be satisfied if the Valgrind test passes :-)

PR #1588, opened 2018-02-14
Fixing c-tor in MeshBase

andrsd:
The member _count_lower_dim_elems_in_point_locator was not initialized
in the copy-constructor.  Also, was not set in the alternative c-tor.


andrsd:
Found this by running RELAP-7 through valgrind.

jwpeterson:
Thanks! Looks like the bug was introduced in 8fd59cb78b6cb (January 2017), so the bug was present in the v1.2.0 and v1.2.1 releases.

dknez:
Yes, looks like my mistake, thanks for the fix!

PR #1589, opened 2018-02-14
Add custom line search callback

lindsayad:
I've already implemented a `ComputeLinesearchObject` in MOOSE that demonstrates excellent convergence behavior for contact, so this would be very helpful.

I've also removed the double leading underscores for our Petsc methods since it's reserved for the compiler according to the ANSI-C standard. Maybe that should be a separate PR/commit.

The removal of the "dunder" will break MOOSE because of [this](https://github.com/idaholab/moose/blob/devel/framework/src/base/NonlinearSystem.C#L313).

### DIFF
--- a/include/solvers/petsc_linear_solver.h
+++ b/include/solvers/petsc_linear_solver.h
@@ -50,6 +50,24 @@ extern "C"
    * This function is called by PETSc to initialize the preconditioner.
    * ctx will hold the Preconditioner.
    */
+  PetscErrorCode libmesh_petsc_preconditioner_setup (void * ctx);
+
+  /**
+   * This function is called by PETSc to actually apply the preconditioner.
+   * ctx will hold the Preconditioner.
+   */
+  PetscErrorCode libmesh_petsc_preconditioner_apply(void * ctx, Vec x, Vec y);
+#else
+  PetscErrorCode libmesh_petsc_preconditioner_setup (PC);
+  PetscErrorCode libmesh_petsc_preconditioner_apply(PC, Vec x, Vec y);
+#endif
+
+#if LIBMESH_ENABLE_DEPRECATED
+#if PETSC_RELEASE_LESS_THAN(3,0,1)
+  /**
+   * This function is called by PETSc to initialize the preconditioner.
+   * ctx will hold the Preconditioner.
+   */
   PetscErrorCode __libmesh_petsc_preconditioner_setup (void * ctx);
 
   /**
@@ -60,6 +78,7 @@ extern "C"
 #else
   PetscErrorCode __libmesh_petsc_preconditioner_setup (PC);
   PetscErrorCode __libmesh_petsc_preconditioner_apply(PC, Vec x, Vec y);
+#endif
 #endif
 } // end extern "C"
 
--- a/include/solvers/petsc_nonlinear_solver.h
+++ b/include/solvers/petsc_nonlinear_solver.h
@@ -58,6 +58,7 @@ extern "C"
                                                 SNESLineSearch, Vec x, Vec y, Vec w, PetscBool * changed_y, PetscBool * changed_w, void * context
 #endif
                                                 );
+  PetscErrorCode libmesh_petsc_linesearch_shellfunc(SNESLineSearch linesearch, void * ctx);
 }
 
 /**
@@ -157,6 +158,22 @@ public:
    * Set to true to use the libMesh's default monitor, set to false to use your own
    */
   void use_default_monitor(bool state) { _default_monitor = state; }
+
+  /**
+   * Abstract base class to be used to implement a custom line-search algorithm
+   */
+  class ComputeLineSearchObject
+  {
+  public:
+    virtual ~ComputeLineSearchObject () {}
+
+    virtual void linesearch (SNESLineSearch linesearch) = 0;
+  };
+
+  /**
+   * A callable object that can be used to specify a custom line-search
+   */
+  std::unique_ptr<ComputeLineSearchObject> linesearch_object;
 
 protected:
 
--- a/include/solvers/petsc_nonlinear_solver.h
+++ b/include/solvers/petsc_nonlinear_solver.h
@@ -40,11 +40,31 @@ class ResidualContext;
 // need access to these most of the time as they are used internally by this object.
 extern "C"
 {
+  PetscErrorCode libmesh_petsc_snes_monitor (SNES, PetscInt its, PetscReal fnorm, void *);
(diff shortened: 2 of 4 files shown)


jwpeterson:
> The removal of the "dunder" will break MOOSE because of this.

So we definitely can't merge this as-is, since every libmesh PR will fail MOOSE testing until we have a chance to do another libmesh update in MOOSE... I'd suggest either dropping the double underscore removal from this, or even better, keeping the old function names around and adding a `libmesh_deprecated()` call before simply forwarding on their arguments to the newly named functions.

roystgnr:
> Maybe that should be a separate PR/commit.

Definitely should have been a separate commit; the situation here where we'd like to merge half the changeset ASAP but rework the other half is a good example of why!

> keeping the old function names around and adding a libmesh_deprecated() call before simply forwarding on their arguments to the newly named functions.

:+1:

Let's get one deprecated-marked release out before removing them entirely.

As for the line search callback... it would be nice if we could encapsulate it better so as to bring it up to NonlinearSolver eventually, but that sounds like it would be 10x as much work for no short-term benefit, so I'm fine with it as-is.

lindsayad:
Split commits here now, so that gives us the option of splitting PRs. Added the deprecated methods.

lindsayad:
This will be tested in MOOSE eventually. For the record, I ran valgrind on an implementation, and it came back clean. Wanted to make sure I avoided a replay of the last callback implementation 😄

include/solvers/petsc_linear_solver.h
```diff
@@ -45,6 +45,24 @@
 // Give them an obscure name to avoid namespace pollution.
 extern "C"
 {
+#if PETSC_RELEASE_LESS_THAN(3,0,1)
```
jwpeterson: This check should be `PETSC_VERSION_LESS_THAN`. 

The `PETSC_RELEASE_LESS_THAN` macro is used when you need to check *both* a minimum version *and* that you are using an actual released version of PETSc (as opposed to some branch like master, next, etc.). It's usually used with negation (`!PETSC_RELEASE_LESS_THAN`) to compile with code that is currently only in PETSc master.

I see a few other places where the macros are wrong in this file, so I can fix that in a separate PR.

include/solvers/petsc_nonlinear_solver.h
```diff
@@ -158,6 +179,22 @@ class PetscNonlinearSolver : public NonlinearSolver<T>
...
+  public:
+    virtual ~ComputeLineSearchObject () {}
+
+    virtual void linesearch (SNESLineSearch linesearch) = 0;
+  };
+
+  /**
+   * A callable object that can be used to specify a custom line-search
+   */
+  std::unique_ptr<ComputeLineSearchObject> linesearch_object;
```
jwpeterson: Back when we were using "fake" C++03 unique_ptr stand-ins, it was bad to put them in the public interface like this because it was possible for them to be accidentally "stolen" too easily. For example, the `BoundaryInfo` in `MeshBase` is only accessible through the `get_boundary_info()` interface. Now that we are using real unique_ptrs it's probably fine though?

@roystgnr any opinion on this?

src/solvers/petsc_linear_solver.C
```diff
@@ -95,6 +95,39 @@ extern "C"
...
+    libmesh_deprecated();
+    PetscErrorCode ierr = libmesh_petsc_preconditioner_apply(ctx, x, y);
+    return ierr;
+  }
+
+#else
+  PetscErrorCode __libmesh_petsc_preconditioner_setup (PC pc)
+  {
+    libmesh_deprecated();
+    PetscErrorCode ierr = libmesh_petsc_preconditioner_setup(pc);
```
jwpeterson: I think 
```
return libmesh_petsc_preconditioner_setup(pc);
```
should be sufficient, no need to declare the `ierr` variable.

include/solvers/petsc_linear_solver.h
lindsayad: 👍

src/solvers/petsc_nonlinear_solver.C
```diff
@@ -801,14 +891,25 @@ PetscNonlinearSolver<T>::solve (SparseMatrix<T> &  pre_in,  // System Preconditi
...
 #if !PETSC_RELEASE_LESS_THAN(3, 8, 4)
   // Resue the residual vector from SNES
   ierr = MatSNESMFSetReuseBase(J, PETSC_TRUE);
   LIBMESH_CHKERR(ierr);
 #endif
 
+  SNESLineSearch linesearch;
+  if (linesearch_object)
+  {
+    ierr = SNESGetLineSearch(_snes, &linesearch);
```
jwpeterson: @lindsayad is currently working on some backwards compatibility guards for these lines of code.

include/solvers/petsc_nonlinear_solver.h
jwpeterson: I think we decided to just go with this change, so this comment should be considered "resolved"

src/solvers/petsc_nonlinear_solver.C
```diff
@@ -128,12 +134,21 @@ extern "C"
     return 0;
   }
 
+#ifdef LIBMESH_ENABLE_DEPRECATED
+  PetscErrorCode
+  __libmesh_petsc_snes_monitor (SNES, PetscInt its, PetscReal fnorm, void *)
+  {
+    libmesh_deprecated();
+    libmesh_petsc_snes_monitor(nullptr, its, fnorm, nullptr);
```
jwpeterson: This callback returns an actual error code, so why are we always returning 0?

jwpeterson: Oh, I see now that the error code returned by the original version of `__libmesh_petsc_snes_monitor` was always 0. This version should simply 
```
return libmesh_petsc_snes_monitor(nullptr, its, fnorm, nullptr);
```
though.

jwpeterson:
The latest failure is not your fault, I just had @brianmoose add an `[InstallCheck]` recipe for libMesh PRs. If you merge my `test_installed_headers` branch from #1597 into this PR, it should start passing again.

lindsayad:
Do we not care about an extraneous branch merge? If we do care, I can just
wait for that other PR to be merged...

On Mon, Feb 19, 2018 at 12:20 PM, John W. Peterson <notifications@github.com

jwpeterson:
> I can just wait for that other PR to be merged

I'd say if you were going to push again for some other reason then do the merge, otherwise waiting is fine. Once #1597 is merged, you may be able to just invalidate that one target instead of rerunning everything...

Issue #1512, opened 2017-12-01
UnstructuredMesh::copy_nodes_and_elements() requires matching _n_parts


jwpeterson:
We have a case where we're hitting the following assert 

https://github.com/libMesh/libmesh/blob/06410d04d134290b911558d1382b559220edfa57/src/mesh/unstructured_mesh.C#L79

in `UnstructuredMesh::copy_nodes_and_elements()` via a call to `ReplicatedMesh::stitch_meshes()` while running in parallel (np=4), because the mesh being "attached" in this particular case only has 3 elements total. (The application is constructing a "master" mesh by stitching together several smaller parts, some of which have very few elements...) 

The app runs fine in optimized mode, and this makes sense to me in principle, i.e. if you are deep copying one mesh over another, should it matter if they are initially partitioned onto the same number of procs? This also seems like it might be a relatively common use case when stitching meshes (?). So, I'm wondering if the assert is overzealous in at least some use cases of `copy_nodes_and_elements()` but required in others?

roystgnr:
Hmm... we either need the communicator on the target mesh to have at least as many ranks as the number of partitions on the source mesh, or we need to "wrap around" any source DofObject processor_id() which exceeds n_processors() on the target mesh.  But I don't see any reason we need the exact same number of partitions, if either that weaker assertion or that wrap-around code is in place instead.

stimpsonsg:
Was there any decision made on a path forward on this?

jwpeterson:
> Was there any decision made on a path forward on this?

I think the simplest approach is to try changing the line above to 
```
libmesh_assert_greater_equal_msg(_n_parts, other_mesh._n_parts);
```
and seeing if that gets us any farther.

stimpsonsg:
ping

permcody:
@jwpeterson, @roystgnr - This issue came up on our CASL call today. The Oakridge guys have had to turn off debug builds for BISON because of this issue so we'd like to pin it down and get it fixed. So first things, first, we need a test case to replicate the issue. 

@stimpsonsg - Can you tell us how to reproduce? If you have BISON input that does it under certain conditions, I'll start from there and see if I can distill it down into a MOOSE or libMesh only problem and then we should be able to get it fixed.

Was there any consensus on whether or not this might be an over-zealous assertion?

jwpeterson:
@gardnerru I believe you were the one looking into this problem. Did you ever have a chance to check whether the problem above fixes the issue you were seeing?

[...]

PR #1590, opened 2018-02-15
Weaken copy_node_and_elements partition assertion

roystgnr:
This will hopefully be sufficient to fix current #1512 problems.


permcody:
@gardnerru  - Can you pull this patch and see if it fixes the debug runs of the CASL problem with BISON?

gardnerru:
yea, looking now

gardnerru:
Looks like a winner. I ran the input that I received when this was first presented in serial, 4 procs and 8 procs and they all ran. Thanks

jwpeterson:
Thanks @gardnerru for testing it.

PR #1592, opened 2018-02-16
Reactivate fem_system_ex2

roystgnr:
There doesn't seem to be any way to restore the old reinit() behavior without regressing the performance improvements made by https://github.com/libMesh/libmesh/commit/3b4b2fbce7aae208974132ec033b9df137439a7f - so let's perform that operation manually where we need it in fem_system_ex2, and turn that example back on.

### DIFF
--- a/examples/fem_system/fem_system_ex2/fem_system_ex2.C
+++ b/examples/fem_system/fem_system_ex2/fem_system_ex2.C
@@ -82,7 +82,9 @@ void setup(EquationSystems & systems,
   imms.save_initial_mesh();
 
   // Fill global solution vector from local ones
-  aux_sys.reinit();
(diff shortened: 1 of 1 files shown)


jwpeterson:
Thanks, I was just going to ask about this for #1593. Will cherry-pick these commits over once they are in master.

roystgnr:
Sounds good.

Sorry about the delay.  It took me too long to convince myself that there really was no middle option in between "restore expensive communication that most codes don't make use of" and "omit communication step that a few codes currently rely on".

The fact that that communication step was undocumented is the only thing that will help me sleep at night about the behavior change - I can't even think of a safe way to catch affected codes in a libmesh_assert() without potentially creating a false positive assertion failure for other codes.

Issue #1598, opened 2018-02-20
DistributedMesh ParmetisPartitioner interface performance is awful


roystgnr:
Thanks to @bboutkov for making this easy to replicate with https://github.com/bboutkov/libmesh/commit/fb76306dc34875d45c0503c29575bd11ca53b25a

Splitting a few million elements among a few dozen processors in that uniform refinement study requires about 20 seconds spent in find_global_indices().  Splitting them among a few hundred processors requires about *2000* seconds spent in find_global_indices().

Obviously the nicest way to resolve this would be to fix find_global_indices() (and thereby speed up I/O using that method as well!) but if that's not possible we can probably avoid calling for perfectly Hilbert-sorted DofObject ids in the first place; unlike in the case of Xdr I/O we don't care about partitioning independence or N-to-M or any such thing here, we only need to meet the ParMETIS API requirements.

roystgnr:
Parallel::sum() is a noticeable time sink in the 32 processor case, but the cost growth is O(log2(Nproc)), which may be as good as we can hope for.

Parallel::probe() calls (in the send_receive() calls in find_global_indices()) are taking a large minority of the PerfLog time at 320 processors, suggesting some load imbalance.

The majority of time spent is in the Parallel::Sort operation, and apparently *not* in communication subroutines...

roystgnr:
Ah, never mind, it's in communication.  Sort::communicate_bins() is responsible for all the time, and makes direct MPI calls which aren't tracked by the PerfLog.

jwpeterson:
IIRC `Parallel::Sort` is something we implemented ourselves, but it might not be the best (or even a good) algorithm for sorting in parallel? Perhaps there is a more performant "standard" implementation we could use instead?

roystgnr:
I tried changing find_global_indices() from the naive round-robin send_receive() to sophisticated asynchronous I/O... and that actually ends up nearly doubling the amount of time I see spent in that communication!

Which is pretty dumbfounding.  Clearly my intuition here is way off.  I might still get to a faster solution along these lines by optimizing for the high N case (where most communications are empty and we can skip them) but I'm no longer optimistic.

It may be time to refactor all our instances of that sort of communication algorithm, similarly to what we already have in parallel_ghost_sync.h.  Then we can swap algorithms in and out without rewriting a hundred lines of code each time, and see what works fastest where.

jwpeterson:
[...]

PR #1600, opened 2018-02-22
Use alltoall to filter out non-communicating-procs

roystgnr:
An alltoall operation is O(P log(P)), whereas naively sending all
possible messages must have been O(P^2).

For me on a 320 processor half-cluster run with OpenMPI, this speeds up
this stretch of code by a factor of 4, which speeds up the total
repartitioning time by 25%.  This is not a sufficient fix for #1598 but
it's a good start.

And there are several other DistributedMesh communications which would
benefit from this same idiom...

### DIFF
--- a/src/mesh/mesh_communication_global_indices.C
+++ b/src/mesh/mesh_communication_global_indices.C
@@ -800,8 +800,8 @@ void MeshCommunication::find_global_indices (const Parallel::Communicator & comm
         if (requested_ids[procup].empty())
           continue;
 
-	communicator.send(procup, requested_ids[procup],
-			  key_send_requests[procup], key_send_tag);
+        communicator.send(procup, requested_ids[procup],
+                          key_send_requests[procup], key_send_tag);
 
         non_empty_requests++;
       }
@@ -812,8 +812,8 @@ void MeshCommunication::find_global_indices (const Parallel::Communicator & comm
         if (!request_non_empty[pid])
           continue;
 
-	std::vector<Parallel::DofObjectKey> request_to_fill;
-	communicator.receive (pid, request_to_fill, key_send_tag);
+        std::vector<Parallel::DofObjectKey> request_to_fill;
+        communicator.receive (pid, request_to_fill, key_send_tag);
 
         std::vector<dof_id_type> & global_ids = global_id_returns[pid];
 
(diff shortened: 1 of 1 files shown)


src/mesh/mesh_communication_global_indices.C
```diff
@@ -858,9 +889,24 @@ void MeshCommunication::find_global_indices (const Parallel::Communicator & comm
             global_ids.push_back (cast_int<dof_id_type>(std::distance(my_bin.begin(), pos) + my_offset));
           }
 
-        // and trade back
-        communicator.send_receive (procdown, global_ids,
-                                   procup,   filled_request[procup]);
+        // and send back
+        id_return_requests.push_back(Parallel::Request());
+	communicator.send (pid, global_ids,
```
jwpeterson: Is this line indented slightly differently than the line above (1 char)?

jwpeterson:
Do you expect this behavior to be MPI implementation dependent at all? Is there a simple test we could try running on our cluster to get before/after data?

src/mesh/mesh_communication_global_indices.C
roystgnr: Oh, hell.  I let a tab slip in somehow.  I'll fix it.  In a separate commit, if you don't mind, so I can merge with other branches easier.

roystgnr: Oh, I get it.  I was editing code on a cluster where I haven't yet copied over my .vimrc.  There's a few tabs I need to fix.  Thanks for catching that one!

roystgnr:
> Do you expect this behavior to be MPI implementation dependent at all?

From what I've read, I expect it to be so MPI-implementation-dependent that it makes me want to cry.  IMHO we ought to see speedup everywhere but the magnitude of the speedup could depend on your MPI version, your interconnect topology, and the phase of the moon.

> Is there a simple test we could try running on our cluster to get before/after data?

Try introduction ex3 in the https://github.com/roystgnr/libmesh/tree/partitioning_speed branch to get after data - it does nothing but a whole lot of refinement.  Then just checkout HEAD^ to get before data.

I'd like to see that too, actually.  We can wait on it before merging.  I showed @bboutkov a partial workaround for the performance problems plaguing his GMG work, so the long-term fix isn't quite as urgent as we thought it was at first.

roystgnr:
I'll wait for that last CI test to finish before pushing the whitespace fixes, then I can just cancel the new round of tests they trigger.

pbauman:
Job [GRINS-dbg](https://femputer.eng.buffalo.edu:8000/job/1198/) on 9dde0c2 : canceled by @roystgnr

pbauman:
Job [GRINS-devel](https://femputer.eng.buffalo.edu:8000/job/1197/) on 9dde0c2 : canceled by @roystgnr

roystgnr:
"The optimization in #1600 looks like a win for all processor counts and all refinement levels to me" - @jwpeterson 

I feel like that should go on the back of a dust jacket for this PR.

PR #1601, opened 2018-02-23
Ignore a few more warnings

permcody:
Don't ignore warngings that don't exist on older clang.

@brianmoose - I'm not exactly sure how to test this. Ideally, we'd test this with your existing PR (https://github.com/idaholab/moose/pull/10852).

### DIFF
--- a/include/utils/ignore_warnings.h
+++ b/include/utils/ignore_warnings.h
@@ -32,7 +32,11 @@
 #pragma clang diagnostic ignored "-Wnested-anon-types"
 #pragma clang diagnostic ignored "-Wsign-compare"
 #pragma clang diagnostic ignored "-Wunused-private-field"
+#pragma clang diagnostic ignored "-Wsign-compare"
+#pragma clang diagnostic ignored "-Wextra"
+#if (__clang_major__ > 3) || ((__clang_major__) == 3 && (__clang_minor__ >= 5))
 #pragma clang diagnostic ignored "-Winconsistent-missing-override"
+#endif
 #endif
 
 #if defined(__GNUC__) && !defined(__INTEL_COMPILER) && !defined(__clang__)


include/utils/ignore_warnings.h
```diff
@@ -32,8 +32,12 @@
 #pragma clang diagnostic ignored "-Wnested-anon-types"
 #pragma clang diagnostic ignored "-Wsign-compare"
 #pragma clang diagnostic ignored "-Wunused-private-field"
+#pragma clang diagnostic ignored "-Wsign-compare"
+#pragma clang diagnostic ignored "-Wextra"
+#if (__clang_major__ < 3) || ((__clang_major__) == 3 && (__clang_minor__ >= 5))
```
brianmoose: `> 3` ?

permcody: oh... this is totally wrong...

include/utils/ignore_warnings.h
```diff
@@ -32,8 +32,12 @@
 #pragma clang diagnostic ignored "-Wnested-anon-types"
 #pragma clang diagnostic ignored "-Wsign-compare"
 #pragma clang diagnostic ignored "-Wunused-private-field"
+#pragma clang diagnostic ignored "-Wsign-compare"
+#pragma clang diagnostic ignored "-Wextra"
+#if (__clang_major__ > 3) || ((__clang_major__) == 3 && (__clang_minor__ >= 5))
```
brianmoose: Sorry, should have caught this before, the parens on the second clause seem off, maybe:
`#if (__clang_major__ > 3) || ((__clang_major__ == 3) && (__clang_minor__ >= 5))`

include/utils/ignore_warnings.h
```diff
@@ -32,8 +32,12 @@
 #pragma clang diagnostic ignored "-Wnested-anon-types"
 #pragma clang diagnostic ignored "-Wsign-compare"
 #pragma clang diagnostic ignored "-Wunused-private-field"
+#pragma clang diagnostic ignored "-Wsign-compare"
+#pragma clang diagnostic ignored "-Wextra"
```
brianmoose: Looks like trilinos also has some `-Wexpansion-to-defined`

brianmoose: Oops, that should be added to the GCC section

brianmoose:
Looks like `-Wexpansion-to-defined` that we get on the MOOSE trilinos build can't be ignored:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=84127
and 
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53431

Another option is for libmesh to include the trilinos include directory with `-isystem`

include/utils/ignore_warnings.h
jwpeterson: I agree the paren is off, but I don't think it matters because `==` takes precedence over `&&` anyway.

jwpeterson:
OK, cool. On that other thread I noticed they also were able to fix the problem by passing `-Wno-expansion-to-defined`? This would be a must shorter (2-3 line) fix that would probably be easier to get merged quickly if you think it's worth a try.

brianmoose:
You want to set that in the default cxx flags? That would silence that warning for all files when we just want to silence it for trilinos headers.

jwpeterson:
> You want to set that in the default cxx flags?

Hmm... [from what I have read](https://gcc.gnu.org/ml/gcc-patches/2016-08/msg00738.html), it sounds like it is actually a pretty useful warning (I didn't realize you could not use `defined` outside of `#if`) so you're right, we probably shouldn't blanket disable it like that.

include/utils/ignore_warnings.h
brianmoose: Yeah, those parens just don't do anything.
Also, it looks like `-Winconsistent-missing-override` doesn't work on clang 3.5.1 which is our current minimum clang version.
I couldn't pinpoint what clang version it was added. I saw the commit [here](https://reviews.llvm.org/rL218925) but not sure when it made it in

jwpeterson: I've seen a [couple](https://github.com/CGAL/cgal/issues/2026)  [sources](https://stackoverflow.com/questions/29145476/requiring-virtual-function-overrides-to-use-override-keyword) that say "since 3.5" but since that's apparently not quite right let's try the next point release and restrict that line to clang 3.6.0 and greater?

brianmoose:
I confirmed, by changing libmesh-config, that using `-isystem` on the trilinos include directory eliminates all the warnings we were seeing.
So I think the only change needed here is just the clang version guard around `-Winconsistent-missing-override`

jwpeterson:
> I confirmed, by changing libmesh-config, that using -isystem on the trilinos include directory eliminates all the warnings we were seeing.

OK, thanks for testing that. I agree that change can go in a separate PR.

include/utils/ignore_warnings.h
brianmoose: Looks like it went in on 3.6.0
[3.5.2](https://llvm.org/svn/llvm-project/cfe/tags/RELEASE_352/final/include/clang/Basic/DiagnosticGroups.td) does not have it
but [3.6.0](https://llvm.org/svn/llvm-project/cfe/tags/RELEASE_360/final/include/clang/Basic/DiagnosticGroups.td) does

permcody:
I'm not updating anymore. Go nuts

jwpeterson:
Replaced by #1605.

Issue #1596, opened 2018-02-19
Support for Petsc 3.4.5


kevindugan:
Some of the recent updates to Libmesh do not account for older versions of Petsc. 

I'm proposing adding more checks on the Petsc version, specifically in the PetscNonlinearSolver::solve method.

jwpeterson:
@kevindugan sorry about this, we definitely try to maintain backwards compatibility for older versions of PETSc. There have been some recent changes in petsc_nonlinear_solver.C, but note that some of them were removed in #1587, so if you are using the libmesh version currently in MOOSE's submodule (cdeee5648) the problem may already be fixed?

kevindugan:
@jwpeterson I completely understand the difficulty in keeping backward compatibility. I'm actually pulling a fresh version of MOOSE with the new submodule. When I build with my older version of Petsc, I get segmentation faults while running the MOOSE test suite inside of the SNESSetUp routine (which made an appearance in petsc_nonlinear_solver.C). 

My initial thought was that this routine didn't exist in v3.4.5 of Petsc, but now that I dig deeper I see that it does exist. Do you think this may be a MOOSE issue instead of a LibMesh issue?

lindsayad:
@kevindugan Can you pull libmesh/master? The most recent changes in `PetscNonlinearSolver::solve` post-date the submodule update. You should still be able to build MOOSE with libmesh/master

kevindugan:
@lindsayad Yes. I'll report what happens.

lindsayad:
The last commit that touched the SNESSetUp interface was in 2010, so I'm thinking it was probably another change in that method

jwpeterson:
> My initial thought was that this routine didn't exist in v3.4.5 of Petsc, but now that I dig deeper I see that it does exist.

Right... you would have had a compilation issue if the routine didn't exist!

> so if you are using the libmesh version currently in MOOSE's submodule (cdeee56) the problem may already be fixed?

Sorry, this sentence was confusing. What I meant was what Alex said: the problem may be fixed in libmesh master, although I see now that libmesh master still has a call to `SNESSetUp`. If you are being forced to rebuild anyway, you might consider building a debug-enabled version of PETSc to help you track down the segfault.

kevindugan:
> Right... you would have had a compilation issue if the routine didn't exist!

Fair point!

[...]

Issue #1593, opened 2018-02-16
v1.3.0 branch started


jwpeterson:
Tag this issue in the commit message of any commits you'd like cherry-picked over.

As usual, let's try to restrict the cherry-picked commits to bugfixes as much as possible and fight the urge to sneak in new features at the last minute.

PR #1602, opened 2018-02-26
Avoid SNESSetUp/SNESGetJacobian/MatMFFDSetFunction calls in PETSc < 3.5.0.

jwpeterson:
Refs #1596.

cc: @kevindugan

### DIFF
--- a/src/solvers/petsc_nonlinear_solver.C
+++ b/src/solvers/petsc_nonlinear_solver.C
@@ -883,6 +883,19 @@ PetscNonlinearSolver<T>::solve (SparseMatrix<T> &  pre_in,  // System Preconditi
       this->_solver_configuration->configure_solver();
     }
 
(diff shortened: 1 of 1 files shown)


jwpeterson:
I'm going to merge this because I've tested it with PETSc 3.4.5 locally on my system and nearly all the MOOSE tests pass, the remaining ones are fixed by restricting them to run with newer PETSc over in idaholab/moose#10910.

jwpeterson:
I'd also classify this as a bugfix that should be cherry-picked onto v1.3.0 (#1593) since that branch will currently segfault for anyone using nonlinear solvers with PETSc < 3.5.0...

roystgnr:
> I'd also classify this as a bugfix that should be cherry-picked onto v1.3.0

Definitely, thanks.

PR #1603, opened 2018-02-27
find_local_indices() for Parmetis setup

roystgnr:
This PR replaces the "find_global_indices over and over for each rank's elements" strategy with a new "find_local_indices for your own elements, then sync" strategy.  It turns out that one round of communication is much, much cheaper than n_processors rounds.

On the test case from @bboutkov, refining to ~2.5M elements on 320 processors, the combination of this PR and #1600 speeds up repartitioning by a little more than an order of magnitude, which IMHO qualifies as a fix for #1598.  There's definitely a little room for more improvement, but probably "10%" not "10x".

This probably shouldn't be backported to 1.3.0 since it's a performance fix rather than a correctness fix and since it makes a major change to library internals.  On the other hand, *order of magnitude*, so I could be talked into changing my mind.

### DIFF
--- a/include/partitioning/parmetis_partitioner.h
+++ b/include/partitioning/parmetis_partitioner.h
@@ -23,11 +23,11 @@
 // Local Includes
 #include "libmesh/id_types.h"
 #include "libmesh/partitioner.h"
-#include "libmesh/vectormap.h"
 #include "libmesh/auto_ptr.h" // libmesh_make_unique
 
 // C++ Includes
 #include <cstddef>
+#include <unordered_map>
 #include <vector>
 
 namespace libMesh
@@ -117,7 +117,7 @@ private:
   /**
    * Maps active element ids into a contiguous range, as needed by ParMETIS.
    */
-  vectormap<dof_id_type, dof_id_type> _global_index_by_pid_map;
+  std::unordered_map<dof_id_type, dof_id_type> _global_index_by_pid_map;
 
   /**
    * Pointer to the Parmetis-specific data structures.  Lets us avoid
--- a/src/mesh/mesh_communication_global_indices.C
+++ b/src/mesh/mesh_communication_global_indices.C
@@ -648,6 +648,46 @@ void MeshCommunication::check_for_duplicate_global_indices (MeshBase &) const
 
 #if defined(LIBMESH_HAVE_LIBHILBERT) && defined(LIBMESH_HAVE_MPI)
 template <typename ForwardIterator>
+void MeshCommunication::find_local_indices (const libMesh::BoundingBox & bbox,
+                                            const ForwardIterator & begin,
+                                            const ForwardIterator & end,
+                                            std::unordered_map<dof_id_type, dof_id_type> & index_map) const
+{
+  LOG_SCOPE ("find_local_indices()", "MeshCommunication");
+
+  // This method determines id-agnostic local indices
+  // for nodes and elements by sorting Hilbert keys.
+
+  index_map.clear();
+
+  //-------------------------------------------------------------
+  // (1) compute Hilbert keys
+  // These aren't trivial to compute, and we will need them again.
+  // But the binsort will sort the input vector, trashing the order
+  // that we'd like to rely on.  So, two vectors...
+  std::map<Parallel::DofObjectKey, dof_id_type> hilbert_keys;
+  {
+    LOG_SCOPE("local_hilbert_indices", "MeshCommunication");
+    for (ForwardIterator it=begin; it!=end; ++it)
+      {
+#ifdef LIBMESH_ENABLE_UNIQUE_ID
+        const Parallel::DofObjectKey hi(get_hilbert_index ((*it), bbox));
+#else
+        const Parallel::DofObjectKey hi(get_hilbert_index ((*it), bbox), (*it)->unique_id());
+#endif
+        hilbert_keys.emplace(hi, (*it)->id());
+      }
+  }
+
+  {
+    dof_id_type cnt = 0;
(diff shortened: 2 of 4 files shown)


roystgnr:
@jwpeterson, I'd be curious to see what this does on your systems too.  Unlike with #1600 I'd like to merge ASAP, since in this case I'm confident we'll see speedup on any MPI stack, but it would still be interesting to know whether the 10x on my cluster corresponds to 2x or 20x on yours.  These commits are also on the roystgnr/partitioning_speed branch, so you can test the before/post-1600/post-1603 performance by just trying out introduction_ex3 at HEAD~3, HEAD~2, and HEAD from that branch.

pbauman:
Awesome, great improvement. I was hoping we could do something along these lines (local work, then one communication).

roystgnr:
> merge ASAP

By which I mean "after tests all pass", of course.  Those failures don't look like false positives.  I'll see if I can replicate them.

roystgnr:
And once that's done I think it's time for some processor sweeps to be safe...

jwpeterson:
> This probably shouldn't be backported to 1.3.0 since it's a performance fix rather than a correctness fix

Agreed.

> @jwpeterson, I'd be curious to see what this does on your systems too.

OK, I will try and look at this and #1600 today, sorry I haven't had time to get to it yet.

include/partitioning/parmetis_partitioner.h
```diff
@@ -117,7 +117,7 @@ class ParmetisPartitioner : public Partitioner
   /**
    * Maps active element ids into a contiguous range, as needed by ParMETIS.
    */
-  vectormap<dof_id_type, dof_id_type> _global_index_by_pid_map;
```
jwpeterson: Looks like `ParmetisPartitioner` and `MetisPartitioner` might be the only places where `vectormap` is currently used. If we eventually replace the one in `MetisPartitioner` as well, I think we should probably deprecate the whole class and remove it completely some day?

roystgnr: Hmm... maybe.  IIRC Ben wrote it for a use where std::map would have been inefficient at a time when we weren't guaranteed to have unordered_map available, so maybe it's obsolete now?

I'd like to swap it in and out and test performance vs unordered_map on a few systems before we go to that trouble, though, I think.

roystgnr:
Starting the distributed mesh sweep now while I wait for my own MOOSE rebuild.  I can't believe I'm hoping for test failures, but I bet I could use more or easier leads to follow up on...

src/mesh/mesh_communication_global_indices.C
```diff
@@ -891,6 +927,19 @@ void MeshCommunication::find_global_indices (const Parallel::Communicator & comm
   libmesh_assert_equal_to(index_map.size(), n_objects);
 }
 #else // LIBMESH_HAVE_LIBHILBERT, LIBMESH_HAVE_MPI
+template <typename ForwardIterator>
+void MeshCommunication::find_local_indices (const libMesh::BoundingBox &,
+                                            const ForwardIterator & begin,
+                                            const ForwardIterator & end,
+                                            std::vector<dof_id_type> & index_map) const
```
roystgnr: This needs to be fixed; I wrote the fallback code here before settling on the final API and I forgot to update it after.

roystgnr:
Found the bug, trying out a fix now, hope to have it fixed and tested later this afternoon.

roystgnr:
> Found the bug

Did I say "bug", singular?  I mean "Found the bug**s**", of course.

> OK, I will try and look at this and #1600 today

Looking at #1600 would be good; that code seems solid.

Looking at this PR right now is only a good idea if you want a chance to laugh at me before I have time to ```rebase -i``` away the evidence.

jwpeterson:
> Did I say "bug", singular? I mean "Found the bugs", of course.

Careful, this now makes it sound like you have definitely found *all* of them. 😉

roystgnr:
Oh, I definitely didn't mean to imply that.  Two bugs fixed so far and I'm still not passing that handful of tests.  I retract "I could be talked into changing my mind"; even after I get this working we're *definitely* not putting it straight into a release.

roystgnr:
I just force-pushed the current version, which fixes 4 embarrassingly obvious bugs, succeeds at the previously assertion-failing test on my cluster, and changes the 320 proc 2.5M element ParmetisPartitioner speedup from 1OOM to 2OOM.

If this passes all tests (both Civet and my own) then I'd love to merge to master ASAP, because the performance improvement is phenomenal.  1.3.0 will have to do without, though, because somehow that big wad of bugs still passed 99.5% of MOOSE tests, which makes me wonder whether there might be some even more subtle corner case that 100% of our test coverage misses.

@pbauman, it looks like the GRINS tests in CI include a 4 processor run but no DistributedMesh runs?  I'll do a GRINS processor sweep with DistributedMesh by hand, but one of these days we should add stuff like that as additional optional Civet recipes on femputer.

pbauman:
> @pbauman, it looks like the GRINS tests in CI include a 4 processor run but no DistributedMesh runs?

I think there's one or 2 in there, but the vast majority are ReplicatedMesh

 > I'll do a GRINS processor sweep with DistributedMesh by hand, 

Thanks

> but one of these days we should add stuff like that as additional optional Civet recipes on femputer.

Yeah, it's on my todo list. :( I'll try and make it a priority.

roystgnr:
@jwpeterson in #1598 reports only 1OOM rather than 2OOM, and his system is probably a much better reflection of real-world performance.  But that's still pretty nice, so I'll merge this as soon as I've done a little more private testing to raise my confidence levels.

PR #1604, opened 2018-02-28
Use -isystem on Trilinos include directories

brianmoose:
@jwpeterson mentioned he might want to do this for other libraries (like boost)

Primarily put this up to verify that it silences all the warnings in moose. I created and added a "Test Trilinos" to this PR that has `-Wall -Wextra` on the moose build steps.

### DIFF
--- a/configure
+++ b/configure
@@ -35387,7 +35387,7 @@ $as_echo "<<< Configuring library with Trilinos Epetra support >>>" >&6; }
 
     #echo "Makefile_config_trilinos="
     #cat Makefile_config_trilinos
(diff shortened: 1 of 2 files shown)


brianmoose:
Looks like Trilinos on linux gcc has an additional `-I/usr/include`  which translates to
`-isystem /usr/include` which seems to cause [problems](https://stackoverflow.com/questions/37218953/isystem-on-a-system-include-directory-causes-errors)

Don't know a good way to handle this since we really don't know what include directories Trilinos will add.

I think for now we might just solve this problem by patching trilinos in the moose environment package....

PR #1606, opened 2018-03-01
Bugfix for ExactSolution::_compute_error with NodeElems in the mesh

dknez:
ExactSolution::_compute_error was reporting large errors for me in a mesh that included NodeElems, even though the solutions matched almost perfectly. Skipping NodeElems in the element loop fixes the issue for me.

I'm not completely sure what goes wrong when the NodeElems are included in the loop, I didn't investigate that in detail.

### DIFF
--- a/src/error_estimation/exact_solution.C
+++ b/src/error_estimation/exact_solution.C
@@ -701,6 +701,12 @@ void ExactSolution::_compute_error(const std::string & sys_name,
               // FIXME: Needs to be updated for vector-valued elements
               DenseVector<Number> output(1);
               (*coarse_values)(q_point[qp],time,output,&subdomain_id);
+
+              if(output.size() == 0)
+              {
+                libmesh_error_msg("Mesh function failed to find point");
+              }
+
               exact_val = output(0);
             }
           const typename FEGenericBase<OutputShape>::OutputNumber val_error = u_h - exact_val;


roystgnr:
This seems like it can introduce a new bug, if users have a mesh with NodeElem points whose values may be in error, right?

I'd rather track down and fix what's going wrong, if we can.  Can you set up a test case that replicates the problem?

dknez:
Yeah, we should track down the bug... OK, I'll stop being lazy and set up a test case!

jwpeterson:
What norm were you using? Gradient-based norms probably make sense for NodeElems?

dknez:
I was using L2 norm

dknez:
I ran my test case in debug mode, and it hits an assertion:

```
Assertion `_out_of_mesh_mode == true' failed.
_out_of_mesh_mode = 0
true = 1


Stack frames: 18
1: libMesh::MacroFunctions::report_error(char const*, int, char const*, char const*)
2: libMesh::PointLocatorTree::operator()(libMesh::Point const&, std::__debug::set<int, std::less<int>, std::allocator<int> > const*) const
3: libMesh::MeshFunction::find_element(libMesh::Point const&, std::__debug::set<int, std::less<int>, std::allocator<int> > const*) const
4: libMesh::MeshFunction::operator()(libMesh::Point const&, double, libMesh::DenseVector<double>&, std::__debug::set<int, std::less<int>, std::allocator<int> > const*)
5: void libMesh::ExactSolution::_compute_error<double>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__debug::vector<double, std::allocator<double> >&)
6: libMesh::ExactSolution::compute_error(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
7: /home/dknez/akselos-dev/scrbe/build/bin/component_visualization-dbg_real() [0x54d5de]
8: /home/dknez/akselos-dev/scrbe/build/bin/component_visualization-dbg_real() [0x54c57d]
9: /home/dknez/akselos-dev/scrbe/build/bin/component_visualization-dbg_real() [0x5b8d9e]
10: MessageHandler::dispatch_message(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, Json::Value const&, std::unique_ptr<MessageHandler, std::default_delete<MessageHandler> >&)
11: MessageHandler::extract_message(Json::Value&, std::unique_ptr<MessageHandler, std::default_delete<MessageHandler> >&)
12: MessageHandler::parse_message(char const*, char const*, std::unique_ptr<MessageHandler, std::default_delete<MessageHandler> >&)
13: /home/dknez/akselos-dev/scrbe/build/bin/component_visualization-dbg_real() [0x5e14ed]
14: /home/dknez/akselos-dev/scrbe/build/bin/component_visualization-dbg_real() [0x5d95a7]
15: /home/dknez/akselos-dev/scrbe/build/bin/component_visualization-dbg_real() [0x5be051]
16: __libc_start_main
17: /home/dknez/akselos-dev/scrbe/build/bin/component_visualization-dbg_real() [0x546d79]
[0] /home/dknez/software/libmesh-src/src/utils/point_locator_tree.C, line 236, compiled Mar  1 2018 at 12:18:02
```

When I skip the NodeElems it runs fine in both dbg and opt modes. Looks like the issue is that the PointLocator that ExactSolution uses fails with the NodeElems.

dknez:
I've tried some other cases with NodeElems and I don't hit the problem with those, so this doesn't seem to be a general problem with NodeElems in ExactSolution.

Also, this issue is apparently related to MeshFunction and PointLocator, not ExactSolution per se, so I don't think the fix I proposed is relevant. My preference would be to close this PR without merging, and separately I can look into what causes the PointLocator to fail in my specific case (I think I could increase the point locator tolerance, or something).

roystgnr:
There shouldn't be any point locator tolerance issues on a NodeElem, though!  So I suspect that just increasing the tolerance would again be a workaround that would mask some real underlying error.

I assume your ExactSolution is coming from a refined mesh, and that's what's causing it to go to the MeshFunction code path?  I've never actually tried doing uniform refinement on a NodeElem, but it's supposed to just create a single child element associated with the same node.  So trying to evaluate on a coarse NodeElem should query the precise point of the fine NodeElem!

dknez:
I'm not using a refined mesh, I'm comparing one solution with another (reference) solution on the same mesh. I agree that the point locator issue I was hitting was surprising, but I can't replicate it with a simpler model so it's not straightforward to debug further.

dknez:
Actually, I've updated this PR so that it just throws an error if the mesh function fails to find a point. I think this is an improvement compared to before --- I much prefer hitting an error than getting confusing/incorrect results from compute_error.

dknez:
Actually, I found the cause of this issue, and it was a bug in my code, not in libMesh.

The issue was that the NodeElems in one mesh and the NodeElems in the other mesh had different subdomain IDs, and we pass the subdomain ID in the call to the MeshFunction inside compute_error, so that's why the mesh function evaluation failed.

So I think this PR is not needed after all, and I'd be OK to just close it without merging. Let me know if you guys agree.

dknez:
As per my previous comment, I'll go ahead and close this.

PR #1609, opened 2018-03-05
Possibly fix issue with "EDGE" elements in Exodus files.

jwpeterson:
Since the Exodus file format is completely customizable, we don't
always know exactly what strings will be present in Exodus files
written by a particular mesh generator. According to #1546, it sounds
like the Python meshio/pygmsh writes "EDGE" element, so this commit
adds another equivalence to our EDGE2 type for this string.

Refs #1546.

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -61,6 +61,7 @@ void init_element_equivalence_map()
       element_equivalence_map["SPHERE"] = NODEELEM;
 
       // EDGE2 equivalences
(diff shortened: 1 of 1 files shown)


roystgnr:
Looks good, but let's wait to hear from @Chaztikov and make sure it works before merging?

jwpeterson:
> Looks good, but let's wait to hear from @Chaztikov and make sure it works before merging?

I'd rather not, that issue is from last December so he may not even be working on the same problem any more... I will leave the issue open, however.

PR #1611, opened 2018-03-06
Use unique_id() iff we have it, not iff we don't!

### LINKED ISSUE #1610: No Unique Indices Compile Issue (--disable-unique-id)
@roystgnr, I think this line is in error because the corresponding #ifdef lives inside the get_hilbert_index method?  I'm a little hesitant to submit the fix because I don't have anything to test it.  Can you confirm or deny my suspicion?

https://github.com/libMesh/libmesh/blob/ed3dc66c224a59f03bc4d23dffc3512bfc79a51f/src/mesh/mesh_communication_global_indices.C#L676

Specifically, there's no 3-argument implementation of get_hilbert_index required.
```
Parallel::DofObjectKey
get_hilbert_index (const Elem * e,
                   const libMesh::BoundingBox & bbox)
{
  static const unsigned int sizeof_inttype = sizeof(Hilbert::inttype);

  Hilbert::HilbertIndices index;
  CFixBitVec icoords[3];
  Hilbert::BitVecType bv;
  get_hilbert_coords (e->centroid(), bbox, icoords);
  Hilbert::coordsToIndex (icoords, 8*sizeof_inttype, 3, bv);
  index = bv;

  return std::make_pair(index, e->unique_id());
#else
  return index;
#endif
}



// Compute the hilbert index
Parallel::DofObjectKey
get_hilbert_index (const Node * n,
                   const libMesh::BoundingBox & bbox)
{
  static const unsigned int sizeof_inttype = sizeof(Hilbert::inttype);

  Hilbert::HilbertIndices index;
  CFixBitVec icoords[3];
  Hilbert::BitVecType bv;
  get_hilbert_coords (*n, bbox, icoords);
  Hilbert::coordsToIndex (icoords, 8*sizeof_inttype, 3, bv);
  index = bv;

  return std::make_pair(index, n->unique_id());
#else
  return index;
#endif
}
```

roystgnr:
Hopefully this fixes #1610 

We clearly need some test coverage of the use case this code is intended to handle: meshes where elements and/or nodes overlap precisely, forcing us to use unique_id() to disambiguate between identical hilbert indices when sorting.

### DIFF
--- a/src/mesh/mesh_communication_global_indices.C
+++ b/src/mesh/mesh_communication_global_indices.C
@@ -670,11 +670,7 @@ void MeshCommunication::find_local_indices (const libMesh::BoundingBox & bbox,
     LOG_SCOPE("local_hilbert_indices", "MeshCommunication");
     for (ForwardIterator it=begin; it!=end; ++it)
       {
-#ifdef LIBMESH_ENABLE_UNIQUE_ID
         const Parallel::DofObjectKey hi(get_hilbert_index ((*it), bbox));
-#else
-        const Parallel::DofObjectKey hi(get_hilbert_index ((*it), bbox), (*it)->unique_id());
-#endif
         hilbert_keys.emplace(hi, (*it)->id());
       }
   }
--- a/src/mesh/mesh_communication_global_indices.C
+++ b/src/mesh/mesh_communication_global_indices.C
@@ -74,7 +74,7 @@ void get_hilbert_coords (const Point & p,
 
 
 Parallel::DofObjectKey
-get_hilbert_index (const Elem * e,
+get_dofobject_key (const Elem * e,
                    const libMesh::BoundingBox & bbox)
 {
   static const unsigned int sizeof_inttype = sizeof(Hilbert::inttype);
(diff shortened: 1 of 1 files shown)


roystgnr:
We actually have a unit test with a slit mesh, but I guess that only creates overlapping nodes, whereas find_local_indices currently only gets called to sort elements?

roystgnr:
And... I'm still an idiot.  get_hilbert_index() already handles the unique_id() call.

roystgnr:
In my defense, get_hilbert_index has a confusingly atavistic name which doesn't properly reflect its current return type.

Of course, the reason for that is because I didn't change the name back when I changed the return type.  Past-Roy is always screwing things up for me.

That's an anonymous namespace function, so there's no API/ABI issues, so I'll change the name to get_dofobject_key... in another commit, which I won't push until after the current CI and my local compile of the second commit are both complete.

roystgnr:
Although I don't *think* ```:%s/get_hilbert_index/get_dofobject_key``` could have possibly broken anything, under the embarrassing circumstances I think I'll wait until CI is done again before merging.

PR #1613, opened 2018-03-07
Make PointLocatorTree threadsafe again

### LINKED ISSUE #1612: PointLocatorTree may not be thread safe at all
Servant point locators of the `PointLocatorTree` type share the `Tree` allocated by the master point locator. `TreeNode` contains a `mutable std::vector<bool> searched_child;` member which is set and resized (!) in `TreeNode<N>::find_element_in_children`, which in turn is called by the point locator `operator ()` (which obviously can be called from parallel threads).

This looks bad, but maybe it means that I am not "full of shit" after all @friedmud ...

dschwen:
Implements @roystgnr's original suggestion for using a stack based `std::array` instead of a `mutable` member vector. I confirmed this to fix the threading crashes in magpie which heavily pounds this PL in a threaded region.

Closes #1612

### DIFF
--- a/include/utils/tree_node.h
+++ b/include/utils/tree_node.h
@@ -220,11 +220,6 @@ private:
    * Does this node contain any infinite elements.
    */
   bool contains_ifems;
-
-  /**
-   * Used in find_element_in_children
-   */
-  mutable std::vector<bool> searched_child;
 };
 
 
--- a/src/utils/tree_node.C
+++ b/src/utils/tree_node.C
@@ -19,6 +19,7 @@
 
 // C++ includes
 #include <set>
+#include <array>
 
 // Local includes
 #include "libmesh/libmesh_config.h"
@@ -498,7 +499,8 @@ const Elem * TreeNode<N>::find_element_in_children (const Point & p,
 {
   libmesh_assert (!this->active());
 
-  searched_child.assign(children.size(), false);
+  // value-initialization sets all array members to false
+  std::array<bool,N> searched_child{};
 
   // First only look in the children whose bounding box
   // contain the point p.


dschwen:
@roystgnr, the initialization in ea67fe7 was wrong. This works for raw C arrays, but not for `std::array`. I have since learned about _value-initialization_.  e3885d9 should be the correct way of zero (i.e. _false_) initializing the _entire array (the old version just set the first element).

permcody:
OK - good, I was just about to comment when you pushed the new version. I'm not sure I like the new version much better, but if it works, fine...

Can you do this? `std::array<bool,N> searched_child(false, N)`;

dschwen:
I'm guessing value-initialization is potentially faster and I did leave a comment...

dschwen:
Actually, your constructor doesn't even exist. So there's that...

jwpeterson:
I do agree that it looks odd. 

My main concern is whether our "minimum" compilers correctly support this syntax since it would be the first use of it in the library.

dschwen:
I guess the test harness will find out. There is a minimum compiler test, right?

jwpeterson:
> There is a minimum compiler test, right?

There's optional ones which I just added.

roystgnr:
I dug into page after page of cppreference.com trying to figure out if that was a valid trick to play, and I *thought* I'd convinced myself it was.  Sorry I missed the array-vs-std::array distinction.

jwpeterson:
It looks like [this](https://stackoverflow.com/questions/31181568/does-stdarray-default-initialize-or-value-initialize) might be another valid value initialization syntax, IMO it's even weirder looking than your current version.
```
auto searched_child = std::array<bool, N>();
```

jwpeterson:
The C++11 issue I encountered previously on this line required me to use double curly braces to *assign* values.

https://github.com/libMesh/libmesh/blob/5ed60acfd248ddf5357243f6df6c2cbc8f596210/src/geom/face_quad.C#L230

This is discussed on the [cppreference std::array page](http://en.cppreference.com/w/cpp/container/array), but may not be relevant to this new line of code you're adding...

dschwen:
@roystgnr I resorted to writing a test code and the results are quite surprising. I tested this under clang and g++

```
  std::array<int,4> a{};
  std::array<int,4> b;
  std::array<int,4> c{1};
  std::array<int,4> d = {1};
```

```
[13][schwd][~/Programs]> clang++ --std=c++11 -o test1 test1.C
[13][schwd][~/Programs]> ./test1
a = 0 0 0 0
b = 898 899 900 901
c = 1 0 0 0
d = 1 0 0 0
[13][schwd][~/Programs]> g++ --std=c++11 -o test1 test1.C
[13][schwd][~/Programs]> ./test1
a = 0 0 0 0
b = 896 897 898 899
c = 1 0 0 0
d = 1 0 0 0
```

Code here: the trick is to spam the stack with non-zero values first to see where initialization occurs

```
#include <iostream>
#include <array>

void spamStack()
{
  // fill stack
  constexpr int N = 1000;
  std::array<int,N> z;

  for (int i = 0; i < N; i++)
    z[i] = i;
}

void output(std::array<int,4> & a, char aname)
{
  std::cout << aname << " = ";
  for (int i = 0; i < 4; i++)
    std::cout << a[i] << ' ';
  std::cout << '\n';
}

void test()
{
  // pad
  std::array<int,100> z;

  // sample
  std::array<int,4> a{};
  std::array<int,4> b;
  std::array<int,4> c{1};
  std::array<int,4> d = {1};

  output(a, 'a');
  output(b, 'b');
  output(c, 'c');
  output(d, 'd');
}

int main()
{
  spamStack();
  test();
  return 0;
}
```

dschwen:
Btw with 

```
  std::array<int,4> a;
  std::array<int,4> b;
  std::array<int,4> c;
  std::array<int,4> d;

```

The outputs are:

```
[13][schwd][~/Programs]> clang++ --std=c++11 -o test1 test1.C
[13][schwd][~/Programs]> ./test1
a = 902 903 904 905
b = 898 899 900 901
c = 894 895 896 897
d = 890 891 892 893
[13][schwd][~/Programs]> g++ --std=c++11 -o test1 test1.C
[13][schwd][~/Programs]> ./test1
a = 900 901 902 903
b = 896 897 898 899
c = 892 893 894 895
d = 888 889 890 891
```

jwpeterson:
Is the flag `--std=c++11` legit? I thought it was 1 dash.

jwpeterson:
If you believe the cppreference page, this one
```
std::array<int,4> c{1};
```
is technically C++14 only, but newer compilers may simply allow it.

dschwen:
Both newer gcc and clang accept the double dash version, but it actually was a typo on my part :o)

jwpeterson:
> Both newer gcc and clang accept the double dash version

TIL

jwpeterson:
The new code warns produces a `-Wmissing-field-initializers` warning on GCC 4.9.2.
```
../src/utils/tree_node.C: In instantiation of ‘const libMesh::Elem* libMesh::TreeNode<N>::find_element_in_children(const libMesh::Point&, const std::__debug::set<short unsigned int>*, libMesh::Real) const [with unsigned int N = 2u; libMesh::Real = double]’:
../src/utils/tree_node.C:554:16:   required from here
../src/utils/tree_node.C:503:37: warning: missing initializer for member ‘std::__debug::array<bool, 2ul>::_M_elems’ [-Wmissing-field-initializers]
   std::array<bool,N> searched_child{};
                                     ^
../src/utils/tree_node.C: In instantiation of ‘const libMesh::Elem* libMesh::TreeNode<N>::find_element_in_children(const libMesh::Point&, const std::__debug::set<short unsigned int>*, libMesh::Real) const [with unsigned int N = 4u; libMesh::Real = double]’:
../src/utils/tree_node.C:555:16:   required from here
../src/utils/tree_node.C:503:37: warning: missing initializer for member ‘std::__debug::array<bool, 4ul>::_M_elems’ [-Wmissing-field-initializers]
../src/utils/tree_node.C: In instantiation of ‘const libMesh::Elem* libMesh::TreeNode<N>::find_element_in_children(const libMesh::Point&, const std::__debug::set<short unsigned int>*, libMesh::Real) const [with unsigned int N = 8u; libMesh::Real = double]’:
../src/utils/tree_node.C:556:16:   required from here
../src/utils/tree_node.C:503:37: warning: missing initializer for member ‘std::__debug::array<bool, 8ul>::_M_elems’ [-Wmissing-field-initializers]
```

While the warning is certainly the product of a not-quite-C++11-conforming compiler, it's also pretty annoying. If this older compiler is happier with the alternative but equivalent syntax
```
auto searched_child = std::array<bool, N>();
```
I think I might change to that.

PR #1616, opened 2018-03-07
Datatype leak fix

roystgnr:
This ought to fix at least some of, and with luck all of, the leaks caught by https://github.com/idaholab/moose/issues/10937

I'm running tests locally now, but we should also put this through a DistributedMesh processor count sweep here before we trust it.

### DIFF
--- a/include/parallel/parallel.h
+++ b/include/parallel/parallel.h
@@ -335,14 +335,16 @@ public:
   void commit ()
   {
 #ifdef LIBMESH_HAVE_MPI
-    MPI_Type_commit (&_datatype);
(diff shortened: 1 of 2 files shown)


roystgnr:
@jwpeterson reports that getting rid of the leaky "optimization" doesn't make a noticeable difference in his timings, using on the "hammer libMesh internal communication as much as possible" test @bboutkov set up.

https://xkcd.com/1691/

jwpeterson:
FYI I'm also using this PR as a guinea pig for a newly-added "--disable-unique-id" test on CIVET.

PR #1618, opened 2018-03-07
Avoid -Wmissing-field-initializers warning on GCC 4.9.2.

jwpeterson:
The previous syntax is correct and should be accepted by C++11
conforming compilers, unfortunately we still have to support some
not-quite-standards-conforming compilers.

Refs #1613.

### DIFF
--- a/src/utils/tree_node.C
+++ b/src/utils/tree_node.C
@@ -500,7 +500,7 @@ const Elem * TreeNode<N>::find_element_in_children (const Point & p,
   libmesh_assert (!this->active());
 
   // value-initialization sets all array members to false
-  std::array<bool,N> searched_child{};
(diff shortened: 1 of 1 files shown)


roystgnr:
Does "= {}" instead of plain "{}" avoid the warning?

If not, I'm ok with this idiom in this case... but I think it might force a copy construction, and that with more complicated or non-header-only classes instead of std::array that copy construction might not optimize away, so it would be nice not to get into the habit.

jwpeterson:
> Does "= {}" instead of plain "{}" avoid the warning?

No, "= {}" produces the same warning as "{}" in GCC 4.9.2.

> so it would be nice not to get into the habit.

I think this weirdness is particular to `std::array`, and I don't see us needing to use it all the time.

Issue #1619, opened 2018-03-08
Copy/paste bug in sparsity_pattern.C, line 306


jwpeterson:
This *looks* like a bug to me (both `it` and `end` are initialized to `begin()`) but it seems like it would have to cause a memory leak and we haven't seen anything in our valgrind testing... it also seems like it's been there a while; the last thing to touch it was an automated reindentation commit.

```
        for (std::set<CouplingMatrix *>::iterator
               it  = temporary_coupling_matrices.begin(),
               end = temporary_coupling_matrices.begin();
             it != end; ++it)
          delete *it;
```

PR #1620, opened 2018-03-08
Fix copy/paste bug in sparsity_pattern.C

jwpeterson:
Refs #1619.


roystgnr:
We ought to replace this with ```set<unique_ptr>``` now that that's guaranteed to be available.

But more urgently, we ought to construct some serious "one ghosting functor creates a non-trivial CouplingMatrix and another creates a non-superset-and-non-subset CouplingMatrix" test cases and make sure this code path is actually working in the first place.  It would be nice if I find the bugs before a user on a deadline does.

Issue #1617, opened 2018-03-07
MPI idiom optimizations


roystgnr:
A few MPI-performance-optimization factoids I've run into over the last week:

1.  @bboutkov reports that, although #1603 is also a huge win on his cluster, #1600 has exactly the opposite effect as it does on PECOS and INL: a slight slowdown rather than a slight speedup!  There's also a lot of variance in timings from run to run.

2.  https://www.clustermonkey.net/MPI/mpi-the-top-ten-mistakes-to-avoid-part-1.html lists the "post an MPI_PROBE and then use the size that is returned to allocate a buffer of the correct size and then MPI_RECV into it" as a *mistake*, saying it often forces allocation of an otherwise unnecessary temporary buffer for the entire message, and recommending our original naive "send a length message first" idiom instead!

3.  https://www.clustermonkey.net/MPI/mpi-the-top-ten-mistakes-to-avoid-part-2.html lists our use of MPI_ANY_SOURCE as another common mistake, claiming it can cause thread contention and unnecessary system calls which can be avoided by using MPI_Waitany() instead.

I'm not sure *any* of this means we should be changing our code to accommodate, but I am now sure we need some more systematic way of comparing performance.  I'm currently leaning toward using MOOSE, reworking something like test/tests/functions/image_function/threshold_adapt_parallel.i to do a bunch of DistributedMesh adaptivity into some initial condition function, and calling that our "performance standard", but I'm open to other suggestions.

roystgnr:
Oh, almost forgot one!

4.  @pbauman recently reminded me that our current rubric for selecting which processor owns a node, "look at all the attached elements and find the element owner of lowest rank", isn't a smart way to load balance.  I don't believe there's ever going to be *much* of an imbalance, at least not on problems large enough to warrant using tons of processors, but maybe it's time to try a change and test that belief.

permcody:
Actually, @fdkong has been looking at this very issue recently and we _DO_ think it's a bigger issue than you might expect. It really starts to matter when you are running a problem with several DOFs per node (e.g. phase_field in 3D so like ~30), combined with suboptimal partitions that we get from Metis and wham! Our scalability goes in the toilet when we get up into the ~1000 processor ranges at ~20K DOFs per proc. A few dozen extra elements on a partition with all of those extra nodes assigned to the same processor leads to load imbalanced in the neighborhood of 10%. I believe Fande has some hard numbers to back this up.

[...]

PR #1621, opened 2018-03-08
Set node processor ids less unevenly

roystgnr:
Hopefully this will reduce the load imbalance we can get with nodes when
using large numbers of processors without extremely large numbers of
elements.

@fdkong, can you try this on the same problem you gave results for in #1617?  I can set up a new benchmark myself but it would be nice to get apples-to-apples numbers on a known-bad case.

### DIFF
--- a/src/mesh/mesh_base.C
+++ b/src/mesh/mesh_base.C
@@ -210,10 +210,12 @@ void MeshBase::prepare_for_use (const bool skip_renumber_nodes_and_elements, con
 
   // Mesh modification operations might not leave us with consistent
   // id counts, but our partitioner might need that consistency.
-  if (!_skip_renumber_nodes_and_elements)
-    this->renumber_nodes_and_elements();
-  else
-    this->update_parallel_id_counts();
+  //
+  // We just call renumber_nodes_and_elements() here because it
+  // currently handles stripping of disconnected nodes and updating of
+  // parallel id counts; it will skip renumbering if the mesh is so
+  // configured.
+  this->renumber_nodes_and_elements();
 
   // Let all the elements find their neighbors
   if (!skip_find_neighbors)
--- a/src/mesh/mesh_tools.C
+++ b/src/mesh/mesh_tools.C
@@ -2077,9 +2077,8 @@ void MeshTools::correct_node_proc_ids (MeshBase & mesh)
 
       SyncNodeSet syncv(valid_nodes, mesh);
 
-      Parallel::sync_node_data_by_element_id
-        (mesh, mesh.elements_begin(), mesh.elements_end(),
-         Parallel::SyncEverything(), Parallel::SyncEverything(), syncv);
+      Parallel::sync_dofobject_data_by_id
+        (mesh.comm(), mesh.nodes_begin(), mesh.nodes_end(), syncv);
     }
 
   // We build up a set of compatible processor ids for each node
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -828,7 +828,18 @@ bool UnstructuredMesh::contract ()
           }
       }
 
-  // Strip any newly-created NULL voids out of the element array
+  // Strip any newly-created NULL voids out of the element array, if
+  // we're allowed to.
+  //
+  // We just call renumber_nodes_and_elements() here because it
+  // currently handles stripping of disconnected nodes and updating of
+  // parallel id counts; it will skip renumbering if the mesh is so
+  // configured.
+  //
+  // Note that renumbering may cause node processor ids to no longer
+  // match an id-number-based canonical processor partitioning
+  // algorithm, but it's probably not a safe time for us to go back
+  // and fix that.
   this->renumber_nodes_and_elements();
 
   // FIXME: Need to understand why deleting subactive children
--- a/src/partitioning/partitioner.C
+++ b/src/partitioning/partitioner.C
@@ -581,6 +581,7 @@ void Partitioner::set_node_processor_ids(MeshBase & mesh)
 
 #ifdef DEBUG
   MeshTools::libmesh_assert_valid_procids<Node>(mesh);
+  MeshTools::libmesh_assert_canonical_node_procids(mesh);
 #endif
 }
 


fdkong:
Thanks, I will take a try.

roystgnr:
Gah.  I figured that the libHilbert canonical ordering would save us from having to worry about backwards compatibility with xda/xdr files, but I didn't even think about solver robustness potentially being an issue.

roystgnr:
But worse, it looks like there's an actual bug being triggered by adjoints_ex4 in some cases.  Definitely not worth trying to benchmark this until I figure that out.

permcody:
OK - this is embarrassingly simple and seems like it'll be highly effective. Nice fix!

roystgnr:
The bug appears to be that uniform refinement and coarsening in AdjointRefinementErrorEstimator (and presumably UniformRefinementErrorEstimator too, though I haven't seen that triggered) no longer leaves us with unchanged node processor ids.

What's baffling me is that, although I see a URefEE bug that would cause this (we forgot to disable renumbering between refinement and re-coarsening), there is no such bug in ARefEE.  Going to have to push my current state and try again tomorrow.

roystgnr:
Ok, I think I get it now.  MeshCommunication::make_node_proc_ids_parallel_consistent and MeshTools::correct_node_proc_ids are *also* setting Node::processor_id(), and when the setting from those functions isn't done in the same way as the Partitioner setting, it causes inconsistencies between node owners before and after refinement.

I think we can skip futzing with make_node_proc_ids_parallel_consistent, for now since:
1) We have to; node ids aren't yet consistent when it's called so we can't rely on them.
2) We'll end up correct in the end, because correct_node_proc_ids gets called afterward.
3) The load imbalance in between (1) and (2) will only affect a couple communication sweeps, not a whole solve.

roystgnr:
The correction for that last failure is definitely going to be a fixup commit, so I'm going to interactive-rebase the whole branch while I'm at it.

roystgnr:
I would chide @permcody for the hubris of the phrase "this is embarrassingly simple", except that I just noticed the phrase "**last** failure" in my ~~last~~, er, previous comment, so who am I to talk.

Nevertheless, I think this might be in working order now; hopefully @fdkong can give it a try as soon as CI gives it a clean bill of health.

permcody:
Well, I meant the simplicity and beauty of the algorithm. I had no idea that the actual change would be this difficult ;)

roystgnr:
Gah.  I was so happy about fixing actual bugs that I completely forgot about the solver sensitivity issues.

roystgnr:
...and, there's still one DistributedMesh issue.  In MeshCommunication::make_nodes_parallel_consistent() we assume that we can sync up new nodes' processor ids first, then ask the owning processors for the canonical DofObject id.  This is necessary because new nodes on a partition border may have been created independently on multiple processors, and it used to work because the processor assignment of a node was independent of the node's id.

Now we have a chicken-and-egg situation, where we can't determine the processor assignment of a node without knowing its id, but with new nodes we still can't determine which of multiple ids to use until we decide which processor owns the node.

One fix would be to make two sweeps when determining processor id: one to get a DofObject-id-independent owner so we know who to query to disambiguate the DofObject id, the second to get the load-balancing-optimal owner.

An alternate fix would be to choose the DofObject id by a different heuristic than "the id assigned by its eventual owner"; I think "the smallest id assigned by any processor ghosting it" would work.

I'm not going to try either of those tonight, though.  IIRC this was some of the ugliest corner case code, because it's easy to have a "fan" mesh where processor A and processor B both have an element linked to node n, but processor A isn't a neighbor of B, only a neighbor-of-a-neighbor-of- to an arbitrary degree.

I really need to write the DistributedMesh paper one of these days.

permcody:
> I really need to write the DistributedMesh paper one of these days.

Not just the paper but also more documentation. Perhaps we can help. A good brain dump of the basic functionality and implications would get us down the road quite a ways. We could also help with running problems, getting results, making graphs, editing. I'm sure a few us would love to contribute to such an endeavor.

roystgnr:
So, ```MeshCommunication::make_nodes_parallel_consistent()``` appears to be unaffected - we essentially already implement my two-sweep fix.  It's ```MeshCommunication::make_new_nodes_parallel_consistent()``` where we lose sync after the first sweep, because we only apply that first sweep to nodes that we suspect might be new.  If one processor can't tell that a node isn't new, but another one can, then the first might set processor_id via the id-independent heuristic and the second will leave things unchanged (using the load-balancing heuristic) and then they'll disagree.

I'd love collaboration on a DistributedMesh paper.  Setting up and running large-scale benchmark problems is honestly the most tedious thing that's dissuading me right now; I'd love to just write up the algorithms and have someone else generate results.  Ask around INL and see who else is interested?

fdkong:
@roystgnr, I definitely will generate some results you may be interested in.  My original plan is to push a Phase-Field application to scale to more than 10K cores.  We have tried with 1K cores, and found that the node assignment did not do a good job. The nodes are imbalanced. When this PR passes these tests, I will try again.

BTW, what extra data (tests) you might be interested in?

fdkong:
@roystgnr, Is the branch  ready for me to take a try?

roystgnr:
@fdkong I'm afraid not.  I'll let you know as soon as it is.

roystgnr:
The last step that's stymieing me, written down to keep my own thoughts straight:

After we finish a projection involving subactive coarsened-away elements, we contract the mesh, eliminating those elements.

After we eliminate elements, unless we've been instructed not to do renumbering, we renumber ids, to save space in any vectors that are indexed by element and/or node number (including the ReplicatedMesh data storage itself).

But, once we renumber node ids, then nodal partitioning **no longer** conforms to an id-based load balancing heuristic.

So we have a few options:

Trying to repartition to restore that conformity would require an extra renumbering of DoF ids too, since we need those to be contiguous on each processor, and *that* would require re-allocating all our vectors.

Avoiding renumbering entirely would basically be a valgrind-proof memory leak, as more and more null DofObject pointers get stuck in our containers.

Not requiring consistent node partitioning currently would make AdjointRefinementErrorEstimator and UniformRefinementErrorEstimator broken, because as soon as a refinement and re-coarsening caused node partitioning to *become* consistent w.r.t. our heuristic, the nodes would then *fail* to be consistent with saved coarse solutions.

Changing ARefEE and UREE to not depend on consistent node partitioning would require us to avoid ever repartitioning nodes, but repartitioning nodes to a consistent heuristic is how DistributedMesh figures out who owns new refinement-generated nodes in the first place!

We might be able to keep new refinement-generated nodes marked as unpartitioned until the last minute, so we can tell which really need to be partitioned and which need to be left alone... but partitioning nodes is a recursive affair (see above "fan" mesh example) so we'll at least need a temporary auxilliary way to mark new nodes during the partition sync'ing process.

friedmud:
I can provide some good news about this PR: it _definitely_ makes a huge difference in nodal load balance.

Here is a plot of HEAD (current) vs this patch for a problem with ~2M elements on 288 procs.  It's showing the number of nodes on each PID:

better_node_partitioning

Here are the stats for the number of nodes per processor:

|           | current     | patched     |
|-----------|-------------|-------------|
| min_nodes | 5534        | 6886        |
| max_nodes | 8401        | 7558        |
| average   | 7112.440972 | 7112.440972 |
| stddev    | 601.8283612 | 161.9056808 |

All-in-all... looking good!

(all this stuff comes via my MOOSE PR over here: https://github.com/idaholab/moose/pull/11216 and thanks to Peacock for the ability to visualize this stuff)

permcody:
Now that’s a cool plot!

jessecarterMOOSE:
Cool! What kind of problem are you running? Is it something that you purposefully made hard-to-balance or something more or less "uniform"? Do you notice a difference in runtime?

friedmud:
@jessecarterMOOSE perfect cube: 126x126x126

Unfortunately - I don't notice much runtime difference.  I was _really_ hoping for some (the reason I'm looking at all of this is to do some scalability optimization).

As long as we're looking at graphs of partitionings: check out this one of METIS vs Hilbert SFC:

hilbert_vs_metis

Hilbert SFC is able to produce a near _uniform_ partitioning... while still keeping the number of edge cuts similar to METIS (although - with a smaller stddev on the edge cuts).

Unfortunately - even that is not fixing the current scalability problem I'm seeing - but it's still cool to play with!

roystgnr:
> I can provide some good news about this PR: it definitely makes a huge difference in nodal load balance.

Thanks, I needed to hear that.  I didn't even open my Github emails this weekend because I couldn't imagine that the chat on this issue was anything other than "38 commits and a month of work?  Maybe just give up already!?!"

I can (after a few red herrings) reproduce the MOOSE failures now, although I don't yet understand them.

My current plan is still "understand the errors, fix the errors, then clean up the branch history", but if that doesn't start bearing fruit quickly then I think plan B is going to be "start cherry-picking chunks of related commits into separate branches, make sure my MOOSE build is happy with them, then submit them as separate PRs".  Mixed in with the "change node partitioning" commits and the "fix everything that was broken by changed node partitioning" commits are a few swaths of "fix an unrelated bug that was exposed by changed node partitioning" commits, a few plain "make things more robust" commits, and a few "add more functionality which was useful for debugging changed node partitioning" commits, and I bet I can at least get those in without breaking anybody, and they'll make it easier to bisect the rest.

roystgnr:
And now I think I get it.  On that very third commit, "Even contract() shouldn't renumber if disallowed", we create the possibility for coarsened meshes to have gaps between node numbers.

But many of our output methods use EquationSystems::build_parallel_solution_vector, which currently assumes (and tries to assert, but the assert is broken for ReplicatedMesh) that there are no gaps between node numbers.

I can make build_parallel_solution_vector more robust, and I'll try that first since it's obviously the right thing to do anyway, but we'll see whether or not that just opens up the door to more "we didn't expect gaps in node numbers" breakage later.

The only other solution is to resume ignoring users' node renumbering settings when doing adaptive coarsening, but that just seems evil.  Our ```MeshBase::allow_renumbering()``` returns ```bool```, meaning ```true``` or ```false``` but never ```false_just_kidding_true```.

roystgnr:
Fewer test failures, now.  And it looks like the biggest remaining problem here is that, when we're not renumbering nodes, we're also failing to delete disconnected nodes, because that was previously done in the same swoop.  That should be easy to fix.

roystgnr:
Fewer test failures, now.  And it looks like the biggest remaining problem here is that, when we're not renumbering nodes, we're also failing to delete disconnected nodes, because that was previously done in the same swoop.  That should be easy to fix.

roystgnr:
I don't think the utils/random.test_par_mesh regression is going to be fixable - any change in node partitioning or numbering is sure to end up putting PRNG results in different places.

permcody:
Yeah, that's fine. Random numbers on a distributed mesh are an enormous pain. However there is hope! Derek recently pointed me to this (used in PETSc): http://www.deshawresearch.com/resources_random123.html

We should be able to generate parallel consistent random numbers even on distributed mesh with that PRNG.

I'd be fine either regolding or dropping. If we want to keep testing, we can restrict the hell out of that test so it only runs on one platform, one version of PETSc, etc.

roystgnr:
vectorpostprocessors/work_balance.test is also ironically impossible to please here - if we *passed* that test it would be evidence of a bug!

Instead of 67 nodes on proc 0 and 54 on proc 1 we now have 61 on proc 0 and 60 on proc 1, and that's awesome, but obviously the CSV differ thinks it's suspicious.

roystgnr:
I wonder - would it be sane to add an option to restore the old libMesh behavior?  IIRC there are a lot of MOOSE apps tests which rely on PRNG initial conditions...

roystgnr:
I think my breaking point would be adding an option to undo the "skip_repartitioning should apply to nodes too" commit.  I can almost justify adding a "--use_node_partitioning==dumb_old_rule" option, under the theory that we might want to try a whole assortment of heuristics later and so it couldn't hurt to add the original heuristic as one of those, but I can't even figure out how to name a "--dont_skip_node_repartitioning_even_if_told" option such that it doesn't sound insane.

roystgnr:
postprocessors/all_print_pps.test also appears to be a sort of false positive: evaluating a residual 27 times instead of 26 is simply the sort of thing you'd expect to happen when you juggle around which DoFs are on which processor and you use a partitioning-dependent preconditioner.

friedmud:
@roystgnr I have some bad news.  Sit down for this one...

I did fairly extensive testing with your branch all weekend... testing it against the old method... your branch is almost always _slower_ than the original way.

After working on it all weekend I think I figured out why: it destroys data locality in the preconditioner.  With your branch the preconditioning is always worse - I have to turn up the number of levels of overlap or use less aggressive coarsening options in Hypre BoomerAMG to get the same solver history.

What is happening is that nearly every element now has at least one node _missing_ on the processor.  What this means is that any preconditioner that runs a domain-decomposed process now sees a less complete picture of the matrix.  These include Block Jacobi, ASM and Hypre BoomerAMG (especially using PMIS or HMIS aggressive coarsening - those are almost totally local operations).

Under the old numbering it was much more likely that "complete" element's nodes would all be on the same processor (on the lower one) - which gives the preconditioners a good view of what's going on.

I talked extensively with @jwpeterson and @permcody about this yesterday and today and we think the right thing to do is, for now, have libMesh default to the old numbering - but allow a switch to the new numbering so that we can compare them more thoroughly on a larger test problem.  There is no doubt that this new numbering gives a better balancing of nodal work - and a better distribution of DoFs in PETSc matrices / vectors... but it might come at too high of a cost of quality preconditioning.

I have a few crazy ideas about what to do about it - one of the wilder ones includes running another partitioner on the lower dimensional surfaces - and we have @fdkong who has money right now to look into this kind of thing.  @jwpeterson has expressed interest in using a "random" process (I think Barry Smith suggested this).

How possible do you think it would be to put in a generic interface for setting node ownership?  Maybe it's time for a new Functor system?

We are now thinking that we might try multiple different methods and publish the results.  I think it would be a nice contribution to the community.

I'm really sorry to bring this bad news to you - I really didn't want it to be true (especially with how much work you put into this!).  I do think this is still useful to have as an option (if you have REALLY heavy node-based calculations this could make a huge difference) - and we still need to do more studies to look at the impact on different problems.  For instance, this might make sense in an _explicit_ solve when you're not using any preconditioner.

friedmud:
BTW: I don't have the exact numbers on me right now - but I can generate some if you want to see the numbers before we make a decision.  I just wanted to let you know what I was seeing before this gets merged.

roystgnr:
> With your branch the preconditioning is always worse

That's a surprise, but not a huge one.  I've noticed that ILU subblock preconditioning on saddle-point problems can be ruined, and I suppose it's not hard to make other preconditioners deteriorate in more subtle ways.

> How possible do you think it would be to put in a generic interface for setting node ownership? Maybe it's time for a new Functor system?

This is probably more hubris, but: it sounds relatively easy.  All the hard work has been done.

The question is: how much data would it need to have access to?

Currently, Node::choose_processor_id() has access to the node it's called on and nothing more.

I suppose if we have a functor abstract base class, and if we make the library always init() that class before it starts asking for processor ids, that's always *technically* enough, isn't it?  You could always have a functor subclass which saves a MeshBase reference (or a MyApplicationStuff reference, etc), and you can do cacheable calculations (like building up a point locator, or even handing off to a third party partitioner) in the init, and after that you can query cached pids or element patches or whatever else you need.

In fact, we probably want the functor to attach to MeshBase and we can just pass that to its init() call by default.

But on the other hand, the point locator lookup is O(log(N)) (or log(N/Nproc) on distributed meshes), whereas just getting a patch once you already have an element in hand is O(1), so if we need to use element data when partitioning nodes, maybe it's better if the API forces us to always provide that first element?

It seems to me that for the most scalable results we may need to see element data when partitioning nodes.  We want to increase the likelihood that all nodes of an element land on the same processor, and our obvious options include "keep track of new pids of nearby nodes", "look at all elements in a node's patch at once rather than one-by-one", etc.

On the gripping hand, there is one place in MeshTools::correct_node_proc_ids() where we don't actually *have* an element handy to look at, because we're often looking at remote element processor ids rather than ids from semilocal elements.

Okay, I take back what I said about "easy".  Here's what *would* be easy, though: modify Node::choose_processor_id() to default to the old behavior, but to enable the new behavior (and scream libmesh_experimental()!) if some new command line flag is set.  Sound good?

That would allow me to get this PR cleaned up and merged before the fork grows much bigger, and then I can start work on the functor interface on a smaller easier-to-bisect new branch.

> I'm really sorry to bring this bad news to you

Never apologize for supplying new data!  Ironically, I suppose I am a little offended that you'd think I might blame the messenger, but I'll get over that.  ;-)

roystgnr:
That stateful_prop failed assert I can fix.  Can anyone figure out what went wrong in the rattlesnake run?

friedmud:
All good to hear:

> Never apologize for supplying new data!

Hah - I wasn't worried you would be angry with me... but I still felt bad to lay down some bad news when you've done so much good work!

I'm not at all sure what the interface should be.  I think if I were building one of these I would probably build a node->elem list and use that to impact things.

Just to give you another example of something @jwpeterson and I were discussing yesterday is the idea of just using a "linear partitioner" on the boundary nodes.  Just order them - and assign half to each processor.  It does get tricky when things are touching many processors though.  Just trying to put some use-cases out there so you can think about the interface.

I definitely like the idea of getting this in there without doing more work on it though (with the default being the old numbering) and then we can do more testing and see what we come up with as far as a new interface later.

roystgnr:
> I think if I were building one of these I would probably build a node->elem list and use that to impact things.

That would be another point for "just provide the node to the API and we can find anything else during init() if we need to", I think...

Except that you've now got me scared about how badly users might get themselves into trouble with a custom functor if we force them to do too much themselves.  Note, for example, that in many distributed meshes there will exist at least one node for which *no processor will have a complete list of elements touching it*!  Trying to do anything based on complete element patches will require users to use the trickiest parallel_ghost_sync.h method just to *find* those patches!  The assertions I've currently put into place for the Node::choose_processor_id() API catch inconsistencies, but if some other API makes more sense then we'd better have the same level of backstopping of it.

permcody:
| Roy | Buzzword | Bingo |
| --------- | --------- | --------- |
|  active semilocal | IIRC | other hand |
| ~"easy"~ | ~gripping hand~ | ~functor~ |
| interior parent | DIE A HORRIBLE DEATH HERE | ~libmesh_assert()~ |

**BINGO!**

pbauman:
This made laugh so hard. I wish I wasnt sick so it didn’t hurt to laugh.

roystgnr:
I think it's a great idea.

Possible within the current infrastructure, no, but would be possible within the functor API I'm currently mulling over.

It would be a giant pain in the neck, but much of the giant pain in the neck is just writing the new partitioner shim, which is something we want to do *anyway* to give us a better alternative to METIS/ParMETIS, and the extra marginal effort after that wouldn't be too bad.

friedmud:
Cool.  It should be easy to do as a small modification to the current
METIS/ParMETIS partitioners.  Just pass in the node connectivity graph
instead of the element connectivity graph...

roystgnr:
The only remaining work from this PR is the ReplicatedMesh "no renumbering means no renumbering" fix in #1666

Issue #1622, opened 2018-03-08
Still bugs in metaphysicl configure


jwpeterson:
@roystgnr Definitely after #1615 (maybe before, not sure) I am getting errors during metaphysicl configure

```
checking for OpenCL library... no
../../../../contrib/metaphysicl/0.2.0/configure: line 22307: test: -ge: unary operator expected
configure: Boost not available, not searching for the Boost chrono library
configure: Boost not available, not searching for the Boost date_time library
../../../../contrib/metaphysicl/0.2.0/configure: line 23190: test: -ge: unary operator expected
configure: Boost not available, not searching for the Boost filesystem library
configure: Boost not available, not searching for the Boost system library
../../../../contrib/metaphysicl/0.2.0/configure: line 24074: test: -ge: unary operator expected
../../../../contrib/metaphysicl/0.2.0/configure: line 24372: test: -lt: unary operator expected
```

The lines of code in question all seem to involve `$boost_major_version` being used unguarded, e.g.
```
if test $boost_major_version -ge 135; then
  if test x"$boost_cv_inc_path" = xno; then
  { $as_echo "$as_me:${as_lineno-$LINENO}: Boost not available, not searching for the Boost system library" >&5
$as_echo "$as_me: Boost not available, not searching for the Boost system library" >&6;}
```
so the problem is presumably in contrib/metaphysicl/0.2.0/m4/common/boost.m4.

jwpeterson:
This is also in v1.3.0, so I'm going to take a crack at fixing it myself...

PR #1623, opened 2018-03-08
Metaphysicl:  Check for unset variable before comparing with -ge, -lt, etc.

jwpeterson:
Refs #1622.


roystgnr:
I'd say we should put this in an upstream PR, too, but upstream seems to date back to Oct 2016, despite having a developer who's still active on other things and a repo with a bunch of simple and useful PRs outstanding.  :-P

PR #1624, opened 2018-03-11
Convert some m4 code to use AS_IF, AS_CASE macros

jwpeterson:
These macros should be preferred to hard-coded `if` and `case` statements for compatibility with the widest possible range of POSIX shells. The macros are also a bit more terse (no closing `fi` or `esac` needed), easier to remember how to use than raw shell syntax, easier to Google when you can't remember exactly how to use them, and Emacs is much better at matching the opening/closing parentheses used by the macros than it is at matching `if`/`fi` in m4 mode. 

The current history has a lot of commits because I wanted to be able to verify that an equivalent configure script was still being generated after each change, but I will squash that down before merging if everything passes.

### DIFF
--- a/configure
+++ b/configure
@@ -7272,35 +7272,36 @@ fi
   # First the flags for gcc compilers
   if test "$GXX" = yes -a "x$REAL_GXX" != "x"; then :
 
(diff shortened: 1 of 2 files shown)


jwpeterson:
We will be merging #1625 instead.

PR #1625, opened 2018-03-11
(Rebased) Convert some m4 code to use AS_IF, AS_CASE macros.

jwpeterson:
Same as #1624 with a simpler history

### DIFF
--- a/m4/compiler.m4
+++ b/m4/compiler.m4
@@ -13,18 +13,15 @@ AC_DEFUN([LIBMESH_SET_COMPILERS],
      AC_ARG_ENABLE(mpi,
                    AS_HELP_STRING([--disable-mpi],
                                   [build without MPI message passing support]),
-                   [case "${enableval}" in
-                     yes)  enablempi=yes ;;
-                     no)  enablempi=no ;;
-                     *)  AC_MSG_ERROR(bad value ${enableval} for --enable-mpi) ;;
-                   esac],
+                   [AS_CASE("${enableval}",
+                            [yes], [enablempi=yes],
+                            [no],  [enablempi=no],
+                            [AC_MSG_ERROR(bad value ${enableval} for --enable-mpi)])],
                    [enablempi=yes])
 
-  if  (test "$enablempi" != no) ; then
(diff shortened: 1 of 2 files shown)


brianmoose:
The moose test harness seems to have failed to completely kill some parallel jobs on moose PRs.
This left moose binaries running on some of the machines over the weekend (ie there was a constant ~12 load on some of the machines)
I imagine that is what was causing the timeouts...

roystgnr:
That's a relief to hear, actually!  I was starting to wonder if there was some recent performance regression in debug mode that was only now adding to enough noise to trigger frequent failures, or if there actually might be some sort of LIBMESH_ENABLE_FASTER_CODE which a broken m4 macro here was failing to define.

m4/compiler.m4
```diff
@@ -13,18 +13,15 @@ AC_DEFUN([LIBMESH_SET_COMPILERS],
...
                    [enablempi=yes])
 
-  if  (test "$enablempi" != no) ; then
-    CXX_TRY_LIST="mpicxx mpiCC mpicc $CXX_TRY_LIST"
-  else
-    AC_MSG_RESULT(>>> Disabling MPI per user request <<<)
-  fi
+  AS_IF([test "$enablempi" != no],
+        [CXX_TRY_LIST="mpicxx mpiCC mpicc $CXX_TRY_LIST"],
+        AC_MSG_RESULT([>>> Disabling MPI per user request <<<]))
```
roystgnr: Any particular reason we're wrapping this AC_MSG_RESULT argument and not the similar others below?

jwpeterson: Oh yeah... I was playing around with different amounts of quoting (inside the argument to `AC_MSG_RESULT`, around the entire `AC_MSG_RESULT` call, etc.) and in the end I don't think it made any difference. We could change this one to be consistent with all the others if you prefer...

roystgnr: Nah, it's fine; I just wanted to make sure there wasn't some subtle m4 gotcha I was missing.

PR #1630, opened 2018-03-13
Fixed bug in discontinuous plotting with Exodus format.

### LINKED ISSUE #1628: write_element_data after write_timestep_discontinuous gives incorrect results
If I do:
```
  ExodusII_IO exo_out(mesh);
  exo_out.write_equation_systems("name.exo", es, &empty_list);
  exo_out.write_element_data(es);
```
then I get the correct element data plot. But if I do:
```
  ExodusII_IO exo_out(mesh);
  exo_out.write_discontinuous_exodusII("name.exo", es, &empty_list);
  exo_out.write_element_data(es);
```
then in that case the element data plot is all mixed up.

Personally, I'd like the latter case to work since I want to include a discontinuous solution plot in the same exo file as the element data.

dknez:
This fixes #1628 for me. We need to use the same node numbering in ExodusII_IO_Helper::write_elements as we use in ExodusII_IO_Helper::write_nodal_coordinates.

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -1399,8 +1399,24 @@ void ExodusII_IO_Helper::write_elements(const MeshBase & mesh, bool use_disconti
   // This counter is used to fill up the libmesh_elem_num_to_exodus map in the loop below.
   unsigned libmesh_elem_num_to_exodus_counter = 0;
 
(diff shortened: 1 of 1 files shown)


dknez:
Thanks for checking this @jwpeterson, I'll go ahead and merge now.

PR #1631, opened 2018-03-15
Bootstrap metaphsyicl (#1627), misc. fixes for non-standard builds

jwpeterson:
The metaphysicl autoconf files were updated in 0767791 but the generated files were not updated.

### DIFF
--- a/examples/adjoints/adjoints_ex1/adjoints_ex1.C
+++ b/examples/adjoints/adjoints_ex1/adjoints_ex1.C
@@ -285,6 +285,11 @@ int main (int argc, char ** argv)
   libmesh_example_requires(libMesh::default_solver_package() != INVALID_SOLVER_PACKAGE,
                            "--enable-petsc, --enable-trilinos, or --enable-eigen");
 
+  // This example relies on exceptions to control which Partitioner gets built.
+#ifndef LIBMESH_ENABLE_EXAMPLES
+  libmesh_example_requires(false, "--enable-exceptions");
+#endif
+
   // Skip adaptive examples on a non-adaptive libMesh build
 #ifndef LIBMESH_ENABLE_AMR
   libmesh_example_requires(false, "--enable-amr");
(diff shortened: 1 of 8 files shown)


jwpeterson:
Note: This is also a test of the new civet.inl.gov host, which appears to be working OK... Thanks @brianmoose!

I've also attached a swath of `--disable-foo` tests to see if we've picked up any recent regressions in our non-standard builds and will investigate those...

examples/adjoints/adjoints_ex1/adjoints_ex1.C
```diff
@@ -285,6 +285,11 @@ int main (int argc, char ** argv)
   libmesh_example_requires(libMesh::default_solver_package() != INVALID_SOLVER_PACKAGE,
                            "--enable-petsc, --enable-trilinos, or --enable-eigen");
 
+  // This example relies on exceptions to control which Partitioner gets built.
+#ifndef LIBMESH_ENABLE_EXAMPLES
+  libmesh_example_requires(false, "--enable-exceptions");
+#endif
+
```
roystgnr: This looks good for now, but in the long run we ought to add a Factory method which tests for a name in the map, then switch adjoints_ex1 to use that rather than attempting a build and catching the error exception.

jwpeterson: What am I even doing, this should be `#ifndef LIBMESH_ENABLE_EXCEPTIONS`! Will fix right now.

roystgnr: We both stared *right* at it...

jwpeterson: Must be that thing where your brain can still read words even when several of the letters are missing.

roystgnr: I'm always tempted to be annoyed with my youngest when she just reads the first letter or two of a word and then guesses wrong at the rest, but damn, at least *she* has an *excuse*...

PR #1632, opened 2018-03-15
LIBMESH_BROKEN_IOSTREAM no longer exists.

jwpeterson:
This #define only applied to really old versions of GCC (2.95 and 2.96) which we no longer support.


roystgnr:
Good catch!

Although, I must say, I was a bit startled when I first glanced at my phone email notification last night:

"John W. Peterson - LIBMESH_BROKEN..."

PR #1634, opened 2018-03-17
Enable periodic BCs with transformation between variables

dknez:
Added a DenseMatrix to PeriodicBoundaryBase which defines a transformation between the dofs on the two sides of the periodic boundary. This is relevant for periodic boundaries on a sector of a circular domain with vector-valued quantities such as velocity or displacement, for example.

I think this will need more work, but I wanted to share what I've done so far to get feedback.

### DIFF
--- a/src/base/periodic_boundary_base.C
+++ b/src/base/periodic_boundary_base.C
@@ -39,6 +39,12 @@ PeriodicBoundaryBase::PeriodicBoundaryBase(const PeriodicBoundaryBase & o) :
   pairedboundary(o.pairedboundary),
   variables(o.variables)
 {
+  // Make a deep copy of _transformation_matrix, if it's not null
+  if(o._transformation_matrix)
+  {
+    this->_transformation_matrix = libmesh_make_unique<DenseMatrix<Real>>();
+    *(this->_transformation_matrix) = *(o._transformation_matrix);
+  }
 }
 
 
@@ -63,6 +69,44 @@ bool PeriodicBoundaryBase::is_my_variable(unsigned int var_num) const
   return a;
 }
 
+
+
+bool PeriodicBoundaryBase::has_transformation_matrix() const
+{
+  return bool(_transformation_matrix);
+}
+
+
+
+const DenseMatrix<Real> & PeriodicBoundaryBase::get_transformation_matrix() const
+{
+  if(!has_transformation_matrix())
+  {
+    libmesh_error_msg("Transformation matrix is not defined");
+  }
+
+  return *_transformation_matrix;
+}
+
+
+
+void PeriodicBoundaryBase::set_transformation_matrix(const DenseMatrix<Real> & matrix)
+{
+  // Make a deep copy of matrix
+  this->_transformation_matrix = libmesh_make_unique<DenseMatrix<Real>>();
+  *(this->_transformation_matrix) = matrix;
+
+  // if _transformation_matrix is defined then it must be the same sie as variables.
+  libmesh_assert_equal_to(_transformation_matrix->m(), variables.size());
+  libmesh_assert_equal_to(_transformation_matrix->n(), variables.size());
+}
+
+
+
+const std::set<unsigned int> & PeriodicBoundaryBase::get_variables() const
+{
+  return variables;
+}
 
 } // namespace libMesh
 
--- a/src/fe/fe_base.C
+++ b/src/fe/fe_base.C
@@ -1789,6 +1789,23 @@ compute_periodic_constraints (DofConstraints & constraints,
                   dof_map.dof_indices (neigh, neigh_dof_indices,
                                        variable_number);
 
+                  // We use neigh_dof_indices_all_variables in the case that the
+                  // periodic boundary condition involves mappings between multiple
+                  // variables.
+                  std::vector<std::vector<dof_id_type>> neigh_dof_indices_all_variables;
+                  if(periodic->has_transformation_matrix())
+                    {
+                      const std::set<unsigned int> & variables = periodic->get_variables();
+                      neigh_dof_indices_all_variables.resize(variables.size());
+                      unsigned int index = 0;
+                      for(unsigned int var : variables)
+                        {
+                          dof_map.dof_indices (neigh, neigh_dof_indices_all_variables[index],
+                                               var);
+                          index++;
+                        }
+                    }
+
                   const unsigned int n_qp = my_qface.n_points();
 
                   // Translate the quadrature points over to the
@@ -2337,8 +2354,33 @@ compute_periodic_constraints (DofConstraints & constraints,
                           if (std::abs(their_dof_value) < 10*TOLERANCE)
                             continue;
 
-                          constraint_row->insert(std::make_pair(their_dof_g,
-                                                                their_dof_value));
+                          if(!periodic->has_transformation_matrix())
+                            {
+                              constraint_row->insert(std::make_pair(their_dof_g,
+                                                                    their_dof_value));
+                            }
+                          else
+                            {
+                              // In this case the current variable is constrained in terms of other variables.
+                              // We assume that all variables in this constraint have the same FE type (this
+                              // is asserted below), and hence we can create the constraint row contribution
+                              // by multiplying their_dof_value by the corresponding row of the transformation
+                              // matrix.
+
+                              const std::set<unsigned int> & variables = periodic->get_variables();
+                              neigh_dof_indices_all_variables.resize(variables.size());
+                              unsigned int index = 0;
+                              for(unsigned int other_var : variables)
+                                {
+                                  libmesh_assert_msg(base_fe_type == dof_map.variable_type(other_var), "FE types must match for all variables involved in constraint");
+
+                                  Real var_weighting = periodic->get_transformation_matrix()(variable_number, other_var);
+                                  constraint_row->insert(std::make_pair(neigh_dof_indices_all_variables[index][i],
+                                                                        var_weighting*their_dof_value));
+                                  index++;
+                                }
+                            }
+
                         }
                     }
                 }
--- a/include/base/periodic_boundary.h
+++ b/include/base/periodic_boundary.h
@@ -79,8 +79,6 @@ public:
   virtual std::unique_ptr<PeriodicBoundaryBase> clone(TransformationType t = FORWARD) const libmesh_override;
 
 protected:
-  // One of these days we'll support rotated boundaries
-  // RealTensor rotation_matrix;
 
   // The vector which is added to points in myboundary
   // to produce corresponding points in pairedboundary
--- a/include/base/periodic_boundary_base.h
+++ b/include/base/periodic_boundary_base.h
@@ -26,6 +26,7 @@
 // Local Includes
 #include "libmesh/point.h"
 #include "libmesh/auto_ptr.h" // deprecated
+#include "libmesh/dense_matrix.h"
 
 // C++ Includes
 #include <set>
@@ -101,12 +102,49 @@ public:
 
   bool is_my_variable(unsigned int var_num) const;
 
+  /**
+   * @return true if _transformation_matrix is not null.
+   */
+  bool has_transformation_matrix() const;
+
+  /**
+   * Get the transformation matrix, if it is defined.
+   * Throw an error if it is not defined.
+   */
+  const DenseMatrix<Real> & get_transformation_matrix() const;
+
+  /**
+   * Set the transformation matrix. When calling this method we
+   * require the following conditions:
+   *  1) \p matrix is square with size that matches this->variables.size()
+   *  2) the list of variables in this->variables set must all have the same FE type
+   * Both of these conditions are asserted in DBG mode.
+   */
+  void set_transformation_matrix(const DenseMatrix<Real> & matrix);
+
+  /**
+   * Get the set of variables for this periodic boundary condition.
+   */
+  const std::set<unsigned int> & get_variables() const;
+
 protected:
 
   /**
    * Set of variables for this periodic boundary, empty means all variables possible
    */
   std::set<unsigned int> variables;
+
+  /**
+   * A DenseMatrix that defines the mapping of variables on this
+   * boundary and the counterpart boundary. This is necessary for
+   * periodic-boundaries with vector-valued quantities (e.g.
+   * velocity or displacement) on a sector of a circular domain,
+   * for exaple, since in that case we must map each variable
+   * to a corresponding linear combination of all the variables.
+   * We store the DenseMatrix via a unique_ptr, and an uninitialized
+   * pointer is treated as equivalent to the identity matrix.
+   */
+  std::unique_ptr<DenseMatrix<Real>> _transformation_matrix;
 };
 
 } // namespace libmesh


dknez:
FYI, I tested this on the test case below, which is a modified version of systems_of_equations_ex6. The solution is hard to interpret due to the weird (and arbitrary) mapped periodic BC that I imposed, but the solution doesn't look unreasonable to me. (Note that I used a direct solver, e.g. MUMPS, during testing to avoid any potential convergence issues.)

```
// C++ include files that we need
#include <iostream>
#include <algorithm>
#include <math.h>

// libMesh includes
#include "libmesh/libmesh_config.h"
#include "libmesh/libmesh.h"
#include "libmesh/mesh.h"
#include "libmesh/mesh_generation.h"
#include "libmesh/exodusII_io.h"
#include "libmesh/gnuplot_io.h"
#include "libmesh/linear_implicit_system.h"
#include "libmesh/equation_systems.h"
#include "libmesh/fe.h"
#include "libmesh/quadrature_gauss.h"
#include "libmesh/dof_map.h"
#include "libmesh/sparse_matrix.h"
#include "libmesh/numeric_vector.h"
#include "libmesh/dense_matrix.h"
#include "libmesh/dense_submatrix.h"
#include "libmesh/dense_vector.h"
#include "libmesh/dense_subvector.h"
#include "libmesh/perf_log.h"
#include "libmesh/elem.h"
#include "libmesh/boundary_info.h"
#include "libmesh/zero_function.h"
#include "libmesh/dirichlet_boundaries.h"
#include "libmesh/string_to_enum.h"
#include "libmesh/getpot.h"
#include "libmesh/solver_configuration.h"
#include "libmesh/petsc_linear_solver.h"
#include "libmesh/petsc_macro.h"
#include "libmesh/periodic_boundaries.h"
#include "libmesh/periodic_boundary.h"

// boundary IDs
#define BOUNDARY_ID_MIN_Z 0
#define BOUNDARY_ID_MIN_Y 1
#define BOUNDARY_ID_MAX_X 2
#define BOUNDARY_ID_MAX_Y 3
#define BOUNDARY_ID_MIN_X 4
#define BOUNDARY_ID_MAX_Z 5
#define CLAMPED_ID 10

// Bring in everything from the libMesh namespace
using namespace libMesh;

class LinearElasticity : public System::Assembly
{
private:
  EquationSystems & es;

public:

  LinearElasticity (EquationSystems & es_in) :
    es(es_in)
  {}

  /**
   * Kronecker delta function.
   */
  Real kronecker_delta(unsigned int i,
                       unsigned int j)
  {
    return i == j ? 1. : 0.;
  }

  /**
   * Evaluate the fourth order tensor (C_ijkl) that relates stress to strain.
   */
  Real elasticity_tensor(unsigned int i,
                         unsigned int j,
                         unsigned int k,
                         unsigned int l)
  {
    // Hard code material parameters for the sake of simplicity
    const Real poisson_ratio = 0.3;
    const Real young_modulus = 1.;

    // Define the Lame constants
    const Real lambda_1 = (young_modulus*poisson_ratio)/((1.+poisson_ratio)*(1.-2.*poisson_ratio));
    const Real lambda_2 = young_modulus/(2.*(1.+poisson_ratio));

    return lambda_1 * kronecker_delta(i, j) * kronecker_delta(k, l) +
      lambda_2 * (kronecker_delta(i, k) * kronecker_delta(j, l) + kronecker_delta(i, l) * kronecker_delta(j, k));
  }

  /**
   * Assemble the system matrix and right-hand side vector.
   */
  void assemble()
  {
    const MeshBase & mesh = es.get_mesh();

    const unsigned int dim = mesh.mesh_dimension();

    LinearImplicitSystem & system = es.get_system<LinearImplicitSystem>("Elasticity");

    const unsigned int u_var = system.variable_number ("u");

    const DofMap & dof_map = system.get_dof_map();
    FEType fe_type = dof_map.variable_type(u_var);
    std::unique_ptr<FEBase> fe (FEBase::build(dim, fe_type));
    QGauss qrule (dim, fe_type.default_quadrature_order());
    fe->attach_quadrature_rule (&qrule);

    std::unique_ptr<FEBase> fe_face (FEBase::build(dim, fe_type));
    QGauss qface(dim-1, fe_type.default_quadrature_order());
    fe_face->attach_quadrature_rule (&qface);

    const std::vector<Real> & JxW = fe->get_JxW();
    const std::vector<std::vector<Real>> & phi = fe->get_phi();
    const std::vector<std::vector<RealGradient>> & dphi = fe->get_dphi();

    DenseMatrix<Number> Ke;
    DenseSubMatrix<Number> Ke_var[3][3] =
      {
        {DenseSubMatrix<Number>(Ke), DenseSubMatrix<Number>(Ke), DenseSubMatrix<Number>(Ke)},
        {DenseSubMatrix<Number>(Ke), DenseSubMatrix<Number>(Ke), DenseSubMatrix<Number>(Ke)},
        {DenseSubMatrix<Number>(Ke), DenseSubMatrix<Number>(Ke), DenseSubMatrix<Number>(Ke)}
      };

    DenseVector<Number> Fe;

    DenseSubVector<Number> Fe_var[3] =
      {DenseSubVector<Number>(Fe),
       DenseSubVector<Number>(Fe),
       DenseSubVector<Number>(Fe)};

    std::vector<dof_id_type> dof_indices;
    std::vector<std::vector<dof_id_type>> dof_indices_var(3);

    for (const auto & elem : mesh.active_local_element_ptr_range())
      {
        dof_map.dof_indices (elem, dof_indices);
        for (unsigned int var=0; var<3; var++)
          dof_map.dof_indices (elem, dof_indices_var[var], var);

        const unsigned int n_dofs   = dof_indices.size();
        const unsigned int n_var_dofs = dof_indices_var[0].size();

        fe->reinit (elem);

        Ke.resize (n_dofs, n_dofs);
        for (unsigned int var_i=0; var_i<3; var_i++)
          for (unsigned int var_j=0; var_j<3; var_j++)
            Ke_var[var_i][var_j].reposition (var_i*n_var_dofs, var_j*n_var_dofs, n_var_dofs, n_var_dofs);

        Fe.resize (n_dofs);
        for (unsigned int var=0; var<3; var++)
          Fe_var[var].reposition (var*n_var_dofs, n_var_dofs);

        for (unsigned int qp=0; qp<qrule.n_points(); qp++)
          {
            // assemble \int_Omega C_ijkl u_k,l v_i,j \dx
            for (unsigned int dof_i=0; dof_i<n_var_dofs; dof_i++)
              for (unsigned int dof_j=0; dof_j<n_var_dofs; dof_j++)
                for (unsigned int i=0; i<3; i++)
                  for (unsigned int j=0; j<3; j++)
                    for (unsigned int k=0; k<3; k++)
                      for (unsigned int l=0; l<3; l++)
                        Ke_var[i][k](dof_i,dof_j) +=
                          JxW[qp] * elasticity_tensor(i,j,k,l) * dphi[dof_j][qp](l) * dphi[dof_i][qp](j);

            // assemble \int_Omega f_i v_i \dx
            DenseVector<Number> f_vec(3);
            f_vec(0) =  0.;
            f_vec(1) =  0.;
            f_vec(2) = -1.;
            for (unsigned int dof_i=0; dof_i<n_var_dofs; dof_i++)
              for (unsigned int i=0; i<3; i++)
                Fe_var[i](dof_i) += JxW[qp] * (f_vec(i) * phi[dof_i][qp]);
          }

        dof_map.constrain_element_matrix_and_vector (Ke, Fe, dof_indices);

        system.matrix->add_matrix (Ke, dof_indices);
        system.rhs->add_vector    (Fe, dof_indices);
      }
  }
};


// Begin the main program.
int main (int argc, char ** argv)
{
  // Initialize libMesh and any dependent libraries
  LibMeshInit init (argc, argv);

  // This example requires a linear solver package.
  libmesh_example_requires(libMesh::default_solver_package() != INVALID_SOLVER_PACKAGE,
                           "--enable-petsc, --enable-trilinos, or --enable-eigen");

  // Initialize the cantilever mesh
  const unsigned int dim = 3;

  // Make sure libMesh was compiled for 3D
  libmesh_example_requires(dim == LIBMESH_DIM, "3D support");

  // Create a 3D mesh distributed across the default MPI communicator.
  Mesh mesh(init.comm(), dim);
  MeshTools::Generation::build_cube (mesh,
                                     32,
                                     8,
                                     4,
                                     0., 1,
                                     0., 0.3,
                                     0., 0.1,
                                     HEX8);

  // Print information about the mesh to the screen.
  mesh.print_info();

  // Add CLAMPED_ID to the boundary_info so that we can clamp the model to
  // make sure it is well-posed.
  for (const auto & elem : mesh.element_ptr_range())
    {
        {

          if (mesh.get_boundary_info().has_boundary_id(elem, side, BOUNDARY_ID_MIN_Z))
            {
              Real centroid_x = elem->centroid()(0);
              if( (centroid_x>0.4) && (centroid_x<0.5))
                {
                  mesh.get_boundary_info().add_side(elem, side, CLAMPED_ID);
                }
            }
        }
    }

  // Create an equation systems object.
  EquationSystems equation_systems (mesh);

  // Declare the system and its variables.
  // Create a system named "Elasticity"
  LinearImplicitSystem & system =
    equation_systems.add_system<LinearImplicitSystem> ("Elasticity");

  // Add three displacement variables, u and v, to the system
  unsigned int u_var = system.add_variable("u", FIRST, LAGRANGE);
  unsigned int v_var = system.add_variable("v", FIRST, LAGRANGE);
  unsigned int w_var = system.add_variable("w", FIRST, LAGRANGE);

  LinearElasticity le(equation_systems);
  system.attach_assemble_object(le);

  PeriodicBoundary periodic_bc(RealVectorValue(1.0, 0., 0.));
  periodic_bc.myboundary = BOUNDARY_ID_MIN_X;
  periodic_bc.pairedboundary = BOUNDARY_ID_MAX_X;
  DenseMatrix<Number> transformation_matrix(3,3);
  transformation_matrix(0,0) = 0.5;
  transformation_matrix(0,1) = 0.5;
  transformation_matrix(1,0) = 0.5;
  transformation_matrix(1,1) = 0.5;
  transformation_matrix(2,2) = 1.;
  periodic_bc.set_transformation_matrix(transformation_matrix);
  periodic_bc.set_variable(u_var);
  periodic_bc.set_variable(v_var);
  periodic_bc.set_variable(w_var);
  system.get_dof_map().add_periodic_boundary(periodic_bc);

  std::vector<unsigned int> variables;
  variables.push_back(u_var);
  variables.push_back(v_var);
  variables.push_back(w_var);
  ZeroFunction<> zf;
  std::set<boundary_id_type> clamped_boundary_ids;
  clamped_boundary_ids.insert(CLAMPED_ID);
  DirichletBoundary clamped_bc(clamped_boundary_ids, variables, zf);
  system.get_dof_map().add_dirichlet_boundary(clamped_bc);

  // Initialize the data structures for the equation system.
  equation_systems.init();

  // Print information about the system to the screen.
  equation_systems.print_info();

  // Solve the system
  system.solve();

  // Plot the solution

  // Use single precision in this case (reduces the size of the exodus file)
  ExodusII_IO exo_io(mesh, /*single_precision=*/true);
  exo_io.write_discontinuous_exodusII("solution.exo", equation_systems);

#endif // #ifdef LIBMESH_HAVE_EXODUS_API

  // All done.
  return 0;
}
```

dknez:
I've also made a more realistic test case: elasticity on a sector of a circular domain. I made AzimuthalPeriodicBoundary (see below) to impose the periodic boundary conditions, and it seems to work (i.e. it gives a reasonable looking solution, but would need to do more validation to check if the answer is actually correct).

I think AzimuthalPeriodicBoundary doesn't really belong inside libMesh (at least in the form below) since it assumes that we have exactly 3 variables in the system so it isn't very generic. I'd be happy to add a new example that incorporates AzimuthalPeriodicBoundary, though.

```
class AzimuthalPeriodicBoundary : public PeriodicBoundaryBase
{
public:
  /**
   * Constructor
   */
  AzimuthalPeriodicBoundary(
    Point center,
    Point axis,
    Real angle)
  :
    PeriodicBoundaryBase(),
    _center(center),
    _axis(axis),
    _theta(angle)
  {
    set_variable(0);
    set_variable(1);
    set_variable(2);
    set_transformation_matrix(get_rotation_matrix());
  }

  /**
   * Copy constructor, with option for the copy to represent an inverse transformation.
   */
  AzimuthalPeriodicBoundary(const AzimuthalPeriodicBoundary & o, TransformationType t = FORWARD)
    :
    PeriodicBoundaryBase(o),
    _center(o._center),
    _axis(o._axis),
    _theta(o._theta)
  {
    if (t == INVERSE)
      {
        std::swap(myboundary, pairedboundary);
        _theta *= -1.0;
      }

    set_variable(0);
    set_variable(1);
    set_variable(2);
    set_transformation_matrix(get_rotation_matrix());
  }

  /**
   * Destructor
   */
  virtual ~AzimuthalPeriodicBoundary() {}

  /**
   * Get the rotation matrix for this transformation.
   */
  DenseMatrix<Real> get_rotation_matrix() const
  {
    // Formula for rotation matrix about an axis is given on wikipedia:
    // en.wikipedia.org/wiki/Rotation_matrix
    // We rotate by angle theta about the axis defined by u, which is a
    // unit vector in the direction of _axis.
    Point u = _axis.unit();
    Real u_x = u(0);
    Real u_y = u(1);
    Real u_z = u(2);
    DenseMatrix<Real> R(3,3);
    R(0,0) = cos(_theta) + u_x*u_x*(1.0 - cos(_theta));
    R(0,1) = u_x*u_y*(1.0 - cos(_theta)) - u_z*sin(_theta);
    R(0,2) = u_x*u_z*(1.0 - cos(_theta)) + u_y*sin(_theta);
    R(1,0) = u_y*u_x*(1.0 - cos(_theta)) + u_z*sin(_theta);
    R(1,1) = cos(_theta) + u_y*u_y*(1.0 - cos(_theta));
    R(1,2) = u_y*u_z*(1.0 - cos(_theta)) - u_x*sin(_theta);
    R(2,0) = u_z*u_x*(1.0 - cos(_theta)) - u_y*sin(_theta);
    R(2,1) = u_z*u_y*(1.0 - cos(_theta)) + u_x*sin(_theta);
    R(2,2) = cos(_theta) + u_z*u_z*(1.0 - cos(_theta));

    return R;
  }

  /**
   * This function should be overridden by derived classes to
   * define how one finds corresponding nodes on the periodic
   * boundary pair.
   */
  virtual Point get_corresponding_pos(const Point & pt) const libmesh_override
  {
    DenseVector<Real> translated_pt(3);
    for(unsigned int i=0; i<3; i++)
    {
      translated_pt(i) = pt(i) - _center(i);
    }

    DenseVector<Real> rotated_pt;
    get_transformation_matrix().vector_mult(rotated_pt, translated_pt);

    Point corresponding_pos;
    for(unsigned int i=0; i<3; i++)
    {
      corresponding_pos(i) = rotated_pt(i) + _center(i);
    }
    return corresponding_pos;
  }

  /**
   * If we want the DofMap to be able to make copies of references and
   * store them in the underlying map, this class must be clone'able,
   * i.e. have a kind of virtual construction mechanism.
   */
  virtual std::unique_ptr<PeriodicBoundaryBase> clone(TransformationType t = FORWARD) const libmesh_override
  {
    return libmesh_make_unique<AzimuthalPeriodicBoundary>(*this, t);
  }

private:

  // Define the properties needed for the azimuthal periodic boundary
  Point _center;
  Point _axis;
  Real _theta;
};
```

roystgnr:
> I think this will need more work

I'll mark it do-not-merge until you're happy; feel free to remove when ready.

dknez:
Actually, now that I've tested it, I think it works fine. Do you have any comments or suggestions for change? Otherwise I think it is OK to merge. The only other thing to change might be to remove the "// One of these days we'll support rotated boundaries" comment at the bottom of periodic_boundary.h.

roystgnr:
You had the right instincts with the Number-vs-Real stuff, because *maybe* there'll be someone who wants to do a complex transformation in a constraint equation one of these days... but until that happens, all our internal constraint calculations always give Real coefficients and so I'm loathe to make calculations 2x-4x more expensive for complex AMR/BC users before one of them actually needs the versatility.

We should probably squash that commit.

dknez:
> You had the right instincts with the Number-vs-Real stuff

Actually, I implemented it with Number initially, but it didn't compile, so I changed it. I'll squash before merging, but I wanted to get feedback before I squash.

src/base/periodic_boundary_base.C
```diff
@@ -39,6 +39,13 @@ PeriodicBoundaryBase::PeriodicBoundaryBase(const PeriodicBoundaryBase & o) :
   pairedboundary(o.pairedboundary),
   variables(o.variables)
 {
+  // Make a deep copy of _transformation_matrix, if it's not null
+  if(o._transformation_matrix)
+  {
+    DenseMatrix<Number> * new_matrix = new DenseMatrix<Number>;
```
roystgnr: Could we use libmesh_make_unique here?

src/base/periodic_boundary_base.C
```diff
@@ -64,6 +71,45 @@ bool PeriodicBoundaryBase::is_my_variable(unsigned int var_num) const
 }
 
 
+
+bool PeriodicBoundaryBase::has_transformation_matrix() const
+{
+  return (_transformation_matrix.get() != nullptr);
```
roystgnr: Is this three times as wordy as necessary?  unique_ptr has a conversion to bool, and even if it didn't then the raw pointer would have a conversion to bool;

src/base/periodic_boundary_base.C
```diff
@@ -64,6 +71,45 @@ bool PeriodicBoundaryBase::is_my_variable(unsigned int var_num) const
...
+}
+
+
+
+const DenseMatrix<Number> & PeriodicBoundaryBase::get_transformation_matrix() const
+{
+  if(!has_transformation_matrix())
```
roystgnr: Let's make this a libmesh_assert_msg, skip the test in opt mode?

src/base/periodic_boundary_base.C
```diff
@@ -64,6 +71,45 @@ bool PeriodicBoundaryBase::is_my_variable(unsigned int var_num) const
...
+
+  return *_transformation_matrix;
+}
+
+
+
+void PeriodicBoundaryBase::set_transformation_matrix(const DenseMatrix<Number> & matrix)
+{
+  // Make a deep copy of matrix
+  DenseMatrix<Number> * new_matrix = new DenseMatrix<Number>;
```
roystgnr: Likewise on libmesh_make_unique

src/fe/fe_base.C
```diff
@@ -2337,8 +2354,32 @@ compute_periodic_constraints (DofConstraints & constraints,
...
+                              constraint_row->insert(std::make_pair(their_dof_g,
+                                                                    their_dof_value));
+                            }
+                          else
+                            {
+                              // In this case the current variable is constrained
+                              // in terms of other variables. We assume that all
+                              // variables in this constraint have the same FE
+                              // type, and hence we can create the constraint
+                              // row contribution by multiplying their_dof_value
```
roystgnr: If we're going to assume that (which seems reasonable!), then we need to document it (in the header doxygen for set_transformation_matrix() should do it) and we need to assert it here.

src/base/periodic_boundary_base.C
dknez: I tried "return _transformation_matrix;" initially, but that didn't compile. What did you have in mind?

src/fe/fe_base.C
dknez: Agreed. This is what I had in mind when I said this PR needs some more work. I'll make that change.

dknez:
Thanks for the feedback @roystgnr. I'll update the PR shortly based on your comments.

Regarding test coverage: My plan was to merge this PR then add a new example in a subsequent PR. It seems like a lot of work to set up good unit tests in this case, so I'd rather just start by adding an example. Sound OK?

roystgnr:
Sounds good.

dknez:
I've updated the PR based on your comments.

dknez:
But I introduced a compilation error...

src/base/periodic_boundary_base.C
roystgnr: Ah!  unique_ptr::operator bool is declared *explicit*, and that auto-applies in "if" and a dozen other cases but "return" isn't one of them.  ```return _transformation_matrix.get();``` should work, though, right?  That's what I usually do out of force of habit.  ```bool(_transformation_matrix)``` ought to be kosher, too, and is probably the most readable idiom.

jwpeterson:
```
return !!_transformation_matrix;
```
I'm joking, but would that really be so bad?

dknez: I changed it to bool(_transformation_matrix)

dknez:
I think it's ready now. Let me know what you think.

src/base/periodic_boundary_base.C
roystgnr: We'll go with bool().

@jwpeterson:
Bang bang?  I shot you down.

roystgnr:
libmesh_assert_msg and libmesh_make_unique?

dknez:
Done.

src/base/periodic_boundary_base.C
```diff
@@ -64,6 +70,44 @@ bool PeriodicBoundaryBase::is_my_variable(unsigned int var_num) const
...
+  return *_transformation_matrix;
+}
+
+
+
+void PeriodicBoundaryBase::set_transformation_matrix(const DenseMatrix<Real> & matrix)
+{
+  // Make a deep copy of matrix
+  this->_transformation_matrix = libmesh_make_unique<DenseMatrix<Real>>();
+  *(this->_transformation_matrix) = matrix;
```
roystgnr: I was expecting that three lines to turn into one line, not two, but apparently that's our fault - we don't have a DenseMatrix copy constructor!

Presumably that was an intentional decision to prevent people from accidentally passing a copy instead of passing-by-reference, but still...

dknez: Yes, I also expected there to be a copy constructor.

dknez:
Thanks for the review. This already passed the tests in the previous incarnation. Should I wait for the tests to run again, or just merge?

roystgnr:
Let's be paranoid; thanks.

PR #1636, opened 2018-03-18
Small fix for send/receive strings

permcody:
Not a super extensive test, but does fix the bug!

### DIFF
--- a/include/parallel/parallel_implementation.h
+++ b/include/parallel/parallel_implementation.h
@@ -2533,13 +2533,13 @@ inline void Communicator::receive (const unsigned int src_processor_id,
   req.add_post_wait_work
     (new Parallel::PostWaitCopyBuffer<std::vector<T>,
      std::back_insert_iterator<std::basic_string<T>>>
-     (tempbuf, std::back_inserter(buf)));
+     (*tempbuf, std::back_inserter(buf)));
 
   // Make the Request::wait() then handle deleting the buffer
   req.add_post_wait_work
     (new Parallel::PostWaitDeleteBuffer<std::vector<T>>(tempbuf));
 
-  this->receive(src_processor_id, tempbuf, req, tag);
+  this->receive(src_processor_id, *tempbuf, req, tag);
 }
 
 
--- a/tests/parallel/parallel_test.C
+++ b/tests/parallel/parallel_test.C
@@ -400,18 +400,20 @@ public:
       TestCommWorld->size();
 
     std::vector<unsigned int> src_val(3), recv_val(3);
+    std::string recv_val_str;
 
     src_val[0] = 0;
     src_val[1] = 1;
     src_val[2] = 2;
 
-    Parallel::Request request;
+    Parallel::Request request, request_str;
 
     if (TestCommWorld->size() > 1)
       {
         // Default communication
         TestCommWorld->send_mode(Parallel::Communicator::DEFAULT);
 
+        // Numbers
         TestCommWorld->receive (procdown,
                                 recv_val,
                                 request);
@@ -425,6 +427,19 @@ public:
 
         for (std::size_t i=0; i<src_val.size(); i++)
           CPPUNIT_ASSERT_EQUAL( src_val[i] , recv_val[i] );
+
+        // Strings
+        // Nonblocking
+        TestCommWorld->receive (procdown,
+                                recv_val_str,
+                                request_str);
+
+        TestCommWorld->send (procup,
+                             _number[0]);
+
+        Parallel::wait (request_str);
+
+        CPPUNIT_ASSERT_EQUAL( recv_val_str, _number[0] );
 
         // Synchronous communication
         TestCommWorld->send_mode(Parallel::Communicator::SYNCHRONOUS);


jwpeterson:
I think this passed debug mode testing because unit tests are not run in debug mode...
```
1) test: ParallelTest::testIrecvSend (F) line: 442 ../../tests/parallel/parallel_test.C
equality assertion failed
- Expected: 
- Actual  : Zero
```

permcody:
Yeah, I clearly didn't think through all the parallel cases. I'll work on a better unit test later. At least I can move forward on what I need this for.

include/parallel/parallel_implementation.h
```diff
@@ -2533,13 +2533,13 @@ inline void Communicator::receive (const unsigned int src_processor_id,
   req.add_post_wait_work
     (new Parallel::PostWaitCopyBuffer<std::vector<T>,
      std::back_insert_iterator<std::basic_string<T>>>
-     (tempbuf, std::back_inserter(buf)));
+     (*tempbuf, std::back_inserter(buf)));
```
permcody: @roystgnr - This seems like the right fix to me. However, when I do some debugging, I get a message like this from MPI: `Message from rank 1 and tag 0 truncated; 5 bytes received but buffer size is 0`. Looks like my message was properly received ("Hello"), but the buffer size is zero... 

I'm not sure why the buffer size would be zero since we are using a back_inserter. I must admit, I haven't used a `back_insert_iterator` before and this is the only place in all of `parallel[_implementation].h` where this exact construct is used but I don't see the problem. I'm tempted to try the solution that you have below for std::set even though that's not the right solution just to see if something works. BTW - I verified that the blocking version works fine so it does have something to do with all of this "post wait work" stuff. Any insight here?

roystgnr: Oh, hell.

Non-blocking receives of vectors require the vector object to be resized before the receive is called.

Our excuse for that was that the only mechanisms we could think of to communicate data size would require either busy-looping or a separate blocking size datum message.  Additional suggestions would be welcome, though; I haven't thought about the problem in years.

None of this is documented in parallel.h, but only in https://sourceforge.net/p/libmesh/mailman/message/20734660/

I have no excuse for that.

roystgnr: In the meantime, we could impose the same requirement on non-blocking receives of strings, if that would work for you.  Look at the string length when receive is called, and use that to size tempbuf?

permcody: > Look at the string length when receive is called, and use that to size tempbuf?

You'll have to excuse my ignorance, but are you suggesting that I add an MPI_Probe call in that method?

I'm pretty flexible. I need this capability to communicate binary blobs from one processor to another. These binary blobs will always just look like a vector of characters. Actually, I might end up just needed a series of gather operations, since I believe in all my cases, they will actually need to be gathered (but on different ranks other than 0).

roystgnr: MPI_Probe is what we do for blocking receives, but MPI_IProbe wouldn't work for non-blocking, because you'd need the probe to succeed before you could post the MPI_Receive, so you'd have to do a blocking probe, so you might as well do a blocking receive in the first place.

Are you sending lots of blobs at once?  If so, then the most efficient solution I can think of would be to:

1. In libMesh, change our basic_string code to stop using temporary buffers.  That was technically required to be C++98 compliant, but C++11 forces strings to give us writeable contiguous arrays of characters.  https://stackoverflow.com/questions/1986966/does-s0-point-to-contiguous-characters-in-a-stdstring

2.  In your code when the sends are ready, do an alltoall on message lengths, then do non-blocking sends of any messages with greater than 0 length.

3a.  In your code after the alltoall, resize strings for incoming data accordingly.

4a.  In your code after the resize, post non-blocking receives for those strings

5a.  In libMesh, add a ```Parallel::waitany(vector<Request>)``` wrapper to MPI_Waitany

6a.  In your code, loop around waitany and handle each string as soon as it comes in.

Or for a slightly easier and potentially less scalable solution:

3b.  In your code after the alltoall, skip the resize; loop over the number of processors with incoming message lengths greater than 0, post a blocking receive from any_source each time, and handle each string as it comes in.  Make sure you use a unique tag for sends and receives so you don't accidentally pick up a later send from a processor you've already handled.

Or for a much easier but potentially much less scalable solution:

3c.  In your code after the alltoall, skip the resize; just loop over processors, skip those with incoming message lengths greater than 0, post a blocking receive for each, and handle each string in processor order as it comes in.

1, 2, 3c is what I did in #1600, and it was a big improvement over the send_receive round-robin which preceded it.  I think a or b would be much better but I don't want to start microoptimizing until I'm sure I can factor things out well enough to try the optimizations everywhere at once, which in turn is waiting for me to finally get the node partitioning mess solid.

friedmud: @roystgnr There is no need to send the message sizes first.  Look at `Communicator::nonblocking_receive_packed_range()`: it takes a `Status` (which came from `Communicator::packed_range_probe`)... that `Status` _knows_ the size of the incoming message and allows you to size the receiving datastructure correctly.

There is no "single function" way to do a non-blocking receive of an arbitrarily sized array.  You have to probe until you find the message... then receive.

jwpeterson:

roystgnr:
Closing in favor of https://github.com/libMesh/TIMPI/pull/155

PR #1637, opened 2018-03-19
Convert more for loops to C++11 range-based for loops

jwpeterson:


roystgnr:
That is some *quality* patch writing!

I have a Dr's appt this morning but I'll try to have this looked through early this afternoon.

src/base/dof_map_constraints.C
```diff
@@ -3149,9 +3081,8 @@ void DofMap::process_constraints (MeshBase & mesh)
...
   RCSet unexpanded_set;
 
-  for (DofConstraints::iterator i = _dof_constraints.begin();
-       i != _dof_constraints.end(); ++i)
-    unexpanded_set.insert(i->first);
+  for (const auto & i : _dof_constraints)
+    unexpanded_set.insert(i.first);
 
   while (!unexpanded_set.empty())
     for (RCSet::iterator i = unexpanded_set.begin();
```
roystgnr: Wow; finally I see a traditional for loop that couldn't be turned into a range for!

include/systems/generic_projector.h
```diff
@@ -672,10 +662,8 @@ void GenericProjector<FFunctor, GFunctor, FValue, ProjectionAction>::operator()
 
       // Loop over all the variables we've been requested to project, to
       // do the projection
-      for (std::size_t v=0; v!=variables.size(); v++)
+      for (const auto & var : variables)
```
roystgnr: I know it's trivial outside of the innermost inner loops, but - is there a performance implication to using a reference rather than a copy here?

The reference will be a pointer under the hood, so unless the compiler can prove that the pointee never changes it will have to hang on to and dereference 8 bytes rather than just hanging on to 4.

And we have a bunch of function calls, including one that I'm sure will never be inlined, so unless the compiler does ridiculous levels of inter-procedural optimization it won't be able to prove that those calls don't change the variables container in mid-iteration.

roystgnr: I know, this particular example called size() on every loop iteration, so the reference-based range for is probably *still* faster, but in theory it seems like removing the & would buy a couple more nanoseconds on top of that.

jwpeterson: I always prefer `const auto &` because somebody could potentially come along later and change the underlying type in `variables` to e.g. a `shared_ptr<T>`, in which case `const auto foo` would then trigger a reference-counted copy construction (and destruction) at each iteration of the loop. Even worse would be if the underlying type changed to simply `T`...

I'm willing to be convinced that there's a performance difference between `const auto` and `const auto &` for the case of dumb pointers, but I don't think it's worth the possibility of introducing a different bug or different performance penalty due to a later refactoring down the line...

jwpeterson: > Even worse would be if the underlying type changed to simply T...

By this I meant some case where `T` is expensive to copy, not the `unsigned int` case here!

src/mesh/mesh_modification.C
```diff
@@ -1834,29 +1801,18 @@ void MeshTools::Modification::flatten(MeshBase & mesh)
...
+    {
+      Elem * added_elem = mesh.add_elem(new_elem);
 
-#ifndef NDEBUG
-        dof_id_type added_id = added_elem->id();
-#endif
+      // If the Elem, as it was re-added to the mesh, now has a
+      // different ID (this is unlikely, so it's just an assert)
+      // the boundary information will no longer be correct.
+      libmesh_assert_equal_to (new_elem->id(), added_elem->id());
```
roystgnr: This actually isn't equivalent to the original uglier code.  IIRC we had a bug at one point where adding a new element changed the id of that element.  We have to save the original id before add_elem to be sure that hasn't happened.

jwpeterson: OK, I can fix that.

src/solvers/petscdmlibmeshimpl.C
```diff
@@ -282,20 +284,19 @@ static PetscErrorCode  DMCreateFieldDecomposition_libMesh(DM dm, PetscInt * len,
...
-      for (std::map<std::string, unsigned int>::const_iterator bit = dlm->blockids->begin(); bit != dlm->blockids->end(); ++bit) {
-        unsigned int b = bit->second;
-        MeshBase::const_element_iterator el     = sys->get_mesh().active_local_subdomain_elements_begin(b);
-        MeshBase::const_element_iterator end_el = sys->get_mesh().active_local_subdomain_elements_end(b);
+      // Iterate only over this DM's blocks.
+      for (const auto & pr : *(dlm->blockids)) {
+        MeshBase::const_element_iterator
+          el     = sys->get_mesh().active_local_subdomain_elements_begin(pr.second),
+          end_el = sys->get_mesh().active_local_subdomain_elements_end(pr.second);
         for ( ; el != end_el; ++el) {
```
roystgnr: Why not change this to another as_range() use?

src/solvers/petscdmlibmeshimpl.C
```diff
@@ -512,12 +510,12 @@ PetscErrorCode DMlibMeshCreateFieldDecompositionDM(DM dm, PetscInt dnumber, Pets
...
-  else { /* Empty splits indicate default: split all variables with one per split. */
+  else { // Empty splits indicate default: split all variables with one per split.
     PetscInt d = 0;
-    for (std::map<std::string, unsigned int>::const_iterator vit = ddlm->varids->begin(); vit != ddlm->varids->end(); ++vit) {
+    for (const auto & pr : (*ddlm->varids)) {
       ddlm->decomposition->push_back(std::set<unsigned int>());
-      unsigned int vid = vit->second;
-      std::string vname = vit->first;
+      unsigned int vid = pr.second;
+      std::string vname = pr.first;
```
roystgnr: How is this not throwing up an unused variable warning?

src/solvers/petscdmlibmeshimpl.C
```diff
@@ -565,12 +563,12 @@ PetscErrorCode DMlibMeshCreateDomainDecompositionDM(DM dm, PetscInt dnumber, Pet
...
-  else { /* Empty splits indicate default: split all blocks with one per split. */
+  else { // Empty splits indicate default: split all blocks with one per split.
     PetscInt d = 0;
-    for (std::map<std::string, unsigned int>::const_iterator bit = ddlm->blockids->begin(); bit != ddlm->blockids->end(); ++bit) {
+    for (const auto & pr : (*ddlm->blockids)) {
       ddlm->decomposition->push_back(std::set<unsigned int>());
-      unsigned int bid = bit->second;
-      std::string bname = bit->first;
+      unsigned int bid = pr.second;
+      std::string bname = pr.first;
```
roystgnr: And again

src/solvers/petscdmlibmeshimpl.C
jwpeterson: Must not have matched any of the various regexes I attempted to employ, will update now.

roystgnr:
:+1:

src/solvers/petscdmlibmeshimpl.C
jwpeterson: No clue, I thought maybe they were i a section of code that was ifdef'd out but that does not appear to be the case. I'll drop the unused vars...

PR #1638, opened 2018-03-19
Systems of equations ex9

dknez:
Added a new example.


dknez:
The commits should probably be squashed, but I would like to squash commits 1 and 3 together and 2 and 4 together, and I wasn't sure how to do that. Maybe someone else knows, or we can just leave it as is.

dknez:
Looks like the tests are failing to open the exo file, but it works fine for me. I added "periodiccomponent.exo" to "data" in Makefile.am, and as far as I know that's all I need to do. Any suggestions?

jwpeterson:
> Maybe someone else knows, or we can just leave it as is.

Interactive rebase (`git rebase -i`) can be used to move commits around and squash them.

> Looks like the tests are failing to open the exo file, but it works fine for me.

I don't see anything obvious that you've missed in your PR... I'll take a look later tonight and see if I can figure out what's going on.

examples/systems_of_equations/systems_of_equations_ex9/Makefile.am
```diff
@@ -0,0 +1,11 @@
+example_name  = systems_of_equations_ex9
+check_SCRIPTS = run.sh
+install_dir   = $(examples_install_path)/systems_of_equations/ex9
+data          = systems_of_equations_ex9.C run.sh periodiccomponent.exo
+sources       = $(data) run.sh
+
+CLEANFILES = solution.exo
+
```
jwpeterson: Have a look at miscellaneous_ex9's Makefile.am. I think you need a section in the Makfile.am like the following to ensure that the input mesh gets linked to the build directory.
```
# also need to link files for VPATH builds
if LIBMESH_VPATH_BUILD
  BUILT_SOURCES = .linkstamp
.linkstamp:
        -rm -f miscellaneous_ex9.exo && $(LN_S) -f $(srcdir)/miscellaneous_ex9.exo
        $(AM_V_GEN)touch .linkstamp

  CLEANFILES += miscellaneous_ex9.exo .linkstamp
endif
```

dknez:
@jwpeterson thanks for the tips, it should be in better shape now. (I have been using "rebase -i" to squash, but I didn't realize it's so useful for reordering commits as well, thanks for letting me know!).

A separate issue is: Is this example correct, i.e. is the "cyclic periodic" condition implemented correctly. I'm not really sure about that, and I think that needs some more investigation. Ideally we'd plot 12 copies of the solution to make a full circle to see if the solution is continuous, but that seems like a lot of work, so any assistance with the validation of this example would be appreciated.

dknez:
The example runs now, which is good.

However, I did the following test:
- solve the model as per this example
- make another version where the mesh and load are rotated by 30 degrees and solve again
- load the solutions side-by-side in paraview

When I did this, the solutions were not continuous. So there is something wrong with the periodic boundary condition here. I'll need to look into it further.

dknez:
This latest commit fixes the discontinuous BC issue. The problem was that the transformation matrix used in get_corresponding_pos should be the inverse of the one used to rotate the variables on the periodic BC. This is because in get_corresponding_pos we map from "my boundary" to "paired boundary", whereas inside compute_periodic_constraints it's the other way around.

I think this is ready to merge now.

roystgnr:
:+1:

But are there any obvious asserts you could throw in there?  It would be nice if any future regressions make themselves known via "make check" failures, not manual paraview juggling.

dknez:
I don't think the asserts would be easy to add, unfortunately. We'd basically have to extract entries from the solution vector corresponding matching points on each side of the boundary condition, and assert that they've been rotated correctly.

PR #1640, opened 2018-03-25
Enabled adjacent periodic boundary conditions

### LINKED ISSUE #1639: Cannot use adjacent periodic boundary conditions
When I was debugging systems_of_equations_ex9, one test I did was to create a new mesh that is "two copies" of the current mesh, with one rotated by 30 degrees so that we had two distinct segments of the periodic domain. I then impose a separated periodic condition on each segment, with the goal of checking that the overall solution was continuous. However, this failed with the error below:

```
Assertion `s_neigh != libMesh::invalid_uint' failed.
s_neigh = 4294967295
libMesh::invalid_uint = 4294967295


Stack frames: 23
1: libMesh::MacroFunctions::report_error(char const*, int, char const*, char const*)
2: libMesh::FEAbstract::compute_periodic_node_constraints(libMesh::NodeConstraints&, libMesh::PeriodicBoundaries const&, libMesh::MeshBase const&, libMesh::PointLocatorBase const*, libMesh::Elem const*)
3: /home/dknez/software/libmesh_install/dbg_real/libmesh/lib/libmesh_dbg.so.0(+0x1678384) [0x7fffee8d7384]
4: /home/dknez/software/libmesh_install/dbg_real/libmesh/lib/libmesh_dbg.so.0(+0x16a9b63) [0x7fffee908b63]
5: libMesh::DofMap::create_dof_constraints(libMesh::MeshBase const&, double)
6: libMesh::System::reinit_constraints()
7: libMesh::System::init_data()
8: libMesh::ImplicitSystem::init_data()
9: libMesh::LinearImplicitSystem::init_data()
10: libMesh::System::init()
11: libMesh::EquationSystems::init()
12: FeSystem::try_equation_systems_init(libMesh::EquationSystems&)
13: FeSolverFunctions::call_linear_elasticity_fe_solver(libMesh::MeshBase&, TypeMaps&, FeSolverInit const&, FeSolverSolve const&, SolveInfo<FeSolverResponse>&, unsigned int, AppOutputHelper&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, std::__debug::map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, ComponentVisualizationSummary, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, ComponentVisualizationSummary> > >*)
14: /home/dknez/akselos-dev/scrbe/build/bin/fe_solver-dbg_real() [0x4a1fe5]
15: MessageHandler::dispatch_message(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, Json::Value const&, std::unique_ptr<MessageHandler, std::default_delete<MessageHandler> >&)
16: MessageHandler::extract_message(Json::Value&, std::unique_ptr<MessageHandler, std::default_delete<MessageHandler> >&)
17: MessageHandler::parse_message(char const*, char const*, std::unique_ptr<MessageHandler, std::default_delete<MessageHandler> >&)
18: /home/dknez/akselos-dev/scrbe/build/bin/fe_solver-dbg_real() [0x4a41f7]
19: /home/dknez/akselos-dev/scrbe/build/bin/fe_solver-dbg_real() [0x4a3c85]
20: /home/dknez/akselos-dev/scrbe/build/bin/fe_solver-dbg_real() [0x4a3c43]
21: __libc_start_main
22: /home/dknez/akselos-dev/scrbe/build/bin/fe_solver-dbg_real() [0x49fa29]
[0] /home/dknez/software/libmesh-src/src/fe/fe_abstract.C, line 986, compiled Mar 22 2018 at 13:18:15
```

I gather that the issue is that the point locator that we use fails because we have two overlapping periodic surfaces from the two different segments.

It'd be nice if we could handle this case more gracefully, IMO.

dknez:
Added set_point_locator_subdomains to PeriodicBoundaryBase. This allows us to tell the PointLocator to look over a specific set of subdomains rather than over the whole mesh. This enables us to fix the issue with adjacent periodic boundary conditions, since we can now tell the PointLocator which subdomains to consider for each periodic BC.

Updated systems_of_equations_ex9 to now use two adjacent and unconnected domains to show that the new approach works.

This resolves #1639.


include/base/periodic_boundary_base.h
```diff
@@ -145,6 +155,14 @@ class PeriodicBoundaryBase
    * pointer is treated as equivalent to the identity matrix.
    */
   std::unique_ptr<DenseMatrix<Real>> _transformation_matrix;
+
+  /**
+   * Imposing a periodic boundary uses a PointLocator to match points on
+   * the two sides of the boundary. This set specifies the subdomains
+   * that the PointLocator searches over. If it is empty, then it is
+   * ignored and we search over all subdomains.
+   */
```
roystgnr: Could we move this comment to the setter function (in an ideal world users would just read about the APIs and ignore private internals) and add an example (or a pointer to the new example app) describing its use?

dknez:
@roystgnr I updated the comment for get_point_locator_subdomains based on your suggestion.

As per the discussion in #1639, let me know how you want to proceed on this. The fix in this PR works for me, but if you prefer a different approach, that's fine with me. The key thing I want to ensure is that the updated version of systems_of_equations_ex9 in this PR works.

roystgnr:
IMHO this isn't the best fix for this case, but it's an adequate fix for this case, I'm too far behind on other issues to put together an alternative in the near future, and even if we *do* implement my ideal fix at some point in the future it will merely render this API redundant, not broken, so there's no serious obstacle to merging now.

dknez:
OK, thanks.

I agree that your proposed fix would be better since it wouldn't require the user to specify subdomain IDs. But I also agree that the current fix is adequate, so it makes sense to merge now and improve later.

PR #1641, opened 2018-03-26
Improvement to how we handle overlapping periodic BCs

dknez:
Roy Stogner suggested a better way to handle overlapping periodic BCs: use point_locator.operator()(p, candidate_elements) and find the element that contains the correct boundary ID. This is more user-friendly and less error-prone since it does not require the users to specify subdomain IDs themselves.

systems_of_equations_ex9 has been updated by removing the part where we set the subdomain IDs, and it works.


roystgnr:
Well, this violates ```libmesh_assert_not_equal_to(beggars, choosers)```, but I'd be happy to just comment that out.

:+1: once Civet is done.

PR #1642, opened 2018-04-05
Add periodic boundaries *before* Mesh::read()

roystgnr:
Otherwise by the time the DistributedMesh knows you have periodic
boundaries, it's too late to recover already-deleted periodic
neighbors.

This fixes failures with --enable-parmesh for me.

Pinging @dknez in case he's got the same issue in any Akselos code.

### DIFF
--- a/examples/systems_of_equations/systems_of_equations_ex9/systems_of_equations_ex9.C
+++ b/examples/systems_of_equations/systems_of_equations_ex9/systems_of_equations_ex9.C
@@ -356,10 +356,6 @@ int main (int argc, char ** argv)
 
   // Create a 3D mesh distributed across the default MPI communicator.
   Mesh mesh(init.comm());
-  mesh.read("systems_of_equations_ex9.exo");
-
-  // Print information about the mesh to the screen.
-  mesh.print_info();
 
   // Create an equation systems object.
   EquationSystems equation_systems (mesh);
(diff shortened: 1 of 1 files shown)


dknez:
Thanks for the ping. We're still using ReplicatedMesh everywhere, so this hasn't been an issue at our end, but thanks for this fix since it'll be relevant for us if/when we start using DistributedMesh.

friedmud:
In general: this isn't possible.  We have lots of period BC autodetection in MOOSE... that requires that you have the mesh first!

Our plan is to output pre-split checkpoint files that contain the proper ghosting by using GhostingFunctors during the splitting (we're already doing this now for extra ghosting layers... but we haven't tried periodic BCs yet).

permcody:
> I think the short term solution for MOOSE has to be temporarily disabling distribution during mesh setup, then reenabling it (and manually calling MeshCommunication::delete_remote_elements()) once you're sure all the ghosting functors are set up.

This is precisely the idea we've implemented in the newest MOOSE system "RelationshipManagers". Objects can basically set them up through validParams (not during construction as we do with everything else in MOOSE). We can then construct them (just abstractions of your GhostingFunctor) and prevent the deletion of remote elements for a little while.

We have a few examples of these in MOOSE already, more work is needed but the idea seems to be a nice abstraction for users. 

This object just ghosts extra layers (just a layer on the existing libMesh object):
https://github.com/idaholab/moose/blob/devel/framework/src/relationshipmanagers/ElementSideNeighborLayers.C

Here's how it's used:
https://github.com/idaholab/moose/blob/devel/test/src/userobjects/GhostUserObject.C

The test:
https://github.com/idaholab/moose/blob/devel/test/tests/mesh/ghost_functors/geometric_neighbors.i

The logic for this system is somewhat scattered, due to MOOSE's Action system but we basically attach these things and make decisions about deleting elements or not.
https://github.com/idaholab/moose/blob/devel/framework/src/relationshipmanagers/RelationshipManager.C

The end result is users can just register RelationshipManagers with objects that need more "stencil" data of any kind. It's not an input file thing, but a file thing that allows them to tap into the very complex setup process (and timing) of getting all of this right.

PR #1644, opened 2018-04-11
Moving discontinuous data setup to MeshOutput.

### LINKED ISSUE #1503: Discontinuous Time Dependent Solution Output

kevindugan:
These changes make the outputting of discontinuous nodal data more uniform with continuous nodal data. This should resolve #1503 and enable the merge in idaholab/moose#11210. 

These changes have been tested with the moose test suite using the `--distributed-mesh -p 4` option.

### DIFF
--- a/include/mesh/exodusII_io.h
+++ b/include/mesh/exodusII_io.h
@@ -202,7 +202,7 @@ public:
    */
   void write_nodal_data_discontinuous (const std::string &,
                                        const std::vector<Number> &,
-                                       const std::vector<std::string> &);
+                                       const std::vector<std::string> &) libmesh_override;
 
   /**
    * Write out global variables.
--- a/include/mesh/mesh_output.h
+++ b/include/mesh/mesh_output.h
@@ -90,12 +90,30 @@ public:
                                        const std::set<std::string> * system_names=libmesh_nullptr);
 
   /**
+   * This method implements writing a mesh with discontinuous data to a
+   * specified file where the data is taken from the \p EquationSystems
+   * object.
+   */
+  virtual void write_discontinuous_equation_systems (const std::string &,
+                                                     const EquationSystems &,
+                                                     const std::set<std::string> * system_names=libmesh_nullptr);
+
+  /**
    * This method implements writing a mesh with nodal data to a
    * specified file where the nodal data and variable names are provided.
    */
   virtual void write_nodal_data (const std::string &,
                                  const std::vector<Number> &,
                                  const std::vector<std::string> &)
+  { libmesh_not_implemented(); }
+
+  /**
+   * This method implements writing a mesh with discontinuous data to a
+   * specified file where the nodal data and variables names are provided.
+   */
+  virtual void write_nodal_data_discontinuous (const std::string &,
+                                               const std::vector<Number> &,
+                                               const std::vector<std::string> &)
   { libmesh_not_implemented(); }
 
   /**
@@ -114,6 +132,15 @@ public:
   virtual void write_nodal_data (const std::string &,
                                  const NumericVector<Number> &,
                                  const std::vector<std::string> &);
+
+  /**
+   * This method should be overridden by output formats wanting to
+   * export discontinuous data.
+   */
+  virtual void write_nodal_data_discontinuous (const std::string &,
+                                               const NumericVector<Number> &,
+                                               const std::vector<std::string> &)
+  { libmesh_not_implemented(); }
 
   /**
    * Return/set the precision to use when writing ASCII files.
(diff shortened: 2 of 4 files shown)


jwpeterson:
So we can't just remove this public API, it's used in `systems_of_equations_ex6` and possibly GRINS (I can't see exactly why that failed at the moment). Is there any way to make this change backwards compatible?

kevindugan:
@jwpeterson I hadn't realized that this was being used anywhere. I'm in the process of putting it back in; running the tests locally now.

pbauman:
GRINS-devel failed because I run `make check` on libMesh in that build, so the relevant libMesh example failed

jwpeterson:
@kevindugan thanks. Yeah, since it's generally not possible to know who uses *any* of the public APIs, we typically try to provide a longish period of deprecation before any of them are permanently removed.

roystgnr:
My personal standard is "The API needs to be marked libmesh_deprecated() and given a non-deprecated replacement, and those changes have to make it into at least one (preferably two) official libMesh releases, and then we can remove the deprecated API after the final transitional release is branched.  We can't always live up to that for various technical reasons, and sometimes we go overboard because we *forget* to remove the deprecated API after we've all stopped using it (looking at you, optional prepare_for_use() argument), but it's what to shoot for.

jwpeterson:
This looks OK to me. Just so that I'm clear, `ExodusII_IO::write_discontinuous_exodusII()` can now be marked `libmesh_deprecated()` and implemented in terms of the new `MeshOutput` function, correct?

kevindugan:
@jwpeterson That is correct.

include/mesh/exodusII_io.h
```diff
@@ -200,9 +200,9 @@ class ExodusII_IO : public MeshInput<MeshBase>,
   /**
    * Write out a discontinuous nodal solution.
    */
-  void write_nodal_data_discontinuous (const std::string &,
+  void write_discontinuous_nodal_data (const std::string &,
```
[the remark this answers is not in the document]
kevindugan: Ok, I'm testing the changes locally now.

permcody:
Timeouts are unfortunate, but do happen. Often times we just invalidate the test target and try again. We try to revisit tests that timeout often to resize them or make them smaller to avoid those kinds of problems in the future.

jwpeterson:
Yeah I have seen this test (`misc/line_source.test`) timeout in the past, there's no way it's due to your recent changes. I will invalidate.

roystgnr:
line-freaking-source times out frequently.  If it had just been me I wouldn't even have invalidated, I'd just say we're good to merge.

Actually, should we squash these 3 commits before merging?  I don't want to be too anal about that sort of thing (I say, while eyeing #1621 nervously) but the latter 2 are basically fixups.

kevindugan:
@roystgnr I have no problem with people being anal when it comes to software.

jwpeterson:
> Actually, should we squash these 3 commits before merging?

Was planning to use Github's "Squash and Merge" option.

roystgnr:
> Github's "Squash and Merge" option.

That's excellent.

include/mesh/mesh_output.h
```diff
@@ -115,6 +133,15 @@ class MeshOutput
                                  const NumericVector<Number> &,
                                  const std::vector<std::string> &);
 
+  /**
+   * This method should be overridden by output formats wanting to
+   * export discontinuous data.
+   */
+  virtual void write_nodal_data_discontinuous (const std::string &,
+                                               const NumericVector<Number> &,
+                                               const std::vector<std::string> &)
```
roystgnr: Wait a minute - why do we have this overload?  We're only overloading the std::vector version, and that gives me -Woverloaded-virtual warnings when the NumericVector version is hidden in the derived class.

kevindugan: I'm looking into this now; I thought I had checked for warnings before I pushed. I may have forgotten to include the `using MeshOutput<MeshBase>::write_nodal_data_discontinuous` line in the final version.

kevindugan:
@roystgnr I've pushed a commit to my branch which specifies the `MeshBase` class in the `exodus_io.h` file. This resolves the warnings about virtual functions. Will I need to open a new pull request, or can we piggyback off this one?

I must have some settings on my Mac which override the warnings from gcc. I could only see these warnings produced when I switched to my Ubuntu machine. I'll know for the future.

jwpeterson:
@kevindugan I'll rebase this most recent commit on master and open a new PR. We did a squash merge here so the history of this branch is not in master.

jwpeterson:
Oh, never mind I guess you already did base your commit on current master... So I'd say just open a new PR with that commit.

kevindugan:
@jwpeterson Yes, I did a rebase on this branch. I'll open up a new PR then. I'll also squash the commits on this branch while I'm at it.

PR #1647, opened 2018-04-12
Run example methods in debugging-friendly order

roystgnr:
If something fails, we want to know as much about the failure as
possible, not have "make check" end after a less-informative build
fails first.


roystgnr:
You mean drop the lines *removing* trailing whitespace?

jwpeterson:
Sorry, read the diff backwards

PR #1653, opened 2018-04-16
Fixing Warnings about Virtual Methods

kevindugan:
These changes resolve warnings about hidden virtual methods. Related to #1503, #1644 idaholab/moose#2716

### DIFF
--- a/include/mesh/exodusII_io.h
+++ b/include/mesh/exodusII_io.h
@@ -189,6 +189,7 @@ public:
    * avoid warnings about hidden overloaded virtual functions.
    */
   using MeshOutput<MeshBase>::write_nodal_data;
+  using MeshOutput<MeshBase>::write_nodal_data_discontinuous;
 
   /**
    * Write out a nodal solution.
--- a/include/mesh/mesh_output.h
+++ b/include/mesh/mesh_output.h
@@ -134,15 +134,6 @@ public:
                                  const std::vector<std::string> &);
 
   /**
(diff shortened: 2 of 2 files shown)


roystgnr:
This fixes the warning, but I still don't know what the answer is to my question: why do we have that second overload?  I'd rather not promise users support in an abstract base class for an interface which has *no* implementation in any derived classes.

kevindugan:
>I'd rather not promise users support in an abstract base class for an interface which has no implementation in any derived classes

That's fair. I was concentrated on getting the specific Exodus case to work through this interface. I was not accessing the `MeshOutput` api using the `NumericVector` method. I can implement this method to localize the parallel solution and call the method using `std::vector`, similar to what is done for the continuous case. Would that be acceptable? Or would you prefer removing that method in the `MeshOutput` class?

roystgnr:
IMHO we ought to just remove it until we have a use case for it.

kevindugan:
Sounds good. I'll push those changes.

roystgnr:
Thanks!

> I can implement this method to localize the parallel solution and call the method using std::vector, similar to what is done for the continuous case.

Just to be clear, though I'm definitely *not against* this, if you ever think that's a valuable refactoring in the future; I just think it ought to wait until we're ready to use it, not implement it for the sake of implementing it.  Methods without test coverage are bitrot waiting to happen.  And since your needs are satisfied by the std::vector option, adding a NumericVector overload plus test before anyone needs it would feel like just makework.

PR #1656, opened 2018-04-16
Don't use LIBMESH_TLS in contrib/metis.

jwpeterson:
This change was also made in PETSc's "pkg-metis" repo
(https://bitbucket.org/petsc/pkg-metis) in c27a7c3 (tag: v5.1.0-p3),
originally in response to a problem with Xcode 7.3, and it has never
been reenabled, so I don't think it can be very important.

It is also nice to be remove libmesh-specific stuff from Metis, and
get rid of the LIBMESH_IS_COMPILING_METIS define. For now I have left
a comment on the variables that originally used thread local storage,
just in case we need to switch it back on later.

### DIFF
--- a/contrib/metis/GKlib/error.c
+++ b/contrib/metis/GKlib/error.c
@@ -15,42 +15,20 @@ This file contains functions dealing with error reporting and termination
 
 #include <GKlib.h>
 
-/**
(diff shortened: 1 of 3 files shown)


roystgnr:
:+1:

PR #1658, opened 2018-04-17
Make solution output more robust to discontiguous ids

roystgnr:
This does *not* yet handle Xdr mesh output, which I'm told is what thwarted @jwpeterson when he first created those first commits, but I think I see and can fix the issue with that.

If I can do so quickly, I'll add it to the end of this PR.  If I can't, then we're not going to want to allow users to wholly disable node renumbering until we completely support discontiguous node ids, but partial support is better than nothing.


roystgnr:
Got an attempt at XdrIO done, trying it out myself (and letting Civet Test debug finish) before pushing.

PR #1660, opened 2018-04-19
Delay new node partitioning

roystgnr:
The second half of the work in #1659; this delays partitioning nodes newly created by refinement, allowing us to partition them with any heuristic but without accidentally repartitioning old nodes if the user has disallowed that.

### DIFF
--- a/include/mesh/mesh_refinement.h
+++ b/include/mesh/mesh_refinement.h
@@ -300,8 +300,10 @@ public:
 
   /**
    * Add a node to the mesh.  The node should be node n of child c of
-   * parent Elem parent.  The processor_id is assigned to all newly
-   * created nodes.
+   * parent Elem parent.  The processor id \proc_id is used if
+   * necessary to help determine numbering of newly created nodes, but
+   * newly created nodes are left unpartitioned until a node
+   * partitionining sweep is done later.
    *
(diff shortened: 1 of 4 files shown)


roystgnr:
Still passing Rattlesnake?  In that case I'll merge once the rest of the CI checkboxes are happy.

roystgnr:
Hmm.. or maybe use this excuse to add some more expensive optional MOOSE tests, since GRINS-dbg might take a while.

roystgnr:
DistributedMesh recover exodiff failures at 3 processors with variables/fe_hier.test_hier_2_1d, at 10 with mesh/named_entities.test_periodic_names, at 12 with restart/restart.test_nodal_var_2, and at 16 with executioners/executioner.test_steady, misc/exception.parallel_exception_jacobian_transient_non_zero_rank, and misc/exception.parallel_exception_residual_transient_non_zero_rank...

And I can't replicate a single one of those failures.

We still have those tests marked "failed but allowed" (for good reason; there was one long-standing failure there that I never managed to replicate) and this PR may not be what broke them (when's the last time we did a distributed recover pass on Civet?  The https://civet.inl.gov/recipe_events/23169/ log doesn't show any of the previous runs.) so I'm going to merge anyway, but I'm despairing trying to figure out how to bisect test failures I can't reproduce.

Issue #1661, opened 2018-04-19
Bug when calling EquationSystems::reinit with NodeElem and discontinuous variables


dknez:
1. Make or read in a mesh, and add a NodeElem to it. Make an EquationSystems object, es, based on the mesh.

2. Add a system to es that only has discontinuous basis functions, e.g. CONSTANT MONOMIALs, or FIRST L2_LAGRANGEs.

3. Do es.init(); followed by es.reinit(); On the call to reinit, I get the stack trace below. I gather the issue is that reinit fails because the NodeElem has no variables on it associated with the system added in point 2 (though it may have variables in other systems in es).

```
Assertion `var < this->n_vars(s)' failed.
var = 0
this->n_vars(s) = 0


Stack frames: 17
0: libMesh::print_trace(std::ostream&)
1: libMesh::MacroFunctions::report_error(char const*, int, char const*, char const*)
2: libMesh::DofObject::n_comp(unsigned int, unsigned int) const
3: libMesh::DofMap::_dof_indices(libMesh::Elem const&, int, std::__debug::vector<unsigned int, std::allocator<unsigned int> >&, unsigned int, libMesh::Node const* const*, unsigned int, unsigned long&) const
4: libMesh::DofMap::dof_indices(libMesh::Elem const*, std::__debug::vector<unsigned int, std::allocator<unsigned int> >&) const
5: libMesh::FEMContext::pre_fe_reinit(libMesh::System const&, libMesh::Elem const*)
6: libMesh::GenericProjector<libMesh::OldSolutionValue<double, &(void libMesh::FEMContext::point_value<double>(unsigned int, libMesh::Point const&, double&, double) const)>, libMesh::OldSolutionValue<libMesh::VectorValue<double>, &(void libMesh::FEMContext::point_gradient<libMesh::VectorValue<double> >(unsigned int, libMesh::Point const&, double&, double) const)>, double, libMesh::VectorSetAction<double> >::operator()(libMesh::StoredRange<libMesh::MeshBase::const_element_iterator, libMesh::Elem const*> const&) const
[...]

PR #1664, opened 2018-04-19
Unit test for Github Issue #1661.

dknez:


roystgnr:
Thanks!

And now I'm completely baffled!

The element that's failing an assert here doesn't have any variables on it, because the variables are on its *parent* element.  Why does it have a parent element, when we haven't actually made any refinement calls, you ask?  Very good question!  I'll try to figure that out next...

roystgnr:
Ha!  EquationSystems assumes you might have done some flagging, so it fixes up flag smoothness and then refines and coarsens if there are any flags there.

Flagging "smoothness" includes getting rid of "unrefined islands", defined as elements which don't have any neighbors who will be at a coarser level after the current refinement step.

Obviously this broken definition includes elements which don't have any neighbors period.

Testing the fix now.

dknez:
Looks good! Thanks for fixing this!

PR #1665, opened 2018-04-20
Add some more header only boost to allow getting a circular_buffer

friedmud:
Needed for an application I'm working on.  Tried to copy over as little as necessary...


friedmud:
Nevermind - this is entirely too slow for my purposes... just going to write my own....

PR #1669, opened 2018-04-21
Updated ExodusII_IO to skip element variable writing on blocks that are inactive

dknez:
We use the "truth table" feature of the ExodusII API to skip writing element variables on inactive blocks.

### DIFF
--- a/include/mesh/exodusII_io_helper.h
+++ b/include/mesh/exodusII_io_helper.h
@@ -298,7 +298,7 @@ public:
    * Sets up the nodal variables
    */
   void initialize_element_variables(std::vector<std::string> names,
-                                    std::vector<std::set<subdomain_id_type>> vars_active_subdomains);
+                                    const std::vector<std::set<subdomain_id_type>> & vars_active_subdomains);
 
   /**
    * Sets up the nodal variables
--- a/include/mesh/exodusII_io_helper.h
+++ b/include/mesh/exodusII_io_helper.h
@@ -297,7 +297,8 @@ public:
   /**
    * Sets up the nodal variables
    */
-  void initialize_element_variables(std::vector<std::string> names);
+  void initialize_element_variables(std::vector<std::string> names,
+                                    std::vector<std::set<subdomain_id_type>> vars_active_subdomains);
 
   /**
    * Sets up the nodal variables
@@ -317,7 +318,10 @@ public:
   /**
    * Writes the vector of values to the element variables.
    */
-  void write_element_values(const MeshBase & mesh, const std::vector<Real> & values, int timestep);
+  void write_element_values(const MeshBase & mesh,
+                            const std::vector<Real> & values,
+                            int timestep,
+                            std::vector<std::set<subdomain_id_type>> vars_active_subdomains);
 
   /**
    * Writes the vector of values to a nodal variable.
@@ -374,6 +378,13 @@ public:
    * starting with r_, i_ and a_ respectively.
    */
   std::vector<std::string> get_complex_names(const std::vector<std::string> & names) const;
+
+  /**
+   * returns a "tripled" copy of \p vars_active_subdomains, which is necessary in the
+   * complex-valued case.
+   */
+  std::vector<std::set<subdomain_id_type>> get_complex_vars_active_subdomains(
+    const std::vector<std::set<subdomain_id_type>> & vars_active_subdomains) const;
 
   /**
    * This is the \p ExodusII_IO_Helper Conversion class.  It provides
--- a/include/mesh/exodusII_io_helper.h
+++ b/include/mesh/exodusII_io_helper.h
@@ -321,7 +321,7 @@ public:
   void write_element_values(const MeshBase & mesh,
                             const std::vector<Real> & values,
                             int timestep,
-                            std::vector<std::set<subdomain_id_type>> vars_active_subdomains);
+                            const std::vector<std::set<subdomain_id_type>> & vars_active_subdomains);
 
   /**
    * Writes the vector of values to a nodal variable.
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -1878,7 +1878,7 @@ void ExodusII_IO_Helper::write_timestep(int timestep, Real time)
 void ExodusII_IO_Helper::write_element_values(const MeshBase & mesh,
                                               const std::vector<Real> & values,
                                               int timestep,
-                                              std::vector<std::set<subdomain_id_type>> vars_active_subdomains)
+                                              const std::vector<std::set<subdomain_id_type>> & vars_active_subdomains)
 {
   if ((_run_only_on_proc0) && (this->processor_id() != 0))
     return;
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -1887,6 +1887,7 @@ void ExodusII_IO_Helper::write_element_values(const MeshBase & mesh,
   // For each variable, create a 'data' array which holds all the elemental variable
   // values *for a given block* on this processor, then write that data vector to file
   // before moving onto the next block.
+  libmesh_assert_equal_to(vars_active_subdomains.size(), static_cast<unsigned>(num_elem_vars));
   for (unsigned int i=0; i<static_cast<unsigned>(num_elem_vars); ++i)
     {
       // The size of the subdomain map is the number of blocks.
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -2105,13 +2105,13 @@ std::vector<std::set<subdomain_id_type>> ExodusII_IO_Helper::get_complex_vars_ac
 {
   std::vector<std::set<subdomain_id_type>> complex_vars_active_subdomains;
 
-  for (auto it : vars_active_subdomains)
+  for (auto & s : vars_active_subdomains)
     {
       // Push back the same data three times to match the tripling of the variables
       // for the real, imag, and modulus for the complex-valued solution.
-      complex_vars_active_subdomains.push_back(it);
-      complex_vars_active_subdomains.push_back(it);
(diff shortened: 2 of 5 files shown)


include/mesh/exodusII_io_helper.h
```diff
@@ -297,7 +297,8 @@ class ExodusII_IO_Helper : public ParallelObject
   /**
    * Sets up the nodal variables
    */
-  void initialize_element_variables(std::vector<std::string> names);
+  void initialize_element_variables(std::vector<std::string> names,
+                                    std::vector<std::set<subdomain_id_type>> vars_active_subdomains);
```
roystgnr: IMHO this "helper" class isn't exactly a public API so we should be able to change it willy-nilly, but I'd like to hear others thoughts too.

In hindsight we should probably have enforced that by making the constructor private and making the user-friendly Exodus I/O class a friend...

dknez:
This works for me in all the cases I've tried, and it passes "make check". I'd like to get this passing the tests so that it can be merged, since I think this is an improvement compared to the previous approach of always writing all element variables to all "Exodus blocks," even in blocks where the variable is inactive.

Can someone with knowledge of the moose tests help me with the debugging for those tests?

roystgnr:
Both failures are in ExodusII::copy_elemental_solution(), which we don't have any libMesh-internal test coverage for, so it's probably a real failure.

Let me see if I can hack up a unit test that exposes that.

include/mesh/exodusII_io_helper.h
jwpeterson: `vars_active_subdomains` should be passed in by const reference, if I understand correctly it is only read from.

src/mesh/exodusII_io_helper.C
```diff
@@ -1730,18 +1731,21 @@ void ExodusII_IO_Helper::initialize_element_variables(std::vector<std::string> n
...
+      for(auto block_id : vars_active_subdomains[var_num])
+        {
+          unsigned int block_index =
+            std::distance(block_ids.begin(), std::find(block_ids.begin(), block_ids.end(), block_id));
+
+          unsigned int truth_tab_index = block_index*num_elem_vars + var_num;
+          truth_tab[truth_tab_index] = 1;
+        }
+    }
+
```
jwpeterson: Is this the backwards-incompatible change that is causing MOOSE tests to fail? We are now defaulting to all zeros in the `truth_tab`?

src/mesh/exodusII_io_helper.C
```diff
@@ -2067,6 +2082,28 @@ std::vector<std::string> ExodusII_IO_Helper::get_complex_names(const std::vector
...
 
+std::vector<std::set<subdomain_id_type>> ExodusII_IO_Helper::get_complex_vars_active_subdomains(
+  const std::vector<std::set<subdomain_id_type>> & vars_active_subdomains) const
+{
+  auto it = vars_active_subdomains.begin();
+  auto it_end = vars_active_subdomains.end();
+
+  std::vector<std::set<subdomain_id_type>> complex_vars_active_subdomains;
+
+  for (; it != it_end; ++it)
```
jwpeterson: Definitely use a range-based for loop here.

src/mesh/exodusII_io_helper.C
dknez: We currently default to "all ones" in the truth table, i.e. write out element data for all blocks.

This change should only affect cases where you have CONSTANT MONOMIAL variables which are not active on all subdomains, since otherwise the truth table should be "all ones" still, just like before. Is that the case in the MOOSE tests that are failing?

jwpeterson: Both tests which are failing are attempting to read in Exodus files written out by other tests. Both tests have CONSTANT, MONOMIAL auxiliary variables in them, but they aren't subdomain restricted as far as I can tell. Also the first parts of each test still pass, so presumably the outputs from those are identical to before?

roystgnr: Could something be going wrong with the writing?  I grabbed a copy of the mesh from the failing-in-Civet MOOSE test, and I'm reading it and copying a solution variable from it in a new unit test, and it's working fine.

I'll try rebuilding MOOSE and reproducing the failure and see if *that* output breaks the unit test.

roystgnr: Yeah, something's got to be an issue with the writing or with the reading in of this branch's output.  tests/restart/restart/elem_part1_out.e went from 86100 bytes to 66840 bytes.

And there aren't any subdomains in that MOOSE test!

dknez: OK, sounds like we're that we're closing in on the issue. Can you compare the ncdump of the two files (or send them to me so I can do it)? Perhaps it's not writing out some of the element variables now?

dknez: @roystgnr do you have any further info on this? Is there anything that I can run to reproduce the issue?

roystgnr: Sorry for the delay; I've been stuck on other work after getting what I *thought* should be a decent unit test case set up.  But my unit test doesn't even work on master:

```
  void testExodusCopyElementSolution () 
  {  
    {  
      Mesh mesh(*TestCommWorld); 
 
      EquationSystems es(mesh); 
      System &sys = es.add_system<System> ("SimpleSystem"); 
      sys.add_variable("e", CONSTANT, MONOMIAL); 
 
      MeshTools::Generation::build_square (mesh, 
                                           3, 3, 
                                           0., 1., 0., 1.); 
 
      es.init(); 
 
      ExodusII_IO exii(mesh); 
 
      exii.write_equation_systems("mesh_with_soln.e", es); 
    } 
 
    // copy_elemental_solution currently requires ReplicatedMesh 
    {  
      ReplicatedMesh mesh(*TestCommWorld); 
 
      EquationSystems es(mesh); 
      System &sys = es.add_system<System> ("SimpleSystem"); 
      sys.add_variable("teste", CONSTANT, MONOMIAL); 
 
      ExodusII_IO exii(mesh); 
 
      exii.read("mesh_with_soln.e"); 
      mesh.prepare_for_use(); 
 
      es.init(); 
 
      // Read the solution e on time step 6 into variable teste 
      exii.copy_elemental_solution(sys, "teste", "e"); 
    } 
  } 
```

roystgnr: I'll email you the working and failing versions of the MOOSE exodus output.

roystgnr: But boy, I'd really like to have a proper test for this API, because if neither this nor my previous two attempts at writing unit tests for it worked, something's wrong.  "Roy's an idiot" can't be ruled out, I'm afraid, but I'd say "the API is too brittle" and "the API is too poorly documented" are at least in the running.

dknez: I can replicate the error with this, thanks.

dknez: I pushed a fix. The issue was that I was supposed to interpret an empty list of active subdomains as "active on all subdomains".

Also, @roystgnr, note that you need to add "exii.write_element_data(es);" after "exii.write_equation_systems("mesh_with_soln.e", es);" to get your test case to work.

roystgnr: Ha!  I tried the one, I tried the other; I didn't try both.

dknez:
I think I now have a bug in the case with inactive subdomains... I'll need to do another push.

dknez:
I believe it's in good shape now. Let's see what the tests say.

src/mesh/exodusII_io_helper.C
dknez: P.S. You also need to change "exii.write_equation_systems("mesh_with_soln.e", es);" to:

```
    std::set<std::string> sys_list;
    exii.write_equation_systems("mesh_with_soln.e", es, &sys_list);
```
otherwise you try to write out the element variables twice.

dknez:
Looks good now. OK to "squash and merge"? Roy, could you add your unit test as well, perhaps after this PR is merged?

src/mesh/exodusII_io_helper.C
dknez: Yes, exactly. It writes out a nodal version of "e" in your test case, so you have to skip that in order to not clash with the element version that we write out when we call write_element_data.

src/mesh/exodusII_io_helper.C
```diff
@@ -2067,6 +2100,28 @@ std::vector<std::string> ExodusII_IO_Helper::get_complex_names(const std::vector
 
 
 
+std::vector<std::set<subdomain_id_type>> ExodusII_IO_Helper::get_complex_vars_active_subdomains(
+  const std::vector<std::set<subdomain_id_type>> & vars_active_subdomains) const
+{
+  auto it = vars_active_subdomains.begin();
```
jwpeterson: Drop these two lines as they are no longer needed

src/mesh/exodusII_io_helper.C
```diff
@@ -2067,6 +2100,28 @@ std::vector<std::string> ExodusII_IO_Helper::get_complex_names(const std::vector
...
+  auto it_end = vars_active_subdomains.end();
+
+  std::vector<std::set<subdomain_id_type>> complex_vars_active_subdomains;
+
+  for (auto it : vars_active_subdomains)
```
jwpeterson: Here I think we want `for (auto & s : vars_active_subdomains)` and `push_back(s)` below, just so there's no extra copying of the sets or confusion about whether the loop variable is an iterator.

src/mesh/exodusII_io_helper.C
jwpeterson: Looks good to me, thanks!

dknez:
@roystgnr is it also OK with you if we merge now? Then we can figure out what's going wrong in the new unit test in #1670.

roystgnr:
Let's wait for Civet, just to be paranoid, but then yeah, good to merge.

dknez:
OK, it's passed, so I'll merge now.

PR #1671, opened 2018-04-26
include/parallel/ header reorganization

roystgnr:
Finally getting around to the request from @friedmud that our Parallel:: code *not* be all lumped into a couple giant kitchen-sink headers.

This PR isn't done - there will be at least one rebase+force-push coming, plus several more additional commits, but it's lunch time and I'd like to hand what's done so far to Civet while I'm AFK.


roystgnr:
We're good up to packing.h!  Now to force-push and watch the passed test hash disappear...

friedmud:
@roystgnr this may be too far.  `post_wait_work.h` is pretty anemic.  Is there not something in-between "everything in one file" and "everything in its own file"?  Some sort of smart grouping?

If not - this is better than it was...

roystgnr:
MOOSE is actually what sold me on the "each class in its own file" paradigm!  I assume that's not a 100% accurate description of its organization, but when I was trying to get up to speed it was just so *nice* to look at the name of a class and immediately know what file it would be in, even if I was looking at a github page or something else without ctags/IDE style hyperlinks.

That said, I completely concede that PostWaitWork is a pretty extreme case.  I'd already been considering stuffing it and its within-libMesh subclasses all into the same file.  I'll do so if you prefer.

roystgnr:
I'll also be moving a few parallel_implementation.h methods to other existing files, so the Attributes and OpFunction and StandardType headers will get a bit more filled out.  I didn't think that was as high a priority, though: "kitchen-sink file for implementations" is annoying to other developers editing the Parallel:: code and to myself, whereas "kitchen sink file for declarations and documentation" is annoying to **everyone**.

friedmud:
It's not a hard and fast rule.  "Helper" objects can be "co-located" with
the objects they help with.

But yeah: we do try to keep things separate (and I'm sure we have a few
that are just as small as `post_wait_work.h`).  If you think having them
all separate is more consistent then I won't stand in your way!

roystgnr:
> If you think having them all separate is more consistent then I won't stand in your way!

Aside from consistency there's scattered other advantages:

With PostWaitWork, the abstract base class is in theory something that users might want to subclass to set up their own fancy communication paradigms, the way we have with container serialization and with packed ranges.  I'd rather pad that file out with documentation than just by throwing in a couple example subclasses.

With our "Helper" objects, like the OpFunction and Attributes and StandardType stuff, not only might users want to add partial specializations around their own data types, but they might need to make sure their partial specializations come *in between* the general declaration and any headers where uses occur... in https://github.com/libantioch/antioch the only sane way I found to get recursive use of template specializations working nicely was to break *everything* out into "declarations" and "implementations" headers, so you could be sure you had all of the former properly preceding all of the latter.

So that all being said, I think I'll leave this PR finely divided for now, but don't think of it as a decision set in stone; if we find the small headers too annoying later (or if anyone else wants to chime in and register annoyance now!) then we can try for more of a middle ground.

PR #1673, opened 2018-04-27
Get rid of --enable-default-comm-world option

roystgnr:
It's been 4 years since we made --enable-default-comm-world a legacy,
deprecated, warning-triggering, use-at-your-own-risk option; time to
remove it entirely.


roystgnr:
Ha!  GRINS doesn't actually require default COMM_WORLD, but it does depend on that macro being defined?  I'll put the macro back in case anyone else did the same thing, and we can update GRINS later.

PR #1674, opened 2018-04-27
Use MPI_IN_PLACE for reductions when available

roystgnr:
This ought to make sum/max/min of vectors a tiny bit faster.

See #1617 for general MPI optimization and "we still can't assume
everybody at least has MPI-2 already?!?" discussion

I'd like to test this against an MPI-1-only library before we merge; if nobody's Civet has such a module handy then we can wait until I build one myself.

### DIFF
--- a/include/parallel/parallel_algebra.h
+++ b/include/parallel/parallel_algebra.h
@@ -76,29 +76,7 @@ public:
 #ifdef LIBMESH_HAVE_MPI
         StandardType<T> T_type(&((*ex)(0)));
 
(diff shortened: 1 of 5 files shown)


roystgnr:
> I'd like to test this against an MPI-1-only library

And now I've done so, and it turns out that my last such test was inadvertently made against an MPI-2 library.  Our MPI-1 support has actually been broken for a year and a half.  Nobody's using it, so let's shave off a couple hundred lines of code by removing it.

PR #1675, opened 2018-05-01
Run bootstrap.

jwpeterson:
This should have been included with #1673.


roystgnr:
Yes it should; thanks!

PR #1677, opened 2018-05-03
Support dash or underscore in on_command_line, command_line_next

### LINKED ISSUE #1269: Support hyphenated versions of all long command line options
We are pretty consistent about this already, but there are a few non-conforming flags like `--node_major_dofs`, `--solver_system_names`, `--solver_variable_names`, `--implicit_neighbor_dofs`, `--read_ifem_systems`, etc. that use an underscore for the separator, which is a bit inconsistent.  Basically we should make all these like `--print-trace`, which already supports both underscore and hyphenated versions.

roystgnr:
This should resolve #1269.

Hopefully the new unit tests are enough to make sure that's true.  We don't seem to actually have any previous CI coverage of these options, and I don't know how often folks are currently running FIN-S with an updated libMesh.

### DIFF
--- a/include/base/libmesh.h
+++ b/include/base/libmesh.h
@@ -158,6 +158,9 @@ void enableSEGV(bool on);
  * then with all underscores changed to dashes, then with all dashes
  * (except any leading dashes) changed to underscores, and returns
  * true if any of the above finds a match.
+ *
+ * This routine manipulates the command_line cursor and should not be
+ * called concurrently with similar utilities in multiple threads.
  */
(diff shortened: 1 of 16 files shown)


roystgnr:
Actually, it looks like ```on_command_line()``` *is* the utility function we need, thanks to its side effects.

I'll try using it that way, and also document the non-thread-safe side effects.

roystgnr:
Looks like I jumped the gun; the XFEM fixes are in MOOSE origin/devel now, but not yet in origin/master...

Issue #1680, opened 2018-05-05
VTKIO reader: processor_id is invalid


ataber:
It looks like when I read a VTK file all the cells and their nodes have invalid processor ids. Furthermore, the neighbor relation is not properly filled out (easily fixed by doing `mesh.find_neighbors();`)

My code looks like:
```
  Mesh mesh(init.comm(), 3);
  VTKIO(mesh).read("input.vtu");
```

ataber:
It looks like this is fixed if I call `mesh.prepare_for_use();` after reading. Is there any reason not to put this call within the `VTKIO::read` function?

PR #1681, opened 2018-05-05
Add prepare_for_use call after VTK read.

ataber:
To fix [this issue](https://github.com/libMesh/libmesh/issues/1680)

### DIFF
--- a/src/mesh/vtk_io.C
+++ b/src/mesh/vtk_io.C
@@ -235,6 +235,8 @@ void VTKIO::read (const std::string & name)
                       << mesh.mesh_dimension()                  \
                       << "D support.");
 #endif // LIBMESH_DIM < 3
+  
+  mesh.prepare_for_use();
 }
 
 


dknez:
Calling prepare_for_use internally seems like a good idea to me. I've been caught by this before too.

pbauman:
I vote "yes" as well. I think if the lack of flexibility becomes as issue, I would think we could introduce a functor interface to allow the user to modify the mesh before `prepare_for_use()`.

jwpeterson:
I agree that in 90%+ of cases, one wants the mesh to be "prepared" automatically on read. A couple of other reasons we haven't done this in the past are:

1. It does parallel communication (which the user may want to postpone until a later time in the interest of overlapping computation and communication).
1. It renumbers nodes and elements by default, so if there was some information the user wished to preserve about the original numbering of the mesh, that becomes more difficult.
1. It passes default arguments to `prepare_for_use()` which the user may wish to override (this is partially solved now by the addition of the `MeshBase::allow_renumbering()` interface)
1. For `DistributedMesh`, remote elements are deleted on each processor.

I'd be in favor of `prepare_for_use()` by default, we would need to provide a way, either a `MeshInput` level flag or by adding default argument to `FooIO::read()`, similar to what is currently done in `MeshBase::read()`, in order for it to be disabled...

jwpeterson:
@ataber I send you an invite to join the libMesh "Associates" Team. That just allows your libmesh PRs to begin CI testing automatically.

roystgnr:
> I'd be in favor of prepare_for_use() by default, we would need to provide a way, either a MeshInput level flag or by adding default arguments to FooIO::read(), similar to what is currently done in MeshBase::read(), in order for it to be disabled...

This sounds good to me, if @permcody is happy with it too.  Then most users leave out "skip_prepare" and get the most user-friendly behavior, but MOOSE apps can manually avoid redundancy with a small change at the framework level.

permcody:
This sounds reasonable to me.

ataber:
As a novice user of this package, I can see why `prepare_for_use` isn't called automatically, but I did lose a couple hours tracking down the errors. I'm fine with either approach discussed above, but if the code does stay the same there should be some more visible documentation about the necessity of calling `prepare_for_use` after reading.

jwpeterson:

PR #1685, opened 2018-05-07
send, receive, non-blocking send for vectors of vectors

roystgnr:
The implementation on this needs to be redone before we merge, but it's passing unit tests so it should get @friedmud up and running.

### DIFF
--- a/include/parallel/parallel_implementation.h
+++ b/include/parallel/parallel_implementation.h
@@ -131,122 +131,9 @@ inline void send_receive_vec_of_vec(const unsigned int dest_processor_id,
       return;
     }
 
-  // temporary buffers - these will be sized in bytes
-  // and manipulated with MPI_Pack and friends
-  std::vector<char> sendbuf, recvbuf;
-
-  // figure out how many bytes we need to pack all the data
-  int packedsize=0, sendsize=0;
-
-  // The outer buffer size
-  libmesh_call_mpi
-    (MPI_Pack_size (1,
-                    libMesh::Parallel::StandardType<unsigned int>(),
-                    comm.get(),
-                    &packedsize));
-
-  sendsize += packedsize;
-
-  for (std::size_t i=0; i<send.size(); i++)
-    {
-      // The size of the ith inner buffer
-      libmesh_call_mpi
-        (MPI_Pack_size (1,
-                        libMesh::Parallel::StandardType<unsigned int>(),
-                        comm.get(),
-                        &packedsize));
-
-      sendsize += packedsize;
-
-      // The data for each inner buffer
-      libmesh_call_mpi
-        (MPI_Pack_size (libMesh::cast_int<int>(send[i].size()),
-                        libMesh::Parallel::StandardType<T1>
-                        (send[i].empty() ? libmesh_nullptr : &send[i][0]),
-                        comm.get(),
-                        &packedsize));
-
-      sendsize += packedsize;
-    }
-
-  libmesh_assert (sendsize /* should at least be 1! */);
-  sendbuf.resize (sendsize);
-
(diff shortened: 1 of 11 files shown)


include/parallel/parallel_implementation.h
```diff
@@ -1374,6 +1375,147 @@ inline void Communicator::send (const unsigned int dest_processor_id,
...
+
+  std::size_t datasize = 0;
+  for (auto & subvec : buf)
+    datasize += subvec.size();
+
+  const std::size_t fullsize = (headersize+datasize)*sizeof(T);
+
+  // Because of those evil casts we don't even want to trigger
+  // constructors/destructors
+  T * tempbuf = (T*)malloc(fullsize);
```
jwpeterson: Ugh, we're purposely putting malloc/free pairs into the library 😭

roystgnr: Yeah, I'm going to take that out.

roystgnr: Wouldn't have uploaded first except that Derek needs to use this for something and I had already nearly finished the dumb implementation before realizing what I should have done instead.

include/parallel/parallel_implementation.h
```diff
@@ -1374,6 +1375,147 @@ inline void Communicator::send (const unsigned int dest_processor_id,
...
+                                Request & req,
+                                const MessageTag & tag) const
+{
+  this->send(dest_processor_id, buf,
+             StandardType<T>((buf.empty() || buf.front().empty()) ?
+                             libmesh_nullptr : &(buf.front().front())), req, tag);
+}
+
+
+
```
jwpeterson: I know this is the "implementation" part of the file, but it would be helpful (at least for me) if you could put a comment at this point in the file saying which of these overloads are blocking and which are non-blocking.

include/parallel/parallel_implementation.h
jwpeterson: Oh, I see 👍

roystgnr: I am going to leave that PostWaitFreeBuffer class in, because I've known the pain of using a C library that returns heap allocated pointers to C++ code, but I don't intend to use it internally.

include/parallel/parallel_implementation.h
roystgnr: Will do.

friedmud:
Trying this - will let you know.

roystgnr:
We don't need it for testing purposes, assuming you're not doing hybrid MPI+threads right now.

We do need it in the final code for people who are running hybrid.

IMHO adding a template parameter for allocators is the right fix... although we should do it *everywhere* and it'll be a big diff.

roystgnr:
Well, wait.  I *think* we need it in the final code.  It's possible that the default C++ allocator is now thread-friendly, though, since C++11 officially recognizes that threads are a thing people use?  We should check on that.

friedmud:
@roystgnr currently getting some segfaults with these methods.

Investigating... if I can't see it quickly... I'll put up my branch so you can see what I'm doing.

friedmud:
Wait - might not be you :-)

roystgnr:
Well, let's not jump to conclusions.  The first version of this branch failed so badly it's not even funny.

I think I'm nearly done with the *right* implementation, too, in which case if you can't find anything obvious in your code then you'll maybe want to try the next commit on this branch...

roystgnr:
Ok, give this one a try.  It still passes unit tests in parallel for me.  That's not to say that Civet won't scream and die right off the bat, but it's at least a good start.

AFK for a while but I'll be back online to work on any problems tonight.

friedmud:
Testing now

PR #1686, opened 2018-05-07
Sparsity pattern parallel sync

friedmud:
Building off of #1685 .

Trying to undo some brain damage in sparsity pattern `parallel_sync`.

Unfortunately it's not quite working right now.  Any ideas @roystgnr ?

I'm actually seeing an assert fail in `dof_map.C`:

```
Assertion `requested->processor_id() == this->processor_id()' failed.
requested->processor_id() = 1
this->processor_id() = 0


[0] src/base/dof_map.C, line 397, compiled nodate at notime
```

It's weird - and seems unrelated... but it's also funny that that's another "round robin" communication spot that needs to be undone! 😄

### DIFF
--- a/src/base/sparsity_pattern.C
+++ b/src/base/sparsity_pattern.C
@@ -23,6 +23,8 @@
 #include "libmesh/elem.h"
 #include "libmesh/ghosting_functor.h"
 #include "libmesh/sparsity_pattern.h"
+#include "libmesh/communicator.h"
+#include "libmesh/parallel_algebra.h"
 
 
 namespace libMesh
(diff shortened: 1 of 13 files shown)


roystgnr:
I can actually get a segfault in the unit tests if I run them in parallel...

roystgnr:
And I can get that same dof_map.C:418 assertion failure on introduction_ex2 on 2 procs

roystgnr:
And if I run the unit tests in dbg or devel mode first (which ought to be the default ordering; I'll fix that soon like I did for the examples) then I get a different-but-still-"unrelated" assertion failure, this time Partitioner communication failing to match up.

src/base/sparsity_pattern.C
```diff
@@ -448,123 +448,164 @@ void Build::parallel_sync ()
...
-                  n_nz[my_r]++;
-            }
-          else
-            {
-              for (std::size_t j=0; j<their_row.size(); j++)
-                if ((their_row[j] < local_first_dof) || (their_row[j] >= local_end_dof))
-                  n_oz[my_r]++;
-                else
-                  n_nz[my_r]++;
+  std::vector<Parallel::Request> dof_sends(num_sends);
```
roystgnr: Should we check that num_sends is the same as std::accumulate of will_send_to?  I think this should be a symmetric communication but I wouldn't want to swear to it.

roystgnr: (by which I mean "of will_receive_from", after the alltoall)

roystgnr: No, it shouldn't be a symmetric communication, but that doesn't matter; your code looks like it should be doing the correct number of receives.

roystgnr:
I feel like we need a better version of parallel_object_only() to debug things like this.  parallel_object_only() will scream if you hit it on different lines in multiple processes out of sync, and it'll typically trigger clear MPI stack errors if you hit it on one process while another process is out of sync trying to do a different reduction, but it won't test for unresolved communication, which would be a nice sort of test to be able to do if that's possible...

friedmud:
Just a dumb logic error that was keeping the receives from happening....

permcody:
Did you mean to commit this with all those prints to std::cout?

permcody:
> Did you mean to commit this with all those prints to std::cout?

NVM - Github took me to the previous commit.

roystgnr:
Let's squash ba9bab9 onto affae9a before we merge?  Otherwise :+1: from me.  Thanks!

pbauman:
@bboutkov Just showed me a case in 3D with the projection matrix that is tripping PETSc malloc errors for 3D in DistributedMesh. He'll be posting case, but I wonder if it stemmed from this?

YaqiWang:
No. Just trying to think what is causing `MatMPIAIJSetPreallocation` in `PetscMatrix<T>::init() ` crash on my run. I am using 64-bit version of PETSc and libMesh, PETSc version is 3.8.4, libMesh is very recent after this fix. I start suspecting something could be wrong on petsc side. My memory usage is very low before the crash. Before the crash, I printed the size of the matrix passed in `MatSetSizes`, which are m_local=918144 n_local=918144 m_global=1671200640 n_global=1671200640. If I down size the problem a little, with the size as m_local=459072 n_local=459072 m_global=835600320 n_global=835600320, the problem runs fine. Any thoughts?

roystgnr:
Is @bboutkov using this SparsityPattern::build() method to preallocate the projection matrix?  I'm not actually clear at all on how that works.  In the unit tests we initialize proj_mat with global and local sizes, and then we pass it straight to System::projection_matrix() to fill without ever assigning sparsity information to it, and then somehow that works rather than screaming and dying but I don't see how.

YaqiWang:
I am running on HPC, bellow is what I got:
```
===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   PID 30279 RUNNING AT r5i3n14.ib0.ice.inl.gov
=   EXIT CODE: 11
=   CLEANING UP REMAINING PROCESSES
=   YOU CAN IGNORE THE BELOW CLEANUP MESSAGES
===================================================================================
```
It looks like a segfault. The problem is that I cannot get the backtrace, otherwise it would be much more helpful.

bboutkov:
> Is @bboutkov using this SparsityPattern::build() method to preallocate the projection matrix? 

Not explicitly, I follow the same paradigm as in the projection matrix unit tests and just prepare properly sized matrixes. Actually, I havent rebased my GMG code in a while so I think my problem is independent of this PR. I just created #1696 a few minutes too early; it will shortly contain more details about my specific situation.

roystgnr:
@YaqiWang, we have yet another report of sparsity pattern issues now, one where the failures go away by reverting this PR.  You might want to try that until we figure out what's going on.

YaqiWang:
Thanks @roystgnr . More of my runs indicates this PR could be an issue. I will keep a close eye on this.

YaqiWang:
Yes, currently running another case, will possibly get to the run in two hours. Will let you posted.

YaqiWang:
Sorry, it will take longer. Something happened on our machine and I lost my interactive job and need to regain one before I can test.

permcody:
I highly recommend you start using "screen" so that you don't lose your interactive jobs.

Issue #1679, opened 2018-05-04
XDR libraries not automatically discovered on Fedora 28


milljm:
Fedora 28 (and possibly other _newer_ distributions) are removing xdr libraries from the common location of /usr/include/rpc in favor of /usr/include/tirpc/rpc

Details of this change for Fedora: https://fedoraproject.org/wiki/Changes/SunRPCRemoval

An internet search for "sunrpc removal from glibc-headers" brings up a few distros where this is also happening.

This is causing the following error:
```pre
ERROR: Functionality is not available.
Make sure LIBMESH_HAVE_XDR is defined at build time
The XDR interface is not available in this installation
```

This is of course a user's issue to solve and not libMesh's... Just wanting to create an issue to 'FYI' the libMesh devs.

jwpeterson:
Hm... we've apparently never had a configure test for the XDR headers since they've been in the same place on every system we've ever used! Shouldn't be too hard to fix by adding one.

PR #1689, opened 2018-05-10
Improve/generalize configure test for XDR headers.

jwpeterson:
Fedora 28 only installs the rpc.h and xdr.h headers if you install the
"libtirpc-devel" RPM; this functionality is no longer built in to
glibc. The headers are also installed in a new location,
/usr/include/tirpc/rpc instead of the old location, /usr/include/rpc,
and you have to link against a library, -ltirpc, in order to use the
XDR functions.

For more information, see: https://fedoraproject.org/wiki/Changes/SunRPCRemoval

The original code was running a mix of AC_CHECK_FUNC and
AC_CHECK_HEADER checks, and it was a bit redundant, so I changed it to
a single AC_LINK_IFELSE test.  I also moved the XDR testing from
libmesh_core_features.m4 to libmesh_optional_packages.m4 since that
made it easier to optionally append to the libmesh_optional_INCLUDES
and libmesh_optional_LIBS variables.

We now no longer have separate defines for the rpc.h and xdr.h
headers, you either have both or you have neither, and this is
controlled by the LIBMESH_HAVE_XDR define.

Refs #1679.


roystgnr:
:+1:

PR #1690, opened 2018-05-11
More Parallel:: template arguments

roystgnr:
We should try to support whatever containers the users hand us, even ones with custom allocators and/or comparators.

This also lets us re-enable the scalable_allocator that was disabled by #1686.

This passes my initial tests but even if our standard CI is happy we'll want to at least make sure we run it through tests with MPI disabled too.


roystgnr:
The CI failure was almost certainly a false positive fixed by https://github.com/idaholab/moose/pull/11502, so I'm not going to waste CPU time repeating the tests here.  Assuming it passes this time in #1691 we'll just merge that PR and close this one.

Issue #1688, opened 2018-05-09
Regression in MOOSE XFEM module test


jwpeterson:
@roystgnr, we seem to have picked up a regression in the `init_solution_propagation` test in MOOSE's xfem module due to commit 9844a86206005266ed1bb690f8fc9aefca00b43c. To reproduce with master, build everything up to the xfem module and run
```
cd $MOOSE_DIR/modules/xfem/test/tests/init_solution_propagation
mpiexec -n 4 ../../../xfem-opt -i init_solution_propagation.i
```
This happened because we previously weren't running the MOOSE modules tests on libmesh PRs, although @brianmoose has subsequently rectified that situation.

Before this commit, the code runs fine, after this commit, the test code hangs with some processes in `MPI_Probe` as best as I can tell. Based on the comment in the offending commit, I'd say some logic error is causing us to be in an infinite loop... In other words, instead of quitting syncing "too soon" we are now never quitting 😄 

@bwspenc I'm not sure exactly what the test in question is doing, but if you could briefly describe it, it might shed some light on what the problem is.

jwpeterson:
The failures can also be seen (they appear as timeouts) over on idaholab/moose#11469.

bwspenc:
When we cut the mesh with XFEM, we have to be careful to copy the solution (current, old, and older as applicable) from the original nodes of the duplicated elements to the newly created nodes at those locations. This test is making sure that happens correctly both in serial and parallel.

bwspenc:
The code that is being tested is in `XFEM::initSolution()` in `moose/modules/xfem/src/base/XFEM.C`. If you comment out all the code in that method, my guess is that it won't hang (but you'll also break a lot of tests because they run incorrectly).

jwpeterson:
OK, thanks for the info @bwspenc!

roystgnr:
Thanks!  I'll try to replicate it this morning.

roystgnr:
This is easy enough to replicate, and it's definitely a libMesh bug.  The trouble is that the offending commit isn't actually a bug; it just triggers the bug.  I think I'll have a real fix soon.

roystgnr:
[...]

PR #1693, opened 2018-05-12
Modifying data read in from an exodus file

dknez:
Added a function that copies data to a new ExodusII_IO object and optionally modifies the data, and writes it out to disk.

The motivation for this is that I want to be able to change some specific values in an exodus file and write out the modified data, without having to create an EquationSystems object in order to do the data modification.

### DIFF
--- a/include/mesh/exodusII_io.h
+++ b/include/mesh/exodusII_io.h
@@ -199,6 +199,17 @@ public:
                                  const std::vector<std::string> &) libmesh_override;
 
   /**
+   * Copy data stored in exio_helper to \p new_exii_io, modify the data according to
+   * \p new_nodal_vals, and then use new_exii_io to write to disk.
+   *
+   * The map \p new_nodal_vals gives a map of index -> new nodal value for
+   * each (timestep,var_name) key.
+   */
+  void modify_and_write_nodal_data (ExodusII_IO & new_exii_io,
+                                    const std::string & fname,
+                                    const std::map<std::pair<int,std::string>, std::map<unsigned int,Real>> & new_nodal_vals);
+
+  /**
    * Write out a discontinuous nodal solution.
    */
   void write_nodal_data_discontinuous (const std::string &,
--- a/src/mesh/exodusII_io.C
+++ b/src/mesh/exodusII_io.C
@@ -791,6 +791,49 @@ void ExodusII_IO::write_nodal_data (const std::string & fname,
 
 
 
+void ExodusII_IO::modify_and_write_nodal_data (ExodusII_IO & new_exii_io,
+                                               const std::string & fname,
+                                               const std::map<std::pair<int,std::string>, std::map<unsigned int,Real>> & new_nodal_vals)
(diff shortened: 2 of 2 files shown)


dknez:
Is that MOOSE XFEM failure related to this PR? I can't see how this would be related, since this PR is just an extra function which should not change any existing behavior.

jwpeterson:
> Is that MOOSE XFEM failure related to this PR?

No, the reason for this failure is being discussed over in #1688. 

I don't have time right at this moment to look at this PR, but I would like to before it gets merged. Can you give me a day or two to look at it?

dknez:
Sure, no problem.

jwpeterson:
I have a few questions/concerns about this PR...

1. Is this really generally applicable enough that it should become part of the `ExodusII_IO` API? We were just lamenting [the other day](https://github.com/libMesh/libmesh/pull/1670) about how inconsistent the `ExodusII_IO` API is... this feels like one more tacked on thing.
1. Only the nodal vars are copied, any fundamental reason to omit elemental vars and global data?
1. It's not really clear to me from the documentation how to build up the fairly complicated `std::map<std::pair<int,std::string>, std::map<unsigned int,Real>>` data structure that must be passed in to the function.
1. This adds a capability to `ExodusII_IO` with no corresponding capability on the `Nemesis_IO` side... not strictly necessary for this PR to be merged, but I'd like to maintain more consistency between these two formats going forward.

dknez:
I agree with your points.

An alternative solution would be to give public access to the helper object so that we can move this kind of functionality into app codes. What do you think about that?

jwpeterson:
> An alternative solution would be to give public access to the helper object so that we can move this kind of functionality into app codes.

Yes I think we've been heading this way for quite some time, actually. People have wanted access to members on the helper at various times in the past, and instead of making the entire helper public, in the past we've just added more interfaces at the `ExodusII_IO` level. I think I'd be OK with this as long as we maintain the "Pimpl" idiom that is there currently, which allows us to avoid installing the Exodus library's internal headers.

jwpeterson:
> I was thinking of just adding a "getter" to ExodusII_IO to access the helper object. Would that be OK with you? If so, I'll make a PR for that.

Seems like the right way to get started, yes.

dknez:
@jwpeterson: I tried to add

```ExodusII_IO_Helper & get_exio_helper();```

to ExodusII_IO, but I can't use that from app codes because exodusII_io_helper.h is apparently not installed. Do you have any suggestion on what the best approach is here?

jwpeterson:
> Do you have any suggestion on what the best approach is here?

This is controlled by the `include/rebuild_include_HEADERS.sh` script, where that file is added to a blacklist. If you remove it from there and then follow the instructions for [adding/removing source files](https://github.com/libMesh/libmesh/wiki/Adding-or-removing-source-files), it should get added to the list of headers to install.

dknez:
Closing this, use #1695 instead.

PR #1695, opened 2018-05-15
Get exio helper

dknez:
In order to have full flexibility with ExodusII input/output, it can be useful to have access to the ExodusII_IO_helper object. This PR provides a "getter" for that object.

We also had to remove exodusII_io_helper.h from the noinst_blacklist so that it is accessible from app codes.

### DIFF
--- a/include/Makefile.in
+++ b/include/Makefile.in
@@ -572,7 +572,6 @@ DISTCLEANFILES = libmesh_config.h
 # details which are not required for the public interface
 noinst_HEADERS = \
         base/libmesh_augment_std_namespace.h \
-        mesh/exodusII_io_helper.h \
         mesh/nemesis_io_helper.h \
         numerics/laspack_matrix.h \
         numerics/laspack_vector.h \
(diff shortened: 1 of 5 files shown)


dknez:
@jwpeterson, is this OK with you to merge?

dknez:
I could add a comment to get_exio_helper in the header to explain that this gives access to a non-public API and may change in the future. Does that sound OK?

jwpeterson:
I think the "likely to see future API changes" clause of the `libmesh_experimental` macro is applicable here.

dknez:
So do you suggest a comment in the header, or a print statement, or calling a macro? I can update it with whatever you prefer.

roystgnr:
If John thinks it's not overkill and if you don't mind the stdout clutter, then let's go with libmesh_experimental().  It was exactly the intent of that macro to warn people about APIs in flux; my only qualm is that I hadn't pictured it being used on APIs which we expect to be in that state for such a *long* time.

jwpeterson:
`libmesh_experimental()` in the getter is fine, it's wrapped in `libmesh_do_once()` so you'll only see it once in the output.

dknez:
OK, done now. Sorry for the extra commits, we can squash if you like?

jwpeterson:
> squash if you like?

Just the last two commits, yeah. And we don't need to wait for testing.

Issue #1697, opened 2018-05-16
PETSc sparsity pattern error in libMesh examples in parallel


dknez:
If I run introduction_ex4 with "mpirun -np 2" I get a PETSc sparsity pattern error:

```
[0]PETSC ERROR: --------------------- Error Message --------------------------------------------------------------
[0]PETSC ERROR: Argument out of range
[0]PETSC ERROR: New nonzero at (0,409) caused a malloc
Use MatSetOption(A, MAT_NEW_NONZERO_ALLOCATION_ERR, PETSC_FALSE) to turn off this check
[0]PETSC ERROR: See http://www.mcs.anl.gov/petsc/documentation/faq.html for trouble shooting.
[0]PETSC ERROR: Petsc Release Version 3.8.0, unknown 
[0]PETSC ERROR: ./example-opt on a arch-linux2-c-opt named david-Lenovo by dknez Wed May 16 09:51:58 2018
[0]PETSC ERROR: Configure options --with-shared-libraries=1 --with-debugging=0 --download-suitesparse --download-blacs --download-ptscotch=yes --with-blaslapack-dir=/opt/intel/compilers_and_libraries_2018.1.163/linux/mkl --CXXFLAGS=-Wl,--no-as-needed --download-scalapack --download-mumps --download-metis --prefix=/home/dknez/software/libmesh_install/opt_real/petsc --download-hypre --download-ml
[0]PETSC ERROR: #1 MatSetValues_MPIAIJ() line 610 in /home/dknez/software/petsc-src/src/mat/impls/aij/mpi/mpiaij.c
[0]PETSC ERROR: #2 MatAssemblyEnd_MPIAIJ() line 712 in /home/dknez/software/petsc-src/src/mat/impls/aij/mpi/mpiaij.c
[0]PETSC ERROR: #3 MatAssemblyEnd() line 5213 in /home/dknez/software/petsc-src/src/mat/interface/matrix.c
```

I'm on the libMesh HEAD, and using PETSc 3.8. I'll git bisect to figure out what triggers this.

friedmud:
This could be related to the parallel_sync changes. I’m not sure if I’ll
have time to take a look today...

roystgnr:
I'll at least see if I can replicate and see if reverting #1686 has any effect.

dknez:
Reverting to c7933d3 works for me, which is the commit right before #1686, so I think #1686 is the issue.

jwpeterson:
[...]

PR #1698, opened 2018-05-16
Revert "Use MPI_Pack/MPI_Unpack for vec-of-vec..."

roystgnr:
This reverts commit 2d9377d37c149cab814d626b78d697b024b6295b, with
conflicts manually fixed.

With some good luck, this patch will fix the breakage @dknez ran into in #1697.  I'm submitting this PR before doing any testing myself so that he can get on that ASAP, since I still can't reproduce the problems he encountered.

With the luck I've been having, this patch won't even compile, this PR will trigger a Github server bug and set their datacenter on fire, and this very joke will cause some reader to have a fatal brain aneurysm.

### DIFF
--- a/include/parallel/parallel_implementation.h
+++ b/include/parallel/parallel_implementation.h
@@ -131,9 +131,122 @@ inline void send_receive_vec_of_vec(const unsigned int dest_processor_id,
       return;
     }
 
+  // temporary buffers - these will be sized in bytes
+  // and manipulated with MPI_Pack and friends
+  std::vector<char> sendbuf, recvbuf;
+
+  // figure out how many bytes we need to pack all the data
+  int packedsize=0, sendsize=0;
+
+  // The outer buffer size
+  libmesh_call_mpi
+    (MPI_Pack_size (1,
+                    libMesh::Parallel::StandardType<unsigned int>(),
+                    comm.get(),
+                    &packedsize));
+
+  sendsize += packedsize;
+
+  for (std::size_t i=0; i<send.size(); i++)
+    {
+      // The size of the ith inner buffer
+      libmesh_call_mpi
+        (MPI_Pack_size (1,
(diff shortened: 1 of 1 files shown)


dknez:
introduction_ex4 now works in parallel for me with this PR, great!

I will run our test suite on this. That will take a couple of hours, though.

roystgnr:
@pbauman, is femputer down?  It's not showing up in the CI box here and I can't seem to connect to the webserver.

pbauman:
Crap. The web page is not pulling up for me, but the machine is up and the server is running. Let me try rebooting.

roystgnr:
I tried digging into git history for the send_receive_vec_of_vec method I broke, and it turns out that the tricky code I was trying and failing to refactor was written in 2bf2691 - by Ben, in 2008.  Any other MPI_Pack/MPI_Unpack experts in the crowd?

Unless I ever manage to figure out how to reproduce the error in #1697, I think I'll probably just leave this hackish alternative in place indefinitely.  Its disadvantages (it won't work on heterogenous architectures that nobody ever uses) are greatly outweighed by its advantages (it probably won't pass CI but then break horribly for other users).

roystgnr:
> Unless I ever manage to figure out how to reproduce the error in #1697

It turns out to be as simple as "use OpenMPI".  I still think we should merge this PR as soon as GRINS-dbg and David's test suite are both happy, though; just because I can trigger the error doesn't mean I'll be able to fix it quickly.

dknez:
This PR fixes the failures I was seeing before, so it's ready to merge as far as I'm concerned.

roystgnr:
Actually, it looks like the real fix may be nearly trivial.  Would you mind testing that too in a minute?

dknez:
Sure

roystgnr:
I'm still running tests on it, but it looks like #1699 will be a much safer fix.

PR #1701, opened 2018-05-17
Add missing return in switch statement

roystgnr:
Previously all the many many people who use VTK output of infinite
quads would hit errors!

This bug dates back to commit 4859f74, a refactoring in July 2004.

Hopefully this record for "longest-unfixed libMesh bug ever" will
never be beaten.


roystgnr:
> Hopefully this record for "longest-unfixed libMesh bug ever" will
never be beaten.

In the cold light of day that sentence looks ominously ambiguous.

I do mean that hopefully no other bugs will go unfixed for so long, not that hopefully no other long-unfixed bugs will get fixed.  We do plan to get to them all *eventually*, I swear!

PR #1702, opened 2018-05-17
New parallel_sync.h algorithms for parallel_ghost_sync.h backend

roystgnr:
The upcoming goal here is to get rid of every single O(N_proc^2) round-robin communication in the whole library, but getting rid of every round-robin in the parallel_ghost_sync.h algorithms actually covers about half of our big communication operations, which is a good start.

The underlying algorithm here is a slight improvement over the one in #1600, and that was a 4-fold speedup for me on the affected sections of code on 320 processors, so I've got a good feeling about this.  I'll want some empirical confirmation or contradiction of that, though.  I'll be running my own tests tomorrow, but if @jwpeterson wants to cook up graphs as useful as the ones in #1598 I'd appreciate it.  More worrisome is that @bboutkov saw contradictory results from #1600, so I'd very much like to find out whether this PR is also a pessimization rather an optimization for him.  At least we now have our communication strategy in one place, so if there turns out to still be a performance issue we can fix it once rather than over and over and over again.

Most importantly, after the #1697 debacle, I'd also like it if such massive refactorings of our communications get tested as widely as possible.  This PR is passing tests so far on my standard configuration, and I'll try the basic libMesh test suite with openmpi myself this time, but @dknez might want to double-check with his applications too.

### DIFF
--- a/include/parallel/parallel_sync.h
+++ b/include/parallel/parallel_sync.h
@@ -441,13 +441,15 @@ void pull_parallel_vector_data(const Communicator & comm,
     response_data, received_data;
   std::vector<Request> response_reqs;
 
+  StandardType<datum> datatype;
+
   auto gather_functor =
-    [&comm, &gather_data, &response_data, &response_reqs]
+    [&comm, &gather_data, &response_data, &response_reqs, &datatype]
     (processor_id_type pid, query_type query)
     {
       Request sendreq;
       gather_data(pid, query, response_data[pid]);
-      comm.send(pid, response_data[pid], sendreq);
+      comm.send(pid, response_data[pid], datatype, sendreq);
       response_reqs.push_back(sendreq);
     };
 
@@ -464,7 +466,7 @@ void pull_parallel_vector_data(const Communicator & comm,
       Request req;
       auto & incoming_data = received_data[proc_id];
       incoming_data.resize(querydata.size());
-      comm.receive(proc_id, incoming_data, req);
+      comm.receive(proc_id, incoming_data, datatype, req);
       receive_reqs.push_back(req);
       receive_procids.push_back(proc_id);
     }
--- a/include/parallel/parallel_sync.h
+++ b/include/parallel/parallel_sync.h
@@ -25,6 +25,7 @@
 
 // C++ includes
 #include <map>
+#include <type_traits>
 #include <vector>
 
 
@@ -192,13 +193,31 @@ void push_parallel_vector_data(const Communicator & comm,
   // give it a good name
   auto & will_receive_from = will_send_to;
 
+  // This function only works for "flat" data that we can pre-size
+  // receive buffers for: a map to vectors-of-standard-types, not e.g.
+  // vectors-of-vectors.
+  //
+  // Trying to instantiate a StandardType<T> gives us a compiler error
+  // where otherwise we would have had a runtime error.
+  //
+  // Creating a StandardType<T> manually also saves our APIs from
+  // having to do a bunch of automatic creations later.
+  //
+  // This object will be free'd before all non-blocking communications
+  // complete, but the MPI standard for MPI_Type_free specifies "Any
+  // communication that is currently using this datatype will
+  // complete normally." so we're cool.
+  typedef decltype(data.begin()->second.front()) ref_type;
+  typedef typename std::remove_reference<ref_type>::type nonref_type;
+  StandardType<typename std::remove_const<nonref_type>::type> datatype;
+
   // Post all of the sends, non-blocking
   for (auto & datapair : data)
     {
       processor_id_type destid = datapair.first;
       auto & datum = datapair.second;
       Request sendreq;
-      comm.send(destid, datum, sendreq);
+      comm.send(destid, datum, datatype, sendreq);
       reqs.insert(reqs.end(), sendreq);
     }
 
@@ -212,7 +231,7 @@ void push_parallel_vector_data(const Communicator & comm,
         Request req;
         auto & incoming_data = received_data[proc_id];
         incoming_data.resize(will_receive_from[proc_id]);
-        comm.receive(proc_id, incoming_data, req);
+        comm.receive(proc_id, incoming_data, datatype, req);
         receive_reqs.push_back(req);
         receive_procids.push_back(proc_id);
       }
--- a/include/parallel/parallel_sync.h
+++ b/include/parallel/parallel_sync.h
@@ -1,0 +1,331 @@
+// The libMesh Finite Element Library.
+// Copyright (C) 2002-2018 Benjamin S. Kirk, John W. Peterson, Roy H. Stogner
+
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+
+// You should have received a copy of the GNU Lesser General Public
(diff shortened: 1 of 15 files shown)


roystgnr:
Marking as "do not merge" for now - even assuming all the CI passes, we'll want to look at performance etc. before letting this into master.

permcody:
FYI @roystgnr - Our TestHarness capability continues to grow and we are routinely using our new PBS scheduler module to run the test suite. This ticket spawned a discussion with the team about enabling large job testing so that we could do something like "min_cpus = 500" or perhaps "pbs_chunks = 14" to test these kinds of changes for select problems that we hand curate. If we tie this into our benchmarking capability (http://mooseframework.org/docs/timing/), we might be able to get qualitative information about the impacts of these types of changes without having to ask individuals to try out some new change. Your thoughts or ideas are welcome.

include/parallel/standard_type.h
```diff
@@ -27,6 +27,7 @@
 
 // C++ includes
 #include <numeric>
+#include <tuple>
```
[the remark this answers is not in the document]
roystgnr: Actually, what's annoying here is that this PR will *not* be the first use in the library, because that specialization doesn't yet get used.  I misread one of the chunks of code I intended to refactor, I concluded *erroneously* that I'd need tuple communication support for it right away, and I wrote up the Parallel tuple support before realizing my mistake.

If you'd like I can remove the tuple commit and save it for a PR where it actually gets exercised.

src/partitioning/parmetis_partitioner.C
```diff
@@ -203,12 +203,12 @@ struct SyncLocalIDs
...
 
   void gather_data (const std::vector<dof_id_type> & ids,
-                    std::vector<datum> & local_ids)
+                    std::vector<datum> & local_ids) const
   {
     local_ids.resize(ids.size());
 
     for (std::size_t i=0, imax = ids.size(); i != imax; ++i)
-      local_ids[i] = id_map[ids[i]];
+      local_ids[i] = id_map.at(ids[i]);
```
jwpeterson: If you switch to `map::at()`, you might want to think about `try`/`catch` block to go with it... this code will now throw instead of silently setting a default value for `local_ids[i]` in case `ids[i]` is not found for some reason.

roystgnr: It's intended to throw, actually!  Silently setting a default value (which actually happened to me once in a previous even-buggier iteration of this refactoring) is a horrible result which leads to tricky bugs triggered within ParMETIS code later.  I could add a try / catch-then-libmesh_error block if you like, to make any internal errors even easier to debug.  But this is a spot where any error would truly be *internal*, not triggerable by an error in user code.

roystgnr:
@permcody - automated benchmarking is definitely a hard problem, but manual benchmarking is even harder, so anything you can do to enable the former would be greatly appreciated.  We wouldn't want to make it part of regular CI, but being able to select optional "single-processor performance", "parallel performance", etc. tests from Civet would be wonderful!

src/partitioning/parmetis_partitioner.C
jwpeterson: I like catch-then-libmesh error personally, the more local the error information, the better.

roystgnr:
I just finished a fault analysis on this.  A tricky analysis of the bug reveals the proximate cause, "char is good at holding true/false values, but you have to change it to size_t when you decide to use it for vector lengths instead".  A simple analysis of the proximate cause reveals the root cause, "Roy is an idiot".

pbauman:
OK, in theory femputer is back up. That failure is a compilation failure in libMesh. I'll try and pay attention to femputer to make sure it stays up.

roystgnr:
And a header test failure too?  AFK right now but will fix both this afternoon.

roystgnr:
gcc/6.1 gives the same error.  Now, as long as the fix doesn't break other compilers...

roystgnr:
Ok, I've figured out the last bug.  Just haven't figured out the best fix...

pbauman:
@bboutkov, @roystgnr mentioned to me that it would be good if you could give this a whirl and see how it does.

roystgnr:
To be specific - the remaining bug is *only* triggered by Nemesis reads.  It'll take me a little while to fix, but since you're doing your tests with refinement from a coarse mesh and not using Nemesis you don't need to wait for the fix to be done.

friedmud:
@roystgnr hah!  I didn't see this PR earlier... and I wrote a bunch of similar parallel send/receive patterns here:

https://github.com/friedmud/moose/blob/parmetis_distributed_generated_mesh_11485/framework/src/mesh/DistributedGeneratedMesh.C#L1270
https://github.com/friedmud/moose/blob/parmetis_distributed_generated_mesh_11485/framework/src/mesh/DistributedGeneratedMesh.C#L1432

While I was writing those I was thinking of how to generalize that algorithm - and was just getting ready to implement a general version similar to what you have here... but you beat me to it!

Let me try to reimplement my stuff using this new code (best way to 👍 or 👎 a new set of code is to try to use it!)

include/parallel/parallel_sync.h
```diff
@@ -0,0 +1,341 @@
...
+      libmesh_assert_less(destid, num_procs);
+      auto & datum = datapair.second;
+      Request sendreq;
+      comm.send(destid, datum, sendreq);
+      reqs.insert(reqs.end(), sendreq);
+    }
+
+  // Post all of the receives, non-blocking
+  std::vector<Request> receive_reqs;
+  std::vector<processor_id_type> receive_procids;
```
friedmud: nit-picky - you could process `will_receive_from` to know how many requests and proc_ids you'll be receiving so you can `reserve()` space for the `push_back()`s

roystgnr:
@friedmud, hold off on "try to reimplement".  The remaining bug here is only triggered by Nemesis reads not because it's a Nemesis-specific bug but because Nemesis is the only parallel_ghost_sync.h use of vector<vector>.  SparsityPattern::parallel_sync will hit the exact same bug.

I've got the fix half finished (and tested enough to know at least that half works) but I've got family stuff in 5 minutes, and this afternoon, so it will probably be tonight before I get the full fix pushed.

roystgnr:
There'll also be an API change to make the fix easier: instead of ```pull_parallel_vector_data<foo>(args)``` it'll be ```pull_parallel_vector_data(args, (foo*)nullptr)```.

roystgnr:
Ok, this passes everything for me now, and it's got additional safeguards to make it impossible to compile a bug of the same variety in the future.  Unless I screwed up the rebase afterward I bet Civet should be happy with it too.  That "tonight" turned out to be overly pessimistic.

roystgnr:
Ok, I'm happy with this and we can merge once we have some performance data.

I'm going to start replacing round-robin calls with parallel_sync.h invocations, but I'll make a separate branch for that and save it for a later PR.

roystgnr:
Actually, let's revisit "happy" after some of the expensive tests have all passed too.

roystgnr:
This is seeing extensive use in my branch in #1708 without causing any failures, and @bboutkov reports his stuff running ~12% faster (total runtime, not just communication, and just this branch, not all the communication optimization it enables in #1708), so we ought to be ready to merge here ... after Civet confirms that I didn't screw anything up when I fixed those merge conflicts today.

#1708 still has a couple more commits I've yet to finish and push, so there'll still be a little more delay there.

roystgnr:
Yeah, I screwed up the merge conflict fix.  Just because part of a merge is clean doesn't mean it's correct...

roystgnr:
Huh, or maybe I didn't.  The problem seems to come from #1715 - @fdkong moved the ParmetisPartitioner::build_graph() declaration outside the ifdef LIBMESH_HAVE_PARMETIS, but didn't move the definition, so configurations without ParMETIS no longer build.

I'll fix that in a separate PR.

roystgnr:
I've already rebased this on top of #1722, on the theory that we should resume CI tests right away because my fix there is correct and complete and will be merged unchanged.

(Remember, kids, don't just ignore your non-STEM classes; concepts like "hubris" and "nemesis" have real-world value too!)

PR #1706, opened 2018-05-19
Add ASCII art axes markers indicating reference element coordinates.

jwpeterson:
I had to refresh my memory of the orientation used for Hexes, so I
added ASCII art axes for the other elements as well.

Note: Hex8::side_nodes_map (see cell_hex8.C) is consistent with the
sideset names assigned in MeshTools::Generation::build_cube(), which
are:

x: left/right
y: bottom/top
z: back/front

From the ASCII art, you might conclude that bottom/top would refer to
the z-direction on the reference element, but it actually refers to
the y-direction.

### DIFF
--- a/include/geom/cell_hex20.h
+++ b/include/geom/cell_hex20.h
@@ -39,11 +39,11 @@ namespace libMesh
  *          o    :                        o    |
  *         /     :                       /     |
  *        /    15o                      /    14o
- *       /       :                     /       |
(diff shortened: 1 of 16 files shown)


roystgnr:
I'm told kids these days put stuff like this into interactive WebGL on cloud-hosted Jupyter notebooks, but back in *my* day ASCII art was all we had and we liked it!

PR #1708, opened 2018-05-21
Replace more round robin communications with parallel_sync

roystgnr:
This branches off of and can wait on #1702; only the last five commits are new.

The first two tweak parallel_sync and add more assertions to its implementations; the final three replace three more round-robin communication processes with invocations of ```Parallel::pull_parallel_vector_data()```


roystgnr:
This is complete, and @bboutkov reported that at least the first 80% of the refactoring was giving him a ~5% speedup (on top of #1702), so I'll probably merge tomorrow unless Civet or my own tests or any other developers begin screaming.

roystgnr:
That external app failure is clearly a false positive.  The no-MPI failure is me being an idiot.  I can replicate it and will be testing the fix shortly.

PR #1710, opened 2018-05-23
Added a DenseMatrix function to compute outer product of 2 vectors

joshuahansel:
This adds a method to `DenseMatrix` to compute the outer product of 2 vectors and store it in the matrix object.

### DIFF
--- a/include/numerics/dense_matrix.h
+++ b/include/numerics/dense_matrix.h
@@ -166,6 +166,25 @@ public:
   void get_principal_submatrix (unsigned int sub_m, DenseMatrix<T> & dest) const;
 
   /**
+   * Computes the outer (dyadic) product of two vectors and stores in (*this).
+   *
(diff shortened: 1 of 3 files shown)


jwpeterson:
One thing that I just read is that the [outer product of complex-valued vectors typically uses the conjugate transpose of the second vector](https://en.wikipedia.org/wiki/Outer_product). This would be pretty easy to implement using the `libmesh_conj()` function which is defined in libmesh_common.h, and then your code would give the expected results for real- and complex-valued builds.

roystgnr:
Great, thanks!

:+1: once Civet is done.

joshuahansel:
@jwpeterson Do you want the complex-value changes on this PR or a subsequent one?

jwpeterson:
> Do you want the complex-value changes on this PR

This PR would be good. It's fine with me if you want to skip adding a complex-valued unit test specifically for the time being...

PR #1717, opened 2018-06-04
Fix compound preprocessor if-statement in eigenproblems_ex3.

jwpeterson:
Otherwise, we get

warning: extra tokens at end of #ifdef directive [-Wextra-tokens]


jwpeterson:
> I can't believe this just gave a warning, not an error...

Yes, we do have an optional SLEPc test recipe (which I've attached) but it definitely doesn't have `-Werror` turned on so this was never caught.

PR #1721, opened 2018-06-05
Forward declare enumerations instead of including headers

jwpeterson:
The forward enumerations are currently enabled by default, but for backwards compatibility, I have also added the `--disable-forward-declare-enums` configure option, which restores the previous enum inclusions and will give downstream apps time to fix up their includes on their own schedule. MOOSE, MOOSE-based apps, and GRINS have already been updated, and even in these relatively large codebases I didn't have to add many new includes.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -4655,6 +4655,7 @@ datarootdir = @datarootdir@
 docdir = @docdir@
 dvidir = @dvidir@
 enabledeprecated = @enabledeprecated@
+enablefwdenums = @enablefwdenums@
 enablelegacyincludepaths = @enablelegacyincludepaths@
 enablepetsc = @enablepetsc@
 enableuniqueptr = @enableuniqueptr@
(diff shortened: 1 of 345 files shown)


jwpeterson:
@pbauman GRINS compiles OK on my system, so most likely I missed updating some optional packages that are being built by your CI. I am unable to see the errors (usual timeout issue for me with femputer) so I'm not sure exactly what the error is...

jwpeterson:
@brianmoose Any chance there is something wrong with the "Test Trilinos" recipe? Possibly an environment variable is not set, as it tries to create a directory at the filesystem root:
```
/bin/mkdir -p '/include/libmesh'
/bin/mkdir: cannot create directory '/include': Permission denied
```

pbauman:
Just landed in DC, will check it out in a couple of hours

brianmoose:
@jwpeterson Yeah, `LIBMESH_DIR` apparently was not getting set. Fixed and now installs as expected.

jwpeterson:
> Just landed in DC, will check it out in a couple of hours

I think the problem was on the libmesh side, so no worries.

Issue #1714, opened 2018-06-01
Build dual graph in partitioner


fdkong:
There are some functions, `build_graph` and `assign_partitioning`, in `parmetis_paritioner` . The functions can be used in other partitioners as well. We refactor the code, and move the part unrelated to parmetis up to base class, `partitioner`.

roystgnr:
Technical question: ```ParmetisPartitioner::build_graph()``` outputs into ```ParmetisHelper::xadj``` and ```ParmetisHelper::adjncy``` vectors.  ```ParmetisPartitioner::assign_partitioning()``` uses input from the ```ParmetisHelper::part``` vector.  Are those all in formats which are usable by other partitioners?  Otherwise at first glance "the part unrelated to parmetis" is the null set.

Meta question: are you guys preparing an interface to a better parallel partitioner?  That would be *awesome*, let me know if there's anything other than asking nitpicky questions which I can do to help.

jwpeterson:
> Meta question: are you guys preparing an interface to a better parallel partitioner? That would be awesome, let me know if there's anything other than asking nitpicky questions which I can do to help.

Yes @fdkong is writing an interface that will allow us to tie into all partitioners supported by PETSc.

fdkong:
We are adding PETSc interface to PTSCOTCH, and it looks much better than parmetis.

PR will be very soon

fdkong:
parmetis:
![parmetis](https://user-images.githubusercontent.com/10508374/40851989-3fbef108-6586-11e8-86af-88f4e1995486.jpg)

ptscotch:

![ptscotch](https://user-images.githubusercontent.com/10508374/40852007-4f096198-6586-11e8-9f64-0d0984a51d6a.jpg)

pbauman:
Just a quick followup: will it be easy to update libMesh to support PETSc's partitioner interface or is that really an application space thing? Would be nice to be able to suck in the PTScotch partitioning.

jwpeterson:
I believe that @fdkong decided to add this support at the MOOSE level. You can check out what he has done at the MOOSE level over in https://github.com/idaholab/moose/pull/11629.

PR #1722, opened 2018-06-05
ParmetisPartitioner::build_graph needs ParMETIS

roystgnr:
This should hopefully fix a bug from #1714 that's causing the new No-MPI build failure in #1702


jwpeterson:
:+1: I also need to incorporate this into idaholab/moose#11565

jwpeterson:
This passes the No MPI test, which was all we were really testing.

PR #1725, opened 2018-06-06
Delete trailing whitespace from source, headers, and .tex

brianmoose:


jwpeterson:
Also passes the new Precheck, so I'm merging as-is.

PR #1729, opened 2018-06-07
Add enum header to rb_scm_construction

roystgnr:
This fixes a regression (presumably due to #1721 not being tested with
SLEPc?) in my configuration.


jwpeterson:
I did test SLEPc, but not the 
```
#if defined(LIBMESH_HAVE_SLEPC) && (LIBMESH_HAVE_GLPK)
```
combination, unforutnately.

Issue #1730, opened 2018-06-07
Problems configuring with PGI


RonRahaman:
I've got some problems configuring libmesh with new-ish versions of the PGI compilers (v17.10) and was wondering if you had any pointers.  

## Compiler identification 
I think I've got this fixed (see below).  Looks like the issue is that, with the `-V` flag, newer versions call themselves the "PGI Compilers and Tools"  instead of "Portland Group Something Something".   
``` Diff
--- a/m4/compiler.m4
+++ b/m4/compiler.m4
@@ -274,7 +274,7 @@ AC_DEFUN([DETERMINE_CXX_BRAND],
   dnl Portland Group C++?
   AS_IF([test "x$compiler_brand_detected" = "xno"],
         [
-          is_pgcc="`($CXX -V 2>&1) | grep 'Portland Group'`"
+          is_pgcc="`($CXX -V 2>&1) | grep 'Portland Group\|PGI'`"
           AS_IF([test "x$is_pgcc" != "x"],
           [
             AC_MSG_RESULT(<<< C++ compiler is Portland Group C++ >>>)
```

## Enabling C++ 11 for conftests
I'm stuck on this issue.  The conftest for `unordered_multimap` fails because I need to add the `-std=c++11` flag.  However, if i set `CXXFLAGS='-std=c++11`, it doesn't seem to get passed along to the conftest.  See the error from config.log below.  Note the compile line doesn't include the `-std` flag.  
```
configure:29427: checking whether the compiler supports std::unordered_multimap
configure:29455: mpicxx -c -g --no_using_std     conftest.cpp >&5
"/usr/include/c++/4.8.5/bits/c++0x_warning.h", line 32: catastrophic error:
          #error directive: This file requires compiler and library support for
          the ISO C++ 2011 standard. This support is currently experimental,
          and must be enabled with the -std=c++11 or -std=gnu++11 compiler
          options.
  #error This file requires compiler and library support for the \
   ^

1 catastrophic error detected in the compilation of "conftest.cpp".
Compilation terminated.
```

jwpeterson:
The configure tests should definitely detect and use the `-std=c++11` flag, I'm not sure why it's missing for you... it would probably be helpful if you could send me your entire config.log file.

I will also say that we have been unable to use the Portland Group C++ compiler for many years, so it would be nice if it is finally working!

The change to compiler.m4 looks like a good one.

RonRahaman:
[...]

PR #1731, opened 2018-06-08
Fixing build system for PGI compilers

RonRahaman:
This PR includes fixes for recent PGI compilers.  It follows-up discussions from Issue #1730.


m4/libmesh_method.m4
```diff
@@ -40,17 +40,17 @@ AC_DEFUN([LIBMESH_SET_METHODS],
  build_oprof=no
 
  dnl define compiler flags for all methods
- CPPFLAGS_OPT="$CPPFLAGS_OPT $libmesh_CPPFLAGS"
- CXXFLAGS_OPT="$CXXFLAGS_OPT $libmesh_CXXFLAGS"
- CFLAGS_OPT="$CFLAGS_OPT $libmesh_CFLAGS"
+ CPPFLAGS_OPT="$CPPFLAGS_OPT $libmesh_CPPFLAGS $CPPFLAGS"
+ CXXFLAGS_OPT="$CXXFLAGS_OPT $libmesh_CXXFLAGS $CXXFLAGS"
```
jwpeterson: In the past, we have always purposely avoided appending the `CXXFLAGS` environment variable to the list of flags used by libmesh, under the assumption that the user's `CXXFLAGS` might not work/be correct for libmesh... it could be time to revisit this decision, however. Anyone who is familiar with autoconf-based build systems is going to be used to using `CXXFLAGS` for this purpose, and it's more likely we are just confusing them than preventing errors in dubious environments...

jwpeterson:
Two comments:
1. In order for your m4 file changes to take effect you'll have to run bootstrap (with the same version of the autotools we are currently using) and check in the results. There are some instructions for doing this [here](https://github.com/libMesh/libmesh/wiki/Updating-configure-tests%2C-Makefiles%2C-etc.).
1. Starting your PR title with "WIP" means that the CI testing won't run, but you may already be aware of this... regardless I've sent you an invite to join the developers group whose PRs are tested automatically.

m4/libmesh_method.m4
roystgnr: https://sourceforge.net/p/libmesh/mailman/message/8105557/ - we made that change more than a decade ago, and now I don't recall why.

I'd personally be in favor of changing it back, but if so let's do it carefully; I think GRINS and probably other packages set libmesh_CXXFLAGS etc. and we'll need to give everyone time to get the non-prefixed counterparts consistent.

RonRahaman: Yes, I agree it's a pretty big change.  It was meant to solve a particular issue where the `-std=c++11` flag was auto-detected for PGI but was not included in some conftests.  The issue came from this series of events:

1. `-std=c++11` is corretly auto-detected and added to `CXXFLAGS`
https://github.com/libMesh/libmesh/blob/c2501bfbb26e7da832aa9fbea63ff7b27b1b3a39/m4/ax_cxx_compile_stdcxx_11.m4#L124-L150
2. In `libmesh_method.m4`, `CXXFLAGS_DEVEL` is set such that it does *not* include `CXXFLAGS`.  Hence, `CXXFLAGS_DEVEL` does not include `-std=c++11`.
  https://github.com/libMesh/libmesh/blob/c2501bfbb26e7da832aa9fbea63ff7b27b1b3a39/m4/libmesh_method.m4#L52
3. Finally, in `libmesh_compiler_features.m4`, the conftests are run *after* `CXXFLAGS="$CXXFLAGS_DEVEL"`.  This means that, for these conftests, `CXXFLAGS` will not include `-std=c++11`.  At this point, conftests for C++11 features will fail.  
  https://github.com/libMesh/libmesh/blob/c2501bfbb26e7da832aa9fbea63ff7b27b1b3a39/m4/libmesh_compiler_features.m4#L12

RonRahaman:
@jwpeterson: 

1. I've been been testing my changes after bootstrapping locally.  Based on my understanding [this guideline](https://github.com/libMesh/libmesh/wiki/Updating-configure-tests%2C-Makefiles%2C-etc.#commit-your-changes), I wasn't going to commit the new configure script until we'd settled on all the fixes.
2. Nope!  I actually didn't know that "WIP" prevented CI testing.  I work mostly in GitLab, where "WIP" prevents the PR from being merged but still lets the CI run.

Shall I commit the new configure script (and other bootstrapped files) and remove "WIP" from the PR title?

m4/libmesh_method.m4
RonRahaman: Here's a possible alternative to adding `CXXFLAGS` to the libmesh flags.   You might be able to add `-std=c++11` to `libmesh_CXXFLAGS` instead of (or in addition to) `CXXFLAGS`.  That should ensure it gets used where necessary.

permcody:
Yeah, take the WIP off. Rest assured we won’t merge until all tests pass
and decisions have been made.

jwpeterson:
> Shall I commit the new configure script (and other bootstrapped files) and remove "WIP" from the PR title?

Yes, absolutely. The guideline you mentioned mainly applies to confining bootstrap-generated changes to a separate commit from the others you've made. Definitely also do this before removing WIP so that your configure changes will actually be tested by the CI.

m4/libmesh_method.m4
jwpeterson: I'm still not sure I understand how `-std=c++11` does not make it into the tests. From what I can tell, there are 4 places where `CXXFLAGS_DEVEL` is changed:

1. configure.ac, line 174: AX_CXX_COMPILE_STDCXX_11() (ax_cxx_compile_stdcxx_11.m4)
2. configure.ac, line 182: LIBMESH_SET_CXX_FLAGS                    (compiler.m4)
3. configure.ac, line 195: LIBMESH_SET_METHODS                      (libmesh_method.m4)
4. configure.ac, line 343: LIBMESH_COMPILER_FEATURES                (libmesh_compiler_features.m4)

At each step, the following happens:

1. Note: `$switch` here is equal to `-std=c++11`
```
CXXFLAGS_DEVEL="$CXXFLAGS_DEVEL $switch"
```

2. GCC-specific flags (for example) are appended
```
CXXFLAGS_DEVEL="$CXXFLAGS_DEVEL -O2 -felide-constructors -g -pedantic -W -Wall -Wextra -Wno-long-long -Wunused"
CXXFLAGS_DEVEL="$CXXFLAGS_DEVEL -Wpointer-arith -Wformat -Wparentheses -Wuninitialized -funroll-loops -fstrict-aliasing -Woverloaded-virtual -Wdisabled-optimization"
```

3. User-defined flags get appended:
```
CXXFLAGS_DEVEL="$CXXFLAGS_DEVEL $libmesh_CXXFLAGS"
```

4. We temporarily set the `CXXFLAGS` for the `ACX_BEST_UNORDERED_MULTIMAP` and friends tests. 
```
CXXFLAGS="$CXXFLAGS_DEVEL"
```

Since `-std=c++11` was appended in step 1 and the `CXXFLAGS_DEVEL` was subsequently only appended to, it should be getting passed to the unordered tests in step 4. Indeed, on my system's config.log I see:

```
configure:29362: checking whether the compiler supports std::unordered_multimap
configure:29390: mpicxx -c  -std=gnu++11 -O2 -felide-constructors -g -pedantic -W -Wall -Wextra -Wno-long-long -Wunused-parameter -Wunused -Wpointer-arith -Wformat -Wparentheses -Wuninitialized -Qunused-arguments -Woverloaded-virtual -fno-limit-debug-info          conftest.cpp >&5
```
As you can see, the `-std=gnu++11` flag is the first one there, as expected.

I'm not sure what in this series of events is breaking for you, and unfortunately, dropbox is blocked for us here, so I haven't had a chance to look at your config.log file yet. I meant to do it from home this weekend, but forgot. Will try and look again tonight.

RonRahaman: How should I point you to my config.log?  I wanted to post it here somehow, but I'd be happy to just email you.

jwpeterson: Email would be fine.

jwpeterson:
> Email would be fine.

So I think what happened is that the CXXFLAGS you specified to configure:

```
../configure ... CXXFLAGS=-std=c++11
```

was used by the basic C++11 test, and caused it to conclude that *no* flags were required:
```
configure:6772: checking whether mpicxx supports C++11 features by default
configure:6808: mpicxx -c -std=c++11  conftest.cpp >&5
configure:6808: $? = 0
configure:6815: result: yes
```

Therefore, we never actually entered this if-statement where $switch is set for all methods:
```
        AS_IF([test "x$enablecxx11" = "xyes"],
              [
                CXXFLAGS="$CXXFLAGS $switch"
                CXXFLAGS_OPT="$CXXFLAGS_OPT $switch"
                CXXFLAGS_DEVEL="$CXXFLAGS_DEVEL $switch"
                CXXFLAGS_DBG="$CXXFLAGS_DBG $switch"
              ])
```
and therefore `CXXFLAGS_DEVEL` does not contain `-std=c++11` at this point. The initial C++11 tests subsequently pass because they still use the contents of `CXXFLAGS`, but the unordered tests do not (they use CXXFLAGS_DEVEL), and therefore all fail because they use the CXXFLAGS_DEVEL variable for their flags.

The simplest fix to your problem should be to remove `CXXFLAGS=-std=c++11` from your configure line. I think everything will just work correctly then, but I need to think about how to prevent this problem from happening in the future...

RonRahaman:
Makes sense.  But I realized I sent the wrong `config.log`.  I meant to send a log from a build where I didn't set `CXXFLAGS`.  I'm sending you another log called `config.58a0faa.no_user_CXXFLAGS.out`.  That build just has the fix from 58a0faa (for the compiler identification) and does *not* have `CXXFLAGS` set in the configure command.  Could you take a look at that, too?

RonRahaman:
I have a more lightweight and equally-effective solution in PR #1745.  May I close this one?

PR #1732, opened 2018-06-08
 Be more explicit about some defaulted/deleted functions

jwpeterson:
These changes currently only affect the MeshBase hierarchy and the
ExactSolution class, but eventually every class in the library should
be more specific about which of the 5 "special functions" (copy/move
ctor, copy/move assignment operator, destructor) are deleted or
defaulted.

### DIFF
--- a/configure
+++ b/configure
@@ -901,8 +901,6 @@ HAVE_DOUBLE_UNDERSCORE_ATTRIBUTE_FALLTHROUGH_FALSE
 HAVE_DOUBLE_UNDERSCORE_ATTRIBUTE_FALLTHROUGH_TRUE
 HAVE_CXX17_FALLTHROUGH_ATTRIBUTE_FALSE
 HAVE_CXX17_FALLTHROUGH_ATTRIBUTE_TRUE
-HAVE_CXX11_NOEXCEPT_FALSE
-HAVE_CXX11_NOEXCEPT_TRUE
 HAVE_CXX11_TO_STRING_FALSE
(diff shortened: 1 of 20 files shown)


pbauman:
Please forgive my C++11-best-practices ignorance. The `default` makes sense to me, but why is specifying `delete` better than just making those functions private?

Nilooab:
I am just beginner! Unfortunately I do not get what do you mean. maybe if
you give me more explanation, It may underestand.

jwpeterson:
It's true that the end result is basically the same, but IMO the compiler diagnostic is a bit cleaner in the `delete`d function case, something along the lines of "attempt to call deleted function" instead of a sometimes really big block of text about "is private within this context". There's also the fact that private+unimplemented was always just a hack to achieve the deleted function effect, and it's always nice to remove a hack with an explicitly supported feature.

permcody:
I can think of two reasons: The first is because `private` often isn't good enough. It means that those methods could still be (accidentally) called from member functions, and if you really don't want that, `delete` is absolute. This comes up quite a bit in self referential class types (`Elem` and `Node`) where you have vectors of neighbors and other similar structures. I made a class last year and caught a few bugs when I explicitly deleted the copy constructor and found a section of code that wouldn't compile. The second is consistency and clarity. It's nice that since we now have 5 methods that we "should" deal with in creating new classes we can knock several of them out with simple keywords without having to use different side effects to get the behavior that we want.

PR #1733, opened 2018-06-09
Make QBase-derived classes explicitly default copy/move constructible…

jwpeterson:
… and assignable.

For these super simple classes, there is no reason they should not
be. One small change required was to make the protected _dim and
_order members non-const.  I don't think there's a compelling reason
for these to be const, other than we did not really understand all the
ramifications of making things const when this class was written.

### DIFF
--- a/include/quadrature/quadrature.h
+++ b/include/quadrature/quadrature.h
@@ -67,15 +67,20 @@ protected:
    * Constructor. Protected to prevent instantiation of this base
    * class.  Use the build() method instead.
    */
-  QBase (const unsigned int _dim,
-         const Order _order=INVALID_ORDER);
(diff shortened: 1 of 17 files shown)


dschwen:
I think this introduced a compile warning:
```
./include/libmesh/quadrature_jacobi.h:74:13: warning: explicitly defaulted copy assignment operator is implicitly deleted [-Wdefaulted-function-deleted]
```

jwpeterson:
Huh, yeah these parameters don't need to be const, the compiler is correct. I wonder why no compiler complained about this before... did you recently switch to a new one?

dschwen:
Yeah, clang 8.0.0

Issue #1709, opened 2018-05-21
MessageTag copies should be attached to Requests using them


roystgnr:
We haven't actually triggered this issue yet, but with parallel communications that return Request objects for later completion it's now a risk.  The current lifetime of a unique tag ends when the MessageTag object is destructed, so if we create a local MessageTag, use it for non-blocking Parallel method, then exit the scope without a wait() on all corresponding Request(s), it is possible for a subsequent MessageTag creation to allocate the same tag number.

Non-blocking Parallel:: methods should probably copy the MessageTag they're given and only delete the copy when they're done with a wait(), so the reference counting will let the Communicator know not to reuse the same tag value.

We should also have a unit test that makes sure this is true.

friedmud:
Nice catch!  Hadn’t thought of that.

In my own code I create the tags in a global scope actually (or at least class scope) so they live during the whole execution of the program.

roystgnr:
I can't take too much credit; I investigated this while tracking down a refactoring bug.  Tag reuse wasn't the cause of that bug, but I looked into it closely enough to realize that we were bug-free in this respect only by luck and timidity, not by design.

roystgnr:
Resolved by #1741

PR #1741, opened 2018-06-14
Request should hold on to MessageTag where appropriate

roystgnr:
This should fix the (theoretical, hopefully not yet triggered) bug possibility opened up by #1709

### DIFF
--- a/include/Makefile.in
+++ b/include/Makefile.in
@@ -855,6 +855,7 @@ include_HEADERS = \
         parallel/parallel_sync.h \
         parallel/post_wait_copy_buffer.h \
         parallel/post_wait_delete_buffer.h \
+        parallel/post_wait_dereference_tag.h \
         parallel/post_wait_free_buffer.h \
         parallel/post_wait_unpack_buffer.h \
         parallel/post_wait_work.h \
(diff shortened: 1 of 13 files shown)


roystgnr:
Well, this was fortuitous.  This PR doesn't have a bug, it *exposes* a bug in Parallel::waitany()... and alerts me to a second bug in Parallel::waitany()...

pbauman:
Job [GRINS-devel](https://femputer.eng.buffalo.edu:8000/job/1740/) on 5d647ee : canceled by @roystgnr

Hanging

pbauman:
Job [GRINS-dbg](https://femputer.eng.buffalo.edu:8000/job/1741/) on 5d647ee : canceled by @roystgnr

Hanging

roystgnr:
Well, that backfired.  In system_io.C we request a tag for some non-blocking sends, then we *let that tag object destruct*, then we try to request a tag with the same suggested number again later.  Before, when we let those numbers expire along with the MessageTag, they were available again for the re-request.  Now, when we let the Request hold on to them, they aren't.

IMHO the best way to solve this is to give MessageTag longer lifetimes, but other users may have done the same thing elsewhere.  After this is passing libMesh CI I'll throw all the app tests we can at it and try to make sure we won't be biting anyone else.  If we are then I'll just pick out the bugfix commit for a separate PR and put this on the back burner.

PR #1743, opened 2018-06-15
Add configure test for determining -march settings

jwpeterson:
If you configure with `--enable-march`, the proper `-march=cpu-type` flag will be added to the optimized and devel mode flags based on your system's CPU. The script does not try to autodetect the CPU when cross-compiling; in such cases, the user can specify `--with-gcc-arch=cpu-type` to compile for a specific architecture. These m4 scripts were pulled from the autoconf-archive without changes.


m4/ax_gcc_archflag.m4
```diff
@@ -0,0 +1,265 @@
...
+            *1?6c?:*:*:*|*2?6[[67]]?:*:*:*|*3?6[[56]]?:*:*:*) ax_gcc_arch="bonnell atom core2 pentium-m pentium3 pentiumpro" ;;
+            *3?67?:*:*:*|*[[45]]?6[[ad]]?:*:*:*) ax_gcc_arch="silvermont atom core2 pentium-m pentium3 pentiumpro" ;;
+            *000?f[[012]]?:*:*:*|?f[[012]]?:*:*:*|f[[012]]?:*:*:*) ax_gcc_arch="pentium4 pentiumpro" ;;
+            *000?f[[346]]?:*:*:*|?f[[346]]?:*:*:*|f[[346]]?:*:*:*) ax_gcc_arch="nocona prescott pentium4 pentiumpro" ;;
+            # fallback
+            *5??:*:*:*) ax_gcc_arch=pentium ;;
+            *??6??:*:*:*) ax_gcc_arch="core2 pentiumpro" ;;
+            *6??:*:*:*) ax_gcc_arch=pentiumpro ;;
+            *00??f??:*:*:*|??f??:*:*:*|?f??:*:*:*|f??:*:*:*) ax_gcc_arch="pentium4 pentiumpro" ;;
+          esac ;;
```
permcody: Holy cow this is archaic! This script will require regular updates but I suppose that's OK. Turns out you can directly get this with clang but if this works for gcc/clang that'll be better.

```
llc --version
```

m4/ax_gcc_archflag.m4
```diff
@@ -0,0 +1,265 @@
...
+#   the flag is set to -mtune (or equivalent) so that the architecture is
+#   only used for tuning, but the instruction set used is still portable. In
+#   the latter case, the flag is set to -march (or equivalent) so that
+#   architecture-specific instructions are enabled.
+#
+#   The user can specify --with-gcc-arch=<arch> in order to override the
+#   macro's choice of architecture, or --without-gcc-arch to disable this.
+#
+#   When cross-compiling, or if $CC is not gcc, then ACTION-FAILURE is
+#   called unless the user specified --with-gcc-arch manually.
```
permcody: I'd recommend dropping "gcc" from the `--with-gcc-arch` this option. The `-march` flag works just as well with clang compilers and probably others as well. It'll likely cause confusion with people thinking they can only use this option, or it only works with GCC.

jwpeterson:
> On Jun 15, 2018, at 8:20 AM, Cody Permann <notifications@github.com> wrote:
> 
> @permcody commented on this pull request.
> 
> In m4/ax_gcc_archflag.m4:
> 
> > +            *1?6[[aef]]?:*:*:*|*2?6e?:*:*:*) ax_gcc_arch="nehalem corei7 core2 pentium-m pentium3 pentiumpro" ;;
> +            *2?6[[5cf]]?:*:*:*) ax_gcc_arch="westmere corei7 core2 pentium-m pentium3 pentiumpro" ;;
> +            *2?6[[ad]]?:*:*:*) ax_gcc_arch="sandybridge corei7-avx corei7 core2 pentium-m pentium3 pentiumpro" ;;
> +            *3?6[[ae]]?:*:*:*) ax_gcc_arch="ivybridge core-avx-i corei7-avx corei7 core2 pentium-m pentium3 pentiumpro" ;;
> +            *3?6[[cf]]?:*:*:*|*4?6[[56]]?:*:*:*) ax_gcc_arch="haswell core-avx2 core-avx-i corei7-avx corei7 core2 pentium-m pentium3 pentiumpro" ;;
> +            *3?6d?:*:*:*|*4?6[[7f]]?:*:*:*|*5?66?:*:*:*) ax_gcc_arch="broadwell core-avx2 core-avx-i corei7-avx corei7 core2 pentium-m pentium3 pentiumpro" ;;
> +            *1?6c?:*:*:*|*2?6[[67]]?:*:*:*|*3?6[[56]]?:*:*:*) ax_gcc_arch="bonnell atom core2 pentium-m pentium3 pentiumpro" ;;
> +            *3?67?:*:*:*|*[[45]]?6[[ad]]?:*:*:*) ax_gcc_arch="silvermont atom core2 pentium-m pentium3 pentiumpro" ;;
> +            *000?f[[012]]?:*:*:*|?f[[012]]?:*:*:*|f[[012]]?:*:*:*) ax_gcc_arch="pentium4 pentiumpro" ;;
> +            *000?f[[346]]?:*:*:*|?f[[346]]?:*:*:*|f[[346]]?:*:*:*) ax_gcc_arch="nocona prescott pentium4 pentiumpro" ;;
> +            # fallback
> +            *5??:*:*:*) ax_gcc_arch=pentium ;;
> +            *??6??:*:*:*) ax_gcc_arch="core2 pentiumpro" ;;
> +            *6??:*:*:*) ax_gcc_arch=pentiumpro ;;
> +            *00??f??:*:*:*|??f??:*:*:*|?f??:*:*:*|f??:*:*:*) ax_gcc_arch="pentium4 pentiumpro" ;;
> +          esac ;;
> Holy cow this is archaic! This script will require regular updates but I suppose that's OK. Turns out you can directly get this with clang but if this works for gcc/clang that'll be better.
> 
> llc --version
> In m4/ax_gcc_archflag.m4:
> 
> > +#   found, the cache variable $ax_cv_gcc_archflag is set to this flag and
> +#   ACTION-SUCCESS is executed; otherwise $ax_cv_gcc_archflag is set to
> +#   "unknown" and ACTION-FAILURE is executed. The default ACTION-SUCCESS is
> +#   to add $ax_cv_gcc_archflag to the end of $CFLAGS.
> +#
> +#   PORTABLE? should be either [yes] (default) or [no]. In the former case,
> +#   the flag is set to -mtune (or equivalent) so that the architecture is
> +#   only used for tuning, but the instruction set used is still portable. In
> +#   the latter case, the flag is set to -march (or equivalent) so that
> +#   architecture-specific instructions are enabled.
> +#
> +#   The user can specify --with-gcc-arch=<arch> in order to override the
> +#   macro's choice of architecture, or --without-gcc-arch to disable this.
> +#
> +#   When cross-compiling, or if $CC is not gcc, then ACTION-FAILURE is
> +#   called unless the user specified --with-gcc-arch manually.
> I'd recommend dropping "gcc" from the --with-gcc-arch this option. The -march flag works just as well with clang compilers and probably others as well. It'll likely cause confusion with people thinking they can only use this option, or it only works with GCC.
> 

I’m not changing the script just for that, I don’t want to get out of sync with upstream.

m4/ax_gcc_archflag.m4
jwpeterson: It’s not “archaic,” that implies something outdated. It’s just been around for a while and has support for lots of different processors.

permcody: ok fair

permcody:
ok

PR #1745, opened 2018-06-15
PGI compiler identification and flags

RonRahaman:
This addresses problems with the PGI compiler from Issue #1730.  It is a much more lightweight solution than PR #1731.  

There are still some linker issues that I discovered in libtool itself.  After I reported the issue, PGI said they'd follow-up with the libtool team ([link](https://www.pgroup.com/userforum/viewtopic.php?p=23789)).  In particular, libtool doesn't identify the PGI compiler correctly when it's discovering the PIC flags.  I'd like to wait for libtool to fix that, instead of hacking something into the libtool now.  For now, I can complete `configure` by manually specifying `-fpic` in CFLAGS, etc.

### DIFF
--- a/m4/compiler.m4
+++ b/m4/compiler.m4
@@ -532,15 +532,15 @@ AC_DEFUN([LIBMESH_SET_CXX_FLAGS],
                        ],
 
             [portland_group], [
-                                CXXFLAGS_DBG="-g --no_using_std"
-                                CXXFLAGS_OPT="-O2 --no_using_std -fast -Minform=severe"
+                                CXXFLAGS_DBG="$CXXFLAGS_DBG -g --no_using_std"
+                                CXXFLAGS_OPT="$CXXFLAGS_OPT -O2 --no_using_std -fast -Minform=severe"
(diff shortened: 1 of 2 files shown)


roystgnr:
This is good.  I'd like to split the re-bootstrap in ee2c5c5 into a separate commit from the substantive changes, though.

m4/compiler.m4
```diff
@@ -532,15 +532,15 @@ AC_DEFUN([LIBMESH_SET_CXX_FLAGS],
                        ],
 
             [portland_group], [
-                                CXXFLAGS_DBG="-g --no_using_std"
-                                CXXFLAGS_OPT="-O2 --no_using_std -fast -Minform=severe"
+                                CXXFLAGS_DBG="$CXXFLAGS_DBG -g --no_using_std"
```
jwpeterson: Ah, so we weren't correctly appending flags in the PGI case, we were just setting them. Good catch!

RonRahaman:
Gotcha.  I made a separate commit for the `bootstrap` in PR #1746.  Third times the charm!  :blush:

PR #1748, opened 2018-06-18
Add tag to receive_packed_range call

roystgnr:
We had this tag on the send but if we want to use it to avoid race
conditions then we need to use it on the *receive* too.

This would fix a bug from #1708


roystgnr:
Looks like GRINS tests are hanging, probably because some earlier CI run today got bit by the same bug you did.  If @pbauman is around to kick it then I'd say wait for them but if not then let's merge this as soon as MOOSE is happy.  A libMesh test that used to trigger the race condition every few runs for me is now running dozens of times with no trouble.

PR #1749, opened 2018-06-18
Detect partitioners available in PETSc, set defines

jwpeterson:
cc: @fdkong


pbauman:
Job [GRINS-devel](https://femputer.eng.buffalo.edu:8000/job/1760/) on 2a190e2 : invalidated by @jwpeterson

Invalidating as I don&#39;t see how a GRINS test failure could be related to this configure change.

pbauman:
Thanks @jwpeterson. You're correct it wasn't. It's testing heisenbug/race condition I thought I'd gotten rid of a long time ago. Apparently I didn't...

fdkong:
Thanks,  @jwpeterson. I will take a try.

PR #1750, opened 2018-06-19
NumericVector::operator/= argument should be const

roystgnr:
We never modify it.  Perhaps at some point we had a shim to a
non-const-correct numerics library?  But even if that were the case we
should have used const_cast rather than inheriting the incorrectness.

Thanks to @dmcdougall for pointing this out.


dmcdougall:
> Thanks to @dmcdougall for pointing this out.

So what you're saying is I should complain more often?  :)

PR #1751, opened 2018-06-19
Node processor Id assignment

fdkong:
This is not ready to merge.  I pushed the code here to ask some suggestions from @roystgnr  and @jwpeterson. Where different node assignment algorithms should live. 

BTW, different algorithms could generate totally different nodes partitions.

### DIFF
--- a/src/partitioning/partitioner.C
+++ b/src/partitioning/partitioner.C
@@ -522,6 +522,10 @@ void Partitioner::set_node_processor_ids_on_interface_using_queue(MeshBase & mes
     {
       unsigned int n_own_nodes = pmap.second.size()/2;
 
+      // Initialize node assignment
+      for (auto it = pmap.second.begin(); it != pmap.second.end(); it++)
+        mesh.node_ref(*it).processor_id() = pmap.first.second;
+
       visted_nodes.clear();
       for (auto it = pmap.second.begin(); it != pmap.second.end(); it++)
       {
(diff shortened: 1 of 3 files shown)


fdkong:
Even_odd scheme:

node_balance



 BFS scheme:

node_balance_queue

roystgnr:
What's BFS stand for?

jwpeterson:
> What's BFS stand for?

Badass F*ckin Scheme

fdkong:
> What's BFS stand for?

breadth first search

fdkong:
Third scheme: apply a partitioner to the interface meshes. The preconditioner based on this algorithm is competitive with the default `min` scheme.

src/partitioning/partitioner.C
```diff
@@ -405,6 +405,306 @@ void Partitioner::set_parent_processor_ids(MeshBase & mesh)
...
+        }
+
+      }
+     queue_done:
+      for (auto node: visted_nodes)
+        mesh.node_ref(node).processor_id() = pmap.first.first;
+    }
+}
+
+#include "petscmat.h"
```
jwpeterson: The `std::complex` test fails because of this line (PETSc is disabled in that build). This seems like an odd place to include a PETSc header... it should be wrapped in ifdefs at the top of the file instead.

src/partitioning/partitioner.C
```diff
@@ -405,6 +405,306 @@ void Partitioner::set_parent_processor_ids(MeshBase & mesh)
...
+    {
+      auto & node = mesh.node_ref(*it);
+      if (indices[i])
+        node.processor_id() = pmap.first.second;
+      else
+        node.processor_id() = pmap.first.first;
+    }
+    ISRestoreIndices(is, &indices);
+    ISDestroy(&is);
+#endif
```
jwpeterson: This entire function should be wrapped in `#ifdef LIBMESH_HAVE_PETSC` and should throw a `libmesh_error_msg()` in the `#else` branch.

jwpeterson:
> @fdkong I think you still need to push the active_element_ptr_range() -> active_local_element_ptr_range() fix that we figured out yesterday before this can be merged...

OK, I spoke with @fdkong locally and he fixed that issue in a slightly different way.

jwpeterson:
I will just fix some small code formatting things and then merge by hand.

PR #1756, opened 2018-06-22
Add support for writing CONSTANT, MONOMIAL variables in Nemesis

jwpeterson:
This is passing the simple regression test I created while developing it, and once it has been merged into into MOOSE upstream I will also add more tests for parallel elemental variable writing to the MOOSE test suite.


friedmud:
Yassss! 👍

PR #1763, opened 2018-06-25
Detect when PETSc is built with ParMETIS.

### LINKED ISSUE #1762: Add Parmetis to the list of partitioners we detect during configure
We now need to know whether the Parmetis we are using comes from PETSc or from the one in contrib.

Refs #1749

jwpeterson:
Fixes #1762.

cc: @fdkong


fdkong:
Thanks

PR #1765, opened 2018-06-26
Fixed a bug when using vinewtonssls and vinewtonrsls

### LINKED ISSUE #1752: Regression with --use-petsc-dm
This was reported by Ata Mesgarnejad last month but I didn't find time to look into it until now.

> The new implementation of the libMeshDM unfortunately fails when using variational inequality solvers VINEWTONRSLS or VINEWTONSSLS (for example miscellaneous/ex7). This happens when PETSc tries to VecDuplicate the solution vector for lower bound in src/snes/impls/vi/vi.c: 383 ?! 

To replicate, in miscellaneous_ex7, run
```
./example-opt --verbose dim=1 N=1024 initial_state=strip initial_center=0.5 initial_width=0.1 dt=1e-10 max_time=.5e-8 -snes_type vinewtonssls --use-petsc-dm
```

> This generates a seg fault on my mac with the PETSc and libMesh revisions I provided before.
>
> My guess is the seg fault happens probably because the solution vector is not initiated correctly.

The failure in misc ex7 was triggered by #1572

@fdkong, any ideas?  Any other takers?  I'm afraid I'm not familiar with either the code that's breaking *or* the code that broke it.

fdkong:
The vinewtonssls and vinewtonrsls setup functions require
a solution vector avalaible.

Closes #1752

### DIFF
--- a/src/solvers/petsc_nonlinear_solver.C
+++ b/src/solvers/petsc_nonlinear_solver.C
@@ -896,6 +896,10 @@ PetscNonlinearSolver<T>::solve (SparseMatrix<T> &  pre_in,  // System Preconditi
   // (petsc/petsc@154060b), so this code block should be safe to use
   // in 3.5.0 and later.
 #if !PETSC_VERSION_LESS_THAN(3,5,0)
+#if !PETSC_VERSION_LESS_THAN(3,6,0)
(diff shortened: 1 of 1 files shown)


jwpeterson:
Looks good to me. For reference, this is also related to #1596.

PR #1766, opened 2018-06-27
Disable HDF5 detection by default

### LINKED ISSUE #1585: Don't automatically enable HDF5
There may be [compatibility issues](https://groups.google.com/d/msg/moose-users/S8Wq7-Ep3bs/X2_CZf6ECQAJ) with our vendored NetCDF and system HDF5 installs, and our configure script is apparently not sophisticated enough to catch them at configure time. So I'd like to stop enabling HDF5 support by default in hdf5.m4...

jwpeterson:
We require an HDF5 which is compatible with our bundled NetCDF, but I don't want to maintain a sophisticated configure test for this since HDF5 is still an optional dependency.

Closes #1585.

### DIFF
--- a/configure
+++ b/configure
@@ -2030,7 +2030,8 @@ Optional Features:
   --disable-glpk          build without GLPK support
   --disable-nlopt         build without NLOPT support
   --disable-capnproto     build without Cap'n Proto support
-  --disable-hdf5          build without HDF5 support
+  --enable-hdf5           build libmesh with HDF5 support, the selected HDF5
+                          must be compatible with contrib/netcdf
   --disable-netcdf        build without netCDF binary I/O
   --disable-exodus        build without ExodusII API support
   --enable-exodus-fortran build with ExodusII Fortran API support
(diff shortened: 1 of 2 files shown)


roystgnr:
Do you think it would be safe to turn this back on if --enable-everything is specified?  I'm in the middle of a massive recompile because I just discovered that the ExodusII files I've generated don't work with h5dump or even with (my version of?) paraview...

jwpeterson:
On with `--enable-everything` is fine with me.

As an aside, if you have the MOOSE package installed, you should also have `/opt/moose/seacas/bin/ncdump` that will dump those Exodus files.

pbauman:
> As an aside, if you have the MOOSE package installed, you should also have /opt/moose/seacas/bin/ncdump that will dump those Exodus files.

I thought we also built those with libMesh now and installed them in `$LIBMESH_DIR/bin`?

jwpeterson:
Huh, it appears we do. I had forgotten about that...

pbauman:
> On with --enable-everything is fine with me.

Looks like we never did this? I'm updating my build scripts now, but looks this feature never crept back in.

roystgnr:
Nah; I'm a bit ambivalent about it myself, since it's actually given people compilation failures.  Are you also in favor of turning HDF5 back on in the --enable-everything case?

pbauman:
I would've been more in favor of it before I updated my build scripts (out laziness) :) But if you're feeling ambivalent about it, let's just leave it as is.

PR #1767, opened 2018-06-27
 Make sure we _find_global_index_by_pid_map in sync

roystgnr:
If we don't, and ranks disagree about whether we need to, then we
hang.

This fixes a bug from #1751 which was causing --distributed-mesh timeouts in some adaptive MOOSE tests.

I'm a little concerned - could there still be a bug here?  What happens if we coarsen the same number of elements we refine, so that ```mesh.n_active_elem()``` remains unchanged?

### DIFF
--- a/src/partitioning/partitioner.C
+++ b/src/partitioning/partitioner.C
@@ -976,7 +976,13 @@ void Partitioner::build_graph (const MeshBase & mesh)
   std::vector<const Elem *> neighbors_offspring;
 #endif
 
-   if (!_global_index_by_pid_map.size() || _global_index_by_pid_map.size() != mesh.n_active_elem())
+   bool need_global_index_by_pid_map =
+     !_global_index_by_pid_map.size() ||
+     _global_index_by_pid_map.size() != mesh.n_active_elem();
+
+   mesh.comm().max(need_global_index_by_pid_map);
+
+   if (need_global_index_by_pid_map)
      _find_global_index_by_pid_map(mesh);
 
    dof_id_type first_local_elem = 0;
--- a/src/partitioning/partitioner.C
+++ b/src/partitioning/partitioner.C
@@ -976,14 +976,11 @@ void Partitioner::build_graph (const MeshBase & mesh)
   std::vector<const Elem *> neighbors_offspring;
 #endif
 
-   bool need_global_index_by_pid_map =
-     !_global_index_by_pid_map.size() ||
-     _global_index_by_pid_map.size() != mesh.n_active_elem();
-
-   mesh.comm().max(need_global_index_by_pid_map);
-
-   if (need_global_index_by_pid_map)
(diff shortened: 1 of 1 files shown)


roystgnr:
If we match coarsening-deletion with equal refinement-creation then we need a different graph but I'm not sure we can't get away with reusing the same indexing.  Hilbert indexing ought to work better than slightly-scrambled indexing but perhaps as long as we're consistent between e.g. what we pass to what we receive from ParMETIS then the result shouldn't be *broken*?

jwpeterson:
I added a `--distributed-mesh -p 4` step to the "Test" target, but it went live slightly after this PR started testing so it won't show up here unless maybe I invalidate (?). I will try that momentarily.

jwpeterson:
Also, I'm going to temporarily push this branch into upstream so it can be immediately tested over on idaholab/moose#11754.

src/partitioning/partitioner.C
```diff
@@ -976,7 +976,13 @@ void Partitioner::build_graph (const MeshBase & mesh)
...
 #endif
 
-   if (!_global_index_by_pid_map.size() || _global_index_by_pid_map.size() != mesh.n_active_elem())
+   bool need_global_index_by_pid_map =
+     !_global_index_by_pid_map.size() ||
+     _global_index_by_pid_map.size() != mesh.n_active_elem();
+
+   mesh.comm().max(need_global_index_by_pid_map);
+
+   if (need_global_index_by_pid_map)
```
fdkong: Talked to @jwpeterson  a little on this. It looks like for me the best solution is remove `if` totally. Then we always construct the global map.

jwpeterson:
@fdkong also noted that calling `DistributedMesh::n_active_elem()` followed by `mesh.comm().max()` is already doing 2 global reductions, so we may not even be saving all that much in the end by avoiding the call to `_find_global_index_by_pid_map`...

src/partitioning/partitioner.C
fdkong: There may be some cases we can not reuse map even the n_ative_elems does not change.  For example, one core does a refinement while another core does coarsening.

jwpeterson:
BTW, this PR does get idaholab/moose#11754 passing, but it could be because we are not encountering a case where the number of active elements on a proc doesn't change even though the mesh (and hence the global mapping) does change in some way...

roystgnr:
We do an ```allgather()``` in the map reconstruction, which is asymptotically more expensive than an allreduce, then we do a sync (which currently includes another allgather in addition to a ton of sends and receives).  It's definitely worth avoiding map reconstruction whenever possible.

But I don't think we can safely avoid map reconstruction here, not without a much more sophisticated cache validity recognition mechanism than "did the number of active elements change?", so I'll take out the if entirely.

PR #1768, opened 2018-06-27
Avoid unused lambda capture warnings.

jwpeterson:
I see these on newer versions of clang (3.5) but not older (3.9) ones.

### DIFF
--- a/src/mesh/mesh_communication_global_indices.C
+++ b/src/mesh/mesh_communication_global_indices.C
@@ -833,8 +833,8 @@ void MeshCommunication::find_global_indices (const Parallel::Communicator & comm
 #ifndef NDEBUG
      & upper_bounds,
      & communicator,
-#endif
      & bbox,
+#endif
      & my_bin,
        my_offset
     ]
--- a/src/mesh/distributed_mesh.C
+++ b/src/mesh/distributed_mesh.C
@@ -1085,7 +1085,6 @@ DistributedMesh::renumber_dof_objects(mapvector<T *, dof_id_type> & objects)
   auto action_functor =
     [
 #ifndef NDEBUG
-     this,
      &first_object_on_proc,
      &objects_on_proc,
 #endif


src/mesh/mesh_communication_global_indices.C
```diff
@@ -841,6 +841,8 @@ void MeshCommunication::find_global_indices (const Parallel::Communicator & comm
     (processor_id_type, const std::vector<Parallel::DofObjectKey> & keys,
      std::vector<dof_id_type> & global_ids)
     {
+      // Avoid warnings about unused lambda captures in optimized builds.
+      libmesh_ignore(bbox);
```
roystgnr: Shouldn't we just move bbox into the ```#ifndef NDEBUG``` block?

jwpeterson: That would be fine with me. IIRC `#ifndef NDEBUG` is equivalent to `#ifdef DEBUG` when we're building libmesh...

roystgnr: Thanks!

> IIRC ```#ifndef NDEBUG``` is equivalent to ```#ifdef DEBUG``` when we're building libmesh...

Nope.  NDEBUG is turned off in devel mode; DEBUG isn't turned on until dbg mode.  This is actually super important because removing NDEBUG merely turns on all our constant-time-cost assertions while DEBUG turns on all our asymptotically-slower tests too.  So devel mode is still appropriate for debugging arbitrarily large problems while dbg mode is just intended to make debugging as easy as possible after you've gotten down to a cheap-enough test case.

jwpeterson: Hmm... then I think `bbox` is still an unused variable in devel mode in the current version of the code, as it is only used in an `#ifdef DEBUG` section.

roystgnr: Ah, hell.  Can you check?  I don't have a recent clang version handy.

jwpeterson: Yep, definitely back so I'm going to go back to the `libmesh_ignore()` strategy.

PR #1770, opened 2018-06-28
Add -Wdeprecated-declarations to the list of warnings ignored on clang.

jwpeterson:
After a bit of Internet detective work, I figured out that the ability
to disable this warning has likely been around since at least April
2012, which corresponds to Clang version 3.1. I've therefore decided
not to guard this line with specific clang versions, but that could be
added later if we decide it's necessary.


jwpeterson:
It shouldn't be possible for this to break any of the builds, but I do want to look at the test output from the build box and make sure it is definitely free of warnings...

PR #1774, opened 2018-06-29
Fix test for Hypre in $PETSC_ARCH builds

jwpeterson:
@brianmoose will attach an all-static recipe which builds PETSc to this PR...


jwpeterson:
I'm going to clean this up and merge since I'm convinced that is working now.

PR #1778, opened 2018-07-03
Optimize Parallel::Sort::communicate_bins()

roystgnr:
This method was taking an insane amount of time when used on a 512 core machine with a slow interconnect across two halves of it.

Using the current parallel_sync.h implementation instead of N_procs gather calls, Parallel::Sort is literally a thousand times faster for me on that case.  I'm also seeing about 75% speedup on a few more realistic cases, and I managed to cut out a hundred lines of code.  I'll merge this as soon as Civet is happy.

@bboutkov might be interested to see if this is any help on his stuff, though I believe we only go through this code path during repartitioning and I/O, not in the multigrid loop.


jwpeterson:
Is parallel sort unit tested/a good candidate for unit testing? Other than that, 👍

roystgnr:
"No" and "Wow yes", respectively.

I'll merge this as is, because I'm pretty confident in the test coverage, but in hindsight there were a few bugs in my first crack at this PR that would have been much easier to fix if they'd manifested as unit test failures rather than weirdness in subsequent find_global_indices() behavior.

jwpeterson:
Do you mind if we attach some parallel sweeps to this PR?

PR #1780, opened 2018-07-05
Don't set "I_duped_it" to true for undefined

permcody:
When using MPI_UNDEFINED, you don't expect to get a new communicator
object so when deleting the Communicator you end up tripping an assertion.


permcody:
Yeah, well I run in devel mode. I catch a lot of crap!

PR #1785, opened 2018-07-09
ErrorVector::plot_error Nemesis support

roystgnr:
Plotting error indicator output is often nice.  Plotting it to a
standard format that viz tools can read is nice.  Being unable to do
that on a distributed mesh without serializing first is not nice, so
let's fix that.

### DIFF
--- a/src/mesh/nemesis_io_helper.C
+++ b/src/mesh/nemesis_io_helper.C
@@ -1890,18 +1890,10 @@ void Nemesis_IO_Helper::compute_border_node_ids(const MeshBase & pmesh)
 
     // The number of node communication maps is the number of other processors
     // with which we share nodes. (I think.) This is just the size of the map we just
-    // created, minus 1.
+    // created, minus 1 unless this processor has no nodes of its own.
     this->num_node_cmaps =
-      cast_int<int>(proc_nodes_touched.size() - 1);
-
-    // If we've got no elements on this processor and haven't touched
-    // any nodes, however, then that's 0 other processors with which
-    // we share nodes, not -1.
-    if (this->num_node_cmaps == -1)
-      {
-        libmesh_assert (pmesh.active_elements_begin() == pmesh.active_elements_end());
-        this->num_node_cmaps = 0;
-      }
+      cast_int<int>(proc_nodes_touched.size() -
+                    proc_nodes_touched.count(this->processor_id()));
 
     // We can't be connecting to more processors than exist outside
     // ourselves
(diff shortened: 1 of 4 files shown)


roystgnr:
Huh.  Wonder why my runs didn't trigger *that*...

roystgnr:
This actually seems to be exposing a preexisting bug.  Not a very serious one, just another "more processors than elements" corner case.  I'll throw in the fix here when I'm done with it.

src/mesh/nemesis_io_helper.C
```diff
@@ -621,11 +621,14 @@ void Nemesis_IO_Helper::put_node_cmap(std::vector<std::vector<int>> & node_cmap_
...
+        nullptr : &node_cmap_node_ids_in[i][0];
+      int * proc_ids_ptr = node_cmap_proc_ids_in[i].empty() ?
+        nullptr : &node_cmap_proc_ids_in[i][0];
+
       nemesis_err_flag =
-        Nemesis::ne_put_node_cmap(ex_id,
-                                  this->node_cmap_ids[i],
-                                  &node_cmap_node_ids_in[i][0],
-                                  &node_cmap_proc_ids_in[i][0],
+        Nemesis::ne_put_node_cmap(ex_id, this->node_cmap_ids[i],
```
jwpeterson: Is it definitely safe to call `ne_put_node_cmap` with `nullptr` inputs? It looks like we weren't even guarding against accessing `[0]` of an empty vector before, so this is definitely an improvement.

src/mesh/nemesis_io_helper.C
```diff
@@ -574,18 +574,18 @@ void Nemesis_IO_Helper::put_cmap_params(std::vector<int> & node_cmap_ids_in,
...
-                                    &node_cmap_ids_in[0],
-                                    &node_cmap_node_cnts_in[0],
-                                    &elem_cmap_ids_in[0],
-                                    &elem_cmap_elem_cnts_in[0],
-                                    this->processor_id());
-    }
+  libmesh_assert(!node_cmap_ids_in.empty());
+  libmesh_assert(!node_cmap_node_cnts_in.empty());
+  libmesh_assert(!elem_cmap_ids_in.empty());
+  libmesh_assert(!elem_cmap_elem_cnts_in.empty());
```
jwpeterson: Maybe these should be `libmesh_error`s instead?

src/mesh/nemesis_io_helper.C
roystgnr: Definitely safe, as in there's a well-documented standard which says we can do this?  No.

Probably safe, as in I've hit a corner case which passes nullptr and our contrib Exodus was okay with that?  Yes.

src/mesh/nemesis_io_helper.C
roystgnr: Nah, this is an internal consistency check.  The library itself will (now) never trigger it, and it's in the write path not read path, so there's no way for a user to trigger this with a corrupt file or anything other than by manually misusing a low-level method.

roystgnr:
DistributedMesh is also happy with it on up to 24 processors.

PR #1789, opened 2018-07-10
Fix Make.common

### LINKED ISSUE #1776: Make.common missing references to METHOD=prof
Hi all, after building the library with `--enable-everything --with-methods="devel prof oprof"` I was trying to build / profile some examples and noticed that my Make.common didn't seem to have any reference to prof/pro/profiling builds:
```
/install/examples/introduction/ex1$ METHOD=prof make
/bender1/data/users/borisbou/software/master/libmesh/install/Make.common:115: *** Invalid environment variable METHOD=prof specified.
```
For now as a quick workaround I drop in the following snippet directly into the installed Make.common and things build as expected, but since I wasn't sure of the proper fix, I thought I'd log this issue for later. Thanks!

```# If METHOD=prof, compile in profile mode.                                                                                                                                                                                                     
ifeq ($(METHOD),prof)
  prof-mode := on
  mode       := "prof"
endif
```

roystgnr:
Looks like the set of people using our Make.common and the set of people using prof mode just never overlapped?

This resolves #1776, albeit nearly a week late.

### DIFF
--- a/contrib/utils/Make.common.in
+++ b/contrib/utils/Make.common.in
@@ -1,6 +1,6 @@
 ###############################################################################
 ## Please note:
-##   The ./configure script generates this Make.common from 
+##   The ./configure script generates this Make.common from
 ##   contrib/utils/Make.common.in, so if you want to change something,
 ##   then do it in the latter file (here!) and re-run
 ##   ./configure in the top level directory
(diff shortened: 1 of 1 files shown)


bboutkov:
Oh. That's embarrassingly simple - I should have done that myself, so thanks. 

But I suppose part of the reason I held off was that I was unsure of the pro vs prof vs profiling synonyms which I've seen for example in /contrib/bin/libmesh-config.in, as compared to the recommended 'pro' in the README and wasn't sure if we were handling these things at a higher level somewhere, or phasing out one name in favor of another.

roystgnr:
You're right - I just use "prof" myself, and I was simply looking at the oprofile case to guess the synonyms we supported, when I should have been looking at libmesh_method.m4.  pro/prof/profiling it is.

I'll fix that up and then merge.

PR #1794, opened 2018-07-12
Implemented densematrix inverse based on svd (#1786)

### LINKED ISSUE #1786: Dense Matrix Inverse based on SVD
Implemented a new function for taking inverse of a dense matrix using svd

deepaky7131:
Close #1786

### DIFF
--- a/include/numerics/dense_matrix.h
+++ b/include/numerics/dense_matrix.h
@@ -62,16 +62,12 @@ public:
               const unsigned int new_n=0);
 
   /**
-   * The 5 special functions can be defaulted for this class, as it
-   * does not manage any memory itself.
+   * Destructor.  Empty.
    */
-  DenseMatrix (DenseMatrix &&) = default;
-  DenseMatrix (const DenseMatrix &) = default;
-  DenseMatrix & operator= (const DenseMatrix &) = default;
-  DenseMatrix & operator= (DenseMatrix &&) = default;
-  virtual ~DenseMatrix() = default;
+  virtual ~DenseMatrix() {}
 
-  virtual void zero() override;
+
+  virtual void zero() libmesh_override;
 
   /**
    * \returns The \p (i,j) element of the matrix.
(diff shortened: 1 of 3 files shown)


deepaky7131:
@jwpeterson : Can you please review the code

include/numerics/dense_matrix_impl.h
```diff
@@ -808,6 +793,34 @@ void DenseMatrix<T>::svd (DenseVector<Real> & sigma,
...
+
+Inverse.resize(VT.n(),U.m());
+
+for (unsigned i =0; i<VT.n(); ++i)
+{
+  for (unsigned j = 0; j<U.m(); ++j)
+  {
+    for (unsigned k = 0; k<VT.m(); ++k)
+    {
+      Inverse(i,j) += VT(k,i)*U(j,k)/sigma(k);
```
jwpeterson: This isn't right for the case where some of the singular values are nearly zero, is it?

You don't want to divide by `sigma(k)` in that case in order to avoid introducing round-off error. See also the `_svd_solve_lapack` function where the user supplies an `rcond` value...

deepaky7131:
Yes. may be i will add an if condition based on the record value.

On Thu, Jul 12, 2018 at 12:29 PM, John W. Peterson <notifications@github.com

include/numerics/dense_matrix.h
```diff
@@ -422,6 +406,16 @@ class DenseMatrix : public DenseMatrixBase<T>
    * Requires PETSc >= 3.1 since this was the first version to provide
    * the LAPACKgelss_ wrapper.
    */
+
+   void svdinv(DenseMatrix<Number> & Inverse);
```
jwpeterson: You may also want to rethink this interface. If the pseudo-inverse will be computed many times, it will be more efficient to reuse the sigma, U, and VT arrays rather than creating and destroying them each time. Passing 4 arguments to the function might be a better API:
```
void svdinv(DenseMatrix<Number> & Inverse,
const DenseVector<Real> & sigma,
const DenseMatrix<Number> & U,
const DenseMatrix<Number> & VT);
```

jwpeterson:
Closing this PR due to inactivity, feel free to reopen if you are able to address the issues noted.

Issue #1793, opened 2018-07-12
Compiler Warning parmetis_partitioner.C


pbauman:
Saw this while compiling from scratch. Don't have time ATM to cook up PR, sorry.

<pre>
../../pbauman/src/partitioning/parmetis_partitioner.C: In member function ‘virtual void libMesh::ParmetisPartitioner::_do_repartition(libMesh::MeshBase&, unsigned int)’:
../../pbauman/src/partitioning/parmetis_partitioner.C:194:95: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
   this->assign_partitioning (mesh, reinterpret_cast<std::vector<dof_id_type> &> (_pmetis->part));
</pre>

<pre>
[10:57:01][pbauman@fry:~]$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/fry1/data/shared/software/apps/gcc/7.2.0/libexec/gcc/x86_64-pc-linux-gnu/7.2.0/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: /fry1/data/shared/software/sourcesdir/gcc/gcc-7.2.0-src/configure --prefix=/fry1/data/shared/software/apps/gcc/7.2.0 --with-gmp=/fry1/data/shared/software/libs/gmp/6.1.1 --with-mpfr=/fry1/data/shared/software/libs/mpfr/3.1.5 --with-mpc=/fry1/data/shared/software/libs/mpc/1.0.3 --disable-multilib --enable-languages=c,c++,fortran --enable-lto
Thread model: posix
gcc version 7.2.0 (GCC) 
</pre>

jwpeterson:
That's... annoying. We should be able to see this on CIVET as well, since I believe we have GCC 7.x targets. Do you see this warning in opt, dbg, and devel modes?

roystgnr:
I get this too, only in devel mode.  I'm looking at warnings in another branch already; let me see if I can find a fix for this to slip in there.

PR #1792, opened 2018-07-12
Update boundary_project_solution to project on edge and node boundary IDs

dknez:
Follow the approach from DofMap::apply_dirichlet_impl to make boundary_project_solution work with edge and node boundary IDs.

PR #1795, opened 2018-07-12
More warning fixes

roystgnr:
This cleans up the warning from #1793 as well as a couple others that had been bothering me.  libMesh (except for contrib/) now compiles cleanly for me.

### DIFF
--- a/include/partitioning/parmetis_helper.h
+++ b/include/partitioning/parmetis_helper.h
@@ -73,16 +73,20 @@ public:
   std::vector<Parmetis::idx_t>  vtxdist;
   std::vector<Parmetis::idx_t>  xadj;
   std::vector<Parmetis::idx_t>  adjncy;
+
+  // We use dof_id_type for part so we can pass it directly to
+  // Partitioner:: methods expecting that type.
   std::vector<dof_id_type>  part;
+
+  // But we plan to pass a pointer to part as a buffer to ParMETIS, so
+  // it had better be using a simply reinterpretable type!
+  static_assert(sizeof(Parmetis::idx_t) == sizeof(dof_id_type),
+                "ParMETIS and libMesh ID sizes must match!");
+
   std::vector<Parmetis::real_t> tpwgts;
   std::vector<Parmetis::real_t> ubvec;
   std::vector<Parmetis::idx_t>  options;
   std::vector<Parmetis::idx_t>  vwgt;
-
-  // We plan to pass part as a buffer to ParMETIS, so it had better be
-  // using the same type!
-  static_assert(sizeof(Parmetis::idx_t) == sizeof(dof_id_type),
-                "ParMETIS and libMesh ID sizes must match!");
 
   Parmetis::idx_t wgtflag;
   Parmetis::idx_t ncon;
--- a/include/partitioning/parmetis_helper.h
+++ b/include/partitioning/parmetis_helper.h
@@ -73,11 +73,16 @@ public:
   std::vector<Parmetis::idx_t>  vtxdist;
   std::vector<Parmetis::idx_t>  xadj;
   std::vector<Parmetis::idx_t>  adjncy;
-  std::vector<Parmetis::idx_t>  part;
+  std::vector<dof_id_type>  part;
   std::vector<Parmetis::real_t> tpwgts;
   std::vector<Parmetis::real_t> ubvec;
   std::vector<Parmetis::idx_t>  options;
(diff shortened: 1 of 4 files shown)


src/partitioning/parmetis_partitioner.C
```diff
@@ -185,13 +185,13 @@ void ParmetisPartitioner::_do_repartition (MeshBase & mesh,
                                        &itr,
                                        &_pmetis->options[0],
                                        &_pmetis->edgecut,
-                                       _pmetis->part.empty()    ? nullptr : &_pmetis->part[0],
+                                       _pmetis->part.empty()    ? nullptr : reinterpret_cast<Parmetis::idx_t *>(&_pmetis->part[0]),
```
jwpeterson: I suppose this avoids the warning because there's no dereferencing happening? Sounds like a reasonable fix to me, but I would mind if you also documented why `part` does not use `Parmetis::idx_t` like all the other things we're passing to Parmetis.

include/partitioning/parmetis_helper.h
```diff
@@ -73,12 +73,17 @@ class ParmetisHelper
...
-  std::vector<Parmetis::idx_t>  part;
+  std::vector<dof_id_type>  part;
   std::vector<Parmetis::real_t> tpwgts;
   std::vector<Parmetis::real_t> ubvec;
   std::vector<Parmetis::idx_t>  options;
   std::vector<Parmetis::idx_t>  vwgt;
 
+  // We plan to pass part as a buffer to ParMETIS, so it had better be
+  // using the same type!
+  static_assert(sizeof(Parmetis::idx_t) == sizeof(dof_id_type),
```
jwpeterson: This static assert is now in two different places, no?

src/partitioning/parmetis_partitioner.C
roystgnr: The comment in the header isn't good enough?

roystgnr: Ah, no, because that discusses passing it to parmetis not to libMesh.  Will fix.

tests/driver.C
```diff
@@ -1,5 +1,8 @@
+// Ignore overloaded-virtual warnings coming from cppunit headers
+#include <libmesh/ignore_warnings.h>
 #include <cppunit/extensions/TestFactoryRegistry.h>
 #include <cppunit/ui/text/TestRunner.h>
+#include <libmesh/restore_warnings.h>
```
jwpeterson: Nice. Just the other day I was seeing new warnings from cppunit headers which I thought I fixed. It's really unbelievable to me how many warnings are generated by that codebase.

include/partitioning/parmetis_helper.h
roystgnr: Whoops; I'd forgotten I'd insisted on it before the reinterpret_cast before.  I'll leave it in the header next to the documentation of the type choices and I'll remove the old one in the C file.

roystgnr:
Hahaha, no, this doesn't get rid of all my warnings, because we *just* merged an unused variable warning in #1792 for BOUNDARY_ID_MIN_Z

We're close though.  We ought to figure out how to stay close.  Simply configuring in CI with libmesh_CXXFLAGS='-Werror' isn't going to work, I don't think, because we can't apply that flag to contrib/ without either massive forking or massive upstream bugfixes of a lot of code in there.  What if I add a WERROR_FLAGS (analogous to PROFILING_FLAGS, ASSEMBLY_FLAGS) to our compiler.m4, then we use it on library code only whenever the user configures with ```--enable-werror```?

We'd probably still want to leave it out of Makefile.common.in and libmesh-config and anything else users scarf?

Then we can then ratchet up our code standards (via some ```--enable-paranoid-warnings``` setting) permanently.  Every few years I turn on a zillion warning options and fix what I can, but it would be awesome to be able to do that and know the result wouldn't start to bit rot as soon as I looked away.

pbauman:
> We're close though. We ought to figure out how to stay close. Simply configuring in CI with libmesh_CXXFLAGS='-Werror' isn't going to work, I don't think, because we can't apply that flag to contrib/ without either massive forking or massive upstream bugfixes of a lot of code in there.

This is the only reason I haven't turned it on in my femputer libMesh builds.

> What if I add a WERROR_FLAGS (analogous to PROFILING_FLAGS, ASSEMBLY_FLAGS) to our compiler.m4, then we use it on library code only whenever the user configures with --enable-werror?

That would be awesome! And I would enable it right away on femputer.

PR #1797, opened 2018-07-13
configure --enable-werror option

roystgnr:
As discussed in #1795 

This builds cleanly for me, so hopefully @pbauman can just turn it on right away in his GRINS tests, which I think use the same compiler and similar stack.  Turning it on in MOOSE tests *ought* to be safe too, since we don't pass -Werror anywhere outside the library, but I'm less confident there.


pbauman:
Job [GRINS-devel](https://femputer.eng.buffalo.edu:8000/job/1913/) on a5bd9f3 : invalidated by @pbauman

Turned on --enable-werror for libMesh devel build

pbauman:
OK, so it doesn't show up in CXXFLAGS at the end of configure and only gets enabled on a subset of directories? Perhaps we add a line to the configure output for non-contrib-cxxflags or something for that sanity check?

pbauman:
(I ask because I was trying to sanity check -Werror was showing up in the GRINS-devel build running now, but I didn't see the flag show up at the end of configure, but did confirm I passed `--enable-werror` to configure in that build)

roystgnr:
Hah, thanks!  I'll fix that.  I just did a "make" but not a "make check" to verify things myself.

I'll add a "werror flag" entry line to the config summary too.

pbauman:
I'll go ahead and enable werror in the dbg build as well. If there's interest, I can add an opt version at a later date.

roystgnr:
We've got opt testing in the MOOSE CI; it might be nice to --enable-werror there, both to catch any unused-when-asserts-are-off parameters and to halt those builds right away if there's anything GRINS will catch that will need fixing.  I'll leave that up to @jwpeterson etc. though; since they're using clang instead of gcc IIRC it's not an *obvious* decision.

roystgnr:
Meanwhile, I'm thinking of adding --enable-paranoid-warnings next, and we can turn that on in CI too.  Out of the ~60 extra warnings options I used to manually check out, 32 still seem to pass out-of-the-box with flying colors, and we could start to enforce that permanently.

PR #1798, opened 2018-07-13
Enable hypre matrix type

### LINKED ISSUE #1385: Take advantage of other Mat types in PETSc
We currently build `PetscMatrix` objects with underlying `MATAIJ` and `MATBAIJ` types in LibMesh, but there are other types, including `MATHYPRE` (petsc/petsc@63c07aad33d943fe85193412d077a1746a7c55aa) which (I assume) could save quite a bit of memory for people who are using BoomerAMG as a preconditioner/solver. Based on the relevant [PETSc example](https://github.com/petsc/petsc/blob/master/src/mat/examples/tests/ex115.c), one also needs to call the appropriate preallocation function (`MatHYPRESetPreallocation`) and there may be some other limitations:
```
/* HYPRE_IJMatrix does not support INSERT_VALUES with off-proc entries */
```
```
/* MatAXPY_Basic further exercises MatSetValues_HYPRE */
      /* however there are still some issues
         with rectangular matrices.
         try, e.g., mpiexec -n 3 ./ex115 -M 7 -N 6 -test_rectangular */
```
so this needs further investigation.

@fdkong will be looking into this in the somewhat near future.

fdkong:
Closes #1385

### DIFF
--- a/include/numerics/petsc_matrix.h
+++ b/include/numerics/petsc_matrix.h
@@ -60,6 +60,10 @@ namespace libMesh
 // Forward Declarations
 template <typename T> class DenseMatrix;
 
+enum PetscMatrixType : int {
+                 AIJ=0,
+                 HYPRE};
+
 
 /**
  * This class provides a nice interface to the PETSc C-based data
@@ -108,6 +112,8 @@ public:
   PetscMatrix & operator= (const PetscMatrix &) = delete;
   PetscMatrix & operator= (PetscMatrix &&) = delete;
   virtual ~PetscMatrix ();
+
+  void set_matrix_type(PetscMatrixType mat_type);
 
   virtual void init (const numeric_index_type m,
                      const numeric_index_type n,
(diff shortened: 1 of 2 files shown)


jwpeterson:
This will probably pass our testing because it's not actually using the PETSc 3.9.3 code paths, but according to @fdkong this should not be merged until those APIs in PETSc are finalized a bit more.

pbauman:
>  APIs in PETSc are finalized 

hahahahaha

fdkong:
@jwpeterson The patch is already in PETSc-maint and PETSc-master. I think it is fine right now to add this into libmesh

jwpeterson:
> The patch is already in PETSc-maint and PETSc-master.

The preprocessor guards in this PR are for PETSc >= 3.9.4. Is there actually going to be a 3.9.4, and would your changes be in it or in the 3.10 release series?

fdkong:
My change will go to PETSc3.9.4 and higher. The current tag is 3.9.3, and my change is already maint. Next tag should be 3.9.4. Even next tag is 3.10.x, this code is still fine, I guess.

PR #1800, opened 2018-07-16
configure --enable-paranoid-warnings option

roystgnr:
My preference would be for us to enable both --enable-werror *and* this in CI, as mentioned in #1797, so that we can easily enforce (compiler-recognized) stricter code standards by adding the warning flags for those standards to the list as soon as we compiler cleanly with each.

I'd be willing to back down on adding this to CI (although IMHO it'd be no worse than when we auto-reject trailing whitespace), but I'd still at least want to add this to libMesh to make manual checks easier.


pbauman:
Want me to turn this on in one of the femputer builds?

pbauman:
Job [GRINS-devel](https://femputer.eng.buffalo.edu:8000/job/1923/) on f616f53 : invalidated by @pbauman

Turned on --enable-paranoid-warnings for devel build

pbauman:
Cool. I'll leave these paranoid warnings on for the devel build.

roystgnr:
Sweet.  Let's leave them on for a while, with the understanding that we may want to turn them back off again or shave off a few warnings if other developers find the Civet failures to be too much of a PITA.  You could probably turn them on in dbg, too.  I'm 99% sure that will pass out-of-the-box (since it passed on-my-box) and it'll add coverage for our handful of ```#ifdef DEBUG``` sections.

pbauman:
Cool, I've turned paranoid warnings on for dbg as well

PR #1801, opened 2018-07-16
Ghost API updates

roystgnr:
I'm not sure the new versions of these comments are as clear as they should be, but since the parts I'm replacing were outright wrong, "unclear" should at least be a big improvement.

Thanks to @dschwen for catching this and pointing it out in #1028

### DIFF
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -367,8 +367,13 @@ public:
   bool on_boundary () const;
 
   /**
-   * \returns \p true if this element is semilocal to the calling
+   * \returns \p true if this element is "semilocal" to the calling
    * processor, which must specify its rank.
+   *
+   * This method is discouraged, as it uses the *old* definition of
+   * semilocal (elements which are not local but which are point
+   * neighbors of something local) rather than any of the new
+   * definitions discussed in ghosting_functor.h
    */
   bool is_semilocal (const processor_id_type my_pid) const;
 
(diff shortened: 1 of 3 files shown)


pbauman:
Job [GRINS-devel](https://femputer.eng.buffalo.edu:8000/job/1925/) on 93c7714 : invalidated by @pbauman

This looks like the &quot;not-as-heisny-but-still-there-heisenbug&quot; I&#39;d thought I&#39;d gotten rid of.

include/geom/elem.h
```diff
@@ -399,8 +399,13 @@ class Elem : public ReferenceCountedObject<Elem>,
   bool on_boundary () const;
 
   /**
-   * \returns \p true if this element is semilocal to the calling
+   * \returns \p true if this element is "semilocal" to the calling
    * processor, which must specify its rank.
+   *
+   * This method is discouraged, as it uses the *old* definition of
```
dschwen: This is used in MOOSE (for example here: https://github.com/idaholab/moose/blob/devel/framework/include/mesh/MooseMesh.h#L149-L156). There are applications where iterating over all non local point neighbors (and _just_ those) would be super useful.

dschwen:
What ever happened to this? @roystgnr

dschwen:
Stale? Superseded and fixed elsewhere?

roystgnr:
Neither, sorry; this just got bumped too low on my TODO list.

roystgnr:
> Superseded and fixed elsewhere?

I finally got around to merging the comment updates while digging through old branches ... but are there any remaining real fixes to make?  We got a fix for the `evaluable_elements_*` iterators in #1847, and I don't recall anything else outstanding, but it's been so long since these threads I might be forgetting something important.

PR #1802, opened 2018-07-17
Make the laplacian smoother respect block boundaries too

friedmud:
Also update find_boundary_nodes() so it doesn't build a std::vector<bool> that is n_nodes in length

### DIFF
--- a/include/mesh/mesh_tools.h
+++ b/include/mesh/mesh_tools.h
@@ -39,6 +39,7 @@ enum ElemType : int;
 #include <vector>
 #include <set>
 #include <limits>
+#include <unordered_set>
 
 namespace libMesh
 {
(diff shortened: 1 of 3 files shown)


include/mesh/mesh_tools.h
```diff
@@ -134,12 +134,27 @@ void build_nodes_to_elem_map (const MeshBase & mesh,
...
+/**
+ * Returns a std::set containing Node IDs for all of the boundary nodes
+ */
+std::set<dof_id_type> find_boundary_nodes(const MeshBase & mesh);
+
+/**
+ * Returns a std::set containing Node IDs for all of the block boundary nodes
+ *
+ * A "block boundary node" is a node that is connected to elemenents from 2
+ * or more blockse
```
roystgnr: blockse?

include/mesh/mesh_tools.h
```diff
@@ -134,12 +134,27 @@ void build_nodes_to_elem_map (const MeshBase & mesh,
...
  * is on the domain boundary (true)) or not (false).
  */
 void find_boundary_nodes (const MeshBase & mesh,
                           std::vector<bool> & on_boundary);
+#endif
+
+/**
+ * Returns a std::set containing Node IDs for all of the boundary nodes
+ */
+std::set<dof_id_type> find_boundary_nodes(const MeshBase & mesh);
```
roystgnr: Should we be using unordered_set for these?  We never expect them to be sorted, but they might be large enough for O(1) lookup to make a difference.

roystgnr:
:+1:, go ahead and merge after those two tweaks.

friedmud:
Made the changes - I'll let testing finish though

PR #1803, opened 2018-07-18
GCC 8 warning fixes

roystgnr:
This is the dark side to ```configure --enable-werror``` - just because we've fixed all the warnings that one compiler found doesn't mean we'll have fixed all the warnings that the very next version of the very same compiler will find with the very same compiler arguments.

On the bright side, now that I'm building with --enable-error I can actually notice these things and fix them right away.

### DIFF
--- a/include/base/getpot.h
+++ b/include/base/getpot.h
@@ -1718,8 +1718,9 @@ GetPot::_internal_managed_copy(const std::string& Arg) const
     return *it;
 
   // Otherwise, create a new one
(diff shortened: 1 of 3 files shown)


pbauman:
Job [GRINS-devel](https://femputer.eng.buffalo.edu:8000/job/1931/) on 69fd152 : invalidated by @pbauman

I really wish I could reliably trigger this failure (that&#39;s obviously something wrong with the testing infrastructure for that test).

PR #1804, opened 2018-07-18
Fix new unit test failures with --enable-parmesh

roystgnr:
The test @dknez added in #1792 was breaking for me on distributed meshes, but the fix is simple.

@pbauman, should we switch the GRINS-devel CI to use --enable-parmesh, so we get regular coverage of that?  We definitely don't want to change GRINS-dbg or MOOSE Test Debug (there are some stupidly slow dbg-mode-only tests on large distributed problems), but changing GRINS-devel should give us pretty full coverage at little to negative net cost.

### DIFF
--- a/tests/systems/systems_test.C
+++ b/tests/systems/systems_test.C
@@ -347,6 +347,10 @@ public:
     ZeroFunction<> zf;
     sys.boundary_project_solution(boundary_ids, variables, &zf);
 
+    // On a distributed mesh we may not have every node on every
+    // processor
+    TestCommWorld->set_union(projected_nodes_set);
+
     // We set the solution to be 1 everywhere and then zero on specific boundary
     // nodes, so the final l1 norm of the solution is the difference between the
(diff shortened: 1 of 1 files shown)


pbauman:
It will honestly probably be easier just to add another recipe (don't ask). I'll try and get to it today.

dknez:
Thanks for fixing the distributed mesh issue from #1792. I do not have an --enable-parmesh build at the moment.

roystgnr:
> It will honestly probably be easier just to add another recipe (don't ask).

Gah, but I have to ask: are you sure?  There have been a few times just recently when femputer has been on the critical path in our "getting junk tested and merged" PERT chart, and giving it 50% more work to do could be a problem in those cases.

> I'll try and get to it today.

No rush.  There may be other bit rot to catch before the tests pass, assuming you've been as lax about manual testing as I have, and fixing that's not too urgent.

PR #1810, opened 2018-07-24
 BoundaryInfo::build_foo_list(): Sort before returning

jwpeterson:
We ran into an issue while developing idaholab/moose#11800 where the vectors returned on different processors were in different orders (even with ReplicatedMesh). This was caused because the underlying containers are stored in memory address order, which is obviously not guaranteed to be consistent across different processors.

This adds an extra O(N lg N) operations to the existing O(N) algorithm, but N here is the number of boundary ids, and IMO ~~correctness~~ ease of use trumps performance in this case.


roystgnr:
Well, not correctness, but definitely ease of use.  :+1:

PR #1812, opened 2018-07-24
Add more paranoid warnings

roystgnr:
These commits include fixes (at least on my system; we'll see if femputer agrees) for many instances of warning triggers for:
```
-Wcast-qual -Wredundant-decls -Wshadow -Wstrict-aliasing -Wswitch-default -Wlogical-op
```
(as well as for a few conversion issues, but I started separating those out after realizing how vastly more prevalent they were)

And the last fix adds those warnings to our ```--enable-paranoid-warnings``` option, so that Paul's CI can scream at us if we backslide in the future.

I'm also wrapping more of our third-party headers in ignore_warnings/restore_warnings; the goal here is to keep our own stuff cleaner, not fix the world.

### DIFF
--- a/configure
+++ b/configure
@@ -7843,12 +7843,12 @@ fi
           CFLAGS_DBG="-g -Wimplicit"
           ASSEMBLY_FLAGS="$ASSEMBLY_FLAGS -fverbose-asm"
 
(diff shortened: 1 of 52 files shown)


roystgnr:
Going to merge these tomorrow unless anyone screams.  IMHO these should be uncontroversial improvements; it's the subsequent (-Wconversion) PR that will have to overuse an ugly idiom.

PR #1819, opened 2018-08-07
Make default constructors public for TypeVector and TypeTensor

lindsayad:
In the course of trying to get automatic differentiation into MOOSE, I've found it advantageous and perhaps necessary to modify some of the operator overloads in `MetaPhysicL` as shown for example in the `NumberArray` code starting [here](https://github.com/lindsayad/moose/blob/ad-moose-wip/framework/contrib/metaphysicl/include/metaphysicl/numberarray.h#L335). The default construction of a `NumberArray` has some consequences, specifically it leads to default construction of its data of type `T`. An operation like `NumberArray<N, TensorValue<Real>> * Real` yields a resulting type `NumberArray<N, TypeVector<Real>`, which means that in my overloaded `operator *` I  want to default construct `TypeVector`. I would like to be able to do this...

### DIFF
--- a/include/numerics/type_tensor.h
+++ b/include/numerics/type_tensor.h
@@ -38,9 +38,7 @@ template <unsigned int N, typename T> class TypeNTensor;
 
 /**
  * This class defines a tensor in \p LIBMESH_DIM dimensional space of
- * type T.  T may either be Real or Complex.  The default constructor
- * for this class is protected, suggesting that you should not
- * instantiate one of these directly.
+ * type T.  T may either be Real or Complex.
  *
  * \author Roy Stogner
  * \date 2004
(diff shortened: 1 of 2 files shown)


roystgnr:
:+1: from me and feel free to merge, but I fear this is a [Chesterton's fence](https://en.wikipedia.org/wiki/Wikipedia:Chesterton%27s_fence) situation.  Were we really just trying to make it harder for users to accidentally construct zero-objects that would then be immediately rewritten??  If so then that's not worth the hassle, but maybe there was another reason I've forgotten.

lindsayad:
I like the reference. I guess my argument in favor would be that this doesn't appear to affect any of our tests...but then again why should it, since we've never allowed it before.

I don't think I have merge rights for libMesh.

roystgnr:
Clearly @pbauman is falling down on the "dictatorial powers while Roy and John are gone" job!  I'll merge now but ping @jwpeterson so he can have a chance to tell us what kind of mistake we've made later.  :-)

pbauman:
> Clearly @pbauman is falling down on the "dictatorial powers while Roy and John are gone" job!

Psh, I was just going to push my own stuff through, everyone else can wait. :P

PR #1820, opened 2018-08-07
Introduce *= with tensor in type_tensor

lindsayad:
Would like to have the opportunity to do `*=(Tensor)` operations

### DIFF
--- a/include/numerics/type_tensor.h
+++ b/include/numerics/type_tensor.h
@@ -250,8 +250,15 @@ public:
    *
    * \returns A reference to *this.
    */
-  template <typename Scalar>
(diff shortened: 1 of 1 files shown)


pbauman:
Job [GRINS-dbg](https://femputer.eng.buffalo.edu:8000/job/2095/) on 015dcd6 : invalidated by @pbauman

I believe this was test heisenbug.

roystgnr:
:+1: since this looks like a pure improvement as-is, but is there any reason to require that both tensors have the same underlying type, rather than templating around a T2?

lindsayad:
Nope, no reason 😄

PR #1823, opened 2018-08-17
Make StandardType<Tuple> work?

friedmud:
The `?` above is because I don't really understand how this code has existed for this long like it is.

Are these changes only required for my compiler?  Or did someone write this and never test it - ever?

At any rate... it appears to be working now!

### DIFF
--- a/include/parallel/standard_type.h
+++ b/include/parallel/standard_type.h
@@ -24,6 +24,7 @@
 
 // libMesh Includes
 #include "libmesh/libmesh_common.h"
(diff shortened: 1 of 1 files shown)


roystgnr:
Wrote it because I thought I needed it for an algorithm, realized I had a better way to implement the algorithm without it, did that instead, then was too frugal to delete already-written code but too lazy to write test code for an otherwise-unused capability.  Sorry for the hassle and thanks for the fixes!

Issue #1824, opened 2018-08-18
Slowness in loose_bounding_box


friedmud:
Seeing some slowness with point locator that I think is associated with Cell::loose_bounding_box().

The offender is really `build_side_ptr()`... we really need to not call that as much as possible - the dynamic memory creation / destruction is just nasty.

![profile001](https://user-images.githubusercontent.com/870705/44295231-46049400-a262-11e8-8351-0b2b03f8353d.png)

I think the fix here for me is that there needs to be a specialization of `loose_bounding_box()` for `Hex8` that is way simpler.  We know that the sides can't be bowed out or anything - so we don't need to build them.

friedmud:
Ok - I fixed this in #1825 by making all of the 'linear' elements use a simple method for calculating a bounding box.  Here's what this profiling looks like now:

![profile002](https://user-images.githubusercontent.com/870705/44295672-10b07400-a26b-11e8-8d4d-f1a5e350c219.png)

Somewhere in the 4x speedup or so (looking at `TrackReader::execute()` which heavily utilizes the point locator).

This will also make the point locator scale better with threads - that memory thrashing in `Cell::loose_bounding_box()` will kill threading.

roystgnr:
Let's leave this open since it's still a problem for higher-order geometric elements?

roystgnr:
And we can discuss possible solutions for that too.  Could you try sticking a ```has_affine_map()``` test into ```Cell::loose_bounding_box()``` and falling back to ```Elem::loose_bounding_box``` if it returns true?  That would be pretty cheap in terms of man-hours if it turns out to also save cpu-hours in the common "we have second order elements with mostly-or-all-affine maps that we use for the higher order dof storage" case.

The longer-term fixes I see may be harder.  Side has to be templated around SideType to give access to all the methods overridden by the side... but SideType is indeterminate at the Cell level, so any method there has to either itself be templated (which loose_bounding_box obviously isn't and won't be) or has to construct Side dynamically (as build_side_ptr does).

[...]

PR #1825, opened 2018-08-18
Simplify the loose bounding box for 'linear' elements

friedmud:
Use the "simple" method for computing loose bounding boxes for all 'linear' elements.  I know that the 3D ones can technically have curved sides - but this is such a huge optimization... I'm interested in discussion about it.

You can see the before and after over here: #1824

### DIFF
--- a/src/geom/cell_hex8.C
+++ b/src/geom/cell_hex8.C
@@ -368,4 +368,10 @@ Real Hex8::volume () const
     triple_product(q[2], q[4], q[5]) / 64.;
 }
 
+BoundingBox
+Hex8::loose_bounding_box () const
+{
+  return Elem::loose_bounding_box();
+}
+
 } // namespace libMesh
--- a/include/geom/cell_hex8.h
+++ b/include/geom/cell_hex8.h
@@ -170,6 +170,11 @@ public:
    */
   virtual Real volume () const override;
 
+  /**
+   * Builds a bounding box out of the nodal positions
+   */
+  virtual BoundingBox loose_bounding_box () const override;
+
 protected:
 
   /**
--- a/include/geom/cell_prism6.h
+++ b/include/geom/cell_prism6.h
@@ -168,6 +168,11 @@ public:
    */
   virtual Real volume () const override;
 
+  /**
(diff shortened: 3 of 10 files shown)


pbauman:
Job [GRINS-devel](https://femputer.eng.buffalo.edu:8000/job/2119/) on 37500cb : invalidated by @pbauman

False negative on test I believe

roystgnr:
The 3D ones can have curved sides, but I don't believe any coordinate of the sides can end up outside the bounding box of the nodes.  The coefficients when converting from master to physical coordinates are all positive in the bi/tri-linear cases, right?

Unless someone embarrasses me with a counterexample, IMHO we should merge this as-is.  No matter how we take care of the build_side_ptr memory thrashing issue in generall, in this specific use case we'll still do better with the simpler code here.

src/geom/cell_hex8.C
```diff
@@ -368,4 +368,10 @@ Real Hex8::volume () const
     triple_product(q[2], q[4], q[5]) / 64.;
 }
 
+BoundingBox
+Hex8::loose_bounding_box () const
+{
+  return Elem::loose_bounding_box();
```
jwpeterson: I'm probably going to make myself look like an idiot here, but I just don't get it. `Elem::loose_bounding_box()` is already virtual and is not overridden by `Hex8` or any of its bases, so what is this actually doing differently from before?

roystgnr:
Skimming over a couple papers on generalized barycentric coordinates makes me more confident that there aren't any counterexamples here.  So I'll merge right away and we'll upgrade "embarrasses me" to "horribly embarrasses me" if someone finds a counterexample anyway.

src/geom/cell_hex8.C
roystgnr: See Cell::loose_bounding_box()

jwpeterson: Ah, forgot about `Cell`. OK, now I get it :)

friedmud:
Thanks for working through this and merging it - I came to the same conclusion about the coordinates of bilinears always falling within the extremum of the nodes....

PR #1826, opened 2018-08-19
Re-implement parallel sync using the NBX algorithm.

friedmud:
Also adds nonblocking_barrier() and possibly_receive()

Switch to a true sparse parallel send/receive using the NBX algorithm from: https://htor.inf.ethz.ch/publications/img/hoefler-dsde-protocols.pdf

That algorithm is similar to "my" algorithm I'm working on - but works better (in my benchmarking) for this case of one-sided, single hop, single data packet, point-to-point transfers.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -3557,7 +3557,6 @@ void DofMap::scatter_constraints(MeshBase & mesh)
       pushed_keys_vals_to_me[pid] = data;
     };
 
-  // Trade pushed dof constraint rows
   Parallel::push_parallel_vector_data
     (this->comm(), pushed_ids_rhss, ids_rhss_action_functor);
   Parallel::push_parallel_vector_data
--- a/src/parallel/communicator.C
+++ b/src/parallel/communicator.C
@@ -197,6 +197,19 @@ void Communicator::barrier () const
 void Communicator::barrier () const {}
 #endif
 
+#ifdef LIBMESH_HAVE_MPI
+void Communicator::nonblocking_barrier (Request & req) const
+{
+  if (this->size() > 1)
+    {
+      LOG_SCOPE("nonblocking_barrier()", "Parallel");
+      libmesh_call_mpi(MPI_Ibarrier (this->get(), req.get()));
+    }
+}
+#else
+void Communicator::barrier () const {}
+#endif
+
 
 MessageTag Communicator::get_unique_tag(int tagvalue) const
 {
--- a/src/parallel/communicator.C
+++ b/src/parallel/communicator.C
@@ -21,6 +21,7 @@
 
 // libMesh includes
 #include "libmesh/libmesh_logging.h"
+#include "libmesh/print_trace.h"
 
 namespace libMesh
 {
@@ -62,7 +63,11 @@ Communicator::Communicator () :
   _size(1),
   _send_mode(DEFAULT),
   used_tag_values(),
-  _I_duped_it(false) {}
+  _next_tag(0),
+  _max_tag(std::numeric_limits<int>::max()),
+  _I_duped_it(false)
+{
+}
 
 
 Communicator::Communicator (const communicator & comm) :
@@ -73,6 +78,8 @@ Communicator::Communicator (const communicator & comm) :
   _size(1),
   _send_mode(DEFAULT),
   used_tag_values(),
+  _next_tag(0),
+  _max_tag(std::numeric_limits<int>::max()),
   _I_duped_it(false)
 {
   this->assign(comm);
@@ -170,6 +177,15 @@ void Communicator::assign(const communicator & comm)
 
       libmesh_assert_greater_equal (i, 0);
       _rank = cast_int<processor_id_type>(i);
+
+      // Get the maximum tag value
+      {
+        MPI_Aint * maxTag;
+        int flag;
+        libmesh_call_mpi(MPI_Comm_get_attr(MPI_COMM_WORLD, MPI_TAG_UB, &maxTag, &flag));
+
+        _max_tag = *maxTag;
+      }
     }
   else
     {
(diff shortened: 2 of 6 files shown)


roystgnr:
The timeout is scary, the NaNs above are scarier.  I'll try to get a look at this shortly.

src/parallel/communicator.C
```diff
@@ -197,6 +197,19 @@ void Communicator::barrier () const
 void Communicator::barrier () const {}
 #endif
 
+#ifdef LIBMESH_HAVE_MPI
+void Communicator::nonblocking_barrier (Request & req) const
+{
+  if (this->size() > 1)
+    {
+      LOG_SCOPE("nonblocking_barrier()", "Parallel");
+      libmesh_call_mpi(MPI_Ibarrier (this->get(), req.get()));
```
roystgnr: MPI_Ibarrier is an MPI3-only method, isn't it?  We almost certainly want to keep MPI-2 capability as a fallback, even if builds with MPI-3 work faster.  Didn't the code you showed me before use non-blocking reductions rather than non-blocking barriers?

friedmud: It did - but this is more efficient.  I do believe that it's MPI-3.

Why do we want to keep MPI-2?  MPI-2 is CRAZY old at this point.  Just like we moved on to C++11... we should move on.

friedmud:
I have no idea why this is failing BTW: I've been using it for days without any issues...

friedmud:
@roystgnr is it possible that some of the `send()` specializations aren't honoring the synchronous send flag?  That would definitely lead to badness!

src/parallel/communicator.C
friedmud: MPI-3 is 2012 BTW... and it's WAY easy to upgrade MPI

roystgnr:
Definitely thought that bug was possible, since IIRC we don't have any test coverage for send_mode(), but grepping through parallel_implementation.h doesn't find anything wrong.  Every send implementation either tests send_mode or hands off to a different send implementation; nothing just blindly calls MPI_Send or MPI_Isend.

friedmud:
Hmmm - thanks for checking.  I don't have time to look into these failures
right now - but I might be able to get to them this week.

friedmud:
@roystgnr check this out.  I think this is good to go now.

I didn't implement the `vector<vector<T>>` case yet.  Do you see a way to unify those two at all - or will it just need to pretty much be a duplication?  I tried to see if I could get a unified one to go... but I couldn't come up with the correct template foo to make it happen.

Note that I changed the way unique tag IDs are chosen now.  The reason is fairly subtle... but basically you need two adjacent calls to push/pull_parallel_vector_data() to use _definitely_ different tag IDs.

I went with @fdkong and looked at what PETSc does for this... and they do something close to what I've coded here (except that they count _down_ and they don't bother to keep a list of currently used ones... mostly just hoping that if they wrap around there are no collisions).

In my testing yesterday this algorithm can make a BIG impact at 200M Dofs on 4k+ procs.

roystgnr:
Thanks!

> @roystgnr check this out.

Will do now, literally.  Multiple git checkouts, even.  I'd like to run on a few different systems and make sure the "works on one system, breaks on another" sort of issues are really gone.

Assuming I can't shake loose any problems, we'll want to merge this ASAP.  But at some point I may be tempted to add back "--sync_algorithm old" and "--sync_algorithm older" command line options for debugging and benchmarking purposes.

roystgnr:
I am seeing hangs.  Going to see if I can get a sweep to trigger them via Civet too.

roystgnr:
Huh.  At least my first hang seems like it's an unrelated bug that my tests just happened to trigger now.  I'll put in a fix in a different PR and keep going.

src/base/dof_map_constraints.C
```diff
@@ -3557,7 +3557,6 @@ void DofMap::scatter_constraints(MeshBase & mesh)
       pushed_keys_vals_to_me[pid] = data;
     };
 
-  // Trade pushed dof constraint rows
```
roystgnr: Why'd this vanish?

roystgnr:
I'm seeing one more failure, but I can't reproduce it everywhere, and I *can* get the same failure from a different branch.  If you want this in master urgently I'm okay with merging now, but I'd prefer to have a little more time to look at it over the holiday.

friedmud:
Interesting - I'm not in any hurry... I'll be banging on this patch myself this weekend.

We were able to run some pretty big jobs with this patch today already... but it needs to be perfect before we put it in so let me know if you see something....

friedmud:
Also: thanks for looking at it!

src/base/dof_map_constraints.C
friedmud: Well - I did have some back and forth on revisions that edited this part of code - just a messup on my part to leave this like this though.  I'll fix it.

src/parallel/communicator.C
```diff
@@ -197,27 +213,40 @@ void Communicator::barrier () const
...
+  auto new_tag = _next_tag++;
+
+  // We don't want to simply get the largest one
+  // because it might be near MPI_TAG_UB and then
+  // we can end up in a sticky situation
+  // Most-likely we will amost never hit a
+  // tag that is in use because we have
+  // ~2 billion of them
+  while(used_tag_values.count(new_tag))
+    new_tag++;
```
roystgnr: Ignoring tagvalue prevents use cases where the user wants to request tags in different orders on different processors; e.g. to use a task management system where ranks grab new bits of work as they complete prior tasks.  We don't actually *do* that or intend to at any time in the near future, but it was the original reasoning.

So what's the reasoning behind removing it?  We already had a "grab the next larger tag if the one we looked at first is already in use" mechanism.  Why didn't it work for your case?

roystgnr:
Well, **damn**, that was a good answer.

And now I see the obvious - the old code was incrementing for ids already in use, then going back to the requested id as soon as it thought it could; the new code is incrementing for ids already in use but that's basically pointless because it *stays* incremented until wrapping around MAX_INT.

I still don't want to break anyone who decided to try a ```isend(tag=1); isend(tag=2); recv(tag=2); recv(tag=1)``` if we don't have to... but now I'm much less confident that we don't have to.

Could we make user tag requests an optional feature?
* Make the get_unique_tag default value be MessageTag::invalid_tag
* If we see invalid_tag we use _next_tag++ (or increment that further if that's already taken)
* If we see a user-requested tag we use it (or increment it further if that's already taken)
* Initialize _next_tag to 1B, so user requests are less likely to conflict with _next_tag
* Make every libMesh internal get_unique_tag() call use the default, so even if we use those tags for asynchronous stuff we have fewer internal "user" requests likely to conflict with user requests.

src/parallel/communicator.C
```diff
@@ -170,6 +177,15 @@ void Communicator::assign(const communicator & comm)
...
       libmesh_assert_greater_equal (i, 0);
       _rank = cast_int<processor_id_type>(i);
+
+      // Get the maximum tag value
+      {
+        MPI_Aint * maxTag;
+        int flag;
+        libmesh_call_mpi(MPI_Comm_get_attr(MPI_COMM_WORLD, MPI_TAG_UB, &maxTag, &flag));
+
+        _max_tag = *maxTag;
```
roystgnr: This appears to be incorrect.  On MPICH 3.2 I get -4026531841 this way, which doesn't fit in int and isn't even positive.  The MPI_Comm_get_attr man page seems to think we need to be passing in a ```void*``` and then casting the pointer to ```int*``` before dereferencing.

I'll do that; I've got a fork of this branch in which I'm trying out my "see if I can retain backwards tagvalue compatibility without breaking Derek's stuff" plan.

src/parallel/communicator.C
```diff
@@ -207,30 +223,30 @@ void Communicator::nonblocking_barrier (Request & req) const
...
+    new_tag++;
 
-  // #ifndef NDEBUG
-  //   // Make sure everyone called get_unique_tag and make sure
-  //   // everyone got the same value
-  //   int maxval = tagvalue;
-  //   this->max(maxval);
-  //   libmesh_assert_equal_to (tagvalue, maxval);
-  // #endif
+  used_tag_values[new_tag] = 1;
```
roystgnr: I dug into SVN history to see why this was commented out, and apparently Ben had a use case in which it was expected to fail, because tags were being grabbed on a subset of processors.  So I feel a little better about going to the effort of respecting tagvalue where it's used while avoiding using it where possible.

Actually, that two-part system ought to make sanity checking easier.  Where the user requests an automatic tag value, we can be sure they want to keep _next_tag in sync, and we can restore this check.

roystgnr:
This algorithm made it in via #1965

PR #1831, opened 2018-08-23
Problem with degenerate BoundingBox intersection

jwpeterson:
We have a case where the following bounding boxes (coordinates printed in`%f` format) were found not to intersect:
```
1: Min: (x,y,z)=(       0,  51.6667,        0), Max: (x,y,z)=( 6.66667,  63.3333,        0) 
2: Min: (x,y,z)=(       0,  51.6667,        0), Max: (x,y,z)=( 6.66667,  63.3333,        0) 
```
At first I thought it was something weird with the (x,y) coordinates being slightly off, but after a bit of testing, the problem is definitely that there is some floating point noise in the z-coordinates of the boxes that causes them to not intersect (basically they are two flat planes separated by a small distance epsilon). The unit test in this PR demonstrates the issue.

It seems to me that we should ignore the intersection test results in dimensions where the bounding box is nearly degenerate, but that will involve a floating point tolerance check. Since there aren't any floating point tolerances in the `BoundingBox` right now, I thought I'd start a PR to discuss the right approach.


tests/geom/bbox_test.C
```diff
@@ -0,0 +1,96 @@
...
+      BoundingBox lhs(/*min=*/Point(eps, c1+eps, 0.), /*max=*/Point(c2-eps, c3-eps, 0.));
+      BoundingBox rhs(/*min=*/Point(0., c1, 0.), /*max=*/Point(c2, c3, 0.));
+      CPPUNIT_ASSERT(lhs.intersects(rhs));
+      CPPUNIT_ASSERT(rhs.intersects(lhs));
+    }
+    {
+      // Test with slightly perturbed (upward) lhs (zmin, zmax).
+      BoundingBox lhs(/*min=*/Point(0., c1, 0.+eps), /*max=*/Point(c2, c3, 0.+eps));
+      BoundingBox rhs(/*min=*/Point(0., c1, 0.), /*max=*/Point(c2, c3, 0.));
+      CPPUNIT_ASSERT(lhs.intersects(rhs));
```
jwpeterson: The failing test is here.

roystgnr:
Technically that failing test is a test we expect to fail, right?  So the problem isn't that ```BoundingBox::intersects()``` is giving wrong results, it's that we're using an exact ```BoundingBox::intersects()``` in some context where we want to be asking a different question.

If it's an issue of some epsilon, it still seems like it should be easier than the ```Elem::point_test()``` case since we don't have any inexact inverse map calculation to worry about.  I'd be happy with a ```intersects(bbox, tol)``` overload when applies tol to the calculation.  It'd have to be an absolute rather than a relative tolerance, which needs to be documented because that's only an obvious choice for this particular case.

The alternative useful thing to add might be a ```projection_intersects(bbox)``` method, which ignores any degenerate-in-both-boxes dimensions.  That would be another exact operation, just differently defined.

I couldn't say which is preferable without knowing where the non-intersecting-but-doing-something-else-we-want-to-know-about pair of bounding boxes came from.  If it's possible that a "degenerate dimension" might really have a thickness of epsilon rather than just a displacement of epsilon then the ```projection_intersects``` idea won't work.

jwpeterson:
> I couldn't say which is preferable without knowing where the non-intersecting-but-doing-something-else-we-want-to-know-about pair of bounding boxes came from.

I believe the application here was grain tracking in phase-field simulations with periodic BCs on a Cartesian domain. The bounding boxes are being built on the boundary of the domain, hence the "degeneracy". 

> The alternative useful thing to add might be a projection_intersects(bbox) method, which ignores any degenerate-in-both-boxes dimensions. That would be another exact operation, just differently defined.

By exact do you mean the "width" of the box is `== 0.0` using floating point arithmetic?

`intersects(bbox, abstol=0.0)` seems like a reasonable approach.

roystgnr:
> The bounding boxes are being built on the boundary of the domain, hence the "degeneracy".

One bounding box is being built with actual nodes on the z=0 boundary, and the other bounding box is being built with the nodes on the periodic boundary and then translated by the domain width-in-z?  But the domain width isn't exactly correct so there's epsilon difference in the z direction?  And we never noticed this before because the other periodic boundary code uses element locator methods that have a tolerance and handle being epsilon off?

> By exact do you mean the "width" of the box is == 0.0 using floating point arithmetic?

Right.  But it sounds like this isn't a safe approach in your case - one node perturbation and suddenly an exactly-degenerate box is no longer technically degenerate, and if the node perturbation epsilon doesn't exceed your translation arithmetic epsilon then your bounding boxes no longer intersect.

> intersects(bbox, abstol=0.0) seems like a reasonable approach.

Agreed.  A separate overload would be more efficient than a default value.  Yes, a priori I'd expect the costs of adding and subtracting a bunch of zeros to be trivial, but if you're hitting ```intersects()``` a lot then my expectations might be wrong.

PR #1833, opened 2018-08-24
Added unit test for DofCoupling with Variable Groups

dknez:
This unit test currently fails due to a sparsity pattern error. This error is apparently due to the interplay of DofCoupling and Variable Groups. If you change the "theta" variables to not use subdomains then all variables are part of one Variable Group and the test passes.

### DIFF
--- a/tests/systems/systems_test.C
+++ b/tests/systems/systems_test.C
@@ -17,6 +17,10 @@
 #include "libmesh/string_to_enum.h"
 #include <libmesh/cell_tet4.h>
 #include <libmesh/zero_function.h>
+#include <libmesh/linear_implicit_system.h>
+#include <libmesh/quadrature_gauss.h>
+#include <libmesh/node_elem.h>
+#include "libmesh/edge_edge2.h"
 
 #include "test_comm.h"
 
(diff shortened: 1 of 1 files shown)


dknez:
@roystgnr could you please have a look at this unit test that I added? As I described above it currently fails, but can be made to work by changing the code so that all variables are in the same variable group.

dknez:
P.S. I'm not sure why the tests aren't running, "Won't run due to failed dependencies"?

pbauman:
From femputer side (sorry if you can't see it because eof self-signed certificate, that's still not sorted yet). Looks like you have a warning in the test that's tripping -Werror in the libMesh build stage.

<pre>
CXX      src/fe/libmesh_devel_la-fe_bernstein_shape_1D.lo
/femputer/pbauman/civet_build_testing_root/libMesh/tests/systems/systems_test.C: In member function ‘void SystemsTest::testDofCouplingWithVarGroups()’:
/femputer/pbauman/civet_build_testing_root/libMesh/tests/systems/systems_test.C:571:18: error: unused variable ‘u_var’ [-Werror=unused-variable]
     unsigned int u_var = system.add_variable ("u", libMesh::FIRST);
                  ^~~~~
/femputer/pbauman/civet_build_testing_root/libMesh/tests/systems/systems_test.C:572:18: error: unused variable ‘v_var’ [-Werror=unused-variable]
     unsigned int v_var = system.add_variable ("v", libMesh::FIRST);
                  ^~~~~
/femputer/pbauman/civet_build_testing_root/libMesh/tests/systems/systems_test.C:573:18: error: unused variable ‘w_var’ [-Werror=unused-variable]
     unsigned int w_var = system.add_variable ("w", libMesh::FIRST);
                  ^~~~~
/femputer/pbauman/civet_build_testing_root/libMesh/tests/systems/systems_test.C:577:18: error: unused variable ‘theta_x_var’ [-Werror=unused-variable]
     unsigned int theta_x_var = system.add_variable ("theta_x", libMesh::FIRST, &theta_subdomains);
                  ^~~~~~~~~~~
/femputer/pbauman/civet_build_testing_root/libMesh/tests/systems/systems_test.C:578:18: error: unused variable ‘theta_y_var’ [-Werror=unused-variable]
     unsigned int theta_y_var = system.add_variable ("theta_y", libMesh::FIRST, &theta_subdomains);
                  ^~~~~~~~~~~
/femputer/pbauman/civet_build_testing_root/libMesh/tests/systems/systems_test.C:579:18: error: unused variable ‘theta_z_var’ [-Werror=unused-variable]
     unsigned int theta_z_var = system.add_variable ("theta_z", libMesh::FIRST, &theta_subdomains);
</pre>

dknez:
I fixed the unused variable warnings, thanks.

dknez:
Also, I see that the Civet Precheck says "The following files contain trailing whitespace after applying your patch: tests/systems/systems_test.C". I don't see any trailing whitesapce in my diff though?

jwpeterson:
> Also, I see that the Civet Precheck says "The following files contain trailing whitespace after applying your patch: tests/systems/systems_test.C". I don't see any trailing whitesapce in my diff though?

git diff --check 2b2196d..408602d
tests/systems/systems_test.C:590: trailing whitespace.
+    // We set the solution to be 1 everywhere, so the final l1 norm of the

dknez:
Thanks. I was looking for it in the github diff, which normally highlights trailing whitespace, but I guess it doesn't do that for comments. Should be fixed now.

roystgnr:
Can this be simplified any further on the variables side?  Do you need 3 of each type to trigger the problem?

dknez:
I simplified to to use 2 variables instead of 6. It triggers the same error in that case.

roystgnr:
Thanks!  Don't have time to fix this right now but I'll at least see if it replicates.

roystgnr:
Well, it fails with DistributedMesh on 24 processors for obvious reasons.  Let's try it on 1...

roystgnr:
Handmaking a distributed mesh is trickier than this, but that was a small fix.  And yeah, I see a malloc out of range now.

Thanks for setting this up; hopefully it's small enough to figure out what's going on.

dknez:
@roystgnr have you had a chance to look into this at all? I'd be happy to help out with any debugging, if there's anything I can help with.

roystgnr:
Just finished looking at it, and at first glance it looks like NOTABUG.  Let me know if I'm wrong:

We have eight degrees of freedom here: u has DoFs 0,1,2,3; theta has DoFs 4,5,6,7

We have four elements here: elem0 is an Edge2 supporting DoFs 0,1,4,5; elem1 is an Edge2 supporting DoFs 2,3,6,7; elem2 is a NodeElem supporting DoF 1, and elem3 is a NodeElem supporting DoF 2.

Note that, although Node 1 defines the position of elem2, and Node 1 can be queried to uncover that it supports DoFs 1 and 5, that *doesn't* mean elem2 supports DoF 5 - the variable subdomain setting specifically told it not to!

When we build our sparsity pattern:
All the DoFs within each element are coupled to each other.  Additionally, the ghosting functor you've added couples elem2 with elem3, so there should be a coupling from DoF1 to DoF2.  So our n_nz on one processor should look like [4,5,5,4,4,4,4,4], which is what we see.  Victory!

Except, in your assembly function:
You try to add Jacobian entries coupling DoF 5 to 6 and 2, and coupling DoF 6 to 1 and 5.  But there's no ghosting functor that suggested such coupling existed.  You coupled the two NodeElem elements, but they aren't in a subdomain for which theta is active, so we've explicitly said they don't have any theta support!  Trying to couple theta anyway means you're trying to fill a matrix with n_nz [4,6,6,4,4,6,6,4], so naturally PETSc screams when you overflow the n_nz libMesh provided it.

There doesn't seem to be any VariableGroup interaction here - once you got rid of the identical variables there were only two one-variable-each groups left.

roystgnr:
So anyway this feels like we've hit XY Problem territory - you're going to extra work to disable variable on the elements you're adding coupling for, and you're trying to couple them anyway, and the solution for the helpfully-distilled-toy-problem case is "don't disable those variables on the coupled elements"... but presumably there's some *other* reason they're disabled that just got distilled away when you made the problem small enough to make sense of in gdb, in which case the solution for the toy problem isn't a sufficient solution for the real case?

dknez:
OK, thanks for this info, this makes sense!

The issue here was my misunderstanding of how the coupling would work, I figured that coupling the NodeElems would couple all the dofs on those NodeElems, but that is incorrect. The DofCoupling uses the active-variables-on-subdomain info in the coupling, as it should, and hence it identifies the theta variables as inactive on the NodeElems and doesn't couple them.

So I agree, there is no bug here, and I will update my non-toy problems accordingly (e.g. by making sure that theta is active on the NodeElems).

Separate question: Do you want to close this PR, or keep the extra unit test (after making an update to ensure that it passes)?

roystgnr:
Honestly, the more coverage the merrier.  Take away the theta terms in the assembly function, maybe add back the other 4 variables, and it'd be great to have this in the test suite.  Oh, and also either manually force ReplicatedMesh, or if you're feeling brave you can try to fix it on DistributedMesh.  ;-)

roystgnr:
:+1: Thanks!

Can you think of any way to improve the ghosting functor documentation to make the coupling definition clearer?  Or should I just add an extra separate paragraph about overlapping elements?

dknez:
The documentation is already detailed, which is good, but I think an extra paragraph about overlapping elements would be helpful.

pbauman:
> Can you think of any way to improve the ghosting functor documentation to make the coupling definition clearer? Or should I just add an extra separate paragraph about overlapping elements?

Since you asked, one thing that would've helped me immensely is pointing out where one plugs in to get a ghosting functor vs. algebraic functor vs. a coupling functor (that you explained to me in telecon). The description of the types of coupling was very clear to me though.

permcody:
Ghosting Functors plug into the mesh. Algebraic/Coupling Functors plug into
the Dof_map. They all use the same interface but the associated coupling
matrix is ignored in the Ghosting functor interface.

roystgnr:
@dknez, @pbauman, check out be47f1c and let me know if it's clear enough?  Kind of terse but I fear that if I make that encyclopedic comment too much longer then users will just tune it out.

pbauman:
> check out be47f1c and let me know if it's clear enough?

Works for me. Thanks!

dknez:
Looks good, thanks!

PR #1834, opened 2018-08-25
RB change: improve how we handle degenerate sides

dknez:
Previously we got many warnings on meshes with degenerate sides. If skip_degenerate_sides is true, this change first checks if a side is degenerate and skips it if necessary.


dknez:
From what I can tell this has passed all the tests. It looks like "Test debug" is stuck on "Pending" even though it has passed all the tests.

Also, this is a very minor change and only affects the reduced basis code, so I'll go ahead and merge.

PR #1836, opened 2018-08-27
Only build discontinuous solution vector when variable is active on subdomain

lindsayad:
Should hopefully resolve issue experienced on mailing thread here: https://sourceforge.net/p/libmesh/mailman/message/36400029/

### DIFF
--- a/src/systems/equation_systems.C
+++ b/src/systems/equation_systems.C
@@ -1143,6 +1143,8 @@ void EquationSystems::build_discontinuous_solution_vector (std::vector<Number> &
                         }
                       }
                     }
+                    else
+                      nn += elem->n_nodes();
                   }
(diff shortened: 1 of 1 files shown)


jwpeterson:
Looks reasonable to me, thanks @lindsayad!

lindsayad:
Renato's exodus output:
lmerr

My exodus output:
pview

lindsayad:
So stupid...I had overload a `vtkBlockColors` plot over the `p` plot. I'm now seeing the same results as Renalto.

Do not merge as is!

lindsayad:
I think this looks more like it...
pview

lindsayad:
Renato has confirmed that this works for him, so this is good to go

PR #1837, opened 2018-08-27
Warning message fix

roystgnr:
Thanks to @jwpeterson for catching this:

"I do appreciate the humor of warning you that there are too many warnings about compiler warnings just being warnings"


jwpeterson:
Looks good to me, I now get one copy of each status message:
```
<<< Compiler warnings are just warnings >>>
<<< Disabling extra paranoid compiler warnings >>>
```

PR #1838, opened 2018-08-27
Small change to Reduced Basis logging

dknez:
Move "quiet" flag from RBConstruction to RBConstructionBase, and skip printing out some (usually excessive) logging in RBConstructionBase by default.


dknez:
This is a very small change and only affects the reduced basis code, so I'll go ahead and merge

PR #1839, opened 2018-08-28
Optimization for apply_dirichlet_impl()

dknez:
Currently the code does a bunch of work on elements that do not have any dirichlet constraints. It's faster if we check for the presence of constraints and skip if none are present.


dknez:
I saw a 10% speedup in some cases. Those cases were quite unusual though, e.g. I was creating a bunch of vectors using constraints with enforce_constraints_exactly, so I was setting a lot of different Dirichlet constraints.

roystgnr:
Thanks!

Issue #1753, opened 2018-06-20
Optimize DofObject::var_to_vg() uses


roystgnr:
This is showing up a lot in profiling by @friedmud, and I've seen it show up previously though I forget where.  I guess even without a lot of variable groups to loop over, the test-and-jump instructions in that function cause slowdowns?  He could probably experiment with __builtin_expect and see if that's the case.

Our two current ideas for fixing this brainstormed on the mailing list:

Add an alternative to ```DofObject::dof_number()``` which takes the variable group and variable index-within-group rather than the variable index-within-system, so the lookup from latter to former can be done once per loop rather than once per ```dof_number()``` call.

Cache the lookup within ```DofMap```, then add a ```DofObject::dof_number()``` overload which takes a DofMap reference and uses that cache to avoid the loop.

I'll try those.  Hopefully @friedmud can come up with a test case that exposes the problem heavily enough so I can quantify results without me having to hassle him to try his app code on each iteration...

friedmud:
Update on this - so, as I mentioned in email I need the _first_ dof for a group of dofs (then I do my own manual indexing past that).  It also happens to be that I know that there is only one variable group in this (actually in _two_ systems)... so I know that the variable group ID is: `0`

So... to get the first dof on a DofObject I replaced my call to `dof_number(sys, 0, 0)` and replaced it with `vg_dof_base(sys, 0)` (where that `0` is the `var_group`).  This is WAY faster... and that whole thing is gone from profiling and my code is 4-5% faster.

Not huge gains I agree - but more than anything it was just bothering me that 4-5% of my total runtime was going to such a simple function as getting the first dof for a system on a DofObject.

This trick won't work too well for MOOSE in general though - because it needs to get dofs for individual variables within variable groups - but we can definitely store off which variable group each variable is assigned to and call a version of `dof_number()` that takes in a `var_group` as a short-cut.  I definitely think that would fix it.

roystgnr:
#1842 should resolve this in a more general way.

PR #1842, opened 2018-08-29
DofObject::dof_number alternative

roystgnr:
This still has quite a bit more work incoming, but since it's a major refactoring of one of the most basic and pervasive libMesh functions, I'm thinking we'll test it one optimization step at a time.

The original motivation comes from @friedmud via #1753 but anything with noticeable time spent in dof_indices() (including some of the RELAP-7 cases, IIRC?) ought to benefit from some of the later optimization stages.

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -1279,19 +1279,24 @@ private:
 
   /**
    * Helper function that gets the dof indices on the current element
-   * for a non-SCALAR type variable.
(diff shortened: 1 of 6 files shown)


roystgnr:
This seems to be passing CI but I think I'm seeing some test failures in private tests...

roystgnr:
Nope; test failures were red herrings from a misconfiguration in my test environment.  This should be ready to go, although I think I'll try to use the optimized method in one or two more places first.

PR #1843, opened 2018-08-30
AbaqusIO: Add more Abaqus -> libMesh element type maps

jwpeterson:
I'm not sure why we never ran into this problem before, this issue was
raised in a recent thread on the MOOSE mailing list:

https://groups.google.com/d/msg/moose-users/T0XOeTbByXE/ym41oODEAgAJ

### DIFF
--- a/src/mesh/abaqus_io.C
+++ b/src/mesh/abaqus_io.C
@@ -521,7 +521,8 @@ void AbaqusIO::read_elements(std::string upper, std::string elset_name)
       elems_of_dimension[2] = true;
     }
   else if (upper.find("CPS3") != std::string::npos ||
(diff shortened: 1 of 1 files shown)


roystgnr:
Should we be adding more while we're at it?  There's a whole list in http://www2.mae.ufl.edu/haftka/finite-element/ABAQUS-elements.pdf of geometric-and-finite-element types, many of which we're missing in our tests.

roystgnr:
:+1: for this change, of course, just wondering if we can preempt future similar problems.

jwpeterson:
> many of which we're missing in our tests.

Wow, good find. Yeah we should probably add others...

PR #1845, opened 2018-08-30
Make metaphysicl a submodule

lindsayad:
I'd like to make metaphysicl a submodule to incorporate upstream changes as quick as possible for downstream automatic differentiation support in MOOSE


lindsayad:
Looks like I'll need to run `boostrap` in metaphysicl

lindsayad:
Can someone tell me what's failing with GRINS? I was able to build/test/install GRINS in `dbg` mode locally

roystgnr:
I wouldn't blame INL firewall rules for this one; "don't connect to sites with invalid SSL certificates" is actually a pretty decent rule.

The GRINS tests are failing in the Fetch and Branch step!

With a bunch of other "Removing contrib/metaphysicl/0.2.0" foo omitted,

```

BUILD_ROOT/libMesh/: git reset --hard
warning: unable to rmdir contrib/metaphysicl: Directory not empty
HEAD is now at d40935c Merge pull request #1843 from jwpeterson/abaqus_CPE3
BUILD_ROOT/libMesh/: git clean -xffdq
BUILD_ROOT/libMesh/: git fetch origin
Warning: No xauth data; using fake authentication data for X11 forwarding.

BUILD_ROOT/libMesh/: git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
BUILD_ROOT/libMesh/: git reset --hard origin/master
HEAD is now at d40935c Merge pull request #1843 from jwpeterson/abaqus_CPE3
BUILD_ROOT/libMesh/: git reset --hard
HEAD is now at d40935c Merge pull request #1843 from jwpeterson/abaqus_CPE3
BUILD_ROOT/libMesh/: git clean -xffdq
BUILD_ROOT/libMesh/: git submodule foreach --recursive git reset --hard
BUILD_ROOT/libMesh/: git submodule update --init --recursive
BUILD_ROOT/libMesh/: git submodule foreach --recursive git clean -xffdq
BUILD_ROOT/libMesh/: git submodule status --recursive
BUILD_ROOT/libMesh/: git checkout master
Already on 'master'
Your branch is up-to-date with 'origin/master'.
BUILD_ROOT/libMesh/: git reset --hard origin/master
HEAD is now at d40935c Merge pull request #1843 from jwpeterson/abaqus_CPE3
BUILD_ROOT/libMesh/: git submodule update --init --recursive
BUILD_ROOT/libMesh/: git branch -D test
Deleted branch test (was d40935c).


Checking out git@github.com:lindsayad/libmesh.git:metaphysicl-submodule:6e510938bc56713299f5382491ee77feb5396ad2
BUILD_ROOT/libMesh/: git checkout -b test origin/master
Switched to a new branch 'test'
Branch test set up to track remote branch master from origin.
BUILD_ROOT/libMesh/: git pull git@github.com:lindsayad/libmesh.git metaphysicl-submodule
Warning: No xauth data; using fake authentication data for X11 forwarding.

From github.com:lindsayad/libmesh
 * branch            metaphysicl-submodule -> FETCH_HEAD
Removing contrib/metaphysicl/0.2.0/test/vector_pde_unit.C
...
Removing contrib/metaphysicl/0.2.0/AUTHORS
CONFLICT (directory/file): There is a directory with name contrib/metaphysicl in HEAD. Adding contrib/metaphysicl as contrib/metaphysicl~6e510938bc56713299f5382491ee77feb5396ad2
Automatic merge failed; fix conflicts and then commit the result.
ERROR: exiting with code 1
```

pbauman:
> I wouldn't blame INL firewall rules for this one; "don't connect to sites with invalid SSL certificates" is actually a pretty decent rule.

Sorry about that. I'm currently low priority on this with our IT people. I'm hoping I can get their attention next week about a proper cert.

> The GRINS tests are failing in the Fetch and Branch step!

This recipe should be a straight copy of some original CIVET recipes, but let me know if I need to do something different for the submodules since I don't use them ever.

Also, wasn't there a discussion at some point about basically "avoiding submodules if you can" (maybe  I'm misremembering). Is the switch here to just make it easier to suck in changes upstream from MetaPhysicL?

permcody:
On Wed, Sep 5, 2018 at 12:40 PM Paul T. Bauman <notifications@github.com>
wrote:

> I wouldn't blame INL firewall rules for this one; "don't connect to sites
> with invalid SSL certificates" is actually a pretty decent rule.
>
> Sorry about that. I'm currently low priority on this with our IT people.
> I'm hoping I can get their attention next week about a proper cert.
>
> The GRINS tests are failing in the Fetch and Branch step!
>
> This recipe should be a straight copy of some original CIVET recipes, but
> let me know if I need to do something different for the submodules since I
> don't use them ever.
>
> Also, wasn't there a discussion at some point about basically "avoiding
> submodules if you can" (maybe I'm misremembering). Is the switch here to
> just make it easier to suck in changes upstream from MetaPhysicL?
>

We love submodules. We use them in nearly every project since they
basically version the projects together so you can be assured that you have
a working "coupled" piece of software at all times. The one hangup you can
run into is that if you need to use one URL method over another (because
only one type works) then using an absolute URL for the submodule can be
problematic (e.g. I setup SSH keys and clone via SSH, but my submodule was
added absolutely was HTTPS). It turns out that you can use relative paths
for submodules if projects are hosted in the same place so even that
problem can be resolved. The CIVET recipe should automatically handle
checking out and syncing the submodule if one is used in your project.

brianmoose:
We get the same merge conflicts: https://civet.inl.gov/job/226594/
It doesn't like that the existing contrib/metaphysicl is getting replaced by a submodule.
Not sure why it is only happening on some of our build boxes.
Using a "-s resolve" instead of the default recursive strategy seems to work but not sure what the best, most robust, solution is.

brianmoose:
The reason the conflict only happened on some build boxes seems to be the existence of the file `contrib/metaphysicl/.git`. This file gets produced after doing a submodule init. 
So build machines that ran previous versions of this PR with the submodule update would fail on subsequent runs. 
It doesn't seem to be cleared by a `git clean -xffdq`. Since civet deinits submodules before doing stuff, I just removed .git files (we need the .git directory of course). Hopefully that won't cause any problems...

pbauman:
OK, so I'll need to login to my CIVET server and remove the contrib/metaphysicl/.git file and then invalidate?

jwpeterson:
I believe we are also finding that even after checking in the build products, metaphysicl does not build on systems without autotools installed somewhere in the user's path. One one of our build boxes, [the error is](https://civet.inl.gov/job/226594/):
```
CDPATH="${ZSH_VERSION+.}:" && cd ../../../contrib/metaphysicl && /bin/bash /opt/civet/build_0/libmesh/contrib/metaphysicl/build-aux/missing aclocal-1.15 -I m4 -I m4/common
/opt/civet/build_0/libmesh/contrib/metaphysicl/build-aux/missing: line 81: aclocal-1.15: command not found
WARNING: 'aclocal-1.15' is missing on your system.
         You should only need it if you modified 'acinclude.m4' or
         'configure.ac' or m4 files included by 'configure.ac'.
         The 'aclocal' program is part of the GNU Automake package:
         <http://www.gnu.org/software/automake>
         It also requires GNU Autoconf, GNU m4 and Perl in order to run:
         <http://www.gnu.org/software/autoconf>
         <http://www.gnu.org/software/m4/>
         <http://www.perl.org/>
Makefile:597: recipe for target '../../../contrib/metaphysicl/aclocal.m4' failed
make[2]: *** [../../../contrib/metaphysicl/aclocal.m4] Error 127
```

I could have sworn that this was due to the whole `AM_MAINTAINER_MODE` thing that we figured out back in 7619611a, but metaphysicl's configure.ac looks the same as libmesh's in this regard.

roystgnr:
Wait, how is that the same?  MetaPhysicL doesn't use AM_MAINTAINER_MODE, regardless of arguments.

jwpeterson:
Line 16 of your configure.ac, at least in 0.2.0, is
```
AM_MAINTAINER_MODE # git trashes timestamps, forcing autoreconf needlessly
```
which is apparently equivalent to `AM_MAINTAINER_MODE([disable])`.

brianmoose:
@pbauman Yes, that should do it. It might continue to be a problem until this gets merged, unless you put something in your scripts.

roystgnr:
That's in 0.2.0 in the imported version in libMesh.  I never turned that on in the main repo, so it's not on in the submodule.

roystgnr:
I can replicate the problem if I uninstall automake.  If I can't find a better way to fix it then we'll just add "turn on maintainer mode" to the set of stuff that gets done to change e.g. v0.3.0 to v0.3.0_bootstrapped

roystgnr:
Ah, and I'd forgotten about the "git trashes timestamps" thing.  So AM_MAINTAINER_MODE it is.

jwpeterson:
OK, good. If it wasn't that then I didn't have any other guesses about what the problem might be.

roystgnr:
Just force pushed an updated tag.  Let's try it again?

jwpeterson:
> Just force pushed an updated tag. Let's try it again?

I assume you are referring to `v0.3.0_bootstrapped` roystgnr/MetaPhysicl@390a660 and not the original v0.3.0, which still appears to be there?

roystgnr:
Right - v0.3.0 is the "here's version 0.3.0" commit, and v0.3.0_bootstrapped is the "here's version 0.3.0 plus a bunch of autogenerated files that don't belong in version control" commit.

I wouldn't get too attached to roystgnr/MetaPhysicL@390a660 though, since the git hash will change once I force push a fix to that comment...

roystgnr:
There we go.  roystgnr/MetaPhysicL@b4169be is it now.

jwpeterson:
> There we go. roystgnr/MetaPhysicL@b4169be is it now.

I got the force-pushed `branch_0.3.0`, but the `v0.3.0_bootstrapped`tag still points to the previous head. If you want to delete the old tag to avoid confusion, see, for example, [this gist](https://gist.github.com/mobilemind/7883996).

roystgnr:
Seriously?  I can't just delete a tag locally and "git push --force --tags"?

Ok, I changed the subsequent commit message while I'm at it (although it's stupidly ambiguous either way...) and now roystgnr/MetaPhysicL@1779e10 should be the tag target.

jwpeterson:
> Seriously? I can't just delete a tag locally and "git push --force --tags"?

I'm reading (some stuff)[https://stackoverflow.com/questions/25849019/move-tag-one-commit-ahead] now that suggests that will work... I might have been using an older git client or something.

> now roystgnr/MetaPhysicL@1779e10 should be the tag target.

Doh, @lindsayad had already restarted the testing on the other one...

roystgnr:
> Doh, @lindsayad had already restarted the testing on the other one...

And we'll need to switch because now the other one is orphaned and will eventually get garbage collected, right?  Sorry, that wasn't worth it for a dubious commit message change.

At least we'll want to do something to kick the CI anyway; GRINS is now saying
```
CONFLICT (directory/file): There is a directory with name contrib/metaphysicl in HEAD. Adding contrib/metaphysicl as contrib/metaphysicl~ad8aaa1e1a83c634fb2ea4917fd91591338fa7e4
```

jwpeterson:
I think our only choice is to burn the GRINS CI server to the ground/nuke it from orbit and start again.

lindsayad:
K, will update to the new commit. `tag: v0.3.0_bootstrapped` and `branch_0.3.0` still point to different commits for what it's worth

lindsayad:
K, we either have to update our CIVET recipe to `git fetch --tags` or the tag needs to be included in a branch history...perhaps both should be done...

roystgnr:
> tag: v0.3.0_bootstrapped and branch_0.3.0 still point to different commits for what it's worth

TIL that "git push --tags" means "--only-tags", not "--also-tags".  It's either fixed or even-further-screwed-up now.

lindsayad:
`git fetch --tags` is the opposite behavior 🙄

jwpeterson:
The last time I did a libmesh update, I had to `--disable-metaphysicl` in MOOSE's configure because it would not build with our minimum-clang target. I attached the min-clang target here, but it won't actually tell us anything... What we need is a MOOSE PR (after this one is merged) that does a libmesh update and removes `--disable-metaphysicl` from the `update_and_rebuild_libmesh.sh` script, to make sure that works with min-clang.

pbauman:
> I think our only choice is to burn the GRINS CI server to the ground/nuke it from orbit and start again.

It basically sounds like I'm going to have to do that after this gets merged.

lindsayad:
So @pbauman, are you good with merging this?

jwpeterson:
@roystgnr I think this is good to go now if you are OK with it. We still don't have feedback from GRINS CI, but I think we have confirmed with local builds that it does build...

jwpeterson:
> Actually I think we still need to test whether libmesh's configure works both with and without the metaphysicl submodule init'd (automatically disabling it in the latter case).

Yeah, this won't work as-is, since libmesh_metaphysicl.m4 *doesn't actually test anything* when you configure with `--enable-metaphysicl`, it just assumes that it will be there. Without even initializing the new submodule, I get:
```
<<< Configuring library with MetaPhysicL support >>>
```
We didn't see this in any of our testing because we explicitly `--disable-metaphysicl` in MOOSE, but an error should be triggered if I attach a `--disable-optional` build to this PR.

jwpeterson:
Since the GRINS tests presumably explicitly or implicitly `--enable-metaphysical`, they should also fail for the same reason (once they get back online).

jwpeterson:
> an error should be triggered if I attach a --disable-optional build to this PR.

Nope, I'm wrong, `--disable-optional` disables metaphysicl so this doesn't test anything new... on my own machine I can confirm I simply get:
```
Making all in metaphysicl
make[2]: *** No rule to make target `all'.  Stop.
```
so we will need to add a test for this.

roystgnr:
Gah, yes, sorry about the weak libmesh_metaphysicl.m4.  I figured there was no need to test for the presence of something in contrib and the idea of using an external install (much less a submodule) was way down on my TODO list.  Unless there's some .git_always_sync_these_submodules setting we can fiddle with, we're going to need a test now.

jwpeterson:
One problem existing users will run into when they try to pull down the submodule. For me there is build junk sitting around in contrib/metaphysicl/0.2.0 (autom4te.cache and other files): 
```
$ git submodule update --init
fatal: destination path 'contrib/metaphysicl' already exists and is not an empty directory.
Clone of 'git@github.com:roystgnr/MetaPhysicL' into submodule path 'contrib/metaphysicl' failed
```
Running
```
rm -rf contrib/metaphysicl/0.2.0/
```
fixes the issue for me, however.

jwpeterson:
I'm pretty sure there's no way to test the normal user's experience (git pull and then no submodule update) with Civet because it is very thorough and always does: 
```
git submodule sync --recursive
git submodule update --init
```
So you'll just have to trust me that the new test works to automatically disable metaphysicl in that situation 😉 . 

I'm going to remove the `--disable-metaphysicl` flag from the min-clang test and try invalidating that one before I merge this PR, however.

pbauman:
> Since the GRINS tests presumably explicitly or implicitly --enable-metaphysicl, they should also fail for the same reason 

I use `--enable-everything` which I presume enables metaphysicl.

> (once they get back online).

It's running, just test this branch is FUBAR'ed because I would need to wipe out the contrib/metaphysicl directory before the git submodule init goes (I think). I was thinking that once this is merged, I'll deal with getting a clean update master tree with metaphysicl submodule.

jwpeterson:
The min-clang (3.5.1) compiler doesn't seem to play nicely with the `-std=c++1y` flag that is used by metaphysicl:
```
/opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/cstdio:120:11: error: no member named 'gets' in the global namespace
  using ::gets;
        ~~^
In file included from ../../../../contrib/metaphysicl/src/utilities/src/metaphysicl_version.C:29:
In file included from ../src/utilities/include/metaphysicl/metaphysicl_version.h:47:
In file included from /opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/iostream:39:
In file included from /opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/ostream:38:
In file included from /opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/ios:42:
In file included from /opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/bits/ios_base.h:41:
In file included from /opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/bits/locale_classes.h:40:
In file included from /opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/string:52:
In file included from /opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/bits/basic_string.h:2815:
In file included from /opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/ext/string_conversions.h:43:
/opt/min_moose/gcc-4.8.4/lib/gcc/x86_64-unknown-linux-gnu/4.8.4/../../../../include/c++/4.8.4/cstdio:120:11: error: no member named 'gets' in the global namespace
  using ::gets;
        ~~^
```

The stack overflow article @lindsayad found suggests the issue is caused by using GCC's STL with the clang compiler, and suggests using clang's own "libc++" implementation will fix the problem. This seems like it would probably work, but from memory I believe we had trouble building clang's libc++ back in the day, and the franken-compiler we ended up with was the best we could do. I don't know if it's really worth bending over backwards to support this old version any longer, my preference would be to bump the minimum clang version to 3.9.0 or so.

jwpeterson:
@roystgnr (Apologies as I think I have already asked you some of this before.)

It seems that using metaphysicl within libmesh is (currently) possible with only the `-std=c++11` flags currently required by libmesh. On the other hand, there are some unit tests (?) and utilities in metaphysicl that require a `-std=c++14` compatible compiler in order to build. 

Therefore, it may be possible to *use* metaphysicl on systems/compilers where it is not possible to *build* all the metaphysicl tests and utilities. This situation currently causes the *entire* libmesh build to fail, even though it seems like it might actually work OK were it not for these extra utilities. This is also what is currently happening our min-clang environment if I understand correctly. Would it be possible to restrict the tests/utilities that are run to a certain standard, like C++11? This might make it easier to use the library overall...

roystgnr:
Damn it, yes, that's possible, and I was almost going to suggest it, except that I was greedily hoping you'd be up for just bumping your clang/std:: requirements instead.

Let me give it a shot.

jwpeterson:
> Let me give it a shot.

OK, we are also still going to upgrade our min-clang environment anyway; I think doing both things is useful.

roystgnr:
Hmm... while I'm at it I ought to add a test that actually triggers the ::gets error.  A --disable-cxx14 option is a good idea in general but there's no reason to force users to use it in a case where we can autodetect the necessity.

roystgnr:
76d762c should handle both the gets error check and manual --disable-cxx14 options.  If it actually works then we can use it as-is; it's currently the v0.3.0.1_bootstrapped tag.

jwpeterson:
```
BUILD_ROOT/libmesh/: git submodule update --init
fatal: reference is not a tree: 76d762c8b086617cb98e6143af3fcb6a5d4fdfcc
Unable to checkout '76d762c8b086617cb98e6143af3fcb6a5d4fdfcc' in submodule path 'contrib/metaphysicl'
ERROR: exiting with code 1
```
Looks like the issue is a tag without a corresponding branch, and the min-clang recipe does not do `git fetch --tags` yet? Would you mind pushing a branch corresponding to the tag, and then I will invalidate all this?

roystgnr:
Pushed.

jwpeterson:
Looks promising, the build passed on the min-clang target and I see
```
metaphysicl...................... : yes
```
in the configure summary.

jwpeterson:
Also this:
```
checking whether mpicxx supports C++14 features with -std=c++14... no
checking whether mpicxx supports C++14 features with +std=c++14... no
checking whether mpicxx supports C++14 features with -h std=c++14... no
checking whether mpicxx supports C++14 features with -std=c++1y... no
checking whether mpicxx supports C++14 features with +std=c++1y... no
checking whether mpicxx supports C++14 features with -h std=c++1y... no
configure: No compiler with C++14 support was found
checking whether mpicxx supports C++11 features with -std=c++11... yes
```
seems to have worked as desired.

jwpeterson:
I think I'm happy with the current status of this, but before I merge, @roystgnr has to make a solemn promise to stop rewriting metaphysicl history. Once this metaphysicl hash becomes part of libmesh history, it needs to live on in metaphysicl forever.

jwpeterson:
BTW, are you definitely going with the 4-digit "major.minor.subminor.ohshit" version/tag naming scheme going forward?

roystgnr:
No promises on "no rewriting metaphysicl history" in general (although that *is* the dream!), but I'm definitely not rewriting any tag that gets incorporated into libMesh or any branch carrying it.

That's definitely the naming scheme going forward, with the caveat that hopefully most releases won't have a ".ohshit" component.  We haven't had to do that in libMesh since like 2013, but past performance may not predict future results; we now have a lot more testing here than we do there.

roystgnr:
Should we be updating http://libmesh.github.io/installation.html to mention ```git submodule update --init --recursive``` in "External Software", or should we update .gitmodules to set something (```fetch.recurseSubmodules = true```??) that enables that by default?  I'm pretty happy with the default "Roy has a libMesh tree elsewhere and wants to use it instead"-friendly behavior in MOOSE, but here we don't currently even have a mechanism for using an external MetaPhysicL install right now.

jwpeterson:
> Should we be updating http://libmesh.github.io/installation.html to mention git submodule update --init --recursive in "External Software"

Sounds reasonable to me, although I think the page you want to edit is: http://libmesh.github.io/externalsoftware.html, there are blurbs about the various other external software packages we support there.

> should we update .gitmodules to set something (fetch.recurseSubmodules = true??

That option [doesn't seem to be a thing](https://stackoverflow.com/questions/31674177/can-i-use-fetch-recursesubmodules-on-git-clone-set-in-gitconfig)? Maybe it will be in later versions of git. And it's up to you whether you want to start having metaphysicl enabled by default for everyone that clones libmesh... do you have the time to deal with all the weird issues that they will encounter?

Issue #1840, opened 2018-08-28
(evaluable_elements_begin,evaluable_elements_end) does not work as we thought


fdkong:
It looks like` (evaluable_elements_begin,evaluable_elements_end) ` still gives the same results as `(mesh.active_local_elements_begin(),mesh.active_local_elements_end())` even after we attach `algebraic_ghosting_functor`.

Is it supposed to work in this way? `(evaluable_elements_begin,evaluable_elements_end) ` should be always the same as `(mesh.active_local_elements_begin(),mesh.active_local_elements_end())`??? But from my understanding, `(evaluable_elements_begin,evaluable_elements_end) `  somehow should give us ghosted elements as well while `(mesh.active_local_elements_begin(),mesh.active_local_elements_end())` does not give any ghosting information. 

The issue can be reproduced using the following code:

```
diff --git a/tests/base/point_neighbor_coupling_test.C b/tests/base/point_neighbor_coupling_test.C
index 48f7992..8356347 100644
--- a/tests/base/point_neighbor_coupling_test.C
+++ b/tests/base/point_neighbor_coupling_test.C
@@ -148,6 +148,18 @@ public:
                 }
             }
         }
+
+
+        std::cout<<" Number of evaluable elements"<<std::endl;
+        for (const auto & elem: as_range(mesh.evaluable_elements_begin(sys.get_dof_map()),mesh.evaluable_elements_end(sys.get_dof_map())))
+        {
+          std::cout<<elem->id()<<std::endl;
+        }
+        std::cout<<" Number of local elements"<<std::endl;
+        for (const auto & elem: as_range(mesh.active_local_elements_begin(),mesh.active_local_elements_end()))
+        {
+          std::cout<<elem->id()<<std::endl;
+        }
   }
```

roystgnr:
No, this is a known bug that I let get buried.  I think we worked out the right way to do things in some of the #1028 discussion; I'll try to get a fix ready soon.

fdkong:
Thanks, @roystgnr Your patch is beautiful.  We should be fine to close this issue.

PR #1847, opened 2018-08-31
Fix evaluable_* iterators

roystgnr:
Aside from documenting the GhostingFunctor APIs better, we should probably have them working correctly too.

This should fix (and test) the broken iterators @fdkong got hit by in #1840


roystgnr:
Going to merge this straight away since it's obviously an improvement over the prior garbage, but we won't close the issue until @fdkong has had time to try this out next week.

PR #1850, opened 2018-09-05
NemesisIO: subdomain_map should contain vectors of dof_id_type.

jwpeterson:
These are element ids, so using `unsigned int` here is the wrong approach.

I don't have a `--with-64-bit-indices` PETSc build lying around at the moment, so this might not be the only regression, it was the first one found by idaholab/moose#12075, though.


jwpeterson:
This looks like it's going to work so I'm going to merge it now and get the MOOSE PR restarted...

PR #1854, opened 2018-09-06
Don't assert GhostingFunctor presence before removing.

jwpeterson:
Since the DofMap removes GhostingFunctors from both itself and the
underlying mesh, the asserts make it an error if the user subsequently
tries to remove a GhostingFunctor from the Mesh manually. This assert
seems overzealous, though, if there's no GhostingFunctor there we
should just be able to do nothing.

### DIFF
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -1809,7 +1809,6 @@ DofMap::add_coupling_functor(GhostingFunctor & coupling_functor)
 void
 DofMap::remove_coupling_functor(GhostingFunctor & coupling_functor)
 {
-  libmesh_assert(_coupling_functors.count(&coupling_functor));
   _coupling_functors.erase(&coupling_functor);
   _mesh.remove_ghosting_functor(coupling_functor);
 }
--- a/src/mesh/mesh_base.C
+++ b/src/mesh/mesh_base.C
@@ -283,9 +283,6 @@ void MeshBase::clear ()
 
 void MeshBase::remove_ghosting_functor(GhostingFunctor & ghosting_functor)
 {
-  // We should only be trying to remove ghosting functors we actually
-  // have
-  libmesh_assert(_ghosting_functors.count(&ghosting_functor));
   _ghosting_functors.erase(&ghosting_functor);
 }
 


jwpeterson:
@fdkong @permcody

fdkong:
Thanks @jwpeterson.

permcody:
@roystgnr - I'm curious what you were after with this assert. We are tripping it because in MOOSE we almost always have two equation systems tied to a single mesh (one nonlinear system and one explicit system). Right now we are attaching the same ghosting functor to the dof_map in each system, which cascades down to the mesh for the ghosting functor. When the simulation terminates, there's some double removal going on.

Perhaps we shouldn't be using completely identical ghosting functors, but there may be further work needed on the MOOSE side with the way we handle reinits. We treat all variables equally so it's really easy to couple between the systems. Anyway, let us know if you think the removal of this assert is the right course of action.

roystgnr:
I was just trying to catch what I *thought* would imply a user code mistake - e.g. inadvertently making a copy of a GhostingFunctor and trying to remove the copy.  I hadn't thought about the multi-System use case, which is entirely valid, and this is the right fix.

roystgnr:
Although...

roystgnr:
Two concerns:

Why do we need to remove the assert in the DofMap code too?  Are there legitimate reasons to double-remove a GhostingFunctor there too?

Should we be using a reference count here instead?  I'm pretty sure that if I added a coupling (or algebraic ghosting) functor to one DofMap, then added it to another, then removed it from the first and discovered that my geometric ghosting was suddenly insufficient for the second, I would consider that a bug!

permcody:
> Why do we need to remove the assert in the DofMap code too? Are there legitimate reasons to double-remove a GhostingFunctor there too?

For consistency, it should probably be the same, but I don't see us running into this scenario.

Should we be using a reference count here instead? I'm pretty sure that if I added a coupling (or algebraic ghosting) functor to one DofMap, then added it to another, then removed it from the first and discovered that my geometric ghosting was suddenly insufficient for the second, I would consider that a bug!

Maybe, I think I'd honestly like a system that allows us to transfer ownership of these objects. Right now in MOOSE, we build these GhostingFunctors (RelationshipManagers) and attach them to the Application so that libMesh can hold onto the references. It feels a little bit clunky that I just have to have a container of these functors sitting around so that a different object can hold onto them. Maybe as this capability evolves we'd consider passing a smart pointer to these objects? In many simulation cases, it seems like we'd want to build these, attach them and forget about it.

roystgnr:
That's an orthogonal question to the reference count (except with a clever use of multimap...), but I agree, there's not really a performance reason to force apps to handle memory management here.  The current API design was about 50% "this is how Mesh/EquationSystems work" and about 50% "we didn't require C++11 or have shared_ptr until last October but the GhostingFunctor APIs went in in August".

IMHO we should probably take these asserts out now to get you up and running, switch to the reference count shortly afterwards to fix the bug I described above.  We could add a shared_ptr interface but I'm afraid that's going to be an uglier (and thus should be a separate) upgrade - the trouble is that we don't have a good way to just *switch* to shared_ptr, so we'd have to keep containers of shared_ptr *alongside* the current system for a while for backwards compatibility.

permcody:
Perhaps we should hold off on really messing around with this until you travel here in October. We have several topics to talk including this system on our list. I'm getting Fande up to speed with it and Alex is now working with it as well. We will learn more about the system and might have better ideas about how to address these issues moving forward.

jwpeterson:
> IMHO we should probably take these asserts out now to get you up and running

:+1: Will merge after debug testing is done.

Issue #1851, opened 2018-09-05
configure: Disable Trilinos when configuring --with-dof-id-bytes=8


jwpeterson:
I'm not sure if this is our fault or Trilinos' (surely they don't have all indices hard-coded to `int`?) but our `numeric_trilinos_cast` helper function assumes this:
```
int * numeric_trilinos_cast(const numeric_index_type * p)
{
  return reinterpret_cast<int *>(const_cast<numeric_index_type *>(p));
}
```
and we also assert it in a few places, e.g.
```
libmesh_assert_equal_to (sizeof(numeric_index_type), sizeof(int));
```

The right solution is probably to just disable Trilinos at configure time if the user configures with `--with-dof-id-bytes=8`.

roystgnr:
Neither our fault nor theirs, I think; they didn't add a "long long" option until Trilinos 11, so it's reasonable that we hardcoded int too, but they do have it working now and we should eventually upgrade.

But in the meantime, simply disabling Trilinos when our type sizes don't match is the right workaround.

PR #1855, opened 2018-09-06
Disable Trilinos when libmesh is configured with --with-dof-id-bytes=8.

jwpeterson:
I decided to handle this by disabling Trilinos and printing a status message, rather than throwing an error during configure (as we do with PETSc when those index sizes mismatch). If people want a Trilinos index size mismatch to trigger a configure error as well, we could always change it, but I think disabling is more appropriate since it is currently not possible for us to use a Trilinos which is compiled with 64-bit indices.

Refs #1851.

### DIFF
--- a/configure
+++ b/configure
@@ -35718,6 +35718,14 @@ else
 fi
 
 
+        if test "$enabletrilinos" = yes && test "$dof_bytes" != "4"; then :
+
+          enabletrilinos=no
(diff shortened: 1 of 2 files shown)


jwpeterson:
> it is currently not possible for us to use a Trilinos which is compiled with 64-bit indices.

After looking into this in a bit more detail, it seems that the default, at least for the Epetra objects that we use in libmesh, is for Trilinos to be compiled with support for both `int` and `long long int` indices simultaneously. It might be possible for us to template the `numeric_trilinos_cast()` helper function so that it can work with both 32 and 64-bit indices, but that's a project for a different PR.

PR #1856, opened 2018-09-10
Add an index type to vector/tensor structures

lindsayad:
This is useful for generic programming, particularly for upcoming automatic differentiation work

### DIFF
--- a/include/numerics/type_vector.h
+++ b/include/numerics/type_vector.h
@@ -105,6 +105,11 @@ public:
   typedef T value_type;
 
   /**
+   * Helper typedef for generic index programming
+   */
+  typedef unsigned int index_type;
+
+  /**
    * Copy-constructor.
    */
   template <typename T2>
--- a/include/Makefile.in
+++ b/include/Makefile.in
@@ -833,6 +833,7 @@ include_HEADERS = \
         numerics/trilinos_epetra_matrix.h \
         numerics/trilinos_epetra_vector.h \
         numerics/trilinos_preconditioner.h \
+        numerics/tuple_of.h \
         numerics/type_n_tensor.h \
         numerics/type_tensor.h \
         numerics/type_vector.h \
--- a/include/include_HEADERS
+++ b/include/include_HEADERS
@@ -256,6 +256,7 @@ include_HEADERS =  \
         numerics/trilinos_epetra_matrix.h \
         numerics/trilinos_epetra_vector.h \
         numerics/trilinos_preconditioner.h \
+        numerics/tuple_of.h \
         numerics/type_n_tensor.h \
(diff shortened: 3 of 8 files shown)


include/base/libmesh_common.h
```diff
@@ -586,6 +586,17 @@ inline Tnew libmesh_cast_int (Told oldvar)
 #define libmesh_fallthrough() ((void) 0)
 #endif
 
+// Recursive tuple scheme
+template <size_t I, typename T>
+struct libmesh_tuple_n {
+    template< typename...Args> using type = typename libmesh_tuple_n<I-1, T>::template type<T, Args...>;
+};
```
jwpeterson: This doesn't match our usual indentation style for classes/structs (opening curly brace on separate line + indent two spaces).

jwpeterson: I don't think libmesh_common.h is the right place for this... I'd say either make a new header file or maybe put it in `type_vector.h` since `type_tensor.h` and `type_n_tensor.h` both include that header. Also the new structs should be in  the libMesh namespace IMO.

include/numerics/type_vector.h
```diff
@@ -104,6 +104,11 @@ class TypeVector
    */
   typedef T value_type;
 
+  /**
+   * Helper typedef for generic index programming
+   */
+  typedef std::tuple<unsigned int> index_type;
```
jwpeterson: Can `index_type` just be a simple `unsigned int` in this case?

include/base/libmesh_common.h
lindsayad: They are in the libMesh namespace, but yes, moving to another location makes sense to me

include/numerics/type_vector.h
jwpeterson: WTF

include/base/libmesh_common.h
jwpeterson: > They are in the libMesh namespace,

Ah, yeah, I thought we just had a bunch of macros in that file, but the whole thing is wrapped in a namespace. In that case, the name of the struct doesn't really need to have `libmesh_` prepended.

roystgnr:
Sorry, I was going to jump in but by the time I got to it you'd beaten me to my few complaints.  :+1: to the current version (assuming it's C++11-clean; I accidentally use C++14-isms myself sometimes so I'm certain I'd miss catching them in others' code).

jwpeterson:
> assuming it's C++11-clean;

I wondered about that myself, so I attached the min-clang and min-gcc targets to this PR and they seem happy. We also already use the `<typename... Args>` syntax in our `libmesh_make_unique` macro and have a required-to-pass configure test for that so I think we are good.

PR #1862, opened 2018-09-15
Packed range with buffer

friedmud:
This PR is about reducing memory churn (something I'm on a bit of a crusade about lately).  The idea is that you can pass in re-usable buffers for packing.

This works by using shared_ptrs because the request needs to hang onto the buffer until the communication is complete - but the caller also might want to hang onto it so that they can reuse it later.  Also: this works _perfectly_ with an object pool like the one I'm adding to MOOSE for keeping track of the buffers: https://github.com/idaholab/moose/pull/12031

I'm already using these new function in my ray tracing code to good effect:

screen shot 2018-08-21 at 11 33 57 am

The "before" is before re-using these buffers... and the "after" is (obviously) after re-using buffers.  The y axis is performance (lower is better)... the x axis is basically solver iterations.  I re-ran it twice on two different sets of nodes to make sure the results were real... and they are.

That said: you have to be doing a LOT of sends and receives (as I am) before this matters...


roystgnr:
This looks good at first glance, so if you're in a hurry to use it in MOOSE then I'd say go ahead and merge.  If you can wait a week, I'd appreciate a chance to look at it again when I have more time.

friedmud:
No hurry here - I have it on my own branches I’m using right now.  Thanks for taking a look at it!

roystgnr:
I'd love a unit test for this but as far as the design goes it looks solid, thanks.

PR #1866, opened 2018-09-25
Return coupled pairs from variable coupling matrix

fdkong:
It is useful when the number of variables is large. We do not
need to query if (i,j) is coupled in moose any more.

### DIFF
--- a/include/numerics/coupling_matrix.h
+++ b/include/numerics/coupling_matrix.h
@@ -96,6 +96,10 @@ public:
    */
   bool empty() const;
 
+  void print(std::ostream & os) const;
+
+  std::vector<std::pair<std::size_t, std::size_t>> & coupled_pairs();
+
   CouplingMatrix & operator&= (const CouplingMatrix & other);
 
 private:
@@ -119,6 +123,8 @@ private:
   typedef std::pair<std::size_t, std::size_t> range_type;
   typedef std::vector<range_type> rc_type;
   rc_type _ranges;
+
+  rc_type _coupled_pairs;
 
   /**
    * The size of the matrix.
@@ -615,6 +621,7 @@ void CouplingMatrix::resize(const unsigned int n)
   _size = n;
 
   _ranges.clear();
+  _coupled_pairs.clear();
 }
 
 
@@ -633,6 +640,33 @@ bool CouplingMatrix::empty() const
   return (_size == 0);
 }
 
+inline
+void CouplingMatrix::print(std::ostream & os) const
+{
+  for (auto range: _ranges)
+    os<<"("<<range.first<<", "<<range.second<<")"<<std::endl;
+}
+
+inline
+std::vector<std::pair<std::size_t, std::size_t>> & CouplingMatrix::coupled_pairs()
+{
+  if (!empty() && !_coupled_pairs.size())
+  {
+    std::size_t n_nonzeros = 0;
+    for (auto range: _ranges)
+      n_nonzeros += range.second-range.first+1;
+
+    _coupled_pairs.reserve(n_nonzeros);
+    for (auto range: _ranges)
+      for (std::size_t location = range.first; location<=range.second; location++)
+      _coupled_pairs.push_back(std::make_pair(location/_size, location%_size));
+  } else if (empty())
+  {
+    _coupled_pairs.clear();
+  }
+
+  return _coupled_pairs;
+}
 
 } // namespace libMesh
 


fdkong:
A moose change based on this PR has an improvement on matrix assembly. 

Original:

profile002

Improved result:

profile003

roystgnr:
Do you have the MOOSE branch uploaded to look at?

roystgnr:
This looks correctly done and I can't argue with the results, I'm just surprised at their magnitude.

fdkong:
@roystgnr https://github.com/idaholab/moose/pull/12199

roystgnr:
Why do we save _coupled_pairs rather than just leaving it to the user?  Saving it doesn't actually cache anything, since it'll be recomputed on subsequent calls, and modifying that member is forcing you to strip a lot of consts out on the MOOSE side.

fdkong:
> Why do we save _coupled_pairs rather than just leaving it to the user? Saving it doesn't actually cache anything, since it'll be recomputed on subsequent calls, and modifying that member is forcing you to strip a lot of consts out on the MOOSE side.

Just talked to @jwpeterson. I would like to to return `_ranges`, and then we do not need to do any extra stuffs.  That is, something like:

` CouplingMatrix::Ranges() { return _ranges; }`

roystgnr:
Do you have any results using the ConstCouplingRow interface?  I'm no longer surprised at the magnitude of the difference between these two results: it seems like in the MOOSE code you've skipped directly from "Nv^2 nested loop over all variables" to "add new API" without trying "use the API that was specifically designed to avoid Nv^2 nested loops over all variables".  coupled_pairs() would work better in *super* sparse cases, where most variables don't couple to *anything*, but that wouldn't be hard to add another iterator for, and in the common "coupling matrices have at least one entry per row" case I'd expect "iterate over non-zero entries" with the existing API to work at least as well as "copy non-zero entries to the heap, then iterate over the copy" with the new API.

roystgnr:
Every internal data structure we directly expose to the user is a data structure that we can never change for optimization without breaking user code.  I'd hope to avoid that to whatever extent possible.

roystgnr:
We may have a tradeoff here, but we don't know that, do we?  Try the ConstCouplingRow interface and see how it performs, please?

roystgnr:
I know I suck at documentation, but the usage in the library (search for CouplingRow in dof_map.C, sparsity_pattern.C, rb_construction.C) is pretty readable.

fdkong:
> We may have a tradeoff here, but we don't know that, do we? Try the ConstCouplingRow interface and see how it performs, please?

We still have 

```
    _begin_it = std::upper_bound
      (_mat._ranges.begin(), _mat._ranges.end(),
       std::make_pair(_begin_location, max_size));
```

This is what we are trying to avoid.  But it might acceptable because the algorithm become O(N) when using `ConstCouplingRow`. I will try it, and report back.

roystgnr:
Thanks!  Right now the O(Nv^2) cost is horrible yet the O(Nv*Ns) cost is basically invisible, so I'm *hoping* that the O(Nv*(Ns+log(Nv))) option will be tolerable and you can get right underway.  If that cost isn't tolerable, then the fix is the same as for super-sparse cases - an iterator-over-rows so that each row can be initialized based on information from the preceding row rather than based on a binary search from scratch.

fdkong:
I think it is fine to stay with `ConstCouplingRow` because we already cached the data at the MOOSE side. To be clear, the caching data thing was done years ago that is not related my MOOSE PR or libMesh pR. I am just using CoupledPair to cache data instead of Nv^2 thing. Caching data does not take much time. The cost was that we had `adding Jacobian` and `caching Jacobian ` query coupling matrix too often.   

Before, we did not use our cached data to avoid querying.  In MOOSE PR, I try to use reuse the cached data.

fdkong:
I think we can tolerate `ConstCouplingRow`. BTW, I did not know ConstCouplingRow exists.

roystgnr:
> BTW, I did not know ConstCouplingRow exists.

That's probably my fault.  I ought to split those out into separate headers, add some range methods to CouplingMatrix (regardless of whether we need them for this specific optimization) to make their construction more efficient and natural, and get some better documentation in coupling_matrix.h for the whole spiel.

fdkong:
I think a bit more. We can tolerate it right now because we call `Assembly::init `once, but it may be a different story if we call it often. 

Here is a solution: we can cheaply construct a `(pair_begin, pair_end)` on the fly based on `ranges`. Or even cheaper we provide a API to access ranges.

I would leave  all these for the future effort.

PR #1867, opened 2018-09-26
Update metaphysicl

lindsayad:
- Make map container member of NDDualNumbers use shared_ptrs to DualNumberSurrogates
  instead of ordinary instances since the former are copy constructible


roystgnr:
:+1: once CI is happy

jwpeterson:
So we are at least building MOOSE with `--enable-metaphysicl` now, but I don't think anything in MOOSE explicitly uses it yet. This is just a compilation/libmesh unit test check I think.

PR #1868, opened 2018-09-27
Macro tweaks

roystgnr:
In theory this PR is actually only dozens of lines of additions, hundreds of deletions, and thousands of moves (a bunch of code that had with-mpi and without-mpi versions now has a common implementation outside the ifdef LIBMESH_HAVE_MPI block), but my git version doesn't support --color-moved and I can't figure out how to make Github use it either.

In practice I did have a few rebase conflicts; hopefully Civet will verify that I didn't screw up any of the resolutions.

### DIFF
--- a/include/base/libmesh_common.h
+++ b/include/base/libmesh_common.h
@@ -458,7 +458,7 @@ extern bool warned_about_auto_ptr;
 // A function template for ignoring unused variables.  This is a way
 // to shut up unused variable compiler warnings on a case by case
 // basis.
-template<class T> inline void libmesh_ignore( const T & ) { }
+template<class ...Args> inline void libmesh_ignore( const Args&... ) { }
 
 
 // cast_ref and cast_ptr do a dynamic cast and assert
--- a/include/parallel/data_type.h
+++ b/include/parallel/data_type.h
@@ -78,8 +78,7 @@ public:
     // FIXME - if we nest an inner type here will we run into bug
     // https://github.com/libMesh/libmesh/issues/631 again?
     libmesh_call_mpi(MPI_Type_contiguous(count, other._datatype, &_datatype));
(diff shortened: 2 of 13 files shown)


jwpeterson:
There are reams of template error messages in the CIVET output, here is the first one for `Communicator::min()` FWIW.
```
In file included from ./include/libmesh/parallel.h:47:0,
                 from ./include/libmesh/libmesh.h:27,
                 from ../src/base/libmesh_common.C:20:
./include/libmesh/parallel_implementation.h:2103:13: error: prototype for ‘void libMesh::Parallel::Communicator::min(bool&) const’ does not match any in class ‘libMesh::Parallel::Communicator’
 inline void Communicator::min(bool & r) const
             ^~~~~~~~~~~~
In file included from ./include/libmesh/parallel.h:24:0,
                 from ./include/libmesh/libmesh.h:27,
                 from ../src/base/libmesh_common.C:20:
./include/libmesh/communicator.h:237:8: error: candidate is: template<class T> void libMesh::Parallel::Communicator::min(T&) const
   void min(T & r) const;
        ^~~
```

roystgnr:
Thanks; I can reproduce the problems and I've got them mostly fixed.  This PR got interrupted by my trip while I was 80% done with it and when I got back I'd forgotten that the other 20% included --disable-mpi testing...

roystgnr:
Planning to merge this soon; it's only a minor improvement but moving all that code is going to cause no end of conflicts when anyone tries to play with moved sections of parallel_implementation.h while it's outstanding.  Fortunately the new method in #1826 is surrounded by unmoved code so should be safe.

jwpeterson:
The no-MPI test should already be covered by the `--enable-complex` case, but since there were so many changes I thought it would be good to test the no MPI and `!LIBMESH_USE_COMPLEX_NUMBERS` configuration...

PR #1870, opened 2018-10-01
Try an unordered_set to speed up count() calls

roystgnr:
This also avoids the lookup in Mesh::elem_ptr

Hopefully this adds to the CheckpointIO optimization in #1859 

@friedmud, give this a try?  I'd expect it to be a mixed bag: optimization on some problems, pessimization on others.  But the huge problems that are giving you trouble ought to be in the former category, whereas medium problems where tree search is negligible compared to I/O time ought to be in the latter...

### DIFF
--- a/src/mesh/checkpoint_io.C
+++ b/src/mesh/checkpoint_io.C
@@ -17,17 +17,6 @@
 
 #include "libmesh/checkpoint_io.h"
 
-// C++ includes
-#include <iostream>
-#include <iomanip>
-#include <cstdio>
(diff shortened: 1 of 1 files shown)


roystgnr:
If this doesn't work (or if it works but not by as much as I'd hope) then we can try rejiggering the incoming set type instead, but I'm hoping we can speed up write_bcs without any API changes.

roystgnr:
This appears to be a decent improvement, so I'll put incoming-set-rejiggering on the back burner for now.

PR #1871, opened 2018-10-01
Use unordered_set in GhostPointNeighbors functor

roystgnr:
This is another crack at continuing to optimize the case from #1859


roystgnr:
Looks like a win, even if not as big a win as I'd hoped.

Issue #1817, opened 2018-08-07
BoundaryInfo::build_node_list_from_side_list() is really slow when using a large number of processor cores


fdkong:
`BoundaryInfo::build_node_list_from_side_list()` is really slow when using a large number of processor cores since we are using blocking send and receive.  I tested this using 10K, 16K and 32K cores.

roystgnr:
This was resolved in #2607

PR #1872, opened 2018-10-01
parallel_sync in build_node_list_from_side_list

roystgnr:
Hopefully not send/receiving with every single other processor will
help with the scalability issues in #1817, and this will automatically
suck in any subsequent parallel_sync re-optimization too.

### DIFF
--- a/src/mesh/boundary_info.C
+++ b/src/mesh/boundary_info.C
@@ -29,6 +29,7 @@
 #include "libmesh/mesh_communication.h"
 #include "libmesh/mesh_serializer.h"
 #include "libmesh/parallel.h"
+#include "libmesh/parallel_sync.h"
 #include "libmesh/partitioner.h"
 #include "libmesh/remote_elem.h"
 #include "libmesh/unstructured_mesh.h"
@@ -1784,11 +1785,11 @@ BoundaryInfo::build_node_list_from_side_list()
   // might have a boundary side.
   const bool mesh_is_serial = _mesh.is_serial();
 
-  typedef std::set<std::pair<dof_id_type, boundary_id_type>> set_type;
-
-  const processor_id_type n_proc     = this->n_processors();
   const processor_id_type my_proc_id = this->processor_id();
-  std::vector<set_type> nodes_to_push(n_proc);
+
+  typedef std::pair<dof_id_type, boundary_id_type> pair_type;
+  typedef std::set<pair_type> set_type;
+  std::map<processor_id_type, set_type> node_sets;
 
   // Pull objects out of the loop to reduce heap operations
   std::unique_ptr<const Elem> side;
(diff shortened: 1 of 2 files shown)


jwpeterson:
Huh, I'm surprised this doesn't conflict with #1869, but I guess you are good. If you end up doing more development here, you might want to rebase first...

roystgnr:
Ignore that reference; accidentally typed the wrong issue number in the first version of a comment.

roystgnr:
@fdkong, got any time to try this out with whatever you used when noting #1817?  I believe it should show clear improvement at the scales you use, so if you're busy I'll just merge, but I'd feel safer with numbers rather than just a priori beliefs: my last four optimization attempts turned out to be three optimizations and one pessimization...

fdkong:
> @fdkong, got any time to try this out with whatever you used when noting #1817? I believe it should show clear improvement at the scales you use, so if you're busy I'll just merge, but I'd feel safer with numbers rather than just a priori beliefs: my last four optimization attempts turned out to be three optimizations and one pessimization...

Give me one or two days. I will try out.

fdkong:
Thanks so much for working this thing.

roystgnr:
Ping?

fdkong:
Sorry..., I am aggressively working on upgrading PETSc. I haven't had time back here.  

I am trying my best to get it back to you.

friedmud:
Oh damn.  I've been getting bit by this - and I didn't know it!  Going to pull  this and try the fix now.

Damn. Damn. Damn.

friedmud:
Well - this fixes the "hang" - but causes an enormous memory spike that causes me to run out of memory.

Still investigating...

friedmud:
@roystgnr am I insane - or does this push all boundary nodes to _all_processors_!?  That doesn't work with 200M elements on 5k procs...

Unfortunately - we are severely handicapped here by not having node->elem pointers and a guaranteed 1 level of geometric ghosting.  Then we would know exactly who needs to be sent what.

I think that I'm going to implement a slimmer version of this function that uses a node_to_elem_map and depends on one level of nodal neighbors ghosting.

friedmud:
Ok - I see the algorithm now - it doesn't push to all processors - it pushes to owners and then pulls ghosts.  Seems reasonable... I'm going to try to hunt down which part is blowing up in memory...

roystgnr:
*Every* processor stuck on the same allgather?  That shouldn't be possible, right?  Are one or more procs out of step with the rest?

friedmud:
It eventually made it past that... but that AllGather is seriously expensive at 4k procs...

BTW: Reverting from the NBX algorithm seemed to help - but the code is now way slower - the AllToAll in all of the parallelsync stuff really shows up at 4k procs and 200M elements...

I'm still debugging to figure out what's up with the NBX algorithm.  I would like to fix it and put it in more of the parallel sync places.

jwpeterson:

PR #1874, opened 2018-10-01
Move UnstructuredMesh::all_{first,second}_order() functions.

jwpeterson:
These were originally MeshTools::Modification functions IIRC, but they
are UnstructuredMesh member functions now so they should be in
unstructured_mesh.C.


roystgnr:
:+1: assuming you didn't slip in any changes.  I (and github) really need to get into the modern ```git diff --color-moved``` era one of these days...

Issue #1821, opened 2018-08-08
Optimization of Mesh Reading / Initialization / Splitting


friedmud:
I'm currently splitting a 2.2GB Exodus mesh with about 55 Million elements using 48 MPI procs.  I'm splitting it a few ways: ~9000 procs, ~500 procs, ~100 procs, etc.

Doing so has uncovered some pretty large inefficiencies that I'd appreciate some discussion on.

Here is a profile run (using Google Perf Tools - showing just processor 0) for splitting it for 144 procs:

![profile001](https://user-images.githubusercontent.com/870705/43816106-5dac33b2-9a90-11e8-81f5-386fb01bfbc2.png)

And here are the "top" functions from that profiling

```
(pprof) top
Showing nodes accounting for 462.35s, 53.41% of 865.65s total
Dropped 282 nodes (cum <= 4.33s)
Showing top 10 nodes out of 118
      flat  flat%   sum%        cum   cum%
   122.81s 14.19% 14.19%    122.81s 14.19%  std::_Rb_tree::equal_range
    98.43s 11.37% 25.56%    122.04s 14.10%  libMesh::GhostPointNeighbors::operator()
    52.40s  6.05% 31.61%    113.26s 13.08%  MooseMesh::cacheInfo
    47.91s  5.53% 37.15%     47.91s  5.53%  __memset_sse2
    28.58s  3.30% 40.45%     28.80s  3.33%  _int_malloc
    27.16s  3.14% 43.58%     27.16s  3.14%  std::_Rb_tree::find
    23.06s  2.66% 46.25%     23.06s  2.66%  libMesh::Predicates::pid::operator()
    22.24s  2.57% 48.82%     22.41s  2.59%  std::_Rb_tree::_M_get_insert_unique_pos
    20.29s  2.34% 51.16%     58.02s  6.70%  variant_filter_iterator::Pred::operator()
    19.47s  2.25% 53.41%     19.87s  2.30%  MPIDI_CH3I_MRAILI_Get_next_vbuf
```

## GhostingFunctor

The first major one is that `libMesh::GhostPointNeighbors::operator()` is a full loop over the full mesh on every processor for every split.  So it's `n_procs * n_elements` kind of operation.  So when you're talking about ~20k procs and 1 billion elements... just this piece gets prohibitive.  In the above timing you can see that this ends up taking up 14% of the runtime - but if you are splitting for many more procs this gets prohibitve fast.

[...]

Issue #1787, opened 2018-07-09
Use consistent ordering for BoundaryInfo::_boundary_side_id multimap


jwpeterson:
This container is currently sorted by `operator<` for pointers, resulting in different orderings on different processors when the code is run in parallel. How this issue has never come up before, I'm not so sure, but we first noticed it over in idaholab/moose#11800 when trying to add lower-dimensional elements from sidesets based on the output of `BoundaryInfo::build_side_list()` which simply extracts information from the multimap.

A similar fix should be done for `_boundary_shellface_id`, `_boundary_edge_id`, and `_boundary_node_id` multimaps, which are also all sorted by pointer address currently.

jwpeterson:
This issue was partially addressed in 438f9bcbe without changing the types of the underlying containers (i.e. they are still multimaps sorted on memory address) by sorting the vectors before they are returned to the user. We might be able to do even better by sorting on some combination of the elem/side/id information, but we still need to be able to look up elements quickly by pointer, so this approach will take a little bit more design/thought.

roystgnr:
It looks as if, even if we use a custom comparator, we'll *still* need to do sorting of the result.  The Compare argument in a multimap only acts on keys, so ```multimap<Key, T>``` is basically unordered with respect to T.

I'll try a custom comparator anyways.  If the results @friedmud got from the unordered_map attempt are to be believed then we *might* get a tiny performance boost from better sorting, and in any case if we *know* we have a perfect sort-by-Key then we could next try replacing "sort the whole list" with "loop over the whole list and just sort the sections with matching keys".

roystgnr:
Hmmm... C++11 *does* mandate that equivalent elements retain their insertion order, so "unordered" is no longer true.  But I don't think it's safe to bet that we're always inserting boundary ids on the same order on different ranks, in which case for our purposes we have an unordered range associated with each key.

Except... if 99% of codes will actually end up with our data already perfectly sorted, and the worst case will still end up with our data nearly sorted, it's kind of tempting to go past "just sort the sections with matching keys" to "just bubble sort the sections with matching keys"!  Best case O(n), and the godawful "computing for MechEs" class that unironically taught me bubble sort would be vindicated!

roystgnr:
[...]

PR #1875, opened 2018-10-01
Use unordered_map instead of map in BoundaryInfo

roystgnr:
We don't use the useless order-by-pointer-address behavior of maps of
pointers, and O(1) average lookup time may be more efficient than
just using a saner ordering.

Hopefully this will eat into the 14% of #1821 runtime currently spent
in equal_range calls from BoundaryInfo

This would make #1787 moot

### DIFF
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -29,8 +29,9 @@
 #include <cstddef>
 #include <map>
 #include <set>
-#include <vector>
 #include <tuple>
+#include <unordered_map>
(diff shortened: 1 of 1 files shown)


roystgnr:
Like most of my PRs today, this ought to wait on performance results from @friedmud even if CI is happy with it.

roystgnr:
So this actually replaces that equal_range runtime with much-slower introsort runtime, plus extra slowness to spare.  John probably had the right idea in #1787 after all.

Issue #1841, opened 2018-08-28
DofMap::local_variable_indices appears incorrect


YaqiWang:
I am running a case with multiple processors and with *block restricted* variables in LAGRANGE FIRST. Unfortunately the case is not in just libMesh so I cannot post it here. The size of indices for one variable obtained with this function is drastically smaller than the size with `System::local_dof_indices` (which returns a set) (5 vs 281 on master processor). I am guessing the logic at https://github.com/libMesh/libmesh/blob/master/src/base/dof_map.C#L1109 is not sufficient to catch my case.

I attempted a fix like this:
```
diff --git a/src/base/dof_map.C b/src/base/dof_map.C
index 5e259cc19..eac3d4551 100644
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -1076,6 +1076,8 @@ void DofMap::local_variable_indices(std::vector<dof_id_type> & idx,
 {
   const unsigned int sys_num       = this->sys_number();
 
+  std::set<dof_id_type> temp;
+  unsigned int counter = 0;
   // If this isn't a SCALAR variable, we need to find all its field
   // dofs on the mesh
   if (this->variable_type(var_num).family != SCALAR)
@@ -1100,9 +1102,15 @@ void DofMap::local_variable_indices(std::vector<dof_id_type> & idx,
                   const dof_id_type index = node.dof_number(sys_num,var_num,i);
                   libmesh_assert_greater_equal (index, this->first_dof());
                   libmesh_assert_less (index, this->end_dof());
-
-                  if (idx.empty() || index > idx.back())
+                  temp.insert(index);
+                  if (temp.size() > counter)
+                  {
                     idx.push_back(index);
+                    ++counter;
+                  }
+
+//                  if (idx.empty() || index > idx.back())
+//                    idx.push_back(index);
                 }
             }
 
@@ -1111,8 +1119,14 @@ void DofMap::local_variable_indices(std::vector<dof_id_type> & idx,
           for (unsigned int i=0; i<n_comp; i++)
             {
               const dof_id_type index = elem->dof_number(sys_num,var_num,i);
-              if (idx.empty() || index > idx.back())
+              temp.insert(index);
+              if (temp.size() > counter)
+              {
                 idx.push_back(index);
+                ++counter;
+              }
+//              if (idx.empty() || index > idx.back())
+//                idx.push_back(index);
             }
         } // done looping over elements
 
[...]

PR #1877, opened 2018-10-02
Fixes for local_*_indices methods

roystgnr:
And a new unit test from @YaqiWang which (prior to the fixes) triggered the #1841 problems that he caught.

If this works for his application code too then we declare victory; if not then IMHO even a partial fix would be a big improvement over the previous behavior.

### DIFF
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -1092,14 +1092,23 @@ void DofMap::local_variable_indices(std::vector<dof_id_type> & idx,
                                     const MeshBase & mesh,
                                     unsigned int var_num) const
 {
+  // Count dofs in the *exact* order that distribute_dofs numbered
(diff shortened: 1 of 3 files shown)


roystgnr:
FYI, although that unit test is showing up first in the list, git log tells me it is the last commit in the PR, as I intended to keep "make check" working on intermediate commits.  Not sure why github is showing me those commits in chronological order rather than post-interactive-rebase order.

YaqiWang:
I will test this out.

YaqiWang:
This appears fixing my issue.

roystgnr:
I'll merge tonight, then.  Thanks for the help digging into this, and sorry for the delay fixing it!

YaqiWang:
Thanks to you @roystgnr .

PR #1878, opened 2018-10-02
Heap vectors of local node data can instead be stack arrays

roystgnr:
This ought to shave 4 or 5 seconds off of the results @friedmud shared
in #1821 

I was kind of hoping to find more than a couple spots where this trick would be useful.

### DIFF
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -87,6 +87,8 @@ Threads::spin_mutex parent_bracketing_nodes_mutex;
 const subdomain_id_type Elem::invalid_subdomain_id = std::numeric_limits<subdomain_id_type>::max();
 
 // Initialize static member variables
+const unsigned int Elem::max_n_nodes;
+
 const unsigned int Elem::type_to_n_nodes_map [] =
   {
     2,  // EDGE2
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -19,6 +19,7 @@
 
 // C++ includes
 #include <algorithm> // for std::sort
+#include <array>
 #include <iterator>  // for std::ostream_iterator
 #include <sstream>
 #include <limits>    // for std::numeric_limits<>
@@ -401,16 +402,16 @@ dof_id_type Elem::key () const
 {
   const unsigned short n_n = this->n_nodes();
 
-  std::vector<dof_id_type> node_ids(n_n);
+  std::array<dof_id_type, Elem::max_n_nodes> node_ids;
 
   for (unsigned short n=0; n != n_n; ++n)
     node_ids[n] = this->node_id(n);
 
   // Always sort, so that different local node numberings hash to the
   // same value.
-  std::sort (node_ids.begin(), node_ids.end());
+  std::sort (node_ids.begin(), node_ids.begin()+n_n);
 
-  return Utility::hashword(node_ids);
+  return Utility::hashword(node_ids.data(), n_n);
 }
 
 
@@ -424,11 +425,9 @@ bool Elem::operator == (const Elem & rhs) const
   const unsigned short n_n = this->n_nodes();
   libmesh_assert_equal_to(n_n, rhs.n_nodes());
 
-  // Make two sorted vectors of global node ids and compare them for
+  // Make two sorted arrays of global node ids and compare them for
   // equality.
-  std::vector<dof_id_type>
-    this_ids(n_n),
-    rhs_ids(n_n);
+  std::array<dof_id_type, Elem::max_n_nodes> this_ids, rhs_ids;
 
   for (unsigned short n = 0; n != n_n; n++)
     {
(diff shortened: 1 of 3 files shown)


src/geom/elem.C
```diff
@@ -399,7 +400,7 @@ dof_id_type Elem::key () const
 {
   const unsigned short n_n = this->n_nodes();
 
-  std::vector<dof_id_type> node_ids(n_n);
+  std::array<dof_id_type, Elem::max_n_nodes> node_ids;
```
jwpeterson: Wait, we are now sorting (and hashing) an array with 27 entries every time regardless of the number of nodes the Elem has, and this is faster?

jwpeterson:
Is this our compilers, or did you not actually try to compile before pushing?
```
../src/geom/elem.C:412:36: note:   ‘std::array<unsigned int, 27>’ is not derived from ‘const Container<T, Args ...>’
Makefile:23623: recipe for target 'src/geom/libmesh_opt_la-elem.lo' failed
make[1]: *** [src/geom/libmesh_opt_la-elem.lo] Error 1
make[1]: *** Waiting for unfinished jobs....
```

src/geom/elem.C
roystgnr: No, this is me forgetting to rejigger the end iterator of the sort call.

roystgnr:
Gah, no, I compiled but had to head home and didn't get to the end of the compile.

src/geom/elem.C
```diff
@@ -399,16 +402,16 @@ dof_id_type Elem::key () const
...
   for (unsigned short n=0; n != n_n; ++n)
     node_ids[n] = this->node_id(n);
 
   // Always sort, so that different local node numberings hash to the
   // same value.
-  std::sort (node_ids.begin(), node_ids.end());
+  std::sort (node_ids.begin(), node_ids.begin()+n_n);
 
-  return Utility::hashword(node_ids);
+  return Utility::hashword(node_ids.begin(), n_n);
```
jwpeterson: So this is calling one of the overloads
```
uint32_t hashword(const uint32_t * k, size_t length, uint32_t initval=0)
uint64_t hashword(const uint64_t * k, size_t length);
```
depending on `dof_id_type`. Is it guaranteed that `std::array<T>::begin()` returns `T*`?

The docs [here](https://en.cppreference.com/w/cpp/container/array/begin) and [here](https://en.cppreference.com/w/cpp/container/array) refer to an `iterator` type but I don't see any guarantees of this...

jwpeterson: Seems to me that `node_ids.data()` is the more correct thing to use, though I expect in almost all cases they are going to be equivalent.

roystgnr: Oh, hell, I am passing begin() to the hashword method.  Thanks; I'll fix that.  Bugs that work fine on most compilers until things scream and die on newer compilers are not friendly sorts of bugs.  At least this one would have been a compile-time error.  Thanks for catching it!

jwpeterson: The line I commented on is passing begin() to hashword, unless I need to hit refresh...

roystgnr: I did remember to use data() in hashword.h so I haven't *completely* lost my mind...

jwpeterson:
Do you have any performance numbers for this change? I'm planning to pull it myself and see, but I figured I'd run the same test as you if you had already developed one.

roystgnr:
Nah; I was hoping that @friedmud would rerun his case from the top of #1821.

In theory the 5.57s and 4.38s from Elem::operator== calling new and free should be gone now, and I can't think of anything that might be countervailing.  (a couple hundred bytes of bloat elsewhere might lead to L1 cache misses, but not if it's at the very top of the stack, right?)

roystgnr:
We really need a nice parameterized benchmarking suite at some point, though, and a way to archive results from it.

jwpeterson:
This optimization looks like a win to me. I wrote a simple [test code](https://gist.github.com/jwpeterson/5b4f0fd0d92939c618ea3a976b3808cc) that just calls `Elem::key()` a bunch of times, and tested it on different numbers of threads. I focused mainly on `HEX27` elements as they are the most expensive to hash, and seemed to show the most improvement from the `std::array` optimization. In single-threaded mode, the `std::array` version of the code is about 1.4X faster than the `std::vector` version. As the number of threads is increased, the two methods seem to scale similarly. Although the gap between them has narrowed by the time we reach `n_threads=4`, the `std::array` version is always the faster of the two.

key_vector_vs_array

jwpeterson:
Continuing the previous results out to 8 threads. Besides being faster, the `std::array` results also look a bit smoother/more predictable than the `std::vector` ones. I can't really explain the kink in the results from 3-5 procs; all these times are averaged over 5 runs.

key_vector_vs_array

roystgnr:
Thanks!  I guess at 8 threads we're starting to hit some kind of Amdahl's law w.r.t. cache contention?  Except we're only reading from shared memory, never writing, so I'm not sure how that would be a problem.

Regardless, this seems like a uniform improvement.  Thanks for the tests!

jwpeterson:
> so I'm not sure how that would be a problem.

I'm a bit troubled by the poor scaling beyond 2 threads in the `std::array` case, as I believe they should be able to scale almost perfectly. This is using the openmp thread model, so I'd like to try an even simpler test (like summing the entries of a vector using multiple threads) and compare the openmp, pthread, and tbb implementations of `parallel_reduce()`.

jwpeterson:
I think most of the poor scalability I was observing before can be explained by the combination of a relatively fast `operator()` function and the fact that I was running multiple loops to get meaningful timing data. This in turn meant that some of my reported time was being spent in thread startup and teardown. If I make the `operator()` calls artificially more expensive (but still scalable) by calling `usleep(range.size());` and then run only 1 loop, the scalability results are much closer to optimal:

usleep_range_size

PR #1881, opened 2018-10-04
Added unit test for DGFEMContext.

dknez:
The test doesn't make any assertions, but it does check that we can successfully assemble and solve a system with discontinuous basis functions using DGFEMContext.

### DIFF
--- a/tests/systems/systems_test.C
+++ b/tests/systems/systems_test.C
@@ -21,6 +21,7 @@
 #include <libmesh/quadrature_gauss.h>
 #include <libmesh/node_elem.h>
 #include "libmesh/edge_edge2.h"
+#include "libmesh/dg_fem_context.h"
 
 #include "test_comm.h"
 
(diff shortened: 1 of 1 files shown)


dknez:
This is in reference to requests for tests and/or examples of DGFEMContext.

In the new test we assemble a DG system using DGFEMContext and then solve. We do not make any CPPUNIT assertions (since I didn't know what I could assert) but I think it's already helpful to check that we can assemble and solve without error.

tests/systems/systems_test.C
```diff
@@ -583,6 +728,30 @@ public:
...
+    sys.add_variable("u", FIRST, L2_LAGRANGE);
+
+    MeshTools::Generation::build_cube (mesh,
+                                       5, 5, 5,
+                                       0., 1., 0., 1., 0., 1.,
+                                       HEX8);
+
+    es.init();
+    sys.attach_assemble_function (assembly_with_dg_fem_context);
+    sys.solve();
```
jwpeterson: I'd like to pull this PR and see how long it takes to run the unit tests in dbg mode. 

Do you think the test would still be sufficient if we just called `sys.assemble()`? Would it be OK if I changed it to 2D?

dknez: Sure, please proceed as you like. It's fine with me if you change this to 2D.

jwpeterson: Actually I think I'll just merge as is, the test only increased the dbg mode unit test runtime by about 1% (47.7s -> 48.2s).

PR #1884, opened 2018-10-05
return type workaround for possible gcc 7.1.0 bug

roystgnr:
With the previous code, at least gcc 6.3, 7.2, and 7.3 were happy with
us, but gcc 7.1.0 didn't think the declaration and definition matched.


lindsayad:
Interesting failure. I think this code looks cleaner anyways!

jwpeterson:
~~cleaner~~ more beautiful

lindsayad:
Truly beautiful code

PR #1885, opened 2018-10-08
Remove more iterator boilerplate

jwpeterson:
Another round of maintenance/refactoring to convert more traditional for-loops into range-based for-loops.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -4324,19 +4324,13 @@ DofMap::get_adjoint_dirichlet_boundaries(unsigned int q)
 void DofMap::remove_dirichlet_boundary (const DirichletBoundary & boundary_to_remove)
 {
   // Find a boundary condition matching the one to be removed
-  std::vector<DirichletBoundary *>::iterator it = _dirichlet_boundaries->begin();
-  std::vector<DirichletBoundary *>::iterator end = _dirichlet_boundaries->end();
-  for (; it != end; ++it)
-    {
-      DirichletBoundary * bdy = *it;
+  auto lam = [&boundary_to_remove](const DirichletBoundary * bdy)
+    {return bdy->b == boundary_to_remove.b && bdy->variables == boundary_to_remove.variables;};
 
-      if ((bdy->b == boundary_to_remove.b) &&
-          bdy->variables == boundary_to_remove.variables)
-        break;
-    }
+  auto it = std::find_if(_dirichlet_boundaries->begin(), _dirichlet_boundaries->end(), lam);
 
   // Delete it and remove it
-  libmesh_assert (it != end);
+  libmesh_assert (it != _dirichlet_boundaries->end());
   delete *it;
   _dirichlet_boundaries->erase(it);
 }
@@ -4348,22 +4342,15 @@ void DofMap::remove_adjoint_dirichlet_boundary (const DirichletBoundary & bounda
   libmesh_assert_greater(_adjoint_dirichlet_boundaries.size(),
                          qoi_index);
 
-  // Find a boundary condition matching the one to be removed
-  std::vector<DirichletBoundary *>::iterator it =
-    _adjoint_dirichlet_boundaries[qoi_index]->begin();
-  std::vector<DirichletBoundary *>::iterator end =
-    _adjoint_dirichlet_boundaries[qoi_index]->end();
-  for (; it != end; ++it)
-    {
-      DirichletBoundary * bdy = *it;
+  auto lam = [&boundary_to_remove](const DirichletBoundary * bdy)
+    {return bdy->b == boundary_to_remove.b && bdy->variables == boundary_to_remove.variables;};
 
-      if ((bdy->b == boundary_to_remove.b) &&
-          bdy->variables == boundary_to_remove.variables)
-        break;
-    }
+  auto it = std::find_if(_adjoint_dirichlet_boundaries[qoi_index]->begin(),
+                         _adjoint_dirichlet_boundaries[qoi_index]->end(),
+                         lam);
 
   // Delete it and remove it
-  libmesh_assert (it != end);
+  libmesh_assert (it != _adjoint_dirichlet_boundaries[qoi_index]->end());
   delete *it;
   _adjoint_dirichlet_boundaries[qoi_index]->erase(it);
 }
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -3855,18 +3855,16 @@ void DofMap::scatter_constraints(MeshBase & mesh)
   // Each ghost-coupled element's owner should get a request for its dofs
   std::set<dof_id_type> requested_dofs;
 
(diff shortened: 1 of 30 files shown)


jwpeterson:
@pbauman I was thinking this would be an uncontroversial change but apparently GRINS doesn't think so... Would you mind letting me know if these test failures are real?

pbauman:
<pre>
/femputer/pbauman/civet_build_testing_root/libMesh/src/mesh/fro_io.C: In member function ‘virtual void libMesh::FroIO::write(const string&)’:
/femputer/pbauman/civet_build_testing_root/libMesh/src/mesh/fro_io.C:195:31: error: declaration of ‘id’ shadows a previous local [-Werror=shadow]
             for (const auto & id : node_list)
                               ^~
/femputer/pbauman/civet_build_testing_root/libMesh/src/mesh/fro_io.C:108:27: note: shadowed declaration is here
         for (const auto & id : bc_ids)
</pre>

pbauman:
Score one for -Werror

src/base/dof_map_constraints.C
```diff
@@ -4350,22 +4342,15 @@ void DofMap::remove_adjoint_dirichlet_boundary (const DirichletBoundary & bounda
...
+  auto lam = [&boundary_to_remove](const DirichletBoundary * bdy)
+    {return bdy->b == boundary_to_remove.b && bdy->variables == boundary_to_remove.variables;};
 
-      if ((bdy->b == boundary_to_remove.b) &&
-          bdy->variables == boundary_to_remove.variables)
-        break;
-    }
+  auto it = std::find_if(_adjoint_dirichlet_boundaries[qoi_index]->begin(),
+                         _adjoint_dirichlet_boundaries[qoi_index]->end(),
+                         lam);
```
roystgnr: Now you're just showing off.

jwpeterson: The `get_` and `remove_` versions of this function could also theoretically share the same lambda, avoiding code duplication, but I was not sure how to do that...

src/mesh/gmv_io.C
```diff
@@ -1046,7 +1042,11 @@ void GMVIO::write_ascii_old_impl (const std::string & fname,
           for (const auto & elem : mesh.active_element_ptr_range())
             {
               // Find the unique index for elem->subdomain_id(), print that to file
-              sbdid_map_iter map_iter = sbdid_map.find( elem->subdomain_id() );
+              auto map_iter = sbdid_map.find(elem->subdomain_id());
+
+              if (map_iter == sbdid_map.end())
+                libmesh_error_msg("Entry for subdomain " << elem->subdomain_id() << " not found.");
```
roystgnr: Shouldn't this be an assert?  It should be currently guaranteed never to fail and if it ever does that'll be to a change causing an internal error.

src/mesh/gmv_io.C
```diff
@@ -2208,8 +2155,9 @@ ElemType GMVIO::gmv_elem_to_libmesh_elem(std::string elemname)
   elemname.erase(std::remove_if(elemname.begin(), elemname.end(), isspace), elemname.end());
 
   // Look up the string in our string->ElemType name.
-  std::map<std::string, ElemType>::iterator it = _reading_element_map.find(elemname);
+  auto it = _reading_element_map.find(elemname);
 
+  // Look up the string in our string->ElemType name.
```
roystgnr: Accidental copy and paste here?

src/mesh/gmv_io.C
```diff
@@ -2247,32 +2195,20 @@ void GMVIO::copy_nodal_solution(EquationSystems & es)
...
       // For each var entry in the _nodal_data map, try to find
       // that var in the system
-      std::map<std::string, std::vector<Number>>::iterator it = _nodal_data.begin();
-      const std::map<std::string, std::vector<Number>>::iterator end = _nodal_data.end();
-      for (; it != end; ++it)
+      for (const auto & pr : _nodal_data)
         {
-          std::string var_name = it->first;
-          // libMesh::out << "Searching for var " << var_name << " in system " << sys << std::endl;
+          std::string var_name = pr.first;
```
roystgnr: While we're playing with the code here anyway, should this be a reference-to-const?

src/mesh/gmv_io.C
jwpeterson: An assert would be fine with me... Before we were using the iterator without any checks so I figured something would be better than nothing.

src/reduced_basis/rb_construction_base.C
```diff
@@ -230,17 +226,15 @@ void RBConstructionBase<Base>::initialize_training_parameters(const RBParameters
   // allowable discrete value
   if (get_n_discrete_params() > 0)
     {
-      std::map<std::string, std::unique_ptr<NumericVector<Number>>>::iterator it           = training_parameters.begin();
-      std::map<std::string, std::unique_ptr<NumericVector<Number>>>::const_iterator it_end = training_parameters.end();
-      for ( ; it != it_end; ++it)
+      for (const auto & pr : training_parameters)
         {
-          std::string param_name = it->first;
+          std::string param_name = pr.first;
```
roystgnr: This could be reference-to-const too, right?

src/reduced_basis/rb_construction_base.C
```diff
@@ -270,29 +264,25 @@ void RBConstructionBase<Base>::load_training_set(std::map<std::string, std::vect
...
+      pr.second->init(n_global_training_samples, n_local_training_samples, false, PARALLEL);
     }
 
-  it = training_parameters.begin();
-  for ( ; it != it_end; ++it)
+  for (auto & pr : training_parameters)
     {
-      std::string param_name = it->first;
-      NumericVector<Number> * training_vector = it->second.get();
+      std::string param_name = pr.first;
```
roystgnr: And here?

src/reduced_basis/rb_construction_base.C
```diff
@@ -391,38 +381,33 @@ void RBConstructionBase<Base>::generate_training_parameters_random(const Paralle
   }
 
   // finally, set the values
-  {
-    std::map<std::string, std::unique_ptr<NumericVector<Number>>>::iterator it           = training_parameters_in.begin();
-    std::map<std::string, std::unique_ptr<NumericVector<Number>>>::const_iterator it_end = training_parameters_in.end();
+  for (auto & pr : training_parameters_in)
+    {
+      std::string param_name = pr.first;
```
roystgnr: Okay, maybe these all belong in a separate PR...

src/mesh/gmv_io.C
jwpeterson: Definitely, no reason to make a copy.

roystgnr:
I'd say change that error to an assert before merging, but I'm not wedded to that.  Otherwise :+1: - The string copies turn out to be way too plentiful to bother with "while we're at it" in this PR.

src/reduced_basis/rb_construction_base.C
jwpeterson: I'll just fix them now, I found only like one other one so far...

PR #1886, opened 2018-10-09
Acquire BC lists outside the CheckPointIO id loop

roystgnr:
This should be just as fast for restarts and O(Nsplits/Nprocs) times
faster for mesh splitting.  Thanks to @friedmud for the idea.

I'm about to head home so I haven't yet made 100% sure this optimizes things.  Or works correctly.  Or compiles.  Hopefully @friedmud can check up on the first question and Civet will scream at me if I've flubbed the latter two.

### DIFF
--- a/include/mesh/checkpoint_io.h
+++ b/include/mesh/checkpoint_io.h
@@ -214,13 +214,15 @@ private:
    * Write the side boundary conditions for part of a mesh
    */
   void write_bcs (Xdr & io,
-                  const std::set<const Elem *, CompareElemIdsByLevel> & elements) const;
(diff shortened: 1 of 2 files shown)


roystgnr:
Ping #1859 where the idea was floated.

friedmud:
Testing this momentarily

friedmud:
Also: put in a PR for the new node_to_elem map based ghosting functor: #1887

roystgnr:
https://github.com/libMesh/libmesh/pull/1859#issuecomment-428388588 is awesome - merging.

PR #1887, opened 2018-10-09
Add map based point neighbor ghosting functor

friedmud:
Adds a superfast point neighbor ghosting functor that uses the nodes_to_elem_map.  Apparently it does have some limitations with respect to some element types and refinement (see: https://github.com/libMesh/libmesh/pull/1859#issuecomment-427091562 ) - but I think it should still be useful in 99% of cases.


friedmud:
Damn - I messed up something with metaphysicl.  Any ideas @roystgnr and @lindsayad ?

contrib/netcdf/4.4.1.1/include/netcdf_meta.h
```diff
@@ -38,10 +38,10 @@
 #define NC_VERSION       "4.4.1.1"
 
 #define NC_HAS_NC2       1 /*!< API version 2 support. */
-#define NC_HAS_NC4       1 /*!< API version 4 support. */
```
roystgnr: This file's changes look like a different-autotools-version false positive?

include/libmesh/Makefile.am
```diff
@@ -204,6 +205,54 @@ BUILT_SOURCES = \
         unv_io.h \
         vtk_io.h \
         xdr_io.h \
+        compare_types.h \
```
roystgnr: And how did this happen?  You don't actually have these files in include/ anywhere, right?  They're all in contrib?

roystgnr:
It looks as if somehow you had metaphysicl headers in include/*/ somewhere before running include/libmesh/rebuild_makefile.sh - could that possibly be the case?  I just tried running that myself but it was a no-op (as it should be when I haven't added any headers).  The script is hand-crafted but ```find .. -name '*.h'``` etc. doesn't leave much ambiguity.

permcody:
Custom in-tree build script is getting in the way!

friedmud:
Yep - I had to basically copy all of the metaphysicl headers into `include/metaphysicl` to make my in-tree build work.  I guess I then ran the scripts and accidentally added junk I didn't mean to add.  Let me strip out all of that crap and re-push.

friedmud:
Don't merge this yet.  I want to do some more testing of it first....

jwpeterson:

PR #1888, opened 2018-10-10
reconnect_nodes with an unordered_set

roystgnr:
We don't actually need an ordering here, and count/insert/etc
operations in O(1) instead of O(log(N)) might be nice.

@friedmud, try this on the case from #1859 and see if it chops anything off the write_nodesets time?

This might be another pessimization instead but it's an easy enough change to be worth trying.

### DIFF
--- a/include/mesh/checkpoint_io.h
+++ b/include/mesh/checkpoint_io.h
@@ -28,7 +28,9 @@
 #include "libmesh/parallel_object.h"
 
 // C++ includes
+#include <set>
 #include <string>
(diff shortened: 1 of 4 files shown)


roystgnr:
Pinging @friedmud - got any time to check this out?

jwpeterson:

roystgnr:
We might still do this but if so it'll need to be rewritten on top of https://github.com/libMesh/libmesh/pull/3364

Issue #1889, opened 2018-10-10
Potential issue with PETSc 3.9 debug builds


jwpeterson:
I have a preliminary report from @andfranklin (no stack trace yet, unfortunately) that we are once again incorrectly accessing a locked PETSc vector in PETSc 3.9 from one of the libmesh examples, basically the same thing that was reported in #1294. The problem does not seem to be present in PETSc 3.8. I'll try to add more information to this ticket as it becomes available.

cc: @fdkong

jwpeterson:
To reproduce the problem, you must configure PETSc with `--with-debugging=1`.

fdkong:
I will try to smooth MOOSE and libMesh with PETSc-3.9.4

fdkong:
I saw some similar issues before in MOOSE, and will try again

pbauman:
Dammit. All my CIVET tests are using PETSc opt. It's a simple change to the module statement on my CIVET server to use PETSc dbg. I can turn that on when folks are ready.

andfranklin:
I was able to reproduce the failure using PETSc 3.9.4 configured with:

```
./configure CC="$CC" CXX="$CXX" --download-mpich --download-fftw --download-hdf5 --download-fblaslapack --download-chaco --download-exodusii --download-hypre --download-metis --download-ml --download-mumps --download-netcdf --download-parmetis --download-scalapack --download-suitesparse --download-superlu --download-superlu_dist --download-zlib --download-pnetcdf --with-c2html=0
```
 and SLEPc 3.9.2. The following was used to configure libmesh:

```
../configure --prefix=$PWD/install --enable-cxx11-required --enable-nodeconstraint --enable-petsc-required --with-cxx="$CXX" --with-cc="$CC" --with-metis=PETSc
```

As far as I could tell nothing surprising occurred when running `make`. The following error was thrown when running `make check`:

```
Making check in optimization/optimization_ex1
/Applications/Xcode.app/Contents/Developer/usr/bin/make  check-am
/Applications/Xcode.app/Contents/Developer/usr/bin/make  example-dbg example-devel example-opt   run.sh
make[4]: Nothing to be done for `../../../../examples/optimization/optimization_ex1/run.sh'.
  CXX      example_dbg-optimization_ex1.o
  CXX      example_devel-optimization_ex1.o
  CXX      example_opt-optimization_ex1.o
  CXXLD    example-dbg
  CXXLD    example-opt
[...]

PR #1898, opened 2018-10-15
Avoid modifying solution vector in TaoOptimizationSolver callbacks.

jwpeterson:
As discussed in #1889 (and previously in #630 and #1294) we cannot modify the solution vector by enforcing constraints in the PETSc callback functions but we can enforce the constraints on `current_local_solution`, which is not locked read only.

cc: @andfranklin


jwpeterson:
I verified using a local PETSc-debug build that the optimization solvers both run with the changes on this branch.

PR #1899, opened 2018-10-15
Fix more GCC 8.1 warnings

roystgnr:
I got buried in "a bunch of our software modules are broken" nonsense today, but at least some of the debugging there gave me an opportunity to see what gcc 8 thinks of a different libMesh configuration.


roystgnr:
Could we turn on ```--enable-paranoid-warnings --enable-werror``` for the ```Test enable complex``` build?

jwpeterson:
> Could we turn on `--enable-paranoid-warnings --enable-werror` for the `Test enable complex` build?

Sure... do you know if that passes currently?

roystgnr:
Oh, hell, no I don't.  The fixes here include a bunch of warnings that would be triggered by that build but I don't know if that list is exhaustive.  How about we just do ```--enable-paranoid-warnings``` and I'll check it manually and make sure it's clean here before we turn on the CI-kill-switch.

jwpeterson:
> How about we just do `--enable-paranoid-warnings` and I'll check it manually and make sure it's clean here before we turn on the CI-kill-switch.

I opened idaholab/civet_recipes#1063 with both flags, but at the moment it looks like the new unit test doesn't pass the complex build either...

jwpeterson:
Problem seems to be a linking error for some unsigned types...
```
parallel/unit_tests_opt-parallel_test.o: In function `ParallelTest::testMaxloc()':
parallel_test.C:(.text._ZN12ParallelTest10testMaxlocEv[_ZN12ParallelTest10testMaxlocEv]+0x19c): undefined reference to `int libMesh::Parallel::dataplusint_type<unsigned short>()'
parallel/unit_tests_opt-parallel_test.o: In function `ParallelTest::testMinloc()':
parallel_test.C:(.text._ZN12ParallelTest10testMinlocEv[_ZN12ParallelTest10testMinlocEv]+0x19c): undefined reference to `int libMesh::Parallel::dataplusint_type<unsigned int>()'
```

PR #1900, opened 2018-10-16
Avoid [-Wunused-private-field] warnings.

jwpeterson:
Several private members of the ExodusII_IO_Helper class are only used
in asserts, so libmesh_ignore() them to silence the compiler. Newer
versions of clang (6.0.1) started to warn about these.

### DIFF
--- a/include/mesh/exodusII_io_helper.h
+++ b/include/mesh/exodusII_io_helper.h
@@ -738,11 +738,11 @@ public:
   {
     // libmesh_ignore variables that are only used in asserts to avoid
     // compiler warnings.
-    libmesh_ignore(node_map_size);
-    libmesh_ignore(inverse_node_map_size);
-    libmesh_ignore(inverse_side_map_size);
-    libmesh_ignore(shellface_map_size);
-    libmesh_ignore(inverse_shellface_map_size);
+    libmesh_ignore(node_map_size,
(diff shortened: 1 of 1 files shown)


jwpeterson:
I'm just gonna wait and rebase this on #1899 once its merged :-)

roystgnr:
I only now realize that libmesh_dbg_var should have been named libmesh_devel_var.  Not sure what I was thinking.

roystgnr:
Actually, those members are all in the same class.  Can we just wrap their definitions in ```#ifndef NDEBUG```?

jwpeterson:
> Can we just wrap their definitions in `#ifndef NDEBUG`?

Sure.

jwpeterson:
> Can we just wrap their definitions in `#ifndef NDEBUG`?

That turned out to be pretty ugly/too many new lines of code unless I wanted to reorder class variables and initialization lists so I decided not to pursue it (the commit is still in the history). Since this is just to fix a warning, let's keep it as simple as possible. The `libmesh_ignores()` are also nicely forward-compatible: in case someone starts using one of the members outside of an assert, they won't have to also remember to go remove `#ifndef` tags.

roystgnr:
Ooh, I like moving all the ignores to one place.  Can you also use the new variadic libmesh_ignore there?

PR #1904, opened 2018-10-16
Do component_to_var lookups in OldSolution code

roystgnr:
This is to handle the case (as triggered all-too-easily in
https://github.com/idaholab/moose/issues/12309) where we have
multi-component variables which *precede* other single-component
variables in the system being projected.

I can't believe nobody ever ran into this before.  We have a habit of declaring SCALAR variables last but that should never have been a *requirement*.

### DIFF
--- a/include/systems/generic_projector.h
+++ b/include/systems/generic_projector.h
@@ -211,12 +211,22 @@ public:
     sys(sys_in),
     old_context(sys_in)
   {
+    // We'll be queried for components but we'll typically be looking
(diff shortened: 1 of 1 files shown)


roystgnr:
Going to merge this now since it at least fixes one instance of the problem for me; if @dschwen turns out to have a more complicated still-unfixed case in hand then we can revisit that later.

dschwen:
Thanks @roystgnr

PR #1906, opened 2018-10-17
Finer ghosting control

roystgnr:
EquationSystems::enable_default_ghosting(false/true) will now remove/restore default ghosting functors created by Mesh and DofMap objects, and will disable/reenable the creation of default ghosting functors by the creations of new add_system calls.

@friedmud you can take a look at these first 5 commits and see if enable_default_ghosting(false) is a start for your codes; the finer DofMap method control will come in later commits and the unit testing to see if any of this is actually working will be later still.

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -266,10 +266,16 @@ public:
    * for codes where no evaluations on neighbor cells (e.g. no jump
    * error estimators) are done.
    *
-   * Default ghosting can be restored manually, or is restored
+   * Defaults can be restored manually via add_default_ghosting(), or
    * automatically if clear() returns the DofMap to a default state.
    */
   void remove_default_ghosting();
+
+  /**
+   * Add the default functor(s) for coupling and algebraic ghosting.
+   * User-added ghosting functors will be unaffected.
+   */
+  void add_default_ghosting();
 
   /**
    * Adds a functor which can specify coupling requirements for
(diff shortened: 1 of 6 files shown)


friedmud:
@roystgnr Thanks for all of this - trying it now.

friedmud:
With all ghosting turned off... running the MOOSE test suite with 4 MPI procs... there are only ~50 failures and 987 passed tests.

So: as suspected most problems simply don't need ghosting.  And: it's trivial to see the sets of tests that do nee ghosting:

1.  Adaptive
2.  Multiapp Transfers
3.  Interface Kernels
4.  DG

So: we'll just add ghosting when we're using those features (which is what our RelationshipManager system does) and we'll be golden.

For now - I'm going to an input file option `Problem/default_ghosting` which I'm going to default to `true` for now.  That will allow us to turn off default ghosting for problems where we know that's important and turn it off to work toward a time when we can default it off.

I definitely think this particular PR can be merged.

roystgnr:
Let me finish up some unit tests and make sure CI is happy with those too and I'll merge tonight.

roystgnr:
While working on this I've noticed an embarrassing over-ghosting regression, but I'll try that fix out in a separate PR.  Hopefully a fix won't break anything at all but it's reassuring to see an upper bound of 50 failures.

friedmud:
Sounds good - I'm working on the MOOSE PR that will use this capability... it will obviously break until you merge this and then I'll set it to have a libMesh submodule update.

PR #1907, opened 2018-10-17
Add reverse order * and / ops for type_n_tensor compilation

lindsayad:
This is for compilation only; just like the `type_n_tensor` code, there is no effort to actually compute anything (accurately).


roystgnr:
Should we be throwing libmesh_not_implemented() for these, then?

lindsayad:
good plan

PR #1908, opened 2018-10-18
Upgrade MetaPhysicL to version 0.3.1

roystgnr:
https://github.com/roystgnr/MetaPhysicL/pull/22 would be good to
incorporate right away, assuming @lindsayad sees no problems.


jwpeterson:
This was definitely bootstrapped using the libmesh autotools versions and not the one-off Ubuntu packages? I can double-check that if you want...

roystgnr:
I checked my $PATH before bootstrap and I did a git diff in contrib/metaphysicl after, but yeah, I'd appreciate a double-check anyway.

jwpeterson:
./bootstrap is clean for me so :+1:

lindsayad:
I think this is good to go

PR #1911, opened 2018-10-18
Revert "Add version-check of MPI in configure"

roystgnr:
Reverts libMesh/libmesh#1822

We need to figure out what is going wonky after a bootstrapped build.


roystgnr:
Not going to wait for CI; we already *know* the previous state is a time bomb.

PR #1913, opened 2018-10-19
API: set_snesmf_reuse_base

### LINKED ISSUE #1912: Add API: set_snesmf_reuse_base()
https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/SNES/MatSNESMFSetReuseBase.html

It allows us to reuse some data. But for some physics, we can not reuse base, and hence we need an option to turn on/off.

fdkong:
Closes #1912


jwpeterson:
I'm guessing you will eventually also want a getter for this parameter, but this is probably fine as-is for now.

PR #1914, opened 2018-10-22
Create partial class specialization for DenseMatrix<DualNumber>

lindsayad:
Any instantiation of a `DenseMatrix<DualNumber>` currently fails because of missing symbols. We could fix this by explicitly instantiating in `dense_matrix.C`, but this will fail because we can't call through to PETSc routines with dual numbers. We cannot partially specialize a template class method to work with all possible `DualNumbers` so my solution at this point is to partially specialize the template class.

### DIFF
--- a/include/numerics/dense_matrix.h
+++ b/include/numerics/dense_matrix.h
@@ -30,6 +30,10 @@
 # include <petscsys.h>
 #endif
 
+#ifdef LIBMESH_HAVE_METAPHYSICL
+#include "metaphysicl/dualnumber.h"
+#endif
+
 // C++ includes
 #include <vector>
 #include <algorithm>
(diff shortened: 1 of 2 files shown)


jwpeterson:
So this might be the right way to handle things, but note that: 

1. The `DenseMatrixBase` API is fairly limited, so you will likely be re-inventing/duplicating stuff if you want to do any solves, decompositions, vector multiplication, etc. with DenseMatrices of DualNumbers.
1. We already have a similar situation with `DenseMatrix<std::complex>`, since we can't call into PETSc's BLAS/LAPACK implementations for those either. The entire dense_matrix_blas_lapack.C is therefore wrapped in
```
#if (LIBMESH_HAVE_PETSC && LIBMESH_USE_REAL_NUMBERS)
```
and we have hand-coded versions of things like `cholesky_solve()` that work for both real and complex numbers, and seem like they could potentially also work with `DualNumbers` (see dense_matrix_impl.h).

The `#ifdef` above is obviously not sufficient once we have `DenseMatrix<DualNumber>`, since presumably then `LIBMESH_USE_REAL_NUMBERS` would be defined at the same time, but perhaps this could be fixed up?

roystgnr:
This can't be the right way to do things, can it?  Aside from John's concerns, every libmesh_not_implemented() is a runtime timebomb waiting to go off.

We already avoid *calling* blas/lapack routines unless we're of a type for which they're defined, right?  So would it be possible to put some "cast data to Real or scream and die at runtime if our datatype is DualNumber" bits in that code to get them to compile?

pbauman:
Same comment here as #1916 - this didn't even attempt to notify femputer (no sent "payload" in webhook settings).

jwpeterson:
Yeah, the central server is what's down so the clients aren't getting any work to do...

pbauman:
I understand the INL ones are down, but I run my own server which should function independently. It's up and running, but it looks like GitHub is not even attempting to deliver to my server. Or did you mean GitHub's servers are down? Oh, GitHub's servers are down. https://status.github.com/messages OK. That's the deal.

lindsayad:
@roystgnr If you don't like this solution, then we will need to move everything into headers because it seems untenable to add an explicit instantiation in the `.C` file every time a different `DualNumber<T, D>` is needed.

roystgnr:
Yeah, that's the cost of compile-time polymorphism.  Fortunately we've already pretty much paid that cost - that's what dense_matrix_impl.h is for.
...
Wait a minute.  We created dense_matrix_impl.h to handle DenseMatrix of the MetaPhysicL::DynamicSparseNumberArray class we use in system_projection.C.  How is that sort of trick not working in your case?  It looks like all we had to do was manually instantiate the two functions we actually used.

lindsayad:
You don't actually instantiate a `DenseMatrix<DSANN>` you just instantiate a couple of template methods of `DenseMatrix<Real>` that take arguments related to the `DSANN`. If I explicitly instantiate a `DenseMatrix<DualNumber>` then I get compile errors out of methods like `min` and `max` which attempt conversions of type `T` to `Real`.

Of course I can't actually track down where the hell in MOOSE we are even implicitly instantiating `DenseMatrix<DualNumber>`. Does anyone have a good suggestion for tracking down the origin of undefined references? Errors like:
```
`/home/lindad/projects/moose/test/lib/libmoose_test-opt.so: undefined reference to `libMesh::DenseMatrix<MetaPhysicL::DualNumber<double, MetaPhysicL::NumberArray<100ul, double> > >::left_multiply(libMesh::DenseMatrixBase<MetaPhysicL::DualNumber<double, MetaPhysicL::NumberArray<100ul, double> > > const&)'
```
just aren't very illuminating

lindsayad:
Ok, I fixed this on the MOOSE side...found where I was instantiating, and there's just no need for it, or more strongly put, it just shouldn't be done. So I'm closing this...

PR #1916, opened 2018-10-22
SparseMatrix::add() ref should be const

### LINKED ISSUE #1915: Parameter in `libMesh::EigenSparseMatrix<T>::add` should be `const`
The argument `X` in `void libMesh::EigenSparseMatrix<T>::add(const T a, SparseMatrix<T> & X)` ought to be `const`.

[Here's](https://libmesh.github.io/doxygen/classlibMesh_1_1EigenSparseMatrix.html#a5d2bdb0f718d4b440941f8471672aba7) the part of the documentation I am looking at.

roystgnr:
This should resolve #1915 

I haven't tested this with Trilinos yet; I'll do so now.

### DIFF
--- a/src/numerics/petsc_matrix.C
+++ b/src/numerics/petsc_matrix.C
@@ -1104,7 +1104,7 @@ void PetscMatrix<T>::add_matrix(const DenseMatrix<T> & dm,
 
 
 template <typename T>
-void PetscMatrix<T>::add (const T a_in, SparseMatrix<T> & X_in)
+void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
 {
   libmesh_assert (this->initialized());
 
@@ -1114,7 +1114,7 @@ void PetscMatrix<T>::add (const T a_in, SparseMatrix<T> & X_in)
   libmesh_assert_equal_to (this->n(), X_in.n());
 
   PetscScalar a = static_cast<PetscScalar>      (a_in);
-  PetscMatrix<T> * X = cast_ptr<PetscMatrix<T> *> (&X_in);
+  const PetscMatrix<T> * X = cast_ptr<const PetscMatrix<T> *> (&X_in);
 
   libmesh_assert (X);
 
--- a/include/numerics/eigen_sparse_matrix.h
+++ b/include/numerics/eigen_sparse_matrix.h
@@ -122,7 +122,7 @@ public:
   virtual void add_matrix (const DenseMatrix<T> & dm,
                            const std::vector<numeric_index_type> & dof_indices) override;
 
-  virtual void add (const T a, SparseMatrix<T> & X) override;
+  virtual void add (const T a, const SparseMatrix<T> & X) override;
 
   virtual T operator () (const numeric_index_type i,
                          const numeric_index_type j) const override;
--- a/include/numerics/laspack_matrix.h
+++ b/include/numerics/laspack_matrix.h
@@ -139,7 +139,7 @@ public:
    * so a hand-coded version with hopefully acceptable performance
    * is provided.
    */
-  virtual void add (const T a, SparseMatrix<T> & X) override;
+  virtual void add (const T a, const SparseMatrix<T> & X) override;
 
   virtual T operator () (const numeric_index_type i,
                          const numeric_index_type j) const override;
--- a/include/numerics/petsc_matrix.h
+++ b/include/numerics/petsc_matrix.h
@@ -205,7 +205,7 @@ public:
    * \note \p X will be closed, if not already done, before performing
    * any work.
    */
-  virtual void add (const T a, SparseMatrix<T> & X) override;
+  virtual void add (const T a, const SparseMatrix<T> & X) override;
 
   virtual T operator () (const numeric_index_type i,
                          const numeric_index_type j) const override;
--- a/include/numerics/sparse_matrix.h
+++ b/include/numerics/sparse_matrix.h
@@ -257,7 +257,7 @@ public:
   /**
    * Compute \f$ A \leftarrow A + a*X \f$ for scalar \p a, matrix \p X.
    */
-  virtual void add (const T a, SparseMatrix<T> & X) = 0;
+  virtual void add (const T a, const SparseMatrix<T> & X) = 0;
 
   /**
    * \returns A copy of matrix entry \p (i,j).
--- a/include/numerics/trilinos_epetra_matrix.h
+++ b/include/numerics/trilinos_epetra_matrix.h
@@ -159,7 +159,7 @@ public:
    * \note \p X will be closed, if not already done, before performing
    * any work.
    */
-  virtual void add (const T a, SparseMatrix<T> & X) override;
+  virtual void add (const T a, const SparseMatrix<T> & X) override;
 
   virtual T operator () (const numeric_index_type i,
                          const numeric_index_type j) const override;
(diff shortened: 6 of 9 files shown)


jwpeterson:
Oh, nice, uh I hope you didn't step on @dmcdougall's toes though.

roystgnr:
I told him over IM that I was already setting up a PR.  I think he just didn't trust me to credit him unless he opened the ticket himself.  ;-)

roystgnr:
Either that, or the "PRs should be preceded by Issues" MOOSE policy is an actual good idea that good developers come to independently, and which I'm merely weird and lazy for ever being annoyed by...

roystgnr:
So... what happened to our CI?

roystgnr:
Ah, I remember seeing that email, but didn't realize it applied to CI too.  I was expecting *some* kind of "CI pending waiting on foo" message rather than nothing, though.  And what about @pbauman and femputer?

src/numerics/petsc_matrix.C
```diff
@@ -1114,7 +1114,7 @@ void PetscMatrix<T>::add (const T a_in, SparseMatrix<T> & X_in)
   libmesh_assert_equal_to (this->n(), X_in.n());
 
   PetscScalar a = static_cast<PetscScalar>      (a_in);
-  PetscMatrix<T> * X = cast_ptr<PetscMatrix<T> *> (&X_in);
+  const PetscMatrix<T> * X = cast_ptr<const PetscMatrix<T> *> (&X_in);
```
dmcdougall: So, @jwpeterson's comment in #1915 about PETSc expecting a non-`const` pointer to a `Mat` isn't an issue here?

roystgnr: Hmmm... my PETSc seemed happy with it but I suppose I should dig through old docs to make sure that wasn't a super recent change.

pbauman:
> And what about @pbauman and femputer?

WTF? I just looked in the settings and it didn't try to deliver to femputer that there was a PR. In fact, some settings were changed (SSL verification turned off). Did someone mess with the webhook settings? Is GitHub screwing up?

src/numerics/petsc_matrix.C
roystgnr: Ah, I get it.  A Mat is a pointer-to-PETSc-stuff, and even if our PetscMatrix is const it can still give us a const-pointer-to-non-const-PETSc-stuff, so it doesn't matter whether or not PETSc's API here is philosophically const-correct (which is good, because it isn't).

dmcdougall: I've read this comment about 14 times and I'm still confused.

The change you're making on this line makes `X` a pointer-to-(const-PetscMatrix).  The thing you pass to PETSc is `X->_mat` which is of type `Mat`.  The PETSc API doesn't guarantee that `X->_mat` won't change during the call to `MatAXPY`.  So as far as I understand this should not compile.  What am I missing?

roystgnr:
Builds fine with Trilinos too.  Their API here seems const-friendly for as far back as we support.

src/numerics/petsc_matrix.C
roystgnr: C++ changes that X->_mat won't change during the call, because what MatAXPY gets is a *copy* of X->_mat.  Under the hood that Mat is itself a pointer to something else, which means the something else can change, even if (our copy of) the Mat can't.

dmcdougall: Oh, right.  Of course.  Nothing to see here.  Move along.

pbauman:
OK, in theory femputer is up and running again.

friedmud:
I think I'm seeing some build failure because of this... on Falcon using GCC 7.3:

```
src/numerics/laspack_matrix.C: In instantiation of ‘void libMesh::LaspackMatrix<T>::add(T, const libMesh::SparseMatrix<T>&) [with T = double]’:
src/numerics/laspack_matrix.C:469:16:   required from here
src/numerics/laspack_matrix.C:407:45: error: invalid conversion from ‘const QMatrix* {aka const QMatrixType*}’ to ‘QMatrix* {aka QMatrixType*}’ [-fpermissive]
       libmesh_assert_equal_to (len, Q_GetLen(&(X->_QMat), row+1));
                                     ~~~~~~~~^~~~~~~~~~~~~~~~~~~~
./include/libmesh/libmesh_common.h:294:20: note: in definition of macro ‘libmesh_assert_equal_to_msg’
     if (!(expr1 == expr2)) {                                            \
                    ^~~~~
src/numerics/laspack_matrix.C:407:7: note: in expansion of macro ‘libmesh_assert_equal_to’
       libmesh_assert_equal_to (len, Q_GetLen(&(X->_QMat), row+1));
       ^~~~~~~~~~~~~~~~~~~~~~~
In file included from ./include/libmesh/laspack_matrix.h:31:0,
                 from src/numerics/laspack_matrix.C:27:
./contrib/laspack/qmatrix.h:68:8: note:   initializing argument 1 of ‘size_t Q_GetLen(QMatrix*, size_t)’
 size_t Q_GetLen(QMatrix *Q, size_t RoC);
        ^~~~~~~~
In file included from ./include/libmesh/libmesh.h:25:0,
                 from ./include/libmesh/sparse_matrix.h:25,
                 from ./include/libmesh/laspack_matrix.h:28,
                 from src/numerics/laspack_matrix.C:27:
src/numerics/laspack_matrix.C:407:45: error: invalid conversion from ‘const QMatrix* {aka const QMatrixType*}’ to ‘QMatrix* {aka QMatrixType*}’ [-fpermissive]
       libmesh_assert_equal_to (len, Q_GetLen(&(X->_QMat), row+1));
./include/libmesh/libmesh_common.h:295:121: note: in definition of macro ‘libmesh_assert_equal_to_msg’
       libMesh::err << "Assertion `" #expr1 " == " #expr2 "' failed.\n" #expr1 " = " << (expr1) << "\n" #expr2 " = " << (expr2) << '\n' << msg << std::endl; \
                                                                                                                         ^~~~~
src/numerics/laspack_matrix.C:407:7: note: in expansion of macro ‘libmesh_assert_equal_to’
       libmesh_assert_equal_to (len, Q_GetLen(&(X->_QMat), row+1));
       ^~~~~~~~~~~~~~~~~~~~~~~
In file included from ./include/libmesh/laspack_matrix.h:31:0,
                 from src/numerics/laspack_matrix.C:27:
./contrib/laspack/qmatrix.h:68:8: note:   initializing argument 1 of ‘size_t Q_GetLen(QMatrix*, size_t)’
 size_t Q_GetLen(QMatrix *Q, size_t RoC);
```

friedmud:
Reverting 5c8f9e92c74ab634 seems to have fixed it.

roystgnr:
Oh, hell.  I double-checked the Trilinos build but I forgot that we're disabling non-LGPL-compatible stuff by default now and that Laspack isn't LGPL-compatible.  I'll fix this, but we ought to add --disable-strict-lgpl to one of the CI configurations to expand coverage a bit.

jwpeterson:
> we ought to add --disable-strict-lgpl to one of the CI configurations to expand coverage a bit.

There's an optional "Test Non-LGPL Contrib" target. I could make this run for every PR if you prefer...

roystgnr:
Ah, I forgot about that.  No, an extra target every PR sounds like overkill.

roystgnr:
I just need to remember to turn it on when we're actually touching that code.

PR #1918, opened 2018-10-23
Partitioner unit tests

### LINKED ISSUE #1905: Partitioning a 0 element mesh throws an assert
Currently, if someone calls `prepare_for_use()` on an empty (zero element) mesh, the number of partitions gets set to 0 by this line:

https://github.com/libMesh/libmesh/blob/5e3620b01bc1ab478de8ae8c6786714095f2148a/src/partitioning/partitioner.C#L69-L71

Then we end up calling `_do_partition` with `n_parts==0`,

https://github.com/libMesh/libmesh/blob/5e3620b01bc1ab478de8ae8c6786714095f2148a/src/partitioning/partitioner.C#L86

and this eventually triggers an assert in `MetisPartitioner`:

https://github.com/libMesh/libmesh/blob/5e3620b01bc1ab478de8ae8c6786714095f2148a/src/partitioning/metis_partitioner.C#L61

It seems to me, however, that this doesn't need to be an error, but simply an early return. @leorachapuis ran into this while working on some mesh generation stuff...

roystgnr:
IIRC we have an open ticket for "METIS sucks", but I didn't realize how bad it was until I was futzing around to fix #1905.  What kind of partitioners can't figure out how to put 27 elements into even 14 nonempty groups?  Is there something wrong with our shims??

Anyway, this PR fixes a couple easier minor bugs, and adds unit tests for at least the level of capability that we *can* expect.

### DIFF
--- a/src/partitioning/centroid_partitioner.C
+++ b/src/partitioning/centroid_partitioner.C
@@ -31,12 +31,19 @@ void CentroidPartitioner::partition_range(MeshBase & mesh,
                                           MeshBase::element_iterator end,
                                           unsigned int n)
 {
-  // Check for an easy return
+  // Check for easy returns
+  if (it == end)
+    return;
+
   if (n == 1)
     {
       this->single_partition_range (it, end);
       return;
     }
+
+  // Make sure the user has not handed us an
+  // invalid number of partitions.
+  libmesh_assert_greater (n, 0);
 
   // We don't yet support distributed meshes with this Partitioner
   if (!mesh.is_serial())
@@ -92,10 +99,6 @@ void CentroidPartitioner::partition_range(MeshBase & mesh,
     default:
       libmesh_error_msg("Unknown sort method: " << this->sort_method());
     }
-
-  // Make sure the user has not handed us an
-  // invalid number of partitions.
-  libmesh_assert_greater (n, 0);
 
   // Compute target_size, the approximate number of elements on each processor.
   const dof_id_type target_size = cast_int<dof_id_type>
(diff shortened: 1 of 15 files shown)


roystgnr:
I *thought* this was going to be dead-simple, but it took me enough iterations that I don't think I want to merge until it gets another set of eyes, even if CI is finally happy.

roystgnr:
> This might be OK, though as it looks like you are not diffing against "gold" files or asserting that a particular element ends up with a particular processor ID?

Oh, definitely not.  My *goal* with tests is to always have them pass regardless of processor count or partitioner choice or phase of the moon.

> I'm guessing you developed the new unit tests on Linux and then CIVET ran them on a Linux box... any chance you've tested them on a Mac? I might be able to give it a shot later today.

Yes, yes, and no, respectively.  That's a very good idea; it'd be embarrassing if METIS performed even worse on Mac and that tripped the tests and that didn't show up until after we'd merged.  Is the Test Mac  CI option enough?  If not then we'll want to wait for you to try it by hand.

jwpeterson:
> Is the Test Mac CI option enough?

That will test with a pretty new clang compiler on Mac and should be sufficient, yeah. Thanks for remembering we had that optional recipe.

roystgnr:
Hmm.. as my finger was clicking on the merge button, my brain was clicking on a possible problem: that Test Mac run surely only hits one processor count.  If you could do something like
```
for num in $(seq 1 24); do LIBMESH_RUN="mpiexec -n $num" make -C tests/ check || break; done
```
on Mac some time I'd feel much relieved to hear that it still passes.

jwpeterson:
OK, I can try that out.

jwpeterson:
That command worked for me so I'd say we're good.

PR #1910, opened 2018-10-18
Re-bootstrap after #1822

roystgnr:
Can't believe we missed this...

Issue #1827, opened 2018-08-20
MPI-3


friedmud:
We should probably start to at least test for MPI-3... even if, for now, all we do is switch some internal algorithms.

Here is a good list of the new capabilities we could look for:

https://htor.inf.ethz.ch/blog/index.php/2012/02/06/mpi-3-0-is-coming-an-overview-of-new-and-old-features/

Hi on my personal list are:

`Ibarrier`, `Ireduce`, `Iallreduce`, `mprobe`, `mrecv`

Stretch goals with be one-sided communication and windowing - I don't have plans to use those in libMesh right now though.

It would be great to get a configure test for this stuff... so we can start to see what people have on their machines.

pbauman:
> We should probably start to at least test for MPI-3... even if, for now, all we do is switch some internal algorithms.

Agree 100%. Could we maybe log either here or in #1826 or somewhere what MPI library/version support MPI3? Mpich, Mvapich, and Intel-MPI are the most commonly used around here. I don't have any sense of version cutoffs etc.

roystgnr:
Pinging @BalticPinguin for opinions, since #1822 is obviously relevant.  If nearly everybody is on MPI-3 then perhaps we could make MPI-2 support non-default for a while, with something like ```configure --enable-mpi2``` required to build with it and a ```MPI-2 support is deprecated; please reconfigure with --enable-mpi2 and email libmesh-users@``` message printed for people who don't have MPI-3 support?  That way we could avoid maintaining multiple code paths for too long but not leave anyone in the lurch.

Oh, and if you guys want to join me in reviewing #1822 ... between vacation and backlog of other things I've barely glanced at it yet, but ideally I'd like to drop support for MPI-2 only after we *finish* dropping support for MPI-1.

BalticPinguin:
I like this idea as well.
If you agree on the testing-scheme suggested in #1822 , this can be extended straight forward to test for MPI-3 as well...

BalticPinguin:
ok; probably it is good to shift the discussion from #1822 to this thread:
As I see it, the MPI-config currently is done as follows:
1. Try to find PETSc with MPI and take theirs without further checks
2. If this doesn't work, search `mpi.h` and take this 
3. If it is not found, try whether the linker/compiler can find it and if it works, take it as it is.
4. configure without MPI.

According to my understanding, 
  --> in case of 1 and 3 we cannot assume to have access to the header-file explicitly. But if this works, 
[...]

PR #1920, opened 2018-10-24
"Added version-check for the MPI-version in use."

BalticPinguin:
Let me have another try with the MPI-version check, hopefully resolving the problems in #1910.
Since this is anyway a new PR, I hope that there is no problem that I squashed the commits in #1822.
It is basically the changes in #1822 but with the additional check to reject PETSc if needed.

Also, this time I added the new configure to trigger problems in the civet-checks.

### DIFF
--- a/configure
+++ b/configure
@@ -33126,7 +33126,7 @@ $as_echo "<<< PETSc did not define MPIEXEC.  Will try configuring MPI now... >>>
 
 
 if test "x$MPIHOME" = "x"; then :
-  MPIHOME="/usr"
+  MPIHOME="/.."
 fi
 
 
@@ -33135,13 +33135,40 @@ if test "${with_mpi+set}" = set; then :
   withval=$with_mpi; MPI="$withval"
 else
 
-              $as_echo "note: MPI library path not given... trying prefix=$MPIHOME"
+              $as_echo "note: MPI library path not given... trying to find it myself"
               MPI=$MPIHOME
 
 fi
 
 
 if test -z "$MPI"; then :
+  MPI="/.."
+fi
+
+if test "$MPI" = "/.."; then :
+
+    for i in $(echo $LD_LIBRARY_PATH | tr ':' '\n')
+    do
+       if  test -e $i/libmpi.a || test -e $MPI_LIBS_PATH/libmpi.so ; then :
+  MPI=$(dirname $i)
+       break
+fi
+    done
+
+fi
+if test "$MPI" = "/.."; then :
+
+    for i in $(echo $INCLUDE | tr ':' '\n')
+    do
+       if  test -e $i/mpi.h ; then :
+  MPI=$(dirname $i)
+       break
+fi
+    done
+
+fi
+
+if test "$MPI" = "/.."; then :
   MPI="/usr"
 fi
 
@@ -33532,9 +33559,8 @@ ac_compiler_gnu=$ac_cv_cxx_compiler_gnu
                 CPPFLAGS="-I$MPI_INCLUDES_PATH $CPPFLAGS"
                 ac_fn_cxx_check_header_mongrel "$LINENO" "mpi.h" "ac_cv_header_mpi_h" "$ac_includes_default"
 if test "x$ac_cv_header_mpi_h" = xyes; then :
-
-$as_echo "#define HAVE_MPI 1" >>confdefs.h
-
+  { $as_echo "$as_me:${as_lineno-$LINENO}: result: MPI-headers are working as expected." >&5
+$as_echo "MPI-headers are working as expected." >&6; }
 else
   { $as_echo "$as_me:${as_lineno-$LINENO}: result: Could not compile in the MPI headers..." >&5
 $as_echo "Could not compile in the MPI headers..." >&6; }; enablempi=no
@@ -33565,16 +33591,14 @@ ac_compiler_gnu=$ac_cv_cxx_compiler_gnu
                 CPPFLAGS="-I$MPI_INCLUDES_PATH $CPPFLAGS"
                 ac_fn_cxx_check_header_mongrel "$LINENO" "mpi.h" "ac_cv_header_mpi_h" "$ac_includes_default"
 if test "x$ac_cv_header_mpi_h" = xyes; then :
-
-$as_echo "#define HAVE_MPI 1" >>confdefs.h
-
+  { $as_echo "$as_me:${as_lineno-$LINENO}: result: MPI-headers are working as expected." >&5
+$as_echo "MPI-headers are working as expected." >&6; }
 else
   { $as_echo "$as_me:${as_lineno-$LINENO}: result: Could not compile in the MPI headers..." >&5
 $as_echo "Could not compile in the MPI headers..." >&6; }; enablempi=no
 fi
 
-
-                MPI_INCLUDES_PATHS="-I$MPI_INCLUDES_PATH"
+ MPI_INCLUDES_PATHS="-I$MPI_INCLUDES_PATH"
                 ac_ext=cpp
 ac_cpp='$CXXCPP $CPPFLAGS'
 ac_compile='$CXX -c $CXXFLAGS $CPPFLAGS conftest.$ac_ext >&5'
@@ -33610,9 +33634,6 @@ if ac_fn_cxx_try_link "$LINENO"; then :
                        { $as_echo "$as_me:${as_lineno-$LINENO}: result: $CXX Compiler Supports MPI " >&5
 $as_echo "$CXX Compiler Supports MPI " >&6; }
 
-$as_echo "#define HAVE_MPI 1" >>confdefs.h
-
-
 else
   { $as_echo "$as_me:${as_lineno-$LINENO}: result: $CXX Compiler Does NOT Support MPI..." >&5
 $as_echo "$CXX Compiler Does NOT Support MPI..." >&6; }; enablempi=no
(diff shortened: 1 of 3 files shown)


jwpeterson:
> I hope that there is no problem that I squashed the commits in #1822.

It's fine. We generally try to keep the build script changes produced by running `bootstrap` in a separate commit, but it's not strictly necessary.

BalticPinguin:
Ok, great. If this is accepted, I can re-push without the configure-update. I just wanted to avoid the situation as before that the actual changes are not tested...

jwpeterson:
Yeah, the changes to configure definitely need to be on the branch that gets merged, they should just be in a separate commit with a message like "Run bootstrap." or something along those lines.

BalticPinguin:
I removed two commas that were wrong;
the failing GRINS seems to be rather in the test-suit, I guess: The GRINS-test uses PETSc, no matter if libMesh does; at least it looks like this for me...

pbauman:
GRINS failures here are us not properly guarding the lack of `PetscDiffSolver`.  We need to fix that.

jwpeterson:
> GRINS failures here are us not properly guarding the lack of `PetscDiffSolver`. We need to fix that.

True, but I guess we don't want this new configure test to disable a previously-working PETSc build like this in the first place???

roystgnr:
Yeah - is that an overzealous test or is that really properly rejecting an outdated MPI?

pbauman:
<pre>
Currently Loaded Modules:
  1) autotools/1.0       6) gcc/6.2.0         11) hdf5/1.8.16
  2) git/2.8.1           7) mpich/3.1.4       12) vtk-dbg/7.1.0
  3) gdb/7.11            8) openblas/0.2.18   13) cppunit-dbg/1.13.2
  4) valgrind/3.11.0     9) petsc/3.7.4-opt   14) gsl/1.16
  5) anaconda/anaconda  10) boost-dbg/1.61.0  15) antioch-dbg/master
</pre>

I would've though this was a valid MPI, but is it not now? I'm behind on following this PR.

jwpeterson:
Would be good to see the output of your config.log if you can get ahold of it. There may be a print statement about what went wrong/why PETSc failed to be found.

roystgnr:
```
Checking Version of MPI now:
configure: WARNING: "MPI-version seems to be too low: Need MPI 2.X or 3.X. Disable MPI now..."
configure: WARNING: MPI is disabled. Disable PETSc as well.
```
Appears to be the important bit.

roystgnr:
Oh, wait, that's just screen output.

jwpeterson:
Yeah, but it's enough. There's no way that mpich/3.1.4 isn't 2.X compatible is there? I'll check now.

roystgnr:
My mpich2/3.1.4 module has in mpi.h:
```
#define MPI_VERSION    3
#define MPI_SUBVERSION 0
```
on line 455.

jwpeterson:
Ah, that's it then. "mpich" vs. "mpich2":
```
peterson(~/software/mpich/mpich-3.1.4)$ find . -name '*.h' -type f | xargs grep MPI_VERSION
./test/mpi/util/mtest_datatype.h:#ifndef MPI_VERSION
./test/mpi/util/mtest_datatype.h:#define MPI_VERSION 1
./test/mpi/include/mpitest.h: *     make CPPFLAGS='-DMTEST_MPI_VERSION=X -DMTEST_MPI_SUBVERSION=Y'
./test/mpi/include/mpitest.h:#ifndef MTEST_MPI_VERSION
./test/mpi/include/mpitest.h:#define MTEST_MPI_VERSION MPI_VERSION
./test/mpi/include/mpitest.h:#if MTEST_HAVE_MIN_MPI_VERSION(2,2)
./test/mpi/include/mpitest.h:#define MTEST_HAVE_MIN_MPI_VERSION(major_,minor_) \
./test/mpi/include/mpitest.h:    ((MTEST_MPI_VERSION == (major_) && MTEST_MPI_SUBVERSION >= (minor_)) ||   \
./test/mpi/include/mpitest.h:    (MTEST_MPI_VERSION > (major_)))
./src/mpi/romio/adio/include/mpipr.h:#ifndef USE_MPI_VERSIONS
```

roystgnr:
Huh.  If @pbauman is using the same, then I'd actually count this as a point in favor of this PR.  If we're going wild with newer MPI features then I think it really would be safest to reject any implementations that manage to compile but that don't actually claim to support those features.

roystgnr:
Well, there's some software that precisely supports a standard and that only adds half-complete support for newer standards if you enable a non-default option (e.g. gcc -std=cxx1z), and then there's other software that adds dribs and drabs of newer standards, as they're implemented, immediately adding each change to the default behavior (e.g. nearly everything everywhere); mpich 1 was probably in the latter category.

jwpeterson:
> mpich 1 was probably in the latter category

Well, in that case I tend to doubt the usefulness of any kind of blanket `MPI_VERSION` check then. We need specific configure tests for MPI features we are already using (#1822) and ones we'd like to use at some point (#1827) if we want to avoid false negatives like this particular case.

jwpeterson:
Note to self: [Our `MPI_Pack` usage is the one that requires MPI-3 standard](https://github.com/libMesh/libmesh/pull/1822#discussion_r211622727).

pbauman:
No, it's mpich2, I just don't have in the module name (since years and years ago).

From the `mpi.h` file:

<pre>
#define MPI_VERSION    3                                                                                                                                                                                                                      
#define MPI_SUBVERSION 0
#define MPICH_NAME     3
#define MPICH         1
</pre>

so it seems to been an erroneous exclusion. Let me try and dig up the config.log file.

BalticPinguin:
Independent from the failure in these checks, I have accidantally found another issue: In the petsc.m4, it sais
``
47   # Trump --enable-petsc with --disable-mpi
 48   AS_IF([test "x$enablempi" = xno],
 49         [enablepetsc=no])
 50 
``
which is circumvented by the current PR; so I need to change it to conform with it.
BTW:  What is the reason for this? (if you remember it; seems to be >4 years old...}

BalticPinguin:
So, I have some updates;
I tested against some mpich-libraries as well, but could not reproduce this behaviour.
Anyway, I realized that the mpi.m4-file in the previous PR was an old version; I don't understand how it came there...
Anyway, I would not expect that the behaviour changes, but most importantly the output should be more verbose. Maybe this helps understanding the issue...

BalticPinguin:
the actual error message is
```
transfers/multiapp_nearest_node_transfer.parallel: Working Directory: /opt/civet/build_0/moose/test/tests/transfers/multiapp_nearest_node_transfer
transfers/multiapp_nearest_node_transfer.parallel: Running command: mpiexec -n 2 /opt/civet/build_0/moose/test/moose_test-dbg -i parallel_master.i --error --error-unused --error-override --no-gdb-backtrace
transfers/multiapp_nearest_node_transfer.parallel: 
transfers/multiapp_nearest_node_transfer.parallel: [mpiexec@hpcbuild4] HYD_pmcd_pmiserv_proxy_init_cb (../../../../src/pm/hydra/pm/pmiserv/pmiserv_cb.c:459): cannot find proxy with ID 33620758
transfers/multiapp_nearest_node_transfer.parallel: [mpiexec@hpcbuild4] HYDT_dmxu_poll_wait_for_event (../../../../src/pm/hydra/tools/demux/demux_poll.c:76): callback returned error status
transfers/multiapp_nearest_node_transfer.parallel: [mpiexec@hpcbuild4] HYD_pmci_wait_for_completion (../../../../src/pm/hydra/pm/pmiserv/pmiserv_pmci.c:198): error waiting for event
transfers/multiapp_nearest_node_transfer.parallel: [mpiexec@hpcbuild4] main (../../../../src/pm/hydra/ui/mpich/mpiexec.c:344): process manager error waiting for completion
transfers/multiapp_nearest_node_transfer.parallel: 
transfers/multiapp_nearest_node_transfer.parallel: ################################################################################
transfers/multiapp_nearest_node_transfer.parallel: Tester failed, reason: CRASH
transfers/multiapp_nearest_node_transfer.parallel: 
    FAIL transfers/multiapp_nearest_node_transfer.parallel FAILED (CRASH)
```
maybe SO can help me understanding this message; but to m understanding (and my [google-expertise](https://lists.mpich.org/pipermail/discuss/2012-December/000205.html) this should not be directly related to the code ?

permcody:
That seems like a spurious failure. It's odd that there isn't any output at all so it's likely something "crashed" before it even started the test. I wouldn't worry about it.

BalticPinguin:
hmm... I would assume that something is broken in the config?
At least at the place where the `VERSION_MPI`-part is called, we seem to test against some other MPI:
```
Checking Version of MPI now:
configure: WARNING: "ERROR: MPI_VERSION is not defined: You seem to have MPI< 1.5 but need MPI 2.X or compatible. Disable MPI now..."
configure: WARNING: "MPICH is not defined."
configure: WARNING: "MPICH is version 3."
configure: WARNING: "MPI 3 is not supported."
```
I realized that the test for MPICH_NAME is not working: it gives this message also, when MPICH_NAME is undefined. (I read somewhere that it would evaluate to 0 than...) Therefore, the "MPICH is version 3" part is not valid (at least misleading).

The idea behind that test is, that my `mpi.h` (mpich-3.2.1) has:
```
 457 #define MPI_VERSION    3
 458 #define MPI_SUBVERSION 1
 459 #define MPICH_NAME     3
 460 #define MPICH         1
 461 #define MPICH_HAS_C2F  1
 462 
```
Any thoughts are apprechiated.

PR #1921, opened 2018-10-24
Add BoundingBox::signed_distance().

jwpeterson:
Sign convention:
1. If Point p is inside the BoundingBox B: d(p, B) < 0
1. If Point p is outside the BoundingBox B: d(p, B) > 0
1. If Point p is on the BoundingBox B: d(p, B) = 0

@roystgnr let me know if you think the formulas/approach seem reasonable. We talked about the "point outside" case; the "point inside" case also seems pretty straightforward. Also adds some simple unit tests.

### DIFF
--- a/include/geom/bounding_box.h
+++ b/include/geom/bounding_box.h
@@ -58,7 +58,7 @@ public:
     this->invalidate();
   }
 
-  /*
+  /**
    * Sets the bounding box to encompass the universe.
    */
   void invalidate ()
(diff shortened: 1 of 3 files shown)


jwpeterson:
> This breaks --enable-1D-only, but we've had a slew of other regressions since I stopped doing CI on that. I'll fix them all together at some point.

Looks like it wouldn't work great with `LIBMESH_DIM==2` either now that you mention it. Since I have to invalidate the "Test debug" target anyway, I'll take a crack at fixing that too...

PR #1922, opened 2018-10-26
1D/2D-only fixes

roystgnr:
This gets all our unit tests working for me in --enable-1D-only and --enable-2D-only configurations, it adds additional assertions protecting TypeTensor users there, and it reenables a few methods (one critical) that were previously unusable there.


roystgnr:
This should be ready to merge once Civet signs off.

PR #1924, opened 2018-10-28
Homepage: mention Github issues and pull requests

roystgnr:
The Github tracker system is both more accessible to users and more
convenient to developers than the old mailing lists, so we shouldn't
only be recommending the latter.


jwpeterson:
Looks good, I will also update the website repo with this change.

PR #1926, opened 2018-11-01
Assert that PetscVector is closed before calling _get_array().

jwpeterson:
@fdkong ran into an issue with this recently that was fixed by
ensuring the vector was closed before trying to get values from it,
but it originally looked like some kind of threading issue even in
debug mode. This assert should let us figure out the same problem
faster in the future.

### DIFF
--- a/src/numerics/petsc_vector.C
+++ b/src/numerics/petsc_vector.C
@@ -1345,6 +1345,7 @@ template <typename T>
 void PetscVector<T>::_get_array(bool read_only) const
 {
   libmesh_assert (this->initialized());
+  libmesh_assert (this->closed());
 
 #ifdef LIBMESH_HAVE_CXX11_THREAD
   std::atomic_thread_fence(std::memory_order_acquire);


jwpeterson:
Interesting... looks like we've been calling `_get_array()` on unclosed vectors on at least this example... I'll see if i can also get a stack trace. 
```
* Running Example adaptivity_ex2:
*  mpiexec -np 4 ./example-dbg -n_timesteps 25 -n_refinements 5 -output_freq 10 -init_timestep 0 
 
Usage:
	 /opt/civet/build_0/libmesh/build/examples/adaptivity/adaptivity_ex2/.libs/lt-example-dbg -init_timestep 0 -n_timesteps 25 [-n_refinements 5]
OR
	 /opt/civet/build_0/libmesh/build/examples/adaptivity/adaptivity_ex2/.libs/lt-example-dbg -read_solution -init_timestep 26 -n_timesteps 25

Running: /opt/civet/build_0/libmesh/build/examples/adaptivity/adaptivity_ex2/.libs/lt-example-dbg -n_timesteps 25 -n_refinements 5 -output_freq 10 -init_timestep 0

 Mesh Information:
  elem_dimensions()={2}
  spatial_dimension()=2
  n_nodes()=6273
    n_local_nodes()=1450
  n_elem()=13650
    n_local_elem()=3677
    n_active_elem()=10240
  n_subdomains()=2
  n_partitions()=4
  n_processors()=4
  n_threads()=1
  processor_id()=0

 EquationSystems
  n_systems()=1
   System #0, "Convection-Diffusion"
    Type "TransientLinearImplicit"
    Variables="u" 
    Finite Element Types="LAGRANGE" 
    Approximation Orders="FIRST" 
    n_dofs()=6273
    n_local_dofs()=1450
    n_constrained_dofs()=0
    n_local_constrained_dofs()=0
    n_vectors()=3
    n_matrices()=1
    DofMap Sparsity
      Average  On-Processor Bandwidth <= 7.52001
      Average Off-Processor Bandwidth <= 0.142197
      Maximum  On-Processor Bandwidth <= 11
      Maximum Off-Processor Bandwidth <= 6
    DofMap Constraints
      Number of DoF Constraints = 0

 Solving time step  0, time=0.0250...
H1 norm = 1.58843
  Refining the mesh...
Assertion `this->closed()' failed.
```

jwpeterson:
The assert seems to be tripped during a call to `DofMap::enforce_constraints_exactly()` inside of `System::project_vector()`. It's possible we just need to add a close() in the `project_vector()` routine...
```
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
  * frame #0: 0x000000010883e080 libmpi.12.dylib`MPI_Abort
    frame #1: 0x000000010044499b libmesh_dbg.0.dylib`libMesh::libmesh_terminate_handler() at libmesh.C:316
    frame #2: 0x00007fff6415cdfe libc++abi.dylib`std::__terminate(void (*)()) + 8
    frame #3: 0x00007fff6415cbe2 libc++abi.dylib`__cxxabiv1::failed_throw(__cxxabiv1::__cxa_exception*) + 27
    frame #4: 0x00007fff6414fbb9 libc++abi.dylib`__cxa_throw + 113
    frame #5: 0x00000001017d74a1 libmesh_dbg.0.dylib`libMesh::PetscVector<double>::_get_array(this=0x0000000108e09650, read_only=true) const at petsc_vector.C:1348
    frame #6: 0x00000001017dd322 libmesh_dbg.0.dylib`libMesh::PetscVector<double>::operator(this=0x0000000108e09650, i=4)(unsigned int) const at petsc_vector.h:1069
    frame #7: 0x00000001002ea229 libmesh_dbg.0.dylib`libMesh::DofMap::enforce_constraints_exactly(this=0x0000000108e09040, system=0x0000000108e08d60, v=0x0000000108e09650, homogeneous=false) const at dof_map_constraints.C:2153
    frame #8: 0x0000000101d6909a libmesh_dbg.0.dylib`libMesh::System::project_vector(this=0x0000000108e08d60, old_v=0x0000000108e25960, new_v=0x0000000108e09650, is_adjoint=-1) const at system_projection.C:407
    frame #9: 0x0000000101d65b2f libmesh_dbg.0.dylib`libMesh::System::project_vector(this=0x0000000108e08d60, vector=0x0000000108e09650, is_adjoint=-1) const at system_projection.C:220
    frame #10: 0x0000000101c9e917 libmesh_dbg.0.dylib`libMesh::System::restrict_vectors(this=0x0000000108e08d60) at system.C:334
    frame #11: 0x0000000101c9f679 libmesh_dbg.0.dylib`libMesh::System::prolong_vectors(this=0x0000000108e08d60) at system.C:384
    frame #12: 0x0000000101be5ca0 libmesh_dbg.0.dylib`libMesh::EquationSystems::reinit_solutions(this=0x00007ffeefbfe738) at equation_systems.C:211
    frame #13: 0x0000000101be3919 libmesh_dbg.0.dylib`libMesh::EquationSystems::reinit(this=0x00007ffeefbfe738) at equation_systems.C:123
    frame #14: 0x0000000100006a2e tmp`main(argc=9, argv=0x00007ffeefbff840) at adaptivity_ex2.C:410
    frame #15: 0x00007fff66a1e085 libdyld.dylib`start + 1
```

jwpeterson:
@fdkong: After further investigation, I don't think there's any way we can actually add this assert, as it may not even be logically correct to do so...

In `DofMap::enforce_constraints_exactly()` we hit the following logic when running `adaptivity_ex2`:
```
  else if (v->type() == GHOSTED)
    {
      v_local = v;
      v_global = v;
    }
```
That is, the pointers are set equal: `v_local == v_global`. Then, at every iteration of the following loop:
```
for (const auto & pr : _dof_constraints)
  {
    dof_id_type constrained_dof = pr.first;
    if (!this->local_index(constrained_dof))
      continue;

    const DofConstraintRow constraint_row = pr.second;

    Number exact_value = 0;
    if (!homogeneous)
      {
        DofConstraintValueMap::const_iterator rhsit =
          _primal_constraint_values.find(constrained_dof);
        if (rhsit != _primal_constraint_values.end())
          exact_value = rhsit->second;
      }
    for (const auto & j : constraint_row)
      exact_value += j.second * (*v_local)(j.first);

    v_global->set(constrained_dof, exact_value);
  }
```
we are calling `v_global->set()` which un-closes the vector (see, e.g. PetscVector::set() on line 150 of petsc_vector.C). In the next iteration of the loop, we access into the same vector with `(*v_local)(j.first)` which is what trips the assert.

roystgnr:
Yeah; in theory that code is correct even in parallel (we already expanded constraints so we're never writing to the same values we're reading from).  In practice though I'm now scared; is this is approved-by-PETSc behavior or is this just behavior that we get away with now but that can fail on us with any version upgrade?

roystgnr:
IIRC there was a similar issue with threaded container access in C++ standards?  The current standard says (with some exceptions like ```vector<bool>```) that it's safe to simultaneously do writes and reads to different entries in the same container, but in either previous standard drafts or previous threaded container libraries that was undefined behavior?

jwpeterson:
> is this is approved-by-PETSc behavior or is this just behavior that we get away with now but that can fail on us with any version upgrade?

Not sure, might be a good question for @jedbrown or the PETSc dev mailing list.

fdkong:
> Yeah; in theory that code is correct even in parallel (we already expanded constraints so we're never writing to the same values we're reading from). In practice though I'm now scared; is this is approved-by-PETSc behavior or is this just behavior that we get away with now but that can fail on us with any version upgrade?

If I wrote the code, I would not do in this way. I prefer to have a copy of `Vec`, and then read data from the copy and update data to the original `Vec`. 

There is no hurt to ask @jedbrown's input:

Hi Jed,

We have a code pattern similarly  as follows in `moose/libmesh`:

```
VecGetArray(vec, &array)
for (i=0; i<n; i++)
{
  the_old_value = array[i];

  the_new_value = modify  the_old_value

  VecSetValue(vec, a_location_different_from_i, & the_new_value,  INSERT_VALUES or ADD_VALUES
);
}
VecRestoreArray(vec, &array);

 VecAssemblyBegin(vec);
 VecAssemblyEnd(vec);
```

Is it a good practice to use` PETSc Vec` in this way? Does `PETSc` support this usage?  Is it safe to do so?

fdkong:
And also interested in what is going to happen if we reorder the last three lines as???:

```
 VecAssemblyBegin(vec);
 VecAssemblyEnd(vec);
VecRestoreArray(vec, &array);

```

jedbrown:
I don't think these are broken, but I wouldn't recommend that usage. Is the different location in the owned or off-process parts? The latter usage could result in extra bus traffic if using device vectors, or we could make it an error in a future version of PETSc.
Cc: @BarrySmith

fdkong:
> I don't think these are broken, but I wouldn't recommend that usage. Is the different location in the owned or off-process parts? 

The different location could be arbitrary (either owned or  off-process parts ).

>The latter usage could result in extra bus traffic if using device vectors, or we could make it an error in a future version of PETSc.
> Cc: @BarrySmith

roystgnr:
In libMesh, at least in the code John found here (and honestly in every other current use case I can remember), the locations being read from are a mix of owned and ghosted coefficients, but the locations being written to are all owned.

However, I'm *right now* in the middle of writing code that would (under some conditions) write to ghosted coefficients and not just local ones.  If that might throw an error in the future then I'll see about using a different temporary data structure rather than directly writing into PetscVector.

BarrySmith:
With the PETSc native vector type a VecSetValues() to an on process location takes place immediately while to a off process location it doesn't take place until the VecAssemblyBegin/End occurs. 

Thus completely generic code that reads and writes to the same locations in the vector can produce different results depending on how many processes it runs on it.

If you know that the code only writes to entries distinct from the entries it reads from or only writes to local entries then the code is safe and will produce the same results regardless of the number of processes

Currently there is no code that prevents calling VecSetValues() when a VecGetArray[Read]() is active. Logically it makes sense that VecSetValues() be an error if VecGetArray[Read]() is outstanding. Currently we do not even keep track of if a VecGetArray[Read]() is outstanding so that would need to be added. Of course if VecSetValues() is blocked the user needs either a copy of the vector or they must manage the temporary storage of the changed values and not set them until the VecRestoreArray[Read]() is called. 

If designing from scratch today I would likely block the VecSetValues (when a VecGetArray[Read]() is outstanding but I'm not sure if we should change the current code. 

Note: one could be more clever and only block the VecSetValues() for remote processes inside a VecGetArray[Read]() to also insure safe code for the native vector type but this may be too much cleverness.

jedbrown:
What happens when a write occurs to a location that hasn't been visited yet? If the location is off-process, it won't be observed but if it's on-process it will be seen as updated. Do you not care about that inconsistency? (What sort of algorithm is this?)

roystgnr:
> If you know that the code only writes to entries distinct from the entries it reads from ... then the code is safe

This would be the case for my new code; thank you for the clarification.

> Logically it makes sense that VecSetValues() be an error if VecGetArrayRead is outstanding.

Would that still be the case if the user knows they're only writing to entries they're not reading (until after the next VecAssemblyBegin/End)?  I agree that this strikes me as dangerous ground for users to tread on but in some cases it may be the only alternative to creating an unnecessary temporary vector.

> If designing from scratch today I would likely block the VecSetValues (when a VecGetArrayRead is outstanding but I'm not sure if we should change the current code.

Looking at our interface, we'd actually be unaffected - every libMesh function calling VecSetValues first checks to see if we have an outstanding array and then does VecRestoreArray or VecRestoreArrayRead as appropriate... that sounds like it could utterly *wreck* my new code's efficiency; maybe I should bite the bullet and use a temporary vector after all?

roystgnr:
> What sort of algorithm is this?

Projection code.  In intermediate phases for some element types it sets DoFs on element edges based on values previously set at the neighboring vertices, sets DoFs on faces based on the values previously set on edges, and sets DoFs on interiors based on values previously set on sides.  Right now we do that element-by-element with some (ok, a lot of) redundant computation, but I was hoping to change it to do separate sweeps.

jedbrown:
"Local vectors" (including the halo) are nice for this. Just get the array once and do all updates with direct array indexing. (I'm inferring that you're using `VecSetValues` instead of direct indexing because you want to think about updates in terms of global indices.)

PR #1928, opened 2018-11-01
Fix issue with constness in LaspackMatrix::add().

jwpeterson:
Refs #1916.

### DIFF
--- a/src/numerics/laspack_matrix.C
+++ b/src/numerics/laspack_matrix.C
@@ -384,8 +384,14 @@ void LaspackMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
   libmesh_assert_equal_to (this->m(), X_in.m());
   libmesh_assert_equal_to (this->n(), X_in.n());
 
-  const LaspackMatrix<T> * X =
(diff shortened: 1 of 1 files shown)


roystgnr:
I swear I hadn't forgotten about this.  But testing the fix uncovered a few other Laspack related bugs; the PR with all of those is coming shortly.

jwpeterson:
OK, thanks! Are you talking about runtime bugs? AFAICT everything compiled correctly with my changes.

roystgnr:
Yup.  Even the unit tests had a swath of errors.  I've figured out the *source* of the errors, but I haven't (unless my just-completed changes pass) *solved* all the errors yet; in my branch the unit tests currently pass but I get a failure as soon as adaptivity_ex2.

roystgnr:
Huh, that last change actually did pass.  Let me run through the rest of the examples before I post the PR, though; even if they don't all pass yet I want to be confident I haven't made anything *worse*.  My prior attempted fix had the effect of "some Laspack uses which previously diverged now converge, but others which previously converged (perhaps too slowly) now segfault", and this did not seem like a worthwhile tradeoff.

Issue #1925, opened 2018-10-30
inconsistent configure of curl


BalticPinguin:
Dear all,
I newly encountered a problem related to the curl-library: The main issue is that during linking of my program the '-lcurl'-option is set, while libmesh is configured without curl and thus the linker does not know where the library is.
As far as I understand it, it is introduced by contrib/netcdf, which configures with curl. 
Unfortunately, I don't understand, where the issue comes from all of a sudden; I used this repo with the same options for quite a while.

A second aspect is that I use the '--enable-everything'-option, so I would expect that curl is enabled? When '--enable-curl', it is added without further changes in the configure-output (i.e. there seems to be no option that is disabled instead).

The good thing is: with '--enable-curl' I have a consistent library again, but probably someone has an idea what actually happened and we can fix it?
Best,
Hubert

BalticPinguin:
Ok; one reason I see this now and didn't see before might be that some but not all of our nodes have curl; but the question of consistency should not be affected by this...

jwpeterson:
> A second aspect is that I use the '--enable-everything'-option, so I would expect that curl is enabled? 

No, if you look at our curl.m4 file, you'll see that `enablecurl=no` is the default. We ended up disabling it by default due to issues we were having in linking to curl on various systems. I think what you are seeing is probably related, but in this case it is NetCDF's fault and not ours that curl is attempting to be linked in. After looking through the `contrib/netcdf/4.4.1.1/configure.ac` file, I see:
```
# Curl support is required if and only if any of these flags are set:
# 1. --enable-dap

if test "x$enable_dap" = "xyes" ; then
require_curl=yes
else
require_curl=no
fi
```
No idea what "dap" is, but perhaps if we pass `--disable-dap` to NetCDF's configure we could avoid the issue with curl. It looks like we use the environment variable `$libmesh_subpackage_arguments` to control what flags are passed to NetCDF configure (see m4/netcdf.m4). You could try setting this variable to `--disable-dap` in your environment and seeing if that fixes the issue for you...

roystgnr:
Fixed by #1930

PR #1930, opened 2018-11-06
fix for #1925 

BalticPinguin:
Dear all,
What do you think of this patch as a fix for #1925 ?
In my viewpoint, this is a bit overspecific, but according to my tests it does its job.
Best regards,
Hubert


roystgnr:
Looked good to me, if it fixes the issue for you.  Thanks!

Issue #1927, opened 2018-11-01
UnstructuredMesh::contract() Clearing Point Locators


friedmud:
I'm running into an issue where `UnstructuredMesh::contract()` is clearing the PointLocator from the Mesh... but then it's needed by `DefaultCoupling` during a threaded loop - which causes an error in debug:

![screen shot 2018-10-30 at 11 05 36 am](https://user-images.githubusercontent.com/870705/47863582-360a4400-ddbd-11e8-91a1-535b5194b2b0.png)

I've tracked it down to this `contract()` call:

![screen shot 2018-10-31 at 12 19 06 pm](https://user-images.githubusercontent.com/870705/47863567-2ab71880-ddbd-11e8-95ed-ed4aae776acf.png)

The issue is that there is no way for me to fix this.  It all happens within the same call to `EquationSystems::reinit()`.  If I add a `mesh.sub_point_locator()` call to `DofMap::build_sparsity()` just before the threaded call... then it works fine.

One thing we could do is actually do a doubly-checked lock (with memory fencing) in `MeshBase::sub_point_locator()` so that it is actually thread safe... then we could remove the restriction.

roystgnr:
> If I add a mesh.sub_point_locator() call to DofMap::build_sparsity() just before the threaded call... then it works fine.

Hmm... that's our "standard" way to handle this, but I don't think it's appropriate here.  DefaultCoupling doesn't even need a PointLocator unless there are periodic BCs in play, and I'd hate to construct one unnecessarily!

> One thing we could do is actually do a doubly-checked lock (with memory fencing) in MeshBase::sub_point_locator() so that it is actually thread safe... then we could remove the restriction.

Like what we do in PetscVector::_get_array?  That sounds great to me!  Would you be up to writing this instance too?  Or better yet, encapsulate the idiom so it's more accessible to mortals?  The details of C++11 memory fence APIs aren't *too* awful, but boy the failure mode is.  Runtime errors that might only happen one time in a million occupy the lowest circle in runtime error hell.

friedmud:
Yeah - I can write this one up.  I'm not sure about abstracting it though - but I'll think about it while I do it.

It will be easier now for sure though: because now that we officially support C++11 we can just write it in pure C++11 and avoid all the C++03 compatible hacks I had to do back in the day for `PetscVector::_get_array()`

friedmud:
Well crap.  I put in a doubly-checked lock... but ran into the next problem:

[...]

PR #1932, opened 2018-11-09
Allow grabbing a sub_point_locator within a threaded section

friedmud:
This might look kinda ugly - but it works.  Looking for feedback!

refs #1927

### DIFF
--- a/src/mesh/mesh_base.C
+++ b/src/mesh/mesh_base.C
@@ -54,7 +54,6 @@ MeshBase::MeshBase (const Parallel::Communicator & comm_in,
   boundary_info  (new BoundaryInfo(*this)),
   _n_parts       (1),
   _is_prepared   (false),
-  _point_locator (),
   _count_lower_dim_elems_in_point_locator(true),
   _partitioner   (),
 #ifdef LIBMESH_ENABLE_UNIQUE_ID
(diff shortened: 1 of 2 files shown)


src/mesh/mesh_base.C
```diff
@@ -478,45 +476,70 @@ unsigned int MeshBase::recalculate_n_partitions()
...
-  libmesh_deprecated();
+  auto point_locator_ptr = _point_locator->load(std::memory_order_acquire);
 
-  if (_point_locator.get() == nullptr)
+  // If there's no master point locator, then we need one.
+  if (point_locator_ptr == nullptr)
     {
-      // PointLocator construction may not be safe within threads
-      libmesh_assert(!Threads::in_threads);
+      Threads::spin_mutex::scoped_lock lock(Threads::spin_mtx);
```
roystgnr: Is it safe to use Threads::spin_mtx here rather than _point_locator_mutex?  What if this ends up getting called from within other code holding the same global mutex?

src/mesh/mesh_base.C
```diff
@@ -478,45 +476,70 @@ unsigned int MeshBase::recalculate_n_partitions()
...
+
+      if (point_locator_ptr == nullptr)
+        {
+          // And it may require parallel communication
+          parallel_object_only();
```
roystgnr: If we do things this way we pretty much have to switch in libmesh.C from MPI_THREAD_FUNNELED to MPI_THREAD_SERIALIZED or MPI_THREAD_MULTIPLE as per the standard.  There's no good way for us to guarantee that the thread which ends up building the point locator is the same as the thread which called MPI_Init_thread.

roystgnr: @friedmud, based on the above and on the discussion last week, should we close this PR?  If you're compiling some MPI implementation(s) without threads to improve performance then there's no way we can safely risk making MPI calls from non-main threads.

PR #1933, opened 2018-11-12
infinite improvements

BalticPinguin:
Dear all,
I'd like to suggest some changes related to infinite elements:
There are three main issues I try to address in this PR:

- [b86d346, 45478c0,  8b8556f ]: added some documentation and assertions where infinite elements cannot be assumed (I just ran into these cases and think it might be nice to inform the user) 

- [dac368c ]: Change in the infFE:inverse_map(): I realized that the implemented newton iteration fails for v=0 (2* the radius of the FEM-region).
  So I removed the iterative scheme completely, as it anyway was a fallback for the analytical case only, which I think might be not needed (the analytic expression is quite stable, I guess).
  Also, I removed the search within the base-element as it just resembled the inverse_map of the respective 2D-elements; now this function is called directly.

- [commit  d70c659 ] Finally, I think it is convenient to have both the frequency and the phase available when dealing with infinite elements; independent of the use of complex numbers; so I removed the respective clauses.

Best, 
Hubert

### DIFF
--- a/include/fe/fe.h
+++ b/include/fe/fe.h
@@ -837,6 +837,7 @@ public:
   /**
    * Explicitly call base class method.  This prevents some
    * compilers being confused by partially overriding this virtual function.
+   * \note: pts need to be in reference space coordinates, not physical ones.
    */
   virtual void reinit (const Elem * elem,
                        const std::vector<Point> * const pts = nullptr,
--- a/include/fe/fe_compute_data.h
+++ b/include/fe/fe_compute_data.h
@@ -79,21 +79,17 @@ public:
   std::vector<Number> shape;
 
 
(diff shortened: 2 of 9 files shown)


roystgnr:
The documentation improvements are good.  The inverse_map() refactoring looks good, though I wouldn't swear to it.  The d70c659 likewise.

We need more test coverage - there's currently no Civet recipe for turning on --enable-ifem --enable-complex at the same time.  I assume you've done so by hand but we really ought to make ifem on by default in the "Test enable complex" build.

jwpeterson:
> we really ought to make ifem on by default in the "Test enable complex" build.

Good idea, I'll put in a CIVET recipe update PR for that now...

roystgnr:
Good enough for me.  Thanks everyone!

BalticPinguin:
Great; thanks a lot.

jwpeterson:
Also, I forgot to say :+1: for best issue title ever.

PR #1935, opened 2018-11-13
Update MetaPhysicL

lindsayad:
- Support for std::complex functions for some classes
- Bugfix in DynamicSparseNumber*::operator[]
- More unit tests


roystgnr:
Let's wait until @pbauman gets femputer back up before merging this one.

jwpeterson:
Let me all double-check that running bootstrap in libmesh on this branch doesn't modify and autoconf files...

jwpeterson:
> Let me also double-check

Looks good to me!

roystgnr:
@pbauman seems to be AFK right now, and the only thing I'm worried about triggering on femputer is the -Werror stuff, so we'll risk a merge.

Issue #1883, opened 2018-10-05
Use query_node_ptr() in BoundaryInfo::add_node(dof_id_type, boundary_id_type)


jwpeterson:
As it stands now, if the user hands us a bad node id, the code simply segfaults in opt mode. In dbg mode, the error message is a bit cryptic, i.e. something like:
```
Assertion `i < this->n_nodes()' failed.
i = 534
this->n_nodes() = 534
```
whereas if we query the node first, we can put a more meaningful description of the problem in a `libmesh_error_msg()`.

cc: @arovinelli

jwpeterson:
Fixed by #1937.

PR #1937, opened 2018-11-13
Warn if user tries to add a boundary id for a non-existent node.

jwpeterson:
We could also make this a fatal error if people think that would be
safer... previously the code would just segfault in opt mode and
assert in dbg mode.

Refs #1883.

### DIFF
--- a/src/mesh/boundary_info.C
+++ b/src/mesh/boundary_info.C
@@ -579,10 +579,17 @@ void BoundaryInfo::add_elements(const std::set<boundary_id_type> & requested_bou
 
 
 
-void BoundaryInfo::add_node(const dof_id_type node,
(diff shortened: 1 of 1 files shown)


roystgnr:
This is only going to be triggered by broken user code, right?  My personal preference would be to leave it as-is (especially if we're typically getting segfaults rather than corrupt data in the worst case), but it's not a performance-critical segment so we can hold users' hands in opt mode too.  A fatal error would probably be more helpful than a warning people will miss or ignore.

jwpeterson:
The segfault and corresponding assert were very confusing to @arovinelli, so I think we should change it *somehow*. I'm happy making it a fatal error instead. Note: performance-critical code can still call the version of `add_node()` that this one is forwarding to, which doesn't do any extra checks.

Issue #1919, opened 2018-10-23
Trouble with AMR/reinit with subdomain restricted variables in TransientExplicitSystem


ajamar1:
- I am using libmesh 1.3.0, but I think I have been fighting this issue unknowingly for years.
- The trouble is with the processing of old and older solution vectors in a TransientExplicitSystem when calling EquationSystems::reinit() after a call to MeshRefinement::refine_and_coarsen_elements() for parallel problems only.  
- Further, the issue is isolated to when I have subdomain restricted variables.   For example, subdomain 1 has variable 0, and subdomain 2 has variable 1. The two subdomains interface such that common nodes should have DoFs associated with both of the two variables.
- After a call to AMR/reinit, part of the processor boundary changes to be aligned at the interface of the two subdomains.
- When all is said and done, the data in the old and older solution vectors is corrupted (value set to zero) at the nodes on that specific part of the processor boundary.  The data in the current solution looks ok.

*This goes away if I run on 1 processor.
*This goes away if I put both variables in the entire domain
*I am seeing this on a TransientExplicitSystem, and I am not sure if it would happen on a TransientImplicitSystem.
*I am not convinced that its an issue associated with AMR.  Rather, I think it just took AMR to move the processor boundary around such that is aligned at the variable interface.  So I think its an issue embedded in all the reinit logic.

I have attached a stand-alone code, which manifests the bug, in the attached tarball.  The tarball includes a README file, Makefiles, an example problem with tecplot layouts to view the important output files, and an archived results folder to show you what I was getting on my system.

Inevitably, I am sure you will be asking for more details.  Roy knows how to reach out to me, and I am happy to discuss over the phone.  I hope I am doing something wrong and that its not a bug.

[libmesh_bug_report.tar.zip](https://github.com/libMesh/libmesh/files/2508151/libmesh_bug_report.tar.zip)

roystgnr:
Thank you very much; this is enough for me to reproduce the problem.

I'm afraid I'm struggling trying to figure it out, though!

> I am using libmesh 1.3.0, but I think I have been fighting this issue unknowingly for years.

I can replicate it with the current master/HEAD.

> I am not convinced that its an issue associated with AMR. Rather, I think it just took AMR to move the processor boundary around such that is aligned at the variable interface. So I think its an issue embedded in all the reinit logic.

This appears to be correct.

> I hope I am doing something wrong and that its not a bug.

[...]

PR #1938, opened 2018-11-15
GenericProjector rewrite

roystgnr:
I'm posting this at an early stage, both so it can get into CI (it's passing my local tests now but this is a huge change that could have all kinds of corner cases) and so @ajamar1 can try it out (it passes the test case in #1919 for me, at least), but anybody who *isn't* running the impacted transient-repartitioning-of-subdomain-restricted-variables case isn't going to want to use the current unbenchmarked bag of optimizations mixed with pessimizations in the first commits here.  This is going to need more optimization work before it's ready to merge.

The goals:

Make projections faster by removing redundant calculations

Make projections correct even in the corner cases where subdomain-restricted
variables are on DoFs owned by processors which don't own that
subdomain.

The means:

Sort projection operations by vertices, edges, faces, interiors,
avoiding duplication.

Test ghost nodes' patches to see what their owners can compute, and
push our computed data to the owners if necessary, in between the
steps where it's computed and the steps where it's used.

The current problems:

Looping over vertices/edges/nodes rather than over elements is forcing
more context data initialization.

This initial commit is missing some of the common case optimizations that
were in the previous version; those need to be re-added.

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -716,18 +716,40 @@ public:
 
   /**
    * Fills the vector \p di with the global degree of freedom indices
-   * for the node.
+   * for the \p node.
    */
   void dof_indices (const Node * const node,
                     std::vector<dof_id_type> & di) const;
 
   /**
    * Fills the vector \p di with the global degree of freedom indices
-   * for the node.   For one variable \p vn.
+   * for the \p node, for one variable \p vn.
    */
   void dof_indices (const Node * const node,
                     std::vector<dof_id_type> & di,
                     const unsigned int vn) const;
+
+  /**
+   * Appends to the vector \p di the global degree of freedom indices
+   * for \p elem.node_ref(n), for one variable \p vn.  On hanging
+   * nodes with both vertex and non-vertex DoFs, only those indices
+   * which are directly supported on \p elem are included.
+   */
+  void dof_indices (const Elem & elem,
+                    unsigned int n,
+                    std::vector<dof_id_type> & di,
+                    const unsigned int vn) const;
+
+  /**
+   * Appends to the vector \p di the old global degree of freedom
+   * indices for \p elem.node_ref(n), for one variable \p vn.  On
+   * hanging nodes with both vertex and non-vertex DoFs, only those
+   * indices which are directly supported on \p elem are included.
+   */
+  void old_dof_indices (const Elem & elem,
+                        unsigned int n,
+                        std::vector<dof_id_type> & di,
+                        const unsigned int vn) const;
 
   /**
    * Fills the vector \p di with the global degree of freedom indices
@@ -1410,6 +1432,16 @@ private:
                      std::size_t & tot_size
 #endif
                      ) const;
+
+  /**
+   * Helper function that implements the element-nodal versions of
+   * dof_indices and old_dof_indices
+   */
+  void _node_dof_indices (const Elem & elem,
+                          unsigned int n,
+                          const DofObject & obj,
+                          std::vector<dof_id_type> & di,
+                          const unsigned int vn) const;
 
   /**
    * Builds a sparsity pattern
--- a/include/systems/generic_projector.h
+++ b/include/systems/generic_projector.h
@@ -85,6 +85,7 @@ private:
   const ProjectionAction & master_action;
   const std::vector<unsigned int> & variables;
   std::unordered_map<dof_id_type, std::vector<dof_id_type>> * nodes_to_elem;
+  bool done_saving_ids;
 
 public:
   GenericProjector (const System & system_in,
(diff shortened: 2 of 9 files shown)


roystgnr:
Those test failures might just be because one of the optimizations (interpolating rather than projecting non-vertex LAGRANGE DoFs) can also give slightly different results.  I guess that's the optimization I'll add back in first.

roystgnr:
Now that I've started more extensive testing:

The good news: at least with complicated FE types this already seems to be ~25% faster than the previous code, so I'm now much more confident that even the simple cases will be faster after I restore those optimizations.

The bad news: this may pass most tests here, but there's still bugs.  I haven't found any incorrect results yet, but at least one or two tests I've run hang.

ajamar1:
Roy,

We are in the middle of a code release so it’ll likely be next week before I grab the patch and start testing for you.  Thanks for the update.  I hope this works out.

Adam

On Nov 16, 2018, at 6:46 PM, roystgnr <notifications@github.com<mailto:notifications@github.com>> wrote:


Now that I've started more extensive testing:

The good news: at least with complicated FE types this already seems to be ~25% faster than the previous code, so I'm now much more confident that even the simple cases will be faster after I restore those optimizations.

The bad news: this may pass most tests here, but there's still bugs. I haven't found any incorrect results yet, but at least one or two tests I've run hang.

milljm:
unrelated to this PR, @roystgnr would it help if I modified the civet recipe to not stop at the first failed target?

 https://civet.inl.gov/job/311338/

roystgnr:
Nah, in general an early exit is ideal.  *Sometimes* I find that I can't replicate a CI failure and then having subsequent failed cases to try is helpful, but that's very rare.

roystgnr:
Well that's interesting.  In that initial_adapt test it's not the solution variables that look weird, it's the "box" variable that got set by the BoxMarker.  In the gold standard it's 2 in the box and 1 everywhere else; in the broken output it's basically equal to ```elem->level()``` (2 in the box, 1 in the transitional layer, 0 everywhere else).

roystgnr:
```fatal error: petsc.h: No such file or directory```

Okay, I'm actually pretty sure this one *wasn't* my fault...

**Comment by roystgnr** [comment]
No, the 64bit and debug tests still can't find petsc.h.  Anyone have any ideas?

Oh, and no GRINS tests yet, but that's just the usual "femputer is down again".  @pbauman?

**Comment by roystgnr** [comment]
I may have finally exhausted the bugs here, so if the plain Test CI passes then I'll do one last pass at interactive rebase cleanup and start throwing parallel sweeps at this thing.

**Comment by roystgnr** [comment]
Nope, plain Test CI has the same petsc.h problem

**Comment by permcody** [comment]
We are actively working with CIVET. We didn't just invalidate because we are trying to understand what was going on. Summary: We've recently added a "clean machine" target to every single event so that the build machines will ensure they have enough space _before_ they run a job. Some of our targets require upwards of 40GBs of space starting from a fresh disk! This will give us a more reliable testing infrastructure moving forward. What this has exposed is obscure problems in our cloning/resetting logic. With nested submodules, it's actually pretty easy to confuse git when you "force delete" a parent submodule, which can cause bad things to happen downstream.

**Comment by permcody** [comment]
Actually hang on... I see your failure. This has nothing to do with nested submodules... It looks like configure properly detected and configured for PETSc. Oh goodie, another problem to look into.

**Comment by milljm** [comment]
I am not exactly sure how to fix this. Here is what is happening.

As civet launches, it reads a file which instructs it on which 'moose-environment' package to load. The file in question is called 'package_version'. Civet searches for this file in the following order:

```bash
if 
  $MOOSE_DIR/package_version
elif
  $REPO_DIR/moose/package_version
elif
  $REPO_DIR/package_version
else:
  use latest available version
fi
```

While libMesh does not have a package_version file sitting in the root of its repo, civet will use the latest available version. But later on in the civet recipe, we are doing something else. MOOSE is instructing civet to use a different set of compilers.

I need to devise a way, that once a recipe begins, it needs to 'stick' with that set of compilers, regardless what other repos might instruct civet to do.

roystgnr:
> I am not exactly sure how to fix this.

Ping me when you come up with a fix or a workaround?  I've got no shortage of other stuff I can be working on, but half of it will also hit the "needs MOOSE CI tests to make sure Roy's not being an idiot" stage sooner or later.

milljm:
A quick fix, is for you to add a file, instructing civet to use the following package version:

```bash
echo "repo-hash:56178942a6f4cf47dfc6bb8b678d26bb6c5a32ed" > package_version
```

This will allow you to work on this PR today...  While we try and solve this for tomorrow.

Note: I have a PR into moose which will update it's package_version file... So... unfortunately, once that goes in later tonight, this PR will start failing _again_... requiring you to update that package_version file to match what moose/package_version contains.

milljm:
@roystgnr hold off on adding package_version to your repo (if you were heading that direction)... we are close to figuring out a way to do this without a change on your end.

roystgnr:
This was passing the GRINS tests even before those last bugfixes, so I'll take that as gospel even if femputer isn't back up soon.

I'm working on unit testing that will improve the libMesh test coverage (I added special code paths to optimize some common MOOSE use cases but then half those weren't properly covered by libMesh test or example cases) and that will in particular cover the original bug report that got me into this mess, and I'll run some sweeps in CI here while I do so, but once that's all done this ought to be finally ready to merge.

pbauman:
Sorry, rebooted femputer, it's running now.

roystgnr:
Gah.  Both those segfaults are in the --enable-parmesh unit tests, but on my own machines I can sweep those from 1 proc through 24 procs, in debug mode even, with no problems.

roystgnr:
Huh.  That debug check is catching a nullptr which should be a remote_elem pointer on a first-order clone of a DistributedMesh, which should be straightforward enough... except:

* both DistributedMesh::clone() and UnstructuredMesh::all_first_order() have code specifically setting all remote_elem pointers
* there's no way a GenericProjector change should have even triggered a preexisting bug here.  Maybe it was just a preexisting bug that we never managed to check for before?  The failing assertion is a DEBUG-only test on data structures in a code path that might not trigger segfaults if skipped, and we've never run that automated distributed sweep with debug mode on before now, and for some reason I can only replicate the bug on my own machines when using some custom flags.

roystgnr:
Ah, I think I see now.  This looks like a bug triggered by #2028.  When we use the skip_find_neighbors code path we might be ending up accidentally *overwriting* remote_elem neighbor links with nulls because we never put remote_elem in the lookup table we use to copy neighbor links over.

Pinging @lindsayad to warn him - this looks like my bug, and I'll fix it ASAP, but if he's using DistributedMesh copy construction or cloning extensively enough to want to improve those copy constructors, there's a chance that his use cases are more direly affected by this bug than mine are.

roystgnr:
I'd love to get more eyes on this first (@jwpeterson, @bboutkov, @ajamar1 ?), but at this point I think it's ready to merge assuming CI is still happy with the last swath of new cppunit assertions.  Most of the recent failures in the new test coverage turned out to be bugs in master too, and at least in my own runs this PR is now passing all the new testing with flying colors (which is more than I can say for master, even after the first rounds of fixes, because the third fix would be harder to backport).

roystgnr:
Ok, now I've got test coverage that seems to fail reliably on master *in the same way* as #1919 did and that works reliably on this PR for me, so hopefully we can avoid any future regressions in the bug that started this whole mess!  If CI is still happy too then I'll merge.

bboutkov:
So I gave these changes (without the last 5 commits) a whirl yesterday, and the overall results are looking good. First of all, in terms of correctness,  I was not able to produce any failures after testing various configurations targeting GMG + extensive use of  ``projection_matrix()``, so looks good to me on that end. Second, I really like the the logging improvements for projection related calls, it looks so much cleaner now, so thanks for those changes!

I also took a look at overall weak scaling timings for GMG on Poisson and how I was affected by this PR. For overall timing, these changes shave off ~0.5s off each run up to np256, so that seems to be in line with the cleanup. Since I didn't see too much overall change I also looked at the timing of the GenericProjector itself and noticed a bit of a curiosity (see plot below). While the speedup was fairly consistent up to np64, afterwards it wasn't as noticeable. I wasn't quite able to track down any particular reason for this, partly because these timings are subject to about as much variability from run to run as the difference between the pre vs post changes,  but after a few tries I was able to consistently replicate this difference at np256. Perhaps its OK, since after all these timings are less than a couple of seconds for 55k local elem for the big run, but I figured I would make the note since I wasn't able to get any higher proc count runs in time to ensure this issue doesn't get even worse at higher proc counts.

gen_proj_proj

roystgnr:
Thanks for that!

The last five commits are all new tests, new features for use in new tests, or fixes for new tests, so shouldn't affect any of your results.

What's nasty here is that I'm afraid we *expect* some loss of scalability, because the fix for #1919 necessarily entailed communication steps that didn't exist in the original algorithm; better to get the right answer slowly than to get the wrong answer quickly.  I was hoping for the crossover point to be in the tens of thousands of processors, though, not in the hundreds.

We might be able to mitigate the problem slightly (and improve some other areas) by enabling more overlap of communication and calculation.  #1965 was actually a tiny regression in that respect (though *utterly* worth it for the asymptotic speedup in the common case), so we should have some room to improve, and I've been thinking about how.

Oh, and wait a minute - those timings of yours are for the GMG projection matrix constructions, right?  I'm afraid one of the sends for that case *doesn't* use the new algorithm from #1965, because that currently is only available when sending vectors of fixed-length types, whereas the projection matrix rows have to get sent as vectors-of-vectors, which falls back on my intermediate alltoall-limited algorithm, which means we expect a serious loss of scalability.  So the good news from most users' perspective is that you're kind of the worst-case scenario here, but the bad news from your perspective is you're kind of the worst-case scenario here!  I apologize!  If you have any plans to do GMG on thousands of cores in the near future, let me know, and I'll bump that optimization up higher on my todo list.

PR #1941, opened 2018-11-15
Default processor_id_type to 4 bytes

### LINKED ISSUE #1939: Default processor_id_type to 4 bytes
We have users who need this, and it's really hard to figure out that you've misconfigured something when your 70K core opt-mode run crashes....

And we aren't even getting anything out of it!  DofObject currently sticks _processor_id in between _id and _idx_buf and so there's going to be sizeof(size_t)-2 bytes of padding after every _processor_id anyways!

roystgnr:
Our configure script was already lying and *saying* it defaulted to 4
bytes; now let's *actually* default to 4 bytes.

This fixes #1939

### DIFF
--- a/configure
+++ b/configure
@@ -31375,7 +31375,7 @@ $as_echo "configuring size of dof_id... $dof_bytes" >&6; }
 if test "${with_processor_id_bytes+set}" = set; then :
   withval=$with_processor_id_bytes; processor_bytes="$withval"
 else
-  processor_bytes=2
+  processor_bytes=4
 fi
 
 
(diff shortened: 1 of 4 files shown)


roystgnr:
I may move that last commit to a separate PR; it fixes a major bug that was triggered by the 2->4 byte default processor_id_type change but it's urgent to apply whether we can do the type change or not.

roystgnr:
Hmmm... I might be able to get CheckPointIO to autodetect old broken-format checkpoint files and workaround the flaw.  That would only fix non-broken files in the broken format, but that would be enough for all the tiny files in MOOSE tests.

I have no idea what's going on with the weighted partitioner failures, though.

roystgnr:
(Especially considering the weighted partitioner *non*-failures in Test debug!)

PR #1944, opened 2018-11-16
Mark new side_ptr functions override to avoid warnings.

jwpeterson:
clang seems to be more strict on warning about this than GCC...

### DIFF
--- a/include/geom/cell_hex.h
+++ b/include/geom/cell_hex.h
@@ -144,7 +144,7 @@ public:
   /**
    * Rebuilds a primitive (4-noded) quad for face i.
    */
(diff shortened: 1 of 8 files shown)


roystgnr:
Sorry I forgot that; thanks!

clang has a warning option that lets you flag any overridden virtual method not marked ```override```?

jwpeterson:
> clang has a warning option that lets you flag any overridden virtual method not marked `override`?

One that's on by default anyway... it doesn't seem to get triggered unless you've marked at least one other virtual function in a class as override.

jwpeterson:
The 64-bit test failure was due to a MOOSE package update and shouldn't have anything to do with this PR.

PR #1948, opened 2018-11-21
eval_old_dofs needs var, not var_component

roystgnr:
Hopefully this fixes the newer bug found in
https://github.com/idaholab/moose/issues/12309


pbauman:
Job [GRINS-devel](https://femputer.eng.buffalo.edu:8000/job/2562/) on cdbb482 : invalidated by @roystgnr

Kicking this test; it doesn&#39;t look like it&#39;s using the changed code so may be a fluke?

Issue #850, opened 2016-03-04
Build fails with --disable-amr


bwspenc:
There are issues with systems/fem_context.C, maybe other files too.

jwpeterson:
This should be fixed as of 7ec01f7.

andrsd:
This seems to be an issue again. Here is what we did to fix it. Feel free to apply or use a better fix.

```
diff --git a/src/base/dof_map.C b/src/base/dof_map.C
index ea89e7e4b..0ec7624ac 100644
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -1855,13 +1855,14 @@ void DofMap::extract_local_vector (const NumericVector<Number> & Ug,
                                    const std::vector<dof_id_type> & dof_indices_in,
                                    DenseVectorBase<Number> & Ue) const
 {
+  const unsigned int n_original_dofs = dof_indices_in.size();
+
 #ifdef LIBMESH_ENABLE_AMR
 
   // Trivial mapping
   libmesh_assert_equal_to (dof_indices_in.size(), Ue.size());
   bool has_constrained_dofs = false;
 
-  const unsigned int n_original_dofs = dof_indices_in.size();
   for (unsigned int il=0; il != n_original_dofs; ++il)
     {
       const dof_id_type ig = dof_indices_in[il];
```

PR #1951, opened 2018-11-26
Fix --disable-amr build.

jwpeterson:
Refs #850.

### DIFF
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -1855,13 +1855,14 @@ void DofMap::extract_local_vector (const NumericVector<Number> & Ug,
                                    const std::vector<dof_id_type> & dof_indices_in,
                                    DenseVectorBase<Number> & Ue) const
 {
+  const unsigned int n_original_dofs = dof_indices_in.size();
+
 #ifdef LIBMESH_ENABLE_AMR
 
   // Trivial mapping
   libmesh_assert_equal_to (dof_indices_in.size(), Ue.size());
   bool has_constrained_dofs = false;
 
-  const unsigned int n_original_dofs = dof_indices_in.size();
   for (unsigned int il=0; il != n_original_dofs; ++il)
     {
       const dof_id_type ig = dof_indices_in[il];
--- a/src/mesh/xdr_io.C
+++ b/src/mesh/xdr_io.C
@@ -406,8 +406,9 @@ void XdrIO::write_serialized_connectivity (Xdr & io, const dof_id_type libmesh_d
       pack_element (xfer_conn, elem);
 #ifdef LIBMESH_ENABLE_AMR
       parent_id_map[elem->id()] = std::make_pair(this->processor_id(),
-                                                 my_next_elem++);
+                                                 my_next_elem);
 #endif
+      ++my_next_elem;
     }
   xfer_conn.push_back(my_next_elem); // toss in the number of elements transferred.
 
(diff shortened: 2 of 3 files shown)


jwpeterson:
Also attached the optional "Test No AMR" recipe...

jwpeterson:
Looks like a unit test is also failing with AMR disabled. I'll have to take a closer look, but it may be that this test should not be run in this configuration...
```
[0] ../src/base/dof_map_constraints.C, line 3389, compiled Nov 26 2018 at 08:17:07
[0]PETSC ERROR: ------------------------------------------------------------------------
[0]PETSC ERROR: Caught signal number 11 SEGV: Segmentation Violation, probably memory access out of range
```

roystgnr:
Serial unit tests pass for me now but I'm still getting a failure in parallel.

roystgnr:
And none of our CI is running the libMesh tests in parallel.  @pbauman, @jwpeterson, could we change the GRINS-devel and "Test debug" recipes to use, say, LIBMESH_RUN='mpiexec -n 8'?  The speedup on bigger examples will probably outweigh the extra overhead on tiny cases.

roystgnr:
Or maybe we use a different processor count on each; 8 on one and 16 on another?  I know John just added the sweep recipes but I don't want to use those often and it'd be good to get a *little* less redundancy in the different tests.

jwpeterson:
> could we change the GRINS-devel and "Test debug" recipes to use, say, LIBMESH_RUN='mpiexec -n 8'?

You talking specifically about unit tests now, or all of the tests that get run by "make check"?

roystgnr:
All the "make check" tests.  The unit tests are all tiny enough that I'm sure running in parallel is just going to make them slower, and they don't give as much corner case coverage as the whole swath of adaptivity and adjoints examples do.

jwpeterson:
OK, so a single parallel (-n 8) "make check" step added to the "Test debug" recipe? Should be easy enough...

roystgnr:
I'd just replace the existing step.  It's much rarer for a bug to affect serial but not parallel runs than for it to break the other way around.

jwpeterson:
:+1: I can't formally approve this PR since I opened it. Thanks for finding the remaining issues @roystgnr!

jwpeterson:
I just looked at the Test_debug recipe and it already does this.
```
[Check]
script = scripts/build.sh
abort_on_failure = True
allowed_to_fail = False
APP_SUBDIR = build
MAKE_ARGS = check
LIBMESH_RUN = mpiexec -np 4
```
Looks like the change was made in idaholab/civet_recipes@09aeaab37bd070d77dbcaecb18ba46d7140e53b3 (Oct 2017).

jwpeterson:
If you drill down into the CIVET output for the Test_debug recipe, you should see e.g.:
```
* Done Running Example introduction_ex1:
*  mpiexec -np 4 ./example-dbg -d 3 output.xdr 
```

jwpeterson:
Anyway, I think this can be merged whenever you are ready.

roystgnr:
Aha!  We're passing Test debug not because it's running in serial, but because it doesn't have any unit tests being run!  No good way to fix that unless you happen to have a mixing-glibcxx_debug-settings-compatible cppunit install on your build boxes, and that would be pretty fragile.  Or unless we turn off the glibcxx debugging, but I think that would lose more coverage than it added.

roystgnr:
Anyway, the parallel fix worked for me.  Thanks for getting this started!

PR #1953, opened 2018-11-26
Prefer vector::data() to &v[0].

jwpeterson:
Since vector::data() is technically a C++11 standard, also add a configure test to ensure that compilers support this API.

### DIFF
--- a/src/numerics/trilinos_epetra_matrix.C
+++ b/src/numerics/trilinos_epetra_matrix.C
@@ -97,14 +97,14 @@ void EpetraMatrix<T>::update_sparsity_pattern (const SparsityPattern::Graph & sp
   if (m==0)
     return;
 
-  _graph = new Epetra_CrsGraph(Copy, *_map, &n_nz_tot[0]);
+  _graph = new Epetra_CrsGraph(Copy, *_map, n_nz_tot.data());
 
   // Tell the matrix about its structure.  Initialize it
   // to zero.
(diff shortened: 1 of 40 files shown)


src/numerics/trilinos_epetra_matrix.C
```diff
@@ -97,14 +97,14 @@ void EpetraMatrix<T>::update_sparsity_pattern (const SparsityPattern::Graph & sp
...
-  _graph = new Epetra_CrsGraph(Copy, *_map, &n_nz_tot[0]);
+  _graph = new Epetra_CrsGraph(Copy, *_map, n_nz_tot.data());
 
   // Tell the matrix about its structure.  Initialize it
   // to zero.
   for (numeric_index_type i=0; i<n_rows; i++)
     _graph->InsertGlobalIndices(_graph->GRID(i),
                                 cast_int<numeric_index_type>(sparsity_pattern[i].size()),
-                                const_cast<int *>((const int *)&sparsity_pattern[i][0]));
+                                const_cast<int *>(sparsity_pattern[i].data()));
```
roystgnr: This broke compilation with Trilinos.  const_cast won't handle the unsigned->signed conversion.

jwpeterson: Oops, there was a hidden `reinterpret_cast` cast happening there, eh. I can fix that.

PR #1954, opened 2018-11-27
Fix failure to sync nodes in some AMR cases

### LINKED ISSUE #1952: Infinite loop in MeshCommunication::make_node_proc_ids_parallel_consistent
I'm not sure how this never got triggered before - to trigger a failure all I need to do is run with --enable-parmesh on 6 processors, and the failing test got added to adaptivity_ex3 two years ago.

@brianmoose, could we add a CIVET test that does a distributed mesh processor count sweep over libMesh "make check"?  CI has been doing that for the MOOSE test suite and I've been doing it with the GRINS test suite but clearly that's not enough... probably not enough 1D tests in either suite, and this corner case is much harder to trigger in 2D or 3D.

Anyway: make_node_proc_ids_parallel_consistent currently uses a "pull" algorithm: every processor looks at all their ghosted elements, then asks those elements' owners for the updated pid of every node on the element.  We have to loop, because in some literal corner cases proc A has to find out a pid from proc B only after B has found it out from proc C who had to ask proc D, but eventually the loop quits.

Except with enough adaptivity involved.  In the failing case I have here, proc 3 owns the leftmost node of a 1D domain, and every other processor needs to know about that node because it's on the single element in the coarse domain.  Proc 5 manages to make that query, because it has a ghosted coarse element from proc 3, but proc 3's subpartition is remote on every other processor, and since proc 5 doesn't own any elements touching that node it doesn't get to propagate the information further.

I think the way to fix this will be to expand the queries that get sent, to include non-local nodes on local elements.

roystgnr:
This should resolve #1952

### DIFF
--- a/include/parallel/parallel_ghost_sync.h
+++ b/include/parallel/parallel_ghost_sync.h
@@ -577,11 +577,25 @@ bool sync_node_data_by_element_id_once(MeshBase & mesh,
           if (!node_check(elem, n))
             continue;
 
-          if (i_have_elem &&
-              elem->node_ref(n).processor_id() == comm.rank())
+          const processor_id_type node_pid =
+            elem->node_ref(n).processor_id();
+
+          if (i_have_elem && (node_pid == comm.rank()))
             continue;
 
-          ghost_objects_from_proc[proc_id]++;
+          if (i_have_elem)
+            {
+              libmesh_assert_not_equal_to
+                (node_pid, DofObject::invalid_processor_id);
+              ghost_objects_from_proc[node_pid]++;
+            }
+          else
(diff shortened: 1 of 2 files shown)


roystgnr:
Assuming this passes the basic Civet tests I'm going to hammer it with our extended stuff too before merging.

roystgnr:
Ok, *definitely* going to hammer it before merging.  That hang was a real bug, and it wasn't a bug that was triggered by more than a single one of the default CI tests or by my own private tests.

roystgnr:
This took a lot of hammering and it fixes #1952 in my tests.

PR #1955, opened 2018-11-28
Remove yet more iterator boilerplate

jwpeterson:
Thankfully, I think these patches are getting smaller over time. Keeping this split up into separate commits since just yesterday @roystgnr found an issue with one of these refactorings in the --disable-amr config. I'll be attaching multiple CIVET recipes to this one... 😄

### DIFF
--- a/src/reduced_basis/rb_scm_evaluation.C
+++ b/src/reduced_basis/rb_scm_evaluation.C
@@ -417,17 +417,13 @@ void RBSCMEvaluation::legacy_write_offline_data_to_files(const std::string & dir
       Xdr C_J_out(file_name.str(), mode);
 
       for (std::size_t i=0; i<C_J.size(); i++)
-        {
-          RBParameters::const_iterator it     = C_J[i].begin();
-          RBParameters::const_iterator it_end = C_J[i].end();
-          for ( ; it != it_end; ++it)
-            {
-              // Need to make a copy of the value so that it's not const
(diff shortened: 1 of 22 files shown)


jwpeterson:
The Test debug failure is due to a recent CIVET configuration change, not a real issue. Will be invalidating shortly.

src/reduced_basis/rb_scm_evaluation.C
```diff
@@ -534,17 +530,13 @@ void RBSCMEvaluation::legacy_read_offline_data_from_files(const std::string & di
...
-        {
-          std::string param_name = it->first;
-          Real param_value;
-          C_J_in >> param_value;
-          C_J[i].set_value(param_name, param_value);
-        }
-    }
+    for (const auto & pr : get_parameters())
+      {
+        std::string param_name = pr.first;
```
roystgnr: Should this be another one we change to const reference while we're at it?

jwpeterson: Yep, good catch.

jwpeterson: Oh no, wait a minute. The problem here was that `C_J_in` in this code chunk does not have an `>>` overload for const references if I recall correctly?

jwpeterson: Ah, misread similarly named parameters. I will change that.

PR #1957, opened 2018-11-29
Fixed compiling with icc 19.

hrittich:
Hi,

a change is necessary for the ICC 19.

According to [this question on Stack Overflow](https://stackoverflow.com/questions/53527682/) the old version used an undefined behavior of the STL.

Feel free to remove the comment I made in the code.

Regards

### DIFF
--- a/src/systems/fem_context.C
+++ b/src/systems/fem_context.C
@@ -80,8 +80,20 @@ void FEMContext::init_internal_data(const System & sys)
 {
   // Reserve space for the FEAbstract and QBase objects for each
   // element dimension possibility (0,1,2,3)
-  _element_fe.resize(4);
-  _side_fe.resize(4);
+
+  // Below is a workaround for the ICC 19. The original code was:
+  //
+  // _element_fe.resize(4);
+  // _side_fe.resize(4);
+
+  _element_fe.clear();
+  for (int i=0; i<4; ++i)
+    _element_fe.push_back(std::map<FEType, std::unique_ptr<FEAbstract>>());
+
+  _side_fe.clear();
+  for (int i=0; i<4; ++i)
+    _side_fe.push_back(std::map<FEType, std::unique_ptr<FEAbstract>>());
+
   _element_fe_var.resize(4);
   _side_fe_var.resize(4);
 
--- a/configure
+++ b/configure
@@ -5575,6 +5575,10 @@ fi
 
                   GXX_VERSION_STRING="`($CXX -V 2>&1) | grep 'Version '`"
                   case "$GXX_VERSION_STRING" in #(
+  *19.*) :
+    { $as_echo "$as_me:${as_lineno-$LINENO}: result: <<< C++ compiler is Intel(R) icc 19 >>>" >&5
+$as_echo "<<< C++ compiler is Intel(R) icc 19 >>>" >&6; }
+                            GXX_VERSION=intel_icc_v19.x ;; #(
   *18.*) :
     { $as_echo "$as_me:${as_lineno-$LINENO}: result: <<< C++ compiler is Intel(R) icc 18 >>>" >&5
 $as_echo "<<< C++ compiler is Intel(R) icc 18 >>>" >&6; }
--- a/m4/compiler.m4
+++ b/m4/compiler.m4
@@ -225,6 +225,8 @@ AC_DEFUN([DETERMINE_CXX_BRAND],
                 [
                   GXX_VERSION_STRING="`($CXX -V 2>&1) | grep 'Version '`"
                   AS_CASE("$GXX_VERSION_STRING",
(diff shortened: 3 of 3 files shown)


roystgnr:
Wow, sorry to put you through that, and thanks for finding the fix!

That's a fascinating issue.  So the vector resize can't use a copy constructor because the embedded unique_ptr doesn't allow it, and it can't use a noexcept move constructor because the embedding map doesn't have one?

I never knew about the latter limitation.  I wonder if there are many places where we're forcing needless copy construction because our zillion libmesh_assert statements have kept us out of the habit of marking even move constructors as noexcept... at first glance the only similar use case I see is a ```vector<MessageTag>```, and we never do anything to resize that.

src/systems/fem_context.C
```diff
@@ -80,8 +80,20 @@ void FEMContext::init_internal_data(const System & sys)
 {
   // Reserve space for the FEAbstract and QBase objects for each
   // element dimension possibility (0,1,2,3)
-  _element_fe.resize(4);
-  _side_fe.resize(4);
+
+  // Below is a workaround for the ICC 19. The original code was:
+  //
+  // _element_fe.resize(4);
+  // _side_fe.resize(4);
```
jwpeterson: We've seen this error before with the Intel compiler, and chalked it up to incompatibility between the GCC stdlib installed on the machine and the Intel backend.  I tend to believe the code is actually standards-compliant, but am by no means an expert.

 At least according to my reading of  [this](https://en.cppreference.com/w/cpp/container/vector/resize), as of C++11, `vector<T>::resize()` requires that `T` meet the requirements of "MoveInsertable" and "DefaultInsertable" in order to use overload (1). I would think a `map<FEType, unique_ptr<X>>` meets these requirements but admit I'm not an expert on these two concepts. 

The `vector<T>::push_back()` in your version requires *only* [MoveInsertable](https://en.cppreference.com/w/cpp/container/vector/push_back), I assume we are calling overload (2) there, so that may be why the workaround works...

jwpeterson:
> it can't use a noexcept move constructor because the embedding map doesn't have one?

Is `noexcept` a requirement of either MoveInsertable or DefaultInsertable? I don't see an explicit mention of this in the docs (and from my reading it appears that none of the std containers have noexcept move constructors because [reasons](https://stackoverflow.com/questions/12332772/why-arent-container-move-assignment-operators-noexcept)) but I'm willing to believe.

src/systems/fem_context.C
roystgnr: That page does seem to explicitly accept the possibility of a throwing move constructor, when referring to the strong exception guarantee:

> In overload (1), if T's move constructor is not noexcept and T is not CopyInsertable into *this, vector will use the throwing move constructor. If it throws, the guarantee is waived and the effects are unspecified.

roystgnr: And honestly even the standard line pointed out on StackOverflow has similar wording:

"If an exception is thrown other than by the move constructor of a non-CopyInsertable T there are no effects."

If a non-CopyInsertable T was required to have a noexcept move constructor then that whole "other than" clause would be meaningless.

roystgnr: We should still merge this because it's a cheap workaround for a problem with a major compiler, but I'm now leaning towards this being an icpc bug rather than a libmesh bug.

jwpeterson:
I couldn't push the bootstrapping changes directly to your branch, so I made a [PR](https://github.com/hrittich/libmesh/pull/1).

src/systems/fem_context.C
hrittich: My explanation for why the workarounds do work is as follows (but I am no expert either).

The `resize` method fails, because it cannot copy a pair containing an `unique_ptr`. Thus, the tree must have already one storage element allocated. I am assuming, that `map` already creates the root node or something like that (the map is empty), and this root node cannot be copied. Hence, the workaround works because each time a new `map` is created and then moved into the `vector`. In contrast to trying to copy exactly the empty map multiple times into the vector.

I might be wrong, though (and not willing to read more of the STL code, than what I had to read yesterday. ;) ).

jwpeterson: Agree, one of the commenters also mentioned MSVC 18.00.40629 has the same issue. The SO seems to have generated a lot of interest!

jwpeterson:
I'd like to merge this by hand (rather than click green button) and include the SO link as it may continue to be updated over time as standards evolve, etc.
```
https://stackoverflow.com/questions/53527682/deleted-function-in-stdpair-when-using-a-unique-ptr-inside-a-map
```

hrittich:
> I'd like to merge this by hand (rather than click green button) and include the SO link as it may continue to be updated over time as standards evolve, etc.
> 
> ```
> https://stackoverflow.com/questions/53527682/deleted-function-in-stdpair-when-using-a-unique-ptr-inside-a-map
> ```

As you wish. I'll pull your changes when you are done.

PR #1959, opened 2018-11-29
Add index_range() helper function.

jwpeterson:
This helper allows us to rewrite loops over `std::vectors` in which indices are required in such a way that:

1. No explicit index type is used at the call site.
1. We avoid repeated calls to `vector::size()` in the loop termination test (although it's not clear if that's really much of an optimization...)

So, for example, a typical for-loop could be rewritten as:
```diff
-      for (unsigned int s = 0; s != sysnames.size(); ++s)
+      for (auto s : index_range(sysnames))
```
Note that the underlying index type is `std::size_t` as that is the type returned by `vector::size()`. It is still possible to do explicit signed integer iteration by constructing an `IntRange<int>` on the fly, and I have also done that in a few cases in this patch. I still need to test how well compilers are able to vectorize `IntRange<int>` loops... if it doesn't seem to work, I'll revert to non range-based loops instead in those cases.  

As you might imagine, we have many places in the library where we can use this helper, I've only fixed about 10% of them in this initial PR. I'm planning to keep all these small commits on this branch just in case I've broken something and need to go back and bisect.

### DIFF
--- a/src/apps/meshavg.C
+++ b/src/apps/meshavg.C
@@ -18,13 +18,12 @@
 
 // Open the solution files named on standard input, take the average
 // of their fields' values, and output to a new solution file.
-
 #include "libmesh/libmesh.h"
-
 #include "libmesh/equation_systems.h"
 #include "libmesh/mesh.h"
 #include "libmesh/namebased_io.h"
 #include "libmesh/numeric_vector.h"
+#include "libmesh/int_range.h"
 
 using namespace libMesh;
 
@@ -73,7 +72,7 @@ int main(int argc, char ** argv)
                EquationSystems::READ_BASIC_ONLY);
       libMesh::out << "Loaded next solution " << argv[i] << std::endl;
 
-      for (unsigned int s = 0; s != sysnames.size(); ++s)
+      for (auto s : IntRange<std::size_t>(0, sysnames.size()))
         {
           if (!es2.has_system(sysnames[s]))
             libmesh_error_msg("EquationSystems object does not have " << sysnames[s]);
@@ -84,7 +83,7 @@ int main(int argc, char ** argv)
 
   int n_solutions = argc - 3;
 
-  for (unsigned int s = 0; s != sysnames.size(); ++s)
+  for (auto s : IntRange<std::size_t>(0, sysnames.size()))
     {
       (*summed_solutions[s]) /= n_solutions;
       es1.get_system(s).solution->swap(*summed_solutions[s]);
--- a/src/apps/meshavg.C
+++ b/src/apps/meshavg.C
@@ -72,7 +72,7 @@ int main(int argc, char ** argv)
                EquationSystems::READ_BASIC_ONLY);
       libMesh::out << "Loaded next solution " << argv[i] << std::endl;
 
-      for (auto s : IntRange<std::size_t>(0, sysnames.size()))
+      for (auto s : index_range(sysnames))
         {
           if (!es2.has_system(sysnames[s]))
             libmesh_error_msg("EquationSystems object does not have " << sysnames[s]);
@@ -83,7 +83,7 @@ int main(int argc, char ** argv)
 
   int n_solutions = argc - 3;
 
-  for (auto s : IntRange<std::size_t>(0, sysnames.size()))
+  for (auto s : index_range(sysnames))
     {
       (*summed_solutions[s]) /= n_solutions;
       es1.get_system(s).solution->swap(*summed_solutions[s]);
--- a/src/base/default_coupling.C
+++ b/src/base/default_coupling.C
@@ -19,11 +19,11 @@
 
 // Local Includes
 #include "libmesh/default_coupling.h"
-
 #include "libmesh/coupling_matrix.h"
 #include "libmesh/elem.h"
 #include "libmesh/periodic_boundaries.h"
 #include "libmesh/remote_elem.h"
+#include "libmesh/int_range.h"
 
 // C++ Includes
 #include <unordered_set>
@@ -160,7 +160,7 @@ void DefaultCoupling::operator()
               active_neighbors.push_back(neigh);
 #endif
 
-              for (std::size_t a=0; a != active_neighbors.size(); ++a)
+              for (auto a : index_range(active_neighbors))
                 {
                   const Elem * neighbor = active_neighbors[a];
 
--- a/src/base/default_coupling.C
+++ b/src/base/default_coupling.C
@@ -160,10 +160,8 @@ void DefaultCoupling::operator()
               active_neighbors.push_back(neigh);
 #endif
 
-              for (auto a : index_range(active_neighbors))
+              for (const auto & neighbor : active_neighbors)
                 {
-                  const Elem * neighbor = active_neighbors[a];
-
                   if (!elements_checked.count(neighbor))
                     next_elements_to_check.insert(neighbor);
 
--- a/src/error_estimation/adjoint_refinement_estimator.C
+++ b/src/error_estimation/adjoint_refinement_estimator.C
@@ -45,7 +45,7 @@
 #include "libmesh/adjoint_refinement_estimator.h"
 #include "libmesh/enum_error_estimator_type.h"
 #include "libmesh/enum_norm_type.h"
-
+#include "libmesh/utility.h"
 
 #ifdef LIBMESH_ENABLE_AMR
 
@@ -507,11 +507,9 @@ void AdjointRefinementEstimator::estimate_error (const System & _system,
               // We will have to manually do the dot products.
               Number local_contribution = 0.;
 
-              for (std::size_t j=0; j != dof_indices.size(); j++)
-                {
-                  // The contribution to the error indicator for this element from the current QoI
-                  local_contribution += (*localized_projected_residual)(dof_indices[j]) * (*localized_adjoint_solution)(dof_indices[j]);
-                }
+              // Sum the contribution to the error indicator for each element from the current QoI
+              for (const auto & dof : dof_indices)
+                local_contribution += Utility::pow<2>((*localized_projected_residual)(dof));
 
               // Multiply by the error weight for this QoI
               local_contribution *= error_weight;
--- a/src/error_estimation/adjoint_refinement_estimator.C
+++ b/src/error_estimation/adjoint_refinement_estimator.C
@@ -509,7 +509,7 @@ void AdjointRefinementEstimator::estimate_error (const System & _system,
 
               // Sum the contribution to the error indicator for each element from the current QoI
               for (const auto & dof : dof_indices)
-                local_contribution += Utility::pow<2>((*localized_projected_residual)(dof));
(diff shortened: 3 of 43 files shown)


jwpeterson:
Some promising results. At least with clang 6.0.1, use of range-based for loops does not preclude vectorization for simple fixed-size loops.
```
test_clang_vectorizer.cc:69:3: remark: vectorized loop (vectorization width: 2, interleaved count: 1) [-Rpass=loop-vectorize]
  for (int i = 0; i < N; ++i)
  ^
test_clang_vectorizer.cc:78:3: remark: vectorized loop (vectorization width: 2, interleaved count: 1) [-Rpass=loop-vectorize]
  for (auto i : IntRange<int>(0, N))
  ^
```

jwpeterson:
And perhaps quite interestingly, clang also reports that it vectorizes a similar loop over a `std::vector` using unsigned (`std::size_t`) iteration and the new `index_range` helper function.
```
test_clang_vectorizer.cc:92:3: remark: vectorized loop (vectorization width: 2, interleaved count: 1) [-Rpass=loop-vectorize]
  for (auto i : index_range(vec1))
  ^
```

jwpeterson:
@pbauman is there maybe an obvious error I've done which is causing GRINS to fail?

src/error_estimation/adjoint_refinement_estimator.C
```diff
@@ -507,11 +507,9 @@ void AdjointRefinementEstimator::estimate_error (const System & _system,
...
               Number local_contribution = 0.;
 
-              for (std::size_t j=0; j != dof_indices.size(); j++)
-                {
-                  // The contribution to the error indicator for this element from the current QoI
-                  local_contribution += (*localized_projected_residual)(dof_indices[j]) * (*localized_adjoint_solution)(dof_indices[j]);
-                }
+              // Sum the contribution to the error indicator for each element from the current QoI
+              for (const auto & dof : dof_indices)
+                local_contribution += Utility::pow<2>((*localized_projected_residual)(dof));
```
roystgnr: No pow here.  Those are two different vectors being combined.

roystgnr: Although, and this is damning, this looks like it might be *triggering* but not *causing* the GRINS error.  I'll have to try this change out on master myself some time and see if I can replicate.

jwpeterson: Oh crap, my eyes must have been going crossed when I changed that line. Will fix.

jwpeterson: GRINS seems happier after this fix. I'm glad we have the extra coverage!

roystgnr: No kidding.  I can't even quantify the extra coverage here, though - libMesh "make check" uses this code in two examples!  I guess the trouble is it's an error indicator and libMesh examples aren't have good enough regression tests to tell us if the refinement patterns are hosed.

src/apps/meshavg.C
```diff
@@ -73,7 +72,7 @@ int main(int argc, char ** argv)
                EquationSystems::READ_BASIC_ONLY);
       libMesh::out << "Loaded next solution " << argv[i] << std::endl;
 
-      for (unsigned int s = 0; s != sysnames.size(); ++s)
+      for (auto s : index_range(sysnames))
```
roystgnr: This patch is going to be full of -Wconversion warning regressions, isn't it?

I'd say merge it anyway, though.  I never got that cleanup finished and IIRC a proper cleanup would end up including a 99.9%-backwards-compatible-but-damn-it-MOOSE-uses-the-other-0.1%-too API change.

src/error_estimation/adjoint_refinement_estimator.C
jwpeterson: Agreed, our reliance on dependent libraries is pretty embarrassing, but who really has the time/desire to make the examples into a real regression test suite with gold files, etc.?

src/apps/meshavg.C
jwpeterson: > This patch is going to be full of -Wconversion warning regressions, isn't it?

Damn, I really hope not. Let me try compiling with that flag turned on, but what about the new code do you think would trigger the warning?

src/base/default_coupling.C
```diff
@@ -160,10 +160,8 @@ void DefaultCoupling::operator()
               active_neighbors.push_back(neigh);
 #endif
 
-              for (std::size_t a=0; a != active_neighbors.size(); ++a)
+              for (const auto & neighbor : active_neighbors)
```
roystgnr: At some point I want to look at ```g++ -S``` output and verify that we don't actually lose anything by using a reference-to-pointer rather than copy-of-pointer in these range for loops over containers of ```Elem *```.  In theory I could imagine the compiler deciding that it can't prove the container's contents aren't being changed by function calls inside the loop and so it needs to do a double-dereference instead of a single for later Elem accesses.

jwpeterson: For the following simple test code:
```C++
#include <vector>

int main()
{
  std::vector<int *> v = {new int(1), new int(2)};
  int sum = 0;
  for (auto & ptr : v)
    sum += *ptr;
  
  return sum;
}
```
there is no difference in the generated assembly regardless of the `&` when compiling with `-O3` with g++ (GCC) 7.3.1 20180902.

src/apps/meshavg.C
roystgnr: We have a number of places where our APIs take "unsigned int", and allowing C++ to silently downcast a size_t to that can give a warning.  I noticed some more that really stuck out later - anywhere you see me using ```cast_int<unsigned int>``` or ```IntRange<unsigned int>``` in a for loop, you can rest assured I didn't write longer code specifying a smaller int type for fun.  :-P

But like I said, don't worry about it.  This isn't like our paranoid warnings where we got everything ship-shape and plan to keep it that way forever, this is a warning that we've never fully avoided and that I just like to check on every year or so to help myself look for potential 64-bit truncation bugs.

src/base/default_coupling.C
roystgnr: Oh, there's definitely not going to be a difference there.  But create a function that's in a separate translation unit and that has access to v, call it inside the loop, and see if *that* makes a difference?

src/solvers/nlopt_optimization_solver.C
```diff
@@ -78,16 +78,15 @@ double __libmesh_nlopt_objective(unsigned n,
...
+          solver->gradient_object->gradient(*(sys.current_local_solution), *(sys.rhs), sys);
 
           // we've filled up sys.rhs with the gradient data, now copy it
           // to the nlopt data structure
           libmesh_assert(sys.rhs->size() == n);
 
           std::vector<double> grad;
           sys.rhs->localize_to_one(grad);
-          for (unsigned i=0; i<n; ++i)
+          for (auto i : IntRange<dof_id_type>(0, n))
```
roystgnr: Good catch... but shouldn't we be changing the type of n to dof_id_type as well?

src/solvers/nlopt_optimization_solver.C
```diff
@@ -128,7 +127,8 @@ void __libmesh_nlopt_equality_constraints(unsigned m,
   if (sys.solution->size() != n)
     libmesh_error_msg("Error: Input vector x has different length than sys.solution!");
 
-  for (unsigned int i=sys.solution->first_local_index(); i<sys.solution->last_local_index(); i++)
+  for (auto i : IntRange<dof_id_type>(sys.solution->first_local_index(),
+                                      sys.solution->last_local_index()))
```
roystgnr: Should we add an index_range overload that takes NumericVector?

src/solvers/nlopt_optimization_solver.C
```diff
@@ -232,7 +233,7 @@ void __libmesh_nlopt_inequality_constraints(unsigned m,
       // TODO: Even better would be if we could use 'result' directly
       // as the storage of ineq_constraints.  Perhaps a serial-only
       // NumericVector variant which supports this option?
-      for (unsigned i=0; i<m; ++i)
+      for (auto i : IntRange<dof_id_type>(0, m))
```
roystgnr: Same question in this function about the types of m and n in the declaration.

src/solvers/nlopt_optimization_solver.C
jwpeterson: Good catch, will fix this and the other one.

src/solvers/nlopt_optimization_solver.C
jwpeterson: Sure. For that one I'd probably hard code `numeric_index_type`? That seems to be what we're using with `NumericVectors`...

roystgnr: Right.

src/solvers/nlopt_optimization_solver.C
jwpeterson: So here we are somewhat constrained by the [nlopt interface](https://nlopt.readthedocs.io/en/latest/NLopt_Reference/#objective-function) itself which is expecting an `unsigned int`.
```
double f(unsigned n, const double* x, double* grad, void* f_data);
```
That probably means that we should `static_assert` that we aren't using the nlopt code when `sizeof(dof_id_type) != 4`...

src/base/default_coupling.C
jwpeterson: No difference in assembly for this simple case either...
```C++
#include <vector>

// Function defined in different TU. Note: the function's not really
// defined anywhere, as we are just generating assembly.
int f(int x);

int main()
{
  std::vector<int *> v = {new int(1), new int(2)};
  int sum = 0;
  for (auto & ptr : v)
    sum += f(*ptr);
  
  return sum;
}
```

roystgnr: Looks like my theory was correct; you just have to get a little more complicated to show it:
```
#include <vector>

// Function defined in different TU. Note: the function's not really
// defined anywhere, as we are just generating assembly.
template <typename T>
int f(T & v, int x);

int main()
{
  std::vector<int *> v = {new int(1), new int(2)};
  int sum = 0;
  for (auto & ptr : v)
    {
      sum += f(v, *ptr);
      sum += f(v, *ptr);
    }

  return sum;
}
```

If you leave the reference in the loop iterator declaration, then (with gcc 8.1 defaults at least) the assembly code does a double-dereference before each use of ```*ptr```; if you remove the reference then a single-dereference gets pulled up to the top of the loop body, for one fewer dereference per ```*ptr```-use-after-the-first per loop iteration.

If I turn on -O3 with gcc 8.1 I can't make head or tail of the assembly anymore, but (after supplying a "return 1" stub f in a separate file and making the loops longer) callgrind seems to think a loop with N iterations over M ```ptr``` uses each entails ```(M-1)*N-2``` more Ir (instruction read) events with the reference in the code.  (Unless M is 1 or 0, in which case whatever that "-2" is goes away and the codes are identical)

roystgnr:
:+1: for the new stuff too.  Don't be distracted by my worries about which iterator style can save a nanosecond over which other iterator style.  ;-)

PR #1960, opened 2018-11-30
Add back in missing reinterpret_cast.

jwpeterson:
Make it more explicit this time, so it isn't missed in the future...

Refs #1953.


roystgnr:
Looks like that was the only breakage.

Do you or @pbauman have Trilinos modules available?  Regular testing using its solvers might be excessive (well, *is* excessive for MOOSE), but we should at least be catching build failures in CI.

PR #1962, opened 2018-12-04
Fix for example when Exodus II is not available.

hrittich:
The example `systems_of_equations_ex9` fails, when Exodus II is not available. This PR adds a corresponding check to the example.


jwpeterson:
Does it really "fail" or just not write output? The I/O appears to be properly guarded by `#ifdef LIBMESH_HAVE_EXODUS_API`.

Would it be OK if we instead wrote out a different file format (say xda) instead of not running the example at all when Exodus is not available?

hrittich:
> Does it really "fail" or just not write output? The I/O appears to be properly guarded by `#ifdef LIBMESH_HAVE_EXODUS_API`.

It fails, because it tries to load the file `systems_of_equations_ex9.exo`.
 
> Would it be OK if we instead wrote out a different file format (say xda) instead of not running the example at all when Exodus is not available?

This idea would only work, if the mesh comes from a different place, e.g., is generated.

jwpeterson:
> It fails, because it tries to load the file `systems_of_equations_ex9.exo`.

Ah, I missed that. In that case your patch is the right way to go.

PR #1964, opened 2018-12-04
Fix output data format in legacy SCM writes

roystgnr:
This fixes reduced_basis_ex2 for me, from a bug in 7b63e4a which
somehow made it through CI.

Apologies to @dknez and I hope this didn't bite any of his users.  Is there any easy way we could get that example running in parallel?  I apparently go many months in between running "make check" on one processor.

@jwpeterson, could we add SLEPC and GLPK modules to the environment of "Test"?

### DIFF
--- a/src/reduced_basis/rb_scm_evaluation.C
+++ b/src/reduced_basis/rb_scm_evaluation.C
@@ -395,7 +395,7 @@ void RBSCMEvaluation::legacy_write_offline_data_to_files(const std::string & dir
       file_name << directory_name << "/C_J_length" << suffix;
       Xdr C_J_length_out(file_name.str(), mode);
 
-      std::size_t C_J_length = C_J.size();
+      unsigned int C_J_length = cast_int<unsigned int>(C_J.size());
       C_J_length_out << C_J_length;
       C_J_length_out.close();
 


dknez:
> Is there any easy way we could get that example running in parallel? I apparently go many months in between running "make check" on one processor.

We use "-eps-type lapack" (which is serial only) in that example for the eigensolves. We can omit "-eps-type lapack" and hence run it in parallel, but I think in that case the eigensolver doesn't converge, so I don't know of a good way to get this working in parallel unfortunately. IIRC the eigenproblems for SCM often have convergence issues, which is why we resorted to LAPACK in that example.

jwpeterson:
> @jwpeterson, could we add SLEPC and GLPK modules to the environment of "Test"?

I think SLEPc builds already exist in the MOOSE package, and we may even already have the ability to attach SLEPc tests to libMesh PRs. If not, that could probably be added.

GLPK is a GNU library. It's not something we currently have on our build boxes, but there are pre-built (glpk-doc, glpk-utils, libglpk-dev) packages for Ubuntu that we could probably add to our VM setup routine. I'll have @milljm take a look into adding that first, then we could think about adding a recipe that uses it.

jwpeterson:
FYI: optional SLEPc test added to this PR, so that bit is definitely already possible. I'd probably just modify that recipe in the future to append `--enable-glpk` and we'd go from there.

milljm:
I failed to mention I added glpk libs to the containers running the show on Civet. A little late, but mentioning this now, here, for good measure.

Issue #1950, opened 2018-11-26
Memory Usage and time in Partitioner::set_node_processor_ids()


friedmud:
I'm seeing some pretty massive memory usage and time spent when trying to split a mesh with 100M+ elements during `Partitioner::set_node_processor_ids()`.

This is well before the communication portion of that routine...

The memory usage when the program entered that routine was around 50GB of RAM.  It's currently climbed to 155GB and counting (I only have 187GB on this node so hopefully it stops soon!).  It has also taken a REALLY long time.  (maybe an hour?).

Here is a current stack trace:

```
#0  0x00002aaab0a5d451 in __memmove_ssse3_back () from /lib64/libc.so.6
#1  0x00002aaaacb90112 in libMesh::Partitioner::set_node_processor_ids(libMesh::MeshBase&) () from /home/gastdr/projects/lemhi/libmesh/lib/libmesh_opt.so.0
#2  0x00002aaaacb923fa in libMesh::Partitioner::partition(libMesh::MeshBase&, unsigned int) () from /home/gastdr/projects/lemhi/libmesh/lib/libmesh_opt.so.0
#3  0x00002aaaac93a388 in libMesh::split_mesh(libMesh::MeshBase&, unsigned int) () from /home/gastdr/projects/lemhi/libmesh/lib/libmesh_opt.so.0
#4  0x00002aaaab89c031 in SplitMeshAction::act() () from /home/gastdr/projects/lemhi/moose/framework/libmoose-opt.so.0
#5  0x00002aaaab899134 in Action::timedAct() () from /home/gastdr/projects/lemhi/moose/framework/libmoose-opt.so.0
#6  0x00002aaaab8a3c69 in ActionWarehouse::executeActionsWithAction(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) () from /home/gastdr/projects/lemhi/moose/framework/libmoose-opt.so.0
#7  0x00002aaaab8a5348 in ActionWarehouse::executeAllActions() () from /home/gastdr/projects/lemhi/moose/framework/libmoose-opt.so.0
#8  0x00002aaaabec90ae in MooseApp::runInputFile() () from /home/gastdr/projects/lemhi/moose/framework/libmoose-opt.so.0
#9  0x00002aaaabec7660 in MooseApp::run() () from /home/gastdr/projects/lemhi/moose/framework/libmoose-opt.so.0
#10 0x0000000000404377 in main ()
```

[...]

PR #1965, opened 2018-12-04
Re-implement parallel sync using the NBX algorithm

roystgnr:
This takes the algorithm from @friedmud in #1826, but fixes the MPI_TAG_UB bug, retains the ability to generate a MessageTag with a specified value, and adds a dbg/devel mode verification of automatic tag values when those are included.

This still needs a couple unit tests and I'm just getting started with my own private testing but I think it's in shape to throw at Civet and I'm done for today.

Also, we need to look at the tag choices in System::write_serialized_blocked_dof_objects - is it just me, or is that going to risk failure once num_blks exceeds 100, with only a few hundred million elements?

### DIFF
--- a/include/parallel/communicator.h
+++ b/include/parallel/communicator.h
@@ -215,6 +215,11 @@ public:
   void barrier () const;
 
   /**
+   * Start a barrier that doesn't block
+   */
+  void nonblocking_barrier (Request & req) const;
+
+  /**
    * Verify that a local variable has the same value on all processors.
    * Containers must have the same value in every entry.
    */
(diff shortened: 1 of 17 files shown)


friedmud:
@roystgnr sorry I dropped out on this - I've had my head buried in my dissertation for the last couple of weeks.  Let me review this for a bit and get back to you...

roystgnr:
I'm guessing that Bison ```Node #427 specified in 'top_disp_r_fuel' not found in the mesh!``` in serial wasn't likely to be due to a change in parallel codes that pass all the other tests.  Hopefully it'll have been fixed by this next push.

roystgnr:
So, N->M pushes and pulls are now working properly in unit tests, which means there's a chance that this PR also fixes the underlying bug(s) behind #1950 and partially obviates #1815... but now that I'm moving on to try something more complicated than a unit test, I find I can't replicate the problem in #1950.  At least not on less than a hundred thousand elements; I'm trying out half a million next.

In hindsight this shouldn't have surprised me, since the trigger for the bug must have been a hell of a corner case for us to not hit it sooner.  @friedmud, can you send me a test case (MOOSE input file + mesh file(s)?) that goes nuts in either fashion?  Something I can fit on 24 cores and 64GB would be ideal; preferably still something that triggers within 640 cores and 1280GB otherwise.

Even if we don't get that testing done right away or this turns out to be an insufficient fix, I'm increasingly confident that it's a solid improvement, but since it's also a major change IMHO we should still wait until after 1.4.0 branches before merging so our git users can shake it down before our release users get it.

roystgnr:
@jwpeterson recently got a new release out with the older more tested code, and @bboutkov recently reported some solid performance improvement results for this on libmesh-devel, and I recently wasted some time after accidentally triggering a bug that this branch fixes, so even if this branch isn't a fix-all I do think it's enough of an improvement to be merged and it's the ideal time to merge it.  I'll do so shortly unless anyone screams.

PR #1969, opened 2018-12-07
gradient with infinite elements

BalticPinguin:
Dear all,
I am sorry to mix these things up in one PR:
- 32ae391 contains some minor fixes in the documentations: The main changes are to name the variables in the declaration (``.h``-file) and the function (``.C``-file) consistent.
The main purpose is, however, to adapt the MeshFunction::gradient()-function to infinite elements.
Respectively, several functions and objects are added to FEInterface.

I have one open question related to this, denoted by the FIXME-tag: To get the gradient in physical coordinates, I need to hold the transformation in the FEInterface; therefore, I added the `` std::vector<std::vector<Real>> local_transform`` variable; but maybe there is a better way to do so?
I would think of some Tensor- or matrix-type or so? 

Any other thoughts about this?
I hope, you are all ok with the introduction of this new feature?

Best,
Hubert

### DIFF
--- a/include/fe/fe_compute_data.h
+++ b/include/fe/fe_compute_data.h
@@ -21,6 +21,7 @@
 
 // Local includes
 #include "libmesh/libmesh.h"
+#include "libmesh/fe_base.h" // required for the type OutputGradient.
 
 // C++ includes
 #include <vector>
@@ -57,7 +58,8 @@ public:
   FEComputeData (const EquationSystems & es,
                  const Point & pin) :
     equation_systems(es),
-    p(pin)
+    p(pin),
+    _need_dshape(false)
   {
     this->clear();
   }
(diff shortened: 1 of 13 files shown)


BalticPinguin:
oh; I have seen a typo:
```
@@ -933,7 +933,7 @@ void FEInterface::compute_data(const unsigned int dim,
       data.local_transform.resize(dim);
 
       for (unsigned int d=0; d<dim; d++)
-         data.local_transform[dim].resize(dim);
+         data.local_transform[d].resize(dim);
```
I hope, this is the only one...
My test of this part obviously did not work as expected.

BalticPinguin:
Ok; so there are some properties where my tests didn't work properly.
I'll come back to this later; but for the moment I'll close it.
Sorry for that.

PR #1968, opened 2018-12-06
Temporarily turn off testing of Hilbert SFC partitioner.

jwpeterson:
Refs #1967.

Issue #1967, opened 2018-12-06
Hilbert SFC Partitioner test fails sporadically on OSX


jwpeterson:
These partitioner unit tests were originally merged in 0dc5c33ba (Oct 25) and for some reason they haven't failed on any of the CIVET testing we've done in the interim. I can get the Hilbert SFC test to fail fairly reliably on my Mac (but not every time, maybe some type of memory corruption?). It may be a platform-dependent issue. Unfortunately I can't seem to get a good stack trace down in the Hilbert library code itself-the last line of my trace, `hilbert.c:0`, is completely bogus. The error itself is a `SIGABRT`, so there's not much to go on.

```
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007fff66b5cb86 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff66c12c50 libsystem_pthread.dylib`pthread_kill + 285
    frame #2: 0x00007fff66ac6268 libsystem_c.dylib`__abort + 144
    frame #3: 0x00007fff66ac6af6 libsystem_c.dylib`__stack_chk_fail + 205
    frame #4: 0x00000001031c3568 libmesh_dbg.0.dylib`hilbert(x=0x000000010a408bb0, y=0x000000010a408c90, z=0x000000010a408d70, N=0x00007ffeefbfb894, table=0x000000010a408e50) at hilbert.c:0
    frame #5: 0x000000010284d9f3 libmesh_dbg.0.dylib`libMesh::SFCPartitioner::partition_range(this=0x00007ffeefbfd998, mesh=0x00007ffeefbfda40, beg=element_iterator @ 0x00007ffeefbfcb28, end=element_iterator @ 0x00007ffeefbfcb08, n=2) at sfc_partitioner.C:125
    frame #6: 0x000000010284e96d libmesh_dbg.0.dylib`libMesh::SFCPartitioner::_do_partition(this=0x00007ffeefbfd998, mesh=0x00007ffeefbfda40, n=2) at sfc_partitioner.C:173
    frame #7: 0x00000001005730fe tmp-dbg`libMesh::HilbertSFCPartitioner::_do_partition(this=0x00007ffeefbfd998, mesh=0x00007ffeefbfda40, n=2) at hilbert_sfc_partitioner.h:75
[...]

PR #1970, opened 2018-12-11
Fix buffer overrun in Hilbert SFC Partitioner

jwpeterson:
This PR fixes a buffer overrun that we probably would never have caught except for the fact that it was causing the unit tests to occasionally fail when libmesh was configured with `--disable-strict-lgpl`. Note that I'm not sure if the fix is correct as far as the algorithm is concerned, but it does fix the problem with the aborts and the partitions are unchanged from before (when the code actually ran).

This PR also adds the ability to pass a regular expression (via "--re foo") to the unit tests executables so that only a subset of all the unit tests are run. Now that we have so many unit tests, this can greatly help the turnaround time when you are just trying to debug a single one. This new capability requires C++11 `<regex>` support to work, and falls back on running all tests if your compiler happens to not support that.

Refs #1967 
Refs #1968

### DIFF
--- a/contrib/sfcurves/hilbert.c
+++ b/contrib/sfcurves/hilbert.c
@@ -1,6 +1,7 @@
 #include <stdlib.h>
 #include <limits.h>
 #include <math.h>
+#include <stdio.h>
 #include "sfcurves_internal.h"
 
 
@@ -25,7 +26,7 @@ void hsfc3d(
 
   unsigned i ;
   unsigned char axis[3+3] ;
-  
+
   /* GRAY coding */
 
   if ( ! init ) {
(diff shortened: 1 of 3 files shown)


roystgnr:
This is all pretty sweet, but could we merge some of these commits together?  At first glance it looks like 68e210a should be squashed onto 695647b for one commit, and everything else should be squashed together for a second commit (or if it works cleanly, just keep 22ce754 and drop all the "add debugging statements, remove debugging statements again" commits?).

jwpeterson:
Yeah... I tried to fix up the history but forgot to squash the driver.C commits together in the process. I did kind of want to keep the debugging print statements in the history just in case I ever have to do this again. Let me try rebasing and let me know what you think.

jwpeterson:
Some 2D space filling curve partitioner results on different meshes. Definitely not perfect, but overall looks about right. I'll try and post some with different numbers of procs later...

sfc_subdomain_test
sfc_subdomain_test_roundhole
sfc_subdomain_test_hole

roystgnr:
Huh.  That pid==3 bit in the top on the right has got to be a bug.  Probably not worth digging into though.

roystgnr:
> I did kind of want to keep the debugging print statements in the history just in case I ever have to do this again.

Sounds fair.

PR #1971, opened 2018-12-12
Convert more traditional for-loops to use index_range() helper function

jwpeterson:
This is fairly granular to make bisection a bit easier in case any of the testing fails. I can squash it down later if desired.

### DIFF
--- a/src/geom/elem_cutter.C
+++ b/src/geom/elem_cutter.C
@@ -283,11 +283,11 @@ void ElemCutter::cut_2D (const Elem & elem,
   _inside_elem.clear();
   _outside_elem.clear();
 
-  for (const auto & elem : _inside_mesh_2D->element_ptr_range())
-    _inside_elem.push_back (elem);
+  for (const auto & el : _inside_mesh_2D->element_ptr_range())
+    _inside_elem.push_back (el);
 
-  for (const auto & elem : _outside_mesh_2D->element_ptr_range())
-    _outside_elem.push_back (elem);
+  for (const auto & el : _outside_mesh_2D->element_ptr_range())
+    _outside_elem.push_back (el);
 
 #endif
 }
@@ -364,13 +364,13 @@ void ElemCutter::cut_3D (const Elem & elem,
   _inside_elem.clear();
   _outside_elem.clear();
 
-  for (const auto & elem : _inside_mesh_3D->element_ptr_range())
-    if (elem->volume() > std::numeric_limits<Real>::epsilon())
-      _inside_elem.push_back (elem);
+  for (const auto & el : _inside_mesh_3D->element_ptr_range())
+    if (el->volume() > std::numeric_limits<Real>::epsilon())
+      _inside_elem.push_back (el);
 
-  for (const auto & elem : _outside_mesh_3D->element_ptr_range())
-    if (elem->volume() > std::numeric_limits<Real>::epsilon())
-      _outside_elem.push_back (elem);
+  for (const auto & el : _outside_mesh_3D->element_ptr_range())
+    if (el->volume() > std::numeric_limits<Real>::epsilon())
+      _outside_elem.push_back (el);
 
 #endif
 }
--- a/src/numerics/distributed_vector.C
+++ b/src/numerics/distributed_vector.C
@@ -31,6 +31,7 @@
 #include "libmesh/parallel.h"
 #include "libmesh/parallel_sync.h"
 #include "libmesh/tensor_tools.h"
+#include "libmesh/int_range.h"
 
 namespace libMesh
 {
@@ -51,8 +52,8 @@ T DistributedVector<T>::sum () const
 
   T local_sum = 0.;
 
-  for (numeric_index_type i=0; i<local_size(); i++)
-    local_sum += _values[i];
+  for (auto & val : _values)
+    local_sum += val;
 
   this->comm().sum(local_sum);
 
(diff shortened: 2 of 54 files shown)


pbauman:
GRINS failure:

<pre>
src/partitioning/libmesh_devel_la-partitioner.lo
/femputer/pbauman/civet_build_testing_root/libMesh/src/parallel/request.C: In function ‘void libMesh::Parallel::wait(std::vector<libMesh::Parallel::Request>&)’:
/femputer/pbauman/civet_build_testing_root/libMesh/src/parallel/request.C:215:15: error: declaration of ‘request’ shadows a global declaration [-Werror=shadow]
   for (auto & request : r)
               ^~~~~~~
In file included from /femputer/pbauman/civet_build_testing_root/libMesh/src/parallel/request.C:19:0:
./include/libmesh/request.h:48:21: note: shadowed declaration is here
 typedef MPI_Request request;
</pre>

jwpeterson:
Thanks @pbauman. Same problem with the complex numbers build. I guess I need to add `-Werror -Wshadow` to my `CXXFLAGS`... 🙄

roystgnr:
Just configure with --enable-paranoid-warnings --enable-werror so you get everything we enforce.

jwpeterson:
I fixed all the paranoid warnings that showed up for clang, but the one pointed out by @pbauman wasn't even there... if there are more failures, I'll compile with GCC and fix those as well.

jwpeterson:
> I'll compile with GCC and fix those as well.

So at least with GCC 6.2.0 it's not complaining about any more `-Wshadow` errors for me... I'll see what the build boxes say. And I can actually see the femputer output since I'm working from home so I should be able to eventually fix things by trial and error...

jwpeterson:
Squashed quite a bit and rebased, if this still passes I will go ahead and merge.

src/geom/elem_cutter.C
```diff
@@ -283,11 +283,11 @@ void ElemCutter::cut_2D (const Elem & elem,
   _inside_elem.clear();
   _outside_elem.clear();
 
-  for (const auto & elem : _inside_mesh_2D->element_ptr_range())
-    _inside_elem.push_back (elem);
```
roystgnr: How did this *not* trigger -Wshadow before?

src/partitioning/subdomain_partitioner.C
```diff
@@ -56,14 +57,12 @@ void SubdomainPartitioner::_do_partition (MeshBase & mesh,
...
   // subdomains in question, and pass it to the internal Partitioner.
-  for (std::size_t c=0; c<chunks.size(); ++c)
-    {
-      MeshBase::element_iterator
-        it = mesh.active_subdomain_set_elements_begin(chunks[c]),
-        end = mesh.active_subdomain_set_elements_end(chunks[c]);
-
-      _internal_partitioner->partition_range(mesh, it, end, n);
-    }
+  for (auto c : index_range(chunks))
```
roystgnr: Do we want an index_range here?  Looks like a range for over chunks would work.

src/reduced_basis/rb_scm_evaluation.C
```diff
@@ -529,7 +529,7 @@ void RBSCMEvaluation::legacy_read_offline_data_from_files(const std::string & di
 
   // Resize C_J based on C_J_stability_vector and Q_a
   C_J.resize( C_J_length );
-  for (std::size_t i=0; i<C_J.size(); i++)
+  for (auto i : index_range(C_J))
```
roystgnr: This could also just be a plain range-for, right?

roystgnr:
:+1: regardless of whether you fix those quibbles

roystgnr:
Merging; this is a solid improvement and there's no telling when John's going to have time to address my nitpicks.

src/geom/elem_cutter.C
jwpeterson: Code is only active in `--disable-strict-lgpl` builds, not sure if we have one of those that *also* has `-Werror -Wshadow`

src/numerics/distributed_vector.C
```diff
@@ -234,13 +231,19 @@ void DistributedVector<T>::add (const NumericVector<T> & v)
...
 
 template <typename T>
-void DistributedVector<T>::add (const T a, const NumericVector<T> & v)
+void DistributedVector<T>::add (const T a, const NumericVector<T> & v_in)
 {
   libmesh_assert (this->initialized());
   libmesh_assert_equal_to (_values.size(), _local_size);
   libmesh_assert_equal_to ((_last_local_index - _first_local_index), _local_size);
 
-  add(a, v);
```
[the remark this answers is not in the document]
jwpeterson: If someone had called it, yeah. It was found by the `-Winfinite-recursion` warning.

src/numerics/sum_shell_matrix.C
```diff
@@ -30,7 +31,7 @@ numeric_index_type SumShellMatrix<T>::m () const
   libmesh_assert(!matrices.empty());
   const numeric_index_type n_rows = matrices[0]->m();
 #ifndef NDEBUG
-  for (std::size_t i=1; i<matrices.size(); ++i)
+  for (auto i : index_range(matrices))
```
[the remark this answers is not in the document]
jwpeterson: Yeah.

src/reduced_basis/rb_scm_evaluation.C
jwpeterson: Yep, I will change that one.

src/partitioning/subdomain_partitioner.C
jwpeterson: Yes, will change.

PR #1972, opened 2018-12-12
Add const accessor methods for dpsi_maps in FEMap

lindsayad:
Allows accessing this data from `const FEMap` objects


lindsayad:
Is there a reason why the grins-dbg target was canceled?

PR #1946, opened 2018-11-19
Change vector and tensor return types

lindsayad:
This allows compatibility with automatic differentiation calculations

PR #1973, opened 2018-12-12
Move TypeVector::unit into the header

lindsayad:
More general programming, e.g. with `T=DualNumber<T2,D>`

### DIFF
--- a/include/numerics/type_vector.h
+++ b/include/numerics/type_vector.h
@@ -1100,6 +1100,31 @@ T cross_norm(const TypeVector<T> & b,
   return std::sqrt(cross_norm_sq(b,c));
 }
 
+template <typename T>
+inline
+TypeVector<T> TypeVector<T>::unit() const
+{
+
+  auto && length = norm();
+
+  libmesh_assert_not_equal_to (length, static_cast<Real>(0.));
+
+#if LIBMESH_DIM == 1
+  return TypeVector<T>(_coords[0]/length);
+#endif
+
+#if LIBMESH_DIM == 2
+  return TypeVector<T>(_coords[0]/length,
+                       _coords[1]/length);
+#endif
+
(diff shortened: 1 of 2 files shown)


include/numerics/type_vector.h
```diff
@@ -1100,6 +1100,31 @@ T cross_norm(const TypeVector<T> & b,
...
 }
 
+template <typename T>
+inline
+TypeVector<T> TypeVector<T>::unit() const
+{
+
+  auto && length = norm();
+
+  libmesh_assert_not_equal_to (length, static_cast<Real>(0.));
```
jwpeterson: Does this line still work if `length` is non-`Real`?

include/numerics/type_vector.h
```diff
@@ -1100,6 +1100,31 @@ T cross_norm(const TypeVector<T> & b,
   return std::sqrt(cross_norm_sq(b,c));
 }
 
+template <typename T>
+inline
+TypeVector<T> TypeVector<T>::unit() const
+{
+
+  auto && length = norm();
```
jwpeterson: You haven't changed `norm()`? So that's always returning `Real` regardless of `T`... maybe you plan to change that later, but I am curious about the use of `auto &&` instead of simply `auto`? Does it avoid a copy or something?

include/numerics/type_vector.h
lindsayad: This works for all `DualNumber` types at least because we overload the comparison operators.

include/numerics/type_vector.h
lindsayad: `norm()` is changed in #1946 (which should be much better off after idaholab/moose#12599). Since `norm` just returns a value, the addition of `&&` makes no difference. I guess I've just been getting in the habit of writing this kind of general code to, in general, avoid copies.

PR #1974, opened 2018-12-12
Pass by reference instead of value in our numerics structures

lindsayad:
This can be discussed, but if we're keeping the code very general, pass-by-value is painful for `T=DualNumber<T2,D>`

### DIFF
--- a/include/numerics/tensor_value.h
+++ b/include/numerics/tensor_value.h
@@ -52,32 +52,32 @@ public:
    * Constructor-from-T.  By default sets higher dimensional
    * entries to 0.
    */
-  explicit TensorValue  (const T xx,
-                         const T xy=0,
-                         const T xz=0,
-                         const T yx=0,
-                         const T yy=0,
-                         const T yz=0,
-                         const T zx=0,
-                         const T zy=0,
-                         const T zz=0);
+  explicit TensorValue  (const T & xx,
+                         const T & xy=0,
+                         const T & xz=0,
+                         const T & yx=0,
+                         const T & yy=0,
+                         const T & yz=0,
+                         const T & zx=0,
+                         const T & zy=0,
+                         const T & zz=0);
 
   /**
    * Constructor-from-scalars.  By default sets higher dimensional
    * entries to 0.
    */
   template <typename Scalar>
-  explicit TensorValue  (const Scalar xx,
-                         const Scalar xy=0,
-                         const Scalar xz=0,
-                         const Scalar yx=0,
-                         const Scalar yy=0,
-                         const Scalar yz=0,
-                         const Scalar zx=0,
-                         const Scalar zy=0,
+  explicit TensorValue  (const Scalar & xx,
+                         const Scalar & xy=0,
+                         const Scalar & xz=0,
+                         const Scalar & yx=0,
+                         const Scalar & yy=0,
+                         const Scalar & yz=0,
+                         const Scalar & zx=0,
+                         const Scalar & zy=0,
                          typename
                          boostcopy::enable_if_c<ScalarTraits<Scalar>::value,
-                         const Scalar>::type zz=0);
+                         const Scalar>::type & zz=0);
 
   /**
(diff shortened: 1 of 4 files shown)


roystgnr:
So (excepting where the compiler can optimize things out), this will greatly speed up a slight fraction of cases, but will slightly slow down a great fraction of cases.  Seems like a wash.

Should we use something like boost::call_traits to try and do the right thing in all cases?

lindsayad:
That sounds like a good idea to me. AFAICT, the only boost dependency in libMesh is in the `pool_allocator`. Should I add another boost dependency or add to `namespace boostcopy` which currently has `enable_if_c`?

roystgnr:
I'd say add to namespace boostcopy assuming it's simple enough.  It's basically just a single template class plus overrides for built-in data types, right?

lindsayad:
Well I don't think we're going to be able to make everything work easily with `call_traits`. Take 
```c++
template <typename T>
template <typename Scalar>
typename boostcopy::enable_if_c<
  ScalarTraits<Scalar>::value,
  TypeVector<typename CompareTypes<T, Scalar>::supertype>>::type
TypeVector<T>::operator * (typename boostcopy::call_traits<Scalar>::param_type scalar) const;
```
for instance. We won't be able to deduce the nested parameter type `Scalar`

lindsayad:
So this program:
```c++
#include "boost/call_traits.hpp"
#include <iostream>
#include <type_traits>

class A {};

template <typename T,
          typename std::enable_if<
              std::is_same<const T,
                           typename boost::call_traits<T>::param_type>::value,
              int>::type = 0>
void f(const T) {
  std::cout << "Called built-in\n";
}

template <typename T,
          typename std::enable_if<
              std::is_same<const T &,
                           typename boost::call_traits<T>::param_type>::value,
              int>::type = 0>
void f(const T &) {
  std::cout << "Called user type\n";
}

int main() {
  A a;
  int b;
  f(a);
  f(b);
}
```
produces
```
Called user type
Called built-in
```
so it does the right thing. So if we want the best of all worlds we'll have to double our function count it seems 😢

roystgnr:
> Is there strong support for the fact that passing a builtin by reference incurs a performance penalty?

Yeah.  There's two penalties: a non-inlined callee has to hit cache one extra time (2-4 cycles instead of 1 or 0 depending on whether function arguments all fit in registers or not), and *any* callee has to assume that any function *it* calls might change the value behind the reference, forcing it to reload that value from cache before reusing it afterwards (2-4 cycles per instance instead of 0).

See https://github.com/libMesh/libmesh/pull/1959#discussion_r238330809 for an empirical example of the latter effect.

That said, I certainly see the circular reasoning involved in trying to deduce a type based on the as-yet-un-deduced type, and I can't see a better way to fix it than your two-versions-of-every-function suggestion, for which I certainly share your distaste.  So I think my final "Don't be distracted by my worries about which iterator style can save a nanosecond over which other iterator style." suggestion there applies here too.

Sorry about the red herring.  I thought that would be an easy improvement to make.  I didn't look in the fine print at https://www.boost.org/doc/libs/1_67_0/libs/utility/call_traits.htm and notice that they were having to call ```do_fill<I,T>``` with explicit template arguments.

lindsayad:
> So I think my final "Don't be distracted by my worries about which iterator style can save a nanosecond over which other iterator style." suggestion there applies here too.

So to be clear, we are favoring passing by reference to multiplying our function count? I would say that if things show up in real application profiling, then that would definitely warrant adding the additional SFINAE overloads.

roystgnr:
> So to be clear, we are favoring passing by reference to multiplying our function count?

Yes.  Both options feel wrong, but the "might be a fraction of a percent slower" wrong doesn't feel as bad as the "have to keep a bunch of redundant code in sync and properly covered by testing" wrong.

I feel like there ought to be some other answer here that's as simple and clever as the "just default the constructor" solution we found to our previous unsolvable problem, but I just don't see it.

PR #1975, opened 2018-12-12
Allow changing the tolerance for an element Jacobian mapping failure

lindsayad:
This is a slippery slope perhaps, but in some ALE simulations I was running into issues where the (element) Jacobian would be > 0 during residual calculations, and then between 0 and -machine_epsilon during the (matrix) Jacobian calculations. PETSc has a graceful way to stop a solve during an issue in the residual computation routine; not so during the Jacobian calculation. The result is an aborted simulation due to some numeric fluctuations leading to a very small negative element Jacobian. This patch made the simulations much more robust. @jwpeterson

### DIFF
--- a/include/fe/fe_map.h
+++ b/include/fe/fe_map.h
@@ -907,7 +907,7 @@ protected:
   /**
    * The Jacobian tolerance used for determining when the mapping fails. The mapping is
    * determined to fail if jac <= jacobian_tolerance. If not set by the user, this number
-   * defaults to -TOLERANCE*TOLERANCE (-1e-12 for double precision calculations)
+   * defaults to 0
    */
   Real jacobian_tolerance;
 
--- a/include/fe/fe_map.h
+++ b/include/fe/fe_map.h
@@ -519,6 +519,12 @@ public:
   { libmesh_assert(!calculations_started || calculate_dxyz);
     calculate_dxyz = true; return JxW; }
 
+  /**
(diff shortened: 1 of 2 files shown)


roystgnr:
My vote would be user-selectable defaulting to 0, but I'm OK with -TOLERANCE*TOLERANCE.

But the way I'd usually handle these sorts of things in my own applications is with a try/catch block that takes the libMesh error and then backtracks on time step length or line search distance or whatever parameter gets you closer to a well-defined problem.  Would that not work for you?

I'm also a little baffled at the description of the problem.  Under what circumstances is your nonlinear solver asking for a Jacobian without first/also asking for a residual?

roystgnr:
Merging because I don't want to hold you up (and because you've now got a patch with obvious utility and total backwards compatibility), but I'm still curious:

> with a parallel run, it is possible that my residual produces say: 0 < jac < machine_epsilon (and so the solve continues on its merry way), while my Jacobian then produces jac <= 0. This has happened.

Does this mean you mean you have a failure occurring on only one subpartition, so only one rank throws the exception, so by the time you catch the exception you're out of sync on the other ranks?

We hit that problem in FIN-S too, and since tolerating the failure was never safe in that app, our only solution was to do something like ```comm.max(threw_exception)``` after every try/catch block, so that if one processor had to backtrack they would all know to backtrack.  That might be overkill for your case but I just wanted to bring it up in case you get stuck in the same corner later.

Issue #1949, opened 2018-11-22
Infinite elements don't find their finite neighbors


BalticPinguin:
Dear all,
after my recent update, the infinite elements miss their neighbors at side 0 (the interfacing finite element): ``elem->neighbor(0)`` is ``0``.
In my case, these are second-order ``Tets`` and ``InfPrism``, respectively.

According to my understanding, the problem is introduced by the change of the used ``side_ptr``-function in called by ``UnstructuredMesh::find_neighbors``:
   * The previously used  ``side_ptr(const unsigned int i)`` always puts a first-order side element; in the ``Tet``-class, the function ``me->side_ptr(i)`` should always give a ``TRI3``; in the ``InfPrism``-class, this is put explicitly.
   * using the respective  ``side_ptr  (std::unique_ptr< Elem > & side, const unsigned int i)``, the ``InfPrism``-version ensures that afterwards the side is of type ``TRI3`` while the ``Tet``-version calls ``build_side_ptr``, which should either be ``TRI3`` or ``TRI6``, depending on the order being used.
 
Without understanding the  ``key``-based part,  I could imagine that the keys don't match and the function ``find_neighbors`` does not work anymore, leaving the neighbor nr. 0 to be the null-pointer.

Do you agree with this argumentation or would you rather see the problem somewhere else? (I came to this because the only recent change related to the problem seems to be 7f7e62ea).
If you do, the is to make "The ``side`` is always the first-order" or " is always the real order" scheme consistent. Which one is preferable in this case?

Best,
Hubert

BalticPinguin:
Ok, I have played with this and found that indeed, making side_ptr() and build_side_ptr()-calls consistent fixes this issue.
I would like to make a PR with this respect, but would like to hear some oppinion about which one to use? (I would assume that the side_ptr()-function is faster, but I would like to avoid breaking something due to wrong order).

Thanks in advance

BalticPinguin:
Since my initial fix was not very good, I am working further on this. However, I came across a behaviour that I don't understand: The function
```Elem::simple_build_side_ptr (std::unique_ptr<Elem> & side, const unsigned int i,  ElemType sidetype) ``` (in `include/libmesh/elem.h:2254`) just ignores the sidetype given.
I would consider this as unexpected.
Probably @roystgnr or @jwpeterson could comment on this?

[...]

PR #1977, opened 2018-12-14
build_simple_side_ptr fix, unit tests

roystgnr:
Hopefully this fixes the underlying problem behind #1949 too.  @BalticPinguin would you give it a try?


roystgnr:
Going to merge this right away if it passes Civet; even if it turns out not to completely fix the bug in question, it does fix a serious bug and add a ton of unit tests.

BalticPinguin:
Oh, this somehow went under my radar; sorry.
But it fixed the issue.
Thank you very much.

PR #1981, opened 2018-12-19
gradient with infinite elements (2)

BalticPinguin:
This is the corrected version of #1969, having some minor changes (and rebased to current master).

 * 671fe8b  contains some minor fixes in the documentations: The main changes are to name the variables in the declaration (.h-file) and the function (.C-file) consistent.
    The main purpose is, however, to adapt the MeshFunction::gradient()-function to infinite elements.
    Respectively, several functions and objects are added to FEInterface.

I have one open question related to this, denoted by the FIXME-tag: To get the gradient in physical coordinates, I need to hold the transformation in the FEInterface; therefore, I added the std::vector<std::vector<Real>> local_transform variable; but maybe there is a better way to do so?
I would think of some Tensor- or matrix-type or so?

Any other thoughts about this?
I hope, you are ok with the introduction of this new feature?

Best,
Hubert

### DIFF
--- a/src/fe/inf_fe.C
+++ b/src/fe/inf_fe.C
@@ -85,8 +85,9 @@ void InfFE<Dim,T_radial,T_map>::attach_quadrature_rule (QBase * q)
   // in radial direction, always use Gauss quadrature
   radial_qrule.reset(new QGauss(1, radial_int_order));
 
-  // currently not used. But maybe helpful to store the QBase *
-  // with which we initialized our own quadrature rules
+  // Maybe helpful to store the QBase *
+  // with which we initialized our own quadrature rules.
+  // Used e.g. in \p InfFE::reinit(elem,side)
   qrule = q;
 }
 
--- a/src/fe/inf_fe.C
+++ b/src/fe/inf_fe.C
@@ -217,7 +217,7 @@ void InfFE<Dim,T_radial,T_map>::reinit(const Elem * inf_elem,
       // right now, and it will generalize a bit, and it won't break
       // the assumptions elsewhere in InfFE.
       std::vector<Point> radial_pts;
-      for (auto p : index_range(*pts))
+      for (std::size_t p=1; p < pts->size(); ++p)
         {
           Real radius = (*pts)[p](Dim-1);
           //IMHO this is a dangerous check:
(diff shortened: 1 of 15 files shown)


src/fe/inf_fe.C
```diff
@@ -217,7 +217,7 @@ void InfFE<Dim,T_radial,T_map>::reinit(const Elem * inf_elem,
       // right now, and it will generalize a bit, and it won't break
       // the assumptions elsewhere in InfFE.
       std::vector<Point> radial_pts;
-      for (auto p : index_range(*pts))
+      for (std::size_t p=1; p < pts->size(); ++p)
```
roystgnr: What happened here?  Why are you skipping p=0?

src/mesh/mesh_function.C
```diff
@@ -556,19 +590,55 @@ void MeshFunction::discontinuous_gradient (const Point & p,
...
+          // Otherwive it is wrong...
+#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+          if (!element->infinite())
+            {
+#endif
+              std::unique_ptr<FEBase> point_fe (FEBase::build(dim, fe_type));
+              const std::vector<std::vector<RealGradient>> & dphi = point_fe->get_dphi();
+              point_fe->reinit(element, & point_list);
+
+              temp_output[index] = grad;
```
roystgnr: Should this be here?  It looks like it's redundantly zeroing out something we'll write the *correct* value to later.

src/mesh/mesh_function.C
```diff
@@ -597,6 +667,13 @@ void MeshFunction::hessian (const Point & p,
     }
   else
     {
+#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+      if(element->infinite())
+        libmesh_warning("Warning: Requested the Hessian of an Infinit element."
```
roystgnr: s/Infinit/Infinite

roystgnr:
Those are the only flaws I can find.  :+1: after they're fixed, though I'd feel at least 10 times safer if you could add some unit test coverage for the new features.

src/fe/inf_fe.C
BalticPinguin: Oh; you are right. Here I must have mixed up the commit my merging the master...

BalticPinguin:
Yes, I completely agree with you: I think none of the changes currently are tested.
I have still some things in my application code which are not working as expected, but I think the given changes will be incorporated to the miscellaneous example 14.
Many thanks for the review.

src/mesh/mesh_function.C
BalticPinguin: you are right; this is not necessary at all.

Issue #1956, opened 2018-11-28
second derivative quantities


BalticPinguin:
Dear all,
I have a rather general question about the second-derivative quantities:
Many of them are in an ``if LIBMESH_ENABLE_SECOND_DERIVATIVES``-clause (e.g. ``get_d2xyzdzeta2()``) while others, such as ``get_d2xyzdeta2()`` are not (see ``include/fe/fe_abstract.h``).
Is there some hidden reason behind it?
Naively, I would expect that without second derivatives, none of these quantities should be accessible, right?
Thanks in advance,
Hubert

BalticPinguin:
Ok, thank you for that confirmation. I just wanted to avoid starting to add those guards and realizing only later that these quantities are still used somewhere.
But than I will work on this.

roystgnr:
On Wed, 28 Nov 2018, Hubert Weißmann wrote:

> I have a rather general question about the second-derivative quantities:
> Many of them are in an if LIBMESH_ENABLE_SECOND_DERIVATIVES-clause (e.g. get_d2xyzdzeta2()) while others, such as get_d2xyzdeta2() are not (see
> include/fe/fe_abstract.h).
> Is there some hidden reason behind it?

Possibly?  I vaguely recall there was some non-second-derivative
related code (curvature of boundaries???) that used second derivative
information, and I didn't want that disabled along with the rest?

But "Roy made an oversight" would be a pretty likely theory too.

> Naively, I would expect that without second derivatives, none of
> these quantities should be accessible, right?

That would be my first guess in your shoes.

To be honest, though, --disable-second is a bit of an atavism now.  It
*originally* allowed 4th order problems to be solved without forcing
2nd order codes to calculate a bunch of terms they didn't need, but
later on I added prerequest semantics to the FE class so there's now
a natural way to enable or disable those calculations at runtime
rather than configure time.
---
Roy

PR #1983, opened 2018-12-30
Hiding second derivative quantities in their macro

BalticPinguin:
Dear all,
in reference to #1956: This is supposed to be the patch: 
Except for the calculate_d2phi in the FEAbstract class (which enters some function arguments), all second derivative quantities are now optional via the respective macro.
I hope, I didn't forget something; but for me, it runs well (probably all the test cases have second derivatives enabled?)
Best,
Hubert


roystgnr:
Yeah, we don't even have any optional tests with --disable-second.  Let me test it manually while I look it over.

roystgnr:
Thanks!

Issue #1985, opened 2019-01-03
Resolution of hanging-node constraints incorrect for matrix-free/finite difference formation of Jacobian


lindsayad:
My understanding of the current set-up for a Newton solve (with PETSc at least) *with user-provided Jacobian and preconditioner*:

1. `SNESComputeFunction` gets called -> callback to libMesh
2. Enforce constraints on `current_local_solution`
3. Call the residual function or functor
    1. While computing the residual, make sure to call `constrain_element_vector` during assembly. This zeros the residual of the constrained dof (and modifies the residuals of the un-constrained dofs which determine the constrained dof)
4. `SNESComputeJacobian` gets called -> callback to libMesh
5. Enforce constraints on `current_local_solution`
6. Call the Jacobian function or functor
    1. While computing the Jacobian, make sure to call `constrain_element_matrix` during assembly. This puts a 1 on the diagonal for the constrained dofs and zeros elsewhere. In conjunction with 3i) this enforces delta_u_constrained_dof = 0
7. `SNESSolve` finishes and then we enforce the constraints exactly on the parallel solution vector

So this works great. The problems occur when we approximate the action of the Jacobian with a matrix-free method **or** any time the preconditioner/Jacobian is computed using finite differences. In that case if one is using the same function for computing the "nonlinear" residual and either the "linear" residual (for matrix-free) or columns of the Jacobian matrix/preconditioner (for a finite-differenced matrix), then it will appear to the linear operator as if the constrained dof depends on **nothing**. 

Consider this mesh:

![mesh](https://user-images.githubusercontent.com/10248304/50662625-4dec7880-0f64-11e9-8eb1-c1e6a9160dbc.png)

The resulting hand-coded Jacobian row for the hanging-node using `constrain_element_matrix` (note that the diagonal is doubled because two elements contribute to it):
row 6: (1, 0.)  (2, 0.)  (4, 0.)  (5, 0.)  (6, 2.)  (9, 0.) 

And the resulting finite-differenced Jacobian row for the hanging-node:
row 6: (1, 0.)  (2, 0.)  (4, 0.)  (5, 0.)  (6, 0.)  (9, 0.) 

So our operator is now singular because we zero this dof's residual every time we call our function! My suggestion to fix this:

- Don't enforce constraints exactly
- Form an actual residual for the "constrained" dof, e.g. .5 * u1 + .5 * u2 - u6 = 0
[...]

PR #1986, opened 2019-01-04
Different constraint enforcement for hanging nodes

lindsayad:
Throwing this up for a (likely disastrous) CIVET test...

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -2168,7 +2168,8 @@ void DofMap::enforce_constraints_exactly (const System & system,
 
 void DofMap::enforce_constraints_on_residual (const NonlinearImplicitSystem & system,
                                               NumericVector<Number> * rhs,
-                                              NumericVector<Number> const * solution) const
+                                              NumericVector<Number> const * solution,
+                                              bool homogeneous) const
 {
   parallel_object_only();
 
@@ -2212,6 +2213,13 @@ void DofMap::enforce_constraints_on_residual (const NonlinearImplicitSystem & sy
       for (const auto & j : constraint_row)
         exact_value += j.second * (*solution_local)(j.first);
       exact_value -= (*solution_local)(constrained_dof);
+      if (!homogeneous)
+        {
+          DofConstraintValueMap::const_iterator rhsit =
+            _primal_constraint_values.find(constrained_dof);
+          if (rhsit != _primal_constraint_values.end())
+            exact_value += rhsit->second;
+        }
 
       rhs->set(constrained_dof, exact_value);
     }
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -2180,6 +2180,24 @@ void DofMap::enforce_constraints_on_residual (const NonlinearImplicitSystem & sy
   if (!solution)
     solution = system.solution.get();
 
+  NumericVector<Number> const * solution_local  = nullptr; // will be initialized below
+  std::unique_ptr<NumericVector<Number>> solution_built;
+  if (solution->type() == SERIAL || solution->type() == GHOSTED)
+      solution_local = solution;
+  else if (solution->type() == PARALLEL)
+    {
+      solution_built = NumericVector<Number>::build(this->comm());
+      solution_built->init (solution->size(), solution->size(), true, SERIAL);
+      solution->localize(*solution_built);
+      solution_built->close();
+      solution_local = solution_built.get();
+    }
+  else // unknown solution->type()
+    libmesh_error_msg("ERROR: Unknown solution->type() == " << solution->type());
+
+  // We should never hit these asserts because we should error-out in
+  // else clause above.  Just to be sure we don't try to use solution_local
+  libmesh_assert(solution_local);
   libmesh_assert_equal_to (this, &(system.get_dof_map()));
 
   for (const auto & pr : _dof_constraints)
@@ -2192,8 +2210,8 @@ void DofMap::enforce_constraints_on_residual (const NonlinearImplicitSystem & sy
 
       Number exact_value = 0;
       for (const auto & j : constraint_row)
-        exact_value += j.second * (*solution)(j.first);
-      exact_value -= (*solution)(constrained_dof);
+        exact_value += j.second * (*solution_local)(j.first);
+      exact_value -= (*solution_local)(constrained_dof);
 
       rhs->set(constrained_dof, exact_value);
     }
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -45,6 +45,7 @@
 #include "libmesh/raw_accessor.h"
 #include "libmesh/sparse_matrix.h" // needed to constrain adjoint rhs
 #include "libmesh/system.h" // needed by enforce_constraints_exactly()
+#include "libmesh/nonlinear_implicit_system.h"
 #include "libmesh/threads.h"
 #include "libmesh/tensor_tools.h"
 #include "libmesh/int_range.h"
@@ -2165,6 +2166,65 @@ void DofMap::enforce_constraints_exactly (const System & system,
   v->close();
 }
 
+void DofMap::enforce_constraints_on_residual (const NonlinearImplicitSystem & system,
+                                              NumericVector<Number> * rhs,
+                                              NumericVector<Number> const * solution) const
+{
+  parallel_object_only();
+
+  if (!this->n_constrained_dofs())
+    return;
+
+  if (!rhs)
+    rhs = system.rhs;
+  if (!solution)
+    solution = system.solution.get();
+
+  libmesh_assert_equal_to (this, &(system.get_dof_map()));
+
+  for (const auto & pr : _dof_constraints)
+    {
+      dof_id_type constrained_dof = pr.first;
+      if (!this->local_index(constrained_dof))
+        continue;
+
+      const DofConstraintRow constraint_row = pr.second;
+
+      Number exact_value = 0;
+      for (const auto & j : constraint_row)
+        exact_value += j.second * (*solution)(j.first);
+      exact_value -= (*solution)(constrained_dof);
+
+      rhs->set(constrained_dof, exact_value);
+    }
+}
+
+void DofMap::enforce_constraints_on_jacobian (const NonlinearImplicitSystem & system,
+                                              SparseMatrix<Number> * jac) const
+{
+  parallel_object_only();
+
+  if (!this->n_constrained_dofs())
+    return;
+
+  if (!jac)
+    jac = system.matrix;
+
+  libmesh_assert_equal_to (this, &(system.get_dof_map()));
+
+  for (const auto & pr : _dof_constraints)
+    {
+      dof_id_type constrained_dof = pr.first;
+      if (!this->local_index(constrained_dof))
+        continue;
+
+      const DofConstraintRow constraint_row = pr.second;
+
+      for (const auto & j : constraint_row)
+        jac->set(constrained_dof, j.first, j.second);
+      jac->set(constrained_dof, constrained_dof, -1);
+    }
+}
 
 
 void DofMap::enforce_adjoint_constraints_exactly (NumericVector<Number> & v,
--- a/src/solvers/petsc_nonlinear_solver.C
+++ b/src/solvers/petsc_nonlinear_solver.C
@@ -181,7 +181,8 @@ extern "C"
     else
       libmesh_error_msg("Error! Unable to compute residual and/or Jacobian!");
 
-    rc.sys.get_dof_map().enforce_constraints_on_residual(rc.sys, &R, rc.sys.current_local_solution.get());
+    PetscVector<Number> X(x, rc.sys.comm());
+    rc.sys.get_dof_map().enforce_constraints_on_residual(rc.sys, &R, &X);
 
     R.close();
 
@@ -219,7 +220,8 @@ extern "C"
     else
       libmesh_error_msg("Error! Unable to compute residual for forming finite difference Jacobian!");
 
-    rc.sys.get_dof_map().enforce_constraints_on_residual(rc.sys, &R, rc.sys.current_local_solution.get());
+    PetscVector<Number> X(x, rc.sys.comm());
+    rc.sys.get_dof_map().enforce_constraints_on_residual(rc.sys, &R, &X);
 
     R.close();
 
@@ -259,7 +261,8 @@ extern "C"
       libmesh_error_msg("Error! Unable to compute residual for forming finite differenced"
                         "Jacobian-vector products!");
 
-    rc.sys.get_dof_map().enforce_constraints_on_residual(rc.sys, &R, rc.sys.current_local_solution.get());
+    PetscVector<Number> X(x, rc.sys.comm());
+    rc.sys.get_dof_map().enforce_constraints_on_residual(rc.sys, &R, &X);
 
     R.close();
 
--- a/src/solvers/petsc_nonlinear_solver.C
+++ b/src/solvers/petsc_nonlinear_solver.C
@@ -181,6 +181,8 @@ extern "C"
     else
       libmesh_error_msg("Error! Unable to compute residual and/or Jacobian!");
 
+    rc.sys.get_dof_map().enforce_constraints_on_residual(rc.sys, &R, rc.sys.current_local_solution.get());
+
     R.close();
 
     return rc.ierr;
@@ -216,6 +218,8 @@ extern "C"
 
     else
       libmesh_error_msg("Error! Unable to compute residual for forming finite difference Jacobian!");
+
+    rc.sys.get_dof_map().enforce_constraints_on_residual(rc.sys, &R, rc.sys.current_local_solution.get());
 
     R.close();
 
@@ -254,6 +258,8 @@ extern "C"
     else
       libmesh_error_msg("Error! Unable to compute residual for forming finite differenced"
                         "Jacobian-vector products!");
+
+    rc.sys.get_dof_map().enforce_constraints_on_residual(rc.sys, &R, rc.sys.current_local_solution.get());
 
     R.close();
 
@@ -366,6 +372,8 @@ extern "C"
     else
       libmesh_error_msg("Error! Unable to compute residual and/or Jacobian!");
 
+    sys.get_dof_map().enforce_constraints_on_jacobian(sys, &PC);
+
     PC.close();
     Jac.close();
 #if PETSC_RELEASE_LESS_THAN(3,5,0)
@@ -453,9 +461,8 @@ extern "C"
     // It's also possible that we don't need to do anything at all, in
     // that case return early...
     NonlinearImplicitSystem & sys = solver->system();
-    DofMap & dof_map = sys.get_dof_map();
 
-    if (!dof_map.n_constrained_dofs() && !solver->postcheck && !solver->postcheck_object)
+    if (!solver->postcheck && !solver->postcheck_object)
       return ierr;
 
     // We definitely need to wrap at least "w"
@@ -495,25 +502,6 @@ extern "C"
 
     if (changed_new_soln)
       *changed_w = PETSC_TRUE;
-
-    if (dof_map.n_constrained_dofs())
-      {
-        PetscVector<Number> & system_soln = *cast_ptr<PetscVector<Number> *>(sys.solution.get());
-
-        // ... and swap it in before enforcing the constraints.
-        petsc_w.swap(system_soln);
-
-        dof_map.enforce_constraints_exactly(sys);
-
-        // If we have constraints, we'll assume that we did change the
-        // solution w (hopefully slightly).  Enforcing constraints
-        // does not change the search direction, y, but the user may
-        // have, so we leave it alone.
-        *changed_w = PETSC_TRUE;
-
-        // Swap back
-        petsc_w.swap(system_soln);
-      }
 
     return ierr;
   }
@@ -940,12 +928,6 @@ PetscNonlinearSolver<T>::solve (SparseMatrix<T> &  pre_in,  // System Preconditi
 
   ierr = SNESGetLinearSolveIterations(_snes, &_n_linear_iterations);
   LIBMESH_CHKERR(ierr);
-
-  // Enforce constraints exactly now that the solve is done.  We have
-  // been enforcing them on the current_local_solution during the
-  // solve, but now need to be sure they are enforced on the parallel
-  // solution vector as well.
-  this->system().get_dof_map().enforce_constraints_exactly(this->system());
 
   // SNESGetFunction has been around forever and should work on all
   // versions of PETSc.  This is also now the recommended approach
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -68,6 +68,7 @@ class MeshBase;
 class PeriodicBoundaryBase;
 class PeriodicBoundaries;
 class System;
+class NonlinearImplicitSystem;
 template <typename T> class DenseVectorBase;
 template <typename T> class DenseVector;
 template <typename T> class DenseMatrix;
@@ -1155,6 +1156,12 @@ public:
    */
   void enforce_adjoint_constraints_exactly (NumericVector<Number> & v,
                                             unsigned int q) const;
+
+  void enforce_constraints_on_residual (const NonlinearImplicitSystem & system,
+                                        NumericVector<Number> * rhs,
+                                        NumericVector<Number> const * solution) const;
+  void enforce_constraints_on_jacobian (const NonlinearImplicitSystem & system,
+                                        SparseMatrix<Number> * jac) const;
 
 
 
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -1159,7 +1159,8 @@ public:
 
   void enforce_constraints_on_residual (const NonlinearImplicitSystem & system,
                                         NumericVector<Number> * rhs,
-                                        NumericVector<Number> const * solution) const;
+                                        NumericVector<Number> const * solution,
+                                        bool homogeneous = true) const;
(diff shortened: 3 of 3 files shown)


pbauman:
FYI, this is the failure during `make check` of libMesh (GRINS-dbg passed because it doesn't run `make check` for libMesh, but does mean the GRINS AMR tests (ReplicatedMesh) passed at least).

<pre>
* Running Example miscellaneous_ex3:
*   ./example-devel -r 3 -o FIRST 
 
Running /femputer/pbauman/civet_build_testing_root/libmesh-devel/examples/miscellaneous/miscellaneous_ex3/.libs/lt-example-devel -r 3 -o FIRST

 Mesh Information:
  elem_dimensions()={2}
  spatial_dimension()=2
  n_nodes()=225
    n_local_nodes()=225
  n_elem()=255
    n_local_elem()=255
    n_active_elem()=192
  n_subdomains()=1
  n_partitions()=1
  n_processors()=1
  n_threads()=1
  processor_id()=0

 EquationSystems
  n_systems()=1
   System #0, "Laplace-Young"
    Type "NonlinearImplicit"
    Variables="u" 
    Finite Element Types="LAGRANGE", "JACOBI_20_00" 
    Infinite Element Mapping="CARTESIAN" 
    Approximation Orders="FIRST", "THIRD" 
    n_dofs()=225
    n_local_dofs()=225
    n_constrained_dofs()=0
    n_local_constrained_dofs()=0
    n_vectors()=1
    n_matrices()=1
    DofMap Sparsity
      Average  On-Processor Bandwidth <= 8.11111
      Average Off-Processor Bandwidth <= 0
      Maximum  On-Processor Bandwidth <= 9
      Maximum Off-Processor Bandwidth <= 0
    DofMap Constraints
      Number of DoF Constraints = 0
      Number of Node Constraints = 0

  NL step  0, |residual|_2 = 2.000000e-01
[0]PETSC ERROR: Not for unassembled vector
[0]PETSC ERROR: See http://www.mcs.anl.gov/petsc/documentation/faq.html for trouble shooting.
[0]PETSC ERROR: Petsc Release Version 3.7.4, Oct, 02, 2016 
[0]PETSC ERROR: /femputer/pbauman/civet_build_testing_root/libmesh-devel/examples/miscellaneous/miscellaneous_ex3/.libs/lt-example-devel on a gcc-6.2.0-mpich-3.1.4-openblas-0.2.18-opt named femputer.eng.buffalo.edu by pbauman Fri Jan  4 13:22:59 2019
[0]PETSC ERROR: Configure options --with-make-np=32 --prefix=/femputer/pbauman/swstack/libs/petsc/3.7.4/gcc/6.2.0/mpich/3.1.4/openblas/0.2.18/opt --with-debugging=false --COPTFLAGS="-O3 -mavx" --CXXOPTFLAGS="-O3 -mavx" --FOPTFLAGS=-O3 --with-shared-libraries=1 --with-mpi-dir=/femputer/pbauman/swstack/libs/mpich/3.1.4/gcc/6.2.0 --with-mumps=true --download-mumps=1 --with-metis=true --download-metis=1 --with-parmetis=true --download-parmetis=1 --with-superlu=true --download-superlu=1 --with-superludir=true --download-superlu_dist=1 --with-blacs=true --download-blacs=1 --with-scalapack=true --download-scalapack=1 --with-hypre=true --download-hypre=1 --with-blas-lib="[/femputer/pbauman/swstack/libs/openblas/0.2.18/gcc/6.2.0/lib/libopenblas.so]" --with-lapack-lib="[/femputer/pbauman/swstack/libs/openblas/0.2.18/gcc/6.2.0/lib/libopenblas.so]" --LDFLAGS=
[0]PETSC ERROR: #1 VecCopy() line 1638 in /femputer/pbauman/swstack/builddir/petsc-2NPUKo/petsc-3.7.4/src/vec/vec/interface/vector.c
[0]PETSC ERROR: #2 KSPInitialResidual() line 65 in /femputer/pbauman/swstack/builddir/petsc-2NPUKo/petsc-3.7.4/src/ksp/ksp/interface/itres.c
[0]PETSC ERROR: #3 KSPSolve_GMRES() line 239 in /femputer/pbauman/swstack/builddir/petsc-2NPUKo/petsc-3.7.4/src/ksp/ksp/impls/gmres/gmres.c
[0]PETSC ERROR: #4 KSPSolve() line 656 in /femputer/pbauman/swstack/builddir/petsc-2NPUKo/petsc-3.7.4/src/ksp/ksp/interface/itfunc.c
[0]PETSC ERROR: #5 SNESSolve_NEWTONLS() line 230 in /femputer/pbauman/swstack/builddir/petsc-2NPUKo/petsc-3.7.4/src/snes/impls/ls/ls.c
[0]PETSC ERROR: #6 SNESSolve() line 4005 in /femputer/pbauman/swstack/builddir/petsc-2NPUKo/petsc-3.7.4/src/snes/interface/snes.c

 ----------------------------------------------------------------------------------------------------
| Time:           Fri Jan  4 13:23:03 2019                                                           |
| OS:             Linux                                                                              |
| HostName:       femputer.eng.buffalo.edu                                                           |
| OS Release:     2.6.32-754.3.5.el6.x86_64                                                          |
| OS Version:     #1 SMP Tue Aug 14 20:46:41 UTC 2018                                                |
| Machine:        x86_64                                                                             |
| Username:       pbauman                                                                            |
| Configuration:  /femputer/pbauman/civet_build_testing_root/libMesh/configure  '--enable-everything'|
|  '--with-metis=PETSc'                                                                              |
|  '--with-cppunit-prefix=/femputer/pbauman/swstack/libs/cppunit/1.13.2/gcc/6.2.0'                   |
|  '--with-vtk-lib=/femputer/pbauman/swstack/libs/vtk/7.1.0/gcc/6.2.0/mpich/3.1.4/lib'               |
|  '--with-vtk-include=/femputer/pbauman/swstack/libs/vtk/7.1.0/gcc/6.2.0/mpich/3.1.4/include/vtk-7.1'|
|  '--with-methods=devel'                                                                            |
|  '--prefix=/femputer/pbauman/civet_build_testing_root/libmesh-devel-install'                       |
|  '--enable-hdf5'                                                                                   |
|  '--disable-glibcxx-debugging'                                                                     |
|  '--enable-werror'                                                                                 |
|  '--enable-paranoid-warnings'                                                                      |
|  'CXX=g++'                                                                                         |
|  'CC=gcc'                                                                                          |
|  'FC=gfortran'                                                                                     |
|  'F77=gfortran'                                                                                    |
|  'METHODS=devel'                                                                                   |
|  'PETSC_DIR=/femputer/pbauman/swstack/libs/petsc/3.7.4/gcc/6.2.0/mpich/3.1.4/openblas/0.2.18/opt'  |
|  'PETSC_ARCH='                                                                                     |
|  'VTK_INCLUDE=/femputer/pbauman/swstack/libs/vtk/7.1.0/gcc/6.2.0/mpich/3.1.4/include/vtk-7.1'      |
|  'VTK_DIR=/femputer/pbauman/swstack/libs/vtk/7.1.0/gcc/6.2.0/mpich/3.1.4'                          |
|  'HDF5_DIR=/femputer/pbauman/swstack/libs/hdf5/1.8.16/gcc/6.2.0'                                   |
 ----------------------------------------------------------------------------------------------------
 -----------------------------------------------------------------------------------------------------------------
| libMesh Performance: Alive time=4.29426, Active time=0.022451                                                   |
 -----------------------------------------------------------------------------------------------------------------
| Event                              nCalls     Total Time  Avg Time    Total Time  Avg Time    % of Active Time  |
|                                               w/o Sub     w/o Sub     With Sub    With Sub    w/o S    With S   |
|-----------------------------------------------------------------------------------------------------------------|
|                                                                                                                 |
|                                                                                                                 |
| DefaultCoupling                                                                                                 |
|   operator()                       194        0.0004      0.000002    0.0004      0.000002    1.86     1.86     |
|                                                                                                                 |
| DofMap                                                                                                          |
|   add_neighbors_to_send_list()     1          0.0000      0.000043    0.0003      0.000339    0.19     1.51     |
|   build_sparsity()                 1          0.0004      0.000392    0.0007      0.000729    1.75     3.25     |
|   create_dof_constraints()         1          0.0001      0.000126    0.0001      0.000126    0.56     0.56     |
|   distribute_dofs()                1          0.0002      0.000241    0.0010      0.001009    1.07     4.49     |
|   dof_indices()                    768        0.0004      0.000000    0.0004      0.000000    1.63     1.63     |
|   prepare_send_list()              2          0.0000      0.000000    0.0000      0.000000    0.00     0.00     |
|   reinit()                         1          0.0004      0.000428    0.0004      0.000428    1.91     1.91     |
|                                                                                                                 |
| FE                                                                                                              |
|   compute_shape_functions()        448        0.0005      0.000001    0.0005      0.000001    2.33     2.33     |
|   init_shape_functions()           66         0.0000      0.000000    0.0000      0.000000    0.14     0.14     |
|   inverse_map()                    192        0.0001      0.000001    0.0001      0.000001    0.64     0.64     |
|                                                                                                                 |
| FEMap                                                                                                           |
|   compute_affine_map()             448        0.0002      0.000000    0.0002      0.000000    0.72     0.72     |
|   compute_face_map()               64         0.0001      0.000002    0.0002      0.000004    0.45     1.11     |
|   init_face_shape_functions()      1          0.0000      0.000002    0.0000      0.000002    0.01     0.01     |
|   init_reference_to_physical_map() 66         0.0001      0.000001    0.0001      0.000001    0.32     0.32     |
|                                                                                                                 |
| Mesh                                                                                                            |
|   find_neighbors()                 2          0.0004      0.000221    0.0004      0.000221    1.97     1.97     |
|   renumber_nodes_and_elem()        4          0.0000      0.000006    0.0000      0.000006    0.11     0.11     |
|                                                                                                                 |
| MeshBase                                                                                                        |
|   prepare_for_use()                2          0.0001      0.000051    0.0006      0.000296    0.46     2.64     |
|                                                                                                                 |
| MeshRefinement                                                                                                  |
|   _refine_elements()               3          0.0002      0.000066    0.0007      0.000245    0.88     3.27     |
|   add_node()                       1008       0.0003      0.000000    0.0003      0.000000    1.45     1.45     |
|                                                                                                                 |
| Parallel                                                                                                        |
|   allgather()                      1          0.0000      0.000001    0.0000      0.000001    0.00     0.00     |
|                                                                                                                 |
| Partitioner                                                                                                     |
|   set_node_processor_ids()         4          0.0003      0.000068    0.0003      0.000068    1.22     1.22     |
|   single_partition_range()         2          0.0000      0.000011    0.0000      0.000011    0.10     0.10     |
|                                                                                                                 |
| PetscNonlinearSolver                                                                                            |
|   jacobian()                       1          0.0006      0.000629    0.0010      0.001046    2.80     4.66     |
|   residual()                       1          0.0001      0.000090    0.0001      0.000090    0.40     0.40     |
|   solve()                          1          0.0008      0.000813    0.0000      0.000000    3.62     0.00     |
|                                                                                                                 |
| System                                                                                                          |
|   solve()                          1          0.0160      0.016043    0.0000      0.000000    71.46    0.00     |
|                                                                                                                 |
| TopologyMap                                                                                                     |
|   init()                           3          0.0001      0.000019    0.0001      0.000019    0.26     0.26     |
|                                                                                                                 |
| XdrIO                                                                                                           |
|   read()                           1          0.0003      0.000350    0.0005      0.000487    1.56     2.17     |
|   read_header()                    1          0.0000      0.000027    0.0000      0.000027    0.12     0.12     |
 -----------------------------------------------------------------------------------------------------------------
| Totals:                            3289       0.0225                                          100.00            |
 -----------------------------------------------------------------------------------------------------------------

application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0
FAIL: run.sh
====================================================
1 of 1 test failed
Please report to libmesh-users@lists.sourceforge.net
====================================================
make[4]: *** [check-TESTS] Error 1
make[4]: Leaving directory `/femputer/pbauman/civet_build_testing_root/libmesh-devel/examples/miscellaneous/miscellaneous_ex3'
make[3]: *** [check-am] Error 2
make[3]: Leaving directory `/femputer/pbauman/civet_build_testing_root/libmesh-devel/examples/miscellaneous/miscellaneous_ex3'
make[2]: *** [check] Error 2
make[2]: Leaving directory `/femputer/pbauman/civet_build_testing_root/libmesh-devel/examples/miscellaneous/miscellaneous_ex3'
make[1]: *** [check-recursive] Error 1
make[1]: Leaving directory `/femputer/pbauman/civet_build_testing_root/libmesh-devel/examples'
make: *** [check-recursive] Error 1
cat: test/test-suite.log: No such file or directory
ERROR: exiting with code 1
</pre>

lindsayad:
@pbauman Is GRINS testing down today?

src/base/dof_map_constraints.C
```diff
@@ -2165,6 +2166,65 @@ void DofMap::enforce_constraints_exactly (const System & system,
...
+    {
+      dof_id_type constrained_dof = pr.first;
+      if (!this->local_index(constrained_dof))
+        continue;
+
+      const DofConstraintRow constraint_row = pr.second;
+
+      for (const auto & j : constraint_row)
+        jac->set(constrained_dof, j.first, j.second);
+      jac->set(constrained_dof, constrained_dof, -1);
```
roystgnr: This is going to force the equivalent of the asymmetric_constraint_rows option and it's going to prevent constrained problem Jacobian matrices from being SPD?  Is there any way around that?

src/base/dof_map_constraints.C
```diff
@@ -2165,6 +2166,65 @@ void DofMap::enforce_constraints_exactly (const System & system,
...
+      dof_id_type constrained_dof = pr.first;
+      if (!this->local_index(constrained_dof))
+        continue;
+
+      const DofConstraintRow constraint_row = pr.second;
+
+      Number exact_value = 0;
+      for (const auto & j : constraint_row)
+        exact_value += j.second * (*solution)(j.first);
+      exact_value -= (*solution)(constrained_dof);
```
roystgnr: No _primal_constraint_values here?  Supporting only homogeneous constraints would be a step up from supporting nothing, but we'd at least want to scream and die if anyone ever tried to use this with heterogeneous constraints, not silently return wrong results.

src/base/dof_map_constraints.C
lindsayad: The Jacobian has to follow from the form of the residual statement, otherwise it will just be wrong. We'd need to think up our residual such that it naturally produces symmetric dependence on the dofs, e.g. if our residual function for the constrained dof depends on a couple of unconstrained dofs, then the residual of those unconstrained dofs must depend on the constrained dof...which would be a paradox...right?

src/base/dof_map_constraints.C
lindsayad: I'm not too knowledgeable about the "primal constraint" terminology. Would the primal constraints be the b vector in `u = C x + b` (using your syntax from [here](https://github.com/libMesh/libmesh/issues/1985#issuecomment-451596272))

pbauman:
> @pbauman Is GRINS testing down today?

FFS. Sorry. That's the third time in four days I've had to reboot it. Hopefully that can get resolved, but for the moment at least it's up and running again.

src/base/dof_map_constraints.C
lindsayad: @roystgnr ?

src/base/dof_map_constraints.C
roystgnr: Well, the way we get a symmetric "jacobian" in most codes is to not let the solver worry about constrained dofs at all.  Instead of putting the constraint equation in the row for u_c, we put "u_c=0" in that row, and the Jacobian is just a 1 on the diagonal.  Then after a solve we enforce_constraints_exactly() to get the real value of each u_c correct.  The correct value isn't needed during a linear solve, because the constraint matrix pre and post multiplication zeros out the constrained dof's column, not just it's row.

This works well for most Newton-Krylov solves because even though we haven't tried to optimize enforce_constraints_exactly we only bother to call it after each linear solve step and the cost is trivial compared to that.  I'm not sure how well it would work for JFNK or for finite differenced Jacobians.

src/base/dof_map_constraints.C
roystgnr: Sorry, missed that comment the first time.  Yes, _primal_constraint_values would be the sparse b in that equation.  (well, 99.9% sure it's b; otherwise I've misremembered the sign convention and it's -b)

src/base/dof_map_constraints.C
lindsayad: > Well, the way we get a symmetric "jacobian" in most codes is to not let the solver worry about constrained dofs at all. Instead of putting the constraint equation in the row for u_c, we put "u_c=0" in that row, and the Jacobian is just a 1 on the diagonal. Then after a solve we enforce_constraints_exactly() to get the real value of each u_c correct. The correct value isn't needed during a linear solve, because the constraint matrix pre and post multiplication zeros out the constrained dof's column, not just it's row.
> 
> This works well for most Newton-Krylov solves

Yes I understand the idea and for many cases this works great. But the discussion in #1985 (hopefully now) illustrates that this doesn't work for JFNK or finite differenced jacobians.

src/base/dof_map_constraints.C
lindsayad: Ok @roystgnr I added possible contributions from `_primal_constraint_values`

lindsayad:
@roystgnr I think this is ready for re-review. If the only failure is the four tests in MOOSE modules, then I would advocate for a merge, and I will fix those tests in a shortly to follow libmesh update to MOOSE. Although I do want to make sure I get a PR merged into MOOSE that allows #1946 to go in...

lindsayad:
That's fair. Will do

lindsayad:
We've got the MOOSE tests fixed up

lindsayad:
I'm hoping that since the `GRINS-devel` test made it to cleanup that that is basically a passed test...

roystgnr:
Yeah, femputer is just down again, and it definitely made it far enough for us to trust a merge.  Pinging @pbauman so he can yell at UB IT again for the benefit of future PR updates.

src/base/dof_map_constraints.C
```diff
@@ -2180,6 +2180,24 @@ void DofMap::enforce_constraints_on_residual (const NonlinearImplicitSystem & sy
   if (!solution)
     solution = system.solution.get();
 
+  NumericVector<Number> const * solution_local  = nullptr; // will be initialized below
+  std::unique_ptr<NumericVector<Number>> solution_built;
+  if (solution->type() == SERIAL || solution->type() == GHOSTED)
+      solution_local = solution;
+  else if (solution->type() == PARALLEL)
```
friedmud: Wait.  What's going on here @lindsayad ?  We definitely don't want to serialize solution vectors for constraint enforcement!  This will be a huge scaling problem.

src/solvers/petsc_nonlinear_solver.C
```diff
@@ -181,7 +181,8 @@ extern "C"
     else
       libmesh_error_msg("Error! Unable to compute residual and/or Jacobian!");
 
-    rc.sys.get_dof_map().enforce_constraints_on_residual(rc.sys, &R, rc.sys.current_local_solution.get());
+    PetscVector<Number> X(x, rc.sys.comm());
```
friedmud: Wait.  @lindsayad are we now making a copy of this... and then killing it each iteration?

friedmud:
@lindsayad it may be possible that something about this is now causing the adaptivity input file in step 7 of the MOOSE tutorial to fail.  (This has been reported by @aeslaughter )

I'm having a hard time tracing back to before this PR (too much has changed)... but the issue very much looks like the preconditioner is not liking something about the way constraints are being enforced.  The solve works with LU... but not with AMG...

src/base/dof_map_constraints.C
lindsayad: Well if it's a huge problem now, then it was also a huge problem before. See [here](https://github.com/libMesh/libmesh/pull/1986/files#diff-b3985fa832b099b796eaea1581eb1995L948)

src/solvers/petsc_nonlinear_solver.C
lindsayad: This does not copy the PETSc Vec object. The doxygen for this constructor is [here](https://libmesh.github.io/doxygen/classlibMesh_1_1PetscVector.html#a237268dc452928c69bb5b8485e0f4d7b)

lindsayad:
> it may be possible that something about this is now causing the adaptivity input file in step 7 of the MOOSE tutorial to fail.

<s>Specifically what is failing? We test the tutorials on every push to the next branch.</s> Ah I see it only tests one time step.

> I'm having a hard time tracing back to before this PR (too much has changed)... but the issue very much looks like the preconditioner is not liking something about the way constraints are being enforced. The solve works with LU... but not with AMG...

Now this I could believe. See @roystgnr comment [here](https://github.com/libMesh/libmesh/pull/1986/#discussion_r245788699). Major sigh...I can't believe we didn't hit anything like this in our entire test suite!!!

I wish I could remember the exact genesis for #1985 and this PR. One issue for sure is that I was getting "false positives" for a bad Jacobian when trying to verify AD on my adaptive laser welding problem using `-snes_test_jacobian` because the finite differenced Jacobian would produce a zero row while our SMP matrix did not (e.g. we artificially introduced a 'c' on the diagonal for the constrained dof where 'c' is positive). 

I also could have sworn that I had developed a test case for PJFNK in which hanging nodes would make it impossible for the linear solve to converge because, I reasoned, just as a Jacobian matrix formed from finite differencing was singular using our previous method of constraint enforcement, so too must the approximation of the Jacobian-vector product using finite differences produce a singular action. However, if I go into the `moose/test/tests/kernels/diffusion_with_hanging_node` test suite that I introduced after this PR, and I run the `test_mf` test with our old constraint enforcement, it runs perfectly fine! Moreover, it runs fine even without a preconditioner! Additionally, the fact that the tutorial adaptivity case ran beautifully is a further testament to the fact that the previous constraint enforcement appears not to introduce issues for matrix-free.

So in summary: It really bothers me to have a residual function that, when finite differenced to produce a Jacobian matrix, produces a singular matrix. This makes it so that FDP is actually a terrible choice to debug your residual statements as long as you have a hanging node or periodic constraint. The same issue goes for trying to debug hand-coded Jacobians when those constraints are present. However, it is probably not a very common occurrence that users are debugging residuals/Jacobians when hanging node/periodic constraints are present.

The benefit of being able to maintain an SPD matrix and thus use preconditioners like amg probably outweighs the concerns of mine in the above paragraph. Although it is shocking that we have no tests that cover the case of adaptivity and iterative algorithms that need an SPD matrix and perhaps indicative that that ***also*** is a rare case, it would be nice to use all the best technology we have to offer in one simulation: adaptivity, multigrid, etc.

So in second summary: I could be convinced to revert this. Maybe @fdkong also wants to weigh in though because he was also in favor of this change.

aeslaughter:
I need a recommendation for the workshop which starts on the 17th, because as it is now when we turn on adaptivity the problem doesn't converged, which doesn't look so good for us. I could switch preconditioners, but that doesn't look great either.

friedmud:
I'm not convinced that this is about having a symmetric matrix for Hypre... Hypre certainly works fine without a symmetric matrix.

As for positive-definite... the new constraints should be positive definite as well... right?  It's a 1 on the diagonal and 0.5 is a couple of off-diagonal spots?

As for not seeing this in our test suite: that's not bothering me.  We don't run many large problems with adaptivity in our test suite.  BUT: users should have started to complain immediately.  I know that people doing phase-field are using adaptivity and hypre... and they haven't complained.  So I'm not sure that we have the whole story just yet.

I would like to verify that we can fix this by doing the old thing before we go too far...

src/solvers/petsc_nonlinear_solver.C
friedmud: right - I see now.

friedmud:
Ok - I have verified that the problem runs fine if I revert these changes using:

```
git co 79e6819d2db09780873cef92ebf0094caf805eca -- src/solvers/petsc_nonlinear_solver.C
```

I'm now working on narrowing down exactly where the issue is.

lindsayad:
I'm going to just try swapping the signs of the on-diagonal and off-diagonal...currently putting a negative on the diagonal! Boy that's idiotic

friedmud:
lol - yes... yes it is :-)

lindsayad:
Switching the signs doesn't make the adaptivity tutorial work

friedmud:
Still better to have them switched :-)

lindsayad:
I don't really know what's going on. The eigenvalues of the matrices for the old constraint method and the constraint method after #2155 don't look that different to me. I've attached them here. After #2155 the matrix has all positive eigenvalues.

[initial-refine-1-old-constraint.log](https://github.com/libMesh/libmesh/files/3263499/initial-refine-1-old-constraint.log)
[initial-refine-1-positive-constraint.log](https://github.com/libMesh/libmesh/files/3263500/initial-refine-1-positive-constraint.log)

lindsayad:
Attached are plots of the sparsity. Looking closely, you can see the new asymmetry, but overall the matrices look pretty damn similar.

[beautiful-positive-new-constraints.pdf](https://github.com/libMesh/libmesh/files/3263597/beautiful-positive-new-constraints.pdf)
[beautiful-old-constraint.pdf](https://github.com/libMesh/libmesh/files/3263610/beautiful-old-constraint.pdf)

(Beautiful) names are courtesy of @fdkong

src/base/dof_map_constraints.C
fdkong: I see, @friedmud already knew the problem, but the code still got in!

lindsayad:
For anyone who ever comes on this discussion, this particular poor solve was caused by poor multiphysics scaling which led to a poor condition number. Doing proper scaling allowed the solve to converge (as discussed in https://github.com/idaholab/moose/pull/13527)

roystgnr:
We need to enforce code commenting on PRs, not just code correctness.  @vikramvgarg stumbled across `enforce_constraints_on_residual` yesterday (while adding post-projection constraint enforcement for transient adjoint problems) and was confused as to why we had a method to do something that (in the code paths he's used) we never need to do.

PR #1987, opened 2019-01-10
Pass original FEType object to shape()/shape_deriv() calls.

jwpeterson:
Otherwise the p-level will be bumped twice as the shape() and
shape_deriv() calls bump it internally.

### DIFF
--- a/src/fe/fe_interface.C
+++ b/src/fe/fe_interface.C
@@ -920,11 +920,16 @@ void FEInterface::compute_data(const unsigned int dim,
 
 #endif
 
-  FEType p_refined = fe_t;
-  p_refined.order = static_cast<Order>(p_refined.order + elem->p_level());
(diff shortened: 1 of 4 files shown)


roystgnr:
Yikes, good catch!

Is this an ancient problem?  It doesn't look like it was introduced in #1981 and I can't recall anything else touching the same code recently.

I guess nothing triggered this because we have p refinement in some examples but those don't use compute_data?

jwpeterson:
I will also add a unit test for this case so that it hopefully continues to work...

jwpeterson:
Hmmm... looks like the new unit test might not run correctly in parallel? I will see if I can reproduce/fix that.

jwpeterson:
In hindsight, obviously looping over all mesh nodes was a bad idea in parallel as a bunch of 
```
Assertion `_out_of_mesh_mode' failed.
```
were being thrown.

PR #1988, opened 2019-01-11
Update cppunit m4 test

jwpeterson:
The old test relied on a utility called cppunit-config which has apparently been removed from upstream for some time and is finally no longer a part of the cppunit package on Ubuntu 18. This PR changes the configure test to simply try and compile/link a simple test program and enable/disable cppunit based on the results of that.

### DIFF
--- a/configure
+++ b/configure
@@ -1260,8 +1260,8 @@ enable_fparser
 with_fparser
 enable_fparser_debugging
 enable_cppunit
-with_cppunit_prefix
-with_cppunit_exec_prefix
+with_cppunit_include
+with_cppunit_lib
 enable_nanoflann
 enable_metaphysicl
 enable_metaphysicl_required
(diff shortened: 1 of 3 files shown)


jwpeterson:
Hmm... cppunit is not being detected on the MOOSE build boxes, 
```
checking whether we can build a trivial CppUnit program... no
```
probably because they are using a slightly older custom-built version
```
  7) moose/.cppunit-1.12.1_gcc-7.3.1
```
which still comes with cppunit-config. I should probably still try and use the old `cppunit-config` if it's available, I'll see if I can fix that.

jwpeterson:
OK, seems like that issue is now fixed:
```
checking for cppunit-config... cppunit-config
checking whether we can build a trivial CppUnit program... yes
```

PR #1995, opened 2019-01-16
Using nanoflann in replicated_mesh::stitching_helper.

leorachapuis:
Use the nanoflann library and kd-trees to improve the stitching algorithm (the part with the nearest neighbor search).

### DIFF
--- a/src/mesh/replicated_mesh.C
+++ b/src/mesh/replicated_mesh.C
@@ -24,6 +24,7 @@
 #include "libmesh/metis_partitioner.h"
 #include "libmesh/replicated_mesh.h"
 #include "libmesh/utility.h"
+#include "libmesh/nanoflann.hpp"
 
 // C++ includes
 #include <unordered_map>
@@ -109,6 +110,48 @@ public:
 };
 }
 
+// This class adapts a vector of Nodes (represented by a pair of a Point and a dof_id_type)
+// for use in a nanoflann KD-Tree
+
+class VectorOfNodesAdaptor
+{
+private:
+  const std::vector<std::pair<Point, dof_id_type>> _nodes;
+
+public:
+  VectorOfNodesAdaptor(const std::vector<std::pair<Point, dof_id_type>> & nodes) :
+    _nodes(nodes)
+  {}
+
+  /**
+   * Must return the number of data points
+   */
+  inline size_t kdtree_get_point_count() const { return _nodes.size(); }
+
+  /**
+   * \returns The dim'th component of the idx'th point in the class:
+   * Since this is inlined and the "dim" argument is typically an immediate value, the
+   *  "if's" are actually solved at compile time.
+   */
+  inline Real kdtree_get_pt(const size_t idx, int dim) const
+    {
+      libmesh_assert_less (idx, _nodes.size());
+      libmesh_assert_less (dim, 3);
+
+      const Point & p(_nodes[idx].first);
+
+      if (dim==0) return p(0);
+      if (dim==1) return p(1);
+      return p(2);
+    }
+
+  /*
+   * Optional bounding-box computation
+   */
+  template <class BBOX>
+  bool kdtree_get_bbox(BBOX & /* bb */) const { return false; }
+
+};
 
 
 namespace libMesh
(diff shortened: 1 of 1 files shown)


jwpeterson:
@leorachapuis I sent you an invite to join the "Associates" team so that your libMesh PRs will be tested automatically.

leorachapuis:
@jwpeterson Thanks !

roystgnr:
We have a --disable-nanoflann option, which this PR will break, right?

My preferred solution would be to leave the old algorithm here as a fallback in case we (or worse, a user on a weird system) discover some bug with nanoflann on their environment or their problem.

On the other hand, we've been enabling nanoflann by default for a long time without any reported compilation bugs, and it does get *some* usage so there probably aren't many runtime bugs either, so I could perhaps be talked into "just remove the --disable-nanoflann option" instead.

jwpeterson:
The build error in the complex/infinite elements configuration looks to be due to paranoid warnings as errors rather anything having to do with that particular configuration...would be good to fix this before merge, though, just by renaming the variable.
```
../src/mesh/replicated_mesh.C:1116:21: error: declaration of ‘current_node’ shadows a previous local [-Werror=shadow]
           for (auto current_node : other_boundary_node_ids)
                     ^~~~~~~~~~~~
../src/mesh/replicated_mesh.C:1099:43: note: shadowed declaration is here
           std::set<dof_id_type>::iterator current_node = this_boundary_node_ids.begin();
                                           ^~~~~~~~~~~~
```

leorachapuis:
@jwpeterson Ok, I can fix that.

jwpeterson:
@roystgnr there is still an "old" N^2 algorithm there as a fallback case, which we could use for `--disable-nanoflann` builds, but I agree that this isn't hugely important. 

The algorithm that got removed was the "binary search" optimization which according to @dknez didn't work very well anyway for most of their real world problems, so I am 100% in favor of removing it. It looks like @leorachapuis commented it out, but it should just be deleted.

jwpeterson:
In that same vein, I don't believe the `FuzzyPointCompare` object was being used by anything else so that should just be removed. All together it should be a nice net reduction in lines of code...

leorachapuis:
Ok, so I will just delete the old "binary search" I commented out, as well as the "FuzzyPointCompare" object. Does that sound good ?

jwpeterson:
> Does that sound good ?

Yep sounds great.

src/mesh/replicated_mesh.C
```diff
@@ -24,95 +24,62 @@
...
-  }
 };
-}
 
 
-
-namespace libMesh
-{
+//namespace libMesh
+//{
```
jwpeterson: Delete these two commented out lines since the namespace declaration has been moved up.

src/mesh/replicated_mesh.C
```diff
@@ -971,77 +938,41 @@ void ReplicatedMesh::stitching_helper (const ReplicatedMesh * other_mesh,
...
-              // Not every node on this_sorted_bndry_nodes will necessarily be stitched, so
-              // if its pair is not found on other_mesh, just continue.
-              if (other_iter != other_sorted_bndry_nodes.end())
-                {
-                  // Check that the points do indeed match - should not be necessary unless something
-                  // is wrong with binary_find.  To be on the safe side, we'll check.
-                  {
-                    // Grab the other point from the iterator
-                    Point other_point = other_iter->first;
+          // Storage for neerest neighbor in the loop below
```
jwpeterson: Typo: neerest -> nearest

src/mesh/replicated_mesh.C
```diff
@@ -971,77 +938,41 @@ void ReplicatedMesh::stitching_helper (const ReplicatedMesh * other_mesh,
...
+          for (auto node : other_boundary_node_ids)
+          {
+            const Real query_pt[] = {other_mesh->point(node)(0), other_mesh->point(node)(1), other_mesh->point(node)(2)};
+            this_kd_tree.knnSearch(&query_pt[0], 1, &ret_index[0], &ret_dist_sqr[0]);
+            node_to_node_map[this_mesh_nodes[ret_index[0]].second] = node;
+            other_to_this_node_map[node] = this_mesh_nodes[ret_index[0]].second;
+          }
 
+          // If the 2 maps don't have the same size, it means we have overwritten a value in node_to_node_map
+          // It means one node in this mesh is the neerest neighbor of several nodes in other mesh.
```
jwpeterson: neerest -> nearest

roystgnr:
> @roystgnr there is still an "old" N^2 algorithm there as a fallback case, which we could use for --disable-nanoflann builds, but I agree that this isn't hugely important.

Using the old algorithm as the fallback sounds fine to me, but could we add the LIBMESH_HAVE_NANOFLANN guards now rather than later?  I wouldn't even mind if the fallback was libmesh_not_implemented(), honestly, just as long as it wasn't a compilation error for anyone hitting a nanoflann incompatibility.

jwpeterson:
> could we add the LIBMESH_HAVE_NANOFLANN guards now rather than later?

Uh, sure. If @leorachapuis wants to take a crack at that great, otherwise I can probably pull this branch later tonight and do that.

leorachapuis:
@roystgnr I can do it but can you tell me where to put them ? I've never used such things before...

leorachapuis:
> > could we add the LIBMESH_HAVE_NANOFLANN guards now rather than later?
> 
> Uh, sure. If @leorachapuis wants to take a crack at that great, otherwise I can probably pull this branch later tonight and do that.

@leorachapuis As I'm not sure of what to do, it would probably be better if you could do it

jwpeterson:
I've attached the "Test No Optional" CIVET test. That should hopefully disable nanoflann and test whether the ifdef guards I added are working.

lindsayad:
So this causes a lot of failures in BISON (see idaholab/moose#12796). @friedmud has indicated he can look into this. lindsayad/libmesh@f9ff76e is a start on fixing this. It cuts the failures down from 50 (well actually MAX FAILURES) to 23

PR #1997, opened 2019-01-16
Add DualNumber to libMesh::CompareTypes

lindsayad:
Prevent some potential future include order dependency issues (in MOOSE).

### DIFF
--- a/include/utils/compare_types.h
+++ b/include/utils/compare_types.h
@@ -177,4 +177,70 @@ SUPERTYPE(double, long double);
 
 } // namespace libMesh
 
+#ifdef LIBMESH_HAVE_METAPHYSICL
+
+namespace MetaPhysicL
+{
+template <typename, typename>
+class DualNumber;
+} // namespace MetaPhysicL
+
+namespace libMesh
+{
+template <typename T, typename T2, typename D>
+struct CompareTypes<T, MetaPhysicL::DualNumber<T2, D>>
+{
+  typedef MetaPhysicL::DualNumber<typename CompareTypes<T, T2>::supertype,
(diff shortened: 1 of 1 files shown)


roystgnr:
Looks good to me, once the redundancy with current MOOSE gets worked out.

lindsayad:
Yea, I don't quite know what the best way forward on that is if the goal is to have no period of breakage. I could do a `#ifndef` or something in MOOSE, get that merged, and then add a corresponding `#define` here in this PR. Then once the libmesh hash is updated in MOOSE, I can completely remove the MOOSE version...and then come back and remove the `#define` in libmesh.

Note that I do have a [libmesh update PR into moose](https://github.com/idaholab/moose/pull/12715) right now...

roystgnr:
> I could do a #ifndef or something in MOOSE, get that merged, and then add a corresponding #define here in this PR. Then once the libmesh hash is updated in MOOSE, I can completely remove the MOOSE version...and then come back and remove the #define in libmesh.

That's clever and doesn't sound like too too much hassle.

include/utils/compare_types.h
```diff
@@ -177,4 +177,71 @@ SUPERTYPE(double, long double);
 
 } // namespace libMesh
 
+#ifdef LIBMESH_HAVE_METAPHYSICL
+#define LIBMESH_DUAL_NUMBER_COMPARE_TYPES
```
lindsayad: I probably should have a `HAVE` in there, but the MOOSE PR is already in...and this should be temporary

lindsayad:
@roystgnr ?

PR #2005, opened 2019-01-20
Avoid duplicating LAGRANGE and L2_LAGRANGE implementations

### LINKED ISSUE #1999: Improve L2_LAGRANGE implementation
I'm not sure when I'll have time to work on this, but it should be fairly simple to make this a whole lot better. The L2-Lagrange basis functions were originally added in 418a8b4d54e by simply copy/pasting the standard Lagrange implementation, but we have apparently not maintained both versions over time. For instance when adding some second derivatives in 9796b53d86, I only added them on the Lagrange side. Judging by line counts, there is actually quite a bit missing on the L2_LAGRANGE side (fe_lagrange_shape_3D.C is now 3661 lines, fe_l2_lagrange_shape_3D.C is only 1463). In general, there should be one "shared" implementation of the basis functions that both LAGRANGE and L2_LAGRANGE can call so we can avoid this issue in the future.

jwpeterson:
I will rebase and clean this up later if everything passes.

Closes #1999.


jwpeterson:
We have the same duplication in fe_hierarchic_shape_* and fe_l2_hierarchic_shape_* so I will fix those as well...

PR #2006, opened 2019-01-21
Combine HIERARCHIC, L2_HIERARCHIC implementations

jwpeterson:
Avoid lots of code duplication as was done in #2005 for Lagrange FEs.


roystgnr:
> +718 −3,323

> +677 −2,189

Awesome!

PR #2010, opened 2019-01-29
ExodusII_IO: Add ability to write more general discontinuous data

jwpeterson:
This new API, `ExodusII_IO::write_discontinuous_nodal_data_as_elemental()`, is similar to `ExodusII_IO::write_element_data()`, but instead of only writing (CONSTANT, MONOMIAL) data, it can be used to write general discontinuous solution fields, e.g. (FIRST, L2_LAGRANGE) and (SECOND, MONOMIAL) as a number of elemental data fields matching the number of vertices of a given element.

### DIFF
--- a/include/mesh/exodusII_io.h
+++ b/include/mesh/exodusII_io.h
@@ -196,12 +196,9 @@ public:
    *
    * This may be useful if you have a viz tool which is capable of
    * interpreting this element data as a discontinuous solution field.
-   * This function should generally _not_ be used for writing out
-   * (CONSTANT, MONOMIAL) data as it is always going to be overkill
-   * for that purpose.  (Imagine you have a CONSTANT, MONOMIAL
-   * variable on a Hex8 element, you would get 8 copies of a single
-   * value corresponding to each of the element's nodes, probably not
-   * what you want.)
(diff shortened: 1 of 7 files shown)


jwpeterson:
```
transfers/multiapp_nearest_node_transfer.fromsub: Error! The Exodus file already contains the variables:
transfers/multiapp_nearest_node_transfer.fromsub: And you asked to write:
transfers/multiapp_nearest_node_transfer.fromsub: Cannot overwrite existing variables in Exodus II file.
```
Ha, I was wondering if changing that check would trip anything. I'm guessing that nothing is printed to the screen because the error message is going to stdout instead of stderr? Also, seems like this doesn't have anything to do with 64-bit builds...

jwpeterson:
I think I figured out the problem. Since variable names can only be 32 characters long in Exodus file, the check fails the second time we try writing because the second name has been truncated:
```
Error! The Exodus file already contains the variables:
nodal_source_from_sub_elemental
elemental_source_from_sub_elemen

And you asked to write:
nodal_source_from_sub_elemental
elemental_source_from_sub_elemental
```
I'll fix the error check to only look at the first 32 characters...

PR #2018, opened 2019-01-31
Don't try to stitch things when you shouldn't

lindsayad:
We ran into errors with BISON for a couple reasons:

- trying to access into this_mesh_nodes when it was empty
- stitching nodes that weren't actually at the same point in space

Refs #1995

### DIFF
--- a/src/mesh/replicated_mesh.C
+++ b/src/mesh/replicated_mesh.C
@@ -946,7 +946,9 @@ void ReplicatedMesh::stitching_helper (const ReplicatedMesh * other_mesh,
 #endif
         }
 
-      if (use_binary_search)
+      if (this_boundary_node_ids.size())
+      {
+        if (use_binary_search)
         {
 #ifdef LIBMESH_HAVE_NANOFLANN
           typedef nanoflann::KDTreeSingleIndexAdaptor<nanoflann::L2_Simple_Adaptor<Real, VectorOfNodesAdaptor>, VectorOfNodesAdaptor, 3> kd_tree_t;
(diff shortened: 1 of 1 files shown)


roystgnr:
Is this what's been giving me false positives with the Apps tests?  And shouldn't we enable the Apps tests on this PR before merging?  :+1: once you do and it passes.

Finally: it scares me that this bug made it all the way to Bison before being triggered.  Is there any way we could add a unit test or tweak misc ex10 to get coverage that would have caught this?

lindsayad:
> Finally: it scares me that this bug made it all the way to Bison before being triggered. Is there any way we could add a unit test or tweak misc ex10 to get coverage that would have caught this?

I volunteer someone who does stitched meshes on the reg, maybe someone like @permcody or @friedmud :smile:

permcody:
What was (still is) missing from the coverage is something that can't, or shouldn't be stitched. We have tests that make sure that well behaved things work properly. I guess it didn't dawn on anyone to throw something at the stitcher that really couldn't be stitched!

jwpeterson:
> I guess it didn't dawn on anyone to throw something at the stitcher that really couldn't be stitched!

Is that what the BISON test was doing, or was that just a bad test?

permcody:
Yeah - The failure was in BISON's 1.5D mesh builder. If you've seen that mesh, it's just a bunch of EDGE2 lines stacked up that are disconnected from each other. Apparently that mesh is being fed to the stitcher, because the object that creates it actually has work to do in 2D, but there's nothing to stitch in the 1.5D case. The new logic didn't have a check that the "closest" node found in Nanoflann was actually in the same position so it was stitching crap together all over the place. The mesh after it got through the new stitcher looked like a pile of pick up sticks.

jwpeterson:
> The mesh after it got through the new stitcher looked like a pile of pick up sticks.

lmao. Is that not what you wanted?

PR #2024, opened 2019-02-06
Add static FEInterface::get_continuity()

jwpeterson:
This is useful because the only way to determine a given variable's
continuity is currently to build an FE object and call
get_continuity() on it, which is a bit inconvenient.


roystgnr:
Thanks!

PR #2025, opened 2019-02-06
Add version of FEInterface::n_dofs() taking an Elem *

jwpeterson:
This lets us avoid the fix (hack) introduced in #1987 and might be useful in other places as well but I did not check into it too closely.

### DIFF
--- a/include/fe/fe_interface.h
+++ b/include/fe/fe_interface.h
@@ -100,6 +100,17 @@ public:
                              const ElemType t);
 
   /**
+   * Similar to the function above but takes an Elem * and accounts for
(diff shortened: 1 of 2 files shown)


roystgnr:
I would have called it a hack *removed* in #1987, and honestly having the extra overload is a much saner idea.

I don't think it'll be universally useful until we do the same thing in a few other places, though - usually when we create one of these elevated-p FEType hacks we also use it in n_dofs_at_node, and sometimes in n_dofs_on_elem, dofs_on_edge, dofs_on_side.

It looks like we could already simplify hp_coarsentest.C, but everything else would require adding similar overloads elsewhere.

Issue #2026, opened 2019-02-07
find_neighbors finds a different neighbor?


lindsayad:
This may have something to do with #1678. We had a user come to us with a parallel `DisplacedProblem` simulation running with `--distributed-mesh` that fails with a PETSc error like:
```
[0]PETSC ERROR: --------------------- Error Message --------------------------------------------------------------
[0]PETSC ERROR: Arguments are incompatible
[0]PETSC ERROR: Incompatible vector local lengths parameter # 1 local size 4644 != parameter # 2 local size 4632
```
The problem occurs when we try to sync the `displaced_system.current_local_solution` with the `undisplaced_system.current_local_solution`. The global and local sizes of these vectors are the same; however, the `_send_list` is different. The `_send_list` is larger for the `undisplaced_system` because the number of elements to ghost is larger by two elements (for both processes: 67 vs 65 for processor 0; 65 vs 63 for processor 1). I latched onto one of the elements (element 1448) that is ghosted for the `undisplaced_system` on processor 0 but not for the `displaced_system`. 1448 is a neighbor of element 664 in the undisplaced system, **but not** in the displaced system!

An lldb summary:
```
(lldb) b EquationSystems::init
Breakpoint 1: where = libmesh_dbg.so.0`libMesh::EquationSystems::init() + 19 at equation_systems.C:98, address = 0x00007f55759b9843
(lldb) c
Process 30667 resuming
Process 30667 stopped
* thread #1, name = 'golem-dbg', stop reason = breakpoint 1.1
    frame #0: 0x00007f55759b9843 libmesh_dbg.so.0`libMesh::EquationSystems::init(this=0x0000000001f5a3d8) at equation_systems.C:98
   95  
   96   void EquationSystems::init ()
   97   {
-> 98     const unsigned int n_sys = this->n_systems();
   99  
   100    libmesh_assert_not_equal_to (n_sys, 0);
   101 
```
This corresponds to the the undisplaced equation systems:
<pre>
(lldb) p _mesh.elem_ptr(664)
(libMesh::Tri3 *) $0 = 0x0000000001dbb330
(lldb) p _mesh.elem_ptr(664)->neighbor_ptr(0)
(libMesh::Tri3 *) $1 = 0x0000000001e2aa60
(lldb) p _mesh.elem_ptr(664)->neighbor_ptr(1)
(libMesh::Tri3 *) $2 = 0x0000000001e29800
[...]

PR #2028, opened 2019-02-08
Add more copying to DistributedMesh and ReplicatedMesh copy constructors

lindsayad:
The decision to `skip_find_neighbors` in the call to `UnstructuredMesh::copy_nodes_and_elements` is based partially on #2026 but I also think that it's more consistent with the spirit of a copy constructor...it's copying over the neighbour mapping of the other mesh.

### DIFF
--- a/src/mesh/distributed_mesh.C
+++ b/src/mesh/distributed_mesh.C
@@ -73,7 +73,7 @@ DistributedMesh::DistributedMesh (const DistributedMesh & other_mesh) :
   _next_free_unpartitioned_node_id(this->n_processors()),
   _next_free_unpartitioned_elem_id(this->n_processors())
 {
-  this->copy_nodes_and_elements(other_mesh);
+  this->copy_nodes_and_elements(other_mesh, true);
   _n_nodes = other_mesh.n_nodes();
   _n_elem  = other_mesh.n_elem();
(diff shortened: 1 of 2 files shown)


roystgnr:
The added copying of names looks like a win for correctness.  :+1:

Using skip_find_neighbors looks like a win for consistency and efficiency.  :+1:  We might also want to switch that map to unordered_map if this code path is about to get much heavier use?

But before we merge this - can you explain to me *how* it's fixing #2026?  The find_neighbors operation here ought to be redundant, but not incorrect!

lindsayad:
I don't see this as fixing #2026; I see this as *avoiding* #2026. If I knew what was "wrong" (or just different?) with the result of `find_neighbors`, I would try to fix it...I was hoping someone smarter could do that

lindsayad:
It looks to me like we are [already using](https://github.com/libMesh/libmesh/blob/master/src/mesh/unstructured_mesh.C#L121) `unordered_map`

roystgnr:
Ah, you're right, that got fixed a year and a half ago.  Sorry, I think I was looking at "git log -p" code rather than the latest.

Ok, let's merge this as-is, but definitely not close out #2026.  We might be avoiding the bug but I doubt we'll be able to avoid it forever, and it would sure be great to fix it before a user hits it again!

Issue #2029, opened 2019-02-10
Create v1.4.0 release


jwpeterson:
Reference this ticket in bug fixes, etc. that get cherry-picked over from master onto the release branch.

jwpeterson:
1.4.0 was released on March 11, 2019. There have been a couple of small bug fixes since then, so I'm contemplating a 1.4.1 release at some point if enough of them accumulate. I'll keep using this issue number for those bug fixes.

PR #2030, opened 2019-02-11
Add shape and skew metrics for Hexes.

jwpeterson:
This is a follow-up to work in #1535 (60487b58), where we added the
analogous shape quality metrics for Quads.  The formulas used are
from:

P. Knupp, "Algebraic mesh quality metrics for
unstructured initial meshes," Finite Elements in Analysis
and Design 39, 2003, p. 217-241, Sections 6.2 and 6.3.

### DIFF
--- a/examples/systems_of_equations/systems_of_equations_ex6/systems_of_equations_ex6.C
+++ b/examples/systems_of_equations/systems_of_equations_ex6/systems_of_equations_ex6.C
@@ -99,10 +99,10 @@ public:
   {
     PetscErrorCode ierr = 0;
     ierr = KSPSetType (_petsc_linear_solver.ksp(), const_cast<KSPType>(KSPCG));
-    LIBMESH_CHKERR(ierr);
+    CHKERRABORT(_petsc_linear_solver.comm().get(), ierr);
 
     ierr = PCSetType (_petsc_linear_solver.pc(), const_cast<PCType>(PCBJACOBI));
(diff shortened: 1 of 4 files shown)


jwpeterson:
Attaching a bunch of optional tests for non-standard configurations so we can check whether that stuff is working for the v1.4.0 release (#2029).

jwpeterson:
I'm not really sure what to say about the results on these 20% deformed meshes other than it is possible for a given element to be among the highest quality for a given metric while being one of the lowest quality in another metric. I guess deciding which one you care about the most may depend on your application... 

hex_quality_diagonal
hex_quality_shape
hex_quality_skew
hex_quality_stretch
hex_quality_taper

roystgnr:
:+1: - I wonder if it would be reasonable to add some "meta" metrics eventually?  Even if a, b, and c give very different results for your mesh, you can run min(a,b,c) and feel happy as long as nothing is outstandingly bad there?  That only works if you can normalize them well enough to make them comparable, though, and I wouldn't even bet on the proper normalization being linear.

jwpeterson:
Yeah, in his paper Knupp talks about combining the shape and skew metrics above with a "size" metric he also proposes, just by computing their product. This works for his metrics because they are non-dimensional and guaranteed to be between 0 and 1. We don't currently have a size metric in libmesh because it requires an additional piece of data (desired element size) which is application-dependent.

PR #2031, opened 2019-02-12
Write Exodus files in netCDF-4 format when HDF5 is available.

jwpeterson:
This seems to work even in the relatively old (v5.22) version of
Exodus that we currently distribute with libmesh, nevertheless it
would be good to update to a more recent version of Exodus [0] at some
point.

[0]: The master branch of https://github.com/gsjaardema/seacas comes
     with Exodus v7.14.

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -1102,7 +1102,20 @@ void ExodusII_IO_Helper::create(std::string filename)
             (std::min(sizeof(Real), sizeof(double)));
         }
 
-      ex_id = exII::ex_create(filename.c_str(), EX_CLOBBER, &comp_ws, &io_ws);
+      // By default we just open the Exodus file in "EX_CLOBBER" mode,
+      // which, according to "ncdump -k", writes the file in "64-bit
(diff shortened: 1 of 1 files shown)


jwpeterson:
Actually it looks like the current version of Exodus is [7.18](https://github.com/gsjaardema/seacas/blob/master/packages/seacas/libraries/exodus/include/exodusII.h).

dknez:
:+1:

roystgnr:
That's awesome, thanks.  I had no idea it would be so simple!

Is there any way to get a handle that would let us append to the files with raw HDF5 APIs?  In my fantasies, the next "native" libMesh format *starts* with Exodus+Nemesis, so that we can use the same files for both restarts and low-order visualization, but then we add additional data arrays to encode p refinement levels, higher order basis coefficients, or whatever else the standard formats don't support.

jwpeterson:
> Is there any way to get a handle that would let us append to the files with raw HDF5 APIs?

Definitely not (that I know of) from Exodus. Their API seems to be designed around the fact that you don't really need/want to know what the underlying file type (netcdf3 vs. netcdf with support for large datasets vs. netcdf4 vs. etc.) is.

PR #2034, opened 2019-02-14
Reimplement ExodusII_IO_Helper::NamesData using C APIs.

jwpeterson:
Even though I think the previous API worked fine, the idea with this
change is that it should be guaranteed to be compatible with the
Exodus C API and there are no questions about std::vector<char>
compatibility.

* Use std::malloc, std::free since we include <cstdlib>
* Use std::calloc to zero-initialize strings. This way there is no
  possibility that any of the strings can have garbage characters in
  them.

### DIFF
--- a/include/mesh/exodusII_io_helper.h
+++ b/include/mesh/exodusII_io_helper.h
@@ -1278,11 +1278,15 @@ class ExodusII_IO_Helper::NamesData
 public:
   /**
    * Constructor.  Allocates enough storage to hold n_strings of
-   * length string_length.  (Actually allocates string_length+1 characters
-   * per string to account for the trailing '\0' character.)
+   * length string_length.
    */
   explicit
-  NamesData(size_t n_strings, size_t string_length);
+  NamesData(std::size_t n_strings, std::size_t string_length);
+
+  /**
+   * Frees memory allocated in the constructor.
+   */
+  ~NamesData();
 
   /**
    * Adds another name to the current data table.
@@ -1300,12 +1304,12 @@ public:
   char * get_char_star(int i);
 
 private:
-  // C++ data structures for managing string memory
-  std::vector<std::vector<char>> data_table;
-  std::vector<char *> data_table_pointers;
+  // Underlying C data structure managed by this class.
+  char ** data_table;
 
-  size_t counter;
-  size_t table_size;
+  std::size_t counter;
+  std::size_t table_size;
+  std::size_t string_len;
 };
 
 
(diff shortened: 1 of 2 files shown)


roystgnr:
Doesn't seem necessary to me.  The calloc change would forgo any question of reads past the terminating NULL, but the API has got to be designed to handle arrays of string literals, in which case it couldn't be dumb enough to try that, right?  I don't even see a string_length in the Exodus API, just in ours.  And compatibility between vector-char and char strings is just standard C++, as long as (like we are) you're careful about null terminators.

permcody:
Well there was the whole "burnup📞" incident... 

This might not be related at all, but we definitely have seen some reading past the end of the string crap with Exodus before.

permcody:
Ha! Here's that old ticket:
https://github.com/idaholab/moose/issues/4815

roystgnr:
> we definitely have seen some reading past the end of the string crap with Exodus before

Hah!  But the flaw there has to be something to do with a string length instead of a null terminator, right?  You can see the \000 in those strings.  So using calloc would just have given a bunch of incorrect \000 characters on the backend instead of garbage, not *no* characters instead of garbage.

permcody:
Yeah - I agree, those are null terminated so it's a string length problem.

jwpeterson:
> Doesn't seem necessary to me.

I agree, I think you should be able to achieve the same end result using either C++ or C strings. 

I think maybe there was an issue with the previous code, though. On this line, 
```
data_table[i].resize(string_length + 1);
```
we were making the strings 1 extra character long to account for the null terminator, when in fact they should be exactly `string_length` characters long, with the last character reserved for a '\0'. So, it makes sense that those 2 MOOSE tests with super long variable names failed, because we were comparing the first `MAX_STR_LENGTH` characters, when in fact only the first `MAX_STR_LENGTH-1` characters should match. 

I don't know if this was also causing the "burnup telephone" issues of the past, but it's possible. I can also report that if you write netCDF4 Exodus files they don't have any of those weird characters in the element variable names any more either.

jwpeterson:
OK, so the diffs make sense to me, and I think the code was wrong before but is right now. In the previous version of the code, we wrote a variable name with 32 (`== MAX_STR_LENGTH`) actual characters and no null terminator (count them and see!)
```
elemental_source_from_sub_elemen
```
whereas now we are correctly writing a 31 character string with a null terminator as the 32nd character:
```
elemental_source_from_sub_eleme
```
hence the diff. I believe the previous strings which didn't have a null terminator in the last position could have been causing problems with ncdump output previously but I can't say for sure.

jwpeterson:
Since there are only two failing tests, the easiest way to move forward might be to change the MOOSE test to temporarily write out shorter variable names (and regold it) until you guys have time to do another libmesh update, at which time you can switch back to the original longer variable names, because I think the long variable name test is definitely a good one to have.

PR #2035, opened 2019-02-14
Upgrade bundled netCDF to version 4.6.2

jwpeterson:
While experimenting with reading Exodus files written in netCDF4 format in Python, we ran across some issues with h5py that seemed to be related to UTF-8 encodings. While looking through the netCDF release notes, I came across this issue related to UTF-8/netCDF4:

​https://github.com/Unidata/netcdf-c/issues/298

Which was apparently fixed by this PR:

​https://github.com/Unidata/netcdf-c/pull/316

And that made it into netCDF release 4.5.0-rc1 (June 5, 2017):

​https://www.unidata.ucar.edu/software/netcdf/docs/RELEASE_NOTES.html

Since our current netCDF 4.4.1.1 predates this release, it seems that netCDF4 files we write are affected by this bug. Following roughly the same update procedure as @benkirk did for the 4.4.1.1 update, I was able to update to 4.6.2 locally without too much difficulty.

I don't think this needs to go in our v1.4.0 release, but if there is enough demand for it we can always release a v1.4.1, similar to what was done for v1.2.1.


jwpeterson:
+400k, -350k :roll_eyes: 

If we have to start updating netCDF more often, we might be better off making it into a git submodule.

permcody:
That's some serious code - I guess you can take the rest of the year off!

jwpeterson:
@pbauman, I've pasted the GRINS failure below. Somehow one of the files in the netCDF source dir gets a local modification:
```
Successfully removed /femputer/pbauman/civet_build_testing_root/libmesh-devel.
BUILD_ROOT/libMesh/: git checkout master
error: Your local changes to the following files would be overwritten by checkout:
	contrib/netcdf/netcdf-c-4.6.2/libsrc/ncx.c
Please, commit your changes or stash them before you can switch branches.
Aborting
ERROR: exiting with code 1
```
Would it be possible for you to log in there and tell me what the diff is? I don't see it on my local system and it's possible the file in question shouldn't even be under source control if it is meant to be a generated file or something...

jwpeterson:
Yeah, sure enough it says that file is generated:
```
/* Do not edit this file. It is produced from the corresponding .m4 source */
```
Hopefully it's generated during configure and not bootstrap. I will check on this.

jwpeterson:
@pbauman I'm also curious if you know how the dbg target could have passed even though the devel target failed? I'm unable to click on the "Details" links at the moment (femputer timeout) so I'm not exactly sure what the differences are between the two recipes.

pbauman:
> @pbauman I'm also curious if you know how the dbg target could have passed even though the devel target failed? I'm unable to click on the "Details" links at the moment (femputer timeout) so I'm not exactly sure what the differences are between the two recipes.

@jwpeterson No idea. There should be no difference in that step (literally calling the same script in that step of both recipes). I've restarted femputer and redelivered the last push you did so it'll rerun in a bit and we can see.

jwpeterson:
> I've restarted femputer and redelivered the last push

Thanks, seems to be working again!

jwpeterson:
This passes all the tests and doesn't appear to have broken anything new in "make distcheck" according to my local testing, so I'm going to go ahead and merge. There were a few other issues with the distcheck target, however, so I will be opening a separate PR for that.

PR #2036, opened 2019-02-15
Fix issue with PETSc that has METIS but no ParMETIS

jwpeterson:
We previously assumed the former implied the latter, but that is not always the case.


jwpeterson:
This should be cherry-picked onto the v1.4.0 branch (#2029).

PR #2042, opened 2019-02-20
unit tests for new infinite element features

BalticPinguin:
With some delay, but finally I finished the unit test for the new features for infinite elements.
I also fixed some related things; I hope, finally I got all issues...

### DIFF
--- a/src/mesh/mesh_function.C
+++ b/src/mesh/mesh_function.C
@@ -512,7 +512,7 @@ void MeshFunction::gradient (const Point & p,
                   {
                     // local coordinates
                     for (std::size_t v=0; v<dim; v++)
-                      for (std::size_t xyz=0; xyz<dim; xyz++)
+                      for (std::size_t xyz=0; xyz<LIBMESH_DIM; xyz++)
                         {
                           // FIXME: this needs better syntax: It is matrix-vector multiplication.
                           grad(xyz) += data.local_transform[v][xyz]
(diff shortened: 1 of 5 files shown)


roystgnr:
Awesome; thanks!

src/mesh/mesh_function.C
```diff
@@ -512,7 +512,7 @@ void MeshFunction::gradient (const Point & p,
                   {
                     // local coordinates
                     for (std::size_t v=0; v<dim; v++)
-                      for (std::size_t xyz=0; xyz<dim; xyz++)
+                      for (std::size_t xyz=0; xyz<LIBMESH_DIM; xyz++)
```
roystgnr: This looks like a good catch... we weren't getting MeshFunction gradients correct for lower-dimensional elements in higher-dimensional spaces?

roystgnr:
Not technically unit tests but clearly *any* testing is an improvement.  Since there's an exact solution to misc_ex14, would it be possible to throw a ```libmesh_assert_less(abs(soln-exact_solution), some_tolerance)``` in somewhere (the ```line_print``` function?) there?

src/mesh/mesh_function.C
BalticPinguin: to be honest, I never tested. But I think, there would have been some error.
Be aware that this part is only reached for infinite elements. Since I didn't use it for their surfaces, no one did, I guess.

BalticPinguin:
yes, this is a good idea. I will add it.

roystgnr:
Thanks!

PR #2047, opened 2019-02-21
PetscDMWrapper Geometric Multigrid + Fieldsplit capabilities

bboutkov:
After far too long, this feature set has finally reached a point where I believe it is ready for further external scrutiny! The idea behind this PR implements step 2 of the discussion in #1567, and also enables the use of fieldsplit capabilities from the command line. I've been testing this fairly extensively on both replicated and distributed meshes up to `np 256` through GRINS for a while now, and have verified that the infrastructure behaves as expected for vanilla GMG on Poisson problems, and also when restricting the GMG construction for example to the velocity block of a Stokes problem.

The change set incorporates quite a long period of development; I've put in a fair amount of effort to condense the commits to logical chunks of core functionalities, but occasionally I include either small fixup commits, or optimization type commits ( such as 26e654c ) which at one point in time exposed certain difficulties that needed to be resolved. Also please note that the occasional documentation improvement slips in incrementally through the commit range, if any such things are troublesome and should be either squashed in or separated please let me know and I can make the necessary adjustments.

While certain core components have been individually tested (mainly the generic projector usage in #1507 and friends), this PR certainly lacks an overall testing suite. I plan to shortly shortly throw together some GRINS based regression testing which will attempt to capture things like the number of linear and nonlinear iterations on some benchmark problems, but am certainly open to suggestions for things that can or should be verified on this end.

Please let me know of things that can be improved, that need fixing, or explaining, and thanks in advance for the review!

### DIFF
--- a/include/solvers/petsc_dm_wrapper.h
+++ b/include/solvers/petsc_dm_wrapper.h
@@ -120,7 +120,7 @@ private:
    * init_dm_data() should be called before this function.
    */
   DM & get_dm(unsigned int level)
-  { libmesh_assert(level < _dms.size());
+    { libmesh_assert_less(level, _dms.size());
     return *(_dms[level].get()); }
 
   //! Get reference to PetscSection for the given mesh level
@@ -128,7 +128,7 @@ private:
    * init_dm_data() should be called before this function.
    */
   PetscSection & get_section(unsigned int level)
-  { libmesh_assert(level < _sections.size());
+    { libmesh_assert_less(level, _sections.size());
     return *(_sections[level].get()); }
 
   //! Get reference to PetscSF for the given mesh level
@@ -136,7 +136,7 @@ private:
    * init_dm_data() should be called before this function.
    */
   PetscSF & get_star_forest(unsigned int level)
-  { libmesh_assert(level < _star_forests.size());
+    { libmesh_assert_less(level, _star_forests.size());
     return *(_star_forests[level].get()); }
 
   //! Takes System, empty PetscSection and populates the PetscSection
--- a/include/solvers/petsc_dm_wrapper.h
+++ b/include/solvers/petsc_dm_wrapper.h
@@ -73,7 +73,7 @@ private:
   std::vector<std::unique_ptr<PetscSF>> _star_forests;
 
   //! Init all the n_mesh_level dependent data structures
-  void init_dm_data(unsigned int n_levels);
+  void init_dm_data(unsigned int n_levels, const Parallel::Communicator & comm);
 
   //! Get reference to DM for the given mesh level
   /**
--- a/include/solvers/petsc_dm_wrapper.h
+++ b/include/solvers/petsc_dm_wrapper.h
@@ -44,18 +44,27 @@ namespace libMesh
   struct PetscDMContext
   {
     int n_dofs;
+    int mesh_dim;
     DM * coarser_dm;
     DM * finer_dm;
+    DM * global_dm;
     libMesh::PetscMatrix<libMesh::Real > * K_interp_ptr;
+    libMesh::PetscMatrix<libMesh::Real > * K_sub_interp_ptr;
     libMesh::PetscMatrix<libMesh::Real > * K_restrict_ptr;
     libMesh::PetscVector<libMesh::Real > * current_vec;
+
+    //! Stores local dofs for each var for use in subprojection matrixes
+    std::vector<std::vector<numeric_index_type>> dof_vec;
 
     PetscDMContext()
     {
       n_dofs = -12345;
+      mesh_dim = -12345;
       coarser_dm = nullptr;
       finer_dm = nullptr;
+      global_dm = nullptr;
       K_interp_ptr = nullptr;
+      K_sub_interp_ptr = nullptr;
       K_restrict_ptr = nullptr;
       current_vec = nullptr;
     }
@@ -99,6 +108,9 @@ private:
   //! Vector of projection matrixes for all grid levels
   std::vector<std::unique_ptr<PetscMatrix<Real>>> _pmtx_vec;
 
+  //! Vector of sub projection matrixes for all grid levels for fieldsplit
+  std::vector<std::unique_ptr<PetscMatrix<Real>>> _subpmtx_vec;
+
   //! Vector of internal PetscDM context structs for all grid levels
   std::vector<std::unique_ptr<PetscDMContext>> _ctx_vec;
 
@@ -110,7 +122,6 @@ private:
 
   //! Stores n_local_dofs for each grid level, to be used for projection vector sizing
   std::vector<unsigned int> _mesh_dof_loc_sizes;
-
 
   //! Init all the n_mesh_level dependent data structures
   void init_dm_data(unsigned int n_levels, const Parallel::Communicator & comm);
--- a/include/solvers/petsc_dm_wrapper.h
+++ b/include/solvers/petsc_dm_wrapper.h
@@ -53,11 +53,11 @@ namespace libMesh
     PetscDMContext()
     {
       n_dofs = -12345;
-      coarser_dm = NULL;
-      finer_dm = NULL;
-      K_interp_ptr = NULL;
-      K_restrict_ptr = NULL;
-      current_vec = NULL;
+      coarser_dm = nullptr;
+      finer_dm = nullptr;
+      K_interp_ptr = nullptr;
+      K_restrict_ptr = nullptr;
+      current_vec = nullptr;
     }
 
   };
--- a/include/solvers/petsc_dm_wrapper.h
+++ b/include/solvers/petsc_dm_wrapper.h
@@ -19,8 +19,10 @@
 #define LIBMESH_PETSC_DM_WRAPPER_H
 
 #include "libmesh/libmesh_common.h"
+#include "libmesh/petsc_macro.h"
 
 #ifdef LIBMESH_HAVE_PETSC
+#if !PETSC_VERSION_LESS_THAN(3,7,3)
 
 #include <vector>
 #include <memory>
@@ -216,6 +218,7 @@ private:
 
 }
 
+#endif // #if PETSC_VERSION
 #endif // #ifdef LIBMESH_HAVE_PETSC
 
 #endif // LIBMESH_PETSC_DM_WRAPPER_H
--- a/include/solvers/petsc_dm_wrapper.h
+++ b/include/solvers/petsc_dm_wrapper.h
@@ -193,7 +193,7 @@ private:
    * The PetscSection contains local dof information. This helper function just facilitates
    * sanity checking that in fact it only has n_local_dofs.
    */
-  dof_id_type check_section_n_dofs( const System & system, PetscSection & section );
+  dof_id_type check_section_n_dofs( PetscSection & section );
 
   //! Helper function to reduce code duplication when setting dofs in section
   void add_dofs_helper (const System & system,
--- a/include/solvers/petsc_dm_wrapper.h
+++ b/include/solvers/petsc_dm_wrapper.h
@@ -31,12 +31,36 @@
 #include "libmesh/ignore_warnings.h"
 #include <petsc.h>
 #include "libmesh/restore_warnings.h"
+#include "libmesh/petsc_matrix.h"
+#include "libmesh/petsc_vector.h"
 
 namespace libMesh
 {
   // Forward declarations
   class System;
   class DofObject;
+
+  //! Struct to house data regarding where in the mesh hierarchy we are located
+  struct PetscDMContext
+  {
+    int n_dofs;
+    DM * coarser_dm;
+    DM * finer_dm;
+    libMesh::PetscMatrix<libMesh::Real > * K_interp_ptr;
+    libMesh::PetscMatrix<libMesh::Real > * K_restrict_ptr;
+    libMesh::PetscVector<libMesh::Real > * current_vec;
+
+    PetscDMContext()
+    {
+      n_dofs = -12345;
+      coarser_dm = NULL;
+      finer_dm = NULL;
+      K_interp_ptr = NULL;
+      K_restrict_ptr = NULL;
+      current_vec = NULL;
+    }
+
+  };
 
 /**
  * This class defines a wrapper around the PETSc DM infrastructure.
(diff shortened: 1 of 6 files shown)


jwpeterson:
Hi Boris,

This sounds like a very promising set of improvements. I don't have time to review it in detail right at this moment, but at first glance I'd say your commit history looks fine to me as-is, I wouldn't worry too much about changing it.

As far as the build failures, it looks like you might be using a newer (or older?) version of PETSc than is being used by MOOSE (current 3.9.4)?
```
../src/solvers/petsc_dm_wrapper.C: In function ‘PetscErrorCode libMesh::__libmesh_petsc_DMCoarsen(DM, MPI_Comm, _p_DM**)’:
../src/solvers/petsc_dm_wrapper.C:187:18: error: ‘DMGetSection’ was not declared in this scope
           ierr = DMGetSection(*(p_ctx->coarser_dm), &section);
                  ^~~~~~~~~~~~
```

bboutkov:
Yes, actually all of my code is using PETSc master as of a couple of days ago. I see that for GRINS Civet is using 3.7.4. Is there a natural way for me to bump the Civet PETSc version? In fact some of the code I recently added is in response to changes from PETSc from only a couple of weeks ago. How far back should I go to start making version-specific fixes?

jwpeterson:
> I see that for GRINS Civet is using 3.7.4. Is there a natural way for me to bump the Civet PETSc version?

You'd need access to the server and knowledge of how and where to build stuff, but yeah this is definitely possible.

> How far back should I go to start making version-specific fixes?

I don't think you can "fix" stuff for older versions if you need code that was recently committed to PETSc master?  The new capability should all just be wrapped in ifdefs so that it will be disabled on all older PETScs. 

This can be tricky to get right, though. Since PETSc master is obviously a moving target, code that compiled with 
```
#ifndef PETSC_VERSION_RELEASE
```
for you last week won't work for someone who is using petsc master from two months ago.

If you know that the code you need will be in e.g. PETSc 3.11 you can definitely safely guard it with 
```
#if !PETSC_VERSION_LESS_THAN(3,11,0)
```
but this makes it hard for you to develop, because that #if statement also fails for you.

include/solvers/petsc_dm_wrapper.h
```diff
@@ -31,13 +33,46 @@
...
+    DM * global_dm;
+    libMesh::PetscMatrix<libMesh::Real > * K_interp_ptr;
+    libMesh::PetscMatrix<libMesh::Real > * K_sub_interp_ptr;
+    libMesh::PetscMatrix<libMesh::Real > * K_restrict_ptr;
+    libMesh::PetscVector<libMesh::Real > * current_vec;
+
+    //! Stores local dofs for each var for use in subprojection matrixes
+    std::vector<std::vector<numeric_index_type>> dof_vec;
+
+    PetscDMContext()
```
jwpeterson: Minor gripe: your struct's constructor should use an initialization list.

include/solvers/petsc_dm_wrapper.h
```diff
@@ -31,13 +33,46 @@
...
 
+  //! Struct to house data regarding where in the mesh hierarchy we are located
+  struct PetscDMContext
+  {
+    int n_dofs;
+    int mesh_dim;
+    DM * coarser_dm;
+    DM * finer_dm;
+    DM * global_dm;
+    libMesh::PetscMatrix<libMesh::Real > * K_interp_ptr;
```
jwpeterson: You are in namespace libMesh, so these `libMesh::` qualifications should be unnecessary? If they are just there for human code readability that's OK with me.

src/solvers/petsc_dm_wrapper.C
```diff
@@ -16,39 +16,355 @@
...
+  //
+  // Since they must have C linkage they have no knowledge of a namespace.
+  // We give them an obscure name to avoid namespace pollution.
+  //--------------------------------------------------------------------
+  extern "C"
+  {
+
+    //! Help PETSc create a subDM given a global dm when using fieldsplit
+#if PETSC_VERSION_LESS_THAN(3,9,0)
+    PetscErrorCode __libmesh_petsc_DMCreateSubDM(DM dm, PetscInt numFields, PetscInt fields[], IS *is, DM *subdm)
```
jwpeterson: We've gone away from double-leading-underscore names for PETSc callback functions, see for example petsc_nonlinear_solver.h where they are now called `libmesh_petsc_snes_monitor`, `libmesh_petsc_snes_residual`, etc. The only reason for this is that all double-leading-underscore names are technically reserved for the compiler... there's really no danger of a name collision in this case, though.

src/solvers/petsc_dm_wrapper.C
```diff
@@ -16,39 +16,355 @@
...
 
 #include "libmesh/libmesh_common.h"
+#include "libmesh/petsc_macro.h"
 
 #ifdef LIBMESH_HAVE_PETSC
+#if !PETSC_VERSION_LESS_THAN(3,7,3)
 
 #include "libmesh/ignore_warnings.h"
 #include <petscsf.h>
+#include <petsc/private/dmimpl.h>
```
jwpeterson: It's usually inadvisable to include or rely on `petsc/private` headers or any of their internal implementation stuff. But this may be unavoidable if you are using relatively new features that aren't otherwise accessible...

src/solvers/petsc_dm_wrapper.C
```diff
@@ -16,39 +16,355 @@
...
+    {
+      PetscErrorCode ierr;
+      // Basically, we copy the PETSc ShellCreateSubDM implementation,
+      // but also need to set the embedding dim and we also propagate
+      // the relevant function pointers to the subDM for GMG purposes.
+      // Since this is called by PETSc we gotta pull some of this info
+      // from the context in the DM.
+
+      // Get our context
+      void * ctx = NULL;
```
jwpeterson: I think you can just use `nullptr` here, it should interact seamlessly with C APIs AFAIK.

src/solvers/petsc_dm_wrapper.C
```diff
@@ -16,39 +16,355 @@
...
+
+      // If the probed number of fields is less than the number of
+      // global fields, this amounts to PETSc 'indicating' to us we
+      // are doing FS. So, we must create subsections for the coarser
+      // DMs.
+      if ( nfieldsf < nfieldsg )
+        {
+          PetscSection section;
+          PetscSection subsection;
+          PetscInt * subfields = new PetscInt[nfieldsf]; // extracted fields
```
jwpeterson: I'd prefer a `std::vector` here, and then just pass the `.data()` array to PETSc to use. Then you don't have to remember to `free []` it later.

src/solvers/petsc_dm_wrapper.C
```diff
@@ -16,39 +16,355 @@
...
+          // Loop over the fields and merge their index sets.
+          std::vector<std::vector<numeric_index_type>> allrows,allcols;
+          std::vector<numeric_index_type> rows,cols;
+          allrows = p_ctx_f->dof_vec;
+          allcols = p_ctx_c->dof_vec;
+
+          // For internal libmesh submat extraction need to merge all
+          // field dofs and then sort the vectors so that they match
+          // the Projection Matrix ordering
+          static const int n_subfields = nfieldsc;
```
jwpeterson: Should this be `static`? Usually function static variables are used as "stateful" counters or flags or something, but here you are just initializing it to some value, so `const` should be sufficient.

src/solvers/petsc_dm_wrapper.C
```diff
@@ -16,39 +16,355 @@
...
+                  rows.insert(rows.end(), allrows[i].begin(), allrows[i].end());
+                  cols.insert(cols.end(), allcols[i].begin(), allcols[i].end());
+                }
+              std::sort(rows.begin(),rows.end());
+              std::sort(cols.begin(),cols.end());
+            }
+
+          // Now that we have merged the fine and coarse index sets
+          // were ready to make the submatrix and pass it off to PETSc
+          Mat  submat;
```
jwpeterson: Is `submat` actually used here? Does PETSc have ownership of it somehow, or does it need to be destroyed?

src/solvers/petsc_dm_wrapper.C
```diff
@@ -57,40 +373,273 @@ void PetscDMWrapper::init_and_attach_petscdm(System & system, SNES & snes)
...
+
+  // How many MG levels did the user request?
+  int usr_requested_mg_lvls = 0;
+  usr_requested_mg_lvls = command_line_next("-pc_mg_levels", usr_requested_mg_lvls);
+
+  // Only construct however many levels were requested if something was actually requested
+  if ( usr_requested_mg_lvls != 0 )
+    {
+      // Dont request more than avail num levels on mesh, require at least 2 levels
+      libmesh_assert_less_equal( (unsigned int)usr_requested_mg_lvls, n_levels );
```
jwpeterson: C++ style `static_cast<unsigned int>` here.

src/solvers/petsc_dm_wrapper.C
roystgnr: It does, yeah.  Only difference between NULL and nullptr is typing, not representation.

src/solvers/petsc_dm_wrapper.C
roystgnr: Plus even if the "static" is not hurting anything this time, it's a bad habit to get into when writing code some of which might be threaded someday.

src/solvers/petsc_dm_wrapper.C
roystgnr: MatCreate usually needs to be matched with MatDestroy, yeah.  This looks like a memory leak?

src/solvers/petsc_dm_wrapper.C
roystgnr: Or could we just declare that as unsigned int to begin with?

pbauman:
> about what you guys are doing in particular

#1567 has longer discussion, but basically this is the last of the hooks to PETSc to do Galerkin type geometric multigrid using an existing mesh hierarchy generated by libMesh via command line options. And to be able to do it only on certain fieldsplits (this part took awhile/lots of interaction with Knepley to get right).

> It would be really nice to have some unit testing or example coverage of this if we don't already...

Agreed and we're open to suggestions here. We'll can some GRINS tests and add them in so they get hit after this is merged, but how would folks like to see this in here? We could probably augment an existing example to illustrate the options, but actual unit testing here is a real pain. On the GRINS side, we'll test that we converge in expected number of nonlinear and linear iterations, but it's hard to do much better than that.

This will also be the time to start discussion (#1567?) on how to add this in `PetscNonlinearSolver` so all libMesh apps can use this, not just FEMSystem based apps.

src/solvers/petsc_dm_wrapper.C
pbauman: In this case, it's currently unavoidable because we need this function and PETSc put it as private. Will lobby @knepley next week to make it public (or tell us what the better thing to do is).

roystgnr:
We already have a few examples (see adaptivity_ex3 for the most extreme case) where we just rerun one of the same codes over and over again with different FE spaces or AMR heuristics or whatever to get more test coverage.  Could we do the same thing with one of the femsystem_ex or adjoints_ex entries?  At least a couple of those are doing velocity+pressure or velocity+pressure+transport so you could get the fieldsplit capability covered too.

src/solvers/petsc_dm_wrapper.C
bboutkov: Cool that's exactly what I worried about, thanks for the info.

src/solvers/petsc_dm_wrapper.C
bboutkov: Ahh, I hated this part of the code and didn't know about this trick, thanks!

src/solvers/petsc_dm_wrapper.C
bboutkov: Oops, left around from one of the initial attempts.

jwpeterson:
I don't mind merging this without libmesh test coverage provided there is some in GRINS, but I also like @roystgnr's idea of having one of the existing examples possibly use the new solver.

roystgnr:
This looks good and it's a new feature rather than code likely to break anything existing, so we can merge now.

However: I don't run with GRINS as often as I used to, and I'm in the middle of massively refactoring the GenericProjector code, so I'm hoping @bboutkov will set up a new PR adding some GMG solver use to the existing examples ASAP, out of self-preservation.

bboutkov:
Ill look into some examples, thanks for the merge and review comments!

friedmud:
BTW: This fails with newer PETSc (from git):

```
../src/solvers/petsc_dm_wrapper.C: In member function ‘void libMesh::PetscDMWrapper::init_and_attach_petscdm(libMesh::System&, _p_SNES*&)’:
../src/solvers/petsc_dm_wrapper.C:609:18: error: ‘class libMesh::System’ has no member named ‘projection_matrix’
           system.projection_matrix(*_ctx_vec[i-1]->K_interp_ptr);
                  ^~~~~~~~~~~~~~~~~
  CXX      src/solvers/libmesh_dbg_la-steady_solver.lo
../src/solvers/petsc_linear_solver.C: In member function ‘virtual libMesh::LinearConvergenceReason libMesh::PetscLinearSolver<T>::get_converged_reason() const’:
../src/solvers/petsc_linear_solver.C:1722:10: error: ‘KSP_DIVERGED_PC_FAILED’ was not declared in this scope
     case KSP_DIVERGED_PC_FAILED        : return DIVERGED_PCSETUP_FAILED;
          ^~~~~~~~~~~~~~~~~~~~~~
../src/solvers/petsc_linear_solver.C:1722:10: note: suggested alternative: ‘KSP_DIVERGED_PCSETUP_FAILED’
     case KSP_DIVERGED_PC_FAILED        : return DIVERGED_PCSETUP_FAILED;
          ^~~~~~~~~~~~~~~~~~~~~~
          KSP_DIVERGED_PCSETUP_FAILED
```

friedmud:
@fdkong do you want to weigh in here?

roystgnr:
Kind of AFK right now so feel free to revert if we can't straighten this out quickly.

That System::projection_matrix() call is only enabled with LIBMESH_HAVE_METAPHYSICL - I guess all the CI has that turned on at this point so we didn't notice that the usage there wasn't properly protected?  The solution then would be to wrap the dependant solver stuff in LIBMESH_HAVE_METAPHYSICL too and either fallback-to-non-GMG or (my preference and probably much easier too) libmesh_not_implemented() in the alternative case.

knepley:
It changed slightly: https://bitbucket.org/petsc/petsc/src/73e1f9d0148528059d8b8285257d8c60b34ef8fd/include/petscksp.h#lines-449

friedmud:
Ok - so I may just be on a version of PETSc before the change - but after @bboutkov .  So this might not be something to fix.

Unfortunately - I'm stuck on this version of PETSc until my dissertation is done!  (One more month?)

For now, I just checked out the version of libMesh just before this got merged and everything is good.

PR #2053, opened 2019-02-26
Explicitly specify scope in else branch: libmesh_make_unique macro

lindsayad:
Currently if not LIBMESH_HAVE_CXX14_MAKE_UNIQUE we define libmesh_make_unique
to be equivalent to make_unique without namespace scoping. I believe
explicitly scoping to libMesh::make_unique makes sense.

This was a actually kind of a saga. See idaholab/moose#12968 for more detail. I guess it's a good thing that the macro doesn't currently explicitly scope because I would have never discovered that issue...but on the other hand I also would have avoided a couple hours of rabbit chasing if we had explicit libMesh scoping in the macro.

### DIFF
--- a/include/base/auto_ptr.h
+++ b/include/base/auto_ptr.h
@@ -47,7 +47,7 @@ std::unique_ptr<T> make_unique(Args &&... args)
   return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
 }
 }
-#define libmesh_make_unique make_unique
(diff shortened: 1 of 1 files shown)


jwpeterson:
I can't see any drawbacks to this... I'm wondering why it wasn't like that before? We always relied on the macro being called in files that had `using namespace libMesh;` at the top? Seems like this should also go in v1.4.0 (#2029).

roystgnr:
:+1: - good catch

PR #2054, opened 2019-02-27
Fix an issue with unique_id update

jiangwen84:
@roystgnr thanks!


roystgnr:
Thanks!  Sorry it took me so long to see this and kick Civet!

I take it the patch worked for your case?  Then we'll merge immediately once CI is happy, and I'll beg @jwpeterson to backport this onto the upcoming release branch if I forget.

I should hack up a unit test that triggers the original bug, but I don't know if or when I'll find time.

jiangwen84:
@roystgnr yes, the fix works for my case.

PR #2055, opened 2019-03-01
Add MeshBase interface for specifying PointLocator customizations.

jwpeterson:
Since you only get access to "sub" PointLocators via calls to
MeshBase::sub_point_locator(), there's not really a good way to set
values on the "real" PointLocator stored by the Mesh.  One approach is
to let the Mesh handle it, which is what we're doing in this PR.

* Set close_to_point tolerance when building PointLocator.
* Inherit the master PointLocator's tolerances if possible.

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -981,6 +981,18 @@ public:
   std::unique_ptr<PointLocatorBase> sub_point_locator () const;
 
   /**
+   * Set value used by PointLocatorBase::close_to_point_tol().
+   *
+   * Defaults to 0.0. If nonzero, calls close_to_point_tol() whenever
+   * a new PointLocator is built for use by this Mesh.  Since the Mesh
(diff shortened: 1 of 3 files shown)


roystgnr:
:+1: Yeah, this works for me.  I wish there was a solution that's less hackish than "individually shim every important point locator setter" but less dangerous than "encourage users to grab a reference to the master point locator".

I guess we could make PointLocatorBase inherit from a PointLocatorSettings superclass, move setters to the latter, and let MeshBase return a reference to the latter?  Sounds like the "right" way to do things but too much work.

jwpeterson:
Thanks! I was thinking along the same lines re: settings for the `PointLocator`. I was thinking of a standalone "Settings" class but the inheritance idea is good as well. 

I am also interested in improving the robustness/redesigning the `PointLocator` to use nanoflann/KD-Trees in some capacity, but not sure when I would have the time...

PR #2057, opened 2019-03-05
Add GMG+Fieldsplit testing for fem_sys_ex1 

bboutkov:
This PR exercises the capabilities added in #2047 by adding a minimal working example for fem_system ex1 which solves a steady Navier-Stokes system by utilizing GMG on the velocity block for both Distributed and Serial meshes.

The cd44332 commit captures the iteration counts from PETSc in the DiffSolver which will shortly be used in some regression testing being done through GRINS.

### DIFF
--- a/examples/fem_system/fem_system_ex1/fem_system_ex1.C
+++ b/examples/fem_system/fem_system_ex1/fem_system_ex1.C
@@ -36,7 +36,7 @@
 #include "libmesh/exodusII_io.h"
 #include "libmesh/kelly_error_estimator.h"
 #include "libmesh/mesh.h"
-#include "libmesh/serial_mesh.h"
+#include "libmesh/replicated_mesh.h"
 #include "libmesh/distributed_mesh.h"
 #include "libmesh/mesh_generation.h"
 #include "libmesh/mesh_refinement.h"
@@ -98,7 +98,7 @@ int main (int argc, char ** argv)
   const unsigned int max_adaptivesteps = infile("max_adaptivesteps", 10);
   const unsigned int dim               = infile("dimension", 2);
   const std::string slvr_type          = infile("solver_type", "newton");
-  const std::string mesh_type          = infile("mesh_type"  , "serial");
+  const std::string mesh_type          = infile("mesh_type"  , "replicated");
 
 #ifdef LIBMESH_HAVE_EXODUS_API
   const unsigned int write_interval    = infile("write_interval", 5);
@@ -116,8 +116,8 @@ int main (int argc, char ** argv)
 
   if (mesh_type == "distributed")
     mesh.reset(new DistributedMesh(init.comm()));
-  else if (mesh_type == "serial")
-    mesh.reset(new SerialMesh(init.comm()));
+  else if (mesh_type == "replicated")
+    mesh.reset(new ReplicatedMesh(init.comm()));
   else
     libmesh_error_msg("Error: specified mesh_type not understood");
 
--- a/examples/fem_system/fem_system_ex1/fem_system_ex1.C
+++ b/examples/fem_system/fem_system_ex1/fem_system_ex1.C
@@ -190,12 +190,10 @@ int main (int argc, char ** argv)
   if (slvr_type == "newton")
     system.time_solver->diff_solver() = libmesh_make_unique<NewtonSolver>(system);
   else if (slvr_type == "petscdiff")
-#ifdef LIBMESH_HAVE_PETSC
-#ifdef LIBMESH_HAVE_METAPHYSICL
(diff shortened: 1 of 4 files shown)


src/solvers/petsc_diff_solver.C
```diff
@@ -337,6 +337,15 @@ unsigned int PetscDiffSolver::solve()
...
 
+  PetscInt l_its, nl_its;
+  ierr = SNESGetLinearSolveIterations(_snes,&l_its);
+  LIBMESH_CHKERR(ierr);
+  this->_inner_iterations = l_its;
+
+  ierr = SNESGetIterationNumber(_snes,&nl_its);
+  LIBMESH_CHKERR(ierr);
+  this->_outer_iterations = nl_its;
+
```
roystgnr: Wow, this was a big oversight on my part.

examples/fem_system/fem_system_ex1/fem_system_ex1.C
```diff
@@ -36,6 +36,8 @@
 #include "libmesh/exodusII_io.h"
 #include "libmesh/kelly_error_estimator.h"
 #include "libmesh/mesh.h"
+#include "libmesh/serial_mesh.h"
```
roystgnr: Could you switch this to replicated_mesh.h (and later to ReplicatedMesh)?  The old misleading names are deprecated.

examples/fem_system/fem_system_ex1/fem_system_ex1.C
```diff
@@ -102,10 +112,17 @@ int main (int argc, char ** argv)
 
   // Create a mesh, with dimension to be overridden later, distributed
   // across the default MPI communicator.
-  Mesh mesh(init.comm());
+  std::shared_ptr<UnstructuredMesh> mesh;
+
+  if (mesh_type == "distributed")
+    mesh.reset(new DistributedMesh(init.comm()));
+  else if (mesh_type == "serial")
```
roystgnr: Ah, and s/serial/replicated here and in the input file too?

roystgnr:
This looks like exactly what I was hoping for, thanks!

bboutkov:
OK thanks for the review, and comments!

Sorry bout that missing PETSc snafu. I've been bitten by that one enough times that I really ought to just have a local config of that sort to avoid such a waste of CI cycles... If need be, I can squash those all up, otherwise I think were all set if Travis is happy.

bboutkov:
GRINS failure is due to PETSc updating the PC_MG_GALERKIN command line option from bool to an enum which was introduced in: https://github.com/petsc/petsc/commit/2134b1e4b80e7da7410bde701042d857987f373d 

Rather than try and play tricks with PETSc versions and options in the libMesh example run script I discussed with @pbauman and it seemslike the plan is to just boost Femputer PETSc version and older PETSc users can just use the older ` -pc_mg_galerkin` without the `both` for the PETSc options.

pbauman:
Job [GRINS-devel](https://femputer.eng.buffalo.edu:8000/job/3003/) on ad3ecbf : invalidated by @pbauman

Updated femputer petsc to 3.8.4. This should now run correctly with the new GMG foo.

pbauman:
Job [GRINS-dbg](https://femputer.eng.buffalo.edu:8000/job/3004/) on ad3ecbf : invalidated by @pbauman

Run with updated femputer petsc

pbauman:
Ack. I had femputer start using a debug build of PETSc instead of an opt one and now we're tripping PETSc errors in the Cahn-Hilliard example. @roystgnr that example is yours I presume?

pbauman:
Oh, the dbg recipe passes because I'm only running `make check` on libMesh in the devel recipe.

roystgnr:
No, misc ex7 was Dmitry Karpeev, not me.  I wonder if it's *ever* been run with a debug PETSc.  I'm ashamed to admit I might never have done so myself, but that seems like the sort of thing @jwpeterson has done from time to time?

jwpeterson:
> that seems like the sort of thing @jwpeterson has done from time to time?

If I did it was only locally on my workstation... we don't currently have an optional test for debugging-enabled PETSc on libmesh PRs on moosebuild, but it would probably be good to add. IIRC correctly that was the best/only way to catch code that tries to modify PETSc vectors that are locked read-only...

roystgnr:
Looks like the bug might only date back to #1986

pbauman:
@bboutkov Can you rebase on master and force push here? I believe #2066 should fix the libMesh issue and I believe I've fixed the femputer issue so we can have femputer give this another whirl.

pbauman:
Huzzah!

bboutkov:
Thanks for the help with those surrounding issues and the merge!

PR #2059, opened 2019-03-06
Fixes for greatly speeding up mesh stitching

friedmud:
Related to #1815

Takes the generation of a 2M element mesh from 138 seconds down to 43 seconds.  Will be an even bigger win on larger meshes!


friedmud:
This obviously went a bit far!  Going to back some of this out...

friedmud:
Nevermind - this isn't necessary.  All that's necessary is #1815

PR #2061, opened 2019-03-06
Skip partitioning in copy_nodes_and_elems if we should

roystgnr:
```skip_noncritical_partitioning()``` as well as the old "reset the partitioner pointer trick" should still have the old/safe behavior; "skip_partitioning()" has the literal-as-the-name dangerous/fast behavior.

It turns out that the refinement estimators were using the partitioner pointer trick, so hopefully this will fix the GRINS CI with no subsequent changes, although I really should update those eventually.

If this works it should be safe to merge and would supercede #1815.  I'm still begging for those test cases to replicate the original problem, though.  The correctness problem has to be due to a bug that I'd rather fix than hide, and the performance problem may be hurting us elsewhere even if we can avoid it here.

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -792,24 +792,47 @@ public:
   bool allow_remote_element_removal() const { return _allow_remote_element_removal; }
 
   /**
-   * If true is passed in then this mesh will no longer be
-   * (re)partitioned.  It would probably be a bad idea to call this on
-   * a DistributedMesh _before_ the first partitioning has
-   * happened... because no elements would get assigned to your
-   * processor pool.
+   * If true is passed in then the elements on this mesh will no
+   * longer be (re)partitioned, and the nodes on this mesh will only
+   * be repartitioned if they are found "orphaned" via coarsening or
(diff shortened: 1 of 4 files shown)


roystgnr:
Huh, it can't merge past the contrib/metaphysicl submodule change automatically?  Let me manually rebase on master then.

friedmud:
This looks good to me!  Thanks!

roystgnr:
Huh.  What's going on with that "Test debug" check?  It didn't even get started until 15 minutes ago?

Anyway, we'll merge when that's done... unless you want to run parallel apps tests too?  I wasn't actually able to find a good way to catch the "user mixes element deletion with skip_partitioning and thereby sometimes breaks libMesh" error, so it might be a good idea to double-check that nobody's actually doing that.

friedmud:
@roystgnr the regular "Test" target just "stopped" last night for some reason (it said "Cancelled").  I invalidated it this morning so it ran - then the debug testing fired off.

This PR really looks good to me.  There's no reason why this can't be merged!

friedmud:
Thanks for getting this in!

PR #2063, opened 2019-03-07
Fix left multiply

lindsayad:
I guess this is why one should always test

### DIFF
--- a/tests/numerics/type_tensor_test.C
+++ b/tests/numerics/type_tensor_test.C
@@ -6,6 +6,7 @@
 
 // libmesh includes
 #include <libmesh/tensor_value.h>
+#include <libmesh/vector_value.h>
 
 // THE CPPUNIT_TEST_SUITE_END macro expands to code that involves
 // std::auto_ptr, which in turn produces -Wdeprecated-declarations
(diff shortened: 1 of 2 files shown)


lindsayad:
@roystgnr @jwpeterson @pbauman? This one should be pretty easy...

tests/numerics/type_tensor_test.C
```diff
@@ -60,6 +62,18 @@ private:
         CPPUNIT_ASSERT_DOUBLES_EQUAL(inverse(i,j), true_inverse(i,j), 1.e-12);
   }
 
+  void testLeftMultiply()
+  {
+    TensorValue<Real> tensor(1, 2, 0, 3, 4, 0);
+    VectorValue<Real> vector(5, 6, 0);
+    auto left_mult = vector * tensor;
+    auto right_mult = tensor * vector;
+    CPPUNIT_ASSERT_DOUBLES_EQUAL(left_mult(0), 23, 1e-12);
```
jwpeterson: It looks like we have several hard-coded 1e-12 constants in our unit tests. They should probably all be changed to `TOLERANCE*TOLERANCE`, but I'm OK merging as-is.

PR #2064, opened 2019-03-11
Prefer PETSc dof_id sizes over implicitly set sizes

### LINKED ISSUE #2060: Detect dof-id-bytes
Right now libMesh will thrown an error if I compile PETSc with 64bit indices and try to use `--with-dof-id-bytes=4` (or vice versa).

It would be great if we didn't need to specify "dof-id-bytes" at all... and just have it be autodetected to use the same thing PETSc is using.

roystgnr:
If the user doesn't specify otherwise then we can default to PETSc's index size rather than default to 4.

This should resolve #2060

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -35,7 +35,7 @@ function check_autotool_prerequisites ()
 {
     # prerequisite versions
     automake_major=1
-    automake_minor=15
+    automake_minor=16
 
     autoconf_major=2
     autoconf_minor=69
--- a/configure
+++ b/configure
@@ -31539,8 +31539,10 @@ $as_echo "configuring size of boundary_id... $boundary_bytes" >&6; }
 # Check whether --with-dof_id_bytes was given.
 if test "${with_dof_id_bytes+set}" = set; then :
   withval=$with_dof_id_bytes; dof_bytes="$withval"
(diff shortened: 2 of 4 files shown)


bootstrap
```diff
@@ -35,7 +35,7 @@ function check_autotool_prerequisites ()
 {
     # prerequisite versions
     automake_major=1
-    automake_minor=15
+    automake_minor=16
```
jwpeterson: Nice catch, I'll make a note to do this the next time we change autotools versions.

roystgnr:
@pbauman, any idea what's going on with that regression/ozone_flame_antioch_kinetics_theory.sh divergence?  I'm trying to reproduce the libMesh example-failure-with-debug-PETSc now, but I'm not even sure where to begin on that GRINS failure.

roystgnr:
Anyway, going to ignore the unrelated CI problems and merge.

pbauman:
> any idea what's going on with that regression/ozone_flame_antioch_kinetics_theory.sh divergence?

No not sure what that's about. Will have a look at it tomorrow.

friedmud:
Sweet!  Thanks for implementing this @roystgnr and @jwpeterson !  Should be a large help!

PR #2066, opened 2019-03-11
Close residual/vector before applying constraints

roystgnr:
Otherwise debug-mode PETSc builds (rightly) scream at us for starting
set operations without finishing add operations.

This fixes a bug introduced in #1986 and caught when @pbauman updated his CI in #2057.


roystgnr:
This fixed the libMesh example bug in CI, and both the failing tests died at the same preexisting GRINS regression (which @pbauman now tells me he thinks he's figured out), so this should be good to merge.

PR #2069, opened 2019-03-13
Add is_zero to TypeVector and TypeTensor

aeslaughter:


jwpeterson:
Error during fetch and branch... did you guys change up the recipes recently?
```
error: unknown option `shallow-submodules'
```

include/numerics/tensor_tools.h
```diff
@@ -99,6 +99,10 @@ inline
 auto norm_sq(const VectorValue<T> & a) -> decltype(std::norm(T()))
 {return a.norm_sq();}
 
+template<typename T>
+inline
+auto is_zero(const T & a){ return a.is_zero();}
```
jwpeterson: Return type should just be `bool`. IIRC automatic return type deduction is a C++14 thing. The other places where it is used in this class are for trailing return types.

permcody:
Bet you wish you were here - we messed stuff up and we are at a "Full Stop" with CIVET right now! @#%@ is getting serious in here.

pbauman:
Heh, for once it's not femputer being the trouble maker.

permcody:
Yeah - we decided to throw in a few new options to "git clone" to keep our disk space usage down. Of course those options only work in newer version of git so all the clients just started dying on every job!

include/numerics/tensor_tools.h
roystgnr: You can do automatic return type deduction in C++11 but only with that awful redundant trailing arrow notation, not with just a single "auto".

roystgnr:
Looks good after the C++11 compatibility is fixed IMHO.

include/numerics/tensor_tools.h
aeslaughter: Good, I don't like all the auto stuff anyway.

PR #2076, opened 2019-03-18
FEInterface::get_continuity() should take a const reference.

jwpeterson:
This API was present in 1.4.0, so this commit should also be
cherry-picked onto the in-progress v1.4.1_branch.

Refs #2029.


roystgnr:
Whoops!  Can't believe I didn't notice that in the original PR.

PR #2078, opened 2019-03-19
Don't clear stashed DofConstraints

jwpeterson:
We need a way of switching back and forth between a "stashed" subset of the Constraints. There should probably eventually be an API for allowing users to do this, but for now simply not clearing the stashed constraints is sufficient for our needs...

### DIFF
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -889,7 +889,7 @@ void DofMap::clear()
 #ifdef LIBMESH_ENABLE_AMR
 
   _dof_constraints.clear();
-  //_stashed_dof_constraints.clear();
+  _stashed_dof_constraints.clear();
   _primal_constraint_values.clear();
   _adjoint_constraint_values.clear();
   _n_old_dfs = 0;
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -889,7 +889,7 @@ void DofMap::clear()
 #ifdef LIBMESH_ENABLE_AMR
 
   _dof_constraints.clear();
-  _stashed_dof_constraints.clear();
+  //_stashed_dof_constraints.clear();
   _primal_constraint_values.clear();
   _adjoint_constraint_values.clear();
   _n_old_dfs = 0;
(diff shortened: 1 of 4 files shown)


src/base/dof_map.C
```diff
@@ -889,7 +889,7 @@ void DofMap::clear()
 #ifdef LIBMESH_ENABLE_AMR
 
   _dof_constraints.clear();
-  _stashed_dof_constraints.clear();
+  //_stashed_dof_constraints.clear();
```
roystgnr: Surely we at least want to clear stashed constraints when we're clearing the whole DofMap, no?

src/reduced_basis/rb_construction.C
```diff
@@ -702,15 +708,12 @@ void RBConstruction::add_scaled_matrix_and_vector(Number scalar,
...
-      if (apply_dof_constraints)
-        {
-          // Apply constraints, e.g. Dirichlet and periodic constraints
-          this->get_dof_map().constrain_element_matrix_and_vector
-            (context.get_elem_jacobian(),
-             context.get_elem_residual(),
-             context.get_dof_indices(),
-             /*asymmetric_constraint_rows*/ false );
-        }
+      // Apply constraints, e.g. Dirichlet and periodic constraints
```
roystgnr: This comment should mention that Dirichlet constraints have been removed, if that's what the stashing has done, right?

roystgnr:
Would it make sense for there to simply be a swap_dof_constraints() API that switches between stashed and current constraints?  Changing the behavior of "stash" to "stash and replace with whatever's there already" and "unstash" to "unstash but also place current stuff there" seems confusing to me.

src/base/dof_map.C
jwpeterson: Yeah I think that one can go back in :)

jwpeterson:
> Would it make sense for there to simply be a swap_dof_constraints() API that switches between stashed and current constraints? Changing the behavior of "stash" to "stash and replace with whatever's there already" and "unstash" to "unstash but also place current stuff there" seems confusing to me.

It's true that this PR changes the "stash" API from a "temporarily delete and restore later" behavior to a "swap" behavior instead... so yeah I'll add that and keep the asserts in the stash API to maintain the existing behavior as well.

Issue #2079, opened 2019-03-20
Why not more mesh copy construction?


lindsayad:
Snippet of `MeshBase` copy constructor:
```c++
MeshBase::MeshBase (const MeshBase & other_mesh) :
  ParallelObject (other_mesh),
  boundary_info  (new BoundaryInfo(*this)),
  _n_parts       (other_mesh._n_parts),
  _is_prepared   (other_mesh._is_prepared),
  _point_locator (),
  _count_lower_dim_elems_in_point_locator(other_mesh._count_lower_dim_elems_in_point_locator),
  _partitioner   (),
#ifdef LIBMESH_ENABLE_UNIQUE_ID
  _next_unique_id(other_mesh._next_unique_id),
#endif
  _skip_noncritical_partitioning(false),
  _skip_all_partitioning(libMesh::on_command_line("--skip-partitioning")),
  _skip_renumber_nodes_and_elements(false),
  _allow_remote_element_removal(true),
  _elem_dims(other_mesh._elem_dims),
  _spatial_dimension(other_mesh._spatial_dimension),
  _default_ghosting(new GhostPointNeighbors(*this)),
  _ghosting_functors(other_mesh._ghosting_functors),
  _point_locator_close_to_point_tol(0.)
```
What is the reason that we don't copy many of the data member of `other_mesh`? I'm currently particularly interested in the `allow_remote_element_removal` flag but the general question stands. It might be nice to document reasons for why we don't copy construct some of these members because I'm sure there are good reasons for many/all of them that might be very instructive regarding the mesh construction process.

roystgnr:
> What is the reason that we don't copy many of the data member of other_mesh?

...

Answering the question about proximate reasons:  I'm embarrassed to say that I don't remember.  Looking through the git log suggests that those entries are 2.5y old or more, which reduces my embarrassment slightly.

_default_ghosting is a fixed thing (you change its effects by pulling it out of the ghosting functors list not by actually changing *it*) stored by unique_ptr to avoid sucking in more header dependencies, so a new initialization is equivalent to a deep copy there.

The _skip_noncritical_partitioning behavior was aping the previous _skip_partitioning behavior, and probably ought to be changed.  Likewise for the previous (now _skip_all_partitioning) behavior; it looks like I just changed that thoughtlessly when we added the command line option.

_point_locator_close_to_point_tol came from @jwpeterson so I'll let him comment.

_allow_remote_element_removal was me, and probably ought to be changed.

[...]

PR #2081, opened 2019-03-21
More copy construction in MeshBase

lindsayad:
Refs #2079


roystgnr:
:+1: but I feel like this should get more than just the default level of testing.  Would it be paranoid to run the MOOSE app tests and distributed mesh sweep before we merge?

lindsayad:
Nope!

lindsayad:
Segmentation fault! 😆

PR #2082, opened 2019-03-21
Add support for reading gmsh-4

lindsayad:
I suppose somebody's probably going to make me write a unit test 😢

### DIFF
--- a/src/mesh/gmsh_io.C
+++ b/src/mesh/gmsh_io.C
@@ -21,6 +21,7 @@
 #include <cstring> // std::memcpy
 #include <numeric>
 #include <unordered_map>
+#include <cstddef>
 
 // Local includes
 #include "libmesh/libmesh_config.h"
@@ -184,6 +185,11 @@ void GmshIO::read_mesh(std::istream & in)
   // note the the nodes can be non-consecutive
   std::map<unsigned int, unsigned int> nodetrans;
 
+  // Map from entity tag to physical id. The key is a pair with the first
+  // item being the dimension of the entity and the second item being
+  // the entity tag/id
+  std::map<std::pair<unsigned, int>, int> entity_to_physical_id;
+
   // For reading the file line by line
   std::string s;
 
@@ -199,7 +205,7 @@ void GmshIO::read_mesh(std::istream & in)
           if (s.find("$MeshFormat") == static_cast<std::string::size_type>(0))
             {
               in >> version >> format >> size;
-              if ((version != 2.0) && (version != 2.1) && (version != 2.2))
+              if (version < 2.0)
                 {
                   // Some notes on gmsh mesh versions:
                   //
@@ -214,6 +220,9 @@ void GmshIO::read_mesh(std::istream & in)
                   //
                   // Mesh version 2.2 tested by Manav Bhatia; no other
                   // libMesh code changes were required for support
+                  //
+                  // Mesh version 4.0 is a near complete rewrite of the previous mesh version
+                  //
                   libmesh_error_msg("Error: Unknown msh file version " << version);
                 }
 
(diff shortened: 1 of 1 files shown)


jwpeterson:
Damn, I guess the file format changed a lot, eh. I will try and take a look at this later tonight...

src/mesh/gmsh_io.C
```diff
@@ -275,10 +283,96 @@ void GmshIO::read_mesh(std::istream & in)
...
+                else if (num_physical_tags)
+                {
+                  in >> physical_tag;
+                  entity_to_physical_id[std::make_pair(1, curve_tag)] = physical_tag;
+                }
+
+                // Read to end of line; this captures bounding information that we don't care about
+                std::getline(in, s);
+              }
+              for (size_t n = 0; n < num_surface_entities; ++n)
```
jwpeterson: We should try to be consistent and always use `std::size_t`. We haven't always done this in the past, but I'm trying to be better about it. We should always be #includeing `<cstddef>` which puts that type into the std namespace, and some compilers are pickier about this than others...

src/mesh/gmsh_io.C
```diff
@@ -275,10 +283,96 @@ void GmshIO::read_mesh(std::istream & in)
...
+              in >> num_point_entities >> num_curve_entities >> num_surface_entities >> num_volume_entities;
+
+              for (size_t n = 0; n < num_point_entities; ++n)
+              {
+                int point_tag, physical_tag;
+                Real x, y, z;
+                size_t num_physical_tags;
+                in >> point_tag >> x >> y >> z >> num_physical_tags;
+                if (num_physical_tags > 1)
+                  libmesh_error_msg("I don't believe that we can specify multiple subdomain or " <<
```
jwpeterson: We definitely can't (currently) do this for the Elem::subdomain_id, but if we find ourselves erroring out a lot of files we might be able to handle this by storing the additional IDs in some other data structure that we could make available to the user.

Anyway I'd change the language of these error messages from "I don't believe we can" to simply "Cannot"!

src/mesh/gmsh_io.C
```diff
@@ -290,20 +384,62 @@ void GmshIO::read_mesh(std::istream & in)
...
+
+              size_t node_counter = 0;
+
+              // Now loop over entities
+              for (size_t i = 0; i < num_entities; ++i)
+              {
+                int entity_dim, entity_tag, parametric;
+                size_t num_nodes_in_block = 0;
+                in >> entity_dim >> entity_tag >> parametric >> num_nodes_in_block;
+                libmesh_assert_msg(!parametric, "We don't currently support reading parametric gmsh entities");
```
jwpeterson: Should this be a `libmesh_error_msg()` as well? If this is a parametric entity, presumably something else is going to fail later on and it's better to get a simple error now rather than a cryptic error later...

src/mesh/gmsh_io.C
```diff
@@ -275,10 +283,96 @@ void GmshIO::read_mesh(std::istream & in)
                 }
             }
 
+          else if (s.find("$Entities") == static_cast<std::string::size_type>(0))
+          {
+            if (version >= 4.0)
```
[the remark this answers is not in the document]
lindsayad: I don't believe there ever really was a mesh version 3. You can look at their documentation [here](http://gmsh.info/doc/texinfo/gmsh.html). They have documentation for legacy format mesh version 2 and 1, but there is nothing for a 3. There does seem to be some loose coupling between their code and mesh versioning, but it's clearly not exact. If I run `gmsh --version` for example I current get 4.2.3, but the generated `.msh` file lists version 4.1.

lindsayad:
Comments hopefully addressed

Issue #2084, opened 2019-03-26
VecScatterCreateWithData is not happening


jwpeterson:
@fdkong I think we can remove this ifdef for now. Based on the PETSc commits below, it looks like they changed their mind. What do you think? I will also backport the change to 1.4.1 which is in progress. 

Refs #2029.

```
commit 221b176569a4b92f5c39b90df909a2772efcd8db
Author: Junchao Zhang <jczhang@mcs.anl.gov>
Date:   Thu Mar 14 10:17:37 2019 -0500

    Update the document and dev.html

commit 9448b7f1ca3c3549e6b999e23e45403553ba4027
Author: Junchao Zhang <jczhang@mcs.anl.gov>
Date:   Mon Mar 18 09:22:25 2019 -0500

    Revert the function name where it is referenced

commit 130e142e396ec444a1c9967b4b57499e4af98d0a
Author: Junchao Zhang <jczhang@mcs.anl.gov>
Date:   Thu Mar 14 10:25:46 2019 -0500

    Restore the old VecScatterCreate() API
    
    The old interface VecScatterCreate(Vec,IS,Vec,IS,VecScatter*) is used by many applications. We do not want to break it.
```

PR #2086, opened 2019-03-26
Drop VecScatterCreateWithData() workaround.

jwpeterson:
This function did not wind up making it into any PETSc release.  This
should probably also be cherry-picked onto the v1.4.1 branch. Even
though the likelihood of any tagged libmesh release continuing to work
with PETSc master is negligibly small, the tagged release _might_ work
with a future PETSc 3.11 so we should try and support that if
possible.

Refs #2084.


fdkong:
It is fine with me.

PR #2088, opened 2019-03-26
DofMap::reinit_send_list(), Overlapping Coupling Test

pbauman:
DO NOT MERGE!

This is exposing a failing use case as a unit test to help decipher what's going on.

The first two commits are additions to the DofMap API to facilitate the following use case of the GhostingFunctor. (The physical context for the following is fluid-structure interaction.)

Suppose I have two subdomains that are not topologically connected, but overlap each other. I have different, subdomain-restricted variables declared on each subdomain. The variables on the subdomains are coupled. So I want to use the GhostingFunctor interface to express the algebraic ghosting and the coupling ghosting. However, the ghosting that I want to generate depends on the solution (think deformed solid). So I have to be able to evaluate the solution when the GhostingFunctor is going to be called. Which means that I can't attach the GhostingFunctor until after the System has been initialized. So the idea is to add the DofMap API so that I can attach the GhostingFunctor after the system is initialized, call DofMap::reinit_send_list(), and then reinitialize any ghosted vectors and the sparsity pattern (if doing coupling ghosting). 

In my application, I make the following observations:

1. If I pass an explicit CouplingMatrix, the algebraic ghosting appears to be incorrect.
2. If I pass a null CouplingMatrix, the algebraic ghosting does not throw an error, but I get incorrect results after a linear solve for certain processor counts.

So this WIP explicitly shows issue 1. (I'm hoping if we fix issue 1, issue 2 will also magically get fixed). It fails on the simple 3 element mesh I construct by hand in the test and on 2 (or more) processors. Everything passes on one processor.

The unit test is a little involved (I need the GhostingFunctor mimicing my use case, a Partitioner to make sure it will always trip the algebraic ghosting error by making sure the subdomains always sit on different processors (which gets its own unit testing) and then the actual testing that triggers failures), but it does mimic my use case pretty closely. I'll enrich it a bit more and can clean it up, but wanted to get this put up now so that it at least exposes the failing case.

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -440,6 +440,16 @@ public:
   void prepare_send_list ();
 
   /**
+   * Clears the \p _send_list vector. This should be done in order to completely
+   * rebuild the send_list from scratch rather than merely adding to the existing
+   * send_list.
+   */
+  void clear_send_list ()
+  {
+    _send_list.clear();
+  }
+
+  /**
    * \returns A constant reference to the \p _send_list for this processor.
    *
    * The \p _send_list contains the global indices of all the
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -450,6 +450,18 @@ public:
   }
 
   /**
+   * Clears the \p _send_list vector and then rebuilds it. This may be needed
+   * in special situations, for example when an algebraic coupling functor cannot
+   * be added to the \p DofMap until after it is completely setup. Then this method
+   * can be used to rebuild the send_list once the algebraic coupling functor is
+   * added. Note that while this will recommunicate constraints with the updated
+   * send_list, this does assume no new constraints have been added since the previous
+   * reinit_constraints call.
+   */
+  void reinit_send_list (MeshBase & mesh);
+
+
+  /**
    * \returns A constant reference to the \p _send_list for this processor.
    *
    * The \p _send_list contains the global indices of all the
--- a/include/base/ghosting_functor.h
+++ b/include/base/ghosting_functor.h
@@ -130,6 +130,23 @@ class Elem;
  * (e.g. subdomain-restricted variables on a neighboring subdomain)
  * will be unaffected.
  *
+ * Typical usage of the GhostingFunctor would be to add a geometric ghosting
+ * functor before the mesh preparation is completed; progmatically, this would
+ * be before MeshBase::prepare_for_use() is called, but many different libMesh
+ * idioms internally call this function. The algebraic and coupling ghosting
+ * functors normally are added before EquationSystems::init() is called.
+ * However, in some circumstances, solution evaluation may be needed within the
+ * GhostingFunctor in order to determine the ghosting, in which case the appropriate
+ * functor would need to be added after EquationSystems::init(). In this case,
+ * the user will need to reinitialize certain parts of the DofMap for
+ * algebraic and coupling functors. For algebraic ghosting functors, the
+ * user will need to call DofMap::reinit_send_list() and then reinitialize
+ * any NumericVectors that are GHOSTED, e.g. the System::current_local_solution.
+ * For coupling ghosting, the user will also need to recompute the sparsity
+ * pattern via DofMap::clear_sparsity() and then DofMap::compute_sparsity() and
+ * then reinitialize any SparseMatrix objects attached to the System, e.g.
+ * the system.get_matrix("System Matrix").
+ *
  * \author Roy H. Stogner
  * \date 2016
  */
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -1678,6 +1678,17 @@ void DofMap::prepare_send_list ()
   std::vector<dof_id_type> (_send_list.begin(), new_end).swap (_send_list);
 }
 
+void DofMap::reinit_send_list (MeshBase & mesh)
+{
+  this->clear_send_list();
+  this->add_neighbors_to_send_list(mesh);
+
+  // This is assuming that we only need to recommunicate
+  // the constraints and no new ones have been added since
+  // a previous call to reinit_constraints.
+  this->process_constraints(mesh);
+  this->prepare_send_list();
+}
 
 void DofMap::set_implicit_neighbor_dofs(bool implicit_neighbor_dofs)
 {
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -882,7 +882,7 @@ void DofMap::clear()
   _first_df.clear();
   _end_df.clear();
   _first_scalar_df.clear();
-  _send_list.clear();
+  this->clear_send_list();
   this->clear_sparsity();
   need_full_sparsity_pattern = false;
 
@@ -935,7 +935,7 @@ void DofMap::distribute_dofs (MeshBase & mesh)
   dof_id_type next_free_dof = 0;
 
   // Clear the send list before we rebuild it
-  _send_list.clear();
+  this->clear_send_list();
 
   // Set temporary DOF indices on this processor
   if (node_major_dofs)
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -24,6 +24,7 @@ unit_tests_sources = \
   base/default_coupling_test.C \
   base/getpot_test.C \
   base/point_neighbor_coupling_test.C \
+  base/overlapping_coupling_test.C \
   fe/fe_bernstein_test.C \
   fe/fe_clough_test.C \
   fe/fe_hermite_test.C \
--- a/tests/Makefile.in
+++ b/tests/Makefile.in
@@ -185,7 +185,8 @@ PROGRAMS = $(noinst_PROGRAMS)
 am__unit_tests_dbg_SOURCES_DIST = driver.C test_comm.h \
 	stream_redirector.h base/dof_object_test.h base/dof_map_test.C \
 	base/default_coupling_test.C base/getpot_test.C \
-	base/point_neighbor_coupling_test.C fe/fe_bernstein_test.C \
+	base/point_neighbor_coupling_test.C \
+	base/overlapping_coupling_test.C fe/fe_bernstein_test.C \
 	fe/fe_clough_test.C fe/fe_hermite_test.C \
 	fe/fe_hierarchic_test.C fe/fe_l2_hierarchic_test.C \
 	fe/fe_l2_lagrange_test.C fe/fe_lagrange_test.C \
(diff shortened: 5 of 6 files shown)


pbauman:
Annnnddd I realize now that the reason it the case was failing was that the FEMContext will try and build every variable defined on the element, not just the ones I asked for coupling with. I'll tinker with this some more later this week to make the testing more precise to see if I can figure why my solutions are crap in the application and reflect that in this test. Then can talk about merging.

src/base/dof_map.C
```diff
@@ -1678,6 +1678,13 @@ void DofMap::prepare_send_list ()
   std::vector<dof_id_type> (_send_list.begin(), new_end).swap (_send_list);
 }
 
+void DofMap::reinit_send_list (MeshBase & mesh)
+{
+  this->clear_send_list();
+  this->add_neighbors_to_send_list(mesh);
+  this->add_constraints_to_send_list();
```
pbauman: OK, I finally got a working case. Most of it was user error, but this part was incomplete. If I changed this to `scatter_constraints(mesh)` it worked for my case (the failure manifest as an incorrect solve for the Newton step. Not sure how we're going to unit test that...). But I think that's a really special case and I think the correct thing to do here call `process_constraints()`. However, that assumes the user hasn't added any new constraints and they've just changed the ghosting. I will certainly document that, but do we want to do something else, e.g. have a `reinit_constraints_and_send_list()` method that would call `reinit_constraints()` instead of just `process_constraints`? Thoughts @roystgnr? Once we settle this, I'll clean this up for merging.

pbauman: Oops, I said the wrong thing. The first alternative I tried was

<pre>
this->scatter_constraints(mesh);
this->add_constraints_to_send_list();
</pre>

i.e. added the `scatter_constraints(mesh)` call. Then, I replaced both of those with `this->process_constraints(mesh)`. Both work for my case, but I think the latter is more complete. My case only needs to send the constraints to other processors, not gather them.

roystgnr: > the failure manifest as an incorrect solve for the Newton step. Not sure how we're going to unit test that...

Test the residual magnitude after enough steps that we should be within tolerance?  Quadratic vs non-quadratic convergence should be different enough that we can be loose with tolerances.

> do we want to do something else, e.g. have a reinit_constraints_and_send_list() method that would call reinit_constraints() instead of just process_constraints?

That might not be a bad idea, but I'd want to see a use case first.  Usually people who have constraints turning on and off on the fly are going to have inequality constraints (which will have to be handled via the formulation) rather than equality constraints (which can go into our current infrastructure), right?

pbauman: > That might not be a bad idea, but I'd want to see a use case first.

Heh, that's a good point. I'm not sure I can come up with one. I'll stick with just this for now and document it and if the need ever arises, we can add that API for it.

pbauman:
OK, this is ready for formal review, I think. I added some documentation to the `GhostingFunctor` in the last commit to try and spell out what needs to be done in use cases like this one. Please also check  that the documentation for `DofMap::reinit_send_list` is adequate. 

The unit testing is necessary, but not sufficient. It won't capture the error I was seeing when we didn't call `DofMap::process_constraints()`, but I'm hoping that a future GRINS test will capture that. This unit test does catch incomplete algebraic ghosting and incomplete sparsity pattern updates.

include/base/ghosting_functor.h
```diff
@@ -130,6 +130,20 @@ class Elem;
  * (e.g. subdomain-restricted variables on a neighboring subdomain)
  * will be unaffected.
  *
+ * Typical usage of the GhostingFunctor would be to add the appropriate
+ * functor before EquationSystems::init() is called. However, in some
```
roystgnr: This is actually misleading - algebraic and coupling functors should be added before EquationSystems::init(), but geometric ghosting functors are ideally added before the Mesh is filled.

include/base/dof_map.h
```diff
@@ -439,6 +439,28 @@ class DofMap : public ReferenceCountedObject<DofMap>,
...
+   * send_list.
+   */
+  void clear_send_list ()
+  {
+    _send_list.clear();
+  }
+
+  /**
+   * Clears the \p _send_list vector and then rebuilds it. This may be needed
+   * for special situations for example when an algebraic coupling functor cannot
```
roystgnr: As long as I'm making you change comments: comma after "situations"

include/base/ghosting_functor.h
pbauman: > geometric ghosting functors are ideally added before the Mesh is filled.

Meaning before `prepare_for_use()`? Sorry, I'm only messing with the algebraic/coupling functors so I'm less sure about the geometry functors.

roystgnr: That's the right meaning, but I'm not sure of a better way to say it than "filled", since libMesh mesh generation idioms typically do prepare_for_use() implicitly.

pbauman:
I pushed two comment commits that I'll squash in once someone takes a look. Good to merge after that?

roystgnr:
I added a DistributedMesh sweep to be paranoid; :+1: once that's done, thanks!

roystgnr:
Well, yay for paranoia.

pbauman:
OK, it was because I didn't use the replicated mesh header, just `mesh.h`. I'll squash that fix in (after testing with --enable-parmesh locally) and squash in those documentation commits and merge once all CIVET's are happy.

As an aside, I do plan to come back to the unit in the nearish future to update to support `DistributedMesh` (i.e. figure out how to tinker with geometric ghosting) once we have our use case fully up and running on `ReplicatedMesh`.

pbauman:
I cannot locally reproduce the segfault the distributed mesh test is showing.

pbauman:
Oops, I lied (missed the mpiexec -np 2 part). I can reproduce.

pbauman:
Hmmm. This doesn't look like me anymore. Here's the output when running the devel unit test:

<pre>
[02:18:02][pbauman@fry:/fry1/data/users/pbauman/research/libmesh/build/clear-send-list/tests]$ mpiexec -np 2 ./unit_tests-devel --option-with-dashes --option_with_underscores 3
Will run the following tests:
All Tests
.....Constraint loop detected
Constraint loop detected
[0] ../../pbauman/src/base/dof_map_constraints.C, line 3482, compiled Apr  5 2019 at 14:07:33
[1] ../../pbauman/src/base/dof_map_constraints.C, line 3482, compiled Apr  5 2019 at 14:07:33
....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................Assertion `ghost_objects_from_proc[p] <= objects_on_proc[p]' failed.
ghost_objects_from_proc[p] = 5
objects_on_proc[p] = 0


[1] ../../pbauman/src/mesh/distributed_mesh.C, line 1054, compiled Apr  5 2019 at 13:44:14
Assertion `(comm).verify(std::string("./include/libmesh/parallel_sync.h").size())' failed.

Assertion `(this->comm()).verify(std::string("../../pbauman/src/mesh/mesh_base.C").size())' failed.

[0] ./include/libmesh/parallel_sync.h, line 245, compiled Apr  5 2019 at 13:44:14
E.[1] ../../pbauman/src/mesh/mesh_base.C, line 160, compiled Apr  5 2019 at 14:07:44
Assertion `this->comm().verify(unpartitioned_objects)' failed.

Assertion `this->comm().verify(unpartitioned_objects)' failed.

[0] ../../pbauman/src/mesh/distributed_mesh.C, line 1051, compiled Apr  5 2019 at 13:44:14
E.[1] ../../pbauman/src/mesh/distributed_mesh.C, line 1051, compiled Apr  5 2019 at 13:44:14
Assertion `this->comm().verify(unpartitioned_objects)' failed.

Assertion `this->comm().verify(unpartitioned_objects)' failed.

[0] ../../pbauman/src/mesh/distributed_mesh.C, line 1051, compiled Apr  5 2019 at 13:44:14
E.[1] ../../pbauman/src/mesh/distributed_mesh.C, line 1051, compiled Apr  5 2019 at 13:44:14
Assertion `this->comm().verify(unpartitioned_objects)' failed.

Assertion `this->comm().verify(unpartitioned_objects)' failed.

[0] ../../pbauman/src/mesh/distributed_mesh.C, line 1051, compiled Apr  5 2019 at 13:44:14
E.[1] ../../pbauman/src/mesh/distributed_mesh.C, line 1051, compiled Apr  5 2019 at 13:44:14
Assertion `this->comm().verify(unpartitioned_objects)' failed.

Assertion `this->comm().verify(unpartitioned_objects)' failed.

[0] ../../pbauman/src/mesh/distributed_mesh.C, line 1051, compiled Apr  5 2019 at 13:44:14
E.[1] ../../pbauman/src/mesh/distributed_mesh.C, line 1051, compiled Apr  5 2019 at 13:44:14
In CheckpointIO::write, directory 'checkpoint_splitter.cpa/2' already exists, overwriting contents.../../pbauman/src/mesh/checkpoint_io.C, line 123, compiled Apr  5 2019 at 13:44:13 ***
Assertion `unique_id == old_node->unique_id()' failed.
unique_id = 2
old_node->unique_id() = 20


[0] ../../pbauman/src/mesh/checkpoint_io.C, line 1003, compiled Apr  5 2019 at 13:44:13
E.Assertion `(this->comm()).verify(std::string("../../pbauman/src/mesh/mesh_base.C").size())' failed.

Assertion `(mesh.comm()).verify(std::string("../../pbauman/src/mesh/mesh_communication.C").size())' failed.

[0] ../../pbauman/src/mesh/mesh_base.C, line 160, compiled Apr  5 2019 at 14:07:44
E.[1] ../../pbauman/src/mesh/mesh_communication.C, line 1101, compiled Apr  5 2019 at 14:07:47
........Warning:  This MeshOutput subclass only supports meshes which have been serialized!
Warning:  This MeshOutput subclass only supports meshes which have been serialized!
.....................Warning!  Parallel xda/xdr is not yet implemented.
Writing a serialized file instead.
........................................................................................................................................................................................................ERROR: The library has been built without
Space Filling Curve support.  Using a linear
partitioner instead!
.....................................................................................................................................................................................ERROR:  Unsupported PETSC Solver: ERROR:  Unsupported PETSC Solver: JACOBI
Continuing with PETSC defaults
JACOBI
Continuing with PETSC defaults
...........


!!!FAILURES!!!
Test Results:
Run:  1173   Failures: 0   Errors: 7


1) test: BoundaryRefinedMeshTest::testMesh (E) 
setUp() failed
- uncaught exception of type libMesh::LogicError
- 


2) test: BoundaryMeshTest::testMesh (E) 
setUp() failed
- uncaught exception of type libMesh::LogicError
- 


3) test: BoundaryInfoTest::testMesh (E) 
- 


4) test: BoundaryInfoTest::testShellFaceConstraints (E) 
- 


5) test: BoundaryInfoTest::testEdgeBoundaryConditions (E) 
- 


6) test: CheckpointIOTest::testAsciiDistRepSplitter (E) 
- 


7) test: CheckpointIOTest::testBinaryDistRepSplitter (E) 
- 
</pre>

pbauman:
I'm trying with master now to see if it's there as well.

roystgnr:
Yeah, I'm pretty sure it'll be there.  I think this is what I'm seeing in #1938 too.  At least one of the bugs appears to be an effect of making the changes in #2061 without changing some of the internal ```skip_repartitioning()``` uses to ```skip_noncritical_repartitioning()```

roystgnr:
I've got at least one case fixed; I'll put in a new PR with that and see what else we hit.

pbauman:
OK, just confirmed I'm seeing this in the current master.

pbauman:
OK, rebased on master after #2097 was merged. Once everything passes, I'll go ahead and merge.

Issue #2089, opened 2019-03-28
Add DualNumber explicit instantiations to libmesh numerical types if MetaPhysicL is enabled


dschwen:
I stumbled across this when I was unable to use `DenseMatrix<ADReal>` in MOOSE. I'll try to figure out a solution for this. Ping @lindsayad

dschwen:
argh, of course that does not work because the proper type is only known in MOOSE

dschwen:
maybe we can factor out a `dense_matrix_base_impl.h` file...

jwpeterson:
Closing because we do now have a dense_matrix_base_impl.h file, and I think there are more recent open PRs/issues on the same topic.

roystgnr:
Yeah; MOOSE now has the explicit instantiations in src/utils/DenseMatrixDualReal.C, which is probably the optimal way to do things until/unless we have other users clamoring to move the instantiations to libMesh proper to (slightly) reduce their own compile times.

PR #2090, opened 2019-03-28
Factor out a DenseMatrixBase implementation header

dschwen:
Refs #2089 (there might be more cases where this is required to enable external apps to create new template instantiations - for example with dual numbers)


dschwen:
Darn, that header doesn't get installed. I must have missed something...

roystgnr:
Did you run include/rebuild_include_HEADERS.sh?

roystgnr:
Ah, looks like you got it now.

dschwen:
Uhm, no, but I edited libmesh/include/include_HEADERS

dschwen:
Let me make teh change using that script you mentioned to keep the diffs clean going forward

roystgnr:
Ha, I was wondering how there ended up being a white space change in the diff!  Your hand-written changes were better than the script's version...

dschwen:
As an aside, will there ever be a unity build for libmesh... :-) ?

dschwen:
ARGH!!! This is STILL not getting installed!!

jwpeterson:
@dschwen, in case you did not see it? https://github.com/libMesh/libmesh/wiki/Adding-or-removing-source-files

dschwen:
@jwpeterson that link helps. I ran 

```
cd include && ./rebuild_include_HEADERS.sh
cd ..
cd src && ./rebuild_libmesh_SOURCES.sh
cd ..
cd include/libmesh && ./rebuild_makefile.sh
cd -
./bootstrap
```

Bit the git diff showed no changes, so I guess I should be fine. Yet the Test/InstallCheck is failing (and I broke MOOSE test?! -  `petsc.h` not found?!).

What. Up?

jwpeterson:
> What. Up?

There is an issue with CIVET today, I believe @milljm and @permcody are working on that.

dschwen:
DenseMatrix needs some `auto` or `T` in place of `Real` to enable templating on DualNumbers

permcody:
New package release on CIVET today. I'm helping @milljm iron out some of the issues that arise when there are competing packages. We have merged in a new change that should help with stability in this situation. To be clear, we haven't deployed the package officially yet with MOOSE, but it's available to CIVET and being picked up _before_ the repositories are cloned/cleaned (e.g. during pre-check).

jwpeterson:
> DenseMatrix needs some auto or T in place of Real to enable templating on DualNumbers

@lindsayad did some of this already I thought?

dschwen:
Thinks like `Real DenseMatrix::min()` are biting me in the behind here. I want the return type to be 

```
  auto min () const -> decltype(libmesh_real((*this)(0,0)));
```

But that doesn't work because `definition of 'libMesh::DenseMatrix<double>' is not complete until the closing '}'`

dschwen:
This seems to work:

```
auto DenseMatrix<T>::min () const -> decltype(libmesh_real(T(0)))
```

jwpeterson:
OK... I guess I don't get why the `min` should be real-valued if T==std::complex

dschwen:
Yeah, me neither, but I'm not eager to break behavior that somebody may rely on.

jwpeterson:
It's been like that since 2006.... be1f873f3dec1fbd92dc45a3b27f13172de58f61, the comments suggest  @roystgnr wanted it that way at the time... he may have been thinking in terms of matrix norms I guess.

dschwen:
Ok, builds now and allows me to use DenseMatrix<DualReal> in my app.

roystgnr:
I'd guess that my only thought was that having min() return min(real()) was likely to be more useful than having it return NaN or throw an error.  But in hindsight "more useful" is also "less safe" so I wouldn't object if someone wanted to change that behavior.

include/numerics/dense_matrix.h
```diff
@@ -1015,17 +1015,17 @@ Real DenseMatrix<T>::l1_norm () const
...
 template<typename T>
 inline
-Real DenseMatrix<T>::linfty_norm () const
+auto DenseMatrix<T>::linfty_norm () const -> decltype(std::abs(T(0)))
 {
   libmesh_assert (this->_m);
   libmesh_assert (this->_n);
 
-  Real rowsum = 0.;
+  decltype(std::abs(T(0))) rowsum = 0.;
```
jwpeterson: Just curious, does `auto rowsum = std::abs(T(0));` work to initialize this, or do you really need `decltype` here?

dschwen: Yeah, I'm pretty sure that would work. It looks obvious now, but I went through so many iteration with the decltype arguments that initially I didn't think of that. I'll make the change.

lindsayad:
@dschwen How did you get this working in your app? I'm getting undefined references with respect to `DenseMatrix::_multiply_blas` upon instantiation of `DenseMatrix<ADReal>`

lindsayad:
I’m thinking of including the impl files, throwing in some empty voids for the blas lapack private methods, and then explicitly instantiating on ADReal in a MOOSE .C file to get this done. That should work right?

lindsayad:
Well with error messages instead of empty voids...and will have to make sure to manually set the use_blas_lapack flag to false.

dschwen:
If you have Marmot access check the `deep_neural_network_318` branch on my fork.

dschwen:
Also https://github.com/idaholab/moose/pull/13294. I figured I might have to add stubs for all the blas crap. But so far I have not run into this issue, and I have already successfully run simulations using  `DenseMatrix<DualReal>`...

lindsayad:
I am guessing that this can only be on a Mac which does lazy symbol binding...? I get undefined references when linking on linux for the blas/lapack stuff

PR #2091, opened 2019-03-28
Ignore one more warning with GCC 8.x compilers

permcody:
The trigger comes from BOOST distributions

### DIFF
--- a/include/utils/ignore_warnings.h
+++ b/include/utils/ignore_warnings.h
@@ -82,6 +82,9 @@
 #if (__GNUC__ > 6)
 #pragma GCC diagnostic ignored "-Wint-in-bool-context"
 #pragma GCC diagnostic ignored "-Wimplicit-fallthrough"
+#if (__GNUC__ > 8)
+#pragma GCC diagnostic ignored "-Wparentheses"
+#endif // GCC > 8
 #endif // GCC > 6
 #endif // GCC > 5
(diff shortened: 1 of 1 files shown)


include/utils/ignore_warnings.h
```diff
@@ -82,6 +82,9 @@
 #if (__GNUC__ > 6)
 #pragma GCC diagnostic ignored "-Wint-in-bool-context"
 #pragma GCC diagnostic ignored "-Wimplicit-fallthrough"
+#if (__GNUC__ > 8)
```
jwpeterson: Does `> 8` mean GCC 9.x?

permcody:
Oh no! That’s not right. This is for the GCC 8 compiler. The error went
away on my local test because I didn’t wrap it in the if statement. I’ll
fix it later unless you just want to do it. Nice catch!

jwpeterson:
Sure, I can just change it to `> 7`. I can't test it conveniently since I don't have a GCC 8.x readily available, so you'll have to let me know if it still works.

jwpeterson:
BTW, it looks like this warning has been around for a while (e.g. [it's mentioned in these GCC 3.4.4 docs](https://gcc.gnu.org/onlinedocs/gcc-3.4.4/gcc/Warning-Options.html)). I guess the ability to disable it was what was only recently added?

PR #2093, opened 2019-04-02
Remove redundant (and shadowing) fe

roystgnr:
This fixes a -Wshadow warning for me

### DIFF
--- a/examples/miscellaneous/miscellaneous_ex14/miscellaneous_ex14.C
+++ b/examples/miscellaneous/miscellaneous_ex14/miscellaneous_ex14.C
@@ -561,12 +561,6 @@ void assemble_SchroedingerEquation(EquationSystems &es, const std::string &syste
    }
    // loop over NEIGHBOURS OF INFINITE ELEMENTS
    sd=2; {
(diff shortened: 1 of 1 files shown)


roystgnr:
Looks like this slipped through CI because "Test with complex numbers and infinite elements enabled" doesn't have PETSc enabled, much less SLEPc?  We should probably fix that if we can...

jwpeterson:
We did this on purpose because there is no complex-enabled PETSc available in CIVET/the MOOSE package. If they decide to add one, we should definitely add a complex-PETSc target.

I actually kind of like this test target the way it is because it tests Eigen... and complex-valued Eigen at that, something that would get no coverage otherwise.

jwpeterson:
BTW, I checked but this shadowing error isn't in v1.4.0, (misc_ex14 was updated after we branched) so this does not need to be cherry-picked.

roystgnr:
> I actually kind of like this test target the way it is because it tests Eigen... and complex-valued Eigen at that, something that would get no coverage otherwise.

I forgot about that; good point.

PR #2097, opened 2019-04-05
Only skip noncritical partitioning in BoundaryMesh

roystgnr:
Under the new skip_partitioning semantics, on a DistributedMesh we
don't appear to be getting even *initial* nodal partitioning into a
valid state.

This is one of (and if we're really really lucky the only one of) the bugs that seems to be throwing current DistributedMesh sweeps for a loop, foiling #1938 and #2088 tests.

While the initial CI sweeps are running on this fix, I'll start digging for any other ```skip_repartitioning()``` uses that need to be converted to ```skip_noncritical_repartitioning()``` to conform to the change in #2061


pbauman:
Hey look at that! Looks like you got ...

> really really lucky 

:)

PR #2099, opened 2019-04-09
Subdomain-restricted variables hanging nodes fix

roystgnr:
I caught this while working on more unit test coverage for #1938, and it turned out to be a bug in master, not just in the rewrite in that PR.  Previously, when an element with a subdomain-restricted variable had a coarser neighbor on which that variable was not supported, garbage constraint equations would be generated.  This PR adds asserts which would catch similar bugs earlier and adds a fix for the bug.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -1360,6 +1360,12 @@ void DofMap::add_constraint_row (const dof_id_type dof_number,
     if (this->is_constrained_dof(dof_number))
       libmesh_error_msg("ERROR: DOF " << dof_number << " was already constrained!");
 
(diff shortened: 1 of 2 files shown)


jwpeterson:
Ignoring whitespace changes the diff isn't all that big, so I'll try and backport to 1.4.1 (#2029) as well?

roystgnr:
I'd definitely like that; thanks!  This is the fix I told you was coming for a "horribly embarrassing" bug - two heavily used features which break intermittently if used together!?!  And it was just dumb luck that I got a nice error message to track down; since the OOB access *didn't* cause a segfault, I fear the breakage could even cause corrupted data in the worst case.

PR #2100, opened 2019-04-09
Fix 2006 bug with refinement within ES::reinit()

roystgnr:
This was part of a half-baked attempt to support both "refine and
coarsen prior to EquationSystems::reinit for CPU efficiency" and
"refine and coarsen only within EquationSystems::reinit for memory
efficiency" idioms, but the latter support was seriously broken.

The "optimization" being removed here didn't actually optimize any use cases, but did break some.

This is another unrelated bug caught by the unit tests I'm working on for #1938.

### DIFF
--- a/src/systems/equation_systems.C
+++ b/src/systems/equation_systems.C
@@ -183,7 +183,6 @@ bool EquationSystems::reinit_solutions ()
 
 #ifdef LIBMESH_ENABLE_AMR
 
-  bool dof_constraints_created = false;
   bool mesh_changed = false;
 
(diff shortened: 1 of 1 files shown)


jwpeterson:
:+1: looks like a good candidate for backporting to v1.4.1 (#2029) as well.

roystgnr:
Agreed; clearly nobody is actually *using* this capability heavily but it would be nice to have it work, it fails in confusing ways when it doesn't work, and it's a really simple fix.

PR #2102, opened 2019-04-16
MeshBase::add_elem_integer() API

### LINKED ISSUE #2095: Add two integers to Elem
We have a need for two `unsigned int`s to get added to `Elem`.  I propose: `user_id1` and `user_id2`.  This will be controlled with `--enable-extra-elem-ids` at configure time (and configured to `false` by default).  These IDs will be user settable - but otherwise will act just like `subdomain_id` and `unique_id`.

This is something that we've had to engineer around for a long time in one of our applications (Rattlesnake) and there's just no reason for it.

Eventually maybe https://github.com/libMesh/libmesh/issues/1439 will obviate the need for this (maybe)... but there's no reason to wait for that.

roystgnr:
This could use a few more convenience APIs and overloads (looking up an already-added index by name, adding multiple integers in one call), but the basic functionality ought to be sufficiently functional to close #2095, and getting the sneaky _idx_buf work correct was difficult enough that I want to get CI (and @YaqiWang and @friedmud if they have time) taking a look at it right away.

It was passing my own tests fine just a minute ago, so hopefully my interactive rebase cleanup didn't screw anything up and it'll be ready to merge when moosebuild is happy too.

### DIFF
--- a/tests/mesh/extra_integers.C
+++ b/tests/mesh/extra_integers.C
@@ -46,6 +46,8 @@ protected:
 
     // Request some extra integers before building
     unsigned int i1 = mesh.add_elem_integer("i1");
+    unsigned int ni1 = mesh.add_node_integer("ni1");
+    unsigned int ni2 = mesh.add_node_integer("ni2");
 
     MeshTools::Generation::build_line(mesh,
                                       /*nx=*/10,
@@ -60,18 +62,34 @@ protected:
         CPPUNIT_ASSERT_EQUAL(elem->get_extra_integer(i1), dof_id_type(elem->point(0)(0)*100));
       }
 
+    for (const auto & node : mesh.node_ptr_range())
+      {
+        CPPUNIT_ASSERT_EQUAL(node->n_extra_integers(), 2u);
+        CPPUNIT_ASSERT_EQUAL(node->get_extra_integer(ni1), DofObject::invalid_id);
+        CPPUNIT_ASSERT_EQUAL(node->get_extra_integer(ni2), DofObject::invalid_id);
+      }
+
     // Force (in parallel) a different partitioning - we'll simply put
     // everything on rank 0, which hopefully is not what our default
     // partitioner did!
     mesh.partition(1);
 
     CPPUNIT_ASSERT_EQUAL(i1, mesh.add_elem_integer("i1"));
+    CPPUNIT_ASSERT_EQUAL(ni1, mesh.add_node_integer("ni1"));
+    CPPUNIT_ASSERT_EQUAL(ni2, mesh.add_node_integer("ni2"));
 
     // Make sure we didn't screw up any extra integers thereby.
     for (const auto & elem : mesh.element_ptr_range())
       {
         CPPUNIT_ASSERT_EQUAL(elem->n_extra_integers(), 1u);
         CPPUNIT_ASSERT_EQUAL(elem->get_extra_integer(i1), dof_id_type(elem->point(0)(0)*100));
+      }
+
+    for (const auto & node : mesh.node_ptr_range())
+      {
+        CPPUNIT_ASSERT_EQUAL(node->n_extra_integers(), 2u);
+        CPPUNIT_ASSERT_EQUAL(node->get_extra_integer(ni1), DofObject::invalid_id);
+        CPPUNIT_ASSERT_EQUAL(node->get_extra_integer(ni2), DofObject::invalid_id);
       }
 
 #ifdef LIBMESH_ENABLE_AMR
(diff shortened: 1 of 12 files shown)


pbauman:
`-Werror` FTW:

<pre>
In file included from ./include/libmesh/dof_object.h:26:0,
                 from /femputer/pbauman/civet_build_testing_root/libMesh/src/base/dof_object.C:23:
/femputer/pbauman/civet_build_testing_root/libMesh/src/base/dof_object.C: In member function ‘void libMesh::DofObject::unpack_indexing(std::__debug::vector<long unsigned int>::const_iterator)’:
/femputer/pbauman/civet_build_testing_root/libMesh/src/base/dof_object.C:597:61: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
                  (std::abs(dof_id_signed_type(_idx_buf[0])) <= _idx_buf.size()));
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
./include/libmesh/libmesh_common.h:276:11: note: in definition of macro ‘libmesh_assert_msg’
     if (!(asserted)) {                                                  \
           ^~~~~~~~
/femputer/pbauman/civet_build_testing_root/libMesh/src/base/dof_object.C:596:3: note: in expansion of macro ‘libmesh_assert’
   libmesh_assert(_idx_buf.empty() ||
   ^~~~~~~~~~~~~~
</pre>

roystgnr:
Wow, I really need to get in the habit of always turning that on myself.  Thanks!

roystgnr:
In my defense, this instance of -Wsign-compare is pretty clearly overkill.  std::abs() of a signed type will always fit in the corresponding unsigned type.

roystgnr:
On the bright side, since I had a bug to fix anyway I figured I might as well add another feature, and since I was adding another feature I figured I might as well add more new unit tests for it, and while I was adding more new unit tests I discovered multiple horrible bugs in the previous new unit tests, so that'll be nice to fix before committing.

YaqiWang:
Thanks @roystgnr. I want double check one thing: I can call `set_extra_integer` when I add an element in the mesh during mesh generation, providing that I call `add_elem_integer` in the mesh beforehand, right?

roystgnr:
That depends on your definition of "when": the Mesh allocates extra integer space when you call add_elem() or insert_elem(), so you'll have to set_extra_integer() after making one of those calls.

tests/mesh/extra_integers.C
```diff
@@ -0,0 +1,100 @@
...
+    MeshRefinement mr(mesh);
+    mr.uniformly_refine(1);
+
+    // Make sure the old elements have the same integers and the new
+    // elements have be initialized as undefined.
+    for (const auto & elem : mesh.element_ptr_range())
+      {
+        CPPUNIT_ASSERT_EQUAL(elem->n_extra_integers(), 1);
+        if (elem->level())
+          CPPUNIT_ASSERT_EQUAL(elem->get_extra_integer(i1), DofObject::invalid_id);
```
YaqiWang: Can we change the default behavior to let the child element inherit the integers in its parent element?

tests/mesh/extra_integers.C
```diff
@@ -0,0 +1,100 @@
...
+  CPPUNIT_TEST_SUITE_END();
+
+protected:
+  // Helper function called by the test implementations, saves a few lines of code.
+  void test_helper_1D(ElemType elem_type)
+  {
+    Mesh mesh(*TestCommWorld, /*dim=*/2);
+
+    // Request some extra integers before building
+    unsigned int i1 = mesh.add_extra_integer("i1");
```
YaqiWang: Should this be `add_elem_integer`?

YaqiWang: Not sure if a function like `get_extra_integer_index(const std::string &)` will be useful. I guess it does not hurt to add it.

roystgnr: Yeah, this is one of the completely broken things I'm up late fixing right now.  :-)

tests/mesh/extra_integers.C
roystgnr: Hm... yeah, that won't take much more work.

tests/mesh/extra_integers.C
roystgnr: get_elem_integer_index would be helpful for debugging?  add_extra_integer *already* gives that functionality, since you can call it multiple times with the same name and it'll return the same index value, but doing so is dangerous, since following ```add_extra_integer("correct_name")``` with ```add_extra_integer("corect_name")``` will just silently add a new integer with the wrong name and return the index for that.  With a "get_elem_integer_index" we could actually assert that the requested name exists.

YaqiWang: `get_elem_integer_index` can be const. It was in my mind that adding the integer is in the mesh generation, while the integer can be used in a material. The material does not needs to know the integer, but hard-coding a name for the integer is better. Actually this may not be important because I can find a place to store the index and have a meaningful function such as `getMaterialIDIndex` to retrieve it.

roystgnr:
Ok, this should be ready to merge today unless there are any objections or any more good suggestions.

PR #2103, opened 2019-04-17
Mesh stitching fixes for NodeElems

jwpeterson:
The code would previously compute an `h_min` value of 0 in cases where the first elem in the mesh was a `NodeElem`, thereby causing all subsequent relative distance checks to fail.


jwpeterson:
This can also be backported to 1.4.1 (#2029).

PR #2105, opened 2019-04-18
Add FEMContext::interior_rate_gradient

pbauman:
This is needed for computing things like the time derivative
of the deformation gradient in solid mechanics applications.


roystgnr:
My extreme pride at the factoring I've done in those FEMContext evaluation methods is sullied only by the shame I feel for never having done a similar job for ```System::point_*```

PR #2108, opened 2019-04-23
CheckpointIO support for extra DofObject integers

### LINKED ISSUE #2095: Add two integers to Elem

roystgnr:
If this works, it should *actually* close #2095 

My brain is pretty frazzled right now, so it probably doesn't work yet and shouldn't be merged until it gets some test coverage; I just want to let CI and @YaqiWang get an early look before I get started on unit tests.

### DIFF
--- a/include/base/dof_object.h
+++ b/include/base/dof_object.h
@@ -227,7 +227,7 @@ public:
    * with \p index, which should have been obtained via a call to \p
    * MeshBase::add_elem_integer or \p MeshBase::add_node_integer
    */
-  dof_id_type get_extra_integer (const unsigned int index);
+  dof_id_type get_extra_integer (const unsigned int index) const;
 
   /**
    * Adds an additional system to the \p DofObject
@@ -994,7 +994,7 @@ DofObject::set_extra_integer(const unsigned int index,
 
 inline
 dof_id_type
-DofObject::get_extra_integer (const unsigned int index)
+DofObject::get_extra_integer (const unsigned int index) const
 {
   libmesh_assert_less(index, this->n_extra_integers());
   libmesh_assert_less(this->n_systems(), _idx_buf.size());
--- a/include/mesh/checkpoint_io.h
+++ b/include/mesh/checkpoint_io.h
@@ -158,6 +158,15 @@ public:
   std::string &       version ()       { return _version; }
 
   /**
+   * \returns \p true if the current file has an XDR/XDA version that
+   * matches or exceeds 1.5
+   *
+   * As of this version we encode any extra integer data that has been
+   * attached to a mesh.
+   */
+  bool version_at_least_1_5() const;
+
+  /**
    * Get/Set the processor id or processor ids to use.
    *
    * The default processor_id to use is the processor_id() of the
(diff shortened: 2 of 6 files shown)


YaqiWang:
Extra integers are supposed to be set during mesh generation, which should be skipped during recovering. So extra integers need to be written to and load from the checkpoint file. I do not care much how it is done just needs to be transparent to users as the other parts of the mesh. The interface changes in getting the index, writing the integers, etc. are good. Thanks Roy for putting all these in.

roystgnr:
> Extra integers are supposed to be set during mesh generation, which should be skipped during recovering. So extra integers need to be written to and load from the checkpoint file.

That's what we're doing here.  The only limitation is that the recovery should be done into either an empty Mesh or into a mesh that has had the exact same extra integers added in the same order, otherwise we scream and die at runtime.

Which I'm not proud of but honestly reading into a non-empty Mesh probably breaks in all kinds of ways with various MeshInput subclasses, since in user code we generally create a new Mesh object whenever we want to read in a new file.
...
Or we clear an old Mesh.  Damn.  I just realized that MeshBase::clear() isn't clearing the extra_integers metadata.  I'll fix that.

roystgnr:
> Damn. I just realized that MeshBase::clear() isn't clearing the extra_integers metadata. I'll fix that.

Or... maybe I won't.  Our MeshGeneration methods call clear() before creating new elements, so we certainly can't remove add_elem_integer results there!

I don't like the idea of clear() not meaning "restore to a pristine state" but I don't see a way around this other than to document the unintuitive behavior.

YaqiWang:
@roystgnr since you are working on recover here. I have a quick question: can I just read one entry in a recover file, for example, read one variable, assuming mesh, target variable space have been setup properly?

roystgnr:
Hmm... in 2011 I added support for "upconverting" EquationSystems xdr/xda saves: e.g. you could save a reaction problem based on a model with a few chemical species, then read that file into a more complicated chemistry system that had additional species too, and the matching variables would be read from file while the non-matching variables would remain at zero (or IIRC remain at their old values if they'd already been set in some other fashion).

But that only works if the variable(s) you want to read are the only ones in the save file.  I don't think we have any means to extract only some variables' data from a save file that has more, or to save data from only a subset of variables to a file.

pbauman:
> I don't think we have any means to extract only some variables' data from a save file that has more, or to save data from only a subset of variables to a file.

Don't we have a libMesh app that does something along these lines though?

roystgnr:
> Don't we have a libMesh app that does something along these lines though?

Ha!  solution_components, yeah.

My brain's stopped working.  Today in my dad's neurology room I was asking the staff if they had any two-for-the-price-of-one deals...

PR #2113, opened 2019-04-26
shared_ptr<GhostingFunctor> support

### LINKED ISSUE #1876: GhostingFunctor API
Simple: `add_ghosting_functor()` should take a `shared_ptr`... right now it's a weird thing where you have to find somewhere to hold onto them for as long as the Mesh exists...

roystgnr:
This is the API I'm proposing to resolve #1876; if everyone likes it we can merge after I add some unit testing.

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -306,6 +306,18 @@ public:
                             bool to_mesh = true);
 
   /**
+   * Adds a functor which can specify coupling requirements for
+   * creation of sparse matrices.
+   *
+   * GhostingFunctor memory when using this method is managed by the
+   * shared_ptr mechanism.
+   */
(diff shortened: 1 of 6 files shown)


pbauman:
Cool. Do we want to go ahead and deprecate the old interface?

roystgnr:
I don't think so, actually.  Once I hit upon the idea of keeping shared_ptr storage separate, it turned out that having twice the interface options doesn't force us to have twice the code.  And I like still being able to write simple code where we just keep stuff on the stack, even if serious middleware is all going to want to switch to shared_ptr.

Plus I'm not completely certain this should be the final API.  We may want to add a clone() method to help with threading, for instance.

Issue #1312, opened 2017-04-12
Slight issue with adaptivity_ex3 and PETSc 3.7.x


jwpeterson:
If you run adaptivity_ex3 with PETSc 3.7.x (and possibly older versions, have not tested yet) under the following configurations:
```
./example-opt refinement_type=p
./example-opt refinement_type=hp max_r_steps=8
./example-opt dimension=1 refinement_type=p
./example-opt dimension=1 approx_type=HERMITE approx_order=3 refinement_type=p
```
the last couple of refinement steps converge in 0 iterations, i.e. PETSc prints:
```
Linear solver converged at step: 0, final residual: 0
```
and then something else divides by zero in the L<sup>2</sup> and H<sup>1</sup> norm calculations, producing NaNs.

The example itself seems to still run fine, there probably just needs to be a check for these degenerate solve cases to avoid the division by zero.

roystgnr:
Hmmm... I only had to reduce max_r_steps on 3 of these cases when #2117 exposed the problem.  The 1d p-refinement case still worked fine for me.

jwpeterson:
This issue can probably be closed then, I doubt it's relevant any more.

roystgnr:
Yeah, I think we're good.  I just did a 1 through 24 proc sweep to check, and everything was happy.  With DistributedMesh, so it's still theoretically possible a ReplicatedMesh partitioning makes a difference on some case, but I can't imagine even METIS/ParMETIS screwing up a 1-d partitioning that badly.

PR #2117, opened 2019-05-01
Assert consistent ErrorVector values for flagging

roystgnr:
Otherwise we'll happily set inconsistent flags and not discover the
error until we actually start trying mesh refinement.


roystgnr:
Ha!  We have an example there with a failing linear solver; I've just never noticed because I always use some stronger non-default PETSc solver options.

jwpeterson:
#1312 might be related.

PR #2119, opened 2019-05-02
Add member function sub_matrix in DenseMatrix

### LINKED ISSUE #2118: Add a member function sub_matrix in DenseMatrix
I am motivated to do it because I will probably need it for the array kernel stuff that I am working on. I plan to have an interface like
```
DenseMatrix<T> DenseMatrix<T>::sub_matrix(
                                       unsigned int row_id, unsigned int row_size,
                                       unsigned int col_id, unsigned int col_size) const;
```
Even I might not be using it (highly unlikely), it does not hurt to have such a function.

YaqiWang:
Close #2118.

### DIFF
--- a/include/numerics/dense_matrix.h
+++ b/include/numerics/dense_matrix.h
@@ -91,6 +91,12 @@ public:
   virtual void zero() override;
 
   /**
+   * Get submatrix with the smallest row and column indices and the submatrix size.
+   */
+  DenseMatrix sub_matrix(unsigned int row_id, unsigned int row_size,
+                         unsigned int col_id, unsigned int col_size) const;
+
+  /**
    * \returns The \p (i,j) element of the matrix.
    */
   T operator() (const unsigned int i,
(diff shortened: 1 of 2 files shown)


jwpeterson:
Seems like an OK idea to me, although you should consider what @pbauman mentioned and see if the existing `DenseSubMatrix` would potentially work for your needs. Note that you can ask for the `parent ()` `DenseMatrix` and pass that to `add_matrix()` if that makes sense...

The build errors look like simple `-Wshadow` warnings treated as errors that should be easy to fix.

YaqiWang:
Thanks @jwpeterson. I answered @pbauman in the issue. I had a narrow path to make the array variable (variable group) work with MOOSE to create something named as `ArrayMooseVariable`, `ArrayKernel` so that one kernel can operate on bunch (could be hundreds) of related variables simultaneously. Assembling the Jacobian is the most tricky part so far.

roystgnr:
As I said in the issue I don't really like this solution to your use case, but I con't see any better solution, and this might be useful for others in similar problems.  Thanks!

PR #2123, opened 2019-05-03
Exodus improvements

andrsd:
Improving support for scalar variables in ExodusII_IO
   - can get their names
   - can copy them into a System


andrsd:
Looks like the bug I found is already fixed. My bad for being behind...

PR #2124, opened 2019-05-06
Add public setter for _elem_dims.

jwpeterson:
In debug mode, assert that user's elem dimensions agree with true elem
dimensions.


roystgnr:
Looks good; thanks!

PR #2129, opened 2019-05-10
Add Communicator::split_by_type()

roystgnr:
I don't have a proper test case for this; hopefully @permcody can catch me if I've missed anything.


permcody:
Looks ok to me. Our testing systems are down for scheduled maintenance for several days this week. I'll get around to testing this later this week or next. You can merge it if you'd like in the meantime.

PR #2132, opened 2019-05-14
ExodusII: write all nodesets simultaneously.

jwpeterson:
This is similar to what we did in b4ac1df1 (#968) to speed up the
writing of multiple subdomains.

For a mesh with about 26k nodesets, this speeds up a test code that
reads in and writes out the mesh from about 700 seconds to about 30
seconds. I think a similar optimization on the reading side would be
able to get that number down even lower.


roystgnr:
:+1:

> For a mesh with about 26k nodesets

I need to start making a list of "things where we didn't originally care about O(f(N)) scaling because I thought N was going to be 20 or maybe 200, not 20K".  Variables, processors, subdomains, nodesets...

What's the use case?

jwpeterson:
> What's the use case?

We primarily use them when connecting different components together. It might be possible to set up the model differently in this case so that there aren't so many nodesets, but since the fix is (relatively) simple and arguably addresses a shortcoming in our use of the API, there's no reason not to just add it anyway.

PR #2008, opened 2019-01-22
 Enable base-side in InfFE::reinit(elem, side) 

BalticPinguin:
The main issue I try to address is that I want to compute properties of the infinite elements at the boundary to the FEM-region. Thus, I need `side==base_side` in `InfFE::reinit(elem, side)`.

Along the lines, I'd like to fix some errors I introduced in #1981 (sorry for that).
and to make the above-mentioned function working, the `InfFE::inverse_map()` must give points outside the element when I feed it with points not being in the element.

The changes are actually more since I ran the `reindent.sh`-script on the code...

I hope not to introduce some strange or wrong behaviour; my tests were all successfull.

PR #2137, opened 2019-05-18
Fix for memory-leak in inf.elements.

BalticPinguin:
I found that I introduced a memory-leak in #2008.
This should fix it.


jwpeterson:
In fact I think you can simply delete the line in question, as the assignment which occurs on the next line [handles this automatically](https://en.cppreference.com/w/cpp/memory/unique_ptr/operator%3D). I do agree that there was a leak previously, so thanks for fixing this!

Issue #2133, opened 2019-05-14
Feature Request: 3D mesh with 1D elements in GMSH mesh files (.msh)


SpaceSapien:
I wanted to ask if a 3D mesh with 1D elements would be able to be added for GMSH .msh files.

See
https://groups.google.com/forum/#!topic/moose-users/zRud9vjgvzM

Thanks

jwpeterson:
If I recall correctly, there was no gmsh-official way to determine whether lower-dimensional elements are present in the mesh file because a) they represent lower-dimensional elements which should be present in the mesh, or b) they are only being used to represent boundary conditions. 

The approach which Martin Lesueur came up with to handle this was to add the "lower_dimensional_block" string to the PhysicalNames section to indicate that a given element set should be maintained in the final mesh as lower-dimensional elements. So, for example, the PhysicalNames section might look something like the following:
```
2 1 "frac" lower_dimensional_block
2 3 "top"
2 4 "bottom"
3 2 "volume"
```
And then we would know to keep the elements of the "frac" block around for use later. I don't see any reason why this approach wouldn't work for meshes with 1, 2, and 3 dimensional elements present. In fact, it may just "work" if you take out this check that's currently in the code: 
```c++
// Have not yet tested a case where 1, 2, and 3D elements are all in the same Mesh,
// though it should theoretically be possible to handle.
if (n_dims_seen == 3)
  libmesh_error_msg("Reading meshes with 1, 2, and 3D elements not currently supported.")
```
and of course mark all the lower_dimensional_blocks appropriately. I don't have time to try this out myself, but it should be pretty easy to at least get to the next relevant error message and then maybe somebody can pick it up from there.

Note that the test.gmsh file that you linked on the MOOSE mailing list is not a mesh file but a gmsh input file. I have not used this lower_dimensional_block feature before myself so I think it may have to be appended manually once the actual mesh file is generated, as it is a libmesh-specific feature.

lindsayad:
@SpaceSapien I know that this issue came out of the moose users list. For what it's worth, if you have created physical groups in your gmsh input (whether using the old scripting `.geo` files or using gmsh's new cool python/C/C++ APIs) corresponding to boundaries, you can use MOOSE mesh modifiers to generate lower dimensional element blocks from those "boundaries". See [here](https://github.com/idaholab/moose/blob/next/framework/src/meshmodifiers/LowerDBlockFromSideset.C). However,

[...]

PR #2139, opened 2019-05-18
Deleted two checks that disabled blocks that where more than 1 dimens…

SpaceSapien:
I had a 3D mesh with 1D blocks for a MOOSE application and the current gmsh reader wouldn't accept dimensional changes larger than 1 (ex a 2D mesh with 1D blocks or a 3D mesh with 2D blocks). 

Based on the advice of @lindsayad  and @permcody I removed two checks inside of gmsh_io.c. 

I can now verify that the 3D 1D hybrid mesh works with gmsh v2 type meshes. I did not check for the gmsh type 3/4 but I would assume that they would work as well.

### DIFF
--- a/src/mesh/gmsh_io.C
+++ b/src/mesh/gmsh_io.C
@@ -707,10 +707,6 @@ void GmshIO::read_mesh(std::istream & in)
             // libMesh::out << "max_elem_dimension_seen=" << max_elem_dimension_seen << std::endl;
             // libMesh::out << "min_elem_dimension_seen=" << min_elem_dimension_seen << std::endl;
 
-            // If the difference between the max and min element dimension seen is larger than
-            // 1, (e.g. the file has 1D and 3D elements only) we don't handle this case.
-            if (max_elem_dimension_seen - min_elem_dimension_seen > 1)
-              libmesh_error_msg("Cannot handle meshes with dimension mismatch greater than 1.");
 
             // How many different element dimensions did we see while reading from file?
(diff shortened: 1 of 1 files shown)


SpaceSapien:
For anyone interested this is my .msh fiile.
[3D1dmesh.zip](https://github.com/libMesh/libmesh/files/3194522/3D1dmesh.zip)

roystgnr:
That really was sufficient?  Great!

If you find time then we'd love a unit test to protect against regressions here (IIRC there's a newer gmsh format version we don't support yet, in which case I'll be afraid of breaking older features whenever we upgrade to add it), but beggars can't be choosers; thanks for the fix!

lindsayad:
@roystgnr We added support for the newest gmsh version in #2082

jwpeterson:
> Based on the advice of @lindsayad and @permcody I removed two checks inside of gmsh_io.c.

https://github.com/libMesh/libmesh/issues/2133#issuecomment-493121620

???????

SpaceSapien:
> ???????

Also @jwpeterson  :)

permcody:
> ???????

You do know the rule about four exclamation points, right? I wonder how many questions marks constitute insanity.

pbauman:
> I wonder how many questions marks constitute insanity.

N+1

jwpeterson:
> ???????

Bonus fact, this is also my reaction to almost every news story I read these days.

PR #2142, opened 2019-05-21
Fix for NaNs in block restricted vector variables using PetSc 3.8.3

### LINKED ISSUE #2141: NaN in result when vector variables are restricted to a subdomain
When creating idaholab/moose#13359, I experienced some unexpected errors using libmesh (as of c4c9fd5) with PetSc 3.8.3 for a 3D test.

In those MOOSE tests (see [`vector_3d.i`](https://github.com/cticenhour/moose/blob/4290733582c25cd2e07ca7bd9db7a471becc5aca/test/tests/interfacekernels/3d_interface/vector_3d.i)), I have two vector variables, `u` and `v`, with `u` being active in subdomain 0 and `v` being active in subdomain 1. `u` has kernels active only in subdomain 0, while `v` only has kernels in subdomain 1. I am applying a vector penalty interface condition (my MOOSE contribution in the PR is to enable VectorInterfaceKernels in general) to ensure `u = v` along the interface between the subdomains, with natural conditions along all external boundaries.

In practice, I would receive good results in 2D and 3D using libMesh and PetSc 3.9.4, but during testing with 3.8.3, NaNs would occur in the solution for `v` in subdomain 0, but **only** in the Y and Z components of the vector. Fande and I tracked this down to a divide by zero that occurs in `equation_systems.C:819`. The difference between PetSc versions is that the `/=` operator is able to catch divide-by-zero in 3.9.4, but doesn't in 3.8.3. 

I did find a fix for this. It seems as though `repeat_count` zero entries were not being set to 1 consistently for variables that do not exist on a particular subdomain. The expression used to calculate the proper vector index in `equation_systems.C:784` differed from that in other parts of the code just before. Specifically, the `var_inc+d` term was `var+d` here. `var` didn't seem to be tied to the number of components of the vectors in the problem, rather, the number of variables in the problem (via `nv_sys`). This change both improved my test solutions and removed the NaNs for those components it was affecting. So I will be submitting a PR with this fix in a few minutes. 

My question for ya'll - was there some specific intention that you might/might not remember for having `var+d` over `var_inc+d` on Line 784? Just want to make sure I am fixing up the right thing  in the right way and not just making a new problem later. :-)

cticenhour:
closes #2141

### DIFF
--- a/src/systems/equation_systems.C
+++ b/src/systems/equation_systems.C
@@ -781,7 +781,7 @@ EquationSystems::build_parallel_solution_vector(const std::set<std::string> * sy
                   // Only do this if this variable has NO DoFs at this node... it might have some from an adjoining element...
                   if (!elem->node_ptr(n)->n_dofs(sys_num, var))
                     for (unsigned int d=0; d < n_vec_dim; d++)
-                      repeat_count.add(nv*(elem->node_id(n)) + (var+d + var_num), 1);
+                      repeat_count.add(nv*(elem->node_id(n)) + (var_inc+d + var_num), 1);
 
             } // end loop over elements
           var_inc += n_vec_dim;


lindsayad:
Do we need to add @cticenhour to the libMesh organization in order to get CIVET tests to fire off automatically? I think that's what @milljm is telling me.

jwpeterson:
@lindsayad that's correct. @cticenhour: invite sent!

jwpeterson:
Not sure what happened with the failing test, but it doesn't seem related:
```
meshgenerators/image_mesh_generator.image_mesh_generator_test: Error calling 'file' command in ImageMesh::buildMesh()
```
I'll invalidate, see if that fixes it...

jwpeterson:
> I'll invalidate, see if that fixes it...

Nevermind, you beat me to it.

cticenhour:
Thanks @jwpeterson!

PR #2147, opened 2019-05-23
Fixed compiling errors against  with PETSc-3.11.x

### LINKED ISSUE #2146: Compiling errors with PETSc-3.11.x
There is an API change in PETSc-3.11.x.

`DMCreateSubDM_Section_Private` was changed to `DMCreateSectionSubDM`

It is a good change actually since the API is public now.

fdkong:
Closes #2146

### DIFF
--- a/src/solvers/petsc_dm_wrapper.C
+++ b/src/solvers/petsc_dm_wrapper.C
@@ -107,10 +107,14 @@ namespace libMesh
               ierr = DMShellSetContext(*subdm, ctx);
               LIBMESH_CHKERR(ierr);
             }
-
+#if PETSC_VERSION_LESS_THAN(3,11,0)
           // Lastly, Compute the subsection for the subDM
           ierr = DMCreateSubDM_Section_Private(dm, numFields, fields, is, subdm);
           LIBMESH_CHKERR(ierr);
+#else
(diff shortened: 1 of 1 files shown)


fdkong:
@jwpeterson  Could you take a look at this?

jwpeterson:
BTW, I attached a test with PETSc master but we don't run that often so I'm not sure whether it actually passes.

fdkong:
> BTW, I attached a test with PETSc master but we don't run that often so I'm not sure whether it actually passes.

It is good to check. If it does not work, I will try to fix it.

jwpeterson:
@fdkong, looks like the new code compiles fine but there are some failing MOOSE tests... they seem to be unrelated to your specific change, but they could be caused by a new PETSc... I'd be OK with merging this as-is, unless you want me to hold off for some reason.

fdkong:
> @fdkong, looks like the new code compiles fine but there are some failing MOOSE tests... they seem to be unrelated to your specific change, but they could be caused by a new PETSc... I'd be OK with merging this as-is, unless you want me to hold off for some reason.

It is better to merge it in as-is, and the falling is unrelated to this PR. I will try to fix on the moose side.  I need this PR in before I can do some tests in MOOSE.

fdkong:
Thanks, @jwpeterson

Issue #2148, opened 2019-05-28
Different meaning of p_level in Elem vs. QBase


jwpeterson:
Currently you get a `FOURTH`-order quadrature rule on a `(FIRST, LAGRANGE)` FE when reinit'd on an Elem with a boosted `p_level==1`.  This is a bit counter-intuitive, since for a normal `(SECOND, LAGRANGE)` FE you would get a `FIFTH`-order rule based on the formula in `FEType::default_quadrature_order()`. The sequence of events which currently happens is:

1. `Elem::reinit()` is called on a p_level boosted Elem.
1. Internally this calls `qrule->init(elem->type(), elem->p_level())` and that sets a separate `_p_level` flag (equal to 1) on the QBase object itself.
1. The `QBase::get_order()` function then returns `static_cast<Order>(_order + _p_level)` where `_order==THIRD` for a standard (FIRST, LAGRANGE) `FEType`, resulting in an overall order of `FOURTH`. 

So, `p_level==1` has a different meaning for Elems ("increase polynomial order used on this Elem by 1") and Quadrature rules ("increase the order of this rule by 1"). I think it would be good if this was more consistent with `FEType::default_quadrature_order()`, where the default quadrature order is based on the quadrature required to compute the mass matrix exactly, and which we have chosen to be: 2 * polynomial_order + 1.

Note that for tensor product Gauss rules this doesn't matter, because `FOURTH`/`FIFTH`-order rules are the same for those elements, but for other geometric Elem types and quadrature rules it could produce a slightly unexpected result.

I think we should change `QBase::get_order()` to return `_order + 2 * _p_level` based on the following logic:

For consistency with `FEType::default_quadrature_order()`, the total quadrature order returned by `QBase::get_order()` for a given Elem should be:
```
total_order = 2 * (elem->p_level() + "FE_order") + 1       (*)
```
where "FE_order" is the order of the underlying FE object. Now, at the QBase level, we don't know what FE_order is, but we do have an `_order` that we were created with, so for the moment assume that it is consistent with an FE_order obtained using the default formula, i.e.
```
_order = 2 * FE_order + 1
```
and therefore, `FE_order = (_order - 1) / 2`. Plugging this back into (*), we get:
```
total_order = 2 * elem->p_level() + _order
```
as desired. Note that it doesn't matter if the `_order` which the rule was created with isn't actually consistent with `FEType::default_quadrature_order()`, we will never integrate with a rule less than `_order`.

[...]

PR #2149, opened 2019-05-28
Fix formula for computing total quadrature order.

jwpeterson:
Also update the documentation for this function to explain its purpose
a bit better.

It also occurred to me that we should probably be calling get_order()
in all the various rules' Order switch statements instead of repeating
the same formula over and over, so that will be a topic for another
PR.

Refs #2148.

### DIFF
--- a/include/quadrature/quadrature.h
+++ b/include/quadrature/quadrature.h
@@ -198,9 +198,18 @@ public:
                      unsigned int p_level=0);
 
   /**
-   * \returns The order of the quadrature rule.
+   * \returns The current "total" order of the quadrature rule which
+   * can vary element by element, depending on the Elem::p_level(),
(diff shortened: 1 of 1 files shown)


roystgnr:
@pbauman, any chance we can get the GRINS tests running on this?  I can't imagine this change revealing any actual bugs, but I could easily imagine it triggering some false positive failures on just the right combination of tricky physics plus quirky solver options.

jwpeterson:
> I can't imagine this change revealing any actual bugs,

`git grep get_order` comes up empty for GRINS, and I don't think we even use it much internally.

pbauman:
Yup, rebooted the server. Need to start coming up with a longer term solution, though, since `femputer` will go away this week.

PR #2150, opened 2019-05-28
Replace hard-coded formula with get_order()

jwpeterson:
This also revealed that some of our protected and private APIs in the QBase hierarchy were a bit misleading... they made it seem like it was possible for the quadrature rule to be initialized with *any* ElemType and *any* p_level, when in reality only `_type` and `_p_level` were being used... so I cleaned that up as well.

Refs #2149.

### DIFF
--- a/include/quadrature/quadrature.h
+++ b/include/quadrature/quadrature.h
@@ -271,8 +271,7 @@ protected:
    * It is assumed that derived quadrature rules will at least
    * define the init_1D function, therefore it is pure virtual.
    */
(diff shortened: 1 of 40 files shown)


jwpeterson:
Hmm... I didn't think about this, but obviously if someone derived from `QBase` they will have to override these protected members, so even though it's not a public API change, it's still an API change. Let me see if I can think of a workaround.

roystgnr:
:+1:

PR #2151, opened 2019-05-29
PETScDMWrapper Fixups

bboutkov:
Hi all. Here's a PR with various fixup's to things I had discovered in the recent past. While some things are minor and cosmetic, there are a few changes worth outlining.

- Previously fieldsplit without GMG had a bug in it, this is now fixed and tested against with a couple more fem_system_ex1 runs. I also added a tiny transient case as we had no earlier coverage of the teardown/resetting up process.
- Some API changes from PETSc's side were already merged (as @fdkong noticed and fixed in #2147 ; thanks!) while others are still pending an official PETSc release. I checked with PETSc developers and was notified that we can get rid of the remaining PETSc private dependency in 3.12.0, so I've preemptively included these updates.
- Fieldsplit with GMG didnt work earlier in cases where we wanted to use GMG on more than one  subblock. This is fixed now, and I will be creating coverage for this shortly in GRINS with a thermally coupled flow example.

As always, please let me know if theres something I can improve or needs fixing, and thanks in advance for the review!

### DIFF
--- a/src/solvers/petsc_dm_wrapper.C
+++ b/src/solvers/petsc_dm_wrapper.C
@@ -240,7 +240,7 @@ namespace libMesh
       // the subDMs. We do this by checking the number of fields. When
       // less than all the fields are used, we need to create the
       // proper subDMs. We get the number of fields and their names
-      // from the incomming fine DM and the global reference DM
+      // from the incoming fine DM and the global reference DM
       PetscInt nfieldsf, nfieldsg;
       char ** fieldnamesf;
       char ** fieldnamesg;
@@ -401,7 +401,7 @@ namespace libMesh
 
       PetscErrorCode ierr;
 
-      // get a communicator from incomming DM
+      // get a communicator from incoming DM
       MPI_Comm comm;
       PetscObjectGetComm((PetscObject)dmc, &comm);
 
--- a/src/solvers/petsc_dm_wrapper.C
+++ b/src/solvers/petsc_dm_wrapper.C
@@ -100,6 +100,11 @@ namespace libMesh
               ierr = DMShellSetCreateRestriction(*subdm, dm->ops->createrestriction);
               LIBMESH_CHKERR(ierr);
             }
+          if (dm->ops->createsubdm)
+            {
+              ierr = DMShellSetCreateSubDM(*subdm, dm->ops->createsubdm);
+              LIBMESH_CHKERR(ierr);
+            }
           ierr = DMShellGetContext(dm, &ctx);
           LIBMESH_CHKERR(ierr);
           if (ctx)
--- a/src/solvers/petsc_dm_wrapper.C
+++ b/src/solvers/petsc_dm_wrapper.C
@@ -490,11 +490,18 @@ void PetscDMWrapper::init_and_attach_petscdm(System & system, SNES & snes)
   // Now fill the corresponding internal PetscDMContext for each created DM
   for( unsigned int i=1; i <= n_levels; i++ )
     {
+      // Set context dimension
+      (*_ctx_vec[i-1]).mesh_dim = mesh.mesh_dimension();
+
+      // Create and attach a sized vector to the current ctx
+      _vec_vec[i-1]->init( _mesh_dof_sizes[i-1] );
+      _ctx_vec[i-1]->current_vec = _vec_vec[i-1].get();
+
+      // Set a global DM to be used as reference when using fieldsplit
+      _ctx_vec[i-1]->global_dm = &(this->get_dm(n_levels-1));
+
       if (n_levels > 1 )
         {
-          // Set context dimension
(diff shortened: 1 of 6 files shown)


roystgnr:
Everything looks good to me.  I'd feel happier if @pbauman has time to look it over too but if not we can merge anyway.

src/solvers/petsc_dm_wrapper.C
```diff
@@ -80,28 +83,97 @@ namespace libMesh
...
+
+          // Set Refine function pointer
+#if PETSC_VERSION_LESS_THAN(3,12,0)
           if (dm->ops->refine)
             {
               ierr = DMShellSetRefine(*subdm, dm->ops->refine);
               LIBMESH_CHKERR(ierr);
             }
+#else
+          PetscErrorCode (*refine)(DM,MPI_Comm,DM*) = nullptr;
```
jwpeterson: Since it appears that you use this construction in several places, it may be a bit cleaner/easier to read if you use a typedef, e.g.
 ```c++
typedef PetscErrorCode (*FunctionPointer)(DM,MPI_Comm,DM*);
FunctionPointer refine = nullptr;
ierr = DMShellGetRefine(dm, &refine);
```
Of course, the typedef can be placed anywhere in the class and used as many times as needed...

**Comment by pbauman** [comment]
I just looked at this from the point of view of documentation and it looks OK to me in that I could follow it.

**Comment by pbauman** [comment]
> Do you expect that your code will compile with current PETSc master? We can attach an additional test of that if you are interested...

Yes, this should work with master, that would be a good check, thanks!

File: src/solvers/petsc_dm_wrapper.C (hunk 1 above)
Comment (bboutkov) [reply src/solvers/petsc_dm_wrapper.C]: Thanks for the suggestion! While I do really like readability improvements, I'm not so sure this one would qualify. The parameters of these calls vary depending on the use; only really refine and coarsen have similar prototypes while the other three are all differing so I don't think there's enough to be gained here. Unless there's a manner in which I can make a parameter agnostic typedef, in which case I'd love to know!

File: src/solvers/petsc_dm_wrapper.C [hunk 2]
```diff
@@ -80,28 +83,97 @@ namespace libMesh
...
+
+          // Set Coarsen function pointer
+#if PETSC_VERSION_LESS_THAN(3,12,0)
           if (dm->ops->coarsen)
             {
               ierr = DMShellSetCoarsen(*subdm, dm->ops->coarsen);
               LIBMESH_CHKERR(ierr);
             }
+#else
+          PetscErrorCode (*coarsen)(DM,MPI_Comm,DM*) = nullptr;
```
fdkong: I am actually surprised subDM does not inherit the necessary functions (such as coarsen, refine) from DM.

src/solvers/petsc_dm_wrapper.C
```diff
@@ -53,7 +56,7 @@ namespace libMesh
 #if PETSC_VERSION_LESS_THAN(3,9,0)
     PetscErrorCode libmesh_petsc_DMCreateSubDM(DM dm, PetscInt numFields, PetscInt fields[], IS *is, DM *subdm)
 #else
-    PetscErrorCode libmesh_petsc_DMCreateSubDM(DM dm, PetscInt numFields, const PetscInt fields[], IS *is, DM *subdm)
+      PetscErrorCode libmesh_petsc_DMCreateSubDM(DM dm, PetscInt numFields, const PetscInt fields[], IS *is, DM *subdm)
```
fdkong: Remove extra space?

src/solvers/petsc_dm_wrapper.C
fdkong: Never mind. I see

bboutkov:
TIMEOUT dampers/min_damping.min_nodal_damping FAILED (TIMEOUT) [min_threads=2,min_cpus=2,FINISHED]

Is the MOOSE failure a fluke?  Not sure how this code is related to the failure.

fdkong:
> TIMEOUT dampers/min_damping.min_nodal_damping FAILED (TIMEOUT) [min_threads=2,min_cpus=2,FINISHED]
> 
> Is the MOOSE failure a fluke? Not sure how this code is related to the failure.

Does not worry. We have not been testing master much since we have a regular PETSc submodule in moose. We are checking PETSc submodule (that points to the latest release of PETSc).

src/solvers/petsc_dm_wrapper.C
bboutkov: Good catch! Looks like I need to check my emacs config.

jwpeterson: This is just emacs being confused about the correct indentation level because of surrounding ifdefs... I have seen it before.

jwpeterson:
I think this can be merged, unless you are planning to make any more changes @bboutkov?

pbauman:
He's not. I'll go ahead and merge.

PR #2152, opened 2019-05-30
Always add block size for petsc matrix

fdkong:
This information is useful for some PCs such fieldsplit and HMG

### DIFF
--- a/src/numerics/petsc_matrix.C
+++ b/src/numerics/petsc_matrix.C
@@ -154,9 +154,11 @@ void PetscMatrix<T>::init (const numeric_index_type m_in,
   LIBMESH_CHKERR(ierr);
   ierr = MatSetSizes(_mat, m_local, n_local, m_global, n_global);
   LIBMESH_CHKERR(ierr);
+  PetscInt blocksize  = static_cast<PetscInt>(blocksize_in);
+  ierr = MatSetBlockSize(_mat,blocksize);
+  LIBMESH_CHKERR(ierr);
 
 #ifdef LIBMESH_ENABLE_BLOCKED_STORAGE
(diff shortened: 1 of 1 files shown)


jwpeterson:
I haven't looked at this in detail, but are you aware we have the `--enable-blocked-storage` configure option we have? It seems like you are making it non-optional, if so we probably want to change that configure option. 

Also, we have some examples where we have had to explicitly disable this in the past (see, e.g. d5c8b076ff88c3eacbaf219b02bf9c6f2f055abd) so I'd like to be sure those are working before we turn it on for everyone...

fdkong:
> I haven't looked at this in detail, but are you aware of the `--enable-blocked-storage` configure option we have? It seems like you are making it non-optional, if so we probably want to change that configure option.
> 
> Also, we have some examples where we have had to explicitly disable blocked storage in the past (see, e.g. [d5c8b07](https://github.com/libMesh/libmesh/commit/d5c8b076ff88c3eacbaf219b02bf9c6f2f055abd)) so I'd like to be sure those are working before we turn it on for everyone...

This is actually kind of irrelevant to the block storage.  It is a regular parameter for PETSc Mat (MATAIJ, MATBAIJ, MATWhatever) that indicates for example how many unknowns are defined on each mesh vertex. Then this info will be super useful  in a preconditioner, e.g., in field split, we could naturally split  matrix based this info (by default we assume the matrix is ordered in mesh vertex major, that is, all variables for mesh vertex 1, all variables for mesh vertex2 and so on).

Yes for MATBAIJ, the block size is related to the storage.

fdkong:
We do not need to turn on the block storage in libmesh.

PR #2153, opened 2019-05-31
Explicit SubFunctor::join in SortAndCopy::join()

roystgnr:
Otherwise we end up *overriding* the subfunctor join, and then we lose
other threads' list of saved ids, and then we can't figure out how to
do subsequent projection stages properly.

Thanks to @lindsayad for helping catch this bug, but wow it should never have made it into master in the first place.  I'll start using --n_threads habitually myself, but perhaps we should make hybrid-parallel runs standard when running the libMesh "make check" through CI?

### DIFF
--- a/include/systems/generic_projector.h
+++ b/include/systems/generic_projector.h
@@ -1628,6 +1628,8 @@ void GenericProjector<FFunctor, GFunctor, FValue, ProjectionAction>::SortAndCopy
   merge_multimaps(vertices, other.vertices);
   merge_multimaps(edges, other.edges);
   merge_multimaps(sides, other.sides);
+
+  SubFunctor::join(other);
 }
 
 


lindsayad:
Unfortunately, this doesn't appear to have fixed the helgrind issues I've observed. I'm attaching helgrind output for our `moose/test/tests/geomsearch/nearest_node_locator/nearest_node_locator.i` input file. Note that this is ***not*** an urgent issue for us!

[helgrind-nearest-node-locator.log](https://github.com/libMesh/libmesh/files/3254229/helgrind-nearest-node-locator.log)

roystgnr:
Good, since if it had been urgent, I'd have screwed you over pretty badly by this point, no?  :-P

That log seems to suggest that it's actually the values of _array_is_present, _values_read_only, _first, _last, and the _read_only_values pointer that are under contention, not the value of actual array entries!  Which is good, since that means my careful work to keep reads *after* the corresponding writes might actually be working, but also bad, since this pushes the credence of "Roy understands the cool tricks @friedmud used in #856 and #1056" from "probably" down to "probably not".

Just looking at that first "Possible data race" entry, it seems like both threads 26 and 27 are in the same ```PetscVector::get -> _get_array``` stack, but the whole point of the _array_is_present tricks there is to make sure that we can call ```get()``` simultaneously from multiple threads without a data race.

Is it possible that helgrind produces a false positive when encountering a properly-fenced double-checked lock idiom?  That idiom actually *was* supposed to be buggy before C++11 memory fences, and text search through [the Helgrind manual](http://valgrind.org/docs/manual/hg-manual.html) doesn't find the word "fence", and I don't see anything obviously relevant in the "happens-before" discussion there.  What *does* look possibly relevant is "Do not roll your own threading primitives (mutexes, etc) from ... atomic counters": if our ```std::atomic<bool>``` idiom is just too awesome for Helgrind to understand, then our only option for clean output from it may be to add helgrind.h support and use ```ANNOTATE_HAPPENS_BEFORE``` etc. to hold it's hand.

roystgnr:
I'm going to be darkly amused if it actually turns out that we caught a major bug by tracking down a false positive bug report.

lindsayad:
Interesting stuff!

lindsayad:
[This failure](https://civet.inl.gov/job/540396/) still makes me think that we have an issue. When this simulation runs to completion, there are four time steps, but in that CIVET target we crash between time steps 3 and 4.

lindsayad:
I've verified that there was a true race condition to MOOSE console output during adaptivity. I should know the answer to this, but could a console output race condition cause a crash?

PR #2158, opened 2019-06-09
Add QNodal quadrature class

jwpeterson:
After a [recent discussion on the libmesh-users mailing list](https://sourceforge.net/p/libmesh/mailman/message/36685739/), I realized that we did not have nodal quadrature rules for the various "serendipity" elements (QUAD8, HEX20, PRISM15) and therefore we were not getting e.g. diagonal approximations to the mass matrix while using `QSimpson` quadrature on such elements. The new `QNodal` class is a hybrid of `QSimpson`, `QTrap`, and custom quadrature rules, depending on the element type for which it is initialized. 

It's not possible to derive a SECOND-order accurate nodal quadrature rule on serendipity elements without using negative weights. Using negative weights in turn would produce an indefinite approximation to the mass matrix which I felt was bad, so instead I went with lower-order accurate rules which also guarantee positivity.

### DIFF
--- a/include/quadrature/quadrature_nodal.h
+++ b/include/quadrature/quadrature_nodal.h
@@ -1,0 +1,90 @@
+// The libMesh Finite Element Library.
+// Copyright (C) 2002-2019 Benjamin S. Kirk, John W. Peterson, Roy H. Stogner
+
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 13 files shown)


include/quadrature/quadrature_nodal.h
```diff
@@ -0,0 +1,90 @@
...
+
+namespace libMesh
+{
+
+/**
+ * This class implements nodal quadrature rules for various element
+ * types. For some element types, this corresponds to well-known
+ * quadratures such as the trapezoidal rule and Simpson's rule. For
+ * other element types (e.g. the serendipity QUAD8/HEX20/PRISM15) it
+ * implememens a lower-order-accurate nodal quarature which still
```
bboutkov: implements*, quadrature*

jwpeterson: Wow, I must have had a mini stroke during that sentence.

PR #2163, opened 2019-06-20
Dof_map returns the right block size

fdkong:
Dof_map should return the right block size regardless if blocked storage is used or not.

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -606,7 +606,7 @@ public:
   /**
    * \returns The block size, if the variables are amenable to block storage.
    * Otherwise 1.
-   * This rontinue was originally designed to enable a blocked storage, but
+   * This routine was originally designed to enable a blocked storage, but
    * it turns out this information is still super useful for solvers even when
    * we do not use the blocked storage (e.g., MATMPIBAIJ in PETSc). For example (in PCHMG),
    * for a system of PDEs, to construct an efficient multilevel preconditioner, we coarsen
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -600,24 +600,24 @@ public:
    */
   bool has_blocked_representation() const
   {
-#ifdef LIBMESH_ENABLE_BLOCKED_STORAGE
     return ((this->n_variable_groups() == 1) && (this->n_variables() > 1));
-#else
-    return false;
-#endif
   }
 
   /**
    * \returns The block size, if the variables are amenable to block storage.
    * Otherwise 1.
+   * This rontinue was originally designed to enable a blocked storage, but
+   * it turns out this information is still super useful for solvers even when
+   * we do not use the blocked storage (e.g., MATMPIBAIJ in PETSc). For example (in PCHMG),
+   * for a system of PDEs, to construct an efficient multilevel preconditioner, we coarsen
+   * the matrix of one single PDE instead of the entire huge matrix. In order to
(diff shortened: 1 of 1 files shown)


include/base/dof_map.h
```diff
@@ -600,11 +600,7 @@ class DofMap : public ReferenceCountedObject<DofMap>,
    */
   bool has_blocked_representation() const
   {
-#ifdef LIBMESH_ENABLE_BLOCKED_STORAGE
     return ((this->n_variable_groups() == 1) && (this->n_variables() > 1));
```
roystgnr: Is this correct?  If we didn't enable blocked storage then we're not using a blocked representation even if we *could*, right?

include/base/dof_map.h
```diff
@@ -613,11 +609,7 @@ class DofMap : public ReferenceCountedObject<DofMap>,
    */
   unsigned int block_size() const
   {
-#ifdef LIBMESH_ENABLE_BLOCKED_STORAGE
     return (this->has_blocked_representation() ? this->n_variables() : 1);
```
roystgnr: This looks like it's consistent either way, but without the preceding change it becomes effectively the same either way.

include/base/dof_map.h
fdkong: Right. PETSc MAT takes the block size not necessary for storage, and the block size is useful for some preconditioners  such fieldsplit and HMG.

include/base/dof_map.h
fdkong: Without this PR, I always have a block size of 1 regardless. I would like to libMesh to setup a correct block size for me, and I can use it for the matrix structure and the preconditioner.

include/base/dof_map.h
roystgnr: I think I see.  So with this PR we end up calling ```MatSetBlockSize(_mat,something_bigger_than_1)```, and that's useful even if we don't set the type to ```MATBAIJ``` as well?

I guess that's ideal, but then we need to document the new behavior better.  It's probably too late to change ```has_blocked_representation()``` to ```could_have_blocked_representation_but_maybe_doesnt()``` but at least we should point out in the comment above that this function returns whether the matrix *could* be blocked, not whether it *is*, lest someone mistakenly assume it's a quick way to test whether we're using MATBAIJ or MATAIJ.

fdkong: >MatSetBlockSize(_mat,something_bigger_than_1), and that's useful even if we don't set the type to MATBAIJ as well?

Exactly right.

fdkong: MATBAIJ and MATAIJ take the block size. We should not have this restriction at the first place.

One idea is to leave these old interfaces alone.  I am going to change something in petsc_matrix.C

Like this:

```
PetscInt blocksize  = static_cast<PetscInt>(this->_dof_map-> n_variables());

ierr = MatSetBlockSize(_mat, blocksize);
```

roystgnr: Oh, but that would be definitely wrong.  ```n_variables()``` is only a viable block size when every variable is of the same order and same subdomain restriction and same type.

I think the changes here are the ideal way to go, I just think we ought to change the comments to correspond to the change in the behavior.

fdkong:
@roystgnr It is ready to review or merge

roystgnr:
Just tack on a typo fix commit, and we won't bother waiting for it to run through CI again before merging.

PR #2166, opened 2019-06-27
Use nodal map when reading nodal variables

### LINKED ISSUE #2165: Use the node_num_map when reading nodal variable values
This is essentially the same issue as #258, which was fixed quite a while ago, but for nodal variables.  If I have an exodus file with a non-sequential node_num_map and read results from it using `ExodusII_IO_Helper::read_nodal_var_values()`, the results are wrong. I ran into this issue using `SolutionUserObject` in MOOSE.

bwspenc:
closes #2165

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -945,13 +945,25 @@ void ExodusII_IO_Helper::read_nodal_var_values(std::string nodal_var_name, int t
   // Allocate enough space to store the nodal variable values
   nodal_var_values.resize(num_nodes);
 
+  std::vector<Real> unmapped_nodal_var_values(num_nodes);
+
   // Call the Exodus API to read the nodal variable values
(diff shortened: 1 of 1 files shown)


bwspenc:
@jwpeterson You may be interested in this as you fixed a similar issue with the element map several years ago.

jwpeterson:
Thanks. I'm kind of surprised we didn't run into any issues before now, but it could be that the files written out by MOOSE/libmesh don't use node_num_map... I guess they have to if the numbering is allowed to be non-contiguous now, though.

bwspenc:
I think we haven't hit this because we are usually reading in output files produced by MOOSE, which I believe always have the node numbering map stripped out (something which I would love to fix, by the way, because that makes it so the input and output mesh have different numbering).  I was trying to read in data from a .e file that was produced a different way.

I ran the MOOSE tests and modules tests with this, and they all pass.

PR #2167, opened 2019-06-28
Update metaphysicl

lindsayad:
Summary of changes:
- Reduce temporary creation and foster loop fusion for derivative calculations
  with static storage types
- Fix std::pow(0,b>1) derivatives

The former is really quite important for all AD calculations in MOOSE that use `*` or `/`. The latter is at least important for @dschwen


lindsayad:
@rwcarlsen

PR #2169, opened 2019-07-02
Expand PetscMatrix interface

lindsayad:
Methods added:
  - local_m
  - local_n
  - get_local_size
  - reset_preallocation

### DIFF
--- a/include/numerics/petsc_matrix.h
+++ b/include/numerics/petsc_matrix.h
@@ -151,6 +151,11 @@ public:
    */
   void update_preallocation_and_zero();
 
+  /**
+   * Reset matrix to use the original nonzero pattern provided by users
+   */
+  void reset_preallocation();
+
   virtual void clear () override;
 
   virtual void zero () override;
@@ -163,7 +168,24 @@ public:
 
   virtual numeric_index_type m () const override;
 
+  /**
+   * Get the number of rows owned by this process
+   */
+  numeric_index_type local_m () const;
+
   virtual numeric_index_type n () const override;
+
+  /**
+   * Get the number of columns owned by this process
+   */
+  numeric_index_type local_n () const;
+
+  /**
+   * Get the number of rows and columns owned by this process
+   * @param i Row size
+   * @param j Column size
+   */
+  void get_local_size (numeric_index_type & m, numeric_index_type & n) const;
 
   virtual numeric_index_type row_start () const override;
 
--- a/src/numerics/petsc_matrix.C
+++ b/src/numerics/petsc_matrix.C
@@ -470,6 +470,14 @@ void PetscMatrix<T>::update_preallocation_and_zero ()
   libmesh_not_implemented();
 }
 
+template <typename T>
+void PetscMatrix<T>::reset_preallocation()
+{
+  libmesh_assert (this->initialized());
+
+  auto ierr = MatResetPreallocation(_mat);
+  LIBMESH_CHKERR(ierr);
+}
 
 template <typename T>
 void PetscMatrix<T>::zero ()
@@ -1002,7 +1010,18 @@ numeric_index_type PetscMatrix<T>::m () const
   return static_cast<numeric_index_type>(petsc_m);
 }
 
+template <typename T>
+numeric_index_type PetscMatrix<T>::local_m () const
+{
+  libmesh_assert (this->initialized());
 
+  PetscInt m = 0;
+
+  auto ierr = MatGetLocalSize (_mat, &m, PETSC_NULL);
+  LIBMESH_CHKERR(ierr);
+
+  return static_cast<numeric_index_type>(m);
+}
 
 template <typename T>
 numeric_index_type PetscMatrix<T>::n () const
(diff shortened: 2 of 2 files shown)


lindsayad:
If someone knows the other matrix types, then at least the size methods could be moved down into the base class and made virtual

src/numerics/petsc_matrix.C
```diff
@@ -1002,7 +1010,18 @@ numeric_index_type PetscMatrix<T>::m () const
...
 }
 
+template <typename T>
+numeric_index_type PetscMatrix<T>::local_m () const
+{
+  libmesh_assert (this->initialized());
+
+  PetscInt m = 0;
 
+  auto ierr = MatGetLocalSize (_mat, &m, PETSC_NULL);
```
fdkong: Minor issue: we just simply use NULL in PETSc right now. We were using PETSC_NULL, but not any more.

src/numerics/petsc_matrix.C
```diff
@@ -470,6 +470,14 @@ void PetscMatrix<T>::update_preallocation_and_zero ()
   libmesh_not_implemented();
 }
 
+template <typename T>
+void PetscMatrix<T>::reset_preallocation()
+{
+  libmesh_assert (this->initialized());
+
+  auto ierr = MatResetPreallocation(_mat);
```
fdkong: Version control

lindsayad:
Comments hopefully addressed

src/numerics/petsc_matrix.C
jwpeterson: Haha, really? I guess we should actually be changing them all to `nullptr` now but I don't really care either way. Was `PETSC_NULL` ever not `NULL`? What issue was it used to address?

fdkong: I have no idea whey we were using PETSC_NULL. I still have a lot of PETSC_NULL in my PhD code:-)

In PETSc, we have this:

`#define PETSC_NULL NULL`

But most of PETSc code is using `NULL`

Cody forced me to use `nullptr` for PETSc code in MOOSE.

src/numerics/petsc_matrix.C
lindsayad: Ugh @fdkong you told me the wrong version number [here](https://github.com/idaholab/moose/pull/13670#discussion_r299159181). Hence the test failures on idaholab/moose#13682. Looks like it should be 3.9.0 based on `git tag --contains`. I'm guessing you just made a typo sigh.

fdkong: https://www.mcs.anl.gov/petsc/petsc-3.9/docs/manualpages/Mat/MatResetPreallocation.html

It is in 3.9. I am confused here.

jwpeterson: `!PETSC_VERSION_LESS_THAN(3,8,0)` means >= 3.8.0, i.e. including 3.8.0

src/numerics/petsc_matrix.C
```diff
@@ -470,6 +470,20 @@ void PetscMatrix<T>::update_preallocation_and_zero ()
   libmesh_not_implemented();
 }
 
+template <typename T>
+void PetscMatrix<T>::reset_preallocation()
+{
+#if !PETSC_VERSION_LESS_THAN(3,8,0)
```
fdkong: Oh I see.

 #if !PETSC_VERSION_LESS_THAN(3,9,0)

8-->9

I was thinking about 9. Why I gave you 8

src/numerics/petsc_matrix.C
fdkong: Right.

fdkong: We need >=3.9.0

include/numerics/petsc_matrix.h
```diff
@@ -163,8 +168,25 @@ class PetscMatrix final : public SparseMatrix<T>
...
+   * Get the number of rows owned by this process
+   */
+  numeric_index_type local_m () const;
+
   virtual numeric_index_type n () const override;
 
+  /**
+   * Get the number of columns owned by this process
+   */
+  numeric_index_type local_n () const;
```
roystgnr: I'm confused by this one.  My first interpretation here is "the total number of columns with non-empty entries in the sparsity pattern in rows owned by this process", but I'm not actually sure how PETSc would determine that - we give them n_nz and n_oz rather than complete sparsity info, right?  And without the latter there's no way to immediately determine which non-empty columns from which rows overlap?  So is this method only well-defined for matrices that have already been assembled??

include/numerics/petsc_matrix.h
```diff
@@ -163,8 +168,25 @@ class PetscMatrix final : public SparseMatrix<T>
 
   virtual numeric_index_type m () const override;
 
+  /**
+   * Get the number of rows owned by this process
+   */
+  numeric_index_type local_m () const;
```
roystgnr: Can we bump this up to the parent class and make the default implementation ```row_stop()-row_start()```?

include/numerics/petsc_matrix.h
fdkong: It is not related to nnz, noz or matrix assembly. It is a layout thing. It is something we tell PETSc to use in libMesh.

In `PetscMatrix::init()`, we have something like 

```
ierr = MatSetSizes(_mat, m_local, n_local, m_global, n_global);
  LIBMESH_CHKERR(ierr);
```

fdkong: `"n_local"` is `"local_n()"`

include/numerics/petsc_matrix.h
fdkong: We could, but `getting m` directly from Mat looks better for me.

include/numerics/petsc_matrix.h
lindsayad: Thanks @fdkong. This is what I had determined through experimentation this morning :smile:

roystgnr: That certainly clears it up!  Thanks!

PR #2170, opened 2019-07-02
Fix PETSc version number for MatResetPreallocation

lindsayad:
Refs #2169


jwpeterson:
Just going to merge because I don't think we have any libmesh PR tests for PETScs old enough to make a difference here.

lindsayad:
> Just going to merge because I don't think we have any libmesh PR tests for PETScs old enough to make a difference here.

Yea I guess you're right. We have one million different PETSc targets for MOOSE but apparently only one for libMesh.

jwpeterson:
> but apparently only one for libMesh.

I think we have two, "regular" and "PETSc master", unless there is a third one in the min-gcc/clang tests.

PR #2171, opened 2019-07-03
SparseMatrix base implementation of local_m

lindsayad:
Follow on to #2169 per @roystgnr request


lindsayad:
INL's network and thus CIVET is currently under crisis 😢

roystgnr:
:+1: once Civet's happy again.  As long as I'm getting presents under my pillow from the Code Fairy, though: unit test?

No, wait, never mind.  We don't even have *any* proper SparseMatrix unit testing right now, so that wouldn't just be a quick drop-in addition.

PR #2173, opened 2019-07-08
Triangle IO improvements

### LINKED ISSUE #2172: Need to have some improvements in Triangle IO
These improvements include:
1. make `TriangleInterface::Hole` support a hole with disconnected boundaries, for example a ring shape hole. This requires setting up the segments for holes slightly differently;
2. optionally create `BoundaryInfo` in `TriangleWrapper::copy_tri_to_mesh` otherwise the mesh generated from Triangle does not have boundaries by default;
3. add a function to obtain the holes created by a mesh (embedded in the geometry to be triangulated);
4. make `TriangleInterface` take `Aa` command-line options for triangle attribute and maximum area (used for assigning subdomain IDs).

I will have a PR shortly for these improvements. The third item is a little complicated and probably need a libMesh expert to carefully review. Thanks!

YaqiWang:
Close #2172.

I think the function in the last commit could be beyond the original usage. It can be a very useful tool for checking the integrity of a mesh.

The implementation is tested with a MOOSE mesher using these improvements. I am not sure how I can test these in libMesh if desired.

### DIFF
--- a/include/mesh/replicated_mesh.h
+++ b/include/mesh/replicated_mesh.h
@@ -27,6 +27,7 @@
 
 // C++ Includes
 #include <cstddef>
+#include <unordered_map>
 
 namespace libMesh
 {
@@ -227,6 +228,32 @@ public:
                         bool verbose=true,
                         bool use_binary_search=true,
                         bool enforce_all_nodes_match_on_boundaries=false);
+
+  /**
+   * Return IDs of representative elements of all disconnected subdomains.
+   * Subdomains are considered connected only when they are sharing at least
+   * one d-1 dimensional object (side in 2D, face in 3D), where d is
+   * the mesh dimension.
+   * The optional argument can be used for getting the subdomain IDs of all
+   * elements with element IDs as the index.
+   * This function cannot be called for a mesh with hanging nodes from
+   * adaptive mesh refinement.
+   */
+  std::vector<dof_id_type> get_disconnected_subdomains(std::vector<subdomain_id_type> * subdomain_ids = nullptr) const;
+
+  /**
+   * Return all points on boundary.
+   * The key of the returned unordered map is the ID of a representative
+   * element of all disconnected subdomains. Subdomains are considered
+   * connected only when they are sharing at least one d-1 dimensional object
+   * (side in 2D), where d is the mesh dimension.
+   * The size of the unordered map value is the number of disconnected
+   * boundaries for a subdomain. Boundaries are considered
+   * connected only when they are sharing a d-2 dimensional object.
+   * This function currently only works for 2D meshes.
+   * The points of each boundary are ordered to form an enclosure.
+   */
+  std::unordered_map<dof_id_type, std::vector<std::vector<Point>>> get_boundary_points() const;
 
 public:
   /**
--- a/src/mesh/replicated_mesh.C
+++ b/src/mesh/replicated_mesh.C
@@ -1417,5 +1417,184 @@ dof_id_type ReplicatedMesh::n_active_elem () const
                                                  this->active_elements_end()));
 }
 
+std::vector<dof_id_type>
+ReplicatedMesh::get_disconnected_subdomains(std::vector<subdomain_id_type> * subdomain_ids) const
+{
+  // find number of disconnected subdomains
+  std::vector<dof_id_type> representative_elem_ids;
+
+  // use subdomain_ids as markers for all elements to indicate if the elements
+  // have been visited. Note: here subdomain ID is unrelated with element
+  // subdomain_id().
+  std::vector<subdomain_id_type> subdomains;
+  if (!subdomain_ids)
+    subdomain_ids = &subdomains;
+  subdomain_ids->clear();
+  subdomain_ids->resize(max_elem_id() + 1, Elem::invalid_subdomain_id);
+
+  // counter of disconnected subdomains
+  subdomain_id_type subdomain_counter = 0;
+
+  // a stack for visiting elements, make its capacity sufficiently large to avoid
+  // memory allocation and deallocation when the vector size changes
+  std::vector<Elem *> list;
+  list.reserve(n_elem());
+
+  // counter of visited elements
+  dof_id_type visited = 0;
+  do
+  {
+    for (const auto & elem : active_element_ptr_range())
+      if ((*subdomain_ids)[elem->id()] == Elem::invalid_subdomain_id)
+      {
+        list.push_back(elem);
+        (*subdomain_ids)[elem->id()] = subdomain_counter;
+        break;
+      }
+
+    dof_id_type min_id = std::numeric_limits<dof_id_type>::max();
+    while (list.size() > 0)
+    {
+      // pop up an element
+      Elem * elem = list.back(); list.pop_back(); ++visited;
+
+      min_id = std::min(elem->id(), min_id);
+
+      for (auto s : elem->side_index_range())
+      {
+        Elem * neighbor = elem->neighbor_ptr(s);
+        if (neighbor != nullptr && (*subdomain_ids)[neighbor->id()] == Elem::invalid_subdomain_id)
+        {
+          // neighbor must be active
+          libmesh_assert(neighbor->active());
+          list.push_back(neighbor);
+          (*subdomain_ids)[neighbor->id()] = subdomain_counter;
+        }
+      }
+    }
+
+    representative_elem_ids.push_back(min_id);
+    subdomain_counter++;
+  }
+  while (visited != n_elem());
+
+  return representative_elem_ids;
+}
+
+std::unordered_map<dof_id_type, std::vector<std::vector<Point>>>
+ReplicatedMesh::get_boundary_points() const
+{
+  if (mesh_dimension() != 2)
+    libmesh_error_msg("Error: get_boundary_points only works for 2D now");
+
+  // find number of disconnected subdomains
+  // subdomains will hold the IDs of disconnected subdomains for all elements.
+  std::vector<subdomain_id_type> subdomains;
+  std::vector<dof_id_type> elem_ids = get_disconnected_subdomains(&subdomains);
+
+  std::unordered_map<dof_id_type, std::vector<std::vector<Point>>> boundary_points;
+
+  // get all boundary sides that are to be erased later during visiting
+  // use a comparison functor to avoid run-time randomness due to pointers
+  struct boundary_side_compare
+  {
+    bool operator()(const std::pair<const Elem *, unsigned int> & lhs,
+                    const std::pair<const Elem *, unsigned int> & rhs) const
+      {
+        if (lhs.first->id() < rhs.first->id())
+          return true;
+        else if (lhs.first->id() == rhs.first->id())
+        {
+          if (lhs.second < rhs.second)
+            return true;
+        }
+        return false;
+      }
+  };
+  std::set<std::pair<const Elem *, unsigned int>, boundary_side_compare> boundary_elements;
+  for (const auto & elem : active_element_ptr_range())
+    for (auto s : elem->side_index_range())
+      if (elem->neighbor_ptr(s) == nullptr)
+        boundary_elements.insert(std::pair<const Elem *, unsigned int>(elem, s));
+
+  while (!boundary_elements.empty())
+  {
+    // get the first entry as the seed
+    const Elem * eseed = boundary_elements.begin()->first;
+    unsigned int sseed = boundary_elements.begin()->second;
+
+    // get the subdomain ID that these boundary sides attached to
+    subdomain_id_type subdomain_id = subdomains[eseed->id()];
+
+    // start visiting the mesh to find all boundary nodes with the seed
+    std::vector<Point> bpoints;
+    const Elem * elem = eseed;
+    unsigned int s = sseed;
+    std::vector<unsigned int> local_side_nodes = elem->nodes_on_side(s);
+    while (true)
+    {
+      std::pair<const Elem *, unsigned int> side(elem, s);
+      libmesh_assert(boundary_elements.find(side) != boundary_elements.end());
+      boundary_elements.erase(side);
+
+      // push all nodes on the side except the node on the other end of the side (index 1)
+      for (unsigned int i = 0; i < local_side_nodes.size(); ++i)
+        if (i != 1)
+          bpoints.push_back(*static_cast<const Point *>(elem->node_ptr(local_side_nodes[i])));
+
+      // use the last node to find next element and side
+      const Node * node = elem->node_ptr(local_side_nodes[1]);
+      std::set<const Elem *> neighbors;
+      elem->find_point_neighbors(*node, neighbors);
+
+      // if only one neighbor is found (itself), this node is a cornor node on boundary
+      if (neighbors.size() != 1)
+        neighbors.erase(elem);
+
+      // find the connecting side
+      bool found = false;
+      for (const auto & neighbor : neighbors)
+      {
+        for (auto ss : neighbor->side_index_range())
+          if (neighbor->neighbor_ptr(ss) == nullptr && !(elem == neighbor && s == ss))
+          {
+            local_side_nodes = neighbor->nodes_on_side(ss);
+            // we expect the starting point of the side to be the same as the end of the previous side
+            if (neighbor->node_ptr(local_side_nodes[0]) == node)
+            {
+              elem = neighbor;
+              s = ss;
+              found = true;
+              break;
+            }
+            else if (neighbor->node_ptr(local_side_nodes[1]) == node)
+            {
+              elem = neighbor;
+              s = ss;
+              found = true;
+              // flip nodes in local_side_nodes because the side is in an opposite direction
+              auto temp(local_side_nodes);
+              local_side_nodes[0] = temp[1];
+              local_side_nodes[1] = temp[0];
+              for (unsigned int i = 2; i < temp.size(); ++i)
+                local_side_nodes[temp.size() + 1 - i] = temp[i];
+              break;
+            }
+          }
+        if (found)
+          break;
+      }
+      if (!found)
+        libmesh_error_msg("ERROR: mesh topology error on visiting boundary sides");
+
+      // exit if we reach the starting point
+      if (elem == eseed && s == sseed)
+        break;
+    }
+    boundary_points[elem_ids[subdomain_id]].push_back(bpoints);
+  }
+
+  return boundary_points;
+}
 
 } // namespace libMesh
--- a/include/mesh/mesh_triangle_holes.h
+++ b/include/mesh/mesh_triangle_holes.h
@@ -176,6 +176,51 @@ private:
   std::vector<unsigned int> _segment_indices;
 };
 
+
+
+
+
+/**
+ * A class for defining a 2-dimensional region for Triangle.
+ */
+class TriangleInterface::Region
+{
+public:
+  /**
+   * Constructor
+   */
+  Region(const Point & center,
+         Real attribute = 0,
+         Real max_area = -std::numeric_limits<Real>::max())
+  : _center(center),
+    _attribute(attribute),
+    _max_area(max_area) {}
+
+  Point inside() const { return _center; }
+
+  Real attribute() const { return _attribute; }
+
+  Real max_area() const { return _max_area; }
+
+private:
+  /**
+   * Arbitrary (x,y) location inside the region
+   */
+  const Point _center;
+
+  /**
+   * Attribute of the region
+   * Default value for attribute is zero.
+   */
+  Real _attribute;
+
+  /**
+   * Maximum area that is allowed for all triangles in this region
+   * Default negative maximum area means that no area constraint will be imposed.
+   */
+  Real _max_area;
+};
+
 } // namespace libMesh
 
 #endif // LIBMESH_HAVE_TRIANGLE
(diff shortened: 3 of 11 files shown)


jwpeterson:
Hi Yaqi, I am on travel at the moment and don’t have time to look into this closely, but the proposed changes all sound fine to me. One comment is (I’m pretty sure) that we don’t have any CIVET testing of -disable-strict-lgpl builds on libmesh PRs by default so your new code is not actually being tested here. You may want to consider adding some testing either here or on MOOSE if you plan to use this capability more extensively.

YaqiWang:
@jwpeterson Yes, I am going to add a test in MOOSE and ask @permcody if possible to to add a `-disable-strict-lgpl` test target. I need multiple eyes on the third commit because I am not sure how libMesh setup local connectivities. The boundary side visiting algorithm only works for 2D. I have not thought of the interface for 3D carefully.

YaqiWang:
This is ready for review.

include/mesh/replicated_mesh.h
```diff
@@ -228,6 +228,13 @@ class ReplicatedMesh : public UnstructuredMesh
                         bool use_binary_search=true,
                         bool enforce_all_nodes_match_on_boundaries=false);
 
+  /**
+   * Return all points on boundary
+   * Note: The key of the returned map is a representative interior point of all disconnected subdomains.
+   *       The size of vector value is the number of disconnected boundaries for a subdomain.
+   */
+  std::map<Point, std::vector<std::vector<Point>>> get_boundary_points() const;
```
roystgnr: This might be too hard to do soon on a distributed mesh, but should we move it to UnstructuredMesh and just rely on a MeshSerializer?  That might be good for flexibility when users want to triangulate a coarse mesh but expect to do a lot of refinement afterwards.

src/mesh/replicated_mesh.C
```diff
@@ -1417,5 +1417,151 @@ dof_id_type ReplicatedMesh::n_active_elem () const
                                                  this->active_elements_end()));
 }
 
+std::map<Point, std::vector<std::vector<Point>>>
+ReplicatedMesh::get_boundary_points() const
+{
+  // find number of disconnected subdomains
+  std::vector<Point> interior_points;
+  std::vector<subdomain_id_type> subdomains(max_elem_id() + 1, Elem::invalid_subdomain_id);
```
roystgnr: Why copy subdomain_id() values into a vector?  They don't change during this function and you always have easy access to the element to query them directly.

roystgnr: Oh, never mind, you're tracking unvisited elements.

roystgnr: Still kind of confusing.  I'd have either used a vector<bool> or an unordered_set.  Neither is so much better as to make it worth changing this though.

include/mesh/replicated_mesh.h
YaqiWang: Does MeshSerializer sends the mesh parts on other processors to the master so that master can have the full mesh? I was creating this function for mesh generation, the stage where mesh has not been distributed, refined, etc. I used some low-level functions in this function and my understanding on how libMesh sets up the local connectivity for various types of cells may not be accurate. I also have not thought of 3D carefully. Is the function written in a way easily understandable?

src/mesh/replicated_mesh.C
roystgnr: No, I'm misunderstanding again.  You're not using subdomain_id(), you're defining your own subdomains based on mesh connectivity?

I should probably stop the stream-of-consciousness here until I'm finished with the readthrough.

YaqiWang: I should have added an inline comment on this variable like `markers for all elements to indicate if they have been visited`.

src/mesh/replicated_mesh.C
```diff
@@ -1417,5 +1417,151 @@ dof_id_type ReplicatedMesh::n_active_elem () const
...
+            if (found)
+              break;
+          }
+      }
+      if (!found)
+        libmesh_error_msg("ERROR: mesh topology error on visiting boundary sides");
+
+      // exit if we reach the starting point
+      if (elem == eseed && s == sseed)
+        break;
```
YaqiWang: This algorithm provides an ordered boundary points that form an enclosure. The ordering is kind of important. Number of separated boundaries is also a product of this algorithm.

include/mesh/replicated_mesh.h
roystgnr: The second optional argument to MeshSerializer lets you serialize to only rank 0, so you can run the algorithm on one processor with all the mesh parts sent there but without having to send them to other processors too.  If the mesh is currently serial, like when it hasn't been distributed yet, the function just tests that and then does nothing.  The destructor re-distributes the mesh (if it hadn't been serial to begin with) so the state change isn't permanent.  Basically you construct it at the beginning of whatever scope you need to be serial and then you're done.

src/mesh/replicated_mesh.C
roystgnr: A comment that these subdomain ids are unrelated to the subdomain_id() would have been helpful.  This seems obvious on closer reading but some times people like me need help seeing the obvious.  :-)

src/mesh/replicated_mesh.C
```diff
@@ -1417,5 +1417,151 @@ dof_id_type ReplicatedMesh::n_active_elem () const
...
+      }
+    while (list.size() > 0)
+    {
+      // pop up an element
+      Elem * elem = list.back(); list.pop_back(); ++visited;
+
+      for (auto s : elem->side_index_range())
+      {
+        Elem * neighbor = elem->neighbor_ptr(s);
+        if (neighbor != nullptr && subdomains[neighbor->id()] == Elem::invalid_subdomain_id)
```
roystgnr: This looks like a standard "fill" sweep through a component of the mesh, but note that you can *also* find an unencountered neighbor in the case of AMR meshes, where a coarse active element at a hanging node has an inactive neighbor in that direction.  That could actually be a problem, in fact; a sweep starting from level N elements is never going to find level N+1 or higher elements.  We probably don't care (this is to be run before any adaptive refinement) but it might be a good idea to throw ```libmesh_assert(neighbor->active())``` or something in before putting neighbor on the list in case someone tries to use this method where it's inapplicable.

src/mesh/replicated_mesh.C
```diff
@@ -1417,5 +1417,151 @@ dof_id_type ReplicatedMesh::n_active_elem () const
...
+          list.push_back(neighbor);
+          subdomains[neighbor->id()] = subdomain_counter;
+        }
+      }
+    }
+    std::cout << std::endl;
+    subdomain_counter++;
+  }
+  while (visited != n_elem());
+  // now subdomain_counter stores the number of disconnected subdomains
```
roystgnr: Technically we've found the number of components that aren't connected through element sides.  You could also have e.g. a mesh with two elements connected at only one corner, and this algorithm would claim that those are two separate subdomains, even though topologically a connection at one point is still a connection.  Is that okay?

include/mesh/replicated_mesh.h
permcody: Is Point reliable with the less than operator? I don't think so. I've definitely run into issues using Points as keys in maps in the past. I would recommend against using a map there. Perhaps would should implement a hash function and use unordered_map instead?

roystgnr: Point inherits TypeVector::operator<, which does the standard dictionary ordering.  And std::map accepts any total ordering, so Point should be fine as a key there.  Any more detailed recollection on those issues?  The only trouble I ever remember having was in the context of Point values changing: with coarsening and re-refinement I don't think we can guarantee that a fine node will be recreated in exactly the same place, due to floating point error.  That epsilon error would propagate to the centroids used here.  But I don't get the impression that in this context it's important for the choice or exact location of the representative points to be identical from run to run?

src/mesh/replicated_mesh.C
```diff
@@ -1417,5 +1417,151 @@ dof_id_type ReplicatedMesh::n_active_elem () const
...
+
+  std::vector<std::vector<std::vector<Point>>> boundary_points(subdomain_counter);
+
+  // following boundary visiting algorithm only works with 2D
+  if (mesh_dimension() != 2)
+    libmesh_error_msg("Error: get_boundary_points only works for 2D now");
+
+  // get all boundary sides
+  // use a comparison functor to avoid run-time randomness due to pointers
```
roystgnr: Now I'm second-guessing my "it won't matter if the particular representative point changes from run to run" claim earlier.  Elem pointers are going to stay fixed within the lifetime of any one call to a const function, so you actually are worrying about reproducing the exact ordering of these results between one run and another?

src/mesh/replicated_mesh.C
```diff
@@ -1417,5 +1417,151 @@ dof_id_type ReplicatedMesh::n_active_elem () const
...
+    const Elem * elem = eseed;
+    unsigned int s = sseed;
+    std::vector<unsigned int> local_side_nodes = elem->nodes_on_side(s);
+    while (true)
+    {
+      boundary_elements.erase(std::pair<const Elem *, unsigned int>(elem, s));
+
+      // push all nodes on the side except the last one
+      for (unsigned int i = 0; i < local_side_nodes.size() - 1; ++i)
+        bpoints.push_back(*dynamic_cast<const Point *>(elem->node_ptr(local_side_nodes[i])));
```
roystgnr: This shouldn't require the dynamic_cast, should it?

src/mesh/replicated_mesh.C
YaqiWang: Will add that into the inline comment.

src/mesh/replicated_mesh.C
YaqiWang: Yes, subdomains connected through d-1 elems only are considered connected. We can actually separate this part of the code into another function which can have options to choose d-1 to 0 connection.

include/mesh/replicated_mesh.h
permcody: This would have been 8+ years ago. The problem was that I was using it to lookup values where the Point precision might change by something close to machine precision (perhaps because of the way I was generating points using trigonometric functions). I seem to recall that if the x-value (the first test) was off by machine precision sometimes a large change in y or z wouldn't trigger the expected result because I would have taken the "wrong" branch on the x-coordinate first. I don't think this is a problem with Point per se, It's more of an issue with floating point equality, which you have to be cognizant of when using it as a map key.

src/mesh/replicated_mesh.C
```diff
@@ -1417,5 +1417,151 @@ dof_id_type ReplicatedMesh::n_active_elem () const
...
+
+      // use the last node to find next element and side
+      const Node * node = elem->node_ptr(local_side_nodes[local_side_nodes.size() - 1]);
```
roystgnr: I don't think this is going to work on second-order elements.  On Quad8/Quad9/Tri6, the last index nodes_on_side returns is actually the local index of the mid-edge node, not of either of the corners.  Maybe go in the opposite direction?  In theory we haven't promised this, but in practice local_side_nodes[0] will always be a vertex.  Or if direction is important then at least *currently* local_side_nodes[1] is always the other vertex in 2D or another vertex in 3D.

src/mesh/replicated_mesh.C
YaqiWang: Assertion is a good idea although we possibly can make this algorithm work with adatptively refined meshes.

src/mesh/replicated_mesh.C
```diff
@@ -1417,5 +1417,151 @@ dof_id_type ReplicatedMesh::n_active_elem () const
...
+      std::set<const Elem *> neighbors;
+      elem->find_point_neighbors(*node, neighbors);
+
+      // if only one neighbor is found (itself), this node is a cornor node on boundary
+      if (neighbors.size() != 1)
```
roystgnr: Should this be ==, not !=?

src/mesh/replicated_mesh.C
YaqiWang: Yes.

src/mesh/replicated_mesh.C
roystgnr: Oh, no, I get it.  If we *do* have more neighbors then one of them is what we're looking for and we can skip ourselves, if we *don't* then we're turning a corner on this element and we need to examine ourselves.

include/mesh/replicated_mesh.h
roystgnr: Oh, yes, the ordering here is definitely infinitely sensitive to Point coordinates in cases where two points are aligned on the yz plane.

YaqiWang: I see. Looks like MeshSerializer is a really powerful tool.

Using std::map directly creates a mapping for a representative point of a subdomain to its boundaries. We do not have to use map though.

roystgnr: IIRC there's no good way around that, though, because any ordering here is essentially a function from 1D to fill 2D, and the inverse of such a function can't be continuous.

src/mesh/replicated_mesh.C
YaqiWang: Probably not.

src/mesh/replicated_mesh.C
roystgnr: Okay.  Perhaps rather than using a map key of a representative element centroid, then, you want to use something like the lowest element id?  Centroids might change by epsilon whenever we change certain mesh generation algorithms, whereas we can try to keep initial id numbrering consistent and libMesh+MOOSE already allow you to disable later renumbering.

include/mesh/replicated_mesh.h
permcody: I might suggest adding a hash function to Point (or TypeVector) so that it is more amenable to unordered_map usage (perhaps there's already one there, or the default hashing is fine). We have been running benchmarks here in the office and have discovered that `unordered_map` is almost always faster than `map` for both heavy insert and lookup cases. We are starting to lean on unordered_map more often when ordering isn't required (it often isn't).

YaqiWang: Good to know. Here the points should be well separated because each point is within one subdomain of a set of disconnected subdomains.

src/mesh/replicated_mesh.C
roystgnr: And now that I'm done reading back and forth, I think this is the only place where either I'm really confused or there actually is a bug.  All the other changes I've suggested are either minor tweaks, or at worst are debatable API design decisions.

include/mesh/replicated_mesh.h
YaqiWang: Yes, using a representative element pointer `Elem *`  instead of `Point` could be a better idea.

src/mesh/replicated_mesh.C
YaqiWang: Maybe just a `Elem *`?

include/mesh/replicated_mesh.h
roystgnr: Oh, I'm definitely on board with using unordered_map everywhere you don't need the ordering.  Just pointing out that if we *do* need a consistent ordering, there's no way to get it from point locations that isn't sensitive to perturbations.  I suspect an unordered_map based on lowest element ID is the way to go here.

src/mesh/replicated_mesh.C
YaqiWang: Using id will make the value invalid after the mesh is renumbered possibly?

include/mesh/replicated_mesh.h
roystgnr: > Here the points should be well separated because each point is within one subdomain of a set of disconnected subdomains.

Well, the issue here is that well separated in space doesn't imply well separated in ordering.  If you have elements that have the exact same X coordinate, then just rotating your mesh by epsilon degrees can completely switch around the ordering of those elements' centroids.  I think using Elem id (lowest id, if you really want maximum reproducability) is a safer way to go, given that you're already relying on it later and so anybody who wants reproduceable orderings from this is going to be disabling id renumbering already.

src/mesh/replicated_mesh.C
YaqiWang: This is the place I am not certain as well. It will be nice if we can do some unit test on this function, but I do not know how to do it. I think it is probably worthwhile to come up with something.

src/mesh/replicated_mesh.C
roystgnr: Hmm... that may be a concern.  Usually anything you could do that would trigger a renumbering would also change geometry enough to invalidate the previous output, though, in which case you'd have to call this again anyway.  I guess all_second_order() can trigger a renumbering despite not changing element geometry ... but it *would* insert new boundary points and so you wouldn't want to reuse get_boundary_points() output in that case either.  So I don't see any cases where we'd have a *practical* concern.

YaqiWang:
There is a small issue if we want a separate function to return a vector of element IDs each for a disconnected subdomain and we want the function in `UnstructuredMesh`. Because we will need to use `MeshSerializer`, it will be constructed twice one in the separate function another in `get_boundary_points`. I guess constructing and destructing of `MeshSerializer` could be costly. @roystgnr do you still want these functions in `UnstructuredMesh`, or we do not have a separate function, or we duplicate the codes in two functions, or we just construct `MeshSerializer` twice?

roystgnr:
Ah, that's a good point.  When the first MeshSerializer destructs then a DistributedMesh is going to have its remote elements deleted and then it'll have to re-serialize them again pointlessly when the second MeshSerializer is constructed.

Hmmm... no, wait.  We won't be calling the element ids function *before* the boundary points function, we'll be calling it *inside* the boundary points function.  So the boundary MeshSerializer will get constructed first, the element ids MeshSerializer will see that we're already serialized and will be a no-op, and the mesh will remain continuously serialized (to rank 0) until the end of the boundary points function.

That's how it should work in theory anyway; I don't think we've got any current code paths that construct a MeshSerializer while another one is already active.  We can try it that way and then if it doesn't work in practice that'll be a bug for me to fix.

YaqiWang:
I see. The inner construction & destruction will be null operation supposedly.

YaqiWang:
It is very hacky  if we put these functions in `UnstructuredMesh` because we cannot mark them `const` due to the usage of `MeshSerializer`. I think the better way is probably to put these functions into `ReplicatedMesh`.  We can leave a pure virtual function in `UnstructuredMesh`. In `DistributedMesh`, we first construct a replicated mesh and then call the functions in the `ReplicatedMesh`. Not sure if this constructor `ReplicatedMesh (const UnstructuredMesh &other_mesh)` can be used to convert a distributed mesh to replicated mesh.

YaqiWang:
I have another question, are the element IDs unique across processors in a distributed mesh? If not, returning a map with IDs as the keys may not be a good idea.

roystgnr:
Gah, I didn't even think about the const correctness issue.

Perhaps we should move this to UnstructuredMesh, but test ```this->is_serial()``` at the start and die with ```libmesh_not_implemented()``` if not?  That would make it possible for DistributedMesh users to do this without a mesh copy; serializing and unserializing would be inefficient but not quite that inefficient.

Anyway, you've got a reasonable workaround for now, so we can leave that for a later date.  Let's try to figure out the second-order element issue and merge this once that's straight?

> I have another question, are the element IDs unique across processors in a distributed mesh?

They are.  The plain id() values change through renumbering unlike unique_id(), but they're consistent across processors at any given time.

> It will be nice if we can do some unit test on this function, but I do not know how to do it. I think it is probably worthwhile to come up with something.

Agreed.  Maybe just do the sort of thing we do in tests/mesh/slit_mesh_test.C: manually construct a small mesh so we can easily see exactly what output we expect from it?

YaqiWang:
Yeap, I am testing this function with a second-order mesh. I will come up with a test in libmesh for this function. The test may not be able to catch all kinds of meshes where nodes are ordered strangely locally, but it is better than nothing and I will make sure users will hit an error so that we can extend the code when needed.

YaqiWang:
A dummy question: if I add a source file in libmesh/tests/mesh folder, how do I compile the unit test to link that file in?

YaqiWang:
I added a unit test in a source file but I probably miss something so that that unit test can be linked in. I did put the test into an existing source file and made sure the code is behaving as expected. The unit did help on finding a bug when the boundary side flips direction.

jwpeterson:
> A dummy question: if I add a source file in libmesh/tests/mesh folder, how do I compile the unit test to link that file in?

There is a list of source files in tests/Makefile.am, you will need to add your new filename there and then run ./bootstrap in the top level of the libmesh repo. You need to be using the same autotools versions as are in contrib/autotools, otherwise there will be a very large diff. You can run `./bootstrap --build-autotools` to first build the required versions and then add the resulting directory  (~/projects/libmesh/contrib/autotools/bin) to your `$PATH`.

include/mesh/replicated_mesh.h
jwpeterson: > Is Point reliable with the less than operator? I don't think so.

It's definitely not reliable to use `set<Point>`, `map<Point, X>` etc! If there was a way to prevent these from even compiling, I would be in favor of using it...

YaqiWang:
Thanks @jwpeterson. The change in makefile is pushed. @roystgnr I addressed your comments. Let us leave this function in ReplicatedMesh only for a while.

YaqiWang:
Any more comments?

roystgnr:
nodes_on_side can't have its output flipped around, but we can construct side elements that do the sort of node-ordering-independent comparisons you're talking about.

Elem::side_neighbors I'd call edge_neighbors, but either way it sounds like a good idea.  Give me some advance notice when you need to extend this to 3D and I'll take a quick crack at it.

This looks good; nice work!

PR #2174, opened 2019-07-08
Add component-wise *= operator to NumericVector

loganharbour:
Also removed optional flag for `v` [here](https://github.com/loganharbour/libmesh/blob/6a1cadac908c43b5ca76c85ac09bc12ccbfd9e55/include/numerics/numeric_vector.h#L416)

### DIFF
--- a/include/numerics/numeric_vector.h
+++ b/include/numerics/numeric_vector.h
@@ -409,7 +409,7 @@ public:
   NumericVector<T> & operator /= (const T a) { this->scale(1./a); return *this; }
 
   /**
-   * Computes the pointwise division of this vector's entries by another's,
+   * Computes the component-wise division of this vector's entries by another's,
    * \f$ u_i \leftarrow \frac{u_i}{v_i} \, \forall i\f$
    *
    * \returns A reference to *this.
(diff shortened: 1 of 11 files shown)


include/numerics/numeric_vector.h
```diff
@@ -390,6 +390,14 @@ class NumericVector : public ReferenceCountedObject<NumericVector<T>>,
    */
   NumericVector<T> & operator *= (const T a) { this->scale(a); return *this; }
 
+  /**
+   * Computes the pointwise multiplication of this vector's entries by another's,
```
dschwen:
```suggestion
   * Computes the componentwise multiplication of this vector's entries by another's,
```

loganharbour: Few others as well

roystgnr:
Have you tried a build with all those subclasses enabled yet?

:+1: if so, though if beggars can be choosers I'd feel happier to see a unit test too.

loganharbour:
Yep, I've made all of the inherited classes are built without error.

I actually looked around to add a unit test, but it appears that there are no simple operation tests for the `NumericVector` inherited classes. Only localization tests. I'm not really sure where I'd start. Any suggestions?

roystgnr:
Operation tests can go in the same NumericVectorTest class in numeric_vector_test.h; the tiny size of that file is because we've also been too negligent about unit testing, not because we intended it to be localize-specific.

But let's just make sure CI is happy with this PR (sorry I didn't notice it needed manual activation!) and then merge as-is; you can put any unit tests into a new PR if you find time.

jwpeterson:
I added tests for non-LGPL contrib and Trilinos, which should make sure the `LaspackVector` and `EpetraVector` changes at least compile... 

@loganharbour I also sent you an invite to join the "collaborators" team so that your PRs can be tested automatically in the future. Since the mathematics here aren't complicated, I'd say we could merge without adding a unit test...

PR #2176, opened 2019-07-11
New unit test for add_system() with extra ints

roystgnr:
This turned out not to be the explanation for the failures in
https://github.com/idaholab/moose/pull/13706 but now that the test is
already written we should use it to avoid future regressions.

### DIFF
--- a/src/base/dof_object.C
+++ b/src/base/dof_object.C
@@ -172,16 +172,20 @@ void DofObject::set_n_systems (const unsigned int ns)
 
   const unsigned int nei = this->n_extra_integers();
   const unsigned int header_size = ns + bool(nei);
-  index_buffer_t new_buf(header_size + nei, header_size);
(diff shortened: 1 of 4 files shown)


roystgnr:
... this new test isn't actually running.  We disabled node_test.C when we disabled the Node copy constructor, and elem_test.C isn't pulling in dof_object_test.h, so the only extra integers tests being run are the ones in tests/mesh/

I'll fix that and then I'll bet this new test fails.

roystgnr:
This seems to fix https://github.com/idaholab/moose/pull/13706 for me too; I'll merge if it passes CI.  @YaqiWang, want to try it out too?

YaqiWang:
Trying it out, will let you know shortly.

YaqiWang:
Yes, it does fix the test and produce the expected results! Thanks @roystgnr .

PR #2178, opened 2019-07-18
Changed libMesh configuration to support --download-selpc in PETSc

fdkong:
There are a couple of SLEPc use cases:

(1) Use prefix, and SLEPc will be installed to $prefix, and SLEPC_DIR=$prefix. PETSC_ARCH most likely is empty. $SLEPC_DIR/include has `slepcversion.h `

(2) Install SLEPc in place, SLEPC_DIR and PETSC_ARCH will be set properly. $SLEPC_DIR/include has 
`slepcversion.h`

(3) Let PETSc configuration --download-slepc, and make a prefix installation.  PETSC_DIR=$prefix, SLEPC_DIR=$PETSC_DIR and PETSC_ARCH should be empty.  $SLEPC_DIR/include has 
`slepcversion.h`

(4) PETSc configure with --download-slepc, and make an in-place installation. I would like to set SLEPC_DIR=$PETSC_DIR, and PETSC_ARCH is NOT empty. And then $SLEPC_DIR/include does not have  `slepcversion.h`. Instead, `slepcversion.h` lives in $SLEPC_DIR/$PETSC_ARCH.

libMesh was able to support (1), (2) and (3). The current PR is trying to support (4) that is what I usually use.

Not sure is it a reasonable use case or not?

### DIFF
--- a/m4/slepc.m4
+++ b/m4/slepc.m4
@@ -31,11 +31,12 @@ AC_DEFUN([CONFIGURE_SLEPC],
                 [
                   AC_CHECK_HEADER([$SLEPC_DIR/include/slepcversion.h],
                                   [SLEPC_INCLUDE="-I$SLEPC_DIR/include"],
-                                  [
-                                    AC_MSG_RESULT(<<< Invalid "\$SLEPC_DIR" detected (slepcversion.h not found). SLEPc disabled. >>>)
-                                    unset SLEPC_DIR
-                                    enableslepc=no
-                                  ])
+                                  [AC_CHECK_HEADER([$SLEPC_DIR/$PETSC_ARCH/include/slepcversion.h],
+                                                   [SLEPC_INCLUDE="-I$SLEPC_DIR/$PETSC_ARCH/include"],
+                                                   [AC_MSG_RESULT(<<< Invalid "\$SLEPC_DIR" detected (slepcversion.h not found). SLEPc disabled. >>>)
+                                                    unset SLEPC_DIR
+                                                    enableslepc=no]
+                                  )])
                 ])
 
             dnl I didn't check this code branch since I don't use a
@@ -51,11 +52,20 @@ AC_DEFUN([CONFIGURE_SLEPC],
           [
             dnl Similar to petsc, we need the slepc version number.
             dnl Note slepc will most likely only work with the corresponding version of petsc
-            slepcmajor=`grep "define SLEPC_VERSION_MAJOR" $SLEPC_DIR/include/slepcversion.h | sed -e "s/#define SLEPC_VERSION_MAJOR[ ]*//g"`
-            slepcminor=`grep "define SLEPC_VERSION_MINOR" $SLEPC_DIR/include/slepcversion.h | sed -e "s/#define SLEPC_VERSION_MINOR[ ]*//g"`
-            slepcsubminor=`grep "define SLEPC_VERSION_SUBMINOR" $SLEPC_DIR/include/slepcversion.h | sed -e "s/#define SLEPC_VERSION_SUBMINOR[ ]*//g"`
-            slepcversion=$slepcmajor.$slepcminor.$slepcsubminor
+            AC_CHECK_HEADER([$SLEPC_DIR/include/slepcversion.h],
+                            [
+                               slepcmajor=`grep "define SLEPC_VERSION_MAJOR" $SLEPC_DIR/include/slepcversion.h | sed -e "s/#define SLEPC_VERSION_MAJOR[ ]*//g"`
+                               slepcminor=`grep "define SLEPC_VERSION_MINOR" $SLEPC_DIR/include/slepcversion.h | sed -e "s/#define SLEPC_VERSION_MINOR[ ]*//g"`
+                               slepcsubminor=`grep "define SLEPC_VERSION_SUBMINOR" $SLEPC_DIR/include/slepcversion.h | sed -e "s/#define SLEPC_VERSION_SUBMINOR[ ]*//g"`
+                             ],
+                             [
+                               slepcmajor=`grep "define SLEPC_VERSION_MAJOR" $SLEPC_DIR/$PETSC_ARCH/include/slepcversion.h | sed -e "s/#define SLEPC_VERSION_MAJOR[ ]*//g"`
+                               slepcminor=`grep "define SLEPC_VERSION_MINOR" $SLEPC_DIR/$PETSC_ARCH/include/slepcversion.h | sed -e "s/#define SLEPC_VERSION_MINOR[ ]*//g"`
+                               slepcsubminor=`grep "define SLEPC_VERSION_SUBMINOR" $SLEPC_DIR/$PETSC_ARCH/include/slepcversion.h | sed -e "s/#define SLEPC_VERSION_SUBMINOR[ ]*//g"`
+                             ]
+            )
 
+            slepcversion=$slepcmajor.$slepcminor.$slepcsubminor
             dnl Older PETSc (3.3 and earlier) might not work if the version numbers don't match exactly.
             oldpetsc=no
 
@@ -78,6 +88,7 @@ AC_DEFUN([CONFIGURE_SLEPC],
                   [test -r ${SLEPC_DIR}/conf/slepc_common_variables],    [includefile=${SLEPC_DIR}/conf/slepc_common_variables],
                   [test -r ${SLEPC_DIR}/conf/slepc_variables],           [includefile=${SLEPC_DIR}/conf/slepc_variables],
                   [test -r ${SLEPC_DIR}/lib/slepc/conf/slepc_variables], [includefile=${SLEPC_DIR}/lib/slepc/conf/slepc_variables], dnl 3.6.0
+                  [test -r ${SLEPC_DIR}/${PETSC_ARCH}/lib/slepc/conf/slepc_variables], [includefile=${SLEPC_DIR}/${PETSC_ARCH}/lib/slepc/conf/slepc_variables],
                   [AC_MSG_RESULT([<<< SLEPc configuration failed.  Could not find slepcconf/slepc_variables file. >>>])
                    enableslepc=no])
 
--- a/m4/slepc.m4
+++ b/m4/slepc.m4
@@ -16,56 +16,65 @@ AC_DEFUN([CONFIGURE_SLEPC],
 
   AS_IF([test "$enableslepc" !=  no],
         [
-          dnl If SLEPC_DIR is not set, we should try PETSC_DIR
-          dnl since SLPEc could live in the same place as PETSc
-          AS_IF([test "x$SLEPC_DIR" = x],
+          dnl If SLEPC_DIR is not set, but $PETSC_DIR is, it's possible
+          dnl that there is a SLEPc within PETSc.  Not sure about using
+          dnl "export" here but the same thing is done in petsc.m4 so it
+          dnl probably doesn't hurt anything.
+          AS_IF([test "x$SLEPC_DIR" = "x" && test "x$PETSC_DIR" != "x"],
                 [export SLEPC_DIR=$PETSC_DIR])
 
-          dnl Test to see if SLEPC_DIR and petscversion were set by user or
-          dnl autodetection.  If not set, then disable slepc, print a message.
-          AS_IF([test "x$SLEPC_DIR" = "x" || test "x$petscversion" = "x"],
-                 [
-                   dnl Test to see if SLEPC_DIR set by user.  If not set, then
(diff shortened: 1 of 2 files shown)


m4/slepc.m4
```diff
@@ -51,11 +52,20 @@ AC_DEFUN([CONFIGURE_SLEPC],
           [
             dnl Similar to petsc, we need the slepc version number.
             dnl Note slepc will most likely only work with the corresponding version of petsc
-            slepcmajor=`grep "define SLEPC_VERSION_MAJOR" $SLEPC_DIR/include/slepcversion.h | sed -e "s/#define SLEPC_VERSION_MAJOR[ ]*//g"`
-            slepcminor=`grep "define SLEPC_VERSION_MINOR" $SLEPC_DIR/include/slepcversion.h | sed -e "s/#define SLEPC_VERSION_MINOR[ ]*//g"`
-            slepcsubminor=`grep "define SLEPC_VERSION_SUBMINOR" $SLEPC_DIR/include/slepcversion.h | sed -e "s/#define SLEPC_VERSION_SUBMINOR[ ]*//g"`
-            slepcversion=$slepcmajor.$slepcminor.$slepcsubminor
+            AC_CHECK_HEADER([$SLEPC_DIR/include/slepcversion.h],
```
jwpeterson: Hmm... calling `AC_CHECK_HEADER` twice on the same file will probably work OK, but this actually triggers a [test compilation](https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Generic-Headers.html) so it's good to not call it more than is strictly necessary. 

Would you mind setting a flag, e.g. `slepc_in_petsc_arch`, and checking that subsequently instead of calling `AC_CHECK_HEADER` again? Otherwise, this change seems fine to me.

fdkong: Thanks, John. I will do that.

I am thinking should we just check  PETSC_DIR and PETSC_ARCH if slepc is just in the same place as PETSc, and then we can simply set SLEPC_DIR=PETSC_DIR, and then continue configuration.

jwpeterson: Sounds fine to me. I don't use this type of configuration at all, so as far as I'm concerned, it's up to you exactly how you want to handle it.

roystgnr: You mean set SLEPC_DIR=PETSC_DIR/PETSC_ARCH so our later tests are looking in the right place?  That sounds like it would make the code slightly simpler but significantly more confusing, since conceptually SLEPC_DIR is a const global variable.  I like the existing PR better, and I like John's suggestion better still.

But I second "as far as I'm concerned, it's up to you exactly how you want to handle it".  I hate m4 and I'm not proud of our existing m4 and I'm hardly going to hold new m4 to arbitrarily higher standards.

fdkong: > You mean set SLEPC_DIR=PETSC_DIR/PETSC_ARCH so our later tests are looking in the right place? That sounds like it would make the code slightly simpler but significantly more confusing, since conceptually SLEPC_DIR is a const global variable. I like the existing PR better, and I like John's suggestion better still.

What I meant was that if SLEPC_DIR is not set by users, should we try to detect if PETSc carries with SLEPc or not?  In order to accomplish this goal, we could simply set SLEPC_DIR=$PETSC_DIR when users do not set SLEPC_DIR.

roystgnr: Oh, sorry, I get it now.  Yeah, I like that.  The more autodetection the better, at least in cases like this where options like ```--disable-slepc``` exist to allow users to override the autodetection.

fdkong:
Now we are able to auto pick up SLEPc from PETSC_DIR.  I am not familiar  with m4, so could you guys help me check?

fdkong:
@roystgnr Could you take a look at this again?

m4/slepc.m4
```diff
@@ -16,26 +16,47 @@ AC_DEFUN([CONFIGURE_SLEPC],
...
-                ],
-                [
-                  AC_CHECK_HEADER([$SLEPC_DIR/include/slepcversion.h],
-                                  [SLEPC_INCLUDE="-I$SLEPC_DIR/include"],
+                 [
+                   dnl Test to see if SLEPC_DIR set by user.  If not set, then
+                   dnl try to autodetect in a default directory
+                   AC_CHECK_HEADER([/usr/lib/slepc/slepcversion.h],
+                                  [
+                                   export SLEPC_DIR=/usr/lib/slepc
```
jwpeterson: Hmm... I don't know what the effect (if any) of `export`ing variables in m4 files is, but we don't normally do this. I *don't* think this will set `SLEPC_DIR` in the user's environment, in case that's what you were attempting to do.

m4/slepc.m4
```diff
@@ -16,26 +16,47 @@ AC_DEFUN([CONFIGURE_SLEPC],
...
-                  enableslepc=no
-                  AC_MSG_RESULT(<<< SLEPc disabled.  Please set your "\$SLEPC_DIR" environment variable correctly to enable SLEPc. >>>)
-                ],
-                [
-                  AC_CHECK_HEADER([$SLEPC_DIR/include/slepcversion.h],
-                                  [SLEPC_INCLUDE="-I$SLEPC_DIR/include"],
+                 [
+                   dnl Test to see if SLEPC_DIR set by user.  If not set, then
+                   dnl try to autodetect in a default directory
+                   AC_CHECK_HEADER([/usr/lib/slepc/slepcversion.h],
```
jwpeterson: Looking in `/usr/lib/slepc` as a default is what we did before, but is that really what we want to do in general? 

It seems that it runs the risk of picking up some system installation of SLEPc when the user didn't want SLEPc enabled at all. If this ever happened it would probably be extremely rare, but also very confusing to the user.

m4/slepc.m4
```diff
@@ -16,26 +16,47 @@ AC_DEFUN([CONFIGURE_SLEPC],
...
+                                  ],
                                   [
-                                    AC_MSG_RESULT(<<< Invalid "\$SLEPC_DIR" detected (slepcversion.h not found). SLEPc disabled. >>>)
-                                    unset SLEPC_DIR
                                     enableslepc=no
+                                    AC_MSG_RESULT(<<< SLEPc disabled.  Please set your "\$SLEPC_DIR" environment variable correctly to enable SLEPc. >>>)
                                   ])
+                ],
+                [
+                  AC_CHECK_HEADER([$SLEPC_DIR/include/slepcversion.h],
```
jwpeterson: I think I follow your logic here but it's a bit convoluted. It's also a bit unsatisfying to repeat the code that looks in `/usr/lib/slepc` for headers like this. I can probably take a crack at improving it if you are getting sick of this PR...

m4/slepc.m4
roystgnr: > it would probably be extremely rare,

/usr/lib/slepc is the default install directory in Debian and Ubuntu.

> but also very confusing to the user.

If they installed slepc, I think using it would be natural and failing to do so would be confusing, no?

jwpeterson: > If they installed slepc, I think using it would be natural and failing to do so would be confusing, no?

Installed SLEPc but *didn't* set `$SLEPC_DIR`? IMO we can't assume they want SLEPc just because it's on their system, sometimes you don't have 100% control over what's installed on your system...

m4/slepc.m4
fdkong: As said before, I am not familiar with m4 so that I can not make the thing more efficient. I would appreciate if you could help.

m4/slepc.m4
fdkong: I think if you do not have SLEPC_DIR, and do not have PETSC_DIR, and do not enable SLEPc, then we should not do anything.

I want to let libMesh take a try to check if PETSc is complied with SLEPc. Then SLEPc will be picked up in the same place as PETSc.  I assume if users pass ``--download-slepc" in PETSc, then users want to use SLEPc. Note that ` --download-slepc` does not set up SLEPC_DIR.

roystgnr: I'll consider myself outvoted.  I still think asking users who installed-but-don't-want SLEPc to use "--disable-slepc" is more user-friendly than asking users who installed-and-want SLEPc to manually specify paths.  Should we make the same change to PETSc autodetection for consistency?  Or for **real** consistency, I notice that we also automatically activate CPPUnit, curl, GLPK, Tecplot, Trilinos, VTK, and XDR based on what we find in /usr...  ;-)

jwpeterson:
Hi @fdkong, would you mind trying the latest version I pushed to your fork? This works for me using a "standalone" (make installed) SLEPc, but I didn't test with the `$PETSC_ARCH` use case. I did also end up keeping the `/usr/lib/slepc` case, as we support that for PETSc so it makes sense to also support it for SLEPc.

The main change, besides hopefully making the header checking logic a bit simpler, is that in the case where `$PETSC_DIR` is not set by the user *and* PETSc has not been found in `/usr/lib/petsc` (otherwise we would have set `$PETSC_DIR`!) then I think we can safely stop looking for SLEPc. Hopefully that logic makes sense to you as well...

fdkong:
> Hi @fdkong, would you mind trying the latest version I pushed to your fork? This works for me using a "standalone" (make installed) SLEPc, but I didn't test with the `$PETSC_ARCH` use case. I did also end up keeping the `/usr/lib/slepc` case, as we support that for PETSc so it makes sense to also support it for SLEPc.
> 
> The main change, besides hopefully making the header checking logic a bit simpler, is that in the case where `$PETSC_DIR` is not set by the user _and_ PETSc has not been found in `/usr/lib/petsc` (otherwise we would have set `$PETSC_DIR` ourselves!) then I think we can safely stop looking for SLEPc. Hopefully that logic makes sense to you as well...

Thanks so much @jwpeterson. I tried:

(1) standalone

(2) $PETSC_ARCH

(3) $PETSC_DIR (where SLEPc lives in the same place as PETSc)

Your beautiful m4 code works for all these cases. Let us merge it in because I cannot wait.

roystgnr:
I think this might have regressed SLEPc detection for me?  I have $PETSC_DIR and $SLEPC_DIR defined as natural, I have $SLEPC_DIR/$PETSC_ARCH/include/slepcconf.h, but when I try configure I get

```
configure:36076: checking whether we can compile a trivial SLEPc program
configure:36103: mpicxx -c  -std=gnu++11 -I/home/roystgnr/LIBRARIES/petsc-3.9.3/include -I/home/roystgnr/LIBRARIES/petsc-3.9.3/linux-gnu/include -I/
home/roystgnr/LIBRARIES/petsc-3.9.3/include -I/home/roystgnr/LIBRARIES/petsc-3.9.3/linux-gnu/include -I/home/roystgnr/LIBRARIES/slepc-3.9.2/include
 conftest.cpp >&5
In file included from /home/roystgnr/LIBRARIES/slepc-3.9.2/include/slepcst.h:16:0,
                 from /home/roystgnr/LIBRARIES/slepc-3.9.2/include/slepceps.h:16,
                 from /home/roystgnr/LIBRARIES/slepc-3.9.2/include/slepcsvd.h:16,
                 from /home/roystgnr/LIBRARIES/slepc-3.9.2/include/slepc.h:14,
                 from conftest.cpp:157:
/home/roystgnr/LIBRARIES/slepc-3.9.2/include/slepcsys.h:23:23: fatal error: slepcconf.h: No such file or directory
```

roystgnr:
Yeah, the regression definitely dates to this PR.  a20bc3d still autodetects the installation in my ```$SLEPC_DIR```; 09565cc fails to.

m4/slepc.m4
```diff
@@ -16,46 +16,80 @@ AC_DEFUN([CONFIGURE_SLEPC],
...
+                        ])
                 ])
-
-            dnl I didn't check this code branch since I don't use a
-            dnl PETSC_ARCH, but running AC_CHECK_HEADER on slepcconf.h
-            dnl should be pretty safe... from what I can tell it just
-            dnl #defines a few things.
-            AS_IF([test "x$PETSC_ARCH" != "x"],
-                  [AC_CHECK_HEADER([$SLEPC_DIR/$PETSC_ARCH/include/slepcconf.h],
-                                   [SLEPC_INCLUDE="$SLEPC_INCLUDE -I$SLEPC_DIR/$PETSC_ARCH/include"])])
```
roystgnr: Okay, this deletion looks like the problem.  Without it we never notice that slepcconf.h is under the $PETSC_ARCH subdirectory, we never add that entry to $SLEPC_INCLUDE, and the later build fails.

jwpeterson:
Sorry about this. Would you mind pushing a fix since your system is already set up to test it?

roystgnr:
Already testing it now, yeah.

roystgnr:
And it should be safe for non-$PETSC_ARCH setups, I think, since it's the same code we had working with them before.

PR #2180, opened 2019-07-21
A small bug fix revealed by uniformly refining a mesh

YaqiWang:
This is a follow-on fix for PR #2173.

### DIFF
--- a/src/mesh/replicated_mesh.C
+++ b/src/mesh/replicated_mesh.C
@@ -1451,6 +1451,8 @@ ReplicatedMesh::get_disconnected_subdomains(std::vector<subdomain_id_type> * sub
         (*subdomain_ids)[elem->id()] = subdomain_counter;
         break;
       }
+    // we should be able to find a seed here
+    libmesh_assert(list.size() > 0);
 
     dof_id_type min_id = std::numeric_limits<dof_id_type>::max();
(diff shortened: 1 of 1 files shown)


src/mesh/replicated_mesh.C
```diff
@@ -1476,7 +1478,7 @@ ReplicatedMesh::get_disconnected_subdomains(std::vector<subdomain_id_type> * sub
     representative_elem_ids.push_back(min_id);
     subdomain_counter++;
   }
-  while (visited != n_elem());
+  while (visited != n_active_elem());
```
jwpeterson: `n_active_elem()` is a relatively expensive (O(N. elem)) function compared to `n_elem()` so it would be best to avoid calling it at every loop iteration. Since I don't think the number of active elements in the mesh changes during this `do` loop, it would be better to compute this value once beforehand.

YaqiWang: Good to know. Thought it is just a getter to a member variable... I can submit a further fix if desired.

jwpeterson: I already have one ready. Github went down for some reason, but I will submit it shortly.

YaqiWang: Great!

PR #2181, opened 2019-07-21
Improve quality of Legendre polynomial implementation

jwpeterson:
My long term goal is to have have some Legendre-polynomial-based elements available for DG. We currently only optionally use Legendre polynomials in the infinite elements (there they are the so-called "radial" basis functions). The previous formulation was based on [Horner's method](https://en.wikipedia.org/wiki/Horner%27s_method) with hard-coded coefficients, but this was both difficult to verify and somewhat inaccurate at higher p levels, at least based on my tests which compared it to C++17's `std::legendre()` and analytic values computed via Wolfram alpha. 

This PR adds a unit test for Legendre infinite elements and replaces the Horner method with an implementation based on Legendre polynomial recurrence relations which is shorter, simpler, and more numerically stable. In a later PR, I plan to add the corresponding 1D (and 2D/3D tensor product) DG elements, which will share this implementation.

### DIFF
--- a/src/fe/inf_fe_legendre_eval.C
+++ b/src/fe/inf_fe_legendre_eval.C
@@ -22,300 +22,108 @@
 #include "libmesh/inf_fe.h"
 #include "libmesh/inf_fe_macro.h"
 
-namespace libMesh
-{
+using namespace libMesh;
 
 // Anonymous namespace for local helper functions
 namespace {
 
-Real legendre_eval(Real v,  unsigned i)
+// Evaluate P_n(x) using the three term recurrence relation.  Note
+// that using the three term recurrence is more accurate than the
+// unrolled Horner scheme that was used here previously and requires a
+// lot less code.
+// TODO: C++17 will have std::legendre() in <cmath>, so eventually we
+// can just use that.
+Real legendre_eval(unsigned int n, Real x)
 {
-  libmesh_assert (-1.-1.e-5 <= v && v < 1.);
+  Real p0 = 1;
+  Real p1 = x;
+  if (n == 0)
+    return p0;
 
-  switch (i)
+  unsigned int i = 1;
+  while (i < n)
     {
-    case 0:
-      return 1.;
-
-    case 1:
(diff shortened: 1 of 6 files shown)


roystgnr:
I see a ton of compilation errors, all apparently of the form
```
../../tests/fe/inf_fe_legendre_test.C: In member function ‘void InfFELegendreTest::testDifferentOrders()’:
../../tests/fe/inf_fe_legendre_test.C:58:7: error: converting to ‘std::tuple<unsigned int, unsigned int, double>’ from initializer list would use explicit constructor ‘constexpr std::tuple< <template-parameter-1-1> >::tuple(_UElements&& ...) [with _UElements = {int, int, double}; <template-parameter-2-2> = void; _Elements = {unsigned int, unsigned int, double}]’
       };
```

jwpeterson:
OK, weird. I tested it on GCC 7.4 locally and of course whatever CIVET is using. The use of tuples is only for convenience, we can take them out if their is a backwards compatibility issue...

jwpeterson:
Oh, maybe just try changing the tuple type to `<int,int,double>` and see if that fixes it. I guess it is complaining about an int -> unsigned int conversion maybe.

roystgnr:
Looks like the vector<tuple> construction is requiring implicit tuple constructors, but constructing a tuple directly from arguments was still only an explicit constructor in C++11?

roystgnr:
I tried changing the literals from e.g. "1" to "1u" and no luck.  Even the non-converting direct initialization seems to be explicit.

roystgnr:
https://stackoverflow.com/questions/12436586/tuple-vector-and-initializer-list

jwpeterson:
OK... we have some code elsewhere in the library that does `vec.emplace_back(i,j,val);` so I'm pretty sure that works in GCC 5.4. Maybe I could just change it to that.

roystgnr:
Tedious, but yeah, that should work.  Or just use a struct?  Either way, thanks!

PR #2182, opened 2019-07-22
Fixes for Elem::second_order_equivalent_type().

jwpeterson:
* QUADSHELL8 is currently the 'full ordered' element for the QUADSHELL
  family.
* There are currently no 2nd-order triangular shell elements, so
  TRISHELL3 should just return itself.
* More descriptive error message for unsupported element types.


jwpeterson:
Merging because I think CIVET may have stopped running the Test debug job.

PR #2183, opened 2019-07-22
Change from std::tuple to struct

jwpeterson:
Apparently the previous code was illegal in C++11 since std::tuples use [explicit constructors](https://stackoverflow.com/questions/12436586/tuple-vector-and-initializer-list)? It may have worked on the newer compilers I tried it on due to changes in later standards filtering down into -std=c++11 code.

Refs #2181.

### DIFF
--- a/tests/fe/inf_fe_legendre_test.C
+++ b/tests/fe/inf_fe_legendre_test.C
@@ -43,8 +43,26 @@ public:
   void setUp() {}
   void tearDown() {}
 
-  typedef std::vector<std::tuple<unsigned, unsigned, Real>> TabulatedVals;
-  typedef std::vector<std::tuple<unsigned, unsigned, Point>> TabulatedGrads;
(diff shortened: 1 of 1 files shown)


jwpeterson:
@roystgnr can you give this a try with your antique hipster compiler and let me know if it works?

roystgnr:
Recompiling now.  Might take a minute, because my CPU had to underclock itself to compensate for that *burn*.

roystgnr:
Yeah, gcc 5.4.0 is happy now.

jwpeterson:
Merged before debug mode finished; this was a unit test change and unit tests don't even run in debug mode.

permcody:
FYI - The INL experienced a power issue this morning which killed one phase of power. That issue has been resolved but we are still experiencing issues in our data center.  You may not see activity for CIVET for a few hours until we get things resolved.

PR #2184, opened 2019-07-22
Only call n_active_elem() once outside the loop predicate.

jwpeterson:
Refs #2180.


jwpeterson:
@permcody is CIVET back to working? I've tried invalidating this PR a couple times but it doesn't seem to have any effect.

jwpeterson:
Merging as this is basically a bugfix for recently added code.

PR #2185, opened 2019-07-22
Boost multiprecision

### LINKED ISSUE #617: Quadruple precision support is broken

roystgnr:
Sorry this is so huge; it turned out to be much more work than I anticipated when, a couple months ago, I thought "I should work on some old tickets in the down time between juggling other tasks".

I'm in no great hurry to merge it if anyone has the desire to review it but not the time, or if anyone finds problems or has disagreements.  It fixes #617 and has a number of more general improvements, but it's been a work in progress for a while and it's not urgent.

### DIFF
--- a/tests/quadrature/quadrature_test.C
+++ b/tests/quadrature/quadrature_test.C
@@ -191,8 +191,13 @@ public:
     // There are 3 different families of quadrature rules for tetrahedra
     QuadratureType qtype[3] = {QCONICAL, QGRUNDMANN_MOLLER, QGAUSS};
 
+    int end_order = 7;
+    // Our higher order tet rules were only computed to double precision
+    if (quadrature_tolerance < 1e-16)
+      end_order = 2;
+
     for (int qt=0; qt<3; ++qt)
-      for (int order=0; order<7; ++order)
+      for (int order=0; order<end_order; ++order)
         {
           std::unique_ptr<QBase> qrule = QBase::build(qtype[qt],
                                                       /*dim=*/3,
--- a/Makefile.am
+++ b/Makefile.am
@@ -44,7 +44,7 @@ AM_CPPFLAGS      = -DLIBMESH_IS_COMPILING_ITSELF \
                    $(libmesh_optional_INCLUDES) \
                    -I$(top_builddir)/include # required for libmesh_version.h
 
-LIBS             = $(libmesh_optional_LIBS)
+LIBS             = $(libmesh_optional_LIBS) $(libmesh_precision_LIBS)
 
 # additional files which must be included in 'make dist'
 EXTRA_DIST = reference_elements \
--- a/configure
+++ b/configure
@@ -870,6 +870,7 @@ NODEPRECATEDFLAG
 CFLAGS_DVL
 CXXFLAGS_DVL
 GXX_VERSION
+libmesh_precision_LIBS
 PKG_CONFIG
 libmesh_LDFLAGS
 CXXCPP
@@ -29366,6 +29367,8 @@ else
 fi
 
 
+libmesh_precision_LIBS=""
+
 if test "x$enablesingleprecision" != "xno"; then :
 
         if test "x$enabletripleprecision" != "xno"; then :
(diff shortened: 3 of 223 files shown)


jwpeterson:
tenor

tests/quadrature/quadrature_test.C
```diff
@@ -511,6 +515,10 @@ public:
   template <QuadratureType qtype, Order order, unsigned int exactorder>
   void test3DWeights ()
   {
+    // Our Gauss-Lobatto rules are only computed to double precision
```
jwpeterson: Wait, why? They should have like 32 *accurate* decimal digits.

roystgnr: We've got a bunch of 32 decimal place literals, but they're all suffixed with L, which truncates to 20 digits when it interprets them as long double.  If we really want all the precision we can get, we'll need a new suffix that we can shim to the boost suffix when we have it or to L when we don't.

Since those unit tests use ```TOLERANCE*sqrt(TOLERANCE)``` and I only set ```TOLERANCE = 1e-11```, 20 digits was good enough for most cases.  But we don't have 20 digits everywhere.

For Gaussian quadrature on Tets we only have 15 digits at second order:
https://github.com/libMesh/libmesh/blob/17be2c62003504c8555d7af2400fc7fd2d5e5924/src/quadrature/quadrature_gauss_3D.C#L85
and likewise on Triangles at seventh order:
https://github.com/libMesh/libmesh/blob/17be2c62003504c8555d7af2400fc7fd2d5e5924/src/quadrature/quadrature_gauss_2D.C#L338

And... what the hell.  I could swear that I saw a similar case in the Gauss-Lobatto rules, and that that prompted me to rule out those specific unit tests, but now I can't find it.  Let me reenable them and see what happens.

roystgnr: I do get failures again.  In both 2D and 3D, for both ```SIXTH``` and ```SEVENTH``` order.

OH HELL.  There's a ```0.1``` in gauss_lobatto_quadrature_1D.C I need to change to ```1/Real(10)```, isn't there?

roystgnr: And that fixes the tests for me.  I'll upload in a minute.

roystgnr: Thanks for catching this!!

jwpeterson: > Thanks for catching this!!

Sure, but I think you are the one who caught the but that I apparently left as an easter egg in 3998bf26f.

jwpeterson: > If we really want all the precision we can get, we'll need a new suffix that we can shim to the boost suffix when we have it or to L when we don't.

I scanned the boost.multiprecision introduction and "Literal types support" docs. It looks like only the only supported user-defined literals are currently `_cppi`, `_cppui`, etc. which are all for integral types. 

According to [this](https://stackoverflow.com/questions/33043809/what-is-the-precision-of-cpp-dec-float-50) it looks like you can pass a string to e.g. the `boost::multiprecision::cpp_dec_float_50` constructor. Maybe it's possible to come up with a simple `to_real()` macro that would do the right thing (stringification, concatenation, etc.) in all cases from the same input?

```C
#ifdef LIBMESH_DEFAULT_QUAD_PRECISION
#define to_real(x) boost::multiprecision::cpp_dec_float_50("#x")
#else
#define to_real(x) long double(x##L)
#endif
```
```
_weights[0] = to_real(6.6666666666666666666666666666667e-02);
```

jwpeterson:
This PR also reminds me that we are still on Boost 1.61 (May 2016) internally. Not suggesting we should also update that in *this* PR, but they are up to version 1.70 now.

tests/quadrature/quadrature_test.C
roystgnr: Just doing stringification in a macro seems clever.  I'd just been *assuming* that we needed to use user-defined literals, but I guess as long as the cpp_dec_float constructor and the conversion to float128 are both constexpr then it'll work out just fine in the end.

roystgnr:
Not a bad idea.  If we do that eventually though then we might as well do the user defined literal thing; IIRC 1.70 is where I noticed they had the Q suffix defined.

roystgnr:
Oh, good boy Civet.  I hadn't thought to do "make installcheck" or to try a complex build, and although I had been testing without MPI it looks like I didn't repeat that build after my final changes to the parallel headers.

roystgnr:
I'm going to avoid doing any interactive rebasing, just toss fix commits on the end, for fear of screwing up any reviewing in progress, but I can do a final reorders+fixups cleanup on the commit history after the final state is in a happy place.

roystgnr:
Throwing on some more tests for compiler variations, downstream apps, valgrind etc.

roystgnr:
Those seem to be unrelated failures.  Looks like the default tests are currently on PETSc 3.10.5, but the tests with old gcc+clang are using PETSc 3.7.6, and the latter doesn't support one of the command line options that @bboutkov used this spring in a new fem_system_ex1 test in #2151.

The PETSc documentation suggests that PCMGSetGalerkin was added in PETSc 3.8, so we can just test for that and skip those last couple tests if we have something older.  While we're at it we probably ought to be skipping the earlier petsc option tests in cases where we don't have PETSc at all.

roystgnr:
Hopefully I didn't screw up the conflict resolution.

After exchanging IMs with @jwpeterson I'm going to feel free to start doing fixups and merge (or merges, if I split some more logically independent parts into separate PRs) soon.

roystgnr:
Okay: everything that has value outside the extended precision context has been split out and merged, and for the handful of ugly ```Point(Real(scalar))``` commits I was dubious about I found a better solution and got rid of them, so the rest I'll merge all in one batch sometime this weekend after confirming that both CI and my own tests are still happy.

PR #2186, opened 2019-07-24
Fix OOB array access in miscellaneous_ex14

roystgnr:
This appears to have been broken for half a year or more.  I don't
imagine we'll be putting a SLEPc+IFEM+Complex build into CI any time
soon, though.


jwpeterson:
Nice catch!

PR #2188, opened 2019-07-25
Fix --disable-amr builds

### LINKED ISSUE #850: Build fails with --disable-amr

roystgnr:
This resolves #850.  Again.  Probably temporarily again.

### DIFF
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -1171,7 +1171,7 @@ void Elem::remove_links_to_me()
 #else
               unsigned int my_s = neigh->which_neighbor_am_i(this);
               libmesh_assert_less (my_s, neigh->n_neighbors());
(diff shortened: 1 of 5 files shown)


src/geom/elem.C
```diff
@@ -1171,7 +1171,7 @@ void Elem::remove_links_to_me()
 #else
               unsigned int my_s = neigh->which_neighbor_am_i(this);
               libmesh_assert_less (my_s, neigh->n_neighbors());
-              libmesh_assert_equal_to (neigh->neighbor(my_s), this);
+              libmesh_assert_equal_to (neigh->neighbor_ptr(my_s), this);
```
jwpeterson: Ha, I thought we had found all these!

PR #2189, opened 2019-07-25
Assert real parents

### LINKED ISSUE #1047: Deprecate using parent() for interior_parent()
We've supported proper interior_parent() for a couple releases now, we use it exclusively internally; at some point we should start asserting that parent() is really a parent, then we can remove the code which has special cases to handle interior_parents in disguise.

roystgnr:
We long ago added interior_parent() and removed any old support for the previous parent()-really-returning-interior_parent() hack, so let's add a few more assertions to make debugging any old third party code easier, and that should completely resolve #1047.

### DIFF
--- a/include/base/libmesh_common.h
+++ b/include/base/libmesh_common.h
@@ -287,42 +287,42 @@ extern bool warned_about_auto_ptr;
 
 #define libmesh_assert_equal_to_msg(expr1,expr2, msg)                   \
   do {                                                                  \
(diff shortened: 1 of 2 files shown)


roystgnr:
Oh, this first crack doesn't work at all; we can't test derived functions in the output of parent() when we may call parent() from base class constructors.  Let me just limit the tests to set_parent().

PR #2190, opened 2019-07-26
Generalize subdomain comparison for var groups

lindsayad:
This allows user code to be more general. They can create
a std::set instance in all cases and pass it in; if the set
is empty then this code will do the right thing and
recognize that there are no restrictions. The
their_active_subdomains was already doing the right thing
because it is a nullptr if the previously added subdomain
set was empty (e.g. see Variable::implicity_active())

### DIFF
--- a/include/systems/system.h
+++ b/include/systems/system.h
@@ -1085,7 +1085,9 @@ public:
 
   /**
    * Adds the variable \p var to the list of variables
-   * for this system.
+   * for this system. If \p active_subdomains is either \p nullptr
+   * (the default) or points to an empty set, then it will be assumed that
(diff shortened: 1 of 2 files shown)


roystgnr:
:+1: Looks fine by me, but let's document the new behavior in system.h?  Not that our previous add_variable documentation was any good, but IMHO at least the previous "nullptr means forget about subdomain restrictions" behavior was semi-intuitive.  But "treat the empty set as if it were the universal set" is a bit confusing as a user API, even if it's pretty sensible as an internal optimization.

lindsayad:
Documented. If I don’t hear any screams for a bit I’m gonna merge

PR #2191, opened 2019-07-26
Provide two functions to inquire the existence of mesh integers

YaqiWang:
It is typically the adder and the getter are in different places, and the getter can fail when the adder has not been called. These functions provide users the way to gracefully handle the case of missing integers.

### DIFF
--- a/src/mesh/mesh_base.C
+++ b/src/mesh/mesh_base.C
@@ -192,6 +192,17 @@ unsigned int MeshBase::get_elem_integer_index(const std::string & name) const
 
 
 
+bool MeshBase::has_elem_integer(const std::string & name) const
+{
+  for (auto & entry : _elem_integer_names)
+    if (entry == name)
+      return true;
+
+  return false;
+}
+
+
+
 unsigned int MeshBase::add_node_integer(const std::string & name)
 {
   for (auto i : index_range(_node_integer_names))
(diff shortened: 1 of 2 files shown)


YaqiWang:
It is must be very difficult for people to survive in old days without git, fully-featured compiler and automatic code testing...

src/mesh/mesh_base.C
```diff
@@ -192,6 +192,17 @@ unsigned int MeshBase::get_elem_integer_index(const std::string & name) const
 
 
 
+bool MeshBase::has_elem_integer(const std::string & name) const
+{
+  for (auto i : index_range(_elem_integer_names))
```
roystgnr: Why use indices here, rather than a range for over the container?  I wasn't planning on changing the container type but better safe than sorry.

roystgnr:
I recently learned from friends as far apart as Australia and TACC that there are still numerical software authors out there who resist using version control.  Not git specifically, version control in general.  I was dumbfounded.

src/mesh/mesh_base.C
YaqiWang: Only because I was copying the code in another function without much thinking. Changed to the range loop.

YaqiWang:
@roystgnr any more comment? I ‘d like this to be included in the coming libmesh update in moose.

roystgnr:
Oh, sorry, I missed your update somehow.  Of course this is good to go now, thanks and sorry for the delay.

YaqiWang:
Not a problem. Thanks @roystgnr .

PR #2194, opened 2019-07-27
Improve Jacobi polynomials implementation

jwpeterson:
This is related to the work in #2181, and is the real payoff in terms of line removal. All three sets of radial basis function values (`LEGENDRE`, `JACOBI_20_00`, `JACOBI_30_00`) for `InfFE`s can be generated more accurately by a single implementation of 30 or so lines based on the three term recurrence relation than 1000+ lines of tabulated coefficients.


tests/fe/inf_fe_jacobi_test.C
```diff
@@ -0,0 +1,369 @@
...
+
+  void setUp()
+  {
+    ////////////////////////////////////////////////////////////////////////////////
+    // Arbitrarily selected JACOBI_20_00 reference values.
+    ////////////////////////////////////////////////////////////////////////////////
+    tabulated_vals[std::make_pair(FIRST, JACOBI_20_00)] =
+      {
+        // These values are the same as Legendre
+        {0,9,0.0550016}, {1,5,0.41674},   {2,8,0.0147376}, {3,7,0.111665},
```
roystgnr: Not that it's insufficient for this purpose, but why only 5 or 6 signficant decimals?

jwpeterson: Laziness? I just copy/pasted the values that were printed to `std::cout` by the test, so they were in that default format which seemed to have about 4 or 5 digits by default.

PR #2195, opened 2019-07-27
QNodal update

jwpeterson:
I changed the metric for computing the optimal weights slightly. These rules, which are for serendipity elements, are only exact for linears. Their main purpose is to guarantee an SPD approximation to the true mass matrix when nodal quadarture is employed, so the weights are simply chosen to be as closely proportional to the true mass matrix values as possible.


roystgnr:
Could you go into more detail in the comments?  Even with the context from our IMs these still look like "magic numbers" to me; with only the context from the comments in the code I'd be *completely* lost.

roystgnr:
Thanks!

PR #2202, opened 2019-07-31
Why are you always failing?

jwpeterson:
This fixes an issue where, after the first time you catch a negative Jacobian exception and keep going, you no longer get exceptions for subsequent negative Jacobians. Instead, the code just returns with incomplete results whenever a negative Jacobian occurs since `failing==true` from then on.

`¯\_(ツ)_/¯`

### DIFF
--- a/src/fe/fe_map.C
+++ b/src/fe/fe_map.C
@@ -513,6 +513,7 @@ void FEMap::compute_single_point_map(const unsigned int dim,
                   {
                     failing = true;
                     elem->print_info(libMesh::err);
+                    failing = false;
                     if (calculate_xyz)
                       {
                         libmesh_error_msg("ERROR: negative Jacobian " \
@@ -589,6 +590,7 @@ void FEMap::compute_single_point_map(const unsigned int dim,
                   {
                     failing = true;
                     elem->print_info(libMesh::err);
+                    failing = false;
                     libmesh_error_msg("Encountered invalid 1D element!");
                   }
                 else
@@ -635,6 +637,7 @@ void FEMap::compute_single_point_map(const unsigned int dim,
                   {
                     failing = true;
                     elem->print_info(libMesh::err);
+                    failing = false;
                     libmesh_error_msg("Encountered invalid 1D element!");
                   }
                 else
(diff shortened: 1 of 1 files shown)


jwpeterson:
@permcody, would you mind taking a look at these [test failures](https://civet.inl.gov/job/375714/)? 

I think what's happening is that they used to somehow not "crash" because only a single exception would be thrown (not sure whether they used to somehow converge or what). Now that I fixed the logic, there are multiple exceptions thrown (correctly, as there are multiple elements with negative/zero Jacobians) and for some reason the tests now "crash," apparently due to dt being cut below dtmin. The test spec contains `should_crash = False` but maybe that should really be set to `True`? I can't really see how the test would ever reach the `end_time` without crashing, given the displacement function that is being used.

permcody:
Do you remember how hard we tried to get a stinkin' negative Jacobian? It
was almost impossible. IIRC - this test has an inverted element or some
other extreme cash that probably should have never converged. Now that you
have fixed the logic, we might have to look at a less drastic test to see
if we can introduce a "recoverable" failure. I'd be fine with temporarily
disabling that test if we can't figure it out. Ideally we'd create
something a little better instead. Glad this is heading in the right
direction. Perhaps if we disable that test, we might make it further and
uncover a few tests in the contact module that fail that might be useful
for building a test case.

Switching the Boolean will make the test pass but the test is just bad.
There's no way to recover from the failure, which is what the solver is
trying to do.

jwpeterson:
I think another idea that would work would be to change `end_time` to 0.75 for those 2 tests. That way they will still hit the negative Jacobian exception when they try and take the first full step out to time t=1, the timestep will be cut to dt=0.25 and then the solve will succeed, and the test will finish. I think this will still whether we can catch the negative Jacobian exception and continue, which is the point of the test.  With the displacement function used, we'll never be able to reach time t=1, just come arbitrarily close to it as is seen in the error message. I'll see if I can make a MOOSE PR...

roystgnr:
> when they try and take the first full step out to time t=1

Is that documented behavior?  I remember when writing adaptive time steppers I always had the final step get truncated if it would take you past tmax.  That's pretty much mandatory if you intend to use u(tmax) for QoI evaluation or error evaluation and you're letting t vary continuously.  But I guess if you're only doing time step reduction upon convergence failure (another option I found helpful) then it's more reasonable to assume that your user has set (tmax-tmin)/deltat to be non-integral on purpose.

jwpeterson:
I'm not sure why, but somehow this pulled idaholab/moose@cac8d4a26a and not the latest master? I'm trying again.

jwpeterson:
Failed test is unrelated/seems to have been failing recently in debug mode.

PR #2203, opened 2019-08-01
Add jacobian_tolerance on DirichletBoundary class.

jwpeterson:
Set Jacobian tolerance of FEMap from DirichletBoundary object while
applying DirichletBCs.

### DIFF
--- a/include/base/dirichlet_boundaries.h
+++ b/include/base/dirichlet_boundaries.h
@@ -181,6 +181,20 @@ public:
   std::unique_ptr<FEMFunctionBase<Gradient>> g_fem;
 
   const System * f_system;
+
+  /**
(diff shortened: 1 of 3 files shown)


jwpeterson:
> Looks good, if it actually fixes the problem. Won't the degenerate J just screw with JxW values and ruin the edge projections on higher order elements, though?

Uh, I really don't know how all the different projections work. I thought that for C0 Lagrange elements  (regardless of order) you don't really have to do projections? It seems to fix the issue for the couple of cases I've tested it on.

roystgnr:
Hmmm...  I guess as long as JxW is never NaN it'll be the same for both the 1x1 mass matrix and forcing in the Lagrange case and you'll get something semi sane.  I wouldn't trust it to give O(h^3) or anything though.

PR #2206, opened 2019-08-02
Bugfix in rb_data_deserialization.C

dknez:
Need to close the file descriptor once we're finised with it.

### DIFF
--- a/src/reduced_basis/rb_data_deserialization.C
+++ b/src/reduced_basis/rb_data_deserialization.C
@@ -99,6 +99,10 @@ void RBEvaluationDeserialization::read_from_file(const std::string & path,
 #endif
 
   load_rb_evaluation_data(_rb_eval, rb_eval_reader, read_error_bound_data);
+
+  int error = close(fd);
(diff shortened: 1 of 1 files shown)


roystgnr:
The capnp APIs only take raw file descriptors, not fstream or anything, so we can't use RAII here?

dknez:
We followed example code such as on here:
https://capnproto.org/cxx.html#example-usage
See after "You might write code like".

They pass "int fd" to the functions, so I guess we need to use raw file descriptors.

jwpeterson:
I think if this became a maintenance issue in the future, you could always write a simple custom wrapping class that calls `close(fd)` in its destructor... unfortunately there [doesn't appear to be a portable way of creating a std filestream from a file descriptor](https://stackoverflow.com/questions/2746168/how-to-construct-a-c-fstream-from-a-posix-file-descriptor).

jwpeterson:
We don't have any libmesh PR testing with capnproto enabled, so I'm just going to merge this.

PR #2207, opened 2019-08-02
Unit test updates

roystgnr:
A new shim function and some new unit tests written as part of #2185, broken apart since they're independent of the rest of the changes there (except for the changes already merged in #2204)


roystgnr:
Oh, and one unit test fix (a failing test should always *fail* in parallel, not lose sync)

roystgnr:
No, wait, there's a few more unit test specific updates that belong here.

PR #2210, opened 2019-08-05
Make mallocs uniformly an error in PetscMatrix

lindsayad:
This adds the setting to the other two init methods that
if a new malloc occurs during MatSetValues then it is an
error. I think this is an improvement because it establishes
uniformity across our init methods and it can prevent users
from running extremely slow simulations. They might hate
it at first, but it would force them to correct their sparsity,
which would definitely be a win for them in the long run.

I fully expect this to cause failures in MOOSE, but they can
manually override if they need to (e.g. I will make those
changes in MOOSE before this gets merged).

### DIFF
--- a/src/numerics/petsc_matrix.C
+++ b/src/numerics/petsc_matrix.C
@@ -333,6 +333,14 @@ void PetscMatrix<T>::init (const numeric_index_type m_in,
 
     }
 
+  // Make it an error for PETSc to allocate new nonzero entries during assembly
+#if PETSC_VERSION_LESS_THAN(3,0,0)
+  ierr = MatSetOption(_mat, MAT_NEW_NONZERO_ALLOCATION_ERR);
+#else
+  ierr = MatSetOption(_mat, MAT_NEW_NONZERO_ALLOCATION_ERR, PETSC_TRUE);
+#endif
+  LIBMESH_CHKERR(ierr);
+
   // Is prefix information available somewhere? Perhaps pass in the system name?
   ierr = MatSetOptionsPrefix(_mat, "");
   LIBMESH_CHKERR(ierr);
@@ -453,6 +461,14 @@ void PetscMatrix<T>::init ()
         default: libmesh_error_msg("Unsupported petsc matrix type");
       }
     }
+
+  // Make it an error for PETSc to allocate new nonzero entries during assembly
+#if PETSC_VERSION_LESS_THAN(3,0,0)
+  ierr = MatSetOption(_mat, MAT_NEW_NONZERO_ALLOCATION_ERR);
+#else
+  ierr = MatSetOption(_mat, MAT_NEW_NONZERO_ALLOCATION_ERR, PETSC_TRUE);
+#endif
+  LIBMESH_CHKERR(ierr);
 
   // Is prefix information available somewhere? Perhaps pass in the system name?
   ierr = MatSetOptionsPrefix(_mat, "");


fdkong:
@lindsayad By default, PETSc will error out when you add new nonzeros without the right preallocation. We do not see that because we turn it off in moose. We have something like this in moose:

` ierr = MatSetOption(_mat, MAT_NEW_NONZERO_ALLOCATION_ERR, PETSC_FALSE);`


What I meant you are doing here will be overwritten by moose. 



```
void
NonlinearSystemBase::computeJacobianInternal(const std::set<TagID> & tags)
{
  // Make matrix ready to use
  activeAllMatrixTags();

  for (auto tag : tags)
  {
    if (!hasMatrix(tag))
      continue;

    auto & jacobian = getMatrix(tag);
#ifdef LIBMESH_HAVE_PETSC
// Necessary for speed
#if PETSC_VERSION_LESS_THAN(3, 0, 0)
    MatSetOption(static_cast<PetscMatrix<Number> &>(jacobian).mat(), MAT_KEEP_ZEROED_ROWS);
#elif PETSC_VERSION_LESS_THAN(3, 1, 0)
    // In Petsc 3.0.0, MatSetOption has three args...the third arg
    // determines whether the option is set (true) or unset (false)
    MatSetOption(
        static_cast<PetscMatrix<Number> &>(jacobian).mat(), MAT_KEEP_ZEROED_ROWS, PETSC_TRUE);
#else
    MatSetOption(static_cast<PetscMatrix<Number> &>(jacobian).mat(),
                 MAT_KEEP_NONZERO_PATTERN, // This is changed in 3.1
                 PETSC_TRUE);
#endif
#if PETSC_VERSION_LESS_THAN(3, 3, 0)
#else
    if (!_fe_problem.errorOnJacobianNonzeroReallocation())
      MatSetOption(static_cast<PetscMatrix<Number> &>(jacobian).mat(),
                   MAT_NEW_NONZERO_ALLOCATION_ERR,
                   PETSC_FALSE);
#endif

```

fdkong:
And this will be helpful when users directly create matrices bypass moose.

lindsayad:
> By default, PETSc will error out when you add new nonzeros without the right preallocation.

I didn't know this was the default...so what you're saying is that these lines of code are not currently necessary? I guess it would still be good to have these lines in case PETSc were to ever change the default.

> What I meant you are doing here will be overwritten by moose.

Yea I'm aware of that. We have three places in `NonlinearSystemBase` where we do that.

dknez:
Yeah, PETSc throws an error by default, so I'm not sure this change is needed? Based on what I read on the PETSc mailing list, older versions of PETSc did not throw an error by default but they switched the default since they figured it's better to throw an error than to run unnecessarily slowly.

fdkong:
> > By default, PETSc will error out when you add new nonzeros without the right preallocation.
> 
> I didn't know this was the default...so what you're saying is that these lines of code are not currently necessary? I guess it would still be good to have these lines in case PETSc were to ever change the default.
> 
> >

The changes are still very useful to make libmesh and moose behavior consistent regardless PETSc version.

jwpeterson:
> The changes are still very useful to make libmesh and moose behavior consistent regardless PETSc version.

@fdkong in your opinion, are they also beautiful changes?

roystgnr:
My recollection is the same as @dknez, but I agree with @fdkong that we should merge this anyway.

PR #2213, opened 2019-08-08
Slepc $PETSC_ARCH fix

roystgnr:
This was mistakenly deleted by #2178, which broke SLEPc configuration
for installs with a separate $PETSC_ARCH subdirectory.

### DIFF
--- a/configure
+++ b/configure
@@ -36010,6 +36010,16 @@ fi
 
 fi
 
+                        if test "x$PETSC_ARCH" != "x"; then :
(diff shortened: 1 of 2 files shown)


fdkong:
Thanks @roystgnr,  I will test your branch to check if it is still working for me.

fdkong:
It works for me.

roystgnr:
Thanks for double-checking, @fdkong!  Sorry I hadn't done the same for you!

PR #2214, opened 2019-08-08
Implement MONOMIAL_VEC

lindsayad:
Pretty straightforward synthesis from MONOMIAL and LAGRANGE_VEC. I added an example similar to vector_fe_ex1 but using MONOMIAL_VEC.


lindsayad:
For @aeslaughter

include/fe/fe.h
```diff
@@ -58,13 +58,19 @@ struct FEOutputType
 template<>
 struct FEOutputType<LAGRANGE_VEC>
 {
-  typedef RealGradient type;
+  typedef RealVectorValue type;
```
roystgnr: This is just for readability, right?  The old code was confusing, not actually broken in some subtle way?

lindsayad: correct

src/fe/fe_monomial_vec.C
```diff
@@ -0,0 +1,845 @@
...
+template <>
+RealGradient
+FE<0, MONOMIAL_VEC>::shape_deriv(const ElemType type,
+                                 const Order order,
+                                 const unsigned int i,
+                                 const unsigned int j,
+                                 const Point & p)
+{
+  Real value = FE<0, MONOMIAL>::shape_deriv(type, order, i, j, p);
+  return libMesh::RealGradient(value);
```
roystgnr: Ha!  But here you're back to writing RealGradient again.  The confusion will never be eradicated.

lindsayad: Haha I will change it!

examples/vector_fe/vector_fe_ex5/vector_fe_ex5.C
```diff
@@ -0,0 +1,143 @@
...
+    libMesh::out << " " << argv[i];
+
+  libMesh::out << std::endl << std::endl;
+
+  // Skip this 2D example if libMesh was compiled as 1D-only.
+  libmesh_example_requires(2 <= LIBMESH_DIM, "2D support");
+
+  // Create a mesh, with dimension to be overridden later, on the
+  // default MPI communicator.
+  DistributedMesh mesh(init.comm());
```
roystgnr: Do we want to force DistributedMesh here?

roystgnr:
:+1:, but I'm confused about a couple bits:

Why are you using a nonlinear solver here?  Problem looks linear to me.

You mean "match LAGRANGE_VEC version" loosely, right?  There's no weird isomorphism between the C0 P2 and DG P1 approximations?

examples/vector_fe/vector_fe_ex5/vector_fe_ex5.C
lindsayad: I suppose not

lindsayad:
> Why are you using a nonlinear solver here? Problem looks linear to me.

It is linear, but

- A non-linear configuration is what I'm most used to
- It seems like there's a lot of linear examples in libMesh, and... 

> You mean "match LAGRANGE_VEC version" loosely, right? There's no weird isomorphism between the C0 P2 and DG P1 approximations?

Yes I do mean loosely, as in the approximate solutions both approach the same true solution. Also since the `LAGRANGE_VEC` example uses a linear solver, I thought it might be a nice demo to show solving the same problem not only with a different basis but with a different solver.

examples/vector_fe/vector_fe_ex5/vector_fe_ex5.C
lindsayad: Changed to `Mesh`

PR #2216, opened 2019-08-16
Install metis.h

roystgnr:
This fixes a MOOSE compilation issue (against out of tree libMesh
builds) for me, and would probably be useful for other users too.


jwpeterson:
Wait, we don't want to install metis.h if we are using PETSc's metis.h, as they might be different. Does this just always install metis.h or does it check to see whether we are using PETSc's header somehow?

roystgnr:
If we're using PETSc's header, then BUILD_METIS is false, and metis doesn't get added to SUBDIRS in contrib/Makefile.am, and nothing in contrib/metis/Makefile.am (including this new header installation) happens.

(I'd like to say I thought of this use case and ensured it would work properly, but no, that was just dumb luck.)

PR #2219, opened 2019-08-19
Reduce header dependencies

roystgnr:
This greatly reduces the number of unnecessary inclusions of the huge parallel.h swath of headers, among some more minor improvements.

As usual for this sort of refactoring, this can break codes which use Foo but rely on including "bar.h" to include "foo.h", but the MOOSE fixes in https://github.com/idaholab/moose/pull/13919 weren't bad.


roystgnr:
And apparently I can even break other branches I was working on simultaneously!

roystgnr:
Looks like I'll need to kick this after those MOOSE changes make it from next to master.

PR #2221, opened 2019-08-20
Disabled glibc++ debug defines by default.

hrittich:
By default libMesh enables the flags -D_GLIBCXX_DEBUG and -D_GLIBCXX_DEBUG_PEDANTIC for the debug version of the library. These flags are dangerous.

These flags require that all C++ libraries that are linked together with the debug version of libMesh are also compiled with these flags. When a user combines libMesh with a library that was compiled without these flags, no warning or error message is shown. The resulting application will however segfault in various places, which makes it hard to trace back the origin of the error. (It just took me two hours to do so.)

I would suggest, to disable these flags by default and to show an prominent warning message when using them. I am aware that these flags can be very useful for debugging and also for running unit tests. Since they require special care and proper treatment by users, however, they should only be used by users that know what they are doing.

### DIFF
--- a/configure
+++ b/configure
@@ -1937,9 +1937,9 @@ Optional Features:
   --enable-cxx11-required Error if compiler does not support C++11
   --enable-sanitize="opt dbg devel prof oprof"
                           turn on sanitizer flags for the given methods
-  --disable-glibcxx-debugging
-                          omit -D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_PEDANTIC even
-                          in dbg mode
+  --enable-glibcxx-debugging
+                          enable -D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_PEDANTIC in
+                          dbg mode
   --enable-glibcxx-debugging-cppunit
                           Use GLIBCXX debugging flags for unit tests
   --enable-coverage       configure code coverage analysis tools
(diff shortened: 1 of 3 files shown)


roystgnr:
We need to tack on a bootstrap commit before merging; assuming you've allowed pushes to your branch, I'll do that shortly.

Talking with @jwpeterson over IM, we're both in favor of this change.  An easily fixed "whoops, accidentally mixed std::__debug with normal containers" to someone who's familiar with the incompatibility isn't a problem, but a 2 hour hunt by someone who's not expecting the problem is awful; our default options should be more robust.

permcody:
We (MOOSE) really like these flags! They can help find indexing problems very quickly in places where we don't already have asserts or other protections.

However, I'm fine with the default changing but I'd like to turn these back on MOOSE's "update_and_rebuild_libmesh.sh" script when this goes in.

jwpeterson:
> I am aware that these flags can be very useful for debugging and also for running unit tests. 

Ironically we disable unit testing in dbg mode builds for precisely this reason, issues arise due to the cppunit library on the system not being complied with the pedantic flags.

> However, I'm fine with the default changing

One idea would be to have a new `METHOD`, "edbg" (extra/extreme debugging) that would turn on the pedantic flags. Even if we don't add another `METHOD`, I'm in favor of "dbg" being something rock solid as far as linking against other applications, and simple to understand. devel has always been a bit more nuanced in its meaning, and a brand new user is not likely to know when or why they would need it...

roystgnr:
> One idea would be to have a new METHOD, "edbg" (extra/extreme debugging) that would turn on the pedantic flags.

I'm not against that, but personally I think --enable-glibcxx-debugging is enough.  IMHO most users know they won't ever be linking to non-GLIBCXX_DEVEL libraries or suspect they will be, and the former can use the configure flag and the latter avoid it.  The only benefit to building both dbg and edbg modes would be for users who are simultaneously developing multiple applications, at least one of which passes STL containers to/from other C++ libraries that can't be conveniently recompiled, and at least one other of which avoids doing so but greatly benefits from debug STL container usage. ... and honestly if you're *that* hardcore a developer you're probably not going to have a problem just doing two separate libMesh installs with different configure options.

PR #2222, opened 2019-08-23
Fix missing includes in --disable-optional builds

jwpeterson:
The libmesh_make_unique macro is still required even with fparser disabled.

### DIFF
--- a/include/numerics/parsed_function.h
+++ b/include/numerics/parsed_function.h
@@ -20,6 +20,7 @@
 
 #include "libmesh/libmesh_config.h"
 #include "libmesh/function_base.h"
+#include "libmesh/auto_ptr.h" // libmesh_make_unique
 
 #ifdef LIBMESH_HAVE_FPARSER
 
@@ -27,7 +28,6 @@
 #include "libmesh/dense_vector.h"
 #include "libmesh/vector_value.h"
 #include "libmesh/point.h"
-#include "libmesh/auto_ptr.h" // libmesh_make_unique
 
 // FParser includes
 #include "libmesh/fparser_ad.hh"
--- a/include/reduced_basis/rb_assembly_expansion.h
+++ b/include/reduced_basis/rb_assembly_expansion.h
@@ -25,7 +25,7 @@
 
 // C++ includes
 #include <vector>
-
+#include <memory>
 
 namespace libMesh
 {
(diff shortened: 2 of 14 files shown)


jwpeterson:
This seems like it should have been caught by "Test No Optional" previously, but maybe that doesn't really disable fparser for some reason...

jwpeterson:
Oh, most likely it was uncovered by shuffling around of headers in #2219.

roystgnr:
Wow, thanks for catching these.  I should have been smart enough to try Test No Optional on #2219.

roystgnr:
But wait, what's going on with all those LIBMESH_HAVE_PETSC wrappers?  I could swear I've run through the unit tests with Eigen sparse solvers recently.

jwpeterson:
> But wait, what's going on with all those LIBMESH_HAVE_PETSC wrappers? I could swear I've run through the unit tests with Eigen sparse solvers recently.

Yes, technically you can run those unit tests as long as *any* linear solver is available. So the logic should be
```
 #if defined(LIBMESH_HAVE_PETSC) || defined(LIBMESH_HAVE_EIGEN) || defined...
```
but I was "compiling by CIVET" last night so I just went for the quick and dirty solution of checking for one particular linear solver package...

roystgnr:
Ah, gotcha.  Let's wrap that logic into a single LIBMESH_HAVE_SOLVER, though, so we can keep the extra test coverage with non-PETSc solvers?  I know we'll only use it once in a blue moon, but it was pretty useful for me just recently.

jwpeterson:
> Let's wrap that logic into a single LIBMESH_HAVE_SOLVER,

This sounds like it would be useful. It feels like it would naturally fit into libmesh_config.h, but we'd need configure to somehow write it for us. Maybe in libmesh_common.h?

roystgnr:
Oh, I hate m4 too much to want to do this in configure.  I was just thinking
```
#if defined(LIBMESH_HAVE_PETSC) || defined(LIBMESH_HAVE_EIGEN) || defined...
#define LIBMESH_HAVE_SOLVER 1
#endif
```

I'd been picturing it in linear_solver.h, but I guess some of the nonlinear solver headers encapsulate that away?  libmesh_common.h seems too extreme.  system.h, since we're generally wondering whether we can expect anything out of system.solve()?

jwpeterson:
> since we're generally wondering whether we can expect anything out of system.solve()?

In this case, the error comes from the constructor of `LinearImplicitSystem`, when it tries to call `LinearSolver<Number>::build()`. So, yeah, system.h would most likely be included anywhere that we'd reasonably want to use the macro.

PR #2233, opened 2019-08-29
ExodusII: Read and write sideset data

jwpeterson:
Variables (data) on sidesets are conceptually similar to elemental data fields defined on lower-dimensional subdomains of elements.  They can be useful for defining boundary condition data, such as e.g. piecewise-constant normal forces on each face of a sideset.

### DIFF
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -695,8 +695,8 @@ public:
    * (elem-id, side-id, bc-id) triplets and returns it to the user,
    * taking advantage of guaranteed RVO.
    */
-  std::vector<std::tuple<dof_id_type, unsigned short int, boundary_id_type>>
-  build_side_list() const;
+  typedef std::tuple<dof_id_type, unsigned short int, boundary_id_type> BCTuple;
+  std::vector<BCTuple> build_side_list() const;
 
   /**
    * Creates a list of active element numbers, sides, and ids for those sides.
--- a/include/mesh/exodusII_io.h
+++ b/include/mesh/exodusII_io.h
@@ -266,6 +266,31 @@ public:
                        const std::set<std::string> * system_names=nullptr);
 
   /**
+   * FIXME: This tuple should be defined in the BoundaryInfo class/header file
+   * since it is convenient to have wherever BoundaryInfo functions are called...
+   */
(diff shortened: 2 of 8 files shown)


roystgnr:
> e.g. piecewise-constant normal forces on each face of a sideset.

How does that work?  If you have a piecewise constant defined on your elements then you're in trouble if one of those elements has two different boundary sides with different normal forces.

jwpeterson:
> one of those elements has two different boundary sides with different normal forces.

Maybe I don't understand what you are asking, but everything is indexed by (elem, side, boundary_id, variable_name) tuples, so at the moment you can have one value per side per boundary id per variable in each Elem.

roystgnr:
Oh, no, I'm just being an idiot.  I assumed from the description that you were pulling values from solution fields somehow; didn't notice that they were coming in via the bc_vals argument.

jwpeterson:
Right... at the moment, the "data" is outside of the EquationSystems/System hierarchy because it's not something we usually store, so I just made up a container that I thought was convenient...

roystgnr:
I'm now seeing
```
Error writing sidesets
Exodus Library Error: [ex_put_sets]
	Error: failed to store side set id 1 in file id 65536
	[-40] Index exceeds dimension bound
```
from adaptivity_ex3 after rebasing a branch on master today.  Could be something I've broken (I'm testing adaptivity_ex3 in particular because that triggered a different failure with an earlier version of this branch), but that message sure *looks* like it might be related to this PR.

jwpeterson:
Hmm... that code (`ExodusII_IO_Helper::write_sidesets`) hasn't changed recently as far as I know. And you likely aren't even calling the new code that did change (`ExodusII_IO_Helper::write_sideset_data`). 

Is the file id really 65536? That seems suspicious, like maybe the file has not been opened successfully for writing for some reason. Is there a way to reproduced this easily? Just run adaptivity_ex3?

roystgnr:
Looks like this is probably a red herring related to my own changes; I tried to reproduce it in master and couldn't.  Sorry for the false alarm.

roystgnr:
Yeah, definitely a red herring.  Found the problem, it was a mistake with my new MPI configuration, and it had nothing to do with any libMesh bug, much less with this PR.

Issue #2179, opened 2019-07-19
libMesh IsoGeometric Analysis support


roystgnr:
I'm working with the [Coreform](https://coreform.com/) folks on adding basic IGA support to libMesh (and MOOSE), via their tech for converting isogeometric U-spline meshes into Bezier patches that we can handle more easily.

The basic plan on the libMesh side is, in chronological order:

- [ ] Add a FE specialization/subclass for calculating at least low-order (quadratic and easier) rational Bezier basis functions
- [ ] Add Node-accessible storage for Bezier weight and/or projective space coordinate storage, to put the "rational" into "rational Bezier"
- [ ] Add library support for choosing between bases for master->physical space mappings
- [ ] Add a MeshInput option for their LS-DYNA format, to read Bezier patch meshes
- [ ] Add benchmark cases to test convergence on e.g. a perfect circle or sphere and ensure we really aren't seeing any geometry discretization error.

(and then at that point remaining work is on the MOOSE side)

Biggest subissues:
* How to best store nodal coefficients.  I can see three ways of doing this; I'll probably do the most easy hack (configure-time option) in the short term while doing the best thing (arbitrary *non*-integer POD DofObject additions, one step from #2095 closer toward #1439) in the medium term.
* How to properly handle C1 continuity.  We'd really like to have this both for certain physics and for some really useful benchmark problems, but I'm only about 90% sure that the "element extraction operator" in the IGA->Bezier papers can be implemented as simply as new constraint equations in libMesh.
* The Elem::loose_bounding_box() heuristics are probably going to be garbage when we throw IGA mappings into the mix.  For the work I'm contracted to do our solution is "don't run any simulations that rely on MeshBase::point_locator()", but that offends me and might cause problems for more sophisticated use cases for others and I'm hoping to come up with a better fix.

roystgnr:
Forgot to tag @cgoates on this.

roystgnr:
Okay, there seems to be only one clear option.  We always have our hands on an Elem when we're evaluating rational basis functions, and multiple IGA meshes would have distinct Elems, so that's the obvious place to put a weight index if we can fit one in without overhead for non-IGA users.

[...]

PR #2234, opened 2019-08-30
FEMap refactor

roystgnr:
```
// TODO: PB: We should consider moving this to the FEMap class
```

// TADA: All moved to the FEMap class.

Clearing out those years-old TODOs should make a few things infinitesimally more efficient, but more urgently it's a clean up that should make implementing the next tick box in #2179 cleaner too.

Begging @BalticPinguin to try this on his own infinite element codes; doing this cleanly required me to do an analogous refactoring on the InfFE class at the same time, but I'm much less confident about that part of it, and we don't have as much test coverage there as we should.

### DIFF
--- a/src/fe/inf_fe_map.C
+++ b/src/fe/inf_fe_map.C
@@ -20,6 +20,7 @@
 // Local includes
 #include "libmesh/libmesh_config.h"
 #ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+#include "libmesh/inf_fe_map.h"
 #include "libmesh/inf_fe.h"
 #include "libmesh/fe.h"
 #include "libmesh/elem.h"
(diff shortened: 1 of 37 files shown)


BalticPinguin:
Sorry for the delay; but I finally tried this in my code; didn't experience any problem so far.
So I think, there should be no problem with this...

roystgnr:
Thanks!  I'll bug you when I put up the next stage of the process.  And a week isn't much delay for something like this; it's easy to revert a mistake after a week.  Mostly I just want to make sure that I don't write some bug that hits you way down the line when there are patches depending patches depending on the bug.

src/fe/inf_fe_map.C
```diff
@@ -36,34 +37,32 @@ namespace libMesh
...
     {
     case 1:
       base_point = inf_elem->point(0);
       break;
     case 2:
-      base_point = FE<1,LAGRANGE>::map (base_elem.get(), reference_point);
-      break;
     case 3:
-      base_point = FE<2,LAGRANGE>::map (base_elem.get(), reference_point);
+      base_point = FEMap::map (dim, base_elem.get(), reference_point);
```
BalticPinguin: Oh; I found an issue: it must be ``dim-1`` because we are trying to map the base_elem which is the face towards the finite element region.

BalticPinguin: ping @roystgnr ; I am not sure whether notifications are sent for comments in code that is merged already...

roystgnr: Thanks!  I'll fix it ASAP.  Any idea why our "Test with complex numbers and infinite elements" CI didn't pick it up?

BalticPinguin: no, not really, I also came across this only by accident, even though I use this function at many places in my code...

Issue #2236, opened 2019-09-03
Create v1.5.0 release


jwpeterson:
@roystgnr will be submitting some larger changesets soon; he'd like to have a new release before that happens. Tag this issue number in any (mainly bugfix) commits that should be cherry-picked over from master before the final release is tagged.

jwpeterson:
https://github.com/libMesh/libmesh/releases/tag/v1.5.0

PR #2237, opened 2019-09-04
Add new warning to ignore in 3rd party code

roystgnr:
On Ubuntu 19.04 the use of g++ 8.3 and OpenMPI 3.1.3 together trigger
this warning for me.

Judging by https://github.com/open-mpi/ompi/issues/5157 there isn't an
underlying bug here, the overzealous warning isn't going to be fixable
upstream, and even if it was there'd still be users on the existing
releases, so let's just make sure we ignore it.


roystgnr:
This should probably be added to #2236; it's only a warning and it's only on that particular configuration, but it's a warning that triggers on literally every file, and it's a configuration that's one of the most popular Linux distros.

PR #2239, opened 2019-09-05
Support non-Lagrange mapping bases

roystgnr:
This is the next checkbox on the #2179 list: the ability to specify and then use different function spaces for the master to physical element maps.

This shouldn't be merged until I've actually got some rational basis function unit tests to *cover* the new feature, but it's at least at the point where we can throw CI at it and make sure I haven't regressed anything else in the process.


roystgnr:
Still want to improve test coverage on this, but it does appear to be working.

roystgnr:
After talking to @cgoates et al, it sounds like the best thing to do w.r.t. expanded unit testing is going to be going straight to manufactured solution tests, which is more heavyweight than I want to add to this one PR, so we'll merge this as-is and bring in the extra testing alongside the new MeshInput subclass.

Issue #2240, opened 2019-09-06
Communicator::sum(T&) not defined


BalticPinguin:
Hi,
I am currently trying to run the newest version; but the checks failed for me:
```
from ../../../../master/examples/subdomains/subdomains_ex3/subdomains_ex3.C:32:
../../../include/libmesh/communicator.h:328:8: warning: inline function ‘void libMesh::Parallel::Communicator::sum(T&) const [with T = double]’ used but never defined
   void sum(T & r) const;
```
I think, this might be due to the new 'inline'-character from commit 7b58040354 (sorry; I didn't find the PR-nr now) ? At least, [this discussion](
https://stackoverflow.com/questions/9338152/must-the-definition-of-a-c-inline-functions-be-in-the-same-file) supports this. However, this might have shown up in some other checks already?

jwpeterson:
Hi, 

You should be able to fix the issue by adding
```
#include "libmesh/parallel.h"
```
to that example? I'm not really sure why our `--disable-strict-lgpl` CI testing missed this... it seems to work fine for the example in question. Let us know if including that header fixes it and we'll apply the fix in upstream.

roystgnr:
Sorry, this was my fault.  It should have been fixed yesterday, though, when we merged #2238 - is it still failing for you after that?

BalticPinguin:
thanks; you were right: now the issue is resolved.

PR #2241, opened 2019-09-07
Hiding second derivatives in macro 2

BalticPinguin:
Hi, 
by updating, in addition to issue #2240, I had some compile-problems of the form
```
./.libs/libmesh_dbg.so: undefined reference to `libMesh::FE<3u, ...>:: shape_second_deriv( ElemType ,FEOrder<,3u ,unsigned int22,>
shape_second_deriv:(Point, const &ElemType),'
./.libs/libmesh_dbg.so: undefined reference to `libMesh::FE<3u, (libMesh.::/FEFamily.)libs22/>libmesh_dbg.so::: shape_second_derivundefined( libMeshreference: :toElemType ,` libMeshlibMesh::::FEOrder<,3u ,unsigned  (intlibMesh,: :unsignedFEFamily )int22,> :libMesh::shape_second_deriv:(PointlibMesh :const:&ElemType),'
libMesh.:/:.Orderlibs,/ libmesh_dbg.sounsigned:  intundefined,  referenceunsigned  toint ,` libMeshlibMesh::::FEPoint< 2uconst,& )('libMesh
:.:/FEFamily.)libs2/>libmesh_dbg.so::: shape_second_derivundefined( libMeshreference: :toElem  `constlibMesh*:,: FElibMesh<:2u:,Order ,( libMeshunsigned: :intFEFamily,) 2unsigned> :int:,shape_second_deriv (libMeshlibMesh::::PointElem  constconst&*),'
```
(using ``--disable-second``). For some reason, they showed up only now, and I must have forgotten them in #1983.

I hope, not to introduce some problems for some other configs by this...

### DIFF
--- a/include/fe/fe.h
+++ b/include/fe/fe.h
@@ -157,6 +157,7 @@ public:
                                  const unsigned int j,
                                  const Point & p);
 
+#ifdef LIBMESH_ENABLE_SECOND_DERIVATIVES
   /**
    * \returns The second \f$ j^{th} \f$ derivative of the \f$ i^{th} \f$
    * shape function at the point \p p.
@@ -209,6 +210,7 @@ public:
                                         const unsigned int j,
                                         const Point & p);
 
+#endif //LIBMESH_ENABLE_SECOND_DERIVATIVES
   /**
    * Build the nodal soln from the element soln.
    * This is the solution that will be plotted.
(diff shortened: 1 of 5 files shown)


roystgnr:
Looks good; let me play with it a bit before merging though.  Those are static functions that shouldn't be called in the --disable-second case, so I'm having trouble understanding why the linker is seeing any reference to them at all.  I guess that's a system-dependent question?

roystgnr:
Well, I can't understand it still, but I can replicate it (with gcc8; I wonder if older compilers were more forgiving?) and I can verify that this fixes it.  Thanks!

PR #2244, opened 2019-09-10
Add DiagonalMatrix

### LINKED ISSUE #2168: Diagonal PETSc matrix
We've recently been doing some automatic scaling of our residuals and Jacobians in MOOSE (see idaholab/moose#12601 and idaholab/moose#13547) in order to improve the condition  number of the linear system, and we've seen some pretty awesome improvements. Take this Jacobian-free non-linear solve without scaling:
```
 0 Nonlinear |R| = 1.326650e+04
      0 Linear |R| = 1.326650e+04
      1 Linear |R| = 2.944220e-05
 1 Nonlinear |R| = 1.864784e+00
      0 Linear |R| = 1.864784e+00
      1 Linear |R| = 3.385636e-06
 2 Nonlinear |R| = 6.259735e-04
      0 Linear |R| = 6.259735e-04
      1 Linear |R| = 9.361485e-07
      2 Linear |R| = 1.749268e-08
      3 Linear |R| = 9.788406e-10
 3 Nonlinear |R| = 1.039767e-04
      0 Linear |R| = 1.039767e-04
      1 Linear |R| = 1.515572e-10
 4 Nonlinear |R| = 3.621991e-07
      0 Linear |R| = 3.621991e-07
      1 Linear |R| = 3.113164e-11
      2 Linear |R| = 1.267239e-12
 5 Nonlinear |R| = 1.073857e-08
```
vs with scaling:
```
 0 Nonlinear |R| = 1.326650e+04
      0 Linear |R| = 1.326650e+04
      1 Linear |R| = 7.073560e+01
      2 Linear |R| = 2.495484e+00
      3 Linear |R| = 8.710280e-02
 1 Nonlinear |R| = 1.866817e+00
      0 Linear |R| = 1.866817e+00
      1 Linear |R| = 2.314249e-02
      2 Linear |R| = 8.128605e-04
      3 Linear |R| = 4.823874e-05
      4 Linear |R| = 2.617779e-06
 2 Nonlinear |R| = 7.210617e-05
      0 Linear |R| = 7.210617e-05
      1 Linear |R| = 7.244895e-07
      2 Linear |R| = 4.770732e-08
      3 Linear |R| = 2.199447e-09
      4 Linear |R| = 1.075420e-10
 3 Nonlinear |R| = 1.609251e-09
```
However, with the way we're implementing this scaling, we end up doing mallocs in `MatSetValues` if we're using discontinuous Galerkin calculations for example because the non-zero structure is different between the matrix computed for scaling purposes and the true matrix. In truth we only care about the diagonal for scaling (eventually we want other scaling choices for say saddle-point problems). So I'm proposing a `PetscDiagonalMatrix` that inherits from `SparseMatrix` but actually uses a `Vec` to store the data. Calls to things like `PetscDiagonalMatrix::add(i, j, value)` would simply return if `i` doesn't equal `j`. Implementing this new class is definitely the simplest option for accomplishing what we want, and in general I think that simplest is best :smile:

lindsayad:
`DiagonalMatrix` holds its data in an underlying `NumericVector` data member, and shims all its methods down into it.

Also:
- add a templated `add_matrix` method to `ImplicitSystem` that allows a user to specify the type of matrix they want to add, since I may want to simultaneously have a `PetscMatrix` and `DiagonalMatrix` held by the system (and because `DiagonalMatrix` itself is not associated with a solver package (although its underlying `NumericVector` is))

Closes #2168

### DIFF
--- a/include/numerics/dense_matrix_base.h
+++ b/include/numerics/dense_matrix_base.h
@@ -31,6 +31,7 @@ namespace libMesh
 
 // Forward declarations
 template <typename T> class DenseVectorBase;
+template <typename T> class DenseVector;
 
 /**
  * Defines an abstract dense matrix base class for use in Finite Element-type
@@ -138,6 +139,11 @@ public:
     ScalarTraits<T2>::value, void >::type
   add (const T2 factor,
        const DenseMatrixBase<T3> & mat);
+
+  /**
+   * Return the matrix diagonal
+   */
+  DenseVector<T> diagonal() const;
 
 protected:
 
--- a/include/numerics/dense_matrix_base_impl.h
+++ b/include/numerics/dense_matrix_base_impl.h
@@ -23,6 +23,7 @@
 // Local Includes
 #include "libmesh/dense_matrix.h"
 #include "libmesh/dense_vector_base.h"
+#include "libmesh/dense_vector.h"
 
 // C++ includes
 #include <iomanip> // for std::setw()
@@ -30,7 +31,15 @@
 namespace libMesh
 {
 
-
+  template <typename T>
+  DenseVector<T>
+  DenseMatrixBase<T>::diagonal() const
+  {
+    DenseVector<T> ret;
+    for (decltype(_m) i = 0; i < _m; ++i)
+      ret(i) = el(i, i);
+    return ret;
+  }
 
   template<typename T>
   void DenseMatrixBase<T>::multiply (DenseMatrixBase<T> & M1,
--- a/include/numerics/dense_matrix_base_impl.h
+++ b/include/numerics/dense_matrix_base_impl.h
@@ -35,7 +35,7 @@ namespace libMesh
   DenseVector<T>
   DenseMatrixBase<T>::diagonal() const
   {
-    DenseVector<T> ret;
+    DenseVector<T> ret(_m);
     for (decltype(_m) i = 0; i < _m; ++i)
       ret(i) = el(i, i);
     return ret;
--- a/include/numerics/numeric_vector.h
+++ b/include/numerics/numeric_vector.h
@@ -133,6 +133,12 @@ public:
   NumericVector & operator= (NumericVector &&) = default;
 
   /**
+   * While this class doesn't manage any memory, the derived class might and
+   * users may be deleting through a pointer to this base class
+   */
+  virtual ~NumericVector() = default;
+
+  /**
    * Builds a \p NumericVector on the processors in communicator
    * \p comm using the linear solver package specified by
    * \p solver_package
--- a/tests/numerics/diagonal_matrix_test.C
+++ b/tests/numerics/diagonal_matrix_test.C
@@ -40,7 +40,7 @@ public:
   void setUp()
   {
     _comm = TestCommWorld;
-    _matrix = libmesh_make_unique<DiagonalMatrix<Real>>(*_comm);
+    _matrix = libmesh_make_unique<DiagonalMatrix<Number>>(*_comm);
 
     numeric_index_type root_block_size = 2;
     _local_size = root_block_size + static_cast<numeric_index_type>(_comm->rank());
@@ -82,8 +82,8 @@ public:
     _matrix->set(beginning_index + 1, beginning_index, 1);
 
     // Test that off-diagonal elements are always zero
-    LIBMESH_ASSERT_FP_EQUAL((*_matrix)(beginning_index, beginning_index + 1), 0, _tolerance);
-    LIBMESH_ASSERT_FP_EQUAL((*_matrix)(beginning_index + 1, beginning_index), 0, _tolerance);
+    LIBMESH_ASSERT_FP_EQUAL(libmesh_real((*_matrix)(beginning_index, beginning_index + 1)), 0, _tolerance);
+    LIBMESH_ASSERT_FP_EQUAL(libmesh_real((*_matrix)(beginning_index + 1, beginning_index)), 0, _tolerance);
 
     // Test add
     {
@@ -101,9 +101,9 @@ public:
       for (numeric_index_type i = beginning_index; i < end_index; ++i)
       {
         if (i)
-          LIBMESH_ASSERT_FP_EQUAL((*_matrix)(i, i), i, _tolerance);
+          LIBMESH_ASSERT_FP_EQUAL(libmesh_real((*_matrix)(i, i)), i, _tolerance);
         else
-          LIBMESH_ASSERT_FP_EQUAL((*_matrix)(i, i), _comm->size(), _tolerance);
+          LIBMESH_ASSERT_FP_EQUAL(libmesh_real((*_matrix)(i, i)), _comm->size(), _tolerance);
       }
     }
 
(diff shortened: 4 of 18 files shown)


lindsayad:
I can add a unit test for this in a bit

include/numerics/dense_matrix_base.h
```diff
@@ -139,6 +140,11 @@ class DenseMatrixBase
   add (const T2 factor,
        const DenseMatrixBase<T3> & mat);
 
+  /**
+   * Return the matrix diagonal
+   */
+  DenseVector<T> diagonal() const;
```
jwpeterson: If I recall correctly we can't get away with a forward declare here, since `DenseVector<T>` is returned by value, the compiler needs to know the complete type. We may not actually see any failures unless dense_matrix_base.h is included all by itself somewhere...

lindsayad: Based on [this SO post](https://stackoverflow.com/a/553869/4493669) I think we should be ok here

jwpeterson: Huh, well the whole situation is a bit more complicated than I remembered I guess, especially considering this line

    When it comes to templates, there is no absolute rule: whether you can use an incomplete type as a template parameter is dependent on the way the type is used in the template.

I guess let's just keep it as-is until some compiler tells us otherwise...

roystgnr:
Looks good once a unit test is in.

fdkong:
Looks really nice. In the future, I would like to have `DiagonalMatrix<T>::inverse()` that will be helpful for explicit schemes in MOOSE

include/numerics/dense_matrix_base_impl.h
```diff
@@ -23,14 +23,23 @@
...
 
-
+  template <typename T>
+  DenseVector<T>
+  DenseMatrixBase<T>::diagonal() const
+  {
+    DenseVector<T> ret;
+    for (decltype(_m) i = 0; i < _m; ++i)
+      ret(i) = el(i, i);
+    return std::move(ret);
```
jwpeterson: There's an Effective C++11 item about returning by `std::move`. Basically, you aren't supposed to do it, because it can inhibit RVO. Instead just `return ret;`.

jwpeterson: https://stackoverflow.com/questions/4986673/c11-rvalues-and-move-semantics-confusion-return-statement

permcody: Haha - We've had a number of debates on the "decltype" question. It boils down to whether you trust contributors to use the right index type. In my experience the answer is no. Depending on the background and experience level of your contributor, you'll get either signed or unsigned variants of ints. That might result in a sign mismatch warning if you turn on enough compiler warning or worse you might actually overflow the int type if you are working with really big ranges.

In MOOSE we created a "mooseindex(type)" call that will always give you the right indexing type back for the type you pass in. It works for containers as well so you don't have to do even ugly tricks to get a non-const POD type back in those cases. Still many in our community (and our team) still dislike it...

tests/numerics/diagonal_matrix_test.C
```diff
@@ -0,0 +1,222 @@
...
+
+namespace
+{
+template <typename T>
+T
+termial(T n)
+{
+  if (n == 1)
+    return 1;
+  return n + termial(n - 1);
```
roystgnr: Is this just a really expensive way to calculate ```n*(n+1)/2```?

tests/numerics/diagonal_matrix_test.C
```diff
@@ -0,0 +1,222 @@
...
+}
+}
+
+class DiagonalMatrixTest : public CppUnit::TestCase
+{
+public:
+  void setUp()
+  {
+    _comm = TestCommWorld;
+    _matrix = libmesh_make_unique<DiagonalMatrix<Real>>(*_comm);
```
roystgnr: Maybe try clearing this in the tearDown?  Perhaps CppUnit isn't destroying the DiagonalMatrixTest object until after the LibMeshInit destruction, for some reason, so then the tests pass but the destructor throws some kind of error?

lindsayad:
Ah the issue is that the test is passing on process 0 (and that's the only output that is printed on the screen) but not on other processes...

tests/numerics/diagonal_matrix_test.C
lindsayad: Lol. Will change

include/numerics/numeric_vector.h
```diff
@@ -132,6 +132,12 @@ class NumericVector : public ReferenceCountedObject<NumericVector<T>>,
   NumericVector (const NumericVector &) = default;
   NumericVector & operator= (NumericVector &&) = default;
 
+  /**
+   * While this class doesn't manage any memory, the derived class might and
+   * users may be deleting through a pointer to this base class
+   */
+  virtual ~NumericVector() {}
```
lindsayad: @roystgnr @jwpeterson Is there anything I'm missing here? Are you guys ok with this addition?

tests/numerics/diagonal_matrix_test.C
jwpeterson: Thanks, I just learned a [new math word/notation](https://en.wikipedia.org/wiki/Termial)!

include/numerics/numeric_vector.h
jwpeterson: I'm trying to use the `= default;` notation for empty special functions. I'm wondering how we never had a virtual destructor from NumericVector before now??? Seems like a good one to backport.

lindsayad: switched to `default`

roystgnr: I'm actually seriously confused here.  Don't we almost always delete NumericVector subclass objects via a pointer to base?  Don't all our subclasses do heap memory deallocation in their destructors?

Oh, I get how it works now.  ParallelObject has a virtual destructor, so all of its derived classes end up having a virtual destructor, whether they're explicitly declared to or not.  I'm a fan of either this or the "= default" solution to make that more bulletproof, though.  Looks like we originally did have an explicit virtual NumericVector destructor, too, at least in 0.5.0 (first ancient version I had handy).

lindsayad: Ah...that makes sense. Yea I was definitely perplexed as well

jwpeterson: Oof, looks like I just randomly deleted it in the process of updating the special functions in f71e04fd58d0618bd5903b239f66d121e3f57b0b

lindsayad:
Why can't cppunit properly destroy tests when selecting a subset with `--re`. It's kind of obnoxious always getting "Memory leak detected!"

jwpeterson:
> Why can't cppunit properly destroy tests when selecting a subset with `--re`? It's kind of obnoxious always getting "Memory leak detected!"

Oh, so --re is what causes that? I had been seeing it recently but couldn't figure it out. It's possible we are doing something wrong in driver.C. I kind of just hacked that together to support --re but didn't carefully check whether I had also introduced memory leaks...

lindsayad:
From a little time in the debugger it seems as if when running with `--re` ***all*** test classes get instantiated and constructed, but it seems as if only the tests selected with `--re` are destructed.

lindsayad:
Alright https://github.com/libMesh/libmesh/pull/2244/commits/99791bed963fa48a7aca7bf0c6293a0a6b2f6e95 should stop the memory leak reports. Note that I would still expect this to be valgrind un-clean because `registry->makeTest` returns a pointer to a heap allocated `TestSuite`. However we can't actually delete this object because it would attempt to free tests which stand to be freed when the `TestRunner` is destructed.

But at least we won't be leaking libMesh objects any more!

jwpeterson:
> Alright [99791be](https://github.com/libMesh/libmesh/commit/99791bed963fa48a7aca7bf0c6293a0a6b2f6e95) should stop the memory leak reports.

Clever solution making the "rejects" object. Thanks for fixing that!

roystgnr:
Thanks!

PR #2252, opened 2019-09-17
Odd config fixes

roystgnr:
This branch was *originally* supposed to be for finding and fixing any 1.5.0 errors in little-used libMesh configurations, but it turned out that I caught a couple recent (post-1.5.0 branch) outright errors in an actually-used configuration too.  If there are any disagreements about something in the bulk of these commits then I can cherry-pick out the handful that are actually important and submit those separately.

Unlike the errors, most of the warnings fixes apply to 1.5.0 code, but IMHO none of them are important enough to bother backporting; any users who try ```--enable-werror --enable-paranoid-warnings --disable-something-everyone-uses``` can pick and settle for two out of three.

### DIFF
--- a/examples/fem_system/fem_system_ex3/elasticity_system.C
+++ b/examples/fem_system/fem_system_ex3/elasticity_system.C
@@ -27,6 +27,9 @@
 
 using namespace libMesh;
 
+
+#if LIBMESH_DIM > 2
+
 void ElasticitySystem::init_data()
 {
   _u_var = this->add_variable ("Ux", FIRST, LAGRANGE);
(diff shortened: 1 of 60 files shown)


jwpeterson:
> but IMHO none of them are important enough to bother backporting

You didn't happen to tag the 1.5.0 release issue (#2236) in any of the commits you want backported, did you?

roystgnr:
I didn't want anything backported, because I figured most of the fixes are for unimportant warnings and the rest are for errors post-branch, but that was a mistake, because I forgot that you got that SerialMesh-incompatible unit test in pre-branch.  So *now* I think there isn't anything *else* that ought to be backported other than that one unit test fix, but I'm not feeling nearly as confident.

PR #2254, opened 2019-09-18
Add PETSc snes_divergence_tolerance option to libmesh

### LINKED ISSUE #2253: Add support for PETSc non-linear divergence tolerance
An option for a non-linear divergence tolerance `-snes_divergence_tolerance` was added to PETSc 3.8.0.  I would like to add this option to MOOSE [(see Moose Issue #13991](https://github.com/idaholab/moose/issues/13991)) but need to make an addition to `petsc_nonlinear_solver.C` in order to set the default tolerance.

More details about this PETSc option can be found here:
     https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/SNES/SNESSetFromOptions.html
     https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/SNES/SNESSetDivergenceTolerance.html

frombs:
Closes #2253

### DIFF
--- a/include/solvers/nonlinear_solver.h
+++ b/include/solvers/nonlinear_solver.h
@@ -312,6 +312,15 @@ public:
   double relative_residual_tolerance;
 
   /**
+   * The NonlinearSolver should exit if the residual becomes greater
+   * than the initial residual times the divergence_tolerance.
+   *
+   * Users should adjust this tolerances to prevent divergence of the
+   * NonlinearSolver.
+   */
+  double divergence_tolerance;
+
+  /**
    * The NonlinearSolver should exit after the full nonlinear step norm is
    * reduced to either less than absolute_step_tolerance
    * or less than relative_step_tolerance times the largest
@@ -407,6 +416,7 @@ NonlinearSolver<T>::NonlinearSolver (sys_type & s) :
   max_function_evaluations(0),
   absolute_residual_tolerance(0),
   relative_residual_tolerance(0),
+  divergence_tolerance(0),
   absolute_step_tolerance(0),
   relative_step_tolerance(0),
   max_linear_iterations(0),
--- a/src/solvers/petsc_nonlinear_solver.C
+++ b/src/solvers/petsc_nonlinear_solver.C
@@ -857,6 +857,12 @@ PetscNonlinearSolver<T>::solve (SparseMatrix<T> &  pre_in,  // System Preconditi
                            this->relative_step_tolerance, this->max_nonlinear_iterations, this->max_function_evaluations);
   LIBMESH_CHKERR(ierr);
 
+  // Set the divergence tolerance for the non-linear solver
+#if !PETSC_VERSION_LESS_THAN(3,8,0)
+  ierr = SNESSetDivergenceTolerance(_snes, this->divergence_tolerance);
+  LIBMESH_CHKERR(ierr);
+#endif
+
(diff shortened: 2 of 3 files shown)


jwpeterson:
Would it make more sense to add a class member `divergence_tolerance`, similarly to what is done for `absolute_residual_tolerance` and `relative_residual_tolerance`, and then set PETSc's divergence tolerance to *that*? It's not clear to me what is gained by setting the value to `PETSC_DEFAULT`, as it seems like that would already be... the default. 

Furthermore, I would have expected that the command line option, `-snes_divergence_tolerance`, should always be picked up regardless, since we call `SNESSetFromOptions(_snes);` Is that not the case? Maybe @fdkong could also comment on this issue.

frombs:
@jwpeterson, thanks for the comments. I based my edits on [lines 851-852](https://github.com/libMesh/libmesh/blob/6f29281056d40b200d2db445b55c9392aafb6dc6/src/solvers/petsc_nonlinear_solver.C#L851-L852) that used `PETSC_DEFAULT` to set some of the linear tolerances so I thought that would be appropriate.  I have no preference for how to set the tolerance and can change the code to a class member if you prefer.

I see your point regarding the command line option but the same case could be made for the PETSc linear tolerances currently set in libmesh.  I added the edits for completeness but if you would rather not include them that is fine.  Let me know how you want to proceed.

fdkong:
@jwpeterson is right. We need to create a member variable and allow it to be changed through the `parameters` class. At the moose side, we should create an input parameter in `FEProblemSove.C` so that end-users are able to change this parameter in a moose input file. Of course, the petsc native option is still a valid path. 

BTW, `PETSC_DEFAULT` is not `10^4` that is the default value of the divergence tolerance.

frombs:
@fdkong, thanks for the clarification. I misunderstood what `PETSC_DEFAULT` does. I will work on the changes.

frombs:
Can the tests be reset?  Several of them failed with exit code 2.  I think they may have failed because I force pushed a correction while they were running. Thanks.

jwpeterson:
I can re-start the tests, but note that there does seem to have been a real error which caused the failure:
```
ERROR: no double parameter named "nonlinear solver divergence tolerance" found.

Known parameters:
Name	 Type	 Value
---------------------
 linear solver maximum iterations	 unsigned int	 10000
 linear solver minimum tolerance	 double	 1e-05
 linear solver tolerance	 double	 1e-05
 nonlinear solver absolute residual tolerance	 double	 1e-35
 nonlinear solver absolute step tolerance	 double	 1e-08
 nonlinear solver maximum function evaluations	 unsigned int	 10000
 nonlinear solver maximum iterations	 unsigned int	 50
 nonlinear solver relative residual tolerance	 double	 1e-08
 nonlinear solver relative step tolerance	 double	 1e-08
 nonlinear solver tolerance	 double	 1e-12
```

jwpeterson:
Yeah, it failed with the same error in `misc_ex3` which is referring to the new parameter. Let me know if you are not able to see the test output on Civet.

frombs:
@jwpeterson, thanks for your help.  I was looking for the error message to be colored like Civet does on Moose so I failed to see the error for `misc_ex3`.  I just pushed an update that should resolve it.

fdkong:
@frombs  Could you please add a test in MOOSE?

PR #2256, opened 2019-09-23
Add MAP_FIND macro

jwpeterson:
This actually accomplishes what I was trying to do in #2255. There's no way to get the right file and line number without using a macro at some level. An example of the generated error message is now:
```
map_find() error: key not found in file ../src/mesh/exodusII_io_helper.C on line 2219
```

### DIFF
--- a/include/mesh/vtk_io.h
+++ b/include/mesh/vtk_io.h
@@ -195,13 +195,13 @@ private:
     // Find an entry in the writing map, or throw an error.
     vtkIdType find(ElemType libmesh_type)
     {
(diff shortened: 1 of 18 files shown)


roystgnr:
:-P  Hate to make you put in a third PR, but could we change this to libmesh_map_find?  MAP_FIND seems generic enough that I'm not going to be surprised to see conflict with some user or third party header code out there somewhere...

jwpeterson:
> could we change this to libmesh_map_find?

Sure, no problem. After merging, I realized I probably should have done that...  :expressionless:

PR #2259, opened 2019-09-26
DynaIO MeshInput subclass

roystgnr:
Part of the next phase of #2179; this reads geometry (and has stubs for later reading of constraint equations) from the BEXT format Bezier meshes that CoreForm produces.

I slipped in a few fixes for unrelated compiler warnings that cropped up while I was testing this branch with a --disable-mpi build; IMHO nothing worth backporting.


roystgnr:
Damn, I only tested with DistributedMesh, and it only works in parallel with DistributedMesh.

I think that's the first time *that's* ever happened to me, rather than the other way around.

roystgnr:
This shouldn't mess with anybody who isn't using these features (except that the extra integer metadata broadcast might make that feature a bit more robust for others), so going to merge now.

PR #2264, opened 2019-10-08
Nonblocking packed range push

loganharbour:
@friedmud has already used something very similar, I just cleaned it up to match `Parallel::push_parallel_vector_data`. There's quite a bit of duplication and I'm not sure if there's a way to cleanly avoid it.

I'll work on some unit tests in a bit, but would like your input @roystgnr on this so far.

### DIFF
--- a/include/parallel/parallel_implementation.h
+++ b/include/parallel/parallel_implementation.h
@@ -1168,10 +1168,6 @@ inline void Communicator::nonblocking_receive_packed_range (const unsigned int s
   // Make the Request::wait() then handle deleting the buffer
   req.add_post_wait_work
     (new Parallel::PostWaitDeleteBuffer<std::vector<buffer_t>>(buffer));
-
-  // The MessageTag should stay registered for the Request lifetime
-  req.add_post_wait_work
-    (new Parallel::PostWaitDereferenceTag(tag));
 }
 
 
@@ -3315,20 +3311,14 @@ inline bool Communicator::possibly_receive_packed_range (unsigned int & src_proc
 {
   LOG_SCOPE("possibly_receive_packed_range()", "Parallel");
 
-  Status stat((StandardType<typename Packing<T>::buffer_type>()));
+  bool int_flag = 0;
 
-  int int_flag = 0;
-
-  libmesh_assert(src_processor_id < this->size() ||
-                 src_processor_id == any_source);
-
-  libmesh_call_mpi(MPI_Iprobe(src_processor_id,
-                              tag.value(),
-                              this->get(),
-                              &int_flag,
-                              stat.get()));
+  auto stat = packed_range_probe<T>(src_processor_id, tag, int_flag);
 
   if (int_flag)
+  {
+    src_processor_id = stat.source();
+
     nonblocking_receive_packed_range(src_processor_id,
                                      context,
                                      out,
@@ -3336,6 +3326,15 @@ inline bool Communicator::possibly_receive_packed_range (unsigned int & src_proc
                                      req,
                                      stat,
                                      tag);
+
+     // The MessageTag should stay registered for the Request lifetime
+     req.add_post_wait_work
+       (new Parallel::PostWaitDereferenceTag(tag));
+  }
+
+  libmesh_assert(!int_flag || (int_flag &&
+                               src_processor_id < this->size() &&
+                               src_processor_id != any_source));
 
   return int_flag;
 }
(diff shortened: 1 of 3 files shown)


include/parallel/parallel_implementation.h
```diff
@@ -1168,6 +1168,10 @@ inline void Communicator::nonblocking_receive_packed_range (const unsigned int s
   // Make the Request::wait() then handle deleting the buffer
   req.add_post_wait_work
     (new Parallel::PostWaitDeleteBuffer<std::vector<buffer_t>>(buffer));
+
+  // The MessageTag should stay registered for the Request lifetime
+  req.add_post_wait_work
+    (new Parallel::PostWaitDereferenceTag(tag));
```
roystgnr: This is a straight-up bug fix, right, having nothing to do with the rest of your changes?

include/parallel/parallel_sync.h
```diff
@@ -100,6 +100,34 @@ void pull_parallel_vector_data(const Communicator & comm,
                                ActionFunctor & act_on_data,
                                const datum * example);
 
+/**
+* Send and receive and act on vectors of data. Similar to
+* push_parallel_vector_data, except the vectors are packed and unpacked
+* using the Parallel::Packing routines.
```
roystgnr: We don't actually want to use vector inputs in this case, do we?  There's no performance benefit since we're serializing into buffers anyway, and working with iterator ranges is a lot more flexible.

include/parallel/parallel_sync.h
```diff
@@ -309,6 +337,182 @@ void push_parallel_vector_data(const Communicator & comm,
...
+{
+  // This function must be run on all processors at once
+  libmesh_parallel_only(comm);
+
+  // This function implements the "NBX" algorithm from
+  // https://htor.inf.ethz.ch/publications/img/hoefler-dsde-protocols.pdf
+
+  // This function only works for "flat" data that we can pre-size
+  // receive buffers for: a map to vectors-of-standard-types, not e.g.
+  // vectors-of-vectors.
```
roystgnr: This comment is an atavism in this context, right?  Whether we do vectors-of-packable-types or iterator-pairs-of-packable-types or whatever, the whole point of this function is going to be *not* needing standard-types, pre-sized receive buffers, etc.

roystgnr:
This is definitely long-overdue, thank you.

Whether we stick to the vectors requirement or loosen it to work with maps of iterator ranges, would it be reasonable to use a common implementation for push_parallel_packed_range and push_parallel_data?  Out of about a hundred lines of code there, the only differences at the moment are the StandardType-vs-output_type construction, the send-vs-nonblocking_send_packed_range calls, and the receive-vs-nonblocking_receive_packed_range calls.  It should be possible to have two shim functions that each perform the proper construction and set up the proper calls as functors, then hand off to a single common core function.

(note that this is "input" as requested, not "criticism"; using copied-and-pasted-and-tweaked code with a vague intention of refactoring it years later is a time-honored libMesh tradition)

include/parallel/parallel_implementation.h
loganharbour: Correct.

include/parallel/parallel_sync.h
loganharbour: Yep... copy pasta failure

loganharbour:
> Whether we stick to the vectors requirement or loosen it to work with maps of iterator ranges, would it be reasonable to use a common implementation for push_parallel_packed_range and push_parallel_data? 

Yes, that's what we were thinking originally. I'll see if I'm able to get something like that working in a reasonable amount of time.

> (note that this is "input" as requested, not "criticism"; using copied-and-pasted-and-tweaked code with a vague intention of refactoring it years later is a time-honored libMesh tradition)

And if not, this'll work too 😬

include/parallel/parallel_sync.h
loganharbour: I think this is reasonable, but might be a little more than I can wrap my head around. I believe @friedmud wanted to add some input to this

roystgnr:
If you're short on time, I'd say switching from vectors to ranges is more urgent than refactoring.  We can refactor innards "for free" at our leisure, whereas if we change the APIs then we're going to want to leave a shim of the older API around for a while.

friedmud:
I'm fairly against changing the interface to iterators.  The interface is already complicated (and lengthy) enough.  I think the vector interface is nice and simple and easy to approach.

I wouldn't be against the internal version getting changed to iterators - but I would want to maintain a shell API that takes vectors directly.

roystgnr:
Okay.  In that case I guess the only things that need to be changed now are a couple of the incorrect comments.

roystgnr:
The subpackage is just about ready to merge; we'll edit this to match afterwards.

loganharbour:
Sounds good. We're ready for this whenever possible so I'll be on the lookout to change when the subpackage is merged.

loganharbour:
Closing due to #2333

PR #2265, opened 2019-10-08
Fix PETSc deprecation warnings.

jwpeterson:
I did not test this with older PETSc yet, just going with the versions recommended in the PETSc deprecation messages:
```
PETSC_STATIC_INLINE PETSC_DEPRECATED_FUNCTION("Use DMSetSection() (since v3.9)")
PetscErrorCode DMSetDefaultSection(DM dm, PetscSection s) {return DMSetSection(dm,s);}
```
```
PETSC_STATIC_INLINE PETSC_DEPRECATED_FUNCTION("Use DMSetSectionSF() (since v3.12)") 
PetscErrorCode DMSetDefaultSF(DM dm, PetscSF s) {return DMSetSectionSF(dm,s);}
```

### DIFF
--- a/src/solvers/petsc_dm_wrapper.C
+++ b/src/solvers/petsc_dm_wrapper.C
@@ -517,7 +517,11 @@ namespace libMesh
 
         // Build the PetscSection and attach it to the DM
         this->build_section(system, section);
+#if PETSC_VERSION_LESS_THAN(3,9,0)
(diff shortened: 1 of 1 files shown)


jwpeterson:
Looks like the Min GCC test still uses PETSc 3.7.6 so I threw it in as well.

PR #2267, opened 2019-10-10
GCC 9.0 warning fixes

permcody:
- explicitly declare used ctors and methods if they are implicitly being called
- ignore this same warning for contribs.

### DIFF
--- a/include/geom/point.h
+++ b/include/geom/point.h
@@ -64,6 +64,11 @@ public:
   {}
 
   /**
+   * Copy-assignment operator.
+   */
+  Point& operator=(const Point & p) = default;
+
+  /**
    * Disambiguate constructing from non-Real scalars
    */
   template <typename T,
--- a/include/utils/ignore_warnings.h
+++ b/include/utils/ignore_warnings.h
@@ -86,6 +86,9 @@
 #if (__GNUC__ > 7)
 #pragma GCC diagnostic ignored "-Wparentheses"
 #pragma GCC diagnostic ignored "-Wcast-function-type"
+#if (__GNUC__ > 8)
+#pragma GCC diagnostic ignored "-Wdeprecated-copy"
+#endif // GCC > 8
 #endif // GCC > 7
 #endif // GCC > 6
 #endif // GCC > 5


roystgnr:
Nice, thanks!  Have you tried it with the --enable-paranoid-warnings turned on yet?  Looks like there's a handful of new warnings we'll want to start turning on, too...

Damn, now I want to build a gcc 9 module...

permcody:
@roystgnr - I highly doubt that this is the only new warning with GCC 9. The latest GCC is now warning if you have a subset of the "copy" type methods defined if you have others defined and are using them. It says that the "implicit declaration is deprecated". This is really handy since it helps developers deliver complete classes that end up being copied. I only fixed the single warning that popped up when compiling MOOSE with warnings as errors turned on (our default). We have several more in MOOSE that I'm fixing up. I'm sure more work is needed but this takes care of our immediate need.

permcody:
Whoa... Looks like we aren't doing enough testing on libMesh? Looks like we've picked up odd build regressions. These appear to be repeatable too:
https://civet.inl.gov/pr/9446/

roystgnr:
Wow.  I'm sure all the "PETSc Alt" failures can be fixed easily, but I'm not sure what to do about that clang internal error.

permcody:
I'm not sure that the clang internal error is repeatable. I'll investigate a bit more. I'll push up another PR shortly with the patches that all went in on Thursday and see what's left.

permcody:
@roystgnr - I ran a git bisect using Clang 3.5.1 (one of MOOSE's current minimum compilers) and narrowed down the current failure to this: https://github.com/libMesh/libmesh/commit/1684e753dd291354ed4e79ef3e26ebc4ff1fbca6

I'm not too attached to any particular version of the compiler but we will either have to bump our minimum now or come up with a workaround if we want to maintain that version. The diagnostics aren't bad, but they aren't great for what's happening. It's definitely a call to your newly "improved" `libmesh_assert_less` macro from `transient_system.C:126`, but I can't immediately see why, or what the problem is. I have the build sitting on my linux system if you want me to try anything different. I might experiment with a few extra parenthesis or trying to narrow down exactly what part of that construct is killing the compiler. If you have any ideas, let me know.

permcody:
FYI - I was unable to find a simple workaround to this. I tried adding additional sets of parenthesis around the expressions and even using a function object instead of the raw < operator (since the compiler was complaining about the ">" token when it crashed. No go. 3.5.1 isn't going to compile that code or a variation of that code. 

@milljm is going to look at bumping the min version of the clang compiler to 3.6.x tomorrow to see if the fix is in that version. The other alternative is to put a conditional in so that the logic remains as it was for just old compiler. That seems like a simple enough fix if libMesh wants to support the older compiler as well.

roystgnr:
I'm happy with a conditional as a workaround once we figure out what versions of the compiler we need the workaround for.  Using the old macro for invocations of the new macro will undoubtedly cause some signed/unsigned comparison warnings but we don't need to support --enable-werror on every old compiler, so long as things do at least compile.

PR #2268, opened 2019-10-10
DMSetSection requires 3.12 or newer.

jwpeterson:
Refs #2265.

### DIFF
--- a/src/solvers/petsc_dm_wrapper.C
+++ b/src/solvers/petsc_dm_wrapper.C
@@ -517,7 +517,7 @@ namespace libMesh
 
         // Build the PetscSection and attach it to the DM
         this->build_section(system, section);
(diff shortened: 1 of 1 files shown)


jwpeterson:
I backported the previous fix, so this one also needs to be backported (#2236).

src/solvers/petsc_dm_wrapper.C
```diff
@@ -517,7 +517,7 @@ namespace libMesh
 
         // Build the PetscSection and attach it to the DM
         this->build_section(system, section);
-#if PETSC_VERSION_LESS_THAN(3,9,0)
+#if PETSC_VERSION_LESS_THAN(3,12,0)
```
permcody: This API appeared in 3.10 as I mentioned on your original commit. Setting this Macro for 3.12 should work though.

PR #2271, opened 2019-10-11
adapted example ReducedBasisEx7 for PETSc >= 3.9.5

### LINKED ISSUE #1890: issue with example reduced_basis_7: PETSc >= 3.9 / no MUMPs
I am compiling against the git-branch (3.10.1) from PETSc and saw that
the option ``-pc_factor_mat_solver_package`` is deprecated (since 3.9.) and is replaced by ``-pc_factor_mat_solver_type``.
Moreover, when Petsc is configured without Mumps, this option should not be used.

As I see it, the 'run.sh'-files are not touched by any configure-step. Is there some other (easy) way to switch between the name of arguments? 
The more important issue is: If PETSc is configured without Mumps, this should not be used at all.
One can of course skip this example, but it runs with just 
``options="-online_mode 0 -ksp_type preonly -pc_type lu"``.
Would you still skip it or do you have some standard way to run examples differently depending on the configuration?

BalticPinguin:
... finally I have some suggestion to fix  #1890 
Maybe it is not the most elegant solution, but it seems to fix the issue.

### DIFF
--- a/examples/reduced_basis/reduced_basis_ex7/reduced_basis_ex7.C
+++ b/examples/reduced_basis/reduced_basis_ex7/reduced_basis_ex7.C
@@ -84,6 +84,43 @@ int main (int argc, char** argv)
   // FIXME: This example currently segfaults with Trilinos?
   libmesh_example_requires(libMesh::default_solver_package() == PETSC_SOLVERS, "--enable-petsc");
 
+#if LIBMESH_HAVE_PETSC
+  // lets parse the input to check which set of arguments was provided.
+  // Skip the runs with wrong arguments:
+  GetPot input(argc, argv);
+
+  // skip if the specification of the solver-package is given in wrong syntax:
+#if PETSC_VERSION_LESS_THAN(3,9,0)
+  if (input.search("-pc_factor_mat_solver_type"))
+   {
+      out<<"LibMesh was configured with PETSc < 3.9, but comman-line options"<<std::endl;
+      out<<"use syntax understood by later vernions only. Skipping now."<<std::endl;
+      return 77;
+   }
+#else
+  if (input.search("-pc_factor_mat_solver_package"))
+    {
+      out<<"LibMesh was configured with PETSc >= 3.9, but comman-line options"<<std::endl;
+      out<<"use deprecated syntax. Skipping now."<<std::endl;
+      return 77;
+    }
+#endif
+
+  // check that we have the required solver:
+  std::string solver;
+#ifndef PETSC_HAVE_MUMPS
+  solver = input.next("mumps");
+  if (solver == "mumps")
+    libmesh_example_requires(false, "petsc having mumps");
+#endif
+#ifndef PETSC_HAVE_SUPERLU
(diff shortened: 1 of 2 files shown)


examples/reduced_basis/reduced_basis_ex7/reduced_basis_ex7.C
```diff
@@ -84,6 +84,43 @@ int main (int argc, char** argv)
...
+    {
+      out<<"LibMesh was configured with PETSc >= 3.9, but comman-line options"<<std::endl;
+      out<<"use deprecated syntax. Skipping now."<<std::endl;
+      return 77;
+    }
+#endif
+
+  // check that we have the required solver:
+  std::string solver;
+#ifndef PETSC_HAVE_MUMPS
```
jwpeterson: Hmm... I guess this uses the #define directly from petscconf.h which should be fine, but note that we also set `LIBMESH_PETSC_HAVE_MUMPS` and `LIBMESH_PETSC_HAVE_SUPERLU_DIST` in our own config.

examples/reduced_basis/reduced_basis_ex7/reduced_basis_ex7.C
```diff
@@ -84,6 +84,43 @@ int main (int argc, char** argv)
...
+#if LIBMESH_HAVE_PETSC
+  // lets parse the input to check which set of arguments was provided.
+  // Skip the runs with wrong arguments:
+  GetPot input(argc, argv);
+
+  // skip if the specification of the solver-package is given in wrong syntax:
+#if PETSC_VERSION_LESS_THAN(3,9,0)
+  if (input.search("-pc_factor_mat_solver_type"))
+   {
+      out<<"LibMesh was configured with PETSc < 3.9, but comman-line options"<<std::endl;
```
jwpeterson: Print statements should full `libMesh::out` and typo: "comman" -> "command"

examples/reduced_basis/reduced_basis_ex7/reduced_basis_ex7.C
```diff
@@ -84,6 +84,43 @@ int main (int argc, char** argv)
...
+    libmesh_example_requires(false, "petsc having mumps");
+#endif
+#ifndef PETSC_HAVE_SUPERLU
+  solver = input.next("superlu");
+  if (solver == "superlu")
+    libmesh_example_requires(false, "petsc having superlu");
+#endif
+
+#endif //LIBMESH_HAVE_PETSC
+
```
jwpeterson: Since this isn't an `#else`, I think you are going to end up calling `input.next` here twice which probably won't do the right thing? Also, the input to `next` is meant to be a default value, which you can just leave as an empty string, i.e. call `solver = input.next(solver)` outside all the `#if` blocks and then handle unrecognized values appropriately below.

BalticPinguin:
thank you for the comments; I fully agree.

jwpeterson:
Thanks for this. I will make a couple of cosmetic changes and then merge.

PR #2277, opened 2019-10-14
Add --disable-openmp configure option

### LINKED ISSUE #2131: Add --disable-openmp configure option
We currently "enable" OpenMP (i.e. set the define `LIBMESH_HAVE_OPENMP` and pass the corresponding flags) any time the compiler supports one of the possible OpenMP flags. Basically in threads.m4, we call
```
AX_OPENMP([],[enableopenmp=no])
```
no matter what without any option to disable it. When using helgrind for debugging, it is apparently necessary to have pthreads support turned on but *not* be using OpenMP (@lindsayad made a thread about this on libmesh-devel). So, we need to be able to control these two things independently.

There might also be an even more subtle middle ground where one would want to pass the `-fopenmp` flags to the compiler, but not actually use any OpenMP pragmas in the code... that should also be possible to support if necessary, but I think for now the ability to simply turn off OpenMP is all that's needed.

lindsayad:
Closes #2131

### DIFF
--- a/m4/threads.m4
+++ b/m4/threads.m4
@@ -17,12 +17,12 @@ AC_DEFUN([ACX_BEST_THREAD],
 
   AC_MSG_RESULT([<<< User requested thread model: $requested_thread_model >>>])
 
-  dnl We *always* detect the compiler's OpenMP support, because libMesh
-  dnl is frequently compiled with other libraries that use OpenMP
-  dnl directives (it can also use a few itself [0]) and if you don't set
-  dnl OMP_NUM_THREADS in the environment or call omp_set_num_threads()
-  dnl programmatically, you get the questionable default of "1 thread
-  dnl per CPU" [1]. This can lead to drastic over-subscription of
+  dnl We *always* detect the compiler's OpenMP support unless explicitly told not
+  dnl to with --disable-openmp, because libMesh is frequently compiled with other
+  dnl libraries that use OpenMP directives (it can also use a few itself [0]) and
+  dnl if you don't set OMP_NUM_THREADS in the environment or call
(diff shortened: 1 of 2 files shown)


lindsayad:
@permcody This is to be able to use helgrind meaningfully

m4/threads.m4
```diff
@@ -32,39 +32,47 @@ AC_DEFUN([ACX_BEST_THREAD],
...
-  dnl If the user explicitly requested the openmp threading model and
-  dnl this compiler doesn't support OpenMP for some reason, we throw an
-  dnl error instead of silently continuing.
-  AS_IF([test "x$requested_thread_model" = "xopenmp"],
-        [
-          AS_IF([test "x$enableopenmp" = "xno"],
-                [AC_MSG_ERROR([requested openmp threading model, but compiler does not support openmp.])])
-        ])
+  AS_IF([test "x$enableopenmp" = "xyes"],
+        [AX_OPENMP([],[enableopenmp=no])]
```
jwpeterson: For instance, here there is a closing `]` but no comma, so I guess it doesn't start the third (else) clause of the `AS_IF` statement?

lindsayad: Correct. If `enableopenmp = yes`, then we execute the `AX_OPENMP` macro and then we also execute the following macro (the `AS_IF` on whether the macro successfully set `OPENMP_CXXFLAGS`)

lindsayad: The corresponding `else` for this `AS_IF` (`test "x$enableopenmp" = "xyes"`) is the `Error if the user simultaneously requested --with-thread-model=openmp and --disable-openmp` bit

PR #2280, opened 2019-10-15
DynaIO fixes, calculate_map fix/optimization

roystgnr:
Bug fixes for #2259, plus a more complicated unit test that caught the original bugs, plus a performance fix for the use of FE in a mapping-only context (which doubles as a bug fix when using rational basis mapping functions).


roystgnr:
Going to merge this, with fingers crossed that nobody else is relying on our previous "you can prerequest one kind of mapping data, then reinit, then request a different kind of mapping data too after the fact because we precomputed everything whether you asked for it or not" behavior.  We'll still be supporting "if you prerequest nothing then we got into backwards compatibility mode and compute everything" uses, just not that mixed scenario.

PR #2281, opened 2019-10-15
Workaround for older clang compiler bug.

permcody:
The libmesh assertion macros fail to parse on older clang compilers. We'll
juse fall back to the previous behavior when using those compilers.

### DIFF
--- a/include/base/libmesh_common.h
+++ b/include/base/libmesh_common.h
@@ -280,7 +280,7 @@ extern bool warned_about_auto_ptr;
 #else
 
 #define libmesh_assertion_types(expr1,expr2)                            \
-  typedef typename std::decay<decltype(expr1)>::type libmesh_type1;  \
+  typedef typename std::decay<decltype(expr1)>::type libmesh_type1;     \
   typedef typename std::decay<decltype(expr2)>::type libmesh_type2
 
 #define libmesh_assert_msg(asserted, msg)                               \
@@ -310,6 +310,8 @@ extern bool warned_about_auto_ptr;
       libMesh::err << "Assertion `" #expr1 " != " #expr2 "' failed.\n" #expr1 " = " << (expr1) << "\n" #expr2 " = " << (expr2) << '\n' << msg << std::endl; \
       libmesh_error();                                                  \
     } } while (0)
+
+#if defined(__clang__) && (__clang_major__ > 3)
 
 #define libmesh_assert_less_msg(expr1,expr2, msg)                       \
   do {                                                                  \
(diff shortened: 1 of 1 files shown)


permcody:
@roystgnr - is this acceptable, or are you even interested in maintaining Clang 3.5.1 support? I haven't verified that Clang 3.6 or even 4.0 fixes this issue yet, but we will look at that today.

jwpeterson:
Just curious if you already tried making a version of the `libmesh_assertion_types()` macro that does "nothing" for older versions of clang, since the error seems to be with the `std::decay` call itself? At least that way you wouldn't have to duplicate every version of the macro... I think all of this was just to avoid warnings in debug mode, so if there are still warnings in old versions of clang that is probably fine.

permcody:
> Just curious if you already tried making a version of the libmesh_assertion_types() macro that does "nothing" for older versions of clang, since the error seems to be with the std::decay call itself? 

I did. That didn't work since the `static_cast` failed for several existing expressions in libMesh. The decay is important for the cast to work in all cases. I really did try a lot of combinations yesterday since I had the environment available to me to play with. To be clear, it's not really the decay that's causing the problem. It has something to do with the parser when it encounters all of the "< >" tokens that get expanded. It believe (but can't determine for sure in the LLVM stack trace), that the actual issues is with the pointer in the expression being tested. 

I couldn't come up with a way to avoid the duplication, but maybe somebody more clever could.

jwpeterson:
I am using `g++ (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0` , and therefore this statement: 
```
#if defined(__clang__) && (__clang_major__ > 3)
```
is false for me, so I'm now getting the fallback code, and I have a bunch of  warnings like the following:
```
./include/libmesh/libmesh_common.h:359:19: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     if (!((expr1) < (expr2))) {                                         \
          ~~~~~~~~~^~~~~~~~~~
./include/libmesh/libmesh_common.h:393:42: note: in expansion of macro ‘libmesh_assert_less_msg’
 #define libmesh_assert_less(expr1,expr2) libmesh_assert_less_msg(expr1,expr2, "")
                                          ^~~~~~~~~~~~~~~~~~~~~~~
./include/libmesh/meshfree_interpolation.h:262:7: note: in expansion of macro ‘libmesh_assert_less’
       libmesh_assert_less (dim, PLDim);
       ^~~~~~~~~~~~~~~~~~~
```
Would it work right if we put the fallback code first, guarded with:
```
#if defined(__clang__) && (__clang_major__ <= 3)
```
then every other compiler would get the new code, right?

PR #2287, opened 2019-10-16
test for mpi version

BalticPinguin:
after the unsuccessfull approach in #1920, I'd suggest to test the MPI-version via the APIs itself:
  * we don't rely on ``MPI_VERSION`` strings
  * in case of incomplete implementation of standards, this test still holds
  * is easily extensible to further functions that may or may not be implemented.

I hope, it behaves as expected for other configurations as well...

### DIFF
--- a/configure
+++ b/configure
@@ -33969,6 +33969,48 @@ fi
 
 
 
+  if test "x$enablempi" != xno; then :
+
+                                cat confdefs.h - <<_ACEOF >conftest.$ac_ext
+/* end confdefs.h.  */
+#include <mpi.h>
+int
+main ()
+{
+
+                                        const int i=1;
+                                        int position, j, a2;
+                                        int myrank;
+                                        char buff[ 10];
+                                        MPI_Comm_rank(MPI_COMM_WORLD, &myrank);
+                                        MPI_Pack(&i, 1, MPI_INT, buff, 10, &position, MPI_COMM_WORLD)
+
+  ;
+  return 0;
+}
+_ACEOF
+if ac_fn_cxx_try_link "$LINENO"; then :
+
+
+$as_echo "#define HAVE_MPI 1" >>confdefs.h
+
+
+else
+
+                       { $as_echo "$as_me:${as_lineno-$LINENO}: WARNING: \"MPI-version seems to be too low: Need MPI 3.X or compatible. Disable MPI now...\"" >&5
+$as_echo "$as_me: WARNING: \"MPI-version seems to be too low: Need MPI 3.X or compatible. Disable MPI now...\"" >&2;}
+                       enablempi=no
+
+fi
+rm -f core conftest.err conftest.$ac_objext \
+    conftest$ac_exeext conftest.$ac_ext
+
+fi
+
+
+
+
+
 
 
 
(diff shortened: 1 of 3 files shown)


BalticPinguin:
hmm... similar to #1920, it fails with
 ``configure: WARNING: "MPI-version seems to be too low: Need MPI 3.X or compatible. Disable MPI now..."``
(but seems to compile fine otherwise!)

Is there a way to look at the ``compile.log``, there the precise output of the test should be given...

roystgnr:
We really ought to be testing the output of configure in those parallel recipes.  There's no point in putting a non-MPI build through mpiexec over and over again until some I/O race condition finally trips.

I don't know how to get at the Civet configure.log, but let me try a few environments locally and see if I can reproduce the problem.

roystgnr:
On one system your branch works fine for me, but on the other I get
```
conftest.cpp:157:10: fatal error: mpi.h: No such file or directory
 #include <mpi.h>
```

The failing system is building with
```
g++ -o conftest -O0 -g -DDEBUG   conftest.cpp -ldl
```
so no mpicxx, no include paths, nothing, so I'm not surprised it's not finding my MPI module.

permcody:
> I don't know how to get at the Civet configure.log, but let me try a few environments locally and see if I can reproduce the problem.

A quick hack would be to cat it when we get a configure failure. That way it's just included in the log that you download. It would be handy to have something a little more sophisticated, but that doesn't exist and we don't really have the resources to make that improvement right now.

BalticPinguin:
I think, this one is outdated due to the `timpi` submodule and can be closed?

roystgnr:
Yeah, I'm afraid so.  Should we try to get this feature into https://github.com/libMesh/autoconf-submodule/ ?  What's the current status with old MPI versions?  I dug back through old tickets until I hit your comment:

> I think the discussion is distributed among too many threads and comment-sections by now; even for me constantly following it becomes hard to track the different branches :-)

and I gave up.

BalticPinguin:
>  Should we try to get this feature into https://github.com/libMesh/autoconf-submodule/ 

In some form, I guess it should be done so; I just tried to run configure with `openmpi-1.6.5` (which has mpi 2.1) where `configure` runs fine but `make` fails.

However, the correct place for this is than the configure of `timpi`, right?

roystgnr:
> However, the correct place for this is than the configure of timpi, right?

Maybe in the long run, and then libMesh would pull its configuration from TIMPI?  But right now, we use the same m4 files in both TIMPI and libMesh (by keeping them in the same submodule), and they get called from the configure.ac of each.

jwpeterson:

BalticPinguin:
I am not sure if there is sense to keep this anyway?
The `m4/mpi.m4`-file (containing most of the modifications) is not present anymore in libMesh.
If at all, I think this now belongs to `TIMPI`; but after all the time, this might be not relevant anymore?

Probably it is better to close it as is?

roystgnr:
It's actually one submodule past TIMPI, now; acsm_mpi.m4 in https://github.com/libMesh/autoconf-submodule

But yeah, this isn't going to be mergeable here anymore; I like the idea but we'll need to rework it for there if we want the upgrades.

PR #2288, opened 2019-10-16
Fix ifdef logic so it works for GCC.

jwpeterson:
Refs #2281.

### DIFF
--- a/include/base/libmesh_common.h
+++ b/include/base/libmesh_common.h
@@ -312,7 +312,40 @@ extern bool warned_about_auto_ptr;
       libmesh_error();                                                  \
     } } while (0)
 
-#if defined(__clang__) && (__clang_major__ > 3)
+// Older Clang has a parsing bug that can't seem to handle the handle the decay statement when
+// expanding these macros. We'll just fall back to the previous behavior with those older compilers
+#if defined(__clang__) && (__clang_major__ <= 3)
+
+#define libmesh_assert_less_msg(expr1,expr2, msg)                       \
+  do {                                                                  \
+    if (!((expr1) < (expr2))) {                                         \
+      libMesh::err << "Assertion `" #expr1 " < " #expr2 "' failed.\n" #expr1 " = " << (expr1) << "\n" #expr2 " = " << (expr2) << '\n' << msg << std::endl; \
+      libmesh_error();                                                  \
+    } } while (0)
+
+#define libmesh_assert_greater_msg(expr1,expr2, msg)                    \
+  do {                                                                  \
+    if (!((expr1) > (expr2))) {                                         \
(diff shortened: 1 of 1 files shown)


roystgnr:
Damn, I can't believe I didn't see that.  How did #2281 even pass CI?  I could have sworn that at least one of our default recipes was a gcc build with --enable-werror and --enable-paranoid-warnings turned on.

permcody:
> I could have sworn that at least one of our default recipes was a gcc build with --enable-werror and --enable-paranoid-warnings turned on.

I don't think so. We can add it though. Do you typically just pass those flags when invoking "make" or do you add this to the configure line? Let me know what you'd like to do and I'll make the change.

permcody:
BTW - I added four new required targets to libMesh yesterday (doubling the number). There's now minimum clang and minimum gcc, both in opt and dbg.

jwpeterson:
> I don't think so. We can add it though.

This flag is turned on for the complex enabled testing IIRC. It definitely seems like it should have failed...

jwpeterson:
Those tests use GCC 8.3.0 and they apparently [compiled without any warnings](https://civet.inl.gov/job/409914/). Not really sure why we see warnings on GCC 7.4.0 but not 8.3.0.

jwpeterson:
The full configure flags for the complex testing are: 
```
../configure --with-methods=opt --prefix= --enable-silent-rules --enable-unique-id --disable-warnings --enable-openmp --disable-maintainer-mode --with-metis=PETSc --enable-complex --disable-petsc --enable-paranoid-warnings --enable-werror --enable-ifem
```
so possibly the warnings were triggered from PETSc code, and this test with `--disable-petsc` doesn't produce them?

roystgnr:
That could explain it, yeah.

IMHO at this point the --enable-paranoid-warnings --enable-werror combo ought to be solid enough to just add it to configure on every recipe.  That'd also make me feel more confident about adding new paranoid warnings (e.g. for GCC 9) without worrying that my own test environment alone isn't sufficient to prevent me from pulling the rug out from under other developers.  But the major negative is that those warnings are more often than not false positives, so enforcing them forces developers to be more obsessive about avoiding otherwise-unimportant errors.

permcody:
OK - so those are the pros and cons. Would you like me to take specific action? How about adding it to the "Test debug" target at least for starters if you don't want to go nuts on all targets. You could also opt to add it to a "new" target that is allowed to fail. There are many options.

roystgnr:
Yeah, sure, let's just bump up coverage one target at a time, starting with "Test debug".  That way if there are any other surprise warnings that show up in some configs but not others, we'll at least not be swamped with them all at once.

PR #2289, opened 2019-10-16
In Elem::level check interior_parent->level()

lindsayad:
In MOOSE mortar when building the mortar segment mesh,
we set the interior_parent for the mortar segments, but
we do not set the parent, because parent implies the child
was formed through adaptivity. However, when a mesh has been
refined and the interior_parent has been refined, then there
are failed assertions in debug mode during prepare_for_use
because the interior_parent level != 0 while previously if the
parent was nullptr the mortar segment would report as being level 0.
Hence I think it makes sense to check the interior_parent level
if the parent is nullptr.

We'll see if this breaks things...

### DIFF
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -2484,7 +2484,12 @@ unsigned int Elem::level() const
   // or by the user, so I am a
   // level-0 element
   if (this->parent() == nullptr)
-    return 0;
+  {
+    if (this->interior_parent() == nullptr)
+      return 0;
+    else
+      return interior_parent()->level();
+  }
 
   // if the parent and this element are of different
   // dimensionality we are at the same level as


lindsayad:
Oh well, looks like that doesn't work...

permcody:
> Wow so many tests now

Yeah, I think we had three separate regressions that made it in since the last libMesh update. I guess we need to expand testing on the libMesh side, or face that problem again in the future.

lindsayad:
Ugh I can see that this would infinitely recurse because `interior_parent()` calls `level()` (at least in debug mode)

lindsayad:
I think I'm being dense on the MOOSE side...I'm going to try setting the mortar segment mesh parent to the lower dimensional element parent...

roystgnr:
Definitely don't like this change, or anything conceptually adjacent to it.  An element with non-zero level() but with no parent is a bug waiting to happen.

If it's possible to create a parent mortar segment such that that parent's interior_parent is the interior_parent's parent, then that's the way to go.

If that's not always possible, which would be reasonable, then we need to just get rid of those assertions, but I fear we'll also have to make changes in some algorithms they were asserting preconditions for.

lindsayad:
Closing for the reasons @roystgnr cited

PR #2296, opened 2019-10-18
When building node list from side list, first clear node containers

lindsayad:
To me it's all in the name...we're building a node list from the
side list, and hence the former should depend only on the latter.
Consequently the node lists should be cleared before looping
over the sides. I had to do this for some GeoemetricSearch work
in MOOSE combined with mesh coarsening. The node list contained
some nodes that did not exist on the coarsened mesh and it was
because the BoundaryInfo node containers were not getting cleared.

### DIFF
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -175,24 +175,30 @@ public:
 
   /**
    * Add \p Node \p node with boundary id \p id to the boundary
-   * information data structures.
+   * information data structures. \p from_side indicates whether
+   * this node is being added from a sideset
    */
   void add_node (const Node * node,
-                 const boundary_id_type id);
+                 const boundary_id_type id,
+                 bool from_side = false);
 
   /**
    * Add node number \p node with boundary id \p id to the boundary
-   * information data structures.
+   * information data structures. \p from_side indicates whether
+   * this node is being added from a sideset
    */
   void add_node (const dof_id_type node,
-                 const boundary_id_type id);
+                 const boundary_id_type id,
+                 bool from_side = false);
 
   /**
    * Add \p Node \p node with boundary ids \p ids to the boundary
-   * information data structure.
+   * information data structure. \p from_side indicates whether
+   * this node is being added from a sideset
    */
   void add_node (const Node * node,
-                 const std::vector<boundary_id_type> & ids);
+                 const std::vector<boundary_id_type> & ids,
+                 bool from_side = false);
 
   /**
    * Clears all the boundary information from all of the nodes in the mesh
@@ -920,6 +926,17 @@ private:
                 boundary_id_type> _boundary_node_id;
 
   /**
+   * Map from nodeset nodes to their boundary IDs
+   *
+   * \note Unlike \p _boundary_node_id, this structure only contains
+   * nodes coming from nodesets. E.g. this will
+   * not contain nodes coming from \p build_node_list_from_side_list
+   */
+  std::multimap<const Node *,
+                boundary_id_type> _boundary_node_id_from_nodeset;
+
+
+  /**
    * Data structure that maps edges of elements
    * to boundary ids. This is only relevant in 3D.
    */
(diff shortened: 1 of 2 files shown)


jwpeterson:
Hmm... looks like there are quite a few tests in MOOSE that relied on the old (incorrect) behavior? Or are those legitimate failures? Either way, :+1: to the idea.

roystgnr:
The documentation in boundary_info.h does specifically say *adds*, so I wouldn't fault anyone for relying on that behavior.  Kind of wish it had been reflected in the name rather than just the comment, though.

lindsayad:
Sigh. Should I add a new method then? `build_node_list_from_just_side_list`? 😄

roystgnr:
Hmm... usually I'd say the "add" version of an API is enough because users can always get the "replace" version by manually calling clear() themselves first, and that way we need half as many (container-expanding) APIs.  But can users actually call a selective clear() in this case?  Did we give them sufficiently flexible accessors?

lindsayad:
No, it's just a full-blown clear everything right now. Perhaps yea it would be better to add something like `clear_nodal(_containers)` and `clear_side(_containers)` as opposed to adding another `build` API.

roystgnr:
> Perhaps yea it would be better to add something like clear_nodal(_containers) and clear_side(_containers) as opposed to adding another build API.

That'd be my preference.

lindsayad:
Ok, I actually added some additional containers meant to represent nodes coming only from nodesets. Then when calling `build_node_list_from_side_list`, I start by setting `_boundary_node_id` to `_boundary_node_id_from_nodeset`.

I think this will work, but I feel like the best solution to my problem would actually be to call `BoundaryInfo::remove(Node *)` when performing coarsening similarly to how `BoundaryInfo::remove(Elem *)` is called. However, I guess the tricky thing would be that we only want to remove the nodes that won't be present after coarsening...

roystgnr:
This isn't yet going to work with parallel redistribution, is it?  When we pack and unpack a Node, we simply toss any boundary ids associated with it into the recipient's BoundaryInfo.  You'll need to add new data to the serialization to specify whether each id is nodeset or sideset associated.  Restart files would have the same problem and need yet another solution.

This is getting kind of Rube Goldberg now.  I think we've graduated from XY problem all the way to XYZ problem.  Could you describe the problem X you're trying to solve?  It sounds like you're generating nodesets from sidesets on a refined mesh, and some of those node boundary ids should go away again when coarsening eliminates the node, and we're not yet doing that?  But it looks like both DistributedMesh::delete_node and ReplicatedMesh::delete_node are remembering to call BoundaryInfo::remove.

lindsayad:
> This is getting kind of Rube Goldberg now.

First I learned Chesterton's fence from you, and now I've learned about Rube Goldberg. You're a fountain of great references @roystgnr.

> Could you describe the problem X you're trying to solve? It sounds like you're generating nodesets from sidesets on a refined mesh, and some of those node boundary ids should go away again when coarsening eliminates the node, and we're not yet doing that? But it looks like both DistributedMesh::delete_node and ReplicatedMesh::delete_node are remembering to call BoundaryInfo::remove.

So the problem is as simple as this: after coarsening the mesh (via `uniformly_coarsen` in this case), `_boundary_node_id` contains nodes that existed only on the previous refinement level, e.g. nodes that should no longer exist. When is `Replicated(Distributed)Mesh::delete_node` supposed to be called? Is that supposed to be the job of me as the user? So from within the debugger, I've checked whether `delete_node` gets called during the scope of `MeshRefinement::uniformly_coarsen`; the answer is that it doesn't.

roystgnr:
> You're a fountain of great references

My two secrets are reading a lot and getting old.

> When is Replicated(Distributed)Mesh::delete_node supposed to be called?

The catch here is the idea of "subactive elements".  When we first coarsen the mesh, we don't immediately delete the finest level of elements, much less their nodes.  This is partly because we sometimes want to so something special with them later (e.g. the Geometric MultiGrid stuff works by coarsening for work on coarsened meshes but then backing out of the coarsening to the original mesh without deleting elements in between), and mostly because we usually need to hang on to them for a little while to enable later projecting solution vectors (we couldn't do the projection calculation if we were to have already thrown away the old degree of freedom indexing).

> Is that supposed to be the job of me as the user?

Sort of.  Only indirectly.  Generally MeshBase::contract() is what does deletions (via renumber_nodes_and_elements() for nodes, for atavistic reasons), and usually the user code hits that via EquationSystems::reinit().

I can see how this would be annoying if you're trying to do work after the coarsening but before the reinit.  We have a simple Elem::is_subactive() that you can use to test those, but there's no simple way to see if a Node is only attached to subactive elements.  There's not even a non-parallel complicated way to do that in general, considering distributed meshes with nodes that might be attached to remote elements.

lindsayad:
Ok, I think I was able to get everything worked out. After coarsening, I had to make sure I worked in the following order:

1. `EquationSystem::reinit_solutions` // Restrict vectors and then contract mesh
2. `MooseMesh::meshChanged` // Update node/side lists
3. `GeometricSearchData::reinit` // Compute geometric information which will be needed for sparsity building
4. `EquationSystems::reinit_systems` // Build sparsity

So I'm going to consider this closed!

PR #2307, opened 2019-10-23
Update tetgen to version 1.5.1

BalticPinguin:
the new version is out since August this year and it seems to me as if there are several nice changes; at least for my program, the mesh generated with the new version has better properties.

But the lincense has changed: It is now a GNU Affero GPL and sais
``` 
You can
redistribute it and/or modify it under the terms of the GNU Affero
General Public License as published by the Free Software Foundation,
either version 3 of the License, or (at your option) any later
version. A copy of the GNU Affero General Public License is reproduced
below.
```
so I hope it is fine; but maybe I have overseen something?


jwpeterson:
I think there's no change for us as far as the license is concerned, it is still non-LGPL so we will only build it when configured with `--disable-strict-lgpl`. I will also turn on our CI testing for that config...

jwpeterson:
Way back in 2014 we fixed a stack buffer overflow in tetgen in 4a054cb38c97967b34e7a5b8fa5a74b0d57816ab. This fix never made it into upstream tetgen AFAIK and it's probably no longer relevant, but I thought I would mention it just in case.

jwpeterson:
> it's probably no longer relevant, but I thought I would mention it just in case.

Finally got around to looking at the diff, which was too big to be displayed on GitHub. Looks like our fix is still relevant and this patch undoes it, so I will have to reapply that patch before merging...
```diff
@@ -1072,24 +991,10 @@ int fast_expansion_sum_zeroelim(int elen, REAL *e, int flen, REAL *f, REAL *h)
     while ((eindex < elen) && (findex < flen)) {
       if ((fnow > enow) == (fnow > -enow)) {
         Two_Sum(Q, enow, Qnew, hh);
-        /**
-         * [JWP] Increment 'eindex,' but only update 'enow' if it is
-         * safe.  Otherwise, 'enow' remains unchanged from the previous
-         * iteration...
-         */
-        ++eindex;
-        if (eindex < elen)
-          enow = e[eindex];
+        enow = e[++eindex];
       } else {
         Two_Sum(Q, fnow, Qnew, hh);
-        /**
-         * [JWP] Increment 'findex,' but only update 'fnow' if it is
-         * safe.  Otherwise, 'fnow' remains unchanged from the previous
-         * iteration...
-         */
-        ++findex;
-        if (findex < flen)
-          fnow = f[findex];
+        fnow = f[++findex];
       }
       Q = Qnew;
       if (hh != 0.0) {
```

PR #2308, opened 2019-10-24
Fix namespace issue in libmesh_assert_xxx macros.

jwpeterson:
* In files without "using namespace libMesh;" at the top, these macros
  were generating the following:

  error: 'libmesh_comparator' was not declared in this scope

* Also include `<functional>` in libmesh_common.h now that we are using
  std::less and friends.


roystgnr:
Thanks again for catching this

PR #2312, opened 2019-10-29
Fix gcc9 warnings

roystgnr:
Looks like there's only a few of these that we hadn't already fixed; I can compile with --enable-werror --enable-paranoid-warnings with gcc 9.2 after this PR.

### DIFF
--- a/include/utils/mapvector.h
+++ b/include/utils/mapvector.h
@@ -57,8 +57,7 @@ public:
     veclike_iterator(const typename maptype::iterator & i)
       : it(i) {}
 
-    veclike_iterator(const veclike_iterator & i)
-      : it(i.it) {}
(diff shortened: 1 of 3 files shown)


roystgnr:
The catch wasn't defaulted-behavior-but-not-marked-default, the catch was another -Wdeprecated-copy warning:  Since we *hadn't* marked that effectively-default-copy-constructor as default, the compiler complained that we were trying to use the default copy-assignment-operator, on the theory that if the former is truly non-default then the latter's default is probably missing something important.

Looks like that's actually going to turn from a warning into an error in future C++ versions:

> The implicit definition of a copy assignment operator as defaulted is deprecated if the class has a user-declared copy constructor or a user-declared destructor (15.4, 15.8). In a future revision of this International Standard, these implicit definitions could become deleted (11.4).

PR #2316, opened 2019-10-30
Update abaqus_io.C

yw5aj:
Added support for reading ABAQUS .inp file with *Surface referring to element sets

### DIFF
--- a/src/mesh/abaqus_io.C
+++ b/src/mesh/abaqus_io.C
@@ -859,25 +859,25 @@ void AbaqusIO::read_sideset(std::string sideset_name, sideset_container_t & cont
     {
       // The strings are either of the form: "391, S2" or "Elset_1, S3"
 
-      // Read the element ID and the leading comma
-      _in >> elem_id >> c;
-      if (_in)
-      {
-        // For the form of "391, S2"
-        // Read another character (the 'S') and finally the side ID
-        _in >> c >> side_id;
+      // Read the first string (either elem_id or elset_name) as elem_id_or_set
+      std::getline(_in, elem_id_or_set, ',');
+      // Read the second string (side_id)
+      _in >> c >> side_id;
 
-        // Store this pair of data in the vector
+      std::istringstream iss_elem_id_or_set (elem_id_or_set);
+      // Assume the side set is of the form of "391, S2"
+      iss_elem_id_or_set >> elem_id;
+
+      if (iss_elem_id_or_set.eof())
+      {
+        // if the side set is of the form of "391, S2"
         id_storage.push_back( std::make_pair(elem_id, side_id) );
       }
         else
         {
-            // For the form of "Elset_1, S3"
-            _in.clear();
-            std::getline(_in, elset_name, ',');
-
-            // Read another character (the 'S') and finally the side ID
-            _in >> c >> side_id;
+            // if the side set is of the form of "Elset_1, S3"
+            iss_elem_id_or_set.clear();
+            iss_elem_id_or_set >> elset_name;
 
             const auto & vec = libmesh_map_find(_elemset_ids, elset_name);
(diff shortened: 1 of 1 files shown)


jwpeterson:
You may have already fixed this, but the complex numbers build failed because there was a warning and we have -Werror turned on for that build.
```
../src/mesh/abaqus_io.C:886:27: error: declaration of 'elem_id' shadows a previous local [-Werror=shadow]
         for (const auto & elem_id : it->second)
                           ^~~~~~~
../src/mesh/abaqus_io.C:852:15: note: shadowed declaration is here
   dof_id_type elem_id=0;
               ^~~~~~~
```

yw5aj:
@jwpeterson thanks a lot for looking into this for me! Just got this fixed :)

src/mesh/abaqus_io.C
```diff
@@ -852,30 +852,49 @@ void AbaqusIO::read_sideset(std::string sideset_name, sideset_container_t & cont
...
-
-      // Store this pair of data in the vector
-      id_storage.push_back( std::make_pair(elem_id, side_id) );
+        // Store this pair of data in the vector
+        id_storage.push_back( std::make_pair(elem_id, side_id) );
+      }
+        else
+        {
+            // For the form of "Elset_1, S3"
+            _in.clear();
```
jwpeterson: Hmm... are you are using `clear()` here to clear some bad or fail bit on the stream that was set during the previous read? If you have tested this and it works, I guess it's OK, but my preference would be to call `std::getline` to extract the entire line from the stream and then do some string processing to figure out which type of sideset is present.

src/mesh/abaqus_io.C
```diff
@@ -852,30 +852,49 @@ void AbaqusIO::read_sideset(std::string sideset_name, sideset_container_t & cont
...
+            std::getline(_in, elset_name, ',');
+
+            // Read another character (the 'S') and finally the side ID
+            _in >> c >> side_id;
+
+            auto it = _elemset_ids.find(elset_name);
+            if (it == _elemset_ids.end())
+            libmesh_error_msg(
+                "The element set referred to by the surface definition does not exist.");
```
jwpeterson: If you include "libmesh/utility.h" at the top of the file, you can shorten this bit to:
```
const auto & vec = libmesh_map_find(_elemset_ids, elset_name);
for (const auto & elem_id_in_elset : vec)
  ...
```
Also drop the curly braces for 1-line for loop. I can fix this before merge if you prefer.

yw5aj: Thanks for the comment! I'll fix this one.

src/mesh/abaqus_io.C
yw5aj: Hi John - yes. We try to read in the `elem_id`, which is a number; if it failed, then we reset the fail bit and try to read as a string. Just to make sure I got it right: you are suggesting that I use `std::getline` to read the whole line. And then I will try to convert the string to `istream` & use `>>` to treat the first part as a number; if that failed, I will treat the first part as a string (`elset_name`). In this case, no fail bit resetting is required. Is that understanding correct?

yw5aj: I edited the above paragraph a bit to make it clearer - you might want to check the webpage version to see the updates :)

yw5aj: Hi @jwpeterson ! I just modified the code per my understanding. Please let me know what you think - or feel free to edit directly as well; I can always test them out before merging :)

jwpeterson:
FYI @yw5aj I sent you an invite to join a libMesh group whose PRs are tested automatically. I just started all the tests manually just now but I'm confused, yesterday your PR seemed to be testing just fine on its own...

permcody:
No, I was activating them.

jwpeterson:
Replaced by #2317

PR #2317, opened 2019-10-31
Add support for reading ABAQUS .inp file with *Surface referring to element sets

jwpeterson:
@yw5aj I think this patch works for me on the test mesh you sent. I squashed together all the commits into one and since the original PR was based on your master branch (which I did not want to force-push) I opened a new PR.

Refs #2316.

### DIFF
--- a/src/mesh/abaqus_io.C
+++ b/src/mesh/abaqus_io.C
@@ -861,6 +861,13 @@ void AbaqusIO::read_sideset(std::string sideset_name, sideset_container_t & cont
       // Read first string up to and including the comma, which is discarded.
       std::getline(_in, elem_id_or_set, ',');
 
+      // Strip any leading or trailing trailing whitespace from this
+      // string, since some Abaqus files may have this.
+      elem_id_or_set.erase
+        (std::remove_if(elem_id_or_set.begin(),
+                        elem_id_or_set.end(), isspace),
+         elem_id_or_set.end());
+
       // Read the character "S", followed by the side id. Note: the >> operator
       // eats whitespace until it reaches a valid character, so this should work
       // whether or not there is a space after the previous comma.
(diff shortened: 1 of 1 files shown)


yw5aj:
Thanks a lot @jwpeterson ! A quick test showed that latest code failed but I'm not sure whether it's me doing something wrong during testing. I'll look into it later tonight and get back to you!

src/mesh/abaqus_io.C
```diff
@@ -849,24 +849,43 @@ void AbaqusIO::read_sideset(std::string sideset_name, sideset_container_t & cont
...
+        (std::strtol(elem_id_or_set.c_str(), &endptr, /*base=*/10));
+
+      if (elem_id != 0 || endptr != elem_id_or_set.c_str())
+        {
+          // if the side set is of the form of "391, S2"
+          id_storage.push_back( std::make_pair(elem_id, side_id) );
+        }
+      else
+        {
+          // if the side set is of the form of "Elset_1, S3"
```
yw5aj: Hi John - this seems unable to handle the whitespaces before and after the elset name in "   Elset_1  , S3", which ABAQUS do allow. Maybe switch to `>>`?

jwpeterson: Thanks for catching this. We have definitely seen this before and in that case simply removing the whitespace from the string after reading it in seemed to work. I pushed a commit which does this, hopefully it fixes your remaining tests.

yw5aj: Thanks so much again @jwpeterson ! This works great.

yw5aj:
Just in case you might find it helpful - I also attached my test files.

[testcube.zip](https://github.com/libMesh/libmesh/files/3795921/testcube.zip)

PR #2318, opened 2019-11-02
AbaqusIO: Add helper functions

jwpeterson:
Motivated by the changes in #2317, these just make it a bit easier to do some common string manipulations in the Abaqus reader.


yw5aj:
This would be super helpful. Thanks! :)

PR #2319, opened 2019-11-05
ignore_warnings.h before tetgen.h

roystgnr:
Otherwise I can't use --enable-werror --enable-paranoid-warnings without
hitting -Werror=switch-default

Oddly I'm not sure what's changed here; I caught this with a gcc 8 build; it's not a new gcc 9 warning.


jwpeterson:
> I'm not sure what's changed here

We updated to Tetgen 1.5.1 on October 23rd. Have you made any `--disable-strict-lgpl --enable-werror --enable-paranoid-warnings` builds since then?

roystgnr:
Ha!  No, this is almost certainly the first.  That'd explain it.

PR #2320, opened 2019-11-05
Move non-libMesh-specific Parallel:: code to a subpackage

roystgnr:
Once this works, we'll put that subpackage into a separate repo, and the upcoming papers from me and @friedmud can tell people "write better message-passing code using this message-passing library", which should elicit fewer dirty looks than "write better message-passing code using this finite element library".

I'm quite happy with the library name PassMess ("It's for message passing!"  "You're moving past messy hand-coded MPI!"  "We're passing a mess of C++ code along to you!"  Choose your own level of irony!) but it's easy to search-and-replace if anyone has other suggestions.

This works for me but I'll tack on a ton of the optional CI tests before we consider merging; I fear won't be surprised at all if e.g. the separate MPI configuration breaks on one system or another.

This currently disables the scoped PerfLog entries in parallel communication functions.  I'd like to restore that eventually, even if it is a bit dubious (we throw away the data on all processors except rank 0...), but none of my ideas for how to do so ("move PerfLog into the subpackage", "play tricky macro games with careful header ordering to insert logging into inlined functions", "add an optional global callback function for use when logging is enabled") seems very elegant.


friedmud:
Not a huge fan of calling any code a "mess" :-)

How about:
MPI++
ObjectMPI
MPII (MPI Interface)
CPPMPI

roystgnr:
Personally I've never seen any sufficiently large code I *wouldn't* call a mess, but I admit that's not the most PR-friendly thing to put in a name.  People still use GIMP though!

MPI++ seems too presumptuous, MPII too prone to Google-typo-correction.  CPPMPI I don't like but it's acceptable if you hate PassMess.

ObjectMPI isn't perfect (I'm more proud of the generic programming than of the OOP here) but I like it almost as much as PassMess which I assume means everyone else will like it much better.  ;-)  Let's make that Plan A unless anyone comes up with a better name.

roystgnr:
I think the only other thing I don't like about ObjectMPI is the syllable count.  5, 3 of which are letters, vs 2, both words, seems clumsier.  Still Plan A but there's got to be something catchier we could pick.

friedmud:
TMPI (Templated MPI)
MMPI (Metaprogramming MPI)
MetaMPI
FWMPI (Freakin' Works MPI)
Pigeon (like the "carrier" kind... they carry messages :-)
?

friedmud:
BTW: "MetaMPI" we could pronounce as "me-tam-pee" just to be annoying :-)

roystgnr:
Okay, @jwpeterson has come up with my new favorite suggestion: "TIMPI", "Templated Interface to MPI".  It's a Tagalog word for "temperance"/"moderation"/"self-control" so there's a zillion hits, but "timpi software" and "timpi mpi" don't show anything relevant, and "timpi c++" has zero hits (until this page gets indexed presumably).  It's got the gravitas of an acronym but it's still two syllables and they're more euphonic to boot.

roystgnr:
Kinda liked MetaMPI (although here it would have the "one level up" meaning rather than the metaprogramming meaning in MetaPhysicL), but there's already software by that name.  TMPI likewise.  FWMPI is kind of like BFR was; it just puts the PR-unfriendliness back one level.  At least SpaceX had a better fig leaf with "Falcon".  MMPI is a famous psych thing, but we could use it here.  Pigeon is a software thing but nothing too close so it's not bad.  I just like TIMPI much better than either.

roystgnr:
Looks like this is going to pass, in which case it's just time to clean up the branch.  I'm going to leave in as many intermediate working states as I can but I'll try to push some of those fixes further back up the branch history.

And I'll search+replace the name.  At least in a single new commit, though I'll do the whole branch if I can get [this cool gsr-branch thing](https://github.com/da-x/git-search-replace) to work.  TIMPI okay with everyone?

roystgnr:
Oh, never mind on the whole-branch-edit idea.  If I do a search-and-replace on file contents without actually moving files to matching directories then it'll just break all the intermediate builds.

roystgnr:
Going once, going twice?  John was happy with this (although wise enough to say so over IM rather than putting his official imprimatur on via github review) so I'm going to hit the button tonight if there are no more concerns/questions/objections.

jwpeterson:
@roystgnr Is this the cleaned up version of the branch? I was going to look at it closer after it was cleaned up but it still said "31 hidden items" in the list of commits (there must be nearly 100?) Is that what you plan to merge?

roystgnr:
Yeah, I know it's huge, but this is it.  I merged bugfixes in with the commits that introduced the bugs wherever I could, but despite the clean CI I want to keep the transition broken into as many steps as possible in case something went wrong for other users.

One thing I was on the fence about - I left in intermediate re-bootstrap commits, to make automatic bisecting easier, but that may be more paranoia than it's worth; I'll get rid of those if you like.

lindsayad:
Are we going to incorporate this as a git submodule or just have it copied intro contrib?

roystgnr:
Submodule.

I've got one last round of futzing with autotools first, and if you guys are planning on doing another MOOSE libMesh update to get those recent optimizations then I'd say do it *before* the submodule gets broken out, but I'm hoping to break out the submodule shortly.

lindsayad:
> Submodule.

Awesome.

> you guys are planning on doing another MOOSE libMesh update

idaholab/moose#14543 will be merged today

PR #2321, opened 2019-11-06
Adding "tellp" to ostream_proxy class

permcody:
We'd like to be able to use libMesh::out (Moose::out) in our TimedPrint
class. For formatting purposes we need this method.

### DIFF
--- a/include/utils/ostream_proxy.h
+++ b/include/utils/ostream_proxy.h
@@ -222,6 +222,11 @@ public:
     return _target;
   }
 
+  /**
+   * Returns the position of the character in the current stream.
+   */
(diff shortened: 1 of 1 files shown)


permcody:
BTW - I do realize that there is a .get() method on the proxy class but since our TimedPrint class is templated we'd still have to specialize it for this one-off class.

roystgnr:
Seems reasonable to me.  Those proxy classes are *supposed* to be drop-in replacements, I was just too lazy to add shims for every single proxyable feature whether we used it or not.  Er, I mean, too concerned about test coverage.  Too concerned about test coverage.

Issue #1307, opened 2017-04-07
Let's eventually dump cppunit?


jwpeterson:
CPPUnit:
* Is still distributed from sourceforge.net ☹️  
* Uses deprecated C++ features (e.g. `std::auto_ptr`) internally which we have to ignore warnings about.
* Can't run individual tests by name, in random order, or which match a regular expression easily. Can't turn on more verbose output for each test that passes/fails (You can do these by making changes to `main()`, but it would be nice if they worked out of the box.)
* Has some weird issue in MOOSE where exceptions thrown during tests somehow cannot be caught in cppunit.
* Has some weird issue in MOOSE where we can't run unit tests in dbg mode (although this works in libmesh somehow).

[googletest](https://github.com/google/googletest) seems like it would be a better choice. The downside is of course refactoring all the existing tests into a new format, and making it all work with the automake build system.

roystgnr:
The dbg mode thing works for me with some cppunit installs and not with others; I guess at some point they converted a header-only-API-using-containers into a library-using-containers and thereby got hit by the GLIBCXX_DEBUG binary incompatibility problems?

I'd say let's not immediately refactor, let's augment.  If someone jumps over the first hurdle of putting in a googletest subdirectory with a skeleton and a couple tests, then the barrier to entry for converting the others, a few at a time, is much reduced.  If we imagine ourselves converting all the hundreds of unit tests at once then it'll never happen.

jwpeterson:
> If we imagine ourselves 

You're right, I was definitely imagining someone else doing it.

> converting all the hundreds of unit tests at once then it'll never happen.

Good point.  I was kind of hoping the test bodies would transfer over basically as-is, won't know until we try it.

Issue #1008, opened 2016-06-22
CPPUNIT_ASSERT_FOO aren't MPI-aware


roystgnr:
And it's a little confusing when our unit_tests report
```
OK (419 tests)
```
Shortly followed by
```
===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   PID 68343 RUNNING AT zappa.ices.utexas.edu
=   EXIT CODE: 1
```

We should make a wrapper macro that handles the "something succeeded on processor 0 but failed elsewhere" case.

jwpeterson:
Something to keep in mind for #1307 as well.

PR #2322, opened 2019-11-07
Use ReplicatedMesh in new edgeset unit test

roystgnr:
Otherwise we get a failure from --enable-parmesh builds, because
something about either this test (added in #2298) or the underlying
code isn't liking DistributedMesh.

This is a short-term workaround so that I don't again run into what
turned out to be an astonishingly hard bug to track down.

As a medium-term fix I'm hoping @jwpeterson can figure out that "isn't
liking DistributedMesh" problem and then we can revert this.

As a long-term improvement, see #1307.  We need something better than
CppUnit for parallel tests, to avoid future "astonishingly hard bug to
track down" issues.  This was another #1008 case, and because I was
working in a branch I had to bisect with --re just to determine that it
really was a new test failing rather than a bug in my branch.

### DIFF
--- a/tests/mesh/write_edgeset_data.C
+++ b/tests/mesh/write_edgeset_data.C
@@ -1,7 +1,7 @@
 // Basic include files
 #include "libmesh/equation_systems.h"
 #include "libmesh/exodusII_io.h"
-#include "libmesh/mesh.h"
+#include "libmesh/replicated_mesh.h"
 #include "libmesh/mesh_generation.h"
 #include "libmesh/parallel.h" // set_union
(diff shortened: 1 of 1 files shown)


jwpeterson:
Posting your comments from IMs before I forget:

> Here it is
> Or rather here's what it isn't:
```
// If we have a distributed mesh, write_sideset_data wants our
// ghost data too; we'll just serialize everything here.
if (!mesh.is_serial())
TestCommWorld->set_union(vals);
```
> Doesn't look like there's any analogous operation in write_edgeset_data.C

PR #2323, opened 2019-11-07
Move forward declaration into its proper namespace

roystgnr:
Not sure why most compilers weren't screaming at us about this long ago,
but @vikramvgarg has a branch which got his compiler to error out when
it couldn't resolve the ambiguity.

### DIFF
--- a/include/error_estimation/adjoint_refinement_estimator.h
+++ b/include/error_estimation/adjoint_refinement_estimator.h
@@ -29,13 +29,13 @@
 #include <cstddef>
 #include <vector>
 
-// Forward declarations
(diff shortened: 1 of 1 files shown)


jwpeterson:
Offending line of code appears to be from 2017, so it has survived multiple compiler generations without complaints. I'm not an expert on compiler name resolution rules, but I could see an argument for that line of code actually being conforming (though admittedly confusing to a reader of the code). Theoretically the compiler could put off resolving the name until it came across the first use, which would be inside a `namespace libMesh`. Anyway, good to keep everything consistent.

PR #2327, opened 2019-11-12
Sparsity pattern optimization for "spider" nodes

jwpeterson:
We have often noticed that building the sparsity pattern for meshes with "spider" nodes (a single node which is connected to hundreds or thousands of 1D elements) when Constraints are active on those nodes seems to take longer than it should, and quickly becomes intractable as the number of connected Elems increases. For example, in a relatively small mesh (n_nodes()=4735, n_elem()=3359), `build_sparsity()` took almost 10 seconds:
```
 --------------------------------------------------------------------------------------------------------------------------------
| libMesh Performance: Alive time=17.3783, Active time=17.3716                                                                   |
 --------------------------------------------------------------------------------------------------------------------------------
| Event                                             nCalls     Total Time  Avg Time    Total Time  Avg Time    % of Active Time  |
|                                                              w/o Sub     w/o Sub     With Sub    With Sub    w/o S    With S   |
|--------------------------------------------------------------------------------------------------------------------------------|
|                                                                                                                                |
| DofMap                                                                                                                         |
|   add_neighbors_to_send_list()                    2          0.0009      0.000459    0.0100      0.005004    0.01     0.06     |
|   build_constraint_matrix_and_vector()            3765       0.3753      0.000100    0.3753      0.000100    2.16     2.16     |
|   build_sparsity()                                1          9.3444      9.344351    9.3562      9.356176    53.79    53.86    |
```

After digging into it, we realized that much of the time was being spent doing repeated in-place merge sorts of the same sets of dofs, so that the end result was you would do a bunch of work but it would have no effect on the resulting sparsity. We found that hashing "large" sets of dofs and skipping sorting them on subsequent encounters sped things up quite a bit:
```
 --------------------------------------------------------------------------------------------------------------------------------
| libMesh Performance: Alive time=8.2667, Active time=8.25377                                                                    |
 --------------------------------------------------------------------------------------------------------------------------------
| Event                                             nCalls     Total Time  Avg Time    Total Time  Avg Time    % of Active Time  |
|                                                              w/o Sub     w/o Sub     With Sub    With Sub    w/o S    With S   |
|--------------------------------------------------------------------------------------------------------------------------------|
|                                                                                                                                |
| DofMap                                                                                                                         |
|   add_neighbors_to_send_list()                    6          0.0019      0.000325    0.0239      0.003989    0.02     0.29     |
|   build_constraint_matrix_and_vector()            4354       0.2548      0.000059    0.2548      0.000059    3.09     3.09     |
|   build_sparsity()                                1          0.7280      0.728025    0.7351      0.735086    8.82     8.91     |
```
Obviously, the time spent computing hashes is wasted unless there are lots of duplicate sets of dofs. In our testing we did find that hashing all the time was more expensive on problems without spider nodes, but not prohibitively so relative to the time it saved on problems with them. At the moment, the optimization is currently only enabled if there over 256 connected dofs for a given Elem, but this is a fairly arbitrarily-selected value. We could also consider making the optimization disableable at configure time if necessary.

### DIFF
--- a/src/base/sparsity_pattern.C
+++ b/src/base/sparsity_pattern.C
@@ -65,6 +65,7 @@ Build::Build (Build & other, Threads::split) :
   coupling_functors(other.coupling_functors),
   implicit_neighbor_dofs(other.implicit_neighbor_dofs),
   need_full_sparsity_pattern(other.need_full_sparsity_pattern),
+  hashed_dof_sets(other.hashed_dof_sets),
   sparsity_pattern(),
   nonlocal_pattern(),
   n_nz(),
@@ -455,6 +456,10 @@ void Build::join (const SparsityPattern::Build & other)
           my_row.erase(std::unique (my_row.begin(), my_row.end()), my_row.end());
         }
     }
+
+  // Combine the other thread's hashed_dof_sets with ours.
+  hashed_dof_sets.insert(other.hashed_dof_sets.begin(),
+                         other.hashed_dof_sets.end());
 }
 
 
--- a/src/base/sparsity_pattern.C
+++ b/src/base/sparsity_pattern.C
@@ -26,7 +26,7 @@
 #include "libmesh/communicator.h"
 #include "libmesh/parallel_algebra.h"
 #include "libmesh/parallel.h"
-
+#include "libmesh/hashword.h"
 
 namespace libMesh
 {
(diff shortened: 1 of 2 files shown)


src/base/sparsity_pattern.C
```diff
@@ -112,10 +112,28 @@ void Build::handle_vi_vj(const std::vector<dof_id_type> & element_dofs_i,
...
+  // number larger will disable the hashing optimization in more
+  // cases.
+  bool dofs_seen = false;
+  if (n_dofs_on_element_j > 0 && n_dofs_on_element_i > 256)
+    {
+      auto hash_i = Utility::hashword(element_dofs_i);
+      auto hash_j = Utility::hashword(element_dofs_j);
+      auto final_hash = Utility::hashword2(hash_i, hash_j);
+      auto result = hashed_dof_sets.insert(final_hash);
+      // if insert failed, we have already seen these dofs
```
roystgnr: ... or we've seen something that hashes to the same value.  I'm sure that only happens one in 2^64 times or whatever but it still scares me a little.  What happens if there **is** a hash collision here?

jwpeterson: You're right, if there's a hash collision, then we would skip setting some sparsity entries that we would otherwise need. If we store more information, we might be able to make a more thorough check in `DEBUG` mode? Since performance is of the essence here, I was hoping to avoid anything too heavy...

roystgnr: Hmmm... I was originally trying to avoid a DEBUG-mode-only check, but I guess if we have an inadequate sparsity pattern then PETSc will scream at us and die even in opt mode?  That's fine then.  I just don't want any possibility of *wrong* results.

roystgnr: (And yes, I'm aware that cosmic rays strike more often than one in N_elem/2^64 runs.  But that's the hardware guys' problem.)

roystgnr:
I'm still skeptical that this is the right solution to your problem, but I happened to be running benchmarks just before the merge, and the build_sparsity() cost was literally unchanged by this PR, so I'm no longer too fearful that unnecessary hashing going to slow down most users.

jwpeterson:
> the build_sparsity() cost was literally unchanged by this PR

Most likely there were no hashes computed, so just an extra `vector::size()` check being done, which must be unmeasurable amidst all the other stuff that is going on.

roystgnr:
Ah, of course; I didn't have enough DoFs per element.  Let me crank that up and see if I can notice a difference.

roystgnr:
Nah, even when I crank up to p=6 hexes (because you need lots of DoFs per element *per variable* even) the build_sparsity() time is still trivial.

jwpeterson:
OK, thanks for checking on this @roystgnr!

PR #2335, opened 2019-11-15
Threads fixes

roystgnr:
We appear to be woefully undertesting multithreaded capabilities in CI.


roystgnr:
Merging ASAP; two of those commits are pretty serious bug fixes.

Issue #2067, opened 2019-03-12
Nemesis Struggles With Large Files


friedmud:
Nemesis is getting quite stuck on a PetscVector::localize() (I thought we removed that years ago?!?):

```
#11 0x00002aaaaed4be69 in PMPI_Allreduce () from /home/gastdr/projects/lemhi_nomarch/gcc_7.3_mvapich_2.3/lib/libmpi.so.12
#12 0x00002aaaad56dccb in PetscSplitOwnership (comm=1140850688, n=n@entry=0x173c59b8, N=N@entry=0x173c59c0) at /home/gastdr/projects/lemhi_nomarch/petsc/src/sys/utils/psplit.c:80
#13 0x00002aaaad63bdc0 in PetscLayoutSetUp (map=0x173c59b0) at /home/gastdr/projects/lemhi_nomarch/petsc/src/vec/is/utils/pmap.c:172
#14 0x00002aaaad651234 in ISGeneralSetIndices_General (is=0x16e1c2b0, n=346, idx=0x16e18100, mode=PETSC_USE_POINTER) at /home/gastdr/projects/lemhi_nomarch/petsc/src/vec/is/is/impls/general/general.c:698
#15 0x00002aaaad651633 in ISGeneralSetIndices (is=0x16e1c2b0, n=n@entry=346, idx=idx@entry=0x16e18100, mode=mode@entry=PETSC_USE_POINTER) at /home/gastdr/projects/lemhi_nomarch/petsc/src/vec/is/is/impls/general/general.c:684
#16 0x00002aaaad651726 in ISCreateGeneral (comm=<optimized out>, n=346, idx=0x16e18100, mode=PETSC_USE_POINTER, is=0x7fffffffb038) at /home/gastdr/projects/lemhi_nomarch/petsc/src/vec/is/is/impls/general/general.c:657
#17 0x00002aaaacb188df in libMesh::PetscVector<double>::localize(std::vector<double, std::allocator<double> >&, std::vector<unsigned long, std::allocator<unsigned long> > const&) const () from /home/gastdr/projects/lemhi_nomarch/libmesh/install/lib/libmesh_opt.so.0
[...]

PR #2338, opened 2019-11-18
Nemesis localize fix

roystgnr:
This isn't a complete solution for #2067 (it fixes NemesisIO::write_element_data, which is relatively self-contained; MeshOutput::write_equation_systems is going to need a similar fix too), but it's enough to throw at CI and make sure I didn't break anything before proceeding to the more complicated half of the fix, and it's hopefully enough that @friedmud can measure *some* speedup.

### DIFF
--- a/include/mesh/mesh_output.h
+++ b/include/mesh/mesh_output.h
@@ -117,7 +117,7 @@ public:
   { libmesh_not_implemented(); }
 
   /**
-   * This method should be overridden by "parallel" output formats for
+   * This method may be overridden by "parallel" output formats for
    * writing nodal data.  Instead of getting a localized copy of the
    * nodal solution vector, it is passed a NumericVector of
    * type=PARALLEL which is in node-major order i.e.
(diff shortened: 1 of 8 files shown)


roystgnr:
Ha!  I didn't think we had any discontinuous Nemesis output test coverage within libMesh itself; I forgot that it's one of the code paths plot_error can take.

roystgnr:
Going to add on the more complicated half of the fix here.  If CI reveals that it was *too* much more complicated then we could always backtrack and just merge at a3068c3 in the worst case.

roystgnr:
Okay, *now* this might be a complete solution to #2067.  For Nemesis at least.  Anyone using .pvtk output on large problems might want to consider updating that to use the newest APIs too.

roystgnr:
Planning to merge today unless anyone screams; I'll have other Nemesis changes in the future that I want to build on top of these.

permcody:
Oops - I guess we don't have all the same error flags on for MOOSE and libMesh:
https://civet.inl.gov/job/439283/

PR #2340, opened 2019-11-19
Fix a bug on allocating segmentlist in Triangle interface

### LINKED ISSUE #2339: A bug in Triangle interface
If both segments and holes are given, currently the interface does not allocate the segment list in a proper size. The offending line is at https://github.com/libMesh/libmesh/blob/master/src/mesh/mesh_triangle_interface.C#L138. I will send in a quick fix.

YaqiWang:
Close #2339.

I cannot find a unit test for the interface so I did not try to add any test. But I did verify that this fixes the problem.

### DIFF
--- a/src/mesh/mesh_triangle_interface.C
+++ b/src/mesh/mesh_triangle_interface.C
@@ -108,11 +108,21 @@ void TriangleInterface::triangulate()
   // If the holes vector is non-nullptr (and non-empty) we need to determine
   // the number of additional points which the holes will add to the
   // triangulation.
+  // Note that the number of points is always equal to the number of segments
+  // that form the holes.
   unsigned int n_hole_points = 0;
 
   if (have_holes)
     for (const auto & hole : *_holes)
+    {
       n_hole_points += hole->n_points();
+      // A hole at least has one enclosure.
+      // Points on enclosures are ordered so that we can add segments implicitly.
+      // Elements in segment_indices() indicates the starting points of all enclosures.
+      // The last element in segment_indices() is the number of total points.
+      libmesh_assert_greater(hole->segment_indices().size(), 1);
+      libmesh_assert_equal_to(hole->segment_indices().back(), hole->n_points());
+    }
 
   // Triangle data structure for the mesh
   TriangleWrapper::triangulateio initial;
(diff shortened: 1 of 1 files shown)


src/mesh/mesh_triangle_interface.C
```diff
@@ -106,13 +106,18 @@ void TriangleInterface::triangulate()
...
   unsigned int n_hole_points = 0;
+  unsigned int num_hole_segments = 0;
 
   if (have_holes)
     for (const auto & hole : *_holes)
+    {
       n_hole_points += hole->n_points();
+      for (unsigned int i=0; i<hole->segment_indices().size()-1; ++i)
+        num_hole_segments += hole->segment_indices()[i+1] - hole->segment_indices()[i];
+    }
```
jwpeterson: It appears that `segment_indices()` function always returns a vector with exactly 2 entries? And even if it didn't this loop would add/subtract the intermediate values anyway so you would only end up with the difference between the first and last entry. Not saying this is wrong, just trying to understand what's going on.

YaqiWang: Yes, the size of `segment_indices()` is `n`+1, where `n` is the number of segments described by nodes on the segments.

YaqiWang: For example of a hole created by two concentric circles in the ring. In such a case, `n` will be 2. The entries in `segment_indices()` will be 3. The nodes are ordered properly so that we can have implicit rule on adding segments with just these three numbers.

YaqiWang: I will add an assertion to make sure we have a peaceful mind ;-)

jwpeterson: What I'm saying is, if segment_indices = [a,b,c,d,..., z] then the result of your calculation above is simply `num_hole_segments` = z-a because all the intermediate terms cancel. So there's no reason to do this loop?

YaqiWang: Yes, I think we just need to fix Line 138.

YaqiWang:
@jwpeterson You are right. The fix should be just one line. I left the two assertions since that can possibly increase the readability of the code and assure the correctness of hole generation.

PR #2341, opened 2019-11-19
Add get_row method to Sparse Matrix

lindsayad:
Currently only implementation is in PetscMatrix because that's what I know...

### DIFF
--- a/include/numerics/sparse_matrix.h
+++ b/include/numerics/sparse_matrix.h
@@ -406,6 +406,20 @@ public:
    */
   virtual void get_transpose (SparseMatrix<T> & dest) const = 0;
 
+  /**
+   * Get a row from the matrix
+   * @param i The matrix row to get
+   * @param indices A container that will be filled with the column indices
+   *                corresponding to (possibly) non-zero values
+   * @param values A container holding the column values
+   */
+  virtual void get_row(numeric_index_type /*i*/,
+                       std::vector<numeric_index_type> & /*indices*/,
+                       std::vector<T> & /*values*/) const
+  {
+    libmesh_not_implemented();
+  }
+
 protected:
 
   /**
--- a/src/numerics/petsc_matrix.C
+++ b/src/numerics/petsc_matrix.C
@@ -1278,6 +1278,23 @@ void PetscMatrix<T>::get_row (numeric_index_type i_in,
   // to run on one processor.
   libmesh_assert(this->closed());
 
+  // PETSc makes no effort at being thread safe. Helgrind complains about
+  // possible data races even just in PetscFunctionBegin (due to things
+  // like stack counter incrementing). Perhaps we could ignore
+  // this, but there are legitimate data races for Mat data members like
+  // mat->getrowactive between MatGetRow and MatRestoreRow. Moreover,
+  // there could be a write into mat->rowvalues during MatGetRow from
+  // one thread while we are attempting to read from mat->rowvalues
+  // (through petsc_cols) during data copy in another thread. So
+  // the safe thing to do is to lock the whole method
+
+#ifdef LIBMESH_HAVE_CXX11_THREAD
+  std::lock_guard<std::mutex>
+#else
+  Threads::spin_mutex::scoped_lock
+#endif
+    lock(_petsc_matrix_mutex);
+
   ierr = MatGetRow(_mat, i_val, &ncols, &petsc_cols, &petsc_row);
   LIBMESH_CHKERR(ierr);
 
--- a/src/numerics/petsc_matrix.C
+++ b/src/numerics/petsc_matrix.C
@@ -1256,7 +1256,45 @@ T PetscMatrix<T>::operator () (const numeric_index_type i_in,
   return value;
 }
 
+template <typename T>
+void PetscMatrix<T>::get_row (numeric_index_type i_in,
+                              std::vector<numeric_index_type> & indices,
+                              std::vector<T> & values) const
+{
+  libmesh_assert (this->initialized());
 
+  // PETSc 2.2.1 & newer
+  const PetscScalar * petsc_row;
+  const PetscInt    * petsc_cols;
+
+  PetscErrorCode ierr=0;
+  PetscInt
+    ncols=0,
+    i_val = static_cast<PetscInt>(i_in);
+
+  // the matrix needs to be closed for this to work
+  // this->close();
+  // but closing it is a semiparallel operation; we want operator()
+  // to run on one processor.
+  libmesh_assert(this->closed());
+
+  ierr = MatGetRow(_mat, i_val, &ncols, &petsc_cols, &petsc_row);
+  LIBMESH_CHKERR(ierr);
+
+  // Copy the data
+  indices.resize(static_cast<std::size_t>(ncols));
+  values.resize(static_cast<std::size_t>(ncols));
+
+  for (std::size_t i = 0; i < indices.size(); ++i)
+  {
+    indices[i] = static_cast<numeric_index_type>(petsc_cols[i]);
+    values[i] = static_cast<T>(petsc_row[i]);
+  }
+
+  ierr  = MatRestoreRow(_mat, i_val,
+                        &ncols, &petsc_cols, &petsc_row);
+  LIBMESH_CHKERR(ierr);
+}
 
 
 template <typename T>
--- a/tests/numerics/petsc_matrix_test.C
+++ b/tests/numerics/petsc_matrix_test.C
@@ -9,8 +9,10 @@
 #include "test_comm.h"
 
 #include <vector>
+#ifdef LIBMESH_HAVE_CXX11_THREAD
 #include <thread>
 #include <algorithm>
+#endif
 
 using namespace libMesh;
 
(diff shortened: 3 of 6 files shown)


include/numerics/sparse_matrix.h
```diff
@@ -406,6 +406,17 @@ class SparseMatrix : public ReferenceCountedObject<SparseMatrix<T>>,
...
+  /**
+   * Get a row from the matrix
+   * @param i The matrix row to get
+   * @param indices A container that will be filled with the column indices
+   *                corresponding to (possibly) non-zero values
+   * @param values A container holding the column values
+   */
+  virtual void get_row(numeric_index_type /*i*/,
+                       std::vector<numeric_index_type> & /*indices*/,
+                       std::vector<T> & /*values*/) const {}
```
roystgnr: Stick a libmesh_not_implemented() in the base class to make sure anyone accidentally using it gets an error rather than silently-mistaken values?

src/numerics/petsc_matrix.C
```diff
@@ -1256,7 +1256,45 @@ T PetscMatrix<T>::operator () (const numeric_index_type i_in,
...
+  const PetscScalar * petsc_row;
+  const PetscInt    * petsc_cols;
+
+  PetscErrorCode ierr=0;
+  PetscInt
+    ncols=0,
+    i_val = static_cast<PetscInt>(i_in);
+
+  // the matrix needs to be closed for this to work
+  // this->close();
```
fdkong: You are right. `close matrix` is not a job of` get row`

tests/numerics/petsc_matrix_test.C
```diff
@@ -0,0 +1,93 @@
...
+
+  Parallel::Communicator * _comm;
+  std::unique_ptr<PetscMatrix<Number>> _matrix;
+  numeric_index_type _local_size, _global_size;
+  std::vector<numeric_index_type> _i;
+  const Real _tolerance = TOLERANCE * TOLERANCE;
+
+};
+
+CPPUNIT_TEST_SUITE_REGISTRATION(PetscMatrixTest);
```
fdkong: I did not see do any preallocation for the matrix. `preallocation` is critical, and we do not want users think that no preallocation is just fine.

You might do that in other places already?

jwpeterson:
This reminds me of all the multithreading issues we ran into with `PetscVector<T>::_get_array` since it has that same `Get`/`Restore` flavor. What are the implications if multiple threads call `get_row()`?

tests/numerics/petsc_matrix_test.C
lindsayad: When we call `PetscMatrix::init` we allocate the non-zero structure

src/numerics/petsc_matrix.C
```diff
@@ -1256,7 +1256,45 @@ T PetscMatrix<T>::operator () (const numeric_index_type i_in,
...
 }
 
+template <typename T>
+void PetscMatrix<T>::get_row (numeric_index_type i_in,
+                              std::vector<numeric_index_type> & indices,
+                              std::vector<T> & values) const
+{
+  libmesh_assert (this->initialized());
+
+  // PETSc 2.2.1 & newer
```
jwpeterson: I'm guessing this comment is copy-pasted and you didn't really test with PETSc 2.2.1 ;-)

lindsayad:
> This reminds me of all the multithreading issues we ran into with `PetscVector<T>::_get_array` since it has that same `Get`/`Restore` flavor. What are the implications if multiple threads call `get_row()`?

I am definitely not a threads expert, and most of my knowledge came within my last hour of reading. From a black-box/outside perspective I would say that this should indeed have thread protection.

However, I'll start by saying that `PetscMatrix::operator()` has the same design pattern as that here, e.g. no thread protection and use of `MatGetRow` and `MatRestoreRow`, and it seems like that's been ok to-date. But I'm wondering if that's a bit of luck. I've looked at the 3.12 version of PETSc `SeqAIJ`, `MPIAIJ`, `SeqBAIJ` and `MPIBAIJ`. All but `SeqAIJ` have the potential to malloc when `MatGetRow` is called. Only `SeqBAIJ` frees when `MatRestoreRow` is called, which I have to say I don't really understand. But from that investigation it would seem like only `SeqBAIJ` has the potential to be thread-unsafe, and I'm guessing we don't really have any `SeqBAIJ` users. 

So in summary, I guess I should add some safe-guards to cover our bases.

lindsayad:
@jwpeterson What is purpose of `std::atomic_thread_fence` in `PetscVector::_get_array` and `PetscVector::_restore_array`?

jwpeterson:
Oof, I don't remember all the details, but it's part of the implementation of a doubly-checked lock. You can read about some of it over on #853. 

BTW, I don't recommend you try and shoot for absolute thread safety in this PR, that is impossible to do without a semi-reliably failing test case anyway... but you might want to put a comment in the new API that it could be a potential source of issues for multithreading. It depends on how you are planning to use `get_row()`... if you never use it in a threaded region then you should have no issues :-)

lindsayad:
@jwpeterson Look at the most recent commit. I think I decided it's best to lock the whole method...

jwpeterson:
> I think I decided it's best to lock the whole method

OK, I agree that is probably the safest thing to do. Hopefully the locking isn't a performance issue for you.

lindsayad:
I wonder how many thought hours we've spent (wasted?) on threading. How many of our users actually use threading? It seems like a pretty uphill battle when one of the primary libraries we use makes no attempt to be thread-safe (or is rather openly fine with being thead-unsafe). For example I'm guessing @friedmud spent at least 40 hours researching/developing #841, #853, and #856.

Don't get me wrong; I personally think there can be great value in shared-memory parallel programming and I think concurrency is a fascinating subject, hence why I spent the last day reading about it. I just feel like sometimes we spend a lot of mind-share on something that none of our users use...

roystgnr:
> How many of our users actually use threading?

Not enough, yet, or someone would have screamed about the GenericProjector-with-threading-and-high-p bug or the MeshFunction-threading-race-condition bug fixed in #2335 long ago.  But IIRC the lack of interest in threading was heavily due to PETSc solves not scaling with threads, and if that hasn't been fixed by now surely it'll be fixed eventually.

In the past our efficient alternative to threading has been DistributedMesh, but that's got its own set of more-difficult-to-program-for characteristics, and "efficient alternative" is very debatable, and the parameters of that debate change with the underlying hardware.

lindsayad:
> But IIRC the lack of interest in threading was heavily due to PETSc solves not scaling with threads, and if that hasn't been fixed by now surely it'll be fixed eventually.

I don't think this is true...see PETSc's position [on threads](https://www.mcs.anl.gov/petsc/miscellaneous/petscthreads.html). Upon reading it seems to me that the only efficient way to use threads and PETSc is using the model where each thread manages its own sequential PETSc object. And that only makes sense for multiple small systems, not for a single large system.

fdkong:
> Not enough, yet, or someone would have screamed about the GenericProjector-with-threading-and-high-p bug or the MeshFunction-threading-race-condition bug fixed in #2335 long ago. But IIRC the lack of interest in threading was heavily due to PETSc solves not scaling with threads, and if that hasn't been fixed by now surely it'll be fixed eventually.

The biggest issue is that it is nontrivial to make the  threading algorithm scalable. The threading model may need to support more features.  Furthermore,
there may not be that many cases where `n threads` are actually faster than `n MPI threads`. 
I do not think PETSc developers will add money to do threading things in any near future.

fdkong:
> I wonder how many thought hours we've spent (wasted?) on threading. How many of our users actually use threading? It seems like a pretty uphill battle when one of the primary libraries we use makes no attempt to be thread-safe (or is rather openly fine with being thead-unsafe). For example I'm guessing @friedmud spent at least 40 hours researching/developing #841, #853, and #856.

It is nice to have the threading support, but I personally do not think it is a good way to use my energy because I do not have any cases where I have to employ threading.

Thanks to @jwpeterson  and @roystgnr 's great work,  libMesh is able to be scalable to more than tens of thousands of cores without much effort. Especially the wonderful data structure: `distributed mesh ` enables memory-efficient calculations.

fdkong:
We may need to reevaluate this issue when hardware change in the future. For example, some specific features only can be achieved using threading model.

lindsayad:
> OK, I agree that is probably the safest thing to do. Hopefully the locking isn't a performance issue for you.

Yea I'm not seeing any performance issues when using this in the Contact application

tests/numerics/petsc_matrix_test.C
```diff
@@ -0,0 +1,121 @@
...
+                                  _tolerance);
+          CPPUNIT_ASSERT_EQUAL(cols_to_get[local_count], local_index);
+        }
+      }
+    };
+
+#ifdef LIBMESH_HAVE_CXX11_THREAD
+    auto num_threads = std::min(unsigned(2),
+                                std::max(
+                                  std::thread::hardware_concurrency(),
```
jwpeterson: I will add `std::thread::hardware_concurrency()` to our C++11 thread configure test. It apparently works with all the "min" compiler versions so this can definitely be merged as-is.

include/numerics/sparse_matrix.h
roystgnr: Could you add this before we merge?

lindsayad: I did...

roystgnr: Sorry, I didn't notice that GitHub now just adds a tiny "outdated" tag rather than collapsing/hiding conversations about outdated code entirely.

PR #2343, opened 2019-11-20
Enable PETSc shell matrices for SLEPc eigenvalue solver 

fdkong:


roystgnr:
Gah, sorry about the huge diff.  I re-ran bootstrap after that PassMess->TIMPI search-and-replace, to fix alphabetization issues from there, but I forgot to re-run the rebuild scripts.

src/solvers/petscdmlibmeshimpl.C
```diff
@@ -1203,7 +1203,7 @@ PetscErrorCode  DMCreate_libMesh(DM dm)
   dm->ops->getinjection       = 0; // DMGetInjection_libMesh;
   dm->ops->getaggregates      = 0; // DMGetAggregates_libMesh;
 #else
-  dm->ops->createinjection = 0;
+  //dm->ops->createinjection = 0;
```
roystgnr: What's going on here?

include/numerics/petsc_shell_matrix.h
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * This class allows to use a PETSc shell matrix.
+ * All overridden virtual functions are documented in
+ * shell_matrix.h.
+ *
+ * \author Fande Kong (fdkong.jd@gmail.com)
+ * \date 2019
+ */
+template <typename T>
+class PetscShellMatrix : public ShellMatrix<T>
```
roystgnr: What's the advantage of this over using a SparseShellMatrix with a PetscMatrix?

src/solvers/petscdmlibmeshimpl.C
fdkong: Accidentally  changed this when I was trying different versions of PETSc.

I will change it back .

include/numerics/petsc_shell_matrix.h
fdkong: `PetscMatrix` suppose to work with nonshell matrices, e.g., SeqAIJ, MPIAIJ.

`PetscShellMatrix` is partially attached to shell matrix. Shell matrix does not need a preallocation, but needs to attach constimized `MatVec`.

BTW, I do not completely understand how to use `a SparseShellMatrix with a PetscMatrix` 
to achieve what I have done here. 

It may help you see more if you would like to take a look at a use case here https://github.com/idaholab/moose/pull/14406

I am open to any other options, if you have other thoughts on how to achieve my goal.

roystgnr: > PetscMatrix suppose to work with nonshell matrices, e.g., SeqAIJ, MPIAIJ.

How hard would it be to remove that requirement, though?  I see we have a PetscMatrix constructor that takes an externally-managed Mat object, and I don't see any internally-used methods except init() where it enforces the type of matrix.  If we can replace 300 lines of PetscShellMatrix code with 30 more lines of switch(_mat_type) cases in PetscMatrix::init(), that'd be worth it.

fdkong: Do you mean `PetscMatrix` take a shell mat of PETSc?

The big concern is that almost all the APIs in PetscMatrix will be invalid for` shell matrix`. For example, the following APIs will no sense for shell matrix:


  ```
virtual void set (const numeric_index_type i,
                    const numeric_index_type j,
                    const T value) override;

  virtual void add (const numeric_index_type i,
                    const numeric_index_type j,
                    const T value) override;

  virtual void add_matrix (const DenseMatrix<T> & dm,
                           const std::vector<numeric_index_type> & rows,
                           const std::vector<numeric_index_type> & cols) override;

  virtual void add_matrix (const DenseMatrix<T> & dm,
                           const std::vector<numeric_index_type> & dof_indices) override;

  virtual void add_block_matrix (const DenseMatrix<T> & dm,
                                 const std::vector<numeric_index_type> & brows,
                                 const std::vector<numeric_index_type> & bcols) override;

  virtual void add_block_matrix (const DenseMatrix<T> & dm,
                                 const std::vector<numeric_index_type> & dof_indices) override
  { this->add_block_matrix (dm, dof_indices, dof_indices); }

  /**
   * Compute A += a*X for scalar \p a, matrix \p X.
   *
   * \note The matrices A and X need to have the same nonzero pattern,
   * otherwise \p PETSc will crash!
   *
   * \note It is advisable to not only allocate appropriate memory
   * with \p init(), but also explicitly zero the terms of \p this
   * whenever you add a non-zero value to \p X.
   *
   * \note \p X will be closed, if not already done, before performing
   * any work.
   */
  virtual void add (const T a, const SparseMatrix<T> & X) override;

  virtual T operator () (const numeric_index_type i,
                         const numeric_index_type j) const override;

  virtual Real l1_norm () const override;

  virtual Real linfty_norm () const override;

```

I do not think it will be better than the current implementation.

fdkong: Similarly, we did not merge `ShellMatrix` into `SparseMatrix`

roystgnr: Yeah, good point.

In hindsight perhaps we should have made something like ShellMatrix into a SparseMatrix superclass, though, rather than relying on overload after overload for the cases where we only need the former's capabilities but we want to handle the latter too.

src/solvers/petscdmlibmeshimpl.C
roystgnr: :+1: IMHO after this change is reverted.

roystgnr: Ping?

fdkong:
@roystgnr ?

roystgnr:
Sorry, didn't get a notification on that last push.

PR #2347, opened 2019-11-21
Add a missing header inclusion.

drwells:
We use START_LOG below, which is a macro defined in `libmesh_logging.h`.

I cannot compile 1.5.0 without this patch. I am not sure why but I *can* compile master without this patch. I don't know what y'all normally do for releases; would it help if I put up a second PR for this onto `v1.5.1_branch`?

Edit: Clarification: I was able to compile master without this patch at about the time 1.5.0 was released. I need this patch to compile master right now (i.e., as of November 21st).

### DIFF
--- a/src/fe/fe_xyz_map.C
+++ b/src/fe/fe_xyz_map.C
@@ -17,6 +17,7 @@
 
 #include "libmesh/fe_xyz_map.h"
 #include "libmesh/elem.h"
+#include "libmesh/libmesh_logging.h"
 
 namespace libMesh
 {
--- a/src/mesh/boundary_info.C
+++ b/src/mesh/boundary_info.C
@@ -22,6 +22,7 @@
 
 // Local includes
 #include "libmesh/libmesh_config.h"
+#include "libmesh/libmesh_logging.h"
 #include "libmesh/boundary_info.h"
 #include "libmesh/distributed_mesh.h"
 #include "libmesh/elem.h"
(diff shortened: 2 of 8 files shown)


roystgnr:
We'll cherry-pick manually; thanks for catching this in time!

drwells:
I updated my local copy of `master` and I got a bunch more compilation errors due to the same missing header. I'll force push an updated fix - I went through and added the header in every file that `LOG_SCOPE` shows up.

jwpeterson:
I'm not sure why we aren't seeing these compilation failures in our CI. Must be a configuration we don't test or something...

drwells:
I ran `configure` in the following way:
```
../configure                                   \
    CCFLAGS="-Wno-implicit-fallthrough"        \
    CXXFLAGS="-Wno-implicit-fallthrough"       \
  --prefix=$HOME/Applications/libmesh-dev.g/   \
  --with-methods=dbg                           \
  --enable-triangle                            \
  --enable-exodus=yes                          \
  --enable-timestamps=no                       \
  --disable-glibcxx-debugging                  \
  --disable-openmp                             \
  --disable-perflog                            \
  --disable-pthreads                           \
  --disable-tbb                                \
  --with-thread-model=none                     \
  --disable-boost                              \
  --disable-reference-counting                 \
  --disable-slepc                              \
  --disable-strict-lgpl                        \
  --disable-vtk                                \
  --disable-eigen                              \
  --with-metis=PETSc                           \
  --disable-hdf5
```

My best guess is that the problem will show up if everything is disabled.

jwpeterson:
Thanks, that is helpful information. Looks like I need to start your PR jobs and add you to the developers list...

drwells:
Sure!

PR #2354, opened 2019-11-27
Sync nodes after all_second_order if not replicated

lindsayad:
Previously we only synced if we were not serial. However,
even if we haven't deleted remote elements in our distributed mesh
we still need to sync the nodes because even if we are adding
nodes on all procs they are getting temporary out-of-sync ids.
If we don't do this sync, then I get failed assertions
in libmesh_assert_valid_parallel_ids. Explicitly, even for the very
first new node, I can fail `(min_proc_id == this->processor_id() && obj)`
because the non-owning proc may have assigned the node
ID 14 while the owning proc assigned it ID 17 or whatever. Then
the owning proc will fail the assertion when `i == 14` because
the LHS will be true while `obj` is `nullptr`

### DIFF
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -1306,7 +1306,7 @@ void UnstructuredMesh::all_second_order (const bool full_ordered)
   // partitioning of new nodes may not be well balanced.
   //
   // make_nodes_parallel_consistent() will fix all this.
-  if (!this->is_serial())
+  if (!this->is_replicated())
     MeshCommunication().make_nodes_parallel_consistent (*this);
 
   // renumber nodes, elements etc


roystgnr:
Good catch, thanks!

roystgnr:
But maybe not a perfect catch.

If we have a DistributedMesh that hasn't yet reached prepare_for_use(), then previously it still returned is_serial() as true and skipped the sync.  Now it goes to the sync, where the final MeshTools::correct_node_proc_ids() step sees unpartitioned elements and screams and dies.

This is what's killing the distributed sweep in #2363 

I haven't yet decided what the "right" fix is.

roystgnr:
Although I have decided what the right test suite would have been: even when a change like this looks obviously correct, if it's a DistributedMesh change we don't run it without a distributed sweep in CI first.

lindsayad:
A temporary bad fix would be to check `is_prepared` :-(

roystgnr:
Hah, that's not *too* far off from what I'm doing.  My check should be much more robust, though.

roystgnr:
Damn, I just realized that if my check *reverts* your fix I'll have no way to know about it.  Any chance you could set up a unit test that would have broken before this PR?

lindsayad:
I'll look into it

PR #2356, opened 2019-12-02
Allow MPI_PROC_NULL in broadcast

### LINKED ISSUE #2353: Allow MPI_NULL_PROC broadcast
Currently, `root_id` in the `Communicator::broadcast` methods is checked to see if its within range of the size of the `Communicator`. I'd like to modify the assertions to allow `MPI_NULL_PROC` so that we can allow broadcasts where we don't need to know the root ahead of time as long as only one proc does the broadcast.

Thoughts?

loganharbour:
Closes #2353 

There's a check for `LIBMESH_HAVE_MPI` hanging around in parallel_unit. I moved it over to `TIMPI_HAVE_MPI`. Should this `#ifdef` just come out?

### DIFF
--- a/contrib/timpi/src/parallel/include/timpi/parallel_implementation.h
+++ b/contrib/timpi/src/parallel/include/timpi/parallel_implementation.h
@@ -1495,15 +1495,15 @@ inline void Communicator::broadcast (bool & data, const unsigned int root_id) co
       return;
     }
 
-  timpi_assert_less (root_id, this->size());
+  timpi_assert (root_id < this->size() ||
+                root_id == (unsigned int)MPI_PROC_NULL ||
+                root_id == (unsigned int)MPI_ROOT);
 
   TIMPI_LOG_SCOPE("broadcast()", "Parallel");
 
   // We don't want to depend on MPI-2 or C++ MPI, so we don't have
   // MPI::BOOL available
   char char_data = data;
-
-  timpi_assert_less(root_id, this->size());
 
   // Spread data to remote processors.
   timpi_call_mpi
@@ -1525,7 +1525,9 @@ inline void Communicator::broadcast (std::basic_string<T> & data,
       return;
     }
 
-  timpi_assert_less (root_id, this->size());
+  timpi_assert (root_id < this->size() ||
+                root_id == (unsigned int)MPI_PROC_NULL ||
+                root_id == (unsigned int)MPI_ROOT);
 
   TIMPI_LOG_SCOPE("broadcast()", "Parallel");
 
@@ -1564,15 +1566,15 @@ inline void Communicator::broadcast (std::vector<T,A> & data,
       return;
     }
 
-  timpi_assert_less (root_id, this->size());
+  timpi_assert (root_id < this->size() ||
+                root_id == (unsigned int)MPI_PROC_NULL ||
+                root_id == (unsigned int)MPI_ROOT);
 
   TIMPI_LOG_SCOPE("broadcast()", "Parallel");
 
   // and get the data from the remote processors.
   // Pass nullptr if our vector is empty.
   T * data_ptr = data.empty() ? nullptr : data.data();
-
-  timpi_assert_less(root_id, this->size());
 
   timpi_call_mpi
     (MPI_Bcast (data_ptr, cast_int<int>(data.size()),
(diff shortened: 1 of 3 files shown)


roystgnr:
> There's a check for LIBMESH_HAVE_MPI hanging around in parallel_unit. I moved it over to TIMPI_HAVE_MPI.

That atavism was deliberate, actually - I'm sure we have user code relying on LIBMESH_HAVE_MPI still and I wanted to keep some test coverage on it.

> Should this #ifdef just come out?

Oh, no.  If it did then we'd get an undefined symbol error from MPI_COMM_TYPE_SHARED and MPI_INFO_NULL whenever someone configures without MPI.

contrib/timpi/test/parallel_unit.C
```diff
@@ -123,15 +123,21 @@ std::vector<std::string> pt_number;
     if (TestCommWorld->rank() == 0)
       dest = src;
 
-    TestCommWorld->broadcast(dest);
+    unsigned int root_id = 0;
+#ifdef LIBMESH_HAVE_MPI
+    if (use_proc_null && TestCommWorld->rank() != 0)
+      root_id = (unsigned int)MPI_PROC_NULL;
```
roystgnr: These MPI_PROC_NULL uses are going to need to be ifdef-protected too.  Or we could add a TIMPI::null_processor_id shim and use that instead, I suppose?  That would be even better if we'd ever use it, but I can't quickly imagine any algorithms that would need to use MPI_PROC_NULL/null_processor_id in parallel but where we wouldn't just want to ifdef out the entire section of algorithm in serial.

loganharbour:
Ahh, I see. Went ahead and switched back to `LIBMESH_HAVE_MPI`. I also wrapped my usage for `MPI_PROC_NULL` around the same `#ifdef`.

I'm done with this now. Works with my use case as well.

contrib/timpi/src/parallel/include/timpi/parallel_implementation.h
```diff
@@ -1495,16 +1495,14 @@ inline void Communicator::broadcast (bool & data, const unsigned int root_id) co
       return;
     }
 
-  timpi_assert_less (root_id, this->size());
+  timpi_assert (root_id < this->size() || root_id == (unsigned int)MPI_PROC_NULL);
```
roystgnr: We shouldn't need the explicit casts in these tests, should we?

roystgnr:
> I also wrapped my usage for MPI_PROC_NULL around the same #ifdef.

So *that's* why my comment didn't submit until I reloaded the page, and why now I see an ifdef there that I could have sworn wasn't there before?  Thought I was losing my mind for a second.

loganharbour:
Ha. I do need to relax with the squashing and force pushing immediately after a comment.

roystgnr:
I've added the No MPI/No PETSc CI tests.  After CI is happy and the explicit casts are gone this looks good to me.

roystgnr:
(so go ahead and force push again when you need to - I won't be in mid-comment again)

loganharbour:
👍 I'm all done here

roystgnr:
The casts, though?  If we need them I don't yet understand why; if we don't then we should remove them.

loganharbour:
Ahh, yes. `MPI_NULL_PROC` is of type `int` so you end up with a comparison warning between `unsigned int` (for `root_id`) and `int`. I just used the casts to get rid of the warnings.

Would you prefer to have the warnings?

roystgnr:
Oh, I'm an idiot.  You're not casting because of the varying processor_id_type sizes, it's a signed/unsigned issue.  No, we definitely don't want the warnings.  I think we have a CI test or two with --enable-werror that would scream at us if we did.

loganharbour:
I just now got the notifications and are now _able_ to view your comments. Makes much more sense now. Silly GitHub.

loganharbour:
Hold off on this. Found an issue. We also need to add `MPI_ROOT` to the assertion. If you're using `MPI_PROC_NULL`, the root must be given the id `MPI_ROOT`, not the actual root proc.

loganharbour:
I need to look into this more--will get around to it when I have a chance.

roystgnr:
> Hold off on this.

Ironically, I just discovered that I need and we don't yet have ```broadcast<vector<vector>>```, so I'm planning a PR that'll stomp right on your toes here.  So no rush, but fair warning: first one ready-to-merge gets to make the other person deal with fixing up merge conflicts.  :-D

loganharbour:
Sounds good. I haven't touched this and won't have time to until next week anyway 😬

PR #2358, opened 2019-12-04
Specify override to avoid clang compiler warning

lindsayad:
I've added the MOOSE Mac Test recipe, which as we know from idaholab/moose#14464 will fail run-time testing, but hopefully with this change we can at least build the framework (see https://civet.inl.gov/job/439270/)


lindsayad:
It only happens with clang...

lindsayad:
Merging because I think it's the MOOSE developers jobs to figure out this try-catch phenomenon in their test suite

PR #2360, opened 2019-12-06
Barebones-build fixes

roystgnr:
This was going to be incidental to testing I've been doing of an upcoming PR for #2179, but it snowballed and it's logically separate and it's much less important, so let's test and merge it separately.


roystgnr:
> If it would be simpler maybe we could make them non-optional?

We wouldn't really get a full win on simplicity unless we also made AMR non-optional, which would have runtime costs too.  I do think I'm about 80% of the way towards "screw it, PBCs/DBCs/AMR compile-time support is mandatory now", though, so if I do get to 100% I'll now expect you to be on board too.  ;-)

PR #2362, opened 2019-12-06
Let make_links_to_me_local handle fake neighbors

roystgnr:
This fixes the test case in https://github.com/idaholab/moose/pull/14456 for me, although as discussed in that ticket there are still a few other parts of libMesh that I'd expect fake neighbor links to break.

### DIFF
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -1024,13 +1024,19 @@ void Elem::make_links_to_me_local(unsigned int n)
     return;
 
   // What side of neigh are we on?  We can't use the usual Elem
-  // method because we're in the middle of restoring topology
+  // method because we're in the middle of restoring topology.  We
+  // can't compare side_ptr nodes because users want to abuse
+  // neighbor_ptr to point to not-technically-neighbors across mesh
+  // slits.  Let's try side centroids.
   const std::unique_ptr<Elem> my_side = this->side_ptr(n);
+  const Point my_centroid = my_side->centroid();
+  const Real my_tol = TOLERANCE * this->hmin();
   unsigned int nn = 0;
   for (; nn != neigh->n_sides(); ++nn)
     {
       const std::unique_ptr<Elem> neigh_side = neigh->side_ptr(nn);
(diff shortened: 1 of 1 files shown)


roystgnr:
Assuming this passes the standard CI I'm going to want to do some DistributedMesh sweeps before we merge.

roystgnr:
@arovinelli, what was the answer to "Wait, are you planning to move these element pairs apart after displacement?"  If it's "yes" then I might as well just cancel this PR; we'll definitely have to just communicate more topology data if we can't use nodal ids *or* nodal coordinates to reconstruct it.

arovinelli:
@roystgnr sorry I didn’t noticed this. Yes the plan is to torn it apart, which again, already works for replicated cases

Sent from my iPhone

On Dec 6, 2019, at 1:13 PM, roystgnr <notifications@github.com> wrote:

﻿

@arovinelli<https://github.com/arovinelli>, what was the answer to "Wait, are you planning to move these element pairs apart after displacement?" If it's "yes" then I might as well just cancel this PR; we'll definitely have to just communicate more topology data if we can't use nodal ids or nodal coordinates to reconstruct it.

arovinelli:
> @arovinelli, what was the answer to "Wait, are you planning to move these element pairs apart after displacement?" If it's "yes" then I might as well just cancel this PR; we'll definitely have to just communicate more topology data if we can't use nodal ids _or_ nodal coordinates to reconstruct it.

@roystgnr  thinking on node id lines, what might help here is a map with fake-neighbor elem,side pairs. Or, could we subclass Node or add a method to keep track of intially coincident nodes and compare by such list in the test??

roystgnr:
Okay, I'll see what else I can do then.

lindsayad:
I feel like this might have been the right solution since we always have an undisplaced reference mesh in every MOOSE calculation (and this work was done for MOOSE)

lindsayad:
Also if it is related, for adaptive meshing we always undisplace the displaced mesh beforehand. This was done way back in 2014 by @jwpeterson in idaholab/moose@f9561331

PR #2363, opened 2019-12-06
Communicate make_links_to_me_local neighbor_side

roystgnr:
With user code playing tricks with neighbor topology while disconnecting
neighbor geometry, there's really no other way to make sure we don't
accidentally break that topology.

Like #2362 this fixes the test case in idaholab/moose#14456 for me; unlike #2362 those problems should *stay* fixed even if the geometry of the non-neighbor "neighbors" gets displaced out of alignment.

@arovinelli, could you confirm that the fix here works for you too?

PR #2364, opened 2019-12-09
make_nodes_parallel_consistent only if valid

roystgnr:
This fixes the bug from #2354 for me on cases where we have a mesh that
isn't yet prepared for use, which hopefully will fix the distributed
sweep failures in #2363


roystgnr:
That failure is unrelated, but I can reproduce it and I'll tack the fix onto the end here.

roystgnr:
And now *that* failure is unrelated; Parmetis just isn't expecting a processor to start out with a swath of disconnected-from-others NodeElem elements.

roystgnr:
Going to test the fix for that separately rather than waiting for a whole sweep to make it to np 16.

PR #2366, opened 2019-12-10
Fix mesh reference in VTKIO write

tvincent056:
If the "writing" constructor is used, MeshInput has an empty pointer to mesh. Calling write_nodal_data then causes dereferencing null pointer. Should be fine to use the const mesh reference from MeshOutput.


jwpeterson:
Thanks, this looks obviously correct to me. I guess it's not surprising that we don't have any test coverage of this because the normal use case is that the VTKIO object is created from a non-const reference.

Issue #2369, opened 2019-12-11
Optimize periodic boundaries ?


lindsayad:
We have a user on our [mailing list](https://groups.google.com/forum/#!topic/moose-users/7ZwXz6Ztlrc) who is spending a lot of time in his simulation finding periodic boundary neighbors as shown by this profile that I generated for him:

![periodic-boundaries-profile](https://user-images.githubusercontent.com/10248304/70651622-035c8d00-1c06-11ea-9fef-4a25014eda76.png)

You can see that we have a diamond pattern where we branch in `EquationSystems::reinit_solutions` to `System::reinit_constraints` and `DofMap::distribute_dofs` and then end up calling `PeriodicBoundaries::neighbor` from both of those branches. Is there a way we could cache information so that one of these branches just uses the information generated by the other?

Also perhaps this deserves its own issue...but should we make the effort to detect when we're running with only one process and then skip adding algebraic ghosting? I ran this profile serially so in an ideal world I feel like I shouldn't see the second branch happening at all. (The stack from DofMap::distribute_dofs down to PeriodicBoundaries::neighbor runs through DofMap::merge_ghost_functor_outputs)

roystgnr:
Ghosting is just one of a few things we should be skipping in serial but aren't, IIRC.  I'll take a look at that first since it should be an easier and more obvious fix.

With those periodic neighbors, I think my first big question is: why exactly are we falling back so heavily on a linear search?!  Of course hunting across the domain is going to be slow if it's O(N) rather than O(log N).

At first glance - I see an interface added in #1012 that **only** does a linear search?  We're not actually using that now, are we?

jwpeterson:
I did have [concerns](https://github.com/libMesh/libmesh/pull/1012#issuecomment-229767034) about this linear search at the time. Apparently performance was not an issue for @snschune's application at the time, but now it seems like it is? Without looking at it in too much detail, would it be possible in this application to not call the version of `PointLocatorTree::operator()` that triggers this linear search?

roystgnr:
We are actually using that now.  When the linear search implementation was added for the candidate_elements search, I didn't worry about it because surely none of *my* code was going to use it.  Then when we had issues with PeriodicBoundaries and overlapping elements, I must have seen that the candidate_elements API and it was an obvious fix, but forgotten that the current implementation was too slow to be usable.

Let me see how hard this is going to be to fix properly.

jwpeterson:
[...]

PR #2370, opened 2019-12-11
Avoid calling n_vertices virtual a bunch of times

lindsayad:
In basic elem calculations like hmin and hmax these calls can actually become a substantial portion of run-time. For instance in the simulation case referenced by #2369, applying this patch results in a 18% reduction in overall simulation time, and in particular cuts the cumulative time spent in `Elem::hmax` and its callees from 55 seconds to 24 seconds.

### DIFF
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -346,10 +346,12 @@ Point Elem::centroid() const
 {
   Point cp;
 
-  for (unsigned int n=0; n<this->n_vertices(); n++)
+  const auto n_vertices = this->n_vertices();
+
+  for (unsigned int n=0; n<n_vertices; n++)
     cp.add (this->point(n));
 
-  return (cp /= static_cast<Real>(this->n_vertices()));
+  return (cp /= static_cast<Real>(n_vertices));
 }
 
 
@@ -358,10 +360,13 @@ Real Elem::hmin() const
 {
   Real h_min=std::numeric_limits<Real>::max();
 
-  for (unsigned int n_outer=0; n_outer<this->n_vertices(); n_outer++)
-    for (unsigned int n_inner=n_outer+1; n_inner<this->n_vertices(); n_inner++)
+  // Avoid calling a virtual a lot of times
+  const auto n_vertices = this->n_vertices();
+
+  for (unsigned int n_outer=0; n_outer<n_vertices; n_outer++)
+    for (unsigned int n_inner=n_outer+1; n_inner<n_vertices; n_inner++)
       {
-        const Point diff = (this->point(n_outer) - this->point(n_inner));
+        const auto diff = (this->point(n_outer) - this->point(n_inner));
 
         h_min = std::min(h_min, diff.norm_sq());
       }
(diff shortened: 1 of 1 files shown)


jwpeterson:
> cuts the cumulative time spent in `Elem::hmax` and its callees from 55 seconds to 24 seconds.

Whoa. And this is all overhead due to `n_vertices()` being a virtual function?! Because it is an inline function returning a constant value for all derived Elem types, which shouldn't take any time at all.

permcody:
Yes - this is insane. Makes we wonder how much long hanging optimization fruit there is hanging on the the tree.

lindsayad:
> Because it is an inline function

IIUC, virtuals can only rarely be in-lined: https://stackoverflow.com/a/733791/4493669 and https://www.geeksforgeeks.org/inline-virtual-function/

roystgnr:
Making this kind of optimization tersely was part of what motivated our range-for helpers.  I *thought* I had done a sweep through the code back then to find likely candidates for replacement, but guess I missed this one; thanks!

> it is an inline function returning a constant value for all derived Elem types, which shouldn't take any time at all.

As long as the compiler can prove the virtual function isn't being overridden, anyway.  So if the object is of known type (not just a reference to a possibly-overridden class) or if it's a reference/pointer to a type in which n_vertices() has been declared final, then the method can be inlined.  But that basically never happens.

src/geom/elem.C
```diff
@@ -358,10 +360,13 @@ Real Elem::hmin() const
...
-  for (unsigned int n_outer=0; n_outer<this->n_vertices(); n_outer++)
-    for (unsigned int n_inner=n_outer+1; n_inner<this->n_vertices(); n_inner++)
+  // Avoid calling a virtual a lot of times
+  const auto n_vertices = this->n_vertices();
+
+  for (unsigned int n_outer=0; n_outer<n_vertices; n_outer++)
+    for (unsigned int n_inner=n_outer+1; n_inner<n_vertices; n_inner++)
       {
-        const Point diff = (this->point(n_outer) - this->point(n_inner));
+        const auto diff = (this->point(n_outer) - this->point(n_inner));
```
roystgnr: Why bother changing this?

lindsayad: Avoid an extra stack frame of constructor call? :laughing:

lindsayad: Or maybe one day the `point` method may not return a `Point` ;-) I can definitely remove this change...

roystgnr: Wait, does this actually remove a call?  diff will still get declared as a const Point, won't it?

roystgnr: I just like avoiding auto except where it adds flexibility (e.g. in templated code) or removes verbosity (e.g. in iterators to crazy long nested container types) because it can remove type-safety (when otherwise the compiler might scream that the type you expected and the type you got aren't the same).

roystgnr:
> guess I missed this one

You know, there are a lot of other cases I've missed where we might as well use the newer idioms; I'll do a sweep through and set up a PR now.

src/geom/elem.C
lindsayad: > Wait, does this actually remove a call? diff will still get declared as a const Point, won't it?

No I believe it will actually be a `TypeVector` based on the `operator -` method that `Point` inherits from `TypeVector`

jwpeterson: I think the generated assembly for this line wouldn't change due to this diff. You might consider 
```
h_min = std::min(h_min, (this->point(n_outer) - this->point(n_inner)).norm_sq());
```
but again I wouldn't really expect a big difference.

lindsayad: > I just like avoiding auto except where it adds flexibility (e.g. in templated code) or removes verbosity (e.g. in iterators to crazy long nested container types) because it can remove type-safety (when otherwise the compiler might scream that the type you expected and the type you got aren't the same).

That's fair. Explicit typing can also make the code more clear and legible to a later developer...

roystgnr:
Wow, even in DenseSubVector::size() we have a virtual method that isn't marked final...

Issue #1053, opened 2016-08-03
Fix our for loops


roystgnr:
That means, in the conditions, we cache the result of any function call which we know won't change during the progress of the loop but which the compiler can't prove won't change during the progress of the loop.

E.g. see an example https://github.com/libMesh/libmesh/pull/1052#discussion_r73400173 where we know that the compiler should be doing a no-op in opt mode but where the compiler has no choice except to make n_neighbors virtual function calls.

jwpeterson:
While we're at it, I'd like to enforce having exactly one space between `for` and the opening `(` in for loop declarations, i.e.
```
for (int foo=bar; bar<N; ++bar)
```
for consistency.  Spacing between operators I don't care as much about.

jwpeterson:
Consistency between loop counter and termination clauses was addressed in #1229.

roystgnr:
This issue was mostly addressed by the move to ranged for, but I figured I'd look at the remaining cases.

One bit of weirdness we still have in a few places: casting things to unsigned int so that we can use unsigned int as a loop iterator.

I figured that this should accomplish basically nothing (there's no speed gain from using shorter integers unless you can vectorize a bunch of them in a row, right?) but that [perhaps using signed integers would help](https://kristerw.blogspot.com/2016/02/how-undefined-signed-overflow-enables.html), as I speculated about in #1229.

So [I decided to check](https://github.com/roystgnr/test-for-iterators).

And what I see at first glance is:

With gcc 4.8.5, everything is equally slow.

With clang 3.4.2, using 64-bit integers in the vectors we operate on makes those cases ~30% faster, regardless of iterator type.

With gcc 7.2.0, the still-slow combinations are uint->size_t, int->anything-but-ptrdiff_t, and ptrdiff_t->anything-64-bit.  There's a ~10% speedup for int->ptrdiff_t, ~15% for uint->ptrdiff_t, 25% for size_t->anything-64-bit, and a ~35% speedup for anything-but-int->anything-32-bit.

I can't make any sense of any of that, but since our real code has been written with my same lack of understanding, empirically it seems like the thing to do is to stick with size_t everywhere and not try to bother using an unsigned type or a shorter type for loop counters, even for the many vectors which we know aren't going to have a size past 2^8, much less 2^32.  Trying to use unsigned int as a loop counter might even be a pessimization!

[...]

PR #2371, opened 2019-12-12
More range for loops

roystgnr:
The PR I promised in #2370, converting as many traditional for loops to range for loops as seemed reasonable, focusing only on those loops where there was a function call in the counter test so that there would be a good chance of getting a little optimization out of it.

I haven't touched the src/reduced_basis loops - after accidentally misconverting one loop (connected_elements.rend() should not be cached in connect_families(), and I even had a comment to that effect...) I decided to avoid loops where I wasn't semi-familiar with the code.  I may get braver once Akselos applications are added to our regular CI runs.

### DIFF
--- a/include/base/dof_object.h
+++ b/include/base/dof_object.h
@@ -22,6 +22,7 @@
 
 // Local includes
 #include "libmesh/id_types.h"
+#include "libmesh/int_range.h"
 #include "libmesh/libmesh_config.h"
 #include "libmesh/libmesh_common.h"
 #include "libmesh/libmesh.h" // libMesh::invalid_uint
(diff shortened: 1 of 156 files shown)


roystgnr:
I tried to keep index types unchanged as much as possible (obviously it's not possible with index_range()), but that was mostly because I didn't want to have to scan so carefully for width-dependent binary I/O calls.

As far as vectorization is concerned I don't think 32-bit vs 64-bit counters matter, and although I've been told that signed is better than unsigned, my results in #1053 (which I really should have linked from this PR before now) seemed to show the opposite!

jwpeterson:
> I've been told that signed is better than unsigned

Yep, that is what I was referring to. In #1053 I can't recall if you actually checked the vectorization output of the compilers in the different cases, though I agree that what really matters in the end is speed, vectorized or not.

roystgnr:
Will merge after app tests pass.  In hindsight I should have done the header refactoring in a separate PR, but I didn't realize we were using a header where a forward declaration should do until I found myself needing to use header after header in the same place.

PR #2372, opened 2019-12-13
Better error message when invalid node id is found in nodeset.

jwpeterson:
Previously the error message for this case was the following:
```
BoundaryInfo::add_node(): Could not retrieve pointer for node 4294967295, no boundary id was added.
```
because the invalid array access happened to return 0 instead of e.g. segfaulting. Then we subtracted 1 and cast the result to unsigned, which made the error message more confusing than necessary.

### DIFF
--- a/src/mesh/exodusII_io.C
+++ b/src/mesh/exodusII_io.C
@@ -401,6 +401,14 @@ void ExodusII_IO::read (const std::string & fname)
           {
             int exodus_id = exio_helper->node_sets_node_list[i + offset];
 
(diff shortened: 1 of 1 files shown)


roystgnr:
This is exactly the sort of use case I was pushing to make terser with my ```libmesh_always_assert``` etc. proposal way back when; just sayin'.

A couple days ago my daughter asked me how people hacked into computers, and I decided to only explain the technical "how buffer overflows and stack smashing work" side of the problem, because the sociological "why lazy bums like daddy don't sanity-check external inputs enough" side of the problem was too depressing.

PR #2373, opened 2019-12-13
Simplify+inline 1D Lagrange shape function calls

roystgnr:
This gives me nearly a 40% speedup in 3D, on a problem that was only using Lagrange for *mapping* functions.  Problems without tensor product elements or with meshes where mapping functions are often cached probably won't do so well though.

I'd like to thank @permcody for his "[Makes we wonder how much long hanging optimization fruit there is hanging on the the tree.](https://github.com/libMesh/libmesh/pull/2370#issuecomment-564708754)" comment - the benchmarking leading to this PR was begun by indignant anger and ended in shamed acceptance.

We still need to vectorize this code, but I think that'll be a much bigger refactoring.

We could get an analogous tensor product elements speedup with similar refactoring on other FE families: BERNSTEIN, HIERARCHIC, and SZABAB at first glance.  Going to throw CI at the LAGRANGE refactoring first.

### DIFF
--- a/include/Makefile.in
+++ b/include/Makefile.in
@@ -667,6 +667,7 @@ include_HEADERS = \
         fe/fe_base.h \
         fe/fe_compute_data.h \
         fe/fe_interface.h \
+        fe/fe_lagrange_shape_1D.h \
         fe/fe_macro.h \
         fe/fe_map.h \
         fe/fe_transformation_base.h \
--- a/include/fe/fe_lagrange_shape_1D.h
+++ b/include/fe/fe_lagrange_shape_1D.h
@@ -125,85 +125,104 @@ Real fe_lagrange_1D_shape(const Order order,
 
 
 inline
-Real fe_lagrange_1D_shape_deriv(const ElemType,
-                                const Order order,
-                                const unsigned int i,
-                                const unsigned int libmesh_dbg_var(j),
-                                const Point & p)
+Real fe_lagrange_1D_linear_shape_deriv(const unsigned int i,
+                                       const unsigned int libmesh_dbg_var(j),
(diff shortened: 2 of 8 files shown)


jwpeterson:
Wait is it "long hanging" or "low hanging"? Guess it depends on what kind of fruit...

jwpeterson:
Can you comment more on the 40% number? Was that 40% of the total runtime? What type of problem (size, algorithm, etc.) did you run?

roystgnr:
Both.  40% speedup from an 800 line diff counts as "low hanging", and refactoring a design that has been part of libMesh since the beginning counts as "long hanging".

roystgnr:
In adaptivity_ex3,
```
example-oprof dimension=3 approx_type=HERMITE approx_order=3 refinement_type=h max_r_steps=4 -pc_type asm -pc_asm_overlap 1 -sub_pc_factor_mat_ordering_type nd -sub_pc_type ilu -sub_pc_factor_levels 1 -sub_pc_factor_zeropivot 0 -sub_pc_factor_shift_type nonzero
```

Nearly 40% of total runtime, but that's because this case is selected for its inefficiency: we spend most of our runtime in the error estimator (a separate PR to improve that aspect will be coming along in a bit) and we spend most of *that* runtime in ```FE::reinit(elem, side)```.

jwpeterson:
FWIW I ran the test several times before and after this PR. There was a lot of variation in the timings but the fastest times were 71s (before optimization) and 42s (after optimization), or ~1.7x speedup. Nice!

lindsayad:
I guess because these are all inlined functions there's no practical reason for include guards

PR #2374, opened 2019-12-13
Reduce default JumpErrorEstimator quadrature

roystgnr:
I fear this is going to break people's 2D/3D AMR regression tests
whenever slightly-differing error estimator values lead to differing
refinement patterns, but IMHO sorting that out will be worth it; we can
get nearly-as-good results at roughly half (2D) or a quarter (3D) of the
cost by not overintegrating.

This shouldn't be merged until we've run as many apps through it was we can to see how much hassle there will be.


roystgnr:
I changed my mind about the "how much hassle" question - there's not enough reason to make the more efficient quadrature opt-out rather than opt-in.

PR #2380, opened 2019-12-17
Check elem->infinite() in FEMap::map*

### LINKED ISSUE #2272: FEMap::map() fails for infinite elements
I found out, that the map() function (most probably some others as well?) fail when called on infinite elements. The reason is that the ``FEInterface::shape_function(dim, fe_type)`` function cannot handle infinite elements: having only ``dim`` and ``fe_type`` available, we cannot distinguish finite from infiite elements.
I currently don't see an easy fix for the FEInterface-function, I would fix this on the level of ``FEMap``.

Or does someone have a more elegant idea? I think, one of the reasons for the FEInterface-class is to provide a uniform interface for finite and infinite elements...

roystgnr:
We don't have an InfFEMap::map_deriv to call but we can give a saner
error line in FEMap::map_deriv, and we can get FEMap::map working on
infinite elements easily.  This fixes #2272 for me.

### DIFF
--- a/src/fe/fe_map.C
+++ b/src/fe/fe_map.C
@@ -2046,6 +2046,11 @@ Point FEMap::map (const unsigned int dim,
 {
   libmesh_assert(elem);
 
+#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+  if (elem->infinite())
+    return InfFEMap::map(dim, elem, reference_point);
+#endif
+
   Point p;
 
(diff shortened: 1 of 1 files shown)


roystgnr:
Going to pull this today if nobody objects: it's a no-op in the --disable-infinite case, it's at least an improvement in the --enable-infinite case, and we can reopen #2272 if it turns out not to be a *complete* fix.

PR #2382, opened 2019-12-18
Always try to snoop MPI from PETSc in TIMPI

lindsayad:
Without this PETSc snooping, the user has to specify MPIHOME or `--with-mpi=<dir>` in order to pick up the correct MPI installation directory.

### DIFF
--- a/contrib/timpi/configure
+++ b/contrib/timpi/configure
@@ -667,10 +667,20 @@ enablewarnings
 enabledeprecated
 timpi_optional_LIBS
 timpi_optional_INCLUDES
+MPI_INCLUDES_PATHS
+MPI_INCLUDES_PATH
+MPI_LIBS_PATHS
+MPI_LIBS_PATH
+MPI_LIBS
+MPI
+CXXCPP
+MPI_IMPL
 PETSC_CC_INCLUDES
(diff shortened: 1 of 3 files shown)


roystgnr:
Great, thank you!

roystgnr:
I may have one more swath of config changes to TIMPI (there's a new system I set up where our autodetection isn't working the way I expected, and I also want to see the effects of this change there), and then I plan to split TIMPI out into its own submodule.  Might be best to do the MOOSE libMesh update pre-new-submodule so that you get these recent improvements without *immediately* getting potental new hassle alongside them.

lindsayad:
👍

lindsayad:
You know what's comical is that this PR didn't actually successfully snoop MPI at all. All it does is throw PETSc include and lib directories into `timpi_optional_INCLUDES` and `timpi_optional_LIBS`. The only reason this made MOOSE testing work again is because it prevented `ACX_MPI` from running ([which was the thing throwing in `-L/usr/lib`](https://github.com/idaholab/moose/pull/14464#issuecomment-563790153)). We're getting good include and library directories for MPI when compiling on `moosebuild` (and on my own local system generally) because we have `CC=mpicc` and `CXX=mpicxx`.

PR #2386, opened 2019-12-20
TIMPI submodule

roystgnr:
This deletes the libMesh/libmesh copies of TIMPI files, and gets them directly from the new libMesh/TIMPI repo instead.

We might need to rework our libMesh documentation a little, though, if user can no longer build libMesh from git without a submodule init first.  MetaPhysicL had the same issue but that was an optional rather than a required dependency.


roystgnr:
Huh.  Wasn't actually expecting that to work on the first go.  I guess all the MOOSE CI already does submodule initialization properly to get the MetaPhysicL dependent parts of MOOSE working?

I'll probably merge this afternoon if nobody screams first.

jwpeterson:
I'm not sure what the best place to announce this is, but anyone with an existing libmesh git repo will need to run
```
git submodule update --init contrib/timpi
```
after pulling this update. If that doesn't work because you already have some existing build byproducts, etc. in the contrib/timpi directory, you can try manually deleting it:
```
rm -rf contrib/timpi
git checkout contrib/timpi
```
Finally, the submodule was added with an SSH URL hard-coded (git@github.com:libMesh/TIMPI) so if you are not running commands as a user with an SSH key uploaded to GitHub, the update won't work. @roystgnr we may want to use a "relative" URL for TIMPI as is done for MetaPhysicL,
```
$ cat .gitmodules 
[submodule "contrib/metaphysicl"]
	path = contrib/metaphysicl
	url = ../../roystgnr/MetaPhysicL
[submodule "contrib/timpi"]
	path = contrib/timpi
	url = git@github.com:libMesh/TIMPI
```
I think this works for moosebuild because it has ssh keys.

PR #2388, opened 2019-12-30
External netcdf

roystgnr:
Thanks to Edgar (at openmail dot cc) for coming up with this:

> I think that I found a way to use my system's netcdf with libMesh (make check gave no errors).

I haven't had time to test it out myself yet, but we'll see what CI thinks of it now and we can take a closer look at it later.


roystgnr:
What the heck?  That's just an unused variable warning (on a --enable-werror configuration) - nothing to do with these commits.  How did that get past CI previously?

roystgnr:
Ah.  It's an unused variable iff ```!defined(LIBMESH_HAVE_EXODUS_API)```, and this patch is causing us to end up with ```exodus........................... : no``` in the configure summary.

jwpeterson:
Just FYI, one reason we previously bundled our own netcdf was that you had to manually set some constants in order for everything to work correctly with Exodus (see e.g. 656a50628ab95a5943e35313c1bc10e21cceff4f) but that no longer seems to be necessary in newer versions of netcdf, as those manually set constants are no longer used.

roystgnr:
Closing this; looks like #2602 has fixed the problems with it.

PR #2389, opened 2019-12-30
RB Change: Add support for POD training in RBConstruction

dknez:
Proper Othogonal Decompositoin (POD) provides an alternative method for training a reduced basis space. Here we add support for POD as an alternative (non-default) approach.

It can be tested with reduced_basis_ex1 by setting `RB_training_type = POD` in reduced_basis_ex1.in.


dknez:
This is an RB-only change that introduces a new code-path and doesn't affect any old code, so I'll go ahead an merge this now that it has passed the tests.

Issue #2359, opened 2019-12-05
negative Jacobian


BalticPinguin:
Dear all,
I am generating my mesh with tetgen and from time to time the calculation breaks due to 'negative Jacobian', so I would like to finally get this fixed now, but need help in figuring out what actually happens. (I am not sure this is related to #2274 somehow, but at least this is still persistent.)

I created a (not really) [minimum example](https://gist.github.com/BalticPinguin/5e6f7da0b46957a028dc4372664bdc05.js), where I see the error.
Unfortunately, it seems to critically depend on ``all`` coordinates, so when I tried to shorten the list of points, I could not reproduce this failure.

The error I than get is
````
Elem Information
   id()=1, unique_id()=177682, processor_id()=0
   type()=TET4
   dim()=3
   n_nodes()=4
    0  Node id()=4, processor_id()=0, Point=(x,y,z)=(-0.0350807, 0.00516588, -0.0571602)
    DoFs=
    1  Node id()=5, processor_id()=0, Point=(x,y,z)=(-0.0367887, 0.00678192, -0.0559037)
    DoFs=
    2  Node id()=6, processor_id()=0, Point=(x,y,z)=(-0.0338311, 0.00778059, -0.0576155)
    DoFs=
    3  Node id()=7, processor_id()=0, Point=(x,y,z)=(0.0260922, 0.00804063, 0.0647749)
    DoFs=
   n_sides()=4
    neighbor(0)=864761
    neighbor(1)=519169
    neighbor(2)=337711
    neighbor(3)=702062
   hmin()=0.00266602, hmax()=0.13645
   volume()=-1.72416e-07
   active()=1, ancestor()=0, subactive()=0, has_children()=0
   parent()=nullptr
   level()=0, p_level()=0
   refinement_flag()=DO_NOTHING
   p_refinement_flag()=DO_NOTHING
   infinite()=0
   DoFs=
  -0.001708  0.0012496  0.0611729
  0.00161604  0.00261471  0.00287475
  0.0012565  -0.0004553  0.121935
ERROR: negative Jacobian -1.0345e-06 at point index 0 in element 1
Stack frames: 9
0: libMesh::print_trace(std::ostream&)
1: libMesh::MacroFunctions::report_error(char const*, int, char const*, char const*)
[...]

PR #2394, opened 2020-01-03
Add consistency-check to default-options for tetgen

BalticPinguin:
fix for #2359.


jwpeterson:
This looks fine to me but I also attached testing for non-strict LGPL builds, since otherwise the Tetgen stuff is not compiled.

PR #2396, opened 2020-01-06
[WIP] Make geometric calculations more generic

### LINKED ISSUE #2121: Automatic Differentiation for geometric entities?
So to run moving mesh problems with perfect Jacobians in MOOSE I've developed methods like [`Assembly::computeSinglePointMapAD`](https://github.com/idaholab/moose/blob/next/framework/src/base/Assembly.C#L756) that essentially reproduces `FEMap::compute_single_point_map` but makes changes like:
<pre>
      for (auto i : index_range(elem_nodes))
      {
        libmesh_assert(elem_nodes[i]);
        <b>libMesh::VectorValue&ltDualReal&gt elem_point = *elem_nodes[i];</b>
        unsigned dimension = 0;
        for (const auto & disp_num : _displacements)
          <b>elem_point(dimension++)
              .derivatives()
              .insert(disp_num * _sys.getMaxVarNDofsPerElem() + i) = 1.;</b>

        _ad_dxyzdxi_map[p].add_scaled(elem_point, dphidxi_map[i][p]);

        if (_calculate_xyz)
          _ad_q_points[p].add_scaled(elem_point, phi_map[i][p]);
      }
</pre>
Doing this allows me to propagate the dependence on non-linear displacement variables down into quantities like JxW, normals, curvatures, and shape function gradients (and consequently variable solution gradients). I'm pained, however, by the amount of code I've duplicated. By dream would be to be able to do these calculations within `libMesh::FE` objects. 

Realizing that dream would likely mean an overhaul of the `FE` object. In MOOSE I've introduced an enum `ComputeStage` that can take on `RESIDUAL` and `JACOBIAN` values. AD (automatic differentiation) objects are then templated on this enum so that their data members essentially take on `Real` and `DualNumber` character for residual and jacobian objects respectively. I imagine that something similar would have to be done to `FE`: introduction of another template parameter. Such an overhaul could not break existing applications so it might actually mean leaving `FE` alone (at least its APIs) and creating a new class that could share code (probably through inheritance).

I dunno, do folks thing it's possible to avoid all this code duplication in a reasonable way?

lindsayad:
Classes that need to be templated:

- [x] FE*
- [x] Node
- [x] Point
- [x] Elem*
- [x] Mesh*

Other To-Do items:

- [x] Compile
- [ ] Unit tests
- [ ] Fix merge conflicts
- [ ] Have someone review/approve this monstrosity

We'll have to talk about what we want to do for backwards compatability.

Closes #2121


roystgnr:
"Work in progress" means I can merge #2395 and undercut all your hard work without feeling *too* guilty?

lindsayad:
Sigh...haha yes go ahead.

roystgnr:
Ouch, I barely changed browser tabs before github started yelling at us here about conflicting files.

Sorry about the hassle.  On the bright side, the expanded unit test coverage should make it easier to do this work without regressions!

roystgnr:
@jwpeterson pointed out to me that our _impl.h convention (where most users *don't* need to include them because we instantiate the common case(s) manually) isn't intuitively obvious.  Let's come up with some boilerplate comment to document that convention in the _impl.h headers as part of this PR, then when we're happy with that we can add it to the existing dense_matrix*_impl.h headers too?

lindsayad:
Sounds good to me

jwpeterson:
So @lindsayad I think you are going to have to template every class in libmesh, like EquationSystems, System, etc.? I'm starting to get scared.

lindsayad:
> So @lindsayad I think you are going to have to template every class in libmesh, like EquationSystems, System, etc.? I'm starting to get scared.

Haha, this is why I threw this up here (with a WIP label) so that you guys could give your input as I go. So I think that I should only have to class template geometric entities. Algebraic entities should escape relatively unscathed, except that algebraic class methods which include geometric entities in their argument list will probably have to become templated; you can see https://github.com/libMesh/libmesh/pull/2396/commits/d2b9728f27ce8088eaf5d1c7d5004944b480d929 for an example of this. If an algebraic class has a geometric entity as a data member or a method with a geometric return type that is unrelated to the argument list (e.g. I wouldn't be able to deduce a function template argument from the function args), then I have to get a little more clever. So for the `DofMap` for example, I ended up creating a `MeshGhosting` class which `MeshBase` inherits from that contains the APIs that the `DofMap` needs to use and then changed the `DofMap::_mesh` data member to be of `MeshGhosting` type. 

I've been thinking about how to do this with any less work (and any less destruction) than what I've been doing but so far I've yet to have any bright ideas.

lindsayad:
I never thought I'd see the day...my branch compiles!!! Now of course compiling a `RealType = DualNumber` unit test...that will be the next battle

lindsayad:
Has anyone ever seen an error like this before?
```
In file included from ../include/libmesh/dof_map_impl.h:40,
                 from ../include/libmesh/instantiate_real_type.h:22,
                 from /home/lindad/projects/moose/scripts/../libmesh/tests/ad/geometric_ad_test.C:2:
/home/lindad/projects/moose/scripts/../libmesh/contrib/timpi/src/algorithms/include/timpi/parallel_sync.h:502:6: error: ‘void TIMPI::push_parallel_vector_data(const TIMPI::Communicator&, const MapType<unsigned int, std::__debug::vector<std::__debug::vector<ValueType, A1>, A2>, ExtraTypes ...>&, const ActionFunctor&) [with MapType = std::__debug::map; ValueType = std::pair<unsigned int, double>; A1 = std::allocator<std::pair<unsigned int, double> >; A2 = std::allocator<std::__debug::vector<std::pair<unsigned int, double> > >; ExtraTypes = {std::less<unsigned int>, std::allocator<std::pair<const unsigned int, std::__debug::vector<std::__debug::vector<std::pair<unsigned int, double>, std::allocator<std::pair<unsigned int, double> > >, std::allocator<std::__debug::vector<std::pair<unsigned int, double>, std::allocator<std::pair<unsigned int, double> > > > > > >}; ActionFunctor = libMesh::DofMap::allgather_recursive_constraints(libMesh::MeshBaseTempl<RealType>&) [with RealType = MetaPhysicL::DualNumber<double, MetaPhysicL::DynamicSparseNumberArray<double, unsigned int> >]::<lambda(libMesh::processor_id_type, const std::__debug::vector<std::__debug::vector<std::pair<unsigned int, double> > >&)>]’, declared using local type ‘const libMesh::DofMap::allgather_recursive_constraints(libMesh::MeshBaseTempl<RealType>&) [with RealType = MetaPhysicL::DualNumber<double, MetaPhysicL::DynamicSparseNumberArray<double, unsigned int> >]::<lambda(libMesh::processor_id_type, const std::__debug::vector<std::__debug::vector<std::pair<unsigned int, double> > >&)>’, is used but never defined [-fpermissive]
```
Used but never defined...I don't get it. There's clearly a definition for the template `TIMPI::push_parallel_vector_data` and you can't declare a lambda without a definition...

lindsayad:
I'm seeing two paths forward with this:

1. Don't template algebraic classes at all. This would essentially necessitate the user always having two meshes (which is what MOOSE does) and the algebraic objects (`EquationSystems, System, DofMap`) would always interact with the non-displaced mesh. This would lead to a lot less content in this PR, but would make it rather inconvenient to query the displaced mesh's  elements and nodes for dof information.
2. Basically template the whole world, e.g. template `EquationSystems, System, DofMap`. I don't think it makes sense to just say template `EquationSystems` because then `System` would have to `get_equation_systems().get_mesh()` in order to get a complete mesh object (e.g. not some dumbed-down base class object that you can't query for elements/nodes etc.)

lindsayad:
On second thought, I'm leaning towards not templating algebraic objects at all. You know, Occam's Razor and stuff. Then I'd just have to do something smart for cloning DofObject information on to the displaced mesh's Nodes and Elems

roystgnr:
So that "used but never defined" looks like my bug?  At one point I had a KeyType parameter in the push vector<vector> specialization, but now it looks like the KeyType parameter exists in the declaration but not in the definition, and so your compiler may rightly be confused by that.  Not sure why my compiler and the CI compilers haven't already hit that or how you set it off, though.

roystgnr:
There's got to be a better phrase than "algebraic classes" for System/DofMap/EquationSystems ... if only because I've already abused that vocabulary in the opposite direction in parallel_algebra.h ... let me call them "system classes" for lack of a better word.

I agree with your dichotomy, and my vote would be to avoid templating the system classes if possible, and I think it would be possible, if we're willing to allow for a little dynamic_cast hackery in user space to get AD geometry when that's available.

lindsayad:
> There's got to be a better phrase than "algebraic classes" for System/DofMap/EquationSystems

Seemed consistent with algebraic ghosting :sweat_smile: 

> I agree with your dichotomy, and my vote would be to avoid templating the system classes if possible, and I think it would be possible, if we're willing to allow for a little dynamic_cast hackery in user space to get AD geometry when that's available.

Cool :+1:

roystgnr:
> Seemed consistent with algebraic ghosting :sweat_smile:

Oh, well, see, there's your problem.  If you try to come up with your own names then consistency is difficult, but if you try to match my naming schemes then consistency is *impossible*.

jwpeterson:

lindsayad:
Closing in favor of the research work on #2576

PR #2401, opened 2020-01-10
n_levels = MeshTools::n_levels()

roystgnr:
This shortens the code and fixes a bug for me on high processor counts
(e.g. more than 21 processors on the fem_system_ex1 GMG example)

I've left the old code (commented out) to make fixes/reversions easier
in case there's any subtlety to the previous n_levels definition that
@pbauman or @bboutkov understood but I didn't; however this patch is
passing tests for me.


permcody:
Need to kick CIVET on this one too.

PR #2402, opened 2020-01-10
Cheaper adaptivity ex3 DiscontinuityMeasure

roystgnr:
This shaves off some runtime and adds slightly more coverage to the
unweighted quadrature rule option

### DIFF
--- a/examples/adaptivity/adaptivity_ex3/adaptivity_ex3.C
+++ b/examples/adaptivity/adaptivity_ex3/adaptivity_ex3.C
@@ -356,6 +356,14 @@ int main(int argc, char ** argv)
       // Compute any discontinuity.  There should be none.
       {
         DiscontinuityMeasure disc;
+
+        // This is a subclass of JumpErrorEstimator, based on
+        // measuring discontinuities across sides between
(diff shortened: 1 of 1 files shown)


permcody:
Looks like you'll need to resend the payload to kick CIVET.

permcody:
The power outage here was very serious - we still aren't at 100% after 5 days. Hopefully tomorrow!

jwpeterson:
> The power outage here was very serious - we still aren't at 100% after 5 days.

osprey

There's your problem right there.

permcody:
In all seriousness, I guess it was worse than that. Apparently some underground conduit carrying the big power cables to the EROB data center failed (by that I mean they stuck cameras in them and discovered they were full of dirt and rocks). Enough foreign debris made its way into those conduits to eventually short out those cables (that mechanism isn't fully understand or hasn't been relayed to us). 

They are attempting to bypass those conduits but I imagine they will be digging a big hole in the parking lot and replacing everything soon.

permcody:
In other news the C3 data center is running smoothly - too bad we only have Sawtooth in there and that's not available for general use yet!

PR #2405, opened 2020-01-16
Add ability to set a tolerance for point containment in the point_locator_tree

pshriwise:
@friedmud and I were looking at the settings for the point_locator tree and noticed that for the `find_element` call, the tolerance for finding points is always the default `TOLERANCE` value. These changes add the ability to set the tolerance for point containment in the tree search for applications that require a lower tolerance than the default.

First libmesh PR, so let me know if there's anything I'm missing.

### DIFF
--- a/include/utils/point_locator_base.h
+++ b/include/utils/point_locator_base.h
@@ -176,19 +176,19 @@ public:
    * Set a tolerance to use when checking
    * if a point is within an element in the mesh.
    */
-  virtual void set_find_element_tol(Real find_element_tol);
+  virtual void set_contains_point_tol(Real contains_point_tol);
 
   /**
    * Specify that we do not want to use a user-specified tolerance to
    * determine if a point is inside an element in the mesh.
    */
-  virtual void unset_find_element_tol();
+  virtual void unset_contains_point_tol();
 
   /**
    * Get the tolerance for determining element containment
    * in the point locator.
    */
-  virtual Real get_find_element_tol() const;
+  virtual Real get_contains_point_tol() const;
 
   /**
    * Get a const reference to this PointLocator's mesh.
@@ -233,12 +233,12 @@ protected:
    * \p true if we will use a user-specified tolerance for locating
    * the element.
    */
-  bool _use_find_element_tol;
+  bool _use_contains_point_tol;
 
   /**
    * The tolerance to use when locating an element in the tree.
    */
-  Real _find_element_tol;
+  Real _contains_point_tol;
 };
 
 } // namespace libMesh
--- a/include/utils/point_locator_base.h
+++ b/include/utils/point_locator_base.h
@@ -173,6 +173,24 @@ public:
   virtual void unset_close_to_point_tol();
 
   /**
+   * Set a tolerance to use when checking
+   * if a point is within an element in the mesh.
+   */
+  virtual void set_find_element_tol(Real find_element_tol);
+
+  /**
+   * Specify that we do not want to use a user-specified tolerance to
+   * determine if a point is inside an element in the mesh.
+   */
+  virtual void unset_find_element_tol();
+
+  /**
+   * Get the tolerance for determining element containment
+   * in the point locator.
+   */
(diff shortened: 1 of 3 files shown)


jwpeterson:
@pshriwise I sent you an invite to join a group whose PRs are tested automatically. For now I started all the testing by hand.

include/utils/point_locator_base.h
```diff
@@ -202,14 +220,25 @@ class PointLocatorBase : public ReferenceCountedObject<PointLocatorBase>
...
   /**
    * The tolerance to use.
    */
   Real _close_to_point_tol;
+
+  /**
+   * \p true if we will use a user-specified tolerance for locating
+   * the element.
+   */
+  bool _use_contains_point_tol;
```
jwpeterson: Both new class members (but especially the boolean) should be initialized in the constructor so they don't have indeterminate values if the user never calls `set_contains_point_tol()`. I'm not sure if this is what's causing the unit tests to fail, but it's possible that your new functionality is being used by default because this bool is not properly initialized.

src/utils/point_locator_tree.C
```diff
@@ -225,7 +229,7 @@ const Elem * PointLocatorTree::operator() (const Point & p,
               this->_element =
                 this->perform_linear_search(p,
                                             allowed_subdomains,
-                                            /*use_close_to_point*/ true,
```
jwpeterson: Is this comment actually wrong? If so, please fix it, if not, please don't delete it because it lets us know what the parameter is used for at a glance...

src/utils/point_locator_tree.C
```diff
@@ -290,8 +294,7 @@ const Elem * PointLocatorTree::perform_linear_search(const Point & p,
         {
           if (!use_close_to_point)
             {
-              if (elem->contains_point(p))
-                return elem;
+              if (elem->contains_point(p, _contains_point_tol)) { return elem; }
```
jwpeterson: Extra curly braces are not needed.

src/utils/point_locator_tree.C
pshriwise: Sorry, I interpreted that as the actual variable _use_close_to_point_tol rather than a comment and thought I was doing a little cleanup.

include/utils/point_locator_base.h
pshriwise: Good call. :man_facepalming: Adding that change now...

jwpeterson:
You may have already figured this out, but the error is that you have to initialize parameters in the same order they are declared:
```
./include/libmesh/point_locator_base.h: In constructor 'libMesh::PointLocatorBase::PointLocatorBase(const libMesh::MeshBase&, const libMesh::PointLocatorBase*)':
./include/libmesh/point_locator_base.h:236:8: error: 'libMesh::PointLocatorBase::_use_contains_point_tol' will be initialized after [-Werror=reorder]
  236 |   bool _use_contains_point_tol;
      |        ^~~~~~~~~~~~~~~~~~~~~~~
./include/libmesh/point_locator_base.h:230:8: error:   'libMesh::Real libMesh::PointLocatorBase::_close_to_point_tol' [-Werror=reorder]
  230 |   Real _close_to_point_tol;
      |        ^~~~~~~~~~~~~~~~~~~
../src/utils/point_locator_base.C:34:1: error:   when initialized here [-Werror=reorder]
   34 | PointLocatorBase::PointLocatorBase (const MeshBase & mesh,
      | ^~~~~~~~~~~~~~~~
```

jwpeterson:
That was a bit painful, but it looks like we are good now. I will squash your commits and fix some minor formatting issues before merging.

roystgnr:
Was this closed by accident via that commit message?

jwpeterson:
> Was this closed by accident via that commit message?

It was closed on purpose after being merged... I could have had it test again but didn't feel it was necessary.

jwpeterson:
After the squashed version was merged, that is.

pshriwise:
Thanks for the review and for following up on the test issues, @jwpeterson.

PR #2409, opened 2020-01-21
Adjusting the shape_ptr API

### LINKED ISSUE #2404: FEInterface::shape_function() fails for infinite elements
I mentioned this already in #2272: The ``FEInterface``-functions that return a ``FEInterface::shape_ptr`` cannot handle infinite elements and currently cannot be adapted easily.

Since this affects many places (currently I try to make the ``FEMap::init_reference_to_physical_map``-function compatible with inf elements) I'd prefer to adapt the interface of that function instead of circumventing it everywhere.

My suggestion would be to to write another set of functions, taking not only ``dim`` and ``FEType`` but also ``ElemType`` and deprecate the version with two arguments.
Any thoughts or better ideas?

Best regards

BalticPinguin:
This is supposed to fix #2404.
I am not fully convinced by the code, but lets see.
I am open to hear some oppinions!

### DIFF
--- a/src/fe/fe_bernstein_shape_1D.C
+++ b/src/fe/fe_bernstein_shape_1D.C
@@ -201,6 +201,19 @@ Real FE<1,BERNSTEIN>::shape(const Elem * elem,
 }
 
 
+template <>
+Real FE<1,BERNSTEIN>::shape(const FEType fet,
+                            const Elem * elem,
+                            const unsigned int i,
+                            const Point & p,
+                            const bool add_p_level)
+{
+  libmesh_assert(elem);
+  return FE<1,BERNSTEIN>::shape
+    (elem->type(),
+     static_cast<Order>(fet.order + add_p_level * elem->p_level()), i, p);
+}
+
 
 template <>
 Real FE<1,BERNSTEIN>::shape_deriv(const ElemType,
@@ -375,6 +388,20 @@ Real FE<1,BERNSTEIN>::shape_deriv(const Elem * elem,
      static_cast<Order>(order + add_p_level*elem->p_level()), i, j, p);
 }
 
+template <>
+Real FE<1,BERNSTEIN>::shape_deriv(const FEType fet,
+                                  const Elem * elem,
+                                  const unsigned int i,
+                                  const unsigned int j,
+                                  const Point & p,
+                                  const bool add_p_level)
+{
+  libmesh_assert(elem);
+  return FE<1,BERNSTEIN>::shape_deriv
+    (elem->type(),
+     static_cast<Order>(fet.order + add_p_level * elem->p_level()), i, j, p);
+}
+
 
 #ifdef LIBMESH_ENABLE_SECOND_DERIVATIVES
 
@@ -499,7 +526,7 @@ Real FE<1,BERNSTEIN>::shape_second_deriv(const ElemType,
                   + (45./16.)*pow<2>(1.-xi)*pow<2>(1.+xi);
         case 6:
           return -(15./16.)*pow<4>(1.+xi)
-                 + (15./8.)*pow<3>(1.+xi)*(1.-xi);
+                  + (15./8.)*pow<3>(1.+xi)*(1.-xi);
         default:
           libmesh_error_msg("Invalid shape function index i = " << i);
         }
@@ -572,6 +599,20 @@ Real FE<1,BERNSTEIN>::shape_second_deriv(const Elem * elem,
   return FE<1,BERNSTEIN>::shape_second_deriv
     (elem->type(),
      static_cast<Order>(order + add_p_level*elem->p_level()), i, j, p);
+}
+
+template <>
+Real FE<1,BERNSTEIN>::shape_second_deriv(const FEType fet,
+                                         const Elem * elem,
+                                         const unsigned int i,
+                                         const unsigned int j,
+                                         const Point & p,
+                                         const bool add_p_level)
+{
+  libmesh_assert(elem);
+  return FE<1,BERNSTEIN>::shape_second_deriv
+    (elem->type(),
+     static_cast<Order>(fet.order + add_p_level * elem->p_level()), i, j, p);
 }
 
 #endif
--- a/src/fe/fe_bernstein_shape_1D.C
+++ b/src/fe/fe_bernstein_shape_1D.C
@@ -526,7 +526,7 @@ Real FE<1,BERNSTEIN>::shape_second_deriv(const ElemType,
                   + (45./16.)*pow<2>(1.-xi)*pow<2>(1.+xi);
         case 6:
           return -(15./16.)*pow<4>(1.+xi)
-                  + (15./8.)*pow<3>(1.+xi)*(1.-xi);
+                 + (15./8.)*pow<3>(1.+xi)*(1.-xi);
         default:
           libmesh_error_msg("Invalid shape function index i = " << i);
         }
--- a/src/fe/fe_rational_shape_1D.C
+++ b/src/fe/fe_rational_shape_1D.C
@@ -36,18 +36,6 @@ namespace libMesh
 {
 
 template <>
-Real FE<1,RATIONAL_BERNSTEIN>::shape(const ElemType,
-                                     const Order,
-                                     const unsigned int,
-                                     const Point &)
-{
-  libmesh_error_msg("Rational bases require the real element \nto query nodal weighting.");
-  return 0.;
-}
-
-
-
-template <>
 Real FE<1,RATIONAL_BERNSTEIN>::shape(const Elem * elem,
                                      const Order order,
                                      const unsigned int i,
@@ -93,16 +81,26 @@ Real FE<1,RATIONAL_BERNSTEIN>::shape(const Elem * elem,
 
 
 template <>
-Real FE<1,RATIONAL_BERNSTEIN>::shape_deriv(const ElemType,
-                                           const Order,
-                                           const unsigned int,
-                                           const unsigned int,
-                                           const Point &)
+Real FE<1,RATIONAL_BERNSTEIN>::shape(const ElemType,
+                                     const Order,
+                                     const unsigned int,
+                                     const Point &)
 {
   libmesh_error_msg("Rational bases require the real element \nto query nodal weighting.");
   return 0.;
 }
 
+
+template <>
+Real FE<1,RATIONAL_BERNSTEIN>::shape(const FEType fet,
+                                     const Elem * elem,
+                                     const unsigned int i,
+                                     const Point & p,
+                                     const bool add_p_level)
+{
+  libmesh_assert(elem);
+  return FE<1,RATIONAL_BERNSTEIN>::shape(elem, fet.order, i, p, add_p_level);
+}
 
 
 template <>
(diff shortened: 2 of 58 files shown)


BalticPinguin:
... Ok; that din't work that well: Without infinite elements, in the following 't' is not used:
```
FEInterface::shape_function(const unsigned int dim,
                            const FEType & fe_t,
                            const ElemType t)
{
  if (is_InfFE_elem(t))
   {
    inf_fe_switch(shape);
   }
#endif
  fe_switch(shape);
}

```
The straight-forward idea would be to make
```
FEInterface::shape_function(const unsigned int dim,
                            const FEType & fe_t,
                            const ElemType
                                           t
#endif
                                              )
{
  if (is_InfFE_elem(t))
   {
    inf_fe_switch(shape);
   }
#endif
  fe_switch(shape);
}

```
which is unreadable.

Any good ideas how to handle this?
 Probably creating a ``libmesh_inf_var(t)``-macro, similar to the ``libmesh_dbg_var()`` one?

jwpeterson:
> Probably creating a libmesh_inf_var(t)-macro, similar to the libmesh_dbg_var() one?

Sounds like a good idea to me.

BalticPinguin:
... btw: In the tests the functions ``LIBMESH_ASSERT_FP_EQUAL`` call not as in the [documentation](http://cppunit.sourceforge.net/doc/cvs/group___assertions.html) first the expected, than the actual but other way round. So the messages one gets are
```
double equality assertion failed
- Expected: 1.60000000000001
- Actual  : 2
- Delta   : 0.001
```
with expected and actual being swaped.
I'm not sure if you noticed that already?

jwpeterson:
> Probably you can clarify this for me?

Yeah, I'm sure we definitely call the macro in the wrong order in several places, I know I can never remember which is supposed to be first. That would be great if you could update any that you come across... it might also help remind people if, in the definition of the `LIBMESH_ASSERT_FP_EQUAL`, instead of "first" and "second" we named the parameters "expected" and "actual" respectively.

BalticPinguin:
>it might also help remind people if, in the definition of the LIBMESH_ASSERT_FP_EQUAL, instead of "first" and "second" we named the parameters "expected" and "actual" respectively.

... sounds like a good idea. I can implement it.

But what I'd still like to understand is:
How do I run these tests? ``make check`` seems not to run them in my environment.

jwpeterson:
> But what I'd still like to understand is:
> How do I run these tests? `make check` seems not to run them in my environment.

Most likely it means that cppunit was not found by libmesh's configure script. If you have a recent config.log file laying around, you should be able to find a line in there which says
```
checking for cppunit-config
```
followed by the reason it failed.

Also, in the configure output summary, you might see something like 
```
cppunit..........................no
```
but this does not get written to config.log

BalticPinguin:
```
configure:42086: checking for cppunit-config
configure:42114: result: none
```
great!  Thank you very much.

roystgnr:
I created libmesh_dbg_var and libmesh_mpi_var, and libmesh_inf_var works well along those lines, but note that if you get in any more complicated situations (should we create libmesh_inf_with_mpi_but_not_dbg_var?) you can always just throw a ```libmesh_ignore(var1, var2);``` somewhere in the body of the function and the compiler will treat those variables as "used" despite the use compiling to a no-op.

We've been putting CppUnit arguments in the wrong order since back when we started using CppUnit, IIRC.  At some point I'd love to just go through the whole unit test suite and swap them all, but short of that I haven't felt motivated to put just a handful at a time in the correct order, since the bug leads to confusion but never to false positives or negatives.

roystgnr:
That's an interesting set of failing unit tests right now.  I'd say look at FETest_FIRST_MONOMIAL_EDGE2 first: since every single second derivative in that case ought to be 0, it ought to be very easy to dig into exactly where that -2 is coming from.

BalticPinguin:
> I'd say look at FETest_FIRST_MONOMIAL_EDGE2 first: since every single second derivative in that case ought to be 0, it ought to be very easy to dig into exactly where that -2 is coming from.

you are right; this is a good idea. As soon as my UnitTest lib is working, I'll do.

BalticPinguin:
I guess I need some more help on running the testes: Finally I have them around, but so-far, I can only run all of them at once, which takes quite some time. But the output of 
```
Will run the following tests:
All Tests
```
implies that one can run them separately; but how do I accomplish that? (in best case with option to run them in gdb...)
Many thanks!

roystgnr:
To run them in gdb use LIBMESH_RUN (this is also good for mpiexec); to run a subset of tests use LIBMESH_OPTIONS to set the --re (regular expression) option that John added to our unit tests (this is also good for trying different PETSc options).  So in your case try something like:
```
LIBMESH_RUN='libtool --mode=execute gdb --args' LIBMESH_OPTIONS='--re FETest_FIRST_MONOMIAL_EDGE2' make check -C tests/
```

BalticPinguin:
So, I hope with this now all issues are fixed.
Let me know if you find some left-over errors or non-optimal design; before mergeing, probably I should adjust the commits for cleaner history?

src/fe/fe_bernstein_shape_1D.C
```diff
@@ -465,13 +492,13 @@ Real FE<1,BERNSTEIN>::shape_second_deriv(const ElemType,
           return -(5./4.)*pow<3>(1.-xi) + (15./8.)*(1.+xi)*pow<2>(1.-xi);
         case 3:
           return -(15./ 4.)*(1.+xi)*pow<2>(1.-xi) + (5./ 8.)*pow<3>(1.-xi)
-          + (15./8.)*pow<2>(1.+xi)*(1.-xi);
+            + (15./8.)*pow<2>(1.+xi)*(1.-xi);
```
roystgnr: What's going on with the whitespace changes here?

src/fe/fe_rational_shape_1D.C
```diff
@@ -139,7 +137,7 @@ Real FE<1,RATIONAL_BERNSTEIN>::shape_deriv(const Elem * elem,
       elem->node_ref(n).get_extra_datum<Real>(datum_index);
 
   Real weighted_shape_i = 0, weighted_sum = 0,
-       weighted_grad_i = 0, weighted_grad_sum = 0;
+    weighted_grad_i = 0, weighted_grad_sum = 0;
```
roystgnr: Another few whitespace changes here

src/fe/fe_rational_shape_2D.C
```diff
@@ -134,7 +132,7 @@ Real FE<2,RATIONAL_BERNSTEIN>::shape_deriv(const Elem * elem,
       elem->node_ref(n).get_extra_datum<Real>(datum_index);
 
   Real weighted_shape_i = 0, weighted_sum = 0,
-       weighted_grad_i = 0, weighted_grad_sum = 0;
+    weighted_grad_i = 0, weighted_grad_sum = 0;
```
roystgnr: And here

src/fe/inf_fe_map.C
```diff
@@ -203,11 +203,20 @@ Point InfFEMap::inverse_map (const unsigned int dim,
       libmesh_error_msg("Invalid dim = " << dim);
     }
 
+#ifdef DEBUG
+  // In debug mode, the validity of base_elems id is checked.
```
roystgnr: This is a little surprising.  We do assert validity in DofObject::id() ... but where in inverse_map do we end up *querying* an id?  Just in the error messages when the mapping doesn't converge?

src/fe/inf_fe_map.C
```diff
@@ -203,11 +203,20 @@ Point InfFEMap::inverse_map (const unsigned int dim,
...
+#endif
+
   // 3.)
   // Now we have the intersection-point (projection of physical point onto base-element).
   // Lets compute its internal coordinates (being p(0) and p(1)):
   p= FEMap::inverse_map(dim-1, base_elem.get(), intersection,
                         tolerance, secure);
+#ifdef DEBUG
+  base_elem->invalidate_id();
+#endif
```
roystgnr: This is a temporary proxy Elem that's getting deleted, isn't it?  So do we need to bother invalidating its id?

src/fe/inf_fe_map.C
roystgnr: And in any case, won't the same test end up happening in devel mode?  So instead of ifdef DEBUG we'd want ifndef NDEBUG?

src/fe/inf_fe_static.C
```diff
@@ -471,40 +489,40 @@ void InfFE<Dim,T_radial,T_map>::compute_data(const FEType & fet,
           data.shape[i] = (InfFERadial::decay(Dim,v)                                    /* (1.-v)/2. in 3D          */
                            * FEInterface::shape(Dim-1, fet, base_el.get(), i_base, p)   /* S_n(s,t)                 */
                            * InfFE<Dim,T_radial,T_map>::eval(v, o_radial, i_radial))    /* L_n(v)                   */
-                           * time_harmonic;                                             /* e^(i*k*phase(s,t,v)      */
+            * time_harmonic;                                             /* e^(i*k*phase(s,t,v)      */
```
roystgnr: Fairly extreme whitespace changes here.

roystgnr:
I'd say merge those "fixed" commits into the commits they're fixing; likewise with the remaining whitespace reversions and any other minor fixes.

IMHO this looks ready to merge after those minor bits are taken care of, and I have to thank you so much for figuring out how to fix my mess but I also have to warn you: unless we also add some new unit tests, or some new assertions to our infinite element examples, *something* to improve the test coverage gaps that let InfFE get broken without a peep from CI to begin with, I'm afraid I'm going to break this code again some day.  Possibly soon; there are further FE/InfFE optimizations that have been creeping up my to-do list.  I'll try my best not to break it again, but I'd feel a lot more confident in my efforts if make check was ready to yell at me and moosebuild ready to block me from merging if I slip up.

src/fe/inf_fe_map.C
BalticPinguin: yes, it is queried only in case of non-convergence. But than the program stops; this might be anoying because if one point does not converge (e.g. in postprocessing), this is not a big issue...

src/fe/inf_fe_map.C
BalticPinguin: ... I wanted to be accurate.
But I agree, we can skip it.

BalticPinguin:
Oh, thank you for the review; the white spaces should have been introduced by ``contrib/bin/reindent.sh`` and I did'nt go though it carefully.

Anyway, here is the cleaned history; I just hope, I didn't mess up something by this...

BalticPinguin:
>unless we also add some new unit tests, or some new assertions to our infinite element examples, something to improve the test coverage gaps that let InfFE get broken without a peep from CI

yes, I have already started adding more checks, even though not yet for the ``FEInterface``-class; but I hope to improve on this step by step.

src/fe/inf_fe_map.C
roystgnr: Oh, it's definitely an issue, and you put in the right fix; I'd just wanted to make sure I understood the issue.

roystgnr:
Looks great; thanks again!

PR #2410, opened 2020-01-23
Create parallel backing vector for DiagonalMatrix

lindsayad:
Before I was calling the backing NumericVector init method with just a global N argument which lead to creation of a serial vector. Now I pass an additional argument corresponding to the local size (taken from the DofMap) that will automatically create a parallel vector when the global size and the local size are different.

EDIT: The first change was not enough. https://github.com/libMesh/libmesh/pull/2410/commits/1f4a32dd8c58d21dc5bec6cb448f3e3611ed9962 alone is susceptible to a corner case where a proc owns all dofs. In that case that proc would create a `SERIAL` type while all other procs would create a `PARALLEL` type, leading to hangs at the first attempted communication. So https://github.com/libMesh/libmesh/pull/2410/commits/b74f36e47f4f86c5d280ce3c7c2f64d2f821b50b mirrors what is done for System::add_vector. We also keep track of the matrix types in a _matrix_types data member. This also adds ParallelType as a default argument to the formerly zero arg SparseMatrix::init method. For almost all concrete SparseMatrix data types this parameter will be unused, but its useful for DiagonalMatrix and it could be useful for other derived matrix types in the future as it is useful for vectors now.

### DIFF
--- a/include/numerics/diagonal_matrix.h
+++ b/include/numerics/diagonal_matrix.h
@@ -80,7 +80,7 @@ public:
             const numeric_index_type noz = 10,
             const numeric_index_type blocksize = 1) override;
 
-  void init() override;
+  void init(ParallelType type = PARALLEL) override;
 
   /**
    * Initialize with a NumericVector \p other, e.g. duplicate the storage
--- a/include/numerics/eigen_sparse_matrix.h
+++ b/include/numerics/eigen_sparse_matrix.h
@@ -91,7 +91,7 @@ public:
                      const numeric_index_type noz=10,
                      const numeric_index_type blocksize=1) override;
 
-  virtual void init () override;
+  virtual void init (ParallelType = PARALLEL) override;
 
   virtual void clear () override;
 
--- a/include/numerics/laspack_matrix.h
+++ b/include/numerics/laspack_matrix.h
@@ -101,7 +101,7 @@ public:
                      const numeric_index_type noz=10,
                      const numeric_index_type blocksize=1) override;
 
-  virtual void init () override;
+  virtual void init (ParallelType = PARALLEL) override;
 
   virtual void clear () override;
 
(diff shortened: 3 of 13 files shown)


lindsayad:
@permcody This will be the long term fix for the parallel failure on idaholab/moose#14422, but I'll drum up a short-term MOOSE fix in the mean-time

lindsayad:
There's a little more I want to do for this PR, so don't merge yet

roystgnr:
Certainly looks straightforward so far.

lindsayad:
Ok, this should be ready for review now

lindsayad:
I think there’s some value in labeling it const per https://stackoverflow.com/a/6125466

roystgnr:
If a passed-by-value parameter can be const there's value labeling it const, but there's also a cost: if you ever want to *change* that constness to use a different algorithm, even though that *ought* to be a purely internal change, C++ thinks it counts as an API change, which means you either have to change the header too or you have to make an unnecessary internal copy of the parameter and hope it gets optimized away.

roystgnr:
So I'm happy either way; just saying it's not a purely one-sided tradeoff.

lindsayad:
Valid point....I’ll go ahead and make it non-const

boyceg:
> If a passed-by-value parameter can be const there's value labeling it const, but there's also a cost: if you ever want to _change_ that constness to use a different algorithm, even though that _ought_ to be a purely internal change, C++ thinks it counts as an API change, which means you either have to change the header too or you have to make an unnecessary internal copy of the parameter and hope it gets optimized away.

For pass-by-value, I believe that you are free to make the parameter non-`const` in the header but `const` in the implementation. (I think that the function signature is the same whether the parameter is `const` or not.)

lindsayad:
Indeed! We can get the best of both worlds then :-)

lindsayad:
Anyone wanna review this fellow?

PR #2411, opened 2020-01-24
Copy interior parents during copy construction

lindsayad:
Interior parent data is necessary for a multiapp and mortar
test of @YaqiWang's


lindsayad:
Hmm, that's fun. My own MOOSE tests are failing :laughing:

YaqiWang:
Sounds like I am making mortar implementation more robust ;-)

PR #2414, opened 2020-01-27
Fix syntax error in reduced_basis ex7.

BalticPinguin:
I just found this error (I think I introduced it myself)...
I think, it should usually if not always fail to run with ``Error: Solver  is unknown``.

So I used the chance to also try for hypre-solver; I hope, this does not break in case of some other configuration.


BalticPinguin:
well, I have the options for both the recent PETSc-version and the older ones.
Consider we run an installation with recent PETSc, we check if the options for old PETSc were given. If so, we stop with a respective message.
But in case we don't stop there, (i.e. we have the options given for recent PETSc) we need to seek for these recent options to have the cursor correctly.

PR #2416, opened 2020-01-28
Allow GenericProjector to work with vector finite elements

lindsayad:
@jiangwen84 The work here should solve https://github.com/idaholab/moose/issues/13030

### DIFF
--- a/include/numerics/tensor_tools.h
+++ b/include/numerics/tensor_tools.h
@@ -202,9 +202,10 @@ struct MakeNumber
 template <typename T>
 struct MakeNumber<std::complex<T>>
 {
-  // Compile-time error: we shouldn't need to make numbers out of
-  // numbers
-  //typedef std::complex<T> type;
+  // Should this be a compile-time error? we shouldn't need to make numbers out of
+  // numbers, but then again having the typedef below enables more generic
+  // programming
+  typedef std::complex<T> type;
 };
 
 
@@ -305,6 +306,60 @@ Number div_from_grad( const TensorValue<Number> & grad );
 /*! Place holder needed for ExactSolution to compile. Will compute the
   divergence of a tensor given the gradient of that tensor. */
 VectorValue<Number> div_from_grad( const TypeNTensor<3,Number> & grad );
+
+/**
+ * This helper structure is used to determine whether a template class is one of
+ * our mathematical structures, like TypeVector, TypeTensor and their descendents
+ */
+template <typename T>
+struct MathWrapperTraits
+{
+  static constexpr bool value = false;
+};
+
+template <typename T>
+struct MathWrapperTraits<TypeVector<T>>
+{
+  static constexpr bool value = true;
+};
+
+template <typename T>
+struct MathWrapperTraits<VectorValue<T>>
+{
+  static constexpr bool value = true;
+};
+
+template <typename T>
+struct MathWrapperTraits<TypeTensor<T>>
+{
+  static constexpr bool value = true;
+};
+
+template <typename T>
+struct MathWrapperTraits<TensorValue<T>>
+{
+  static constexpr bool value = true;
+};
+
+template <unsigned int N, typename T>
+struct MathWrapperTraits<TypeNTensor<N,T>>
+{
+  static constexpr bool value = true;
+};
+
+template <typename T, typename enable = void> struct MakeBaseNumber {};
+
+template <typename T>
+struct MakeBaseNumber<T, typename std::enable_if<ScalarTraits<T>::value>::type> {
+  typedef typename MakeNumber<T>::type type;
+};
+
+template <template <typename> class Wrapper, typename T>
+struct MakeBaseNumber<
+    Wrapper<T>,
+    typename std::enable_if<MathWrapperTraits<Wrapper<T>>::value>::type> {
+  typedef typename MakeBaseNumber<T>::type type;
+};
 
 }//namespace TensorTools
 
--- a/include/systems/generic_projector.h
+++ b/include/systems/generic_projector.h
@@ -38,6 +38,8 @@
 #include "libmesh/system.h"
 #include "libmesh/threads.h"
 #include "libmesh/auto_ptr.h" // libmesh_make_unique
+#include "libmesh/tensor_tools.h"
+#include "libmesh/libmesh_common.h"
 
 // TIMPI includes
 #include "timpi/parallel_sync.h"
@@ -142,7 +144,7 @@ public:
   // We generally need to hang on to every value we've calculated
   // until we're all done, because later projection calculations
   // depend on boundary data from earlier calculations.
-  std::unordered_map<dof_id_type, FValue> ids_to_save;
+  std::unordered_map<dof_id_type, typename FFunctor::ValuePushType> ids_to_save;
 
   // This needs to be sorted so we can get sorted dof indices cheaply
   // later
@@ -162,11 +164,13 @@ public:
     // We keep track of which dof ids we might need to send, and what
     // values those ids should get (along with a pprocessor_id to
     // leave invalid in case *we* can't compute those values either).
-    std::unordered_map<dof_id_type, std::pair<FValue, processor_id_type>> new_ids_to_push;
+    std::unordered_map<dof_id_type,
+                       std::pair<typename FFunctor::ValuePushType,
+                                 processor_id_type>> new_ids_to_push;
 
     // We'll hang on to any new ids to save on a per-thread basis
     // because we won't need them until subsequent phases
-    std::unordered_map<dof_id_type, FValue> new_ids_to_save;
+    std::unordered_map<dof_id_type, typename FFunctor::ValuePushType> new_ids_to_save;
 
     // Helper function for filling new_ids_to_push
     void find_dofs_to_send (const Node & node,
@@ -176,12 +180,32 @@ public:
 
     // When we have new data to act on, we may also need to save it
     // and get ready to push it.
-    void insert_id(dof_id_type id,
-                   const FValue & val,
-                   processor_id_type pid);
+    template <typename InsertInput,
+              typename std::enable_if<
+                  std::is_same<typename ProjectionAction::InsertInput, InsertInput>::value,
+                  int>::type = 0>
+    void insert_id(dof_id_type id, const InsertInput & val, processor_id_type pid);
 
+    template <typename InsertInput,
+              typename std::enable_if<
+                  !std::is_same<typename ProjectionAction::InsertInput, InsertInput>::value,
+                  int>::type = 0>
+    void insert_id(dof_id_type id, const InsertInput & val, processor_id_type pid);
+
+    template <typename InsertInput,
+              typename std::enable_if<
+                  std::is_same<typename ProjectionAction::InsertInput, InsertInput>::value,
+                  int>::type = 0>
     void insert_ids(const std::vector<dof_id_type> & ids,
-                    const std::vector<FValue> & vals,
+                    const std::vector<InsertInput> & vals,
+                    processor_id_type pid);
+
+    template <typename InsertInput,
+              typename std::enable_if<
+                  !std::is_same<typename ProjectionAction::InsertInput, InsertInput>::value,
+                  int>::type = 0>
+    void insert_ids(const std::vector<dof_id_type> & ids,
+                    const std::vector<InsertInput> & vals,
                     processor_id_type pid);
 
     // Our subclasses will need to merge saved ids
@@ -199,6 +223,7 @@ public:
 
     // These get used often enough to cache them
     std::vector<FEContinuity> conts;
+    std::vector<FEFieldType> field_types;
 
     const System & system;
   };
@@ -219,6 +244,7 @@ public:
     using SubFunctor::system;
     using SubFunctor::context;
     using SubFunctor::conts;
+    using SubFunctor::field_types;
     using SubFunctor::insert_id;
     using SubFunctor::insert_ids;
 
@@ -231,7 +257,7 @@ public:
        const std::vector<unsigned int> & involved_dofs,
        unsigned int var_component,
        const Node * node,
-       const FEBase & fe);
+       const FEGenericBase<typename FFunctor::RealType> & fe);
   };
 
 
@@ -275,10 +301,10 @@ public:
     using SubProjector::system;
     using SubProjector::context;
     using SubProjector::conts;
+    using SubFunctor::field_types;
 
     using SubProjector::insert_id;
     using SubProjector::insert_ids;
-    using SubProjector::construct_projection;
 
     void operator() (const node_range & range);
   };
@@ -295,10 +321,10 @@ public:
     using SubProjector::system;
     using SubProjector::context;
     using SubProjector::conts;
+    using SubFunctor::field_types;
 
     using SubProjector::insert_id;
     using SubProjector::insert_ids;
-    using SubProjector::construct_projection;
 
     void operator() (const node_range & range);
   };
@@ -314,10 +340,10 @@ public:
     using SubProjector::system;
     using SubProjector::context;
     using SubProjector::conts;
+    using SubFunctor::field_types;
 
     using SubProjector::insert_id;
     using SubProjector::insert_ids;
-    using SubProjector::construct_projection;
 
     void operator() (const node_range & range);
   };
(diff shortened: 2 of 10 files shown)


roystgnr:
Oh, this is fantastic.  Give me a little time to review it.

In the meantime, though - could you add a unit test exercising it?  Or maybe add some projection work plus assertions to one of the vector fe example codes?

This code has to deal with an incredible number of corner conditions (both metaphorical and literal!), and it's a performance hotspot, so I'm definitely going to be refactoring it again and again in the future, and I really don't want to accidentally revert this capability when I do.

include/systems/generic_projector.h
```diff
@@ -2122,9 +2119,14 @@ void GenericProjector<FFunctor, GFunctor, FValue, ProjectionAction>::ProjectEdge
...
-            (dof_indices, edge_dofs, var_component,
-             &edge_node, proj_fe);
+          if (FEInterface::field_type(proj_fe.get_fe_type()) == TYPE_SCALAR)
+            this->construct_projection
+              (dof_indices, edge_dofs, var_component,
+               &edge_node, static_cast<const FEGenericBase<Real> &>(proj_fe));
+          else
+            this->construct_projection
+              (dof_indices, edge_dofs, var_component,
+               &edge_node, static_cast<const FEGenericBase<RealVectorValue> &>(proj_fe));
```
roystgnr: Could we use libmesh_cast here?  I can't see these casts failing but it'd be nice to double check when debugging anyway.

include/systems/generic_projector.h
```diff
@@ -2284,9 +2286,14 @@ void GenericProjector<FFunctor, GFunctor, FValue, ProjectionAction>::ProjectSide
...
-            (dof_indices, side_dofs, var_component,
-             &side_node, proj_fe);
+          if (FEInterface::field_type(proj_fe.get_fe_type()) == TYPE_SCALAR)
+            this->construct_projection
+              (dof_indices, side_dofs, var_component,
+               &side_node, static_cast<const FEGenericBase<Real> &>(proj_fe));
+          else
+            this->construct_projection
+              (dof_indices, side_dofs, var_component,
+               &side_node, static_cast<const FEGenericBase<RealVectorValue> &>(proj_fe));
```
roystgnr: Likewise libmesh_cast here if possible?

include/systems/generic_projector.h
```diff
@@ -2405,9 +2412,14 @@ void GenericProjector<FFunctor, GFunctor, FValue, ProjectionAction>::ProjectInte
...
-            (dof_indices, all_dofs, var_component,
-             nullptr, *fe);
+          if (FEInterface::field_type(fe->get_fe_type()) == TYPE_SCALAR)
+            this->construct_projection
+              (dof_indices, all_dofs, var_component,
+               nullptr, static_cast<const FEGenericBase<Real> &>(*fe));
+          else
+            this->construct_projection
+              (dof_indices, all_dofs, var_component,
+               nullptr, static_cast<const FEGenericBase<RealVectorValue> &>(*fe));
```
roystgnr: And here.

include/systems/generic_projector.h
```diff
@@ -2573,17 +2585,20 @@ GenericProjector<FFunctor, GFunctor, FValue,
...
    const std::vector<unsigned int> & involved_dofs,
    unsigned int var_component,
    const Node * node,
-   const FEBase & fe)
+   const FEGenericBase<OutputShape> & fe)
 {
-  const std::vector<Real> & JxW = fe.get_JxW();
-  const std::vector<std::vector<Real>> & phi = fe.get_phi();
-  const std::vector<std::vector<RealGradient>> * dphi = nullptr;
+  typedef FEGenericBase<OutputShape> FEBase;
```
roystgnr: I know reusing "FEBase" makes the diff a lot shorter, but could we pick a new name to use here?  I'm surprised this doesn't trigger -Wshadow already, and I don't want to be caught out as soon as a compiler gets more thorough about that warning.

lindsayad: 👍

lindsayad:
So @roystgnr this is getting a little more in-depth than I was hoping. The assumption that the finite element solution and the degrees of freedom have the same type is baked into a lot of places. Consider this code:
<pre>
template<>
inline
Number
OldSolutionValue<Number, &FEMContext::point_value, Number>::
eval_at_node(const FEMContext & c,
             unsigned int i,
             unsigned int /* elem_dim */,
             const Node & n,
             bool extra_hanging_dofs,
             Real /* time */)
{
  LOG_SCOPE ("Number eval_at_node()", "OldSolutionValue");

  // This should only be called on vertices
  libmesh_assert_less(c.get_elem().get_node_index(&n),
                      c.get_elem().n_vertices());

  // Handle offset from non-scalar components in previous variables
  libmesh_assert_less(i, this->component_to_var.size());
  unsigned int var = this->component_to_var[i];

  // Optimize for the common case, where this node was part of the
  // old solution.
  //
  // Be sure to handle cases where the variable wasn't defined on
  // this node (due to changing subdomain support) or where the
  // variable has no components on this node (due to Elem order
  // exceeding FE order) or where the old_dof_object dofs might
  // correspond to non-vertex dofs (due to extra_hanging_dofs and
  // refinement)

  const Elem::RefinementState flag = c.get_elem().refinement_flag();

  if (n.old_dof_object &&
      (!extra_hanging_dofs ||
       flag == Elem::JUST_COARSENED ||
       flag == Elem::DO_NOTHING) &&
      n.old_dof_object->n_vars(sys.number()) &&
      n.old_dof_object->n_comp(sys.number(), var))
    {
      const dof_id_type old_id =
        n.old_dof_object->dof_number(sys.number(), var, 0);
      <b>return old_solution(old_id);</b>
    }

  <b>return this->eval_at_point(c, i, n, 0, false);</b>
}
</pre>
In the first return we are accessing the `NumericVector` which is always going to hold `Number/Real` whereas in the second return we are going to evaluate at a point, which most of the time is going to call through to the templated `FEMContext::point_value` which is going to set the value according to whether it is a standard or vector finite element basis type. Do you have a suggestion for cleanest way to refactor this code?

include/systems/generic_projector.h
```diff
@@ -2634,13 +2649,13 @@ GenericProjector<FFunctor, GFunctor, FValue, ProjectionAction>::SubProjector::co
   for (unsigned int qp=0; qp<n_qp; qp++)
     {
       // solution at the quadrature point
-      FValue fineval = f.eval_at_point(context,
+      auto fineval = f.eval_at_point(context,
```
roystgnr: After much staring I find I'm still confused.  The FValue -> auto switch here is because FValue is still being set as Number, where as f's return value is now Gradient?  If that's the case then we definitely shouldn't still be using the name "FValue"; that's just anti-self-documenting.

What I think I'd prefer is to change FValue to fit the return value type of f, then use TensorTools::MakeNumber to get the other (DoFValue?) type that in the non-vector-case is synonymous with FValue.

roystgnr:
Speaking of DoFValue-vs-FValue distinctions ...

Here we don't just need to worry about clean refactoring, but about clean-and-fast refactoring.  All the complicated hacks here are basically trying to drop us into cheap special cases as often as we can get away with that, even though the general code paths would get us to the same answer, because the general code paths get there more slowly.

So obvious refactors (just return type FValue in each case, but only use the first component of it in some cases...) aren't obviously good ideas.

I think what we might want to do first is pull out the nodal interpolation entirely: get that massive hunk of HERMITE-specific code out of the main algorithm, and then also use the same calling site for whatever vector-valued-specific code we want to use in that case.

After that, we can just do two more template specializations on eval_at_node (or damn, four more specializations? for the projection matrix calculation case); we've already got Number+point_value and Gradient+point_gradient; we'd just add Gradient+point_value and Tensor+point_gradient.

Not sure this is the best way to go (we've now got 2^n specializations where I'd hope we could compose n bits of code, which is annoying even when n is only 3) but it's the first idea that comes to mind.

lindsayad:
> I think what we might want to do first is pull out the nodal interpolation entirely: get that massive hunk of HERMITE-specific code out of the main algorithm, and then also use the same calling site for whatever vector-valued-specific code we want to use in that case.

And put it where?

include/systems/generic_projector.h
lindsayad: > After much staring I find I'm still confused. The FValue -> auto switch here is because FValue is still being set as Number, where as f's return value is now Gradient? If that's the case then we definitely shouldn't still be using the name "FValue"; that's just anti-self-documenting.
> 
> What I think I'd prefer is to change FValue to fit the return value type of f, then use TensorTools::MakeNumber to get the other (DoFValue?) type that in the non-vector-case is synonymous with FValue.

Do we have documented what the `F` in `FFunctor` and `G` in `GFunctor` mean? I'm supplying `OldSolutionValue` as a functor. My best guess is that `F` should stand for finite element solution (value) *function* and `G` shold stand for finite element solution *gradient*? Or are `G` and `F` just supposed to stand for generic `f(x)`, `g(x)`?

roystgnr: Function and Gradient, I think, but honestly this dates back to when I first added C1 element support, and I have no personal recollection at this point.

roystgnr:
Well, if I could figure out the *right* way to refactor things, such that the DSNA code could piggy-back off the plain code, then the plain code would definitely belong in generic_projector.h, even for the Hermite- and vector- specific bits.

As it is, I can't, so even a lot of the existing code's presence in generic_projector.h is a case of me being too lazy to set up a generic_projector.C, in which we ought to have any specialization that no longer has a free template argument and that's too long to inline.  Right now that's just two eval_at_node specializations, but if that bumps up to four specializations then the laziness is becoming even more blatant.

include/systems/generic_projector.h
lindsayad: > What I think I'd prefer is to change FValue to fit the return value type of f, then use TensorTools::MakeNumber to get the other (DoFValue?) type that in the non-vector-case is synonymous with FValue.

This is fine with me. However, I'll probably add another helper struct in `TensorTools` for determining `DofValue` since `MakeNumber<Gradient>` would just return `Gradient`, e.g. it maintains the rank of the incoming structure (vector -> vector, tensor -> tensor). Wait...there should be no reason for an additional `DofValue` type...that type should always just be `Number`...

roystgnr: > However, I'll probably add another helper struct in TensorTools for determining DofValue since MakeNumber<Gradient> would just return Gradient

Yeah, I'm an idiot; I forgot that was the "upgrade Real->Number" function; I *thought* I had a "DecrementRank until you can't Decrement it no more" function but I guess not.

> Wait...there should be no reason for an additional DofValue type...that type should always just be Number...

Nope.  About a third of the reason for doing the GenericProjector refactoring in the first place was to let it handle more complicated things than Number, like the DynamicSparseNumberArray case in projection_matrix().

lindsayad:
> After that, we can just do two more template specializations on eval_at_node (or damn, four more specializations? for the projection matrix calculation case); we've already got Number+point_value and Gradient+point_gradient; we'd just add Gradient+point_value and Tensor+point_gradient.

So I guess moving forward on this PR, I'm going to look at crafting these new specializations. Overall this `GenericProjector` code seems pretty impressive, and I am no where close to understanding it all. I have a few possibly stupid questions:

1. Why wouldn't we want to maintain `FValue` as the return type of `eval_at_node`, even in the vector case? In what cases would we only want to use some of the dof values returned?
2. If we were to maintain `FValue` as the return type of `eval_at_*`, then maybe I'll also need to specialize `eval_at_point`? The answer to this question would probably  fall out to me after solving question 1.
3. I would describe my understanding of projection vs. interpolation as tenuous. I believe that we perform interpolation for nodal bases and projection otherwise? Please correct me if I'm wrong. This question is motivated primarily by your quote below. Don't we perform interpolation for families other than `HERMITE`? Even for vanilla `LAGRANGE`? 

> I think what we might want to do first is pull out the nodal interpolation entirely: get that massive hunk of HERMITE-specific code out of the main algorithm, and then also use the same calling site for whatever vector-valued-specific code we want to use in that case.

include/systems/generic_projector.h
lindsayad: It seems unfortunate that when doing projection with `OldSolutionCoefs` as the functor that `FValue` does not correspond to the `Output` template argument which I think helped lead to my confusion. But we have decided...`FValue` should correspond to the return type of `eval_at_node`/`eval_at_point`, and not to the first template argument to `FFunctor` which corresponds to the `Output` parameter provided to `FEMContext::point_value`. Correct?

lindsayad: <pre>
template <typename Output,
          void (FEMContext::*point_output) (unsigned int,
                                            const Point &,
                                            Output &,
                                            const Real) const>
class OldSolutionCoefs : public OldSolutionBase<Output, point_output>
{
public:
  <b>typedef typename DSNAOutput&ltOutput&gt::type DSNA;</b>
</pre>

I like what you did here with the `DSNAOutput struct`

roystgnr:
No probably with stupid questions.  Just forgive possibly-stupid answers.

1.  So, I've trying to keep both LAGRANGE_VEC and NEDELEC_ONE elements in mind here ... but I shouldn't have been.  Nedelec elements aren't even continuous at vertices; we're not going to support them merely by handling vector element projections, we're going to need a whole new continuity() case.  In which case... I suppose we really do want to return FValue from eval_at_node, don't we?  We're always going to have d dof values there for a C0 vector element in R^d, aren't we?

2.  eval_at_point should definitely be returning FValue==Gradient in the vector case, shouldn't it?  And in this case it looks to me like OldSolutionValue::eval_at_point ought to still be correct.

3.  Here we're always talking about "projection" in a mathematical sense of "an idempotent operator", not specifically an L2 projection or an H1 projection or whatever.  In *general*, what we do for any element is we make the projection calculations as local as possible: we interpolate any DoFs at nodes, then we do a local Hilbert projection (L2 for C0 or H1 for C1) values on edges while holding node DoFs fixed, then on faces while holding edges and nodes fixed, then on interiors while holding faces edges and nodes fixed.  The exception to this is Lagrange, where we do interpolations instead of projections for the sake of backwards compatibility and speed.

include/systems/generic_projector.h
roystgnr: I think that's correct.  FValue ought to correspond to the output of f, and eval_at_* are just ways of getting that value.  But FFunctor doesn't even technically have to be a templated class, much less be templated around an argument matching FValue.

lindsayad:
>     1. We're always going to have d dof values there for a C0 vector element in R^d, aren't we?

Yes, I think so!

>     3\. In _general_, what we do for any element is we make the projection calculations as local as possible: we interpolate any DoFs at nodes, then we do a local Hilbert projection (L2 for C0 or H1 for C1) values on edges while holding node DoFs fixed, then on faces while holding edges and nodes fixed, then on interiors while holding faces edges and nodes fixed.  The exception to this is Lagrange, where we do interpolations instead of projections for the sake of backwards compatibility and speed.

Ok this is a great explanation, thank you. Is it more accurate to substitute "vertex" for "node" here?

roystgnr:
Oh, sorry - yes, I *definitely* meant to say "vertex" there; non vertex nodes are only interpolated in the Lagrange case.

lindsayad:
Yea, @jiangwen84 it looks like your example from over in idaholab/moose##13030 runs well now (I did remove one one of the initial adaptivity steps so that the non-linear residual so that there was actual adaptivity after the first solve)

jiangwen84:
@lindsayad thanks for your PR! I cannot wait to give it a shot. @roystgnr thank you too!

lindsayad:
@roystgnr I think this is ready for review

include/numerics/tensor_tools.h
```diff
@@ -306,6 +307,56 @@ Number div_from_grad( const TensorValue<Number> & grad );
   divergence of a tensor given the gradient of that tensor. */
 VectorValue<Number> div_from_grad( const TypeNTensor<3,Number> & grad );
 
+template <typename T>
```
roystgnr: Could we get a comment here?  I assume this is to distinguish our math classes to allow template specialization based on them, but that's not obvious from the name.

include/systems/generic_projector.h
```diff
@@ -1179,17 +1345,37 @@ void GenericProjector<FFunctor, GFunctor, FValue, ProjectionAction>::SubFunctor:
...
+              !std::is_same<typename ProjectionAction::InsertInput, InsertInput>::value,
+              int>::type>
+void
+GenericProjector<FFunctor, GFunctor, FValue, ProjectionAction>::SubFunctor::insert_ids(
+    const std::vector<dof_id_type> &,
+    const std::vector<InsertInput> &,
+    processor_id_type)
+{
+  libmesh_error_msg("ID insertion should only occur when the provided input matches that "
+                    "expected by the ProjectionAction");
```
roystgnr: This is weird.  Couldn't you have just left off this specialization, and then any mismatch would have been a compile-time error rather than a run-time error?

lindsayad: If I don't have this, then I get compile errors right now in  `ProjectVertices`, `ProjectEdges`, `ProjectSides`, and `ProjectInteriors` when I instantiate a vector projection because I thought we were open to using the same calling sites for both scalar and vector field variables.

roystgnr: Oh, I get it.  You're choosing which to call with an if test, so we still get an instantiation of the incompatible option, but that's okay because its an instantiation we'll never hit.

roystgnr:
Well, I'm going to be spending weeks kicking myself over whether I gave you the right design advice here, but I certainly still can't think of a better way to do it.  Thanks for all this!

lindsayad:
Well if it isn't right, we can always revert :-)

PR #2419, opened 2020-01-31
fem_system IGA example

roystgnr:
This gives more extensive testing (it's already caught a few bugs) and a more easily reusable example than the unit tests for DynaIO.

I still need to futz with the boundary conditions here, but those bug fixes might affect other codes too so I want to get it into CI.

Refs #2179


jwpeterson:
Did the testing run for this? If not, did you want it to? I don't see any pass/fail status for CI...

jwpeterson:
Would it make sense to put 3250a318a6b536c0479ab8b8ede6e7ccd9c57dce into a separate PR? That one looks like an important bugfix (?) that could potentially get merged before the example is 100% done.

jwpeterson:
Yep, one random payload delivery failure for this PR, I'll try resending it.

fail_2149

jwpeterson:
```
ERROR: The following files contain trailing whitespace after applying your patch:
	examples/fem_system/fem_system_ex3/elasticity_system.h
	examples/fem_system/fem_system_ex5/elasticity_system.h
```

roystgnr:
Yeah, I've fixed that in my copy of the branch; will force-push after a while.  Trying to debug something else first.

roystgnr:
IMHO 3250a31 isn't a bugfix, it's a matter of preference.  I like the consistency of the new assignment better, but it wouldn't be worth making the change except that I've got Rational Bernstein stuff depending on that consistency and it'd be far uglier and harder to add workarounds there.

PR #2423, opened 2020-02-04
Add MeshBase::insert_elem(std::unique_ptr<Elem>)

jwpeterson:
Follow on work to #2420, #2422.

### DIFF
--- a/include/mesh/distributed_mesh.h
+++ b/include/mesh/distributed_mesh.h
@@ -287,6 +287,7 @@ public:
   virtual Elem * add_elem (Elem * e) override;
   virtual Elem * add_elem (std::unique_ptr<Elem> e) override;
   virtual Elem * insert_elem (Elem * e) override;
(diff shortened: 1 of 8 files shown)


roystgnr:
How the heck could this have messed up misc ex11?

jwpeterson:
> How the heck could this have messed up misc ex11?

Not sure but I confirmed on my system that the error is in  https://github.com/libMesh/libmesh/pull/2423/commits/72240a3752bf6732c3d0b541325aa5d3cda69c68. Since it's a subdivision element example, presumably it doesn't like the changes to `MeshTools::Subdivision::all_subdivision()`, but I don't see anything obviously wrong so I will have to take a closer look...

jwpeterson:
There was a bug in one of the `Tri3Subdivision` constructors, specifically in the constructor called by `Elem::build()`, which explains why it was triggered by this PR. Should hopefully be fixed in https://github.com/libMesh/libmesh/pull/2423/commits/1f0b03758f3a34a934059a30e5a060cefdf1f160

PR #2426, opened 2020-02-06
Helper structure for determining DenseMatrix<T>::use_blas_lapack

lindsayad:
Currently in MOOSE we have to create our own header for `DenseMatrix` which declares a specialization for the `DenseMatrix<DualReal>::DenseMatrix`, and then use our own header any place in MOOSE that we want to make use of `DenseMatrix`. This is prone to developer error and can also be prone to issues with header include order in the case that a libMesh include pulls in dense_matrix and we somehow implicitly instantiate `DenseMatrix<DualReal>::DenseMatrix` before we include our own header (unlikely but still).

### DIFF
--- a/include/numerics/dense_matrix.h
+++ b/include/numerics/dense_matrix.h
@@ -554,6 +554,14 @@ public:
    */
   bool use_blas_lapack;
 
+  /**
+   * Helper structure for determining whether to use blas_lapack
+   */
+  struct UseBlasLapack
+  {
+    static const bool value = false;
+  };
+
 private:
 
   /**
@@ -763,9 +771,13 @@ typedef DenseMatrix<Complex> ComplexDenseMatrix;
 
 using namespace DenseMatrices;
 
-
-
-
+#ifdef LIBMESH_HAVE_PETSC
+template <>
+struct DenseMatrix<double>::UseBlasLapack
+{
+  static const bool value = true;
+};
+#endif
 
 // ------------------------------------------------------------
 // Dense Matrix member functions
@@ -774,11 +786,7 @@ inline
 DenseMatrix<T>::DenseMatrix(const unsigned int new_m,
                             const unsigned int new_n) :
   DenseMatrixBase<T>(new_m,new_n),
-#if defined(LIBMESH_HAVE_PETSC) && defined(LIBMESH_USE_REAL_NUMBERS) && defined(LIBMESH_DEFAULT_DOUBLE_PRECISION)
-  use_blas_lapack(true),
-#else
-  use_blas_lapack(false),
-#endif
+  use_blas_lapack(DenseMatrix<T>::UseBlasLapack::value),
   _val(),
   _decomposition_type(NONE)
 {


jwpeterson:
:+1: Looks like a major improvement over the ifdefs that were there before.

dschwen:
So.. how does this change address the implicitly instantiation issue?

lindsayad:
> So.. how does this change address the implicitly instantiation issue?

We will no longer need `DenseMatrix.h` in MOOSE because will no longer need to declare a specialization for `DenseMatrix<DualReal>::DenseMatrix` because the primary template is going to create what we want now, e.g. it will set `use_blas_lapack = true` when `T = double` (when #ifdef LIBMESH_HAVE_PETSC) and `false` otherwise.

dschwen:
Oh, I see, but we'd still have to include the *impl header somewhere, right?

lindsayad:
yea for sure. We'll be able to get rid of `DenseMatrix.h`, but `DenseMatrixDualReal.C` will still exist minus the `DenseMatrix<DualReal>::DenseMatrix` definition

jwpeterson:
I figured out that this PR is causing an issue for us. The problem seems to be that the logic for setting `use_blas_lapack` to true was changed from:
```
#if defined(LIBMESH_HAVE_PETSC) && defined(LIBMESH_USE_REAL_NUMBERS) && defined(LIBMESH_DEFAULT_DOUBLE_PRECISION)
```
to simply
```
#ifdef LIBMESH_HAVE_PETSC
```
but the ifdef logic in dense_matrix_blas_lapack.C itself was not changed. 

In my case, I have a complex numbers build (LIBMESH_USE_REAL_NUMBERS is not defined) but I still create a `DenseMatrix<Real>` and call `lu_solve()` on that. Apparently this used to work because it would fall back on the hand-coded (non-Lapack) `_lu_decompose()` function, but now it tries to call `_lu_decompose_lapack()`. Unfortunately, this function is an error unless:
```
#if (LIBMESH_HAVE_PETSC && LIBMESH_USE_REAL_NUMBERS)
```
which isn't true for my complex numbers build. 

I'm not quite sure exactly what the right fix should be, but I think both the ifdefs currently around `_lu_decompose_lapack()` and the explicit instantiations at the bottom of dense_matrix_blas_lapack.C could use some updating. Also it seems like there was maybe an old  assumption that a `DenseMatrix<Real>` wouldn't ever be instantiated for complex-enabled builds but this seems unnecessarily restrictive (it has been working OK for us up to this point) so that might have to be fixed as well.

roystgnr:
So you've got a ```DenseMatrix<Real>``` which means it's hitting the ```use_blas_lapack=true``` for that ... and you're using lu_solve(), so you're giving it a Real-valued right hand side too since we didn't template that around a separate T2 type ... so this is the sort of scenario that our changed ifdef logic was supposed to be *for*!  Allowing you to use BLAS/Lapack on double even when our default Number wasn't double.

But let's see what's going wrong ... lu_solve() dispatches to _lu_decompose_lapack() ... and _lu_decompose_lapack still uses the old ifdef logic.

So what we want is probably to replace the ifdef logic in dense_matrix_blas_lapack.C, to use partial template specialization instead: _lu_decompose_lapack() should scream and die in the default case, but should be specialized to use BLAS+Lapack in the T==Real case.  Right?

jwpeterson:
That's what I was trying first. I think I just want a specialization of `_lu_decompose_lapack()` for `double` and then comment out the explicit instantiations in dense_matrix_blas_lapack.C entirely? I can't really remember how specialization works in the library context, is a specialization guaranteed to be instantiated?

jwpeterson:
Ah, wait, I'm starting to remember a bit more. It looks like the PETSc Lapack wrappers (LAPACKgetrf_, etc.) that we use are only created for the current `PetscScalar`... in other words in complex-enabled PETSc, `LAPACKgetrf_` wraps `zgetrf` while in real-valued PETSc it wraps `dgetrf`. So I think we do need to fall back on hand-coded LU, etc. in the Number==std::complex case...

roystgnr:
Ah, hell.  The first problem can be worked around (you can only do partial specialization of classes, and there's nothing to do overloading on for this function, so we'd want an anonymous helper class to partially specialize), but if PETSc only gives us PetscScalar, then we only have PetscScalar.  At that point maybe the right thing to do really is to just restore the old ifdef logic.  It's either that or try to detect BLAS+Lapack ourselves, which would be ten times more hassle, and only worth it if this Real-lu_solve-in-complex-build case is a significant part of your runtime.

jwpeterson:
> At that point maybe the right thing to do really is to just restore the old ifdef logic.

This is what I'm testing now. FWIW it should be possible to get e.g. a Lapack-enabled `DenseMatrix<Number>::lu_solve()` when PETSc/libMesh are compiled with complex number support, since in this case the types are not mixed. But that would require some additional work and it's not something we currently need so I'm not planning to try and add it at the moment.

jwpeterson:
> only worth it if this Real-lu_solve-in-complex-build case is a significant part of your runtime.

Calling the hand-coded Real-lu_solve-in-complex-build has never showed up as a performance hotspot for us AFAIK. If it was I'd probably refactor into an implementation that uses an `Eigen` matrix before making direct BLAS/Lapack calls, although the latter isn't out of the question either... just because PETSc only wraps one of `{s,d,z,c}getrf`, it might still be possible to call one of the others directly, I've never tried it.

PR #2427, opened 2020-02-07
Additional PETSc+METIS compatibility options

roystgnr:
These were written to let me build with the Ubuntu 19.10 system PETSc and METIS, but the PETSc configuration fixes might improve compatibility with other installs too, and the option to use a third-party METIS is just generally the right thing to offer users.


roystgnr:
Yay CI.  This is clearly my bug, but I don't understand why it's only in the 64-bit environment that Parmetis configuration would now be failing.

roystgnr:
Never mind, now the other PETSc METIS configs are failing too, and I think I see why.

roystgnr:
Can anyone reproduce that fem_system_ex5 failure?  I cajoled @pbauman into trying on his Mac, but there it ran and completed fine.

pbauman:
I didn't realize that was related to this PR. Note that I used `--with-metis=PETSc` in my build. Is there another configuration I should try? (My entire configure line was `--enable-everything --with-metis=PETSc`)

roystgnr:
Oh, it (almost certainly) wasn't related to this PR.  This PR is just the first time I turned on the optional CI "MOOSE Mac Test" recipe since merging the new example.

Configuration there is
```
../configure --with-methods=opt --prefix=/opt/civet/build_0/libmesh/installed --enable-silent-rules --enable-unique-id --disable-warnings --enable-openmp --disable-maintainer-mode --with-metis=PETSc
```

But I don't see anything there that raises my eyebrows.

roystgnr:
I think I've found the bug, which is indeed a heinous bug anywhere but which dumb luck didn't let us trigger anywhere except Mac.  If this works then I'll move the fix to a new PR.  If *that* works then I'll rebase this post-bugfix-merge, get rid of the debugging commits, and merge this once CI is happy again.

Issue #2431, opened 2020-02-10
Add SparseMarix::clone(), zero_clone() methods


jwpeterson:
Currently there is no way (that I am aware of) to make a copy of an existing `SparseMatrix`. The current approach is to cast to a more derived type e.g. `PetscMatrix` and then call `MatCopy` on the raw `Mat`, which definitely seems like it could be improved. The `clone()` and `zero_clone()` names are consistent with existing `NumericVector` interfaces.

roystgnr:
Yeah, there's no good reason we don't have this interface already, other than "none of the main developers needed it before" and "users were willing to put up with a PetscMatrix-specific hack instead".

jwpeterson:
Added in #2435

PR #2435, opened 2020-02-11
Add SparseMatrix::clone(), zero_clone()

jwpeterson:
The LaspackMatrix and EpetraMatrix implementations are currently
marked as libmesh_not_implemented(), although they include a possible
implementation. In case someone ever needs this capability, they could
use this code as a starting point.

Refs #2431


jwpeterson:
I will add some unit tests for this... at the moment I am just testing whether it even compiles in our non-standard configs.

PR #2438, opened 2020-02-14
libmesh_vector_at

roystgnr:
@jwpeterson pointed out w.r.t. #2437 that std::vector::at() doesn't give a very useful error message, and he's already been using an upgraded replacement, so let's steal that and put it in the library.

This PR uses it in DynaIO in place of the recent at() calls there, but at some point we should go over the other MeshInput subclasses, find places where (except in dbg mode) we're trusting our input files, and fix that.  It'd be nice if someday we got libMesh to pass fuzz testing without segfaults.

### DIFF
--- a/include/utils/utility.h
+++ b/include/utils/utility.h
@@ -35,14 +35,14 @@ namespace libMesh
  * idiom, which is similar to calling map.at(), but gives a more
  * useful error message with a line number.
  */
-#define libmesh_map_find(map, key) Utility::map_find((map), (key), __FILE__, __LINE__)
+#define libmesh_map_find(map, key) libMesh::Utility::map_find((map), (key), __FILE__, __LINE__)
 
 /**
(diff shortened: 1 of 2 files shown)


include/utils/utility.h
```diff
@@ -148,6 +155,45 @@ map_find(const Map & map,
...
+template<typename Vector>
+inline
+typename Vector::const_reference &
+vector_at(const Vector & vec,
+          typename Vector::size_type i,
+          const char * filename,
+          int line_number)
+{
+  if (i >= vec.size())
+    libmesh_error_msg("vec_at() error: Index " + std::to_string(i) +
```
jwpeterson: I'd just using streaming `<<`, cut out all the `std::to_string` which shouldn't be necessary.

roystgnr: Good point.  Will do.

PR #2440, opened 2020-02-18
Call enforce_constraints_on_residual using local solution

fdkong:
Do not use parallel global solution that will automatically localize the whole solution. This causes scaling issue and memory issue.

This simply makes the simulation non-scalable.

Here is a test using 1536 cores for a problem with 10,100,279 DoFs.

Original performance:

```
SNESSolve              3 1.0 2.7287e+02 1.0 2.42e+09 2.8 9.1e+07 3.7e+04 7.8e+02 93100 93100 86  93100 93100 87  9061
SNESSetUp              3 1.0 1.7834e-04 8.4 0.00e+00 0.0 0.0e+00 0.0e+00 0.0e+00  0  0  0  0  0   0  0  0  0  0     0
SNESFunctionEval      13 1.0 2.3555e+02 1.0 1.87e+06 1.7 8.1e+07 4.0e+04 7.8e+01 80  0 83 95  9  80  0 83 95  9    10
SNESJacobianEval      10 1.0 2.8432e+01 1.0 3.57e+07 2.3 1.1e+06 1.1e+05 1.5e+02 10  2  1  3 17  10  2  1  3 17  1362
SNESLineSearch        10 1.0 1.6866e+02 1.0 3.87e+07 2.6 6.3e+07 4.0e+04 1.0e+02 57  2 64 73 11  57  2 64 73 11   237
```

With this PR:

```
SNESSolve              3 1.0 3.7522e+01 1.0 2.42e+09 2.8 1.5e+07 1.1e+04 7.2e+02 63100 68 91 85  63100 68 91 86 65895
SNESSetUp              3 1.0 2.3532e-0411.5 0.00e+00 0.0 0.0e+00 0.0e+00 0.0e+00  0  0  0  0  0   0  0  0  0  0     0
SNESFunctionEval      13 1.0 1.6213e+00 1.0 1.87e+06 1.7 4.6e+06 3.2e+03 2.6e+01  3  0 22  8  3   3  0 22  8  3  1444
SNESJacobianEval      10 1.0 2.7916e+01 1.0 3.57e+07 2.3 1.1e+06 1.1e+05 1.5e+02 47  2  5 67 18  47  2  5 67 18  1387
SNESLineSearch        10 1.0 1.4380e+00 1.1 3.87e+07 2.6 3.8e+06 3.2e+03 6.0e+01  2  2 18  7  7   2  2 18  7  7 27849
```

The simulation is almost ten times faster with this PR

### DIFF
--- a/src/solvers/petsc_nonlinear_solver.C
+++ b/src/solvers/petsc_nonlinear_solver.C
@@ -182,10 +182,19 @@ extern "C"
     else
       libmesh_error_msg("Error! Unable to compute residual and/or Jacobian!");
 
-    PetscVector<Number> X(x, rc.sys.comm());
+
+    // Synchronize PETSc x to local solution since the local solution may be changed due to the constraints
+    PetscVector<Number> & X_sys = *cast_ptr<PetscVector<Number> *>(rc.sys.solution.get());
+    PetscVector<Number> X_global(x, rc.sys.comm());
+
+    X_global.swap(X_sys);
+    rc.sys.update();
+    X_global.swap(X_sys);
 
     R.close();
-    rc.sys.get_dof_map().enforce_constraints_on_residual(rc.sys, &R, &X);
+
+    rc.sys.get_dof_map().enforce_constraints_on_residual(rc.sys, &R, rc.sys.current_local_solution.get());
+
     R.close();
 
     return rc.ierr;
@@ -222,9 +231,17 @@ extern "C"
     else
       libmesh_error_msg("Error! Unable to compute residual for forming finite difference Jacobian!");
 
+    // Synchronize PETSc x to local solution since the local solution may be changed due to the constraints
+    PetscVector<Number> & X_sys = *cast_ptr<PetscVector<Number> *>(rc.sys.solution.get());
+    PetscVector<Number> X_global(x, rc.sys.comm());
+
+    X_global.swap(X_sys);
+    rc.sys.update();
+    X_global.swap(X_sys);
+
     R.close();
-    PetscVector<Number> X(x, rc.sys.comm());
-    rc.sys.get_dof_map().enforce_constraints_on_residual(rc.sys, &R, &X);
+
+    rc.sys.get_dof_map().enforce_constraints_on_residual(rc.sys, &R, rc.sys.current_local_solution.get());
 
     R.close();
 
@@ -264,9 +281,17 @@ extern "C"
       libmesh_error_msg("Error! Unable to compute residual for forming finite differenced"
                         "Jacobian-vector products!");
 
+    // Synchronize PETSc x to local solution since the local solution may be changed due to the constraints
+    PetscVector<Number> & X_sys = *cast_ptr<PetscVector<Number> *>(rc.sys.solution.get());
+    PetscVector<Number> X_global(x, rc.sys.comm());
+
+    X_global.swap(X_sys);
+    rc.sys.update();
+    X_global.swap(X_sys);
+
     R.close();
-    PetscVector<Number> X(x, rc.sys.comm());
-    rc.sys.get_dof_map().enforce_constraints_on_residual(rc.sys, &R, &X);
+
+    rc.sys.get_dof_map().enforce_constraints_on_residual(rc.sys, &R, rc.sys.current_local_solution.get());
 
     R.close();
 


jwpeterson:
I think this was changed in 6667d894f7735fc9aaec592b5e6265137f4ed3b6. You may want to ask your buddy @lindsayad about it, I'm guessing it didn't work without this change...

fdkong:
> I think this was changed in [6667d89](https://github.com/libMesh/libmesh/commit/6667d894f7735fc9aaec592b5e6265137f4ed3b6). You may want to ask your buddy @lindsayad about it, I'm guessing it didn't work without this change...

Ok, then we need to understand why it did not work at the first place. Because this change will make simulation non-scalable for sure.

@lindsayad Any thoughts? Do you want to run simulation using at most two cores? Haha

lindsayad:
I believe I had to make it that way or else the constraint equations were not correct, as suggested by over 50 failures in the MOOSE test harness

lindsayad:
@fdkong Look at the bottom of the [old PR](https://github.com/libMesh/libmesh/pull/1986/files). We've always had to do this parallel constraint enforcement

fdkong:
> I believe I had to make it that way or else the constraint equations were not correct, as suggested by over 50 failures in the MOOSE test harness

OK, I see. I will look at these use cases (failing tests).

lindsayad:
> @fdkong Look at the bottom of the [old PR](https://github.com/libMesh/libmesh/pull/1986/files). We've always had to do this parallel constraint enforcement

But...we only enforced the constraints once at the end of each nonlinear solve...now we do it every residual and Jacobian evaluation.

So if we want to revert that old PR and go back to having a singular  function for constraints in order to restore scalability, then I can be on-board with that. It seems like a justified trade-off

roystgnr:
So where exactly is the parallel global solution?  I know it sure *looks* like that's what X is, but that constructor is supposed to making a shim of x, not a copy of X.  See https://github.com/libMesh/libmesh/pull/1986/files#r290943705

On the other hand, if you're seeing 10-fold speedup then surely *something* slower is happening in the x case vs the current_local_solution case, and I don't see what.

fdkong:
> So where exactly is the parallel global solution? I know it sure _looks_ like that's what X is, but that constructor is supposed to making a shim of x, not a copy of X. See https://github.com/libMesh/libmesh/pull/1986/files#r290943705
> 
> On the other hand, if you're seeing 10-fold speedup then surely _something_ slower is happening in the x case vs the current_local_solution case, and I don't see what.

Not only speedup. The old code will run out of memory very soon. The timings here are just 3 time steps. The old code could not run more than 3 time steps, and new code can run more than 20 time steps.

The problem is here:

```
  if (solution->type() == SERIAL || solution->type() == GHOSTED)
      solution_local = solution;
  else if (solution->type() == PARALLEL)
    {
      solution_built = NumericVector<Number>::build(this->comm());
      solution_built->init (solution->size(), solution->size(), true, SERIAL);
      solution->localize(*solution_built);
      solution_built->close();
      solution_local = solution_built.get();
    }
```
`
solution_built->init (solution->size(), solution->size(), true, SERIAL);`

means we gather  the whole world to everybody!!!

lindsayad:
Or actually I'm betting the answer is to apply your PR here @fdkong, and then reverse this diff from #1986 (which occurs after the nonlinear solve):
```
-  // Enforce constraints exactly now that the solve is done.  We have
-  // been enforcing them on the current_local_solution during the
-  // solve, but now need to be sure they are enforced on the parallel
-  // solution vector as well.
-  this->system().get_dof_map().enforce_constraints_exactly(this->system());
```
I'm thinking maybe that will give us a good function as well as  good scalability

lindsayad:
> means we gather the whole world to everybody!!!

Oooh, beautiful code!

fdkong:
> On the other hand, if you're seeing 10-fold speedup then surely _something_ slower is happening in the x case vs the current_local_solution case, and I don't see what.

The code was slow  because we call `VecScatter` to gather the global solution to everybody.

lindsayad:
Ok @fdkong I pushed a commit to your branch restoring the parallel constraint enforcement at the end of the solve. let's see how this does with testing

roystgnr:
If that doesn't work, could we just replace that serial vector with a ghosted vector?

lindsayad:
Ok now we're just pack to the singular function 😆

src/solvers/petsc_nonlinear_solver.C
```diff
@@ -943,6 +943,12 @@ PetscNonlinearSolver<T>::solve (SparseMatrix<T> &  pre_in,  // System Preconditi
   ierr = SNESGetLinearSolveIterations(_snes, &_n_linear_iterations);
   LIBMESH_CHKERR(ierr);
 
+  // Enforce constraints exactly now that the solve is done.  We have
+  // been enforcing them on the current_local_solution during the
+  // solve, but now need to be sure they are enforced on the parallel
+  // solution vector as well.
+  this->system().get_dof_map().enforce_constraints_exactly(this->system());
+
```
fdkong: Why do we need to enforce it after solve? We manually to change the solution to match prescribed constraints?

fdkong: OK, I believe I see the whole picture now.  I will add more changes, and then we do not need this anymore.

It should be very easy

lindsayad:
Currently, we:

1. `enforce_constraints_exactly` on the `current_local_solution` before actually computing the residual. The rationale for this is a little vague to me, but I believe it's to prevent weird things from happening in displaced mesh problems (like inverted elements).
2. While computing the residual, we call `constrain_element_vector` to move the physics residuals from constrained dofs to unconstrained dofs
3. Post residual computation, call `enforce_constraints_on_residual` which computes a residual based on the global parallel solution

With @fdkong first commit here, the constraint residual would always be zero because we are evaluating it based on the `current_local_solution` which we already constrained in step 1. Hence at least in PJFNK there's no driver to get the constrained dof correct...even if the constrained dof is "wrong" in the global solution vector, we are going to see a zero in the residual That is why we saw all those MOOSE test failures in the initial PR and why it was fixed in the commit I pushed up which enforces the constraint on the global solution vector post-solve.

The problem is that we're back to a singular function now...we enforce the constraint exactly on the current local solution, and then we evaluate the constraint residual with an already constrained solution, so that the constraint residual is always zero no matter how we perturb the involved dofs.

My idea would be to remove 1. and see what happens...

lindsayad:
Oh I also like @roystgnr GHOSTED suggestion...

fdkong:
Trying to synchronize PETSc solution to system current local solution, and so we can directly use the system current local solution.

Let us see if we have a better chance.

fdkong:
> My idea would be to remove 1. and see what happens... If that turned out to be a bad idea then I am fine reverting #1986

I do not think we need to revert the code. We could easily fix this by making enforce_constraints_on_residual scalable. (1) we could just gather the part of the solution needed for building constraints. It can be done by creating a VecScatter using constraint DoFs. 

Or (2) We could synchronize the PETSc solution again so that the current local solution does not include constraints any more.

fdkong:
> Or (2) We could synchronize the PETSc solution again so that the current local solution does not include constraints anymore.

This works. @roystgnr @lindsayad  Could you review?

roystgnr:
Will do; give me a little while.

On a more embarrassing note, @jwpeterson just explained to me what I should have seen clearly already: DofMap::enforce_constraints_exactly() has the exact same sort of scalability problem, so this PR doesn't even fix every code path.  I'll put together a separate PR fixing that.

In my defense, I'm pretty sure I wrote the offending serialization code in like 2007, before we even had ghosted vector support, so at the time it wasn't completely unreasonable.  I still should have noticed and fixed it a decade ago, though.

src/solvers/petsc_nonlinear_solver.C
```diff
@@ -182,10 +182,19 @@ extern "C"
...
 
-    PetscVector<Number> X(x, rc.sys.comm());
+
+    // Synchronize PETSc x to local solution since the local solution may be changed due to the constraints
+    PetscVector<Number> & X_sys = *cast_ptr<PetscVector<Number> *>(rc.sys.solution.get());
+    PetscVector<Number> X_global(x, rc.sys.comm());
+
+    X_global.swap(X_sys);
+    rc.sys.update();
+    X_global.swap(X_sys);
```
lindsayad: This is interesting. So we `enforce_constraints_exactly` on the `current_local_solution` before the residual evaluation. But here we are restoring the `current_local_solution` to PETSc `x` so that when perturbing `x` we will actually see those effects appear when we `enforce_constraints_on_residual` post residual evaluation. I'm down for this. I assume that you've checked the performance on this and it looks good?

fdkong:
> In my defense, I'm pretty sure I wrote the offending serialization code in like 2007, before we even had ghosted vector support, so at the time it wasn't completely unreasonable. I still should have noticed and fixed it a decade ago, though.

I think a better way is to extract needed DoFs only for example constraints. That will be more efficient.  Another concern, we create VecScatter on the fly and destroy it, but I think we should reuse VecScatter as much as possible.

However, the current fix is a quick fix for me, and I will do more tests to see if we have any performance penalty.

fdkong:
> On a more embarrassing note, @jwpeterson just explained to me what I should have seen clearly already: DofMap::enforce_constraints_exactly() has the exact same sort of scalability problem, so this PR doesn't even fix every code path. I'll put together a separate PR fixing that.

It would be great if you could fix all paths. There are a lot. Everything related to `serial` should be revisited  again.

lindsayad:
> On a more embarrassing note, @jwpeterson just explained to me what I should have seen clearly already: DofMap::enforce_constraints_exactly() has the exact same sort of scalability problem, so this PR doesn't even fix every code path. I'll put together a separate PR fixing that.

This PR doesn't fix *any* code paths in `DofMap`. It circumvents the problem completely because `current_local_solution` is GHOSTED as opposed to PARALLEL, so we don't do any localization. A follow-up PR will be great, so that other code that unwittingly calls these `DofMap` constraint methods won't automatically degrade their scalability.

lindsayad:
> Beautiful code

It's acceptable code 😆

roystgnr:
> It would be great if you could fix all paths.

It would, but some will be easier than others; correctness is even more important than performance, as I was ironically reminded when my web browser crashed while I was typing the *original* version of this comment.  (even Typio didn't save it, what the hell?)

In MeshFunctionSolutionTransfer, ExactSolution, and ExactErrorEstimator, as well as for a MOOSE object or two IIRC, there are code paths where we have two arbitrary meshes, no good a priori way of deciding their overlaps, and so we'll be relying on serial vectors indefinitely.  Not going to call those "WONTFIX", but definitely "WONTFIX_UNTIL_SOMEONE_NEEDS_IT_BADLY_ENOUGH" to justify the development time necessary.

But you're absolutely right that we needed to check out everything.  UniformRefinementEstimator looks like it'll be an easy fix.  And there are a couple code paths in System::calculate_norm and System::project_vector, where we aren't overcommunicating but I think we *are* over allocating, which I expect will be easy to fix.

I'd also recommend @jwpeterson look at RBEIMConstruction - I see a couple places there where I feel confident enough to suggest there's a problem but not confident enough to quickly fix it myself.

pbauman:
`PetscDiffSolver` should also be low hanging fruit for this (and I of course would like to see it there).

jwpeterson:
> I'd also recommend @jwpeterson look at RBEIMConstruction - I see a couple places there where I feel confident enough to suggest there's a problem but not confident enough to quickly fix it myself.

Will do.

fdkong:
> However, the current fix is a quick fix for me, and I will do more tests to see if we have any performance penalty.

Looks like the performance is great for me now!

PR #2441, opened 2020-02-18
Skip redundant resize in init_shape_functions

roystgnr:
This shaves 2-5% off benchmark runtimes for me.


roystgnr:
That race conditions report is an unrelated false positive from the point of view of this PR, yes?

permcody:
Most likely  I think all patches are in next but if you branched from devel
you’ll be a little behind. Just ignore that

PR #2443, opened 2020-02-19
 RBEIMConstruction: Move FE getter calls outside of elem loops

jwpeterson:
Not sure how we got away with this before, but after recently updating libmesh, we started hitting the
```
Assertion `!calculations_started || calculate_xyz' failed.
```
in `FEMap` in some of the `reduced_basis` code, and this change fixes it.


jwpeterson:
Failure is due to MOOSE race condition detection and unrelated to this PR. Going to go ahead and merge since this code is not tested by MOOSE test anyway.

roystgnr:
That's actually a new assertion, only dating back to the FEMap refactoring, when I noticed that we were overcomputing in FEMap.

I was hoping that nobody else (there was one MOOSE test failure IIRC) would trip it, since it's directly analogous to the FE behavior we've had for a decade; failing that I was hoping that whoever tripped it would appreciate the optimization enough to not be annoyed by the technical failure of backwards compatibility.  :sweat_smile:

PR #2444, opened 2020-02-19
Control sorting method used in BoundaryInfo::build_node_list()

jwpeterson:
I have a case where it is useful to sort the list of tuples returned by `BoundaryInfo::build_node_list()` by boundary id instead of node id since I later need to search in it by boundary id. This PR adds the ability to control the type of sort, and defaults to the current approach which is by node id.


jwpeterson:
Failure was again due to race condition detection in MOOSE, unrelated to PR.

PR #2446, opened 2020-02-20
Avoid serial vectors

roystgnr:
These are all the "easy" fixes I can find to cases of overserialization, based on the failure case @jwpeterson pointed out to me and on the discussion of the case in #2440

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -2128,7 +2128,9 @@ void DofMap::enforce_constraints_exactly (const System & system,
   else if (v->type() == PARALLEL)
     {
       v_built = NumericVector<Number>::build(this->comm());
-      v_built->init (v->size(), v->size(), true, SERIAL);
+      v_built->init (v->size(), v->local_size(),
+                     this->get_send_list(), true,
+                     GHOSTED);
       v->localize(*v_built);
       v_built->close();
       v_local = v_built.get();
@@ -2206,7 +2208,8 @@ void DofMap::enforce_constraints_on_residual (const NonlinearImplicitSystem & sy
   else if (solution->type() == PARALLEL)
     {
       solution_built = NumericVector<Number>::build(this->comm());
-      solution_built->init (solution->size(), solution->size(), true, SERIAL);
+      solution_built->init (solution->size(), solution->local_size(),
+                            this->get_send_list(), true, GHOSTED);
       solution->localize(*solution_built);
       solution_built->close();
       solution_local = solution_built.get();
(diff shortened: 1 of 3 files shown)


jwpeterson:
Thanks, this looks like it should fix our issue with `DofMap::enforce_constraints_exactly()`. We'll re-run our test case after this is merged and let you know how it goes.

src/base/dof_map_constraints.C
```diff
@@ -2206,7 +2208,8 @@ void DofMap::enforce_constraints_on_residual (const NonlinearImplicitSystem & sy
   else if (solution->type() == PARALLEL)
     {
       solution_built = NumericVector<Number>::build(this->comm());
-      solution_built->init (solution->size(), solution->size(), true, SERIAL);
+      solution_built->init (solution->size(), solution->local_size(),
+                            this->get_send_list(), true, GHOSTED);
       solution->localize(*solution_built);
```
fdkong: This may be still a problem. You could look at 

```
template <typename T>
void PetscVector<T>::localize (NumericVector<T> & v_local_in) const
{
  this->_restore_array();

  // Make sure the NumericVector passed in is really a PetscVector
  PetscVector<T> * v_local = cast_ptr<PetscVector<T> *>(&v_local_in);

  libmesh_assert(v_local);
  libmesh_assert_equal_to (v_local->size(), this->size());

  PetscErrorCode ierr = 0;
  const PetscInt n = this->size();

  IS is;
  VecScatter scatter;

  // Create idx, idx[i] = i;
  std::vector<PetscInt> idx(n);
  std::iota (idx.begin(), idx.end(), 0);

  // Create the index set & scatter object
  ierr = ISCreateLibMesh(this->comm().get(), n, idx.data(), PETSC_USE_POINTER, &is);
  LIBMESH_CHKERR(ierr);

  ierr = LibMeshVecScatterCreate(_vec,          is,
                          v_local->_vec, is,
                          &scatter);
  LIBMESH_CHKERR(ierr);

  // Perform the scatter
  ierr = VecScatterBegin(scatter, _vec, v_local->_vec,
                         INSERT_VALUES, SCATTER_FORWARD);
  LIBMESH_CHKERR(ierr);

  ierr = VecScatterEnd  (scatter, _vec, v_local->_vec,
                         INSERT_VALUES, SCATTER_FORWARD);
  LIBMESH_CHKERR(ierr);

  // Clean up
  ierr = LibMeshISDestroy (&is);
  LIBMESH_CHKERR(ierr);

  ierr = LibMeshVecScatterDestroy(&scatter);
  LIBMESH_CHKERR(ierr);

  // Make sure ghost dofs are up to date
  if (v_local->type() == GHOSTED)
    v_local->close();
}
```

`  const PetscInt n = this->size()` should be changed to   `const PetscInt n = this->local_size();`. And then change anything else if necessary.

roystgnr: No, the bug here is that I didn't change which overload is being called!  We have a send_list, we need to be using it.

Thanks for catching this.

PR #2449, opened 2020-02-21
Vectorized shape, shape_deriv

roystgnr:
I intend to actually start trying to *optimize* the vectorized forms of these functions at some point, but as soon as I used FEInterface::shapes in H1FETransformation I saw a 10% speedup in one benchmark (just by avoiding so many redundant switch statements), so let's get the easy win merged ASAP and continue on the harder stuff afterward.

### DIFF
--- a/include/fe/fe.h
+++ b/include/fe/fe.h
@@ -22,6 +22,7 @@
 
 // Local includes
 #include "libmesh/fe_base.h"
+#include "libmesh/int_range.h"
 #include "libmesh/libmesh.h"
 
 // C++ includes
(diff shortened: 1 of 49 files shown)


roystgnr:
I've been trying optimized versions of shapes, and shape_derivs, and I'm pretty confident that there's no low-hanging fruit there; I can move a few more switch statements outside of loops but I get something like 0.5% speedup on benchmarks this time, not 10%; I can't even be sure the "speedup" is statistically significant without a lot more trials.  So the only win here was this first reduction of function calls and switch statements.

I do have one more idea that might be worth trying...

PR #2455, opened 2020-02-26
FE, FEMap optimizations

roystgnr:
I only get about 6% speedup out of this on a typical benchmark, where I had been kind of hoping for 16% with a few further changes (none of which actually helped), but 6% is better than nothing.


roystgnr:
I can replicate this and I think I understand what the problem is.  There's definitely a shortcoming in one of the commits here, and I think there might be an optimization possible in MOOSE that we've been missing but that I just noticed since the less optimized code path triggered my bug.

PR #2456, opened 2020-03-03
If user specifies particular MPI installation, then use it

lindsayad:
This is different from what we've done historically where we always try to snoop MPI from PETSc (when PETSc is available). This is also consistent with what we want to do in TIMPI, which from a general dependency standpoint cares nothing about PETSc.


m4/compiler.m4
```diff
@@ -6,22 +6,40 @@ AC_DEFUN([LIBMESH_SET_COMPILERS],
...
   AS_IF([test "$enablempi" != no],
-        [CXX_TRY_LIST="mpicxx mpiCC mpicc $CXX_TRY_LIST"],
-        AC_MSG_RESULT([>>> Disabling MPI per user request <<<]))
+        [
+          dnl Ok we want to enable mpi. Let's think about what compilers
+          dnl we want to try.
+          AS_IF([test x"$MPI" != x],
+                [
+                  dnl only try to set CXX if the user hasn't specified one already
+                  AS_IF([test -z "$CXX"],
```
roystgnr: Is this safe to do *before* the --with-cxx support below?

lindsayad: No it's not safe. It wouldn't mess up the setting of CXX but it could lead to incorrectly setting other flags like `MPI_USING_WRAPPERS`

lindsayad:
I just realized that this calls `CONFIGURE_PETSC` twice...that seems pretty dumb

m4/compiler.m4
```diff
@@ -2,6 +2,11 @@
 # -------------------------------------------------------------
 AC_DEFUN([LIBMESH_SET_COMPILERS],
 [
+  dnl We need to verify that we've done AC_ARG_ENABLE(mpi), AC_ARG_WITH(mpi), AC_ARG_ENABLE(petsc)
+  dnl which occur in LIBMESH_COMPILER_CONTROL_ARGS
+  AS_IF([test x"CALLED_COMPILER_CONTROL_ARGS" != x1],
+        [AC_MSG_ERROR([the compiler control args macro must be called before setting compilers])])
+
```
roystgnr: Isn't this sort of thing what ```AC_REQUIRE([LIBMESH_COMPILER_CONTROL_ARGS])``` is designed to do, but more flexibly?

m4/libmesh_compiler_control_args.m4
```diff
@@ -0,0 +1,71 @@
+AC_DEFUN([LIBMESH_COMPILER_CONTROL_ARGS],
```
roystgnr: Should we call this LIBMESH_ rather than a more general name if we're going to try to use the same logic (the same m4 file, right?) in TIMPI?

m4/compiler.m4
lindsayad: It looks like you commented on an outdated commit here...indeed while it took me a while, I did come to `AC_REQUIRE` on my own 😆

m4/libmesh_compiler_control_args.m4
lindsayad: Yea this does look like it could be plugged directly into libMesh. I'll remove the prefix

lindsayad:
> This is looking like it's ready to go. Before we merge, though, can we sync up [libMesh/TIMPI#21](https://github.com/libMesh/TIMPI/pull/21) to use the new m4 refactoring here? I'd like to test this on a bunch of different systems and module stacks, and it'd be easier to do that with TIMPI where the configure/make/check cycle is a lot faster.

Yep!

roystgnr:
So here's a crazy question, inspired by this PR but external to it.  Since TIMPI is a required package for libMesh and always will be ... would it be sane to just add the TIMPI m4 directory in the libMesh autoconf path and delete the dozen-odd redundant files from libMesh m4?  We'll still have the generated configure script to provide an informative error message when the TIMPI submodule hasn't been updated, and with AM_MAINTAINER_MODE disabled there shouldn't be any chicken-and-egg problems there.  That'll make it much easier to keep things in sync if/when we need to make future changes.

lindsayad:
> So here's a crazy question, inspired by this PR but external to it. Since TIMPI is a required package for libMesh and always will be ... would it be sane to just add the TIMPI m4 directory in the libMesh autoconf path and delete the dozen-odd redundant files from libMesh m4? We'll still have the generated configure script to provide an informative error message when the TIMPI submodule hasn't been updated, and with AM_MAINTAINER_MODE disabled there shouldn't be any chicken-and-egg problems there. That'll make it much easier to keep things in sync if/when we need to make future changes.

That seems like an excellent idea to me. Doing something like that almost eases maintainability and also robustifies the m4 used IMO

Issue #2457, opened 2020-03-04
BoundaryInfo::sync(): Store parent side index in BoundaryMesh using "extra_integers"


jwpeterson:
This information is almost always useful when creating a BoundaryMesh, and currently the only way to recover it is via `Elem::which_node_am_i()` calls after the fact. It would be better to store this information when it is available during BoundaryInfo::sync(), using the new flexible "extra_integers" storage that was added recently.

jwpeterson:
I plan to make storing these indices optional with the default being "on" but IMO they could also just be non-optional, the way that interior_parents are always stored for BoundaryMeshes...

roystgnr:
My vote is still for optional, but I think I'm starting to come up with reasons why that's my gut hunch:

We sometimes generate multidimensional meshes by starting with a volume mesh and using BoundaryInfo::add_elements to add surface mesh elements on some side(s).  In that case, because of the current limitations of extra_integers storage, we usually don't want to add an integer that's useful on ```N^{2/3}``` boundary elements but would get allocated and manipulated on all $N$, but sometimes that might be worth it, so we'd want to leave it up to the user.  And if we've already done the work to make it optional in that case, we might as well let it be optional in every case.

jwpeterson:
Merged in ef7dd34ad

PR #2458, opened 2020-03-05
Set extra integer on boundary_mesh elems corresponding to parent side index

jwpeterson:
Refs #2457

### DIFF
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -161,7 +161,8 @@ public:
    * Only boundary elements with the specified ids are created.
    */
   void add_elements (const std::set<boundary_id_type> & requested_boundary_ids,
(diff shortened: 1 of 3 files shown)


jwpeterson:
This is not yet optional as we discussed in #2457, I want to see if there are any major issues with the idea before updating the relevant `BoundaryInfo` APIs.

roystgnr:
Looks good to me once you add the optionality.

jwpeterson:
@roystgnr please review again when you have the time, the setting of parent side ids is now optional.

roystgnr:
Looks good enough to me.  We might go back and add the flag to sync() too if anyone complains in the future.

jwpeterson:
Thanks!

PR #2460, opened 2020-03-09
Update TIMPI and MetaPhysicL submodules

lindsayad:
- Update to new DualNumber template signature
- Update StandardType specializations to use `is_fixed_type` static member API

### DIFF
--- a/include/numerics/dense_matrix.h
+++ b/include/numerics/dense_matrix.h
@@ -42,19 +42,16 @@
 #include <algorithm>
 
 #ifdef LIBMESH_HAVE_METAPHYSICL
-namespace MetaPhysicL
-{
-template <typename, typename>
-class DualNumber;
-}
+#include "metaphysicl/dualnumber_forward.h"
+
 namespace std
 {
 // These declarations must be visible to the DenseMatrix method declarations that use
 // a std::abs trailing return type in order to instantiate a DenseMatrix<DualNumber>
-template <typename T, typename D>
-MetaPhysicL::DualNumber<T, D> abs(const MetaPhysicL::DualNumber<T, D> & in);
-template <typename T, typename D>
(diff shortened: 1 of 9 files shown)


roystgnr:
Well that's fun.  Wonder why it didn't scream about the incompatibility on non-Mac MOOSE builds.

lindsayad:
Looks like it happily fails on the Linux boxes as well (as one would hope). Just had to give it time. We knew this was an API breaking change. For this one to go in, we're just going to have to live with the failures (until I get the libMesh update into MOOSE)

roystgnr:
I hate non-overlapping updates, even *aside* from the loss of CI.  The problem here is forward declarations, no?  Could we first update MetaPhysicL to the earliest version with dualnumber_forward.h, then update MOOSE to use that header for forward declarations, then do the final update?

lindsayad:
I should have thought more critically about this at the time...The third template parameter and `dualnumber_forward.h` were introduced in the same commit. We could revert that commit, and then I can split the commit into two.

lindsayad:
Actually, the least painful would be for me to to do a `MetaPhysicL` version check and if `MAJOR_VERSION < 1`, then use the existing forward declaration, else include `metaphysicl/dualnumber_forward.h`

lindsayad:
Well, this shows that my current way of doing `--with-future-timpi-dir` isn't very fail-safe... Going to have to revert https://github.com/idaholab/civet_recipes/pull/1620 while I think about the right thing to do. Is it a crazy idea to create a libtimpi.so which points to the most optimized extension of TIMPI that gets built?

lindsayad:
Looks like this is ready to review/merge

roystgnr:
My "only configures when I pass in --enable-entrails-of-a-chicken" system is now autodetecting things perfectly, thank you!

lindsayad:
That’s cool to hear!

PR #2461, opened 2020-03-10
Only checking mapping is not enough to determine if a vec is ghosted

fdkong:
We need to check if vec has a local representation as well

### DIFF
--- a/include/numerics/petsc_vector.h
+++ b/include/numerics/petsc_vector.h
@@ -577,8 +577,13 @@ PetscVector<T>::PetscVector (Vec v,
       ierr = VecGetLocalToGlobalMapping(_vec, &mapping);
       LIBMESH_CHKERR(ierr);
 
+      Vec localrep;
+      ierr = VecGhostGetLocalForm(_vec,&localrep);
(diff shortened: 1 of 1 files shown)


fdkong:
@roystgnr @jwpeterson Any comments?

roystgnr:
Looks good, but I'm trying to understand how this bug came to be in the first place.  Did PETSc not originally return a LocalToGlobalMapping for VecCreateMPI vectors but now it does?  Are you setting up a Vec in a special way that gives it a LocalToGlobalMapping without have created it with VecCreateGhost?

fdkong:
> Looks good, but I'm trying to understand how this bug came to be in the first place. Did PETSc not originally return a LocalToGlobalMapping for VecCreateMPI vectors but now it does? Are you setting up a Vec in a special way that gives it a LocalToGlobalMapping without have created it with VecCreateGhost?

`VecCreateMPI` still behaves like before.


I am using `DMCreateGlobalVector` that calls `DMCreateGlobalVector_DA(DM da,Vec *g)`. Here DM adds `LToGMapping` in order to use  `VecSetValuesLocal`

PR #2463, opened 2020-03-11
Add edge to side map for all cells

### LINKED ISSUE #2462: Add edge to side maps for cells
We use these in ray tracing. It makes sense to have them in libMesh instead.

loganharbour:
Closes #2462

### DIFF
--- a/include/geom/cell_hex.h
+++ b/include/geom/cell_hex.h
@@ -165,6 +165,7 @@ public:
    */
   virtual std::pair<Real, Real> qual_bounds (const ElemQuality q) const override;
 
+  virtual std::vector<unsigned int> sides_on_edge(const unsigned int e) const override final;
 
 protected:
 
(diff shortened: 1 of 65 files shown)


loganharbour:
There's a 2 everywhere that is the number of sides that contain an edge... should this be a static variable?

loganharbour:
I agree. The routines we're using have direct access to the type, therefore they just access the static variables, so I didn't think about routines.

I'll add some to this and ping when done.

roystgnr:
We can merge after that, then.

If you have time, a unit test wouldn't hurt either.

<<Strange, you think.  Who would write that sentence down and tie it to a stone?  Perhaps you should investigate the glass house nearby.>>

loganharbour:
What's an appropriate unit test here?  Checking all of the maps with respect to one another?

roystgnr:
Yeah, that's about all I can think of.  Assert is_node_on_edge && is_edge_on_side => is_node_on_side.

loganharbour:
Alright. This is good now.

roystgnr:
Looks good to me too, thanks!  I'll hit the button if I'm still awake when it passes CI.

PR #2464, opened 2020-03-12
Deprecate reinit of FE without prerequests

roystgnr:
This PR introduces FEAbstract::get_nothing(), for when you really want an FE object not to compute anything on quadrature points, and it deprecates the old behavior where if you tried not prerequesting anything then we'd happily compute *everything*.

This was originally intended to look for more performance bugs in MOOSE like https://github.com/idaholab/moose/issues/14811 - but in the process I found multiple prerequest failures in libMesh itself, a couple of them serious.

This is passing my basic test runs; I have more testing to do but I think we're ready to see what CI thinks.


roystgnr:
Damn, I was hoping this would work on the first try.  I'm getting 25% speedup on AMR benchmarks, and it's passing libMesh tests for *me*...

roystgnr:
Going to start modules and external apps and such once the basic CI is finished.

I'm hopeful that this will pass the basic tests this time, because I'd just screwed up a couple of the reduced_basis example upgrades, but on that note I'd rather not merge until @dknez or @jwpeterson or someone has a chance to test this against the Akselos code.  I believe the results will be "pass with a default build" and "fail if libMesh is built with --disable-deprecated, but in informative ways in dbg mode that point to optimization opportunities in Akselos", but I'd like to have evidence for that rather than just belief.

jwpeterson:
soupnazi

FE if you don't pre-request data properly.

jwpeterson:
> I'd rather not merge until @dknez or @jwpeterson or someone has a chance to test this against the Akselos code.

I will test this out. Ideally I think I'll do one update with all the TIMPI/Metaphysical submodule updates first to make sure that's not broken, then move on to looking at this.

roystgnr:
In theory, aside from the one ```libmesh_deprecated()``` warning message, you should still be allowed to avoid proper prerequests so long as you configure without --disable-deprecated, so people whose code doesn't prerequest properly still have a loophole.

elaine

jwpeterson:
> I will test this out.

That said, no need to wait on merging if this is ready, it seems like a long-overdue fix!

roystgnr:
Long-overdue, so it can wait a little longer.  I'd hate to risk introducing an actual bug in the attempt to fix a performance flaw.

That 25% speedup was also kind of a best-case scenario (although I didn't plan it that way, I just used the same benchmark I've been favoring lately): 3D + AMR + LAGRANGE + JumpErrorEstimator hits both the major internal performance flaws while simultaneously being as cheap as possible in other code paths.  Codes which don't use AMR may see no "free" speedup, though I'd still recommend trying a --disable-deprecated build to see if speedup can be had with a few lines' patch in user code.

roystgnr:
Oh, but that reminds me, I still haven't started the optional CI runs.  Will do now.

roystgnr:
Okay, I'm pretty sure that "Clean Machine" failure wasn't me.

jwpeterson:
Error message for the failed build is
```
Release 6f3c438e838564d48bf591191986ef747f50c8e1 not found
```
may be something for @milljm to take a look at when he has time.

milljm:
Ubuntu-14? Wow... Old recipe me thinks. I am going to adjust this to Ubuntu-16. I can't remember why this was originally set on 14.

milljm:
The more I think about it, the more I think we only got Tiamat to work on Ubuntu-14. So this still might fail.

milljm:
No ones worked on Tiamat for over a year in a half. I am going to add a bunch of echos to this target, just so it is nice and green. Right now, it is failing to _fetch_ Tiamat. So... yeah. Not gonna happen.

jwpeterson:
OK, I kinda figured that might be the case, thanks for checking into it!

roystgnr:
Okay, I'm good to merge whether or not that last kitchen-sink test runs, then.  John, did you get a chance to check on Akselos stuff yet?

jwpeterson:
I confirmed that current master (before merging this PR) is working for us last night, so let's go ahead and merge this (and I'll do another update) and that way if there are any changes they'll hopefully be isolated to this PR.

PR #2465, opened 2020-03-13
 Forward declare BoundaryInfo in mesh_base.h

jwpeterson:
I noticed while working on #2458 that touching boundary_info.h causes dozens of files to recompile, but it doesn't seem like that should be necessary... I'm guessing that this may break some downstream apps, but hopefully that shouldn't be too hard to fix.


roystgnr:
Forget downstream apps; we have lazy downstream example codes!

jwpeterson:
It's possible that this change will break some downstream apps, but the fix will be simple, just include "libmesh/boundary_info.h" where necessary. I'm going to merge this as-is for now because it's not currently possible for me to attach optional App testing to PRs... I think something got messed up with moosebuild permissions.

PR #2469, opened 2020-03-17
Delete old DofObject only if it exists

### LINKED ISSUE #2468: DofObject only delete old if it exists
The `DofObject` destructor calls `clear_old_dof_object()`.

If the old dof object is null... this won't work. Hopefully it's appropriate to check if it exists before calling delete on it.

loganharbour:
Closes #2468

I hit an issue with the `clear_old_dof_object()` call when using a vector of `Node` in an unusual way.

### DIFF
--- a/src/base/dof_object.C
+++ b/src/base/dof_object.C
@@ -141,8 +141,11 @@ DofObject & DofObject::operator= (const DofObject & dof_obj)
 
 void  DofObject::clear_old_dof_object ()
 {
-  delete this->old_dof_object;
-  this->old_dof_object = nullptr;
+  if (this->old_dof_object)
+  {
+    delete this->old_dof_object;
+    this->old_dof_object = nullptr;
+  }
 }
 
 


loganharbour:
Doh.

Ended up with an uninitialized vector of `DofObject` which led to some pretty awful undefined behavior when going to clear/resize said vector.

loganharbour:
Actually - this may still be meaningful in the uninitialized case and worth putting in. The issue being that the uninitialized `DofObject` also does not have its old object initialized, so the underlying old object in the uninitialized object is also uninitialized...

...does that make sense?

roystgnr:
Not sure what you mean, sorry.  Our constructor initializes old_dof_object to nullptr, and then we're still save to call clear() and delete nullptr.  And if we actually did have an uninitialized old_dof_object, then the fix would be to make sure it was always initialized, not try to test it against nullptr (redundantly or unsafely or both, depending on the luck of the compiler).

jwpeterson:
> when using a vector of Node
>  uninitialized vector of DofObject

Just to clarify, you are talking about a vector of *pointers* here, correct? If you call delete on an uninitialized pointer then that's undefined behavior no matter what, the change in this PR shouldn't make any difference although it may seem to have, since undefined behavior is hard to predict.

loganharbour:
Sorry - this is the result of my not fully understanding my issue.

I ended up with a class that had a `std::vector<std::unique_ptr<Node>>` as a member variable that was not initialized to be empty on construction of the class. When I went to call `clear()` on said vector, I seg faulted on the `delete` that I guarded against here. I resolved my issue by initializing said vector to be a size of zero upon construction of the class. This also fixed it... but I'm not sure it's necessary anymore.

roystgnr:
> that was not initialized to be empty

Well, that just leaves me even more confused.  std::vector isn't like a built-in; default initialization of one should already leave you with a well-defined empty vector.

jwpeterson:
> member variable that was not initialized to be empty on construction of the class.

Agreed, even if not explicitly initialized in the initializer list, the vector would still be default-initialized. 

Is it possible that something like the following happened: 
```
std::vector<std::unique_ptr<Node>> vec;
// fill in vec somehow...
mesh.add_node(vec[0].get());
```
In this case, the Mesh takes ownership of the pointer but `vec` doesn't know about this. If the mesh desctructor subsequently runs before `vec` is destroyed, you would potentially see a segfault (undefined behavior) in the `vec` desctructor, or if you called `vec.clear()`. Anyway, not sure you are even using a `Mesh` here but this idea could apply to anything where ownership is accidentally transferred through a dumb pointer.

PR #2471, opened 2020-03-18
Enable Exodus II IO to load elemental variables into element IDs

### LINKED ISSUE #2470: Enhance Exodus IO a little to allow using elemental variables in the file to set extra integers
Now the mesh has been supporting the extra element integers. To make this even more useful with Exdous mesh files, we can load an elemental variable in the file and use it to set an extra integer in the loader. Exodus does not support integer variables, but we can use double and simply round the double. We found in the past in our reactor analysis application that letting the mesh contain the extra integers is super handy. I will send in a PR shortly.

YaqiWang:
Close #2470.

Note that I did not have a test in libMesh (will be adding one in MOOSE). If I have to add one, please just let me know.

Another thing is that `NameBasedIO` does not take extra arguments, so when using this capability, users will have to directly construct `ExodusII_IO`. I thought of using naming conventions like any elemental variables in Exodus file with name like `[anything]_id` will be loaded automatically, but thought that may not be a good idea (for example if later a user want to add an extra integer in the same name, there could some implications).

### DIFF
--- a/include/mesh/exodusII_io.h
+++ b/include/mesh/exodusII_io.h
@@ -304,6 +304,13 @@ public:
                      std::vector<std::map<BoundaryInfo::BCTuple, Real>> & bc_vals);
 
   /**
+   * Set the elemental variables in the Exodus file that to be read into extra
+   * element integers. The names of these elemental variables will be used to
+   * name the extra element integers.
+   */
+  void set_reading_extra_integer_vars(const std::vector<std::string> & extra_integer_vars);
+
+  /**
    * Sets the list of variable names to be included in the output.
    * This is _optional_.  If this is never called then all variables
    * will be present. If this is called and an empty vector is supplied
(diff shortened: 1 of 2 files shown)


YaqiWang:
@roystgnr or @jwpeterson can any of you review this?

include/mesh/exodusII_io.h
```diff
@@ -60,7 +60,8 @@ class ExodusII_IO : public MeshInput<MeshBase>,
    */
   explicit
   ExodusII_IO (MeshBase & mesh,
-               bool single_precision=false);
+               bool single_precision=false,
+               const std::vector<std::string> & extra_integer_vars = {});
```
jwpeterson: I think I'd prefer not to change the constructor if possible since nothing else is really handled this way? Can we handle it similar to `ExodusII_IO::set_output_variables()`, where the user can pass in a reference to a vector of strings that tells which variables are extra integer variables?

YaqiWang: Will do.

YaqiWang:
@jwpeterson Your comment was addressed. Btw, the PR in MOOSE shows how this new capability is used.

YaqiWang:
Thanks John. If this can be merged so that I can catch next libMesh update in MOOSE, it will be great.

PR #2473, opened 2020-03-18
Add allow_find_neighbors APIs to MeshBase

### LINKED ISSUE #2472: skip* deprecation inconsistency
The `skip_renumber_nodes_and_elements` argument to `MeshBase::prepare_for_use` is deprecated but the second default argument (`skip_find_neighbors`) is not. Yet we do not provide an interface for the user to set `skip_find_neighbors` other than through the `prepare_for_use` call. So we are essentially requiring that if the user wants to pass `skip_find_neighbors` then they will also have to provide the deprecated `skip_renumber_nodes_and_elements` argument. 

~~Additionally, if the first argument is going to be deprecated then we should issue a deprecation warning regardless of the value passed (we currently only issue the warning if `skip_renumber_nodes_and_elements=true`)~~ Duh, we can't do that...there's no c++ `isParamSetByUser` :rofl:

lindsayad:
This mirrors the APIs for allow_renumbering, which will allow
a more consistent setup before calling `prepare_for_use`.

Closes #2472

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -1066,6 +1066,13 @@ public:
   bool allow_renumbering() const { return !_skip_renumber_nodes_and_elements; }
 
   /**
+   * If \p false is passed then this mesh will no longer work to find element
+   * neighbors when being prepared for use
+   */
+  void allow_find_neighbors(bool allow) { _skip_find_neighbors = !allow; }
+  bool allow_find_neighbors() const { return !_skip_find_neighbors; }
+
+  /**
    * If false is passed in then this mesh will no longer have remote
    * elements deleted when being prepared for use; i.e. even a
    * DistributedMesh will remain (if it is already) serialized.
@@ -1784,6 +1791,11 @@ protected:
    * This is set when prepare_for_use() is called.
    */
   bool _skip_renumber_nodes_and_elements;
+
+  /**
+   * If this is \p true then we will skip \p find_neighbors in \p prepare_for_use
+   */
+  bool _skip_find_neighbors;
 
   /**
    * If this is false then even on DistributedMesh remote elements
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -1016,13 +1016,17 @@ public:
    *  4.) call \p cache_elem_dims()
    *
    * The argument to skip renumbering is now deprecated - to prevent a
-   * mesh from being renumbered, set allow_renumbering(false).
+   * mesh from being renumbered, set allow_renumbering(false). The argument to skip
+   * finding neighbors is also deprecated. To prevent find_neighbors, set
+   * allow_find_neighbors(false)
    *
    * If this is a distributed mesh, local copies of remote elements
    * will be deleted here - to keep those elements replicated during
    * preparation, set allow_remote_element_removal(false).
    */
-  void prepare_for_use (const bool skip_renumber_nodes_and_elements=false, const bool skip_find_neighbors=false);
+  void prepare_for_use (const bool skip_renumber_nodes_and_elements, const bool skip_find_neighbors);
+  void prepare_for_use (const bool skip_renumber_nodes_and_elements);
+  void prepare_for_use ();
 
   /**
    * Call the default partitioner (currently \p metis_partition()).
--- a/src/mesh/mesh_base.C
+++ b/src/mesh/mesh_base.C
@@ -66,6 +66,7 @@ MeshBase::MeshBase (const Parallel::Communicator & comm_in,
   _skip_noncritical_partitioning(false),
   _skip_all_partitioning(libMesh::on_command_line("--skip-partitioning")),
   _skip_renumber_nodes_and_elements(false),
+  _skip_find_neighbors(false),
   _allow_remote_element_removal(true),
   _spatial_dimension(d),
   _default_ghosting(libmesh_make_unique<GhostPointNeighbors>(*this)),
@@ -95,7 +96,8 @@ MeshBase::MeshBase (const MeshBase & other_mesh) :
 #endif
   _skip_noncritical_partitioning(false),
   _skip_all_partitioning(libMesh::on_command_line("--skip-partitioning")),
-  _skip_renumber_nodes_and_elements(false),
+  _skip_renumber_nodes_and_elements(other_mesh._skip_renumber_nodes_and_elements),
+  _skip_find_neighbors(other_mesh._skip_find_neighbors),
   _allow_remote_element_removal(true),
   _elem_dims(other_mesh._elem_dims),
   _spatial_dimension(other_mesh._spatial_dimension),
@@ -350,6 +352,12 @@ void MeshBase::prepare_for_use (const bool skip_renumber_nodes_and_elements, con
       this->allow_renumbering(false);
     }
 
+  if (skip_find_neighbors)
+  {
+    libmesh_deprecated();
+    this->allow_find_neighbors(false);
+  }
+
   // Mesh modification operations might not leave us with consistent
   // id counts, but our partitioner might need that consistency.
   if (!_skip_renumber_nodes_and_elements)
@@ -358,7 +366,7 @@ void MeshBase::prepare_for_use (const bool skip_renumber_nodes_and_elements, con
     this->update_parallel_id_counts();
 
   // Let all the elements find their neighbors
-  if (!skip_find_neighbors)
+  if (!_skip_find_neighbors)
     this->find_neighbors();
 
   // The user may have set boundary conditions.  We require that the
--- a/src/mesh/mesh_base.C
+++ b/src/mesh/mesh_base.C
@@ -321,7 +321,13 @@ void MeshBase::prepare_for_use (const bool skip_renumber_nodes_and_elements, con
 {
   libmesh_deprecated();
 
-  this->allow_renumbering(!skip_renumber_nodes_and_elements);
+  // We only respect the users wish if they tell us to skip renumbering. If they tell us not to
+  // skip renumbering but someone previously called allow_renumbering(false), then the latter takes
+  // precedence
+  if (skip_renumber_nodes_and_elements)
+    this->allow_renumbering(false);
+
+  // We always accept the user's value for skip_find_neighbors, in contrast to skip_renumber
   this->allow_find_neighbors(!skip_find_neighbors);
 
   this->prepare_for_use();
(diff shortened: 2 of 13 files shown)


lindsayad:
I don't know how these failures are happening. I didn't do anything!

lindsayad:
This is rough. I think that these failures are illustrating that we were relying on `skip_renumber_nodes_and_elements=false` not being honored in a lot of tests

lindsayad:
This is insanity. I see that I can't get rid of the default argument API because we often implicitly rely on `skip_find_neighbors=false`

jwpeterson:
So you're saying we can't not skip finding neighbors?

lindsayad:
always never skip finding neighbors

jwpeterson:
> always never skip finding neighbors

more like never start passing tests

lindsayad:
image

lindsayad:
@jwpeterson all tests pass. I'm faster than the infinite monkey

include/mesh/mesh_base.h
```diff
@@ -1016,13 +1016,17 @@ class MeshBase : public ParallelObject
...
+   * finding neighbors is also deprecated. To prevent find_neighbors, set
+   * allow_find_neighbors(false)
    *
    * If this is a distributed mesh, local copies of remote elements
    * will be deleted here - to keep those elements replicated during
    * preparation, set allow_remote_element_removal(false).
    */
-  void prepare_for_use (const bool skip_renumber_nodes_and_elements=false, const bool skip_find_neighbors=false);
+  void prepare_for_use (const bool skip_renumber_nodes_and_elements, const bool skip_find_neighbors);
+  void prepare_for_use (const bool skip_renumber_nodes_and_elelments);
```
jwpeterson: typo: "elelments"

src/mesh/mesh_base.C
```diff
@@ -96,6 +97,7 @@ MeshBase::MeshBase (const MeshBase & other_mesh) :
   _skip_noncritical_partitioning(false),
   _skip_all_partitioning(libMesh::on_command_line("--skip-partitioning")),
   _skip_renumber_nodes_and_elements(false),
+  _skip_find_neighbors(false),
```
jwpeterson: Looks like you followed what was done for the `_skip_renumber_nodes_and_elements` parameter, but I wonder if both of these shouldn't get their values from `other_mesh`?

include/mesh/mesh_base.h
```diff
@@ -1016,13 +1016,17 @@ class MeshBase : public ParallelObject
...
+  void prepare_for_use ();
```
jwpeterson: OK, so if I understand correctly, calling `prepare_for_use()` with no args is equivalent to allowing *both* renumbering and finding neighbors, i.e. the previous default values? I think I am good with that. It should be backwards compatible with all possible existing ways of calling `prepare_for_use()`, yes?

src/mesh/mesh_generation.C
```diff
@@ -1406,7 +1406,8 @@ void MeshTools::Generation::build_cube(UnstructuredMesh & mesh,
 
 
   // Done building the mesh.  Now prepare it for use.
-  mesh.prepare_for_use (/*skip_renumber =*/ false);
+  mesh.allow_find_neighbors(true);
```
jwpeterson: Based on `_skip_find_neighbors(false)` being the default, `allow_find_neighbors(true)` is the default right? Do we need this extra line? As far as I can tell this `prepare_for_use()` call was using all default args before, so calling it with no args should be equivalent...

include/mesh/mesh_base.h
lindsayad: Calling with no args will respect whatever the current values of `_skip_renumbering_nodes_and_elements` and `_skip_find_neighbors` are. If no one has touched these since normal mesh construction (e.g. not copy construction, since I'm going to try making the change you suggested), then these values are false, so we *allow* both.

src/mesh/mesh_generation.C
lindsayad: This particular line may be unnecessary, it's true. But I got paranoid after I discovered a libMesh test case where I think we did an initial `prepare_for_use(blah, skip_find_neighbors=true)` and then we performed mesh refinement later and called `prepare_for_use()` in which we relied on the implicit setting of `skip_find_neighbors=false`, so I went through libMesh and any place where we called `prepare_for_use` without any arguments (previously relying on the implicit setting of `skip_find_neighbors=false`), I added this line ahead of it to make sure that the previous behavior was replicated

lindsayad: To be clear, calling with no args after this PR is generally not equivalent to calling with no args before this PR.

jwpeterson: OK, paranoid (and more expressive in general) is always better when there's any chance of ambiguity.

src/mesh/mesh_base.C
jwpeterson: Thanks for fixing both of these.

lindsayad: :+1:

roystgnr:
Thinking about this some more ... what was the point of it?  Under what conditions do we want to tell a mesh to skip find_neighbors *by default*?  It feels like "I generated a mesh and I know it has all its neighbor links correct" was a reasonable thing to say, but "I have a mesh and I know that no future combination of operations on the mesh will leave it without valid neighbor links" would require psychic powers.

lindsayad:
https://github.com/idaholab/moose/pull/14456#issuecomment-603462594 was the motivator

roystgnr:
Ah, damn, that's a reasonable motivator.

src/mesh/mesh_base.C
```diff
@@ -316,6 +318,35 @@ bool MeshBase::has_node_integer(const std::string & name) const
...
+  libmesh_deprecated();
+
+  // We only respect the users wish if they tell us to skip renumbering. If they tell us not to
+  // skip renumbering but someone previously called allow_renumbering(false), then the latter takes
+  // precedence
+  if (skip_renumber_nodes_and_elements)
+    this->allow_renumbering(false);
+
+  this->prepare_for_use();
+}
```
roystgnr, line 347: These two implementations are no longer actually backwards compatible.  The old behavior, which we're going to revert to (albeit still deprecate) in a new PR shortly, obeys the skip_renumber_nodes_and_elements and skip_find_neighbors values for *one* prepare_for_use call, it doesn't change mesh settings *permanently*.

src/mesh/replicated_mesh.C
```diff
@@ -1428,7 +1428,8 @@ void ReplicatedMesh::stitching_helper (const ReplicatedMesh * other_mesh,
         }
     }
 
-  this->prepare_for_use( /*skip_renumber_nodes_and_elements= */ false, skip_find_neighbors);
+  this->allow_find_neighbors(!skip_find_neighbors);
```
roystgnr, line 1431: And here the same bug as https://github.com/libMesh/libmesh/pull/2473/files#r435257295 is being made *manually*: when we call ReplicatedMesh::stitching_helper now with skip_find_neighbors, instead of making the call slightly cheaper, it breaks the Mesh.  This was being covered up in CI before because the use cases were always being followed with *another* old-style prepare_for_use call that inadvertently permanently restored allow_find_neighbors(), but if code paths try to use the non-deprecated prepare_for_use overload then things simply stay broken.

src/mesh/mesh_base.C
```diff
@@ -335,20 +366,11 @@ void MeshBase::prepare_for_use (const bool skip_renumber_nodes_and_elements, con
...
-  //
   // Instances where you if prepare_for_use() should not renumber the nodes
   // and elements include reading in e.g. an xda/r or gmv file. In
   // this case, the ordering of the nodes may depend on an accompanying
   // solution, and the node ordering cannot be changed.
 
-  if (skip_renumber_nodes_and_elements)
-    {
-      libmesh_deprecated();
-      this->allow_renumbering(false);
```
lindsayad, line 350: @roystgnr you're talking about how the new code makes permanent settings while the old code did not. I agree on the `find_neighbors` front. But this is a permanent setting, isn't it? And what's worse to me is that even if a user *explicitly* passes `skip_renumber_nodes_and_elements=false`, if someone in the past called it with `skip_renumber_nodes_and_elements`, then the new request is not going to be honored

lindsayad: Moreover, we would give a deprecation warning to the first caller who passed `skip_renumber_nodes_and_elements=true` and  then honor his request. But for the guy who actually passes `skip_renumber_nodes_and_elements=false`, we give no warning at all before we potentially proceed to not do what he asks. (And I understand why we give no warning in the second case since it's a default argument...we have no way to know whether `skip_renumber_nodes_and_elements=false` has been explicitly requested or we're getting the default)

roystgnr: Oh, wow, you're right.  What an annoyingly inconsistent old behavior.  I'm not thrilled with the new APIs either (since temporary changes to the find_neighbors setting now require a variable and a call to save the old setting, then the prepare_for_use, then a call to restore the old setting), but I can't help but admit that the old API was worse.

src/mesh/unstructured_mesh.C
```diff
@@ -643,8 +644,8 @@ void UnstructuredMesh::read (const std::string & name,
     }
 
   // Done reading the mesh.  Now prepare it for use.
-  this->prepare_for_use(/*skip_renumber (deprecated)*/ false,
-                        skip_find_neighbors);
+  this->allow_find_neighbors(!skip_find_neighbors);
```
roystgnr, line 647: and here

src/mesh/unstructured_mesh.C
```diff
@@ -234,12 +234,13 @@ void UnstructuredMesh::copy_nodes_and_elements(const UnstructuredMesh & other_me
   //partitioning for now.
   this->allow_renumbering(false);
   this->allow_remote_element_removal(false);
+  this->allow_find_neighbors(!skip_find_neighbors);
```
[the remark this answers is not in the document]
lindsayad: By bug you mean I've introduced a different behavior than what we had previously? E.g. I'm setting `allow_find_neighbors` indefinitely as opposed to just for one `prepare_to_use` call? Looking back, I should have tried even harder to keep behavior consistent especially given that the APIs appear unchanged. However, I felt a little painted into a corner given that methods which take two default arguments can intercept calls with no args, one arg, or two args.

Based on your comments over on idaholab/moose#15389, I guess you're going to revert this? So it's safe to permanently not renumber, but its unsafe to permanently not allow finding of neighbors?

I guess that we are just going to give up on idaholab/moose#14456?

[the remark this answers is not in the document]
roystgnr: Not going to revert the whole PR, just introduced #2594 to fix the bugs in it.

>  So it's safe to permanently not renumber,

Not safe for efficiency in transient AMR cases, but safe for correctness at least.

> but its unsafe to permanently not allow finding of neighbors?

Yeah, in most cases.

> I guess that we are just going to give up on idaholab/moose#14456?

No, reverting the bugs here should still allow that to work, in the case where it works.  libMesh automatically disabling find_neighbors is *insane*, but that doesn't mean libMesh is going to automatically enable it either.  If the user manually disables it, on their head be it.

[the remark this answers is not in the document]
lindsayad: This all sounds good to me

PR #2474, opened 2020-03-19
Change Elem::operator== comparison

lindsayad:
Compare with centroids instead of global node ids. This is extremely
useful when comparing side elements that are coincident but may
not actually share the same nodes. See idaholab/moose#12033 for
application and #2362 for more inspiration. This probably won't
work but let's give it a shot

### DIFF
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -443,28 +443,14 @@ bool Elem::operator == (const Elem & rhs) const
   if (this->type() != rhs.type())
     return false;
 
-  const unsigned short n_n = this->n_nodes();
-  libmesh_assert_equal_to(n_n, rhs.n_nodes());
+  const Real tolerance = TOLERANCE * this->hmin();
 
-  // Make two sorted arrays of global node ids and compare them for
-  // equality.
-  std::array<dof_id_type, Elem::max_n_nodes> this_ids, rhs_ids;
+  const Real centroid_distance = (rhs.centroid() - this->centroid()).norm();
 
-  for (unsigned short n = 0; n != n_n; n++)
-    {
-      this_ids[n] = this->node_id(n);
-      rhs_ids[n] = rhs.node_id(n);
-    }
-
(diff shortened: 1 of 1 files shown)


lindsayad:
I see that this alone has no chance of finding a neighbor element through coincident sides that do not share the same global node ids since `UnstructuredMesh::find_neighbors` uses keys calculated from global node ids...

lindsayad:
I like that this fails tests where we build elements whose nodes are all located at the origin

jwpeterson:
I don't think we should change the behavior of `Elem::operator==()`, at least not without deprecating it, etc. first, since as you saw there will probably be code relying on it. Can you add a different comparison operator that does what you want?

jwpeterson:
Also, does this test even make sense? If I have two "concentric" Hexes that share the same centroid but one is 100x larger than the other they would compare equal in your test, no?

lindsayad:
Yea they would. And even if checking both the centroid and the volume wouldn't be sufficient to determine geometric equality since say for a quad, one could just be rotated by 45 degrees or something.

What  about using node coordinates  instead of IDs? I believe we do that  for the hilbert keys for adaptivity...

jwpeterson:
> What about using node coordinates instead of IDs? I believe we do that for the hilbert keys for adaptivity...

For your comparison operator, yeah, something like that would probably make sense though you will have to do some kind of "max min" check (for each node of Elem A, find the min distance node of Elem B, then take the max of all those distances) since you want to allow the nodes to be in any order.

For `Elem::operator==()` I think we have to keep it as-is (no floating point comprisons, etc.) for backwards compatibility. It was created to check whether two elements have exactly the same connectivity (the assumption being that they are in the same Mesh) whereas the comparison you are talking about the Elems can be from different meshes, etc.

lindsayad:
That's fair

lindsayad:
Per https://github.com/idaholab/moose/pull/14456#issuecomment-601709543 and the reasons here, closing this

PR #2475, opened 2020-03-20
Parameters support for std::map

roystgnr:
This should handle the libMesh side of the support required for https://github.com/idaholab/moose/issues/14894

The print_helper expansion here is in line with what we're already doing, but in the long term we should factor it into a separate class as discussed in the MOOSE issue.


roystgnr:
This is pretty straightforward and adds test coverage, so merging straight away.

PR #2479, opened 2020-03-24
More robust Partitioner::set_node_processor_ids

roystgnr:
Hopefully this will fix the partitioning inconsistencies that we've been
managing to trigger with new mesh generators.

This is kind of a low level change, and I've only done some basic testing with this so far, so even if @fdkong finds it working on his problem I'm going to want to hammer it with other tests before we merge.

### DIFF
--- a/src/mesh/mesh_generation.C
+++ b/src/mesh/mesh_generation.C
@@ -17,11 +17,6 @@
 
 
 
-// C++ includes
-#include <cstdlib> // *must* precede <cmath> for proper std:abs() on PGI, Sun Studio CC
-#include <cmath> // for std::sqrt
-
-
 // Local includes
 #include "libmesh/mesh_generation.h"
 #include "libmesh/unstructured_mesh.h"
@@ -54,6 +49,13 @@
 #include "libmesh/enum_order.h"
 #include "libmesh/int_range.h"
 #include "libmesh/parallel.h"
+#include "libmesh/parallel_ghost_sync.h"
+
+// C++ includes
+#include <cstdlib> // *must* precede <cmath> for proper std:abs() on PGI, Sun Studio CC
+#include <cmath> // for std::sqrt
+#include <unordered_set>
+
 
 namespace libMesh
 {
@@ -1538,6 +1540,9 @@ void MeshTools::Generation::build_sphere (UnstructuredMesh & mesh,
 
   BoundaryInfo & boundary_info = mesh.get_boundary_info();
 
+  // Building while distributed is a little more complicated
+  const bool is_replicated = mesh.is_replicated();
+
   // Sphere is centered at origin by default
   const Point cent;
 
@@ -1867,8 +1872,10 @@ void MeshTools::Generation::build_sphere (UnstructuredMesh & mesh,
   // Loop over the elements, refine, pop nodes to boundary.
   for (unsigned int r=0; r<nr; r++)
     {
-      // A DistributedMesh needs a little prep before refinement
-      if (!mesh.is_replicated())
+      // A DistributedMesh needs a little prep before refinement, and
+      // may need us to keep track of ghost node movement.
+      std::unordered_set<dof_id_type> moved_ghost_nodes;
+      if (!is_replicated)
         mesh.prepare_for_use();
 
       mesh_refinement.uniformly_refine(1);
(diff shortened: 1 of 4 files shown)


fdkong:
I will try with my new mesh generator to check if it is able to fix the issue

roystgnr:
Well, no rush; it's definitely caused regressions.  I can replicate a failure here, so let me see how hard it'll be to fix.

roystgnr:
> let me see how hard it'll be to fix.

And the answer is: "hard".  I was trying to sync everything here with a series of "pull" operations, but I think I've got a corner case where a chicken-and-egg problem is going to force an initial "push", but it's not immediately obvious how to avoid that push being huge.

fdkong:
@roystgnr  Node ID is always unique? The nodes with the same ID on different MPI ranks are the same node?

roystgnr:
Victory!  At least on the default CI tests!

Okay, @fdkong; I'm going to add some more optional moosebuild tests, but now that this is passing the basics and passing the stuff I've thrown at it manually it's probably in good enough shape for you to try it on the previously-failing cases on your generator.

> Node ID is always unique? The nodes with the same ID on different MPI ranks are the same node?

If you mean "is" in the normative sense, then **yes**: the Node object with the same ID on different ranks should always refer to the same node.

If you mean "is" in the positive sense, then **hopefully**, but getting nodes and/or elements out of sync an easy mistake for DistributedMesh modification code to make, so we spend a crazy amount of time in dbg mode trying to catch any such mistakes ASAP.

roystgnr:
Since when do we get 5MB->output_size_limit_exceeded spews of inverse_map errors on every dbg mode "make check"???

Pinging @jwpeterson - this looks like it predates #2498 but I just don't understand what's going on here; I don't get anything like this when running tests myself.

roystgnr:
I can replicate this, and ironically it looks like the extra testing here revealed a bug in the build_sphere code I thought I'd just finished fixing...

fdkong:
@roystgnr When you are ready. I will take try to see if my cases work with this change.

roystgnr:
Inside-view Roy says "I've finally looked at this from every angle; it's completely robust and ready for you to test!"

Outside-view Roy says "That's what I thought the last two times too, and I was wrong then, so I'm probably wrong now."

It'd be safest to wait at least until the Distributed make check sweep makes it to "Check 10" before you spend any time running tests yourself.  I'll also comment again here if any of my private testing hits a failure.

roystgnr:
Hit another bug, but it seems to be specific to subactive element handling, in which case I'd expect CI to fail here sooner or later but it would still be safe to test this hash on a mesh generator.

roystgnr:
Where by "hash" I of course mean both [this](https://github.com/libMesh/libmesh/pull/2479/commits/3fdd5c974de0454bb5edf774aa13e697cd2174ab) and [this](https://www.merriam-webster.com/dictionary/make%20a%20hash%20of).

roystgnr:
Okay, that last test failure ought to be fixed by https://github.com/idaholab/moose/pull/15089 - once that makes it to next+CI, I'll clean up the git history here, then we can merge (assuming no other regressions found).

src/mesh/mesh_generation.C
```diff
@@ -54,6 +49,13 @@
 #include "libmesh/enum_order.h"
 #include "libmesh/int_range.h"
 #include "libmesh/parallel.h"
+#include "libmesh/parallel_ghost_sync.h"
+
+// C++ includes
+#include <cstdlib> // *must* precede <cmath> for proper std:abs() on PGI, Sun Studio CC
```
dschwen, line 55: 🤡

roystgnr: Yeah, I doubt that comment has remained relevant ever since someone (Ben?) started that workaround a decade ago; I can't imagine any compiler version with half-broken std::abs is managing to handle C++11 fine.  But meh, couldn't kill us to preserve non-alphabetical-order in a few places.

roystgnr:
Well, this is annoying.  I can no longer replicate that test failure, even when specifically using the same MOOSE commit (b3aa59bad2) that's checked out in the CI logs.

fdkong:
Did you check if the same petsc was used?

roystgnr:
Oh, wow, good point.  You're relying on them for partitioning, right?  And CI is using 3.11.4 while I'm on 3.12.4.  Let me just try cycling through a bunch of different mesh sizes, then and see if I can get any to scream at me.

roystgnr:
Looping over nproc from 2 to 16, nx==ny==nz from 3 to 20, and dbg-mode tests still passed for me every time.

fdkong:
> Oh, wow, good point. You're relying on them for partitioning, right?

Strictly speaking, for this particular test case, we use parmetis via the PETSc interfance.  The same parmetis is used in the libmesh as well.

For certain cases, partition from PETSc-3.12.5 could be different from PETSc-3.11.4 because the underlying parmetis is different.  Did you try 3.11.4?

roystgnr:
Okay, tried with 3.11.4 finally ... and same deal: the failing unit test still passes on my system.

Could you give this PR a try, see if you can replicate the problem?  If not then at this point I'm going to have to try printf-debugging iterating over Civet-output, which might take a while...

roystgnr:
My printf-debugging PR simply *passed*, so maybe that last CI run here simply hadn't pulled the relevant fixes in MOOSE after all?  Rebasing and pushing to kick off another run here; if this works too then I'll merge.

roystgnr:
Looks like the distributed_rectilinear_mesh_generator test worked, but the new unit test from #2526 failed to account for DistributedMesh.  Easily fixed.

fdkong:
Cool!

PR #2481, opened 2020-03-25
Fix BoundingBox intersection test for QuadTree

jwpeterson:
Previously the code used generic BoundingBox::intersects() checks but this isn't robust in the QuadTree (LIBMESH_DIM==2 or is_planar_xy) cases because the BoundingBoxes are infinitesimally thin in the z-direction, and likely to produce spurious intersection failures.


jwpeterson:
Also added a unit test that fails before this change, but passes after it.

PR #2482, opened 2020-03-25
MeshGeneration unit tests, fix ReplicatedMesh::n_* for holes in numbering

roystgnr:
Aside from adding a little more test coverage for basic code that should really never ever break, this exposes the bugs fixed by #2480 - the broken cases now fail if I don't have that patch included.

### DIFF
--- a/include/mesh/replicated_mesh.h
+++ b/include/mesh/replicated_mesh.h
@@ -101,10 +101,10 @@ public:
   virtual void renumber_nodes_and_elements () override;
 
   virtual dof_id_type n_nodes () const override
-  { return cast_int<dof_id_type>(_nodes.size()); }
+  { return _n_nodes; }
 
   virtual dof_id_type parallel_n_nodes () const override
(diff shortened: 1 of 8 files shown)


roystgnr:
No, wait, they fail both ways.  I miscounted something on the node tests for those cases.

roystgnr:
No, it's not a miscount on *my* end, it's a miscount in ReplicatedMesh::n_nodes() when there are "holes" in the list of nodes by id, which is inevitable whenever we delete nodes but we aren't allowed to renumber nodes above them.  Not sure what to do about that except make a ReplicatedMesh::_n_nodes to keep track of things separately from the vector length.  (and similarly for _n_elem)  Hopefully that won't be too bad, since most of the mistaken n_nodes == max_node_id assumptions have already been caught years ago by DistributedMesh usage.

roystgnr:
This is a pretty scarily low-level change, but to two functions that have massive test coverage, and it's definitely a bugfix in the deleting-without-renumbering case, so I'm merging now.

PR #2485, opened 2020-03-26
Reset ReplicatedMesh::_n_* during clear()

roystgnr:
This bug just got introduced by #2482, but somehow doesn't get triggered
by our usual test suite unless we configure with --disable-lgpl-only
(after which point the ElemCutter is using clear() to reuse its internal
Mesh).

### DIFF
--- a/src/mesh/replicated_mesh.C
+++ b/src/mesh/replicated_mesh.C
@@ -643,6 +643,7 @@ void ReplicatedMesh::clear ()
   for (auto & elem : _elements)
     delete elem;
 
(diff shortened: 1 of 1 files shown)


roystgnr:
Throwing the Non-LGPL Contrib CI at it this time, as well as more app tests just to make up for my inadequate paranoia earlier.

permcody:
Our internal gitlab server got upgraded last night. Jason just upgraded the CIVET server. Hopefully things stabilize in a bit.

Issue #2484, opened 2020-03-26
I seem to find some problems with uniqueID？


ermore:
// I understand that elems returned by  the function (mesh.active_element_ptr_range() ) include local elems and ghost elems
    // in each process The activate Elem should have a different unqiueid number

When I compile and execute the following code with command `mpiexec -n 2 ./ex`:

```c++

int main(int argc, char **argv)
{
    LibMeshInit init(argc, argv);
    Mesh mesh(init.comm(), 3);

    MeshRefinement mesh_refinemesh(mesh);

    int K = 3;
    MeshTools::Generation::build_cube(mesh,
                                      4 * K,
                                      2 * K,
                                      2 * K,
                                      0., 1. * 2,
                                      0., 1,
                                      0., 1,
                                      HEX8);

    // I understand that elems returned by  the function (mesh.active_element_ptr_range() ) include local elems and ghost elems
    // in each process The activate Elem should have a different unqiueid number

    int adapt = 1;
    for (adapt = 1; adapt < 4; ++adapt)
    {
        libMesh::out << "adaption:" << adapt << std::endl;
        for (auto &elem : mesh.active_local_element_ptr_range())
        {
            if (elem->processor_id() == 0)
            {
                elem->set_refinement_flag(Elem::REFINE);
            }
        }
        MPI_Barrier(MPI_COMM_WORLD);
        mesh_refinemesh.refine_and_coarsen_elements();

        processor_id_type processor_rank = mesh.comm().rank();
        processor_id_type processor_size = mesh.comm().size();
        std::set<unique_id_type> uniqueSet;

        unsigned int active_elem_size = std::distance(mesh.active_elements_begin(), mesh.active_elements_end());

        for (auto const &elem : mesh.active_element_ptr_range())
        {
            uniqueSet.insert(elem->unique_id());
        }

        std::cout << "rank: " << processor_rank << " uniqueSset size: " << uniqueSet.size() << " ; activate size: " << active_elem_size << std::endl;
    }

    return 0;
}
```
I get the following results:
```
adaption:1
adaption:1
adaption:1
rank0 uniqueSset size: 684 ; activate size: 684
adaption:2
rank1 uniqueSset size: 621 ; activate size: 621
adaption:2
rank2 uniqueSset size: 706 ; activate size: 706
adaption:2
rank0 uniqueSset size: 2126 ; activate size: 2126
adaption:3
rank1 uniqueSset size: 2140 ; activate size: 2140
adaption:3
[...]

PR #2486, opened 2020-03-28
Fixes for DistributedMesh unique_id generation

roystgnr:
This is much better than our previous usually_unique_id generation.

I'll throw a heavyweight dbg-only test for uniqueness in shortly to make
sure that this is the last of the necessary fixes and that it doesn't
regress again.

Refs #2484


roystgnr:
I'm hoping to merge this right away, since it fixes at least some cases that trigger serious errors; I'll get the added test coverage in a later PR.

roystgnr:
And by "right away" of course I mean "right away *after* completing the appropriate optional CI" which I probably should have remembered to turn on last night...

PR #2487, opened 2020-03-28
macrofice second-derivative quantities

BalticPinguin:
Again, some second derivatives sneaked outside the ``ENABLE_SECOND_DERIVATIVE``-macro.

In ``FEMContext::build_new_fe`` I am not sure if one should probably throw a ``libmesh_error()`` instead of just skipping?


roystgnr:
libmesh_not_implemented() might be most appropriate?  We don't really have a macro for "you need different configure options" except for the one that skips examples.

BalticPinguin:
ok, than I'll use this one. Just to have an error when requesting ``get_derivative_level`` of ``2``...

roystgnr:
Thanks; sorry to let this slip.

PR #2491, opened 2020-03-30
Testing unique_id uniqueness

roystgnr:
This adds (dbg-mode-only) test coverage that would have caught the bugs in #2484, and a fix for another bug that it's already caught.  I'm still running tests myself too; once those and/or CI passes then I'll start the heavyweight optional CI tests.

### DIFF
--- a/include/base/dof_object.h
+++ b/include/base/dof_object.h
@@ -146,9 +146,16 @@ public:
   unique_id_type unique_id () const;
 
   /**
-   * \returns The globally \p unique_id for this \p DofObject as a writable reference.
+   * \returns The globally \p unique_id for this \p DofObject as a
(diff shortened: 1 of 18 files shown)


roystgnr:
This is finding more bugs but it's also way too slow for regular use as-is...

roystgnr:
And that's because I'm an idiot and missed an ampersand in a refactor.

roystgnr:
This latest swath of commits still won't pass CI until after https://github.com/idaholab/moose/pull/15012 gets into MOOSE, but between them and the MOOSE PR I'm no longer seeing any of the failures in that last CI run.

roystgnr:
Okay, we should be good to go; planning to merge after CI is all done.

PR #2492, opened 2020-03-30
Add capability to read Exodus file header only

jwpeterson:
This is useful if you just need to know how many Nodes or Elems a Mesh has but don't want to spend the time reading/processing the entire Exodus file. Also added a unit test of the new capability.

### DIFF
--- a/include/mesh/exodusII_io_helper.h
+++ b/include/mesh/exodusII_io_helper.h
@@ -27,6 +27,7 @@
 #include "libmesh/point.h"
 #include "libmesh/boundary_info.h" // BoundaryInfo::BCTuple
 #include "libmesh/enum_elem_type.h" // INVALID_ELEM
+#include "libmesh/exodus_header_info.h"
 
 // C++ includes
 #include <iostream>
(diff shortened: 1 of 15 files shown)


include/mesh/exodusII_io_helper.h
```diff
@@ -501,36 +509,42 @@ class ExodusII_IO_Helper : public ParallelObject
...
 
-  // Number of global variables
-  int num_global_vars;
+  // Problem title (Use vector<char> to emulate a char *)
+  std::vector<char> & title;
 
-  // Number of sideset variables
-  int num_sideset_vars;
+  // Number of dimensions in the mesh
+  int & num_dim;
```
roystgnr: These are basically backwards compatibility shims to the ExodusHeaderInfo entries?  Do we need them?  Was anyone ever using ExodusII_IO_Helper directly?

I'm in favor of the shims either way, but maybe we could wrap them in an ```#ifdef LIBMESH_ENABLE_DEPRECATED``` and so hopefully get rid of them in another year?

jwpeterson: We don't *need* them but it made the current diff much smaller, and I think I like the code better with them (easier to read and less typing) but if you really want to we could get rid of them.

roystgnr: Hmm... I guess I'm paying attention to the visible redundant lines here and not to the hypothetical redundant tokens you'd need to add elsewhere otherwise.  Okay, I'm happy with it.

jwpeterson:
Thanks!

PR #2495, opened 2020-04-03
FileSolutionHistory class

roystgnr:
This is going to see more work for adjoint-based adaptivity and async I/O in the future, but the proof of concept is working now for adjoint-based sensitivities.


roystgnr:
Merging - this doesn't really mess with any code outside the SolutionHistory classes.

PR #2498, opened 2020-04-07
Add 'extra_checks' parameter to FEMap::inverse_map()

jwpeterson:
Narrows the scope of the `secure` flag, which was previously used for all the different kinds of checks, so that it now does not include the "sanity" checks which occur after the Newton iteration has converged. These additional post-convergence checks are now controlled by the `extra_checks` parameter. By default, `extra_checks == secure == true` so this is backwards-compatible with the previous behavior.

Also, improved the documentation of the `secure` flag, which was somewhat lacking previously.


jwpeterson:
Thanks!

PR #2502, opened 2020-04-10
The 'extra_checks' parameter should match 'secure' for backwards comp…

jwpeterson:
…atibility.

This means we have to explicitly set 'extra_checks' to false in cases
where 'secure' is set to false. Also fixed up some documentation that
still referred to the now deprecated FEInterface::inverse_map().

Refs #2498

### DIFF
--- a/include/fe/fe.h
+++ b/include/fe/fe.h
@@ -424,7 +424,7 @@ public:
                             const Real tolerance = TOLERANCE,
                             const bool secure = true) {
     // libmesh_deprecated(); // soon
-    return FEMap::inverse_map(Dim, elem, p, tolerance, secure);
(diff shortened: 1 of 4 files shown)


jwpeterson:
This fixes the issue of too much output in debug mode which @roystgnr found in #2479, and was definitely caused by the change in #2498 not being thorough enough. 

@milljm Do you think it would be possible to have a moosebuild option where we consider output files > X MB as a "failure" (default to X=5, say) even if no tests actually failed? That would have led me to catch this issue a bit earlier. Nothing urgent; I can open a ticket for it if you prefer.

milljm:
@jwpeterson Yeah, I can set that up. I thought we already had a system like this in place for moose. I just need to find it.

Issue #2496, opened 2020-04-05
Error running MeshTools::Genertion::build_sphere() !


ermore:
When I run the following code:
```c++
MeshTools::Generation::build_sphere (mesh, 10.0, 2, HEX27, 2, true);
```
I get an error:
```
[0]PETSC ERROR: ------------------------------------------------------------------------
[0]PETSC ERROR: Caught signal number 11 SEGV: Segmentation Violation, probably memory access out of range
[0]PETSC ERROR: Try option -start_in_debugger or -on_error_attach_debugger
[0]PETSC ERROR: or see https://www.mcs.anl.gov/petsc/documentation/faq.html#valgrind
[0]PETSC ERROR: or try http://valgrind.org on GNU/linux and Apple Mac OS X to find memory corruption errors
[0]PETSC ERROR: configure using --with-debugging=yes, recompile, link, and run 
[0]PETSC ERROR: to get more information on the crash.
[0]PETSC ERROR: --------------------- Error Message --------------------------------------------------------------
[0]PETSC ERROR: Signal received
[0]PETSC ERROR: See https://www.mcs.anl.gov/petsc/documentation/faq.html for trouble shooting.
[0]PETSC ERROR: Petsc Release Version 3.12.1, Oct, 22, 2019 
[0]PETSC ERROR: ./example-dbg on a arch-linux2-c-opt named ermore by ermore Sun Apr  5 09:26:57 2020
[0]PETSC ERROR: Configure options -with-debugging=0 --download-hdf5 --download-hypre --download-metis --download-parmetis --download-ptscotch --download-zoltan --download-mumps --download-scalapack --download-suitesparse --download-superlu --download-superlu_dist --download-slepc --download-sundials --download-silo --download-zliba
[0]PETSC ERROR: #1 User provided function() line 0 in  unknown file
--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
with errorcode 59.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
```

roystgnr:
We don't seem to have any test coverage for that function.  Let me add some real quick and see if I can replicate this.

In the meantime - are you running in dbg mode?  Regardless of regressions I'd hope we at *least* hit an assert before getting a segmentation fault.

roystgnr:
[...]

PR #2503, opened 2020-04-10
Fixes for build_sphere(DistributedMesh)

roystgnr:
Ironically, this fixes every supported element type *except* the HEX27 case which prompted this investigation in #2496 in the first place.  There's still one bug remaining in the all_second_order distributed unique_id handling that I haven't yet sorted out.  But I'm done for the day so we can at least get the other fixes into CI first.


roystgnr:
Ok, with that last fix everything's happy for me with 16-proc DistributedMesh runs too.  This should finally be ready to merge once CI is happy.

PR #2504, opened 2020-04-12
Fix adjoint Dirichlet constraints with AMR

vikramvgarg:
We were not applying hanging nodes constraints correctly for nodes connected to adjoint Dirichlet boundaries. Adjoints example 6 was showing spurious values at such hanging nodes. 

Fixed by correctly expanding adjoint dirichlet constraints. 

Adjoints example 6 works properly with AMR now.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -3346,6 +3346,22 @@ void DofMap::process_constraints (MeshBase & mesh)
     check_for_constraint_loops();
   }
 
+  // Adjoints will be constrained where the primal is
+  // Therefore, we will expand the adjoint_constraint_values
+  // map whenever the primal_constraint_values map is expanded
+
+  // First, figure out the total number of QoIs
+  const unsigned int max_qoi_num =
+    _adjoint_constraint_values.empty() ?
+    0 : _adjoint_constraint_values.rbegin()->first+1;
+
+  // A vector to hold pointers to the adjoint constraints map for each adjoint variable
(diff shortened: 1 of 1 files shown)


roystgnr:
@vikramvgarg has some ideas for better test coverage on this, but we'll have a separate PR for those.

roystgnr:
Turns out we *did* have some better test coverage on this, just not inside libMesh itself.

src/base/dof_map_constraints.C
```diff
@@ -3346,6 +3346,22 @@ void DofMap::process_constraints (MeshBase & mesh)
...
+  const unsigned int max_qoi_num =
+    _adjoint_constraint_values.empty() ?
+    0 : _adjoint_constraint_values.rbegin()->first+1;
+
+  // A vector to hold pointers to the adjoint constraints map for each adjoint variable
+  std::vector<DofConstraintValueMap *> _adjoint_constraints_maps;
+  _adjoint_constraints_maps.resize(max_qoi_num);
+
+  for ( unsigned int j = 0; j < max_qoi_num; j++)
+    _adjoint_constraints_maps[j] = &(_adjoint_constraint_values.find(j)->second);
```
roystgnr, line 3363: All is madness!

This is an obvious bug whenever find() returns end(), but not only did I not see that before, the code worked fine in CI for me too.  I'm not sure what about my OS or compiler or phase of the moon has changed since, but I get crashes in 4 or 5 GRINS tests now when the later code here tries to dereference end().  I'll fix it.

PR #2453, opened 2020-02-24
Update copyright notice for 2020.

jwpeterson:
Nothing to see here, just using CI to make sure perl scripts didn't do anything crazy.

PR #2506, opened 2020-04-15
WIP: Infinite Elements: setup of properties

BalticPinguin:
Finally, I have some fix for #2453, even though it is not ready yet.

So far, I have adapted the main structure of the class to work correctly. The main problem is:
  - Jacobian diverges with r^4 (in general r^(dim+1))
  - phi goes with 1/r (in general r^{(dim-1)/2} )
  - dphi goes with 1/r^2 ( r^{(dim+1)/2} )

 With this, a concise notation of the quantities actually used (J over r^4, phi times r etc.) is not very easy.
I decided to introduce some ``decay`` which is r^(dim+1)/2; and ``phi``-> ``phi_over_decayxr`` should mean  ``phi *r / decay``.  (see changes in misc_1)

Any better ideas are welcome!

Unfortunately, I didn't find some consise literature for dealing with these integrals, so I started to prepare a document myself to which one can refer the user.

In this context, open tasks are
- [ ]  adjust second example
- [ ] adjust further fe_map-related quantities:
      * get_tangents
      * get_normals
      * ...
 - [ ] adjust tests
 - [ ]  API of map: remove false genarality

The last step refers to ``InfFEMap::eval`` which maps ``v`` to ``r`` using effectively the form ``r=2*a/(1-v)`` (pretty intransparent, though); while the notation comes from a more general form, this generality anyway is not supported. So maybe one can remove it completely or change to a form that actually supports it...


BalticPinguin:
So, finally I am back with this one: It was more than I expected, but finally everything is implemented, I guess.
I will provide also some documentation/derivation of the terms to make it easier to follow what is behind these changes.

Most of the new variables and functions have VERY clumbsy names; If anyone has good ideas for a better scheme, you are very welcome.

include/fe/fe_abstract.h
```diff
@@ -244,35 +244,68 @@ class FEAbstract : public ReferenceCountedObject<FEAbstract>
...
+  virtual
+#endif
   const std::vector<Point> & get_xyz() const
   { calculate_map = true; return this->_fe_map->get_xyz(); }
 
+
+#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+  /**
+   * This function is the variant of \p get_JxW() for \p InfFE.
+   * Since J diverges there, a respectize decay-function must be
```
jwpeterson, line 261: Typo: "respectize" -> may be a better word here is "suitable" or something along those lines.

include/fe/fe_abstract.h
```diff
@@ -244,35 +244,68 @@ class FEAbstract : public ReferenceCountedObject<FEAbstract>
...
+   * applied to obtain well-defined quantities.
+   *
+   * For FE, it is equivalent to the common \p get_JxW().
+   */
+  virtual const std::vector<Real> & get_JxWxdecay_sq () const
```
jwpeterson, line 266: Is it true that someone using `InfFEs` should *only* use the "decay" JxW or do they need to have access to the non-decay JxW as well? As for a name I would prefer `JxW_decay` or `decay_JxW` to this one.

include/fe/fe_abstract.h
```diff
@@ -244,35 +244,68 @@ class FEAbstract : public ReferenceCountedObject<FEAbstract>
...
+#endif
+
   /**
    * \returns The element Jacobian times the quadrature weight for
    * each quadrature point.
+   *
+   * For \p InfFE, use \p get_JxWxdecay_sq() instead.
    */
+#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+  virtual
```
jwpeterson, line 277: IMO it's OK if these functions become `virtual` regardless of whether infinite elements are enabled. There may be other drawbacks but this is not usually called in performance-sensitive parts of the code and it would certainly look a lot cleaner to drop the `ifdef`.

include/fe/inf_fe.h
```diff
@@ -528,8 +534,229 @@ class InfFE : public FEBase
...
+   * \returns The element tangents in xi-direction at the quadrature
+   * points.
+   */
+  const std::vector<RealGradient> & get_dxyzdxi() const override
+  { calculate_map = true; libmesh_not_implemented();}
+  /**
+   * \returns The element tangents in eta-direction at the quadrature
+   * points.
+   */
+  const std::vector<RealGradient> & get_dxyzdeta() const override
```
jwpeterson, line 612: Isn't this the same as the base class implementation? I'm not sure why you need all these overrides, probably I am missing something obvious though.

src/fe/inf_fe_map.C
```diff
@@ -27,11 +27,43 @@
...
 #include "libmesh/libmesh_logging.h"
 
+#include "libmesh/dense_matrix.h"
+#include "libmesh/analytic_function.h" // for DenseVector
+
 namespace libMesh
 {
 
+   // local helper-function solving Ax=b in 3D.
+bool system_solve_3x3(DenseMatrix<libMesh::Real> & A, DenseVector<libMesh::Real> & b, DenseVector<libMesh::Real> & x)
```
jwpeterson, line 37: If it's possible to use a `TypeTensor` here instead of a `DenseMatrix`, there is already a `solve()` member on that. It would be good not to repeat this code for solving a 3x3 system if possible.

tests/fe/inf_fe_radial_test.C
```diff
@@ -16,9 +18,43 @@
 #include <vector>
 
 #include "libmesh_cppunit.h"
+#include "libmesh/dense_matrix.h"
+#include "libmesh/analytic_function.h" // for DenseVector
+#include "libmesh/fe.h"
 
 using namespace libMesh;
 
+bool system_solve_3x3(DenseMatrix<libMesh::Real> & A, DenseVector<libMesh::Real> & b, DenseVector<libMesh::Real> & x)
```
jwpeterson, line 27: And now `system_solve_3x3` is repeated twice... would definitely be good to avoid this.

jwpeterson:
Also if the refactoring in #2597 gets merged before this PR, it will most likely conflict. If that happens, I can take a shot at fixing the conflicts on this branch.

include/fe/fe_abstract.h
BalticPinguin: oh, I think I just meant respective :-) Thanks

src/fe/inf_fe_map.C
BalticPinguin: great, I was not aware of this.

include/fe/inf_fe.h
BalticPinguin: well, the difference is that usually this calls ``_fe_map->get_dxyzdzeta()`` but for infinite elements, ``_fe_map`` is not in use (anymore)...

include/fe/fe_abstract.h
BalticPinguin: > I would prefer JxW_decay or decay_JxW to this one.

You are right, this looks much cleaner (I especially like the idea of adding ``decay`` in front), but I wanted the ``sq`` because the ``decay``-factor here is the square of that which is compensated for in the shape functions.

> Is it true that someone using InfFEs should only use the "decay" JxW 

yes: JxW is a function that (when written in physical space) which grows like r^4; this behaviour is the main reason for all these changes:
  * Before, it was computed via Lagrange-interpolation on the reference-element, where the local coordinate ``zeta = 1-2a/r``, i.e. JxW goes like ``1/zeta^4``: being non-polynomial, the interpolation won't work properly
  * proper evaluation of the integral **must** yield a divergent ``JxW`` (because the volume is infinitely large)
  * the integral over the full matrix elements does not diverge because the shape functions decay fast enough. So I introduced some scaling in ``JxW`` which is compensated in ``phi``, avoiding divergences and still having the correct matrix elements.

**I'll try to get the 'documentation' mentioned above into shape asap; this might help to grasp the ideas behind**, than it probably is easier to discuss the naming-schemes and you have a chance to follow my changes. _I think I should have started with sharing the document and not purely the code; sorry for that_

include/fe/fe_abstract.h
roystgnr: I like leaving the ifdef, but I'm happy either way.  Indeed every call to this should either be outside the element loop (where the performance difference is trivial) or at least outside a loop over all test functions on all quadrature points, where the performance difference is nearly trivial.

roystgnr: Although... could we turn the zillion ifdef statements into just *one*?  Something like

```
#define virtual_for_inffe virtual
#else
#define virtual_for_inffe
#endif
```

BalticPinguin:
Sorry that this took so long now; but finally I have added some tests and corrected several mistakes I have found.
Among others, I have added the ``get_JxW()`` again: (At least for the given examples), the Jacobian does not contain 'Inf's, but you cannot sum them up without geting 'NaN's. So it seems to work but I'd expect it to be numerically rather instable.

I am sure you'll find some places where the code can be improved; I think, the main changes to come are names for the new functions. The extra weights of these quantities are summarized on page 4 of this document: [radial_mapping.pdf](https://github.com/libMesh/libmesh/files/5122007/radial_mapping.pdf) Probably this helps finding a better naming scheme.
The document summarizes my thoughts behind the changes implemented here; I wanted to have it all in one, more or less self-explaining document. I will upload it to Archive to make it available as reference for later users. (so if you have comments on the text, they are very welcome as well).

Best regards

BalticPinguin:
pinging @jwpeterson for comments

BalticPinguin:
I know I took quite some time on this, but probably you can help me now to finally get this PR done, @jwpeterson @roystgnr ?
As I see it, the main question is to get a more convenient naming scheme for the new functions. But I'm sure you find several further things to be improved...

Best regards

BalticPinguin:
Dear @jwpeterson , @roystgnr ,
I'd like to discuss the content of this PR with you.
I am aware that some parts of this code don't look very well but probably you can help me to go through these things to obtain a good fix for the addressed problems?
Best regards,
Hubert

jwpeterson:
Hi @BalticPinguin sorry for the lack of response on my end. I will try and take a look at this and comment tonight!

jwpeterson:
I rebased and cleaned up your branch, but since you did all this work on your repo's master branch, I didn't want to push directly to it. Instead I opened up #2750. Do you mind if we close this and move the discussion over to there instead?

PR #2508, opened 2020-04-17
Bugfix in rb_construction_base.C

dknez:
Previously the index of n_training_samples_per_param was going out of bounds.

### DIFF
--- a/src/reduced_basis/rb_construction_base.C
+++ b/src/reduced_basis/rb_construction_base.C
@@ -464,7 +464,11 @@ void RBConstructionBase<Base>::generate_training_parameters_deterministic(const 
       count++;
     }
 
-  std::vector<unsigned int> n_training_samples_per_param(num_params);
+  // n_training_samples_per_param has 3 entries, but entries after "num_params"
+  // are unused so we just set their value to 1. We need to set it to 1 (rather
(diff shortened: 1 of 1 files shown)


roystgnr:
Reminder: whenever we merge this, we ought to change one of the Civet recipes (Test debug?  Minimum gcc version debug?) to use --enable-glibcxx-debugging so we don't have a regression of this type again.

jwpeterson:
> Reminder: whenever we merge this, we ought to change one of the Civet recipes (Test debug? Minimum gcc version debug?)

I'd prefer not to change an existing recipe, since that disables the dbg-mode unit tests from running, which are also useful for catching asserts before code gets merged. Maybe add a new optional "glibcxx" CIVET recipe?

roystgnr:
My hope was that if we changed one debug recipe (to get GLIBCXX_DEBUG coverage on the examples) and left the other unchanged (to get dbg coverage on the unit tests) that'd give us a strict increase in coverage without increasing CI expense.

jwpeterson:
> that'd give us a strict increase in coverage without increasing CI expense.

OK, sounds good. Does Test Debug actually use GCC? If not, then I guess we should go with Min gcc.

PR #2509, opened 2020-04-19
Fix flux error estimates when flux weights not in approximation space

vikramvgarg:
To capture boundary data representation errors with adjoints, we needed to avoid applying adjoint Dirichlet boundary conditions on the coarse scale adjoint vectors in the projection step after enriching the mesh.

Also fixed a bug in adjoints example 6 which was preventing the coarse adjoints from being projected at all.

Tested on fixed adjoints example 6, matches error estimation identities, expected effectivities and gives an OOM improvement in boundary flux QoI convergence.

### DIFF
--- a/examples/adjoints/adjoints_ex6/adjoints_ex6.C
+++ b/examples/adjoints/adjoints_ex6/adjoints_ex6.C
@@ -283,9 +283,6 @@ int main (int argc, char ** argv)
 
   equation_systems.init ();
 
-  // Add an adjoint_solution0 vector to the system
-  system.add_vector("adjoint_solution0", false, GHOSTED);
-
   // Print information about the mesh and system to the screen.
   mesh.print_info();
   equation_systems.print_info();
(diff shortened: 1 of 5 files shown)


examples/adjoints/adjoints_ex6/adjoints_ex6.C
```diff
@@ -283,9 +283,6 @@ int main (int argc, char ** argv)
 
   equation_systems.init ();
 
-  // Add an adjoint_solution0 vector to the system
-  system.add_vector("adjoint_solution0", false, GHOSTED);
```
roystgnr, line 287: This is arguably a bug in main() in this particular case, but in general shouldn't we be fixing this incompatibility inside ARefEE?  If somebody's turned off projections of any vector we need projected, the right thing to do is "temporarily turn projections back on before we do our refinement, then restore the old setting after we're done", not "silently return bad error indicator values", no?

vikramvgarg: I didnt do that inside ARefEE since projections on is default behavior. But you are right, we can save the existing projection setting, temporarily turn projections on and restore back like we do for other settings.

Let me make and commit those changes in.

roystgnr:
Let me do a little cleanup, but otherwise this looks ready.  Thanks!

PR #2511, opened 2020-04-20
Update to post_process_elem_matrix_and_vector in RBConstruction

dknez:
We should only call post_process_elem_matrix_and_vector if apply_dof_constraints is true because we want to be able to compute matrices with no dof constraints or dof transformations included in them.


dknez:
This is a one line change that only affects new code in RBConstruction (post_process_elem_matrix_and_vector was added recently) so I'm going to go ahead and merge.

PR #2513, opened 2020-04-21
Introduce std::reference_wrapper in fem_system_ex3

jwpeterson:
This lets us loop over containers of references to subvectors/matrices
instead of hard-coding the names of each reference, so in the end it
results in less code.

std::reference_wrapper is in the C++11 standard and is defined in the
`<functional>` header.

### DIFF
--- a/examples/fem_system/fem_system_ex3/elasticity_system.C
+++ b/examples/fem_system/fem_system_ex3/elasticity_system.C
@@ -15,8 +15,10 @@
 // License along with this library; if not, write to the Free Software
 // Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
+// Example includes
 #include "elasticity_system.h"
 
+// libMesh includes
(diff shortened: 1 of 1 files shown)


examples/fem_system/fem_system_ex3/elasticity_system.C
```diff
@@ -179,20 +184,26 @@ bool ElasticitySystem::element_time_derivative(bool request_jacobian,
   const std::vector<std::vector<Real>> & phi = u_elem_fe->get_phi();
   const std::vector<std::vector<RealGradient>> & grad_phi = u_elem_fe->get_dphi();
 
-  // We set _w_var=_v_var etc in lower dimensions so this is sane:
```
roystgnr, line 182: Let's keep this comment around here and in side_time_derivative?  It's not a priori obvious that using a variable labeled "the index of w" in a 2D system with only "u" and "v" isn't a disaster waiting to happen.

Issue #2510, opened 2020-04-20
Edge3::volume failure for not-quite-linear edges


roystgnr:
Wen Jiang hit this case:

```
point(0) = (x,y,z)=(0.810213, 0, 0),
point(1) = (x,y,z)=(1.01, 0, 0),
point(2) = (x,y,z)=(0.910055, 0, 0)
a = 1.06426e-08
c = 0.00997871
2. * std::sqrt(c) = 0.199787
b = 2.06106e-05
ba = 1936.61
ca = 937619
s1 = 967.307
s2 = 969.307
log_term = nan
Assertion `!libmesh_isnan(log_term) && log_term > 0.' failed.
```

Round-off error in the point values prevents me from reproducing that exactly, but it turns out to be *easy* to reproduce the same effect: just tweak the middle point in a line far enough to avoid the ```TOLERANCE*TOLERANCE``` check but not much farther, and you get NaN in log_term.

roystgnr:
Fixing log_term in the a->0 limit turns out to be simple enough:

```
  const Real root_a = std::sqrt(a);
  
  libmesh_assert (1.-ba+ca>0.);
  
  const Real r1 = std::sqrt(a - b + c); // sqrt(c) if linear
  const Real r2 = std::sqrt(a + b + c); // sqrt(c) if linear
  const Real r2_minus_r1 = 2*b/(r2+r1);
  
  Real log_term = // 1 if linear 
    (a - 0.5*b + r1*root_a) / (a - 0.5*b + r2*root_a);
  libmesh_assert(!libmesh_isnan(log_term) && log_term > 0.);

  return 0.5*(r1 + r2 + ba*b/(r2+r1) +
              root_a*(ca - 0.25*ba*ba)*std::log(log_term)
             );
```

But the astute reader will note that there are still four divisions by zero in that final expression the a->0 limit.  The particularly astute reader will notice that ```ba*b``` in the third term in the final answer isn't just a 0/0 case that itself approaches 0, it's a *path-dependent* limit that doesn't vanish in 2D or 3D.  This implies that either my math is wrong, the original formula was wrong, or we the fourth term must *also* be a path-dependent limit which exactly cancels out the path-dependence in the third term, despite the forms looking nothing alike.

Pinging @jwpeterson

roystgnr:
[edge3.pdf](https://github.com/libMesh/libmesh/files/4506548/edge3.pdf)

John's derivation of the current formula.

roystgnr:
[...]

PR #2514, opened 2020-04-21
Make Edge3::volume() calculation more robust.

jwpeterson:
Refs #2510

### DIFF
--- a/src/geom/edge_edge3.C
+++ b/src/geom/edge_edge3.C
@@ -191,25 +191,40 @@ Real Edge3::volume () const
   const Real c = B.norm_sq();
 
   // Degenerate straight line case
-  if (a < TOLERANCE*TOLERANCE)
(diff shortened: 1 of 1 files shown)


roystgnr:
Or not.  Doesn't help to test for NaNs after the fact if they've already triggered a floating point exception, I suppose.

jwpeterson:
Hah, debug mode to the rescue, but I'm not sure how that test wasn't triggering a floating point error (or an assert) previously? Must have been prevented by the tolerance check... 

I'm guessing the issue here is that:
```
Real log_term = (1. - 0.5*ba + s1) / (-1. - 0.5*ba + s2);
```
is either `inf` or `nan`, since I think both operations raise FPEs. I'll add another check for denominator not exactly equal to 0, falling back on the straight line volume otherwise...

PR #2515, opened 2020-04-21
Don't swap LIBS and CPPFLAGS

roystgnr:
This may be at the root of how a bad CPPFLAGS setting was getting
incorporated into libMesh builds using Conda

### DIFF
--- a/configure
+++ b/configure
@@ -45247,25 +45247,25 @@ $as_echo
 for method in ${METHODS}; do
      case "${method}" in #(
   opt) :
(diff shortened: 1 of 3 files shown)


roystgnr:
Same fix went into TIMPI via https://github.com/libMesh/TIMPI/pull/28

roystgnr:
Shoot, while we're at it I should be fixing the config_summary.m4 bug I caught while debugging this.

lindsayad:
Thanks @roystgnr for fixing this terrible bug!

PR #2519, opened 2020-04-22
Multiple file support in meshtool

roystgnr:
I hacked this together last night to use for visualizing a collection of Bezier patches, but it might be useful to others too.

It won't be *really* generally useful without options to do mesh stitching rather than just mesh union, but I didn't need that so it's not implemented.


roystgnr:
You're correct (bearing in mind that file1.exo itself isn't modified, only the Mesh in memory constructed from it), and you're right that the help message isn't enough.  I'm not sure source code comments are a great place to document the usage of a tool, but I guess until we start making man pages it's the best we've got.

PR #2524, opened 2020-04-23
Respect empty variables in Nemesis_IO::write_nodal_data

### LINKED ISSUE #2523: Respect empty variables in Nemesis_IO::write_nodal_data
I came across this when trying to write Nemesis output for a system that contains only CONSTANT MONOMIAL variables.

I was creating output like this:
```
Nemesis_IO io(mesh);
io.set_output_variables(std::vector<std::string>());
io.write_timestep(filename, es, 1, 1);
io.write_element_data(es);
```

Output variables were set to blank so that in `write_timestep()` it'll write no variables (which is fine, I have no nodal variables), and in `write_element_data()`, it will grab all of the relevant variables. This works for `ExodusII_IO` but not `Nemesis_IO`.

This issue is because in the `Nemesis_IO::write_nodal_data()` call, if no output variables are given, we pass an empty vector of variable names to `EquationSystems::find_variable_numbers()` [here](https://github.com/libMesh/libmesh/blob/master/src/mesh/nemesis_io.C#L1373) which then gives us _all_ of the variable numbers (because the variable name vector is empty).

loganharbour:
Closes #2523

Passing this to CI first to see what happens.

### DIFF
--- a/src/mesh/nemesis_io.C
+++ b/src/mesh/nemesis_io.C
@@ -1370,8 +1370,11 @@ void Nemesis_IO::write_nodal_data (const std::string & base_filename,
 
   this->prepare_to_write_nodal_data(base_filename, output_names);
 
-  std::vector<std::pair<unsigned int, unsigned int>> var_nums =
-    es.find_variable_numbers(output_names);
+  std::vector<std::pair<unsigned int, unsigned int>> var_nums;
+  // If we pass in an empty vector below, it will return all of the
+  // var nums in es, which we don't want.
+  if (!output_names.empty())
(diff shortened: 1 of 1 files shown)


loganharbour:
Do you want a modified test for this?

roystgnr:
This looks like a straightforward fix; we can merge now if you like.

On the other hand, how's the saying go?  "You don't have to test all your code, only the code that's supposed to work."  I'm going to be doing some work on Nemesis too soon, and whether that should sound like a gift or a threat depends on how much test coverage all your preferred features have.  :smile:

loganharbour:
I say go for it.

PR #2526, opened 2020-04-23
Use numerically stable Edge3::volume() algorithm

### LINKED ISSUE #2510: Edge3::volume failure for not-quite-linear edges

roystgnr:
Fixes #2510 for me

Or at least the octave version did.  Hopefully CI will make sure I didn't make any typos in the C++ version.

### DIFF
--- a/tests/geom/volume_test.C
+++ b/tests/geom/volume_test.C
@@ -1,0 +1,56 @@
+// libmesh includes
+#include <libmesh/elem.h>
+#include <libmesh/enum_elem_type.h>
+#include <libmesh/mesh_generation.h>
+#include <libmesh/mesh.h>
+
+// unit test includes
+#include "test_comm.h"
+#include "libmesh_cppunit.h"
+
+using namespace libMesh;
+
+class VolumeTest : public CppUnit::TestCase
+{
+
+public:
+  CPPUNIT_TEST_SUITE( VolumeTest );
+  CPPUNIT_TEST( testEdge3Volume );
+  CPPUNIT_TEST_SUITE_END();
+
+public:
+  void setUp()
+  {
+  }
+
+  void tearDown()
+  {
+  }
+
+  void testEdge3Volume()
+  {
+    Mesh mesh(*TestCommWorld);
+    MeshTools::Generation::build_line (mesh, /*nelem=*/1, 0., 1., EDGE3);
+    CPPUNIT_ASSERT_EQUAL(static_cast<dof_id_type>(3), mesh.n_nodes());
+
+    auto edge3 = mesh.elem_ptr(0);
+
+    // Check unperturbed, straight edge case
+    LIBMESH_ASSERT_FP_EQUAL(1.0, edge3->volume(), TOLERANCE*TOLERANCE);
+
+    // Check middle node perturbed in +x direction case. This should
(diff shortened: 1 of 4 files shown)


roystgnr:
Wow, thanks again for catching that!  We might want to check over the other subclasses that override Elem::volume()... or maybe add some unit tests that compare this->volume() to this->Elem::volume()?

tests/geom/volume_test.C
```diff
@@ -0,0 +1,56 @@
...
+    LIBMESH_ASSERT_FP_EQUAL(1.0, edge3->volume(), TOLERANCE*TOLERANCE);
+
+    // Check middle node perturbed in +x direction case. This should
+    // not change the volume because it's still a straight line
+    // element.
+    Real & middle_node_x = edge3->node_ref(2)(0);
+    middle_node_x += 1.e-3;
+    LIBMESH_ASSERT_FP_EQUAL(1.0, edge3->volume(), TOLERANCE*TOLERANCE);
+
+    // Check middle node perturbed in +x direction case. This should
```
roystgnr, line 48: -x direction

jwpeterson: The comment is fixed in the latest version.

lindsayad:
I hadn't seen @roystgnr use much `auto` before I rebased my AD work on master and saw the new `volume_test` (yea ok three `autos` isn't that many). Do we have a design document or anything about when developers should use `auto`? I started out full-steam on the `auto` train but have gradually swung in the other direction of almost never using `auto` except when the type is basically explicitly stated on the RHS. There are definitely valid arguments in both directions I think. For example if there was a lot more `auto` in libMesh, then creating this `Dual/GeomReal` in libMesh would require a lot less work; but then again the code would be much less self-documenting.

roystgnr:
> I hadn't seen @roystgnr use much auto before I rebased my AD work on master and saw the new volume_test

That one was @jwpeterson; look at the commit log, not just the PR.  :laughing:

> Do we have a design document or anything about when developers should use auto?

Nope.  I'm a bit undecided on it myself, and I'm agreed on the tradeoffs: *if* you're sure what your return type is going to be (if only in a templated sense) then it's more self-documenting to use that type, but even when you think you're sure, it's easy to later discover you're wrong.

I'm not sure the Real->AD/Dual/Geom/SurReal changes fall entirely on the pro- vs anti-auto side, though.  There's something to be said for having the compiler point out to you every place your implementation is changing under the hood, rather than just having it guess what you intended to happen.

PR #2528, opened 2020-04-23
Debugging node pid sync

roystgnr:
Going to try to figure out here what's going on with #2479 and the failing test I can't reproduce locally.


roystgnr:
Well, I can't exactly debug checks that all *passed*.

PR #2533, opened 2020-04-28
Add Quad::quality(TWIST) test

jwpeterson:
This is useful for detecting invalid elements. A Quad either fails this quality test (0) or passes it (1), there is not a continuous range of output values like for the other quality measures.


jwpeterson:
I forgot to update the enum_to_string map for the new enumeration, but rather than pushing that trivial change and redoing all the testing, I'll just merge it by hand.

PR #2534, opened 2020-04-28
Use common m4 files from a submodule

### LINKED ISSUE #2459: Add TIMPI m4 to libMesh autoconf path
This would allow removal of duplicate m4 between TIMPI and libMesh. TIMPI will always be a required submodule of libMesh, so it's safe to do this. This eases maintainability. Also will probably force creation of some new, smaller macros (which we can `AC_REQUIRE`) which is better/safer/easier to maintain in its own right.

roystgnr:
If this works for everybody, this should resolve #2459 - m4 files that we use in multiple subprojects are now in their own submodule, so we can more easily develop them without inadvertently forking them with each change.


.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "contrib/timpi"]
 	path = contrib/timpi
 	url = ../../libMesh/TIMPI
+[submodule "m4/autoconf-submodule"]
+	path = m4/autoconf-submodule
+	url = git@github.com:libMesh/autoconf-submodule
```
jwpeterson, line 9: We need to use the relative path here, for people that don't use ssh keys... the same as the TIMPI url above.

configure.ac
```diff
@@ -125,13 +125,13 @@ AC_SUBST(PERL)
 # Check for whether to enable MPI and PETSc which will influence our
 # choice of compiler
 #--------------------------------------------------------------------
-COMPILER_CONTROL_ARGS
+ACSM_COMPILER_CONTROL_ARGS
```
jwpeterson, line 128: ACSM == Autoconf Submodule? Should we somehow indicate it's the libmesh autoconf submodule? "LAS_"?

jwpeterson:
One other question, does merging this PR then require the `--recursive` flag to `git submodule update`? I already updated the README, but we may also want to make some kind of announcement on libmesh-users (for what it's worth) as this will probably affect people's automated update scripts.

configure.ac
roystgnr: I'd been hoping to keep the name more agnostic (while still being [as unique as practical](https://en.wikipedia.org/wiki/ACSM) to avoid namespace conflicts) but I'm not wedded to it.  I just don't want to accidentally imply that this is *only* useful for libMesh.

.gitmodules
roystgnr: Huh, I agree; I'm not sure why it didn't give me a relative path.  IIRC I didn't do anything with the submodule add command here that I hadn't done elsewhere.

roystgnr:
Yeah, it definitely requires --recursive.  I really wish there was a way to set that flag in the repo rather than making users remember it or update their git config manually.

.gitmodules
roystgnr: And I've got the same non-relative paths elsewhere, except for MetaPhysicL and TIMPI themselves, so I guess I was using a different setting back then?

jwpeterson: Yeah I think you have to go in and edit it by hand. I did that for the TIMPI one in 3746e900a3d4a906a1b656a50814096389d66aee for example.

configure.ac
jwpeterson: OK, hopefully we don't need to add a second autoconf submodule in libmesh (!!), as then "ACSM" might be ambiguous...

jwpeterson:
One other request: can you please ensure that "make dist" continues working before merging this PR? We are completely reliant on that for making release tarballs at the moment. You can see the approach used for TIMPI in Makefile.am, basically we just throw an error if the required submodules aren't init'd when "make dist" is run.

roystgnr:
Will do.

configure.ac
roystgnr: For now if we need more shared autoconf files, I think we'll put them in the autoconf-submodule repo, even if (like the existing ax_* files in there) they're copies of upstream, just so we can make local edits easily enough.

Frankly I'd be thrilled if git some day enabled pulling file-by-file from other repos (still qualified by commit hash), though.  If that happens then I'd love to just pull straight from the autoconf archive where possible; IIRC the maintainers there have been pretty responsive if we ever need anything changed upstream, and in the worst case scenario it'd be easy to temporarily replace a file with a local fork *until* changes were merged upstream.

src/fe/fe_map.C
```diff
@@ -1662,7 +1662,11 @@ Point FEMap::inverse_map (const unsigned int dim,
                           const Point & physical_point,
                           const Real tolerance,
                           const bool secure,
-                          const bool extra_checks)
+                          const bool
+#ifdef DEBUG
+                          extra_checks
```
jwpeterson, line 1667: This is what your `libmesh_dbg_var()` macro is for, right?

jwpeterson:
It looks like all tests failed configure with the error
```
configure: error: You must run "git submodule update --init --recursive" before configuring libmesh
```
but I'm pretty sure that CIVET does run that. Is the new check wrong maybe?

jwpeterson:
I think I know what it is, you need another "test" directive after the `-a`:
```
AS_IF([test -r $top_srcdir/contrib/timpi/README -a test -r $top_srcdir/contrib/timpi/m4/autoconf-submodule/acsm_mpi.m4],
```

jwpeterson:
> you need another "test" directive after the -a

Hmm... I'm now thinking that might only be the case if you use the `&&` syntax for the test? (From m4/scrapte_petsc.m4):
```
[test -r ${PETSC_DIR}/makefile && test -r ${PETSC_DIR}/${PETSC_ARCH}/conf/variables],
```
For some reason I switched all the compound tests from `-a` to the `&&` syntax at some point... can't remember why.

roystgnr:
Hmm... I assumed that we really weren't doing the update recursively (which wouldn't have prevented passing before - it's just m4 files), but maybe there is some weirdness about -a?  It looks like [they are going to deprecate -a](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html) eventually, though, so we might as well change to the more basic syntax here now, whether or not that's the reason you changed it in other areas before.

jwpeterson:
Hah, I did know that at some point, as this commit message proves! 99fb8dcc6d3edc42d20dd0c2f83d988f41ba7cb4

src/fe/fe_map.C
jwpeterson: Ah, the infamous `#ifndef NDEBUG` strikes again. OK, I guess the current macro could remain as-is and we could add e.g. `libmesh_DEBUG_var()` which only triggers `#ifdef DEBUG`.

jwpeterson: I like this spelling since it appears to be shouting at you.

roystgnr:
As per https://github.com/idaholab/civet_recipes/issues/1780 there seems to be a bit of a chicken-and-egg problem with recursive submodule init in Civet.

The only workaround I can think of so far:

1. Check the submodules' AM_MAINTAINER_MODE setting too (which prevents autotools from trying to rebuild itself when it sees newer m4 files, and which thereby doesn't actually *need* the m4 files except for explicit bootstrap invocations) and if it's on (as it is on our submodule hashes) don't error out of configure when we've got the submodules but not the submodules' submodules.
2. To make sure this doesn't let us inadvertently ruin "make dist" output, add a test for recursive submodules' m4 files in Makefile.am instead (as we have with contrib/metaphysicl/README).

roystgnr:
There was a failure in my "make distcheck", but it turned out to be unrelated; I'll fix it in a separate PR once I confirm that there's no other failure that *is* related.

PR #2535, opened 2020-04-29
Add an option to ignore the cached element in point_locator.

pshriwise:
These changes add an option to bypass the cached element that was last found by the point locator. In the Monte Carlo application I work on, [OpenMC](https://openmc.org), I found that relying on the cached element caused bias in our collision estimator. It's fairly common for particles in our code to move very short distances, possibly just across the boundary of an element. In these cases virtually any tolerance we apply to the search is still likely return the previously cached value even if the particle has numerically moved into an other element. The ability to ignore the cached element provides the option of a fresh search each time to avoid this issue.

This additionally contains a fix which ensures the user-specified element containment tolerance is used when checking the cached element on the point locator -- a missed change from #2405.

### DIFF
--- a/include/utils/point_locator_base.h
+++ b/include/utils/point_locator_base.h
@@ -103,8 +103,8 @@ public:
    * the subdomains searched.
    */
   virtual const Elem * operator() (const Point & p,
-                                   bool use_cache = true,
-                                   const std::set<subdomain_id_type> * allowed_subdomains = nullptr) const = 0;
+                                   const std::set<subdomain_id_type> * allowed_subdomains = nullptr,
+                                   bool use_cache = true) const = 0;
 
   /**
    * Locates a set of elements in proximity to the point with global coordinates
--- a/include/utils/point_locator_base.h
+++ b/include/utils/point_locator_base.h
@@ -103,8 +103,7 @@ public:
    * the subdomains searched.
    */
   virtual const Elem * operator() (const Point & p,
-                                   const std::set<subdomain_id_type> * allowed_subdomains = nullptr,
-                                   bool use_cache = true) const = 0;
+                                   const std::set<subdomain_id_type> * allowed_subdomains = nullptr) const = 0;
 
   /**
    * Locates a set of elements in proximity to the point with global coordinates
--- a/include/utils/point_locator_base.h
+++ b/include/utils/point_locator_base.h
@@ -103,6 +103,7 @@ public:
    * the subdomains searched.
    */
   virtual const Elem * operator() (const Point & p,
+                                   bool use_cache = true,
                                    const std::set<subdomain_id_type> * allowed_subdomains = nullptr) const = 0;
 
   /**
--- a/include/utils/point_locator_tree.h
+++ b/include/utils/point_locator_tree.h
@@ -105,6 +105,7 @@ public:
    * operator().
    */
   virtual const Elem * operator() (const Point & p,
+                                   bool use_cache = true,
                                    const std::set<subdomain_id_type> * allowed_subdomains = nullptr) const override;
 
   /**
--- a/include/utils/point_locator_tree.h
+++ b/include/utils/point_locator_tree.h
@@ -105,8 +105,8 @@ public:
    * operator().
    */
   virtual const Elem * operator() (const Point & p,
-                                   bool use_cache = true,
-                                   const std::set<subdomain_id_type> * allowed_subdomains = nullptr) const override;
+                                   const std::set<subdomain_id_type> * allowed_subdomains = nullptr,
+                                   bool use_cache = true) const override;
 
   /**
    * Locates a set of elements in proximity to the point with global coordinates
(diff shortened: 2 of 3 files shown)


include/utils/point_locator_base.h
```diff
@@ -103,6 +103,7 @@ class PointLocatorBase : public ReferenceCountedObject<PointLocatorBase>
    * the subdomains searched.
    */
   virtual const Elem * operator() (const Point & p,
+                                   bool use_cache = true,
```
roystgnr, line 106: This definitely doesn't work here.  You can't add a new parameter in the *middle* of a function signature, even with a default value, without breaking API backwards compatibility for anyone using the old signature.

I like the option (especially being able to easily select it per-evaluation rather than making it a property of the locator), though; try just moving use_cache to be after allowed_subdomains instead of before?

We'll also need some kind of documentation in the comment above.  I think pointing out the TOLERANCE-scale cache bias at element boundaries is a necessity here; if you hadn't explained that in the PR I'd have never imagined why the cache would ever cause a problem or why a reasonable solution would be to avoid caching rather than fix the problem.

roystgnr:
So although my first impulse was to admit "a reasonable solution would be to avoid caching rather than fix the problem", I have to ask - are we sure about that?  Hitting that octree is an O(log(N)) operation, whereas using the candidate_elements overload and choosing the closest candidate would be O(1) (albeit with a much worse constant - it'd still be too expensive to do by default but a reasonable option where necessary).

include/utils/point_locator_base.h
pshriwise: Fair point. That was a little shortsighted. I'll move the new argument to the end of signature.

I'll look into updating the documentation as well.

pshriwise:
Thanks for the prompt review and discussion @roystgnr.

The main reason I didn't dig deeper in to this issue was based on an assumption that other applications are just fine with using the cached element and considering a point as contained within some tolerance in order to, as you stated, avoid the additional O(log(N)) search in the octree. So whether or not it's a problem may be use dependent.

It isn't clear to me how the `candidate_element` overload would help with the complexity/cost issue. Looking at the code, that call appears to require a search of the tree for each query to construct the list of candidate elements. Perhaps what you're getting at though is that I could use that overload to select the nearest candidate without this option for ignoring the cached element?

roystgnr:
Sorry, I wasn't being very clear.  The trick would be to cache a whole set of candidate elements, rather than one element at a time.  The set would then only need reevaluation when the new point had moved so far as to no longer be contained in one of the set's elements.

Except ... no, that wouldn't work.  Picturing a moving point near an element boundary, we move into a new candidate *before* we leave an old candidate, so there's no good test for reevaluation unless we *also* cache a layer of point neighbors, and now this is just sounding immensely unwieldy.

> Perhaps what you're getting at though is that I could use that overload to select the nearest candidate without this option for ignoring the cached element?

... Actually, what I'm fearing right now is that that might be what you should be doing anyways.  If you use the single-element overload, then even ignoring the cache you're still going to hit a code path where we try to save time by breaking out of a loop as soon as we hit a success from contains_point() - but in that loop we loop over an octree node's element list in the same order every time, which means that along boundaries between elements you may see a bias toward whichever element was added to the octree first.  That bias might be acceptable whereas a bias toward an old element was unacceptable, but that's up to you.

pshriwise:
Ah I see, yeah using the candidate elements as a cache would be troublesome for this purpose. But for others it might be useful and is something that could be done external to the point_locator without any changes. In fact, any caching could be done externally if needed, but that's another matter.

> ... Actually, what I'm fearing right now is that that might be what you should be doing anyways. If you use the single-element overload, then even ignoring the cache you're still going to hit a code path where we try to save time by breaking out of a loop as soon as we hit a success from contains_point() - but in that loop we loop over an octree node's element list in the same order every time, which means that along boundaries between elements you may see a bias toward whichever element was added to the octree first. That bias might be acceptable whereas a bias toward an old element was unacceptable, but that's up to you.

Very true. I noticed that as well when looking over the code earlier.

The barrier I come to when thinking about checking a set of elements coming back from the `candidate_elements` overload is how to select the "closest" element? The best way I can think of is to look at the point as mapped to a reference element, but even then that is computed within some tolerance.

pshriwise:
Welp, what I think all that boils down to is that no matter what we're relying on a tolerance for point containment, so there are going to be regions where positive a point containment result for more than one element is valid. In theory, we could potentially find a "best" element based on the point's location in the reference element, but I don't think that's necessary as long as one can control the tolerance governing these ambiguous regions.

In my case, what I found after a considerable amount of digging is that I chose a rather unfortunate test case in which several geometric boundaries, mesh elements faces, and most likely octree boundaries are all coincident. When running very few particles for a unit test, this will almost certainly result in what looks like significant differences in the results due to repeated use of the cached value rather than a fresh search. But, as I should have realized earlier, the tolerances we're applying to the search shouldn't cause any appreciable difference in the results as long as enough particles are run.

In the end, it's probably best to reduce this PR to the fix for the cached value check and leave it at that. Thank you for the discussion @roystgnr!

roystgnr:
Going to squash these commits when merging; we don't need the back-and-forth on an ultimately-unused option in the history.

Thanks for the fix!

pshriwise:
Makes sense. Thanks for the review!

Issue #2537, opened 2020-04-29
Too many boost warnings when building moose stochastic_tools module in clang debug


lynnmunday:
I'm getting 5mb of these types of warnings when I compile the blackbear moose code with clang in debug and it obscures other warnings:

warning: non-standard suffix on floating constant [-Wpedantic]
  275 |   BOOST_DEFINE_MATH_CONSTANT(ln_ln_two, -3.665129205816643270124391582326694694e-01, "-3.66512920581664327012439158232669469454263447837105263053677713670561615319352738549455822856698908358302523045e-01")

These warnings come from the moose stochastic_tools module that includes boost.  The standard way of including boost in moose is to include it through libmesh.  I tried to ignore these warnings with something like this in the stochastic_tools code where it includes boost::

#pragma GCC diagnostic ignored "-Wparentheses"
#pragma clang diagnostic ignored "-Wpedantic"
#include <boost/math/distributions.hpp>
#include "libmesh/restore_warnings.h"

But this does not work.  I think these boost warnings will be ignored if libmesh included boost with "-isystem" instead of "-I".   I was changing the m4/ax_boost_base.m4 file to use -isystem and I'll submit a pull request if it works.  This would fix the moose ticket ref #15158

jwpeterson:
Hmm... it may not be possible to ignore `-Wpedantic` warnings? Not sure about this particular case, but I have seen warnings that could not be ignored in the past. The lowercase `-i` inclusion seems to be a good workaround and relatively standard AFAICT.

lynnmunday:
is "-i" different than "-isystem"?

jwpeterson:
I've never used it so I'm not sure, but I was looking at [this](https://gcc.gnu.org/onlinedocs/cpp/System-Headers.html) which mentions both `-isystem` and `-idirafter`, the latter of which I interpreted could be used as `-icontrib/boost/include`.

lynnmunday:
We're trying -isystem because that is what I saw on stack overflow.

lynnmunday:
This is the first time I've used them

lynnmunday:
I will close this issue.  We can suppress it with the #pragma GCC system_header on the moose side.

lynnmunday:
will close this issue. We can suppress it with the #pragma GCC system_header on the moose side.

PR #2538, opened 2020-04-29
change to boost m4 file to ignore pedantic warnings from boost

lynnmunday:
replace -I with -isystem to ignore warnings coming from boost.  Ben Spencer checked that this does suppress the warnings we were seeing when building grizzly with the moose stochastic_tools which were included boost.  @bwspenc @dschwen @aeslaughter @jwpeterson 

This closes Issue #2537

### DIFF
--- a/configure
+++ b/configure
@@ -33612,7 +33612,11 @@ fi
           if test "x$install_internal_boost" = "xyes"; then :
   libmesh_contrib_INCLUDES="$BOOST_INCLUDE $libmesh_contrib_INCLUDES"
 else
-  libmesh_optional_INCLUDES="$BOOST_CPPFLAGS $libmesh_optional_INCLUDES"
+
+                   BOOST_CPPFLAGS=$(sed -e "s/^-I/-isystem/g" <<< $BOOST_CPPFLAGS)
+                   BOOST_CPPFLAGS=$(sed -e "s/ ^-I/ -isystem/g" <<< $BOOST_CPPFLAGS)
+                   libmesh_optional_INCLUDES="$BOOST_CPPFLAGS $libmesh_optional_INCLUDES"
+
 fi
 
 fi
--- a/m4/boost.m4
+++ b/m4/boost.m4
@@ -69,7 +69,11 @@ AC_DEFUN([CONFIGURE_BOOST],
         [
           AS_IF([test "x$install_internal_boost" = "xyes"],
                 [libmesh_contrib_INCLUDES="$BOOST_INCLUDE $libmesh_contrib_INCLUDES"],
-                [libmesh_optional_INCLUDES="$BOOST_CPPFLAGS $libmesh_optional_INCLUDES"])
+                [
+                   BOOST_CPPFLAGS=$(sed -e "s/^-I/-isystem/g" <<< $BOOST_CPPFLAGS)
+                   BOOST_CPPFLAGS=$(sed -e "s/ ^-I/ -isystem/g" <<< $BOOST_CPPFLAGS)
+                   libmesh_optional_INCLUDES="$BOOST_CPPFLAGS $libmesh_optional_INCLUDES"
+                ])
         ])
 
   AM_CONDITIONAL(LIBMESH_INSTALL_INTERNAL_BOOST, test x$install_internal_boost = xyes)


bwspenc:
This fix does indeed silence the warnings we are getting.  It seems like the right fix to me.

dschwen:
for cosmetic reasons I'd prefer a space after the `-isystem`. That's how the docs show it. But as long as it works it's fine.

dschwen:
Btw. `-isystem` does take precedence over `-I` if both are specified for the same directory!

roystgnr:
Couldn't we do this in boost.m4, where we incorporate BOOST_CPPFLAGS into libmesh_optional_INCLUDES?  Then it would be one new line (a sed command) rather than 14 lines of changes, and more importantly it wouldn't inadvertently get reverted whenever we update ax_boost_base.m4

bwspenc:
@roystgnr Are you proposing something like this:
```
                [libmesh_optional_INCLUDES="-isystem $BOOST_CPPFLAGS $libmesh_optional_INCLUDES"])

**Comment by roystgnr** [comment]
No, I'm afraid that wouldn't work; you'd just get -isystem -I /foo/bar.  Something like preceding that line with a sed command:
```
[libmesh_BOOST_CPPFLAGS=`echo $BOOST_CPPFLAGS | sed -e "s/^-I/-isystem/g" -e "s/ -I/ -isystem/g"
libmesh_optional_INCLUDES="$libmesh_BOOST_CPPFLAGS $libmesh_optional_INCLUDES"])
```

**Comment by bwspenc** [comment]
@roystgnr Somehow I was thinking that it would work if you threw `-isystem` in front of the `-I`, but you're right, we can't do that.
@lynnmunday put together a solution using sed that is similar to what you have, and I just verified that it silences the warnings. It seems a little more confusing to me than his original fix of just replacing the instances of -I with -isystem, but either way works. He's going to update this in just a second.

**Comment by bwspenc** [comment]
@lynnmunday Looks like you have the original two commits in the history.

**Comment by jwpeterson** [comment]
@bwspenc this history's OK with me if it's OK with you.

**Comment by lynnmunday** [comment]
@bwspenc I removed the old commits and pushed a new commit.  @jwpeterson thanks for adding me for testing

**Comment by bwspenc** [comment]
Yes, looks good now.

**Comment by bwspenc** [comment]
@jwpeterson or @roystgnr -- The tests passed.  If you guys are good with this, if you could merge this in sometime soon, that would be great. We've got some build issues with a MOOSE app that is exceeding the limit of the output that we save, so we're anxious to get this in to hopefully solve that.

**Comment by bwspenc** [comment]
Thanks!

**Comment by jwpeterson** [comment]
@bwspenc @lynnmunday I think we have to revert this unfortunately. I'm not exactly sure how it passed all the CI testing, but it breaks the compilation for me and @roystgnr 
```
In file included from ./include/libmesh/libmesh_common.h:36:0,
                 from ./include/libmesh/function_base.h:24,
                 from ./include/libmesh/parsed_function.h:22,
                 from ../src/apps/fparser_parse.C:23:
/usr/include/c++/7/cstdlib:75:15: fatal error: stdlib.h: No such file or directory
 #include_next <stdlib.h>
               ^~~~~~~~~~
compilation terminated.
```

There is a [Stack Overflow post](https://stackoverflow.com/questions/2579576/i-dir-vs-isystem-dir) that seems like it is related, but I don't really know what the resolution of the issue is. I am using GCC 7.4.0 on Ubuntu 18.

**Comment by roystgnr** [comment]
GCC 9.2.1 on Ubuntu 19.10 here, but the same failure message.

**Comment by jwpeterson** [comment]
Reverted in 06dae2334

**Comment by bwspenc** [comment]
That's a bummer.  FYI, @lindsayad (because he was working on updating the libmesh version in MOOSE).

**Comment by jwpeterson** [comment]
Is it possible that the moosebuild CI doesn't even use the internal boost?

**Comment by jwpeterson** [comment]
> Is it possible that the moosebuild CI doesn't even use the internal boost?

No, at least the min-GCC target detects and uses the internal boost... and apparently works just fine.
```
checking for boostlib >= 1.57.0... configure: We could not detect the boost libraries (version 1.57 or higher). If you have a staged boost library (still not installed) please specify $BOOST_ROOT in your environment and do not give a PATH to --with-boost option.  If you are sure you have boost installed, then check your version number looking in <boost/version.hpp>. See http://randspringer.de/boost for more documentation.
<<< External boost installation *not* found... will try to configure for libmesh's internal boost >>>
checking for boostlib >= 1.57.0... yes
<<< Using libmesh-provided boost in ./contrib >>>
```

**Comment by jwpeterson** [comment]
@bwspenc @lynnmunday Did you guys already try putting
```
#pragma GCC system_header
```
before the boost header inclusions in the source file? It's not clear to me that this will work... it seems like that pragma only has an effect if it is itself located in a header, but it could be worth trying.

lynnmunday:
@jwpeterson.  We didn't try that pragma. I saw that in some boost code, before boost used it, they checked for gcc, would it still work for clang?

jwpeterson:
> would it still work for clang?

A similar thing (`#pragma clang system_header`) should work for clang I believe (if this does even work). There's an annoying thing where compilers warn about unknown pragmas... so you end up having to `#ifdef` them to avoid that.

jwpeterson:
FYI it appears we at least aren't alone in dealing with this issue. As this [thread](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92826) mentions, some of the boost headers have taken to pragma'ing themselves as system headers to work around the problem. One possibility is that there is a newer boost version in which distributions.hpp has already been updated in this manner, and thus updating boost would also solve this problem.

bwspenc:
@jwpeterson We have a solution based on `#pragma GCC system_header` that is working, and which we can deal with on the MOOSE side: https://github.com/idaholab/moose/pull/15199/files

jwpeterson:
OK, glad you were able to get it resolved.

PR #2540, opened 2020-05-01
Improve BoundingBox initialization

andrsd:
BoundingBox class assumes that x-, y- and z- coordinates are monotonically
increasing. "Sort" the coordinates before constructing the bounding, so
that users do not have to think about this.

### DIFF
--- a/include/geom/bounding_box.h
+++ b/include/geom/bounding_box.h
@@ -42,13 +42,9 @@ class BoundingBox : public std::pair<Point, Point>
 public:
 
   BoundingBox (const Point & new_min,
-               const Point & new_max) :
-    std::pair<Point, Point>(new_min, new_max)
-  {}
+               const Point & new_max);
 
-  BoundingBox (const std::pair<Point, Point> & bbox) :
-    std::pair<Point, Point> (bbox)
-  {}
+  BoundingBox (const std::pair<Point, Point> & bbox);
 
   /**
    * Default constructor sets invalid bounds.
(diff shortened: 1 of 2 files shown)


andrsd:
This is just a suggestion. If you hate it, we can deal with this on MOOSE level 😃

jwpeterson:
Sounds reasonable to me. I wonder how the `meshgenerators/tiled_mesh_generator.tiled_mesh_generator_test` test failure is related to BoundingBoxes though...

andrsd:
maybe somebody contributed a bad input file and were bit by the same error I was. they assumed the box was correct, but the swapped coordinates yielded an empty bounding box. with the fix, the test probably does what was the original intention - it now produces 1000 nodes, comparing to the original 729.

andrsd:
nope, I was wrong. it wants to create "empty" bbox and it does a series of unions to come up with minimal bbox. I think if I use the c-tor with no params, it should recover the intended behavior...

andrsd:
Ok, that fixed it...

loganharbour:
Nice 👍

roystgnr:
Seems like a bit of a waste of CPU to force everybody to do six redundant ternary operations, just to allow people to write

```BoundingBox bb(p1, p2);```

instead of

```BoundingBox bb; bb.union_with(p1); bb.union_with(p2);```

Would it be sufficient to instead load the constructor up with assertions, so users who don't think about the sorting issue get a clean error message at the error rather than confusing failures later?

andrsd:
Well, there is 8 combinations how to specify a bounding box and this class will accept only 1. All 8 combinations are valid, but the user has to know which one is the correct one. To me that is not right. Users should not need to know about the internal implementation.

Thanks for considering, I will go patch MOOSE instead.

PR #2542, opened 2020-05-01
Fix node unique_id in xda for libMesh 0.9.2

roystgnr:
This was working fine in new libMesh, and in older libMesh, but not in
one of our intermediate file formats, which was breaking a CI test in
https://github.com/idaholab/moose/pull/15182

Refs https://github.com/idaholab/moose/issues/15193

### DIFF
--- a/src/mesh/xdr_io.C
+++ b/src/mesh/xdr_io.C
@@ -1736,37 +1736,38 @@ void XdrIO::read_serialized_nodes (Xdr & io, const dof_id_type n_nodes)
         }
     }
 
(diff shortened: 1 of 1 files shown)


roystgnr:
Technically we don't know this succeeded without a moose modules test in **dbg** mode, but that will fail until https://github.com/idaholab/moose/pull/15200 makes it through, so when this passes the standard CI I'm planning to merge based on it-fixed-the-bug-on-my-system, then we'll double-check the debug modules CI when these both get incorporated into the MOOSE libMesh update.

PR #2543, opened 2020-05-01
Fixed valgrind issue with VecSetDM

fdkong:
We should not compose object twice

### DIFF
--- a/src/solvers/petscdmlibmeshimpl.C
+++ b/src/solvers/petscdmlibmeshimpl.C
@@ -785,7 +785,13 @@ static PetscErrorCode DMCreateGlobalVector_libMesh(DM dm, Vec *x)
   else {
     ierr = VecDuplicate(v,x); CHKERRQ(ierr);
   }
+
+#if PETSC_RELEASE_LESS_THAN(3,13,0)
   ierr = PetscObjectCompose((PetscObject)*x,"DM",(PetscObject)dm); CHKERRQ(ierr);
+#else
+  ierr = VecSetDM(*x, dm);CHKERRQ(ierr);
+#endif
+
   PetscFunctionReturn(0);
 }
 


jwpeterson:
@fdkong You'll have to rebase this on my revert, it has conflicts.

fdkong:
Do not merge it. We need a way to test against both PETSc-3.12 and PETSc-3.13.

I believe this is caused by increasing reference count twice. Then DMDestroy won't delete anything.

fdkong:
@jwpeterson Could you give me some permission? So I can label PR as "do not merge"

lindsayad:
This appears to fix the MOOSE test valgrind failures with PETSc 3.12.5

jwpeterson:
@fdkong that's too much power, I added a label for you :-)

fdkong:
@jwpeterson Do not worry about me, I am fine

fdkong:
OK. I am going to check if it is Valgrind clean using this change against both` PETSc-3.13 and PETSc-3.12`

https://github.com/idaholab/moose/pull/15202

lindsayad:
This also fixes the MOOSE contact module valgrind failure as well, which also used the standard differencing version of FDP. I guess those were the only tests that did. Additionally based on the vi ssls failures that must also use this DMSetUp business.

fdkong:
One single line of code broke the whole world. It is NOT beautiful.

fdkong:
We are testing the whole world here https://github.com/idaholab/moose/pull/15182

fdkong:
It looks like good for me  according to https://civet.inl.gov/job/517103/
and https://civet.inl.gov/job/517145/

This will obviously work for PETSc-3.12 or older since we did not anything for ancient PETSc (`PETSC_RELEASE_LESS_THAN(3,13,0)`).

jwpeterson:
@fdkong you are saying to merge this PR now, is that correct?

fdkong:
> @fdkong you are saying to merge this PR now, is that correct?

Yes, Thanks @jwpeterson.

PR #2544, opened 2020-05-02
Make "NumericVector localize()" scalable if possible

fdkong:
`std::vector<PetscInt> idx(n)` is not scalable since `n` is the global size of vectors

### DIFF
--- a/src/numerics/petsc_vector.C
+++ b/src/numerics/petsc_vector.C
@@ -674,46 +674,51 @@ void PetscVector<T>::localize (NumericVector<T> & v_local_in) const
   PetscVector<T> * v_local = cast_ptr<PetscVector<T> *>(&v_local_in);
 
   libmesh_assert(v_local);
+  // v_local_in should be closed
+  libmesh_assert(v_local->closed());
   libmesh_assert_equal_to (v_local->size(), this->size());
+  // 1) v_local_in is a large vector to hold the whole world
+  // 2) v_local_in is a ghosted vector
+  // 3) v_local_in is a parallel vector
+  // Cases 2) and 3) should be scalable
+  libmesh_assert(this->size()==v_local->local_size() || this->local_size()==v_local->local_size());
 
   PetscErrorCode ierr = 0;
-  const PetscInt n = this->size();
 
-  IS is;
-  VecScatter scatter;
+  if (v_local->type() == SERIAL && this->size() == v_local->local_size())
+  {
+    VecScatter scatter;
 
-  // Create idx, idx[i] = i;
-  std::vector<PetscInt> idx(n);
-  std::iota (idx.begin(), idx.end(), 0);
-
-  // Create the index set & scatter object
-  ierr = ISCreateGeneral(this->comm().get(), n, idx.data(), PETSC_USE_POINTER, &is);
-  LIBMESH_CHKERR(ierr);
-
-  ierr = VecScatterCreate(_vec,          is,
-                          v_local->_vec, is,
-                          &scatter);
-  LIBMESH_CHKERR(ierr);
-
-  // Perform the scatter
-  ierr = VecScatterBegin(scatter, _vec, v_local->_vec,
-                         INSERT_VALUES, SCATTER_FORWARD);
-  LIBMESH_CHKERR(ierr);
-
-  ierr = VecScatterEnd  (scatter, _vec, v_local->_vec,
-                         INSERT_VALUES, SCATTER_FORWARD);
-  LIBMESH_CHKERR(ierr);
-
-  // Clean up
-  ierr = ISDestroy (&is);
-  LIBMESH_CHKERR(ierr);
-
-  ierr = VecScatterDestroy(&scatter);
-  LIBMESH_CHKERR(ierr);
+    ierr = VecScatterCreateToAll(_vec,&scatter,NULL);
+    LIBMESH_CHKERR(ierr);
+    ierr = VecScatterBegin(scatter,_vec,v_local->_vec,INSERT_VALUES,SCATTER_FORWARD);
+    LIBMESH_CHKERR(ierr);
+    ierr = VecScatterEnd(scatter,_vec,v_local->_vec,INSERT_VALUES,SCATTER_FORWARD);
+    LIBMESH_CHKERR(ierr);
+    ierr = VecScatterDestroy(&scatter);
+    LIBMESH_CHKERR(ierr);
+  }
+  // Two vectors have the same size, and we should just do a simple copy.
+  // v_local could be either a parallel or ghosted vector
+  else if (this->local_size() == v_local->local_size())
+  {
+    ierr = VecCopy(_vec,v_local->_vec);
+    LIBMESH_CHKERR(ierr);
+  }
+  else
+  {
+    libmesh_error_msg("Vectors are inconsistent");
+  }
 
   // Make sure ghost dofs are up to date
   if (v_local->type() == GHOSTED)
-    v_local->close();
+  {
+    // We do not call "close" here to save a global reduction
+    ierr = VecGhostUpdateBegin(v_local->_vec,INSERT_VALUES,SCATTER_FORWARD);
+    LIBMESH_CHKERR(ierr);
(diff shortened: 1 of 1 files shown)


fdkong:
@roystgnr  Please let me know if I understand ` localize()` correctly or not?

jwpeterson:
@fdkong looks like this mostly worked, except for several `solution_function` tests in MOOSE? Maybe there is an issue with `SolutionFunction` in MOOSE that could be fixed first, and then this PR would pass?

roystgnr:
I have to say I'm baffled.  SolutionFunction doesn't use localize() directly, and any indirect usage seems to be via the same code paths that I'd have expected to break a hundred other tests if they had problems.

We should only ever be using this function for PARALLEL->SERIAL or GHOSTED->SERIAL copies in libMesh, I think (since for PARALLEL->GHOSTED we always supply a send_list now), in which case I thought it should indeed boil down to VecCopy with non-ancient PETSc.  But it looks like people have using this overload with PARALLEL->GHOSTED anyway?  If so then that'd be the first place I'd be suspicious of only-affects-a-few-tests breakage.

fdkong:
> I have to say I'm baffled. SolutionFunction doesn't use localize() directly, and any indirect usage seems to be via the same code paths that I'd have expected to break a hundred other tests if they had problems.

It is a dead-lock (MPI). I am working on now.

> 
> We should only ever be using this function for PARALLEL->SERIAL or GHOSTED->SERIAL copies in libMesh, I think (since for PARALLEL->GHOSTED we always supply a send_list now), in which case I thought it should indeed boil down to VecCopy with non-ancient PETSc. But it looks like people have using this overload with PARALLEL->GHOSTED anyway? If so then that'd be the first place I'd be suspicious of only-affects-a-few-tests breakage.

PARALLEL->GHOSTED should be able to be accomplished without "send_list". That is, we should be able to do that without using any info from `system`.

It is why I want to make this work.

fdkong:
I was wondering if `libmesh_assert` will be always called regardless of dbg or opt?

roystgnr:
It gets called when NDEBUG isn't defined, which with our standard flags means it will be called in dbg and devel modes, but not in opt mode.

jwpeterson:
`libmesh_assert` is enabled whenever `#ifndef NDEBUG`. In compiler.m4, we set the following flags:
```
  CPPFLAGS_OPT="-DNDEBUG"
  CPPFLAGS_DBG="-DDEBUG"
  CPPFLAGS_DEVEL=""
```
So: both dbg and devel mode don't define `NDEBUG`. Hope that clears things up :laughing:

fdkong:
`localize` is valid when `this` vector is a serial vector?

roystgnr:
Well... in theory nobody should ever call localize() from a serial vector.  In practice, there can be reasons to call localize() from a ghosted vector, which can be serial under the hood on some of our weaker linear algebra shims, so I don't want to actually assert that this isn't serial.

I just looked at your forced push - the only change was that previously you weren't initializing ierr at declaration time, and yet now it's passing the tests it failed before??  Now I'm even *more* baffled - ierr was never actually checked until after it was initialized by the return value of a PETSc call.

fdkong:
> Well... in theory nobody should ever call localize() from a serial vector. In practice, there can be reasons to call localize() from a ghosted vector, which can be serial under the hood on some of our weaker linear algebra shims, so I don't want to actually assert that this isn't serial.

Then I have to rethink my code. We basically have all combinations.

What did you mean by "ghosted vector could be a serial vector"? What is vector->type()? Ghosted or SERIAL?

fdkong:
> I just looked at your forced push - the only change was that previously you weren't initializing ierr at declaration time, and yet now it's passing the tests it failed before?? Now I'm even _more_ baffled - ierr was never actually checked until after it was initialized by the return value of a PETSc call.

What you saw was not what was really happening. I have force push several times. it was my bad.

roystgnr:
> Then I have to rethink my code. We basically have all combinations.

Really?  Under what conditions would you localize a serial vector to something else?  Isn't that just an unnecessary data copy?

> What did you mean by "ghosted vector could be a serial vector"? What is vector->type()? Ghosted or SERIAL?

Hmm... looks like vector->type() will return SERIAL.  But it'll be a vector that someone *tried* to construct as GHOSTED, except that our EpetraVector shim didn't support that.

> I have force push several times.

Ah, that makes much more sense.

> There was a logical error in my libmesh change

What was it?

fdkong:
> > There was a logical error in my libmesh change
> 
> What was it?

There is a parallel vector `with local_size=8 on mpirank=0 and local_size=0 on mpi rank=1`, and I simply thought it was a serial vector. It was my bad

fdkong:
Not sure if I can you one question here: @roystgnr.

What was the motivation in the first place to introduce ghost vectors in libmesh?

fdkong:
> `libmesh_assert` is enabled whenever `#ifndef NDEBUG`. In compiler.m4, we set the following flags:
> 
> ```
>   CPPFLAGS_OPT="-DNDEBUG"
>   CPPFLAGS_DBG="-DDEBUG"
>   CPPFLAGS_DEVEL=""
> ```
> 
> So: both dbg and devel mode don't define `NDEBUG`. Hope that clears things up 😆

Thanks. I got it now :-)

roystgnr:
> There is a parallel vector with local_size=8 on mpirank=0 and local_size=0 on mpirank=1, and I simply thought it was a serial vector. It was my bad

Oh, of course!  You can have local_size==size for a non-serial vector simply because you're on a one-element mesh and you're looking at those values on rank 0.  I should have seen that; I've fixed tons of the same sort of bug (and written too many of the ones I fixed...) myself.

Okay, that makes sense.  This should be ready to merge now, then?

> What was the motivation in the first place to introduce ghost vectors in libmesh?

Just look at the pathetic alternative: originally (as well as *currently* if you're not using PETSc!) current_local_solution (and anything like it) was a serial vector.  We were careful to only copy ghost DoFs to it, and to only use parallel vectors for solves and matvecs and such, so CPU scaling wasn't garbage, but memory scaling still was.  On the 16-processor scale problems we were running a dozen years ago serial vectors still weren't the biggest use of RAM, but obviously that was never going to scale to thousands or hundreds of thousands of processors.

friedmud:
Can I suggest a name change for `localize()`?  I was initially really confused about this whole thing... because I couldn't see why we would want to create a local copy of parallel/ghosted solution vectors... but apparently there are a whole bunch of `localize()` methods I had forgotten about.

How about we rename the ones that are not about copying things locally to: `synchronize()` or something like that....

fdkong:
> Can I suggest a name change for `localize()`? I was initially really confused about this whole thing... because I couldn't see why we would want to create a local copy of parallel/ghosted solution vectors... but apparently there are a whole bunch of `localize()` methods I had forgotten about.
> 
> How about we rename the ones that are not about copying things locally to: `synchronize()` or something like that....

What I really like is:

scalable version (synchronize): parallel vector --> paralle ghosted vector

nonscalable version (serialize): parallel (ghosted) vector  --> a huge serial vector.

roystgnr:
Ironically, that's not far from what we originally had: localize(target) was the nonscalable version, localize(target, send_list) was the scalable version.  I guess adding flexibility muddles clarity.  I like serialize() for the nonscalable version.  synchronize() still sounds ambiguous, but off the top of my head I can't think of anything better.

fdkong:
> Ironically, that's not far from what we originally had: localize(target) was the nonscalable version, localize(target, send_list) was the scalable version.

Haha, you are right. But it would be still nice if we could do a scalable update without using `send_list` .  `send_list` is an unnecessary parameter for my use case.

 >I guess adding flexibility muddles clarity. I like serialize() for the nonscalable version. synchronize() still sounds ambiguous, but off the top of my head I can't think of anything better.

`parallelToGhostUpdate`?

fdkong:
More candidates:

`parallelSynchronize`, `distributedSynchronize`, `scalableSynchronize`

roystgnr:
Well, let me merge this now, regardless of what we decide on for future names.  sync_ghosts and sync_everything?

PR #2546, opened 2020-05-02
Add Tri::quality(SHAPE) metric.

jwpeterson:
This is analogous to the SHAPE metric for the Quad originally added in #1535 (019737632).


jwpeterson:
tri_quality

The results are very similar to the existing `DISTORTION`/`STRETCH` metric that we already have but the formulas are much simpler so this metric should probably be preferred.

PR #2548, opened 2020-05-03
Adding number of threads option to LibMeshInit.

pshriwise:
Adding a parameter to set the number of threads used by libMesh explicitly to allow consistency in the number of threads used when using libMesh with other threaded applications. Specifically, if both the parent application and libMesh are using the `openmp` threading model, the call to `openmp_set_num_threads` in `LibMeshInit` can change the number of threads to an unintended value.

An alternative is to always make sure the `LibMeshInit` object gets created before the number of `openmp` threads is set in the parent application, but this seems to be a more flexible option.

The changes here will prefer the number of threads set by the `--n-threads` option in the command line arguments over the new parameter (`n_threads`) passed into `LibMeshInit`. If the `--n-threads` option is absent and the `n_threads` is it's default value (`-1`) then the number of threads will default to 1.

### DIFF
--- a/include/base/libmesh.h
+++ b/include/base/libmesh.h
@@ -98,9 +98,9 @@ public:
    * parameter to use a user-specified MPI communicator.
    */
   LibMeshInit(int argc, const char * const * argv,
-              MPI_Comm COMM_WORLD_IN=MPI_COMM_WORLD);
+              MPI_Comm COMM_WORLD_IN=MPI_COMM_WORLD, int n_threads=-1);
 #else
-  LibMeshInit(int argc, const char * const * argv);
+  LibMeshInit(int argc, const char * const * argv, int n_threads=-1);
 #endif
 
   /**
--- a/include/base/libmesh.h
+++ b/include/base/libmesh.h
@@ -100,7 +100,8 @@ public:
   LibMeshInit(int argc, const char * const * argv,
               MPI_Comm COMM_WORLD_IN=MPI_COMM_WORLD, int n_threads=-1);
 #else
-  LibMeshInit(int argc, const char * const * argv, int n_threads=-1);
+  LibMeshInit(int argc, const char * const * argv,
+              int COMM_WORLD_IN=0, int n_threads=-1);
 #endif
 
   /**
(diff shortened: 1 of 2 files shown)


include/base/libmesh.h
```diff
@@ -98,9 +98,9 @@ class LibMeshInit
    * parameter to use a user-specified MPI communicator.
    */
   LibMeshInit(int argc, const char * const * argv,
-              MPI_Comm COMM_WORLD_IN=MPI_COMM_WORLD);
+              MPI_Comm COMM_WORLD_IN=MPI_COMM_WORLD, int n_threads=-1);
 #else
-  LibMeshInit(int argc, const char * const * argv);
+  LibMeshInit(int argc, const char * const * argv, int n_threads=-1);
```
roystgnr, line 103: So ... this doesn't work well.  I should have thought about this harder ahead of time, but the combination of this new argument and the previous optional-argument-in-some-configs means that user code can't be made MPI-agnostic anymore if it's also setting n_threads, because the position of the n_threads argument isn't MPI-agnostic.

Maybe what we need to do is simply finally make that optional-argument-in-some-configs into an optional-argument in every config, just using the TIMPI::communicator shim instead of MPI_Comm.  Let me hijack your PR branch and I'll try that.

pshriwise: Ah, yeah I can see how that would be a problem. We had some MPI-dependent code on our side anyway, so it fit in pretty well for us but that definitely won't always be the case.

Thanks for digging into it further!

roystgnr:
My fix isn't as nice as I'd hope (we'll have to add a TIMPI shim for MPI_COMM_WORLD before we can get rid of those ifdefs entirely) but it seems like it's at least an improvement over the baseline, and it makes your new parameter straightforward.  So this should be good to merge once it passes CI.

I've been thinking more about the long-term fix too; maybe instead of creating a whole new LibMeshParameters object, the more flexible way to do things is to just let LibMeshInit take a reference to GetPot?  Then users would still be free to override command line parameters, or ignore command line parameters, or whatever they wanted to, in code, but the change in the library would be a hundred times smaller.  GetPot::set isn't quite as clean as I'd like under the hood but it's sufficient.

PR #2551, opened 2020-05-04
Use push_parallel_packed_range in scatter_constraints

### LINKED ISSUE #2549: Use push_parallel_packed_range somewhere
We really need some test coverage for `push_parallel_packed_range` in CI.

loganharbour:
Closes #2549

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -3828,7 +3828,7 @@ void DofMap::scatter_constraints(MeshBase & mesh)
         ids_offsets = pushed_node_ids_offsets[pid];
       keys_vals.resize(ids_size);
       ids_offsets.resize(ids_size);
-      std::set<const Node *> & nodes = pushed_nodes[pid];
+      std::set<const Node *> nodes;
 
       std::size_t push_i;
       std::set<dof_id_type>::const_iterator it;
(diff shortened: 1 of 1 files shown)


loganharbour:
@roystgnr this isn't really the _ideal_ implementation, but it allows for some testing.

loganharbour:
```
Making all in timpi
make[2]: Entering directory '/opt/civet/build_0/libmesh/build/contrib/timpi'
make[2]: *** No rule to make target 'all'.  Stop.
make[2]: Leaving directory '/opt/civet/build_0/libmesh/build/contrib/timpi'
```

How in the hell did I break this?

roystgnr:
libMesh has "maintainer mode" turned on, so that people can use git checkouts without having autotools installed, but a consequence of that is that we have to check in autotools-output files ourselves.  For submodules we only add those files on branches.  Typically on release branches, which is probably worth it for this bugfix.  Let me see if I can get a change of my own then and then I'll put out a 1.1.1.

roystgnr:
Make that a 1.1.2.  And the trouble is that 1.1.1 was used for the m4 submodule, which means that we need to get #2534 in to handle that change.  I've updated #2534 to use TIMPI 1.1.2, so after that's merged we can just rebase a457ae4 on it.

src/base/dof_map_constraints.C
```diff
@@ -3856,6 +3856,10 @@ void DofMap::scatter_constraints(MeshBase & mesh)
           ids_offsets[push_i].first = *it;
           ids_offsets[push_i].second = _node_constraints[constrained].second;
         }
+
+      auto & pid_nodes = pushed_nodes[pid];
+      pid_nodes.reserve(nodes.size());
```
roystgnr, line 3861: Shoot.  Didn't I mention that the stuff inside ifdef LIBMESH_ENABLE_NODE_CONSTRAINTS wasn't a good idea for test coverage since it doesn't get hit by CI?  I swear I *intended* to mention that.  Anyway, this is broken, starting with std::set::reserve not being a thing.  I'll see if I can fix.

Issue #2550, opened 2020-05-04
Add dual basis


dewenyushu:
The dual basis is often utilized in the dual Lagrange approach. Here, the shape functions for the Lagrange multipliers (LMs) satisfy the biorthogonal condition across the interface. The biorthogonal condition results in a diagonalized local mass matrix, which allows to condense out the LMs and transform the saddle point problem in to a positive definite one. 

- [x] Add `dual_phi`, `dual_dphi`, and `dual_d2phi` in the `fe_base` class
- [x] Add `compute_dual_shape_functions()` to compute the dual shape functions
- [x] Initialize dual shape functions in `fe` for `lowerD` element type

dewenyushu:
Dual basis unit test fails with:

- First order, Lagrange, Hex8

- Second order, Lagrange, QUAD9

- All 3D second order elements

Issue reported by @jwpeterson 
https://github.com/libMesh/libmesh/pull/2552#issuecomment-627581010

jwpeterson:
@dewenyushu I think this issue can be closed if you are happy with it, I don't have any other outstanding issues.

PR #2552, opened 2020-05-04
Add dual shape functions

dewenyushu:
address part of the issue #2550.


dewenyushu:
Hi @lindsayad, please help me review this PR since you are the one who is the most familiar of this :)

lindsayad:
@roystgnr are you open to the concepts here?

include/fe/fe_base.h
```diff
@@ -218,9 +226,18 @@ class FEGenericBase : public FEAbstract
...
+  { libmesh_assert(!calculations_started || calculate_dphi);
+    calculate_dphi = calculate_dual = calculate_dphiref = true; return dual_dphi; }
+
   void request_dphi() const override
   { get_dphi(); }
 
+  void request_dual_dphi() const override
+  { get_dual_dphi(); }
+
+  const DenseMatrix<Real> get_dual_coeff() const;
```
jwpeterson, line 239: Looks like this isn't defined? I gather this is not a complete PR yet...

dewenyushu: Thanks! Nice catch! This should be fixed now.

roystgnr:
> @roystgnr are you open to the concepts here?

The dual shape functions (in a particular norm, generally fixed to be L2) are the shape functions psi_i such that the integral over the element of psi_i * phi_j is 1 for i==j and 0 for i != j, right?

(If your answer is yes then so is mine; likewise if no!)

When I saw #2550 I was prepared to groan at a half dozen ways I could picture modifications to address it going wrong, but so far the WIP has already preemptively addressed two of the things I was expecting to have to complain about, so I'm now looking forward to the rest.

dewenyushu:
> The dual shape functions (in a particular norm, generally fixed to be L2) are the shape functions psi_i such that the integral over the element of psi_i * phi_j is 1 for i==j and 0 for i != j, right?

Exactly!

dewenyushu:
> When I saw #2550 I was prepared to groan at a half dozen ways I could picture modifications to address it going wrong, but so far the WIP has already preemptively addressed two of the things I was expecting to have to complain about, so I'm now looking forward to the rest

Really.. what would be your suggestions going forward?

roystgnr:
Well, my first concerns were with expense for non-users (which you're already handling with calculate_dual) and lack of implementation for most elements (which you're handling via the generic calculation in compute_dual_shape_functions as opposed to hand-coding shape function equations).  Those were the most important things.  I was also initially worried about compatibility with the vector-valued elements, about what sounded like it might be an overly-ambitious initial changeset for a new contributor, and about whether there would be much API bloat added to what's already a fairly bloated class hierarchy, but based on this first commit I was just being paranoid.  The worst thing I see in the code so far is that you don't use index_range in *every* place it would be appropriate, which is a trivial nitpick not a serious problem.

dewenyushu:
> Does this computation make sense for FEs other than Lagrange (non-interpolary, discontinuous, C1, etc.)?

Great question. I think the computation is general enough that shape functions of any FE type can have their corresponding dual shape functions. According to my literature search, however, a dual shape function would be helpful mostly under the circumstance when Lagrange multipliers (LMs) are involved, where a mass matrix of the form phi_i * phi_j can be diagonalized. This allows for condensing of the LMs and thus a positive definite system. I guess the advantage of using dual shape functions for other types of variables may not be prominent in this sense.

dewenyushu:
> I was also initially worried about compatibility with the vector-valued elements,

Thanks for your comment! I actually tried to avoid touching anything associated with the vector-value elements :) And one of the reasons was because I was not sure if the dual shape functions would be beneficial there. I would happy to add the capabilities if needed, after I get an initial proof of concept working. 

> The worst thing I see in the code so far is that you don't use index_range in every place it would be appropriate, which is a trivial nitpick not a serious problem.

Thanks for pointing this out. Will be changed!

jwpeterson:
OK, I was mainly asking from the library implementation perspective rather than the math. modeling one. It might make sense to warn the user and/or not compute dual basis functions (even when they are explicitly pre-requested) in cases where the dual basis functions don't make sense, would lead to numerical issues, etc. for some reason.

dewenyushu:
> OK, I was mainly asking from the library implementation perspective rather than the math. modeling one. It might make sense to warn the user and/or not compute dual basis functions (even when they are explicitly pre-requested) in cases where the dual basis functions don't make sense, would lead to numerical issues, etc. for some reason.

I see. Currently the implementation has been verified for first order Lagrange (which LMs use the most common). I am yet to test higher order or other FE types. I will put that as a warning there until other types being tested and verified. Thanks for pointing this out!

src/fe/fe.C
```diff
@@ -283,6 +283,8 @@ void FE<Dim,T>::reinit(const Elem * elem,
         this->compute_shape_functions (elem,*pts);
       else
         this->compute_shape_functions(elem,this->qrule->get_points());
+      if (this->calculate_dual)
+        this->compute_dual_shape_functions();
```
lindsayad, line 287: We we have something like a `libmesh_do_once`? @dewenyushu you asked about where you could put the warning...off the top of my head, I would put it here, something like:
```c++
{
  static bool warned = false;
  if (T != LAGRANGE && !warned)
  {
    libmesh_warning("dual calculations have only been verified for the LAGRANGE family");
    warned = true;
  }
  this->compute_dual_shape_functions();
}

Comment (jwpeterson) [reply src/fe/fe.C:294]: The `libmesh_warning()` macro is already wrapped in a `libmesh_do_once()` (see include/base/libmesh_common.h) so that outer static bool isn't necessary.

Comment (dewenyushu) [reply src/fe/fe.C:290]: Thanks! I just corrected it.

**Comment by dewenyushu** [comment]
@lindsayad @roystgnr @jwpeterson please let me know if further changes need to be made in order for this to be merged. Thanks!

**Comment by jwpeterson** [comment]
> please let me know if further changes need to be made in order for this to be merged.

This really needs either a unit test or an example. I know there will eventually be coverage in MOOSE but something relatively straightforward so that we have coverage on the libmesh level is important to have. I doesn't have to solve a real problem, but it should verify that the new code is working.

**Comment by jwpeterson** [comment]
@dewenyushu I sent you an invite to join the libmesh Associates team so that your PRs will be tested automatically in the future.

**Comment by dewenyushu** [comment]
> > please let me know if further changes need to be made in order for this to be merged.
> 
> This really needs either a unit test or an example. I know there will eventually be coverage in MOOSE but something relatively straightforward so that we have coverage on the libmesh level is important to have. I doesn't have to solve a real problem, but it should verify that the new code is working.

Thanks for the comment, John! I will work with @lindsayad to add a unit test.

**Comment by jwpeterson** [comment]
The complex numbers build failure also looks to be real, it seems that you may have mixed up some of the numeric types in the `DenseMatrix::lu_solve()` lines.
```
../src/fe/fe_base.C: In member function 'void libMesh::FEGenericBase<T>::compute_dual_shape_functions() [with OutputType = double]':
../src/fe/fe_base.C:765:22: error: cannot bind non-const lvalue reference of type 'libMesh::DenseVector<double>&' to an rvalue of type 'libMesh::DenseVector<double>'
  765 |     A.lu_solve(Dcol, coeffcol);
      |                      ^~~~~~~~
In file included from ../src/fe/fe_base.C:29:
./include/libmesh/dense_vector.h:300:1: note:   after user-defined conversion: 'libMesh::DenseVector<T>::DenseVector(const libMesh::DenseVector<T2>&) [with T2 = std::complex<double>; T = double]'
  300 | DenseVector<T>::DenseVector (const DenseVector<T2> & other_vector) :
      | ^~~~~~~~~~~~~~
In file included from ./include/libmesh/fe_base.h:33,
                 from ./include/libmesh/fe.h:24,
                 from ../src/fe/fe_base.C:21:
./include/libmesh/dense_matrix.h:394:35: note:   initializing argument 2 of 'void libMesh::DenseMatrix<T>::lu_solve(const libMesh::DenseVector<T>&, libMesh::DenseVector<T>&) [with T = double]'
  394 |                  DenseVector<T> & x);
      |                  ~~~~~~~~~~~~~~~~~^
../src/fe/fe_base.C:768:32: error: cannot convert 'std::complex<double>' to 'double' in assignment
  768 |     dual_coeff(row, j)=coeffcol(row);
      |                        ~~~~~~~~^~~~~
      |                                |
      |                                std::complex<double>
In file included from ../src/fe/fe_base.C:29:
./include/libmesh/dense_vector.h: In instantiation of 'libMesh::DenseVector<T>::DenseVector(const libMesh::DenseVector<T2>&) [with T2 = std::complex<double>; T = double]':
../src/fe/fe_base.C:765:30:   required from here
./include/libmesh/dense_vector.h:311:5: error: no matching function for call to 'std::vector<double>::push_back(const value_type&)'
  311 |     _val.push_back(other_vals[i]);
      |     ^~~~
```

**Comment by lindsayad** [comment]
@dewenyushu I've run into the following conundrum while working on a test for you. Just speaking about Lagrange, what do you suggest doing when your quadrature points lie along an element boundary such that some of the shape functions are zero? This leads to A being singular.

**Comment by dewenyushu** [comment]
> @dewenyushu I've run into the following conundrum while working on a test for you. Just speaking about Lagrange, what do you suggest doing when your quadrature points lie along an element boundary such that some of the shape functions are zero? This leads to A being singular.

Just to record our discussion progress here: only do `compute_dual_shape_functions()` when `pts==NULL`, and do `reinit` before sampling the finite element solution

**Comment by lindsayad** [comment]
I think I've verified that we can do dual computations for any FE type without at least raising an error :smile: I'll think I'll add an FP comparison test for an elem/FE combo for which I can do the dual calculations by hand

**Comment by dewenyushu** [comment]
> I think I've verified that we can do dual computations for any FE type without at least raising an error 😄

 Great! That is exciting! We do not have any FE type assumptions in the computation and it should mathematically work for any FE type. Glad you could verify this!

>  I think I'll add an FP comparison test for an elem/FE combo for which I can do the dual calculations by hand

What is the FP comparison exactly?

File: src/fe/fe.C [hunk 3]
```diff
@@ -287,6 +287,13 @@ void FE<Dim,T>::reinit(const Elem * elem,
       {
         if (T != LAGRANGE)
           libmesh_warning("dual calculations have only been verified for the LAGRANGE family");
+
+        // In order for the matrices for the biorthogonality condition to be
+        // non-singular, the dual basis coefficients must be computed when the
+        // primal shape functions are evaluated with a quadrature rule
+        if (!pts)
+          this->compute_dual_shape_coeffs();
```
dewenyushu, line 295: Nice change here!

lindsayad:
> What is the FP comparison exactly?

Floating point. I think I'll just do an `EDGE2` and `FE<1,LAGRANGE>` and check to make sure that the dual coefficients are `2,-1,-1,2` and that the dual shapes evaluated at the quadrature points correspond to  dual1 = 1/2 - 3x/2 and dual2 = 3x/2 + 1/2

src/fe/fe_xyz.C
```diff
@@ -700,6 +700,9 @@ void FEXYZ<Dim>::init_shape_functions(const std::vector<Point> & qp,
 
   }
 #endif // ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+
+  if (this->calculate_dual)
+    this->init_dual_shape_functions(n_approx_shape_functions, n_qp);
```
dewenyushu, line 705: Since you are adding the initializations, do we need to add this to `InfFE<Dim,T_radial,T_map>::init_shape_functions` as well?

lindsayad: I don't know much about infinite elements; if dual has meaning there, then I will add it.

roystgnr: dual has meaning there, but not the *same* meaning; you can't use the L2 norm anymore without some kind of weighting function.  It should be fine to have an attempt to calculate_dual there end up at libmesh_not_implemented() for now.

dewenyushu:
> > What is the FP comparison exactly?
> 
> Floating point. I think I'll just do an `EDGE2` and `FE<1,LAGRANGE>` and check to make sure that the dual coefficients are `2,-1,-1,2` and that the dual shapes evaluated at the quadrature points correspond to dual1 = 1/2 - 3x/2 and dual2 = 3x/2 + 1/2

Okay. I did check in an MOOSE example exactly the same case you are checking here haha. I would be surprised if the numbers do not match here

lindsayad:
That failure isn't my fault!

roystgnr:
No, that failure is killing everything now:

lindsayad:
> No, that failure is killing everything now:
> 
> https://civet.inl.gov/recipe_events/27297/
> 
> What the heck happened?

The thing is that the seemingly troublesome idaholab/moose#15327 (or could it have been in idaholab/moose#15223?) wouldn't have made it into MOOSE master unless it passed all our CIVET testing with MOOSE's current libmesh submodule hash. So there must be something different from our libMesh hash and libMesh master that incurs the failure

lindsayad:
I'll look into this failure now

lindsayad:
Currently unable to reproduce this with the same MOOSE hash, the libMesh hash resulting from merge of this PR, and PETSc 3.12.5. MOOSE testing these days is only using PETSc 3.12.5 so I'm wondering if that's the difference here...?

fdkong:
> Currently unable to reproduce this with the same MOOSE hash, the libMesh hash resulting from merge of this PR, and PETSc 3.12.5. MOOSE testing these days is only using PETSc 3.12.5 so I'm wondering if that's the difference here...? (This is using PETSc 3.11.4) @fdkong on the MOOSE side is PETSc 3.11.4 still tested anywhere?

We test both PETSc-3.11.4 (petsc alt) and PETSc-3.12.5 (regular)

lindsayad:
K. But I'm seeing that we don't do any distributed mesh testing with PETSc alt, so maybe that's it. @miljm @roystgnr I should make this MOOSE test pass regardless of the PETSc version, but this is bringing me to the question of the future of libmesh testing...are we going to start using conda?

lindsayad:
Alright successfully reproduced with PETSc 3.11.4!

lindsayad:
@fdkong any speculation about this silliness? There are 200 dofs, 50 local dofs:
```
ime Step 1, time = 0.25, dt = 0.25
 0 Nonlinear |R| = 2.757015e-01
      0 Linear |R| = 2.757015e-01
  Linear solve did not converge due to DIVERGED_PC_FAILED iterations 0
                 PC_FAILED due to FACTOR_OUTMEMORY 
Nonlinear solve did not converge due to DIVERGED_LINE_SEARCH iterations 0
```
MUMPS seems like it's terrible

lindsayad:
Runs fine with `-pc_factor_mat_solver_type superlu_dist`. I don't know why this was using `-pc_type lu` anyways; going to remove that option. Runs fine with default `bjacobi`

fdkong:
> @fdkong any speculation about this silliness? There are 200 dofs, 50 local dofs:
> 
> ```
> ime Step 1, time = 0.25, dt = 0.25
>  0 Nonlinear |R| = 2.757015e-01
>       0 Linear |R| = 2.757015e-01
>   Linear solve did not converge due to DIVERGED_PC_FAILED iterations 0
>                  PC_FAILED due to FACTOR_OUTMEMORY 
> Nonlinear solve did not converge due to DIVERGED_LINE_SEARCH iterations 0
> ```
> 
> MUMPS seems like it's terrible

OMG. It is time for me to write an `HLU`?

lindsayad:
https://github.com/idaholab/moose/pull/15254

dewenyushu:
> Runs fine with `-pc_factor_mat_solver_type superlu_dist`. I don't know why this was using `-pc_type lu` anyways; going to remove that option. Runs fine with default `bjacobi`

So `-pc_type lu` will be deprecated sometime soon? @fdkong

fdkong:
> > Runs fine with `-pc_factor_mat_solver_type superlu_dist`. I don't know why this was using `-pc_type lu` anyways; going to remove that option. Runs fine with default `bjacobi`
> 
> So `-pc_type lu` will be deprecated sometime soon? @fdkong

`-pc_type lu` will live forever. Do not worry! The issue here is MUMPS is unstable. In moose, most of the time we use superlu_dist

roystgnr:
Ouch, you learn something new every day.  I became familiar with MUMPS first, so it's always been my go-to for parallel LU when I'm trying to diagnose convergence failures in iterative solvers on problems that I couldn't scale down to serial.  I guess it's just dumb luck that I've never hit an invertible matrix it couldn't handle, because I wouldn't have even thought to try SuperLU instead; I just assumed the effective differences between them were performance related.

pbauman:
FWIW, I've traditionally found mumps to be a bit faster than superlu_dist, but mumps can have a harder time with problems when the matrices get more ill-conditioned, but that's not uniformly true. I've found examples where mumps failed and superlu_dist succeeded and vice-versa. Parallel, sparse-direct solvers are hard.

milljm:
> K. But I'm seeing that we don't do any distributed mesh testing with PETSc alt, so maybe that's it. @milljm @roystgnr I should make this MOOSE test pass regardless of the PETSc version, but this is bringing me to the question of the future of libmesh testing...are we going to start using conda?

I can add a new target that uses moose-petsc easily enough.

src/fe/fe.C
lindsayad: That seems bad. The points being passed for reinitializing the lower-D element won't correspond to a quadrature rule for the lower-D element; they will correspond to a quadrature rule for the mortar element. I thought you told me that reinit needed to be called for the "full" element in order to compute the correct coefficients? We're going to end up back in the same boat where a user could very easily construct a singular matrix

dewenyushu: For the mortar approach where we integrate on the projected points, the non-singularity of the local matrices depends on the correctness of the projected integration points.. Please see https://github.com/idaholab/moose/pull/15216#discussion_r423129248

lindsayad: So instead of `if (!pts)`, the check should be `if (you're giving me a set of points I approve of)` :smiley:

lindsayad: Ugh, so I guess I'm going to have to remove this check and then just not use the preexisting `fe_test` for coverage :frowning_face:

dewenyushu: > So instead of `if (!pts)`, the check should be `if (you're giving me a set of points I approve of)` 😃

Haha that is definitely true! I am just not sure what is the mathematical equivalence of 'you're giving me a set of points I approve of' .. maybe putting a non-singularity check in the `compute_dual_coeff` would be the easiest fix

jwpeterson: Should we check (maybe only in debug mode or if the user requests it) whether the matrix that gets constructed is singular? Unfortunately DenseMatrix::det() overwrites the matrix with an LU decomposition to compute the determinant... so this check would not be cheap or fast.

lindsayad: What do we do if the check fails? Warn the user and continue? Maybe the hard failure out of the LU factorization is the most appropriate thing

dewenyushu: Thanks John! We are actually doing `lu_solve` later in the code. So it is already doing a LU decomposition, can we make use of that? 

Or instead of computing the determinate, is there a warning in `lu_solve` so that we throw an error when the solve fails?

dewenyushu: > What do we do if the check fails? Warn the user and continue? Maybe the hard failure out of the LU factorization is the most appropriate thing

I agree. The code should throw an error and quit when the LU solve fails.

jwpeterson: OK, sorry I kind of forgot that you already do an LU solve, but yeah both the hand-coded and Lapack-based `DenseMatrix::_lu_decompose`, routines will throw an exception when the matrix is singular, so you could then potentially recover from that in some way at the MOOSE level.

jwpeterson: It may also be worth computing the `det()` after the `lu_solve()` is done since it's very cheap relative to the cost of `lu_solve()` (just the product of diagonal entries). This may alert you to cases which are nearly singular but which don't otherwise throw an exception... not sure how relevant that is in practice but I imagine it could affect the accuracy of the method if the `det()` is 1.e-12 or something...

lindsayad:
Beautiful code ready for review

lindsayad:
@jwpeterson does this look good to you?

jwpeterson:
@dewenyushu I turned the `det()` check into a normal print statement and ran the unit tests just to see what would happen. Note that you can run e.g.
```
./unit_tests-opt --re 'FETest_FIRST_LAGRANGE_.*::testDualDoesntScreamAndDie'
```
to test just the first-order Lagrange elements, for example. These all seemed to work fine except for the Hex8 case, for which I got:
```
The determinant of the coefficient matrix for dual calculations is 1.12157e-13, |det(A)| < 1e-12. Are you using a proper set of integration points for your element?.
```
does that seem correct to you? Do you expect the dual basis function computation to work for 3D elements?

For 2D SECOND-order LAGRANGE elements, I get a nearly singular matrix for QUAD9:
```
The determinant of the coefficient matrix for dual calculations is 1.65195e-13, |det(A)| < 1e-12. Are you using a proper set of integration points for your element?.
```
For 3D SECOND-order elements, all are singular:
```
TET10
The determinant of the coefficient matrix for dual calculations is 2.78977e-30, |det(A)| < 1e-12. Are you using a proper set of integration points for your element?.
HEX20
The determinant of the coefficient matrix for dual calculations is 3.06069e-39, |det(A)| < 1e-12. Are you using a proper set of integration points for your element?.
HEX27
The determinant of the coefficient matrix for dual calculations is 3.02683e-58, |det(A)| < 1e-12. Are you using a proper set of integration points for your element?.
PRISM15
The determinant of the coefficient matrix for dual calculations is 7.02682e-32, |det(A)| < 1e-12. Are you using a proper set of integration points for your element?.
PRISM18
The determinant of the coefficient matrix for dual calculations is 1.73306e-40, |det(A)| < 1e-12. Are you using a proper set of integration points for your element?.
PYRAMID13
The determinant of the coefficient matrix for dual calculations is 4.44295e-33, |det(A)| < 1e-12. Are you using a proper set of integration points for your element?.
PYRAMID14
The determinant of the coefficient matrix for dual calculations is 2.50749e-36, |det(A)| < 1e-12. Are you using a proper set of integration points for your element?.
```
Also note that there is an issue with the `det()` check currently, we need to take the absolute value. I will open a PR for this shortly...

dewenyushu:
@jwpeterson thanks for checking different element types, really appreciate it! I have not verified the implementation of different element types other than first order Lagrange. 

However, I do not expect the calculation would fail for different element types. The concept for the dual shape functions is pretty straightforward and does not assume special element types. The only thing I would think of is we are probably not passing the proper integration points in the calculation...

Other than this, I did observe an issue of an MOOSE example that does not converge while using 2nd order dual Lagrange element. I assumed the error came from my implementation on the MOOSE side, and I must be wrong about it. 

Anyways, I will take it from here and report back what I will find.

jwpeterson:
I think the simplest explanation is that there is some issue with our `DenseMatrix::det()` implementation... I did not realize previously that you are just building a mass matrix, so unless there is something extremely wrong with the `phi` values at the time you are using them, I really don't see how any element can have a zero determinant for its mass matrix...

dewenyushu:
> I think the simplest explanation is that there is some issue with our `DenseMatrix::det()` implementation... I did not realize previously that you are just building a mass matrix, so unless there is something extremely wrong with the `phi` values at the time you are using them, I really don't see how any element can have a zero determinant for its mass matrix...

Yeah. I agree. So this suggests that `phi` may not be initialized when I access it during the calculation, or I am probably integrating using the wrong `qrule`/ `JxW`, right?

lindsayad:
> I think the simplest explanation is that there is some issue with our `DenseMatrix::det()` implementation... I did not realize previously that you are just building a mass matrix, so unless there is something extremely wrong with the `phi` values at the time you are using them, I really don't see how any element can have a zero determinant for its mass matrix...

Lol. I can bust out the debugger for this @jwpeterson unless you're already on it. Should be pretty fast to track down the issue looking at Lagrange/Hex8 for example

jwpeterson:
> I can bust out the debugger for this @jwpeterson unless you're already on it.

Sorry, just saw this... I am looking into whether there is something wrong with the `det()` computation now.

jwpeterson:
I printed the matrix A before the lu_solve for linear HEX8 and computed the determinant with numpy:
```
import numpy as np
a = np.array([[3.70370370e-02,  1.85185185e-02,  9.25925926e-03,  1.85185185e-02,  1.85185185e-02,  9.25925926e-03,  4.62962963e-03,  9.25925926e-03], 
 [1.85185185e-02,  3.70370370e-02,  1.85185185e-02,  9.25925926e-03,  9.25925926e-03,  1.85185185e-02,  9.25925926e-03,  4.62962963e-03], 
 [9.25925926e-03,  1.85185185e-02,  3.70370370e-02,  1.85185185e-02,  4.62962963e-03,  9.25925926e-03,  1.85185185e-02,  9.25925926e-03], 
 [1.85185185e-02,  9.25925926e-03,  1.85185185e-02,  3.70370370e-02,  9.25925926e-03,  4.62962963e-03,  9.25925926e-03,  1.85185185e-02], 
 [1.85185185e-02,  9.25925926e-03,  4.62962963e-03,  9.25925926e-03,  3.70370370e-02,  1.85185185e-02,  9.25925926e-03,  1.85185185e-02], 
 [9.25925926e-03,  1.85185185e-02,  9.25925926e-03,  4.62962963e-03,  1.85185185e-02,  3.70370370e-02,  1.85185185e-02,  9.25925926e-03], 
 [4.62962963e-03,  9.25925926e-03,  1.85185185e-02,  9.25925926e-03,  9.25925926e-03,  1.85185185e-02,  3.70370370e-02,  1.85185185e-02], 
 [9.25925926e-03,  4.62962963e-03,  9.25925926e-03,  1.85185185e-02,  1.85185185e-02,  9.25925926e-03,  1.85185185e-02,  3.70370370e-02]]) 
np.linalg.det(a)
```
The answer is 1.1215665428215297e-13. So this basically agrees with our internal `det()` calculation. The main problem that i see is that this matrix is not even symmetric...

jwpeterson:
I also printed the LU-factorized matrix A and computed the det as in the code by taking the product of diagonal elements
```
LU = np.array(
 [[3.70370370e-02,  5.00000000e-01,  2.50000000e-01,  5.00000000e-01,  5.00000000e-01,  2.50000000e-01,  1.25000000e-01,  2.50000000e-01], 
 [1.85185185e-02,  2.77777778e-02,  5.00000000e-01,  3.70074342e-17,  9.94574793e-17,  5.00000000e-01,  2.50000000e-01,  4.97287396e-17], 
 [9.25925926e-03,  1.38888889e-02,  2.77777778e-02,  5.00000000e-01,  1.18605186e-32, -3.46944695e-18,  5.00000000e-01,  2.50000000e-01], 
 [1.85185185e-02,  1.34922937e-18,  1.38888889e-02,  2.08333333e-02,  4.93432455e-17, -5.39691748e-18,  8.32667268e-17,  5.00000000e-01], 
 [1.85185185e-02,  3.08395285e-18,  3.42387546e-34,  1.02798428e-18,  2.77777778e-02,  5.00000000e-01,  2.50000000e-01,  5.00000000e-01], 
 [9.25925926e-03,  1.38888889e-02, -1.92747053e-19,  9.63735264e-20,  1.38888889e-02,  2.08333333e-02,  5.00000000e-01,  7.40148683e-17], 
 [4.62962963e-03,  6.94444444e-03,  1.38888889e-02,  2.02384406e-18,  6.94444444e-03,  1.04166667e-02,  2.08333333e-02,  5.00000000e-01], 
 [9.25925926e-03,  1.54197642e-18,  6.94444444e-03,  1.04166667e-02,  1.38888889e-02,  1.44560290e-18,  1.04166667e-02,  1.56250000e-02]])
np.prod(np.diagonal(LU))
```
This gives the same result, so I'm fairly confident in the `DenseMatrix` side of things. Next I will compare to an element mass matrix computed in a "normal" assembly loop...

src/fe/fe.C
roystgnr: Oh, good point.  Using LU is currently required for that test+warning to work, because our DenseMatrix::det() implementation depends on LU, but it wouldn't be hard to just upgrade det().

jwpeterson:
For the linear HEX8 reference element I get the following mass matrix:
```
 2.96296296e-01  1.48148148e-01  7.40740741e-02  1.48148148e-01  1.48148148e-01  7.40740741e-02  3.70370370e-02  7.40740741e-02 
 1.48148148e-01  2.96296296e-01  1.48148148e-01  7.40740741e-02  7.40740741e-02  1.48148148e-01  7.40740741e-02  3.70370370e-02 
 7.40740741e-02  1.48148148e-01  2.96296296e-01  1.48148148e-01  3.70370370e-02  7.40740741e-02  1.48148148e-01  7.40740741e-02 
 1.48148148e-01  7.40740741e-02  1.48148148e-01  2.96296296e-01  7.40740741e-02  3.70370370e-02  7.40740741e-02  1.48148148e-01 
 1.48148148e-01  7.40740741e-02  3.70370370e-02  7.40740741e-02  2.96296296e-01  1.48148148e-01  7.40740741e-02  1.48148148e-01 
 7.40740741e-02  1.48148148e-01  7.40740741e-02  3.70370370e-02  1.48148148e-01  2.96296296e-01  1.48148148e-01  7.40740741e-02 
 3.70370370e-02  7.40740741e-02  1.48148148e-01  7.40740741e-02  7.40740741e-02  1.48148148e-01  2.96296296e-01  1.48148148e-01 
 7.40740741e-02  3.70370370e-02  7.40740741e-02  1.48148148e-01  1.48148148e-01  7.40740741e-02  1.48148148e-01  2.96296296e-01 
```
which has `det() = 1.88167642e-06`. So, I'm not sure exactly why I was expecting an O(1) determinant for some reason? The eigenvalues are:
```
(1.00000000e+00, 0.00000000e+00)
(3.70370370e-02, 0.00000000e+00)
(3.33333333e-01, 0.00000000e+00)
(3.33333333e-01, 0.00000000e+00)
(1.11111111e-01, 3.80058872e-17)
(1.11111111e-01, -3.80058872e-17)
(1.11111111e-01, 0.00000000e+00)
(3.33333333e-01, 0.00000000e+00)
```
(condition number = 27) so otherwise all seems normal to me.

jwpeterson:
Also if I scale the reference HEX8 by 1/2 in all directions so that it matches the unit test, I get `det()=1.12156655e-13`, an exact match. In conclusion, I don't think there's anything wrong with the dual basis function coefficients, but the `det()` test I proposed was wildly wrong so we should remove that.

dewenyushu:
Thanks @jwpeterson for working on this! I tried to follow but I was not quite familiar with libmesh, while you are moving so quickly! 

Just to make sure I understand this: so there does not seems to be a problem in the dual basis function coefficients, but the `det()` computation was problematic? Does it pass the unit testing for higher order elements if the `det()` check gets removed?

jwpeterson:
> there does not seems to be a problem in the dual basis function coefficients

Yes, they seem to be working fine as far as I can tell. The only thing that was broken was my understanding for how small mass matrix determinants are in general, but I now realize that it doesn't affect the accuracy of the L2-projection, and is very sensitive to changes in element size, especially for higher-order elements, since det(c*A) = c^n * det(A), where A is n x n. I think I will open a PR to change the check to ensure that the `det()` is simply non-negative, since we always expect that for SPD matrices. 

Also, the suggestion by @variscarey to use cholesky_solve() for this is a good one so if you don't mind I'd like to tack that on to #2550, but I can take care of that myself at some point.
 
> Does it pass the unit testing for higher order elements if the det() check gets removed?

Yeah, it was already passing before as well, I was just seeing a warning that prompted the further investigation.

dewenyushu:
Great. Thanks for the explanation.
> Also, the suggestion by @variscarey to use cholesky_solve() for this is a good one so if you don't mind I'd like to tack that on to #2550, but I can take care of that myself at some point.

Sure, let's use cholesky as it seems to be cheaper in this case.

jwpeterson:
> Does it pass the unit testing for higher order elements if the det() check gets removed?

Regarding this, there are still actually a couple of failures when I merely assert that `A.det() > 0`. I'm looking into those now, but they are all specialized FEs on a particular element so we may just end up skipping them in this test. 
```
1) test: FETest_FOURTH_BERNSTEIN_HEX27::testDualDoesntScreamAndDie (E)
uncaught exception of type std::exception (or derived).
- Mass matrix had non-positive determinant = 0


2) test: FETest_THIRD_HERMITE_HEX27::testDualDoesntScreamAndDie (E)
uncaught exception of type std::exception (or derived).
- Mass matrix had non-positive determinant = 0


3) test: FETest_FOURTH_HIERARCHIC_HEX27::testDualDoesntScreamAndDie (E)
uncaught exception of type std::exception (or derived).
- Mass matrix had non-positive determinant = 0


4) test: FETest_FOURTH_L2_HIERARCHIC_HEX27::testDualDoesntScreamAndDie (E)
uncaught exception of type std::exception (or derived).
- Mass matrix had non-positive determinant = 0
```

jwpeterson:
After some more digging, @roystgnr found that while the mass matrices are actually being computed OK for these elements, their determinants are so small that it underflows the double precision number to actually try and compute them, and therefore det() returns 0.0. A better test would be to check the minimum diagonal entry of A (after the lu_solve, this is the smallest eigenvalue of A) is strictly > 0, which it seems that we can do with sufficient accuracy. I will open a PR with that change hopefully shortly.

PR #2556, opened 2020-05-05
Try appending to a VariableGroup in add_variables

roystgnr:
@fdkong ran into a case where adding two variable arrays in MOOSE gave
him false from has_blocked_representation(), apparently because MOOSE
arrays use the libMesh add_variables() API and that API doesn't do the
same variable group optimization that add_variable() does.  Let's try
fixing that.

Marking this do-not-merge because it's completely untested; uploading it anyway so Fande and CI can try it out ASAP.


roystgnr:
This is fixing the test case @fdkong tried, and it's not breaking anything in CI, so merging.  We should still get some better test coverage for this case in libMesh itself rather than relying on MOOSE indefinitely.

PR #2559, opened 2020-05-07
Fix an overzealous assertion in unpack_indexing

roystgnr:
Doing distributed mesh distribution with extra integers in debug mode
would previously fail at i=1 here when the first assertion failed to
take into account our overloading of the sign bit at i=0.

This fixes https://github.com/idaholab/moose/pull/15241 for me.

Not sure what we should do to get better test coverage on this.  I'm considering just adding some extra integers to the mesh in one of our AMR examples, which would at least have triggered this bug in my own tests and in some of our non-default CI recipes.


roystgnr:
The only failing test appears to be from that MOOSE compatibility regression that @lindsayad is still hunting down, so I'm merging.

PR #2560, opened 2020-05-07
Update to MetaPhysicL 1.0.2

roystgnr:
This changes the submodule's submodule path to be a relative path,
fixing submodule init for users who aren't using ssh keys


jwpeterson:
I'm planning to merge this if the only failure is that `fv_2D_constant_scalar_advection` test...

roystgnr:
Yeah, that's unrelated.  If that's the only failure then we should be good to go when Test debug is done.

PR #2561, opened 2020-05-07
Add hash specialization for FEType

lindsayad:
I was looking into changing all our `std::maps` in MOOSE's Assembly class to `std::unordered_maps` because I was seeing `std::map::operator[]` show up in profiling, so I needed to make `FEType` hashable. Unfortunately it appears that it's actually a little slower in the simple diffusion case I was profiling with `_Hashtable::_M_find_before_node` coming from `std::unordered_map::find` being the main culprit. So probably not gonna switch in that case. However, maybe this could be useful someday...

### DIFF
--- a/include/Makefile.in
+++ b/include/Makefile.in
@@ -1019,6 +1019,7 @@ include_HEADERS = \
         utils/compare_types.h \
         utils/enum_to_string.h \
         utils/error_vector.h \
+        utils/hashing.h \
         utils/hashword.h \
         utils/ignore_warnings.h \
         utils/int_range.h \
--- a/include/fe/fe_type.h
+++ b/include/fe/fe_type.h
@@ -26,15 +26,18 @@
 #include "libmesh/enum_order.h"
 #include "libmesh/enum_fe_family.h" // LAGRANGE
 #include "libmesh/enum_inf_map_type.h" // CARTESIAN
+#include "libmesh/hashing.h"
 
 // C++ includes
 #include <memory>
+#include <functional>
 
 namespace libMesh
 {
 
 // Forward declarations
 class QBase;
+class FEType;
 
 /**
  * This provides a shim class that wraps the Order enum.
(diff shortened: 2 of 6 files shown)


jwpeterson:
I guess you based your example on [this](https://en.cppreference.com/w/cpp/utility/hash) but I find it a bit strange that their advice is to create two `std::hash` objects and use each one once rather than creating 1 object and calling `operator()` twice. Maybe their approach is faster somehow :man_shrugging:

jwpeterson:
Also, maybe we should combine the values [the way they do it in boost](https://stackoverflow.com/questions/35985960/c-why-is-boosthash-combine-the-best-way-to-combine-hash-values)?

lindsayad:
> Also, maybe we should combine the values [the way they do it in boost](https://stackoverflow.com/questions/35985960/c-why-is-boosthash-combine-the-best-way-to-combine-hash-values)?

It looks like the boost way will also create two hash objects. I'll implement the boost way although the guy who wrote the [second most accepted answer](https://stackoverflow.com/a/50978188/4493669) (two years later) doesn't like it particularly much.

jwpeterson:
> I'll implement the boost way although the guy who wrote the [second most accepted answer](https://stackoverflow.com/a/50978188/4493669) (two years later) doesn't like it particularly much.

Yeah, I saw that after my original post. Seems like he researched it... I'm fine with whatever, including no changes.

lindsayad:
Maybe I'll just leave as is then for consistency with the c++ reference

roystgnr:
I was going to suggest that we use whatever's in our hashword.h, but then I actually *looked* at uint32_t hashword2, and I'm guessing that 0xdeadbeef was not a constant chosen to optimize maxmin of relative entropy.

jwpeterson:
> I'm guessing that 0xdeadbeef was not a constant chosen to optimize maxmin of relative entropy.

Maybe not but I'm pretty sure `hashword2` (hard-coded for two inputs) is a legit hash and would also be a good choice to use here.

lindsayad:
This is ready to review/merge

PR #2563, opened 2020-05-11
Support shell preconditioning matrix in eigen system

fdkong:
If users are going to add a customized preconditioner, we should allow them to do whatever they want. Sometimes we do not like to store the matrix explicitly.

For example, @YaqiWang wants to construct a transport sweeper where the preconditioning matrix is not formed.

### DIFF
--- a/include/solvers/eigen_solver.h
+++ b/include/solvers/eigen_solver.h
@@ -198,6 +198,19 @@ public:
                                                                  const double tol,
                                                                  const unsigned int m_its) = 0;
 
+   /**
+    * Solves the standard eigenproblem involving the ShellMatrix \p matrix_A and
+    * the shell preconditioning matrix \p precond.
+    *
+    * \returns The number of converged eigenpairs and the number of
+    * iterations.
+    */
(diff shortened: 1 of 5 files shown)


fdkong:
I did something wrong:?

```
ERROR  : Failed to obtain user identity information
ABORT  : Retval = 255
/tmp/: rm -rf /tmp/Ubuntu-16
```

roystgnr:
That's an MPI initialization failure on the build system, definitely nothing to do with this PR.

roystgnr:
I only kicked one of the failing CI targets, because if this isn't specific to a few random nodes having problems then I don't want to just start spamming new jobs in, but if Test debug makes it past the unit tests' startup this time then we should restart the other two too.  I hope that's what happens; the pattern of which CI failed vs passed looks worryingly configuration-specific, not random.

fdkong:
@roystgnr  Any more comments?

roystgnr:
Little annoyed that that FV test fix hasn't caught up yet, but let's not let that hold this up.

Issue #2564, opened 2020-05-12
Nemesis Append Fix


frombs:
When the append option is invoked for Nemesis file outputs in MOOSE, a seg fault occurs when the nodal_solution is written to file.  A screenshot of the fault along with the trace file is attached.  Please note that this occurs when the "--recover" option is used in MOOSE.

![Nemesis_append_seg_fault](https://user-images.githubusercontent.com/8171022/81644785-a3048e00-93e5-11ea-8be6-019ca0f43842.png)

This was previously reported by @permcody last year on the [mailing list](https://sourceforge.net/p/libmesh/mailman/message/36674583/). Also reference [Idaholab#13396](https://github.com/idaholab/moose/issues/13396) and the conversation at [Idaholab#13415](https://github.com/idaholab/moose/pull/13415).

I am not sure if this issue only applies to Moose but was able to resolve it by rebuilding the `exodus_node_num_to_libmesh` map within the `Nemesis_IO::prepare_to_write_nodal_data` function call after the `append` flag is invoked.

You can see the changes in my branch linked below.  I am not sure if there is a better way to handle this issue and would appreciate your feedback in order to resolve it.

jwpeterson:
Hi @frombs this sounds like a good improvement. There's probably other stuff that's not working quite right with appending to Nemesis files (for example I don't know what will happen if the mesh partitioning changes between writes) but this sounds like a step in the right direction. Would you mind opening a PR to see how it does with the rest of the CI testing?

roystgnr:
> for example I don't know what will happen if the mesh partitioning changes between writes

IIRC even in Exodus you have to start a new file after a mesh change, right?  Or do we support appends there now?

jwpeterson:
IIRC even in Exodus you have to start a new file after a mesh change, right? Or do we support appends there now?

No, that's correct as far as changing the number/positions of `Nodes`/`Elems`, but for `ReplicatedMesh` the partitioning doesn't matter because we use the `MeshSerializer`... but it would matter for Nemesis.

frombs:
> Would you mind opening a PR to see how it does with the rest of the CI testing?

@jwpeterson, I opened the PR but the checks need to be activated.

PR #2566, opened 2020-05-12
Allow Nemesis Appending During MOOSE Recover

frombs:
Ref #2564

### DIFF
--- a/include/mesh/nemesis_io_helper.h
+++ b/include/mesh/nemesis_io_helper.h
@@ -297,6 +297,20 @@ public:
   virtual void initialize(std::string title, const MeshBase & mesh, bool use_discontinuous=false) override;
 
   /**
+   * This function uses global communication routines to determine the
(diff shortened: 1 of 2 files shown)


jwpeterson:
@frombs I sent you an invite to join the group which gets their PRs tested automatically. I'll start these ones by hand now.

roystgnr:
After this is in, can you get a test case covering the previously-broken behavior into MOOSE CI?  I fear this would be an easy regression to make otherwise.

frombs:
@jwpeterson, thanks for adding me to the group.

frombs:
> After this is in, can you get a test case covering the previously-broken behavior into MOOSE CI? I fear this would be an easy regression to make otherwise.

@roystgnr, this case should be covered by [idaholab#13415](https://github.com/idaholab/moose/pull/13415). I will contact @permcody and see how he wants to proceed.

PR #2567, opened 2020-05-12
Initial refactoring of ConstrainDirichlet class

jwpeterson:
The eventual goal is to change the order of the loops from elem-loop-innermost:
```
for (dirichlet : dirichlets)
  for (var : dirichlet->variables)
    apply_dirichlet_impl<T>(range, var, *dirichlet, ...);
      for (elem : range)
```
to an order in which there is only one total loop over the elems:
```
for (elem : range)
  for (dirichlet : dirichlets(elem))
    for (var : dirichlet->variables)
      apply_dirichlet_impl<T>(elem, var, *dirichlet, ...);
```
This new approach should work better in cases where there are many DirichletBoundary objects, but each one is defined for only a few (or 1) nodes or elements, since we will go from N loops over the elements down to 1.

This initial round of refactoring does not yet re-order the loops, but makes the code a bit more modular by employing a FEMContext object. The FEMContext is used to store all the required elem, side, and edge FE objects, quadrature rules, DOF indices, etc.

Note: the FEMContext used in ConstrainDirichlet is assigned an algebraic_type of DOFS_ONLY because we aren't allowed to access current_local_solution vector values when ConstrainDirichlet is called...since they may not be initialized yet.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -1183,29 +1183,33 @@ public:
      * input function \p f gives the arbitrary solution.
      */
 
+    // Check for a quick return in case there's no work to be done.
+    if (dirichlets.empty())
+      return;
+
     // Figure out which System the DirichletBoundary objects are
-    // defined for. Verify that all Variables are defined on the same
-    // System.
+    // defined for. We break out of the loop as soon as we encounter a
+    // valid System pointer, the assumption is thus that all Variables
+    // are defined on the same System.
     System * system = nullptr;
     for (const auto & dirichlet : dirichlets)
       for (const auto & var : dirichlet->variables)
         {
           const Variable & variable = dof_map.variable(var);
-          auto current_system = variable.system();
+          system = variable.system();
 
-          if (!system)
-            system = current_system;
-          else if (current_system != system)
-            libmesh_error_msg("All variables should be defined on the same System");
+          if (system)
+            goto done;
         }
 
-    // If we found no System it could be because there are no
-    // DirichletBoundary objects or there are DirichletBoundaries but
-    // they have no Variables defined on them? This seems like a
-    // strange situation but it would also imply there is nothing to
-    // be done, so let's just return in that case.
+    // If we found no System, it could be because none of the
+    // Variables have one defined, or because there are
+    // DirichletBoundary objects with no Variables defined on
+    // them. These situations both indicate a likely error in the
+    // setup of a problem, so let's throw an error in this case.
+  done:
     if (!system)
-      return;
+      libmesh_error_msg("Valid System not found for any Variables.");
 
     // Construct a FEMContext object for the System on which the
     // Variables in our DirichletBoundary objects are defined. This
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -254,7 +254,7 @@ private:
   DofMap                  & dof_map;
   const MeshBase          & mesh;
   const Real               time;
-  const DirichletBoundary  dirichlet;
+  const DirichletBoundaries & dirichlets;
 
   const AddConstraint     & add_fn;
 
@@ -296,7 +296,9 @@ private:
   void apply_dirichlet_impl(const ConstElemRange & range,
                             const unsigned int var,
                             const Variable & variable,
-                            const FEType & fe_type) const
+                            const FEType & fe_type,
+                            const DirichletBoundary & dirichlet,
+                            FEMContext & fem_context) const
   {
     typedef OutputType                                                      OutputShape;
     typedef typename TensorTools::IncrementRank<OutputShape>::type          OutputGradient;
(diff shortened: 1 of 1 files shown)


jwpeterson:
@milljm have you seen any errors like this on the Test Mac box before? It's not related to my PR.
```
/var/folders/zz/zyxvpxvq6csfxvn_n00003x80000z_/T/tmp3ncpgts2: line 319: printf: `{': invalid format character
PROMPT={$(conda_env)$(me_at)$(git_current_branch) } ############################ END ENVIRONMENT ############################
ERROR: exiting with code 77
```

milljm:
Was it the Mac test that failed? I am assuming so due to the temp directory path... But no, I have not seen that error before :/ Odd. And if it was the Mac target, it was odd enough there as well as an invalidation solved the issue.

jwpeterson:
> Was it the Mac test that failed?

Yeah, and it passed when I invalidated so who knows... :man_shrugging:

jwpeterson:
@roystgnr it looks like only the `fv_2D_constant_scalar_advection` failed, so not related to this PR. Unfortunately the MOOSE modules tests that would normally run after that did not run, so not sure what we should do. @lindsayad tells me that fv_2D test fix should be merged now so maybe we can just wait awhile and try again once it's in MOOSE master.

lindsayad:
It's in MOOSE master now. Just invalidate

src/base/dof_map_constraints.C
```diff
@@ -1133,32 +1183,80 @@ public:
...
+        {
+          const Variable & variable = dof_map.variable(var);
+          auto current_system = variable.system();
 
-        const FEType & fe_type = variable.type();
+          if (!system)
+            system = current_system;
+          else if (current_system != system)
+            libmesh_error_msg("All variables should be defined on the same System");
+        }
```
roystgnr, line 1200: This part seems a bit costly in your extreme thousand-DirichletBoundary cases.  Sure you don't want it wrapped in ifndef NDEBUG?

src/base/dof_map_constraints.C
```diff
@@ -1133,32 +1183,80 @@ public:
...
 
-        if (fe_type.family == SCALAR)
-          continue;
+    // If we found no System it could be because there are no
+    // DirichletBoundary objects or there are DirichletBoundaries but
+    // they have no Variables defined on them? This seems like a
+    // strange situation but it would also imply there is nothing to
+    // be done, so let's just return in that case.
+    if (!system)
+      return;
```
roystgnr, line 1208: This, on the other hand, seems a bit far in the other direction; we can test for those cases from within library code so if we find ourselves here anyway I'd rather libmesh_assert(system) and figure out what went wrong.

roystgnr:
Looks good for step 1.  I reserve the right to revert in a panic if @vikramvgarg discovers one of these steps breaking the adjoint dirichlet boundaries somehow, though; I don't think we have any test coverage asserting correctness of that.

vikramvgarg:
@roystgnr adjoints example 6 uses adjoint dirichlet boundaries. But we havent put in any asserts to test the output so far.

roystgnr:
Yeah, I noticed that we have assertions there but commented out.  Are those atavisms from us forking that code from a previous adjoints example, changing the problem and therefore the solution, and then not recalculating "gold" results to use for regression testing?  Would it be hard to fix that?  I'm paranoid that the adjoint dirichlet BCs are too fragile without any test coverage.

src/base/dof_map_constraints.C
jwpeterson: You're right... this started as a loop with a `break` as soon as a valid System was found but then it went through a couple iterations... I can put it back the way it was originally.

src/base/dof_map_constraints.C
jwpeterson: OK, I will change this.

vikramvgarg:
I think we abandoned that example once we realized that the solution to the stabilization adjoint consistency issues lay elsewhere.

As it stands, that example is not very useful unfortunately. We should replace it with a real model adaptivity problem that uses both residual and adjoint evaluation physics. 

Example 1 and 2 are ideal for adjoint Dirichlet coverage. Example 2 will then test adjoint dirichlet + adjoint sensitivities, which is probably the most fragile part of the adjoint capability.

I can make a branch for modifying example 1 and 2 to use adjoint dirichlet.

roystgnr:
That'd be great; thanks!

Ideally we'd have the adjoint dirichlet vs interior QoI be a config entry, to make any issues that do crop up easier to debug (and so we could just run both options if we decide we need a little more regression test coverage).

src/base/dof_map_constraints.C
```diff
@@ -1183,29 +1183,33 @@ public:
      * input function \p f gives the arbitrary solution.
      */
 
+    // Check for a quick return in case there's no work to be done.
+    if (dirichlets.empty())
+      return;
```
roystgnr, line 1188: You could probably drop these lines.  In the unexpected dirichlets.empty() case that would take an extra branch (one for the empty loop over dirichlets but then a second to test ```!system```), but in the common case it'd save a branch, and of course it makes the code a couple lines shorter.

roystgnr: That's assuming we want to consider "applying an empty set of dirichlet constraints" to be a bug, though.  Which is reasonable as long as we're just calling it internally, but I guess external usage might want to be more flexible.  So you could leave it.

jwpeterson: I can't drop these lines because I no longer `return` when a valid System isn't found, I now throw an error. So I need to distinguish between the valid cases where there are no DirichletBoundary objects (e.g. introduction_ex2 whose failure originally brought this to my attention) and cases where there are DirichletBoundary objects but they aren't set up correctly in one way or another.

roystgnr:
Since updating to post this PR I'm seeing a segfault, triggered by fem_system_ex3, at dof_map_constraints.C:812 when phi is empty.

Not sure how we're making it to that line (on a HEX8!) in the first place, but I'm suspicious of the fact that I still see a 1 in the free_dof array.  I'm looking into it now.

jwpeterson:
OK... I wonder why we didn't see it on any of the CI runs though...

jwpeterson:
Possibly it's an FE pre-request issue. If you are building with `--disable-deprecated` locally, missing FE pre-requests behave differently.

roystgnr:
Hmm.  This is a ```--disable-strict-lgpl --enable-everything --enable-parmesh --enable-paranoid-warnings --enable-werror``` build.  No ```--disable-deprecated``` this time; libmesh_config.h still says LIBMESH_ENABLE_DEPRECATED is 1.

I'm also baffled that we don't see it on the CI runs.  I wouldn't expect parmesh or any of the everything options to make a difference here.

More suspicious than free_dof here is dof_is_fixed - all still false!  It looks like we've hit a place where is_boundary_node is false for all nodes, but is_boundary_edge is true for one edge.

...

And that's happening because we don't run through is_node_on_edge earlier, whereas we do run through is_node_on_side.  And that didn't break before because of the later edge-vs-side typos you fixed in this PR.

Still not sure why it's not breaking in CI though.  OOB undefined behavior that's just happening to work with some compilers but not others??

jwpeterson:
> because of the later edge-vs-side typos you fixed in this PR.

So maybe those weren't typos...? It's possible you were intentionally using the same container for this purpose and I just misunderstood that. If we un-fix the typo, does the segfault go away?

> OOB undefined behavior

Definitely possible in opt mode runs. I think we turned off the pedantic debugging stuff on CI (?) so maybe also getting UB in Debug Tests.

PR #2568, opened 2020-05-12
Check whether abs(det(A)) is small, not det(A).

jwpeterson:
Refs #2552.

### DIFF
--- a/src/fe/fe_base.C
+++ b/src/fe/fe_base.C
@@ -767,12 +767,10 @@ void FEGenericBase<Real>::compute_dual_shape_coeffs ()
   }
 
 #ifndef NDEBUG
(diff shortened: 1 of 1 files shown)


roystgnr:
How does this help?  Even an L2 seminorm matrix ought to be symmetric positive semidefinite, so even with roundoff error we shouldn't see any large negative eigenvalues.  Are there cases where det(A) is turning out to be negative but with high absolute value?

jwpeterson:
My gut reaction was that det(A) can be negative in general but tou're right, I now realized A is a mass matrix so if det(A) < 0 then that indicates some other type of error condition which we should also warn about, not take the absolute value of.

PR #2569, opened 2020-05-12
Make Elem::interior_parent always safe to call

recuero:
In the most general case, when calling `interior_parent` we may not know whether the element's dimension coincides with the system's dimension (`LIBMESH_DIM == elem->dim()`). This PR aims to make `interior_parent` always safe to call.

### DIFF
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -1056,7 +1056,7 @@ public:
    * by storing the D-dimensional parent.  This method provides access to that
    * element.
    *
-   * This method is not safe to call if this->dim() == LIBMESH_DIM; in
+   * This method returns nullptr if this->dim() == LIBMESH_DIM; in
    * such cases no data storage for an interior parent pointer has
    * been allocated.
    */
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -770,7 +770,8 @@ void Elem::find_interior_neighbors(std::set<Elem *> & neighbor_set)
 const Elem * Elem::interior_parent () const
 {
   // interior parents make no sense for full-dimensional elements.
-  libmesh_assert_less (this->dim(), LIBMESH_DIM);
+  if (this->dim() >= LIBMESH_DIM)
+      return nullptr;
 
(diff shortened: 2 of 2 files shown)


roystgnr:
What's the advantage of the duplicated work here over
```bool Elem::has_interior_parent() { return this->interior_parent(); }```?

For that matter, what's the advantage of adding a new API over user code just testing interior_parent()?  We distinguish e.g. MeshBase::elem_ptr from MeshBase::query_elem_ptr because we want to assert non-null in the common case, but we never added that level of assertions to Elem::interior_parent.

recuero:
I believe `has_interior_parent()` will allow to find out if any element, whether or not it fulfills the conditions to be have an interior element, has an interior parent. `interior_parent()` may throw in one of the asserts if the element is not even a candidate.

lindsayad:
> What's the advantage of the duplicated work here over
> `bool Elem::has_interior_parent() { return this->interior_parent(); }`?

Our documentation explicitly states that `interior_parent` is not safe to call if `elem->dim() == LIBMESH_DIM`. We want a method that's safe to call at all times

lindsayad:
If we make `interior_parent()` safe to call in all cases, e.g. just return `nullptr` for example if `LIBMESH_DIM == elem->dim()`, then it makes sense *not* to add this new API

roystgnr:
> If we make interior_parent() safe to call in all cases, e.g. just return nullptr for example if LIBMESH_DIM == elem->dim(), then it makes sense not to add this new API

Yeah, that sounds like a better option to me.

There's a second assertion in interior_parent() too, but whereas the first assertion is basically "assert that the user isn't calling this where it's guaranteed to return null" and can safely be removed, the second is "assert that the mesh isn't incomprehensibly corrupted", and can safely be kept.

jwpeterson:
I think we need to update the documentation as well... the stuff about "not safe to call" should be changed to "returns nullptr".

jwpeterson:
I'm fine merging without waiting for the CI again if @roystgnr is.

roystgnr:
Yeah, this works for me.  Thanks!

PR #2571, opened 2020-05-15
Fix adjoints example 5

vikramvgarg:
We had a few bugs in adjoints example 5.

1) The analytic solution had an error in an exponent, did not cause a big error in the QoI evaluation, but was making the sensitivity off by a factor of 2.
2) We were not scaling the mass matrix and adjoint weighted residuals correctly by deltat.

Bugs have been fixed, and gold value corrected.

### DIFF
--- a/examples/adjoints/adjoints_ex5/adjoints_ex5.C
+++ b/examples/adjoints/adjoints_ex5/adjoints_ex5.C
@@ -31,11 +31,16 @@
 // T(boundary;t) = 0
 
 // For these initial and boundary conditions, the exact solution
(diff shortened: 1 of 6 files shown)


vikramvgarg:
Wait, looks like I ended up combining commits from two separate branches here.

jwpeterson:
In case it's helpful, you can start a new branch and use `git cherry-pick` to grab the individual commits off of this branch that you want to keep.

vikramvgarg:
> In case it's helpful, you can start a new branch and use `git cherry-pick` to grab the individual commits off of this branch that you want to keep.

Thanks, that'll help.

PR #2573, opened 2020-05-15
Fix adjoints example 5

vikramvgarg:
We had a few bugs in adjoints example 5.

1. The analytic solution had an error in an exponent, did not cause a big error in the QoI evaluation, but was making the sensitivity off by a factor of 2.
2. We were not scaling the mass matrix and adjoint weighted residuals correctly by deltat.

Bugs have been fixed, and gold value corrected.


roystgnr:
Well this is good to find out.  Is it possible we're not thread safe somewhere in the QoI here?

Let me see if I can replicate the gold standard and/or the failure.

roystgnr:
Well, I can replicate the failure.  I'm getting -5.30953 even in serial.

PR #2574, opened 2020-05-16
Second part of the DirichletBoundary refactoring

jwpeterson:
This PR reorders the loops in `ConstrainDirichlet::operator()` so that there is only one loop over the elements. The diff looks large but almost all of that is reindentation of the `apply_dirichlet_impl()` function. For a more manageable diff you can pull the branch and run:
```
git diff dee0a3ec8..23d7cf5eb
```
The history is also messy so once we this is working and we are happy with it, I'll squash everything down to 2 commits, one with actual changes and one with reindenting.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,7 +1,7 @@
 This is the output of "git shortlog -ns", updated periodically.
 
-  7514  Roy H. Stogner
-  6598  John W. Peterson
+  7517  Roy H. Stogner
+  6607  John W. Peterson
   2730  Benjamin S. Kirk
    792  David Knezevic
    391  Paul T. Bauman
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -571,678 +571,672 @@ private:
           }
       }
 
-    // Iterate over all the elements in the range
-//    for (const auto & elem : range)
-//      {
+    // There's a chicken-and-egg problem with FEMFunction-based
+    // Dirichlet constraints: we can't evaluate the FEMFunction
+    // until we have an initialized local solution vector, we
+    // can't initialize local solution vectors until we have a
+    // send list, and we can't generate a send list until we know
+    // all our constraints
+    //
+    // We don't generate constraints on uninitialized systems;
+    // currently user code will have to reinit() before any
+    // FEMFunction-based constraints will be correct.  This should
+    // be fine, since user code would want to reinit() after
+    // setting initial conditions anyway.
+    if (f_system && context.get())
+      context->pre_fe_reinit(*f_system, elem);
 
-        // There's a chicken-and-egg problem with FEMFunction-based
-        // Dirichlet constraints: we can't evaluate the FEMFunction
-        // until we have an initialized local solution vector, we
-        // can't initialize local solution vectors until we have a
-        // send list, and we can't generate a send list until we know
-        // all our constraints
-        //
-        // We don't generate constraints on uninitialized systems;
-        // currently user code will have to reinit() before any
-        // FEMFunction-based constraints will be correct.  This should
-        // be fine, since user code would want to reinit() after
-        // setting initial conditions anyway.
-        if (f_system && context.get())
-          context->pre_fe_reinit(*f_system, elem);
+    // Also pre-init the fem_context() we were passed on the current Elem.
+    fem_context.pre_fe_reinit(fem_context.get_system(), elem);
 
-        // Also pre-init the fem_context() we were passed on the current Elem.
-        fem_context.pre_fe_reinit(fem_context.get_system(), elem);
+    // Get a reference to the DOF indices for the current element
+    const std::vector<dof_id_type> & dof_indices =
+      fem_context.get_dof_indices(var);
 
-        // Get a reference to the DOF indices for the current element
-        const std::vector<dof_id_type> & dof_indices =
-          fem_context.get_dof_indices(var);
+    // The number of DOFs on the element
+    const unsigned int n_dofs =
+      cast_int<unsigned int>(dof_indices.size());
 
-        // The number of DOFs on the element
-        const unsigned int n_dofs =
-          cast_int<unsigned int>(dof_indices.size());
+    // Fixed vs. free DoFs on edge/face projections
+    std::vector<char> dof_is_fixed(n_dofs, false); // bools
+    std::vector<int> free_dof(n_dofs, 0);
 
-        // Fixed vs. free DoFs on edge/face projections
-        std::vector<char> dof_is_fixed(n_dofs, false); // bools
-        std::vector<int> free_dof(n_dofs, 0);
+    // The element type
+    const ElemType elem_type = elem->type();
 
-        // The element type
-        const ElemType elem_type = elem->type();
+    // Zero the interpolated values
+    Ue.resize (n_dofs); Ue.zero();
 
-        // Zero the interpolated values
-        Ue.resize (n_dofs); Ue.zero();
+    // In general, we need a series of
+    // projections to ensure a unique and continuous
+    // solution.  We start by interpolating boundary nodes, then
+    // hold those fixed and project boundary edges, then hold
+    // those fixed and project boundary faces,
 
-        // In general, we need a series of
-        // projections to ensure a unique and continuous
-        // solution.  We start by interpolating boundary nodes, then
-        // hold those fixed and project boundary edges, then hold
-        // those fixed and project boundary faces,
-
-        // Interpolate node values first. Note that we have a special
-        // case for nodes that have a boundary nodeset, since we do
-        // need to interpolate them directly, even if they're non-vertex
-        // nodes.
-        unsigned int current_dof = 0;
-        for (unsigned int n=0; n!= sebi.n_nodes; ++n)
+    // Interpolate node values first. Note that we have a special
+    // case for nodes that have a boundary nodeset, since we do
+    // need to interpolate them directly, even if they're non-vertex
+    // nodes.
+    unsigned int current_dof = 0;
+    for (unsigned int n=0; n!= sebi.n_nodes; ++n)
+      {
+        // FIXME: this should go through the DofMap,
+        // not duplicate dof_indices code badly!
+        const unsigned int nc =
+          FEInterface::n_dofs_at_node (dim, fe_type, elem_type,
+                                       n);
+        if ((!elem->is_vertex(n) || !sebi.is_boundary_node[n]) &&
+            !sebi.is_boundary_nodeset[n])
           {
-            // FIXME: this should go through the DofMap,
-            // not duplicate dof_indices code badly!
-            const unsigned int nc =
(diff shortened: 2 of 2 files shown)


roystgnr:
Hmm... adjoints_ex3 is using first order, so that failure can't be a projection-vs-interpolation thing.  (it also looks like 100% relative error).  Looks like an actual bug.

jwpeterson:
Yeah, it appears that `QoI_0_computed` is zero... I'm looking into this now. By the way, I added some print statements in `DofMap::create_dof_constraints()` and it seems that sometimes when it is called there are 0 DirichletBoundary objects, but then one time there are 5: 
```
Solving the adjoint problem
_dirichlet_boundaries->size()=0
_adjoint_dirichlet_boundaries.size()=0
_dirichlet_boundaries->size()=0
_adjoint_dirichlet_boundaries.size()=0
_dirichlet_boundaries->size()=0
_adjoint_dirichlet_boundaries.size()=0
_dirichlet_boundaries->size()=0
_adjoint_dirichlet_boundaries.size()=0
_dirichlet_boundaries->size()=0
_adjoint_dirichlet_boundaries.size()=0
_dirichlet_boundaries->size()=5
_adjoint_dirichlet_boundaries.size()=0
Adaptive step 5
```
Does that seem reasonable to you?

roystgnr:
Hmm... we don't actually add the DirichletBoundary objects until we get to CoupledSystem::init_data.  I'd have thought it possible for us to hit one or two create_dof_constraints() beforehand, but *five* sure sounds excessive.

vikramvgarg:
We dont use heterogenous adjoint boundary conditions here, so the number of adjoint dirichlet boundaries looks correct. However, we use dirichlet boundary conditions multiple variables (2 velocity components and concentration) on more than one type of boundary.

vikramvgarg:
> Hmm... we don't actually add the DirichletBoundary objects until we get to CoupledSystem::init_data. I'd have thought it possible for us to hit one or two create_dof_constraints() beforehand, but _five_ sure sounds excessive.

There are indeed five primal Dirichlet boundary conditions. No slip and penetration for velocity on wall, prescribed velocity and concentration at right and left inlets.

jwpeterson:
Yeah, there are definitely 5 BCs added in coupled_system.C,
```
(DirichletBoundary (wall_bdy, uv, &zero));
(DirichletBoundary (left_inlet_bdy, uv, &inflow_left));
(DirichletBoundary (left_inlet_bdy, C_only, &one));
(DirichletBoundary (right_inlet_bdy, uv, &inflow_right));
(DirichletBoundary (right_inlet_bdy, C_only, &zero));
``` 
so the amount seems right, just a bit surprising how many times `create_dof_constraints()` is called before we actually have any DirichletBoundaries.

roystgnr:
Vikram pointed out over IM that we do elevate the order on the velocity elements, so there could be projection-vs-interpolation result changes, even if not of the magnitude we're seeing.

Still no idea where those many 0-boundary calls are coming from.  On a lot of codes we do some uniform refinement to start out, but I don't see it here, only adaptive refinement subsequent to the first solve.

jwpeterson:
> so there could be projection-vs-interpolation result changes, even if not of the magnitude we're seeing.

This refactoring hasn't changed the projection code (not intentionally, anyway). Just reordered the loops so that there is one loop over the Elems instead of one per DirichletBC.

jwpeterson:
I turned `output_exodus` on and looked at the primal output files in Paraview... they are showing u=v=p=0, while C is nonzero. So it's like the velocity inlet conditions aren't being applied at all.

vikramvgarg:
The u and v inlets are the bcs for which we supply a user defined BdyFunction (inflow_left and inflow_right). So these might not be getting attached or evaluated.

jwpeterson:
I think I have some idea what's going on... it's related to the fact that I've incorrectly combined the `is_boundary_side[s]` check which used to be per boundary id into a union over all boundary ids. So in this particular example, on the coarse mesh there is one inlet element, but it has three boundary sides, one of which is the actual inlet and the other two of which are no-slip surfaces. 

When trying to set u values on the inlet (boundary id 1), the following happens: 
```
DirichletBoundary object has boundary ids: 
1 
Applying boundary condition for var = 0
Variable name is u
Projecting side values...
Projecting side values for elem 4, side 0
At qp 0, f_accessor(0)=-0
At qp 1, f_accessor(0)=-0
At qp 2, f_accessor(0)=-0
At qp 3, f_accessor(0)=-0
Projecting side values for elem 4, side 1
At qp 0, f_accessor(0)=-0.06461106321354794
At qp 1, f_accessor(0)=-0.22110322250073797
At qp 2, f_accessor(0)=-0.22110322250073797
At qp 3, f_accessor(0)=-0.064611063213547551
Projecting side values for elem 4, side 2
At qp 0, f_accessor(0)=0
At qp 1, f_accessor(0)=0
At qp 2, f_accessor(0)=0
At qp 3, f_accessor(0)=0
```
In reality, only side1 is on boundary id 1 and the other two sides are on boundary id 3 (no-slip boundary) so I think this is the source of the problem.

jwpeterson:
OK, so to expand on the previous comment, I confirmed that the issue is I have a single `is_boundary_side` array for a given Elem, and therefore in the failing test when I should have:
```
is_boundary_side[DirichletBoundary_0] = {false, true, false, false}
is_boundary_side[DirichletBoundary_1] = {true, false, true, false}
```
for the `Elem` in question, I instead have:
```
is_boundary_side = {true, true, true, false}
```
which is like the component-wise OR of the two vectors above. Then what happens is DirichletBoundary_0 is applied to sides 0, 1, and 2, followed by DirichletBoundary_1 being applied to sides 0, 1, and 2, but since we forbid constraint overwrites, only DirichletBoundary_0, which in this case happens to be the no-slip BC, is imposed, leading to the current issue.

I think in general, I need to have an `is_boundary_side` array *per* `DirichletBoundary` object which is active for a given Elem, in other words a 
```
map<DirichletBoundary*, vector<bool>> is_boundary_side_map;
```
and similarly for `is_boundary_node`, `is_boundary_edge`, `is_boundary_shellface`, etc. Then in apply_dirichlet_impl I'll have to consult this map based on the current DirichletBoundary object we are processing, and that should fix the problem.

roystgnr:
> I think in general, I need to have an is_boundary_side array per DirichletBoundary object which is active for a given Elem

This is exactly the sort of thing we were trying to avoid in your zillion-DirichletBoundary case, though, isn't it?

I still think my original idea is the way to go: look up relevant DirichletBoundary objects, in a multimap keyed off boundary_id_type, while iterating over each ids_vec.

jwpeterson:
> This is exactly the sort of thing we were trying to avoid in your zillion-DirichletBoundary case, though, isn't it?

I don't think it will be too bad, `is_boundary_side_map` I'm proposing is recomputed per Elem, and so only has entries for the DirichletBoundaries that are active on the current Elem.

> I still think my original idea is the way to go: look up relevant DirichletBoundary objects, in a multimap keyed off boundary_id_type, while iterating over each ids_vec.

I think that's basically what the `boundary_id_to_dirichlet_boundaries` in my implementation is, although it's a map from boundary_id -> `std::set<DirichletBoundary>` which are active on that id...

roystgnr:
> I don't think it will be too bad

Okay.  Just to be clear, I'm not trying to insist on or veto or even really push anything; since you're the one with the performance-stress-testing case I'll be happy with whatever your benchmarking is happy with in the end.

> only has entries for the DirichletBoundaries that are active on the current Elem.

You're right, I wasn't taking that into account.  I'd still guess that my suggestion would be faster but my guess is now an uncertain "by a tiny constant factor" rather than a confident "asymptotically".

jwpeterson:
This latest push is now passing adjoints_ex3 for me.

@roystgnr I decided it would probably be easier to temporarily revert your fixes in #2578 since they conflicted with my larger set of changes, and then reapply them on top of this latest work if possible. I may require your assistance with this (or at least with getting a reproducible failing test case) as `is_boundary_side`, `is_boundary_node` etc. flags are now stored per `DirichletBoundary` object for the current Elem, but otherwise the fixes you made should still be applicable.

jwpeterson:
Also I haven't rebased this branch yet, but at this point the history is too confusing to be useful/helpful in the future, so once we get your fixes re-applied I'll probably just make this whole branch into a single commit, possibly 2 if I can figure out a way to separate out the reindentation, which must still be over 70% of the overall change.

jwpeterson:
The following 4 invocations of fem_system_ex3 all run fine for me on my laptop:
```
./example-dbg deltat=0.25 n_timesteps=5 time_solver=newmark write_interval=1 solver_quiet=false relative_step_tolerance=1e-3 relative_residual_tolerance=1.e-3 -ksp_type cg -pc_type jacobi
./example-dbg time_solver=steady n_timesteps=1 write_interval=1 solver_quiet=false relative_step_tolerance=1e-3 relative_residual_tolerance=1.e-3 -ksp_type cg -pc_type jacobi
./example-dbg deltat=0.25 n_timesteps=5 time_solver=euler theta=0.5 write_interval=1 solver_quiet=false relative_step_tolerance=1e-3 relative_residual_tolerance=1.e-3 -ksp_type gmres -pc_type bjacobi -sub_pc_type ilu
./example-dbg deltat=0.25 n_timesteps=5 time_solver=euler2 theta=0.5 write_interval=1 solver_quiet=false relative_step_tolerance=1e-3 relative_residual_tolerance=1.e-3 -ksp_type gmres -pc_type bjacobi -sub_pc_type ilu
```
so @roystgnr please let me know if you can still reproduce the segfault/OOB array access you described in #2578 (and how to reproduce it) with this branch.

roystgnr:
Distilled version for public and posterity of my frustrated ranting to John on IM:

Thanks perhaps to upgrading systems and futzing with compilers, I was no longer able to get that bug to turn into a segfault.  Doing some rebasing gets the previous bugfix into this branch, though I'd feel more comfortable if we had better testing on the failure.

jwpeterson:
@roystgnr here's a simplified version of the branch which has the conflicts with your bugfixes. Hopefully the rebase is a bit easier for you now.

roystgnr:
Rebase is done.  Whether it's *working* is another question, though.

jwpeterson:
@roystgnr Thanks! This compiles and adjoints_ex3 works fine for me, but it looks like CIVET did not run again after your push, maybe it got confused by the rebasing. I'll push a dummy commit to see if I can get it to run again...

jwpeterson:
Hurray! I'm happy with this if you are @roystgnr

roystgnr:
Can't say I'm *happy* yet, since we don't have enough test coverage on some cases, but I'm definitely much *happier* than I was with the previous code.  Thanks for all this!

jwpeterson:
OK. I'm hoping the third refactoring (to allow interpolation for Lagrange elements) will be ready soon. BTW we ran a few small test cases already with just this refactoring, and in one of them the time taken for `create_dof_constraints()` went from 28% to 12% of the total runtime, so it already looks like a good improvement.

PR #2577, opened 2020-05-19
Refactor unsteady adjoints FSH usage

vikramvgarg:
Fixing the incorrect analytic solution in adjoints example 5 lead to further corrections in the way unsteady solver uses solution histories, and how the user code utilizes the advance_timestep and adjoint_advance_timestep methods. 

Also the final adjoint sensitivity integral is calculated using the trapezoidal rule now.


roystgnr:
The failure with complex numbers looks real; there are a few places you're using Real where you needed Number.

src/solvers/unsteady_solver.C
```diff
@@ -151,14 +151,28 @@ void UnsteadySolver::solve ()
...
-      // User has to attach appropriate solution_history object for this to
-      // actually store anything anywhere
+      // We call advance_timestep in user code after solve, so any solutions
+      // we will be storing will be for the next time instance
+      _system.time += _system.deltat;
+
+      // If the user has attached a memory or file solution history
+      // object to the solver, this will store the current solution
+      // indexed with the current time
+      solution_history->store(false);
```
roystgnr, line 165: Why not bring this outside the if/else?

src/solvers/unsteady_solver.C
```diff
@@ -177,24 +191,36 @@ void UnsteadySolver::advance_timestep ()
...
-      solution_history->store(true);
-      // Decrement the system time
+      // The adjoint system has been solved. We need to store the adjoint solution and
+      // load the primal solutions for the next time instance (t - delta_ti).
       _system.time -= _system.deltat;
+
+      // Retrieve the primal solution vectors at this new time instance (t - delta_ti)). These provide the
+      // data to solve the adjoint problem for the next time instance (t - delta_ti - delta_ti+1).
+      solution_history->retrieve(true);
+
```
roystgnr, line 206: Likewise for the retrieve+stores here

examples/adjoints/adjoints_ex5/heatsystem.h
```diff
@@ -55,7 +55,16 @@ class HeatSystem : public FEMSystem
   // Sensitivity Calculation
   Number & compute_final_sensitivity()
   {
-    final_sensitivity = -(R_plus_dp - R_minus_dp)/(2*dp);
+    final_sensitivity = 0.0;
+
+    // Use the trapezoidal rule to compute the sensitivity integral
```
roystgnr, line 60: Good catch with this, but it's only correct here because we're doing first-order time stepping with constant deltat.  In the future we ought to try to figure out some way to abstract this into the TimeSolver (for the integration rule) combined with the SolutionHistory (for the deltat changes).

roystgnr:
This looks good, aside from the couple redundancies in branches.  INL's looking into the Test Mac failures ("Probably an XCode update thing"), but if they can't fix it quickly then IMHO we can merge regardless; nothing here should have had a Mac-specific effect so I'm not too worried if we don't get a proper test.

examples/adjoints/adjoints_ex5/heatsystem.h
vikramvgarg: IMO, we should further abstract our adjoint senstivity methods to a class like we do for error estimation, where the right function to compute sensitivity is used depending on the TimeSolver and SolutionHistory.

src/solvers/unsteady_solver.C
vikramvgarg: Thanks for taking care of this !

Issue #2345, opened 2019-11-20
Refactor FParser code generation and just in time compilation


dschwen:
I need this to be more flexible to support JIT with dual numbers and batch compilation. In batch compilation multiple functions will be compiled in a single pass. This should reduce overhead from multiple compiler launches. This functionality will require a global storage (or material object level management of batches) for function objects, which will most likely be implemented on the MOOSE side.

roystgnr:
Sounds great, but I'm kind of surprised this is a problem.  How many function objects per executable are you guys playing with?

dschwen:
Potentially hundreds! In a multiphase system with many components you need derivatives and second derivatives (with cross derivatives) of free energies for multiple phases, in certain scenarios you also need third order derivatives...

roystgnr:
That'd do it, then!

Would it be possible to make batch compilation available to everyone, and avoid the global management problem, by doing some sort of registration when parsers are instantiated but postponing compilation until the first evaluation call?  Typically apps (including MOOSE) build all their ParsedFunction+ParsedFEMFunction objects during a "setup" phase, right?  So by the time an evaluation gets called, every object would be registered and we'd be able to do all the compilation in one sweep then.

dschwen:
Hm, this would have to be thought out carefully. Eval can be called in threaded regions, and the cache files are named based on a hash of the function. In moose JIT is triggered in the non-threaded setup phase, which allows me for threads >0 to copy construct functions from thread zero versions. 

It sounds worthwhile to at least think about it. My current refactoring should make an implementation of a libmesh batch JIT compilation easier.

roystgnr:
> Hm, this would have to be thought out carefully.

I would be offended by your implication that I'm not thinking carefully, if it hadn't been followed by undeniable proof that I'm not thinking carefully...

> Eval can be called in threaded regions, and the cache files are named based on a hash of the function.

Yeah, I see now.  That's easy to fix with an extra lock, but then we trash eval performance.  I guess that since we know the "cache_exists" bool would only ever be changing from false to true, though, we would be able to trust a "true" and we'd only have to worry about locking *after* we saw a "false"?

PR #2579, opened 2020-05-19
Refactor compiled function pointer and pImmed update

dschwen:
Some more small changes to support dual numbers in parsed functions through JIT compilation. 
DualNumbers are C++ objects, which means they cannot legally be returned from `extern "C"` functions (it can work, but with output an ugly compiler warning), so I'm passing in a pointer to a return value (old school).

There is an upcoming Moose PR that depends on these changes.

Refs #2345

### DIFF
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -1,7 +1,6 @@
 #include "fparser_ad.hh"
 #include "extrasrc/fpaux.hh"
 #include "extrasrc/fptypes.hh"
-#include "extrasrc/fpaux.hh"
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


roystgnr:
I'm amazed this works.  C++ name mangling doesn't do ruinous things to you here?

dschwen:
hahah, nope, that's what `extern "C"` prevents. Or were you concerned about type names like DualReal being mangled? That doesn't happen.

jwpeterson:
Looks like we have some libmesh fparser unit tests in: 
```
numerics/composite_function_test.C
numerics/parsed_fem_function_test.C
numerics/parsed_function_test.C
```

PR #2580, opened 2020-05-22
Recognize more intel compilers

roystgnr:
Not sure how useful this will be, since the newest patches don't support MetaPhysicL, the less-patched 19 that @lindsayad was using has bugs, and the new 20 I tried seems to have other bugs, but at least it's a step in the right direction.

### DIFF
--- a/configure
+++ b/configure
@@ -6984,6 +6984,10 @@ fi
 
                   GXX_VERSION_STRING="`($CXX -V 2>&1) | grep 'Version '`"
                   case "$GXX_VERSION_STRING" in #(
+  *20.*) :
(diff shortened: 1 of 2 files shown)


dschwen:
Is there really an icc 20? I was told that parallel studio 2020 comes with
icc 19.1 ...

roystgnr:
Huh, looks like you're right.
```
$ icpc --version
icpc (ICC) 19.1.1.217 20200306
```
and so forth, and similarly for ```-V```

Oh, well.  They'll have an *actual* 20 out one of these days and we might as well give it the same flags as a start when it comes.

This makes me even more frustrated by those segfaults I was seeing with Intel "20", though.  That wasn't a failure of some brand spanking new major-version-incrementing optimizer, that was a regression from 19.0.5 to 19.1.1?  How disappointing.

PR #2582, opened 2020-05-26
Use max_node_id() in build_nodes_to_elem_map

roystgnr:
And also scream that this API should be deprecated in any use cases where the distinction is important.

Thanks to @fdkong for catching this.


fdkong:
Thanks

PR #2583, opened 2020-05-26
Respect original order of DirichletBoundary objects.

jwpeterson:
The refactoring in #2574 caused the DirichletBoundary objects'
constraints to be applied in an "arbitrary" order rather than the
order in which they were actually added to the DofMap.  Since users
may have been implicitly or explicitly relying on this ordering to
break "ties" in cases where multiple constraints were applied to the
same DOF, we should try to maintain that order, even though it was not
guaranteed previously. I also updated the documentation to make this
ordering guarantee an official part of the DofMap API.

* Add 'boundary_id_to_ordered_dirichlet_boundaries' which is ordered
  based on the original vector ordering.
* Add 'ordered_dbs' container to SingleElemBoundaryInfo struct.
* Update comment for DofMap::add_dirichlet_boundary()

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -303,9 +303,9 @@ private:
   struct SingleElemBoundaryInfo
   {
     SingleElemBoundaryInfo(const BoundaryInfo & bi,
-                           const std::map<boundary_id_type, std::set<DirichletBoundary *>> & map_in) :
+                           const std::map<boundary_id_type, std::set<std::pair<unsigned int, DirichletBoundary *>>> & ordered_map_in) :
       boundary_info(bi),
-      boundary_id_to_dirichlet_boundaries(map_in),
+      boundary_id_to_ordered_dirichlet_boundaries(ordered_map_in),
       elem(nullptr),
       n_sides(0),
       n_edges(0),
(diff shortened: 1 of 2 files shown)


src/base/dof_map_constraints.C
```diff
@@ -1406,17 +1406,20 @@ public:
...
+    // Map from boundary_id -> set<pair<id,DirichletBoundary*>> objects which
     // are active on that boundary_id. Later we will use this to determine
     // which DirichletBoundary objects to loop over for each Elem.
-    std::map<boundary_id_type, std::set<DirichletBoundary *>>
-      boundary_id_to_dirichlet_boundaries;
+    std::map<boundary_id_type, std::set<std::pair<unsigned int, DirichletBoundary *>>>
+      boundary_id_to_ordered_dirichlet_boundaries;
 
-    for (const auto & dirichlet : dirichlets)
+    for (auto dirichlet_id : index_range(dirichlets))
```
roystgnr, line 1415: Unrelated to this PR, but I wonder if it would make sense to create an enumerate_range() function to iterate over both range and value at once, in a std::pair?  Aside from saving a line of code, that would enable optimizers to stop fretting about whether the vector was getting sneakily resized by any functions called from inside the loop.

Or would that be too much of a pain to use and/or to read (e.g. dirichlet.first and dirichlet.second being much less self-documenting than dirichlet_id and dirichlet) without C++17 structured bindings?

jwpeterson: Yeah I think your idea makes good sense once we require C++17 and structured bindings. IIRC there are several places in the library where we currently use an `index_range` but an `enumerate_range` is what we really want.

jwpeterson:
Thanks for reviewing!

PR #2584, opened 2020-05-27
Enforce Ordering of Nodal Variables During Nemesis Output

frombs:
closes libMesh#2581

### DIFF
--- a/src/systems/equation_systems.C
+++ b/src/systems/equation_systems.C
@@ -38,6 +38,8 @@
 #include "libmesh/elem.h"
 #include "libmesh/libmesh_logging.h"
 
+#include <numeric> // std::iota
+
 // Include the systems before this one to avoid
 // overlapping forward declarations.
 #include "libmesh/equation_systems.h"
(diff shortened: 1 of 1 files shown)


frombs:
@roystgnr, the `nemesis_hide.i` test in moose needs to be regolded to allow the tests to pass. However, if I create a new PR in Moose, the test will also fail there without this PR being merged.  How should I proceed?

permcody:
Your PR to MOOSE needs to include the libmesh submodule update too. CIVET
will build _that_ version of libmesh with your patch and then your regolded
test will work.

jwpeterson:
There's the usual chicken-and-egg problem though, since @frombs's commit isn't in libmesh upstream, he can't use it for a libmesh submodule update in MOOSE. Unless you guys found a way around that issue? I can push a copy of this branch to libmesh upstream if that would be helpful...

lindsayad:
> There's the usual chicken-and-egg problem though, since @frombs's commit isn't in libmesh upstream, he can't use it for a libmesh submodule update in MOOSE. Unless you guys found a way around that issue? I can push a copy of this branch to libmesh upstream if that would be helpful...

I've found that the history of the git repo on the CIVET boxes can play a role here. Some libmesh submodules appear to fetch pull request HEADs while others do not

frombs:
@jwpeterson, it looks like moving this to upstream would be best [see comment here](https://github.com/idaholab/moose/pull/15352#pullrequestreview-420284535).

jwpeterson:
@frombs I pushed your branch to libmesh upstream, so it should be possible to include it as a libmesh submodule update in a MOOSE PR now. If such an update gets merged in MOOSE you'll have to let us know so we can make sure and merge *this* PR into libmesh as well.

lindsayad:
If we merge something like idaholab/moose#15352 with a submodule hash corresponding to this PR HEAD, then I guess what we would merge this as soon as the moose changes hit moose master? Otherwise testing on other libmesh PRs will show red right?

frombs:
@jwpeterson, this should be ready to merge now. The regolded Moose tests were merged and made it to the master branch and all tests are passing.

PR #2585, opened 2020-05-28
Remove long-deprecated code from libMesh

roystgnr:
A MOOSE app turned out to be breaking node unique_ids after accidentally using a Node copy constructor when it should have been using the constructor-from-Point.  So I'm looking at that (it's still deprecated, and it's now slicing to better mimic constructor-from-Point in --enable-deprecated configuration builds) and other deprecated code paths.

A couple methods in libMesh weren't being "properly" deprecated, and so get a reprieve for another year or so.  I also left in anything deprecated less than two years ago, and a couple base class APIs that have non-deprecated overrides in some subclasses but not *every* subclass.

But with those few exceptions, this PR removes anything that's been deprecated since 2017.

Or 2016.

Or 2015, 2014, 2013... 2012?  2011???  *2009?!?*  **2008?!?  Seriously?!?**


jwpeterson:
> Remove deprecated logging macros

You know I had to click through to find out what the 2008 deprecated code was! Very impressive that there was no deprecated code in MOOSE... I will check our stuff but no need to wait on merging for that.

One thing you may consider is adding the "Test No Deprecated" optional test to this PR? It might catch something, who knows.

PR #2586, opened 2020-05-28
Don't assert that ip level is <= boundary elem level

lindsayad:
@jwpeterson actually has a potential use case for this, so I think he
can decide whether he agrees with this or not. His libMesh mortar
element generation code builds maps that would be voided upon mesh
refinement. Hence mortar generation should occur after the higher
dimensional mesh has been refined. If mortar generation happens after
mesh refinement then obviously the interior parent will have a higher
level than the mortar element. I think this is a valid use case

### DIFF
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -793,11 +793,6 @@ const Elem * Elem::interior_parent () const
                   (interior_p == remote_elem) ||
                   (interior_p->dim() > this->dim()));
 
-  // We require consistency between AMR of interior and of boundary
-  // elements
-  if (interior_p && (interior_p != remote_elem))
-    libmesh_assert_less_equal (interior_p->level(), this->level());
-
   return interior_p;
 }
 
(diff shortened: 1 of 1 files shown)


lindsayad:
I guess we probably would be in trouble in the general case though if performing adaptivity after adding a boundary elem with a higher level interior parent...?

roystgnr:
How will the interior_parent have a higher level?  Even if the interior is more refined than the boundary, that just means that the interior_parent on the boundary should be pointing to an ancestor element of the same level in the interior, not to a random one of it's descendants.

lindsayad:
This is a mortar mesh that is generated on the fly. They are brand new elements. You can think of them as being equally refined as the interior (e.g. the size of a mortar element will never exceed the size of a (refined) interior element face)

roystgnr:
Okay, I remember this discussion now.  IIRC my suggested solution was to generate ancestors for the mortar elements.  But that would only work for the perfectly aligned interface case, wouldn't it?

So :+1: to merging this.  But don't be shocked if there turn out to be bugs down the line; I don't remember whether these assertions were intended to be *postconditions* for boundary element generators (which would be okay to remove) or were intended to be *preconditions* for other multi-dimensional mesh code (which would mean we've just opened Pandora's box).

PR #2587, opened 2020-05-29
Solution History with adaptive timestepping

vikramvgarg:
Extended the file and memory solution history retrieve methods to work with variable time stepping.
Adjoints example 5 regression expanded to test both uniform and adaptive time stepping.

### DIFF
--- a/examples/adjoints/adjoints_ex5/adjoints_ex5.C
+++ b/examples/adjoints/adjoints_ex5/adjoints_ex5.C
@@ -173,16 +173,8 @@ void write_output(EquationSystems & es,
 #endif
 }
 
-void set_system_parameters(HeatSystem & system,
-                           FEMParameters & param)
+void set_system_parameters(FEMSystem &system, FEMParameters &param)
 {
-  // Use the prescribed FE type
-  system.fe_family() = param.fe_family[0];
-  system.fe_order() = param.fe_order[0];
-
-  // Use analytical jacobians?
-  system.analytic_jacobians() = param.analytic_jacobians;
-
   // Verify analytic jacobians against numerical ones?
   system.verify_analytic_jacobians = param.verify_analytic_jacobians;
   system.numerical_jacobian_h = param.numerical_jacobian_h;
@@ -195,13 +187,20 @@ void set_system_parameters(HeatSystem & system,
   system.print_jacobian_norms    = param.print_jacobian_norms;
   system.print_jacobians         = param.print_jacobians;
   system.print_element_jacobians = param.print_element_jacobians;
-  system.print_element_residuals = param.print_element_residuals;
 
   // Solve this as a time-dependent or steady system
   if (param.transient)
     {
       UnsteadySolver *innersolver;
-      if (param.timesolver_core == "euler")
+      if (param.timesolver_core == "euler2")
+        {
+          Euler2Solver *euler2solver =
+            new Euler2Solver(system);
+
+          euler2solver->theta = param.timesolver_theta;
+          innersolver = euler2solver;
+        }
+      else if (param.timesolver_core == "euler")
         {
           EulerSolver *eulersolver =
             new EulerSolver(system);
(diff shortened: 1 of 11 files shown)


vikramvgarg:
This looks like its ready to merge. @roystgnr

examples/adjoints/adjoints_ex5/heatsystem.h
```diff
@@ -140,6 +138,9 @@ class HeatSystem : public FEMSystem
   std::vector<Number> R_plus_dp;
   std::vector<Number> R_minus_dp;
 
+  // A vector to hold the possibly adaptive deltats
+  std::vector<Number> deltat_vector;
```
roystgnr, line 142: Hmm... this works for now, but in the future we definitely want to have this accessed via some SolutionHistory APIs.  Note that we'd have to change this user code for any adaptive case where we're doing multiple sweeps through time, for example.

vikramvgarg: Yeah, I mentioned in the other PR that we might want to do our adjoint sensitivities in a separate class like we do for ARef and ARes. Elements of the calculation like these can be members of that class, that interact with the timesolver and mutate accordingly.

examples/adjoints/adjoints_ex5/adjoints_ex5.C
```diff
@@ -195,13 +187,20 @@ void set_system_parameters(HeatSystem & system,
...
 
   // Solve this as a time-dependent or steady system
   if (param.transient)
     {
       UnsteadySolver *innersolver;
-      if (param.timesolver_core == "euler")
+      if (param.timesolver_core == "euler2")
+        {
+          Euler2Solver *euler2solver =
+            new Euler2Solver(system);
```
roystgnr, line 198: I'm going to want to change this to avoid the raw heap pointers.

roystgnr: But we were *already* using raw heap pointers for the old eulersolver, so I won't futz with it in this PR.

examples/adjoints/adjoints_ex5/adjoints_ex5.C
```diff
@@ -271,15 +310,15 @@ void set_system_parameters(HeatSystem & system,
     {
 #ifdef LIBMESH_HAVE_PETSC
       PetscDiffSolver *solver = new PetscDiffSolver(system);
-      system.time_solver->diff_solver() = std::unique_ptr<DiffSolver>(solver);
+      system.time_solver->diff_solver() = UniquePtr<DiffSolver>(solver);
```
roystgnr, line 313: What's up here?  Were you working off an older libMesh version?  This undoes a change that got rid of a use of a deprecated shim.

examples/adjoints/adjoints_ex5/adjoints_ex5.C
```diff
@@ -271,15 +310,15 @@ void set_system_parameters(HeatSystem & system,
...
 #else
-      libmesh_error_msg("This example requires libMesh to be compiled with PETSc support.");
+      libmesh_error();
```
roystgnr, line 315: Likewise here - we put nice error messages in on purpose, I swear!  This definitely needs fixing before merge.

examples/adjoints/adjoints_ex5/adjoints_ex5.C
```diff
@@ -173,16 +173,8 @@ void write_output(EquationSystems & es,
...
-                           FEMParameters & param)
+void set_system_parameters(FEMSystem &system, FEMParameters &param)
 {
-  // Use the prescribed FE type
-  system.fe_family() = param.fe_family[0];
-  system.fe_order() = param.fe_order[0];
-
-  // Use analytical jacobians?
-  system.analytic_jacobians() = param.analytic_jacobians;
-
```
roystgnr, line 185: Okay, this whole commit needs a little reworking.  And get into the habit of looking at "git diff" before committing, and use "git add --patch" if it turns out that desired and undesired changes got mixed together some how?

examples/adjoints/adjoints_ex5/adjoints_ex5.C
vikramvgarg: Sorry, yeah, I borrowed some code from an older application that was using adaptive time stepping.

include/solvers/memory_solution_history.h
```diff
@@ -31,6 +32,9 @@
 namespace libMesh
 {
 
+// Forward Declarations
+class DifferentiableSystem;
```
roystgnr, line 36: This is redundant with the above include.

include/solvers/memory_solution_history.h
```diff
@@ -22,6 +22,7 @@
 
 // Local includes
 #include "libmesh/numeric_vector.h"
+#include "libmesh/diff_system.h"
```
roystgnr, line 25: This should be in the .C file, since only the .C file needs it; no need to force every other memory_solution_history.h-using file to include it too.

src/solvers/memory_solution_history.C
```diff
@@ -128,10 +128,47 @@ void MemorySolutionHistory::store(bool /* is_adjoint_solve */)
...
+
+    // Recovering deltats needs two different entries from the the
+    // stored solutions map
+    if(stored_sols_decrement_time != stored_solutions.begin())
+    {
+      stored_sols_decrement_time--;
+
+      Real _decremented_time = stored_sols_decrement_time->first;
+
+      dynamic_cast<DifferentiableSystem &>(_system).deltat = _current_time - _decremented_time;
```
roystgnr, line 153: This breaks MemorySolutionHistory compatibility with any non-diffsystem.  There's no way to actually make adaptive adjoint time stepping work with those right now, but we should at least be catching std::bad_cast (or using a dynamic_cast of pointer and checking for nullptr) and reverting to a uniform-timestepping code path to keep that working as generally as possible.

src/solvers/file_solution_history.C
```diff
@@ -142,13 +142,46 @@ void FileSolutionHistory::retrieve(bool is_adjoint_solve)
...
+
+    // Recovering deltats needs two different entries from the the
+    // stored solutions map
+    if(stored_sols_decrement_time != stored_solutions.begin())
+    {
+      stored_sols_decrement_time--;
+
+      Real _decremented_time = stored_sols_decrement_time->first;
+
+      dynamic_cast<DifferentiableSystem &>(_system).deltat = _current_time - _decremented_time;
```
roystgnr, line 163: Same problem here as in MemorySolutionHistory

include/solvers/file_solution_history.h
```diff
@@ -22,6 +22,7 @@
 
 // Local includes
 #include "libmesh/numeric_vector.h"
+#include "libmesh/diff_system.h"
```
roystgnr, line 25: And here; this header didn't need to change.

src/solvers/adaptive_time_solver.C
```diff
@@ -86,15 +87,38 @@ void AdaptiveTimeSolver::reinit()
...
+
   NumericVector<Number> & old_nonlinear_soln =
     _system.get_vector("_old_nonlinear_solution");
   NumericVector<Number> & nonlinear_solution =
     *(_system.solution);
 
   old_nonlinear_soln = nonlinear_solution;
 
-  if (!first_solve)
-    _system.time += last_deltat;
```
roystgnr, line 97: Wait - how was this ever working before, if we didn't have a ```first_solve = false``` case until just now?

roystgnr: Was it that we were always setting first_solve = false in TwoStepTimeSolver?

vikramvgarg: Yup.

src/solvers/memory_solution_history.C
vikramvgarg: Btw, is system the right owner for deltat, shouldnt the timesolver own it ?

roystgnr: Hmm... IIRC I wanted the physics to be able to integrate without querying the time solver.  But yeah, this was a mistake.  I'd change it right now if I could figure out a nice temporarily-backwards-compatibility-preserving path to do so.

vikramvgarg:
@roystgnr, this should be ready to merge now.

roystgnr:
There's a bit of refactoring that ought to be done in ex5 that I didn't notice earlier, and a couple unintended whitespace changes you missed reverting, but yeah, good enough to merge.  Thanks!

PR #2588, opened 2020-05-29
Add apply_lagrange_dirichlet_impl() optimization

jwpeterson:
This is the third part of the DirichletBoundary refactor
(see #2567, #2574, #2583 for the other parts). This part adds an
optimization for the important special case of specifying BCs for
Lagrange variables.

The basic idea is that Lagrange FEs only have nodal DOFs, so we should
only have to interpolate the boundary function -- projection should
not be required. To implement this, we needed to add some additional
code for shellface BCs which now induce nodal BCs on all the nodes of
the shell element.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -535,6 +535,7 @@ private:
       for (unsigned short shellface=0; shellface != 2; ++shellface)
         {
           boundary_info.shellface_boundary_ids (elem, shellface, ids_vec);
+          bool do_this_shellface = false;
 
           for (const auto & bc_id : ids_vec)
             {
@@ -542,6 +543,7 @@ private:
               if (it != boundary_id_to_ordered_dirichlet_boundaries.end())
                 {
                   has_dirichlet_constraint = true;
+                  do_this_shellface = true;
 
                   // We need to loop over all DirichletBoundary objects associated with bc_id
                   ordered_dbs.insert(it->second.begin(), it->second.end());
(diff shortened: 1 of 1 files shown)


jwpeterson:
Looks like fem_system_ex4 is failing... I will see if I can reproduce the issue locally.

jwpeterson:
Hmm... in dbg mode, I hit an assert:
```
Assertion `nc == n_vec_dim' failed.
nc = 0
n_vec_dim = 1
```

Stack trace:
```
5: libMesh::DofMap::create_dof_constraints(libMesh::MeshBase const&, double)
6: libMesh::System::reinit_constraints()
7: libMesh::System::init_data()
8: libMesh::ImplicitSystem::init_data()
9: libMesh::DifferentiableSystem::init_data()
10: libMesh::FEMSystem::init_data()
11: /home/jwpeterson/projects/libmesh/build/examples/fem_system/fem_system_ex4/.libs/example-dbg(+0x73663) [0x55fe7cd9a663]
12: libMesh::System::init()
13: libMesh::EquationSystems::init()
14: /home/jwpeterson/projects/libmesh/build/examples/fem_system/fem_system_ex4/.libs/example-dbg(+0x4d583) [0x55fe7cd74583]
15: __libc_start_main
16: /home/jwpeterson/projects/libmesh/build/examples/fem_system/fem_system_ex4/.libs/example-dbg(+0x4cbaa) [0x55fe7cd73baa]
[0] ../src/base/dof_map_constraints.C, line 703, compiled May 29 2020 at 12:57:13
```
I believe `nc=0` will happen when there are no DOFs at a node, for example a linear Lagrange FE on a quadratic Elem? In that case I think we just need to `continue` to the next Node as there's no DOF to set... but I'd like to figure out exactly what's going on there.

jwpeterson:
Ha, and that would make sense for fem_system_ex4, looks like it builds a HEX27/QUAD9 mesh and then instantiates linear FEs on it.

jwpeterson:
This seems to be working with most of my tests so I'm going to go ahead and merge. The performance payoff wasn't nearly as large as the loop reordering optimization, but this change is still beneficial because it will allow us to apply Dirichlet BCs on Lagrange elements without reinit'ing on Nodes.

roystgnr:
I thought it was a miracle when the MOOSE tests all passed, but I just spent half a day (wait, it's 11pm? maybe more than half...) discovering that this PR triggered 3 GRINS regression test failures.  Exactly the sort we'd expect to fail (discontinous Dirichlet BCs being projected-but-now-interpolated into the FE space), and the solution diffs aren't huge, and the results look fine in the eyeball norm, so I'm just going to regold and not worry about it ... but we'd better put a big screaming "this may change your solutions" warning into the 1.6.0 release NEWS, at the very least.

PR #2589, opened 2020-05-29
Use double curly braces constructing std::array

lindsayad:
Help out old clang compilers for which -Werror -Wall triggers an
error, see https://civet.inl.gov/job/535324/


lindsayad:
This only appears to be triggered on CIVET when building MOOSE because that's the place where we   pick in `ADDITIONAL_CPPFLAGS='-Wall -Wextra -Werror'` in the `Makefile`

jwpeterson:
Thanks, I was wondering if we still had any compilers that needed the double curly brace syntax... guess this answers that question.

PR #2590, opened 2020-06-01
Fix distributed periodic constraints bug

roystgnr:
I noticed some time ago that there was a regression in periodic boundary conditions on distributed mesh on some weird corner case (I could run libMesh "make check" on up to 20 procs, before triggering an assertion failure in one example with 21 procs), but I've only just now managed to track down the problem - when I reduced the amount of default ghosting libMesh does (to include side neighbors but not all point neighbors) I missed one bit of code that was still occasionally depending on 
seeing point neighbors.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -1703,6 +1703,7 @@ void DofMap::add_constraint_row (const dof_id_type dof_number,
     if (this->is_constrained_dof(dof_number))
       libmesh_error_msg("ERROR: DOF " << dof_number << " was already constrained!");
 
+  // We don't allow nonsensical constraints
   libmesh_assert_less(dof_number, this->n_dofs());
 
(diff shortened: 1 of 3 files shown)


roystgnr:
I can reproduce the failure here; it's definitely a real regression, if not due to this patch then triggered by this patch.

roystgnr:
Ouch.  The regression appears to be merely *triggered* by this patch; the old code could hypothetically end up failing on corners with multiple periodic boundaries as well, if the element ids ended up being assigned in a particularly perverse way; we've just been lucky enough (hopefully!?!) not to see that.

In hindsight this probably should have been a fix or at least an assertion rather than a comment...
```
                  // FIXME: This code doesn't yet properly handle
                  // cases where multiple different periodic BCs
                  // intersect.
```

jwpeterson:

PR #2591, opened 2020-06-03
apply_dirichlet_impl(): use p-refined FEType

jwpeterson:
This is the same type of fix I had to make in 2afcb0f94, but for the
old apply_dirichlet_impl() code path. It only affects apps which are
imposing Dirichlet BCs on non-Lagrange p-refined elems.

Refs #2588.


jwpeterson:
Thanks!

PR #2592, opened 2020-06-03
Non-proxy sides should have same p_level() as Elem they are built from.

jwpeterson:
Proxy sides built through the Side<X,Y> helper class should already
report a matching p_level() since they "inherit" it from the class
which creates them.


jwpeterson:
This change is in preparation for hopefully deprecating some `FEInterface` calls which currently take an `ElemType` rather than an `Elem*`, and therefore rely on the caller to account for `Elem::p_level()` correctly, which can be tricky to do correctly.

jwpeterson:
Thanks!

PR #2596, opened 2020-06-04
Fix value return in JIT function

### LINKED ISSUE #2595: Fix JIT compile error
As noted in
https://hpcgitlab.hpc.inl.gov/idaholab/bison/-/merge_requests/2048#note_277895
the use of the `if` conditional in parsed ADReal functions will lead to a JIT compile error. The generated code contains a return statement with the wrong type (ADReal instead of void). I must have missed that when converting the JIT functions from returning values to taking a writable reference for the computed value. This was necessary to permit returning non-intrinsic types.

dschwen:
Closes #2595

### DIFF
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -966,7 +966,7 @@ bool FunctionParserADBase<Value_t>::JITCodeGen(std::ostream & ccout, const std::
           ccout << "if (s[" << sp-- << "] < 0.5) ";
 
         if (ip >= ByteCode.size())
-          ccout << "return s[" << sp << "];\n";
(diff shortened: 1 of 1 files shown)


jwpeterson:
More JITTY code

dschwen:
Are you JITTING you pants over it?

jwpeterson:
I've run out of JITs to give.

dschwen:
If this doesn't fix it I'm getting JITFACED!

dschwen:
A ... baJITlion... ?

...

... I'll see myself out.

jwpeterson:
> How many libmesh updates can we do in a month?

A JITload

PR #2597, opened 2020-06-04
Add new InfFE::n_dofs() overload, deprecate old version

jwpeterson:
This is similar to the refactoring in #2593, but for `InfFEs`. This makes it so that `InfFE::n_dofs()` and related helper functions now take an `Elem*` instead of an `ElemType`, and deprecates previous versions. This at least compiles for me, but it would be good if people that actually use `InfFEs`, like @BalticPinguin, can also take a look.

### DIFF
--- a/src/fe/fe_interface.C
+++ b/src/fe/fe_interface.C
@@ -479,8 +479,7 @@ unsigned int FEInterface::n_dofs(const unsigned int dim,
                                  const FEType & fe_t,
                                  const ElemType t)
 {
-  // TODO:
-  // libmesh_deprecated();
+  libmesh_deprecated();
 
 #ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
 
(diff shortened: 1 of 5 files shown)


src/fe/fe_interface.C
```diff
@@ -545,9 +543,9 @@ FEInterface::n_dofs(const FEType & fe_t,
 
 #ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
 
-  // FIXME: Can/should InfElems account for p_level()?
+  // InfElems currently don't support p_level()
   if (is_InfFE_elem(elem->type()))
-    return ifem_n_dofs(dim, fe_t, elem->type());
+    return ifem_n_dofs(fe_t, elem);
```
BalticPinguin, line 548: Since mesh refinement currently is not working with infinite elements at all (#1149), one probably doesn't have to care about this here?
Also, there are two separate polynomial orders: the ``base`` order and ``radial`` order; an implementation of p-refinement thus might take care of this?

jwpeterson: Yeah, this PR is mainly about having a consistent `FEInterface::n_dofs()` API, so the fact that InfFEs don't really support p-refinement at the moment isn't critical.

PR #2598, opened 2020-06-05
Fix clang warnings

lindsayad:
Clang correctly detects that these are operators are implicitly deleted
because the compiler has no idea what we want to do with the reference
members of the class. So we should either remove these defaulted operators
or write our own assignment operators if that's what we want.

Also fix an unused lambda capture warning

### DIFF
--- a/include/mesh/exodusII_io_helper.h
+++ b/include/mesh/exodusII_io_helper.h
@@ -95,12 +95,19 @@ public:
   /**
    * Special functions. This class does not manage any dynamically
    * allocated resources (file pointers, etc.) so it _should_ be
-   * default copy/move constructable and assignable, but I don't know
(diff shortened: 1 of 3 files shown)


jwpeterson:
Can we explicitly `=delete` the copy/move assignment operators instead of removing those lines? One of the main points behind explicitly defaulting/deleting these functions is that you don't have to remember/guess which ones the compiler generates for you and which ones you can't call.

jwpeterson:
I'm also a bit surprised that GCC doesn't warn about this...

jwpeterson:
Would also be good if we could put a comment regarding why copy/move assignment is deleted... actually I can take care of all these small changes if you prefer.

lindsayad:
Yea, I'll let you do it because I don't want to recompile :-)

PR #2600, opened 2020-06-08
Deprecate FEInterface::n_dofs_per_elem(), add version taking an Elem*

jwpeterson:
Related to work in #2597, #2599.


jwpeterson:
The bugs in commits ca534a3 and c169a52 were introduced by an earlier PR (#2593) but were not caught at the time, I think because of a configuration that is not tested. If this passes, I will make a separate bugfix PR from just those commits, and then rebase this one on that...

PR #2601, opened 2020-06-08
Add new versions of FEInterface::n_shape_functions() taking an Elem *

jwpeterson:
Similar to the changes in #2600, #2599, #2593.

### DIFF
--- a/src/fe/fe_bernstein.C
+++ b/src/fe/fe_bernstein.C
@@ -51,7 +51,8 @@ void bernstein_nodal_soln(const Elem * elem,
   const Order totalorder = static_cast<Order>(order + elem->p_level());
 
   // FEType object to be passed to various FEInterface functions below.
-  FEType fe_type(totalorder, BERNSTEIN);
+  FEType fe_type(order, BERNSTEIN);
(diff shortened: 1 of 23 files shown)


src/fe/fe_bernstein.C
```diff
@@ -97,8 +96,7 @@ void bernstein_nodal_soln(const Elem * elem,
             // u_i = Sum (alpha_i phi_i)
             for (unsigned int i=0; i<n_sf; i++)
               nodal_soln[n] += elem_soln[i] *
-                // FE<Dim,T>::shape(elem, order, i, mapped_point);
-                FEInterface::shape(Dim, fe_type, elem, i, refspace_nodes[n]);
+                FEInterface::shape(Dim, p_refined_fe_type, elem, i, refspace_nodes[n]);
```
roystgnr, line 99: The switch from fe_type to p_refined_fe_type here - was that a bug in the previous code?

jwpeterson: No, that's a temporary workaround until I have a chance to update `FEInterface::shape()` so that it handles Elem::p_level() internally. That should be my next PR.

jwpeterson:
Thanks for reviewing!

PR #2603, opened 2020-06-08
Add version of FEInterface::shape() which accounts for Elem::p_level() internally

jwpeterson:
There are two existing `FEInterface::shape()` implementations which differ in what they expect the input `FEType`'s order to be (total order for `ElemType` version, base order for `Elem*` version). This is a bit confusing/easy to forget, so these functions are now deprecated and instead one should call one of:
```
static Real shape(const FEType & fe_t,
                  const Elem * elem,
                  const unsigned int i,
                  const Point & p);

static Real shape(const FEType & fe_t,
                  int extra_order,
                  const Elem * elem,
                  const unsigned int i,
                  const Point & p);
```
The former accounts for `Elem::p_level()` internally automatically, the latter ignores `Elem::p_level()` and uses the specified `extra_order` instead. The idea is that you will only call this latter version if you know what you are doing, a majority of users will simply call the first version.

### DIFF
--- a/src/fe/fe_monomial.C
+++ b/src/fe/fe_monomial.C
@@ -32,8 +32,7 @@ namespace {
 void monomial_nodal_soln(const Elem * elem,
                          const Order order,
                          const std::vector<Number> & elem_soln,
-                         std::vector<Number> &       nodal_soln,
-                         const unsigned Dim)
+                         std::vector<Number> & nodal_soln)
 {
   const unsigned int n_nodes = elem->n_nodes();
 
@@ -65,7 +64,6 @@ void monomial_nodal_soln(const Elem * elem,
       {
         // FEType object to be passed to various FEInterface functions below.
         FEType fe_type(order, MONOMIAL);
-        FEType p_refined_fe_type(totalorder, MONOMIAL);
 
         const unsigned int n_sf =
           FEInterface::n_shape_functions(fe_type, elem);
@@ -84,7 +82,7 @@ void monomial_nodal_soln(const Elem * elem,
             // u_i = Sum (alpha_i phi_i)
             for (unsigned int i=0; i<n_sf; i++)
               nodal_soln[n] += elem_soln[i] *
-                FEInterface::shape(Dim, p_refined_fe_type, elem, i, refspace_nodes[n]);
+                FEInterface::shape(fe_type, elem, i, refspace_nodes[n]);
           }
 
         return;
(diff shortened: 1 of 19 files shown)


src/fe/fe_monomial.C
```diff
@@ -84,7 +82,7 @@ void monomial_nodal_soln(const Elem * elem,
             // u_i = Sum (alpha_i phi_i)
             for (unsigned int i=0; i<n_sf; i++)
               nodal_soln[n] += elem_soln[i] *
-                FEInterface::shape(Dim, p_refined_fe_type, elem, i, refspace_nodes[n]);
+                FEInterface::shape(fe_type, elem, i, refspace_nodes[n]);
```
jwpeterson, line 85: @roystgnr in the course of updating this (and the other FEs) `nodal_soln()` code, I realized that this `FEInerface::shape` call was being made with *both* a "p-refined" `FEType` (i.e. based on `totalorder`) *and* it was being passed an `Elem` pointer. In fe_interface.C, this calls:
```
const Order o = fe_t.order; // this is already totalorder
fe_switch(shape(elem,o,i,p));
```
and that `fe_switch` macros then calls the 4-parameter FE::shape() overload (see fe.h)
```
static OutputShape shape(const Elem * elem,
                         const Order o,
                         const unsigned int i,
                         const Point & p,
                         const bool add_p_level = true);
```
Note that the default value of `add_p_level` is true, so this function is being called with an already-inflated `Order`, and then accounting for the `p_level` *again* internally, for example the 1D monomial instantiation is:
```
template <>
Real FE<1,MONOMIAL>::shape(const Elem * elem,
                           const Order order,
                           const unsigned int i,
                           const Point & p,
                           const bool add_p_level)
{
  libmesh_assert(elem);

  return FE<1,MONOMIAL>::shape(elem->type(), static_cast<Order>(order + add_p_level * elem->p_level()), i, p);
}
```
This definitely seems like a bug to me, but apparently we've never encountered it, possibly because the `n_sf` is still correct and the lower-p DOFs are "nested" within the higher-p DOFs? Or maybe my interpretation of what's happening is just wrong. Either way, it should be fixed in this refactoring.

roystgnr: We might not have any p_level() refinement being tested with MONOMIAL elements.  IIRC our test coverage is all for H1 problems with continuous Galerkin.  But you're also right that even solid regression tests (or application code, thankfully) wouldn't have triggered the bug here, since the MONOMIAL shape functions are hierarchic ("nested") with respect to p.  Good catch!

PR #2605, opened 2020-06-09
Centroid partitioner stable sort

drwells:


jwpeterson:
Certainly seems reasonable to me. We do have a unit test of `CentroidPartitioner`, but from scanning it I don't think this change should have any effect on it... I think there are also some `CentroidPartitioner` tests in MOOSE so we'll see if any of those diff.

drwells:
is there a place that we should record incompatible changes?

jwpeterson:
> is there a place that we should record incompatible changes?

Not that I'm aware of. Since this isn't an API change and it fixes buggy/non-reproducible behavior it should be merged even if tests diff IMO. We'll just need to temporarily disable/regold them so that the CI is happy. Kinda looks like all is going to pass though...

roystgnr:
The GitHub reaction emoji options need to add :crossed_fingers:

PR #2606, opened 2020-06-10
More FEInterface refactoring

jwpeterson:
I added a few `\todos` on FEInterface APIs we should eventually update to do away with the unnecessary `dim` arg, but didn't go for a full-blown deprecation of these since they did not otherwise have confusing/easy to misuse `p_level()` behavior. My preference would be to simply update those APIs in a backwards-incompatible way at some point, possibly after a new release is tagged.

Still remaining: extend the refactoring to cover the related `InfFE` APIs.

### DIFF
--- a/include/fe/fe_interface.h
+++ b/include/fe/fe_interface.h
@@ -492,7 +492,6 @@ public:
   shape_function(const FEType & fe_t,
                  const Elem * elem);
 
-
   /**
    * \returns The \f$ j^{th} \f$ coordinate of the gradient of
    * the \f$ i^{th} \f$ shape function at point \p p.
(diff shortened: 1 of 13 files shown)


roystgnr:
I'm thrilled with the pace of these refactoring PRs, but I'm less thrilled that I'm now hitting a failed assertion in DofMap::old_dof_indices(), from adaptivity_ex3, in a 2 processor DistributedMesh 1D hp refinement run.  I haven't yet tracked down exactly what the regression is or figured out why it wouldn't be affecting our CI runs, though.

jwpeterson:
What command line arguments for adaptivity_ex3 exactly? It should be possible to bisect... all the intermediate refactoring commits compile and are pretty small. Looks like old_dof_indices was changed most recently in the following commits:
6f58aa759573ca9a361f7b6cd64b86ae2f2a7e32
7a96b518068e629992af112014badaf32605a2c4
137f630942e39f17da49e15653e3664bb6811ca6
2f6a409f866e79b00a76cd016a99ec26293dbd54

jwpeterson:
Thanks!

PR #2607, opened 2020-06-10
build_node_list_from_side_list via parallel_sync

roystgnr:
TIMPI vector<vector> options are flexible enough to do this now, and it
should be much faster, and @fdkong is working at processor counts where
the gross inadequacy of the previous code was becoming an issue.

### DIFF
--- a/src/mesh/boundary_info.C
+++ b/src/mesh/boundary_info.C
@@ -17,9 +17,6 @@
 
 
 
-// C++ includes
-#include <iterator>  // std::distance
-
 // Local includes
 #include "libmesh/libmesh_config.h"
 #include "libmesh/libmesh_logging.h"
@@ -32,6 +29,12 @@
 #include "libmesh/partitioner.h"
 #include "libmesh/remote_elem.h"
 #include "libmesh/unstructured_mesh.h"
+
+// TIMPI includes
+#include "timpi/parallel_sync.h"
+
+// C++ includes
+#include <iterator>  // std::distance
 
 namespace
 {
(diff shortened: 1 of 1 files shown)


roystgnr:
The code is much cleaner now and could hardly get any slower, so this ought to be okay to merge if and when CI verifies that it's *correct*, but I'd love it if @fdkong manages to also verify that it solves his build_node_list_from_side_list performance problems.

fdkong:
It is what I got with this PR

profile006

jwpeterson:
That name mangling though :eyes:

roystgnr:
Looks like we're down to 50s, versus 56-60 in some earlier profiles you showed me?  So this is an improvement and we should merge, but it's not nearly as much as I hoped it would be.

I do think I see a nasty performance issue in the new push(vec<vec>) code that's sucking up most of the improvement we'd hoped to see from that.  If I'm right then one more PR ought to give us a pretty big win...

roystgnr:
https://github.com/libMesh/TIMPI/pull/38 is my crack at that performance issue; if you could run the same 12,288-core problem for an apples-to-apples comparison, that would be ideal.

fdkong:
> [libMesh/TIMPI#38](https://github.com/libMesh/TIMPI/pull/38) is my crack at that performance issue; if you could run the same 12,288-core problem for an apples-to-apples comparison, that would be ideal.

I will do that, give me a bit of time.

roystgnr:
Sure; not trying to rush you!  Just wanted to point out exactly which run of yours I'm hoping I fixed this time; there are some other optimization ideas still in the pipeline.

PR #2610, opened 2020-06-11
Side objects should set interior_parent(), not parent()

roystgnr:
For me this fixes the regression we were talking about in https://github.com/idaholab/moose/pull/15389 , which turned out to be caused by the interaction of parent() pointer settings on side objects with the ancestor-editing code in set_p_level(), as triggered by https://github.com/libMesh/libmesh/pull/2592

This PR is the right long-term fix: to use interior_parent() instead of parent() for that link on side elements, just as we've done for years on boundary mesh elements.  Hopefully that doesn't break compatibility for user codes; in the library itself there were only two or three places checking parent().  If there is any breakage then we'll need to work out a more tedious transition plan...

### DIFF
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -2308,7 +2308,10 @@ Elem::simple_build_side_ptr (const unsigned int i,
         face->set_node(n) = this->node_ptr(Subclass::side_nodes_map[i][n]);
     }
 
-  face->set_parent(nullptr);
+#ifdef LIBMESH_ENABLE_DEPRECATED
+  if (!proxy) // proxy sides used to leave parent() set
+#endif
+    face->set_parent(nullptr);
   face->set_interior_parent(this);
 
   return face;
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -69,6 +69,8 @@ class Elem;
 class PeriodicBoundaries;
 class PointLocatorBase;
 #endif
+template <class SideType, class ParentType>
+class Side;
 
 
 /**
@@ -1716,6 +1718,14 @@ protected:
   /**
    * An implementation for simple (all sides equal) elements
    */
+  template <typename Sideclass, typename Subclass>
+  std::unique_ptr<Elem>
+  simple_build_side_ptr(const unsigned int i,
+                        bool proxy);
+
+  /**
+   * An implementation for simple (all sides equal) elements
+   */
   template <typename Subclass>
   void simple_build_side_ptr(std::unique_ptr<Elem> & side,
                              const unsigned int i,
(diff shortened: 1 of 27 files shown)


roystgnr:
> If there is any breakage then we'll need to work out a more tedious transition plan...

Those failing tests look like plausible ```side->parent()``` users, so I guess plan B it is.

roystgnr:
To be specific: "Plan B" will be to set *both* parent and interior_parent on sides (making sure any p_level() changes don't trickle up to the "parent").  I'm afraid there won't be any way to mark side->parent() calls as ```libmesh_deprecated()```, but we can at least set parent to nullptr in the --disable-deprecated case.

include/geom/elem.h
```diff
@@ -2280,6 +2290,32 @@ Elem::build_side_ptr (std::unique_ptr<const Elem> & elem,
 
 
 
+template <typename Sideclass, typename Subclass>
+inline
+std::unique_ptr<Elem>
+Elem::simple_build_side_ptr (const unsigned int i,
```
jwpeterson, line 2296: This doesn't set the subdomain id? The functions you are replacing all set the subdomain id of the side... maybe something is relying on this behavior?

src/fe/fe_boundary.C
```diff
@@ -639,7 +639,7 @@ void FEMap::compute_face_map(int dim, const std::vector<Real> & qw,
 
         // We need to look back at the full edge to figure out the normal
         // vector
-        const Elem * elem = side->parent();
+        const Elem * elem = side->interior_parent();
```
jwpeterson, line 642: So... "side->parent" with "side->interior_parent" throughout? I think we're pretty consistent about naming sides "side" but just curious if you also searched for all possible `foo->parent()` calls as well?

include/geom/elem.h
jwpeterson: Ah, I'm guessing sides now get their subdomain_id from their interior_parent... if that's the case, nevermind!

roystgnr: They get their subdomain_id via the parent argument to their constructor, which we later turn into their interior_parent, yeah.  I admit I'd originally forgotten about that behavior too - I was all ready to fix a "bug" in the optimized build_side_ptr overload where it was explicitly resetting subdomain_id when modifying a side_ptr but seemed to be failing to do so when first constructing one.

src/fe/fe_boundary.C
roystgnr: Not a text replace, rather a manual search, but oh, no, I didn't search for *every* ```->parent()``` call, I'm afraid.  But we'll leave parent() set in default non --disable-deprecated builds for a couple years, and hopefully people will play with --disable-deprecated frequently enough in development to catch any incompatibilities before deployment.

I wish I could figure out how to make those incompatibilities into nice clean libmesh_deprecated() errors, but at least a nullptr dereference most often leads to a segfault plus stack trace rather than a corrupt solution.

src/geom/cell_hex20.C
```diff
@@ -184,23 +184,7 @@ Order Hex20::default_order() const
...
                                              bool proxy )
 {
-  libmesh_assert_less (i, this->n_sides());
-
-  if (proxy)
-    return libmesh_make_unique<Side<Quad8,Hex20>>(this,i);
-
-  else
-    {
-      std::unique_ptr<Elem> face = libmesh_make_unique<Quad8>();
```
roystgnr, line 194: So, we've only been setting parent() on proxy sides, not on non-proxy sides.  I guess that's why my backwards compatibility commit was overkill?  I still can't replicate the MOOSE test failures, they keep passing for me, but I can certainly try being more backwards compatible in my backwards compatibility.

roystgnr:
Ok, I'm an idiot.  I was trying to reproduce these failures with my $LIBMESH_DIR set to a different install directory.  This should be much easier to debug once I stop trying to do so blind.

PR #2611, opened 2020-06-12
Use new non-deprecated n_dofs overload

roystgnr:
This fixes an error for me in our unit tests for --disable-deprecated builds.


jwpeterson:
Merging right away because I just ran into this locally in my `--disable-deprecated` build as well.

PR #2612, opened 2020-06-12
Reduced Basis change: Optionally pre-evaluate theta functions

dknez:
For the sake of efficiency we can pre-evaluate the theta functions and store the results. This can be helpful for the RB greedy algorithm since in that case we re-evaluate many times for the same set of training parameters.


dknez:
Test debug gave an error earlier, but I couldn't see the error. Any idea what the issue is?

Issue #2613, opened 2020-06-16
Disable -Wimplicit-fallthrough warnings in FParser code


jwpeterson:
If I recall correctly, this is a GCC-only warning, but when compiling FParser code I see hundreds of lines of warnings:
```
./extrasrc/fp_opcode_add.inc:10662:2: warning: this statement may fall through [-Wimplicit-fallthrough=]
  if(!ByteCodePtr)
  ^~
```
I don't propose trying to actually fix these, but maybe they could be disabled with compiler pragmas.

roystgnr:
This is one of the ones our ignore_warnings.h disables, and the placement of the pragma there says it works for GCC 6 or later.

Of course that doesn't help fparser, which doesn't include any libMesh headers, but we could reasonably make a copy (ideally a stripped down copy, since ignore_warnings.h disables *any* warning that any third-party headers have ever given us) to use there.

jwpeterson:
I'll try just copying the pragma and compiler version checks from ignore_warnings.h

jwpeterson:
Fixed in #2617

PR #2617, opened 2020-06-17
Disable warnings in fparser builds

jwpeterson:
This gets rid of the warnings described in #2613, plus the 
```
the ABI of passing structure with complex float member has changed in GCC 4.4
```
one that has been showing up for a while now.

### DIFF
--- a/contrib/fparser/Makefile.orig
+++ b/contrib/fparser/Makefile.orig
@@ -95,6 +95,7 @@ CPPFLAGS=$(FEATURE_FLAGS) -I$(LIBMESH_DIR)/include/libmesh
 CXXFLAGS=-Wall -W -pedantic -ansi $(OPTIMIZATION)
 #CXXFLAGS += -Wunreachable-code
 #CXXFLAGS += -std=c++0x
+CXXFLAGS += -std=c++11
 
 #CXXFLAGS += -Weffc++
 
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -22,6 +22,16 @@ using namespace FPoptimizer_CodeTree;
 
 #include "lib/sha1.h"
 
+// There are several case statements in this file where we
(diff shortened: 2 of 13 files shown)


contrib/fparser/Makefile.orig
```diff
@@ -95,6 +95,7 @@ CPPFLAGS=$(FEATURE_FLAGS) -I$(LIBMESH_DIR)/include/libmesh
 CXXFLAGS=-Wall -W -pedantic -ansi $(OPTIMIZATION)
 #CXXFLAGS += -Wunreachable-code
 #CXXFLAGS += -std=c++0x
+CXXFLAGS += -std=c++11
```
roystgnr, line 98: Does this makefile actually still get used?  I thought this was just what we renamed their Makefile before we put our own Makefile.am in place.

contrib/fparser/fparser_ad.cc
```diff
@@ -22,6 +22,16 @@ using namespace FPoptimizer_CodeTree;
...
+// There are several case statements in this file where we
+// intentionally want to fall through, so let's not get warned about
+// each one.
+#ifdef __GNUC__
+#if (__GNUC__ > 6)
+#  pragma GCC diagnostic ignored "-Wimplicit-fallthrough"
+#endif
+#endif
+
+
```
roystgnr, line 34: Why not just include the new header here (and in the other half dozen .cc files)?  Trying to keep the other two warnings active?

contrib/fparser/Makefile.orig
jwpeterson: Yes, I had to use it to regenerate fpoptimizer.cc

contrib/fparser/fparser_ad.cc
jwpeterson: Yeah, principle of least change I guess. Since we aren't currently getting those other warnings from this file, I'd rather not disable them yet.

PR #2620, opened 2020-06-22
Edge::side_ptr(): remove inadvertant set_p_level() call

jwpeterson:
This change was made by accident in f1782255be while I was trying to
update the various Elem::build_side_ptr() routines in #2592, so this
simply reverts that change.

There appears to be larger issues with the consistency of the
Elem::side_ptr() method and setting of the side's p_level, but this
change can at least be reverted before those are addressed.


roystgnr:
Thanks!

PR #2621, opened 2020-06-23
Use Mesh::add_point to ensure new Node has a valid ID

dknez:


roystgnr:
This is good, but @jwpeterson points out that the underlying flaw is a bug in ReplicatedMesh::add_node(); we should fix that too in another PR.

jwpeterson:
@roystgnr I don't get why this compiles (maybe it is in a config we don't test?). We need to pass a `Point` to the `add_point()` routine but three values are passed here. I'm worried the second and third args are being converted to id and proc_id somehow.

jwpeterson:
> (maybe it is in a config we don't test?)

Yeah, this is inside an `#if defined(LIBMESH_HAVE_CAPNPROTO)` section, and configure says:
```
<<< Required header files not found, Cap'n Proto support disabled. >>>
```

roystgnr:
Whoa, that'll teach me to read diffs more closely.

> I'm worried the second and third args are being converted to id and proc_id somehow.

That'd have been caught via ```-Werror -Wconversion``` in our ```--enable-werror --enable-paranoid-warnings``` CI builds - if only at least one of those hadn't ifdef'ed out the broken code.

jwpeterson:
I will push a fix for this, I have a capnp build locally.

dknez:
Oops, thanks!

PR #2626, opened 2020-06-25
Fix concurrency bug in "make test_headers"

roystgnr:
I noticed this was broken while swiping the code for TIMPI in
https://github.com/libMesh/TIMPI/pull/37 - we should fix it here too.

In 2014 Ben added the capability to test 20 headers at a time; what nobody realized before is that we would then *skip the 21st*.


jwpeterson:
Thanks! I guess the headers we've been skipping testing for all this time were OK, so that's good.

PR #2627, opened 2020-06-26
Improvements to SolutionHistory API

vikramvgarg:
1) SolutionHistory operation functions now take time as an argument, no more assuming sequential operations.
2) SolutionHistory data structure changed from list to map. This will help with adaptive and multi step time stepping schemes.
3) SolutionHistory::erase(time) added, which will streamline discarding saved time steps when an adaptive time stepping scheme is used.

### DIFF
--- a/src/solvers/file_solution_history.C
+++ b/src/solvers/file_solution_history.C
@@ -75,7 +75,7 @@ void FileSolutionHistory::store(bool is_adjoint_solve, Real time)
   // In an empty history we create the first entry
   if (stored_solutions.begin() == stored_solutions.end())
     {
-      stored_solutions.emplace_back( time, std::string()  );
+      stored_solutions[time] = std::string();
       stored_sols = stored_solutions.begin();
     }
 
@@ -86,7 +86,7 @@ void FileSolutionHistory::store(bool is_adjoint_solve, Real time)
       ++stored_sols;
       libmesh_assert (stored_sols == stored_solutions.end());
 #endif
-      stored_solutions.emplace_back( time, std::string() );
+      stored_solutions[time] = std::string();
       stored_sols = stored_solutions.end();
       --stored_sols;
     }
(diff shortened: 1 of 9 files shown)


src/solvers/file_solution_history.C
```diff
@@ -33,81 +33,73 @@ FileSolutionHistory::~FileSolutionHistory ()
...
-    }
+  // Set the stored sols iterator to whichever key is within TOLERANCE of time
+  if(std::abs(lower_bound_it->first - time) < TOLERANCE)
+  {
+    stored_sols = stored_solutions.find(time);
+  }
+  else if(std::abs((--lower_bound_it)->first - time) < TOLERANCE)
+  {
+    stored_sols = stored_solutions.find(time);
+  }
```
roystgnr, line 57: Should we scream and die here if we don't find anything?  At least a libmesh_assert()?

src/solvers/memory_solution_history.C
```diff
@@ -31,74 +31,66 @@ MemorySolutionHistory::~MemorySolutionHistory ()
...
-    }
+  // Set the stored sols iterator to whichever key is within TOLERANCE of time
+  if(std::abs(lower_bound_it->first - time) < TOLERANCE)
+  {
+    stored_sols = stored_solutions.find(time);
+  }
+  else if(std::abs((--lower_bound_it)->first - time) < TOLERANCE)
+  {
+    stored_sols = stored_solutions.find(time);
+  }
```
roystgnr, line 55: Likewise here

roystgnr:
Yeah, this looks much cleaner.  I'll merge as-is, since I think the adjoints_ex5 gold value tests ought to be enough to ensure we're not messing anything up, but we should probably slip those extra asserts into subsequent PRs.

src/solvers/file_solution_history.C
jwpeterson: Is it not possible for `--lower_bound_it` to step off the front of the array?

roystgnr: Oh, wait a minute, this is just wrong.  If find() was going to work, we could have just used it in the first place!  The point of my lower_bound suggestion was that you can then be sure of finding the nearest two entries, and that you can then *use* one of those entries.  If you don't actually use the entry then if we have even a hint of floating point error stored_sols will get set to stored_solutions.end() and we'll either get an assertion failure at a completely different part of the code (e.g. from the assertion above here at the *next* step) or we'll get a segfault or a garbage value when end() gets dereferenced.

We don't need to revert this PR since the bug didn't happen to trigger anything in CI, but we should get this fixed ASAP.

PR #2631, opened 2020-06-30
Remove extra flags from pkgconfig Cflags.

drwells:
Followup to #2604.

The convention for this field is to only list necessary (i.e., preprocessor) flags. I moved the other compilation flags into cxxflags_extra, which is analogous to where PETSc puts them.


roystgnr:
Thanks!  I don't think anyone had given our pkgconfig support a good hard look since 2012...

drwells:
:+1:

PR #2632, opened 2020-06-30
Generalize DifferentiableSystem::adjoint_solve

vikramvgarg:
1) Adjoint solve methods can now be overloaded by TimeSolver subclasses, will mirror the forward timestepping scheme.
2) Fixed handing of corner cases generated by use of map::lower_bound in SoiutionHistory subclasses.

### DIFF
--- a/src/solvers/file_solution_history.C
+++ b/src/solvers/file_solution_history.C
@@ -32,7 +32,8 @@ FileSolutionHistory::~FileSolutionHistory ()
 }
 
 // This function finds, if it can, the entry where we're supposed to
-// be storing data
+// be storing data, leaves stored_sols unchanged if it cant find an entry
+// with the key corresponding to time.
 void FileSolutionHistory::find_stored_entry(Real time)
 {
   if (stored_solutions.begin() == stored_solutions.end())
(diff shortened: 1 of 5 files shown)


src/solvers/file_solution_history.C
```diff
@@ -32,29 +32,57 @@ FileSolutionHistory::~FileSolutionHistory ()
...
+    stored_sols = stored_solutions.begin();
+    return;
+  }
+  else // A decremented iterator, to perform the sandwich test for the key closest to time
+  {
+    lower_bound_it_decremented = std::prev(lower_bound_it);
+  }
+
+  // Set the stored sols iterator as per the key which is within TOLERANCE of time
   if(std::abs(lower_bound_it->first - time) < TOLERANCE)
```
roystgnr, line 73: Man I hate having to use an absolute tolerance here.  We'd like for the same code to work with transient EM propagation and with cosmology, if possible, spanning orders of magnitude.  Is there anything sane we can normalize by here?  The difference between lower_bound_it->first and lower_bound_it_decremented->first, maybe?

src/solvers/file_solution_history.C
```diff
@@ -32,29 +32,57 @@ FileSolutionHistory::~FileSolutionHistory ()
...
+  // begin -> we are looking for the solution at the initial time
+  // end -> we are creating a new entry
+  // anything else, we are looking for an existing entry
   stored_solutions_iterator lower_bound_it = stored_solutions.lower_bound(time);
 
-  // Set the stored sols iterator to whichever key is within TOLERANCE of time
+  // For the key right before the lower bound
+  stored_solutions_iterator lower_bound_it_decremented;
+
+  // If we are at end, we are creating a new entry, nothing more to do
```
roystgnr, line 57: Is this correct?  If our last stored_solutions time is t, and we ask for t-epsilon^2, then the lower_bound is going to return end, even though we surely want t.

roystgnr:
If you have time to consider these issues soon I'd love to have the improvements tacked onto this PR, however the code here is still a clear improvement on the *previous* code so we can merge any time.

src/solvers/file_solution_history.C
vikramvgarg: We could scale and map the interval (lower_bound_it_decremented->first, lower_bound_it->first) to (0, 1) and require that the scaled time be within TOLERANCE of one of the endpoints ?

src/solvers/file_solution_history.C
vikramvgarg: In that case, we need to add an if check in SH::retrieve, which moves stored_sols to the entry before end, if the call find_stored_entry in retrieve ends up pointing stored_sols to the end.

src/solvers/file_solution_history.C
roystgnr: That'd effectively be what I was thinking, yes (although multiplying tolerance by the interval length rather than dividing the time difference).

src/solvers/file_solution_history.C
roystgnr: Could you explain?  Right now, if we find_stored_history(final_time-epsilon^2), it would fail to find it.  With my suggestion, it would find it.  Why would the latter result require an extra if check that the former would not?

src/solvers/file_solution_history.C
```diff
@@ -32,29 +32,57 @@ FileSolutionHistory::~FileSolutionHistory ()
...
+    return;
+  }
+  else if(lower_bound_it == stored_solutions.begin()) // At the beginning, so we cant go back any further
+  {
+    stored_sols = stored_solutions.begin();
+    return;
+  }
+  else // A decremented iterator, to perform the sandwich test for the key closest to time
+  {
+    lower_bound_it_decremented = std::prev(lower_bound_it);
```
jwpeterson, line 69: `std::prev` is a good change here. It is in the `<iterator>` [header](https://en.cppreference.com/w/cpp/iterator/prev) which we don't currently include here, though, so please add that for the sake of correctness.

roystgnr:
Okay, I'll merge once CI is through.  We can save the relative tolerances and the end() tolerance test for a later PR.

Issue #2630, opened 2020-06-29
A copy constructor bug in DistributedMesh


YaqiWang:
The `BoundaryInfo` in `DistributedMesh` is not currently copied with multiple processors. I have not looked into the detail but @roystgnr knows the exact problem. He mentioned he was going to fix it and add unit test for it to avoid regression in the future.

PR #2634, opened 2020-06-30
Boundaryinfo equals fix

roystgnr:
A new unit test, intended to replicate the failure noted in #2630, in a commit after a bugfix for that failure.

@YaqiWang, could you try this and see if it works for your case too?

I'll merge this when CI is done either way; even if this turns out not to fix *that* bug, it at least fixes **a** bug.

### DIFF
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -69,9 +69,12 @@ protected:
 
 public:
   /**
-   * Actual copying operation.
+   * Copy assignment operator
    *
-   * \note The copy will have a reference to the same Mesh as the original.
+   * \note \p this will still reference the same MeshBase it was
+   * constructed with.  Boundary data copied from other_boundary_info
+   * will refer to objects in this mesh which have the same
+   * DofObject::id() as the corresponding objects in the other mesh.
    */
   BoundaryInfo & operator=(const BoundaryInfo & other_boundary_info);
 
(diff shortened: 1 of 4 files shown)


jwpeterson:
I have a comment about the documentation for this copy assignment operator:
```
  /**
   * Actual copying operation.
   *
   * \note The copy will have a reference to the same Mesh as the original.
   */
  BoundaryInfo & operator=(const BoundaryInfo & other_boundary_info);
```

I think it should instead say something like: "the mesh reference held by this BoundaryInfo will remain unchanged by assignment" or something along those lines. That's really the only thing that differentiates it from the default copy assignment operator in this case.

Also, while you're at it, it appears that `_ss_id_to_name` and friends are also not properly cleared in `BoundaryInfo::clear()`.

roystgnr:
Damn, good catches.  I'll fix those.

jwpeterson:
```
FAIL meshgenerators/mesh_extruder_generator.extruder/preserve_sideset_names FAILED (ERRMSG)
```
Huh, I guess there was a test actually relying on the behavior of not clearing sideset names in `clear()`. I would not have guessed this was the case. I guess if you want to comment that out to get this passing for now, and maybe we can add it back in later, that would be simplest?

roystgnr:
Nah, I think I finally figured it out.  The bug you noticed while I was fixing the other bug was covering up a third bug.  Looks like the third bug is in libMesh too, so I'll fix that and tack on the fix here.

YaqiWang:
Do not know why I did not notice this. Sorry. Just a quick question, do we need to fix MOOSE for this? It will give me some headache to find the case and rerun it.

roystgnr:
We don't *need* to fix MOOSE; the failure case you noticed should have been completely fixed by this PR.

There's now some MooseMesh copy constructor code that's become redundant, so we should eventually delete a few lines there and just rely on this assignment operator working correctly under the hood, but the old code is just redundant, not incorrect or noticeably slow, so there's no rush.

PR #2637, opened 2020-07-01
Optimize Elem::find_point_neighbors

### LINKED ISSUE #2636: Optimize Elem::find_point_neighbors
`Elem::find_point_neighbors` should (1) check if an element was checked already before performing a point check and (2) use `emplace_hint`.

loganharbour:
Closes #2636

@roystgnr here's the more general optimization I posted about in Slack. I haven't compiled this locally... just copied some of my code. Maybe it'll work the first time? :-)

### DIFF
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -590,6 +590,10 @@ void Elem::find_point_neighbors(const Point & p,
   std::set<const Elem *> untested_set, next_untested_set;
   untested_set.insert(this);
 
+#ifdef LIBMESH_ENABLE_AMR
+  std::vector<const Elem *> active_neighbor_children;
+#endif // #ifdef LIBMESH_ENABLE_AMR
+
   while (!untested_set.empty())
     {
       // Loop over all the elements in the patch that haven't already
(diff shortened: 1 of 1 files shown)


src/geom/elem.C
```diff
@@ -602,34 +606,34 @@ void Elem::find_point_neighbors(const Point & p,
                 {
                   if (current_neighbor->active())                // ... if it is active
                     {
-                      if (current_neighbor->contains_point(p))   // ... and touches p
+                      auto it = neighbor_set.lower_bound(current_neighbor);
```
roystgnr, line 609: lower_bound can't be right here, can it?  There's no sensible ordering on a ```set<T*>```.  Surely we want find?

roystgnr: Wait, I see, you're not directly using it, you're setting it up for emplace_hint later?

roystgnr: Because we don't want to redundantly call contains_point() on an element that's already in the neighbor set, but since we have to hunt through the neighbor set to check that we might as well save the results of that hunt to speed up the insert later.  Sorry, I think I get it now.

loganharbour: Correct. If we know from a search that we don't have it... why not use the failed search for the correct insertion which is guaranteed to happen?

loganharbour: Yes - that was my thought.

Surely for most cases the set search is cheaper than contains_point().

src/geom/elem.C
```diff
@@ -602,34 +606,34 @@ void Elem::find_point_neighbors(const Point & p,
...
+                          neighbor_set.emplace_hint(it, current_neighbor);
                         }
                     }
 #ifdef LIBMESH_ENABLE_AMR
                   else                                 // ... the neighbor is *not* active,
                     {                                  // ... so add *all* neighboring
                                                        // active children that touch p
-                      std::vector<const Elem *> active_neighbor_children;
-
+                      active_neighbor_children.clear();
```
roystgnr, line 622: Are compilers still not smart enough to optimize a container declared in a loop as well as a container cleared at the start of a loop?  Has anyone tested that recently?  I admit I've made these sorts of optimizations myself and IIRC years ago they were a big deal, but I'd hoped that by 2020 we'd be able to just write purer code and let a smarter default allocator handle things.

loganharbour: Hmm. I definitely haven't. I'm fine with this either way.

roystgnr: I'd say leave it in.  I'd love some new benchmarks, but in a conflict between even half-recollected ancient benchmarks and hope, my money's not on hope.

PR #2641, opened 2020-07-07
Fix statistics *.py for python 3

roystgnr:
Python 3 doesn't understand xrange; range is what we want (although this
is slower in Python 2 and they could easily have retained the name
xrange as an option for backwards compatibility COME ON)

numpy byte strings now need to be decoded to give Python unicode
strings.

@jwpeterson - you have a Python 2 install you can check this on?  Everything I have root on seems to be new enough that python2 matplotlib isn't in the repos anymore; everything old I can find doesn't have matplotlib installed.

### DIFF
--- a/doc/statistics/github_traffic_base.py
+++ b/doc/statistics/github_traffic_base.py
@@ -68,7 +68,7 @@ class PlotData(object):
     # Convert date strings into numbers.
     date_nums = []
     for d in date_strings:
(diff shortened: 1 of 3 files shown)


roystgnr:
As long as it's not too inconvenient, I'd like to retain python 2 compatibility, so let's leave the shebang line as plain "python".  This isn't the most computationally intensive set of scripts, so if someone who has ```python->python2``` and ```python3``` in /usr/bin gets the slower version then they'll survive.

Issue #2642, opened 2020-07-08
Sparse matrix addition and multiplication


dewenyushu:
These two functionalities are needed in a preconditioner type as part of the dual mortar development (ref. MOOSE [issue 15215](https://github.com/idaholab/moose/issues/15215)). This preconditioner type needs to condense the original system of equation. During this process, matrix addition and matrix-matrix multiplication will be necessary.

PR #2643, opened 2020-07-08
Add matrix-matrix multiplication and addition

dewenyushu:
Address issue #2642

### DIFF
--- a/include/numerics/petsc_matrix.h
+++ b/include/numerics/petsc_matrix.h
@@ -242,16 +242,16 @@ public:
   /**
    * Compute Y = A*X for matrix \p X.
    */
-  void mat_mult (SparseMatrix<T> & X, SparseMatrix<T> & Y);
+  virtual void matrix_matrix_mult (SparseMatrix<T> & X, SparseMatrix<T> & Y) override;
 
   /**
-   * Add \p scalar* \p spm to the rows and cols of this matrix (A):
-   * A(rows[i], cols[j]) += scalar * spm(i,j)
-   */
+    * Add \p scalar* \p spm to the rows and cols of this matrix (A):
+    * A(rows[i], cols[j]) += scalar * spm(i,j)
+    */
   void add_sparse_matrix (const SparseMatrix<T> & spm,
-                          const std::vector<numeric_index_type> & rows,
-                          const std::vector<numeric_index_type> & cols,
-                          const T scalar);
+                          const std::map<numeric_index_type,numeric_index_type> & row_ltog,
+                          const std::map<numeric_index_type,numeric_index_type> & col_ltog,
+                          const T scalar) override;
 
   virtual T operator () (const numeric_index_type i,
                          const numeric_index_type j) const override;
--- a/include/numerics/petsc_matrix.h
+++ b/include/numerics/petsc_matrix.h
@@ -239,6 +239,20 @@ public:
    */
   virtual void add (const T a, const SparseMatrix<T> & X) override;
 
+  /**
+   * Compute Y = A*X for matrix \p X.
+   */
+  void mat_mult (SparseMatrix<T> & X, SparseMatrix<T> & Y);
+
+  /**
+   * Add \p scalar* \p spm to the rows and cols of this matrix (A):
+   * A(rows[i], cols[j]) += scalar * spm(i,j)
+   */
+  void add_sparse_matrix (const SparseMatrix<T> & spm,
+                          const std::vector<numeric_index_type> & rows,
+                          const std::vector<numeric_index_type> & cols,
+                          const T scalar);
+
   virtual T operator () (const numeric_index_type i,
                          const numeric_index_type j) const override;
 
--- a/src/numerics/petsc_matrix.C
+++ b/src/numerics/petsc_matrix.C
@@ -1201,6 +1201,53 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
 }
 
 
+template <typename T>
+void PetscMatrix<T>::mat_mult (SparseMatrix<T> & X_in, SparseMatrix<T> & Y_out)
+{
+  libmesh_assert (this->initialized());
+
+   // sanity check.
+   libmesh_assert_equal_to (this->n(), X_in.m());
+   libmesh_assert_equal_to (this->m(), Y_out.m());
+   libmesh_assert_equal_to (X_in.n(), Y_out.n());
+
+   const PetscMatrix<T> * X = cast_ptr<const PetscMatrix<T> *> (&X_in);
+   PetscMatrix<T> * Y = cast_ptr<PetscMatrix<T> *> (&Y_out);
+
+   libmesh_assert (X);
+   libmesh_assert (Y);
+
+   PetscErrorCode ierr=0;
+
+   // the matrix from which we copy the values has to be assembled/closed
+   libmesh_assert(X->closed());
+
+   semiparallel_only();
+
+   ierr = MatMatMult(_mat, X->_mat, MAT_INITIAL_MATRIX, PETSC_DEFAULT, &Y->_mat);
+   LIBMESH_CHKERR(ierr);
+
+}
+
+template <typename T>
+void PetscMatrix<T>::add_sparse_matrix (const SparseMatrix<T> & spm,
+                        const std::vector<numeric_index_type> & rows,
+                        const std::vector<numeric_index_type> & cols,
+                        const T scalar)
+{
+  libmesh_assert_equal_to(spm.m(), rows.size());
+  libmesh_assert_equal_to(spm.n(), cols.size());
+
+  for(unsigned int i = 0; i<rows.size(); ++i)
+  {
+    std::vector<numeric_index_type> indices;
+    std::vector<T> values;
+    spm.get_row(i, indices, values);
+    for (unsigned int j=0; j<indices.size(); ++j)
+      this->add (rows[i], cols[indices[j]], scalar*values[j]);
+  }
+}
+
 
 
 template <typename T>
--- a/src/numerics/petsc_matrix.C
+++ b/src/numerics/petsc_matrix.C
@@ -30,6 +30,7 @@
 #include "libmesh/dense_matrix.h"
 #include "libmesh/petsc_vector.h"
 #include "libmesh/parallel.h"
+#include "libmesh/utility.h"
 
 
 #ifdef LIBMESH_ENABLE_BLOCKED_STORAGE
@@ -1202,7 +1203,7 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
 
 
 template <typename T>
-void PetscMatrix<T>::mat_mult (SparseMatrix<T> & X_in, SparseMatrix<T> & Y_out)
+void PetscMatrix<T>::matrix_matrix_mult (SparseMatrix<T> & X_in, SparseMatrix<T> & Y_out)
 {
   libmesh_assert (this->initialized());
 
@@ -1213,9 +1214,6 @@ void PetscMatrix<T>::mat_mult (SparseMatrix<T> & X_in, SparseMatrix<T> & Y_out)
 
    const PetscMatrix<T> * X = cast_ptr<const PetscMatrix<T> *> (&X_in);
    PetscMatrix<T> * Y = cast_ptr<PetscMatrix<T> *> (&Y_out);
-
-   libmesh_assert (X);
-   libmesh_assert (Y);
 
    PetscErrorCode ierr=0;
 
@@ -1231,24 +1229,53 @@ void PetscMatrix<T>::mat_mult (SparseMatrix<T> & X_in, SparseMatrix<T> & Y_out)
 
 template <typename T>
 void PetscMatrix<T>::add_sparse_matrix (const SparseMatrix<T> & spm,
-                        const std::vector<numeric_index_type> & rows,
-                        const std::vector<numeric_index_type> & cols,
+                        const std::map<numeric_index_type, numeric_index_type> & row_ltog,
+                        const std::map<numeric_index_type, numeric_index_type> & col_ltog,
                         const T scalar)
 {
-  libmesh_assert_equal_to(spm.m(), rows.size());
-  libmesh_assert_equal_to(spm.n(), cols.size());
+  libmesh_assert_equal_to(spm.m(), row_ltog.size());
+  libmesh_assert_equal_to(spm.n(), col_ltog.size());
+  // make sure matrix has larger size than spm
+  libmesh_assert_greater_equal(this->m(), spm.m());
+  libmesh_assert_greater_equal(this->n(),spm.n());
 
-  for(unsigned int i = 0; i<rows.size(); ++i)
+  if(!this->closed())
+    this->close();
+
+  PetscErrorCode ierr=0;
+
+  const PetscMatrix<T> * pscm=cast_ptr<const PetscMatrix<T> *>(&spm);
+
+  PetscInt ncols=0;
+
+  const PetscInt *lcols;
+  const PetscScalar *vals;
+
+  std::vector<PetscInt> gcols;
+  std::vector<PetscScalar> values;
+
+  for (auto ltog: row_ltog)
   {
-    std::vector<numeric_index_type> indices;
-    std::vector<T> values;
-    spm.get_row(i, indices, values);
-    for (unsigned int j=0; j<indices.size(); ++j)
-      this->add (rows[i], cols[indices[j]], scalar*values[j]);
+    PetscInt grow[]={static_cast<PetscInt>(ltog.second)}; // global row index
+
+    ierr = MatGetRow(pscm->_mat, static_cast<PetscInt>(ltog.first), &ncols, &lcols, &vals);
+    LIBMESH_CHKERR(ierr);
+    // get global indices (gcols) from lcols, increment values = vals*scalar
+    gcols.resize(ncols);
+    values.resize(ncols);
+    for (auto i : index_range(gcols))
+    {
+      gcols[i] = libmesh_map_find(col_ltog, lcols[i]);
+      values[i] = scalar*vals[i];
+    }
+    ierr=MatSetValues(_mat, 1, grow, ncols, gcols.data(), values.data(),  ADD_VALUES);
+    LIBMESH_CHKERR(ierr);
+    ierr  = MatRestoreRow(pscm->_mat, static_cast<PetscInt>(ltog.first), &ncols, &lcols, &vals);
+    LIBMESH_CHKERR(ierr);
   }
+  // Note: We are not closing the matrix because it is expensive to do so when adding multiple sparse matrices.
+  //       Remember to manually close the matrix once all changes to the matrix have been made.
 }
-
-
 
 template <typename T>
 T PetscMatrix<T>::operator () (const numeric_index_type i_in,
--- a/include/numerics/sparse_matrix.h
+++ b/include/numerics/sparse_matrix.h
@@ -281,6 +281,22 @@ public:
   virtual void add (const T a, const SparseMatrix<T> & X) = 0;
 
   /**
+   * Compute Y = A*X for matrix \p X.
+   */
+  virtual void matrix_matrix_mult (SparseMatrix<T> & /*X*/, SparseMatrix<T> & /*Y*/)
+  { libmesh_not_implemented(); }
+
+  /**
+   * Add \p scalar* \p spm to the rows and cols of this matrix (A):
+   * A(rows[i], cols[j]) += scalar * spm(i,j)
+   */
+  virtual void add_sparse_matrix (const SparseMatrix<T> & /*spm*/,
+                          const std::map<numeric_index_type, numeric_index_type> & /*rows*/,
+                          const std::map<numeric_index_type, numeric_index_type> & /*cols*/,
+                          const T /*scalar*/)
+  {libmesh_not_implemented();}
+
+  /**
    * \returns A copy of matrix entry \p (i,j).
    *
    * \note This may be an expensive operation, and you should always
--- a/src/numerics/sparse_matrix.C
+++ b/src/numerics/sparse_matrix.C
@@ -50,7 +50,6 @@ SparseMatrix<T>::SparseMatrix (const Parallel::Communicator & comm_in) :
 
 
 
-
 // default implementation is to fall back to non-blocked method
 template <typename T>
 void SparseMatrix<T>::add_block_matrix (const DenseMatrix<T> & dm,


include/numerics/petsc_matrix.h
```diff
@@ -239,6 +239,19 @@ class PetscMatrix final : public SparseMatrix<T>
    */
   virtual void add (const T a, const SparseMatrix<T> & X) override;
 
+  /**
+   * Compute Y = A*X for matrix \p X.
+   */
+  void mat_mult (SparseMatrix<T> & X, SparseMatrix<T> & Y);
+
+  /**
+   * Add \p scalar* \p spm to the rows and cols of this matrix
```
roystgnr, line 248: Could we expand this explanation a bit?  Something like ```A_rows[i]cols[j] += scalar * spm_ij```?  "to the rows and cols" seems ambiguous until you go look at the implementation.

dewenyushu: Good point! This has been taken care of.

src/numerics/petsc_matrix.C
```diff
@@ -1201,6 +1201,53 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
...
+
+   semiparallel_only();
+
+   ierr = MatMatMult(_mat, X->_mat, MAT_INITIAL_MATRIX, PETSC_DEFAULT, &Y->_mat);
+   LIBMESH_CHKERR(ierr);
+
+}
+
+template <typename T>
+void PetscMatrix<T>::add_sparse_matrix (const SparseMatrix<T> & spm,
```
jwpeterson, line 1233: We already have
```
  /**
   * Compute \f$ A \leftarrow A + a*X \f$ for scalar \p a, matrix \p X.
   */
  virtual void add (const T a, const SparseMatrix<T> & X) = 0;
```
Would it be possible for you to use that instead? If not, then there's nothing here which is specific to PETSc so it would make more sense to add this new API at the SparseMatrix level.

dewenyushu: I do not think I will be able to use 

` virtual void add (const T a, const SparseMatrix<T> & X) = 0;`

because in my case X has smaller size than A, and I would like to add to certain rows and cols of A.

dewenyushu:
> I'd prefer to add virtual implemented-as-error stubs to the SparseMatrix API rather than one-off capabilities at the `PetscMatrix` level.

Thanks for the comment! Just want to make sure: by this you mean adding pure virtual functions in SparseMatrix, i.e., `virtual add_sparse_matrix(...) = 0;` and keep the actual realization in PetscMatrix or do you suggest moving the functions to SparseMatrix?

src/numerics/petsc_matrix.C
jwpeterson: OK, looks like we should also assert that `*this` has at least as many rows/cols as `spm` then? I also think you may find that this implementation is inefficient since it calls `add()` once per entry rather than calling `MatSetValues` with an entire row at a time. Therefore this could serve as a base class implementation, but @fdkong may be able to suggest an improvement for the `PetscMatrix` specialization.

One other small point is that I think you should declare `indices` and `values` outside the i-loop, no reason to create/destroy them `rows.size()` times.

jwpeterson:
> Thanks for the comment! Just want to make sure: by this you mean adding pure virtual functions in SparseMatrix, i.e., `virtual add_sparse_matrix(...) = 0;` and keep the actual realization in PetscMatrix or do you suggest moving the functions to SparseMatrix?

Since there is no "generic" version of mat_mult(), it should be implemented as an error in sparse_matrix.h, i.e.:
```
virtual void mat_mult (SparseMatrix<T> & X, SparseMatrix<T> & Y)
{ libmesh_not_implemented(); }
```
Then you would add your existing implementation as a virtual `override` in petsc_matrix.h/C.

The `add_sparse_matrix()` routine that you have here is generic enough to go directly into sparse_matrix.C/h. Then, at some later point, a version optimized for `PetscMatrix` (for example that sets entire rows values at a time) could be used to `override` the base class version.

Generally I am against implemented-as-error APIs in base classes but in this case the plan is that sparse matrix multiplication would eventually be implemented for EigenSparseMatrix, EpetraMatrix, etc. in the future when someone has time/motivation, not that it would be permanently unimplemented.

src/numerics/petsc_matrix.C
```diff
@@ -1201,6 +1201,53 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
...
+  libmesh_assert_equal_to(spm.m(), rows.size());
+  libmesh_assert_equal_to(spm.n(), cols.size());
+
+  for(unsigned int i = 0; i<rows.size(); ++i)
+  {
+    std::vector<numeric_index_type> indices;
+    std::vector<T> values;
+    spm.get_row(i, indices, values);
+    for (unsigned int j=0; j<indices.size(); ++j)
+      this->add (rows[i], cols[indices[j]], scalar*values[j]);
```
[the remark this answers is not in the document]
fdkong: I am not sure about where you are using this function, but for me, `spm` seems to be small since you are using `std::vector<numeric_index_type>` to map local indices to global indices.

OK, if `spm` is a parallel matrix, we should not use `std::vector`   since `cols` is sparse. If you do this way, you will run out of memory really soon. 

Hopefully, I overreact.

dewenyushu:
> > Thanks for the comment! Just want to make sure: by this you mean adding pure virtual functions in SparseMatrix, i.e., `virtual add_sparse_matrix(...) = 0;` and keep the actual realization in PetscMatrix or do you suggest moving the functions to SparseMatrix?
> 
> Since there is no "generic" version of mat_mult(), it should be implemented as an error in sparse_matrix.h, i.e.:
> 
> ```
> virtual void mat_mult (SparseMatrix<T> & X, SparseMatrix<T> & Y)
> { libmesh_not_implemented(); }
> ```
> 
> Then you would add your existing implementation as a virtual `override` in petsc_matrix.h/C.
> 
> The `add_sparse_matrix()` routine that you have here is generic enough to go directly into sparse_matrix.C/h. Then, at some later point, a version optimized for `PetscMatrix` (for example that sets entire rows values at a time) could be used to `override` the base class version.
> 
> Generally I am against implemented-as-error APIs in base classes but in this case the plan is that sparse matrix multiplication would eventually be implemented for EigenSparseMatrix, EpetraMatrix, etc. in the future when someone has time/motivation, not that it would be permanently unimplemented.

Makes sense. Thanks for the clarification!

src/numerics/petsc_matrix.C
```diff
@@ -1201,6 +1201,53 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
...
+template <typename T>
+void PetscMatrix<T>::add_sparse_matrix (const SparseMatrix<T> & spm,
+                        const std::vector<numeric_index_type> & rows,
+                        const std::vector<numeric_index_type> & cols,
+                        const T scalar)
+{
+  libmesh_assert_equal_to(spm.m(), rows.size());
+  libmesh_assert_equal_to(spm.n(), cols.size());
+
+  for(unsigned int i = 0; i<rows.size(); ++i)
```
fdkong, line 1241: `i` should have the same data type as `rows.size()`

roystgnr:
Should we warn people in the mat_mult (or mat_mat_mult, I like Fande's suggestion) comment about the sparsity issues in play?  With a typical 3D stencil the product of two matrices has roughly 4-8 times the fill of each matrix alone.  Just how most people who think they want to invert a dense matrix really want to factorize it, I suspect most people who think they want to multiply sparse matrices really want to treat them as a two-stage operator.

include/numerics/petsc_matrix.h
```diff
@@ -239,6 +239,20 @@ class PetscMatrix final : public SparseMatrix<T>
    */
   virtual void add (const T a, const SparseMatrix<T> & X) override;
 
+  /**
+   * Compute Y = A*X for matrix \p X.
+   */
+  void mat_mult (SparseMatrix<T> & X, SparseMatrix<T> & Y);
+
```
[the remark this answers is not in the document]
jwpeterson: Good point. If we're changing it "mat" doesn't seem to be prevalent currently in other libmesh APIs but `mult` is, so maybe `matrix_matrix_mult()`.

src/numerics/petsc_matrix.C
roystgnr: Agreed.  Perhaps ```for (auto i : index_range(rows))``` here?

jwpeterson: The current pseudo-standard loop style for vector looping is `for (auto i : index_range(rows))` which handles the data typing correctly and only calls `size()` exactly once.

src/numerics/petsc_matrix.C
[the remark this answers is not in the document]
jwpeterson: @dewenyushu It might be good if you could explain what you are trying to do, since sparse matrix-matrix multiplication is rather unusual. I had assumed that @fdkong would be familiar with the preconditioner you are trying to implement, but it sounds like he's not. I guess it is somehow related to the Schur complement/condensing out the Lagrange multiplier DOFs, so there may be a more efficient way to go about this.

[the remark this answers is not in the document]
dewenyushu: @jwpeterson you are correct, both functions are related to condensing out the LM dofs. In order to do so, I need to extract part of the original Jacobian matrix (e.g., D and M shown below), do some calculations (e.g., MDinv*K1ci, etc), and put the new block to the corresponding indices of the condensed Jacobian (shown on the right). 

Screen Shot 2020-07-08 at 1 19 15 PM

include/numerics/petsc_matrix.h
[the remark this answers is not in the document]
dewenyushu: Yea, I agree you guys about the naming. I will change it to `matrix_matrix_mult()`

src/numerics/petsc_matrix.C
```diff
@@ -1201,6 +1201,53 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
...
+  {
+    std::vector<numeric_index_type> indices;
+    std::vector<T> values;
+    spm.get_row(i, indices, values);
```
fdkong, line 1245: This is not quite right in parallel. `i ` won't start from zero for nonzero MPI ranks.

src/numerics/petsc_matrix.C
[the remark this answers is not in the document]
fdkong: Oh, my fault, I ignored the context here. Yes, then `spm` is a regular parallel sparse matrix.

`GetRow, and MatSetValues` should work fine. 

But you have to change to use 


```
const std::map<numeric_index_type> & rows,
const std::map<numeric_index_type> & cols,
```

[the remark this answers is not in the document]
dewenyushu: > I am not sure about where you are using this function, but for me, `spm` seems to be small since you are using `std::vector<numeric_index_type>` to map local indices to global indices.
> 
> OK, if `spm` is a parallel matrix, we should not use `std::vector` since `cols` is sparse. If you do this way, you will run out of memory really soon.
> 
> Hopefully, I overreact.

 The row number of spm is the number of dofs on the master contact interface, col number is either the number of dofs of the master interior or contact interface

fdkong:
> Should we warn people in the mat_mult (or mat_mat_mult, I like Fande's suggestion) comment about the sparsity issues in play? With a typical 3D stencil the product of two matrices has roughly 4-8 times the fill of each matrix alone. Just how most people who think they want to invert a dense matrix really want to factorize it, I suspect most people who think they want to multiply sparse matrices really want to treat them as a two-stage operator.

MatMatMult is a valid operation. Yes, we know it could be expensive in memory, time and communcation. It is totally problem-dependent. In AMG, we have to do this operation (PtAP or MatMatMatMult).

fdkong:
`this` (matrix) should be preallocated correctly before we call `add_sparse_matrix`, otherwise, the code will be unbelievably slow.

src/numerics/petsc_matrix.C
dewenyushu: > This is not quite right in parallel. `i ` won't start from zero for nonzero MPI ranks.

That is a good point. So what to do here?

fdkong: Could you push up your moose changes? We should start from the caller

fdkong: `const std::map<numeric_index_type> & row_ltog,`

and then

```
for (auto ltog: row_ltog)
   spm.get_row(ltog.first, indices, values);
```

dewenyushu: > Could you push up your moose changes? We should start from the caller

Please look at an initial implementation for condensing out the LMs [here](https://github.com/idaholab/moose/pull/15565)

src/numerics/petsc_matrix.C
```diff
@@ -1201,6 +1201,53 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
...
+   libmesh_assert (Y);
+
+   PetscErrorCode ierr=0;
+
+   // the matrix from which we copy the values has to be assembled/closed
+   libmesh_assert(X->closed());
+
+   semiparallel_only();
+
+   ierr = MatMatMult(_mat, X->_mat, MAT_INITIAL_MATRIX, PETSC_DEFAULT, &Y->_mat);
```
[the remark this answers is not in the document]
dewenyushu: > We should leave `MatReuse` as an in parameter, and set its default value as `MAT_INITIAL_MATRIX`
> 
> For some cases, if your matrix sparsity does not change, it is a big deal to reuse the sparsity pattern using `MAT_REUSE_MATRIX`

I am keeping a virtual version in `sparse_matrix.h`:

```
virtual void matrix_matrix_mult (SparseMatrix<T> & /*X*/, SparseMatrix<T> & /*Y*/, const MatReuse & /*Reuse=MAT_INITIAL_MATRIX*/)
  { libmesh_not_implemented(); }
```

Therefore, I included the `petscmat.h` in the header. It compiles fine on my own machine while causing some failed builds on GitHub (complaining about not being able to find the `petscmat.h`). Any ideas on how to eliminate the error here? @fdkong

src/numerics/petsc_matrix.C
```diff
@@ -1201,7 +1201,86 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
...
+  const PetscScalar *vals;
+
+  for (auto ltog: row_ltog)
+  {
+    PetscInt grow[]={static_cast<PetscInt>(ltog.second)}; // global row index
+
+    ierr = MatGetRow(pscm->_mat, static_cast<PetscInt>(ltog.first), &ncols, &lcols, &vals);
+    LIBMESH_CHKERR(ierr);
+    // get global indices (gcols) from lcols, increment values = vals*scalar
+    PetscInt gcols[static_cast<std::size_t>(ncols)];
```
jwpeterson, line 1266: I know this is a work in progress, but just wanted to point out: this is creating a variable length array on the stack, which some (many?) compilers will warn about and which runs the risk of exceeding the available stack space, so you should use a std::vector for these and then pass the `.data()` pointer of the vector to PETSc APIs.

src/numerics/petsc_matrix.C
```diff
@@ -1201,7 +1201,86 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
...
+    PetscScalar values[static_cast<std::size_t>(ncols)];
+    for (auto i=0; i<static_cast<std::size_t>(ncols); ++i)
```
jwpeterson, line 1268: Since `ncols` is a `PetscInt`, just declare `i` to be a `PetscInt` as well. IIRC, the important thing is as far as efficiency is concerned is that the loop iteration counter type matches the loop end type, so no reason to involve the unsigned `std::size_t` type here.

src/numerics/petsc_matrix.C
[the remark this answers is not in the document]
jwpeterson: We can't introduce the PETSc type `MatReuse` at the base class level (and we can't include PETSc headers without properly guarding them) since libMesh supports `--disable-petsc` builds. You can work around this issue by creating a new enum in libmesh which you can then map to the PETSc enumeration values. (see, for example: enum_convergence_flags.h). Since this is kind of an optional/nice-to-have feature, I'd say you could skip it in your current PR and then myself or one of the other devs can go back and implement it a bit later, once your current PR is merged.

src/numerics/petsc_matrix.C
dewenyushu: I was worried about the variable size warnings but was not sure how to pass by this. This makes sense. I will make the change. Thanks!

src/numerics/petsc_matrix.C
```diff
@@ -1201,7 +1201,89 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
...
+
+    ierr = MatGetRow(pscm->_mat, static_cast<PetscInt>(ltog.first), &ncols, &lcols, &vals);
+    LIBMESH_CHKERR(ierr);
+    // get global indices (gcols) from lcols, increment values = vals*scalar
+    gcols.resize(static_cast<std::size_t>(ncols));
+    values.resize(static_cast<std::size_t>(ncols));
+    for (auto i : index_range(gcols))
+    {
+      numeric_index_type lcol=static_cast<numeric_index_type>(lcols[i]);
+      gcols[i] = static_cast<PetscInt>(col_ltog.at(lcol));
```
jwpeterson, line 1274: Instead of `map::at()` which throws a difficult to trace exception, so let's replace these two lines with:
```
gcols[i] = libmesh_map_find(col_ltog, lcols[i]);
```

src/numerics/petsc_matrix.C
```diff
@@ -1201,7 +1201,89 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
...
+
+  if(!this->closed())
+    this->close();
 
+  PetscErrorCode ierr=0;
+
+  ierr=MatSetOption(_mat, MAT_NEW_NONZERO_ALLOCATION_ERR, PETSC_FALSE);
+  LIBMESH_CHKERR(ierr);
+
+  const PetscMatrix<T> * pscm=dynamic_cast<const PetscMatrix<T> *>(&spm);
```
jwpeterson, line 1252: Use `cast_ptr` here instead of `dynamic_cast`.

src/numerics/petsc_matrix.C
```diff
@@ -1201,7 +1201,89 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
...
+    ierr = MatGetRow(pscm->_mat, static_cast<PetscInt>(ltog.first), &ncols, &lcols, &vals);
+    LIBMESH_CHKERR(ierr);
+    // get global indices (gcols) from lcols, increment values = vals*scalar
+    gcols.resize(static_cast<std::size_t>(ncols));
+    values.resize(static_cast<std::size_t>(ncols));
+    for (auto i : index_range(gcols))
+    {
+      numeric_index_type lcol=static_cast<numeric_index_type>(lcols[i]);
+      gcols[i] = static_cast<PetscInt>(col_ltog.at(lcol));
+      values[i] = static_cast<PetscScalar>(scalar*vals[i]);
```
jwpeterson, line 1275: This `static_cast` should be removed unless you were somehow getting warnings about this line of code?

src/numerics/petsc_matrix.C
```diff
@@ -1201,7 +1201,89 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
...
+  std::vector<PetscScalar> values;
+
+  for (auto ltog: row_ltog)
+  {
+    PetscInt grow[]={static_cast<PetscInt>(ltog.second)}; // global row index
+
+    ierr = MatGetRow(pscm->_mat, static_cast<PetscInt>(ltog.first), &ncols, &lcols, &vals);
+    LIBMESH_CHKERR(ierr);
+    // get global indices (gcols) from lcols, increment values = vals*scalar
+    gcols.resize(static_cast<std::size_t>(ncols));
```
jwpeterson, line 1269: static_cast should not be needed here or in the following line.

src/numerics/petsc_matrix.C
```diff
@@ -1201,7 +1201,89 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
...
+
+   // sanity check.
+   libmesh_assert_equal_to (this->n(), X_in.m());
+   libmesh_assert_equal_to (this->m(), Y_out.m());
+   libmesh_assert_equal_to (X_in.n(), Y_out.n());
+
+   const PetscMatrix<T> * X = cast_ptr<const PetscMatrix<T> *> (&X_in);
+   PetscMatrix<T> * Y = cast_ptr<PetscMatrix<T> *> (&Y_out);
+
+   libmesh_assert (X);
```
jwpeterson, line 1217: The `cast_ptr` call already asserts whether the cast worked in debug mode, so these additional asserts are redundant.

src/numerics/petsc_matrix.C
```diff
@@ -1201,7 +1201,89 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
...
+  // make sure matrix has larger size than spm
+  libmesh_assert_greater_equal(this->m(), spm.m());
+  libmesh_assert_greater_equal(this->n(),spm.n());
+
+  if(!this->closed())
+    this->close();
 
+  PetscErrorCode ierr=0;
+
+  ierr=MatSetOption(_mat, MAT_NEW_NONZERO_ALLOCATION_ERR, PETSC_FALSE);
```
jwpeterson, line 1249: @fdkong this is going to make the `add_sparse_matrix()` routine really slow in case the matrix is not pre-allocated correctly, right? Do we have no way of properly allocating the sparsity ahead of time in the application? I prefer not to allow people to do something that will make their code really slow, since then they will blame us for that :)

fdkong: Right, we should remove this line of code. In the preconditioner, we definitely can figure out the preallocation even though it is not straightforward. 

We will preallocate memory correctly before we call `add_sparse_matrix`

dewenyushu: Good point @jwpeterson @fdkong . I agree it would be ideal to pre-allocate memory before calling this function. I have no idea how to do that though... hopefully @fdkong can give me some directions on this.

dewenyushu:
> Thanks! If @fdkong also approves, I will just make a few small whitespace/formatting changes and then merge.

Sounds great! Thanks for your comments. Really appreciate all the help here!

src/numerics/petsc_matrix.C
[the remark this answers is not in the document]
fdkong: @dewenyushu Do you like to address John's comments? Just introduce `enum`.

It is will be very useful because you reuse expensive calculations during the preconditioner application.

src/numerics/petsc_matrix.C
```diff
@@ -1201,7 +1202,82 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
...
+    }
+    ierr=MatSetValues(_mat, 1, grow, ncols, gcols.data(), values.data(),  ADD_VALUES);
+    LIBMESH_CHKERR(ierr);
+    ierr  = MatRestoreRow(pscm->_mat, static_cast<PetscInt>(ltog.first), &ncols, &lcols, &vals);
+    LIBMESH_CHKERR(ierr);
+  }
+  ierr=MatAssemblyBegin(_mat, MAT_FINAL_ASSEMBLY);
+  LIBMESH_CHKERR(ierr);
+  ierr=MatAssemblyEnd(_mat, MAT_FINAL_ASSEMBLY);
+  LIBMESH_CHKERR(ierr);
```
fdkong, line 1279: You do not want to call MatAssembly until you have done all `add_sparse_matrix`. For example,

```
J-> add_sparse_matrix(A1, rows1, cols1)
J-> add_sparse_matrix(A2, rows2, cols2)
.
..
J-> add_sparse_matrix(A100, rows100, cols100),

J->close()
```

That is, we will call `close` in moose.

jwpeterson: > Just introduce `enum`.

@fdkong let me make a separate ticket for this one, I can take care of it.

jwpeterson: > You do not want to call MatAssembly

This seems like an important thing to fix before merge...

dewenyushu: > You do not want to call MatAssembly until you have done all `add_sparse_matrix`. For example,
> 
> ```
> J-> add_sparse_matrix(A1, rows1, cols1)
> J-> add_sparse_matrix(A2, rows2, cols2)
> .
> ..
> J-> add_sparse_matrix(A100, rows100, cols100),
> 
> J->close()
> ```
> 
> That is, we will call `close` in moose.

I thought it would be safer to close the matrix here every time when finishing adding a sparse matrix to the original matrix (also, to prevent needing to close the matrix manually). Is it very inefficient to do so?

fdkong: We can not do that since the matrix memory will be shrunk after you call `MatAssembly`.  If you did this, there were two issues:

1) You would use much more memory than you should 

2) The code would be really slow because of many allocs.

dewenyushu: > We can not do that since the matrix memory will be shrunk after you call `MatAssembly`. If you did this, there were two issues:
> 
> 1. You would use much more memory than you should
> 2. The code would be really slow because of many allocs.

okay I see. In this case it makes sense to avoid calling `MatAssembly` many times by putting it here in this function. Thanks!

dewenyushu: > > Just introduce `enum`.
> 
> @fdkong let me make a separate ticket for this one, I can take care of it.

@jwpeterson thanks!

PR #2646, opened 2020-07-11
Check whether nonlinear residual vector and base vector are equal

lindsayad:
Reusing the nonlinear residual vector as the base vector for matrix-free
finite differencing was an optimization we introduced in #1572. However,
per idaholab/moose#14644 we would like to be sure that it is a safe
thing to do. So I propose that in debug mode we check and make sure that
they really are the same.

Closes idaholab/moose#14644

### DIFF
--- a/src/solvers/petsc_nonlinear_solver.C
+++ b/src/solvers/petsc_nonlinear_solver.C
@@ -297,7 +297,78 @@ extern "C"
     PetscNonlinearSolver<Number> * solver =
       static_cast<PetscNonlinearSolver<Number> *> (ctx);
 
-    return libmesh_petsc_snes_mffd_residual(solver->snes(), x, r, ctx);
+    PetscErrorCode ierr = libmesh_petsc_snes_mffd_residual(solver->snes(), x, r, ctx);
+    CHKERRABORT(solver->comm().get(), ierr);
+
+#if !PETSC_VERSION_LESS_THAN(3,8,4)
+#ifndef NDEBUG
+
+    // When the user requested to reuse the nonlinear residual as the base for doing matrix-free
+    // approximation of the Jacobian, we'll do a sanity check to make sure that that was safe to do
+    if (solver->snes_mf_reuse_base() && (solver->comm().size() == 1) && (libMesh::n_threads() == 1))
+    {
+      SNES snes = solver->snes();
+
+      KSP ksp;
+      ierr = SNESGetKSP(snes, &ksp);
+      CHKERRABORT(solver->comm().get(), ierr);
+
+      PetscInt ksp_it;
+      ierr = KSPGetIterationNumber(ksp, &ksp_it);
+      CHKERRABORT(solver->comm().get(), ierr);
+
+      SNESType snes_type;
+      ierr = SNESGetType(snes, &snes_type);
+      CHKERRABORT(solver->comm().get(), ierr);
+
+      libmesh_assert_msg(snes_type, "We're being called from SNES; snes_type should be non-null");
+
+      Mat J;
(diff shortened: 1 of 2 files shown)


roystgnr:
Looks like it was a good thing you checked...

lindsayad:
Ok once idaholab/moose#15566 gets in, then we should be able to invalidate the "Test debug" target and move forward here

lindsayad:
May I get a review on this?

src/solvers/petsc_nonlinear_solver.C
```diff
@@ -297,7 +297,78 @@ extern "C"
...
+                "You requested to reuse the nonlinear residual vector as the base vector for "
+                "computing the action of the matrix-free Jacobian, but the vectors are not "
+                "the same. Your physics must have states; either remove the states "
+                "from your code or make sure that you set_mf_reuse_base(false)");
+        }
+
+        // There are always exactly two function evaluations for the zeroth ksp iteration when doing
+        // matrix-free approximation of the Jacobian action: one corresponding to the evaluation of
+        // the base vector, and the other corresponding to evaluation of the perturbed vector. So we
+        // toggle back and forth between states
```
roystgnr, line 364: This seems fragile to rely on, but I gather that, if someone tries to do anything more complicated, the above error checking would catch it?

lindsayad: I also don't like it, but I don't believe there's currently any way to query Petsc for "are we calling the residual function because we are evaluating the base for approximating the Jacobian action". @fdkong ?

You would definitely get an error if something screwy was going on...it just may not be the user's fault. e.g. if PETSc somehow decided that they were going to do more than two function evaluations during the zeroth linear iteration.

fdkong: I do not follow this thread carefully. Could you explain what you want?

fdkong: There is no API for this right now.  You could check if you are on the first linear iteration of the first nonlinear iteration

jwpeterson:
> I'd say wait a bit longer before merging in case @jwpeterson wants a look too.

Don't know enough to comment intelligently... although it looks like the extra checking you are doing is only in debug mode when running in serial. Is that right?

src/solvers/petsc_nonlinear_solver.C
lindsayad: > There is no API for this right now.

Ok, that's good to know!

>  You could check if you are on the first linear iteration of the first nonlinear iteration

That's not enough. When not reusing the base, there will be *two* calls to `SNESComputeFunction` even during the zeroth (or first depending on what you want to call it) linear iteration, one to compute the base and one to compute the perturbation in order to approximate the Jacobian action. We generally hope that the first call (the base evaluation) corresponds to the last nonlinear residual evaluation. However, the second call (for the perturbation) definitely should not match the nonlinear residual evaluation. That's why you have to be very careful about when you perform the comparison. That's why I introduced this `computing_base_vector boolean`. 

Note that you also want to do this check in general for every nonlinear iteration. For instance if you don't come into mechanical contact until later in your nonlinear solve, you wouldn't see a discrepancy between the nonlinear residual and the matrix-free base residual until that point, e.g. you would have false confidence that the function evaluations are the same if you only did the check for the initial nonlinear iteration.

lindsayad:
> Don't know enough to comment intelligently... although it looks like the extra checking you are doing is only in debug mode when running in serial. Is that right?

Correct!

src/solvers/petsc_nonlinear_solver.C
fdkong: OK, the whole thing is for debugging, and so I won't worry much.

PR #2648, opened 2020-07-15
TIMPI 1.1.3 update

roystgnr:
@grmnptr is now also hitting the warnings we fixed in https://github.com/libMesh/TIMPI/pull/33 but haven't yet brought downstream, and it'd be a good idea to pull in that sync optimization too.

We can also now use the new TIMPI gather/allgather_packed_range
buffer size arguments to avoid the bug @fdkong hit, when we a sufficiently large mesh
on sufficiently many processors.  Handling the redundancy in incoming
buffers one chunk at a time should keep our total memory allocation
reasonable.


fdkong:
Thanks @roystgnr

PR #2653, opened 2020-07-20
No more _find_global_index_by_pid_map n_procs loop

roystgnr:
The old code was probably fine when Ben wrote it in 2003 (when it was
only for Parmetis and only used on small processor counts), but moving
it up to Partitioner was making partitioning at high processor counts
take forever.

I've barely started testing this, but @fdkong and @YaqiWang are getting held up by the slowdown.  Fande, give this version a try with your slow case, let me know how much speedup we get?


roystgnr:
Fande says his problem case just went from 3100 seconds to 234 seconds, so I'm planning to merge this as soon as CI finishes.

fdkong:
This PR saves me! Thanks

PR #2654, opened 2020-07-20
Copy extra integers and mapping data whenever "copying" elements

### LINKED ISSUE #2644: UnstructuredMesh::all_second_order not copying the extra element integers
The function creates second order elements and copies values from the original first order elements. But the function missed copying the extra element integers. The memory for holding these integers seems indeed allocated. Tag @roystgnr

roystgnr:
This should fix #2644, in which @YaqiWang noticed that we hadn't updated all_second_order() to properly copy extra elem integer data from low order elements to the high order elements replacing them.

While I was at it I looked through the code for any similar failures elsewhere, and fixed those: extra integer data and/or mapping data wasn't being copied in flatten(), copy_nodes_and_elements(), create_submesh(), and all_first_order().

There's still one remaining fix to be made: mapping data isn't making it into Elem packed_range serialization, which should break parallel repartitioning on isogeometric meshes.  That's a slightly trickier fix so I'll leave it for a future PR.

### DIFF
--- a/src/mesh/mesh_modification.C
+++ b/src/mesh/mesh_modification.C
@@ -939,13 +939,24 @@ void MeshTools::Modification::all_tri (MeshBase & mesh)
             }
           } // end switch (etype)
 
-
-
-        // Be sure the correct IDs are also set for all subelems.
+        // Be sure the correct data is set for all subelems.
+        const unsigned int nei = elem->n_extra_integers();
         for (unsigned int i=0; i != max_subelems; ++i)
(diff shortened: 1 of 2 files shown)


lindsayad:
@YaqiWang can you look at this soon? Otherwise I'm going to have idaholab/moose#15656 merged.

roystgnr:
This may not fix the distributed IGA case, but it doesn't regress anything, and I'm confident enough about the other fixes that I want them in before our next release.

PR #2655, opened 2020-07-23
Major refresh of the EIM framework

jwpeterson:
Mainly this change reorders how we store EIM basis functions so that it's easier to return values for all quad points on an element and the code overall runs faster. This change is not backwards-compatible but it's unlikely there are many external codes using these APIs. We've updated the relevant libmesh examples, so people can also use those to update their codes if required. Finally, if required, once this is merged I can always retroactively start a 1.6.0 release branch from the commit just before the merge in order to have one more release with the old API.

### DIFF
--- a/src/reduced_basis/rb_eim_construction.C
+++ b/src/reduced_basis/rb_eim_construction.C
@@ -58,8 +58,7 @@ RBEIMConstruction::RBEIMConstruction (EquationSystems & es,
     best_fit_type_flag(PROJECTION_BEST_FIT),
     _Nmax(0),
     _rel_training_tolerance(1.e-4),
-    _abs_training_tolerance(1.e-12),
-    _perturb_size(1.e-6)
+    _abs_training_tolerance(1.e-12)
 {
   // The training set should be the same on all processors in the
   // case of EIM training.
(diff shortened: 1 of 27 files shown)


jwpeterson:
FYI @roystgnr this update is working well for us internally, so as far as we're concerned it's good to merge.

roystgnr:
Let me throw on a --disable-deprecated test while I finish looking it over...

src/reduced_basis/rb_eim_construction.C
```diff
@@ -136,797 +111,831 @@ void RBEIMConstruction::set_best_fit_type_flag (const std::string & best_fit_typ
...
 
-ExplicitSystem & RBEIMConstruction::get_explicit_system()
+std::vector<std::unique_ptr<ElemAssembly>> & RBEIMConstruction::get_eim_assembly_objects()
 {
-  return get_equation_systems().get_system<ExplicitSystem>(_explicit_system_name);
+  return _rb_eim_assembly_objects;
 }
 
-void RBEIMConstruction::load_basis_function(unsigned int i)
+void RBEIMConstruction::init_context(FEMContext & c)
```
roystgnr, line 378: Ah, never mind.  My search just was turning up empty before because I hadn't yet hit "Load diff" on the big changes.

jwpeterson:
Thanks!

PR #2656, opened 2020-07-24
Link to libvtkCommonExecutionModel

roystgnr:
This is the solution suggested by Charles Taylor to VTK linking problems people have occasionally encountered:

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/moose-users/B-MoFSnFCKg/ipRdBmwnBAAJ

https://groups.google.com/g/moose-users/c/7lDXjHVnRgQ

I'm not sure what code is *selectively* trying to request those particular symbols; it might be internal to certain VTK configurations.  But this does seem to be the right library to add to fix them.

I'm just starting to build this myself, so hopefully there's no embarrassing mistakes, but I wanted it to upload it ASAP so someone who can reproduce the original problem can make sure this fixes it.

### DIFF
--- a/configure
+++ b/configure
@@ -38464,12 +38464,14 @@ else
                                                    VTK_LIBRARY_WITH_VERSION="-L$VTK_LIB -lvtkIOCore-$vtkmajorminor -lvtkCommonCore-$vtkmajorminor -lvtkCommonDataModel-$vtkmajorminor \
                                            -lvtkFiltersCore-$vtkmajorminor -lvtkIOXML-$vtkmajorminor -lvtkImagingCore-$vtkmajorminor \
                                            -lvtkIOImage-$vtkmajorminor -lvtkImagingMath-$vtkmajorminor -lvtkIOParallelXML-$vtkmajorminor \
-                                           -lvtkParallelMPI-$vtkmajorminor -lvtkParallelCore-$vtkmajorminor"
+                                           -lvtkParallelMPI-$vtkmajorminor -lvtkParallelCore-$vtkmajorminor \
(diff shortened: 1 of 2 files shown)


jwpeterson:
My only question would be whether this change actually applies to VTK-6.2 and above or if it's a more recently added library? That may not be easy to check, though.

roystgnr:
I've used both old and new VTK without problems.  So there's got to be some system- or configuration-dependent trigger for the bug.  But VTK 6.2 at least does have this library, so no harm supplying it to the linker just in case.

roystgnr:
Let's wait until we hear back from Charles Taylor before merging; it sure looked to me like the configure diff matched his patch, but I'd like to be certain.

ca-taylor:
I'd be happy to verify the commits for you but I'm not sure how to cherry pick them into my local branch.   It seems like I should just be able to cherry pick them from Roy's fork/clone but so far my attempts have failed.    I'm willing to learn though.

jwpeterson:
In your libmesh submodule (assuming you have ssh keys, otherwise modify the URLs appropriately):
```
git remote add roystgnr git@github.com:roystgnr/libmesh.git
git pull roystgnr vtk_c_e_m
```
This will merge the changes from Roy's branch into yours so it's not the same as cherry-picking... but simpler IMO.

ca-taylor:
I managed to figure out the cherry-pick approach but you are right.  You’re suggestion is *much* more straightforward.   

Thank you for the help.

Charlie

ca-taylor:
Roy's commits seem to have done the trick and I was able to update and rebuild libmesh without hitting the unresolved references.

roystgnr:
Thanks again!

PR #2658, opened 2020-07-24
Add geometric constants for EDGE elems

### LINKED ISSUE #2657: Add geometric constants for EDGE elem types
Add `num_sides`, `num_edges`, `nodes_per_side`, and `nodes_per_edge` to the EDGE derived types for consistency.

loganharbour:
Closes #2657

### DIFF
--- a/include/base/libmesh.h
+++ b/include/base/libmesh.h
@@ -245,9 +245,15 @@ const Number zero = 0.;
 
 /**
  * A number which is used quite often to represent
- * an invalid or uninitialized value.
+ * an invalid or uninitialized value for an unsigned integer.
  */
 const unsigned int invalid_uint = static_cast<unsigned int>(-1);
+
+/**
+ * A number which is used quite often to represent
+ * an invalid or uninitialized value for an integer.
+ */
+const int invalid_int = std::numeric_limits<int>::max();
 
 } // namespace libMesh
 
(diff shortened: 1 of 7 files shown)


loganharbour:
@roystgnr this is good.

I tried to find some edge methods that could use this to no avail.

jwpeterson:
Seems fine to me but I'm curious about the distinction between 0 and `invalid_int`? If we say that an Edge has `num_edges=0` (seems reasonable) could we also just say it has `nodes_per_edge=0` for simplicity?

loganharbour:
The thought here by @roystgnr was that if `num_edges = 0`, that `nodes_per_edge` should be some not-applicable value. Hence, the invalid_int.

I see reasons for both choices (invalid/0) - so I'll let you guys choose 😬

jwpeterson:
My main concern would be some code that accidentally creates an array of size `Edge2::nodes_per_edge` but since there's no virtual API/accessor for it I guess that's pretty unlikely.

roystgnr:
> My main concern would be some code that accidentally creates an array of size Edge2::nodes_per_edge

Ah, I see.  *My* concern had been code that's accidentally been called for the "edge of an edge", and making that loop forever rather than continue failing silently seemed like the best way to go.

I think anything trying to allocate an Elem2::nodes_per_edge array, but not as part of a loop over all 0 edges whose internals thereby don't get triggered, is probably likewise a "already broken so let's die ASAP where the breakage is easy to diagnose" case, in which case invalid_int is still the way to go, but I could be talked out of that.

PR #2660, opened 2020-07-25
Add libmesh_error_msg_if(cond, msg) macro

jwpeterson:
The idea here is to make error checking more self-contained while also allowing for the possibility of later adding a configure option which disables _all_ error checking.
    
Another option would have been to add:
```
libmesh_always_assert_msg(asserted, msg)
```    
but in that case in order to use the new macro we'd have to reverse the logic of all the checks that are already present in the code, and that seemed too error prone to me.

So far I have updated the files in src/mesh to use the new macro. If all goes well, I will eventually update the rest of the code.

### DIFF
--- a/include/base/libmesh_common.h
+++ b/include/base/libmesh_common.h
@@ -430,6 +430,12 @@ struct casting_compare {
 
 #define libmesh_error() libmesh_error_msg("")
 
+#define libmesh_error_msg_if(cond, msg)         \
+  do {                                          \
+    if (cond)                                   \
+      libmesh_error_msg(msg);                   \
+  } while (0)
+
 #define libmesh_exceptionless_error_msg(msg)                            \
   do {                                                                  \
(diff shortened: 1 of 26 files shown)


roystgnr:
My argument for always_assert would be that, laziness being what it is, we probably have a lot more libmesh_assert(foo) that should be turned on always than we have if (foo) error() that should be simplified.  But I'm okay either way.

jwpeterson:
> But I'm okay either way.

OK, thanks for reviewing. I think there's a case to be made for having both macros actually, if only because I don't trust myself to DeMorgan-ize complicated compound logical expressions to convert from an `assert` to an `error_if`, especially in a big PR. 

I also think that because every `libmesh_assert` -> `libmesh_always_assert` change can have performance implications, it should therefore be looked at closely, whereas in this PR I'm just making existing error checks a bit more terse.

PR #2662, opened 2020-07-28
Use libmesh_error_msg_if() where appropriate

jwpeterson:
Refs #2660.

### DIFF
--- a/src/apps/compare.C
+++ b/src/apps/compare.C
@@ -116,10 +116,9 @@ void process_cmd_line(int argc,
            */
         case 'm':
           {
-            if (names.empty())
-              names.push_back(optarg);
-            else
-              libmesh_error_msg("ERROR: Mesh file name must precede left file name!");
+            libmesh_error_msg_if(!names.empty(), "ERROR: Mesh file name must precede left file name!");
+
+            names.push_back(optarg);
             break;
           }
 
@@ -137,13 +136,10 @@ void process_cmd_line(int argc,
            */
         case 'l':
           {
-            if (!left_name_set)
-              {
-                names.push_back(optarg);
-                left_name_set = true;
-              }
-            else
-              libmesh_error_msg("ERROR: Mesh file name must precede right file name!");
+            libmesh_error_msg_if(left_name_set, "ERROR: Mesh file name must precede right file name!");
+
+            names.push_back(optarg);
+            left_name_set = true;
             break;
           }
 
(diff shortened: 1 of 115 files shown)


src/apps/compare.C
```diff
@@ -152,10 +148,10 @@ void process_cmd_line(int argc,
            */
         case 'r':
           {
-            if ((!names.empty()) && (left_name_set))
-              names.push_back(optarg);
-            else
-              libmesh_error_msg("ERROR: Mesh file name and left file name must precede right file name!");
+            libmesh_error_msg_if(names.empty() || !left_name_set,
```
roystgnr, line 151: > [I don't trust myself](https://github.com/libMesh/libmesh/pull/2660#issuecomment-664539859) to DeMorgan-ize complicated compound logical expressions

I thought that line was a good explanation for your decision; now I fear it was also foreshadowing...

src/systems/fem_system.C
```diff
@@ -1392,8 +1392,7 @@ void FEMSystem::mesh_position_get()
 {
   // This function makes no sense unless we've already picked out some
   // variable(s) to reflect mesh position coordinates
-  if (!_mesh_sys)
-    libmesh_error_msg("_mesh_sys was nullptr!");
```
roystgnr, line 1396: In hindsight I probably should have made this an assert.

src/apps/compare.C
jwpeterson: Not gonna lie, I have the [wikipedia page](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) open in one of my tabs right now.

jwpeterson:
Thanks for reviewing!

src/reduced_basis/rb_eim_evaluation.C
```diff
@@ -292,10 +280,8 @@ Number RBEIMEvaluation::get_eim_basis_function_value(unsigned int basis_function
                                                      unsigned int comp,
                                                      unsigned int qp) const
 {
-  if (basis_function_index >= _local_eim_basis_functions.size())
-  {
-    libmesh_error_msg("Invalid basis function index: " + std::to_string(basis_function_index));
-  }
+  libmesh_error_msg_if(basis_function_index >= _local_eim_basis_functions.size(),
+                       "Invalid basis function index: " + basis_function_index);
```
roystgnr, line 284: You dropped the std::to_string() here.

Perhaps more alarmingly, this made it through CI!?!  Even the --enable-paranoid-warnings --enable-werror runs!?!

It looks like ```-Wstring-plus-int``` is a clang warning but not yet a gcc warning, and we only use --enable-werror on gcc builds.  Right now we're in "we'll take what we can get" mode on CI, but after that's straightened out we ought to just turn on --enable-paranoid-warnings --enable-werror on every CI build.

jwpeterson: Ha, I meant to change the` +` into a `<<`, I'll push that now...

PR #2664, opened 2020-07-30
Clone ghosting functors 

fdkong:
https://github.com/idaholab/moose/pull/15655

Clone ghosting functors when we clone mesh because the operations in ghosting functors are mesh dependent.

### DIFF
--- a/include/base/default_coupling.h
+++ b/include/base/default_coupling.h
@@ -22,6 +22,7 @@
 
 // Local Includes
 #include "libmesh/ghosting_functor.h"
+#include "libmesh/auto_ptr.h" // libmesh_make_unique
 
 namespace libMesh
 {
@@ -54,6 +55,26 @@ public:
     _n_levels(0)
   {}
 
+  /**
+   * Constructor.
+   */
+  DefaultCoupling(const DefaultCoupling & other) :
+    GhostingFunctor(other),
+    _dof_coupling(other._dof_coupling),
+#ifdef LIBMESH_ENABLE_PERIODIC
+    _periodic_bcs(other._periodic_bcs),
+#endif
+    _mesh(other._mesh),
+    _n_levels(other._n_levels)
+  {}
+
+  /**
+   * A clone() is needed because GhostingFunctor can not be shared between
+   * different meshes. The operations in  GhostingFunctor are mesh dependent.
+   */
+  virtual std::unique_ptr<GhostingFunctor> clone () const override
+  { return libmesh_make_unique<DefaultCoupling>(*this); }
+
   // Change coupling matrix after construction
   void set_dof_coupling(const CouplingMatrix * dof_coupling);
 
@@ -72,7 +93,8 @@ public:
 #endif
 
   // Set MeshBase for use in checking for periodic boundary ids
-  void set_mesh(const MeshBase * mesh)
+  // It should be called if a GhostingFunctor is created via clone
+  virtual void set_mesh(const MeshBase * mesh) override
   { _mesh = mesh; }
 
   /**
--- a/include/base/default_coupling.h
+++ b/include/base/default_coupling.h
@@ -51,7 +51,6 @@ public:
 #ifdef LIBMESH_ENABLE_PERIODIC
     _periodic_bcs(nullptr),
 #endif
-    _mesh(nullptr),
     _n_levels(0)
   {}
 
@@ -64,7 +63,6 @@ public:
 #ifdef LIBMESH_ENABLE_PERIODIC
     _periodic_bcs(other._periodic_bcs),
 #endif
-    _mesh(other._mesh),
     _n_levels(other._n_levels)
   {}
 
@@ -91,11 +89,6 @@ public:
   void set_periodic_boundaries(const PeriodicBoundaries * periodic_bcs)
   { _periodic_bcs = periodic_bcs; }
 #endif
-
-  // Set MeshBase for use in checking for periodic boundary ids
-  // It should be called if a GhostingFunctor is created via clone
-  virtual void set_mesh(const MeshBase * mesh) override
-  { _mesh = mesh; }
 
   /**
    * If we have periodic boundaries, then we'll need the mesh to have
@@ -130,7 +123,6 @@ private:
 #ifdef LIBMESH_ENABLE_PERIODIC
   const PeriodicBoundaries * _periodic_bcs;
 #endif
-  const MeshBase * _mesh;
   unsigned int _n_levels;
 };
 
--- a/include/base/ghosting_functor.h
+++ b/include/base/ghosting_functor.h
@@ -25,6 +25,7 @@
 #include "libmesh/id_types.h"
 #include "libmesh/mesh_base.h"
 #include "libmesh/reference_counted_object.h"
+#include "libmesh/auto_ptr.h" // libmesh_make_unique
 
 // C++ Includes
 #include <unordered_map>
@@ -160,9 +161,27 @@ public:
   GhostingFunctor() {}
 
   /**
+   * Copy Constructor
+   */
+  GhostingFunctor(const GhostingFunctor & other) : ReferenceCountedObject<GhostingFunctor>(other) {}
+
+  /**
    * Virtual destructor; this is an abstract base class.
    */
   virtual ~GhostingFunctor() {}
+
+  /**
+   * A clone() is needed because GhostingFunctor can not be shared between
+   * different meshes. The operations in  GhostingFunctor are mesh dependent.
+   */
+  virtual std::unique_ptr<GhostingFunctor> clone () const
+  { libmesh_not_implemented(); }
+
+  /**
+   * It should be called after cloning a ghosting functor.
+   * Ghosting functor is mesh dependent
+   */
+  virtual void set_mesh(const MeshBase * /*mesh*/) { libmesh_not_implemented(); }
 
   /**
    * What elements do we care about and what variables do we care
--- a/include/base/ghosting_functor.h
+++ b/include/base/ghosting_functor.h
@@ -158,12 +158,20 @@ public:
   /**
    * Constructor.  Empty in the base class.
    */
-  GhostingFunctor() {}
+  GhostingFunctor(): _mesh(nullptr) {}
+
+  /**
+   * Constructor using mesh
+   */
+  GhostingFunctor(const MeshBase & mesh): _mesh(&mesh) {}
 
   /**
    * Copy Constructor
    */
-  GhostingFunctor(const GhostingFunctor & other) : ReferenceCountedObject<GhostingFunctor>(other) {}
+  GhostingFunctor(const GhostingFunctor & other) :
+   ReferenceCountedObject<GhostingFunctor>(other),
+   _mesh (other._mesh)
+  {}
 
   /**
    * Virtual destructor; this is an abstract base class.
@@ -175,13 +183,21 @@ public:
    * different meshes. The operations in  GhostingFunctor are mesh dependent.
    */
   virtual std::unique_ptr<GhostingFunctor> clone () const
-  { libmesh_not_implemented(); }
+  // Let us return nullptr for backward compatibility.
+  // We will come back to mark this function as pure virtual
+  // once the API upgrade is done.
+  { return nullptr; }
 
   /**
    * It should be called after cloning a ghosting functor.
    * Ghosting functor is mesh dependent
    */
-  virtual void set_mesh(const MeshBase * /*mesh*/) { libmesh_not_implemented(); }
+  virtual void set_mesh(const MeshBase * mesh) { _mesh = mesh; }
+
+  /**
+   *  Return the mesh associated with ghosting functor
+   */
+  const MeshBase * get_mesh() const { return _mesh; }
 
   /**
    * What elements do we care about and what variables do we care
@@ -238,6 +254,9 @@ public:
    * after a redistribution is complete.
    */
   virtual void delete_remote_elements () {};
+
+protected:
+  const MeshBase * _mesh;
 };
 
 } // namespace libMesh
--- a/include/base/ghosting_functor.h
+++ b/include/base/ghosting_functor.h
@@ -197,7 +197,7 @@ public:
   /**
    *  Return the mesh associated with ghosting functor
    */
-  const MeshBase * get_mesh() const { return _mesh; }
+  const MeshBase * get_mesh() const { libmesh_assert(_mesh); return _mesh; }
 
   /**
    * What elements do we care about and what variables do we care
--- a/src/base/default_coupling.C
+++ b/src/base/default_coupling.C
@@ -78,7 +78,12 @@ void DefaultCoupling::operator()
 {
   LOG_SCOPE("operator()", "DefaultCoupling");
 
-  libmesh_assert(_mesh);
+  // Let us not do assertion at this moment for API upgrade.
+  // There is a functor inside of ElementSideNeighborLayers.
+  // We can not set mesh for that functor because there is no handle
+  // in libmesh. We need to override set_mesh in moose for setting a mesh for the functor.
+  // The set_mesh overridden will not happen until the current change gets in.
+  //libmesh_assert(_mesh);
 
 #ifdef LIBMESH_ENABLE_PERIODIC
   bool check_periodic_bcs =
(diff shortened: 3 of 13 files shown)


fdkong:
It is ready for review @roystgnr

fdkong:
Oh, at least `precheck` passed. It was not bad at all

roystgnr:
We've all been there.

But this is actually making it to the hp refinement case in adaptivity_ex3, in the first CI recipe I checked, so you've got to be almost there.  Let me see if I can replicate the problem.

include/base/ghosting_functor.h
```diff
@@ -159,11 +160,29 @@ class GhostingFunctor : public ReferenceCountedObject<GhostingFunctor>
...
    * Virtual destructor; this is an abstract base class.
    */
   virtual ~GhostingFunctor() {}
 
+  /**
+   * A clone() is needed because GhostingFunctor can not be shared between
+   * different meshes. The operations in  GhostingFunctor are mesh dependent.
+   */
+  virtual std::unique_ptr<GhostingFunctor> clone () const
+  { libmesh_not_implemented(); }
```
roystgnr, line 178: This is your problem.  As soon as we hit a subclass that hasn't be modified in sync with this API change and the next, we'll scream and die.  Better to make clone() a pure virtual function so at least the failure happens at compile time (in this case, as soon as you hit include/base/sibling_coupling.h) instead of runtime!

On the other hand, it occurs to me that the MOOSE update here is going to be a lot smoother than I'd expected.  There's nothing stopping us from implementing clone() in the RelationshipManager subclasses there *before* we do the libMesh update, and the result will be compatible with libMesh both pre- and post- this PR; we just can't mark those implementations as override until after this PR gets merged.

include/base/ghosting_functor.h
```diff
@@ -159,11 +160,29 @@ class GhostingFunctor : public ReferenceCountedObject<GhostingFunctor>
...
+
+  /**
+   * It should be called after cloning a ghosting functor.
+   * Ghosting functor is mesh dependent
+   */
+  virtual void set_mesh(const MeshBase * /*mesh*/) { libmesh_not_implemented(); }
```
roystgnr, line 184: Let's just move the mesh pointer that so many subclasses use into this class, and then we can make this non-virtual?

include/base/ghosting_functor.h
roystgnr: Oh, and although of course the MOOSE-based CI in this PR won't pass until those subclass clone methods are merged to MOOSE, let's save that until after we've got the code here passing libMesh-only tests and reviewed; it would be silly to merge something into MOOSE only to risk realizing afterwards that we need a slight design change, like my suggestion about set_mesh() below.

Oh, and speak of the devil:  We should have a set_mesh() marked virtual in this class, shouldn't we, even though we'll be able to implement it here and that implementation will be good enough for most/all subclasses today?  There may be subclasses that want to override that implementation to re-cache mesh-based data when needed, and it's not a function that will ever be called in any inner loops so the virtual overhead should be negligible.

include/base/ghosting_functor.h
roystgnr: Well, not non-virtual, but at least we can give it a good default implementation.

include/base/ghosting_functor.h
fdkong: OK, we have to implement clone correctly in libmesh before my MOOSE PR can get in since a mesh could have both RMs and libmesh style ghosting functors. I can not miss any of them in the displaced mesh.

fdkong: Oh, never mind.

fdkong:
I could see it is very tricky to update APIs here.  We need one back-compability implementation: -)

include/base/ghosting_functor.h
fdkong: I intended to put `libmesh_not_implemented` to force us to update downstream but at the same time, I do not need to update all of them at once.

I am not sure what a good default implementation looks like. We may just move up mesh here since most of the functors are mesh-dependent. 

We might have something like this in `ghosting_funcgor.`h:?

```
virtual void set_mesh(const MeshBase * mesh) { _mesh = mesh; }

const mesh * getMesh() const { return _mesh; }

private:

const MeshBase * _mesh;
```

include/base/ghosting_functor.h
fdkong: In order to have a back-compatibility, I think we might do this in `ghosting_functor.h `:

```
  virtual std::unique_ptr<GhostingFunctor> clone () const
  {  return nullprt; }
```

and then do this in `meh_base.C`:


```
   for (const auto & gf : other_mesh._ghosting_functors )
    {
      std::shared_ptr<GhostingFunctor> clone_gf = gf->clone();
      if (clone_gf)
        {
           clone_gf->set_mesh(this);
           add_ghosting_functor(clone_gf);
        } else 
       {
           add_ghosting_functor(gf);
       }
    }
```

include/base/ghosting_functor.h
roystgnr: Exactly.  But make it get_mesh; don't mix snake-case for the setter and camel-case for the getter!

We can then upgrade the libMesh subclasses with mesh pointers, and then updating MOOSE subclasses can come after the libMesh PR is merge.

include/base/ghosting_functor.h
fdkong: Once the integration is done, we could come back and change this function to pure virtual

fdkong:
The PR is backward compatible now. It is ready for a serious review. 

@roystgnr

src/mesh/mesh_base.C
```diff
@@ -102,9 +102,27 @@ MeshBase::MeshBase (const MeshBase & other_mesh) :
...
   _point_locator_close_to_point_tol(other_mesh._point_locator_close_to_point_tol)
 {
+   for (const auto & gf : other_mesh._ghosting_functors )
+   {
+     std::shared_ptr<GhostingFunctor> clone_gf = gf->clone();
+     // Some subclass of GhostingFunctor does not implement clone
+     // GhostingFunctor will return nullptr by default.
+     // It is written for APIs upgrade
+     // Once moose is updated, we should come back
+     // and change this code
```
roystgnr, line 114: This is extremely clever; good job!  I'd say let's throw in a libmesh_deprecated() in the else case, then we can leave it for a release or two and actually give everyone else a smooth upgrade path too.

fdkong:
OK, let me check if I can test my MOOSE PR https://github.com/idaholab/moose/pull/15655 without merging this into master.

fdkong:
I can not even test my PR if the libmesh change is not in master yet https://civet.inl.gov/job/578314/

fdkong:
@lindsayad @jwpeterson Do you have any suggestions on this PR?

roystgnr:
Civet thinks there are merge conflicts blocking use of this hash in MOOSE?  I guess you could rebase this on master and then try again.

But no, let's just merge this as soon as CI is happy, and not hold up that MOOSE PR.  This looks great to me, and although it's slightly different than the ideas I ran by John before he went on vacation, I think the differences are all clear improvements.

include/base/ghosting_functor.h
```diff
@@ -157,13 +158,47 @@ class GhostingFunctor : public ReferenceCountedObject<GhostingFunctor>
...
+   * Constructor using mesh
+   */
+  GhostingFunctor(const MeshBase & mesh): _mesh(&mesh) {}
+
+  /**
+   * Copy Constructor
+   */
+  GhostingFunctor(const GhostingFunctor & other) :
+   ReferenceCountedObject<GhostingFunctor>(other),
+   _mesh (other._mesh)
```
lindsayad, line 173: This was the "bug" that incited this whole PR right? Would it even make sense to clone the same ghosting functor onto the same mesh? would it be safer to leave `_mesh` as `nullptr` and then put in protective asserts in places where `_mesh` is used? (If they're not already there, they should be anyways right?)

lindsayad: I get that in the code you've added you immediately call `set_mesh` after `clone`, but maybe someone else will use this in a different way in the future. Unlikely, but I'm always paranoid

fdkong: Great point!!! 

If it is not too evil to assume that a ghosting functor highly likely has  a mesh, then it might not be a bad idea to have clone function like:

`std::unique_ptr<GhostingFunctor> clone (const MeshBase * mesh) `

But this will contradict the definition of `clone`?? Possibly not.

```
  GhostingFunctor(const GhostingFunctor & other, const MeshBase * mesh) :
   ReferenceCountedObject<GhostingFunctor>(other),
   _mesh (mesh)
```

roystgnr: > Would it even make sense to copy/clone the same ghosting functor onto the same mesh?

As a use case, usually no, though I could see a case where the same general functor might be copied and then the two copies given different (but neither superceding the other) settings.

As a general principle, though, I think it's the only thing that makes sense - when you make a copy of an object you want it to be as indistinguishable from the original as possible until/unless you start changing things afterwards.

> put in protective asserts in places where _mesh is used?

This, on the other hand, is such a good idea that I feel stupid for not having done so already.  I'd love it if, for any ghosting functor subclass that uses ```_mesh``` (and maybe even the ones that don't!), whenever the functor receives a non-empty range, it does something like ```libmesh_assert(_mesh.query_elem_ptr((*it)->id()) == *it);``` on each element ```*it``` in the range.  If we'd had that to begin with then it wouldn't have taken so long to track this bug down.

include/base/default_coupling.h
```diff
@@ -50,10 +51,28 @@ class DefaultCoupling : public GhostingFunctor
...
   {}
 
+  /**
+   * Constructor.
+   */
+  DefaultCoupling(const DefaultCoupling & other) :
+    GhostingFunctor(other),
+    _dof_coupling(other._dof_coupling),
+#ifdef LIBMESH_ENABLE_PERIODIC
+    _periodic_bcs(other._periodic_bcs),
```
roystgnr, line 64: On second glance, this shallow ```PeriodicBoundaries *``` copy (here and elsewhere) is an issue, isn't it?  other._periodic_bcs, if it exists, is pointing to the periodic boundary conditions associated with an old system, and if those get changed or if different PBCs get used with the new system, then we're out of sync with what we'd intended.

I don't think that should hold up this PR (which is at least a huge improvement, and should be all the improvement we need for the displaced mesh case in MOOSE) but it's at least a problem to puzzle over.

fdkong: I was thinking about this loudly easier. I should add some comments here. For the displaced system and the reference system, we are using the same PBC. I am not advocating it is a great idea. Here is some code from `AddPeriodicBCAction`

   ```
   if (component >= 0)
      {
        const std::pair<BoundaryID, BoundaryID> * boundary_ids =
            _mesh->getPairedBoundaryMapping(component);
        RealVectorValue v;
        v(component) = _mesh->dimensionWidth(component);
        PeriodicBoundary p(v);

        if (boundary_ids == NULL)
          mooseError(
              "Couldn't auto-detect a paired boundary for use with periodic boundary conditions");

        p.myboundary = boundary_ids->first;
        p.pairedboundary = boundary_ids->second;
        setPeriodicVars(p, getParam<std::vector<VariableName>>("variable"));
        nl.dofMap().add_periodic_boundary(p);
        if (displaced_problem)
          displaced_problem->nlSys().dofMap().add_periodic_boundary(p);
      }


```

I should put some comments here before we figure out an acceptable solution.

Comment (fdkong) [reply include/base/default_coupling.h:64]: Possibly need a clone for PBC?

File: include/base/ghosting_functor.h (hunk 4 above)
Comment (fdkong) [reply include/base/ghosting_functor.h:173]: Yes, I might add these changes

File: include/base/ghosting_functor.h (hunk 1 above)
Comment (roystgnr) [reply include/base/ghosting_functor.h]: I think the cleverness here means we can afford to wait on going pure virtual - let's let a release get out where the old API just screams ```libmesh_deprecated();``` then we can go to pure virtual for the release after that.

File: include/base/default_coupling.h (hunk 5 above)
Comment (roystgnr) [reply include/base/default_coupling.h:64]: Looks like we already have a clone for PeriodicBoundary itself (IIRC the Akselos folks needed a subclass to do weirder transformations?), so it should be easy to add a whole PeriodicBoundaries::clone().

Trouble is, that might not be what we want.  It's at least not a complete solution.  Typically PeriodicBoundaries-aware functors are associated with the DofMap of the system with periodic variables, not with the underlying Mesh.  If we make a cloned copy of the PeriodicBoundaries associated with system1, then that'll solve the half of the problem where we don't want subsequent changes in system1's PBCs to affect ghosting functors on system2, but it won't solve the half of the problem where we do want subsequent changes in system2's PBCs to affect its ghosting functors, because the cloned PeriodicBoundaries won't be the same as the set that gets attached to system2 via it's DofMap::add_periodic_boundary().

Comment (fdkong) [reply include/base/default_coupling.h:64]: Yes, we have to leave it as-is for a moment. "as-is" (sharing PBC across systems) is what moose uses right now. 

But should we at least create an issue on this? We could revisit it when we think we are smart :-)

File: src/base/default_coupling.C [hunk 6]
```diff
@@ -78,7 +78,12 @@ void DefaultCoupling::operator()
 {
   LOG_SCOPE("operator()", "DefaultCoupling");
 
-  libmesh_assert(_mesh);
+  // Let us not do assertion at this moment for API upgrade.
+  // There is a functor inside of ElementSideNeighborLayers.
+  // We can not set mesh for that functor because there is no handle
+  // in libmesh. We need to override set_mesh in moose for setting a mesh for the functor.
```
roystgnr, line 84: Ah - set_mesh() is getting called by MOOSE *only* if _app.getExecutioner() returns non-null?

So we can leave the check that comes later in the get_periodic_bcs case, because that's actually checking if something's broken, but the check here would be overzealous?  Makes sense to me.

include/base/default_coupling.h
roystgnr: [Issue created.](https://github.com/libMesh/libmesh/issues/2670)

> We could revisit it when we think we are smart :-)

Gonna need a time machine for that.

When I was 11 taking honors Algebra I, I thought I was smart.  Eventually that sort of thing got me to places where I wasn't  the smartest kid in the building, then to where I wasn't the smartest student in the class, then to where I wasn't the smartest researcher in the group ... at this point I find it safest to assume I'm not the smartest person in any given conversation.  It's most demoralizing that I haven't been able to make an exception for talking to my kids.

roystgnr:
This looks ready to merge to me.  Anything else we might want to add or change first?

src/base/default_coupling.C
lindsayad: So we will be able to uncomment this eventually though right?

src/base/default_coupling.C
```diff
@@ -93,8 +100,11 @@ void DefaultCoupling::operator()
   if (!this->_n_levels)
     {
       for (const auto & elem : as_range(range_begin, range_end))
+      {
+        //libmesh_assert(_mesh->query_elem_ptr(elem->id()) ==elem);
```
lindsayad, line 104: And same here? Right now MOOSE would actually fail this assertion because of our badness with displaced mesh?

fdkong: > So we will be able to uncomment this eventually though right?

Yes, when the MOOSE PR is merged https://github.com/idaholab/moose/pull/15655

fdkong: The displaced mesh and the reference mesh need to be synced correctly in a geometric sense. `Default_coupling` seems to be fine even when they are inconsistent, but it is still a good idea to make then consistent.

src/base/default_coupling.C
fdkong: Yes, we will uncomment all these.

include/base/default_coupling.h
fdkong: LOL

fdkong:
> This looks ready to merge to me. Anything else we might want to add or change first?

I do not have anything to add. If you want to change something, just let me know.

lindsayad:
This passed all testing huh? How did that happen? Our `RelationshipManager` class inherits from `GhostingFunctor` and it already has a `_mesh` member of a different type

fdkong:
> This passed all testing huh? How did that happen? Our `RelationshipManager` class inherits from `GhostingFunctor` and it already has a `_mesh` member of a different type

RM just overwrites. RM uses what it has

PR #2665, opened 2020-07-31
External Parmetis support

roystgnr:
This seems to be necessary for me to build with Ubuntu 20.04 PETSc, which links to the system Metis, meaning I can't build contrib metis or parmetis without conflicts.

This makes our parmetis.m4 a little more similar to metis.m4 to boot, but with more checks to keep the tricky "PETSc has Metis but not Parmetis so we can't use the latter" cases handled.

Would very much appreciate more eyes on this; it works for me and we'll see if it works for Civet but that's still just 2 out of the half dozen categories of environments we'd like to keep supporting here.

### DIFF
--- a/m4/parmetis.m4
+++ b/m4/parmetis.m4
@@ -12,6 +12,55 @@ AC_DEFUN([CONFIGURE_PARMETIS],
                          [AC_MSG_ERROR(bad value ${enableval} for --enable-parmetis)])],
                 [enableparmetis=$enableoptional])
 
+  AC_ARG_WITH(parmetis,
+             AS_HELP_STRING([--with-parmetis=<internal,PETSc,/some/libdir>],
+                            [internal: build from contrib, PETSc: rely on PETSc]),
+             [AS_CASE("${withval}",
+                      [internal], [build_parmetis=yes],
+                      [PETSc],    [build_parmetis=petsc],
+                      [PARMETIS_LIB="-L${withval} -lparmetis"
+                       build_parmetis=no])
+              enableparmetis=yes],
+             [build_parmetis=yes])
+
+  AC_ARG_WITH(parmetis-include,
+             AS_HELP_STRING([--with-parmetis-include=</some/includedir>]),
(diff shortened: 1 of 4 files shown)


m4/parmetis.m4
```diff
@@ -12,63 +12,79 @@ AC_DEFUN([CONFIGURE_PARMETIS],
...
+                build_parmetis=petsc],
+               [AC_MSG_RESULT(<<< Disabling ParMETIS support to avoid PETSc conflict>>>)
+                enableparmetis=no])
+        ])
+
+  dnl Conversely, if:
+  dnl .) ParMETIS is enabled in libmesh,
+  dnl .) PETSc does not have a ParMETIS or METIS, or we aren't using PETSc, and
+  dnl .) build_parmetis=petsc because user said --with-parmetis=PETSc,
+  dnl then we need to make sure that libmesh builds its own ParMETIS!
```
jwpeterson, line 54: Holy moly! In our builds we disable ParMETIS because it has a different (and less LGPL-compatible) license than METIS, so I should be able to pull your branch and test that configuration...

jwpeterson:
The current error looks like an unset environment variable during linking:
```
g++: error: /libparmetis: No such file or directory
```
Not sure why it is specific to the complex numbers case, although PETSc is disabled in that build so that might be a clue...

m4/parmetis.m4
roystgnr: Okay, this should be ready for that test.  No rush, though; there's only one system I'm using external Parmetis on right now and I can just rebase branches there on this for the near future.

jwpeterson: Finally got around to trying this with an `--enable-petsc --with-metis=PETSc --disable-parmetis` build and looks like it works fine. I did not test what I believe is the more challenging case, with system installs of METIS/Parmetis that could conflict with stuff, but it sounds like that is the case which motivated this Issue for you in the first place, so I assume it's working as well...

PR #2666, opened 2020-07-31
Inline a more efficient implementation of BoundingBox::intersects.

drwells:
This ends up being about 6x faster due to inlining and improved short circuiting.


jwpeterson:
Hmm... any idea how much of the performance improvement is due to the inlining and how much is due to the revamped logic? I generally prefer to avoid having implementations in headers since it can introduce additional compilation dependencies, etc. but for demonstrable performance improvements, it's worthwhile.

drwells:
I'll check.

include/geom/bounding_box.h
```diff
@@ -147,6 +147,60 @@ class BoundingBox : public std::pair<Point, Point>
...
+{
+  BoundingBox expanded_my_box = *this;
+  BoundingBox expanded_other_box = other_box;
+  for (unsigned int dir=0; dir<LIBMESH_DIM; ++dir)
+    {
+      expanded_my_box.first(dir) -= abstol;
+      expanded_my_box.second(dir) += abstol;
+      expanded_other_box.first(dir) -= abstol;
+      expanded_other_box.second(dir) += abstol;
+    }
```
roystgnr, line 200: Should we only be expanding one box or the other?  I think that'll match the previous abstol behavior better, and it'll save a few flops.

roystgnr:
No additional header dependencies here, since we already include point.h for RVO.  I agree that if we can get most of the speedup without the inlining then I'd prefer to settle for that and leave the definition in the .C, but 6x is huge and I'm betting at least half of that is from the inlining.

include/geom/bounding_box.h
drwells: You're absolutely right: I'll fix this.

jwpeterson:
> I'm betting at least half of that is from the inlining

I'm curious about that... in 3D we used to check all the dimensions before returning, now there is a possibility of detecting an intersection sooner.

roystgnr:
> in 3D we used to check all the dimensions before returning, now there is a possibility of detecting non-intersection sooner.

Sure, but on the other hand the loop over dimensions used to be unrollable to a fixed length but now it definitely needs a test-and-branch at each dimension.  Might be a wash.

Or might be a wash with some processors and compiler settings and a win on others.  IIRC "how badly do branches screw up your instruction pipelining" is a question whose answer has swung back and forth over the years.

jwpeterson:
After reading a bit more (e.g. this [SO answer](https://stackoverflow.com/a/1759575/659433)) I finally understood it's not so much the `inline` keyword itself but the fact that the compiler has the full function definition available at compile time that allows it to optimize better. I'm now leaning towards your "at least half of that" estimate more strongly than before.

drwells:
Yup. Measuring these things is a bit tricky - I am using callgrind to count instructions, which I believe is a very good measure of actual performance. In particular I am measuring the way we call this function in [IBTK::FEDataManager::collectActivePatchElements()](https://github.com/IBAMR/IBAMR/blob/master/ibtk/src/lagrangian/FEDataManager.cpp#L2872). This is important since if the function is inlined I can no longer measure how long it takes directly.

- With the `master` branch of libMesh: `BoundingBox::intersects` is measured to take 178.4k instructions.
- With this patch, but not inlined: `BoundingBox::intersects` is measured to take 36.7k instructions.
- With this patch (so inlined): the calling function requires 14.9k fewer instructions, so I estimate this version takes 21.8k instructions.

Hence it looks like the inlined version is actually about 9x faster, and roughly a third of that comes from actually inlining the function. My original benchmark used a much larger mesh - I'm not sure why the function ended up being even cheaper with this benchmark but its still much faster than the original.

Nevertheless - would you prefer to move the definition back to the source file?

jwpeterson:
Thanks for checking, these are interesting results. IMO it's worth keeping inlined especially since you have a real world use case that benefits. I think I will add a comment that the function is inlined for performance reasons before merging, just so I don't forget later and try moving it to the C file... but there's no reason to trigger the CI again just for that small change.

drwells:
well, I also need to fix the fuzzy version :)

I'll force push in a few minutes.

jwpeterson:
@roystgnr Looks like overall improvement is 85% algorithm, 15% inlining, based on these numbers?

roystgnr:
> not so much the inline keyword itself but the fact that the compiler has the full function definition available at compile time that allows it to optimize better

Yup.  The process of inlining itself avoids stack buildup/teardown at the cost of bloating your assembled code, so inlining alone is only worth it for tiny functions.  But making it possible for the compiler to double-check that a function call has no side effects (like, *really* has no side effects, not just "the user said const and maybe he's not using const_cast or mutable to belie that"), or for the compiler to pull repeated calculations from inside a function to the outside of a loop calling that function, etc, can be big deals.

> I am using callgrind to count instructions, which I believe is a very good measure of actual performance.

In this context, where we're not making any significant change to memory access patterns, I think so.  In general it's sadly possible (and increasingly likely) for your instruction count to be meaningless because your CPU spends most of its time spinning it's wheels waiting for cache fetches.  You can still profile cache misses (with one set of simulated-cache options or another) with callgrind, but I think it's a non-default option.

> Looks like overall improvement is 90% algorithm, 10% inlining, based on these numbers?

Yeah.  In hindsight I would have been smarter to expect "50/50" improvement *geometrically*, not arithmetically, and that would have looked more like 75%/25% when measured algorithm-first like this ... but 90%/10% is still definitely more than I'd have ever imagined.  Which is reassuring, I guess: ever since I got a good speedup by specially-inlining the 1D Lagrange shape function evaluations, I've been wondering whether I ought to be hunting much harder for other should-have-been-inlined cases.

roystgnr:
> IMO it's worth keeping inlined especially since you have a real world use case that benefits

Agreed.

include/geom/bounding_box.h
roystgnr: Weird; I'd have expected Github to have marked this conversation as outdated after the force push.

drwells:
You're welcome. It's nice to move improvements upstream!

PR #2667, opened 2020-08-02
Updates to RB EIM code to enable further vectorization

dknez:


dknez:
This is a follow-up to the recent PR RB_EIM_refresh.pr2. This follow-up adds a new method, RBEIMEvaluation::rb_eim_solves(), and makes some minor changes to methods that are used internally by RBEIMConstruction and RBEIMEvaluation. This doesn't affect user code, e.g. reduced_basis_ex4 and reduced_basis_ex6 are unchanged here.

@jwpeterson is on vacation this week. @roystgnr would you like to review this, or else let me know if we should go ahead and merge.

dknez:
Yes, good advice, and I agree! After this point I think it should be stabilized, but if there are further changes we will start scattering libmesh_experimental(), as you suggested.

Thanks for checking! I'll go ahead and merge now.

PR #2671, opened 2020-08-06
Fixed node balancing algorithms

fdkong:


fdkong:
Do you have any comment? @roystgnr

roystgnr:
Are you kidding me?  This was a 4-character fix!?  That's wonderful!

We really need some kind of regression testing if we don't want this to decay again in the future, but the libMesh testing philosophy (tests need to work on any number of processors, and with any partitioner, but *maybe* you can require ReplicatedMesh if you're really lazy) doesn't make it easy to test this sort of thing.  Can you add something to MOOSE later?

roystgnr:
(merging because even if I didn't trust your private testing, this diff clearly isn't going to make the situation any *worse*, even though I'd still love to see tests in CI proving that we're making things better)

fdkong:
> Can you add something to MOOSE later?

https://github.com/idaholab/moose/pull/15722

fdkong:
> though I'd still love to see tests in CI proving that we're making things better)

Tests in idaholab/moose#15722 will fail without this PR. So I think it  makes things better.

But keep in mind, I do not trust myself most of the time :-)

PR #2672, opened 2020-08-07
Take out of libmesh_assert for an API integration

fdkong:
https://github.com/idaholab/moose/pull/15727


fdkong:
When something was wrong, it has to be wrong for a while before we can it right.

lindsayad:
Are you channeling some inner Yoda?

mangerij:
ask not what the code does, but what the code was intended to do
________________________________

fdkong:
@roystgnr Could you click the button to get my evil in?

PR #2674, opened 2020-08-12
Subdomain expansion projection fixes

roystgnr:
Previously, when reaching System::project_vector() after expanding a subdomain with a subdomain-restricted variable, we were getting segfaults.  We should now be getting correct projections of old DoFs combined with zeroing of new DoFs.

Thanks to @dewenyushu for setting up a test case demonstrating this problem.  And for catching this problem in the first place.  And for listening to me when I recommended trying this because I wasn't expecting any problems ...

I'm not confident that we don't have any more bugs in this use case (either nobody was doing this in parallel before or the regressions here were shockingly old), but at least that first Element_Activation/diffuse_activate_element_mesh2.i claims to be converging for me now.  Once we're more confident this is working we'll definitely need some unit testing, because it looks like it would be easy to inadvertently break this during future performance optimizations, just as happened last time, or during future projection feature additions.

### DIFF
--- a/include/systems/generic_projector.h
+++ b/include/systems/generic_projector.h
@@ -819,9 +819,27 @@ public:
 
     libmesh_assert_equal_to (old_indices.size(), indices.size());
 
-    values.resize(old_indices.size());
+    // We may have invalid_id in cases where no old DoF existed, e.g.
+    // due to expansion of a subdomain-restricted variable's subdomain
+    bool invalid_old_index = false;
+    for (const auto & di : old_indices)
+      if (di == DofObject::invalid_id)
+        invalid_old_index = true;
 
-    old_solution.get(old_indices, values);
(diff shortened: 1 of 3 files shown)


dewenyushu:
@roystgnr thanks for working on the patch! This works fine with the problems that I created for testing expanding a subdomain. I met some issues yesterday with displaced mesh using tensor mechanics, but that is not related to this.

I will be working a a unit test for this :)

roystgnr:
Thank you!  I'd be comfortable merging this before your unit test is ready, so no rush on that, but I'm afraid the changes here are low-level enough that I'm not comfortable merging until after Civet is back online and has a chance to report success with the existing CI test suite.

roystgnr:
Going to try closing and reopening this to see if our new "Civet, but on somebody's RPi or something" webhook is working.

PR #2676, opened 2020-08-14
Eim lookup table

dknez:
This is another update to the Reduced Basis EIM code. In this case the changes do not affect any existing code, they just provide a new option of defining an RBParametrizedFunction based on a "lookup table" rather than assuming that it is based on an analytical formula.

### DIFF
--- a/src/reduced_basis/rb_eim_construction.C
+++ b/src/reduced_basis/rb_eim_construction.C
@@ -250,6 +250,31 @@ void RBEIMConstruction::set_rb_construction_parameters(unsigned int n_training_s
   set_rel_training_tolerance(rel_training_tolerance_in);
   set_abs_training_tolerance(abs_training_tolerance_in);
 
+  if (get_rb_eim_evaluation().get_parametrized_function().is_lookup_table)
+    {
+      const std::string & lookup_table_param_name =
+        get_rb_eim_evaluation().get_parametrized_function().lookup_table_param_name;
+
+      libmesh_error_msg_if(!discrete_parameter_values_in.count(lookup_table_param_name),
+        "Lookup table parameter should be discrete");
+
+      std::vector<Real> & lookup_table_param_values =
+        libmesh_map_find(discrete_parameter_values_in, lookup_table_param_name);
+      unsigned int lookup_table_size = lookup_table_param_values.size();
+
+      // Overwrite the discrete values for lookup_table_param to make sure that
+      // it is: 0, 1, 2, ..., lookup_table_size-1.
+      for (unsigned int i=0; i<lookup_table_size; i++)
+        {
+          lookup_table_param_values[i] = i;
+        }
+
+      // Also, overwrite n_training_samples_in to make sure it matches
+      // lookup_table_size so that we will get full coverage of the
+      // lookup table in our training set.
+      n_training_samples_in = lookup_table_size;
+    }
+
   // Initialize the parameter ranges and the parameters themselves
   initialize_parameters(mu_min_in, mu_max_in, discrete_parameter_values_in);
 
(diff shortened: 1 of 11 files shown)


src/reduced_basis/rb_eim_construction.C
```diff
@@ -250,6 +250,31 @@ void RBEIMConstruction::set_rb_construction_parameters(unsigned int n_training_s
...
+
+      std::vector<Real> & lookup_table_param_values =
+        libmesh_map_find(discrete_parameter_values_in, lookup_table_param_name);
+      unsigned int lookup_table_size = lookup_table_param_values.size();
+
+      // Overwrite the discrete values for lookup_table_param to make sure that
+      // it is: 0, 1, 2, ..., lookup_table_size-1.
+      for (unsigned int i=0; i<lookup_table_size; i++)
+        {
+          lookup_table_param_values[i] = i;
```
jwpeterson, line 269: This could just be another `std::iota` call I think.

src/reduced_basis/rb_parametrized_function.C
```diff
@@ -31,6 +31,8 @@ namespace libMesh
 RBParametrizedFunction::RBParametrizedFunction()
 :
 requires_xyz_perturbations(false),
+is_lookup_table(false),
+lookup_table_param_name(""),
```
jwpeterson, line 35: An empty string is constructed by default so this one line isn't needed, but definitely do need to initialize the bool!

roystgnr: It might not hurt to leave the line in (IIRC there are some ultraparanoid warning levels we aren't close to being able to turn on that want you to explicitly write a constructor for every member), but I wonder if the compiler has an easier time optimizing () (default constructor, clearly no data to copy) over ("") (const char * overload, but hopefully inlined so maybe it can figure out that in this particular case there's no data to copy?).

jwpeterson:
Oh, I just remembered that the CI is not running, so we shouldn't merge this yet because of that.

src/reduced_basis/rb_parametrized_function.C
jwpeterson: Any compiler warning about not default-initializing a string (or any STL container) is definitely in the wrong IMO. I'm trying to imagine all the STL member variables we have and then imagine the compiler warning about every one of them that's not explicitly default-initialized...

dknez:
> Looks good to me, I just had a couple minor comments. If you agree, I can go ahead and make those and push an update.

Thanks for reviewing. Yes, please clean up as you like and merge when you're ready, thanks!

jwpeterson:
As I understand it, there's no ETA on CIVET and these changes won't really be tested by CIVET anyway, so I'm gonna go ahead and merge.

src/reduced_basis/rb_eim_construction.C
```diff
@@ -436,6 +475,43 @@ Real RBEIMConstruction::get_max_abs_value_in_training_set() const
...
+      auto local_pf = _local_parametrized_functions_for_training[i];
+
+      unsigned int RB_size = get_rb_eim_evaluation().get_n_basis_functions();
+      if (RB_size > 0)
+        {
+          // Get the right-hand side vector for the EIM approximation
+          // by sampling the parametrized function (stored in solution)
+          // at the interpolation points.
+          DenseVector<Number> EIM_rhs(RB_size);
+          for (unsigned int i=0; i<RB_size; i++)
```
roystgnr, line 500: We've now got a -Wshadow warning (error, in my usual configuration) in libMesh master.

Man I hope Civet is up and running again soon.

roystgnr: The index-change fix is simple enough *if* the code is currently correct, but I'll leave it to @jwpeterson and @dknez just in case there's an actual error being flagged by the warning.

jwpeterson: I'll push a fix for this shortly... looks to me like all the instances of `i` in the inner loop can be changed to `j` without issue, the outer `i` is only used for setting the `local_pf` reference.

PR #2677, opened 2020-08-14
GhostPointNeighbors optimization

roystgnr:
In theory this should give *much* faster results when doing mesh splitting, and slightly faster results for distributed mesh redistribution.

In practice, we should try some performance testing on common distributed mesh applications to make sure performance really is improved or unchanged there.

This change removes support for meshes where point_neighbors searches don't work because two sections of a 2D or 3D manifold only meet at a single node.  Such configurations haven't been supported for many years, and probably already break in other code paths, yet it turns out we had a mesh like that in our own unit tests.  That I wrote.  Embarrassing.

To make the new code (and other code) work more robustly on IGA meshes, I added the ability to default IGA node weights to values other than 0.0 on newly created elements, which incidentally means we also have the ability to set default values for any other extra elem or node data as well.  So that's nice.

This is passing my tests, but these changes are invasive enough and low-priority enough that we definitely should not merge this until Civet says it's okay, probably after tacking on a distributed mesh debug sweep too.

### DIFF
--- a/include/base/dof_object.h
+++ b/include/base/dof_object.h
@@ -403,10 +403,26 @@ public:
    * now be \p n_integers associated; this *replaces*, not augments,
    * any previous count.
    *
+   * Any newly-added values will initially be DofObject::invalid_id
+   *
    * If non-integer data is in the set, each datum of type T should be
    * counted sizeof(T)/sizeof(dof_id_type) times in \p n_integers.
    */
   void add_extra_integers (const unsigned int n_integers);
+
+  /**
+   * Assigns a set of extra integers to this \p DofObject.  There will
+   * now be \p n_integers associated; this *replaces*, not augments,
+   * any previous count.
+   *
+   * Any newly-added values will be copied from \p default_values.
+   *
+   * If non-integer data is in the set, each datum of type T should be
+   * counted sizeof(T)/sizeof(dof_id_type) times in \p n_integers, and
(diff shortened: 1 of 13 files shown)


roystgnr:
Tweaked the commit wording to force CI.  If this passes then I'd like to merge it before we branch off for the next release.

lindsayad:
Alrighty https://github.com/libMesh/libmesh/pull/2677/commits/ac995c7a515f4e924b50c65cb8aecbae1ff72505 caused this failure https://civet.inl.gov/job/588896/ I did a little more robust bisect this time

lindsayad:
That test meshes a pyramid-looking thing with tet4s AFAICT and then it does a whole bunch of subdomain bounding boxes with `BlockDeleters`. Presumably `prepare_for_use` is called after reading in the pyramid file mesh, and then again after 6 block deletion stages are conducted. I would say this is a bad test, but it's worrisome to me that we produce different results with different proc counts, e.g. the serialized Exodus mesh has has 87 nodes for all processor counts except the 15 proc run in our testing.

roystgnr:
We could revert that commit if we have to, since it's just a performance optimization, but it ought to be a pretty significant optimization, so I'd like to know what's going wrong.  Any chance this is the same never-technically-supported situation as the SlitMesh test, where two subsets of the mesh are connected at and only at a single node?

lindsayad:
Ill investigate tomorrow.

PR #2680, opened 2020-08-18
Fix -Wshadow warning.

jwpeterson:
Refs #2676.


dknez:
This looks good, but while we're at it we should change `auto local_pf = _local_parametrized_functions_for_training[i];` to `const auto & local_pf = _local_parametrized_functions_for_training[i];` and remove the comment about "will modify this below".

PR #2681, opened 2020-08-19
NemesisIO input

roystgnr:
This adds a basic NemesisIO(DistributedMesh) read() unit test, followed by enough fixes to get that read() working.

I'll be adding more serious testing (and realistically more fixes) in subsequent PRs but this should hopefully be a straightforward improvement already.

However, this shouldn't be merged until we have Civet up and running again - I think the Nemesis CI tests in MOOSE are going to be robust enough to not be fazed by the numbering issue I fixed here without us having to re-gold first, but I won't swear to that.

### DIFF
--- a/src/mesh/nemesis_io_helper.C
+++ b/src/mesh/nemesis_io_helper.C
@@ -694,31 +694,6 @@ void Nemesis_IO_Helper::put_elem_map(std::vector<int> & elem_mapi_in,
 
 
 
-
-
-
-void Nemesis_IO_Helper::put_n_coord(unsigned start_node_num,
-                                    unsigned num_nodes_in,
-                                    std::vector<Real> & x_coor,
-                                    std::vector<Real> & y_coor,
-                                    std::vector<Real> & z_coor)
-{
-  if (num_nodes_in)
-    {
-      nemesis_err_flag =
-        Nemesis::ne_put_n_coord(ex_id,
-                                start_node_num,
-                                num_nodes_in,
-                                x_coor.empty() ? nullptr : x_coor.data(),
-                                y_coor.empty() ? nullptr : y_coor.data(),
-                                z_coor.empty() ? nullptr : z_coor.data());
-
-      EX_CHECK_ERR(nemesis_err_flag, "Error writing coords to file!");
-    }
-}
-
-
-
 void Nemesis_IO_Helper::initialize(std::string title_in, const MeshBase & mesh, bool /*use_discontinuous*/)
 {
   // Make sure that the reference passed in is really a DistributedMesh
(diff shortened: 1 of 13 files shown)


src/mesh/nemesis_io_helper.C
```diff
@@ -2385,7 +2336,7 @@ void Nemesis_IO_Helper::write_nodal_solution(const std::vector<Number> & values,
 
       for (int i=0; i<num_nodes; ++i)
         {
-          Number value = values[this->exodus_node_num_to_libmesh[i]*num_vars + c];
+          Number value = values[(this->exodus_node_num_to_libmesh[i]-1)*num_vars + c];
```
jwpeterson, line 2339: I haven't thought about this in detail, but my initial reaction is "he's subtracting 1 from a libmesh node id" which seems like it would be bad if the libmesh node id is 0?

roystgnr: Sadly, it's not that straightforward.  Subtracting 1 from a libmesh-node-id-plus-one (which gets done because we want 1-based numbering in the file and we directly pass that vector::data() to an Exodus API.  See the "push_back(id+1)" call and preceding comment.

Should we rename that vector?  And move it while we're at it; I'm not sure why it's in exodusII_io_helper.h if it's only ever used by the subclass.

jwpeterson: > Should we rename that vector? And move it while we're at it; I'm not sure why it's in exodusII_io_helper.h if it's only ever used by the subclass.

I'd be in favor of both the renaming and moving to the Nemesis helper. Not sure why it was added to the Exodus helper... maybe it was used there at some point.

roystgnr:
In theory (and in my own libMesh testing) this branch now also has nodal data being written and re-read correctly.

In practice, we'll see what the MOOSE tests have to say about that.  If there's any regressions there then I'm going to regret not letting the original branch go through CI before I tacked on dozens more commits...

Since I've had to futz quite a bit with the heavily-used ExodusII_IO_Helper class here, and still I'm not yet familiar enough nor happy enough with that code, I'm leaning towards not including this PR in our next release branch, even if I'm lucky and CI can't find any problems with it.

roystgnr:
Those unit tests are passing for me on 1 through 16 processors now.  Hopefully --enable-complex works too.  I'll bet I don't have my ifdefs for --disable-nemesis/--disable-exodus in perfect order, so I'll check on that before we merge, but otherwise I think this is at a good break point.

jwpeterson (approved):
Looks good to me assuming you can figure out what's going on in the failing complex-variables build, but it's quite large (GitHub refused to even display the diffs in nemesis_io.C and nemesis_io_helper.C) so I probably haven't given it the most thorough possible review...

roystgnr:
I don't have enough hubris to say I've replicated **the** problem, but I can trigger **a** problem in two of the new unit tests in a complex build, so I'll see if I can fix those and we'll let CI take another crack at it from there.

roystgnr:
If you do want to review those huge diffs, going commit by commit might work.  I've been trying to keep each individual edit as small and self-contained as possible, and I've already tried to wipe out any blind alleys with fixup commits so hopefully nothing left in the intermediate commits is redundant/obsoleted either.

roystgnr:
The complex case was a pretty clean fix.  I'll merge tomorrow morning unless anyone screams.

Issue #2679, opened 2020-08-17
Removal of irrelevant pkgconfig files.


pshriwise:
I remember discussing this earlier in the year with some devs, but sadly that conversation was lost due to slack message history limitations. This is my account of that converstaion as best I can recall it:

Currently, it appears that all of the `libmesh-*.pc` files get installed with libmesh even if some of the library builds are disabled. This makes it impossible for an external application to determine which of these libraries are actually present in the install until it potentially fails to find them during linking. In particular, it would be nice to external apps to be able to select which libMesh library to link to based on its own configuration.

It seems that an appropriate fix would be to only install the `libmesh-*.pc` files that are relevant to the build configuration with an always-present `libmesh.pc` that refers to an available configuration in the installation with a preference of: `opt`, `devel`, `dbg`, `oprof`, `prof`. (Not sure that's an ideal order for the last two configuration options honestly, just my best guess.)

jwpeterson:
Definitely sounds like something it would be good to fix. Within our autoconf files, once the `LIBMESH_SET_METHODS` (m4/libmesh_methods.m4) function has been called, the Makefile variables `LIBMESH_OPT_MODE`, `LIBMESH_DBG_MODE`, `LIBMESH_DEVEL_MODE`, `LIBMESH_PROF_MODE`, `LIBMESH_OPROF_MODE`. 

The relevant lines in Makefile.am are:
```
# Support for pkgconfig
pkgconfigdir     = $(libdir)/pkgconfig
pkgconfig_DATA   = contrib/utils/Make.common \
                   contrib/utils/libmesh.pc \
                   contrib/utils/libmesh-opt.pc \
                   contrib/utils/libmesh-dbg.pc \
                   contrib/utils/libmesh-devel.pc \
                   contrib/utils/libmesh-prof.pc \
                   contrib/utils/libmesh-oprof.pc

configdir        = $(sysconfdir)/libmesh
config_DATA      = contrib/utils/Make.common \
                   contrib/utils/libmesh.pc \
                   contrib/utils/libmesh-opt.pc \
                   contrib/utils/libmesh-dbg.pc \
                   contrib/utils/libmesh-devel.pc \
                   contrib/utils/libmesh-prof.pc \
                   contrib/utils/libmesh-oprof.pc
```
so one thing you could try is wrapping these Makefile.am lines in conditional statements, e.g.
``` 
if LIBMESH_OPT_MODE
[...]

PR #2682, opened 2020-08-19
Updates to pkgconfig files

pshriwise:
This PR addresses #2679. 

With these changes Pkgconfig files are only installed for build methods which are enabled. 

Additionally, the input file used to configure the `libmesh.pc` file is chosen based on the enabled build methods as well with a preference order of `opt`, `dbg`, `devel`, `prof`, `oprof`. This ensures that the `libmesh.pc` is always valid and can be used by external applications consistently.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -115,21 +115,33 @@ AUX_DIST = build-aux/install-sh \
 # Support for pkgconfig
 pkgconfigdir     = $(libdir)/pkgconfig
 pkgconfig_DATA   = contrib/utils/Make.common \
-                   contrib/utils/libmesh.pc \
-                   contrib/utils/libmesh-opt.pc \
-                   contrib/utils/libmesh-dbg.pc \
-                   contrib/utils/libmesh-devel.pc \
-                   contrib/utils/libmesh-prof.pc \
-                   contrib/utils/libmesh-oprof.pc
+                   contrib/utils/libmesh.pc
 
 configdir        = $(sysconfdir)/libmesh
 config_DATA      = contrib/utils/Make.common \
-                   contrib/utils/libmesh.pc \
-                   contrib/utils/libmesh-opt.pc \
-                   contrib/utils/libmesh-dbg.pc \
-                   contrib/utils/libmesh-devel.pc \
-                   contrib/utils/libmesh-prof.pc \
-                   contrib/utils/libmesh-oprof.pc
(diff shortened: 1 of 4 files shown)


pshriwise:
Wondering if `devel` should come before `dbg` in order of preference for the `libmesh.pc` file. Any thoughts @jwpeterson?

jwpeterson:
You'll have to run `./bootstrap` in the top level of the repo and commit the results in order for your changes to make their way into the "generated" autoconf files. We do this so that regular users just have to run configure/make/make install and don't have to worry about bootstrapping. And you'll want to be sure and use the version of the autotools that comes with libmesh for this... there are some instructions [here](https://github.com/libMesh/libmesh/wiki/Updating-configure-tests%2C-Makefiles%2C-etc.)

> Wondering if devel should come before dbg in order of preference

Are you asking for the case when the user does not build opt but does build both debug and devel methods? I'd say dbg, but I don't feel too strongly about it either way.

pshriwise:
Took care of the bootstrapping I think. Let me know if something went wrong though!

>Are you asking for the case when the user does not build opt but does build both debug and devel methods? I'd say dbg, but I don't feel too strongly about it either way.

Yeah that's what I had in mind. I was wondering if performance might take precedence there, but I think you're right.

jwpeterson:
Thanks, since our main CI is down at the moment, it might be a little while before this gets merged. 

By the way, if I wanted to test the pkg-config files on my system, what exactly would I run? I guess something like 
```
pkg-config --cflags --libs $LIBMESH_DIR/etc/libmesh/libmesh.pc
```
is sufficient? I gather that by placing that directory in `$PKG_CONFIG_PATH` I could shorten this to:
```
pkg-config --cflags --libs libmesh
```
Do you use this in your project's Makefiles or something?

pshriwise:
Sounds good. No big rush. It'd be great to have this in before the next release though. :crossed_fingers: 

That's what I've been using to test it, yeah. I've been using CMake's `pkg_check_modules` primarily, but I have other test projects using Makefile where it will be nice to be able to always rely on the `libmesh.pc` file too.

jwpeterson:
I had a chance to test this in a build directory where only dbg mode is built and it seems to work fine. I also verified that re-running ./bootstrap produces no changes. So I think it's probably safe to merge even though CI hasn't had a chance to run yet... Thanks @pshriwise!

PR #2683, opened 2020-08-20
Fix indexing in EquationSystems::get_vars_active_subdomains

lindsayad:
Without this fix, output is actually broken anytime the variable
numbering in the system doesn't match the ordering in the `names`
parameter. This refs a MOOSE user reported issue
[here](https://groups.google.com/d/msgid/moose-users/7f996a4a-1a97-4836-84fc-e86e32cac868o%40googlegroups.com?utm_medium=email&utm_source=footer)


lindsayad:
I could actually believe that there may be a MOOSE test that unwittingly has this bad behavior embedded in it, but I guess even then I would have to break CI at some point (unless going the circuitous route of updating MOOSE with a libmesh PR head)

PR #2686, opened 2020-08-25
Unsteady Adjoint Error Estimation Implementation (Model Error)

vikramvgarg:
This PR adds capability to do unsteady adjoint error estimation for model error, for both uniform and TwoStep timesolvers.

1) TS::integrate_adjoint_refinement_error_estimate takes in references to an ARefEE object and  ErrorVector as arguments.
2) The above method has been defined for both uniform and 2Step time solvers, like we had for adjoint sensitivity.
3) TS::integrate_qoi_timestep time integrates QoIs specified by user consistently with the integration method used for the primal and adjoint solves
4) Verified for calculating model error for an instantaneous and a smooth (temporally) QoI, illustrated in new adjoints example 7.


src/error_estimation/adjoint_refinement_estimator.C
```diff
@@ -111,8 +112,25 @@ void AdjointRefinementEstimator::estimate_error (const System & _system,
...
   // Only if the user didn't already solve it for us
+  // If _adjoint_evaluation_physics pointer is not null, swap
+  // the current physics with the one held by _adjoint_evaluation_physics
+  // before assembling the adjoint problem
   if (!system.is_adjoint_already_solved())
-    system.adjoint_solve(_qoi_set);
+  {
+    // Swap the physics if needed, note that the current physics pointer
+    // held by the system will be pointed to by _adjoint_evalation_physics after swapping
+    const bool swapping_physics = _adjoint_evaluation_physics;
```
roystgnr, line 122: Instead of recreating this bool a zillion times (with the same name but with two different meanings! we previously used it for _residual_evaluation_physics), can we either just directly test the physics pointers (they do implicitly convert to bool inside if()), or make two different bools, once each, with distinct names?

This annoys the compiler, too: with --enable-paranoid-warnings I get -Wshadow yelling at me.

vikramvgarg: Makes sense.

include/solvers/adaptive_time_solver.h
```diff
@@ -82,7 +82,9 @@ class AdaptiveTimeSolver : public FirstOrderUnsteadySolver
 
   virtual void integrate_adjoint_sensitivity(const QoISet & qois, const ParameterVector & parameter_vector, SensitivityData & sensitivities) override = 0;
 
-  virtual Real last_complete_deltat() override { return completedtimestep_deltat; }
+  virtual void integrate_adjoint_refinement_error_estimate(AdjointRefinementEstimator & adjoint_refinement_error_estimator, ErrorVector & QoI_elementwise_error, std::vector<Real *> QoI_time_instant = std::vector<Real *>()) = 0;
+
+  virtual Real last_complete_deltat() { return completedtimestep_deltat; };
```
roystgnr, line 87: Why'd you take off ```override``` here?

vikramvgarg: Sorry, this was probably mess up due to a conflict. I had been futzing with this function and had thought I had restricted it to AdaptiveTS and subclasses only. I'll correct this.

include/solvers/file_solution_history.h
```diff
@@ -97,7 +97,7 @@ class FileSolutionHistory : public SolutionHistory
   stored_solutions_iterator stored_sols;
 
   // A helper function to locate entries at a given time
-  void find_stored_entry(Real time);
+  void find_stored_entry(Real time, bool storing = false);
```
roystgnr, line 100: We ought to document ```\p storing``` (and the reason for it) in the comment here.

include/solvers/memory_solution_history.h
```diff
@@ -96,7 +96,7 @@ class MemorySolutionHistory : public SolutionHistory
   stored_solutions_iterator stored_sols;
 
   // A helper function to locate entries at a given time
-  void find_stored_entry(Real time);
+  void find_stored_entry(Real time, bool storing = false);
```
roystgnr, line 99: Likewise in the comment here.

include/solvers/time_solver.h
```diff
@@ -267,6 +278,18 @@ class TimeSolver : public ReferenceCountedObject<TimeSolver>
    */
   virtual Real last_complete_deltat();
 
+  /**
+   * Setter and getter for the final time variable
```
roystgnr, line 282: Needs more documentation, too.  At first glance it just raises a ton of "Do I have to set this?", "What's it get used for?", "What if I want to restart a simulation to an extended time?" sorts of questions.

src/solvers/unsteady_solver.C
```diff
@@ -306,4 +481,12 @@ Real UnsteadySolver::du(const SystemNorm & norm) const
   return _system.calculate_norm(*solution_copy, norm);
 }
 
+void UnsteadySolver::update()
```
roystgnr, line 484: In hindsight this was enough of a bugfix that I'd have been happy committing it in its own PR.

src/error_estimation/adjoint_refinement_estimator.C
```diff
@@ -367,6 +403,26 @@ void AdjointRefinementEstimator::estimate_error (const System & _system,
         }
     }
 
+  // We are all done with Dirichlet lift vectors and they should be removed, lest we run into I/O issues later
```
roystgnr, line 406: Should we remove them, or just set_vector_preservation(false) here?  If we're likely to use them later, and we are, then there's no point in churning memory, right?

vikramvgarg: Good point, they may get used later for user defined QoI evaluations.

src/solvers/file_solution_history.C
```diff
@@ -53,10 +53,21 @@ void FileSolutionHistory::find_stored_entry(Real time)
...
 
-  // If we are at end, we are creating a new entry, nothing more to do
+  // If we are at end, we could be creating a new entry (depends on the storing bool), return
+  // Otherwise, get a decremented iterator for the sandwich test
   if(lower_bound_it == stored_solutions.end())
   {
-    return;
+    if(storing)
+    {
+      return;
```
roystgnr, line 62: So at this point we haven't updated stored_sols?  Isn't that potentially a bug if we don't have an empty history?

src/solvers/file_solution_history.C
```diff
@@ -77,9 +88,16 @@ void FileSolutionHistory::find_stored_entry(Real time)
   {
     stored_sols = lower_bound_it_decremented;
   }
-  else
+  else // Neither of the two candidate keys matched our time
   {
-    libmesh_error_msg("Failed to set stored solutions iterator to a valid value.");
+    if(storing) // If we are storing, this is fine, we need to create a new entry, so just return
+    {
+      return;
```
roystgnr, line 95: Same question here - are we just getting away with this because we only create new entries at the end of the history?

src/solvers/twostep_time_solver.C
```diff
@@ -120,14 +123,16 @@ void TwostepTimeSolver::solve()
       old_deltat = _system.deltat;
       _system.deltat *= 0.5;
 
+      _system.update();
```
roystgnr, line 126: Neither of these are redundant?  Really?

include/systems/system.h
```diff
@@ -1573,6 +1573,11 @@ class System : public ReferenceCountedObject<System>,
    */
   std::vector<Number> qoi;
 
+  /**
+   * Vector to hold error estimates for qois
+   */
+  std::vector<Number> qoi_error_estimates;
```
roystgnr, line 1579: Does this belong in the System, like qoi, or as a separate output, like our ErrorVector results?

With System::qoi we have other code that needs to calculate it, so we can get a little savings by keeping it around in a well-defined place to avoid making that calculation redundantly.  Does the same apply here?

src/solvers/twostep_time_solver.C
```diff
@@ -339,4 +340,60 @@ void TwostepTimeSolver::integrate_adjoint_sensitivity(const QoISet & qois, const
...
+
+  // Add the spatially integrated errors for all the QoIs in the QoI set
+  Real spatially_integrated_qoi_error;
+
+  for (auto j : make_range(_system.n_qois()))
+  {
+    // Skip this QoI if not in the QoI Set
+    if (adjoint_refinement_error_estimator.qoi_set().has_index(j))
+    {
+      (_system.qoi_error_estimates)[j] = qoi_error_estimates_first_half[j] + qoi_error_estimates_second_half[j];
```
roystgnr, line 394: There's got to be a better nomenclature we can use here.  Correct me if I'm wrong, but:

```qoi_error_estimates``` isn't an estimate of the QoI error, but only the component of that estimate which comes from a single time step?

```integrate_adjoint_refinement_error_estimate``` integrates it over the current time step (or two sub time steps), but doesn't add it to the time integration in progress?

Those are defensible choices but the names aren't self-documenting, they're almost anti-self-documenting.  And regardless of name we're going to need much, much more documentation in the headers.

File: src/solvers/unsteady_solver.C [hunk 13]
```diff
@@ -281,6 +294,168 @@ void UnsteadySolver::integrate_adjoint_sensitivity(const QoISet & qois, const Pa
...
+  {
+    // For evaluating the residual, we need to use the deltat that was used
+    // to get us to this solution, so we save the current deltat as next_step_deltat
+    // and set _system.deltat to the last completed deltat.
+    next_step_deltat = _system.deltat;
+    _system.deltat = last_step_deltat;
+
+    // The adjoint error estimate expression for a backwards facing step
+    // scheme needs the adjoint for the last time instant, so save the current adjoint for future use
+    NumericVector<Number> & current_adjoint = _system.get_adjoint_solution(0);
```
roystgnr, line 330: This only works if we have a single QoI at index 0?  That needs to be fixed.  Maybe with a QoISet argument to this function?

roystgnr: Ah, sorry, didn't see that you're using the QoISet attached to the ARefEE.

roystgnr: But we should still be doing that same loop here, no?

src/solvers/unsteady_solver.C
```diff
@@ -281,6 +294,168 @@ void UnsteadySolver::integrate_adjoint_sensitivity(const QoISet & qois, const Pa
...
+        (_system.qoi_error_estimates)[j] = ((qoi_error_estimates_left[j] + qoi_error_estimates_right[j])/2.)*(time_right - time_left);
+      }
+      else if(time_right <= *(QoI_time_instant[j]) + TOLERANCE)
+      {
+        (_system.qoi_error_estimates)[j] = qoi_error_estimates_right[j]*last_step_deltat;
+      }
+      else
+      {
+        (_system.qoi_error_estimates)[j] = 0.0;
+        libmesh_warning("Error estimate contribution for a time instant past the instantaneous QoI evaluation time has been set to zero.");
```
roystgnr, line 453: When do we expect this to happen?  Should this be an error rather than a warning?

src/systems/diff_system.C
```diff
@@ -354,6 +354,11 @@ bool DifferentiableSystem::have_second_order_scalar_vars() const
 void DifferentiableSystem::swap_physics ( DifferentiablePhysics * & swap_physics )
 {
   std::swap(this->_diff_physics, swap_physics);
+
+  // If the physics has been swapped, we will reassemble
+  // the matrix from scratch before doing an adjoint solve
+  // rather than just transposing
+  this->disable_cache();
```
roystgnr, line 361: Great catch.  This is another bit that could have been its own bugfix PR.

src/systems/implicit_system.C
```diff
@@ -711,7 +711,7 @@ void ImplicitSystem::adjoint_qoi_parameter_sensitivity (const QoISet & qoi_indic
   // if we havent already or dont have an initial condition for the adjoint
   if (!this->is_adjoint_already_solved())
     {
-      this->adjoint_solve(qoi_indices);
+      this->ImplicitSystem::adjoint_solve(qoi_indices);
```
roystgnr, line 714: Why don't we want to let derived systems override this?

include/solvers/twostep_time_solver.h
```diff
@@ -78,6 +78,15 @@ class TwostepTimeSolver : public AdaptiveTimeSolver
    */
   virtual void integrate_adjoint_sensitivity(const QoISet & qois, const ParameterVector & parameter_vector, SensitivityData & sensitivities) override;
 
+  /**
+   * A method to compute the adjoint refinement error estimate at the current timestep.
+   * int_{tstep_start}^{tstep_end} R(u^h,z) dt
+   * Fills in an ErrorVector that contains the weighted sum of errors from all the QoIs and can be used to guide AMR.
```
roystgnr, line 84: If we're using the ARefEE::qoi_set() then "all the QoIs" isn't always correct here.

roystgnr:
> Verified for calculating model error for an instantaneous QoI.

Can we get that added to the examples in this same PR?  There's way too many places for possible errors to crop up in this code for me to be confident in it just based on inspection, and even if I was then I'd still be afraid of regressing it as soon as someone sneezed.

src/solvers/twostep_time_solver.C
vikramvgarg: They both probably are, sorry this is a leftover from a debugging hack.

include/systems/system.h
vikramvgarg: This isnt going to be a NumericVector elementwise vector like ErrorVector. The global QoI_elementwise_error is a weighted sum of contributions from each QoI, so we dont need to maintain separate vectors for each QoI.

vikramvgarg:
> > Verified for calculating model error for an instantaneous QoI.
> 
> Can we get that added to the examples in this same PR? There's way too many places for possible errors to crop up in this code for me to be confident in it just based on inspection, and even if I was then I'd still be afraid of regressing it as soon as someone sneezed.

I think we can extend adjoints example 5 to include testing the error estimation.

src/systems/implicit_system.C
vikramvgarg: The this pointer here can be for a TS object.

roystgnr: The this pointer in ImplicitSystem::anything() is always going to be an ImplicitSystem or a subclass of one, which TimeSolver isn't.

vikramvgarg: Yeah, I guess went overboard with the disambiguation here.

src/solvers/unsteady_solver.C
vikramvgarg: You are right, this needs to be generalized.

src/error_estimation/adjoint_refinement_estimator.C
vikramvgarg: Was resolved in commit 3ee8a03494626d7e3c85561baccf59800a187ccb.

include/solvers/adaptive_time_solver.h
vikramvgarg: Fix commit:

include/solvers/time_solver.h
vikramvgarg: Fix commit:

src/solvers/twostep_time_solver.C
vikramvgarg: Fix commit:

src/solvers/file_solution_history.C
vikramvgarg: The only time find_stored_entry needs to change stored_sols is if we actually manage to find an existing entry for that time. This should never happen during a timestepping loop unless erase has not been to remove garbage entries. 

This could happen if we embed the timestepping loop inside an adaptive or continuation loop for example. So perhaps we should add a method to clear the stored_solutions map ? This would also delete old files ideally.

src/solvers/unsteady_solver.C
vikramvgarg: QoI0 is instantaneous and QoI2 is smooth and has support past the time instant.

src/error_estimation/adjoint_refinement_estimator.C
vikramvgarg: Reading in breaks if the lift vectors are not removed.

include/solvers/time_solver.h
```diff
@@ -279,7 +279,8 @@ class TimeSolver : public ReferenceCountedObject<TimeSolver>
   virtual Real last_complete_deltat();
 
   /**
-   * Setter and getter for the final time variable
+   * The final_time variable can be set by the user or in another library
+   * function. For use, see documentation of final_time variable.
```
roystgnr, line 283: This tells us *what* we can do with it, but we didn't need the comment for that.  set_final_time() clearly sets a final_time() variable, get_final_time() clears one.  All is solid there.

What we're missing in the comment is *why* we want or need to do this.  Suppose I'm time marching to steady state (via some ||du/dt|| < tolerance).  I don't know my final time.  Is it impossible for me to do adjoint time stepping and QoI postprocessing?  What happens if I get final_time wrong?

src/solvers/unsteady_solver.C
roystgnr: QThat seems awfully specific and awfully cryptic.  I assume you mean that, because one smooth QoI exists, the user will be calling this function at every time step, and because an instantaneous QoI also exists, the library will notice that it's being asked for error terms coming from time steps after that QoI instant?  That's not even worth a warning, then, except in the case where the QoI set doesn't have any other QoIs.

Speaking of awfully cryptic - I just noticed now that QoI_time_instant isn't even documented anywhere yet.  Right now the user base of this code is basically "you", and maybe I have a fighting chance because I can go back through IM history to figure out what various things are for, but I'm pretty sure if I showed these APIs to another core developer, much less a typical user, they'd have no idea where to begin.  Hopefully the new example will help with that, but examples are a good addition to documentation, not a replacement for it.

include/solvers/file_solution_history.h
roystgnr: Indeed, passing different values to a method's argument is expected to change its behavior.  We knew that before the redundant comment.  What's important is *how* the behavior changes.  In this case, the issues we have with floating-point comparisons and the manner in which this fixes them needs to be here for everyone to read, not in IM logs for just you and me.

include/solvers/unsteady_solver.h
```diff
@@ -204,9 +204,14 @@ class UnsteadySolver : public TimeSolver
 private:
 
   /**
-   * A pointer to a vector to hold the adjoint solution at the last time step
+   * A vector of pointers to vectors holding the adjoint solution at the current time step
    */
-  std::unique_ptr<NumericVector<Number>> old_adjoint;
+  std::vector< NumericVector<Number>* > current_adjoints;
```
roystgnr, line 209: Do we need an extra variable to hold pointers to the things we're already storing in the System?  Even if it was useful for clarity in the actual method there'd be no reason to keep it in the class rather than as a local variable.

include/solvers/file_solution_history.h
vikramvgarg: I've added a description of how the behaviour of the function changes depending on whether we are storing or not-storing.

src/solvers/unsteady_solver.C
vikramvgarg: Added documentation for QoI_time_instant in time_solver.h with the comments for the base integrate_adjoint_refinement_error_estimate function declaration.

include/solvers/time_solver.h
vikramvgarg: In the constructor, final_time is set to std::nan("1"). Not setting it does not preclude adjoint time stepping or QoI postprocessing. If you want to evaluate a final time QoI and compute a corresponding adjoint, you need to set this variable to a finite value.

include/solvers/unsteady_solver.h
vikramvgarg: Removed.

examples/adjoints/adjoints_ex7/SigmaPhysics.C
```diff
@@ -0,0 +1,194 @@
...
+
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+
+
+#include "libmesh/getpot.h"
+
+#include "SigmaPhysics.h"
```
roystgnr, line 22: We don't use CamelCase for anything else; can this become sigma_physics.h and sigma_physics.C?

examples/adjoints/adjoints_ex7/SigmaPhysics.C
```diff
@@ -0,0 +1,194 @@
...
+
+  GetPot infile("heat.in");
+  _k = infile("k", 1.0);
+  _analytic_jacobians = infile("analytic_jacobians", true);
+
+  // The temperature is evolving, with a first order time derivative
+  this->time_evolving(T_var, 1);
+
+  const boundary_id_type all_ids[4] = {0, 1, 2, 3};
+  std::set<boundary_id_type> all_bdys(all_ids, all_ids+(2*2));
```
roystgnr, line 51: Okay, I'll bite... why "2*2" and not just "4"?

examples/adjoints/adjoints_ex7/SigmaPhysics.C
```diff
@@ -0,0 +1,194 @@
...
+
+  ZeroFunction<Number> zero;
+
+  ConstFunction<Number> one(1.0);
+
+  // Most DirichletBoundary users will want to supply a "locally
+  // indexed" functor
+  sys.get_dof_map().add_dirichlet_boundary(DirichletBoundary (all_bdys, T_only, &one));
+
+  // Set adjoint boundary condition z(1,y) = y * (1 - y)
```
roystgnr, line 63: Huh?  Looks like "z(partial Omega) = 0" to me.

examples/adjoints/adjoints_ex7/SigmaPhysics.C
vikramvgarg: I guess I didnt want to extend to reach the backspace key ...

examples/adjoints/adjoints_ex7/SigmaPhysics.C
```diff
@@ -0,0 +1,194 @@
...
+  std::set<boundary_id_type> all_bdys(all_ids, all_ids+(2*2));
+
+  std::vector<unsigned int> T_only(1, T_var);
+
+  ZeroFunction<Number> zero;
+
+  ConstFunction<Number> one(1.0);
+
+  // Most DirichletBoundary users will want to supply a "locally
+  // indexed" functor
```
roystgnr, line 60: Does this even mean anything in this context, where the functor output is independent of index?  Or is it a copy-and-paste (probably of a copy-and-paste) from a different example where it did mean something?

examples/adjoints/adjoints_ex7/SigmaPhysics.C
vikramvgarg: Sorry, that needs to be updates.

examples/adjoints/adjoints_ex7/SigmaPhysics.C
vikramvgarg: Yeah, I started with an existing example.

examples/adjoints/adjoints_ex5/adjoints_ex5.C
```diff
@@ -502,7 +502,7 @@ int main (int argc, char ** argv)
 
           // Output the H1 norm of the computed solution
           libMesh::out << "|U("
-                      << system.time + system.time_solver->last_complete_deltat()
+                      << system.time + system.time_solver->last_completed_deltat()
```
roystgnr, line 505: I like the new name better too, but is it worth API breakage?  I guess we haven't had the old name around very long?  But if it's worth changing then it's worth keeping a libmesh_deprecated() shim for the old name for a while.

examples/adjoints/adjoints_ex7/SigmaPhysics.C
```diff
@@ -0,0 +1,194 @@
...
+      NumericVector<Number> & adjoint_solution1 =
+        const_cast<System &>(sys).get_adjoint_solution(1);
+
+      // Add this adjoint solution to the vectors that diff context should localize
+      c.add_localized_vector(adjoint_solution1, sys);
+    }
+
+}
+
+#define optassert(X) {if (!(X)) libmesh_error_msg("Assertion " #X " failed.");}
```
roystgnr, line 112: Is this getting used anywhere?

examples/adjoints/adjoints_ex5/adjoints_ex5.C
vikramvgarg: The only other place we have used this is adjoints example 5, and I updated that. We can make a shim, but I doubt anyone is using this.

examples/adjoints/adjoints_ex7/SigmaPhysics.C
vikramvgarg: Its a workaround for Fedora OS issue IIRC.

examples/adjoints/adjoints_ex7/SigmaPhysics.h
```diff
@@ -0,0 +1,55 @@
...
+// FEMSystem, TimeSolver and  NewtonSolver will handle most tasks,
+// but we must specify element residuals
+class SigmaPhysics : public FEMPhysics
+{
+public:
+  // Constructor
+  SigmaPhysics():FEMPhysics(){}
+
+  std::string & fe_family() { return _fe_family; }
+  unsigned int & fe_order() { return _fe_order; }
```
roystgnr, line 19: Are these accessors or the variables they access getting used anywhere?  They look like they're copy-and-paste cruft from a System subclass where they would have been used.

vikramvgarg: Yeah, all this excess needs to be trimmed.

examples/adjoints/adjoints_ex7/SigmaPhysics.h
```diff
@@ -0,0 +1,55 @@
...
+
+  // Variables to hold the computed QoIs
+  Number computed_QoI[1];
+
+  // The FE type to use
+  std::string _fe_family;
+  unsigned int _fe_order;
+
+  // Index for T variable
+  unsigned int T_var;
```
roystgnr, line 50: You've got T_var here, uninitialized, and unused, while 0 is hardcoded in the physics kernel itself?  Is it possible that at some point I said that our example codes are to teach new users a lesson, and you didn't realize that I meant that literally, it wasn't a euphemism for punishment?  :laughing:

examples/adjoints/adjoints_ex7/adjoint_initial.C
```diff
@@ -0,0 +1,77 @@
...
+
+using namespace libMesh;
+
+void adjoint_read_initial_parameters()
+{
+}
+
+void adjoint_finish_initialization()
+{
+}
```
roystgnr, line 30: I'm not understanding what either of these are for.

examples/adjoints/adjoints_ex7/SigmaPhysics.h
vikramvgarg: Well, my cleanup for the example focused on the adjoints_ex7.C and the element_qoi/derivative files. Forgot to take care of the other ones.

examples/adjoints/adjoints_ex7/adjoint_initial.C
```diff
@@ -0,0 +1,77 @@
...
+
+
+
+// Initial conditions
+Number adjoint_initial_value0(const Point & p,
+                             const Parameters &,
+                             const std::string &,
+                             const std::string &)
+{
+  Real x = p(0), y = p(1);
```
roystgnr, line 40: Okay, I'm going to stop here for now, at the unused variables.  I need to push a re-bootstrap commit so that our CI actually runs this example, and you need to reconfigure with --enable-paranoid-warnings --enable-werror so that the compiler can have more to harass you about and I can have less.

vikramvgarg: Yeah, some cleanup is needed.

examples/adjoints/adjoints_ex7/SigmaPhysics.C
roystgnr: Okay, I can find it in my svn logs now.  The comment later in the file is "Workaround for weird FC6 bug".  That's Fedora Core 6, released 2006.  The reason I had to go back to svn logs is because the compiler bug nearly predated git!

We don't support anything without C++11 anymore, much less anything with such weird bugs.  Remove it.

examples/adjoints/adjoints_ex7/sigma_physics.C
```diff
@@ -0,0 +1,183 @@
...
+                                          DiffContext & context)
+{
+  bool compute_jacobian = request_jacobian && _analytic_jacobians;
+
+  FEMContext & c = cast_ref<FEMContext &>(context);
+
+  // First we get some references to cell-specific data that
+  // will be used to assemble the linear system.
+  FEBase * elem_fe = nullptr;
+  c.get_element_fe(0, elem_fe);
```
roystgnr, line 121: Either get rid of T_var or use T_var everywhere we need the index of the T variable.  Pretending that index is a variable when it's really a hard-coded constant is just misleading.

examples/adjoints/adjoints_ex7/factoryfunction.C
```diff
@@ -0,0 +1,73 @@
...
+// So we can look up hand-coded functions by name string
+template<>
+std::map<std::string, Factory<FunctionBase<Number>> *> &
+Factory<FunctionBase<Number>>::factory_map()
+{
+  static std::map<std::string, Factory<FunctionBase<Number>> *> _map;
+  return _map;
+}
+
+FactoryImp<ExampleOneFunction, FunctionBase<Number>> example_one_factory ("example_one");
```
roystgnr, line 69: Where is this getting used?

examples/adjoints/adjoints_ex7/femparameters.C
```diff
@@ -0,0 +1,753 @@
+
+#include "femparameters.h"
```
roystgnr, line 2: Does this differ from the adjoints_ex1 femparameters.C?  If not, could we just use the same symlink trick that we use in the other adjoints examples?

examples/adjoints/adjoints_ex7/heatsystem.h
```diff
@@ -0,0 +1,111 @@
...
+  // The FE type to use
+  std::string _fe_family;
+  unsigned int _fe_order;
+
+  // Calculate Jacobians analytically or not?
+  bool _analytic_jacobians;
+
+  // Variables to hold the perturbed residuals
+  std::vector<Number> R_plus_dp;
+  std::vector<Number> R_minus_dp;
```
roystgnr, line 98: These never get used, do they?  Can you go through and remove everything that never gets used?

examples/adjoints/adjoints_ex7/initial.C
```diff
@@ -0,0 +1,51 @@
...
+
+using namespace libMesh;
+
+void read_initial_parameters()
+{
+}
+
+void finish_initialization()
+{
+}
```
roystgnr, line 30: These too.  Having a single main() that called differing functions in differing codes made some sense at the time.  But once main() is no longer general we don't gain anything by spending dozens of lines of code pretending otherwise.

include/solvers/euler2_solver.h
```diff
@@ -70,6 +70,21 @@ class Euler2Solver : public FirstOrderUnsteadySolver
...
+   * A method to integrate the system::QoI functionals.
+   */
+  virtual void integrate_qoi_timestep() override;
+
+  /**
+   * A method to compute the adjoint refinement error estimate at the current timestep.
+   * int_{tstep_start}^{tstep_end} R(u^h,z) dt
+   * Fills in an ErrorVector that contains the weighted sum of errors from all the QoIs and can be used to guide AMR.
+   * Also fills in a map that links QoI indices to spatially integrated error estimates for the QoI with that index.
+   * The trapezoidal rule (for temporally smooth QoI) or left Riemann sum (for an instantaneous QoI) is used
```
roystgnr, line 83: If we're using the trapezoidal rule regardless of theta, that's inconsistent and we need to fix it or at the very least error out when we see theta!=0.5.

If we're weighting our integration consistently with theta then this comment should say so instead.

include/solvers/euler_solver.h
```diff
@@ -92,6 +92,21 @@ class EulerSolver : public FirstOrderUnsteadySolver
...
+   * A method to integrate the system::QoI functionals.
+   */
+  virtual void integrate_qoi_timestep() override;
+
+  /**
+   * A method to compute the adjoint refinement error estimate at the current timestep.
+   * int_{tstep_start}^{tstep_end} R(u^h,z) dt
+   * Fills in an ErrorVector that contains the weighted sum of errors from all the QoIs and can be used to guide AMR.
+   * Also fills in a map that links QoI indices to spatially integrated error estimates for the QoI with that index.
+   * The trapezoidal rule (for temporally smooth QoI) or left Riemann sum (for an instantaneous QoI) is used
```
roystgnr, line 105: Here, on the other hand, we shouldn't be using the trapezoidal rule for any theta; even theta=0.5 should be doing a midpoint evaluation; that's the whole distinction between EulerSolver and Euler2Solver.  It would be better to have nothing but a libmesh_not_implemented() override here, so we know to fix it if/when we need it, than to have something subtly inconsistent.

examples/adjoints/adjoints_ex7/heatsystem.h
vikramvgarg: Will do.

include/solvers/euler2_solver.h
vikramvgarg: The comment is outdated, will bring it up to speed with the more general implementation we have now.

jwpeterson:
Before this PR gets merged I'd like to clean up the history quite a bit... If one of can you can ping me when it is ready I will take a crack at that.

include/solvers/time_solver.h
```diff
@@ -132,12 +134,29 @@ class TimeSolver : public ReferenceCountedObject<TimeSolver>
...
 
+  /**
+   * A method to compute the adjoint refinement error estimate at the current timestep.
+   * int_{tstep_start}^{tstep_end} R(u^h,z) dt
+   * The user provides an initialized ARefEE object.
+   * Fills in an ErrorVector that contains the weighted sum of errors from all the QoIs and can be used to guide AMR.
+   * As an option, the user can also supply a QoI_time_instant vector where non-null pointers imply an instantaneous
+   * QoI evaluation at the time held by the pointer.
+   * Also fills in a map that links QoI indices to spatially integrated error estimates for the current timestep.
+   * ONLY SUPPORTED for Backward Euler.
```
roystgnr, line 156: Let's put "ONLY CURRENTLY SUPPORTED", and in the subclasses put an
```
if (theta != 1.0)
  libmesh_not_implemented();
```
to ensure it?  People should read documentation, and should read warnings, but they don't, and what people *do* read is the final message saying why their program stopped.

include/solvers/time_solver.h
```diff
@@ -261,12 +280,33 @@ class TimeSolver : public ReferenceCountedObject<TimeSolver>
   { _is_adjoint = _is_adjoint_value; }
 
   /**
-   * Function to return 'last_deltat()', returns system.deltat if
-   * fixed timestep solver is used, completed_deltat if the adaptive
+   * Returns system.deltat if fixed timestep solver is used,
+   * the complete timestep size (sum of all substeps) if the adaptive
```
roystgnr, line 284: This is just confusing to anyone who hasn't been buried in TwoStepTimeSolver for a month.  As far as we're concerned at this level of abstraction, the time step length is deltat.  Maybe there's some high p trickery going on, or it's been subdivided under the hood, or we tried some different deltats and settled on one that worked, but that's supposed to be effectively hidden at the level of the base class.  "Returns the change in system.time, deltat, for the last timestep which was successfully completed."  You could then elaborate with "This only returns the outermost step size in the case of nested time solvers." if you want.  But you should also add "If no time step has yet been successfully completed, then returns X", because that's not yet defined by any version of this comment, and at this point you're the only person who knows what X is.  :-D

include/solvers/time_solver.h
```diff
@@ -261,12 +280,33 @@ class TimeSolver : public ReferenceCountedObject<TimeSolver>
...
    */
-  virtual Real last_complete_deltat();
+  virtual Real last_completed_timestep_size();
+
+  /**
+   * Returns the system.deltat timestep used to get to the current time.
+   * For a uniform timestepper, it will simply return the last used system.deltat, for an
+   * adaptive time solver, it will return the last used substep size.
+   */
+  virtual Real last_completed_deltat();
```
roystgnr, line 294: And this smells like we're doing something wrong.  This is supposed to be the TimeSolver class.  Users at this level should be protected from digging into N details of N different subclasses except where that's absolutely unavoidable.

This description doesn't seem to match what you've got going on in the new example, either.  The time advanced by a solve is supposed to be the complete deltat size, not the size of the last substep.

include/solvers/time_solver.h
```diff
@@ -261,12 +280,33 @@ class TimeSolver : public ReferenceCountedObject<TimeSolver>
...
 
+  /**
+   * The final_time variable can be set by the user or in another library
```
roystgnr, line 297: Indeed, that is what "public:" means.

But what's important here isn't whether I *can* set this, it's whether I *have to* set this.  As written, this looks like a confusing booby trap.  I can't do adjoint time stepping without it?  Then how did ex5?  I can't do QoI postprocessing without it?  What are all those other examples doing?

I get that you're using this because we're too lazy to come up with a better API yet for QoI functionals that have delta functions in t, but that needs to be explained in the documentation, not just in my chat logs.  And if there are any *other* uses then add them too.  A new user should be able to read this comment and answer every question I asked before, and know exactly when set_final_time() needs to be vs doesn't need to be called.

include/solvers/time_solver.h
```diff
@@ -261,12 +280,33 @@ class TimeSolver : public ReferenceCountedObject<TimeSolver>
...
+   * function. By default the final_time variable is set to be std::nan("1").
```
roystgnr, line 298: Nobody cares about the default setting value for it's own sake what's important is the default *behavior*, and how to restore that if necessary.

Although... is this a safe default value?  C++ doesn't technically guarantee that an implementation supports quiet NaN, and we don't want to throw a SIGFPE if someone has "scream if we ever see NaN" settings enabled somehow.  If the only thing we really *need* this for is to handle delta function QoIs at the final_time then couldn't we just default it to ```numeric_limits<Real>::max()```?

include/solvers/twostep_time_solver.h
```diff
@@ -71,13 +71,27 @@ class TwostepTimeSolver : public AdaptiveTimeSolver
...
    * The midpoint rule is used to integrate each substep
    */
   virtual void integrate_adjoint_sensitivity(const QoISet & qois, const ParameterVector & parameter_vector, SensitivityData & sensitivities) override;
 
+  /**
+   * A method to compute the adjoint refinement error estimate at the current timestep.
+   * int_{tstep_start}^{tstep_end} R(u^h,z) dt
+   * Fills in an ErrorVector that contains the weighted sum of errors from ARefEE::qoi_set() and can be used to guide AMR.
+   * Also fills in a map that links QoI indices to spatially integrated error estimates for the QoI with that index.
+   * The midpoint rule is used for the numerical integration.
```
roystgnr, line 91: This is no longer correct, though, right?  We're now summing the substeps integrals, and using whatever our nested time solvers use, which in general might not be the midpoint rule?

src/solvers/unsteady_solver.C
```diff
@@ -241,7 +264,7 @@ void UnsteadySolver::retrieve_timestep()
 
 void UnsteadySolver::integrate_adjoint_sensitivity(const QoISet & qois, const ParameterVector & parameter_vector, SensitivityData & sensitivities)
 {
-  // We are using the midpoint rule to integrate each timestep
+  // We are using the trapezoidal rule to integrate each timestep
```
roystgnr, line 267: Good fix.  In hindsight - is this the right thing to do for all theta?  I'd think that we'd end up with significantly different weightings if we have an adaptive deltat and we want our integration rule to be consistent with the integrator we use for the transient solves.

include/solvers/unsteady_solver.h
```diff
@@ -115,13 +117,28 @@ class UnsteadySolver : public TimeSolver
...
    */
   virtual void integrate_adjoint_sensitivity(const QoISet & qois, const ParameterVector & parameter_vector, SensitivityData & sensitivities) override;
 
+  /**
+   * A method to compute the adjoint refinement error estimate at the current timestep.
+   * int_{tstep_start}^{tstep_end} R(u^h,z) dt
+   * Fills in an ErrorVector that contains the weighted sum of errors from all the QoIs and can be used to guide AMR.
+   * Also fills in a map that links QoI indices to spatially integrated error estimates for the QoI with that index.
+   * The trapezoidal rule (for temporally smooth QoI) or left Riemann sum (for an instantaneous QoI) is used
+   * for the numerical integration depending on the nature of the QoI.
```
roystgnr, line 138: Let's either just copy and paste the time_solver.h comment into its subclasses, or have any differences be due to subclass-specific underlying differences?  E.g., the documentation for QoI_time_instant is missing, so I can't quickly see it when looking at this class, yet there's so much documentation here that neither can I conclude that the real documentation must be elsewhere.  It's like antimissile chaff for doxygen users!

src/solvers/euler2_solver.C
```diff
@@ -257,6 +261,232 @@ bool Euler2Solver::_general_residual (bool request_jacobian,
   return jacobian_computed;
 }
 
+void Euler2Solver::integrate_qoi_timestep()
+{
+  // We are using the trapezoidal rule to integrate each timestep
+  // (f(t_j) + f(t_j+1))/2 (t_j+1 - t_j)
```
roystgnr, line 267: Looks like we're integrating consistently with theta below.  Update this comment accordingly?

examples/adjoints/adjoints_ex7/sigma_physics.C
vikramvgarg: Went with T_var everywhere.

examples/adjoints/adjoints_ex7/factoryfunction.C
vikramvgarg: Removed.

src/solvers/unsteady_solver.C
vikramvgarg: Actually, this function should probably be in Euler/2Solver.C and follow the scheme of integrate_qoi_timestep.

vikramvgarg:
This PR getting reviewed by Roy: https://www.youtube.com/watch?v=OlZsCYnnICA&t=6m48s

roystgnr:
I was trying for a much more metaphorical skewering...

And not to raise any anxiety, but there's still 6 files I haven't finished going over yet, so there may be more coming tonight.  :stuck_out_tongue:

src/solvers/file_solution_history.C
```diff
@@ -28,14 +28,34 @@
...
+  : stored_sols(stored_solutions.end()),
+  _system(system_), localTimestamp(0),
+  timeTotimestamp()
+  {
+    dual_solution_copies.resize(system_.n_qois());
+
+    for(auto j : make_range(system_.n_qois()))
+    {
+      dual_solution_copies[j] = nullptr;
+    }
```
roystgnr, line 45: Doesn't resize default-initialize the new entries to nullptr already?

src/solvers/time_solver.C
```diff
@@ -124,9 +135,39 @@ void TimeSolver::integrate_adjoint_sensitivity(const QoISet & qois, const Parame
...
+void TimeSolver::integrate_adjoint_refinement_error_estimate
+  (AdjointRefinementEstimator & adjoint_refinement_error_estimator,
+   ErrorVector & QoI_elementwise_error,
+   std::vector<Real *> /* QoI_time_instant */)
+{
+  // Make sure the system::qoi_error_estimates vector is of the same size as system::qoi
+  if(_system.qoi_error_estimates.size() != _system.qoi.size())
+      _system.qoi_error_estimates.resize(_system.qoi.size());
+
+  // Base class assumes a direct steady state error estimate
```
roystgnr, line 147: So let's move this to SteadySolver, and put libmesh_not_implemented() in the base class?  Better to get an error message than an incorrect solution if someone's got a third-party derived class or if we miss an implementation in one of our own derived classes.

src/solvers/twostep_time_solver.C
```diff
@@ -300,27 +301,67 @@ std::pair<unsigned int, Real> TwostepTimeSolver::adjoint_solve (const QoISet & q
...
-  // Tell our timesolver the combined deltat before adjoint advance timestep updates it
-  this->completedtimestep_deltat = 2.0*_system.deltat;
+  // Record the sub step deltat we used for the last adjoint solve.
+  last_deltat = _system.deltat;
 
-  // Adjoint advance again so solution history saves the adjoint
-  core_time_solver->adjoint_advance_timestep();
+  // Record the total size of the last timestep, for a 2StepTS, this is
+  // simply twice the deltat for each sub(half) step.
+  this->completed_timestep_size = 2.0*_system.deltat;
```
roystgnr, line 318: We are totally handling system.deltat in an inconsistent way here, but that's an old design flaw and my fault, not anything wrong with this PR.

src/solvers/twostep_time_solver.C
```diff
@@ -300,27 +301,67 @@ std::pair<unsigned int, Real> TwostepTimeSolver::adjoint_solve (const QoISet & q
...
+  for (auto j : make_range(_system.n_qois()))
+  {
+    (_system.qoi)[j] = qois_first_half[j] + qois_second_half[j];
+  }
+}
+
 void TwostepTimeSolver::integrate_adjoint_sensitivity(const QoISet & qois, const ParameterVector & parameter_vector, SensitivityData & sensitivities)
 {
-  // We are using the midpoint rule to integrate each timestep
+  // We are using the trapezoidal rule to integrate each timestep, and then pooling the contributions here.
```
roystgnr, line 364: Not trapezoidal, but rather the core time solver rule, whatever that may be.

src/solvers/twostep_time_solver.C
```diff
@@ -339,4 +380,57 @@ void TwostepTimeSolver::integrate_adjoint_sensitivity(const QoISet & qois, const
      sensitivities[i][j] = sensitivities_first_half[i][j] + sensitivities_second_half[i][j];
 }
 
+void TwostepTimeSolver::integrate_adjoint_refinement_error_estimate(AdjointRefinementEstimator & adjoint_refinement_error_estimator, ErrorVector & QoI_elementwise_error, std::vector<Real *> QoI_time_instant)
+{
+  // We are using the midpoint rule or right sided Riemman rule to integrate each timestep
```
roystgnr, line 385: Likewise here, right?

src/solvers/twostep_time_solver.C
roystgnr: I still haven't come up with much better names, so we can leave it.  And the qoi_error_estimates documentation in system.h is clear enough.  Can you edit the integrate_adjoint_refinement_error_estimate doc to say something clearer than "A method to compute the adjoint refinement error estimate at the current timestep", though?  We're not actually estimating the error here, we're estimating the contribution of the error from the current timestep, right?  That needs to be absolutely clear in the doxygen.

src/solvers/file_solution_history.C
vikramvgarg: Fixed.

src/solvers/time_solver.C
vikramvgarg: We will need to do this for the other integrate_timestep functions as well, right ?

src/solvers/twostep_time_solver.C
vikramvgarg: Yes, this was fixed but see above.

src/solvers/twostep_time_solver.C
vikramvgarg: For the sensitivity I have not yet moved things to the core timesolver classes. I will modify the comment here and we do that in another PR ?

src/solvers/euler_solver.C
```diff
@@ -190,5 +194,232 @@ bool EulerSolver::_general_residual (bool request_jacobian,
   return jacobian_computed;
 }
 
+void EulerSolver::integrate_qoi_timestep()
+{
+  // We are using the trapezoidal rule to integrate each timestep
```
roystgnr, line 199: Another spot to point out that we're using theta now.

vikramvgarg: Yup, got this.

src/solvers/time_solver.C
roystgnr: Right.

vikramvgarg: Done.

src/solvers/euler2_solver.C
vikramvgarg: Done.

include/solvers/unsteady_solver.h
vikramvgarg: Done.

examples/adjoints/adjoints_ex7/femparameters.C
vikramvgarg: Done.

examples/adjoints/adjoints_ex7/initial.C
vikramvgarg: Done.

include/solvers/euler2_solver.h
vikramvgarg: Done.

include/solvers/euler_solver.h
vikramvgarg: Done.

include/solvers/time_solver.h
vikramvgarg: Done.

include/solvers/time_solver.h
vikramvgarg: Purged this function and replaced its use with last_completed_timestep_size in adjoints example 5.

include/solvers/time_solver.h
vikramvgarg: We initialize last_deltat and completed_timestep_size with s.deltat  in the constructor now.

include/solvers/time_solver.h
vikramvgarg: Done.

src/solvers/unsteady_solver.C
vikramvgarg: Added a comment to this end.

src/solvers/twostep_time_solver.C
roystgnr: Sounds good.

roystgnr (approved):
It took a full 60 commits and 111 comments on this PR, but I can no longer find anything to complain about that's significant enough to impede a merge.

@jwpeterson - if you want to try cleaning up the history here (and oh boy is there a lot of cleanup possible), I think we're ready to go.

vikramvgarg:
> It took a full 60 commits and 111 comments on this PR, but I can no longer find anything to complain about that's significant enough to impede a merge.
> 
> @jwpeterson - if you want to try cleaning up the history here (and oh boy is there a lot of cleanup possible), I think we're ready to go.

@jwpeterson let me know if I can do anything to help with the cleanup.

jwpeterson:
I definitely do still plan on taking a crack at cleaning up the history here. Mainly just to eliminate as many of the intermediate failed builds as possible (preferably all of them). I may be able to get to it some time this evening.

jwpeterson:
OK, I've rebased all the commits on top of current master and combined some together (e.g. the Example 7 files are all added in a single commit now). In the process of rebasing, there was only one file which had conflicts, adaptive_time_solver.h. Since I wasn't sure exactly how to fix them during the rebase, I just applied the version of that file from the original branch (see e9124a0), at the end of the rebase, so hopefully that's fine.

vikramvgarg:
> OK, I've rebased all the commits on top of current master and combined some together (e.g. the Example 7 files are all added in a single commit now). In the process of rebasing, there was only one file which had conflicts, adaptive_time_solver.h. Since I wasn't sure exactly how to fix them during the rebase, I just applied the version of that file from the original branch (see [e9124a0](https://github.com/libMesh/libmesh/commit/e9124a0b95892f6307b100989e86316dc63cb6f2)), at the end of the rebase, so hopefully that's fine.

@jwpeterson is the branch ready to merge ?

jwpeterson:
Yeah I just figured you guys would want to review the rebase one last time before merging.

PR #2687, opened 2020-08-25
Fix PeriodicBoundaries where elements have multiple sides on boundary

roystgnr:
@arovinelli hit a case the other day that I hadn't accounted for but should have: it's possible (and with Tets or Tris, easy) to have a mesh in which two sides of the same element are on the same sideset of the boundary.  Previously, if this was the case with a periodic boundary, our PeriodicBoundaries code would fail to set up the periodic constraint equations correctly.

This PR adds a new BoundaryInfo helper function to make it easier to examine multiple sides on a boundary, plus an optional PeriodicBoundaries::neighbor() entry to return side index information in cases where we need to know exactly which side is the corresponding side across a periodic boundary, and uses the latter to make sure we get the correct side when building periodic dof constraints and node constraints.

This needs testing before merge.

### DIFF
--- a/include/base/periodic_boundaries.h
+++ b/include/base/periodic_boundaries.h
@@ -59,12 +59,18 @@ public:
 
   ~PeriodicBoundaries();
 
-  // The periodic neighbor of \p e in direction \p side, if it
+  // \returns the periodic neighbor of \p e in direction \p side, if it
   // exists, nullptr otherwise.
+  //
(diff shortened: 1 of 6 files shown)


arovinelli:
@roystgnr this is cool. Thanks for working on this
Can we add a test with the shark-teeth mesh I sent you the other day?

roystgnr:
Obsoleted by #2695.

PR #2688, opened 2020-08-26
Subdomain expansion

roystgnr:
Recreating #2674 to see if we can get a CI check on it triggered.


roystgnr:
Kicking this one more time, now that the webhook settings are fixed and it seems to have connected.

PR #2690, opened 2020-08-26
Subdomain expansion, new branch

roystgnr:
One more attempt to trigger CI

### DIFF
--- a/include/systems/generic_projector.h
+++ b/include/systems/generic_projector.h
@@ -819,9 +819,27 @@ public:
 
     libmesh_assert_equal_to (old_indices.size(), indices.size());
 
-    values.resize(old_indices.size());
+    // We may have invalid_id in cases where no old DoF existed, e.g.
+    // due to expansion of a subdomain-restricted variable's subdomain
+    bool invalid_old_index = false;
+    for (const auto & di : old_indices)
+      if (di == DofObject::invalid_id)
+        invalid_old_index = true;
 
-    old_solution.get(old_indices, values);
+    values.resize(old_indices.size());
+    if (invalid_old_index)
+      {
+        for (auto i : index_range(old_indices))
+          {
+            const dof_id_type di = old_indices[i];
+            if (di == DofObject::invalid_id)
+              values[i] = 0;
+            else
+              values[i] = old_solution(di);
+          }
+      }
+    else
+      old_solution.get(old_indices, values);
   }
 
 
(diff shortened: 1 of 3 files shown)


roystgnr:
Still no pass.  Unless we're getting lucky elsewhere there might be something in this PR tripping things?  I'll kick it one more time just in case we're getting bitten by a race condition (I'm pretty sure we *have* a race condition in one new unit test, in fact, but I've never triggered it on less than 3 processors and it manifests for me as a nice failure message not a SIGKILL), then I'll try bisecting this.

roystgnr:
Since other branches are passing, I'm going to try bisecting here a bit and see if we can narrow down what commit was causing this one to fail.  The original subomain_expansion branch has all the commits; I'll be stripping off and adding on to this one.

roystgnr:
That was a pass on https://github.com/libMesh/libmesh/pull/2690/commits/6624e69af4af0e03ebf262eecff503ff8f39623f - for future reference in case I accidentally clear the green check with a later force push.

roystgnr:
And then a quick failure from f8109f7

roystgnr:
And then a pass from 29323bd.  There might be a bit of a red herring here, if f8109f7 created a problem that a later commit fixed.  But those later commits should reorder cleanly, and we don't yet have test coverage in here depending on f8109f7, so I'll try bisecting with that one commit left out.

roystgnr:
Okay, then - df9624a includes every commit *except* f8109f7, and that passes.  I'll try one last time with the whole set to make sure it's failing.

roystgnr:
Still failing, but now with actual assertion failures instead of SIGKILL, and I can actually reproduce the failures now without getting sidetracked by the other unit test bugs, so I'm finally optimistic here.

roystgnr:
Hopefully I've got it now.  The fixup for the last commit here is supposed to be backwards compatible in cases other than the one (h-refinement with high-p elements) that broke, but @dewenyushu is going to want to double-check and make sure I didn't just regress the subdomain expansion capability that motivated this PR in the first place.

dewenyushu:
> Hopefully I've got it now. The fixup for the last commit here is supposed to be backwards compatible in cases other than the one (h-refinement with high-p elements) that broke, but @dewenyushu is going to want to double-check and make sure I didn't just regress the subdomain expansion capability that motivated this PR in the first place.

Hi @roystgnr , thanks for checking with me! I just tested this PR with my existing test cases for subdomain expansion. Everything looks fine so far!

jwpeterson (approved):

PR #2692, opened 2020-08-27
DO NOT MERGE TO MASTER

roystgnr:
This reverts weeks of commits, going back to the state at the merge of
PR 2668, to see if that works on the new fallback Civet

I'm setting the pull request to target "not_master" just in case someone accidentally mis-clicks.


roystgnr:
Okay, this PR is pointless.  #2693 has now passed, and it doesn't include any bug fixes, so our master branch must be our-current-ducttape-based-CI compatible.  Either #2690 introduced a bug in its own branch, or the CI itself is now working better than it was previously.

Issue #2635, opened 2020-07-01
Upgrade C++ standard support


roystgnr:
Much discussion on Slack with @lindsayad and @jwpeterson seems to boil down to:

1.  In the short term, we need --enable-cxx14 and --enable-cxx14-required arguments, working the way their cxx11 equivalents currently do, so MOOSE can make it easier for its users to identify the problem and recover if their upcoming C++14 code runs into a user's environment which lacks (and/or requires compiler flags for) C++14 support

2.  We ought to be generalizing this in the long term.  Something like e.g. ```--enable-cxx-maximum c++20 --enable-cxx-minimum c++14``` if we want configure to try to enable C++20 but to settle for C++17 or C++14 if that's all that's supported.  With no maximum argument we'd just use our current behavior, where we let the compiler defaults rule except when we find that our minimum standard requirement is unsupported without an explicit compiler argument.

3.  At that point we ought to have a simple ```--enable-cxx-std``` for people who want to tersely set the same minimum and maximum: i.e. use the exact std we specify or scream and die.

4.  Modern C++ compilers (especially g++/clang++) are pretty good about not half-supporting core language features unless you specifically request a still-under-development standard, so we can probably get away with testing a few features (and possibly the ```__cplusplus``` macro) and calling a passed result "C++XX support" for any XX newer than 11.

5.  Modern C++ standard libraries (even libstdc++/libc++) seem to get more piecemeal support, so for those we might have to do what we did with C++11, adding more tests as we or our users rely on more library features.

6.  C++14 doesn't seem quite tempting enough to make it a minimum requirement for libMesh even when MOOSE upgrades.

7.  C++17 has some syntactic sugar which is very tempting, so we may want to upgrade libMesh directly to that once we're comfortable that we won't be leaving any users out in the cold, but that's not going to be in the near future.

roystgnr:
A quick update on (7):

One of my reasons for wanting to stick with C++11 for so long was that I know we had users stuck on RHEL7 and derivative systems, for which GCC 4.8.x was the standard and there was no C++14 support out of the box.

But, RHEL8 came out mid-2020, it has GCC 8 (which supports all of the C++17 language improvements and most of the standard library improvements) by default, and it makes GCC 9 (which supports it all) an easy install, so unless there are other widely used environments with "C++14 but not C++17" support, I think "just jump to C++17" really ought to be our plan, I'd say by the end of this year or early next.

PR #2693, opened 2020-08-27
Add methods to synchronize side boundary ids on the fly

dewenyushu:
This PR adds the capability to synchronize boundary side changes across processors. This capability is realized via `sync_pull_boundary_side_id()`. In the case when boundary sides of a ghost element is removed, the `sync_push_remove_boundary_side_id()` method is utilized to communicate with the processor that owns this element.

This is an initial implementation of this method that is just enough to fulfill my use case. @fdkong @roystgnr I would really appreciate it if you would provide some suggestions for making it mergeable :)

### DIFF
--- a/src/mesh/boundary_info.C
+++ b/src/mesh/boundary_info.C
@@ -1806,24 +1806,7 @@ BoundaryInfo::build_node_list_from_side_list()
      node_bcid_action_functor, datum_type_ex);
 }
 
-void BoundaryInfo::sync_push_remove_boundary_side_id(
-  std::unordered_map<processor_id_type, std::vector<std::pair<dof_id_type, unsigned int>>>
-  & elems_to_push)
-{
-  auto elem_action_functor =
-    [this]
-    (processor_id_type,
-     const std::vector<std::pair<dof_id_type, unsigned int>> & received_elem)
-    {
-      for (const auto & pr : received_elem)
-        this->remove_side(_mesh.elem_ptr(pr.first), pr.second);
-    };
-
-  Parallel::push_parallel_vector_data
-    (this->comm(), elems_to_push, elem_action_functor);
-}
-
-void BoundaryInfo::sync_pull_boundary_side_id()
+void BoundaryInfo::sync_boundary_side_ids()
 {
   // we need BCs for ghost elements.
   std::unordered_map<processor_id_type, std::vector<dof_id_type>>
--- a/src/mesh/boundary_info.C
+++ b/src/mesh/boundary_info.C
@@ -1806,7 +1806,7 @@ BoundaryInfo::build_node_list_from_side_list()
      node_bcid_action_functor, datum_type_ex);
 }
 
-void BoundaryInfo::sync_boundary_side_ids()
+void BoundaryInfo::parallel_sync_side_ids()
 {
   // we need BCs for ghost elements.
   std::unordered_map<processor_id_type, std::vector<dof_id_type>>
@@ -1862,7 +1862,62 @@ void BoundaryInfo::sync_boundary_side_ids()
      elem_id_action_functor, datum_type_ex);
 }
 
+void BoundaryInfo::parallel_sync_node_ids()
+{
+  // we need BCs for ghost nodes.
+  std::unordered_map<processor_id_type, std::vector<dof_id_type>>
+    node_ids_requested;
 
+  // Determine what nodes we need to request
+  for (const auto & node : _mesh.node_ptr_range())
+  {
+    const processor_id_type pid = node->processor_id();
+    if (pid != this->processor_id())
+      node_ids_requested[pid].push_back(node->id());
+  }
+
+  typedef std::vector<boundary_id_type> datum_type;
+
+  // gather the node ID and boundary_id_type for the ghost nodes
+  auto node_id_gather_functor =
+  [this]
+  (processor_id_type,
+    const std::vector<dof_id_type> & ids,
+    std::vector<datum_type> & data)
+    {
+      data.resize(ids.size());
+      for (auto i : index_range(ids))
+      {
+        Node * node = _mesh.node_ptr(ids[i]);
+        for (const auto & pr : as_range(_boundary_node_id.equal_range(node)))
+        data[i].push_back(pr.second);
+      }
+    };
+
+    // update the _boundary_node_id on this processor
+    auto node_id_action_functor =
+      [this]
+      (processor_id_type,
+       const std::vector<dof_id_type> & ids,
+       std::vector<datum_type> & data)
+    {
+        for (auto i : index_range(ids))
+        {
+          Node * node = _mesh.node_ptr(ids[i]);
+          //clear boundary node
+          _boundary_node_id.erase(node);
+          // update boundary node
+          for (const auto & pr : data[i])
+            _boundary_node_id.insert(std::make_pair(node, pr));
+        }
+    };
+
+
+    datum_type * datum_type_ex = nullptr;
+    Parallel::pull_parallel_vector_data
+      (this->comm(), node_ids_requested, node_id_gather_functor,
+       node_id_action_functor, datum_type_ex);
+}
 
 void BoundaryInfo::build_side_list_from_node_list()
 {
--- a/src/mesh/boundary_info.C
+++ b/src/mesh/boundary_info.C
@@ -1806,6 +1806,78 @@ BoundaryInfo::build_node_list_from_side_list()
      node_bcid_action_functor, datum_type_ex);
 }
 
+void BoundaryInfo::sync_push_remove_boundary_side_id(
+  std::unordered_map<processor_id_type, std::vector<std::pair<dof_id_type, unsigned int>>>
+  & elems_to_push)
+{
+  auto elem_action_functor =
+    [this]
+    (processor_id_type,
+     const std::vector<std::pair<dof_id_type, unsigned int>> & received_elem)
+    {
+      for (const auto & pr : received_elem)
+        this->remove_side(_mesh.elem_ptr(pr.first), pr.second);
+    };
+
+  Parallel::push_parallel_vector_data
+    (this->comm(), elems_to_push, elem_action_functor);
+}
+
+void BoundaryInfo::sync_pull_boundary_side_id()
+{
+  // we need BCs for ghost elements.
+  std::unordered_map<processor_id_type, std::vector<dof_id_type>>
+    elem_ids_requested;
+
+  // Determine what elements we need to request
+  for (const auto & elem : _mesh.element_ptr_range())
+  {
+    const processor_id_type pid = elem->processor_id();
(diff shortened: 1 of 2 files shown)


roystgnr:
First of all - don't futz with this until CI is finished, please!?  It looks as if you're somehow immune to whatever problem plagued #2690 and that I haven't yet been able to reproduce in #2692 by bisecting backwards to the Before-broken-CI Times, so if this passes then either there's something *specifically* wrong with #2690 or there was simply something wrong with the new CI server but Jason's configuration changes have now fixed it.

roystgnr:
I'm liking sync_pull_boundary_side_id.  I think I'd make that name plural, and we definitely need a better comment documenting it, but no real criticisms.

But sync_push_remove_boundary_side_id might be too your-one-use-case specific to put in libMesh?  I'm not sure what the right solution to that is though.  If we leave it as is then it does look like the most efficient implementation for your problem, but you'll probably be the only one using it, so we might as well just put it into your code rather than into a libMesh header.  If we were to make it into a general API for updating boundary ids on ghost element sides (for arbitrary mixtures of adding and removing them) then I think that'd be widely usable and worth making a general API for ... but that API might be less efficient for your particular case (it'd involve the same amount of latency but a little more bandwidth) and it'd be more tricky to get test coverage of the cases you're not needing to exercise.

But don't take the above paragraph too seriously until we hear what @fdkong thinks, since he was talking about using these APIs in his applications too.  If he only needs the pull, or if he needs a push but his cases are more general, then I stand by what I said, but if he chimes in with "actually yeah I also sometimes need to push remove_side calls alone" then I'm going to find it hard to keep arguing that it's a rare use case.

dewenyushu:
Thanks @roystgnr ! I agree the push is not general enough to put into libmesh. A pull function that synchronizes all the side info for the ghost elements (remove and add at the same time) would probably be more useful in this case. But yes, let's hear what @fdkong thinks, and I am glad to make the change.

src/mesh/boundary_info.C
```diff
@@ -1806,6 +1806,78 @@ BoundaryInfo::build_node_list_from_side_list()
...
+    [this]
+    (processor_id_type,
+     const std::vector<std::pair<dof_id_type, unsigned int>> & received_elem)
+    {
+      for (const auto & pr : received_elem)
+        this->remove_side(_mesh.elem_ptr(pr.first), pr.second);
+    };
+
+  Parallel::push_parallel_vector_data
+    (this->comm(), elems_to_push, elem_action_functor);
```
fdkong, line 1823: Where you are using `sync_push_remove_boundary_side_id` @dewenyushu ? I thought boundary sides could be removed locally (without communication). And then `sync_pull_boundary_side_id` should update ghosted elements correctly. 

I might be missing something here.

src/mesh/boundary_info.C
```diff
@@ -1806,6 +1806,78 @@ BoundaryInfo::build_node_list_from_side_list()
...
+}
+
+void BoundaryInfo::sync_pull_boundary_side_id()
+{
```
fdkong, line 1827: `sync_boundary_side_ids`?

src/mesh/boundary_info.C
```diff
@@ -1806,6 +1806,78 @@ BoundaryInfo::build_node_list_from_side_list()
...
+    const processor_id_type pid = elem->processor_id();
+    if (pid != this->processor_id())
+      elem_ids_requested[pid].push_back(elem->id());
+  }
+
+  typedef std::vector<std::pair<unsigned short int, boundary_id_type>> datum_type;
+
+  // gather the element ID, side, and boundary_id_type for the ghost elements
+  auto elem_id_gather_functor =
+    [this]
```
fdkong, line 1844: Change` this to _mesh`? I might  not need to expose the entire object.

src/mesh/boundary_info.C
```diff
@@ -1806,6 +1806,78 @@ BoundaryInfo::build_node_list_from_side_list()
...
+    for (auto i : index_range(ids))
+    {
+      Elem * elem = _mesh.elem_ptr(ids[i]);
+      for (const auto & pr : as_range(_boundary_side_id.equal_range(elem)))
+        data[i].push_back(std::make_pair(pr.second.first, pr.second.second));
+    }
+  };
+  // update the _boundary_side_id on this processor
+  auto elem_id_action_functor =
+    [this]
```
fdkong, line 1859: `& _boundary_side_id, &_mesh`

fdkong:
If I understand correctly, `sync_push_remove_boundary_side_id` is not necessary even for @dewenyushu.

src/mesh/boundary_info.C
dewenyushu: The push is needed in my application because I need to remove the neighbor element side from the boundary. The neighbor element is often a ghost element so I need to communicate this change by push. 

Roy and I are curious if you may need similar functionalities in your application (i.e., remove/add sides for the ghost elements)? So we could discuss whether this function should be kept in libmesh.

roystgnr:
I think the sync_push code is for cases where it's really "push, then sync":  The owner of element e needs its side s removed, but it doesn't *know* that, only the owner of ```e->neighbor_ptr(s)``` does.  So a pure "sync", where we update ghosts to match their owners, would be backwards - what we want is for, in some specific cases, ghosts to be changed and update their owners, and only after that happens do we do the sync step.

fdkong:
> If I understand correctly, `sync_push_remove_boundary_side_id` is not necessary even for @dewenyushu.

I, obviously, did not understand correctly :-)

I see, if a boundary surface lies on a processor-boundary, we need to change the current element as well as the remote neighbors.

roystgnr:
I guess "Fande can't immediately conceive of use cases for push_remove" implies "Fande doesn't have more use cases for push_remove"?  ;-)  So it sounds like the way to go from here is tighten up the lambda function variable capture, strip the push function out of here to use in MOOSE or application code, then we can merge the pull function (which I also prefer the shorter name for, though either way works) into libMesh?

dewenyushu:
> I guess "Fande can't immediately conceive of use cases for push_remove" implies "Fande doesn't have more use cases for push_remove"? ;-) So it sounds like the way to go from here is tighten up the lambda function variable capture, strip the push function out of here to use in MOOSE or application code, then we can merge the pull function (which I also prefer the shorter name for, though either way works) into libMesh?

lol I agree it seems Fande does not have a use case for the 'push_remove' right now. I do feel that a more generalized form of "push_remove" (e.g., push_remove_and_add) would be worthwhile putting into libmesh in the future. But yeah, I will remove this function from this PR and make the changes based on your comments.

dewenyushu:
@fdkong Emm seems I am not understanding lambda well enough for this:
`error: invalid use of non-static data member 'libMesh::BoundaryInfo::_boundary_side_id'`
`error: invalid use of non-static data member 'libMesh::BoundaryInfo::_mesh'`

Does this mean instead of '[&_mesh]', I should be using something like '[&_mesh = this->get_mesh()]' and similarly for '_boundary_side_id'? But it seems like we do not have the corresponding get methods... should I just switch back to '[this]'?

fdkong:
> @fdkong Emm seems I am not understanding lambda well enough for this:
> `error: invalid use of non-static data member 'libMesh::BoundaryInfo::_boundary_side_id'`
> `error: invalid use of non-static data member 'libMesh::BoundaryInfo::_mesh'`
> 
> Does this mean instead of '[&_mesh]', I should be using something like '[&_mesh = this->get_mesh()]' and similarly for '_boundary_side_id'? But it seems like we do not have the corresponding get methods... should I just switch back to '[this]'?

Yes, I did not notice they are private members.  Just go back to `[this]`.

My head does not work these days maybe because of the coming winter that makes a bit sad.

Sorry!

dewenyushu:
> > @fdkong Emm seems I am not understanding lambda well enough for this:
> > `error: invalid use of non-static data member 'libMesh::BoundaryInfo::_boundary_side_id'`
> > `error: invalid use of non-static data member 'libMesh::BoundaryInfo::_mesh'`
> > Does this mean instead of '[&_mesh]', I should be using something like '[&_mesh = this->get_mesh()]' and similarly for '_boundary_side_id'? But it seems like we do not have the corresponding get methods... should I just switch back to '[this]'?
> 
> Yes, I did not notice they are private members. Just go back to `[this]`.
> 
> My head does not work these days maybe because of the coming winter that makes a bit sad.
> 
> Sorry!

lol this or lambda is just hard!

roystgnr:
> I should be using something like '[&_mesh = this->get_mesh()]'

No, I'm afraid that's C++14-only, and we're still requiring C++11 compatibility.  You could declare my_mesh above the lambda and then just capture my_mesh, but that's more hassle for just a bit of encapsulation; I'd say [this] is fine too.

fdkong:
> No, I'm afraid that's C++14-only, and we're still requiring C++11 compatibility. You could declare my_mesh above the lambda and then just capture my_mesh, but that's more hassle for just a bit of encapsulation; I'd say [this] is fine too.

Does [&_mesh = _mesh] work? Will involve a copy constructor?

roystgnr:
If it worked it definitely shouldn't invoke a copy constructor, but pre-C++14 I don't think it's legal syntax.

fdkong:
> If it worked it definitely shouldn't invoke a copy constructor, but pre-C++14 I don't think it's legal syntax.

haha, we decided to move to C++14 in moose a while ago. Do you have any plan on libmesh side?

roystgnr:
We have some discussion notes in #2635 - basically we need to improve our C++14 detection in autotools, but we'll probably jump straight from C++11 to C++17 when our users are ready for that.

roystgnr:
Actually, in light of the recent Slack discussion - should we just make the code here part of a single ```sync_boundary_ids()``` method that does a parallel pull of node (and edge and shellface) bcids at the same time?

roystgnr:
Of course, the issue with that is there's no equally-simple and equally-efficient way to do it (separate communication steps for each type of bcid would be less efficient, encoding them into the same data structure would be less simple).  Maybe we'll just want to add individual node/edge/shellface options, knowing that most applications will be able to only call the one or two they need?

roystgnr:
And as a last thought - if that's what we end up doing, is "sync_boundary_side_ids" the best name schema?  "sync" is a bit overloaded with our other BoundaryInfo methods that sync info from an interior mesh to a boundary (and/or internal sideset) mesh, so perhaps "parallel_sync" would clear that up.  It also now occurs to me that "boundary" is redundant in a method of a class already named ```BoundaryInfo```, but that ship has **clearly** already sailed so we might as well leave it in for consistency.

dewenyushu:
@roystgnr thanks for the suggestion! I add a method to sync boundary nodes. I have not yet added the edge/shellface options because: 1) I currently do not have a use case for it so it is hard to tell how that would work, 2) it seems like templating these with the side option could avoid some code redundancy.

Also, I agree the naming without 'boundary' in it :)

PR #2697, opened 2020-09-01
Eigen solver initial guess setup

fdkong:
Add an API to allow us to set up an initial guess in slepc eigen solver.

### DIFF
--- a/src/solvers/slepc_eigen_solver.C
+++ b/src/solvers/slepc_eigen_solver.C
@@ -39,7 +39,8 @@ namespace libMesh
 
 template <typename T>
 SlepcEigenSolver<T>::SlepcEigenSolver (const Parallel::Communicator & comm_in) :
-  EigenSolver<T>(comm_in)
+  EigenSolver<T>(comm_in),
+  _initial_space(nullptr)
 {
   this->_eigen_solver_type  = ARNOLDI;
   this->_eigen_problem_type = NHEP;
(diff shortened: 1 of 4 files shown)


src/solvers/slepc_eigen_solver.C
```diff
@@ -970,21 +989,14 @@ void SlepcEigenSolver<T>::set_initial_space(NumericVector<T> & initial_space_in)
...
   libmesh_error_msg("SLEPc 3.1 is required to call EigenSolver::set_initial_space()");
 #else
-  this->init();
-
-  PetscErrorCode ierr = 0;
-
   // Make sure the input vector is actually a PetscVector
   PetscVector<T> * initial_space_petsc_vec =
-    dynamic_cast<PetscVector<T> *>(&initial_space_in);
+  dynamic_cast<PetscVector<T> *>(&initial_space_in);
```
jwpeterson, line 994: I see this is just a whitespace change, but while you are at it, you may as well change this to a `cast_ptr` call and drop the following `libmesh_error_msg_if` line, since the error check is then done internally. So the whole thing would be in one line:
```
_initial_space = cast_ptr<PetscVector<T> *>(&initial_space_in);
```

fdkong: Sorry, I somehow missed this. I thought we could not merge anything in libmesh either and did not pay attention to my PR.

Thanks for the fix @jwpeterson

jwpeterson: No worries, I will merge once everything passes again.

jwpeterson (approved):

PR #2700, opened 2020-09-08
Fix index selection in PeriodicBoundary code

roystgnr:
This alone wasn't sufficient to fix the problem that @arovinelli ran
into with their hand-altered neighbor_ptr topologies, but it might at
least be a corner-case fix for someone else.

### DIFF
--- a/src/fe/fe_base.C
+++ b/src/fe/fe_base.C
@@ -2114,7 +2114,7 @@ compute_periodic_constraints (DofConstraints & constraints,
                           if (!elem->is_node_on_side(n,new_s))
                             continue;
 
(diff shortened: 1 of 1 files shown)


roystgnr:
I can reproduce this, and it looks like it actually interacts with a *different* bug, so by fixing one case I've made a more common case worse.  Going to close this and get back to work on actually replacing this code with something better rather than just patching it up.

arovinelli:
> I can reproduce this, and it looks like it actually interacts with a _different_ bug, so by fixing one case I've made a more common case worse. Going to close this and get back to work on actually replacing this code with something better rather than just patching it up.

It sounds like a good idea

PR #2711, opened 2020-09-22
Add some vector CompareTypes

lindsayad:
Documentation for `CompareTypes` is:
```
// Operators using different but compatible types need a return value
// based on whichever type the other can be upconverted into.  For
// instance, the proper return type for
// TypeVector<float>::operator*(double) is TypeVector<double>.  In
// general, an operation using types S and T should return a value
// based on CompareTypes<S,T>::supertype
```
So what I've implemented here wouldn't be valid for all elementary binary operations, e.g. for multiplication, but I think this holds true to the "upconverted" concept, and obviously would hold true for + and - operations (which is what motivated this PR).

### DIFF
--- a/include/numerics/type_vector.h
+++ b/include/numerics/type_vector.h
@@ -1131,6 +1131,18 @@ TypeVector<T> TypeVector<T>::unit() const
 #endif
 
 }
+
+template <typename T>
+struct CompareTypes<TypeTensor<T>, TypeTensor<T>>
+{
+  typedef TypeTensor<T> supertype;
+};
+
+template <typename T, typename T2>
+struct CompareTypes<TypeTensor<T>, TypeTensor<T2>>
+{
+  typedef TypeTensor<typename CompareTypes<T,T2>::supertype> supertype;
+};
 } // namespace libMesh
 
 namespace std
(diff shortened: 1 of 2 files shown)


roystgnr:
I'm okay with this.  The one thing that really bothers me is the inconsistency with operator* - we've always defined that (perhaps being too clever for our own good, but too late now) to mean "inner product", but the result here is going to interpret it as an operator into the space of vectors e.g. element-wise multiplication.  That should be a compile-time error if/when we ever trigger it, though, not anything disastrous.  @jwpeterson?

jwpeterson:
> @jwpeterson?

Looks like template black magic to me, I'm in.

include/numerics/type_vector.h
```diff
@@ -1131,6 +1131,18 @@ TypeVector<T> TypeVector<T>::unit() const
...
+struct CompareTypes<TypeTensor<T>, TypeTensor<T>>
+{
+  typedef TypeTensor<T> supertype;
+};
+
+template <typename T, typename T2>
+struct CompareTypes<TypeTensor<T>, TypeTensor<T2>>
+{
+  typedef TypeTensor<typename CompareTypes<T,T2>::supertype> supertype;
+};
```
lindsayad, line 1145: Omg, I guess I never actually do addition/subtraction of `TypeVectors`. What a terrible typo

PR #2712, opened 2020-09-23
Use -> to get a member from a pointer

roystgnr:
This regressed in #2677

This method is templated and not being instantiated, so I guess my
compiler and our CI compilers didn't bother to check the code's validity
when seeing vector<T> with an indeterminate T, rather than realizing
that a pointer to *any* object can't have a size() method.

This will need to be backported to the release branch, for the sake of
users who want to use this method or who have smarter compilers than
mine.  Thanks to @dschwen for reporting this.

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -2032,7 +2032,7 @@ std::vector<unsigned int> MeshBase::add_elem_data(const std::vector<std::string>
                                                   bool allocate_data,
                                                   const std::vector<T> * default_values)
 {
-  libmesh_assert(!default_values || default_values.size() == names.size());
+  libmesh_assert(!default_values || default_values->size() == names.size());
 
   std::vector<unsigned int> returnval(names.size());
 
@@ -2081,7 +2081,7 @@ std::vector<unsigned int> MeshBase::add_node_data(const std::vector<std::string>
                                                   bool allocate_data,
                                                   const std::vector<T> * default_values)
 {
-  libmesh_assert(!default_values || default_values.size() == names.size());
+  libmesh_assert(!default_values || default_values->size() == names.size());
 
   std::vector<unsigned int> returnval(names.size());
 


jwpeterson:
I just realized that there was a TIMPI submodule update in with this PR. Was that intentional?

jwpeterson:
TIMPI submodule restored in 079271645

jwpeterson:
Also, this commit (minus the submodule change) was cherry-picked onto branch_v1.6.0 in 049aa0361.

roystgnr:
> Was that intentional?

To some extent every one of my mistakes is at least indirectly intentional, based on my repeated conscious decision to still write software rather than retiring in ignominy after events as shameful as this, but there was no *direct* intent behind this one.

roystgnr:
Thanks for catching+fixing that.

roystgnr:
On one of my systems I tried configuring git submodule.recurse=true, to try and prevent myself from forgetting the submodule update when I did a pull ... but then that turned out to be a PITA, because it would also try to recurse when doing a *push*, and I'd have to manually disable that every time to avoid whining about missing branches in the submodules.  So I didn't set submodule.recurse on any other systems, including the one from which I flubbed this PR, so here we are.  I don't suppose there's some even more esoteric way to get automatic recursive pulls *without* automatic recursive pushes?

jwpeterson:
@milljm would it be possible for CIVET to post a warning on any libmesh PR where there's a TIMPI submodule update? Similar to what's done for MOOSE PRs with a submodule update? That would probably help us prevent stuff like this from happening in the future as well.

jwpeterson:
> I don't suppose there's some even more esoteric way to get automatic recursive pulls without automatic recursive pushes?

In .git/hooks there's a pre-rebase.sample, but no post-rebase? My idea would be to have a post-rebase or post-pull hook that does "git submodule update" automatically, not sure if that's possible though.

milljm:
@jwpeterson Sorry, been crazy busy lately. Yeah I can look into a warning message.

milljm:
How would _any_ submodule change causing a warning sound to you?

If not, it shouldn't take me long to figure out how to allow a recipe to specify _what_ submodule should trigger a warning.

jwpeterson:
> How would _any_ submodule change causing a warning sound to you?

Sounds fine to me. They are still infrequent enough that it would be good to get warned about them when they happen.

milljm:
I believe the script used for MOOSE, is generic enough for any repo. If so, I feel bad I just didn't take the five minutes it took me to do this 5 days ago.

jwpeterson:
You are all good, this was a very low priority issue!

Issue #2710, opened 2020-09-21
error: static_assert failed due to requirement '__is_cpp17_move_insertable<std::__1::allocator<FunctionParserBase<double>>, false>::value' "The specified type does not meet the requirements of Cpp17MoveInsertable"


yurivict:
clang-11 fails to compile libmesh-1.4.1 on FreeBSD:

```
In file included from src/apps/calculator.C:24:
In file included from /usr/include/c++/v1/map:481:
In file included from /usr/include/c++/v1/__tree:15:
/usr/include/c++/v1/memory:1634:13: error: static_assert failed due to requirement '__is_cpp17_move_insertable<std::__1::allocator<FunctionParserBase<double>>, false>::value' "The specified type does not meet the requirements of Cpp17MoveInsertable"
            static_assert(__is_cpp17_move_insertable<allocator_type>::value,
            ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/v1/vector:954:21: note: in instantiation of function template specialization 'std::__1::allocator_traits<std::__1::allocator<FunctionParserBase<double>>>::__construct_backward_with_exception_guarantees<FunctionParserBase<double> *>' requested here
    __alloc_traits::__construct_backward_with_exception_guarantees(
                    ^
/usr/include/c++/v1/vector:1631:5: note: in instantiation of member function 'std::__1::vector<FunctionParserBase<double>, std::__1::allocator<FunctionParserBase<double>>>::__swap_out_circular_buffer' requested here
    __swap_out_circular_buffer(__v);
    ^
/usr/include/c++/v1/vector:1644:9: note: in instantiation of function template specialization 'std::__1::vector<FunctionParserBase<double>, std::__1::allocator<FunctionParserBase<double>>>::__push_back_slow_path<const FunctionParserBase<double> &>' requested here
        __push_back_slow_path(__x);
        ^
./include/libmesh/parsed_fem_function.h:632:15: note: in instantiation of member function 'std::__1::vector<FunctionParserBase<double>, std::__1::allocator<FunctionParserBase<double>>>::push_back' requested here
      parsers.push_back(fp);
              ^
./include/libmesh/parsed_fem_function.h:349:9: note: in instantiation of member function 'libMesh::ParsedFEMFunction<double>::partial_reparse' requested here
  this->partial_reparse(expression);
        ^
[...]

PR #2713, opened 2020-09-24
MeshFunction: use unique_ptr to manage PointLocator

jwpeterson:
The main reason behind this change is to make the class default
move-constructible, which should theoretically allow it to be stored
by value in a std::vector via push_back().

Refs #2710

### DIFF
--- a/src/mesh/mesh_function.C
+++ b/src/mesh/mesh_function.C
@@ -50,9 +50,7 @@ MeshFunction::MeshFunction (const EquationSystems & eqn_systems,
   _vector              (vec),
   _dof_map             (dof_map),
   _system_vars         (vars),
-  _point_locator       (nullptr),
-  _out_of_mesh_mode    (false),
-  _out_of_mesh_value   ()
+  _out_of_mesh_mode    (false)
 {
 }
 
@@ -69,26 +67,13 @@ MeshFunction::MeshFunction (const EquationSystems & eqn_systems,
   _vector              (vec),
   _dof_map             (dof_map),
   _system_vars         (1,var),
-  _point_locator       (nullptr),
-  _out_of_mesh_mode    (false),
-  _out_of_mesh_value   ()
+  _out_of_mesh_mode    (false)
 {
-  //   std::vector<unsigned int> buf (1);
-  //   buf[0] = var;
-  //   _system_vars (buf);
 }
 
 
 
-
-
-
-
-MeshFunction::~MeshFunction ()
-{
-  delete this->_point_locator;
-}
-
+MeshFunction::~MeshFunction () = default;
 
 
 
@@ -100,7 +85,7 @@ void MeshFunction::init (const Trees::BuildType /*point_locator_build_type*/)
   // Don't do twice...
   if (this->_initialized)
     {
-      libmesh_assert(this->_point_locator);
+      libmesh_assert(_point_locator);
       return;
     }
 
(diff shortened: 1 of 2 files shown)


src/mesh/mesh_function.C
```diff
@@ -50,9 +50,7 @@ MeshFunction::MeshFunction (const EquationSystems & eqn_systems,
   _vector              (vec),
   _dof_map             (dof_map),
   _system_vars         (vars),
-  _point_locator       (nullptr),
-  _out_of_mesh_mode    (false),
-  _out_of_mesh_value   ()
```
roystgnr, line 55: Wait a sec - doesn't this leave _out_of_mesh_value uninitialized?  I could easily imagine some user code relying on just getting 0 by default without manually saying so.

jwpeterson: out_of_mesh_value is a DenseVector so it will be default constructed no matter what, no reason to explicitly do so?

roystgnr: You're right, never mind.

dschwen:
This adds a _pessimizing move_ warning:

```
  CXX      src/mesh/libmesh_dbg_la-mesh_output.lo
../src/mesh/mesh_function.C:99:20: warning: moving a temporary object prevents copy elision [-Wpessimizing-move]
  _point_locator = std::move(mesh.sub_point_locator());
                   ^
```

just get rid of that `std::move`?

jwpeterson:
Hmm... yes I think this warning makes sense since there's no named variable to `std::move` from. I fixed it in #2723 but don't we have `-Werror` turned on in some of our builds? Or maybe this is a newer clang? Not clang-11 (#2710) by any chance?

dschwen:
Clang 9.0.1

dschwen:
Some more possibly related warnings:

```
In file included from ../src/reduced_basis/rb_eim_assembly.C:22:
./include/libmesh/rb_eim_construction.h:73:3: warning: explicitly defaulted move constructor is implicitly deleted [-Wdefaulted-function-deleted]
  RBEIMConstruction (RBEIMConstruction &&) = default;
  ^
./include/libmesh/rb_eim_construction.h:53:27: note: move constructor of 'RBEIMConstruction' is implicitly deleted because base class 'RBConstructionBase<libMesh::System>' has a deleted move constructor
class RBEIMConstruction : public RBConstructionBase<System>
                          ^
./include/libmesh/rb_construction_base.h:73:3: note: 'RBConstructionBase' has been explicitly marked deleted here
  RBConstructionBase (const RBConstructionBase &) = delete;
  ^
In file included from ../src/reduced_basis/rb_eim_assembly.C:22:
./include/libmesh/rb_eim_construction.h:76:23: warning: explicitly defaulted move assignment operator is implicitly deleted [-Wdefaulted-function-deleted]
  RBEIMConstruction & operator= (RBEIMConstruction &&) = default;
                      ^
./include/libmesh/rb_eim_construction.h:53:27: note: move assignment operator of 'RBEIMConstruction' is implicitly deleted because base class 'RBConstructionBase<libMesh::System>' has a deleted move assignment operator
class RBEIMConstruction : public RBConstructionBase<System>
                          ^
./include/libmesh/rb_construction_base.h:74:24: note: 'operator=' has been explicitly marked deleted here
  RBConstructionBase & operator= (const RBConstructionBase &) = delete;
```

jwpeterson:
OK, this one is caused by the changes in 68f985274. I suppose I could just undo that... I think at least `RBEIMConstruction` cannot be default move-assigned because one of its base classes, `System`, contains references.  It seems like default move construction should be possible though. I will have to look into this a bit more.

PR #2714, opened 2020-09-24
Add/update special functions in RB classes.

jwpeterson:
* RBConstructionBase
* RBParametrized
* RBEIMConstruction

In general we should not call clear() in any class destructor unless
we need to manually clean up memory managed by that class. Instead, we
should just let the member destructors do their thing.

Refs #2710


jwpeterson:
The error here doesn't seem to be real... only one test failed and it has nothing to do with my change.

PR #2715, opened 2020-09-25
Fix scatter for node constraints

roystgnr:
I'm not sure how this got left in #2558 but it was made redundant by the
upgrade there and it's a horrible deadlock to leave it in.


jwpeterson:
:+1: Seems like a good candidate for backporting as well.

roystgnr:
Definitely needs to be backported, sorry.  I could swear I tested that PR but maybe I screwed up a rebase?

PR #2716, opened 2020-09-28
Fix possible overflow in PeriodicBoundaries::neighbor

lindsayad:
When compiling with 64-bit indices, I get the compiler warning:
```
periodic_boundaries.C:116:30: warning: conversion from ‘libMesh::dof_id_type’
{aka ‘long unsigned int’} to ‘unsigned int’ changes value from
‘18446744073709551615’ to ‘4294967295’ [-Woverflow]
```
and rightfully so


jwpeterson (approved):

jwpeterson:
I will also cherry-pick this onto branch_v1.6.0 once it's merged.

roystgnr (approved):
Wait - I actually documented this correctly, with invalid_uint, but then I implemented it with invalid_id?  Good catch.

PR #2717, opened 2020-09-29
Dummy PR to test non-standard configs

jwpeterson:
Any fixes we make on here can be cherry-picked over onto branch_v1.6.0

### DIFF
--- a/include/numerics/distributed_vector.h
+++ b/include/numerics/distributed_vector.h
@@ -572,7 +572,11 @@ void DistributedVector<T>::set (const numeric_index_type i, const T value)
   libmesh_assert_less (i-first_local_index(), local_size());
 
   _values[i - _first_local_index] = value;
+
+
+  this->_is_closed = false;
 }
+
 
 
 
@@ -587,6 +591,9 @@ void DistributedVector<T>::add (const numeric_index_type i, const T value)
   libmesh_assert_less (i-first_local_index(), local_size());
 
   _values[i - _first_local_index] += value;
+
+
+  this->_is_closed = false;
 }
 
 
--- a/src/reduced_basis/rb_construction.C
+++ b/src/reduced_basis/rb_construction.C
@@ -92,12 +92,7 @@ RBConstruction::RBConstruction (EquationSystems & es,
   assemble_before_solve = false;
 }
 
-
-RBConstruction::~RBConstruction ()
-{
-  this->clear();
-}
-
+RBConstruction::~RBConstruction () = default;
 
 void RBConstruction::clear()
 {


jwpeterson:
fat-cat-running-580x405

Behold the power of a fully-operational CIVET.

jwpeterson:
@lindsayad, there is a unit test failure in the `--disable-optional` configuration

2) test: DiagonalMatrixTest::testNumerics (F) line: 120 ../../tests/numerics/diagonal_matrix_test.C
assertion failed
- Expression: !_matrix->closed()

The most likely issue is that there are no solver packages available in this config (if I recall correctly) so the fix is just to disable the test in that case. Would you mind taking a look since I think you were the one who added the test?

jwpeterson:
Some notes on the failing "all-static" building/linking test:

There is a linker failure caused by many undefined references, all of which seem to stem from the "libxml2.a" library, e.g.:
```
/usr/lib/x86_64-linux-gnu/libxml2.a(xzlib.o): In function `xz_head':
(.text+0x69e): undefined reference to `lzma_auto_decoder'
/usr/lib/x86_64-linux-gnu/libxml2.a(xzlib.o): In function `__libxml2_xzclose':
(.text+0x1f95): undefined reference to `lzma_end'
```
I'm not sure what would cause libxml to be linked in to libmesh, but upon inspecting one of the app executable's link lines, there is definitely a "-lxml2" there at the end:
```
ccache mpicxx -O2 -felide-constructors -funroll-loops
-fstrict-aliasing -Wdisabled-optimization -fopenmp -static -o amr-opt
src/apps/amr_opt-amr.o -Wl,-rpath
-Wl,/opt/civet/build_0/petsc/arch-civet-c-opt/lib -Wl,-rpath
-Wl,/opt/moose/mpich-3.3/gcc-9.2.0/lib ./.libs/libmesh_opt.a
-L/opt/civet/build_0/petsc/arch-civet-c-opt/lib
-L/opt/moose/gcc-9.2.0/lib64 -L/opt/moose/gcc-9.2.0/lib -L/usr/lib
/opt/civet/build_0/libmesh/build/contrib/netcdf/v4/liblib/.libs/libnetcdf.a
/opt/civet/build_0/libmesh/build/contrib/timpi/src/.libs/libtimpi_opt.a
/opt/moose/mpich-3.3/gcc-9.2.0/lib/libmpicxx.a
-L/opt/moose/mpich-3.3/gcc-9.2.0/lib
-L/opt/moose/gcc-9.2.0/lib/gcc/x86_64-pc-linux-gnu/9.2.0
-L/lib/x86_64-linux-gnu -L/usr/lib/x86_64-linux-gnu -lglpk -lz -lpetsc
-lHYPRE -lcmumps -ldmumps -lsmumps -lzmumps -lmumps_common -lpord
-lscalapack -lsuperlu_dist -lflapack -lfblas -lparmetis -lmetis -lX11
-lxcb -lXau -lXdmcp /opt/moose/mpich-3.3/gcc-9.2.0/lib/libmpifort.a
/opt/moose/gcc-9.2.0/lib/../lib64/libgfortran.a
/opt/moose/gcc-9.2.0/lib/../lib64/libquadmath.a
/opt/moose/gcc-9.2.0/lib/../lib64/libstdc++.a -ldl
/opt/moose/mpich-3.3/gcc-9.2.0/lib/libmpi.a -lm -lxml2 -lpthread -lrt -fopenmp
```
Its location within the link line makes it look like it's linked in by to Mpich, but since we do some link line re-ordering that might just be a coincidence. Could @milljm or someone else who uses the MOOSE compilers post the output of "mpicxx -show"? That would at least help narrow it down.

I'm not relying on this build configuration currently, so I probably won't be able to spend too much time tracking down the issue, mainly just wanted to document it while it was fresh on my mind.

lindsayad:
> Its location within the link line makes it look like it's linked in by to Mpich, but since we do some link line re-ordering that might just be a coincidence. Could @milljm or someone else who uses the MOOSE compilers post the output of "mpicxx -show"? That would at least help narrow it down.

I don't think this is coming from MPI:
```
(base) lindad@pop-os:/opt/moose/mpich-3.3/gcc-9.2.0/lib$ ldd libmpi.so
	linux-vdso.so.1 (0x00007ffe6a160000)
	libgfortran.so.5 => /opt/moose/gcc-9.2.0/lib64/libgfortran.so.5 (0x00007f613796a000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f61377f6000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f61377d3000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f61377c8000)
	libgcc_s.so.1 => /opt/moose/gcc-9.2.0/lib64/libgcc_s.so.1 (0x00007f61375b0000)
	libquadmath.so.0 => /opt/moose/gcc-9.2.0/lib64/libquadmath.so.0 (0x00007f6137369000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f6137175000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f6138381000)
(base) lindad@pop-os:/opt/moose/mpich-3.3/gcc-9.2.0/lib$ cat libmpi.la
# libmpi.la - a libtool library file
# Generated by libtool (GNU libtool) 2.4.6
#
# Please DO NOT delete this file!
# It is necessary for linking the library.

# The name that we can dlopen(3).
dlname='libmpi.so.12'

# Names of this library.
library_names='libmpi.so.12.1.6 libmpi.so.12 libmpi.so'

# The name of the static archive.
old_library='libmpi.a'

# Linker flags that cannot go in dependency_libs.
inherited_linker_flags=''

# Libraries that this one depends upon.
dependency_libs=' -lm -lpthread -lrt'

# Names of additional weak libraries provided by this library
weak_library_names=''

# Version information for libmpi.
current=13
age=1
revision=6

# Is this an already installed library?
installed=yes

# Should we warn about portability when linking against -modules?
shouldnotlink=no

# Files to dlopen/dlpreopen
dlopen=''
dlpreopen=''

# Directory that this library needs to be installed in:
libdir='/opt/moose/mpich-3.3/gcc-9.2.0/lib'
(base) lindad@pop-os:/opt/moose/mpich-3.3/gcc-9.2.0/lib$ mpicc -show
gcc -I/opt/moose/mpich-3.3/gcc-9.2.0/include -L/opt/moose/mpich-3.3/gcc-9.2.0/lib -Wl,-rpath -Wl,/opt/moose/mpich-3.3/gcc-9.2.0/lib -Wl,--enable-new-dtags -lmpi
```

jwpeterson:
> I don't think this is coming from MPI:

Yeah, me neither... that was kind of a wild guess. "-lxml2" also doesn't show up in the `libmesh_optional_LIBS` that gets printed by the configure summary, so no help there.

lindsayad:
> ```
> 
> ```
> 
> 
> The most likely issue is that there are no solver packages available in this config (if I recall correctly) so the fix is just to disable the test in that case. Would you mind taking a look since I think you were the one who added the test?

That would be one fix. Another fix option would be to make `add` and `set` operations in `DistributedVector` toggle `_is_closed` to `false` to be consistent with our other `NumericVector` derivatives.

lindsayad:
> Yeah, me neither... that was kind of a wild guess. "-lxml2" also doesn't show up in the `libmesh_optional_LIBS` that gets printed by the configure summary, so no help there.

It's a reasonable guess. We no longer use the MOOSE environment except on CIVET. My own local mpich build links to xml2

jwpeterson:
> Another fix option would be to make add and set operations in DistributedVector toggle _is_closed to false to be consistent with our other NumericVector derivatives.

Hmm, yeah that definitely seems reasonable. 

I guess our thinking was that `DistributedVector` was always essentially "closed" since it doesn't have any fancy caching or communications to do for add/set operations. But good to make it consistent with the other NumericVector types IMO. There's no efficiency issue I'm aware of since DistributedVector isn't actually used in real code...

jwpeterson:
> My own local mpich build links to xml2

OK, well maybe that is the culprit on CIVET then too? Either way I think it's more of an operating system/packaging issue. [This person](https://forum.crystal-lang.org/t/building-static-apps-using-libxml-with-official-docker-image/251) reported getting similar errors while linking against the static libxml2 libraries on Ubuntu as well, albeit while doing something completely unrelated. I think there's no static linking regressions from the standpoint of things we can control anyway.

lindsayad:
> OK, well maybe that is the culprit on CIVET then too? Either way I think it's more of an operating system/packaging issue. [This person](https://forum.crystal-lang.org/t/building-static-apps-using-libxml-with-official-docker-image/251) reported getting similar errors while linking against the static libxml2 libraries on Ubuntu as well, albeit while doing something completely unrelated. I think there's no static linking regressions from the standpoint of things we can control anyway.

I wouldn't rule it out. I'm trying to figure out how to get on the build boxes after all this DMZ business to verify.

lindsayad:
> I guess our thinking was that `DistributedVector` was always essentially "closed" since it doesn't have any fancy caching or communications to do for add/set operations. But good to make it consistent with the other NumericVector types IMO. There's no efficiency issue I'm aware of since DistributedVector isn't actually used in real code...

Hope you don't mind I pushed a patch to your branch for this... If you do, too late!

jwpeterson:
> Hope you don't mind I pushed a patch to your branch for this... If you do, too late!

I was planning to avoid that since there are like 20 checks on this PR. I was only going to invalidate the ones that failed (after the fixes were merged to master) to try and save some CI cycles.

lindsayad:
Ok noted. No more pushes to this branch. My b!

lindsayad:
Maybe I should read the PR title. Kind of late to avoid more CI cycles, but feel free to force push my changes away, in which case I'll open a new PR.

jwpeterson:
> feel free to force push my changes away, in which case I'll open a new PR.

Even though it says "dummy" I plan to actually merge this PR at some point, so no need to create a new one.

jwpeterson:
@lindsayad looks like your change worked BTW, there is now only one `--disable-optional` failure which @roystgnr is going to look into.
```
1) test: MeshInputTest::testDynaReadPatch (E) 
- ERROR:  Unrecognized solver package: 6
```

lindsayad:
Beautiful code always works

lindsayad:
Boom:
```
[Ubuntu-16][/opt/moose/mpich-3.3/gcc-9.2.0/lib]> cat libmpi.la
# libmpi.la - a libtool library file
# Generated by libtool (GNU libtool) 2.4.6
#
# Please DO NOT delete this file!
# It is necessary for linking the library.

# The name that we can dlopen(3).
dlname='libmpi.so.12'

# Names of this library.
library_names='libmpi.so.12.1.6 libmpi.so.12 libmpi.so'

# The name of the static archive.
old_library='libmpi.a'

# Linker flags that cannot go in dependency_libs.
inherited_linker_flags=''

# Libraries that this one depends upon.
dependency_libs=' -lm -lxml2 -lpthread -lrt'
```
Thanks @milljm for the civet image!

jwpeterson:
> dependency_libs=' -lm -lxml2 -lpthread -lrt'

OK, thanks for following up on that. 

I have no idea why either mpich would have an xml2 dependency in the first place, or why Ubuntu would have trouble linking against its own static xml2 libs, but that's where we're at on that. Usually people needing to do an all-static build have been on some weird system to begin with, so probably not a big deal that it doesn't work on Ubuntu, but good to have the issue documented.

lindsayad:
I wonder if a fix for this would be to use a conda environment instead of a MOOSE environment. The conda mpicc doesn't appear to link to xml

roystgnr:
> why Ubuntu would have trouble linking against its own static xml2 libs

Because they're not installed by default?  On 20.04 it looks like /usr/lib/x86_64-linux-gnu/libxml2.so.2 is in the libxml2 package, but /usr/lib/x86_64-linux-gnu/libxml2.a is only provided if you also install libxml2-dev.

jwpeterson:
Wouldn't that be a different error message, like "can't find libxml2.a"? We're seeing undefined references in an otherwise existing library...
```
/usr/lib/x86_64-linux-gnu/libxml2.a(xzlib.o): In function `xz_head':
(.text+0x69e): undefined reference to `lzma_auto_decoder'
```

roystgnr:
Damn, good point.  Let me see what that archive looks like on my desktop.

roystgnr:
lzma_auto_decoder is listed as an undefined symbol there, but it's *also* listed as an undefined symbol in the .so file...  Looks like that symbol is defined in liblzma.so and liblzma.a ... but that just pushes the question back a step - why isn't the linker complaining about not being able to find liblzma?  I guess there's no libxml2.la anywhere to let libtool know that its dependencies have dependencies, and libxml2.a doesn't give any dependency information to the linker the way libxml2.so does?

jwpeterson:
OK, yeah I did see some stuff online about needing to add "-llzma" but note that there are several undefined references to `ucnv_close_55`:
```
/usr/lib/x86_64-linux-gnu/libxml2.a(encoding.o): In function `xmlFindCharEncodingHandler':
(.text+0x2793): undefined reference to `ucnv_close_55'
```
which I suppose could also be due to the same issue, but the other lzma functions seem to be prefixed with `lzma_`.

roystgnr:
I'm seeing ucnv_ symbols defined in /lib/x86_64-linux-gnu/libicuuc.a (and .so)

lindsayad:
Mm conda only installs shared libraries so that's a dead-end. With that I do get missing file errors :smile:

jwpeterson:
I traced down the "Test No Exceptions" failure to the `buildCubePyramid{5,13,14}` unit tests. The building of the mesh works fine but then we call `prepare_for_use()` -> `delete_remote_elements()` -> `query_ghosting_functors()` -> `elem->find_point_neighbors() -> contains_point() -> point_test()` and within this test the `FEMap::inverse_map()` call throws a `libMesh::ConvergenceFailure()` exception. (This is likely due to the fact that we call find_point_neighbors() for the pyramid's apex, where the Jacobian is singular.) When exceptions are turned on, this works fine since that just means that the `contains_point()` test returns false, but when exceptions are disabled, the code calls `std::abort()` with no error message.

Since this isn't the first time this type of issue has come up, I think it would be good if we could find a better solution. I can think of the following:
1. Stop supporting the `--disable-exceptions` configuration. IIRC this was something that originated from a time where not all compilers had the best support for exceptions, but I think it's fair to say that this support is much better now.
1. Change `TypeTensor::solve()` so that it doesn't throw. Instead we could have it return a true/false `success` flag and then in e.g. `FEMap::inverse_map()` check this flag rather than doing the try/catch block that we have there now.

I guess I'm leaning towards Option 1 since using exceptions in this case feels extremely natural (i.e. it handles an error condition that occurs relatively infrequently) and I'd bet that most downstream apps explicitly depend on exception handling now. For example, it is definitely the only sensible way to recover from deeply-nested nonlinear solver failures. This doesn't have to be rushed into release 1.6.0... for the time being I'll open a PR to disable those Pyramid `build_cube()` tests when exceptions are not enabled.

roystgnr:
I'd vote for Option 2.

I'm not wedded to it, and since it's a pain to maintain I won't argue if you really want to mandate exceptions support, but I would like to explain why I wanted to make it a removable feature:

C++ exception support (along with RTTI, which IIRC we disable concurrently if configured that way) is one of the few things that violates C++'s "you only pay for what you use" principle - binaries compiled with exception support, even when they neither throw nor catch any exceptions, end up being typically at 5%-10% (even as much as 20% in a couple cases I'm skimming) larger.

So the [Google C++ Style Guide disallows them](https://google.github.io/styleguide/cppguide.html#Exceptions), and [the LLVM Coding Standards disallow them](http://llvm.org/docs/CodingStandards.html#do-not-use-rtti-or-exceptions), and IIRC some embedded C++ compilers still have no plans to support them.

And although I like exceptions, and I've been trying to get us to throw useful exceptions to user code where that makes sense, that's why I've been trying to avoid catch blocks where we rely on exceptions for correct behavior internally.  I never turn off exceptions in my own builds, so I'm not going to be too upset if we're too sick of maintaining that option, but I do think it's reasonable for people to want to disable them, I like that we're enabling them to do so, and I think a return value from TypeTensor::solve() would be worth it for that.

I'm also just now realizing the irony of the "ha ha now you have to maintain the move constructor" joke I made to Alex earlier, and I have to grudgingly admit that the obvious compromise here would be "we'll maintain --disable-exceptions support but from now on 'we' means 'just Roy'".

jwpeterson:
All the non-standard configs that could be fixed relatively easily have now been fixed, so I'm going to merge this PR now and hopefully tag a 1.6.0-rc2 soon.

It would be nice to get the "all-static" build working again, but I think I'd do something even more stripped down than what the current recipe does, e.g. build with no MPI and PETSc, just to cut down on the number of external packages that might not be working quite right with all-static linking.

PR #2720, opened 2020-09-29
ParsedFEMFunction: change "parsers" to a vector of unique_ptrs

jwpeterson:
This is an attempt to fix the issue raised in #2710. Since
FunctionParserBase has a deleted move constructor, it is not
considered to be MoveInsertable, which means it cannot be passed to
std::vector::push_back() by value.

### DIFF
--- a/include/numerics/parsed_fem_function.h
+++ b/include/numerics/parsed_fem_function.h
@@ -40,6 +40,7 @@
 #include <sstream>
 #include <string>
 #include <vector>
+#include <memory>
 
 namespace libMesh
 {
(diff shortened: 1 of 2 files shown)


jwpeterson:
The "Test No Optional" failure here is the same one reported in 
```
1) test: MeshInputTest::testDynaReadPatch (E) 
- ERROR:  Unrecognized solver package: 6
```
so I think this change is otherwise fine when Fparser is disabled.

roystgnr:
This looks good at first glance, but I think our only really decent test coverage here was in GRINS.  Give me a day or so to futz around there and try things out?

jwpeterson:
Sure, I'm not even trying to get this into 1.6.0 necessarily. If its proven that it works with clang-11, then I'll be backporting it to the 1.4 and 1.5 releases anyway so I can always tack on a 1.6.x at that time as well.

jwpeterson:
Note to self: invalidate "Test No Optional" build after #2721 is merged.

roystgnr:
Works fine with my tests, including with GRINS.

PR #2721, opened 2020-09-30
Nonstandard-configuration fixes

roystgnr:
The second commit here fixes the Dyna unit test failure in the configurations in #2717; the first commit fixes an overzealous assert that was affecting a swath of other tests in dbg/devel modes.

### DIFF
--- a/examples/fem_system/fem_system_ex3/fem_system_ex3.C
+++ b/examples/fem_system/fem_system_ex3/fem_system_ex3.C
@@ -40,6 +40,7 @@
 #include "libmesh/enum_solver_package.h"
 #include "libmesh/enum_solver_type.h"
 #include "libmesh/auto_ptr.h" // libmesh_make_unique
+#include "libmesh/numeric_vector.h"
 
 // The systems and solvers we may use
 #include "elasticity_system.h"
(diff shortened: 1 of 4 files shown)


jwpeterson:
Definitely some progress. It looks like when no optional packages are enabled we actually have a compile error in fem_system_ex3. I'll push a commit to this branch if it's possible to fix this with a simple inclusion...
```
../../../../examples/fem_system/fem_system_ex3/fem_system_ex3.C: In function 'int main(int, char**)':
../../../../examples/fem_system/fem_system_ex3/fem_system_ex3.C:270:29: error: invalid use of incomplete type 'class libMesh::NumericVector<double>'
       *(v_system->solution) = system.get_vector("_old_solution_rate");
                             ^
In file included from ../../../include/libmesh/dof_object.h:25:0,
                 from ../../../include/libmesh/mesh_base.h:24,
                 from ../../../include/libmesh/elem_range.h:24,
                 from ../../../include/libmesh/system.h:24,
                 from ../../../include/libmesh/equation_systems.h:26,
                 from ../../../../examples/fem_system/fem_system_ex3/fem_system_ex3.C:33:
../../../include/libmesh/int_range.h:34:29: note: declaration of 'class libMesh::NumericVector<double>'
 template <typename T> class NumericVector;
```

jwpeterson:
Hurray, "Test No Optional" passed finally. I think this can be merged any time now, but I'll wait for the rest of the testing to pass first. All of these commits will be cherry-picked onto branch_v1.6.0

jwpeterson (approved):

PR #2722, opened 2020-09-30
Always skip remote neighbors in DefaultCoupling

lindsayad:
Previously we only skipped over remote neighbors if they were a side neighbor. Now we skip whether they are a side neighbor or a periodic boundary neighbor.

### DIFF
--- a/examples/adaptivity/adaptivity_ex5/adaptivity_ex5.C
+++ b/examples/adaptivity/adaptivity_ex5/adaptivity_ex5.C
@@ -36,7 +36,7 @@
 
 // libMesh includes
 #include "libmesh/libmesh.h"
-#include "libmesh/replicated_mesh.h"
+#include "libmesh/mesh.h"
 #include "libmesh/mesh_refinement.h"
 #include "libmesh/gmv_io.h"
 #include "libmesh/exodusII_io.h"
@@ -225,9 +225,7 @@ int main (int argc, char ** argv)
   libmesh_example_requires(2 <= LIBMESH_DIM, "2D support");
 
   // Create a new mesh on the default MPI communicator.
-  // DistributedMesh currently has a bug which is triggered by this
-  // example.
-  ReplicatedMesh mesh(init.comm());
+  Mesh mesh(init.comm());
 
   // Create an equation systems object.
   EquationSystems equation_systems (mesh);
--- a/src/base/default_coupling.C
+++ b/src/base/default_coupling.C
@@ -132,19 +132,9 @@ void DefaultCoupling::operator()
             {
               const Elem * neigh = elem->neighbor_ptr(s);
 
-              // If we have a neighbor here
-              if (neigh)
-                {
-                  // Mesh ghosting might ask us about what we want to
-                  // distribute along with non-local elements, and those
-                  // non-local elements might have remote neighbors, and
-                  // if they do then we can't say anything about them.
-                  if (neigh == remote_elem)
-                    continue;
-                }
 #ifdef LIBMESH_ENABLE_PERIODIC
               // We might still have a periodic neighbor here
-              else if (check_periodic_bcs)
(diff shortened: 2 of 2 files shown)


examples/adaptivity/adaptivity_ex5/adaptivity_ex5.C
```diff
@@ -224,10 +224,10 @@ int main (int argc, char ** argv)
...
   libmesh_example_requires(2 <= LIBMESH_DIM, "2D support");
 
-  // Create a new mesh on the default MPI communicator.
-  // DistributedMesh currently has a bug which is triggered by this
-  // example.
-  ReplicatedMesh mesh(init.comm());
+  // Create a new mesh on the default MPI communicator. Directly test
+  // DistributedMesh, which used to have a bug when combining adaptivity with
+  // periodic boundary conditions
+  DistributedMesh mesh(init.comm());
```
roystgnr, line 230: This I don't want to merge without personal testing, and I'm not expecting those tests to pass.  The other commit is a good fix, but I'm afraid it's not the only fix required for AMR+PBCs.  Last I looked adaptivity_ex5 failed when run on 21 processors, with a *different* bug...

lindsayad: Should I strip off the latter commit?

roystgnr: Yeah, let's.  We're not going to merge it unless the first commit really happened to fix the bug I'm thinking of, and I don't even think we've got a CI option that goes up to 21 processors, so we might as well go with my gut recollection of the bug being a different one.

lindsayad: Let me know when you've completed your own testing. If this is good, I'd like to get it merged to move forward on idaholab/moose#15338

roystgnr: Actually, this might be good... looking through my notes, it was systems_of_equations_ex9 that I was able to get to die on 21 processors.  Whatever was long ago killing adaptivity_ex5 might actually be fixed at this point, in a previous PR if not in this one.  Let me run through things and I'll let you know.

If this is fixed, though, then I'd like to use "Mesh" here (which will test ReplicatedMesh frequently and DistributedMesh occasionally), not just "DistributedMesh" (which would test DistributedMesh frequently and ReplicatedMesh never).  I'd be dumbfounded if we managed to someday regress ReplicatedMesh because we'd stopped giving it as much testing.

lindsayad: I thought about `Mesh`, but was thinking about the fact that we not do any `--enable-parmesh` CIVET testing by default. Do we want to add that as a default target, or just throw it on occasionally?

lindsayad: There are four targets that `--enable-parmesh`. Even/odd distributed make check sweep opt/dbg

roystgnr (approved):
This works on up to 24 processors for me, at least.

examples/adaptivity/adaptivity_ex5/adaptivity_ex5.C
roystgnr: Yeah, I'm afraid that's deliberate.  The easy DistributedMesh failures I generally catch myself because I build with --enable-parmesh more often than not.  The *hard* DistributedMesh failures are the ones that only happen when you're at 15 processors and using the right adaptivity settings and it's the first blue moon after Labor Day and you sneeze right when Parmetis starts up.  So we usually don't bother with those tests until it's time for a release or a MOOSE upgrade (and IMHO not often enough in the latter case...) and we can afford a big sweep or two.

jwpeterson:
Is this something that should go in 1.6.0?

PR #2724, opened 2020-09-30
Generalize default constructors for TypeTensor and TypeVector

rwcarlsen:
Assigning to the (template typed) values in these data structures with
the (integer) zero doesn't work when the template type is not POD nor
has an overloaded assignment operator taking an int/number.  Assigning
"{}" instead works for a larger set of types, while still giving zero
values to POD.  This issue was run into when I tried to use an
eigen::Matrix<DualNumber<...>> in TypeVector.

### DIFF
--- a/include/numerics/type_tensor.h
+++ b/include/numerics/type_tensor.h
@@ -519,20 +519,20 @@ template <typename T>
 inline
 TypeTensor<T>::TypeTensor ()
 {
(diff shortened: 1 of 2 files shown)


jwpeterson:
xzibit

I heard you like linear algebra, so I put some AD inside a matrix inside your vector.

roystgnr (approved):

Issue #2703, opened 2020-09-14
Fix PBC issues for good (hopefully)


arovinelli:
The periodic boundary condition system in libmesh has some issues that are difficult to replicate: an example is 
#2590  - "I'm hurting now to look back at the optimism I had to just create a PR that I thought would fix everything so an issue wasn't even necessary..."

After many tries @roystgnr decided that patching code doesn't seem to be a good option (see #2700).

Issue #2590 is going to be the reference PR for this Issue.
@roystgnr feel free to edit this by writing down all the issues you think we might have and that this PR will resolve.
I will help to prepare test cases (unfortunately in MOOSE) and meshes to make sure we have good coverage

PR #2725, opened 2020-10-02
Ghost periodic point neighbors in GhostPointNeighbors

lindsayad:
This fixes my distributed periodic boundary issue. Currently our
periodic constraint algorithm relies on identifying a primary element in
charge of constraining (vertex) dofs. That designation is determined by
looking at point neighbors and comparing things like ids. If not all the
point neighbors are available, then a process may make an invalid
determination that it owns a primary elem when in fact the primary elem
is off-process. In order to to validly determine primary elems, we must
have all the periodic point neighbors available

### DIFF
--- a/src/base/ghost_point_neighbors.C
+++ b/src/base/ghost_point_neighbors.C
@@ -146,7 +146,7 @@ void GhostPointNeighbors::operator()
                     {
                       // Don't need to ghost RemoteElem or an element we already own or an
                       // element we've already examined
-                      if (appn == remote_elem || appn->processor_id() == _mesh->processor_id() ||
+                      if (appn == remote_elem || appn->processor_id() == p ||
                           periodic_elems_examined.count(appn))
                         continue;
 
--- a/src/base/ghost_point_neighbors.C
+++ b/src/base/ghost_point_neighbors.C
@@ -52,7 +52,8 @@ void GhostPointNeighbors::operator()
 
   std::set<const Elem *> periodic_elems_examined;
   const BoundaryInfo & binfo = _mesh->get_boundary_info();
-  std::vector<boundary_id_type> ppn_bcids;
+  std::vector<boundary_id_type> appn_bcids;
+  std::vector<const Elem *> active_periodic_neighbors;
 #endif
 
   // Using a connected_nodes set rather than point_neighbors() would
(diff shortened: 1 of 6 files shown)


src/base/ghost_point_neighbors.C
```diff
@@ -67,30 +89,97 @@ void GhostPointNeighbors::operator()
   CouplingMatrix * nullcm = nullptr;
 
   for (const auto & elem : as_range(range_begin, range_end))
+  {
```
roystgnr, line 92: libMesh indentation is two spaces to the brace, plus another two spaces to the code in the next block.

src/base/ghost_point_neighbors.C
```diff
@@ -67,30 +89,97 @@ void GhostPointNeighbors::operator()
...
+                periodic_elems_examined.count(ppn))
+              continue;
+
+            // We only need to keep point neighbors that are along the periodic boundaries
+            bool on_periodic_boundary = false;
+            for (const auto ppn_s : ppn->side_index_range())
+            {
+              binfo.boundary_ids(ppn, ppn_s, ppn_bcids);
+              for (const auto & pb_pair : *_periodic_bcs)
+                if (std::find(ppn_bcids.begin(), ppn_bcids.end(), pb_pair.first) != ppn_bcids.end())
```
roystgnr, line 148: Instead of a loop over _periodic_bcs, then a linear std::find for each, should we loop over ppn_bcids and use std::map::find for each?  I think both the constant and the asymptotics (which you wouldn't think are important for BCs but then you hear about the crazy stuff Akselos does...) ought to be better that way.

src/base/ghost_point_neighbors.C
```diff
@@ -67,30 +89,97 @@ void GhostPointNeighbors::operator()
...
+                {
+                  on_periodic_boundary = true;
+                  goto jump;
```
roystgnr, line 151: Of course, even switching loop ordering doesn't get rid of this goto, which I know is the right thing to do in C/C++ in this ONE USE CASE of breaking out of nested loops, but which gives me painful Apple BASIC flashbacks nonetheless.

roystgnr:
Good workaround.  I'm afraid it doesn't fix everything I know is wrong here (systems_of_equations_ex9 still gives me fits with DistributedMesh on 21 processors), but every little bit helps.

Can you get that failing case into CI in MOOSE, though, please?  #2703 is still mocking me, and I swear that at *some* point I'm going to fix distributed PBCs completely, but I fear that before that point happens I'm going to come up with attempted fixes that don't quite work (see #2590), and the more corner cases we have test coverage for, the less likely I am to let a non-working "fix" get merged.

src/base/ghost_point_neighbors.C
lindsayad: :+1:

lindsayad:
> Can you get that failing case into CI in MOOSE, though, please?

Oh definitely. It's actually technically already in! The issue is that after idaholab/moose#15338 we actually delete remote elements after `DistributedRectilinearMeshGeneration`, which we didn't do before, so previous regression tests that were relying on all the elements being there started diffing.

lindsayad:
Hopefully comments are addressed

lindsayad:
Woops. Hold on a second... Copying over the `_periodic_bcs` when copy-constructing `GhostPointNeighbors` is the same kind of error as copying over the `_mesh`...

lindsayad:
Ok now this code makes a MOOSE distributed displaced mesh with adaptivity and periodic boundary conditions work.

lindsayad:
Ready for re-review/merge

lindsayad:
Darn, my redistribute problem is because I’m comparing to the mesh processor id and not the processor id arg. Smh

lindsayad:
Ok after the bug-fix in https://github.com/libMesh/libmesh/pull/2725/commits/3f9015493934a777ff9228616e613144a664de7c this is truly ready for re-review 😆

src/base/ghost_point_neighbors.C
```diff
@@ -36,6 +42,23 @@ void GhostPointNeighbors::operator()
...
   libmesh_assert(_mesh);
 
+#ifdef LIBMESH_ENABLE_PERIODIC
+  bool check_periodic_bcs =
+    (_periodic_bcs && !_periodic_bcs->empty());
+
+  std::unique_ptr<PointLocatorBase> point_locator;
+  if (check_periodic_bcs)
+    {
+      libmesh_assert(_mesh);
```
roystgnr, line 52: Worried the mesh might have disappeared in the last 9 lines?

lindsayad: Uh yea! You can never be too sure. Haha I will fix that

roystgnr (approved):

PR #2727, opened 2020-10-04
Fix --disable-unique-id case in new Nemesis code

roystgnr:
This was another bug from #2681 that crops up in a configuration we
weren't regularly testing.


jwpeterson:
I inadvertently merged essentially the same fix in #2728, I'll merge this and clean up any issues that crop up during the merge...

PR #2728, opened 2020-10-05
Fix: testPeriodicLagrange2() unit test requires Exodus

jwpeterson:
Refs #2717


jwpeterson:
The first commit here should be cherry-picked onto 1.6.0-rc1, the second one was introduced more recently by #2681.

jwpeterson:
Hmm... the last "fix" works around an issue where Nemesis can apparently still be enabled even when the user configures with `--disable-exodus`. I don't think this should be allowed since the Nemesis code reuses parts of the Exodus code... Instead I think the configure tests need to be updated so that `--disable-exodus` implies `--disable-nemesis`.

PR #2729, opened 2020-10-05
Clang 10 warning fixes

roystgnr:
I wasn't able to reproduce the error in #2710 with clang++ 10 on Xubuntu 20.04, but turning on -Werror and our paranoid warnings list and trying out different configurations at least uncovered a half dozen warnings worth fixing.


jwpeterson (approved):
I will attempt to cherry-pick all of these on the 1.6.0 branch

PR #2734, opened 2020-10-06
Delete all mesh move constructors

lindsayad:
Clang warns of explicitly defaulted, implicitly deleted move
construction in `UnstructuredMesh`, which is definitely true. To remove
those warnings, we explicitly delete all move constructors from
`MeshBase` down

### DIFF
--- a/include/mesh/distributed_mesh.h
+++ b/include/mesh/distributed_mesh.h
@@ -73,9 +73,9 @@ public:
   DistributedMesh (const DistributedMesh & other_mesh);
 
   /**
-   * Move-constructor.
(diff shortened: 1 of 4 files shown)


lindsayad:
I've added the MOOSE Mac test, but that's unlikely to reproduce the failure seen on https://civet.inl.gov/job/588191/ since we don't define `-Werror` in the recipe.

jwpeterson:
Hmm... so this should probably have gone in with #2732. Since I cherry-picked that one to 1.6.0 I will need to cherry-pick this one as well.

roystgnr (approved):

PR #2742, opened 2020-10-09
Submodule updates

roystgnr:
This brings us up to date with both TIMPI and MetaPhysicL, including new methods in the former and the latest AD stuff in both.


roystgnr:
I swear I did a --enable-paranoid-warnings build with that last TIMPI PR ... but of course it didn't matter because that's one of the features we haven't moved into the autoconf submodule yet.  And for some reason we don't have -Wshadow in our travis.yml there either (did it not work with the ancient clang++ we're supporting?).

PR #2743, opened 2020-10-09
More IGA tests

roystgnr:
Along with fixes for a bug caught by the new tests.

Apologies to @jwpeterson for the additional work, but 9df4c12 fixes a pretty egregious oversight of mine and it should probably be cherry-picked back to the release.


jwpeterson (approved):
I will cherry-pick 9df4c12 now, thanks for the heads up.

PR #2744, opened 2020-10-12
Avoid manual memory management in System class

jwpeterson:
The `System::_vectors` data structure now more closely matches what we do with the `_matrices`. This makes it possible to default the move constructor for `System`. Also updated a few other things for C++11 that I came across.

### DIFF
--- a/include/systems/system.h
+++ b/include/systems/system.h
@@ -120,13 +120,12 @@ public:
    *   is the C++11 way of achieving the same effect.
    * - The System holds references to Mesh and EquationSystems
    *   objects, therefore it can't be default move-assigned.
-   * - This class manages memory in the _vectors member, so we can't
-   *   default the move constructor.
-   * - The destructor is responsible for cleaning up the _vectors member.
+   * - This class _can_ be default move constructed.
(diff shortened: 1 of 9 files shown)


include/systems/system.h
```diff
@@ -113,6 +113,22 @@ class System : public ReferenceCountedObject<System>,
...
 
+  /**
+   * Special functions.
+   * - The copy constructor and assignment operator were previously
+   *   declared as private libmesh_not_implemented() functions, this
+   *   is the C++11 way of achieving the same effect.
+   * - The System holds references to Mesh and EquationSystems
+   *   objects, therefore it can't be default move-assigned.
+   * - This class _can_ be default move constructed.
+   * - The destructor throws an error if libMesh::closed()
```
roystgnr, line 124: This should say something like "dies with an error in dbg/devel modes"?  We can't throw from a destructor and we don't bother to double-check in opt mode.

jwpeterson: Good point, I will change those docs.

jwpeterson: Fixed in fd6ac4977

PR #2748, opened 2020-10-15
TypeVector overloads for std::norm and friends for Eigen::Matrix

rwcarlsen:
This is needed in MOOSE.  I know this really seems like a moose-specific
hack finding its way into libmesh, but the way moose does unity builds,
we can't guarantee the order of appearance of the declarations of these
overloads.  The compiler keeps finding the wrong ones and ignores the
correct one if we put the overloads anywhere in MOOSE.  We can definitely
mod it a bit to try to generalize it a bit more, stick it in a separate include
file, etc. - but we're going to need something ~like this.

### DIFF
--- a/include/numerics/type_vector.h
+++ b/include/numerics/type_vector.h
@@ -30,6 +30,12 @@
 #include <cmath>
 #include <complex>
 
+#ifdef LIBMESH_HAVE_EIGEN
+#include "libmesh/ignore_warnings.h"
+#include "Eigen/Core"
+#include "libmesh/restore_warnings.h"
+#endif
+
 #ifdef LIBMESH_HAVE_METAPHYSICL
 #include "metaphysicl/dualnumber_forward.h"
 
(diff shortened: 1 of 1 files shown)


jwpeterson:
```
../contrib/eigen/eigen/Eigen/src/Core/Functors.h:977:28: error: 'template<class _Operation> class std::binder1st' is deprecated [-Werror=deprecated-declarations]
  977 | struct functor_traits<std::binder1st<T> >
```

Since we have -Werror turned on, this is an error. I guess it could be time to update the bundled version of Eigen again... but for now I think you'll have to wrap the Eigen header inclusion in
```
...
#include "libmesh/restore_warnings.h"
```
like is done in eigen_core_support.h

roystgnr:
This is ... less horrifying than I'd expected it to be, at least.  I've been trying to figure out what the "right" way would be to do nested specializations in unity builds, and no luck so far.

rwcarlsen:
Want me to it into a `horrifying_moose_specific_hacks.h` file?

roystgnr:
I'm actually a little confused as to how it doesn't *have* to be a horrifying_moose_specific_hacks.h file, that gets included at the top of every unity .C file.  If this *is* the problem I'd guessed it was, where you have to get all relevant overload/specialization declarations to precede all definitions in order to get the definitions to use the right specializations, then getting the overloads at the top of type_vector.h will work for those overloads calls to norm() etc. within TypeVector, but as soon as someone tries to call the same norm() from a file that doesn't include type_vector.h and happens to get placed ahead of type_vector.h in a unity build, you'll just have the same breakage again.

include/numerics/type_vector.h
```diff
@@ -47,6 +53,24 @@ template <typename T, typename D, bool asd>
 MetaPhysicL::DualNumber<T, D, asd> abs(const MetaPhysicL::DualNumber<T, D, asd> & in);
 template <typename T, typename D, bool asd>
 MetaPhysicL::DualNumber<T, D, asd> abs(MetaPhysicL::DualNumber<T, D, asd> && in);
+
+template <typename T, typename D, bool asd>
+using ADRealEigenVector = Eigen::Matrix<MetaPhysicL::DualNumber<T, D, asd>, Eigen::Dynamic, 1>;
```
roystgnr, line 58: Shouldn't this also go inside the LIBMESH_HAVE_EIGEN ifdef?

rwcarlsen: yes

rwcarlsen:
The only reason we need these overloads is because TypeVector looks for them.  I don't foresee ever needing these functions directly for any other reason.  If that changes, then what you suggest will likely be necessary.  I'm happy to put these in a separate file now.  I don't really care either way as long as my code in moose compiles :-)

Issue #2738, opened 2020-10-06
Don't throw from TypeTensor::solve()


jwpeterson:
The idea is to keep C++ exceptions support optional; currently they are essential to the way that `Elem::contains_point()` works so that's not ideal. 

See also the [discussion](https://github.com/libMesh/libmesh/pull/2717#issuecomment-703836224) on #2717.

PR #2749, opened 2020-10-15
Add method for identifying singular nodes and use in inverse_map

### LINKED ISSUE #2747: GhostPointNeighbors has unintended behavior
For this case, where the node in question is the sphere in the middle, the two elements pictured are on different processors, and there are other pyramids filling the gaps between these two pyramids:

Screen Shot 2020-10-12 at 4 20 03 PM

The recent changes introduced to `GhostPointNeighbors` do not properly ghost these elems.

loganharbour:
Closes #2747

### DIFF
--- a/include/geom/cell_pyramid.h
+++ b/include/geom/cell_pyramid.h
@@ -151,6 +151,19 @@ public:
 
   virtual std::vector<unsigned int> sides_on_edge(const unsigned int e) const override final;
 
+  /**
+  * \returns Node 4 (the apex) if \p is at the apex within the tolerance
+  * \p tol. If \p is not at the apex within the tolerance \p tol, will
+  * return invalid_uint.
+  *
+  * The Jacobian at the apex is singular because directional derivatives
+  * are defined in four directions, independently (four edges meet).
+  * Therefore, we can check for this in a caught, failed inverse mapping
+  * and return the correct point without failure if it is the apex.
+  * This behavior _does_ require exceptions to be enabled.
+  */
+  unsigned int local_singular_node(const Point & p, const Real tol = TOLERANCE*TOLERANCE) const override final;
+
 protected:
 
   /**
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -1553,6 +1553,21 @@ public:
   void libmesh_assert_valid_node_pointers() const;
 #endif // DEBUG
 
+  /**
+   * \returns The local node index of the given point IF said node
+   * has a singular Jacobian for this element. If the given point
+   * is not a node or is a node and does not have a singular Jacobian,
+   * this will return invalid_uint.
+   *
+   * The intention is for this to be overriden in derived element
+   * classes that do have nodes that have singular Jacobians. When
+   * mapping failures are caught, we can check this to see if the
+   * failed physical point is actually a singular point and
+   * return the correct master point.
+   */
+  virtual unsigned int local_singular_node(const Point & /* p */, const Real /* tol */ = TOLERANCE*TOLERANCE) const
+  { return invalid_uint; }
+
 protected:
 
   /**
--- a/src/fe/fe_map.C
+++ b/src/fe/fe_map.C
@@ -1878,6 +1878,16 @@ Point FEMap::inverse_map (const unsigned int dim,
               }
             libmesh_catch (ConvergenceFailure &)
               {
+                // If we encountered a singular Jacobian, but are at
+                // a singular node, return said singular node.
+                const unsigned int local_singular_node =
+                  elem->local_singular_node(physical_point, tolerance);
+                if (local_singular_node != invalid_uint)
+                {
+                  libmesh_assert_less(local_singular_node, elem->n_nodes());
+                  return elem->master_point(local_singular_node);
+                }
+
                 // We encountered a singular Jacobian.  The value of
                 // dp is zero, since it was never changed during the
                 // call to RealTensorValue::solve().  We don't want to
--- a/src/geom/cell_pyramid.C
+++ b/src/geom/cell_pyramid.C
@@ -220,4 +220,10 @@ std::vector<unsigned int> Pyramid::sides_on_edge(const unsigned int e) const
 
 
 
+unsigned int Pyramid::local_singular_node(const Point & p, const Real tol) const
+{
+  return this->node_ref(4).absolute_fuzzy_equals(p, tol) ? 4 : invalid_uint;
+}
+
+
 } // namespace libMesh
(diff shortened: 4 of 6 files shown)


include/geom/cell_pyramid.h
```diff
@@ -151,6 +151,8 @@ class Pyramid : public Cell
 
   virtual std::vector<unsigned int> sides_on_edge(const unsigned int e) const override final;
 
+  bool contains_point (const Point & p, Real tol=TOLERANCE) const override;
```
roystgnr, line 154: No final here?

loganharbour: Hmm - I'm actually not accustomed to using final at all. Didn't even think about it. Will add.

roystgnr:
I see Github has no "fingers crossed" reaction emoji...

loganharbour:
Heh - too bad we're playing around with CIVET right now so testing isn't up anyway.

include/geom/cell_pyramid.h
roystgnr: Doesn't actually make a difference in this case, just a good habit to have.  In other cases it can make the difference between "this will be a virtual function call" versus "this is inlined".

loganharbour:
> We'll want to keep an eye on the `buildSphereHex27()` test in the "Test No Exceptions" configuration once CIVET is back up and running...

Yeah, I actually wasn't paying attention to the revert in its entirety and didn't mean to pull those in. I'll just rip out the revert and make a commit taking out just the Pyramid stuff.

jwpeterson:
Hmm... it's not 100% clear to me what happened here, but my guess is that some of the shape function tests for Pyramids (e.g.  `FETest_SECOND_LAGRANGE_PYRAMID14::testU`) have started failing now because they are `reinit()`ing on element vertices? Either that or a point is being inverse-mapped to a pyramid apex. I think the `contains_point()` change we made is a good one, so hopefully we can figure out how to tweak these other Pyramid unit tests so that they continue working as well.

loganharbour:
I'll look more into the singular issues and what's causing/calling them

loganharbour:
The issue is in the inverse mapping at the apex for a pyramid. 

I have a few thoughts on how to get around this, but it will definitely require further discussion. The first of which is to add something like `unsigned int Elem::at_indeterminate_point(const Point &)` what will default to `invalid_uint` except for pyramids which will return 4 when at the apex. This will give the inverse map the hint to go ahead and directly return the reference apex positioning.

I've talked to @roystgnr about this some and he's pretty busy so we haven't gotten too far. Any thoughts @jwpeterson ?

jwpeterson:
> Any thoughts @jwpeterson

Could we roll this `at_indeterminate_point()` (`at_singular_point()`?) idea into the `Elem::contains_point()` routine somehow? Maybe an extra `bool * at_singular_point = nullptr` default argument to `contains_point()` that could be set internally? That seems like a natural time when we discover this condition anyway, and also I think it would let us avoid introducing extra overhead for elements that don't/can't have any singular points in their Jacobian. I'm not exactly sure whether we can then translate this information into the `FEMap::inverse_map()` calls where it is needed, though.

roystgnr:
> Could we roll this at_indeterminate_point() (at_singular_point()?) idea into the Elem::contains_point() routine somehow?

The trouble is that even if we get contains_point() perfectly correct, it ends up not being enough, not even for our unit tests: once we try to evaluate a solution at that point, the FE reinit has to translate it back to the master element, and inverse_map fails there.

I feel like there's got to be a way to make inverse_map work, but I'm just not seeing it.  I have tried some change-of-variables tricks on the expressions for Pyramid derivative components, similar to what we did to make Elem3::volume() numerically stable, but this time nothing I tried seemed to work.

loganharbour:
Any further thoughts @jwpeterson given the discussion above?

I'm very hesitant to start working on this until there is a moderate agreement on a fix so that it's not wasted work. I suspect that my first attempt will not be easily approved and will likely be too hacky.

jwpeterson:
When we hit the singular point, I think we should be in the `libmesh_catch` statement in the following code in fe_map.C:

https://github.com/libMesh/libmesh/blob/785a0dbcb6fa51c0e2c601cebabf6192c4dc96ab/src/fe/fe_map.C#L1847-L1914

If so, then instead of returning a point at "infinity" (aka 1e6) as we do now, could we check to see if we are at a known singular point (using some new `Elem` virtual API) and if so, return the known reference coordinates of that singular point instead? For the pyramid apex that should be (0,0,1) for example.

This issue is likely going to involve some speculative investigation and may not end up leading to a satisfactory conclusion... so it's definitely understandable if you don't want to spend too much time on it. I guess it all depends on how badly the bug in #2747 is messing up the stuff you are actually trying to work on.

Incidentally, this issue is related to #2738 where it's argued that we shouldn't throw from `TypeTensor::solve()` in the first place, but I think the changes here can be made independently of whether or not that gets implemented.

loganharbour:
I'll attempt this @jwpeterson. 

It's important to note that the capability that I need is just for `Pyramid::contains_point()` to work properly, not the inverse mapping. In ray tracing, I back up the intersection checks with `contains_point()` assertions. I could just disable the unit tests for pyramids at the apex, but I'd prefer to get this all working so that it's not forgotten about.

loganharbour:
Well.... here goes something.

loganharbour:
Looks like this fixes the issue with the inverse map. There's still an issue with `FEMap::compute_single_point_map()` that needs to be resolved.

loganharbour:
Okay, I'm not sure if I have time to fix the other issues. 

What are the general thoughts on adding this in addition to in the geom unit tests a inverse_map test on each of the nodes?

loganharbour:
Alright... I've spent far too long on this and don't have any more time to spend on it.

loganharbour:
Ping @jwpeterson - think I got this working. Let me know if this is what you had in mind.

`Pyramid::contains_point()` will fail without exceptions still, because we do the singular check within the catch. With this, I added a unit test that checks for `contains_point()` at nodes. It is disabled if at a singular node, without exceptions.

loganharbour:
Looks like we're in the clear! This is ready for review.

jwpeterson (approved):
PR looks good to me. If I understand correctly, we just skip some `contains_point()` tests when exceptions are disabled. This seems like a good approach until someone has a chance to implement #2738. Thanks for working on this!

loganharbour:
> This seems like a good approach until someone has a chance to implement #2738.

Yep - @roystgnr had a similar thought.

> Thanks for working on this!

Sure thing! Glad to have it done.

roystgnr:
Thanks!  I think this is the best we can do without inventing a new pyramid macroelement, and I'm not even sure *that* could be made to work.

PR #2750, opened 2020-10-16
Rebase #2506 and some minor clean-up

jwpeterson:
Refs #2506 
@BalticPinguin

### DIFF
--- a/examples/miscellaneous/miscellaneous_ex1/miscellaneous_ex1.C
+++ b/examples/miscellaneous/miscellaneous_ex1/miscellaneous_ex1.C
@@ -94,9 +94,6 @@ int main (int argc, char ** argv)
   // Skip this 3D example if libMesh was compiled as 1D/2D-only.
   libmesh_example_requires(3 <= LIBMESH_DIM, "3D support");
 
-  // Tell the user what we are doing.
-  libMesh::out << "Running ex6 with dim = 3" << std::endl << std::endl;
-
   // Create a serialized mesh, distributed across the default MPI
   // communicator.
   Mesh mesh(init.comm());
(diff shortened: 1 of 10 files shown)


tests/fe/inf_fe_radial_test.C
```diff
@@ -410,15 +1000,15 @@ public:
...
+        {15,15,Point(-0.0423204,0.0113397,0.12514)}
       };
 
     tabulated_grads[std::make_pair(FOURTH, JACOBI_30_00)] =
       {
         {16,3,Point(-0.0476508,0.012768,0.771323)},
         {17,10,Point(0.333487,-0.333487,1.01421)},
         {18,14,Point(0,0,0)},
-        {19,27,Point(-0.00698362,0.00187126,0.000667664)},
+        {19,27,Point(-0.00698362,0.00187126,-0.00228396)},
```
jwpeterson, line 1011: The other diffs in this test look like whitespace changes or replacing small numbers with 0, but this line appears to be a real difference? I'm curious how it passed before but also passes now, with a different gold value?

include/fe/fe_base.h
```diff
@@ -250,6 +250,10 @@ class FEGenericBase : public FEAbstract
    * \returns The curl of the shape function at the quadrature
    * points.
    */
+
+#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
```
jwpeterson, line 254: Should we possibly move the definition of the `virtual_for_inffe` macro to fe_base.h (or some other header) so that it can be used here as well?

jwpeterson (commented):
It's difficult for me to comment intelligently on these changes since I am not very familiar with InfFEs... You (@BalticPinguin) are definitely the expert here so I'm inclined to go with your changes since they are passing the tests, etc. I'll wait for @roystgnr's comments.

I made one or two small comments, but I can't approve this PR on GitHub since I opened it...

include/fe/fe_base.h
BalticPinguin: I forgot this part; but of course.
In this case, if I want to add changes, I use `jwpeterson:pr_2506` as reference for ``pull``/``push``

jwpeterson: Yeah, you should be able to fetch my branch, make changes, and push... let me know if you run into any issues with that.

tests/fe/inf_fe_radial_test.C
BalticPinguin: this is wrong: I had changed these values but found an error and changed them back; but forgot this one.
Interesting it runs through: the difference between the correct (deleted) and the new value is `0.002951624` in z-coordinate, the `norm_sq()` of the difference thus `8.712084e-06` which is below the threshold of `1e-4`.

jwpeterson: Ah, yeah I forgot we used a relatively loose tolerance on that test.

examples/miscellaneous/miscellaneous_ex1/miscellaneous_ex1.C
```diff
@@ -358,15 +355,30 @@ void assemble_wave(EquationSystems & es,
...
+      // is used (in 3D this is \f$ r^{-2} \f$).
+      //
+      // To account for this extra weight, \p phi, \p dphi and \p weight are
+      // re-scaled as well:
+      //  * J      --> Jxd^2
+      //  * phi    --> phi/decay x r
+      //  * dphi   --> dphi/decay x r
+      //  * weight --> weight / r^2
+      // With this, the product of Jacobian, test and trial functions (and their derivatives)
+      // can be used as normal since the extra weights cancel.
```
roystgnr, line 369: Thank you for this!

include/fe/inf_fe.h
```diff
@@ -71,21 +71,27 @@ class InfFERadial
    * \returns The first (local) derivative of the
    * decay in radial direction of the infinite element.
    *
-   * This is only valid for 3D?? - RHS
+   * This is only valid for 3D!
```
roystgnr, line 74: So ... could we get any kind of protection on this (and later)?  Require an ```Elem*``` and assert (then eventually switch) based on elem->dim()?  Require base_elem to already be built and check that?

We laugh at the old joke about the dimwitted new programmer who slowly realizes, "I don't think the compiler even *reads* the comments!", but frankly this dimwitted library developer is slowly realizing that you can't count on the app developers reading them all either.

roystgnr (approved):
Sure wish we had more than one InfFE expert around, so this could get a *proper* review.  But at least there aren't any obvious problems, and it is nice to have more than *zero* InfFE experts around again.

include/fe/inf_fe.h
BalticPinguin: you are right; after setting `decay` to valid values in all dimensions, we should adjust the derivative...

jwpeterson:
Those last changes addressed all my concerns, thanks!

BalticPinguin:
Thank you very much for your efforts!

PR #2751, opened 2020-10-19
Add ExodusII_IO::write_nodeset_data() API

jwpeterson:
Also add a new unit test for this. We are now able to read and write variables on node, edge, and sidesets in the Exodus format.


roystgnr:
Works for me.  Though I fear I'll never catch up on porting features over to Nemesis.

PR #2753, opened 2020-10-21
Simplify DiagonalMatrix::zero_clone() slightly

jwpeterson:
@lindsayad what do you think about this? I think it's a bit clearer to reuse `init()` here, while otherwise being the same amount of work overall. Calling `init()` doesn't really work as nicely in the non-zero `clone()` case since it doesn't actually copy values... so I left that one alone.

### DIFF
--- a/include/numerics/diagonal_matrix.h
+++ b/include/numerics/diagonal_matrix.h
@@ -84,14 +84,18 @@ public:
 
   /**
    * Initialize with a NumericVector \p other, e.g. duplicate the storage
-   * allocation of \p other. This in general DOES NOT copy the vector entires
+   * allocation of \p other. This function DOES NOT copy the vector entries.
+   * If you set fast = false, the initialized entries are explicitly zeroed,
+   * otherwise their values are indeterminate.
    */
   virtual void init(const NumericVector<T> & other, const bool fast = false);
 
   /**
    * Initialize with DiagonalMatrix \p other, e.g. duplicate the storage
-   * allocation of the underlying NumericVector in \p other. This in general
-   * DOES NOT copy the vector entires
+   * allocation of the underlying NumericVector in \p other. This function
+   * DOES NOT copy the vector entries. If you set fast = false, the
+   * initialized entries are explicitly zeroed, otherwise their values
+   * are indeterminate.
    */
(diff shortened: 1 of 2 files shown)


lindsayad (changes requested):
I actually don't understand why I made these virtual:
```
  /**
   * Initialize with a NumericVector \p other, e.g. duplicate the storage
   * allocation of \p other. This in general DOES NOT copy the vector entires
   */
  virtual void init(const NumericVector<T> & other, const bool fast = false);

  /**
   * Initialize with DiagonalMatrix \p other, e.g. duplicate the storage
   * allocation of the underlying NumericVector in \p other. This in general
   * DOES NOT copy the vector entires
   */
  virtual void init(const DiagonalMatrix<T> & other, const bool fast = false);
```
My issue with this change is that none of the documentation of the sequence of `init` APIs that are going to be called indicate that zeroing is going to happen if you pass `fast = false`. If you edit the `init(const DiagonalMatrix` doxygen and also the `NumericVector::init(NumericVector` to say that zeroing will happen if `fast = false`, then I'm totally on-board with these changes

jwpeterson:
> I actually don't understand why I made these virtual:

Hmm... the `SparseMatrix` base class `init()` methods are virtual, but these aren't overrides of them. I'm not really sure what the C++ rules are on that.

> My issue with this change is that none of the documentation of the sequence of init APIs that are going to be called indicate that zeroing is going to happen if you pass fast = false.

That's true, I did have to dig down into petsc_vector.C to determine exactly what `fast` did... I'll update the docs.

lindsayad:
> Hmm... the `SparseMatrix` base class `init()` methods are virtual, but these aren't overrides of them. I'm not really sure what the C++ rules are on that.

These end up just being different overloads, but I'm surprised we don't see shadowing warnings when we compile. Usually I end up having to do a `using BaseClass::methodWithOtherOverloads` if I introduce additional overloads in the derived class. Perhaps it's because `SparseMatrix` is a dependent template so the compiler can't be sure we're shadowing methods in the general template? @roystgnr ?

jwpeterson:
Yeah, not "shadowing" but virtuals being "hidden" by overrides is usually warned about, so it could indeed be some template thing that prevents that warning from being emitted.

PR #2757, opened 2020-10-22
UnsteadySolver: old_local_nonlinear_solution should be a std::shared_ptr

jwpeterson:
This fixes an issue where two `unique_ptrs` "owned" the same pointer, which we previously worked around by `release()`ing one copy before it could be destroyed. This is a situation that is easier to handle with `std::shared_ptrs`.

### DIFF
--- a/include/solvers/unsteady_solver.h
+++ b/include/solvers/unsteady_solver.h
@@ -164,8 +164,10 @@ public:
 
   /**
    * Serial vector of _system.get_vector("_old_nonlinear_solution")
+   * This is a shared_ptr so that it can be shared between different
+   * derived class instances, as in e.g. AdaptiveTimeSolver.
    */
-  std::unique_ptr<NumericVector<Number>> old_local_nonlinear_solution;
+  std::shared_ptr<NumericVector<Number>> old_local_nonlinear_solution;
 
   /**
    * Computes the size of ||u^{n+1} - u^{n}|| in some norm.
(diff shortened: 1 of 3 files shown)


roystgnr:
This breaks one line in GRINS, but that's easily fixed.  I'd like @vikramvgarg to get a chance to try this if he wants before we merge; he's done more AdaptiveTimeSolver stuff than anyone lately.

vikramvgarg:
If it passes adjoints example 5 and 7, thats a good sign. Those are two examples where the adaptive time solver is used, and the old_local_nonlinear_solution is exercised.

jwpeterson:
> If it passes adjoints example 5 and 7, thats a good sign.

Yeah, this change shouldn't cause any tests to fail since we are just mimicking the same behavior we had with `unique_ptr` in a slightly cleaner way. The main issue would be if you had an application code that used the public `old_local_nonlinear_solution` member in a unique-ptr specific way.

vikramvgarg:
> > If it passes adjoints example 5 and 7, thats a good sign.
> 
> Yeah, this change shouldn't cause any tests to fail since we are just mimicking the same behavior we had with `unique_ptr` in a slightly cleaner way. The main issue would be if you had an application code that used the public `old_local_nonlinear_solution` member in a unique-ptr specific way.

I dont have any app code that uses old_local_nonlinear_solution. Shouldnt be a problem from my side.

roystgnr (approved):

PR #2758, opened 2020-10-22
Move LinearSolver to ImplicitSystem

jwpeterson:
~~This function has been deprecated since the 1.1.x series of releases (2016), so now's a good time to finally make it an error to call it.~~

In order to maintain the const-ness of `ImplicitSystem::get_linear_solver()`, I had to make the `linear_solver` member mutable. I think it's a reasonable compromise to maintain backwards compatibility while avoiding manual memory management.

### DIFF
--- a/include/systems/continuation_system.h
+++ b/include/systems/continuation_system.h
@@ -370,13 +370,6 @@ private:
   NumericVector<Number> * delta_u;
 
   /**
-   * We maintain our own linear solver interface, for solving
-   * custom systems of equations and/or things which do not require
-   * a full-blown NewtonSolver.
-   */
-  std::unique_ptr<LinearSolver<Number>> linear_solver;
-
-  /**
    * False until initialize_tangent() is called
    */
   bool tangent_initialized;
--- a/include/systems/diff_system.h
+++ b/include/systems/diff_system.h
@@ -118,12 +118,6 @@ public:
   get_linear_solve_parameters() const override;
 
   /**
-   * Releases a pointer to a linear solver acquired by
-   * \p this->get_linear_solver()
-   */
-  virtual void release_linear_solver(LinearSolver<Number> *) const override;
-
-  /**
    * Assembles a residual in \p rhs and/or a jacobian in \p matrix,
    * as requested.
    * Note that in some cases only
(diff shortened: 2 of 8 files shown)


roystgnr:
I fear the FIN-S adjoints capability actually still depends on this.  Looks like it's using TransientNonlinearImplicitSystem under the hood, which doesn't own a LinearSolver, which is probably exactly why I'd added this memory leak of a hack in the first place.

Would it make sense to simply move the unique_ptr<LinearSolver> from LinearImplicitSystem (and ContinuationSystem, it looks like) to ImplicitSystem, then when get_linear_solver() needs to build a linear solver on the fly we build it there, so it gets cleaned up properly if there's a destruction without a preceding release_linear_solver() call?  Subclasses that already have their own LinearSolver would still have their *_linear_solver overrides, so they'd have 8 bytes of nullptr waste, but we never build so many System objects that we'd care about that.

jwpeterson:
I thought you said FIN-S developers don't use an up-to-date libmesh anyway?

roystgnr:
Nah, that's true.  But that's a fork I was hoping to eventually fix, not encourage from our end.

jwpeterson:
OK, one other question: are they actually calling `get_linear_solver()` in practice? I don't think it makes sense to do so... `NonlinearImplicitSystem` has a 
```
std::unique_ptr<NonlinearSolver<Number>> nonlinear_solver;
```
which is presumably what they are using for solving. Seems like there wouldn't be a `LinearSolver` object involved.

roystgnr:
*Everything* doing sensitivity analysis (forward or adjoint) or adjoint error analysis ends up calling sensitivity_solve() and/or adjoint_solve(), which in turn use get_linear_solver() and release_linear_solver() to get a LinearSolver to use.

Sorry I didn't remember the root issue (well, the issue with anything wanting to use these operations on NonlinearImplicitSystem) way back in #1145 - I guess I hadn't been using FIN-S myself for a couple years then, and my other adjoint work was always with my own code using FEMSystem.

jwpeterson:
OK, that (sensitivity analysis) definitely seems like a valid use case and I think it would still make sense at the `ImplicitSystem` level so I'll go ahead and move it.

roystgnr:
Thanks!

PR #2759, opened 2020-10-22
Updates to InfFEBase::build_elem()

jwpeterson:
* Preserve const-ness of input in the output
* Don't hand back allocated dumb pointer to the user
* InfFE::base_elem should be treated as const
* Update docs including the need for a const_cast workaround in debug mode.

### DIFF
--- a/include/base/dof_object.h
+++ b/include/base/dof_object.h
@@ -784,7 +784,6 @@ unsigned int DofObject::n_dofs (const unsigned int s,
 inline
 dof_id_type DofObject::id () const
 {
-  libmesh_assert (this->valid_id());
   return _id;
 }
 
--- a/include/fe/inf_fe.h
+++ b/include/fe/inf_fe.h
@@ -160,11 +160,12 @@ private:
 public:
 
   /**
-   * Build the base element of an infinite element.  Be careful,
-   * this method allocates memory!  So be sure to delete the
-   * new element afterward.
+   * Build the base element of an infinite element. Here "base" refers
+   * to the non-infinite/traditional geometric element face which is
+   * associated with this infinite element. A const base Elem is built
+   * from a const input InfElem.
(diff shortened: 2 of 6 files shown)


roystgnr:
Solid improvement IMHO but let's wait to see what @BalticPinguin thinks before we merge.

jwpeterson:
OK, thanks for reviewing. One thing that I think could be improved is for `Elem::build_side_ptr()` to set a "not invalid" `id()` on the side proxy element that it builds. This would allow us to avoid the `const_cast` that is in this PR. But, after thinking about that idea for a bit, it's not clear exactly what that side id should be. We could set it to the "parent" element's id or 0, but it's possible this duplication of ids would then lead to other issues...

A possibly simpler fix is to make it so that `DofObject::id()` does not throw an exception in debug (or any) mode. This seems safer overall to me, because it avoids the issue where you are already throwing an exception and in the course of trying to print information about the error (as in `inverse_map()`) you go to print the `id()` of the element causing the error, and this *also* throws, leading to an immediate program termination.

roystgnr:
We use id() so often as a key in map (and worse, vector) lookups, and in the past I've really liked the ability to catch invalid_id errors there with an exception rather than a segfault...  but I guess we catch the map case with libmesh_map_find now, and any remaining instances of the vector case (damn ErrorVector) ought to be deprecated for the sake of efficiency in distributed mesh runs ... so yeah, I'd be happy to see that assertion removed.  Sorry I didn't do it *last* time this problem came up, now that I think about it.

jwpeterson:
> so yeah, I'd be happy to see that assertion removed.

OK, great. I'll just tack that change on to this PR and then I can merge it without any `const_cast`.

BalticPinguin (approved):
This looks all fine to me.

PR #2760, opened 2020-10-23
Ubuntu 20.10 (gcc 10.2) fixes

roystgnr:
These fix a couple errors that gcc 10.2.0-13ubuntu1 gives me when trying to build a --enable-paranoid-warnings --enable-werror configuration on a beta of Ubuntu 20.10.

### DIFF
--- a/src/numerics/type_vector.C
+++ b/src/numerics/type_vector.C
@@ -65,8 +65,6 @@ template <typename T>
 void TypeVector<T>::write_unformatted (std::ostream & os,
                                        const bool newline) const
 {
-  libmesh_assert (os);
-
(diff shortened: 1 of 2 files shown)


src/numerics/type_vector.C
```diff
@@ -65,8 +65,6 @@ template <typename T>
 void TypeVector<T>::write_unformatted (std::ostream & os,
                                        const bool newline) const
 {
-  libmesh_assert (os);
```
jwpeterson, line 68: Wow, you'd think at *some* point we'd stop finding little Easter eggs like this...

jwpeterson (approved):

PR #2761, opened 2020-10-23
Re-enable check_dirichlet_bcid_consistency()

roystgnr:
As well as expanding it so we don't get false positives in codes that are only pinning points and/or edges but not sides.

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -1340,6 +1340,12 @@ public:
   DirichletBoundaries *
   get_adjoint_dirichlet_boundaries(unsigned int q);
 
+  /**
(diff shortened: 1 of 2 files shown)


jwpeterson:
Looks good to me. Not sure why it says 

> "roystgnr added 3 commits on May 20, 2016"

Is this an ancient commit that you rebased or something?

roystgnr:
Ha!  Exactly.  I was working on something else, noticed how many zillion undeleted branches are in the roystgnr fork of libMesh, and decided to skim through and triage them, deleting any that had obviously already been merged (or indirectly merged via a different cleanup branch), and in the process I found this branch that had never been merged but looked like it would be trivial to bring up to date.  The TODO that had been sitting in a comment for four years was the icing on the cake there.

PR #2762, opened 2020-10-23
Ubuntu 20.10 clang 11 warning fixes

roystgnr:
This isn't as complete as #2760 was (we'll need a TIMPI ignore_warnings.h update too, and that's too much hassle to be worth it just so we can turn on --enable-werror --enable-paranoid-warnings on one specific compiler), but it's everything that I could get to trigger outside of a submodule.

### DIFF
--- a/include/utils/ignore_warnings.h
+++ b/include/utils/ignore_warnings.h
@@ -33,6 +33,7 @@
 #pragma clang diagnostic ignored "-Wunused-private-field"
 #pragma clang diagnostic ignored "-Wextra"
 #pragma clang diagnostic ignored "-Wredundant-decls"
+#pragma clang diagnostic ignored "-Wcast-align"
 #pragma clang diagnostic ignored "-Wcast-qual"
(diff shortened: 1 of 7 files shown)


include/utils/ignore_warnings.h
```diff
@@ -33,6 +33,7 @@
 #pragma clang diagnostic ignored "-Wunused-private-field"
 #pragma clang diagnostic ignored "-Wextra"
 #pragma clang diagnostic ignored "-Wredundant-decls"
+#pragma clang diagnostic ignored "-Wcast-align"
```
jwpeterson, line 36: Did you happen to check whether `-Wcast-align` is new and hence should only be used with newer versions of clang?

roystgnr: Yeah, it's as old as I could find docs for.

roystgnr: Not sure whether it's only triggering for me now because the new clang catches more instances or because of the new OpenMPI just started using it.

jwpeterson (approved):

PR #2767, opened 2020-10-26
Add API to tell mesh it is not prepared

lindsayad:
Refs idaholab/moose#15944 and idaholab/moose#15945

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -154,6 +154,12 @@ public:
   { return _is_prepared; }
 
   /**
+   * Tells this we have done some operation where we should no longer consider ourself prepared
+   */
+  void set_isnt_prepared()
+  { _is_prepared = false; }
+
+  /**
    * \returns \p true if all elements and nodes of the mesh
    * exist on the current processor, \p false otherwise
    */


roystgnr (approved):
Looks fine to me, as a start.  Right now it won't do anything without careful user management of prepare_for_use calls, but should we be looking at a long-term plan too?  Perhaps code that currently calls prepare_for_use() should instead simply be calling set_isnt_prepared(), and code that *requires* a prepared mesh should be calling prepare_for_use(), and prepare_for_use() should be a no-op in cases where is_prepared() is already true?  Then multiple mesh operations that invalidate mesh data could be done in succession, without redundant calculations in between, without the mesh operations themselves having to know how they're being composed.

But we couldn't move toward that in a hurry, because of all the old code that relies on prepare_for_use() acting on the mesh but doesn't set_isnt_prepared().

Maybe what we do in the long run is mix this with the more fine-grained prepare_for_use() subdivision we were talking about?  prepare_for_use() would remain the backwards-compatible really-prepare-everything method, but each of the substeps would work in the way I outlined...

lindsayad:
> But we couldn't move toward that in a hurry, because of all the old code that relies on prepare_for_use() acting on the mesh but doesn't set_isnt_prepared().

You're referring to checking `_is_prepared` and returning if `true`? Yea, totally agree that we can't do that yet.

> Maybe what we do in the long run is mix this with the more fine-grained prepare_for_use() subdivision we were talking about? prepare_for_use() would remain the backwards-compatible really-prepare-everything method, but each of the substeps would work in the way I outlined...

Yes I think this is definitely where we want to head. I'm wondering about adding APIs like `bool neighbors_found() const` for adding more of that fine-grained control you're talking about.

PR #2768, opened 2020-10-27
Exodus copy fixes

roystgnr:
Fixes and newly-enabled unit tests for DistributedMesh bugs in the new ExodusII_IO code in #2681, one of which had doubled up to become a ReplicatedMesh-with-subdomain-restricted-variable regression too.


jwpeterson (approved):
Thanks, this fixes the issue for me.

jwpeterson:
These Exodus changes were merged in bb3a0db84d9c01bb936fd4676923dbf8fe7c9883, which is not in 1.6.0, so nothing needs to be backported here.

PR #2769, opened 2020-10-28
Default/delete special functions for System classes.

jwpeterson:
* ExplicitSystem
* ImplicitSystem
* CondensedEigenSystem
* DifferentiableSystem
* LinearImplicitSystem
* OptimizationSystem
* FEMSystem
* FrequencySystem
* NewmarkSystem
* ContinuationSystem
* RBConstructionBase
* RBConstruction
* RBEIMConstruction
* RBSCMConstruction
* EigenSystem

The destructors for this class hierarchy can mostly just be default()ed and in general should not call clear().

This change defaults a few move constructors that were previously deleted to avoid new clang compiler warnings... @roystgnr if you still have a clang-11 build laying around, it would be great if you could check whether this branch has any warnings for you. I don't think we have any CI with newer clangs, but I could be wrong about that.

### DIFF
--- a/include/reduced_basis/rb_construction.h
+++ b/include/reduced_basis/rb_construction.h
@@ -64,12 +64,10 @@ public:
 
   /**
    * Special functions.
-   * - This class contains unique_ptrs so it can't be default copy assigned/constructed
-   * - This class inherits from System so it can't be default move assigned/constructed
-   *   (to be defaulted later, after we make the System tree move-compatible)
+   * - This class has the same restrictions/defaults as its base class.
    * - Destructor is defaulted out-of-line
    */
-  RBConstruction (RBConstruction &&) = delete;
+  RBConstruction (RBConstruction &&) = default;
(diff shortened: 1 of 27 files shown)


jwpeterson:
> I don't think we have any CI with newer clangs, but I could be wrong about that.

Just double-checked and it looks like the Test Mac target has clang-9.0.0. It is building opt-mode only and doesn't use `-Werror`, so it won't fail if there are warnings... I'll manually check and see if any get generated though.

jwpeterson:
> I'll manually check and see if any get generated though.

There was actually a warning on Mac Test, as I had forgotten to update `EigenSystem`'s special functions. That should be fixed now, though.

roystgnr:
clang++ 11 is coming up clean for me with --enable-paranoid-warnings --enable-werror

roystgnr (approved):

jwpeterson:
> clang++ 11 is coming up clean for me with

Great, thanks for checking.

PR #2772, opened 2020-10-29
Return hasher instead of hash in FParserAD JIT API

### LINKED ISSUE #2771: Add const reference getter for parsed function bytecode
I need this in MOOSE to build the hash of an fparser object for JIT.

dschwen:
Closes #2771

### DIFF
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -642,13 +642,13 @@ template<>
 bool FunctionParserADBase<long double>::JITCompile() { return JITCompileHelper("long double"); }
 
 template<typename Value_t>
-std::string FunctionParserADBase<Value_t>::JITCodeHash(const std::string & Value_t_name)
+FParserJIT::Hash FunctionParserADBase<Value_t>::JITCodeHash(const std::string & Value_t_name)
 {
   FParserJIT::Hash hasher;
   hasher.addData(std::string("version 2.0"));
   hasher.addData(this->mData->mByteCode);
   hasher.addData(Value_t_name);
-  return hasher.get();
+  return hasher;
 }
 
 template<typename Value_t>
@@ -667,7 +667,7 @@ bool FunctionParserADBase<Value_t>::JITCompileHelper(const std::string & Value_t
     return false;
 
   // compute SHA1 hash of the function
-  std::string hash = JITCodeHash(Value_t_name);
+  std::string hash = JITCodeHash(Value_t_name).get();
 #ifndef NDEBUG
   hash += "_dbg";
 #endif
--- a/contrib/fparser/fparser_ad.hh
+++ b/contrib/fparser/fparser_ad.hh
@@ -9,6 +9,11 @@
 template<typename Value_t>
 class ADImplementation;
 
+namespace FParserJIT
+{
+  class Hash;
+}
+
 template<typename Value_t>
 class FunctionParserADBase : public FunctionParserBase<Value_t>
 {
@@ -138,7 +143,7 @@ protected:
 
 #if LIBMESH_HAVE_FPARSER_JIT
   /// return a SHA1 hash for the current bytecode and value type name
-  std::string JITCodeHash(const std::string & value_type_name);
+  FParserJIT::Hash JITCodeHash(const std::string & value_type_name);
 
   /// write generated C++ code to stream
   bool JITCodeGen(std::ostream & ccout, const std::string & fname, const std::string & Value_t_name);


roystgnr:
I'd rather return just a hash, if that's all you need.  If we add a getter for the bytecode then any future change to the byte code storage type or interpretation or whatever is no longer an internal change, it's an API change.

But I'm okay with this as-is if you disagree.

dschwen:
Hm. Rather than the hash (to which I need to add data) I could return a
hasher object.

roystgnr:
A hasher object sounds verbose.  I'd prefer this over that.  But couldn't you just return a size_t and use something like boost::hash_combine to add more data?

dschwen:
I could theoretically drop the SHA1 hashing and move everything to C++11 std::hash, but then I'd have to use boost for hashing vectors and hash combination and I'd be down to 64bits from 160bits. Still a pretty big space, and I fully acknowledge that my fear of hash collisions is probably unfounded and a result of a basic human inability to properly grasp probabilities...

roystgnr (approved):
Looks great to me now; thanks!

dschwen:
Thanks, in the meantime I have started on removing the SHA1 lib that I added in 2014 before we were using C++11. I think it is time to reevaluate if std::hash can do that job. 

https://github.com/dschwen/libmesh/commit/0dd7be2da6493dd9240b3ac82a64281e24636ecc

I've pushed this to a separate branch, though.

dschwen:
Hm, on two machines (mac and Linux) my `./bootstrap` runs are failing with

```
configure.ac:69: installing 'build-aux/compile'
configure.ac:14: installing 'build-aux/missing'
Makefile.am:81: error: CODE_COVERAGE_ENABLED does not appear in AM_CONDITIONAL
src/Makefile.am:112: error: CODE_COVERAGE_ENABLED does not appear in AM_CONDITIONAL
src/Makefile.am: installing 'build-aux/depcomp'
test/Makefile.am:185: error: CODE_COVERAGE_ENABLED does not appear in AM_CONDITIONAL
autoreconf: automake failed with exit status: 1
```

what might I be missing here?

jwpeterson:
Good question... are you by any chance using a version of autotools different from the one in libmesh/contrib? `which autoconf` for instance should point to  `$LIBMESH_ROOT/contrib/autotools/bin/autoconf`.

dschwen:
Hm, I think libMesh it bringing its own autotools:

```
$ ./bootstrap
Using autoconf (autoreconf (GNU Autoconf) 2.69)
Using automake (automake (GNU automake) 1.16.1)
Inadequate autotools in your path, I'll build new ones...
  building autoconf-2.69 in /home/schwd/Programs/libmesh/contrib/autotools
  building automake-1.16.1 in /home/schwd/Programs/libmesh/contrib/autotools
install-info: warning: no info dir entry in `/home/schwd/Programs/libmesh/contrib/autotools/share/info/automake-history.info'
  building libtool-2.4.6 in /home/schwd/Programs/libmesh/contrib/autotools
ar: `u' modifier ignored since `D' is the default (see `U')
ar: `u' modifier ignored since `D' is the default (see `U')

 --> successfully installed autoreconf in /home/schwd/Programs/libmesh/contrib/autotools/bin - please update your PATH and try again!

Using autoconf (autoreconf (GNU Autoconf) 2.69)
Using libtool (libtool (GNU libtool) 2.4.6)
Using automake (automake (GNU automake) 1.16.1)
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I m4 -I m4/autoconf-submodule
configure.ac:138: warning: ACSM_COMPILER_CONTROL_ARGS is m4_require'd but not m4_defun'd
m4/compiler.m4:3: LIBMESH_SET_COMPILERS is expanded from...
configure.ac:138: the top level
...
```

dschwen:
oh, wait, is it maybe not using them?! I didn't notice the message about the path...

jwpeterson:
Yeah we don't try to pre-empt whatever is in your PATH, so you have to update it.

This probably indicates an issue with our m4 files though. Maybe something that used to set `CODE_COVERAGE_ENABLED` was lost at some point, but bootstrap never complained about it?

dschwen:
Nah, I did `export PATH=/home/schwd/Programs/libmesh/contrib/autotools/bin:$PATH` and ran `./bootstrap` again. Same error.

jwpeterson:
Looks like it was removed in af6b375c89476550b69c2dbdeae193ae53b0d195

jwpeterson:
OK, not removed, but *moved*. Did you happen to run `git submodule update --init --recursive`? Maybe you don't have the new m4/autoconf-submodule, though I'd expect there to be other issues if that was the case.

dschwen:
Ah! Yeah, that was it. Maybe add that to the bootstrap script?

jwpeterson:
> Maybe add that to the bootstrap script?

Yeah that might be a good idea, but only if we are actually in a git repo and not an unpacked tar archive or something...

jwpeterson (approved):

jwpeterson:
Should this be merged before or after #2778 or does it not really matter?

dschwen:
It more of an either or the way I squashed the other PR :-/

roystgnr:
I approve of paranoia, so I'm happy using this instead of #2778 unless John has strong feelings the other way.

roystgnr:
I do think 64 hash bits ought to be enough for anyone.  :-)

jwpeterson:
> I'm happy using this instead of #2778 unless John has strong feelings the other way.

I haven't really been following what's going on here, so please choose the one you like best.

dschwen:
https://en.wikipedia.org/wiki/Option_Paralysis

jwpeterson:
> ... really, that's what that Wikipedia page is about?!

This isn't the disambiguation page you're looking for...

dschwen:
Closing in favor of #2778

PR #2773, opened 2020-10-29
Use custom deleters to automatically control lifetime of PETSc objects

jwpeterson:
We should be able to use the second template argument of `std::unique_ptr<T, Deleter>` to automatically call PETSc `XXXDestroy()` functions for PETSc objects when they go out of scope. Initially this results in slightly more lines of code to define the `Deleter` structs, but I think with some clever macro usage that could be reduced quite a bit. The end result is nice: you don't have to manually clean up PETSc objects any more.

If this initial change passes all the tests, I will eventually work on converting other PETSc code to use it as well.

### DIFF
--- a/src/solvers/petsc_linear_solver.C
+++ b/src/solvers/petsc_linear_solver.C
@@ -450,53 +450,64 @@ PetscLinearSolver<T>::solve (SparseMatrix<T> &  matrix_in,
                                     &subprecond);
       LIBMESH_CHKERR(ierr);
 
-      /* Since removing columns of the matrix changes the equation
-         system, we will now change the right hand side to compensate
-         for this.  Note that this is not necessary if \p SUBSET_ZERO
-         has been selected.  */
+      // Since removing columns of the matrix changes the equation
+      // system, we will now change the right hand side to compensate
+      // for this.  Note that this is not necessary if \p SUBSET_ZERO
+      // has been selected.
       if (_subset_solve_mode!=SUBSET_ZERO)
         {
           this->create_complement_is(rhs_in);
           PetscInt is_complement_local_size =
-            cast_int<PetscInt>(rhs_in.local_size()-is_local_size);
+            cast_int<PetscInt>(rhs_in.local_size() - is_local_size);
 
-          Vec subvec1 = nullptr;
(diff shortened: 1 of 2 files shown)


dschwen:
This is cool. Could this be simplified with a template alias?

```
template <typename T>
using petsc_unique_ptr = std::unique_ptr<T, PETScDeleter<T>>;
```

and then have a templated `PETScDeleter<T>` struct with specializations for operator `()`

roystgnr:
Love the idea, once it's working.

src/solvers/petsc_linear_solver.C
```diff
@@ -158,17 +140,22 @@ void PetscLinearSolver<T>::init (const char * name)
...
       // Create the linear solver context
-      ierr = KSPCreate (this->comm().get(), &_ksp);
-      LIBMESH_CHKERR(ierr);
+      _ksp = [&]()
+        {
+          KSP tmp;
+          ierr = KSPCreate (this->comm().get(), &tmp);
+          LIBMESH_CHKERR(ierr);
+          return std::unique_ptr<KSP, KSPDeleter>(&tmp);
+        }();
```
jwpeterson, line 149: This just doesn't work at all. I kind of thought that `tmp` here is like a heap-allocated pointer for some reason... but that seems to not be the case. In the debugger, I see:
```
(gdb) p _ksp
$1 = std::unique_ptr<_p_KSP *> = {get() = 0x7ffdd3e93ec8}
```
but
```
(gdb) p *_ksp
$2 = (_p_KSP *) 0x0
```
I still think there is an improvement possible that would allow us to avoid making manual `XXXDestroy()` calls, but this ain't it.

jwpeterson:
Probably we don't need to involve unique_ptrs at all and it's better to just wrap a `KSP` in a `KSPClass` with a destructor that calls `KSPDestroy()`. C++03 stuff basically.

jwpeterson:
> This is cool. Could this be simplified with a template alias?

Yeah, this seems like it would help clean up the syntax a bit and make it harder to forget the custom deleter template parameter. If the initial changes (now over in #2774) pass, I'll add this template alias in a future PR.

PR #2775, opened 2020-10-29
RB: Fix an issue where nodes in multiple nodesets were assembled twice

jwpeterson:
Also update and expand on the options for writing debugging ouputs during truth solves.

### DIFF
--- a/src/reduced_basis/rb_construction.C
+++ b/src/reduced_basis/rb_construction.C
@@ -620,49 +620,51 @@ void RBConstruction::add_scaled_matrix_and_vector(Number scalar,
   const MeshBase & mesh = this->get_mesh();
 
   // First add any node-based terms (e.g. point loads)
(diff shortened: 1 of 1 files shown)


jwpeterson:
Huh, looks like Mac doesn't have the version of mkstemp that takes a suffix...
```
../src/reduced_basis/rb_construction.C:1252:20: error: use of undeclared identifier 'mkstemps'
          int fd = mkstemps(temp_file, 4);
                   ^
../src/reduced_basis/rb_construction.C:1269:20: error: use of undeclared identifier 'mkstemps'
          int fd = mkstemps(temp_file, 4);
                   ^
2 errors generated.
```
Let me see if I can find a workaround for that.

jwpeterson:
> Let me see if I can find a workaround for that.

Looks like Linux and BSD-based operating systems define mkstemps in different headers.

PR #2776, opened 2020-10-29
SparsityPattern refactoring

roystgnr:
This is a prerequisite for some more Isogeometric Analysis work, but that's conceptually separate (and not quite done) so it'll be in a later PR.  Hopefully the code quality improvements here (more modularization, less redundancy) stand on their own.  Also hopefully they work on their own, and didn't get shredded by the careful git add --patch fiddling that put them here.

### DIFF
--- a/include/base/sparsity_pattern.h
+++ b/include/base/sparsity_pattern.h
@@ -66,6 +66,27 @@ static void sort_row (const BidirectionalIterator begin,
                       BidirectionalIterator       middle,
                       const BidirectionalIterator end);
 
+
+
+  /**
+   * Abstract base class to be used to add user-defined implicit
+   * degree of freedom couplings.
+   */
+  class AugmentSparsityPattern
+  {
+  public:
+    virtual ~AugmentSparsityPattern () {}
+
+    /**
+     * User-defined function to augment the sparsity pattern.
+     */
+    virtual void augment_sparsity_pattern (SparsityPattern::Graph & sparsity,
+                                           std::vector<dof_id_type> & n_nz,
+                                           std::vector<dof_id_type> & n_oz) = 0;
+  };
+
+
+
 /**
  * This helper class can be called on multiple threads to compute
  * the sparsity pattern (or graph) of the sparse matrix resulting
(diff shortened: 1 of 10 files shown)


include/base/sparsity_pattern.h
```diff
@@ -80,8 +100,57 @@ static void sort_row (const BidirectionalIterator begin,
...
 class Build : public ParallelObject
 {
+public:
+  Build (const DofMap & dof_map_in,
+         const CouplingMatrix * dof_coupling_in,
+         const std::set<GhostingFunctor *> & coupling_functors_in,
+         const bool implicit_neighbor_dofs_in,
+         const bool need_full_sparsity_pattern_in);
+
+  Build (Build & other, Threads::split);
```
jwpeterson, line 110: Consider defaulting the move constructor/deleting the other special functions for `Build` class? Not really necessary, but I feel it's good to have simply as documentation.

jwpeterson (approved):
Looks like a good refactoring to me... I couldn't tell from your last commit, what actually fixed the segfault we were seeing in that RB example?

roystgnr:
I hadn't updated the implementation of DofMap::attach_matrix() to stay compatible with the SparseMatrix::_sp change.  The System-internal matrices that went through init_matrices() were still fine, so most examples worked, and external matrices where I was explicitly attaching a new SparsityPattern in my new code worked, but the previous method where you set up external matrices by attaching them to the DofMap, that one I broke.

include/base/sparsity_pattern.h
```diff
@@ -107,6 +107,20 @@ class Build : public ParallelObject
          const bool implicit_neighbor_dofs_in,
          const bool need_full_sparsity_pattern_in);
 
+  /**
+   * Special functions.
+   * - The Build object holds references to DofMap and coupling
+   *   functors, therefore it can't be move-assigned.
+   */
+  Build (const Build &) = default;
+  Build & operator= (const Build &) = default;
```
jwpeterson, line 116: You can't default copy-assign objects with reference members either, as that would be re-seating a reference. Copy-construct should be OK though. Some compilers correctly warn about this, but others don't.

PR #2778, opened 2020-10-30
Drop SHA1 and use C++11 std::hash

### LINKED ISSUE #2771: Add const reference getter for parsed function bytecode

dschwen:
This is an alternative approach to #2772 per @roystgnr 's suggestion.

Closes #2771


dschwen:
I'm starting to like this one better than #2772 :-/

roystgnr:
I can't say I mind the line count decrease and the added std:: compatibility, as long as you're comfortable with the 160-bit to 64-bit change.  Say the word and we'll merge this one.

dschwen:
Ok, go for it. The "birthday problem" for 2<sup>64</sup> hashes solved for 1 _million_ parsed functions in a single input directory still results in a hash collision in only one in ~37 million cases. For a more realistic 1000 functions in one input (as a generous upper limit) would get a collision in one in 36,930,418,565,985 cases. I'm comfortable with those odds. (all assuming a perfect hashing function of course...)

contrib/fparser/fparser_ad.cc
```diff
@@ -10,18 +10,31 @@ using namespace FUNCTIONPARSERTYPES;
...
+namespace std {
+  template <typename T>
+  struct hash<std::complex<T>> {
+    size_t operator()(const std::complex<T> & val) {
+      return hash<T>{}(val.real()) ^ hash<T>{}(val.imag());
+    }
+  };
+}
+
+#include <libmesh/hashing.h>
```
lindsayad, line 31: @roystgnr this is where we started making libmesh contrib rely on libmesh...

lindsayad:
Naughty @dschwen foregoing the rules in https://en.cppreference.com/w/cpp/language/extending_std

dschwen:
> Naughty @dschwen foregoing the rules in https://en.cppreference.com/w/cpp/language/extending_std

Wait, what? That's exactly one of the permitted exceptions.

dschwen:
> hash of complex? complex is already a std library type. It's not a program type

Ugh. Why did I add that? I guess the specialization did not exist back then. We should check if it is part of c++17

roystgnr:
You probably added that and I probably ignored the sin because the C++ standard just sucks this badly here.  Not only is `std::hash<std::complex<T,U>>` not in C++17, it doesn't even look to be slated for C++20.  There are new hash implementations for various new classes, so they haven't just forgotten `std::hash`, there seems to be a deliberate decision not to specialize it for "compound" types.  No pair, no STL containers, no tuple... and no complex.

I suspect `boost::hash` (which is very thorough and why exactly it hasn't just been moved to the standard yet is beyond me...) is too big to copy into our `boostcopy::`, and I'd still rather not make boost a mandatory dependency (except for quad-precision support where we don't have other good options).  Maybe the thing to do is create a libMesh::hash shim at this point, and add things like complex and pair and tuple to that?

dschwen:
Yeah, a shim sounds like a great idea.

lindsayad:
It makes sense to me that the library hasn't done generic hash<complex> since the library only wants people to instantiate complex with float, double, and long double. To the boost conversation, it's funny that this line is in the std page for hash:

> Note: additional specializations for std::pair and the standard container types, as well as utility functions to compose hashes are available in [boost::hash](https://www.boost.org/doc/libs/release/libs/container_hash/doc/html/hash.html#ref)

roystgnr:
I wouldn't call cppreference the *standard* page.  If I grep through "Working Draft, Standard for Programming
Language C++" version N4140 (you have to pay for the real C++ standards, but cheap jerks like me can get drafts that are just one diff newer or older for free ... and apparently I'm also a lazy jerk since I have C++14 lying around but not C++17 yet?) they don't mention boost.  Boost is the de facto location for stuff that'll go into the official standard library if it's simple well-liked enough, but de jure it wasn't even as official as the "TR1" stuff that became popular when C++11 was dragging its feet.

PR #2780, opened 2020-10-31
Continue petsc_unique_ptr refactor

jwpeterson:
This basically finishes the refactoring started in #2774 and #2777. ~~I would still like to try this with a slightly different (C++03) type of wrapping class instead.~~ This now uses a C++03 style wrapping class, which as @roystgnr pointed out, avoids one extra layer of indirection and heap allocation while still having all the same capabilities.


jwpeterson:
@roystgnr this PR is good to go as far as I'm concerned. I'll merge tomorrow afternoon some time unless I hear from you first.

include/numerics/dense_matrix.h
```diff
@@ -31,7 +31,11 @@
 # ifdef I
 #  define LIBMESH_SAW_I
 # endif
+
+#include "libmesh/ignore_warnings.h"
 # include <petscsys.h>
```
roystgnr, line 36: What's getting triggered by their header that wasn't before?

include/numerics/wrapped_petsc.h
```diff
@@ -0,0 +1,228 @@
...
+
+#include <petscksp.h>
+#include <petscvec.h>
+#include <petscis.h>
+#include <petscmat.h>
+#include <petscviewer.h>
+#include <petscpc.h>
+#include <petscsnes.h>
+#include <petscdm.h>
+#include <petscsf.h>
```
roystgnr, line 42: I'd like to get rid of this lump-of-dependencies, if possible (or at least replace it with forward declarations), so each WrappedPetsc<Foo> will still only need petscfoo.h.  Any way to move them (and the destroy overrides) to a wrapped_petsc.C?

include/numerics/dense_matrix.h
jwpeterson: The warning was:
```
    In file included from ./include/libmesh/petsc_solver_exception.h:29,
                     from ./include/libmesh/petsc_unique_ptr.h:26,
                     from ../src/partitioning/partitioner.C:29:
    /opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/include/petscsys.h: In function 'PetscErrorCode PetscMemcpy(void*, const void*, std::size_t)':
    /opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/include/petscsys.h:1650:32: error: cast from type 'const void*' to type 'char*' casts away qualifiers [-Werror=cast-qual]
     1650 |     memcpy((char*)(a),(char*)(b),n);
          |                                ^
    In file included from ./include/libmesh/petsc_solver_exception.h:29,
                     from ./include/libmesh/petsc_unique_ptr.h:26,
                     from ../src/partitioning/partitioner.C:29:
    /opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/include/petscsys.h: In function 'PetscErrorCode PetscSegBufferGetInts(PetscSegBuffer, PetscInt, PetscInt* __restrict__*)':
    /opt/moose/petsc-3.11.4/mpich-3.3_gcc-9.2.0-opt/include/petscsys.h:2389:166: error: cast from type 'PetscInt* __restrict__*' {aka 'int* __restrict__*'} to type 'void**' casts away qualifiers [-Werror=cast-qual]
     2389 | PETSC_STATIC_INLINE PetscErrorCode PetscSegBufferGetInts(PetscSegBuffer seg,PetscInt count,PetscInt *PETSC_RESTRICT *slot) {return PetscSegBufferGet(seg,(size_t)count,(void**)slot);}
          |                                                                                                                                                                      ^~~~
```
I actually only saw it on the new header file I added, not quite sure why it wouldn't be triggered from the other inclusions of the same header, but I felt it was better to be on the safe side. It also doesn't seem to be triggered in newer PETScs.

include/solvers/petsc_dm_wrapper.h
```diff
@@ -18,30 +18,32 @@
...
 #include "libmesh/petsc_macro.h"
 
 #ifdef LIBMESH_HAVE_PETSC
 #if !PETSC_VERSION_LESS_THAN(3,7,3)
 #if defined(LIBMESH_ENABLE_AMR) && defined(LIBMESH_HAVE_METAPHYSICL)
 
-#include <vector>
-#include <memory>
-#include <unordered_map>
-#include <map>
```
roystgnr, line 31: Let's leave these in?  We are using them directly, not just indirectly.

include/numerics/wrapped_petsc.h
jwpeterson: Yeah I wasn't happy with this either. And there was only an issue on the `make installcheck` stage of the build, when the new header is tested without any other surrounding headers. I guess that could actually be fixed by moving the destroy specializations to a .C file... I'll try it.

src/solvers/petsc_linear_solver.C
```diff
@@ -140,18 +148,17 @@ void PetscLinearSolver<T>::init (const char * name)
       PetscErrorCode ierr=0;
 
       // Can't use libmesh_make_unique with custom deleter
```
roystgnr, line 150: Does this comment still make any sense here?

jwpeterson: No, that should have been removed when I dropped `petsc_unique_ptr`.

roystgnr (approved):

include/numerics/wrapped_petsc.h
jwpeterson: It worked in my local testing, thanks for the suggestion.

PR #2781, opened 2020-11-02
Add sanity check to System::init_data() 

jwpeterson:
Avoid trying to `init()` `System` solution vectors which are larger than the underlying NumericVector representation's index type allows. An example of this is with PETSc, which uses the `PetscInt` type that can be either a signed 4 or signed 8 byte type depending on how PETSc is configured. The idea/hope is that this will be an easier exception to recover from than an error which has to propagate up from the underlying NumericVector representation's library.

### DIFF
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -1081,7 +1081,12 @@ std::size_t DofMap::distribute_dofs (MeshBase & mesh)
   // dependencies to the send_list too.
   // this->sort_send_list ();
 
-  // Return total number of DOFs across all procs.
+  // Return total number of DOFs across all procs. We compute and
+  // return this as a std::size_t so that we can detect situations in
+  // which the total number of DOFs across all procs would exceed the
+  // capability of the underlying NumericVector representation to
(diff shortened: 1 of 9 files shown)


src/base/dof_map.C
```diff
@@ -1080,6 +1080,9 @@ void DofMap::distribute_dofs (MeshBase & mesh)
   // EquationSystems call that for us, after we've added constraint
   // dependencies to the send_list too.
   // this->sort_send_list ();
+
+  // Return total number of DOFs across all procs.
+  return std::accumulate(dofs_on_proc.begin(), dofs_on_proc.end(), static_cast<std::size_t>(0));
```
roystgnr, line 1085: Might want a comment here about the distinction between the dof_id_type calculation we did earlier to get _end_df[n_proc-1] and the forced-to-be-size_t calculation here; neither this code nor its motivation would be obvious to me if we hadn't just been talking about it.

roystgnr (approved):

jwpeterson:
Thanks for review!

PR #2782, opened 2020-11-04
InfFE fixes

roystgnr:
Since I last rebased my own work on master, I've started seeing some warning spew in dbg mode and a segfault in a unit test (at least with --disable-deprecated).  This fixes both for me.


jwpeterson:
Nice! I have been seeing these warnings lately in dbg-mode unit tests as well, but didn't have a chance to look into it yet.

PR #2785, opened 2020-11-05
Add VecGhostLocalFormRAII helper class

jwpeterson:
This approach gives better exception safety and avoids some code
duplication because we can do the GHOSTED check within the helper
class itself rather than before every operation on the Vec.

Add LIBMESH_CHKERR2 macro. This is useful when you want to call the
LIBMESH_CHKERR macro but you aren't in a ParallelObject.

### DIFF
--- a/include/numerics/petsc_solver_exception.h
+++ b/include/numerics/petsc_solver_exception.h
@@ -76,11 +76,21 @@ public:
       throw PetscSolverException(ierr);         \
     } } while (0)
 
+// Two-argument CHKERR macro that takes both a comm and an error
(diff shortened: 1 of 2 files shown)


roystgnr (approved):

roystgnr:
Should we shave off a few more lines by reimplementing LIBMESH_CHKERR in terms of LIBMESH_CHKERR2 now?

jwpeterson:
> Should we shave off a few more lines by reimplementing LIBMESH_CHKERR in terms of LIBMESH_CHKERR2 now?

Ah, yeah that would probably be good... Let me just squash that change into this commit if possible.

PR #2787, opened 2020-11-05
Support IsoGeometric Analysis meshes in System::project_vector()

roystgnr:
This was way trickier than I was expecting.  We can't project constraining DoFs and then just do a forward application of constraint equations to get the constrained DoFs consistent; we have to project constrained DoFs and then do a constrained solve to get the constraining DoFs consistent.

### DIFF
--- a/include/base/sparsity_pattern.h
+++ b/include/base/sparsity_pattern.h
@@ -221,7 +221,9 @@ private:
                              std::vector<dof_id_type> & dofs_vi,
                              unsigned int vi);
 
-public:
+#ifndef LIBMESH_ENABLE_DEPRECATED
+private:
+#endif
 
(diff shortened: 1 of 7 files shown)


roystgnr:
Since 75d0b81ee is a simple fix for a pre-existing bug, I've cherry-picking it onto a new branch_v1.6.1.

Since the bug is really hard to trigger (except with IGA meshes, which are still a somewhat bleeding-edge feature) and the error is within interpolation error levels even when it is triggered, a 1.6.1 tag and release will wait until/unless someone else gets bit by it or more urgent fixes worth backporting come up.

jwpeterson (approved):
I heard you like solves, so we put a solve in your constraints.

PR #2788, opened 2020-11-07
Add variadic template VecScatterBeginEnd() helper function

jwpeterson:
New helper function is in petsc_solver_exceptions.h. I initially tried
to put this in petsc_macro.h, but I got the error:
```
./include/libmesh/petsc_macro.h: In function ‘void libMesh::VecScatterBeginEnd(const libMesh::Parallel::Communicator&, const Args& ...)’:
./include/libmesh/petsc_macro.h:223:3: error: there are no arguments to ‘LIBMESH_CHKERR2’ that depend on a template parameter, so a declaration of ‘LIBMESH_CHKERR2’ must be available [-fpermissive]
   LIBMESH_CHKERR2(comm, ierr);
   ^~~~~~~~~~~~~~~
```
It could instead be in a separate header file that includes petsc_solver_exception.h, if desired.

### DIFF
--- a/include/numerics/petsc_solver_exception.h
+++ b/include/numerics/petsc_solver_exception.h
@@ -22,7 +22,7 @@
 
 #ifdef LIBMESH_HAVE_PETSC
 
-#include "libmesh_exceptions.h"
+#include "libmesh/libmesh_exceptions.h"
 
 #ifdef I
 # define LIBMESH_SAW_I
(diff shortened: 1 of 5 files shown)


include/numerics/petsc_solver_exception.h
```diff
@@ -22,7 +22,7 @@
 
 #ifdef LIBMESH_HAVE_PETSC
 
-#include "libmesh_exceptions.h"
```
roystgnr, line 25: Wait, really?  How was this compiling before?

jwpeterson: Yeah, I have no clue... it should definitely have failed the standalone header check for one thing.

roystgnr (approved):

PR #2790, opened 2020-11-10
Remove potential deadlock with Singletons and RemoteElem

lindsayad:
We had a potential deadlock before as reported by helgrind. We would
acquire the `remote_elem_mtx` first during setup and then
`singleton_mtx`. However during destruction we acquire `singleton_mtx`
first and then `remote_elem_mtx`. This has the potential to deadlock. So
why not let the singleton code handle all the locking? Seems like
singleton code should do that kind of thing and we should leverage it in
`RemoteElem` if we are going to go through the process of inheriting
from `Singleton`. This reports helgrind clean now

### DIFF
--- a/src/base/libmesh_singleton.C
+++ b/src/base/libmesh_singleton.C
@@ -73,6 +73,8 @@ Singleton::Singleton ()
 
 Singleton::Setup::Setup ()
 {
+  SingletonMutex::scoped_lock lock(setup_mtx);
+
   get_setup_cache().push_back (this);
 }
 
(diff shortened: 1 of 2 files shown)


jwpeterson (approved):
Looks good to me. I am also impressed if this is the only thing helgrind complains about?

lindsayad:
Haha yea this is all I saw on one MOOSE input file.

roystgnr:
Thanks!

>  I am also impressed if this is the only thing helgrind complains about?

More like heckgrind, amiright?

PR #2792, opened 2020-11-10
enable AMR with infinite elements.

BalticPinguin:
I am not very certain with this PR, I must admit: The changes work with my code and produce reasonable results but I am not sure how to test these things systematically.

The basic principle that I used is that for infinite elements only the base side (i.e. that side which interfaces finite elements and always has index ``0``) is refined.
When estimating the error, so-far, I just ignore infinite elements (i.e. they are just refined for sanity-reasons).
It might be better to take the surface between finite and infinite elements into account, but this requires some deeper changes in the structure, so I view this as some improvement for future...

Any thoughts on this or do you spot some mistakes in here?


src/fe/fe_abstract.C
```diff
@@ -888,6 +892,14 @@ void FEAbstract::compute_node_constraints (NodeConstraints & constraints,
         elem->neighbor_ptr(s) != remote_elem)
       if (elem->neighbor_ptr(s)->level() < elem->level()) // constrain dofs shared between
         {                                                 // this element and ones coarser
+
+#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+           // for infinite elements, only the base is constrained.
```
roystgnr, line 897: Is this really correct?  I'd have been certain you need constraints on the sides with neighboring infinite elements too.  Although I admit I wouldn't have the slightest idea how to calculate them.

roystgnr:
Well, that segfault has to mean *some* mistake.  Can you replicate it?  I'm afraid I can't see anything at first skim that should admit the possibility of a segfault.

src/fe/fe_abstract.C
BalticPinguin: I am not absolutely certain; but since there is no splitting in radial direction, I don't see how the basis could get discontinuous as long as the basis is not.

BalticPinguin:
After fetching the latest changes from the master, I can reproduce the segfault. I'll come back when I fixed it.

src/error_estimation/jump_error_estimator.C
```diff
@@ -307,6 +318,11 @@ void JumpErrorEstimator::estimate_error (const System & system,
               fine_context->side_fe_reinit();
             }
 
+#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+          if (e->neighbor_ptr(n_e) ->infinite()) // view infinite elements as boundary; don't integrate boundary_sides.
+             continue;
+#endif // LIBMESH_ENABLE_INFINITE_ELEMENTS
+
```
BalticPinguin, line 325: * this is the source of segfault: the `neighbor_ptr` may be `null`.
I just wonder why it didn't fail before...

jwpeterson: I think dereferencing a nullptr only technically results in undefined behavior, so it's not *guaranteed* to crash, although that is usually the result. Hard to say if that was at play for you here, though.

src/fe/fe_abstract.C
roystgnr: Are there more DoFs in the radial direction, though?  That's all you'd need for inconsistency in 3D.  If you have a coarse hex with two fine hexes next to it, all three of which have infinite neighbors, then just because the DoFs on the hanging node at the finite->infinite boundary are properly constrained doesn't mean the DoFs on the extended node (the one "closer to infinity" in cell_inf_hex8.h parlance, whose edge in that diagram leads to the hanging node) will be properly constrained too.

BalticPinguin: I see. You are right; this must be done 'somehow'...

BalticPinguin:
Hi @roystgnr,
I have rewritten most of the code and tried to incorporate your comments as good as I could.

The 'compute_constraints'-part should be okay now (I hope), the `compute_node_constraints` is not implemented. In my viewpoint, this has the problem that If we have multiple DOFs that contribute at a certain node, I would need to constrain them separately (which is done in the current form of the `compute_inf_constraint`) but I don't know how to do that in the `compute_node_constraints`-function...

Therefore, currently the code only runs when either infinite elements or node constraints are disabled.

src/error_estimation/jump_error_estimator.C
```diff
@@ -183,6 +183,12 @@ void JumpErrorEstimator::estimate_error (const System & system,
       const dof_id_type e_id = e->id();
 
 #ifdef LIBMESH_ENABLE_AMR
+
+#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+      if (e->infinite())
+         continue;
```
roystgnr, line 189: Should we have a libmesh_warning() here?  Although it's probably good enough in practice, I can't image that this is the *ideal* way to get an error indicator on a mesh with infinite elements, just taking into account jumps between finite elements.

src/fe/fe_base.C
```diff
@@ -1521,6 +1521,17 @@ FEGenericBase<OutputType>::compute_proj_constraints (DofConstraints & constraint
   if (!elem->active())
     return;
 
+#ifdef LIBMESH_ENABLE_INFINITE_ELEMENTS
+    // for infinite elements, only the base is constrained.
```
roystgnr, line 1525: Is this comment an atavism?  We discussed that in 3D there would be constraints on the off-base DoFs too, right?

But then the code below seems over-zealous, since in *2D* it's true that only the finite element neighboring an infinite element needs to be constrained; the infinite element layer is conforming.

src/error_estimation/jump_error_estimator.C
BalticPinguin: yes, probably a warning would be a nice option; on the other hand, if we account for the jump here, it can be caused by insufficient radial base of the infinite element which would, in turn, lead to repeated refinement without getting improvement there.

src/fe/fe_base.C
BalticPinguin: oh yes, I forgot that one..

BalticPinguin:
Hi @roystgnr,
I have rewritten most of the code and tried to incorporate your comments as good as I could.

The 'compute_constraints'-part should be okay now (I hope), the `compute_node_constraints` is not implemented. In my viewpoint, this has the general problem that there are multiple DOFs that contribute at each node. Thus one needs to constrain them separately (which is done in the current form of the `inf_compute_constraints`) but I don't know how to do that in the `compute_node_constraints`-function...

Therefore, currently the code only runs when either infinite elements or node constraints are disabled.

roystgnr:
> The changes work with my code and produce reasonable results but I am not sure how to test these things systematically.

Systematically is asking too much, I fear.  There are enough "probably works" heuristics in AMR that you can't put in outright "definitely works" assertions without risking some user hitting a weird counterexample.  I think there's three things you'll find worth doing, though:

1. Make sure you quantify "reasonable results".  Something like a DoFs-vs-error graph (whether error measured versus a known/manufactured solution, or an extremely overrefined solution, or via UniformRefinementErrorEstimator, in descending order of desirability), just to verify that you're not just looking good, you're getting faster convergence in the preasymptotic regime (or asymptotically, too, if your problem has a singularity or layer that restricts it to a low enough Hilbert space).

2. Add code exercising the AMR to some libMesh example.  I hate to say it, but there are too many people working on too many parts of the code, including deep parts that we have to rearchitect in sometimes-tricky ways to add new features or to optimize for performance, and so for anything that doesn't have enough test coverage in CI, the question isn't if it will end up being broken, the question is when.  Just having an AMR InfFE example running and giving coverage of our own internal assertions would be better than having no protection against regressions.

3. Assert the AMR solution values in the example.  Even with no systematic a priori estimate on your solution error on an adapted grid, you can still measure some quantities of interest that are relatively stable under slight mesh changes (Hilbert norms, point values at specified points, whatever), and then assert in the example that future runs evaluate those quantities to within some tolerance of your measured "gold" run values.  That's often the best you can do for continuous integration on complicated problems.

You don't need to do all this *now* if you don't want; I don't want to look a gift horse in the mouth and IMHO other than the lack of test coverage this is ready to merge.  But if you want to depend on this code in the future you'll be better off adding something sooner or later to make sure that it stays dependable.

roystgnr (approved):

BalticPinguin:
Thank you very much for the review and the comment!
Than I'd suggest to keep this PR open for a while, I hope that this will increase pressure on me to add examples/test coverage and adjust the [config-scheme](https://github.com/libMesh/libmesh/issues/2798#issuecomment-772614616), if you agree that it is sensefull.

roystgnr:
That works for me.  I'll leave it marked as "approved", since you shouldn't need to rebase/force-push/etc the existing code, but if I've convinced you that test coverage will be worth it then we can wait on that before merging.

BalticPinguin:
So finally I managed to add another example where AMR of infinite elements is used.
I've used that chance to enable `0`-frequency after I once set this as invalid point.
However, allowing for vanishing frequency has the advantage that one can use infinite elements to solve Laplace equation in unbound domains which is done now...

BalticPinguin:
since I cannot see the cause of the failed test: Is it caused by me or did something just go wrong?

jwpeterson:
> since I cannot see the cause of the failed test: Is it caused by me or did something just go wrong?

It's a failure in a MOOSE test, `kernels/hfem.dirichlet` so I don't think it was caused by your changes. I'll invalidate it and run again...

roystgnr:
hfem.robin_displaced is clearly only failing because it depends on MUMPS and Test Mac isn't supplying that.  So I'd say CI is happy with this.

examples/miscellaneous/miscellaneous_ex15/miscellaneous_ex15.C
```diff
@@ -0,0 +1,301 @@
...
+  for (int i=1; i<argc; i++)
+    std::cout << " " << argv[i];
+  std::cout << std::endl << std::endl;
+
+#ifndef LIBMESH_ENABLE_INFINITE_ELEMENTS
+  libmesh_example_requires(false, "--enable-ifem");
+#else
+
+#ifdef LIBMESH_ENABLE_SECOND_DERIVATIVES
+  // UniformRefinement uses second derivatives if they are enabled,
```
roystgnr, line 178: Oh, well that's a bug on my part.  Looks like I'm requesting the kitchen sink, then deciding after the fact which parts of it actually need to be used?  It'll be easy to fix that.  I'll put the fix in a separate PR after this is merged.

examples/miscellaneous/miscellaneous_ex15/miscellaneous_ex15.C
```diff
@@ -0,0 +1,301 @@
...
+#include "libmesh/numeric_vector.h"
+#include "libmesh/dof_map.h"
+#include "libmesh/point_locator_tree.h"
+// for infinite elements:
+#include "libmesh/inf_fe.h"
+#include "libmesh/inf_elem_builder.h"
+//#include "libmesh/fe_interface.h"
+// for refinement:
+#include "libmesh/error_vector.h"
+#include "libmesh/uniform_refinement_estimator.h"
```
roystgnr, line 52: You never ended up actually using URefEE, did you?  You're just doing uniform refinement in the loop below?  So we're not actually getting coverage of hanging nodes on uniform elements?

BalticPinguin: yes. Thank you for pointing that out.
I wanted to make it as simple as possible; but of course I shouldn't minimize code coverage.

BalticPinguin:
Sorry that I took that long for this...

But besides the small change in example misc 15, I found some errors in the computation of constraints and, probably most important, finally have a test-coverage for this, which makes me more confident that this actually works now.

BalticPinguin:
@roystgnr  I don't want to push you; but I don't know if you receive notifications if I make updates here; so I mention you that you see that here are some updates...

roystgnr:
I appreciate it!  I got the notification, but it's been a mess of a week and it got buried in my inbox.  I'll take a look again now.

examples/miscellaneous/miscellaneous_ex15/miscellaneous_ex15.C
```diff
@@ -271,8 +270,16 @@ int main (int argc, char** argv)
   // NOW THE REFINEMENT-LOOP comes; we use uniform_refinement.
```
roystgnr, line 270: This is an atavism now, yes?

tests/fe/inf_fe_radial_test.C
```diff
@@ -22,6 +29,7 @@
 #include "libmesh/fe.h"
 
 #include "libmesh/face_tri6.h"
+//#include "libmesh/cell_tet4.h"
```
roystgnr, line 32: Looks like this line got committed by accident?

src/fe/inf_fe_static.C
```diff
@@ -1267,19 +1267,28 @@ void InfFE<Dim, T_radial, T_map>::inf_compute_constraints (DofConstraints & cons
     return;
 
   if (child_elem->neighbor_ptr(0) == nullptr ||
-        child_elem->neighbor_ptr(0) != remote_elem)
-     return;
```
roystgnr: Yikes, I can't believe I missed this bug.

I'm unsure why you're having to look for remote_elem at all, though.  Are we not looping over local elements only when we get here?  And if so, shouldn't we never see a remote_elem neighbor?

src/fe/inf_fe_static.C
```diff
@@ -1374,10 +1383,25 @@ void InfFE<Dim, T_radial, T_map>::inf_compute_constraints (DofConstraints & cons
       if (self_constraint)
         continue;
 
+      // if the base is not coarser, than only those DOFS shared with respective neighbors need constraints:
+      if (coarser_neighbors[0] != 0)
+        {
+          bool needs_account=false;
+          for (auto & X : coarser_neighbors)
+            if ( child_elem->neighbor_ptr(X)->contains_point(child_side->point(child_side_dof)) )
```
roystgnr, line 1391: I'm having trouble following what the coarser_neighbors code is supposed to do (why does any elem-neighbor pair care about the other neighbors' status?), but this part in particular is greatly confusing me.  You're passing child_side_dof, a dof index, into a function expecting a point index?  What happens if you have more DoFs than points?

src/fe/inf_fe_static.C
BalticPinguin: I must admit that I got paranoid on `remote_elem`s recently, so I started to add checks for it everywhere. But I agree that there should be none...

src/fe/inf_fe_static.C
BalticPinguin: I agree that this is not written well; I'll try to rework it to be more readable. 
> why does any elem-neighbor pair care about the other neighbors' status?

I think, there are two issues: First, I tried some shortcut in the logic of the function; but while trying to write this down, I see that there is too much confinement. So thanks for pointing me there!

> You're passing child_side_dof, a dof index, into a function expecting a point index?
Probably I missed something here: I adapted this from `fe_lagrange.C:803`:
````
              // The support point of the DOF
              const Point & support_point = my_side->point(my_dof);
````
which, to my mind, is the very same: `my_dof` is the dof index of `my_side` which is used to obtain its corresponding support point.
Is there another, safer, way to obtain the point that I want to get?

Comment (roystgnr) [reply src/fe/inf_fe_static.C]: I usually see the "one dof index" == "one point" confusion in people used to working with Lagrange elements, where that actually holds.  But libMesh has a ton of other elements that don't behave that way (e.g. a cubic Hermite element has 2^d dofs on each vertex, a Hierarchic element of degree p has p-1 dofs on each mid-edge node and p-1^2 on each mid-quad-face node) ... and I *thought* our infinite elements were among those!  Don't you end up with more and more dofs on a mid-infinite-edge node as you raise the polynomial degree of the radial basis functions use?

If you're trying to loop over nodes on a side, in general you want to forget dofs exist.  Either loop over all nodes and check `Elem::is_node_on_side` to decide which to skip, or get `Elem::nodes_on_side` and loop over those.

**Comment by BalticPinguin** [comment]
May I bring this back on your list, @roystgnr ?
I feel a bit bad for the pushing, since I took some time as well...

I hope, with the recent changes the code is more understandable. At least, I think that it should be more correct now (before it was consistent, but confined more DOFs than necessary).

**Comment by BalticPinguin** [comment]
ping @roystgnr

File: src/fe/inf_fe_static.C (hunk 10 above)
Comment (BalticPinguin) [reply src/fe/inf_fe_static.C]: Hi @roystgnr , I saw this only by now, sorry.
It is correct that infinite elements are not of "one dof index" = "one point" because the radial polynomials degree can be set independent from the element type. But I wouldn't think of those degrees as belonging to the 'mid-infinite points'. In fact, these mid points are not needed for the basis functions at all (given that the 'origin' is known; which in practice is computed using the outer points). 
Since both the order and the type of polynomial can be set to 'something', there are only very few properties one knows in general. But what they have in common is that the first one is a/r (where a is the distance of the 'inner' face from the origin) and the other polynomials start with 0 at the inner face. 
Thus, if I restrict the DOFs at nodes, I get those with `radial_dof=1` correct and restrict the other DOFs at the outer points, not knowing if the radial polynomials probably have a node there.

Therefore, I implemented the scheme slightly different: Knowing that the radial DOF of the basis functions with `radial_dof==1` is 1 at the base, I can restrict it in the same way as Lagrange elements (This function is only accessed when the non-radial (base) dofs are of Lagrange type).
DOFs with `radial_dof > 1` must be restricted in the same way as those with `radial_dof = 1` to ensure smoothness along the radial axis, i.e. if the first dof is restricted to 1/3 of its max. value, I must restrict all other DOFs that share the same base DOF to 1/3 of their actual maximum.

I hope, I made my point somehow understandable?
Probably this should be explained also in the sources in more detail...

Comment (roystgnr) [reply src/fe/inf_fe_static.C]: > I saw this only by now, sorry.

No worries.  Between "trying to wrap urgent things up by vacation time" and "vacation time" and "trying to catch up on what I missed during vacation time", I'm more behind on getting back to you on this ticket than vice-versa.

> But I wouldn't think of those degrees as belonging to the 'mid-infinite points'.

Where are their indices stored, then?  On the base points?

> Probably this should be explained also in the sources in more detail...

Yeah, even in the current version of the code I'm seeing `child_side->point(child_side_dof)`, and I still don't get it.  We do store all indices on some DofObjects, right, even in infinite elements?  And the number of indices can exceed the number of points, right?  So how can we look up a point by passing in a dof index rather than a point/node index?  How does that not become an OOB index when `n_elem_dofs` exceeds `n_points`?

Comment (BalticPinguin) [reply src/fe/inf_fe_static.C]: I hope the answer in the [the newest commit](https://github.com/libMesh/libmesh/pull/2792/commits/5a66cd08f6dbe8b7b6cbb863ccb3c80f56c426d8) helps to understand my idea.

>Where are their indices stored, then? On the base points?

Good point: No, they are indeed stored on the outer points. But this doesn't mean that these nodes are 'support' points: The radial polynomial may have a node or some small value there; so we may constrain them in a wrong way (or worse don't mark them for constraining when the radial polynomial has a node).

>  even in the current version of the code I'm seeing `child_side->point(child_side_dof)`

I think, the confusion starts with the name `child_side`; I have changed it to `child_base` now: independent of the `side` considered, we use the base face to compute the constraint; and there, we have `Lagrange`-type elements.
I hope this becomes clean from my new part in the documentation: Since the radial direction is not refined, here the basis functions have the same behavior and hence we are (in principle) free to choose a point where the constraints should be set.
But, instead of searching for an explicit point, I assume a point where the radial polynomial takes the value of 1, allowing me to compute the constraint just based on the `child_base`s functions.

Comment (BalticPinguin) [reply src/fe/inf_fe_static.C]: @roystgnr , I hope, with the newest explanation the function is easier to understand!?

**Comment by BalticPinguin** [comment]
Maybe I may send a reminder about this to you, @roystgnr 
Probably my explanations are still not convincing. But I personally think that the logic is correct, so probably you can try to formulate another question? Or did I miss to answer your previous question?
Best regards

File: doc/html/src/examples.html [hunk 11]
```diff
@@ -187,6 +187,8 @@ <h1>A Series of Example Programs</h1>
 
 <li><L1><a href="examples/miscellaneous_ex14.html">Infinite Elements With Imaginary Frequency</a></L1></li>
 
+<li><L1><a href="examples/miscellaneous_ex10.html">Infinite Elements With Zero Frequency: Unbound Laplace Equation</a></L1></li>
```
roystgnr, line 190: Not sure how I missed this before - this should be ex15, no?

src/fe/inf_fe_static.C
roystgnr: Sorry for taking so long to get back to this; I was banging my head against the early drafts of #2985, your variable name change and comment here didn't make things any more understandable at first glance, and before I found time to do another full read through I just let it drop off my attention.

At second glance ... it actually does make sense, doesn't it?  The infinite element DoFs and nodes are always ordered base-first, the base FE type is always Lagrange, and the Lagrange FE DoF indices are equal to local node indices, so of course you can interchange them?

I'd love it if we can eventually weaken "the base FE type is always Lagrange", but for now this looks great.

roystgnr:
More discussion in the resolved comment chain above, but TL;DR: other than that HTML typo, this all looks good to me now.  Ready to merge once that's fixed?

doc/html/src/examples.html
BalticPinguin: thanks. Sure this is wrong.

BalticPinguin:
From my side, it is ready to merge than.
Please let me know if I should do some rebasing to have a cleaner history!?
Best regards
Hubert

src/fe/inf_fe_static.C
BalticPinguin: >  The infinite element DoFs and nodes are always ordered base-first, the base FE type is always Lagrange

yes, this exactly is the point. 

I'd love it if we can eventually weaken "the base FE type is always Lagrange", but for now this looks great.

Yes, that indeed would be nice; but my impression was that most other elements don't compute constraints or use `compute_proj_constraints` which than require some generalizations; but I am not aware of how we could enter this part with non-Lagrange elements...

roystgnr:
I guess a few of those commits should be turned into 'fixup' changes to the previous commits they fix/modify, sure.  I don't think there's any need for more reorganization than that, though.

src/fe/inf_fe_static.C
roystgnr: `compute_proj_constraints` might not be as efficient as per-FE code (which, even though it probably wants to do just those projections, might be able to use more direct calculations rather than instantiating generic quadrature rules and inverting small systems), but it doesn't require much from the FE type and it gives the result we want for the space types it handles.

Can we run with base type != LAGRANGE on conforming meshes already?

BalticPinguin:
This is the 'squashed' version; I hope, you are fine with it!?

src/fe/inf_fe_static.C
BalticPinguin: It seems that it works with Non-lagrange basis as well (I have tested Bernstein, xyz and monomial which seem to to be the only ones that work with Tet-elements) in the sense that I didn't receive errors.
I didn't check if the results make sense, though.

roystgnr:
Thanks again!  Sorry this turned out to be such a grueling process!

BalticPinguin:
Thanks for merging!
I am happy it finally got merged.
And I hope that finally the code quality is good enough that also following InFEM users will understand this :-)

PR #2800, opened 2020-11-12
Test whether we need to call build_constraint_matrix() recursively

jwpeterson:
The hypothesis is that libmesh now expands all "recursive" constraints before calling this function, and therefore it should not need to be called recursively. This PR tests whether that is the case.

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -1586,11 +1586,9 @@ private:
    * Build the constraint matrix C and the forcing vector H
    * associated with the element degree of freedom indices elem_dofs.
    * The optional parameter \p called_recursively should be left at
-   * the default value \p false.  This is used to handle the special
-   * case of an element's degrees of freedom being constrained in
-   * terms of other, local degrees of freedom.  The usual case is for
-   * an elements DOFs to be constrained by some other, external DOFs
-   * and/or Dirichlet conditions.
+   * the default value \p false. It used to be necessary to call this function
+   * recursively to handle "unexpanded" constraints, but we now
+   * expand all the constraints before calling this function.
    *
    * The forcing vector will depend on which solution's heterogenous
    * constraints are being applied.  For the default \p qoi_index this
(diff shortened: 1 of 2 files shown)


jwpeterson:
@roystgnr looks like the failures are related to IGA (assuming because of the "Dyna" stuff) and periodic boundary conditions. So I guess it's not the case that `build_constraint_matrix()` always receives expanded constraints and we do still need the recursion? If so, I will just change this PR to be a documentation update and leave the recursion in...
```
!!!FAILURES!!!
Test Results:
Run:  1916   Failures: 0   Errors: 5


1) test: MeshInputTest::testDynaFileMappingsFEMEx5 (E) 
- All constraints should have been expanded before calling DofMap::build_constraint_matrix(), but dof_set still contains 9 entries.


2) test: MeshInputTest::testDynaFileMappingsBlockWithHole (E) 
- All constraints should have been expanded before calling DofMap::build_constraint_matrix(), but dof_set still contains 27 entries.


3) test: MeshInputTest::testDynaFileMappingsPlateWithHole (E) 
- All constraints should have been expanded before calling DofMap::build_constraint_matrix(), but dof_set still contains 9 entries.


4) test: MeshInputTest::testDynaFileMappingsCyl3d (E) 
- All constraints should have been expanded before calling DofMap::build_constraint_matrix(), but dof_set still contains 27 entries.


5) test: PeriodicBCTest::testPeriodicLagrange2 (E) 
- All constraints should have been expanded before calling DofMap::build_constraint_matrix(), but dof_set still contains 2 entries.
```

roystgnr:
I can't quite understand the IGA failures there.  We are semi-manually adding constraints from DynaIO *after* the existing constraints have all been processed.  But the IGA constraints we add shouldn't actually need any further processing!  Values on FE DoFs get constrained to values on spline node DoFs, but there's only one level of constraint there, nothing recursive.

With the Periodic BC test I have the converse confusion: I can see how we can have recursive constraints there (periodic vs Dirichlet constraints at corner nodes), but both those types of constraints get automatically added before constraints are processed, so I don't see why they're not expanded afterward.

jwpeterson:

jwpeterson:
Closing this because I can't even remember the context and it hasn't been touched for 4+ years.

Issue #2802, opened 2020-11-12
Quad4::contains_point() should work with a non-planar face


loganharbour:
I'm sure there are other cases where this doesn't work properly, but Quad4 is the easy one to fix.

I'm proposing something like this:
```
bool
Quad4::contains_point (const Point & p, Real tol) const
{
  if (this->has_affine_map())
    return Quad::contains_point(p, tol);

  auto tri3_side = Elem::build(TRI3);
  tri3_side->set_node(0) = const_cast<Node *>(&this->node_ref(0));
  tri3_side->set_node(1) = const_cast<Node *>(&this->node_ref(1));
  tri3_side->set_node(2) = const_cast<Node *>(&this->node_ref(3));

  if (tri3_side->contains_point(p, tol))
    return true;

  tri3_side->set_node(0) = const_cast<Node *>(&this->node_ref(2));

  return tri3_side->contains_point(p, tol);
}
```

jwpeterson:
I'm confused. It looks like there is no `Quad::contains_point()` currently? And the stuff with the Tri elements I would assume works in the affine_map case, not the other way around...

loganharbour:
Closing for now per #2803

PR #2803, opened 2020-11-12
Specialize Quad4 contains_point and close_to_point for non-planar faces

loganharbour:
Refs #2802

### DIFF
--- a/src/geom/face_quad4.C
+++ b/src/geom/face_quad4.C
@@ -243,4 +243,44 @@ Quad4::loose_bounding_box () const
 }
 
 
+
+bool
+Quad4::contains_point (const Point & p, Real tol) const
+{
+  if (Quad4::has_affine_map())
+    return Quad::contains_point(p, tol);
+
+  auto tri3_side = Elem::build(TRI3);
+  tri3_side->set_node(0) = const_cast<Node *>(&this->node_ref(0));
+  tri3_side->set_node(1) = const_cast<Node *>(&this->node_ref(1));
+  tri3_side->set_node(2) = const_cast<Node *>(&this->node_ref(3));
+
+  if (tri3_side->contains_point(p, tol))
+    return true;
+
+  tri3_side->set_node(0) = const_cast<Node *>(&this->node_ref(2));
+
+  return tri3_side->contains_point(p, tol);
+}
+
+
+bool
+Quad4::close_to_point (const Point & p, Real tol) const
+{
+  if (Quad4::has_affine_map())
+    return Quad::close_to_point(p, tol);
+
+  auto tri3_side = Elem::build(TRI3);
+  tri3_side->set_node(0) = const_cast<Node *>(&this->node_ref(0));
+  tri3_side->set_node(1) = const_cast<Node *>(&this->node_ref(1));
+  tri3_side->set_node(2) = const_cast<Node *>(&this->node_ref(3));
+
+  if (tri3_side->close_to_point(p, tol))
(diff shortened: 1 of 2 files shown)


src/geom/face_quad4.C
```diff
@@ -243,4 +243,44 @@ Quad4::loose_bounding_box () const
...
+
+  return tri3_side->contains_point(p, tol);
+}
+
+
+bool
+Quad4::close_to_point (const Point & p, Real tol) const
+{
+  if (Quad4::has_affine_map())
+    return Quad::close_to_point(p, tol);
```
loganharbour, line 271: @jwpeterson moving the discussion to here.

You said:
> I'm confused. It looks like there is no Quad::contains_point() currently? And the stuff with the Tri elements I would assume works in the affine_map case, not the other way around...

This could (and probably should) just be
```suggestion
    return this->close_to_point(p, tol);
```

loganharbour:
@jwpeterson now for this:
> And the stuff with the Tri elements I would assume works in the affine_map case, not the other way around...

I'm not sure I follow this. If the map is affine (planar), use the standard `Elem::contains_point()`. Otherwise, use the two tris that are definitely planar instead.

loganharbour:
The ordering is also definitely wrong - I need to fix that. The tri needs to be ordered counter-clockwise I believe for the Tri3::contains_point() to work.

If I get a ✅  on this, I can add a unit test. I'm not against it being thrown out for another idea, but just want the functionality.

jwpeterson:
> Otherwise, use the two tris that are definitely planar instead.

A Quad4 isn't made up of two planar sections, it's a bilinear surface... so using two Tris is only an approximation to that.

src/geom/face_quad4.C
jwpeterson: I was talking about `Quad::contains_point()`. I'm guessing it just calls `Elem::contains_point()`.

loganharbour:
> so using two Tris is only an approximation to that.

Okay, I follow. With that, isn't four sufficient? Where the midpoint is the intersection of the diagonals.

jwpeterson:
Any number of Tris is still going to be an approximation, but probably a very fast converging one. I'm thinking two Tris is probably sufficient, but it of course depends on the tolerance being used.

loganharbour:
Actually, no. The more I think about it - if we really do want this, we should just solve the proper system...

jwpeterson:
> if we really do want this, we should just solve the proper system...

Yeah, that's what's done by the base class `Elem::contains_point()`... after some bounding box checks it calls `inverse_map()`. This generic approach is probably slower than what you could get by hand-coding something for Quad4 though.

loganharbour:
See, that's what I thought at first but I couldn't get it to work originally with my ray-tracing intersections. If you say that it's supported, then I'll trust it and look into my issues more.

Closing for now.

jwpeterson:
> If you say that it's supported, then I'll trust it and look into my issues more.

Definitely possible there's a bug... non-planar quads are definitely used in practice, but not sure how many people are running `contains_point()` on them. I'd guess some issue with the bounding box checks and/or tolerances... it would be good to fix whatever it is once you find a valid test case.

loganharbour:
For now, I don't need to validate those intersections for the time being. It's gonna be a much larger task than I have the time for right now. I was hoping that I could just get it working, but no dice.

I've got an issue up for this in MOOSE and I referenced this as a reminder to return to it when I have the time.

Issue #2801, opened 2020-11-12
Nemesis Hex27 output


manavbhatia:
I am writing a 3D mesh with Hex27 elements to an Exodus and Nemesis files. While the visualization of Exodus output looks fine in Paraview, the Nemesis output does now. Please see attached screenshot (left is Exodus, right is Nemesis).

Note that I have used this for both Hex8 and Hex27 and this visualization issue shows up only for Hex27. 

Attached is a code that can reproduce the issue (I am only including the API calls): 

```
        libMesh::MeshTools::Generation::build_cube(*mesh,
                                                   30, 30, 6,
                                                   0, 1,
                                                   0, 1,
                                                   0,.5,
                                                   libMesh::HEX27);
        libMesh::Nemesis_IO(*mesh).write("mesh.e-s"); // this file is causing issues
        libMesh::ExodusII_IO(*mesh).write("mesh2.exo");
```
<img width="1479" alt="Screen Shot 2020-11-12 at 10 22 07 AM" src="https://user-images.githubusercontent.com/4251244/98966485-f0ff6900-24d0-11eb-99e5-08a6b103ecb2.png">

jwpeterson:
Hmm... Hex27 is the only element with a non-trivial node mapping between the libmesh numbering and the ExodusII numbering:

https://github.com/libMesh/libmesh/blob/223c6da2154833257cb39f670913bec23df712a4/src/mesh/exodusII_io_helper.C#L55-L70

So maybe the issue is that we are not actually applying the node mapping in Nemesis, but we only notice the issue for Hex27.

jwpeterson:
Also it's only the mid-face and central nodes where the numbering is permuted, and it looks like those nodes are the ones which are messed up for you.

manavbhatia:
I looked in the source code for NemesisIO_Helper, and am curious if [this line](https://github.com/libMesh/libmesh/blob/223c6da2154833257cb39f670913bec23df712a4/src/mesh/nemesis_io_helper.C#L1771) could be the source of error. Instead of using `conv.get_node_map(j)`, should it use `conv.get_inverse_node_map(j)`, like `ExodusII_IO_Helper` [here](https://github.com/libMesh/libmesh/blob/223c6da2154833257cb39f670913bec23df712a4/src/mesh/exodusII_io_helper.C#L2124)?

manavbhatia:
I included this fix on my libMesh source installation and that fixed the issue. You may want to put this into the source at your end.

[...]

PR #2804, opened 2020-11-15
Fix NemesisIO Hex27 output issue

jwpeterson:
Refs #2801


roystgnr:
It's at least an improvement over the existing code.  Does the fix handle reads as well as writes?

jwpeterson:
The fix was in `build_element_and_node_maps()` which should only be called during writing.

PR #2805, opened 2020-11-22
Share global-to-local map among numeric vectors

fdkong:
Global-to-local map may uses a lot of memory. One quick way to help that is to share the global-to-local map among vectors.

I need to return code to do comparison, but here is some data.

Before PR:

```
(pprof) list PetscVector::init
Total: 1004.79MB
ROUTINE ======================== libMesh::PetscVector::init in /home/kongf/workhome/sawtooth/moosebm/libmesh/installed/include/libmesh/petsc_vector.h
     0  314.07MB (flat, cum) 31.26% of Total
     .     .  746: this->_type = GHOSTED;
     .     .  747:
     .     .  748: /* Make the global-to-local ghost cell map. */
     .     .  749: for (auto i : index_range(ghost))
     .     .  750:  {
     .  314.07MB  751:   _global_to_local_map[ghost[i]] = i;
     .     .  752:  }
     .     .  753:
     .     .  754: /* Create vector. */
     .     .  755: ierr = VecCreateGhost (this->comm().get(), petsc_n_local, petsc_n,
     .     .  756:             petsc_n_ghost, petsc_ghost,
(pprof)
```

After this PR:

```

(pprof) list PetscVector::init
Total: 191.66MB
```

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -74,7 +74,7 @@ template <typename T> class DenseMatrix;
 template <typename T> class SparseMatrix;
 template <typename T> class NumericVector;
 
-
+typedef std::unordered_map<numeric_index_type,numeric_index_type> GlobalToLocalMap;
 
 // ------------------------------------------------------------
 // Do we need constraints for anything?
@@ -473,6 +473,10 @@ public:
   void clear_send_list ()
   {
     _send_list.clear();
+
+    // _global_to_local_map needs to match _send_list
+    if (_global_to_local_map)
+      _global_to_local_map->clear();
   }
 
   /**
@@ -497,6 +501,8 @@ public:
    * computation.
    */
   const std::vector<dof_id_type> & get_send_list() const { return _send_list; }
+
+  const std::shared_ptr<GlobalToLocalMap> get_global_to_local_map() const { return _global_to_local_map; }
 
   /**
    * \returns A constant reference to the \p _n_nz list for this processor.
@@ -1686,6 +1692,12 @@ private:
    * solution on my processor.
    */
   std::vector<dof_id_type> _send_list;
+
+  /**
+   * Map that maps global to local ghost cells (will be empty if not
+   * in ghost cell mode).
+   */
+  std::shared_ptr<GlobalToLocalMap> _global_to_local_map;
 
   /**
    * Function object to call to add extra entries to the sparsity pattern
--- a/include/numerics/petsc_vector.h
+++ b/include/numerics/petsc_vector.h
@@ -109,6 +109,18 @@ public:
                const std::vector<numeric_index_type> & ghost,
                const ParallelType type = AUTOMATIC);
 
+   /**
+    * Constructor. Set local dimension to \p n_local, the global
+    * dimension to \p n, but additionally reserve memory for the
+    * indices specified by the \p ghost_map argument. \p ghost_map
+    * is a global-to-local map for ghosting elements.
+    */
+   PetscVector (const Parallel::Communicator & comm_in,
+                const numeric_index_type N,
+                const numeric_index_type n_local,
+                const std::shared_ptr<GlobalToLocalMap> ghost_map,
+                const ParallelType type = AUTOMATIC);
+
   /**
    * Constructor.  Creates a PetscVector assuming you already have a
    * valid PETSc Vec object.  In this case, \p v is NOT destroyed by the
@@ -158,6 +170,12 @@ public:
   virtual void init (const numeric_index_type N,
                      const numeric_index_type n_local,
                      const std::vector<numeric_index_type> & ghost,
+                     const bool fast = false,
+                     const ParallelType = AUTOMATIC) override;
+
+  virtual void init (const numeric_index_type N,
+                     const numeric_index_type n_local,
+                     const std::shared_ptr<GlobalToLocalMap> ghost_map,
                      const bool fast = false,
                      const ParallelType = AUTOMATIC) override;
 
@@ -431,15 +449,10 @@ private:
   void _restore_array() const;
 
   /**
-   * Type for map that maps global to local ghost cells.
-   */
-  typedef std::unordered_map<numeric_index_type,numeric_index_type> GlobalToLocalMap;
-
-  /**
    * Map that maps global to local ghost cells (will be empty if not
    * in ghost cell mode).
    */
-  GlobalToLocalMap _global_to_local_map;
+  std::shared_ptr<GlobalToLocalMap> _global_to_local_map;
 
   /**
    * This boolean value should only be set to false
@@ -473,7 +486,7 @@ PetscVector<T>::PetscVector (const Parallel::Communicator & comm_in, const Paral
   _last(0),
   _local_form(nullptr),
   _values(nullptr),
-  _global_to_local_map(),
+  _global_to_local_map(nullptr),
   _destroy_vec_on_exit(true),
   _values_manually_retrieved(false),
   _values_read_only(false)
@@ -492,7 +505,7 @@ PetscVector<T>::PetscVector (const Parallel::Communicator & comm_in,
   _array_is_present(false),
   _local_form(nullptr),
   _values(nullptr),
-  _global_to_local_map(),
+  _global_to_local_map(nullptr),
   _destroy_vec_on_exit(true),
   _values_manually_retrieved(false),
   _values_read_only(false)
@@ -512,7 +525,7 @@ PetscVector<T>::PetscVector (const Parallel::Communicator & comm_in,
   _array_is_present(false),
   _local_form(nullptr),
   _values(nullptr),
-  _global_to_local_map(),
+  _global_to_local_map(nullptr),
   _destroy_vec_on_exit(true),
   _values_manually_retrieved(false),
   _values_read_only(false)
@@ -533,12 +546,31 @@ PetscVector<T>::PetscVector (const Parallel::Communicator & comm_in,
   _array_is_present(false),
   _local_form(nullptr),
   _values(nullptr),
-  _global_to_local_map(),
+  _global_to_local_map(nullptr),
   _destroy_vec_on_exit(true),
   _values_manually_retrieved(false),
   _values_read_only(false)
 {
   this->init(n, n_local, ghost, false, ptype);
+}
+
+template <typename T>
+inline
+PetscVector<T>::PetscVector (const Parallel::Communicator & comm_in,
+                             const numeric_index_type n,
+                             const numeric_index_type n_local,
+                             const std::shared_ptr<GlobalToLocalMap> ghost_map,
+                             const ParallelType ptype) :
+  NumericVector<T>(comm_in, ptype),
+  _array_is_present(false),
+  _local_form(nullptr),
+  _values(nullptr),
+  _global_to_local_map(nullptr),
+  _destroy_vec_on_exit(true),
+  _values_manually_retrieved(false),
+  _values_read_only(false)
+{
+  this->init(n, n_local, ghost_map, false, ptype);
 }
 
 
@@ -553,7 +585,7 @@ PetscVector<T>::PetscVector (Vec v,
   _array_is_present(false),
   _local_form(nullptr),
   _values(nullptr),
-  _global_to_local_map(),
+  _global_to_local_map(nullptr),
   _destroy_vec_on_exit(false),
   _values_manually_retrieved(false),
   _values_read_only(false)
@@ -599,8 +631,13 @@ PetscVector<T>::PetscVector (Vec v,
           ierr = ISLocalToGlobalMappingGetIndices(mapping,&indices);
           LIBMESH_CHKERR(ierr);
 
+          // We need to create a GlobalToLocalMap object to take ghosting elements
+          // from Petsc Vec
+          _global_to_local_map = std::make_shared<GlobalToLocalMap>();
+
           for (numeric_index_type i=ghost_begin; i<ghost_end; i++)
-            _global_to_local_map[indices[i]] = i-my_local_size;
+            (*_global_to_local_map)[indices[i]] = i-my_local_size;
+
           this->_type = GHOSTED;
           ierr = ISLocalToGlobalMappingRestoreIndices(mapping, &indices);
           LIBMESH_CHKERR(ierr);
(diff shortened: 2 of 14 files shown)


fdkong:
@roystgnr

tests/base/overlapping_coupling_test.C
```diff
@@ -678,7 +678,7 @@ private:
     system.current_local_solution = libMesh::NumericVector<libMesh::Number>::build(system.comm());
 
     system.current_local_solution->init(system.n_dofs(), system.n_local_dofs(),
-                                        dof_map.get_send_list(), false,
```
roystgnr, line 681: We ought to leave the unit tests as-is, I think.  We should get more than enough coverage of the new map API option via internal library code, but we also want to continue supporting the old send_list API option, and after upgrading library internals that pretty much leaves these couple unit tests as the last places giving the old API test coverage.

src/base/dof_map.C
```diff
@@ -194,6 +195,8 @@ DofMap::DofMap(const unsigned int number,
 
   this->add_coupling_functor(*_default_coupling);
   this->add_algebraic_ghosting_functor(*_default_evaluating);
+
+  _global_to_local_map = std::make_shared<GlobalToLocalMap>();
```
roystgnr, line 199: Can this go in the initializer list?

include/numerics/petsc_vector.h
```diff
@@ -854,8 +951,6 @@ void PetscVector<T>::clear ()
     }
 
   this->_is_closed = this->_is_initialized = false;
-
-  _global_to_local_map.clear();
```
roystgnr, line 858: I hadn't thought about NumericVector::clear() when I was blithely reassuring you that we wouldn't have any problems with sharing this map ... We definitely can't actually clear a shared map here, but if we leave the map untouched, then we aren't entirely cleared, in a sense.  Would it make sense to reset the map to nullptr?

include/numerics/petsc_vector.h
```diff
@@ -533,14 +546,33 @@ PetscVector<T>::PetscVector (const Parallel::Communicator & comm_in,
   _array_is_present(false),
   _local_form(nullptr),
   _values(nullptr),
-  _global_to_local_map(),
+  _global_to_local_map(nullptr),
```
roystgnr, line 549: I feel like I just concluded, when talking to @jwpeterson a few weeks back, that we didn't have many places in the library where moving to C++ delegating constructors would be useful ... and yet here we are.  Now I wish I could remember *why* I'd come to that conclusion.

include/base/dof_map.h
```diff
@@ -498,6 +502,8 @@ class DofMap : public ReferenceCountedObject<DofMap>,
    */
   const std::vector<dof_id_type> & get_send_list() const { return _send_list; }
 
+  const std::shared_ptr<GlobalToLocalMap> get_global_to_local_map() const { return _global_to_local_map; }
```
roystgnr, line 505: This still lets us modify _global_to_local_map from other classes, doesn't it?  Since ```shared_ptr<T>``` seamlessly converts to ```shared_ptr<const T>```, shouldn't we return the latter here?

include/numerics/petsc_vector.h
jwpeterson: > Would it make sense to reset the map to nullptr?

Yes, or simply `_global_to_local_map.reset();` should do the right thing. If the internal reference count is reduced to zero by that call, then it will call the underlying object's destructor.

include/numerics/petsc_vector.h
jwpeterson: Not sure exactly what you are asking, but AFAIK this line doesn't change anything and could be omitted if desired, as it's just calling the default `shared_ptr` constructor.

fdkong: `_global_to_local_map` will be initialized regardless?  What I want is that if check `_global_to_local_map`, it should return `0`

roystgnr: Oh, not asking about this line in this spot in particular.  Just noticing the same line being added again, and again, and again, in different constructors that I'd naively think could all be delegating most of their work to a single target constructor, then handling only their minor differences themselves.

include/base/dof_map.h
fdkong: We can not change this. In pets_vector.h, we have 

`std::shared_ptr<GlobalToLocalMap> _global_to_local_map;`

It need to be compatible with the old API. If send_list is used, we need to add indices into `_global_to_local_map`.

If `_global_to_local_map` points to a constant object, we can not do any modifications

jwpeterson: Sounds like it should not be a `const` function in that case.

fdkong:
Anyone have any thought on why the tests are failing before I am going to dig into the code?

roystgnr:
I was keeping my eye out while reviewing but sadly I didn't see anything suspicious.  If it were me I'd be running
```
LIBMESH_RUN='mpirunxterm -np 2 libtool --mode=execute gdb --args' make check -C tests
```
rather than trying to figure something out from first principles, though.  See which test is failing and where the failing vector is being initialized then proceed from there.  Also useful in our tests directory is:
```
LIBMESH_OPTIONS='--re some_regex_identifying_the_failing_test'
```

fdkong:
>I was keeping my eye out while reviewing but sadly I didn't see anything suspicious. If it were me I'd be running

> LIBMESH_RUN='mpirunxterm -np 2 libtool --mode=execute gdb --args' make check -C tests

Which directory I should go before I type this magic command line?

I did `make -j16 check`, and it takes forever to run the whole world. I just want to run the failed test.

roystgnr:
> Which directory I should go before I type this magic command line?

I really did pick the wrong afternoon to vanish for a few hours, didn't I?

The key thing there is the standard option "-C tests": ```make -C foo``` tells make to run in directory foo.  So ```make -C tests``` could be run from your build dir, or ```make -C your_build_dir/tests``` could be run from anywhere.

But the real magic is that "mpirunxterm" isn't a standard command, it's my own hackish shell script:
```
#!/bin/bash

numproc=$2

shift
shift

echo "$@"
echo \'"$@"\'" && read -p COMPLETED finalinput || read -p FAILED finalinput; sleep 20"

exec mpirun --oversubscribe -np $numproc xterm -e bash -c \'"$@"\'" && read -p COMPLETED finalinput || read -p FAILED finalinput; sleep 20"
```

Which spawns an xterm for each rank with whatever command I asked for.

lindsayad:
It would be nice to know if this failure was because of mpi or threads...

lindsayad:
Ok yea unit tests run fine with just threads, but fails with just mpi. So threads are not necessary to cause the failure

fdkong:
The issue is that `global_to_local_map` is modified after PetscVector::init is called. So PetscVector is inconsistent with shared `global_to_local_map`

The `global_to_local_map` is first built right here

```
0: 0   libmesh_dbg.0.dylib                 0x00000001048ce302 libMesh::print_trace(std::__1::basic_ostream<char, std::__1::char_traits<char> >&) + 450
1: 1   libmesh_dbg.0.dylib                 0x000000010479bbfd libMesh::DofMap::prepare_send_list() + 61
2: 2   libmesh_dbg.0.dylib                 0x00000001056c1e49 libMesh::System::init_data() + 985
3: 3   libmesh_dbg.0.dylib                 0x00000001056c1a04 libMesh::System::init() + 532
4: 4   libmesh_dbg.0.dylib                 0x0000000105663b26 libMesh::EquationSystems::init() + 1430
5: 5   unit_tests-dbg                      0x000000010282f59e MeshInputTest::testProjectionRegression(libMesh::MeshBase&, libMesh::DynaIO&, std::__1::array<double, 4ul>) + 990
6: 6   unit_tests-dbg                      0x000000010282e4ef MeshInputTest::testDynaFileMappings(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::array<double, 4ul>) + 655
7: 7   unit_tests-dbg                      0x000000010282e0b7 MeshInputTest::testDynaFileMappingsFEMEx5() + 151
```

And then we call 

```
  current_local_solution->init (this->n_dofs(), this->n_local_dofs(),
                                _dof_map->get_global_to_local_map(), false,
                                GHOSTED);
```

And then we modify  `global_to_local_map`, and leave `PetscVector` out of sync 

```
1: 1   libmesh_dbg.0.dylib                 0x00000001064dcbfd libMesh::DofMap::prepare_send_list() + 61
2: 2   libmesh_dbg.0.dylib                 0x0000000106c62fd9 libMesh::DynaIO::add_spline_constraints(libMesh::DofMap&, unsigned int, unsigned int) + 1017
3: 3   unit_tests-dbg                      0x0000000101e26602 MeshInputTest::testProjectionRegression(libMesh::MeshBase&, libMesh::DynaIO&, std::__1::array<double, 4ul>) + 1090
44: : 4   unit_tests-dbg                      0x0000000101e254ef MeshInputTest::testDynaFileMappings(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::array<double, 4ul>) + 655
5: 5   unit_tests-dbg                      0x0000000101e250b7 MeshInputTest::testDynaFileMappingsFEMEx5() + 1514   unit_tests-dbg                      0x000000010282e4ef MeshInputTest::testDynaFileMappings(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::array<double, 4ul>) + 655
5: 5   unit_tests-dbg                      0x000000010282e0b7 MeshInputTest::testDynaFileMappingsFEMEx5() + 151
```

tests/mesh/mesh_input.C
```diff
@@ -604,8 +604,12 @@ public:
     unsigned int n_var =
       sys.add_variable("n", Order(order), RATIONAL_BERNSTEIN);
 
-    es.init();
     dyna.add_spline_constraints(sys.get_dof_map(), sys.number(), n_var);
+    // `add_spline_constraints` secretly changed `send_list` and `global_to_local_map`
+    // by calling `dof_map.prepare_send_list()`
+    // If we initialize system earlier, then PetscVector will be out of sync
+    // We should init system only when everything is ready.
+    es.init();
```
fdkong, line 612: The current fix does work either.

fdkong:
I guess that sharing the data structure `global_to_local_map` might not be the good idea at all. Because it is legal for `libmesh` to modify `send_list` and `global_to_local_map` anytime if `libmesh` wants. That definitely leave PetscVector inconsistent. 

`void testProjectionRegression(MeshBase & mesh, DynaIO & dyna, std::array<Real, 4> expected_norms)` is a good example for this

tests/mesh/mesh_input.C
roystgnr: Ironically, I think one of my other open PRs would end up fixing this issue *by accident*.

But I agree with you that there's a deeper issue that needs to be fixed; if at the moment we have a test exposing the problem that's a good thing, not a bad thing.

roystgnr:
Stupid DynaIO tricks aside, IMHO we'd really intended to only allow users to modify the send_list via callbacks from with the DofMap's own send_list calculations, as provided in attach_extra_send_list_object() or attach_extra_send_list_function(), and so we would only be doing so from within a reinit() where we would be initializing the ghosted solution vector with the new send_list shortly thereafter.

The trouble appears to be that we can also add new constraints at any time, then tell the DofMap to repeat its send_list calculations (thereby taking the new constraints into account), and *this doesn't entail any NumericVector::reinit()*.  So while my first thought had been "if the user tells DofMap to change the send_list after it's in use, let's just Copy-on-Write the thing" (which ought to be easy with shared_ptr), my second is "this is yet a deeper underlying bug waiting to happen, isn't it?"  What we really need is a way to easily ensure, whenever the send_list is changed, we also end up reinitializing any vectors that need it.  We don't necessarily need to *enforce* that, since anyone doing something more low level than "just reinit()" should be thinking in Caveat Emptor mode already ... but we do at least need to be able to mark a change and assert that we're not using an out-of-date map, because this was a pretty awful bug to track down, and it would be even worse if someone ever just got a corrupt permutation that *didn't* lead to an assertion failure or crash.

I'm not yet sure how to do any of that properly...

tests/mesh/mesh_input.C
fdkong: I am curious how you did you fix this test? Could you point me the PR?

fdkong:
Yeah, I am not sure if it is still a good idea to save memory in this way. I could see there would be some really hard-to-manage bugs in the future. 

If https://github.com/idaholab/moose/issues/16339 is addressed, the current PR is not that necessay  because we will end up using only one ghosted  solution vector.

I would like to pursue this PR if we have a clear way to sync data between `dof_map` and `PetscVector`.  

How about we are going to init `PetscVector` whenever  if` DofMap::prepare_send_list` is triggered and the system is already initialized? But in some cases, users may NOT want to change `PetscVector` even `send_list` is changed?

tests/mesh/mesh_input.C
roystgnr: Check out #2808 - we basically don't wait for application-level code to apply IGA constraints any more, we make the spline node constraint connections storable in the Mesh, so the DofMap itself can translate them into dof_constraints (whether raw or by merging them with overlapping Dirichlet constraints) as soon as it needs to, and so we don't end up initializing things with one set of constraints only to have to change those constraints right afterward.

roystgnr:
> Yeah, I am not sure if it is still a good idea to save memory in this way. I could see there would be some really hard-to-manage bugs in the future.

I agree wholeheartedly; we really need to lock this set of APIs down if we want to allow this level of coupling, and doing that will be dangerous and prone to mistakes.  If you can get the same savings from a higher-level fix that would probably be safer.

> How about we are going to init PetscVector whenever DofMap::prepare_send_list is triggered and the system is already initialized?

That was my first thought, but it only works for vectors managed by the system.  We'd have to combine it with clever use of shared_ptr - instead of letting an existing send_list/map be changed, we'd release our pointers to it (causing it to be freed if nobody else has a handle to it but leaving it intact if the user has non-system-managed vectors using it) and generate a new send_list+map in its place. 

> But in some cases, users may NOT want to change PetscVector even send_list is changed?

I think it might be reasonable to require that "if your vector is managed by the System via add_vector, then its send_list will change when the System's changes, and if you don't like that then just don't let your vector be managed by the System".

fdkong:
OK, idaholab/moose#16339  should be better way. 

If we still have memory issue after  idaholab/moose#16339 is done, I will rethink this PR again.

YaqiWang:
How do we handle matrices since matrices are already sharing data in DofMap?

roystgnr:
We handle matrices by making it too hard for users to handle their own matrices, so they're just stuck with us?  :-D

Seriously, though, I had to do some refactoring (#2776, and I think there was one more) this year just to be able to (in #2787 and #2808) solve a linear system without a matrix attached to a System.

PR #2806, opened 2020-11-24
Add bounding box scaling

loganharbour:


src/geom/bounding_box.C
```diff
@@ -133,4 +133,25 @@ Real BoundingBox::signed_distance(const Point & p) const
     }
 }
 
+
+void BoundingBox::scale(const Real factor)
+{
+  auto scale_dim = [&](const unsigned int dim) {
+    if (this->first(dim) != std::numeric_limits<Real>::max())
```
roystgnr, line 140: This can't be right?  In the degenerate case first(dim) would be *negative* max, no?

We should also test second(dim) here, lest someone have a half-infinite box in that direction.

include/geom/bounding_box.h
```diff
@@ -140,6 +140,13 @@ class BoundingBox : public std::pair<Point, Point>
    * d <= 0 if the point is inside the Bounding Box
    */
   Real signed_distance(const Point & p) const;
+
+  /**
+   * Scales each dimension of the bounding box by \p factor.
+   *
+   * Has no effect for dimensions that are invalid.
```
roystgnr, line 147: Maybe replace "invalid" by "at numeric_limits::max in either direction"?  That's still a perfectly valid bounding box, it's just not one we can sensibly apply a relative scaling to in that dimension.

src/geom/bounding_box.C
loganharbour:
```
  void invalidate ()
  {
    for (unsigned int i=0; i<LIBMESH_DIM; i++)
      {
        this->first(i)  =  std::numeric_limits<Real>::max();
        this->second(i) = -std::numeric_limits<Real>::max();
      }
  }
```

loganharbour:
- Added the check for an "invalid" max()
- Updated doxygen comments
- Added assertion for max()(dim) > min()(dim)

loganharbour:
Alright - this is good to go or for another review.

src/geom/bounding_box.C
roystgnr: Ah, okay. I mistook "invalid" for "(half-)infinite", thought you were trying to handle the latter cases.

roystgnr (approved):

PR #2808, opened 2020-11-25
Make IGA constraints automatically applied, DirichletConstraint-compatible

roystgnr:
The automatic application ought to enable us to use DynaIO meshes as a drop-in replacement for non-isogeometric meshes in other formats in many cases; we no longer need user-level or middleware-level code to carefully apply those constraints.  (And I do mean carefully; I wasn't doing it correctly in our own example, even!)

The DirichletConstraint compatibility should let us solve far more problems, since we won't have to backpedal in the physics kernels and use penalty BCs.


jwpeterson:
Do you think this would fix the failure related to DynaIO that we saw in #2800 when removing recursive constraint application?

roystgnr:
Huh... gun to my head I'd guess "no" - it was fem_system_ex5 where I really screwed up the manual constraint application, whereas those failures look like they're in unit tests.  But I wouldn't be very confident in my guess; we should definitely try rebasing that PR after this one gets merged.

PR #2810, opened 2020-11-25
Fix --disable-metaphysicl builds

roystgnr:
I put header includes in the wrong place here; we still need
linear_solver.h for spline projections regardless of whether we're using
MetaPhysicL for GMG projection matrices.


roystgnr:
Merging this since it's a straightforward fix for a dumb bug.  I'll wait for the end of the week with my more extensive PRs.

PR #2811, opened 2020-11-25
Quadruple-precision fixes

roystgnr:
This isn't a complete set of "make check works beautifully again" fixes, it's just what I needed to get things compiling when I was experimenting with IGA code at quadruple precision last month.  I'll try to go about filling out the bugfixes later, but lest I forget about this branch completely, for now I want to at least see if git complains about conflicts or CI about failures.


roystgnr:
I still need to look at unit tests, especially for the FE types dependent on FDM gradients and hessians, but this is probably good to merge after #3049, just to get the compilation fixes in.

PR #2819, opened 2020-12-10
Update submodules

lindsayad:
Refs libmesh/autoconf-submodule#4 and idaholab/moose#16067

This robustifies scraping of PETSc configure if PETSc was configured on a system in which core utilities were found in `/usr/bin` and our system's core utilities are in `/bin`


roystgnr:
I guess this "pr60_bootstrapped" sort of submodule branch is the workflow we want when we've got an update that's important enough to rush downstream but not important enough to justify a new release number for the submodule package?  I like it; I just think it's a new enough process to warrant specific mention.

roystgnr (approved):

PR #2821, opened 2020-12-14
Initialize next_unpartitioned_unique_id

roystgnr:
There's a chance this is responsible for a valgrind warning that @fdkong
managed to trigger.

### DIFF
--- a/src/mesh/distributed_mesh.C
+++ b/src/mesh/distributed_mesh.C
@@ -167,6 +167,8 @@ DistributedMesh::DistributedMesh (const UnstructuredMesh & other_mesh) :
 #ifdef LIBMESH_ENABLE_UNIQUE_ID
   _next_unique_id = other_mesh.parallel_max_unique_id() +
                     this->processor_id();
+  _next_unpartitioned_unique_id = _next_unique_id +
+    (this->n_processors() - this->processor_id());
 #endif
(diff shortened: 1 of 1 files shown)


fdkong:
I am going try this out. Thanks

src/mesh/distributed_mesh.C
```diff
@@ -167,6 +167,8 @@ DistributedMesh::DistributedMesh (const UnstructuredMesh & other_mesh) :
 #ifdef LIBMESH_ENABLE_UNIQUE_ID
   _next_unique_id = other_mesh.parallel_max_unique_id() +
                     this->processor_id();
+  _next_unpartitioned_unique_id = _next_unique_id +
+    (this->n_processors() - this->processor_id());
```
fdkong, line 171: Thanks. I think our code goes through` DistributedMesh::DistributedMesh (const DistributedMesh & other_mesh)` not this constructor. But I will try this out.

roystgnr: Yeah, I think you're right.  But I didn't find anything else that might be getting used while uninitialized, so this was my best shot.

roystgnr:
@fdkong reports that this didn't fix his valgrind issue, but since it might fix issues with other code paths I'm merging anyway.

PR #2824, opened 2020-12-22
Enable and test IGA meshes in reduced_basis code

roystgnr:
This includes some fixes to enable the interaction of our reduced_basis and IGA code, and some bug fixes to the prior round of IGA code, and some new test coverage that helped uncover all the above problems, and some new apps and examples capabilities that made it easy to examine those problems (as well as a few red herrings along the way).


roystgnr:
IM from John, for the permanent record: "I ran all of our tests that matched the "isolated" and "node_elem" regexes on an old master (6594fad14) plus your cherry-picked commit 706115af. We use "isolated nodes" (NodeElems) for things like point masses, point loads, and some contact implementations, so I think these tests are probably the most relevant ones to your change. They all passed."

Not planning to merge until the 4th or 5th, so we've loads of time for more tests and inspection, but I think this has got it!

PR #2825, opened 2020-12-23
Catch exception thrown by capnp to get improved error message.

jwpeterson:
This fixes an issue I had where a capnp exception went uncaught and the libmesh terminate handler was called, but no nice error message with line numbers, etc. was printed.


dknez:
Looks good to me, thanks!

PR #2828, opened 2021-01-05
Eliminate leak in merge_ghost_functor_outputs

lindsayad:
We appear to have a leak from this code as shown [here](https://civet.inl.gov/job/653571/). This patch appears to fix the issue as shown [here](https://civet.inl.gov/job/654071/).

### DIFF
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -1501,7 +1501,10 @@ merge_ghost_functor_outputs(GhostingFunctor::map_type & elements_to_ghost,
                       std::set<CouplingMatrix *>::iterator temp_it =
                         temporary_coupling_matrices.find(const_cast<CouplingMatrix *>(existing_it->second));
                       if (temp_it != temporary_coupling_matrices.end())
+                      {
+                        delete *temp_it;
                         temporary_coupling_matrices.erase(temp_it);
+                      }
 
                       existing_it->second = nullptr;
                     }


jwpeterson:
Nice catch. I haven't looked at this in detail, but it seems to me that `temporary_coupling_matrices` should store `unique_ptrs` so we can avoid calling `new`/`delete`. @roystgnr ?

lindsayad:
I think @roystgnr is on vacation for a couple more weeks...?

jwpeterson:
> I think @roystgnr is on vacation for a couple more weeks...?

OK, well this fix can definitely be merged as-is since it fixes the immediate problem, but it would be good to address the underlying issue of why we are manually managing memory here.

fdkong:
Yes, we should use smart pointers because they are smart and valgrind-clean, lol

src/base/dof_map.C
```diff
@@ -1483,7 +1483,13 @@ merge_ghost_functor_outputs(GhostingFunctor::map_type & elements_to_ghost,
...
                         {
                           CouplingMatrix * cm = new CouplingMatrix(*existing_it->second);
-                          temporary_coupling_matrices.insert(cm);
+#ifndef NDEBUG
+                          auto dbg_pr =
+#endif
+                            temporary_coupling_matrices.insert(cm);
+                          libmesh_assert_msg(dbg_pr.second,
+                                             "We must actually insert this new coupling matrix or "
+                                             "else we are going to leak");
```
roystgnr, line 1492: How is this ever going to fail?  If we just got a pointer from new, then it won't alias any existing pointer and it won't be in the container.  If we didn't, then new should have thrown an exception and we shouldn't have made it to the assert.

src/base/dof_map.C
```diff
@@ -1501,7 +1507,10 @@ merge_ghost_functor_outputs(GhostingFunctor::map_type & elements_to_ghost,
                       std::set<CouplingMatrix *>::iterator temp_it =
                         temporary_coupling_matrices.find(const_cast<CouplingMatrix *>(existing_it->second));
                       if (temp_it != temporary_coupling_matrices.end())
+                      {
+                        delete *temp_it;
```
roystgnr, line 1511: Yeah, not sure how I missed this.

roystgnr (approved):

roystgnr:
The code here was written before we started requiring C++11; otherwise it would have been smart pointers from the start.  We should definitely migrate to them.

roystgnr:
And, yeah, on vacation until the 18th.

src/base/dof_map.C
lindsayad: Yea...that seems pretty obvious when I give any level of thought about it

PR #2829, opened 2021-01-06
ExodusII_IO: Add number of nodes per Elem sanity check.

jwpeterson:
This adds a check for invalid Exodus files with mismatched ""elem_type" and "num_nod_per_el" values. Previously such files would invoke undefined behavior in opt mode and throw an assert in dbg mode. Now both modes die with a reasonable error message like the following: 
```
Error: Exodus file says 3 nodes per Elem, but Elem type EDGE2 has 2 nodes.
```


friedmud (approved):

roystgnr:
That's interesting.  What generated the invalid file?

jwpeterson:
> What generated the invalid file?

I believe it was an Abaqus -> Exodus converter script that we use internally.

PR #2832, opened 2021-01-08
Avoid allocating  matrix when using JFNK

fdkong:


fdkong:
@YaqiWang  Could you try if this works for you? 

We might have a long term solution for this. But this simple fix is not a bad idea since people want to use griffin now.

lindsayad (approved):
This seems logically correct to me and more consistent with the old behavior

PR #2833, opened 2021-01-12
The residual history vector pointer for the PETSc linear solver should…

lcnsqr:
… be declared as const, otherwise the compilation will fail with recent (git master branch) PETSc versions

### DIFF
--- a/src/solvers/petsc_linear_solver.C
+++ b/src/solvers/petsc_linear_solver.C
@@ -1262,7 +1262,7 @@ void PetscLinearSolver<T>::get_residual_history(std::vector<double> & hist)
   // methods, the number of residuals returned in the history
   // vector may be different from what you are expecting.  For
   // example, TFQMR returns two residual values per iteration step.
-  PetscReal * p;
+  const PetscReal * p;
   ierr = KSPGetResidualHistory(_ksp, &p, &its);
   LIBMESH_CHKERR(ierr);
 
@@ -1295,7 +1295,7 @@ Real PetscLinearSolver<T>::get_initial_residual()
   // methods, the number of residuals returned in the history
   // vector may be different from what you are expecting.  For
   // example, TFQMR returns two residual values per iteration step.
-  PetscReal * p;
+  const PetscReal * p;
   ierr = KSPGetResidualHistory(_ksp, &p, &its);
   LIBMESH_CHKERR(ierr);
 
--- a/src/solvers/petsc_linear_solver.C
+++ b/src/solvers/petsc_linear_solver.C
@@ -1263,12 +1263,12 @@ void PetscLinearSolver<T>::get_residual_history(std::vector<double> & hist)
   // vector may be different from what you are expecting.  For
   // example, TFQMR returns two residual values per iteration step.
 
-  // Recent development versions of PETSc require the residual
-  // history vector pointer to be declared as const
-#if PETSC_VERSION_RELEASE == 0
-  const PetscReal * p;
-#else
+  // Recent versions of PETSc require the residual
+  // history vector pointer to be declared as const.
+#if PETSC_VERSION_LESS_THAN(3,14,4) && PETSC_VERSION_RELEASE
   PetscReal * p;
+#else
+  const PetscReal * p;
 #endif
 
   ierr = KSPGetResidualHistory(_ksp, &p, &its);
@@ -1304,13 +1304,14 @@ Real PetscLinearSolver<T>::get_initial_residual()
   // vector may be different from what you are expecting.  For
   // example, TFQMR returns two residual values per iteration step.
 
-  // Recent development versions of PETSc require the residual
-  // history vector pointer to be declared as const
-#if PETSC_VERSION_RELEASE == 0
-  const PetscReal * p;
-#else
+  // Recent versions of PETSc require the residual
+  // history vector pointer to be declared as const.
+#if PETSC_VERSION_LESS_THAN(3,14,4) && PETSC_VERSION_RELEASE
   PetscReal * p;
+#else
+  const PetscReal * p;
 #endif
+
 
   ierr = KSPGetResidualHistory(_ksp, &p, &its);
   LIBMESH_CHKERR(ierr);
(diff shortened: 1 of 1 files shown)


jwpeterson:
We'll probably have to add `#ifdefs` to make sure this continues compiling with non-master PETSc.

jwpeterson:
@lcnsqr FYI I sent you an invite to join a GitHub group whose PRs are tested automatically.

jwpeterson:
For future reference, the error message with an older PETSc is:
```
../src/solvers/petsc_linear_solver.C:1266:10: error: no matching function for call to 'KSPGetResidualHistory'
  ierr = KSPGetResidualHistory(_ksp, &p, &its);
         ^~~~~~~~~~~~~~~~~~~~~
../src/solvers/petsc_linear_solver.C:1545:16: note: in instantiation of member function 'libMesh::PetscLinearSolver<double>::get_residual_history' requested here
template class PetscLinearSolver<Number>;
               ^
/opt/min_moose/petsc/mpich_petsc-3.7.6/clang-opt-superlu/include/petscksp.h:126:29: note: candidate function not viable: no known conversion from 'const PetscReal **' (aka 'const double **') to 'PetscReal **' (aka 'double **') for 2nd argument
PETSC_EXTERN PetscErrorCode KSPGetResidualHistory(KSP,PetscReal*[],PetscInt *);
                            ^
../src/solvers/petsc_linear_solver.C:1299:10: error: no matching function for call to 'KSPGetResidualHistory'
  ierr = KSPGetResidualHistory(_ksp, &p, &its);
         ^~~~~~~~~~~~~~~~~~~~~
../src/solvers/petsc_linear_solver.C:1545:16: note: in instantiation of member function 'libMesh::PetscLinearSolver<double>::get_initial_residual' requested here
template class PetscLinearSolver<Number>;
               ^
/opt/min_moose/petsc/mpich_petsc-3.7.6/clang-opt-superlu/include/petscksp.h:126:29: note: candidate function not viable: no known conversion from 'const PetscReal **' (aka 'const double **') to 'PetscReal **' (aka 'double **') for 2nd argument
PETSC_EXTERN PetscErrorCode KSPGetResidualHistory(KSP,PetscReal*[],PetscInt *);
```

lcnsqr:
@jwpeterson thank you for the feedback! Apparently, this is an issue just for the development versions of PETSc, so I added an #if directive to check whether it's a release version before declaring the vector pointer.

I tested with the newest development version and the last release version of PETSc and now libmesh compiles fine in both.

jwpeterson:
Oops, your second commit introduced trailing whitespace, which we don't allow. I could fix it for you and push to your branch, but I see that you are working on master, so maybe best if you handle it. BTW, I think running "git log --check" should show the lines that have trailing whitespace.
```
commit d62d122c79f2be6c0fff2e78e3fdfb32576b4a32 (lcnsqr/master)
Author: Luciano Siqueira <lcnsqr@gmail.com>
Date:   Tue Jan 12 16:16:23 2021 -0300

    The residual history vector pointer for the PETSc linear solver should be declared as const only for the development (git master branch) PETSc versions

src/solvers/petsc_linear_solver.C:1266: trailing whitespace.
+  // Recent development versions of PETSc require the residual 
src/solvers/petsc_linear_solver.C:1306: trailing whitespace.
+ 
src/solvers/petsc_linear_solver.C:1307: trailing whitespace.
+  // Recent development versions of PETSc require the residual 

```

jwpeterson:
The other issue with your fix is that it will fail to compile for anyone using an older version of PETSc master than you, since PETSC_VERSION_RELEASE == 0 for them as well. We don't have a good fix for this currently, other than "if you're using PETSc master, hopefully you know what you're doing and can update/fix the problem".

lcnsqr:
Oops, sorry about the trailing spaces. I already fixed that.

lcnsqr:
Yeah, people working with PETSc master will probably update their sources once they see the issue.

jwpeterson:
For future reference, this change seems to have happened in petsc/petsc@94a4d4eb42d, which is not yet part of any tagged release. Once it makes it into some release, we can then go back and update this line of code with the appropriate `PETSC_VERSION_LESS_THAN` checks.

Since the commit in question is about some other topic, they may not realize it breaks API compatibility. @fdkong is that something you would be able to inquire about?

src/solvers/petsc_linear_solver.C
```diff
@@ -1262,7 +1262,15 @@ void PetscLinearSolver<T>::get_residual_history(std::vector<double> & hist)
   // methods, the number of residuals returned in the history
   // vector may be different from what you are expecting.  For
   // example, TFQMR returns two residual values per iteration step.
+
+  // Recent development versions of PETSc require the residual
+  // history vector pointer to be declared as const
+#if PETSC_VERSION_RELEASE == 0
```
fdkong, line 1268: Not quite right. We need to use version comparison  as well 

!PETSC_RELEASE_LESS_THAN(major,minor,subminor) should work

fdkong (changes requested):

src/solvers/petsc_linear_solver.C
lcnsqr: Unfortunately, the major, minor and subminor version numbers are the same for the release version and the current PETSc git master. I'm sure it will change in the near future, but for now the only difference is the PETSC_VERSION_RELEASE info.

I added the PETSC_RELEASE_LESS_THAN check in my last commit. I assume the const declaration will be used in PETSc 3.14.4 and above.

jwpeterson: Yeah, but what version should we put? Is it guaranteed to be in 3.15.x since it's in master now and it's not in the 3.14.x series? I can take care of making this change...

jwpeterson: > I can take care of making this change...

Oops, I see that @lcnsqr already fixed it, but note that as @fdkong says we should use `PETSC_RELEASE_LESS_THAN(3,14,4);` which does the same logic as your check.

jwpeterson (approved):
Your most recent version looks good to me, I'll (squash) merge this as-is and then make the minor update to `PETSC_RELEASE_LESS_THAN()` subsequently.

PR #2836, opened 2021-01-18
Fix print_personal

dewenyushu:
- Fix the communicator in PetscMatrix<T>::print_personal() in order to avoid PETSc error
- Make sure that the output matrix is initialized in PetscMatrix<T>::matrix_matrix_mult() (see this fix in PR #2838 instead )

### DIFF
--- a/src/numerics/petsc_matrix.C
+++ b/src/numerics/petsc_matrix.C
@@ -725,7 +725,7 @@ void PetscMatrix<T>::print_personal(std::ostream & os) const
   // Print to screen if ostream is stdout
   if (os.rdbuf() == std::cout.rdbuf())
     {
-      ierr = MatView(_mat, PETSC_VIEWER_STDOUT_SELF);
+      ierr = MatView(_mat, NULL);
       LIBMESH_CHKERR(ierr);
     }
 


src/numerics/petsc_matrix.C
```diff
@@ -725,7 +725,7 @@ void PetscMatrix<T>::print_personal(std::ostream & os) const
   // Print to screen if ostream is stdout
   if (os.rdbuf() == std::cout.rdbuf())
     {
-      ierr = MatView(_mat, PETSC_VIEWER_STDOUT_SELF);
+      ierr = MatView(_mat, NULL);
```
jwpeterson, line 728: I perused the docs for [MatView](https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Mat/MatView.html] briefly, but couldn't see what passing `NULL` for the second arg does?

src/numerics/petsc_matrix.C
```diff
@@ -769,7 +769,7 @@ void PetscMatrix<T>::print_personal(std::ostream & os) const
       // of the file internally.  Since print_personal() takes a reference to
       // an ostream, we have to do an extra step...  print_personal() should probably
       // have a version that takes a string to get rid of this problem.
-      ierr = PetscViewerASCIIOpen( this->comm().get(),
+      ierr = PetscViewerASCIIOpen( PetscObjectComm((PetscObject)_mat),
```
jwpeterson, line 772: Hmm... if there is there some difference between the communicator of the PetscMatrix and the underlying Mat, I believe we will run into other issues as well, no? Can you explain more about the situation where this arises?

jwpeterson:
The two failing tests failed in a weird way which can't be due to your change, probably best to just invalidate them and see if the error goes away.
```
=== configuring in contrib/netcdf/v4 (/opt/civet/build_0/libmesh/build/contrib/netcdf/v4)
configure: running /bin/bash ../../../../contrib/netcdf/v4/configure --disable-option-checking '--prefix=/opt/civet/build_0/libmesh/installed'  '--with-methods=opt' '--enable-silent-rules' '--enable-unique-id' '--disable-warnings' '--enable-openmp' '--disable-maintainer-mode' '--with-metis=PETSc' 'PETSC_DIR=/opt/min_moose/petsc/mpich_petsc-3.7.6/gcc-opt-superlu' 'CC=mpicc' 'FC=mpif90' 'F77=mpif77' 'CXX=mpicxx' 'METHODS=opt' '--disable-netcdf-4' '--disable-dap' '--disable-curl' '--disable-testsets' 'CXX=mpicxx' 'CC=mpicc' 'F77=mpif77' 'FC=mpif90' 'CPPFLAGS= ' 'LIBS= -ldl ' --cache-file=/dev/null --srcdir=../../../../contrib/netcdf/v4
configure: netCDF 4.6.2
checking build system type... x86_64-pc-linux-gnu
checking host system type... x86_64-pc-linux-gnu
checking target system type... x86_64-pc-linux-gnu
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... configure: error: newly created file is older than distributed files!
Check your system clock
configure: error: ../../../../contrib/netcdf/v4/configure failed for contrib/netcdf/v4
ERROR: exiting with code 1
```

src/numerics/petsc_matrix.C
dewenyushu: Thanks @jwpeterson ! I came across error in `MatView` about 'Arguments must have same communicator' while trying to print out a `PetscMatrix` for debugging purposes. @fdkong suggested me to make the two changes in `print_personal`, which resolved the issue.

@fdkong Would you mind explaining a little more on what the NULL does here?

src/numerics/petsc_matrix.C
dewenyushu: @fdkong will we run into other issues due to communicator differences?

src/numerics/petsc_matrix.C
fdkong: `NULL` is the right option to use because we want Mat print out on the screen for both parallel and serial cases.  Here are codes:

```
975: PetscErrorCode MatView(Mat mat,PetscViewer viewer)
976: {
977:   PetscErrorCode    ierr;
978:   PetscInt          rows,cols,rbs,cbs;
979:   PetscBool         isascii,isstring,issaws;
980:   PetscViewerFormat format;
981:   PetscMPIInt       size;

986:   if (!viewer) {PetscViewerASCIIGetStdout(PetscObjectComm((PetscObject)mat),&viewer);}
```

src/numerics/petsc_matrix.C
fdkong: `Viewer` and `Mat` have to use the same communicator. We have PETSc code like

`PetscCheckSameComm(mat,1,viewer,2);`

src/numerics/petsc_matrix.C
```diff
@@ -1197,6 +1197,10 @@ void PetscMatrix<T>::matrix_matrix_mult (SparseMatrix<T> & X_in, SparseMatrix<T>
 
   ierr = MatMatMult(_mat, X->_mat, MAT_INITIAL_MATRIX, PETSC_DEFAULT, &Y->_mat);
   LIBMESH_CHKERR(ierr);
+
+  // Specify that the new matrix is initialized and close it.
+  Y->_is_initialized = true;
+  Y->close();
```
fdkong, line 1203: You might just set closed flag to true. Do not need to really do communication. `_mat` from `MatMatMult` is closed for sure.

fdkong (changes requested):

src/numerics/petsc_matrix.C
dewenyushu: Thanks @fdkong ! In this case, should I just remove `Y->close();` or do `libmesh_assert(Y->closed());` instead?

fdkong: Just remove `Y->close()`

src/numerics/petsc_matrix.C
jwpeterson: @fdkong I agree that `Viewer` and `Mat` should should use the same Communicator! The issue I see is that there should be no difference between `this->comm().get()` (where `this` refers to the `PetscMatrix`) and `PetscObjectComm((PetscObject)_mat` under most normal circumstances. Do you agree? Otherwise it means that someone has changed either the` PetscMatrix` or  `Mat` communicator but not updated the other one about the change...

src/numerics/petsc_matrix.C
jwpeterson: OK, makes sense to me, I was not aware of this `if (!viewer)` logic.

src/numerics/petsc_matrix.C
dewenyushu: It seems that I should keep this line as it originally was then? I am not sure which is better..

jwpeterson: Logically I think it should be the same. If it actually "fixes" a problem, then there is a larger issue that we would need to look into and analyze, which is how the Communicator object of the `PetscMatrix` and its underlying `_mat` can possibly be different.

dewenyushu: Thanks @jwpeterson . IIRC the error I was having actually came from the `MatView()`. Let me try my case without this line of change and see if I still see the error.

fdkong: I do not have time to verify it at this moment. From my understanding (I could be totally wrong), PETSc can not use the same communicator as libmesh or other packages. If PETSc can use  the same comm, I will be puzzled on

`MPI_Isend(comm, .., tag.., ).`

PETSc could generate the same tag as libmesh, and then how we make sure we do not mess up this kind of communication.

fdkong: OK, I understood now. PETSc will retrieve an inner comm from libmesh comm if it exists, otherwise, PETSc will duplicate a communicator and imbed it into libmesh-comm.  So PETSc will always have its own comm. So `MPI_Isend` will be fine.  

`PetscCheckSameComm` will still return true as long as libmesh-comm and PETSc-comm share the same group.

So the bottom line is that  you could leave this code alone

dewenyushu: Thanks @fdkong for looking into this! I checked my case again and it seems this line of change is not necessary. I am reverting it back to what it was.

fdkong: Yes, your case will not touch this line of code at all. It is why you did not see anything. But I do not think we need to change this code

fdkong (approved):

jwpeterson (approved):
Quite a discussion for a 1-line PR!

fdkong:
> Quite a discussion for a 1-line PR!

Beautiful one line of code

PR #2838, opened 2021-01-19
PetscMatrix functions change for static condensation

dewenyushu:
Several changes are made in the PetscMatrix class to enable static condensation in this [PR](https://github.com/idaholab/moose/pull/16017)

- Add an unsorted version for `create_submatrix`, which ensures that no zero diagonal entry appears in the Jacobian matrix
- Make sure `matrix_matrix_mult` initializes and closes the product matrix 
- Make sure that `add_sparse_matrix` works properly in parallel

### DIFF
--- a/include/numerics/sparse_matrix.h
+++ b/include/numerics/sparse_matrix.h
@@ -404,6 +404,9 @@ public:
    * This function creates a matrix called "submatrix" which is defined
    * by the row and column indices given in the "rows" and "cols" entries.
    * Currently this operation is only defined for the PetscMatrix type.
+   * Note: The \p rows and \p cols vectors need to be sorted;
+   *       Use the nosort version below if \p rows and \p cols vectors are not sorted;
+   *       The \p rows and \p cols only contain indices that are owned by this processor.
    */
   virtual void create_submatrix(SparseMatrix<T> & submatrix,
                                 const std::vector<numeric_index_type> & rows,
@@ -413,6 +416,21 @@ public:
                          rows,
                          cols,
                          false); // false means DO NOT REUSE submatrix
+  }
+
+  /**
+   * Similar to the above function, this function creates a \p
+   * submatrix which is defined by the indices given in the \p rows
+   * and \p cols vectors.
+   * Note: Both \p rows and \p cols can be unsorted;
+   *       Use the above function for better efficiency if your indices are sorted;
+   *       \p rows and \p cols can contain indices that are owned by other processors.
+   */
+  virtual void create_submatrix_nosort(SparseMatrix<T> & /*submatrix*/,
+                                         const std::vector<numeric_index_type> & /*rows*/,
+                                         const std::vector<numeric_index_type> & /*cols*/) const
+  {
+    libmesh_not_implemented();
   }
 
   /**
--- a/src/numerics/petsc_matrix.C
+++ b/src/numerics/petsc_matrix.C
@@ -925,6 +925,78 @@ void PetscMatrix<T>::_get_submatrix(SparseMatrix<T> & submatrix,
   petsc_submatrix->close();
 }
 
+template <typename T>
+void PetscMatrix<T>::create_submatrix_nosort(SparseMatrix<T> & submatrix,
+                                             const std::vector<numeric_index_type> & rows,
+                                             const std::vector<numeric_index_type> & cols) const
+{
+  if (!this->closed())
+    {
+      libmesh_deprecated();
+      libmesh_warning("The matrix must be assembled before calling PetscMatrix::create_submatrix_nosort().\n"
+                      "Please update your code, as this warning will become an error in a future release.");
+      const_cast<PetscMatrix<T> *>(this)->close();
+    }
+
+  // Make sure the SparseMatrix passed in is really a PetscMatrix
+  PetscMatrix<T> * petsc_submatrix = cast_ptr<PetscMatrix<T> *>(&submatrix);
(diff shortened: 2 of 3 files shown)


include/numerics/sparse_matrix.h
```diff
@@ -415,6 +415,23 @@ class SparseMatrix : public ReferenceCountedObject<SparseMatrix<T>>,
...
+   * and \p cols vectors.
+   * Note: Both \p rows and \p cols can be unsorted;
+           Use the above function for better efficiency if your indices are sorted;
+   *       \p rows and \p cols contain indices that are owned by all processors.
+   */
+  virtual void create_submatrix_nosort(SparseMatrix<T> & submatrix,
+                                         const std::vector<numeric_index_type> & rows,
+                                         const std::vector<numeric_index_type> & cols) const
+  {
+    this->_get_submatrix_nosort(submatrix,
```
jwpeterson, line 430: Hmm... I can't remember currently why we did this "public function calling a protected virtual" design previously, it may have been something to do with backwards compatibility or something... I see no reason to propagate this strange idiom unnecessarily, `create_submatrix_nosort()` can just be added as a virtual `libmesh_not_implemented()` function at the `SparseMatrix` level without this additional indirection.

include/numerics/sparse_matrix.h
```diff
@@ -415,6 +415,23 @@ class SparseMatrix : public ReferenceCountedObject<SparseMatrix<T>>,
                          false); // false means DO NOT REUSE submatrix
   }
 
+  /**
+   * Similar to the above function, this function creates a \p
+   * submatrix which is defined by the indices given in the \p rows
+   * and \p cols vectors.
+   * Note: Both \p rows and \p cols can be unsorted;
```
jwpeterson, line 422: Would you mind updating the docs for the original `create_submatrix()` function to note that the rows and columns must be sorted in order for the routine to work correctly? There's apparenty no indication of this requirement at the moment. And I guess it means sorted both by rows and by column indices within each row? It would be great to clarify that as well...

src/numerics/petsc_matrix.C
```diff
@@ -925,6 +925,81 @@ void PetscMatrix<T>::_get_submatrix(SparseMatrix<T> & submatrix,
...
+      LIBMESH_CHKERR(ierr);
+      // clear data for this row
+      sub_cols.clear();
+      sub_vals.clear();
+    }
+  }
+  ierr = MatAssemblyBegin(petsc_submatrix->_mat,MAT_FINAL_ASSEMBLY);
+  LIBMESH_CHKERR(ierr);
+  ierr = MatAssemblyEnd(petsc_submatrix->_mat,MAT_FINAL_ASSEMBLY);
+  LIBMESH_CHKERR(ierr);
```
jwpeterson, line 998: I think these 4 lines can be replaced by the recently-added macro:
```
MatAssemblyBeginEnd(petsc_submatrix->comm(), petsc_submatrix->_mat, MAT_FINAL_ASSEMBLY);
```

jwpeterson (commented):
Looks OK to me provided it doesn't conflict with the related changes in #2836.

include/numerics/sparse_matrix.h
dewenyushu: Thanks @jwpeterson . I agree and I will make the change.

include/numerics/sparse_matrix.h
dewenyushu: This is a good point. I will make the change to the docstring to avoid potential confusions in the future. Yes, both row and column indices need to be sorted in order for it to work properly.

src/numerics/petsc_matrix.C
dewenyushu: That is right. I will make the change. Thanks!

src/numerics/petsc_matrix.C
```diff
@@ -1197,6 +1268,10 @@ void PetscMatrix<T>::matrix_matrix_mult (SparseMatrix<T> & X_in, SparseMatrix<T>
 
   ierr = MatMatMult(_mat, X->_mat, MAT_INITIAL_MATRIX, PETSC_DEFAULT, &Y->_mat);
   LIBMESH_CHKERR(ierr);
+
+  // make sure Y is initialized and close it
+  Y->_is_initialized = true;
+  Y->close();
```
roystgnr, line 1274: The trouble with throwing close() calls everywhere is that each one of them is a parallel communication, right?

This seems like a good place to do it (I'm not even sure what you can follow up a MatMatMult with if you don't close it in between), but at some point we really ought to get more rigorous about documenting which of our methods follow up with a close() and which don't.

roystgnr (approved):

src/numerics/petsc_matrix.C
dewenyushu: Thanks @roystgnr . It seems like removing the `Y->close()` is better here, according to @fdkong (https://github.com/libMesh/libmesh/pull/2836#discussion_r560400499).  I tried it and found no issue. (I accidentally made duplicated changes here and in #2836  and I have removed the change in the other PR. I will update this one soon)

I agree that we should have better documentation about when close() is necessary or not.

dewenyushu:
@jwpeterson I suppose this PR is ready to go. Please let me know if there are further comments for its improvement :)

jwpeterson (approved):

jwpeterson:
Thanks @dewenyushu!

PR #2839, opened 2021-01-21
Use Utility::enum_to_string() to improve some error messages.

jwpeterson:
We still have many places where you can get cryptic numbers in error messages, e.g.
```
ERROR: This is not a linear element: type=9
```
which is annoying for normal users as well as developers who haven't memorized all the enumerations. This PR fixes many of them with `Utility::enum_to_string()`, but there might still be a few others lurking.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -15,10 +15,8 @@
 // License along with this library; if not, write to the Free Software
 // Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
-// Local includes
-#include "libmesh/dof_map.h"
-
 // libMesh includes
+#include "libmesh/dof_map.h"
 #include "libmesh/boundary_info.h" // needed for dirichlet constraints
 #include "libmesh/dense_matrix.h"
 #include "libmesh/dense_vector.h"
@@ -50,6 +48,7 @@
 #include "libmesh/system.h" // needed by enforce_constraints_exactly()
 #include "libmesh/tensor_tools.h"
 #include "libmesh/threads.h"
+#include "libmesh/enum_to_string.h"
 
 // TIMPI includes
 #include "timpi/parallel_implementation.h"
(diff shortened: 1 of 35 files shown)


src/base/dof_map_constraints.C
```diff
@@ -15,10 +15,8 @@
 // License along with this library; if not, write to the Free Software
 // Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
-// Local includes
-#include "libmesh/dof_map.h"
-
 // libMesh includes
+#include "libmesh/dof_map.h"
```
roystgnr, line 19: Let's avoid changes like this in the future?  I like our idiom of "the header with our definitions comes before other libMesh headers", since that makes it harder to accidentally write dependencies into a header file which get satisfied by source files' order-of-inclusion rather than by the header's own includes ... but if we just put such headers at the top of the libMesh section rather than in their own section, I'm sure I'll eventually end up breaking that idiom in subsequent PRs by alphabetizing without thinking about it.

src/mesh/mesh_triangle_interface.C
```diff
@@ -20,9 +20,7 @@
 
 #ifdef LIBMESH_HAVE_TRIANGLE
 
-// C/C++ includes
-#include <sstream>
-
+// llibmesh includes
```
roystgnr, line 23: llibmesh?

roystgnr:
Nice improvement.  Did you just notice the unhelpful errors yourself, or did this already bite a user?

src/mesh/mesh_triangle_interface.C
jwpeterson: Nice catch, fixed in 87a981396903dca0148f4ca4d6b9b9abf2040805

src/base/dof_map_constraints.C
jwpeterson: OK, my bad. I was thinking that we had possibly been inconsistently using "// local includes" and "// libmesh includes" in various places and was trying to make it consistent, but I see the idea now. Fixed in 071e2a4119ef01f65240a5d82c1501d215e2dce5

jwpeterson:
> Nice improvement. Did you just notice the unhelpful errors yourself, or did this already bite a user?

Yeah we had an error message from `all_second_order()` talking about "element type 10" which I thought could be improved, so I fixed several others at the same time.

PR #2842, opened 2021-01-26
C++14, C++17 detection options

roystgnr:
This pulls in a newer autoconf archive C++ standard detection macro, and gives users control of what to try to detect via our configure options, as described in points 2 and 3 of #2635.  The default behavior is --with-cxx-std-min=2011 --with-cxx-std-max=2017, but these can be overridden, individually or both at once via e.g. --with-cxx-std=2014.  Our configure now tests for the highest level of C++ support it can find (up to a max year of cxx-std-max), adding flags if necessary, but screams and dies if it can't find any support for C++ of year cxx-std-min.  This way users of libMesh codes that have higher C++ requirements than our minimal C++11 can get those requirements tested for immediately.

Users can still set CXX or CXXFLAGS directly (or use bleeding edge compilers) to get C++20 or future version support, and as long as whatever they request also passes the tests for the cxx-std-max version, the higher version will be respected.  Hopefully someone will add C++20 support to the autoconf archive macro soon and we can incorporate that for more user friendliness.  Hopefully that someone will be a person who hates m4 less than I do.

The behavior of allowing compiler defaults (or user flags as set directly) to override cxx-std-max, as above, is debateable.  If I get much pushback on it, I could be talked into eventually instead implementing "max means max allowed, not max tested" instead, so that --with-cxx-std-max=2017 actually tries to prevent use of a C++20 or higher compiler unless it can find a flag to downgrade that support, rather than merely skipping tests for C++20 and higher.  But the detection macros don't make that easy so I'd rather save that change for later even if it's desired.


roystgnr:
Oh, this is interesting.  With the default behavior, we're inserting -std=c++17 in cases where we weren't previously, and that means we're hitting some code that was deprecated in earlier versions and removed in C++17.  I'll fix that in our codes, but there's no way to be sure user codes don't have the same problem, and that definitely calls into question my default max setting in this new PR.

lindsayad:
I think we're going to have to update our `Eigen` contrib because I'm getting this error with c++17 support:
```
/Users/lindad/projects/moose/scripts/../libmesh/installed/include/Eigen/src/Core/Functors.h:973:28: error: no member named 'binder2nd' in namespace 'std'
```
This was indeed deprecated in C++11

lindsayad:
Looks like they fixed this up in 2015:
```
(base) lindad@localhost:~/eigen(master)$ git log -S "std::bind2nd"
commit 494fa991c36dffdbd02a0d97e48e0b297efa9dc3
Author: Christoph Hertzberg <chtz@informatik.uni-bremen.de>
Date:   Thu May 7 17:28:40 2015 +0200

    bug #872: Avoid deprecated binder1st/binder2nd usage by providing custom functors for comparison operators
```

jwpeterson:
Eigen 3.2.9, which was released on July 18, 2016. I guess it doesn't have that fix, but that is crazy old so it would be good to update. We didn't really do anything special to update last time (28e9c08defb866766fc03e4e32a86fe2716e3a5d) just copy in the new files and update the symbolic link.

lindsayad:
Yea I'm working on that now

lindsayad:
Weird that a commit from May 2015 wasn't included in a July 2016 release. Also weird that the GCC builds here with C++17 didn't complain about this

lindsayad:
It's also unfortunate...The error that I pasted in was a with a `Functors.h` include header. That header appears to no longer exist in eigen-3.3.9. Sounds like we may need to introduce some preprocessing directives to handle this stuff

jwpeterson:
Yeah, we don't include Functors.h so if it's gone, I think we are fine. I think you will need to update contrib/Eigen/Makefile.am. I'm guessing I just pasted the output of "ls" into that file previously.

lindsayad:
Ugh: https://gitlab.com/libeigen/eigen/-/issues/2011 They actually explicitly mention libmesh in that issue.

jwpeterson:
So you're seeing the same error as Thierry Thomas from that thread?

lindsayad:
No I saw the title of the issue with 3.3.8:
```
Eigen::eigen_assert_exception missing from includes of src/Core/products/Parallelizer.h and src/Core/CommaInitializer.h
```
I'm testing with 3.3.9 now...

lindsayad:
3.3.9 works great. So with c++17 with the fixes from MetaPhysicL and 3.3.9 Eigen I am able to build all the way through MOOSE test and all MOOSE tests pass.

roystgnr:
Sounds like the last problem remaining here is going to be the one in https://civet.inl.gov/job/668530/ - somehow with older clang versions it seems that asking for C++14 support *breaks* our test of a C++11 feature.  I haven't looked into the details, maybe there's something wrong with our test, but most likely this may be an argument for simply not defaulting to asking for anything past C++11 unless the user requests it.

lindsayad:
That failure would be fixed with libmesh/autoconf-submodule#5

roystgnr:
@lindsayad, were you able to replicate the "Minimum clang version" failure?  I downloaded clang++ 3.5.2 (which I assumed would be close enough to 3.5.1 to replicate with), and it's not failing for me.  ax_cxx_compile_stdcxx.m4 discovers that -std=gnu++14 works, our cxx11.m4 uses that flag via $CXX (I verified this using a wrapper script as my compiler), and without $switch overriding it, that passes our C++11 tests just fine.

I don't suppose a Civet "download as tarball" link will include a failed config.log in that tarball?

lindsayad:
No it won't. It's just the console output (AFAIK). Are we sure that `libmesh_CXXFLAGS` isn't doing something nefarious?

lindsayad:
> No it won't. It's just the console output (AFAIK). Are we sure that `libmesh_CXXFLAGS` isn't doing something nefarious?

I don't see how it would. I don't see population of `libmesh_CXXFLAGS` anywhere in `civet_recipes`

lindsayad:
Why did you re-push those three commits? Was something missing?

roystgnr:
Somehow it looked like my MetaPhysicL update hadn't updated MetaPhysicL.

roystgnr:
Which should fix Test Mac, but isn't related to the minimum-clang failures.

lindsayad:
@milljm @loganharbour is there a way for us to dig out the `config.log` from the Mac boxes?

lindsayad:
Wait just kidding these are linux-min...

roystgnr:
> Which should fix Test Mac

By which I mean that this *subsequent* force push should fix Test Mac.  I really need to add more sanity checking to my makerelease.sh script.

roystgnr:
I think I'm happy with this now.  Only two things left:

1. Clean up history.  I'm going to keep a lot of those commits separate, since early stages work for most compilers and each fix for weirder cases probably deserves its own explanatory comment, but I'll improve the comments and I'll merge all the bootstrap commits into one final one.

2. Decide whether to keep the C++14/17 switch requests in.  I like asking for the highest standards we know how to ask for, even if we're not going to use anything above C++11 internally, but this is a behavior change, and it might be a hassle for people whose user code isn't C++17 compatible, just like MetaPhysicL previously wasn't, who are using compilers which support C++17 but don't default to C++17.  Conversely, if we just ask for C++11 by default, that won't actually break any existing users who require higher standards, because we'll still get support for those standards the way we did before, via their compiler defaults or their $CXX or $libmesh_CXXFLAGS settings, we never request a downgrade.

I think what pushes me to vote "keep them in" is that we're considering requiring some C++14 or even C++17 support in the not-too-distant future, and if that's going to break anybody then I'd rather them find out ASAP, while we still have the ability to easily revert back to C++11 with a configure option.  But I could be talked out of that if others disagree.

jwpeterson:
I admit to not following along closely with the development of this PR (+18,541 lines?!) but is your question whether to make the default configure options: `--with-cxx-std-min=2011 --with-cxx-std-max=2017`, with then translates into passing `-std=c++17` to the compiler provided that it passes all the tests for that?

jwpeterson:
Oh, I just realized that GitHub lets you filter by file extension, so it's possible to turn off the Makefile.in file changes. That helps a lot.

jwpeterson:
`m4/cxx11.m4 -> m4/acsm_cxx_tests.m4`

This looks like just a name change for the time being, but the name implies that this file will eventually be moved into the autoconf-submodule, is that right?

roystgnr:
> (+18,541 lines?!)

I was afraid someone would call me on that.  Some of that is Makefile.in changes, but there's a lot from the generated configure script - after fighting with interactions between m4 and for loops I got frustrated enough to do some hand-unrolling, which added a few lines to the m4 macro but which bloats the final script pretty badly.  I can try and fix it, if you want.  It might be easier to do so now that I'm starting from a known-working state and I've already hacked together some better configure-debugging tools (e.g. compiler wrapper scripts that simulate selective failures).  But I was really hoping to just be done with this for a while; I have more urgent things on my plate.

> is your question whether to make the default configure options: --with-cxx-std-min=2011 --with-cxx-std-max=2017, which then translates into passing -std=c++17 to the compiler provided that it passes all the tests for that?

Almost.  We pass -std=c++17 to the compiler provided that it passes all the C++17 tests with that and that it *doesn't* pass the C++17 tests without that.  So with the settings as-is, we aren't downgrading anybody who's using a C++20-default compiler, but we are upgrading anybody who's using a C++14-default C++17-capable compiler and who doesn't specify e.g. --with-cxx-std=2014 to opt out of the upgrade.

roystgnr:
> This looks like just a name change for the time being, but the name implies that this file will eventually be moved into the autoconf-submodule, is that right?

Not quite.  The name change is because it's a callback *from* the autoconf submodule, so we want to use their namespace.  I needed the ACSM macro to be able to call libMesh tests from in between its own tests, but I wanted to be able to allow other packages using the ACSM macro (starting with MetaPhysicL for now) to specify fewer tests (appropriate in that case since it's effectively a header-only library) or more tests (in case someone really needs particular C++14 or higher features that we don't).

jwpeterson:
> We pass -std=c++17 to the compiler provided that it passes all the C++17 tests with that and that it doesn't pass the C++17 tests without that.

OK, that makes sense but it's very subtle to really understand the difference IMO. Like, I (think) I understand it right now, but I might not in a month or two. I guess what we don't support (and haven't ever supported) is someone who says, "I want to explicitly compile my code with -std=cxxFF", that is, there's no way to choose an explicit standard at configure time, correct?

jwpeterson:
> after fighting with interactions between m4 and for loops

Oh yeah, you had asked me something about that but I thought we decided that since there was no `AS_FOR` you would just write POSIX-style for loops in m4 files and it would work... is that not the case?

> bloats the final script pretty badly

Our current configure is already over 45k lines, I think the worrying about bloating it ship has long ago sailed... there isn't some hard upper bound on the size of a /bin/sh script that we need to worry about hitting, right?

roystgnr:
>  I guess what we don't support (and haven't ever supported) is someone who says, "I want to explicitly compile my code with -std=cxxFF", that is, there's no way to choose an explicit standard at configure time, correct?

Actually, we still support that, in the same way we used to: setting the flag via CXX or CXXFLAGS.  That's why I reverted that change the autoconf-archive made to the detection macro, so we begin testing with a "plain" environment and only try adding flags if the plain tests failed, because with user intervention "plain" can mean "manually added a flag that the macro didn't know existed".

roystgnr:
> I thought we decided that since there was no AS_FOR you would just write POSIX-style for loops in m4 files and it would work... is that not the case?

I think it might have worked, because after I did the unrolling and figured out what I'd done wrong it didn't look like my bugs were related to the for loop ... it's just that by that point I was so sick of m4 that I didn't want to worry about optimization while I was still not 100% of the way to correctness.

> Our current configure is already over 45k lines, I think the worrying about bloating it ship has long ago sailed...

I suppose.  It's still a trivial size of our total download.  I'm bothered because the bloat is embarrassing, not because of any performance concerns.

> there isn't some hard upper bound on the size of a /bin/sh script that we need to worry about hitting, right?

There is a maximum length for a single shell command, which can be surprisingly short on old Unixes.  But I don't think I've ever heard of a maximum length for command blocks (for, if, etc) and the maximum length for scripts is IIRC just set by filesystem limitations, 2GB even on very old systems.

jwpeterson (approved):

roystgnr:
@jwpeterson noticed a huge jump in fem_system_ex5 runtime with #2845 - I'm going to hold off on merging this until that's been straightened out, just in case we have to revert.

lindsayad:
:cry:

jwpeterson:
For the complex-enabled testing (which uses Eigen Sparse solvers because PETSc is disabled in that config)
[Before Eigen PR](https://civet.inl.gov/pr/12601/): 1 hour 50 minutes 
[After Eigen PR](https://civet.inl.gov/pr/12611/): 5 hours 53 minutes

roystgnr:
Let's move discussion to that PR?

roystgnr:
Looks like #2851 straightened things out orthogonal to the changes and dependencies here, so we're good to go!

PR #2843, opened 2021-01-27
The SHAPE and SKEW bounds for Quad elements should have similar bounds.

jwpeterson:
After updating these metrics in 60487b58446, I should have also updated the bounds. Note that Knupp does not give explicit bounds on these metrics for a "good" quality element in his paper, but both have an upper bound of 1, so the previous values for SKEW were just wrong.


jwpeterson:
There's a known issue with CIVET recipes at the moment which is unrelated to this small change, so I'm merging.

PR #2844, opened 2021-01-28
RBConstruction: Handle exceptions thrown during assembly.

jwpeterson:
This fixes an issue we had where it was difficult to handle exceptions thrown by the `RBConstruction::add_scaled_matrix_and_vector()` function in parallel, since previously the code could throw on some processors but not others. Now it either throws on all procs or none, which makes handling the exception easier for the caller.


jwpeterson:
This works for me in my local testing, and we don't have any CI that would really exercise it, so I'm going to go ahead and merge.

PR #2846, opened 2021-02-01
Only *non-empty* constraint_rows is experimental

### LINKED ISSUE #2841: Experimental warning for all MeshBase users
We seem to have picked up a warning recently:
```
*** Warning, This code is untested, experimental, or likely to see future API changes: ./include/libmesh/mesh_base.h, line 1667, compiled Jan 15 2021 at 09:23:22 ***
```
In my version of libmesh, that line corresponds to this:
```
  const constraint_rows_type & get_constraint_rows() const
  { libmesh_experimental(); return _constraint_rows; }
```
but I don't explicitly call this anywhere in my code... so is it really experimental if I can't avoid it easily?

@roystgnr

roystgnr:
Especially with distributed meshes I still feel like IGA (and hopefully
in the future other) uses of this new constraint_rows data structure are
tricky enough to warn users about, but we have more than enough coverage
of the empty-constraint_rows case to be certain I haven't regressed
anything there.

Fixes #2841

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -1661,10 +1661,10 @@ public:
     constraint_rows_type;
 
   constraint_rows_type & get_constraint_rows()
-  { libmesh_experimental(); return _constraint_rows; }
+  { return _constraint_rows; }
 
   const constraint_rows_type & get_constraint_rows() const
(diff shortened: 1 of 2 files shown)


jwpeterson:
Thanks!

PR #2847, opened 2021-02-02
Turn off gdb backtraces by default

lindsayad:
The user will now have to opt-in to gdb backtraces instead of opting out
with `--without-gdb-command`. There are many times that I forget to
configure with `--without-gdb-command` and then I sadly wait for a
minute to get my prompt back or switch to another terminal to kill the
process. There was some general support for this
change on our Moose developers libmesh slack channel

### DIFF
--- a/configure
+++ b/configure
@@ -31889,7 +31889,7 @@ $as_echo "---------------------------------------------" >&6; }
 if test "${with_gdb_command+set}" = set; then :
   withval=$with_gdb_command; gdb_command="$withval"
 else
-  gdb_command="gdb"
+  gdb_command="no"
 fi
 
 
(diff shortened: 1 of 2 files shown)


roystgnr:
I'd forgotten this, but you can use "--gdb=no" on the libMesh command line even if you've enabled it at compile time.

The reverse works, too, though, so I'm still happy with this PR; if someone does change their mind and want backtraces they can run with --gdb=gdb.

jwpeterson:
Since the GDB backtrace approach relies on attaching to a running process with `-p`, I think for security reasons, on newer systems it no longer even works when GCC toolchains are being used. Definitely should not be on by default.

roystgnr:
You're right about the security reasons - I forgot, that that bit me once.  And I think the resolutions (prctl() calls or /proc/sys changes, on Linux) are all very system-specific, not something I really want to add to libMesh code or maintain.

PR #2849, opened 2021-02-03
MeshFunction updates

jwpeterson:
Previously `MeshFunction::set_point_locator_tolerance()` only set the tolerance used in `PointLocator` `Elem::close_to_point()` checks, but now it will set the tolerance used in `Elem::contains_point()` and `BoundingBox` checks done within the `PointLocator` as well. This is more consistent with the user's likely intention in calling the function in the first place, and helps prevent situations in which the `PointLocator`'s performance degrades dramatically as it falls back on doing many linear searches. I think what happened here is probably that additional tolerances were added on the `PointLocator` side, but the `MeshFunction` API was not updated at that time.

### DIFF
--- a/include/mesh/mesh_function.h
+++ b/include/mesh/mesh_function.h
@@ -101,15 +101,17 @@ public:
   ~MeshFunction ();
 
   /**
-   * Override the FunctionBase::init() member function by calling our
-   * own and specifying the Trees::NODES method.  specifies the method
-   * to use when building a \p PointLocator
+   * Override the FunctionBase::init() member function.
    */
-  virtual void init () override { this->init(Trees::NODES); }
+  virtual void init () override;
 
   /**
    * The actual initialization process.  Takes an optional argument which
    * specifies the method to use when building a \p PointLocator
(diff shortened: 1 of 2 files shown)


include/mesh/mesh_function.h
```diff
@@ -105,11 +105,15 @@ class MeshFunction : public FunctionBase<Number>,
    * own and specifying the Trees::NODES method.  specifies the method
    * to use when building a \p PointLocator
    */
-  virtual void init () override { this->init(Trees::NODES); }
+  virtual void init () override;
```
roystgnr, line 108: The comment here is atavistic now, isn't it?

roystgnr (approved):

include/mesh/mesh_function.h
jwpeterson: Yes, I'll fix that.

jwpeterson:
The machine that runs the Test Mac testing seems to be having some issues, so yeah, don't worry about that little guy.

PR #2850, opened 2021-02-03
More complete BoundaryInfo handling around mesh stitching

lindsayad:
This PR:
- ensures that we have nodeset/sideset/edgeset id to name mapping correct after stitching. We were not copying over the id to name information from `other_mesh` previously
- erases all `BoundaryInfo` information for the stitched boundary ids upon request, as opposed to previously where we were only erasing sideset information
- adds nodes into nodesets in `MeshTools::Generation::build_cube`


src/mesh/replicated_mesh.C
```diff
@@ -1319,6 +1319,23 @@ void ReplicatedMesh::stitching_helper (const ReplicatedMesh * other_mesh,
                                std::get<1>(t),
                                std::get<2>(t));
 
+      const auto & other_ns_id_to_name = other_boundary.get_nodeset_name_map();
+      auto & ns_id_to_name = boundary.set_nodeset_name_map();
+
+      for (const auto & other_pr : other_ns_id_to_name)
+        ns_id_to_name[other_pr.first] = other_pr.second;
```
roystgnr, line 1326: Why not just ```ns_id_to_name.insert(other_ns_id_to_name.begin(), other_ns_id_to_name.end()```?

lindsayad: :+1:

lindsayad:
Looks like there is some MOOSE compatibility I will have to work on first (not surprising)

lindsayad:
This is ready for review

src/mesh/boundary_info.C
```diff
@@ -212,6 +212,11 @@ void BoundaryInfo::regenerate_id_sets()
       _boundary_ids.insert(id);
       _shellface_boundary_ids.insert(id);
     }
+
+  // Handle global data
+  _communicator.set_union(_ss_id_to_name);
+  _communicator.set_union(_ns_id_to_name);
+  _communicator.set_union(_es_id_to_name);
```
roystgnr, line 219: Could we make this a fixup on the previous commit?  The state in between is broken in a way that worries me, letting those maps get out of sync.

include/mesh/boundary_info.h
```diff
@@ -974,6 +982,13 @@ class BoundaryInfo : public ParallelObject
   std::map<boundary_id_type, std::string> _es_id_to_name;
 };
 
+template <typename ModifyFunctor>
+void
+BoundaryInfo::modify_side(ModifyFunctor & modify_functor)
+{
+  modify_functor(_boundary_side_id);
+}
```
roystgnr, line 1006: This is kind of a weird API, isn't it?  It's no better encapsulated than a non-const getter function for _boundary_side_id would be, but it's more complicated syntax.

lindsayad: Haha I was anticipating a response like this. I have no counter-argument. Are you open to a non-const getter? Otherwise I don't know how to do what I want to do in the stitch operation without making things pretty complicated

lindsayad: Or actually I could add a pretty specific `BoundaryInfo::clear_stitched_boundary_ids(boundary_id_type bid_1, boundary_id_type bid_2)`

lindsayad: Now that I've said it, the new method sounds most attractive to me...

roystgnr: I'd be reasonably happy with that, except either name it ```clear_stitched_boundary_side_ids``` or have it work on node/edge/shellface ids too?

But sadly, I'd even be okay with the non-const getter right now.  The const getter we already have is enough to prevent us from making data structure changes without issues backwards compatibility.  I should have fought the camel's nose at that point, at this point most of the damage is already done.

src/mesh/boundary_info.C
lindsayad: done

include/mesh/boundary_info.h
lindsayad: Even though strict encapsulation crimes have already been committed, I elected to go the more encapsulated route and added the `BoundaryInfo` method

jwpeterson:
Is the TIMPI submodule update needed for this PR? I guess that was the discussion about communicating strings that I saw...

lindsayad:
Yes. Otherwise I cannot do the `set_union` with the id to name boundary info maps.

roystgnr:
Man, I hate re-reviewing after a force push.  Is the only thing you changed that BoundaryInfo modifier API and it's use?  If so then I think this is good to merge.  If not then point me to whatever else changed since my last read?

lindsayad:
> Man, I hate re-reviewing after a force push.

Noted and understood for future.

>  Is the only thing you changed that BoundaryInfo modifier API and it's use?

Yea deleted the basically encapsulation violating `BoundaryInfo::modify_side` and added `BoundaryInfo::clear_stitched_boundary_side_ids` instead. Essentially https://github.com/libMesh/libmesh/pull/2850/commits/83ab3a78ed23bee9740729e697548b0df6aba69e is the change

roystgnr:
Well, the trouble is that I *like* the force-pushed history cleanup.  I think the ideal workflow is probably "tack new commits on the end to respond to reviews, then once everybody's happy with the final state of the branch do a single interactive rebase that's nothing but fixups and squashes", but that's not always practical.

This looks good to me now; give @jwpeterson a chance to finish looking over the final state too before we merge, but if he's happy or doesn't have time then we can merge tomorrow.

lindsayad:
I think that your ideal workflow is not too difficult to realize as long as
the developer is making good use of `git commit --fixup`. I'll do my best
to adhere to that philosophy in the future, especially as that is the
workflow that *I* most appreciate when I'm reviewing.

roystgnr:
Either there needs to be a "+2" emoji on Github, or there needs to be a way to attach multiple emoji to different parts of a comment.  I didn't actually know that ```--fixup``` existed; I've been writing "fixup Foo" and "fixup Bar" comments and rebasing manually for years.

include/mesh/boundary_info.h
```diff
@@ -912,6 +923,8 @@ class BoundaryInfo : public ParallelObject
    * See _side_boundary_ids, _edge_boundary_ids, _node_boundary_ids, and
    * _shellface_boundary_ids for sets containing IDs for only sides, edges,
    * nodes, and shell faces, respectively.
+   *
+   * This data is semilocal in nature
```
jwpeterson, line 927: I lazily didn't check, but is the concept of "semilocal data" rigorously defined somewhere?

src/mesh/mesh_generation.C
```diff
@@ -956,9 +1014,25 @@ void MeshTools::Generation::build_cube(UnstructuredMesh & mesh,
...
-                    mesh.add_point(Point(static_cast<Real>(i)/static_cast<Real>(2*nx),
-                                         static_cast<Real>(j)/static_cast<Real>(2*ny),
-                                         static_cast<Real>(k)/static_cast<Real>(2*nz)), node_id++);
+                  {
+                    const Node * const node =
+                        mesh.add_point(Point(static_cast<Real>(i) / static_cast<Real>(2 * nx),
+                                             static_cast<Real>(j) / static_cast<Real>(2 * ny),
+                                             static_cast<Real>(k) / static_cast<Real>(2 * nz)),
+                                       node_id++);
+                    if (k == 0)
```
jwpeterson, line 1023: I guess it's done now, but any reason to not just call `BoundaryInfo::build_node_list_from_side_list()` some place at the end of `build_cube()`?

src/mesh/boundary_info.C
```diff
@@ -2601,5 +2622,76 @@ void BoundaryInfo::_find_id_maps(const std::set<boundary_id_type> & requested_bo
...
 }
 
+void BoundaryInfo::clear_stitched_boundary_side_ids (const boundary_id_type sideset_id,
+                                                     const boundary_id_type other_sideset_id,
+                                                     const bool clear_nodeset_data)
+{
+  auto end_it = _boundary_side_id.end();
+  auto it = _boundary_side_id.begin();
+
+  auto predicate =
```
jwpeterson, line 2632: Maybe a comment about what this `predicate()` lambda does? Either here or when it's used. It is sufficiently complicated looking to warrant that IMO.

jwpeterson (approved):
Thanks for adding unit tests!

jwpeterson:
> I've been writing "fixup Foo" and "fixup Bar" comments and rebasing manually for years.

Yeah in addition to `git commit --fixup <hash>`, you would need to `git rebase -i --autosquash` to get the full benefit.

YaqiWang:
This fixup and autosquash thing sounds pretty handy for some cases.

include/mesh/boundary_info.h
lindsayad: I actually don't think so, and I don't know where to put it. Instead I could say "attached to this process's local or ghosted elements"

src/mesh/mesh_generation.C
lindsayad: I talked about this with @roystgnr in the beginning of working on this PR and he voiced his distaste for that method because it mixes two concepts that aren't necessarily related.

src/mesh/boundary_info.C
lindsayad: comment added in https://github.com/libMesh/libmesh/pull/2850/commits/5a32f242b27eb4a62092d22ede56992b98f8cac3

src/mesh/mesh_generation.C
roystgnr: Sorry I was unclear; I just dislike using that method to modify an arbitrary mesh; it would have made decent sense to use it in this case, to save you some work.

Although I do like the way you've done it here a tiny bit better; I've always imagined that some day we might modify some of our mesh reading and mesh generation capabilities so that, instead of clearing (or doing random garbage...) existing mesh data, they could merge new data in-place into an existing mesh.  Avoiding the shortcut here means that if we ever decide to do that then there's one less thing that will need to be fixed to make it work.

roystgnr:
Looks good to me.  Ready to clean up the history and merge?

lindsayad:
:+1:

jwpeterson (approved):

PR #2851, opened 2021-02-05
Re-reserve non-zero pattern after Eigen::SparseMatrix::setZero

lindsayad:
`Eigen::SparseMatrix::setZero` clears the entire non-zero pattern, so we have to
re-reserve after zeroing in order to not have very expensive recalls to
`Eigen::SparseMatrix::reserveInnerVectors`. This solves the problem reported in
#2845 .

Previous graph after merge of #2845:
image
Graph after this fix:
re-reserve

### DIFF
--- a/src/numerics/eigen_sparse_matrix.C
+++ b/src/numerics/eigen_sparse_matrix.C
@@ -179,7 +179,15 @@ void EigenSparseMatrix<T>::clear ()
 template <typename T>
 void EigenSparseMatrix<T>::zero ()
 {
+  // This doesn't just zero, it clears the entire non-zero structure!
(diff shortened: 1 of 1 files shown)


jwpeterson:
Nice catch! I kinda wish they had something similar to PETSc where it would crash on new non-zero allocation during insert instead of just running really slow.

lindsayad:
Totally agree

jwpeterson:
Test only takes 23 minutes now instead of the original time of 2 hours! Sweet!

roystgnr (approved):

lindsayad:
Nice! Maybe we should replace PETSc :-)

PR #2855, opened 2021-02-08
Fix deprecated use of libmesh_cast_ptr

jwpeterson:
The InterMeshProjection PR introduced some deprecated code.

### DIFF
--- a/src/systems/inter_mesh_projection.C
+++ b/src/systems/inter_mesh_projection.C
@@ -18,8 +18,6 @@
 // Local includes
 #include "libmesh/inter_mesh_projection.h"
 
-// C++ includes
-
 namespace libMesh
 {
     GradientMeshFunction::GradientMeshFunction(MeshFunction * _mesh_function)
@@ -30,7 +28,7 @@ namespace libMesh
     // to reassign its ownership, and then cast this released ptr as a MeshFunction * to complete
     // the assignment. Looks complicated, but other approaches to this assignment led to hard to
     // untangle Seg Faults.
-     mesh_function = std::unique_ptr<MeshFunction>(libmesh_cast_ptr<MeshFunction *>((_mesh_function->clone()).release()));
+     mesh_function = std::unique_ptr<MeshFunction>(cast_ptr<MeshFunction *>((_mesh_function->clone()).release()));
      libmesh_experimental();
     }
 


jwpeterson:
The `InterMeshProjection` code is not indented correctly, doesn't follow our naming conventions, and uses raw new/delete. I don't think this should have been merged as-is.

roystgnr:
Damn, my bad on the raw new/delete, I should have noticed that.

I didn't even think to double-check indentation.  We have that tested in MOOSE PRs as a mandatory-pass precheck, right?  Should we get it added to libMesh CI?  I don't like the mandatory-pass behavior holding up subsequent testing, but I could be talked out of that, and we could at least have it return "failed but allowed" or something to make it easy to catch and fix up in the end.

I'm a little surprised that libmesh_cast_ptr slipped by.  Do we not have --disable-deprecated turned on in any of our default recipes?

roystgnr (approved):

jwpeterson:
> Do we not have --disable-deprecated turned on in any of our default recipes?

We do not currently, but we do have a disable-deprecated CIVET recipe that can be attached on demand

roystgnr:
Should we turn on --disable-deprecated in some test by default?  The only obvious downside I see is that doing so would make it easier to get master into a CI-failing state: PR Alpha deprecates a function and replaces all old uses of it, PR Beta gets started before Alpha is merged and adds a new use of the deprecated function, both end up getting merged based on CI tests which are happy because each only tests one of the two branches' changes at a time.  But even if that ever happens it'll be easy to fix, and it won't be much worse than what's happening now, where we don't realize the need for a fix until someone manually disables deprecated code.

roystgnr:
Oh, the new/delete thing is doubly my fault.  Vikram was basing this off src/apps/projection.C, which I was using new/delete in (in 2011, in my defense!) and which I never cleaned up (this is less defensible...)

jwpeterson:
> Should we turn on --disable-deprecated in some test by default?

I would hesitate to do so since it would then require anyone making a PR that deprecates code to also immediately fix all existing uses of that deprecated code to avoid getting a red X on their PR... The current system in which some devs semi-regularly compile with `--disable-deprecated` configs seems reasonable to me.

PR #2857, opened 2021-02-08
Branch v1.6.1 testing

roystgnr:
Trying to run our Civet tests on v1.6.1.

I naturally can't get moosebuild to play with a PR in a non-default fork (see https://github.com/roystgnr/libmesh/pull/1 ); hopefully it'll still be monitoring PRs to a non-master branch.


roystgnr:
Cool, default tests running.  Presuming those pass (e.g. the recipes are really trying to merge on top of "merge_base_v1.6.0" and haven't hardcoded "master"), I'll add some of the more interesting non-default tests afterward.

roystgnr:
Looks like the merge is done correctly, and there's an actual bug!
```
../src/numerics/eigen_sparse_matrix.C: In instantiation of 'void libMesh::EigenSparseMatrix<T>::zero() [with T = double]':
../src/numerics/eigen_sparse_matrix.C:366:16:   required from here
../src/numerics/eigen_sparse_matrix.C:185:13: error: 'class libMesh::EigenSparseMatrix<double>' has no member named '_sp'
  185 |   if (this->_sp)
      |       ~~~~~~^~~
../src/numerics/eigen_sparse_matrix.C:188:58: error: 'class libMesh::EigenSparseMatrix<double>' has no member named '_sp'
  188 |     const std::vector<numeric_index_type> & n_nz = this->_sp->get_n_nz();
      |                                                    ~~~~~~^~~
Makefile:17707: recipe for target 'src/numerics/libmesh_dbg_la-eigen_sparse_matrix.lo' failed
```

I guess that bug fix depended on the sparse-matrices-without-DofMap dependence feature that I added post-v1.6.0?

jwpeterson:
> I guess that bug fix depended on the sparse-matrices-without-DofMap dependence feature that I added post-v1.6.0?

Is there any other way to reproduce @lindsayad's fix in pre-1.6.0 code? Maybe save the sparsity pattern locally before zeroing so it can be re-applied afterward? Otherwise, we can just drop that fix from the point release branch. It would be unfortunate to lose the performance improvement, but not the end of the world...

roystgnr:
> Is there any other way to reproduce @lindsayad's fix in pre-1.6.0 code?

Ironically I think I recall seeing an original commit of his in that fix PR that was pre-1.6.0 compatible, but I presume it didn't work with the pre-DofMap-projection-matrix code triggered by combining Dirichlet+IGA in a couple of our examples, so he replaced it with the current code and force pushed.  Shouldn't be too hard to replicate though, even if I'm misremembering or if he doesn't have the original commit.

roystgnr:
Let me remove the offending commit first and make sure tests pass again, then I'll try recreating it in a compatible way.

roystgnr:
> Let me remove the offending commit first and make sure tests pass again, then I'll try recreating it in a compatible way.

By which I meant "let me try recreating it first, since it seems so easy to do and I'm full of hubris, then let me try pushing with the recreated fix, then let me try digging through CI failures until I see that one of them seems to include an error message related to the recreated fix, then let me sheepishly delete the recreated commit and do what I said I was going to do in the first place".

roystgnr:
From IM discussion, after discovering that current MOOSE now depends on APIs that didn't make it into v1.6.0:

> (09:41:57 AM) John: So, slight kink in the brilliant plan of tricking CIVET into testing point releases. I guess we'll have to dig through the red X's and make sure they are all MOOSE build related
(09:41:59 AM) Roy: Well, that's the limit of testing, then.  We're definitely not adding those new APIs to a point fix release.
(09:42:37 AM) Roy: But yeah, I think if we at least check everything and make sure it passed the libMesh stages, that's much better than manual testing alone.

I'll wait until the rest of these tests die, and make sure they died at the expected point, before I force-push our next adjustments: a better recreation of that EigenSparseMatrix fix, and we're going to remove the ```MeshFunction::set_point_locator_tolerance()``` changes because, after some discussion, we worry that in what's supposed to be an extra-stable point release, the fix for a performance issue isn't worth the risk of side effects there.

jwpeterson:
Just FYI for anyone else following along, we don't expect the MOOSE testing (or even building) to pass on this branch, since CIVET pulls an up-to-date MOOSE and we are testing a libmesh from roughly last October. It's still better than creating a tag and hoping for the best IMO.

lindsayad:
> Ironically I think I recall seeing an original commit of his in that fix PR that was pre-1.6.0 compatible, but I presume it didn't work with the pre-DofMap-projection-matrix code triggered by combining Dirichlet+IGA in a couple of our examples, so he replaced it with the current code and force pushed. Shouldn't be too hard to replicate though, even if I'm misremembering or if he doesn't have the original commit.

I suspect it was never compatible. I was always relying on the `_sp` member. The commit which I force pushed over did some assertions with the `_dof_map` but then still used `_sp`. I removed the assertion with the `_dof_map` because there were unit tests (and consequently potential user code) that called `zero` before any `DofMap` or `SparsityPattern` was attached.

jwpeterson:
This seems to be passing all the testing we expect it to pass, so I'm going to close this and push a v1.6.1 tag now.

jwpeterson:
[1.6.1 Release Testing](https://civet.inl.gov/event/63525/)

jwpeterson:
Release 1.6.1 is [available now](https://github.com/libMesh/libmesh/releases/tag/v1.6.1).

PR #2858, opened 2021-02-09
Inter mesh projection fixup

vikramvgarg:
Code improvements for inter mesh projection branch which was merged earlier:

1)  Addition of MeshFunction copy constructor, which is now used instead of MeshFunction::clone() to avoid casting.
2) Raw new/delete replaced by unique_ptr.
3) Reindentation of all changed files.

Ref: https://github.com/libMesh/libmesh/pull/2855
https://github.com/libMesh/libmesh/pull/2826

### DIFF
--- a/include/mesh/mesh_function.h
+++ b/include/mesh/mesh_function.h
@@ -85,7 +85,7 @@ public:
                 const FunctionBase<Number> * master=nullptr);
 
   /**
-   * A regular copy constructor. Replicates the steps of MeshFunction::clone().
+   * A regular copy constructor.
    */
   MeshFunction (const MeshFunction & mf);
 
--- a/include/mesh/mesh_function.h
+++ b/include/mesh/mesh_function.h
@@ -85,12 +85,16 @@ public:
                 const FunctionBase<Number> * master=nullptr);
 
   /**
+   * A regular copy constructor. Replicates the steps of MeshFunction::clone().
+   */
+  MeshFunction (const MeshFunction & mf);
+
+  /**
    * Special functions.
    * - This class conains a unique_ptr so it can't be default copy constructed.
    * - This class contains const references so it can't be default copy/move assigned.
    * - The destructor is defaulted out-of-line.
    */
-  MeshFunction (const MeshFunction &) = delete;
   MeshFunction (MeshFunction &&) = default;
   MeshFunction & operator= (const MeshFunction &) = delete;
   MeshFunction & operator= (MeshFunction &&) = delete;
--- a/include/systems/inter_mesh_projection.h
+++ b/include/systems/inter_mesh_projection.h
@@ -87,7 +87,7 @@ public:
 
   virtual std::unique_ptr<FunctionBase<Gradient>> clone () const
   {
-    return libmesh_make_unique<GradientMeshFunction>(*(mesh_function.get()));
+    return libmesh_make_unique<GradientMeshFunction>(*mesh_function);
   }
 
   virtual Gradient operator() (const Point & , const Real)
--- a/include/systems/inter_mesh_projection.h
+++ b/include/systems/inter_mesh_projection.h
@@ -87,7 +87,6 @@ public:
 
   virtual std::unique_ptr<FunctionBase<Gradient>> clone () const
   {
-    //return libmesh_make_unique<GradientMeshFunction>(dynamic_cast<MeshFunction *>(mesh_function.get()));
     return libmesh_make_unique<GradientMeshFunction>(mesh_function.get());
   }
 
--- a/include/systems/inter_mesh_projection.h
+++ b/include/systems/inter_mesh_projection.h
@@ -78,7 +78,7 @@ class GradientMeshFunction : public FunctionBase<Gradient>
 {
 public:
   // Constructor
-  GradientMeshFunction(MeshFunction * _mesh_function);
+  GradientMeshFunction(const MeshFunction & _mesh_function);
 
   // Destructor
   virtual ~GradientMeshFunction () { }
@@ -87,7 +87,7 @@ public:
 
   virtual std::unique_ptr<FunctionBase<Gradient>> clone () const
   {
-    return libmesh_make_unique<GradientMeshFunction>(mesh_function.get());
+    return libmesh_make_unique<GradientMeshFunction>(*(mesh_function.get()));
   }
 
   virtual Gradient operator() (const Point & , const Real)
--- a/include/systems/inter_mesh_projection.h
+++ b/include/systems/inter_mesh_projection.h
@@ -37,71 +37,71 @@
 namespace libMesh
 {
 
-    // Forward declarations
+// Forward declarations
 
-    /**
-     * This class implements inter mesh projection, i.e. projection of
-     * vectors defined on a given mesh (from_mesh associated with from_system)
-     * to another mesh (to_mesh of to_system).
-     */
+/**
+ * This class implements inter mesh projection, i.e. projection of
+ * vectors defined on a given mesh (from_mesh associated with from_system)
+ * to another mesh (to_mesh of to_system).
+ */
 
-    class InterMeshProjection
-    {
-        public:
+class InterMeshProjection
+{
+public:
 
-        // Constructor, specifies the _from_system whose vectors will be
-        // projected onto the _to_mesh
-        InterMeshProjection(System & _from_system, System & _to_mesh);
+  // Constructor, specifies the _from_system whose vectors will be
+  // projected onto the _to_mesh
+  InterMeshProjection(System & _from_system, System & _to_mesh);
 
-        // Projects from_system vectors onto the to_mesh
-        void project_system_vectors();
+  // Projects from_system vectors onto the to_mesh
+  void project_system_vectors();
 
-        static Number fptr(const Point & p, const Parameters &, const std::string & libmesh_dbg_var(sys_name), const std::string & unknown_name);
+  static Number fptr(const Point & p, const Parameters &, const std::string & libmesh_dbg_var(sys_name), const std::string & unknown_name);
 
-        static Gradient gptr(const Point & p, const Parameters &, const std::string & libmesh_dbg_var(sys_name), const std::string & unknown_name);
+  static Gradient gptr(const Point & p, const Parameters &, const std::string & libmesh_dbg_var(sys_name), const std::string & unknown_name);
 
-        private:
+private:
 
-        // Local copy of the _from_system
-        System & from_system;
+  // Local copy of the _from_system
+  System & from_system;
 
-        // Local copy of the _to_system
-        System & to_system;
+  // Local copy of the _to_system
+  System & to_system;
 
-    };
+};
 
-    // This class provides the functor we will supply to System::project_vector
-    // inside InterMeshProjection::project_system_vectors.
-    // Object is constructed by passing in a pointer to the mesh function whose
-    // gradient we want to shim via operator().
-    class GradientMeshFunction : public FunctionBase<Gradient>
-    {
-        public:
-        // Constructor
-        GradientMeshFunction(MeshFunction * _mesh_function);
+// This class provides the functor we will supply to System::project_vector
+// inside InterMeshProjection::project_system_vectors.
+// Object is constructed by passing in a pointer to the mesh function whose
+// gradient we want to shim via operator().
+class GradientMeshFunction : public FunctionBase<Gradient>
+{
+public:
+  // Constructor
+  GradientMeshFunction(MeshFunction * _mesh_function);
 
-        // Destructor
-        virtual ~GradientMeshFunction () { }
+  // Destructor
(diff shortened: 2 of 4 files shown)


src/systems/inter_mesh_projection.C
```diff
@@ -75,23 +75,18 @@ void InterMeshProjection::project_system_vectors()
   (*solution_vector_serial) = solution_vector;
 
   // Construct a MeshFunction for the solution
-  MeshFunction * mesh_func_solution =
-    new MeshFunction(from_system.get_equation_systems(), *solution_vector_serial,
-                     from_system.get_dof_map(), variables_vector);
+  std::unique_ptr<MeshFunction> mesh_func_solution(new MeshFunction(from_system.get_equation_systems(), *solution_vector_serial,
```
roystgnr, line 78: The RAII here is much better than having to manually delete later, but you can use std::make_unique() to get rid of the call to new too.

src/systems/inter_mesh_projection.C
```diff
@@ -75,23 +75,18 @@ void InterMeshProjection::project_system_vectors()
...
+                     from_system.get_dof_map(), variables_vector));
 
   mesh_func_solution->init();
 
   // For some element types (say C1) we also need to pass a gradient evaluation MeshFunction
   // To do this evaluate, a new shim class GradientMeshFunction has been added which redirects
   // gptr::operator evaluations inside projection methods into MeshFunction::gradient calls.
-  GradientMeshFunction * gptr_solution = new GradientMeshFunction(mesh_func_solution);
+  std::unique_ptr<GradientMeshFunction> gptr_solution(new GradientMeshFunction(mesh_func_solution.get()));
```
roystgnr, line 86: Likewise

src/systems/inter_mesh_projection.C
roystgnr: Actually ... do we need heap allocation here at all?  I had to use it in the original projection.C because we needed a separate MeshFunction for each variable, but now that we support that variables_vector and only need one MeshFunction, can't we just have the MeshFunction (and GradientMeshFunction) on the stack?

src/systems/inter_mesh_projection.C
```diff
@@ -110,24 +105,19 @@ void InterMeshProjection::project_system_vectors()
       from_system.get_vector(vec_name).localize(*current_vector_proxy);
 
       // Construct a MeshFunction for the current component
-      MeshFunction * mesh_func =
-        new MeshFunction(from_system.get_equation_systems(), *current_vector_proxy,
-                         from_system.get_dof_map(), variables_vector);
-
+      std::unique_ptr<MeshFunction> mesh_func(new MeshFunction(from_system.get_equation_systems(), *current_vector_proxy,
```
roystgnr, line 108: Here we're having to loop over all the vectors being projected, but since we don't need each MeshFunction outside the loop, we should be able to just keep things on the stack again, right?

roystgnr:
In the future, please keep whitespace changes in their own commits.  I like to be able to review 5 lines of code changes by looking at the 5 lines that really changed, without having to pick them out of 50 lines of whitespace change camouflage first.  :laughing:

src/systems/inter_mesh_projection.C
jwpeterson: > you can use std::make_unique() to get rid of the call to new too.

`libmesh_make_unique<T>` still, unless we officially allow C++14 now that you merged #2842?

include/systems/inter_mesh_projection.h
```diff
@@ -37,71 +37,71 @@
...
 
-        virtual std::unique_ptr<FunctionBase<Gradient>> clone () const
-        {
-          //return libmesh_make_unique<GradientMeshFunction>(dynamic_cast<MeshFunction *>(mesh_function.get()));
-          return libmesh_make_unique<GradientMeshFunction>(mesh_function.get());
-        }
+  virtual std::unique_ptr<FunctionBase<Gradient>> clone () const
+  {
+    //return libmesh_make_unique<GradientMeshFunction>(dynamic_cast<MeshFunction *>(mesh_function.get()));
+    return libmesh_make_unique<GradientMeshFunction>(mesh_function.get());
```
roystgnr, line 91: ... I'm not sure how I missed this before.  Are you "cloning" here, not by making a new copy, but by giving two different unique_ptr objects control over the same raw pointer???

include/mesh/mesh_function.h
```diff
@@ -84,13 +84,17 @@ class MeshFunction : public FunctionBase<Number>,
                 const unsigned int var,
                 const FunctionBase<Number> * master=nullptr);
 
+  /**
+   * A regular copy constructor. Replicates the steps of MeshFunction::clone().
```
jwpeterson, line 88: The copy constructor should not replicate `MeshFunction::clone()`, `MeshFunction::clone()` should just call the copy constructor, e.g.
```
return libmesh_make_unique<MeshFunction>(*this);
```

src/mesh/mesh_function.C
```diff
@@ -71,7 +71,18 @@ MeshFunction::MeshFunction (const EquationSystems & eqn_systems,
...
 
-
+MeshFunction::MeshFunction (const MeshFunction & mf):
+  FunctionBase<Number> (mf._master),
+  ParallelObject       (mf._eqn_systems),
+  _eqn_systems         (mf._eqn_systems),
+  _vector              (mf._vector),
+  _dof_map             (mf._dof_map),
+  _system_vars         (mf._system_vars),
+  _out_of_mesh_mode    (false)
```
jwpeterson, line 81: `_out_of_mesh_mode` needs to be copied from the incoming `MeshFunction`, not hard-coded to `false`.

roystgnr: Agreed, but take a look at our existing code too and you can see what went wrong - he's not adding a new bug here, he's retaining an existing bug.

jwpeterson: > he's retaining an existing bug.

I guess he copy-pasted from one of the existing constructors...? There I'd argue it's not a bug to hard-code `_out_of_mesh_mode` to false.

src/mesh/mesh_function.C
```diff
@@ -71,7 +71,18 @@ MeshFunction::MeshFunction (const EquationSystems & eqn_systems,
...
+{
+  this->init();
```
roystgnr, line 83: I'm not sure we want to call this unless ```mf.initialized()``` - someone might want to copy construct (or clone) a MeshFunction before the underlying Mesh is ready.  Although again this looks like there's a pre-existing bug - in our previous clone() code we checked initialized() before calling init(), but we then set_point_locator_tolerance() willy-nilly regardless of whether we know we've initialized _point_locator or not.

src/systems/inter_mesh_projection.C
roystgnr: libmesh_make_unique still, you're right.

src/mesh/mesh_function.C
roystgnr: Maybe, but I'm not just talking about the existing constructors, rather about the existing ```clone()```.  It doesn't pass ```_out_of_mesh_mode``` to the constructor used there (because there's no constructor that takes it) but it also doesn't use a setter to change ```_out_of_mesh_mode``` afterwards.

roystgnr: Ooh, and this appears to be in the "oldest libMesh bugs" Hall Of Fame.  I added FunctionBase::clone() and implementations in 9a3d8b10308cfb247c55d8c7abcb05f13488bda3 in Feb 2012, and at that time ```_out_of_mesh_mode``` already existed yet somehow I just ignored it.

vikramvgarg: Sorry, this is terrible. Was a bit late in the day.

src/systems/inter_mesh_projection.C
roystgnr: Thinking about this some more: In theory our CI tests are still going to catch any C++14 code that slips into libMesh, despite the compiler flag request changes in #2842, because at least the minimum-clang test ends up falling back on C++11 when it hits that too-old-libstdc++ problem while testing code under clang++ C++14 settings.  But in practice I should probably double-check with @loganharbour that the CI migrations they've been working on aren't going to change anything in that regard.

My personal preference is leaning more and more in favor of us to start requiring C++17 support and using newer features by the end of the 2021, a year and a half after the new stable RHEL ecosystem started supporting it by default (and years after major compilers have been supporting it for power users), but I definitely want any libMesh standards requirement upgrade to be a collective reasoned decision, not an accident, and I want to get at least one more C++11-compatible release out first.

src/mesh/mesh_function.C
roystgnr: Nah, this is a perfectly understandable bug.  (where "perfectly understandable" is super-conveniently defined as "I wrote the same bug too")  Feel free to give yourself hell over the other clone() bug, though.  :wink:

src/systems/inter_mesh_projection.C
vikramvgarg: > Actually ... do we need heap allocation here at all? I had to use it in the original projection.C because we needed a separate MeshFunction for each variable, but now that we support that variables_vector and only need one MeshFunction, can't we just have the MeshFunction (and GradientMeshFunction) on the stack?

This is for each system vector.

include/systems/inter_mesh_projection.h
roystgnr: And the answer is "no, you're not" - that mesh_function.get() isn't a ```GradientMeshFunction*``` that's about to get owned by a second non-unique unique_ptr, it's a ```MeshFunction*``` that's getting used as the constructor argument to a brand new unique ```GradientMeshFunction```.  Sorry about the mix up.  Apparently "how I missed this before" is "I was paying closer attention before".

src/systems/inter_mesh_projection.C
roystgnr: Sure, but do you operate on all the vectors at the same time, so that you have to have all the corresponding MeshFunctions defined at the same time?  If each MeshFunction is getting used only within a single block then you might as well declare it on the stack in that block and let it go out of scope afterwards.

src/mesh/mesh_function.C
```diff
@@ -78,10 +78,18 @@ MeshFunction::MeshFunction (const MeshFunction & mf):
   _vector              (mf._vector),
   _dof_map             (mf._dof_map),
   _system_vars         (mf._system_vars),
-  _out_of_mesh_mode    (false)
+  _out_of_mesh_mode    (mf._out_of_mesh_mode)
 {
-  this->init();
-  this->set_point_locator_tolerance(mf.get_point_locator().get_close_to_point_tol());
+  // Initialize the mf and set the point locator if the
```
jwpeterson, line 83: Reading this comment in Samuel L. Jackson's voice while expanding out the "mf" abbreviation.... :+1:

jwpeterson (approved):
Latest changes look good to me. I shared @roystgnr's confusion about the `GradientMeshFunction::clone()` routine, so that could be worth adding a comment to.

roystgnr:
Glad I'm not the only one who was confused.  I still feel bad for giving Vikram flak over correct code.

But rather than a comment, let's change the syntax itself to be less confusing?  GradientMeshFunction doesn't work without an underlying MeshFunction, so taking a pointer in its constructor rather than a reference is pointless; the pointer should never be null.  Let's take a const reference there and there'll be no implication that memory management might be involved, nor that modifying the input (vs modifying a clone of it) might ever happen.

jwpeterson:
> Let's take a const reference there

Sounds good to me, I assumed it was conforming to some existing API, but if we can change it to a reference easily, let's do it.

vikramvgarg:
> Glad I'm not the only one who was confused. I still feel bad for giving Vikram flak over correct code.
> 
> But rather than a comment, let's change the syntax itself to be less confusing? GradientMeshFunction doesn't work without an underlying MeshFunction, so taking a pointer in its constructor rather than a reference is pointless; the pointer should never be null. Let's take a const reference there and there'll be no implication that memory management might be involved, nor that modifying the input (vs modifying a clone of it) might ever happen.

Yes, const reference is what it should have always been. I am trying to remember why I choose a pointer initially, I should have documented my decisions in the comments there.

include/systems/inter_mesh_projection.h
```diff
@@ -87,7 +87,7 @@ class GradientMeshFunction : public FunctionBase<Gradient>
 
   virtual std::unique_ptr<FunctionBase<Gradient>> clone () const
   {
-    return libmesh_make_unique<GradientMeshFunction>(mesh_function.get());
+    return libmesh_make_unique<GradientMeshFunction>(*(mesh_function.get()));
```
roystgnr, line 90: Just ```*mesh_function``` will do here; ```unique_ptr``` defines ```operator*``` so you don't need to manually ```get()``` the underlying pointer.

roystgnr (approved):

PR #2862, opened 2021-02-13
Reduced Basis + "IGA" fixes

roystgnr:
So, it turns out that our existing reduced_basis code doesn't support the fully general constraint equations that our IGA code currently depends on, and that I never noticed this because I inadvertently set the wrong boundary conditions in the IGA case of reduced_basis_ex5, and that when we give reduced_basis_ex5 the correct boundary conditions then the greedy basis expansion doesn't converge.

This doesn't fix that, but does work around it: instead of always loading a complete IGA Bezier Extraction mesh with all the finite elements plus spline nodes plus constraint equations connecting them, we now have the option to load an "IGA" Bezier mesh with *just* the finite elements.  The resulting finite element space still uses the same Rational Bezier mapping functions, so is still isogeometric in a sense, but it's no longer constrained to have the higher basis function continuity of the original spline basis ... and because of that lack of constraints it now converges in reduced_basis problems (for H1-conforming weak formulations, anyway).


roystgnr:
This seems to be working beautifully (on other IGA mesh files too) for me, so we're ready to merge.  But I can generate the results I need by just working on the branch, so it can wait for review before merging.

roystgnr:
I'm working in brief spurts between Texas' rolling blackouts (and during blackouts, thanks @permcody for suggesting the biggest UPS we could find...), but I don't see @jwpeterson online, and this freeze + grid disaster is supposed to last for days, so I'm going to merge this and #2863 so we don't keep @lindsayad waiting in https://github.com/idaholab/moose/pull/16965 any longer.  If we (or the tests there) catch anything after the fact then we can revert the PR(s).

PR #2864, opened 2021-02-17
Mac warning fixes

roystgnr:
My last clang++ build on OSX spat out a couple reasonable compiler warnings.

### DIFF
--- a/src/reduced_basis/rb_construction.C
+++ b/src/reduced_basis/rb_construction.C
@@ -1293,7 +1293,7 @@ void RBConstruction::enrich_basis_from_rhs_terms(const bool resize_rb_eval_data)
 
       // Debugging: enable this code to print the rhs that was used in
       // the most recent truth solve to a uniquely-named file.
-      if (false)
+#if 0
         {
           char temp_file[] = "truth_rhs_XXXXXX.dat";
           int fd = mkstemps(temp_file, 4);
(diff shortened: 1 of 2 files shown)


src/reduced_basis/rb_construction.C
```diff
@@ -1303,11 +1303,12 @@ void RBConstruction::enrich_basis_from_rhs_terms(const bool resize_rb_eval_data)
               rhs->print_matlab(std::string(temp_file));
             }
         }
+#endif // 0
 
       // Debugging: enable this code to print the most recent truth
       // solution to a uniquely-named file.
 #ifdef LIBMESH_HAVE_EXODUS_API
-      if (false)
```
jwpeterson, line 1310: So clang warns about `if (false)` now? I find that kind of annoying because with `#if 0` it's too easy to introduce typos and other errors while refactoring, as the code is no longer seen by the compiler at all. I think with `if (false)` it at least still reports errors.

jwpeterson (approved):

src/reduced_basis/rb_construction.C
roystgnr: That's a good point.  That's more than just a little annoying.  I don't see any other way around it, though, unless we revoke clang's -Wunreachable-code privileges in compiler.m4.  That might be worth considering, since we'll be getting roughly the same warning coverage in our gcc builds?  But it might just be a matter of time before gcc adopts the same interpretation; it's hard to argue that the stuff inside ```if (false)``` isn't indeed unreachable code.

Issue #2861, opened 2021-02-13
Compilation Failure with PETSc 3.14.4


pbauman:
Just tried building master with PETSc 3.14.4 release tarball:

https://github.com/libMesh/libmesh/blob/master/src/solvers/petsc_linear_solver.C#L1268

```
#if PETSC_RELEASE_LESS_THAN(3,14,4)
  PetscReal * p;
#else
  const PetscReal * p;
#endif

  ierr = KSPGetResidualHistory(_ksp, &p, &its);
```
Looks like API change didn't actually land in 3.14.4 (though I see it's in the PETSc master branch). :(

jwpeterson:
I'm guessing they will not make that API change in a "maint" release... we can probably change it to 
`PETSC_RELEASE_LESS_THAN(3,15,0)`, what do you think?

pbauman:
Yeah, agreed.

roystgnr:
And now I get the same compilation failure in PETSc 3.14.5.

Sorry for not following along with this (in my defense I was busy enjoying my unemployment/vacation on January 13!), but ... doesn't #2833 make no sense, on multiple levels?

1.  You don't need a pointer to be declared pointer-to-const to pass it to an API taking pointer-to-const.  C and C++ will automatically convert `T*` to `const T*` for you.
2.  This is an *output* parameter for `KSPGetResidualHistory`, not an input parameter.  Who would ever change that to be const?  They'd just have to cast it to non-const internally or the API wouldn't work at all.

roystgnr:
Ah, no, I get it now!

roystgnr:
We're passing a pointer-to-a-pointer there.  And in the future that's going to be a pointer-to-a-const-pointer.

jwpeterson:
I thought we fixed this in #2866, how are you still getting the error?

roystgnr:
https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/KSP/KSPGetResidualHistory.html still lists no const for that parameter in the declaration.

roystgnr:
I guess they're going to change it eventually, it does make sense as a change since users shouldn't be using that pointer to alter that data ...

Wait.  How *am* I still getting the error?  Did they make the API change in 3.14.5 instead of waiting for 3.15.0?

roystgnr:
```
$ grep KSPGetResidualHistory LIBRARIES/petsc-3.14.5/include/*.h
[...]

PR #2866, opened 2021-02-22
Fix PETSc version guard

jwpeterson:
PETSc API changes generally don't make it into "maint" releases, so
update this one to the next point release instead.

Refs #2861


roystgnr (approved):
Thanks for getting to this!

PR #2868, opened 2021-03-01
PointLocatorTree: use std::shared_ptr to manage tree

jwpeterson:
The master/sub PointLocator models `std::shared_ptr` ownership semantics, so it makes sense to use that rather than raw new/delete to manage the resource. This also means that the destructor can be defaulted.

Refs #2865

### DIFF
--- a/include/utils/point_locator_tree.h
+++ b/include/utils/point_locator_tree.h
@@ -26,6 +26,7 @@
 
 // C++ includes
 #include <cstddef>
+#include <memory> // std::shared_ptr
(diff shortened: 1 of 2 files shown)


roystgnr (approved):
It would be nice to use weak_ptr for non-master locators in debug+devel modes, to try to detect cases where a user hangs on to one too long and then tries to use it after its master has been deleted.  I can't see how to do that which wouldn't introduce more opportunities for bugs in library code, though.

PR #2871, opened 2021-03-03
modify two unit tests

hugary1995:
I'm demonstrating a current limitation: if a nodeset/sideset is empty, it won't be written to exodus.

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -1663,7 +1663,25 @@ void ExodusII_IO_Helper::initialize(std::string str_title, const MeshBase & mesh
     for (const auto & id : shellface_boundaries)
       unique_side_boundaries.push_back(id);
   }
+  // Add any empty-but-named side boundary ids
+  for (const auto & pair : bi.get_sideset_name_map())
+    {
+      const boundary_id_type id = pair.first;
+
+      if (std::find(unique_side_boundaries.begin(),unique_side_boundaries.end(),id)
+            == unique_side_boundaries.end())
+        unique_side_boundaries.push_back(id);
+    }
+
   bi.build_node_boundary_ids(unique_node_boundaries);
+  for (const auto & pair : bi.get_nodeset_name_map())
+    {
+      const boundary_id_type id = pair.first;
+
+      if (std::find(unique_node_boundaries.begin(),unique_node_boundaries.end(),id)
+            == unique_node_boundaries.end())
+        unique_node_boundaries.push_back(id);
+    }
 
   num_side_sets = cast_int<int>(unique_side_boundaries.size());
   num_node_sets = cast_int<int>(unique_node_boundaries.size());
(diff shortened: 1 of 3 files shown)


roystgnr:
Beautiful; thanks!  Let me make sure this replicates for me and I'll see if the suggestion I was proposing will work on it.

jwpeterson:
Will Exodus actually let you store zero-size nodesets or sidesets in the file? I think this isn't allowed, but not 100% sure.

jwpeterson:
@hugary1995 I sent you an invite to join the team that gets its PRs tested automatically.

roystgnr:
These tests fail as expected for me.  I'm pretty disturbed that we're only seeing failures in CI on the "Minimum gcc version" recipes.  We can't possibly have Exodus disabled on every other recipe, can we??

hugary1995:
It seems to me that most recipes didn't even run unit tests.

roystgnr:
That's got to be it.  We don't have cppunit available for enough of them.

roystgnr:
> Will Exodus actually let you store zero-size nodesets or sidesets in the file? I think this isn't allowed, but not 100% sure.

There's definitely code to handle that case in the current version in contrib/exodus/.  Like, specifically testing for 0-sized sets and taking a particular code path rather than immediately returning with an error.  My first attempt to actually *use* that feature died with a later Exodus error, but it looks like that was a bug on my part rather than on theirs.  I've got the sideset case working now and I think I've almost got the nodeset case working.

hugary1995:
I confirm that this fixes the empty nodeset/sideset issue in my moose PR.

jwpeterson (commented):
Looks OK to me, but assuming that we already have a test of *reading* Exodus node/sidesets, would you mind adding a unit test of reading *empty* side/nodesets? If we don't already have a reading test, then nevermind I guess...

src/mesh/exodusII_io_helper.C
```diff
@@ -1663,7 +1663,25 @@ void ExodusII_IO_Helper::initialize(std::string str_title, const MeshBase & mesh
...
+            == unique_side_boundaries.end())
+        unique_side_boundaries.push_back(id);
+    }
+
   bi.build_node_boundary_ids(unique_node_boundaries);
+  for (const auto & pair : bi.get_nodeset_name_map())
+    {
+      const boundary_id_type id = pair.first;
+
+      if (std::find(unique_node_boundaries.begin(),unique_node_boundaries.end(),id)
```
jwpeterson, line 1681: Usually we put a space after commas, right? Maybe this is not an official coding style guideline.

hugary1995:
> Looks OK to me, but assuming that we already have a test of _reading_ Exodus node/sidesets, would you mind adding a unit test of reading _empty_ side/nodesets? If we don't already have a reading test, then nevermind I guess...

The unit test I modified is like a reading/writing unit test 2-in-1. It first writes some variables on certain nodesets/sidesets, than read the exodus back in. It asserts that the written values match the read-in values.

src/mesh/exodusII_io_helper.C
roystgnr: We do, and it's not but it should be.  Will fix.

jwpeterson:
> The unit test I modified is like a reading/writing unit test 2-in-1

OK, thanks, you're correct. I was thrown off by the test being called `testWrite()`.

jwpeterson (approved):

roystgnr:
I really do need to add Nemesis support for this too, but that can go in another PR.

lindsayad:
@hugary1995 you broke BISON: https://civet.inl.gov/job/713298/ It would be great if you could fix those tests in a forward and backward compatible way such that we don't have to do the submodule update dance with temporarily broken MOOSE CI on next, but if that's not possible, then can you have those regolds prepared for when we update the MOOSE/libmesh hash in BISON?

lindsayad:
I should say that I'm *pretty* sure that this is the PR breaking BISON...

hugary1995:
I don't have access to BISON unfortunately. I don't understand how this PR could fail BISON. If it is really related to this PR, it almost certainly means that some assessment cases in BISON were writing empty nodeset/sidesets to Exodus before. But why?

lindsayad:
The exodiffs show that there are more sidesets/nodesets being written now than exist in the gold file.

PR #2875, opened 2021-03-06
Fix matrix_matrix_mult()

dewenyushu:
- Let user specify if we should reuse the product matrix.
- Make sure we delete the output matrix if we do not reuse the matrix. Otherwise memory leak is possible.

### DIFF
--- a/src/numerics/petsc_matrix.C
+++ b/src/numerics/petsc_matrix.C
@@ -80,6 +80,7 @@ namespace libMesh
 template <typename T>
 PetscMatrix<T>::PetscMatrix(const Parallel::Communicator & comm_in) :
   SparseMatrix<T>(comm_in),
+  _mat(nullptr),
   _destroy_mat_on_exit(true),
   _mat_type(AIJ)
 {}
@@ -1248,7 +1249,7 @@ void PetscMatrix<T>::add (const T a_in, const SparseMatrix<T> & X_in)
 
 
 template <typename T>
-void PetscMatrix<T>::matrix_matrix_mult (SparseMatrix<T> & X_in, SparseMatrix<T> & Y_out)
+void PetscMatrix<T>::matrix_matrix_mult (SparseMatrix<T> & X_in, SparseMatrix<T> & Y_out, bool reuse)
 {
   libmesh_assert (this->initialized());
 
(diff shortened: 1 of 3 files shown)


src/numerics/petsc_matrix.C
```diff
@@ -80,6 +80,7 @@ namespace libMesh
 template <typename T>
 PetscMatrix<T>::PetscMatrix(const Parallel::Communicator & comm_in) :
   SparseMatrix<T>(comm_in),
+  _mat(NULL),
```
jwpeterson, line 83: I'm a bit surprised we didn't run into issues with not initializing `_mat` in the past. Definitely seems like the right thing to do, but we should probably use `nullptr` for consistency with C++11.

src/numerics/petsc_matrix.C
```diff
@@ -1266,9 +1267,18 @@ void PetscMatrix<T>::matrix_matrix_mult (SparseMatrix<T> & X_in, SparseMatrix<T>
...
-  LIBMESH_CHKERR(ierr);
-
+  if (reuse)
+  {
+    ierr = MatMatMult(_mat, X->_mat, MAT_REUSE_MATRIX, PETSC_DEFAULT, &Y->_mat);
+    LIBMESH_CHKERR(ierr);
+  }
+  else
+  {
+    ierr = MatDestroy(&Y->_mat);
```
jwpeterson, line 1277: This I'm not so sure about... the `Y` that is passed in could have been used for anything previously, and, for instance, it could have `_destroy_mat_on_exit == false`, in which case we should not call `MatDestroy` like this. I think it should be OK to call `Y->clear()`, which handles this distinction as well as doing some error checking/setting of flags.

jwpeterson (changes requested):

src/numerics/petsc_matrix.C
roystgnr: I'm not too surprised we haven't run into errors; we generally init() pretty quickly after construction.

As for warnings?  IIRC this isn't caught by -Wuninitialized.  I think it is caught by -Weffc++, but so are so many false positives (even in libstdc++ headers, according to `man gcc`) that we never use that.

src/numerics/petsc_matrix.C
roystgnr: Agreed.  But also: was this a bug before?  Our existing matrix_matrix_mult() would have given us some error or some memory leak if Y had a Mat which had already been initialized?

src/numerics/petsc_matrix.C
jwpeterson: In pretty much any typical PETSc code you would see code like:
```
Mat mat;
```
i.e. without explicit nullptr-initialization, so we were at least mimicking that behavior previously.

src/numerics/petsc_matrix.C
dewenyushu: Thanks. This is a great suggestion.

I think this is a bug but was somehow well handled by PETSc for most cases. The memory issue came up when I was trying to do matrix inversion using `MATSOLVERSUPERLU_DIST`. The original `matrix_matrix_mult()` results in a hanging pointer and `Petsc_Superlu_dist_keyval_Delete_Fn()` was complaining about this upon existing.

jwpeterson (approved):

PR #2876, opened 2021-03-07
Mesh assign

vikramvgarg:
Branch adds mesh reassignment and reinit_mesh capability, so that a previously initialized ES object can be assigned a new mesh object.


include/mesh/distributed_mesh.h
```diff
@@ -89,6 +89,12 @@ class DistributedMesh : public UnstructuredMesh
   virtual std::unique_ptr<MeshBase> clone () const override
   { return libmesh_make_unique<DistributedMesh>(*this); }
 
+  /**
+   * Assignment function, will move argument onto the MeshBase object.
+   * Definition in derived classes.
+   */
+  virtual void assign (MeshBase & other_mesh) override;
+
```
jwpeterson, line 97: But you aren't implementing "move" assignment, you're implementing (non-standard) copy assignment.

jwpeterson (changes requested):
I think you might need to start by telling us what you are trying to achieve here, and why this is the way you have decided to do it. I am hesitant to add hand-coded copy/move assign/construct operators because they are notoriously difficult to maintain. I'm even more hesitant to create a precedent of having "named" copy or move assignment operators when the language provides a standard way of naming those functions.

Also, this PR has no tests of an intricate new feature.

jwpeterson:
It looks like there are some useful bits that should be merged (and unit tested!) separately like `Variable::operator==` and `VariableGroup::operator==`.

roystgnr:
> Also, this PR has no tests of an intricate new feature.

This should be *really* easy to test, too.  Start with any small unit test that sets up a System with variables, build and assign from a new mesh, then reinit_mesh(), then assert that n_variables() and n_dofs() and n_nodes() and everything are as expected.

roystgnr:
> It looks like there are some useful bits that should be merged (and unit tested!) separately like Variable::operator== and VariableGroup::operator==

Those plus the DofMap::add_variable_group change and the System::init_data change would make a small self-contained fix, if you like.  And then a basic unit test would be something like manually trying to repeat add_variable_group() and making sure you still only have the one group afterwards.  Detailed testing might be too tricky to bother with, though; with the original code the problem is that any duplicate variable group confuses the hell out of DofMap dof numbering and then DofObject fails internal assertions way down the line.

include/mesh/distributed_mesh.h
roystgnr: Vikram can correct me if I'm wrong, but I think what he wants here is actual move assignment, not copy assignment.  A swap() might work as an alternative, or as a better alternative for other use cases further down the line?

IIRC putting this into a "named assignment" was my idea, to try to make it harder for users to accidentally call an expensive copy.  If I'm right about move-and-later-swap being the right goals, though, then perhaps "letting C++ give users extra rope" is a worthwhile price to pay for letting C++ do its other magic?  IIRC with C++11 and later, if you make a class move constructable and move assignable, the standard then provides you with a std::swap implementation automatically.  And if we turned this into a proper move assignment operator that'd be most of the work done for when we want to add a move constructor later.

roystgnr:
Before making any other changes, would it be possible to fix the weird, weird git history here?  It looks like somehow you merged a branch onto a slightly modified copy of itself and now there's two copies of every commit?

vikramvgarg:
> Before making any other changes, would it be possible to fix the weird, weird git history here? It looks like somehow you merged a branch onto a slightly modified copy of itself and now there's two copies of every commit?

Yes, I would want to fix this first too. It happened after I did a rebase on master. There was one trivial conflict which I fixed, after which the branches did merge, but I ended up with this weird history.

vikramvgarg:
> I think you might need to start by telling us what you are trying to achieve here, and why this is the way you have decided to do it. 
 
@jwpeterson I should have probably put this in the original PR message, but here is the context. Consider an unsteady primal and adjoint program with AMR enabled. The mesh will be allowed to change with every time step during the primal time march. During the adjoint (backwards) timestepping, we will need to read in the time varying primal solution and mesh at every adjoint timestep from the disk or memory.

The underlying ES and systems will remain the same, but we will need to be able to read in an incoming mesh (upcoming in MeshHistory), replace the existing mesh attached to the ES with the incoming mesh (this branch) and then project the existing system vectors onto this mesh (InterMeshProjection). This is a reassignment of the mesh associated with the ES, which is why an assign rather than copy or swap paradigm was chosen.

vikramvgarg:
> > Also, this PR has no tests of an intricate new feature.
> 
> This should be _really_ easy to test, too. Start with any small unit test that sets up a System with variables, build and assign from a new mesh, then reinit_mesh(), then assert that n_variables() and n_dofs() and n_nodes() and everything are as expected.

Yes, this is pretty much what I have as the test already. I can add it as a unit test.

jwpeterson:
> This is a reassignment of the mesh associated with the ES, which is why an assign rather than copy or swap paradigm was chosen.

You want the Mesh owned by the ES to take over ownership of an outside Mesh, and you don't care what happens to the original Mesh that the ES owned? That is move assignment. IMO this should be implemented as the move-assignment operator for the class, since it is more self-documenting that way. 

We *could* implement only the MeshBase move assignment operator at this time, but it might be worth investigating whether MeshBase is already std::swap'able first. In that case, it would make more sense to implement on the the MeshBase move constructor. Then move assignment (and copy assignment) would come for "free" by using the copy-and-swap and move-and-swap idioms.

There's a decent discussion of move assignment and move construction and their relation to swap [here](https://stackoverflow.com/a/6687520/659433). In particular, if you implemented 

> It happened after I did a rebase on master. There was one trivial conflict which I fixed, after which the branches did merge,

If you do a rebase, then you should not *also* do a merge. Generally speaking, your PRs should always be rebased on a relatively recent master, then we will handle the merging on GitHub, so you should not have to merge (pull) at all on your end.

roystgnr:
> it would make more sense to implement the MeshBase move constructor. Then move assignment (and copy assignment) would come for "free" by using the copy-and-swap and move-and-swap idioms.

Except then where do we get swap?  You need a move constructor and move assignment operator already for that, or you need a manual implementation that'll be as complex as either.

I'd figured the way to go is: move assignment operator for now, move constructor later via empty-construct-then-assign, and that gets us std::swap for free.

roystgnr:
Also ... how far off are we from being able to get correct compiler-generated move construction/assignment?  From our comments it looks like the only issue is the `MeshBase` reference in `BoundaryInfo`?  That seems so close, though (even if we replace that reference with a pointer) I don't see an obvious way around the problem.

jwpeterson:
> Also ... how far off are we from being able to get correct compiler-generated move construction/assignment?

Hmm... the `BoundaryInfo` member in MeshBase is already stored in a `unique_ptr`, so shouldn't that already be move-assignable?

jwpeterson:
I think MeshBase is pretty close to being default move-assignable. The problem IIRC is that the `ReplicatedMesh` and `DistributedMesh` subclasses manually manage memory, so you'd need to set all those to nullptr after they were move assigned, to avoid a double-free when the destructor was eventually called.

jwpeterson:
BTW, I was thinking that `std::swap` would work on a class as long as all its members were std::swap'able, but that is how the default constructors and assignment operators work, not std::swap :(

roystgnr:
> Hmm... the BoundaryInfo member in MeshBase is already stored in a unique_ptr, so shouldn't that already be move-assignable?

Yeah, but the automatic move assignment won't be correct - the moved BoundaryInfo would still have a reference to the old MeshBase rather than the one it's been moved to.

There isn't any way to get the compiler to generate its default operator, but then add code to be called *after* the defaults, is there?

jwpeterson:
> the moved BoundaryInfo would still have a reference to the old MeshBase 

I see, yeah. There'd be a similar issue for the PointLocator.

> There isn't any way to get the compiler to generate its default operator, but then add code to be called after the defaults, is there?

No, it's all or nothing as far as I know.

vikramvgarg:
@jwpeterson @roystgnr I tried to follow the instructions on the installation page (http://libmesh.github.io/installation.html) to run the unit tests,
./configure
make
make -C tests check

However, the subdirectories within tests in my build folder are empty and I keep getting a 0 unit tests passed message when I run the above command. Have the options for compiling unit tests changed ?

roystgnr:
Do you have cppunit installed?  We don't include that in contrib/.

vikramvgarg:
> Do you have cppunit installed? We don't include that in contrib/.

No. Let me install that and try again.

include/mesh/distributed_mesh.h
```diff
@@ -81,7 +81,12 @@ class DistributedMesh : public UnstructuredMesh
    * Copy and move assignment are not allowed.
    */
   DistributedMesh & operator= (const DistributedMesh &) = delete;
-  DistributedMesh & operator= (DistributedMesh &&) = delete;
+
+  /**
+   * Overloaded operator= will move contents of other_mesh to calling
+   * MeshBase object.
+   */
+  virtual MeshBase & operator= (MeshBase && other_mesh) override;
```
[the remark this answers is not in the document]
vikramvgarg: @jwpeterson I gather that move assignment not being virtual is the convention. A very typical use case for the assignment operator here will be along the lines,

system.get_mesh() = std::move(*mesh_two);

Now System::get_mesh() returns type MeshBase, whereas mesh_two can also be of type MeshBase, if for example it was declared to be a shared_ptr to an UnstructuredMesh. This is the case in the unit test I have added. 

With operator=(&&) being virtual, it appears that the proper derived class types for the system mesh and mesh_two (Replicated/Distributed in the unit test) were detected and the program gave the correct output.

If I remove the virtual qualifier, the program reverts to the MeshBase implementation rather than resolving the derived type of the calling object automatically.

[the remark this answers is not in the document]
vikramvgarg: It looks like virtual assignment operators are ok if we ensure that base class move operators are only called once,
https://stackoverflow.com/questions/64325516/move-assignment-operator-and-virtual-inheritance/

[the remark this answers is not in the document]
roystgnr: You're absolutely correct that the move assignment operator shouldn't be virtual ... but notice that it's kind of entirely our fault that he went down that path:

"I'm even more hesitant to create a precedent of having "named" copy or move assignment operators when the language provides a standard way of naming those functions."

As I should have realized at the time, the language *doesn't* provide a standard way of providing virtual move assignment, does it?  And virtual move assignment is what we need, to do a non-slicing move assignment into a target for which our only handle is a reference-to-base-class.  ```virtual assign(MeshBase &&)``` ought to be implemented as shims to ```ReplicatedMesh::operator=(ReplicatedMesh &&)``` and ```DistributedMesh::operator=(DistributedMesh &&)``` under the hood (after a libmesh_cast_ref to ensure that we're not trying to mix the two), but to do the move into a base class reference it's going to need to be virtual, so it ought to be a named assign() function itself, right?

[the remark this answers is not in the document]
jwpeterson: OK, damn, I failed to understand the use case which @vikramvgarg just articulated above, which clearly has to involve virtual functions of some kind or another, and not just standard move assignment. I think we are a lot closer than we were before, however. Your idea of a virtual assign that is overridden in the bases to call move assignment feels like a good approach to me, plus now we will also have normal move assignment for Meshes.

include/mesh/distributed_mesh.h
```diff
@@ -84,9 +84,19 @@ class DistributedMesh : public UnstructuredMesh
...
    */
-  virtual MeshBase & operator= (MeshBase && other_mesh) override;
+  DistributedMesh & operator= (DistributedMesh && other_mesh);
+
+  /**
+   * Shim to call the move assignment operator for this class
+  */
+  virtual MeshBase & assign(MeshBase * other_mesh) override
+  {
+    *(libmesh_cast_ptr<DistributedMesh*>(this)) = std::move(*(libmesh_cast_ptr<DistributedMesh*>(other_mesh)));
```
roystgnr, line 96: No need to cast `this`, only `other_mesh`

include/mesh/mesh.h
```diff
@@ -84,10 +84,14 @@ class Mesh : public DefaultMesh
...
 
   /**
-   * Use the equality operator of the base DefaultMesh class
-   * if an rvalue argument is provided.
+   * Provides move functionality via base class move assignment operator.
    */
-  using DefaultMesh::operator=;
+  virtual MeshBase & assign(MeshBase * other_mesh) override
+  {
+    *(libmesh_cast_ptr<DefaultMesh*>(this)) = std::move(*(libmesh_cast_ptr<DefaultMesh*>(other_mesh)));
```
roystgnr, line 91: likewise

include/mesh/parallel_mesh.h
```diff
@@ -42,10 +42,14 @@ class ParallelMesh : public DistributedMesh
...
-   * if an rvalue argument is provided.
+   * Provides move functionality via base class move assignment operator.
    */
-  using DistributedMesh::operator=;
+  virtual MeshBase & assign(MeshBase * other_mesh) override
+  {
+    DistributedMesh::assign(other_mesh);
+
+    return *this;
+  }
```
roystgnr, line 52: Is this necessary?  Shouldn't we already be inheriting DistributedMesh::assign here?

include/mesh/mesh_base.h
```diff
@@ -110,7 +110,13 @@ class MeshBase : public ParallelObject
...
    */
   MeshBase & operator= (const MeshBase &) = delete;
-  virtual MeshBase & operator= (MeshBase && other_mesh);
+  MeshBase & operator= (MeshBase && other_mesh);
+
+  /**
+   * Shim to allow operator = (&&) to behave like a virtual function
+   * without having to be one.
+   */
+  virtual MeshBase & assign(MeshBase * other_mesh) = 0;
```
roystgnr, line 119: Oh, no.  Make that argument an rvalue reference.  We do not want to move from an input parameter without being sure the user expects it!

src/mesh/unstructured_mesh.C
```diff
@@ -55,6 +55,62 @@ UnstructuredMesh::UnstructuredMesh (const Parallel::Communicator & comm_in,
...
+        }
+      else
+        {
+          libmesh_deprecated();
+          add_ghosting_functor(*(*gf));
+        }
+    }
+
+  if (other_mesh.partitioner())
+    _partitioner = std::move((other_mesh.partitioner())->clone());
```
roystgnr, line 109: We're doing a move here, trying to do as little copying as possible - why did we need the `clone()` first?

src/mesh/unstructured_mesh.C
```diff
@@ -55,6 +55,62 @@ UnstructuredMesh::UnstructuredMesh (const Parallel::Communicator & comm_in,
...
+{
+  // First make a call to MeshBase::clear
+  this->MeshBase::clear();
+
+  this->ParallelObject::operator=(other_mesh);
+
+  // First set this mesh equal to other_mesh as a MeshBase.
+  this->MeshBase::operator=(std::move(other_mesh));
+
+  this->copy_nodes_and_elements(dynamic_cast<UnstructuredMesh &>(other_mesh), true);
```
roystgnr, line 68: This is adding an RTTI dependency; that's a no.

Also: we're trying to do an O(1) move here, not an O(N) copy.  Using an API with "copy" in the name is a good sign that we're failing.  ;-)  To actually do a move, we're going to have to do it in the subclasses with the containers whose contents we're wanting to move.

include/mesh/parallel_mesh.h
vikramvgarg: True, same for BoundaryMesh as well.

src/mesh/unstructured_mesh.C
vikramvgarg: Fixed.

src/mesh/unstructured_mesh.C
vikramvgarg: Oops, guess I got so caught up in boundary functors and virtual functions that I forgot all about the nodes and elements .... Will use the move assignment on the _elements and _nodes vectors.

vikramvgarg:
@roystgnr @jwpeterson , the PR is ready for review.

roystgnr:
Sorry for the delay; I was kind of buried in #2900.

src/mesh/distributed_mesh.C
```diff
@@ -60,14 +60,30 @@ DistributedMesh::DistributedMesh (const Parallel::Communicator & comm_in,
...
 {
-  // First set this mesh equal to other_mesh as an UnstructuredMesh.
+  // First make a call to MeshBase::clear
+  this->MeshBase::clear();
+
+  // Move assign as a ParallelObject.
+  this->ParallelObject::operator=(other_mesh);
+
+  // Move assign as a MeshBase.
+  this->MeshBase::operator=(std::move(other_mesh));
```
roystgnr, line 70: In theory, shouldn't this also do the ParallelObject::operator= ?  In the normative sense of "should", at least.

src/mesh/distributed_mesh.C
```diff
@@ -60,14 +60,30 @@ DistributedMesh::DistributedMesh (const Parallel::Communicator & comm_in,
...
+
+  // Nodes and elements belong to DistributedMesh and have to be
+  // moved before we can move the BoundaryInfo object in UnstructuredMesh::operator=.
+  this->move_nodes_and_elements(std::move(other_mesh));
+
+  // Move assign as an UnstructuredMesh.
   this->UnstructuredMesh::operator=(std::move(other_mesh));
 
+  // The remaining work to move a DistributedMesh
   _is_serial = std::move(reinterpret_cast<DistributedMesh &>(other_mesh).is_serial());
```
roystgnr, line 80: I should have noticed before, but: std::move() here is pointless.  `is_serial()` returns by value, so it's already a temporary, so it's already an rvalue even without std::move().  And even if it wasn't, there's no move optimization possible with a simple bool; copying a bool is already as cheap as it gets.  So we might as well just set `_is_serial = other_mesh.is_serial()` here and make the code cleaner.

roystgnr: Likewise for _is_serial_on_proc_0, _max_node_id, _max_elem_id.

roystgnr: Etcetera.  Basically anything that's not doing a heap allocation anywhere under the hood (i.e. anything that's not a container or a complicated class) can just be copied; move is helpful by letting us just move pointers rather than their contents.

src/mesh/distributed_mesh.C
```diff
@@ -213,6 +229,14 @@ DistributedMesh::DistributedMesh (const UnstructuredMesh & other_mesh) :
   this->update_parallel_id_counts();
 }
 
+void DistributedMesh::move_nodes_and_elements(MeshBase && other_mesh)
+{
+  this->_nodes = std::move((libmesh_cast_ptr<DistributedMesh*>(&other_mesh))->_nodes);
+  this->_n_nodes = dof_id_type(_nodes.size());
```
roystgnr, line 235: No, it is not.  Likewise for this->_n_elem.  On a distributed mesh we aren't storing every single DofObject.  Just copy the other mesh's caches.  We wouldn't have them if they were redundant.

src/mesh/replicated_mesh.C
```diff
@@ -173,6 +173,37 @@ ReplicatedMesh::ReplicatedMesh (const UnstructuredMesh & other_mesh) :
...
+ReplicatedMesh & ReplicatedMesh::operator= (ReplicatedMesh && other_mesh)
+{
+  // First make a call to MeshBase::clear
+  this->MeshBase::clear();
+
+  // Move assign as a ParallelObject.
+  this->ParallelObject::operator=(other_mesh);
+
+  // Move assign as a MeshBase.
+  this->MeshBase::operator=(std::move(other_mesh));
```
roystgnr, line 185: Same question about ParallelObject

src/mesh/replicated_mesh.C
```diff
@@ -173,6 +173,37 @@ ReplicatedMesh::ReplicatedMesh (const UnstructuredMesh & other_mesh) :
...
+  this->UnstructuredMesh::operator=(std::move(other_mesh));
+
+  return *this;
+}
+
+
+void ReplicatedMesh::move_nodes_and_elements(MeshBase && other_mesh)
+{
+  this->_nodes = std::move((libmesh_cast_ptr<ReplicatedMesh*>(&other_mesh))->_nodes);
+  this->_n_nodes = dof_id_type(_nodes.size());
```
roystgnr, line 201: This isn't nearly as inaccurate, but it's still inaccurate.  We keep _n_nodes so we can keep track of nodes which have been nulled out but (because the container is indexed by id and because vector entry removal is O(N)) haven't been removed from the vector.

roystgnr:
Those are the only mistakes I caught.  I can take another pass after they (and anything @jwpeterson finds) are fixed; after that point we'll need to rebase a ton of history before we consider merging.

src/mesh/distributed_mesh.C
vikramvgarg: This PR has way too many embarrassing oversights. Should have looked for the cached methods.

vikramvgarg:
I've fixed the mistakes @roystgnr caught, @jwpeterson let me know if there are more issues. Else I can move on to the rebase.

src/mesh/distributed_mesh.C
```diff
@@ -77,41 +85,22 @@ DistributedMesh & DistributedMesh::operator= (DistributedMesh && other_mesh)
   this->UnstructuredMesh::operator=(std::move(other_mesh));
 
   // The remaining work to move a DistributedMesh
-  _is_serial = std::move(reinterpret_cast<DistributedMesh &>(other_mesh).is_serial());
-  _is_serial_on_proc_0 = std::move(reinterpret_cast<DistributedMesh &>(other_mesh).is_serial_on_zero());
+  _is_serial = reinterpret_cast<DistributedMesh &>(other_mesh).is_serial();
+  _is_serial_on_proc_0 = reinterpret_cast<DistributedMesh &>(other_mesh).is_serial_on_zero();
```
roystgnr, line 89: No need for any reinterpret_cast here on methods that are virtual in MeshBase.

src/mesh/distributed_mesh.C
```diff
@@ -77,41 +85,22 @@ DistributedMesh & DistributedMesh::operator= (DistributedMesh && other_mesh)
...
+  _max_elem_id = reinterpret_cast<DistributedMesh &>(other_mesh).max_elem_id();
 
-  _max_node_id = std::move(reinterpret_cast<DistributedMesh &>(other_mesh).max_node_id());
-  _max_elem_id = std::move(reinterpret_cast<DistributedMesh &>(other_mesh).max_elem_id());
-
-  _next_free_local_node_id = std::move(reinterpret_cast<DistributedMesh &>(other_mesh)._next_free_local_node_id);
-  _next_free_local_elem_id = std::move(reinterpret_cast<DistributedMesh &>(other_mesh)._next_free_local_elem_id);
-  _next_free_unpartitioned_node_id = std::move(reinterpret_cast<DistributedMesh &>(other_mesh)._next_free_unpartitioned_node_id);
-  _next_free_unpartitioned_elem_id = std::move(reinterpret_cast<DistributedMesh &>(other_mesh)._next_free_unpartitioned_elem_id);
+  _next_free_local_node_id = reinterpret_cast<DistributedMesh &>(other_mesh)._next_free_local_node_id;
```
roystgnr, line 94: Actually, I think there's no need for any casts here, even on the non-MeshBase data.  other_mesh is already an lvalue (despite being passed in as an rvalue; basically it didn't "have a name" when it got passed in but we "gave it a name" when we made it a parameter argument).

src/systems/system.C
```diff
@@ -288,7 +295,104 @@ void System::init_data ()
...
+  // magic, whether done directly or via EquationSystems::read()
+  //libmesh_assert(!this->is_initialized());
+
+  // First initialize any required data:
+  // either only the basic System data
+  if (_basic_system_only)
+    System::init_data();
+  // or all the derived class' data too
+  else
+    this->init_data();
```
roystgnr, line 310: What exactly is going on here?  You init_data(), but then the rest of this function is mostly a copy of System::init_data()?  I had hoped we'd be refactoring to avoid code duplication here, not ending up with code duplication and adding *code path* duplication to boot!

roystgnr (changes requested):
Getting closer, at least!

include/mesh/mesh.h
```diff
@@ -82,7 +82,16 @@ class Mesh : public DefaultMesh
...
    */
   Mesh & operator= (const Mesh &) = delete;
-  Mesh & operator= (Mesh &&) = delete;
+
+  /**
+   * Provides move functionality via base class move assignment operator.
+   */
+  virtual MeshBase & assign(MeshBase && other_mesh) override
+  {
+    *(libmesh_cast_ptr<DefaultMesh*>(this)) = std::move(*(libmesh_cast_ptr<DefaultMesh*>(&other_mesh)));
```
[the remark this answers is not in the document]
vikramvgarg: We havent declared an =(&&) operator for Mesh, and it tries to ends up trying to call Mesh::=(&) which is marked deleted.

src/mesh/boundary_info.C
```diff
@@ -138,6 +138,72 @@ BoundaryInfo & BoundaryInfo::operator=(const BoundaryInfo & other_boundary_info)
...
+  _side_boundary_ids = std::move(other_boundary_info._side_boundary_ids);
+  _node_boundary_ids = std::move(other_boundary_info._node_boundary_ids);
+  _edge_boundary_ids = std::move(other_boundary_info._edge_boundary_ids);
+  _shellface_boundary_ids = std::move(other_boundary_info._shellface_boundary_ids);
+
+  _ss_id_to_name = std::move(other_boundary_info._ss_id_to_name);
+  _ns_id_to_name = std::move(other_boundary_info._ns_id_to_name);
+  _es_id_to_name = std::move(other_boundary_info._es_id_to_name);
+
+  //We need to rebuild the boundary side ids
```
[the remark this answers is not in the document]
vikramvgarg: I had implemented the copying of the boundary ids in UnstructuredMesh earlier, and failed to remember that all building boundary side ids entailed was assigning them to names.

src/systems/system.C
vikramvgarg: I forgot to remove the rest of the function after copying over from init and removing the out of place assert.

include/mesh/mesh.h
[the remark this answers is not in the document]
roystgnr: Actually - do we need this at all?  We have declared `=(&&)` for both DefaultMesh options, so if we don't override assign here then it should just call the DefaultMesh override.  This is just a shim so we shouldn't have anything else to move here, right?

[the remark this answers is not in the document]
vikramvgarg: Yes, now that we are using the virtual function, we no longer need this.

include/mesh/mesh.h
```diff
@@ -82,7 +82,7 @@ class Mesh : public DefaultMesh
    * Copy and move assignment are not allowed.
    */
   Mesh & operator= (const Mesh &) = delete;
-  Mesh & operator= (Mesh &&) = delete;
+
```
roystgnr, line 85: @jwpeterson, is getting rid of this enough?  I'd think we'd need to explicitly set it `= default` too.

include/mesh/distributed_mesh.h
```diff
@@ -81,7 +81,27 @@ class DistributedMesh : public UnstructuredMesh
    * Copy and move assignment are not allowed.
```
roystgnr, line 81: Gah, I noticed that this comment was being obsoleted in replicated_mesh.h, but missed it here.

include/mesh/replicated_mesh.h
```diff
@@ -77,7 +77,27 @@ class ReplicatedMesh : public UnstructuredMesh
...
-  ReplicatedMesh & operator= (ReplicatedMesh &&) = delete;
+
+  /**
+   * Use the equality operator of the base UnstructuredMesh class
+   * if an rvalue argument is provided.
+  */
+  ReplicatedMesh & operator= (ReplicatedMesh && other_mesh);
+
+  /**
+   * Shim to call the move assignment operator for this class
```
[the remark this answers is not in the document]
roystgnr: Add to the comment here that it's really only supported for other_mesh of type ReplicatedMesh?

include/mesh/distributed_mesh.h
```diff
@@ -81,7 +81,27 @@ class DistributedMesh : public UnstructuredMesh
...
-  DistributedMesh & operator= (DistributedMesh &&) = delete;
+
+  /**
+   * Overloaded operator= will move contents of other_mesh to calling
+   * DistributedMesh object.
+   */
+  DistributedMesh & operator= (DistributedMesh && other_mesh);
+
+  /**
+   * Shim to call the move assignment operator for this class
```
roystgnr, line 92: This comment also needs an update and the shim also ought to be moved to the .C file

include/mesh/distributed_mesh.h
```diff
@@ -81,7 +81,27 @@ class DistributedMesh : public UnstructuredMesh
...
+  */
+  virtual MeshBase & assign(MeshBase && other_mesh) override
+  {
+    *this = std::move(*(cast_ptr<DistributedMesh*>(&other_mesh)));
+
+    return *this;
+  }
+
+  /**
+   * Move node and elements from other_mesh to this mesh.
```
roystgnr, line 102: Something here about other_mesh needing to be a DistributedMesh.

We actually could support moving nodes and elements from a ReplicatedMesh to a DistributedMesh or vice-versa, but until we do, the function signature here is misleading.

tests/mesh/boundary_info.C
```diff
@@ -260,7 +260,7 @@ public:
       bi.nodeset_name(0) = "ZERO";
       bi.nodeset_name(1) = "ONE";
 
-      BoundaryInfo bi2 {bi};
+      BoundaryInfo & bi2 (bi);
```
roystgnr, line 263: Yuck, how did I miss this before???  This commit isn't disambiguating a copy from a move, it's turning a copy into a no-op reference initialization!  That basically eliminates this unit test!

src/systems/system.C
```diff
@@ -288,7 +295,25 @@ void System::init_data ()
    this->init_matrices();
 }
 
+void System::reinit_mesh ()
```
roystgnr, line 298: So at this level we didn't need anything different than System::init() except for the removed assert()?

So:

A. Have we thought enough about how this should interact with derived systems?  I thought we'd discussed possibly wanting reinit_mesh() to be virtual, so derived systems would have a chance to react to a change in mesh?

B. Should System::init() just be that assert followed by a reinit_mesh()?  Based on your design right now the answer is "of course it should", but the answer to (A) could make this more questionable.

tests/mesh/mesh_assign.C
```diff
@@ -47,9 +47,9 @@ public:
   void testMeshMoveAssign()
   {
 
-    // We will do 4 tests:
-    // {ReplicatedMesh ,DistributedMesh} x {Mesh 2 from Memory, From File}
-    std::array<std::string, 2> mesh_types = {"replicated", "distributed"};
+    // We will do 6 tests:
+    // {Mesh, ReplicatedMesh ,DistributedMesh} x {Mesh 2 from Memory, From File}
```
roystgnr, line 51: Oh, this was a smart addition.  I should have noticed the omission.  I foolishly think of Mesh as a typedef (what we really wanted) rather than a shim class (what we implemented, for backward compatibility with forward declarations)...

include/mesh/mesh.h
roystgnr: I guess it's enough for our immediate purposes - since we only use the move assignment via the virtual assign() function, and that's implemented at the ReplicatedMesh and DistributedMesh level so it explicitly calls those move assignment operators, and Mesh is a shim with no other data to move, this works.  And the new unit tests confirms that.

It might still be a good idea to add an = default here just as a courtesy to anyone who for some reason wants to do a move assignment of Mesh directly.  Or it might not, since that'd just be one more thing to add extra test coverage for...

jwpeterson: Yeah if I recall correctly, the compiler doesn't auto generate the special functions if you have customized any of them. I'd say even if that wasn't the case it would be better to explicitly document that are using default move assignment.

src/systems/system.C
vikramvgarg: > So at this level we didn't need anything different than System::init() except for the removed assert()?

Yes, this seems to be the case. Calling solve() after moving behaves as expected.

> 
> So:
> 
> A. Have we thought enough about how this should interact with derived systems? I thought we'd discussed possibly wanting reinit_mesh() to be virtual, so derived systems would have a chance to react to a change in mesh?

Wouldnt this overlap with the user defined initialization function we call at the end of the reinit_mesh() function ?

> 
> B. Should System::init() just be that assert followed by a reinit_mesh()? Based on your design right now the answer is "of course it should", but the answer to (A) could make this more questionable.

The question is whether there is anything inside System::init_data that does not need execution when we only mean to reinit the mesh. I think there isnt, because almost every data structure owned by System relies on its mesh to provide the scaffolding.

roystgnr: After some thought - I don't think making reinit_mesh() virtual is an overlap here.  In some sense the "user" of a user_initialization() isn't the class designer, it's the class instantiator.  So we'd want the subclass designer to be able to overload reinit_mesh() without messing with any separate user initialization that's been attached.

Plus, what the hell, this is one function call per replacing-the-whole-freaking-mesh, so if it takes an extra nanosecond to go to the vtable that's worth some added flexibility.

vikramvgarg: Sounds reasonable.

include/mesh/mesh.h
```diff
@@ -79,10 +79,16 @@ class Mesh : public DefaultMesh
...
 
   /**
-   * Copy and move assignment are not allowed.
+   * Copy assignment is not allowed.
    */
   Mesh & operator= (const Mesh &) = delete;
 
+  /**
+   * Move assignment is allowed but mesh objects
+   * should be moved using the MeshBase::assign function.
```
roystgnr, line 88: I wouldn't say "should be" here.  This assignment operator ought to work perfectly.  We can't use it because we only have internal access to a MeshBase, but for anyone writing app code with access to the subclass there'd be no reason to prefer the virtual move over the standard move.

roystgnr (approved):

roystgnr:
Happy with the changeset as a whole; IMHO we just need one more rebase to clean up the history before merging.

roystgnr:
I think that's about as clean as I can get the history.  I've also got a saved branch with the original history in case I screwed anything up during all the rebases.  :sweat:

src/base/dof_map.C
```diff
@@ -254,17 +254,23 @@ void DofMap::set_error_on_constraint_loop(bool error_on_constraint_loop)
 
 void DofMap::add_variable_group (const VariableGroup & var_group)
 {
-  const unsigned int vg = cast_int<unsigned int>(_variable_groups.size());
+  // Ensure that we are not duplicating an existing entry in _variable_groups
+  if (std::find(_variable_groups.begin(), _variable_groups.end(), var_group) == _variable_groups.end())
```
jwpeterson, line 258: I was going to suggest that 
```
if (!std::any_of(_variable_groups.begin(), _variable_groups.end(), std::equal_to(var_group)))
```
was terser while doing the same amount of comparisons, but after typing it all out I'm not sure. I do think it reads nicer than "find == end" in general though.

src/mesh/distributed_mesh.C
```diff
@@ -58,7 +58,30 @@ DistributedMesh::DistributedMesh (const Parallel::Communicator & comm_in,
   _partitioner = libmesh_make_unique<ParmetisPartitioner>();
 }
 
+DistributedMesh & DistributedMesh::operator= (DistributedMesh && other_mesh)
+{
+  // Move assign as an UnstructuredMesh.
+  this->UnstructuredMesh::operator=(std::move(other_mesh));
+
+  // Nodes and elements belong to ReplicatedMesh and have to be
```
jwpeterson, line 66: Doc typo: `ReplicatedMesh` -> `DistributedMesh`

src/mesh/distributed_mesh.C
```diff
@@ -58,7 +58,30 @@ DistributedMesh::DistributedMesh (const Parallel::Communicator & comm_in,
...
+
+  // Handle remaining MeshBase moves.
+  this->post_dofobject_moves(std::move(other_mesh));
+
+  return *this;
+}
 
+MeshBase & DistributedMesh::assign(MeshBase && other_mesh)
+{
+  *this = std::move(*(cast_ptr<DistributedMesh*>(&other_mesh)));
```
jwpeterson, line 81: `cast_ref<DistributedMesh&>(other_mesh)`

seems clearer (assuming it compiles?) rather than going from pointer back to reference.

src/mesh/mesh_base.C
```diff
@@ -1519,4 +1561,46 @@ MeshBase::merge_extra_integer_names(const MeshBase & other)
...
+  _default_ghosting = std::move(other_mesh._default_ghosting);
+  _default_ghosting->set_mesh(this);
+
+  // _ghosting_functors
+  _ghosting_functors = std::move(other_mesh._ghosting_functors);
+
+  std::set<GhostingFunctor *>::const_iterator gf_begin_it = _ghosting_functors.begin();
+  std::set<GhostingFunctor *>::const_iterator gf_end_it = _ghosting_functors.end();
+
+  for (auto gf = gf_begin_it; gf != gf_end_it; ++gf )
```
jwpeterson, line 1582: Use a range-based for loop here?

src/mesh/mesh_base.C
```diff
@@ -1519,4 +1561,46 @@ MeshBase::merge_extra_integer_names(const MeshBase & other)
...
+    (*gf)->set_mesh(this);
+  }
+
+  // _shared_functors
+  _shared_functors = std::move(other_mesh._shared_functors);
+
+  std::map<GhostingFunctor *, std::shared_ptr<GhostingFunctor>>::const_iterator sf_begin_it = _shared_functors.begin();
+  std::map<GhostingFunctor *, std::shared_ptr<GhostingFunctor>>::iterator sf_end_it = _shared_functors.end();
+
+  for (auto sf = sf_begin_it; sf != sf_end_it; sf++ )
```
jwpeterson, line 1593: And range-based for loop here.

src/mesh/replicated_mesh.C
```diff
@@ -173,6 +173,38 @@ ReplicatedMesh::ReplicatedMesh (const UnstructuredMesh & other_mesh) :
...
+
+  // Handle those remaining moves.
+  this->post_dofobject_moves(std::move(other_mesh));
+
+  return *this;
+}
+
+MeshBase & ReplicatedMesh::assign(MeshBase && other_mesh)
+{
+  *this = std::move(*(cast_ptr<ReplicatedMesh*>(&other_mesh)));
```
jwpeterson, line 194: Again `cast_ref` seems like it would be cleaner if it compiles.

src/mesh/replicated_mesh.C
```diff
@@ -173,6 +173,38 @@ ReplicatedMesh::ReplicatedMesh (const UnstructuredMesh & other_mesh) :
...
+
+  return *this;
+}
+
+void ReplicatedMesh::move_nodes_and_elements(MeshBase && other_mesh)
+{
+  this->_nodes = std::move((cast_ptr<ReplicatedMesh*>(&other_mesh))->_nodes);
```
jwpeterson, line 201: Rather than repeated checked casts, just do one at the beginning of the function and use that reference.

src/mesh/unstructured_mesh.C
```diff
@@ -56,7 +56,6 @@ UnstructuredMesh::UnstructuredMesh (const Parallel::Communicator & comm_in,
 }
 
 
-
```
jwpeterson, line 59: This blank line deletion shouldn't be part of the diff.

tests/mesh/mesh_assign.C
```diff
@@ -0,0 +1,230 @@
...
+       // Now set the Dirichlet bcs: u = 1.0, when boundary id = 0 and 1.
+       // The function describing the boundary value
+       ConstFunction<Number> one(1.0);
+
+       // The boundary ids of the boundaries where the value is to be set.
+       const boundary_id_type all_bdry_ids[4] = {0, 1, 2, 3};
+       const std::set<boundary_id_type> bottom_right_bdry_ids(all_bdry_ids, all_bdry_ids + 1);
+
+       // The variables for which we are defining the bc.
+       std::vector<unsigned int> variables;
```
jwpeterson, line 120:
```
std::vector<unsigned int> variables = {0};
```

tests/mesh/mesh_assign.C
```diff
@@ -0,0 +1,230 @@
...
+          // based on its mesh
+          system.get_equation_systems().reinit_mesh();
+
+          // Reinit
+          system.get_equation_systems().reinit();
+
+          // Write out the mesh xda, this currently fails with DistributedMesh,
+          // running on more than 8 processors, when the mesh to be moved in is
+          // read from a file.
+          std::cout<<"Write out the mesh, read from memory."<<std::endl<<std::endl;
```
jwpeterson, line 187: We put spaces around "<<" operators and we should write to `libMesh::out` rather than std::cout

tests/mesh/mesh_assign.C
```diff
@@ -0,0 +1,230 @@
...
+
+      }// End loop over mesh two memory vs file case
+
+    }// End loop over type of mesh to be tested
+
+  } // End definition of MeshAssignTest
+
+}; // End definition of class MeshAssignTest
+
+CPPUNIT_TEST_SUITE_REGISTRATION( MeshAssignTest );
```
jwpeterson, line 230: Need a newline at end of file here.

tests/mesh/mesh_assign.C
```diff
@@ -0,0 +1,230 @@
...
+          system.get_equation_systems().reinit_mesh();
+
+          // Reinit
+          system.get_equation_systems().reinit();
+
+          // Write out the mesh xda, this currently fails with DistributedMesh,
+          // running on more than 8 processors, when the mesh to be moved in is
+          // read from a file.
+          std::cout<<"Write out the mesh, read from memory."<<std::endl<<std::endl;
+          system.get_mesh().write("mesh.out.memory.moved.xda");
```
jwpeterson, line 188: If you are writing a mesh from a unit test, you will need to clean it up, otherwise we will get `make distclean` errors. This file should be added to the `CLEANFILES` list in Makefile.am in the unit tests directory.

tests/mesh/mesh_assign.C
```diff
@@ -0,0 +1,230 @@
...
+          */
+
+          MeshTools::Generation::build_square(*mesh_two,
+                                              2, 2,
+                                              0., 1.,
+                                              0., 1.,
+                                              QUAD4);
+
+          // Uniformly refine mesh_two
+          std::unique_ptr<MeshRefinement> mesh_refinement(new MeshRefinement(*mesh_two));
```
jwpeterson, line 161: `libmesh_make_unique` should be used instead of `new`. Basically if you find yourself writing `new Foo`, you should stop and figure out whether it's really necessary. My goal is to have as few raw new/delete in the code as possible.

jwpeterson (changes requested):
This seems to be in much better shape, though I remain concerned about the maintenance burden of `MeshBase::operator=`. I think it would be good if there was a libmesh wiki page that covered everything you need to update if you add a new member variable to MeshBase, or something along those lines.

include/mesh/distributed_mesh.h
```diff
@@ -78,10 +78,25 @@ class DistributedMesh : public UnstructuredMesh
...
+
+  /**
+   * Shim to call the move assignment operator for this class
+  */
+  virtual MeshBase & assign(MeshBase && other_mesh) override;
+
+  /**
+   * Move node and elements from a DistributedMesh.
+   */
+  virtual void move_nodes_and_elements(MeshBase && other_mesh) override;
```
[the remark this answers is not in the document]
vikramvgarg: I cant think of a good reason to keep it public, @roystgnr ?

src/base/dof_map.C
vikramvgarg: I am definitely not up to speed with the latest C++ conventions. I would have thought that the find statement is more readable. I feel the same way about make_range actually. However, if none_of is more efficient/in tune with standards, I can change this.

tests/mesh/mesh_assign.C
vikramvgarg: I was just beginning to get the hang of new ....

vikramvgarg:
> This seems to be in much better shape, though I remain concerned about the maintenance burden of `MeshBase::operator=`. I think it would be good if there was a libmesh wiki page that covered everything you need to update if you add a new member variable to MeshBase, or something along those lines.

Sounds good. The overall conclusion seems to be that the move assignment is pretty much the default move, apart from the complications brought by BoundaryInfo and MeshBase coupling.

include/mesh/distributed_mesh.h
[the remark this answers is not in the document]
roystgnr: Nah; protected makes the most sense to me.

src/base/dof_map.C
roystgnr: >  latest C++ conventions

I'd make fun of you because `std::none_of` is C++11, but I'd forgot it existed until John pointed it out...

I prefer find() but I don't really like any of the options C++ standards give us here.  I want to just call a vector::contains(), but we don't even get set::contains() until C++20, nor string::contains() until C++23, and still no std::contains() in sight?

src/mesh/unstructured_mesh.C
roystgnr: Sorry, that one's my fault.  I caught and fixed one of these in boundary_info.h after I finished squashing commits but I didn't think to check elsewhere afterwards.

src/base/dof_map.C
jwpeterson: Yeah, the STL always uses ranges because that's the most general thing to do... but it ain't that pretty sometimes.

tests/mesh/mesh_assign.C
roystgnr: I had to help some users upgrade a MOOSE app last month to add DistributedMesh capability.  A MOOSE app that, to better match the style of a different application they were coupling to, used malloc().

It was difficult to make only the necessary updates and resist the temptation to prescribe a more deep-rooted solution.

https://www.youtube.com/watch?v=cDfqwHck2Qg

roystgnr:
> I remain concerned about the maintenance burden of MeshBase::operator=.

I've been trying to come up with a good way to add "forwards-compatible" error checking here.

Best idea I've had so far: add a dbg-mode-only code block in prepare_for_use, where we clone(), then clear(), then move-assign from the clone.  But then we'd still be out of luck if someone added a new member variable and didn't properly update the move-assign *or* the clear() methods.  I don't see how it's even possible in theory to handle that sort of bug with C++ support for reflection.

> I think it would be good if there was a libmesh wiki page

The spam threat wasn't worth it when we had a wiki, IIRC.

Perhaps we need a better workflow for https://github.com/libMesh/libMesh.github.io updates?  I find myself reluctant to update the non-autogenerated html files there because I know it won't change anything until the next manual sync anyway, but then I don't bother syncing because nothing but the Doxygen has been updated.

jwpeterson:
> The spam threat wasn't worth it when we had a wiki, IIRC.

The GitHub wiki is fine, you can't post to it unless you have permission to and I have never observed any spam on there.

roystgnr:
I'd forgotten you'd restarted that here.  We should add a link to it in the http://libmesh.github.io/ headers.

I'm not sure it'd be helpful for documenting this kind of maintenance issue, though.  When someone adds a new file and can't figure out why it's not being compiled, that visible failure provokes some web searching and eventually leads them to https://github.com/libMesh/libmesh/wiki/Adding-or-removing-source-files - but when someone adds a new member variable, and everything compiles perfectly, and the only problem is that they forgot to add it to the move assignment and now their new feature is invisibly incompatible with mesh swapping, what will lead them to wiki instructions on how to fix it?

jwpeterson:
Oh... I'm not worried about other people so much as I just want a checklist for myself. That's how I use the "adding and removing source files" anyway.

tests/mesh/mesh_assign.C
vikramvgarg: Basics are always the best.

jwpeterson:
@vikramvgarg I would suggest that you not run `git pull`... ever really, unless you set it up to only perform fast-forward merges. The branch history now contains both @roystgnr's rebased history, the original history which we don't want, and your new changes. I am going to try and fix it now.

vikramvgarg:
> @vikramvgarg I would suggest that you not run `git pull`... ever really, unless you set it up to only perform fast-forward merges. The branch history now contains both @roystgnr's rebased history, the original history which we don't want, and your new changes. I am going to try and fix it now.

Sorry about this John. I have set it up to do only ff-merges now using: git config --global pull.ff only

jwpeterson (approved):

PR #2877, opened 2021-03-08
Include unpartitioned elements in MeshBase::subdomain_ids

loganharbour:
Resolves https://github.com/idaholab/moose/issues/17245.

### DIFF
--- a/tests/mesh/mesh_subdomain_id.C
+++ b/tests/mesh/mesh_subdomain_id.C
@@ -1,0 +1,74 @@
+#include <libmesh/libmesh.h>
+#include <libmesh/elem.h>
+#include <libmesh/mesh.h>
+#include <libmesh/mesh_generation.h>
+
+#include "test_comm.h"
+#include "libmesh_cppunit.h"
+
+
+using namespace libMesh;
+
+class MeshSubdomainIDTest : public CppUnit::TestCase
+{
+  /**
+   * The goal of this test is to verify proper operation of the
+   * MeshBase::subdomain_ids() method.
+   */
+public:
+  CPPUNIT_TEST_SUITE( MeshSubdomainIDTest );
+
+  CPPUNIT_TEST( testUnpartitioned );
+  CPPUNIT_TEST( testMultiple );
+
+  CPPUNIT_TEST_SUITE_END();
+
+public:
+  void setUp() {}
+
+  void tearDown() {}
+
+  void testUnpartitioned()
+  {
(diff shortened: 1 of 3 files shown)


tests/mesh/mesh_generation_test.C
```diff
@@ -86,10 +86,17 @@ public:
...
+
     CPPUNIT_ASSERT_EQUAL(mesh.n_elem(), cast_int<dof_id_type>(n));
     CPPUNIT_ASSERT_EQUAL(mesh.n_nodes(),
                          cast_int<dof_id_type>((Elem::type_to_n_nodes_map[type]-1)*n + 1));
 
+   // Make sure that subdomain ids for unpartitioned elements are included
+   std::set<subdomain_id_type> ids;
+   mesh.subdomain_ids(ids);
+   CPPUNIT_ASSERT_EQUAL(ids.size(), (std::size_t)1);
+   CPPUNIT_ASSERT_EQUAL(*ids.begin(), (subdomain_id_type)0;
```
jwpeterson, line 98: This line (and a similar one below) is missing a trailing parenthesis?

loganharbour: Yeah - I pushed those in haste before testing and actually removed. 

I thought I could get away with testing this with the MeshTools::Generation meshes, but I forgot that we call prepare_for_use() at the end of all of those calls. Therefore, I wouldn't really be testing as we'd partition.

loganharbour:
I could possibly add a unit test that sets `skip_partitioning(true)` and uses `MeshTools::Generation` to build a simple mesh. Should be good then to make sure subdomain 0 exits in the mesh via the (all) unpartitioned elements.

Thoughts? Alternative?

roystgnr:
Code looks good.  Would it make sense to just create a mesh by hand in a unit test?  We've got a few like that in e.g. the slit mesh and the multi-dimensional mesh unit tests.

tests/mesh/mesh_subdomain_id.C
```diff
@@ -0,0 +1,72 @@
...
+  CPPUNIT_TEST_SUITE_END();
+
+public:
+  void setUp() {}
+
+  void tearDown() {}
+
+  void testUnpartitioned()
+  {
+    std::unique_ptr<UnstructuredMesh> mesh = libmesh_make_unique<ReplicatedMesh>(*TestCommWorld);
```
roystgnr, line 33: Why force ReplicatedMesh here?  If we use Mesh it'll test replicated or distributed depending on configure settings, and we'd expect it to work either way.

roystgnr (approved):

loganharbour:
Should be good to merge.

PR #2879, opened 2021-03-11
Add Elem::has_invertible_map() and overrides for some element types

jwpeterson:
This is a branch that I actually worked on last year, but it kind of got buried since then. I envision that the main use for this function will be debugging. For us it came up in a situation where we need to create an `Edge3` element "on the fly" out of three basically co-linear points. In one case when we did this, the mid-node was too close to one of the vertex nodes, and the map was therefore non-invertible, but we didn't have a way of checking that at the time. The result was that we were getting weird `inverse_map()` results -- not errors but wrong/unexpected values.

For 1D (Edge3 and Edge4) elements, I look at the vector 
```
N := d\vec{x}/ d(xi) := (dx/dxi, dy/dxi, dz/dxi)
```
and compute xi_m such that `N(xi_m) \cdot N(0) = 0`. If -1 <= xi_m <= 1, then the element map is not invertible.

For the Quad4, we look at the surface "normal" vector, 
```
N := d\vec{x} / d(xi) \cross d\vec{x} / d(eta)
```
and do basically the same thing.

For "quadratic" elements like Edge3, N varies linearly with xi, so we only need to look at the vertex values to determine whether it goes to zero anywhere in the element. For a cubic element like Edge4, N varies quadratically, so the min can occur internally to the element and we have to be a bit more careful with our check. 

So far this is only implemented for a few of the simpler elements but the goal would be to eventually enable it for all element types.

### DIFF
--- a/include/geom/cell_tet4.h
+++ b/include/geom/cell_tet4.h
@@ -136,6 +136,11 @@ public:
   virtual bool has_affine_map () const override { return true; }
 
   /**
+   * \returns \p true if the element has non-zero volume(), false otherwise.
+   */
+  virtual bool has_invertible_map(Real tol) const override;
+
+  /**
    * \returns \p true if the Lagrange shape functions on this element
    * are linear.
    */
--- a/include/geom/edge_edge2.h
+++ b/include/geom/edge_edge2.h
@@ -111,6 +111,11 @@ public:
   virtual bool has_affine_map () const override { return true; }
 
   /**
+   * \returns \p true if the element has non-zero volume(), false otherwise.
+   */
+  virtual bool has_invertible_map(Real tol) const override;
+
+  /**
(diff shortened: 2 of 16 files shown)


roystgnr:
Throw in some assertions about `this->mapping_type() == LAGRANGE_MAP` to the higher-order elements where your calculations assume that?  I don't think it'll be easy to come up with a case where the standard map is invertible but the Rational Bernstein map isn't or vice-versa, but I'm sure such cases exist.

roystgnr:
Scratch "to the higher-order elements".  You can get non-trivial differences from rational bases even for p=1, and at least in the bi+trilinear cases I think you can turn an invertible Lagrange element into a non-invertible element if you have a non-uniform weighting on the nodes.  So maybe best to just add that assertion everywhere until we have time to properly implement this for the rational-basis mappings.

PR #2880, opened 2021-03-13
Support for writing constant elemental vectors to elements in exofiles

### LINKED ISSUE #2873: Add support for writing constant elemental vectors to elements in exofiles
`ExodusII_IO::write_element_data` filters all variables which are not `FEType(CONSTANT, MONOMIAL)`, but `FEType(CONSTANT, MONOMIAL_VEC)` ought to allowed through as well.

`EquationSystems::build_variable_names()` can be modified to concatenate to a non-empty vector of variable names and called a second time with `CONSTANT MONOMIAL_VEC` types. `EquationSystems::build_parallel_elemental_solution_vector()` can be modified to recognize multiple DOF indices for a particular variable, much like how nodal vars are treated.

crswong888:
`ExodusII_IO::write_element_data` and `Nemesis_IO::write_element_data` now have full support for writing `CONSTANT, MONOMIAL_VEC` variables. I was pretty verbose in the code commentary because I found there to be a lot of rules for using the assosciated functions that developers ought to be aware of, especially in `EquationSystems::find_variable_numbers`.

I'm not sure how my changes affect other functions related to elemental data, such as those for reading and copying, if at all. Maybe some treatment of MONOMIAL_VECs ought to be made for these as well.

For the case of `ExodusII_IO::write_element_data_from_discontinuous_nodal_data`, I actually have no idea what this is for, but it does output constant monomials. I am assuming something can be done to handle constant monomial vecs too, so I left a comment about this.

(closes #2873) (refs idaholab/moose#17240)


crswong888:
@roystgnr when you get the chance could you please activate the tests here?

crswong888:
I can fix this :)

jwpeterson:
Hello @crswong888 I think because you put WIP in the title of your PR, it's not running any CI testing... just wondering whether that was intentional.

crswong888:
@jwpeterson yep, this was intentional. I recently rebased with the master branch and I didn't want to take up any resources to rerun the tests, but I do know for sure that there is one test from the libmesh repo that's failing. 

I've had to put this project on the back burner for the moment. Thanks for checking though. I will try to debug this at some point in the next few weeks.

crswong888:
@roystgnr @jwpeterson, please review this when you get the chance. Note that I have the solution for the failed MOOSE tests in one of my branches: [https://github.com/crswong888/moose/tree/exo-elevecs-17240](https://github.com/crswong888/moose/tree/exo-elevecs-17240).

roystgnr:
> Note that I have the solution for the failed MOOSE tests in one of my branches

Is that safe to get merged downstream first so we can get our tests here green, or do we have a chicken-and-egg problem of some sort?

crswong888:
@roystgnr The latter, it is not safe :(

crswong888:
So I've found a bug in my commit, and it turns out there's no chicken-egg situation with MOOSE after all. Tests are now passing without the need to provide backwards compatibility.

jwpeterson:
This is sufficiently complicated that I'd like to try it with our internal test suite before merging. 

Do any of the existing `vector_fe` examples in libmesh make use of this new capability? It would be great to have at least that or a new libmesh unit test, but it sounds like it will also be tested in MOOSE, so that's probably OK for me.

crswong888:
@jwpeterson There's currently nothing in either LibMesh or MOOSE that specifically tests writing constant vectors to elements. I will surely develop one in MOOSE once I add the compatibility. 

Actually, there is one input file in the MOOSE repo that creates a `MONOMIAL_VEC`: [step 4 of the workshop](https://github.com/idaholab/moose/blob/b5dcc0273d5f12d427c25e76faf65e2b6d7ef8d4/tutorials/darcy_thermo_mech/step04_velocity_aux/problems/step4.i#L26), but to circumvent the lack of support for vector output the variable is decomposed into individual monomials. I am tasked with updating the tutorial and that's how I encountered this problem.

There is only [this one example](https://github.com/libMesh/libmesh/blob/484c8652977b6504e1613633b4c7d87297f94957/examples/vector_fe/vector_fe_ex5/vector_fe_ex5.C#L136) that includes a `MONOMIAL_VEC` type in its equation system, but its order is first, and so it gets processed as a nodal variable.

roystgnr:
Would it make sense to make the default `FIRST` in vector_fe_ex5 into a variable, read that variable from the `input_file` GetPot, call `input_file.parse_command_line(argc, argv)` to allow us to override the default arguments, and add another `run_example` call to `vector_fe_ex5/run.sh` that overrides it with p=0?  Or would that be too much work for something that would technically cover the code path but wouldn't be doing a proper regression test?

crswong888:
@roystgnr I can give what you suggested a shot, sure!

roystgnr:
Sorry for the red herring.  I pinged @lindsayad about the formulation he used in vector_fe_ex5, and failure to converge for p=0 is the *expected* behavior there.

Perhaps what we ought to be doing is just a unit test with a project_solution and then Exodus output?

crswong888:
Ah. Well, I tried. And okay, I agree, a unit test sounds good. 

Is there some sort of diff test for Exodus files? That's really the main thing to check here.

crswong888:
...Also, I could merge the edits I made to example 5. The solution is bogus, but it's still a decent demonstration of how to set up the exodus object.

crswong888:
@roystgnr I developed the unit tests in pretty much the way we discussed. I tried to make use of the `project_solution()` and/or `project_vector()` methods along with a `FunctionBase<Number>` routine, but I ran into some problems. I was fighting with it for too long and ended up just employing a brute-force `current_local_solution->set()` method. I left a comment in the code about all this.

I am also pushing the stuff I tried with 'vector_fe_ex5' in a separate commit. If you want it, keep it! Otherwise, we can just discard that one.

roystgnr:
I like it.  Though I'm going to need to dig into those `project_solution()` issues.  Let's wait for @jwpeterson to give it a second look before considering merging, though.

examples/vector_fe/vector_fe_ex5/vector_fe_ex5.C
```diff
@@ -133,7 +137,24 @@ main(int argc, char ** argv)
...
+      equation_systems.build_variable_names(varnames);
+
+      // Create ExodusII object with an unambigous filename (indicating this solution is special)
+      ExodusII_IO exo_ptr(mesh);
+      exo_ptr.write("out_constant.e");
+      exo_ptr.set_output_variables(varnames);
+      exo_ptr.write_element_data(equation_systems);
+    }
+  else
+    ExodusII_IO(mesh).write_equation_systems("out.e", equation_systems);
```
lindsayad, line 157: I live behind the MOOSE wall by and large. Do we really not handle element vs. nodal writing internally? It seems like in an ideal world there would be no if/else branch here. (Other than to warn about the trivial solution)

crswong888: @lindsayad the `write_equation_systems()` method used here points to the namespace in `mesh/mesh_output.h`, which basically proceeds through the same steps I hard-coded here but for nodal data only (or element data written as nodal). I think this method ought to be made to handle constant element data as well, and I tried to do so, but it was tricky and I ultimately gave up on that.

roystgnr: Yeah, the API design in libMesh here is garbage.  Not my garbage, but I'm not going to name names; I'm certainly responsible for letting it slide at the time rather than insisting that write_equation_systems() be upgraded instead.  I think that a decade ago I still thought of Exodus as a viz format rather than as something we wanted to make into a first class restart option someday.

If the higher-order support going into Exodus (which is what limited my respect for it in the past; I couldn't even properly visualize my own dissertation work without adding fake h refinement to make up for the low p) is usable enough (and not just limited to use in conjunction with the new spline support) then we might have an opportunity to fix all this soon.

For now?  @crswong888 reiterated to me that we can just drop the vector_fe_ex5 commit, but I think the ugly API here makes it more worth leaving in, for didactic reasons, until we make Exodus `write_equation_systems()` more versatile.

jwpeterson:
Sorry @crswong888, this PR kind of fell off my radar. I will try and get back to you soon.

crswong888:
@jwpeterson no worries. Thank you!!

include/systems/equation_systems.h
```diff
@@ -351,28 +353,49 @@ class EquationSystems : public ReferenceCountedObject<EquationSystems>,
...
+   * should be sorted in accordance with ExodusII format, and so the
+   * developer just needs to know to separate dof_indices when
+   * accessing the system solution for vector variables.
+   *
+   * This function is designed to work for either a single type or a
+   * vector of types, but not both. This is because it can't simply be
+   * called a second time with another type because it filters (deletes)
+   * the names of those on the first call which used a different type.
+   * Thus, the 'types' argument is for the case where variables of
+   * multiple types are allowed to pass through.
```
roystgnr, line 373: Remind me - what was the expected behavior if we only pass in nullptr - was it "return all variable numbers for variables of every type"?  My fault for not documenting it initially, but we should get it squared away for the future, especially if we're making this function more complicated.

src/mesh/nemesis_io_helper.C
```diff
@@ -2660,77 +2675,90 @@ Nemesis_IO_Helper::write_element_values(const MeshBase & mesh,
           const std::vector<dof_id_type> & elem_ids =
             (it == subdomain_map.end()) ? empty_vec : it->second;
 
-          // Possibly skip this (variable, subdomain) combination
-          if (active_subdomains.empty() || active_subdomains.count(sbd_id))
+          // Possibly skip this (variable, subdomain) combination. Also check that there is at least
+          // one element on the sbd... Apparently, it is possible to have a sbd with zero elements,
+          // e.g., when running "adaptivity_ex3" in parallel with the 'dimension=1' argument.
+          if ((active_subdomains.empty() || active_subdomains.count(sbd_id)) && elem_ids.size())
```
roystgnr, line 2681: elem_ids is only going to have local elements, right?  So some processors are going to be entering this block now while others skip it?  But that's okay because we're operating only on `current_local_solution` inside the block, not calling anything `parallel_object_only()`?

src/systems/equation_systems.C
```diff
@@ -855,65 +856,104 @@ EquationSystems::build_elemental_solution_vector (std::vector<Number> & soln,
...
-  // If the names vector has entries, we will only populate the soln vector
-  // with names included in that list.  Note: The names vector and var_nums
-  // vector pairs are reordered alphabetically upon exiting this function.
-  std::vector<std::pair<unsigned int, unsigned int>> var_nums;
-  std::vector<std::string> filter_names = names;
-  bool is_filter_names = !filter_names.empty();
-
+  // Resolve class of type input and ensure that at least one of them is null
+  libmesh_error_msg_if(type && types,
+                       "Input either 'type' or 'types' in find_variable_numbers, but not both.");
```
roystgnr, line 878: Should this be libmesh_assert_msg(), since although we fear bad code is going to try to pass in two non-nullptr arguments here, there should be no way for a bad input file to do so?

src/mesh/nemesis_io_helper.C
```diff
@@ -2660,77 +2675,90 @@ Nemesis_IO_Helper::write_element_values(const MeshBase & mesh,
...
+                    std::vector<Real> local_soln_buffer(local_soln.size());
+                    std::transform(local_soln.begin(), local_soln.end(),
+                                   local_soln_buffer.begin(), [](Number n) { return n.real(); });
+                    ex_err = exII::ex_put_elem_var(ex_id,
+                                                   timestep,
+                                                   static_cast<int>(stride*(var_ctr+comp)+1),
+                                                   static_cast<int>(sbd_id),
+                                                   static_cast<int>(local_soln.size()),
+                                                   local_soln_buffer.data());
+                    EX_CHECK_ERR(ex_err, "Error writing element real values.");
```
roystgnr, line 2723: IIRC Exodus can be picky about 0-length data in some APIs, but I can't remember whether the preferred solution is "make sure you write a 0-length vector" or "make sure you *don't* write a 0-length vector", nor can I remember which APIs were the fragile ones.

Could we get some test coverage of this case, perhaps by just assigning every element in your unit test to a different subdomain, so that as soon as we test with `mpiexec -np 2` we'll end up with subdomains that have 0 elements on one processor or the other?

src/systems/equation_systems.C
```diff
@@ -855,65 +856,104 @@ EquationSystems::build_elemental_solution_vector (std::vector<Number> & soln,
...
+  // within that system. This way, we know where to find their data even after we sort them
+  std::vector<std::pair<unsigned int, unsigned int>> var_nums;
+
   for (const auto & pr : _systems)
     {
       const System & system = *(pr.second);
-      const unsigned int nv_sys = system.n_vars();
 
-      for (unsigned int var=0; var < nv_sys; ++var)
+      for (unsigned int var = 0; var < system.n_vars(); ++var)
```
roystgnr, line 910: Why change this around?  n_vars() isn't changing on us, right?

src/systems/equation_systems.C
```diff
@@ -855,65 +856,104 @@ EquationSystems::build_elemental_solution_vector (std::vector<Number> & soln,
...
+                      (std::find(name_filter.begin(), name_filter.end(), name) != name_filter.end()))
+                    component_names.push_back(name);
+                }
+
+              if (! component_names.empty())
+                names.insert(names.end(), component_names.begin(), component_names.end());
+              else
+                continue;
+            }
+          else /*scalar variable*/
```
roystgnr, line 935: Write "scalar-valued variable" here, so we don't get confused and think this branch has something to do with `SCALAR`?

No, we don't have good nomenclature here.

include/systems/equation_systems.h
crswong888: Yes, all variables listed in `names` no matter there type, would be returned with their numbers. Similarly, if the `names` list is empty, all variables of specified `type` OR `types` no matter what there name is, would be returned.

Thus, if `names` is empty and neither `type` nor `types` is specified, every single variable in the system gets returned.

crswong888: This was how it worked before, the only difference is not its possible to provide a list of types so that this function need only be called once. Awhile ago we discussed whether we should make `type` a vector by default, but we were concerned we'd be break people's code if we did that.

src/mesh/nemesis_io_helper.C
crswong888: I think so. `elem_ids` comes from the subdomain object, and we force it to be an empty vector if nothing is found:

```
for (const int sbd_id_int : global_elem_blk_ids)
  {
    const subdomain_id_type sbd_id =
      cast_int<subdomain_id_type>(sbd_id_int);
    auto it = subdomain_map.find(sbd_id);
    const std::vector<dof_id_type> empty_vec;
    const std::vector<dof_id_type> & elem_ids =
      (it == subdomain_map.end()) ? empty_vec : it->second;
```

So I think it is safe to just skip the whole routine if `elem_ids` is empty. That just means that there's no elements on the subdomain, which happens if there's more processors than elements in the global domain.

src/systems/equation_systems.C
crswong888: Yeah it's the developer's fault if this happens. So okay, assert makes more sense.

src/systems/equation_systems.C
crswong888: No it's not. I just thought it looked a bit cleaner this way and avoided having to declare a const unsigned int on each system iteration when the only thing we needed it for was setting up the variable loop.

src/mesh/nemesis_io_helper.C
crswong888: I don't think we would make it this far if there were zero elements on the subdomain.

src/systems/equation_systems.C
crswong888: Okay!

src/systems/equation_systems.C
roystgnr: That's actually good practice in general - if you declare a local variable, the compiler can just test against it.  E.g. if you don't declare a local variable here, then the compiler can't be sure that e.g. the `System::variable_type()` call later doesn't cause `System::n_vars()` here to change, so it actually has to call `n_vars()` on every loop iteration.

In this particular case it doesn't matter - `n_vars()` is inlinable, and even if it wasn't, that's a huge loop body with a small loop counter and one little function call per counter increment wouldn't make a dent in its performance.  But putting unnecessary-calls-that-the-compiler-can't-tell-are-unnecessary in loop tests is still a good habit to avoid getting into.  We actually have an IntRange class (and some helper functions like `make_range` and `index_range`) that let you get clean looking code without any inefficiency: here you could do `for (auto var : make_range(system.n_vars()))` and it would only evaluate `n_vars()` once.

src/mesh/nemesis_io_helper.C
roystgnr: Yeah, looks like we'd be short-circuiting at the `elem_ids.size()` above.  But that just pushes the potential failure point elsewhere; I'd still feel a lot safer with a test case that hits it and doesn't fail.

src/systems/equation_systems.C
crswong888: I understand, thanks for the wisdom! I'll go with the `for (auto var : make_range(system.n_vars()))` solution!

lindsayad:
> Actually, there is one input file in the MOOSE repo that creates a `MONOMIAL_VEC`: [step 4 of the workshop](https://github.com/idaholab/moose/blob/b5dcc0273d5f12d427c25e76faf65e2b6d7ef8d4/tutorials/darcy_thermo_mech/step04_velocity_aux/problems/step4.i#L26), but to circumvent the lack of support for vector output the variable is decomposed into individual monomials. I am tasked with updating the tutorial and that's how I encountered this problem.

This one input is now timing out in places like https://civet.inl.gov/job/817387/

lindsayad:
It looks like the problem here is that there is a CONSTANT MONOMIAL named `velocity_x` and then a `MONOMIAL_VEC` named `velocity`, such that we are indexing out of bounds here
```
              parallel_soln.set(ne * (var_ctr + comp) + elem->id(), sys_soln(dof_indices[comp]));
```
because `var_ctr` is getting too high. So I don't know if this is something we want to try and detect in libMesh and provided a helpful error message? Either way we will have to update the MOOSE test

lindsayad:
Ok I added a commit to https://github.com/idaholab/moose/pull/18705 which updates our MOOSE output to leverage the fact that we should treat `MONOMIAL_VEC` as elemental data now, and then I deleted the duplicate `MONOMIAL` variable. No regolding required!

roystgnr:
> So I don't know if this is something we want to try and detect in libMesh and provided a helpful error message?

Yikes, definitely.  I don't want to error out whenever someone adds variables like that (maybe warn?), because our internal conflict with renaming shouldn't be users' obligation in general, but if we get to the point where we're being asked to write output that relies on our renaming, screaming about our inability is better than trying to plow on and failing more obscurely.  Maybe we `libmesh_error_msg_if` when we see a duplicate in `EquationSystems::build_variable_names()`?

lindsayad:
> Maybe we `libmesh_error_msg_if` when we see a duplicate in `EquationSystems::build_variable_names()`?

Yes I think that's totally reasonable and a good solution. For MOOSE users certainly this is an easy fix once they know what the problem is.

roystgnr:
It's not sufficient to look for duplicates among the currently-requested variable names, is it?  The name list we get out of this is going to be handed to code like `find_variable_numbers()`, and that's going to look at every `System` in the `EquationSystems` and every variable in each, so if we have a duplicate with *anything* that's still going to screw us up, isn't it?

lindsayad:
I believe you are correct

PR #2884, opened 2021-03-17
Add a missing pointer to the DofMap's sparsity pattern.

### LINKED ISSUE #2883: Should SparsityPattern::Build be optional?
I have a bunch of code in IBAMR that looks like (from circa [here](https://github.com/IBAMR/IBAMR/blob/master/ibtk/src/lagrangian/FEProjector.cpp#L227))

```cpp
        std::unique_ptr<PetscMatrix<double> > M_mat(new PetscMatrix<double>(comm));
        M_mat->attach_dof_map(dof_map);
        M_mat->init();

        // Loop over the mesh to construct the system matrix.
        DenseMatrix<double> M_e;
        std::vector<libMesh::dof_id_type> dof_id_scratch;

        // assembly continues here...
```

This doesn't work anymore since with #2776 
```cpp
  const std::vector<numeric_index_type> & n_nz = this->_dof_map->get_n_nz();
  const std::vector<numeric_index_type> & n_oz = this->_dof_map->get_n_oz();
```
became
```cpp
  const std::vector<numeric_index_type> & n_nz = this->_sp->get_n_nz();
  const std::vector<numeric_index_type> & n_oz = this->_sp->get_n_oz();
```
It's not ideal that it looks like, for `PetscMatrix` at least, I need to provide both a `DofMap` and a `SparsityPattern::Build` when the `DofMap` knows everything. I'd rather not compute a bunch of redundant sparsity patterns if I don't need to.

My preferred fix would be to just make the `SparsityPattern::Build` optional (i.e., cover the case where `_sp` is null by falling back to the `DofMap`) - would y'all accept such a patch?

drwells:
The documentation states that this function will

    Set a pointer to the \p DofMap to use.  If a separate sparsity
    pattern is not being used, use the one from the DofMap.

Hence it should explicitly set the pointer to the sparsity pattern computed by the DofMap should one not already be set.

Fixes #2883 - the fix suggested by @roystgnr is much better than my original plan!

### DIFF
--- a/src/numerics/sparse_matrix.C
+++ b/src/numerics/sparse_matrix.C
@@ -56,6 +56,8 @@ template <typename T>
 void SparseMatrix<T>::attach_dof_map (const DofMap & dof_map)
 {
   _dof_map = &dof_map;
+  if (!_sp)
+    _sp = dof_map.get_sparsity_pattern();
 }
 
 


roystgnr:
> the fix suggested by @roystgnr is much better than my original plan!

Ironically, I'm now leaning in the other direction, as I just finished writing on the ticket.  Though, since I'm the one who wrote this bug and you're the one who caught it, unless anyone else chimes in I'll defer to you on what the final fix ought to be.

drwells:
I think older versions of libMesh didn't work in this use case without the `dof_map.compute_sparsity(mesh)` call so its not immediately clear to me what we can expect the `DofMap` to contain at this point. I'll take alook.

roystgnr:
I'd be happy with us making things *more* robust, not just regaining the previous level of robustness or lack thereof.  I'm in the middle of a Human Performance Improvement training session right now and they're really trying their best to impress upon us the moral that designing systems which work correctly if and only if the user makes zero errors is a no-no.

drwells:
With 1.6.1 and older, setting up a `PetscMatrix` required that the `DofMap` had computed its sparsity (so that the `n_nz` and `n_oz` arrays were available). I don't think there is a way to make things more robust (other than adding some assertions around pointers) since a `DofMap` doesn't always have that information available - i.e., we still need to call `dof_map.compute_sparsity()` at the correct point.

With recent patches we now have that `DofMap::get_n_nz()` et al just immediately return the value provided by the sparsity pattern it computes. If a user wants to use a *non-default* sparsity pattern for a particular matrix in the same System then they will need to provide that sparse matrix with a pointer to that sparsity pattern. Most of the time, though, users will just want to use the default. Hence, I believe this patch is right (and it matches what is in the documentation) since it only attaches the default sparsity pattern if one was not already attached (and a user can add their own either before or after if they want).

drwells:
Put another way - I don't know of a way to make this more robust (aside from fixing the bug, naturally) but the documented default seems pretty robust to me - we attach a default sparsity pattern which users can override instead of relying on users to attach a `DofMap` and then also attach a sparsity pattern.

roystgnr (approved):

roystgnr:
Thanks again for this!

This bug never made it into 1.6.1, right?  So we don't need to backport the fix to the 1.6 release branch?

drwells:
You're welcome!

> This bug never made it into 1.6.1, right? So we don't need to backport the fix to the 1.6 release branch?

 AFAICT there is no explicit usage of `PetscMatrix::_sp` in 1.6.1 so the answer is 'no'.

PR #2886, opened 2021-03-21
MeshBase & member in BoundaryInfo changed to MeshBase *

vikramvgarg:
To enable future implementation of a move assignment operation for MeshBase and derived classes, the BoundaryInfo member of
MeshBase should be moved completely. This in turn means that the MeshBase member of BoundaryInfo should be reassignable, a pointer rather than a reference.

### DIFF
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -882,9 +882,9 @@ private:
                       const std::set<subdomain_id_type> & subdomains_relative_to);
 
   /**
-   * The Mesh this boundary info pertains to.
+   * A pointer to the Mesh this boundary info pertains to.
    */
-  MeshBase & _mesh;
+  MeshBase * _mesh;
 
   /**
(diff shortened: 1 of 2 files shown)


roystgnr (approved):
Looks solid to me.  I'll merge at the end of the day unless John or anyone else chimes in.

jwpeterson:
I was expecting to see a setter of some variety, but I guess the point is that `MeshBase` is a friend and can change the internals directly, while others should not be allowed to do so?

roystgnr:
Yup.  I was pleasantly surprised that we actually kept our privates private here; there's no ```give_up_and_expose_raw_data_for_efficiency()``` API like we have in too many of our other classes.  (of course in this case that would be easy enough to change even if it had existed, and there's less reason for it to exist since the private member isn't anything complicated like a container)

jwpeterson (approved):

PR #2888, opened 2021-03-24
Read/write numeric vector projection/type info

lindsayad:
I ran into an issue where I was reading at a ghost dof location with a
vector that should be ghosted, but when reading `System` information
from xdr before this PR we add vectors as default `ParallelType PARALLEL` if the
vectors are not already present in the `System` object; the result was a
failed read and a triggered assertion in debugging modes/a segmentation
fault in optimized modes. So here I add the writing of both vector
projection and type so that users are safe to use ghosted vectors
without considering whether their vector has been created at the time we
read xdr `System` information. I know I'll need to add versioning
information because of this change.

Refs idaholab/moose#17399

### DIFF
--- a/configure
+++ b/configure
@@ -43912,7 +43912,7 @@ _ACEOF
 
 
 cat >>confdefs.h <<_ACEOF
-#define IO_COMPATIBILITY_VERSION "0.7.4"
+#define IO_COMPATIBILITY_VERSION "1.7.0"
 _ACEOF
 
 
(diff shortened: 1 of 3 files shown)


lindsayad:
Silly MOOSE showing the need for versioning the IO :laughing:

roystgnr:
Ha!  I was actually kind of surprised you put this up before adding versioning, but I didn't want to say anything because I really wanted to make sure our CI failed.  It's a pretty damning indictment of the libMesh "make check" that that phase passed.

jwpeterson:
Sounds like `make check` is pretty "robust" :wink:

lindsayad:
No matter what you throw at it, it passes! Yea sounds like a feature to me :smile:

roystgnr (approved):

jwpeterson (approved):
Looks good to me. I guess at some point we should stop using `sprintf` with a fixed `buf` size of 80 in this function :laughing:

PR #2891, opened 2021-03-29
Added set_EIM_rhs_vec() to RBEIMEvaluation

dknez:
Added set_EIM_rhs_vec() to RBEIMEvaluation so that we can explicitly set the RHS vectors that we want to use in rb_eim_solves().


jwpeterson (approved):
I'll do a squash+merge once all the testing finishes

Issue #1456, opened 2017-10-17
Expand support for different vector finite elements


lindsayad:
@cticenhour Could you talk about what type of vector elements you need to solve your applications? We currently have NEDELEC_ONE elements. Do you need Raviart-Thomas, etc.?

cticenhour:
My application (edge plasma/heating antenna modeling for fusion plasma systems) is mostly concerned with electric field structures (where my boundary conditions concern themselves with fields tangential to surfaces/interfaces). So, my current weak form would be considered an H(curl) problem. Accordingly, most of the reading that I have done on vector elements in this subject so far leans heavily on first and second order Nedelec elements depending on the accuracy needed. I've even seen third order in practice, but mostly only as part of an error study, so I'm not sure how commonly that is used.  

Since Raviart-Thomas elements seem to be used for H(div) problems (where you'd care about the continuity/definition of the normal field component in your boundary conditions), I'm not as concerned with them at the moment. However, any modeling concerned with magnetic fields would benefit. The fusion systems I care about use magnetic confinement, so the ability to use Raviart-Thomas could be useful in future.

roystgnr:
We do have LAGRANGE_VEC, too!  It's the perfect tool for debugging a code that will eventually require NEDELEC, or for if you want to write the same code but you don't want any of that nasty solution projection or hanging node constraint support...

More seriously:

Raviart-Thomas was my first thought, too.

On the other hand, even with the Nedelec elements, we only support the first kind, and we only support first-order!  I'm afraid I'm unfamiliar with the pros and cons of the two Nedelec families, but I am familiar enough with p-refinement to suspect that p=1 is suboptimal for many of our users' problems.

pbauman:
Yeah, I only needed first order when I coded up NEDELEC_ONE what is now forever ago. What would really be nice is to just do hierarchical Nedelec elements (ala Demkowicz's book) and be done with it. Similar statement for R-T.

roystgnr:
I also strongly support the "hierarchical" suggestion, but didn't want to say so, since my motivation is "I still haven't coded up adaptive p refinement support for non-hierarchical elements and I would like to continue getting away with avoiding that."  Presumably Dr. Demkowicz had more mathematical and less self-serving justification for such elements, though, so maybe they're really worth going with.

lindsayad:
@pbauman @roystgnr Do you guys have a text you would recommend from which to implement RT?

PR #2893, opened 2021-03-31
Add shape functions for NEDELEC_ONE FEs on TET10 mesh elements

alexanderianblair:
I've added a set of shape functions for NEDELEC_ONE elements on  TET10 meshes in 3D, based on the implementation in MFEM [here](https://mfem.github.io/doxygen/html/fe_8cpp_source.html), working to extend the range of FEs available in libMesh as in #1456. I've also extended the vector_fe/vector_fe_ex4 example to test the curl-curl problem on TET10s, just as was previously done on HEX20 and HEX27 mesh elements in 3D in #130.

There does appear to be some additional artifacts in the Paraview visualisations for these examples over each element, but H_curl error estimates seem reasonable compared to the case for hexehedral meshes (and solutions appear to converge to the right values as mesh size decreases). 

Any comments/recommendations welcome!

### DIFF
--- a/examples/vector_fe/vector_fe_ex4/run.sh
+++ b/examples/vector_fe/vector_fe_ex4/run.sh
@@ -10,6 +10,9 @@ example_name=vector_fe_ex4
 # preconditioners like ILU don't all work very well.  We currently
 # force Jacobi in these examples.
 
+options="element_type=TET10 -pc_type jacobi"
+run_example_no_extra_options "$example_name" "$options"
+
 options="element_type=HEX20 -pc_type jacobi"
 run_example_no_extra_options "$example_name" "$options"
 
--- a/examples/vector_fe/vector_fe_ex4/vector_fe_ex4.C
+++ b/examples/vector_fe/vector_fe_ex4/vector_fe_ex4.C
@@ -76,9 +76,9 @@ int main (int argc, char ** argv)
     command_line_value(std::string("element_type"),
                        std::string("HEX27"));
 
-  libmesh_error_msg_if(elem_str != "HEX20" && elem_str != "HEX27",
+  libmesh_error_msg_if(elem_str != "TET10" && elem_str != "HEX20" && elem_str != "HEX27",
                        "You entered: " << elem_str <<
-                       " but this example must be run with HEX20 or HEX27.");
+                       " but this example must be run with TET10, HEX20 or HEX27.");
 
   MeshTools::Generation::build_cube (mesh,
                                      grid_size,
--- a/src/fe/fe_interface.C
+++ b/src/fe/fe_interface.C
@@ -2536,6 +2536,7 @@ unsigned int FEInterface::max_order(const FEType & fe_t,
         case TRI6:
         case QUAD8:
         case QUAD9:
+        case TET10:
         case HEX20:
         case HEX27:
           return 1;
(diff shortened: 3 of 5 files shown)


jwpeterson:
Thanks, I will start the testing momentarily and send you an invite to join a group of developers that gets their PRs tested automatically.

jwpeterson:
It might be good to also add a comment somewhere with a reference book or paper that these are from. I looked at the MFEM source but did not see an obvious reference there.

alexanderianblair:
Sure, I'll add one in - the reference implementation is more specifically in mfem::Nedelec1TetFiniteElement::CalcVShape() in that source [here](https://mfem.github.io/doxygen/html/classmfem_1_1Nedelec1TetFiniteElement.html#a0f99dd4510c0a34894ad9e81e5dd1fd7).
Worth noting that MFEM has a slightly different node numbering scheme to libMesh, but that should be accounted for already in this commit!

jwpeterson:
@pbauman Any comments on this? It extends your original work on the vector-valued FEs.

pbauman:
I don't have time to verify shape functions, but it looks like the test is being used, so if the correct convergence rate is observed, then probably are good. The test only prints out the error, I manually confirmed the convergence rate when I did the quads and hexes. So I would encourage doing that. Otherwise, I'm super happy someone completed what I never did. Thanks @alexanderianblair!

roystgnr (approved):

jwpeterson:
I ran vector_fe_ex4 with TET10s and got first-order convergence for both L2 and H(curl) error. I'm not sure whether these are the expected rates of convergence for the curl-curl problem, but with HEX20s for example I get 2nd-order convergence in L2.

vector_fe_ex4

pbauman:
First order in HCurl is expected. Never did dig deeper to understand why L2 got second order on the hexes. At any rate, if you're getting the right convergence order in HCurl, that's pretty good evidence.

alexanderianblair:
The first order H(curl) and L2 errors decreasing for the TET10s are expected, and are consistent with those seen in deal.ii [here](https://www.dealii.org/reports/nedelec/nedelec.pdf), FEMPAR [here](https://www.sciencedirect.com/science/article/pii/S096599781831113X) and referenced by FEniCS [here](https://fenicsproject.org/pub/book/book/fenics-book-2011-06-24.pdf). The first order H(curl) error is optimal as far as I'm aware (see, for example, [Monk](https://icerm.brown.edu/materials/Slides/tw-18-7/Finite_Element_Methods_for_Maxwells_Equations_%5D_Peter_Monk,_University_of_Delaware.pdf) and [Hiptmair at al.](https://www.sam.math.ethz.ch/sam_reports/reports_final/reports2009/2009-04_fp.pdf)). It's interesting that the L2 error decreases faster for HEX20s - as noted in the deal.ii [report](https://www.dealii.org/reports/nedelec/nedelec.pdf) above, whilst O(h) convergence is usually expected, the optimal convergence can be faster than that. I'm not sure why they seem to be giving second order here though either yet!

jwpeterson:
> I'm not sure why they seem to be giving second order here though either yet!

I guess it could be some fortuitous cancellation happening because of the uniform grid on a cube... anyway good to hear that the results are generally as expected. Thanks also for those links. I think it would be good to get those into the documentation somewhere, maybe just throw them in vector_fe_ex4.

jwpeterson:
I added those references to vector_fe_ex4 and merged. Thanks again!

PR #2894, opened 2021-04-01
Add global data structures for mesh subdomain and boundary ids

lindsayad:
This kind of information is used significantly in MOOSE where we have
historically unioned this information ourselves. However, I have a case
where I want to query boundary/subdomain information between when the
libmesh mesh is prepared and when we build these MOOSE global data
structures. Moreover, I think having this information available is of
general utility and belongs upstream. Refs my work for
idaholab/moose#17189.

### DIFF
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -731,12 +731,17 @@ public:
    * \returns A set of the boundary ids which exist on semilocal parts
    * of the mesh.
    *
-   * DistributedMesh-compatible code may need a set_union or other
-   * manipulations to work with sets of boundary ids which include ids
-   * on remote parts of the mesh.
+   * Code that wishes to access boundary ids on all parts of the mesh, including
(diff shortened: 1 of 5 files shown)


roystgnr:
My suggestion earlier wasn't "do extra global communication with every prep phase", it was "do extra global communication when the cache is first requested, then clear the cache at every prep phase".  Lazy evaluation, like we do with the point locator.  This isn't remotely as expensive as that, though, so I'm fine with this if @jwpeterson is.

lindsayad:
The 5 BISON failures are due to #2871

lindsayad:
@jwpeterson do you have opinions on this PR?

jwpeterson:
My opinion is... we will probably want to have a lot more global information eventually, right? Looks good to me though.

lindsayad:
I do not have any explicit approvals, but seemingly two implicit approvals. I'll wait a bit but it sounds like I might be able to click the big green button

jwpeterson (approved):

PR #2899, opened 2021-04-13
undef MAJOR_VERSION and MINOR_VERSION before including petsc.h in config tests

finkandreas:
I think the issue could be bigger than what I have fixed. LibMesh defines in `configure.ac` the defines `MAJOR_VERSION` and `MINOR_VERSION`. I have build PETSc with CUDA support, i.e. `petsc.h` pulls in cuda headers, which have an enum defined that looks like this:
```
typedef enum libraryPropertyType_t
{
        MAJOR_VERSION,
        MINOR_VERSION,
        PATCH_LEVEL
} libraryPropertyType;
```
This clearly clashes with `#define MAJOR_VERSION 1` that will be added in the conftest. I changed the test to undef MAJOR/MINOR_VERSION before including `petsc.h`, and it successfully finds PETSc now.
I think the issue could be bigger, because any 3rdparty library that pulls in CUDA headers and you do a compilation test at configure stage will suffer from the same problem.
Later `MAJOR_VERSION` is renamed to `LIBMESH_MAJOR_VERSION`, and there is no problem at all. I think a clean solution should rename `MAJOR_VERSION` already at configure stage to `LIBMESH_MAJOR_VERSION`.

### DIFF
--- a/configure
+++ b/configure
@@ -48927,6 +48927,8 @@ ac_compiler_gnu=$ac_cv_cxx_compiler_gnu
           cat confdefs.h - <<_ACEOF >conftest.$ac_ext
 /* end confdefs.h.  */
 
+          #undef MAJOR_VERSION
+          #undef MINOR_VERSION
           #include <petsc.h>
           static char help[]="";
 
@@ -48983,6 +48985,8 @@ ac_compiler_gnu=$ac_cv_c_compiler_gnu
           cat confdefs.h - <<_ACEOF >conftest.$ac_ext
 /* end confdefs.h.  */
 
+          #undef MAJOR_VERSION
+          #undef MINOR_VERSION
           #include <petsctao.h>
           static char help[]="";
 
--- a/m4/petsc.m4
+++ b/m4/petsc.m4
@@ -97,6 +97,8 @@ AC_DEFUN([CONFIGURE_PETSC],
           CXXFLAGS="$saveCXXFLAGS $PETSCINCLUDEDIRS"
 
           AC_COMPILE_IFELSE([AC_LANG_SOURCE([[
+          @%:@undef MAJOR_VERSION
+          @%:@undef MINOR_VERSION
           @%:@include <petsc.h>
           static char help[]="";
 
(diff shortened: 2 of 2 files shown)


jwpeterson:
>  I think a clean solution should rename MAJOR_VERSION already at configure stage to LIBMESH_MAJOR_VERSION.

I need to check how our automatic `LIBMESH_` prefixing works. If I recall correctly it just gets blanket applied to all the `AC_DEFINES`, so I'm not sure if it's possible to treat MAJOR_VERSION and MINOR_VERSION differently.

jwpeterson:
@finkandreas I sent you an invite to a group that gets their libmesh PRs tested automatically.

finkandreas:
@jwpeterson I joined the group, but I am a little bit uncertain what the comment on Precheck means? Should I get active in any way?

jwpeterson:
@finkandreas sorry I should have mentioned that the servers which run our CI are down for an extended maintenance currently. I'll start up the tests manually once they are back online.

roystgnr:
> If I recall correctly it just gets blanket applied to all the AC_DEFINES, so I'm not sure if it's possible to treat MAJOR_VERSION and MINOR_VERSION differently.

Blanket applied (via m4/autoconf-submodule/ax_prefix_config_h.m4), but more importantly applied only at the last step when generating libmesh_config.h.  I'm not sure how to do anything like proper namespacing earlier.  I do like the workaround here for this specific case, if we can't come up with anything more general.

jwpeterson:
> I do like the workaround here for this specific case, if we can't come up with anything more general.

Agreed, I think since this problem is limited to configure tests, it makes sense to limit the fix to configure tests as well.

jwpeterson:
I will run ./bootstrap before manually merging once all the tests pass...

jwpeterson:
Hmm... I guess there's no point in letting the tests run without a bootstrap commit, since nothing has changed in the configure script. I'll just push the bootstrap commit to this branch...

jwpeterson:
This fix should also be cherry-picked onto branch_v1.6.2

roystgnr (approved):

PR #2900, opened 2021-04-14
SIDE_HIERARCHIC FE type, tensor product implementations

roystgnr:
This adds an FE space which looks like the standard HIERARCHIC shapes on element sides, but which is discontinuous from side to side and which is undefined on interiors.  The goal is to allow users to more efficiently support Hybrid FEM, Discontinuous Petrov-Galerkin, Finite Volume implementations with flux variables, basically any formulation which includes some variables solely supported on interfaces between elements.

Also added: a ton of tests for side variables, a permutation capability to give us better test coverage of odd configurations, and a ton of tests for those permutations.

We still have a lot of work yet to come (the higher-order HEX27 shapes are broken, and triangle support isn't yet implemented, and this really deserves its own example code), but some of the most urgent work is going to have other prerequisites (we need mid-face node options for 3D elements with triangular faces, including a PRISM20 and a TET14), and this is already at the point where it's sufficient for some users' short-term needs.  I've disabled the broken cases for now, so if anyone tries to use a broken high order on a HEX27 they'll get an exception thrown rather than an incorrect result.

I'm hoping I'll have the broken cases working by the time our CI is back online, but we'll see.  The problem with bicubics on HEX27 sides is obvious but hard to fix, and I haven't even figured out *how* biquadratics on HEX27 sides are broken.

Pinging @lindsayad and @YaqiWang


src/geom/cell_pyramid.C
```diff
@@ -46,7 +46,8 @@ const Real Pyramid::_master_points[14][3] =
     {0, -0.5, 0.5},
     {0.5, 0, 0.5},
     {0, 0.5, 0.5},
-    {-0.5, 0, 0.5}
+    {-0.5, 0, 0.5},
+    {0, 0, 0}
```
jwpeterson, line 50: Nice catch. I will cherry-pick this onto [branch_v1.6.2](https://github.com/libMesh/libmesh/tree/branch_v1.6.2) as well.

src/quadrature/quadrature_simpson_3D.C
```diff
@@ -131,6 +131,95 @@ void QSimpson::init_3D(const ElemType, unsigned int)
...
+
+        _points[12](0) = -0.5;
+        _points[12](1) = 0.;
+        _points[12](2) = 0.5;
+
+        _points[13](0) = 0.;
+        _points[13](1) = 0.;
+        _points[13](2) = 0.;
+
+        // These are of dubious value since we can't integrate on the
```
jwpeterson, line 199: I just worry that anyone accidentally using this quadrature rule for integration will now get an extremely singular mass matrix (and other unexpected results) rather than an error message... is there no way we can warn about this somehow?

src/quadrature/quadrature_nodal_3D.C
```diff
@@ -103,11 +104,21 @@ void QNodal::init_3D(const ElemType, unsigned int)
...
     case HEX27:
+    case PYRAMID13:
+    case PYRAMID14:
       {
         QSimpson rule(/*dim=*/3, /*ignored*/_order);
         rule.init(_type, /*ignored*/_p_level);
         _points.swap (rule.get_points());
         _weights.swap(rule.get_weights());
+
+        // We can't do a proper Simpson rule for pyramids regardless
```
jwpeterson, line 115: Not sure I understand this, creating a `QSimpson` rule for a `PYRAMID13` a few lines prior to this will have already thrown a "Unsupported type" exception, so you should never get here...?

src/quadrature/quadrature_simpson_3D.C
jwpeterson: Hmm... I guess they will get an error when they `reinit()` due to `JxW` being zero/undefined at the apex node, so maybe that's good enough.

src/geom/cell_pyramid.C
roystgnr: Thanks!

src/quadrature/quadrature_simpson_3D.C
roystgnr: That's what I'm thinking.  And then if we ever *do* figure out how to ever avoid the undefined JxW (a special case for pyramids where the mapping function turns out to be linear and so the map derivatives at the apex are consistent??) then this will actually be useful for node-lumped integration too.

src/quadrature/quadrature_nodal_3D.C
roystgnr: We don't throw "Unsupported type" there anymore, we just return the nodal points (plus the 14th) with the dubious only-integrate-linears-exactly weights.

roystgnr:
I'm now passing tests in 3D too, even with p=3 and p=4, even if I bump up the number of test permutations ... which is kind of a disaster, because I could swear I don't yet have the gradient and hessian terms correct for the cubic and higher degrees of freedom, so my test coverage must be inadequate.

I'm going to leave p limited to 2 in 3D for now regardless of whether I can get those tests to pass with it or not.

roystgnr:
Even after I rotate the mesh so as to scramble the point-based node orderings, even if I temporarily bump up mesh resolution to test more combinations of permutations, I *still* can't get the higher-order tests to fail if I re-enable them.

At some point I'm going to need to just manually permute two neighbor HEX27s into a combination that I expect to fail, then test gradients on that, then if *that* still passes try it by hand and figure out whether I'm wrong about certain face-bubble-function derivatives still needing work ... but that's not going to be done soon.  For now I think this is ready to merge.

jwpeterson:
I think I have seen that `hfem.robin_displaced` test fail several times. I assume it's not related, but maybe that test should be disabled/fixed on Mac too...

roystgnr:
Yeah, either @lindsayad missed one when he was fixing the others, or commits crossed in the merge?  I put up a fix in https://github.com/idaholab/moose/pull/17623 this morning; pinging to see if someone can review it now.

jwpeterson (approved):
Looks good to me, although I see at least two good PRs here: one with the `Elem::permute` and related changes, and one with the `SIDE_HIERARCHIC` changes. It would be nice to separate that out, but the chances that we'll need to completely revert one part or the other is low.

Oh, and it would also be nice to have some tests we expect to fail actually fail... not sure what is going on there.

PR #2903, opened 2021-04-17
Reduced Basis change: Enable eim-best-fit with lookup table

dknez:
We previously did not support "EIM best fit" in EIM training when we have a "lookup table". This PR resolves that issue.


dknez:
I guess the "Test Mac" failures are unrelated to this PR?

roystgnr:
Those ought to be fixed by https://github.com/idaholab/moose/pull/17623 - nothing to do with this PR, and we can merge without rerunning them.

(Assuming this is good to merge - it's in a part of the reduced basis code that my recent foray didn't cover, so I'll take your word for it on quality unless John can double-check)

jwpeterson (approved):

PR #2904, opened 2021-04-19
Wrap Petsc include for dmimpl.h in ignore and restore warnings.

vikramvgarg:
While compiling libMesh with paranoid warnings and werror, I get an error in PETSc:
include/petsc/private/petscdsimpl.h:62:46:
error: cast from type ‘const void*’ to type ‘PetscHashFormKey* {aka _PetscHashFormKey*}’ casts away qualifiers [-Werror=cast-qual].

Using ignore and restore warnings to ignore this.

### DIFF
--- a/src/solvers/petscdmlibmeshimpl.C
+++ b/src/solvers/petscdmlibmeshimpl.C
@@ -21,12 +21,16 @@
 
 #ifdef LIBMESH_HAVE_PETSC
 
+#include "libmesh/ignore_warnings.h"
+
 // PETSc includes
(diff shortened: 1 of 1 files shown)


jwpeterson:
Sounds reasonable to me. What version of PETSc are you using? If it's relatively new then you might want to raise this issue in their repo so they can fix the warning in upstream.

vikramvgarg:
> Sounds reasonable to me. What version of PETSc are you using? If it's relatively new then you might want to raise this issue in their repo so they can fix the warning in upstream.

The version is 3.15.0.1, I will send then a message about the issue.

PR #2910, opened 2021-04-27
Officially deprecate UniquePtr compatibility shim

jwpeterson:
It's long past time we did this. I plan to remove it soon after the next release.

### DIFF
--- a/examples/adjoints/adjoints_ex7/sigma_physics.C
+++ b/examples/adjoints/adjoints_ex7/sigma_physics.C
@@ -176,8 +176,7 @@ F(i) += JxW[qp] * ( ( -sigma * (grad_T * dphi[i][qp]) ) + (f * phi[i][qp]) );
   return compute_jacobian;
 }
 
-// Copy Constructor
-UniquePtr<DifferentiablePhysics> SigmaPhysics::clone_physics()
+std::unique_ptr<DifferentiablePhysics> SigmaPhysics::clone_physics()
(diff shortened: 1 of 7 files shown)


examples/adjoints/adjoints_ex7/sigma_physics.C
```diff
@@ -176,8 +176,7 @@ F(i) += JxW[qp] * ( ( -sigma * (grad_T * dphi[i][qp]) ) + (f * phi[i][qp]) );
   return compute_jacobian;
 }
 
-// Copy Constructor
-UniquePtr<DifferentiablePhysics> SigmaPhysics::clone_physics()
+std::unique_ptr<DifferentiablePhysics> SigmaPhysics::clone_physics()
 {
-  return UniquePtr<DifferentiablePhysics>(dynamic_cast<DifferentiablePhysics *>(this));
```
roystgnr, line 182: For future reference, we figured out that the reason this code wasn't breaking horribly was just that it was never getting called - this method gets used by `attach_physics()`, but never gets reached in code relying on `swap_physics()`.

In defense of @vikramvgarg , it looks like he just copied this from similarly atrocious code I had in diff_system.h, which was similarly never reached (in that case because those were `libmesh_not_implemented()` overrides of the pure virtual function; the bad code was just to avoid missing-return-value errors).

roystgnr (approved):

PR #2918, opened 2021-05-03
Add Elem::center_node_on_side

loganharbour:


src/geom/cell_prism18.C
```diff
@@ -1036,4 +1036,13 @@ Prism18::permute(unsigned int perm_num)
   }
 }
 
+
+unsigned int Prism18::center_node_on_side(const unsigned int side) const
+{
+  libmesh_assert_greater_equal (side, 1);
```
jwpeterson, line 1042: Hmm... it seems a little bit tricky to use this function. You'd first have to know that it's safe to call on what could be a base class `Elem*` pointer... I don't know your specific use case, but would you consider returning some pre-defined "invalid_id" constant instead of throwing an error?

loganharbour: Yeah, that was actually my biggest concern here. I was between that, or something like

```
bool T::has_center_node_on_side(const unsigned int) const;
```

Which would you prefer? I think I would actually prefer the extra method above.

loganharbour: Yeah, that was actually my biggest concern here. I was between that, or something like

```
bool T::has_center_node_on_side(const unsigned int) const;
```

Which would you prefer? I think I would actually prefer the extra method above.

loganharbour: I actually went ahead with returning invalid_uint otherwise. I think this is fine.

jwpeterson (approved):
Thanks!

roystgnr (approved):

PR #2921, opened 2021-05-05
Compilation fixes

roystgnr:
These fix test builds with --enable-parmesh and example builds with --disable-deprecated, respectively.


jwpeterson (approved):

jwpeterson:
I will also cherry-pick that `libmesh_cast_ptr` fix in misc_ex14 onto branch_v1.6.2. That was the only remaining use of libmesh_cast_ptr in the 1.6.x series.

PR #2922, opened 2021-05-06
Prerequest nothing on side FE in RB ex2

roystgnr:
This has been broken for forever, I just never noticed until now.  I
must have been forgetting to --disable-deprecated on builds with both
SLEPc and GLPK enabled, and without --disable-deprecated this behavior
is just an inefficiency that doesn't throw any errors.


jwpeterson (approved):

jwpeterson:
I'll also cherry-pick this one onto [branch_v1.6.2](https://github.com/libMesh/libmesh/tree/branch_v1.6.2).

jwpeterson:
you-get-nothing

PR #2925, opened 2021-05-12
FEMContext::active_vars()

roystgnr:
This lets us restrict a FEMContext's underlying FE objects to a subset of the variables in the system, not just for efficiency, but for correctness in the --disable-deprecated configuration under our newer "never try to reinit a FE you haven't requested data from" rules.

### DIFF
--- a/include/systems/fem_context.h
+++ b/include/systems/fem_context.h
@@ -65,16 +65,22 @@ public:
 
   /**
    * Constructor.  Allocates some but fills no data structures.
+   *
+   * Optionally specify a limited number of variables to be "active"
+   * and thus calculated on.  If \p active_vars is null then
+   * calculations will be prepared for every variable in \p sys.
    */
   explicit
-  FEMContext (const System & sys);
+  FEMContext (const System & sys,
+              const std::vector<unsigned int> * active_vars = nullptr);
 
   /**
    * Constructor.  Specify the extra quadrature order instead
-   * of getting it from \p sys.  Optionally specify a limited number
-   * of variables to be "active" and thus calculated on.  If
-   * \p active_vars is null then calculations will be prepared for
-   * every variable in \p sys.
+   * of getting it from \p sys.
+   *
+   * Optionally specify a limited number of variables to be "active"
+   * and thus calculated on.  If \p active_vars is null then
+   * calculations will be prepared for every variable in \p sys.
    */
   explicit
   FEMContext (const System & sys,
--- a/include/systems/fem_context.h
+++ b/include/systems/fem_context.h
@@ -994,16 +994,6 @@ public:
   void set_jacobian_tolerance(Real tol);
 
   /**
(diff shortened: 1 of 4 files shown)


roystgnr:
@jwpeterson - this is the FEMContext change I was telling you about a week ago.  I'm more confident in it now than I was then, but I'd still hope we can put it through the Akselos tests before merging.

jwpeterson:
OK, I will give it a try soon on our internal test suite.

src/systems/fem_context.C
```diff
@@ -1840,7 +1878,15 @@ void FEMContext::pre_fe_reinit(const System & sys, const Elem * e)
...
-            }
+            };
+
+          if (_active_vars)
+            for (auto v : *_active_vars)
+              init_localized_var_data(v);
+          else
+            for (auto v : make_range(sys.n_vars()))
+              init_localized_var_data(v);
+
```
jwpeterson, line 1889: Regarding our previous IM discussion, could we have a `make_range` that takes two args (untested pseudo code)
```
template<typename P, typename T>
IntRange<T>
make_range(P * ptr, T end)
{
  if (ptr)
    return {*std::begin(ptr), *std::end(ptr)};
  else
    return {T(0), end};
}
```
You might have to do some template magic to make this compile, but then you'd use it like. 
```
for (auto v : make_range(_active_vars.get(), sys.n_vars()))
  init_localized_var_data(v);
```
It seems a little specific to this use case to be a general utility, but might be good as an anonymous helper function.

jwpeterson (approved):
Our test suite doesn't seem to have any major issues with this PR so far, so I'm good.

include/systems/fem_context.h
```diff
@@ -993,16 +1005,6 @@ class FEMContext : public DiffContext
...
-  /**
-   * System from which to acquire moving mesh information
-   */
-  System * _mesh_sys;
-
-  /**
-   * Variables from which to acquire moving mesh information
-   */
-  unsigned int _mesh_x_var, _mesh_y_var, _mesh_z_var;
-
```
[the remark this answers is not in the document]
roystgnr: Underscore-prefixed variables that we've long had setters and getters for, too.  I intended to make this change years ago after any users had time to switch to the accessor calls, then just forgot until I ran across them last week.

src/systems/fem_context.C
roystgnr: We can't just return an IntRange here because that does contiguous incrementing.  We'd need a whole new class to return, one where the iterator checks the container pointer at every increment, and increments an internal value in the null case or a container iterator in the non-null case.  I think it'd be more of a pain than the existing code here, though I'd love it if I was wrong.

jwpeterson: Oh, yeah, duh. Not sure what I was thinking...

PR #2926, opened 2021-05-12
Mark virtual functions as such

jwpeterson:
Note that this should have no effect on the code, it is basically just documentation for humans.

### DIFF
--- a/include/fe/fe_base.h
+++ b/include/fe/fe_base.h
@@ -217,10 +217,10 @@ public:
     return dual_phi;
   }
 
-  void request_phi() const override
+  virtual void request_phi() const override
   { get_phi(); }
 
-  void request_dual_phi() const override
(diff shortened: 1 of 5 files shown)


jwpeterson:
I don't guarantee that I found all of these... it is kind of a hard thing to grep for.

roystgnr (approved):
Fine by me.  I like to hope that most of our users would be able to figure things out from `override` alone, though.

dschwen:
Isn't virtual completely redundant to override? We've had the discussion on MOOSE PRs and .. let's say opinions still differ.

jwpeterson:
Yeah, this is in response to a specific mailing list question where a user didn't realize that override -> virtual. Also we are like 98% consistent in labeling everything virtual so I'd rather fix the remaining 2% than go the other direction.

PR #2927, opened 2021-05-13
make periodic BC setup work for vector FEFieldType

gridley:
As it stands, if you try to set up periodic BCs for a vector variable, libMesh fails. I have tested this in MOOSE and confirmed this fix does the trick. I plan to add the input file [listed here](https://github.com/idaholab/moose/issues/17823) as a test in MOOSE once this gets incorporated to libMesh.

### DIFF
--- a/src/fe/fe_interface.C
+++ b/src/fe/fe_interface.C
@@ -2187,14 +2187,24 @@ void FEInterface::compute_periodic_constraints (DofConstraints & constraints,
                                                 const unsigned int variable_number,
                                                 const Elem * elem)
 {
-  // No element-specific optimizations currently exist
-  FEBase::compute_periodic_constraints (constraints,
-                                        dof_map,
-                                        boundaries,
-                                        mesh,
-                                        point_locator,
-                                        variable_number,
-                                        elem);
+  const FEType & fe_t = dof_map.variable_type(variable_number);
+  // No element-specific optimizations currently exist, although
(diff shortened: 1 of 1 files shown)


jwpeterson:
> libMesh fails

Oof. I will send you an invite to the group that gets their libMesh PRs tested automatically.

roystgnr:
Pinging @pbauman so he can marvel at what a simple fix this was.  I'm not sure whether to be super-impressed that we got the templated `FEGenericBase<>::compute_periodic_constraints` method to work perfectly in the vector case without even testing it, or just ashamed that we got only a handful of lines of code away from actually invoking the vectorized form of method and making vector PBCs work, and then **stopped**.

I say "we" because our imported svn->git history seems to think that [the templating here](https://github.com/libMesh/libmesh/commit/ee076c656eafb0cdca01cea705f32a9b3c86903f#diff-2d6a37b3b8b0beeac986ca1452f31915d9cf5cc976bfbdd978405c087eb71f74R1798) was my work, but I could have sworn Paul did all the heavy lifting on vector-valued FE, and the surrounding commits like b390f8e5dd947b602bd9906aefb9b12ef7705c03 and 3fd20a8380a661d21b9e7c536ba4076cc3cf6f1f seem to agree, so I doubt if ee076c656 was entirely my work.

roystgnr (approved):

PR #2928, opened 2021-05-17
Small bug-fixes

BalticPinguin:
Hi,
over the last months I have gathered a couple of minor issues, mostly one-liners I'd like to merge.

Is it more convenient if I submit them all together to reduce the number of PRs or is it better to keep them separate; so one can adjust them and discuss the background better?

This is a start with two issues.
Since you might wonder about the replacement of '0' with "0." in vsmoother: I got the error
````
    Assertion `_dist_norm > 0' failed.
    _dist_norm = 0.783142
    0 = 0
````
from this. Even though I don't understand this (probably an issue with compiler?), it was too easy fix...

### DIFF
--- a/src/mesh/mesh_smoother_vsmoother.C
+++ b/src/mesh/mesh_smoother_vsmoother.C
@@ -213,7 +213,7 @@ Real VariationalMeshSmoother::smooth(unsigned int)
   _logfile << "Saving Result" << std::endl;
   writegr(R);
 
-  libmesh_assert_greater (_dist_norm, 0);
+  libmesh_assert_greater (_dist_norm, 0.);
   return _dist_norm;
 }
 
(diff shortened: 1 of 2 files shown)


jwpeterson (approved):

jwpeterson:
Thanks, I recently came across a similar issue with the `libmesh_assert_greater/less` macro. This is caused by some internal casting between the different types that are passed to the macro when the comparison is done, and the literal "0" is treated as an int. Your fix here is the correct one.

jwpeterson:
> Is it more convenient if I submit them all together to reduce the number of PRs or is it better to keep them separate; so one can adjust them and discuss the background better?

Separate PRs or one PR is fine with me, as long as they are in separate commits.

jwpeterson:
Both of these fixes will be backported to 1.6.2 release.

PR #2929, opened 2021-05-17
remove duplicated loops over nodes and elements

BalticPinguin:
The first commit is simple; but in the second I am not completely sure if this is correct: There were two sets of `set_n_systems`-calls; while the first one was conditional, the other one wasn't; but from the comment below that part which I removed I'd assume that the second part should remain!?

### DIFF
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -130,7 +130,7 @@ const unsigned int Elem::type_to_n_nodes_map [] =
     18, // INFHEX18
 
     6,  // INFPRISM6
-    16, // INFPRISM12
+    12, // INFPRISM12
 
     1,  // NODEELEM
 
--- a/src/systems/equation_systems.C
+++ b/src/systems/equation_systems.C
@@ -146,28 +146,10 @@ bool EquationSystems::reinit_solutions ()
   const unsigned int n_sys = this->n_systems();
   libmesh_assert_not_equal_to (n_sys, 0);
 
-  // We may have added new systems since our last
-  // EquationSystems::(re)init call
-  bool _added_new_systems = false;
+  // And any new systems will need initialization
   for (unsigned int i=0; i != n_sys; ++i)
     if (!this->get_system(i).is_initialized())
-      _added_new_systems = true;
-
-  if (_added_new_systems)
-    {
-      // Our DofObjects will need space for the additional systems
-      for (auto & node : _mesh.node_ptr_range())
-        node->set_n_systems(n_sys);
-
-      for (auto & elem : _mesh.element_ptr_range())
-        elem->set_n_systems(n_sys);
-
-      // And any new systems will need initialization
-      for (unsigned int i=0; i != n_sys; ++i)
-        if (!this->get_system(i).is_initialized())
-          this->get_system(i).init();
-    }
-
+      this->get_system(i).init();
 
   // We used to assert that all nodes and elements *already* had
   // n_systems() properly set; however this is false in the case where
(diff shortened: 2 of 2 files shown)


src/geom/elem.C
```diff
@@ -130,7 +130,7 @@ const unsigned int Elem::type_to_n_nodes_map [] =
     18, // INFHEX18
 
     6,  // INFPRISM6
-    16, // INFPRISM12
+    12, // INFPRISM12
```
jwpeterson, line 133: Wow... not sure how that happened. I will backport this to the 1.6.2 release as well.

src/systems/equation_systems.C
```diff
@@ -146,28 +146,10 @@ bool EquationSystems::reinit_solutions ()
...
-      for (auto & elem : _mesh.element_ptr_range())
-        elem->set_n_systems(n_sys);
-
-      // And any new systems will need initialization
-      for (unsigned int i=0; i != n_sys; ++i)
-        if (!this->get_system(i).is_initialized())
-          this->get_system(i).init();
-    }
-
+      this->get_system(i).init();
```
jwpeterson, line 152: I'll let @roystgnr comment here, but it looks to me like you are removing some `set_n_systems()` calls entirely. I guess what you are saying is that this happens somewhere else instead, so this code is redundant?

src/geom/elem.C
roystgnr: Broken ever since Ben committed it in 2007, according to 2b80e9d296dca053a6661ab721db15850a28c7ff

Not sure how it survived this long, though.  I guess we don't have any Xdr/Checkpoint/DistributedMesh test coverage of InfPrism12?

src/systems/equation_systems.C
roystgnr: Looks redundant to me, actually.  We do that later sweep over `set_n_systems()` regardless, to handle the "users added nodes or elements after system initialization", and I guess whoever added that didn't realize that it made the `set_n_systems()` calls from the "new system has been added since last init" case superfluous.

The most efficient thing to do would be to ensure that new nodes and elements get `set_n_systems()` done as they're added, though.  We should probably add a `MeshBase::set_n_systems()` method, which would first do the loops below but would then cache the `n_sys` value and use it on new additions too, and then we'd *really* only need to do a full loop when new post-initialization Systems get added.

roystgnr: (That's an idea for another PR, though; this is still a big improvement even if it could be made better still)

src/systems/equation_systems.C
```diff
@@ -218,6 +200,7 @@ bool EquationSystems::reinit_solutions ()
 
         sys.prolong_vectors();
       }
+    //FIXME: why have this if we set it to 'true' anyway?
```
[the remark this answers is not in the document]
roystgnr: IIRC, once upon a time this was *supposed* to be a part of the forwards compatibility path towards requiring users to not handle the `refine_and_coarsen_elements()` calls themselves, but to merely flag elements and allow the EquationSystems to handle all mesh refinement and coarsening as it was doing projections.  I never figured out how to properly get that to work without breaking old codes.  It might be worth revisiting, though; the current system makes AMR more expensive than it should be.

src/geom/elem.C
BalticPinguin: I don't think there are more tests than the `inf_fe_radial_tests` for inifinite elements? 
By the way: Would you expect that checkpointing works for infinite elements? I once tried, but there were several errors (that above coming first); so I gave up thinking that this just isn't implemented...

roystgnr: I would not expect that.  I don't know of anything specifically precluding checkpointing working for infinite elements, so hopefully there aren't too many bugs to fix, but in my experience if something doesn't have complete test coverage there's never zero bugs.  The closest I've ever seen was the vector-valued FE periodic boundary code that we just turned on; it was utterly astonishing to see that that completely untested instantiation was actually correct and only needed to be enabled..

roystgnr (approved):

jwpeterson:
> roystgnr approved these changes

@roystgnr Did you want to merge in the FIXME comment as well? If not I can just remove that one line.

BalticPinguin:
No, this really was just a question; so it is clarified now.
If it is easy for you to remove, please go ahead; otherwise, I'd re-push without it.

jwpeterson:
It's easy to remove, I'll take care of it.

BalticPinguin:
:+1:  Thank you!

roystgnr:
> @roystgnr Did you want to merge in the FIXME comment as well? If not I can just remove that one line.

I would have been happy merging it - it's kind of embarrassing how long that situation has gone on, and more-embarrassing prodding might just push me over the top to a fix - but it's better removed.

PR #2930, opened 2021-05-19
Improve the performance of BoundingBox::contains_point 

fdkong:
by marking `is_between` as an inline function

```
3.28s  130:      if (bboxes[i_from].contains_point(*node + _to_positions[i_to]))

520ms  130:      if (bboxes[i_from].contains_point(*node + _to_positions[i_to]))
```

### DIFF
--- a/src/geom/bounding_box.C
+++ b/src/geom/bounding_box.C
@@ -27,7 +27,7 @@
 namespace libMesh
 {
 // Small helper function to make contains_point() more readable.
-bool is_between(Real min, Real check, Real max)
+inline bool is_between(Real min, Real check, Real max)
 {
   return min <= check && check <= max;
 }


jwpeterson:
Hmm... I'm kind of surprised that this does anything to be honest. My understanding was that the `inline` keyword is only a "hint" as far as [optimization](https://en.cppreference.com/w/cpp/language/inline) is concerned, and the compiler already "inlines" stuff that it thinks can/should be. 

> The original intent of the inline keyword was to serve as an indicator to the optimizer that inline substitution of a function is preferred over function call, that is, instead of executing the function call CPU instruction to transfer control to the function body, a copy of the function body is executed without generating the call. This avoids overhead created by the function call (passing the arguments and retrieving the result) but it may result in a larger executable as the code for the function has to be repeated multiple times.

> Since this meaning of the keyword inline is non-binding, compilers are free to use inline substitution for any function that's not marked inline, and are free to generate function calls to any function marked inline. Those optimization choices do not change the rules regarding multiple definitions and shared statics listed above.

I don't see any real harm in it for a function already in a .C file, so :+1:

jwpeterson (approved):

fdkong:
@roystgnr  was surprised as well. We even changed the compiler flags from "-O2" to "-O3", but the complier did not change anything. Marked the function as inline could make the code **_6 times faster_**. It is reproducible at this point. There is no easy way to check if the compiler actually treated that function as "inline". 

The compiler might malfunction since we had some weird compiler flags in some corners?

jwpeterson:
> There is no easy way to check if the compiler actually treated that function as "inline".

Good point. For something as simple as `is_between()`, I'd almost recommend using a macro... this would take the guesswork out of whether or not compiler X is going to honor the `inline` keyword at different levels of optimization. But let's go with the C++ approach for now since that has type checking, etc.

roystgnr:
> @roystgnr was surprised as well.

Not just surprised, dismayed.  I found it nearly impossible to believe that a modern compiler (gcc 8.4!) would fail to inline an anonymous-namespace function whose body is probably smaller than its own calling sequence, regardless of what we remembered or forgot to hint to it, even at high optimization levels.

> The compiler might malfunction since we had some weird compiler flags in some corners?

I did check that the `libmesh_CXXFLAGS='-O3'` configure trick I suggested was still working properly; judging by my O3 build's filesize changes (5% to 10% bloat from the extra inlining and unrolling) it is.

YaqiWang:
I am wondering if compiler can output a list of functions that are actually inlined. This small inlines can be a big killer.

roystgnr:
Looks like we can mark a function as inline, and then use `-Winline` to emit a warning when the compiler decides it's too big to be worth inlining regardless of what it was marked.

I wouldn't want to make that a default behavior, though; it's not a "real" enough (or a reproduceable enough) warning to safely use in --enable-werror builds.

PR #2935, opened 2021-06-01
$LIBMESH_BENCHMARK selection of benchmarking examples

roystgnr:
Putting this up for discussion and further editing for a while before we consider merging.

With this PR, whenever an environment variable `$LIBMESH_BENCHMARK` is defined, running `make check` in the examples directory will skip ordinary `run_example` lines and will instead only run `benchmark_example` lines based on the version (currently a number, but we could make this a string or pattern later) in the environment variable.

The chosen examples are scaled up to take in the neighborhood of a minute each to execute (which I'm hoping is enough to let actual code speed not be overwhelmed by startup time but not so much that iterating over the whole set of dozens of examples takes prohibitively long) on one core for me, but also not to spend too much time in the linear solver (so we're actually benchmarking libMesh execution) and also not to be inflexible in the linear solver (so we can run on multiple cores without worrying about something failing to converge).

I haven't actually limited the set to examples that are *scalable* in the linear solver, though - we can run the whole suite in parallel (at least up to 64 cores) to make sure we aren't missing any parallel-only hotspots, but in the future we'll want other benchmark sets that let us properly look at weak scaling too.

I'm tempted to see how far I can backport this, and even to move some tests from Version 1 to Version 2 if they can't be backported very far, so we can check for possible historical performance regressions.

### DIFF
--- a/examples/adaptivity/adaptivity_ex2/adaptivity_ex2.C
+++ b/examples/adaptivity/adaptivity_ex2/adaptivity_ex2.C
@@ -181,15 +181,18 @@ int main (int argc, char ** argv)
       libmesh_error_msg("ERROR: Initial timestep not specified!");
     }
 
-  // This value is also obtained from the command line, and specifies
-  // the number of time steps to take.
+  // This command line value specifies how many time steps to take.
   unsigned int n_timesteps = 0;
 
-  // Again do a search on the command line for the argument
   if (command_line.search("-n_timesteps"))
     n_timesteps = command_line.next(0);
   else
     libmesh_error_msg("ERROR: Number of timesteps not specified");
+
+  // This command line value specifies how far to allow refinement
+  unsigned int max_h_level = 5;
+  if (command_line.search("-max_h_level"))
+    max_h_level = command_line.next(n_timesteps);
 
 
   // Skip this 2D example if libMesh was compiled as 1D-only.
@@ -402,7 +405,7 @@ int main (int argc, char ** argv)
               // coarsened.
               mesh_refinement.refine_fraction() = 0.80;
               mesh_refinement.coarsen_fraction() = 0.07;
-              mesh_refinement.max_h_level() = 5;
+              mesh_refinement.max_h_level() = max_h_level;
               mesh_refinement.flag_elements_by_error_fraction (error);
 
               // This call actually refines and coarsens the flagged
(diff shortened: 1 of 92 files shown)


roystgnr:
My hope of backporting this has diminished - not because it's hard to backport this particular changeset, but because it's hard to compile old libMesh against new PETSc without backporting a bunch of compatibility patches for that first, and the patches in the reduced_basis tree aren't applying cleanly for me.  So we might be stuck looking for future regressions alone here, not past regressions too.

roystgnr:
> So we might be stuck looking for future regressions alone here, not past regressions too.

For future reference in case I lose my local benchmark results file, we haven't had any obvious regressions so far.  On the same workstation, I ran
`LIBMESH_RUN='mpirun -np 16' LIBMESH_OPTIONS='' command time make check -C examples/ METHODS=opt LIBMESH_BENCHMARK=1` today that I ran (among many others) last June, and the runtime (14'50") was 2% faster today.

roystgnr:
14'11" with my upcoming FE optimization branch ... down from 14'31"-14'47" with master ...  And `LIBMESH_BENCHMARK` doesn't even really hit the cases I was optimizing (IGA and 3D `LAGRANGE` on non-affine elements) very hard...

roystgnr:
13'49"-13'54" with master right now.

roystgnr:
We're at 12'48" with master (4 threads x 16 ranks) in a `-march=native` build (with #4500, that being what put the idea in my head, but the diff there shouldn't matter in this benchmark suite), and 13'05" with 1 thread.  I don't want to default our opt builds to something that might work on one computer and crash on another, but we might want to put something in our configure instructions explaining "here's how to get 7% speedup free"...

Edit: that's a `-march=native` libMesh build, but it was still against the same `-O3` PETSc settings I've been using for ages.  7% speedup on the whole benchmark probably means we're well into the double digits on the non-solver part of the benchmark.

roystgnr:
Finally got around to a `-march=native` PETSc build too ... and this is kind of wild, because it's much slower!  16'08".  Now I'm going to have to bisect both that option and the PETSc and libMesh upgrades I've made in the meantime...

Edit: libMesh upgrades actually were upgrades; backing off to just after the #4500 merge takes the time up to 16'15"

roystgnr:
Shoot, *and* the MPI change; I just switched from PETSc-downloaded MPICH to Ubuntu 26.04 OpenMPI for convenience.  That really shouldn't be the culprit when there are no fancy fabrics involved and every rank is on localhost, but that might be the culprit.

Edit: it's at least partly the MPI change.  Switching to Ubuntu 26.04 MPICH gets me to 14'18", and switching to PETSc-built MPICH gets me to 13'25".

PR #2937, opened 2021-06-03
Fix for deprecated warning in SLEPc not-quite-3.15

roystgnr:
PETSc 3.15 includes a non-release SLEPc 3.14.2 that has already
deprecated STPrecondSetMatForPC but that hasn't upgraded its
version number to let us switch to the non-deprecated version.  If
we're in that situation we need to ignore the deprecated warning.

This fixes --disable-deprecated builds for me.


roystgnr:
Cherry-picked onto the 1.6.2 branch

PR #2941, opened 2021-06-10
interior_parent fix on distributed refined multi-elem_dim meshes

roystgnr:
The fem_system_ex4 example (the one with the multiple manifold dimensions in the mesh) took 17 processors to trigger a failure from this, which I guess is why it never showed up on one of our 1-16 processor Civet sweeps?  Or possibly the failure only showed up for me because of some quirk of my compiler; this was a dangling-pointer bug and those are incredibly good at hiding.

### DIFF
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -798,6 +798,20 @@ const Elem * Elem::interior_parent () const
                   (interior_p == remote_elem) ||
                   (interior_p->dim() > this->dim()));
 
+  // If an element in a multi-dimensional mesh has an interior_parent
+  // link, it should be at our level or coarser, just like a neighbor
+  // link.  Our collect_families() code relies on this, but it might
+  // be tempting for users to manually assign something that breaks
+  // it.
+  //
+  // However, we *also* create temporary side elements, and we don't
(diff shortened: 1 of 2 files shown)


roystgnr:
Reverting this for now.  This fix appears to have led to (created? triggered?) a different problem when using *separate* boundary meshes, where the interior_parent() on a boundary mesh element isn't on the boundary mesh.

That problem should have been caught by one of our CI tests, but wasn't.  I'll look into that too.

Also, we appear to have a problem where if I type "I'm not going to work this weekend" into Slack, my hubris gets punished.

roystgnr:
The CI tests aren't running our unit tests right now.  I'm guessing when we moved moosebuild to new systems, the new environment didn't have cppunit installed in /usr?  We should probably add a `--enable-cppunit-required` configure flag, like we have for other technically-optional-but-often-essential dependencies like PETSc and MetaPhysical, so we can avoid this happening again in future upgrades.

As I type this @cticenhour tells me that the new moosebuild systems make cppunit available as a module, so we can simply manually load it in our test recipes.  I'll be AFK shortly but hopefully when I get back I'll find time to do that.

PR #2942, opened 2021-06-11
Begin contains_vertex_of with node equality tests

roystgnr:
This can be a little slower when testing 2-level-different AMR neighbors
and/or non-neighbors, but the extra speed when testing most neighbors
seems to make it a big win in my benchmarks.

Hoping @fdkong will see whether this gives him much improvement with his own distributed mesh refinement cases.


roystgnr:
Finally tried this out with one of the tests Fande recommended for me, and it makes DistributedMesh refinement about 150% (2.5x) faster in that one.  Still not nearly fast *enough* for us to consider the problem solved and done with, but too good not to merge in the meantime.

PR #2943, opened 2021-06-11
chunked_mapvector storage option

roystgnr:
A new chunked_mapvector container class, new abstraction to make it interchangeable with mapvector in DistributedMesh, and a configure option to choose between them as well as to tweak the new one.

I still want to benchmark the combination of this with custom allocators (though I don't expect much *combined* effect), but with around N=64 I'm seeing a few percent benchmark speedup over the mapvector-with-custom-allocator alone, which probably means much more speedup in the iterations and queries that are directly affected.

### DIFF
--- a/configure
+++ b/configure
@@ -1196,6 +1196,8 @@ enable_periodic
 enable_dirichlet
 enable_nodeconstraint
 enable_parmesh
+enable_distmesh
+with_mapvector_chunk_size
 enable_ghosted
 enable_node_valence
 enable_1D_only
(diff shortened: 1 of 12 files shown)


roystgnr:
Bah.  Somewhere in the cleanup for this I screwed up the linking; I'm getting `undefined reference to `libMesh::Predicates::evaluable<libMesh::chunked_mapvector<libMesh::Node*, unsigned long, 64u>::veclike_iterator>::operator()(libMesh::chunked_mapvector<libMesh::Node*, unsigned long, 64u>::veclike_iterator const&) const` and the like.  I was hoping to be done for the weekend but I guess I'll find time to fix this at some point.

roystgnr:
That force push might have fixed it.

roystgnr:
Nope.  It compiles now but isn't passing everything I can throw at it.  I swear this was working great in my original poorly-factored incarnation.  But I don't dare merge until I figure out what's not working perfectly now.

roystgnr:
The failure might be the fault of #2941 - if so then I'll rebase and try my own tests again after we revert that.

PR #2946, opened 2021-06-14
Weaken partitioner_test for SFC on 8+ processors

roystgnr:
Ideally we should strengthen the partitioners, but I don't want to find
out right now whether someone has a SFC-based parallel regression test
golded somewhere.


roystgnr:
Cherry-picked onto the 1.6.2 branch

PR #2947, opened 2021-06-14
Command-line override of default Metis-vs-Parmetis

roystgnr:
Using Parmetis for ReplicatedMesh doesn't seem to make any difference in performance on our benchmark set in #2935, but this might be useful or might make a performance difference in other contexts.

### DIFF
--- a/include/Makefile.in
+++ b/include/Makefile.in
@@ -649,6 +649,7 @@ include_HEADERS = \
         enums/enum_norm_type.h \
         enums/enum_order.h \
         enums/enum_parallel_type.h \
+        enums/enum_partitioner_type.h \
         enums/enum_point_locator_type.h \
         enums/enum_preconditioner_type.h \
         enums/enum_quadrature_type.h \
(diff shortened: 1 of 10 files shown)


roystgnr:
Well, the good news is that we have cppunit-based-unit-test coverage back, right?

jwpeterson (approved):
Seems reasonable to me, but since the command line option *only* applies to the default partitioner that is used by Distributed/ReplicatedMesh, it seems a little broadly named? I would suggest maybe `--default-mesh-partitioner={metis,parmetis}` but I'll leave it up to you.

roystgnr:
> I would suggest maybe --default-mesh-partitioner={metis,parmetis}

I love merging the option names into one, and pointing out that it just overrides the default and not anything specified in code, but your version is slightly verbose, so I figured `--default-partitioner` would be a little better.

Then I realized that the limitation to just metis-or-parmetis feels ridiculous when we're taking a string as argument, so I added a PartitionerType enum, string_to_enum support for it, and a static Partitioner::build() function accepting it, and now we have way more functionality than I ever intended to come out of a one-off "let me quickly make sure Parmetis isn't unexpectedly expensive" test.  :-D

Issue #2951, opened 2021-06-21
fe->reinit (elem, pts, weights) doesn't work properly


bizhishui:
Hi,

Suppose I want to compute some quantities (ex. the normal vector) on each vertex of the mesh (mesh elements are all triangles). Here is my code snippet

```
     std::vector<Point> q_points;
     std::vector<Real> weights;
    q_points.resize(3); weights.resize(3);
    q_points[0](0) = 0.0; q_points[0](1) = 0.0; weights[0] = Real(1)/6;
    q_points[1](0) = 1.0; q_points[1](1) = 0.0; weights[1] = weights[0];
    q_points[2](0) = 0.0; q_points[2](1) = 1.0; weights[2] = weights[0];

    FEType fe_type = dof_map.variable_type(x_var);
    std::unique_ptr<FEBase> fe (FEBase::build(2, fe_type));
    const int extraorder = 3;
    std::unique_ptr<QBase> qrule (fe_type.default_quadrature_rule (2, extraorder));
    fe->attach_quadrature_rule (qrule.get());

    const std::vector<Point> & q_pos = fe->get_xyz();

    for (const auto & elem : mesh.active_local_element_ptr_range()) {
      fe->reinit (elem, &q_points, &weights);
      std::cout <<  q_points.size() <<  std::endl;  //  3
      std::cout <<  q_pos.size() << std::endl;       //   6
    }
```
My question is why *reinit* doesn't work as it declared? After calling reinit, the real quadrature points aren't the ones passed through reinit.

Best regards,
Jinming

roystgnr:
I plugged your code fragment into a unit test in #2954, to give it the support framework (mesh, system, etc) necessary to run it ... and it passes for me.  Can you see if the tests there pass or fail for you, and if it fails can you figure out what's different between my case and yours?

bizhishui:
Thank you very much!
The unit test as you suggested in #2954  passes also for me, as summarised in unit_test_opt.txt
[unit_test_opt.txt](https://github.com/libMesh/libmesh/files/6701296/unit_test_opt.txt)

This means that the quadrature points are really reset by *reinit*, and will be independent with *extraoder*. So the assertion CPPUNIT_ASSERT_EQUAL(q_pos.size(), std::size_t(3)) passes.

In my case, I work with the _Tri3Subdivision_ element for a shell mesh (like a 2D curved surface in a 3D domain). Here is my code (newer to libmesh, this code may be bad)
```
void BEM::compute_normal()
[...]

PR #2954, opened 2021-06-22
Test custom FE reinit

roystgnr:
This fixes a bug (although *not* the bug it was supposed to catch, just an atavism in FEXYZ), and it adds some more coverage to our unit tests, so we could merge it as-is ...

But what I was *hoping* was that it would expose the bug reported in #2951, and at least on my system it doesn't; those assertions pass for me.


roystgnr:
Going to merge this now for the added coverage; it looks like the lack of subdivision-surface testing is what's got us failing to replicate the problem in #2951, but that's not an easy thing to add to this PR so we'll work on it separately later.

PR #2936, opened 2021-06-02
Use pool allocator in mapvector

roystgnr:
I've been wanting to create some clever custom data structure here for a
decade, and I still want to in the long term, but it seems that just
using a better allocator is enough to get my new benchmark tests running
2 or 3% faster.

PR #2955, opened 2021-06-25
Mimic stl and boost constructors in our allocators

lindsayad:
Neither boost  nor stl (at least not in libstdc++ implementation) use
`explicit` on their copy constructors. And more importantly for me to be
able to build on my machine I need to make the copy constructor
templated to allow copy construction from `(Fast)PoolAllocators` with
different `T`. This also mimics the constructors of boost and stl. Refs #2936

### DIFF
--- a/include/utils/pool_allocator.h
+++ b/include/utils/pool_allocator.h
@@ -67,7 +67,8 @@ public:
     boost::pool_allocator<T>()
   {}
 
-  explicit PoolAllocator(const PoolAllocator & o) :
+  template <typename U>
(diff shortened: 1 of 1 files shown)


roystgnr (approved):
Works for me.  Although I do wonder what the performance implications are of a bunch of copies being made.  Should we add move constructors for these?

lindsayad:
Maybe. There are no move constructors defined for `std::allocator` or for the boost allocators. Perhaps there's a good reason for that?

PR #2952, opened 2021-06-22
Try changing max order for L2 families

lindsayad:
Let's see what breaks!

PR #2957, opened 2021-06-28
More contains_vertex_of optimization

roystgnr:
I came up with this in a conversation with @jbadger95 a week ago, when he was asking about `contains_edge_of`.  Sadly I haven't yet actually improved the case *he* cared about (and probably won't soon; yes I'm still on vacation), but the question of "how much would this improve our point neighbor ghosting" kept nagging at me.

Turns out the answer is: our `LIBMESH_BENCHMARK` with `--enable-parmesh` gets 5% faster, and 3D refinement-heavy cases (like the uniform refinement tests @fdkong pointed me to) can get over 50% faster; all this is on top of the previous optimization in #2942, so that's something like a 5x speedup overall for distributed 3D mesh refinement.  I cannot fully express my embarrassment that we had performance that horrible in a common use case.  (We actually *still* have bad performance here, since `delete_remote_elements()` is half the cost of the solve in my test case, but it's no longer a dozen times the cost, so yay?)

### DIFF
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -438,9 +438,12 @@ public:
 
   /**
    * \returns \p true if a vertex of \p e is contained
-   * in this element.
+   * in this element.  If \p mesh_connection is true, looks
+   * specifically for containment possibilities of an element \p e
+   * that is connected to \p this via membership in the same manifold
+   * of the same mesh.
    */
-  bool contains_vertex_of(const Elem * e) const;
+  bool contains_vertex_of(const Elem * e, bool mesh_connection=false) const;
 
   /**
    * \returns \p true if an edge of \p e is contained in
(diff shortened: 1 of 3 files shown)


roystgnr:
Added a couple distributed CI sweeps; if those pass this should be good to merge.

roystgnr:
`refinement_type=p fe_family=L2_HIERARCHIC` failure - not sure if that's from this PR or from #2952.

roystgnr:
It's not from this PR.  Same failure with master.

fdkong:
Thanks for working on this. For uniform refinement, we often do not need original coarse meshes. In other words, parents can be deleted after children's creations. I think this can reduce the cost of finding neighbors. We might consider that as a valuable option when a general solution is impossible.

PR #2959, opened 2021-07-01
Fix distributed p coarsening assertion

roystgnr:
Not sure how we never managed to hit this overzealous assertion until
now, but one of the new tests in #2952 was hitting it on just a few
processors for me.

### DIFF
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -1546,6 +1546,15 @@ public:
   void hack_p_level (const unsigned int p);
 
   /**
+   * Sets the value of the p-refinement level for the element
+   * without altering the p-level of its ancestors; also sets the
+   * p_refinement_flag, simultaneously so that they can be safely
+   * checked for mutual consistency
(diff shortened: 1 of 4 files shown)


roystgnr:
If this passes the distributed sweep then I'll tack it onto branch_v1.6.2 too, but mostly just for completeness' sake, since it doesn't affect opt mode, and since I have trouble imagining a user who's doing adaptive p refinement yet is still using a 2020 release rather than the git head.

roystgnr:
Ha!  Flipping these operations just triggers the converse assertion.  Need to find a different fix.

roystgnr:
And ... the different fix isn't working for me either.  There's a more subtle underlying problem here; it even looks like I'm tripping a serious assertion now rather than an overzealous one.

roystgnr:
Take 3 (and final for today; about to be AFK again for a long while).

I have a good feeling about this one!

PR #2960, opened 2021-07-09
Add a new functor for sync element integers

YaqiWang:
As title.

### DIFF
--- a/src/parallel/parallel_ghost_sync.C
+++ b/src/parallel/parallel_ghost_sync.C
@@ -67,7 +67,7 @@ SyncSubdomainIds::SyncSubdomainIds(MeshBase & m)
 {}
 
 void SyncSubdomainIds::gather_data (const std::vector<dof_id_type> & ids,
-                                        std::vector<datum> & ids_out) const
+                                    std::vector<datum> & ids_out) const
 {
   ids_out.reserve(ids.size());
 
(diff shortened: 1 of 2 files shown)


src/parallel/parallel_ghost_sync.C
```diff
@@ -88,4 +88,32 @@ void SyncSubdomainIds::act_on_data (const std::vector<dof_id_type> & ids,
...
+      ids_out.push_back(elem->get_extra_integer(ind));
+    }
+}
+
+void SyncElementIntegers::act_on_data(const std::vector<dof_id_type> & ids,
+                                      const std::vector<dof_id_type> & integer_ids) const
+{
+  for (auto i : index_range(ids))
+    {
+      Elem * elem = mesh.query_elem_ptr(ids[i]);
```
roystgnr, line 113: Let's just make this `mesh.elem_ref(ids[i])`, which lets you delete a then-redundant assert and a layer of indirection below?

src/parallel/parallel_ghost_sync.C
```diff
@@ -88,4 +88,32 @@ void SyncSubdomainIds::act_on_data (const std::vector<dof_id_type> & ids,
...
+{}
+
+void SyncElementIntegers::gather_data(const std::vector<dof_id_type> & ids,
+                                      std::vector<dof_id_type> & ids_out) const
+{
+  ids_out.reserve(ids.size());
+
+  for (const auto & id : ids)
+    {
+      Elem * elem = mesh.query_elem_ptr(id);
```
roystgnr, line 102: Likewise here.  See the nearby SyncSubdomainIds functor for example.  If you want to demand that you always find the element you're querying, we have a shorter and less indirect API for that.

roystgnr:
Merging this regardless of the "Test Mac" failure - @loganharbour thinks that might be a recent CI client change, and I'm 99.999% sure the changes here could have nothing to do with that failure, nor could provided-but-unused code give us any subsequent failures so long as it compiles fine on Mac.

loganharbour:
> Merging this regardless of the "Test Mac" failure - @loganharbour thinks that might be a recent CI client change, and I'm 99.999% sure the changes here could have nothing to do with that failure, nor could provided-but-unused code give us any subsequent failures so long as it compiles fine on Mac.

I concur 👍🏼

jwpeterson:
FWIW the same Mac Test failure happened on #2961. :man_shrugging:

loganharbour:
Yeah - these failure are going to happen on all mac targets until we look into it more.

@jwpeterson you got lucky with the second run being on the last machine that had a display 😬

PR #2962, opened 2021-07-13
DofMap: add const DofConstraints accessor

jwpeterson:
We have a case where it would be useful to be able to quickly get a (const) reference to the `DofConstraintRow` for a given DOF, but currently the `DofMap` only provides the `constraint_rows_begin()` and `constraint_rows_end()` iterators. This PR adds a const accessor for the `DofConstraints` object itself.


roystgnr (approved):

roystgnr:
That failure looks like another "CI broken on headless Macs" case, so go ahead and merge when ready.

PR #2965, opened 2021-07-21
Update to add EIM observation points

dknez:
Enable the user to set "observation points" where we store the values of EIM basis functions, and allows inspection of the resulting EIM approximation at these points.

This does not change any existing behavior.


roystgnr:
John's on vacation, right?  Need a second pair of eyes on this?

jwpeterson:
I was just about to merge, but feel free to review if you like.

roystgnr:
Nah, go ahead - I just recalled you saying you'd be gone for a week in July and for some reason I thought this was the week.

PR #2963, opened 2021-07-21
Trigger a build

loganharbour:
Just switched libmesh to use new cloning methods, which sandbox things a bit more. Testing the changes.

PR #2966, opened 2021-07-21
--disable-amr fixes

roystgnr:
This should at least fix that one optional test from #2963.  I'm talking with @loganharbour to try to figure out how to make sure we don't wait this long between even oddball CI recipe invocations in the future.

Pinging @lindsayad to ask for a second set of eyes on my attempt to fix his GhostPointNeighbors code; everything else here is pretty trivial.

### DIFF
--- a/src/base/ghost_point_neighbors.C
+++ b/src/base/ghost_point_neighbors.C
@@ -125,6 +125,7 @@ void GhostPointNeighbors::operator()
               if (!equal_level_periodic_neigh || equal_level_periodic_neigh == remote_elem)
                 continue;
 
+#ifdef LIBMESH_ENABLE_AMR
               equal_level_periodic_neigh->active_family_tree_by_topological_neighbor(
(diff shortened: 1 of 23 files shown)


src/base/ghost_point_neighbors.C
```diff
@@ -125,13 +125,17 @@ void GhostPointNeighbors::operator()
...
+#ifdef LIBMESH_ENABLE_AMR
               equal_level_periodic_neigh->active_family_tree_by_topological_neighbor(
                 active_periodic_neighbors,
                 elem,
                 *_mesh,
                 *point_locator,
                 _periodic_bcs,
                 /*reset=*/true);
+#else
+              active_periodic_neighbors = { equal_level_periodic_neigh };
```
lindsayad, line 137: This looks like the right thing to me!

lindsayad (approved):

PR #2969, opened 2021-07-22
Skip ConstrainDirichlet on empty adjoint dirichlet boundaries

roystgnr:
Otherwise we can run into an assertion failure later.

This should fix the regression that was committed in #2761 and reported
in https://github.com/grinsfem/grins/issues/602


roystgnr:
Going to merge this right away; I'm not certain it's the *only* bugfix needed here but it's definitely a bugfix, it's not regressing tests, and it's a straightforward small change.

PR #2970, opened 2021-07-23
Fix all-static build issues in fparser_ad

dschwen:
Refs #2963

### DIFF
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -609,6 +609,17 @@ int ADImplementation<Value_t>::AutoDiff(unsigned int _var, typename FunctionPars
 }
 
 template<typename Value_t>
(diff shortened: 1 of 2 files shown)


dschwen:
```
/usr/bin/ld: cannot find -lXdmcp
```

@roystgnr / @jwpeterson this does not seem FParser related anymore. I was able to successfully build an all-ststic config with my patch on a Linux system (with a custom static PETSc build, like in the civet recipe).

roystgnr:
> this does not seem FParser related anymore.

Agreed.  I'd be happy merging this to fix up our compile-time failures in this config, and waiting on the link-time failures (which of our dependencies or their dependencies or whatever is trying to link to the X Display Manager control protocol!?!?!) until some other time.

dschwen:
I noticed the PETSc tests/examples require X11

PR #2971, opened 2021-07-25
Add override keyword in several places

pbauman:
As suggested by GCC's `-Wsuggest-override`. I want to call out 86f7175 since that changes `fparser`. I'm not sure if we forked `fparser` or not, so if not, we may want to exclude this commit from the PR and try to upstream it instead. This and libMesh/TIMPI#71 get all the libMesh-based warnings silenced when I build GRINS with `-Wsuggest-override`. I didn't drop this warning in libMesh yet so there may be others if we add this warning to libMesh proper.

### DIFF
--- a/contrib/fparser/fparser_ad.hh
+++ b/contrib/fparser/fparser_ad.hh
@@ -196,10 +196,10 @@ protected:
 
   // Exceptions
   class UnknownVariable : public std::exception {
-    virtual const char* what() const throw() { return "Unknown variable"; }
+    virtual const char* what() const throw() override { return "Unknown variable"; }
   } UnknownVariableException;
   class UnknownSerializationVersion : public std::exception {
-    virtual const char* what() const throw() { return "Unknown serialization file version"; }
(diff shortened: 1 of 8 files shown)


roystgnr:
We very much forked fparser, I'm afraid.  I think the whole of fparser_ad.hh is ours, even.

roystgnr (approved):
This is great.  Were these the only spots John had missed?  Should we add -Wsuggest-override to our --enable-paranoid-warnings builds?

pbauman:
>  Were these the only spots John had missed? 

I don't know, honestly. I only enabled this flag on the GRINS side and, thus, this PR reflects the warnings that showed up from libMesh there.

> Should we add -Wsuggest-override to our --enable-paranoid-warnings builds?

I certainly would like to see it added to the flags. Whether it's in the paranoid section or "normal" section or something else is up for discussion, I suppose. IMHO, this warning should just be on all the time, but I could see some folks disagreeing.

PR #2972, opened 2021-07-29
ExodusII v8.11

roystgnr:
I had been hoping that we could support the new IGA extensions to ExodusII without actually upgrading our decade-old ExodusII version, but no such luck.  The old APIs don't give us flexible access to global attribute vectors (even though they *should*, but that's a rant that'd have to go upstream and the fix would require us to upgrade anyway).

So, here's the latest ExodusII+Nemesis from a tagged SEACAS release.  In the long run we really ought to be supporting compilation against a third-party install of these things, but just adding the latest version to contrib/ for now seemed to be the minimally disruptive way to upgrade us to start.

The configure options here default us to v8.11 when they can (because new APIs yay), but fall back on v5.22 when they have to (which may be a *lot*, because v8.11 seems to require NetCDF4, which seems to require HDF5, which we don't autodetect so users have to specify --enable-hdf5 --with-hdf5=foo to get it).


jwpeterson:
I'm definitely in favor of this change. I tried to look back in the history to see what we did the last time... but Exodus apparently hasn't been updated since the change to automake, so it's well overdue!  We currently use `--enable-hdf5` in our builds, so I should be able to at least test that v8.11 is actually working for us, and report back here.

roystgnr:
> We currently use --enable-hdf5 in our builds, so I should be able to at least test that v8.11 is actually working for us, and report back here.

Thanks!  It's passing my tests still with --enable-hdf5, but I'd feel more confident with more than one positive result.

If your tests pass, and if the new APIs turn out to be suitable for my IGA Exodus stuff, then I'll merge.  I can see about getting some HDF5 coverage in CI after that; we'll need it by the time I have the IGA Exodus branch ready, since I'll definitely want the unit tests for that to be double-checked regularly.

YaqiWang:
Do they have any documentation on what is new in this new Exodus release?

roystgnr:
Not that I can find.  The docs' 02-Changes.txt chapter is a couple blurbs, added in 2016 and ambiguous about what earlier date range is being covered, and not updated since.

jwpeterson (approved):
I was able to build and run some of our internal tests with Exodus v8.11 and everything seems to basically be working fine. Removing the exodusII.h header from all of our includes is also a huge improvement; we should have done that change a while ago, but good to have it fixed now at least. 

One thing I discovered is that the "variables" section in the header of the exo file may be written in a different order with this newer version of Exodus... the standard exodiff utility doesn't seem to care about this, but we had some other comparison code that needed to be updated slightly to account for this non-essential difference between files.

roystgnr:
Just finished testing with the new APIs; they seem to be sufficient for everything I need to extract from these new IGA files.

Issue #2976, opened 2021-08-04
Centroid evaluation for certain element types is not accurate


YaqiWang:
Currently the centroid is evaluated as an arithmetic mean of composing vertices. This is correct for simplex, such as triangle or tetrahedra but not accurate for more complicated element shapes, for example quad. In my opinion, we should fix this.

roystgnr:
We should fix this, but we should fix the comments to match.

Before, the function didn't return what it's name said it does.  That's a bug IMHO.  (and a bug that's been there since d7a66045db68b999b3c590360f9ce632ec38c343, so I think this is a new and unbeatable record for Oldest LibMesh Bug Ever!).  But with your PR, it won't return (in the Quad4 case) what it's documentation said it does.  That's still a bug.  Let's update the comment in elem.h to the effect of "may return the element centroid, or may return the average of its nodes"?  Something weasel-wordy like that, so we can get around to updating hex and prism and 2nd order and such elements at some point in the distant future without risking breaking the user contract further.

YaqiWang:
;-) I was just throwing a rock there. Not expecting the commit to be merged by itself. IMO, we should do something similar in `Elem::centroid()` like what we did in `Elem::volume`, i.e. use quadratures for the evaluation. Then we can override each individual types with faster evaluations.

roystgnr:
That's exactly what @jwpeterson suggested - he points out that the linearity of the problem makes the quadrature-based solution surprisingly easy to optimize.

It still won't be nearly as fast as the current "`centroid()`", so we'll want to move that implementation to `vertex_mean()` or something, and transition over the many current use cases that don't require the new level of accuracy ... but at least we'll be able to say that `centroid()` really means *centroid*, at least up to quadrature error, and we can avoid any awful weasel-wordy "sometimes centroid but sometimes not" temporary definition.

jwpeterson:
I will work on adding this soon.

jwpeterson:
Fixed in #2978.

PR #2977, opened 2021-08-04
Add a protection when loading element integers from Exodus

### LINKED ISSUE #2975: Need a protection when loading element integers in Exodus file
Currently we are using elemental variables in Exodus files to store extra element integers. The variables are stored as real numbers (double), while the element integers are in type of `dof_id_type`. `dof_id_type` could be with 64 bits thus exceeds the limit a double can represent because double only has 53-bit precision. When converting a double to `dof_id_type`, we need a protection to prevent from misrepresentation from happening.

YaqiWang:
Close #2975.

And fix quad for #2976.


src/mesh/exodusII_io.C
```diff
@@ -286,7 +286,30 @@ void ExodusII_IO::read (const std::string & fname)
 
           // Assign extra integer IDs
           for (auto & id : extra_ids)
-            elem->set_extra_integer(id, std::round(elem_ids[id][elem->id()]));
+          {
+            Real v = elem_ids[id][elem->id()];
+            if (std::numeric_limits<Real>::digits > std::numeric_limits<dof_id_type>::digits)
```
roystgnr, line 291: No, no - we don't want to compare float/double/quad to int/long here, we want to see if v, specifically, is in the range of numbers that can be exactly represented as a double.

roystgnr: Oh, wait, I get it now - you're trying to short circuit entirely in the case where we have a large Real but a small dof_id_type.  Disregard the above - that's clever!

roystgnr:
Have you tested the new read code on your broken-with-64-bit invalid_id case yet?

src/mesh/exodusII_io.C
```diff
@@ -286,7 +286,30 @@ void ExodusII_IO::read (const std::string & fname)
...
+              elem->set_extra_integer(id, std::round(v));
+            else
+            {
+              if (v == (Real) DofObject::invalid_id)
+                elem->set_extra_integer(id, DofObject::invalid_id);
+              else
+              {
+                dof_id_type max_representation = 1;
```
roystgnr, line 299: You might need to explicitly make this unsigned long to avoid that overflow warning with 32-bit dof_id_type?

Although it does seem like the compiler could be smarter here; you shouldn't get an overflow unless you're in the other branch of the if statement, which is using a constexpr argument so should be evaluable at compile time.

src/mesh/exodusII_io.C
YaqiWang: ;-) yes, otherwise the bit shift operation could create some strange behavior.

src/mesh/exodusII_io.C
YaqiWang: I can do that.

YaqiWang:
@jwpeterson I basically chop the quad into two triangles and evaluate the areas of the two, then weight the centroids of two triangles to get the centroid of the quad. There could be better ways on evaluating centroid for general polygons. If you guys can come up some better formula, I will be happy to remove that commit.

YaqiWang:
@roystgnr No I have not. We can wait until this is confirmed by Hansel.

YaqiWang:
What is that `k=1,2,3`?

jwpeterson:
Just index notation, so x_1 = x, x_2 =y, etc.

YaqiWang:
Oh. Yes. That should be the definition of centroid. Because we know the centroid of triangle is simple the arithmetic average of three vertices, we can use two triangles to find the centroid of a quad. IMO, we should replace the centroid function in the base class to use quadrature and then can override in derived element types if there is a faster formula.

jwpeterson:
OK, I think your approach works for quads with 4 points co-planar, but I am not positive for a non-coplanar Quad4. In any case, I agree that this is similar to `Elem::volume()` in that the base class implementation should use generic quadrature but we should provide optimized derived class implementations. Frequently the optimized formulas are just hard-coded quadrature rules, which can be pretty fast.

YaqiWang:
You are right. I was not considering quad faces in 3d cells. Let me remove this commit so that we do not accidentally get it merged.

roystgnr:
I'm okay with this once Hansol (just joined github, @HansolPark1) can confirm it really works as expected.  I'd be happier if we had some unit test coverage for it, though, especially if we ever want to consider transitioning to the stronger typed options in ExodusII 8.

roystgnr:
On Slack @HansolPark1 tells us:

>  I checked that writing -1 and reading it gives me -1 now. This perfectly works for me, but I don't know why writing -2 and reading it still gives me -1 as well.

Maybe that's unavoidable?  -1 and -2, when cast to the unsigned dof_id_type and from there to Real, give the same result?  In which case my sanity checking ideas aren't at all as failsafe as I thought they would be, but this still seems like an improvement on the existing code.

HansolPark1:
> On Slack @HansolPark1 tells us:
> 
> > I checked that writing -1 and reading it gives me -1 now. This perfectly works for me, but I don't know why writing -2 and reading it still gives me -1 as well.
> 
> Maybe that's unavoidable? -1 and -2, when cast to the unsigned dof_id_type and from there to Real, give the same result? In which case my sanity checking ideas aren't at all as failsafe as I thought they would be, but this still seems like an improvement on the existing code.

I've checked with an arbitrary negative number like -56, and it still reads as -1. But I'm satisfied with the current improvement though.

YaqiWang:
I guess `(Real) -56_uLL` is equal to `(Real) DofObject::invalid_id` thus you did not hit the error message in this change?

roystgnr:
That would be my guess, and gdb confirms, `(double)(unsigned long)(-1) - (double)(unsigned long)(-56)` is 0.  Maybe try `-5600` to be safe?  That should round to a different number.

HansolPark1:
Tried -5600 and exodusII_io.C gave me the following error message.
"Error! An element integer value higher than 9007199254740992 was found! Exodus uses real numbers for storing element  integers, which can only represent integers from 0 to 9007199254740992."

YaqiWang:
> I'm okay with this once Hansol (just joined github, @HansolPark1) can confirm it really works as expected. I'd be happier if we had some unit test coverage for it, though, especially if we ever want to consider transitioning to the stronger typed options in ExodusII 8.

@roystgnr can you give more details about this in your mind? I am happy to add one if we can have something to make future changes less painful.

roystgnr:
I'm afraid a painless transition isn't possible.  Deciding to work directly with manually-written-doubles, rather than using some encapsulated write counterpart to the encapsulated read support, was a design mistake; we can't ever upgrade the implementation of a write API that doesn't exist.  But, in the future we'll want to redo this correctly, with typed data, and it'll be easy to accidentally regress the older support when we do so.  Adding a small tests/meshes/elem_integer_example.e along with a unit test in tests/mesh/extra_integers.C would be a good idea, to prevent that.

YaqiWang:
@roystgnr I created a small exodus with element integers and changed tests/mesh/extra_integers.C but I cannot make the test run. Maybe cppunit test has been outdated?

roystgnr:
> I created a small exodus with element integers and changed tests/mesh/extra_integers.C but I cannot make the test run. 

Do you have cppunit (both library and devel files) installed?

If you're having trouble doing that, you can push the test to your branch here and I can give it a shot.

YaqiWang:
I probably do not have cppunit installed. We are now using gtest. It is a little difficult for me to install cppunit, so possibly easier to let you finish the test. If you have any requirement on the mesh, let me know and I can generate one quickly for you. The mesh I checked in is a 3-by-3 2D mesh, with element integers illustrated in the inline comment.

roystgnr:
I can replicate the failure here; let me see how hard it is to fix.

YaqiWang:
Thanks Roy. Sorry for pushing this to you. If you want to change the mesh, feel free to let me know.

roystgnr:
Compilation fixes were easy, but the unit test dies on me.

It looks like it's supposed to die, yes?  The -5600 is supposed to trigger an error?  So we should just catch that error (and exit happily if we do) or do a CPPUNIT_ASSERT(false) if we failed to catch the error.  And we should probably wrap the test in a LIBMESH_ENABLE_EXCEPTIONS check too.

But that's only half of what we should be testing here: making sure that we scream and die properly when we see a bad datum is important, but let's also have another unit test that makes sure we set the correct values when we read an Exodus file with only good data?

YaqiWang:
Want another mesh file?

roystgnr:
Right.

roystgnr:
This works for me; assuming it passes CI, give it a look over, and if you like the new tests too then we should be good to merge.

YaqiWang:
Looks like 32bit failure is real? I guess both exodus file will be bad in 32bit mode.

roystgnr:
Hooray for comprehensive CI, I guess.

Looking at this in the debugger, I see that this wasn't true:
```
    /*
     This 3-by-3 mesh contains the following element integers:
     material_id = '0 -1    -2
                    1 -5600 4503599627370496
                    2 -3    4503599627370497'
     Real(-5600) != Real(-1) (used for invalid_id), so we can tell
     this is an error.
    */
```

Those negative numbers aren't being stored as negative numbers:

```
(gdb) p elem_ids
$10 = std::vector of length 1, capacity 1 = {std::map with 9 elements = {[0] = 0, [1] = 1.8446744073709552e+19, [2] = 1.8446744073709552e+19, [3] = 1, [4] = 3, [5] = 4503599627370496, [6] = 2, [7] = 1.8446744073709552e+19, [8] = 4503599627370497}}
```

They're being left as the configuration-dependent unsigned `dof_id_type`, then converted directly to FP from that, so naturally they break if read using a different `dof_id_type` configuration.  This needs to be fixed on the write end, which is another reason why creating a library function to read this data but not a library function to write it was a really awful idea.

YaqiWang:
Yes, the file was generated with dof_id_type as 8byte integers. -1 is Real(2^64-1) in the file. How to fix this? Do you need a sparate file for 4byte integer?

roystgnr:
Maintaining separate files which break in different ways on different configurations would just be compounding the problem.

The problem is:

> Those negative numbers aren't being stored as negative numbers

We need to be checking for invalid_id on the write side, not just on the read side, and writing -1 rather than 2^64-1 or 2^32-1.  And then on the read side we should be checking for -1.  Looking back through Slack logs, it looks like you guys just didn't take me literally enough:

> write invalid_id as -1 and test for -1 when reading

YaqiWang:
I literally had that `material_ids` parameter in the input with a mesh generator. The code convert `-1` to unsigned based on the type of `dof_id_type`. This is probably not very right because `material_ids` should be only accepting unsigned numbers. I was using it as the `DofObject::invalid_id` otherwise we do not have a good way to specify the value of `DofObject::invalid_id`. You think we should write `-1.0` instead of `Real(DofObject::invalid_id)`, something like `Real(dof_id_type_signed(DofObject::invalid_id))`?

YaqiWang:
The writing happens in MOOSE `ElemExtraIDAux` aux kernel, where the code
```
Real
ElemExtraIDAux::computeValue()
{
  return _id;
}
```
We can have
```
Real
ElemExtraIDAux::computeValue()
{
  return (_id == DodObject::invalid_id) ? -1 : _id;
}
```
Do you think we should do this? Note that this code change is only for protecting writing large integers in 64-bit that Real (double) cannot handle. I feel we can just limit the unit test in 64-bit build.

roystgnr:
> Do you think we should do this?

Definitely.

> I feel we can just limit the unit test in 64-bit build.

"It breaks, so let's avoid testing the conditions where it breaks" is almost never the right solution.

YaqiWang:
@roystgnr just pushed a PR in MOOSE for the writing. Once that is merged, you should be able to change here.

src/mesh/exodusII_io.C
```diff
@@ -286,7 +286,32 @@ void ExodusII_IO::read (const std::string & fname)
...
+                max_representation = (max_representation << std::numeric_limits<Real>::digits);
+                libmesh_error_msg_if(v > (Real) max_representation,
+                                     "Error! An element integer value higher than "
+                                     << max_representation
+                                     << " was found! Exodus uses real numbers for storing element "
+                                     " integers, which can only represent integers from 0 to "
+                                     << max_representation
+                                     << ".");
+
+                elem->set_extra_integer(id, std::round(v));
```
YaqiWang, line 311: Will need to add `if (std::round(v) > std::numeric_limits<dof_id_type>::max() && std::round(v) < std::numeric_limits<dof_id_type>::min())` here.

src/mesh/exodusII_io.C
```diff
@@ -286,7 +286,32 @@ void ExodusII_IO::read (const std::string & fname)
...
           for (auto & id : extra_ids)
-            elem->set_extra_integer(id, std::round(elem_ids[id][elem->id()]));
+          {
+            Real v = elem_ids[id][elem->id()];
+            // if digits of Real is greater than of dof_id_type, we can do direct conversion
+            if (std::numeric_limits<Real>::digits > std::numeric_limits<dof_id_type>::digits)
+              elem->set_extra_integer(id, std::round(v));
+            else
+            {
+              if (v == (Real) DofObject::invalid_id)
```
YaqiWang, line 296: This line will be `if (v == -1.0)`.

src/mesh/exodusII_io.C
```diff
@@ -286,7 +286,40 @@ void ExodusII_IO::read (const std::string & fname)
...
+          {
+            const Real v = elem_ids[id][elem->id()];
+
+            if (v == Real(-1))
+              {
+                elem->set_extra_integer(id, DofObject::invalid_id);
+                continue;
+              }
+
+            const long long iv = std::llround(v);
```
jwpeterson, line 298: I noticed that we don't explicitly
```
#include <cmath>
```
in exodusII_io.C. Presumably it is already included implicitly somehow or another, but might be good to make it explicit as well.

tests/mesh/extra_integers.C
```diff
@@ -196,6 +205,56 @@ public:
...
   void testExtraIntegersCheckpointHex8() { checkpoint_helper(HEX8, 2, true); }
 
+#ifdef LIBMESH_HAVE_EXODUS_API
+  void testExtraIntegersExodusReading()
+  {
+    Mesh mesh(*TestCommWorld);
+    mesh.allow_renumbering(false);
+
+    // This 3-by-3 mesh contains the following element integers:
+    std::vector<std::ptrdiff_t> material_id = {0, -1, 2,
```
jwpeterson, line 215: `std::ptrdiff_t` technically requires 
```
#include <cstddef>
```

jwpeterson (changes requested):
I don't see any Makefile.am changes here... but we had to add some the last time we added input meshes, e.g.
```
# Make sure any data goes in "make dist"
datadir = $(install_dir)
data = meshes/1_quad.bxt.gz \
       meshes/25_quad.bxt.gz \
       meshes/shark_tooth_tri6.xda.gz
```
It would be a shame to break `make dist` right before trying to tag a new release :)

roystgnr:
Actually, looks like the *last* couple times we (okay, Vikram and I...) added some input meshes we forgot to modify tests/Makefile.am then too.  After we get this merged I'll make a separate PR to add the missing entries.

src/mesh/exodusII_io.C
```diff
@@ -286,7 +287,40 @@ void ExodusII_IO::read (const std::string & fname)
...
+            libmesh_error_msg_if(iv > max_representation,
+                                 "Error! An element integer value higher than "
+                                 << max_representation
+                                 << " was found! Exodus uses real numbers for storing element "
+                                 " integers, which can only represent integers from 0 to "
+                                 << max_representation
+                                 << ".");
+
+            libmesh_error_msg_if(iv < 0,
+                                 "Error! An element integer value less than -1"
```
YaqiWang, line 315: -1 => 0

YaqiWang: nvm, -1 is for invalid id, so there is correct ;-)

roystgnr: Yeah, the test criterion there is a little confusing, but I figured I might as well double-check that we didn't let a -1 get to this point somehow, rather than setting invalid_id and continuing.

roystgnr:
I forgot the re-bootstrap until "make distcheck" yelled at me, but I think that's everything now (except the missing tests/Makefile.am meshes from previous PRs, which I'll add in a separate PR from this).

jwpeterson (approved):

roystgnr:
Waiting to give @YaqiWang a chance to double-check this with his codes, then I'll squash some commits together and let it run through CI again before we merge.

src/mesh/exodusII_io.C
YaqiWang: Possibly we should use dof_id_type here because `v` is supposed to be generated from it.

YaqiWang:
I will re-run Griffin against this to double check and report back soon.

YaqiWang:
@roystgnr This does not have impact on Griffin.

src/mesh/exodusII_io.C
YaqiWang: Actually, uint_64 or unsigned long long because dof_id_type could be in 32-bits.

roystgnr: Using dof_id_type here is tricky, because there's no `std::dof_id_type_round()` and no functions that overload on return type; we'd need some `#if LIBMESH_DOF_ID_BYTES > 4` and `#else` and such to use different functions in different cases.

But more to the point, it shouldn't be necessary; you're right that we'll hit other guards (an FP exception thrown if we're out of the long long range, a libmesh C++ exception thrown if we're in `long long` range but out of the 1-to-1 representation range, and then we have a redundant assert (via `cast_int` at the end) if somehow we made it past those but we're not in `dof_id_type` range.

YaqiWang: Will `std::llround` throw an error if `v` is between `2^63` to `2^64`?

jwpeterson: From the man page:
```
       Domain error: x is a NaN or infinite, or the rounded value is too large
              An invalid floating-point exception (FE_INVALID) is raised.

       These functions do not set errno.
```

PR #2978, opened 2021-08-06
Deprecate Elem::centroid(), add Elem::true_centroid() and Elem::vertex_average() alternatives

jwpeterson:
This PR deprecates `Elem::centroid()`, replacing it with a shim that calls `Elem::vertex_average()` to mimic the previous behavior. It also adds `Elem::true_centroid()` which computes the "geometric" centroid of the element using optimized formulas where possible, and numerical quadrature otherwise. 

In my non-scientific testing, I found that it's quite a bit slower to call `Elem::true_centroid()` when all you need is a point "somehwere" in the interior of the Elem, so after some discussion, @roystgnr and I decided to make people "opt in" to the more expensive calculation only when it is actually needed by an application.

Refs #2976

### DIFF
--- a/examples/miscellaneous/miscellaneous_ex9/augment_sparsity_on_interface.C
+++ b/examples/miscellaneous/miscellaneous_ex9/augment_sparsity_on_interface.C
@@ -31,9 +31,9 @@ void AugmentSparsityOnInterface::mesh_reinit ()
   // Loop over all elements (not just local elements) to make sure we find
   // "neighbor" elements on opposite sides of the crack.
 
-  // Map from (elem, side) to centroid
-  std::map<std::pair<const Elem *, unsigned char>, Point> lower_centroids;
-  std::map<std::pair<const Elem *, unsigned char>, Point> upper_centroids;
+  // Map from (elem, side) to vertex average
+  std::map<std::pair<const Elem *, unsigned char>, Point> lower_avgs;
+  std::map<std::pair<const Elem *, unsigned char>, Point> upper_avgs;
 
   for (const auto & elem : _mesh.active_element_ptr_range())
     for (auto side : elem->side_index_range())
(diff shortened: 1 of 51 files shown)


examples/miscellaneous/miscellaneous_ex9/augment_sparsity_on_interface.C
```diff
@@ -43,14 +43,14 @@ void AugmentSparsityOnInterface::mesh_reinit ()
             {
               std::unique_ptr<const Elem> side_elem = elem->build_side_ptr(side);
 
-              lower_centroids[std::make_pair(elem, side)] = side_elem->centroid();
+              lower_centroids[std::make_pair(elem, side)] = side_elem->vertex_average();
```
roystgnr, line 46: Would it be worthwhile to search-and-replace in all the variable names too?

`centroid = elem->not_a_centroid();` is funny but it might confuse someone years down the line.

jwpeterson: I added a comment in the `CentroidPartitioner` docs, but did not rename the class... I think the current name is more evocative even if less correct than e.g. `ElemVertexAveragePartitioner` or something like that. Also we'd have to keep the old name around for backwards compatibility, etc. regardless.

YaqiWang (changes requested):
@jwpeterson This is really nice! My only minor comment is that `Edge2` can has a specialization. I will test this with Griffin later.

YaqiWang:
I am not so concerned by the cost because users can calculate the centroid (similarly as volume) and save them for later use if they want.

src/geom/elem.C
```diff
@@ -356,6 +356,110 @@ const Elem * Elem::reference_elem () const
...
+  // the centroid more efficiently when possible.
+
+  // If this Elem has an elevated p_level, then we need to generate a
+  // barebones copy of it with zero p_level and call true_centroid()
+  // on that instead.  This workaround allows us to avoid issues with
+  // calling FE::reinit() with a default_order() FEType, and then
+  // having that order incorrectly boosted by p_level.
+  if (_p_level)
+    {
+      auto elem_copy = Elem::build(this->type());
```
roystgnr, line 413: I originally didn't like this idea because of the expense of having to make a heap allocation inside a centroid calculation; it didn't occur to me that you *only* have to make the allocation in the uncommon case.  We still ought to fix this in FE at some point.

src/geom/face_quad4.C
```diff
@@ -265,6 +265,57 @@ void Quad4::connectivity(const unsigned int libmesh_dbg_var(sf),
 
 
 
+Point Quad4::true_centroid () const
+{
+  // Note: this function does a substantially similar behavior to the
+  // volume() calculation, so it would be good if they could share
+  // code somehow without losing any performance by introducing extra
+  // function calls, etc.
```
roystgnr, line 273: Inline function in an anonymous namespace?

roystgnr (approved):

examples/miscellaneous/miscellaneous_ex9/augment_sparsity_on_interface.C
roystgnr: Yeah, I wouldn't want to change any class names, thanks to the added hassle; I'm just thinking about these little internal variables here and there.

src/geom/elem.C
jwpeterson: > We still ought to fix this in FE at some point.

I didn't look into it too carefully, but it just seemed like it would be awkward. Adding p_level support in the first place was already a tricky dance of trying to find all the places where it needed to be accounted for, now we would need to go in and add a "oops, just kidding don't actually use p_level" flag to all of those places.

src/geom/face_quad4.C
jwpeterson: Yes that's probably the way to go, but my thinking has shifted a bit since writing this comment, since only a few lines of code that are unlikely to change actually need to be duplicated. 

What I think will actually be more useful is a stand-alone `compute_quad4_centroid(Point& p0, Point& p1, Point& p2, Point& p3);` function that would be called by Quad4, but also by 3D Elems with Quad faces (Pyramid, Hex, prism) since the formulas for the 3D centroids depend explicitly on the centroids of the faces. But that's an optimization I'd like to save for a later PR.

PR #2980, opened 2021-08-12
Generalize vector operations to always use global forms

lindsayad:
This allows operations in which `this` is `GHOSTED` but the other
    vector(s) are `PARALLEL` while simultaneously supporting the
    previous use cases, e.g. GHOSTED-GHOSTED, PARALLEL-GHOSTED, etc.

### DIFF
--- a/src/numerics/petsc_vector.C
+++ b/src/numerics/petsc_vector.C
@@ -389,10 +389,12 @@ void PetscVector<T>::add (const T a_in, const NumericVector<T> & v_in)
 
   libmesh_assert_equal_to (this->size(), v->size());
 
-  VecGhostLocalFormRAII<T> ghost(*this, this->type() == GHOSTED);
-  VecGhostLocalFormRAII<T> v_ghost(*v, this->type() == GHOSTED);
-  PetscErrorCode ierr = VecAXPY(ghost.local_form, a, v_ghost.local_form);
+  PetscErrorCode ierr = VecAXPY(_vec, a, v->vec());
   LIBMESH_CHKERR(ierr);
+  if (this->type() == GHOSTED)
+    VecGhostUpdateBeginEnd(this->comm(), _vec, INSERT_VALUES, SCATTER_FORWARD);
+
+  this->_is_closed = true;
 }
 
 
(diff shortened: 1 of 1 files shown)


src/numerics/petsc_vector.C
```diff
@@ -390,7 +390,7 @@ void PetscVector<T>::add (const T a_in, const NumericVector<T> & v_in)
   libmesh_assert_equal_to (this->size(), v->size());
 
   VecGhostLocalFormRAII<T> ghost(*this, this->type() == GHOSTED);
-  VecGhostLocalFormRAII<T> v_ghost(*v, this->type() == GHOSTED);
+  VecGhostLocalFormRAII<T> v_ghost(*v, v->type() == GHOSTED);
```
fdkong, line 393: There are several other places where the same issue may exist. Could you fix other places as well?

fdkong (changes requested):

jwpeterson:
Crap, it looks like I broke this in 6d51e68bf50105629d24df4adb78fd3005cfd9ca. Your fix looks good to me though :+1:

jwpeterson (approved):

roystgnr (approved):

roystgnr (changes requested):
Actually ... the code certainly *looks* obviously correct here, and I just assumed someone had let some thread-unsafe code slip into MOOSE ... but those errors all look like PETSc Vec related errors.  What's going on there?

lindsayad:
Wow yea MOOSE no likey

lindsayad:
Hmm yea it seems like @jwpeterson 's new code perfectly emulated the old code. I guess just we've never supported doing this `add` overload when `this` is ghosted and `v` is parallel? And if this is not the fix to support that, then I'll have to expand my PETSc knowledge

jwpeterson:
> perfectly emulated the old code.

Haha, yes I see now what I was doing was just replacing those enclosing `if`-`else` blocks with something equivalent... @fdkong: is attempting to get the local form of a PARALLEL vector really supposed to just give you back a `NULL` pointer? It seems like that could only ever end badly...

lindsayad:
Yea all the failures in MOOSE are for the opposite scenario: `this` is `PARALLEL` and `v` is `GHOSTED`. So yea the fact that my change incurred all those parallel failures makes total sense. When `this` is `PARALLEL` we definitely don't want to `add` with the local form of `v` (which is what my change did)

lindsayad:
I feel like what should be done when `this` is `GHOSTED` and `v` is `PARALLEL` is to do the `VecAXPY` *not* with the local forms and then to do `VecGhostUpdateBegin` and `VecGhostUpdateEnd` after... ?

fdkong:
>@fdkong: is attempting to get the local form of a PARALLEL vector really supposed to just give you back a NULL pointer? It seems like that could only ever end badly...

From my understanding, the local form is created only when we call VecCreateGhost/VecCreateGhostBlock or their friends. 

The local form will be a null pointer if we do not create a ghost vector

lindsayad:
You might say that if you don't create a ghost, then the local form is truly ethereal ... like a 👻

jwpeterson:
> I feel like what should be done when `this` is `GHOSTED` and `v` is `PARALLEL` is to do the `VecAXPY` _not_ with the local forms and then to do `VecGhostUpdateBegin` and `VecGhostUpdateEnd` after... ?

OK... so we would have to add an if-else check back in to handle that case? That's a bit annoying but not the end of the world I guess. Note also that there's a `VecGhostUpdateBeginEnd()` macro which can simplify that part a bit.

lindsayad:
I *think* the new commit should work for all cases and simplifies the code

fdkong (approved):
The code looks so beautiful. 

I like these changes.

jwpeterson:
So can `VecGhostLocalFormRAII` just be done away with completely? I'm not sure I see the point of it now. It looks like `PetscVector<T>::scale`, `abs()`, and `operator=(T)` still use it, but do they need to?

roystgnr (approved):

PR #2985, opened 2021-08-16
ExodusII IGA support

roystgnr:
This shouldn't be merged as-is; the IGA data format is still a bit WIP and there might be a change upcoming in the way the spline node coordinates are stored.

However, it's working with the previous draft of the format, and the remaining update should be much less intrusive, so this is probably a good place to put the new code through CI and make sure I haven't regressed anything elsewhere.


roystgnr:
Refs #2179

roystgnr:
That last unit test is enough to make me happy.  @jwpeterson, want to review?  No rush.  We still definitely don't want to merge this until after we've done a release branch and a Moose submodule update, and not until after I've gotten word that the draft standard we're implementing here is no longer in flux.

PR #2989, opened 2021-08-20
Tri7 support

roystgnr:
The true goal here is to give us face nodes on triangular-faced cells, but the face element itself was tricky enough that it deserves its own PR.


src/quadrature/quadrature_nodal_2D.C
```diff
@@ -145,6 +145,24 @@ void QNodal::init_2D(const ElemType, unsigned int)
         return;
       }
 
+    case TRI7:
+      {
+        // We can't exactly integrate cubics with only seven nodes,
+        // but with w_i = integral(phi_i) for Lagrange shape functions
+        // phi_i, we'll at least get exact integrals of every Lagrange
+        // shape function, including the cubic bubble.
```
jwpeterson, line 153: I agree with the weights you found here, but I don't see how you get them by integrating the TRI7 basis functions? If I do that using sympy, I get the following:
```
import sympy as sy
from fractions import Fraction

xi, eta = sy.sympify('xi, eta')

zta0 = 1 - xi - eta
zta1 = xi
zta2 = eta

# Representative vertex basis function
phi0 = 2 * zta0 * (zta0 - Fraction(1,2)) + Fraction(3,27) * zta0 * zta1 * zta2
wv = sy.integrate(sy.integrate(phi0, (eta, 0, 1-xi)), (xi, 0, 1))
print(f'wv={wv}') # wv=1/1080

# Representative edge basis function
phi3 = 4 * zta0 * zta1 - Fraction(12,27) * zta0 * zta1 * zta2
we = sy.integrate(sy.integrate(phi3, (eta, 0, 1-xi)), (xi, 0, 1))
print(f'we={we}') # we=22/135

# Centroid basis function
phi6 = zta0 * zta1 * zta2
wc = sy.integrate(sy.integrate(phi6, (eta, 0, 1-xi)), (xi, 0, 1))
print(f'wc={wc}') # wc=1/120

# Weights should sum up to 1/2, the area of the reference element
print(f'Weight sum = {3*wv + 3*we + wc}') # weight sum = 1/2
```
Also, I believe that the rule you have here *is* exact for cubics.

jwpeterson: Oops, I was off by a factor of 1/27 in my original script, the corrected version agrees with you now:
```
import sympy as sy
from fractions import Fraction

xi, eta = sy.sympify('xi, eta')

zta0 = 1 - xi - eta
zta1 = xi
zta2 = eta

# Representative vertex basis function
phi0 = 2 * zta0 * (zta0 - Fraction(1,2)) + 3 * zta0 * zta1 * zta2
wv = sy.integrate(sy.integrate(phi0, (eta, 0, 1-xi)), (xi, 0, 1))
print(f'wv={wv}') # wv=1/40

# Representative edge basis function
phi3 = 4 * zta0 * zta1 - 12 * zta0 * zta1 * zta2
we = sy.integrate(sy.integrate(phi3, (eta, 0, 1-xi)), (xi, 0, 1))
print(f'we={we}') # we=1/15

# Centroid basis function
phi6 = 27 * zta0 * zta1 * zta2
wc = sy.integrate(sy.integrate(phi6, (eta, 0, 1-xi)), (xi, 0, 1))
print(f'wc={wc}') # wc=9/40
```

So I guess the only further clarification I would make is that this rule is actually exact for cubics... I'm 99% sure about this, but it also wouldn't hurt my feelings if you wanted to verify that :)

roystgnr: > So I guess the only further clarification I would make is that this rule is actually exact for cubics... I'm 99% sure about this, but it also wouldn't hurt my feelings if you wanted to verify that :)

Whoa, seriously?  That's fantastic.  I didn't even bother to check for that because it seemed too unlikely to bother.

roystgnr:
Whoa.  The new tests here actually caught a pre-existing bug.  `has_affine_map` was broken for a bunch of existing elements - using relative tolerances as it should, but relative to a single point location rather than to the displacement between two points.  I'm going to put fixes for that in a separate PR.

src/quadrature/quadrature_nodal_2D.C
roystgnr: It totally works.  Quadrature rules are magic.

I've updated the comment accordingly.

I can't actually think of any use for this fact, though.  We generally only do nodal quadrature rules when we want a lumped mass matrix, but we only get a lumped mass matrix if we're using a Lagrange basis on that exact set of nodes, so if we try to "cheat" a Tri6 by using a Tri7 rule, we wouldn't get a lumped mass matrix with one extra entry, we'd just get a dense mass matrix that only integrates exact cubics at a cost which could have given us exact quintics.

jwpeterson: Agreed, it's kind of cool that it works, but I'm not sure when it would ever be useful. The only potential benefit that comes to mind is in some compressible flow formulations, where you might use a nodal quadrature to evaluate inviscid flux terms for stability reasons... though such formulations are never based on TRI7 elements. :man_shrugging:

roystgnr:
I caught one chunk of missing code (albeit in a code path only used for viz?) while working on Tet14 support in another branch, so I'm going to mark this as "do not merge" until I'm done there, just in case I run across any other omissions.  However, what's here should be ready for review if @jwpeterson wants to finish that or if anyone else wants to chime in.  Pinging @lindsayad since he's got a bit of a stake; this is the base of what's eventually going to let us expand the `SIDE_HIERARCHIC` type from #2900 onto non-tensor-product elements in 3D.

jwpeterson (approved):

roystgnr:
Forgot to put this back under "do not merge" after we ran into the has_affine_map fiasco.  We shouldn't be merging this until we've got a submodule update into Moose and a new release branch in libMesh.

lindsayad:
Could we do the release before the moose submodule update? I would like to include #3005 for example in the moose submodule update and I don't really see any reason why this couldn't also be included in the moose submodule update

roystgnr:
I'd rather do the release after the submodule, just for the added test coverage we get that way.  The "configure --weird-stuff" options we try out on every release aren't a big deal to most Moose users, whereas the Moose apps CI is (as we've recently found out to our sorrow) a great way to test a release.

I had been intending to get #3005 into the submodule unless anyone objects, though.  That's a fairly small + clean change with an immediate use case and no likely side effects.  But this PR here is a couple thousand lines of code, and it's not really useful for its own sake (we can already store internal DoFs on the Elem DofObject of a Tri6, and that covers everything but Lagrange, which nobody's going to want to use on such a weird element).  Even if we also throw in a few more thousand lines from #3002, we're *still* not quite to the holy grail - that PR adds the geometric Tet14, and the basis LAGRANGE FE support needed to make it work, but unless there's a use case I'm missing, it's still not going to be *useful* until I add HIERARCHIC and SIDE_HIERARCHIC FE support in a later PR.

roystgnr:
Rebased, and we've now got a release branch separate from master, so I'm planning to merge once CI is happy.

PR #2990, opened 2021-08-22
Add Pyramid5::true_centroid()

jwpeterson:
Also, turn on the `true_centroid()` unit test for this element.

### DIFF
--- a/include/geom/cell_hex8.h
+++ b/include/geom/cell_hex8.h
@@ -185,6 +185,17 @@ public:
   static const unsigned int edge_sides_map[num_edges][2];
 
   /**
+   * Class static helper function that computes the centroid of a
+   * hexahedral region from a set of input points which are assumed to
+   * be in counter-clockwise order. Hex8::true_centroid() and
+   * Pyramid5::true_centroid() are implemented in terms of this
+   * function.
+   */
+  static Point centroid_from_points(
+    const Point & x0, const Point & x1, const Point & x2, const Point & x3,
+    const Point & x4, const Point & x5, const Point & x6, const Point & x7);
+
+  /**
    * We compute the centroid of the Hex using a customized numerical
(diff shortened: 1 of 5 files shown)


include/geom/cell_hex8.h
```diff
@@ -184,6 +184,17 @@ class Hex8 final : public Hex
    */
   static const unsigned int edge_sides_map[num_edges][2];
 
+  /**
+   * Class static helper function that computes the centroid of a
+   * hexahedral region from a set of input points which are assumed to
+   * be in counter-clockwise order. Hex8::true_centroid() and
```
roystgnr, line 190: What does it mean for points in 3D to be in counter-clockwise order?  If you can look along a "z" axis with respect to which the points' angle theta is increasing counter-clockwise, can't you just look along the same axis in the opposite direction and see the angle going clockwise?

roystgnr (approved):
Amazingly convenient that this works, but `true_centroid_and_volume` is testing it on 48 distorted pyramids, yes?  And it's within 1e-12 on O(1) sized elements?  That's good enough for me.

include/geom/cell_hex8.h
jwpeterson: Uh, yeah this comment is poorly-worded. The function works on a set of points that are numbered "like a Hex8", including if you e.g. collapse a face down to a single point (Pyramid case) or an edge (Prism6 case). I'm happy to change the comment to whatever.

jwpeterson:
> is testing it on 48 distorted pyramids, yes?

Yes, and importantly some of those pyramids should have non-planar bases so we should be testing the most "general" type of Pyramid5... previously I had used a "1x1x1" `build_cube()` mesh in the unit test, but that gives 6 pyramids with planar bases (since MeshTools::Modification::distort doesn't move boundary nodes).

PR #2991, opened 2021-08-23
Fix has_affine_map()

roystgnr:
As discovered in #2989, we have some subclasses using the wrong type of vector (and thus the wrong fuzzy comparison relative tolerances) in `has_affine_map()`.

Assuming I've cherry picked them out of all the Tri7 work correctly, this should fix and add more testing for those methods.

### DIFF
--- a/src/geom/cell_tet10.C
+++ b/src/geom/cell_tet10.C
@@ -158,23 +158,29 @@ bool Tet10::is_child_on_side(const unsigned int /*c*/,
 bool Tet10::has_affine_map() const
 {
   // Make sure edges are straight
-  if (!this->point(4).relative_fuzzy_equals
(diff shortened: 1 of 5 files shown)


roystgnr:
We're not going to put the Tri7 code into our next release, but this fix should probably go in once we're confident it's working.  The false negatives caught by my new unit tests aren't a big deal, since they just mean that we can spend a handful more flops than we should at elements with non-vertex nodes touching the origin ... but the same bug could lead to false positives for meshes with nonlinear elements (even simple ones like non-parallelogram quads) at extremely large distances (O(1e6)) from the origin.  (Is anyone currently using libMesh for GCMs or other thousand-kilometer-domain scale simulations? If so that's great and I apologize for breaking them.)

jwpeterson (approved):

PR #2997, opened 2021-08-26
Greatly reduce has_affine_map tolerance

roystgnr:
From 1e-6 down to 1e-12 in the default double-precision configuration.

When doing a MOOSE libMesh submodule update in https://github.com/idaholab/moose/pull/18705 we discovered that our existing `has_affine_map()` tolerance is so loose that just tweaking the equations we use in implementations is resulting in noticeable diffs in some tests.

When I started using `relative_fuzzy_equals` in these calculations I didn't really think about tolerance issues, since surely there wouldn't be any elements that aren't affine but are only 1e-7 away from affine!  But mesh deformation problems produce exactly such elements, which means we're throwing as much as O(1e-6) relative error into mesh deformation problem calculations at that point, when the elements are nonlinear enough to need independent mapping gradient calculations at each quadrature point but not nonlinear enough for our loose tolerances to detect that fact.

### DIFF
--- a/include/base/libmesh_common.h
+++ b/include/base/libmesh_common.h
@@ -121,7 +121,7 @@ typedef LIBMESH_DEFAULT_SCALAR_TYPE Real;
 // For example, v == 0 is changed to std::abs(v) < TOLERANCE.
 
 #ifdef LIBMESH_DEFAULT_SINGLE_PRECISION
-static const Real TOLERANCE = 2.5e-3;
+static constexpr Real TOLERANCE = 2.5e-3;
 # if defined (LIBMESH_DEFAULT_TRIPLE_PRECISION) || \
      defined (LIBMESH_DEFAULT_QUADRUPLE_PRECISION)
 #  error Cannot define multiple precision levels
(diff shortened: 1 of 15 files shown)


roystgnr:
Looks like the 32bit/64bit thing is a red herring; what's going on is that only the 64-bit recipe is running the modules tests and thereby seeing their failures?

roystgnr:
1.8e-4 rel diff in tensor_mechanics/test:substepping.power_law_substepping seems to be the worst of it.  According to the exodiff, our stress is significantly higher now, which works as both a technical datum *and* a metaphor!

So what's the best way out of the chicken-and-egg dilemma here?  Cherry-pick this PR onto a new branch on top of the previous submodule hash, tag that, do the update to it simultaneously with the regression test re-golding, then revisit doing a proper submodule update after that?

lindsayad:
That seems more complex than it needs to be? What would be the goal? Intuititively I would think it would be to keep other PR testing to libMesh green but the moment we update MOOSE and re-gold then libMesh master will be red with respect to MOOSE testing. So I would think that we should just merge this when we think we are ready to update the MOOSE submodule hash and then blitz our re-golds through modules and apps

roystgnr:
> Intuititively I would think it would be to keep other PR testing to libMesh green

That was what I was shooting for, but you know what?  I don't have anything in the pipeline right now that shouldn't be delayed until after the submodule update anyways.

lindsayad (approved):
Ok well I'm ready to start the chaos whenever you are 😆

jwpeterson:
Friday afternoon before a 3-day weekend sounds good :grimacing:

lindsayad:
Sounds like the old ninja commit days

roystgnr:
I want to make a "Leeroy Jenkins" joke, but I'm not sure the kids these days get that.  I think the current meme is something like "full send", but I can't find it on knowyourmeme.com, and come to think of it I'm not sure the kids these days still use that.

At this rate I might as well give up on modern memeing entirely.  23 Skidoo!

PR #3005, opened 2021-09-01
Create multi_evaluable predicate

lindsayad:
Anyone got a great idea for a unit test? Conceptually I'd want a test in which I ghost every element around a given element and have one system with Lagrange and one with monomial and then check and see that the element is marked as evaluable for the single predicate with the first system DofMap, and not evaluable when all DofMaps are considered but I don't know how easy this would be to do (especially for different processor counts).

@hugary1995 do you want to play around with this? You'd have to change the MOOSE `FEProblemBase::getEvaluableElementRange` method (which we will want to do anyways)

### DIFF
--- a/tests/base/multi_evaluable_pred_test.C
+++ b/tests/base/multi_evaluable_pred_test.C
@@ -1,0 +1,156 @@
+#include <libmesh/partitioner.h>
+#include <libmesh/mesh.h>
+#include <libmesh/auto_ptr.h>
+#include <libmesh/mesh_generation.h>
+#include <libmesh/equation_systems.h>
+#include <libmesh/elem.h>
+#include <libmesh/system.h>
+#include <libmesh/dof_map.h>
+#include <timpi/communicator.h>
+
+#include "test_comm.h"
+#include "libmesh_cppunit.h"
+
+using namespace libMesh;
+
+class ContrivedPartitioner : public Partitioner
+{
+public:
+  ContrivedPartitioner() = default;
+  ContrivedPartitioner(const ContrivedPartitioner &) = default;
+  ContrivedPartitioner(ContrivedPartitioner &&) = default;
+  ContrivedPartitioner & operator=(const ContrivedPartitioner &) = default;
+  ContrivedPartitioner & operator=(ContrivedPartitioner &&) = default;
+  virtual ~ContrivedPartitioner() = default;
+
+  std::unique_ptr<Partitioner> clone() const override
+  {
+    return libmesh_make_unique<ContrivedPartitioner>(*this);
+  }
+
+protected:
+  void _do_partition(MeshBase & mesh, const unsigned int n) override
+  {
+    libmesh_assert(n > 1);
+
+    unsigned int n_interior_elems = 0;
+    for (Elem * const elem : as_range(mesh.elements_begin(), mesh.elements_end()))
+    {
+      bool internal_elem = true;
+      for (const Elem * const neighbor : elem->neighbor_ptr_range())
+        if (!neighbor)
+        {
+          internal_elem = false;
+          break;
+        }
+
+      // The interior element will go on processor 1. Other elements will go on processor 0. For a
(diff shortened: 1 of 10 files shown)


roystgnr:
Should we copy that vector into the predicate, rather than taking a const reference to it?  It shouldn't be much memory cost, that might be balanced out by the compiler being able to optimize better, and most importantly that'll avoid the looming bug where someone creates a vector in order to get a range of these iterators but then tries to hang on to the range or to the individual iterators after the vector goes out of scope.

roystgnr:
There's no unit test for this that will be worthwhile on 1 processor, but we default to 2 processors in CI and I at least run "make check" manually with 16 processors all the time; coverage that would trigger test failures in either (or ideally both) of those cases would be better than nothing.

hugary1995:
I can make the change in FEProblemBase once the libMesh update is in.

roystgnr:
I feel like we could use custom algebraic ghosting functors in the unit test ... maybe do two levels of point neighbors in the functor for one system, three levels of side neighbors in the functor for the other system, so that on the right partitioning we would have no empty sector of their Venn diagram?  And if we had 2+ processors solving on a build_square mesh, it'd be easy to do the right partitioning manually.

lindsayad:
> vector goes out of scope.

I thought that the lifetime gets extended if you've [bound a const reference to it](https://en.cppreference.com/w/cpp/language/lifetime)? I'm fine making this pass by value; I'm just curious whether my understanding is wrong

lindsayad:
> no empty sector of their Venn diagram?

Are these the three sectors in the Venn?
- elements evaluable for both DofMaps
- elements only evaluable for one DofMap
- elements only evaluable for the other DofMap

lindsayad:
The test I created is not nearly as sophisticated as the one you suggested but it satisfies me... Let me know if you want it to be more adnvaced

tests/base/multi_evaluable_pred_test.C
```diff
@@ -0,0 +1,156 @@
...
+#include <libmesh/system.h>
+#include <libmesh/dof_map.h>
+#include <timpi/communicator.h>
+
+#include "test_comm.h"
+#include "libmesh_cppunit.h"
+
+using namespace libMesh;
+
+class ContrivedPartitioner : public Partitioner
```
roystgnr, line 16: I love this.

tests/base/multi_evaluable_pred_test.C
```diff
@@ -0,0 +1,156 @@
...
+    elem_system.add_variable("elem", CONSTANT, MONOMIAL);
+    auto & elem_dof_map = elem_system.get_dof_map();
+    elem_dof_map.remove_default_ghosting();
+
+    es.init();
+
+    {
+      auto n_evaluable = std::distance(mesh.evaluable_elements_begin(nodal_dof_map),
+                                       mesh.evaluable_elements_end(nodal_dof_map));
+      typedef decltype(n_evaluable) comp_type;
```
roystgnr, line 96: This isn't important enough to change, but it's sad that the already-questionable benefits of "`auto` is shorter than `std::size_t`" were all lost a mere two lines later.

roystgnr (approved):
Not perfect test coverage, but good enough for me, and sadly it's probably still better than average for the library as a whole.

tests/base/multi_evaluable_pred_test.C
lindsayad: Hahaha but not we're protected in case `difference_type` is ever something other than `std::size_t`!

jwpeterson: I think I'm missing the joke here, but `difference_type` is a signed type (`std::ptrdiff_t`)

roystgnr: I think I'm the unwitting butt of my own joke here.  "Why use auto; it's not like we're ignorant of the correct return type, right guys?"

PR #3009, opened 2021-09-02
Upgrade lcov to version 1.15

roystgnr:
Earlier lcov versions don't support the format of the code coverage files generated by gcc 9 and gcc 10.

With this, `--enable-coverage` and `make coverage` generate lcov reports for me.

The reports are lousy right now, 48% function coverage and 57% line coverage, but that's an issue to fix at a later date.  To be fair, those numbers are just from just our unit tests and examples; if we also counted "Moose CI on every PR" and "Moose app CI on every submodule update" and "GRINS tests when Roy's not feeling lazy" then the numbers would hopefully be less embarrassing.


jwpeterson:
I didn't even realize we had an lcov in contrib :smile:

jwpeterson (approved):

jwpeterson:
I didn't even realize we had an lcov in contrib :smile:

roystgnr:
Ha!  At one point PECOS had automatic lcov reports on a number of our standalone libraries there, and I managed to get that ported over to libMesh, but somehow didn't it running in CI, not with Civet nor with BuildBot.  But now @loganharbour is setting up CI coverage tests for Moose and figuring out the backend stuff I never got to.

We're still figuring out how useful it will be, though.  Getting reliable coverage reports for templated inline functions seems to be particularly tricky.

PR #3011, opened 2021-09-07
Use typeid for hashing

### LINKED ISSUE #3010: Rely on typeid rather than type name for FParser JIT hashing.
The type name used for hashing JIT compiled AD function parser objects is always `ADReal` (in MOOSE), but the underlying type (both template and template parameters) are build config dependent. It is safer to use `typeid(Value_t).hash_code()` as it depends on the underlying type (behind all layers of typedefs) and includes template parameters.

dschwen:
Closes #3010

### DIFF
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -609,13 +609,13 @@ int ADImplementation<Value_t>::AutoDiff(unsigned int _var, typename FunctionPars
 }
 
 template<typename Value_t>
-std::size_t FunctionParserADBase<Value_t>::JITCodeHash(const std::string & Value_t_name)
+std::size_t FunctionParserADBase<Value_t>::JITCodeHash()
 {
   // start with a version tag in case the JIT function signature changes
   std::size_t h = std::hash<std::string>{}("v3");
   for (auto b :this->mData->mByteCode)
     libMesh::boostcopy::hash_combine(h, b);
-  libMesh::boostcopy::hash_combine(h, Value_t_name);
+  libMesh::boostcopy::hash_combine(h, typeid(Value_t).hash_code());
   return h;
 }
 
@@ -674,7 +674,7 @@ bool FunctionParserADBase<Value_t>::JITCompileHelper(const std::string & Value_t
     return false;
 
   // compute hash of the function
-  std::string hash = FParserJIT::hashToString(JITCodeHash(Value_t_name));
+  std::string hash = FParserJIT::hashToString(JITCodeHash());
 #ifndef NDEBUG
   hash += "_dbg";
 #endif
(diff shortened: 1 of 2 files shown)


roystgnr (commented):
This is going to break builds with RTTI disabled.  Which I think nobody uses, and as far as libMesh is concerned fparser is also an optional feature and shouldn't be expected to be compatible with every configuration ... but I still hate breaking it.  Is there any other solution?

dschwen:
We could go back to passing in the type name as a string. But I'd have to
tack on yet another string parameter to the JITCompileHelper method.
Another, somewhat dirty option would be to has the sizeof value for the
type in addition to the name. :-/

roystgnr:
I think I'd prefer the sizeof solution if it would actually work, but it's easy to imagine multiple ADReal options that have the same underlying size but not the same underlying code, because one has a pointer to Foo where the other has a pointer to Bar, or whatever.  So we'll go with this.

If anyone claims "I need to disable RTTI to save every possible byte of space, but also I need to do JIT compilation of ASCII functions", then we'll deal with that problem when we come to it.  Probably by laughing at them.

roystgnr (approved):

dschwen:
Lovely. We could always ifdef the sizeof approach for non-RTTI builds, to have at least a somewhat better working approach that works without RTTI.

lindsayad:
It appears that this never got tested. I'm confident that this broke tests in MOOSE

dschwen:
Wait, what? Why is there no CI here? Can you point me to a broken test?

dschwen:
Oh, f*ck, I see what I did now! OMG. `Value_t` is also Real for AD!!! Might as well just revert this. For now. I'll put in a better fix. D'Oh! :facepalm:

lindsayad:
```
materials/derivative_material_interface.ad_analytic_derivatives/jacobian ......... FAILED (INCORRECT JACOBIAN)
materials/derivative_material_interface.ad_analytic_derivatives/derivative_jacobian FAILED (INCORRECT JACOBIAN)
materials/derivative_sum_material.random_ic/ad_validating ................................... FAILED (EXODIFF)
materials/derivative_material_interface.ad_analytic_derivatives/material_chaining ........... FAILED (CSVDIFF)
```

jwpeterson:
I am also curious why there was no CI... was it just this one PR or were there others that weren't tested?

lindsayad:
There was no CI because CIVET was temporarily down (it's back online now)

dschwen:
The new plan is to ass a user specified salt to the hash.
MOOSE requires RTTI (for dynamic casts I guess). So I could use the typeid
hash in moose (and on the correct type this time).

jwpeterson:
> The new plan is to ass a user 

You can edit your post, but I'd like to hear more about this idea...

dschwen:
Relentless

PR #3013, opened 2021-09-09
Set next unique id in copy_nodes_and_elements

lindsayad:
We were currently setting `unique_id` for nodes after the call to
`add_point` which means that `_next_unique_id` will end up incorrect if
we don't also call `set_next_unique_id` after adding all our nodes. Note
that this problem doesn't exist for elements because we set the
`unique_id` before calling `add_elem` which results in the `MeshBase`
object correctly setting `_next_unique_id` on its own

### DIFF
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -230,6 +230,12 @@ void UnstructuredMesh::copy_nodes_and_elements(const UnstructuredMesh & other_me
       }
   }
 
+#ifdef LIBMESH_ENABLE_UNIQUE_ID
+  // We set the unique ids of nodes after adding them to the mesh such that our value of
+  // _next_unique_id may be wrong. So we amend that here
+  this->set_next_unique_id(other_mesh.parallel_max_unique_id() + unique_id_offset + 1);
+#endif
+
   //Finally prepare the new Mesh for use.  Keep the same numbering and
   //partitioning for now.
   this->allow_renumbering(false);
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -76,6 +76,7 @@ unit_tests_sources = \
   mesh/write_nodeset_data.C \
   mesh/write_edgeset_data.C \
   mesh/write_vec_and_scalar.C \
+  mesh/mesh_collection.C \
   mesh/all_second_order.C \
   numerics/composite_function_test.C \
   numerics/coupling_matrix_test.C \
--- a/tests/Makefile.in
+++ b/tests/Makefile.in
@@ -217,8 +217,8 @@ am__unit_tests_dbg_SOURCES_DIST = driver.C libmesh_cppunit.h \
 	mesh/mapped_subdomain_partitioner_test.C \
 	mesh/mesh_function_dfem.C mesh/write_sideset_data.C \
 	mesh/write_nodeset_data.C mesh/write_edgeset_data.C \
-	mesh/write_vec_and_scalar.C mesh/all_second_order.C \
-	numerics/composite_function_test.C \
+	mesh/write_vec_and_scalar.C mesh/mesh_collection.C \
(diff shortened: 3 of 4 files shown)


src/mesh/unstructured_mesh.C
```diff
@@ -116,9 +119,15 @@ void UnstructuredMesh::copy_nodes_and_elements(const UnstructuredMesh & other_me
...
 
 #ifdef LIBMESH_ENABLE_UNIQUE_ID
-        newn->set_unique_id(oldn->unique_id() + unique_id_offset);
-#endif
+        const unique_id_type new_unique_id = oldn->unique_id() + unique_id_offset;
+        newn->set_unique_id(new_unique_id);
+        potentially_next_unique_id = std::max(new_unique_id+1, potentially_next_unique_id);
       }
+
+    this->set_next_unique_id(potentially_next_unique_id);
```
roystgnr, line 127: This isn't safe on a distributed `other_mesh`, is it?  If we can't see the whole node range on every processor, we might be setting a value here that's inconsistent between processors and too low on some of them.

I think we're going to need to just call `other_mesh.parallel_max_unique_id()` and add `unique_id_offset` to that.

Boy.  We really ought to add a `this->comm().verify(id)` inside `DistributedMesh::set_next_unique_id` too.  This seems like a really easy mistake to make and a really hard mistake to catch.

And going one more level up the root cause tree ... we're just woefully neglecting `MeshBase::_is_prepared`, aren't we?  I suspect that `copy_nodes_and_elements` works the way it does because we were expecting code using it to `prepare_for_use` afterwards, but with nothing doing `set_isnt_prepared()` here how are we to know that, and with next to nothing asserting `is_prepared()` elsewhere how are we to verify that?

roystgnr: ...

I'm apparently the confused one.

We do have `prepare_for_use()` at the end of `copy_nodes_and_elements()`, don't we?  The problem is that, while we're always calling `update_parallel_id_counts()` via `prepare_for_use()`, in the guts of `parallel_max_unique_id()` there we still expect to see local values of `_next_unique_id` there which are locally valid, even if they're not globally valid; we don't actually inspect DofObject `unique_id()` values to see if any of them have gotten ahead of us.  And we shouldn't change that; I'd certainly rather not add an additional element loop and node loop inside what was previously an O(1) method.

So, yeah, the solution is still going to be to fix that manually with `other_mesh.parallel_max_unique_id()` plus `unique_id_offset`; my clever ideas about `prepare_for_use()` weren't clever at all.

roystgnr (approved):
Should we add a test for this?

lindsayad:
Probably. I'll try to get to one reasonably soon

lindsayad:
I'm not quite sure what to assert. On replicated meshes I could assert the final `parallel_max_unique_id` for a test in which I combine multiples meshes with `copy_nodes_and_elements`. The assertion is a little trickier for a distributed mesh, unless I wanted to replicated all the math in `DistributedMesh::set_next_unique_id`. I could just make sure that I can combine 3 meshes via two calls to `copy_nodes_and_elements`. I believe this would've failed before this PR on the second call

lindsayad:
Ok *I* think the test I added is pretty good! I verified that it failed with the old code (after some iterations!)

roystgnr:
There's no `CPPUNIT_ASSERT(anything)` in that test, though, is there?  Is it failing (prior to the fix here) on one of our internal asserts somehow?

lindsayad:
>  Is it failing (prior to the fix here) on one of our internal asserts somehow?

correct. That test used to fail when we `libmesh_assert_valid_unique_ids`

roystgnr:
I'd prefer something that's active in opt mode too (is that one even on in devel mode?) but as long as we're covering it somewhere, good enough for me.

PR #3015, opened 2021-09-10
Add elem_num_map and node_num_map accessors to ExodusII_IO

jwpeterson:
These weren't needed by anyone in the past, but we now have a case where we store elemental and nodal values directly from the Exodus file in arrays, and then need to index into those arrays later based on (mapped) elem ids. Having these maps will help make that indexing possible. Also adds a unit test at the request of @roystgnr but there's not much to test here.

### DIFF
--- a/include/mesh/exodusII_io.h
+++ b/include/mesh/exodusII_io.h
@@ -423,6 +423,27 @@ public:
    */
   const std::vector<std::string> & get_global_var_names();
 
+  /**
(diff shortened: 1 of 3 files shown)


roystgnr (approved):

jwpeterson:
It looks like CIVET is down currently so I can't check for sure, but I think the failures here were probably the tests we already knew were failing, so I'm going to go ahead an merge.

PR #3018, opened 2021-09-12
Make FEMap::map_fe_type usable outside library

lindsayad:
It took [Mac testing](https://civet.inl.gov/job/817866/) on idaholab/moose#18395
in order to discover what I think is undesirable behavior. We
define `map_fe_type` as a public static function, which to me makes
me think it should be safe to use outside the library, but then we
put its definition in the source file which makes it unusable outside
of the library since `static` forces internal linkage. So I think
we should move the definition to the header

### DIFF
--- a/src/fe/fe_map.C
+++ b/src/fe/fe_map.C
@@ -42,7 +42,6 @@
 namespace libMesh
 {
 
-inline
 FEFamily
 FEMap::map_fe_type(const Elem & elem)
 {


roystgnr:
This was obviously wrong, but I think you've misdiagnosed it - the problem isn't putting a `static` function in the source file (else we'd also need to put everything from `FEMap::build()` to `FEMap::map()` in the header!), it's declaring it `inline` afterward.  Not sure how I put that there, but it basically told `fe_map.C` "you don't need to generate a symbol for this if you don't want to, everyone's inlining it anyway," and it looks like Mac LLVM took us up on that.

Try leaving the function in the .C file, but removing the `inline` descriptor?

roystgnr (approved):
Sweet, this did fix it then?

Change looks good to me.  Might want to reword the commit message, though.

lindsayad:
> Sweet, this did fix it then?

Actually haven't verified yet (compiling now), but I fully expect it to. The `inline` explanation makes total sense. I don't know how I overthought that one.

> Might want to reword the commit message, though.

I re-worded the commit. Didn't know if I wanted to re-word the title post as then your follow-on comment (and the general thread of conversation) would look off

lindsayad:
Ok I verified that this fixes the issue on my Mac

PR #3020, opened 2021-09-19
Reduced Basis EIM update to add elem_id and qp arguments

dknez:
Add elem_id and qp arguments in RBParametrizedFunction, since this can be needed in some cases (e.g. if the RBParametrizedFunction uses mesh-based data).

### DIFF
--- a/examples/reduced_basis/reduced_basis_ex4/assembly.h
+++ b/examples/reduced_basis/reduced_basis_ex4/assembly.h
@@ -40,6 +40,7 @@ using libMesh::RealGradient;
 using libMesh::Elem;
 using libMesh::FEBase;
 using libMesh::subdomain_id_type;
+using libMesh::dof_id_type;
 using libMesh::Utility::pow;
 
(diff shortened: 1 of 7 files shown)


jwpeterson (approved):
Technically this is a public API change for `RBParametrizedFunction`, but I think I'm OK with allowing it since we already started the new release branch, and this is not a widely used API overall. 

I'm also a bit concerned that the argument list for `vectorized_evaluate()` continues to grow. It also looks like accidentally passing those three vectors of relatively similar type in the wrong order could be a bug waiting to happen... it may be possible to come up with a better design at some point.

roystgnr:
At the moment I think most of our CI runs with 8-byte `dof_id_type`, 4-byte `unsigned int`, and 2-byte `subdomain_id_type`, and when you're using vectors for all of those there's no automatic conversion and any misordering will be screamed about at compile time, so you're okay for library-internal usage.  If you're running apps with 4-byte `dof_id_type` then that'd be a bit more risky...

PR #3022, opened 2021-09-20
Update MetaPhysicL to 1.3.1

roystgnr:
This is necessary to fix "make check" with some compilers, and to fix
"make distcheck"

@jwpeterson - try this with the new libMesh RC?


roystgnr:
No, actually, it looks like the "make check" fixes were in 1.3.0 after all.  So 1.3.1 is *just* a distcheck fix.

jwpeterson:
Sure... to try this I could cherry-pick your commit onto the release branch, tag the RC, and create the release, then watch to see what `make distcheck` does... I'll let you know.

roystgnr:
"The good news is that make distcheck passed on the first try" - good enough for me; with this it worked on my system too.

PR #3023, opened 2021-09-21
Dual shape function modifications for 3D mortar and edge dropping

jbadger95:
This simply separates the computation of dual functions from the reinit routine so that we can compute dual shape functions on a different set of points that we reinit with. I'm not super happy with these fixes, should at least add some protections for making sure dual shape functions have been reinited before computation. MOOSE PR [#18885](https://github.com/idaholab/moose/pull/18885) has the corresponding changes for getting the mortar quadrature to initialize the dual shape functions. 

I'm not intending for this to be merged as is, its going to need a bit more work.

### DIFF
--- a/include/fe/fe.h
+++ b/include/fe/fe.h
@@ -452,6 +452,10 @@ public:
                        const std::vector<Point> * const pts = nullptr,
                        const std::vector<Real> * const weights = nullptr) override;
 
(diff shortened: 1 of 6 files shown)


jwpeterson:
My understanding is that this PR has been superseded by #3067 so I'm going to close it. Feel free to reopen if that's not the case.

PR #3024, opened 2021-09-22
Fix get_refspace_nodes(TET14)

roystgnr:
This wasn't breaking any unit tests that didn't try to do quadrature on
Tet sides, but as soon as I tried to extend HIERARCHIC shape functions
to cubics on Tets and do a project_solution with them, bam, broken.


roystgnr:
I plan to merge this as soon as I've seen it pass CI, since it's such an obvious fix, but remember that even closed tickets on Github can still have new comments added, in case we still need more discussion of the issue or mockery of me.

PR #3025, opened 2021-09-23
Reduced Basis EIM update: Add phi values

dknez:
In some cases we want to be able to interpolate nodal data onto the "EIM interpolation points", which are at quadrature points. To do this, we need access to the shape function values (i.e. the "phi" values). This MR provides this access.


jwpeterson (approved):
Technically this is an API change but again it's not a widely used API and the extra complication to try and make it backwards compatible does not seem like it would be worth it.

PR #3026, opened 2021-09-23
Hierarchic shape function support on tetrahedra

roystgnr:
In particular, we can go up to unlimited polynomial degree (except for roundoff error, which has us cap "unlimited" at around 11...) on Tet14.


roystgnr:
At some point in the near future we may want to revisit these shape functions, because their ordering isn't compatible with multiple element dimensions sharing the same nodes (when I did the 2D support that wasn't yet an option so I didn't even imagine reorienting 2D interior shape functions to match shape functions on sides of 3D elements), and because we might want to scale them differently, and because we might want to change our orientation rules (the comparison of Point values is robust to changes of id, like in the AMR simulations I was doing, but horribly fragile to mesh displacement) - but when we do that we'll want to create a new FEFamily enum for the new functions, rather than risk mangling reads of people's existing .xdr/.xda files.

PR #3028, opened 2021-09-24
Remove an unused variable.

drwells:
I noticed that there was a release candidate so I took a look - I have some more patches incoming. This one is just for master, though.

As of be42425a2ed (which is not in 1.7.0) these are not function arguments when LIBMESH_ENABLE_UNIQUE_ID is not defined.

### DIFF
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -158,7 +158,6 @@ void transfer_elem(const Elem & lo_elem,
 
           hi_node->set_unique_id(new_unique_id);
 #endif
-          libmesh_ignore(max_new_nodes_per_elem);
 
           /*
            * insert the new node with its defining vertex


jwpeterson:
What exactly was the error here? I looked at this briefly and it seems to me that we could get a "set but not used warning" if you remove the line you suggest, since `max_new_nodes_per_elem` is set regardless of `LIBMESH_ENABLE_UNIQUE_ID`, but it is only *used* if that define is set.

drwells:
I get an undefined variable warning:
```
  CXX      src/numerics/libmesh_dbg_la-dense_matrix_base.lo
../src/mesh/unstructured_mesh.C:161:26: error: use of undeclared identifier
      'max_new_nodes_per_elem'
          libmesh_ignore(max_new_nodes_per_elem);
                         ^
  CXX      src/numerics/libmesh_dbg_la-dense_matrix_blas_lapack.lo
```

The cause looks quite clear to me: at the top of the function we have
```cpp
                   unique_id_type max_unique_id,
                   unique_id_type max_new_nodes_per_elem,
#endif
```
so these are definitely not defined when that is not enabled.

jwpeterson:
OK, yes I see. Sorry, I was looking at an older version of unstructured_mesh.C.

jwpeterson (approved):

PR #3030, opened 2021-09-24
Fix clang warnings

drwells:
The version numbers here are for the versions of clang released with Linux distributions, not macs, so this generates a ton of warnings on macOS.


jwpeterson (approved):
This looks like a nice simplification to me. My only concern would be if one of the four new pragmas could itself generate a warning on an old enough compiler, but this does not seem to be the case for our "min clang" and "min gcc" testing, so I think it should be fine.

drwells:
We use a similar set up in deal.II and IBAMR - to the best of my knowledge any version of clang or GCC that supports C++11 will also support those flags.

PR #3033, opened 2021-09-27
Add coverage configuration for CIVET

loganharbour:


jwpeterson:
> * Coverage rate 57.38% is less than the required 60.0%

We set the bar low, but not low enough.

loganharbour:
> > * Coverage rate 57.38% is less than the required 60.0%
> 
> We set the bar low, but not low enough.

Worth a try 😆 

We're also going to see what the result is when we compile with -O0 (running now). We needed to do this in MOOSE in order to capture some complex templated methods that were not picked up with -O2.

loganharbour:
`-O2`: https://mooseframework.inl.gov/libmesh/docs/PRs/3033_O2/coverage/
`-O0`: https://mooseframework.inl.gov/libmesh/docs/PRs/3033/coverage/

Literally no difference!

roystgnr:
> Literally no difference!

At least we won't have to worry about de-optimization making the tests interminable, then?

I went into the fe/ directory on the report, since I was pretty confident we had decent coverage from the unit tests there, but I'd forgotten about all the weird cases we never expect anybody to use.  Like, 0D element support is nice, a great way to couple in some ODEs to your system or to make side elements generalize to 1D meshes or whatever, but we're never going to see someone instantiate them with SZABAB.  And 1D CLOUGH is there for completeness, but it's equivalent to 1D HERMITE so nobody ever uses it unless they have a simplex-based solver with variable dim.

I'll bet there's a bunch of cases like that which could be covered with trival unit tests.  I don't remember if I even bothered to instantiate any of the FE tests in 0D...

loganharbour:
Okay. Looks like this works.

The diff failed (the "no coverage report(s) were found" error) because of course there is no base report. Once this gets into master, @roystgnr you should be able to do a `git commit --amend` with a force push on a PR of your choosing (to get the new recipes) and we should get a diff.

roystgnr (approved):

loganharbour:
Once https://github.com/idaholab/civet_recipes/pull/155 goes in - this can be merged.

PR #3034, opened 2021-09-27
Exodus version choice fix for --enable-exodus

### LINKED ISSUE #3029: netcdf problems with 1.7.0
I get a bunch of link problems with the release candidate like
```
/usr/bin/ld: ./.libs/libmesh_dbg.so: undefined reference to `H5get_libversion'
/usr/bin/ld: ./.libs/libmesh_dbg.so: undefined reference to `nc_inq_grp_full_ncid'
/usr/bin/ld: ./.libs/libmesh_dbg.so: undefined reference to `nc_inq_grpname_full'
/usr/bin/ld: ./.libs/libmesh_dbg.so: undefined reference to `nc_inq_grps'
/usr/bin/ld: ./.libs/libmesh_dbg.so: undefined reference to `nc_inq_grpname'
/usr/bin/ld: ./.libs/libmesh_dbg.so: undefined reference to `nc_inq_grp_parent'
/usr/bin/ld: ./.libs/libmesh_dbg.so: undefined reference to `nc_def_var_deflate'
/usr/bin/ld: ./.libs/libmesh_dbg.so: undefined reference to `nc_def_grp'
/usr/bin/ld: ./.libs/libmesh_dbg.so: undefined reference to `nc_inq_grpname_len'
/usr/bin/ld: ./.libs/libmesh_dbg.so: undefined reference to `nc_inq_grp_ncid'
```

notably, those are all netcdf-4 functions - most of them are in `dgroup.c`, which doesn't seem to be compiled, which makes me think that `USE_NETCDF4` is not being set. However, `dgroup.c` does have a makefile rule generated for it, so I'm not sure why its not getting compiled.

roystgnr:
Unless someone explicitly requests Exodus 8, we don't want to try giving it to them unless we know they have HDF5 support.

This fixes #3029 on my system.

This should be backported to the release branch if it works for others too.

### DIFF
--- a/configure
+++ b/configure
@@ -54752,7 +54752,14 @@ fi
   # Check whether --enable-exodus was given.
 if test "${enable_exodus+set}" = set; then :
   enableval=$enable_exodus; case "${enableval}" in #(
-  yes|v811) :
+  yes) :
+    enableexodus=yes
+                                      if test "x$enablehdf5" = "xyes"; then :
+  exodusversion=v8.11
+else
+  exodusversion=v5.22
+fi ;; #(
+  v811) :
     enableexodus=yes
                                       exodusversion="v8.11" ;; #(
   new|v522) :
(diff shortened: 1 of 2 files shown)


jwpeterson:
Thanks @roystgnr. I will cherry-pick this onto the release branch and tag a new RC.

PR #3037, opened 2021-09-30
Updated how we evaluate observation points for EIM basis functions

dknez:
This update provides a default implementation of `RBParametrizedFunction::evaluate_at_observation_points()` which should work in most cases since it uses `RBParametrizedFunction::vectorized_evaluate()`.

### DIFF
--- a/src/reduced_basis/rb_parametrized_function.C
+++ b/src/reduced_basis/rb_parametrized_function.C
@@ -229,13 +229,71 @@ Number RBParametrizedFunction::get_parameter_independent_data(const std::string 
   return libmesh_map_find(libmesh_map_find(_parameter_independent_data, property_name), sbd_id);
 }
 
-std::vector<std::vector<Number>> RBParametrizedFunction::evaluate_at_observation_points(const RBParameters & /*mu*/,
-                                                                                        const std::vector<Point> & /*observation_points*/,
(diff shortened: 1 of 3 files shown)


src/reduced_basis/rb_parametrized_function.C
```diff
@@ -229,13 +229,71 @@ Number RBParametrizedFunction::get_parameter_independent_data(const std::string
...
+      Point obs_pt_ref_coords =
+        FEMap::inverse_map(
+          elem_ref.dim(),
+          &elem_ref,
+          obs_pt,
+          /*tolerance*/ TOLERANCE,
+          /*secure*/ true,
+          /*extra_checks*/ false);
+
+      std::vector<Point> obs_pt_ref_coords_vec;
```
jwpeterson, line 274: Simplify this initialization to one line:
```
std::vector<Point> obs_pt_ref_coords_vec = {obs_pt_ref_coords};
```

jwpeterson (approved):
It's another API change but it's an API that has recently been in flux anyway, so I'm fine with merging.

src/reduced_basis/rb_parametrized_function.C
dknez: I pushed this change.

jwpeterson:
FYI we have tests for this feature internally, but it is not tested in any libmesh examples, so the code coverage report seems accurate.

PR #3038, opened 2021-10-04
fix encapsulation of enable_unique_ids

BalticPinguin:
When configuring libMesh without unique_ids, I got the following errors (I use `--enable-werror`):
````
../master/src/mesh/unstructured_mesh.C: In member function ‘virtual void libMesh::UnstructuredMesh::all_second_order(bool)’:
../master/src/mesh/unstructured_mesh.C:1286:16: error: variable ‘max_new_nodes_per_elem’ set but not used [-Werror=unused-but-set-variable]
   unsigned int max_new_nodes_per_elem;
                ^~~~~~~~~~~~~~~~~~~~~~
../master/src/mesh/unstructured_mesh.C: In member function ‘virtual void libMesh::UnstructuredMesh::all_complete_order()’:
../master/src/mesh/unstructured_mesh.C:1500:16: error: variable ‘max_new_nodes_per_elem’ set but not used [-Werror=unused-but-set-variable]
   unsigned int max_new_nodes_per_elem;
                ^~~~~~~~~~~~~~~~~~~~~~
````
This is fixed with this PR.

Best regards

### DIFF
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -1279,13 +1279,13 @@ void UnstructuredMesh::all_second_order (const bool full_ordered)
    */
   std::map<std::vector<dof_id_type>, Node *> adj_vertices_to_so_nodes;
 
+#ifdef LIBMESH_ENABLE_UNIQUE_ID
   /*
    * The maximum number of new second order nodes we might be adding,
    * for use when picking unique unique_id values later
    */
   unsigned int max_new_nodes_per_elem;
 
-#ifdef LIBMESH_ENABLE_UNIQUE_ID
     unique_id_type max_unique_id = this->parallel_max_unique_id();
 #endif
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by jwpeterson — APPROVED** [review]
Thanks for this fix. I mentioned that it seemed like we should see this warning in #3028, so good to confirm that I was not crazy. It might be simpler to add a single
```
libmesh_ignore(max_new_nodes_per_elem);
```
line than to try a guard every instance where that variable is used, though? I can try making that change.

jwpeterson:
I pushed an update which just ignores the unused variable rather than guarding every use, but I don't have a build with unique ids disabled handy, so it would be great if you could let me know whether it still fixes the issue for you...

BalticPinguin:
Hi John,
sure, your fix is indeed simpler and cleaner; that option didn't come to my mind.
And yes, the current variant works well.

For the future, don't bother rejecting/closing my PR if there is a better solution, than you don't have to merge + revert just for the statistics :-) 
But I appreciate that :-)

jwpeterson:
@BalticPinguin I saw that your commit was on the master branch of your fork... I did not want to push something that would immediately generate a conflict for you... also I was not sure that my approach would actually get rid of the warning so figured it would be good to keep your fix in the history as well.

PR #3041, opened 2021-10-04
Support SIDE_HIERARCHIC on triangles

roystgnr:
This was originally going to be a trivial speedbump on the way toward adding SIDE_HIERARCHIC Tet support, but it turned out to be trickier than imagined (see the second commit) and I got pulled away from work this week, so let's at least get this bit in a PR.  Smaller PRs are better PRs, right?

Also, before I proceed further I want to ping @lindsayad (you were the one who wanted to use these soon, right?) to ask for a second opinion on my decision in 080c37c.  In the common formulations I've seen that have flux-like variables on element sides, nobody's going to be integrating functions of tangential gradients of those variables on those sides, but if I'm wrong and such integrals will often happen then we might need to reconsider whether it's safe for that to be a user-unfriendly process.


lindsayad:
Thus far we have gone in collocated pressure and velocity in our finite volume work, with Rhie-Chow allowing us to combat the checkerboard effect that arises from collocation. I often do get frustrated with Rhie-Chow and imagine a world in which we do our finite volume work on a staggered grid, but even there I do not think we would leverage this `SIDE` work. We would need something more like a dual mesh I think. I know in the beginning I expressed interest in `SIDE` but that was before I thought about it more. There is still a chance that we use `SIDE` just for storing mass fluxes but we'll see.

I believe that @YaqiWang and Griffin have interest in face variables.

PR #3048, opened 2021-10-07
Ban using broken PETScs with (par)metis.

drwells:
Also related to #3047 - we should just error out if the user asks for something impossible (in this case, use metis with petsc when petsc cannot be used).

### DIFF
--- a/configure
+++ b/configure
@@ -51949,6 +51949,10 @@ if test "${with_metis_include+set}" = set; then :
 fi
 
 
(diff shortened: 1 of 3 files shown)


roystgnr:
Is this the weird thing where we merge one PR into master in between two --enable-coverage CI jobs for another PR, and so those jobs have different base commits that they're tested against, but to merge coverage from multiple jobs they all have to be against the same base?

loganharbour:
You’re good. I can mess around with it after it’s merged. The easy way around this is to invalidate all of them so that they’re synced, but I was lazy and hopeful…

PR #3049, opened 2021-10-07
--enable-singleprecision compilation fixes

roystgnr:
Don't really have time to go through unit tests soon (and @jwpeterson was right, our FDM with single precision is garbage) but at least this gets us compiling again.


roystgnr:
Throwing on decimals is usually the safe thing to do!  I like relying on the raw int->Real conversion when I'm confident it's correct, but if we sometimes throw decimal points into new/changed code then we'll just fix it again later.  I don't know if anyone else has been futzing around with alternative precisions, but in the past we've often only had it in shape for releases and not always even then.  I only use the extended precision myself every few years, when I have some error convergence that is bottoming out at a questionable "low enough that it might be rounding error but high enough that I can't be sure" level.  (I just pushed the update for those to #2811, BTW; I think it ought to be merge-worthy now)

PR #3057, opened 2021-10-15
complement solver/preconditioner to string_to_enum.

BalticPinguin:
I have realized that the `BICGSTAB` and `SPARSELU` solver types were missing in the `solver_to_enum`-list.

Please let me know about the reformatting of `preconditioners`; I can remove this from the PR if prefered.

Best regards


BalticPinguin:
> those enums being added later and the maps not being updated at that time.

That would be also my explanation; and seemingly, they are not used very much anyway, or people don't use these solvers...

jwpeterson:
I'll cherry-pick this on the 1.7.0 branch as well.

PR #3059, opened 2021-10-20
Exodus ex_open fix

roystgnr:
ExodusII 8.11 keeps track of ex_open/ex_close, and we have a leak in parallel where we open files and don't close them before trying to open them a second time, and that leak triggers in examples/transient_ex1.  This should fix it, and also slightly improve ExodusII_IO efficiency in parallel.

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -1645,6 +1645,8 @@ void ExodusII_IO_Helper::initialize(std::string str_title, const MeshBase & mesh
   // The majority of this function only executes on processor 0, so any functions
   // which are collective, like n_active_elem() or n_edge_conds() must be called
   // before the processors' execution paths diverge.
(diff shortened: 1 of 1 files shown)


roystgnr:
Oh, this is just a mess.  Not only do we have redundant reads of nearly everything, we've broken encapsulation in `ExodusII_IO_Helper::read_edge_blocks`, which just stuffs data directly into a `MeshBase`.

roystgnr:
This is way too mammoth a fix to do this way now.  I don't want to backport hundreds of lines in the middle of a release.  I'll instead try sticking with the redundant open calls and just get redundant close calls to match.

roystgnr:
I think that first commit fixes the leak and that second fixes the race condition.  If this passes CI and my own tests then I want to merge and backport ASAP.  I wouldn't be surprised if there are other issues that Exodus 8 exposes that we haven't caught yet, but this at least fixes the biggest.

PR #3066, opened 2021-10-29
Calculator app improvements

roystgnr:
I added a few features to do some testing on IGA meshes, and they were useful enough to clean up and save for future uses too.

### DIFF
--- a/src/apps/L2system.C
+++ b/src/apps/L2system.C
@@ -50,7 +50,7 @@ void L2System::init_context(DiffContext & context)
 {
   FEMContext & c = cast_ref<FEMContext &>(context);
 
-  FEBase * elem_fe = nullptr;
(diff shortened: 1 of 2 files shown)


jwpeterson (approved):
We don't actually test the src/apps, do we? This would probably be non-trivial to set up, but they seem to get outdated quickly...

roystgnr:
Our tests cover the apps' compilation, but not any runs - hence the bugfix here, for a bug that can only trigger at runtime.

Might not be that hard to at least get a run or two out of each, though; just need to set up a test script for them, or one script for each test we want to show up as an independent result in the directory's "make check" summary.  It's been too long since I used most of the apps to come up with a good use case quickly for all of them, but if I get a chance I'll try it out with this calculator as a proof-of-concept next week.

PR #3067, opened 2021-10-29
Dual shape function modification

dewenyushu:
This is a re-opening of Jacob's draft PR on an initial implementation of dual basis modification for 3D contact and edge dropping (#3023)

Improvement on the efficiency and robustness of dual basis calculation will be examined in separate PRs.


dewenyushu:
@lindsayad @jbadger95 I made some changes to the dual basis calculation PR. It's now failing one test in stochastic tools somehow. I will continue to look into it. Aside from it, this should be ready for your review. 

I added back the calculation of dual shape coefficient using the default QRule, which can be turned off on the MOOSE level for msm elements that are partially in contact. This way we can protect the calculation of the dual shape function without dual coefficients being calculated first. Also since we _may_ still need the default dual shape function for the interior msm elements I suppose it's a better idea to retain this capability in libmesh.

jwpeterson:
> It's now failing one test in stochastic tools

That same test is failing on lots of PRs, so probably not related to this one.

jbadger95:
Have you looked into the cost or conditioning at all?

dewenyushu:
> Have you looked into the cost or conditioning at all?

I looked into the conditioning. There's significant increase of condition number (several orders of magnitude) when the element in contact becomes small (see the attached plot, condition number shoots up when the secondary element is about to be fully not in contact). I am working on an initial implementation of element/nodal scaling for improving this behavior. 

[Condition_Number_2D_Slide_Scaling.pdf](https://github.com/libMesh/libmesh/files/7464090/Condition_Number_2D_Slide_Scaling.pdf)

I have not looked deep into the cost yet. I suppose we have not come into a conclusion whether we will need customized quadrature points for the interior elements, so it is not very clear to me right now in terms of where and how to reduce the cost. 

Anyways, my intent is to take smaller steps to cleanup and merge what you've implemented first and optimize things (better scaling & improve efficiency) later, since I realize both tasks would take some more work than I expected.

jbadger95:
Is this the conditioning of the overall system or of the mass matrix for the dual coefficients?

(sorry for the delayed response, busy day)

jbadger95:
My biggest concern with conditioning is that on 3D hexas (quad faces) that if we don’t enrich the quadrature order then the dual mass matrix can potentially be singular. The poor conditioning of edge dropping is expected (without scaling), my concern here is just that for normal mortar problems (without edge dropping) using an inexact quadrature might cause problems.

Can you compare the maximum conditioning number for dual mass matrices (on an aligned 3D hexa mesh and maybe some twisted meshes or something with a bit more exotic mortar segment mesh) using the default mortar quadrature rule vs an enriched quadrature rule (or the default primal quadrature rule if you wish)?  

Do all the MOOSE tests pass with these changes?

dewenyushu:
> Is this the conditioning of the overall system or of the mass matrix for the dual coefficients?
> 
> (sorry for the delayed response, busy day)

This is the conditioning of the overall system.

dewenyushu:
> Can you compare the maximum conditioning number for dual mass matrices (on an aligned 3D hexa mesh and maybe some twisted meshes or something with a bit more exotic mortar segment mesh) using the default mortar quadrature rule vs an enriched quadrature rule (or the default primal quadrature rule if you wish)?

Thanks Jacob!

Just to make sure, by `enriched qrule` you mean the customized qrule computed in `ComputeMortarFunctor` in this PR (https://github.com/idaholab/moose/pull/18885), or do you mean a higher quadrature order is needed while computing the customized qrule?

> Do all the MOOSE tests pass with these changes?

Yes, all the MOOSE tests pass (except for the one stochastic tools that does not seem to be related).

jbadger95:
By enriched quadrature I mean an enriched mortar quadrature (e.g. `quadrature = SECOND` In mortar constraint). A first order quadrature on the triangle mortar segments does not exactly integrate first order quad elements, that’s why I’m worried about dual mass matrices being poorly conditioned.

dewenyushu:
> By enriched quadrature I mean an enriched mortar quadrature (e.g. `quadrature = SECOND` In mortar constraint). A first order quadrature on the triangle mortar segments does not exactly integrate first order quad elements, that’s why I’m worried about dual mass matrices being poorly conditioned.

Okay, I got what you mean here.

I tested this change (together with your MOOSE change) with two non-conforming blocks (hex 8 elements), default qrule, with and without edge dropping. Both work fine. 

I can try to get the condition number of the mass matrix and compare that for default and enriched quadrature to determine if enriched qrule is needed.

jbadger95:
Sounds good, sorry for pestering you I just didn’t get a chance to really test it so just wanted to make sure you had. If you’re convinced it’s not breaking anything I don’t have any objections to merging it. I’ll review it a bit later today.

dewenyushu:
> Sounds good, sorry for pestering you I just didn’t get a chance to really test it so just wanted to make sure you had. If you’re convinced it’s not breaking anything I don’t have any objections to merging it. I’ll review it a bit later today.

Not at all! Glad that we are on the same page. And I really appreciate your help looking into this.

src/mesh/unstructured_mesh.C
```diff
@@ -99,6 +99,9 @@ void UnstructuredMesh::copy_nodes_and_elements(const UnstructuredMesh & other_me
     //Preallocate Memory if necessary
     this->reserve_nodes(other_mesh.n_nodes());
 
+#ifdef LIBMESH_ENABLE_UNIQUE_ID
+    unique_id_type potentially_next_unique_id = _next_unique_id;
```
roystgnr, line 103: This seems like an important and distinct change, doesn't it?  Probably worth backporting?  Is it worth merging separately if this PR gets delayed any further?

include/fe/fe_base.h
```diff
@@ -591,7 +591,7 @@ class FEGenericBase : public FEAbstract
    * Compute the dual basis coefficients \p dual_coeff
    * the default qrule for element dimension and quadrature order is needed
```
roystgnr, line 592: The documentation here is an atavism now, yes?

src/fe/fe_base.C
```diff
@@ -743,38 +743,37 @@ void FEGenericBase<OutputType>::compute_shape_functions (const Elem * elem,
     this->_fe_trans->map_div(this->dim, elem, qp, (*this), this->div_phi);
 }
 
+
+// Note I am passing points instead of using members because size changes between calls
```
roystgnr, line 747: This seems to be the key to the whole commit here, yet this sentence is too cryptic for me to understand it.

include/fe/fe_base.h
```diff
@@ -795,7 +795,7 @@ void FEGenericBase<OutputType>::compute_dual_shape_functions ()
 }
 
 template <typename OutputType>
-void FEGenericBase<OutputType>::compute_dual_shape_coeffs(const std::vector<Real> & /*JxW*/, std::vector<std::vector<OutputShape>> & /*phi_vals*/)
+void FEGenericBase<OutputType>::compute_dual_shape_coeffs(const std::vector<Real> & /*JxW*/, const std::vector<std::vector<OutputShape>> & /*phi_vals*/)
```
roystgnr, line 798: As long as there are more changes to make anyway, let's just turn this commit into a fixup on top of the compilation-breaking commit it fixed?

include/fe/fe_abstract.h
```diff
@@ -150,9 +150,17 @@ class FEAbstract : public ReferenceCountedObject<FEAbstract>
...
+  /**
+   * This re-computes the dual shape function coefficients.
+   * The dual shape coefficients are utilized when calculating dual shape functions.
+   * This does not support InfFE type.
+   */
+  virtual void reinit_dual_shape_coeffs (const Elem * /*elem*/,
+                                         const std::vector<Point> & /*pts*/,
+                                         const std::vector<Real> & /*JxW*/)
+  {
+    libmesh_error_msg("Dual shape coefficient calculation has not been implemented for this FE type.");
```
roystgnr, line 162: Likewise with a fixup for this commit, if I'm understanding correctly and it broke --enable-ifem builds.

src/mesh/unstructured_mesh.C
dewenyushu: Yeah, this is not closely related to this PR. Let me have it in a separate one.

include/fe/fe_base.h
dewenyushu: Good catch. The docstring is outdated. Let me update it

src/fe/fe_base.C
dewenyushu: Sure, let me clarify things a bit more.

src/mesh/unstructured_mesh.C
jbadger95: This was actually a @lindsayad change, [#3013](https://github.com/libMesh/libmesh/pull/3013) (sorry for including it here, I figured it would resolve on rebase anyway)

dewenyushu:
Okay, I check on the conditioning of the mass matrix using FIRST and SECOND quadrature. I pick the maximum condition number in all msm element throughout the simulation. This is done for three cases:

****Case 1**, two block compression**
Maximum condition number:
FIRST order -  8.73517; SECOND order - 8.73517


****Case 2**, two block compression, rotated top block** 
Maximum condition number:
FIRST order -  16.7234; SECOND order - 16.308



****Case 3**, two block sliding**
Maximum condition number:
FIRST order -  236.946; SECOND order - 235.013 (maximum condition number at around the point when one element is to be fully not in contact, and then reduces to ~3 in both cases)

jbadger95:
Awesome, that’s interesting how little it affects the conditioning!

jbadger95 (approved):
Looks good to me, thanks for finishing this up!

include/fe/fe_abstract.h
```diff
@@ -150,6 +150,27 @@ class FEAbstract : public ReferenceCountedObject<FEAbstract>
...
+  }
+
+  /**
+   * This re-computes the dual shape function coefficients using DEFAULT qrule.
+   * This does not support InfFE type.
+   */
+  virtual void reinit_default_dual_shape_coeffs (const Elem * /*elem*/)
+  {
+    libmesh_error_msg("Dual shape coefficient calculation has not been implemented for this FE type.");
```
lindsayad, line 171: Do not all classes go through FE? Just wondering why this is a run time error as opposed to pure virtual

src/fe/fe_base.C
```diff
@@ -743,38 +743,40 @@ void FEGenericBase<OutputType>::compute_shape_functions (const Elem * elem,
...
+// element changes accordingly.
 template <>
-void FEGenericBase<Real>::compute_dual_shape_coeffs (const QBase & default_qrule)
+void FEGenericBase<Real>::compute_dual_shape_coeffs (const std::vector<Real> & JxW, const std::vector<std::vector<OutputShape>> & phi_vals)
 {
   // Start logging the dual coeff computation
   LOG_SCOPE("compute_dual_shape_coeffs()", "FE");
 
-  unsigned int sz=phi.size();
-  libmesh_error_msg_if(!sz, "ERROR:  dual basis should be computed after the primal basis");
```
lindsayad, line 753: const

include/fe/fe_abstract.h
lindsayad: Ah I see that InfFE does not. Still I wonder if it would be nicer to have this be pure virtual and put the not implemented error in InfFE? I don’t know what the style preference is for @jwpeterson and @roystgnr

jwpeterson: I don't have a strong feeling either way for FE since as you mentioned the only other implementation is InfFE

roystgnr: I'm happy either way.  We tend to make our most "critical" methods pure virtual, to try to set an example to subclass authors that implementation is something only to be avoided if you can't help it, but the definition of "critical" is a matter of opinion, and especially in this case where we're not expecting any completely new subclasses any time soon the matter seems just academic.

dewenyushu: Thanks @lindsayad @jwpeterson  @roystgnr for your inputs - Let me make this as pure virtual and add the not implemented error in InfFE

dewenyushu:
I cannot see the details in Civet test for the Coverage for some reason. Could someone help me look into it? - Is it because the functions in InfFE are not tested?

include/fe/fe_abstract.h
dewenyushu: Moving it to InfFE does seem to cause some extra test issue. I could not see the details in Civet to make sure whether this is only coverage though. I'll just keep this inside of `fe_base` instead

src/fe/fe.C
```diff
@@ -298,38 +298,58 @@ void FE<Dim,T>::reinit(const Elem * elem,
...
-          this->compute_dual_shape_coeffs(default_qrule);
-          // only compute the dual coefficient once
-          this->calculate_dual_coeff = false;
-        }
-        // the dual shape functions relies on the customized shape functions
+        // Check if we need to calculate the dual coefficients based on the default QRule
+        // We keep the default dual coeff calculation for the initial stage of the simulation,
+        // as well as interior elements that are fully in contact
+        // Currently, we re-compute dual_coeff for all msm elements by setting `calculate_default_dual_coeff' = false
+        // in `Assembly::reinitDual` and use the customized QRule for calculating the dual shape coefficients
```
jwpeterson, line 305: This comment is referring to MOOSE code "Assembly::reinitDual" and MOOSE jargon "msm". If you would like to keep it here, maybe just clarify that is the case?

jwpeterson (commented):
I don't know enough about the details of this PR to be a final Approver, @lindsayad please merge this when you are ready.

include/fe/fe_abstract.h
lindsayad: The coverage error has been present for every single completed testing of this PR: https://civet.inl.gov/pr/14815/ so the issue is not this single function. I suspect the orange box is due to the fact that the coverage % of this PR is less than the suggested 90%

lindsayad: next time let's not force push so quickly

lindsayad: Yea this is from the coverage testing script:
```
          if coverage_passes $COVERAGE_GENERATE_DIFF_NEW_RATE $WARN_NEW; then
            PASSED_COVERAGE+=("New line coverage rate $COVERAGE_GENERATE_DIFF_NEW_RATE%% passes the suggested $WARN_NEW%%")
          else
            WARNED_COVERAGE+=("New new line coverage rate $COVERAGE_GENERATE_DIFF_NEW_RATE%% is less than the suggested $WARN_NEW%%")
            RETURN_CODE=1
```

lindsayad: @jwpeterson was right ... even though the timing was very suspiciously correlated with the move to InfFE and back it was not caused by it. @loganharbour pushed a commit fixing the storage logic yesterday afternoon (it had broken last Thursday)

dewenyushu: Thanks for checking on this! It was such a coincidence that my update was at the same time frame as Logan's.

PR #3069, opened 2021-11-01
Only write Xdr header info on proc 0

dknez:
Fixed race condition in writing header in RBEvaluation::write_out_vectors().


roystgnr (approved):
Ignore the stochastic_tools failure in Test Moose.  This is like the third false positive I've seen from that recently.

PR #3071, opened 2021-11-03
Add SVD in preconditioner type

dewenyushu:
Add SVD in the preconditioner type in order to monitor condition number for the system. 

This is especially useful in evaluating the conditioner number for the condensed system produced by the variable condensation preconditioner (https://github.com/idaholab/moose/pull/16017)

Ping @fdkong

### DIFF
--- a/include/enums/enum_preconditioner_type.h
+++ b/include/enums/enum_preconditioner_type.h
@@ -45,6 +45,7 @@ enum PreconditionerType : int {
                          USER_PRECOND,
                          SHELL_PRECOND,
                          AMG_PRECOND,
+                         SVD_PRECOND,
                          // Invalid
                          INVALID_PRECONDITIONER};
 }
--- a/src/numerics/petsc_preconditioner.C
+++ b/src/numerics/petsc_preconditioner.C
@@ -212,6 +212,11 @@ void PetscPreconditioner<T>::set_petsc_preconditioner_type (const Preconditioner
       CHKERRABORT(comm,ierr);
       break;
 
+    case SVD_PRECOND:
+      ierr = PCSetType (pc, const_cast<KSPType>(PCSVD));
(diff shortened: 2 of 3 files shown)


roystgnr (approved):

fdkong:
@roystgnr Do we have the oldest petsc libMesh intends to support in civet? 

I am not sure when SVD was introduced into PETSc

jwpeterson:
@fdkong I found this commit from 2010 where "PCSVD" was first added: petsc/petsc@27c671223c1e7d72e35e89362b1f8e2d6e82ab2d
It goes all the way back to PETSc 3.2 so I don't think we need to guard it.


I'm not against adding this, but I'm also not sure whether this is needed/what it does? Normally one would just do `-pc_type svd` on the command line and it would be used, right?

dewenyushu:
Thanks for checking on this @jwpeterson !

> I'm not against adding this, but I'm also not sure whether this is needed/what it does? Normally one would just do `-pc_type svd` on the command line and it would be used, right?

I agree the `-pc_type svd` is what is normally needed in the `Executioner` block. But this would be helpful in the recently developed special preconditioner type (see idaholab/moose#16017), inside of which it is calling different preconditioner types. Specifically, its input block looks like this:

[Preconditioning]
  [vcp]
    type = VCP
    full = true
    lm_variable = 'lm'
    primary_variable = 'u'
    preconditioner = 'AMG' **<----- I would like SVD available here, too, for research purpose**
    is_lm_coupling_diagonal = true
  []
[]

fdkong:
>  -pc_type svd on the command line and it would be used, right?

We can not do that for a libmesh-wrapped petsc preconditioner. We might improve that in the future

fdkong (approved):

jwpeterson (approved):
Thanks, this looks good to me.

PR #3072, opened 2021-11-03
Set next unique id in copy_nodes_and_elements

dewenyushu:
This is @lindsayad 's fix for getting correct `unique_id` for nodes. Here is his comment:

We were currently setting `unique_id` for nodes after the call to `add_point` which means that `_next_unique_id` will end up incorrect if we don't also call `set_next_unique_id` after adding all our nodes. Note that this problem doesn't exist for elements because we set the `unique_id` before calling `add_elem` which results in the `MeshBase` object correctly setting `_next_unique_id` on its own

### DIFF
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -306,6 +306,9 @@ void UnstructuredMesh::copy_nodes_and_elements(const UnstructuredMesh & other_me
     //Preallocate Memory if necessary
     this->reserve_nodes(other_mesh.n_nodes());
 
+#ifdef LIBMESH_ENABLE_UNIQUE_ID
+    unique_id_type potentially_next_unique_id = _next_unique_id;
+#endif
     for (const auto & oldn : other_mesh.node_ptr_range())
       {
         processor_id_type added_pid = cast_int<processor_id_type>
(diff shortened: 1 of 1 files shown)


roystgnr:
This is one of those things we never caught on our own because the existing unique_id() values all turn out fine, it's just that when you then take that mesh through *another* mesh generator or through adaptive refinement, subsequent unique_id() values can be duplicates?

Sorry I didn't notice that commit earlier, and thanks for splitting it out; IMHO we ought to get it in and backported quickly.

roystgnr (approved):

lindsayad (approved):
I'm so spacey. Somehow I thought that I'd already gotten this commit into upstream. Thanks you @dewenyushu

lindsayad (changes requested):
Wait a second ... I think there is not a need for this after #3013

jwpeterson:
Yeah I think this came from someone's old branch that had not been rebased. One way to test would be to rebase this PR on master and see if the commit just vanishes.

lindsayad:
It's not even that. I wish we had more of a history on #3013, but I think @roystgnr actually saw issues with the implementation that is in this commit and and then I fixed it up and that's what we merged. #3013 even adds a test. I believe we actually had some of @jbadger95's mortar tests (now merged in MOOSE) that would not have run without the fix in #3013

jwpeterson:
I see. I was just confused by the commit message being the same in 73cd75e2b. So I think what you are saying is that we should just close this PR.

lindsayad:
Yes I think so.

dewenyushu:
Oh I did not realize this has been merged. Sorry about the confusion. Let me just close this one.

roystgnr:
My apologies.  I remembered working on a "next unique id was left out of sync with new unique ids" PR, but I didn't recall where and I just assumed it was in a different part of the code than this.

PR #3074, opened 2021-11-03
RB update: Enable EIM on element sides

dknez:
Previously we only enabled EIM on element interiors. This update enables EIM on element side as well. This does not change any existing functionality, it just adds new functionality for EIM on element sides.


jwpeterson (approved):
As discussed briefly with @roystgnr over IM, we will have test coverage for this feature internally, it is a bit too involved to make it into a libmesh example or unit test.

PR #3076, opened 2021-11-04
SolutionHistory based on HistoryData

vikramvgarg:
SolutionHistory derived classes were maintaining history with ad hoc maps from time to timestamps and filenames or vectors. To maintain a consistent, less error prone history and ensure a clean interface between SolutionHistory and MeshHistory (later PR):
1) A single map is now maintained by the SolutionHistory base class. 
2) This map takes time to the new HistoryData type objects.

HistoryData derived classes (currently File and Memory) interface between SolutionHistory and the disk or memory. SolutionHistory interfaces between TimeSolver and HistoryData.  In the future, whenever a combined Solution and MeshHistory is being maintained, HistoryData will allow us to maintain a convenient interface between MeshHistory and SolutionHistory.

The PR also includes fixes to TimeSolver handling of initial conditions, and adjoint examples 5 and 7.

### DIFF
--- a/src/solvers/file_solution_history.C
+++ b/src/solvers/file_solution_history.C
@@ -19,6 +19,7 @@
 #include <iostream>
 // Local includes
 #include "libmesh/file_solution_history.h"
+#include "libmesh/file_history_data.h"
 
 #include "libmesh/diff_system.h"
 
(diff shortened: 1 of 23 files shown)


src/solvers/file_solution_history.C
```diff
@@ -284,8 +181,13 @@ void FileSolutionHistory::retrieve(bool is_adjoint_solve, Real time)
   }
   else
   {
-    // Read in the primal solution stored at the current recovery time from the disk
-    _system.get_equation_systems().read (stored_sols->second, READ, EquationSystems::READ_DATA | EquationSystems::READ_ADDITIONAL_DATA);
+    // // If we are not in the adjoint loop, we could be in the primal loop again having solved
+    // // only the primal problem, or in a primal postprocessing stage to evaluate a QoI for example.
+    // // If we can find an adjoint solution file, read that in, else read in the primal file
+    if(dynamic_cast<FileHistoryData &>(*(stored_datum->second)).get_adjoint_filename().empty())
```
roystgnr, line 187: Why are we relying on this rather than is_adjoint_solve?

vikramvgarg: Thanks for catching this. Will fix.

roystgnr: > See lines 510 to 526

From IMs, copying it here for the record.

Yeah, this makes sense now.  In hindsight we really need to rework this API, but this PR isn't creating any new deficiencies, just working around the original flaws ... and I'm not sure what the right long-term fix is yet, either.

roystgnr (approved):

roystgnr:
Let's wait a few days in case anyone else wants to look it over; I'll merge Friday if there are no objections.

PR #3078, opened 2021-11-08
Use more macros from autoconf submodule

roystgnr:
I needed these macros in ACSM for https://github.com/libMesh/TIMPI/pull/77, but as long as they're cleaned up and available in the submodule, we might as well pare a few hundred lines of redundancy out of the libMesh versions.


roystgnr:
@loganharbour says the coverage extraction failure in those tests is because "A WIP script change for TIMPI coverage got in on accident", and that he's "Finishing up donating blood right now - I'll fix it in 10 or so when I'm home"  I swear I read all the way through our Software Quality Assurance Plan, and I felt duly impressed with the level of care that has to be undertaken for nuclear work, but somehow I must have missed the paragraph explaining that anyone who commits a bug must Pay For It ... In Blood!

loganharbour:
> but somehow I must have missed the paragraph explaining that anyone who commits a bug must Pay For It ... In Blood!

Welcome to the MOOSE team 🩸

jwpeterson (approved):
Looks good to me, but I'm wondering if we are going to need PRs/CI/testing on the autoconf-submodule repository as it gets more m4 code pushed to it..? I guess the current development approach is that you just push directly to autoconf-submodule, and then do a submodule update in libmesh to allow your changes to be tested?

jwpeterson:
> if we are going to need PRs/CI/testing on the autoconf-submodule repository

FWIW I have no idea how we would test flat m4 files... but maybe some basic stuff like trailing whitespace checks would be a good start.

loganharbour:
If you're only ever using `autoconf-submodule` as a submodule at a hash (standard users should never checkout and pull latest), what does it matter?

roystgnr:
I think our tests basically would have to run a hundred `configure --enable-foo --with-bar=baz` combinations, then parse the resulting `libmesh_config.h` files and maybe a few `Makefile`s to make sure that the stuff we thought we were turning on actually got turned on.  Maybe even do a `make V=1 just_one_file` if we could figure out how to do that portably, and make sure the flags we expect to be there are still there.

roystgnr:
I was going to hold off on merging this until we were forced to rely on some of the changes anyway in a future TIMPI merge ... but thanks to that embarrassing bug @lindsayad just fixed I guess we'll be needing a TIMPI update very soon regardless, in which case we might as well not delay here.

PR #3080, opened 2021-11-09
Fix Apple M1 JIT

### LINKED ISSUE #3060: FParser JIT fails on Apple Silicon
In opt mode the `FParser` JIT compile (or rather the dlopen) fail. In dbg mode it succeeds. Looks like the only thing that makes a difference here is the`-g` option. So I'll leave debug symbols in in all modes now. 🤷‍♂️

dschwen:
Fixes the issue locally on my M1, but I'd love to hear any suggestions why this failed in the first place...

Closes #3060

### DIFF
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -17,6 +17,7 @@ using namespace FPoptimizer_CodeTree;
 #include <cstdio>
 #include <iomanip>
 #include <sstream>
+#include <algorithm>
 
 // hashing complex numbers
 namespace std {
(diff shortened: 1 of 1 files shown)


cticenhour (approved):
Confirmed that this fixes the failing JIT tests on my M1 build as well.

jwpeterson (approved):
Strange bug indeed.

From [here](https://www.cplusplus.com/doc/tutorial/ntcs/) I found this: 

> Sequences of characters enclosed in double-quotes (") are literal constants. And their type is, in fact, a null-terminated array of characters. This means that string literals always have a null character ('\0') automatically appended at the end.

so maybe there is some weird bug on your compiler where `std::strings` constructed from string literals still retain a null-termination character?

dschwen:
Not just one null character, dozens. The compiler is clang 12.

jwpeterson:
OK, well if you can somehow reproduce it with a simple test code, sounds like it would be a good bug report for clang? Does the behavior change at all if you throw in std::string constructors all over the place, e.g.
```
std::string command = std::string(FPARSER_JIT_COMPILER) + std::string(" -O2 -shared -rdynamic -fPIC ");
```
?

roystgnr:
Yeah, this very much looks like a compiler bug to me.  Admittedly std::string is allowed to preallocate more space than necessary, and can leave null characters or whatever it wants in the extra space, and almost certainly (maybe certainly after C++11?) has at least one null character just to make the `c_str()` implementation trivial ... but `rbegin()` isn't even allowed to point to that first null, much less to point a dozen characters past it.

dschwen:
no success so far in creating an MWE I'll have to dig deeper into the compiler options that conda build sets

PR #3083, opened 2021-11-10
fix vector-index out of range

BalticPinguin:
I have another micro-patch:
I don't understand the context here in more detail, so I am not fully sure that this is the correct fix.
But at least now we don't write out of bounds.


jwpeterson (approved):
Thanks, nice catch! And very strange bug.

PR #3091, opened 2021-11-18
Fix "make coverage -j"

roystgnr:
This should have been an obvious race condition, but I guess I never triggered it before adding coverage support to TIMPI.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -706,6 +706,9 @@ lcov-reset:
 	$(lcov_bin)/lcov --directory . --zerocounters
 
 coverage: lcov-reset check lcov-report
+	$(MAKE) lcov-reset
+	$(MAKE) check
+	$(MAKE) lcov-report
 
 CLEANFILES += src/apps/*.gcda src/apps/*.gcno
 


roystgnr:
That test failure can't possibly have anything to do with this PR, but it looks distressingly real.

Paging @lindsayad since the first error looks related to MetaPhysicL+TIMPI interaction ... any ideas?  I'd say it looked like something was previously relying on an indirect header inclusion that's since gone away, but we haven't replaced any headers with forward declarations recently in TIMPI or MetaPhysicL, and regardless of the problem I don't see how it could have made it through Moose CI to trouble us here.

lindsayad:
Lol, what in the world...

lindsayad:
Oh I know what is going on ... we are not configuring libmesh (and hence metaphysicl) with `--with-future-timpi-dir` whereas in MOOSE I've assumed that libmesh will always be configured that way and haven't added any guards to my code to gracefully compile if its not. So two solutions: configure here with `--with-future-timpi-dir` or I remove my assumption that MOOSE will always have libmesh compiled with `--with-future-timpi-dir` ...

roystgnr:
Well, we certainly don't want to make users configure libMesh with `--complicated-argument-just-to-make-us-self-compatible` every time they want to communicate classes from one of our enabled submodules using classes from another.

But we can probably work around this.  We're already using `AX_SUBDIRS_CONFIGURE` for contrib/metaphysicl; we can just add `--with-future-timpi-dir` to that.

roystgnr:
Except - I'm not sure I actually get that option.  In scripts I see we're setting `--with-future-timpi-dir` to the eventual install directory, but we need `make check` to work pre-installation, too; if I try to just set `--with-future-timpi-dir=$prefix` (the libMesh $prefix) then contrib/metaphysicl/test/ executables fail to link for me.  We aren't allowed to pass both `--with-timpi=/one/dir --with-future-timpi-dir=/another/dir`, so what do we do?.

lindsayad:
If we want to run `make check` before `make install` (which I understand we ***should*** be able to ... that's one of the points of `make check` ... to see if you really want to `make install`), then I really have no solution other than turning TIMPI-required MetaPhysicL testing over to something like `make installcheck`

To the immediate issue at hand, I think I'll just go change the MOOSE code such that we can compile when `METAPHYSICL_HAVE_TIMPI` is not defined

lindsayad:
Oh wait ... that's not an easy solution. I can make MOOSE compile but the tests relying on `METAPHYSICL_HAVE_TIMPI` then won't pass and we don't currently have a way to disable them. We could read `metaphysicl_config.h` though and add an option to the MOOSE test spec for disabling a test if `!defined(METAPHYSICL_HAVE_TIMPI)`

PR #3092, opened 2021-11-22
Automatic support for configuring MetaPhysicL to use TIMPI

roystgnr:
With any luck this will fix the CI regression that we first triggered in #3091

### DIFF
--- a/m4/libmesh_metaphysicl.m4
+++ b/m4/libmesh_metaphysicl.m4
@@ -58,7 +58,28 @@ AC_DEFUN([CONFIGURE_METAPHYSICL],
           METAPHYSICL_INCLUDE="-I\$(top_srcdir)/contrib/metaphysicl/src/numerics/include -I\$(top_srcdir)/contrib/metaphysicl/src/core/include -I\$(top_srcdir)/contrib/metaphysicl/src/utilities/include"
           AC_DEFINE(HAVE_METAPHYSICL, 1, [Flag indicating whether the library will be compiled with MetaPhysicL support])
           AC_MSG_RESULT(<<< Configuring library with MetaPhysicL support >>>)
-          AX_SUBDIRS_CONFIGURE([contrib/metaphysicl],[--with-cxx-std=20$acsm_cxx_version])
+
+          dnl MetaPhysicL needs to know which TIMPI library to link to
+          dnl for "make check"
+          my_method=dbg
+          AS_IF([test "x$build_devel" = xyes],
+                [my_method=devel])
+          AS_IF([test "x$build_prof" = xyes],
+                [my_method=prof])
+          AS_IF([test "x$build_oprof" = xyes],
+                [my_method=oprof])
+          AS_IF([test "x$build_opt" = xyes],
+                [my_method=opt])
+
+          dnl Autoconf doesn't define $abs_top_srcdir at this point;
+          dnl here's a trick from GraphicsMagick:
+          my_top_srcdir="$(cd $srcdir && pwd)"
+
+          dnl FIXME: setting TIMPI_DIR, even to something invalid, is
+          dnl currently the best way to get MetaPhysicL to recognize
+          dnl that we really want TIMPI.
+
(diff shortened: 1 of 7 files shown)


roystgnr:
And of course it doesn't work in dbg-only builds, because we currently hard-code `-ltimpi_opt` in MetaPhysicL timpi.m4 and configure.ac

roystgnr:
Time to get rid of that Minimum gcc version test, I guess, at least until we have a PETSc-with-gcc-7 module to update to for testing C++17 support.

We need to build MetaPhysicL with TIMPI support now to pass the new MOOSE tests, so we need newer MetaPhysicL now to add configure options that let us build with pre-installed TIMPI and still pass contrib/metaphysicl "make check", so we need newer TIMPI to add the new SemiPermanent class that newer MetaPhysicL needs, so we need gcc 5 or better for the C++14 support that newer TIMPI needs.  I think that's the whole of it, but at this point if you told me that we had to swallow gcc 7 to catch the bird to catch the spider to catch the fly I'd just shrug.

The other failures I'm seeing here look real, and I can reproduce them; I'll see if I can figure out what's causing them.

roystgnr:
Is everything actually building?  Really hoping we can get CI going soon on other PRs again.

Last thing I'm worried about is whether the `TIMPI_DIR` submodule configure argument here interferes with the `--with-future-timpi` stuff going on in Moose scripts.

roystgnr:
Testing it myself: adding `--with-future-timpi-dir=FOO` doesn't leave us with a working MetaPhysicL `make check`, so there's clearly still work to do ... but that's the same situation we're in *now*, so this still seems like a uniform upgrade.

roystgnr:
@jwpeterson reports that this works with gcc 7 (which will be our new minimum gcc once we have modules available for CI) so I'm not going to worry about the Minimum gcc tests, but he also reminded me to try out a "make distcheck", and that's failing during the MetaPhysicL tests.  Still trying to figure it out; if I can't do so soon though I might just merge this anyway as an interim step so we can get the Test Moose recipe going again on other PRs.

roystgnr:
With those last m4 changes I'm seeing distcheck pass too; if everything but the min gcc tests here is still fine then we'll merge.

m4/libmesh_metaphysicl.m4
```diff
@@ -58,7 +58,28 @@ AC_DEFUN([CONFIGURE_METAPHYSICL],
...
+
+          dnl Autoconf doesn't define $abs_top_srcdir at this point;
+          dnl here's a trick from GraphicsMagick:
+          my_top_srcdir="$(cd $srcdir && pwd)"
+
+          dnl FIXME: setting TIMPI_DIR, even to something invalid, is
+          dnl currently the best way to get MetaPhysicL to recognize
+          dnl that we really want TIMPI.
+
+          AX_SUBDIRS_CONFIGURE([contrib/metaphysicl],[[--with-cxx-std=20$acsm_cxx_version],[CPPFLAGS=-I$my_top_srcdir/contrib/timpi/src/algorithms/include/ -I$my_top_srcdir/contrib/timpi/src/parallel/include/ -I$my_top_srcdir/contrib/timpi/src/utilities/include/ -I$ac_abs_top_builddir/contrib/timpi/src/utilities/include/ $CPPFLAGS],[LDFLAGS=-L$ac_abs_top_builddir/contrib/timpi/src/ $LDFLAGS],[TIMPI_DIR=$my_top_srcdir/contrib/timpi/],[--with-timpi-method=$my_method]])
```
lindsayad, line 82: I thought that I'd set things up such that `TIMPI_DIR` needed to point to a directory whose sub-directories were `bin, include, lib` but here it looks like we are now pointing to a source directory. What else has changed such that this works? I see all the proper flags getting set here so now I'm wondering if `TIMPI_DIR` is even relevant at all?

roystgnr: A mass of manually-set CPPFLAGS and LDFLAGS changed such that this works.  TIMPI_DIR is only relevant because its existence is at the moment what convinces timpi.m4 that we want to try to configure with TIMPI, so that it has the chance to discover that the manual flags work.  In hindsight the m4 idioms I was using a decade ago when some of these m4 files were originally written were not the appropriate idioms to import into libMesh.  In the long run I want to rewrite probably 75% of timpi.m4 and 25% of timpi_control_args.m4 but for this PR I was shooting for the minimal changeset that would get CI working again.

PR #3097, opened 2021-12-06
Fixes + workarounds for GCC 11 warnings, runtime error

roystgnr:
The first two commits are fixes for sensible warnings that have become more zealous or more clever.  The third commit is a workaround for a buggy warning, which someone already reported in gcc bugzilla.

The fourth commit is a workaround for a buggy combination of optimization flags that somehow makes it possible for `std::stable_sort` to corrupt data.  I've added it to the Ubuntu bug database, and if I don't get much traction there I'll make sure it's replicable with a from-scratch gcc and I'll add it to gcc bugzilla too.

In the meantime, I recommend everyone learn from my mistakes: avoid Ubuntu 21.10 and avoid gcc 11.  Seriously: **makes it possible for `std::stable_sort` to corrupt data!**  This was a really picky bug to distill, and I can't seem to trigger any other assertion failures or segfaults so far when testing with this workaround, but until someone manages to figure out what went wrong it might be safest to steer clear of the risk of finding the next trigger.

If anyone happens to have a non-Ubuntu gcc 11 install handy, I'd love to know whether my failure case breaks for them too:

```
  std::vector<std::pair<std::tuple<double,double>,int>> testvec(1900, {{1,2},12345}); // 18- works
  std::stable_sort(testvec.begin(), testvec.end());

  for (auto & p : testvec)
    if (p.second != 12345)
      abort();
```

fails at runtime whenever compiled with the combination `-O2 -funroll-loops`.

### DIFF
--- a/src/solvers/petscdmlibmeshimpl.C
+++ b/src/solvers/petscdmlibmeshimpl.C
@@ -907,7 +907,9 @@ static PetscErrorCode  DMSetUp_libMesh(DM dm)
     ierr = DMSNESSetFunction(dm, SNESFunction_DMlibMesh, (void *)dm); CHKERRQ(ierr);
     ierr = DMSNESSetJacobian(dm, SNESJacobian_DMlibMesh, (void *)dm); CHKERRQ(ierr);
     if (dlm->sys->nonlinear_solver->bounds || dlm->sys->nonlinear_solver->bounds_object)
-      ierr = DMSetVariableBounds(dm, DMVariableBounds_libMesh); CHKERRQ(ierr);
+      {
+        ierr = DMSetVariableBounds(dm, DMVariableBounds_libMesh); CHKERRQ(ierr);
+      }
   }
   else {
     /*
--- a/tests/parallel/parallel_test.C
+++ b/tests/parallel/parallel_test.C
@@ -151,18 +151,16 @@ public:
 
   void testBroadcast()
   {
-    std::vector<unsigned int> src(3), dest(3);
-
-    src[0]=0;
-    src[1]=1;
-    src[2]=2;
+    // Workaround for spurious warning from operator=
+    // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100366
+    std::vector<unsigned int> src{0,1,2}, dest(3,0);
 
     if (TestCommWorld->rank() == 0)
       dest = src;
 
     TestCommWorld->broadcast(dest);
 
-    for (std::size_t i=0; i<src.size(); i++)
+    for (std::size_t i=0; i<3; i++)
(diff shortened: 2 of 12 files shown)


src/solvers/petscdmlibmeshimpl.C
```diff
@@ -907,7 +907,9 @@ static PetscErrorCode  DMSetUp_libMesh(DM dm)
     ierr = DMSNESSetFunction(dm, SNESFunction_DMlibMesh, (void *)dm); CHKERRQ(ierr);
     ierr = DMSNESSetJacobian(dm, SNESJacobian_DMlibMesh, (void *)dm); CHKERRQ(ierr);
     if (dlm->sys->nonlinear_solver->bounds || dlm->sys->nonlinear_solver->bounds_object)
-      ierr = DMSetVariableBounds(dm, DMVariableBounds_libMesh); CHKERRQ(ierr);
+      {
+        ierr = DMSetVariableBounds(dm, DMVariableBounds_libMesh); CHKERRQ(ierr);
```
jwpeterson, line 911: This looks like a real issue that was caught, nice!

tests/parallel/parallel_test.C
```diff
@@ -151,18 +151,16 @@ public:
...
   void testBroadcast()
   {
-    std::vector<unsigned int> src(3), dest(3);
-
-    src[0]=0;
-    src[1]=1;
-    src[2]=2;
+    // Workaround for spurious warning from operator=
+    // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100366
+    std::vector<unsigned int> src{0,1,2}, dest(3,0);
```
jwpeterson, line 156: I'm always amazed by what C++ compiler writers do. Sometimes in a good way, sometimes not so good.

jwpeterson (approved):

src/solvers/petscdmlibmeshimpl.C
jwpeterson: I would also like to cherry-pick this. I've been sitting on the latest 1.7.0 RC for a while, but there haven't been many fixes recently... probably should just go ahead and pull the trigger.

tests/parallel/parallel_test.C
roystgnr: Yeah, I was shocked by this bug and utterly horrified by the optimization-corruption bug, but I must admit that the reason for my shock and horror is that compiler writers have given me extremely high expectations over the years.  If I tried to write a modern C++ compiler I'm sure it would fail to compile half of the valid codes out there and the ones that compiled would run ten times slower, yet the last time I can recall being affected by a serious compiler bug was before gcc 3 came out.

PR #3099, opened 2021-12-09
Mapping fixes, optimizations

roystgnr:
The edge_map fix here seems to be just what I needed to fix initial condition projections on a complex IsoGeometric Analysis mesh, and it should generally be faster and less FP-error prone in other edge reinit uses too.


loganharbour:
Glad to see that one of my method additions to `Elem` finally came in handy for someone other than myself 😆

PR #3100, opened 2021-12-10
Add helper for building element sides that minimizes construction

loganharbour:
The capability to pass in a possibly exiting element in the form of a `unique_ptr` to `Elem::build_side_ptr()` was added in the past. If the element that is passed in is of the correct type required to build a side, no new allocation is made and the necessary information is transferred in place. 

This helper, the `ElemSideBuilder`, stores a cache that is indexed by `ElemType` and "builds" element sides given an element and a side in a way that requires the construction of an element for a given type only once. With this also comes `Elem::side_type()`, which returns the `ElemType` of a given side and is needed for indexing into the cache in `ElemSideBuilder.`

Additionally, I used this new helper in a few places and also cleaned up some side building where appropriate.

### DIFF
--- a/examples/miscellaneous/miscellaneous_ex9/augment_sparsity_on_interface.C
+++ b/examples/miscellaneous/miscellaneous_ex9/augment_sparsity_on_interface.C
@@ -4,6 +4,7 @@
 // libMesh includes
 #include "libmesh/elem.h"
 #include "libmesh/boundary_info.h"
+#include "libmesh/elem_side_builder.h"
 
 using namespace libMesh;
 
@@ -35,20 +36,23 @@ void AugmentSparsityOnInterface::mesh_reinit ()
   std::map<std::pair<const Elem *, unsigned char>, Point> lower;
   std::map<std::pair<const Elem *, unsigned char>, Point> upper;
 
+  // To avoid extraneous allocation when getting element side vertex averages
+  std::unique_ptr<const Elem> side_elem;
+
   for (const auto & elem : _mesh.active_element_ptr_range())
     for (auto side : elem->side_index_range())
       if (elem->neighbor_ptr(side) == nullptr)
         {
           if (_mesh.get_boundary_info().has_boundary_id(elem, side, _crack_boundary_lower))
             {
-              std::unique_ptr<const Elem> side_elem = elem->build_side_ptr(side);
+              elem->build_side_ptr(side_elem, side);
 
               lower[std::make_pair(elem, side)] = side_elem->vertex_average();
             }
 
           if (_mesh.get_boundary_info().has_boundary_id(elem, side, _crack_boundary_upper))
             {
-              std::unique_ptr<const Elem> side_elem = elem->build_side_ptr(side);
+              elem->build_side_ptr(side_elem, side);
 
               upper[std::make_pair(elem, side)] = side_elem->vertex_average();
             }
--- a/include/geom/elem_side_builder.h
+++ b/include/geom/elem_side_builder.h
@@ -1,0 +1,68 @@
+// The libMesh Finite Element Library.
+// Copyright (C) 2002-2021 Benjamin S. Kirk, John W. Peterson, Roy H. Stogner
+
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+
+
+#ifndef LIBMESH_ELEM_SIDE_BUILDER_H
+#define LIBMESH_ELEM_SIDE_BUILDER_H
+
+#include "libmesh/elem.h"
+
+namespace libMesh
+{
+
+/**
+ * Helper for building element sides that minimizes the construction
+ * of new elements.
+ *
+ * The building of element side pointers has the option to pass an
+ * existing constructed element to avoid extraneous allocation.
+ * If the element that is passed is of the correct type, the
+ * build is done in place, that is, the nodes and subdomain are
+ * done in place.
+ *
+ * This tool contains a cache with an entry per element type.
+ * When a side is requested, the cached element of the desired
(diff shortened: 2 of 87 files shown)


include/geom/elem_side_builder.h
```diff
@@ -0,0 +1,49 @@
...
+#include "libmesh/elem.h"
+
+namespace libMesh
+{
+
+/**
+ * Helper for building element sides that minimizes the construction
+ * of new elements.
+ *
+ * TODO: add more info
```
roystgnr, line 32: Indeed.

include/geom/elem_side_builder.h
```diff
@@ -0,0 +1,49 @@
...
+{
+public:
+  /**
+   * Builds an element side for side \p s of element \p elem.
+   */
+  Elem & operator(const Elem & elem, const unsigned int s);
+
+private:
+  /// Proxy element cache for building side pointers; indexed by ElemType
+  std::vector<std::unique_ptr<Elem>> _cached_elems;
```
roystgnr, line 44: This is going to end up quite sparse, isn't it?  I can't think of a better way to do it, though; typically for containers this small the relative overhead of anything but a vector/array is huge.

tests/geom/elem_test.C
```diff
@@ -139,14 +140,37 @@ public:
...
+
+  void test_elem_side_builder()
+  {
+    ElemSideBuilder cache;
+    for (const auto & elem : _mesh->active_local_element_ptr_range())
+      for (const auto s : elem->side_index_range())
+      {
+        const auto side = elem->build_side_ptr(s);
+        const auto & cached_side = cache(*elem, s);
+        CPPUNIT_ASSERT_EQUAL(e1.type(), e2.type());
```
roystgnr, line 159: I'm guessing this'll be fixed by the time I'm done typing this comment...

include/geom/elem_side_builder.h
loganharbour: Yep. Only other option here is if we want to go static with an array; same sparsity issue though

loganharbour:
Barring any failures on my part, this should be good to go. Take a look at the doctring for `ElemSideBuilder` and let me know what you think.

roystgnr:
Docs look good to me.  Let's wait for @jwpeterson to check it out before merging, though.  I've got a few other things (both the outstanding PRs and one other branch still in progress) that I'm hoping to get into the next Moose libMesh submodule update, myself, so I don't want to rush that too too much.

loganharbour:
Are there tests for inf elements anywhere?

loganharbour:
Few more things, of which objections to any are okay with me:

- Applied `ElemSideBuilder` to the examples that build sides; this may be overkill and it may be more appropriate to just use a single `std::unique_ptr<const Elem>`, but I figured it also wasn't an idea to expose users to the class in the examples
- Applied `ElemSideBuilder` to a few methods in libmesh that _could_ be building element sides for multiple types
- Used a single element in a few places for building sides where it was appropriate

jwpeterson:
> Let's wait for @jwpeterson to check it out before merging, 

After @roystgnr explained the issue to me in terms of Prism elements, where you are switching back and forth between different geometric side types, the idea makes sense to me. So, assuming that the tests start passing, I'd be OK with merging? Would you mind adding a description to the PR as well?

src/mesh/mesh_tools.C
```diff
@@ -366,9 +366,7 @@ void MeshTools::find_boundary_nodes (const MeshBase & mesh,
     for (auto s : elem->side_index_range())
       if (elem->neighbor_ptr(s) == nullptr) // on the boundary
         {
-          std::unique_ptr<const Elem> side = elem->build_side_ptr(s);
```
roystgnr, line 369: What?  How did we never get an unused variable warning for this?  Is the non-trivial destructor enough for the compiler to have to give up and say "maybe they're doing some fancy RAII thing"?

loganharbour: Right?! I had the same thought!

loganharbour:
> So, assuming that the tests start passing, I'd be OK with merging? Would you mind adding a description to the PR as well?

Yep - assuming this is passing, this is good to go. It does require a once-over from one of you to confirm that the places that I added `ElemSideBuilder` are appropriate.

In addition to @roystgnr - I did make a change to `ElemSideBuilder` that allows for building const and non-const sides that should be looked at (uses a little const casting)

examples/miscellaneous/miscellaneous_ex9/augment_sparsity_on_interface.C
```diff
@@ -35,23 +36,26 @@ void AugmentSparsityOnInterface::mesh_reinit ()
...
 
+  // To avoid extraneous allocation when getting element side vertex averages
+  std::unique_ptr<const Elem> side_elem;
+
   for (const auto & elem : _mesh.active_element_ptr_range())
     for (auto side : elem->side_index_range())
       if (elem->neighbor_ptr(side) == nullptr)
         {
           if (_mesh.get_boundary_info().has_boundary_id(elem, side, _crack_boundary_lower))
-            {
```
loganharbour, line 43: Gah... stupid spaces

loganharbour:
This isn't completely ready. I'm going to add infinite elem tests to `elem_test.C`

loganharbour:
Good to go.

src/geom/cell_inf_hex8.C
```diff
@@ -67,7 +67,7 @@ const unsigned int InfHex8::edge_nodes_map[InfHex8::num_edges][InfHex8::nodes_pe
 const unsigned int InfHex8::edge_sides_map[InfHex8::num_edges][2] =
   {
     {0, 1}, // Edge 0
-    {1, 2}, // Edge 1
+    {0, 2}, // Edge 1
```
roystgnr, line 70: Come again?  I could imagine having screwed up this table somehow (or having copied a mistake in the diagram in cell_inf_hex8.h?), but not like this.  An InfHex vertex has three edges connected to it.  There's no way node 0 is part of 4 edges.

loganharbour: Sound logic. This is actually a map that I implemented :-) 

I'll look into it

roystgnr:
That `edge_sides_map` weirdness was the only problem I could find.

src/geom/cell_inf_hex8.C
loganharbour: Oh - this is edge -> side map. Has nothing to do with nodes. It's valid.

loganharbour: Actually, hold up. Need to correct it in the other classes, too.

loganharbour:
Done, again. Consolidated the edge_sides_map for the cell classes, moving the map to the base classes instead. I duplicated the same map for all of the specialized cells when it was unnecessary.

roystgnr (approved):

roystgnr:
@jwpeterson - looks like 2434108 is necessary + sufficient to backport onto the release branch

PR #3106, opened 2021-12-14
MeshTools::clear_spline_nodes() support

roystgnr:
It's becoming clear to me that we need to support condensed linear systems, or *some* sort of better solver support for systems with tons of constraint equations ... but in the meantime we can sometimes work around the problem by avoiding creating tons of constraint equations.  Factoring that support out of the DynaIO IGA code lets us invoke it for Exodus IGA files too.


roystgnr:
Going to merge this right away - we have coverage here and previously for the DynaIO effects, and that's slightly deprecated in favor of Exodus IGA now anyway.

PR #3107, opened 2021-12-14
Fix all_second_order() bug with extra integers

### LINKED ISSUE #3105: transfer_elem seems not transferring extra element integers
The `transfer_elem` in `unstructured_mesh.C` used by `UnstructuredMesh::all_second_order` seems not transferring extra element integers. Because of this, when converting a first order mesh with extra element integers to second order, we will lose these integers. This is a bug reported by @yjung-anl. Possibly care needs to be taken for nodes (not sure). Tag @roystgnr

roystgnr:
This should fix #3105


roystgnr:
Same bug exists on the 1.7.0 branch, but pre-refactoring so the patch here won't apply cleanly.  I'll fix it by hand there.

YaqiWang:
Thanks @roystgnr.

PR #3108, opened 2021-12-15
Fix bug introduced in #3100

roystgnr:
We should probably be running a CI recipe with `--enable-everything --disable-strict-lgpl` ...


loganharbour (approved):
Oof. That's a significant error

PR #3112, opened 2022-01-03
Don't unnecessarily compute the Jacobian for ComputeResidualandJacobian

lindsayad:
In our `PetscNonlinearSolver` we can compute the Jacobian during
residual evaluation if the user has supplied a
`ComputeResidualandJacobian` object. Then, we can simply return during
the PETSc Jacobian evaluation call and avoid duplicating work. I tested this
on example 8 from `systems_of_equations` and the result is unchanged.
This work is necessary in order for adoption of `ComputeResidualandJacobian`
by MOOSE AD to yield gains.


jwpeterson (approved):
Looks good to me, and sounds like it's working for you as well... go ahead and merge when ready.

PR #3114, opened 2022-01-04
Higher precision when printing assert failures

roystgnr:
I managed to trigger a bug where doubles that I expected to be exactly
equal really did differ by an ulp, but increasing the precision here
was what I needed to ensure they weren't actually differing by a billion
ulps.

### DIFF
--- a/include/base/libmesh_common.h
+++ b/include/base/libmesh_common.h
@@ -301,14 +301,18 @@ extern bool warned_about_auto_ptr;
 #define libmesh_assert_equal_to_msg(expr1,expr2, msg)                   \
   do {                                                                  \
     if (!((expr1) == (expr2))) {                                        \
+      std::streamsize oldp = libMesh::err.precision(17);                \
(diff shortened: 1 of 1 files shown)


jwpeterson:
Looks like [std::streamsize is defined in the `<ios>` header](https://en.cppreference.com/w/cpp/io/streamsize) which we don't currently explicitly directly include in libmesh_common.h, though it's possible we get it indirectly. Might be worth adding just for completeness/in case the indirect inclusions change at some point.

roystgnr:
Hmmm... it should be guaranteed to be included via `iostream`, right?  It's being passed and returned by value in APIs there, so there's no way they can forward declare it.

jwpeterson:
Ah, yes, `<iostream>` includes `<ios>` but we don't directly include `<iostream>` in libmesh_common.h either, most likely relying on getting it from ostream_proxy.h.

roystgnr:
True.  Both steps of that inclusion ought to be guaranteed here too, though, no?  We're directly using `libMesh::err` so we won't be removing `ostream_proxy.h`, and `ostream_proxy.h` is supposed to be an inlineable shim so we won't be removing `iostream` from the includes there.

PR #3119, opened 2022-01-07
resolve warning of expectedly unbinned entries.

BalticPinguin:
To start with, I wish you all a happy new year !
it is a small (in my viewpoint) improvement of the `StatisticsVector::histogram`-function, which I use in my program and always got a warning that some entries are not accounted for.
Since this is (at least in my case always) intentional, I would opt for removing that warning, keeping it for cases where something really went wrong.

But maybe there are reasons to inform the user in any case?

### DIFF
--- a/src/utils/statistics.C
+++ b/src/utils/statistics.C
@@ -206,6 +206,13 @@ void StatisticsVector<T>::histogram(std::vector<dof_id_type> & bin_members,
   // This vector will store the number of members each bin has.
   bin_members.resize(n_bins);
 
+#ifdef DEBUG
+  // we may not bin all values.
+  // Those we skip on purpose (e.g. inactive elements in an ErrorVector)
+  // should also not appear in the consistency-check below.
+  unsigned int unbinned=0;
+#endif
+
   dof_id_type data_index = 0;
   for (auto j : index_range(bin_members)) // bin vector indexing
     {
@@ -221,6 +228,9 @@ void StatisticsVector<T>::histogram(std::vector<dof_id_type> & bin_members,
           // ErrorVector.)  We just skip entries like that.
           if (current_val < min)
             {
+#ifdef DEBUG
+               unbinned++;
+#endif
               //     libMesh::out << "(debug) Skipping entry v[" << i << "]="
               //       << (*this)[i]
               //       << " which is less than the min value: min="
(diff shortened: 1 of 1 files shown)


src/utils/statistics.C
```diff
@@ -253,12 +267,12 @@ void StatisticsVector<T>::histogram(std::vector<dof_id_type> & bin_members,
                                                static_cast<dof_id_type>(0),
                                                std::plus<dof_id_type>());
 
-  if (n != n_binned)
+  if (n-unbinned != n_binned)
     {
       libMesh::out << "Warning: The number of binned entries, n_binned="
                    << n_binned
                    << ", did not match the total number of entries, n="
```
jwpeterson, line 274: It might be more accurate for this warning to say "total number of *binnable* entries", since now we don't count the inactive ones.

src/utils/statistics.C
```diff
@@ -206,6 +206,13 @@ void StatisticsVector<T>::histogram(std::vector<dof_id_type> & bin_members,
   // This vector will store the number of members each bin has.
   bin_members.resize(n_bins);
 
+#ifdef DEBUG
+  // we may not count all bins.
```
jwpeterson, line 210: I think maybe this comment should read "We may not bin all values"?

jwpeterson (commented):
Your change seems fine to me, although I have couple of comments.

1. It's not clear to me how `current_val < min` happens? `min` == the minimum value in the entire vector, so by definition nothing should be less than that.
1. I think we also don't bin values which are larger than the  max, i.e. `current_val > bin_bounds[j+1]` and we are on the last bin? Should these also count as unbinned? (Again, it's not clear to me how there could be any values larger than the max, so maybe this is not an issue.

BalticPinguin:
thanks @jwpeterson  for the comments; I have updated the points and amended the commit since it is only a small one.

> how current_val < min happens?

This happens on `ErrorVectors`: Its overridden `min()`-function  only accounts for active elements (inactive elements get `0`-entries but shouldn't be counted).

> current_val > bin_bounds[j+1] and we are on the last bin?

I don't think, this situation could happen. That is how I would understand the warning: There are exactly `num_inactive_elem` reasons to be not within the bounds.

But I agree that, with the changes suggested here, the check for the number of binned entries ends up being rather pointless unless the `sort`-function doesn't work. I think, this is what you were after?

jwpeterson:
> Its overridden min()-function only accounts for active elements

Ah, of course. Thanks I kind of forgot about how this class hierarchy worked.

> the check for the number of binned entries ends up being rather pointless unless the sort-function doesn't work.

I think it's always been kind of a sanity check that we never expect to fail (though we did not consider the inactive entries before, oops) so your changes look fine to me. Thanks!

PR #3122, opened 2022-01-08
Nanoflann update to 1.4.1

### LINKED ISSUE #3121: Update nanoflann contrib
The current nanoflann version is a few years old and development has not stopped. For a MOOSE related work I ran into an exception for a missing feature in nanoflann that has since been added (related to dynamic trees).

dschwen:
Closes #3121


dschwen:
Moose fails are expected, I have a psych ready.

lindsayad:
Is there any expectation that nanoflann might restore the ability to be backwards compatible given your comment on jlblancoc/nanoflann#154 or should we just move forward?

dschwen:
No idea, and tbh I don't understand why this is incompatible yet. At first glance it looked like they were expecting a pointer to a std:: vector::size_type, which would be std::size_t

I'll look some more into it. There is also a compiler warning I submitted a PR for.

dschwen:
And that PR was already merged https://github.com/jlblancoc/nanoflann/pull/163

I'll update this PR accordingly.

dschwen:
@lindsayad the change from `uint64_t` to `uint3_t` for the default index ("accessor") type was made [here](https://github.com/jlblancoc/nanoflann/commit/1bac65cf05c6235a12d14fe4a91cebeeb7486cf8). 

One might reasonably argue that it will be rather unusual (for now) to have trees exceeding 4.29 billion leaves, and if such an application is anticipated it would be reasonable to explicitly request the broader index type. 

I'll check, if member function templating could be used to convert indices for backward compatibility.

lindsayad:
> explicitly request the broader index type.

Is this request available through configure options?

dschwen:
> > explicitly request the broader index type.
> 
> Is this request available through configure options?

No, through an additional template parameter.

dschwen:
Roy, this might be possible. Let me check if the additional index type template parameter is available in the old nanoflann version... 

...and yes, this is possible, so I'll go ahead and make a quick PR for that!

dschwen:
idaholab/moose#20000

jwpeterson:
Should we (in a separate PR) switch nanoflann over to a git submodule? They actually do updates and have git, so it seems like a good candidate.

lindsayad:
image

roystgnr:
It would be a great candidate, if the build systems aren't too incompatible.  I don't know how much editing we've done in our fork ... looks like 3 commits after the last update?  That doesn't sound too bad, especially if all but one were bugfixes that could go in upstream.  Did you see the stupid-games-with-sed I played in the poly2tri submodule?  That wasn't awful but I wouldn't want to do anything much crazier.

jwpeterson:
@lindsayad we do have submodules in our submodules... the autoconf one IIRC, but you were the one that added it!

jwpeterson:
> if the build systems aren't too incompatible.

It's header-only, right?

lindsayad:
Haha indeed. I just wanted to post a meme. I'm a big fan of submodules once you get past the first couple weeks where CIVET fails for confusing reasons

jlblancoc:
Hi all, 
Sorry for the late show up. 
I think you should be able to maintain backwards compatibility by setting the last template parameter to `size_t`... did you already tried it?
Let me know if it doesn't work and need any further change upstream, before we release nanoflann 1.4.2.

dschwen:
Woot! Hey @jlblancoc , I'm doing that in idaholab/moose#20000. I think this should be good enough!

jlblancoc:
Awesome!

dschwen:
is it maybe called `vAcc` now? https://github.com/jlblancoc/nanoflann/blob/fa97cabdc39b88e6df4b7a4ad00a8c965ff4a0da/include/nanoflann.hpp#L1488-L1498

jwpeterson:
The type is different
 ```
    std::vector<AccessorType> vAcc;
```
vs.
```
    std::vector<IndexType> vind;
```
but I see that `AccessorType` is actually an index...     I will try that I guess.

**Comment by jwpeterson** [comment]
Yeah our code compiles simply using `vAcc` instead of `vind` (haven't tested it yet). It would be good to know the specific `NANOFLANN_VERSION` where `vAcc` was introduced so that we could properly guard the code with #ifdefs.

**Comment by jwpeterson** [comment]
I found commit jlblancoc/nanoflann@d0e7852d2a92e4a17baf4f91c634270e31bea465 where `vind` was replaced by `vAcc`. In that hash, I see:
```
#define NANOFLANN_VERSION 0x132
```
so I will use this as an approximate version guard.

PR #3124, opened 2022-01-14
Check processor id of point neighbors

lindsayad:
I think this was probably a simple oversight


lindsayad:
I'm actually curious if there are any nefarious behaviors that could arise from adding a local element?

roystgnr:
I think this was just a performance bug, not a correctness issue.  Pretty embarrassing one, though.

roystgnr (approved):

lindsayad:
Performance due to the cost of an `unordered_map` insertion? Or is there something more?

roystgnr:
Performance due to unnecessary communication when repartitioning, mainly.  With those extra entries in there I'd expect each processor to be sent redundant copies of many of its newly acquired elements.

PR #3128, opened 2022-01-16
Make use of std::condition_variable for _get_array

### LINKED ISSUE #3127: Helgrind error out of PetscVector::_get_array
My understanding of `std::atomic_thread_fence` is basically 0, but it doesn't seem to be preventing these valgrind errors like a lock would. Maybe valgrind doesn't understand `std::atomic_thread_fence`?
```
==2236593== Possible data race during read of size 1 at 0x103C5200 by thread #79
==2236593== Locks held: none
==2236593==    at 0x809552D: load (atomic_base.h:481)
==2236593==    by 0x809552D: operator bool (atomic:87)
==2236593==    by 0x809552D: libMesh::PetscVector<double>::_get_array(bool) const (petsc_vector.C:1164)
==2236593==    by 0x49AA91A: libMesh::PetscVector<double>::get(std::vector<unsigned long, std::allocator<unsigned long> > const&, double*) const (petsc_vector.h:1088)
==2236593==    by 0x830413A: get (numeric_vector.h:856)
==2236593==    by 0x830413A: get (numeric_vector.h:848)
==2236593==    by 0x830413A: libMesh::OldSolutionValue<double, &(void libMesh::FEMContext::point_value<double>(unsigned int, libMesh::Point const&, double&, double) const)>::eval_old_dofs(libMesh::Elem const&, unsigned int, unsigned int, std::vector<unsigned long, std::allocator<unsigned long> >&, std::vector<double, std::allocator<double> >&) (generic_projector.h:858)
==2236593==    by 0x8326191: libMesh::GenericProjector<libMesh::OldSolutionValue<double, &(void libMesh::FEMContext::point_value<double>(unsigned int, libMesh::Point const&, double&, double) const)>, libMesh::OldSolutionValue<libMesh::VectorValue<double>, &(void libMesh::FEMContext::point_gradient<libMesh::VectorValue<double> >(unsigned int, libMesh::Point const&, double&, double) const)>, double, libMesh::VectorSetAction<double> >::SortAndCopy::operator()(libMesh::StoredRange<libMesh::MeshBase::const_element_iterator, libMesh::Elem const*> const&) (generic_projector.h:1742)
==2236593==    by 0x8328336: void* libMesh::Threads::run_body<libMesh::StoredRange<libMesh::MeshBase::const_element_iterator, libMesh::Elem const*>, libMesh::GenericProjector<libMesh::OldSolutionValue<double, &(void libMesh::FEMContext::point_value<double>(unsigned int, libMesh::Point const&, double&, double) const)>, libMesh::OldSolutionValue<libMesh::VectorValue<double>, &(void libMesh::FEMContext::point_gradient<libMesh::VectorValue<double> >(unsigned int, libMesh::Point const&, double&, double) const)>, double, libMesh::VectorSetAction<double> >::SortAndCopy>(void*) (threads_pthread.h:236)
==2236593==    by 0x484E8AA: ??? (in /usr/libexec/valgrind/vgpreload_helgrind-amd64-linux.so)
==2236593==    by 0xE17D926: start_thread (pthread_create.c:435)
==2236593==    by 0xE20D9E3: clone (clone.S:100)
==2236593== 
==2236593== This conflicts with a previous write of size 1 by thread #78
==2236593== Locks held: 1, at address 0x103C5238
==2236593==    at 0x80956A7: store (atomic_base.h:457)
==2236593==    by 0x80956A7: store (atomic:104)
==2236593==    by 0x80956A7: libMesh::PetscVector<double>::_get_array(bool) const (petsc_vector.C:1227)
==2236593==    by 0x49AA91A: libMesh::PetscVector<double>::get(std::vector<unsigned long, std::allocator<unsigned long> > const&, double*) const (petsc_vector.h:1088)
==2236593==    by 0x830413A: get (numeric_vector.h:856)
==2236593==    by 0x830413A: get (numeric_vector.h:848)
==2236593==    by 0x830413A: libMesh::OldSolutionValue<double, &(void libMesh::FEMContext::point_value<double>(unsigned int, libMesh::Point const&, double&, double) const)>::eval_old_dofs(libMesh::Elem const&, unsigned int, unsigned int, std::vector<unsigned long, std::allocator<unsigned long> >&, std::vector<double, std::allocator<double> >&) (generic_projector.h:858)
==2236593==    by 0x8326191: libMesh::GenericProjector<libMesh::OldSolutionValue<double, &(void libMesh::FEMContext::point_value<double>(unsigned int, libMesh::Point const&, double&, double) const)>, libMesh::OldSolutionValue<libMesh::VectorValue<double>, &(void libMesh::FEMContext::point_gradient<libMesh::VectorValue<double> >(unsigned int, libMesh::Point const&, double&, double) const)>, double, libMesh::VectorSetAction<double> >::SortAndCopy::operator()(libMesh::StoredRange<libMesh::MeshBase::const_element_iterator, libMesh::Elem const*> const&) (generic_projector.h:1742)
==2236593==    by 0x8328336: void* libMesh::Threads::run_body<libMesh::StoredRange<libMesh::MeshBase::const_element_iterator, libMesh::Elem const*>, libMesh::GenericProjector<libMesh::OldSolutionValue<double, &(void libMesh::FEMContext::point_value<double>(unsigned int, libMesh::Point const&, double&, double) const)>, libMesh::OldSolutionValue<libMesh::VectorValue<double>, &(void libMesh::FEMContext::point_gradient<libMesh::VectorValue<double> >(unsigned int, libMesh::Point const&, double&, double) const)>, double, libMesh::VectorSetAction<double> >::SortAndCopy>(void*) (threads_pthread.h:236)
==2236593==    by 0x484E8AA: ??? (in /usr/libexec/valgrind/vgpreload_helgrind-amd64-linux.so)
==2236593==    by 0xE17D926: start_thread (pthread_create.c:435)
==2236593==    by 0xE20D9E3: clone (clone.S:100)
```

lindsayad:
I'll keep the below text for historical purposes. TLDR, initially I threw up a PR
that just locked at the top of `_get_array` in order to avoid helgrind errors. But
then I dove into this a bit more, and I think the new code keeps optimal performance
and eliminates helgrind errors.

### Original post

This removed all the helgrind errors coming out of the threaded element
loop in `system_projection.C`. By putting the lock at function scope as
opposed to contained within a `if (!_array_present)` scope, you ensure
that you don't let through any "reader" threads (e.g. threads that exit
this function and go onto read things like the global to local map and
finally the actual read array), while you have a writer thread. This
also makes this function much simpler to read I think. You will have
more lock contention with this change for sure (like a lot more), but
once threads get through the handful of lines under contention, they're
free to race off (reading) uncontested.

I recognize that letting reader threads through and the helgrind errors
associated with this are probably false positives, at least I'm 90% sure
they are false positives 100% of the time (because the only way that
`_array_is_present` can be `true`, which "lets a reader thread through",
is if a writer thread has already essentially made
it's way through `_get_array` ... unless two threads had two different
call stacks ... and if that was the case then the
function level lock in `_get_array` would also be insufficient to
prevent races). So if we don't want to go with this PR, then I
definitely understand. I just get tickled the wrong way by these
helgrind "errors". Maybe if we don't let this get merged, I can do some
helgrind development instead 😆 

Closes #3127

### DIFF
--- a/include/numerics/petsc_vector.h
+++ b/include/numerics/petsc_vector.h
@@ -1099,6 +1099,9 @@ void PetscVector<T>::get(const std::vector<numeric_index_type> & index,
 
   std::unique_lock<std::mutex> read_lock(_petsc_vector_mutex);
   _petsc_vector_cv.wait(read_lock, [this](){ return _array_is_present.load(); });
+  // When wait exits it means we've acquired and locked the mutex, but all we are doing now
+  // is reading, so it's safe to unlock and let other threads continue
+  read_lock.unlock();
   for (std::size_t i=0; i<num; i++)
     {
       const numeric_index_type local_index = this->map_global_to_local_index(index[i]);
--- a/include/numerics/petsc_vector.h
+++ b/include/numerics/petsc_vector.h
@@ -51,10 +51,9 @@
 #include <unordered_map>
 #include <limits>
 
-#ifdef LIBMESH_HAVE_CXX11_THREAD
 #include <atomic>
 #include <mutex>
-#endif
+#include <condition_variable>
 
 namespace libMesh
 {
@@ -414,11 +413,20 @@ private:
    * object to keep down thread contention when reading frmo multiple
    * PetscVectors simultaneously
    */
-#ifdef LIBMESH_HAVE_CXX11_THREAD
   mutable std::mutex _petsc_vector_mutex;
-#else
-  mutable Threads::spin_mutex _petsc_vector_mutex;
-#endif
+
+  /**
+   * Condition variable for _get_array and _restore_array.  This is part of the
+   * object to keep down thread contention when reading frmo multiple
+   * PetscVectors simultaneously
+   */
+  mutable std::condition_variable _petsc_vector_cv;
+
+  /**
+   * Mutex that helps enforce a call_once idiom within a method, where
+   * once means call for only one thread
+   */
+  mutable std::mutex _petsc_vector_do_once_mutex;
 
   /**
    * Queries the array (and the local form if the vector is ghosted)
@@ -460,7 +468,7 @@ private:
   /**
    * Whether or not the data array is for read only access
    */
-  mutable bool _values_read_only;
+  mutable std::atomic<bool> _values_read_only;
 };
 
 
@@ -1089,6 +1097,8 @@ void PetscVector<T>::get(const std::vector<numeric_index_type> & index,
 
   const std::size_t num = index.size();
 
+  std::unique_lock<std::mutex> read_lock(_petsc_vector_mutex);
+  _petsc_vector_cv.wait(read_lock, [this](){ return _array_is_present.load(); });
   for (std::size_t i=0; i<num; i++)
     {
       const numeric_index_type local_index = this->map_global_to_local_index(index[i]);
--- a/include/numerics/petsc_vector.h
+++ b/include/numerics/petsc_vector.h
@@ -410,14 +410,14 @@ private:
 
   /**
    * Mutex for _get_array and _restore_array.  This is part of the
-   * object to keep down thread contention when reading frmo multiple
+   * object to keep down thread contention when reading from multiple
    * PetscVectors simultaneously
    */
   mutable std::mutex _petsc_vector_mutex;
 
   /**
    * Condition variable for _get_array and _restore_array.  This is part of the
-   * object to keep down thread contention when reading frmo multiple
+   * object to keep down thread contention when reading from multiple
    * PetscVectors simultaneously
    */
   mutable std::condition_variable _petsc_vector_cv;
--- a/src/numerics/petsc_vector.C
+++ b/src/numerics/petsc_vector.C
@@ -1151,12 +1151,6 @@ void PetscVector<T>::_get_array(bool read_only) const
 {
   libmesh_assert (this->initialized());
 
-#ifdef LIBMESH_HAVE_CXX11_THREAD
-  std::atomic_thread_fence(std::memory_order_acquire);
-#else
-  Threads::spin_mutex::scoped_lock lock(_petsc_vector_mutex);
-#endif
-
   // If the values have already been retrieved and we're currently
   // trying to get a non-read only view (ie read/write) and the
   // values are currently read only... then we need to restore
@@ -1171,11 +1165,10 @@ void PetscVector<T>::_get_array(bool read_only) const
 
   if (!_array_is_present)
     {
-#ifdef LIBMESH_HAVE_CXX11_THREAD
-      std::lock_guard<std::mutex> lock(_petsc_vector_mutex);
-#endif
+      std::lock_guard<std::mutex> do_once_lock(_petsc_vector_do_once_mutex);
       if (!_array_is_present)
         {
+          std::unique_lock<std::mutex> write_lock(_petsc_vector_mutex);
           PetscErrorCode ierr=0;
           if (this->type() != GHOSTED)
             {
@@ -1222,12 +1215,8 @@ void PetscVector<T>::_get_array(bool read_only) const
             _first = static_cast<numeric_index_type>(petsc_first);
             _last = static_cast<numeric_index_type>(petsc_last);
           }
-#ifdef LIBMESH_HAVE_CXX11_THREAD
-          std::atomic_thread_fence(std::memory_order_release);
-          _array_is_present.store(true, std::memory_order_relaxed);
-#else
           _array_is_present = true;
-#endif
+          _petsc_vector_cv.notify_all();
         }
     }
 }
--- a/src/numerics/petsc_vector.C
+++ b/src/numerics/petsc_vector.C
@@ -1168,7 +1168,6 @@ void PetscVector<T>::_get_array(bool read_only) const
       std::lock_guard<std::mutex> do_once_lock(_petsc_vector_do_once_mutex);
       if (!_array_is_present)
         {
-          std::unique_lock<std::mutex> write_lock(_petsc_vector_mutex);
           PetscErrorCode ierr=0;
           if (this->type() != GHOSTED)
             {


lindsayad:
This is a simple program which has the same concept as current `master` and it yields the same helgrind errors. I feel like there has to be a way to do this, with all the c++11 and later advents in thread technology, to not get errors out of helgrind.
```c++
#include <thread>
#include <vector>
#include <atomic>
#include <mutex>

std::mutex cv_m;
std::atomic<bool> done = false;
std::vector<int> container;

void fillContainer()
{
  if (!done)
  {
    std::lock_guard<std::mutex> lock(cv_m);
    if (!done)
    {
      container = {1, 2, 3, 4};
      done = true;
    }
  }

  const auto a = container.front();
}


int main()
{
  std::thread t1(fillContainer), t2(fillContainer);
  t1.join();
  t2.join();
}
```

lindsayad:
This looks like it does the trick
```c++
#include <thread>
#include <vector>
#include <atomic>
#include <mutex>
#include <iostream>
#include <condition_variable>

int main()
{
  std::vector<int> container;
  std::mutex mx;
  std::mutex do_once_mx;
  std::condition_variable cv;
  std::atomic<bool> container_filled(false);

  auto get = [&container, &cv, &mx, &do_once_mx, &container_filled]()
    {
      {
        // _getArray                                                                                                  
        if (!container_filled)
        {
          std::lock_guard<std::mutex> do_once(do_once_mx);
          if (!container_filled)
          {
            std::unique_lock<std::mutex> lock(mx);
            container = {1, 2, 3, 4};
            container_filled = true;
            cv.notify_all();
          }
        }
      }

      std::unique_lock<std::mutex> lock(mx);
      cv.wait(lock, [&container_filled](){ return container_filled.load(); });
      std::cout << container.front() << std::endl;
    };

  std::thread t1(get), t2(get);
  t1.join();
  t2.join();
}
```

lindsayad:
This looks even better
```c++
#include <thread>
#include <vector>
#include <atomic>
#include <mutex>
#include <iostream>
#include <condition_variable>

int main()
{
  std::vector<int> container;
  std::mutex mx;
  std::condition_variable cv;
  std::atomic<bool> container_filled(false);
  std::once_flag flag;

  auto get = [&container, &cv, &mx, &container_filled, &flag]()
    {
      std::call_once(flag, [&container_filled, &mx, &container, &cv]()
      {
        std::unique_lock<std::mutex> lock(mx);
        container = {1, 2, 3, 4};
        container_filled = true;
        cv.notify_all();
      });

      std::unique_lock<std::mutex> lock(mx);
      cv.wait(lock, [&container_filled](){ return container_filled.load(); });
      std::cout << container.front() << std::endl;
    };

  std::thread t1(get), t2(get);
  t1.join();
  t2.join();
}
```

roystgnr:
Okay, I've got to scream for another set of eyes that understands race conditions better than I do, to make sure this is safe.  Maybe @friedmud has time to review?

lindsayad:
Ok I tried a test where I launched the functions from https://github.com/libMesh/libmesh/pull/3128#issuecomment-1014056722 and https://github.com/libMesh/libmesh/pull/3128#issuecomment-1014057118 (latter slightly modified to now match the current state of this PR and to remove cout because after unlocking it would be unsafe to call cout because it's read-write) 100,000 times with 8 threads and timings were 24.095 seconds and 24.151 seconds respectively. So I would say they are roughly equivalent with the latter also having the benefit of reporting helgrind clean

lindsayad:
Largely for my own memory, I'm posting [this stackoverflow answer](https://stackoverflow.com/a/39038214/4493669) which helped me go down this path

roystgnr:
I think I'm happy with the code here now.  I want to run some tests with it, but unless John or Derek chime in before I'm done I'll merge afterward.

include/numerics/petsc_vector.h
```diff
@@ -414,11 +413,20 @@ class PetscVector final : public NumericVector<T>
...
    */
-#ifdef LIBMESH_HAVE_CXX11_THREAD
   mutable std::mutex _petsc_vector_mutex;
-#else
-  mutable Threads::spin_mutex _petsc_vector_mutex;
-#endif
+
+  /**
+   * Condition variable for _get_array and _restore_array.  This is part of the
+   * object to keep down thread contention when reading frmo multiple
```
jwpeterson, line 420: typo "frmo"

src/numerics/petsc_vector.C
```diff
@@ -1151,12 +1151,6 @@ void PetscVector<T>::_get_array(bool read_only) const
 {
   libmesh_assert (this->initialized());
 
-#ifdef LIBMESH_HAVE_CXX11_THREAD
-  std::atomic_thread_fence(std::memory_order_acquire);
-#else
-  Threads::spin_mutex::scoped_lock lock(_petsc_vector_mutex);
```
jwpeterson, line 1157: This is sort of separate from this PR, but are we still maintaining code for different thread "models" e.g. TBB and pthreads in the `Threads` namespace? That seems like an increasingly difficult proposition as the thread support in `std::` gets more sophisticated.

jwpeterson (approved):
I don't know enough about `std::condition_variable`'s workings to comment intelligently, but I do trust that @lindsayad has tested this with helgrind much more than I have (I have tested 0 times) so if he says it works, I'm happy.

src/numerics/petsc_vector.C
roystgnr: It does seem to finally be redundant at this point, doesn't it?

roystgnr (approved):

friedmud:
Wait - this shouldn't have been necessary.  The memory fencing was correct in order to ensure that the doubly checked lock worked properly.  See https://mortoray.com/2012/02/28/double-checked-locking-why-doesnt-it-work-and-how-to-fix-it/

A condition variable and notifications are WAY overkill for this and could seriously slow down threaded execution of residual evaluations.

friedmud:
Ah - I found the original article I used when I wrote this code: https://preshing.com/20130930/double-checked-locking-is-fixed-in-cpp11/

We should look closely and make sure I didn't mess something up and then revert this.  This was my first time using memory fences and I may not have gotten it right.

I use the same kind of memory fencing in `PerfGraphLivePrint` now and we haven't had any helgrind issues there...

friedmud:
I definitely messed this up originally - working on a fix.

lindsayad:
> I use the same kind of memory fencing in `PerfGraphLivePrint` now and we haven't had any helgrind issues there...

I always have to `--disable-perf-graph-live-print` before running helgrind otherwise I'm bombarded with errors

lindsayad:
Originally reported here: https://github.com/idaholab/moose/discussions/19443#discussioncomment-1768107

Maybe this was overkill but it didn't show up in timings [here](https://github.com/libMesh/libmesh/pull/3128#issuecomment-1015725605). However, that was a simple program, and so would not be a perfect model for MOOSE.

friedmud:
@lindsayad thanks for the reminder about perfgraphliveprint.  I'll take a look at that one next.

I have a fix upcoming here - what's the easiest way to test with helgrind (I'm already on Rod)

lindsayad:
You can trigger helgrind errors wit the pre-this-PR code with any test that ends up calling `System::project_vector` so any test that has adaptivity will do it. I just checked that this reproduces errors:
```
(base) lindad@pop-os:~/projects3/moose/test/tests/adaptivity/initial_adapt((HEAD detached at 6b91c65f13))$ valgrind --tool=helgrind ../../../moose_test-devel -i initial_adapt.i --n-threads=2 --disable-perf-graph-live
```

lindsayad:
That being said, if we want to revert this PR, I just don't think helgrind is going to be a tool we can use and not get false positives. For instance, this code below based on [the Preshing example](https://preshing.com/20130930/double-checked-locking-is-fixed-in-cpp11/) yields helgrind errors but passes if you use clang's `ThreadSanitizer`

```c++
#include <thread>
#include <atomic>
#include <mutex>

class Singleton
{
public:
  Singleton() = default;

  static Singleton * getInstance();

private:
  static std::atomic<Singleton *> m_instance;
  static std::mutex m_mutex;
};

std::atomic<Singleton *> Singleton::m_instance;
std::mutex Singleton::m_mutex;

Singleton *
Singleton::getInstance()
{
  Singleton * tmp = m_instance.load(std::memory_order_relaxed);
  std::atomic_thread_fence(std::memory_order_acquire);
  if (tmp == nullptr)
  {
    std::lock_guard<std::mutex> lock(m_mutex);
    tmp = m_instance.load(std::memory_order_relaxed);
    if (tmp == nullptr)
    {
      tmp = new Singleton;
      std::atomic_thread_fence(std::memory_order_release);
      m_instance.store(tmp, std::memory_order_relaxed);
    }
  }
  return tmp;
}

int
main()
{
  std::thread t1(Singleton::getInstance), t2(Singleton::getInstance), t3(Singleton::getInstance),
      t4(Singleton::getInstance), t5(Singleton::getInstance), t6(Singleton::getInstance),
      t7(Singleton::getInstance), t8(Singleton::getInstance);
  t1.join();
  t2.join();
  t3.join();
  t4.join();
  t5.join();
  t6.join();
  t7.join();
  t8.join();
}
```

lindsayad:
Well actually, Preshing's sequentially consistent atomic pattern registers as both `ThreadSanitizer` and `helgrind` clean, and certainly is beautiful and concise code. But as he says the machine code tends to be less efficient that his other examples
```c++
void
fillContainer()
{
  if (done)
  {
    std::lock_guard<std::mutex> lock(cv_m);
    if (done)
    {
      container = {1, 2, 3, 4};
      done = true;
    }
  }

  container.front();
}
```

lindsayad:
Well actually this code based on Preshing's sequentially consistent atomics pattern is both `ThreadSanitizer` and `helgrind` clean and is very human-readable and concise. But he does say the machine code tends to be less efficient.

```c++
#include <thread>
#include <vector>
#include <atomic>
#include <mutex>

std::mutex cv_m;
std::atomic<bool> done = false;
std::vector<int> container;

void
fillContainer()
{
  if (!done)
  {
    std::lock_guard<std::mutex> lock(cv_m);
    if (!done)
    {
      container = {1, 2, 3, 4};
      done = true;
    }
  }

  container.front();
}

int
main()
{
  std::thread t1(fillContainer), t2(fillContainer);
  t1.join();
  t2.join();
}
```

lindsayad:
Omg, helgrind doesn't error if you pass `-O1` but does if you pass `-O0`. Stupid

friedmud:
Note that passing with Thread Sanitizer doesn't mean it's correct.  Thread Sanitizer only detects data races _if they happen_... so you could run it 100 times with TSan and still have a bug and not know it.

Yeah - using full acquire/release loads/stores (like in your second example since that's the default for atomics) will guarantee that everything works... but is definitely not as fast (threads have to stop and publish on the first read no matter what).

In your last comment are you saying that your second example errors in Helgrind?  That's what I've been reading... that Helgrind can't understand atomic guarded sections.

friedmud:
@dschwen pointed me to a great article on memory operation ordering: https://www.arangodb.com/2021/02/cpp-memory-model-migrating-from-x86-to-arm/

friedmud:
@jwpeterson I agree - and I'm sure that it uses the normal Valgrind suppression mechanism (it's just a module of Valgrind - right?)

lindsayad:
> In your last comment are you saying that your second example errors in Helgrind? That's what I've been reading... that Helgrind can't understand atomic guarded sections.

Correct ... with `-O0` (but not with `-O1` ... I am used to seeing false positives crop up with more optimization ... but not seeing false positives crop up with *less* optimization!)

lindsayad:
> Note that passing with Thread Sanitizer doesn't mean it's correct. Thread Sanitizer only detects data races _if they happen_

Really? The intro paper I read on `ThreadSanitizer` made it seem like it used a lot of valgrind but produced less false positives. Anyway if this is the case, I wish they would document it clearly on their github repo or in the clang doc. It is currently not documented if this is the case.

dschwen:
Sooo... what's the gold standard now? This code (quite old actually) https://github.com/libMesh/libmesh/blob/c503a9a59f6b0169cb81c20cfa02fe665fd0cc09/src/numerics/petsc_vector.C#L1225-L1279 looks like it is doing all the fencing and atomics, or am I missing something?

dschwen:
here is the atomic part in the header https://github.com/libMesh/libmesh/blob/c503a9a59f6b0169cb81c20cfa02fe665fd0cc09/include/numerics/petsc_vector.h#L358-L363

(6yo code)

lindsayad:
For one thing, I *think* that if you are going to use `std::memory_order_relaxed` for loads and stores of your global atomic (`_array_is_present` for us), then you will have to create a local variable in order to avoid races. This is Preshing's example:
```c++
std::atomic<Singleton*> Singleton::m_instance;
std::mutex Singleton::m_mutex;

Singleton* Singleton::getInstance() {
    Singleton* tmp = m_instance.load(std::memory_order_relaxed);
    std::atomic_thread_fence(std::memory_order_acquire);
    if (tmp == nullptr) {
        std::lock_guard<std::mutex> lock(m_mutex);
        tmp = m_instance.load(std::memory_order_relaxed);
        if (tmp == nullptr) {
            tmp = new Singleton;
            std::atomic_thread_fence(std::memory_order_release);
            m_instance.store(tmp, std::memory_order_relaxed);
        }
    }
    return tmp;
}
```
I'm just not sure that a `std::atomic_thread_fence(std::memory_order_acquire);` with nothing ahead of it in the code makes sense. I could be wrong though

lindsayad:
We also have to make sure that we are doing *something* different with `_values_read_only` ... that definitely had races before this PR

dschwen:
And can/should we use C++11 threading stuff in libmesh now? Is there still a reason to keep the  `Threads::spin_mutex` stuff around?

jwpeterson:
I think the idea was that `Threads::spin_mutex` should use the "right" type of mutex under the hood. So if you are using TBB threads, you would get a TBB mutex, if you are using C++11 threads, you would get a `std::mutex`, etc. In practice I don't think it is working like this currently :(

roystgnr:
For my own sake I wouldn't mind getting rid of non-C++11 thread support.  We're way past the point where we might want to worry about having to use a fallback on compilers which screw that up.  I could imagine users deliberately configuring libMesh to use another threading system for interoperability with third-party code, though...

roystgnr:
Oh, and of course even if getting rid of other backends, we'd still keep the shims to C++11 around for a couple years for backward compatibility.

jwpeterson:
I think we were trying to be compatible with the types of threads *other* libraries in the software stack are using. I don't know if this is currently an issue for us... off the top of my head I think boomeramg uses pthreads, but I can't really think of others. Threaded MKL?

lindsayad:
FWIW here is output with `ThreadSanitizer` for MOOSE's `initial_adapt.i` test with the old code

[tsan.txt](https://github.com/libMesh/libmesh/files/7953753/tsan.txt)

dschwen:
Ok, I still don't know what the reference implementation of a double checked lock in "libMesh C++" should look like. We do have `Threads::atomic<T>`, but that is a manual spin lock implementation for pthreads, where C++11 `std::atomic<T>` operations are compiled to machine instructions (`lock` prefixed). 
If our thread model is determined by other libraries that means we cannot use C++11 threads anyways and miss out on the sweet lock free atomics. And what about memory fences? I don't see any shims for those in libmesh. Is it safe/correct to use `std::atomic_thread_fence` for those regardless of the thread model?

lindsayad:
Possibly misleading post in a PR about threads ... but It would be nice if GitHub supported threaded conversations like gitlab does

Anyway, the cpp core guidelines recommend using acquire and release operations as opposed to fences for a doubly-checked lock pattern, see [here](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconc-double-pattern). Moreover, `std::atomic_thread_fence` has the unfortunate feature of [not being supported by ThreadSanitizer](https://github.com/google/sanitizers/issues/1352) and it's certainly not supported by helgrind.

Making this thread even longer by pasting in even more code snippets ... If we consider
```c++
#include <thread>
#include <atomic>
#include <mutex>

std::mutex cv_m;
std::atomic<bool> done = false;
int a;

void
maybeWriteDefinitelyRead()
{
  bool local_done = done.load(std::memory_order_relaxed);
  std::atomic_thread_fence(std::memory_order_acquire);
  if (!local_done)
  {
    std::lock_guard<std::mutex> lock(cv_m);
    local_done = done.load(std::memory_order_relaxed);
    if (!local_done)
    {
      a = 0;
      local_done = true;
      std::atomic_thread_fence(std::memory_order_release);
      done.store(local_done, std::memory_order_relaxed);
    }
  }

  const auto val = a;
}

int
main()
{
  std::thread t1(maybeWriteDefinitelyRead), t2(maybeWriteDefinitelyRead);
  t1.join();
  t2.join();
}
```
this errors with ThreadSanitizer even though it is actually a [stronger guard](https://preshing.com/20131125/acquire-and-release-fences-dont-work-the-way-youd-expect/) than the following code
```c++
#include <thread>
#include <atomic>
#include <mutex>

std::mutex cv_m;
std::atomic<bool> done = false;
int a;

void
maybeWriteDefinitelyRead()
{
  if (!done.load(std::memory_order_acquire))
  {
    std::lock_guard<std::mutex> lock(cv_m);
    if (!done.load(std::memory_order_relaxed))
    {
      a = 0;
      done.store(true, std::memory_order_release);
    }
  }

  const auto val = a;
}

int
main()
{
  std::thread t1(maybeWriteDefinitelyRead), t2(maybeWriteDefinitelyRead);
  t1.join();
  t2.join();
}
```
In that vein, since we get the same thread-safe code with a weaker guard (atomic operation store and release as opposed to fence "store" and "release"), I suggest we go with that, especially considering that it's recognized as being safe by `ThreadSanitizer` (although still not by helgrind).

lindsayad:
See #3146 for new implementation

dschwen:
And this is what we use regardless of the libmesh threading model? (Openmo, pthreads, TBB)

jwpeterson:
It think the new PR is very close to what we were doing before, so in that sense it should work well for all of our current use cases. I don't think we have a good answer larger question of thread interoperability... I think it's important but beyond the scope of this PR.

PR #3130, opened 2022-01-19
Make CompareDofObjectsByID available to all

lindsayad:
This is a useful struct for us in MOOSE mortar

### DIFF
--- a/include/base/dof_object.h
+++ b/include/base/dof_object.h
@@ -1325,6 +1325,22 @@ unsigned int DofObject::system_var_to_vg_var (const unsigned int s,
   return (var - accumulated_sum);
 }
 
+/**
+ * Comparison object to use with DofObject pointers.  This sorts by id(),
+ * so when we iterate over a set of DofObjects we visit the objects in
+ * order of increasing ID.
+ */
+struct CompareDofObjectsByID
+{
+  bool operator()(const DofObject * a,
+                  const DofObject * b) const
+  {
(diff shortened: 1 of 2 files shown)


include/base/dof_object.h
```diff
@@ -1325,6 +1325,31 @@ unsigned int DofObject::system_var_to_vg_var (const unsigned int s,
...
+  bool operator()(const DofObject * a,
+                  const DofObject * b) const
+  {
+    libmesh_assert (a);
+    libmesh_assert (b);
+
+    return a->id() < b->id();
+  }
+
+  bool operator()(DofObject * a,
```
jwpeterson, line 1344: I see that you added a version taking non-const inputs, but I don't understand why it's needed. I guess you got a compiler error that caused you to add it?

lindsayad: I do have a container of pointers to non-const and I preeemptively assumed I would get a compiler error saying that it was unable to convert pointer to non-const to pointer to const ... but I can double-check

lindsayad: Huh I can't believe I didn't know that you can implicitly convert pointer to non-const to pointer to const. I mean it makes sense but for some reason I thought it was illegal

lindsayad: Ok I've removed the non-const pointer overload

jwpeterson (approved):

PR #3131, opened 2022-01-21
Set PETSc mat/vec types at runtime

fdkong:
Improve the interface to allow users to set mat/vec types at runtime.

This is required when we try to leverage PETSc GPU solvers in MOOSE

### DIFF
--- a/include/numerics/petsc_vector.h
+++ b/include/numerics/petsc_vector.h
@@ -658,9 +658,8 @@ void PetscVector<T>::init (const numeric_index_type n,
   // create a sequential vector if on only 1 processor
   if (this->_type == SERIAL)
     {
-      ierr = VecCreateSeq (PETSC_COMM_SELF, petsc_n, &_vec);
-      CHKERRABORT(PETSC_COMM_SELF,ierr);
-
+      ierr = VecCreate(PETSC_COMM_SELF, &_vec);CHKERRABORT(PETSC_COMM_SELF,ierr);
+      ierr = VecSetSizes(_vec, petsc_n, petsc_n); CHKERRABORT(PETSC_COMM_SELF,ierr);
       ierr = VecSetFromOptions (_vec);
       CHKERRABORT(PETSC_COMM_SELF,ierr);
     }
@@ -670,15 +669,14 @@ void PetscVector<T>::init (const numeric_index_type n,
 #ifdef LIBMESH_HAVE_MPI
       PetscInt petsc_n_local=cast_int<PetscInt>(n_local);
       libmesh_assert_less_equal (n_local, n);
-      ierr = VecCreateMPI (this->comm().get(), petsc_n_local, petsc_n,
-                           &_vec);
-      LIBMESH_CHKERR(ierr);
+      // Use more generic function instead of VecCreateSeq/MPI
+      ierr = VecCreate(this->comm().get(), &_vec);LIBMESH_CHKERR(ierr);
+      ierr = VecSetSizes(_vec, petsc_n_local, petsc_n); LIBMESH_CHKERR(ierr);
 #else
       libmesh_assert_equal_to (n_local, n);
-      ierr = VecCreateSeq (PETSC_COMM_SELF, petsc_n, &_vec);
-      CHKERRABORT(PETSC_COMM_SELF,ierr);
+      ierr = VecCreate(PETSC_COMM_SELF, &_vec);CHKERRABORT(PETSC_COMM_SELF,ierr);
+      ierr = VecSetSizes(_vec, petsc_n, petsc_n); CHKERRABORT(PETSC_COMM_SELF,ierr);
 #endif
-
       ierr = VecSetFromOptions (_vec);
       LIBMESH_CHKERR(ierr);
     }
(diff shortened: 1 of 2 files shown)


fdkong:
@roystgnr

fdkong:
We might need to add prefix into mat and vec. I did not add that yet. What are your opinions?

roystgnr:
"But please do not merge it yet. I need to do more tests on HPC"

include/numerics/petsc_vector.h
```diff
@@ -675,15 +674,17 @@ void PetscVector<T>::init (const numeric_index_type n,
 #ifdef LIBMESH_HAVE_MPI
       PetscInt petsc_n_local=cast_int<PetscInt>(n_local);
       libmesh_assert_less_equal (n_local, n);
-      ierr = VecCreateMPI (this->comm().get(), petsc_n_local, petsc_n,
-                           &_vec);
-      LIBMESH_CHKERR(ierr);
+      // Let us a create vector but not specify a vector type
+      // By default, mpivec will be created.
+      // Motivation: to allow users specify vector type at runtime to
```
roystgnr, line 679: What happens to all the users who *don't* specify vector type at runtime?  I guess PETSc must be smart enough now to infer from the last two `VecSetSizes` parameters which backend it should switch to?  That would be awesome if so.  Is it a recent feature or something we need to check version numbers for?

rtmills: Hi @roystgnr, I'm one of the PETSc developers who is working with @fdkong on a project, so I've taken a look at this PR. If users don't specify any vector type, then PETSc will default to creating a `Vec` of type `VECSTANDARD`, which is essentially exactly what would happen before @fdkong's proposed changes (`VECSTANDARD` defaults to a plain-old `VECSEQ` on one MPI process or a `VECMPI` on multiple MPI processes). The type will be set in the `VecSetFromOptions()` call. We generally recommend using `VecCreate()` as in Fande's changes, rather than using `VecCreateSeq()` or `VecCreateMPI()`, actually, since the right thing should then happen automatically based on the size of the MPI communicator.

I should note, however, that even when using `VecCreateSeq()` and `VecCreateMPI()`, if the user is asking for a GPU type via the PETSc options data base (that is, by the command line, in most cases), PETSc will still convert the `Vec` to the specified GPU type when `VecSetFromOptions()` is called. (I just verified this in a debugger.) So, the use of these isn't the cause of whatever issues @fdkong is running into with getting GPU types working. It looks like libMesh does create some vectors from a `DM`. If some of those vectors need to be set to GPU types, then you'll need to specify something like `-dm_vec_type cuda` (using whatever appropriate prefix is appropriate for your DMs, if you've set something). I don't think it's the case with libMesh, but if there are vectors coming from something like `MatCreateVecs()`, then you'll need to make sure that the matrix is one of the GPU types, or, if it is a `MATSHELL`, you may need to add some logic to set the corresponding GPU type appropriately by calling `MatShellSetVecType()`.

fdkong: >I should note, however, that even when using VecCreateSeq() and VecCreateMPI(), if the user is asking for a GPU type via the PETSc options data base (that is, by the command line, in most cases),

I have another half PR that apply different prefixes for Vec and ghostedVec.

The actual reason was that `VecCreateGhost` does not do correct things on GPU

fdkong:
@roystgnr It is ready for review.  

Something like this worked on GPU

`../../../moose_test-opt    -i 2d_diffusion_gpu.i Mesh/dmg/nx=20  Mesh/dmg/ny=20   -use_gpu_aware_mpi 0 -mat_type aijcusparse   -vec_type cuda -pc_type gamg -mg_levels_pc_type jacobi   -ksp_rtol 1e-20     -log_view    -snes_view`

roystgnr (approved):
I don't see any problems myself.  Going to give @jwpeterson a day to look things over if he wants to before we merge, though.

jwpeterson:
> Going to give @jwpeterson a day to look things over

I'm OK with this provided that we did not introduce any code that needs to be guarded in older versions of PETSc. (IIRC we claim to support all the way back to 3.5.0?) Also, it looks like the history could stand to be cleaned up (2 commits with the same commit message) or maybe just a squash-merge is sufficient.

roystgnr:
Yikes, good point.  I'd forgotten what we're trying to support, so I told @loganharbour that the petsc/3.8 module he had lying around was fine so use as a minimum in the new CI environments (we check it in the Min clang debug and Min gcc debug recipes).  We may need a new-old module to test support farther back.

jwpeterson:
IMO there's nothing magic about 3.5.0, it was arbitrarily chosen some time ago. We can just make it 3.8.0 instead if that's convenient. Also I don't really suspect that this PR introduces new code... `VecCreate` must date back to the very beginning of PETSc?

roystgnr:
> We can just make it 3.8.0 instead if that's convenient.

Let's do that, then?  I've already dumped more than enough on Logan's plate.  And as long as we're now requiring C++17 compilers and standard libraries from everybody, requiring a 2017-or-later PETSc version is just as reasonable an upgrade.

fdkong:
>all the way back to 3.5.0?

If we want to support that in libmesh, I would like to see that we have CI in libmesh to check that.  In MOOSE, we do not really spend much time on old PETSc anymore. The oldest one in MOOSE is PETSc-3.11

I am not even sure I can build PETSc-3.5 on my local desktop, lol

loganharbour:
> I am not even sure I can build PETSc-3.5 on my local desktop, lol

Ha, exactly. 3.8 was the easiest I could compile without trying too hard.

fdkong:
>Also, it looks like the history could stand to be cleaned up (2 commits with the same commit message) or maybe just a squash-merge is sufficient.

Done

PR #3134, opened 2022-01-24
Fix non-MPI build

mkhon:


jwpeterson:
Thanks for reporting this issue @mkhon. I sent you an invite to a group that gets their libmesh PRs tested automatically.

roystgnr:
> I think the right fix here may be to move the declarations outside of the #if defined(LIBMESH_HAVE_MPI) region

Agreed.  I'll get a PR up with that ASAP, once my own tests are done running and I'm confident there aren't regressions elsewhere.  We have a non-MPI TIMPI build in our CI recipes, but our non-MPI libMesh build is a non-default recipe.  My apologies for not thinking to enable it for the PR that introduced as bug (as well as for writing the code that introduced this bug).

mkhon:
@roystgnr I updated this PR to have the declarations moved outside of #ifdef block

roystgnr:
Thanks!  I'll merge this rather than #3136, then, so you get credit in the logs.

roystgnr (approved):

PR #3135, opened 2022-01-24
Fix unistd.h issues when building with MS VC

mkhon:
- Use LIBMESH_HAVE_UNISTD_H
- Remove unneeded #include's

This fixes the majority of MS VC build issues related to missing unistd.h

### DIFF
--- a/src/base/print_trace.C
+++ b/src/base/print_trace.C
@@ -46,7 +46,9 @@
 #include "libmesh/print_trace.h"
 #include "libmesh/libmesh.h"
 
+#ifdef LIBMESH_HAVE_UNISTD_H
 #include <unistd.h>  // needed for getpid()
+#endif
 #include <fstream>
 #include <sstream>
 #include <string>
--- a/src/mesh/namebased_io.C
+++ b/src/mesh/namebased_io.C
@@ -46,8 +46,10 @@
 #include <iomanip>
 #include <fstream>
 #include <vector>
-#include <sys/types.h> // getpid
-#include <unistd.h>
+#ifdef LIBMESH_HAVE_UNISTD_H
+#include <sys/types.h>
+#include <unistd.h>  // for getpid()
+#endif
 
 
 namespace libMesh
--- a/examples/adjoints/adjoints_ex3/output.C
+++ b/examples/adjoints/adjoints_ex3/output.C
@@ -1,7 +1,10 @@
 #include <fstream>
-#include <unistd.h>
 
 #include "libmesh/libmesh_common.h"
+
+#ifdef LIBMESH_HAVE_UNISTD_H
+#include <unistd.h> // needed for truncate()
+#endif
 
 #include "output.h"
 
--- a/examples/fem_system/fem_system_ex2/fem_system_ex2.C
+++ b/examples/fem_system/fem_system_ex2/fem_system_ex2.C
@@ -41,7 +41,6 @@
 #include <iostream>
 #include <sstream>
 #include <string>
-#include <unistd.h>
 
 using namespace libMesh;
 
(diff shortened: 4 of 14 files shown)


src/base/print_trace.C
```diff
@@ -46,7 +46,11 @@
 #include "libmesh/print_trace.h"
 #include "libmesh/libmesh.h"
 
+#ifdef LIBMESH_GDB_COMMAND
```
jwpeterson, line 49: `#ifdef LIBMESH_GDB_COMMAND` always evaluates to `true`, it is set to the string "no" by default, so I think those lines you added are redundant.

src/mesh/namebased_io.C
```diff
@@ -46,8 +46,10 @@
 #include <iomanip>
 #include <fstream>
 #include <vector>
-#include <sys/types.h> // getpid
-#include <unistd.h>
+#ifdef LIBMESH_HAVE_UNISTD_H
+#include <sys/types.h>
+#include <unistd.h>  // for getpid()
```
jwpeterson, line 51: This seems to only address half of the issue, since we still call `getpid()` elsewhere in this file. Does Windows have its own `getpid()` that requires no additional header inclusions to call? Or are we getting that inclusion by accident under Windows? According to some very brief investigating that I did, it looks like `getpid` is defined in [<process.h>](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/getpid?view=msvc-170&viewFallbackFrom=vs-2019) on Windows.

jwpeterson (commented):
Looks like a good change overall, but I'm wondering why none of the previous attempts to get libMesh compiling on Windows ran into these issues. Are you using MSys2?

roystgnr:
> I'm wondering why none of the previous attempts to get libMesh compiling on Windows ran into these issues.

Have we ever had serious attempts before to get libMesh compiling on **Win32**?  I had libMesh working with Cygwin for a few years, and I think someone else might have added tweaks for MinGW at some point, and I'd bet we work out of the box with Microsoft's current "forget it, let's just host a Ubuntu distribution on top of us"  WSL ... but I didn't think we ever came close to working with straight-up Win32.  IIRC there was someone who tried it very many years ago and they ran into some C++ idiom that we used everywhere but that Visual C++ didn't yet like, and they gave up shortly thereafter.

Intellectually I'm super curious about this patch (aren't Windows users pretty much uniformly using Win64 now?), but practically I worry a bit: this is exactly the sort of thing we'd really want to test for regressions before every libMesh release, but I don't think we'll be getting Windows CI boxes any time soon.  @mkhon - if you manage to get the rest of the Windows build issues worked out, could you add build instructions to the end of our README.md ?  If they work on 64-bit Windows too, I've at least got a dual-boot laptop I could use to try them out by hand.

dschwen:
@roystgnr not straight up win32, but msys2 (which also requires use of Win32 APIs). We eventually gave up on that (libmesh worked sort of, but we ran into linking issues with MOOSE modules and apps). In the end we confirmed that WSL does indeed work out of the box.

src/base/print_trace.C
mkhon: Thank you. Corrected

src/mesh/namebased_io.C
mkhon: This is only a first set of changes for windows build.

Proper getpid(), gettimeofday(), mkstemp() definitions would require additional automake/autoconf-fu, so this will submitted separately.

mkhon:
@roystgnr This patch is a part of changes to build libmesh with libtool (msys2) and MS VC on Windows.
See microsoft/vcpkg#22810

PR #3136, opened 2022-01-24
Fix builds with MPI disabled/unavailable

roystgnr:
This saves a couple lines over the fix proposed in #3134 for the
libmesh.C regression introduced by #3092

I'm not hitting any other regressions in my own testing.


jwpeterson (approved):

roystgnr:
@mkhon updated #3134 at the same time as I was testing this, so we'll merge that version.

PR #3137, opened 2022-01-24
Fix MS VC 2019 build

mkhon:
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\include\xmemory(714): error C2280: 'std::pair<const libMesh::FEType,std::unique_ptr<libMesh::FEAbstract,std::default_delete<libMesh::FEAbstract>>>::pair(const std::pair<const libMesh::FEType,std::unique_ptr<libMesh::FEAbstract,std::default_delete<libMesh::FEAbstract>>> &)': attempting to reference a deleted function

### DIFF
--- a/src/systems/fem_context.C
+++ b/src/systems/fem_context.C
@@ -193,19 +193,8 @@ void FEMContext::init_internal_data(const System & sys)
   // Reserve space for the FEAbstract and QBase objects for each
   // element dimension possibility (0,1,2,3)
 
-  // Below is a workaround for the ICC 19. The original code was:
-  //
-  // _element_fe.resize(4);
-  // _side_fe.resize(4);
-
-  _element_fe.clear();
-  for (int i=0; i<4; ++i)
-    _element_fe.push_back(std::map<FEType, std::unique_ptr<FEAbstract>>());
-
-  _side_fe.clear();
-  for (int i=0; i<4; ++i)
-    _side_fe.push_back(std::map<FEType, std::unique_ptr<FEAbstract>>());
-
+  _element_fe = std::vector<std::map<FEType, std::unique_ptr<FEAbstract>>>(4);
+  _side_fe = std::vector<std::map<FEType, std::unique_ptr<FEAbstract>>>(4);
   _element_fe_var.resize(4);
   _side_fe_var.resize(4);
 


src/systems/fem_context.C
```diff
@@ -190,25 +194,6 @@ void FEMContext::use_unweighted_quadrature_rules(int extra_quadrature_order)
 
 void FEMContext::init_internal_data(const System & sys)
 {
-  // Reserve space for the FEAbstract and QBase objects for each
```
jwpeterson, line 193: We can't delete this code because it's a workaround for another compiler (Intel). In addition this potentially changes the behavior of the code because `init_internal_data()` could theoretically be called multiple times (not sure if it actually is) and we would need `_element_fe`, `_side_fe`, etc. to be cleared each time.

mkhon: 1) How can I test ICC build (e.g on Linux)?

2) Currently init_internal_data() is private and is only called from the constructor.

3) The problem with MS VC 2019 is that neither of this works:

```
    std::vector<std::map<int, std::unique_ptr<const char *>>> v1;
    v1.resize(1);
    v1.push_back(std::map<int, std::unique_ptr<const char *>>());
    v1.push_back(std::move(std::map<int, std::unique_ptr<const char *>>()));
```

jwpeterson: Maybe try `emplace_back()`?

> How can I test ICC build (e.g on Linux)?

We no longer have CI testing of the Intel compiler, unfortunately. If you can find some other workaround that fixes MSVC then we could consider `ifdef`'ing based on the compiler. Honestly I find it surprising that two different compilers have issues with the same (otherwise innocuous looking) lines of code. I'm pretty sure the code is standards conforming and it's the compilers that are wrong.

mkhon: I think the problem is not in the compiler but in their STL implementations.

The problem with this piece of code is that std::unique_ptr<> is non-copyable (it has copy constructor explicitly deleted), but std::vector::push_back() may involve copying.

The same about std::vector::resize() (which may re-allocate memory and copy objects from old location to new).

resize() and push_back() implementations can use move constructors, but likely ICC 19 and MS VC 2019 got it wrong.

mkhon: @jwpeterson yes, I also tried emplace_back() - produces the same error with MS VC 2019

mkhon: I submitted a less intrusive change, but I don't think that different codepaths and #ifdef-fu is a good thing.

I am almost sure construction of vector of specified size will work in older ICC (ICC 2021 and later is clang and it should be ok even with the original code).

jwpeterson: > I am almost sure construction of vector of specified size will work in older ICC

Yes, it's possible you are correct. Looking back to the original change (7e6abb44e0b2da93e501acf99d4c13fa56a15df7) the only offending lines were the `resize()` calls, no compiler that I'm aware of has complained about the constructor initialization syntax. This raises the question of whether MSVC/ICC19 would accept the syntax
```
_element_fe = std::vector<std::map<FEType, std::unique_ptr<FEAbstract>>>(4);
```
for reinitializing the vector with four empty maps.

roystgnr:
Rather than clearing the vector of maps and then restoring empty maps, would it work to loop over the maps and clear them individually?

jwpeterson (approved):
Latest diff looks good to me. I will just throw in a comment that these two lines are a `resize()` workaround for the Intel 19 and MSVC STL implementations before merging (once the CI all passes again).

PR #3138, opened 2022-01-25
Differentiate tasks for InFEM (InFEM::reinit)

BalticPinguin:
Finally I have adjusted the calculation of `InfFE`s to only compute the required quantities.
Since the overall structure of the calculation differs from the `FE`-variant, I have added some other `calculate_<something>`s:'

While I didn't see much sense in having a separate `calculate_dphiref` here, I made a distinction between `scaled` and unscaled sets, i.e. if the quantity itself is requested or if the `xdecay` or `overdecay` variant needs to be calculated.

Further, I added a separate `calculate_jxw`-flag because after restructuring one must test both the weighted and unweighted JxW for positiveness  (at least it makes sense). Testing the unweighted jacobian (`inv_jac <= 1e-10`) however often resulted in errors of my (otherwise working) application, so I'd like to avoid its computation as long as not explicitly requested.

`include/fe/inf_fe.h` from running `reindent.sh`; if you don't like it, let me know.

Best regards

### DIFF
--- a/src/fe/inf_fe.C
+++ b/src/fe/inf_fe.C
@@ -30,20 +30,7 @@
 #include "libmesh/int_range.h"
 #include "libmesh/auto_ptr.h"
 #include "libmesh/type_tensor.h"
-
-
-namespace {
-  // Put this outside a templated class, so we only get 1 warning
-  // during our unit tests, not 1 warning for each of the zillion FE
-  // specializations we test.
-#ifdef LIBMESH_ENABLE_SECOND_DERIVATIVES
-  void inffe_hessian_warning () {
-    libmesh_warning("Warning: Second derivatives for Infinite elements"
-                    << " are not yet implemented!"
-                    << std::endl);
-  }
-#endif
-}
+#include "libmesh/fe_interface.h"
 
 
 namespace libMesh
@@ -56,6 +43,11 @@ template <unsigned int Dim, FEFamily T_radial, InfMapType T_map>
 InfFE<Dim,T_radial,T_map>::InfFE (const FEType & fet) :
   FEBase       (Dim, fet),
 
+  calculate_map_scaled(false),
+  calculate_phi_scaled(false),
+  calculate_dphi_scaled(false),
+  calculate_xyz(false),
+  calculate_jxw(false),
   _n_total_approx_sf (0),
   _n_total_qp        (0),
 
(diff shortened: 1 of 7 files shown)


roystgnr (approved):
No obvious problems here that I could see.

How much speedup do you typically see from this?

BalticPinguin:
To be honest, in my application I don't see any difference: The time for solving gives ~80% of time and since I do some rather involved calculations on the FEM, the assembly of InFEM is so light that its timings change more between different runs than by this improvement.
And I didn't compare the performance of the examples; if you want, I can make a clean comparison there.

Besides speedup, I see the advantage that now one can evaluate `phi` literally at infinity; before, I had to go "close enough to infinity" which is technically fine, but leads to long discussions when I try to explain the computational scheme...
In that sense, the speedup is not during computation but rather in the communication of the procedure :-)

BalticPinguin:
To answer your question, @roystgnr : In one of my calculations I got faster by a factor of 5 with these changes; I think that most calls here only requested `phi`; so it is a quite optimistic estimate; on the other hand besides the assembly, in most cases point-values are enough to have and they are computationally comparatively cheap.

src/fe/inf_fe.C
```diff
@@ -339,6 +322,78 @@ void InfFE<Dim,T_radial,T_map>::reinit(const Elem * inf_elem,
...
+#endif
+      this->calculate_phi_scaled = this->calculate_dphi_scaled = this->calculate_xyz = true;
+      if (FEInterface::field_type(fe_type.family) == TYPE_VECTOR)
+        {
+          this->calculate_curl_phi = true;
+          this->calculate_div_phi  = true;
+        }
+    }
+#else //LIBMESH_ENABLE_DEPRECATED
+  libmesh_assert (this->calculate_nothing ||
```
roystgnr, line 358: Does this chain of boolean operators make sense?  I'm trying to parse it (after a `-Werror -Wparentheses` warning kept me from compiling) and I'm failing.

On a related note, we must need to add `--enable-paranoid-warnings --enable-werror` to our infinite elements recipe in CI?

BalticPinguin: oh, you are right: I think, the `!this->`-terms must have no negation here and the `&&` should be changed to `||` to make actually sense.
I think, the issue here is that one cannot chain-up statements with mixed `&&` and `||`; so it should be fixed by adjusting this?  But I currently don't see why `-Wparentheses` prevents parsing of that statement...

I obviously use `--enable-deprecated`; so I didn't catch this mistake.

PR #3139, opened 2022-01-25
Add Intel(R) icc 21 support

mkhon:
Requires libMesh/autoconf-submodule#6 and libMesh/TIMPI#89


jwpeterson:
I merged your autoconf-submodule PR, so I think, in addition to this change we'll want to include a submodule update on this branch, but I can take care of it from here.

PR #3141, opened 2022-01-26
outer_product: scalar * vector -> vector

lindsayad:
I'd like the same piece of code to work with different 'a' argument
types in a * b = c, and I want the result to never collapse rank


roystgnr (approved):
This seems entirely reasonable.  But was there any reason for only adding the `scalar outer* vector` case, and not also adding the `vector outer* scalar` case while we're at it?

roystgnr:
Oh, and obviously a unit test would not be amiss if you have time.

jwpeterson:
Not sure if this is relevant for this type of outer product, but do we need to worry about complex numbers here? Looks like we do that currently for outer products between TypeVectors.

"For complex vectors, it is often useful to take the conjugate transpose of v" [Outer_product wikipedia](https://en.wikipedia.org/wiki/Outer_product)

lindsayad:
So my issue with introducing `libmesh_conj` is that it's going to make a copy and in the place I'm using this I'm actually working with `DualNumbers` for which copies are non-trivial. 

On a somewhat separate note, I also realize @roystgnr I think that our practice is to instantiate `DualNumber<std::complex>` for which I think `libmesh_conj` is not going to do the right thing :-/

include/numerics/type_vector.h
```diff
@@ -1168,6 +1168,24 @@ struct CompareTypes<TypeVector<T>, TypeVector<T2>>
...
+
+  return ret;
+}
+
+template <typename T, typename T2, typename std::enable_if<ScalarTraits<T2>::value, int>::type = 0>
+TypeVector<typename CompareTypes<T, T2>::supertype>
+outer_product(const TypeVector<T> & a, const T2 & b)
+{
+  return outer_product(b, a);
+}
```
roystgnr, line 1188: I feel like if I point out "this isn't correct for non-commutative `*`" then I'm being an ass, but if I don't point out "this isn't correct for non-commutative `*`" then karma will determine that Now Is The Time for someone to try and start stuffing quaternions or something into a `TypeVector`.

roystgnr:
Is `libmesh_conj` definitely going to make a copy?  The default implementation is an inline `foo(T x){return x;}`

Trying things out [on Godbolt](https://godbolt.org/z/ava5Wv5rx) it appears that we definitely get a copy constructor call from the parameter being passed by value, but that (at least with a class wrapping int and a constructor that has no side effects) the construction can be completely optimized away.

That said - `libmesh_conj` certainly isn't going to do the right thing here.  We handle `std::complex<T>` with a specialization, we make everything else the identity, and that's not right for anything that's conceptually complex but not `std::complex`.  But my inclination would be to use `libmesh_conj` here anyway and rely on users (which, yeah, might just be "us" here...) to add more overloads/specializations as necessary.

lindsayad:
> rely on users (which, yeah, might just be "us" here...) to add more overloads/specializations as necessary.

You mean add overloads of `libmesh_conj`?

> would be to use `libmesh_conj` here anyway

Ok I'll do that. If it shows up in profiling that the compiler is not optimizing away copies and this is a hot-spot, then we can revisit

roystgnr:
This looks perfect to me now.

jwpeterson (approved):
Beautiful code

PR #3142, opened 2022-01-26
Enable EIM using POD

dknez:
This is equivalent to the "Discrete Empirical Interpolation Method", which is described in the literature. In some cases (mainly with relatively few parameters) the POD-based approach can be more efficient than the Greedy approach, so it is good to have both options.

### DIFF
--- a/src/reduced_basis/rb_eim_construction.C
+++ b/src/reduced_basis/rb_eim_construction.C
@@ -164,13 +164,16 @@ void RBEIMConstruction::set_best_fit_type_flag (const std::string & best_fit_typ
     {
       best_fit_type_flag = PROJECTION_BEST_FIT;
     }
+  else if (best_fit_type_string == "eim")
+    {
+      best_fit_type_flag = EIM_BEST_FIT;
+    }
+  else if (best_fit_type_string == "pod")
+    {
+      best_fit_type_flag = POD_BEST_FIT;
+    }
   else
-    if (best_fit_type_string == "eim")
-      {
-        best_fit_type_flag = EIM_BEST_FIT;
-      }
-    else
-      libmesh_error_msg("Error: invalid best_fit_type in input file");
+    libmesh_error_msg("Error: invalid best_fit_type in input file");
 }
 
 void RBEIMConstruction::print_info()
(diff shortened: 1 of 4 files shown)


src/reduced_basis/rb_eim_construction.C
```diff
@@ -54,6 +54,54 @@
...
+
+// Recall that we have:
+// typedef std::map<dof_id_type, std::vector<std::vector<Number>>> QpDataMap;
+// typedef std::map<std::pair<dof_id_type,unsigned int>, std::vector<std::vector<Number>>> SideQpDataMap;
+
+// Implement u <- u + k*v
+template <typename DataMap>
+void add(DataMap & u, const Number k, const DataMap & v)
+{
+  for (auto & [key,val] : u)
```
jwpeterson, line 70: This is passing all the CI tests, so @roystgnr do we officially allow structured bindings now? I thought it was a C++17 thing, and I can't remember if we went to C++14 or jumped all the way to C++17.

jwpeterson (approved):
Looks good to me though I'm not quite sure about the coverage report showing 0%. Does the coverage count the examples or only the unit tests? @dknez updated reduced_basis_ex4 to use this new code, so there is actually some coverage of it.

src/reduced_basis/rb_eim_construction.C
roystgnr: We're *almost* ready to officially allow *basic* C++17 structured bindings.  Personally I'd rather wait until we have proper CI going before letting them in.

The catches here are:

1. We're not ready yet, only *almost* ready.  We're literally getting new module trees built with new min-gcc and min-clang compilers right now.  Currently CI is testing with a min-clang that handles (and so enables because I foolishly didn't ask it to go lower) C++17, and we're not testing with a min-gcc (because the only older module we already had handy handles C++11 but not C++14), so we get false positives here.

2. We'll be getting *basic* C++17 with our min-clang, not the handful of Defect Reports that have led to changes in the standard since then, one of which is the way structured bindings work (or, in the original standard, fail to work) with non-public member variables.  Not an issue here, and I'd figured it wouldn't be an issue for a while, but if we're chomping at the bit to use structured bindings then maybe I made the wrong call?

dknez: I pushed an update which removed the structured binding, so hopefully that resolves this.

roystgnr:
That 0-new-lines-coverage ... I'm guessing the trouble is that reduced_basis_ex4 won't run without Cap'n Proto, and we don't build with that in CI.

jwpeterson:
> the trouble is that reduced_basis_ex4 won't run without Cap'n Proto

OK, this definitely seems plausible. Since @dknez dropped the structured binding code and reduced_basis_ex4 is working for him, I'm going to merge now.

PR #3145, opened 2022-01-27
Detect PETSc HDF5 when --enable-hdf5

roystgnr:
See https://github.com/idaholab/moose/issues/20033#issuecomment-1012323604 for the discussion

This still doesn't turn on HDF5 when it's not explicitly enabled, but it makes it possible for us to enable it in configure scripts that will then be compatible with all of the basic "$HDF5_DIR is set" and "PETSc built HDF5" and "HDF5 is already in compiler include+lib paths" build scenarios.

In my test builds this is ... *mostly* enough.  `make check` in contrib/netcdf/v4 runs into [this API incompatibility](https://stackoverflow.com/questions/62157364/why-is-hdf5-giving-a-too-few-arguments-error-here) when I try to use the HDF5 1.12.0 that my petsc 3.15.0 tree downloaded for me - but the netcdf library itself built okay, and the HDF5-dependent bits of libMesh still work fine.  Hopefully this is something that was fixed in HDF5 1.12.1 (what I've been using previously), not something particular to PETSc's HDF5 configuration.

### DIFF
--- a/configure
+++ b/configure
@@ -54563,8 +54563,24 @@ fi
 
 
 # --------------------------------------------------------------
-# HDF5 -- enabled by default
+# HDF5 -- disabled by default
 # --------------------------------------------------------------
+
+# PETSc configure can --download-hdf5, and that's a really convenient
+# feature for users ... but libMesh --enable-hdf5 will only detect it
+# if we add PETSc include and lib directory flags *here*; if we wait
+# until build time it's too late.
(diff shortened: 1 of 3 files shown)


aeslaughter:
@roystgnr @brianmoose What else do you need to support the higher-order element output to VTK format from MOOSE?

PR #3146, opened 2022-01-28
Implement doubly-checked lock for _get_array

lindsayad:
This removes the condition variable and `notify_all` and allows us to
remove the `std::atomic` on `_values_read_only`. This reports
ThreadSanitizer clean and is in the spirit of what @friedmud initially
implemented

### DIFF
--- a/include/numerics/petsc_vector.h
+++ b/include/numerics/petsc_vector.h
@@ -416,19 +416,6 @@ private:
   mutable std::mutex _petsc_vector_mutex;
 
   /**
-   * Condition variable for _get_array and _restore_array.  This is part of the
-   * object to keep down thread contention when reading from multiple
-   * PetscVectors simultaneously
-   */
-  mutable std::condition_variable _petsc_vector_cv;
-
-  /**
-   * Mutex that helps enforce a call_once idiom within a method, where
-   * once means call for only one thread
-   */
-  mutable std::mutex _petsc_vector_do_once_mutex;
-
-  /**
    * Queries the array (and the local form if the vector is ghosted)
    * from PETSc.
    *
@@ -468,7 +455,7 @@ private:
   /**
    * Whether or not the data array is for read only access
    */
-  mutable std::atomic<bool> _values_read_only;
+  mutable bool _values_read_only;
 };
 
 
@@ -1097,11 +1084,6 @@ void PetscVector<T>::get(const std::vector<numeric_index_type> & index,
 
   const std::size_t num = index.size();
 
-  std::unique_lock<std::mutex> read_lock(_petsc_vector_mutex);
-  _petsc_vector_cv.wait(read_lock, [this](){ return _array_is_present.load(); });
-  // When wait exits it means we've acquired and locked the mutex, but all we are doing now
-  // is reading, so it's safe to unlock and let other threads continue
-  read_lock.unlock();
   for (std::size_t i=0; i<num; i++)
     {
       const numeric_index_type local_index = this->map_global_to_local_index(index[i]);
--- a/src/numerics/petsc_vector.C
+++ b/src/numerics/petsc_vector.C
@@ -1151,22 +1151,24 @@ void PetscVector<T>::_get_array(bool read_only) const
 {
   libmesh_assert (this->initialized());
 
+  const bool initially_array_is_present = _array_is_present.load(std::memory_order_acquire);
+
   // If the values have already been retrieved and we're currently
   // trying to get a non-read only view (ie read/write) and the
   // values are currently read only... then we need to restore
   // the array first... and then retrieve it again.
-  if (_array_is_present && !read_only && _values_read_only)
+  if (initially_array_is_present && !read_only && _values_read_only)
     _restore_array();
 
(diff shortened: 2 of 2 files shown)


jwpeterson (approved):
I think this ticks all the boxes: 
:heavy_check_mark:  similar to what we had before in terms of performance
:heavy_check_mark:  simpler/cleaner than what we had before
:heavy_check_mark:  ThreadSanitizer safe
but the `std::memory_order` stuff is very complicated IMO, I don't pretend to understand exactly what is going on here.

lindsayad:
I want to take a little time to understand *why* it's safe to read from `_values_read_only` on thread A towards the top of the method when in my mind a writer thread B may be writing to `_values_read_only` later in the method. E.g. the only way I don't see there being a race is if there's a guarantee that if the `acquire` happens on thread A before the `release` on thread B, then thread A is guaranteed to not see the memory writes of thread B ... but I have only ever seen articles about the reverse, e.g. it's guaranteed that if `release` happens on thread B before `acquire` happens on thread A, then A is guaranteed to see all the writes from thread B

lindsayad:
Or at least `ThreadSanitizer` regards this as safe, but I'm wondering if it's a false negative

lindsayad:
Sigh... the program below does indeed report ThreadSanitizer unclean for read and write to `payload`, so it is as @friedmud said ... ThreadSanitizer can report as clean when there's still a possibility for a race. If you replace the arguments to `fib` with 0 and remove the assert, then the program registers as clean, e.g. false negative.
```c++
#include <thread>
#include <atomic>
#include <mutex>
#include <cassert>

std::mutex cv_m;
std::atomic<bool> done = false;
bool payload = false;

int
fib(int n)
{
  if (n <= 1)
    return n;
  return fib(n - 1) + fib(n - 2);
}

void
foo(const int thread_id)
{
  const bool initially_done = done.load(std::memory_order_acquire);
  assert(!initially_done);

  if (!initially_done)
  {
    // Make thread 1 wait longer, but make both threads wait such that initially_done should be                       
    // false for both                                                                                                 
    if (thread_id)
      fib(42);
    else
      fib(30);

    const auto val = payload;
  }

  if (!initially_done)
  {
    std::lock_guard<std::mutex> lock(cv_m);
    if (!done.load(std::memory_order_relaxed))
    {
      payload = true;
      done.store(true, std::memory_order_release);
    }
  }
}

int
main()
{
  std::thread t0(foo, 0), t1(foo, 1);
  t0.join();
  t1.join();
}
```
So even though this PR has reported as ThreadSanitizer clean in my testing, it should be subject to races for `_values_read_only` whenever `acquire` happens before `release`.

lindsayad:
So basically I can never completely trust thread debugging tools 😭

lindsayad:
So I think that we should actually still be good here because we only read `_values_read_only` if `_array_is_present`, e.g. we should not run into issue if we `acquire` before `release`. If we `acquire` before `release` and `!_array_is_present`, then we don't perform any reads of `_values_read_only` (that would then race with writes to `_values_read_only`). And if we `acquire` before `release` and `_array_is_present`, then we read `_values_read_only` but there are no writers to `_values_read_only` so no race. Our situation is akin to this code
```c++
int
fib(int n)
{
  if (n <= 1)
    return n;
  return fib(n - 1) + fib(n - 2);
}

void
getArray(const int thread_id)
{
  const bool initially_array_is_present = array_is_present.load(std::memory_order_acquire);
  assert(!initially_array_is_present);

  // Make thread 1 wait longer, but make both threads wait such that initially_array_is_present                       
  // should be false for both                                                                                         
  if (thread_id)
    fib(31);
  else
    fib(30);

  if (initially_array_is_present)
    const auto value = values_read_only;

  if (!initially_array_is_present)
  {
    std::lock_guard<std::mutex> lock(cv_m);
    if (!array_is_present.load(std::memory_order_relaxed))
    {
      values_read_only = true;
      array_is_present.store(true, std::memory_order_release);
    }
  }
```
where again I've made sure with the calls to `fib` that there would be a read-write race for `values_read_only` ***if*** the read wasn't protected by `initially_array_is_present`. If you don't protected the read of `values_read_only`, I checked and you get the `ThreadSanitizer` errors. So with this, I'm going to convert this back from draft

roystgnr:
I also like the look of this ... and I'd also really like to see @friedmud's opinion before merging.

Should we create a unit test that constructs a petsc vector, reads an iteration count (or falls back on a tiny default), and just does that many `get_array()` and/or `get_array_read()` calls?  So then we could quickly try out something like `LIBMESH_OPTIONS='--re GetArray --get_array_n 500000 --n_threads 64' time make check -C tests` whenever we're afraid of a possible threaded performance regression?

lindsayad:
I assume you'd want to interweave some operations that involve setting `_array_is_present = false`

roystgnr:
Huh, yeah.  I should have thought a little harder before typing.  I'm actually not sure what kind of access pattern you'd want most, though, to maximize the contention triggered.

friedmud:
I like this pattern - it's much cleaner than before and should still be very performant.  Thanks for doing all the work on this @lindsayad !

PR #3147, opened 2022-01-29
Treat *.C files as C++

mkhon:


roystgnr:
Isn't this an autogenerated (well, auto-copied) file?  When I run `bootstrap`, I see a couple lines like `configure.ac:138: installing 'build-aux/compile'` and the next thing I know any manual changes there are overwritten with a copy of the file from my autotools build, `contrib/autotools/share/automake-1.16/compile`.

Your changes here don't include any changes to the copyright date, so I assume you're not just doing something like pulling a file from 1.16.5 to replace our version from 1.16.  Would that be an option, just upgrading our autotools version?  If not then we're going to have to make the changes via something like a sed command in `bootstrap` or they're just going to get reverted the next time someone so much as touches an m4 file.

mkhon:
libtool compile wrapper does not support .C as C++ extension

PR #3160, opened 2022-02-11
WIP: Add support for Bezier elements in VTK output.

### LINKED ISSUE #3152: VTK support for IsoGeometric Analysis visualization
Pinging @aeslaughter to get the Coreform guys in here.

Right now, our best option for IGA visualization is "output to Exodus, and wince as our ExodusII_IO output inappropriately turns RATIONAL_BERNSTEIN control points into LAGRANGE mid-edge/face/cell points"  This works fine as long as our mesh is so fine that there's little difference between the two ... which means that the more advantageous an IGA solve is, the worse our IGA visualization is going to be.

Our best option for quickly fixing that is probably to output the proper element data via VTKIO, using the relatively recent extensions to VTK that support those element types.

Right now, src/mesh/vtk_io.C has an element map from libMesh element types to VTK element types, and that won't be enough - in libMesh the same e.g. HEX27 may be a Lagrange element if it has one mapping_type() or a Rational Bernstein element if it has another mapping_type().  At the very least we'll need to be testing Elem::mapping_type() before deciding which VTK element type to output.  Hopefully this will be the only major complication (in particular, I don't *see* anything in get_local_node_values or node_values_to_vtk that I'd expect to change), but I'm unfamiliar with the VTK rational basis support so I can't swear to it.

Our best bet for testing this out would be to just add VTK output to at least one of the IGA unit tests in tests/mesh/mesh_input.C - possibly testExodusFileMappingsPlateWithHole?  I'm not sure how best to get such a test in CI (as opposed to "just look at it manually in Paraview"), though, not without VTK input as well as output support.

brianmoose:
VTK 9 added support for VTK_BEZIER_* elements.
This writes out those elements in the VTK output when available.

Start with a WIP to make sure this is how the libmesh folks want this done.

For verification, I just look at the result in paraview  with a tesselation. As @roystgnr said in #3152 , not sure how else to check it.

[PlateWithHole_Patch8_0.vtu.gz](https://github.com/libMesh/libmesh/files/8050965/PlateWithHole_Patch8_0.vtu.gz)

closes #3152

### DIFF
--- a/include/mesh/vtk_io.h
+++ b/include/mesh/vtk_io.h
@@ -209,15 +209,15 @@ private:
   };
 
   /**
-   * ElementMaps object that is built statically and used by
+   * ElementMaps objects that are built statically and used by
    * all instances of this class.
    */
-  static ElementMaps _element_maps;
+  static std::map<ElemMappingType, ElementMaps> _element_maps;
 
   /**
-   * Static function used to construct the _element_maps struct.
+   * Static function used to construct _element_maps.
    */
-  static ElementMaps build_element_maps();
+  static std::map<ElemMappingType, ElementMaps> build_element_maps();
 
 #endif
 };
--- a/src/mesh/vtk_io.C
+++ b/src/mesh/vtk_io.C
@@ -115,14 +115,15 @@ void VTKIO::write (const std::string & name)
 // a couple of "stub" functions at the bottom.
 #ifdef LIBMESH_HAVE_VTK
 
-// Initialize the static _element_maps struct.
-VTKIO::ElementMaps VTKIO::_element_maps = VTKIO::build_element_maps();
+// Initialize the static _element_maps map.
+std::map<ElemMappingType, VTKIO::ElementMaps> VTKIO::_element_maps = VTKIO::build_element_maps();
 
 // Static function which constructs the ElementMaps object.
-VTKIO::ElementMaps VTKIO::build_element_maps()
+std::map<ElemMappingType, VTKIO::ElementMaps> VTKIO::build_element_maps()
 {
   // Object to be filled up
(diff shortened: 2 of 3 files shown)


lindsayad:
@brianmoose what's up?!

jwpeterson (approved):
Do we know what version of Paraview supports (or will support) these types of elements?

brianmoose:
Hey @lindsayad  ! Just started at Coreform about a month ago and now trying to bring @brianmoose back to life!

@jwpeterson Based off the VTK submodule in Paraview, looks like 5.8.0 was the first version that used a VTK that had the VTK_BEZIER_* stuff in it.

jwpeterson:
> looks like 5.8.0 was the first version

OK, thanks I will check that out. I guess this PR imples you didn't run into issues with compiling libmesh against VTK 9, so that's good.

lindsayad:
@jwpeterson not scared of WIP and absence of CI

jwpeterson:
Hah, I clicked the wrong button didn't I.

lindsayad:
> Hey @lindsayad ! Just started at Coreform about a month ago and now trying to bring @brianmoose back to life!

It's fun seeing you back in the community!

brianmoose:
No problems with compiling against VTK 9, that seems to be the default with MOOSE anyway.
I can't believe CIVET is still up and running, @jwpeterson don't deprive me of the full experience :smile:

brianmoose:
Haha, looks like I will get the full experience, I forgot to put a VTK guard in the unit tests...

jwpeterson:
Hah, sorry about that. I guess you an always open a second PR to fix the one I accidentally merged.

PR #3161, opened 2022-02-11
Add a VTK check before trying to write out a VTK file

brianmoose:
Hopefully it is enough to just check if VTK is available and not a specific version.
Presumably, with older versions of VTK this will still provide a sanity check that exodus files with BEX don't crash when writing out to VTK.


brianmoose:
Hopefully fixes #3160

jwpeterson (approved):

PR #3163, opened 2022-02-17
Mesh refinement support in Poly2TriTriangulator

roystgnr:
This isn't quite done yet (it's still not handling the case where refinement entails refining Hole polylines, and we can only refine based on an area criterion, not an angle criterion), but it's 2/3 of the way there now, so probably a good time for review and to make sure CI is as happy with it as my workstation is.


roystgnr:
Adding a couple `--disable-deprecated` CI runs, on a PR that doesn't do anything special there, to try and make sure that the failures in #3164 are really due to the changes there, not something that's regressed in master or in the test recipes.

PR #3164, opened 2022-02-17
Allow proper constness for our MeshBase iterators

roystgnr:
In the middle of working on the triangulation code, I wrote a range for loop over `node_ptr_range()`, and out of habit I declared the loop variable `auto & node`, and then later in my first crack at the algorithm I decided to reuse the variable via `node = next_node` in a while loop ...

And of course this leaks memory and corrupts the mesh.  Fun, "you'll finally trigger an assert four hundred lines of code later" corruption.

With our current mesh iterators, a const MeshBase will give you iterators that allow users to change the underlying elements and nodes, and a non-const MeshBase will give you iterators that allow users to corrupt the Mesh's internal pointers!

This PR fixes that: a const MeshBase will now only give you const access to Elem and Node values, and while a non-const MeshBase will give you non-const access it will no longer let you change the MeshBase internal pointers.

This is a very far-reaching API change (see the fix commits in this PR, just for our internal code), so for now we'll only make the change when configured with `--disable-deprecated`).  Most of the fixes are straightforward and sensible.

There are a couple `const_cast` workarounds here, for code that either needs an API rework or code that's not heavily used enough to be worth more than a hack.

There's one `const_cast` that really worries me, though: I'd really appreciate it if @dknez or @jwpeterson find time to look at the "const_cast hack in mesh stitching", 98f2ea6 - that code has so much test coverage that I can't imagine there's an actual bug, but neither can I figure out how to make it properly const-correct.

### DIFF
--- a/include/base/variant_filter_iterator.h
+++ b/include/base/variant_filter_iterator.h
@@ -46,7 +46,9 @@
  * \author John W. Peterson
  * \date 2004
  */
-template<class Predicate, class Type, class ReferenceType = Type &, class PointerType = Type *>
+template<class Predicate, class Type, class ReferenceType = Type &, class PointerType = Type *,
+                          class ConstType = const Type, class ConstReferenceType = const Type &,
+                          class ConstPointerType = const Type *>
 class variant_filter_iterator :
 #if defined(__GNUC__) && (__GNUC__ < 3)  && !defined(__INTEL_COMPILER)
   public std::forward_iterator<std::forward_iterator_tag, Type>
@@ -58,7 +60,8 @@ public:
   /**
    * Shortcut name for the fully-qualified typename.
    */
-  typedef variant_filter_iterator<Predicate, Type, ReferenceType, PointerType> Iterator;
+  typedef variant_filter_iterator<Predicate, Type, ReferenceType, PointerType,
+                                  ConstType, ConstReferenceType, ConstPointerType> Iterator;
 
 
 
@@ -87,7 +90,7 @@ public:
     virtual bool equal(const IterBase * other) const = 0;
 
     // typedef typename variant_filter_iterator<Predicate, Type, const Type &, const Type *>::IterBase const_IterBase;
-    typedef typename variant_filter_iterator<Predicate, Type const, Type const & , Type const *>::IterBase const_IterBase;
+    typedef typename variant_filter_iterator<Predicate, ConstType, ConstReferenceType, ConstPointerType>::IterBase const_IterBase;
 
     /**
      * Similar to the \p clone() function.
@@ -113,7 +116,7 @@ public:
     virtual bool operator()(const IterBase * in) const = 0;
 
     // typedef typename variant_filter_iterator<Predicate, Type, const Type &, const Type *>::PredBase const_PredBase;
-    typedef typename variant_filter_iterator<Predicate, Type const, Type const &, Type const *>::PredBase const_PredBase;
+    typedef typename variant_filter_iterator<Predicate, ConstType, ConstReferenceType, ConstPointerType>::PredBase const_PredBase;
 
     /**
      * Similar to the \p clone() function.
@@ -186,7 +189,7 @@ public:
        * Important typedef for const_iterators.  Notice the weird syntax!  Does it compile everywhere?
        */
       // typedef typename variant_filter_iterator<Predicate, Type, const Type &, const Type *>::template Iter<IterType> const_Iter;
-      typedef typename variant_filter_iterator<Predicate, Type const, Type const &,  Type const *>::template Iter<IterType> const_Iter;
+      typedef typename variant_filter_iterator<Predicate, ConstType, ConstReferenceType, ConstPointerType>::template Iter<IterType> const_Iter;
 
       typename IterBase::const_IterBase * copy =
         new const_Iter(iter_data);
@@ -284,7 +287,7 @@ public:
        * Important typedef for const_iterators.
        */
       //      typedef typename variant_filter_iterator<Predicate, Type, const Type &, const Type *>::template Pred<IterType, PredType> const_Pred;
-      typedef typename variant_filter_iterator<Predicate, Type const, Type const &,  Type const *>::template Pred<IterType, PredType> const_Pred;
+      typedef typename variant_filter_iterator<Predicate, ConstType, ConstReferenceType, ConstPointerType>::template Pred<IterType, PredType> const_Pred;
 
 
       typename PredBase::const_PredBase * copy =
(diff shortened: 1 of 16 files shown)


jwpeterson (approved):
This looks like a good change. There has always been something a little bit "off" with the `const`ness of the elem/node iterators but I don't think I ever put together that it was related to the "const pointer to const object" issue.

roystgnr:
IIRC I realized what the problem was several years ago, and I tried to fix it, but I couldn't figure out *how* to fix it (the conversion from non-const to const stymied me, because it was no longer as simple as tacking const onto a templated typename?) and I gave up.  This time I still couldn't figure out the best way to fix it, but after having been so horribly embarrassed by my code's run-time breakage, I was open to more extreme "add three more template arguments to `variant_filter_iterator`" level solutions if that was what it took to move that sort of breakage to compile-time.

roystgnr:
Hmmm... I'm seeing a failure in my own testing, when I run `--disable-deprecated` in parallel.  Wonder if I screwed something up in the `mesh_communication.C` update.

jwpeterson:
Regarding this:

> There's one const_cast that really worries me, though: I'd really appreciate it if @dknez or @jwpeterson find time to look at the "const_cast hack in mesh stitching", 98f2ea6 - that code has so much test coverage that I can't imagine there's an actual bug, but neither can I figure out how to make it properly const-correct.

I'll have to take a closer look, because I'm not sure I understand the issue currently. Does `neighbor` have to be non-const because we call a non-const version of `side_ptr`
```
neighbor->side_ptr(their_side, ns);
```
a few lines later?

roystgnr:
Nah, that's actually fine - if `their_side` is a `unique_ptr<const Elem>` then we're allowed to build it from a `const Elem`.

jwpeterson:
> Nah, that's actually fine - if `their_side` is a `unique_ptr<const Elem>` then we're allowed to build it from a `const Elem`.

OK, yeah that's what I figured... so what was the compiler error if you didn't cast `neighbor`?

roystgnr:
I'm an idiot.  I'm testing in parallel on a build that I hadn't pointed to PETSc.  Looks like a false alarm, though I'll triple check before taking off the "do not merge" tag.

roystgnr:
Yeah, IIRC the `set_neighbor()` calls are the problem.

jwpeterson:
OK, certainly it seems like a bug to me. Perhaps it is a line of code that is never actually reached or something.

jwpeterson:
> I was wondering if both non-const and const Elems are getting put into that container, but somehow the key search means we only pull non-const Elems out?

It looks to me like we only insert (key, value) pairs for `this` mesh, note the `(i==0)` bit
```
// This side is on the boundary, add its information to side_to_elem
if (skip_find_neighbors && (i==0))
  {
    key_type key = el->key(side_id);
    val_type val;
    val.first = el;
    val.second = cast_int<unsigned char>(side_id);

    key_val_pair kvp;
    kvp.first = key;
    kvp.second = val;
    side_to_elem_map.insert (kvp);
  }
```
so the `side_to_elem_map` could theoretically contain non-const `Elem*`.

roystgnr:
> note the (i==0) bit

Oh, duh, that's what I missed!  So the map should *only* contain non-const Elems.  Maybe a const_cast (probably in the insert, though rather than the extraction) really is the best answer here.

roystgnr:
I can't actually replicate that unit test failure.  Might need to add some more options to CI, to hopefully get something nicer than a bare segfault to look at.

jwpeterson:
> I can't actually replicate that unit test failure.

OK, I actually build with `--disable-deprecated` locally (and in debug mode) so I could pull your branch and try it as well if you don't have any luck.

roystgnr:
If Debug No Deprecated doesn't replicate the failure, I'd love it if you'd try.  I guess (though I expect @loganharbour might be able to disconfirm) that there could have been something weird with the CI environment changes (Test No Deprecated is an optional test still running on the old environment), but it's also possible that there's some weird Undefined Behavior corner case somewhere where my compiler accidentally does what I'd hoped but other compiler versions don't.

roystgnr:
Well, it's not something regressed in the old environment; I made Debug No Deprecated use the new environment, and it fails too.

And it fails with a segfault too.  And I can't replicate this on my laptop either.

loganharbour:
Test on rod?

roystgnr:
It's definitely a gcc 9 thing, at least.  With gcc-9 (9.4.0) I can replicate the problem on my workstation.  11.2.0 and 10.3.0 both work fine.

roystgnr:
By "gcc-9" thing, apparently I mean just "9 thing".  I get basically the same warning from clang 9, even though (after the fixes in #3166) clang 11 and 13 are happy.

roystgnr:
Interestingly enough, clang 11 and 13 give the same warning when I boil down the code as far as I can, though gcc 10 and 11 continue to not complain.

```
#include <vector>

// -Wreturn-stack-address with clang++-9, clang++-11
// -Wreturn-local-addr with gcc++9
// Works with g++-10, g++-11
typedef const int * const & return_type;

return_type testfunc (const std::vector<int *> vec)
{ 
  return *vec.begin();
}

int main (void)
{ 
  int my_int = 5;
  
  std::vector<int *> vecptrs {&my_int, &my_int, &my_int};
  
  return 0;
}
```

I'm still not sure what's wrong with the above code.  I can get this case to work if I drop the `&` from the return type, which you'd think would be fine in the libMesh use case since we hardly care about passing a const pointer-to-const by reference instead of value ... except we have a bunch of range-for loops that declare `const auto &` as the data type, and if we don't actually return a reference then that breaks all of them.

jwpeterson:
> I'm still not sure what's wrong with the above code.

FWIW, GCC 7.4.0 also warns about this code for me
```
$ g++ --version
g++ (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0
```
```
$ g++ test.C -o test
test.C: In function ‘const int* const& testfunc(std::vector<int*>)’:
test.C:10:21: warning: returning reference to temporary [-Wreturn-local-addr]
   return *vec.begin();
                     ^
```

jwpeterson:
I wonder if these older compilers somehow don't recognize `const int * const &` as a "const reference" and therefore don't think the rule about "const references extending the lifetime of a temporary object" applies, so they instead trigger that `-Wreturn-local-addr` warning.

roystgnr:
I'd feel safer assuming this was an overzealous warning if we didn't also get a segfault out of the same configuration.  I'm checking now to see if I can reproduce that too.

roystgnr:
I do get a segfault in the unit tests, and it does look like it's coming from a corrupted elem iterating over a range.

I also see different assembly from gcc 9 vs gcc 11: https://godbolt.org/z/sPrsj8rT1

roystgnr:
And with both gcc and clang there seems to be a significant change in the assembly from versions 9 to 10.

I just wish I knew why!  If it was a compiler bug before you'd think they'd have backported the fix - gcc 9.4 postdates a bunch of 10.x and 11.x releases that give the new assembly - but if it's a code bug you'd think they wouldn't have dropped the warning, even if the current assembly output turns out to be "luckier" than t used to be.

roystgnr:
No, I don't get *exactly* the same assembly.  And I don't know assembly and don't understand the difference.

roystgnr:
Possibly this has something to do with Defect Report 198??

http://www.open-std.org/jtc1/sc22/wg21/docs/lwg-defects.html#198

permcody:
@loganharbour - asked me to look at why only the GrainTracker tests are failing with this PR when nothing else does. Here's what I found:

```
permcj@theoracle{~/projects/moose}[install_examples]$ find . -name "*.i" | xargs egrep "\[(?:\./)?perf_graph\]"                             (moose)
./modules/phase_field/test/tests/grain_tracker_test/grain_tracker_remapping_test.i:  [./perf_graph]
./modules/phase_field/examples/grain_growth/grain_growth_3D.i:  [./perf_graph]
```

The failing tests are the only ones in the entire repository with a sub_block under `Outputs` named `perf_graph`. I think we support duplicate named nodes in MOOSE, but now I'm not so sure. The reason I say this is that we don't typically look at the parse tree, but instead look at the parameter names like this:

```
Kernels/diffusion/type=Diffusion

Outputs/perf_graph/level=2
Outputs/perf_graph=true
```

The two may coexist without stomping on each other since the structure leading up to any parameter is still unique even if there's a duplicate "path".

The big question is why does this PR trigger this failure? I don't have an explanation for that.

loganharbour:
> The big question is why does this PR trigger this failure? I don't have an explanation for that.

It's not just _this_ PR; we recently upped MOOSE testing in libmesh. This happens everywhere in libmesh at the moment. The question is why it happens on libmesh and not on MOOSE...

PR #3168, opened 2022-02-22
Require C++17 support

roystgnr:
@loganharbour reminds me this morning: our new CI is all set up now; we should be ready to make C++17 support mandatory (and thereby allow us to start writing C++14 and C++17 code...) just as soon as we're brave enough to hit the button here.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -2312,15 +2312,12 @@ std::string DofMap::get_local_constraints(bool print_nonlocal) const
   os << "DoF Constraints:"
      << std::endl;
 
-  for (const auto & pr : _dof_constraints)
+  for (const auto & [i, row] : _dof_constraints)
     {
-      const dof_id_type i = pr.first;
-
       // Skip non-local dofs if requested
       if (!print_nonlocal && !this->local_index(i))
         continue;
 
-      const DofConstraintRow & row = pr.second;
       DofConstraintValueMap::const_iterator rhsit =
         _primal_constraint_values.find(i);
       const Number rhs = (rhsit == _primal_constraint_values.end()) ?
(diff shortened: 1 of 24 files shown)


roystgnr:
Once `Test debug` is finished, I'll tack on another commit that switches a bunch of code to use structured bindings; then we can be more confident that even outside autoconf tests our C++17 really is C++17.

Well, "C++17 without Defect Report fixes", anyway.  Turns out we actually have a ton of cases where we'd love to be able to use a structured binding with private member variables...

roystgnr:
No, never mind.  I misunderstood the DR; that won't affect us yet.

jwpeterson:
This branch compiles fine for me on Ubuntu 18 era GCC (7.4.0). :+1:

jwpeterson (approved):

src/base/dof_map_constraints.C
```diff
@@ -2312,15 +2312,12 @@ std::string DofMap::get_local_constraints(bool print_nonlocal) const
   os << "DoF Constraints:"
      << std::endl;
 
-  for (const auto & pr : _dof_constraints)
+  for (const auto & [i, row] : _dof_constraints)
```
loganharbour, line 2315: ugg this has me so excited

roystgnr:
We've tracked the Mac failure down to an environment problem (turns out having '-std=c++14' in our $CXXFLAGS is a bad idea), might take longer to figure out what's screwing up our Mac conda environment and to fix it.

loganharbour:
Mac tests look good now:

```
configure: Found C++17 standard support
checking whether mpicxx supports C++14 features by default... yes
configure: Found C++14 standard support
checking whether mpicxx supports C++11 features by default... yes
configure: Found C++11 standard support
configure: Using support for C++17 standard
```

If they both go green (ignoring any unrelated moose failures, WIP...) I say this is good to go.

PR #3170, opened 2022-02-22
DistributedMesh fixes for new unit tests, new triangulator

roystgnr:


roystgnr:
At least some of those failures are due to recipe updates, not due to this PR.  I wish I knew the exact mechanism for all of them...

roystgnr:
Going to push on top of master, to see if the recent libMesh fixes plus Moose fixes help some of those CI failures.  A TIMPI update and a serialization upgrade is too significant a change to be 100% certain that those are all false positives and that there's no other lurking failures behind them.

roystgnr:
I think everything in the remaining failing recipes is a known issue with a now-known explanation being dealt with in other PRs, and the passing recipes include Mac and MOOSE debug so I'm happy enough with coverage.  I'll merge this early this afternoon unless there are objections.

PR #3171, opened 2022-02-22
Add static TensorValue::rotation_matrix helper

lindsayad:
I want to use this for coordinate transformation work (idaholab/moose#12293) and I figured I might as well reuse what we already have

### DIFF
--- a/include/numerics/tensor_value.h
+++ b/include/numerics/tensor_value.h
@@ -298,6 +298,8 @@ TensorValue<Real>
 TensorValue<T>::rotation_matrix(const Real phi, const Real theta, const Real psi)
 {
 #if LIBMESH_DIM == 3
+  // We apply a negative sign here or else we don't get counter-clockwise/right-hand-rule rotation.
+  // Maybe there's an error/omission in https://mathworld.wolfram.com/EulerAngles.html ?
   const Real p = -phi / 180. * pi;
   const Real t = -theta / 180. * pi;
   const Real s = -psi / 180. * pi;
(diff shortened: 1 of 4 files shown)


include/numerics/tensor_value.h
```diff
@@ -275,6 +287,35 @@ TensorValue<T>::TensorValue (const TypeTensor<Real> & p_re,
 }
 #endif
 
+template <typename T>
+TensorValue<T>
+TensorValue<T>::rotation_matrix(const Real phi, const Real theta, const Real psi)
+{
+#if LIBMESH_DIM == 3
+  const Real p = -phi / 180. * libMesh::pi;
```
lindsayad, line 295:
```suggestion
  const Real p = -phi / 180. * pi;
```
Hmm I set about writing the inverse rotation and now am in this quandary as to why the normal rotation has negative signs applied here

lindsayad: @dknez I think you were the one who originally added the code in `MeshModification::rotate` (way back in 2011 it looks like). I've checked and these negative signs are indeed necessary in order to get counter-clockwise/right-hand-rule rotation. I feel like this means there is an error/omission in the [Wolfram documentation](https://mathworld.wolfram.com/EulerAngles.html) or maybe I am missing a detail in the documentation?

dknez: hmm, good question... I don't remember the details here, sorry!

jwpeterson: Looks like code was originally added in 476198bb789923932767df4fd5884dc759b2113a by @benkirk and possibly @MartinLuethi was also involved. I concur that the Wolfram article seems to be wrong, and we have been using a corrected version in our code all these years. A better reference might be the [Wikipedia](https://en.wikipedia.org/wiki/Euler_angles#Rotation_matrix) article, specifically the last row of the table in the "Proper Euler angles" column.

jwpeterson (approved):

lindsayad:
MOOSE failures are unrelated and will be fixed once https://github.com/idaholab/moose/pull/20384 percolates into master

lindsayad:
wtf happened to libmesh testing? We must have changed the CIVET environment?

include/numerics/tensor_value.h
```diff
@@ -275,6 +295,43 @@ TensorValue<T>::TensorValue (const TypeTensor<Real> & p_re,
...
 #endif
 
+template <typename T>
+TensorValue<Real>
+TensorValue<T>::rotation_matrix(const Real angle1_deg, const Real angle2_deg, const Real angle3_deg)
+{
+#if LIBMESH_DIM == 3
+  const auto angle1 = angle1_deg / 180. * pi;
+  const auto angle2 = angle2_deg / 180. * pi;
+  const auto angle3 = angle3_deg / 180. * pi;
```
loganharbour, line 305:
```suggestion
  const auto angle1 = -angle1_deg / 180. * pi;
  const auto angle2 = -angle2_deg / 180. * pi;
  const auto angle3 = -angle3_deg / 180. * pi;
```

loganharbour (changes requested):

lindsayad:
@jwpeterson it looks to me like we've already got a lot of users relying on our (as-yet) undocumented reverse intrinsic rotation methodology. Yuck. What do you think about having (reverse-)intrinsic and extrinsic methods? `MeshModification::rotate` would use (reverse-)intrinsic

jwpeterson:
> What do you think about having (reverse-)intrinsic and extrinsic methods?
I think this would be even more confusing than what we have now (which is saying something!).

I'm confused though, I thought what you implemented would give the same results as before, but we would no longer have this weird internal negation of angles to document, and we would be able to use an as-is formula from Wikipedia... Guess I'm wrong about that though.

lindsayad:
I think the animation in [this section](https://en.wikipedia.org/wiki/Euler_angles#Chained_rotations_equivalence) shows the intrinsic rotations best. After watching that I think it's fairly clear that you would have to prescribe different angles in an extrinsic rotation in order to reach the same final end point.

lindsayad:
I think I can truthfully say this time that remaining failures are unrelated

roystgnr (approved):

PR #3176, opened 2022-02-24
Lock against concurrent read-write/write-write in Vector::add/insert/set

lindsayad:
A couple other minor changes
- Use `Threads::spin_mutex` instead of `std::mutex` to help address
  @dschwen's concerns about being thread-model agnostic
- Similarly `Threads::spin_mutex::scoped_lock` in place of
  `std::lock_guard<std::mutex>`
- In `_restore_array` use the same pattern as `_get_array`. We were
  actually using memory sequential ordering here before. Going to `release`
  and `acquire` is actually less heavy-weight

The changes here fix the thread errors on the failing moose test
`geomsearch/3d_moving_penetration_smoothing.disjoint/pl_test3nnstt` when
run with threads on https://civet.inl.gov/job/963721/ for #3170. I've also
seen that thread failure crop on another libmesh PR recently (I think it
was my own). That test currently has thread errors because the racing adds
are protected by translation unit local mutexes ... and the mutexes are in
different translation units (Postprocessor unity build vs. UserObject unity build). 
This is a demonstration that users cannot really
be trusted to write thread-safe code, or that they can get races from places
they wouldn't necessarily expect to compete

### DIFF
--- a/contrib/bin/test_installed_examples.sh
+++ b/contrib/bin/test_installed_examples.sh
@@ -63,7 +63,7 @@ function test_example()
     printf '%s' "Testing example installation $exdir ... " > $stdout
     cd $examples_install_path/$exdir
 
-    if $CXX $installed_CXXFLAGS *.C -o $app_to_link $installed_LIBS > $errlog 2>&1 ; then
+    if $CXX $installed_CXXFLAGS $CXXFLAGS *.C -o $app_to_link $installed_LIBS > $errlog 2>&1 ; then
         # See color codes above.  We:
         # .) skip to column 65
         # .) print [ in white
@@ -76,7 +76,7 @@ function test_example()
         # See comment above for OK status
         printf '\e[65G\e[1;37m[\e[1;31m%s\e[1;37m]\e[m\e[m\n' " FAILED " >> $stdout
 	echo "Compile line:"  >> $stdout
-	echo $CXX $installed_CXXFLAGS *.C -o $app_to_link $installed_LIBS >> $stdout
+	echo $CXX $installed_CXXFLAGS $CXXFLAGS *.C -o $app_to_link $installed_LIBS >> $stdout
 	echo ""  >> $stdout
 	echo "Output:"  >> $stdout
 	cat $errlog  >> $stdout
--- a/contrib/bin/test_installed_headers.sh
+++ b/contrib/bin/test_installed_headers.sh
@@ -80,7 +80,7 @@ function test_header()
     echo "int foo () { return 0; }" >> $source_file
 
     #echo $CXX $test_CXXFLAGS $source_file -o $app_file
-    if $CXX $test_CXXFLAGS $source_file -c -o $object_file >$errlog 2>&1 ; then
+    if $CXX $test_CXXFLAGS $CXXFLAGS $source_file -c -o $object_file >$errlog 2>&1 ; then
         # See color codes above.  We:
         # .) skip to column 65
         # .) print [ in white
@@ -96,7 +96,7 @@ function test_header()
         cat $source_file  >> $stdout
         echo ""  >> $stdout
         echo "Command line:" >> $stdout
-        echo $CXX $test_CXXFLAGS $source_file -c -o $object_file  >> $stdout
+        echo $CXX $test_CXXFLAGS $CXXFLAGS $source_file -c -o $object_file  >> $stdout
         echo ""  >> $stdout
         echo "Output:" >> $stdout
         cat $errlog >> $stdout
--- a/include/numerics/distributed_vector.h
+++ b/include/numerics/distributed_vector.h
@@ -33,6 +33,7 @@
 #include <vector>
 #include <algorithm>
 #include <limits>
+#include <mutex>
 
 namespace libMesh
 {
@@ -576,6 +577,7 @@ void DistributedVector<T>::set (const numeric_index_type i, const T value)
   libmesh_assert_less (i, size());
   libmesh_assert_less (i-first_local_index(), local_size());
 
+  std::scoped_lock lock(this->_numeric_vector_mutex);
   _values[i - _first_local_index] = value;
 
 
@@ -595,6 +597,7 @@ void DistributedVector<T>::add (const numeric_index_type i, const T value)
   libmesh_assert_less (i, size());
   libmesh_assert_less (i-first_local_index(), local_size());
 
+  std::scoped_lock lock(this->_numeric_vector_mutex);
   _values[i - _first_local_index] += value;
 
 
--- a/include/numerics/eigen_sparse_vector.h
+++ b/include/numerics/eigen_sparse_vector.h
@@ -33,6 +33,7 @@
 
 // C++ includes
 #include <limits>
+#include <mutex>
 
 namespace libMesh
 {
@@ -481,6 +482,7 @@ void EigenSparseVector<T>::set (const numeric_index_type i, const T value)
   libmesh_assert (this->initialized());
   libmesh_assert_less (i, this->size());
 
+  std::scoped_lock lock(this->_numeric_vector_mutex);
   _vec[static_cast<eigen_idx_type>(i)] = value;
 
   this->_is_closed = false;
@@ -495,6 +497,7 @@ void EigenSparseVector<T>::add (const numeric_index_type i, const T value)
   libmesh_assert (this->initialized());
   libmesh_assert_less (i, this->size());
 
+  std::scoped_lock lock(this->_numeric_vector_mutex);
   _vec[static_cast<eigen_idx_type>(i)] += value;
 
   this->_is_closed = false;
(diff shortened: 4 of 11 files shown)


dschwen:
I'm starting to think threads are threads and they are whatever the operating system gives us. Std::thread, pthreads, openmp are just different interfaces. But the basic thread operations should work interchangeably.

roystgnr:
If we're doing this in PetscVector, shouldn't we be doing the same in every NumericVector subclass?  If we're doing this in localized add() operations, shouldn't we be doing it in set() and insert()?  And shouldn't we document it?  With our previous thread-safety policy of "caveat emptor" at least users could figure out that they had to protect themselves; "you don't have to protect yourself, but ha ha there's exceptions" is in a sense more dangerous.

And I want to ping @jwpeterson and see if he knows any more than I do about "why weren't we doing this in PetscVector already".  The only guesses I could come up with were "maybe an advanced user wants to batch multiple add calls inside a single lock" and "maybe a crazy user wants to create per-thread NumericVectors.

lindsayad:
> I'm starting to think threads are threads and they are whatever the operating system gives us. Std::thread, pthreads, openmp are just different interfaces.

So what interfaces do you think we should use in our code? My preference would be to go through `std` since it feels the most ... standard

lindsayad:
> If we're doing this in PetscVector, shouldn't we be doing the same in every NumericVector subclass? If we're doing this in localized add() operations, shouldn't we be doing it in set() and insert()? And shouldn't we document it?

I think the answer to all of your questions is yes. But I'll wait on @jwpeterson's take. A comment I would make is that in general with our threaded loops it is rare for us to be adding/inserting into the same location in the vector; this would occur most commonly probably with a method like DG. Usually the data races I see in TSAN for `add` are actually for `_is_closed` (well actually all the races I've seen to this point are for `_is_closed`; none for the vector memory locations themselves)

lindsayad:
So after writing that ... maybe it does make the most sense not to put locks in? Maybe instead we could just make `_is_closed` `std::atomic` and then if a user is doing DG then they should be handling locking?

lindsayad:
I'm being dumb ... even CFEM should in theory have data races for vector entries for nodes that are on thread boundaries

roystgnr:
Long ago I looked at node "coloring schemes" for this sort of thing, where you could work independently in most cases because you knew a priori that you weren't going to conflict with a neighboring thread because the neighboring thread was only working a little ways away ...

That's hard and fragile code to write, though.

jwpeterson:
> ping @jwpeterson and see if he knows any more than I do about "why weren't we doing this in PetscVector already". 

I think we only fixed the cases which were giving us issues at the time, i.e. multhreaded assembly that calls `PetscVector::add` and apparently `PetscMatrix<T>::get_row()` which I think is something MOOSE uses to set DirichletBCs. So yeah, ideally everything you mentioned (and more) should be guarded with mutexes.

lindsayad:
@roystgnr what kind of documentation are you looking for? I was thinking I'd just add something like this to method doxygen: "This method is thread-safe because we lock our mutex before writing into the global vector". That work?

lindsayad:
@friedmud might have opinions on the proposed changes here

lindsayad:
More thread failures at https://civet.inl.gov/job/965795/ I suspect the additional two failing tests are due to the same cause

friedmud:
We should have a meeting to talk about this.

There is no need for this _in MOOSE_ because it's taken care of in `Assembly`.

As for `std::thread`... there is nothing more agnostic than `std::thread`... and yes, they all use the same OS threads underneath (usually they compile to pthreads).

friedmud:
Let me say a bit more: we don't want locking in these routines because it's hella slow for everyone not doing threading... and there are better ways to deal with this within your application that are more performant.

friedmud:
@jwpeterson we never call those methods within threaded sections...

lindsayad:
Here are a couple of profiles running a 4 million dof simple diffusion problem with 64 threads (single process). The first is with libmesh master and moose master. The second is with this libmesh branch and then MOOSE edited to remove the lock in `ComputeResidualThread::postElement` since it would be redundant and unnecessary with the lock introduced in this PR. The latter case drops the time spent spinning from 15% of the sim time to 11% of the sim time. I think maybe what @friedmud was saying is that we shouldn't have these calls to `add` within the threaded loop and hence no need for the locks, e.g. we should only be calling `add` post thread loop execution? If that is the case, then I myself wouldn't be too worried about the expense of a lock that won't ever have contention, but I could be wrong.

Anyway I thought this data could be helpful when we have a discussion about this.

### libMesh/MOOSE master

`peek spin_mutex::scoped_lock::scoped_lock` output:
```
                                             9.71s 67.01% |   ComputeJacobianThread::postElement (inline)
                                             4.71s 32.51% |   ComputeResidualThread::postElement (inline)
                                             0.07s  0.48% |   libMesh::Threads::atomic::operator++ (inline)
         0     0%     0%     14.49s 15.17%                | libMesh::Threads::spin_mutex::scoped_lock::scoped_lock
                                            14.49s   100% |   libMesh::Threads::spin_mutex::lock (inline)
```

simple-diffusion-lm-master

### This PR + no lock in ComputeResidualThread::postElement

`peek spin_mutex::scoped_lock::scoped_lock` output:
```
                                            10.01s 97.28% |   ComputeJacobianThread::postElement (inline)
                                             0.12s  1.17% |   libMesh::PetscVector::add (inline)
                                             0.11s  1.07% |   libMesh::PetscVector::add_vector (inline)
                                             0.05s  0.49% |   libMesh::Threads::atomic::operator++ (inline)
         0     0% 0.022%     10.29s 11.26%                | libMesh::Threads::spin_mutex::scoped_lock::scoped_lock
                                            10.29s   100% |   libMesh::Threads::spin_mutex::lock (inline)
```

simple-diffusion-lm-alex-thread-branch-remove-moose-lock

friedmud:
To be clear - leave the lock in `postElement` and make sure there are none in PetscVector (basically: like it used to be).

lindsayad:
The first graph is for what it used to be/MOOSE and libMesh master

friedmud:
That doesn't make sense.  You're locking a lot more now - and it's showing less lock time?  I guess we'll sort it out in the meeting.

Also: any chance you could try running both cases with threads to report the time?  8 or 12 threads would be good.

friedmud:
We would also need to see timings of running a pure libMesh calculation with this change.  Pure libMesh calcs never had locking before - now there will be locking for every add...

lindsayad:
> You're locking a lot more now - and it's showing less lock time? I guess we'll sort it out in the meeting.

You're locking the same amount. It's just now you're locking in libMesh instead of MOOSE, and we're only locking before shared memory access wheres when you lock in MOOSE you're locking out some operations that could be happening in parallel.

> Also: any chance you could try running both cases with threads to report the time? 8 or 12 threads would be good.

Both the profiles were generated with 64 threads. You're looking for timings without tcmalloc?

friedmud:
Ugh - sorry - my reading comprehension is definitely failing me.  Looking forward to hearing more about it all in the meeting.

lindsayad:
Ok benchmark timing rebased on top of #3178
```
time make -j 64 check -C examples METHODS=opt LIBMESH_BENCHMARK=1 SUBDIRS='adjoints/adjoints_ex1 adjoints/adjoints_ex2 adjoints/adjoints_ex3 adjoints/adjoints_ex4 adjoints/adjoints_ex6 adjoints/adjoints_ex7 fem_system/fem_system_ex1 fem_system/fem_system_ex2 fem_system/fem_system_ex3 fem_system/fem_system_ex4 fem_system/fem_system_ex5'
```

- master, single thread: 11m3.346s
- master, 64 threads: 8m30.149s
- this, single thread: 11m10.524s
- this, 64 threads: 8m28.332s

friedmud:
Well - it seems like we have plenty of evidence that this change won't make a huge difference in run-time (and we can bypass it if we need to).

friedmud:
Thanks for running the numbers @lindsayad !

lindsayad:
So I think the question is do we want to preemptively add `unlocked_*` APIs before there's evidence that they're needed? Unfortunately it's not as simple as having the locked version locking and then calling the unlocked version because `_restore_array` may try to lock the same `_petsc_vector_mutex`. Unless we wanted to add another mutex and make calls to `_restore_array` full contentions, reducing code duplication would mean calling `_restore_array` in both locked and unlocked versions, and then (after locking in the locked version) calling a helper function. So at that point we would have tripled the function number for add/set/insert. Maybe given the limited number of function we're tripling, that's ok?

roystgnr:
Agreed.  I wasn't expecting a huge cost, and 0.5% is well within the amount of variation we normally see from run to run.

roystgnr (approved):

friedmud (approved):
This looks good to me. :+1:

PR #3179, opened 2022-03-01
Make adjoint ex5 benchmark pass

lindsayad:
Also respect `METHODS` in `benchmark_example()`

Fixes #3177

### DIFF
--- a/examples/adjoints/adjoints_ex5/adjoints_ex5.C
+++ b/examples/adjoints/adjoints_ex5/adjoints_ex5.C
@@ -725,7 +725,7 @@ int main (int argc, char ** argv)
         libmesh_error_msg_if(std::abs(system.time - (1.0089)) >= 2.e-4,
                              "Mismatch in end time reached by adaptive timestepper!");
 
-        libmesh_error_msg_if(std::abs(total_sensitivity - 4.87767) >= 2.e-4,
+        libmesh_error_msg_if(std::abs(total_sensitivity - 4.87767) >= 3.e-3,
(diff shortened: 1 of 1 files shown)


roystgnr:
I'm cool with loosening the tolerance here (though I really also need to figure out why a failure there wasn't causing "make check" to fail...)

But what's going on with METHODS?  When would you ever want to benchmark with more than one $METHOD?

lindsayad:
I dunno. I've been running with METHODS unthinkingly after you passed me the command in slack
```
make -j 65 check -C examples METHODS=opt LIBMESH_BENCHMARK=1 SUBDIRS='adjoints/adjoints_ex1 adjoints/adjoints_ex2 adjoints/adjoints_ex3 adjoints/adjoints_ex4 adjoints/adjoints_ex5 adjoints/adjoints_ex6 adjoints/adjoints_ex7 fem_system/fem_system_ex1 fem_system/fem_system_ex2 fem_system/fem_system_ex3 fem_system/fem_system_ex4 fem_system/fem_system_ex5'
```
And I had just noticed that METHODS wasn't being recognized/respected. I see what you're saying. I can remove that commit

roystgnr:
Ah, damn it.  I'm sure I'd meant to write `METHOD=opt`.  Yeah, let's take out that commit.  Maybe we should broaden the function here to accept METHOD or METHODS-with-only-one-entry at some point, but definitely don't want to benchmark with 2.

roystgnr (approved):

PR #2931, opened 2021-05-24
libmesh_not_implemented() in all unimplemented TypeNTensor functions

jwpeterson:
The constructors for this class are (sort of) implemented, but almost none of the actual API is... to prevent the possibility of any of the unimplemented functions being called and silently returning invalid results, they should all have `libmesh_not_implemented()` calls in them.

PR #3180, opened 2022-03-02
memcpy warning fix in TypeNTensor

roystgnr:
I'm not sure I understand why gcc 11 thinks the auto-generated functions
here might lead to memcpy-into-buffer-of-size-0, but this is a
compilation shim we should never be truly using anyway, so I won't
bother figuring out the details until we actually need rank-3+ tensors,
I'll just work around the problem.


roystgnr:
So we're actually invoking `TypeNTensor::operator=`, even?!  I guess I really should have looked harder at the failures in #2931 last year.  I'll look into this now.

PR #3184, opened 2022-03-07
Optionally use system provided netcdf

### LINKED ISSUE #1542: Please allow external netcdf
I already have ```netcdf``` installed. ```--enable-netcdf``` attempts to install ```netcdf``` over the existing installation and fails.

You should have the argument ```--with-netcdf=/usr/local``` to set external netcdf.
Currently I just can't use ```netcdf``` with ```libmesh``` for this reason.

StefanBruens:
Detect a system wide installed netcdf using pkgconfig and use it, in
case --enable-netcdf=system is specified.

Fixes #1542.

### DIFF
--- a/m4/netcdf.m4
+++ b/m4/netcdf.m4
@@ -7,6 +7,8 @@ AC_DEFUN([CONFIGURE_NETCDF],
                 AS_HELP_STRING([--disable-netcdf],
                                [build without netCDF binary I/O]),
                 [AS_CASE("${enableval}",
+                  [system],     [enablenetcdf=yes
+                                 netcdfversion=system],
                   [yes|new|v4], [enablenetcdf=yes
                                  netcdfversion=4],
                   [old|v3],     [enablenetcdf=yes
@@ -27,6 +29,8 @@ AC_DEFUN([CONFIGURE_NETCDF],
         ])
 
   AS_CASE("${netcdfversion}",
+          [system], [PKG_CHECK_MODULES([NETCDF], [netcdf >= 4.2])],
+
           [3], [
                  dnl We shouldn't get here, see if test above.
                  AC_MSG_ERROR([>>> Error: netCDF3 is no longer distributed with libMesh <<<])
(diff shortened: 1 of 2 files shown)


m4/netcdf.m4
```diff
@@ -27,6 +29,8 @@ AC_DEFUN([CONFIGURE_NETCDF],
         ])
 
   AS_CASE("${netcdfversion}",
+          [system], [PKG_CHECK_MODULES([NETCDF], [netcdf >= 4.2])],
```
roystgnr, line 32: Do we need to make a local copy of pkg.m4 before we try to use PKG_CHECK_MODULES?  My system has it installed, but we don't want bootstrap to rely on anything more than our own contrib/autotools version.

StefanBruens: It is part of pkgconf, so it should be there already.

From several years of packaging experience, bundled copies only get out of sync with upstream and are a reoccurring PITA.

roystgnr: > It is part of pkgconf, so it should be there already.

Sadly I have committed the ought-is fallacy and been burned by it too many times to not check, and ... it's almost guaranteed to be available on Ubuntu (It's a dependency for *ubuntu-desktop* now, of all things?  The default GUI now depends on autoconf macros??), but the next system I checked was an OSX laptop, where although everything I need to build (and even bootstrap) libMesh is there, but pkg.m4 isn't.

> bundled copies only get out of sync with upstream and are a reoccurring PITA.

I can't disagree with this.  But adding dependencies willy-nilly has gotten me in trouble too.  What does pkg.m4 itself depend on?  If it's self-contained (and looks like it is, in a couple hundred lines?) I'd feel safer working with a copy of it than letting people without it break.

lindsayad:
Per #3390 we will likely want to do some comparison of netcdf and hdf5 versions so that we can error out during configure if incompatible versions are detected ... although admittedly the incompatibility of netcdf 4.6.2 with hdf5 1.12 appears to only be in netcdf testing of itself and not in our uses of it

lindsayad:
we really should try to make some progress on this

PR #3185, opened 2022-03-08
Factor out push_node, respect offset

roystgnr:
This saves a couple lines of code, and will save more when I use the
lambda again in some subsequent changes ... but more importantly, it
makes the coordinate offset parameter get respected by discontinuous
output, not just continuous output.

Let's get that into CI before I go any further, and see if there are any
Moose tests that are comparing to inconsistent gold results.


roystgnr:
John - sorry to pester you about what ought to be a simple fix, but it occurs to me that Akselos might be using this feature too.

jwpeterson (approved):
Ah, so the issue here was that we weren't using `_coordinate_offset` in one place but were in another? This is not something that we use, so I'm fine with merging as-is :+1:

PR #3188, opened 2022-03-10
ExodusII::write_added_sides()

roystgnr:
We're going to want to let users visualize SIDE_DISCONTINUOUS variables, and that's a little tricky to do when we're outputting continuous meshes.  A workaround: output side elements for all sides of the mesh, so we can then output discontinuous side data on those.

This is the first half of that: output of the side elements, and a unit test for the same.  This probably isn't worth merging until the data output is also done, but I want to make sure our CI tests aren't unhappy with this.

### DIFF
--- a/include/mesh/exodusII_io.h
+++ b/include/mesh/exodusII_io.h
@@ -140,6 +140,8 @@ public:
    */
   void write_added_sides (bool val);
 
+  virtual bool get_add_sides () override;
+
   /**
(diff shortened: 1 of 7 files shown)


roystgnr:
Well, I can replicate the failure, at least.

roystgnr:
I'm expecting that last commit to fail; I'll reorder it later once the code to support it is there.

roystgnr:
Replaced by #3233

PR #3191, opened 2022-03-16
RB Update: Enable RBParametrizedFunction to be defined based on external index maps

dknez:
This PR doesn't change any existing behavior or APIs, it just adds virtual methods that users can use in subclasses to define RBParametrizedFunctions (used in RB EIM approximations) based on maps into data that are indexed on mesh-based data like elem IDs or node IDs.


dknez:
@roystgnr Do you know if this is a real failure? I would not expect any issues with this MR since it does not change any existing behavior.

dknez:
The error message I see is "FAIL make_install/moose_test_tests/version.print_version_short FAILED (EXPECTED OUTPUT MISSING)", but I'm not clear if that is related to this PR or not?

roystgnr:
Definitely not related to this PR; that's a new Moose test that someone had already noticed had problems in another context.

"Coverage, New: 0.00%" is a little more worrisome, though.  Is that just because this is all in a CapNProto code path and we don't have that enabled in CI anywhere?

dknez:
@roystgnr it's working on our internal tests, and I think doesn't change any existing behavior. Are you OK with merging this?

@jwpeterson is on vacation this week, so he can't weigh in right now.

dknez:
Thanks!

PR #3200, opened 2022-03-22
Add MeshBase::all_second_order_range()

jwpeterson:
This generalizes the behavior of `MeshBase::all_second_order()` by allowing the user to pass in an iterator range of elements to be converted to `SECOND` order. The original `MeshBase::all_second_order()` function is now a shim that calls this function. 

Note: I considered the possibility of giving the new function the same name but that would have made it an overloaded virtual, which seems to always give warnings about "hidden" functions, so I decided to just avoid this by giving it a new name.

Refs #000

### DIFF
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -37,7 +37,7 @@
 #include <sstream>
 #include <iomanip>
 #include <unordered_map>
-
+#include <algorithm> // std::all_of
 
 namespace {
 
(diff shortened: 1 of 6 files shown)


src/mesh/unstructured_mesh.C
```diff
@@ -1244,17 +1246,19 @@ void UnstructuredMesh::all_second_order (const bool full_ordered)
...
+  // empty range we therefore we set already_second_order to false so
+  // that processors with empty ranges just "do whatever the other
+  // processors are doing".
+  bool already_second_order =
+    (range.begin() != range.end()) &&
+    std::all_of(range.begin(), range.end(),
+                [](const auto & elem){ return elem->default_order() == SECOND; });
+
+  // Check with other processors and possibly return early
   this->comm().max(already_second_order);
```
roystgnr, line 1262: We want `min()` now instead of `max()`, right?  If one processor has a range with all first-order elements that doesn't mean every processor does.

Also, if we use `min()` we can probably get rid of the separate empty-range check above - `std::all_of` should return `true` for an empty range.

jwpeterson: If any processor has a range with (some) first-order elements, then they can't all early return. This is a parallel-only function, or at least that was how I understood the original logic.

roystgnr:
That one problem was the only thing I noticed, we should be good to merge once it's fixed, but I'm throwing on a `--enable-distmesh` processor sweep to be paranoid.

You also might rebase on master and add `LOG_UNIT_TEST;` calls?  (for completeness, if not immediate usefulness - the Valgrind results on that logging are in and it's pretty clear that the FETest suite-of-suites is what's killing us there)

src/mesh/unstructured_mesh.C
roystgnr: The original logic wasn't logical.  I was assuming that all elements in the mesh were of the same order.

jwpeterson: This issue should be fixed now and I've also rebased on latest devel.

jwpeterson:
The Min clang and Min gcc checks appear to be done and green... but I guess that update never made it to GitHub for some reason. @roystgnr I'm going to go ahead and merge unless you want to investigate further.

roystgnr:
Github seems to be having intermittent server issues all the time lately.  Yeah, go ahead.

PR #3204, opened 2022-03-23
Use string_view, std::move(string) where appropriate.

roystgnr:
Definitions of "appropriate" may vary.  In the one direction: I didn't get anywhere close to the "`const std::string &` parameters should all either be `string_view` or pass-by-value-then-move instead" advice.  In the other: we could scale down this PR or ditch it entirely; nothing we ever do with text is performance-critical (except PerfLog, ironically, and we fixed that quite a while ago).

### DIFF
--- a/include/base/variable.h
+++ b/include/base/variable.h
@@ -56,12 +56,12 @@ public:
    * all subdomains.
    */
   Variable (System * sys,
-            const std::string & var_name,
+            std::string var_name,
             const unsigned int var_number,
             const unsigned int first_scalar_num,
             const FEType & var_type) :
     _sys(sys),
-    _name(var_name),
+    _name(std::move(var_name)),
     _active_subdomains(),
     _number(var_number),
     _first_scalar_number(first_scalar_num),
@@ -73,13 +73,13 @@ public:
    * indices for which this variable is active.
    */
   Variable (System * sys,
-            const std::string & var_name,
+            std::string var_name,
             const unsigned int var_number,
             const unsigned int first_scalar_num,
             const FEType & var_type,
             const std::set<subdomain_id_type> & var_active_subdomains) :
     _sys(sys),
-    _name(var_name),
+    _name(std::move(var_name)),
     _active_subdomains(var_active_subdomains),
     _number(var_number),
     _first_scalar_number(first_scalar_num),
@@ -190,7 +190,7 @@ public:
    * all subdomains.
    */
   VariableGroup (System * sys,
-                 const std::vector<std::string> & var_names,
+                 std::vector<std::string> var_names,
                  const unsigned int var_number,
                  const unsigned int first_scalar_num,
                  const FEType & var_type) :
@@ -199,7 +199,7 @@ public:
               var_number,
               first_scalar_num,
               var_type),
-    _names(var_names)
+    _names(std::move(var_names))
   {}
 
 
@@ -208,7 +208,7 @@ public:
    * indices for which this variable is active.
    */
   VariableGroup (System * sys,
-                 const std::vector<std::string> & var_names,
+                 std::vector<std::string> var_names,
                  const unsigned int var_number,
                  const unsigned int first_scalar_num,
                  const FEType & var_type,
@@ -220,7 +220,7 @@ public:
               first_scalar_num,
               var_type,
               var_active_subdomains),
-    _names(var_names)
+    _names(std::move(var_names))
   {}
 
   /**
--- a/include/base/variable.h
+++ b/include/base/variable.h
@@ -297,8 +297,8 @@ public:
    * a very limited window of opportunity - after the user specifies variables
    * but before the system is initialized.
    */
-  void append (const std::string & var_name)
-  { _names.push_back (var_name); }
+  void append (std::string var_name)
+  { _names.push_back (std::move(var_name)); }
 
 protected:
   std::vector<std::string> _names;
--- a/include/error_estimation/exact_solution.h
+++ b/include/error_estimation/exact_solution.h
@@ -197,8 +197,8 @@ public:
    * any value.  For that you need to call the l2_error(), h1_error()
    * or h2_error() functions respectively.
    */
-  void compute_error(const std::string & sys_name,
-                     const std::string & unknown_name);
+  void compute_error(std::string_view sys_name,
+                     std::string_view unknown_name);
 
   /**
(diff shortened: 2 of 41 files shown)


roystgnr:
That error was interesting.  MooseEnum has an implicit conversion operator giving std::string, which was enough to pass a MooseEnum to `string_to_enum(string)`.  string_view has an implicit constructor from std::string, which is enough to let us pass a string to `string_to_enum(string_view)`.  But C++ doesn't like chaining both together:

```
struct Stringable {
  operator std::string () const { return "Converted!"; }
};

void takes_string_view(std::string_view) {}
void takes_cr_string(const std::string &) {}

void test(void)
{
  Stringable cs;
  std::string s;

  takes_cr_string(s);
  takes_cr_string(cs);
  takes_string_view(s);
  // takes_string_view(cs); // fails
}
```

If we try to make a `string_view` API but keep the `const string &` version around for backwards compatibility with this case then we also need to *add* a `const char *` version around to keep the other two from being ambiguous overloads.  So I've just reverted `string_to_enum` entirely; I don't see a better solution.

roystgnr:
And that error was less interesting.  I was hunting for some crazy string_view subtlety that I'd missed; turns out I'd just screwed up the formatting when replacing sprintf calls.

include/error_estimation/exact_solution.h
```diff
@@ -342,15 +342,16 @@ class ExactSolution
    * The name of the unknown is
    * the key for the map.
    */
-  typedef std::map<std::string, std::vector<Real>> SystemErrorMap;
+  typedef std::map<std::string, std::vector<Real>, std::less<>>
+    SystemErrorMap;
```
jwpeterson, line 346: Is there a reason to explicitly specify the map comparison function here? Is this somehow not equivalent to`std::less<std::string>` (which would be the default)?

include/error_estimation/exact_solution.h
```diff
@@ -342,15 +342,16 @@ class ExactSolution
...
 
   /**
    * A map of SystemErrorMaps, which contains entries
    * for each system in the EquationSystems object.
    * This is required, since it is possible for two
    * systems to have unknowns with the *same name*.
    */
-  std::map<std::string, SystemErrorMap> _errors;
+  std::map<std::string, SystemErrorMap, std::less<>> _errors;
```
jwpeterson, line 354: Similar comment about `std::less<>` here.

include/base/variable.h
```diff
@@ -56,12 +56,12 @@ class Variable
...
    */
   Variable (System * sys,
-            const std::string & var_name,
+            std::string var_name,
             const unsigned int var_number,
             const unsigned int first_scalar_num,
             const FEType & var_type) :
     _sys(sys),
-    _name(var_name),
+    _name(std::move(var_name)),
```
jwpeterson, line 64: There is probably an "Effective C++" guideline on this, but I'm curious exactly what the difference in performance is here? When passing by const reference we will make a copy (by calling the copy constructor) to store in the class. When passing by copy and moving, somehow a single copy is still made somewhere... the class doesn't steal the resources of any external string. So I see 1 string copy being made either way. Is the difference that we don't construct a *reference* in the new version? I have always thought of references as basically "free" or an "alias" but presumably there is some internal bookkeeping (similar to a pointer) that has to happen for them.

jwpeterson (approved):

include/error_estimation/exact_solution.h
roystgnr: It's not, actually!  `std::less<std::string>` is restricted to uses with `std::string`.  `std::less<>` (i.e. [less<void>](https://en.cppreference.com/w/cpp/utility/functional/less_void)) is a new "transparent comparator" functor, which basically compares any two things you throw at it for which `operator<(thing1,thing2)` is well-defined.  If you use the default comparator, then, you can't do operations like `find` with a `string_view`; if you use a transparent comparator you can.

Kind of annoying that using this feature forces you to use extra syntax, but I guess the C++ standards people weren't happy with the fact that making transparent comparators the default could change the behavior of existing code - if you try to do a `find(thing3)`, where thing3 is of some type `sneaky_type &` that's implicitly convertable to `const key_type &`, then with the default comparator it does that implicit conversion and then does `operator<` with the result, whereas with a transparent comparator it just goes straight to `operator<(const key_type &, const sneaky_type &)`.

include/base/variable.h
roystgnr: If we construct `Variable(sys,str,m,n,fet)` then there's no difference in performance, we just make the same copy we would have made anyway, slightly earlier.  But now we can also choose to construct `Variable(sys,std::move(str),m,n,fet)`, and then the pass-by-value does a move construction on the value, and then the initialization here does another move, and nowhere in the whole sequence is there a copy made.

Kind of ridiculous in this case, I admit.  I'm mostly just trying to get into modern habits for when someday I'm writing code where either bigger strings or more string passing means being able to avoid some copies actually matters.

include/error_estimation/exact_solution.h
jwpeterson: OK, thanks for the explanation. Not particularly self-documenting so you kind of need to be "in the know" to make sense of it. I guess one could explicitly say `std::less<void>` though this is no clearer in my opinion.

include/base/variable.h
jwpeterson: OK, yeah makes sense for the case where the user constructs a string and then has no further use for it, so might as well let the `Variable` take over ownership and avoid any copies.

roystgnr:
This was mostly valuable as a learning experience, and that's more valuable if I actually document what I learned somewhere more public than IMs, so copying my thoughts here:

---

In all the advice I've read, the rule is now "you never pass `std::string` by const reference" - either you're just going to inspect it (in which case pass `string_view`) or you're going to make a modified copy (in which case pass `string` by value and the copy can be done as a move in some user code).  In practice?  We rely on `.c_str()` a bunch, and `string_view` doesn't let you get that because you might be looking at only a window into another string and there might not be a null character at the end.

And then when you've still got a const string & API lower down on the stack, it doesn't save you anything to convert APIs to string_view higher than that (unless their use of the lower API is optional).

For our use cases?  The benefits aren't worth much.  The changes I made are beneficial changes, but not enough to justify the work studying `string_view` and figuring out where it's appropriate, except that that was something I wanted to learn about anyway.  It seems like it'd be extremely useful to people trying to write certain types of string-heavy editor/parser/etc. code, though.

PR #3205, opened 2022-03-24
We need to free what ExodusII malloc()'ed.

roystgnr:
This fixes a Valgrind error for me.  Now we just need to see if it fixes
the same error for Yaqi, and see why it didn't cause the libMesh
Valgrind CI to fail.

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -1175,6 +1175,10 @@ void ExodusII_IO_Helper::read_elem_in_block(int block)
               const int * as_int = static_cast<int *>(attr.values);
               std::copy(as_int, as_int+attr.value_count, bex_elem_degrees.begin());
 
+              // ex_get_attributes did a values=calloc(); free() is our job.
+              if (attr.values)
+                free(attr.values);
+
               // Right now Bezier extraction elements aren't possible
               // for p>2 and aren't useful for p<2, and we don't
               // support anisotropic p...


src/mesh/exodusII_io_helper.C
```diff
@@ -1175,6 +1175,10 @@ void ExodusII_IO_Helper::read_elem_in_block(int block)
               const int * as_int = static_cast<int *>(attr.values);
               std::copy(as_int, as_int+attr.value_count, bex_elem_degrees.begin());
 
+              // ex_get_attributes did a values=calloc(); free() is our job.
+              if (attr.values)
+                free(attr.values);
```
jwpeterson, line 1180: This is... real ugly. Is this even documented anywhere? I didn't see anything in v8.11/exodus/include/exodusII.h

roystgnr: It is **real** ugly.  It's documented on ex_get_attribute.c in a *comment* (not a Doxygen comment even) on line 270.

Exodus is under active development right now, with a friendly + responsive maintainer, so we should definitely see if things are clarified in the latest hash, and if not then file an issue (or just a PR fixing the documentation, maybe with a comment on the PR suggesting an ex_free_attribute(s) API for forward compatibility?).

I didn't want to fire something off before I could be diplomatic about it, though; my initial sarcastic reactions on Slack or even "real ugly" might not have been conducive toward "friendly + responsive" remaining true in the long term.  ;-)  Even if it wasn't taken personally, there's been a bunch of good work in ExodusII recently and there are some new features on the horizon we could really use later, and I don't mind some short-term underdocumentation if that's the price of rapid development.  I've certainly been more than guilty of making that tradeoff myself...

jwpeterson (approved):

src/mesh/exodusII_io_helper.C
jwpeterson: Yeah I agree that your fix is the needed one, and that `ex_free_attribute()` would be preferable, though still not ideal. If you are creating issues, then this "not accessed" comment in the header is also misleading:
```
typedef struct ex_attribute
{
  ex_entity_type entity_type;
  int64_t        entity_id;
  char           name[NC_MAX_NAME];
  ex_type        type; /* int, double, text */
  size_t         value_count;
  void *         values; /* not accessed if NULL */
} ex_attribute;
```

Maybe wrapping `ex_attribute` in some kind of RAII class could also work, though I don't have time to devote to this.

lindsayad:
sure

jwpeterson:
Is there a known issue with CIVET currently? All the tests are Pending for this PR.

lindsayad:
our build boxes are undergoing some reconfiguration

PR #3206, opened 2022-03-24
Deprecate libmesh_make_unique

roystgnr:
We can just use `std::make_unique` now.

I also got rid of `#include "libmesh/auto_ptr.h"` in headers that only needed `<memory>` - though that might break any downstream code that's still using `libmesh_make_unique` and relying on indirect header inclusion to find it.


include/base/auto_ptr.h
```diff
@@ -35,24 +35,12 @@ template<typename T>
 using UniquePtr = std::unique_ptr<T>;
```
jwpeterson, line 35: I would vote to remove the `UniquePtr` alias now as well. Every once in a while we get copy/pasted code from ancient examples that incorrectly re-introduces `UniquePtr` usage, so it would be good to completely avoid that. I have a branch which does this, but it will likely conflict with this larger branch, so maybe better to do as part of this PR?

jwpeterson (approved):
This looks good (and long overdue) to me. Are you waiting on any other CI testing before merging?

include/base/auto_ptr.h
jwpeterson: jwpeterson/libmesh@9834ba282458c60bd96796d9f233d78f97c40916

roystgnr: I agree; we've had it deprecated forever now.  I'd prefer a separate PR though.  I'll do that if I've inadvertently trampled on yours.

roystgnr:
Nah; I was waiting on the Moose Mac CI to be fixed, and now that that's done and passed we're good to go.

PR #3207, opened 2022-03-24
Use C++17 merge where appropriate

roystgnr:


roystgnr:
Those slackers!

I figured going to C++17 was a no-brainer for clang: "[Clang 5 and later](https://clang.llvm.org/cxx_status.html#cxx17) implement all the features of the [ISO C++ 2017 standard](https://www.iso.org/standard/68564.html)" and we don't need the DR features.

But C++17 support **in libc++** isn't so rosy: [they don't give us "Splicing Maps and Sets" support until version 8](https://libcxx.llvm.org/Status/Cxx17.html)!

I guess we're going to need more fine-grained m4 tests after all.  I should be the one to do them this time; John suffered enough when we moved to C++11.

roystgnr:
That one failing test on Mac is a known issue with our CI right now; we can merge anyway.

PR #3208, opened 2022-03-25
Preconditioner reuse in petsc_nonlinear_solver

reverendbedford:
This alters the logic used in PetscNonlinearSolver to:

1. Not free the SNES instance between solves
2. Use SNESReset by default in between solves
3. Add two new options "reuse preconditioner" and "reuse preconditioner maximum iterations" to support preconditioner reuse during and across nonlinear solves.

This PR would close  issue #3209

### DIFF
--- a/include/solvers/nonlinear_solver.h
+++ b/include/solvers/nonlinear_solver.h
@@ -350,6 +350,16 @@ public:
   bool converged;
 
   /**
+   * Whether we should reuse the linear preconditioner
+   */
+  bool reuse_preconditioner;
+
+  /**
+   * Number of linear iterations to retain the preconditioner
+   */
+  unsigned int reuse_preconditioner_max_its;
+
+  /**
    * Set the solver configuration object.
    */
   void set_solver_configuration(SolverConfiguration & solver_configuration);
@@ -415,6 +425,8 @@ NonlinearSolver<T>::NonlinearSolver (sys_type & s) :
   initial_linear_tolerance(0),
   minimum_linear_tolerance(0),
   converged(false),
+  reuse_preconditioner(false),
+  reuse_preconditioner_max_its(0),
   _system(s),
   _is_initialized (false),
   _preconditioner (nullptr),
--- a/include/solvers/nonlinear_solver.h
+++ b/include/solvers/nonlinear_solver.h
@@ -350,21 +350,42 @@ public:
   bool converged;
 
   /**
-   * Whether we should reuse the linear preconditioner
-   */
-  bool reuse_preconditioner;
-
-  /**
-   * Number of linear iterations to retain the preconditioner
-   */
-  unsigned int reuse_preconditioner_max_its;
-
-  /**
    * Set the solver configuration object.
    */
   void set_solver_configuration(SolverConfiguration & solver_configuration);
 
+  /**
+   *  Get the reuse_preconditioner flag
+   */
+  virtual bool reuse_preconditioner() const;
+
+  /**
+   *  Set the reuse preconditioner flag
+   */
+  virtual void set_reuse_preconditioner(bool reuse);
+
+  /**
+   *  Get the reuse_preconditioner_max_its parameter
+   */
+  virtual unsigned int reuse_preconditioner_max_its() const;
+
+  /**
+   *  Set the reuse_preconditioner_max_its parameter
+   */
+  virtual void set_reuse_preconditioner_max_its(unsigned int i);
+
+
 protected:
+  /**
+   * Whether we should reuse the linear preconditioner
+   */
+  bool _reuse_preconditioner;
+
+  /**
+   * Number of linear iterations to retain the preconditioner
+   */
+  unsigned int _reuse_preconditioner_max_its;
+
   /**
    * A reference to the system we are solving.
    */
@@ -425,8 +446,8 @@ NonlinearSolver<T>::NonlinearSolver (sys_type & s) :
   initial_linear_tolerance(0),
   minimum_linear_tolerance(0),
   converged(false),
-  reuse_preconditioner(false),
-  reuse_preconditioner_max_its(0),
+  _reuse_preconditioner(false),
+  _reuse_preconditioner_max_its(0),
   _system(s),
   _is_initialized (false),
   _preconditioner (nullptr),
--- a/include/solvers/petsc_nonlinear_solver.h
+++ b/include/solvers/petsc_nonlinear_solver.h
@@ -93,7 +93,7 @@ public:
   /**
    * Destructor.
    */
-  ~PetscNonlinearSolver () = default;
+  ~PetscNonlinearSolver ();
 
   /**
    * Release all memory and clear data structures.
--- a/include/solvers/petsc_nonlinear_solver.h
+++ b/include/solvers/petsc_nonlinear_solver.h
@@ -47,6 +47,7 @@ class ResidualContext;
 // need access to these most of the time as they are used internally by this object.
 extern "C"
 {
+  PetscErrorCode libmesh_petsc_recalculate_monitor(SNES snes, PetscInt it, PetscReal norm, void* mctx);
   PetscErrorCode libmesh_petsc_snes_monitor (SNES, PetscInt its, PetscReal fnorm, void *);
   PetscErrorCode libmesh_petsc_snes_residual (SNES, Vec x, Vec r, void * ctx);
   PetscErrorCode libmesh_petsc_snes_fd_residual (SNES, Vec x, Vec r, void * ctx);
@@ -201,6 +202,11 @@ public:
    */
   std::unique_ptr<ComputeLineSearchObject> linesearch_object;
 
+  /**
+    * Setup the default monitor if required
+    */
+  void setup_default_monitor();
+
 protected:
 
   /**
@@ -261,6 +267,11 @@ protected:
    */
   bool _computing_base_vector;
 
+  /**
+    * Whether we've triggered the preconditioner reuse
+    */
+  bool _setup_reuse;
+
 private:
   friend ResidualContext libmesh_petsc_snes_residual_helper (SNES snes, Vec x, void * ctx);
   friend PetscErrorCode libmesh_petsc_snes_residual (SNES snes, Vec x, Vec r, void * ctx);
@@ -268,8 +279,6 @@ private:
   friend PetscErrorCode libmesh_petsc_snes_mffd_residual (SNES snes, Vec x, Vec r, void * ctx);
   friend PetscErrorCode libmesh_petsc_snes_jacobian (SNES snes, Vec x, Mat jac, Mat pc, void * ctx);
 };
-
-
 
 } // namespace libMesh
 
--- a/include/solvers/petsc_nonlinear_solver.h
+++ b/include/solvers/petsc_nonlinear_solver.h
@@ -282,6 +282,11 @@ protected:
     */
   bool _setup_reuse;
 
+  /**
+   *  Persistent storage for the maximum iterations
+   */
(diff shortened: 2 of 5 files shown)


lindsayad:
@fdkong

roystgnr:
Looks reasonable to me.  I'd love to get some test coverage on it, but I know it'd be a pain in the neck to get something that covers the PETSc case but doesn't false positive on other solvers.

include/solvers/petsc_nonlinear_solver.h
```diff
@@ -261,6 +262,16 @@ class PetscNonlinearSolver : public NonlinearSolver<T>
    */
   bool _computing_base_vector;
 
+  /**
+    *  Whether we have a live snes instance
+    */
+  bool _snes_allocated;
```
fdkong, line 268: `  WrappedPetsc<SNES> _snes;`  should have  enough information to detect whether or not SNES is created or not.  For example,

```
  /**
   * User-defined conversion to bool. This is intended to mimic code like:
   * IS is = nullptr;
   * ...
   * if (!is)
   *   ...
   * Note that this comparison is thus concerned with obj itself and
   * not &obj.
   */
  operator bool() const { return obj != nullptr; }
```
Considering that, I think should not need this extra flag `_snes_allocated`

src/solvers/petsc_nonlinear_solver.C
```diff
@@ -105,6 +105,33 @@ libmesh_petsc_snes_residual_helper (SNES snes, Vec x, void * ctx)
...
+  // added to the SNES instance if we're reusing the preconditioner
+  PetscErrorCode
+  libmesh_petsc_recalculate_monitor(SNES snes, PetscInt, PetscReal, void* mctx)
+  {
+    unsigned int * max_iters = (unsigned int *) mctx;
+    PetscErrorCode ierr = 0;
+
+    KSP ksp;
+    ierr = SNESGetKSP(snes, &ksp);
+    if (ierr != 0) return ierr;
```
fdkong, line 120: `LIBMESH_CHKERR(ierr);`

src/solvers/petsc_nonlinear_solver.C
```diff
@@ -825,6 +876,38 @@ PetscNonlinearSolver<T>::solve (SparseMatrix<T> &  pre_in,  // System Preconditi
...
+    }
+  else if (!(this->reuse_preconditioner))
+    // This covers the case where it was enabled but was then disabled
+    {
+      ierr = SNESSetLagPreconditionerPersists(_snes, PETSC_FALSE);
+      LIBMESH_CHKERR(ierr);
+      if (_setup_reuse)
+        {
+          _setup_reuse = false;
+          ierr = SNESMonitorCancel(_snes);
```
fdkong, line 906: We can not really do this. This will kill all monitors

fdkong (changes requested):
Thanks for contributions. If I am wrong, please correct me. 

From my understanding, preconditioner reuse within a single nonlinear solve, can be realized using "`-snes_lag_preconditioner`" without changing libmesh code.

Preconditioner reuse across multiple time steps can be implemented using PETSc options as well such as the combination of `-snes_lag_preconditioner_persists` and `-snes_lag_preconditioner`. The only issue we have in libmesh right now is that we throw aways SNES after each nonlinear solve.

It is great that this PR take an attempt to keep SNES after each nonlinear solve, which introduces more possibilities on PC reuse.

My concern here is that we introduce some PC-reuse parameters  and a few related functions that are not necessary, since these capabilities are accessible via PETSc options. If we can not do that, I am happy to enhance PETSc.  My philosophy  here is that we should maintain as little PETSc code as possible in libmesh/MOOSE.

Therefore, my suggestion would be to keep the capability that can have SNES live across several time steps, but remove all other stuffs such as: `reuse_preconditioner, reuse_preconditioner_max_its,  libmesh_petsc_recalculate_monitor
`

src/solvers/petsc_nonlinear_solver.C
reverendbedford: Currently everything is re-setup each time, so the only monitor that might be canceled is the default monitor.  We should be able to fix this by adding

```
      if (_default_monitor)
        {
          ierr = SNESMonitorSet (_snes, libmesh_petsc_snes_monitor,
                                 this, PETSC_NULL);
          LIBMESH_CHKERR(ierr);
        }
```

after the this, right?  Or better move that to the `make_and_add_default_monitor` method.

include/solvers/petsc_nonlinear_solver.h
reverendbedford: Yep, this is a good idea.

src/solvers/petsc_nonlinear_solver.C
reverendbedford: Yep

reverendbedford:
@fdkong, no PETSc does not do what I want right now.  `-snes_lag_preconditioner` only has options for 

- "recalculate the preconditioner next time you need it and never again"
- "never recalculate"
- "recalculate every X times we calculate the Jacobian" where X is a fixed number (so every time, ever other time, etc.)

What I want is "recalculate whenever it took more than Y linear iterations to solve the system" which isn't provided by current PETSc options, as best as I can tell.  Using the monitor here provides this functionality.  If it's possible to implement this functionality in PETSc that's fine, but it's not currently there.

fdkong:
It might be much clearer if we could implement this in PETSc, when we consider that all information required are at the algebra level.  But the current hookup approach might be an easy one. So I am fine with this PR.

reverendbedford:
I'm not qualified to work on PETSc :) but if this type of heuristic ever gets implemented over there I'd be happy to then go back and update libmesh and MOOSE.  So let me make your suggested changes/fixes here and then figure out why those 2 stupid tests are failing in MOOSE and I'll ping you for re-review.

reverendbedford:
@fdkong I'm sorry I dropped the ball on this.  This commit ought to address your review plus fix the two failing MOOSE tests.  Could you or someone else release the tests?

include/solvers/nonlinear_solver.h
```diff
@@ -349,6 +349,16 @@ class NonlinearSolver : public ReferenceCountedObject<NonlinearSolver<T>>,
    */
   bool converged;
 
+  /**
+   * Whether we should reuse the linear preconditioner
+   */
+  bool reuse_preconditioner;
```
roystgnr, line 355: Can we use getters and setters (defaulting to `libmesh_not_implemented()` outside of PETSc) rather than public member access for these new variables?  It would be nice if, when someone tries to use this feature with Nox or with future subclasses that don't support it right away, they get an error message rather than getting silently ignored.

reverendbedford: Yep, good idea.

reverendbedford: Yep, good idea.

jwpeterson:
@reverendbedford I sent you an invite to join the libMesh Associates group. Members of this group have the PRs tested automatically, so you won't need a developer to activate the tests every time.

reverendbedford:
@jwpeterson Thanks!  I joined.  But I think you need to activate the tests one last time for me :)

reverendbedford:
I would hold off merging until I modify MOOSE to actually use this.  I can at least link the MOOSE PR, which will have tests, here.

jwpeterson:
> I can at least link the MOOSE PR, which will have tests, here.

Your MOOSE PR will need to use a libmesh that includes your branch from this PR... so we would need to push your branch to libmesh upstream and you would have to do a libmesh update on your MOOSE PR in order to test it.

reverendbedford:
Good point, but at least let me test locally first.  I'm starting to think that I can't actually test the truth value of the `_snes` object and may need the flag...

reverendbedford:
Ah, right.  Well at least let me run some tests locally.  I'm having trouble finding a small problem that will actually trigger the behavior we want to test.

reverendbedford:
@jwpeterson Yes, I screwed something up when I made the changes last night.  This commit fixes it and I now have MOOSE tests ready to go.  Those tests check that we can trigger the correct behavior, specifically that you reuse the same preconditioner until the linear iterations exceed the "maximum iterations" parameter.

Please merge when you have a chance, this should now be good to go.

hugary1995:
I can't wait to try this :-)

src/solvers/petsc_nonlinear_solver.C
```diff
@@ -615,15 +642,23 @@ PetscNonlinearSolver<T>::PetscNonlinearSolver (sys_type & system_in) :
...
-PetscNonlinearSolver<T>::~PetscNonlinearSolver () = default;
-
+PetscNonlinearSolver<T>::~PetscNonlinearSolver ()
+{
+  // Take care of the SNES instance, if it was allocated
+  if (_snes)
+    {
+    _snes.destroy();
+    }
+}
```
jwpeterson, line 661: This change should definitely not be necessary, so I'm curious whether some testing you did confirmed that you needed it? The main purpose of the
```
WrappedPetsc<SNES> _snes;
```
object is that it is destroyed automatically when its destructor is called, allowing us to have `default` destructor implementations.

src/solvers/petsc_nonlinear_solver.C
```diff
@@ -825,6 +870,41 @@ PetscNonlinearSolver<T>::solve (SparseMatrix<T> &  pre_in,  // System Preconditi
   // Should actually be a PetscReal, but I don't know which version of PETSc first introduced PetscReal
   Real final_residual_norm=0.;
 
+  // We don't want to do this twice because it resets
+  // SNESSetLagPreconditioner
+  if ((reuse_preconditioner()) && (!_setup_reuse))
+    {
+      _setup_reuse = true;
+      ierr = SNESSetLagPreconditionerPersists(_snes, PETSC_TRUE);
```
jwpeterson, line 878: Just a comment for myself as we may have already discussed this: `SNESSetLagPreconditionerPersists` was added in petsc/petsc@37ec4e1a76a069aca3c25363b0dc0a5311d937ea which dates back to PETSc 3.5, so I don't think any version guards are needed here.

jwpeterson (changes requested):
@reverendbedford sorry for losing track of this PR... it looks like @roystgnr and @fdkong signed off on it already and were just waiting on me. I took another look and my only concern is that you un-defaulted the `PetscNonlinearSolver` destructor, which I don't think should be necessary. If you agree, I can go ahead and make that change and finally get this merged.

src/solvers/petsc_nonlinear_solver.C
reverendbedford: This I don't know.

src/solvers/petsc_nonlinear_solver.C
reverendbedford: I believe it was necessary and I didn't understand why.  I'll check next week -- I'm on conference travel and unfortunately having a great deal of double compiling from scratch on my macbook...

reverendbedford: Hmm, I really should not be needed.  I'll check when I get back (or if I can get everything to compile without conda while traveling).

reverendbedford: My tests in MOOSE on actually using the feature pass with the default destructor.  For some reason the tests aren't running now, but once they do this should be good.

reverendbedford:
@jwpeterson should be good now, though for some reason the tests over here aren't running.

jwpeterson:
> @jwpeterson should be good now, though for some reason the tests over here aren't running.

OK, yeah I think CIVET is down for scheduled maintenance for a few days right now.

jwpeterson (approved):

PR #3210, opened 2022-03-28
Free all the attributes we retrieve with ex_get_attributes

lindsayad:
Previously we only freed the attribute corresponding to bex_elem_degrees


roystgnr (approved):
Yup, I certainly missed that.

PR #3216, opened 2022-04-05
Deprecate non-std:: is_sorted, iota

roystgnr:
Those have been available since C++11, they're not just SGI extensions
anymore.

### DIFF
--- a/include/utils/utility.h
+++ b/include/utils/utility.h
@@ -25,7 +25,7 @@
 // System includes
 #include <string>
 #include <vector>
-#include <algorithm> // for std::lower_bound
+#include <algorithm> // is_sorted, lower_bound
 
 namespace libMesh
 {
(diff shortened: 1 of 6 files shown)


include/utils/utility.h
```diff
@@ -216,58 +215,16 @@ void iota (ForwardIter first, ForwardIter last, T value)
...
-  //   ^    ^
-  // prev first
-  //
-  //  [-    -    -    -    -    -]
-  //        ^    ^
-  //      prev first
-  //
-  //  [-    -    -    -    -    -]
-  //             ^    ^
-  //           prev first
```
jwpeterson, line 241: Beautiful ASCII art

jwpeterson (approved):
Looks good to me. I'd say remove it after the next release maybe. I don't think either of these would have been widely used in user code.

PR #3218, opened 2022-04-05
Fix clang warnings

roystgnr:
This is what I was able to reproduce on Linux, at least.  Pinging @fdkong if he want to see if it works for his Mac builds too.

### DIFF
--- a/src/fe/fe_boundary.C
+++ b/src/fe/fe_boundary.C
@@ -1083,7 +1083,7 @@ void FEMap::compute_edge_map(int dim,
   if (calculate_dxyz)
     for (unsigned int p=0; p<n_qp; p++)
       {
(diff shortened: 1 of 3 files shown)


src/fe/fe_boundary.C
```diff
@@ -1083,7 +1083,7 @@ void FEMap::compute_edge_map(int dim,
   if (calculate_dxyz)
     for (unsigned int p=0; p<n_qp; p++)
       {
-        const Point n  = this->dxyzdxi_map[p].cross(this->dxyzdeta_map[p]);
+        // const Point n  = this->dxyzdxi_map[p].cross(this->dxyzdeta_map[p]);
```
jwpeterson, line 1086: This line of code must be pretty old... strange that it never showed up as a warning before now.

jwpeterson (approved):

PR #3219, opened 2022-04-05
Modern C++ parameter passing

roystgnr:
This is a mix of:

1) more of the pass-by-value-then-move trick for taking copies.  I wasn't able to find any place this would be a big deal for performance, but in the course of looking I found:

2) a bunch places where we can pass braced initializer lists directly rather than laboriously constructing a named std:: container and passing that.

I like the results, but nothing in this PR is critical and everything in it is debatable.


jwpeterson (approved):
Nice cleanup, especially in the examples.

roystgnr:
Interesting - that looks like an actual regression in misc ex6, that I must have missed because I didn't have --disable-strict-lgpl in my own builds.

PR #3220, opened 2022-04-05
set_hdf5_writing(bool) option

roystgnr:
We're going to want to turn this off in Moose, so users' new files will
be as portable as possible to other builds (including new gold
regression test files and including our --disable-hdf5 CI recipes).

We'll leave it on by default in libMesh for backwards compatibility;
anybody who went to the trouble of building with HDF5 manually probably
had a reason to want HDF5 output.

### DIFF
--- a/include/mesh/exodusII_io.h
+++ b/include/mesh/exodusII_io.h
@@ -495,6 +495,14 @@ public:
 #endif
 
   /**
+   * Set to true (the default) to write files in an HDF5-based file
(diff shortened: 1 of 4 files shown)


roystgnr:
I tested out setting hdf writing to false manually - the type reported by Linux `file` does change from HDF5 to NetCDF, and the written files are readable by a non-HDF5-enabled libMesh.

jwpeterson (approved):

PR #3221, opened 2022-04-05
Structured bindings

jwpeterson:
This provides improved syntax only, avoiding `pr.first.second` type code when possible. I worked on this a while back but never got it merged. Since the C++17 commits seem to be heating up, I thought it would be a good time to put it out there. No idea when I will get time to go through every file making these changes.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -1954,17 +1954,17 @@ void DofMap::process_mesh_constraint_rows(const MeshBase & mesh)
 
           const dof_id_type constrained_id =
             node->dof_number(sys_num, var_num, 0);
-          for (auto pr : node_row.second)
+          for (const auto & [pr, val] : node_row.second)
             {
-              const Elem * spline_elem = pr.first.first;
+              const Elem * spline_elem = pr.first;
               libmesh_assert(spline_elem == mesh.elem_ptr(spline_elem->id()));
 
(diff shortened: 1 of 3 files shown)


src/base/dof_map_constraints.C
```diff
@@ -2281,23 +2281,21 @@ std::string DofMap::get_local_constraints(bool print_nonlocal) const
   os << "Node Constraints:"
      << std::endl;
 
-  for (const auto & pr : _node_constraints)
+  for (const auto & [node, pr] : _node_constraints)
```
roystgnr, line 2284: It's a shame there's no movement on nested structured bindings.  I found [a "flatten" function](https://www.reddit.com/r/cpp_questions/comments/eto1rh/nesting_structured_bindings/) that works around the problem for some declarations but I don't think it'd be usable in a range for loop.

roystgnr (approved):

roystgnr:
> No idea when I will get time to go through every file making these changes.

I got [a big swath of them](https://github.com/libMesh/libmesh/pull/3168/commits/da9ef07f3ca7b94c90bc2b1be76be97bab316cc2) already ... but while I don't remember what regex I was using to try to find them, it clearly wasn't broad enough.  I really do love this category of change, though; you get much more readable names and much terser code, and the only cost (in places where we're not already using `auto` out of laziness or genericity) is slightly less readable type info.

PR #3224, opened 2022-04-06
Clang-tidy cleanup

roystgnr:
Most of the fixes here are trivial performance improvements and/or slightly clearer code; the one serious bugfix I moved into #3223

Even the workarounds for false positives might be worth it in the long run.  Ever since I discovered that nothing less would do to catch "used a container after std::move" bugs, I've been thinking of getting a clang-tidy target into our Makefiles and CI recipes; it's annoying when there are false positives to ignore or work around, but #3223 alone has been worth the hassle.

### DIFF
--- a/include/partitioning/parmetis_helper.h
+++ b/include/partitioning/parmetis_helper.h
@@ -26,6 +26,8 @@
 
 #ifdef LIBMESH_HAVE_PARMETIS
 
+#include "libmesh/id_types.h"
+
 // Before we include a header wrapped in a namespace, we'd better make
 // sure none of its dependencies end up in that namespace
 #include <mpi.h>
--- a/src/mesh/distributed_mesh.C
+++ b/src/mesh/distributed_mesh.C
@@ -90,7 +90,7 @@ MeshBase & DistributedMesh::assign(MeshBase && other_mesh)
 
 DistributedMesh::~DistributedMesh ()
 {
-  this->clear();  // Free nodes and elements
+  this->DistributedMesh::clear();  // Free nodes and elements
 }
 
 
@@ -967,7 +967,7 @@ void DistributedMesh::clear ()
   // There is no need to remove them from
   // the BoundaryInfo data structure since we
   // already cleared it.
-  this->clear_elems();
+  this->DistributedMesh::clear_elems();
 
   for (auto & node : _nodes)
     delete node;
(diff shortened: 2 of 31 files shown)


src/mesh/distributed_mesh.C
```diff
@@ -90,7 +90,7 @@ MeshBase & DistributedMesh::assign(MeshBase && other_mesh)
 
 DistributedMesh::~DistributedMesh ()
 {
-  this->clear();  // Free nodes and elements
+  this->DistributedMesh::clear();  // Free nodes and elements
```
jwpeterson, line 93: This is the same behavior as before, yes? I guess maybe I'd remove the `this->` as it seems (even more) redundant now.

include/partitioning/parmetis_helper.h
```diff
@@ -26,6 +26,8 @@
 
 #ifdef LIBMESH_HAVE_PARMETIS
 
+#include "libmesh/id_types.h"
+
```
jwpeterson, line 30: This could be because we get Parmetis from PETSc in any builds where we run `make test_headers`?

src/utils/perf_log.C
```diff
@@ -111,8 +111,9 @@ void PerfLog::push (const std::string & label,
     label_c_str = non_temporary_strings[label];
   else
     {
-      char * newcopy = new char [label.size()+1];
-      strcpy(newcopy, label.c_str());
+      const std::size_t labelsizep1 = label.size()+1;
+      char * newcopy = new char [labelsizep1];
+      strncpy(newcopy, label.c_str(), labelsizep1);
```
jwpeterson, line 116: We include the `<cstring>` in this C file, so I would have thought `std::strncpy` would be more strictly correct here.

jwpeterson (approved):

include/partitioning/parmetis_helper.h
roystgnr: That might be it!  We even autodetect it and prefer it to our own if we find it, don't we?  So even my own `make test_headers` runs would have been fooled.

src/mesh/distributed_mesh.C
roystgnr: It's even more redundant from my standpoint too, but is it from the compiler's standpoint?  I never really managed to understand the whole "you need `this->` if you have a template base class or if you might refactor in the future to have a template base class or if you ever took a deep breath while an editor had a templated class open" philosophy so I've just been using `this->` whenever in the slightest doubt.

PR #3225, opened 2022-04-08
Add support for element sets (elemsets) to MeshBase, Exodus

jwpeterson:
This PR adds support for reading and writing elemsets in Exodus files, and the accompanying data structures in MeshBase. Element sets are similar to the existing sidesets and nodesets in the Exodus writer, although they are not quite so specific to boundary conditions as those other two sets are. They also have some characteristics in common with subdomains, but (1) an elem can be in any number of elemsets whereas it can only be in one subdomain and (2) if you are using Exodus files, elemsets are not tied to geometric element type the way that blocks (and unfortunately subdomains) are. Since elemset ids are not exactly boundary ids and not exactly subdomain ids, I introduced a new type, `elemset_id_type`, which is currently `typedef`'d to `subdomain_id_type`. This avoids some confusion and also allows us to change it to something else in the future if desired.

In consultation with @roystgnr, we decided to use the "extra integer" feature of the Elem class in order to implement this. This way there should be less worries about keeping the elemset ids communicated and up-to-date in parallel, though admittedly this first draft of the PR may have some parallel issues... I have only tested in serial up to this point. 

In order to use the extra_integer approach, you must define the *same* number of extra integers per Elem. So, rather than try and define num_elemsets extra_integers on every Elem, we define a *single* extra integer on every Elem and use a table to map from {sets to which an elem belongs} -> extra_integer and back. Currently I've named this extra_integer `"elemset_code"`, but I'm open to discussion on the name. Whatever name we choose would effectively need to be reserved for use by the internals of libMesh in order to avoid strange behavior in e.g. the Exodus writer, which currently checks for the existence of an extra_integer by this name.

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -283,7 +283,7 @@ public:
    * calling add_elemset_code(). If no such code/set is found, returns
    * the empty set or DofObject::invalid_id, respectively.
    */
-  std::set<subdomain_id_type> get_elemsets(dof_id_type elemset_code) const;
+  void get_elemsets(dof_id_type elemset_code, std::set<subdomain_id_type> & id_set_to_fill) const;
   dof_id_type get_elemset_code(const std::set<subdomain_id_type> & id_set) const;
 
   /**
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -271,6 +271,13 @@ public:
   void add_elemset_code(dof_id_type code, const std::set<subdomain_id_type> & id_set);
 
   /**
+   * Determines the number of unique elemset ids which have been added
+   * via add_elemset_code() calls by looping over the
+   * _elemset_codes_inverse_map and counting them.
+   */
+  unsigned int n_elemsets() const;
+
+  /**
    * Look up the element sets for a given elemset code and
    * vice-versa. The elemset must have been previously stored by
    * calling add_elemset_code(). If no such code/set is found, returns
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -268,7 +268,7 @@ public:
    * specified in \p id_set. Also sets up the inverse mapping, so that if one knows
    * all the element sets an Elem belongs to, one can look up the corresponding code.
    */
-  void add_elemset_code(dof_id_type code, const std::set<subdomain_id_type> & id_set);
+  void add_elemset_code(dof_id_type code, const std::set<elemset_id_type> & id_set);
 
   /**
    * Determines the number of unique elemset ids which have been added
@@ -283,8 +283,8 @@ public:
    * calling add_elemset_code(). If no such code/set is found, returns
    * the empty set or DofObject::invalid_id, respectively.
    */
-  void get_elemsets(dof_id_type elemset_code, std::set<subdomain_id_type> & id_set_to_fill) const;
-  dof_id_type get_elemset_code(const std::set<subdomain_id_type> & id_set) const;
+  void get_elemsets(dof_id_type elemset_code, std::set<elemset_id_type> & id_set_to_fill) const;
+  dof_id_type get_elemset_code(const std::set<elemset_id_type> & id_set) const;
 
   /**
    * \returns The "spatial dimension" of the mesh.
@@ -1975,8 +1975,7 @@ protected:
    * stored as an extra_integer (named "elemset_code") on all elements,
    * and extra_integers are of type dof_id_type. Elements which do not
    * belong to any set should be assigned an elemset code of DofObject::invalid_id.
-   * 2.) We use subdomain_id_type for the element set id type since
-   * element sets can be thought of as a generalization of the concept
+   * 2.) Element sets can be thought of as a generalization of the concept
    * of a subdomain. Subdomains have the following restrictions:
    *   a.) A given element can only belong to a single subdomain
    *   b.) When using Exodus file input/output, subdomains are (unfortunately)
@@ -1989,8 +1988,8 @@ protected:
    * with the one requirement that elements which belong to no sets
    * should have a set code of DofObject::invalid_id.
    */
-  std::map<dof_id_type, std::set<subdomain_id_type>> _elemset_codes;
-  std::map<std::set<subdomain_id_type>, dof_id_type> _elemset_codes_inverse_map;
+  std::map<dof_id_type, std::set<elemset_id_type>> _elemset_codes;
+  std::map<std::set<elemset_id_type>, dof_id_type> _elemset_codes_inverse_map;
 
   /**
    * The "spatial dimension" of the Mesh.  See the documentation for
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -264,6 +264,22 @@ public:
   void set_elem_dimensions(const std::set<unsigned char> & elem_dims);
 
   /**
+   * Tabulate a user-defined "code" for elements which belong to the element sets
+   * specified in \p id_set. Also sets up the inverse mapping, so that if one knows
+   * all the element sets an Elem belongs to, one can look up the corresponding code.
+   */
+  void add_elemset_code(dof_id_type code, const std::set<subdomain_id_type> & id_set);
+
+  /**
+   * Look up the element sets for a given elemset code and
+   * vice-versa. The elemset must have been previously stored by
+   * calling add_elemset_code(). If no such code/set is found, returns
+   * the empty set or DofObject::invalid_id, respectively.
+   */
+  std::set<subdomain_id_type> get_elemsets(dof_id_type elemset_code) const;
+  dof_id_type get_elemset_code(const std::set<subdomain_id_type> & id_set) const;
+
+  /**
    * \returns The "spatial dimension" of the mesh.
    *
    * The spatial dimension is defined as:
@@ -1944,6 +1960,30 @@ protected:
    * We cache the subdomain ids of the elements present in the mesh.
    */
   std::set<subdomain_id_type> _mesh_subdomains;
+
+  /**
+   * Map from "element set code" to list of set ids to which that element
+   * belongs (and vice-versa). Remarks:
+   * 1.) The elemset code is a dof_id_type because (if used) it is
+   * stored as an extra_integer (named "elemset_code") on all elements,
+   * and extra_integers are of type dof_id_type. Elements which do not
+   * belong to any set should be assigned an elemset code of DofObject::invalid_id.
+   * 2.) We use subdomain_id_type for the element set id type since
+   * element sets can be thought of as a generalization of the concept
+   * of a subdomain. Subdomains have the following restrictions:
+   *   a.) A given element can only belong to a single subdomain
+   *   b.) When using Exodus file input/output, subdomains are (unfortunately)
+   *   tied to the concept of exodus element blocks, which consist of a single
+   *   geometric element type, somewhat limiting their generality.
+   * 3.) The user is responsible for filling in the values of this map
+   * in a consistent manner, unless the elemsets are read in from an
+   * Exodus file, in which case the elemset codes will be set up
+   * automatically. The codes can basically be chosen arbitrarily,
+   * with the one requirement that elements which belong to no sets
+   * should have a set code of DofObject::invalid_id.
+   */
+  std::map<dof_id_type, std::set<subdomain_id_type>> _elemset_codes;
+  std::map<std::set<subdomain_id_type>, dof_id_type> _elemset_codes_inverse_map;
 
   /**
    * The "spatial dimension" of the Mesh.  See the documentation for
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -264,11 +264,19 @@ public:
   void set_elem_dimensions(const std::set<unsigned char> & elem_dims);
 
   /**
+   * Typedef for the "set" container used to store elemset ids. The
+   * main requirements are that the entries be sorted and unique, so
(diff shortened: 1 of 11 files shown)


src/mesh/exodusII_io_helper.C
```diff
@@ -1446,7 +1450,7 @@ void ExodusII_IO_Helper::read_nodeset_info()
       EX_CHECK_ERR(ex_err, "Error retrieving nodeset information.");
       message("All nodeset information retrieved successfully.");
 
-      // Resize appropriate data structures -- only do this once outnode the loop
+      // Resize appropriate data structures -- only do this once outside the loop
```
jwpeterson, line 1453: I got a laugh out of this side -> node text replacement that went awry.

include/mesh/mesh_base.h
```diff
@@ -263,6 +263,29 @@ class MeshBase : public ParallelObject
...
+   */
+  unsigned int n_elemsets() const;
+
+  /**
+   * Look up the element sets for a given elemset code and
+   * vice-versa. The elemset must have been previously stored by
+   * calling add_elemset_code(). If no such code/set is found, returns
+   * the empty set or DofObject::invalid_id, respectively.
+   */
+  void get_elemsets(dof_id_type elemset_code, std::set<elemset_id_type> & id_set_to_fill) const;
```
roystgnr, line 286: Should we just return rather than fill in new APIs, now that C++17 mandates copy/move construction?

include/mesh/mesh_base.h
```diff
@@ -1945,6 +1968,29 @@ class MeshBase : public ParallelObject
...
+   *   geometric element type, somewhat limiting their generality.
+   * 3.) The user is responsible for filling in the values of this map
+   * in a consistent manner, unless the elemsets are read in from an
+   * Exodus file, in which case the elemset codes will be set up
+   * automatically. The codes can basically be chosen arbitrarily,
+   * with the one requirement that elements which belong to no sets
+   * should have a set code of DofObject::invalid_id.
+   */
+  std::map<dof_id_type, std::set<elemset_id_type>> _elemset_codes;
+  std::map<std::set<elemset_id_type>, dof_id_type> _elemset_codes_inverse_map;
```
roystgnr, line 1992: Since we're keeping these in sync invariably and encapsulating them properly, could we avoid duplicating the set data somehow?  Make the values in _elemset_codes be just pointers to the keys in _elemset_codes_inverse_map, perhaps?  That wouldn't work with `unordered_map` (rehashing invalidates iterators) but it ought to be well-defined behavior with `map`.

src/mesh/mesh_base.C
```diff
@@ -218,6 +222,43 @@ void MeshBase::set_elem_dimensions(const std::set<unsigned char> & elem_dims)
   _elem_dims = elem_dims;
 }
 
+
+
+void MeshBase::add_elemset_code(dof_id_type code, const std::set<elemset_id_type> & id_set)
```
roystgnr, line 227: Should we pass id_set by value, so we can do move construction and avoid a copy in some cases?

include/mesh/mesh_base.h
jwpeterson: I thought about this, but I generally expect this function to be called for each Elem in an Elem-loop... so I think that would still construct/destroy one std::set per loop iteration (?) whereas with this you could construct the std::set outside the loop and reuse it. Or maybe you are saying that

```
std::set<elemset_id_type> s;
for (elem loop...)
{
  s = get_elemsets(elem->get_extra_integer(elemset_index));
}
```
would move-assign `s` each time, giving the same effect?

include/mesh/mesh_base.h
```diff
@@ -263,6 +263,29 @@ class MeshBase : public ParallelObject
...
+   * Tabulate a user-defined "code" for elements which belong to the element sets
+   * specified in \p id_set. Also sets up the inverse mapping, so that if one knows
+   * all the element sets an Elem belongs to, one can look up the corresponding code.
+   */
+  void add_elemset_code(dof_id_type code, const std::set<elemset_id_type> & id_set);
+
+  /**
+   * Determines the number of unique elemset ids which have been added
+   * via add_elemset_code() calls by looping over the
+   * _elemset_codes_inverse_map and counting them.
```
roystgnr, line 276: That's kind of expensive for what's usually an O(1) type operation, if it might be called repeatedly.  Should we just keep a full `set<elemset_id_type>` in sync?  Or would doing that backfire by killing the asymptotic complexity of a future `remove_elemset_code`?

src/mesh/mesh_base.C
jwpeterson: The current "two maps" approach uses `id_set` twice:
```
void MeshBase::add_elemset_code(dof_id_type code, const std::set<elemset_id_type> & id_set)
{
  _elemset_codes.emplace(code, id_set);
  _elemset_codes_inverse_map.emplace(id_set, code);
}
```
so I think you'd need to `std::move()` it in the first call, leaving it invalid for the second call. Moving to a single map approach of some kind would help with this though.

include/mesh/mesh_base.h
jwpeterson: I would like to look be able to quickly look up in both directions though. Maybe that is possible in the approach you described, not sure. There is a bi-directional map in boost I think, which is basically what I'm going for here with two maps.

include/mesh/mesh_base.h
```diff
@@ -263,6 +263,29 @@ class MeshBase : public ParallelObject
    */
   void set_elem_dimensions(const std::set<unsigned char> & elem_dims);
 
+  /**
+   * Tabulate a user-defined "code" for elements which belong to the element sets
+   * specified in \p id_set. Also sets up the inverse mapping, so that if one knows
+   * all the element sets an Elem belongs to, one can look up the corresponding code.
+   */
+  void add_elemset_code(dof_id_type code, const std::set<elemset_id_type> & id_set);
```
roystgnr, line 271: Bear with me, but: are we sure we want to use `set` to store a set?  A sorted `vector` is typically faster unless the container size gets quite large or the number of insertions quite high, and in those cases `unordered_set` can be faster.

I could see this becoming a seriously useful fundamental feature, and if we're not certain what the best API is now, we might want to wrap it in some `ElemSets` class so that we can change the data structure later.

include/mesh/mesh_base.h
jwpeterson: Yeah, I was imagining that there would only be a few elemsets defined, but your comment makes sense in the case where somebody inevitably wants to define thousands of elemsets... I would be OK with adding an additional `set<elemset_id_type>` member to MeshBase, but I have not yet considered the consequences of a `remove_elemset()` capability. This would trigger a change to all the elemset codes involving that elemset... it seems messy.

include/mesh/mesh_base.h
roystgnr: Oh, that's an interesting point.

I think the construction *doesn't* get elided here, because that `s =` is an assignment rather than an initialization.  It will be a move assignment, but that move assignment won't have the same effect in any `std::set` implementation that reuses node memory, but I'm not sure if there *is* any `std::set` implementation that reuses memory.  If there is then pulling the `set` construction out of the loop would be helpful, but if there isn't (and I think there isn't) then it's not.  But then if we were to switch to an underlying sorted `vector` or an `unordered_set` we're back to "helpful" again.

include/mesh/mesh_base.h
roystgnr: It's possible with the approach I was suggesting (unless we change to `unordered_map`, which is worth considering!), yes.  All that would change is that everywhere you would have read `_elemset_codes[foo]` you instead use `*_elemset_codes[foo]`.

src/mesh/mesh_base.C
roystgnr: Even with the two-maps design we still would go from two set copy constructions down to one.  But yeah, we'd have to make one of those maps into a map-to-pointers to go from one down to zero.

include/mesh/mesh_base.h
jwpeterson: My issue with using a vector is that you are always having to remember to keep it sorted and unique. I don't expect the construction of sets for the purpose of calling `add_elemset_code()` to be a performance bottleneck, so I'm leaning toward the convenience and familiarity of plain old `std::set` currently.

include/mesh/mesh_base.h
roystgnr: Oh, wow, extremely messy; I hadn't considered that.  On the other hand, it means we don't really have to care about how keeping a full `set<elemset_id_type>` would impact it - the difference between O(N_elem+1) and O(N_elem+N_elemsets) should be negligible.

include/mesh/mesh_base.h
roystgnr: It's not just the set construction (that actually works better with `set` than with sorted vectors), it's the lookups - a binary search in a sorted vector can go much faster just from fewer cache misses.

But I can't argue with convenience.  The best alternative I could propose would be to bring Boost's `flat_set` into our `boostcopy` namespace (or into our contrib boost, and then just have our shim fall back on `std::set` for people who disable boost).

include/mesh/mesh_base.h
jwpeterson: OK, so how do you add an entry to a map where the key is a pointer to the value? As mentioned in the other thread, currently I have
```
void MeshBase::add_elemset_code(dof_id_type code, const std::set<elemset_id_type> & id_set)
{
  _elemset_codes.emplace(code, id_set);
  _elemset_codes_inverse_map.emplace(id_set, code);
}
```
are you proposing something like
```
void MeshBase::add_elemset_code(const std::set<elemset_id_type> & id_set)
{
  _elemset_codes.emplace(&id_set, id_set);
}
```
?

jwpeterson: Oh, wait I think I misread your suggestion. 

> Make the values in _elemset_codes be just pointers to the keys in _elemset_codes_inverse_map

So there are still two maps?

roystgnr: Yup, still two maps.  You still `map::emplace` into `_elemset_codes_inverse_map`, but you catch the returned iterator and take the address of the newly emplaced set to use in `_elemset_codes`.

include/mesh/mesh_base.h
jwpeterson: I was not previously familiar with `boost::flat_set`, but it sounds like the best of both worlds. According to some [Benchmarks](https://denisyaroshevskiy.github.io/benchmarking_sets/), for range construction, copying, and erasing entries, boost::flat_set kills std::set and oftentimes std::unordered_set over the entire range of sizes tested. So I think it would be good to have `boost::flat_set` available regardless of whether we end up using it for this specifically. In this PR I will typedef the type of `id_set` (for now to `std::set`) to make it easier to change later.

roystgnr: Sounds like a good compromise - future changes would be technically API-breaking but we'll have given user code a chance at staying forward compatible..  Should we tag this (the add method, I guess) with `libmesh_experimental()` for a while too?

jwpeterson: > Should we tag this (the add method, I guess) with libmesh_experimental() for a while too?

Yeah that's fine with me.

include/mesh/mesh_base.h
jwpeterson: I ended up leaving this one as pass-by-reference for now since I think it's the familiar approach, and its effectiveness does not depend on whatever type we might eventually change `MeshBase::elemset_type` to.

src/mesh/mesh_base.C
jwpeterson: I updated `add_elemset_code` to use pass-by-value semantics.

include/mesh/mesh_base.h
jwpeterson: I implemented the approach of storing a pointer-to-set rather than two copies of the set. I personally don't expect these sets to be "big" or for there to be too many of them in practice... but in case there is I think we have the most efficient storage approach now.

include/mesh/mesh_base.h
jwpeterson: I added a new MeshBase member, `_all_elemset_ids` to support O(1) behavior of `MeshBase::n_elemsets()`.

src/mesh/mesh_base.C
```diff
@@ -104,6 +104,8 @@ MeshBase::MeshBase (const MeshBase & other_mesh) :
   _skip_find_neighbors(other_mesh._skip_find_neighbors),
   _allow_remote_element_removal(other_mesh._allow_remote_element_removal),
   _elem_dims(other_mesh._elem_dims),
+  _elemset_codes(other_mesh._elemset_codes),
```
roystgnr, line 107: This is no longer sufficient if we're storing pointers, though, right?  We need a deep copy here?

src/mesh/mesh_base.C
```diff
@@ -164,6 +166,8 @@ MeshBase& MeshBase::operator= (MeshBase && other_mesh)
   _allow_remote_element_removal = other_mesh.allow_remote_element_removal();
   _block_id_to_name = std::move(other_mesh._block_id_to_name);
   _elem_dims = std::move(other_mesh.elem_dimensions());
+  _elemset_codes = std::move(other_mesh._elemset_codes);
```
roystgnr, line 169: I *think* this should still work correctly in the pointers case, but I'd love some `#ifdef DEBUG` check below just to be certain.

src/mesh/mesh_base.C
jwpeterson: Oh, yeah good catch.

roystgnr (approved):

jwpeterson:
Thanks for reviewing!

PR #3227, opened 2022-04-14
ErrorMap should manage the memory for its ErrorVectors

jwpeterson:
Another cleanup PR that I've had laying around for a while. `ErrorMap` is a public typedef of the `ErrorEstimator` class so technically this is an API change. Practically it was only used internally by a couple of the `ErrorEstimator` classes, so it's possible there's not much code out there (if any) that would be affected by the change.

### DIFF
--- a/include/error_estimation/error_estimator.h
+++ b/include/error_estimation/error_estimator.h
@@ -38,6 +38,7 @@ enum ErrorEstimatorType : int;
 #include <map>
 #include <string>
 #include <vector>
+#include <memory>
 
 namespace libMesh
 {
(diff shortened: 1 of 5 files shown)


jwpeterson:
Test failures are not build related/related to this PR:
```
[5.563s]     DIFF contact/test:mortar_restart.frictional_bouncing_block_action_restart_1 FAILED (EXODIFF) [contact-opt]
[2.987s]     DIFF contact/test:mortar_dynamics.block-dynamics-action FAILED (EXODIFF) [contact-opt]
[2.568s]     DIFF contact/test:mortar_restart.frictional_bouncing_block_action_restart_2 FAILED (EXODIFF) [contact-opt]
[3.517s]     DIFF contact/test:mortar_dynamics.block-dynamics-friction FAILED (EXODIFF) [contact-opt]
[3.529s]     DIFF contact/test:mortar_dynamics.block-dynamics-friction-action FAILED (EXODIFF) [contact-opt]
```

jwpeterson:
The failing tests are being addressed in https://github.com/idaholab/moose/pull/20794

roystgnr (approved):
I was going to complain that this is an API-breaking change ... but it's a very sophisticated API that as far as I can tell is only getting used internally, so I'll bet nobody's affected by the change and if I'm wrong I'll bet anyone who is affected won't have any trouble updating their code to match.

PR #3228, opened 2022-04-14
DTKSolutionTransfer should store std::unique_ptrs

jwpeterson:
This one is not an API change but it's complicated by the fact that I don't have a Trilinos (let alone DTK) build so while the change should be pretty straightforward I'm not sure it actually compiles. Was hoping maybe there's an attachable CIVET recipe that could test this...

### DIFF
--- a/include/solution_transfer/dtk_solution_transfer.h
+++ b/include/solution_transfer/dtk_solution_transfer.h
@@ -24,21 +24,24 @@
 
 #ifdef LIBMESH_TRILINOS_HAVE_DTK
 
+// libMesh includes
 #include "libmesh/solution_transfer.h"
 #include "libmesh/dtk_adapter.h"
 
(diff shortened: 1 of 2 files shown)


jwpeterson:
> Was hoping maybe there's an attachable CIVET recipe that could test this...

Aaand there doesn't seem to be one. I would also be OK with deprecating/eventually deleting this code if it's no longer being used or maintained.

jwpeterson:
Test failures are not related to this PR.

jwpeterson:
The failing tests are being addressed in https://github.com/idaholab/moose/pull/20794

roystgnr (approved):

roystgnr:
We really do need to get Trilinos back in CI.  It's too useful an alternative to deprecate, but it might not have been properly tested since BuildBot days.

PR #3230, opened 2022-04-15
WIP: Some fixes for reading exodus files with BEX elements that cause a segfault

brianmoose:
This contains a minimal exodus file with BEX elements that causes libmesh to segfault.
These are Derek's changes to get it to stop segfaulting and produce sane VTK output.
There is more work to be done on the libmesh side though as the MOOSE results are still not good.
See idaholab/moose#20790


roystgnr:
This is exactly what I was hoping for, thanks.  Didn't get time to dig into the constraint equation issues today, but hopefully early next week.

jwpeterson:
FYI, this didn't test since it was marked WIP.

brianmoose:
:laughing: 
And of course it failed....although that doesn't look like my fault.
I think in the future I will forgo using WIP!

roystgnr:
> FYI, this didn't test since it was marked WIP.

I'm ashamed to admit I didn't actually know or notice that.  In my head "WIP" was simply because we know there are going to be more fixes required along this line, regardless of the current fix being progress.

I pulled the branch to run the new unit test (and look at the ASCII dump of the new test file) myself, and when it worked for me and I was in a hurry to quit for the weekend and I didn't see any big red Xs above the inviting green button.

> And of course it failed....although that doesn't look like my fault.

No, and #3232 looks like it was a simple fix.  But even if it had failed after merging, that would have been my fault, not yours.

roystgnr:
This isn't even sane output at this stage, I'm afraid.

The immediate problem:

If you look at the actual Quad9 nodes we're creating here, there are a ton of duplicate nodes sharing effectively the same points - and they're not getting caught, because in libMesh "duplicate node" never means "we see you have a bug, we should scream", it means "gosh, you must want a slit in your mesh here; we'll make sure the element neighbor topology reflects that for you".  (and this is a deliberate decision; you can run interesting problems on slit meshes and we have unit tests to keep that from regressing)

The first duplicates I see, and the equations that constructed them from spline nodes:

```
Node 4 at (x,y,z)=(-1.20043e-15, -5.82867e-16,        0) = 0.5 * (x,y,z)=(-1.85962e-15,      0.5,        0) + 0.5 * (x,y,z)=(-5.41234e-16,     -0.5,        0)
Node 11 at (x,y,z)=(-5.20417e-16, -2.08167e-16,        0) = 0.125 * (x,y,z)=(    -0.5,      0.5,        0) + 0.25 * (x,y,z)=(-1.85962e-15,      0.5,        0) + 0.125 * (x,y,z)=(     0.5,      0.5,        0) + 0.125 * (x,y,z)=(    -0.5,     -0.5,        0) + 0.25 * (x,y,z)=(-5.41234e-16,     -0.5,        0) + 0.125 * (x,y,z)=(     0.5,     -0.5,        0)
```

I'm assuming these were intended to be the same node, but in our BEXT reader code right now we use those equations as a key to identify identical nodes from different elements (see `local_nodes.find(key)` around exodusII_io.C:517), and two different equations gets interpreted as two different nodes.  If we're going to get files where that isn't accurate, we need some other way to identify-or-distinguish nodes.  "Look through an octree for an existing node nearby, call it the same node if it's near enough", for example, but that would make impossible to read BEXT meshes with actual slits, so hopefully someone else has a better idea.

The longer-term problem:

This file looks like it builds a non-conforming level-0 mesh: one big quad at the bottom, 4 small quads at the top, no parent element for the 4 small quads.  There are probably 50 places scattered randomly around `*/*/*.{C,h}` in libMesh (and maybe more in Moose?) where that's going to break, and I'm not sure how to find them all, except one-by-one as we try to use a mesh like this.

In this particular case, we could avoid the problem by just giving the 4 small quads a parent element (though figuring out how to autodetect if/when some need it is a trickier question), but would backfilling a refinement tree like that work for every mesh we want to read, not just this simplified test case?

PR #3231, opened 2022-04-18
Save old adjoint vectors

vikramvgarg:
This branch gives TimeSolver the capability to save old adjoint solutions during the adjoint time march, like old_nonlinear_solution saves the primal solutions during the primal time march.

Detailed steps:
1) There are multiple adjoint solutions, hence the interface between user code and System was extended to allow the user to set the number of qois (System::init_qois) and hence the number of adjoint related vectors via System::init_adjoints.
2) System members qoi and qoi_error_estimates which were public earlier have now been made private, and accessors added.
3) UnsteadySolver and AdaptiveTimeSolver implementations of adjoint_advance_timestep save all old_adjoint_vectors.

### DIFF
--- a/include/systems/system.h
+++ b/include/systems/system.h
@@ -1603,10 +1603,10 @@ public:
    */
   void init_qois(unsigned int n_qois);
 
-  void fill_qoi(unsigned int qoi_index, Real qoi_value);
+  void fill_qoi(unsigned int qoi_index, Number qoi_value);
   Real get_qoi_value(unsigned int qoi_index);
 
-  void fill_qoi_error_estimate(unsigned int qoi_index, Real qoi_error_estimate);
+  void fill_qoi_error_estimate(unsigned int qoi_index, Number qoi_error_estimate);
   Real get_qoi_error_estimate_value(unsigned int qoi_index);
 
   /**
--- a/include/systems/system.h
+++ b/include/systems/system.h
@@ -1581,6 +1581,7 @@ public:
    */
   unsigned int n_qois() const;
 
+  #ifdef LIBMESH_ENABLE_DEPRECATED
   /**
    * Values of the quantities of interest.  This vector needs
    * to be both resized and filled by the user before any quantity of
@@ -1596,6 +1597,17 @@ public:
    * User code can use this for accumulating error estimates for example.
    */
   std::vector<Number> qoi_error_estimates;
+  #endif
+
+  /** Accessors for qoi and qoi_error_estimates vectors
+   */
+  void init_qois(unsigned int n_qois);
+
+  void fill_qoi(unsigned int qoi_index, Real qoi_value);
+  Real get_qoi_value(unsigned int qoi_index);
+
+  void fill_qoi_error_estimate(unsigned int qoi_index, Real qoi_error_estimate);
+  Real get_qoi_error_estimate_value(unsigned int qoi_index);
 
   /**
    * \returns The value of the solution variable \p var at the physical
@@ -2214,6 +2226,24 @@ private:
    * Do we want to apply constraints while projecting vectors ?
    */
   bool project_with_constraints;
+
+  #ifndef LIBMESH_ENABLE_DEPRECATED
+  /**
+   * Values of the quantities of interest.  This vector needs
+   * to be both resized and filled by the user before any quantity of
+   * interest assembly is done and before any sensitivities are
+   * calculated.
+   */
+  std::vector<Number> qoi;
+
+  /**
+   * Vector to hold error estimates for qois, either from a steady
+   * state calculation, or from a single unsteady solver timestep. Used
+   * by the library after resizing to match the size of the qoi vector.
+   * User code can use this for accumulating error estimates for example.
+   */
+  std::vector<Number> qoi_error_estimates;
+  #endif
 };
 
 
--- a/include/systems/system.h
+++ b/include/systems/system.h
@@ -1603,10 +1603,10 @@ public:
    */
   void init_qois(unsigned int n_qois);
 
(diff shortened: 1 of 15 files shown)


include/systems/system.h
```diff
@@ -1596,6 +1597,17 @@ class System : public ReferenceCountedObject<System>,
    * User code can use this for accumulating error estimates for example.
    */
   std::vector<Number> qoi_error_estimates;
+  #endif
+
+  /** Accessors for qoi and qoi_error_estimates vectors
+   */
+  void init_qois(unsigned int n_qois);
+
+  void fill_qoi(unsigned int qoi_index, Real qoi_value);
```
roystgnr, line 1606: This ought to be taking Number, not Real.

include/systems/system.h
```diff
@@ -1596,6 +1597,17 @@ class System : public ReferenceCountedObject<System>,
...
+  Real get_qoi_value(unsigned int qoi_index);
+
+  void fill_qoi_error_estimate(unsigned int qoi_index, Real qoi_error_estimate);
```
roystgnr, line 1609: This is slightly more interesting.  We want to be able to store directional error estimates here, e.g. for refinement-based estimators that can actually give us more than just a bound?  If so then the API fix is simple, Real->Number again.

src/systems/system.C
```diff
@@ -2144,6 +2144,38 @@ void System::user_QOI_derivative(const QoISet & qoi_indices,
       (qoi_indices, include_liftfunc, apply_constraints);
 }
 
+void System::init_qois(unsigned int n_qois)
+{
+  qoi.resize(n_qois);
+  qoi_error_estimates.resize(n_qois);
+}
+
+void System::fill_qoi(unsigned int qoi_index, Real qoi_value)
```
roystgnr: Let's use "set" rather than "fill" for this and for the setter below?  More consistent with other library setters.

src/solvers/file_solution_history.C
```diff
@@ -164,19 +165,27 @@ void FileSolutionHistory::retrieve(bool is_adjoint_solve, Real time)
   if(is_adjoint_solve)
   {
     // Reading in the primal xdas overwrites the adjoint solution with zero
-    // So swap to retain the old adjoint solution
+    // So swap to retain the adjoint and old adjoint vectors
     for (auto j : make_range(_system.n_qois()))
     {
       dual_solution_copies[j] = _system.get_adjoint_solution(j).clone();
+
+      std::string old_adjoint_solution_name = "_old_adjoint_solution";
```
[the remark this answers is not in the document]
roystgnr: Weird, when I reload the github page I'm not seeing the comment that this comment is referencing.  For the record my other complaint was about History classes assuming the number of adjoint solution vectors that TimeSolver classes need.  When we go higher order we'll be adding more vectors, so we'll need some interface to let the TimeSolver classes communicate the complete list, we can't just hard-code it.

This is a big enough advance that it doesn't need to be fixed in this PR, we just need to start thinking about it before we dig ourselves into some hole where it can't eventually be fixed.

roystgnr:
That's everything I caught.  The History<->TimeSolver interaction is the biggest thing to worry about for the future, but I think the API name and the suboptimal assertion coverage are probably all we need to change before this stage is ready to merge.

roystgnr (approved):
Looks good to me.  We'll merge tomorrow if there are no other comments by then.

jwpeterson:
I'm getting the following error since this was merged:
```
In file included from ./include/libmesh/fem_system.h:24:0,
                 from /home/jwpeterson/software/libmesh-src/src/error_estimation/adjoint_refinement_estimator.C:42:
./include/libmesh/diff_system.h: In member function ‘void libMesh::DifferentiableSystem::attach_qoi(libMesh::DifferentiableQoI*)’:
./include/libmesh/diff_system.h:230:37: error: ‘std::vector<double> libMesh::System::qoi’ is private within this context
     this->diff_qoi->init_qoi( this->qoi );}
                                     ^~~
In file included from ./include/libmesh/equation_systems.h:26:0,
                 from /home/jwpeterson/software/libmesh-src/src/error_estimation/adjoint_refinement_estimator.C:30:
./include/libmesh/system.h:2243:23: note: declared private here
   std::vector<Number> qoi;
                       ^~~
```
I configure and build with `--disable-deprecated` so it may be related to that.

jwpeterson:
The `qoi` and `qoi_error_estimates` members were both wrapped in `#ifndef LIBMESH_ENABLE_DEPRECATED` in 85e81b6928. Not sure what the right fix is... should `DiffSystem::attach_qoi()` become deprecated as well?

roystgnr:
Yeah, I just noticed this blocking the devel->master merge too.  Trying to figure out the fix now.

jwpeterson:
OK, we actually have two copies of this code in system.h
```
  /**
   * Values of the quantities of interest.  This vector needs
   * to be both resized and filled by the user before any quantity of
   * interest assembly is done and before any sensitivities are
   * calculated.
   */
  std::vector<Number> qoi;

  /**
   * Vector to hold error estimates for qois, either from a steady
   * state calculation, or from a single unsteady solver timestep. Used
   * by the library after resizing to match the size of the qoi vector.
   * User code can use this for accumulating error estimates for example.
   */
  std::vector<Number> qoi_error_estimates;
```
one wrapped in `#ifdef LIBMESH_ENABLE_DEPRECATED` and one protected one wrapped in `#ifndef`... this seems really confusing and error prone. Need to have one copy with just public/protected wrapped in ifdefs for starters.

jwpeterson:
A possible quick fix may be to just not deprecate these two variables yet, see jwpeterson/libMesh@04e94e307

PR #3233, opened 2022-04-19
ExodusII::write_added_sides()

roystgnr:
This replaces #3188 - it starts with that branch, but adds data output as well.

I still have more testing to do before merging, but I'm again at a stage where double-checking that CI is happy would be good.

### DIFF
--- a/src/fe/fe_hermite.C
+++ b/src/fe/fe_hermite.C
@@ -273,8 +273,9 @@ unsigned int hermite_n_dofs_per_elem(const ElemType t,
 } // anonymous namespace
 
 
-// Instantiate nodal_soln() function for every dimension
+// Instantiate (side_) nodal_soln() function for every dimension
 LIBMESH_FE_NODAL_SOLN(HERMITE, hermite_nodal_soln)
+LIBMESH_FE_SIDE_NODAL_SOLN(HERMITE)
 
 
 // Do full-specialization for every dimension, instead
--- a/src/fe/fe_hermite.C
+++ b/src/fe/fe_hermite.C
@@ -19,9 +19,10 @@
 
 // Local includes
 #include "libmesh/elem.h"
+#include "libmesh/enum_to_string.h"
 #include "libmesh/fe.h"
 #include "libmesh/fe_interface.h"
-#include "libmesh/enum_to_string.h"
+#include "libmesh/fe_macro.h"
 
 namespace libMesh
 {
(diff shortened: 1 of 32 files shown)


jwpeterson:
Looks like an actual failure in misc_ex5?
```
* Running Example miscellaneous_ex5:
*  mpiexec -np 2 ./example-dbg  --n_threads=2
 
  Beginning Solve 0
Number of elements: 219
 assembling elliptic dg system... done
System has: 768 degrees of freedom.
Linear solver converged at step: 31, final residual: 4.91472e-12
L2-Error is: 0.00666744
  Beginning Solve 1
Number of elements: 827
 assembling elliptic dg system... done
System has: 2896 degrees of freedom.
Linear solver converged at step: 46, final residual: 7.22623e-12
L2-Error is: 0.00264921
  Beginning Solve 2
Number of elements: 3003
 assembling elliptic dg system... done
System has: 10512 degrees of freedom.
Linear solver converged at step: 65, final residual: 1.95985e-11
L2-Error is: 0.0016323
Assertion `num_real_nodes == num_nodes' failed.
num_real_nodes = 21024
num_nodes = 0


[1] ../src/mesh/exodusII_io.C, line 2096, compiled Apr 19 2022 at 15:11:19
application called MPI_Abort(MPI_COMM_WORLD, 1) - process 1
FAIL: run.sh
===============================================================
```
This example calls `ExodusII_IO::write_discontinuous_exodusII()` though, so maybe that makes sense?

jwpeterson (commented):
Looks good to me so far, but I was expecting you to use the `EX_FACE_BLOCK` or `EX_EDGE_BLOCK` support which is in more recent versions of Exodus in some way...  also curious if you have looked at the result in Paraview yet?

roystgnr:
That's definitely an actual failure in misc_ex5, but I'm still trying to replicate it.

roystgnr:
Never mind; it's easy to replicate in parallel.

roystgnr:
And easy to fix; it wasn't an actual failure, just an assertion getting called in parallel that should have only been called on proc 0.

roystgnr:
That MOOSE thing is python acting up again, but the complex configuration's an actual bug in the new unit tests.

tests/mesh/mesh_input.C
```diff
@@ -45,6 +45,54 @@ Number sin_x_plus_cos_y (const Point& p,
...
+
+
+  // x/y/z components of a div-free flux,
+  // curl([x^2yz, xy^2z, xyz])
+  if (is_on_face(x, 0))
+    {
+      libmesh_assert(!is_on_face(y, 1));
+      libmesh_assert(!is_on_face(z, 2));
+//      return (x*z-x*y*y);
+return 3000+100*x+10*y+z;
```
jwpeterson, line 82: Does this line have weird indentation for any particular reason? Also maybe drop the commented out line if it isn't needed or was part of some debugging work, etc.

src/fe/fe_hermite.C
```diff
@@ -272,41 +273,9 @@ unsigned int hermite_n_dofs_per_elem(const ElemType t,
...
-template <>
-void FE<3,HERMITE>::nodal_soln(const Elem * elem,
-                               const Order order,
-                               const std::vector<Number> & elem_soln,
-                               std::vector<Number> & nodal_soln)
-{ hermite_nodal_soln(elem, order, elem_soln, nodal_soln); }
-
-
+// Instantiate (side_) nodal_soln() function for every dimension
+LIBMESH_FE_NODAL_SOLN(HERMITE, hermite_nodal_soln)
```
jwpeterson, line 277: This PR seems to have some less related FE refactoring in it, e.g. macro-izing `FE::nodal_soln` instantiations, that might have worked well as a separate PR, but probably too painful to split up at this point.

jwpeterson (approved):
I'm not familiar enough with the details of what you are doing (and it is too big) to review this really well... but looks good to me on a cursory inspection, has tests, and passes existing tests, so how bad can it be?

tests/mesh/mesh_input.C
roystgnr: Just the opposite - I use completely-unindented lines for debugging work, on the theory that they stick out so badly I won't fail to notice them when I'm skimming over `git diff --staged` before a commit.  This theory is not always correct, clearly.

roystgnr:
Well, that was scary.  After I restored the regular test instead of the debugging test, it started failing on Hexes.  Turned out to be a bug in the test rather than the implementation, though.  After I finish hammering it on more processors and distributed meshes and so on I'll push, then I'll merge once CI is happy too.

PR #3235, opened 2022-04-19
Another round of structured bindings updates

jwpeterson:
Like last time, should be no functionality changes/only prettier syntax changes.


roystgnr (approved):
Removes more lines than it adds, adds much more readability than it removes.  I love these structured binding diffs.

jwpeterson:
Thanks for reviewing!

PR #3236, opened 2022-04-19
L2System should store unique_ptrs

jwpeterson:
This is a change to a public member, thus an API change, but it is a class which I think is only used by some of the codes in apps?


roystgnr (approved):

roystgnr:
> it is a class which I think is only used by some of the codes in apps

Correct - this doesn't even end up in libmesh.so

PR #3238, opened 2022-04-20
Use unique_ptr in DofMap::merge_ghost_functor_outputs() API

jwpeterson:
An API change, so if necessary, we could implement a backwards-compatible shim for this. If nothing in MOOSE calls this directly I think we should probably be OK to skip that. Also, this uses a C++14 capability related to the `is_transparent` type, but @roystgnr and I tried it on a bunch of older compilers and it seems to be pretty well-supported.

### DIFF
--- a/include/utils/utility.h
+++ b/include/utils/utility.h
@@ -19,13 +19,14 @@
 #ifndef LIBMESH_UTILITY_H
 #define LIBMESH_UTILITY_H
 
-// Local includes
+// LibMesh includes
 #include "libmesh/libmesh_common.h" // for Real
 
-// System includes
+// C++ includes
 #include <string>
 #include <vector>
 #include <algorithm> // is_sorted, lower_bound
+#include <memory> // unique_ptr
 
 namespace libMesh
 {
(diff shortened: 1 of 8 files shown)


jwpeterson:
The failure here in `mesh/custom_partitioner.group/custom_linear_partitioner_displacement` definitely seems real and related to my changes, but I'm not exactly sure what the issue would be.
```
../src/base/dof_map.C, line 1587, compiled Apr 20 2022 at 10:58:462
Assertion `ghost_coupling->size() == n_var' failed.
ghost_coupling->size() = 2
n_var = 1
```

include/utils/utility.h
```diff
@@ -464,6 +465,48 @@ T ReverseBytes::operator() (T & data) const
 }
 
 
+
+/**
+ * Struct which defines a custom comparison object that
+ * can be used with std::sets of std::unique_ptrs
+ */
+template <class T>
```
roystgnr, line 473: Do we need this to be templated around T?  Or would templating the member functions work?

roystgnr:
It looks to me like a find() is returning end() (and thus prompting us to create a duplicate of an existing functor) rather than actually finding what it should, but I don't see how.

Maybe we ought to have a couple unit tests for the new comparator?  This seems like the sort of thing we really want to catch at the start of "make check", not deep into ghosting functor operations.

include/utils/utility.h
jwpeterson: I don't know actually... I just assumed this was how you were supposed to do it, but I can check...

jwpeterson:
> Maybe we ought to have a couple unit tests for the new comparator?

OK. I also (force-pushed) a change since the last failure which might fix the problem. I was trying to do fewer `const_casts` but I think I ended up introducing a bug... not sure but the CI will let us know.

include/utils/utility.h
jwpeterson: OK, I tried templating just the member functions and it still worked. This makes the `typedef` declaration nicer since you don't have to repeat the type twice... so I'll go ahead and make that change.

jwpeterson:
OK, nice, there was a real bug found by that MOOSE test but it is now fixed. 

I can add some unit tests for this. I have the test code that I used to develop everything initially, plus I expanded upon that to verif that stuff like the following works as expected (although not something that can be used safely in real code):
```
// Test that inserting the same underlying pointer (after moving) fails
std::set<std::unique_ptr<int>, CompareUnderlying> s3;
auto four = std::make_unique<int>(4);
s3.insert(std::move(four)); // steals resources of four
s3.insert(std::move(four)); // inserts nullptr underlying value, moved-from unique_ptr is guaranteed to be nullptr 
s3.insert(std::move(four)); // verify that insert does not happen, nullptr underlying value already exists
```

roystgnr (approved):

roystgnr:
I'd feel safer with unit tests but this is still a big improvement as-is.

PR #3239, opened 2022-04-20
Replace deprecated Exodus API calls with non-deprecated versions

jwpeterson:
I split this up quite a bit so that if anything fails, it will hopefully be easy to bisect back to the breaking commit. If everything passes I could be convinced to squash the whole thing down into a single commit.

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -3351,8 +3351,9 @@ read_sideset_data(const MeshBase & mesh,
     {
       // Read the sideset data truth table
       std::vector<int> sset_var_tab(num_side_sets * num_sideset_vars);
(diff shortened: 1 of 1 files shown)


roystgnr (approved):
I like it better split up, personally.

jwpeterson:
I ran this command from a test repository on latest devel just to see what would happen:
```
git merge --no-ff roystgnr/exodus_faces_soln jwpeterson/fix_exodus_deprecated
```
and there was a single, relatively trivial looking, merge conflict in exodusII_io_helper.C. If you want I can wait until #3233 is merged before attempting to merge this.

roystgnr:
Nah, let's merge this whenever you're ready.  It's much more straightforward and I don't want to hold it up; #3233 is passing tests now but I'm not yet sure it's *working* working.

PR #3245, opened 2022-04-23
Eigen update to 3.4.0

### LINKED ISSUE #3244: Update libEigen (deprecated `std::result_of`)
Eigen 3.3.9 contains a single use of `std::result_of`, which has been deprecated in C++17. This causes a MOOSE test to barf.

dschwen:
Closes #3244


dschwen:
Ugh, I'll make an alternative PR with just a small backport of the C++17 fix that's in 3.4.0 to 3.3.9. Figuring out the isnan issue might take a while (and possibly fixes to other repos).

roystgnr:
MetaPhysicL needs to be updated to a bootstrapped version, not to master.  I'll update our bootstrapped branch there now.

roystgnr:
Okay, try updating the MetaPhysicL submodule to https://github.com/libMesh/MetaPhysicL/commit/29b776c77b0098c320509877af524603231e3814

dschwen:
Hm, looks like there is another issue that we need to take care of.

dschwen:
All green now!

roystgnr (approved):
I'm happy with this, but with enough shock at "+231,922 −196,029" that I'd like to wait for @jwpeterson to have a chance to chime in before hitting the button.  My text input box is *lagging* as I type right now, the "files changed" page is so big

roystgnr:
Also ... while we're thinking about this ... should we just make contrib/eigen a git submodule?  We crossed the "mandatory dependency in a submodule, everybody has to init submodules" threshold with TIMPI, which was the biggest obstacle in the past.  IIRC we don't have any submodules hosted on gitlab rather than github, but I don't see how that would be a problem.

dschwen:
Yeah, submodule is not a bad idea. But how would that work with the custom Makefile?

roystgnr:
With poly2tri we actually needed custom header changes, so I've got some sed magic that creates contrib/poly2tri/modified/ files from the pristine contrib/poly2tri/poly2tri/ repo.  A custom Makefile.am in another directory can't be a worse hack than that.

dschwen:
already working on it

dschwen:
Well, this naive approach does not work, because `build/contrib/eigen` only contains the `Makefile` but not the `eigen` subdir.

dschwen:
┻━┻︵ \(°□°)/ ︵ ┻━┻

I don't get why the eigen folder is not copied into the build dir. What am I missing? Why does this "just work" with poly2tri for example?

jwpeterson:
> for @jwpeterson to have a chance to chime in before hitting the button. My text input box is lagging as I type right now, the "files changed" page is so big

We recently had that issue with the poly2tri submodule where it took longer to get a fix merged into upstream than you wanted, and it almost caused you to fork poly2tri on GitHub, push the fix to your fork, and change the URL in our .gitmodules, right? I guess in the end it didn't come to that, but I would worry about the same thing possibly happening with Eigen. 

I would also be a bit concerned with using a hard-coded https:// URL in .gitmodules, but only because we haven't had to do this previously. I'm guessing it would probably be fine unless there was some build machine that only allowed ssh traffic for some reason? 

I think the benefits of not having +/-200k line commits in the history probably outweighs either of these concerns, and both could also be addressed if we wanted to add a libMesh/eigen mirror on GitHub... then we could push our own temporary patches there and continue using relative URLs in .gitmodules.

jwpeterson:
@dschwen just a guess but in contrib/Makefile.am we have this:
```
if LIBMESH_INSTALL_INTERNAL_EIGEN
   SUBDIRS += eigen/eigen
endif
```
Are you trying to remove one level of "eigen" from the path?

dschwen:
@jwpeterson no, what was previously the `contrib/eigen/eigen` symling to the eigen copy (e.g. `3.3.9`) is now teh submodule. So everything stays at teh same level. _But_ I cannot add a `Makefile.(am|in)` in that directory anymore as it is now just the upstream  Eigen. So I figured I could pull those automake files out of that level and put them next to the submodule as `contrib/eigen/Makefile*`. When running configure the corresponding `Makefile` is added in `build/contrib/eigen/Makefile`, but `build/contrib/eigen/eigen` is missing.

roystgnr:
> almost caused you to fork poly2tri on GitHub, push the fix to your fork, and change the URL in our .gitmodules, right? I guess in the end it didn't come to that

No almost - it did come to that!  In hindsight it didn't *have to* come to that, because they merged my PR the following week, but for about 6 days we were using a forked poly2tri.  Didn't actually cause us any hiccups.  IMHO even if we had to maintain a fork longer-term, it might be easier to do so via a forked submodule than via our sad "cp -r and also there's been some files changed, try to guess which!" alternative.

> I would also be a bit concerned with using a hard-coded https:// URL in .gitmodules

Non-relative URLs, on the other hand, I do recall causing us hiccups at one point.  But I think it was a problem of the form "trying to hardcode use of ssh broke someone who was relying on https", and I wouldn't expect to see any trouble going the other way around.

PR #3246, opened 2022-04-24
std::result of is deprecated in C++17 

dschwen:
Refs #3244


roystgnr (approved):
I really really hate forking upstream, but under the circumstances this seems like the best short-term option.

roystgnr:
Oh, was isinf the *only* headache you ran into?  We can just push the MetaPhysicL fix downstream right now, then.

jwpeterson:
Danke Schwen.

(I just thought of this great German pun.)

PR #3249, opened 2022-04-26
Don't lock in libmesh_singleton.C

lindsayad:
I think we should be creating all of our singletons in non-threaded
code regions so it shouldn't be necessary to lock. Moreover, we
got a user report of a deadlock due to locking in this file. Testing
MOOSE simple diffusion with these changes reports as thread sanitizer
clean

### DIFF
--- a/src/base/libmesh_singleton.C
+++ b/src/base/libmesh_singleton.C
@@ -32,10 +32,6 @@ namespace
 {
 using namespace libMesh;
 
-// Mutex object for required locking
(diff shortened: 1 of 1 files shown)


roystgnr (approved):
This is probably the best way to avoid the Static Initialization Order fiasco for now.  And as long as we're always doing Singleton setup in LibMeshInit I don't see how we'll run into trouble in the future either.

PR #3251, opened 2022-04-27
Let Poly2TriTriangulator users disallow boundary refinement

roystgnr:
We can't stitch meshes unless boundary sides line up, and we might want to stitch triangulator output to other meshes, so we should give triangulator users finer control over how to refine boundary sides.

### DIFF
--- a/include/mesh/mesh_triangle_holes.h
+++ b/include/mesh/mesh_triangle_holes.h
@@ -87,6 +87,30 @@ public:
     seg.push_back(n_points());
     return seg;
   }
+
+  /**
+   * Set whether or not a triangulator is allowed to refine the
+   * hole boundary when refining the mesh interior.  This is true by
+   * default, but may be set to false to make the hole boundary more
+   * predictable (and so easier to stitch to other meshes) later.
+   */
+  virtual void set_refine_boundary_allowed (bool refine_bdy_allowed)
+  { _refine_bdy_allowed = refine_bdy_allowed; }
+
+  /**
+   * Get whether or not the triangulation is allowed to refine the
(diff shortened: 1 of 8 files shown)


roystgnr:
Hm.  I was under the impression that return value copy elision was mandatory in C++17, but apparently gcc 7.5 wasn't quite there yet.  Good thing we learned about this from a unit test, not serious code.

@vikramvgarg - you put in a ton of work to define move-assignment methods for the Mesh hierarchy, but we still mark the move-constructors as deleted.  Was there a reason we couldn't define those too, or was it just that we never expected to need the capability so there was no point in spending even more time implementing it?

I can't think of any reason why "empty construct, then move-assign" wouldn't be good enough here.  I'm actually kind of tempted to stubbornly add that rather than downgrade my unit test.

vikramvgarg:
> Hm. I was under the impression that return value copy elision was mandatory in C++17, but apparently gcc 7.5 wasn't quite there yet. Good thing we learned about this from a unit test, not serious code.
> 
> @vikramvgarg - you put in a ton of work to define move-assignment methods for the Mesh hierarchy, but we still mark the move-constructors as deleted. Was there a reason we couldn't define those too, or was it just that we never expected to need the capability so there was no point in spending even more time implementing it?
> 
> I can't think of any reason why "empty construct, then move-assign" wouldn't be good enough here. I'm actually kind of tempted to stubbornly add that rather than downgrade my unit test.

I think we simply overlooked the move constructor. The comment for it is out of date. I cant see a reason why we cant have a move constructor, although I am struggling to think of a use case.

PR #3252, opened 2022-04-27
Fix up raw-QoI-access deprecation

roystgnr:
The hard part of this is the need to change the DifferentiableQoI::init_qoi interface, but I don't see any other forwards-compatible way to avoid that.

### DIFF
--- a/examples/adjoints/adjoints_ex7/heatsystem.h
+++ b/examples/adjoints/adjoints_ex7/heatsystem.h
@@ -38,7 +38,7 @@ public:
       _fe_family("LAGRANGE"),
       _fe_order(1),
       _analytic_jacobians(true)
-  { qoi.resize(2); QoI_time_instant.resize(2);}
+  { this->init_qois(2); QoI_time_instant.resize(2);}
 
   Real & k() { return _k; }
   bool & analytic_jacobians() { return _analytic_jacobians; }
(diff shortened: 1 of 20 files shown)


jwpeterson:
Looks like a build error in one of the examples:
```
../../../../examples/adjoints/adjoints_ex2/L-qoi.C:8:6: error: member access into incomplete type 'libMesh::System'
  sys.init_qois(1);
     ^
../../../include/libmesh/diff_context.h:41:7: note: forward declaration of 'libMesh::System'
class System;
      ^
1 error generated.
```

I also attached a "Test No Deprecated" to this PR, so we'll know for sure whether that's passing.

examples/adjoints/adjoints_ex7/heatsystem.h
```diff
@@ -38,7 +38,7 @@ class HeatSystem : public FEMSystem
       _fe_family("LAGRANGE"),
       _fe_order(1),
       _analytic_jacobians(true)
-  { qoi.resize(2); QoI_time_instant.resize(2);}
+  { this->init_qois(2); QoI_time_instant.resize(2);}
```
vikramvgarg, line 41: Thanks for catching and fixing all of these.

vikramvgarg (approved):
Looks good to me.

jwpeterson (approved):

PR #3258, opened 2022-05-03
MeshedHole

roystgnr:
This should make it easier to construct triangulated meshes that can be stitched to other meshes, or that are defined via other mesh generation utilities.

### DIFF
--- a/src/mesh/mesh_triangle_holes.C
+++ b/src/mesh/mesh_triangle_holes.C
@@ -21,7 +21,66 @@
 // Local includes
 #include "libmesh/mesh_triangle_holes.h"
 
+#include "libmesh/boundary_info.h"
+#include "libmesh/elem.h"
 #include "libmesh/int_range.h"
+#include "libmesh/mesh_base.h"
+#include "libmesh/node.h"
+#include "libmesh/simple_range.h"
+
+namespace
+{
+  using namespace libMesh;
+
+  // Returns a positive distance iff the ray from source in the
+  // direction of ray_target intersects the given edge, -1 otherwise
+  Real find_intersection(const Point & source,
+                         const Point & ray_target,
+                         const Point & edge_pt0,
+                         const Point & edge_pt1)
+  {
+    // Calculate intersection parameters (fractions of the distance
+    // along each segment)
+    const Real raydx = ray_target(0)-source(0),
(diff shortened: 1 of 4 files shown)


roystgnr:
No `std::reduce()` from libstdc++ in GCC 7.5, huh?  Damn, that's not until GCC 9 according to https://en.cppreference.com/w/cpp/compiler_support/17 (the Parallelism TS row) - and clang still shows up red there, even, though our Min clang seems happy to compile it here.

I'd been thinking of taking a pass through our code and rewriting All The Things to use the new algorithms, so we could at least experiment a bit with threading them, but I guess instead I'll be going back and replacing `reduce()` with a hand-written loop like a caveman.

That `Test debug` failure looks more serious.  Wonder if I screwed something up in parallel.

src/mesh/mesh_triangle_holes.C
```diff
@@ -159,14 +227,227 @@ Point TriangulatorInterface::ArbitraryHole::point(const unsigned int n) const
...
+  // This is expensive to compute, so only do it when we first need it
+  if (_center(0) == std::numeric_limits<Real>::max())
+    {
+      // Start with the vertex average
+
+      // Turns out "I'm a fully compliant C++17 compiler!" doesn't
+      // mean "I have a full C++17 standard library!"
+      // _center = std::reduce(_points.begin(), _points.end());
+      _center = 0;
+      for (auto p : _points)
```
jwpeterson, line 390: FYI, I tested and 
```
  std::vector<Point> v = {Point(0,0), Point(1,0), Point(1,1), Point(0,1)};
  Point result = std::accumulate(v.begin(), v.end(), Point());
```
compiles and does the right thing on my ancient compiler.

roystgnr: You know, I got so wrapped up in the new exciting parallelizable algorithms that I completely forgot that one.  Good idea.

roystgnr:
I kinda want to make inline `libmesh_` versions of `reduce()`, `transform_reduce()`, everything in that Parallelism TS now.  Serial implementation on older compilers, std:: call with configurable ExecutionPolicy on newer ones.  Probably not worth it unless we have benchmarks that clearly hit a hotspot inside such loops, though.

PR #3262, opened 2022-05-04
Enable stitching with (serialized) DistributedMesh

roystgnr:
I still hope that one of these days we'll do this more efficiently than by serializing, but in the meantime let's at least make this possible.


roystgnr:
There's a bug in that same module that we're trying to hunt down ... but I think it's a completely unrelated bug, and these are real failures here.

PR #3263, opened 2022-05-04
EIM on nodesets

dknez:
We previously implemented EIM on element interiors and sidesets. This PR adds EIM on nodesets. This is useful for things like point loads.


jwpeterson:
This passes our internal testing in both serial and parallel so it should be good to go now.

jwpeterson (approved):

PR #3264, opened 2022-05-04
Drop long-commented-out code in Elem::connectivity() functions

jwpeterson:
Most (if not all) of our Elems have had VTK counterparts for some time now, so it is no longer necessary to keep big commented-out blocks of code around for writing out "sub-elements" to VTK files. I'm not even sure it was all that well-tested to begin with...


roystgnr (approved):

roystgnr:
> +30 −551

simpsons-goodbye

PR #3265, opened 2022-05-04
Allow Edge/Face elements to define polylines to triangulate

roystgnr:
This could be useful for doing fully-anisotropic adaptivity (in 2D, anyway...) in the future, but for the present it's just to add a bit of user-friendliness.


jwpeterson (approved):
Might be worth having a `p2t::Point(const Node & n)` "factory" function that asserts the z-coordinate of `n` == 0. It looks like you are checking this in one place but not others.

PR #3267, opened 2022-05-05
System variable numbers: ushort -> uint

roystgnr:
Yaqi was seeing an error when he tested out adding more than 2^16
variables to a system; this change shouldn't cost much and it should get
us up to 2^24 (due to the ncv_magic packing in DofObject) for 32-bit
dof_id_type builds and 2^32 (due to unsigned int everywhere) for 64-bit
dof_id_type builds.

This should fix compatibility; I can't even imagine what problems we
might find with efficiency at those counts.

### DIFF
--- a/include/systems/system.h
+++ b/include/systems/system.h
@@ -1180,7 +1180,7 @@ public:
    * \returns The variable number associated with
    * the user-specified variable named \p var.
    */
-  unsigned short int variable_number (std::string_view var) const;
+  unsigned int variable_number (std::string_view var) const;
 
   /**
    * Fills \p all_variable_numbers with all the variable numbers for the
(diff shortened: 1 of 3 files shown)


jwpeterson:
65535 variables ought to be enough for anyone. No really, it should be enough.

roystgnr:
The long-term solution here ought to be array-valued variables, but that'll be a bigger change than this was, just just because of the API changes but because the ncv_magic nitty-gritty currently assumes no variable type has more than 256 components at a single DofObject.  Probably just need to make that 2^(8*sizeof(dof_id_type)/2) instead of a fixed 2^8, but who knows what other weirdness is lurking under other `private:` hoods.

roystgnr:
"My problem was able to get past where it was crashing before, so that definitely fixes the error I had"

PR #3268, opened 2022-05-05
We can just use DirichletBoundary objects, no need to use new/delete

jwpeterson:
Came across this much simpler update while investigating whether we could smart-pointify `_diff_physics`, which turned out to be much more complicated/messy.


roystgnr (approved):
Looks good.  Should we also (at least eventually) be changing DirichletBoundaries to be a container of `unique_ptr<DB>`?

jwpeterson:
Yes, it is on my radar.

PR #3272, opened 2022-05-09
allow reading of Abaqus mesh containing TRI6 and QUAD8 elements

jessecarterMOOSE:
As discussed [here](https://github.com/libMesh/libmesh/discussions/3259), libmesh is missing support for reading in some higher-order 2D Abaqus elements. The node and side ordering are the same for libmesh and Abaqus for TRI6 and QUAD8: [see bottom of page here](https://abaqus-docs.mit.edu/2017/English/SIMACAEELMRefMap/simaelm-r-2delem.htm), so it is straight-forward to add the element maps.

I don't have a test for this off-hand, but my end use for this is in MOOSE and I could submit something there, or I could learn the libmesh test system (which I've neglected to look into thus far). I'll leave it up to the developers.

### DIFF
--- a/src/mesh/abaqus_io.C
+++ b/src/mesh/abaqus_io.C
@@ -136,10 +136,24 @@ void init_eletypes ()
       }
 
       {
+        // TRI6
+        const unsigned int   node_map[] = {0,1,2,3,4,5}; // identity
+        const unsigned short side_map[] = {0,1,2}; // identity
+        add_eletype_entry(TRI6, node_map, 6, side_map, 3);
+      }
+
+      {
         // QUAD4
         const unsigned int   node_map[] = {0,1,2,3}; // identity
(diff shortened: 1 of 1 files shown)


jwpeterson (approved):
Thanks! It looks like we already had the relevant search strings for TRI6 and QUAD8 in`AbaqusIO::read_elements()`, so this was just a simple omission of the relevant `ElementDefinition`s, likely due to not having a test.

jessecarterMOOSE:
Yeah I was surprised how it easy it was to get this working.

roystgnr (approved):
I'll beg for a libMesh unit test (cppunit is easy enough to work with, especially if you're just adding a new test to an existing file like tests/mesh/mesh_input.C), as it makes development easier when the test cycle is faster.
...
But beggars can't be choosers.  This looks good to me.

PR #3274, opened 2022-05-09
Use unique_ptr in ParameterVector

roystgnr:
This is cleaner code, and it also fixes a memory leak in
deep_resize()...

### DIFF
--- a/include/systems/parameter_vector.h
+++ b/include/systems/parameter_vector.h
@@ -47,7 +47,7 @@ public:
   /**
    * Default constructor: "no parameters"
    */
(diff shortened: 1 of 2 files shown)


jwpeterson (approved):
Looks good to me unless in `ParameterVector::deep_copy` you want to add 
```
#ifndef NDEBUG
target._is_shallow_copy = false;
#endif
```

roystgnr:
The tests in GRINS seem happy (after a few updates in https://github.com/grinsfem/grins/pull/620 for our recent qoi API changes).

PR #3276, opened 2022-05-10
Always add elements/nodes to mesh when unpacking

lindsayad:
The packed range algorithms only work if the object is added as
soon as it is unpacked. But `push_parallel_packed_range` does all
of its unpacking before the user gets a chance to add objects to
the mesh manually. So we need to make sure we do it automatically
in the unpack routines

### DIFF
--- a/include/mesh/mesh_inserter_iterator.h
+++ b/include/mesh/mesh_inserter_iterator.h
@@ -20,6 +20,10 @@
 #ifndef LIBMESH_MESH_INSERTER_ITERATOR_H
 #define LIBMESH_MESH_INSERTER_ITERATOR_H
 
+#include "libmesh/libmesh_config.h"
+
+#ifdef LIBMESH_ENABLE_DEPRECATED
+
 // Local includes
 #include "libmesh/mesh_base.h"
 
@@ -39,11 +43,13 @@ class Point;
  * arguments, which adds objects to the Mesh.
  * Although any mesh_inserter_iterator can add any object, we
  * template it around object type so that type inference and
- * iterator_traits will work.
+ * iterator_traits will work. This class used to be intended
+ * for use with packed range methods in TIMPI. However, our
+ * packing routines now automatically add elements and nodes
+ * to the mesh, so this class is deprecated
  *
  * \author Roy Stogner
  * \date 2012
- * \brief An output iterator for use with packed_range functions.
  */
 template <typename T>
 struct mesh_inserter_iterator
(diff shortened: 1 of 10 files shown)


lindsayad:
https://github.com/idaholab/moose/pull/19971/commits/a26a238f4c20c2c9d15935cab8cf54a7d20175ea now works great with this change

include/mesh/do_nothing_inserter_iterator.h
```diff
@@ -0,0 +1,73 @@
...
+ * context such that the (un)packing process adds directly into the mesh.
+ * Given such a specialization, we don't want to actually do anything
+ * with the output iterator, which motivates the creation of this class
+ *
+ * \author Alexander Lindsay
+ * \date 2022
+ * \brief A do-nothing output iterator for use with packed_range functions.
+ */
+template <typename T>
+struct do_nothing_inserter_iterator
```
roystgnr, line 44: Is there something that this can do that include/utils/null_output_iterator.h can't?

include/mesh/mesh_inserter_iterator.h
```diff
@@ -39,11 +43,13 @@ class Point;
  * arguments, which adds objects to the Mesh.
  * Although any mesh_inserter_iterator can add any object, we
  * template it around object type so that type inference and
- * iterator_traits will work.
+ * iterator_traits will work. This class used to be intended
+ * for use with packed range methods in TIMPI. However, our
+ * packing routines now automatically add elements and nodes
+ * to the mesh, so this class is deprecated
```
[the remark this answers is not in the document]
lindsayad: I would assume that it would just waste CPU cycles, but I feel like you would know the answer to this better than I... I figured our early return checks would do the job. Things like
```c++
Elem * DistributedMesh::add_elem (Elem * e)
{
  // Don't try to add nullptrs!                                                                                       
  libmesh_assert(e);

  // Trying to add an existing element is a no-op                                                                     
  if (e->valid_id() && _elements[e->id()] == e)
    return e;
```

include/mesh/do_nothing_inserter_iterator.h
lindsayad: I would think not! Let me check! Clearly my searching for pre-existing options was not thorough enough

include/mesh/mesh_inserter_iterator.h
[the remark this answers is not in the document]
roystgnr: Yeah, I think you're probably correct here.  And if not, anyone savvy enough to use this API is probably able to keep up with the changes too.

include/mesh/do_nothing_inserter_iterator.h
lindsayad: looks like `null_output_iterator` worked like a charm

lindsayad:
Looking at that diff report ... should I also deprecate `insert_node`?

roystgnr:
I think so.

lindsayad:
I guess I'll find out soon one way or the other whether coverage reports include deprecated or not...

roystgnr (approved):

PR #3280, opened 2022-05-10
Add support for elemsets when stitching meshes

jwpeterson:
This does not yet handle the case of merging different elemset codes that refer to the same elemset ids, e.g.

1. mesh A has elemset_code 0 referring to elemset ids {a, b}
1. mesh B has elemset_code 1 referring to elemset ids {a, b}
1. mesh B is stitched onto mesh A

It should be easy to handle this simply by retaining mesh A's elemset code and updating the Elems which came from Mesh B accordingly, however this is not a case which I need to handle currently (elemset ids of stitched meshes will always be distinct), so it can be left as future work for now.

### DIFF
--- a/tests/mesh/mesh_stitch.C
+++ b/tests/mesh/mesh_stitch.C
@@ -308,43 +308,14 @@ public:
 
     // Check that the stitched mesh has merged elemset codes and ids as expected
     MeshBase::elemset_type id_set_to_fill;
-    for (const auto elemset_code : {1, 2, 3, 4})
+    const elemset_id_type code_to_type[] = {0,1,2,100,200};
+    for (dof_id_type elemset_code=1; elemset_code<5; ++elemset_code)
       {
         mesh0->get_elemsets(elemset_code, id_set_to_fill);
 
-        // Should be one elemset id in each set
+        // Assert one elemset id in each set, and that set contains the correct id
         CPPUNIT_ASSERT_EQUAL(static_cast<std::size_t>(1), id_set_to_fill.size());
-
-        switch (elemset_code)
-          {
-          case 1:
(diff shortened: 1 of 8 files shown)


jwpeterson:
Hmm... I ran my new unit test before pushing and it worked fine, so my changes must have broken one of the existing unit tests.

tests/mesh/mesh_stitch.C
```diff
@@ -228,6 +224,169 @@ public:
...
+    // Check that the stitched mesh has merged elemset codes and ids as expected
+    MeshBase::elemset_type id_set_to_fill;
+    for (const auto elemset_code : {1, 2, 3, 4})
+      {
+        mesh0->get_elemsets(elemset_code, id_set_to_fill);
+
+        // Should be one elemset id in each set
+        CPPUNIT_ASSERT_EQUAL(static_cast<std::size_t>(1), id_set_to_fill.size());
+
+        switch (elemset_code)
```
roystgnr, line 318: There's about 30 lines here that could be replaced with something like:
```
const MeshBase::elemset_type code_to_type[] = {0,1,2,100,200};
CPPUNIT_ASSERT(id_set_to_fill.count(code_to_type[elemset_code]));
```

Right?

But that's the worst nitpick I have, and it's in a unit test so I hardly care if it's cleaned up.

roystgnr (approved):

tests/mesh/mesh_stitch.C
jwpeterson: Oh, yeah that would work because of the particular elemset_codes I chose, huh. I can make that change once it looks like there aren't any more warnings to address.

PR #3281, opened 2022-05-11
ParameterMultiAccessor should store a vector of unique_ptrs

jwpeterson:
Another hopefully straightforward update similar to #3274, #3273, #3270. We don't have any tests (or even code that includes the header for) `ParameterMultiAccessor` so strictly speaking this should be run through the GRINS regression testing before being merged, and (even more strictly speaking) we should probably add a unit test for this in libmesh.

### DIFF
--- a/include/systems/parameter_multiaccessor.h
+++ b/include/systems/parameter_multiaccessor.h
@@ -106,11 +106,11 @@ public:
    */
   virtual std::unique_ptr<ParameterAccessor<T>> clone() const override
   {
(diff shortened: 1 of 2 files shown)


roystgnr:
Looks good, and this doesn't break anything in GRINS (again, with https://github.com/grinsfem/grins/pull/620 ) for me.

roystgnr:
I totally spoke too soon there - I'd built against an older libMesh install by accident.  There actually is a bug here; I'll put up a separate PR with the fix.

PR #3283, opened 2022-05-11
Fix unused variable warning in --disable-exodus case

jwpeterson:
Noticed this in the devel->master merge, where `-Werror` has recently been added to a few recipes. I will add Test No Exodus and Test No Optional recipes to this PR momentarily.


roystgnr (approved):

jwpeterson:
I'm just going to merge this because it passed the "Test No Exodus" and "Test No Optional" tests, and it is needed for devel -> master merges.

PR #3287, opened 2022-05-12
Do Elem/FE order compatibility testing in optimized modes too 

roystgnr:
This shouldn't be too expensive compared to the rest of the reinit(), and getting a baffling low-level error in opt mode instead is a constant source of confusion for users whenever they create an inconsistent combination for the first time.

Pinging @bwspenc, who's patiently + politely "complained about this a million times"; turns out the million-and-first time did the trick.  In hindsight I should have dug into this when he first talked with me about it long ago, not waited for yet-another-victim to pop up on Slack this week.

### DIFF
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -629,25 +629,25 @@ void DofMap::reinit(MeshBase & mesh)
 
           FEType fe_type = base_fe_type;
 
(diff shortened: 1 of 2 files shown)


bwspenc:
Ok, well "a million" was probably more like 3. Thanks for addressing this!

PR #3290, opened 2022-05-15
testBadElemFECombo requires exception support to be enabled

jwpeterson:
This is currently preventing a devel -> master merge. I'll attach the Test No Exceptions recipe to this PR as well.


roystgnr:
Thanks for catching this.

PR #3291, opened 2022-05-15
Parameters should store a map of std::unique_ptrs

jwpeterson:
* Add typedef for map used by Parameters
* Simplify logic in Parameters::have_parameter()
* Use range-based for-loop
* More readable code in Parameters::set


jwpeterson:
I thought this might happen since the `InputParameters` class in MOOSE derives from `Parameters`... there might be a relatively simple fix though.

roystgnr (approved):

PR #3293, opened 2022-05-17
Always examine function parser error

lindsayad:
@bwspenc has pointed out that `EvalError` just returns an
error code that gets set during `Eval`, hence error checking is
really no more expensive than checking an `if` condition. He's
also pointed out that it may be impossible to do an initial setup
check to check all the possible evaluations of the function that
may occur during simulation execution. Given the almost free nature
of the check, I think it makes good sense to always check the error
code and potentially save users compiling in debug mode etc. in order
to figure out the issue in their simulation


roystgnr (approved):

bwspenc:
Awesome, thanks!

PR #3298, opened 2022-05-19
Support+fixes for more complicated MeshedHole and Poly2Tri refinement cases

roystgnr:
I'm still trying to come up with a unit test to tack on here; these were slightly tricky bugs to trigger reliably, though.

### DIFF
--- a/src/mesh/mesh_triangle_holes.C
+++ b/src/mesh/mesh_triangle_holes.C
@@ -311,6 +311,7 @@ TriangulatorInterface::MeshedHole::MeshedHole(const MeshBase & mesh,
                 {
                   hole_edge_map.emplace(elem->node_ptr(s),
                                         elem->node_ptr((s+1)%ns));
+                  // Do we really need to support flipped 2D elements?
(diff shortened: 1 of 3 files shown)


src/mesh/mesh_triangle_holes.C
```diff
@@ -323,57 +324,94 @@ TriangulatorInterface::MeshedHole::MeshedHole(const MeshBase & mesh,
...
-      auto [next_it_begin, next_it_end] = hole_edge_map.equal_range(n);
+        libmesh_error_msg_if
+          (std::distance(next_it_begin, next_it_end) != 2,
+           "Bad edge topology in MeshedHole");
 
-      libmesh_error_msg_if
-        (std::distance(next_it_begin, next_it_end) != 2,
-         "Bad edge topology in MeshedHole");
+        const Node * next = nullptr;
+        for (auto [key, val] : as_range(next_it_begin, next_it_end))
```
jwpeterson, line 355: `key` and `val` are `const` here, I think.

jwpeterson (approved):

PR #3299, opened 2022-05-19
More PETSc 3.17 compatibility

roystgnr:
I'm not sure how much test coverage we have on these two files, but this was what it took to get compilation working for me.

### DIFF
--- a/include/numerics/petsc_macro.h
+++ b/include/numerics/petsc_macro.h
@@ -94,6 +94,18 @@
 # define LibMeshCreateSubMatrix MatCreateSubMatrix
 #endif
 
+// As of release 3.17.0, SETERRQ was made to use __VA_ARGS__ and
+// SETERRQ1, SETERRQ2, etc. were deprecated
+#if PETSC_VERSION_LESS_THAN(3,17,0)
+# define LIBMESH_SETERRQ1 SETERRQ1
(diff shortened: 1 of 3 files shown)


pbauman:
> This reminds me... we should consider whether we can remove the petscdmlibmeshimpl.C code, as I think it has been superseded by the code in petsc_dm_wrapper.h, which is also what the multigrid preconditioner/solver stuff is implemented on top of.

The MOOSE folks would be the ones to ask, I think. I'm not sure if there is additional functionality in the petscdmlibmeshimpl.C code or not. If they aren't using it, then I would suggest it be removed.

jwpeterson:
@fdkong do you know whether MOOSE is using petscdmlibmeshimpl.C code internally? I think not... it was something added to libmesh a while ago by Dmitry.

fdkong:
I think we do not really use that in moose

jwpeterson:
OK, thanks for letting us know. I think it is only used in misc_ex7, which is a variational inequality solver. So I guess we'd have to either be willing to drop that example or somehow refactor it to use the pestc_dm_wrapper.h approach (not sure whether this is possible).

fdkong:
It would be a good idea to check if we lose any capability or not before we remove this code

PR #3300, opened 2022-05-21
Convert _diff_physics to std::unique_ptr

jwpeterson:
* Move _diff_physics accessors to C file
* Add custom deleter for _diff_physics unique_ptr
* Maintain previous behavior of holding a shallow copy of ourselves

Initially this change was meant to make the code *simpler* but I think it failed in that goal. What it does do is improve the handling of the case where a `DiffSystem` is destroyed (for example, due to exception handling) while holding a shallow copy of a `_diff_physics` pointer. Previously the `DiffSystem` destructor would have called `delete` on such a pointer, but that would have been the wrong behavior if it was a stack pointer or a dumb pointer actually owned (and deleted) by someone else. This PR also adds support for swapping in a `DifferentiablePhysics` object via `unique_ptr`, a use case which we did not previously support. The new versions of `swap_physics()` don't allow deep copies to be swapped away into dumb pointers, or shallow copies to be swapped into owernship by a `unique_ptr`, since both of those situations could potentially lead to issues.

### DIFF
--- a/include/systems/diff_system.h
+++ b/include/systems/diff_system.h
@@ -29,6 +29,7 @@
 
 // C++ includes
 #include <memory>
+#include <functional> // std::bind
 
 namespace libMesh
 {
@@ -181,8 +182,7 @@ public:
    * \note If no external Physics object is attached, the default is
    * \p this.
    */
-  const DifferentiablePhysics * get_physics() const
-  { return this->_diff_physics; }
+  const DifferentiablePhysics * get_physics() const;
 
   /**
    * \returns A reference to a DifferentiablePhysics object.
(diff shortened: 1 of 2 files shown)


roystgnr:
Yeah, what a mess.  Though I admit I would prefer an exception-correct mess over the previous simple-but-leaky code.

Trouble is, we have no path forward here.  The `_diff_physics == this` case is too widely used to push away from quickly, but until it's completely deprecated we still have to expect that shallow copies are a typical use case, which means we can't ever rely on the `swap_physics(unique_ptr)` overload.

There's got to be a better way to rewrite this, long-term, even though I think you've convinced me there's no better backwards-API-compatible way.

My thoughts:

We can use a nullptr unique_ptr to denote the "our physics is ourself" case.

We can probably deprecate the `swap_physics` case entirely, replacing it with a `push_physics` and `pop_physics`.  Pinging @vikramvgarg to correct me if I'm wrong, but isn't that really our only category of use case for `swap_physics`?  We want to use a new physics object in our system temporarily, then we want to restore the old physics object when we're done.  We've been doing that by swapping the two and then swapping the two again, but pushing a new physics on and then popping it off again would accomplish the same thing, and then we would be able to take the new physics as a reference to clone and not worry about memory management of it at all, and keep our old physics in a stack so we never relinquish memory management of that.

jwpeterson:
> We can use a nullptr unique_ptr to denote the "our physics is ourself" case.

This was my initial thought as well, but it seemed to be a non backwards-compatible change? I was also worried that I would introduce a bug where a `nullptr` `_diff_physics` object was somehow accessed, so I ended up sticking with the existing "pointer-to-self" approach.

> Trouble is, we have no path forward here.

One thing which we can do now (and could not do before) is `attach_physics()` followed by `swap_physics(unique_ptr &)`. The first call makes a deep copy and the second swaps ownership with an outside `unique_ptr`. If the `DiffSystem` is destroyed while the swap is in effect, everything is still deleted correctly.

> We can probably deprecate the swap_physics case entirely, replacing it with a push_physics and pop_physics

Yes, I think maybe the simplest/best approach would be if the `DiffSystem` was never responsible for deleting a `_diff_physics` object. (In this case we would need to change `attach_physics()` so that it did not make a deep copy.) Then the push/pop or stack approach would match current use cases well.

vikramvgarg:
> We want to use a new physics object in our system temporarily, then we want to restore the old physics object when we're done.

Yes, that is the only use case I am aware of. This is done in adjoints example 7.

PR #3304, opened 2022-05-27
Convert _diff_physics to vector<unique_ptr>

roystgnr:
This seems cleaner than #3300 to me, though that's certainly debatable.

I tested the new `swap_physics` (and caught bugs in my first cracks at it) against adjoints_ex7 before I upgraded it to push/pop.

I've also put the `_diff_qoi` on a stack rather than just turning it into a single `unique_ptr`, though in that case we don't yet need the full flexibility a stack allows.

### DIFF
--- a/include/systems/diff_system.h
+++ b/include/systems/diff_system.h
@@ -180,7 +180,7 @@ public:
   const DifferentiablePhysics * get_physics() const
   { if (this->_diff_physics.empty())
       return this;
-    return this->_diff_physics.back().get(); }
+    return this->_diff_physics.top().get(); }
 
   /**
    * \returns A reference to a DifferentiablePhysics object.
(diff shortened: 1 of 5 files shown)


jwpeterson (approved):
Since `DifferentiableSystem::swap_physics` is officially deprecated now, it would be good to attach a "Test No Deprecated" recipe to this. We don't use that internally and I don't think MOOSE does, so I'm thinking GRINS would be the main downstream app needing updates.

include/systems/diff_system.h
```diff
@@ -408,6 +411,21 @@ class DifferentiableSystem : public ImplicitSystem,
...
+   * users should create separate physics objects.
+   */
+  std::vector<std::unique_ptr<DifferentiablePhysics>> _diff_physics;
+
+  /**
+   * Pointer to object to use for quantity of interest assembly
+   * evaluations.  Defaults to \p this for backwards compatibility; in
+   * the future users should create separate physics objects.
+   */
+  std::vector<std::unique_ptr<DifferentiableQoI>> _diff_qoi;
```
jwpeterson, line 428: These vectors could be `std::stacks` I guess... though its not clear to me if there are any benefits to that change beyond communicating their purpose to the reader. The runtime performance is probably exactly the same.

roystgnr:
I checked out GRINS manually; there's no direct swap use there.

PR #3308, opened 2022-05-31
Add/use std::vector constructor for abstract_multi_predicate

jwpeterson:
This lets us avoid a lot of boilerplate `push_back` code as well as taking advantage of the constructor's initialization list rather than creating an empty vector and then filling it up with values. Also adds the `pred_ptr` alias declaration to reduce boilerplate further.

Refs #3306

### DIFF
--- a/include/base/multi_predicates.h
+++ b/include/base/multi_predicates.h
@@ -25,6 +25,8 @@
 // C++ includes
 #include <vector>
 #include <memory>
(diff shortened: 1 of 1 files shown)


jwpeterson:
I developed this change using ancient GCC 7.4.0, and it passed both the min-clang and min-gcc testing here, so I think it's probably OK.

PR #3309, opened 2022-06-01
Configure --enable-xdr-required option

roystgnr:
Pinging @loganharbour to try this out on a system that actually lacks xdr.h; I've only tested it in conjunction with `--disable-xdr`, because Ubuntu 20.10 doesn't want to let me `apt remove libtirpc-dev` unless I'm willing to remove a hundred other -dev packages and all my compilers with it.

### DIFF
--- a/configure
+++ b/configure
@@ -1219,6 +1219,7 @@ enable_perflog
 enable_examples
 enable_strict_lgpl
 enable_nested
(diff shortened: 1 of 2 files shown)


roystgnr:
failed_successfully

loganharbour (approved):

PR #3310, opened 2022-06-01
Update MetaPhysicL

roystgnr:
This incorporates a compiler warning fix from Alex, plus a workaround for clang 9 problems from me.


roystgnr:
I still see a failure (segfault when system_io.C spawns a thread to do a vector write) when I try to use clang-9 for testing on my workstation.  Based on the difficulty it gave me with a mere rvalue reference parameter in MetaPhysicL, I suspect that clang-9 isn't quite up to the task of compiling with libstdc++ headers designed for gcc 12.  If this PR doesn't fix our Moose submodule update then I'll try debugging on the proper clang-9-with-contemporary-libc++ environment on those machines next.

PR #3311, opened 2022-06-02
Use declarations header for declarations

roystgnr:
Otherwise we can't adjust the declarations in MetaPhysicL (which we've
started doing in complicated ways, to handle bugs with clang-9) unless
we carefully reflect them here.


roystgnr:
They're in MooseUtils.h too?  Seriously?

PR #3313, opened 2022-06-10
Add support for writing Node/Elem integers to XDR/A files

jwpeterson:
Includes two unit tests (one with elemsets and one without) of the new functionality.

### DIFF
--- a/include/mesh/xdr_io.h
+++ b/include/mesh/xdr_io.h
@@ -268,6 +268,11 @@ private:
 
   //---------------------------------------------------------------------------
   // Read Implementation
+  //
+  // In the function templates below, the "T type_size" argument is only
+  // used for template argument deduction purposes. The value of this
+  // parameter can be arbitary because it will not be used within the
+  // function.
 
   /**
    * Read header information - templated to handle old (4-byte) or new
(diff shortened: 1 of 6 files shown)


jwpeterson:
FYI this PR includes a bump of the XDR file version to 1.8.0, and this jogged my memory that I never did the final libmesh-1.7.0 release, which was very close to happening several months ago. So I will try and take care of that soon.

include/mesh/xdr_io.h
```diff
@@ -201,6 +201,15 @@ class XdrIO : public MeshInput<MeshBase>,
...
   bool version_at_least_1_3_0() const;
 
+  /**
+   * \returns \p true if the current file has an XDR/XDA version that
+   * matches or exceeds 1.8.0.
+   *
+   * In this version we added support for writing the Mesh's "extra"
+   * integer fields to the file.
+   */
+  bool version_at_least_1_8_0() const;
```
roystgnr, line 211: In hindsight we probably should have made this a single API with three int parameters, and some internal asserts to protect against typos.  Not complaining now, just noticing it's something we can consolidate later.

jwpeterson: I agree, sounds like a good refactoring to me. 

I saw a few other things that I would like to potentially try and refactor in `XdrIO` while working on this PR, for example in `write_serialized_nodes()` I simply continued the pattern of re-communicating the node ids for a (now) third time that was started when we added support for writing unique ids to the Xdr file. The node ids should probably just be saved on processor 0 in "chunks" the first time they are communicated, but I didn't want attempt to make that larger refactoring while also adding this new feature.

roystgnr (approved):

PR #3321, opened 2022-06-22
Drop element orientation check for 1D elements

jwpeterson:
According to @roystgnr, this check may simply be unnecessary now:

"I guess we could get rid of it.  We handle flipped 1D and 2D elements these days, as a special case of embedded-in-3D, but that clause might have predated that support."

So let's see if this change passes all the testing. I also added a unit test that would have failed before the update, but passes now.

### DIFF
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -596,15 +596,13 @@ void UnstructuredMesh::find_neighbors (const bool reset_remote_elements,
 
                         // If found a match with my side
                         //
-                        // We need special tests here for 1D:
-                        // since parents and children have an equal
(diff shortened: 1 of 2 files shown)


roystgnr:
Fingers crossed.  adaptivity_ex3 is the one to watch out for; that has both h and hp refinement tests on both LAGRANGE and HERMITE elements in 1D, and I can't imagine us not running into trouble if the change here has any issues.

roystgnr:
No changes there from the last CI run as far as I can see.

jwpeterson:
The Test MOOSE failure here is not related to the PR, it is something in the "Test moose python" step:
```
[41.56s]     FAIL MooseDocs/test/commands.check FAILED (CRASH)
```

PR #3322, opened 2022-06-23
Pass configure args to netcdf in a less hacked-together way.

roystgnr:
Our ac_configure_args hackery was causing those subdir arguments to end
up in our own configure lines as well (and redundantly, too, becoming
longer and longer each time make decided that `config.status --recheck`
was necessary!).  Using the macro here (as we already do with
MetaPhysicL) should be better supported, and it fixes those side effects
for me.

I've tested this with a third-party HDF5, and my initial now-deleted
attempts to fix the problem by making the hackery even more hacked up
have me convinced that my test coverage was solid there.

### DIFF
--- a/configure
+++ b/configure
@@ -710,7 +710,6 @@ LIBMESH_ENABLE_NETCDF_V4_TRUE
 LIBMESH_ENABLE_NETCDF_FALSE
 LIBMESH_ENABLE_NETCDF_TRUE
 NETCDF_INCLUDE
-subdirs
 LIBMESH_ENABLE_HDF5_FALSE
 LIBMESH_ENABLE_HDF5_TRUE
 HDF5_PREFIX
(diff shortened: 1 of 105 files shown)


jwpeterson:
Same goes for the "Test complex and infinite" failure.

roystgnr:
Yeah, I hadn't been looking into the failures here while our CI modules were still in flux, but those look legit.  Let me test the PETSc HDF5 autodetection case myself; I've been doing my own runs and tests with an independent HDF5 install.

roystgnr:
So much for that theory.  On my system we're passing 'CPPFLAGS=a-bunch-of-petsc-stuff' to netcdf configure just fine, and it finds PETSc's HDF5 from that.

roystgnr:
Not sure how that last test didn't notify Github, but tests are all passing now.  I'll probably merge tomorrow if there are no objections.  I'm not sure this is the ideal solution but it's about ten times less ugly than the workaround it's replacing.

PR #3323, opened 2022-06-24
Explicitly use ReplicatedMesh in test of writing elemset data

jwpeterson:
Getting this test working with `DistributedMesh` would probably take more time than I have at the moment, and I don't think there's any pressing desire to use this capability for `DistributedMesh` (certainly not from my end) so for now I'm restricting the test to run with `ReplicatedMesh`.


loganharbour:
Clang crud is fixed; sorry about that

jwpeterson:
@roystgnr brought this issue to my attention and he was OK with the temporary fix, so I'm merging now.

PR #3324, opened 2022-06-25
Poly2TriTriangulator fixes for DistributedMesh

roystgnr:
This is the set of changes needed to fix our Poly2Tri unit tests (and the tests in https://github.com/idaholab/moose/pull/21303) with `--enable-distmesh`, but that `MeshSerializer` update ought to make the serializer potentially more flexible for other use cases too.

### DIFF
--- a/src/mesh/poly2tri_triangulator.C
+++ b/src/mesh/poly2tri_triangulator.C
@@ -189,6 +189,14 @@ void Poly2TriTriangulator::triangulate()
       _elem_type != TRI7)
     libmesh_not_implemented();
 
+  // If we have no explicit segments defined, we may get them from
+  // mesh elements
+  this->elems_to_segments();
+
+  // If we *still* have no explicit segments defined, we get them from
+  // the order of nodes.
+  this->nodes_to_segments(_n_boundary_nodes);
+
   // Insert additional new points in between existing boundary points,
   // if that is requested and reasonable
   this->insert_any_extra_boundary_points();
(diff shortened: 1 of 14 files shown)


roystgnr:
Well, I spoke too soon.  Half the proposed MOOSE poly2tri tests now work with DistributedMesh, but I'll need to look deeper into the rest.  This ought to be merged regardless so we can get `--enable-distmesh` tests in our regular CI here.

roystgnr:
Okay, with these last few commits (plus changes in MOOSE) my tests there are also working on DistributedMesh, and our unit test suite here is find with `--enable-distmesh` builds up to 8 ranks.  I'm still seeing something weird from an Exodus test at 9 ranks, but that looks like a separate issue so I'd like to merge this once CI is happy.

src/mesh/poly2tri_triangulator.C
```diff
@@ -164,6 +164,11 @@ Poly2TriTriangulator::~Poly2TriTriangulator() = default;
 // Primary function responsible for performing the triangulation
 void Poly2TriTriangulator::triangulate()
 {
+  // We only operate on serialized meshes.  And it's not safe to
+  // serialize earlier, because it would then be possible for the user
+  // to re-parallelize the mesh in between there and here.
+  MeshSerializer _serializer(_mesh);
```
jwpeterson, line 170: Leading underscore naming should be reserved for member variables.

jwpeterson (approved):

src/mesh/poly2tri_triangulator.C
roystgnr: Yup.  And this originally *was* a member variable, so I probably just cut and pasted.  Will fix.

roystgnr:
If this passes I'll add some DistributedMesh sweeps to CI, and if *that* works too I'll merge.

roystgnr:
Well, at least it's a new example failing this time.

roystgnr:
Looks like my fixes for Triangle+PSLG+DistributedMesh failed horribly with Triangle+GENERATE_CONVEX_HULL+anything.

PR #3327, opened 2022-06-28
Minor XdrIO cleanup/refactoring

jwpeterson:
These were some things I came across while working on adding support for elemsets to XDR in #3313. I rather like the approach of using a `std::tuple` to [declare multiple for-loop iteration variables with loop scope](https://stackoverflow.com/questions/2687392/is-it-possible-to-declare-two-variables-of-different-types-in-a-for-loop). It requires C++17, but it works with my old compiler just fine, so hopefully it also works with our other "min" compilers.

### DIFF
--- a/src/mesh/xdr_io.C
+++ b/src/mesh/xdr_io.C
@@ -445,40 +445,39 @@ XdrIO::write_serialized_connectivity (Xdr & io,
                 output_buffer.clear();
 
                 // n. nodes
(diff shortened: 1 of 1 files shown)


roystgnr (approved):
We don't want to use post-increment-and-dereference on *every* sort of iterator, since it forces a copy and IIRC our predicate iterators are a bit heavyweight, but it's fine here, and everything else looks good to me too.

PR #3330, opened 2022-06-29
Add memory header to fix compile on CentOS

### LINKED ISSUE #3329: Build fails with gcc 10.2.1 on CentOS
The relatively recent `mesh/poly2tri_triangulator` does not compile on CentOS with gcc 10.2.1.  The compile error is

```
./include/libmesh/poly2tri_triangulator.h:139:31: error: 'unique_ptr' is not a member of 'std'
  139 |   std::map<const Hole *, std::unique_ptr<ArbitraryHole>> replaced_holes;
      |                               ^~~~~~~~~~
./include/libmesh/poly2tri_triangulator.h:31:1: note: 'std::unique_ptr' is defined in header '<memory>'; did you forget to '#include <memory>'?
...
```

The fix is just to `#include <memory>` in the header.  I'll upload a PR in a minute.

reverendbedford:
Adds `#include <memory>` to `poly2tri_triangulator.h` to fix a build issue on CentOS.  Will close #3329.

### DIFF
--- a/include/mesh/poly2tri_triangulator.h
+++ b/include/mesh/poly2tri_triangulator.h
@@ -19,6 +19,7 @@
 #ifndef LIBMESH_POLY2TRI_TRIANGULATOR_H
 #define LIBMESH_POLY2TRI_TRIANGULATOR_H
 
+#include <memory>
 
 #include "libmesh/libmesh_config.h"
 


roystgnr (approved):

roystgnr:
Thanks again!

PR #3332, opened 2022-07-04
StoredRange can store an optionally-deleted unique_ptr

jwpeterson:
* Also add some libmesh_asserts in StoredRange::reset() to catch the case
  where a range is reset() but the _objs pointer was not initialized.
* Use lambdas to define the custom deleter. The decision to delete/not
  is made at construction time, and does not change during the object's
  lifetime.
* One thing I am not 100% happy with is that we still need to call 'new'
  when allocating the class unique_ptr, since one cannot really use
  std::make_unique when a custom deleter is employed.
  https://stackoverflow.com/questions/21788066/using-stdmake-unique-with-a-custom-deleter
* Use ptr_type typedef for improved readability
* Add required C++ includes.

### DIFF
--- a/include/geom/stored_range.h
+++ b/include/geom/stored_range.h
@@ -25,6 +25,8 @@
 
 // C++ includes
 #include <vector>
+#include <memory> // std::unique_ptr
+#include <functional> // std::function
 
 namespace libMesh
 {
@@ -55,7 +57,9 @@ public:
   /**
    * Allows an \p StoredRange to behave like an STL container.
    */
-  typedef typename std::vector<object_type>::const_iterator const_iterator;
+  typedef typename std::vector<object_type> vec_type;
+  typedef typename vec_type::const_iterator const_iterator;
+  typedef typename std::unique_ptr<vec_type, std::function<void (vec_type *)>> ptr_type;
 
   /**
    * Constructor. Optionally takes the \p grainsize parameter, which is the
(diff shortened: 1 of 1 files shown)


roystgnr (approved):

jwpeterson:
So we have a randomly failing Min clang target, is that correct?

roystgnr:
Yup.  In MetaPhysicL.  I'll see if I can replicate it locally.

roystgnr:
I'm getting the same non-deterministic failures from clang 13 and 14, and gcc 9 and 11; this has got to be our problem, not a compiler issue.

Unless it's an issue with something else in the environment?  I went to an older Linux box to try clang 10, the compiler in our failing CI runs, and *that* turns out to be happy to make thousands of runs without a hiccup.

roystgnr:
... `instantiations_unit` is outputting uninitialized values.  That's undefined behavior, and it's non-deterministic in the heap-allocated cases because the heap might be pulling memory we've already used for something else, and some of that memory might be effectively getting a reinterpret_cast to NaN.

I'll fix it.

jwpeterson:
OK, sounds good. From what I could see, the metaphysicl test suite doesn't yet output the name of the test which failed (?) but that would be useful for trying to diagnose problems via CIVET... since I don't think the unit test log can be downloaded as part of the results.

Issue #3328, opened 2022-06-29
No fully-compatible MPI implementations on Ubuntu 22.04 LTS


roystgnr:
OpenMPI has a fix for their regression currently posted at https://github.com/open-mpi/ompi/pull/10527, and MPICH seems to already have their git head working again, but in the meantime, not being able to do an `MPI_MIN` with an arbitrarily large `unsigned long` is a big deal for us.  At the very least it makes `DistributedMesh` unusable with `METHOD=dbg` (with our git head I get failures from the `semiverify` asserted at `src/mesh/mesh_tools.C, line 2164`), and there's no telling where else we might be triggering the same problem.

I'm currently working around the problem myself with a `--download-mpich` PETSc build (which pulls MPICH 3.4.1 right now; that seems to predate the regression there), just posting a warning for others.

jwpeterson:
Just a small follow-up comment on this, since it led to some recent confusion for us. The Ubuntu 22 libmpich-dev package which is labeled ["4.0-3"](https://changelogs.ubuntu.com/changelogs/pool/universe/m/mpich/mpich_4.0-3/changelog)
```
Package: libmpich-dev
Architecture: amd64
Version: 4.0-3
Multi-Arch: same
Priority: extra
Section: universe/libdevel
Source: mpich
Origin: Ubuntu
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: Debian Science Maintainers <debian-science-maintainers@lists.alioth.debian.org>
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Installed-Size: 58262
Depends: gfortran | fortran-compiler, g++, libmpich12 (= 4.0-3), gfortran-11 | gfortran-mod-15, mpich (= 4.0-3)
Filename: pool/universe/m/mpich/libmpich-dev_4.0-3_amd64.deb
Size: 7374790
MD5sum: 73f445f5d0d674ff6b653ffd2f5cf85d
SHA1: 2119398701c2842a8a4f2ca96ad591e08b05e648
SHA256: bfe7155963cef03a800746913b51e8e5489d18cedf5950a2e6b86ab0e03500d3
SHA512: ee551ce69c59039820dca26681b56ee2a0c0b20c9c2fb20cd83496606cc358d03b8832f750197e09bc93cad5cc5b37c1f6a43e2f4dd6c11773ca450fb2899543
```

[...]

PR #3335, opened 2022-07-06
Update TIMPI to v1.8.5

roystgnr:
This avoids an overzealous compiler warning (and fixes --enable-werror
--enable-paranoid-warnings builds) for me.


jwpeterson (approved):

jwpeterson:
Looks like you added some new tests? Just wondering if those fail on the MPI distributions referred to #3328, i.e. can we use TIMPI tests to determine if a distro's MPI has been patched?

roystgnr:
Those tests will catch the implementation bug, but only if they're run with `TIMPI_RUN='mpirun -np 2'` or more.  I'm afraid I couldn't figure out how to trigger a failure in serial.

PR #3339, opened 2022-07-18
Poly2tri error reporting, feature add, refinement fixes

roystgnr:
This branch was originally just trying to clean up some handling of user-input errors, but it snowballed a bit.  Changes are all limited to the triangulation code, though.

### DIFF
--- a/include/mesh/mesh_triangle_holes.h
+++ b/include/mesh/mesh_triangle_holes.h
@@ -288,9 +288,9 @@ private:
  * edges in a subdomain with a listed id will define the hole.
  *
  * If no ids are given, the hole will be defined by all 1D Edge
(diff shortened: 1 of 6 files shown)


dschwen:
Is there any code to (automatically) add interior Steiner points yet? If not, is that planned?

roystgnr:
It's in there.  (though I'm a little rattled - I thought it was working weeks ago, I'd been running tests generating hundreds of thousands of triangles on adaptively graded grids ... and then I just found and fixed another bug in it yesterday)

The biggest limitation right now is that you can only direct the triangulator to add points to reduce triangle area; there's not yet an option to try to increase triangle minimum angle.

PR #2160, opened 2019-06-13
Clear up the old SLEPc solver before we start calling SLEPc EPS.

fdkong:
If we call `solve() `of `slepc_eigen_solver` multiple times, we should obtain the same answer.

If we do not clear up the old EPS, it looks like there are messing things in EPS. And then we will 
get a different answer when we have the second call to EPS.

PR #3342, opened 2022-07-21
Remove the line for resetting the solver type in slepc eigen solver

### LINKED ISSUE #3341: SlepcEigenSolver clear function resetting the default solver type which may prevents calling solve function properly
This:
```
      // SLEPc default eigenproblem solver
      this->_eigen_solver_type = KRYLOVSCHUR;
```
in `void SlepcEigenSolver<T>::clear ()` function is giving me a trouble.

I've called `void set_eigensolver_type (const EigenSolverType est)` to set the desired the solver type and need to call `EigenSystem::solve ()` (that calls one of the solve functions in `SlepcEigenSolver` multiple times). But because that `clear` function resets the default solver type and is called by all the solve functions, I cannot make the second solve call to use the same non-default solver type I set previously. Because of this, I have to use a different way to set the solver type with SLEPc command-line option `-eps_type`, which defeats the purpose of having `set_eigensolver_type` function.

I propose to remove that line because,

1) the clear function is for clearing data changed by `solve` functions but not options set by users to the solver in my opinion;
2) there are other settings in the solver like `_position_of_spectrum`, etc., which should be cleared as well if we think we should reset the solver type in `clear`;
3) we could add a new function like `clearSolveData` and called by solve functions, but I think it is unnecessary. Users can reset the solver type manually if they want.

I've tried removing that line and ran all MOOSE tests. All tests past. I will push up a PR to let civet check if there are any implications.

YaqiWang:
Close #3341.


YaqiWang:
I can add a new function like `clearEPS()` and call it instead of calling `clear()` if you prefer the removal of the line. @jwpeterson and @roystgnr.

YaqiWang:
I can add a new function like `clearEPS()` and call it instead of calling `clear()` if you prefer the removal of the line. @jwpeterson and @roystgnr.

jwpeterson:
I think the right fix here, as @roystgnr suggested, is to try removing the `clear()` from the different `SlepcEigenSolver` solve() APIs. The closest analog I see for this is in `PetscLinearSolver` where we don't call `clear()` before the solves and it seems to work OK. Have you tried this?

roystgnr:
Digging in to the git log: we only added those clear() calls 3 years ago, in #2160 - but they were added to fix a couple actual problems.

YaqiWang:
Yeah, that is the thing I was afraid of. It is very difficult for me to use the slepc solver without relying on setting commandline options to tune the solver even though I am using the EigenSystem. So I decided to just write a very specific small slepc solver for my purpose. So I lost motivation to push this forward. I respect the design to let `clear()` make the object return its original state. Maybe create a new function like `reinit` that can be used by solve functions.

roystgnr:
> Maybe create a new function like reinit that can be used by solve functions.

IMHO this may be the right thing to do medium-term.

In hindsight the right thing to do 3 years ago, as well as in the long term, might have been to require user code to call `clear()` (or a `clear_everything_but_solver_type()`, or whatever it actually needed) manually *as needed*, not force it on every `solve()` call *regardless of need*.  There are typically good performance reasons for wanting to be able to reuse structures (I'm guessing matrix factorizations are the issue, based on skimming threads like https://lists.mcs.anl.gov/pipermail/petsc-users/2016-November/030870.html ?) created in one solve for another similar solve, even if that policy can be pointless in some circumstances and can outright backfire on dissimilar solves.

The trouble with this is I'm not sure where to put that code at the user level.  I guess we could experiment with removing the `clear()` calls entirely, find the failing tests, and look for the cases that need a manual `clear()`.  And if it turns out that we missed some then we'll be teaching downstream users a valuable lesson about writing enough test coverage?

PR #3346, opened 2022-07-22
Fixes a VTK linking issue I experienced with VTK 9.2

makeclean:
I recently installed with VTK 9.2 on my system, and configure failed due a missing symbols from an unrequested library. I think probably happens when the VTK library being used is not the system VTK as otherwise it would likely get picked up automatically. This fixes that issue and I makes the link more pedantic.

### DIFF
--- a/configure
+++ b/configure
@@ -53513,13 +53513,13 @@ else
                                            -lvtkFiltersCore-$vtkmajorminor -lvtkIOXML-$vtkmajorminor -lvtkImagingCore-$vtkmajorminor \
                                            -lvtkIOImage-$vtkmajorminor -lvtkImagingMath-$vtkmajorminor -lvtkIOParallelXML-$vtkmajorminor \
                                            -lvtkParallelMPI-$vtkmajorminor -lvtkParallelCore-$vtkmajorminor \
-                                           -lvtkCommonExecutionModel-$vtkmajorminor"
+                                           -lvtkCommonExecutionModel-$vtkmajorminor -lvtksys-$vtkmajorminor"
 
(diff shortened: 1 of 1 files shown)


jwpeterson:
Hello @makeclean,

Thanks for pointing out this issue. It looks like you edited the configure script directly, but since this script is generated automatically from the contents of m4 files (in this case m4/vtk.m4), we would need to make your update there, otherwise it would be lost the next time someone runs bootstrap. 

If you would like to try this, we have some [instructions](https://github.com/libMesh/libmesh/wiki/Updating-configure-tests%2C-Makefiles%2C-etc.) on how to do it. They are somewhat involved, though so it might be easier if one of the other devs makes the change when they get a chance.

roystgnr:
> configure failed due a missing symbols from an unrequested library

Could you post the exact configure error?  I'm curious as to whether it's just something about your install or whether it's going to hit everyone once all us out-of-date 9.0/9.1 users move up to 9.2.

It looks like libvtksys has been around since at least VTK 5, so hopefully adding it as a linker requirement here won't break any of our really-out-of-date VTK users.

makeclean:
Sorry for taking so long to come back to this, I'm building on our HPC system and I'm hitting the same issues with 9.2, the error message is 
```
/usr/bin/ld: /tmp/cc6CtNnI.o: undefined reference to symbol '_ZN6vtksys18SystemToolsManagerD1Ev'
/home/dc-davi4/rds/rds-ukaea-ap001/moose_dev/vtk/lib64/libvtksys-9.2.so.1: error adding symbols: DSO missing from command line
```
I'll have a bash at doing the m4 files :)

makeclean:
Can also confirm it does *not* occur with VTK 9.1

roystgnr:
Replaced by #3426

PR #3350, opened 2022-07-27
Don't leak memory in bytecoderules_parser.cc

lindsayad:
This was an error when trying to build and link with address sanitizer


jwpeterson:
@dschwen coding this up like "that's a problem for future Daniel to deal with"

jwpeterson (approved):

PR #3352, opened 2022-07-27
TriangulatorInterface::elems_to_segments() work

roystgnr:
Fix elems_to_segments behavior when we're asked to build an outer boundary from only a subset of all our boundary/subdomain ids.

Add elems_to_segments support to our Triangle interface.

Add unit tests for all of the above.

### DIFF
--- a/src/mesh/mesh_triangle_interface.C
+++ b/src/mesh/mesh_triangle_interface.C
@@ -62,6 +62,10 @@ void TriangleInterface::triangulate()
 
   unsigned int n_hole_points = this->total_hole_points();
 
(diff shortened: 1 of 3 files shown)


roystgnr:
Interesting.  I could swear I tested this with Triangle locally.

roystgnr:
Well, I can't yet replicate the problem locally, but I've discovered that `MeshInputTest::testTetgenIO` fails on `DistributedMesh` on 2 processors, so the night hasn't been a total waste?

roystgnr:
I can replicate it on rod, at least, and it breaks in serial too, so hopefully it won't be too much of a hassle to debug.

PR #3359, opened 2022-08-01
Turn on fetchRecurseSubmodules by default

roystgnr:
Going through some old branches of mine (in the dentist's waiting room, too low on laptop battery to do heavy builds/checks...) I ran across one with this little ease-of-use option.  It can still be overridden by --no-recurse-submodules if users ever really want that.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,11 @@
 [submodule "contrib/metaphysicl"]
 	path = contrib/metaphysicl
 	url = ../../libMesh/MetaPhysicL
+        fetchRecurseSubmodules = true
 [submodule "contrib/timpi"]
 	path = contrib/timpi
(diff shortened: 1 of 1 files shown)


jwpeterson (approved):
Any idea how long git has supported this option (and what happens if it doesn't)? I'm guessing the answer may be any version that supports the `--recursive` command line arg, in which case we already require it anyway, so :+1:

roystgnr:
> Any idea how long git has supported this option (and what happens if it doesn't)?

After I was done laughing at myself for Googling "git repository" to look for git's own source code, I found it and did a `git blame Documentation/gitmodules.txt` there, and it looks like fetchRecurseSubmodules dates back to 2010 or earlier.

PR #3362, opened 2022-08-02
Fix up subdir build order

roystgnr:
This doesn't let us get away with removing the hack in
tests/Makefile.am, but it can be a bit more efficient for
highly-parallel builds.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -297,6 +297,8 @@ if LIBMESH_OPROF_MODE
   libmesh_oprof_la_LIBADD   = contrib/libcontrib_oprof.la $(LIBS)
 endif
 
+# Make sure we build the library before we test it
+SUBDIRS += .
 
 if LIBMESH_ENABLE_CPPUNIT
  SUBDIRS += tests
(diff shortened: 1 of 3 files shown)


jwpeterson (approved):
For me, `make -C tests` has never really done anything... I have to do e.g. `make -C tests unit_tests-opt` or `make -C tests check` to have it actually compile something. Does this PR change that behavior? Was there some ticket that discussed the parallel build inefficiency?

roystgnr:
As far as automake is concerned, "only build the test programs if you're explicitly asked for `make check`" is supposed to be a feature, not a bug - faster builds, if you like living on the edge and not testing your builds.  This doesn't change that.

No ticket here; the inefficiency is totally theoretical ... and going back to reread https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html now, I think it's not a practical inefficiency; we're using `$(MAKE)` and GNU make is smarter than I remembered being about it splitting parallel jobs between multiple sub-makes.

PR #3363, opened 2022-08-02
Create MeshBase::reinit_ghosting_functors API

lindsayad:
This is useful for users who wish to do-fine grained calling
of `partition` or `delete_remote_elements` in which case for
some ghosting functors to work well we must first call `mesh_reinit`
on them. This will be used in idaholab/moose#21699

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -219,6 +219,11 @@ public:
   virtual void delete_remote_elements () {}
 
   /**
+   * Loops over ghosting functors and calls mesh_reinit()
+   */
+  void reinit_ghosting_functors();
+
+  /**
    * \returns The logical dimension of the mesh; i.e. the manifold
    * dimension of the elements in the mesh.  If we ever support
    * multi-dimensional meshes (e.g. hexes and quads in the same mesh)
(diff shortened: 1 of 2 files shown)


roystgnr (approved):

roystgnr:
Actually ... could we change the API name on this?  I'm sure `GhostingFunctor::mesh_reinit()` is about a hundred times more useful than `GhostingFunctor::dofmap_reinit()`, but the latter does *exist* and we should make it clear that only the former is being called here.

lindsayad:
I suppose it's good to be 100% clear. But I liked the looks of
```
mesh->reinit_ghosting_functors()
```
vs
```
mesh->mesh_reinit_ghosting_functors()
```
To me the mesh is kind of implicit in that we are calling it through the mesh ... but it's better to be explicit

roystgnr:
Ah, yeah, you're right.

roystgnr:
You were right the first time, I mean.  If we want to encapsulate a `dofmap_reinit()` loop then we'll add a `DofMap::reinit_ghosting_functors()` to do that, and it's clear enough which means which.

lindsayad:
this is lame. I can see that testing is going on civet.inl.gov but github isn't registering it here :-/

lindsayad:
Clear based on https://civet.inl.gov/pr/16437/

PR #3365, opened 2022-08-03
Make Hole::contains() FP-robust, and use it in more verification

roystgnr:
I still need to throw more tests at this, but at the least this version fixes cases that the last version got tricked by, and adds test coverage hitting them.


roystgnr:
I think that covers the "more tests", and a good thing too since the possibly-still-broken case I was worrying about turned out to be an actually-still-broken case.  All fixed now, so I'm planning to merge once CI is happy.

PR #3367, opened 2022-08-06
Fix typos found in docs and examples subdirectories

luzpaz:
Found via `codespell -q 3 -S ./contrib,./m4,./NEWS -L ans,inout,nd,parm,som,theses,ue,wirth`

### DIFF
--- a/doc/html/src/publications.html
+++ b/doc/html/src/publications.html
@@ -16607,7 +16607,7 @@ Lane Carasik, Vishal Patel, Samuel Judd, and Paolo Venneri.
 </td>
 <td class="bibtexitem">
 Chandrakanth Bolisetti, Saran Bodda, Andrew Slaughter, and Justin Coleman.
- Seismic probabalistic risk assessment using MASTODON.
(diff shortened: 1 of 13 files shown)


doc/html/src/publications.html
```diff
@@ -16607,15 +16607,15 @@ <h2>2021</h2>
 </td>
 <td class="bibtexitem">
 Chandrakanth Bolisetti, Saran Bodda, Andrew Slaughter, and Justin Coleman.
- Seismic probabalistic risk assessment using MASTODON.
+ Seismic probabilistic risk assessment using MASTODON.
```
luzpaz, line 16610: Screenshot_20220806_131759

jwpeterson:
Thanks, these changes all look good to me. There are a few which are in machine-generated files like `configure`, so those will need to be fixed in the corresponding "source" files that generate them so they don't come back the next time we run `./bootstrap`. I think it should be OK to merge this PR as-is regardless.

jwpeterson (approved):

PR #3372, opened 2022-08-10
Add support for QUADSHELL4/8 elements in MeshTools::Generation::build_square()

jwpeterson:
These elements can be handled almost identically to `QUAD4` and `QUAD8` so there is not really any new implementation in this PR, just new case blocks.

### DIFF
--- a/src/mesh/mesh_generation.C
+++ b/src/mesh/mesh_generation.C
@@ -89,6 +89,7 @@ unsigned int idx(const ElemType type,
       }
 
     case QUAD8:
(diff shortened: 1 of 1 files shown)


roystgnr (approved):

jwpeterson:
The failures are in Test Moose Python, so I think not relevant to this PR
```
    FAIL moosesqa/test.reports/SQADocumentReport FAILED (CRASH)
    FAIL mooseutils/tests.civet_results FAILED (CRASH)
```

PR #3373, opened 2022-08-11
Configuring with --enable-hdf5-required now implies --enable-hdf5 

jwpeterson:
Previously, the user was required to pass both args:
```
  --enable-hdf5 --enable-hdf5-required
```
to configure in order to achieve this, but now only the latter is necessary.

### DIFF
--- a/configure
+++ b/configure
@@ -33822,7 +33822,7 @@ fi
   # before this can be enabled.
   hardcode_into_libs=yes
 
-  # Ideally, we could use ldconfig to report *all* directories which are
+  # Ideally, we could use ldconfig to report *all* directores which are
   # searched for libraries, however this is still not possible.  Aside from not
   # being certain /sbin/ldconfig is available, command
   # 'ldconfig -N -X -v | grep ^/' on 64bit Fedora does not report /usr/lib64,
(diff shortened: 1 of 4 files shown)


configure
```diff
@@ -33822,7 +33822,7 @@ fi
   # before this can be enabled.
   hardcode_into_libs=yes
 
-  # Ideally, we could use ldconfig to report *all* directories which are
+  # Ideally, we could use ldconfig to report *all* directores which are
```
roystgnr, line 33825: Did you re-bootstrap without the typo fixes in libtool.m4?

roystgnr (changes requested):

roystgnr:
Oh that's a shame about libtool.m4.  Ours is just getting overwritten by whatever's in autotools, huh?

roystgnr (approved):

roystgnr:
I wouldn't bother waiting on the python tests to merge; no idea when the cleanup of MOOSE `next` will be done.

jwpeterson:
> Oh that's a shame about libtool.m4. Ours is just getting overwritten by whatever's in autotools, huh?

Yeah moderately annoying.

PR #3374, opened 2022-08-12
Fix in RB serialization/deserialization

dknez:
Added missing on_mesh_nodes() cases in RB serialization/deserialization


dknez:
I assume this "test moose python" error is not a real failure?

jwpeterson (approved):

PR #3376, opened 2022-08-18
Fix XDR bug introduced in refactoring of 084c949

jwpeterson:
In hindsight, it's obvious that we can't use range-based for-loops in this case since we are also in an outer for-loop over "blocks", but we didn't notice the problem in any of our unit tests because you have to have a mesh larger than the XdrIO block size in order to trigger it. Also added a comment to hopefully prevent the same refactoring from happening again in the future.


roystgnr (approved):
Ouch.  Sorry I didn't notice the problem either.

How long ago was the bug?  Do we need to backport this?

jwpeterson:
Only "Test moose python" failures, so merging.

PR #3381, opened 2022-08-25
print_helper using 3D vector argument

### LINKED ISSUE #3380: 3D Vector Argument for libMesh::print_helper()
`libMesh::print_helper() ` is used by MOOSE for parsing input files. Recently, triple indexed/separated fields were found needed in MOOSE input syntax. To make that happen, `libMesh::print_helper(std::ostream & os, const std::vector<std::vector<std::vector<P> >>* param)` is needed.

miaoyinb:
closes #3380

### DIFF
--- a/include/utils/parameters.h
+++ b/include/utils/parameters.h
@@ -37,7 +37,8 @@
 namespace libMesh
 {
 /**
- * Helper functions for printing scalar, vector and vector<vector> types.  Called from Parameters::Parameter<T>::print(...).
+ * Helper functions for printing scalar, vector, vector<vector> and vector<vector<vector>>
+ * types.  Called from Parameters::Parameter<T>::print(...).
  */
 template<typename P>
 void print_helper(std::ostream & os, const P * param);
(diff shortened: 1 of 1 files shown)


miaoyinb:
@GiudGiud FYI

jwpeterson:
FYI @miaoyinb I sent you an invite to join a group that has its PRs tested automatically. Note that some of the testing may fail currently due to a change that is present in libmesh upstream that is not compatible with MOOSE, we're hopeful that will be updated soon.

miaoyinb:
> FYI @miaoyinb I sent you an invite to join a group that has its PRs tested automatically. Note that some of the testing may fail currently due to a change that is present in libmesh upstream that is not compatible with MOOSE, we're hopeful that will be updated soon.

Got it. Thank you so much.

roystgnr (approved):
I dislike that we're flattening the output here rather than using some kind of delimiter between different sub (and subsub) vectors ... but we're *already* flattening the `vector<vector<P>>` case, so at least we're consistent.

roystgnr:
The Test MOOSE failure is just "Test MOOSE python" again.

PR #3383, opened 2022-08-25
Fix edge cases in preconditioner reuse

reverendbedford:
There are a few edge cases in the new preconditioner_reuse feature:

1. libmesh doesn't make a new preconditioner when the size of the system changes because of mesh refinement.  This is easy to trigger by forcing a new preconditioner in `NonlinearImplicitSystem::reinit ()`
2. petsc doesn't make a new preconditioner when the size of the system changes in the course of solving a variational inequality problem.  This kind of feels like a petsc issue, but it's possible to address in libmesh by checking the size of the system of equations and making a new preconditioner if it differences from the last time through.

### DIFF
--- a/src/systems/nonlinear_implicit_system.C
+++ b/src/systems/nonlinear_implicit_system.C
@@ -82,6 +82,10 @@ void NonlinearImplicitSystem::reinit ()
   // re-initialize the nonlinear solver interface
   nonlinear_solver->clear();
 
+  // force the solver to get a new preconditioner, in
+  // case reuse was set
+  nonlinear_solver->force_new_preconditioner();
+
   // FIXME - this is necessary for petsc_auto_fieldsplit
   // nonlinear_solver->init_names(*this);
 
--- a/src/systems/nonlinear_implicit_system.C
+++ b/src/systems/nonlinear_implicit_system.C
@@ -54,7 +54,7 @@ NonlinearImplicitSystem::NonlinearImplicitSystem (EquationSystems & es,
   es.parameters.set<Real>("nonlinear solver relative step tolerance") = 1e-8;
 
   es.parameters.set<bool>("reuse preconditioner") = false;
-  es.parameters.set<unsigned int>("reuse preconditioner maximum iterations") = 1;
+  es.parameters.set<unsigned int>("reuse preconditioner maximum linear iterations") = 1;
 }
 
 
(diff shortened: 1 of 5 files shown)


src/systems/nonlinear_implicit_system.C
```diff
@@ -82,6 +82,10 @@ void NonlinearImplicitSystem::reinit ()
   // re-initialize the nonlinear solver interface
   nonlinear_solver->clear();
 
+  // force the solver to get a new preconditioner, in
+  // case reuse was set
+  nonlinear_solver->force_new_preconditioner();
```
roystgnr, line 87: Is the clear() above not enough to do that?  That seems like a bug in itself, IMHO.

reverendbedford: No.  In the MOOSE workflow `clear()` gets called after every nonlinear iteration.  So I had to alter what `clear()` means over here to not get rid of the SNES instance (or else there would be no way to save the conditioner).

roystgnr: > No. In the MOOSE workflow clear() gets called after every nonlinear iteration.

Wow ... we render the PetscNonlinearSolver uninitialized, and unless we're reusing the precondtioner we destroy the SNES entirely, after every nonlinear iteration?

What a mess.  At least you don't seem to be digging the hole any deeper here.

roystgnr (approved):
At some point we have a lot to look into regarding the differences between our solver function names and their semantics, but today is not that day.

reverendbedford:
@roystgnr We're still figuring out if and how we can deal with all the oddity variational inequality problems cause, so maybe hold off.  This version does fix adapative mesh refinement though.

roystgnr:
Can do.  Maybe there's a cleaner more general fix that this isn't in the right direction of ... though I don't see it.

reverendbedford:
@roystgnr Can you please consider merging now?  This version fixes the MOOSE issue with AMR but does not fix the variational inequality solver issue.  I don't think there is a way to fix that outside of a change to petsc, so I'll go bother those folks next time I walk across campus.

reverendbedford:
That last push just changes the name of the parameter slightly in response to code from from @lindsayad over in MOOSE.  We can make the change pretty easily now but won't be able to later...

roystgnr:
That name change does make this an API change ... but not for one that made it into a release, or even into master for more than a few months, so I won't be a hardass about a `libmesh_deprecated();` transition shim.  The API did make it into the MOOSE submodule ... I can't find anyone using it in MOOSE itself yet, at least.  But if some random MOOSE app adopts it before we get the name change in then I reserve the right to hit revert here rather than sort out the conflict myself.

reverendbedford:
@roystgnr you should be safe, I'm the one trying to get it into MOOSE and I'm now going to wait for this to migrate.  But fair enough in case some random person noticed it and started using it first.

PR #3384, opened 2022-08-25
Code deduplication

roystgnr:
We definitely want to look at the line-by-line code coverage results on this one, in case I screwed something up in any of the refactoring here and we don't yet have test coverage for it.

There's a lot more still to do in this vein, but I think the ~1000 lines deleted here were the lowest-hanging fruit.


jwpeterson:
I attached a SLEPc test since I don't think we have even compilation coverage of those changes otherwise?

jwpeterson (approved):
I definitely didn't review all 4k lines that you changed, but since you removed a net 2k lines there's also no way I'm complaining :smile:

roystgnr:
I'm not liking how much red is in the code coverage on shell_solve_common, but we're at least hitting the preconditioner+shell case via misc ex4, and that was the only one that I was really worried about the refactor messing up, and it's still hitting the same solution in the same number of iterations.  If preconditioner+shell+indexset-restrict or some combination even more esoteric was broken I'm going to blame that on other people writing features without enough test coverage.

PR #3386, opened 2022-08-30
More code deduplication

roystgnr:
This shaves off another 500-odd more lines, and in theory it adds support for a couple PetscLinearSolver feature combinations that were ignored before.

In practice I hope it doesn't just outright break PETSc solves.  We'll see what CI says.

I think this is the last of the low-hanging fruit.  Much of PatchRecovery/WrappedPatchRecovery ought to be factored out but it'll be tricky to do so without slowing down the non-Wrapped case (but perhaps only trivially?), and there's some bits in fem_system.C that I'd like to consolidate but made a mess of when I tried ... and there's a bunch of projection code bits here and there that ought to be refactored, except that's much less trivial than the rest of this stuff so I'll save it for another time.

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -1585,6 +1585,18 @@ private:
                                          MeshBase & mesh);
 
   /*
+   * Helper method for the above two to count + distriubte SCALAR dofs
+   */
+  void distribute_scalar_dofs (dof_id_type & next_free_dof);
+
+#ifdef DEBUG
+  /*
+   * Internal assertions for distribute_local_dofs_*
+   */
+  void assert_no_nodes_missed(MeshBase & mesh);
+#endif
+
+  /*
    * A utility method for obtaining a set of elements to ghost along
    * with merged coupling matrices.
    */
--- a/include/mesh/mesh_function.h
+++ b/include/mesh/mesh_function.h
@@ -354,6 +354,14 @@ protected:
                                        const std::set<subdomain_id_type> * subdomain_ids = nullptr) const;
 
   /**
+   * Helper function for finding a gradient as evaluated from a
+   * specific element
+   */
+  void _gradient_on_elem (const Point & p,
+                          const Elem * element,
+                          std::vector<Gradient> & output);
+
+  /**
    * The equation systems handler, from which
    * the data are gathered.
    */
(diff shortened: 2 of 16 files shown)


lindsayad:
Did we truly lose all getpot coverage?

roystgnr:
I didn't change anything in getpot.h, I didn't change anything that used GetPot, and the getpot. hyperlink in our coverage diff is broken, so I'm pretty confident that's a false positive.

No idea where it came from, though.  We only generate coverage data based on libMesh tests, so the MOOSE python thing (or anything in MOOSE) shouldn't have affected the results.

roystgnr:
Even the full coverage report for getpot.h looks like nonsense here.

src/systems/system_projection.C
```diff
@@ -1888,7 +1888,7 @@ void BoundaryProjectSolution::operator()(const ConstElemRange & range) const
                             if (!dof_is_fixed[j])
                               freej++;
                           }
-                        Fe(freei) += phi[i][qp] * fineval * JxW[qp];
+                        Fe(freei) += (fineval * phi[i][qp]) * JxW[qp];
```
lindsayad, line 1891: why this change?

lindsayad (approved):
Most seem like very straightforward refactoring, and the more complex refactoring look like very nice improvements to me.

src/systems/system_projection.C
roystgnr: Why is it in that commit?  Because I stupidly had it sitting in my working tree when I did a `git commit -a` on the matrix stuff.  Can't believe I didn't notice it in the list of changed files while writing the commit message.

I'll take it out of the PR entirely; it didn't come to anything in the end.

Why did I make the change in the first place?  Because I was in the process of preparing projection code (perhaps the most often cut-and-paste-and-tweaked code in all of libMesh and MOOSE...) for refactoring, and my first step in the workflows here was generally "make any trivial changes as necessary to get the [Copy/Paste Detector](https://pmd.github.io/latest/pmd_userdocs_cpd.html) to detect longer duplicate blocks", rather than relying on my own reading to be sure that two blocks are duplicates the whole way through.

roystgnr: Thanks for the review, BTW.  Repeatedly looking at duplicate-except-for-whitespace-and-comment code blocks makes my eyes want to glaze over, so while I tried to be careful about bugs I'm definitely happier with more eyes on the diffs.

roystgnr:
I'll let this sit for the weekend unless John finds a chance to look at it earlier.  I do want to get it into the next MOOSE submodule update next week, but before that, nothing here is more urgent than the new support for currently-never-used PetscLinearSolver feature combinations.

PR #3387, opened 2022-08-31
--disable-exodus build fixes

roystgnr:
This fixes the regressions for that configuration from #3233 for me.


roystgnr:
Added the No Exodus and No Optional recipes to CI.

PR #3388, opened 2022-08-31
Add RawType and ReplaceAlgebraicType to TypeTensor/Vector

grmnptr:
To enable flexible templated harmonic interpolation in MOOSE, we need specializations of MetaphysicL::RawValue for TypeTensor/TypeVector.

### DIFF
--- a/tests/numerics/type_tensor_test.C
+++ b/tests/numerics/type_tensor_test.C
@@ -23,7 +23,9 @@ public:
   CPPUNIT_TEST(testRotation);
 #endif
   CPPUNIT_TEST(testIsZero);
-
+#ifdef LIBMESH_HAVE_METAPHYSICL
+  CPPUNIT_TEST(testReplaceAlgebraicType);
+#endif
   CPPUNIT_TEST_SUITE_END();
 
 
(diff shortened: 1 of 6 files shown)


lindsayad:
LGTM but would be nice to get a unit test

roystgnr (approved):

tests/numerics/type_tensor_test.C
```diff
@@ -23,7 +23,9 @@ public:
   CPPUNIT_TEST(testRotation);
 #endif
   CPPUNIT_TEST(testIsZero);
-
+#ifdef LIBMESH_HAVE_METAPHYSICL
+  CPPUNIT_TEST(testCompareTypes);
```
lindsayad, line 27: If you want to wrap these in `ifdef LIBMESH_HAVE_METAPHYSICL` I would suggest renaming to `testReplaceAlgebraicType` since `CompareTypes` is always defined in libMesh.

roystgnr:
@grmnptr - just sent you an invite for the libMesh Associates team, so CI will run automatically in the future rather than waiting for me to notice those "Developer needed to activate" messages.

tests/numerics/type_tensor_test.C
grmnptr: Renamed them!

PR #3389, opened 2022-08-31
Add TensorTraits

lindsayad:
Currently only holds a `rank` attribute, but there could be room for other things in the future perhaps. This would be useful for user code that wants to switch what it's doing for different types of mathematical structures, e.g. in @grmnptr's harmonic interpolation in idaholab/moose#21928

### DIFF
--- a/include/base/libmesh_common.h
+++ b/include/base/libmesh_common.h
@@ -697,6 +697,13 @@ inline Tnew libmesh_cast_int (Told oldvar)
   return cast_int<Tnew>(oldvar);
 }
 
+/**
+ * This is a helper variable template for cases when we want to use a default compile-time
+ * error with constexpr-based if conditions. The templating delays the triggering
+ * of the static assertion until the template is instantiated.
+ */
+template <class T>
+constexpr std::false_type always_false{};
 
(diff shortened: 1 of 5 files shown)


roystgnr (approved):

lindsayad:
I'm pretty curious why code coverage did not change. What flags are we using when compiling and testing for coverage? Are we optimized and inlining? I know template lines that are not instantiated do not count against us, but I would expect lines that we instantiate and cover should count *for* us.

roystgnr:
Note that there's no diff coverage report (and kudos on the coverage tool for not reporting the diff coverage as NaN or 0/0!).

IIRC, code coverage works at runtime by adding a coverage-reporting utility to every function that gets called ... but here we have no functions getting called.  Not even a pre-main() static initialization constructor is getting called, since everything here is constexpr and/or POD.

Which kind of sucks, when you think of the implications: no way to measure code coverage of any compile-time metaprogramming, except indirectly insofar as the metaprogram lines get used in runtime program lines which themselves get covered.  I don't think it's worth adding runtime lines to make those measurements, though, especially since we'd have to put them in include/ or src/ rather than tests/.

PR #3398, opened 2022-09-12
Add partition-of-unity testing

lindsayad:
This exercise was primarily undertaken to increase my understanding of our implemented shape functions. I'm curious why in the literature (the literature I've found at least) the partition-of-unity property (or lack thereof) is relatively little discussed. It seems like this property (or lack thereof) has very important consequences for being able to measure conservation ... or am I off here?

### DIFF
--- a/tests/fe/fe_test.h
+++ b/tests/fe/fe_test.h
@@ -22,6 +22,7 @@
 
 #define FETEST                                  \
   CPPUNIT_TEST( testU );                        \
+  CPPUNIT_TEST( testPartitionOfUnity );         \
   CPPUNIT_TEST( testGradU );                    \
   CPPUNIT_TEST( testGradUComp );                \
   CPPUNIT_TEST( testHessU );                    \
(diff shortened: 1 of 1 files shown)


jwpeterson:
I'm not aware of what the connection between conservation and partition-of-unity property would be, can you elaborate? I have seen the partition of unity property used in super-convergent flux recovery post-processing techniques, which seems related but not exactly the same thing.

lindsayad:
It's not something I've tested, but I guess intuitively if I have a weak form like this:
```
  return _test[_i][_qp] * _f(_q_point[_qp], _t);
```
then if the test functions at a given qp do not sum to unity, then I may impart a total source/sink to the nonlinear system that is not equivalent to a naive global integration of `_f(_q_point[_qp], _t)`

jwpeterson:
OK, well I think the important thing with regard to "conservation" is having the constant 1 in the the test space. For "local" conservation, you have to have "the constant function 1 on the element and zero outside the element", which we don't have in general for continuous FE bases.

jwpeterson:
In your example, there would presumably also be a  `_test[_i][_qp]` on the "left-hand side" of the equation, so even if the test functions did not sum to 1, you would also have the same non-unity sum on the left-hand side and it would "cancel".

lindsayad:
The impetus for me thinking about this was https://github.com/idaholab/moose/discussions/21966 where I am supporting the global conservation of CGFE for fluid-flow calculations. 

Your responses show that I'm not asking the right question. As shown in https://github.com/idaholab/moose/discussions/21966 I can demonstrate/defend global conservation of CGFE by doing simple element integral postprocessing of individual terms (in that example forcing-function, reaction, and advection terms), summing the results, and showing that the sum is near machine epsilon. Those element integral postprocessors do not involve the test functions, and they do not need to, I believe, because of the partition-of-unity property of the Lagrange shape functions.

My real question is: if I am using shape functions that do not have the partition-of-unity property, what is the best way to demonstrate that the finite element discretization is globally conservative? Perhaps it may not be?

jwpeterson:
>  if I am using shape functions that do not have the partition-of-unity property, what is the best way to demonstrate that the finite element discretization is globally conservative?

I'd do essentially the same thing that you did on idaholab/moose#21966 for the global test function v=1, but not assuming partition of unity. Then you have the same situation that "terms multiplied by the gradient of the test function do not contribute to global conservation", but instead of getting e.g.

advection + reaction - reaction = 0,

you'll get:

C*advection + C*reaction - C*source = 0

for some C != 1.

jwpeterson:
> Those element integral postprocessors do not involve the test functions, and they do not need to, I believe, because of the partition-of-unity property of the Lagrange shape functions.

I agree with you. To demonstrate global conservation for non p-o-u bases I guess the postprocessors would need to involve the test function values.

roystgnr (approved):

PR #3404, opened 2022-09-29
Better checking in System::add_variable()

jwpeterson:
This fixes an issue we ran into where a variable with the same name and `FEType` was added to a `System` twice, but with different subdomain restrictions each time. Previously, the code would simply return the number of the existing variable, regardless of whether it had the same subdomain restriction as the variable just added. The previous code therefore effectively ignored attempts by the user to change the subdomain restriction of existing variables, but this new patch makes it an error to add the same variable twice with different subdomain restriction, since IMO this is less confusing than trying to debug why a variable has no DOFs on a subdomain where you expected it to have some. 

It's possible this will cause some tests to fail... but if so I would argue that those tests were only working by accident/not fully understood in the first place.

### DIFF
--- a/src/systems/system.C
+++ b/src/systems/system.C
@@ -1217,16 +1217,6 @@ unsigned int System::add_variable (std::string_view var,
                                    const FEType & type,
                                    const std::set<subdomain_id_type> * const active_subdomains)
 {
-  // Debugging
-  libMesh::out << "Called System::add_variable(), with var = " << var << std::endl;
-  if (active_subdomains)
-  {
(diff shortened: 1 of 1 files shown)


roystgnr (approved):
Could you clean up the history?  Or do you want to leave the debugging bits accessible if they're needed again in the future?

jwpeterson:
> if they're needed again in the future?

My plan was to leave them in the history, yeah. It was just non-trivial enough that I didn't want to have to write it again if there ended up being some issue.

PR #3406, opened 2022-10-03
Fix and unit test write_added_sides(true) with discontinuous writes

roystgnr:
I'd written a code path for this, but I'd missed it in that first wave of unit tests, so of course it didn't work.  And then when I tried to use this feature in an actual solve, where discontinuous output is a main use case, it did not go well.


roystgnr:
Well, that's disheartening.  I swear these are passing on my build.  Wonder if I've again managed to write something that works on DistributedMesh and fails on ReplicatedMesh.

roystgnr:
Yup.  ReplicatedMesh build fails.

jwpeterson (approved):

PR #3407, opened 2022-10-06
Get function calls out of for loop tests

roystgnr:
Refs #1053

I haven't touched the reduced_basis code here, the ratio of importance to CI coverage there is way too high to futz with for a microoptimization, but this cleans up all the other redundantly-called functions I could find in include/ and src/

None of this (except maybe the new RawType specializations?) is likely to be in an inner kernel, so no rush to merge this, I just noticed a few of these cases while I was looking into vectorization reports.


roystgnr:
More unused variable warnings.  Not sure how I let myself start working on a build without `--enable-werror` turned on, but I'll fix that and get back to this.

jwpeterson (approved):

PR #3410, opened 2022-10-18
Basic Prism20, Prism21 element support

roystgnr:
There's still work to be done here (Prism20 full-order Lagrange shape functions, embedding matrices for AMR, connectivity for Tecplot and/or VTK if possible), but this is now at the point where my workstation sees a hundredish new unit tests passing and I'd like to see if CI agrees.


roystgnr:
PRISM20 is working for me too now

jwpeterson (approved):
Do the coverage checks give us any idea what the remaining 30+% of the untested lines of code actually are?

src/quadrature/quadrature_nodal_3D.C
```diff
@@ -122,6 +122,82 @@ void QNodal::init_3D(const ElemType, unsigned int)
...
+        // Dropping the xi^2 zeta^2 constraint gives us a rank-1 null
+        // space to play with ... but adding anything from that null
+        // space to push wte closer to positive just pushes wse
+        // negative.
+        //
+        // Dropping exact integration of xi^3 type terms gives us a
+        // rank-2 null space.  I just found the max(min(weight)) from
+        // that solution space using Octave.  Someone who cares more
+        // than me might want to repeat this exercise with better than
+        // double precision...
```
jwpeterson, line 154: I'll also check what "nodal quadrature mass matrix optimization" approach for this case gives when I get a chance.

roystgnr: That'd be great!  I didn't quite hit the uncanny valley then, huh?

roystgnr:
Uncovered lines include error messages we never expect to hit (but which we leave active in opt mode because they're in switch statement default cases so we never expect a branch to slow us down either), a ton of stuff for adaptivity (which I can't cover yet, because it's not yet correct...), some combinations of FEFamily+order (which I guess we should cover; at one point I was trying to do a sparse representative subset there to keep our unit tests quick to run, but I guess at this point the "unit tests quick to run" ship has sailed away, reached another continent, been decommissioned, and turned into a museum.  "At one point the unit tests were quick to run!" is now a museum plaque prompting bored skeptical kids to inquire "Mommy, can that really be true?"  I think I still want to save exhaustive FE testing for another PR, though, in case it pushes us over any timeout limits.

Looks like there are some things we could cover and run quickly, though: I see a few unit tests that never got refactored or extended for recent element types that would get things like `local_side_node()` and `local_edge_node()`, likewise for nodal quadrature ... oh, ouch, and there *is* a test that got properly refactored where I just didn't add a new instantiation.  I'll fix all that before we merge.

roystgnr:
I was *right* about to merge this before I noticed I hadn't actually registered my new testNodalQuadrature with CppUnit.

PR #3414, opened 2022-10-24
Pyramid quadrature bug fixes and unit tests

roystgnr:
The QGrid+Pyramid bug was embarrassing, even for a case nobody uses.

At some point I just need to start going through our full code coverage report and getting those numbers up to par.

### DIFF
--- a/tests/quadrature/quadrature_test.C
+++ b/tests/quadrature/quadrature_test.C
@@ -270,6 +270,17 @@ private:
     return exactz * tri_integrals(modex, modey, 0);
   };
 
+  const std::function<Real(int,int,int)> pyramid_integrals =
+  [this](int modex, int modey, int modez) {
+
+    const int binom = Utility::binomial(modex+modey+modez+3, modez);
+
+    if (modex%2 || modey%2)
+      return Real(0);
+
+    return Real(4)/((modex+1)*(modey+1)*binom*(modex+modey+modez+3));
+  };
+
 
 public:
   void setUp ()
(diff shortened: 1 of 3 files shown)


tests/quadrature/quadrature_test.C
```diff
@@ -512,20 +528,28 @@ public:
...
+        testPolynomials(qtype, order, PYRAMID14, pyramid_integrals, 0);
+      }
     // QSimpson doesn't get all quadratics on a simplex or its extrusion
     else if (qtype == QSIMPSON)
       {
         testPolynomials(qtype, order, TET10, tet_integrals, std::min(1u,exactorder));
         testPolynomials(qtype, order, PRISM15, prism_integrals, std::min(1u,exactorder));
+
+        // And on pyramids we gave up and redid QTrap
+        testPolynomials(qtype, order, PYRAMID14, pyramid_integrals, std::min(1u,exactorder));
```
jwpeterson, line 546: So... if I understand correctly, this works because we don't have an FE involved and we don't try to evaluate e.g. `dphi[qp]` or anything that would involve the inverse Jacobian in the test code...? I did not feel comfortable providing a `QNodal` implementation for the Pyramid elements because this kind of implies that you can use it for "normal" FE assembly code, but trying to do so in the case of `QNodal` would lead to division-by-zero or NaN type errors rather than a "not implemented" error.

roystgnr: Yeah, I'm not seeing a fix here as a high priority.  Nodal quadrature on Pyramids, except I guess in the very special case of a mass matrix integrated separately from the rest of your physics, just seems like a mistake.

jwpeterson: > just seems like a mistake.

I agree, so what I'm suggesting is that we just don't allow nodal quadrature on pyramids in the first place, by having a `libmesh_not_implemented();` error in that case.

jwpeterson: What about making it accessible depending on the value of a flag, similar to what we currently do with
```
  bool allow_rules_with_negative_weights;
```
That one defaults to `true`, since in the common use case I don't think people often care, but I would prefer to have
```
  bool allow_nodal_pyramid_quadrature; 
```
default to `false`, and let people turn it on for specialized applications (explicit mass matrix lumping) once they are aware of the consequences.

jwpeterson: If that sounds reasonable, I would implement it, not saying it needs to be in this PR.

roystgnr: Oh, yeah, that sounds like a great idea.

jwpeterson (approved):

PR #3415, opened 2022-10-24
More test coverage, NumericVector subclass fixes

roystgnr:
Finding that QGrid+Pyramid bug spooked me.  It's going to take forever until we have really solid test coverage, but I'll try to find time every now and then to take a pass at whatever the lowest-hanging fruit is.


roystgnr:
From 56.46% to 59.48%!  One more PR like this and our coverage will be worthy of a "D-" or even a "D" instead of an "F"!

Issue #3416, opened 2022-10-25
Installation error - build error


XuanQuangPham91:
Hello, 

I have a problem with the installation.

I have downloaded and installed: petsc-3.18.0 and slepc-3.18.0. After that, I download and install libmesh-1.7.1 following instructions in the [link](https://libmesh.github.io/installation.html). I meet the error below while doing the "make" step before the "make install" step.

In file included from /opt/local/lib/petsc-3.18.0/include/petscsys.h:1189,
                 from /opt/local/lib/petsc-3.18.0/include/petscbag.h:5,
                 from /opt/local/lib/petsc-3.18.0/include/petsc.h:5,
                 from ./include/libmesh/petsc_macro.h:66,
                 from src/solvers/petscdmlibmeshimpl.C:20:
src/solvers/petscdmlibmeshimpl.C: In function ‘PetscErrorCode DMSetUp_libMesh(DM)’:
src/solvers/petscdmlibmeshimpl.C:904:120: error: ‘struct _p_PetscObject’ has no member named ‘type’
  904 |     LIBMESH_SETERRQ2(((PetscObject)dm)->comm, PETSC_ERR_ARG_WRONG, "DM of type %s, not of type %s", ((PetscObject)dm)->type, DMLIBMESH);      
|                                                                                                                        ^~~~
/opt/local/lib/petsc-3.18.0/include/petscerror.h:115:127: note: in definition of macro ‘SETERRQ’
  115 |     PetscErrorCode ierr_seterrq_petsc_ = PetscError(comm, __LINE__, PETSC_FUNCTION_NAME, __FILE__, ierr, PETSC_ERROR_INITIAL, __VA_ARGS__); \    
|                                                                                                                               ^~~~~~~~~~~
src/solvers/petscdmlibmeshimpl.C:904:5: note: in expansion of macro ‘LIBMESH_SETERRQ2’
  904 |     LIBMESH_SETERRQ2(((PetscObject)dm)->comm, PETSC_ERR_ARG_WRONG, "DM of type %s, not of type %s", ((PetscObject)dm)->type, DMLIBMESH);
      |     ^~~~~~~~~~~~~~~~
  CXX      src/systems/libmesh_devel_la-diff_context.lo
  CXX      src/systems/libmesh_devel_la-diff_system.lo
make[1]: *** [Makefile:21925: src/solvers/libmesh_devel_la-petscdmlibmeshimpl.lo] Error 1
make[1]: *** Waiting for unfinished jobs....
[...]

PR #3417, opened 2022-10-26
Petsc 3.18 compatibility fixes

roystgnr:
Refs #3416

### DIFF
--- a/src/solvers/petscdmlibmeshimpl.C
+++ b/src/solvers/petscdmlibmeshimpl.C
@@ -342,7 +342,7 @@ static PetscErrorCode  DMCreateFieldDecomposition_libMesh(DM dm, PetscInt * len,
         PetscInt elen, dlen;
         ierr = ISGetLocalSize(emb, &elen); CHKERRQ(ierr);
         ierr = ISGetLocalSize(dis, &dlen); CHKERRQ(ierr);
-        if (elen != dlen) LIBMESH_SETERRQ1(((PetscObject)dm)->comm, PETSC_ERR_PLIB, "Failed to embed subdomain %D", d);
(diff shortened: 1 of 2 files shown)


src/solvers/petscdmlibmeshimpl.C
```diff
@@ -958,7 +958,7 @@ PetscErrorCode  DMCreate_libMesh(DM dm)
 
   PetscFunctionBegin;
   PetscValidHeaderSpecific(dm,DM_CLASSID,1);
-  ierr = PetscNewLog(dm,&dlm);CHKERRQ(ierr);
+  ierr = PetscNew(&dlm);CHKERRQ(ierr);
```
jwpeterson, line 961: No if-defing required here, eh? I guess `PetscNew` should have just been used whenever this was originally written...

jwpeterson (approved):
This looks good to me. I have been wanting to make a PETSc 3.18 build and test it for a while... now that you have already fixed the libmesh compatibility stuff I will try and do that sooner rather than later.

src/solvers/petscdmlibmeshimpl.C
roystgnr: It was written in 2012, and Jed was recommending PetscNewLog to users [as late as 2016](https://scicomp.stackexchange.com/questions/23471/petsc-difference-between-petscnew-and-petscmalloc#comment37122_23471).  I assume they made `PetscNew` do more sophisticated logging without user assistance at some point since, but this code was fine at the time.

roystgnr:
I was hoping to hear from the folks in #3416, but I'll merge now since it works for me.  Hopefully if there are still incompatibilities with other configure options we'll again get pinged as people come across them.

PR #3426, opened 2022-11-03
VTK 9.2 linking fix

roystgnr:
This replaces #3346


jwpeterson (approved):

capitalaslash:
any chance this could be backported to the v1.7.1 branch?

jwpeterson:
> any chance this could be backported to the v1.7.1 branch?

I think it should be possible. It looks like there haven't been any other changes to vtk.m4 so the likelihood of a conflict is small. This would then become a 1.7.2 tag... does that work for you?

capitalaslash:
sure!
I am currently cherry-picking it manually but for deployed applications they require me to use a fixed version.

PR #3432, opened 2022-11-08
Import ray tracing routines

loganharbour:


loganharbour:
@roystgnr everything here is complete and fully tested - this is definitely only a chunk of it, but could be reviewed now.

loganharbour:
I'll also add that I think I have the most of the non-tool changes made, that is the stuff to `Elem` and the derived classes. That was my goal here - get those in first, and then add a few tools.

include/geom/elem.h
```diff
@@ -752,18 +752,27 @@ class Elem : public ReferenceCountedObject<Elem>,
...
   /**
    * \returns the (local) node numbers on the specified edge
    */
   virtual std::vector<unsigned int> nodes_on_edge(const unsigned int /*e*/) const = 0;
 
+  /**
+   * \returns A pointer to the beginning of the array that contains the
+   * (local) node numbers on the edge \p e
+   *
+   * This does not create temproraries, unlike \p nodes_on_edge.
```
roystgnr, line 771: temporaries

include/geom/elem.h
```diff
@@ -745,6 +745,15 @@ class Elem : public ReferenceCountedObject<Elem>,
    */
   virtual std::vector<unsigned int> nodes_on_side(const unsigned int /*s*/) const = 0;
 
+  /**
+   * \returns A pointer to the beginning of the array that contains the
+   * (local) node numbers on the specified side
+   *
+   * This does not create temproraries, unlike \p nodes_on_side.
```
roystgnr, line 757: temporaries

include/geom/elem.h
```diff
@@ -2053,6 +2062,17 @@ class Elem : public ReferenceCountedObject<Elem>,
...
   }
 
+  /**
+   * Helper for overriding nodes_on_side in derived classes.
+  */
+  template <class ElemClass>
+  const unsigned int * _nodes_on_edge_ptr(const unsigned short s) const
+  {
+    static_assert(std::is_base_of<Elem, ElemClass>::value, "Not an Elem");
+    libmesh_assert_less(s, std::extent<decltype(ElemClass::edge_nodes_map)>::value);
```
roystgnr, line 2072: Should we throw in a static assert here to make sure that `std::extent<decltype(ElemClass::edge_nodes_map)>::value` is also consistent with `ElemClass::nodes_per_edge`?   (and similarly in `_nodes_on_side_ptr`)

include/geom/intersection_tools.h
```diff
@@ -69,6 +69,19 @@ WithinSegmentResult within_segment(const Point & s1,
                                    const Point & p,
                                    const Real tol = TOLERANCE);
 
+/**
+ * Checks whether or not the given points are collinear
+ * @param p1 The first point
+ * @param p2 The second point
+ * @param p3 The third point
+ * @param tol The tolerance to use
```
roystgnr, line 79: Here's another one where tol isn't specified as relative or absolute and where we're not worrying about units.

Here it's clearer that performance might be an issue (though I bet with `norm_sq()` you'd be fine) but I'd still prefer a relative tolerance and I'd still insist that we document an absolute tolerance.  I probably sound like an ass about this, but really I should be sounding like an ass and rattling heavy chains: I feel like the Ghost Of Screwed Up Displaced Mesh Problems Past.

src/geom/intersection_tools.C
```diff
@@ -76,6 +78,71 @@ bool collinear(const Point & p1,
...
+  libmesh_assert_less(s, elem.n_sides());
+  libmesh_assert(extrema.is_invalid());
+  libmesh_assert_equal_to(elem.dim(), 3);
+
+  // For higher order than linear without linearization, make sure
+  // that our edges are collinear
+  if (elem.default_order() > 1 && !linearize)
+    for (const auto e : elem.edge_index_range())
+    {
+      // we should expect 3 edges for our higher order elems
```
roystgnr, line 97: 3-noded edges

roystgnr: And actually, could you just assert that `n_nodes_on_edge > 3`?  We do hope to support higher-order Lagrange and Rational Bezier-Bernstein one of these days.

roystgnr (changes requested):

include/geom/elem_extrema.h
```diff
@@ -0,0 +1,176 @@
...
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+
+
+#ifndef LIBMESH_ELEMEXTREMA_H
+#define LIBMESH_ELEMEXTREMA_H
+
+#include "libmesh/libmesh_common.h"
+
+#if LIBMESH_DIM > 1
```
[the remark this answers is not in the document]
loganharbour: Added 1D support for being at a vertex

src/geom/intersection_tools.C
```diff
@@ -0,0 +1,67 @@
...
+                                   const Real tol)
+{
+  libmesh_assert(!s1.absolute_fuzzy_equals(s2, tol));
+  libmesh_assert_less(std::abs((s1 - s2).norm() - length), tol);
+
+  const auto diff1 = p - s1;
+  const auto diff2 = p - s2;
+  const auto tol_scaled = tol * length;
+
+  if (diff1 * diff2 > tol_scaled)
```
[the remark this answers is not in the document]
loganharbour: Changed to a relative tolerance

include/geom/intersection_tools.h
```diff
@@ -0,0 +1,75 @@
...
+                         NOT_WITHIN = 0,
+                         AT_BEGINNING = 1,
+                         AT_END = 2,
+                         BETWEEN = 3 };
+
+/**
+ * Checks whether or not a point is within a line segment
+ * @param s1 The first point on the segment
+ * @param s2 The second point on the segment
+ * @param length The segment length (for optimization if it's already computed)
```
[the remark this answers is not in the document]
loganharbour: Decided to just get rid of the optimization - if this shows up in profiling for real tracing I'll implement an optimization

include/geom/elem.h
loganharbour: I do like this but unfortunately it won't work because in cases like `Edge`, we don't have an `edge_nodes_map`

loganharbour:
@roystgnr - good to go.

loganharbour:
Re the tolerance stuff: the last few tests run these scaled with element features on the order of [1e-8,1e8]

src/geom/intersection_tools.C
```diff
@@ -150,6 +150,32 @@ bool within_edge(const Elem & elem,
   return false;
 }
 
+unsigned short at_vertex(const Elem & elem,
+                         const Point & p,
+                         const Real tol)
+{
+  for (const auto v : elem.vertex_index_range())
+    if (elem.point(v).relative_fuzzy_equals(p, tol))
```
roystgnr, line 158: This isn't a proper relative tolerance.  Here `relative_fuzzy_equals` will use `p+elem.point(v)` to compare against, which is likely to be way too big in some cases (a tiny element far from the origin) and is going to be effectively *zero* in other cases (a huge element with a vertex at the origin).

I'd normally use `elem.hmax()` as a scale here, but that might be too expensive for you?  If so, it wouldn't be unreasonable to have something like `Elem::cheap_scale()` that returns something like an l1 norm (how do we *not* already have that for `TypeVector`...) of `(elem.point(1)-elem.point(0))`.  Kind of arbitrary for anisotropic meshes but so is the choice of `hmax()` over `hmin()`.

src/geom/intersection_tools.C
```diff
@@ -150,6 +150,32 @@ bool within_edge(const Elem & elem,
...
+                                 const Point & p,
+                                 const unsigned short s,
+                                 const Real tol)
+{
+  libmesh_assert_less(s, elem.n_sides());
+  const unsigned int * side_node_map = elem.nodes_on_side_ptr(s);
+  for (const auto i : make_range(elem.n_vertices_on_side(s)))
+  {
+    const auto v = side_node_map[i];
+    if (elem.point(v).relative_fuzzy_equals(p, tol))
```
roystgnr, line 173: Similar tolerance issue here.

roystgnr (changes requested):

gridley:
This is exciting! We'll be able to add volumetric mesh geometries to OpenMC after this is merged (at the moment only surface mesh geometries are allowed). OpenMC already optionally links to libmesh for volume mesh tallies, but this will make multiphysics coupling substantially more straightforward. :smile:

loganharbour:
Oh hi @gridley. Yeah, this is completely OpenMC driven after some conversations with @pshriwise. I've got another branch pulling over the rest of the routines, just haven't made a PR yet.  On that note - thanks for the reminder to finish this up 😆

PR #3436, opened 2022-11-10
Add NodeElem contains_point() and close_to_point() implementations

jwpeterson:
The base class versions of these functions return false for `NodeElem`, even when the `Point` passed in only differs from the `NodeElem`'s position by floating point noise. The main difference is that the tolerance passed to `NodeElem::contains_point()` is used as an absolute tolerance rather than a relative tolerance, since, when you only have a single point, there's no obvious scale (like `hmax()` for an `Elem`) to use.

### DIFF
--- a/include/geom/node_elem.h
+++ b/include/geom/node_elem.h
@@ -274,6 +274,24 @@ public:
     return INVALID_ELEM;
   }
 
+  /**
+   * \returns \p true if the point p is within a distance of "tol" from
+   * the point representing this element, false otherwise.
+   *
+   * The tolerance "tol" is normally treated as a relative tolerance in
+   * contains_point() checks, but in this case there is no relevant length
+   * to use in determing a relative tolerance, so "tol" is treated as an
+   * absolute tolerance. The NodeElem contains_point() and close_to_point()
+   * implementations are identical, whereas they differ for other element
+   * types.
+   */
+  virtual bool contains_point(const Point & p, Real tol) const override;
+
+  /**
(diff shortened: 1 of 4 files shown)


roystgnr (approved):

jwpeterson:
There's a failing unit test
```
!!!FAILURES!!!
Test Results:
Run:  3595   Failures: 1   Errors: 0


1) test: MeshInputTest::testExodusFileMappingsTwoBlocks (F) line: 1233 ../../tests/mesh/mesh_input.C
assertion failed
- Expression: located_elem == elem
```
but I can't reproduce it locally.

jwpeterson:
Just to recap what @roystgnr and I figured out here: the failing unit test was an IGA mesh, so NodeElems were being created on the fly, hence the relevance to this PR. In addition to being an IGA test, the mesh also contains overlapping nodes, and therefore created overlapping `NodeElems` which differed in position by floating point noise. The previous `Elem::contains_point()` check for this mesh passed for only one of the overlapping NodeElems (the one that compared floating point equal) but the new test (which uses an absolute tolerance of 1.e-6) passes for both overlapping NodeElems. The fix for the unit test was to use the PointLocator API that returns a set of candidate Elems rather than a single Elem.

jwpeterson:
:+1: I can't approve my own PR :rofl: but @roystgnr change works for me as well using the Nanoflann based PointLocator, so I will merge once CI is happy.

roystgnr (approved):

PR #3439, opened 2022-11-10
Xdr string_view comments

roystgnr:
This is built on top of (and I noticed the opportunity while making) #3438, but it's not an urgent fix like that one is, just an API upgrade I missed while hunting for good string_view opportunities in #3204.

### DIFF
--- a/include/numerics/laspack_vector.h
+++ b/include/numerics/laspack_vector.h
@@ -35,7 +35,6 @@
 #include <qvector.h>
 
 // C++ includes
-#include <cstdio> // for std::sprintf
 #include <limits>
 #include <mutex>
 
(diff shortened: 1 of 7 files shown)


loganharbour:
Is this ready? :-)

roystgnr:
It's ready, if you want to review.  The important commit is already merged via #3438, though.  That's enough to fix the deprecation issues you saw, right?

I usually don't merge my own feature update PRs unless they've gotten a review or I've let them sit a week or so to give opportunity for review.  Bug fixes I'll get in ASAP, but for anything that's not urgent I like to have ~~someone to share the blame~~ *another set of eyes* for bugs.

PR #3442, opened 2022-11-14
Use boundary_id_type for nodeset/sideset counter types in Abaqus mesh reader

jessecarterMOOSE:
Counting Abaqus nodesets/sidesets was done with a short int, which can overflow for big models. Use `boundary_id_type` instead.


jwpeterson (approved):

jessecarterMOOSE:
Thanks @jwpeterson !

PR #3444, opened 2022-11-21
Use existing BoundaryInfo::BCTuple in a couple more places

jwpeterson:
This was already being done correctly in the .C file, so only the .h file needed updating.


roystgnr:
Well, that's weird.  Jason was working on Civet early this morning but I thought everything went back to status quo hours ago.

PR #3447, opened 2022-11-30
Fix --disable-exceptions builds

roystgnr:
Thanks to @jwpeterson for noticing the failure in our devel->master CI.  This goes a bit farther than the fixes for that; I did a gcc build with `libmesh_CXXFLAGS='-fno-exceptions'` as well and fixed the (non-third-party) errors which cropped up from that too.


jwpeterson:
I attached a "No Exceptions" recipe to this. Not that I don't think it will pass or anything :wink:

jwpeterson:
The Test MOOSE mac failure seems unrelated:
```
[21.42s]     FAIL mooseutils/tests.civet_results FAILED (CRASH)
```

jwpeterson (approved):

PR #3460, opened 2022-12-14
Options for restricting solves to unconstrained DoFs only

roystgnr:
I'm not sure this is the right interface (there's at least some PetscDiffSolver encapsulation that I've broken with `public:` but should redo properly with `friend`), and I'm not sure this is the right implementation (the related code already in PetscLinearSolver was creating a submatrix, but this is mainly for efficiency so maybe we should be building a composite matrix in the first place?), and it's not complete (PetscLinearSolver and NewtonSolver are working; PetscDiffSolver is giving me a zero pivot on the preconditioner; PetscNonlinearSolver isn't implemented).

But the pre-vacation work year is starting to run low for me, so I'm putting this up to see if CI manages to see even more problems than I do.

### DIFF
--- a/examples/adjoints/adjoints_ex3/adjoints_ex3.C
+++ b/examples/adjoints/adjoints_ex3/adjoints_ex3.C
@@ -790,6 +790,17 @@ int main (int argc, char ** argv)
 
   equation_systems.init ();
 
+  if (infile.search("--restrict_solves_to_unconstrained"))
+    {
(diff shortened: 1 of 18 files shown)


roystgnr:
So not only is the new PetscDiffSolver option not working, the old functionality has regressed.  Didn't even think to check for that by hand...

roystgnr:
This doesn't seem to be the right way to go about this, since it won't work for the nonlinear case.  I'll try again with something based on Alex's condensed matrix work.

PR #3461, opened 2022-12-19
Update to check_dirichlet_bcid_consistency()

dknez:
Previously check_dirichlet_bcid_consistency() required that all processors had DirichletBoundary objects defined identically, otherwise one could get a parallel deadlock due to calling comm.max() on a subset of processors. In this update we do a comm.set_union() of boundary IDs with DirichetBoundary defined on them so that we can define DirichletBoundary objects differently on different processors. This is desirable since in general other libMesh mesh data is not required to be identical on different processors, and this update to check_dirichlet_bcid_consistency() means that DirichletBoundary also no longer needs to be identical on different processors.

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -845,6 +845,11 @@ public:
   void set_implicit_neighbor_dofs(bool implicit_neighbor_dofs);
 
   /**
+   * Set the _verify_dirichlet_bc_consistency flag.
+   */
+  void set_verify_dirichlet_bc_consistency(bool val);
+
+  /**
    * Tells other library functions whether or not this problem
    * includes coupling between dofs in neighboring cells, as can
    * currently be specified on the command line or inferred from
@@ -1924,6 +1929,18 @@ private:
    */
   bool _implicit_neighbor_dofs_initialized;
   bool _implicit_neighbor_dofs;
+
+  /**
+   * Flag which determines whether we should do some additional
+   * checking of the consistency of the DirichletBoundary objects
+   * added by the user. Defaults to true, but can be disabled in cases
+   * where you only want to add DirichletBoundary objects "locally"
+   * and can guarantee that no repartitioning will be done, since
+   * repartitioning could cause processors to own new boundary sides
+   * for which they no longer have the proper DirichletBoundary
+   * objects stored.
+   */
+  bool _verify_dirichlet_bc_consistency;
 };
 
 
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -1381,14 +1381,11 @@ public:
   get_adjoint_dirichlet_boundaries(unsigned int q);
 
   /**
-   * Check that all the \p dbc_bcids (which typically come from DirichetBoundary objects)
-   * are actually present in the mesh. If not, this will throw an error.
-   * This method is useful for detecting the case where DirichletBoundary objects have
-   * been defined using IDs that are not present in the mesh, since this usually indicates
-   * that there is some error in the model setup.
+   * Check that all the ids in dirichlet_bcids are actually present in the mesh.
+   * If not, this will throw an error.
    */
   void check_dirichlet_bcid_consistency (const MeshBase & mesh,
-                                         const std::set<boundary_id_type>& dbc_bcids) const;
+                                         const DirichletBoundary & boundary) const;
 #endif // LIBMESH_ENABLE_DIRICHLET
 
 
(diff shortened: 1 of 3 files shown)


jwpeterson (approved):
Technically this is an API change, but I think not in a widely used API, so I would be good with merging it. The function is set up to handle the case where boundary id sets are not consistent among ranks
```
      // On a distributed mesh, boundary id sets may *not* be
      // consistent across all ranks, since not all ranks see all
      // boundaries
      mesh.comm().max(found_bcid);
```
but if that is actually the case, then you cannot loop over `_dirichlet_boundaries` and call `comm().max()` as it will result in a deadlock, like @dknez mentioned.

roystgnr:
> this update to check_dirichlet_bcid_consistency() means that DirichletBoundary also no longer needs to be identical on different processors.

Does it really?  When we redistribute elements to other processors, are we now redistributing any associated DirichletBoundary objects at the same time (as we do with e.g. BoundaryInfo entries)?  If not then we still aren't actually supporting distributed DirichletBoundary information, we're just making its breakage slightly delayed and making fixing it slightly more complicated.

This PR is probably a step in the right direction, so I don't want to say "don't merge it until you've taken the next ten steps too", but could we do something to catch the failure in the repartitioned case?  Even if it's just asserting *then* that boundary data is identical across processors, that would still be preferable to potentially silently dropping BCs.

dknez:
As far as I understand the DirichletBoundary objects lead to creation of dof_constraint_rows, and once those dof_constraint_rows are created, they are communicated in parallel as needed, is that right? Based on this, it seems to me that we should not have to define the same DirichletBoundary on all processors, since all that matters is that the subsequent dof_constraint_rows are communicated properly (which happens automatically already as far as I know). In other cases we aren't required to add the constraints on all processors, since the dof_constraint_rows will be communicated automatically.

As a result it seems to me that we should be allowed to create a DirichletBoundary on one processor and not others, since then the dof_constraint_rows will be communicated as needed. I've tested this and it works fine in parallel as far as I can tell, but it only worked after applying the update from this PR to check_dirichlet_bcid_consistency(), since otherwise we got a parallel deadlock in check_dirichlet_bcid_consistency(), as described above.

dknez:
P.S. The test cases I've run all use ReplicatedMesh. Maybe there are some other issues in the DistributedMesh case, though I think this PR is at least helpful to allow us to add the dof_constraint_rows from DirichletBoundary on a subset of processors.

jwpeterson:
> could we do something to catch the failure in the repartitioned case? Even if it's just asserting then that boundary data is identical across processors, that would still be preferable to potentially silently dropping BCs.

I would be happy to add some additional checks, but I'm not sure exactly what you mean by "repartitioned case"? Is the idea that if you start with inconsistent (aka different local) `DirichletBoundary` information that would be in fact be sufficient to constrain the problem, but then you repartition, you may then end up with processors owning new sides that should be constrained, but not having the right `DirichletBoundary` objects to actually constrain them?

roystgnr:
> all that matters is that the subsequent dof_constraint_rows are communicated properly (which happens automatically already as far as I know)

It's been too long, so don't take my word for this, but IIRC the rows are automatically *communicated* properly, but (at least on a distributed mesh? though we might be doing ReplicatedMesh the same way for efficiency, either now or in the future!) the rows might not be *created* properly; we might only be bothering to create constraint rows as necessary for our local elements.

> Is the idea that if you start with inconsistent (aka different local) DirichletBoundary information that would in fact be sufficient to constrain the problem, but then you repartition, you may then end up with processors owning new sides that should be constrained, but not having the right DirichletBoundary objects to actually constrain them?

This is basically what I'm afraid of.  Not sure what the best way to test it would be.  Maybe use a linear partitioner, then renumber elements and repartition, to force nearly everything to be reassigned?

jwpeterson:
After some discussion with @roystgnr and @dknez we decided to switch approaches slightly and allow the "expert" user to simply circumvent the `check_dirichlet_bcid_consistency()` check entirely in cases where the "repartition issue" is not relevant.  Although such a check would be possible in theory according to @roystgnr:

> (do a global set union of pair<boundary_id_type,variable_num>, and if anyone sees a boundary that's not in their DBCs but is in that global set then they scream and die)

it would still add extra computational complexity, and would be better to avoid if possible.

Note that I am leaving the original commit (followed by a revert) in the history since we have already used it internally as a sort of  "submodule hash" and it would be good to keep that in the history.

roystgnr (approved):
Go ahead and merge if you don't want to wait on the downstream test fixes; there's no way *this* PR should trigger any real errors.

jwpeterson:
OK, thanks for reviewing. I'll wait for the remaining in progress tests to finish but not worry about the failed ones.

PR #3462, opened 2022-12-21
Add named block, sideset, and nodeset support for Nemesis meshes

jasondhales:
I have a small test problem in MOOSE that requires these changes.


roystgnr:
Just got around to looking at https://github.com/idaholab/moose/pull/23007 ; that'll do.

PR #3463, opened 2023-01-04
We no longer use hashword.h inside hashing.h

roystgnr:
And libMesh still builds fine either way, so nothing in the library is relying on indirect inclusion either.  If MOOSE and apps are happy too this should be an easy merge.

### DIFF
--- a/include/utils/hashing.h
+++ b/include/utils/hashing.h
@@ -18,8 +18,6 @@
 #ifndef LIBMESH_HASHING_H
 #define LIBMESH_HASHING_H
 
-#include "libmesh/hashword.h"
-
 #include <functional>
 
 namespace libMesh


jwpeterson:
The failures here don't seem related to this simple change... fine with me if you just want to go ahead and merge?

jwpeterson (approved):

roystgnr:
I'm planning to wait, in case there are any subsequent failures in recipe steps that we haven't reached yet, from some app or other that was relying on indirect inclusion.  @lindsayad is going to need this in his PR but he can cherry-pick for now.

I've pointed out these failures downstream ( https://github.com/idaholab/moose/issues/22711 ) so hopefully we'll be back in shape sooner or later.

lindsayad:
Thanks for this!

PR #3467, opened 2023-01-18
Submodule updates

roystgnr:
The new TIMPI features are cool, but the MetaPhysicL and TIMPI bugfixes are the important things.


roystgnr:
Well, that's a relief: I can at least replicate this locally.

roystgnr:
It's looking like the bug I fixed was *hiding* another bug, another one of those annoying C++-isms where one level of a an overload with a nested type gets instantiated before all the necessary partial specializations are declared.  This is going to be fun to fix...

roystgnr:
Turned out to be fun to fix, just not fun to diagnose.  I'll update submodules here again to kick the CI after the MOOSE level of the fix makes it to next.

roystgnr:
Damn, same failure.  Well, I'll try the MetaPhysicL update that fixed it locally for me.

roystgnr:
@jwpeterson - if you need the `string` compatibility ASAP and don't want to wait for this CI thing to get straightened out, we could try updating to just after https://github.com/libMesh/TIMPI/pull/100 instead; it was the subsequent bugfix PR that's leading to the regressions here.

jwpeterson:
OK, I am compiling libmesh now and I will let you know if the latest TIMPI makes my code compile. If so, I agree it might be worth an intermediate update.

roystgnr:
Whoa - did https://github.com/idaholab/moose/pull/23229 actually fix the incompatibility here?

If this really passes then I'm getting a libMesh submodule going ASAP after merging, before something regresses us again.  In the long run I fear we might have to do something special in MOOSE unity builds - include `packing_decl.h` and the MetaPhysicL packing specializations at the top before any `.C` files are included, maybe.

lindsayad:
Nice!

PR #3476, opened 2023-02-01
Better messages when libmesh_parallel_only() fails

roystgnr:
This category of warnings is in recent years being hit by a large collection of programmers who are writing distributed-parallel codes for the first time and could benefit from clearer messages when they get out of sync, as opposed to the original collection of programmers who the category was designed for, {me}.

We'll get the equivalent updates to the TIMPI version of these when PR 109 there makes it downstream.


loganharbour:
🥳

PR #3470, opened 2023-01-25
Allow boundary be associated with child elements 

dewenyushu:
Taking over #3094

PR #3477, opened 2023-02-03
Allow children to have boundary conditions

grmnptr:
Taking over #3470 which took over #3094

### DIFF
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -557,6 +557,7 @@ public:
    */
   void build_shellface_boundary_ids(std::vector<boundary_id_type> & b_ids) const;
 
+#ifdef LIBMESH_ENABLE_AMR
   /**
   * Update parent's boundary id list so that this information is consistent with
   * its children
@@ -567,9 +568,9 @@ public:
   * Specifically, when we coarsen an element whose children have different boundary ids.
   * In such scenarios, the parent will inherit the children's boundaries if at
   * least 50% them own a boundary while sharing the side of the parent.
-  * Otherwise, we delete the boundary from the children and the parent as well.
   */
   void transfer_boundary_ids_from_children(const Elem * const parent);
+#endif
 
   /**
    * \returns The number of element-side-based boundary conditions.
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -877,6 +877,18 @@ public:
   const std::multimap<const Elem *, std::pair<unsigned short int, boundary_id_type>> & get_sideset_map() const
   { return _boundary_side_id; }
 
+  /**
+   * \returns Whether or not there are some children on boundary sides
+   */
+  bool is_children_on_boundary_side() const
+  { return _children_on_boundary; }
+
+  /**
+   * Whether or not to allow set boundary sides on children elements
+   */
+  void allow_children_on_boundary_side(const bool children_on_boundary)
+  { _children_on_boundary = children_on_boundary; }
+
 private:
 
   /**
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -892,13 +892,13 @@ public:
   { return _boundary_side_id; }
 
   /**
-   * \returns Whether or not there are some children on boundary sides
+   * \returns Whether or not there may be child elements directly assigned boundary sides
    */
   bool is_children_on_boundary_side() const
   { return _children_on_boundary; }
 
   /**
-   * Whether or not to allow set boundary sides on children elements
+   * Whether or not to allow directly setting boundary sides on child elements
    */
   void allow_children_on_boundary_side(const bool children_on_boundary)
   { _children_on_boundary = children_on_boundary; }
@@ -955,7 +955,7 @@ private:
 
   /*
    * Whether or not children elements are associated to any boundary
-   * It is false by default. The flag will be turnned on if add_side
+   * It is false by default. The flag will be turned on if add_side
    * function is called with a child element
    */
   bool _children_on_boundary;
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -481,6 +481,13 @@ public:
                                const unsigned short int side) const;
 
   /**
+   * \returns The number of raw (excludes ancestors) boundary ids associated with the \p side
+   * side of element \p elem.
+   */
+  unsigned int n_raw_boundary_ids (const Elem * const elem,
+                                   const unsigned short int side) const;
+
+  /**
    * \returns The list of boundary ids associated with the \p side side of
    * element \p elem.
    */
@@ -552,15 +559,15 @@ public:
 
   /**
   * Update parent's boundary id list so that this information is consistent with
-  * its children elements
+  * its children
   *
   * This is useful when `_children_on_boundary = true`, and is used when the
   * element is about to get coarsened i.e., in MeshRefinement::_coarsen_elements()
   *
-  * Specifically, when coarsen a child element who is the last child on that
-  * boundary, we remove that boundary on the parent's side accordingly.
-  *
-  * Otherwise, we add the parent's side to the boundary.
+  * Specifically, when we coarsen an element whose children have different boundary ids.
+  * In such scenarios, the parent will inherit the children's boundaries if at
+  * least two of them own a boundary while sharing the side of the parent.
+  * Otherwise, we delete the boundary from the children and the parent as well.
   */
   void transfer_boundary_ids_to_parent(const Elem * const elem);
 
@@ -954,8 +961,8 @@ private:
   _boundary_side_id;
 
   /*
-   * Whether or not children elements are associated to any boundary
-   * It is false by default. The flag will be turned on if add_side
+   * Whether or not children elements are associated with any boundary
+   * It is false by default. The flag will be turned on if `add_side`
    * function is called with a child element
    */
   bool _children_on_boundary;
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -927,6 +927,13 @@ private:
                 std::pair<unsigned short int, boundary_id_type>>
   _boundary_side_id;
 
+  /*
+   * Whether or not children elements are associated to any boundary
+   * It is false by default. The flag will be turnned on if add_side
+   * function is called with a child element
+   */
+  bool _children_on_boundary;
+
   /**
    * A collection of user-specified boundary ids for sides, edges, nodes,
    * and shell faces.
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -551,6 +551,20 @@ public:
   void build_shellface_boundary_ids(std::vector<boundary_id_type> & b_ids) const;
 
   /**
+  * Update parent's boundary id list so that this information is consistent with
+  * its children elements
+  *
+  * This is useful when `_children_on_boundary = true`, and is used when the
+  * element is about to get coarsened i.e., in MeshRefinement::_coarsen_elements()
+  *
+  * Specifically, when coarsen a child element who is the last child on that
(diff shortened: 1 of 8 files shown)


src/mesh/boundary_info.C
```diff
@@ -1240,25 +1247,66 @@ void BoundaryInfo::boundary_ids (const Elem * const elem,
   // Clear out any previous contents
   vec_to_fill.clear();
 
-  // Only level-0 elements store BCs.  If this is not a level-0
-  // element get its level-0 parent and infer the BCs.
+  // In most of the cases only level-0 elements store BCs.
+  // In certain application (such as time-dependent domains), however, children
+  // need to store BSc too. This case is covered with the _children_on_boundary
```
roystgnr, line 1252:
```suggestion
  // need to store BCs too. This case is covered with the _children_on_boundary
```

src/mesh/boundary_info.C
```diff
@@ -1240,25 +1247,66 @@ void BoundaryInfo::boundary_ids (const Elem * const elem,
   // Clear out any previous contents
   vec_to_fill.clear();
 
-  // Only level-0 elements store BCs.  If this is not a level-0
-  // element get its level-0 parent and infer the BCs.
+  // In most of the cases only level-0 elements store BCs.
```
roystgnr, line 1250:
```suggestion
  // In most cases only level-0 elements store BCs.
```

src/mesh/boundary_info.C
```diff
@@ -1240,25 +1247,66 @@ void BoundaryInfo::boundary_ids (const Elem * const elem,
...
+      // Loop over ancestors to check if they have boundary ids on the same side
+      while (searched_elem->parent() != nullptr)
+      {
+        const Elem * parent = searched_elem->parent();
+        if (parent->is_child_on_side(parent->which_child_am_i(searched_elem), side) == false)
+          return;
+
+        searched_elem = parent;
+
+        if (_children_on_boundary)
```
roystgnr, line 1280: This is redundant with the if test above.  The redundant for block above could be eliminated, too, by doing `while (searched_elem)` so you get both the element and its ancestors with the same entry.

src/mesh/boundary_info.C
```diff
@@ -1653,23 +1756,61 @@ BoundaryInfo::sides_with_boundary_id(const Elem * const elem,
...
         }
+        // Otherwise we trust what we got and return the side
+        else
+          returnval.push_back(side);
+      }
+  }
+
+#ifdef LIBMESH_ENABLE_AMR
+  // We might have instances (especially with moving boundary domains) when we
+  // query the paren't boundary ID on a child.
```
roystgnr, line 1778:
```suggestion
  // query the parent boundary ID on a child.
```

include/mesh/boundary_info.h
```diff
@@ -550,6 +557,20 @@ class BoundaryInfo : public ParallelObject
...
+  *
+  * This is useful when `_children_on_boundary = true`, and is used when the
+  * element is about to get coarsened i.e., in MeshRefinement::_coarsen_elements()
+  *
+  * Specifically, when we coarsen an element whose children have different boundary ids.
+  * In such scenarios, the parent will inherit the children's boundaries if at
+  * least two of them own a boundary while sharing the side of the parent.
+  * Otherwise, we delete the boundary from the children and the parent as well.
+  */
+  void transfer_boundary_ids_to_parent(const Elem * const elem);
```
roystgnr, line 572: This is somewhat vague and confusing - I can't tell from the comment whether `elem` is expected to be the parent or each child, from the function name I'd expect the child, from the function implementation I'd expect the parent (except that non-parents are filtered out rather than being errors?).

roystgnr: Should we change the name to `transfer_boundary_ids_from_children`, change the parameter name to `parent`?

include/mesh/boundary_info.h
```diff
@@ -550,6 +557,20 @@ class BoundaryInfo : public ParallelObject
...
+  /**
+  * Update parent's boundary id list so that this information is consistent with
+  * its children
+  *
+  * This is useful when `_children_on_boundary = true`, and is used when the
+  * element is about to get coarsened i.e., in MeshRefinement::_coarsen_elements()
+  *
+  * Specifically, when we coarsen an element whose children have different boundary ids.
+  * In such scenarios, the parent will inherit the children's boundaries if at
+  * least two of them own a boundary while sharing the side of the parent.
```
roystgnr, line 569: Surely "two" should be dimension-dependent?  If we're in 1D, for example, and there's only 1 child on a side to begin with, then we still definitely want to copy that id back to the parent!  In 2D "2" is reasonable but then the equivalent criterion in 3D would be "4"; alternatively "2" would also be reasonable in 3D but in that case the equivalent criterion in 2D would be "1".

src/mesh/boundary_info.C
```diff
@@ -1712,6 +1853,61 @@ BoundaryInfo::build_shellface_boundary_ids(std::vector<boundary_id_type> & b_ids
...
+        for (const auto & pr : as_range(_boundary_side_id.equal_range(elem->child_ptr(child_i))))
+        {
+          // Making sure we are on the same boundary
+          if (pr.second.first == side_i)
+          {
+            // If this is the first time we hit this boundary tag, we create an entry for it in the map
+            if (!boundary_counts.count(pr.second.second))
+                boundary_counts.insert({pr.second.second, 0});
+            // Otherwise, just increment the count
+            boundary_counts[pr.second.second] += 1;
```
roystgnr, line 1890: Aren't these four lines just exactly what `++boundary_counts[pr.second.second]` would do?

roystgnr (changes requested):

roystgnr:
This is getting really close.  We should probably turn on the distributed dbg checks right before we merge,rather than waiting for the devel->master merge, just to be safe.  Modifying `Packing<Elem>` is not easy, and just because I don't see any major mistakes in the new version here doesn't mean there aren't any.

The one thing I'm still worried about there: we're not doing any communication after transferring BCs from children to parents, and strictly speaking we should be, because otherwise we might fall out of sync on newly coarsened ghost elements getting BCs transferred from newly deleted remote children.  That alone wouldn't be a disaster for most apps, since we generally only query BCs on local elements ... but follow that up with a repartitioning (as we generally do after AMR/C) that turns an out-of-sync ghost element into a local element ... the element's original owner will send it to the new owner, but IIRC we generally *ignore* incoming communications of that sort (except to assert consistency in dbg mode) for data we already expect to have.

include/mesh/boundary_info.h
grmnptr: Great idea!

include/mesh/boundary_info.h
grmnptr: Oh yes, this came from my 2D mindset. I'll do "if 50%+" of the children on that side have it.

grmnptr:
Got it, thanks for the review! I'll push the changes later today.

src/mesh/boundary_info.C
grmnptr: cool suggestion, thanks!

src/mesh/boundary_info.C
```diff
@@ -1712,6 +1848,60 @@ BoundaryInfo::build_shellface_boundary_ids(std::vector<boundary_id_type> & b_ids
...
+    }
+
+    // This is where the decision is made. If 50% of the children have the tags,
+    // we propagate them upwards upon coarsening. Otherwise, they are deleted.
+    for (const auto & boundary : boundary_counts)
+    {
+      if (boundary.second / number_of_sides_on_children > 0.5)
+        this->add_side(parent, side_i, boundary.first);
+      else
+        this->remove_side(parent, side_i, boundary.first);
```
roystgnr, line 1897: We should never be doing a `remove_side` here, should we?  If the parent doesn't have the side then it's a no-op; if the parent *does* have the side then by definition so does all its children, whether or not the side is set on them *directly*.

src/mesh/boundary_info.C
```diff
@@ -1712,6 +1848,60 @@ BoundaryInfo::build_shellface_boundary_ids(std::vector<boundary_id_type> & b_ids
...
+    }
+  }
+
+  for (const auto & child_i : make_range(parent->n_children()))
+    this->remove(parent->child_ptr(child_i));
```
roystgnr, line 1902: Isn't this redundant with the `remove()` calls in `MeshRefinement::_coarsen_elements`?

roystgnr (changes requested):
Only two remaining issues I could find.  I wouldn't mind about a trillion more test cases, but this looks to be good enough as a starting point, assuming CI agrees with me later that those 4 lines are redundant.  Pinging @jwpeterson in case he wants to take a look before we merge.

src/mesh/boundary_info.C
grmnptr: I deleted this so coarsening won't delete anything from boundary info anymore.

src/mesh/boundary_info.C
grmnptr: I ended up keeping this and adding a condition for the removal in the coarsen loop. The reason behind this is that we might accidentally remove boundary ids before we transfer them back to the parent. Thus, when we allow boundaries on children we remove them after the transfer process.

src/mesh/boundary_info.C
```diff
@@ -1438,9 +1490,6 @@ void BoundaryInfo::remove_side (const Elem * elem,
 {
   libmesh_assert(elem);
 
-  // Only level 0 elements are stored in BoundaryInfo.
-  libmesh_assert_equal_to (elem->level(), 0);
-
   // Erase (elem, side, *) entries from map.
   erase_if(_boundary_side_id, elem,
            [side](decltype(_boundary_side_id)::mapped_type & pr)
```
roystgnr, line 1495: Okay, I think my only remaining concern is what came to me on Slack yesterday:

> We actually probably ought to throw an error for that remove_side() case, if we're not going to support it yet, the more I think about it.  If a child has id 5 because its parent does, and we try to remove it from the child, then ideally we remove it from the parent and restore it on the other children, less ideally we exit with an error and make the app handle it, but what we do now is silently do nothing, and that's totally wrong.

If someone tries to remove a side that isn't there, our current do-nothing behavior is fine, it's like putting white paint on an already-white wall.  But if someone tries to remove a side that is there via an ancestor, but *isn't a raw side*, our current do-nothing behavior isn't fine.  Put white paint on a blue wall and you don't expect to see blue afterwards, not even if the blue was put on with a large brush and the white is being put on with a smaller brush.

This is a new feature, so it doesn't need to be a *comprehensive* new feature, but we definitely don't want to set up applications for failure if the author mistakenly thinks it's comprehensive or if the author accidentally starts to assume that the missing parts of the feature are missing by design.

Could we modify erase_if to return a std::size_t counting the number of erasures, then (here and in the bcid-specific case below) if we get a 0 from that we'll double-check that `boundary_ids(elem, side, vec)` is empty (or below, that `has_boundary_id(elem, side, id)` is false), then throw `libmesh_not_implemented()` if so?

roystgnr (changes requested):

src/mesh/boundary_info.C
```diff
@@ -1504,6 +1533,25 @@ void BoundaryInfo::remove_side (const Elem * elem,
 {
   libmesh_assert(elem);
 
+#ifdef LIBMESH_ENABLE_AMR
+  // Here we have to stop and check if somebody tries to remove an ancestor's boundary ID
+  // through a child
+  if (elem->level() && _children_on_boundary)
```
roystgnr, line 1539: Let's not check _children_on_boundary here?  Trying to remove a child side even when no child sides have been added is still a case we want to scream about.

roystgnr (approved):
I have that one more trivial issue, but I don't think it's worth a holdup; this looks ready to merge once CI is done.

src/mesh/boundary_info.C
grmnptr: Removed!

PR #3479, opened 2023-02-08
Don't let send_list go unprepared

roystgnr:
This was failing when using PETSc DM on newer PETSc, and it was probably a performance penalty regardless.


roystgnr:
Merging; this fixes parallel `make check -C examples/fem_system/fem_system_ex1` for me, and the parts unrelated to that ought to be optimizations via calls to well-tested code.

PR #3481, opened 2023-02-08
No more alltoall() calls

roystgnr:
I think these are the last communication steps in libMesh that weren't using the NBX algorithm but should have been.

There's still work to be done at the TIMPI level, to enable a little more communication/computation interleaving, but this code is at least a lot shorter and should scale a lot better.


roystgnr:
Adding the expensive DistributedMesh tests, just to be extra cautious before we merge.

lindsayad (approved):
Along with an optimization this is a nice line reduction!

jwpeterson:
> Along with an optimization this is a nice line reduction!

alltonone

roystgnr:
alltosome, but not sometoall

PR #3484, opened 2023-02-14
Add System::has_constraint_object()

jwpeterson:
There was previously no way to safely determine whether a user-defined Constraint object was attached to the System.


jwpeterson:
There are no unit tests of either `System::get_constraint_object()` or this new function, but we will be using it in our internal tests a lot.

PR #3485, opened 2023-02-21
Fix for --disable-amr builds

roystgnr:
Took me too long to realize that I had to make sure to --enable-deprecated to reproduce the build failure here...


jwpeterson (approved):
Looks good to me, assuming the "No AMR" test passes.

PR #3491, opened 2023-03-06
ExodusII_IO: Allow low-order edge blocks to refer to higher-order Elem edges

jwpeterson:
We'd like to be able to define lower-order edge (EDGE2) blocks in exo files and have them "match" with edges of higher-order Elems (e.g. HEX20) provided that the vertices match. Previously, the code would only detect a match with a HEX20 edge if the edge block contained EDGE3 elements. libMesh's primary use of edge blocks in exo files is for defining edgesets in the `BoundaryInfo` object, since there is no direct equivalent of the exo sideset for edges. That existing capability will be slightly expanded (while remaining backwards compatible) by this update.

It would be possible, but a bit awkward, to add a unit test for this, since the test would require an externally generated exo file that has lower-order edge blocks in it... currently I don't think there is a way to do this with pure libmesh code (we can only add full-ordered edges of 3D Elems that actually exist in the mesh). I could add such a test if desired, but I doubt this will be a widely used functionality. This capability will get tested by our internal tests.


roystgnr:
> the test would require an externally generated exo file that has lower-order edge blocks in it

We do have the tests/meshes/ directory for things like that.

jwpeterson:
> We do have the tests/meshes/ directory for things like that.

OK, so if you are interested in having a test case, I can check whether the one I'm using now would be alright to share.

jwpeterson:
I think the failures are due to the fact that the exo file I'm adding is "netCDF-4" type, so it only works with HDF5-enabled builds.

PR #3494, opened 2023-03-09
Moving _values_manually_retrieved to after _get_array in PetscVector

### LINKED ISSUE #3493: PetscVector not allowing raw array access after read-only operations
The following code issues an the subsequent error:
```C++
  // Just in case
  petsc_solution->restore_array();

  // Just print the vector using ::operator()
  for (numeric_index_type i = petsc_solution->first_local_index(); i < petsc_solution->last_local_index(); ++i)
    std::cout << "u(" << i << ") = " << (*petsc_solution)(i) << std::endl;

  // Get array
  PetscScalar * petsc_array = petsc_solution->get_array();

  // restore
  petsc_solution->restore_array();
  petsc_solution->close();
```

```
libMesh terminating:
PetscVector values were manually retrieved but are being automatically restored!
Stack frames: 18
1: libMesh::MacroFunctions::report_error(char const*, int, char const*, char const*, std::ostream&)
2: libMesh::PetscVector<double>::_restore_array() const
3: libMesh::PetscVector<double>::_get_array(bool) const
```

This operation should be allowed `::operator()` should not restricting access to the raw array.

`get_array()` +should+ error out if:
- There is another request that hasn't been restored yet
- Other operations like `operator()` or `set` are being used before calling `restore_array()`

`get_array_read()` should error if:
- There is a `get_array()` that hasn't been restored yet
- Non-`const` operations are being performed before it's been restored.

zachmprince:
Closes #3493

### DIFF
--- a/include/numerics/petsc_vector.h
+++ b/include/numerics/petsc_vector.h
@@ -1110,8 +1110,8 @@ template <typename T>
 inline
 PetscScalar * PetscVector<T>::get_array()
 {
-  _values_manually_retrieved = true;
   _get_array(false);
+  _values_manually_retrieved = true;
 
   return _values;
 }
@@ -1121,8 +1121,8 @@ template <typename T>
 inline
 const PetscScalar * PetscVector<T>::get_array_read() const
 {
-  _values_manually_retrieved = true;
   _get_array(true);
+  _values_manually_retrieved = true;
 
   return _read_only_values;
 }
--- a/tests/numerics/petsc_vector_test.C
+++ b/tests/numerics/petsc_vector_test.C
@@ -77,6 +77,21 @@ public:
     CPPUNIT_ASSERT_EQUAL((intptr_t)read_only_values, (intptr_t)values);
 
     v.restore_array();
+
+    // Test to make sure we can get arrays after other operators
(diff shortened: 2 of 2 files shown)


jwpeterson:
I see that you updated the unit test, but CIVET reports no coverage of the new lines of code. It's possible that's an error with the coverage itself, but probably good to confirm that we know exactly what's going on before merging...

YaqiWang:
@loganharbour or @roystgnr any comments about @jwpeterson 's comment? I checked the full coverage report, which does not make sense on this change.

loganharbour:
It's not hit in the current coverage either: https://mooseframework.inl.gov/libmesh/docs/coverage/include/numerics/petsc_vector.h.gcov.html

I would not be surprised here if we are missing coverage because we're running with -O2

roystgnr (approved):

jwpeterson:
OK, so this patch fixes a correctness issue that is only relevant in debug-mode (or less aggressively optimized) binaries. That sounds good to me, but I think it's more interesting that the compiler can actually make this optimization... setting a class member boolean seems like an "observable side effect" that would not be considered a candidate for elision.

jwpeterson (approved):

PR #3495, opened 2023-03-09
Add toggle for HDF5 for nemesis output

loganharbour:
FYI @MengnanLi91

### DIFF
--- a/include/mesh/nemesis_io.h
+++ b/include/mesh/nemesis_io.h
@@ -201,6 +201,13 @@ public:
                             unsigned int timestep,
                             std::vector<Real> & global_values);
 
+  /**
+   * Set to true (the default) to write files in an HDF5-based file
+   * format (when HDF5 is available), or to false to write files in
+   * the old NetCDF3-based format.  If HDF5 is unavailable, this
+   * setting does nothing.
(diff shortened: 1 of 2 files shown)


roystgnr (approved):

loganharbour:
Can I get a merge? I think you're used to approving and leaving be for people that have write :-)

roystgnr:
Just kicked CI myself, and already it's saying Coverage failed.  Shouldn't it be pending until the gcov recipes run before it decides that?

loganharbour:
Even better, you invalidated the wrong event:

https://civet.inl.gov/pr/17922/

If you look, the most recent event is all green. The changes just didn't push to GH.

loganharbour:
> Just kicked CI myself, and already it's saying Coverage failed. Shouldn't it be pending until the gcov recipes run before it decides that?

This is because you invalidated old recipes. Coverage will be pending until its dependencies are met. When you invalidate everything, like you did, it sends a fail message for all events. Then, as they run they repopulate. But because coverage doesn't actually run until it's ready... it stays as failed until it runs.

jwpeterson (approved):

PR #3497, opened 2023-03-15
Add Tolerance to Floating Point Comparisons in find_intersection()

### LINKED ISSUE #3496: Floating-Point Comparisons without Tolerance in `mesh_triangle_holes.C`
The `find_intersection()` function in `mesh_triangle_holes.C` contains a number of floating-point comparisons without tolerance. The function is used by MOOSE's `XYDelaunayGenerator` through `libMesh::TriangulatorInterface::Hole::find_ray_intersections()` to verify the holes/boundary geometry. And these floating-point comparisons without tolerance sometimes cause "false positive error" in such verification.

miaoyinb:
closes #3496

### DIFF
--- a/src/mesh/mesh_triangle_holes.C
+++ b/src/mesh/mesh_triangle_holes.C
@@ -70,59 +70,95 @@ namespace
     return signof(det);
   }
 
+  // Return true iff the ray from source toward ray_target crosses the
+  // edge from edge_pt0 (non-inclusive) to edge_pt1 (inclusive) in the
+  // forward ray direction.  edge_pt2 is the vertex following edge_pt1
+  // along the polygon; it is used to classify a crossing that lands
+  // exactly on the shared vertex edge_pt1.
+  //
+  // The decision is made purely from orientation predicates on the
+  // vertices, never from a parametric edge coordinate compared against
+  // a tolerance.  This is what guarantees consistency: a shared vertex
+  // yields the exact same orientation when viewed from either of its
+  // two edges, so a ray passing through a vertex is counted exactly
+  // once for a "through" crossing and not at all for a "tangent"
+  // (glancing) touch, regardless of floating-point round-off.
   bool is_intersection(const Point & source,
                        const Point & ray_target,
                        const Point & edge_pt0,
-                       const Point & edge_pt1)
+                       const Point & edge_pt1,
+                       const Point & edge_pt2)
   {
-    int orient_st0 = orientation(source, ray_target, edge_pt0);
-    int orient_st1 = orientation(source, ray_target, edge_pt1);
-    int orient_edge_s = orientation(edge_pt0, edge_pt1, source);
-    int orient_edge_t = ray_orientation(edge_pt0, edge_pt1, source, ray_target);
+    const int orient_st0 = orientation(source, ray_target, edge_pt0);
+    const int orient_st1 = orientation(source, ray_target, edge_pt1);
 
-    // Intersection on interior
-    if ((orient_st0 == -orient_st1) &&
-        (orient_edge_s != orient_edge_t))
-      return true;
+    // Which side of the edge line the source lies on, and which way the
+    // ray turns off the edge, together tell us whether the crossing is
+    // ahead of the source rather than behind it.
+    const int orient_edge_s = orientation(edge_pt0, edge_pt1, source);
+    const int orient_edge_t = ray_orientation(edge_pt0, edge_pt1, source, ray_target);
+    const bool forward = (orient_edge_s != orient_edge_t);
 
-    // Ray intersects edge_pt1
+    // The ray line passes exactly through edge_pt1, the inclusive
+    // endpoint of this edge.  Count it only as a genuine "through"
+    // crossing: the previous vertex (edge_pt0) and the next vertex
+    // (edge_pt2) must lie on opposite sides of the ray line.  A tangent
+    // touch (both neighbors on the same side) or a collinear edge
+    // (edge_pt0 also on the line) does not count.
     if (orient_st1 == 0)
-      return true;
+      {
+        const int orient_st2 = orientation(source, ray_target, edge_pt2);
+        return forward && (orient_st0 != 0) && (orient_st0 == -orient_st2);
+      }
 
-    // Source is on line; we don't count that
-    // if (orient_edge_s == 0)
-    // Ray is parallel to edge; no intersection;
-    // if (orient_edge_t == 0)
-    // Ray intersects edge_pt0; we don't count that
-    // if (orient_st0 == 0)
+    // The ray line passes through edge_pt0, the non-inclusive endpoint;
+    // that vertex is counted (if at all) by the previous edge.
+    //
+    // A run of two or more vertices all lying exactly on the ray line
+    // is therefore never counted, even when the boundary arrives at the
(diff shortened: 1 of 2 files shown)


miaoyinb:
@roystgnr Please take a look to see if the update makes sense. It does solve the issue I am having now with MOOSE but I am not sure it would affect other stuffs. Thanks.

roystgnr:
Definitely makes sense.  Using an absolute tolerance rather than a relative one bugs me (at least in that first test; t is non-dimensionalized already so "absolute" is already relative), but we can scale by the ray or edge length to fix that.

What worries me more is: IIRC, that `(t >= 0 && t < 1)` test was very carefully chosen, with the `t==0` case needing special handling because of the "intersects at exactly one point" case, and turning that from an exact test into a fuzzy test might have broken some of those cases.  It's at least a very good sign that we're still passing all unit tests ... but the very fact that you're running into problems means that there are still corner cases which our unit tests don't cover.

Can you give an example (in the form of a new unit test would be ideal, but if you just have a MOOSE input file with XYDelaunay then I'll work from there) of a case that fails with the current code but works with this?  That would be helpful in figuring out if we can improve the solution here (in particular, I could try to rescale that absolute tolerance, but I have no idea whether the result would still fix your failing case!) or if I can find a better solution.  And a new test case would be indispensable in making sure we never revert the fix.  Imagine if someone later does run into another corner case, and their changes effectively revert this in order to fix it; I'd have no way to realize (except maybe a twinge of _deja vu_ when I say to them "It's at least a very good sign that we're still passing all unit tests") that we were breaking you again.

miaoyinb:
Sure. I am developing a complex mesh generator that uses `XYDelaunay` as sub-generator and met with this issue. Let me see if I can make a simpler example.

miaoyinb:
This is a simple input that has the issue with the current code but can be run without issue after the modifications here.
I did notice that the issue is caused by a hole boundary node at {x1,0,0} and an external boundary node at {x2,0,0}, because a {1,0,0} ray direction is used here.
```
[Mesh]
  [pccmg1]
    type = PolygonConcentricCircleMeshGenerator
    num_sides = 5
    num_sectors_per_side = '4 4 4 4 4'
    background_intervals = 2
    polygon_size = 8.0

    ring_radii = '6.0 7.0'
    ring_intervals = '2 2'
  []
  [bdg1]
    type = BlockDeletionGenerator
    input = pccmg1
    block = 4
    new_boundary = 10000
  []
  [tg1]
    type = TransformGenerator
    input = bdg1
    transform = ROTATE
    vector_value = '90.0 0.0 0.0'
  []  
  [pcg]
    type = ParsedCurveGenerator
    x_formula = 'R*cos(t*2*pi)'
    y_formula = 'R*sin(t*2*pi)'
    section_bounding_t_values = '0 1'
    nums_segments = 60
    constant_names = 'pi R'
    constant_expressions = '${fparse pi} 100'
    is_closed_loop = true
  []
  [xyd]
    type = XYDelaunayGenerator
    boundary = pcg
    holes = tg1
    refine_boundary = false
    stitch_holes = 'true'
    desired_area = 10
  []  
[]
```

roystgnr:
That's perfect; thanks.  I can replicate easily with that.

I'm still hoping to come up with a better fix than your patch here, but I'm also still failing to.  If I don't come up with something soon, we can just merge your patch but add a unit test based on your failing case here to make sure future revisions don't break it again.  Mind if I push to your branch?

miaoyinb:
Sounds good. Please feel free to push to my branch.

miaoyinb:
@roystgnr I think your concerns make sense. I did meet with a false error even with this modification. I have not yet been able to replicate the new error using a simple input. But I think it should be related to the situation you mentioned.

I am thinking if we should use a random ray direction instead of {1,0,0}; or use two ray directions; or do a prescreening of ray direction to make sure the ray is away from a node. I was just thinking out loud here.

miaoyinb:
The best I can do so far is to provide an input using the existing MOOSE MGs and load the mesh created by my under-development MGs from exodus files here ([accs.zip](https://github.com/libMesh/libmesh/files/11024718/accs.zip)).

The example runs well with the original libMesh but fails the hole verification with the modified libMesh.

```
assm_pitch = 100
assm_side = '${fparse assm_pitch/sqrt(3.0)}'
[Mesh]
  [acc1]
    type = FileMeshGenerator
    file = acc1.e
  []
  [acc2]
    type = FileMeshGenerator
    file = acc2.e
  []
  [hex]
    type = PolyLineMeshGenerator
    points = '${fparse
[accs.zip](https://github.com/libMesh/libmesh/files/11024714/accs.zip)
 assm_side/2.0} ${fparse assm_pitch/2.0} 0.0
                  ${fparse assm_side} 0.0 0.0
                  ${fparse assm_side/2.0} ${fparse -assm_pitch/2.0} 0.0
                  ${fparse -assm_side/2.0} ${fparse -assm_pitch/2.0} 0.0
                  ${fparse -assm_side} 0.0 0.0
                  ${fparse -assm_side/2.0} ${fparse assm_pitch/2.0} 0.0'
    loop = true
    num_edges_between_points = 10
  []
  [xyd]
    type = XYDelaunayGenerator
    boundary = hex
    holes = 'acc1 acc2'
    refine_boundary = false
    stitch_holes = 'true true'
    desired_area = 10
    # verify_holes = 'false'
  []
[]

```

roystgnr:
And it's exactly what I was fearing might be a failure case: our ray from one hole point is "intersecting" (within tolerance) at one vertex of the other hole, but this time it's a "tangent" intersection rather than a "through" intersection, and our "check for double-intersections and don't count them" code isn't working after your patch.

I'll turn it into a unit test and we'll see what we can do.

miaoyinb:
Got some time to re-investigate this. My idea is to add a random, small but non-zero shift to the ray direction. It seems to work for my cases but not sure if it is generally a good idea. (obviously it failed some tests 😢) Any insight would be greatly appreciated.

roystgnr:
I think a random shift is a terrifying idea - it won't actually fix the bug, it'll just make the bug harder to replicate.

My vague intuition here is that we've hit a problem where using *any* tolerances is a bad idea.  The goal isn't to catch corner cases here and there, it's just to treat corner cases consistently.  Whether we think we're hitting a boundary vertex or not doesn't actually matter.  What matters is that, when we look at the same boundary vertex from each of the edges sharing it, we always come to the exact same conclusion in both cases about whether our ray intersects it or goes "above" it or "below" it.

Hmm... maybe that's what I'm doing wrong here?  Instead of trying to find the parameter u for the distance along each of the two edges our intersection is at, then check for u∈[0,1], maybe we should be testing whether a ray is clockwise/counterclockwise/intersecting each of the three vertices of those two edges.  By doing a vertex test instead of an edge test we'd guarantee consistency.

miaoyinb:
@roystgnr this has been a while...tried to revisit it.

miaoyinb:
@roystgnr, just a reminder. Let me know if the new approach would solve the issue. Thanks.

roystgnr:
This actually looks good to me now.  I haven't been able to convince myself that there isn't some subtle corner case we might be missing, but even in the worst case scenario I think we've got an improvement over the previous code.  Thanks for sticking with this!

roystgnr (approved):

PR #3500, opened 2023-03-22
Update TIMPI

roystgnr:
This gets us better error messages when timpi_parallel_only fails, Packing<tuple> fixes, better test coverage, better comments, more rvalue support in the NBX algorithm, and the removal of `Packing` specifications for unpackable container types.

Hopefully this goes better than the current Update-MetaPhysicL PR...


roystgnr:
I've got to remember that "fixes" is a Github keyword, not just a comment...

loganharbour:
Following. I would really like to get this to moose next week if possible

PR #3503, opened 2023-03-28
Fix "make check" with MetaPhysicL without MPI

roystgnr:


roystgnr:
Added the failing test.

PR #3508, opened 2023-03-31
Make our use of isspace more robust

roystgnr:
We had some users of a modified libMesh who saw the existing code break here - their fault for bringing in a 3rd party header with `using namespace std` (thereby making `isspace` ambiguous), but the workaround is easy enough, and I'm not a big fan of us including C headers when there's a C++ version around anyway.

Pinging @rainiscold and @LabrosV ; hopefully I haven't typo'd either.

### DIFF
--- a/include/base/getpot.h
+++ b/include/base/getpot.h
@@ -46,6 +46,7 @@
 #include <cmath>
 #include <cstdlib>
 #include <ctime>
+#include <cctype>
 
 extern "C" {
 #include <stdarg.h> // --> va_list and friends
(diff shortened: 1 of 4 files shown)


jwpeterson:
Any chance you'd want to make the same update in `AbaqusIO` and `GMVIO`?

jwpeterson (approved):

roystgnr:
Huh; definitely.  It wasn't a problem there because there's no std algorithm use and so no ambiguous template argument, but let me get them all (and getpot.h) while I'm at it.

roystgnr:
Wait, GMV is also using it in remove_if.  I guess the polluting header just didn't end up in the dependencies there.

PR #3509, opened 2023-03-31
Potentially prepare meshes for use when stitching

lindsayad:
This is to pass an assertion in `copy_nodes_and_elements` that the meshes have matching states of preparedness. I think these changes are fine from a user perspective. `this` mesh is obviously going to be modified during stitching, and I would wager that the majority of the time a user is going to discard the `other_mesh` anyway since they're merging it into `this`. And if they're not, they can still pass to the `const` overload if they want to ensure it's not modified


lindsayad:
closing in favor of #3510

PR #3510, opened 2023-03-31
Remove overzealous assertion in copy_nodes_and_elements

lindsayad:
This code dates from 2007 when we were basically asserting that other data in the mesh matched. Refs failures on idaholab/moose#23923


roystgnr (approved):

roystgnr:
Kicking the failing recipe, since you've already got it into an issue.

lindsayad:
The link I made won't illustrate the error now, but hopefully I can remember/find the test with the problem

PR #3512, opened 2023-04-05
ExodusII_IO: Allow edges with different orientations to match

jwpeterson:
I previously had a check in the Exodus reader to make sure that the orientation of an Edge specified in an edge block matched the orientation of a candidate Elem's edge before adding it to the `BoundaryInfo` object. What I should have realized (since I added a unit test which demonstrated the issue) is that it is possible for two neighboring elements to have the same orientation for a given edge while the edge block specifies the edge in the opposite orientation. In this case, we were (incorrectly) adding no edges to the `BoundaryInfo` object. I have now updated this check so that Edges specified in edge blocks match either possible orientation.

This means that when adding edgesets via Exodus edge blocks, one gets an (Elem, edge, id) tuple in the `BoundaryInfo` object for *each* neighboring Elem that touches it (regardless of orientation). This is a byproduct of the way that we use Exodus edge blocks, since Exodus edge blocks don't specify the Elem which each edge in the block is meant to belong to, but this actually works well for my personal use case since I need to know about all neighboring Elems anyway.

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -1576,21 +1576,6 @@ void ExodusII_IO_Helper::read_edge_blocks(MeshBase & mesh)
                       bi.add_edge(elem_edge_pair.first,
                                   elem_edge_pair.second,
                                   edge_block_id);
-
-                      // Debugging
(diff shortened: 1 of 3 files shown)


roystgnr (approved):
No need to wait for the MOOSE CI fixes, I think.

PR #3513, opened 2023-04-10
TIMPI update, adding `--timpi_sync=` option

roystgnr:
This gets us a few older algorithms to enable, which might be useful as a last-ditch option whenever we're worried that our current NBX implementation might have some issue or that some particular MPI implementation might be having difficulty supporting it.


roystgnr:
I swear I built this with `--enable-paranoid-warnings --enable-werror` already.  Apparently I didn't do so with an old enough gcc, though?

PR #3514, opened 2023-04-11
Make linear partitioner idempotent for distributed meshes

lindsayad:
The blksize was wrong if `partition_range` was called (again) after the mesh had been distributed. Now I'm getting the same partitioning even if I call `prepare_for_use` twice

This is relevant to work on idaholab/moose#24018


roystgnr (approved):
Good catch!

PR #3519, opened 2023-04-17
Apply constraints via element nonlinear residuals

### LINKED ISSUE #3504: Constraint application slow
In #1986 I changed the way we do constraint application. This has led to very large slow downs for @roystgnr's IGA work and likely slows down every problem for which we have constraints at the libMesh level. The slow down is likely (definitely?) due to inserting new nonzeroes in the PETSc matrix.

roystgnr:
This, combined with updates in app codes, ought to fix #3504.

I'm going to need to retest my corresponding MOOSE branch (it was working before, but I've done a lot of cleanup to the libMesh branch since then) before we merge this, but I'm confident - getting MOOSE problems to work was much easier than getting every combination of adjoints and heterogeneous constraints and NewtonSolver-vs-PetscDiffSolver to work in libMesh examples.

### DIFF
--- a/examples/adjoints/adjoints_ex1/adjoints_ex1.C
+++ b/examples/adjoints/adjoints_ex1/adjoints_ex1.C
@@ -67,6 +67,7 @@
 #include "libmesh/newton_solver.h"
 #include "libmesh/numeric_vector.h"
 #include "libmesh/partitioner.h"
+#include "libmesh/petsc_diff_solver.h"
 #include "libmesh/steady_solver.h"
 #include "libmesh/system_norm.h"
 #include "libmesh/enum_solver_package.h"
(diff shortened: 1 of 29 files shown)


roystgnr:
My moose branch seems to be working - and if I disable global constraint handling while not using element residual constraints, it fails, so I think the test coverage is good.

roystgnr:
No changes in a GRINS run either.

jwpeterson (approved):
I think you can go ahead and merge. As discussed over IM, I don't think this change will affect our internal testing because we don't use the `PetscNonlinearSolver`.

roystgnr:
I'll put in two separate PRs for the braces and the new-s.

lindsayad:
Thanks @roystgnr !

roystgnr:
This was the last thing I really wanted to get in before the next libMesh submodule update ... except that it feels like we're getting really close to figuring out our sawtooth MPI troubles, so waiting to see if we can get a fix/workaround into TIMPI is tempting.

PR #3522, opened 2023-04-18
Update TIMPI submodule

roystgnr:
This gets us the cleanup and the error checking that Logan added - which appears to have turned out to be a red herring, because we were really being beset by a mvapich 2.3.5 bug that they've fixed as of 2.3.7 - but still, code improvements.

@loganharbour


loganharbour (approved):
🥳

PR #3527, opened 2023-04-20
Remove setting of skip_noncritical_partitioning in copy_nodes_and_elements

lindsayad:
Pull #3301 abstracted so that we could use MeshBase objects in `copy_nodes_and_elements`. However, this had the unfortunate effect of conceptually changing the effect that calling the `skip_noncritical_partitioning` setter had. The getter for `skip_noncritical_partitioning` checks three booleans. The setter only sets one. This has consequences for `operator==` comparisons for meshes. Since are not modifying the value of `_skip_noncritical_partitioning` before the `prepare_for_use()` here I really see no need for the setter I'm removing. That doesn't seem like a job for `copy_nodes_and_elements`

### DIFF
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -503,7 +503,6 @@ void UnstructuredMesh::copy_nodes_and_elements(const MeshBase & other_mesh,
   this->allow_renumbering(other_mesh.allow_renumbering());
   this->allow_remote_element_removal(other_mesh.allow_remote_element_removal());
   this->skip_partitioning(other_mesh.skip_partitioning());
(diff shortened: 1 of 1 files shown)


roystgnr (approved):
This looks like a real problem, but I don't think this is a complete fix.  We should be copying these sorts of settings properly in `MeshBase` (whereas I'm seeing us just setting it to `false` in the copy constructor and making the same bug in the move-assignment operator), and we ought to be able to get rid of *all* the copying of settings here - it's `copy_nodes_and_elements()`, not `copy_nodes_and_elements_and_some_trivia()`, right?

That said, this is the first step in the right direction.

jwpeterson:
So… `dont_skip_non_critical_partitioning`??

PR #3529, opened 2023-04-24
Avoid using direct `new` calls in examples

jwpeterson:
As mentioned on #3519, we still had some instances of pre-C++14 std::unique_ptr use in the examples; this branch cleans up all the ones I could find.


roystgnr (approved):
I was going to get to this, I swear...

PR #3532, opened 2023-04-25
RBConstruction: Re-throw exceptions on all procs in parallel

jwpeterson:
We already had similar code around the parallel elem assembly loop; the node assembly loop is also parallel so it needs the same exception synchronization/re-throwing to avoid deadlocks in parallel.

### DIFF
--- a/src/reduced_basis/rb_construction.C
+++ b/src/reduced_basis/rb_construction.C
@@ -629,6 +629,17 @@ void RBConstruction::add_scaled_matrix_and_vector(Number scalar,
   for (const auto & t : mesh.get_boundary_info().build_node_list())
     nodes_with_nodesets.insert(std::get<0>(t));
 
+  // It's possible for the node assembly loop below to throw an
+  // exception on one (or some subset) of the processors. In that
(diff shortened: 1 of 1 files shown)


src/reduced_basis/rb_construction.C
```diff
@@ -682,6 +693,18 @@ void RBConstruction::add_scaled_matrix_and_vector(Number scalar,
             }
         }
     }
+  } // libmesh_try
+  libmesh_catch(...)
```
roystgnr, line 697: Would it make sense to have a preceding catch block for `std::exception & e`, so you can preserve `e.what()`?  We don't have any great examples of that in libMesh (maybe one in rb_data_serialization.C) but it's standard in MOOSE and seems to be really helpful for diagnosis.

jwpeterson: I think that's a potential improvement, but we'd have to communicate the `e.what()` string(s) from all processors and combine or uniquify them in some way for it to be sensible to the user. And I would not actually be able to test this easily, since our own internal exception type currently does not derive from std::exception.

Issue #3533, opened 2023-04-25
Allow existing SECOND-order elements in `UnstructuredMesh::all_second_order_range`


jwpeterson:
We currently throw an error in `UnstructuredMesh::all_second_order_range` when encountering an Elem that is not `FIRST` order: 

https://github.com/libMesh/libmesh/blob/098b3da530ae977b2613af14f991c39c9a96de61/src/mesh/unstructured_mesh.C#L1415-L1417

It seems to me that, if the Elem is already `SECOND`-order, we could theoretically just copy it over to the new Mesh of SECOND-order elements as-is, however I don't think this will be as simple as just deleting the error message... so I will look into this further.

PR #3534, opened 2023-04-25
Skip any elements in the range that are already second-order

jwpeterson:
>  however I don't think this will be as simple as just deleting the error message

OK, so I did not delete the error message exactly, it's still there for Elems that are neither FIRST nor SECOND-order, but simply skipping elements that are already SECOND order worked for my test case. I'll see about creating a unit test for this, but in the meantime wanted to get the proposed fix out there.

Refs #3533

### DIFF
--- a/tests/mesh/all_second_order.C
+++ b/tests/mesh/all_second_order.C
@@ -3,7 +3,8 @@
 #include <libmesh/distributed_mesh.h>
 #include <libmesh/replicated_mesh.h>
 #include <libmesh/mesh_generation.h>
-#include <libmesh/edge_edge2.h>
+#include <libmesh/point.h>
+#include <libmesh/elem.h>
 
 // cppunit includes
 #include "test_comm.h"
@@ -19,6 +20,7 @@ public:
   CPPUNIT_TEST( allSecondOrder );
   CPPUNIT_TEST( allSecondOrderRange );
   CPPUNIT_TEST( allSecondOrderDoNothing );
+  CPPUNIT_TEST( allSecondOrderMixed );
 
   CPPUNIT_TEST_SUITE_END();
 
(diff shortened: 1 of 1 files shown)


src/mesh/unstructured_mesh.C
```diff
@@ -1412,7 +1412,16 @@ UnstructuredMesh::all_second_order_range (const SimpleRange<element_iterator> &
...
   for (auto & lo_elem : range)
     {
-      // make sure it is linear order
+      // Skip elements in the range that are already SECOND-order.
+      // Ordinarily one does not have a mixture of FIRST and
+      // SECOND-order elements in an all-2D or all-3D mesh, since
+      // FIRST and SECOND-order elements can't be (conforming)
+      // neighbors, but in a mixed-dimension element mesh, this is
+      // possible.
+      if (lo_elem->default_order() == SECOND)
```
roystgnr, line 1421: Should this be `!= FIRST`, since we may have higher-order?

roystgnr (changes requested):
That's the only issue I see.

src/mesh/unstructured_mesh.C
jwpeterson: The `!= FIRST` case is still handled by the subsequent  `libmesh_error_msg()`. In other words, we only want to `continue` for `SECOND`-order elements, `THIRD`, `FOURTH`, etc. we'll still throw an error.

roystgnr (approved):
Makes sense; I guess the average user here might be surprised to see a *down* conversion or might be surprised to see *no* conversion; there's not one obvious semantics for a third-order element.

roystgnr:
This should be enough to close #3533?

PR #3542, opened 2023-05-02
Add and use copy_constraint_rows

lindsayad:
With constraint comparison occurring when we call `MeshTools::valid_is_prepared`, we need to ensure that we copy constraint rows when copy constructing meshes


roystgnr (approved):
Good catch.

PR #3543, opened 2023-05-03
Updates to RBParameters interface

jwpeterson:
We are working on generalizing the `RBParameters` class in a mostly backwards-compatible way so that it stores a `map<string, vector<Real>>` instead of the current `map<string, Real>`. The prior "scalar" use cases will continue to be supported by using single-entry vectors in the map, but the point is that our various `vectorized_evaluate` routines should become much more memory efficient if we can switch from a vector-of-maps to a map-containing-vectors approach.

This change doesn't affect much libMesh code, but we previously used the deprecated APIs a lot in our internal codebase. The new APIs in this PR are the result of working around calling the deprecated APIs, so they should hopefully be sufficient for anyone else that was actually using `RBParameters::get_parameters_map()` extensively. Follow-up PRs will deal with the actual update to `vector<Real>`

### DIFF
--- a/include/reduced_basis/rb_parameters.h
+++ b/include/reduced_basis/rb_parameters.h
@@ -86,9 +86,16 @@ public:
   const std::map<std::string, Real> & get_extra_parameters_map() const;
 
   /**
-   * Get the value of the specific parameter.
+   * Get the value of the specified parameter, throwing an error if it
+   * does not exist.
    */
   Real get_value(const std::string & param_name) const;
+
+  /**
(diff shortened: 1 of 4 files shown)


roystgnr:
"There’s a current network incident for accessing public hosts (intermittent failures and timeouts). This includes to [github.com](http://github.com/), [github.inl.gov](http://github.inl.gov/), and ANL’s gitlab host for SAM. Our contact will let me know when these are resolved, until then… fetching may be an annoying nightmare" - Logan

roystgnr (approved):

PR #3546, opened 2023-05-11
FE optimization

roystgnr:
More all_shape* specializations (and more usage of them) enables us to take advantage of tensor products in LAGRANGE HEX evaluations and of the special structure of `RATIONAL_BERNSTEIN`.  Overall I've only shaved a couple percent off of `LIBMESH_BENCHMARK`, but it doesn't hit those cases super hard.  IGA is *much* less slow now.

### DIFF
--- a/include/fe/fe.h
+++ b/include/fe/fe.h
@@ -249,6 +249,22 @@ public:
                            std::vector<OutputShape> & v,
                            const bool add_p_level = true);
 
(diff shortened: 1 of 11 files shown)


roystgnr:
Refs #2935, where I've been stashing benchmark results.

jwpeterson (approved):

PR #3547, opened 2023-05-11
Change unsigned int to size_t in coupling matrix to prevent overflow

### LINKED ISSUE #3541: Overflow with huge number of variables
There are overflow issues in libMesh when the number of variables is huge; specifically, when the number of variables is larger than 65536 whose square is the unsigned integer limit.
One overflow point that I found is in `coupling_matrix.h` when computing `_row_i*_mat.size()`, but there seems to be more.

NamjaeChoi:
This fixes an overflow that happens when the number of variables is larger than 65536. That overflow not only causes a coupling matrix to be set incorrectly but also causes a ghosted `PetscVector` to contain wrong ghost indices. Closes #3541.

### DIFF
--- a/tests/numerics/coupling_matrix_test.C
+++ b/tests/numerics/coupling_matrix_test.C
@@ -169,7 +169,7 @@ private:
         for (unsigned int pj = 0; pj != sizeof(jvals)/isize; ++pj)
           {
             CPPUNIT_ASSERT(ccr_it != ccr.end());
-            CPPUNIT_ASSERT_EQUAL(*ccr_it, jvals[pj]);
+            CPPUNIT_ASSERT_EQUAL((unsigned int) *ccr_it, jvals[pj]);
             ++ccr_it;
           }
 
--- a/include/numerics/coupling_matrix.h
+++ b/include/numerics/coupling_matrix.h
@@ -59,32 +59,32 @@ public:
    * Constructor.
    */
   explicit
-  CouplingMatrix (const unsigned int n=0);
+  CouplingMatrix (const std::size_t n=0);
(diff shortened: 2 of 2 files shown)


roystgnr (approved):
Kind of overkill as a fix, but I have to admit it's a lot cleaner than keeping parameters `unsigned int` and trying to manually upcast them whenever a calculation needs it.

tests/numerics/coupling_matrix_test.C
```diff
@@ -169,7 +169,7 @@ private:
         for (unsigned int pj = 0; pj != sizeof(jvals)/isize; ++pj)
           {
             CPPUNIT_ASSERT(ccr_it != ccr.end());
-            CPPUNIT_ASSERT_EQUAL(*ccr_it, jvals[pj]);
+            CPPUNIT_ASSERT_EQUAL((unsigned int) *ccr_it, jvals[pj]);
```
jwpeterson, line 172: Tangentially related minor point: I don't know why but we seem to have a bunch of C-style casts in the unit tests when a `static_cast` is the intent. So that's an Effective C++ thing that I will make a separate PR for.

jwpeterson (approved):

pbauman:
Maybe too little too late here, but maybe `dof_id_type` is more appropriate here?

roystgnr:
No, IMHO.  That would still fail in the case of 32-bit `dof_id_type` and higher-than-16-bit variable counts.  Which does seem kind of crazy to me, but not *much* more crazy than 64-bit `dof_id_type` with a hundred thousand variables, so here we are.

PR #3548, opened 2023-05-12
Do not throw from ExodusII_IO_Helper::close()

jwpeterson:
Since this function is called from the ExodusII_IO destructor, it should not throw (all destructors should be marked noexcept) because it could be called during stack unwinding.

### DIFF
--- a/include/mesh/exodusII_io_helper.h
+++ b/include/mesh/exodusII_io_helper.h
@@ -274,8 +274,11 @@ public:
 
   /**
    * Closes the \p ExodusII mesh file.
+   *
+   * This function is called from the ExodusII_IO destructor, so it should
+   * not throw an exception.
    */
-  void close();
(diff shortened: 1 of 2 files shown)


jwpeterson:
FYI @roystgnr I considered changing from `EX_CHECK_ERR` to `EX_EXCEPTIONLESS_CHECK_ERR` like we discussed, but the latter still calls `std::terminate` anyway, which, if we are in the middle of stack unwinding, would have basically the same effect as throwing a second exception. So I think the most appropriate thing to do is simply log the failure to close the file and keep going.

roystgnr (approved):

roystgnr:
The MOOSE clang failure is a CI environment issue there; this ought to be fine to merge.

jwpeterson:
Thanks for reviewing!

PR #3549, opened 2023-05-12
Fix for Exodus writer when sideset and shellface set have same id

jwpeterson:
* Store side_boundary_ids in a set. This lets us easily avoid duplicates when there is a set containing both sides and shellfaces

* Same fix for overlapping sideset/shellface ids is needed in both ExodusII_IO_Helper::write_sidesets() and ExodusII_IO_Helper::initialize()

* Rename elem -> elem_lists, side -> side_lists These are slightly more descriptive names, since the maps contain lists of ids and we normally use the variable "elem" to refer to an Elem pointer.


jwpeterson:
Previously, if we read an Exodus file with a sideset containing both 3D elem sides and shellfaces, and then wrote it back out, we would get two copies of the original sideset in the resulting file rather than one. That is, you would see something like:
```
 ss_prop1 = 200, 510, 101, 201, 102, 100, 500, 200, 201 ;
 ```
 for the sideset ids (note 200 and 201 appear twice). Theoretically I guess this could have been OK other than the wasted file space, but we ran into an issue where it was actually producing garbage values like the following:
 ```
 elem_ss8 = 371303408, 32574, 371303408, 32574, 984149104, 21982, 984149104, 21982, 3, 3, 3, ...
 side_ss8 = 371301600, 32574, 956180848, 21982, 0, 0, 0, 0, 3, 3, 3, 3, 3, 3,

 elem_ss9 = 371301600, 32574, 984185376, 21982, 0, 0, 0, ...
 side_ss9 = 371301600, 32574, 984185712, 21982, 0, 0, 0, 0, 3, 3, 3, 3, 3, 3, ...
```
so this fixes all those issues.

roystgnr (approved):
I'd prefer test coverage, but I'll trust your internal stuff.

This behavior feels like we're decimating some information ... but it wouldn't be the first or the last time we did so in Exodus...

jwpeterson:
I think we have a perfect "round trip" of data in our test case: single sideset in Exodus -> sideset + shellface set in libMesh -> compressed back down to single sideset in Exodus when writing.

jwpeterson:
Since this fixes a bug for us (and apparently doesn't introduce any new bugs) I'm going to merge as-is, but I'll make a TODO item for myself to add a unit test (the test case I used to develop this PR is too large for a unit test).

PR #3550, opened 2023-05-12
Elem::key() returns dof_id_type, not uint

roystgnr:
Let's keep around *all* the bits of that.

Pinging @NamjaeChoi to see if this fixes the fsanitize warning for him.  For merely zealous warnings I'll get a separate branch warning, but IMHO dropping half of a hash is an outright bug, even if hash collision handling means it's only a performance bug.

### DIFF
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -543,7 +543,7 @@ void UnstructuredMesh::find_neighbors (const bool reset_remote_elements,
   // are neighbors
   {
     // data structures -- Use the hash_multimap if available
-    typedef unsigned int                    key_type;
+    typedef dof_id_type                     key_type;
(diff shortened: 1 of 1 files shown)


NamjaeChoi:
Confirmed that this resolves the message:
```
../src/mesh/unstructured_mesh.C:569:42: runtime error: implicit conversion from type 'libMesh::dof_id_type' (aka 'unsigned long long') of value 14037056135608635786 (64-bit, unsigned) to type 'const unsigned int' changed the value to 254513546 (32-bit, unsigned)
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../src/mesh/unstructured_mesh.C:569:42 in
```

lindsayad (approved):
Too bad `auto` only fixes the second line of this two line PR

PR #3552, opened 2023-05-16
RBParameters: store std::vector<Real> instead of single Real

jwpeterson:
* Add RBParameters const_iterator custom iterator class replacing previous typedef.
* Return a reference to a std::pair when dereferencing const_iterator. This is needed to support code like the following which occurs in several places:
```
for (const auto & pr : mu_min_in) 
  libmesh_error_msg_if(mu_min_in.get_value(pr.first) > mu_max_in.get_value(pr.first), "Error: Invalid mu_min/mu_max in RBParameters constructor.");
```
  Side note: the above would ideally be replaced with a custom `RBParameters::operator>()` function which would then not expose as much of the internals of RBParameters.
* Drop deprecated `RBParameters::get_{extra}_parameters_map()` functions. There is not really a sensible way to implement these in a backwards-compatible way now that we have changed the type of the underlying container. IMO there is not really a point in leaving them in place just to produce a runtime error, it's better to get a compile time error instead.
* Update documentation for `RBParameters::get_{extra}_parameter_names()` functions.  Since the `get_{extra}_parameters_map()` function has now been removed, we should not direct people to call it.
* Drop extra newline from `RBParameters::print()`

### DIFF
--- a/include/reduced_basis/rb_parameters.h
+++ b/include/reduced_basis/rb_parameters.h
@@ -27,6 +27,7 @@
 #include <string>
 #include <map>
 #include <set>
+#include <vector>
 
 namespace libMesh
 {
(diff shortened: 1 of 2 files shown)


roystgnr (approved):
This is starting to look smooth.

PR #3553, opened 2023-05-18
Handle not-quite-orphaned DistributedMesh nodes

roystgnr:
The MOOSE test case that triggered a `--distributed-mesh` failure here was working on 15 or fewer processors, on 16 processors it was working on my own workstation, and it was working 80% of the time (are some Parmetis versions not deterministic?) on INL servers.

I'll see if I can come up with a more reproduceable unit test, but in the meantime this at least fixes the issue for me.

### DIFF
--- a/include/mesh/distributed_mesh.h
+++ b/include/mesh/distributed_mesh.h
@@ -644,6 +644,13 @@ protected:
   bool _is_serial_on_proc_0;
 
   /**
+   * A boolean remembering whether we've recently deleted top-level elements or
+   * not.  If so, we'll need to be extra careful when deleting "unused" nodes
+   * they used to have, because those nodes might still be used by ghost elements.
+   */
+  bool _deleted_coarse_elements;
+
+  /**
    * Cached data from the last renumber_nodes_and_elements call
    */
   dof_id_type _n_nodes, _n_elem, _max_node_id, _max_elem_id;
--- a/src/mesh/distributed_mesh.C
+++ b/src/mesh/distributed_mesh.C
@@ -1511,7 +1511,7 @@ void DistributedMesh::renumber_nodes_and_elements ()
         repartitioned_node_sets_to_push;
 
       auto ids_action_functor =
-        [this, &used_nodes, &repartitioned_node_pids,
+        [&used_nodes, &repartitioned_node_pids,
          &repartitioned_node_sets_to_push]
         (processor_id_type pid,
          const std::vector<std::pair<dof_id_type, boolish>> & ids_and_bools)
--- a/src/mesh/distributed_mesh.C
+++ b/src/mesh/distributed_mesh.C
@@ -715,10 +715,14 @@ void DistributedMesh::delete_elem(Elem * e)
   // Try to make the cached elem data more accurate
   _n_elem--;
 
-  // Was this a coarse element?  We'll have to be more careful with our nodes
-  // in contract() later; no telling if we just locally orphaned a node that
-  // should be globally retained.
-  if (!e->parent())
+  // Was this a coarse element, not just a coarsening where we still
+  // have some ancestor structure?  Was it a *local* element, that we
+  // might have been depending on as an owner of local nodes?  We'll
+  // have to be more careful with our nodes in contract() later; no
+  // telling if we just locally orphaned a node that should be
+  // globally retained.
+  if (e->processor_id() == this->processor_id() &&
+      !e->parent())
     _deleted_coarse_elements = true;
 
   // Delete the element from the BoundaryInfo object
--- a/src/mesh/distributed_mesh.C
+++ b/src/mesh/distributed_mesh.C
@@ -42,6 +42,7 @@ DistributedMesh::DistributedMesh (const Parallel::Communicator & comm_in,
                                   unsigned char d) :
   UnstructuredMesh (comm_in,d), _is_serial(true),
   _is_serial_on_proc_0(true),
+  _deleted_coarse_elements(false),
   _n_nodes(0), _n_elem(0), _max_node_id(0), _max_elem_id(0),
   _next_free_local_node_id(this->processor_id()),
   _next_free_local_elem_id(this->processor_id()),
@@ -75,6 +76,9 @@ DistributedMesh & DistributedMesh::operator= (DistributedMesh && other_mesh)
   // etc. subclasses.
   this->move_nodes_and_elements(std::move(other_mesh));
 
+  // But move_nodes_and_elems misses (or guesses about) some of our
+  // subclass values, and we want more precision than a guess.
+  _deleted_coarse_elements = other_mesh._deleted_coarse_elements;
   _extra_ghost_elems = std::move(other_mesh._extra_ghost_elems);
 
   // Handle remaining MeshBase moves.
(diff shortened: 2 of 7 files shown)


src/mesh/distributed_mesh.C
```diff
@@ -1442,12 +1457,133 @@ void DistributedMesh::renumber_nodes_and_elements ()
...
   std::set<dof_id_type> used_nodes;
 
+  // What used node info should we send from our proc?  Could we take ownership
+  // of each if we needed to?
+  std::map<processor_id_type, std::map<dof_id_type, bool>>
+    used_nodes_on_proc;
+
   // flag the nodes we need
   for (auto & elem : this->element_ptr_range())
     for (const Node & node : elem->node_ref_range())
```
lindsayad, line 1478: Not from this PR but I'd be consistent on explicit typing or `auto`

src/mesh/distributed_mesh.C
```diff
@@ -1442,12 +1457,133 @@ void DistributedMesh::renumber_nodes_and_elements ()
...
+  // nodes from it that are now no longer connected to any elements on it, but
+  // that are connected to their own semilocal elements.  We'll have to
+  // communicate to ascertain if that's the case.
+  this->comm().max(_deleted_coarse_elements);
+
+  // What used nodes do we see on our proc?
   std::set<dof_id_type> used_nodes;
 
+  // What used node info should we send from our proc?  Could we take ownership
+  // of each if we needed to?
```
lindsayad, line 1472: What does "each" mean?

src/mesh/distributed_mesh.C
```diff
@@ -1442,12 +1457,133 @@ void DistributedMesh::renumber_nodes_and_elements ()
...
+      std::map<processor_id_type, std::set<dof_id_type>>
+        repartitioned_node_sets_to_push;
+
+      auto ids_action_functor =
+        [this, &used_nodes, &repartitioned_node_pids,
+         &repartitioned_node_sets_to_push]
+        (processor_id_type pid,
+         const std::vector<std::pair<dof_id_type, boolish>> & ids_and_bools)
+        {
+          for (auto [n, local_to_pid] : ids_and_bools)
```
lindsayad, line 1516: I would rename `local_to_pid` to something like `can_make_owning_pid` or `can_make_sender_owner`. I confused myself for a while thinking that `local_to_pid` was a `processor_id_type`

src/mesh/distributed_mesh.C
```diff
@@ -1442,12 +1457,133 @@ void DistributedMesh::renumber_nodes_and_elements ()
...
+              // If we don't see a use for our own node, but someone
+              // else does, better figure out who should own it next.
+              if (!used_nodes.count(n))
+                {
+                  auto it = repartitioned_node_pids.find(n);
+                  if (local_to_pid)
+                    {
+                      if (it != repartitioned_node_pids.end() &&
+                          pid < it->second)
+                        it->second = pid;
```
lindsayad, line 1527: So we're always going to favor the lower processor id for ownership? Does this have precedence elsewhere? From an occam's razor standpoint I like this choice, and I imagine the load balancing effects are not all that significant (given this only crops up one out of 5 times on one test for only n procs > 15)

lindsayad (changes requested):

src/mesh/distributed_mesh.C
roystgnr: Clarified ("each node"), but I'm honestly unsure what else looked like a candidate antecedent.

src/mesh/distributed_mesh.C
roystgnr: That is the libMesh default, though we've got like three other alternatives that are better for load-balancing.  But but yeah, here we're not trying to load-balance the whole set of interface nodes; `Partitioner` does that elsewhere, we're just trying to handle a literal corner case.

src/mesh/distributed_mesh.C
roystgnr: Done

include/mesh/distributed_mesh.h
```diff
@@ -643,6 +643,13 @@ class DistributedMesh : public UnstructuredMesh
    */
   bool _is_serial_on_proc_0;
 
+  /**
+   * A boolean remembering whether we've recently deleted top-level elements or
+   * not.  If so, we'll need to be extra careful when deleting "unused" nodes
```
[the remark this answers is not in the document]
roystgnr: It's foolish that this seems to be changing as a standard.  Separating connected chunks of text by a level which scales with the strength of the connection is a good thing for readability.  Chapters get a page break and a heading.  Paragraphs get a blank line or at least an indentation.  Sentences get a punctuation and an extra space, and independent clauses only get punctuation.  The semantic connection between `not.  If` is weaker than the connection between `so, we'll`, and so the syntactical connection certainly should be weaker too.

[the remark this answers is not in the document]
roystgnr: In LaTeX, the single-space-between-sentences option is called `\frenchspacing`, because the luminary of computer science and typesetting couldn't imagine that there might someday be *two* countries who thought that was a good idea.

src/mesh/distributed_mesh.C
```diff
@@ -189,6 +192,7 @@ DistributedMesh::DistributedMesh (const DistributedMesh & other_mesh) :
 DistributedMesh::DistributedMesh (const UnstructuredMesh & other_mesh) :
   UnstructuredMesh (other_mesh), _is_serial(other_mesh.is_serial()),
   _is_serial_on_proc_0(other_mesh.is_serial()),
+  _deleted_coarse_elements(true), // better safe than sorry...
```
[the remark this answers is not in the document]
roystgnr: Hmm... in a copy from a `ReplicatedMesh`, the original wouldn't have this problem, because it would naturally see all elements connected to a node and would never try to orphan anything that had even a weird non-local connection ... but in a copy of the exact same elements?  I think the subsequent remote element deletion would just turn `_delete_coarse_elements` back on, so we wouldn't have a bug, but we also wouldn't end up taking advantage of skipping that sync step.

... and now that I think about it, that's actually a performance bug in other contexts.  `delete_remote_elements` is doing its `delete_elem` calls carefully, in a way that isn't going to trigger the problem that that sync step is necessary to repair.  I need to only be turning on `_deleted_coarse_elements` when a *local* element is deleted!

I'll add a commit to do that, and verify that it still fixes the MOOSE problem.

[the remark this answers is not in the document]
roystgnr: Still fixes the MOOSE problem!

include/mesh/distributed_mesh.h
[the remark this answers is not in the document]
lindsayad: Haha I did intentionally not group this with the "requested changes" 😄 as I feel that having it this way *is* valid

lindsayad (approved):
really appreciate the fix!

roystgnr:
Wish I had a better unit test for the fix.  If I keep the test and revert the fix I don't run into failures on less than 7 processors.  At least opt mode dies too, so a reversion would be likely to get caught in the tests we run on all PRs.

Speaking of ... I should turn on the dbg mode sweeps too this time, before we merge, just in case there's anything else subtle I'm missing.

lindsayad:
> in the tests we run on all PRs

We have a recipe that runs on PRs that includes np >= 7?

roystgnr:
>We have a recipe that runs on PRs (by default) that includes np >= 7?

Yeah, just for `METHOD=opt`, though.  So it's good when the unit test failure is actually a cppunit test failure (or at the very least a segfault) rather than coverage that just hits an internal assertion but manages to make it out alive if the assertion is skipped.

PR #3554, opened 2023-05-22
Add support for multi-step RBParameters objects in reduced_basis_ex4

jwpeterson:
This PR updates one of the `reduced_basis` examples to use the new capabilities provided in #3552. We previously had partial support for using "pre-evaluated thetas" in `RBEvaluation::rb_solve()`, and this PR extends that support to the computation of outputs and error bounds. The "pre-evaluated thetas" codepath is relevant to the multi-step `RBParameters` approach because the `RBEvaluation` class itself does not need to know that the `RBParameters` it is using has multiple steps... it just uses the pre-evaluated thetas, and the looping over steps can be handled at a higher level of the code, as is now done in reduced_basis_ex4.


dknez (approved):
This all looks good to me, and this is a good update of reduced_basis_ex4 to cover the case of multiple solves with different parameters.

PR #3555, opened 2023-05-22
Avoid using C-style casts in unit tests

jwpeterson:
I'm not sure exactly how this pattern got started and spread so much in the unit tests, but let's use C++ style casts when writing C++.

I also ended up updating code like "dof_id_type(i)" which I guess is technically C++ (constructing a temporary) and not a C-style cast, however I think using the cast syntax still gives the most clarity.


roystgnr:
> I'm not sure exactly how

Thank you for not typing `git blame` and finding out.  ;-)

PR #3559, opened 2023-05-23
Some clean-up commits for reduced_basis code

jwpeterson:
* Make APIs `const` when possible
* Default/delete special member functions.

I believe there's still some additional APIs that can be made `const` here, especially `RBConstruction`'s matrix accessors, but this is a decent chunk for now.


roystgnr (approved):
This looks beautiful from a quick skim; let me know if you need a closer review.

PR #3561, opened 2023-05-24
Fix deprecated calls to RBEvaluation::eval_output_dual_norm()

jwpeterson:
I only noticed this was an issue on the libMesh master merge, where we have a --disable-deprecated build.


jwpeterson:
This PR is required in order for us to get a master merge, so I'm going to go ahead and hit the green button.

PR #3562, opened 2023-05-24
More RBParameters updates

jwpeterson:
New interfaces that were either missed while working on #3554, or stuff that I only realized I would need after working on things a bit more.


roystgnr (approved):
You could have reorganized that if/elseif/else to save a line of code.  ;-)

PR #3566, opened 2023-05-25
Add EquationSystems read/write capability via istream/ostream

loganharbour:


roystgnr (approved):
Fingers cross on the CI, but I was hoping it would be this simple!

roystgnr:
The finger-crossing thing never works.

lindsayad:
Is this ultimately for https://github.com/idaholab/moose/issues/21011?

loganharbour:
> Is this ultimately for [idaholab/moose#21011](https://github.com/idaholab/moose/issues/21011)?

Yep. Gonna use libMesh for the system loads via the moose restart system.

loganharbour:
Good now. Gonna add unit tests and clean up doxygen.

There's some real ugly `std::function` use here... couldn't use a templated lambda because we don't put the methods in the headers. It's a unique case when you need to show _how_ to load the per-processor data, without actually loading it (because it might not exist)

loganharbour:
Ready.

roystgnr (approved):
I still like it, assuming John doesn't object to the technically-an-API-change or to us getting rid of that hack for reading our (third-most?) ancient file format with infinite elements.

jwpeterson (approved):
I don't really recall exactly the details here, but we aren't dropping support for reading xda/r files with infinite elements, just for really old versions of xda/r files.

roystgnr:
That's correct; only versions so old that we didn't put that info in the header should be affected.

PR #3568, opened 2023-05-26
Get a clean `-fsanitize=integer` run of `make check`

roystgnr:
Pinging @NamjaeChoi

I just ran in serial and on 4 procs in parallel, but this is runtime behavior checking and it's quite possible that corner cases exist at other processors counts, especially with AMR.

A clean run here requires `-fsanitize-ignorelist=yoursrcdir/fsanitize_ignorelist.txt`.  In the most extreme example, there are a few libstdc++ functions that aren't `-fsanitize=integer` clean and are never going to be.  The GCC team seems to believe that unsigned integer overflow and underflow are not undefined behavior (definitely correct), are necessary for optimal code when doing certain kinds of bit fiddling and looping (probably at least partially correct), and that there's no point in looking for them elsewhere (incorrect, but that's what we have the ignore list for).  In addition to upstream stuff like libstdc++ and Eigen, I've got ignore entries to cover the heavy bit-fiddling in our fparser fork and in some hashing headers.

It would be interesting to add a recipe for this to the devel->master merge, but a quick glance through the commits here should show how annoying it is to fix the lines that fsanitize=integer screams about.  90% are obvious false positives, and perhaps 90% of the rest are subtle false positives.  Oh, and runs with UBSan active are pretty slow, though that's probably my fault for combining it with dbg in my own testing.  Maybe we should add a recipe but it only runs opt mode and we don't make it required to pass?

Longer-term: I love the goal here, but I'm not sure how feasible it's going to be to extend it to MOOSE.  AFAIK nobody's ever even tried to get a `-Wconversion -Werror` build there, and I'm betting the libMesh history of sporadically doing things like that has made this cleanup much easier than it could have been.

### DIFF
--- a/contrib/fparser/util/bytecoderules_parser.cc
+++ b/contrib/fparser/util/bytecoderules_parser.cc
@@ -131,7 +131,7 @@ namespace
                       size_t indent)
         {
             std::string codehash, prevlabel;
-            for(size_t a=chain.size(); a-- > 0; )
+            for(ptrdiff_t a=chain.size(); a-- > 0; )
             {
                 codehash += chain[a];
                 std::string label = GenLabel(codehash);
@@ -509,7 +509,7 @@ namespace
             outstream << "\n" << Indent(indent+1);
 
         outstream << '"';
-        for(size_t a=so_far.size(); a-- > 0; )
+        for(ptrdiff_t a=so_far.size(); a-- > 0; )
         {
             if(a+1 != so_far.size()) outstream << " ";
             outstream << so_far[a].name;
@@ -1269,7 +1269,7 @@ namespace
             }
 
             Node* head = &global_head;
-            for(size_t b=sequence.size(); b-->0; )
+            for(ptrdiff_t b=sequence.size(); b-->0; )
             {
                 const Match& m = sequence[b];
                 bool dup = false;
--- a/include/mesh/namebased_io.h
+++ b/include/mesh/namebased_io.h
@@ -125,8 +125,8 @@ NameBasedIO::is_parallel_file_format (std::string_view name)
 {
   return ((name.rfind(".xda") < name.size()) ||
           (name.rfind(".xdr") < name.size()) ||
-          (name.rfind(".nem") + 4 == name.size()) ||
-          (name.rfind(".n") + 2 == name.size()) ||
+          (name.rfind(".nem") == name.size() - 4) ||
+          (name.rfind(".n") == name.size() - 2) ||
           (name.rfind(".cp") < name.size())
           );
 }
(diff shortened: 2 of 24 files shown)


contrib/fparser/util/bytecoderules_parser.cc
```diff
@@ -1269,7 +1269,7 @@ namespace
             }
 
             Node* head = &global_head;
-            for(size_t b=sequence.size(); b-->0; )
+            for(ptrdiff_t b=sequence.size(); b-->0; )
```
jwpeterson, line 1272: How was this working before with an unsigned type and this cutesy `b-->0` BS? When `b==0` the test fails and then you post-decrement so the number wraps to max size_t, but we never see it because the loop is done?

include/mesh/namebased_io.h
```diff
@@ -125,8 +125,8 @@ NameBasedIO::is_parallel_file_format (std::string_view name)
 {
   return ((name.rfind(".xda") < name.size()) ||
           (name.rfind(".xdr") < name.size()) ||
-          (name.rfind(".nem") + 4 == name.size()) ||
-          (name.rfind(".n") + 2 == name.size()) ||
+          (name.rfind(".nem") == name.size() - 4) ||
+          (name.rfind(".n") == name.size() - 2) ||
```
jwpeterson, line 129: I'd have guessed the sanitizer wouldn't like subtracting from an unsigned type that could possibly be 0? Oh, but std::string::rfind() returns std::string::npos which is effectively max size_t so adding to that would also be bad... not sure which is best actually.

src/base/dof_map_constraints.C
```diff
@@ -1717,7 +1717,7 @@ dof_id_type DofMap::n_local_constrained_dofs() const
   const DofConstraints::const_iterator lower =
     _dof_constraints.lower_bound(this->first_dof()),
     upper =
-    _dof_constraints.upper_bound(this->end_dof()-1);
+    _dof_constraints.lower_bound(this->end_dof());
```
jwpeterson, line 1720: This looks like an actual bugfix... was that suggested by the sanitizer?

jwpeterson (approved):
Looks good to me, but sounds like it was pretty annoying to use the sanitizer.

contrib/fparser/util/bytecoderules_parser.cc
roystgnr: Precisely.  It's apparently a popular idiom, too; this is what `-fsanitize=integer` was complaining about in a couple places in libstdc++ `basic_string` code.

src/base/dof_map_constraints.C
roystgnr: It wasn't suggested by the sanitizer; that only reports filename and line number and what type of behavior was detected, and leaves it up to the developer to figure out how to fix it.  Maybe in LLVM 30.

But ... yeah, this looks like a real bug fix.  I recently spent a ton of time fixing a bug that was triggered by an unexpected corner case with element deletion, just like this one could easily have been, and as annoying as the sanitizer is I have to say this fix alone made me feel like the annoyance was worth it so far.

PR #3570, opened 2023-05-31
Never go above the max order when computing FEs

lindsayad:
I started going down the road of modifying APIs but wound up here. I personally think running with the maximum order is better than erroring with a requested order that is too high, but I can definitely understand someone arguing the other way. These changes were enough to make @YaqiWang's test in idaholab/moose#24180 run and show good results (shown below).

When I initially started off modifying APIs and adding more `add_p_level` arguments, I wound up staring at our MOOSE `Assembly` helpers and thinking that if I disabled p refinement for those, then I would be implicitly disabling p refinement for actual user Lagrange variables of the same order. And that would be sad for a user who does want to p-refine a Lagrange basis up to second order from a mesh order of unity and initial variable order of unity.
```
Postprocessor Values:
+----------------+----------------+----------------+----------------+
| time           | dofs           | h              | l2_err         |
+----------------+----------------+----------------+----------------+
|   0.000000e+00 |   0.000000e+00 |   0.000000e+00 |   0.000000e+00 |
|   1.000000e+00 |   1.200000e+01 |   7.071068e-01 |   1.012842e-02 |
|   2.000000e+00 |   2.400000e+01 |   7.071068e-01 |   1.963216e-03 |
|   3.000000e+00 |   4.000000e+01 |   7.071068e-01 |   2.547371e-04 |
+----------------+----------------+----------------+----------------+
```

### DIFF
--- a/src/fe/fe.C
+++ b/src/fe/fe.C
@@ -323,16 +323,19 @@ void FE<Dim,T>::reinit_dual_shape_coeffs(const Elem * elem,
   this->elem_type = elem->type();
   this->_p_level = elem->p_level();
 
+  const auto order = Order(
+      std::min(static_cast<int>(this->get_order()),
+               static_cast<int>(FEInterface::max_order(this->get_fe_type(), this->get_type()))));
   const unsigned int n_shapes =
     this->n_shape_functions(this->get_type(),
-                            this->get_order());
+                            order);
 
   std::vector<std::vector<OutputShape>> phi_vals;
   phi_vals.resize(n_shapes);
   for (const auto i : make_range(phi_vals.size()))
     phi_vals[i].resize(pts.size());
 
-  all_shapes(elem, this->get_order(), pts, phi_vals);
+  all_shapes(elem, order, pts, phi_vals);
   this->compute_dual_shape_coeffs(JxW, phi_vals);
 }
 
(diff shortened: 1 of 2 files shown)


roystgnr:
Want to collapse that revert, at least in the branch for this PR?  (I'd save it elsewhere for now while we look at this)

roystgnr:
I think what bothered me about this idea in the past:

It leads to infinite loops when doing adaptive p refinement.  Instead of being told that your mesh is too coarse for any supported p level to work, you just keep p refining forever without actually improving your solution.

It doesn't preserve LBB constraints with mixed elements.  If you try to p-refine a Taylor-Hood element on a geometric element where we don't do p=3, then instead of getting an error message that that's unsupported, you just get a singular (or nearly, to FP) matrix and either a solver divergence or a garbage solution.

But most importantly:

Silently ignoring p refinement is kind of mean to users, isn't it?  If I have an input deck that asks for p refinement, and I have an isoparametric Lagrange variable, I want to be told that I need to use a different variable type to get what I asked for, I don't want to just silently get something asymptotically worse than I wanted.

I'd say we need to at least add some warnings here, and from the libMesh perspective that would be right, but those warnings wouldn't be very useful for MOOSE users, would they?  They'd be triggered for *every* p-refinement case, not just p-refinement of user Lagrange variables.

> if I disabled p refinement for those, then I would be implicitly disabling p refinement for actual user Lagrange variables of the same order

We don't have separate `FE` objects there?  We were failing to support p refinement even with no user Lagrange variables, because we had a framework-instantiated FELagrange regardless.  I'd assumed that was separate from whatever we created for users ... though I guess it wouldn't be, for optimization reasons.  That's a bit of a pain.

lindsayad:
> though I guess it wouldn't be, for optimization reasons. That's a bit of a pain.

Yea, that's the crux of it

lindsayad:
K I will explore the way that wouldn't hold any surprises for users. In MOOSE I will probably add some new data when we have p refinement in our assembly class, e.g. a helper that is not lie within all the `_fe` data of `Assembly`

lindsayad:
Went with #3574 instead

PR #3572, opened 2023-05-31
Poly2TriTriangulator refinement fixes

roystgnr:
This fixes some failure cases where triangulation refinement runs away creating sliver elements ... and it fixes more failure cases where it turns out we can't trust poly2tri to give us a true Delaunay triangulation.

The more careful behavior at boundaries will probably force us to regold tests in MOOSE.  Here's hoping it doesn't force too much of that further downstream.

This fixes https://github.com/idaholab/moose/issues/24474 for me.

### DIFF
--- a/include/geom/cell_hex.h
+++ b/include/geom/cell_hex.h
@@ -172,7 +172,7 @@ public:
    */
   virtual unsigned int n_permutations() const override final { return 24; }
 
-  virtual void orient(BoundaryInfo *) override final;
+  virtual bool is_flipped() const override final;
 
   /**
    * This maps each edge to the sides that contain said edge.
--- a/include/geom/cell_inf_hex.h
+++ b/include/geom/cell_inf_hex.h
@@ -187,7 +187,7 @@ public:
    */
   virtual unsigned int n_permutations() const override final { return 4; }
 
-  virtual void orient(BoundaryInfo *) override final;
+  virtual bool is_flipped() const override final;
 
   /**
    * This maps each edge to the sides that contain said edge.
--- a/include/geom/cell_inf_prism.h
+++ b/include/geom/cell_inf_prism.h
@@ -171,7 +171,7 @@ public:
 
   std::vector<unsigned int> sides_on_edge(const unsigned int e) const override final;
 
-  virtual void orient(BoundaryInfo *) override final;
+  virtual bool is_flipped() const override final;
 
   /**
    * This maps each edge to the sides that contain said edge.
--- a/include/geom/cell_prism.h
+++ b/include/geom/cell_prism.h
@@ -151,7 +151,7 @@ public:
    */
   virtual unsigned int n_permutations() const override final { return 6; }
 
-  virtual void orient(BoundaryInfo *) override final;
+  virtual bool is_flipped() const override final;
 
   /**
    * This maps each edge to the sides that contain said edge.
--- a/include/geom/cell_pyramid.h
+++ b/include/geom/cell_pyramid.h
@@ -175,7 +175,7 @@ public:
    */
   virtual unsigned int n_permutations() const override final { return 4; }
 
-  virtual void orient(BoundaryInfo *) override final;
+  virtual bool is_flipped() const override final;
 
   /**
    * This maps each edge to the sides that contain said edge.
--- a/include/geom/cell_tet.h
+++ b/include/geom/cell_tet.h
@@ -196,7 +196,7 @@ public:
    */
   virtual unsigned int n_permutations() const override final { return 12; }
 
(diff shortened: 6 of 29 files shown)


roystgnr:
The compiler warnings are easy to fix, but looks like we have at least three regoldings coming in MOOSE.

olinwc:
I'm pulling this branch to work on some of my test cases, will report the results once I have them

olinwc:
@roystgnr Devel mode reactor compiled with this branch I was able to mesh my gHPMR_2d_core_v33.i without any assertions or errors.

olinwc:
Unfortunately my MHTGR mesh seems to error out, but it does that with the current version of moose/libmesh, so I need to troubleshoot that.

olinwc:
Ok I was able to run my MHTGR mesh with this branch also. Error was being caused by a boundary I defined being renamed during stitching and me trying to use the old name. Works fine now.

olinwc:
@roystgnr Decided to run gHPMR_2d_core_v33.i in parallel and it just hangs or is just taking a lot longer, in serial it runs in 15 seconds, with four processors I've been waiting for several minutes just getting

```
Setting Up....................
```

roystgnr:
In serial even in devel mode it runs in 9 seconds for me, on 4 procs I get segfaults or assertion failures (boy I love parallel race conditions!) after more like 40 seconds.

Looking into it now.

roystgnr:
2 procs is enough to trigger failure too; hooray for small blessings.

roystgnr:
Whoa.  The failure is coming from `LaplaceMeshSmoother`??

roystgnr:
One failure was coming from `LaplaceMeshSmoother`.  If I turn off smoothing in XYDelaunay then I see assertion failures from ... `Partitioner::partition_unpartitioned_elements`?  Weirder and weirder.  There's got to be some underlying failure elsewhere.

roystgnr:
Okay; the underlying failure is in Poly2TriTriangulator.  I've found one obvious bug where it can get out of sync in parallel; still hunting for others.

@lindsayad wrote [an ironic three hours ago](https://github.com/idaholab/moose/pull/24619#issuecomment-1582907802) that "There may be some algorithms that require iterating in sync across processes for which you do have to have a deterministic ordering and I think we have those in the places we need them", but what he forgot is that I'm always writing **new** bugs to replace the old ones!

lindsayad:
Hahaha how timely. We have so many development intersections and cross-references these days. This from @olinwc included:

> Error was being caused by a boundary I defined being renamed during stitching and me trying to use the old name.

roystgnr:
@olinwc - give it a try now?

olinwc:
Since this is my personal Github I don't get notified on my work email... Trying this now!

olinwc:
So good news, `gHPMR_2d_core_v33.i` did run on 4 cores, it was slower, but it's also a small problem so that might just be expected

4 cores:
```
Finished Setting Up                                                                      [ 75.41 s] [  521 MB]
```
Serial:
```
Finished Setting Up                                                                      [ 14.61 s] [  210 MB]
```

Will try MHTGR next

olinwc:
MHTGR

Serial:
```
Finished Setting Up                                    [143.10 s] [ 1814 MB]
```
4 Cores:
```
Finished Setting Up                                    [790.19 s] [ 3770 MB]
```
No more hanging!

roystgnr:
> and performance fixes can wait.

I say, right before I dive into the code to look for fixes.

With a swath of low-hanging microoptimizations, I managed to make things faster by ... a fraction of a second?  So that's not the way to go.

But during the process I couldn't help but notice that gHPMR_2d_core_v33.i is triggering FOUR HUNDRED AND NINETEEN repartitionings, which for a mesh with no extrusions and no refinements is probably approximately four hundred eighteen too many.  We've long had vague plans to make the `is_prepared()`/`prepare_for_use()` system more fine-grained, but this seems to be the perfect motivational and test case; it's triggering an order of magnitude more overkill than I was imagining.

lindsayad:
That is truly brutal

olinwc:
@roystgnr That parallel behavior was all running with `METHOD=devel`

I'm also on a 2013 Mac Pro running locally, so people with more up-to-date local machines, like the new M1 Macs, are a lot faster per core.

roystgnr:
Rebased on `master`, to kick off new CI now that the tests to regold are temporarily disabled.

roystgnr:
That timeout on the `Test MOOSE debug` modules worries me.  Looked like the timing-out test was unrelated, so I just kicked it, but I'm also going to see if I get any weirdness in a local dbg modules `make test`.

PR #3574, opened 2023-06-01
Add way to disable use of p-refinement when doing FE::reinit

lindsayad:
I think this is the better way to handle this compared to #3570. With some new `Assembly` code I have this working for @YaqiWang's https://github.com/idaholab/moose/pull/24180


roystgnr (approved):
This is pretty much exactly what I was hoping would work!

PR #3576, opened 2023-06-05
Add some missing RBParameters APIs for "extra" parameters

jwpeterson:
Also add some unit tests so hopefully the coverage is not 0%.


jwpeterson:
Same failure in `mesh/preparedness.test` that we've been seeing.

PR #3582, opened 2023-06-13
global_indices tweaks

roystgnr:
None of these optimizations helped very much, but they didn't hurt either; if CI doesn't show any hiccups I'd just as soon keep them, and if it does (perhaps we really *needed* that `long double` intermediate calculation, at least for `--disable-unique-id` builds?) I'd at least be curious to know about it.


roystgnr:
Going to be taking another look at Hilbert-indices related performance issues shortly; unless there are objections I'll merge this first just to make sure I'm not duplicating or forking from it.

PR #3584, opened 2023-06-13
Error in EigenSystem::solve if we have constraints

lindsayad:
At least with SLEPc's eigen solver and the assembly routines in MOOSE, the result of a solve does not honor constraints such as hanging nodes/the solve is totally hosed. We do have `CondensedEigenSystem` which is a class intended for use with constraints. Unrelated to this PR that class will need to be updated in order to support shell matrices if MOOSE is to be able to use it

Here is the result of running `moose/test/tests/problems/eigen_problem/eigensolvers/ne.i` with half of the domain refined

Screenshot from 2023-06-13 12-10-15

### DIFF
--- a/src/systems/eigen_system.C
+++ b/src/systems/eigen_system.C
@@ -208,6 +208,10 @@ void EigenSystem::reinit ()
 
 void EigenSystem::solve ()
 {
+  if (get_dof_map().n_constrained_dofs())
+    libmesh_error_msg("EigenSystem does not support constrained degrees of freedom. If you wish to "
+                      "perform a solve on a system with constraints, then please use the "
+                      "CondensedEigenSystem class instead");
 
   // A reference to the EquationSystems
   EquationSystems & es = this->get_equation_systems();


dknez:
CondensedEigenSystem is used to remove the "constrained dofs" from the system so that you can solve the eigenproblem with constrained dofs removed. This is necessary because constrained dofs lead to spurious eigenvalues.

It looks like the condensed eigensolve is working correctly here. My guess is that you just need an "enforce_constraints_exactly" call after the condensed system solution is copied back to the original system. That should set the constrained dofs to their "correct" values.

lindsayad:
> t looks like the condensed eigensolve is working correctly here

My post probably wasn't very clear. That was the solve with `EigenSystem`, not `CondensedEigenSystem`. As you say `EigenSystem` gives poor results because 

> constrained dofs lead to spurious eigenvalues.

Most of the eigen work done with MOOSE uses shell matrices, so we will need to figure out how to adapt `CondensedEigenSystem` to work with those

roystgnr (approved):

dknez:
OK, understood.

> although maybe it wouldn't look so bad if we did add an enforce_constraints_exactly call after the solve?

Yes, I think it'd be interesting to try that. Visually based on the plot it looks to me like that might at least give a "correct looking" solution here.

lindsayad:
Ok this is with an `enforce_constraints_exactly`

Screenshot from 2023-06-13 16-08-19

dknez:
Looks good. I think it works fine, the only issue is that you have constrained dofs so you get spurious eigenvalues, but as long as those aren't within the part of the spectrum that you requested, then that doesn't affect your result.

lindsayad:
well in the first realistic case I looked at it seems like a spurious eigenvalue is in the spectrum we requested. @YaqiWang this is your Griffin example. Left with hanging nodes and right without

Screenshot from 2023-06-13 16-40-59

lindsayad:
given that, I think merging this error is the right thing to do especially given that we have a system class purposefully meant to handle these conditions (even if it will require some new development to support e.g. shell matrices)

dknez:
Agreed.

lindsayad:
Opened https://github.com/libMesh/libmesh/issues/3585 for future development

roystgnr:
> this is with an enforce_constraints_exactly

Man, I thought you were **joking** about trying that.

If we didn't already have a better way to do things, I guess it wouldn't be crazy to try rescaling the constraint lines in a matrix, so as to push spurious eigenvalues to the opposite end of the spectrum from the ones you're looking for ... but condensing them out entirely is a much much safer way to do things.

lindsayad:
> Man, I thought you were **joking** about trying that.

Haha well @dknez was curious and I was too. And for the simple test, the result looks great. But not for the first real test I tried 😢

PR #3588, opened 2023-06-20
Delete non-ASCII characters from header

jwpeterson:
This is kind of a real fix, but also a test of whether CIVET is working again.

### DIFF
--- a/include/solvers/time_solver.h
+++ b/include/solvers/time_solver.h
@@ -265,9 +265,9 @@ public:
   void set_solution_history(const SolutionHistory & _solution_history);
 
   /**
-￼   * A getter function that returns a reference to the solution history
-￼   * object owned by TimeSolver
-￼   * */
+   * A getter function that returns a reference to the solution history
+   * object owned by TimeSolver
+   */
   SolutionHistory & get_solution_history();
 
   /**


roystgnr (approved):

roystgnr:
So weird.  Personally I have to hold myself back from going crazy with UTF-8 in comments, `d∈ℝ³` and so forth, but where'd we get non-ASCII whitespace?

dschwen:
Why are we still so picky about unicode? Don't all compilers support it in the source code by now? Are we worried about copy pasting invisible characters that could lead to all kinds of confusion? It sure would be nice to use greek letters in comments with equations (or even in identifier names!)

jwpeterson:
> Why are we still so picky about unicode?

In this case, the use of non-ASCII characters wasn't intentional, so it made sense to remove it. I'm not aware if we have a libMesh rule explicitly allowing or disallowing unicode characters either in comments or in actual code? Personally I think it's a little gimmicky and probably wouldn't use them much, but if people really want to do it, I guess it would be fine, maybe after one more ASCII-only release?

roystgnr:
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2295r6.pdf

Clang basically assumes everything is UTF-8, but GCC and MSVC require a special flag or environment, and other compilers ... were apparently not worth mentioning directly?  But it looks like the "Edison Design Group" did the frontend for Portland Group compilers, from which NVidia's HPC stuff is descended, so maybe nvc++ is solid?

I love writing with Unicode, but I'd just as soon wait until Unicode support is in an official standard, not just a common extension.  I'd also worry a little bit about users' editor support, especially if we want to use unicode in identifiers.  XCompose is what makes me happy to write with Unicode, but last I looked you have to get third party programs to make that work on Windows or Mac, and even on Linux I've been maintaining my own `.XCompose` file to hit all the fun math characters that don't have compose sequences by default.  In the default Compose file, ironically, ¬∃ `∃`!

roystgnr:
I'd also worry a little bit about visual conflicts.  Does `ι` clearly look like an iota in everybody's editor?  If I'm using `α` and `Β` as identifiers, are you going to try to reference `B` or `Β` later?  They look indistinguishable to me in both my browser and my terminal, though one is ASCII "bee" (0x42) and the other is Greek "beta" (U+0392, 0xCE92).

Issue #3589, opened 2023-06-21
build fails: error in format-handler


BalticPinguin:
I recently updated my toolchain and ran into a compilation error.
I configure libMesh with  `  --enable-werror   --enable-petsc `  (I consider those to be the relevant flags) and during `make` I got the error
````
../master/src/solvers/petscdmlibmeshimpl.C: In function ‘PetscErrorCode DMCreateDomainDecomposition_libMesh(DM, PetscInt*, char***, _p_IS***, _p_IS***, _p_DM***)’:
../master/src/solvers/petscdmlibmeshimpl.C:447:85: error: format ‘%d’ expects argument of type ‘int’, but argument 8 has type ‘long unsigned int’ [-Werror=format=]
  447 |         if (elen != dlen) LIBMESH_SETERRQ1(((PetscObject)dm)->comm, PETSC_ERR_PLIB, "Failed to embed field %" LIBMESH_PETSCINT_FMT, d);
      |                                                                                                                                     ~
      |                                                                                                                                     |
      |                                                                                                                                     long unsigned int
/data/tm162/bin/DSK/petsc_opt/include/petscerror.h:57:127: note: in definition of macro ‘SETERRQ’
   57 |     PetscErrorCode ierr_seterrq_petsc_ = PetscError(comm, __LINE__, PETSC_FUNCTION_NAME, __FILE__, ierr, PETSC_ERROR_INITIAL, __VA_ARGS__); \
      |                                                                                                                               ^~~~~~~~~~~
../master/src/solvers/petscdmlibmeshimpl.C:447:27: note: in expansion of macro ‘LIBMESH_SETERRQ1’
  447 |         if (elen != dlen) LIBMESH_SETERRQ1(((PetscObject)dm)->comm, PETSC_ERR_PLIB, "Failed to embed field %" LIBMESH_PETSCINT_FMT, d);
      |                           ^~~~~~~~~~~~~~~~
In file included from /data/tm162/bin/DSK/petsc_opt/include/petscsys.h:44,
                 from /data/tm162/bin/DSK/petsc_opt/include/petscbag.h:5,
                 from /data/tm162/bin/DSK/petsc_opt/include/petsc.h:5,
                 from ./include/libmesh/petsc_macro.h:66,
                 from ../master/src/solvers/petscdmlibmeshimpl.C:20:
/data/tm162/bin/DSK/petsc_opt/include/petscsystypes.h:293:25: note: format string is defined here
[...]

PR #3590, opened 2023-06-22
PETSC_RELEASE_LESS_THAN -> PETSC_VERSION_LESS_THAN

roystgnr:
The former macro was only ever intended to be used temporarily, and it has the weird side effect of causing incompatibility with future non-release PETSc hashes if it's kept around too long.

This was aimed at #3589 but turned out to be a total red herring there.  I still think it's worth merging.


jwpeterson (approved):
Change looks fine to me, but as discussed there should be no change for anyone using a moderately recent PETSc.

PR #3594, opened 2023-06-22
Make Metis and Parmetis IGA-aware

roystgnr:
I'm not sure how to add regression tests for this sort of thing, since those partitioners are somewhat black boxes as far as we're concerned and there are always legitimate cases where they can't come up with a good partitioning ... but the newly-output `max(local unconstrained dofs)` numbers from Metis on `fem_example_ex5` on 16 ranks go down from 208 pre-partitioner-modifications to 94 post-modifications, on a problem where the theoretical optimum would be 80, and that's good enough for me for now.  With Parmetis the numbers go from 80 to 84, which is less good, but at least we're not clearly regressing ... and if I bump up the number of ranks then the `master` behavior degrades badly while the `parmetis_iga` behavior doesn't.

### DIFF
--- a/src/partitioning/metis_partitioner.C
+++ b/src/partitioning/metis_partitioner.C
@@ -202,10 +202,20 @@ void MetisPartitioner::partition_range(MeshBase & mesh,
 
             libmesh_assert_less (elem_global_index, vwgt.size());
 
-            // maybe there is a better weight?
             // The weight is used to define what a balanced graph is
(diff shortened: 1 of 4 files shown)


roystgnr:
That reported diff coverage is crummy, but I think the fix is to enable coverage on our distributed-sweep recipe rather than to add an extra parmetis run to our regular tests.

PR #3597, opened 2023-06-26
In FE, use _add_p_level_in_reinit whenever assigning _p_level

lindsayad:
Refs idaholab/moose#24180

### DIFF
--- a/src/fe/fe.C
+++ b/src/fe/fe.C
@@ -161,7 +161,7 @@ void FE<Dim,T>::reinit(const Elem * elem,
         {
           // Set the type and p level for this element
           this->elem_type = elem->type();
-          this->_p_level = elem->p_level();
+          this->_p_level = this->_add_p_level_in_reinit * elem->p_level();
 
           // Initialize the shape functions
           this->_fe_map->template init_reference_to_physical_map<Dim>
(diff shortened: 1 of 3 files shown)


src/fe/fe.C
```diff
@@ -58,10 +58,10 @@ FE<Dim,T>::FE (const FEType & fet) :
 
 
 template <unsigned int Dim, FEFamily T>
-unsigned int FE<Dim,T>::n_shape_functions () const
+unsigned int FE<Dim,T>::n_shape_functions (const bool add_p_level) const
 {
   return FE<Dim,T>::n_dofs (this->elem_type,
-                            static_cast<Order>(this->fe_type.order + this->_p_level));
+                            static_cast<Order>(this->fe_type.order + add_p_level * this->_p_level));
```
roystgnr, line 64: This seems weird to me.  The premise of caching `elem_type` and `_p_level` is that they should represent whatever we did our last `reinit()` with, isn't it?  So we should never have to override the `n_shape_functions()` behavior, because "return the number of shape functions we would have had if we had done a thing we didn't do" isn't a thing anyone needs to know, is it?

It looks to me like we're setting `this->_p_level = elem->p_level()` even when `this->_add_p_level_in_reinit` is false.  Isn't that a bug, and if we fixed that would it obviate the need for this PR?

lindsayad: Do you regard `get_order` as cached data as well? If so, I will fix that API as well

lindsayad: I would suppose yes since it's using `this->_p_level`

roystgnr (approved):
If this works for the MOOSE issues, this definitely looks like better code at the libMesh level.

lindsayad:
yep it works for MOOSE!

Issue #3580, opened 2023-06-12
Make mesh stitcher issue warnings or errors


YaqiWang:
@roystgnr brought this up during a meeting; when the stitcher tries to stitch two meshes, even though the two side sets to be stitched are not connected, the stitcher will return without any error. For flexibility, we might do not want the two side sets exactly overlap geometrically, but such a case should be considered as a warning or an error. We could also have cases that nodes on one side geometrically within the side set on the other side but do not have any nodes to be merged with. It could be a good idea to issue a warning at the end to tell users that a list of such nodes.

dknez:
We do have errors for this, e.g. see `enforce_all_nodes_match_on_boundaries`.

It's important to be able to turn that off, though, since in some cases we want to allow the nodes to be "non-conforming" and to instead add constraints to connect the meshes.

PR #3599, opened 2023-06-28
Improve handling of mesh stitching for non-conforming meshes

pmoseley:
This adds a new flag "merge_boundary_nodes_all_or_nothing" (open to better names?) to `stitch_meshes` and `stitch_surfaces`, which can help to avoid scenarios where you would otherwise create a mix of merged and unmerged nodes at a stitched interface.

I've also add a boolean return value to help identify if nodes were actually stitched; I think this should help with #3580 .

### DIFF
--- a/include/mesh/unstructured_mesh.h
+++ b/include/mesh/unstructured_mesh.h
@@ -192,8 +192,10 @@ public:
    * There is no simple a priori relationship between node IDs in "this" mesh
    * and other_mesh and node IDs in the stitched mesh because the number of nodes (and hence
    * the node IDs) in the stitched mesh depend on how many nodes are stitched.
+   *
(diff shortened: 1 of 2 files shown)


roystgnr (approved):
I like this the way it is ... but as long as we're changing the return value anyway, would it make sense to return the number of matching nodes, rather than just a boolean for whether that number is non-zero?  `map::size()` is supposed to be `O(1)` (and likewise if we switch to `unordered_map` later).

pmoseley:
It's a good comment; there's not much reason not to return the number.

I think it will only be useful in some rare cases, because if the user doesn't already know the total number of nodes on the boundary they'll be missing some context. But I'm sure there are some cases where it could be useful.

jwpeterson (approved):

PR #3601, opened 2023-06-29
Exodus tri 3d fix

roystgnr:
I'm trying to get much better unit test coverage here before we merge, but I've already had to make a Nemesis change that I'm afraid might break some downstream exodiffs, so let's see how bad the damage in CI is now.

Refs https://github.com/idaholab/moose/issues/24631


roystgnr:
Probably shouldn't have pushed that 3rd commit yet ... I bet having that without a re-bootstrap breaks "make installcheck"...

roystgnr:
And there's the regolding failures: "dimension doesn't agree" from Nemesis exodiffs.  We'll disable those temporarily once this is ready to merge.

roystgnr:
I'm finally getting back to this now.  I've given up on replicating it on my own machine, but an environment with the same set of modules gives me the same failure, so hopefully I can boil it down there.

roystgnr:
And the good news is that a test run without using multiple threads or multiple MPI ranks fails the same way.  I was afraid that I might be having trouble replicating because I don't use the same multithreading backend...

roystgnr:
The new tests here are failing even without the new code.  For some reason, in non-HDF5 builds, we're passing the correct `node_sets_node_list` node indices (translated to Exodus 1-based) to `ex_put_concat_sets`, and yet we're getting zero vectors in the output Exodus file.  We then scream when reading the output back in, because 0 doesn't make sense as a 1-based index.

roystgnr:
Ok, that was the bug.  All that's left is the nemesis exodiffs.

roystgnr:
Ooh, I think we've got it this time.  I'll merge this as soon as the last CI recipe confirms.

PR #3602, opened 2023-06-30
Set operators also for unrestricted solve & without preconditioner.

BalticPinguin:
I have discussed in  #3598, that solving linear systems with `ShellMatrix` makes problems in the current code.
The missing `KSPSetOperators`-call is now added here; for me, this fixes the problems.


jwpeterson (approved):
Looks like a good bugfix to me, I guess we never ran into this because we don't do many subset solves.

roystgnr:
Thanks so much for the fix; this was probably from my latest refactoring sweeps.  I tried my best to reproduce the existing behavior in the cases for which we don't have test coverage, but at some point we really just need test coverage...

PR #3603, opened 2023-07-10
Add in-place L2 norm difference method

loganharbour:


loganharbour:
@roystgnr please

loganharbour:
@jwpeterson - Roy is looking for your opinions for naming `readable()` and `comparable()` in https://github.com/libMesh/libmesh/pull/3603/commits/9acd8d3ce15307c75966c8bd9a22928c5ee552fb; mind taking a look?

jwpeterson:
OK, so the question is just if those are good names for the functions? Seems OK to me, might change `comparable` -> `compatible` since the word "comparable" is commonly used in the context of STL strict weak orderings.

loganharbour:
> OK, so the question is just if those are good names for the functions? S

Yep - that's it. We hit some stupid mistakes not understanding a result when we forgot to close, so decided to clean that up here.

> might change `comparable` -> `compatible` since the word "comparable" is commonly used in the context of STL strict weak orderings.

Works for me.

loganharbour:
@roystgnr good to go.

roystgnr (approved):

PR #3607, opened 2023-08-01
Remove unnecessary indirection from DiffContext

roystgnr:
This gives us a little bit of speedup and a surprisingly large reduction in memory use, especially when we have thousands of variables as in https://github.com/idaholab/moose/issues/25007

### DIFF
--- a/include/systems/diff_context.h
+++ b/include/systems/diff_context.h
@@ -124,8 +124,7 @@ public:
   const DenseSubVector<Number> & get_elem_solution( unsigned int var ) const
   {
     libmesh_assert_greater(_elem_subsolutions.size(), var);
-    libmesh_assert(_elem_subsolutions[var]);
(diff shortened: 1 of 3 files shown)


lindsayad (commented):
It is surprising indeed that this significantly reduces memory

roystgnr:
In that thousand-variable case, we've got a million subjacobian entries, so you'd expect at least 8MB less memory usage, maybe 16MB or even 24MB when you consider heap overhead from a bunch of individual allocations getting replaced by single blocks.

But yeah, I'm seeing MOOSE report hundreds of MB less peak RAM use with that test file and this patch.  I have no explanation for that.

The next obvious step is to not actually allocate element Jacobians or sub-Jacobians at all in cases (like GenericProjector, where our linear systems are smaller even than the sub-Jacobians and can be solved one-at-a-time) where we don't use them.  Trying that out now.

PR #3611, opened 2023-08-04
More work on allowing selective p refinement

lindsayad:
This is in support of issue idaholab/moose#25105 and PR idaholab/moose#25108


roystgnr:
I like the implementation, not yet thrilled with the API.

1. As long as we can turn this on, let's also add the ability to turn it off?

2. Doing this on a per-variable-group basis would be much more flexible.  Would it be too unreasonable to do a loop over variable groups in https://github.com/idaholab/moose/pull/25108 and disable p refinement for each of the `LAGRANGE` ones?

3. There the potential to let users get themselves into real trouble here.  p-refine, reinit systems, *then* turn this on, and suddenly you've got a bunch of allocated DoF indices that don't correspond to anything your FE objects can see.  Is there any way we can catch that?  I don't see one immediately.  "*Don't* p-refine, reinit systems, then turn this on because you plan to p-refine later" is a perfectly reasonable process, yet unless we loop over elements and test every `p_level()` I don't see how we can distinguish the reasonable process from the bug generator.  I guess we already rely on downstream code being smart about knowing when it needs a `reinit()` of one thing or another, so this doesn't get any worse?  And even "p-refine, reinit, turn this on, reinit again" actually might work out-of-the-box with the sorts of p-hierarchic bases that benefit from that refinement.

lindsayad:
> * As long as we can turn this on, let's also add the ability to turn it off?

Sounds good

> * Doing this on a per-variable-group basis would be much more flexible.  Would it be too unreasonable to do a loop over variable groups in [[WIP] Allow selective exclusion of Lagrange from p-refinement idaholab/moose#25108](https://github.com/idaholab/moose/pull/25108) and disable p refinement for each of the `LAGRANGE` ones?

I think we can do that

lindsayad:
How about this little guy?

roystgnr (approved):

roystgnr:
MOOSE failures are unrelated, so let's live dangerously.

PR #3613, opened 2023-08-07
remap_subdomain_ids capability in stitch_meshes

roystgnr:
I'm going to try to default it to "true", on the theory that if we don't see a lot of CI failures then this will be a cheap feature to enable by default whereas if we do see a lot of CI failures then this will be an important feature to enable by default.

This catches the failure cases in https://github.com/idaholab/moose/issues/25070 for me.  Shame that it also catches cases in MOOSE for me; I'm going to let it rip here to see what the full CI damage is.


roystgnr:
Well this is looking nice so far.  CI *only* has the 14 failures I could reproduce locally, and 13 of those failures turn out to be due to a bug in this PR.  If I can figure out the 14th then we're in business.

roystgnr:
Two steps forward, one step back.  But those are more than just "we renumbered subdomain ids" exodiffs; in dbg mode we're actually exceeding max subdomain_id_type somehow... looks like that's a test I need to be doing in opt too, whatever the resolution is.

roystgnr:
> we're actually exceeding max subdomain_id_type somehow...

Okay, no "somehow" about it.  https://github.com/idaholab/moose/pull/19950/ is the culprit.  `UINT16_MAX - 1`? `UINT16_MAX - 1 - _peripheral_regions`?  Once one of those is in the mesh (ncdump of an intermediate `AssemblyMeshGenerator` mesh tells me it's "65534" in the failing core_hex.i case), we basically have no room (either in this new libMesh code or in `MooseMeshUtils::getNextFreeSubdomainID()`) to automatically allocate other subdomain ids past the existing range.

Pinging @shikhar413 - is there any way we can just stop using hard coded subdomain ids, and instead get auto allocated ids for these sorts of purposes?  There's just not a big enough namespace for every mesh generator to pick its own magic numbers and pray there will be no conflicts.

shikhar413:
Possibly, I remember there was a lot of contention over this with @GiudGiud when we were originally developing this capability since we wanted a way to be able to tag elements with a unique subdomain ID to reference where they were in the mesh later on in `generate` to be able to modify mesh properties and tag any necessary EEIDs, and we chose a namespace specifically large enough to avoid possible collisions with other mesh generators. This topic could use some re-visiting since it has been a few years and now maybe we can be smarter about using EEIDs instead of subdomain IDs to figure out what type of elements (pins, assembly background, or assembly duct?) are we looking to modify the mesh properties of in `generate`. This will take some time though, and I can't comment right now about whether using auto allocated ids would work instead.

Issue #2882, opened 2021-03-16
Prism18-mesh with exodus-2


BalticPinguin:
Dear all,
I need some help for further analyzing a problem I just encountered:
I generate and write a mesh with
```  
  MeshTools::Generation::build_cube (mesh, /*nx= */5, /*ny=*/5, /*nz=*/3,
                                        /*xmin=*/ -5, /*xmax=*/5,
                                        /*ymin=*/ -5, /*ymax=*/5,
                                        /*zmin=*/ -3, /*zmax=*/3,
                                        PRISM18 /* this works*/           //PRISM18 /*THIS FAILS! */
                                    );

  ExodusII_IO(mesh).write ("finite_cube.e");
```
but when I open the `finite_cube.e` with ParaView, it looks quite odd: First, only one element in z-direction is present but also some elements overlap and some (e.g. element with id= 2) have weird shapes.

Unfortunately, I don't know how to further analyze the issue to see, e.g., if it is a problem with printing the file or if it is on ParaViews side!?
When using `PRISM15`, everything looks good. Also, according to `mesh.print_info()`, the generated mesh itself looks fine: It has the correct number of nodes and elements.
Has someone some experience with tools to analyze exodus-2 files or can give some other useful hints?
Many thanks in advance,
Hubert

roystgnr:
Could you loop over the mesh and check whether each element is affine?  If the answer is yes then the problem has to be on our end or Paraview's; if no then we're doing something wrong in mesh generation.

BalticPinguin:
>  check whether each element is affine? 

You mean to check for `elem->has_affine_map()`? yes, they are all affine. Also, if I generate the `PRISM18`-mesh and than convert all elements to first order, I get the same output as for `PRISM6`.

roystgnr:
```has_affine_map()```, right.  Converting to first order could simply hide a bug, but unless we have an exactly-canceling-it-out bug in ```has_affine_map()```, that should have caught any problems in mesh generation.

Perhaps our libmesh-to-exodus node numbering conversion is broken for PRISM18?  Looking at exodusII_io_helper.C I don't actually see any node map; we seem to believe that their WEDGE18 numbering should match our PRISM18 numbering exactly.  Which ... maybe?

BalticPinguin:
yes, that indeed is a very good hint.
It seems [this](https://gsjaardema.github.io/seacas-docs/html/element_types.html#wedge) is the official documentation where unfortunately the wedge18 is missing.
[...]

PR #3616, opened 2023-08-09
Fixes for 2d Nédélec one implementation and provided example

nmnobre:
Hi,

Just a few small fixes pertaining Nédélec one elements.
To my untrained eyes - I've just started playing with libMesh - it seems like TRI7 wasn't being properly distinguished from TRI6.
Apologies if I got anything wrong. :)

Cheers,
-Nuno

### DIFF
--- a/examples/vector_fe/vector_fe_ex3/vector_fe_ex3.C
+++ b/examples/vector_fe/vector_fe_ex3/vector_fe_ex3.C
@@ -69,15 +69,15 @@ int main (int argc, char ** argv)
   Mesh mesh(init.comm());
 
   // Use the MeshTools::Generation mesh generator to create a uniform
-  // grid on the square [-1,1]^D. We must use TRI6 elements for the
-  // Nedelec triangle elements.
+  // grid on the square [-1,1]^D. We must use at least TRI6 or QUAD8 elements
+  // for the Nedelec triangle or quadrilateral elements, respectively.
   std::string elem_str =
     command_line_value(std::string("element_type"),
                        std::string("TRI6"));
 
-  libmesh_error_msg_if(elem_str != "TRI6" && elem_str != "QUAD8" && elem_str != "QUAD9",
+  libmesh_error_msg_if(elem_str != "TRI6" && elem_str != "TRI7" && elem_str != "QUAD8" && elem_str != "QUAD9",
                        "You selected: " << elem_str <<
-                       " but this example must be run with TRI6, QUAD8, or QUAD9.");
+                       " but this example must be run with TRI6, TRI7, QUAD8, or QUAD9.");
 
   MeshTools::Generation::build_square (mesh,
                                        grid_size,
--- a/examples/vector_fe/vector_fe_ex3/vector_fe_ex3.in
+++ b/examples/vector_fe/vector_fe_ex3/vector_fe_ex3.in
@@ -1,4 +1,4 @@
-# The coarse rid size from which to start adaptivity
+# The coarse grid size from which to start adaptivity
 grid_size = 20 
 
 # Turn this off to see the NewtonSolver chatter
--- a/examples/vector_fe/vector_fe_ex4/vector_fe_ex4.C
+++ b/examples/vector_fe/vector_fe_ex4/vector_fe_ex4.C
@@ -71,9 +71,8 @@ int main (int argc, char ** argv)
   Mesh mesh(init.comm());
 
   // Use the MeshTools::Generation mesh generator to create a uniform
-  // grid on the square [-1,1]^D. We must use TRI6 elements for the
-  // Nedelec triangle elements.
-
+  // grid on the square [-1,1]^D. We must use at least TET10 or HEX20 elements
+  // for the Nedelec tetrahedral or hexahedral elements, respectively.
   std::string elem_str =
     command_line_value(std::string("element_type"),
                        std::string("HEX27"));
@@ -90,9 +89,9 @@ int main (int argc, char ** argv)
   // [3]: FEniCS, https://fenicsproject.org/pub/book/book/fenics-book-2011-06-24.pdf
   // [4]: Monk, https://icerm.brown.edu/materials/Slides/tw-18-7/Finite_Element_Methods_for_Maxwells_Equations_%5D_Peter_Monk,_University_of_Delaware.pdf
   // [5]: Hiptmair at al., https://www.sam.math.ethz.ch/sam_reports/reports_final/reports2009/2009-04_fp.pdf
-  libmesh_error_msg_if(elem_str != "TET10" && elem_str != "HEX20" && elem_str != "HEX27",
+  libmesh_error_msg_if(elem_str != "TET10" && elem_str != "TET14" && elem_str != "HEX20" && elem_str != "HEX27",
                        "You entered: " << elem_str <<
-                       " but this example must be run with TET10, HEX20 or HEX27.");
+                       " but this example must be run with TET10, TET14, HEX20 or HEX27.");
 
   MeshTools::Generation::build_cube (mesh,
                                      grid_size,
--- a/examples/vector_fe/vector_fe_ex4/vector_fe_ex4.in
+++ b/examples/vector_fe/vector_fe_ex4/vector_fe_ex4.in
@@ -1,4 +1,4 @@
-# The coarse rid size from which to start adaptivity
+# The coarse grid size from which to start adaptivity
 grid_size = 4 
 
 # Turn this off to see the NewtonSolver chatter
--- a/src/fe/fe_nedelec_one.C
+++ b/src/fe/fe_nedelec_one.C
@@ -65,7 +65,11 @@ void nedelec_one_nodal_soln(const Elem * elem,
           case TRI7:
             {
               libmesh_assert_equal_to (elem_soln.size(), 3);
-              libmesh_assert_equal_to (nodal_soln.size(), 6*2);
+
+              if (elem_type == TRI6)
+                libmesh_assert_equal_to (nodal_soln.size(), 6*2);
+              else
+                libmesh_assert_equal_to (nodal_soln.size(), 7*2);
               break;
             }
           case QUAD8:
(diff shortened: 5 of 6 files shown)


roystgnr:
This all looks good to me.  The failing "Test MOOSE min gcc" is an unrelated downstream problem we're clearing up; this should be good to merge once everything else is clear.

nmnobre:
I think there might be something wrong with the output.
When trying to visualise the solution in Paraview, with TRI7 I can only see the boundaries, whereas with TRI6 I can see the whole domain as expected...

nmnobre:
> I think there might be something wrong with the output of ex3.
> When trying to visualise the solution in Paraview, with TRI7 I can only see the solution at the boundaries, whereas with TRI6, QUAD8 and QUAD9 I can see the solution over the whole domain as expected...

I'm guessing that's why @pbauman wrote this way back then:
https://github.com/libMesh/libmesh/blob/4ae34d8f6c088ce992abcff8b31fef7077ae4c47/examples/vector_fe/vector_fe_ex3/vector_fe_ex3.C#L71-L73

The same happens with TET14. I'll revert the changes that allow the example to be run with TRI7, I'd just like to understand why this is a limitation? Do we need to forbid TRI7 and TET14 more assertively for Nedelec elements? What happens if someone tries to use them in their own problem specification? It _looks_ like a limitation when writing the output exodus file, but I can't seem to understand what's happening. :/

jwpeterson:
I think the @pbauman comment may be referring to using `TRI6` vs `TRI3`? The `TRI7` element was added fairly recently by @roystgnr, and it's possible that there's some issue with the exo output for it. Can you let us know exactly how to reproduce the problem using vector_fe_ex3 and your branch?

pbauman:
> I think the @pbauman comment may be referring to using TRI6 vs TRI3? 

Yeah, you need TRI6 because you need mid-edge nodes to use Nédélec elements.

roystgnr:
I'd guess this is an Exodus output issue.  I have test coverage for internal correctness with `TRI7` but I don't think I've ever tried to visualize with it ... and it's possible that users haven't done that either.  There's not much point in using a `Tri7` by itself (the extra mid-element node does give you an extra bubble function to use for the geometric mapping, but for non-Lagrange FE shape functions we could already just store their mid-element shape function indices on the `Elem` itself); `Tri7` is mostly useful indirectly as a side element to allow higher-order 3D elements with triangular faces to have cubic and higher face DoFs.

nmnobre:
> I think the @pbauman comment may be referring to using `TRI6` vs `TRI3`? The `TRI7` element was added fairly recently by @roystgnr, and it's possible that there's some issue with the exo output for it. Can you let us know exactly how to reproduce the problem using vector_fe_ex3 and your branch?

That actually makes a lot of sense, thank you. :)

To reproduce:
1) Build as per the readme;
2) Go to `<install_dir>/examples/vector_fe/ex3` and run `METHOD=dbg make`;
3) Run w/ `./example-dbg element_type=TRI6 -pc_type jacobi` and `mv out.e TRI6.e`;
4) Run w/ `./example-dbg element_type=TRI7 -pc_type jacobi` and `mv out.e TRI7.e`;
5) Open both `TRI6.e` and `TRI7.e` in Paraview and notice there's no bulk info for the latter.

roystgnr:
> notice there's no bulk info for the latter

I'm not sure what that means, but I'm certainly noticing a mess.  If I use the IOSS Reader option in Paraview to open TRI7.e, I get a failure from vtkIOSSUtilities.cxx, "Element of topology 'tri7' with 7 nodes is not supported.".  If I use the Legacy ExodusII Reader option, I get garbage.  It's as if something isn't advancing fast enough through node indices, and so we're just getting a bunch of tangled triangles attached to half the nodes.

This is with Paraview 5.10.1 - what version (and what reader option) are you using?

roystgnr:
I'm looking at the `ncdump` output of the most simple case, `grid_size=1`, and it looks fine to me.  Nodes all in the right places, `num_nod_per_el1 = 7`, connectivity matching the `TRI7` picture in https://sandialabs.github.io/seacas-docs/exodusII-new.pdf

jwpeterson:
So dow we think this is a Paraview bug? When we previously had an issue with PRISM18 elements in #2882, [acbauer](https://github.com/acbauer) was able to get a fix into Paraview upstream for us.

nmnobre:
> > notice there's no bulk info for the latter
> 
> I'm not sure what that means

The boundary data is visualisable and correct, everything else, i.e. on the inner points of the domain, is missing.

> This is with Paraview 5.10.1 - what version (and what reader option) are you using?

5.11.1, but I don't get an option to select a reader?

> Oh, and what Exodus version? I'm on a 5.22 build here.

Same, as provided with libMesh w/o any tricks.

> I'm looking at the ncdump output of the most simple case, grid_size=1, and it looks fine to me.

Oh I was wondering if something like that existed! I was using a hex editor... 😰
How does this work though, is an Exodus file format basically a netCDF file?

> So do we think this is a Paraview bug?

VisIt doesn't seem to like it either. They both use VTK though...

roystgnr:
> The boundary data is visualisable and correct, everything else, i.e. on the inner points of the domain, is missing.

This sounds almost as if it's only plotting EDGE3 boundaries and just plain giving up on the TRI7s.  It's weird that you wouldn't get an error message, though.  Unless maybe you're just not an old person like me, so you're using one of those new-fangled "gooeys" to launch Paraview?  I actually only see the error message in the console from where I've been running `paraview foo.e`; in the Paraview window itself I just get what looks like an empty file loaded.

> 5.11.1, but I don't get an option to select a reader?

That requires a bit of hassle.  You have to go to Tools, Manage Plugins, and load (and/or set to auto-load) the LegacyExodusReader plugin there.  They managed a nice quick fix for the IOSS bug that had us figure this all out, but I'm still using the older Paraview (prebuilt Ubuntu package) so I'm still using the legacy reader.

> How does this work though, is an Exodus file format basically a netCDF file?

Depends on the version and configuration, but yeah, it's always either NetCDF or HDF5 under the hood.

jwpeterson:
FYI @nmnobre I sent you an invite to join the group whose libMesh PRs are tested automatically. For now I will start them all manually.

nmnobre:
> This sounds almost as if it's only plotting EDGE3 boundaries and just plain giving up on the TRI7s. It's weird that you wouldn't get an error message, though. Unless maybe you're just not an old person like me, so you're using one of those new-fangled "gooeys" to launch Paraview?

I am, I feel like a WIMP now. 😩

> That requires a bit of hassle. You have to go to Tools, Manage Plugins, and load (and/or set to auto-load) the LegacyExodusReader plugin there. They managed a nice quick fix for the IOSS bug that had us figure this all out, but I'm still using the older Paraview (prebuilt Ubuntu package) so I'm still using the legacy reader.

Thanks! I've enabled the LegacyExodusReader, and as you said, it looks like a mess:

Screenshot 2023-08-17 at 20 41 54

When I use the IOSS Reader, it's as if there's no data (I have to select the boundary sets to actually see something).

roystgnr:
Okay, your experience and mine seem to match.  Should we file a Paraview bug report?  If I strip out variables/nodes/sidesets on a 2-triangle mesh, I get down to 60 lines of ncdump output, and yet it's still clear that we hit an assertion with IOSS and we're missing a triangle with the legacy display.

nmnobre:
> Okay, your experience and mine seem to match. Should we file a Paraview bug report? If I strip out variables/nodes/sidesets on a 2-triangle mesh, I get down to 60 lines of ncdump output, and yet it's still clear that we hit an assertion with IOSS and we're missing a triangle with the legacy display.

Hang on, see https://gitlab.kitware.com/vtk/vtk/-/commit/f80cc5c6e206b3a9f0a442566cf4ceb3c4d8b8f8. It seems Paraview is just shipping with an old version of VTK w/o the newly added support for TRI7.

roystgnr:
Ooh, that could do it.  I wouldn't want to bother to file a bug report for the legacy reader alone.

roystgnr:
I'll hang on to my test cases, though; if you upgrade and it's still giving you problems then we can still report them after that.  With 75 lines we can get an 8-triangle file, and that's enough to exhibit tangling triangles, not just missing ones.

nmnobre:
> Ooh, that could do it. I wouldn't want to bother to file a bug report for the legacy reader alone.

I'll try to build Paraview from source (wish me luck) and see if that sorts it!

nmnobre:
Paraview @[31dbe2d6](https://gitlab.kitware.com/paraview/paraview/-/tree/31dbe2d6832af83b07449c6e764cd48c47d9ba70) w/ VTK @[40c19ea4](https://gitlab.kitware.com/vtk/vtk/-/tree/40c19ea4e8f67c0edd104b4578ba0bc8efcac4ed). 🎉

Plotted as points so you can see (maybe with some effort) all the nodes in their rightful place!

Screenshot 2023-08-17 at 23 30 17

jwpeterson (approved):
The changes in this PR look good to me, unless you wanted to add a comment somewhere in the example noting that the output will not look correct unless one is using a sufficiently new version of Paraview.

nmnobre:
Unfortunately, TET14 isn't a thing in VTK, they only have TET15.
Maybe in the same way they treat TET10 and TET11 the same, they could also treat TET14 and TET15 the same...
Or conversely, why do we have TET14 and not TET15, but then have no HEX26 but do have HEX27?

roystgnr:
The root answer to "why don't we have" is usually either "someone made a contrary assumption in like 2007 and now we'd break a hundred classes if we changed it" (e.g. anisotropic h refinement) or "Roy's lazy" (e.g. probably everything else).

The proximate answer in the case of TET14 vs TET15, though, is that TET14 gives us mid-side nodes, which we need if we want to support FE types with degrees of freedom whose support is restricted to the two element sharing a side, but TET15 only adds a central node, which we don't need because we can already store bubble function degrees of freedom on the element itself.  Technically we could also get away with HEX26 the same way, but we have HEX27 because people do like the full triquadratic LAGRANGE mapping functions, and for LAGRANGE we keep all our DoFs nodal.

PR #3619, opened 2023-08-15
Delete dangerous implicit conversions for should_p_refine

lindsayad:
I immediately shot myself in the foot with my own code per https://github.com/idaholab/moose/pull/25108#discussion_r1294127650. Let's not let myself nor anyone else do that in the future

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -1621,6 +1621,12 @@ public:
    */
   bool should_p_refine(unsigned int g) const;
 
+  // Prevent bad user implicit conversions
+  void should_p_refine(FEFamily, bool) = delete;
+  void should_p_refine(Order, bool) = delete;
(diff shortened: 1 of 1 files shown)


roystgnr (approved):
Should we just turn things like FEFamily into scoped enums?  I suspect if C++11 had been out when we defined all our enums only Order would have been allowed to implicitly convert.

roystgnr:
I'm not seeing CI kicking on here?

lindsayad:
> Should we just turn things like FEFamily into scoped enums? I suspect if C++11 had been out when we defined all our enums only Order would have been allowed to implicitly convert.

That would probably be a good idea

> I'm not seeing CI kicking on here?

Yea GitHub is having problems with webhooks again

lindsayad:
failure unrelated

PR #3621, opened 2023-08-19
Fixes for EIM in the case of no parameters

dknez:
The recent PR related to EIM caused some errors in the case where we don't have any parameters. This fixes those errors.


jwpeterson (approved):
The "Test MOOSE mac" failures 
```
    FAIL TestHarness/tests.dry_run FAILED (CRASH)
```
are unrelated to this change.

PR #3624, opened 2023-08-21
Add MeshRefinement::_allow_unrefined_patches flag and accessor

jwpeterson:
This flag defaults to false in order to maintain the original behavior of the code, which was to always eliminate unrefined element patches. If you set it to true, then you can allow such patches to persist, which may be useful in e.g. 1D meshes where having unrefined patches does not introduce additional hanging nodes.

### DIFF
--- a/include/mesh/mesh_refinement.h
+++ b/include/mesh/mesh_refinement.h
@@ -480,6 +480,16 @@ public:
    */
   signed char & underrefined_boundary_limit();
 
+  /**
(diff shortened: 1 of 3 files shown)


jwpeterson:
I discussed this briefly with @roystgnr over IM yesterday and it's a relatively small/straightforward change that fixes an issue we had in our internal testing, so I'm going to go ahead and merge.

PR #3625, opened 2023-08-21
Add absolute tolerance

### LINKED ISSUE #3623: Adding an absolute tolerance for the `LinearSolver`-s
I encountered a problem where I would like to set an absolute tolerance for `PetscLinearSolver` in MOOSE. 
Right now, the solve function takes a relative tolerance only:

```
std::pair< unsigned int, Real > libMesh::PetscLinearSolver< T >::solve( SparseMatrix< T > &  matrix,
		SparseMatrix< T > &  preconditioner,
		NumericVector< T > &  solution,
		NumericVector< T > &  rhs,
		const double  tol,
		const unsigned int  m_its 
	) 	
```

and assigns the default petsc option for the absolute tolerance:

```
// Set the tolerances for the iterative solver.  Use the user-supplied
// tolerance for the relative residual & leave the others at default values.
ierr = KSPSetTolerances (_ksp, tol, PETSC_DEFAULT,
                            PETSC_DEFAULT, max_its);
LIBMESH_CHKERR(ierr);
```

I am thinking about either of these two:
1. Adding a new parameter to the `solve` function with a default value of 0 or a small number supported by all derived classes.
2. Creating a new virtual function which has both toleraces and errors for derived classes if the one with the abolute tolerance is not implemented. 

What do you think @lindsayad , @roystgnr , @jwpeterson ?

grmnptr:
Closes #3623


jwpeterson (commented):
It looks like you are adding a bunch of new `solve` overrides, but as I mentioned previously there are already so many `solve` overrides (7 if you include adjoint_solve) that this will likely result in a very large changeset if you want to support all the different ways of solving. What I suggest is instead to add a new API for specifying/attaching the `LinearSolverControl` object to a LinearSolver object. Then change the existing LinearSolver::solve APIs so that the tolerance and max iterations become optional parameters, i.e.

```
const std::optional<double> tol = std::nullopt,
const std::optional<unsigned int> max_its = std::nullopt,
```

Hopefully, this will be a backwards-compatible change. Then, within the derived-class solve() overrides, you can check whether the user actually provided a `tol` and `max_its` (old approach) and if not, instead use the values from the attached `LinearSolverControl` object (new approach).

grmnptr:
Ahhh okay, I think I completely misinterpreted the message. Thanks for the clarification, I am on it.

include/solvers/eigen_sparse_linear_solver.h
```diff
@@ -75,8 +75,8 @@ class EigenSparseLinearSolver : public LinearSolver<T>
   solve (SparseMatrix<T> & matrix,
          NumericVector<T> & solution,
          NumericVector<T> & rhs,
-         const double tol,
-         const unsigned int m_its) override;
+         const std::optional<double> tol = std::nullopt,
```
roystgnr, line 78: I always struggled to see a point to `std::optional<T>`, vs just using a `T*` and interpreting `nullptr` as the null option ... but this sort of backwards API compatibility is really cool.

src/solvers/petsc_linear_solver.C
```diff
@@ -356,28 +356,36 @@ PetscLinearSolver<T>::solve (SparseMatrix<T> &  matrix_in,
...
                              NumericVector<T> & rhs_in,
-                             const double tol,
-                             const unsigned int m_its)
+                             const std::optional<double> tol,
+                             const std::optional<unsigned int> m_its)
 {
   LOG_SCOPE("solve()", "PetscLinearSolver");
 
+  const double rel_tol = this->get_real_solver_setting("rel_tol", tol);
+  const double abs_tol = this->get_real_solver_setting("abs_tol", std::nullopt, (double)PETSC_DEFAULT);
```
jwpeterson, line 365: I'd prefer `static_cast<Real>(PETSC_DEFAULT)` assuming that works instead of the C-style cast with `double`.

src/solvers/petsc_linear_solver.C
```diff
@@ -356,28 +356,36 @@ PetscLinearSolver<T>::solve (SparseMatrix<T> &  matrix_in,
...
                                      NumericVector<T> & rhs_in,
-                                     const double tol,
-                                     const unsigned int m_its)
+                                     const std::optional<double> tol,
+                                     const std::optional<unsigned int> m_its)
 {
   LOG_SCOPE("adjoint_solve()", "PetscLinearSolver");
 
+  const double rel_tol = this->get_real_solver_setting("rel_tol", tol);
+  const double abs_tol = this->get_real_solver_setting("abs_tol", std::nullopt, (double)PETSC_DEFAULT);
```
jwpeterson, line 383: Same comment about C-style cast.

jwpeterson (approved):
This looks good to me, I like the way that you used the existing `_solver_configuration` stuff rather than create a completely new class. Assuming @roystgnr doesn't have any changes he wants to see and that this works for your MOOSE application, I'd say we're good to merge.

src/solvers/petsc_linear_solver.C
roystgnr: I'm cool with either `static_cast` or with a construction rather than a cast ... but we should really figure out whether we need this all to be `double` or `Real`.  I thought PETSc still uses `double` for tolerances even when the underlying `PetscReal` has been switched to `float` or `__float128`, so we'd want `double` here ... but if I'm wrong then I'd really worry about whether we can convert `PETSC_DEFAULT` from one of these to another without inadvertently trashing it.

jwpeterson: [KSPSetTolerances](https://petsc.org/release/manualpages/KSP/KSPSetTolerances/) takes `PetscReal`. My assumption was that we make our `Real` match `PetscReal`, but I don't know exactly how you handled this.

roystgnr: Okay, I was wrong: it's `PetscReal` for everything in https://petsc.org/release/manualpages/KSP/KSPSetTolerances/

So we want to be careful to use `Real` for anything that we can't be certain will be happy about a conversion, like `PETSC_DEFAULT`.

roystgnr: For loose definitions of "match"; IIRC they use pure `__float128` in the quad precision case, while we use `boost::multiprecision::float128` to get some `std::` function overloads from there.

roystgnr: Anyway, once the `PETSC_DEFAULT` stuff is changed from `double` to `Real`, I'm happy with this.

grmnptr:
Got it, thanks. I added the modifications. Should we rerun the tests?

src/solvers/petsc_linear_solver.C
grmnptr: Ahh okay, github does not load new conversations autoamtically. I'll make it Real.

grmnptr:
Accoding to @loganharbour , we don't test mingcc on MOOSE anymore and he will remove this test from libmesh CI too.

jwpeterson:
> Should we rerun the tests?

Looks like they are running now. I'd be OK with not waiting for everything to finish before merging. Since you changed the 'LaspackLinearSolver`, it might be good to throw in a `--disable-strict-lgpl` test, but if not I think that runs on the devel -> master fast forward merge anyway. My only other question was if you tested the Trilinos changes? They look straightforward enough, so I don't think it's absolutely necessary.

grmnptr:
I didn't test the Trilinos changes. Can we test them on this PR somehow?

jwpeterson:
> I didn't test the Trilinos changes.

A long time ago there was a Trilinos Civet recipe that we could attach to PRs, but I don't see it now.

roystgnr:
We used to have a Trilinos Civet recipe?  I though I was the only one testing Trilinos configurations, via BuildBot.

roystgnr (approved):

jwpeterson:
> We used to have a Trilinos Civet recipe?

Yeah, looks like it has been [inactive](https://github.com/idaholab/civet_recipes/blob/48091b1410b25e466a3056167f75c9087a332997/inactive_recipes/moosebuild/package_builder/Trilinos_mpich_gcc_petsc_default_vtk.cfg) for at least three years, though.

Thanks @grmnptr for the update!

PR #3630, opened 2023-08-31
Extend add_vector to support upgrades on re-add

roystgnr:
This way we can have multiple code components asking for the same vector and they all get a superset of the capabilities they ask for.

### DIFF
--- a/src/systems/system.C
+++ b/src/systems/system.C
@@ -732,7 +732,58 @@ NumericVector<Number> & System::add_vector (std::string_view vec_name,
   // Return the vector if it is already there.
   auto it = this->_vectors.find(vec_name);
   if (it != this->_vectors.end())
-    return *it->second;
+    {
+      // If the projection setting has *upgraded*, change it.
+      if (projections) // only do expensive lookup if needed
+        libmesh_map_find(_vector_projections, vec_name) = projections;
+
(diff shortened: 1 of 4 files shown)


lindsayad (commented):
Nice!

lindsayad (approved):

src/systems/system.C
```diff
@@ -733,13 +733,64 @@ NumericVector<Number> & System::add_vector (std::string_view vec_name,
   // Return the vector if it is already there.
   auto it = this->_vectors.find(vec_name);
   if (it != this->_vectors.end())
-    return *it->second;
+    {
+      // If the projection setting has *upgraded*, change it.
```
jwpeterson, line 737: We discussed marking `System::add_vector` parallel_only, right?

jwpeterson (approved):
I can't remember what lead to us having separate storage for the vector types even though the types are stored on the vectors themselves, but that definitely seems good to be rid of that :+1:

src/systems/system.C
roystgnr: Yeah; sorry it slipped my mind.

Looking through, there's quite a few methods of similar type that ought to be marked parallel_only(); I'll make a separate PR for the lot.

PR #3631, opened 2023-09-01
Elem refinement unit tests, Inf FE fixes, mesh iterator refactor + more ranges

roystgnr:
Some results from downstream had me worried about inadequate test coverage for refinement of Tri7 and derivatives, so I added it here.

I then became frustrated that the new tests wouldn't compile because `local_element_ptr_range()` wasn't actually a thing, so I refactored all our iterator declarations to use macros for brevity and to give a range() option for every existing iterator type.

I cleaned up a couple more minor things along the way, but more importantly I caught a ton of old bugs in infinite-Elem implementations of inadequately-tested-with-infinite-Elem virtual functions.


roystgnr:
Bah.  This has the same "breaks on Pyramids with --disable-exceptions" bug that my previous unit testing PR did.  I'll add a fix for that after I see whether the current PR CI is happy this time.

roystgnr:
Okay, this ought to be solid (including "Test No Exceptions").  I'll merge once CI finishes agreeing.

PR #3640, opened 2023-09-11
Respect Abaqus node and element numbers a little

roystgnr:
We're not going to respect them enough to not convert from 1-based to 0-based, of course[0], but we can at least stick to just that conversion and not give them whatever (sequential for nodes and replicated meshes, sequential with spacing for distributed meshes) ids we choose instead.

[0] Dijkstra, E.W. "Why numbering should start at zero"

This is my crack at making life easier for @Traiwit in https://github.com/idaholab/moose/discussions/25440 ... who I'm really hoping can help test it, because it looks like we have *zero* test coverage of AbaqusIO right now?

### DIFF
--- a/include/mesh/abaqus_io.h
+++ b/include/mesh/abaqus_io.h
@@ -212,23 +212,6 @@ private:
   std::set<ElemType> _elem_types;
 
   /**
-   * Map from libmesh element number -> abaqus element number,
-   * and the converse.
-   */
-  // std::map<dof_id_type, dof_id_type> _libmesh_to_abaqus_elem_mapping;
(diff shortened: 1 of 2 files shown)


lindsayad (approved):

roystgnr:
> it looks like we have zero test coverage of AbaqusIO right now?

Well at least this wasn't true.

roystgnr:
I'd used `node_ptr` where I meant `query_node_ptr`; fixing that the tests pass for me.  Fingers crossed for CI.

roystgnr:
Going to merge this, lest I forget to do so before the next submodule update.  It's at least an improvement in the few tests we do have; if we hit regressions elsewhere we can fix them then.

PR #3642, opened 2023-09-13
Optimise RT orientation code dealing w/ cyclic permutations

nmnobre:
Hi @roystgnr,

So my question regarding continuous deformations had a second intent behind: to confirm if we did need a tiny piece of code I put in to deal with the fact that cyclic permutations are not even for quad faces, just to make sure the node we start the sequence at does not matter. I'm now convinced we should have it, so I've made sure it isn't used for simplices and got rid of the inefficient while loop. I've also switched from dynamic to static arrays to make sure we aren't leaving any performance on the table. Semantically, this PR changes nothing.

Cheers,
Nuno

### DIFF
--- a/examples/vector_fe/vector_fe_ex6/vector_fe_ex6.C
+++ b/examples/vector_fe/vector_fe_ex6/vector_fe_ex6.C
@@ -31,9 +31,10 @@
 // The solver packages supported by libMesh.
 #include "libmesh/enum_solver_package.h"
 
-// The mesh object and mesh generation utilities.
+// The mesh object and mesh generation and modification utilities.
 #include "libmesh/mesh.h"
 #include "libmesh/mesh_generation.h"
+#include "libmesh/mesh_modification.h"
 
 // Matrix and vector types.
 #include "libmesh/dense_matrix.h"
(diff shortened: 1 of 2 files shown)


roystgnr:
Just to be safe, would you try adding a `MeshModification::permute_elements()` call to the RT example, and make sure it doesn't effect any results?  I should have thought of doing that with the initial PR...

roystgnr (approved):
Looks good to me; fingers crossed that CI agrees.

nmnobre:
> Just to be safe, would you try adding a `MeshModification::permute_elements()` call to the RT example, and make sure it doesn't effect any results?

What does this do exactly?

roystgnr:
`permute_elements()` calls `Elem::permute()` to act randomly on every element, and the subclasses' overloads of that enable all the valid permutations on each element type.  So for a Hex you could think of the permutation as picking one of 6 random sides, then one of 4 random rotations keeping that side in place, for all the 24 valid ways you can get the same physical geometry from a different local node ordering.

Do that on a decent-sized mesh and you're practically guaranteed to catch any orientation-based bugs.  We've got it in our unit test for scalar elements, but I didn't think about applying it to the vector elements before now.  It might be smart for us to use it in one or two Nedelec vector_fe examples too...

roystgnr:
Oh, and yes, I'll be on to talk to you at 8:30 my time tomorrow.

nmnobre:
> `permute_elements()` calls `Elem::permute()` to act randomly on every element, and the subclasses' overloads of that enable all the valid permutations on each element type. So for a Hex you could think of the permutation as picking one of 6 random sides, then one of 4 random rotations keeping that side in place, for all the 24 valid ways you can get the same physical geometry from a different local node ordering.

That's perfect, very, very good shout (as always).
I did test commenting out the line that 'resets' the cyclic permutations and it did fail to produce the expected solution, so it's  needed, as expected! :)

PR #3645, opened 2023-09-19
Remove fortran/cxx hdf5 detection/linkage

### LINKED ISSUE #3644: We can detect different HDF5 C and CXX libraries
An example of how this can easily happen: PETSc paths are ahead of system paths for inclusion/linking and PETSc has built HDF5 without CXX. Meanwhile there is a system installation of HDF5 that *does* have CXX. This happened in https://github.com/idaholab/moose/discussions/25506

lindsayad:
We don't use it anywhere in libMesh. We also don't use it in MOOSE and @jwpeterson has indicated he doesn't use it either

Closes #3644


roystgnr (approved):
I'm certainly cool without this if John is.

jwpeterson:
The failing tests seem to be a Python issue unrelated to this change.
```
======================================================================
ERROR: test (test.TestAverageStencil)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/civet0/build/moose/test/tests/fvkernels/mms/skewness-correction/adv-diff-react/test.py", line 9, in test
    fig = mms.ConvergencePlot(xlabel='Element Size ($h$)', ylabel='$L_2$ Error')
  File "/opt/civet0/build/moose/python/mms/ConvergencePlot.py", line 44, in __init__
    tick.label.set_fontsize(fontsize)
AttributeError: 'XTick' object has no attribute 'label'
```

jwpeterson (approved):

PR #3650, opened 2023-09-28
Added RBParametrizedFunction::preevaluate_parametrized_function_cleanup()

dknez:
Added RBParametrizedFunction::preevaluate_parametrized_function_cleanup(), which is a no-op by default. It can be overridden in subclasses in order to do relevant cleanup during EIM training, such as clearing cached data.


jwpeterson (approved):
We have some internal tests that will use this capability.

PR #3651, opened 2023-09-28
DO NOT MERGE - test revert to master

roystgnr:
This sets us back to the exact same code as in 52f92e583f, so I can add the `Test (no threads)` recipe that's been failing since https://civet.inl.gov/job/1786204/ and double-check that it's something in CI that's changed rather than something in libMesh.  The first failing change in libMesh was to a *comment*, so I'm not expecting a pass here, but I'm also not finding any changes to the CI environment that could explain the failures, so here we are.


roystgnr:
Figured out the problem; a fix is coming via https://github.com/idaholab/moose/pull/25647

PR #3655, opened 2023-10-03
Improve the description of centroid nodes

nmnobre:


jwpeterson (commented):
Looks like the "equidistant from" terminology was originally introduced in the TET14 update (3d44cbfcd446b5d40453fd59ea6529db0d4541c9) and then got copied from there to other element docs. Change looks fine to me but it's @roystgnr doc so I'll let him approve.

jwpeterson:
BTW the [TRI7 docs](
https://github.com/libMesh/libmesh/blob/ca8cf871b45ce438a3a8032ff3701cef04caab89/include/geom/face_tri7.h) don't mention the centroid node separately, so might be good to add a comment there for consistency.

nmnobre:
> Looks like the "equidistant from" terminology was originally introduced in the TET14 update ([3d44cbf](https://github.com/libMesh/libmesh/commit/3d44cbfcd446b5d40453fd59ea6529db0d4541c9)) and then got copied from there to other element docs. Change looks fine to me but it's @roystgnr doc so I'll let him approve.

Thanks John. The wording "equidistant from" just sounded weird to me, but let's wait for Roy as you suggest. :-)

nmnobre:
> BTW the [TRI7 docs](https://github.com/libMesh/libmesh/blob/ca8cf871b45ce438a3a8032ff3701cef04caab89/include/geom/face_tri7.h) don't mention the centroid node separately, so might be good to add a comment there for consistency.

But in that case the diagram does depict the node, so I don't think there's a need for the extra comment.

roystgnr (approved):
Of course "equidistant" is nonsense in general ... and it's not even true on our master element, so I'm not sure what I was thinking.  It might have been true on the equilateral triangles I was scribbling on scratch paper while trying to figure out the code here?

nmnobre:
> Of course "equidistant" is nonsense in general ... and it's not even true on our master element, so I'm not sure what I was thinking. It might have been true on the equilateral triangles I was scribbling on scratch paper while trying to figure out the code here?

Yeah, for equilateral triangles, the circumcenter does indeed match the centroid.
The centroid is preserved by affine maps as those between the reference and physical elements, so centroid nodes on the reference element will still be centroid nodes on the physical element.

PR #3657, opened 2023-10-12
Use a relative tolerance in norm-of-diff test

roystgnr:
On enough processors we exceed the absolute tolerance here with the newest PETSc.

Hopefully this fixes the failures @loganharbour has been seeing.


roystgnr:
Or it might just catch the Wasp switchover breaking all our recipes.  That's good too.

roystgnr:
Going to reopen this to make sure it hits the new Civet recipes.

PR #3658, opened 2023-10-13
Use a relative tolerance in norm-of-diff test

roystgnr:
On enough processors we exceed the absolute tolerance here with the newest PETSc.


roystgnr:
And of course Civet is smart enough to recognize that my new PR has the same old git hash.

roystgnr:
Going to merge; that canceled test made it to MOOSE modules and there's no way a cppunit test change affects anything there or beyond.

PR #3661, opened 2023-10-17
Fix missing header, refactor with Threads::DisablePerfLogInScope

roystgnr:
In poly2tri_triangulator.C we were getting libmesh_logging.h indirectly from threads_foo.h of all places, in a way that broke if neither of TBB and pthreads were available.  See #3659.

After fixing that, I figured I'd avoid the problem happening again, by getting that header out of the threads headers and into threads.C ... which turns out to be a cleaner and more exception-safe solution too.

We should wait on merging this until Logan's done with the CI backend transition.

### DIFF
--- a/include/parallel/threads.h
+++ b/include/parallel/threads.h
@@ -70,8 +70,8 @@ class DisablePerfLogInScope
 {
 public:
 #ifndef LIBMESH_ENABLE_PERFORMANCE_LOGGING
-  DisablePerfLogInScope() {}
-  ~DisablePerfLogInScope() {}
+  DisablePerfLogInScope() = default;
+  ~DisablePerfLogInScope() = default;
 #else
   DisablePerfLogInScope();
   ~DisablePerfLogInScope();
--- a/include/parallel/threads.h
+++ b/include/parallel/threads.h
@@ -61,6 +61,25 @@ private:
   bool & _b;
 };
 
+/**
+ * We use a class to turn perf logging off and on within threads, to
+ * be exception-safe and to avoid forcing indirect inclusion of
+ * libmesh_logging.h everywhere.
+ */
+class DisablePerfLogInScope
+{
+public:
+#ifndef LIBMESH_ENABLE_PERFORMANCE_LOGGING
+  DisablePerfLogInScope() {}
+  ~DisablePerfLogInScope() {}
+#else
+  DisablePerfLogInScope();
+  ~DisablePerfLogInScope();
+private:
+  const bool _logging_was_enabled;
+#endif
+};
+
 
 /**
  * Simple compatibility class for std::thread 'concurrent' execution.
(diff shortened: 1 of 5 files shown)


include/parallel/threads.h
```diff
@@ -61,6 +61,25 @@ class BoolAcquire
...
+ * We use a class to turn perf logging off and on within threads, to
+ * be exception-safe and to avoid forcing indirect inclusion of
+ * libmesh_logging.h everywhere.
+ */
+class DisablePerfLogInScope
+{
+public:
+#ifndef LIBMESH_ENABLE_PERFORMANCE_LOGGING
+  DisablePerfLogInScope() {}
+  ~DisablePerfLogInScope() {}
```
jwpeterson, line 74: `= default;` for the "non logging" implementations rather than empty brackets? 

The [reasoning](https://stackoverflow.com/questions/13576055/how-is-default-different-from-for-default-constructor-and-destructor) is related to whether the compiler can consider it a "trivial" class or not.

roystgnr: Oh, man, I'd forgotten about that.  Will fix.

roystgnr:
"Unused variable"?  Oh, of course.  If we build with --disable-perflog then the constructor and destructor do nothing, and g++ is smart enough to recognize that.  Clever.

roystgnr:
Pinging @loganharbour in case he wants to look at the coverage diff.  A bunch of new misses, but all on `continue;` keywords or `[this]` lambda captures, so it looks to me like the new Civet environment is just optimizing a little differently, not missing any configuration details that would eat into our test coverage.

loganharbour:
> Pinging @loganharbour in case he wants to look at the coverage diff. A bunch of new misses, but all on `continue;` keywords or `[this]` lambda captures, so it looks to me like the new Civet environment is just optimizing a little differently, not missing any configuration details that would eat into our test coverage.

Yeah, this looks fine to me. I thought that the new base would already be updated, but looks like this one has a different base. Good to go

roystgnr:
Thanks!  I'll wait for John to have a chance to review if he wants; last I talked to him I was still pushing "let's not merge anything until the Civet updates are done".

In fact ... do we have any Civet coverage using TBB?  I used to use that heavily, but looking at my own local builds, the newest thing I can find that isn't instead relying on the pthread backend is from 2021... and looking at the LCOV results here I see `threads_pthread.h` but not `threads_tbb.h`, so that's a no.  I'll see if I can test this locally, explicitly with TBB, and make sure I haven't broken that.

jwpeterson (approved):
Sorry, meant to approve this when I commented before.

I think this is the first time we've used the `[[maybe_unused]]` attribute, so ideally we should have a configure test for that, even though it *should* be supported by all C++17 compliant compilers.

roystgnr:
> ideally we should have a configure test for that

I think we do!  It's in the test body used by ax_cxx_compile_stdcxx.m4

I'm not sure how to *test* the test, though.  Wrap a compliant C++17 compiler in a shell script that first greps for maybe_unused and dies if it finds it?

jwpeterson:
> I think we do! It's in the test body used by ax_cxx_compile_stdcxx.m4

OK, thanks for pointing that out. I didn't find it with a cursory search since it's actually in the autoconf-submodule and not libmesh's m4 directory. That's definitely good enough for me.

PR #3663, opened 2023-10-19
Autotools update in dependencies

roystgnr:
This gets MetaPhysicL, TIMPI, and the older NetCDF, all of which still had pre-generated files with the older autotools.

Hopefully *now* this is enough to get users building on M1/M2 without having to autoreconf themselves first.


dschwen:
Tested on my M2 laptop and configures and builds without an autoreconf 🎉

PR #3667, opened 2023-10-23
Fix missing libmesh_logging.h includes

jwpeterson:
These were not caught by CI because we don't have any CIVET recipes that define `LIBMESH_HAVE_CAPNPROTO`.


roystgnr (approved):

roystgnr:
Nor any builds of mine, apparently.  Thanks for the fix.

PR #3671, opened 2023-10-23
Test for *new* autotools when deciding to build it

roystgnr:
Otherwise we let older autotools slip by and then the next thing you know Alex's PR is all messed up.


lindsayad (approved):
Great, thanks!

jwpeterson (approved):

PR #3674, opened 2023-10-25
Compute L2 error for the scalar field in Vector FE ex6

nmnobre:
I was curious about what Alex (@lindsayad) noted yesterday in #3638 and couldn't resist trying to reproduce it.
My observations are exactly the same, and upping the quadrature order for the error computation does bring back the linear trend (phew.....).

I've done this the lazy way, i.e., I'm just creating new `ExactSolution` and `FunctionBase` objects... It seems that if I reverse the order in which I add the vector and scalar variables to the system, add thus the way in which I assemble the system as well, that I could then use the same `ExactSolution` and `FunctionBase` objects we already have... But what if I had more than one vector variable in the system? If I understand the following correctly,

https://github.com/libMesh/libmesh/blob/edc105866bac6e574ddc182352fcd032bda19b0e/include/numerics/function_base.h#L50-L53

I can only have one vector variable, and it must be the last one to be added to the system?
I'll update the html file with the new outputs if you're both happy with this. If Alex already has a better implementation, I'm happy to drop this as well, no worries. :)

Cheers,
-Nuno

### DIFF
--- a/src/error_estimation/exact_solution.C
+++ b/src/error_estimation/exact_solution.C
@@ -469,12 +469,19 @@ void ExactSolution::_compute_error(std::string_view sys_name,
     _equation_systems_fine->get_system(sys_name) :
     _equation_systems.get_system (sys_name);
 
+  const MeshBase & mesh = computed_system.get_mesh();
+
   const Real time = _equation_systems.get_system(sys_name).time;
 
   const unsigned int sys_num = computed_system.number();
   const unsigned int var = computed_system.variable_number(unknown_name);
-  const unsigned int var_component =
-    computed_system.variable_scalar_number(var, 0);
+  unsigned int var_component = 0;
+  for (const auto var_num : make_range(var))
+  {
+    const auto & var_fe_type = computed_system.variable_type(var_num);
+    const auto var_vec_dim = FEInterface::n_vec_dim(mesh, var_fe_type);
+    var_component += var_vec_dim;
+  }
 
   // Prepare a global solution, a serialized mesh, and a MeshFunction
   // of the coarse system, if we need them for fine-system integration
@@ -517,8 +524,6 @@ void ExactSolution::_compute_error(std::string_view sys_name,
 
   // Get a reference to the dofmap and mesh for that system
   const DofMap & computed_dof_map = computed_system.get_dof_map();
-
-  const MeshBase & mesh = computed_system.get_mesh();
 
   // Grab which element dimensions are present in the mesh
   const std::set<unsigned char> & elem_dims = mesh.elem_dimensions();
--- a/doc/html/examples/vector_fe_ex6.html
+++ b/doc/html/examples/vector_fe_ex6.html
@@ -88,9 +88,12 @@
     DofMap Constraints
       Number of DoF Constraints = 0
 
-L2 error is: 0.134332
-HDiv semi-norm error is: 0.344122
-HDiv error is: 0.369411
+~~ Vector field (u) ~~
+L2 error is: 0.134373
+HDiv semi-norm error is: 0.344042
+HDiv error is: 0.369353
+~~ Scalar field (p) ~~
+L2 error is: 0.0697862
 
 ***************************************************************
 * Done Running Example vector_fe_ex6:
@@ -141,9 +144,12 @@ HDiv error is: 0.369411
     DofMap Constraints
       Number of DoF Constraints = 0
 
-L2 error is: 0.134332
-HDiv semi-norm error is: 0.344122
-HDiv error is: 0.369411
+~~ Vector field (u) ~~
+L2 error is: 0.134373
+HDiv semi-norm error is: 0.344042
+HDiv error is: 0.369353
+~~ Scalar field (p) ~~
+L2 error is: 0.0697862
 
 ***************************************************************
 * Done Running Example vector_fe_ex6:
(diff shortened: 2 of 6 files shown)


jwpeterson:
> I can only have one vector variable, and it must be the last one to be added to the system?

Looks like that comment was added by @pbauman in 9d05132d42af46c1efda2130b27c374b632d9e02 when he was starting to add the support for vector-valued FE, so he might be able to comment further.

roystgnr:
> I can only have one vector variable, and it must be the last one to be added to the system?

In theory, no; if you then added a v_var at index 4, its component indices in 3D would be expected to be 6, 7, and 8.

In practice?  I don't think we have any test coverage here, which means we've almost certainly got bugs here.

nmnobre:
> In theory, no; if you then added a v_var at index 4, its component indices in 3D would be expected to be 6, 7, and 8.

Oh... that's not what that comment says though? What you're telling me is that there is code to avoid indexing collisions...
And how does that work in practice to address the `DenseVector` storing the solution? Each variable needs to know the amount and extent of previous variables? Maybe we could just index variables at multiples of 3 by default and that would avoid collisions, at the expense of indexing gaps (if there are scalar variables or the problem is < 3D).

roystgnr:
> Oh... that's not what that comment says though?

The comment is talking about a hypothetical case with one vector-valued variable, right?  It doesn't say either-way what happens with two, but I think "components are indexed contiguously" is the intended meaning, not "everything breaks".

> What you're telling me is that there is code to avoid indexing collisions...

In theory, everything working with a `FunctionBase` whose output is supposed to match a system's indexing should be using something like `Variable::first_scalar_number()` to figure out what component index is the first for a particular variable.  In practice I'm sure there's still lots of code that will scream `libmesh_not_implemented()` if it sees a vector-valued variable, or worse fail to handle it properly at all.

The trouble is that there's nothing in the API *mandating* that that indexing be done correctly; in fact `FunctionBase` isn't actually restricted to use cases where indexing matches system variables at all! 
 It's perfectly valid to write code like e.g. `TriangulatorInterface` where we get a single scalar-valued desired-area function via a `FunctionBase`, in a context where a `System` may not even exist.

> Maybe we could just index variables at multiples of 3 by default

We "could have indexed" this way: it would have worked for vector-valued variables, and then broken down the line for rank-2+ tensor-valued variables.

But "could index" suggests it's a change that we could make now, and no, that would break every code using FunctionBase everywhere.

lindsayad:
Man we're colliding on work here. I will put up a PR to show you what I did. I modified `ExactSolution` to properly handle vector variables

nmnobre:
> Man we're colliding on work here. I will put up a PR to show you what I did. I modified `ExactSolution` to properly handle vector variables

No worries, I'm happy to drop this if need be. :)

lindsayad (approved):
This is great. Thanks @nmnobre !

nmnobre:
> This is great. Thanks @nmnobre !

Sorry for the overlap, I just panicked when you said super-convergence and couldn't help myself. I'll try to be more patient next time. Please accept my apologies.

> I modified ExactSolution to properly handle vector variables

Thanks again for this, this is what I was asking @roystgnr about, so it couldn't have been more timely! :)

> The comment is talking about a hypothetical case with one vector-valued variable, right?

Not super clear to me, I'll try to amend the comment so it makes more sense to me... Let me know if it makes sense to you.

> In theory, everything working with a FunctionBase whose output is supposed to match a system's indexing should be using something like Variable::first_scalar_number() to figure out what component index is the first for a particular variable.

Okay, so can we use that here to avoid hard-coding the indices?

lindsayad:
I do think that @roystgnr is going to shortly up a PR that will automatically add the extra quadrature during error calculations

lindsayad:
> > In theory, everything working with a FunctionBase whose output is supposed to match a system's indexing should be using something like Variable::first_scalar_number() to figure out what component index is the first for a particular variable.
> 
> Okay, so can we use that here to avoid hard-coding the indices?

@roystgnr this is what we talked about yesterday on Slack. `first_scalar_number()` is not going to be correct for any variable coming after a vector variable, which is what necessitates things like the workaround in [28e3e89](https://github.com/libMesh/libmesh/pull/3674/commits/28e3e894997a91a4e2019a55c6b0759c949a2b0a)

lindsayad:
BTW if you're interested @nmnobre I would add you to our slack space (would just need the email address you'd want to have associated with it)

nmnobre:
> I do think that @roystgnr is going to shortly up a PR that will automatically add the extra quadrature during error calculations

Okay, there are other examples that use this mechanism of asking for the order from an input file (the ones I copied this from haha)... maybe when he creates that PR he'd be able to clean all such examples?

nmnobre:
> BTW if you're interested @nmnobre I would add you to our slack space (would just need the email address you'd want to have associated with it)

That'd be grand!
You can just use the one associated with my commits. :)

src/error_estimation/exact_solution.C
```diff
@@ -469,12 +469,19 @@ void ExactSolution::_compute_error(std::string_view sys_name,
...
+  const MeshBase & mesh = computed_system.get_mesh();
+
   const Real time = _equation_systems.get_system(sys_name).time;
 
   const unsigned int sys_num = computed_system.number();
   const unsigned int var = computed_system.variable_number(unknown_name);
-  const unsigned int var_component =
-    computed_system.variable_scalar_number(var, 0);
+  unsigned int var_component = 0;
+  for (const auto var_num : make_range(cast_int<decltype(var)>(0), var))
```
lindsayad, line 479:
```suggestion
  for (const auto var_num : make_range(var))
```
I don't know why I added the unnecessary `begin` argument 🙁

lindsayad:
@nmnobre I'm going to amend my commit with the suggested change and push

lindsayad:
Jk maintainers cannot push to organization repositories. Just realized you submitted this from farscape-project

roystgnr (approved):

PR #3676, opened 2023-10-25
Don't underintegrate with default ErrorEstimator settings

roystgnr:
We only triggered the problem here with ExactSolution, but there are a bunch of theoretically-susceptible classes, plus a probably-not-susceptible-but-lets-be-safe class.  This fixes them all.

I'm only changing the default setting here; anyone who set an extra quadrature order explicitly should still be getting what they asked for.


jwpeterson (approved):

lindsayad (approved):
beautiful code, beautiful math

lindsayad:
@nmnobre FYI

PR #3677, opened 2023-10-27
Get the proper rational weights for nodal_soln

roystgnr:
Good news: this way we get the same answer from visualizations on 32-bit and 64-bit, and it should be the correct answer to boot!

Bad news: I've had to disable the MOOSE IGA regression tests (at least the ones with non-trivial bex_weight meshes) to get this to pass, because neither answer was correct before.

Refs idaholab/moose#25819

### DIFF
--- a/src/fe/fe_rational.C
+++ b/src/fe/fe_rational.C
@@ -83,8 +83,11 @@ void rational_nodal_soln(const Elem * elem,
 
         std::vector<Real> node_weights(n_nodes);
 
+        const unsigned char weight_index = elem->mapping_data();
+
         for (unsigned int n=0; n<n_nodes; n++)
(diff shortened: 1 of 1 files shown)


jwpeterson (approved):
You had tests of this, so was the bug just not bad enough to make the solutions look "obviously" wrong?

roystgnr:
Yup.  The catches are that:

1. The weight just changes the ratios of how different "raw" basis functions get composed into each final rational basis function, so there's still a bound on how bad things can get.  I was seeing diffs of a fraction of a percent.

2. It's in code that only gets used for output file writes, so it doesn't even affect any internal error calculations.

3. Not all IGA meshes even use non-constant weights, so some test cases didn't even end up being wrong.

PR #3678, opened 2023-10-28
Streamline exact solution specification in Vector FE ex6

nmnobre:
Hey,

I decided to play a bit with #3676, and for that I defined a wavenumber-like variable to allow me to change the characteristics of the solution function easily. Other people might find this ability useful, this is for them.

Now, some examples read an input parameter, `extra_error_quadrature`, to adjust the quadrature order for error estimation purposes. From what I've seen, it's default value is often 2, which is *not* better than 1 (the new global default), and worse, can induce people in error I feel...

In libMesh, the quadrature order is the order of the highest polynomial which can be exactly integrated, not the number of points in the quadrature rule. As the default quadrature order is 2p+1, i.e. odd, adding an even number, m, as the extra order, is no better than just adding m-1. From what I've seen, all examples which take an input parameter to set the extra order use an even number as its default, and none of these examples need it anymore with the new global default anyway...

The new global default means you'll only need an extra order if you're using an order p basis to estimate a solution with order p+2 on each element, at which point you probably want a finer mesh anyway, so the option is now kinda irrelevant? (Which I think is awesome and a huge merit of #3676 btw!)

It might be useful to illustrate how you can change the error quadrature order, but none of our examples make useful use of that option atm, thoughts?

### DIFF
--- a/examples/vector_fe/vector_fe_ex6/div_grad_exact_solution.h
+++ b/examples/vector_fe/vector_fe_ex6/div_grad_exact_solution.h
@@ -31,52 +31,52 @@ public:
 
   Real scalar(Real x, Real y)
   {
-    return cos(.5*pi*x)*sin(.5*pi*y);
+    return cos(k*x)*sin(k*y);
   }
 
   Real scalar(Real x, Real y, Real z)
   {
-    return cos(.5*pi*x)*sin(.5*pi*y)*cos(.5*pi*z);
+    return cos(k*x)*sin(k*y)*cos(k*z);
   }
 
   RealGradient operator() (Real x, Real y)
   {
-    const Real ux =  .5*pi*sin(.5*pi*x)*sin(.5*pi*y);
-    const Real uy = -.5*pi*cos(.5*pi*x)*cos(.5*pi*y);
+    const Real ux =  k*sin(k*x)*sin(k*y);
+    const Real uy = -k*cos(k*x)*cos(k*y);
 
     return RealGradient(ux, uy);
   }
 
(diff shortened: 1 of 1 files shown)


examples/vector_fe/vector_fe_ex6/div_grad_exact_solution.h
```diff
@@ -25,58 +25,61 @@ using namespace libMesh;
 
 class DivGradExactSolution
 {
+private:
+  const Real k = .5*pi;
```
lindsayad, line 29: I prefer to read

```
public:
protected:
private:
```

lindsayad (changes requested):

lindsayad (approved):

nmnobre:
@lindsayad Any thoughts on my observations regarding the extra error quadrature order input parameter?

lindsayad:
> As the default quadrature order is 2p+1, i.e. odd, adding an even number, m, as the extra order, is no better than just adding m-1.

Is this statement true for all quadrature rules or just those of Gaussian type? I believe it may not be generally true but @jwpeterson and @roystgnr have been talking a lot about quadrature lately and their knowledge base is deeper than mine

roystgnr:
Even for Gaussian quadrature that statement doesn't hold on triangles, tets, and prisms.  (Probably on pyramids, too, but pyramids are so crazy in general that we don't even try to do anything fancy there)

nmnobre:
You're obviously correct, I was stupidly blindfolded by edges, quads and hexes. :)

PR #3682, opened 2023-10-30
More parallel synchronization assertions

roystgnr:
I'm not sure whether the hung jobs @YaqiWang has been seeing are actually due to loss of sync in libMesh/MOOSE code, rather than something lower-level, but it is a bit embarrassing that we're seeing a loss of sync in devel mode runs that should have been caught with an assertion and is instead manifesting as a hang.

### DIFF
--- a/src/numerics/petsc_vector.C
+++ b/src/numerics/petsc_vector.C
@@ -375,6 +375,9 @@ void PetscVector<T>::add (const T a_in, const NumericVector<T> & v_in)
 
   PetscErrorCode ierr = VecAXPY(_vec, a, v->vec());
   LIBMESH_CHKERR(ierr);
+
+  libmesh_assert(this->comm().verify(int(this->type())));
+
   if (this->type() == GHOSTED)
     VecGhostUpdateBeginEnd(this->comm(), _vec, INSERT_VALUES, SCATTER_FORWARD);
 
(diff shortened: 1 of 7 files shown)


src/numerics/petsc_vector.C
```diff
@@ -353,6 +375,9 @@ void PetscVector<T>::add (const T a_in, const NumericVector<T> & v_in)
 
   PetscErrorCode ierr = VecAXPY(_vec, a, v->vec());
   LIBMESH_CHKERR(ierr);
+
+  libmesh_assert(this->comm().verify(this->type()));
```
lindsayad, line 379: nice

roystgnr: Oh, but hell, I meant to put that in `add_vector()` too.  After looking at the MOOSE patch again I don't think this is the problem, but if it is let's find it at its root.

roystgnr:
This isn't going to work without a custom OpFunction, huh?  Okay, that's doable.

YaqiWang:
@roystgnr Please let me know when this is ready for a try.

roystgnr:
@YaqiWang - this should be in the MOOSE libMesh submodule now, if you want to try it out on a `METHOD=devel` run.  I wouldn't be hopeful that it to affect one of the stack trace sets you showed us (the PetscVector init), but the other (the FEProblemBase::outputStep) is what inspired this and I'd be very curious to know if you manage to trip any assertions.

roystgnr:
Well, "in the MOOSE libMesh submodule" in next, anyway.  Looks like it hasn't made it to devel yet.

PR #3684, opened 2023-10-31
Update Nanoflann to v1.5.0

### LINKED ISSUE #3683: Update Nanoflann to v1.5.0
The current Nanoflann version is almost a couple years old and there have been a few additions such as concurrent tree building which I want to use. There are a few API changes and bug fixes since v1.4.1 as well.

parikshitbajpai:
Closes #3683


roystgnr:
Thanks for this!  I'm not sure why our CI hasn't yet kicked in on it...

Nanoflann is currently on github, isn't it?  Would it be reasonable to switch over to using a git submodule, so long as we're updating?

parikshitbajpai:
> Thanks for this! I'm not sure why our CI hasn't yet kicked in on it...

I can `git commit --amend --no-edit` and see if that triggers the CI.

> Nanoflann is currently on github, isn't it? Would it be reasonable to switch over to using a git submodule, so long as we're updating?

I think adding it as a submodule would be the way to go. It'll make pulling future changes easier for sure but I'm not sure what we would do for files which we currently don't include in the contrib version such as their testing / gui examples, etc.

roystgnr:
I sent you an invite to the group that automatically gets CI run on PRs; that's not what went wrong before but it's still a good idea.  I've activated testing manually here.

I'm not sure what headaches using a submodule here would cause; I'd say just try it out and if you run into any trouble we'll just merge this instead?  Usually we'd have no problem with pulling extra testing/example codes.  Where things get tricky is when the official repository doesn't have quite the layout or build system or something that we need; see the poly2tri submodule for an example of that.

parikshitbajpai:
> I'm not sure what headaches using a submodule here would cause; I'd say just try it out and if you run into any trouble we'll just merge this instead? Usually we'd have no problem with pulling extra testing/example codes.

Let me try it out. I might need some figuring out the Makefile but I'll have a look at poly2tri first.

parikshitbajpai:
Closing this since going the submodule way in #3689.

PR #3685, opened 2023-10-31
Add parameter for a user supplied type hash

### LINKED ISSUE #3654: Recompile JIT code if the value type changes
When rebuilding MOOSE with different AD container sizes the cached JIT objects do not get invalidated. I had a PR for that two years ago, but had to revert it (#3014). It's high time to put in a proper fix as this is tormenting @lindsayad .

I will most likely have the value typeid contribute to the function hash.

dschwen:
Closes #3654

### DIFF
--- a/contrib/fparser/fparser_ad.cc
+++ b/contrib/fparser/fparser_ad.cc
@@ -454,8 +454,8 @@ int FunctionParserADBase<Value_t>::AutoDiff(const std::string& var_name)
     const std::string jitdir = ".jitcache";
     if (cached)
     {
-      // generate a hash of the Value type size, byte code, immediate list, and registered derivatives
-      auto h = JITCodeHash();
+      // generate a hash of the size, byte code, immediate list, and registered derivatives (derivative form is type independent)
+      auto h = JITCodeHash(0);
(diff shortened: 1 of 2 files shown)


dschwen:
MOOSE PR is here https://github.com/idaholab/moose/pull/25914 and will need a libmesh submodule update.

lindsayad (approved):

roystgnr:
This feels less flexible - before you could pass in any string, and if you needed a unique string you could append as much as you needed to get.  Now you have 2^64 options and if you need something more unique it seems you have to essentially have your own hash algorithm?

dschwen:
> before you could pass in any string,

No you couldn't pass any string. The string was directly used as the type in the generated C code for the function. 

I suppose you could pass something like `ADReal/*derivative_size=50*/` as a type...

roystgnr:
> The string was directly used as the type in the generated C code for the function.

Where's this happen?  All I'm seeing in `JITCodeHash` is a bunch of `libMesh::boostcopy::hash_combine` calls.

dschwen:
JITCodeHash is not supposed to be an external interface (I should check if it's accidentally public...). The string is used in the JIT compile helper method.

roystgnr (approved):

PR #3686, opened 2023-10-31
Vector example 7 follow-up

lindsayad:
- Add HTML
- Fix svd solve guard

### DIFF
--- a/doc/html/examples/vector_fe_ex7.html
+++ b/doc/html/examples/vector_fe_ex7.html
@@ -1,0 +1,175 @@
+<!doctype html>
+<html lang="en-US">
+<head>
+  <meta http-equiv="Content-Type" content="text/html;charset=utf-8">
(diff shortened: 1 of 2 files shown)


jwpeterson (approved):
Looks good to me, thanks for the html output for the example, I will update the libmesh webpage with it. 

As discussed over slack with @roystgnr, we may have a missing FE pre-request somewhere which is triggering the following "not currently being computed" warnings for me. Something to look into for a separate PR...
```
WARNING: Shape function gradients for HDiv elements are not currently being computed!./include/libmesh/hdiv_fe_transformation.h, line 88, compiled Oct 31 2023 at 13:52:00 ***
WARNING: Shape function curls for HDiv elements are not currently being computed!./include/libmesh/hdiv_fe_transformation.h, line 121, compiled Oct 31 2023 at 13:52:00 ***
```

roystgnr (approved):

PR #3688, opened 2023-11-01
Single Face Hybridizable Discontinuous Galerkin example

lindsayad:
For k = 1, we get k + 1 error in the L2 norm for both the primal solution and the flux, and k + 2 error in the L2 norm for a postprocessed primal field

### TRI6 
Screenshot 2023-11-01 at 4 21 44 PM


lindsayad:
### QUAD8/9

It's interesting that QUAD8/9 seems to have non-optimal convergence

image

### TET14

image

### HEX27

And HEX27 seems to be slightly non-optimal as well

image

lindsayad:
I think this is ready for review. In the paper from which this example is constructed, Cockburn's "A superconvergent ldg-hybridizable galerkin method for second-order elliptic problems", they only considered simplexes

lindsayad:
I do not understand some of the coverage report changes...

roystgnr (approved):
I'm still flabbergasted by the "just pick an arbitrary side to stabilize, whatever" method, but if it works it works.  And I'm thrilled to see `L2_LAGRANGE_VEC` go in.

roystgnr:
Oh, and to recap Slack discussion here; we're pretty sure coverage report diffs are due to this PR overlapping with changes in our CI compilers, not anything real.

jwpeterson:
> Oh, and to recap Slack discussion here;

My only comment regarding #3688 is that since it adds another example, it would be good if you could commit the make doc output in the doc/html/examples dir

roystgnr:
I should not have merged this yet.  Those failing recipes looked like more of the CI problems we'd been having, but they're legitimate failures with --enable-distmesh.  I'll figure out the fix.

roystgnr:
I'm actually not seeing how this isn't failing in parallel on replicated meshes too.  We're looping over `mesh.active_local_element_ptr_range()`, and in the `!global_solve` case we're calling `system.solution->close()` and `system.update()` for each?  How is this code not supposed to hang when we have different numbers of elements on different processors?

roystgnr:
It is failing in parallel on replicated meshes too.  I get `Different ranks are at different libmesh_parallel_only points` from a 2 processor run.  What I'm not seeing is how we don't get that in CI.

lindsayad:
Wow I don't know how that snuck inside of the element loop. It's correct for the matrix-based assembly routine

lindsayad:
> It's interesting that QUAD8/9 seems to have non-optimal convergence

I wonder if this is a consequence of the single-face implementation in this example. In https://github.com/idaholab/moose/pull/26477 I do not limit the penalty term to a single face and I am seeing optimal convergence with QUAD9

PR #3689, opened 2023-11-02
Switch to Nanoflann submodule

### LINKED ISSUE #3683: Update Nanoflann to v1.5.0

parikshitbajpai:
Closes #3683 
Supersedes #3684


roystgnr:
Oh, I like this.  Thanks!

Looks like you still need to add a couple more tweaks to Makefile.am to get our `make check` running their examples from the deeper-nested directory.

It looks like you're starting us with their current git head instead of 1.5.0.  That seems like a reasonable choice to me; just curious: was it a deliberate choice?

parikshitbajpai:
> Looks like you still need to add a couple more tweaks to Makefile.am to get our `make check` running their examples from the deeper-nested directory.

Yeah I just realized I had a few missing intermediate `nanoflann/`. Fixing them now

> It looks like you're starting us with their current git head instead of 1.5.0. That seems like a reasonable choice to me; just curious: was it a deliberate choice?

I noticed they had a missing argument `SearchParameters` in `knnSearch` which was fixed in fixes post v1.5.0. But having checked them again right now, it looks like they are failing the tests in nanoflann repo. I'll see if they cause our tests to fail or not and change it to v1.5.0 if I have to.

roystgnr:
Nanoflann uses threads now.  `std::atomic`, `std::async`, `std::mutex`, the works.  Some C++ compilers appear to automatically link to libpthread or whatever they need as the backend here, even when we don't pass them a flag to indicate that; other C++ compilers ... do not.

We ought to be able to fix this by passing our libMesh-detected flags to nanoflann `make check`.  I'll give that a try.

roystgnr:
Ah, *that* was the failure I was afraid I'd see.  MOOSE API calls that don't fit a changed API name.  I guess we can put in a PR there that uses one name or the other depending on `NANOFLANN_VERSION`

parikshitbajpai:
> Ah, _that_ was the failure I was afraid I'd see. MOOSE API calls that don't fit a changed API name. I guess we can put in a PR there that uses one name or the other depending on `NANOFLANN_VERSION`

I'm thinking I can try
```cpp
#include "libmesh/nanoflann.hpp"
#include "libmesh/point.h"

#ifdef NANOFLANN_VERSION == 0x150
  \\ Syntax1
#else
  \\ Syntax2
#endif
```
I'm not sure what the best way for `Syntax1` and `Syntax2` is yet. Also, since this will happen in multiple files, should we factor it out once and have it in a separate file which I can include?

parikshitbajpai:
I am afraid some MOOSE checks here will still fail. The `searchParameters()` syntax in Nanoflann 1.5.0 doesn't have the obsolete `int checks_IGNORED_` param which was part of `searchParams()` arguments. This would cause `nanoflann::searchParameters(10)` calls in the MOOSE patch (https://github.com/idaholab/moose/pull/25962) to result in failing tests.

roystgnr:
Oh, wonderful.  I should have known when I saw a magic number that it was going to be necromancy.

I really need to get the libMesh submodule update PR going, in case we run into any unexpected problems there from other libMesh updates.  Want to take a crack at an updated Nanoflann compatibility PR for MOOSE and see if you can beat the submodule update?

parikshitbajpai:
> Oh, wonderful. I should have known when I saw a magic number that it was going to be necromancy.
> 
> I really need to get the libMesh submodule update PR going, in case we run into any unexpected problems there from other libMesh updates. Want to take a crack at an updated Nanoflann compatibility PR for MOOSE and see if you can beat the submodule update?

@roystgnr https://github.com/idaholab/moose/pull/25905 is open. Don't like having a hard coded arbitrary value in there but trying to beat that submodule update.

roystgnr:
Looks like the fix-to-the-fix made it from next to devel, so I'm kicking this again.

roystgnr:
And CI is finally happy.  Here we go.

PR #3691, opened 2023-11-02
Don't try to cache when using non-Lagrange mapping

roystgnr:
Most IGA problems hit a different reason to invalidate FE caches, so this bug wasn't triggering ... but the simple meshes I was using to test discontinuous-Bezier-extraction IGA were setting it off.

And the most obvious symptom was simply that we were getting different results in parallel in a new MOOSE discontinuous-IGA test, which sent me chasing after a dozen red herrings before finding the non-parallel non-MOOSE non-discontinuous-extraction bug.  The ratio of "work spent finding the problem" to "work spent fixing the problem" here was insane.

### DIFF
--- a/src/fe/fe.C
+++ b/src/fe/fe.C
@@ -190,9 +190,12 @@ void FE<Dim,T>::reinit(const Elem * elem,
           if (this->qrule->shapes_need_reinit())
             this->shapes_on_quadrature = false;
 
+          // We're not going to bother trying to cache nodal
+          // points *and* weights for fancier mapping types.
           if (this->elem_type != elem->type() ||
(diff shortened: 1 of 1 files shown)


lindsayad (approved):
sounds like quite the goose chase

roystgnr:
Logan says, regarding `http.client.RemoteDisconnected: Remote end closed connection without response`, "We’re getting a lot of incomplete messages from [civet.inl.gov](http://civet.inl.gov/) to the build boxes, which is causing even more python failures for querying civet results than normal. Looking into it…", so I'm not going to bother kicking that failed python recipe again, I'm just going to merge.  `Test MOOSE mac` did pass everything affected by this PR.

PR #3694, opened 2023-11-03
Average side nodal solutions for H_DIV elements

nmnobre:
Forgot this when we initially implemented H(div) conforming elements, but I'm pretty sure the tangential component would need to be averaged in the same way the normal component needs to for H(curl) elements.

Cheers,
-Nuno


lindsayad (approved):

roystgnr (approved):

nmnobre:
Is there an easy way we could test this in ex6?

roystgnr:
It would be a good idea to add test coverage for this, but it's inexcusable to let a fix for a visualization bug drop off the radar for a year because I hoped to get around to adding test coverage.

PR #3697, opened 2023-11-08
Update configuration to add default --enable-fpe-safety

roystgnr:
clang 15 and 16 will generate floating-point-exception emitting behavior at -O2 by default, which breaks even the MetaPhysicL submodule's `make check` tests.  I see the attraction of that (presumably they're packing N float instructions into N/4 or N/8 SIMD instructions, and then if N%4 or N%8 but the instruction is division there'll be some NaNs ignored at the end?) but we have a lot of users relying on trapping FPEs, so let's make the default behavior safe for them.

This ought to fix our newly-clang-15-using `Test clang` recipe.


roystgnr:
Well, damn.  `error: overriding currently unsupported use of floating point exceptions on this target [-Werror,-Wunsupported-floating-point-opt]` looks like it might actually be related.

roystgnr:
And it's *only* on ARM mac that it's failing, so I can't reproduce locally either...

roystgnr:
Okay - with that hack I'm seeing some expected failures and some weird IM failures but the fix is working where it's supposed to; I'm going to head back upstream to do it properly now.

roystgnr:
And now what should be the proper fix is in.  Cross your fingers.

roystgnr:
This appears to have finally fixed new Linux clang without having broken ARM Mac clang, and that "Test mac" failure is some new environment issue, so if this passes I'm merging.

roystgnr:
And that "Failed but allowed" on `Test MOOSE` is the network issues hitting a python test again.  This should be good to go.

PR #3698, opened 2023-11-08
Scrape $CC from PETSc

roystgnr:
This is the safest way to get mpicc if PETSc built it, which may be necessary if we're building an Exodus configuration that needs to find mpi.h

This is going to conflict with #3697, but that one may take a while to figure out on ARM/mac so I can just rebase and fix it afterward.


roystgnr:
The "Test clang" failure is a result of not getting #3697 in yet; the "Test MOOSE mac" failure a result of INL network issues vs the python tests.  This should still be safe to merge.

PR #3699, opened 2023-11-09
Fix off-by-one error in Abaqus element renumbering

roystgnr:
I translated 1-based-to-0-based for Abaqus-to-libMesh numbering when trying to look up elements for e.g. subdomain ids, but I forgot to translate that when *creating* the elements in the first place.

I want to add some more test coverage before merging this, but I'm putting it up now to get it available and to make sure there are no side-effects.


roystgnr:
The `Test clang` failure is still just waiting on #3697.

This PR should probably reference #3640 (where I introduced the bug) and https://github.com/idaholab/moose/discussions/26001 (our first victim of the bug).

roystgnr:
And the `Test MOOSE mac` failure is just another network issue.

roystgnr:
> Not much we can do about this now

Not sure if by "this" you mean the bug (which might be worth a new libMesh submodule update ASAP; that would also get us the Nanoflann update and the fix for clang 15+ FPE issues, neither of which I would mind seeing rushed downstream) or the recent "respect Abaqus numbering" behavior change (which we could make optional if it's causing anyone problems, though I'd still want the old "just number sequentially and ignore the file" behavior to be non-default).

dschwen:
I didn't see that this is a fix for #3640. Carry on then :-)

PR #3703, opened 2023-11-10
Error if mismatched p-levels for Lagrange between neighbors

lindsayad:
Otherwise we will silently break the continuity of the space

### DIFF
--- a/src/fe/fe_lagrange.C
+++ b/src/fe/fe_lagrange.C
@@ -823,6 +823,7 @@ void lagrange_compute_constraints (DofConstraints & constraints,
 #endif
 
   FEType fe_type = dof_map.variable_type(variable_number);
+  const bool add_p_level = dof_map.should_p_refine_var(variable_number);
 
   // Pull objects out of the loop to reduce heap operations
   std::vector<dof_id_type> my_dof_indices, parent_dof_indices;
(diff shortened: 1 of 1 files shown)


lindsayad:
most of this diff is indentation

src/fe/fe_lagrange.C
```diff
@@ -831,134 +832,149 @@ void lagrange_compute_constraints (DofConstraints & constraints,
...
+          {
+            // p refinement constraints:
+            // constrain dofs shared between
+            // active elements and neighbors with
+            // lower polynomial degrees
+            const unsigned int min_p_level =
+              neigh->min_p_level_by_neighbor(elem, elem->p_level());
+            if (min_p_level < elem->p_level())
+              libmesh_error_msg(
+                "Mismatched p-refinement levels for LAGRANGE is not currently supported");
```
jwpeterson, line 975: We often use mismatched p-level in 1D, since there you don't have a continuity problem. Can we have this new error message allow a mismatch if both neighbors are 1D?

lindsayad: This method returns at the top if `Dim == 1`

jwpeterson: Beautiful early `return` statement

roystgnr (approved):
Github *really* doesn't want to make that mostly-whitespace diff easy to read, does it?

But after some back and forth, this looks great to me.

PR #3704, opened 2023-11-12
Fix expanded $CC detection

roystgnr:
This should fix the "libMesh no longer configures without PETSc" bug that's currently preventing devel->master merges.


roystgnr:
Adding the failing recipes here too; once this passes it should be good to merge.

PR #3707, opened 2023-11-14
Better order-upgrading methods

### LINKED ISSUE #3706: Support repair of mixed-order meshes in `UnstructuredMesh::all_second_order()`
And we should support that in `all_complete_order()` too while we're at it.

Per discussion with @eshemon (feel free to tag your team), we should be able to get useful higher-order (for now, quadratic) geometry support much faster if we can do the upgrades one mesh generator at a time, relying on the ability to combine the new higher-order-enabled output with non-upgraded first-order-only output in the end.

Right now the `UnstructuredMesh` code handles upgrading mixes of lower and higher order elements, but it looks like it should only work correctly in cases (all lower-order elements are 1D, or all lower-order elements are in a separate component) where no newly-higher-order element should share a new-to-it local node with an existing higher-order element.

That limitation looks like it'll be relatively easy to fix - just do an initial sweep through the range and put any existing higher-order-only nodes into the same map we're using for new higher-order-only nodes.

roystgnr:
If this works it will resolve #3706.  I haven't added test cases yet, though; I'm just putting the PR up early so I can let CI help me make sure the refactoring involved hasn't created any obvious *regressions*.

### DIFF
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -1229,6 +1229,19 @@ public:
    */
   static ElemType first_order_equivalent_type (const ElemType et);
 
+  /**
+   * \returns The ElemType of the associated "complete" order element
+   * (which will be the same as the input if the input is already a
+   * complete-order ElemType), or INVALID_ELEM for elements that cannot be
+   * converted into complete-order equivalents.
+   *
+   * The "complete" version of an element is an element which can
+   * represent the same geometry but which has nodes available to
+   * restore degrees of freedom on any vertex, edge, or face.
+   *
+   * For example, when \p this is a \p TET4, then \p TET14 is returned.
+   */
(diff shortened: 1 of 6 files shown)


roystgnr:
Two obvious regressions ... and while writing unit tests I've also discovered that the mixed-upgrade capability needs to be more robust still if we want it to work in the range case as well as the whole-mesh case.

roystgnr:
And there's an anti-regression, where we're fixing a bug in libMesh but MOOSE has test cases depending on the bug.  I'll see about fixing that downstream.

roystgnr:
John is reporting that this isn't regressing any of the partial-order-upgrading tests in the Akselos suite either; I'll merge later this week if there are no other issues.

PR #3710, opened 2023-11-16
Fix for writing XDR files of adaptively refined meshes with extra elem integers

jwpeterson:
Also includes a unit test that segfaults for me without the fix in jwpeterson/libmesh@de15955c5.


roystgnr (approved):

jwpeterson:
Min clang does not seem to want to start at the moment, but this isn't the type of PR that's likely to break on an older compiler, no fancy C++ syntax here.

PR #3711, opened 2023-11-16
Add configure support for --with-xdr-include option

### LINKED ISSUE #3709: Hardcoded /usr/include/tirpc path
Your configure script tests for XDR in a hard-coded system path `/usr/include/tirpc`, which is not the right location to check for this on our system. Any way to include a `TIRPC_DIR` environment variable or something similar ?

```
              { $as_echo "$as_me:${as_lineno-$LINENO}: checking for XDR support in /usr/include/tirpc" >&5
$as_echo_n "checking for XDR support in /usr/include/tirpc... " >&6; }```

jwpeterson:
RPC headers can be installed in places other than `/usr/include/tirpc` so let the user specify the correct PATH.

Closes #3709.

### DIFF
--- a/configure
+++ b/configure
@@ -1256,6 +1256,8 @@ enable_strict_lgpl
 enable_nested
 enable_xdr_required
 enable_xdr
+with_xdr_include
+with_xdr_libname
 enable_boost
 with_boost
 with_boost_libdir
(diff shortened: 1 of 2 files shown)


mboisson:
Will this also work from an environment variable ? Unfortunately, some packages bundle `libMesh` in their build script, and there's no way to change the options passed to `configure` without patching the build scripts.

jwpeterson:
> Will this also work from an environment variable ?

No this patch won't work with an environment variable as-is. I think I can fix that, but FYI if you are building the libmesh used by MOOSE, it should be possible to pass additional configure flags to the script you call to build libmesh (at least it was possible at one time), so in that case you wouldn't need to patch the build scripts.

mboisson:
ah, indeed there are options to the script (though I did not see it in the documentation, they just say to run the script as is without referring to any option).

PR #3712, opened 2023-11-16
Gcc 13 warning fixes

roystgnr:
This gets me running `make check` cleanly on `--enable-paranoid-warnings --enable-werror` builds on Ubuntu 23.10.


jwpeterson (approved):
GCC 13 eh? Back in my day we were on GCC 2.95 for 6 years, and _we liked it_.

PR #3713, opened 2023-11-16
Introduced VectorizedEvalInput to encapsulate vectorized evaluate inputs

dknez:
Previously the RB EIM "vectorized evaluate" functions had a large number of separate inputs, which was tedious to deal with and means that it is an API change any time we want to enable more data to be passed to these functions. This update encapsulates these inputs in the `VectorizedEvalInput` class.


jwpeterson (approved):
:+1: This gives us a nice clean-up of lines in libmesh, and in our internal code as well.

PR #3732, opened 2023-11-28
HDG Navier Stokes

lindsayad:
Based on work by Nguyen and Cockburn. I've implemented MMS tests for both lid and channel flow type boundary value problems.

### Cavity MMS results
grid_size=2
L2 error for u is: 1.543638e-01
L2 error for v is: 1.543638e-01
L2 error for pressure is: 2.879229e-01
grid_size=4
L2 error for u is: 4.007067e-02
L2 error for v is: 4.007067e-02
L2 error for pressure is: 5.268862e-02
grid_size=8
L2 error for u is: 1.012986e-02
L2 error for v is: 1.012986e-02
L2 error for pressure is: 1.111670e-02
grid_size=16
L2 error for u is: 2.543457e-03
L2 error for v is: 2.543457e-03
L2 error for pressure is: 2.581651e-03

### Channel MMS results
grid_size=2
L2 error for u is: 2.318716e-01
L2 error for v is: 3.545995e-01
L2 error for pressure is: 6.271178e-01
grid_size=4
L2 error for u is: 6.192154e-02
L2 error for v is: 9.922736e-02
L2 error for pressure is: 1.775775e-01
grid_size=8
L2 error for u is: 1.577982e-02
L2 error for v is: 2.597410e-02
L2 error for pressure is: 4.410009e-02
grid_size=16
L2 error for u is: 3.977834e-03
L2 error for v is: 6.505690e-03
L2 error for pressure is: 1.091284e-02


lindsayad:
Marking this as ready for review. These current failures seem unrelated?

roystgnr:
Definitely unrelated.  Logan's looking atone; I have a hypothesis about and an attempted fix for the other.

lindsayad:
I've opened https://github.com/libMesh/libmesh/issues/3742 as an acknowledgment that I should try using `FEMContext`

lindsayad:
I was curious what a profile for this would look like so I made one for the 64x64 lid driven cavity case. I'm pleased that the dominant cost is the global solve as opposed to local inverses during assembly. This means there will be a lot to gain from using field split preconditioning

hdg

roystgnr:
What was your ksp/pc type for that profile?

But regardless, 64x64 is pretty small; if the assembly was already negligible there then I wouldn't worry about it in general.

The only exception I can think of is really high polynomial degree.  We already do an O(N*p^2d) assembly (p^d bases, p^d quadrature point values for each on each of N elements), even on tensor product elements where there are asymptotically cheaper tricks ... but these local inverses look like O(N*p^3d), don't they?  You've got m=p^d entries in those matrices and then you're doing O(m^3) inverses on each element?

PR #3733, opened 2023-11-29
Add missing quadrature types to string/enum map

nmnobre:
Hey Roy (@roystgnr),

Any reason for omitting the other quadrature types from 62e6b6d?

Cheers,
-Nuno


roystgnr:
No; just that I was working from the previous file rather than going back through every enum header.

roystgnr (approved):

PR #3736, opened 2023-12-05
Keep track of command line queries

roystgnr:
This gives us a mechanism by which we can tell PETSc to ignore them (right before PetscFinalize), so user codes can avoid spurious PETSc unused-option warnings.

I've tested this a bit by hand but only in one configuration; I'll throw the CI kitchen sink at it before we think about merging.

### DIFF
--- a/examples/adaptivity/adaptivity_ex1/adaptivity_ex1.C
+++ b/examples/adaptivity/adaptivity_ex1/adaptivity_ex1.C
@@ -79,11 +79,7 @@ int main(int argc, char ** argv)
   // default MPI communicator.
   Mesh mesh(init.comm());
 
-  GetPot command_line (argc, argv);
-
-  int n = 4;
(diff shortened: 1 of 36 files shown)


jwpeterson (approved):
I'd be in favor of the `libMesh::command_line_next` and `libMesh::on_command_line` refactoring even if it wasn't combined with the PETSc stuff.

roystgnr:
This should reference https://github.com/idaholab/moose/issues/25577 - though we'll need to start throwing in `libMesh::add_command_line_name()` calls in MOOSE after the libMesh submodule update before that's fully fixed.

roystgnr:
The three failing tests seem to be CI issues.  I'm disturbed by "Test with PETSc main", but it looks like that hasn't passed since ... ever?  https://civet.inl.gov/recipe_events/40576/  If the rest are happy then I'll merge.

PR #3744, opened 2023-12-08
HIERARCHIC_VEC and L2_HIERARCHIC_VEC families

roystgnr:
I'm not yet thrilled with the test coverage here (our massive FE unit test coverage *still* only covers scalar-valued FE types), but a handful of type/order combinations in two example apps is probably good enough for a new types that are basically make-it-vector wrappers around existing types.

I could probably get rid of a thousand lines of not-quite-duplicated code here via a hellish tree of macros, but I hate macros only slightly less than I hate cut-and-pasted-and-edited code, so I didn't bother.

I'm not sure if there's anything that's supposed to be tweaked to use HDG at higher order, so I *didn't* tweak anything ... which seems to be fine!  I'm seeing higher-order convergence just by bumping up all variable orders and making sure quadrature order matches.


lindsayad (commented):
I did not look in great depth, but skimming this it looks good. It is great to hear that the HDG convergence is scaling as expected! Thanks for adding that to the example

PR #3747, opened 2023-12-12
Add more preconditioner support to Eigen

roystgnr:
This started as an attempt to figure out why Eigen was mis-converging in a #3744 CI test, and it got to be significant enough for its own PR; it's working (when combined with GMRES) in the new high-p vector_fe_ex1 convergence tests now, but let's make sure it doesn't regress anywhere in the process.

### DIFF
--- a/src/solvers/eigen_sparse_linear_solver.C
+++ b/src/solvers/eigen_sparse_linear_solver.C
@@ -144,7 +144,9 @@ EigenSparseLinearSolver<T>::solve (SparseMatrix<T> & matrix_in,
           case ICC_PRECOND:
           case ILU_PRECOND: // If asked for CG we must be symmetric, right?
             {
-              Eigen::ConjugateGradient<EigenSM,UPLO,IncompleteCholesky<Number>> solver (matrix._mat);
+              Eigen::ConjugateGradient<EigenSM,UPLO,
+                IncompleteCholesky<Number,Eigen::Lower,Eigen::AMDOrdering<eigen_idx_type>>>
+                solver (matrix._mat);
               retval = do_solve(solver, "Eigen CG solver with Incomplete Cholesky preconditioning");
               break;
             }
@@ -177,7 +179,8 @@ EigenSparseLinearSolver<T>::solve (SparseMatrix<T> & matrix_in,
           case ICC_PRECOND:
           case ILU_PRECOND:
             {
-              Eigen::BiCGSTAB<EigenSM,IncompleteLUT<Number>> solver (matrix._mat);
+              Eigen::BiCGSTAB<EigenSM,IncompleteLUT<Number, eigen_idx_type>>
+                solver (matrix._mat);
               retval = do_solve(solver, "Eigen BiCGSTAB solver with ILU preconditioning");
               break;
             }
@@ -228,7 +231,8 @@ EigenSparseLinearSolver<T>::solve (SparseMatrix<T> & matrix_in,
           case ICC_PRECOND:
           case ILU_PRECOND:
             {
-              Eigen::GMRES<EigenSM,IncompleteLUT<Number>> solver (matrix._mat);
+              Eigen::GMRES<EigenSM,IncompleteLUT<Number, eigen_idx_type>>
+                solver (matrix._mat);
               retval = set_restart_and_solve(solver, "Eigen GMRES solver with ILU preconditioning");
               break;
             }
--- a/src/solvers/eigen_sparse_linear_solver.C
+++ b/src/solvers/eigen_sparse_linear_solver.C
@@ -101,8 +101,29 @@ EigenSparseLinearSolver<T>::solve (SparseMatrix<T> & matrix_in,
 
   std::pair<unsigned int, Real> retval(0,0.);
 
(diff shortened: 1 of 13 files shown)


roystgnr:
Looks like I might be passing with 32-bit indices and failing with 64-bit?

roystgnr:
Yup; my local build was 32-bit; switch to 64 replicates the compilation failure.

src/solvers/eigen_sparse_linear_solver.C
```diff
@@ -101,47 +101,109 @@ EigenSparseLinearSolver<T>::solve (SparseMatrix<T> & matrix_in,
...
   std::pair<unsigned int, Real> retval(0,0.);
 
-  const int max_its = this->get_int_solver_setting("max_its", m_its);
-  const double abs_tol = this->get_real_solver_setting("abs_tol", tol);
+  // Eigen doesn't give us a solver base class?  We'll just use a
+  // generic lambda, then.
+  auto do_solve = [this, &rhs, &solution, tol, m_its]
+    (auto & e_solver, std::string_view msg) {
+    const int max_its = this->get_int_solver_setting("max_its", m_its);
+    const double abs_tol = this->get_real_solver_setting("abs_tol", tol);
```
lindsayad, line 109: Do these need to be hard-typed or can we use `auto` here?

lindsayad: It does seem a little surprising to me that `get_real_solver_setting` returns a `double` instead of a `Real`

src/solvers/eigen_sparse_linear_solver.C
```diff
@@ -101,47 +101,109 @@ EigenSparseLinearSolver<T>::solve (SparseMatrix<T> & matrix_in,
...
+        switch (this->_preconditioner_type)
+        {
+          case IDENTITY_PRECOND:
+            {
+              Eigen::ConjugateGradient<EigenSM,UPLO,IdentityPreconditioner> solver (matrix._mat);
+              retval = do_solve(solver, "Eigen CG solver without preconditioning");
+              break;
+            }
+          case ICC_PRECOND:
+          case ILU_PRECOND: // If asked for CG we must be symmetric, right?
```
lindsayad, line 145: I would not have this much trust in a user...

src/solvers/eigen_sparse_linear_solver.C
```diff
@@ -101,47 +101,109 @@ EigenSparseLinearSolver<T>::solve (SparseMatrix<T> & matrix_in,
...
-        libMesh::out << "estimated error: " << solver.error() << std::endl;
-        retval = std::make_pair(solver.iterations(), solver.error());
-        _comp_info = solver.info();
+        const int UPLO = Eigen::Lower|Eigen::Upper;
+
+        switch (this->_preconditioner_type)
+        {
+          case IDENTITY_PRECOND:
+            {
+              Eigen::ConjugateGradient<EigenSM,UPLO,IdentityPreconditioner> solver (matrix._mat);
```
lindsayad, line 140: How does this work? I would have assumed that `UPLO` would had to have been a `constexpr` in order for it to be used as a template argument

src/solvers/eigen_sparse_linear_solver.C
roystgnr: It's not so much "I trust the user", it's "If the user's already screwed up this badly then we're not going to screw them up any worse".  But sure, I'll change that case to an error.

src/solvers/eigen_sparse_linear_solver.C
roystgnr: Heh; I want to blame @grmnptr for that, in #3625, but both John *and* I reviewed that one and neither of us noticed the poor name choice.  (name choice, not return value choice, IMHO, since we generally use `double`-typed tolerances regardless of `Real` type for a few reasons).

That's a new enough change and this is an internal enough API that I think we can just change the name.  I'll want to do that in a separate PR though.

roystgnr: I prefer the hard type here over auto; requires less effort to read and understand.  I use auto when a type is way too verbose, or when it's a dependent type in a generic function, or when I'm too lazy to remember exactly which type I want.

src/solvers/eigen_sparse_linear_solver.C
roystgnr: It has to be a [constant expression](https://en.cppreference.com/w/cpp/language/constant_expression), but doesn't have to be a `constexpr`.

I like to imagine that some day somebody's going to design a *good* programming language.

lindsayad: Are `Eigen::Upper` and `Eigen::Lower` `constexpr`?

src/solvers/eigen_sparse_linear_solver.C
lindsayad: Yea I think the primary argument for `auto` here would be if you were going to go change the return type of either of these functions, and then we would be doing a conversion perhaps without knowing it. But then again you could be doing a conversion when passing these arguments in the following lines so 🤷

PR #3750, opened 2023-12-14
Avoid unused var warning with --disable-eigen

roystgnr:

PR #3751, opened 2023-12-14
Fix -Wunused-local-typedefs with --disable-second

roystgnr:


jwpeterson (approved):

jwpeterson:
Looks like a similar fix is needed for one of the other typedefs too. I can push an update...
```
../src/systems/system.C: In lambda function:
../src/systems/system.C:1848:84: error: typedef 'OutputTensor' locally defined but not used [-Werror=unused-local-typedefs]
           typedef typename std::remove_reference<decltype(fe)>::type::OutputTensor OutputTensor;
                                                                                    ^~~~~~~~~~~~
```

roystgnr:
I hate to have to ask it, but would you mind pushing an update?  I'm going to be AFK again, for a few hours this time, shortly.

roystgnr:
We're not going to get this to pass without the changes in #3750 too, but we should at least *build*...

jwpeterson:
Looks like "Test No Second Derivatives" build passed now. I'll probably just merge before other stuff finishes since we are still waiting on #3750 anyway.

PR #3752, opened 2023-12-14
Use vertices to define low_order_key() for neighbor-finding and mesh stitching

### LINKED ISSUE #3738: Support mixed-order meshes in `find_neighbors()`, `stitch_meshes()`
While working on the unit tests for #3706 I ran into an issue: a mixed-order mesh is currently a very fragile thing to play with, because `UnstructuredMesh::find_neighbors()` treats mixed-order neighbors as **non**-neighbors - the higher order element has more nodes on their shared side, those additional nodes create a different `elem.key(side)`, and those different keys don't get recognized as a side that should be shared.  `UnstructuredMesh::stitch_meshes()` has the same problem: it will happily handle merging of nodes along mixed-order shared sides, but then when it comes time to create the `neighbor_ptr()` links between those sides it'll use `key()` on each and fail to recognize the identity there.

I'm thinking this might be as easy to fix as creating a `low_order_key()` (to preserve the existing API's behavior) that only considers vertices, not mid-edge or mid-face nodes, and then using that instead in these two places.

Pinging @eshemon and @miaoyinb so they can keep track of progress here too.  IMHO they shouldn't wait on this before beginning to upgrade mesh generators to second-order, though; the inadequacy here only affects stitched-together mixed-order compositions, and it ought to be possible to make a lot of progress even while stuck only doing single-component-level testing.

roystgnr:
This resolves #3738 for me

### DIFF
--- a/include/geom/cell_hex.h
+++ b/include/geom/cell_hex.h
@@ -125,10 +125,17 @@ public:
 
   /**
    * \returns An id associated with the \p s side of this element.
-   * The id is not necessarily unique, but should be close.  This is
-   * particularly useful in the \p MeshBase::find_neighbors() routine.
+   * The id is not necessarily unique, but should be close.
    */
   virtual dof_id_type key (const unsigned int s) const override;
+
+  /**
+   * \returns An id associated with the \p s side of this element, as
+   * defined solely by element vertices.  The id is not necessarily
+   * unique, but should be close.  This is particularly useful in the
+   * \p MeshBase::find_neighbors() routine.
+   */
+  virtual dof_id_type low_order_key (const unsigned int s) const override;
 
   /**
    * \returns \p Hex8::side_nodes_map[side][side_node] after doing some range checking.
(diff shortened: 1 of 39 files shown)


roystgnr:
Looks like the new tests are passing in serial but still failing in parallel, even on a ReplicatedMesh.  That's unexpected.

roystgnr:
Our `_boundary_side_id` is sorted by pointer, so it differs from processor to processor, but in `BoundaryInfo::clear_stitched_boundary_side_ids` with `clear_nodeset_data` we just operate on the first element from which we see a side ... so if that's the higher-order element then we clear the higher-order side nodes' ids and if it's not we don't.  And it looks like we're clearing *all* node boundary ids there, even if there are more than just the ones we used to stich.  Should be easy enough to fix both problems at once.

roystgnr:
The `Test MOOSE min clang` failures are an unrelated problem downstream.  If everything else seems happy here then we're ready to merge.

PR #3754, opened 2023-12-20
Removed unused RB EIM code

dknez:
Removed RB EIM code related to "observation points" since this is unused code.


jwpeterson (approved):
This is an API change, but likely an API only used by our internal code (and even that was a small fix) so I'm going to go ahead and merge. The Min clang failure is unrelated.

PR #3757, opened 2023-12-30
Do same vector operations in solver precheck as resid/jac eval

lindsayad:
These include
- Most importantly, updating the `current_local_solution` based on the vector passed in by PETSc
- Then performing constraint enforcement on the `current_local_solution`

This should allow a lot more code reuse in MOOSE where it's assumed that any computations using the degrees of freedom in the nonlinear system should use the `current_local_solution` vector. As an indication of the possbility for more code re-use, the library level changes allowed removal of the `ghosted_old_solution` data member in the HDG NS example.

Unrelated to the library changes I also removed the `parallel_increment` data member in the HDG NS example because it's simply not needed. This furthers the example of what I will implement in MOOSE, e.g. the number of new vectors I will need to add for HDG should only be one.


roystgnr (approved):

roystgnr:
That failing test is just warnings we need to fix downstream.

PR #3759, opened 2024-01-04
Make is_prepared() more fine-grained

roystgnr:
I'm still trying to figure out how to make this easy to use, but I'd like to at least test that what I've got so far still passes CI.

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -145,7 +145,7 @@ public:
   virtual ~MeshBase ();
 
   /**
-   * A partitioner to use at each prepare_for_use()
+   * A partitioner to use at each partitioning
    */
   virtual std::unique_ptr<Partitioner> & partitioner() { return _partitioner; }
 
@@ -174,7 +174,7 @@ public:
    * No Node is removed from the mesh, however even NodeElem elements
    * are deleted, so the remaining Nodes will be considered "unused"
    * and cleared unless they are reconnected to new elements before
-   * the next \p prepare_for_use()
+   * the next preparation step.
    *
    * This does not affect BoundaryInfo data; any boundary information
    * associated elements should already be cleared.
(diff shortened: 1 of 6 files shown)


lindsayad:
Does "do not merge" also mean "do not review"?

roystgnr:
> Does "do not merge" also mean "do not review"?

It's already bad enough to need fixing, huh?  ;-)

Nah, please jump in as you see fit.  I've got more I'm wanting to add here, but if there's flaws in the foundation then it's probably better to fix them before I put more layers on top.

lindsayad (commented):
Just looking at the header this seems like a great foundation to me

roystgnr:
This might actually be in shape to merge soon.  My remaining concerns:

1.  The additional internal testing I've added might trigger dbg and/or devel failures downstream, in tests of code that doesn't properly prepare a mesh before solving (so it'll fail the new assertions) but either doesn't rely on the unprepared data or just compares to a gold solution that mis-relied on it (so it's not already failing tests).  I'll add every recipe I think might be affected to the CI here.

2.  The additional internal testing I've added might be too expensive in dbg mode, where it tests preparation validity before every assembly.  I *think* we can afford that (`prepare_for_use()` is much cheaper if we don't repartition, which we don't when checking validity).  If we see CI timeouts then I'll try moving my testing from assembly methods to `solve()` methods.  If we still see timeouts after that then I'll remove the intensive testing entirely for now.

roystgnr:
I'm moving the "incredibly careful assertions" part of this PR to a new branch, https://github.com/roystgnr/libmesh/tree/mesh_preparation_assertions, which I'll put into a new PR for after this one goes in.  That (plus the other PRs I've just factored out, plus one more bugfix PR that'll wait on #4394 because it requires a commit there) should get this PR down to *just* the functionality promised in the title, and should leave it in a ready-to-review state.

include/mesh/mesh_base.h
```diff
@@ -1844,6 +1987,58 @@ class MeshBase : public ParallelObject
                                      const boundary_id_type b2);
 #endif
 
+  /**
+   * Flags indicating in what ways a mesh has been prepared for use.
+   */
+  struct Preparation {
```
jwpeterson: We usually do all opening curly brace on next line.

roystgnr:
I feel like 90% of the work on this PR has been fixes, not for bugs in the PR, but for bugs exposed by the PR.

Along with the API name changes I figured I'd throw in a couple more unit tests, despite the fact that they would *surely* be redundant with all the test coverage I got in the branch I'm following this one with ... and, long story short, there'll be a couple more non-trivial commits coming along with the name changes and the new tests: a bug fix in the MeshBase copy constructor, and some changes to allow copying and cloning of some types of badly-unprepared meshes.

jwpeterson (approved):
Thanks for doing that renaming 👍

PR #3761, opened 2024-01-09
Support interior_parent() in Elem order upgrades

roystgnr:
This fixes a segfault (or in dbg mode, an assertion failure) in MALAMUTE tests for me.

Refs https://github.com/idaholab/moose/pull/26462

### DIFF
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -1056,7 +1056,8 @@ Elem * Elem::interior_parent ()
 void Elem::set_interior_parent (Elem * p)
 {
   // interior parents make no sense for full-dimensional elements.
-  libmesh_assert_less (this->dim(), LIBMESH_DIM);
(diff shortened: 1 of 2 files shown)


roystgnr:
Test MOOSE mac failure is something wonky with CI, but I'm going to re-run Test MOOSE debug, just in case the slowdown there was an actual regression.

roystgnr:
If that passes I'll merge; I'm still working on replicating the other half-dozen new failures in MOOSE apps, but this at least fixes one of them and that one was a real (long-standing, if newly triggered) bug.

lindsayad (approved):
If you think this is too expensive I’d also be fine saying something like “you should only make your level 0 lower d elements after you’re done meddling with your level 0 higher d elements” and the malamute use case continues to be invalid

PR #3765, opened 2024-01-11
Avoid creating duplicate entries in BoundaryInfo multimaps

roystgnr:
This caused a half dozen of the exodiffs in https://github.com/idaholab/moose/pull/26462, after it was triggered by the stitching changes in https://github.com/libMesh/libmesh/pull/3752/

### DIFF
--- a/src/mesh/boundary_info.C
+++ b/src/mesh/boundary_info.C
@@ -3271,4 +3271,31 @@ BoundaryInfo::get_global_boundary_ids() const
   return _global_boundary_ids;
 }
 
+void
+BoundaryInfo::libmesh_assert_valid_multimaps() const
+{
+#ifndef NDEBUG
+  auto verify_multimap = [](const auto & themap) {
+    for (const auto & [key, val] : themap)
+      {
+        auto range = themap.equal_range(key);
+
+        int count = 0;
+        for (auto it = range.first; it != range.second; ++it)
+          if (it->second == val)
+            ++count;
+
+        libmesh_assert(count == 1);
+      }
+  };
+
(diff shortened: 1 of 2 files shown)


src/mesh/boundary_info.C
```diff
@@ -1829,6 +1829,9 @@ void BoundaryInfo::renumber_id (boundary_id_type old_id,
   for (auto & p : _boundary_node_id)
     if (p.second == old_id)
       {
+        // If we already have this id on this node, we don't want to
+        // create a duplicate in our multimap
+        this->remove_node(p.first, new_id);
```
lindsayad, line 1834: What’s the removal expense? Can we not just keep the old duplicate and not add the new duplicate?

roystgnr: I did think about replacing the range-for with explicit iterators, then reusing those iterators to avoid the `O(N log N)` cost of descending into the tree a second and third time redundantly.  But after that there's still the same subiteration over the same conceptually-`equal_range()` results.  We'd be a factor of 3 faster at best, we could be slower at worst (if you've already stepped past the start of the range you have to get back to the beginning, either iterating backwards or repeatedly testing for range starts to cache them), and the code gets a lot uglier.

I wouldn't have even thought about it ... except the `FuelPin3DMeshGenerator` does have **eighty-four** calls to `change_boundary_id()`.  I'd fear premature optimization if I tried to make this run any faster without a slow test case to use for before-and-after profiling, but if someone does come up with such a test case my current plan is to apologize meekly and start running it.

lindsayad (approved):

PR #3768, opened 2024-01-16
No unneeded subdomain-restricted var constraints

roystgnr:
We're already doing this check in `compute_proj_constraints`, for all the *non*-Lagrange variable types, but for subdomain-restricted Lagrange variables we were creating unnecessary constraints in the case of hanging nodes whose neighboring coarse elements were entirely in subdomains that don't support the variable.

### DIFF
--- a/src/fe/fe_lagrange.C
+++ b/src/fe/fe_lagrange.C
@@ -822,7 +822,8 @@ void lagrange_compute_constraints (DofConstraints & constraints,
    }
 #endif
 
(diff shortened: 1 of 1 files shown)


lindsayad (approved):
This is great, thanks @roystgnr

roystgnr:
The "Test MOOSE debug"==Canceled here seems to be just the usual execution time variation that's hit us repeatedly.  I'll change the recipe to give it more leeway in the future.

PR #3770, opened 2024-01-22
Correct overzealous assertion to work with a NodeElem-only mesh

loganharbour:


roystgnr (approved):

loganharbour:
Needs another, unfortunately. Poor std::max type comparison failure

roystgnr (approved):

roystgnr:
You could have just used 1u, to retain `max()`-compatibility?  But yeah, IMHO `max(int, unsigned int)` should have been in `namespace std` long ago, among other overloads.  We still have `libmesh_augment_std_namespace.h` cheating on behalf of mixed floating-point types, though we don't let it pollute anyone else's code, and I still regularly vacillate between "we should just move this stuff to the libMesh namespace and respect the C++ standard" versus "if the C++ standard wanted that level of respect it should have earned it".

PR #3776, opened 2024-02-05
Matrix read/write to matlab format, unit test coverage improvements

roystgnr:
My original impetus here was just to get a `PetscMatrix::read_matlab()` working for some prototyping of other features, but then I decided to get some proper test coverage of it, and that entailed adding a handful of other `SparseMatrix` methods plus adding non-PETSc implementations of some of the existing methods.

The git history here is the result of a ton of interactive rebasing to clean up a very trashy original history; I think it's now at the point where it should pass tests at all intermediate stages, but I wouldn't guarantee that.

### DIFF
--- a/include/base/sparsity_pattern.h
+++ b/include/base/sparsity_pattern.h
@@ -159,6 +159,11 @@ public:
   { return nonlocal_pattern; }
 
   /**
+   * The total number of nonzeros in the global matrix.
+   */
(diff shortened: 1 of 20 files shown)


lindsayad:
this will be nice for more quickly generating eigenvalue plots

roystgnr:
Mac "Test Python" failure unrelated

PR #3778, opened 2024-02-08
RBParameter vector-valued parameter support - Part1 - Foundation

pmoseley:
This is some foundational work to add support in RBParameters for vector-valued parameters, e.g. `_parameters` defined as a `  std::map<std::string, std::vector<std::vector<Real>>>`. That change will happen in a future PR, here I'm just putting some work in place.

In this PR, I've:
- clarified some docs
- added new and improved existing error checking (in opt)
- changed some arguments to be const references, in order to avoid needless copies
- Removed NumericVector from rb_construcion_base, because it will not be able to function in the future as `NumericVector<std::vector<Real>>`. It's use was quite limited here, so this was relatively simple.
- changed the user API in two ways (described below)

The changes to the API are:

- renaming "steps" to "samples", e.g. `RBParameters::n_samples()`
This better reflects the way these values are typically used; during RB training we typically have a vector of "samples" that does not correspond to the definitions we commonly use for "loadsteps" or "timesteps".

- taking input vectors of Reals instead of Numbers
Previously, we took some inputs as Numbers and converted them to Real internally (without making it clear to the user that any imaginary parts would be lost). Now we explicitly take Reals as inputs, which should make that limitation more clear and avoids some unnecessary copies. Support for Number is not inherently difficult if someone wants it at some point in the future, the primary limitation to address would be the capnproto serialization/deserialization.

Normally, other than the changes to the API, existing tests should continue to function. Backwards-compatibility will be maintained after the transition to vector-valued parameters as much as possible (the biggest upcoming breaking changes will be related to the iterators).

### DIFF
--- a/include/reduced_basis/rb_parametrized.h
+++ b/include/reduced_basis/rb_parametrized.h
@@ -211,9 +211,13 @@ private:
                                                 std::map<std::string, std::vector<Real>> & discrete_parameter_values_in);
 
   /**
-   * Helper function to check that \p params is valid.
+   * Helper function to check that \p params is valid:
+   *  - same number of parameters (error)
+   *  - parameter values are within the min/max range (warning)
+   *  - discrete values are correctly discrete within tolerance (warning)
+   * Warnings are only printed if "verbose_mode" is true.
    */
-  bool valid_params(const RBParameters & params);
+  bool check_if_valid_params(const RBParameters & params) const;
 
   /**
    * Helper function to check if the specified value
--- a/include/reduced_basis/rb_parametrized.h
+++ b/include/reduced_basis/rb_parametrized.h
@@ -27,6 +27,7 @@
 #include "libmesh/reference_counted_object.h"
 
 // C++ includes
+#include <cstddef>
 #include <vector>
 
 namespace libMesh
@@ -69,6 +70,7 @@ public:
 
   /**
    * Initialize the parameter ranges and set current_parameters.
+   * The input min/max parameters should have exactly 1 sample each.
    */
   void initialize_parameters(const RBParameters & mu_min_in,
                              const RBParameters & mu_max_in,
@@ -111,8 +113,13 @@ public:
 
   /**
    * Set the current parameters to \p params
+   * The parameters are checked for validity; an error is thrown if
+   * the number of parameters or samples is different than expected.
+   * We \return a boolean true if the new parameters are within the min/max
+   * range, and false otherwise (but the parameters are set regardless).
+   * Enabling the "verbose_mode" flag will also print more details.
    */
-  void set_parameters(const RBParameters & params);
+  bool set_parameters(const RBParameters & params);
 
   /**
    * Get an RBParameters object that specifies the minimum allowable value
--- a/src/reduced_basis/rb_construction_base.C
+++ b/src/reduced_basis/rb_construction_base.C
@@ -65,7 +65,7 @@ std::pair<unsigned int, unsigned int> calculate_n_local_samples_and_index(
   unsigned int n_local_samples = n_global_samples;
   unsigned int first_local_index = 0;
 
-  if(serial || communicator.size()==1)
+  if (serial || communicator.size() == 1)
     return {n_local_samples, first_local_index};
 
   // Calculate the number of training parameters local to this processor
@@ -230,12 +230,15 @@ RBParameters RBConstructionBase<Base>::get_params_from_training_set(unsigned int
   // If the _training_parameters are empty, return an empty RBParameters.
   // Otherwise, create a new RBParameters object from the single sample requested.
   RBParameters params;
-  if(!_training_parameters.empty())
+  if (!_training_parameters.empty())
     {
-      libmesh_error_msg_if((global_index<this->get_first_local_training_index()) ||
-                           (global_index>=this->get_last_local_training_index()),
-                           "Error: index " << global_index << " must be within range: "
-                           << this->get_first_local_training_index() << " - "
+      libmesh_error_msg_if((global_index < this->get_first_local_training_index()) ||
+                           (global_index >= this->get_last_local_training_index()),
+                           "Error: index "
+                           << global_index
+                           << " must be within range: "
+                           << this->get_first_local_training_index()
+                           << " - "
                            << this->get_last_local_training_index());
 
       const numeric_index_type local_index = global_index - get_first_local_training_index();
@@ -326,7 +329,8 @@ void RBConstructionBase<Base>::initialize_training_parameters(const RBParameters
       _n_local_training_samples = last_local_index-first_local_index;
     }
   _n_global_training_samples = _n_local_training_samples;
-  if(!serial_training_set)
+
+  if (!serial_training_set)
     this->comm().sum(_n_global_training_samples);
 
   // For each parameter that only allows discrete values, we "snap" to the nearest
@@ -382,7 +386,8 @@ void RBConstructionBase<Base>::load_training_set(const std::map<std::string, std
       _n_local_training_samples =
         cast_int<numeric_index_type>(new_training_set.begin()->second.size());
       _n_global_training_samples = _n_local_training_samples;
-      if(!serial_training_set)
+
+      if (!serial_training_set)
         {
           this->comm().sum(_n_global_training_samples);
 
@@ -398,11 +403,9 @@ void RBConstructionBase<Base>::load_training_set(const std::map<std::string, std
         }
 
       // Ensure that the parameters are the same.
-      for (auto & it : _training_parameters)
-        {
-          libmesh_error_msg_if(new_training_set.find(it.first)==new_training_set.end(),
-            "Parameters must be identical in order to overwrite dataset.");
-        }
+      for (const auto & pr : _training_parameters)
+        libmesh_error_msg_if(!new_training_set.count(pr.first),
+                             "Parameters must be identical in order to overwrite dataset.");
 
       // Copy the values from the new_training_set to the internal training_parameters.
       _training_parameters = new_training_set;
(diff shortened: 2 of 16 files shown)


tests/utils/rb_parameters_test.C
```diff
@@ -96,17 +99,47 @@ public:
     CPPUNIT_ASSERT_EQUAL(m["c"], 3.);
   }
 
+  void testIteratorsWithSteps()
```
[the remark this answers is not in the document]
pmoseley: Great, yeah I tried to update pretty thoroughly in the code and comments but wasn't careful for the unit tests. Updated, thanks!

src/reduced_basis/rb_parametrized.C
```diff
@@ -374,31 +374,46 @@ void RBParametrized::print_discrete_parameter_values() const
...
+  for(const auto & [param_name, param_val] : params)
     {
-      const std::string & param_name = pr.first;
-      valid = valid && ( (get_parameter_min(param_name) <= params.get_value(param_name)) &&
-                         (params.get_value(param_name) <= get_parameter_max(param_name)) );
+      // Check every parameter value (including across samples) to ensure
+      // it's within the min/max range.
+      const bool outside_range =
+        ((param_val < get_parameter_min(param_name)) &&
+         (param_val > get_parameter_max(param_name)));
```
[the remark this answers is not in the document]
pmoseley: Yes, good catch. I fixed that, and added some unit tests for RBParametrized.

Which also revealed that RBParametrized didn't really support multi-sample parameters, so I've updated (at least partially) that as well.

src/reduced_basis/rb_construction_base.C
```diff
@@ -398,9 +398,9 @@ void RBConstructionBase<Base>::load_training_set(const std::map<std::string, std
         }
 
       // Ensure that the parameters are the same.
-      for (auto & [param_name, _] : _training_parameters)
+      for (auto & it : _training_parameters)
```
jwpeterson, line 401: I really prefer not to use the name `it` here, since, elsewhere in the code, that usually refers to an iterator variable, whereas in range-based for-loops like this, it is a reference to a `std::pair`, hence the original variable name `pr`.

src/reduced_basis/rb_construction_base.C
```diff
@@ -469,115 +505,67 @@ void RBConstructionBase<Base>::generate_training_parameters_random(const Paralle
...
 
           // Generate log10 scaled training parameters
-          if (log_param_scale[param_name])
+          if (log_param_scale.at(param_name))
             {
               Real log_min   = log10(min_parameters.get_value(param_name));
               Real log_range = log10(max_parameters.get_value(param_name) / min_parameters.get_value(param_name));
 
-              training_vector->set(index, pow(10., log_min + random_number*log_range ) );
+              sample_vector[i] = {pow(10., log_min + random_number*log_range )};
```
jwpeterson, line 527: Looks like this was wrong in the original as well, but pow -> std::pow

src/reduced_basis/rb_construction_base.C
```diff
@@ -469,115 +505,67 @@ void RBConstructionBase<Base>::generate_training_parameters_random(const Paralle
...
-      numeric_index_type first_index = training_vector->first_local_index();
-      for (auto i : make_range(training_vector->local_size()))
+      for (auto i : make_range(n_local_training_samples))
         {
-          numeric_index_type index = first_index + i;
           Real random_number = static_cast<Real>(std::rand()) / RAND_MAX; // in range [0,1]
 
           // Generate log10 scaled training parameters
-          if (log_param_scale[param_name])
+          if (log_param_scale.at(param_name))
```
jwpeterson, line 522: Use `libmesh_map_find(log_param_scale, param_name)` instead of `at` here. The helper function gives a better error message that includes a stack trace when `param_name` is not in the map.

src/reduced_basis/rb_construction_base.C
```diff
@@ -607,21 +595,23 @@ void RBConstructionBase<Base>::generate_training_parameters_deterministic(const
...
   // then we take a tensor product to obtain the final set of training samples.
   std::vector<std::vector<Real>> training_samples_per_param(num_params);
   {
     unsigned int i = 0;
-    for (const std::string & param_name : param_names)
+    for (const auto & it : min_parameters)
       {
-        bool use_log_scaling = log_param_scale[param_name];
+        const std::string &param_name = it.first;
+        const bool use_log_scaling = log_param_scale.at(param_name);
```
jwpeterson, line 614: at -> libmesh_map_find

src/reduced_basis/rb_construction_base.C
```diff
@@ -607,21 +595,23 @@ void RBConstructionBase<Base>::generate_training_parameters_deterministic(const
...
+                         << total_samples_check << " samples instead?");
   }
 
   // First we make a list of training samples associated with each parameter,
   // then we take a tensor product to obtain the final set of training samples.
   std::vector<std::vector<Real>> training_samples_per_param(num_params);
   {
     unsigned int i = 0;
-    for (const std::string & param_name : param_names)
+    for (const auto & it : min_parameters)
```
jwpeterson, line 611: it -> pr

src/reduced_basis/rb_construction_base.C
```diff
@@ -673,18 +663,18 @@ void RBConstructionBase<Base>::generate_training_parameters_deterministic(const
             for (indices[2]=0; indices[2]<n_training_samples_per_param[2]; indices[2]++)
               {
                 unsigned int param_count = 0;
-                for (const std::string & param_name : param_names)
+                for (const auto & [param_name, _] : min_parameters)
```
jwpeterson, line 666: We don't currently use underscore elsewhere in the library to refer to unused parameters. Although I think it might become standardized at some point, I'd prefer to use `pr` and `pr.first` for consistency for now. For now, it doesn't even save much (if any) code since we still have to explicitly libmesh_ignore it.

src/reduced_basis/rb_construction_base.C
```diff
@@ -673,18 +663,18 @@ void RBConstructionBase<Base>::generate_training_parameters_deterministic(const
...
                   {
-                    libmesh_error_msg_if(!training_parameters_in.count(param_name), "Invalid parameter name: " + param_name);
+                    libmesh_ignore(_);
+                    const auto & it = local_training_parameters_in.find(param_name);
```
jwpeterson, line 669: Minor comment but `find` returns by value so we don't need the reference here.

src/reduced_basis/rb_construction_base.C
```diff
@@ -713,12 +706,12 @@ void RBConstructionBase<Base>::broadcast_parameters(unsigned int proc_id)
   // do the broadcast
   this->comm().broadcast(current_parameters_vector, proc_id);
 
-  // update the copy of the RBParameters object
+  // update the copy of the RBParameters object.
+  // TODO - this only works with single-sample parameter objects.
   unsigned int count = 0;
-  for (const auto & pr : current_parameters)
+  for (const auto & it : current_parameters)
```
jwpeterson, line 712: it -> pr

src/reduced_basis/rb_parameters.C
```diff
@@ -150,30 +150,35 @@ void RBParameters::push_back_extra_value(const std::string & param_name, Real va
...
+    "Requesting value for extra parameter " << param_name << ", but parameter contains multiple samples.");
+  return sample_vec[0];
 }
 
 Real RBParameters::get_extra_value(const std::string & param_name, const Real & default_val) const
 {
   // same as get_value(param_name, default_val) but for the map of extra parameters
   auto it = _extra_parameters.find(param_name);
-  return ((it != _extra_parameters.end() && it->second.size() != 0) ? it->second[0] : default_val);
+  if(it==_parameters.end())
```
jwpeterson, line 163: Space after `if` in if-statements and preferably around comparison operators, so:
```
if (it == _parameters.end())
```
We do have some [source code formatting conventions](https://github.com/libMesh/libmesh/wiki/Source-code-formatting-conventions) but nothing that automatically enforces them... I can push a commit fixing stuff like this if you want.

src/reduced_basis/rb_construction_base.C
```diff
@@ -469,115 +505,67 @@ void RBConstructionBase<Base>::generate_training_parameters_random(const Paralle
...
-          }
-        else
-          {
-            training_parameters_in[param_name]->init(n_training_samples_in, false, SERIAL);
-          }
-      }
+  const auto &[n_local_training_samples, first_local_index] =
+      calculate_n_local_samples_and_index(communicator, n_global_training_samples_in,
+                                          serial_training_set);
+  for (const auto & it : min_parameters)
```
jwpeterson, line 511: it -> pr

include/reduced_basis/rb_parametrized.h
```diff
@@ -27,6 +27,7 @@
 #include "libmesh/reference_counted_object.h"
 
 // C++ includes
+#include <cstddef>
```
jwpeterson, line 30: Was this include needed for `std::size_t` or something?

src/reduced_basis/rb_construction_base.C
```diff
@@ -469,115 +505,67 @@ void RBConstructionBase<Base>::generate_training_parameters_random(const Paralle
...
+    local_training_parameters_in[it.first] = std::vector<Real>(n_local_training_samples);
 
   // finally, set the values
-  for (auto & pr : training_parameters_in)
+  for (auto & [param_name, sample_vector]: local_training_parameters_in)
```
jwpeterson, line 515: Space before and after : in range-based for loops.

jwpeterson (commented):
I think I have mostly minor comments remaining at this stage so I can push an update fixing those if that's OK with you.

src/reduced_basis/rb_construction_base.C
pmoseley: Sure. I switched to `.at()` from `[]` when I made the `log_param_scale` const, not for any explicit need of bounds checking.

include/reduced_basis/rb_parametrized.h
pmoseley: Yeah probably. I run IWYU along with my linter, and it automatically adds these sometimes. It probably compiles without this line, by pulling it in from somewhere els.

pmoseley:
> I think I have mostly minor comments remaining at this stage so I can push an update fixing those if that's OK with you.

Sounds good to me, if you don't mind. Sorry about the formatting inconsistencies, typically I haven't been very careful about spaces.

jwpeterson (approved):

jwpeterson:
Failing MOOSE Clang test not related to this PR.

PR #3780, opened 2024-02-08
Read mesh constraint rows from an arbitrary matrix

roystgnr:
I don't like the parallel scalability here, but I don't see how to do better without very strict assumptions about partitioning.

The test coverage here needs improvement, but it's at a state where it should be reasonable to hand it to CI and make sure there aren't any obvious issues I've missed.


roystgnr:
I'm actually content with the test coverage for now; at least it's been enough to catch several mistakes in my first few attempts at the new feature...

PR #3782, opened 2024-02-13
Only do CPPUNIT_ASSERT_THROW tests if libmesh exceptions are enabled

jwpeterson:
This was causing a failure on the devel -> master merge testing.


jwpeterson:
Test MOOSE Mac error is unrelated Python error:
```
ERROR: testNotHeavy (test_SoftHeavyDependency.TestHarnessTester)
Heavy test is skipped while non-heavy tests are not
```

_So you're saying that non-heavy tests are not not being skipped?_

PR #3784, opened 2024-02-14
Support and use simple -pc_type on non-PETSc solvers

roystgnr:
I'm on some flaky internet today, on a new laptop where I'd not yet gotten around to building PETSc, and while working on fixing that I stumbled across some too-slow-in-dbg Eigen sparse solver behavior that had an IMHO natural fix.

Assuming I haven't broken anything else in the process, of course.  I still can't test this with PETSc...

### DIFF
--- a/examples/transient/transient_ex2/run.sh
+++ b/examples/transient/transient_ex2/run.sh
@@ -7,7 +7,7 @@ source $LIBMESH_DIR/examples/run_common.sh
 example_name=transient_ex2
 example_dir=examples/transient/$example_name
 
-options="pipe-mesh.unv"
+options="pipe-mesh.unv -pc_type jacobi"
 run_example "$example_name" "$options"
 
 # No benchmark here; "result_node = 274" makes mesh modification iffy
--- a/examples/transient/transient_ex2/transient_ex2.C
+++ b/examples/transient/transient_ex2/transient_ex2.C
@@ -260,6 +260,8 @@ int main (int argc, char** argv)
         fill_dirichlet_bc(equation_systems, "Wave");
 
       // Solve the system "Wave".
+      // Tell the user what we are doing.
+      libMesh::out << "Solving time step " << time_step << '/' << n_time_steps << std::endl;
       t_system.solve();
 
       // After solving the system, write the solution
--- a/include/solvers/linear_solver.h
+++ b/include/solvers/linear_solver.h
@@ -85,9 +85,9 @@ public:
   bool initialized () const { return _is_initialized; }
 
   /**
-   * Release all memory and clear data structures.
+   * Reset data
    */
-  virtual void clear () {}
+  virtual void clear ();
 
   /**
    * Initialize data structures if not done so already.
@@ -125,6 +125,12 @@ public:
    * Sets the type of preconditioner to use.
    */
   void set_preconditioner_type (const PreconditionerType pct);
+
+  /**
+   * Sets the preconditioner to use a default type or to use a type
+   * specified via a PETSc-style command line argument..
+   */
+  void set_default_preconditioner_type ();
 
   /**
    * Attaches a Preconditioner object to be used
--- a/src/solvers/eigen_sparse_linear_solver.C
+++ b/src/solvers/eigen_sparse_linear_solver.C
@@ -55,13 +55,8 @@ EigenSparseLinearSolver(const Parallel::Communicator & comm_in) :
 template <typename T>
 void EigenSparseLinearSolver<T>::clear ()
 {
-  if (this->initialized())
-    {
-      this->_is_initialized = false;
-
-      this->_solver_type         = BICGSTAB;
-      this->_preconditioner_type = ILU_PRECOND;
-    }
+  this->LinearSolver<T>::clear ();
+  this->_solver_type = BICGSTAB;
 }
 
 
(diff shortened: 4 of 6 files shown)


roystgnr:
Or maybe I'm not the one with internet problems?  3 failures to `git fetch` is kind of unprecedented...

jwpeterson:
This seems potentially confusing/inconsistent? 

Confusing since someone could send you a command line with `-pc_type lu` that's not working/crashing for them, but works fine for you, and it's because they are configured without PETSc and didn't know they were actually getting Eigen's LU solver. 

Inconsistent since we only do this for one particular flag, i.e. why not also support `-ksp_type` or any of the thousands of other PETSc command line args? 

Finally, just as a preference I would rather not further propagate PETSc's "single-dash for long command line args" since I think it's more standard to do e.g. `--pc-type lu`.

roystgnr:
The confusing behavior can happen in master right now - if you follow instructions to set -pc_type lu but you're using Eigen, since they have a LU solver but not preconditioner we'll happily continue onward with ILU instead, and if someone went to the trouble to avoid defaulting to ILU in PETSc then there's a good chance it won't work on the same problem in Eigen either.

With the PR, because the command line option would set LU_PRECOND, the convergence situation can be even worse (our EigenSparseLinearSolver behavior for an unsupported preconditioner type is to fallback on the old Eigen default of Jacobi; I should probably change that to the libMesh default of ILU...), but the user experience is probably better, because when a solve call sees an unsupported preconditioner setting it'll at least scream a "No EigenSparseLinearSolver support for LU_PRECOND" warning before trying its best with a fallback.

There's definitely still a possibility of asking for a supported preconditioner and one user getting a better result from PETSc than another gets from Eigen (maybe PETSc has a better default pivoting threshold, maybe there were other options like `-pc_factor_mat_ordering_type` that Eigen can't support) ... but again I don't see how this PR makes that situation *worse*.  It's a problem already (e.g. because PETSC defaults to GMRES but that's still experimental enough in Eigen that we default to BICGSTAB there) and It used to be a much worse problem (because we just used Jacobi with Eigen...)

> why not also support -ksp_type or any of the thousands of other PETSc command line args?

Because I'm lazy for question 1, because other solver packages often can't for question 2.  I'm willing to be less lazy and add `-ksp_type` to this PR for consistency (probably should have done that from the get-go...).  I can't add Eigen interfaces to PETSc options unsupported by Eigen, but with all the third-party-dependency (and dependency-of-dependency...) configuration options for PETSc it's already not uncommon to have environments where you run into other users' PETSc options unsupported by (your build of) *PETSc*.

> Finally, just as a preference I would rather not further propagate PETSc's "single-dash for long command line args" since I think it's more standard to do e.g. --pc-type lu

Yeah, I admit I don't like that at all either.  I just considered "make non-PETSc users suffer through PETSc's CLI option prefix" to be a lesser sin than "require users to use a different CLI option depending on their backend".  We could try to add a layer of GNU standards, intercepting e.g. "--pc_type jacobi" and translating it to "-pc_type jacobi" first if we're passing it on to PETSc?  But I fear that nobody will use it that way, and then instead of worrying about how PETSC's Jacobi compares to Eigen's we'll be back to worrying about how PETSc's Jacobi compares to Eigen's ILU.

jwpeterson:
> but the user experience is probably better, because when a solve call sees an unsupported preconditioner setting it'll at least scream a "No EigenSparseLinearSolver support for LU_PRECOND" warning before trying its best with a fallback.

OK, I agree that it's definitely possible to get unexpected/confusing behavior in master now, namely *silently* getting a different preconditioner or solver from the one you requested. So if the PR is addressing that by printing a warning or error then I agree it's an improvement.

> I'm willing to be less lazy and add -ksp_type to this PR for consistency

I'm definitely not pushing to add even more cross-solver solver support for PETSc's very idiosyncratically named (and recognizable by most people in the field) options... I was just hoping to use that as an example to dissuade you from this approach entirely :)

> We could try to add a layer of GNU standards, intercepting e.g. "--pc_type jacobi" and translating it to "-pc_type jacobi" first if we're passing it on to PETSc? But I fear that nobody will use it that way,

Well, presumably our own examples would at least use the options that way? And I think by specifying a small set of cross-solver compatible flags (e.g. make a table of what the libmesh options `--preconditioner`, `--linear-solver` etc. map to in PETSc, Eigen, and other solver packages) then it's more obvious to the user that they are requesting the "current solver package's Jacobi preconditioner" vs. requesting `-pc_type jacobi`, which IMO indicates that they want a specific preconditioner from PETSc only.

roystgnr:
> And I think by specifying a small set of cross-solver compatible flags (e.g. make a table of what the libmesh options --preconditioner, --linear-solver etc. map to in PETSc, Eigen, and other solver packages) then it's more obvious to the user that they are requesting the "current solver package's Jacobi preconditioner" vs. requesting -pc_type jacobi, which IMO indicates that they want a specific preconditioner from PETSc only.

That's a really good idea.  `-pc_type` would leave people wondering which PETSc options we've ported to other solvers and which we haven't; `--pc_type` would just confuse people, but `--preconditioner` would obviously be a general libMesh option.  Okay; whenever I come back to this I'll use that syntax instead.  And instead of translating it into a PETSc CLI option like I'd been imagining with `--pc_type`, I'll just use it to set the PreconditionerType programmatically regardless of solver package; that way we'll get some more test coverage for the programmatic settings, plus more importantly it'll naturally get overridden if `-pc_type` appears too (which IMHO is the right behavior, more specific flags overriding more general ones, even if here "more specific" just means "more package-specific").

lindsayad:
beautiful discussion

PR #3785, opened 2024-02-19
RBParameter vector-valued parameter support - Part2 (final) - Implementation

pmoseley:
This is part 2 (final part), part 1 was here: https://github.com/libMesh/libmesh/pull/3778 .

This adds support in RBParameters for vector-valued parameters. I've changed the underlying datastructure for parameters from `std::map<std::string, std::vector<Real>>` to `std::map<std::string, std::vector<std::vector<Real>>` .

I've tried to maintain compatibility as much as possible by maintaining the old setter/getter functions, which work with the new vectors behind-the-scenes, and generally throw errors if you're trying to use functions expecting single-values on multi-valued data.

There are a few notable breakages of backwards compatibility:
- Cap'n Proto Interfaces
  Because now we need to serialize additional data, the interfaces have changed. However, I've implemented it as a union, so that existing data should be possible to read. Unfortunately, any code which reads/writes a Cap'nProto `RBParameter` needs to be updated.

- Iterators
  Previously we provided custom `.begin()` and `.end()` iterators (and similarly for the "extra parameters") for the RBParameters class. Those iterators iterated over the whole parameter vector, including the steps. Now they've been renamed to `.begin_serialized()` and `.end_serialized()`, and new iterators take the place of the old ones in `.begin()/.end()` which provide a more "standard" map iteration.

  The new iterators are handy for getting parameter names, and also for iterating over the parameters with a better understanding of steps/values. The old "serialized" iterators are occasionally useful when you just want to check all the values without caring where you are in the step/value vector.

### DIFF
--- a/include/reduced_basis/rb_parameters.h
+++ b/include/reduced_basis/rb_parameters.h
@@ -33,14 +33,18 @@ namespace libMesh
 {
 
 /**
- * Typedef for an individual RB parameter.
+ * Typedef for an individual RB parameter. Each parameter is now stored
+ * as a vector of values (different from the vector of samples).
  */
-using RBParameter = Real;
+using RBParameter = std::vector<Real>;
 
 /**
  * This class is part of the rbOOmit framework.
  *
- * This class defines a set of parameters index by strings.
+ * This class defines a set of parameters indexed by strings.
+ * Multiple samples can be defined, where a sample is a set of values for
+ * each parameter. The parameters themselves can be multi-valued, e.g.
+ * for storing a matrix-type parameter.
  *
  * \author David J. Knezevic
  * \date 2012
@@ -75,8 +79,20 @@ public:
   RBParameters(const std::map<std::string, Real> & parameter_map);
 
   /**
-   * Define a constant iterator for this class. This custom iterator
-   * design is copied from the chunked_mapvector class.
+   * Return const_iterators to the internal parameter map, as a convenient
+   * way to access the parameter names and values.
+   * For example: for(const auto & [param_name, sample_vec] : my_parameters)
+   */
+  std::map<std::string,std::vector<RBParameter>>::const_iterator begin() const;
+  std::map<std::string,std::vector<RBParameter>>::const_iterator end() const;
+  std::map<std::string,std::vector<RBParameter>>::const_iterator extra_begin() const;
+  std::map<std::string,std::vector<RBParameter>>::const_iterator extra_end() const;
+
+  /**
+   * Define a constant iterator for iterating over the map of parameters.
+   * This will iterate over every individual value in the map, meaning all
+   * three levels (param name, sample vector, value vector).
+   * This custom iterator design is copied from the chunked_mapvector class.
    */
   class const_iterator
   {
@@ -93,9 +109,11 @@ public:
 
     // Constructor
     const_iterator(const MapIter & it,
-                   const std::size_t vec_index)
+                   const std::size_t sample_vec_index,
+                   const std::size_t value_vec_index)
       : _it(it),
-        _vec_index(vec_index)
+        _sample_vec_index(sample_vec_index),
+        _value_vec_index(value_vec_index)
     {}
 
     // Copy ctor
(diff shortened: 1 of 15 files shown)


src/reduced_basis/rb_eim_construction.C
```diff
@@ -404,8 +406,10 @@ void RBEIMConstruction::set_rb_construction_parameters(unsigned int n_training_s
       const std::string & lookup_table_param_name =
         get_rb_eim_evaluation().get_parametrized_function().lookup_table_param_name;
 
-      std::vector<Real> lookup_table_training_samples(n_training_samples_in);
-      std::iota(lookup_table_training_samples.begin(), lookup_table_training_samples.end(), 0);
+      Real val = 0.0;
+      std::vector<RBParameter> lookup_table_training_samples(n_training_samples_in, {val});
+      for(auto it = lookup_table_training_samples.begin(); it != lookup_table_training_samples.end(); ++it, ++val)
+        (*it)[0] = val;
```
jwpeterson, line 412: I was a bit thrown off seeing that `val` is a `Real` and then seeing the `++val` however I see now that this is also what we were doing before behind the scenes with `std::iota`. Maybe add a comment here just to make it clear what's going on?
```
// Fill lookup_table_training_samples with the sequential single-entry vectors {0.0}, {1.0}, {2.0}, ...
```

src/reduced_basis/rb_parameters.C
```diff
@@ -153,38 +225,84 @@ Real RBParameters::get_extra_value(const std::string & param_name) const
...
+  libmesh_error_msg_if(sample_vec.size() == 0, "Error getting value for extra parameter " << param_name);
   return sample_vec[0];
 }
 
 Real RBParameters::get_extra_value(const std::string & param_name, const Real & default_val) const
 {
   // same as get_value(param_name, default_val) but for the map of extra parameters
   auto it = _extra_parameters.find(param_name);
-  if (it == _parameters.end())
+  if (it == _extra_parameters.end())
```
jwpeterson, line 245: This looks like an actual bugfix, not sure how long that has been wrong...

src/reduced_basis/rb_parametrized.C
```diff
@@ -393,35 +395,45 @@ bool RBParametrized::check_if_valid_params(const RBParameters &params) const
...
+              // For discrete params, make sure params.get_value(param_name) is sufficiently
+              // close to one of the discrete parameter values.
+              // Note that vector-values not yet supported in discrete parameters,
+              // and the .get_sample_value() call will throw an error if the user
+              // tries to do it.
+              if (is_discrete_parameter(param_name))
+                {
+                  const bool is_value_discrete =
+                    is_value_in_list(params.get_sample_value(param_name, sample_idx),
+                                     get_discrete_parameter_values().find(param_name)->second,
```
jwpeterson, line 428: I see that this is just re-indentation, but (as a note to myself) it would be good to avoid this double-lookup idiom by doing:
```
if (auto it = get_discrete_parameter_values().find(param_name);
    it != get_discrete_parameter_values().end())
{
  const bool is_value_discrete =
    is_value_in_list(params.get_sample_value(param_name, sample_idx),
                     it->second,
                     ...
```

jwpeterson (approved):
Looks good, I just had a question about whether we really want to deprecate `get_parameter_names()` since I did try to do that once, but it was just more convenient to have it in the end...

include/reduced_basis/rb_parameters.h
```diff
@@ -300,20 +372,20 @@ class RBParameters
...
    * Fill \p param_names with the names of the parameters.
    *
-   * Note: this function was previously deprecated, but now that
-   * multi-step RBParameters objects are possible, this is actually
-   * the most efficient way to get a list of the parameter names
-   * stored on this object.
+   * \deprecated instead of creating a new set of strings here, it's
+   * better to iterate over the RBParameters object directly, using
+   * the iterators from the begin()/end() functions. The .first
+   * will provide the parameter name, .second can be ignored as needed.
```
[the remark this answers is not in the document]
pmoseley: I found 2 legitimate instances in the Akselos code-base where `get_parameter_names()` could not be trivially improved by using iterators instead of making this new object of string copies.

In those cases, it's a simple replacement to iterate yourself over the map, and populate the set with the results.

I could go either way. I think this function has some limited usefulness, but mostly I want to discourage people from using it unnecessarily.

src/reduced_basis/rb_parameters.C
pmoseley: Looks like I introduced it in my last PR :( I probably made a mistake with cherry-picking/rebasing between the two branches.

So at least it hasn't been a bug very long...

include/reduced_basis/rb_parameters.h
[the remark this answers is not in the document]
pmoseley: I've removed the deprecation comment.

I've also changed the function so it just returns a std::set with copy-elision, rather than taking it as a parameter and clearing it.

src/reduced_basis/rb_eim_construction.C
pmoseley: Yeah, it's a good comment. I've changed the code to be explicit about adding 1.0 instead, and added a comment.

src/reduced_basis/rb_parameters.C
pmoseley: I've split this out into a separate commit, to make it more clear that there was a correction here.

src/reduced_basis/rb_parametrized.C
pmoseley: Good eye, I've gone ahead and updated this.

In general I haven't really touched the discrete parameters.

PR #3787, opened 2024-02-21
Don't duplicate mapping code for xyz

lindsayad:
AFAICT there's not really a strong reason for the XYZ map class to have it's own implementation for `compute_face_map` other than to error for dim=1. I could be wrong though; my untrained eye thinks maybe there was an optimization in there which would make some conceptual sense to me given that XYZ has its points in the physical space. This refs segmentation faults on idaholab/moose#26456 (see https://civet.inl.gov/job/2068980/ for example) which were occurring because the XYZ `compute_face_map` implementation was not doing checking of `if (calculate_foo)`


examples/vector_fe/vector_fe_ex8/vector_fe_ex8.C
```diff
@@ -327,7 +327,7 @@ compute_enriched_soln(const MeshBase & mesh,
   const auto & scalar_phi_face = scalar_fe_face.get_phi();
   const auto & enriched_scalar_phi_face = enriched_scalar_fe_face->get_phi();
   const auto & lambda_phi_face = lambda_fe_face.get_phi();
-  const auto & JxW_face = vector_fe_face.get_JxW();
+  const auto & JxW_face = scalar_fe_face.get_JxW();
```
dschwen, line 330: Why are these changes required?

lindsayad: I did this to make sure the JxW we're getting from the XYZ map is accurate

roystgnr (approved):
This is some seriously old code, isn't it?  `git blame` wants me to go bug Paul Bauman about a commit in 2012, but he was just doing some (major) refactoring of duplication that Ben Kirk put in in the initial FEXYZ commit in 2004!?  Ben's commit message says he "added optimizations for computing edge data in the case of XYZ finite elements", but I don't see us avoiding any virtual function calls or unnecessary work here (to the contrary, the later pre-request optimizations never made it into this copy of the code path...).

Maybe there's some actual optimization here that I'm just missing, but if so we can put it back after somebody else notices it and screams.

PR #3788, opened 2024-02-22
Replace $options variable name in run_common.sh

roystgnr:
This was conflicting with $options in vector_fe_ex8/run.sh, causing options strings there to be appended when they were intended to be replaced.

Is this really the best way to fix global-vs-"local" variable-name conflicts in bash?  It seems to be the fix adopted by everybody's m4 files, so I assume it's the best sh-compatible fix, but you'd think bash extensions would enable *something* better...

Thanks to @lindsayad for catching this.

### DIFF
--- a/examples/run_common.sh
+++ b/examples/run_common.sh
@@ -6,11 +6,11 @@ message_running() {
     shift
     executable=$1
     shift
-    options=$@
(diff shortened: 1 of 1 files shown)


lindsayad (approved):
this is great, thanks Roy

PR #3790, opened 2024-02-27
Respect allow_renumbering(false)

roystgnr:
With an error message, we can find a fix or a workaround for the incompatibility with ExodusII.

With the current mere warning message and forced renumbering, we're liable to break whatever code wasn't compatible with renumbering.

### DIFF
--- a/src/mesh/mesh_output.C
+++ b/src/mesh/mesh_output.C
@@ -52,17 +52,7 @@ void MeshOutput<MT>::write_equation_systems (const std::string & fname,
       // been properly renumbered...
       libmesh_assert(!my_mesh.allow_renumbering());
 
-      libmesh_do_once(libMesh::out <<
-                      "Warning:  This MeshOutput subclass only supports meshes which are contiguously renumbered!"
-                      << std::endl;);
-
-      my_mesh.allow_renumbering(true);
-
-      my_mesh.renumber_nodes_and_elements();
-
-      // Not sure what good going back to false will do here, the
-      // renumbering horses have already left the barn...
-      my_mesh.allow_renumbering(false);
+      libmesh_error_msg("This MeshOutput subclass only supports meshes which are contiguously numbered!");
     }
 
   if (!_is_parallel_format)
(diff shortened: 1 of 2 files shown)


roystgnr:
I was expecting some MOOSE test failures, but I wasn't expecting a MOOSE test failure from a perfectly reasonable use case.  Clearly what we need to be doing in the `allow_renumbering(false)` case for `ExodusII` output in particular is to use the file's `elem_num_map` to record the sparse numbering, after using a virtual function to let `MeshOutput` know that we have that capability.  That'll take longer, but this isn't an urgent update.

src/mesh/mesh_output.C
```diff
@@ -52,17 +52,7 @@ void MeshOutput<MT>::write_equation_systems (const std::string & fname,
...
-                      << std::endl;);
-
-      my_mesh.allow_renumbering(true);
-
-      my_mesh.renumber_nodes_and_elements();
-
-      // Not sure what good going back to false will do here, the
-      // renumbering horses have already left the barn...
-      my_mesh.allow_renumbering(false);
+      libmesh_error_msg("This MeshOutput subclass only supports meshes which are contiguously numbered!");
```
jwpeterson, line 55: Hmm... I think in our internal code, we may be relying on being allowed to get away with renumbering even when a mesh is set to `allow_renumbering(false)`. At least, I see the "This MeshOutput class only supports..." warnings in our logfiles. I think it happens right before we do exodus output, so I'm worried that if you remove this "hack" we'll attempt to write exodus files with non-contiguous numbering (from stitched meshes) and then that will fail? I'd like to try out this change and see if it breaks any of our tests...

jwpeterson: OK, while I was composing that beautiful paragraph of prose, it looks like you closed the PR :smile:

roystgnr: I'd say do try it out, because the whole reason for this PR is that I ran across a bug where a problem was setting `allow_renumbering(false)` and *actually really meant it* ... but as per my other comment, I think the real solution here is going to be to finally just support ExodusII output of meshes with sparse numbering.

PR #3791, opened 2024-02-27
GMG doesn't require these restrictions

roystgnr:
Or at least, I can't find any reason why it should require them (even when we coarsen to create projection operators we're guaranteed our parent elements won't be remote), it's been working without some of them accidentally (ExodusII output has been forcing renumbering), and it's passing test sweeps I run manually.

I've pulled this commit out of #3790; hopefully it'll fare better than that PR as a whole did.


roystgnr:
Test MOOSE failure is unrelated

PR #3793, opened 2024-03-02
Netgen submodule

roystgnr:
This isn't in working order (I'm sure it'll fail installcheck, most of the features aren't there...) but it's passing the most basic test now and probably good enough to let others see it and to make sure I haven't regressed anything in other tests.


roystgnr:
Well, at least I didn't break installcheck in the default --disable-netgen configuration; that's a plus.  Let me try enabling netgen by default and make sure CI catches the issues then.

roystgnr:
Okay, *now* we're seeing the real issues; enough to prompt cleanup of the cmake options and a stray clang warning, at least.  And I think a proper manual install (with patchelf on Linux, install_name_tool on OSX, thankfully this was already figured out in moose.mk at one point...) will fix the remaining issues.

roystgnr:
The Mac failure is the same eigenproblems_ex4 issue that's been hitting all libMesh PRs since the environment update.

roystgnr:
Likewise for the umat issue in Test MOOSE mac

jwpeterson (approved):

PR #3795, opened 2024-03-05
More read_*() options for PetscMatrix

roystgnr:
I want to add one more unit test before merging this, but first I want to see just how inflexible the PETSCVIEWERBINARY format is - it breaks when I move from 32-bit to 64-bit; does it also break from endianness issues when going from Linux/Intel to Mac/ARM?


roystgnr:
There's definitely something going wonky with my PETSc's HDF5/MatView interaction; I get weird header-only data-free output not only from the write code here but also even when doing something as simple as `-mat_view hdf5:introduction_ex3.hdf5`.  So that's going to be going untested for now.  The rest of this should be ready to go; I'll merge in a few days unless anyone has comments.

PR #3801, opened 2024-03-13
Update printing for vector-valued RBParameters

pmoseley:
Now for vector-valued RBParameters, by default we only print 5 values then an ellipsis.
Otherwise, printing the parameter values can result in a lot of output. Users can still specify to print all values by setting max_values to a negative number.


dknez:
This is a minor change related to how we print info during RB training, and has passed CI, so I'll go ahead and merge.

PR #3806, opened 2024-03-19
Handle the case where we encounter linearly dependent data in EIM

dknez:
In this case we should call the enrichment function again without asking to add a basis function.


jwpeterson (approved):
This is a slight API update, but it should be backwards compatible since it is just adding a return type where there was none before.

jwpeterson:
The failures here are not related to this PR, see https://github.com/idaholab/moose/pull/26588 for further discussion.

PR #3807, opened 2024-03-19
Don't try to interpolate Dirichlet values on non-Lagrange non-vertices

roystgnr:
In the best case we notice that the non-vertex hosts multiple DoFs and we fail an assertion; in the worst case just setting a coefficient value to a function value would give garbage here.

I've never tripped this with libMesh/GRINS/etc codes, but MOOSE hits it as soon as I try adding a DirichletBoundary there, since MOOSE generates nodesets from sidesets.

Pinging @dknez, since this basically reverts some of his #672 changes.  Fortunately, in the time since that PR we tried to do some optimization by splitting apart Lagrange from non-Lagrange code paths, and we only have to revert the broken non-Lagrange case, which shouldn't change the behavior Akselos is depending on in the Lagrange case.

It wouldn't be crazy to support the non-Lagrange case (we'd end up pinning one linear combination of the DoFs at a node in a Dirichlet nodeset that wasn't part of a side in a Dirichlet sideset) but I don't have time or a use case for that right now.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -871,10 +871,13 @@ private:
     // hold those fixed and project boundary edges, then hold
     // those fixed and project boundary faces,
 
-    // Interpolate node values first. Note that we have a special
-    // case for nodes that have a boundary nodeset, since we do
-    // need to interpolate them directly, even if they're non-vertex
-    // nodes.
+    // Interpolate node values first.
+    //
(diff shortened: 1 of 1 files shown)


dknez:
Since this only affects non-Lagrange cases, this should be fine, thanks. We'll let you know if we encounter any issues with this change, though.

roystgnr:
Test failures are unrelated; I'll merge this now, but do let me know if you find cause to revert it.

PR #3809, opened 2024-03-19
Wrap `_local_size` into atomic to avoid unwanted modifications while `operator()` is used with `set()` in petsc vector.

### LINKED ISSUE #3808: Operator `()` of a petsc vector is not safe if used together with `set()` in a threaded loop.
If this is used in a threaded loop:
```
const auto normalized_value = (*grad_container[dim_index])(dof_id_elem) / volume;
grad_container[dim_index]->set(dof_id_elem, normalized_value);
```
We get the following error:
```
libMesh terminating:
Assertion `local_index < _local_size' failed.
local_index = 24
_local_size = 0
```
which comes from the fact that the `set` function can set `_local_size` to 0.

grmnptr:
Closes #3808

### DIFF
--- a/tests/numerics/petsc_vector_test.C
+++ b/tests/numerics/petsc_vector_test.C
@@ -3,11 +3,18 @@
 #ifdef LIBMESH_HAVE_PETSC
 
 #include "numeric_vector_test.h"
+#include <libmesh/equation_systems.h>
+#include <libmesh/mesh_generation.h>
+#include <libmesh/linear_implicit_system.h>
+#include <libmesh/mesh.h>
+#include <libmesh/dof_map.h>
+#include <libmesh/elem.h>
 
 
 using namespace libMesh;
 
 class PetscVectorTest : public NumericVectorTest<PetscVector<Number>> {
+
 public:
   PetscVectorTest() :
     NumericVectorTest<PetscVector<Number>>() {
@@ -24,6 +31,8 @@ public:
   CPPUNIT_TEST( testGetArray );
 
   CPPUNIT_TEST( testPetscOperations );
+
+  CPPUNIT_TEST( testPetscVectorGetSetThread );
 
   CPPUNIT_TEST_SUITE_END();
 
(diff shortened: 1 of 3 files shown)


roystgnr:
I'm scared of whether these are the *only* fixes we need, but if this fixes your MOOSE case then it's definitely a step in the right direction.

Could you add something like your MOOSE failure case to a new unit test in `tests/numerics/numeric_vector_test.h`?  Threading bugs are annoyingly hard to reproduce and if there's a chance of regression then I'd hope to hit it as far upstream as possible.  I don't see any good threaded unit tests to point you to for an example, but `git grep Threads::parallel_for` points to a dozen-odd use cases in the library itself.

grmnptr:
I can confirm that this change fixes the issue, based on 100 runs with 10 threads. The issue usually pooped up in 5-10 runs with this many threads.

jwpeterson:
I'm curious how you figured out that this was the fix, for example did you use a debugging tool that suggested it? Also, you made _first and _last std::atomic but we don't .load() or store them anywhere?

Finally, general question: is this (declaring a variable atomic) something that needs to be done differently depending on if TBB is used vs. std::threads?

grmnptr:
To be honest I am not a threading expert, but I'll try to answer your questions as accurately as I can. 
I got the following error from libmesh in debug mode:

```
libMesh terminating:
Assertion `local_index < _local_size' failed.
local_index = 24
_local_size = 0
```

The debugger pointed to the `operator()` line discussed in #3808 . 
In `set()` , `_local_size` is set to 0 (through `restore_array`) and that can be the source of the problem. To be honest, @roystgnr pointed these out and I just followed his suggestions.

The reason I used `load()` is because `make_range()` has a `T range(const T & range)` underneath and this constructor has been deleted for atomic-wrapped quantities.

I have no idea if this would change based on different threading libraries, I was wrapping the member variables in atomic based on the examples on the c++ standard website so I hope oneTBB respects that.

I am also open to other suggestions for assigning a value to an already existing petsc vector that doesn't rely on other operations which manipulate the size.

grmnptr:
I created a test for this and it exhibits out of memory access in the petsc vector. It is not deterministic, so I think this change did not solve all the problems. The new error is coming from 
```
template <typename T>
inline
T PetscVector<T>::operator() (const numeric_index_type i) const
{
  this->_get_array(true);

  const numeric_index_type local_index = this->map_global_to_local_index(i);

#ifndef NDEBUG
  if (this->type() == GHOSTED)
    {
      libmesh_assert_less (local_index, _local_size);
    }
#endif

  return static_cast<T>(_read_only_values[local_index]); <<<< This line
}
```
with 
```
(gdb) p _read_only_values
$1 = (const PetscScalar *) 0x0
```
In restore array, this gets set to nullptr too. I am not sure I can pull an atomic here.

tests/numerics/petsc_vector_test.C
```diff
@@ -129,6 +138,89 @@ public:
...
+    Threads::parallel_for(active_local_elem_range, test_thread);
+
+    sys.solution->close();
+
+    const libMesh::dof_id_type first = sys.solution->first_local_index();
+    const libMesh::dof_id_type last  = sys.solution->last_local_index();
+
+    for (libMesh::dof_id_type n=first; n != last; n++)
+      LIBMESH_ASSERT_FP_EQUAL(libMesh::libmesh_real((*sys.solution)(n)),
+                              libMesh::Real(1.0),
```
lindsayad, line 182: should this be 0.5?

grmnptr: Nice, thanks! I was altering things in there and forgot to reset it back to 0.5.

roystgnr:
> a debugging tool that suggested it

I prefer the phrase "a debugging assistant", thank you very much!

But seriously, this whole section of code terrifies me; I don't understand the C++ memory-order standards well at all.  I tried to figure out all the variables that might get caught out by a read-only get-array transitioning to a read-write one, but aside from the miss we've already caught, I'm not even 100% certain `std::atomic` is a guaranteed fix; one reason I was begging for a unit test was so we'd get a chance to see if OSX and Linux behave differently.

I'm not yet sure what to do about `_read_only_values`.

lindsayad:
Sigh, we'll have to dig into the double checked locking pattern again. With this PR, I still get these errors out of thread sanitizer on the MOOSE tests (using grmnptr/moose@63f614deae)

grmnptr:
Thanks for running the sanitizer on the PR. I disabled threading on that new capability for the time being.

grmnptr:
I'll close this PR for now. I don't think I'll dive deeper, threading is not yet a requirement for the capability I am working on.

PR #3810, opened 2024-03-19
Fix p=2 Lagrange nodal_soln on Tri7-sided elems

roystgnr:
I must have missed this because I still think of those elements as just
ways to host Hierarchic bases, and although they need Lagrange support
for mappings, we only hit nodal_soln when decimating solution variables
for visualization.

This fixes (and tests the fix for) the bug in https://github.com/libMesh/libmesh/discussions/3794#discussioncomment-8843171


jwpeterson (approved):

roystgnr:
CI failures are unrelated.

PR #3811, opened 2024-03-20
Edge3 compatibility with poly2tri

roystgnr:
Pinging @miaoyinb and @eshemon, both to report this and to ask a couple questions at the end.

The results of my preliminary investigation into poly2tri's interaction with quadratic elements:

Bad news: There are definitely a number of real bugs preventing our triangulators from generating output based on higher-order boundary and hole elements.

Good news: Getting output instead of bugs doesn't look to be very hard; it's easy enough to decimate boundary data before passing it along to poly2tri.  This PR does that for the case of Edge3, and it took me an hour or two.  Handling Tri6/Tri7/Quad8/Quad9 will be harder but I'd guess a day's work.

Bad news: Decimating boundary data decimates it.  To get a geometrically-correct Tri6 triangulation from our output here, which for your use cases is kind of the whole point, we'll need to not only add a Tri3->Tri6 conversion (easy) but also snap our mid-edge nodes to the geometry defined by the original mid-edge nodes which, at that point, *are already deleted* by the decimation.

Good news: Preserving that data and doing the snapping afterward ... won't be an hour and a half job, but it will be doable without too much work, cleanly enough for my tastes.  I ought to be able to squeeze this in to my meshing time in the next few months.

Bad news: The more sophisticated we want to make this, the harder it's going to be to do.  "Snap existing boundary nodes to the correct geometry after all the interior geometry is done" will be some work.  "Snap all boundary nodes afterward, because we want to be able to refine the boundaries" will be harder.  "Snap all boundary nodes as soon as each boundary edge is refined, because we don't want the interior elements to be too stretched near the boundary" will be harder still.

How useful are each of those levels of sophistication for you?  To put it into XYDelaunay terms, it's all about `refine_boundary` and `refine_holes` options.  Level 0 is "if you use quadratic elements, the triangulated domain borders can't even be curved" - I can't imagine that being useful to you but if I'm wrong let me know and I'll knock it out ASAP.  Level 1 is "if you use quadratic elements, you can't refine borders".  Level 2 is "if you use quadratic elements and you refine curved borders, elements near the border will be stretched".  Level 3 is "this all actually works the way it should".

### DIFF
--- a/src/mesh/triangulator_interface.C
+++ b/src/mesh/triangulator_interface.C
@@ -111,26 +111,43 @@ void TriangulatorInterface::elems_to_segments()
           point_id_map.emplace(*node, node->id());
         }
 
+      // We don't support directly generating Tri6, so for
+      // compatibility with future stitching we need to be working
+      // with first-order elements.  Let's get rid of any non-vertex
+      // nodes we just added.
+      for (Elem * elem : _mesh.element_ptr_range())
+        for (auto n : make_range(elem->n_vertices(), elem->n_nodes()))
+          point_id_map.erase(elem->point(n));
+
       // We'll steal the ordering calculation from
       // the MeshedHole code
       const TriangulatorInterface::MeshedHole mh { _mesh, this->_bdy_ids };
-
-      // And now we're done with elements.  Delete them lest they have
-      // dangling pointers to nodes we'll be deleting.
-      _mesh.clear_elems();
 
       // If we've specified only a subset of the mesh as our outer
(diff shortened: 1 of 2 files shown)


miaoyinb:
Thank you @roystgnr. 

In my opinion,  Level 1 looks reasonable. Usually we want quadratic elements generated by `XYDelaunay` because we want to stitch the hole meshes with the mesh generated by `XYDelaunay`.

@eshemon Any comments?

roystgnr:
Downstream failures here are unrelated.

PR #3813, opened 2024-03-26
Adding Auto Area Function for Poly2Tri

miaoyinb:
Thinking about adding an automatically generated element area function for poly2tri that can be called by `XYDelaunay` in MOOSE. The auto function is based on inverse Distance Interpolation of the EDGE elements sizes of the outer boundary as well as the holes boundaries.

I also modified the interpolation a little bit to add a "background" value and the corresponding effective distance so that a default area value can be used in those areas away from boundaries.

### DIFF
--- a/include/mesh/triangulator_interface.h
+++ b/include/mesh/triangulator_interface.h
@@ -254,11 +254,26 @@ public:
 
   /**
   *  Generate an auto area function based on spacing of boundary points.
+  *  The external boundary as well as the hole boundaries are taken into consideration
+  *  to generate the auto area function based on inverse distance interpolation.
+  *  For each EDGE element on these boundaries, its centroid (midpoint) is used as the point
+  *  position and the desired area is calculated as 1.5 times of the area of the equilateral
+  *  triangle with the edge length as the length of the EDGE element.
+  *  For a given position, the inverse distance interpolation only considers a number of nearest
+  *  points (set by num_nearest_pts) to calculate the desired area. The weight of the value at
+  *  each point is calculated as 1/distance^power.
+  * 
+  *  In addition to these conventional inverse distance interpolation features, a concept of
+  *  "background value" and "background effective distance" is introduced. The background value
+  *  belongs to a virtual point located at a constant distance (background effective distance)
+  *  from the given position. The weight of the value at this virtual point is calculated as
+  *  1/background_effective_distance^power. Effectively, the background value is the value when
+  *  the given position is far away from the boundary points.
   */
   void generate_auto_area_function(const Parallel::Communicator &comm,
                                    const unsigned int num_nearest_pts,
                                    const unsigned int power,
-                                   const Number background_value,
+                                   const Real background_value,
                                    const Real  background_eff_dist);
 
   /**
--- a/include/mesh/triangulator_interface.h
+++ b/include/mesh/triangulator_interface.h
@@ -260,12 +260,12 @@ public:
                                    const unsigned int power,
                                    const Number background_value,
                                    const Real  background_eff_dist);
-  
+
   /**
   *  Whether or not an auto area function has been generated.
   */
   bool has_auto_area_function() {return _auto_area_function != nullptr;}
-  
+
   /**
   *  Caluclate the local desired area based on the auto area function.
   */
--- a/include/mesh/triangulator_interface.h
+++ b/include/mesh/triangulator_interface.h
@@ -26,7 +26,6 @@
 #include "libmesh/libmesh.h"
 #include "libmesh/point.h"
 
-#include "libmesh/meshfree_interpolation.h"
 #include "libmesh/function_base.h"
 
 // C++ includes
@@ -38,6 +37,8 @@ namespace libMesh
 
 // Forward Declarations
 class UnstructuredMesh;
+template <unsigned int KDDim>
+class InverseDistanceInterpolation;
 enum ElemType : int;
 
 // Helper function for auto area calculation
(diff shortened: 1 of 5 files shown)


miaoyinb:
@roystgnr If the idea makes sense, I can formalize things, remove the draft label, and also submit the MOOSE PR.

@eshemon FYI

miaoyinb:
areaFunc

An example using the auto function.

include/solution_transfer/meshfree_interpolation.h
```diff
@@ -330,13 +332,17 @@ class InverseDistanceInterpolation : public MeshfreeInterpolation
    */
   InverseDistanceInterpolation (const libMesh::Parallel::Communicator & comm_in,
                                 const unsigned int n_interp_pts = 8,
-                                const Real  power               = 2) :
+                                const Real  power               = 2,
+                                const Number background_value   = 0.,
```
roystgnr, line 336: We also need more documentation here, for the benefit of users who aren't using it indirectly via a triangulator.

src/mesh/poly2tri_triangulator.C
```diff
@@ -1500,7 +1502,8 @@ bool Poly2TriTriangulator::should_refine_elem(Elem & elem)
   // vertices first
   for (auto v : make_range(elem.n_vertices()))
     {
-      const Real local_area_target = (*area_func)(elem.point(v));
+      // If we have an auto area function, we'll use it and override other area options
+      const Real local_area_target = this->has_auto_area_function() ? this->get_auto_desired_area(elem.point(v)) : (*area_func)(elem.point(v));
```
roystgnr, line 1506: It's still taking a `Point` as input and outputting a scalar as output, so couldn't we encapsulate the interpolation in a `FunctionBase` subclass rather than in a member function here?  The extra class definition would be a little bit more boilerplate but it would also eliminate half of this diff.

src/solution_transfer/meshfree_interpolation.C
```diff
@@ -281,9 +281,21 @@ void InverseDistanceInterpolation<KDDim>::interpolate (const Point
...
+  _vals.resize(n_fv);
+
+  std::fill (_vals.begin(), _vals.end(), Number(0.));
+  Real tot_weight(0.);
+  // The background value is optional
+  // If background value option is enabled, add it to the total weight
+  // If not, a "zero" weight is added
+  if (_background_eff_dist > 0.0)
+  {
+    const Number background_wt = _background_eff_dist * _background_eff_dist > std::numeric_limits<Real>::epsilon()
```
roystgnr, line 293: We'd normally use some power of `TOLERANCE` here.

src/mesh/triangulator_interface.C
```diff
@@ -314,5 +315,67 @@ unsigned int TriangulatorInterface::total_hole_points()
...
 }
 
+void TriangulatorInterface::generate_auto_area_function(const Parallel::Communicator &comm,
+                                                        const unsigned int num_nearest_pts,
+                                                        const unsigned int power,
+                                                        const Number background_value,
+                                                        const Real  background_eff_dist)
+{
+  // Get the hole mesh of the outer boundary
+  // Holes should already be attached if applicable when this function is called
```
roystgnr, line 325: That's more fragile than it needs to be, for a function for users to call.  Let the users call something (I'd still push for `set_desired_area_function(OurNewSubclass(its, args))` to set up the function, then we can call its `init()` ourselves from within `triangulate()` and not rely on them getting ordering correct.

miaoyinb:
Thanks for the comments @roystgnr  I will work on this early next week.

include/mesh/triangulator_interface.h
```diff
@@ -241,6 +243,25 @@ class TriangulatorInterface
    */
   void attach_region_list(const std::vector<Region*> * regions) { _regions = regions; }
 
+  /**
+  *  Generate an auto area function based on spacing of boundary points.
```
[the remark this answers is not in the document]
miaoyinb: I have moved things into a Function. I do not think we need `std::complex` support for this function but get the related CIVER test failures. Any suggestion about best practice to disable `std::complex` for this? Thanks.

jwpeterson:
`MeshfreeInterpolation get_source_vals()` returns a writeable reference to a `std::vector<Number>`. When libMesh is compiled with complex variables enabled, `Number == std::complex`. To fix this error
```
./include/libmesh/triangulator_interface.h: In member function 'void libMesh::AutoAreaFunction::init_mfi(const std::vector<libMesh::Point>&, const std::vector<double>&)':
./include/libmesh/triangulator_interface.h:76:41: error: no match for 'operator=' (operand types are 'std::vector<std::complex<double> >' and 'const std::vector<double>')
     _auto_area_mfi->get_source_vals() = input_vals;
                                         ^~~~~~~~~~
``` 
the `input_vals` would also need to be a `std::vector<Number>` to conform to the interface. In your case, you would just be using `Numbers` with zero imaginary part.

roystgnr:
The mac test failure is unrelated.  I'll try to re-review this tomorrow afternoon or Thursday.

src/solution_transfer/meshfree_interpolation.C
miaoyinb: I was trying to prevent doing 1/0. Maybe it's better to put a `libmesh_assert` here.

miaoyinb:
@roystgnr Please take a look at this PR when convenient and let me know if any additional revisions are needed. Thanks

src/solution_transfer/meshfree_interpolation.C
roystgnr: That makes sense.  An assert wouldn't; it's not unreasonable to want to "interpolate" at zero distance.

include/mesh/triangulator_interface.h
```diff
@@ -73,25 +73,36 @@ class AutoAreaFunction : public FunctionBase<Real>
...
     target_pts.push_back(p);
     target_vals.resize(1);
 
     _auto_area_mfi->interpolate_field_data(_auto_area_mfi->field_variables(), target_pts, target_vals);
 
+#ifdef LIBMESH_USE_COMPLEX_NUMBERS
+    return target_vals.front().real();
+#else
     return target_vals.front();
+#endif
```
roystgnr, line 105: We have a `libmesh_real(T)` function that ought to let us turn these 5 lines into 1.

roystgnr (approved):
I'd appreciate it if you could fix the header inclusion and the `libmesh_real()` use, but if either of those give you trouble then IMHO this is good enough to merge as-is; those are the sort of minor things I fix every few years anyway, in frenzies of sorting through `grep` results or other automated sweeps.

include/mesh/triangulator_interface.h
miaoyinb: I will definitely fix the `libmesh_real(T)` use today.

Let me try to fix header inclusion (I thought it was no longer needed after using the FunctionBase approach). If I cannot fix it by Monday, may be we can merge this first so that the MOOSE part of this feature can be reviewed and merged soon. Thanks.

roystgnr: Agreed.  I'm going to be on leave for much of Monday but I'll try to keep an eye out for updates here; I should have time to check around noon.

The header inclusion is only a mild annoyance, but it should be easy to fix - just move the destructor definition (even "= default" counts as a definition) into the .C file, leaving only a declaration in the header, and then we can remove that include and replace it with a forward declaration.

roystgnr: Oh!  I should have realized you had other functions that would require more than a forward declaration.  The constructor had to be moved because `std::make_unique` needs to see the constructor of the class inside the `unique_ptr`.

IMHO those functions are complicated enough to belong in the .C file anyway.  Thanks!

roystgnr (approved):

PR #3814, opened 2024-03-28
FEMContext::hardest_fe_type() fix

roystgnr:
Starting with var 0 even when we have an `active_vars` that doesn't include it can cause us to use too many quadrature points, which can be wasteful in some cases and a disaster in others.  See https://github.com/idaholab/moose/issues/26605

### DIFF
--- a/src/systems/fem_context.C
+++ b/src/systems/fem_context.C
@@ -68,6 +68,7 @@ FEMContext::FEMContext (const System & sys,
 {
   if (active_vars)
     {
+      libmesh_assert(!active_vars->empty());
       auto vars_copy =
         std::make_unique<std::vector<unsigned int>>(*active_vars);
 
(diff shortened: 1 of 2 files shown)


roystgnr:
Pushed this to PR immediately because I want to see what regression tests in MOOSE need regolding even if it works; fingers crossed that it works.

roystgnr:
Code never respects the fingers.

roystgnr:
We can merge this regardless of the failing tests, but I'm seeing some weird behavior with (enough levels of) adaptive p refinement when running locally; I'd like to get that figured out before this goes in.

roystgnr:
The weirdness is predating these commits, and we're not seeing it in CI, so I'll keep looking into this but I won't let it delay this merge.

I got pushback on temporarily disabling the broken MOOSE tests, so those'll be broken until we get this to master and the submodule update to MOOSE.

PR #3815, opened 2024-04-01
We require C++17 now, so std::map::insert_or_assign() is allowed

jwpeterson:
Simple PR to test that devel -> master merges are working again.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -380,8 +380,8 @@ private:
           bool do_this_side = false;
           for (const auto & bc_id : ids_vec)
             {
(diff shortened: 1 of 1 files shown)


roystgnr (approved):
My `README` commit verified that the updated devel->master CI recipes are working fine, so we don't need this urgently, but on the other hand it's a good change and the only failures we're seeing from PR CI are the expected-until-submodule-update ones.

jwpeterson:
OK, well maybe we can wait until those tests are fixed and we need to test another devel -> master merge? Then I'll invalidate the failed builds here and we can merge if everything's passing.

PR #3816, opened 2024-04-02
Quadratic Elem support in TriangulatorInterface

roystgnr:
This is enough to let us mix and match lower- and higher-order elements in boundary and hole definitions for the same triangulation, and get a triangulation output that matches the higher-order midpoint nodes where they're available.

This should be everything we need on the libMesh end to support what I called "level 1" in https://github.com/libMesh/libmesh/pull/3811

### DIFF
--- a/include/mesh/mesh_triangle_holes.h
+++ b/include/mesh/mesh_triangle_holes.h
@@ -64,9 +64,22 @@ public:
   virtual unsigned int n_points() const = 0;
 
   /**
+   * The number of geometric midpoints along *each* of the sides
+   * defining the hole.
+   */
+  virtual unsigned int n_midpoints() const { return 0; }
+
+  /**
    * Return the nth point defining the hole.
    */
   virtual Point point(const unsigned int n) const = 0;
+
+  /**
+   * Return the midpoint \p m along the side \p n defining the hole.
+   */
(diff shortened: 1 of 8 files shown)


roystgnr:
Broken in parallel, huh?

jwpeterson (approved):
Looks good to me aside from whatever parallel issue you are tracking down

roystgnr:
I rebased so it's hard to see, but the parallel fix was one line; I forgot that I had split a broadcast() into two places and when I had to add a `broadcast(_midpoints)` I forgot to add it to both.

Kind of an awkward code decision, actually.  I did:
```
if (pid != 0)
{
  broadcast(foo);
  return();
}

hundreds_of_lines();
of_pid_0_stuff();

broadcast(foo);
```

... pretty much solely because I didn't want to indent `hundreds_of_lines()` inside an `if (pid == 0)` block or create a separate `hundreds_of_lines_of_pid_0_stuff()` function.  I think my subconscious heuristic might be "code with smaller `wc -c` output is better" and I think it might have failed me here...

roystgnr:
Okay; I'll merge this once "Test 32bit" is done so we can get into the submodule update ASAP.

roystgnr:
It's done with unit tests; the examples don't cover this code.

PR #3818, opened 2024-04-02
Update to plotting EIM variables

dknez:
Updates to enable more than one EIM variable to be plotted. We specify variables to be plotted via the new `EimVargroupPlottingInfo` struct.


jwpeterson (approved):

jwpeterson:
Failures are due to known failing MOOSE tests.

PR #3821, opened 2024-04-04
Increased functionality for CondensedEigenSystem

### LINKED ISSUE #3585: Add shell matrix handling for CondensedEigenSystem
This would be a precursor to allowing MOOSE to switch from `EigenSystem` to `CondensedEigenSystem`, which MOOSE would need to do if it wants to be able to perform eigen solves with CFEM with hanging nodes, etc.

lindsayad:
Closes #3585 and refs [idaholab/moose](https://github.com/idaholab/moose/issues/27262). 

As one piece of supporting evidence that this works (at least to some degree), I re-ran the neutron diffusion test case @YaqiWang gave me back in June of last year. We failed utterly with hanging nodes back then as shown in https://github.com/libMesh/libmesh/pull/3584#issuecomment-1590200529. However, it looks good now as shown below

image

I will try @pbalest's test case soon for further verification


lindsayad:
Accompanying MOOSE PR at https://github.com/idaholab/moose/pull/27276

src/numerics/petsc_vector.C
```diff
@@ -1384,8 +1384,46 @@ void PetscVector<T>::_restore_array() const
...
+  WrappedPetsc<IS> parent_is;
+  auto ierr = ISCreateGeneral(this->comm().get(),
+                              cast_int<PetscInt>(rows.size()),
+                              numeric_petsc_cast(rows.data()),
+                              PETSC_USE_POINTER,
+                              parent_is.get());
+  LIBMESH_CHKERR(ierr);
 
+  Vec subvec;
+  ierr = VecGetSubVector(_vec, parent_is, &subvec);
```
YaqiWang, line 1401: We allocate memory for storing this sub vector here, I guess?

include/numerics/petsc_matrix.h
```diff
@@ -124,10 +124,12 @@ class PetscMatrix final : public SparseMatrix<T>
    */
   PetscMatrix (PetscMatrix &&) = delete;
   PetscMatrix (const PetscMatrix &) = delete;
-  PetscMatrix & operator= (const PetscMatrix &) = delete;
   PetscMatrix & operator= (PetscMatrix &&) = delete;
   virtual ~PetscMatrix ();
 
+  PetscMatrix & operator= (const PetscMatrix &);
+  virtual SparseMatrix<T> & operator= (const SparseMatrix<T> & v) override;
+
```
jwpeterson, line 132: I know it's not really the main point of the PR, but since we are using virtual assignment operators here, I was wondering if you had [read up on them in detail](https://stackoverflow.com/questions/669818/virtual-assignment-operator-c)? Seems like it might be tricky to get exactly correct.

jwpeterson (commented):
We use the `CondensedEigenSystem` in our code, so I'd like to try out this branch with some of our internal tests, just to make sure we get the same answer before/after the changes.

jwpeterson:
In case it's relevant, we are using a slightly older PETSc (`v3.17.4`) than what you may be testing with.

src/numerics/petsc_vector.C
lindsayad: It may or it may not. Sometimes PETSc provides a direct view into the memory of the existing vector in which case there is no new memory allocation

lindsayad:
Thanks @jwpeterson for looking into this. I have not looked into all the test failures here yet. Maybe once I do I will also fix the failures in your internal testing. I think I jumped the gun bit here on not initially marking this as draft while seeing what CI said

lindsayad:
I believe failures are unrelated so I think this is ready for some review

lindsayad:
I should try to add some testing

jwpeterson:
@lindsayad I can pull your branch and try it again with our internal tests... unless you think I should wait?

lindsayad:
I think it's ready for you to test

jwpeterson:
> I think it's ready for you to test

OK, the tests that were previously giving that `New nonzero at (0,0)` error on this branch are now working again. :+1:

include/systems/condensed_eigen_system.h
```diff
@@ -152,6 +152,12 @@ class CondensedEigenSystem : public EigenSystem
    */
   void copy_super_to_sub(const SparseMatrix<Number> & super, SparseMatrix<Number> & sub);
 
+  /**
+   * Instructs not to create the condensed submatrices from the global matrices in the solve
```
roystgnr, line 156: Do we definitely understand *why* we might need this?  Is it's just because it forces us to `close()` earlier and thereby trigger PETSc's "you gave us a sparsity pattern but since we see zeroes we're going to assume you gave us the wrong sparsity pattern and shrink it" policy?

If so, some documentation of that here would be good.

Also if so, is there still no better workaround at this point?  Feels like someone should have eventually added a `PetscDidIStutter(false)` we could be calling...

roystgnr (changes requested):
No non-documentation-related complaints here.

src/systems/condensed_eigen_system.C
```diff
@@ -39,6 +39,7 @@ CondensedEigenSystem::CondensedEigenSystem (EquationSystems & es,
   : Parent(es, name_in, number_in),
     condensed_matrix_A(&this->add_matrix("Condensed Eigen Matrix A")),
     condensed_matrix_B(&this->add_matrix("Condensed Eigen Matrix B")),
+    condensed_precond_matrix(&this->add_matrix("Condensed Eigen Preconditioner")),
```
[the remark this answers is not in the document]
lindsayad: I was just following the pattern that existed before (lines 40, 41). I can change the behavior for all

include/systems/condensed_eigen_system.h
lindsayad: The reason to not create submatrices in the solve is that the submatrix creation will create an entirely new matrix and consequently wipe away the callbacks I had attached to the condensed matrix. SLEPc will not be able to communicate with us if that is the case.

Also it seems that all previous users of this class (the RB code) have already assembled their system before handing off to the solver so `create_submatrix` is actually meaningful for those users. We (MOOSE) generally do not do assembly until we are within the solve so `create_submatrix` is truly useless. It's creating a submatrix from a matrix with no fill

lindsayad:
This should be ready for more review

roystgnr (approved):

PR #3825, opened 2024-04-08
Don't do nontrivial calculations just in DEBUG for Slepc solver

lindsayad:
This removes the printing of eigenvalues and associated computation of errors from the `solve` method for `SlepcEigenSolver`, which was previously only occurring for `dbg` mode. Instead I've added a `print_eigenvalues` routine that can be used to print eigenvalues, not attached to any given compilation method. This code necessitated assembly of the B matrix for the inverse power method in MOOSE when it is not strictly needed

### DIFF
--- a/include/solvers/slepc_eigen_solver.h
+++ b/include/solvers/slepc_eigen_solver.h
@@ -256,6 +256,11 @@ public:
    */
   EPS eps() { this->init(); return _eps; }
 
+  /**
+   * Print the eigenvalues and associated error
+   */
+  void print_eigenvalues() const;
+
 private:
 
   /**
--- a/src/solvers/slepc_eigen_solver.C
+++ b/src/solvers/slepc_eigen_solver.C
@@ -508,14 +508,6 @@ SlepcEigenSolver<T>::_solve_helper(Mat precond,
   PetscInt its=0;
   ST st=nullptr;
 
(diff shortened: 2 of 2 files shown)


lindsayad:
Associated MOOSE change that this PR would enable: https://github.com/idaholab/moose/pull/27323/commits/dc8cd3280aea7ed0c2dfc6ee4b4d1a4ceb7fa4ed

lindsayad:
I can also drop the `print_eigenvalues` method entirely. There is only one example that casts to `SlepcEigenSolver` but that requires both SLEPc and IFEM and I don't think we have any recipe that does Slepc and IFEM, at least not for devel coverage testing. Soooo, my next thought would be to make a virtual `print_eigenvalues` method in the base class but I have to admit I'm not particularly motivated to do that

roystgnr (approved):

jwpeterson:
Our "Test complex and infinite" recipe (always thought that was an amazing name, btw) doesn't even test with PETSc enabled, let alone SLEPc. We use complex-valued PETSc and SLEPc in our internal testing so I'm confident that it's working, but I'm surprised we don't get more regressions given that I don't think we have any test recipes with this configuration.

lindsayad:
We don't get any regressions because we only write beautiful code

PR #3827, opened 2024-04-12
Add gzip option to splitter app

roystgnr:
I added this while doing some testing, and it might be useful to others too.


roystgnr:
Merging; there's no way this made a Navier-Stokes parallel run diff.

PR #3829, opened 2024-04-15
Use if-statements with initializers to limit variable scope

jwpeterson:
This is a [C++17 feature](https://en.cppreference.com/w/cpp/language/if) which all of our "min" compilers correctly support.

### DIFF
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -1486,10 +1486,8 @@ merge_ghost_functor_outputs(GhostingFunctor::map_type & elements_to_ghost,
           // if we had to fix up gf output to get here.
           libmesh_assert(elem->active());
 
-          GhostingFunctor::map_type::iterator existing_it =
-            elements_to_ghost.find (elem);
-
-          if (existing_it == elements_to_ghost.end())
+          if (auto existing_it = elements_to_ghost.find(elem);
+              existing_it == elements_to_ghost.end())
             elements_to_ghost.emplace(elem, elem_cm);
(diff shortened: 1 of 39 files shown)


roystgnr (approved):
You stripped a few `const auto` down to `auto` when I'd have been happier the other way, but honestly I'm so impressed you did all the hard cases too and not just the "bring `auto it = foo.find` into the if" cases, I'm happy to go along with your preference.

Especially since my real semantic preference is `const map_type::const_iterator` in most cases but that syntax is so much more awkward than `auto`...

Since you did do so many of the hard cases by hand I'm tempted to throw every CI recipe we have at this to try and maximize coverage, but if you want to just merge now and fix any devel->master breakage after the fact, that's fine too.

jwpeterson:
Go this error from the Coverage recipe:
```
Mutiple base shas were found when retrieving head reports:
  6baf4e8503c86a8c7f3336b8e5e16860f5027769
  ab641a5a8c31db63fa81e3dd8ae481e201422080
Failed to generate the full coverage report
```

jwpeterson:
> You stripped a few const auto down to auto

Ah, yes I agree this was a bit overzealous. I'll put the consts back in (and possibly add some more?) before merging.

roystgnr:
I've seen that coverage error before, and it's always been a false positive (something about an unrelated merge to master?), so I would ignore it, but if you're pushing more changes then it'll be good to see the coverage stats go through properly afterward.

Issue #3489, opened 2023-03-02
Add API to check and/or assert that multiple bcids don't have the same sideset/nodeset name


roystgnr:
This may be leading to problems in apps using MOOSE, which treats names as the "first-class" identifier rather than numeric ids, in the case of complex compound-constructed meshes.

PR #3830, opened 2024-04-16
Assert bcid set names are different

andcarl1364:
This solves problems found when using MOOSE in tandem with libmesh due to MOOSE’s implementation treating names as the “first-class” identifier.

Resolves: #3489

### DIFF
--- a/include/reduced_basis/rb_construction.h
+++ b/include/reduced_basis/rb_construction.h
@@ -839,7 +839,7 @@ protected:
   /*
    * Pre-evaluate the theta functions on the entire (local) training parameter set.
    */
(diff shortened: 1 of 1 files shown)


roystgnr:
This definitely doesn't resolve #3489 - it's adding a check to a particular app, not in an API that can be directly called from MOOSE.

And I'm not actually seeing how it's useful in that app.  The names this PR allows the user to specify as command line arguments aren't even related to the names set in the mesh.  In opt mode the new options do nothing; in devel/dbg modes they just assert that you didn't give the same name twice on the command line; in neither case do they seem to have anything to do with the mesh id->name map.

PR #3832, opened 2024-04-18
Drop workaround that was only required on 'older' compilers

jwpeterson:
Our minimum compilers may now be new enough to handle this type of converting constructor: https://stackoverflow.com/questions/58181090/why-does-unique-ptrderived-implicitly-cast-to-unique-ptrbase


jwpeterson:
Looks like our minimum compilers are happy with the simpler syntax here, so I'll go ahead and merge once all the testing is done.

Issue #3714, opened 2023-11-16
Enhance MeshFunction to handle vector variables


lindsayad:
We would resize the output vector to be equal to `vec_dim * n_vector_vars + n_scalar_vars`

andcarl1364:
so we can easily get the vec_dim using this->_system_vars.size() as already implemented, however which specific vector vars are we counting to calculate n_vector_vars and the same for scaler vars with n_scalar_vars?

roystgnr:
`vec_dim` isn't `_system_vars.size()`; it would be the dimensionality of the vector variables, not the total number of variables.

`_system_vars` is a vector of variable indices; vector variables with any of those indices would be included in `n_vector_vars` and scalar variables with any of those indices would be included in `n_scalar_vars`.

andcarl1364:
thank you! I guess what i'm confused about is which specific variables am I checking for having those indices, are they in the mesh_function.C file? and is there some sort of map of vector variables and their dimensionality in order to calculate `vec_dim`?

roystgnr:
Variables are added dynamically to a `System`; `DofMap::n_variables()` returns the total count of them for its associated system.  Each variable's information is available from `DofMap::variable(i)`; that information includes a `FEType`, and `FEInterface::n_vec_dim(mesh, fe_type)` indicates how many dimensions a particular non-`SCALAR` variable has.

You might want to consider looking through some of the libMesh example programs; e.g. the distinction between variables in the C++ (might be named in a library .C file) context versus variables in the mathematical (added and named by an application) context is shown in introduction_ex2.C; later introduction_ex4.C is the first to use the returned index of a newly-added variable.

PR #3835, opened 2024-04-22
resizes output vector to better handle vectors

andcarl1364:
enhances MeshFunction to better handle vector variables which in turn helps general field transfers work with vectors as well

resolves: #3714
see also: #26084, #26045

### DIFF
--- a/include/mesh/mesh_function.h
+++ b/include/mesh/mesh_function.h
@@ -409,7 +409,7 @@ protected:
 
   /**
    * the new size to resize output to  
-  */
+   */
   int _new_resize;
 };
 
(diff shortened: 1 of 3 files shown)


roystgnr:
What does this do?

`_new_resize` gets initialized to 0, then I don't see it ever being used again.

`vec_dim`, `n_vector_vars`, and `n_scalar_vars` get calculated (somewhat confusingly, but that's an existing problem with the overlapping terminology for scalar-valued-functions vs discrete-per-system-scalars) ... and then I don't see them ever being used again?

andcarl1364:
my file didn't save correctly, let me do it again

PR #3842, opened 2024-04-24
Innocuous "fix" for FE<3,{HIERARCHIC,LAGRANGE}_VEC>::n_dofs_at_node

nmnobre:


jwpeterson (approved):
This seems like an obvious bug, I guess that means we don't have any unit testing of this function for 3D HIERARCHIC_VEC elements...

nmnobre:
> This seems like an obvious bug, I guess that means we don't have any unit testing of this function for 3D HIERARCHIC_VEC elements...

It’s a conceptual bug, but not a bug in practice because all specialisations of the method for the scalar classes call the same function anyway. Thus the quotation marks around “fix”.

PR #3843, opened 2024-04-26
Test (in debug mode) partitioner graph symmetry

roystgnr:
If we don't catch a problem here, we can end up getting hit by it in Parmetis with a much more confusing error message.

I'm using this to debug other work, but assuming it doesn't give us any false positives it's probably worth making permanent.


roystgnr:
Passing the basic tests is a good start.  I'll add a debug-mode distributed sweep to try to be extra cautious.

PR #3844, opened 2024-04-26
Keep mesh prepared when an extraction matrix has us create new NodeElems

roystgnr:
I can work around this at the application level, but we shouldn't have to.


roystgnr:
Closing this in favor of a much bigger branch with diagnostics and fixes for the bugs I uncovered.

PR #3845, opened 2024-04-28
Use std::make_unique instead of raw new when possible

jwpeterson:
Follow-up to #3834.


jwpeterson:
The only failure here is due to an unrelated environment issue in the Test MOOSE mac target, so I think it should be ready for your review, @roystgnr.

roystgnr (approved):

PR #3846, opened 2024-04-29
Eigenproblem example fixes

lindsayad:
- Add html doc
- Add correct inline documentation in example
- Don't do refinement if adaptivity is not available


jwpeterson (approved):
Thanks!

PR #3847, opened 2024-04-30
Assert the assumptions made at the top of DofMap::local_variable_indices

lindsayad:
These assertions need to hold true or else we are not actually adding all our indices


roystgnr:
My (slow) response from Slack:

> Aren't we looping over all local node pointers there?  So don't we expect to see a ton of indices that we've already seen when the same nodes were hit in the local element loop?

If we're lucky in a particular case (well, if we've renumbered our own nodes) then the node numbering will still have us passing those asserts - but that's *un*lucky in the general case, because then we may break as soon as we load a mesh with renumbering disabled and an inconsistent parallel partitioning.

lindsayad:
~Can we do the assertion in the element loop~ We are likely to touch the same node multiple times in the element loop

PR #3849, opened 2024-05-01
Return the unmodified normalization from get_eim_error_indicator()

dknez:
It's better to return the unmodified normalization from get_eim_error_indicator() since that can be used to indicate if the EIM approximation is identically zero or not.


jwpeterson (approved):

jwpeterson:
Test MOOSE mac failure is unrelated

PR #3851, opened 2024-05-02
Allow 1D elements to break symmetry of the "is neighbor of" relationship

jwpeterson:
Previously if we had an odd number of 1D elements which met at the same Node, one of them would end up not being paired off as neighbors with any other elements, and therefore, from its perspective, would appear to be on a boundary. This PR changes that behavior by allowing multiple 1D elements to have the same neighbor in that situation. I had to weaken some of the checks in `Elem::libmesh_assert_valid_neighbors()` to make this work, and I'm curious what else will break due to this change.

### DIFF
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -1219,26 +1219,37 @@ void Elem::libmesh_assert_valid_neighbors() const
           else
             {
               unsigned int rev = neigh->which_neighbor_am_i(elem);
-              libmesh_assert_less (rev, neigh->n_neighbors());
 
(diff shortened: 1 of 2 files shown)


jwpeterson:
Closing this since it was mainly just a test to see how much stuff would break if we allowed non-mutual neighbors, and the answer is "a lot". Basically anything that calls `Elem::which_neighbor_am_i()` is broken by this change when there are > 2 1D Elems meeting at a single Node. For reference, we have an existing unit test, `testPostInitAddElem`, that sets up such a Mesh and it gets broken by this change.

PR #3853, opened 2024-05-03
Relax some find_neighbors() assertions that can be overzealous for 1D Elems

jwpeterson:
I don't think it will be easy to make #3851 work correctly in the short term, so this PR takes a different approach which is basically to just allow incomplete/incorrect neighbor information on meshes with multiple 1D Elems meeting in a single Node. To be clear, we were already doing this so the PR doesn't make it any worse, but we did have some overzealous asserts in `find_neighbors()` which are currently being tripped when refining such meshes.

### DIFF
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -1047,11 +1047,26 @@ void UnstructuredMesh::find_neighbors (const bool reset_remote_elements,
                       // Let's make sure that "had children made remote"
                       // situation is actually the case
                       libmesh_assert(neigh->has_children());
+
+                      // This assert is overzealous in the case where
+                      // N 1D elements (N>2, even) meet in a common
+                      // node, but an odd number are flagged for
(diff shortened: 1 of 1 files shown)


jwpeterson:
> - Error opening ExodusII mesh file: meshes/find_neighbors_junction_1D_only.e

Ah, I think the failures I'm seeing from the unit tests are due to using a netCDF-4/HDF5 exo file... I think there is a way to convert it to something more portable to our CI, I just need to remember what that is...
```
ncdump -k tests/meshes/find_neighbors_junction_1D_only.e
netCDF-4
```

roystgnr:
IIRC you can use h5dump to get ASCII out of the ExodusII-HDF5 file and pipe that to ncgen to put it into an ExodusII-NetCDF3 file.

jwpeterson:
> IIRC you can use h5dump

I used `ncdump`/`ncgen`, but otherwise yes this approach worked for me.

PR #3859, opened 2024-05-24
More testing and support for IsoGeometric Analysis on distributed meshes

roystgnr:
We've been running IGA on distributed meshes for a while, but it turned out that there were a number of corner cases that could trigger bugs, including simple *re*-distribution.  This adds test coverage (when run in a --enable-distmesh sweep, as we do on master merges and as I'll do here) as well as fixes for those cases, plus a lot of additional diagnostic tools and internal assertions that were helpful along the way.

This also includes the improvements from #3844, which were the first things to trigger a couple of these bugs and send me down this rabbit hole.

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -1716,6 +1716,22 @@ public:
   void copy_constraint_rows(const SparseMatrix<T> & constraint_operator);
 
   /**
+   * Prints (from processor 0) all mesh constraint rows.  If \p
+   * print_nonlocal is true, then each constraint is printed once for
+   * each processor that knows about it, which may be useful for \p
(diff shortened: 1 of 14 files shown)


roystgnr:
In theory the cppunit tests should catch every bug I hit even in an opt mode sweep, but to be safe I'm adding the dbg mode sweeps too.

roystgnr:
I was ready to talk smack about a MOOSE test that tries to do an exodiff on element ghosting patterns, ready to sneeze at the next Parmetis point upgrade ... but not only does it use a linear partitioner for reproducibility, it found a real bug here -the new code can overghost after adaptive mesh refinement.

jwpeterson (approved):

PR #3861, opened 2024-05-28
Install additional FParser headers

### LINKED ISSUE #3860: Install FParser `fptypes.hh` header
Derived function parser objects may implement their own evaluation logic. For that to work the internal `Data` structure must be known. The header `fptypes.hh` where it is defined is currently not installed.

dschwen:
It's possible to derive from the function parser (or its AD version) to [implement your own evaluation](https://github.com/libMesh/libmesh/blob/ecc44b12009d7bbc963f1c515c72513ed2c9a5b4/contrib/fparser/fparser.hh#L142-L145) based on the function byte code. I need to do that to implement evaluation using libtorch tensors as the datatype. For that to work I need the header defining the internal data structures of FParser. 

Closes #3860

### DIFF
--- a/contrib/fparser/extrasrc/Makefile.am
+++ b/contrib/fparser/extrasrc/Makefile.am
@@ -15,3 +15,10 @@ fp_opcode_add.inc: $(srcdir)/fp_opcode_add.inc.release
 	$(AM_V_GEN)$(LN_S) -f $< $@
 
 endif
+
+# We need to install some additional headers to permit
+# developers to derive from FParser to implement evaluation
+# on custom types.
+includedir = $(prefix)/include/libmesh/extrasrc
+
+include_HEADERS = fptypes.hh
--- a/contrib/fparser/Makefile.am
+++ b/contrib/fparser/Makefile.am
@@ -38,7 +38,7 @@ noinst_PROGRAMS = # none, append below
 # parsed_function.h
 includedir = $(prefix)/include/libmesh
 
-include_HEADERS = fparser.hh fparser_ad.hh
+include_HEADERS = fparser.hh fparser_ad.hh fpconfig.hh
 
 
 EXTRA_DIST += util/bytecoderules.dat util/bytecoderules_header.txt
--- a/contrib/fparser/Makefile.in
+++ b/contrib/fparser/Makefile.in
@@ -1106,7 +1106,7 @@ CLEANFILES = fpoptimizer/grammar_data.cc $(am__append_7)
 # in case they weren't conditionally cleaned:
 # DISTCLEANFILES += util/cpp_compress
 DISTCLEANFILES = util/bytecoderules_parser util/tree_grammar_parser
-include_HEADERS = fparser.hh fparser_ad.hh
+include_HEADERS = fparser.hh fparser_ad.hh fpconfig.hh
 FPOPTIMIZER_CC_FILES = \
 	    lib/crc32.hh \
 	    lib/autoptr.hh \
--- a/contrib/fparser/extrasrc/Makefile.in
+++ b/contrib/fparser/extrasrc/Makefile.in
@@ -18,6 +18,7 @@
 # nothing to do here, and fp_opcode_add.inc will get
 # generated. however, if we are building the release
 # version, we need to link it.
+
 VPATH = @srcdir@
 am__is_gnu_make = { \
   if test -z '$(MAKELEVEL)'; then \
@@ -155,7 +156,8 @@ am__aclocal_m4_deps =  \
 	$(top_srcdir)/configure.ac
 am__configure_deps = $(am__aclocal_m4_deps) $(CONFIGURE_DEPENDENCIES) \
 	$(ACLOCAL_M4)
-DIST_COMMON = $(srcdir)/Makefile.am $(am__DIST_COMMON)
+DIST_COMMON = $(srcdir)/Makefile.am $(include_HEADERS) \
+	$(am__DIST_COMMON)
 mkinstalldirs = $(install_sh) -d
 CONFIG_HEADER = $(top_builddir)/include/libmesh_config.h.tmp
 CONFIG_CLEAN_FILES =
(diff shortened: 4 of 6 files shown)


dschwen:
I should also get rid of this [friend declaration](https://github.com/libMesh/libmesh/blob/ecc44b12009d7bbc963f1c515c72513ed2c9a5b4/contrib/fparser/fparser.hh#L152) and use `getParserData()` instead.

roystgnr:
Every newly-exposed interface is to some small extent a new failure of modularity ...

But that said, this seems perfectly reasonable to me; the right thing to do, even.  That's a public header, just not one we expected to need.

roystgnr (approved):

contrib/fparser/extrasrc/Makefile.am
```diff
@@ -15,3 +15,10 @@ fp_opcode_add.inc: $(srcdir)/fp_opcode_add.inc.release
 	$(AM_V_GEN)$(LN_S) -f $< $@
 
 endif
+
+# We need to install some additional headers to permit
+# developers to derive from FParser to implement evaluation
+# on custom types.
+includedir = $(prefix)/include/libmesh/extrasrc
```
jwpeterson, line 22: fparser doesn't really follow the pattern of our other contrib libraries that have multiple headers, e.g. timpi, poly2tri, metaphysicl, which get installed in `include/libmesh/timpi`, etc. I don't really like the idea of having an "extrasrc" directory at the same level as these other libraries, with the other fparser headers being at include/libmesh/fparser.hh and include/libmesh/fparser_ad.hh... Given that we are shaking things up anyway, I think this would be a good time to make an `include/libmesh/fparser` top level include directory. I think we could also make it backwards compatible by setting the include search paths correctly in the Makefile.

roystgnr: There's no `include/libmesh/timpi` - we install to `include/timpi`.  So `include/fparser` would be the way to go there ... except that, for `include/timpi` and the like, they all have one-way dependencies - they can be installed separately, they can be compiled against and linked to even if the rest of libMesh isn't there, they don't require anything from the libMesh directory, etc.  But as long as our fparser fork is still demanding `libmesh_config.h` rather than being operable as a standalone fork, I think `include/libmesh/fparser` might be a reasonable way to organize it.

dschwen:
My main concern is that we'd be making more modifications on the files we got from upstream, which would make it more difficult to update FParser (which admittedly hasn't happened in a long time now).

jwpeterson:
> My main concern is that we'd be making more modifications on the files we got from upstream,

At the moment, fptypes.hh does a relative inclusion of the fpconfig.hh header.
```
#include "../fpconfig.hh"
```
I don't think we need to change that, the headers can still be in the same relative positions but be installed to `include/libmesh/fparser` instead of directly in `include/libmesh` the way they are now.

contrib/fparser/extrasrc/Makefile.am
dschwen: There's `LIBMESH_USING_THREADS`, `LIBMESH_HAVE_FPARSER_JIT`, and `LIBMESH_HAVE_CXX11_ERF` which we use in FParser.

dschwen: > I think we could also make it backwards compatible by setting the include search paths correctly in the Makefile.

Ideally we would want `libmesh-config` spit out the additional `fparser` include path. I have no idea where to make that change.

dschwen: Actually.. sigh, not that easy. I see where the path is set in `contrib/bin/libmesh-config.in`, but we do not add `libmesh` to that path (unless in legacy include mode). So the include statements are

currently:
```
#inlcude "libmesh/fparser.hh"
```

proposed:
```
#inlcude "libmesh/fparser/fparser.hh"
```
There is not additional path we can add that would make that change backwards compatible.

jwpeterson: I see... the issue is that fparser doesn't have it's own `./configure`, so it's relying on a bunch of flags determined by libmesh's configure... other than that, the coupling is basically a "one-way" coupling like the rest of the contrib libraries.

jwpeterson: > There is not additional path we can add that would make that change backwards compatible.

OK, I guess the backwards compatibility layer will have to be us making a symbolic link from the existing
`include/libmesh/fparser.hh` -> `include/libmesh/fparser/fparser.hh`
?

Then we'd give people time to update their codes from:
```
#include "libmesh/fparser.hh"
```
to 
```
#include "libmesh/fparser/fparser.hh"
```
and eventually remove the links after some deprecation period.

dschwen: Uhm, and how do we make that symlink?

jwpeterson: That's a good question. We have this script, `include/libmesh/rebuild_makefile.sh` that you normally have to run whenever you add/remove files. It generates the `include/libmesh/Makefile.am` file. If you look at that Makefile.am, you should see that we already do some special handling of the fparser headers... I think maybe here is where we could set up additional links, but it would definitely take me some trial and error to figure this out. 
```
#
# contrib rules
if LIBMESH_ENABLE_FPARSER

fparser.hh: $(top_srcdir)/contrib/fparser/fparser.hh
	$(AM_V_GEN)rm -f $@ && $(LN_S) -f $< $@

fparser_ad.hh: $(top_srcdir)/contrib/fparser/fparser_ad.hh
	$(AM_V_GEN)rm -f $@ && $(LN_S) -f $< $@

  BUILT_SOURCES  += fparser.hh fparser_ad.hh
  DISTCLEANFILES += fparser.hh fparser_ad.hh

endif
```

dschwen: We need to set up the links when we install though not before. What should have been a small PR is now turning into an autotools nightmare for me :-(

jwpeterson: OK, well since @roystgnr is OK with the change as-is, I guess it's fine with me. I am wondering though if you have actually tested including this new header in some way at the MOOSE level? We don't have any unit tests or examples that use the new fparser header, so just passing the CI testing here doesn't tell us much.

dschwen: > I am wondering though if you have actually tested including this new header in some way at the MOOSE level?

I have a MOOSE app for an LDRD in which I've been using this libmesh branch for almost a week to implement a GPU accelerated function parser.

dschwen:
What also occurred to me is that I could try adding the missing headers to my app (a nasty workaround). But that would allow me to proceed without this PR and would give us some time to figure out the fparser header reorganization.

roystgnr:
An extremely nasty workaround in a typical use case, where we like to allow people to use newer versions than ours of 3rd party dependencies, but since we've already massively forked parser, there is no newer version, there's just ours, so that's not crazy.  I'd still prefer merging this PR over copying headers downstream, even if we change our minds about directories later.

CBqk3762:
Hi, I've been having problems compiling libmesh due to fparser:
```
making all in libmesh
make[3]: *** No rule to make target '../../../contrib/fparser/fparser.hh', needed by 'fparser.hh'.  Stop.
```
I can see this is a very recent thread so I am guessing the fix is not in the main branch yet? Could I fix my repo with specific files that were merged into devel here? I don't completely follow your discussion but it sounds like it might be possible to just replace certain files with the updated ones and maybe move some around so they're in the right directories?

Even if it's a bit of a clunky fix that's fine, I just really need libmesh working for a proof-of-concept!

jwpeterson:
This PR proposes an extension to the existing fparser functionality, and thus won't be related the bug you are reporting. It might be good if you open a separate Issue for the problem you are encountering, and let us know exactly what version of libmesh you are using, what steps you have done so far, etc.

CBqk3762:
Okay, my apologies, I will go and do that now

PR #3865, opened 2024-06-04
Strict petsc error code compatibility

lindsayad:
Some PETSc devs only configure with `--with-strict-petscerrorcode` and if I'd ever like them to look at something, it's a lot easier if our stack builds out of the box for them

### DIFF
--- a/include/numerics/petsc_vector.h
+++ b/include/numerics/petsc_vector.h
@@ -571,7 +571,7 @@ PetscVector<T>::PetscVector (Vec v,
 
   /* We need to ask PETSc about the (local to global) ghost value
      mapping and create the inverse mapping out of it.  */
-  PetscErrorCode ierr=0;
+  PetscErrorCode ierr = (PetscErrorCode)0;
   PetscInt petsc_local_size=0;
   ierr = VecGetLocalSize(_vec, &petsc_local_size);
   LIBMESH_CHKERR(ierr);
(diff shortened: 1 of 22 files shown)


roystgnr (approved):
I thought maybe the m4 changes were more `CXXFLAGS=-Werror configure` stuff I missed, but that wouldn't explain having to explicitly cast from int to PetscErrorCode.  What PETSc/compiler version was complaining about the implicit casts there?

lindsayad:
You get these errors/warnings when you configure PETSc `--with-strict-petscerrorcode`

include/numerics/petsc_vector.h
```diff
@@ -1187,7 +1187,7 @@ Real PetscVector<T>::min () const
 
   this->_restore_array();
 
-  PetscErrorCode ierr=0;
+  PetscErrorCode ierr = (PetscErrorCode)0;
```
jwpeterson, line 1190: @lindsayad does PETSc `--with-strict-petscerrorcode` warn/error about about C++ style casts, e.g.
```
auto ierr = static_cast<PetscErrorCode>(0);
```
It is a small thing but I prefer to use C++ style casts whenever possible.

lindsayad: It works fine. I've opened https://github.com/libMesh/libmesh/pull/3867 to change it

jwpeterson: OK, thanks, I was going to just update it at some later point, didn't mean to imply that you should do it :)

PR #3866, opened 2024-06-04
fix read VTK_IO bug

ermore:
When I read in a mesh file using VTK, I encounter that the volume of the triangular prisms is negative.

I compared the node connection order distribution of the VTK `VTK_WEDGE` and the libmesh `PRISM6`,their connection order is the same.
image
image

### DIFF
--- a/src/geom/cell_prism6.C
+++ b/src/geom/cell_prism6.C
@@ -344,11 +344,11 @@ void Prism6::connectivity(const unsigned int libmesh_dbg_var(sc),
       {
         conn.resize(6);
         conn[0] = this->node_id(0);
-        conn[1] = this->node_id(2);
-        conn[2] = this->node_id(1);
+        conn[1] = this->node_id(1);
+        conn[2] = this->node_id(2);
         conn[3] = this->node_id(3);
-        conn[4] = this->node_id(5);
-        conn[5] = this->node_id(4);
+        conn[4] = this->node_id(4);
+        conn[5] = this->node_id(5);
         return;
       }
 
--- a/src/geom/cell_pyramid5.C
+++ b/src/geom/cell_pyramid5.C
@@ -257,10 +257,10 @@ void Pyramid5::connectivity(const unsigned int libmesh_dbg_var(sc),
     case VTK:
       {
         conn.resize(5);
-        conn[0] = this->node_id(3);
-        conn[1] = this->node_id(2);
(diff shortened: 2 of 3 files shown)


roystgnr:
Those images don't show the same order.  In the VTK doc image, the 012 triangle is numbered counter-clockwise when viewed from outside; in our PRISM6 it's numbered clockwise (though in our image it looks counter-clockwise because the view direction is from inside).

Also, what's with the `vtkXMLPUnstructuredGridReader` -> `vtkXMLUnstructuredGridReader` change?

jwpeterson:
> When I read in a mesh file using VTK, I encounter that the volume of the triangular prisms is negative.

Can you please also share the code that you used to make this determination?

ermore:
> Those images don't show the same order. In the VTK doc image, the 012 triangle is numbered counter-clockwise when viewed from outside; in our PRISM6 it's numbered clockwise (though in our image it looks counter-clockwise because the view direction is from inside).
> 
> Also, what's with the `vtkXMLPUnstructuredGridReader` -> `vtkXMLUnstructuredGridReader` change?

----
When I use vtkXMLPUnstructuredGridReader to read mesh(mesh.vtu), I get an error:
```
2024-06-06 10:13:26.676 (   0.784s) [        9C6EAC80]       vtkXMLReader.cxx:997    ERR| vtkXMLPUnstructuredGridReader (0x1abff40): Cannot find PUnstructuredGrid element in file.
2024-06-06 10:13:26.676 (   0.784s) [        9C6EAC80]       vtkExecutive.cxx:741    ERR| vtkCompositeDataPipeline (0x1ac0b70): Algorithm vtkXMLPUnstructuredGridReader (0x1abff40) returned failure for request: vtkInformation (0x1ac8210)
  Debug: Off
  Modified Time: 112
  Reference Count: 1
  Registered Events: (none)
  Request: REQUEST_INFORMATION
  ALGORITHM_AFTER_FORWARD: 1
  FORWARD_DIRECTION: 0
```
However, when I read it into pvtu (multi-process file), there is no error.

roystgnr:
Yeah, I'm afraid we regressed .vtu support long ago when we accepted a patch to add .pvtu support and didn't realize the object used wasn't backwards compatible.  I'm not happy with that, but a regression in the other direction (I think?  Let me try activating CI here to see.) wouldn't be an improvement.

PR #3864, opened 2024-06-03
Fixes for configure with CXXFLAGS=-Werror

roystgnr:
@lindsayad never mentioned any names, but I strongly approve of whatever crazy freak has `$CXXFLAGS` set to `-Werror` in their environment, and I apologize for the false negatives it was giving for some of our configure tests, and I hope this fixes the lot of them.

PR #3867, opened 2024-06-04
Use static_cast instead of C-cast for PetscErrorCode

lindsayad:
Follow-on to #3864


jwpeterson (approved):

roystgnr:
Sorry for not getting to this yesterday, but ... couldn't we just use `PETSC_SUCCESS` instead of casting 0?

roystgnr:
Whoa, never mind!  `PETSC_SUCCESS` was only introduced in December 2022!

PR #3868, opened 2024-06-06
Don't ruin Petsc error stack

lindsayad:
I believe `PetscFunctionBegin` pushes and `PetscFunctionReturn` pops. So I messed that up with my recent changes

### DIFF
--- a/src/solvers/petsc_diff_solver.C
+++ b/src/solvers/petsc_diff_solver.C
@@ -46,6 +46,8 @@ extern "C"
                                        PetscReal fnorm,
                                        void * ctx)
   {
+    PetscFunctionBegin;
+
     PetscDiffSolver & solver =
       *(static_cast<PetscDiffSolver *> (ctx));
 
@@ -87,6 +89,8 @@ extern "C"
   PetscErrorCode
   __libmesh_petsc_diff_solver_residual (SNES, Vec x, Vec r, void * ctx)
   {
+    PetscFunctionBegin;
+
     libmesh_assert(x);
     libmesh_assert(r);
     libmesh_assert(ctx);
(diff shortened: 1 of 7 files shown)


jwpeterson:
I would assume you are meant to pair every  `PetscFunctionBegin` with a `PetscFunctionReturn`, but that's not what this diff is showing, i.e. you're just adding `PetscFunctionReturn`s here?

lindsayad:
we're on the same page

include/numerics/petsc_solver_exception.h
```diff
@@ -39,12 +40,6 @@
 namespace libMesh
 {
 
-// Forward declarations
-namespace Parallel
-{
-class Communicator;
-}
-
```
jwpeterson, line 47: I don't see any other changes in this header that would require a full inclusion over a forward declaration. (I understand this PR might still be under development, so maybe there's another change I'm not seeing yet.)

lindsayad: This is failing on devel due to my changes in the previous PRs which is why you probably don't see the changes that created the need for this change

lindsayad: Probably should have separated this PR into two (still can if you would prefer)

jwpeterson: OK, it's up to you. I guess one reason to do it would be to make it easier to revert just the PetscFunctionBegin/Return stuff at some later time.

lindsayad: Kk opened https://github.com/libMesh/libmesh/pull/3869

lindsayad:
should be ready for review

jwpeterson (approved):

PR #3869, opened 2024-06-06
Fixes for configuring libmesh without exceptions

lindsayad:
Refs failures on devel

### DIFF
--- a/src/numerics/wrapped_petsc.C
+++ b/src/numerics/wrapped_petsc.C
@@ -26,7 +26,7 @@
 
 // PETSc includes
 #ifdef I
-# define LIBMESH_SAW_I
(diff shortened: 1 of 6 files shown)


src/numerics/wrapped_petsc.C
```diff
@@ -26,7 +26,7 @@
 
 // PETSc includes
 #ifdef I
-# define LIBMESH_SAW_I
```
roystgnr, line 29: I think the indentation here was intentional; we typically do that for preprocessor directives inside preprocessor if/else blocks.

Not that any extra nesting clarity is needed for a one-line block, though.

roystgnr (approved):

PR #3871, opened 2024-06-10
Adding an example to the testing suite to see if coverage changes Ref…

kyriv1980:
…s #000

Just a test, to see if coverage changes,

### DIFF
--- a/examples/miscellaneous/miscellaneous_ex8/miscellaneous_ex8.C
+++ b/examples/miscellaneous/miscellaneous_ex8/miscellaneous_ex8.C
@@ -15,8 +15,6 @@
 // License along with this library; if not, write to the Free Software
 // Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
-
-
 // <h1>Miscellaneous Example 8 - Meshfree Interpolation Utilities.</h1>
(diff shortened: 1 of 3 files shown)


kyriv1980:
@roystgnr

roystgnr:
Yeah; got it.

Adding some context from the Slack discussion:

Vasily here was looking through our test coverage reports, and noticed this afternoon that we have a class with 0% test coverage in LCOV despite it being invoked in an example code.  This PR is a bit of a hack, moving some of that example into a unit test to see if *that* gets our code coverage data to see it.

Though ... looking at it now ... I think I see a different root of the problem.

roystgnr:
Yeah, I see what happened now.  At some point somebody copy-and-pasted code that they should have reused.

(fun fact: the above sentence has probably been true a hundred million different times by now)

We have a `MeshlessInterpolationFunction` defined in meshfree_solution_transfer.C, where LCOV sees it not being used in our test runs, and we also have *another* one defined in misc ex8, where it's getting used but (because it's not library code) LCOV doesn't care that it's being used.  I'll let these recipes finish since it's still possible that our coverage report will see something I didn't, but I think the actual fix here is to factor the library function out into its own header+source, then use *it* in misc ex8 to get the extra coverage.

PR #3875, opened 2024-06-11
Re-bootstrap

roystgnr:
Somehow not all the m4 changes from #3865 made it into `configure`?

### DIFF
--- a/configure
+++ b/configure
@@ -51979,8 +51979,10 @@ ac_compiler_gnu=$ac_cv_cxx_compiler_gnu
 
                             int main(int argc, char **argv)
                             {
-                              SlepcInitialize(&argc, &argv, (char*)0, help);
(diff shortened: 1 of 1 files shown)


roystgnr:
Going to ignore those Mac failures and merge if the rest of CI passes.  I can kick runs that fail to fetch due to the sudden network failures this afternoon, but I've never seen `Mac sandbox test error` before...

jwpeterson (approved):

roystgnr:
Speaking of CI issues ... why the heck does our `Test 32 bit` recipe take 5 times as long as it should??  This one is *not* a new failure ... looks like it started nearly a year ago...

roystgnr:
Ah, there it is.  We used to only run that case in `opt` mode, now we use `dbg` too.  False alarm.

PR #3876, opened 2024-06-11
Use newer TIMPI with `set_union(*multi*)` support

roystgnr:
We don't need to use this in libMesh yet but it makes more flexibility available downstream.


roystgnr:
`controls/web_server_control.control/vec_string` failure looks like a CI environment issue, not PR-related.

PR #3877, opened 2024-06-11
Kyriv/increase test coverage

kyriv1980:
This is a Pr that creates the new MeshlessInterpolationFunction class and source and build system entries, switches the MeshfreeSolutionTransfer to use the new official version and removes the internal version from that source file, switches misc ex8 to use the new official class and removes the internal version from its source file and build system, and updates the re-bootstrap results.


kyriv1980:
@roystgnr Let me know if everything is in order. Thank you.

roystgnr:
It's a start.  Issues:

You still need to remove that first test commit ... and that's going to force you to manually redo the later miscellaneous_ex8.C changes, because those were made *on top* of the accidental-reformatting-to-MOOSE-standards changes.

Did you have whatever "reformat changes to MOOSE standards" editor setting still turned on when you were working on `meshfree_solution_transfer.C`?  I kinda like the alphabetization of the `#include` directives, especially since it didn't mess with the outer ordering based on category, but the reformatting of `MeshfreeSolutionTransfer::transfer` was wrong.

You mixed real changes with bootstrap changes in https://github.com/libMesh/libmesh/pull/3877/commits/b55f5fd27c8fdbeaf4deeb8e91da94cda39bbd29

roystgnr:
Ah, and now I see the new revert commit.  That fixes the end state of leaving that test commit in, but we don't really need the test commit in the history, even; you ought to just `rebase -i` and get rid of it entirely.

kyriv1980:
I think it would be easier to scrap this all together and redo from scratch?

kyriv1980:
Scratch PR and redo.

PR #3878, opened 2024-06-11
Kyriv/refactor solution transfers

kyriv1980:
Pr that creates the new MeshlessInterpolationFunction class and refactors as needed.


kyriv1980:
@roystgnr I re-did the Pr from the beginning cause i didn't want to get into the weeds with git history etc...

roystgnr:
This looks good.  Only one change I can see to make, but let's wait until we get the coverage report for this version of the branch first.

src/solution_transfer/meshfree_solution_transfer.C
```diff
@@ -37,68 +37,6 @@ namespace libMesh
...
-   */
-  MeshlessInterpolationFunction (MeshlessInterpolationFunction &&) = default;
-  MeshlessInterpolationFunction (const MeshlessInterpolationFunction &) = default;
-  virtual ~MeshlessInterpolationFunction () = default;
-
-  /**
-   * This class contains const references so it can't be assigned.
-   */
-  MeshlessInterpolationFunction & operator= (const MeshlessInterpolationFunction &) = delete;
-  MeshlessInterpolationFunction & operator= (MeshlessInterpolationFunction &&) = delete;
```
roystgnr, line 61: We should probably add the above 5 move/copy/assignment methods to the new consolidated class.  If @jwpeterson was enough of a stickler to put them in this internal-library-use-only version, I can't imagine him letting us get away with leaving them out of the exposed version.

roystgnr (changes requested):

include/solution_transfer/meshless_interpolation_function.h
```diff
@@ -29,126 +27,55 @@
...
+template <typename T> class DenseVector;
 
 // ------------------------------------------------------------
 // MeshlessInterpolationFunction class definition
-class MeshlessInterpolationFunction : public FunctionBase<Number>
-{
+class MeshlessInterpolationFunction : public FunctionBase<Number> {
 private:
-  const MeshfreeInterpolation & _mfi;
+  const MeshfreeInterpolation &_mfi;
```
jwpeterson, line 39: Don't delete this space.

include/solution_transfer/meshless_interpolation_function.h
```diff
@@ -29,126 +27,55 @@
...
   mutable std::vector<Point> _pts;
   mutable std::vector<Number> _vals;
-  Threads::spin_mutex & _mutex;
+  Threads::spin_mutex &_mutex;
```
jwpeterson, line 42: Or this one.

jwpeterson (changes requested):
Please keep all the spacing the way it was originally. There's no reason to have any whitespace only diffs in this file, that way we can focus on things which actually changed.

include/solution_transfer/meshless_interpolation_function.h
kyriv1980: at some point i deactivated the clang formatting that edits the code when i save. But maybe some edits got through before that. Is there something equivalent for the libmesh world?

include/solution_transfer/meshfree_solution_transfer.h
```diff
@@ -21,6 +21,7 @@
 #define MESHFREESOLUTIONTRANSFER_H
```
jwpeterson, line 21: Although It's not part of your change, I just noticed that this include guard doesn't follow our usual pattern. So, these two lines could be changed to:
```
#ifndef MESHFREE_SOLUTION_TRANSFER_H
#define MESHFREE_SOLUTION_TRANSFER_H
```
if you felt so inclined.

include/solution_transfer/meshless_interpolation_function.h
jwpeterson: I think @lindsayad was going to throw a .clang-format file into the libmesh repo at some point, but it doesn't look like he did so yet. Other than that, we just have some [formatting conventions](https://github.com/libMesh/libmesh/wiki/Source-code-formatting-conventions) that we use, but aren't rigidly enforced by anything.

lindsayad: Yea I haven't thrown anything in yet, to my chagrin

roystgnr: I tried to throw in one once, but it kind of backfired on me as soon as I started saving files with my vim "auto-apply .clang-format" settings.  clang-format doesn't have (and IIRC has a FAQ about not having) an option to tell it to enforce rules A1, B3, and C2 while simply preserving the existing formatting for rules D through Z; it's an all-or-nothing reformatter.

If we want to decide on formatting for every option they have, I guess we could do all the changes in one giant auto-reformat-everything commit and then proceed from there, but that's the only way to avoid editors turning every `:wq` into a giant mess of whitespace changes.

roystgnr:
John asked me about something that I agree we might want to nip in the bud - since this is the first time we're exposing this class to users, let's fix up the name consistency while we can still do so without breaking anybody's code?  If we replace `s/MeshlessInterpolationFunction/MeshfreeInterpolationFunction/g`, and similarly on all the `meshless_` filenames and `MESHLESS_` macros, that'll match the terminology choice in the existing `MeshfreeInterpolation` and `MeshfreeSolutionTransfer` classes.

jwpeterson (approved):
This looks much better. I have one or two very minor formatting nitpicks still but I can just fix those before merge.

include/solution_transfer/meshless_interpolation_function.h
lindsayad: I tell emacs to exclude certain directories like emacs for my clang formatting hook. I would assume you could tell vim the same thing, but if not, then I guess emacs must be better 😺

roystgnr (approved):

roystgnr:
Looks good to me; I'll leave it to John to hit the button after he pushes whatever formatting tweaks to the branch.  Thanks!

include/solution_transfer/meshless_interpolation_function.h
roystgnr: :fishing_pole_and_fish:

I can tell vim to ignore libMesh directories, but then what's the point of having a .clang-format file there?  Just to look pretty?

lindsayad: What I do when developing new code in libMesh is to write it, then select the new code region, and then `M-x clang-format`

PR #3880, opened 2024-06-13
A few more LOG_SCOPE entries

roystgnr:
I noticed a too-slow setup with a 33MB constraint matrix file; this helped me track it down more finely.

### DIFF
--- a/src/numerics/petsc_matrix.C
+++ b/src/numerics/petsc_matrix.C
@@ -24,6 +24,7 @@
 // Local includes
 #include "libmesh/petsc_matrix.h"
 
+// libMesh includes
 #include "libmesh/dof_map.h"
 #include "libmesh/dense_matrix.h"
 #include "libmesh/libmesh_logging.h"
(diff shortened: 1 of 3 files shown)


roystgnr:
`Stale file handle` in those failing tests is a false positive, environment problem.

src/numerics/petsc_matrix.C
```diff
@@ -23,8 +23,10 @@
 
 // Local includes
 #include "libmesh/petsc_matrix.h"
+
```
jwpeterson, line 26: Are you wanting to add a blank line after the header which is named after the current source file? I don't think we've done this in the past, and I prefer not to change it since then the comment `// Local includes` does not seem to apply to all the other headers.

jwpeterson (approved):
I'm good to merge unless you are hoping to try and fix the stale file handle thing...

src/numerics/petsc_matrix.C
roystgnr: I like making it clear that "this one header isn't in the same category as all the others for a reason", vs "we can't *quite* get alphabetical order right.

I should have added another comment, though; what we've done in some other .C files (e.g. dof_map.C),  is `// Local includes` for the source-file-specific header and `// libMesh includes` for the rest.  I'll fix that now and see if that kick resolves the file handle issues too.

PR #3887, opened 2024-06-25
Fixes for `make dist` and `make distcheck`

roystgnr:
I'm almost certainly going to need to do this with the Netgen branch too, but first let's at least get *master* working.

The trickiest bit here is the optional NetCDF subdirectories - if we want a tarball that lets users choose v4.6.2 or v4.9.2 at their `configure` time, then we need the ability to choose `all` at *our* configure time before the `make dist`.


roystgnr:
Everything but the (dbg) `Test 32 bit` has passed, but I'm going to need to tack on another commit or two; apparently there were *4* major changes that aren't passing distcheck.

roystgnr:
Okay, this is finally passing a full `make distcheck` for me.

jwpeterson (approved):
Thanks for looking into this, I had no idea `make dist` had gotten so far out of date. As you mentioned, it would definitely be good to have CI testing for this, possibly on the devel -> master merge.

roystgnr:
Thanks!  I'll force-push my rebased netgen branch (which just passed distcheck locally too!) now.

PR #3895, opened 2024-07-01
Add fuzzy equals comparisons

lindsayad:
Split off from #3883 

- [x] Add unit tests

### DIFF
--- a/include/numerics/numeric_vector.h
+++ b/include/numerics/numeric_vector.h
@@ -786,6 +786,11 @@ public:
    */
   bool compatible(const NumericVector<T> & v) const;
 
+  /**
+   * checks whether the vector \p v is fuzzy equal to this vector
+   */
+  bool fuzzy_equal(const NumericVector<T> & v) const;
+
 protected:
 
   /**
--- a/include/numerics/numeric_vector.h
+++ b/include/numerics/numeric_vector.h
@@ -315,6 +315,12 @@ public:
   Real l2_norm_diff (const NumericVector<T> & other_vec) const;
 
   /**
+   * \returns The \f$ \ell_1 \f$-norm of \f$ \vec{u} - \vec{v} \f$, where
+   * \f$ \vec{u} \f$ is \p this.
+  */
+  Real l1_norm_diff (const NumericVector<T> & other_vec) const;
+
+  /**
    * \returns The size of the vector.
    */
   virtual numeric_index_type size () const = 0;
@@ -786,15 +792,6 @@ public:
    */
   bool compatible(const NumericVector<T> & v) const;
 
-  /**
-   * checks whether the vector \p v is fuzzy equal to this vector by doing element-wise comparisons.
-   * A given element will be deemed fuzzy equal if either a relative or absolute tolerance fuzzy
-   * equal comparison returns true
-   */
-  bool fuzzy_equals(const NumericVector<T> & v,
-                    const Real rel_tol = TOLERANCE,
-                    const Real abs_tol = TOLERANCE) const;
-
 protected:
 
   /**
@@ -1077,8 +1074,24 @@ void  NumericVector<T>::swap (NumericVector<T> & v)
 }
 
 
-} // namespace libMesh
 
+template <typename T>
+auto
+l1_norm(const NumericVector<T> & vec) -> decltype(vec.l1_norm())
+{
+  return vec.l1_norm();
+}
+
+
+
+template <typename T>
+auto
+l1_norm_diff(const NumericVector<T> & vec1,
+             const NumericVector<T> & vec2) -> decltype(vec1.l1_norm_diff(vec2))
+{
+  return vec1.l1_norm_diff(vec2);
+}
+} // namespace libMesh
 
 // Workaround for weird boost/NumericVector interaction bug
 #ifdef LIBMESH_DEFAULT_QUADRUPLE_PRECISION
--- a/include/numerics/numeric_vector.h
+++ b/include/numerics/numeric_vector.h
@@ -789,7 +789,7 @@ public:
   /**
    * checks whether the vector \p v is fuzzy equal to this vector
    */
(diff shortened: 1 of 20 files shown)


include/numerics/numeric_vector.h
```diff
@@ -786,6 +786,15 @@ class NumericVector : public ReferenceCountedObject<NumericVector<T>>,
    */
   bool compatible(const NumericVector<T> & v) const;
 
+  /**
+   * checks whether the vector \p v is fuzzy equal to this vector by doing element-wise comparisons.
+   * A given element will be deemed fuzzy equal if either a relative or absolute tolerance fuzzy
+   * equal comparison returns true
+   */
+  bool fuzzy_equal(const NumericVector<T> & v,
```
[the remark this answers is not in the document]
roystgnr: I'd prefer to maintain that.  "If you don't know whether you want absolute or relative tolerance, you don't know what you're doing" is a tautology on one level and a critically important statement on another.  My "eh, it's all fuzzy equality" attitude in the libMesh shape-function-caching code long ago turned into a disaster when it eventually interacted with the MOOSE displaced-mesh code.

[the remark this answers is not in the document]
lindsayad: I took @jwpeterson's comment to be more aimed at the plural naming vs. the separation of `relative` and `absolute`. I've addressed that. I added distinct tolerances for relative and absolute to the APIs in this PR. I think this is better than having separate APIs. i don't care to loop over the vector or matrix twice when I can do it once. We don't run SNES twice to see whether we meet the relative convergence criteria or absolute convergence criteria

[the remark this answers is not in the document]
roystgnr: Ah, looks like I misinterpreted; thanks for clarifying.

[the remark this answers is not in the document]
roystgnr: I read "plural" as "there's two of them", and the comment on the main page cuts off the function arguments below it.  This looks good to me.

include/utils/fuzzy_equals.h
```diff
@@ -0,0 +1,203 @@
...
+ * @param tol The tolerance to be used
+ * @return true if var1 and var2 are equal within tol
+ */
+template <typename T,
+          typename T2,
+          typename T3 = T,
+          typename std::enable_if<ScalarTraits<T>::value && ScalarTraits<T2>::value &&
+                                      ScalarTraits<T3>::value,
+                                  int>::type = 0>
+bool absolute_fuzzy_equals(const T & var1, const T2 & var2, const T3 & tol = TOLERANCE * TOLERANCE);
```
roystgnr, line 41: `tol` should always conceptually be a positive real, right?  What's the use case for `const T3 &` vs just `Real`?

include/utils/fuzzy_equals.h
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * Function to check whether both the real and imaginary parts of two complex variables are equal
+ * within an absolute tolerance
+ * @param var1 The first complex variable to be checked
+ * @param var2 The second complex variable to be checked
+ * @param tol The real tolerance to be used for comparing both real and imgainary parts
```
roystgnr, line 48:
```suggestion
 * @param tol The real tolerance to be used for comparing both real and imaginary parts
```

I feel like I just successfully found the "No Brown M&Ms" clause.

include/utils/fuzzy_equals.h
```diff
@@ -0,0 +1,203 @@
...
+      return absolute_fuzzy_equals(
+          var1,
+          var2,
+          tol * (std::abs(MetaPhysicL::raw_value(var1)) + std::abs(MetaPhysicL::raw_value(var2))));
+    else if constexpr (TensorTools::TensorTraits<T>::rank == 1)
+    {
+      for (const auto i : make_range(libmesh_dim))
+        if (!relative_fuzzy_equals(var1(i), var2(i), tol))
+          return false;
+
```
roystgnr, line 113: Won't this conclude that the vectors [1e9, 1] and [1e9, 1+1e-5] aren't relatively equal?  That doesn't seem like the behavior we want.

roystgnr: The simplest comparable thing to do here is to get the l∞ norm in one pass, then use that as the scale for each difference comparison on a second pass.  We could avoid a 2-pass solution by keeping track of l∞ of both the values and the differences in a single pass and then comparing at the end.  That doesn't let us take advantage of early-return cases, but I don't think I see a way around that.

roystgnr (changes requested):

lindsayad:
I'm thinking that I might drop the `fuzzy_equals` routines for the global data structures

roystgnr:
Those are the trickiest, I admit.  That's exactly the reason we *want* library routines for them IMHO, but if you want to get the obviously-acceptable parts merged right away then I'm fine with postponing anything that will take longer.

include/utils/fuzzy_equals.h
lindsayad: I need a confused emoji

include/utils/fuzzy_equals.h
lindsayad: Does the norm we use really matter? I think I'm going to do a bit of consolidation. It looks like we're using the L1 norm in `TypeVector`

include/utils/fuzzy_equals.h
lindsayad: the poor spelling is fixed though 😄

lindsayad:
Ok I've attempted a fairly generic implementation that supports the more complicated global data structures. Let's see what you think

lindsayad:
Failures are unrelated. I think this should be ready for review

roystgnr (approved):
Where/how do we end up instantiating `TypeVector<bool>`?

lindsayad:
MOOSE solid mechanics module does it 🤷‍♂️

PR #3896, opened 2024-07-01
A couple small changes for subvectors

lindsayad:
- Take a `unique_ptr` by value when restoring instead of an rvalue of the object. I think this makes the `get` and `restore` APIs pair better. I know that this is changing a publicly available API but as @roystgnr pointed out in #3883 it's only been around for a couple months so I'm 🤞 that he and @jwpeterson will allow it. If not, nbd, I'll go through the deprecation process
- Make sure when restoring the subvector to communicate ghost data. I ran into the bad effects of not doing this when working on #3883

### DIFF
--- a/include/numerics/petsc_vector.h
+++ b/include/numerics/petsc_vector.h
@@ -349,7 +349,7 @@ public:
   virtual std::unique_ptr<NumericVector<T>>
   get_subvector(const std::vector<numeric_index_type> & rows) override;
 
-  virtual void restore_subvector(NumericVector<T> && subvector,
+  virtual void restore_subvector(std::unique_ptr<NumericVector<T>> && subvector,
                                  const std::vector<numeric_index_type> & rows) override;
 
 private:
--- a/include/numerics/petsc_vector.h
+++ b/include/numerics/petsc_vector.h
@@ -349,7 +349,7 @@ public:
   virtual std::unique_ptr<NumericVector<T>>
   get_subvector(const std::vector<numeric_index_type> & rows) override;
 
-  virtual void restore_subvector(std::unique_ptr<NumericVector<T>> && subvector,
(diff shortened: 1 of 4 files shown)


include/numerics/petsc_vector.h
```diff
@@ -349,7 +349,7 @@ class PetscVector final : public NumericVector<T>
   virtual std::unique_ptr<NumericVector<T>>
   get_subvector(const std::vector<numeric_index_type> & rows) override;
 
-  virtual void restore_subvector(NumericVector<T> && subvector,
+  virtual void restore_subvector(std::unique_ptr<NumericVector<T>> && subvector,
```
jwpeterson, line 352: I would just pass the `std::unique_ptr` by value instead of rvalue reference. This is what we do in `MeshBase::add_elem()` for example, and it [guarantees that the function takes ownership of the pointer](https://stackoverflow.com/questions/71930850/why-not-to-accept-stdunique-ptr-by-rvalue-reference).

lindsayad:
If you're ok with the API changes @roystgnr, this would be a PR that I'd like included in a libmesh update since I have multiple in-flight MOOSE PRs waiting on the libmesh update for other reasons that use these APIs

lindsayad:
So if we were to put an update in without this and then the MOOSE PRs got merged, this could no longer pass CI

lindsayad:
well ironically this is failing all the MOOSE targets, but for unrelated reasons

roystgnr (approved):

roystgnr:
If we can get this in ASAP then we should definitely update the current submodule update PR.  I'm chomping at the bit to get multiple features into MOOSE, though; if those failures are unrelated then let's merge anyway and if they're related then let me know if there's anything I can do to help fix them.

lindsayad:
I'm 99% sure they are unrelated but let's ensure the remaining 1%

PR #3898, opened 2024-07-08
Bump coverage of petscdmlibmeshimpl

lindsayad:
I wanted to check whether the hooks to this code still exist and they seem like they do. In that case, let's ensure that we have *some* coverage so that the likelihood of those hooks continuing to function over time is a bit higher


lindsayad:
we're slowly but surely inching our coverage up!

PR #3900, opened 2024-07-10
System::is_initialized() should be const

roystgnr:
I forget who pointed this out to me, but at least I didn't lose the branch I left lying around for it...


dschwen:
> I forget who pointed this out to me
✋

PR #3901, opened 2024-07-10
Fix elem dof distribution for `--node-major-dofs`

lindsayad:
I guess that this has flown under the radar since hardly anyone uses `--node-major-dofs`?


roystgnr:
... Ben Kirk used `--node-major-dofs` religiously to do some cool trickery in FIN-S, but we probably haven't had test coverage on it at any point in the last 5 years.

But this bug looks like it might *predate* that?  It could be that people have only been using `--node-major-dofs` with elements that have no elem dofs!  (Or, less likely, it could be that they've been disabling the automatic variable group creation at the same time...)

roystgnr (approved):

Issue #3899, opened 2024-07-09
Import from MSH-File: 


grunerjmeier:
Dear Community,

When importing the gmsh-file given in [1] into Moose, "PointA" gets assigned wrong coordinates. The FE elements (volumes, plates and beams) are imported fine.

[1] https://github.com/grunerjmeier/Suzanne/blob/117edc4794f077663d7b55f0292e5554ad0fb4db/suzanne.msh

As per the MSH-File, "PointA" should be to top of the head of Suzanne (coordinates '0 3.4988 9.28'). Please see image below.
https://github.com/grunerjmeier/Suzanne/blob/117edc4794f077663d7b55f0292e5554ad0fb4db/suzanne.msh#L35

Instead in Moose PointA is set to (-1.05469,  3.65505, -14.0167):
```
FileMeshGenerator 'file':  Mesh Nodesets:
FileMeshGenerator 'file':   Nodeset 1 (PointA), 1 nodes
FileMeshGenerator 'file':    Bounding box minimum: (x,y,z)=(-1.05469,  3.65505, -14.0167)
FileMeshGenerator 'file':    Bounding box maximum: (x,y,z)=(-1.05469,  3.65505, -14.0167)
FileMeshGenerator 'file':    Bounding box delta: (x,y,z)=(       0,        0,        0)
```

The msh file is imported to Moose via
```
[Mesh]
  [file]
    type = FileMeshGenerator
    file = "suzanne.msh"
    show_info = true
  []
[]
```

![image](https://github.com/libMesh/libmesh/assets/57180520/81f5d112-f193-408a-9edd-6e55f92ff169)

roystgnr:
Looks like you've got a GMSH 4.1 format file, and it's hitting our `if (version >= 4.0)` block, but people are saying [they've changed the format from 4.0 to 4.1](https://github.com/HomerReid/scuff-em/issues/184)?

If there are major changes then I don't know if we'll have time to update to them; if you could try converting your file to 4.0 then I'd love to know if it's still breaking or not and I'm sure you'd love to know if not.

But that might be a red herring - looking at [the current format](https://gmsh.info/doc/texinfo/gmsh.html#MSH-file-format) I don't see anything major out of place for point $Entities or $Elements.

[...]

PR #3903, opened 2024-07-11
Bounding box testing for $Entities in GmshIO

roystgnr:
This is the best solution we could come up with to #3899 - continue our existing behavior, but scream and die if the .msh file we're reading is definitely inconsistent with that behavior.

I'm ashamed that this is the first .msh file we've added to our test suite - LCOV confirms that our CI is not hitting `GmshIO::read()`!  I'm almost hoping that this PR triggers test failures from inconsistent meshes downstream, just so I can see that there's been *something* hitting that code.


roystgnr:
> I'm almost hoping that this PR triggers test failures from inconsistent meshes downstream, just so I can see that there's been something hitting that code.

Oh, better yet, it trigger tested failures from *consistent* meshes and let me catch a bug in it.

roystgnr:
`s/a bug/bugs/`

On the bright side, I'm getting more and more happy about our downstream coverage here.

roystgnr:
Okay, I'm not sure what's going on in the updated Mac environments, but this PR does not put a segfault in eigenproblems_ex4 and does not put compiler warnings in solid_mechanics/examples/umat_build.build

We ought to be ready to merge.

src/geom/bounding_box.C
```diff
@@ -62,6 +62,62 @@ bool BoundingBox::contains_point (const Point & p) const
 }
 
 
+
+Real BoundingBox::max_size() const
+{
+  Real size = 0;
+  for (unsigned int d = 0; d != LIBMESH_DIM; ++d)
```
lindsayad, line 69: In the future, after #3895 goes in, we could use `make_range` with `libmesh_dim` here

src/geom/bounding_box.C
```diff
@@ -62,6 +62,62 @@ bool BoundingBox::contains_point (const Point & p) const
...
+    {
+      Real sized = this->second(d) - this->first(d);
+      if (!libmesh_isinf(sized))
```
lindsayad, line 72: This would return `false` if `sized` corresponded to `std::numeric_limits<Real>::max()`, wouldn't it? I just read https://en.cppreference.com/w/cpp/numeric/math/isinf to check. If I'm correct, then the behavior of this function doesn't match its doxygen

include/geom/bounding_box.h
```diff
@@ -116,6 +116,17 @@ class BoundingBox : public std::pair<Point, Point>
...
+  /**
+   * \returns \p true if the bounding box contains the given point,
+   * to within at least one of the given tolerance(s).  At least one
+   * tolerance should be set to greater than zero; otherwise use the
+   * other \p contains_point overload for efficiency.
+   *
+   * Relative tolerances are computed relative to the maximum finite
+   * extent of the bounding box, \p max_size()
+   */
+  bool contains_point (const Point &, Real abs_tol, Real rel_tol) const;
```
lindsayad, line 128:
```suggestion
  bool contains_point (const Point & p, Real abs_tol, Real rel_tol) const;
```

lindsayad: Without it, I assume you don't actually use that argument in the implementation (which you do)

src/geom/bounding_box.C
```diff
@@ -62,6 +62,62 @@ bool BoundingBox::contains_point (const Point & p) const
...
+        size = std::max(size, sized);
+    }
+  return size;
+}
+
+
+
+bool BoundingBox::contains_point
+  (const Point & p, Real abs_tol, Real rel_tol) const
```
lindsayad, line 81:
```suggestion
  (const Point & p, const Real abs_tol, const Real rel_tol) const
```

src/geom/bounding_box.C
```diff
@@ -62,6 +62,62 @@ bool BoundingBox::contains_point (const Point & p) const
...
+  // Just use the other contains_point overload
+  libmesh_assert_greater(rel_tol+abs_tol, Real(0));
+
+  // Find absolute tolerance from relative tolerance
+  const Real tol = std::max(abs_tol, this->max_size()*rel_tol);
+
+  // Make local variables first to make things more clear in a moment
+  const Real my_min_x = this->first(0) - tol;
+  const Real my_max_x = this->second(0) + tol;
+  bool x_int = is_between(my_min_x, p(0), my_max_x);
```
lindsayad, line 95:
```suggestion
  const bool x_int = is_between(my_min_x, p(0), my_max_x);
```

lindsayad: You could early return if `x_int = false` and so on through this function if you wanted to avoid some work

src/geom/bounding_box.C
```diff
@@ -62,6 +62,62 @@ bool BoundingBox::contains_point (const Point & p) const
...
+
+  bool intersection_true = x_int;
+
+#if LIBMESH_DIM > 1
+  const Real my_min_y = this->first(1) - tol;
+  const Real my_max_y = this->second(1) + tol;
+  bool y_int = is_between(my_min_y, p(1), my_max_y);
```
lindsayad, line 102:
```suggestion
  const bool y_int = is_between(my_min_y, p(1), my_max_y);
```

src/geom/bounding_box.C
```diff
@@ -62,6 +62,62 @@ bool BoundingBox::contains_point (const Point & p) const
...
+
+  intersection_true = intersection_true && y_int;
+#endif
+
+
+#if LIBMESH_DIM > 2
+  const Real my_min_z = this->first(2) - tol;
+  const Real my_max_z = this->second(2) + tol;
+  bool z_int = is_between(my_min_z, p(2), my_max_z);
```
lindsayad, line 111:
```suggestion
  const bool z_int = is_between(my_min_z, p(2), my_max_z);
```

src/mesh/gmsh_io.C
```diff
@@ -625,19 +641,50 @@ void GmshIO::read_mesh(std::istream & in)
...
                     {
+                      Node * node = mesh.node_ptr(nodetrans[gmsh_node_id]);
+
+                      // Make sure the file is consistent about entity
+                      // placement.  Well, mostly consistent.  We have
+                      // files that claim a bounding box but still
+                      // have points epsilon outside it.
+                      libmesh_error_msg_if
+                        (!expected_bounding_box.contains_point
+                           (*node, 0, /* relative */ TOLERANCE),
```
lindsayad, line 672:
```suggestion
                           (*node, /* abs */ 0, /* relative */ TOLERANCE),
```

src/mesh/gmsh_io.C
```diff
@@ -665,8 +712,37 @@ void GmshIO::read_mesh(std::istream & in)
...
+                      expected_bounding_box = it->second;
+
+                    Node * node = mesh.node_ptr(nodetrans[gmsh_node_id]);
+
+                    // We'll accept *mostly* consistent.  We have
+                    // files that claim a bounding box but still have
+                    // points epsilon outside it.
+                    libmesh_error_msg_if
+                      (!expected_bounding_box.contains_point
+                         (*node, 0, /* relative */ TOLERANCE),
```
lindsayad, line 737:
```suggestion
                         (*node, /* abs */ 0, /* relative */ TOLERANCE),
```

src/mesh/gmsh_io.C
```diff
@@ -665,8 +712,37 @@ void GmshIO::read_mesh(std::istream & in)
...
+                       "$Elements dim " << entity_dim << " element "
+                       << gmsh_element_id << " (entity " << entity_tag <<
+                       ") has node at " << *static_cast<Node*>(node)
+                       << "\n outside entity physical bounding box " <<
+                       expected_bounding_box);
```
lindsayad, line 742: Seems like two repetitions of this code. *Optionally* could add a method/lambda

src/geom/bounding_box.C
roystgnr: We generally don't pass parameters by `const` value.  `const` on a reference parameter is making a promise to the user, but `const` on a value parameter means nothing to the API semantics (because the value is a copy or a temporary anyway depending on input, and in neither case will the calling code ever see it) but can require future changes to the implementation to change the API syntax.

src/geom/bounding_box.C
roystgnr: Yeah, I think you're right.  Pushed a fix.

src/geom/bounding_box.C
lindsayad: You can leave it non-const in the header declaration for the reason you stated but mark it as `const` in the implementation as that helps communicate to the reader that the parameters (at least in the current implementation) do not change

lindsayad (approved):

PR #3904, opened 2024-07-16
Revert "TypeVector overloads for std::norm and friends for Eigen::Matrix

lindsayad:
This reverts commit f23f44a06df66a1f10ce0539a76623f4bbf34eb1.

### DIFF
--- a/include/numerics/type_vector.h
+++ b/include/numerics/type_vector.h
@@ -30,37 +30,6 @@
 #include <cmath>
 #include <complex>
 
-#ifdef LIBMESH_HAVE_EIGEN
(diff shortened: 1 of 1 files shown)


jwpeterson (approved):
I guess this didn't break anything in MOOSE... did you guys find some other workaround for whatever caused [f23f44a](https://github.com/libMesh/libmesh/commit/f23f44a06df66a1f10ce0539a76623f4bbf34eb1) to be needed in the first place?

lindsayad:
we have first-class support for Eigen data structures holding dual numbers now. Additionally, I don't think this was ever the right fix having these forward declarations here that don't involve `TypeVector`

PR #3909, opened 2024-07-23
Don't leak temporary Netgen LDFLAGS into libmesh_LDFLAGS

roystgnr:
Thanks to @lindsayad for catching this - it looks like it was hitting my Netgen-enabled MOOSE build too but somehow the build was completing anyway??


jwpeterson:
Error message according to downloading the CIVET output:
```
/opt/rh/gcc-toolset-12/root/usr/libexec/gcc/x86_64-redhat-linux/12/ld: cannot find -lnglib: No such file or directory
/opt/rh/gcc-toolset-12/root/usr/libexec/gcc/x86_64-redhat-linux/12/ld: cannot find -lngcore: No such file or directory
make[1]: *** [Makefile:13041: libmesh_opt.la] Error 1
make[1]: Leaving directory '/data/civet1/build/libmesh/build'
make: *** [Makefile:33870: all-recursive] Error 1
```

roystgnr:
Yeah, I can reproduce that.  Looks like I just forgot an AC_SUBST for the new variable.

lindsayad (approved):

Issue #3157, opened 2022-02-10
Recent gcc won't let us use packed ranges with TypeVector, Point


roystgnr:
Declaring a `std::map<processor_id_type , std::vector<TypeVector<Real>>> vals` and then trying to perform a parallel `set_union(vals)` is enough to demonstrate the problem.  `set_union` correctly detects that we need to turn the `map::value_type` into a packed range, our `Packing` overload for `std::vector` correctly detects `IsFixed<TypeVector>::value` and correctly concludes that it's safe to pack each individual `TypeVector` with a `memcpy` ... and then gcc screams a warning at us about `copying an object of non-trivial type`.

`TypeVector` is indeed not TriviallyCopyable, and the [C++ standard correctly says that should be undefined behavior](https://en.cppreference.com/w/cpp/string/byte/memcpy) for `std::memcpy`, so I tried giving `TypeVector` a default copy constructor and default destructor (both of which are trivial), and I verified that it then passed `is_trivially_copyable` static assertions ...

And that's not enough.  GCC doesn't just warn about `!is_trivially_copyable`, it warns about `!is_trivial`, and [they won't fix that](https://gcc.gnu.org/legacy-ml/gcc-patches/2017-07/msg00299.html), and we can't satisfy that (because we have our default constructors zero out their data).  I don't want to just pragma away the warning here either, because when it's not a false positive it's actually super useful.

My plan B was going to be to manually add `Packing` overloads for such classes, but as I type this out I'm thinking instead that we might just insist on more precise warnings manually: in TIMPI packing.h we'd cast our `T*` to `char*` to avoid gcc's warning but we'd instead add an `is_trivially_copyable` static assertion to give us a saner warning, and then in libMesh classes we'd be able to get away with just satisfying `is_trivially_copyable`.

Pinging @bwspenc because he caught the problem today and because I expect him to remind me about this issue next week when I have time to work on it.

roystgnr:
Okay - the problem is now fixed for TypeVector and Point, but it's not fixed in general: the C++ standard doesn't mandate that `is_trivially_copyable<tuple<SomethingTriviallyCopyable>>::value == true`, and that's too important a case for future use, so we'll need our own metafunction instead.  We'll have `ActuallyTriviallyCopyable` default to false, specialize it to be true whenever `is_trivially_copyable` is, but then add specializations to our own metafunction for whichever containers might give us false negatives from the `std::` metafunction.

[...]

PR #3910, opened 2024-07-30
Fixes for compilations using single/triple/quadruple precision

nmnobre:
Hey,

This completely fixes single/triple precision compilations for me.

Compilations using quadruple precision, however, will still fail:
1) The boost installation we ship with libMesh is missing some necessary definitions:
`./include/libmesh/point.h:104:73: error: use of deleted function ‘std::hash<boost::multiprecision::number<boost::multiprecision::backends::float128_backend, boost::multiprecision::et_off> >::hash()’`
2) Even with the latest version of boost, we have an incompatible assertion (`__float128` is trivially copyable but boost's wrapper unfortunately isn't? After writing that, I checked, and the docs claim it should be...?! boostorg/multiprecision#635):
`../src/geom/point.C:28:80: error: static assertion failed: Someone made Point non-TriviallyCopyable`

Cheers,
-Nuno

### DIFF
--- a/src/apps/embedding.C
+++ b/src/apps/embedding.C
@@ -179,7 +179,7 @@ int main(int argc, char ** argv)
             }
           else
             {
-              int oldnumerator = std::round(shape*denominator);
+              int oldnumerator = int(std::round(shape*denominator));
               int newnumerator = oldnumerator;
               int newdenominator = denominator;
 
--- a/configure
+++ b/configure
@@ -50525,14 +50525,16 @@ fi
 fi
 
 
-          if test "x$enableboost" = "xyes"
+              if test "x$enableboost" = "xyes"
 then :
 
           if test "x$install_internal_boost" = "xyes"
 then :
   libmesh_contrib_INCLUDES="$BOOST_INCLUDE $libmesh_contrib_INCLUDES"
+                 export timpi_CPPFLAGS="-I$(realpath $top_srcdir)/contrib/boost/include"
 else $as_nop
   libmesh_optional_INCLUDES="$BOOST_CPPFLAGS $libmesh_optional_INCLUDES"
+                 export timpi_CPPFLAGS="$BOOST_CPPFLAGS"
 fi
 
 fi
--- a/configure
+++ b/configure
@@ -50531,10 +50531,10 @@ then :
           if test "x$install_internal_boost" = "xyes"
 then :
   libmesh_contrib_INCLUDES="$BOOST_INCLUDE $libmesh_contrib_INCLUDES"
-                 export timpi_CPPFLAGS="-I$(realpath $top_srcdir)/contrib/boost/include"
+                 export timpi_CPPFLAGS="$timpi_CPPFLAGS -I$(realpath $top_srcdir)/contrib/boost/include"
 else $as_nop
   libmesh_optional_INCLUDES="$BOOST_CPPFLAGS $libmesh_optional_INCLUDES"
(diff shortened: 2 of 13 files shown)


jwpeterson:
Is the `static_assert` there essential to have? If not, maybe we could just `ifdef` it out for higher-precision builds.

nmnobre:
> Is the `static_assert` there essential to have? If not, maybe we could just `ifdef` it out for higher-precision builds.

I don't know why it's there. :)

roystgnr:
The reason why it's there is discussed (in rambling fashion) in https://github.com/libMesh/libmesh/issues/3157

Basically: in TIMPI we're copying certain things into buffers via `reinterpret_cast<char*>` followed by `memcpy`, and we just test for a `StandardType<Thing>` to see if we're allowed to do that, so we don't want to accidentally define `StandardType<Thing<T>>` unless `T` is trivially copyable.

Unfortunately, a `T` can be trivially copyable *without* `is_trivially_copyable<T>::value==true`, because there are loads of ways (almost any way) to write a constructor that can convince C++ that something non-trivial is going on without actually breaking `memcpy`.  Sounds like Boost is hitting one of them now?  If we're sure this is an overzealous assert then we can remove it; I just wish there was a way to assert what we actually want to assert here.

src/apps/embedding.C
```diff
@@ -179,7 +179,7 @@ int main(int argc, char ** argv)
             }
           else
             {
-              int oldnumerator = std::round(shape*denominator);
```
roystgnr, line 182: This is wild.  Does boost 128 convert to int explicitly but not implicitly?

roystgnr (approved):

src/apps/embedding.C
nmnobre: Indeed, https://godbolt.org/z/ve9zEaqza.

nmnobre:
> Sounds like Boost is hitting one of them now?

Yep. Assuming I can get boostorg/multiprecision#636 merged, the newest commit, [3e80340](https://github.com/libMesh/libmesh/pull/3910/commits/3e80340fb522aee3e7e954258fe5a2c7e4c204ac), fixes compilations using quadruple precision if the user disables both fparser (linking issue 'cause fparser doesn't know about boost's float128, so it never specialises anything for that type) and MetaPhysicL (because we check at points if the scalar type is built-in, and it obviously isn't marked as such https://github.com/libMesh/MetaPhysicL/blob/e67ec423140498ed65e5824c4de0f9f7143bbbdc/src/core/include/metaphysicl/compare_types.h#L163-L177).

roystgnr:
Could you cherry-pick my last two commits from the roystgnr/fixes branch?  I don't have permission to push to yours.  That'll fix a minor issue I noticed and hopefully also wake up Civet.

nmnobre:
> Could you cherry-pick my last two commits from the roystgnr/fixes branch? I don't have permission to push to yours. That'll fix a minor issue I noticed and hopefully also wake up Civet.

Thanks, well noticed. :)

roystgnr:
Most of our tests get run with 64-bit integers; that one with 32-bit.  It's also one of the places where we run tests in dbg mode.

nmnobre:
Thanks, I see, you change `id size (dofs)`.

roystgnr:
We really ought to be testing the variations on `subdomain_id_type` and `boundary_id_type` too ... IIRC at least one of those has been bumped up by users recently, and they're easier to test since they don't require any changes in PETSc for full coverage.

PR #3915, opened 2024-08-02
Implement improved ASPECT_RATIO quality metric for Quad elements

jwpeterson:
The previous `ASPECT_RATIO` implementation had a couple of shortcomings:
1. Rhombus-like quadrilaterals with arbitrarily small internal angles still reported "perfect" aspect ratio = 1.
2. Degenerate quadrilaterals with one zero-length side previously reported 0 aspect ratio, now this should only happen if the quadrilateral actually has 0 area.

Both of these issues are addressed by the current PR.


roystgnr (changes requested):
The previous `ASPECT_RATIO` is obviously inferior to this one in a mathematical sense, but while one of the things I like about this one in a practical sense is "matches ANSYS", I think the previous one's "matches CUBIT" is worth something too.

Can we either add the new metric as something like `ASPECT_RATIO_ANSYS` (which I prefer from a backwards-compatible-results standpoint), or add the new `ASPECT_RATIO` but retain the old metric as something like `ASPECT_RATIO_CUBIT` or `EDGE_LENGTH_RATIO` (which I prefer from a usability, "give you the best option consistent with the name" sense)?  Slightly leaning toward the latter but I'd be happy with either.

roystgnr (approved):

PR #3917, opened 2024-08-06
Smoother unit tests

roystgnr:
This should help a bit with avoiding regressions in LaplaceMeshSmoother, and hopefully with fixing the regressions (which are worse than I thought? I was expecting corrupt output but got a segfault) in VariationalMeshSmoother.

Pinging @oanaoana

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -73,6 +73,7 @@ unit_tests_sources = \
   mesh/mesh_function_dfem.C \
   mesh/mesh_generation_test.C \
   mesh/mesh_input.C \
+  mesh/mesh_smoother_test.C \
   mesh/mesh_stitch.C \
   mesh/mesh_tet_test.C \
   mesh/mesh_triangulation.C \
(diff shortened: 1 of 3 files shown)


roystgnr:
Not sure why CI hasn't been getting all the github triggers here...

jwpeterson (approved):
When I first saw the PR title, I thought you were making our unit tests run "smoother" (i.e. better) but this is good too :)

roystgnr:
Well, now you've got me wondering if it's possible to do that too...  I'm picturing something where, although `unit_test_nprocs` *defaults* to whatever `TestCommWorld->size()` is, we can instead do e.g. `LIBMESH_RUN='mpirun -np 64' UNIT_TEST_NPROCS=4` to run 16 tests at a time... and then we have a hideous macro in each unit test that creates the appropriate subcommunicator for that test ... and makes rank 0 wait for other ranks' results sent via asynchronous MPI calls ... and the more I think about this idea the dumber it gets.

PR #3918, opened 2024-08-07
Fix for Tri7::is_face()

jwpeterson:
This makes it consistent with Quad9::is_face() which returns true for the central Node of the Quad9.

Also improve the documentation for Elem::is_face() and Elem::is_edge() to make our conventions clear for elements of specific dimension.

### DIFF
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -730,11 +730,14 @@ public:
 
   /**
    * \returns \p true if the specified (local) node number is an edge node.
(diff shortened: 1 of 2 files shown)


jwpeterson:
@roystgnr Any changes on the comments I added here? I don't think we had any unit test coverage of `Tri7::is_face()`, and likely this was not called in MOOSE, but I will be calling it in a subsequent PR.

roystgnr (approved):

PR #3919, opened 2024-08-07
Better error when asked to tetrahedralize a broken boundary

roystgnr:
In https://github.com/idaholab/moose/pull/28298#discussion_r1707391079 @miaoyinb reported a segfault from such user input, and the naked assertion failure we'd report in dbg/devel modes wasn't much better.  Hopefully the new error message here is meaningful enough for most users.


roystgnr:
Two completely different DistributedMesh test failures.  At least this one was only a bug in the test itself.

PR #3920, opened 2024-08-08
Add Elem::edges_adjacent_to_node() API

jwpeterson:
This function is a companion to `Elem::nodes_on_edge()` in that, given an elem-local node id, it returns the elem-local edge ids that are adjacent to that node. This new API is not tested yet, but I'm in the process of working on another PR that will use it in a generic implementation of the `MIN_ANGLE` and `MAX_ANGLE` `ElemQuality` metrics, and it was self-contained/at a good enough stopping point to at least be checked by CI.

### DIFF
--- a/include/geom/cell_hex.h
+++ b/include/geom/cell_hex.h
@@ -181,6 +181,8 @@ public:
 
   virtual bool is_flipped() const override final;
 
+  virtual std::vector<unsigned int> edges_adjacent_to_node(const unsigned int n) const override;
+
   /**
    * This maps each edge to the sides that contain said edge.
    */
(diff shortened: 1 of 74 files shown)


roystgnr (approved):
I'm okay with this as-is, but I'd really love to see a new method in tests/geom/elem_test.C for it, even if it's just a check for consistency with `Elem::nodes_on_edge`.

jwpeterson:
> a check for consistency with Elem::nodes_on_edge.

That's a good idea, I'll add such a test after I've finished the MIN/MAX_ANGLE Elem quality metric update I'm currently working on.

PR #3922, opened 2024-08-09
Add generic `Elem::quality(MIN,MAX_ANGLE)` implementation

jwpeterson:
* Add specific unit tests for `MIN_ANGLE` and `MAX_ANGLE` metric for Quads.
* Add unit test that calls the new `Elem::quality()` implementation for all element types in the `elem_test.C` unit test suite, and check that the return values make sense.
* As requested in #3920, this PR also adds a unit test of the basic `Elem::edges_adjacent_to_node()` functionality that was added in #3920.

### DIFF
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -1613,6 +1613,88 @@ Real Elem::quality (const ElemQuality q) const
           return *max / *min;
       }
 
+    case MIN_ANGLE:
+    case MAX_ANGLE:
+      {
+        // 1D elements don't have interior angles, so just return some
+        // dummy value in that case.
+        if (this->dim() < 2)
+          return 0.;
+
+        // Initialize return values
+        Real min_angle = std::numeric_limits<Real>::max();
+        Real max_angle = -std::numeric_limits<Real>::max();
+
+        for (auto n : this->node_index_range())
+          {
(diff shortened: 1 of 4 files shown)


jwpeterson:
@roystgnr this should be ready for review. Your suggestion to check that `Elem::edges_adjacent_to_node()` was consistent with `Elem::nodes_on_edge()` was a good one, looks like we have 100% coverage of the new code through the elem_test.C unit tests.

roystgnr (approved):
This is all good, except that now that we're expanding from 2D to 3D I think we need some documentation making it clear that `MIN_ANGLE` and `MAX_ANGLE` are referring to edge angles in the 3D case, not (more commonly used in the literature I've seen) dihedral angles.

I'm not sure where to document that, though.  The Elem::quality() doxygen?  Probably the Quality::describe() output?

jwpeterson:
Hmm... the `Elem::quality()` documentation is very sparse, possibly because not all element types support all the quality metrics. But now that we have a couple generic ones that *are* supported for all elements, it would probably be worth documenting that (including exactly what angles are checked for the `MIN/MAX_ANGLE` metric). I'll also update the ElemQuality::describe() documentation, though to be perfectly honest I have never looked at that before and kind of forgot it existed...

PR #3925, opened 2024-08-14
Fix bad calls to LibmeshPetscCall when we don't have this->comm()

lindsayad:
For the static method we don't have a communicator from anyone so we choose to except/abort across all ranks

### DIFF
--- a/examples/eigenproblems/eigenproblems_ex4/eigenproblems_ex4.C
+++ b/examples/eigenproblems/eigenproblems_ex4/eigenproblems_ex4.C
@@ -707,7 +707,7 @@ form_matrixA(SNES /*snes*/, Vec x, Mat jac, Mat pc, void * ctx)
     PetscMatrix<Number> sub(pc, eigen_system.comm());
     eigen_system.copy_super_to_sub(pc_super, sub);
 #else
-    auto sub = PetscMatrixBase<Number>::get_context(pc);
+    auto sub = PetscMatrixBase<Number>::get_context(pc, eigen_system.comm());
     libmesh_assert(sub);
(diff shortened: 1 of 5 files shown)


roystgnr:
Being able to use subcommunicators is a real use case.  I don't know anyone who's used subcommunicators *and* caught `LogicError` to recover the jobs on them, I admit, but it's at least a "real theoretical use case" I've seen discussed from people interested in hardware reliability on large distributed systems.

If we don't want to make these methods non-static, could we use https://petsc.org/main/manualpages/Sys/PetscObjectGetComm/ to try to get the communicator associated with `mat`?  And if *that* doesn't work (or if it seems pointless to you, in the context of a `mat` that's failing PETSc calls...) then fall back on `libMesh::GLOBAL_COMM_WORLD` so we at least get any subcommunicator that's been set at the `LibMeshInit` level?

lindsayad:
When I was away from comp, I decided that `PetscObjectGetComm` was the right thing to do here. we're on the same page

lindsayad:
although I am going to use the world communicator for checking the ierr from `PetscObjectGetComm` 😄

lindsayad:
the other option would be to add a communicator argument to the static method. I am not going to make the method non-static because the point of this method is to recover the libMesh wrapper object for the Mat object, so chicken-and-egg

roystgnr (approved):

lindsayad:
Mac failure is independent of this PR and will be fixed by https://github.com/libMesh/libmesh/pull/3924/commits/cfaa3ca45385d22403e427cad972012a127f20e9

PR #3927, opened 2024-08-14
Add generic `Elem::quality(MIN,MAX_DIHEDRAL_ANGLE)` implementation

jwpeterson:
After some discussion with @roystgnr, it seems clear to me that, for 3D elements, the dihedral angles (angles between adjacent Elem sides, where the sides are approximated by planes) are a more fundamental measure of Elem quality than the edge angles implemented in #3922. For example, it is easy to imagine a "squashed" tetrahedral element that has "nice" edge angles, but arbitrarily small dihedral angles, while the converse is not possible. This PR adds new `ElemQuality` enumerations `MIN_DIHEDRAL_ANGLE` and `MAX_DIHEDRAL_ANGLE` and a generic implementation for computing them, which falls back on the `MIN_ANGLE` and `MAX_ANGLE` code for 2D Elems. The new metrics are also unit tested for all Elem types, and give the expected results.

### DIFF
--- a/include/enums/enum_elem_quality.h
+++ b/include/enums/enum_elem_quality.h
@@ -38,6 +38,8 @@ enum ElemQuality : int {
                   SHAPE,
                   MAX_ANGLE,
                   MIN_ANGLE,
+                  MAX_DIHEDRAL_ANGLE,
+                  MIN_DIHEDRAL_ANGLE,
                   CONDITION,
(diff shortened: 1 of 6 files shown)


jwpeterson:
Verified that the "Test mac" failure is the `-Werror` one that is being fixed soon.

PR #3928, opened 2024-08-14
Don't build unit tests with plain "make"

roystgnr:
This is still more consistent than we were before #3911, but now it'll be consistent in the other direction: only build unit tests as part of `make check` before running them or as individually directed via e.g. `make unit_tests-devel`.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -220,7 +220,6 @@ if LIBMESH_ENABLE_FPARSER
 endif
 
 check_PROGRAMS = # empty, append below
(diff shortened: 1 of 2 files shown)


roystgnr:
Ooh, CI doesn't seem to have kicked off automatically.  We'll see if Logan's script gets it going.

jwpeterson (approved):
Aside from tests failing for unrelated reasons, this looks good to me, would be good to merge ASAP since I'd like it in place for our next internal libmesh update.

PR #3930, opened 2024-08-16
Netgen handling for unprepared inputs, non-contained holes

roystgnr:
If we have an unprepared input mesh we want to prepare it, not die in confusing ways inside Netgen code.

If we have a hole that definitely isn't contained in the boundary we're trying to tetrahedralize, we want to give a sensible error message, not die in confusing ways inside Netgen code.

Refs https://github.com/idaholab/moose/pull/28298

### DIFF
--- a/include/geom/bounding_box.h
+++ b/include/geom/bounding_box.h
@@ -59,7 +59,8 @@ public:
   }
 
   /**
-   * Sets the bounding box to encompass the universe.
+   * Sets the bounding box to be "inverted".  This is a useful
+   * starting point for future \p union_with operations.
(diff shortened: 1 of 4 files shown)


roystgnr:
"If it compiles for you locally, no need to re-run all the tests" - @jwpeterson, regarding that last commit

I appreciate that (maybe we can squeeze this into https://github.com/idaholab/moose/pull/28411 under the wire?), but I'm quoting for the record here, so that if it turns out that our min clang isn't as happy as gcc 11 then we can *both* feel foolish.

PR #3932, opened 2024-08-18
Make sure to destroy PetscShellMatrix `_mat`

lindsayad:
With the change in #3914 to make the `_mat` data member the same as the `PetscMatrix(Base)` class, the manual destruction is now required. This refs the valgrind failures on idaholab/moose#28411

### DIFF
--- a/include/numerics/petsc_shell_matrix.h
+++ b/include/numerics/petsc_shell_matrix.h
@@ -78,7 +78,7 @@ public:
 
   PetscShellMatrix (const Parallel::Communicator & comm_in);
 
-  virtual ~PetscShellMatrix () = default;
+  virtual ~PetscShellMatrix ();
 
   virtual numeric_index_type m () const override;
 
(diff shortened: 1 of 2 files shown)


roystgnr (approved):

roystgnr:
Any chance we could accomplish this with a `WrappedPetsc<Mat>`?

lindsayad:
That's what we used to do for this class and at one point while working on #3914 I switched `_mat` in the `PetscMatrix(Base)` to also be a `Wrapped` class. However, this didn't play nicely with some PETSc APIs for a "good" technical reason (unfortunately which I can't remember) so I changed both back to not be wrapped. This is something I could try exploring again in the future so at least I can publicly share why it doesn't work (or realize that it should work)

lindsayad:
It's coming back to me a little. There was definitely infinite recursion, I believe with `operator=` but I that bit I am fuzzy on

PR #3933, opened 2024-08-19
Update the netgen submodule

nmnobre:
At least on my machine, this change is sufficient to configure/build libMesh from scratch with gcc 14. :)

### DIFF
--- a/configure
+++ b/configure
@@ -55401,6 +55401,13 @@ then :
 
                   NETGEN_INCLUDE="-I\$(top_srcdir)/contrib/netgen/ -I\$(top_builddir)/contrib/netgen/build/"
                   NETGEN_LIBS="-lnglib -lngcore"
+                  if test "x$GCC" = "xyes"
+then :
+  if test "$($CC -dumpversion | cut -d'.' -f1)" = "8"
+then :
+  NETGEN_LIBS="$NETGEN_LIBS -lstdc++fs"
+fi
+fi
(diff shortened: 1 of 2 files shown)


roystgnr:
We need an extra linker line here?  See, *that's* the sort of thing that had me worry about the update.  It's already failing `Test ARM mac` with:

`ld: library not found for -lstdc++fs`

I'm kind of confused, honestly.  A C++ compiler invoked as a linker shouldn't need an extra `-l` to link to a standard C++ library, and IIRC we've been using `std::filesystem` at the MOOSE level for a while without any extra linker arguments.

We haven't yet been requiring compiler support for `std::filesystem` at the libMesh level.  Maybe this is the time to start, if there aren't any Netgen commits that support g++ 14 but don't require it?

nmnobre:
> We need an extra linker line here? See, _that's_ the sort of thing that had me worry about the update. It's already failing `Test ARM mac` with:
>
> `ld: library not found for -lstdc++fs`

I noticed. :/

> I'm kind of confused, honestly. A C++ compiler invoked as a linker shouldn't need an extra `-l` to link to a standard C++ library, and IIRC we've been using `std::filesystem` at the MOOSE level for a while without any extra linker arguments.

With gcc 8? I think that's the issue here.

> We haven't yet been requiring compiler support for `std::filesystem` at the libMesh level. Maybe this is the time to start, if there aren't any Netgen commits that support g++ 14 but don't require it?

I just switched to the tip on master, so I doubt it.

roystgnr:
Oh, it's just a very-old-gcc thing?  Cool, then.

And I guess this doesn't *technically* mean we're requiring `std::filesystem` now, then, just requiring it for one optional contrib package, which I'm definitely fine with.

roystgnr (approved):

PR #3934, opened 2024-08-19
Avoid SIGFPE in HDF5 1.14.3 init

roystgnr:
This was a short-lived bug (nonexistent in 1.14.2, fixed in 1.14.4) but it's biting MOOSE users right now and we're having trouble (for other reasons) downgrading or upgrading to a non-buggy version

This patch fixes the problem for me, as tested with `LIBMESH_OPTIONS='--re ExodusTest_EDGE2 --enable-fpe' make check -C tests` before and after.

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -49,8 +49,9 @@ extern "C" {
 
 // C++ includes
 #include <algorithm>
-#include <sstream>
+#include <cfenv> // workaround for HDF5 bug
 #include <cstdlib> // std::strtol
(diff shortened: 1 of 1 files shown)


jwpeterson (approved):

roystgnr:
I think I'm going to factor out that "temporarily disable SIGFPE" class under a better name later; it looks like it'll be useful in other places, too, if only to ensure that we don't trigger a simple FPE in places where we would normally throw an exception with a nice error message instead.

But for now I'll settle for "not getting broken by an HDF5 bug".

PR #3935, opened 2024-08-22
Expose FPEDisabler class, get `make check` running FPE-clean

roystgnr:
Well, mostly FPE-clean, anyway.  We have two examples where I had to use the disabler for at least one run condition.


jwpeterson (approved):
I would probably have left the comments about why `FPEDisabler` is needed for HDF 1.14 in exodusII_io_helper.C, but otherwise looks good to me.

PR #3940, opened 2024-08-29
Disable -march=native in Netgen

roystgnr:
This should definitely not be enabled by default, as it causes builds compiled on newer architecture versions to die with illegal instruction errors on older architecture versions.  And since Netgen has static objects with non-trivial constructors, we can see those failures even in codes that don't ever call Netgen!

### DIFF
--- a/configure
+++ b/configure
@@ -55395,7 +55395,7 @@ then :
                                         mkdir -p contrib/netgen/install
 
                                         if (cd contrib/netgen/build && \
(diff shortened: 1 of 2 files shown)


roystgnr:
Not going to kick those failing schemadiff tests; Logan assures me that's just a problem in MOOSE next.

roystgnr:
Civet went down for IT reasons before we could finish that last 1% of testing, but I like my odds.

PR #3942, opened 2024-08-30
Avoid misleading warning about actually used cmd line options

nmnobre:
This patch avoids misleading warnings for two IO options, `-keep-cout` and `--drop-cerr`, by checking for the command line option on all processors (including the zeroth).

```
WARNING! There are options you set that were not used!
WARNING! could be spelling mistake, etc!
There is one unused database option. It is:
Option left: name:--keep-cout (no value) source: command line
```

### DIFF
--- a/src/base/libmesh.C
+++ b/src/base/libmesh.C
@@ -626,16 +626,16 @@ LibMeshInit::LibMeshInit (int argc, const char * const * argv,
   // redirect libMesh::out to nothing on all
   // other processors unless explicitly told
   // not to via the --keep-cout command-line argument.
-  if (libMesh::global_processor_id() != 0)
-    if (!libMesh::on_command_line ("--keep-cout"))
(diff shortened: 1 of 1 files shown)


nmnobre:
> I thought this was a warning that came from PETSc

I think it is, but only because we're passing it on to PETSc by failing to consume it within libMesh.

roystgnr (approved):
Good catch.  I saw the issue and thought this must be a regression; it didn't occur to me that we were only calling `on_command_line` (and hence `add_command_line_name`) on a subset of processors.

roystgnr:
Test failures are from a snafu downstream

nmnobre:
Note to future self, we didn't see this in MOOSE because we're calling `libMesh::add_command_line_name()` there as well:

https://github.com/idaholab/moose/blob/f5681a9d7aa9ca34b44531bf6331eb7ae4d24886/framework/include/utils/InputParameters.h#L1569-L1586

PR #3944, opened 2024-09-04
Handle constraint operator empty+zero-sum columns

roystgnr:
If our heuristic for where to place a new NodeElem doesn't work out, then we can use an alternative, but if the operator itself has a problem then we want to report that and abort.


jwpeterson (approved):
It looks like we already have unit testing for this, nice. Is there a MOOSE Issue that triggered this fix? If so, it would probably be good to tag it here.

roystgnr:
There isn't; the breakage was something I hit with a new operator matrix I just got from Coreform.  In theory we should be ready for other users to start playing with this stuff as of a couple months ago; on our official timetable we deliver a beta version in December and I'm hoping to catch any bugs in the pipeline myself before we encourage anybody but power users to join in.

PR #3947, opened 2024-09-04
Minor addition in EimVarGroupPlottingInfo struct

dknez:
Added the option to enforce min or max values in the EimVarGroupPlottingInfo struct.

### DIFF
--- a/include/reduced_basis/rb_eim_evaluation.h
+++ b/include/reduced_basis/rb_eim_evaluation.h
@@ -52,6 +52,23 @@ class Elem;
  */
 struct EIMVarGroupPlottingInfo
 {
+
+  /**
+   * Default constructor. Initialize values that would otherwise
+   * be uninitialized. We do not initialize the FEType or string
+   * members since they already have default constructors.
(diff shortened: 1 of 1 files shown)


jwpeterson (approved):

jwpeterson:
Unrelated to this PR, but @roystgnr any idea what is causing these `SCHEMADIFF` failures only on the ARM recipe?
```
[0.647s]     DIFF meshgenerators/rename_block_generator.rename/all_ids FAILED (SCHEMADIFF)
```

roystgnr:
Apparently conda put out a new version of deepdiff, the software we base our schemadiff tests on, that broke us.  It didn't break our containerized CI recipes, where the whole system is fixed, but with conda we don't pin the version number on every package we request, so recipes using conda broke.  We then fixed the incompatibility in MOOSE, but a different PR broke MOOSE next so it took way too long for the schemadiff fixes to make it into master.

Anyway, it should all be fixed as of this morning, so ignore older failures but scream if you see it again.

PR #3948, opened 2024-09-05
Sanitizer fixes

roystgnr:
I'm trying out various `-fsanitize` options on various compiler versions, and some are triggering on what looks like actual Undefined Behavior.  Let's fix that.

There'll be more to come, but I'd like to get a run going in Civet right away to see how the MOOSE schemadiff problem is currently affecting our CI.

### DIFF
--- a/include/geom/cell_hex.h
+++ b/include/geom/cell_hex.h
@@ -42,7 +42,7 @@ public:
    * parent. Derived classes implement 'true' elements.
    */
   Hex(const unsigned int nn, Elem * p, Node ** nodelinkdata) :
-    Cell(nn, Hex::n_sides(), p, _elemlinks_data, nodelinkdata)
+    Cell(nn, num_sides, p, _elemlinks_data, nodelinkdata)
   {
     // Make sure the interior parent isn't undefined
     if (LIBMESH_DIM > 3)
@@ -66,6 +66,13 @@ public:
                  _master_points[i][1],
                  _master_points[i][2]);
   }
+
+  /**
+   * Geometric constants for all Hexes
+   */
+  static const int num_sides = 6;
+  static const int num_edges = 12;
+  static const int num_children = 8;
 
   /**
    * \returns 6.
--- a/include/geom/cell_hex20.h
+++ b/include/geom/cell_hex20.h
@@ -76,7 +76,7 @@ public:
    */
   explicit
   Hex20 (Elem * p=nullptr) :
-    Hex(Hex20::n_nodes(), p, _nodelinks_data)
+    Hex(num_nodes, p, _nodelinks_data)
   {}
 
   Hex20 (Hex20 &&) = delete;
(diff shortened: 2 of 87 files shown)


include/geom/cell_hex.h
```diff
@@ -42,7 +42,7 @@ class Hex : public Cell
    * parent. Derived classes implement 'true' elements.
    */
   Hex(const unsigned int nn, Elem * p, Node ** nodelinkdata) :
-    Cell(nn, Hex::n_sides(), p, _elemlinks_data, nodelinkdata)
+    Cell(nn, num_sides, p, _elemlinks_data, nodelinkdata)
```
jwpeterson, line 45: So, what exactly was the undefined behavior here? Can you not call `Hex::n_sides()` from the `Hex` constructor or something?

roystgnr: From g++ 11.4 with `-fsanitize=address -fsanitize=undefined`:

```
../include/libmesh/cell_hex.h:45:26: runtime error: member call on address 0x614000006870 which does not point to an object of type 'Hex'
0x614000006870: note: object has invalid vptr
 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00
```

I'm not 100% sure this is undefined behavior; UbSan has had bugs with this warning in the past (albeit they seem to be limited to very old gcc versions or to combinations of flags like `-fvisibility=hidden` we don't use).  And calling virtual functions from constructors is definitely defined and permitted (though it's frowned upon, as the behavior of "dispatch goes to the currently-constructed-class definition rather than to the eventually-constructed-class definition" is pretty non-intuitive.

But I'm also not 100% sure this *isn't* undefined behavior.  You should be able to call `Hex::n_sides()` from the `Hex` constructor body, but is it legal to call it from the initializer list, especially at a place where even `DofObject::DofObject()` hasn't been called yet?  I would have guessed (I suppose I did guess) that the answer is still "yes" because qualifying with `Hex::` should avoid virtual dispatch.  But maybe it's still disallowed because the compiler hasn't even officially had time to set up the sort of trickery that allows `this` to be type converted with automatic offsets in cases of multiple inheritance?

Either way I liked the new code better than the old so I didn't worry about it too much at the time.  Now you've got my curiosity itching through.

jwpeterson: OK, thanks for the explanation. I agree the new code is cleaner anyway, I had no idea we were duplicating `num_sides` etc. in all the Hex-derived classes, definitely no reason to do that.

roystgnr: Whoa, here we are: https://stackoverflow.com/a/63391516

"It is not safe to call any member function (virtual or not virtual) before all base have been initialized."

You can call member functions from the rest of the initializer list entries afterward, but not to evaluate arguments for the base class initializer(s).

jwpeterson: OK, sounds like a legitimate find by the sanitizer then.

PR #3956, opened 2024-09-19
ASCII .stl support

roystgnr:
@maxnezdyur reported we've had users ask for STL support, and I had some time today while waiting for long tests to finish running, so here's a first crack at it.

It's only ASCII and only Tri3 support currently, and I'm done for the week, but adding binary support and Tri6/7 subdivision-write support shouldn't be much extra work.

### DIFF
--- a/src/mesh/stl_io.C
+++ b/src/mesh/stl_io.C
@@ -1,0 +1,358 @@
+// The libMesh Finite Element Library.
+// Copyright (C) 2002-2024 Benjamin S. Kirk, John W. Peterson, Roy H. Stogner
+
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+// Local includes
+#include "libmesh/stl_io.h"
+
+#include "libmesh/distributed_mesh.h"
+#include "libmesh/enum_elem_type.h"
+#include "libmesh/face_tri3.h"
+#include "libmesh/libmesh_config.h"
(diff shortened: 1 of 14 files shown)


src/mesh/stl_io.C
```diff
@@ -0,0 +1,358 @@
...
+
+
+void STLIO::read (const std::string & filename)
+{
+  LOG_SCOPE("read()", "STLIO");
+
+#ifndef LIBMESH_HAVE_CXX11_REGEX
+  libmesh_not_implemented();  // What is your compiler?!?  Email us!
+  libmesh_ignore(filename);
+#else
```
jwpeterson, line 154: Should we make C++11 regex support officially required now? Looking at configure.ac, I see:
```
# --------------------------------------------------------------
# Test for optional modern C++ features, which libMesh offers shims
# for and/or which libMesh applications may want to selectively use.
# --------------------------------------------------------------
LIBMESH_TEST_CXX14_MAKE_UNIQUE
LIBMESH_TEST_CXX11_REGEX
```
and we obviously use `std::make_unique` freely everywhere now, so I think these could probably both be moved to the `ACSM_TEST_CXX_ALL` test, which is what sets the `have_cxx_all` flag. A topic for another PR though, either way.

src/mesh/stl_io.C
```diff
@@ -0,0 +1,358 @@
...
+          libmesh_error_msg_if
+            (c != '\n' && c != '\r' &&
+             (c < ' ' || c > '~'),
+             "Found non-ASCII character " << int(c) << " on line " <<
+             line_num << "\nSTLIO does not yet support binary files.");
+        }
+
+      if (std::regex_search(line, sm, start_regex))
+        {
+          if (std::regex_search(line, sm, start_with_name_regex))
```
jwpeterson, line 253: Our current `std::regex_search` configure test does not test the version of that [function](https://en.cppreference.com/w/cpp/regex/regex_search) taking a `std::match_results<T>`, so if we do decide to make `std::regex` required, we should probably also beef up the test a little, so it at least tests all the functionality we use.

jwpeterson (approved):

src/mesh/stl_io.C
roystgnr: Yeah, that sounds good to me.

[cppreference claims](https://en.cppreference.com/w/cpp/compiler_support/11) that Embarcadero C++ Builder and Sun/Oracle C++ are the only libraries that didn't add regex support ... but [later claims](

src/mesh/stl_io.C
roystgnr: Oh, yikes, definitely.  I'm already using `match_results` in our new `SparseMatrix::read_matlab` too.

PR #3963, opened 2024-09-26
Output EDGE to exodus as BAR

roystgnr:
"EDGE" isn't actually in the ExodusII standard, as far as I can tell, despite Paraview and even *exodiff* seemingly supporting it.  But apparently Python meshio is stricter, so we should be more standards-conforming here.

Thanks to Adriaan Riet (@ke7kto) for catching this.


roystgnr:
Well, that's fun Civet reporting:
```
Failed to save output:
A string literal cannot contain NUL (0x00) characters.
```
I'll see if I can reproduce.

PR #3964, opened 2024-10-02
MeshBase::elem_orders()

roystgnr:
I want to add some testing to this, and fix the devel->master merges, before we merge this, but I'm worried that that NodeElem change might break something somewhere so let's get this into CI now.

### DIFF
--- a/include/enums/enum_order.h
+++ b/include/enums/enum_order.h
@@ -83,6 +83,7 @@ enum Order : int {
             FORTYFIRST   = 41,
             FORTYSECOND  = 42,
             FORTYTHIRD   = 43,
+            MAXIMUM = 255,
             // Invalid
             INVALID_ORDER};
 
--- a/src/mesh/mesh_base.C
+++ b/src/mesh/mesh_base.C
@@ -31,6 +31,7 @@
 #include "libmesh/mesh_tools.h"
 #include "libmesh/parallel.h"
 #include "libmesh/parallel_algebra.h"
+#include "libmesh/parallel_fe_type.h"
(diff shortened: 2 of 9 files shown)


include/enums/enum_order.h
```diff
@@ -83,6 +83,7 @@ enum Order : int {
             FORTYFIRST   = 41,
             FORTYSECOND  = 42,
             FORTYTHIRD   = 43,
+            MAXIMUM = 255,
```
jwpeterson, line 86: "255th degree should be enough for anyone." - Roy Stogner

src/mesh/mesh_base.C
```diff
@@ -2148,6 +2182,11 @@ MeshBase::copy_constraint_rows(const SparseMatrix<T> & constraint_operator)
 
       Elem * added_elem = this->add_elem(std::move(elem));
       this->_elem_dims.insert(0);
+      this->_elem_default_orders.insert(added_elem->default_order());
+      this->_max_nodal_order =
+        static_cast<Order>
+          (std::min(static_cast<int>(this->_max_nodal_order),
+                    static_cast<int>(added_elem->max_nodal_order())));
```
jwpeterson, line 2189: Since `_max_nodal_order` actually the *minimum* max nodal order, should we maybe name that parameter differently? I know that this is explained in the parameter comments, but I think it is an important distinction that could be easy to miss.

jwpeterson (approved):
Looks good to me, other than if you wanted to name `_max_nodal_order` a bit more specifically e.g. `_smallest_max_nodal_order` or something along those lines.

PR #3967, opened 2024-10-08
SimplexRefiner for mesh refinement via edge bisection

roystgnr:
Playing around with STL files revealed how many of them are good at capturing a geometry with the fewest triangles possible, which is good for file size but not good for constructing a good tetrahedralization.  This lets us fix that, and is the first step toward doing similar things on tets directly.

### DIFF
--- a/include/mesh/simplex_refiner.h
+++ b/include/mesh/simplex_refiner.h
@@ -1,0 +1,151 @@
+// The libMesh Finite Element Library.
+// Copyright (C) 2002-2024 Benjamin S. Kirk, John W. Peterson, Roy H. Stogner
+
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+
+#ifndef LIBMESH_SIMPLEX_REFINER_H
+#define LIBMESH_SIMPLEX_REFINER_H
+
+#include "libmesh/libmesh_config.h"
+
+// Local Includes
+#include "libmesh/elem.h"
+#include "libmesh/function_base.h"
+
+#include <map>
+#include <memory>
+
+namespace libMesh
+{
+
+// Forward Declarations
+class UnstructuredMesh;
+class Node;
+
+/**
+ * A C++ class to refine a simplicial mesh via splitting edges that
+ * exceed a given metric.
+ *
+ * \author Roy H. Stogner
+ * \date 2024
+ */
+class SimplexRefiner
+{
+public:
+  /**
+   * The constructor.  A reference to the mesh containing the elements
+   * which are to be split by edge subdivision must be provided.
+   *
+   * At the time of refining this mesh should be conforming.
+   */
+  explicit
+  SimplexRefiner(UnstructuredMesh & mesh);
+
+  /**
+   * Finds elements which exceed the requested metric and refines them
+   * via subdivision into new simplices as necessary.
+   *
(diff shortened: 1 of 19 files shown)


include/mesh/simplex_refiner.h
```diff
@@ -0,0 +1,198 @@
...
+  void fill_refinement_datum(std::pair<Node *, Node *> vertices,
+                             refinement_datum & vec);
+
+  /**
+   * Keep track of elements added within each original element, so we
+   * can answer queries about them from other processors.  If the
+   * element originally with id i is split, its replacements e1 and e2
+   * will be in added_elements[i][ei.vertex_average()]
+   */
+  std::unordered_map<dof_id_type, std::unordered_map<Point, Elem *>> added_elements;
```
jwpeterson, line 181: Huh, I didn't realize we had a `std::hash<Point>` specialization. Is this how you want to handle it whenever we need to make an `unordered_map` with a non-trivial Key type? In #3975, I used the "class with operator()" approach, and this approach is also used in e.g. the `ElemHashUtils` class. Basically I'm just curious what is considered the best practice for user-defined hash functions... should they all just be put in namespace `std`?

src/mesh/mesh_generation.C
```diff
@@ -2752,5 +2752,44 @@ void MeshTools::Generation::build_delaunay_square(UnstructuredMesh & mesh,
 #endif // LIBMESH_HAVE_TRIANGLE && LIBMESH_DIM > 1
 
 
+void MeshTools::Generation::surface_octahedron
+  (UnstructuredMesh & mesh,
+   Real xmin, Real xmax,
+   Real ymin, Real ymax,
+   Real zmin, Real zmax,
+   bool flip_tris)
```
jwpeterson, line 2760: `MeshTools::Generation::surface_octahedron()` seems a little bit specific? I could see having a `MeshTools` implementation for triangulating a regular polygon with n sides... but this looks more like unit test code to me.

src/mesh/replicated_mesh.C
```diff
@@ -425,6 +425,10 @@ void ReplicatedMesh::delete_elem(Elem * e)
 void ReplicatedMesh::renumber_elem(const dof_id_type old_id,
                                    const dof_id_type new_id)
 {
+  // This could be a no-op
+  if (old_id == new_id)
+    return;
+
```
jwpeterson, line 431: This (and the same update in distributed_mesh.C) seem like good optimizations to have independent of the rest of the PR, I wouldn't be opposed to merging them in a separate branch but also would understand if you don't feel like doing that. I assume they are already in a separate commit here, so that's probably fine as well.

src/mesh/simplex_refiner.C
```diff
@@ -0,0 +1,598 @@
...
+  _mesh(mesh),
+  _desired_volume(0.1)
+{}
+
+
+
+bool SimplexRefiner::refine_elements()
+{
+  if (_mesh.is_prepared())
+    _mesh.prepare_for_use();
```
jwpeterson, line 55: Are you missing a `!` here?

jwpeterson (approved):

include/mesh/simplex_refiner.h
roystgnr: In #3975 you did the only legal thing, I think - our key is a `vector<dof_id_type>`, and `vector` isn't a user class, and `dof_id_type` is a user typedef but not a user class, so it's not legal to put our own specialization around it into namespace `std`.  With `Point`, because we created an actual class, doing our own specializations for it is specifically allowed ... though I'm not sure if it's considered best practice.  IMHO the main benefit of it is only that it makes generic algorithms easier if you don't have to figure out how to use a custom hash function template argument for some cases but not others.

src/mesh/simplex_refiner.C
roystgnr: I assume `!` is a euphemism for "brain cell", in which case the answer is yes.

If you just meant a literal `!` the answer is also yes.

src/mesh/mesh_generation.C
jwpeterson: I'm actually good with having the ability to mesh 3D *octahedrons* in MeshTools::Generation. For some reason, the words "surface" and "octahedron" combined in my brain to equal "octagon".

roystgnr:
Rebased on devel to fix conflicts in autotools bits; the only new thing here is the `!` bug fix.

src/mesh/simplex_refiner.C
jwpeterson: @permcody Why are you always preparing?

PR #3970, opened 2024-10-14
Make AbaqusIO extendable

### LINKED ISSUE #3969: Make AbaqusIO more extendable
I would like to derive from `AbaqusIO` to add support for some exotic features. For that I need to change private to protected access and move two helpers from the anonymous namespace into the class. I'm also adding a virtual callback to parse sections not recognized by the main parser.

dschwen:
Closes #3969

### DIFF
--- a/include/mesh/abaqus_io.h
+++ b/include/mesh/abaqus_io.h
@@ -66,7 +66,7 @@ public:
    */
   bool build_sidesets_from_nodesets;
 
-private:
+protected:
   /**
    * The type of data structure used to store Node and Elemset IDs.
    */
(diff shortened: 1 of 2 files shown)


include/mesh/abaqus_io.h
```diff
@@ -219,6 +219,26 @@ class AbaqusIO : public MeshInput<MeshBase>
...
+   * Removes all whitespace characters from "line". Simpler than trying
+   * to remember the erase-remove-if idiom.
+   */
+  void strip_ws(std::string &line);
+
+  /**
+   * Override this callback to implement support for more sections in
+   * derived classes.
+   */
+  virtual void extendedParsing(const std::string & /*header*/) {}
```
jwpeterson, line 241: We don't use camel case in libMesh

jwpeterson (commented):
Maybe I don't understand the full context here, but can you not just update the existing `AbaqusIO` class with the support needed? Is this just a way of avoiding the need to do a libmesh update in MOOSE in order to bring in new libmesh changes?

dschwen:
> Maybe I don't understand the full context here, but can you not just update the existing `AbaqusIO` class with the support needed? Is this just a way of avoiding the need to do a libmesh update in MOOSE in order to bring in new libmesh changes?

Yeah, there is more context, namely the implementation of Abaqus user defined elements, which do not quite fit the mold in libmesh.

PR #3971, opened 2024-10-15
Normal normals

roystgnr:
A user suggested that they might have seen normal vectors inverted on a flipped Tri element.  I couldn't reproduce that, but I *did* see normal vectors inverted on a flipped Edge element.  This PR fixes that, and adds more unit tests to make sure such a gross error shouldn't recur.  More subtle potential errors are going to need more subtle tests at some point.

### DIFF
--- a/src/fe/fe_boundary.C
+++ b/src/fe/fe_boundary.C
@@ -664,7 +664,7 @@ void FEMap::compute_face_map(int dim, const std::vector<Real> & qw,
     {
     case 1:
       {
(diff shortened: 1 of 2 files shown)


jwpeterson (approved):

roystgnr:
Rebased on master to see if the MOOSE fixes have made it in yet.

roystgnr:
And again

roystgnr:
Yeah, I'd say so.  Our only problem with it was the downstream app code that had been relying on the broken behavior.

PR #3973, opened 2024-10-21
Refactor with LibmeshPetscCall

nmnobre:
Hi,

Would this be of interest? I can probably do the rest of the code in a morning if it would.

Cheers,
-N

### DIFF
--- a/include/numerics/petsc_solver_exception.h
+++ b/include/numerics/petsc_solver_exception.h
@@ -102,7 +102,7 @@ public:
 
 // Two argument version of the function above where you pass in the comm
 // instead of relying on it being available from the "this" pointer.
-#define LIBMESH_CHKERR2(comm, ierr) CHKERRABORT(comm.get(), ierr);
+#define LIBMESH_CHKERR2(comm, ierr) CHKERRABORT(comm, ierr);
 
 // Alternative macro to be used in functions returning a datatype convertible
 // to PetscErrorCode.
@@ -118,11 +118,11 @@ public:
   inline                                                                \
   void Function ## BeginEnd(const Parallel::Communicator & comm, const Args&... args) \
   {                                                                     \
-    PetscErrorCode ierr = LIBMESH_PETSC_SUCCESS;               \
+    PetscErrorCode ierr = LIBMESH_PETSC_SUCCESS;                        \
     ierr = Function ## Begin(args...);                                  \
-    LIBMESH_CHKERR2(comm, ierr);                                        \
+    LIBMESH_CHKERR2(comm.get(), ierr);                                  \
     ierr = Function ## End(args...);                                    \
-    LIBMESH_CHKERR2(comm, ierr);                                        \
+    LIBMESH_CHKERR2(comm.get(), ierr);                                  \
   }
 
 PETSC_BEGIN_END(VecScatter) // VecScatterBeginEnd
@@ -139,6 +139,14 @@ PETSC_BEGIN_END(VecGhostUpdate) // VecGhostUpdateBeginEnd
   } while (0)
 
 #define LibmeshPetscCall2(comm, ...)                                                               \
+  do                                                                                               \
+  {                                                                                                \
+    PetscErrorCode libmesh_petsc_call_ierr;                                                        \
+    libmesh_petsc_call_ierr = __VA_ARGS__;                                                         \
+    LIBMESH_CHKERR2(comm.get(), libmesh_petsc_call_ierr);                                          \
+  } while (0)
+
+#define LibmeshPetscCallA(comm, ...)                                                               \
   do                                                                                               \
   {                                                                                                \
     PetscErrorCode libmesh_petsc_call_ierr;                                                        \
--- a/include/numerics/petsc_solver_exception.h
+++ b/include/numerics/petsc_solver_exception.h
@@ -89,6 +89,11 @@ public:
     LIBMESH_CHKERR(ierr);                       \
   } while (0)
 
+// Macro which we call in functions returning a datatype convertible
+// to PetscErrorCode after every PETSc function that returns an error code.
+#define LIBMESH_CHKERRQ(ierr)                   \
+  LIBMESH_CHKERR(ierr);
+
 #else
 
 // If we don't have exceptions enabled, just fall back on calling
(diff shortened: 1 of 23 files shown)


jwpeterson:
I assume that it was @lindsayad's intention to eventually replace all of the PETSc function call/error checking code with `LibmeshPetscCall`, so this looks like a good change to me, but I'll let him comment whether there is some reason to hold off on updating the remaining ones.

@roystgnr this PR was marked as a Draft on GitHub but the CI testing still ran. Is that the intended behavior in CIVET for Draft PRs? I seem to remember that marking a PR as Draft would prevent the tests from running before the user was ready to do so.

nmnobre:
> I assume that it was @lindsayad's intention to eventually replace all of the PETSc function call/error checking code with `LibmeshPetscCall`, so this looks like a good change to me, but I'll let him comment whether there is some reason to hold off on updating the remaining ones.

I think I've completed all replacements for the macros we already have.
Unfortunately, I think we should have named `LibmeshPetscCall(2)`, `LibmeshPetscCallAbort(2)` instead, so that `LibmeshPetscCall` would be free for idioms using `CHKERRQ`. As a workaround, I can create `LibmeshPetscCall3` for the latter case and redefine `LibmeshPetscCall2` to get an MPI communicator, so I can cover more cases.
This way, when we think we're ready to drop support for PETSc < 3.17.0, we could simply:
- redefine `LibmeshPetscCall` to wrap `PetscCallAbort`;
- drop `LibmeshPetscCall2` and call `PetscCallAbort` directly;
- drop `LibmeshPetscCall3` and call `PetscCall` directly.

> @roystgnr this PR was marked as a Draft on GitHub but the CI testing still ran. Is that the intended behavior in CIVET for Draft PRs? I seem to remember that marking a PR as Draft would prevent the tests from running before the user was ready to do so.

FWIW, I actually find this very useful. :)

roystgnr:
Running CI for Draft PRs is definitely the behavior I'd intend.  I write ~~too much~~ occasional code that I want to run through the whole CI gauntlet even when I know there are other things that still need to be improved or even fixed before merging.

roystgnr:
I'd rather not use `PetscCall` ever, personally.  `LIBMESH_CHKERR` doesn't avoid using `CHKERR` not because the latter is a recent invention, but because throwing an exception that can either be caught or can trigger a terminate handler is strictly more useful than directly triggering a terminate handler.

nmnobre:
> I'd rather not use `PetscCall` ever, personally. `LIBMESH_CHKERR` doesn't avoid using `CHKERR` not because the latter is a recent invention, but because throwing an exception that can either be caught or can trigger a terminate handler is strictly more useful than directly triggering a terminate handler.

Yes, of course. Brain fart -.-, I was looking through the non-`LIBMESH_ENABLE_EXCEPTIONS` code path for a few hours to make sure I was calling the right macro, and completely forgot about the default code path. I think tomorrow I'll create `LibmeshPetscCallQ` which I think is a bit more descriptive than what I suggested earlier with `LibmeshPetscCall3`. (I think you guys might've named it `LibmeshPetscCall2` because it takes 2 args).

nmnobre:
Done. All yours @roystgnr.

include/base/libmesh.h
```diff
@@ -119,7 +119,7 @@ class LibMeshInit
    * handling.  Finalizes any of Slepc, Petsc, and MPI which were
    * initialized by LibMeshInit.
    */
-  virtual ~LibMeshInit();
+  virtual ~LibMeshInit() noexcept(false);
```
jwpeterson, line 122: I'd rather not be throwing exceptions from the `LibMeshInit` destructor (or any destructor for that matter) since it's generally considered bad practice (destructors are called during stack unwinding, throwing another exception during stack unwinding immediately terminates the program). We use `libmesh_exceptionless_assert()` and `CHKERRABORT` in the `LibMeshInit` destructor for this reason.

nmnobre: Agreed.

nmnobre: I chose to just revert to using `CHKERRABORT` instead of calling `PetscCallAbort` for compatibility with PETSc < 3.17.0.

lindsayad:
You can use `[WIP]` or `WIP:` in the PR title for disabling CIVET testing

include/numerics/petsc_solver_exception.h
```diff
@@ -130,13 +139,29 @@ PETSC_BEGIN_END(VecGhostUpdate) // VecGhostUpdateBeginEnd
...
+
+#define LibmeshPetscCallA(comm, ...)                                                               \
   do                                                                                               \
   {                                                                                                \
     PetscErrorCode libmesh_petsc_call_ierr;                                                        \
     libmesh_petsc_call_ierr = __VA_ARGS__;                                                         \
     LIBMESH_CHKERR2(comm, libmesh_petsc_call_ierr);                                                \
   } while (0)
 
+#define LibmeshPetscCallQ(...)                                                                     \
```
lindsayad, line 157: With all these macros now, please add comments about when developers should use which

lindsayad (changes requested):

lindsayad (approved):

include/numerics/petsc_solver_exception.h
```diff
@@ -89,6 +89,10 @@ class PetscSolverException : public SolverException
     LIBMESH_CHKERRQ(ierr);                      \
   } while (0)
 
+// Remove me: for backward compatibility with MOOSE only
+#define LIBMESH_CHKERR(ierr) LIBMESH_CHKERRQ(ierr)
+#define LIBMESH_CHKERR2(comm, ierr) LIBMESH_CHKERRA(comm, ierr)
```
lindsayad, line 94: It's confusing that the one vs. two argument macros are equivalent to `PetscCall` and `PetscCallAbort` respectively. Is this only because MOOSE is "being dumb"?

nmnobre: Sorry, I don't think I understood the question, `PetscCall` and `PetscCallAbort` are also one and two argument macros?

lindsayad: `PetscCallAbort` will abort immediately if the error code is non-zero. This does not happen with `PetscCall`

nmnobre: Agreed. Point being? (sorry, it's late over here...)

lindsayad: I would not expect that `LIBMESH_CHKERR` and `LIBMESH_CHKERR2` have fundamentally different behavior when the only difference is that one takes a comm and the other does not

nmnobre: Right, thanks, apologies. I'm with you now.

Yes, it's true that before, *if* `LIBMESH_ENABLE_EXCEPTIONS` wasn't defined, they'd both wind up calling `CHKERRABORT`, and that behaviour is now reserved to `LIBMESH_CHKERRA`. Note, however, that the fallback is only defined for the case where `LIBMESH_ENABLE_EXCEPTIONS` is indeed defined, for which the behaviour was already the same before.

lindsayad: Got ya. Yea I thought you didn't likely *add* any silliness that wasn't there before. Still would be nice if we as a collective make this more consistent. Probably can do this after updating MOOSE

PR #3975, opened 2024-10-22
Add NonManifoldGhostingFunctor and associated SidesToElemMap classes

jwpeterson:
We often have structural solves involving shells where more than two shell elements meet at a single Edge (see screenshots of some cases where 5 shells meet at a single Edge). In such cases, we don't have complete neighbor information, but it is still possible to correctly ghost the required DOFs when the different neighbors are on different processors. This PR adds a new `GhostingFunctor` which enables this type of ghosting, as well as several unit tests.

non_manifold_junction3
non_manifold_junction2
non_manifold_junction1
non_manifold_junction0


jwpeterson:
The warning that I'm seeing from the Min clang testing makes no sense
```
mpicxx -std=gnu++17 -DHAVE_CONFIG_H -I. -I../../tests -I../include  -DNDEBUG  -I/usr/include -pthread -I/opt/petsc/include -I/opt/petsc//include -I/usr/include/tirpc -I../include -I../../contrib/metaphysicl/src/numerics/include -I../../contrib/metaphysicl/src/core/include -I../../contrib/metaphysicl/src/utilities/include -I../../contrib/nanoflann/nanoflann/include -I../../contrib/fparser -I../../contrib/libHilbert/include -I../../contrib/nemesis/v8.11/nemesis -I../../contrib/exodusii/v8.11/exodus/include -I../../contrib/exodusii/v8.11/exodus/sierra -I../../contrib/netcdf/netcdf-c/include -I../contrib/netcdf/netcdf-c/include -I../../contrib/eigen/eigen -I../../contrib/gmv -I../../contrib/qhull/qhull/src -I../../contrib/qhull/qhull/src/libqhullcpp -I../../contrib/triangle -I../../contrib/tetgen -I../../contrib/netgen/ -I../contrib/netgen/build/ -I../contrib/poly2tri/modified   -I../../contrib/gzstream -I../../contrib/sfcurves -I../../contrib/laspack -I../../contrib/boost/include -I../contrib/timpi/src/utilities/include -I../../contrib/timpi/src/utilities/include -I../../contrib/timpi/src/parallel/include -I../../contrib/timpi/src/algorithms/include   -DLIBMESH_IS_UNIT_TESTING  -O2 -felide-constructors -Qunused-arguments -Wunused-parameter -Wunused -ftrapping-math    -fopenmp -Werror -Wall -Wextra -Wcast-align -Wdisabled-optimization -Wformat=2 -Wformat-nonliteral -Wformat-security -Wformat-y2k -Winvalid-pch -Wmissing-field-initializers -Wmissing-include-dirs -Wpacked -Wstack-protector -Wtrigraphs -Wunreachable-code -Wunused-label -Wunused-parameter -Wunused-value -Wvariadic-macros -Wvolatile-register-var -Wwrite-strings  -MT fe/unit_tests_opt-fe_lagrange_test.o -MD -MP -MF fe/.deps/unit_tests_opt-fe_lagrange_test.Tpo -c -o fe/unit_tests_opt-fe_lagrange_test.o `test -f 'fe/fe_lagrange_test.C' || echo '../../tests/'`fe/fe_lagrange_test.C
../../tests/base/nonmanifold_coupling_test.C:69:13: error: loop will run at most once (loop increment never executed) [-Werror,-Wunreachable-code-loop-increment]
            for (auto s : elem->side_index_range())
            ^~~
```
It's like it doesn't recognize that it's a range-based for-loop for some reason.

We use this construction in many places in the library, though, so no idea why it only `-Werrors` on this one.

jwpeterson:
It's the same `-Werror` on Test clang as well, what the heck?

include/mesh/sides_to_elem_map.h
```diff
@@ -0,0 +1,145 @@
...
+
+  /**
+   * Return a const reference to the list of Elems connected to "side"
+   * of "elem". The returned list of Elems will also include "elem"
+   * itself. Throws an error if the requested (elem, side) combination
+   * is not found in the map (this should not happen because every
+   * side in the map should be attached to at least one Elem).
+   */
+  const std::vector<const Elem *> &
+  get_connected_elems(const Elem * elem, unsigned int side) const;
```
roystgnr, line 77: I'm not thrilled about returning a reference from a const function.  That forces us to either make sure that those vectors will always part of our internal data structure (as opposed to any means of finding that set of elements on the fly) or to do nasty things with `mutable` if we change that.

But I certainly don't want to return the vector by value either and force a copy with heap allocation on every call.

Could we perhaps return a pair of begin/end iterators, of a type that for now is just a `typedef` of `vector::const_iterator`?  That'll still be ABI-breaking to change but it'll be easy to keep API compatible.

src/ghosting/non_manifold_coupling.C
```diff
@@ -0,0 +1,101 @@
...
+
+NonManifoldGhostingFunctor::
+NonManifoldGhostingFunctor(const MeshBase & mesh) :
+  GhostingFunctor(mesh),
+  _stem(MeshTools::SidesToElemMap::build(mesh))
+{
+  // TODO: find out if we should build the cache here in the constructor
+  // or if there will always be at least one mesh_reinit() call before the
+  // NonManifoldGhostingFunctor is called. Roy's initial suggestion is to
+  // build the cache in both places.
```
roystgnr, line 33: I checked; we do a `mesh_reinit()` call in `prepare_for_use()`, before we ever try to use ghosting functors for anything on the mesh or its systems.  If the ghosting functor is attached before the mesh is initialized, which it very much should be because otherwise we can end up deleting remote elements that can't be recovered without serializing, then skipping the map build here should be safe.

On the other hand, if someone's *really* doing things properly then they should be attaching the ghosting functor before ever putting elements in the mesh (in which case the build should be free to leave in), and if someone's doing things a bit dubiously (cough, MOOSE) then better safe than sorry.

roystgnr (changes requested):
Only one huge change request:

It looks like we're not supporting meshes with hanging nodes here?  E.g. if we have a refined element with a coarser neighbor, I don't think the side to elem map will catch that.

If I'm right, that's fine for a first version of this, but we need to at least be screaming if we see that situation (maybe check for parents that have entries in the map where we don't?) rather than just reporting empty connections when we shouldn't.

jwpeterson:
@roystgnr I believe I have addressed all your review comments here.

include/mesh/sides_to_elem_map.h
roystgnr: Yeah, it's not that it forces the function to be conceptually non-const now, it's that if we ever change the data structure it'd force the function to do evil mutable or const_cast things later.  Thanks for the change.

roystgnr (approved):

PR #3874, opened 2024-06-11
Refactor code for ND1 and RT shape functions

nmnobre:
Pending benchmarking...

PR #3977, opened 2024-10-23
New utility functions and slight refactoring

nmnobre:
This is in preparation for the new hypre preconditioner(s) for ND1 (and RT) systems coming soon.
It's also a dependence for the shape functions refactoring in #3874 which now uses the new functions in the `Elem` class.

Cheers,
-Nuno

### DIFF
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -1065,6 +1065,21 @@ public:
    */
   virtual bool close_to_point(const Point & p, Real tol) const;
 
+  /**
+   * \returns \p true if edge \p i is positively oriented. An edge is
+   * positively oriented iff its first vertex (i.e. zeroth node) is
+   * lexicographically greater than its second vertex (i.e. first node).
+   */
+  bool edge_orientation(const unsigned int i) const;
+
+  /**
+   * \returns \p true if face \p i is positively oriented. A face with N
+   * vertices is positively oriented iff its 3 lexicographically greatest
+   * vertices (i.e. 3 lexicographically greatest nodes amongst the N leading
+   * nodes) are an odd permutation relative to their lexicographic ordering.
+   */
+  bool face_orientation(const unsigned int i) const;
+
 private:
   /**
    * Shared private implementation used by the contains_point()
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -1070,7 +1070,7 @@ public:
    * positively oriented iff its first vertex (i.e. zeroth node) is
    * lexicographically greater than its second vertex (i.e. first node).
    */
-  bool edge_orientation(const unsigned int i) const;
+  bool positive_edge_orientation(const unsigned int i) const;
 
   /**
    * \returns \p true if face \p i is positively oriented. A face with N
@@ -1078,7 +1078,7 @@ public:
    * vertices (i.e. 3 lexicographically greatest nodes amongst the N leading
    * nodes) are an odd permutation relative to their lexicographic ordering.
    */
-  bool face_orientation(const unsigned int i) const;
+  bool positive_face_orientation(const unsigned int i) const;
 
 private:
   /**
--- a/src/geom/elem.C
+++ b/src/geom/elem.C
@@ -3404,4 +3404,37 @@ Elem::is_internal(const unsigned int i) const
   }
 }
 
+bool
+Elem::edge_orientation(const unsigned int i) const
+{
+  libmesh_assert_less (i, this->n_edges());
+
+  return this->point(this->local_edge_node(i, 0)) >
+         this->point(this->local_edge_node(i, 1));
+}
+
+bool
+Elem::face_orientation(const unsigned int i) const
+{
+  libmesh_assert_less (i, this->n_faces());
+
+  const unsigned int N = Elem::type_to_n_sides_map[this->side_type(i)];
+
+  const std::vector<unsigned int> nodes = this->nodes_on_side(i);
+  std::vector<Point> vertices(N);
+  std::transform(nodes.begin(), nodes.begin() + N, vertices.begin(),
+                 [&](unsigned int n) { return this->point(n); });
+
+  for (unsigned int j = 0; j < N - 3; j++)
+    std::rotate(vertices.begin() + j,
+                std::min_element(vertices.begin() + j, vertices.end()),
+                vertices.end());
+
+  unsigned int cnt = 0;
+  for (unsigned int j = N - 3; j < N; j++)
+    for (unsigned int k = j + 1; k < N; k++)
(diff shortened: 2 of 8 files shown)


roystgnr (approved):

src/systems/equation_systems.C
```diff
@@ -79,25 +79,11 @@ void EquationSystems::clear ()
...
-  for (auto & elem : _mesh.element_ptr_range())
-    elem->set_n_systems(n_sys);
-
-  for (auto i : make_range(this->n_systems()))
-    this->get_system(i).init();
-
-#ifdef LIBMESH_ENABLE_AMR
-  MeshRefinement mesh_refine(_mesh);
-  mesh_refine.clean_refinement_flags();
-#endif
```
roystgnr, line 100: Wait a minute.  Where did the refinement flag cleanup go?

roystgnr (changes requested):

src/geom/elem.C
```diff
@@ -3404,4 +3404,37 @@ Elem::is_internal(const unsigned int i) const
...
+
+bool
+Elem::face_orientation(const unsigned int i) const
+{
+  libmesh_assert_less (i, this->n_faces());
+
+  const unsigned int N = Elem::type_to_n_sides_map[this->side_type(i)];
+
+  const std::vector<unsigned int> nodes = this->nodes_on_side(i);
+  std::vector<Point> vertices(N);
```
jwpeterson, line 3424: This sequence of lines confused me because it looks like `N` should be a number of sides, since it comes from the `type_to_n_sides_map`, but you are getting a number of vertices, based on the fact that "the number of sides of a side" is equal to the number of vertices on that side for 3D elements. My suggestions are to name the variable more descriptively, e.g. `n_side_vertices`, and even better, put a comment explaining what you are doing.

src/geom/elem.C
```diff
@@ -3404,4 +3404,37 @@ Elem::is_internal(const unsigned int i) const
...
+  libmesh_assert_less (i, this->n_edges());
+
+  return this->point(this->local_edge_node(i, 0)) >
+         this->point(this->local_edge_node(i, 1));
+}
+
+bool
+Elem::face_orientation(const unsigned int i) const
+{
+  libmesh_assert_less (i, this->n_faces());
```
jwpeterson, line 3419: I think we should assert that the Elem's dimension == 3 here, and I'm actually in favor of making it an opt-mode `libmesh_error_msg_if()` check instead of just an assert. If this function is called on a 2D element (in opt mode), it's possible for N == 2, at which point computing N - 3 will wrap to a huge number in the loops below, likely leading to a segfault.

include/geom/elem.h
```diff
@@ -1065,6 +1065,21 @@ class Elem : public ReferenceCountedObject<Elem>,
    */
   virtual bool close_to_point(const Point & p, Real tol) const;
 
+  /**
+   * \returns \p true if edge \p i is positively oriented. An edge is
+   * positively oriented iff its first vertex (i.e. zeroth node) is
+   * lexicographically greater than its second vertex (i.e. first node).
+   */
+  bool edge_orientation(const unsigned int i) const;
```
jwpeterson, line 1073: I'd prefer to name these functions `positive_edge_orientation()` and `positive_face_orientation()`, respectively, so that it's more clear what the `true`/`false` return value is referring to.

src/systems/equation_systems.C
nmnobre: Nowhere.

https://github.com/farscape-project/libmesh/blob/833c85b501f242a90dd21397402948f7d881fcf9/src/systems/equation_systems.C#L118-L121

jwpeterson (changes requested):
It's not obvious to me that calling `EquationSystems::reinit_mesh()` from `EquationSystems::init()` does the same thing that this function did before, but it did pass all the CI testing... 

I just think it's just conceptually confusing that calling `EquationSystems::init()` no longer means that we call `System::init()` for each one? I agree that it would be good to get rid of some of the redundancy here though.

src/geom/elem.C
```diff
@@ -3404,4 +3404,37 @@ Elem::is_internal(const unsigned int i) const
...
+
+  for (unsigned int j = 0; j < N - 3; j++)
+    std::rotate(vertices.begin() + j,
+                std::min_element(vertices.begin() + j, vertices.end()),
+                vertices.end());
+
+  unsigned int cnt = 0;
+  for (unsigned int j = N - 3; j < N; j++)
+    for (unsigned int k = j + 1; k < N; k++)
+      if (vertices[j] > vertices[k]) cnt++;
```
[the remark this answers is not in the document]
nmnobre: Done.

include/geom/elem.h
nmnobre: Done.

src/geom/elem.C
nmnobre: > I think we should assert that the Elem's dimension == 3 here

That's implicit since `n_faces` is zero if that's not the case.

> I'm actually in favor of making it an opt-mode `libmesh_error_msg_if()` check instead of just an assert.

Help me here, how do you decide? I can similarly segfault on a call to `local_edge_node` if I ask for a non-existent edge.

src/systems/equation_systems.C
```diff
@@ -79,25 +79,11 @@ void EquationSystems::clear ()
 
 void EquationSystems::init ()
 {
-  const unsigned int n_sys = this->n_systems();
-
-  libmesh_assert_not_equal_to (n_sys, 0);
+  libmesh_assert(std::all_of(make_range(this->n_systems()).begin(),
+                             make_range(this->n_systems()).end(),
+                             [&](unsigned i) { return !this->get_system(i).is_initialized(); }));
```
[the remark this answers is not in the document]
nmnobre: Done. This renders 833c85b unnecessary (strictly speaking), should I drop it?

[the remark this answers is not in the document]
jwpeterson: I think it's good to leave those iterator typedefs in... we should have always had those.

src/geom/elem.C
jwpeterson: It's true that we should be consistent (if we assert in other places, we should assert here too) and performant (so, perfer debug-mode asserts to opt-mode checks) so I'm OK keeping it as an assert. 

I thought it would clearer to someone seeing an assert from this function if it said "You can only call Elem::positive_face_orientation() on 3D elements." instead of something about `i` not being less than the number of faces, but I don't feel to strongly about it.

jwpeterson (approved):
Looks good to me, unless @roystgnr has further questions/concerns about the `System::init()` refactoring.

roystgnr (approved):

PR #3978, opened 2024-10-24
Optimize optional RB-EIM data storage

apicard95:
In this update we reshape RB-EIM optional data to be element based instead of quadrature point based in their indexing. 
This helps reducing the number of entries as we remove duplicates. 

This update only affects optional data as it would break too many existing training data if we port that change to other data structure that are not optional (ElemType for instance).


dknez (approved):
This looks good to me, should remove some redundant data that is stored per element rather than per qp.

PR #3979, opened 2024-10-24
Generalize MeshFunction::find_element()

jwpeterson:
Previously, we handled the case of non-local Elems being returned by the `MeshFunction`'s `PointLocator` differently depending on whether the `MeshFunction`'s `_vector` was `SERIAL` or not. In the case where it was not `SERIAL`, we searched the point neighbors of the non-local Elem returned by the PointLocator until a local one was found, or returned `nullptr` if no such local point neighbor could be found. 

There is another important case, however, which is where `MeshFunction`'s `_vector` is `GHOSTED`, and the non-local Elem is still "evaluable" on the local processor, due to e.g. additional ghosting set up by the user. In that case, it should be safe to return the non-local Elem from `MeshFunction::find_element()`. 

This change (together with the one in #3975) fixes an issue in our internal code, but it is possible that it will cause some existing tests to break, since it changes slightly the behavior of `MeshFunction::find_element()`.

### DIFF
--- a/src/mesh/mesh_function.C
+++ b/src/mesh/mesh_function.C
@@ -647,41 +647,8 @@ const Elem * MeshFunction::find_element(const Point & p,
   // locate the point in the other mesh
   const Elem * element = (*_point_locator)(p, subdomain_ids);
 
-  // If the PointLocator did not find an Element containing Point p,
-  // OR if the PointLocator found a local element containting Point p,
-  // we can simply return it.
-  if (!element || (element->processor_id() == this->processor_id()))
-    return element;
-
-  // Otherwise, the PointLocator returned a valid but non-local
-  // element.  Therefore, we have to do some further checks:
-  //
-  // 1.) If we have a SERIAL _vector, then we can just return the
-  //     non-local element, because all the DOFs are available.
-  if (_vector.type() == SERIAL)
-    return element;
-
-  // 2.) If we have a GHOSTED _vector, then we can return a non-local
-  //     Element provided that all of its DOFs are ghosted on this
-  //     processor. That should be faster than the other option, which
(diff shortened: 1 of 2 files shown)


src/mesh/mesh_function.C
```diff
@@ -647,6 +647,45 @@ const Elem * MeshFunction::find_element(const Point & p,
...
+
+  // locate the point in the other mesh
+  std::set<const Elem *> candidate_elements;
+  std::set<const Elem *> final_candidate_elements;
+  (*_point_locator)(p, candidate_elements, subdomain_ids);
+
+  // For each candidate Elem, if it is evaluable, add it to the set of
+  // final candidate Elems.
+  for (const auto & element : candidate_elements)
+    final_candidate_elements.insert(this->check_found_elem(element, p));
```
roystgnr, line 681: We should be filtering out `nullptr` return values (which `check_found_elem` can return) at some point, right?  Since this is a `std::set` I'd just `erase(nullptr)` after the loop here.

roystgnr (changes requested):

src/mesh/mesh_function.C
roystgnr: Nope, never mind.  We were already inserting `nullptr` in that else clause.  This is stupid and we should change it, but probably not in this PR and probably not until we've had some time to make sure users don't report any fallout from this PR that our test suites missed.

roystgnr (approved):

jwpeterson:
I think I agree with you that we should filter out `nullptr`s from the std::set  returning version of the function. I doubt there is existing code that would be affected by the change from "getting a std::set with a single nullptr entry" to "getting an empty std::set", but the change is best left to a different PR.

PR #3983, opened 2024-10-30
Allow creating matrices based on hash tables

lindsayad:
This is only supported by the PETSc backend


roystgnr:
Those look like real failures - somehow in a bunch of use cases we're now trying to zero a matrix before it's initialized?

lindsayad:
Yea, I'll get back to this particular PR soon

lindsayad:
this is ready for review

include/numerics/sparse_matrix.h
```diff
@@ -583,6 +583,32 @@ class SparseMatrix : public ReferenceCountedObject<SparseMatrix<T>>,
...
 
+  /**
+   * @returns Whether the matrix supports hash table assembly
+   */
+  virtual bool supports_hash_table() const { return false; }
+
+  /**
+   * @returns Whether a preference for hash table vs. preallocation has been set
+   */
+  bool assembly_preference_set() const { return _assembly_preference_set; }
```
roystgnr, line 594: How does this differ from just `use_hash_table()`?  If `use_hash_table()` returns `true`, doesn't that mean I've set that as a preference?  If it's `false`, doesn't that in general mean that I've left the default `_use_hash_table` setting on rather than that I've specifically called `do_what_you_were_already_going_to()`?

src/systems/system.C
```diff
@@ -342,18 +344,28 @@ void System::init_matrices ()
...
         this->get_dof_map().attach_matrix(m);
+
+      if (!pr.second->assembly_preference_set())
+        {
+          const bool use_hash =
+              this->_prefer_hash_table_matrix_assembly && pr.second->supports_hash_table();
+          pr.second->use_hash_table(use_hash);
+          if (!use_hash)
+            this->_require_sparsity_pattern = true;
+        }
```
roystgnr, line 355: It looks like we definitely need some test coverage of a system with a mix of hash and non-hash matrices.  Even if it's not hit in CI until we upgrade PETSc there, we'll at least get some early warning on our own machines.

roystgnr (changes requested):

src/systems/system.C
lindsayad: We already have a PETSc sufficient for testing in apptainer. I will look into making something

src/numerics/petsc_matrix.C
```diff
@@ -343,11 +323,33 @@ void PetscMatrix<T>::init (const numeric_index_type m_in,
...
+{
+  // Make it an error for PETSc to allocate new nonzero entries during assembly. For old PETSc
+  // versions this option must be set after preallocation for MPIAIJ matrices
   LibmeshPetscCall(MatSetOption(this->_mat, MAT_NEW_NONZERO_ALLOCATION_ERR, PETSC_TRUE));
-  this->set_context();
-  this->zero();
+
+  this->_is_initialized = true;
+  if (!this->_use_hash_table)
+    this->zero();
```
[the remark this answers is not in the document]
lindsayad: I'm just going to try removing the zeroing

roystgnr:
I take it there's no way via PETSc to get around the need for `copy_from_hash()`?  I'd have thought the way to read `hashed_matrix(i,j)` would be to hash the pair `(i,j)` and read what you find (possibly after a `close()` in parallel), not to copy it into a CSR matrix or whatever and do all reads from that.

Other than that this is looking good now.

lindsayad:
The thing is that there is no "hash table" matrix in PETSc. There is just a "hash table until you call `MatAssemblyBegin/End` and then you have your CSR" matrix. So the moment you call `close` on the "temporarily hash table" matrix you create your CSR matrix and your nonzero pattern is set. So this `copy_from_hash` is necessary to get intermediate snapshots of your matrix before its nonzero pattern has been completely filled. These intermediate snapshots are needed for things like kinematic contact. Maybe reading over [this PETSc issue](https://gitlab.com/petsc/petsc/-/issues/852) would give some more context

src/systems/system.C
lindsayad: I still need to add some testing for this

lindsayad:
As the new example test is currently written, we will need https://gitlab.com/petsc/petsc/-/merge_requests/8063 to fix the failure. I'll probably just skip resetting the preallocation for the "non-hash-table" matrix in the mean time

lindsayad:
I have no idea how this broke MetaPhysicL

lindsayad:
Huzzah!

roystgnr (approved):
This looks ready to go to me, but it's a big enough changeset that I'd say let's let @jwpeterson also have a chance to agree or object before merging.

lindsayad:
Ping on this. Would be nice for me to get moving on the dependent MOOSE PR for BISON support

roystgnr:
Yeah, 2 weeks was enough of a chance.  😆

jwpeterson:
Sorry for not replying sooner, have been out of the country the last week or so. Definitely good with merging!

Issue #3740, opened 2023-12-07
Spurious -I/usr/include/tirpc is added when autodetecting libtirpc in linker path


roystgnr:
A Cardinal+OpenMC user ran into this, on a system where libtirpc is necessary for XDR support but where the headers are still installed in `/usr/include/rpc/`, not `/usr/include/tirpc/rpc/`.  Our first attempt to detect XDR support fails (because we don't yet add `-ltirpc` and the linker has missing symbols), then our next attempt (where we add `-I/usr/include/tirpc` to the compiler flags and `-ltirpc` to the linker's) succeeds (because we aren't giving the compiler anything like `-Wmissing-include-dirs -Werror`), so our m4 concludes that those flags are needed and adds them both.  Our `libmesh-config` script and our pkgconfig files then have that flag, and anything that screams about invalid include directories (like the compiler options above, or reportedly the cmake configuration for Cardinal with OpenMC, which dies with `Imported target "PkgConfig::LIBMESH" includes non-existent path "/usr/include/tirpc"`) will fail.

This isn't urgent, since there's a workaround: `--with-xdr-include=/usr/include` goes straight to the "try to link with `-ltirpc`" configure path and succeeds without adding any invalid include directories.  But this was annoying to debug so I figured I'd at least get the error message and the workaround googleable right away.

PR #3985, opened 2024-10-30
Add --with-xdr-libdir configure option

roystgnr:
This is starting to pass my own tests, so we'll see what CI thinks of it.  (by looking at CI results manually; AFAIK MOOSE has replaced all its XDR/CPR-based tests with XDA/CPA versions now, though it still uses --enable-xdr-required for downstream's sake)


roystgnr:
I need to manually test the case in https://github.com/libMesh/libmesh/issues/3740 too.

roystgnr:
This stuff is nasty to test, but I think it fixes #3740 for me now, as well as doing what it should in the obvious cases I've tried.

jwpeterson (approved):

PR #3994, opened 2024-11-04
Fix lines ending in comma

jwpeterson:
I don't think this was a bug, but it is certainly strange looking, so let's fix it.


roystgnr (approved):
It wasn't technically a bug but it was definitely Obfuscated C++ at best.

PR #3998, opened 2024-11-06
Enable entities (elems, nodes, etc) to be flagged during RB mesh pre-evaluation

apicard95:
* This update introduces a generic map in RB that can store various sets of dof_id_types. This can be used to identify elems, nodes, components and apply a specific treatment to those entities during EIM assembly.
* The map is stored in VectorizedEvalInput and transferred to RBParametrizedFunctions during mesh pre-evaluation (all qp data).
* This map is then used to fill a second map that will store the same properties but only for EIM points. This second map is stored in RBEIMEvaluation and can be serialized and deserialized during the training.

### DIFF
--- a/contrib/capnproto/rb_data.capnp
+++ b/contrib/capnproto/rb_data.capnp
@@ -45,6 +45,11 @@ struct IntPair @0xa9e786708685aed5 {
   second @1 :Integer;
 }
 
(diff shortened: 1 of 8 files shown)


jwpeterson (approved):
This looks good to me, but I think I will give you a chance to do some internal testing before merging? Let me know if you prefer to just merge now.

PR #4000, opened 2024-11-07
Fixes for NetCDF 4.6.2 tests run with HDF > 1.12

jwpeterson:
This is a patch by @ostueker which he developed on the 1.7.x release branch. I tested it there (albeit only with older HDF) and it worked fine, so it should probably also be on the 1.8.x branch as well as master.


ostueker:
This patch is based on Unidata/netcdf-c#1671 and was adapted so that it can be applied to `libmesh/contrib/netcdf/netcdf-c-4.6.2`.

roystgnr (approved):

PR #4001, opened 2024-11-07
DofMap members to retrieve the no. of dofs per variable

nmnobre:
Open to renaming or adding shortcuts to `System` as well if people would like so.

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -696,7 +696,8 @@ public:
 
   /**
    * \returns The number of degrees of freedom on this processor for a
-   * particular variable \p vn.
+   * particular variable \p vn.  This is an O(N) operation on serialized or
+   * O(N/Nproc) operation on distributed meshes.
    */
   dof_id_type n_local_dofs(const unsigned int vn) const
   {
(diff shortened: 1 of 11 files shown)


include/base/dof_map.h
```diff
@@ -679,6 +694,17 @@ class DofMap : public ReferenceCountedObject<DofMap>,
   dof_id_type n_local_dofs () const
   { return this->n_dofs_on_processor (this->processor_id()); }
 
+  /**
+   * \returns The number of degrees of freedom on this processor for a
+   * particular variable \p vn.
```
roystgnr, line 699:
```suggestion
   * \returns The number of degrees of freedom on this processor for a
   * particular variable \p vn.  This is an O(N) operation on serialized or
   * O(N/Nproc) operation on distributed meshes.
```

Allocating the full vector here just to return its size and destruct it isn't as awful an idea as it looks at first glance (since the alternatives are to either duplicate code or do some fancy templating in the backend), but we should probably warn users that this call is a lot more expensive than the similarly-named O(1) calls around it.

roystgnr (changes requested):

roystgnr (approved):
This is great.

PR #4003, opened 2024-11-08
Add Netgen to distributed tarballs

roystgnr:
We seem to have been passing `make distcheck` only because `configure` gracefully disabled Netgen for us?

I'm still testing builds of tarballs-now-with-netgen myself, but I think it's working.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -21,4 +21,4 @@
 	url = https://github.com/jlblancoc/nanoflann.git
 [submodule "contrib/netgen/netgen"]
 	path = contrib/netgen/netgen
-	url = ../../NGSolve/netgen.git
+	url = ../../libMesh/netgen.git
--- a/contrib/netgen/Makefile.am
+++ b/contrib/netgen/Makefile.am
@@ -35,9 +35,9 @@ install-data-local: install-ngliblibDATA
 	@if which install_name_tool 2>/dev/null; then \
 	  echo "install_name_tool exists on OSX but is unneeded for rpath-based libnglib there"; \
 	elif which patchelf 2>/dev/null; then \
-	  patchelf --set-rpath "$(libdir)" "$(libdir)/libnglib$(LIBMESH_LIBRARY_SUFFIX)"; \
+	  patchelf --set-rpath "$(libdir)" "$(DESTDIR)$(libdir)/libnglib$(LIBMESH_LIBRARY_SUFFIX)"; \
 	elif which chrpath 2>/dev/null; then \
-	  chrpath -r "$(libdir)" "$(libdir)/libnglib$(LIBMESH_LIBRARY_SUFFIX)"; \
+	  chrpath -r "$(libdir)" "$(DESTDIR)$(libdir)/libnglib$(LIBMESH_LIBRARY_SUFFIX)"; \
 	else \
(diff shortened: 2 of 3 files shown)


jwpeterson:
Once this is merged to master, it will also need to be cherry-picked onto branch_v1.8.0. It was while testing `make dist` on the release branch that we first realized the netgen sources were not in the distribution tarball. Once that problem was fixed, further complications arose with actually building netgen from the release tarball.

jwpeterson (approved):

roystgnr:
Wait, did we just *pass* the `distcheck` test in CI?  I swear I'm still having some issues locally.

jwpeterson:
Difficult to say what happened exactly. The output of the DistCheck step [here](https://civet.inl.gov/job/2540813/) exceeded the output size limit:

```
*****************************************************

CIVET: Output size exceeded limit (5242880 bytes), further output will not be displayed!

*****************************************************
```

jwpeterson:
Looks like Netgen was configured and built during the DistCheck step:
```
  Netgen :  Automatic configuration OK.
```
...
```
[ 99%] Building CXX object CMakeFiles/nglib.dir/nglib/nglib.cpp.o
[100%] Linking CXX shared library libnglib.so
```
It looks like it was in the process of deleting everything when the output size limit was exceeded, so it kind of seems like it worked.
```
Making uninstall in examples
```

jwpeterson:
@roystgnr what issues are you still seeing locally with this branch? I pulled this branch and did the following:

1. `make dist`
2. Untar tarball in fresh directory.
3. `configure`, ensure that netgen was enabled
4. `make install`
5. `make check`
6. Go to unit test directory and run: `./unit_tests-opt --re testNet`

The last step produced the following output:
```
Will run the following tests:
MeshTetTest::testNetGen
MeshTetTest::testNetGenError
MeshTetTest::testNetGenTets
MeshTetTest::testNetGenFlippedTris
MeshTetTest::testNetGenHole
MeshTetTest::testNetGenSphereShell
--- Running 6 tests in total.
......


OK (6 tests)
```

So, as far as I can tell, this fixes all the issues that we had before with `make dist` and `netgen`. 

I think this is the last thing (or one of the last things) I'm going to cherry-pick onto `branch_v1.8.0` before tagging `1.8.0-rc1` so let me know ASAP if there is still some issue to be resolved here.

jwpeterson:
@roystgnr I tested your `$DESTDIR` update and it still worked fine for the sequence of tests I mentioned in this [comment](https://github.com/libMesh/libmesh/pull/4003#issuecomment-2474695418), so if it also fixes the `distcheck` errors you were seeing locally, I think we should be good to merge.  (I've already approved this PR, but if I could approve it again, I would :face_holding_back_tears:)

roystgnr:
I wish we were good to merge.  The `$DESTDIR` fix gets my build working fine, but then the subsequent *distclean* fails because I didn't delete my build directory.  I'm actually not seeing anything outside the build directory (the only failure there was fixed in our netgen branch and not general to cmake) so this ought to be an easy fix too.

jwpeterson:
I confirmed that `make distclean` also works when building from the tarball. I will go ahead and merge when the testing is done, unless you get to it first.

PR #4004, opened 2024-11-11
DMSection changes in PETSc 3.23

lindsayad:
Also simplify the PETSc version macros. Don't know if I'll get some pushback on that


roystgnr (approved):
Jed's "fix PETSC_VERSION_LT() to require PETSC_VERSION_RELEASE=1" fix was back in 2013?  I'm okay with just relying on their macro now.

PR #4005, opened 2024-11-12
Fixes for RBEIMConstruction::get_random_point() in parallel.

dknez:
Need to do a set_union() in order to ensure we have data from all processors.


jwpeterson (approved):
Looks good to me. There seems to be something wrong with the CI right now, I'll check on that.

PR #4011, opened 2024-11-18
Amend previous workaround for LibmeshPetscCall

nmnobre:
@jwpeterson


jwpeterson:
Thanks @nmnobre, I attached the "Test No Exceptions" recipe to this PR, that should let us know whether the devel -> master merge is also going to pass.

jwpeterson (approved):

PR #4013, opened 2024-11-20
Optimize SparseMatrix::read_matlab()

roystgnr:
I finally had occasion to try out `read_matlab()` on some really sizable matrices (half a gigabyte unzipped, converted from ~150M 32-bit-PETSc-format files), and it was way slower than it should have been.  It's *still* maybe an order of magnitude slower than it should be, but even iostream isn't as badly optimized as regex, so it's now about 15x faster than it was originally, and 5 seconds is tolerable whereas 75 seconds was not.


jwpeterson (approved):
Pretty bad when iostreams looks fast in comparison to anything.

Issue #4014, opened 2024-11-21
dynamic_cast failures on MacOS: A consequence of Apple's Command Line Tools' update. 


colegruninger97:
Hello,

Just a heads up: after updating to MacOS command line tools 16.1, I encountered dynamic_cast failures in IBAMR (which uses libmesh to handle finite elements) when compiler optimizations are enabled. The issue occurs in the following function:

```
inline void
copy_and_synch(libMesh::NumericVector<double>& v_in,
               libMesh::NumericVector<double>& v_out,
               const bool close_v_in = true,
               const bool close_v_out = true)
{
#if defined(NDEBUG)
    auto v_in_petsc = static_cast<libMesh::PetscVector<double>*>(&v_in);
    auto v_out_petsc = static_cast<libMesh::PetscVector<double>*>(&v_out);
#else
    auto v_in_petsc = dynamic_cast<libMesh::PetscVector<double>*>(&v_in);
    auto v_out_petsc = dynamic_cast<libMesh::PetscVector<double>*>(&v_out);
    TBOX_ASSERT(v_in_petsc);
    TBOX_ASSERT(v_out_petsc);
#endif
    if (close_v_in) v_in.close();
    PetscErrorCode ierr = VecCopy(v_in_petsc->vec(), v_out_petsc->vec());
    IBTK_CHKERRQ(ierr);
    if (close_v_out) v_out.close();
}
```

The assertion error appears to be caused by a compiler bug associated with how the updated compiler handles virtual tables for shared libraries (see: https://stackoverflow.com/questions/79192304/macos-xcode-16-breaks-dynamic-cast-for-final-types-defined-in-shared-library).

Following the stackoverflow post, I found that removing the `final` keyword from the PetscVector class declaration in libmesh/include/numerics/petsc_vector.h resolves the issue. I suppose this issue will crop up whenever an object from a class marked as 'final' is dynamically cast with optimizations turned on and so removing the final keyword could provide a workaround until this compiler bug is fixed.

roystgnr:
Oh, what a travesty.  We still have a `libmesh_final` macro, for backwards compatibility with really old application codes, but looking through our logs, it was all the way back in 2018 that we decided we'd given compilers enough time to support C++11 correctly and we switched all our internal uses to plain `final`.

[...]

PR #4015, opened 2024-11-21
Test for dynamic_cast compiler bug

roystgnr:
See issue #4014.

I considered adding a workaround in another commit here, where we move all our defaulted virtual destructors into the .C files to try to convince the new XCode that that really is the proper place to create a single RTTI instantiation for each class.  One downside would have been that this would prevent inlining of the destructors of `final` classes, but a bigger downside is that grep says we have 144 header files with defaulted virtual destructors that would need to be moved.  Gross.

### DIFF
--- a/tests/numerics/numeric_vector_test.h
+++ b/tests/numerics/numeric_vector_test.h
@@ -13,6 +13,7 @@
 #include <memory>
 
 #define NUMERICVECTORTEST                       \
+  CPPUNIT_TEST( testCasting );                  \
   CPPUNIT_TEST( testLocalize );                 \
   CPPUNIT_TEST( testLocalizeBase );             \
   CPPUNIT_TEST( testLocalizeIndices );          \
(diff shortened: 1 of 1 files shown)


jwpeterson:
> I considered adding a workaround in another commit here, where we move all our defaulted virtual destructors into the .C files

I don't think the issue is specific to whether the virtual destructor is inlined or not, right? 

> Those StackOverflow answers are suggesting that XCode is now generating RTTI information in the compilation unit with the definition of the first non-inline virtual function and that the problem thus only occurs for classes with no non-inline virtual functions, and that can't be true here because NumericVector and PetscVector do have non-inline virtual functions

The SO article doesn't include an example where the virtual destructor is declared out-of-line, but *other* virtual functions are still declared inline (the case we have with NumericVector), so I'm curious if the compiler would still have the bug in that case. If so, then moving the destructors out-of-line wouldn't change anything...

jwpeterson:
BTW, there is at least one good reason to declare destructors out-of-line. If your class `X` contains a `std::unique_ptr<T>`, then you only need a forward declaration of `T` in `X.h` unless you try to define `~X()` (e.g. by defaulting it) in the header file. As I understand it, the `X` destructor automatically generates a call to the destructor of `T`, which then requires a complete type for `T` to be available.

colegruninger97:
> > I considered adding a workaround in another commit here, where we move all our defaulted virtual destructors into the .C files
> 
> I don't think the issue is specific to whether the virtual destructor is inlined or not, right?
> 
> > Those StackOverflow answers are suggesting that XCode is now generating RTTI information in the compilation unit with the definition of the first non-inline virtual function and that the problem thus only occurs for classes with no non-inline virtual functions, and that can't be true here because NumericVector and PetscVector do have non-inline virtual functions
> 
> The SO article doesn't include an example where the virtual destructor is declared out-of-line, but _other_ virtual functions are still declared inline (the case we have with NumericVector), so I'm curious if the compiler would still have the bug in that case. If so, then moving the destructors out-of-line wouldn't change anything...


In the petsc_vector.h header file, I tried the following potential solutions:

1. Moving virtual destructors out-of-line (to .C files): dynamic_cast fails
2. Moving both virtual destructors AND virtual functions out-of-line: dynamic_cast fails
3. Removing the "final" keyword: works in all cases, regardless of whether functions are inline or out-of-line

This suggests the issue with the compiler might have something to do specifically with the "final" keyword.

PR #4017, opened 2024-11-24
Fix in enrich_eim_approximation_on_sides().

dknez:
The broadcast of optimal_side_index was missing, which could give invalid results in parallel. Fix is to add the broadcast in for this property, similar to the other properties that we broadcast.


jwpeterson (approved):
The CI test failure is unrelated.

PR #4018, opened 2024-11-26
Scrape Fortran compiler from PETSc too

roystgnr:
Failing to do this was causing me a ton of linker problems in MOOSE modules on a system that had both a /usr/bin/mpif90 and a PETSc-specific mpif90.


jwpeterson:
Does this need to be on the 1.8.0 release branch as well?

roystgnr:
I'd bring it in, I think.  It doesn't do anything for libMesh since we don't build any Fortran ourselves, but it's pretty important for any applications relying on our `libmesh_config --fc` in any way.

PR #4029, opened 2024-12-02
Remove ACX_CHOOSE_PRECISION

nmnobre:
And what about this @roystgnr? Safe to remove? xD

I'd probably also delete this line since I don't think it's being used by none of libMesh, MetaPhysicL or TIMPI and just avoids the side-effect in the future.
https://github.com/libMesh/autoconf-submodule/blob/e7e18c44be9615ec4f35907a8f6a3b2f8a6678f7/acsm_precision.m4#L103


roystgnr:
This isn't safe to remove.  It's not a feature that got replaced by a newer version of the same feature or otherwise obviated, it's the entire reason why we write `Real` everywhere instead of `double`.  It's not a commonly used feature, which is why we don't have it in CI, but apparently I need to fix that before someone tries to remove it.  ;-)

nmnobre:
> This isn't safe to remove. It's not a feature that got replaced by a newer version of the same feature or otherwise obviated, it's the entire reason why we write `Real` everywhere instead of `double`. It's not a commonly used feature, which is why we don't have it in CI, but apparently I need to fix that before someone tries to remove it. ;-)

No, of course, I understand that (we discussed single, triple and quadruple precision support not too long ago), I just replaced the call with the respective acsm call, the functionality remains.

nmnobre:
Extra context:
```
diff m4/precision.m4 m4/autoconf-submodule/acsm_precision.m4
2c2
< dnl Accept options for single-precision or triple-precision scalars
---
> dnl Accept options for single-, triple-, quadruple-precision scalars
5c5
< AC_DEFUN([ACX_CHOOSE_PRECISION],
---
> AC_DEFUN([ACSM_CHOOSE_PRECISION],
25c25
< libmesh_precision_LIBS=""
---
> acsm_precision_LIBS=""
60c60
<               [libmesh_precision_LIBS="-lquadmath"
---
>               [acsm_precision_LIBS="-lquadmath"
78c78
<               [libmesh_precision_LIBS=""
---
>               [acsm_precision_LIBS=""
103c103
<   AC_SUBST(libmesh_precision_LIBS)
---
>   AC_SUBST(acsm_precision_LIBS)
```

nmnobre:
Just to be clear: I didn't try to remove functionality, just a redundant file. :-)

roystgnr:
Oh!  I didn't see that you'd left the ACSM version.

nmnobre:
> Okay, *this* time I think I'm closing this one for real; I didn't have permission to push a re-bootstrap commit to your fork, so I put it in #4030

Thanks! I'm away from office most of this week and I managed to lock myself out of the VPN I need to access my typical development machine, so I submitted a bit blindly and forgot to re-bootstrap.

> 
> 
> Anyway, this passes my tests with single precision, at least, by which I mean it reveals that we've managed to regress some code in single precision.

Ouch! Hopefully that's not a byproduct #3910?

PR #4031, opened 2024-12-04
Delete boost

lindsayad:
Our current boost is not compatible with c++17


jwpeterson:
> Our current boost is not compatible with c++17

Just curious, what exactly isn't working? Also, are the diff statistics correct? Looks like we are removing a lot more lines than we're adding...

roystgnr:
We are removing a ton of files here.

On the one hand that isn't obviously crazy for our internal purposes: if boost has gotten better about avoiding unnecessary includes, then `bcp` of just the 5 headers our README asks for could suddenly lose a lot of dependencies.

On the other hand, that README is clearly out of date, still asking for boost smart pointers we no longer use, and not bothering to ask for quad-precision support we try to make available.

On the gripping hand ... why are we still providing contrib/boost?  Feels like it might have been critical at some point before C++11 smart pointers, but internally it's all purely optional quad-precision and pool-allocator stuff, and externally we don't know what our users are relying on.  Looking through MOOSE I see the stochastic tools module already having to check `LIBMESH_HAVE_INTERNAL_BOOST` to make sure they don't assume the existence of advanced math functions we don't provide.

Would it be crazy to drop contrib/boost entirely?  Non-external-boost-users might lose a touch of performance when modifying a DistributedMesh, and wouldn't be able to build with `--enable-quadrupleprecision`, and that would be it.

jwpeterson:
We always use an external boost for our internal code, I'd be OK with getting rid of contrib/boost.

lindsayad:
I was hoping to add a boost header such that downstream users do not have an additional external library dependency

lindsayad:
> Just curious, what exactly isn't working?

This is the compilation error I got on my ARM Mac yesterday:

> libmesh/installed/include/boost/numeric/conversion/detail/converter.hpp:453:47: error: no template named 'unary_function' in namespace 'std'; did you mean '__unary_function'

lindsayad:
I personally think it's useful to just have some fall-back boost capability in libMesh. I would be all for adding the boost headers that stochastic tools relies on

roystgnr:
> and downstream users do not have an additional external library dependency

Do we count optional dependencies as dependencies here?  If so then we'd better figure out how to saddle them with hunting down capnproto.  If not then we're fine.  ;-)

Though, I should probably benchmark `DistributedMesh` with and without a real pool allocator before blithely making it harder to enable.  My intuition is that the performance degradation of falling back on ordinary allocation would be less than the (also relatively minor) performance improvement of just finally switching our default backend to a conservative `chunked_mapvector setting`, but I've never actually measured it.

>I personally think it's useful to just have some fall-back boost capability in libMesh.

The trouble with this is deciding *what* fall-back to provide.  "Everything most users would benefit from" or "everything a substantial minority of users would benefit a lot from" is zero.  "Everything we ever use in libMesh itself" is not zero, but it's small enough that it may be too small for many downstream users.

> I would be all for adding the boost headers that stochastic tools relies on

This isn't a crazy idea.  Go through all the "standard" MOOSE apps too to accumulate "Everything we know typical downstream users might need", and then bcp the union of those?  John, could you give us a list of boost headers Akselos uses, to possibly add?  I don't really have any other suggestions myself; GRINS and my personal research codes never used Boost, and the only thing my old copy of FIN-S uses is `boost/test` for unit testing.

jwpeterson:
> John, could you give us a list of boost headers Akselos uses, to possibly add?

I ran `git grep -h "boost/" | sort | uniq` and it returned the following list:
```
#include <boost/algorithm/string/classification.hpp>
#include <boost/algorithm/string/compare.hpp>
#include <boost/algorithm/string/constants.hpp>
#include <boost/algorithm/string.hpp>
#include <boost/algorithm/string/predicate.hpp>
#include <boost/algorithm/string/split.hpp>
#include <boost/filesystem.hpp>
#include <boost/filesystem/operations.hpp>
#include <boost/foreach.hpp>
#include <boost/format.hpp>
#include <boost/lexical_cast/bad_lexical_cast.hpp>
#include <boost/lexical_cast.hpp>
#include <boost/make_shared.hpp>
#include <boost/math/special_functions/bessel.hpp>
#include <boost/math/special_functions/sign.hpp>
#include <boost/optional.hpp>
#include <boost/ref.hpp>
#include <boost/scope_exit.hpp>
#include <boost/uuid/uuid_generators.hpp>
#include <boost/uuid/uuid.hpp>
#include <boost/uuid/uuid_io.hpp>
```
however I'm guessing some of these aren't relevant for us any longer (e.g. `<boost/make_shared.hpp>`) so that's likely not an authoritative list.

> libmesh/installed/include/boost/numeric/conversion/detail/converter.hpp:453:47: error: no template named 'unary_function' in > namespace 'std'; did you mean '__unary_function'

OK, interesting, it looks like that was indeed [removed in C++17](https://en.cppreference.com/w/cpp/utility/functional/unary_function) but I guess all of our other STL implementations have decided to keep it around for a while..

jwpeterson:
And I haven't checked recently, but in the past, `boost/filesystem` had a compiled library component, so we couldn't just provide header file only support for it in libmesh.

lindsayad:
Ha, you guys also use `lexical_cast` 😄

lindsayad:
@jwpeterson if you wanted to say https://github.com/libMesh/libmesh/pull/4031#issuecomment-2521026023 is an authoritative list or you want to update that list, then I would vote to compile those and what we use in the MOOSE herd to re-generate the boost headers I add in this PR.

> And I haven't checked recently, but in the past, `boost/filesystem` had a compiled library component, so we couldn't just provide header file only support for it in libmesh.

Definitely would be nice to avoid this. Especially given `std::filesystem` support

jwpeterson:
I cleaned up our includes a bit and regenerated the list in https://github.com/libMesh/libmesh/pull/4031#issuecomment-2521026023. We probably won't refactor all of our uses of `boost::filesystem` into `std::filesystem` calls any time soon, so we'll continue to have an external boost dependency for now. So, you are welcome to include the dependencies of these headers in the libmesh internal boost, but it's not critical because we probably won't end up using it.

lindsayad:
You know what ... I think this is a bigger maintenance headache than it's worth. I agree with you guys, let's just burn it to the ground 😆

jwpeterson:
Looks like there is a build failure in TIMPI when libmesh has no internal boost.
```
cc1plus: error: /include/boost-0: No such file or directory [-Werror=missing-include-dirs]
```
@roystgnr is there boost code in TIMPI that doesn't have a non-boost fallback? It's possible that we also have some in libmesh too, but the pool allocator and the `boost::multiprecision` stuff looks like it is guarded correctly.

lindsayad:
I went to reproduce but of course like a dummy didn't think about my external boost installation

lindsayad:
Merp went to rod to reproduce with the container but that didn't help me since rod has a system boost install

lindsayad:
I guess I'll just temporarily uninstall boost on my local machine

jwpeterson:
In theory, you could configure with `--disable-boost` and see if TIMPI is happy, but it's possible that TIMPI configure would then go and find an external boost installation? I think we are proving that it's actually harder to find a system *without* boost installed than *with* boost :laughing: 

One comment is: I see that there are changes to our boost.m4 on this PR, but GitHub is having trouble displaying them, possibly due to the large size of whole the diff. It might be good to put the boost.m4 changes in a separate commit, but if it's too much trouble don't worry about it.

roystgnr:
Was that build failure definitely in TIMPI?  TIMPI only knows boost exists if `--enable-quadrupleprecision` is used, in which case it just assumes it can ask for boost headers.

`/include/boost-0: No such file or directory` is weird ... it's as if it thinks it detected boost, but with an empty prefix and no version info.  In libMesh `boost.m4` we do hand our `$BOOST_CPPFLAGS` to TIMPI configure, but only if `$enableboost` is still `yes` at the end.

jwpeterson:
> Was that build failure definitely in TIMPI?

You can check one of the failing logs, but I think It was configuring TIMPI at the time that error occurred.

lindsayad:
It was an issue in the `boost.m4` that I introduced which was giving TIMPI silly flags

jwpeterson (approved):
Looks good to me, thanks for making a separate commit for the boost.m4 changes.

lindsayad:
@roystgnr I'll probably merge this in a couple hours unless you indicate otherwise

nmnobre:
> Would it be crazy to drop contrib/boost entirely? Non-external-boost-users might lose a touch of performance when modifying a DistributedMesh, and wouldn't be able to build with --enable-quadrupleprecision, and that would be it.

In fact, according to what I wrote in #3910, users weren't already able to build with `--enable-quadrupleprecision` with the built-in boost, so I think this was a good decision. :)

PR #4034, opened 2024-12-07
Add mpi thread command line option

lindsayad:
Pairs with https://github.com/libMesh/TIMPI/pull/137. Error message without these changes when running strumpack LU with slate (for GPU builds even if using run-time cpu):
```
[0]PETSC ERROR: No support for this operation on this system
[0]PETSC ERROR: SLATE requires MPI_THREAD_MULTIPLE
```

### DIFF
--- a/src/base/libmesh.C
+++ b/src/base/libmesh.C
@@ -383,6 +383,11 @@ LibMeshInit::LibMeshInit (int argc, const char * const * argv,
       libMesh::perflog.disable_logging();
   }
 
+  auto check_empty_command_line_value = [](auto & cl, const std::string & option) {
+    libmesh_error_msg_if(cl.search(option),
+                         "Detected option " << option << " with no value.  Did you forget '='?");
+  };
+
   // Build a task scheduler
   {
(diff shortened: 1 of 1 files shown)


lindsayad:
@roystgnr This is one way of approaching the slate failures. However, per [this discussion](https://discord.com/channels/1119324534303109172/1314739078754861056/1314739078754861056) on PETSc discord (you may not be able to follow the link; let me know if you want me to add you to the PETSc discord), I feel like a better approach is probably just to automatically request `MPI_THREAD_MULTIPLE` when slate is being used for strumpack, like is done in [this PETSc test](https://discord.com/channels/1119324534303109172/1314739078754861056/1314851242694152252), without any intervention by the general user who likely knows little/nothing about different thread support levels in the mpi library

roystgnr (approved):
I definitely want command-line control over this, but I'd love it if the default was determined by things like autodetection of slate/strumpack.  My only reason for avoiding an aggressive default long ago was that some MPI implementations wouldn't handle it back then; that might still be the case but it surely won't apply to users who are building a stack that requires an aggressive request.

PR #4043, opened 2025-01-06
Add backward compatibility for older RB-EIM data (read-only)

apicard95:
In summary, we convert old format training data to the new format during the reading stage. 

Some older training data are qp based for several optional fields (JxW_all_qp, phi_i_all_qp, dxyzdxi, dxyzdeta...), the new version is elem based. To maintain backward compatibility when reading older training data, we check if n_elems is present in the set. If so and if the field is not empty, we lookup based on elems, otherwise we lookup based on qp but we make sure to remove potential duplicates and store that data based on elems. 
The remaining optional field required is an elem_id_to_local_indexing map that is built during reading if optional fields are present and the training data is old format.

### DIFF
--- a/src/reduced_basis/rb_data_deserialization.C
+++ b/src/reduced_basis/rb_data_deserialization.C
@@ -869,6 +869,135 @@ void load_rb_eim_evaluation_data(RBEIMEvaluation & rb_eim_evaluation,
   }
 
   unsigned int n_elems = rb_eim_evaluation_reader.getNElems();
+  bool build_elem_id_to_local_index_map = false;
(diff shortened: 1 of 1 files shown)


jwpeterson (approved):
@apicard95 Are you ready for this to be merged now or are you still doing some internal testing?

apicard95:
@jwpeterson Almost all the tests are passing locally with local training. Only 1 or 2 have shown to be diffing compared to master but considering the current state of the pipelines, I don't think it is directly related to those changes. This PR should be more of a pass or break type of update.

PR #4045, opened 2025-01-06
Added option to normalize RB solution snapshots

dknez:
It can be useful to normalize solution snapshots in RB training, e.g. this is relevant if snapshots have different magnitudes and we want to treat them all on "equal footing."


src/reduced_basis/rb_construction.C
```diff
@@ -1453,6 +1453,20 @@ void RBConstruction::train_reduced_basis_with_POD()
     }
   libMesh::out << std::endl;
 
+  if (normalize_solution_snapshots)
+  {
+    std::cout << "Normalizing solution snapshots" << std::endl;
```
jwpeterson, line 1458: We use `libMesh::out` for print statements inside the library.

src/reduced_basis/rb_eim_construction.C
```diff
@@ -644,6 +620,51 @@ Real RBEIMConstruction::train_eim_approximation_with_greedy()
   return greedy_error;
 }
 
+void RBEIMConstruction::apply_normalization_to_solution_snapshots()
+{
+  LOG_SCOPE("apply_normalization_to_solution_snapshots()", "RBEIMConstruction");
+
+  std::cout << "Normalizing solution snapshots" << std::endl;
```
jwpeterson, line 627: libMesh::out

jwpeterson (commented):
This is the second PR where I'm seeing failures during fetch and branch for ARM machines. @roystgnr @lindsayad any ideas?
```
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
Python runtime state: core initialized
ModuleNotFoundError: No module named 'encodings'
```

src/reduced_basis/rb_construction.C
dknez: oops, thanks, fixed in the update I just pushed

jwpeterson (approved):

lindsayad:
looks like someone invalidated it and now it's stuck. Let's try again

PR #4047, opened 2025-01-13
Refactor some code

nmnobre:


include/base/dof_map.h
```diff
@@ -719,7 +719,7 @@ class DofMap : public ReferenceCountedObject<DofMap>,
    * \returns The number of degrees of freedom on each partition for a
    * particular variable \p vn.
    */
-  std::vector<dof_id_type> n_dofs_on_each_processor(const unsigned int vn) const
```
roystgnr, line 722: What's the reason for this?  I found the old name just as understandable, and even if it wasn't we wouldn't want to instantly break the API compatibility with it with no transition period for the migration.

include/numerics/petsc_vector.h
```diff
@@ -179,6 +179,8 @@ class PetscVector final : public NumericVector<T>
 
   virtual T sum () const override;
 
+  template <NormType N> Real lx_norm () const;
```
roystgnr, line 182: No documentation, and if there was it would be "the argument here is a raw PETSc value"?

We need to document this and move it from `public:` to `private:` or `protected:`.

include/geom/elem.h
```diff
@@ -1080,6 +1080,12 @@ class Elem : public ReferenceCountedObject<Elem>,
    */
   bool positive_face_orientation(const unsigned int i) const;
 
+  /**
+   * \returns \p true if the vertices of edge \p e appear in edge_nodes_map in
+   * the same order they appear in side_nodes_map for side \p s.
```
roystgnr, line 1085: Pretend `edge_nodes_map` and `side_nodes_map` don't exist.  (I'm currently writing element classes where they don't.)  Rewrite with that in mind?

Something like "returns true iff the local node numbering of edge e matches the local node numbering of that edge when viewed as a side of side s"?

I'm also struggling a bit to understand the use of this.  Presumably you need a fallback code path for where it returns false?  Is it that much of an optimization to test it and use a faster code path when it returns true?  But it can't be, if you use it on every element, and yet not making it a `static` method suggests it might be true or false on an element-by-element basis.

I'd also hope to see some unit testing for this, perhaps something that demonstrates why it would be useful.

roystgnr (changes requested):
I'd like the `lx_norm()` refactor if you get it out of `public:`.  I'm not sure the rest is salvageable, but I'm open to being talked out of that.

include/base/dof_map.h
nmnobre: The reason is simple, but weak: the name is long and it was taking too much of my lines of code, that's all.
Since I introduced the function two months ago and I couldn't see no usage otherwise, I thought I might be in the grace period. Happy to drop.

include/numerics/petsc_vector.h
nmnobre: So, this was something I did when I thought I needed l2 for matrices, and noticed the redundancy in the code, that's all. I agree it should be private, that's an oversight. Let me change that around and add some docs.

include/geom/elem.h
nmnobre: > I'm also struggling a bit to understand the use of this.
> perhaps something that demonstrates why it would be useful.

Yep, my bad. It was always a risk to put this up without proper motivation.
In short, I'm trying to remove commits out the preconditioners PR-to-be (it's probably time I put this up, because I need your help with some design choices anyway).
For the ADS preconditioner, I need to know the relative orientation of faces to their defining edges and so I need to assess both orientations as if I'm cycling (clockwise for instance) through the face, one vertex after the other. That's the reason for this functionality (it works).

include/base/dof_map.h
roystgnr: Oh, wow, Nov. 6 commit, merged even later than that?  Never mind, then, I'm cool with the change.  Anyone keeping up with libMesh *that* closely can just keep doing so.  ;-)

include/geom/elem.h
nmnobre: Okay, I've removed that commit.

roystgnr (approved):
Looks good.  Thanks, and thanks for your patience with me ragging on the first version!

roystgnr:
I definitely don't want to call it `l2_norm()`, since to me that suggests the maximum singular value (the operator norm induced by using the l2 norm on both the domain and codomain), but I'd consider `frobenius_norm()` an improvement.

jwpeterson (approved):

roystgnr (approved):

PR #4048, opened 2025-01-13
netcdf: remove an autogenerated file.

drwells:
This shouldn't be checked in and, due to the order of includes in `NETCDF_INCLUDE`, prevents the correct header from being read.

I found this while working on 1.7.6: it looks like the other half of this bug present on that branch (not adding `-I/$(top_srcdir)/contrib/netcdf/netcdf-c-4.6.2/include/` to `NETCDF_INCLUDE`) was already fixed.


roystgnr (approved):

roystgnr:
Looks correct to me; if @jwpeterson doesn't chime in shortly please ping me again so I don't forget to merge.

jwpeterson:
It looks like this file has been there since 5fad6180ef58be8135309b4997767bfb20c53bac... Definitely agree that it should be removed, just not sure how it hasn't broken anything else before now.

PR #4051, opened 2025-01-17
Add `operator[]` for `DenseVector`

lindsayad:
I think of `operator[]` as being most appropriate for already allocated data whereas I usually associate `operator()` with on-the-fly functor evaluations. Obviously we are not going to remote `operator()` (and I believe some would make a math notation argument for keeping it anyway which I'm not against), but I think we should support `operator[]`. This allows some templated code that takes things like `std::vector` and `std::array` to also work with `DenseVector`

### DIFF
--- a/include/numerics/dense_vector.h
+++ b/include/numerics/dense_vector.h
@@ -121,6 +121,16 @@ public:
    */
   T & operator() (const unsigned int i);
 
+  /**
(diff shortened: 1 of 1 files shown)


roystgnr (approved):
I'd even add [] to TypeVector while we're at it.  The ()/[] discrepancy never bothered me much but I never understood the reason for it either.

PR #4052, opened 2025-01-19
Fix 2d H(curl) and H(div) formulations embedded in 3d space

nmnobre:


roystgnr:
From Slack just now:

> Sorry for the wait but good news, the changes with the commit fixed the pure-2D with rotation eigenvalue discrepancy. 

> For reference, before the change, the first three non-zero eigenvalue were:

> * No Rotation: 5.7854890369939, 14.687709716912, 26.384752214715 / With Rotation: 8.6489407338644, 18.264283560678 25.571975954959

> Now with the changes, the first three non-zero eigenvalue are:

> * No Rotation: 5.7854890369939, 14.687709716912, 26.384752214715 / With Rotation: 5.785489036994, 14.687709716912, 26.384752214714

Anything else we should wait for before merging?

nmnobre:
> Anything else we should wait for before merging?

Depends on how much more effort we are willing to put into this. I looked a bit into supporting what you call "mixed meshes" in the code in 8f01f22, in particular regarding output/visualisation. It seems to work, but I don't think we can escape modifying the interface to `FEInterface::nodal_soln()` to also take `n_vec_dim`, and I don't know if you're okay with that? Or do you have other ideas? Is this going to open a can of worms?

I also think Vector Ex. 3 looks a bit weird now, it would be nicer to allow an arbitrary rotation and evaluate the error on the rotated mesh (which we can do), but then if we don't do the above, we can't plot the result. I was thinking maybe I set the default to no rotation (i.e. the identity) and keep the output but add a note saying that if you rotate the mesh, you have to rotate the mesh back into the xy plane and hope the orientations still hold.

nmnobre:
So, here's the plan: this PR will only fix the math, i.e. the previously incomplete Piola mappings.

roystgnr (approved):

PR #4055, opened 2025-01-21
Fix timesteps in exodus output

nmnobre:
This fixes the 9.96921e+36 timestamp in ParaView for exodus outputs with a single time step: we now set it to time 0, instead of leaving it undefined.

This only affects builds configured with `--enable-hdf5` such that `LIBMESH_HAVE_HDF5` is defined, which in turn causes the Exodus output to be in NETCDF4-compatible format. In those cases, `ncdump` revealed that `time_whole = _ ;` which then caused the problem. We now write `time_whole = 0 ;`.

I believe the ten redundant time steps are a separate issue, and an issue with ParaView itself, which I reported here: https://gitlab.kitware.com/paraview/paraview/-/issues/22854.

Screenshot 2025-01-14 at 15 17 17

### DIFF
--- a/include/mesh/exodusII_io.h
+++ b/include/mesh/exodusII_io.h
@@ -314,6 +314,16 @@ public:
                        const std::set<std::string> * system_names=nullptr);
 
   /**
+   * Writes out the solution for no specific time or timestep.
+   * \param fname Name of the file to write to
+   * \param es EquationSystems object which contains the solution vector.
+   * \param system_names Optional list of systems to write solutions for.
+   */
+  virtual void write_equation_systems (const std::string & fname,
+                                       const EquationSystems & es,
+                                       const std::set<std::string> * system_names=nullptr) override;
+
+  /**
    * Write elemsets stored on the Mesh to file.
    *
    * \note An elemset is a concept which is related to (but distinct
--- a/src/mesh/exodusII_io.C
+++ b/src/mesh/exodusII_io.C
@@ -2001,12 +2001,20 @@ void ExodusII_IO::write_timestep (const std::string & fname,
                                   const std::set<std::string> * system_names)
 {
   _timestep = timestep;
-  write_equation_systems(fname,es,system_names);
+  MeshOutput<MeshBase>::write_equation_systems(fname,es,system_names);
 
   if (MeshOutput<MeshBase>::mesh().processor_id())
     return;
 
   exio_helper->write_timestep(timestep, time);
+}
+
+
+void ExodusII_IO::write_equation_systems (const std::string & fname,
+                                          const EquationSystems & es,
+                                          const std::set<std::string> * system_names)
+{
+  write_timestep(fname, es, 1, 0, system_names);
 }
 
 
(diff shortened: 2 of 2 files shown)


include/mesh/exodusII_io.h
```diff
@@ -313,6 +313,16 @@ class ExodusII_IO : public MeshInput<MeshBase>,
...
 
+  /**
+   * Writes out the solution for no specific time or timestep.
+   * \param fname Name of the file to write to
+   * \param es EquationSystems object which contains the solution vector.
+   * \param system_names Optional list of systems to write solutions for.
+   */
+  virtual void write_equation_systems (const std::string & fname,
+                                       const EquationSystems & es,
+                                       const std::set<std::string> * system_names=nullptr);
```
jwpeterson, line 324: This should be marked `override`, there are some compilers that warn when that's missing IIRC.

jwpeterson (commented):
I'm pretty sure I have seen that `time_whole = _;` many times, and although it looked unusual in the ncdump output, I wasn't aware that it also displayed weirdly in Paraview.

roystgnr (approved):
Looks good to me, once the `override` is fixed.

I wish we had some test coverage, but it wouldn't really fit into either our unit tests or examples scaffolding.  At some point in the future we need a "test scripts" directory to do things like tests with our mini-apps, and maybe we could squeeze some `if $have_hdf5; then ncdump foo.e | grep etc` in with that.

include/mesh/exodusII_io.h
nmnobre: Quite right. :)

nmnobre:
> I'm pretty sure I have seen that `time_whole = _;` many times, and although it looked unusual in the ncdump output, I wasn't aware that it also displayed weirdly in Paraview.

Do you mind testing if you also see the weird timestamp without the fix?
I'm pretty sure this hasn't always been a problem, maybe since some HDF5 update in PETSc, that's what I suspect.

jwpeterson:
> Do you mind testing if you also see the weird timestamp without the fix?

I searched through a few Exodus files that I happened to have laying around at the moment, and could only find well-defined values of `time_whole`, e.g. `time_whole = 0 ;` If you can attach a "bad" Exodus file on this PR I could verify that it also shows the same timestamp in whatever version of Paraview I currently have installed?

nmnobre:
> If you can attach a "bad" Exodus file on this PR I could verify that it also shows the same timestamp in whatever version of Paraview I currently have installed?

Here's the output of vector fe ex1: [bad.zip](https://github.com/user-attachments/files/18493225/bad.zip). If you run that example with your current installation, is `time_whole` also ill-defined?

jwpeterson:
> If you run that example with your current installation, is time_whole also ill-defined?

Yes, I can confirmed that I get the undefined `time_whole` for that example. 

And here is how it is displayed in Paraview 5.11.1:
bad_time_whole

I have a self-compiled (not from PETSc) `hdf5-1.10.0` installation, so I don't think it's anything related to recent HDF5 versions (or newer Paraviews).

nmnobre:
> I believe the ten redundant time steps are a separate issue, and an issue with ParaView itself, which I reported here: https://gitlab.kitware.com/paraview/paraview/-/issues/22854.

For completeness: I've been told and confirmed this has been fixed in newer (nightly) versions of ParaView.

PR #4056, opened 2025-01-21
Skip misc_ex17 when PETSc is not enabled

jwpeterson:
FYI @lindsayad something like this is necessary to fix the libmesh devel -> master merge. I think the easiest thing is to skip the example when PETSc is not available.


roystgnr (approved):

lindsayad:
man that's probably been failing for a while. Sorry guys. Feel free to DM me when stuff like this happens

PR #4061, opened 2025-01-27
Add Raviart-Thomas grad-div example

nmnobre:
The (real) motivation is that of other PRs, to be opened soon.
However, since this example is self-contained and should have some value in and of itself, I thought I'd alleviate those other PRs of these changes (including the always massive re-bootstrap and new html).


examples/vector_fe/vector_fe_ex10/vector_fe_ex10.C
```diff
@@ -0,0 +1,480 @@
...
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+
+// <h1>Vector Finite Elements Example 10 - Raviart-Thomas elements (grad-div)</h1>
+// \author Nuno Nobre
+// \date 2025
+//
+// This example uses Raviart-Thomas elements to solve a model grad-div problem
+// in H(div) in both 2d and 3d: -\nabla (\nabla \cdot \vec{u}) + \vec{u} = f.
+
```
jwpeterson, line 25: Would you mind adding some comments about how this example compares/contrasts with vector_fe_ex6? Obviously they are solving different PDEs, but e.g. ex6 uses a "mixed" (u,p) formulation, whereas in this case we don't do that. I guess this information is fairly basic, but since the examples are meant for explanation/teaching, I think it would be appropriate to include additional information.

nmnobre: Done?

jwpeterson: Yeah, that's an improvement... I was also just wondering what practical applications are related the grad-div problem? Obviously div-grad (aka Poisson) has many applications in physics so we don't need to elaborate on them, but I have never run into grad-div before. If it's just a standard/convenient test for RT elements, that's fine too of course.

jwpeterson (approved):

roystgnr (approved):

jwpeterson:
This already passed before the comment update so I'm just going to merge without waiting for the latest round of testing to finish.

PR #4065, opened 2025-01-29
Update _max_abs_value_in_training_set after normalizing snapshots

dknez:
_max_abs_value_in_training_set is used in the EIM greedy algorithm, and it needs to be updated if we normalize the solution snapshots.

### DIFF
--- a/src/reduced_basis/rb_eim_construction.C
+++ b/src/reduced_basis/rb_eim_construction.C
@@ -640,9 +640,10 @@ void RBEIMConstruction::apply_normalization_to_solution_snapshots()
 
   for (unsigned int i=0; i<n_snapshots; i++)
     {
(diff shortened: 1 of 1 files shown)


jwpeterson:
We are seeing the following error message on some of the CI machines:
```
[11:39:45] Executing sandbox-exec

[11:39:45] Not creating Conda environment; already available
/public Conda channel not available.

[11:39:45] Execution in sandbox-exec completed with return code 1
```
@loganharbour or @milljm any ideas? I won't invalidate unless you guys think we should...

jwpeterson (approved):

loganharbour:
Jason was screwing around with stuff. Should be good now...

milljm:
me mucking around with Civet
image

PR #4068, opened 2025-02-03
Refactor FEInterface macros

nmnobre:
Hi @jwpeterson,

This again has bigger motivations, but I think it can stand on its own:

1) Make sure CI is happy
2) Fixes a few of oversights:

`CLOUGH` is missing from:
https://github.com/libMesh/libmesh/blob/7e3b4e3a24c8374654927f7190a2ff1ca5898283/src/fe/fe_interface.C#L229-L266

`L2_HIERARCHIC_VEC` is missing from:
https://github.com/libMesh/libmesh/blob/7e3b4e3a24c8374654927f7190a2ff1ca5898283/src/fe/fe_interface.C#L349-L387

The first error message has a typo, `shape` -> `func_and_args`, in:
https://github.com/libMesh/libmesh/blob/7e3b4e3a24c8374654927f7190a2ff1ca5898283/src/fe/fe_interface.C#L229-L266

### DIFF
--- a/src/fe/fe_interface.C
+++ b/src/fe/fe_interface.C
@@ -70,229 +70,94 @@ FEInterface::is_InfFE_elem(const ElemType et)
     case INFHEX18:
     case INFPRISM6:
     case INFPRISM12:
-      {
-        return true;
-      }
+      return true;
 
     default:
-      {
-        return false;
-      }
+      return false;
     }
 }
 
 #endif //ifndef LIBMESH_ENABLE_INFINITE_ELEMENTS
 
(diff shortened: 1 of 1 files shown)


jwpeterson (approved):
This looks like a good change to me but I think @roystgnr originally wrote these big switch statement macros (?), so I'll let him comment/review before merging.

roystgnr:
I'm kind of scared that there's some sort of `--disable-foo` case we just broke, but I'm happy waiting for the devel->master CI to find out.  I want to get that TIMPI fix downstream ASAP and we might as well get everything outstanding into master at the same time.

PR #4071, opened 2025-02-04
Refactor nodal_soln code

nmnobre:
Another round. :)

### DIFF
--- a/src/numerics/petsc_preconditioner.C
+++ b/src/numerics/petsc_preconditioner.C
@@ -136,7 +136,7 @@ void PetscPreconditioner<T>::set_petsc_preconditioner_type (const Preconditioner
 
     default:
       libMesh::err << "ERROR:  Unsupported PETSC Preconditioner: "
-                   << preconditioner_type       << std::endl
+                   << Utility::enum_to_string(preconditioner_type) << std::endl
                    << "Continuing with PETSC defaults" << std::endl;
     }
 


roystgnr:
Gah.  Any chance I could talk you into getting back to this one in a couple weeks?  I can already see it's going to have merge conflicts with a huge branch of mine.  You've got priority, though, both in the timing sense and in the "removing hundreds of lines of code beats adding thousands" sense.

roystgnr:
I take it back - you're hitting a bunch of the same files, but in different enough places to avoid conflict.

nmnobre:
I noticed we use `enum_to_string()` in `set_petsc_solver_type` but not in `set_petsc_preconditioner_type`, any reason?
Granted the former is missing a lot more solvers than the latter is missing preconditioners, so `enum_to_string()` is more likely to be useful for the former than the latter, is that why?

roystgnr (approved):

roystgnr:
From the git logs, it looks like I added the `enum_to_string` while futzing with `--enable-petsc --enable-mpi` configurations, and my best guess is that I always used PETSc preconditioning options via the command line but sometimes used solver options programmatically and so I was only paying attention to the latter.

nmnobre:
> and so I was only paying attention to the latter

Right, so I think I'll change it? I guess if we are only expecting the user to pass existing/enumerated solvers/preconditioners, it makes sense to switch to `enum_to_string`. For `set_petsc_preconditioner_type` that only applies for `SSOR_PRECOND`, but still... _consistency_... I suppose. The bigger problem comes if the user just passes a wild integer in, for which the code will crash, but that's probably unintended anyway.

roystgnr:
> Right, so I think I'll change it?

Sounds good.  Separate PR?  If @jwpeterson doesn't chime in with objections we can merge this as-is soon.

> I guess if we are only expecting the user to pass existing/enumerated solvers/preconditioners, it makes sense to switch to enum_to_string.

We're definitely expecting that.  If anyone casts a non-enumerated int to an enum to try to evade C++ type-checking, they deserve whatever's coming to them.  :laughing:

nmnobre:
I squeezed it in, it's just a line.

PR #4072, opened 2025-02-06
Added error when stitching fails

DanielYankura:
In no matching nodes are found, libmesh throws an error now

refs idaholab/moose#27758

### DIFF
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -2164,6 +2164,7 @@ UnstructuredMesh::stitching_helper (const MeshBase * other_mesh,
               node_to_elems_map.clear();
             }
         }
+      libmesh_error_msg_if(node_to_node_map.empty(), "Error: Stitch failed, no matching nodes found. Ensure boundaries are properly aligned before trying again.");
     }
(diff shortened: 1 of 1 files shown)


roystgnr:
I like the idea, but it looks like we're interfering here with higher-level error handling in MOOSE CI.

roystgnr:
We could probably add a try/catch there to be able to add compatibility with this with a smooth transition?

jwpeterson:
I would definitely need to test this before we could merge. I think there may be some situations where we rely on the current behavior of not throwing an error when nothing is stitched.

DanielYankura:
I'll go ahead and try adding a try/catch. Also I know users often ignore warnings but if there are cases where it relies on it not throwing an error then maybe making it a warning instead might work?

roystgnr:
I could be talked into a warning.  Not literally `libmesh_warning()`, but something informative to `libMesh::out`.

DanielYankura:
Sorry this took so long, but I changed it to a warning statement (using libmesh::out). I was going to do a try/catch but I think that might still interfere with MOOSE's error handling?

PR #4073, opened 2025-02-07
Add transient_ex3 example demonstrating explicit DG/FV formulation of 2D advection equation

jwpeterson:
This example is based on code originally written by @dknez for a class on numerical methods. We do have a DG example (misc_ex5) but it is for an elliptic equation and may or may not support "finite volume" mode (i.e. running with `(CONSTANT, MONOMIAL)` FEs), but the present example does allow this. This example also lets us demonstrate one way of implementing an explicit timestepping discretization in libmesh, which I'm not sure we have currently. The example runs very fast and gives consistent output for me in both serial and parallel, so it will not add appreciably to test times and shouldn't give us a bunch of test failures.


jwpeterson:
The failure in the "Test MOOSE ARM mac" recipe:
```
[0.033s]     FAIL meshgenerators/break_mesh_by_block_generator.hanging_nodes FAILED (CRASH) [min_cpus=12]
```
 is unrelated to this new test.

examples/transient/transient_ex3/advection_system.C
```diff
@@ -0,0 +1,137 @@
...
+    _fe_order(CONSTANT),
+    _fe_family(MONOMIAL)
+{
+  // Allocate NumericVectors in _Fh. I could not figure out
+  // how to do this in the initialization list, I don't think it's
+  // possible.
+  // TODO: the number of problem dimensions is hard-coded here, we should
+  // make this depend on the input file parameters instead.
+  for (unsigned int i=0; i<2; ++i)
+    _Fh.push_back(NumericVector<Number>::build(es.comm()));
```
roystgnr, line 46: Would it make sense to postpone this until init_data()?

examples/transient/transient_ex3/claw_system.C
```diff
@@ -0,0 +1,971 @@
...
+          fe_elem_face->reinit(elem, side);
+
+          // Find their locations on the neighbor
+          std::vector<Point> qface_neighbor_points;
+          FEInterface::inverse_map(
+            elem->dim(),
+            fe_type,
+            neighbor,
+            qface_points,
+            qface_neighbor_points);
```
roystgnr, line 507: Could we switch this to FEMap::inverse_map(), in preparation for #4074?

examples/transient/transient_ex3/claw_system.C
```diff
@@ -0,0 +1,971 @@
...
+          fe_elem_face->reinit(elem, side);
+
+          // Find side QP locations in neighbor side reference space
+          std::vector<Point> qface_neighbor_points;
+          FEInterface::inverse_map(
+            elem->dim(),
+            fe_type,
+            neighbor,
+            qface_points,
+            qface_neighbor_points);
```
roystgnr, line 676: Likewise for FEMap::inverse_map here

roystgnr (approved):
I can fix up the to-be-deprecated calls after rebasing #4074 later if needed.

examples/transient/transient_ex3/claw_system.C
jwpeterson: Updated in [9e2be1a](https://github.com/libMesh/libmesh/pull/4073/commits/9e2be1a9f044314b2fed387eb57f7762a17a2b58)

examples/transient/transient_ex3/claw_system.C
jwpeterson: Updated in [9e2be1a](https://github.com/libMesh/libmesh/pull/4073/commits/9e2be1a9f044314b2fed387eb57f7762a17a2b58)

examples/transient/transient_ex3/advection_system.C
jwpeterson: Yeah, I think that makes sense. I did the update in [05ef170](https://github.com/libMesh/libmesh/pull/4073/commits/05ef17074d7135c2b43ab1e687b2973192456328) and likewise for the extra `SparseMatrices` in 
24a5ac0.

jwpeterson:
The MOOSE compilation failures here are due to idaholab/moose@08c5990288ae1f444af3ad534f41150629bb92d8, and don't have anything to do with this PR.

PR #4076, opened 2025-02-11
Quadrature API changes

roystgnr:
This strips out just the quadrature-related changes that #4074 depends on, as per @jwpeterson 's request for easier testing and review.

A couple of those commits had Polygon code leak in; that should be fixed here via interactive rebases but it does mean that I'll need to do a little of the same work in #4074 afterward.

### DIFF
--- a/src/fe/inf_fe_boundary.C
+++ b/src/fe/inf_fe_boundary.C
@@ -78,13 +78,13 @@ void InfFE<Dim,T_radial,T_base>::reinit(const Elem * inf_elem,
       if (s > 0)
         {
           current_fe_type.radial_order = fe_type.radial_order;
-          radial_qrule->init(EDGE2, inf_elem->p_level(), true);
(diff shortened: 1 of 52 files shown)


src/fe/inf_fe_boundary.C
```diff
@@ -78,21 +78,21 @@ void InfFE<Dim,T_radial,T_base>::reinit(const Elem * inf_elem,
       if (s > 0)
         {
           current_fe_type.radial_order = fe_type.radial_order;
-          radial_qrule->init(EDGE2, inf_elem->p_level());
+          radial_qrule->init(EDGE2, inf_elem->p_level(), true);
```
jwpeterson, line 81: I'd prefer it if we had an inline comment for the new hard-coded boolean flags, e.g.
```
radial_qrule->init(EDGE2, inf_elem->p_level(), /*simple_type_only=*/true);
```
but that's a minor comment.

jwpeterson (approved):
Looks good to me assuming you can figure out the transition path for MOOSE codes that use this API.

PR #4079, opened 2025-02-12
Precondition constraint matrix

roystgnr:
I'm hoping to add more sophisticated options soon, but this simple diagonal preconditioning seems to be extremely helpful for some Coreform mesh+extraction-operator combinations.

We're not going to be able to merge this until https://github.com/idaholab/moose/pull/29861 hits MOOSE devel - I really want to be sure that it doesn't break any of the test cases there in any configuration.

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -1723,13 +1723,19 @@ public:
    * existing node, a new NodeElem will be added to the mesh on which
    * to store the new unconstrained degree(s) of freedom.
    *
+   * If \p precondition_constraint_operator is true, then the values
+   * of those new unconstrained degrees of freedom may be scaled to
+   * improve the conditioning of typical PDE matrices integrated on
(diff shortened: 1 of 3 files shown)


roystgnr:
Oh, no.  It doesn't break tests, but of course changing the scaling of spline node values creates an exodiff for the test that outputs spline node values.  I'm going to need to make preconditioning non-default here for backwards compatibility and then turn it on at the MOOSE level.

roystgnr:
I keep seeing a crash in meshgenerators/break_mesh_by_block_generator.hanging_nodes in "Test MOOSE ARM mac".  There's no way this PR caused it, but it's disturbing that it's repeating.

roystgnr:
It's happening in #4081 too, at least.

roystgnr:
I'm a little worried that we had some earlier PR trigger that ARM Mac error, but it's not this one and I want to get this downstream ASAP.

PR #4080, opened 2025-02-13
Add thread safety for when multiple threads are failing point inversion

lindsayad:
Refs a failure that briefly cropped up on idaholab/moose#29700 and that I can get to trigger 100% of the time with a build with thread sanitizer


roystgnr (approved):

jwpeterson (approved):
I will trust your expertise and testing on this one, don't know much about `thread_local`. 

But, this PR also makes me think that there are probably a lot of places where we have carelessly used a `static bool` variable in the same way that need to be updated.

PR #4081, opened 2025-02-13
Wrap copied pids based on n_proc, not n_part

### LINKED ISSUE #1512: UnstructuredMesh::copy_nodes_and_elements() requires matching _n_parts

roystgnr:
This fixes a bug where, when copying a distributed mesh into another distributed mesh that's too small to use all its processors, the latter mesh is left in an invalid state with element owners that can't see all the elements they own.

This should probably be the PR that finally closes #1512


roystgnr:
This is a real bug.  Not a new bug, but a bug in the `all_increased_order_range` helper function that's being newly exposed.  Working on a fix now.

roystgnr:
Rebased after the #4089 merge, so hopefully this won't trigger failures in the element order tests anymore.

PR #4084, opened 2025-02-19
Fix combination of mesh constraints with subdomain-restricted variables

roystgnr:
We put new constraining nodes into a new subdomain id as they're created, but then if they're being used to constrain degrees of freedom from variables with subdomain restrictions those variables weren't being properly given the correct unconstrained DoFs.

This patch fixes that, at the cost of changing internal DofMap APIs that really shouldn't be public rather than protected; if we're lucky enough not to break anybody's code then the next step ought to be better encapsulation to avoid that risk in the future.

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -1835,6 +1835,23 @@ private:
                                 dofobject_accessor objects);
 
   /**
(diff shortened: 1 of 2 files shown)


roystgnr:
Test failure unrelated

PR #4085, opened 2025-02-20
MeshBase::get_info(): Print all Orders using Utility::enum_to_string()

jwpeterson:
Previously, we cast the final entry to an unsigned int before printing, which gave inconsistent looking results like the following:
```
  Mesh Information:
   elem_dimensions()={1, 3}
   elem_default_orders()={FIRST, 2}
```


jwpeterson:
The Test MOOSE ARM mac failure is not related:
```
meshgenerators/break_mesh_by_block_generator.hanging_nodes: 
There are not enough slots available in the system to satisfy the 12
slots that were requested by the application:
```

PR #4089, opened 2025-02-25
Don't renumber_nodes_and_elements to change order

roystgnr:
This might have been necessary with the original version of these methods, which probably predate e.g. the introduction of `max_elem_id()` to query on meshes where that might differ from `n_elem()`, but it shouldn't be necessary now, and it can *break* the cases where we construct a range prior to doing renumbering and thereby invalidate the range (e.g. moving valid elements to before the range beginning)

This fixes the errors triggered by #4081 for me.


jwpeterson (approved):
We definitely use `all_second_order_range()` (which calls `all_increased_order_range()`) but I don't think we're doing it for an unprepared mesh... I'm good with merging but will probably double check that this doesn't break anything for us.

PR #4090, opened 2025-02-25
Fix range estimation bug

### LINKED ISSUE #4087: FParser constant folding issue
There is a problem in FParser related to constant folding (in conjunction with exponentials and potential precision loss), which will transform this expression
```
1 - (1.81e+16 * (2.71828 ^ (-1000 * abs(c)))) ^ -1
```

to 

```
1
```

The magnitude of the pre-exponential factor matters. Setting it to `1.80e+16` results in the correct folding.

Subexpressions like this can easily emerge from common "smooth interface expressions like:

```
    expression = 'r := ((x-${x_coord})^2 + (y-${y_coord})^2)^0.5;
                  int_factor := exp(-8*(r-${radius_wear})/${interface_width});
                  1- (1/(1 + int_factor))'
```

For small length scales where the interface width `${interface_width}` is a small value compared to the radius `${radius_wear}` this leads to factoring out a potentially large exponential, leading to the (potential) precision loss issue above.

dschwen:
Closes #4087

We should update from 4.5.1 to 4.5.2 (that release from 2015 includes an optimizer fix that I patched in and reported and several other regressions. In fact 4.5.2 might fix this range estimation bug, too. At first glance it will avoid the issue by not estimating the range at all).

### DIFF
--- a/contrib/fparser/examples/optimizer_bug.cc
+++ b/contrib/fparser/examples/optimizer_bug.cc
@@ -2,7 +2,8 @@
 // (c) 2014 by Daniel Schwen
 // ========================================================
 
-// $CXX -o optimizer_bug optimizer_bug.cc -I ../../../installed/include/libmesh/ ../../../build/contrib/fparser/.libs/libfparser_la-fp* 
+// $CXX -o optimizer_bug optimizer_bug.cc -I ../../../installed/include/
+// ../../../build/contrib/fparser/.libs/libdbg.a
 
 #include "../fparser.hh"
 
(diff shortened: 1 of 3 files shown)


dschwen:
> So, are we not forked from fparser upstream? Just curious how it would go if we did try to update.

Painfully. I noticed that I did the last update myself over 10 years ago. I attempted the update last week and ran into some issues building the release `fpoptimizer.cc` (which we ship). The fparser makefile produces a compressed version, but we skip the compression. I don't remember the other changes to the build process. Anypoop, I'll give it another shot later this week or next week. I did prepare a patchset by comparing upstream fparser 4.5.1 to what we have in libmesh. But there will be manual work required to make sure everything applies.

PR #4077, opened 2025-02-11
FE API deprecation/changes

roystgnr:
As with #4076, this is a fraction of the changes in #4074, split by subsystem into its own PR for easier testing and review.

PR #4091, opened 2025-02-26
Add APIs for name prefixing

lindsayad:
Refs https://github.com/idaholab/moose/issues/29951

### DIFF
--- a/include/systems/system.h
+++ b/include/systems/system.h
@@ -1927,6 +1927,22 @@ public:
    */
   void prefer_hash_table_matrix_assembly(bool preference);
 
+  /**
+   * Instructs this system to prefix solve options with its name for solvers that leverage prefixes
+   */
+  void prefix_with_name(bool value) { _prefix_with_name = value; }
+
+  /**
+   * @returns Whether we are name prefixing
+   */
+  [[nodiscard]] bool prefix_with_name() const { return _prefix_with_name; }
+
+  /**
+   * @returns A prefix that may be applied to solver options. Note that this
+   * prefix is only used if \p prefix_with_name()
+   */
+  std::string prefix() const { return this->name() + "_"; }
+
 protected:
 
   /**
@@ -2307,6 +2323,11 @@ private:
    * Whether any of our matrices require an initial sparsity pattern computation in order to determine preallocation
    */
   bool _require_sparsity_pattern;
+
+  /**
+   * Whether we are name prefixing solver options
+   */
+  bool _prefix_with_name;
 };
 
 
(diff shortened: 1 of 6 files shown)


lindsayad:
If folks are good with this as-is, I'd love to slip this into https://github.com/idaholab/moose/pull/29817, but if we want significant revisions then I definitely don't want this to hold that up

include/systems/system.h
```diff
@@ -1927,6 +1927,22 @@ class System : public ReferenceCountedObject<System>,
...
 
+  /**
+   * Instructs this system to prefix solve options with its name for solvers that leverage prefixes
+   */
+  void prefix_with_name(bool value) { _prefix_with_name = value; }
+
+  /**
+   * @returns Whether we are name prefixing
+   */
+  [[nodiscard]] bool prefix_with_name() const { return _prefix_with_name; }
```
jwpeterson, line 1938: This is the first/only place we use the `[[nodiscard]]` attribute in the library, should we add a configure test for it? I just remembered we have all this stuff
```
// Define backwards-compatible fallthrough attribute.  We could
// eventually also add support for other compiler-specific fallthrough
// attributes.
#ifdef LIBMESH_HAVE_CXX17_FALLTHROUGH_ATTRIBUTE
#define libmesh_fallthrough() [[fallthrough]]
#elif defined(LIBMESH_HAVE_DOUBLE_UNDERSCORE_ATTRIBUTE_FALLTHROUGH)
#define libmesh_fallthrough() __attribute__((fallthrough))
#else
#endif
```
for the `[[fallthrough]]` attribute, but these attributes have been standard since C++17 so maybe time to get rid of `libmesh_fallthrough()`

roystgnr (approved):
At some point I need to go through all the `git grep on_command_line` output and fix everything else that has a command line interface but no C++ interface...

include/systems/system.h
jwpeterson: :+1: I `git grepped` for a test in the `m4` directory and didn't find one. I guess that type of grep doesn't descend into submodules.

jwpeterson:
Should we wait until CIVET is closer to 100% before merging stuff?

include/systems/system.h
roystgnr: And it looks like we are running it.  Checking one of my `config.log` files I see the `[[nodiscard]]` function in the `failed program was:` output for a case where my `mpicxx` originally failed before being retried with `-std=gnu++17`.

lindsayad:
We'll need to figure out what to do about that openmpi slot issue at some point for Mac arm

roystgnr:
It should be as simple as telling openmpi's `mpirun` to `--oversubscribe`, shouldn't it?

lindsayad:
Tossed up https://github.com/idaholab/moose/pull/29966

roystgnr:
> I'd love to slip this into https://github.com/idaholab/moose/pull/29817

I'll wait until that's done and merged, then, before we merge #4076 and #4077, just in case we need any fixes to go in first.  Those are big enough and the submodule update delayed enough that I don't want to get them into it by accident lest something cause further delay.

PR #4094, opened 2025-03-03
Add comments with approximate floating point values of magic constants

jwpeterson:
This is also related to a brief discussion I had with @roystgnr where it was pointed out that some of our "L" suffixes are redundant while others aren't.


jwpeterson:
Test failure is not related to PR:
```
[0.029s]     FAIL meshgenerators/break_mesh_by_block_generator.hanging_nodes FAILED (CRASH) [min_cpus=12]
```

roystgnr (approved):

PR #4095, opened 2025-03-03
Fix DistributedMesh::add_point on deleted node id

roystgnr:
If we try to add a new node with an id that once existed but was recently deleted, we might find that the entry for that id still exists as a nullptr on the _nodes container.  Previously this would lead to a failed assertion in dbg mode and a segfault in opt mode.

This fixes a test failure in
https://github.com/idaholab/moose/pull/29953 for me.


roystgnr:
`[0.033s]     FAIL meshgenerators/break_mesh_by_block_generator.hanging_nodes FAILED (CRASH) [min_cpus=12]` is downstream and unrelated

PR #4096, opened 2025-03-04
Avoid creating temporary array for hashword2() overload taking 64-bit inputs

jwpeterson:
I was playing around with this yesterday while trying to optimize some code using a `std::unordered_map`. In that application, `std::map` turned out to be faster so we ended up not needing this, but it might be useful at some point. The speed-up here (if any) comes from not needing to pack the inputs into a separate array just to conform to the `fnv_64_buf()` API, the hash function itself should be exactly the same.

### DIFF
--- a/include/utils/hashword.h
+++ b/include/utils/hashword.h
@@ -132,6 +132,41 @@ uint64_t fnv_64_buf(const void * buf, size_t len)
   return hval;
 }
 
+/**
+ * Same hash as fnv_64_buf(), but specialized for exactly two inputs.
+ * This function is called from the "hashword2()" overload for 64-bit
+ * integers, and lets us avoid constructing a temporary array in order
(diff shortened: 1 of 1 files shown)


roystgnr (approved):
Couldn't hurt.  I wouldn't expect avoiding a copy onto a stack array to be a noticeable optimization, but getting rid of conditional jumps by unrolling now-fixed-length loops might be helpful.  I take it from "if any" that you didn't benchmark, though?

jwpeterson:
> I take it from "if any" that you didn't benchmark, though?

Yeah, we ended up going with a different approach before I actually got a chance to try this hash function. As you pointed out, there's probably not a big savings from avoiding the stack array creation (as I was hoping), but the compiler should be able to better unroll the loops in the new version since it knows the number of iterations at compile-time now.

PR #4098, opened 2025-03-05
Add begin/end members to DenseVector

lindsayad:
Serves me right for not compiling https://github.com/idaholab/moose/pull/29889 in debug mode. This further helps me to use `DenseVector` in templated code

### DIFF
--- a/include/numerics/dense_vector.h
+++ b/include/numerics/dense_vector.h
@@ -282,6 +282,18 @@ public:
    */
   const std::vector<T> & get_values() const { return _val; }
 
+  /**
+   * \returns An iterator pointing to the beginning of the vector
+   */
+  auto begin() const { return _val.begin(); }
+  auto begin() { return _val.begin(); }
+
+  /**
+   * \returns An iterator pointing to the end of the vector
+   */
+   auto end() const { return _val.end(); }
+   auto end() { return _val.end(); }
+
 private:
 
   /**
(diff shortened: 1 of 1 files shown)


lindsayad:
@roystgnr I don't think you've posted the next libmesh update PR, but if you're on board for this PR I'd love to have it included

jwpeterson (commented):
I don't think we use `auto` return types for class APIs anywhere else? I guess I'm fine with it, but I don't know what C++ standards support this officially (maybe ever since C++11)?

roystgnr:
I guess my only real annoyance about the `auto` is it's less self-documenting; figuring out the return type means scrolling down to the definition of `_val`, which could eventually be a long scroll, the way we tend to expand our classes' method lists.

roystgnr:
Alex is AFK until next week (which will be busy), though, and would probably prefer that *some* version of this PR get merged rather than have any argument back-and-forth.  I'll let John make the final call.

jwpeterson (approved):
I think this should be good to go now, we can wait for all the testing again if you want to @roystgnr, but I tested that the compilation works locally so should be safe to merge.

lindsayad:
My contention would be that I personally never write code in which I catch a returned iterator with anything other than auto as I usually don't care if it's an iterator over a vector, deque, array, etc. And if for some reason anyone ever wanted to change the type of `_val`, they'll now have to go change multiple lines of code and perhaps because the code was more self-documenting, users will also have decided to explicitly type the return type instead of using auto, making it more difficult to make non-breaking changes in the future.

But as it's difficult to envision `_val` ever changing from a `std::vector`, I'm not too bothered

PR #4101, opened 2025-03-10
Remove declaration for removed API

roystgnr:
Alex took the definition of this out, reasonably since it's redundant with the local m/n methods, but he forgot to remove the declaration.


roystgnr:
Thanks to @YaqiWang for catching this.

PR #4102, opened 2025-03-11
Clean up new log file

roystgnr:
Our distcheck recipe in CI is complaining, e.g. https://civet.inl.gov/job/2741387/ , about

```
ERROR: files left in build directory after distclean:
./tests/smoother.out
```


jwpeterson (approved):
The `distcheck` target has been failing for longer than this test has been turned back on, so I think there's still something else going on there, but this is obviously needed as well.

PR #4103, opened 2025-03-11
Allow instantiation of print with dual numbers

lindsayad:
New versions of google test appear to make use of `operator <<` which calls `print` for `TypeVector`. We have MOOSE unit tests that instantiate `TypeVector` with dual numbers, e.g. something other than `Real` or `Complex` for which we have explicit instantiations for


roystgnr (approved):

jwpeterson (approved):
Would it make sense to add explicit instantiations of `TypeVector` for `DualNumber` too (when `MetaPhysicL` is available)?

lindsayad:
Dual number is a template with infinitely many instance choices, eg the size of a static derivative container so I don't think that will work

PR #4112, opened 2025-03-20
Add mutex to setEpsilon

### LINKED ISSUE #4111: Make the global FParser epsilon value `thread_local`
The fuzzy comparison epsilon in FParser is a global variable (sigh). In MOOSE we will change it from it's default value a lot more often now (through https://github.com/idaholab/moose/pull/30014). In the course of this @lindsayad came across an thread sanitizer race warning when multiple threads try to set the epsilon value at the same time. This is all kinds of bad, but at the very minimum we should make that change operation atomic. This doesn't address the possibility of different threads setting epsilon to different values, but at least in stock MOOSE that should not happen. 

I'm falling more and more out of love with that library...

dschwen:
Closes #4111

Thanks @lindsayad for pointing me to `thread_local`

### DIFF
--- a/contrib/fparser/fparser.cc
+++ b/contrib/fparser/fparser.cc
@@ -17,13 +17,14 @@
 #include "fpconfig.hh"
 #include "fparser.hh"
 
-#include <set>
-#include <cstdlib>
-#include <cstring>
+#include <cassert>
 #include <cctype>
 #include <cmath>
-#include <cassert>
+#include <cstdlib>
+#include <cstring>
 #include <limits>
+#include <mutex>
+#include <set>
 
 #include "extrasrc/fptypes.hh"
 #include "extrasrc/fpaux.hh"
@@ -931,6 +932,8 @@ Value_t FunctionParserBase<Value_t>::epsilon()
 template<typename Value_t>
 void FunctionParserBase<Value_t>::setEpsilon(Value_t value)
 {
+    static std::mutex epsilon_mutex;
+    std::lock_guard<std::mutex> lock(epsilon_mutex);
     Epsilon<Value_t>::value = value;
 }
 


roystgnr:
This scares me a bit.  If I try to set epsilon to a new value while I'm in a non-threaded section of code, and then I enter into threads, what happens?  My first guess would be "if new threads are being forked then they inherit the new value, but if we're using an underlying thread pool then only one of the threads in the pool has the new value", and if that's the case then that would be horrifying.

If there's already a setter function and we're only ever using that, could we just add a lock in the setter instead?

lindsayad:
excellent point

dschwen:
We'll my first commit just makes epsilon atomic. That seems like the easiest way.

lindsayad (dismissed):

dschwen:
arithmetic operators aren’t overloaded for `std::atomic` types.

:sob:

dschwen:
> This scares me a bit. If I try to set epsilon to a new value while I'm in a non-threaded section of code, and then I enter into threads, what happens? My first guess would be "if new threads are being forked then they inherit the new value, but if we're using an underlying thread pool then only one of the threads in the pool has the new value", and if that's the case then that would be horrifying.

The right answer would be to set epsilon in the thread you expect it to be used (which is what we do in MOOSE). `thread_local` would be the smallest possible change to fix the underlying issue, and probably the best performing...

jwpeterson (approved):
My only question was whether making the std::mutex itself static creates another kind of race condition, but [that appears to be safe to do](https://stackoverflow.com/questions/38860572/does-making-stdmutex-as-static-creates-race-condition-for-the-mutex-itself).

dschwen:
> My only question was whether making the std::mutex itself static creates another kind of race condition, but [that appears to be safe to do](https://stackoverflow.com/questions/38860572/does-making-stdmutex-as-static-creates-race-condition-for-the-mutex-itself).

Love the first reply:

PR #4115, opened 2025-03-24
reset_memory -> restore_original_nonzero_pattern

lindsayad:
I also added a pretty important comment to the doxygen string. This reflects what happens in the PETSc implementation. I now have some concern about the method name ... a name like `reset_memory` makes me think that there's no way something like a no-op could happen. Perhaps a name like `restore_(original_)nonzero_pattern` would be better? Would this still conceptually allow its use for hash table based memory allocation? I think so since resetting/clearing the hash table restores the original sparsity pattern ... that pattern being whatever you want it to be

### DIFF
--- a/examples/miscellaneous/miscellaneous_ex17/miscellaneous_ex17.C
+++ b/examples/miscellaneous/miscellaneous_ex17/miscellaneous_ex17.C
@@ -160,8 +160,8 @@ int main (int argc, char ** argv)
   // MatResetHash added in PETSc version 3.23
 #if !PETSC_VERSION_LESS_THAN(3, 23, 0)
   // reset the memory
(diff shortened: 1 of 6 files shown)


lindsayad:
Will do. I'm converting this to draft because there are upstream changes to PETSc that I need to make for the work here

lindsayad:
The Mac ARM test target is pretty obnoxious

lindsayad:
We are not going to change the API in PETSc so I can remove the WIP draft label from this. Instead we are going to honor the documentation of `MatResetPreallocation`. https://gitlab.com/petsc/petsc/-/merge_requests/8236

roystgnr (approved):

PR #4116, opened 2025-03-24
verify() doesn't actually do assert() by itself.

roystgnr:
I was going to make fun of @lindsayad for making this mistake, but it turns out it's a popular mistake to make.

Pinging @jwpeterson to run the RB code through their tests.

I'll add some distributed sweeps here just in case those accidentally-do-nothing distributed consistency tests actually were missing inconsistencies.

### DIFF
--- a/include/parallel/parallel_algebra.h
+++ b/include/parallel/parallel_algebra.h
@@ -144,11 +144,11 @@ struct TypeVectorAttributes
   static const bool has_min_max = true;
   static void set_lowest(V & x) {
     for (int d=0; d != LIBMESH_DIM; ++d)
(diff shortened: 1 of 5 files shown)


roystgnr:
Well, that's not good.  Those look like actual verification failures, for node coordinate consistency, in three unit tests, two of which are for features that we either heavily use downstream in MOOSE or plan to heavily use downstream soon.  Looking into it.

jwpeterson (approved):

lindsayad (commented):
I'm mildly proud of the productive changes that came out of what I thought was an innocuous question

PR #4117, opened 2025-03-25
Refactor Tet::choose_diagonal()

nmnobre:
Hi,

This still doesn't resolve the main issue, but removes the uncertainty from nodal permutations out of the equation...

Cheers,
-N

### DIFF
--- a/src/geom/cell_tet.C
+++ b/src/geom/cell_tet.C
@@ -203,23 +203,17 @@ bool Tet::is_child_on_side_helper(const unsigned int /*c*/,
 
 void Tet::choose_diagonal() const
 {
-  // Check for uninitialized diagonal selection
+  // If uninitialized diagonal selection, select the shortest octahedron diagonal
   if (this->_diagonal_selection==INVALID_DIAG)
     {
       Real diag_01_23 = (this->point(0) + this->point(1) - this->point(2) - this->point(3)).norm_sq();
       Real diag_02_13 = (this->point(0) - this->point(1) + this->point(2) - this->point(3)).norm_sq();
       Real diag_03_12 = (this->point(0) - this->point(1) - this->point(2) + this->point(3)).norm_sq();
 
-      this->_diagonal_selection=DIAG_02_13;
+      std::array<Real, 3> D = {diag_02_13, diag_03_12, diag_01_23};
 
-      if (diag_01_23 < diag_02_13 || diag_03_12 < diag_02_13)
-        {
-          if (diag_01_23 < diag_03_12)
-            this->_diagonal_selection=DIAG_01_23;
-
-          else
-            this->_diagonal_selection=DIAG_03_12;
-        }
+      this->_diagonal_selection =
+        Diagonal(std::distance(D.begin(), std::min_element(D.begin(), D.end())));
     }
 }
 


src/geom/cell_tet.C
```diff
@@ -206,20 +206,27 @@ void Tet::choose_diagonal() const
...
-      if (diag_01_23 < diag_02_13 || diag_03_12 < diag_02_13)
-        {
-          if (diag_01_23 < diag_03_12)
-            this->_diagonal_selection=DIAG_01_23;
-
-          else
-            this->_diagonal_selection=DIAG_03_12;
-        }
+      auto M = [&](const unsigned i, const unsigned j) -> Point
+               { return this->point(i) + this->point(j); };
```
jwpeterson, line 210: Is this missing a 1/2? Assuming you are computing midpoints here...

src/geom/cell_tet.C
```diff
@@ -206,20 +206,27 @@ void Tet::choose_diagonal() const
...
+
+      Real diag_02_13 = (P[0] - P[3]).norm_sq();
+      Real diag_03_12 = (P[1] - P[4]).norm_sq();
+      Real diag_01_23 = (P[2] - P[5]).norm_sq();
+
+      std::array<Real, 3> D = {diag_02_13, diag_03_12, diag_01_23};
+
+      // If all diagonals are of ~equal length, we choose the one with
+      // the lexicographically smallest endpoint to break the tie
+      if (abs(D[0] - D[1]) < TOLERANCE * TOLERANCE && abs(D[0] - D[2]) < TOLERANCE * TOLERANCE)
```
jwpeterson, line 223: `abs` -> `std::abs`

src/geom/cell_tet.C
nmnobre: We are. And we were before too. But we avoid the division by two for performance I imagine. Because that doesn't change the order.

src/geom/cell_tet.C
nmnobre: Done.

nmnobre:
Actually, John, if you wouldn't mind. Can you test applying the following changes to `introduction_ex4.C`, running with say, `./example-opt -d 3 -n 2 -o FIRST` and looking at the result in Paraview? Do you see weird artifacts on the sides of the cube which are supposed to be identically zero?

```
53c53,54
< 
---
> #include "libmesh/mesh_refinement.h"
> #include "libmesh/mesh_modification.h"
197c198
<                                          ((dim == 2) ? QUAD4 : HEX8));
---
>                                          ((dim == 2) ? QUAD4 : TET10));
213c214,216
< 
---
>   MeshTools::Modification::permute_elements(mesh);
>   MeshRefinement mesh_refinement (mesh);
>   mesh_refinement.uniformly_refine(1);
```

nmnobre:
Right, I can't really decide if this useful or not. `permute_elements()` is clearly a debugging/testing utility only (and in fairness, this is clearly stated in its docstring). So, can it even happen that the elements in the same mesh are locally numbered differently? If the user manually permutes their mesh before it's ingested, sure, but that's unlikely - though the user would probably still expect the same results in that case. Are there any other circumstances?

jwpeterson:
I guess you deleted the comment (?), but I ran intro_ex4 with your suggested modifications and the solution looks like this:
ex4

> Do you see weird artifacts on the sides of the cube which are supposed to be identically zero?
Yeah, I think I see some non-zero values on the sides... is that what you were referring to?

nmnobre:
> I guess you deleted the comment (?)

I did.

> Yeah, I think I see some non-zero values on the sides... is that what you were referring to?

It was.

But this is not a good test because `permute_elements()` does nothing to fix boundary ids after the permutation.
And if you disable the permutations, everything looks good. Thus my recent comment about the possible uselessness of this PR.

jwpeterson:
> Thus my recent comment about the possible uselessness of this PR.

What was your motivation in creating the PR? Just because we can't easily recreate the scenario in a unit test doesn't mean that the change is useless...

>> I guess you deleted the comment (?)

> I did.

OK, well it would be easier for me (and anyone viewing this PR in the future!) to follow along with what's going on if you leave the comment history intact (maybe use strikethrough text or something) and avoid force-pushing the branch while it's under discussion.

nmnobre:
> What was your motivation in creating the PR? Just because we can't easily recreate the scenario in a unit test doesn't mean that the change is useless...

In a sentence: I'm not sure our tetrahedral refinement is trustworthy. I started playing with vector fe ex4, and noticed that the solution would be all weird with `uniform_refine(2)` and started digging... @roystgnr suggested we tried permuting elements to see if you'd still get the same results, but we don't, even with `uniform_refine(0)`, because of the way we select the diagonal (in case multiple diagonals are the same length, it's a coin flip as to whether which gets selected - this PR removes this uncertainty).

> OK, well it would be easier for me (and anyone viewing this PR in the future!) to follow along with what's going on if you leave the comment history intact (maybe use strikethrough text or something) and avoid force-pushing the branch while it's under discussion.

Yes, you're right, I was hoping you hadn't seen the comment yet, I apologise.

BUT, this PR has already revealed something at least, choosing a different diagonal is causing the MOOSE non-conformal tests to fail - and I have no idea why, because the new nodes are _exactly_ the same. But it feels like progress.

nmnobre:
> But it feels like progress - or worst case scenario, the MOOSE checks are not robust.

Yep, that was MOOSE: idaholab/moose#30196.

roystgnr:
> `permute_elements()` does nothing to fix boundary ids after the permutation.

This is true, but it wouldn't be hard to fix.  We'd need to pass an optional `BoundaryInfo *` into `Elem::permute()`, the way we do into `Elem::flip()`, but I didn't avoid that because it was a bad idea, I just avoided it because it wasn't necessary for the new unit tests I was writing.

I'm not sure this PR is useful (there's always going to be some grid resolution where floating point error works out differently at different places, and instead of getting to a tie we can disambiguate lexicographically, we get to an almost-tie that goes one way or another), but I'm still stubbornly wondering if "permute elements after changing the diagonal selection" can teach us more - we wouldn't want to make a more sweeping change *permanent*, but selecting diagonals based on something like *only* physical point ordering might still help you with debugging.

nmnobre:
> I'm not sure this PR is useful (there's always going to be some grid resolution where floating point error works out differently at different places, and instead of getting to a tie we can disambiguate lexicographically, we get to an almost-tie that goes one way or another),

Yes, of course. The usefulness of this, if any, comes only from pure node reorderings. With this, you get the same mesh, no matter how you count. That's all. For instance:
https://github.com/idaholab/moose/blob/09b3c83c02189c5d90355d400588060dc48d2d21/framework/src/meshgenerators/MeshDiagnosticsGenerator.C#L1272-L1274

> but I'm still stubbornly wondering if "permute elements after changing the diagonal selection" can teach us more

I'm not sure I understand, sorry.

nmnobre:
We've decided spending the extra effort for the tie-break is probably not worth it, so this PR is now just a very minor refactor.

roystgnr (approved):

src/geom/cell_tet.C
nmnobre: Resolving since not applicable anymore.

src/geom/cell_tet.C
nmnobre: Resolving since not applicable anymore.

PR #4119, opened 2025-03-27
Add orientation-dependent caching checks and option to disable caching

nmnobre:
Hi,

It turns out the problem I initially thought was to do with tetrahedral refinement is not specifically related to our refinement procedures, it is just more likely to happen when we refine.

There are cases in which two consecutive elements are sufficiently similar to trigger and reuse cached shape functions when, in fact, the orientation of the second element doesn't match that of the first. For example, two parallel edges aligned with the y-axis on neighbouring elements might not have the same orientation because the second has got a O(1e-16) difference in the x-coordinates of its two vertices for example. This is bound to happen when we refine and the resulting coordinates can't be exactly represented numerically (in base 2).

I've added a case to Vector FE Ex. 3: you expect `grid_size=5 refine=2` to be the same as `grid_size=20` (our default), but unless we disable caching, that isn't the case.

Of course an alternative to this would be, just like we do for caching, to do fuzzy comparisons for orientations as well. Opinions?

Cheers,
Nuno

### DIFF
--- a/include/fe/fe.h
+++ b/include/fe/fe.h
@@ -806,7 +806,7 @@ protected:
    * Check if the node locations, edge and face orientations
    * held in the element cache match those of element \p elem.
    */
-  bool match(const Elem * elem);
+  bool matches_cache(const Elem * elem);
 
   /**
    * The last side and last edge we did a reinit on
--- a/include/fe/fe.h
+++ b/include/fe/fe.h
@@ -790,10 +790,23 @@ protected:
                                       const unsigned vdim = 1);
 
   /**
-   * An array of the node locations on the last
-   * element we computed on
+   * Vectors holding the node locations, edge and
+   * face orientations of the last element we cached.
    */
   std::vector<Point> cached_nodes;
+  std::vector<bool> cached_edges, cached_faces;
+
+  /**
+   * Repopulate the element cache with the node locations,
+   * edge and face orientations of the element \p elem.
+   */
+  void cache(const Elem * elem);
+
+  /**
+   * Check if the node locations, edge and face orientations
+   * held in the element cache match those of element \p elem.
+   */
+  bool match(const Elem * elem);
 
   /**
    * The last side and last edge we did a reinit on
--- a/src/fe/fe.C
+++ b/src/fe/fe.C
@@ -27,6 +27,7 @@
 #include "libmesh/tensor_value.h"
 #include "libmesh/enum_elem_type.h"
 #include "libmesh/quadrature_gauss.h"
+#include "libmesh/libmesh_singleton.h"
 
 namespace {
   // Put this outside a templated class, so we only get 1 warning
@@ -40,6 +41,18 @@ namespace {
 
 namespace libMesh
 {
+// ------------------------------------------------------------
+// Whether we cache the node locations on the last element we computed on
+// to try to avoid calling init_shape_functions and compute_shape_functions
+static const bool * caching = nullptr;
+
+class CachingSetup: public Singleton::Setup
+{
+  private:
+    void setup() { caching = new bool(!on_command_line("--disable-caching")); }
+  public:
+    ~CachingSetup() { delete caching; caching = nullptr; }
+} caching_setup;
 
 
 // ------------------------------------------------------------
@@ -225,7 +238,7 @@ void FE<Dim,T>::reinit(const Elem * elem,
             }
 
           // Replace cached nodes if no longer fitting
-          if (this->shapes_need_reinit() && !cached_nodes_still_fit)
+          if (this->shapes_need_reinit() && !cached_nodes_still_fit && *caching)
             {
               cached_nodes.resize(elem->n_nodes());
               for (auto n : elem->node_index_range())
--- a/src/fe/fe.C
+++ b/src/fe/fe.C
@@ -171,7 +171,7 @@ void FE<Dim,T>::cache(const Elem * elem)
 
 
 template <unsigned int Dim, FEFamily T>
-bool FE<Dim,T>::match(const Elem * elem)
+bool FE<Dim,T>::matches_cache(const Elem * elem)
 {
   bool m = cached_nodes.size() == elem->n_nodes();
   for (unsigned n = 1; m && n < elem->n_nodes(); n++)
@@ -271,7 +271,7 @@ void FE<Dim,T>::reinit(const Elem * elem,
               this->_elem = elem;
 
               // Check if cached element's nodes, edge and face orientations still fit
-              cached_elem_still_fits = this->match(elem);
+              cached_elem_still_fits = this->matches_cache(elem);
 
               // Initialize the shape functions if needed
               if (this->shapes_need_reinit() && !cached_elem_still_fits)
--- a/src/fe/fe.C
+++ b/src/fe/fe.C
@@ -42,8 +42,9 @@ namespace {
 namespace libMesh
 {
 // ------------------------------------------------------------
-// Whether we cache the node locations on the last element we computed on
-// to try to avoid calling init_shape_functions and compute_shape_functions
+// Whether we cache the node locations, edge and face orientations of the last
+// element we computed on as needed to avoid calling init_shape_functions and
+// compute_shape_functions
 static const bool * caching = nullptr;
 
 class CachingSetup: public Singleton::Setup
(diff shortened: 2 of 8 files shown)


roystgnr:
> Of course an alternative to this would be, just like we do for caching, to do fuzzy comparisons for orientations as well.

I don't see how this would work.  There has to be some single point at which an orientation flips; it's not something that can be non-transitive like the similarity we use in caching.  I'd love to find out how I'm wrong, though, particularly if it's a foundational sort of mistake on my part.  At *some* point we need to come up with orientation rules that can work with mesh displacement, and I haven't thought of anything better than "use unique_id() when it's enabled".

My first thought was to start doing non-fuzzy exact comparisons for caching, but that won't work well: a cartesian grid which should cache cleanly would then start hitting false negatives with any rotation or even with most grid spacings.

My only other idea is that we could make the "fuzzy" tolerance controlled by the FE type.  Using a nodal or discontinuous type, it specifies `<= TOLERANCE*TOLERANCE`, and you still get the maximum reasonable caching; using a type that depends on orientations for vector or high-order scalar shapes, it specifies `<= 0`, and you get safety.

nmnobre:
> There has to be some single point at which an orientation flips; it's not something that can be non-transitive like the similarity we use in caching.

So, similarity is a relation between two elements, and I agree it's non-transitive. The question that needs to be asked is: if two elements are similar, can we guarantee they have the same orientations. If two elements are similar, we know vectors (0) -> (m) and (0) -> (n), where m and n are vertices of some edge, in the two elements are similar, i.e. their difference has l1-norm <= tol. It's obviously the case this doesn't guarantee (m) -> (n), i.e. the edge, is similar by the same standard. But, by the triangle inequality, we know it's similar if we relax the constraint to <= 2*tol. So could we not define our lexicographic comparisons with this in mind?

> My first thought was to start doing non-fuzzy exact comparisons for caching, but that won't work well

Agreed.

>  At some point we need to come up with orientation rules that can work with mesh displacement

Since this is a worry, and I agree we'll need to use a unique node id, and since in that case our similarity test is not good enough, i.e. it's no guarantee the orientation is preserved, I'm tempted to go with:

> My only other idea is that we could make the "fuzzy" tolerance controlled by the FE type. Using a nodal or discontinuous type, it specifies `<= TOLERANCE*TOLERANCE`, and you still get the maximum reasonable caching; using a type that depends on orientations for vector or high-order scalar shapes, it specifies `<= 0`, and you get safety.

Do we still keep `--disable-caching` or do I remove that?

roystgnr:
> If we decide to go with this, do we still keep --disable-caching or do I remove that?

I definitely like having an option to disable caching.  It's kind of like the option to use higher than double precision: we should never *expect* to need it, but if we ever *suspect* we need it then it's invaluable to be able to easily test that suspicion, either so we can quickly find and fix the problem or so we can rule out that suspicion.  I guess we would want it to be a `configure` option rather than a command line option if we discovered that the extra runtime test hurt performance, but if we do the runtime option right then I'd be so surprised if someone saw a performance hit from it that I wouldn't trust their numbers without testing it myself.  (I'll add some review comments re: doing it right)

> the second element needs to be similar AND preserve orientation.

This isn't crazy.  It'd be much more productive than the "set tolerance to 0 for orientation-dependent elements" idea.

include/fe/fe.h
```diff
@@ -793,6 +793,13 @@ class FE : public FEGenericBase<typename FEOutputType<T>::type>
                                       std::vector<Number> & nodal_soln_on_side,
                                       bool add_p_level = true);
 
+  /**
+   * Whether we cache the node locations on the last
+   * element we computed on to try to avoid calling
+   * init_shape_functions and compute_shape_functions
+   */
+  bool caching;
```
roystgnr, line 801: This should probably be a static member - even if anyone is doing something as weird as turning caching on for some FE objects and off for others, that means they're deep enough in the weeds that they could figure out how to do that by manipulating the static member.

nmnobre: I've removed this member entirely. I can put it back if you think we should?

src/fe/fe.C
```diff
@@ -47,6 +47,7 @@ namespace libMesh
 template <unsigned int Dim, FEFamily T>
 FE<Dim,T>::FE (const FEType & fet) :
   FEGenericBase<typename FEOutputType<T>::type> (Dim,fet),
+  caching(!libMesh::on_command_line("--disable-caching")),
```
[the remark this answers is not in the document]
nmnobre: I'm now (partially) using the `Singleton` functionality though I actually never create a `Singleton` object as it looks unnecessary here.

nmnobre:
This is ready for another look, it implements what we discussed, it's just a matter of style now I think.

1) Do we want to initialise a static data member in the `FE` classes for the caching toggle?
2) Did I miss any orientation-dependent families in the if conditions? Would we rather define a `is_orientation_dependent()` method similar to `is_hierarchic()`?

src/fe/fe.C
```diff
@@ -202,45 +216,63 @@ void FE<Dim,T>::reinit(const Elem * elem,
...
+              // Check if cached elem geometry is sufficiently similar
+              cached_elem_still_fits = cached_nodes.size() == elem->n_nodes();
+              for (unsigned n = 1; cached_elem_still_fits && n < elem->n_nodes(); n++)
+                cached_elem_still_fits = (elem->point(n) - elem->point(0)).relative_fuzzy_equals
+                                         (cached_nodes[n] - cached_nodes[0]);
+
+              // Check if cached edge and face orientations are equal
+              if constexpr (T == HIERARCHIC || T == L2_HIERARCHIC || T == HIERARCHIC_VEC ||
+                            T == L2_HIERARCHIC_VEC || T == BERNSTEIN || T == SZABAB ||
+                            T == NEDELEC_ONE || T == RAVIART_THOMAS)
```
roystgnr, line 238: Could we move this logic into a static Elem:: array?

roystgnr (approved):
I also think the cache validity checking is getting to a level of complexity that deserves being broken out into its own method, but IMHO this is still a step in the right direction as-is.

nmnobre:
I've benchmarked Vector FE Ex.4 with
`for i in {1..5}; do (time ./example-opt grid_size=20 element_type=HEX20 -pc_type lu > /dev/null) 2>&1 | grep real; done`, and saw no performance regression:
Before:
```
real	0m20.275s
real	0m20.247s
real	0m20.344s
real	0m20.391s
real	0m20.318s
```
After:
```
real	0m20.003s
real	0m20.288s
real	0m20.424s
real	0m20.201s
real	0m20.380s
```
After w/ `--disable-caching`:
```
real	0m21.396s
real	0m21.389s
real	0m21.371s
real	0m21.446s
real	0m21.565s
```

src/fe/fe.C
nmnobre: >I also think the cache validity checking is getting to a level of complexity that deserves being broken out into its own method

Done.

> Could we move this logic into a static Elem:: array?

Moved to an `FEInterface` member function instead as suggested in (2) above.

roystgnr (changes requested):

src/fe/fe.C
```diff
@@ -135,6 +149,50 @@ void FE<Dim,T>::dofs_on_edge(const Elem * const elem,
...
+}
+
+
+
+template <unsigned int Dim, FEFamily T>
+bool FE<Dim,T>::match(const Elem * elem)
+{
+  bool m = cached_nodes.size() == elem->n_nodes();
+  for (unsigned n = 1; m && n < elem->n_nodes(); n++)
+    m = (elem->point(n) - elem->point(0)).relative_fuzzy_equals(cached_nodes[n] - cached_nodes[0]);
```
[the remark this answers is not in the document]
nmnobre: > For correctness: shouldn't this be `&=` rather than `=`? Likewise in the for loops below?

We know `m` is true in the loop bodies, so the assignment is equivalent.

> And if that's right, then, for performance: should we bother tracking `m` to return it? If only 99% of cached geometry and orientation matches then we can't reuse cached data, so we might as well `return false` as soon as we see any mismatch or `return true` at the end otherwise.

Yeah, I thought about that, but I really liked the symmetry with the code in the sister method. Besides, the loops short-circuit if `m` is false. For orientation-independent elements this means the code should be optimal (right?), for orientation-dependent elements, I can check `m` in the if-statement if you think it's needed.

> And either way, if I'm right and we don't want to merge this without changes: how about `matches_cache()` for the name?

I can change the name anyway. :)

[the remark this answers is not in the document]
roystgnr: Oh, man, I missed the short-circuiting in the for loop parentheses.  This is a really weird idiom but yeah, looks correct, and it's basically optimal - if you break out of the first for loop you're still requesting 2 unnecessary tests of m, but I'll bet the optimizer figures that out and gives the same code regardless, and either way you're not doing any unnecessary floating-point.

Damn it, I knew that if I thought I caught you in a correctness error then I must have overlooked something.

nmnobre:
> I'd prefer the name change but I'm happy either way.

Done.

PR #4120, opened 2025-03-27
Remove some no longer necessary changes for PETSc/hypre memory location

lindsayad:
This is now handled upstream in PETSc

### DIFF
--- a/src/solvers/petsc_linear_solver.C
+++ b/src/solvers/petsc_linear_solver.C
@@ -36,10 +36,6 @@
 #include "libmesh/enum_solver_type.h"
 #include "libmesh/enum_convergence_flags.h"
 
-#ifdef LIBMESH_HAVE_PETSC_HYPRE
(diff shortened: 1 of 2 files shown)


roystgnr (approved):
It's *now* handled upstream in PETSc; is there any reason to leave this in (in version guards) for older PETSc compatibility?

lindsayad:
My older PETSc code that worked with this libMesh code that I'm now removing was still post 3.22 release, so this would be dead code for any PETSc release

lindsayad:
I really think my brain is atrophying. I looked back at a PR from 2020 and thought to myself I was a lot smarter then. As I was falling asleep last night, I thought to myself ... "are you sure your code would be meaningless for all PETSc versions? Did you even check for PETSc version macros before responding to Roy?" And clearly I did not. It looks like this is *not* dead code between PETSc 3.12 and 3.23 😆

PR #4121, opened 2025-03-27
Fix for overwriting nodesets

DanielYankura:
- When creating sidesets from nodesets if the sideset shares an id with a nodeset they are combined
- This checks first that the two are actually the same (same id and same nodes)
- If they are not the same the instead of overwriting a new id is assigned to the new nodeset
- If this is to be merged some tests in MOOSE would have to be re-golded

refs idaholab/moose#30085

### DIFF
--- a/src/mesh/boundary_info.C
+++ b/src/mesh/boundary_info.C
@@ -141,6 +141,7 @@ BoundaryInfo & BoundaryInfo::operator=(const BoundaryInfo & other_boundary_info)
   _global_boundary_ids = other_boundary_info._global_boundary_ids;
   _side_boundary_ids = other_boundary_info._side_boundary_ids;
   _node_boundary_ids = other_boundary_info._node_boundary_ids;
+  _sideset_to_nodeset_conversion = other_boundary_info._sideset_to_nodeset_conversion;
   _edge_boundary_ids = other_boundary_info._edge_boundary_ids;
   _shellface_boundary_ids = other_boundary_info._shellface_boundary_ids;
 
@@ -348,6 +349,7 @@ void BoundaryInfo::clear()
   _boundary_ids.clear();
   _side_boundary_ids.clear();
   _node_boundary_ids.clear();
+  _sideset_to_nodeset_conversion.clear();
   _edge_boundary_ids.clear();
   _shellface_boundary_ids.clear();
   _ss_id_to_name.clear();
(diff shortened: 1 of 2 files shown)


DanielYankura:
I checked the tests that failed and it looks like it was because it was tested on those three MOOSE tests I fixed in idaholab/moose#30291 before the PR made it to next. I invalidated them to run them again. Hopefully that should fix it.

roystgnr:
`meshmodifiers/element_subdomain_modifier.boundary/moving_boundary FAILED (TIMEOUT) [FINISHED]`

A timeout, so probably random, but I'm going to kick it again to be safe.

roystgnr:
And of course the MOOSE ARM mac failure is no more than the same civet breakage; we'll ignore that.

DanielYankura:
That's what I figured. Those two tests weren't failing when I first opened the PR and they aren't failing when I run them locally.

src/mesh/boundary_info.C
```diff
@@ -2358,20 +2405,26 @@ BoundaryInfo::build_node_list_from_side_list()
...
+          // Check for equivalent nodeset here. If it has one the next portion can be skipped
           // Add each node node on the side with the side's boundary id
           for (auto i : side->node_index_range())
+          {
+            const boundary_id_type bcid = id_pair.second;
+            auto bcid_renum = bcid;
+            // If grouping by name check if nodeset id needs to be renumbered
+            if (_node_boundary_ids.find(bcid) != _node_boundary_ids.end() && !has_equivalent_nodeset(side, bcid))
+              // This sideset overlaps with a nodeset. Throw an error
+              libmesh_error_msg("Sideset " << bcid << " has the same ID as a preexisting nodeset. Rename one in order to build a nodeset from this sideset");
```
roystgnr, line 2417: I love this error checking in spirit.  In practice we're not going to call it nearly as often as we should, right?

Consider again `A--a--B--b--C`, with BCID 5 on node `A` and on sides `a` and `b`.

Suppose our element loop hits side a first.

We try to `build_node_list_from_side_list()`.  On side `a`, `has_equivalent_nodeset()` returns true because it sees node `A`, so it happily adds BCID 5 to node `B`.  Then on side `b`, `has_equivalent_nodeset()` returns true because it sees node `B`, so it happily adds BCID 5 to node `C`, and it finishes, having avoided emitting an error that IMHO it definitely should have.

Now suppose our loop takes us to `b` before `a`.  On `b`, `has_equivalent_nodeset()` doesn't see any of the BCID 5 nodes that it knows exist, so it returns false, and *now* we emit the error.

But our results shouldn't depend on element numbering.

roystgnr (changes requested):
I think we need unit tests here, maybe before even bothering to rewrite the code that will address them.  Come up with cases that should throw an error, and cases that shouldn't, and add a little test using each.

PR #4124, opened 2025-03-31
Refactor orientation-determining code

nmnobre:
Hi,

Though this keeps lexicographic ordering on the point coordinates as the orientation-determining factor, there is now both a single place where we determine edge orientations, `Elem::positive_edge_orientation()`, and a single place where we determine face orientations, `Elem::positive_face_orientation()`, for the entire codebase. This should make things easier if and when we decide to switch to comparing ids instead.

Cheers,
Nuno

### DIFF
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -1098,10 +1098,11 @@ public:
   bool positive_edge_orientation(const unsigned int i) const;
 
   /**
-   * \returns \p true if face \p i is positively oriented. A face with N
-   * vertices is positively oriented iff its 3 lexicographically greatest
-   * vertices (i.e. 3 lexicographically greatest nodes amongst the N leading
-   * nodes) are an odd permutation relative to their lexicographic ordering.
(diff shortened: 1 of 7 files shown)


nmnobre:
> I'd love to take credit for something clever but this code doesn't look familiar at all ;-)

Oh, I think you might be right @jwpeterson, the furthest I can go is to a massive commit by David Knezevic in 2011. The code I based the new face orientation function on seems to be from "the before times".

roystgnr (approved):
I'm not as confident as I should be that we have sufficient test coverage to detect any sign errors (well, `!` errors) here, but if we don't then that's a bug in our test coverage; I say we go for it.

nmnobre:
> I'm not as confident as I should be that we have sufficient test coverage to detect any sign errors (well, `!` errors) here

I doubled checked these a couple of times: right after doing it and with a fresh mind the next morning, so hopefully together with the tests we do have across libMesh and MOOSE, I haven't missed any signs. :)

PR #4127, opened 2025-04-01
Set nullspace for both J and P operators

lindsayad:
These can be different for instance when doing PJFNK. This now matters post https://gitlab.com/petsc/petsc/-/merge_requests/8247


roystgnr (approved):
I'd love it if we had some unit test coverage of these sorts of features, or at least *something* above the MOOSE level.

lindsayad:
Sigh, I 100% agree. But you've given me the green check mark this time and I feel like I have little time at this time in my life to do the right thing 😦

roystgnr:
I sometimes feel like it's cruel to think, "That's future-Roy's problem", but really I'm just the product of my upbringing; that jerk past-Roy really didn't make it easy for me to develop better habits.

PR #4128, opened 2025-04-02
Don't let GMSH physicals overwrite each other

roystgnr:
See https://github.com/libMesh/libmesh/discussions/4125 for an example case.


roystgnr:
No breakage other than the broken CI.  I'll add a unit test and then this should be ready to merge.

PR #4133, opened 2025-04-08
Fix for --disable-amr builds

roystgnr:


roystgnr:
I'm seeing valgrind tests pass postdating the failure, so hopefully that was a fluke.  I'll add them here along with the NoAMR recipe to be sure.

roystgnr:
Valgrind failures aren't related to this; still looking into them.

PR #4134, opened 2025-04-09
Netgen reconfigure fix

roystgnr:
Previously, if you re-ran configure (or had it re-run for you after a build system change) in a build directory where you had built Netgen and where your new configuration included Netgen, your subsequent make would fail at link time, because we were cleaning up the build directory without cleaning up the build stamp file.

### DIFF
--- a/configure
+++ b/configure
@@ -55270,7 +55270,8 @@ then :
 
                               my_top_builddir="$(pwd)"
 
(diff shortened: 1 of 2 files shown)


jwpeterson (approved):
I noticed recently that I had to manually `rm .buildstamp` in the netgen build dir, but had no idea why... sounds like you found the fix!

PR #4137, opened 2025-04-10
Fix FE::n_quadrature_points() when not using the qrule

roystgnr:
We weren't doing anything to track quadrature point counts in cases where we supplied a vector of points rather than relied on the qrule, and this was causing n_quadrature_points() to return potentially-erroneous values.

This was causing errors in JumpErrorEstimator subclasses, including a valgrind failure in one of our slit mesh unit tests.

This change fixes the OOB; hopefully CI doesn't find anything it regresses.  I'm worried about the possibility of fixed calculation errors in the indicators us exodiffs on MOOSE adaptivity tests.

### DIFF
--- a/include/fe/fe.h
+++ b/include/fe/fe.h
@@ -634,13 +634,6 @@ public:
    */
   virtual void attach_quadrature_rule (QBase * q) override;
 
-  /**
-   * \returns The total number of quadrature points.  Call this
-   * to get an upper bound for the \p for loop in your simulation
-   * for matrix assembly of the current element.
-   */
-  virtual unsigned int n_quadrature_points () const override;
-
 #ifdef LIBMESH_ENABLE_AMR
   /**
    * Computes the constraint matrix contributions (for
--- a/include/fe/fe_abstract.h
+++ b/include/fe/fe_abstract.h
@@ -483,10 +483,10 @@ public:
   virtual unsigned int n_shape_functions () const = 0;
 
   /**
-   * \returns The total number of quadrature points.  Useful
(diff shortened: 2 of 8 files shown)


jwpeterson:
Nice fix. Do you think this is a recent regression or something that has been there a while?

jwpeterson:
Also, I will plan to cherry-pick both of these commits onto the 1.8.1 branch unless you say otherwise.

roystgnr:
The bug has been there for forever, but I think it's only triggered by the relatively recent feature addition for testing jumps across slits.  That's actually a good point - if my answer is right, then the buggy code path is likely to only have seen use in our internal tests and in some of my Flex IGA analyses, so there shouldn't be any users impacted and there shouldn't be any exodiffs downstream.

roystgnr:
With this fixed I'm no longer seeing any screaming about our own code from valgrind, just a couple "blocks are definitely lost" complaints about PetscMallocAlign and a couple uninitialized value complaints about VTK.

roystgnr:
It's not a tiny change IMHO since it's ABI breaking, but it's definitely worth cherry-picking at least the commit with the fixes, if not the commit with the added assertions.  I tried to make it as small as possible with cherry-picking in mind; there's now some room for refactoring in fe.C that I may get to too someday but that I don't want to risk futzing with while I'm behind on other things.

jwpeterson:
I cherry-picked f1c3dbcea onto branch_v1.8.x, so it will be in 1.8.1 whenever we get around to making it.

PR #4138, opened 2025-04-11
Move demangle into libmesh_common

lindsayad:
Then we can use it in `cast_ref` and `cast_ptr`


jwpeterson (approved):
This looks like a good update to me that doesn't really have any downsides.

roystgnr (approved):
I don't like having to include `<string>` in as ubiquitous a file as `libmesh_common.h`, but at least on my system we're already including it from there indirectly ... and I'm betting the dependency is coming from something like `<iostream>` that's essentially unavoidable with our `OStreamProxy` declarations here.

PR #4140, opened 2025-04-11
StaticCondensation can also be a DofMap

lindsayad:
My goal is to be able to do field splits on the statically condensed system. This should sit in draft since it has meaningful impacts like converting `DofMap` methods into virtuals until I have some useful demonstration of capability


src/numerics/sparse_matrix.C
```diff
@@ -165,15 +165,11 @@ void SparseMatrix<Complex>::print(std::ostream & os, const bool sparse) const
...
 SparseMatrix<T>::build(const Parallel::Communicator & comm,
-                       const SolverPackage solver_package,
-                       const MatrixBuildType matrix_build_type /* = AUTOMATIC */)
+                       const SolverPackage solver_package)
 {
   // Avoid unused parameter warnings when no solver packages are enabled.
   libmesh_ignore(comm);
 
-  if (matrix_build_type == MatrixBuildType::DIAGONAL)
-    return std::make_unique<DiagonalMatrix<T>>(comm);
```
roystgnr, line 175: Deprecation-at-most reminder here.

examples/miscellaneous/miscellaneous_ex16/miscellaneous_ex16.C
```diff
@@ -171,7 +171,7 @@ main(int argc, char ** argv)
   libmesh_error_msg_if(!libMesh::relative_fuzzy_equals(*sys.solution, *sc_sys.solution, 1e-4),
                        "mismatching solution");
   libMesh::out << "Static condensation reduced problem size to "
-               << sc_sys.get_static_condensation().get_condensed_mat().m() << std::endl;
```
roystgnr, line 174: This old idiom should still work, right?  Could we leave it around, either in one example or in a unit test, to cover that?

lindsayad:
> E.g. I can't follow vector_fe_ex9 - we're querying `system.has_static_condensation()`, but I don't see anywhere we might be enabling static condensation in that example.

Static condensation is setup from the command line in `run.sh`

src/systems/system.C
```diff
@@ -73,6 +75,7 @@ System::System (EquationSystems & es,
   time                              (0.),
   qoi                               (0),
   qoi_error_estimates               (0),
+  _sc_dof_map                       (nullptr),
```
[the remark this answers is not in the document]
lindsayad: The `StaticCondensation` matrix which, if being used, will be the system matrix that users add into. The dof indices that users use for adding into this are the global dofs, hence its conceptual reliance on the full `DofMap`. However, internally the `StaticCondensation` class will need to map global system matrix data to statically condensed matrix data for which it must rely on the statically condensed dof map (`StaticCondensationDofMap`). 

Having the full system hold both full and condensed dof maps makes `StaticCondensation` class construction easy. Additionally, construction of sparsity patterns for the condensed matrices currently routes through the full dof map because it implements methods that the `SparsityPattern::Build` class calls like `find_connected_dofs` that the `StaticCondensationDofMap` doesn't currently have implementations for. But the `SparsityPattern::Build` does also query the `StaticCondensationDofMap` for things like `dont_condense_vars`. So this is another instance where we require both the `DofMap` and the `StaticCondensationDofMap` at the same time. I think sometime in the future `StaticCondensationDofMap` could be more full featured such that it can handle sparsity pattern construction for the statically condensed system all by itself, but I'm not there yet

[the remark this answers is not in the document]
lindsayad: That isn't a pro-con list. It's perhaps a "current necessity due to a degree of laziness" list? I have not yet made a serious attempt to see if I can make `SparsityPattern::Build` use only the statically condensed dof map. I think the one of the logical outcomes of such an attempt would be to change the current constructor in this PR
```c++
Build::Build (const DofMap & dof_map_in,
              const CouplingMatrix * dof_coupling_in,
              const std::set<GhostingFunctor *> & coupling_functors_in,
              const bool implicit_neighbor_dofs_in,
              const bool need_full_sparsity_pattern_in,
              const bool calculate_constrained_in,
              const StaticCondensationDofMap * const sc_in)
```
to
```c++
Build::Build (const DofMapBase & dof_map_in,
              const CouplingMatrix * dof_coupling_in,
              const std::set<GhostingFunctor *> & coupling_functors_in,
              const bool implicit_neighbor_dofs_in,
              const bool need_full_sparsity_pattern_in,
              const bool calculate_constrained_in)
```
which would in turn meaning enhancing `DofMapBase/StaticCondensationDofMap`.

I can attempt a serious run at this if you would like me to

examples/miscellaneous/miscellaneous_ex16/miscellaneous_ex16.C
lindsayad: What do you mean? `get_static_condensation()` now returns a `DofMapBase` instead of the `SparseMatrix` derived class. So the options are either (what is currently used in that example)
```c++
sc_sys.get_static_condensation_system_matrix().get_condensed_mat().m()
```
or
```c++
sc_sys.get_static_condensation().n_dofs()
```

lindsayad:
I've restored the `matrix_build_type`

examples/miscellaneous/miscellaneous_ex16/miscellaneous_ex16.C
roystgnr: Ugh.  I misunderstood part of the API change here.

Is there anything we can do to preserve backwards compatibility here?  Maybe make `get_static_condensation()` do what `get_static_condensation_system_matrix()` now does, and make `get_static_condensation_map()` the new API that returns the `DofMapBase` subclass?  If we've got what's effectively a new API then we can give it a new name, and then we've still got the old name left over to try to preserve compatibility.

The old API made it into the 1.8.0 branch, and wasn't marked `libmesh_experimental()`, and I just don't like yanking users around.  I like to be one of the people who ameliorates others' API consistency problems, despite the ugliness of `#if VERSION > foo #elif VERSION > bar #etc` code; I don't like to be the one of the ones *creating* the problems.

src/systems/system.C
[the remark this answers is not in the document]
roystgnr: Oh, definitely don't attempt a serious run at this yet.  I'm still just spitballing here.  Trying to understand the design, not to criticize it.

Even in my idea, we'd still have two DofMapBase subclasses, it's just that one of them would own the other, to make the APIs cleaner.  An Elem is-a DofObject, which at times has-a second DofObject accessible via `old_dof_object()`, which is non-nullptr in cases where we need to track both an old and a current numbering at once.  The analogous thing to do here would be a DofMap which has-a StaticCondensationDofMap, non-null when we're doing static condensation; but perhaps the better solution in this case would be to have a StaticCondensationDofMap for the advanced cases, which then always has-a plain DofMap that doesn't itself know about static condensation.

Even that might not be an improvement.  If you're 99% sure it wouldn't be an improvement, I'm fine with us not trying.  If you're 90% sure it wouldn't be an improvement, I'd like to get something in the new code paths marked libmesh_experimental() but I'm still fine with us not trying *yet*.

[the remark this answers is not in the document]
lindsayad: So you're thinking that the `DofMap`, not the `System`, should own `StaticCondensationDofMap`?

examples/miscellaneous/miscellaneous_ex16/miscellaneous_ex16.C
lindsayad: This should be hopefully resolved by https://github.com/libMesh/libmesh/pull/4140/commits/6015c7f0be4cb036fa5aa6a1c93fa374b860b07e

lindsayad:
Cannot reproduce this single test failure on my ubuntu container so I'm going to say this is a bad test specification. Will relax the tolerancing comparisons in MOOSE

lindsayad:
this should be ready for further review

roystgnr (approved):
If @jwpeterson isn't still on vacation or buried by more urgent things to catch up on, then I'd like to give him time to look at this too, but if he doesn't chime in by Monday I'd say let's merge this then.

jwpeterson:
> then I'd like to give him time to look at this too,

To be honest, I haven't followed the design discussion here closely, so I'm going to defer to you guys' expertise. 

It's a little concerning to me that we are making `DofMap::dof_indices()` virtual... in the past we have put some effort into squeezing every last drop of performance out of that function, since it is called millions of times and often showed up in profiling. At the same time, we also won't be directly measuring any slowdowns, since I removed the logging from `DofMap::dof_indices()` in  be55a073911bfe31d640351c18b72cc6969f29fa. 

So I guess my main request would be if you guys have any performance regression testing on CIVET, to keep an eye on that before/after this is merged just to get an idea of whether there is any overall performance loss.

lindsayad:
Will definitely monitor our stuff in MOOSE. Doing some libmesh benchmarking

After this PR
```
real    25m11.291s
user    24m9.356s
sys     0m49.729s
```

Before this PR
```
real    25m26.384s
user    24m7.176s
sys     1m9.175s
```

So that looks the same to me

roystgnr:
Huh.  That `sys` difference is kind of huge, but IIRC that measures time spent in the kernel (e.g. waiting for the filesystem to report a read or write as complete), not anything this PR would have affected.  A `user` time 0.2% slower might be real, but it's definitely within the `LIBMESH_BENCHMARK` margin of error and also within my personal margin of "I don't care".

> Will definitely monitor our stuff in MOOSE.

IIRC the one application to worry about here is Relap-7 - at least I think that was the one commonly doing solves on a zillion Edge2 elements where the actual calculations per element were thin enough that every little bit of overhead showed up.  A bunch of the micro-optimizations in things like `dof_indices` date back to my attempt to optimize that kind of workload 8 or 9 years ago.

lindsayad:
@joshuahansel you'll need to help us monitor performance in THM. Perhaps we can setup some performance logging stuff for THM case(s) with what @loganharbour has been working on recently?

joshuahansel:
> @joshuahansel you'll need to help us monitor performance in THM. Perhaps we can setup some performance logging stuff for THM case(s) with what @loganharbour has been working on recently?

Sounds good, lemme know whenever the capability is ready.

loganharbour:
> > @joshuahansel you'll need to help us monitor performance in THM. Perhaps we can setup some performance logging stuff for THM case(s) with what @loganharbour has been working on recently?
> 
> Sounds good, lemme know whenever the capability is ready.

Metrics are already being collected, although not used.

See https://github.com/idaholab/moose/tree/next/modules/combined/performance for examples. Test spec is named "performance" instead of "tests" which are ran on master: https://civet.inl.gov/job/2938624/

PR #4141, opened 2025-04-12
Add our new unit test mesh to EXTRA_DIST

roystgnr:
This should fix a `make distcheck` regression.


roystgnr:
We're still seeing distcheck failures, but now in distclean in netgen; this seems to have fixed the unit test installation.

PR #4146, opened 2025-04-18
Drop LOG_SCOPE calls from DofMap::dof_indices()

jwpeterson:
I would like to propose that we drop the `LOG_SCOPE` calls from the various `DofMap::dof_indices()` functions, because it appears to be both adding overhead to the simulation time while also not providing particularly useful timing information. For example, we often see cases like the following:
```
 -------------------------------------------------------------------------------------------------------------------------------
| Event                                            nCalls     Total Time  Avg Time    Total Time  Avg Time    % of Active Time  |
|                                                             w/o Sub     w/o Sub     With Sub    With Sub    w/o S    With S   |
|-------------------------------------------------------------------------------------------------------------------------------|
|   dof_indices()                                  31640150   7.6276      0.000000    7.6276      0.000000    0.93     0.93     |
```
where the average time per call ends up being less than 1 microsecond. It is therefore not really possible to tell how much of the reported time is just the time taken by the logging code vs. the time taken by `dof_indices()` itself.

### DIFF
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -2192,7 +2192,12 @@ void DofMap::dof_indices (const Elem * const elem,
   // active)
   libmesh_assert(!elem || elem->active());
 
-  LOG_SCOPE("dof_indices()", "DofMap");
+  // dof_indices() is a relatively light-weight function that is
+  // called millions of times in normal codes. Therefore, it is not a
+  // good candidate for logging, since the cost of the logging code
+  // itself is roughly on par with the time required to call
+  // dof_indices().
+  // LOG_SCOPE("dof_indices()", "DofMap");
 
   // Clear the DOF indices vector
   di.clear();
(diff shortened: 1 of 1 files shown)


roystgnr (approved):
On my CPU the most basic `dof_indices()` call (1 variable, on tets) seems to take 41ns to log and 57ns to actually run, just based on some before and after computations on large meshes.

I'd argue that means it's *either* adding overhead to the simulation time *or* not providing particularly useful timing information.  If PerfLog shows 7.6s on logged dof_indices() in a 3D simulation then probably at least 4.6s of that is real, and if that number scales up to the point where the logging cost is annoying then the larger non-logging cost is probably worth looking at.  I'm going to have to uncomment this to productively look for performance improvement opportunities in an explicit-timestepping code at some point.

On the other hand, with most methods the PerfLog overhead is so cheap that we can just run with it on all the time and not worry about it, which I really like.  That's clearly not true here, so I'm happy commenting it out.

PR #4148, opened 2025-04-21
Vtk 9.4 fixes

roystgnr:
We can't even configure against VTK 9.4.x without updated detection code.


roystgnr:
I'm still building this to test locally, but on the off chance that we still have runtime compatibility and merely lost configure-time compatibility, let's see what CI thinks of it right away.

configure
```diff
@@ -56132,14 +56132,21 @@ fi
...
 
-                              if test -r $VTK_INC/vtkVersionMacros.h
+               if                                                                                                                               test -r $VTK_INC/vtkVersionQuick.h
+then :
+
+                       vtkmajor=`grep "define VTK_MAJOR_VERSION" $VTK_INC/vtkVersionQuick.h | sed -e "s/.*#define VTK_MAJOR_VERSION[ ]*//g"`
+                       vtkminor=`grep "define VTK_MINOR_VERSION" $VTK_INC/vtkVersionQuick.h | sed -e "s/.*#define VTK_MINOR_VERSION[ ]*//g"`
+                       vtkbuild=`grep "define VTK_EPOCH_VERSION" $VTK_INC/vtkVersionQuick.h | sed -e "s/.*#define VTK_EPOCH_VERSION[ ]*//g"`
+
+elif                      test "x$vtkmajor" = "x" && test -r $VTK_INC/vtkVersionMacros.h
```
lindsayad, line 56142: why the introduction of the `vtkmajor` test?

m4/vtk.m4
```diff
@@ -119,22 +119,41 @@ AC_DEFUN([CONFIGURE_VTK],
...
+                     dnl replacing it with a "subminor/patch numbers
+                     dnl are not epoch numbers" problem.
+                     [test -r $VTK_INC/vtkVersionQuick.h],
+                     [
+                       vtkmajor=`grep "define VTK_MAJOR_VERSION" $VTK_INC/vtkVersionQuick.h | sed -e "s/.*#define VTK_MAJOR_VERSION[ ]*//g"`
+                       vtkminor=`grep "define VTK_MINOR_VERSION" $VTK_INC/vtkVersionQuick.h | sed -e "s/.*#define VTK_MINOR_VERSION[ ]*//g"`
+                       vtkbuild=`grep "define VTK_EPOCH_VERSION" $VTK_INC/vtkVersionQuick.h | sed -e "s/.*#define VTK_EPOCH_VERSION[ ]*//g"`
+                     ],
+                     dnl If we have the vtkVersionMacros.h (VTK 6.x) and have yet to find the version, look there
+                     [test "x$vtkmajor" = "x" && test -r $VTK_INC/vtkVersionMacros.h],
```
lindsayad, line 136: why the introduction of the `vtkmajor` test? This is the else condition so we didn't set vtkmajor in the if branch. No harm in testing; mostly asking out of curiosity

roystgnr: Ah, this is the else condition.  I guess I was thinking about fallthrough (which would cause the valid variables to then get overwritten), but that won't happen.

loganharbour:
So are we good with this?

roystgnr:
My own tests aren't done either, but they've at least hit some VTK coverage and nothing's failing; CI has hit all our VTK coverage in those sweeps and nothing's failing.

jwpeterson (approved):
This looks good to me. I am surprised (in a good way) that we didn't have to change anything with what libraries we need to link with.

PR #4152, opened 2025-04-23
More rigorous implementation and verification of mesh clone()

roystgnr:
The second commit here fixes the underlying problem uncovered in https://github.com/idaholab/moose/pull/30155#discussion_r2056875687

The first commit is a bit of refactoring for simplicity, the third gives us more rigorous clone() behavior, the fourth *checks* for rigorous clone() behavior, and the last checks that our fixes were sufficient in the tet meshing case that first triggered the problem.

### DIFF
--- a/include/mesh/distributed_mesh.h
+++ b/include/mesh/distributed_mesh.h
@@ -117,7 +117,13 @@ public:
    * Virtual copy-constructor, creates a copy of this mesh
    */
   virtual std::unique_ptr<MeshBase> clone () const override
-  { return std::make_unique<DistributedMesh>(*this); }
+  {
+    auto returnval = std::make_unique<DistributedMesh>(*this);
+#ifdef DEBUG
+    libmesh_assert(*returnval == *this);
+#endif
+    return returnval;
+  }
 
   /**
    * Destructor.
--- a/src/mesh/distributed_mesh.C
+++ b/src/mesh/distributed_mesh.C
@@ -192,6 +192,12 @@ DistributedMesh::DistributedMesh (const MeshBase & other_mesh) :
   _next_free_unpartitioned_elem_id(this->n_processors())
 {
   this->copy_nodes_and_elements(other_mesh, true);
+
+  // The prepare_for_use() in copy_nodes_and_elements() is going to be
+  // tricky to remove without breaking backwards compatibility, but it
+  // updates some things we want to just copy.
+  this->copy_cached_data(other_mesh);
+
   this->copy_constraint_rows(other_mesh);
 
   auto & this_boundary_info = this->get_boundary_info();
--- a/src/mesh/distributed_mesh.C
+++ b/src/mesh/distributed_mesh.C
@@ -149,17 +149,12 @@ DistributedMesh::~DistributedMesh ()
 // make sure the compiler doesn't give us a default (non-deep) copy
 // constructor instead.
 DistributedMesh::DistributedMesh (const DistributedMesh & other_mesh) :
-  UnstructuredMesh (other_mesh), _is_serial(other_mesh._is_serial),
-  _is_serial_on_proc_0(other_mesh._is_serial_on_proc_0),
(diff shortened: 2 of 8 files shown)


src/mesh/distributed_mesh.C
```diff
@@ -149,17 +149,12 @@ DistributedMesh::~DistributedMesh ()
...
-  _next_free_local_elem_id(this->processor_id()),
-  _next_free_unpartitioned_node_id(this->n_processors()),
-  _next_free_unpartitioned_elem_id(this->n_processors())
+  DistributedMesh(static_cast<const MeshBase &>(other_mesh))
 {
-  this->copy_nodes_and_elements(other_mesh, true);
-  this->copy_constraint_rows(other_mesh);
+  _is_serial = other_mesh._is_serial;
+  _is_serial_on_proc_0 = other_mesh._is_serial_on_proc_0;
+  _deleted_coarse_elements = other_mesh._deleted_coarse_elements;
```
jwpeterson, line 156: Any reason we don't also initialize `_is_serial`, `_is_serial_on_proc_0`, `_deleted_coarse_elements`, etc. in the initialization list rather than setting them in the constructor body? I see that we weren't setting those fields in the initialization list before either, so there might be a good reason for it...

jwpeterson (approved):
Looks good to me, it might be worth documenting that we do additional verification checks when cloning in `METHOD=dbg` builds.

roystgnr:
It would be good if it wasn't breaking at least one MOOSE reactor module class.  I think that's got to be a downstream bug, but I don't want to merge until I've found and fixed it.

include/mesh/distributed_mesh.h
```diff
@@ -117,7 +117,13 @@ class DistributedMesh : public UnstructuredMesh
    * Virtual copy-constructor, creates a copy of this mesh
    */
   virtual std::unique_ptr<MeshBase> clone () const override
-  { return std::make_unique<DistributedMesh>(*this); }
+  {
+    auto returnval = std::make_unique<DistributedMesh>(*this);
+#ifdef DEBUG
+    libmesh_assert(*returnval == *this);
```
jwpeterson, line 123: It took me a while to realize that `MeshBase::operator==` eventually calls the virtual `MeshBase::subclass_locally_equals()` function. It might be worth updating the docs for `MeshBase::operator==` to explain that this non-virtual function internally calls a customized equivalence comparison which is appropriate to the 
underlying type of `other_mesh`.

src/mesh/distributed_mesh.C
roystgnr: They get set in the delegated constructor initializer list, but then `copy_nodes_and_elements()` in that constructor does a `prepare_for_use()`, and so I want to be safe and not trust that they're still set the same way.  To be able to meet the new `assert(*returnval == *this)` standard in `clone()` we really want everything to be matching even when we may be cloning an unprepared mesh where "matching" is sometimes at odds with "accurate".

I'm tempted to eventually assert that these values are already correct except in the `!other_mesh.is_prepared()` case, or to refactor most of `copy_nodes_and_elements()` into a (differently named for backwards compatibility) helper function that does everything but the `prepare_for_use()`, but this is such a deep change already (it just exposed a bug in some of my MOOSE framework code that only gets triggered in a module!) and it's such an important change (it fixes a bug that leads to really rare+subtle+confusing failures way downstream) that I wanted to make the critical bits as small as possible for now.

roystgnr:
I'll need to rerun testing anyway after https://github.com/idaholab/moose/pull/30392 gets merged, so I'll push the documentation improvements then.

roystgnr:
And by "after https://github.com/idaholab/moose/pull/30392 gets merged" I of course mean "after https://github.com/idaholab/moose/pull/30399 gets merged".  It would turn out to be two separate bugs.

roystgnr:
Force pushing a rebased-on-master branch to kick CI; it looks like we still didn't have both the patches fully merged downstream yet last time.

PR #4156, opened 2025-04-25
Mark restore_array const

### LINKED ISSUE #4155: Mark PETSC Vector `restore_array` function as `const`
All that function does is update a `mutable` member and call another `const` function. This would resolve the conundrum of being able to call `get_array_read` on a const reference to a `PetscVector`, but being unable to call the _mandatory_ `restore_array` after the read. Certain MOOSE apps circumvent this by casting `const` away, but the elegant solution here is to just mark `restore_array` as `const`.

dschwen:
Closes #4155

### DIFF
--- a/include/numerics/petsc_vector.h
+++ b/include/numerics/petsc_vector.h
@@ -229,7 +229,7 @@ public:
    * \note This MUST be called after get_array() or get_array_read()
    * and before using any other interface functions on PetscVector.
    */
-  void restore_array();
+  void restore_array() const;
 
   virtual NumericVector<T> & operator += (const NumericVector<T> & v) override;
 
@@ -1160,7 +1160,7 @@ const PetscScalar * PetscVector<T>::get_array_read() const
 
 template <typename T>
 inline
-void PetscVector<T>::restore_array()
+void PetscVector<T>::restore_array() const
 {
   // \note \p _values_manually_retrieved needs to be set to \p false
   // \e before calling \p _restore_array()!


roystgnr (approved):

dschwen:
Ok, let's pump the brakes on this. @lindsayad pointed out that there may be issues with race conditions in threaded regions. I'll check if we need to add locking.

lindsayad:
A few thoughts:

1. I wish that we had some record about why we use `VecGetArray(Read)` and `VecRestoreArray(Read)` in the first place as opposed to just using things like `VecGetValues` and `VecSetValues`. But when I do a `git log -S VecGetArray --reverse` I see the first libMesh git commit haha. The natural hypothesis would be performance, but how much performance do we gain? I suppose having the raw array may also be helpful for doing larger global vector operations compared to element assembly.
2. Let's presume it's a good idea to have these raw array getter APIs. My vote would be to split `_get_array` and `_restore_array` into methods that either call the non-const `VecGetArray/VecRestoreArray` or the const `VecGetArrayRead/VecRestoreArrayRead`, but not both. Another option would be to eliminate these helper methods and put the code into the APIs. For APIs I think we should have `void get_array_read() const`, `void get_array()`, `void restore_array_read() const` (new), and `void restore_array()`

lindsayad (changes requested):
Let's workout the discussion a bit

lindsayad:
I think that's a nice idea

roystgnr:
IIRC Derek's changes were purely for performance.  I'd agree that figuring out exactly how much performance would be nice.

I like the idea of splitting our helper methods into const and non-const versions rather than letting read vs write be a parameter.

I hate the idea of exposing these methods to the user.  That would require users to add Petsc-specific calls for non-Petsc-specific code, and to manually manage more complication.

I love the sentinel pattern idea in theory.  It's kind of embarrassing that we still have pairs of C functions like this that don't yet have an RAII wrapper, now that I think about it.  I guess the names are "Set/Restore" instead of the usual "Create/Destroy" and that was enough for it to slip past us?

lindsayad:
> I hate the idea of exposing these methods to the user. That would require users to add Petsc-specific calls for non-Petsc-specific code, and to manually manage more complication.

What are you referring to? The sentinel pattern idea?

dschwen:
I'm sure Roy dislikes the idea of special treatment for PETSc vectors. I do understand that from his library developer standpoint, but from my side I wouldn't mind having an API for my 99-100% use-case.

lindsayad:
> IIRC Derek's changes were purely for performance

`_get_array` and `_restore_array` were added in 6fa3a3b1cfd19edb77dca7b4eaddde23b2e273d2, which was well before @friedmud added the public `get_array, get_array_read, restore_array` in https://github.com/libMesh/libmesh/issues/1056. But as the message in 6fa3a3b1cfd19edb77dca7b4eaddde23b2e273d2 says, it was for performance

roystgnr:
I'm referring to "eliminate these helper methods and put the code into the APIs".  Or maybe I misunderstood you?  We've got exposed `get_array()`, `restore_array()` already, I thought you were talking about requiring users to manually do the work of `_get_array()` rather than relying on the helper methods to be called correctly from the more general APIs.

lindsayad:
> I thought you were talking about requiring users to manually do the work of `_get_array()`

Definitely not. I was suggesting that the non-const branch of `_get_array()` be moved into `get_array` and the const branch of `_get_array` be moved into `get_array_read`. Other methods on `PetscVector` can just as easily call `get_array/restore_array` as `_get_array/_restore_array`, just remove a character 😄

PR #4158, opened 2025-04-28
Add const where we can

roystgnr:
I noticed these while debugging some polyhedra test failures


jwpeterson (approved):
These are non-reference parameters so I usually don't think about making such things `const`. Did this lead to you finding a bug or something?

roystgnr:
This is just the code I happened to be stepping through while I found an unrelated bug (now fixed) in the new polyhedra code.  Not seeing `const` where I expected it made me wonder whether there was something going on I wasn't understanding, so I added it to check.

PR #4160, opened 2025-04-28
Make set_node() a proper setter

roystgnr:
Just returning a reference to the user doesn't let us do anything complicated, and for polyhedra we'll need to do something complicated.


jwpeterson (approved):
This would be good to get merged, I will need to fix some of our code to use the non-deprecated API ASAP.

PR #4164, opened 2025-05-05
More static maps

roystgnr:
One last feature pulled out of #4150

### DIFF
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -626,6 +626,15 @@ public:
   virtual ElemType type () const = 0;
 
   /**
+   * This array maps the integer representation of the \p ElemType enum
+   * to the geometric dimension of the element.
+   *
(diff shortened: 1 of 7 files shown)


jwpeterson (approved):
I assume the unit test will catch the case where someone adds a new element type but forgets to update the `type_to_default_order_map`? If so, :+1:

roystgnr:
It will (combined with -Werror on semirecent compilers, at least) when someone remembers to `INSTANTIATE_ELEMTEST(THEIR_NEW_ELEM_TYPE)`.  I'm actually not sure how to ensure *that*, though.  The unit tests are supposed to be runnable independently and selectively, so it's hard to distinguish between "the author forgot to add a unit tests instantiation for ElemType t" and "there's an instantiation but the user used a `--re Foo` option that skipped it".

PR #4165, opened 2025-05-06
Vector Variable Output for Mesh Function

csdechant:
This PR adds additions and tests for MeshFunction::operator() that allows for Lagrange, Nedelec, and Raviart Thomas vector variables.


src/mesh/mesh_function.C
```diff
@@ -305,6 +305,84 @@ void MeshFunction::operator() (const Point & p,
...
+              FEComputeData data (this->_eqn_systems, mapped_point);
+              const Point & pt = data.p;
+
+              // where the solution values for the var-th variable are stored
+              std::vector<dof_id_type> dof_indices;
+              this->_dof_map.dof_indices (element, dof_indices, var);
+
+              // interpolate the solution
+              {
+                RealGradient value = 0.;
```
roystgnr, line 371: Try just `Gradient` here - the solution may be complex-valued.

include/fe/fe_interface.h
```diff
@@ -31,242 +29,239 @@
...
+  class Point;
+  class MeshBase;
+  enum FEFamily : int;
+  enum Order : int;
+  enum FEFieldType : int;
+  enum ElemType : int;
+  enum FEContinuity : int;
 
 #ifdef LIBMESH_ENABLE_PERIODIC
-class PeriodicBoundaries;
```
roystgnr, line 51: What fraction of the diff here is whitespace changes?  Those are fine if they're motivated and restricted to a separate commit (or ideally PR), but if they're mixed in to a commit with substantive changes then that's basically turning PR review (as well as any future git diffs that cross the commit boundary) into an Obfuscated C Code contest entry.

I'm guessing this was an automated editor thing?  I told vim a long time ago to automatically run clang-format whenever I saved a MOOSE file (or any file in a subdirectory of a directory with a `.clang-format` file), but soon afterward I was very unpleasantly surprised the first time I tried to save a libMesh file that I was working on from *within the MOOSE submodule*.

csdechant: You're correct in that this is an automated editor thing. I also noticed the drastic edits with using the auto editing in libMesh files. I thought I corrected those unwanted edits, but I missed this file. I will fix the unwanted whitespace edits.

src/fe/fe_interface.C
```diff
@@ -756,6 +756,227 @@ bool FEInterface::on_reference_element(const Point & p,
...
+RealGradient FEInterface::vectorshape_hcurl(const unsigned int dim,
+                                            const FEType & fe_t,
+                                            const Elem * elem,
+                                            const unsigned int i,
+                                            const Point & p)
+{
+  auto fe = FEVectorBase::build(dim, fe_t);
+  std::vector<Point> pt = {p};
+  fe->get_dphideta(); // to compute the map
+  fe->reinit(elem, &pt);
```
roystgnr, line 786: This is a little horrifying. If we're on a Hex27 evaluating 6 p=1 Raviart-Thomas shape functions, ideally we'd be able to get away with evaluating 27 Lagrange mapping gradients, but if we generate a new FE every time then we do a ton of heap allocation/deallocation *and* we have to do those Lagrange evaluations 27*6=162 times.

Not too long ago I had to do an optimization pass on the `RATIONAL_BERNSTEIN` FE initialization because the non-vector code path there has a similar problem; it does show up in profiling.  But, just like here, I couldn't figure out how to actually fix the non-vector code path, only how to make the typical user's initialization evaluate shape functions all at once rather than one at a time so that we could reuse some of the intermediate results.  We ought to look for the possibility of doing something similar here.

include/fe/fe_interface.h
```diff
@@ -356,6 +356,57 @@ class FEInterface
...
+   * point \p p for H_CURL continuity.. This method allows you to
+   * specify the dimension, element type, and order directly.
+   * Automatically passes the request to the appropriate finite
+   * element class member.
+   */
+  static RealGradient vectorshape_hcurl(const unsigned int dim,
+                                        const FEType & fe_t,
+                                        const Elem * elem,
+                                        const unsigned int i,
+                                        const Point & p);
```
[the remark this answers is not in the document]
roystgnr: Ah, never mind, I see it - that's what `correct_vectorshape` is for.  IMHO that should be our only public API, renamed `vectorshape` so we don't leave users wondering which APIs return *incorrect* shapes, and we can hide these three in the `private:` section.

src/mesh/mesh_function.C
```diff
@@ -305,6 +305,84 @@ void MeshFunction::operator() (const Point & p,
...
+              // where the solution values for the var-th variable are stored
+              std::vector<dof_id_type> dof_indices;
+              this->_dof_map.dof_indices (element, dof_indices, var);
+
+              // interpolate the solution
+              {
+                Gradient value = 0.;
+
+                for (auto i : index_range(dof_indices))
+                  value += this->_vector(dof_indices[i]) * FEInterface::correct_vectorshape(dim, fe_type, element, i, pt);
```
roystgnr, line 374: We're looping over all indices, so let's not bother going through `FEInterface` at all here, let's just use an `FE` so we can pull its properly-transformed `phi()` directly and not do `n_dofs` times as much mapping work.

roystgnr (changes requested):
We're getting closer now, and this is important enough and new enough that I could wince and merge something with bad asymptotic performance that we can fix later, but we've got to be able to handle the mixed-vector-and-scalar use case, not add a new API that'll make adding that handling harder later.

tests/mesh/mesh_function.C
```diff
@@ -220,7 +227,211 @@ public:
...
 #endif // LIBMESH_ENABLE_AMR
-};
 
+#ifdef LIBMESH_HAVE_PETSC
+  // Tests the projection of Lagrange Vectors using MeshFunction
+  void vectorMeshFunctionLagrange()
+  {
+    LOG_UNIT_TEST;
+
+    // Reading mesh and solution infromation from XDA files
```
roystgnr:
```suggestion
    // Reading mesh and solution information from XDA files
```

tests/mesh/mesh_function.C
```diff
@@ -220,7 +227,211 @@ public:
...
+    ReplicatedMesh mesh(*TestCommWorld);
+    mesh.read("solutions/lagrange_vec_solution_mesh.xda");
+    EquationSystems es(mesh);
+    es.read("solutions/lagrange_vec_solution.xda", READ,
+            EquationSystems::READ_HEADER |
+                EquationSystems::READ_DATA |
+                EquationSystems::READ_ADDITIONAL_DATA);
+    es.update();
+
+    // Pulling the correct system and variable infromation from
```
roystgnr:
```suggestion
    // Pulling the correct system and variable information from
```

tests/mesh/mesh_function.C
```diff
@@ -220,7 +227,211 @@ public:
...
+    ReplicatedMesh mesh(*TestCommWorld);
+    mesh.read("solutions/nedelec_one_solution_mesh.xda");
+    EquationSystems es(mesh);
+    es.read("solutions/nedelec_one_solution.xda", READ,
+            EquationSystems::READ_HEADER |
+                EquationSystems::READ_DATA |
+                EquationSystems::READ_ADDITIONAL_DATA);
+    es.update();
+
+    // Pulling the correct system and variable infromation from
```
roystgnr:
```suggestion
    // Pulling the correct system and variable information from
```

tests/mesh/mesh_function.C
```diff
@@ -220,7 +227,211 @@ public:
...
+    LIBMESH_ASSERT_FP_EQUAL(libMesh::libmesh_real(output_vec(1)), libMesh::libmesh_real(output_expected(1)),
+                            TOLERANCE * TOLERANCE);
+  }
+
+  // Tests the projection of Raviart Thomas Vectors using MeshFunction
+  void vectorMeshFunctionRaviartThomas()
+  {
+    LOG_UNIT_TEST;
+
+    // Reading mesh and solution infromation from XDA files
```
roystgnr:
```suggestion
    // Reading mesh and solution information from XDA files
```

tests/mesh/mesh_function.C
```diff
@@ -220,7 +227,211 @@ public:
...
+  // Tests MeshFunction for mixed scalar and vector variable outputs
+  // In this case, the variables types are:
+  //  - variable index: 0, variable family: Raviart Thomas, Order: First
+  //  - variable index: 1, variable family: Monomial, Order: Constant
+  //  - variable index: 2, variable family: Scalar, Order: First
+  void mixedScalarAndVectorVariables()
+  {
+    LOG_UNIT_TEST;
+
+    // Reading mesh and solution infromation from XDA files
```
roystgnr:
```suggestion
    // Reading mesh and solution information from XDA files
```

tests/mesh/mesh_function.C
```diff
@@ -220,7 +227,211 @@ public:
...
+    ReplicatedMesh mesh(*TestCommWorld);
+    mesh.read("solutions/raviart_thomas_solution_mesh.xda");
+    EquationSystems es(mesh);
+    es.read("solutions/raviart_thomas_solution.xda", READ,
+            EquationSystems::READ_HEADER |
+                EquationSystems::READ_DATA |
+                EquationSystems::READ_ADDITIONAL_DATA);
+    es.update();
+
+    // Pulling the correct system and variable infromation from
```
roystgnr:
```suggestion
    // Pulling the correct system and variable information from
```

Or, could we just refactor the new tests?  Seeing the same typo copy-and-pasted a half dozen times is generally a sign that we've been doing too much copy and pasting.

Not a requirement (I don't care about copy-and-pasted code in unit tests nearly so much as I do when it's in the library) but I'll bet we could shave off a hundred lines here.

include/mesh/mesh_function.h
```diff
@@ -27,6 +27,7 @@
 #include "libmesh/tensor_value.h"
 #include "libmesh/tree_base.h"
 #include "libmesh/parallel_object.h"
+#include "libmesh/fe.h"
```
roystgnr: If we're not using `FE` in the header, let's not include `fe.h` here.  Do we even need it in the `.C` file?

roystgnr (changes requested):

include/mesh/mesh_function.h
csdechant: I think I was using something from `FE` in a previous iteration, but I am not using it now. This line of code has been removed.

tests/mesh/mesh_function.C
csdechant: I liked the idea of refactoring (it usually makes a code more readable, which I am always in favor for). I made the edits by including a helper function.

PR #4169, opened 2025-05-09
NumericVector::print/read_matlab

roystgnr:
This is easier than the SparseMatrix versions were, and it's useful for some utilities and debugging I'm working on.


roystgnr:
Well, that's clearly a real bug.  Looks like I should have tested the DistributedVector instantiations, not just the PetscVector ones.

PR #4171, opened 2025-05-15
Remove warnings guards around gzstream.h

roystgnr:
The shadowing warnings we used to get were a gcc 8 bug, and should be fixed by now.

I say "should be", of course, which does not imply "were".  gcc 9.5 works fine for me, and hopefully 8.x is fine now too, but if our "Min gcc" CI recipes aren't happy then we'll just let this branch languish a while.

### DIFF
--- a/src/mesh/abaqus_io.C
+++ b/src/mesh/abaqus_io.C
@@ -25,9 +25,7 @@
 
 // gzstream for reading compressed files as a stream
 #ifdef LIBMESH_HAVE_GZSTREAM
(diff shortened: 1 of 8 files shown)


roystgnr:
Huh.  8.5.0 on our min gcc tests, and it's definitely getting the --enable-paranoid-warnings --enable-werror testing, and it's passing.  I think we're good to go here.

jwpeterson (approved):

PR #4176, opened 2025-05-20
Matrix solve app

roystgnr:
I had some need to test various matrices against various PETSc linear solver options, *after* doing manual tweaking of those matrices, so I wrote this little utility to read a matrix and a vector from files and throw them at a `LinearSolver`; I figure it might be of general usefulness too.


roystgnr:
Test failure unrelated

PR #4177, opened 2025-05-22
Use non-deprecated set_node overload

roystgnr:
This slipped in via a PR predating the API change.


roystgnr:
Going to merge this right away: it passes on my system, the deprecated call is holding up devel->master, and that's holding up other things downstream.

PR #4181, opened 2025-05-27
ignore additional Abaqus keywords

jessecarterMOOSE:
The current Abaqus parser reads `*NODE` but ignores `*NODE OUTPUT` sections (and similarly for `*ELEMENT`). There are more valid Abaqus keyword sections that start with *NODE/*ELEMENT that should be ignored. Otherwise it tries to read these sections as mesh data and inevitably throws an error.

This PR adds more keywords to ignore.


roystgnr:
Yeah, that'd work fine; good idea.

jessejcarter:
FYI I added the following lines to the bottom of one of the test meshes:

```
** Sections below test if libmesh will ignore them
*NODE OUTPUT
x
y
*ELEMENT OUTPUT
x
y
*NODE FILE
foo
*ELEMENT MATRIX
bar
```

jessecarterMOOSE:
I'm guessing these failures are unrelated

PR #4182, opened 2025-05-28
Track a MeshBase::interior_mesh() to enable BoundaryMesh serialization

roystgnr:
I don't recall whether @YaqiWang hit this bug or whether I found it while investigating the issues he did hit in #4180, but either way thanks go to him for pushing at a mix of capabilities that hadn't been properly tested and working before.

Previously, we couldn't correctly serialize a distributed boundary mesh, because the Elem Packing code would try to find interior parents by id in the wrong mesh.  Now we can, by keeping track of the "interior mesh" (which defaults to `this`) for any given mesh.

Unit tests here fail (when running on a few ranks) without the preceding commits' fixes.

### DIFF
--- a/tests/mesh/boundary_mesh.C
+++ b/tests/mesh/boundary_mesh.C
@@ -37,9 +37,10 @@ protected:
   std::unique_ptr<UnstructuredMesh> _left_boundary_mesh;
   std::unique_ptr<UnstructuredMesh> _internal_boundary_mesh;
 
-  // We create an internal sideset ID that does not conflict with
+  // We create internal sideset IDs that do not conflict with
   // sidesets 0-3 that get created by build_square().
-  static constexpr boundary_id_type bid = 5;
+  static constexpr boundary_id_type bid1 = 5;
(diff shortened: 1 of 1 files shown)


YaqiWang:
Is this related the assertion I hit with mesh refinement when using `BoundaryInfo::sync`?

roystgnr:
The first three commits fix and add test coverage for subsequent serialization of a distributed `BoundaryInfo::sync()` target mesh, but the second three now fix and add test coverage for the bug you hit with `BoundaryInfo::sync()` of a refined source mesh.

roystgnr:
This isn't as sweeping as I'd hoped it would be, because as soon as we mix interior and exterior sidesets we run into the "libMesh doesn't properly support non-manifold connections" problem for which I've never been able to come up with an efficient fix, but it does fix a bunch of use cases and add much better test coverage.

roystgnr:
The test coverage is *so* much better, in fact, that it's failing on the devel->master merge.  I can replicate it and am looking into it.

PR #4183, opened 2025-06-02
Remove redundant prepare_send_list() calls

roystgnr:
Thanks to @lindsayad for catching one of these.


lindsayad:
Ah I see that the MOOSE test is biting others as well. I have a change at https://github.com/idaholab/moose/pull/30658

roystgnr:
That new exodiff is consistently the only failure, so this should be good to go.

PR #4184, opened 2025-06-03
Adaptivity improvements

lindsayad:
- Add support for adaptivity with static condensation. Follow-on from #4140 and in response to https://github.com/idaholab/moose/pull/27188#pullrequestreview-2879037069
- Clean refinement flags after projecting vectors. This ensures that if there is no mesh refinement before another
call to project vectors that vector projection will work. Refs idaholab/moose#30742

### DIFF
--- a/include/numerics/static_condensation_dof_map.h
+++ b/include/numerics/static_condensation_dof_map.h
@@ -120,16 +120,55 @@ private:
    * be uncondensed
    * @param full_dof_number The dof id in the full (uncondensed + condensed) system
    * @param involved_in_constraints Whether the \p full_dof_number is involved in constraints
+   * @param uncondensed_local_to_global_map A map from uncondensed full dof numbering to local
+   * (element) numbering for a given element
+   * @param local_uncondensed_dofs_set Set of local uncondensed degrees of freedom (numbered in the
+   * full system)
+   * @param nonlocal_uncondensed_dofs Map from processor IDs to sets of nonlocal uncondensed degrees
+   * of freedom (numbered in the full system)
+   * @param elem_uncondensed_dofs Used for temporary storage of uncondensed degrees of freedom
+   * active on an element
+   * @param uncondensed_local_dof_number Counter for number of uncondensed dofs on an element
+   * @param constraint_dofs Set of degrees of freedom numbered in the full system involved in
+   * constraints
    */
-  void add_uncondensed_dof_plus_constraint_dofs(dof_id_type full_dof_number,
-                                                bool involved_in_constraints);
+  void add_uncondensed_dof_plus_constraint_dofs(
+      dof_id_type full_dof_number,
+      bool involved_in_constraints,
+      std::unordered_map<dof_id_type, dof_id_type> & uncondensed_global_to_local_map,
+      std::unordered_set<dof_id_type> & local_uncondensed_dofs_set,
+      std::unordered_map<processor_id_type, std::unordered_set<dof_id_type>> &
+          nonlocal_uncondensed_dofs,
+      std::vector<dof_id_type> & elem_uncondensed_dofs,
+      dof_id_type & uncondensed_local_dof_number,
+      std::unordered_set<dof_id_type> & constraint_dofs);
 
   /**
    * Add an uncondensed dof
    * @param full_dof_number The dof id in the full (uncondensed + condensed) system
    * @param involved_in_constraints Whether the \p full_dof_number is involved in constraints
+   * @param uncondensed_local_to_global_map A map from uncondensed full dof numbering to local
+   * (element) numbering for a given element
+   * @param local_uncondensed_dofs_set Set of local uncondensed degrees of freedom (numbered in the
+   * full system)
+   * @param nonlocal_uncondensed_dofs Map from processor IDs to sets of nonlocal uncondensed degrees
+   * of freedom (numbered in the full system)
+   * @param elem_uncondensed_dofs Used for temporary storage of uncondensed degrees of freedom
+   * active on an element
+   * @param uncondensed_local_dof_number Counter for number of uncondensed dofs on an element
+   * @param constraint_dofs Set of degrees of freedom numbered in the full system involved in
+   * constraints
    */
-  void add_uncondensed_dof(dof_id_type full_dof_number, bool involved_in_constraints);
+  void add_uncondensed_dof(
+      dof_id_type full_dof_number,
+      bool involved_in_constraints,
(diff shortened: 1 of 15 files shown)


src/numerics/static_condensation_dof_map.C
```diff
@@ -42,16 +42,57 @@ StaticCondensationDofMap::StaticCondensationDofMap(MeshBase & mesh,
...
+  const bool is_constrained = it != full_dof_constraints.end();
+  involved_in_constraints = involved_in_constraints || is_constrained;
+
+  this->add_uncondensed_dof(full_dof_number, involved_in_constraints);
+  if (is_constrained)
+    for (const auto [full_unconstrained_dof, weight] : it->second)
+      {
+        libmesh_ignore(weight);
+        // Our constraining dofs may themselves be constrained
+        this->add_uncondensed_dof_plus_constraint_dofs(full_unconstrained_dof,
```
lindsayad:
```suggestion
    for (const auto [full_constraining_dof, weight] : it->second)
      {
        libmesh_ignore(weight);
        // Our constraining dofs may themselves be constrained
        this->add_uncondensed_dof_plus_constraint_dofs(full_constraining_dof,
```

lindsayad:
MetaPhysicL failure again 😦

include/numerics/static_condensation_dof_map.h
```diff
@@ -166,6 +189,32 @@ class StaticCondensationDofMap : public DofMapBase
 
   /// Number of off-diagonal nonzeros per row in the reduced system
   std::vector<dof_id_type> _reduced_noz;
+
+  /// Set of local uncondensed degrees of freedom (numbered in the full system). This data member is
+  /// used only during the reinit phase and is cleared at its conclusion
```
roystgnr: It looks like you're also using it in the constructor, but in each case wouldn't it make sense to have this be a local variable rather than a member variable?  You'd need to pass it to `add_uncondensed_dof()` as another argument, but then it's a little easier to follow program flow.

include/numerics/static_condensation_dof_map.h
```diff
@@ -166,6 +189,32 @@ class StaticCondensationDofMap : public DofMapBase
...
+  std::unordered_set<dof_id_type> _local_uncondensed_dofs_set;
+
+  /// Map from processor IDs to sets of nonlocal uncondensed degrees of freedom (numbered in the
+  /// fully system). This data member is used only during the reinit phase and is cleared at its
+  /// conclusion
+  std::unordered_map<processor_id_type, std::unordered_set<dof_id_type>> _nonlocal_uncondensed_dofs;
+
+  /// Used for temporary storage of uncondensed degrees of freedom active on an element
+  std::vector<dof_id_type> _elem_uncondensed_dofs;
```
roystgnr: Likewise for these two.  Though I admit we're now up to three more arguments, and my suggestion is getting less compelling, it just feels confusing to have temporary storage saved as members.

roystgnr (approved):
I think I'm happy with this, though I'd love for @jwpeterson to take a look too.

include/numerics/static_condensation_dof_map.h
lindsayad: So you would prefer this?
```c++
void StaticCondensationDofMap::add_uncondensed_dof(
    const dof_id_type full_dof_number,
    const bool involved_in_constraints,
    std::unordered_map<dof_id_type, dof_id_type> & uncondensed_global_to_local_map,
    std::unordered_set<dof_id_type> & local_uncondensed_dofs_set,
    std::unordered_map<processor_id_type, std::unordered_set<dof_id_type>> &
        nonlocal_uncondensed_dofs,
    std::vector<dof_id_type> & elem_uncondensed_dofs,
    dof_id_type & uncondensed_local_dof_number,
    std::unordered_set<dof_id_type> constraint_dofs)
```

lindsayad: https://github.com/libMesh/libmesh/pull/4184/commits/c2bd8d6a9558cff4496f553bd543aecdfb818a65 uses local variables

roystgnr: Yeah, honestly.  It feels like that "scratch space" could be members of a nested class, even?

I wouldn't go that far now, though.  If this was a public API I'd be anxious over getting the best design before we exposed it to users and had to support it, but with a private API we can redo at any time I'm just expressing mild preferences.

src/numerics/static_condensation_dof_map.C
```diff
@@ -42,16 +42,86 @@ StaticCondensationDofMap::StaticCondensationDofMap(MeshBase & mesh,
...
+  if (is_constrained)
+    for (const auto [full_constraining_dof, weight] : it->second)
+      {
+        libmesh_ignore(weight);
+        // Our constraining dofs may themselves be constrained
+        this->add_uncondensed_dof_plus_constraint_dofs(full_constraining_dof,
+                                                       /*involved_in_constraints=*/true,
+                                                       uncondensed_global_to_local_map,
+                                                       local_uncondensed_dofs_set,
+
```
jwpeterson: Is clang format inserting these blank lines for some reason?

src/numerics/static_condensation_dof_map.C
```diff
@@ -42,16 +42,86 @@ StaticCondensationDofMap::StaticCondensationDofMap(MeshBase & mesh,
...
+  const auto & full_dof_constraints = _dof_map.get_dof_constraints();
+  auto it = full_dof_constraints.find(full_dof_number);
+  const bool is_constrained = it != full_dof_constraints.end();
+  involved_in_constraints = involved_in_constraints || is_constrained;
+
+  this->add_uncondensed_dof(full_dof_number,
+                            involved_in_constraints,
+                            uncondensed_global_to_local_map,
+                            local_uncondensed_dofs_set,
+
```
jwpeterson: Another blank line

jwpeterson (approved):

roystgnr:
We can merge without kicking that test - MetaPhysicL failure again.  I'm so baffled by that one.  `LIBMESH_RUN` doesn't affect MetaPhysicl `make check` and we don't set `METAPHYSICL_RUN` so we shouldn't be getting anything non-deterministic from MPI, and none of the unit tests there respect `--n_threads=` either.

src/numerics/static_condensation_dof_map.C
lindsayad: no I think that was just me somehow

lindsayad:
I force pushed the last commit just with the blank line removals that @jwpeterson pointed out. I'll merge if/once this passes CI

PR #4187, opened 2025-06-10
Fixes for static condensation dof map

lindsayad:
Refs #4140


roystgnr:
Test failure (in MetaPhysicL) is unrelated.  Though it is something I've seen once or twice before, and I wish I understood it or at least could reproduce it...

PR #4188, opened 2025-06-11
Eliminate obsolete use of parent() to store interior_parent(), deprecated side/edge proxy option

roystgnr:
This should fix some issues @YaqiWang ran into when syncing a 1D mesh to a NodeElem boundary mesh, and while I was digging into that I got rid of the proxy code that's been redundant for a decade and deprecated since 2021 (and that was still using the old behavior for backwards compatibility).


roystgnr:
I should promise that in the future I'll actually wait for all the unit tests to finish running locally on a branch before I make a PR, no matter how confident I am that everything's correct.

I won't, but I should.

roystgnr:
This will need to wait on https://github.com/idaholab/moose/pull/30719

roystgnr:
Rebased on master to try again

roystgnr:
Okay - that last refactor shaved off another 60 lines of code, fixed the distributed netgen regression for me, and probably fixed a bunch of never-yet-triggered bugs too.  If this passes CI and I don't see any screams from @jwpeterson or @lindsayad tomorrow, I'll merge in the afternoon.

Issue #4082, opened 2025-02-18
Revamp the variational mesh smoother


pbehne:
The `VariationalMeshSmoother` does not have any test coverage and has fallen out of date. This issue delineates the tasks required to bring it up to date. Tasks will be edited/added as needed.

- [ ] Get the (commented out) `MeshSmootherTest::testVariationalQuad` and `MeshSmootherTest::testVariationalTri` tests running and passing.
- [ ] Add test coverage for other element types.
- [ ] Add test coverage for the adaptivity feature of the smoother.
- [ ] Add test coverage for metric types that are not uniform.
- [ ] Add test coverage for the `area_of_interest` feature of the smoother.
- [ ] General code cleanup/consolidation/refactor to use libMesh APIs instead of hard-coded methods
  - [ ] Replace hard-coded quadrature sets with libMesh-provided quadrature sets.
  - [ ] Consolidate all the branched logic in a more concise way.
  - [ ] Use libMesh-provided solvers instead of hard coded solvers.
- [ ] Extend sliding boundary node logic to 3D meshes.
  - [ ] Identify sliding boundary nodes in 3D meshes.
  - [ ] Extend boundary node correction to 3D meshes.
- [ ] Add/Update test coverage for 3D meshes.

PR #4191, opened 2025-06-12
Variational smoother now uses libMesh solver and constraint system

pbehne:
Previously, the VariationalMeshSmoother set up and solved the
optimization problem using logic branches for each element type and a
separate solver than what libMesh provides.
    
This PR introduces two new objects to incorporate the variational
smoother's solve process into the rest of libMesh: the FEMSystem-derived
VariationalSmootherSystem and the Constraint-derived
VariationalSmootherConstraint.
    
The VariationalSmootherSystem computes the residual and analytic
Jacobian for the optimization solve. The analytic Jacobian was verified
to be correct when compared to finite difference estimations.
    
The VariationalSmootherConstraint currently constrains boundary nodes
and, if requested, subdomain boundary nodes to not move during the
smoothing process. Less strict sliding boundary node constraints will be
introduced in a future PR.

These changes allowed for a massive cleanup of the VariationalMeshSmootherClass.

Ref. #4082

### DIFF
--- a/src/mesh/mesh_smoother_vsmoother.C
+++ b/src/mesh/mesh_smoother_vsmoother.C
@@ -64,13 +64,33 @@ void VariationalMeshSmoother::smooth(unsigned int)
   if (_mesh.elem_dimensions().size() > 1)
     libmesh_not_implemented_msg("Meshes containing elements of differing dimension are not yet supported.");
 
+  /*
+  Ideally we'd want to update _mesh directly even if it already has an
+  EquationSystems attached and doing work on it ... and that's thwarted by the
+  problems:
+
+  1. We can't easily tell if there's already an EquationSystems attached.
+  2. We can't attach a second EquationSystems safely (if it already has a system 0)
+     because the DoF indexing will need to be overwritten by our system.
+  3. The destructor of es won't even clean up after itself (we generally expect
+     a mesh to go unused after its EquationSystems is destroyed), much less know
+     how to restore anything from a previous EquationSystems.
+
+  To avoid these issues, we'll just construct a new DistributedMesh mesh_copy
+  from _mesh, then do the solve on mesh_copy, then copy its node locations back
+  to _mesh after the solve is done. That'll be slightly less memory-efficient
+  and somewhat more CPU-efficient in the case where _mesh is serial, though
+  it'll be significantly less memory-efficient when _mesh is already distributed,
+  but either way the robustness is probably worth it.
+  */
+
   // Create a new mesh, EquationSystems, and System
   DistributedMesh mesh_copy(_mesh);
   EquationSystems es(mesh_copy);
   VariationalSmootherSystem & sys = es.add_system<VariationalSmootherSystem>("variational_smoother_system");
 
   // Set this to something > 0 to add more quadrature points than the default
-  // rule that integrates order 2 * nq_points + 1 polynomials exactly.
+  // rule that integrates order 2 * fe_order + 1 polynomials exactly.
   // Using higher quadrature orders has not had a significant effect on observed solutions.
   //sys.extra_quadrature_order = 0;
 


roystgnr:
Ooh - we never talked about the parallel case.  That's going to be slightly harder to fix than bad whitespace or a shadowing error.

For now, what we want to do is copy the variational smoother solution vector (which is `PARALLEL`, with a `GHOSTED` `current_local_solution` copy) into a `SERIAL` vector before using the results from there.  That's also the long-term solution for serial (including `ReplicatedMesh`) meshes, but at some point we might also want to write a code path that sets up the proper ghosting (algebraic ghosting will need to match geometric ghosting, even if the latter has been expanded) on distributed meshes.

include/systems/variational_smoother_constraint.h
```diff
@@ -0,0 +1,66 @@
...
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+#ifndef LIBMESH_VARIATIONAL_SMOOTHER_CONSTRAINT_H
+#define LIBMESH_VARIATIONAL_SMOOTHER_CONSTRAINT_H
+
+// Local Includes
+#include "libmesh/system.h"
+#include <libmesh/dof_map.h>
```
roystgnr:
```suggestion
#include "libmesh/dof_map.h"
```

It would take some *weird* compiler flags for this to make a difference, but let's be consistent.

src/systems/variational_smoother_system.C
```diff
@@ -0,0 +1,467 @@
...
+}
+
+void VariationalSmootherSystem::init_data ()
+{
+  auto & mesh = this->get_mesh();
+  // Add a variable for each dimension of the mesh
+  // "r0" for x, "r1" for y, "r2" for z
+  for (const auto & d : make_range(mesh.mesh_dimension()))
+    this->add_variable ("r" + std::to_string(d), static_cast<Order>(_fe_order),
+                        Utility::string_to_enum<FEFamily>(_fe_family));
```
roystgnr: Let's just leave `LAGRANGE` implicit here, and query `mesh.elem_default_orders()` to get the FE order?  (if the set is a singleton use that order; if not then scream and die).  The `_fe_family`/`_fe_order` stuff only made sense in systems where we can let the user control those.

include/systems/variational_smoother_system.h
```diff
@@ -0,0 +1,119 @@
...
+
+  std::string & fe_family() { return _fe_family; }
+  unsigned int & fe_order() { return _fe_order; }
+
+  // We want to be able to project functions based on *other* systems'
+  // values.  For that we need not only a FEMFunction but also a
+  // reference to the system where it applies and a separate context
+  // object (or multiple separate context objects, in the threaded
+  // case) for that system.
+  libMesh::System * input_system;
```
roystgnr: Get rid of this?  It's just left over from `L2System`, right?

include/systems/variational_smoother_system.h
```diff
@@ -0,0 +1,119 @@
...
+
+protected:
+  std::map<libMesh::FEMContext *, std::unique_ptr<libMesh::FEMContext>>
+    input_contexts;
```
roystgnr: Likewise we don't seem to be using this.

include/systems/variational_smoother_system.h
```diff
@@ -0,0 +1,119 @@
...
+                                        libMesh::DiffContext & context) override;
+
+  /* Computes the element reference volume used in the dilation metric
+   * The reference value is set to the averaged value of all elements' average |J|.
+   */
+  void compute_element_reference_volume();
+
+  /// The FE type to use
+  std::string _fe_family;
+  unsigned int _fe_order;
```
roystgnr: And these aren't ever going to be changed, so we don't need them as variables.

src/mesh/mesh_smoother_vsmoother.C
```diff
@@ -44,4470 +49,62 @@ namespace libMesh
...
-          } // end dim==3
-#endif // LIBMESH_DIM > 2
-      }
-
-  // Done with the metric file
-  metric_file.close();
-
-  std::ofstream grid_file(grid.c_str());
-  libmesh_error_msg_if(!grid_file.good(), "Error opening file: " << grid);
+  Real dilation_weight = _dilation_weight;
```
roystgnr: Can we not just directly use `_dilation_weight` later?

src/mesh/mesh_smoother_vsmoother.C
```diff
@@ -44,4470 +49,62 @@ namespace libMesh
...
 
-  grid_file << _dim << "\n" << _n_nodes << "\n" << Ncells << "\n0" << std::endl;
+  // Create a new mesh, EquationSystems, and System
+  EquationSystems es(_mesh);
```
roystgnr: Ideally we'd want this to work even if `_mesh` already has an `EquationSystems` attached and doing work on it ... and that's thwarted by the problems:

1. We can't easily tell if there's already an `EquationSystems` attached.
2. We can't attach a second `EquationSystems` safely (if it already has a system 0) because the DoF indexing will need to be overwritten by our system.
3. The destructor of `es` won't even clean up after itself (we generally expect a mesh to go unused after its `EquationSystems` is destroyed), much less know how to restore anything from a previous `EquationSystems`.

I'm thinking that what we want to do here is just construct a new `DistributedMesh mesh_copy` from `_mesh`, then do the solve on `mesh_copy`, then copy its node locations back to `_mesh` after the solve is done.  That'll be slightly less memory-efficient and somewhat more CPU-efficient in the case where `_mesh` is serial, though it'll be significantly less memory-efficient when `_mesh` is already distributed, but either way I think the robustness is probably worth it.

src/mesh/mesh_smoother_vsmoother.C
```diff
@@ -44,4470 +49,62 @@ namespace libMesh
...
+  VariationalSmootherSystem & sys = es.add_system<VariationalSmootherSystem>("variational_smoother_system");
 
-  // Use scientific notation with 6 digits
-  grid_file << std::scientific << std::setprecision(6);
+  // Set this to something > 0 to add more quadrature points than the default
+  // rule that integrates order 2 * nq_points + 1 polynomials exactly.
+  sys.extra_quadrature_order = 0;
```
roystgnr: You could comment this out since 0 is the default; also maybe comment that we're not seeing major differences in results at higher quadrature orders.

src/mesh/mesh_smoother_vsmoother.C
```diff
@@ -44,4470 +49,62 @@ namespace libMesh
...
-            // quad
-            grid_file << cells[i][0] << " " << cells[i][1] << " " << cells[i][2] << " -1 -1 -1 0" << std::endl;
-            grid_file << cells[i][1] << " " << cells[i][3] << " " << cells[i][0] << " -1 -1 -1 0" << std::endl;
-            grid_file << cells[i][3] << " " << cells[i][2] << " " << cells[i][1] << " -1 -1 -1 0" << std::endl;
-            grid_file << cells[i][2] << " " << cells[i][0] << " " << cells[i][3] << " -1 -1 -1 0" << std::endl;
-          }
+  DiffSolver & solver = *(sys.time_solver->diff_solver().get());
+  sys.time_solver->diff_solver()->relative_residual_tolerance = 1e-10;
+  solver.quiet = false;
+  solver.verbose = true;
```
roystgnr, line 123: Shouldn't we be quiet by default?

src/mesh/mesh_smoother_vsmoother.C
```diff
@@ -44,4470 +49,62 @@ namespace libMesh
...
-        if ((_dim == 2) && (nvert == 4))
-          {
-            // quad
-            grid_file << cells[i][0] << " " << cells[i][1] << " " << cells[i][2] << " -1 -1 -1 0" << std::endl;
-            grid_file << cells[i][1] << " " << cells[i][3] << " " << cells[i][0] << " -1 -1 -1 0" << std::endl;
-            grid_file << cells[i][3] << " " << cells[i][2] << " " << cells[i][1] << " -1 -1 -1 0" << std::endl;
-            grid_file << cells[i][2] << " " << cells[i][0] << " " << cells[i][3] << " -1 -1 -1 0" << std::endl;
-          }
+  DiffSolver & solver = *(sys.time_solver->diff_solver().get());
+  sys.time_solver->diff_solver()->relative_residual_tolerance = 1e-10;
```
roystgnr: Let's default these to `TOLERANCE*std::sqrt(TOLERANCE)` (1e-9 with double-precision) or something, to be independent of compile-time precision settings.

src/systems/variational_smoother_system.C
```diff
@@ -0,0 +1,467 @@
...
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+#include "libmesh/variational_smoother_system.h"
+
+#include "libmesh/elem.h"
+#include "libmesh/fe_base.h"
+#include "libmesh/fe_interface.h"
+#include "libmesh/fem_context.h"
+#include "libmesh/getpot.h"
```
roystgnr: We shouldn't need the getpot header here.

src/systems/variational_smoother_system.C
```diff
@@ -0,0 +1,467 @@
...
+      // _epsilon_squared to zero to get chi(det(S)) = det(S)
+      const Real epsilon_sq = det > 0. ? 0. : _epsilon_squared;
+      const Real chi = 0.5 * (det + std::sqrt(epsilon_sq + det_sq));
+      const Real chi_sq = chi * chi;
+      const Real chi_cube = chi_sq * chi;
+      const Real sqrt_term = std::sqrt(epsilon_sq + det_sq);
+      // dchi(x) / dx
+      const Real chi_prime = 0.5 * (1. + det / sqrt_term);
+      // dchi(det(S) / dS
+      RealTensor dchi_dS = chi_prime * det * S_inv_T;
```
roystgnr: This could be `const` too, right?  Likewise with `dmu_dS` below?

Normally when I see something that isn't const but could be I figure the programmer was just being lazy, but you're being so careful above that it makes me wonder if I'm missing a subtlety here.

roystgnr (changes requested):

src/mesh/mesh_smoother_vsmoother.C
pbehne: I understand what `relative_residual_tolerance` is, but what is `relative_step_tolerance`, and do we need to set it?

With the `relative_residual_tolerance=1e-10`:
```
MeshSmootherTest::testVariationalQuad.Assembling the System
Nonlinear Residual: 47.5325
Linear solve starting, tolerance 1e-12
Linear solve finished, step 7, residual 3.49955e-16
Trying full Newton step
  Current Residual: 14.2283
  Nonlinear step: |du|/|u| = 0.0424404, |du| = 0.183073
Assembling the System
Nonlinear Residual: 14.2283
Linear solve starting, tolerance 1e-12
Linear solve finished, step 7, residual 7.46886e-18
Trying full Newton step
  Current Residual: 1.41691
  Nonlinear step: |du|/|u| = 0.0203602, |du| = 0.0881424
Assembling the System
Nonlinear Residual: 1.41691
Linear solve starting, tolerance 1e-12
Linear solve finished, step 7, residual 1.71678e-18
Trying full Newton step
  Current Residual: 0.00915126
  Nonlinear step: |du|/|u| = 0.00224677, |du| = 0.00972878
Assembling the System
Nonlinear Residual: 0.00915126
Linear solve starting, tolerance 1e-12
Linear solve finished, step 7, residual 1.44068e-20
Trying full Newton step
  Current Residual: 3.21233e-07
  Nonlinear step: |du|/|u| = 1.46944e-05, |du| = 6.36284e-05
Assembling the System
Nonlinear Residual: 3.21233e-07
Linear solve starting, tolerance 1e-12
Linear solve finished, step 8, residual 7.28137e-22
Trying full Newton step
  Current Residual: 8.33185e-15
```

Taking it out and letting it use the default:
```
MeshSmootherTest::testVariationalQuad.Assembling the System
Nonlinear Residual: 47.5325
Linear solve starting, tolerance 1e-12
Linear solve finished, step 7, residual 3.49955e-16
Trying full Newton step
  Current Residual: 14.2283
  Nonlinear step: |du|/|u| = 0.0424404, |du| = 0.183073
Assembling the System
Nonlinear Residual: 14.2283
Linear solve starting, tolerance 1e-12
Linear solve finished, step 7, residual 7.46886e-18
Trying full Newton step
  Current Residual: 1.41691
  Nonlinear step: |du|/|u| = 0.0203602, |du| = 0.0881424
Assembling the System
Nonlinear Residual: 1.41691
Linear solve starting, tolerance 1e-12
Linear solve finished, step 7, residual 1.71678e-18
Trying full Newton step
  Current Residual: 0.00915126
  Nonlinear step: |du|/|u| = 0.00224677, |du| = 0.00972878
Assembling the System
Nonlinear Residual: 0.00915126
Linear solve starting, tolerance 1e-12
Linear solve finished, step 7, residual 1.44068e-20
Trying full Newton step
  Current Residual: 3.21233e-07
  Nonlinear step: |du|/|u| = 1.46944e-05, |du| = 6.36284e-05
Assembling the System
Nonlinear Residual: 3.21233e-07
Linear solve starting, tolerance 1e-12
Linear solve finished, step 8, residual 7.28137e-22
Trying full Newton step
  Current Residual: 8.33185e-15
  Nonlinear step: |du|/|u| = 5.16679e-10, |du| = 2.23728e-09
Assembling the System
Nonlinear Residual: 8.33185e-15
Linear solve starting, tolerance 1e-12
Linear solve finished, step 13, residual 3.19296e-32
Trying full Newton step
  Current Residual: 1.04148e-14
  Shrinking Newton step to 0.5
  Current Residual: 1.04148e-14
  Shrinking Newton step to 0.25
  Current Residual: 1.04148e-14
  Shrinking Newton step to 0.125
  Current Residual: 1.04148e-14
  Shrinking Newton step to 0.0625
  Current Residual: 1.04148e-14
  Shrinking Newton step to 0.03125
  Current Residual: 1.04148e-14
  Shrinking Newton step to 0.015625
  Current Residual: 1.04148e-14
  Shrinking Newton step to 0.0078125
  Current Residual: 1.04148e-14
  Shrinking Newton step to 0.00390625
  Current Residual: 1.04148e-14
  Shrinking Newton step to 0.00195312
  Current Residual: 1.04148e-14
  Shrinking Newton step to 0.000976562
  Current Residual: 1.04148e-14
  Shrinking Newton step to 0.000488281
  Current Residual: 1.04148e-14
  Shrinking Newton step to 0.000244141
  Current Residual: 1.04148e-14
  Shrinking Newton step to 0.00012207
  Current Residual: 1.04148e-14
  Shrinking Newton step to 6.10352e-05
  Current Residual: 1.04148e-14
  Shrinking Newton step to 3.05176e-05
  Current Residual: 1.04148e-14
  Shrinking Newton step to 1.52588e-05
  Current Residual: 1.04148e-14
Inexact Newton step FAILED at step 5
 : errorE
```

src/systems/variational_smoother_system.C
pbehne: Nah, I just missed the const. Good catch!

src/mesh/mesh_smoother_vsmoother.C
```diff
@@ -63,15 +64,15 @@ void VariationalMeshSmoother::smooth(unsigned int)
   if (_mesh.elem_dimensions().size() > 1)
     libmesh_not_implemented_msg("Meshes containing elements of differing dimension are not yet supported.");
 
-  Real dilation_weight = _dilation_weight;
-
   // Create a new mesh, EquationSystems, and System
```
roystgnr, line 87: Heh; last review I missed that this comment was an atavism, but now we're actually creating a new mesh.

We should explain *why* we're creating a new mesh rather than solving directly on the input, though; the desire to be compatible even with the possibility of a preexisting `EquationSystems` took a long time to occur to me.

src/mesh/mesh_smoother_vsmoother.C
```diff
@@ -44,4470 +50,73 @@ namespace libMesh
...
-
-  for (dof_id_type i=0; i<_n_nodes; i++)
-    {
-      if ((mask[i]==1) || (mask[i]==2))
-        maskf[i] = 1;
-      else
-        maskf[i] = 0;
-    }
+  // Set this to something > 0 to add more quadrature points than the default
+  // rule that integrates order 2 * nq_points + 1 polynomials exactly.
```
roystgnr:
```suggestion
  // rule that integrates order 2 * fe_order + 1 polynomials exactly.
```

The version with `nq_points` is only valid in 1D

PR #4192, opened 2025-06-17
Remove redundant const_cast/release/etc

roystgnr:
There's a unique_ptr constructor that works just fine for the implicit conversion here.


jwpeterson (approved):
Beautiful code cleanup

PR #4193, opened 2025-06-17
Update doxygen for max_node_id/max_elem_id

lindsayad:
Should help user code trying to select node/element IDs for new nodes/elements


roystgnr:
Well, someone might want to look at `problems/eigen_problem/arraykernels.eigen/two_variables_precond_include_eigen_kernels` for race conditions, but I'm pretty sure the doc change didn't introduce one.

PR #4194, opened 2025-06-18
Document somewhat odd behavior of `ReplicatedMesh::renumber_nodes_and_elements`

lindsayad:
This is better than looking up old slack history or at least gives another possibility of discovering this information

### DIFF
--- a/include/mesh/replicated_mesh.h
+++ b/include/mesh/replicated_mesh.h
@@ -127,7 +127,12 @@ public:
   virtual void clear_elems() override;
 
   /**
-   * Remove nullptr elements from arrays
+   * @copydoc MeshBase::renumber_nodes_and_elements()
+   *
+   * Note that regardless of whether \p _skip_renumber_nodes_and_elements is true, this method will
+   * always remove nullptr nodes and elements from arrays and consequently will always renumber. So in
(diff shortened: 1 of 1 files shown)


roystgnr (approved):

include/mesh/replicated_mesh.h
```diff
@@ -127,7 +127,12 @@ class ReplicatedMesh : public UnstructuredMesh
   virtual void clear_elems() override;
 
   /**
-   * Remove nullptr elements from arrays
+   * @copydoc MeshBase::renumber_nodes_and_elements()
+   *
+   * Note that regardless of whether \p _skip_renumber_nodes_and_elements is true, this method will
+   * always remove nullptr nodes elements from arrays and consequently will always renumber. So in
```
jwpeterson: typo? Did you mean "remove nullptr nodes *and* elements"?

lindsayad:
```suggestion
   * always remove nullptr nodes and elements from arrays and consequently will always renumber. So in
```

jwpeterson (approved):
I don't think it's necessary to wait for all the testing to finish unless you really want to.

PR #4198, opened 2025-06-24
Fix missing FE pre-request in VariationalSmootherSystem::compute_element_reference_volume()

jwpeterson:
Otherwise, in --disable-deprecated builds, each of the MeshSmootherTest unit tests throws an error.

Refs #4191

Example of the error fixed by this PR:
```
1) test: MeshSmootherTest::testVariationalQuad (E) 
uncaught exception of type std::exception (or derived).
- 
Stack frames: 36
0: libMesh::print_trace(std::ostream&)
1: libMesh::MacroFunctions::report_error(char const*, int, char const*, char const*, std::ostream&)
2: libMesh::FEGenericBase<double>::determine_calculations()
3: libMesh::FE<2u, (libMesh::FEFamily)0>::reinit(libMesh::Elem const*, std::vector<libMesh::Point, std::allocator<libMesh::Point> > const*, std::vector<double, std::allocator<double> > const*)
4: libMesh::FEMContext::elem_fe_reinit(std::vector<libMesh::Point, std::allocator<libMesh::Point> > const*)
5: libMesh::VariationalSmootherSystem::compute_element_reference_volume()
6: libMesh::VariationalSmootherSystem::init_data()
7: libMesh::System::reinit_mesh()
8: libMesh::EquationSystems::reinit_mesh()
9: libMesh::EquationSystems::init()
10: libMesh::VariationalMeshSmoother::smooth(unsigned int)
11: libMesh::VariationalMeshSmoother::smooth()
```

PR #4200, opened 2025-06-25
get_nothing() from FE objects we want nothing from

roystgnr:
*This* fixes the --disable-deprecated errors for me

### DIFF
--- a/src/systems/variational_smoother_system.C
+++ b/src/systems/variational_smoother_system.C
@@ -102,7 +102,7 @@ void VariationalSmootherSystem::compute_element_reference_volume()
 
   Real elem_averaged_det_J_sum = 0.;
 
-  // Must pre-request JxW before reinit() for efficiency in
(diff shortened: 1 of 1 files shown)


jwpeterson:
For reference, the issue here was that calling `FEMap::get_JxW()` directly, rather than calling the "forwarding" function `FE::get_JxW()`, sets a different flag. The latter sets the `FE::calculate_map` flag, which is checked in the `determine_calculations()` assertion, while the former sets the `FEMap::calculate_dxyz` flag, which is not (currently) checked. This seems to be a bit of an inconsistency, so it is probably worth updating the `determine_calculations()` checks to also take into account `FEMap` pre-requests.

jwpeterson:
Refs #4198

roystgnr:
We're past all the unit tests that exercise this; merging now so we can get master going again ASAP.

PR #4203, opened 2025-06-26
Have MeshGeneration functions take MeshBase

lindsayad:
This would allow avoiding the static cast in user code in idaholab/moose#30874


roystgnr:
IMHO we want the `UnstructuredMesh` requirement here.  I know, in practice every UnstructuredMesh is a MeshBase and vice-versa, but in my long-lost dream of doing things like `CartesianMesh` (elements created on the fly, storage O(1) instead of O(N)), there's just no way to run a smoother or a generator on every possible MeshBase subclass.

lindsayad:
Surely we could relax the requirement on some of these apis, like `build_square`?

roystgnr:
If I hand a hypothetical `PolarMesh` to `build_square`, what's it do with it?

I'd say that the right thing to do here is to have everything that really wants an arbitrary unstructured mesh hold on to a reference (/smart pointer, whatever) to an `UnstructuredMesh` rather than to a `MeshBase`.  The only big catch I see with that is that right now `clone()` always returns `unique_ptr<MeshBase>` even when `UnstructuredMesh` and subclasses should be able to return `unique_ptr<UnstructuredMesh>`, and that ought to be fixable.

lindsayad:
> and that ought to be fixable

How so? Can't do it with unique pointers. Can only leverage covariance with raw pointers/references

lindsayad:
anyway seems like I should close this PR

roystgnr:
I admit my first idea was to just provide an `unstructured_clone()` for `UnstructuredMesh` users.

For a slightly-less ugly option: [the first idiom here](https://www.fluentcpp.com/2017/09/12/how-to-return-a-smart-pointer-and-use-covariance/).  We make a virtual `clone_impl()` returning raw pointers to leverage covariance, and we make our `clone()` non-virtual to wrap the pointers in the properly typed smart pointer at each level.  It'll mean a few raw `new` that we can't get rid of, but at least they'll only be in a private method and their outputs will be wrapped immediately.

PR #4209, opened 2025-07-08
Avoid out of bounds access when getting address

nmnobre:
Courtesy of the newest versions of gcc/libstdc++ which are now more assertive by default, here's an example: https://godbolt.org/z/3YE5a4zed.

### DIFF
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -1012,16 +1012,17 @@ void ExodusII_IO_Helper::read_bex_cv_blocks()
               for (auto i : IntRange<std::size_t>(0, bex_num_dense_cv_blocks))
                 {
                   bex_dense_constraint_vecs[i].resize(bex_dense_cv_info[2*i]);
+                  const int vecsize = bex_dense_cv_info[2*i+1];
                   for (auto & vec : bex_dense_constraint_vecs[i])
                     {
-                      const int vecsize = bex_dense_cv_info[2*i+1];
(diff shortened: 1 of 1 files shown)


jwpeterson:
https://stackoverflow.com/questions/45724767/does-stdnext-check-if-we-are-already-at-the-end-of-the-container

jwpeterson:
Example (https://godbolt.org/z/KoxfK4hY4) confirming that `std::next` will return an invalid iterator ( != end()) if you advance it too far.

nmnobre:
> Is the idea that std::next(it, bignum) will never return an iterator past the end() of the container that it comes from, no matter how large bignum is?

Not really. I wasn't questioning code correctness.

> and whereas before that would have been (possibly) a segfault, now it's going to (again, possibly) be a more silent error?

Potentially... Again, I wasn't really thinking of the case where the code might actually be wrong :)

>  I guess we might want to libmesh_assert() (via the iterator returned by std::copy) that the right amount of data was copied?

I've added an assertion as suggested.

roystgnr:
So the issue is that, although I *want* an "end" iterator in some cases, it's not legal to obtain a reference to the end position via `operator[]` even if all you're going to do is take a pointer?  Thanks for the fix and the review, guys!

nmnobre:
> So the issue is that, although I _want_ an "end" iterator in some cases, it's not legal to obtain a reference to the end position via `operator[]` even if all you're going to do is take a pointer?

Precisely.

PR #4213, opened 2025-07-15
Iterate over GhostingFunctor* in vectors, not sets

roystgnr:
It's entirely fair for GhostingFunctor subclasses to want to do parallel operations in e.g. mesh_reinit(), but we can't do that if they're not executed in the same order on all processors, and a set of pointers is definitely not guaranteed to be in the same order on all processors.

I think I may have actually fixed this bug before anyone got bitten by it (while investigating other issues, though).

This is technically an API change, but hopefully everybody uses "auto" for any ugly iterators that they have to declare, in which case this won't be an API change that breaks anyone's compile.

### DIFF
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -1613,8 +1613,8 @@ void
 DofMap::
 merge_ghost_functor_outputs(GhostingFunctor::map_type & elements_to_ghost,
                             CouplingMatricesSet & temporary_coupling_matrices,
-                            const std::vector<GhostingFunctor *>::const_iterator & gf_begin,
-                            const std::vector<GhostingFunctor *>::const_iterator & gf_end,
+                            const GhostingFunctorIterator & gf_begin,
+                            const GhostingFunctorIterator & gf_end,
                             const MeshBase::const_element_iterator & elems_begin,
                             const MeshBase::const_element_iterator & elems_end,
                             processor_id_type p)
--- a/src/base/dof_map.C
+++ b/src/base/dof_map.C
@@ -1613,8 +1613,8 @@ void
 DofMap::
 merge_ghost_functor_outputs(GhostingFunctor::map_type & elements_to_ghost,
                             CouplingMatricesSet & temporary_coupling_matrices,
-                            const std::set<GhostingFunctor *>::iterator & gf_begin,
-                            const std::set<GhostingFunctor *>::iterator & gf_end,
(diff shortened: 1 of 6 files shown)


jwpeterson:
I think I'm OK with the API change, however I would like to get a coupon for one free future API change of mine to be redeemed later :laughing: 

Any idea if the MOOSE test failures here are real/actually due to this PR?
```
[1.303s]     FAIL dgkernels/2d_diffusion_dg.proper_ghosting_with_action_serial FAILED (CRASH)
[1.835s]     FAIL dgkernels/2d_diffusion_dg.proper_ghosting_with_action_parallel FAILED (CRASH) [min_cpus=2]
```

jwpeterson:
BTW, would it make sense to put a public typedef on MeshBase like:
```
typedef std::vector<GhostingFunctor *>::const_iterator GhostingFunctorIterator;
```
This would give people something to use in cases where it's not possible to use `auto`, and would also make the next diff smaller in case it changes to a std::list or something in the future.

roystgnr:
The failures are definitely real - using `set` was automatically correcting double-insertions, and MOOSE was depending on that in more than one code path.  I'm going to restore support for that behavior.

If I can find all the *other* points where MOOSE was depending on that support, then I'll mark the revised support for it as deprecated for now, but if not then I'll just leave it supported indefinitely.   Supporting it with vector instead of set gets us from O(log(N)) to O(N), but here N is the tiny number of ghosting functors and we're already unavoidably O(N*N_elem) when we actually query ghosting functors elsewhere.

I thought about adding that typedef but decided to be lazy.  Thanks for calling me on it; my superego now feels vindicated and my id chastened.

src/base/dof_map.C
```diff
@@ -2033,6 +2033,14 @@ void
 DofMap::add_coupling_functor(GhostingFunctor & coupling_functor,
                              bool to_mesh)
 {
+  // We used to implicitly support duplicate inserts to std::set
+#ifdef LIBMESH_ENABLE_DEPRECATED
+  _coupling_functors.erase
+    (std::remove(_coupling_functors.begin(),
+                 _coupling_functors.end(),
+                 &coupling_functor));
```
jwpeterson: `std::vector::erase(pos)` erases only the element at `pos`... are you trying to erase *all* the elements that get std::removed here?

https://en.cppreference.com/w/cpp/container/vector/erase.html

roystgnr: Yeah, that was a stupid bug.

PR #4214, opened 2025-07-15
Refactor add_periodic_boundary, adding more error checking

roystgnr:
@loganharbour noticed the opportunities here; hopefully I didn't miss anything.


roystgnr:
From Slack, for public reference:

"Okay, just pushed the polymorphic approach. Let me know which one you prefer and we'll stick with that."

"I think I like the std::variant version better.  Sorry to make you go to so much work!  I'd be okay with either version."

"Yes, the polymorphic feels a bit more cluttered, doesn’t it?  No worries, I was interested in how it would turn out"

I think my only remaining real complaint is that, even if we end up relying on exceptions here, at a minimum we need to disable the tests relying on them until the --disable-exceptions CI passes.

PR #4216, opened 2025-07-17
Added sliding boundary nodes to VariationalMeshSmoother

pbehne:
#### Summary
This pull request implements sliding external and subdomain boundary node constraints in the `VariationalMeshSmoother`
In addition to expanding test coverage to test sliding boundary nodes, 1D and 3D test cases were added to the unit test suite.

#### Highlights
- **Constraint Framework Refactor**
  - Introduced `PointConstraint`, `LineConstraint`, and `PlaneConstraint` structs to represent fixed nodes, nodes constrained to a line, and nodes constrained to a plane
  - Implemented a unified `intersect()` interface for combining constraints using `std::variant`
  - Added robust geometric logic for line-line, plane-line, and plane-plane intersections
  - Fallback to `PointConstraint` if constraints fail to intersect cleanly

- **Sliding Boundary Enhancements**
  - Implemented sliding boundary node constraints: previously, boundary and subdomain boundary nodes were fully fixed and not allowed to slide
  - Nodes on external boundaries internal subdomain boundaries now slide along planes and lines unless constrained by geometry

- **Expanded Tests and Validation**
  - Added new smoother test cases in 1D and 3D
  - Replaced `DistortSquare` with `DistortHypercube`, supporting 1D, 2D, and 3D skewing
  - Boundary nodes are now distorted within the faces and edges they belong to (while preserving corner vertices)
  - Increased complexity of subdomain boundary test to cover more complex cases such a an element having neighbors of multiply different subdomain ids

#### References
- Related to #4082

### DIFF
--- a/include/systems/variational_smoother_constraint.h
+++ b/include/systems/variational_smoother_constraint.h
@@ -24,6 +24,138 @@
 
 namespace libMesh
 {
+
+struct PointConstraint;
+struct LineConstraint;
+struct PlaneConstraint;
+
+/// Type used to store a constraint that may be a PlaneConstraint,
+/// LineConstraint, or PointConstraint
+using ConstraintVariant =
+    std::variant<PointConstraint, LineConstraint, PlaneConstraint>;
+
+/// Represents a fixed point constraint.
+struct PointConstraint {
+  Point location;
+
+  PointConstraint() = default;
+
+  /// Constructor
+  /// @parm p The point defining the constraint.
+  PointConstraint(const Point &p);
+
+  bool operator<(const PointConstraint &other) const;
+
+  bool operator==(const PointConstraint &other) const;
+
+  /// Computes the intersection of this point with another constraint.
+  /// Handles intersection with PointConstraint, LineConstraint, or
+  /// PlaneConstraint.
+  /// @param other The constraint to intersect with.
+  /// @return The most specific ConstraintVariant that satisfies both
+  /// constraints.
+  ConstraintVariant intersect(const ConstraintVariant &other) const;
+};
+
+/// Represents a line constraint defined by a base point and direction vector.
+struct LineConstraint {
+  Point r0;
+  Point dir;
+
+  LineConstraint() = default;
+
+  /// Constructor
+  /// @parm p A point on the constraining line.
+  /// @param d the direction of the constraining line.
+  LineConstraint(const Point &p, const Point &d);
+
+  bool operator<(const LineConstraint &other) const;
+
+  bool operator==(const LineConstraint &other) const;
+
+  /// Query whether a point lies on the line.
+  /// @param p The point in question
+  /// @return bool indicating whether p lies on the line.
+  bool contains_point(const PointConstraint &p) const;
+
+  /// Query whether a line is parallel to this line
+  /// @ param l The line in question
+  /// @ return bool indicating whether l is parallel to this line.
+  bool is_parallel(const LineConstraint &l) const;
+
+  /// Query whether a plane is parallel to this line
+  /// @ param p The plane in question
+  /// @ return bool indicating whether p is parallel to this line.
+  bool is_parallel(const PlaneConstraint &p) const;
+
+  /// Computes the intersection of this line with another constraint.
+  /// Handles intersection with LineConstraint, PlaneConstraint, or
+  /// PointConstraint.
+  /// @param other The constraint to intersect with.
+  /// @return The most specific ConstraintVariant that satisfies both
+  /// constraints.
+  ConstraintVariant intersect(const ConstraintVariant &other) const;
+};
+
+/// Represents a plane constraint defined by a point and normal vector.
+struct PlaneConstraint {
+  Point point;
+  Point normal;
+
+  PlaneConstraint() = default;
+
+  /// Constructor
+  /// @parm p A point on the constraining plane.
+  /// @param n the direction normal to the constraining plane.
+  PlaneConstraint(const Point &p, const Point &n);
+
+  bool operator<(const PlaneConstraint &other) const;
+
+  bool operator==(const PlaneConstraint &other) const;
+
+  /// Query whether a point lies on the plane.
+  /// @param p The point in question
+  /// @return bool indicating whether p lies on the plane.
+  bool contains_point(const PointConstraint &p) const;
+
+  /// Query whether a line lies on the plane.
+  /// @param l The line in question
+  /// @return bool indicating whether l lies on the plane.
+  bool contains_line(const LineConstraint &l) const;
+
+  /// Query whether a plane is parallel to this plane
+  /// @ param p The plane in question
+  /// @ return bool indicating whether p is parallel to this plane.
+  bool is_parallel(const PlaneConstraint &p) const;
+
+  /// Query whether a line is parallel to this plane
+  /// @ param l The line in question
+  /// @ return bool indicating whether l is parallel to this plane.
+  bool is_parallel(const LineConstraint &l) const;
+
+  /// Computes the intersection of this plane with another constraint.
+  /// Handles intersection with PlaneConstraint, LineConstraint, or
+  /// PointConstraint.
+  /// @param other The constraint to intersect with.
+  /// @return The most specific ConstraintVariant that satisfies both
+  /// constraints.
+  ConstraintVariant intersect(const ConstraintVariant &other) const;
+};
+
+/// Dispatch intersection between two constraint variants.
+/// Resolves to the appropriate method based on the type of the first operand.
+/// @param a First constraint.
+/// @param b Constraint to combine with a.
+/// @return Combination (intersection) of constraint a and b.
+inline ConstraintVariant intersect_constraints(const ConstraintVariant &a,
+                                               const ConstraintVariant &b) {
+  return std::visit(
+      [](const auto &lhs, const auto &rhs) -> ConstraintVariant {
+        return lhs.intersect(rhs);
+      },
+      a, b);
+}
+
 /*
  * Constraint class for the VariationalMeshSmoother.
  *
@@ -39,15 +171,6 @@ private:
 
   /// Whether nodes on subdomain boundaries are subject to change via smoothing
   const bool _preserve_subdomain_boundaries;
-
-  /*
-   * Identifies and imposes the appropriate constraints on a node.
-   * @param node The node to constrain.
-   * @param neighbors Vector of neighbors to use to identify the constraint to
-   * impose.
-   */
-  void impose_constraints(const Node &node,
-                          const std::vector<const Node *> neighbors);
 
   /*
    * Constrain (i.e., fix) a node to not move during mesh smoothing.
@@ -87,21 +210,56 @@ private:
       const Elem & containing_elem,
       const BoundaryInfo & boundary_info);
 
-  static void filter_neighbors_for_subdomain_constraint(
-      const Node & node,
-      std::vector<const Node *> & neighbors,
-      const subdomain_id_type sub_id1,
+  /// Get the relevant nodal neighbors for a subdomain constraint.
+  /// @param mesh The mesh being smoothed.
+  /// @param node The node (on the subdomain boundary) being constrained.
+  /// @param sub_id1 The subdomain id of the block on one side of the subdomain
+  /// boundary.
+  /// @param sub_id2 The subdomain id of the block on the other side of the
+  /// subdomain boundary.
+  /// @param nodes_to_elem_map A mapping from node id to containing element ids.
+  /// @return A set of node pointer sets containing nodal neighbors to 'node' on
+  /// the sub_id1-sub_id2 boundary. The subsets are grouped by element faces
+  /// that form the subdomain boundary. Note that 'node' itself does not appear
+  /// in this set.
+  static std::set<std::set<const Node *>>
+  get_neighbors_for_subdomain_constraint(
+      const MeshBase &mesh, const Node &node, const subdomain_id_type sub_id1,
       const subdomain_id_type sub_id2,
-      std::unordered_map<dof_id_type, std::vector<const Elem *>> & nodes_to_elem_map
-    );
+      const std::unordered_map<dof_id_type, std::vector<const Elem *>>
+          &nodes_to_elem_map);
 
-  static void filter_neighbors_for_boundary_constraint(
-      const Node & node,
-      std::vector<const Node *> & neighbors,
-      std::unordered_map<dof_id_type, std::vector<const Elem *>> & nodes_to_elem_map,
-      const std::unordered_set<dof_id_type> & boundary_node_ids,
-      const BoundaryInfo & boundary_info
-    );
+  /// Get the relevant nodal neighbors for an external boundary constraint.
+  /// @param mesh The mesh being smoothed.
+  /// @param node The node (on the external boundary) being constrained.
+  /// @param boundary_node_ids The set of mesh's external boundary node ids.
+  /// @param boundary_info The mesh's BoundaryInfo.
+  /// @param nodes_to_elem_map A mapping from node id to containing element ids.
+  /// @return A set of node pointer sets containing nodal neighbors to 'node' on
+  /// the external boundary. The subsets are grouped by element faces that form
+  /// the external boundary. Note that 'node' itself does not appear in this
+  /// set.
+  static std::set<std::set<const Node *>> get_neighbors_for_boundary_constraint(
+      const MeshBase &mesh, const Node &node,
+      const std::unordered_set<dof_id_type> &boundary_node_ids,
+      const BoundaryInfo &boundary_info,
+      const std::unordered_map<dof_id_type, std::vector<const Elem *>>
+          &nodes_to_elem_map);
+
+  /// Determines the appropriate constraint (PointConstraint, LineConstraint, or
+  /// PlaneConstraint) for a node based on its neighbors.
+  /// @param node The node to constrain.
+  /// @param dim The mesh dimension.
+  /// @return The best-fit constraint for the given geometry.
+  static ConstraintVariant determine_constraint(
+      const Node &node, const unsigned int dim,
+      const std::set<std::set<const Node *>> &side_grouped_boundary_neighbors);
+
+  /// Applies a given constraint to a node (e.g., fixing it, restricting it to a
+  /// line or plane).
+  /// @param node The node to constrain.
+  /// @param constraint The geometric constraint variant to apply.
+  void impose_constraint(const Node &node, const ConstraintVariant &constraint);
 
 public:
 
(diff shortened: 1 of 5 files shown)


include/systems/variational_smoother_constraint.h
```diff
@@ -24,6 +24,138 @@
...
+struct PointConstraint {
+  Point location;
+
+  PointConstraint() = default;
+
+  /// Constructor
+  /// @parm p The point defining the constraint.
+  PointConstraint(const Point &p);
+
+  bool operator<(const PointConstraint &other) const;
```
roystgnr, line 70: Docs on all these comparators?  Even if they're just for use with std::set or something we should know what they mean.  There's one obvious interpretation for PointConstraint, but it's a little ambiguous for the others.

include/systems/variational_smoother_constraint.h
```diff
@@ -24,6 +24,138 @@
...
+  LineConstraint(const Point &p, const Point &d);
+
+  bool operator<(const LineConstraint &other) const;
+
+  bool operator==(const LineConstraint &other) const;
+
+  /// Query whether a point lies on the line.
+  /// @param p The point in question
+  /// @return bool indicating whether p lies on the line.
+  bool contains_point(const PointConstraint &p) const;
```
roystgnr, line 159: Any chance we can get away with moving these struct definitions to the .C file?

If they're basically invisible to the rest of the library, then the code here is fine, great way to refactor stuff for internal use.

If not, then we may want to do things like change them into classes encapsulating the members, change `TOLERANCE` into an (optional, defaulting to `TOLERANCE*TOLERANCE` or whatever) `tol` argument on all the functions that rely on comparison to within a tolerance, etc. etc., before we start getting unexpected external uses of APIs we didn't intend to support directly.

include/systems/variational_smoother_constraint.h
```diff
@@ -41,11 +173,91 @@ class VariationalSmootherConstraint : public System::Constraint
...
    */
   void fix_node(const Node & node);
 
+  /*
+   * Constrain a node to remain in the given plane during mesh smoothing.
+   * @param node Node to constrain
+   * @param ref_normal_vec Reference normal vector to the constraining plane.
+   * This, along with the coordinates of node, are used to define the
+   * constraining plane.
+   */
```
roystgnr, line 412: Could we consistently use this (or better, with the leading `/**`) style for doxygen comments, rather than randomly switching to `///` depending on how ChatGPT was feeling each day?  ;-)

roystgnr: Not to knock ChatGPT, of course; I'm feeling a little called out by the consistent use of `@param`...

include/systems/variational_smoother_constraint.h
```diff
@@ -24,6 +24,138 @@
...
+  bool operator<(const PointConstraint &other) const;
+
+  bool operator==(const PointConstraint &other) const;
+
+  /// Computes the intersection of this point with another constraint.
+  /// Handles intersection with PointConstraint, LineConstraint, or
+  /// PlaneConstraint.
+  /// @param other The constraint to intersect with.
+  /// @return The most specific ConstraintVariant that satisfies both
+  /// constraints.
```
roystgnr: Looks like you're throwing an error if the intersection is empty; document that?

src/systems/variational_smoother_constraint.C
```diff
@@ -18,10 +18,253 @@
...
+
+  const auto dot_prod = d * canonical;
+  libmesh_assert(!absolute_fuzzy_equals(dot_prod, 0.));
+  dir = (dot_prod > 0) ? d.unit() : -d.unit();
+}
+
+bool LineConstraint::operator<(const LineConstraint &other) const {
+  if (!(dir.absolute_fuzzy_equals(other.dir, TOLERANCE)))
+    return dir < other.dir;
+  return (dir * r0) < (other.dir * other.r0) - TOLERANCE;
```
roystgnr: This isn't an antisymmetric relation - is that intentional, to get `set` etc. to consider nearly-identical lines as identical and drop one of them?  If so we should document that.

src/systems/variational_smoother_constraint.C
```diff
@@ -18,10 +18,253 @@
...
+}
+
+bool LineConstraint::operator==(const LineConstraint &other) const {
+  if (!(dir.absolute_fuzzy_equals(other.dir, TOLERANCE)))
+    return false;
+  return this->contains_point(other.r0);
```
roystgnr: This isn't mathematically compatible with operator< above - we can have two lines that are equal (same direction, each contains the other's point) but also one is less than the other (the points differ).  We definitely need to document what these are supposed to do so we can figure out for sure whether they actually do it.  ;-)

src/systems/variational_smoother_constraint.C
```diff
@@ -18,10 +18,253 @@
...
+  return std::visit(
+      [&](auto &&o) -> ConstraintVariant {
+        using T = std::decay_t<decltype(o)>;
+        if constexpr (std::is_same_v<T, LineConstraint>) {
+          if (*this == o)
+            return *this;
+          libmesh_error_msg_if(this->is_parallel(o),
+                               "Lines are parallel and do not intersect.");
+
+          // Solve for t in the equation p1 + t·d1 = p2 + s·d2
```
roystgnr: I'm thrilled with the idea of unleashing unicode in comments, and we've got it in a bunch of our contrib/ codes already, but I think this will be the first time it's in libMesh proper, so I want to make sure @jwpeterson is fine with it too.

If we can talk the C++ people into giving us `operator×` and `operator⊗` then I'm stripping the library of `cross()` and `outer_product()` calls approximately five minutes after we upgrade.

src/systems/variational_smoother_constraint.C
```diff
@@ -33,82 +276,109 @@ VariationalSmootherConstraint::~VariationalSmootherConstraint() = default;
...
+      // subdomain_constraint
+      try
+      {
+        const auto combined_constraint =
+            intersect_constraints(subdomain_constraint, boundary_constraint);
+        this->impose_constraint(node, combined_constraint);
+      }
+      catch (const std::exception & e)
+      {
+        // This will catch cases where constraints have no intersection
```
roystgnr: Two things:

This should probably be `libmesh_try`/`libmesh_catch` to compile on builds with exceptions disabled.

How are we going to hit this at all?  Every constrant we're creating contains the node itself, right?  So at the very least shouldn't we already get something within epsilon of `PointConstraint(node)` as the intersection?

src/systems/variational_smoother_constraint.C
```diff
@@ -124,4 +394,397 @@ void VariationalSmootherConstraint::fix_node(const Node & node)
...
+  // Fall back to point constraint
+  if (valid_planes.empty())
+    return PointConstraint(node);
+
+  // Combine all the planes together to get a common constraint
+  auto it = valid_planes.begin();
+  ConstraintVariant current = *it++;
+  for (; it != valid_planes.end(); ++it)
+  {
+    try
```
roystgnr: Same "how are we possibly getting an empty constraint" question here

src/systems/variational_smoother_constraint.C
pbehne: > to get set etc. to consider nearly-identical lines as identical and drop one of them?

That was the AI's idea. Good point on the asymmetry. I'm taking that out and putting in a check for fuzzy equality prior to checking less than.

src/systems/variational_smoother_constraint.C
pbehne: Great question. I asked myself the same thing when an error was thrown and crashed the program. This is from the `testVariationalHexMultipleSubdomains` test. First, I think it is work including a picture of the results. I'll update the PR description to do so, and will copy the result here:
image
Consider this view of the z=0 boundary of the distorted mesh:
image
Let's look at subdomain 0 only:
image
Zoom in on this element:
image
The top of this element, which forms part of the subdomain boundary is the union of two different planar surfaces. `valid_planes`, when determining the constraint for the front left node, ends up holding 3 planes: The two seen in the figure, plus their "average". The average plane does not intersect both of the first two planes, and we end up with the case of a line (the combination of two of the planes) being parallel but not coincident with the third plane. Thus, an exception is thrown and we fall back to fixing the node.

src/systems/variational_smoother_constraint.C
pbehne: See the next comment regarding the other place I use try/catch. I included it here as well to avoid issues similar to what I explained below.

include/systems/variational_smoother_constraint.h
pbehne: The `determine_constraint` and `impose_constraint` methods of `VariationalSmootherConstraint` use `ConstraintVariant`, so it doesn't seem like I can move the `Point/Line/PlaneConstraint` structs fully into `src`. I'll go ahead and lock them down.

src/systems/variational_smoother_constraint.C
roystgnr: > The top of this element, which forms part of the subdomain boundary is the union of two different planar surfaces.

No, it's not.  The top is non-planar.  Paraview just decimates bilinear quads into pairs of linear triangles.

But the non-planarity gives you the exact same problem.  Thanks for the explanation!  We should put something into comments about that.

I'm not sure this is the right place to use exceptions, though.  They're ideally exception-al, not something that applies to all quad faces that aren't planar.  I'd have hoped we'd simply end up with a PointConstraint here.

I guess the code is fine as is, it just feels like a code smell to me, and we're going to have to turn on some additional recipes to make sure that builds with exceptions disabled don't die in the unit tests.

roystgnr:
From Slack, for public reference:

"Okay, just pushed the polymorphic approach. Let me know which one you prefer and we'll stick with that."

"I think I like the std::variant version better. Sorry to make you go to so much work! I'd be okay with either version."

"Yes, the polymorphic feels a bit more cluttered, doesn’t it? No worries, I was interested in how it would turn out"

I think my only remaining real complaint is that, even if we end up relying on exceptions here, at a minimum we need to disable the tests relying on them until the --disable-exceptions CI passes.

roystgnr (changes requested):
I'm thrilled with this now except for a couple of the whitespace changes; @pbehne is going to rewrite history now, and has sworn he won't edit anything else but whitespace in the process.

include/systems/variational_smoother_constraint.h
```diff
@@ -22,36 +22,469 @@
...
+    libmesh_assert_msg(false, _err_msg);
+    return false;
+  }
+
+private:
+  std::string _err_msg =
+      std::string("We should never get here! The InvalidConstraint object ") +
+      std::string(
+          "should be detected and replaced with a valid ConstraintVariant ") +
+      std::string("prior to calling any class methods.");
```
jwpeterson: This looks a bit unusual... is there any reason to make 3 std::string constructor calls like this? Our usual approach is to initialize things in the constructor initialization list, which I think would be better here.

src/systems/variational_smoother_constraint.C
```diff
@@ -18,110 +18,822 @@
...
+  const bool base_on_plane = this->contains_point(PointConstraint(l.point()));
+  const bool dir_orthogonal = std::abs(_normal * l.direction()) < _tol;
+  return base_on_plane && dir_orthogonal;
+}
+
+ConstraintVariant PlaneConstraint::intersect(const ConstraintVariant & other) const
+{
+  return std::visit(
+      [&](auto && o) -> ConstraintVariant {
+        using T = std::decay_t<decltype(o)>;
```
jwpeterson, line 231: What is the purpose of the `std::decay_t<decltype>` call here, exactly? I guess it's so that you have a type `T` to use in the subsequent `std::is_same_v` calls... but it raises the question of whether we can just use `decltype(o)` directly. This might be worth commenting about unless it is utterly standard practice with `std::variant` code that I'm just not familiar with.

src/systems/variational_smoother_constraint.C
```diff
@@ -18,110 +18,822 @@
...
+      const auto side_grouped_boundary_neighbors = get_neighbors_for_boundary_constraint(
+          mesh, node, boundary_node_ids, boundary_info, nodes_to_elem_map);
+
+      // Determine which constraint should be imposed
+      const auto boundary_constraint =
+          determine_constraint(node, dim, side_grouped_boundary_neighbors);
+
+      // Check for the case where this boundary node is also part of a subdomain id boundary
+      const auto it = subdomain_boundary_map.find(bid);
+      if (it != subdomain_boundary_map.end())
```
jwpeterson: You could consider stuffing the `it` declaration inside the `if`-statement here, just to restrict that variable's scope.

src/systems/variational_smoother_constraint.C
```diff
@@ -18,110 +18,822 @@
...
+  // Constrain the node to it's current location
+  if (dim == 1 || neighbors.size() == 1)
+    return PointConstraint{node};
+
+  if (dim == 2 || neighbors.size() == 2)
+    {
+      // Determine whether node + all neighbors are colinear
+      bool all_colinear = true;
+      const Point ref_dir = (*neighbors[0] - node).unit();
+      for (const auto i : make_range(size_t(1), neighbors.size()))
```
jwpeterson: I think we generally use `std::size_t` elsewhere so it would be good to keep that consistent.

src/systems/variational_smoother_system.C
```diff
@@ -108,24 +109,97 @@ void VariationalSmootherSystem::compute_element_reference_volume()
...
+
   for (const auto * elem : mesh.active_local_element_ptr_range())
   {
     femcontext.pre_fe_reinit(*this, elem);
     femcontext.elem_fe_reinit();
 
-    const auto elem_integrated_det_J = std::accumulate(JxW.begin(), JxW.end(), 0.);
+    // Add target element info, if applicable
+    if (_target_inverse_jacobians.find(elem->type()) ==
+        _target_inverse_jacobians.end()) {
```
jwpeterson: We put opening curly brace of if-statements on a separate line.

src/systems/variational_smoother_system.C
```diff
@@ -108,24 +109,97 @@ void VariationalSmootherSystem::compute_element_reference_volume()
...
+      const auto nq_points = femcontext.get_element_qrule().n_points();
+
+      // If the target element is the reference element, Jacobian matrix is
+      // identity, det of inverse is 1. This will only be overwritten if a
+      // different target elemen is explicitly specified.
+      target_elem_inverse_jacobian_dets[elem->type()] =
+          std::vector<Real>(nq_points, 1.0);
+
+      switch (elem->type()) {
+      case TRI3: {
```
jwpeterson: Curly braces on new lines here as well.

include/systems/variational_smoother_constraint.h
```diff
@@ -22,36 +22,469 @@
...
+ * Type used to store a constraint that may be a PlaneConstraint,
+ * LineConstraint, or PointConstraint
+ */
+using ConstraintVariant = std::variant<PointConstraint, LineConstraint,
+                                       PlaneConstraint, InvalidConstraint>;
+
+/**
+ * Represents a fixed point constraint.
+ */
+class PointConstraint {
```
jwpeterson: Curly braces for classes should be on next line.

src/systems/variational_smoother_constraint.C
pbehne: The purpose of std::decay_t is to strip references/const/volatile from o's type so that we have a plant type to used in std::is_same. I'll add the comment!

roystgnr (approved):

jwpeterson:
FYI @pbehne this PR seems to have broken the `--disable-deprecated` build (see here, for example: https://civet.inl.gov/branch/14145/). The error message is:

```
../src/systems/variational_smoother_system.C: In member function 'void libMesh::VariationalSmootherSystem::prepare_for_smoothing()':
../src/systems/variational_smoother_system.C:158:37: error: no matching function for call to 'libMesh::Tri3::set_node(int)'
               target_elem.set_node(0) = &node_0;
                                     ^
In file included from ../src/systems/variational_smoother_system.C:20:
./include/libmesh/elem.h:2571:6: note: candidate: 'virtual void libMesh::Elem::set_node(unsigned int, libMesh::Node*)'
 void Elem::set_node (const unsigned int i,
      ^~~~
./include/libmesh/elem.h:2571:6: note:   candidate expects 2 arguments, 1 provided
../src/systems/variational_smoother_system.C:159:37: error: no matching function for call to 'libMesh::Tri3::set_node(int)'
               target_elem.set_node(1) = &node_1;
                                     ^
In file included from ../src/systems/variational_smoother_system.C:20:
./include/libmesh/elem.h:2571:6: note: candidate: 'virtual void libMesh::Elem::set_node(unsigned int, libMesh::Node*)'
 void Elem::set_node (const unsigned int i,
      ^~~~
./include/libmesh/elem.h:2571:6: note:   candidate expects 2 arguments, 1 provided
../src/systems/variational_smoother_system.C:160:37: error: no matching function for call to 'libMesh::Tri3::set_node(int)'
               target_elem.set_node(2) = &node_2;
                                     ^
In file included from ../src/systems/variational_smoother_system.C:20:
./include/libmesh/elem.h:2571:6: note: candidate: 'virtual void libMesh::Elem::set_node(unsigned int, libMesh::Node*)'
 void Elem::set_node (const unsigned int i,
      ^~~~
./include/libmesh/elem.h:2571:6: note:   candidate expects 2 arguments, 1 provided
```
so the fix should be simple, just use the non-deprecated `Elem::set_node()` API instead.

pbehne:
Thanks for the heads up, I'll add the fix to #4218!

pbehne:
FYI, only the deprecated version shows up in doxygen. I had to go look at `elem.h` to see the non-deprecated version.

jwpeterson:
> FYI, only the deprecated version shows up in doxygen. I had to go look at `elem.h` to see the non-deprecated version.

OK, that's interesting... I guess Doxygen is somehow seeing the "pre-processed" version of the source (and the defines are somehow set such that it doesn't generate docs for the non-deprecated version)? Maybe that's a normal Doxygen thing, I don't know.

pbehne:
Yeah, I figured I'd mention it since I recall @roystgnr mentioning that he and you typically just use the header files themselves for documentation. As someone less familiar with the library, I prefer doxygen so I don't have to dig through parent headers to see what a given class does and doesn't inherit.

roystgnr:
I think the problem is that we only update [the libMesh Doxygen](https://libmesh.github.io/doxygen/) manuallly, by which I mean "almost never", by which I mean "Oh my God the files [still have copyright 2002-2018 headers](https://libmesh.github.io/doxygen/dof__map_8C_source.html)".  The libMesh Doxygen [on the MOOSE site](https://mooseframework.inl.gov/docs/doxygen/libmesh/dof__map_8C_source.html) gets autoupdated, which I guess I thought was good enough ... but it doesn't matter if the MOOSE copy [marks something as deprecated](https://mooseframework.inl.gov/docs/doxygen/libmesh/classlibMesh_1_1Elem.html#a0d25454251fb8f3518211ccd9f467b91) if that's not the copy everybody is reading.

PR #4217, opened 2025-07-22
Fix compatibility with AddressSanitizer flags

roystgnr:
This looks like it's been broken for years, since the autoconf-submodule refactoring.


roystgnr:
This seemed to be working for me with g++, but we didn't emit any errors (not surprising since we're regularly testing with valgrind); I'll wait for @loganharbour to play with it with clang++ before we merge.

roystgnr:
Looks like this got dropped; I'll rebase to make sure it's not breaking anything and merge if so.

loganharbour:
Looks good?

PR #4218, opened 2025-07-28
VariationalMeshSmoother: Support for higher order elements

pbehne:
This pull request extends the `VariationalMeshSmoother` to handle higher-order elements, as part of the work on issue #4082.

- **Improved Test Coverage**:

  - Updated the `VariationalMeshSmoother` tester to better handle symmetric configurations. Nodes previously unaffected by the `DistortHypercube` due to lying on planes of symmetry now receive a 5% forced distortion.
  - The above forced distortion removed the need to skip distortion checks on nodes with coordinate value 0.5, simplifying the `distortion_is` logic.

- **Extended Element Support**:

  - **EDGE3 and HEX20**: Added tests verifying smoother functionality on these higher-order elements. Minor adjustments to `distortion_is` were made to correctly map midpoint nodes to integer values using `fe_order`.
  - **TRI6**: Introduced a target equilateral quadratic triangle to support smoothing for TRI6 elements, along with corresponding tests.
  - **HEX27**: Implemented logic to determine face neighbors for face-center nodes when no nodal neighbors are available, enabling full support and test coverage for HEX27 elements.

Ref #4082.

### DIFF
--- a/src/systems/variational_smoother_constraint.C
+++ b/src/systems/variational_smoother_constraint.C
@@ -536,10 +536,6 @@ void VariationalSmootherConstraint::find_nodal_or_face_neighbors(
   if (!neighbors.size())
     {
       // Grab the element containing node
-      // It doesn't make sense for a node having no edge neighbors to belong to
-      // multiple elements, right? Otherwise, we will figure out those cases as
-      // they occur.
-      libmesh_assert(nodes_to_elem_map.at(node.id()).size() == 1);
       const auto *elem = nodes_to_elem_map.at(node.id()).front();
       // Find the element side containing node
       for (const auto &side : elem->side_index_range())
--- a/src/systems/variational_smoother_constraint.C
+++ b/src/systems/variational_smoother_constraint.C
@@ -521,6 +521,51 @@ void VariationalSmootherConstraint::constrain_node_to_line(const Node & node,
     }
 }
 
+void VariationalSmootherConstraint::find_nodal_or_face_neighbors(
+    const MeshBase &mesh, const Node &node,
+    const std::unordered_map<dof_id_type, std::vector<const Elem *>>
+        &nodes_to_elem_map,
+    std::vector<const Node *> &neighbors)
+{
+  // Find all the nodal neighbors... that is the nodes directly connected
+  // to this node through one edge.
+  MeshTools::find_nodal_neighbors(mesh, node, nodes_to_elem_map, neighbors);
+
+  // If no neighbors are found, use all nodes on the containing side as
+  // neighbors.
+  if (!neighbors.size())
+    {
+      // Grab the element containing node
+      // It doesn't make sense for a node having no edge neighbors to belong to
+      // multiple elements, right? Otherwise, we will figure out those cases as
+      // they occur.
+      libmesh_assert(nodes_to_elem_map.at(node.id()).size() == 1);
+      const auto *elem = nodes_to_elem_map.at(node.id()).front();
+      // Find the element side containing node
+      for (const auto &side : elem->side_index_range())
+        {
+          bool found_node = false;
+          const auto &nodes_on_side = elem->nodes_on_side(side);
+          for (const auto &local_node_id : nodes_on_side)
+            if (elem->node_id(local_node_id) == node.id())
+            {
+              found_node = true;
+              break;
+            }
+          if (found_node)
+            {
+              for (const auto &local_node_id : nodes_on_side)
+                // No need to add node itself as a neighbor
+                if (const auto *node_ptr = elem->node_ptr(local_node_id);
+                    *node_ptr != node)
+                  neighbors.push_back(node_ptr);
+              break;
+            }
+        }
+    }
+  libmesh_assert(neighbors.size());
+}
+
 // Utility function to determine whether two nodes share a boundary ID.
 // The motivation for this is that a sliding boundary node on a triangular
 // element can have a neighbor boundary node in the same element that is not
@@ -588,9 +633,11 @@ VariationalSmootherConstraint::get_neighbors_for_subdomain_constraint(
 {
 
   // Find all the nodal neighbors... that is the nodes directly connected
-  // to this node through one edge
+  // to this node through one edge, or if none exists, use other nodes on the
+  // containing face
   std::vector<const Node *> neighbors;
-  MeshTools::find_nodal_neighbors(mesh, node, nodes_to_elem_map, neighbors);
+  VariationalSmootherConstraint::find_nodal_or_face_neighbors(
+      mesh, node, nodes_to_elem_map, neighbors);
 
   // Each constituent set corresponds to neighbors sharing a face on the
   // subdomain boundary
@@ -672,9 +719,11 @@ VariationalSmootherConstraint::get_neighbors_for_boundary_constraint(
 {
 
   // Find all the nodal neighbors... that is the nodes directly connected
-  // to this node through one edge
+  // to this node through one edge, or if none exists, use other nodes on the
+  // containing face
   std::vector<const Node *> neighbors;
-  MeshTools::find_nodal_neighbors(mesh, node, nodes_to_elem_map, neighbors);
+  VariationalSmootherConstraint::find_nodal_or_face_neighbors(
+      mesh, node, nodes_to_elem_map, neighbors);
 
   // Each constituent set corresponds to neighbors sharing a face on the
   // boundary
--- a/src/systems/variational_smoother_constraint.C
+++ b/src/systems/variational_smoother_constraint.C
@@ -522,10 +522,10 @@ void VariationalSmootherConstraint::constrain_node_to_line(const Node & node,
 }
 
 void VariationalSmootherConstraint::find_nodal_or_face_neighbors(
-    const MeshBase &mesh, const Node &node,
-    const std::unordered_map<dof_id_type, std::vector<const Elem *>>
-        &nodes_to_elem_map,
-    std::vector<const Node *> &neighbors)
+    const MeshBase & mesh,
+    const Node & node,
+    const std::unordered_map<dof_id_type, std::vector<const Elem *>> & nodes_to_elem_map,
+    std::vector<const Node *> & neighbors)
 {
   // Find all the nodal neighbors... that is the nodes directly connected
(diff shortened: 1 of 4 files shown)


src/systems/variational_smoother_constraint.C
```diff
@@ -521,6 +521,51 @@ void VariationalSmootherConstraint::constrain_node_to_line(const Node & node,
...
+  MeshTools::find_nodal_neighbors(mesh, node, nodes_to_elem_map, neighbors);
+
+  // If no neighbors are found, use all nodes on the containing side as
+  // neighbors.
+  if (!neighbors.size())
+    {
+      // Grab the element containing node
+      // It doesn't make sense for a node having no edge neighbors to belong to
+      // multiple elements, right? Otherwise, we will figure out those cases as
+      // they occur.
```
roystgnr: Won't any face-centered node on the "boundary" between two subdomains belong to two elements?

src/systems/variational_smoother_system.C
```diff
@@ -113,90 +114,135 @@ void VariationalSmootherSystem::prepare_for_smoothing()
...
+          const auto nq_points = femcontext.get_element_qrule().n_points();
+
+          // If the target element is the reference element, Jacobian matrix is
+          // identity, det of inverse is 1. This will only be overwritten if a
+          // different target elemen is explicitly specified.
+          target_elem_inverse_jacobian_dets[elem->type()] =
+              std::vector<Real>(nq_points, 1.0);
+
+          // Elems deriving from Tri
+          if (dynamic_cast<const Tri*>(elem))
```
roystgnr: Let's rely on the switch statement here?  I know it's not an inner kernel and so this might be the most appropriate idiom but I still want to avoid non-optional RTTI.

src/systems/variational_smoother_system.C
```diff
@@ -113,90 +114,135 @@ void VariationalSmootherSystem::prepare_for_smoothing()
...
+                    // equilateral triangle side length that preserves area of reference
+                    // element
+                    ref_area = target_elem->reference_elem()->volume();
+
+                    break;
+                  }
+
+                case TRI6:
+                  {
+                    target_elem = std::make_unique<Tri6>();
```
roystgnr: You can probably consolidate some of this with `Elem::build(elem->type())`

roystgnr (changes requested):

src/systems/variational_smoother_constraint.C
roystgnr: But you've already instantiated the HEX27 tests!?  I'd have expected that to be the iconic example of the problem.

pbehne: Yeah, so these high order hex tests take a bit to run in dbg mode, so I always ran them in opt and didn't see the assertion hit in dbg until I put this PR up. I don't know what I was thinking when I wrote the assertion. I took it out.

roystgnr: How'd we pass tests in CI?  I could have sworn the 32-bit recipe was running with unit tests enabled and in dbg mode.

pbehne:
> I want to figure out why exactly we weren't tripping that bogus assert before! Even if the library code is right, maybe there's something screwed up in the test coverage?

Is this something you want to check, or want me to look into it?

roystgnr:
> Is this something you want to check, or want me to look into it?

Between errands this morning and tomorrow I'm running stupidly behind, so ... wait a minute.  Emphasis on "stupidly".  That commit was hitting that assertion failure left and right!  E.g.  https://civet.inl.gov/job/3073780/

I could swear I saw it pass but maybe that was only after the assertion-removal commit.

Okay, we'll merge tomorrow morning unless John screams.

src/systems/variational_smoother_system.C
```diff
@@ -195,7 +195,7 @@ void VariationalSmootherSystem::prepare_for_smoothing()
                   // Scale the nodal positions to conserve area
                   auto node_ptr = std::make_unique<Node>(
                       equilateral_points[node_id] * side_length, node_id);
-                  target_elem->set_node(node_id) = node_ptr.get();
+                  target_elem->set_node(node_id, node_ptr.get());
```
jwpeterson: I think I'll make a separate PR with just this fix as well, it shouldn't conflict with #4218 and it will give us a working `--disable-deprecated` build before #4218 just in case that is useful for debugging for some reason...

jwpeterson:
I fixed a conflict with devel (that I introduced, sorry!) and updated the branch. I also haven't reviewed this yet, but I'll do so now.

src/systems/variational_smoother_constraint.C
```diff
@@ -521,6 +521,47 @@ void VariationalSmootherConstraint::constrain_node_to_line(const Node & node,
     }
 }
 
+void VariationalSmootherConstraint::find_nodal_or_face_neighbors(
+    const MeshBase &mesh, const Node &node,
```
jwpeterson: Sorry to continue harping on this, but we put spaces both before and after the `&` and `*`, so here it would be:
```
const MeshBase & mesh, const Node & node,
```
In addition we usually do a hanging indent when there are multiple lines of parameters rather than indenting a fixed four spaces.

I think the clang-format checked in to the repo handles this automatically (?), but I've never used it.

src/systems/variational_smoother_constraint.C
```diff
@@ -521,6 +521,47 @@ void VariationalSmootherConstraint::constrain_node_to_line(const Node & node,
...
+  // Find all the nodal neighbors... that is the nodes directly connected
+  // to this node through one edge.
+  MeshTools::find_nodal_neighbors(mesh, node, nodes_to_elem_map, neighbors);
+
+  // If no neighbors are found, use all nodes on the containing side as
+  // neighbors.
+  if (!neighbors.size())
+    {
+      // Grab the element containing node
+      const auto *elem = nodes_to_elem_map.at(node.id()).front();
```
jwpeterson: Always use `libmesh_map_find()` instead of `map::at()`. The former gives a nicely formatted error message in the exception when the key is not found, the latter just throws a generic std::exception.

src/systems/variational_smoother_constraint.C
```diff
@@ -521,6 +521,47 @@ void VariationalSmootherConstraint::constrain_node_to_line(const Node & node,
...
+      for (const auto &side : elem->side_index_range())
+        {
+          bool found_node = false;
+          const auto &nodes_on_side = elem->nodes_on_side(side);
+          for (const auto &local_node_id : nodes_on_side)
+            if (elem->node_id(local_node_id) == node.id())
+            {
+              found_node = true;
+              break;
+            }
```
jwpeterson: I'd refactor this into a `std::find_if(nodes_on_side.begin(), nodes_on_side.end(), [&](auto local_node_id){...});` call rather than hand-coding the linear search.

src/systems/variational_smoother_system.C
```diff
@@ -113,90 +114,124 @@ void VariationalSmootherSystem::prepare_for_smoothing()
...
-              Node node_1 = Node(s, 0.);
-              Node node_2 = Node(0.5 * s, 0.5 * s * sqrt_3);
-              target_elem.set_node(0, &node_0);
-              target_elem.set_node(1, &node_1);
-              target_elem.set_node(2, &node_2);
+              for (const dof_id_type node_id : index_range(equilateral_points))
+                {
+                  // Scale the nodal positions to conserve area
+                  auto node_ptr = std::make_unique<Node>(
+                      equilateral_points[node_id] * side_length, node_id);
```
jwpeterson: `Node::build()` can be used here, and I'd emplace_back() the new Node directly into `owned_nodes`, catching the reference returned by that to then pass to `target_elem->set_node(...)`. This just saves the step of defining the `node_ptr` temporary variable.

src/systems/variational_smoother_system.C
```diff
@@ -113,90 +114,124 @@ void VariationalSmootherSystem::prepare_for_smoothing()
...
+
+                  _target_inverse_jacobians[target_elem->type()][qp] = H_inv;
+                  target_elem_inverse_jacobian_dets[target_elem->type()][qp] =
+                      H_inv.det();
+                }
+            }// if Tri
+        }// if find == end()
+
+      // Reference volume computation
+      Real elem_integrated_det_J(0.);
```
jwpeterson: Minor comment but I don't think we use this type of initialization for `Real`s elsewhere (we typically use boring `Real foo = 0.;`) so it looks a bit out of place IMO.

tests/mesh/mesh_smoother_test.C
```diff
@@ -101,12 +103,12 @@ private:
                     modulation *= (pj - 0.5) * (pj - 0.5) * 4.; // quadratic bump centered at 0.5
                   }
               }
-            output(i) = p(i) + (std::pow(xi, 3) - xi) * modulation;
+            const auto delta = (std::pow(xi, 3) - xi) * modulation;
```
jwpeterson: Please use `libMesh::Utility::pow<N>` for all integer powers. I know this isn't a new line of code in this PR, but I forgot to mention it on the earlier one. It is just a template function that compiles down to pure multiplication/no function calls at runtime.

tests/mesh/mesh_smoother_test.C
```diff
@@ -269,15 +278,25 @@ public:
   {
     LOG_UNIT_TEST;
 
-    const auto dim = ReferenceElem::get(type).dim();
+    // Get mesh dimension, determine whether type is triangular
+    const auto * ref_elem = &(ReferenceElem::get(type));
+    const auto dim = ref_elem->dim();
+    const bool type_is_tri = dynamic_cast<const Tri*>(ref_elem);
```
jwpeterson, line 286: Not necessary for this PR, but this feels so clunky that we should probably add a helper function that does this by looking only at an `ElemType` enum. Not sure what the best place would be, but maybe a helper function like `enum_to_string()` would probably make sense. It would need to be updated as new Tri types are added, but we have a few things like that already so I wouldn't mind one more.

jwpeterson (changes requested):

tests/mesh/mesh_smoother_test.C
pbehne: Created #4220 for this.

PR #4219, opened 2025-07-30
Fix deprecated Elem::set_node() call

jwpeterson:
This fix is also made in a slightly different form in #4218. Hopefully this fix does not conflict with that one, but even if it does it should be pretty simple to resolve.


jwpeterson:
I attached the two "no-deprecated" builds to this PR. We can probably merge it once those two pass without waiting for all the other testing to also pass.

jwpeterson:
This passed all the "No Deprecated" builds so I'm going to go ahead an merge it, just to see if there's any conflicts with #4218 that I need to then clean up.

PR #4221, opened 2025-07-31
Variational smoother: target element refactor

pbehne:
Moved the target element logic from `prepare_for_smoothing` into static helper functions.

Ref #4082.

### DIFF
--- a/tests/mesh/mesh_smoother_test.C
+++ b/tests/mesh/mesh_smoother_test.C
@@ -160,7 +160,7 @@ class ParallelogramToSquare : public FunctionBase<Real>
   void operator()(const Point & p, const Real, DenseVector<Real> & output)
   {
     output.resize(3);
-    output(0) = p(0) + p(1) / std::sqrt(Real(3).);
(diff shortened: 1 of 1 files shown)


src/systems/variational_smoother_system.C
```diff
@@ -724,4 +628,156 @@ bool VariationalSmootherSystem::element_time_derivative (bool request_jacobian,
...
+VariationalSmootherSystem::get_target_elem(const ElemType & type)
+{
+  // Build target element
+  auto target_elem = Elem::build(type);
+
+  // Volume of reference element
+  const auto ref_vol = target_elem->reference_elem()->volume();
+
+  // Update the nodes of the target element, depending on type
+  const Real sqrt_3 = std::sqrt(3.);
```
roystgnr: Sorry I missed this before - let's make this `std::sqrt(Real(3))`?  Might not matter as much for a mesh smoother, but elsewhere in the library that can make the difference between "converges correctly for higher than double precision" and "ha ha quadruple precision is pointless".

PR #4225, opened 2025-08-04
Linkstamp fix

roystgnr:
I'm hoping the Makefile.am fix here has something to do with our failures on ARM Mac CI recipes.


roystgnr:
Yeah, this looks like it's on track to be the first ARM Mac passes since #4165 was merged.  I shouldn't have just assumed that "Failed but allowed" was another intermittent environment issue.

jwpeterson (approved):

roystgnr:
@loganharbour reports this fixes the problem on his system.  I'd be tempted to merge now, with the ARM Mac recipes reporting success and so many other recipes stuck at "Waiting" for a slot, but if there's one thing I've learned from all this it's to not just assume that a seemingly-straightforward build system change won't have any unexpected side effects...

loganharbour (approved):
🥳 🍻

PR #4226, opened 2025-08-06
VariationalMeshSmoother: Enhancements for Tangled Mesh Handling

pbehne:
This PR introduces several enhancements to the VariationalMeshSmoother system, primarily focused on improving support for tangled meshes and enabling more flexible testability and diagnostics.

*Key Changes*

**Support for Tangled Meshes**

- Enabled handling of tangled meshes by allowing negative Jacobians.
- Set epsilon dynamically based on the minimum determinant value to improve stability during smoothing.

**Mesh Quality Tracking**

- The VariationalSmootherSystem now tracks mesh quality metrics during the smoothing process.
- This includes the ability to detect tangled configurations and adapt internal parameters accordingly.
- Mesh quality info is available throughout the solve process for inspection and validation.

**Refactor of VariationalMeshSmoother**

- Refactoring to expose and persist internal objects (EquationSystems, VariationalSmootherSystem, constraints, and mesh copies).
- A new setup method was added to prepare these objects outside the smooth() method.
- This allows external inspection of the smoother system (e.g., mesh quality pre/post smoothing), which is now leveraged in unit tests.

**Testing Improvements**

- Added new unit tests for tangled mesh scenarios.
- Enhanced multi-subdomain test logic:
  * Previously: constrained subdomain boundary nodes to remain fixed.
  * Now: compares subdomain volumes before and after smoothing, allowing more realistic behavior with sliding boundaries.
- Helps verify preservation of key properties without overly constraining node motion.

**Miscellaneous**

- Set solver absolute tolerance to prevent failures with near-smooth initial guesses.
- Moved target Jacobian generation to a dedicated function.

### DIFF
--- a/include/mesh/mesh_smoother_vsmoother.h
+++ b/include/mesh/mesh_smoother_vsmoother.h
@@ -27,7 +27,10 @@
 #include "libmesh/libmesh_common.h"
 #include "libmesh/mesh_smoother.h"
 #include "libmesh/variational_smoother_system.h"
+#include "libmesh/variational_smoother_constraint.h"
 #include "petsc_diff_solver.h"
+#include "libmesh/distributed_mesh.h"
+#include "libmesh/equation_systems.h"
 
 // C++ Includes
 #include <cstddef>
@@ -79,6 +82,12 @@ public:
   virtual ~VariationalMeshSmoother() = default;
 
   /**
+   * Setup method that creates equation systems, system, and constraints, to be
+   * called just prior to smoothing.
+   */
+  virtual void setup();
+
+  /**
    * Redefinition of the smooth function from the
    * base class.  All this does is call the smooth
    * function in this class which takes an int, using
@@ -93,6 +102,10 @@ public:
    */
   void smooth(unsigned int n_iterations);
 
+  /**
+   * Getter for the _system attribute
+   */
+  VariationalSmootherSystem * get_system() { return _system; }
 
 private:
 
@@ -101,8 +114,35 @@ private:
    */
   const Real _dilation_weight;
 
-  /// Whether subdomain boundaries are subject to change via smoothing
+  /**
+   * Whether subdomain boundaries are subject to change via smoothing
+   */
   const bool _preserve_subdomain_boundaries;
+
+  /**
+   * EquationsSystems object associated with the smoother
+   */
+  std::unique_ptr<EquationSystems> _equation_systems;
+
+  /**
+   * System used to smooth the mesh.
+   */
+  VariationalSmootherSystem * _system;
+
+  /**
+   * Constraints imposed on the smoothing process.
+   */
+  std::unique_ptr<VariationalSmootherConstraint> _constraint;
+
+  /**
+   * Mesh copy to avoid multiple EquationSystems.
+   */
+  std::unique_ptr<DistributedMesh> _mesh_copy;
+
+  /**
+   * Attribute the keep track of whether the setup method has been called.
+   */
+  bool _setup_called;
 };
 
 } // namespace libMesh
--- a/src/systems/variational_smoother_system.C
+++ b/src/systems/variational_smoother_system.C
@@ -132,7 +132,15 @@ void VariationalSmootherSystem::assembly (bool get_residual,
   compute_mesh_quality_info();
 
   // Update _epsilon_squared_assembly based on whether the mesh is tangled
-  _epsilon_squared_assembly = _mesh_info.mesh_is_tangled ? _epsilon_squared : 0.;
+  if (_mesh_info.mesh_is_tangled)
+    {
+      const Real & min_S = _mesh_info.min_qp_det_S;
+      // This component is based on what Larisa did in the original code.
+      const Real variable_component = 100. * Utility::pow<2>(_ref_vol * min_S);
+      _epsilon_squared_assembly = _epsilon_squared + variable_component;
+    }
+  else
+    _epsilon_squared_assembly = 0.;
 
   FEMSystem::assembly(get_residual, get_jacobian, apply_heterogeneous_constraints, apply_no_constraints);
 }
--- a/src/systems/variational_smoother_system.C
+++ b/src/systems/variational_smoother_system.C
@@ -168,6 +168,9 @@ void VariationalSmootherSystem::init_context(DiffContext & context)
       fe_map.get_dxyzdzeta();
       fe_map.get_JxW();
 
+      // Mesh may be tangled, allow negative Jacobians
+      fe_map.set_jacobian_tolerance(std::numeric_limits<Real>::lowest());
+
       c.get_side_fe( 0, my_fe, dim );
       my_fe->get_nothing();
     }
--- a/src/systems/variational_smoother_system.C
+++ b/src/systems/variational_smoother_system.C
@@ -39,6 +39,14 @@ namespace libMesh
 {
 
 /**
+ * Function to prevent dividing by zero for degenerate elements
+ */
+Real chi_epsilon(const Real & x, const Real epsilon_squared)
+{
+  return 0.5 * (x + std::sqrt(epsilon_squared + Utility::pow<2>(x)));
+}
+
+/**
  * Given an fe_map, element dimension, and quadrature point index, returns the
  * Jacobian of the physical-to-reference mapping.
  */
@@ -353,7 +361,7 @@ bool VariationalSmootherSystem::element_time_derivative (bool request_jacobian,
                          0, 0, 1);
 
       // The chi function allows us to handle degenerate elements
-      const Real chi = 0.5 * (det + std::sqrt(_epsilon_squared_assembly + det_sq));
+      const auto chi = chi_epsilon(det, _epsilon_squared_assembly);
       const Real chi_sq = chi * chi;
       const Real sqrt_term = std::sqrt(_epsilon_squared_assembly + det_sq);
       // dchi(x) / dx
@@ -718,7 +726,7 @@ void VariationalSmootherSystem::compute_mesh_quality_info()
           const auto tr = trace(S.transpose() * S, dim);
 
           // The chi function allows us to handle degenerate elements
-          const Real chi = 0.5 * (det + std::sqrt(_epsilon_squared_assembly + det_sq));
+          const auto chi = chi_epsilon(det, _epsilon_squared_assembly);
 
           // distortion
           const Real beta = std::pow(tr / dim, half_dim) / chi;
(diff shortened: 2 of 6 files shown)


include/mesh/mesh_smoother_vsmoother.h
```diff
@@ -93,6 +102,10 @@ class VariationalMeshSmoother : public MeshSmoother
    */
   void smooth(unsigned int n_iterations);
 
+  /**
+   * Getter for the _system attribute
+   */
+  VariationalSmootherSystem * get_system() { return _system; }
```
roystgnr: This exposes a lot.  Would a shim to `compute_mesh_quality_info()` be sufficient?

roystgnr: Or maybe just to `get_mesh_info()`, if we either keep our settings such that we always check the final residual (and hence compute a final quality info in that last assembly) or add a manual `compute_mesh_quality_info()` after our solve?

include/mesh/mesh_smoother_vsmoother.h
```diff
@@ -101,8 +114,35 @@ class VariationalMeshSmoother : public MeshSmoother
...
+
+  /**
+   * Constraints imposed on the smoothing process.
+   */
+  std::unique_ptr<VariationalSmootherConstraint> _constraint;
+
+  /**
+   * Mesh copy to avoid multiple EquationSystems.
+   */
+  std::unique_ptr<DistributedMesh> _mesh_copy;
```
roystgnr: If we put this in the class then forgetting to destruct the class gets a lot more expensive.  If we can get away with just exposing the mesh_info, can we move this back to a local variable?

src/systems/variational_smoother_system.C
```diff
@@ -628,6 +670,133 @@ bool VariationalSmootherSystem::element_time_derivative (bool request_jacobian,
...
+      info.total_combined += combined_int;
+      if (combined_int > info.max_elem_combined.second)
+        info.max_elem_combined = std::make_pair(elem->id(), combined_int);
+      else if (combined_int < info.min_elem_combined.second)
+        info.min_elem_combined = std::make_pair(elem->id(), combined_int);
+
+    } // for elem
+
+  // Get contributions from elements on other processors
+  mesh.comm().max(info.max_elem_det_S);
```
roystgnr: Does this work??  If you're using `pair<dof_id_type, Real>`, and we're defaulting to lexicographic ordering, then aren't you just picking whichever processor's local max has the higher id rather than which has the higher max?

Also we definitely need some regression coverage for this in the unit tests.  If I'm right about this being a bug, an assert could have caught it in one of the processor sweep recipes, because a "gold" value in serial would then differ too greatly in parallel.

tests/mesh/mesh_smoother_test.C
```diff
@@ -275,31 +265,23 @@ public:
...
                                const ElemType type,
-                               const bool multiple_subdomains = false)
+                               const bool multiple_subdomains = false,
+                               const bool tangle_mesh=false)
   {
     LOG_UNIT_TEST;
 
+    libmesh_error_msg_if(
+        multiple_subdomains && tangle_mesh,
+        "Arbitrary mesh tangling with multiple subdomains is not yet supported.");
```
roystgnr: This is the sort of thing that ought to be a `libmesh_deprecated_msg()` where users can be informed by it, not just here to rescue developers.

roystgnr: I of course mean `libmesh_not_implmented_msg()`.

When I was young, an incident of getting two simple related-but-very-different words solidly switched around in my head for no apparent reason was one of my most embarrassing memories.  These days I treasure that memory, as proof that I'm not going senile, I just always had a couple of screws loose.

src/systems/variational_smoother_system.C
pbehne: Nice catch. I doubt it works as I intended. For now, I'll change the `std::pair<dof_id_type, Real>` to `Real` because we don't currently use the elem id information.

include/mesh/mesh_smoother_vsmoother.h
pbehne: The reason I split `VariationalMeshSmoother::smooth` into `setup` and `smooth` is so we can compute mesh quality info BEFORE attempting to solve the system. Instead of having only the `VariationalSmootherSystem::_mesh_info` attribute, I had `_mesh_info_presmoothing` and `_mesh_info_postsmoothing`, where the later is computed and saved at the end of `prepare_for_smoothing`. This approach would allow us to recombine `setup` and `smooth` and avoid having `mesh_copy` as an attribute. The downside is that we can't look at `_mesh_info_presmoothing` until we've actually called `solve`, which I think is silly. But, it seems like keeping `mesh_copy` as a local variable might justify that structure.

Based on this, what would you prefer?

src/systems/variational_smoother_system.C
roystgnr: Could you just change it to `pair<Real, dof_id_type>`?  I do think it's sometimes super useful to be able to examine just how bad the worst element in a mesh is.

include/mesh/mesh_smoother_vsmoother.h
roystgnr: I really like computing `_mesh_info_presmoothing` in `setup()` (either in advance by the user or in `solve()` if the user didn't call it, as you're doing with that flag) and `_mesh_info_postsmoothing` at the end of `smooth()`.  I agree that not being able to look at `_mesh_info_presmoothing` before `solve()` is silly, but letting the user manually call `setup()` fixes that ...

In fact, in the back of my mind I'm wondering whether we should just turn the most generally useful of these calculations into `ElemQuality` metric options and make the quality queries available independent of any smoother code.  Maybe not in this PR.  ;-)

pbehne: So should we leave this as-is for now?

src/systems/variational_smoother_system.C
```diff
@@ -802,6 +802,10 @@ void VariationalSmootherSystem::compute_mesh_quality_info()
   mesh.comm().min(info.min_elem_combined);
   mesh.comm().sum(info.total_combined);
 
+  unsigned int mesh_is_tangled_int = info.mesh_is_tangled;
```
roystgnr: Why a uint here?  Just use a `bool` with `max` instead of `sum`.

tests/mesh/mesh_smoother_test.C
```diff
@@ -272,9 +272,9 @@ public:
   {
     LOG_UNIT_TEST;
 
-    libmesh_error_msg_if(
-        multiple_subdomains && tangle_mesh,
-        "Arbitrary mesh tangling with multiple subdomains is not yet supported.");
+    if (multiple_subdomains && tangle_mesh)
+      libmesh_not_implemented_msg(
```
roystgnr, line 278: I think I emphasized the wrong point here - using a more-specific error call is preferable to using a less-specific one, sure, but the big thing is that the error message is in the wrong place.  We don't need to protect developers from overusing the unit test refactoring here; we need to protect users from attempting an unsupported operation without getting a sane error message out of it.  The `libmesh_not_implemented_msg` ought to be checking for multiple subdomains in the `mesh_is_tangled` case in `VariationalSmootherSystem::assembly` in `src/`; if we put anything in `tests/` it would just be a `try/catch` to make sure that the error message gets triggered properly.

pbehne: I think that in general, if a user has a tangled mesh with multiple subdomains, the smoother should be able to handle it in the same manner as a single-subdomain mesh. If `preserve_subdomain_boundaries=true`, then subdomain boundary nodes will be constrained and smoother will try to move non-subdomain boundary nodes around to untangle. If `preserve_subdomain_boundaries=false`, then who cares, this is the same as untangling a single subdomain mesh. The main reason I put the check in the unit test is that I didn't want to come up with an intelligent way to tangle a mesh with multiple subdomains without touching the subdomain boundary nodes. Doing so would have the side effect that this assertion is tripped:

```
// Make sure our DistortHyperCube transformation has distorted the mesh
for (auto node : mesh.node_ptr_range())
  CPPUNIT_ASSERT(distortion_is(*node, true));

```

include/mesh/mesh_smoother_vsmoother.h
```diff
@@ -65,8 +65,9 @@ class UnstructuredMesh;
  *
  * Notes:
  *
- * 1) The smoother supports tangled meshes. However, it cannot untangle all such
- * meshes. This will be investigated further in the future.
+ * 1) The smoother supports tangled meshes. However, not all untangling solves
+ * converge. Some will converge to a local optima, resulting in an untangled,
+ * but still suboptimal mesh that cannot be smoothed much further.
```
roystgnr: Wait what now?  I thought our problem was that there were some meshes that we couldn't untangle.  If there are meshes that we can untangle but that don't converge to a global optimum afterward, doesn't that mean that not all solves on *untangled* meshes converge?

include/mesh/mesh_smoother_vsmoother.h
roystgnr: Sorry I missed this reply.  I was thinking that "make the quality queries available independent of any smoother code" shouldn't go in this PR, but I was still hoping we could pull off "make the temporary copy of the mesh go away when we're done using it, not when user code remembers to destruct the smoother".

include/mesh/mesh_smoother_vsmoother.h
pbehne: I tried to be ambiguous here on what we can and cannot do as a way to say "no promises"

> doesn't that mean that not all solves on untangled meshes converge?

For tangled meshes, we either
1. succeed in untangling the mesh (converge to SOME optima).
2. fail in untangling the mesh (fail to converge to ANY optima).

If we succeed in untangling, we either
1. succeed in smoothing the mesh (converge to a GLOBAL optima)
2. fail in smoothing the mesh (fail to converge to a GLOBAL optima). This could possibly happen if we untangle to a local optima.

Of course, any failure to converge to any optima at all could be caused by bad numerics (super small jacobians, etc....)

Should I update the documentation to be more like my above comment?

include/mesh/mesh_smoother_vsmoother.h
pbehne: See my proposed workaround in f628328.

pbehne: > make the quality queries available independent of any smoother code

I've added this as a miscellaneous task to #4082.

include/mesh/mesh_smoother_vsmoother.h
pbehne: I'll add another thought to my comment above.

In the case that a tangled mesh is untangled to a non-global optima, I think it is quite possible that it cannot be smoothed much further, especially if the smooth global optima cannot be reached without traveling through tangled mesh space.

pbehne: Well, after trying to find a case where we untangle a mesh, but smooth it to a local optima, I am prepared to eat my words. I couldn't find such a case occurring after I patched all the bugs. I'll update the wording in the documentation.

roystgnr (approved):

roystgnr:
MOOSE test failures are downstream and unrelated; we're still trying to diagnose them.

PR #4228, opened 2025-08-14
hp-Adaptivity: Smoothness percentage control flexibility

skvarjun:
User can control the percentage of smoothness values above which p-refinement is favored in hp-adaptivity. For instance, default value is set as 0.9, means p refining favored on an element only if local smoothness is above 90% when compared to the reported smoothness across all other elements.
Important change is that an option argument `Real smoothness_threshold = 0.9` is added to base class of `select_refinement()`.

### DIFF
--- a/src/utils/error_vector.C
+++ b/src/utils/error_vector.C
@@ -52,14 +52,9 @@ ErrorVectorReal ErrorVector::minimum() const
 
   for (dof_id_type i=0; i<n; i++)
     {
-      // Only positive (or zero) values in the error vector
-      libmesh_assert_greater_equal ((*this)[i], 0.);
       if (this->is_active_elem(i))
         min = std::min (min, (*this)[i]);
     }
-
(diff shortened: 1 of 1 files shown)


include/error_estimation/hp_selector.h
```diff
@@ -70,7 +70,7 @@ class HPSelector
    * refinement and potentially change the desired
    * refinement type.
    */
-  virtual void select_refinement (System & system) = 0;
+  virtual void select_refinement (System & system, Real smoothness_threshold = 0.9) = 0;
```
roystgnr, line 73: Can we move `smoothness_threshold` into a member variable of the subclass with an accessor?  It's only appropriate in the base class if it's going to be used by every possible subclass, and there are at least some selection possibilities that don't rely on a posteriori smoothness testing.

skvarjun: Yes, I agree with this.

PR #4230, opened 2025-08-19
Netgen interface performance logging

roystgnr:
I was hoping to get some improvements (in the valgrind case) out of this, and I couldn't, but the logging alone is still better included than omitted.


roystgnr:
Test failures are an unrelated problem downstream.

PR #4232, opened 2025-08-21
Variational Smoothing for Pyramids

pbehne:
Added support to the `VariationalMeshSmoother` for pyramids.

Ref. #4082

### DIFF
--- a/src/systems/variational_smoother_system.C
+++ b/src/systems/variational_smoother_system.C
@@ -931,7 +931,7 @@ VariationalSmootherSystem::get_target_elem(const ElemType & type)
   else if (type_str.compare(0, 7, "PYRAMID") == 0)
     {
 
-      // Ideally, the target element would be a pyramid with an square base and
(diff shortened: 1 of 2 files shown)


pbehne:
From the failing min clang recipe:

```
1) test: MeshSmootherTest::testVariationalHex27Tangled (E) 
uncaught exception of type std::exception (or derived).
- Unrecoverable failure to converge
```

Funny, it converges on my machine :)

roystgnr:
The "Test MOOSE" (inc. clang) issues are downstream and unrelated.

The prior "Test with threads" timeouts still worry me.  Getting through unit tests in CI opt mode should take like 20 minutes tops; we shouldn't be coming anywhere near a 60 minute timeout.  Cutting the mesh sizes from 4^n to 2^n in the case of higher order elements gets us down to 26.5 minutes, which seems fine, but I may kick that a few times before we merge just to see what the variation is.

PR #4234, opened 2025-08-26
Fix for the case when we encounter linearly-dependent data in EIM training

dknez:


jwpeterson (approved):
This looks OK to me, but I think we are going to test it locally and make sure it fixes the bug we are seeing before merging.

Speaking of testing, it looks like none of the CIVET tests are running for this currently. @roystgnr any chance that CIVET is down/experiencing issues at the moment?

PR #4237, opened 2025-08-27
Make libMesh::out and libMesh::err thread safe

lindsayad:
This is based on @roystgnr's idea to make thread local buffers that we send to the wrapped output sink whenever we receive a flush indicator, at which time we lock while writing the the thread local buffer to the shared sink. This also defaults to flushing if the end of an operation is `\n`

### DIFF
--- a/src/base/libmesh.C
+++ b/src/base/libmesh.C
@@ -29,6 +29,7 @@
 #include "libmesh/print_trace.h"
 #include "libmesh/enum_solver_package.h"
 #include "libmesh/perf_log.h"
+#include "libmesh/thread_buffered_syncbuf.h"
 
 // TIMPI includes
 #include "timpi/communicator.h"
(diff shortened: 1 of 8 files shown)


lindsayad:
*LOL* still getting those damn non-unicode failures

lindsayad:
what a joke/time-suck

lindsayad:
I'm going to try locking in the get pot search because that's the other thing that shows tsan races during `print_trace`. If that doesn't do it that will suck

src/base/libmesh.C
```diff
@@ -220,6 +221,54 @@ int                GLOBAL_COMM_WORLD = 0;
...
+std::streambuf * _out_prewrap_buf = nullptr;
+std::streambuf * _err_prewrap_buf = nullptr;
+
+// Install after redirection/drop decisions
+void install_thread_buffered_sync()
+{
+  // libMesh::out
+  if (auto * ob = libMesh::out.rdbuf(); ob)
+    {
+      _out_prewrap_buf = ob;
```
roystgnr, line 244: If we call this twice in a row we've just lost the actual pre-wrap buffer.  I know you're not doing that and not exposing this API, but could we throw in an assert here in case someone refactors or shims to it later?

roystgnr (approved):
I am happy with this, but it is a lot of new complication; since it's not actually *fixing* the `non-unicode characters in output` failures I think I'd rather wait until we find the real fix (or real combination of fixes, if this fixes some cases and something else fixes the rest) before we merge.

lindsayad:
> I am happy with this, but it is a lot of new complication; since it's not actually _fixing_ the `non-unicode characters in output` failures I think I'd rather wait until we find the real fix (or real combination of fixes, if this fixes some cases and something else fixes the rest) before we merge.

100% agree. TBH if something else fixes the non-unicode characters stand-alone I would consider closing this PR. What are your thoughts? This certainly fixes real thread races. I think the question is whether there are real consequences from the thread races this PR fixes or if the races are benign enough that the added complexity and possible performance reduction from this PR are worth it?

roystgnr:
I'm fine with the performance reduction; it's just console output.  I could go either way on the complexity-vs-value question, though.

Actually, here's where my thoughts are leaning: could we just make the thread-buffered sync *optional*?  The complexity is encapsulated pretty well, so I'm only worried about it being a red herring if we have issues with console output in the future, and if we can just turn it off with a command line option then that's not a worry.

roystgnr (changes requested):
Let's make `could we just make the thread-buffered sync optional?` (where, to be clear, I'd be happy to default the option to "on") an official review.  I could be talked into merging this as-is if John thinks the extra option is just pointless worry, though.

roystgnr:
As of this PR, we now seem to be segfaulting whenever we throw an exception and exit.  @bwspenc was the first to notice (or at least the first to be smart enough to say something about?) the effects on `mooseError()` calls, but it affects both libMesh and MOOSE exceptions, both Linux with libstdc++ and Mac with libc++, both my homebuilt libraries and the usual Conda packages.

On either sort of system, after the terminate() calls MPI_Abort, when we're getting into iostreams cleanup, a library-specific destructor makes a `basic_ostream` `flush()` call that segfaults.

I'm kind of baffled as to *why* it would be segfaulting, though.  When we hit the `libmesh_terminate_handler()`, it should be after stack unwinding, therefore after the `~LibMeshInit()` destructor, therefore after we've done any necessary cleanup. ... except, maybe it doesn't?  If I set gdb breakpoints in `libmesh_terminate_handler()` and `~LibMeshInit()` and run something that throws an exception, I'm only seeing the terminate handler, not the LibMeshInit destructor ... and now that I think about it, we'd be in trouble if we *did* hit `~LibMeshInit` when unwinding, because the first thing that does is a `barrier()` and there's no way an unhandled exception is going to be thrown on every rank in parallel in general.

Oh!  ["It is implementation-defined whether any stack unwinding takes place for uncaught exceptions."](https://en.cppreference.com/w/cpp/language/throw.html)  So a C++ compiler is within its rights to notice that we've got no catch block and not bother with unwinding when we throw an error that will just hit the terminate handler.  We ought to be testing `std::uncaught_exceptions()` in `~LibMeshInit` so we don't get stuck on that barrier if we are unwinding, and we ought to be doing any necessary iostreams cleanup in `libmesh_terminate_handler` so we don't corrupt the OS cleanup if we didn't unwind.

PR #4239, opened 2025-08-30
Skip remote children in bracketing_nodes search

roystgnr:
This fixes a downstream failure for me doing adaptivity with Quad8 elements, where the nasty search we do to find bracketing nodes of incomplete-order elmeents can hit remote children if they're ghosted.

I'm not completely confident this fixes the problem without just opening us up to even harder-to-trigger problems, but it's at least an improvement.

Thanks to @skvarjun for catching this.


roystgnr:
Downstream failure unrelated.

I really need to add some Quad8/Hex20/Hex26 coverage at the libMesh level, but after this gets downstream we'll at least have coverage from Arjun's PR there.

PR #4240, opened 2025-09-04
Set PetscVector ParallelType in Vec constructor

jwpeterson:
This mimics the pre-3.21.0 behavior by setting the ParallelType to PARALLEL in the case where it does identify as "MPI" but is not GHOSTED.


jwpeterson:
This fixes an error for me, and looks like it passes ~most~ all of the CIVET testing, so I'd be in favor of merging it. As @roystgnr points out over slack, it might be good to replace our
```
  if (std::strcmp(ptype, VECMPI) == 0)
```
with a check on the wrapped Vec's communicator size, since that is in general accessible from PETSc objects by calling:
```
PetscObjectGetComm((PetscObject)foo, &comm);
```

PR #4241, opened 2025-09-04
Use MatGetValue instead of MatGetRow in PetscMatrix::operator()

lindsayad:
MatGetValue already does what I am removing in this PR. And I am fixing MatGetValue for a particular PETSc matrix type in [petsc/petsc#6892](https://gitlab.com/petsc/petsc/-/merge_requests/8692) so it seems wasteful to implement the fix twice. It also feels much more logical to call this targeted API

### DIFF
--- a/src/numerics/petsc_matrix.C
+++ b/src/numerics/petsc_matrix.C
@@ -1123,15 +1123,10 @@ T PetscMatrix<T>::operator () (const numeric_index_type i_in,
 {
   libmesh_assert (this->initialized());
 
-  // PETSc 2.2.1 & newer
-  const PetscScalar * petsc_row;
-  const PetscInt    * petsc_cols;
-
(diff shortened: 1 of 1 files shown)


lindsayad:
I don't know. I looked for MatGetValue in our git history and didn't see anything. I was hoping to see that and then a later commit removing it for X reason. MatGetValue has been present in PETSc since at least 2007 when it was converted from a macro to an inline function. I was thinking maybe it was a recent add but clearly not

lindsayad:
The `MatGetRow` exists in this method from the initial import of source in 2003

lindsayad:
And I now found that `MatGetValue` has existed in PETSc since 1997. So I think my conclusion is IDK

lindsayad:
Well if we find that there is a reason, maybe in some downstream breakage, then we can always revert

PR #4242, opened 2025-09-05
SparseMatrix::read_coreform_hdf5()

roystgnr:
This is a little classier than reading matrices from an ASCII matlab script, much more portable than 32-vs-64-bit-specific PETSc binary formats, and ought to be easier to integrate into (version 8 HDF5-based) ExodusII file input in the future.


jwpeterson (approved):
I chatted with @roystgnr on slack, and he's planning to eventually add support for writing these types of files as well, which should help close the loop on this as a feature.

PR #4243, opened 2025-09-07
Transpose + norm unit tests, Laspack implementations, Eigen fixes

roystgnr:
This adds implementations to allow us to expand the test coverage of the implementation I added to allow us to expand the test coverage of the new code I actually cared about.

### DIFF
--- a/src/numerics/laspack_matrix.C
+++ b/src/numerics/laspack_matrix.C
@@ -412,8 +412,8 @@ void LaspackMatrix<T>::get_transpose (SparseMatrix<T> & dest) const
   for (numeric_index_type col : make_range(N))
     {
       auto len = col_offsets[col];
-      libmesh_assert_equal_to(target._row_start[col+1] -
-                              target._row_start[col], len);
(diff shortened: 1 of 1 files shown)


jwpeterson (approved):

roystgnr:
I forgot to add the `Test Non-LGPL Contrib` recipe earlier, but if/when that passes I'll merge.

roystgnr:
Well, hooray for being paranoid about testing.  I'd been working with `--disable-strict-lgpl`, of course, but forgot to combine it with `--enable-werror`.

jwpeterson:
FYI @roystgnr this PR looks like it has the same issue where the yellow status symbol doesn't turn green even though all the tests are green. So I'd say merge whenever you are ready...

jwpeterson:
I'd also like to pull these changes into the `branch_v1.9.x` that was recently started. I could even be convinced to try and port them to 1.8.x, at least the Eigen fixes seem like they are important.

roystgnr:
Those last two running recipes passed, just didn't report it.

PR #4244, opened 2025-09-07
Initialize epsilon_squared_assembly

roystgnr:
Otherwise it doesn't get initialized until the first time we test whether the mesh is tangled, but since we do that test at the same time as we calculate mesh info including element-wise residual contributions, those calculations use uninitialized data.

IMHO what we need here is to separate out a tangling test from the rest of the mesh info calculations, but I'll leave that for Patrick to consider and just fix valgrind's screaming for now.


roystgnr:
That cleaned up the VariationalSmoother unit tests as far as valgrind is concerned.

PR #4245, opened 2025-09-08
Don't double-create a PetscViewer

roystgnr:
The docs for PetscViewerASCIIOpen say that it does the PetscViewerCreate, and gdb agrees, and doing a redundant PetscViewerCreate makes valgrind yell at us about a memory leak.

### DIFF
--- a/src/numerics/petsc_matrix.C
+++ b/src/numerics/petsc_matrix.C
@@ -521,14 +521,12 @@ void PetscMatrix<T>::print_matlab (const std::string & name) const
       const_cast<PetscMatrix<T> *>(this)->close();
     }
 
-  WrappedPetsc<PetscViewer> petsc_viewer;
-  LibmeshPetscCall(PetscViewerCreate (this->comm().get(),
-                                      petsc_viewer.get()));
-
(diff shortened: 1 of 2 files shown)


roystgnr:
> My only thought was whether it has always been the case that PETSc itself calls PetscViewerCreate, or if there were some earlier versions where they expected you to call it?

`git log -p --follow src/sys/classes/viewer/impls/ascii/vcreatea.c` says they were doing `PetscNew(Viewer_ASCII)` in the very first commit that renamed `ViewerFileOpenASCII` to `ViewerASCIIOpen`, in 1998.

> Presumably we copied this code from some example somewhere

Maybe?  Our vector case looks like it was copied from our matrix case, but no telling where Ben got that from - it was part of the massive development in Jan 2003, interleaved between other commit messages like "will work as a first cut", so I'd assume he just made a reasonable misinterpretation of a doc page.

roystgnr:
The leak was just 640 bytes per call, too, nothing anyone would ever notice unless they had test coverage on that method (we didn't) which they were trying to get valgrind-clean (we weren't).

PR #4246, opened 2025-09-09
Make prepare_for_use optional after stitching

roystgnr:
If we stitch 2 meshes together this is a handy user-friendly default; if we stitch 200 meshes together this is an O(N^2) performance issue, with a nasty constant in parallel for repartitioning (including our defaults) that begins with a libHilbert sweep.

I still need to add test coverage to this, but I'm putting it up now to make sure it doesn't already make anything scream.


roystgnr:
After toggling this new option in one of the MOOSE reactor mesh generators, the "pessimized" mesh generation problem @YaqiWang sent me runs 6x faster.  I think a slightly more extensive MOOSE change should improve that further.

jwpeterson (approved):
Longer term I think we should refactor the stitching function into an object... don't functions with  10+ parameters, many of which are defaulted and/or for niche use cases.

PR #4249, opened 2025-09-10
Avoid memory leak in unit test driver

roystgnr:
If we add only a subset of tests to the runner (via --re or --deny_re options), we're careful to move the other tests to a "rejects" suite so they'll get deleted there, but the "supertest", the suite holding all the other tests, wasn't getting deleted in those cases.

This (on top of my earlier fixes, and using the MOOSE suppressions file for third-party library issues) gets selective valgrind runs of our unit tests clean for me.  (all-tests runs are clean either way)

This isn't urgent to merge; I still need to run through our examples to look for valgrind issues too.  This tiny leak is *only* a problem because it upsets valgrind, and until we're sure we're ready to make that be a big deal (add a `--error-exitcode=` option to our valgrind recipes), "do our unit tests upset valgrind" isn't a critical question.

### DIFF
--- a/tests/driver.C
+++ b/tests/driver.C
@@ -56,6 +56,9 @@ private:
   CppUnit::Test & _shimmed_test;
 };
 
+// "Magic number" for when we're not doing allow or deny regex
+static constexpr int added_whole_suite = -12345;
+
 // Add Tests to runner that match user-provided regex.
 int add_matching_tests_to_runner(CppUnit::Test * test,
                                  const std::string & allow_r_str,
@@ -72,7 +75,7 @@ int add_matching_tests_to_runner(CppUnit::Test * test,
   {
     libMesh::out << test->getName() << std::endl;
     runner.addTest(test);
-    return -12345;
+    return added_whole_suite;
   }
 
   if (test->getChildTestCount() == 0)
@@ -182,7 +185,7 @@ int main(int argc, char ** argv)
 
   // If we didn't add the whole suite to the runner, we need to clean
   // it up ourselves
-  if (n_tests_added != -12345)
+  if (n_tests_added != added_whole_suite)
     owned_suite.reset(suite);
 #else
   // If no C++11 <regex> just run all the tests.


roystgnr:
Scratch "gets selective valgrind runs of our unit tests clean for me" - I'm still seeing something else, and I'm not sure if it's just something I missed before or a regression from this PR.

tests/driver.C
```diff
@@ -134,15 +140,17 @@ int main(int argc, char ** argv)
...
   libMesh::out << "Will run the following tests:" << std::endl;
   const int n_tests_added =
-    add_matching_tests_to_runner(registry.makeTest(),
+    add_matching_tests_to_runner(suite,
                                  allow_regex_string, allow_regex,
                                  deny_regex_string, deny_regex,
                                  runner, rejects);
   if (n_tests_added >= 0)
     libMesh::out << "--- Running " << n_tests_added << " tests in total." << std::endl;
+  if (n_tests_added != -12345)
```
jwpeterson: Minor comment, but since this is our own magic number that we now actually have to refer to, it might make sense to use `libMesh::invalid_uint` or some other named constant.

tests/driver.C
```diff
@@ -134,15 +140,17 @@ int main(int argc, char ** argv)
...
+    owned_suite.reset(suite);
```
jwpeterson, line 189: I guess I don't understand why we don't need to clean up `suite` when there's no tests added? From a surface level reading of the code it just looks like `registry.makeTest()` returns a dumb pointer whose lifetime we are expected to manage, and we were always leaking it before...

jwpeterson (approved):
I had some questions that could probably be cleared up if I read the cppunit docs, but I chose not to :man_shrugging:

tests/driver.C
roystgnr: This should clean up `suite` when there's no tests added.  That'll return `n_tests_added == 0`, and `0 != -12345`, so we put `suite` in our `unique_ptr` and it gets cleaned up.

roystgnr:
There's definitely something weird going on here.  This patch cleans up the leaked suite from `TestFactoryRegistry::makeTest()`, but it starts complaining about a test suite destructor hitting invalid accesses to already-freed memory from `TestSuiteFactory<AllSecondOrderTest>::makeTest()` in particular.

If *every* test class was complaining then I'd be sure I'm trying to do a double-free here, once from the runner and then once from recursive destruction from the suite, but it's *just* `AllSecondOrderTest` complaining?

No, wait, it's too much of a coincidence that the failing `AllSecondOrderTest` is first alphabetically.  I must be trying to do a double-free here, but something about the first UB causes the destructor to skip the rest.

roystgnr:
I think I've managed to square the circle of "we can't add a test to a runner without passing ownership of the pointer to the runner" and "when we get a test from a suite there's no way to take ownership of it away from the suite": we get each test from the suite, we wrap it in our own `TestShim` that passes along every API call *except* the destructor, and we hand the shim to the runner.  The runner deletes the shims, then we can delete the suite to delete the non-shimmed tests, and nothing gets destroyed 2 times or 0 times.

I've added our `Valgrind (unit tests)` recipe, and I'm testing with valgrind myself; if nothing screams then this will finally be ready to merge.

PR #4250, opened 2025-09-15
Some array variable support

lindsayad:
Also reuses variable and group storage from the System in the DofMap

### DIFF
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -676,6 +676,13 @@ public:
    * that can be leveraged by the \p DofMap to build degrees of freedom
    * containers corresponding to all the variables in this variable array
    *
+   * An 'array variable' is simply a sequence
+   * of contiguous variable numbers defined by pair where the first member of the pair
+   * is the first number in the variable sequence and the second member of the pair is
+   * the number of the last variable in the sequence plus one. Array variables may be
+   * used in tandem with variable grouping by downstream code to build optimized physics
+   * kernels since each variable in the array will have the same shape functions.
+   *
    * \returns The index number for the last of the new variables.
    */
   unsigned int add_variable_array (System & sys,
@@ -1803,7 +1810,16 @@ private:
 
   /**
    * Retrieve the array variable bounds for a given variable \p vi. This variable may
-   * lie anywhere within an array variable range
+   * lie anywhere within an array variable range. An 'array variable' is simply a sequence
+   * of contiguous variable numbers defined by pair where the first member of the pair
+   * is the first number in the variable sequence and the second member of the pair is
+   * the number of the last variable in the sequence plus one. Array variables may be
+   * used in tandem with variable grouping by downstream code to build optimized physics
+   * kernels since each variable in the array will have the same shape functions.
+   *
+   * We note that we store array variables as a container of the above described pairs. Within
+   * this API we will do a binary search such that the complexity is O(log(N)) where N is the
+   * number of array variables present in \p this
    */
   const std::pair<unsigned int, unsigned int> &
   get_variable_array(unsigned int vi) const;
--- a/include/systems/system.h
+++ b/include/systems/system.h
@@ -1172,6 +1172,13 @@ public:
    * that can be leveraged by the \p DofMap to build degrees of freedom
    * containers corresponding to all the variables in this variable array
    *
+   * An 'array variable' is simply a sequence
+   * of contiguous variable numbers defined by pair where the first member of the pair
+   * is the first number in the variable sequence and the second member of the pair is
+   * the number of the last variable in the sequence plus one. Array variables may be
+   * used in tandem with variable grouping by downstream code to build optimized physics
+   * kernels since each variable in the array will have the same shape functions.
+   *
    * \returns The index number for the last of the new variables.
    */
   unsigned int add_variable_array (const std::vector<std::string> & vars,


include/base/dof_map.h
```diff
@@ -572,17 +572,25 @@ class DofMap : public DofMapBase,
...
+    VariableKey() = default;
+    VariableKey(const VariableKey &) = delete;
+    VariableKey(VariableKey &&) = delete;
+  };
+
+  /**
+   * Add our knowledge of variables and their groups
+   */
+  void
+  add_variables_and_groups(const std::vector<Variable> & variables, const std::vector<VariableGroup> & variable_groups, VariableKey &&);
```
loganharbour:
```
namespace libMesh {
struct pass_key ...
}
```

time?!

lindsayad: huh? Are you looking for a general library passkey like we have in MOOSE? I can do that. It probably makes sense in tandem with me constructing the key simply as `{}` such that giving the pass key a meaningful name doesn't even matter

loganharbour: Yeah, I was just making a bit of a joke here but it's still a good idea.

And I agree - `libMesh::pass_key<System>` is more readable

include/base/libmesh_common.h
```diff
@@ -666,6 +666,12 @@ static constexpr std::size_t libmesh_dim = LIBMESH_DIM;
 #define libmesh_fallthrough() ((void) 0)
 #endif
 
+template <typename T>
+class PassKey
+{
+  friend T;
+  constexpr PassKey() = default;
```
loganharbour, line 673:
```suggestion
  constexpr PassKey() = default;
  constexpr PassKey(const PassKey &) {}
```

I think?

lindsayad: why? The copy constructor should be implicitly defined

include/base/dof_map.h
```diff
@@ -572,17 +572,17 @@ class DofMap : public DofMapBase,
...
+  void
+  add_variables_and_groups(const std::vector<Variable> & variables, const std::vector<VariableGroup> & variable_groups, PassKey<System>);
 
   /**
-   * Add a group of unknowns of order \p order and finite element type
-   * \p type to the system of equations.
+   * Add knowledge of array variables
    */
-  void add_variable_group (VariableGroup var_group);
+  void add_array_variables(const std::vector<ArrayVariableBlock> & array_variables)
```
roystgnr: This isn't an "add" (append to whatever already exists), it's a "set" (overwrite anything that already exists).  Let's rename and update the comment accordingly?

include/base/dof_map.h
```diff
@@ -1947,14 +1974,14 @@ class DofMap : public DofMapBase,
...
    */
-  std::vector<Variable> _variables;
+  const std::vector<Variable> * _variables = nullptr;
 
   /**
-   * The finite element type for each variable group.
+   * The groups of variables
    */
-  std::vector<VariableGroup> _variable_groups;
+  const std::vector<VariableGroup> * _variable_groups = nullptr;
```
roystgnr: Instead of storing these in the System and having pointers to that from the DofMap, couldn't we store these in the DofMap to which we already have a pointer from the System?

loganharbour:
@lindsayad @roystgnr have an ETA? The moose update is ready

include/base/dof_map.h
```diff
@@ -1687,6 +1707,13 @@ class DofMap : public DofMapBase,
 
 private:
 
+  /**
+   * Retrieve the array variable bounds for a given variable \p vi. This variable may
```
roystgnr, line 1812: Let's describe in more detail what these are?  If we hadn't discussed it already I'd have no idea.

include/base/dof_map.h
```diff
@@ -2702,6 +2738,71 @@ StaticCondensationDofMap & DofMap::get_static_condensation()
   return *_sc;
 }
 
+inline std::pair<unsigned int, unsigned int>
+DofMap::get_variable_array(const unsigned int vi) const
+{
+  auto it = std::upper_bound(
```
roystgnr, line 2864: Not sure if it's worth the trouble to make arrays to make this O(1), probably not until we profile it, but if we don't then we should mention in the doc string that it's O(log2(Nvar)) so people can know to keep it out of inner loops.

include/base/libmesh_common.h
roystgnr: I think if we'd manually defined the default constructor then that would disable default definitions of other constructors, but with just the `= default` we should still get the other autogenerated constructors.

Is the autogenerated default constructor not sufficient, though?  https://en.cppreference.com/w/cpp/language/default_constructor.html#Implicitly-defined_default_constructor claims the autogenerated default will still be `constexpr`.

include/base/dof_map.h
```diff
@@ -2173,6 +2200,9 @@ class DofMap : public DofMapBase,
 
   /// Static condensation class
   std::unique_ptr<StaticCondensationDofMap> _sc;
+
+  /// Pointer to array variable information held by the \p System
+  const std::vector<ArrayVariableBlock> * _array_variables = nullptr;
```
roystgnr: Likewise for this; it seems like less rigamarole to hold it in the DofMap.

roystgnr:
My vote would be to get the submodule updated ASAP; we can always rush the next one.  I'll try to keep an eye on work this weekend if we think we can get this squeezed in by Sunday though.

lindsayad:
> @lindsayad @roystgnr have an ETA? The moose update is ready

let's just go ahead with the moose update. As @roystgnr says we can do a quick follow-on once this is ready

include/base/libmesh_common.h
lindsayad: > I think if we'd manually defined the default constructor then that would disable default definitions of other constructors, but with just the `= default` we should still get the other autogenerated constructors.

Not true. You could declare arbitrarily many constructor types, and as long as you do not declare your own copy constructor, the compiler will do so for you. From https://en.cppreference.com/w/cpp/language/copy_constructor.html: 

> If no user-defined copy constructors are provided for a class type, the compiler will always declare a copy constructor as a non-[explicit](https://en.cppreference.com/w/cpp/language/explicit.html) inline public member of its class.

Regarding the default constructor

> Is the autogenerated default constructor not sufficient, though?

From https://en.cppreference.com/w/cpp/language/default_constructor.html:

> If there is no user-declared constructor or constructor template for a class type, the compiler will implicitly declare a default constructor as an inline ***public*** member of its class.

***public*** is the problematic part there for the pass-key idiom

include/base/dof_map.h
lindsayad: I will add doc. I think the common use case is that we have relatively few array variables but each array variable could have order of hundred component variables. So to me the logical thing to do is O(log(N)) computational complexity over a small N instead of O(N) memory use for large N

lindsayad: Either way the cost should be small and practically unimportant but the way I've done it feels morally correct haha

include/base/dof_map.h
```diff
@@ -573,16 +573,115 @@ class DofMap : public DofMapBase,
...
+  /**
+   * Adds variables \p vars to the list of variables
+   * for this system. If \p active_subdomains is either \p nullptr
+   * (the default) or points to an empty set, then it will be assumed that
+   * the \p vars have no subdomain restrictions. This API will end up
+   * calling \p this->add_variables(). However, we will additionally store data
+   * that can be leveraged by the \p DofMap to build degrees of freedom
+   * containers corresponding to all the variables in this variable array
+   *
+   * \returns The index number for the last of the new variables.
```
roystgnr, line 686: Let's add something here about what it means for a variable to be an array variable?

Also: should we have a shim for this in `System`?

lindsayad: I already added one; see https://github.com/libMesh/libmesh/pull/4250/files#diff-ed6efbeb218f859f2083e9e2d3ffdb6b9dcdba76c9f7aed5c8f4eca398a294efR1184

roystgnr: Sorry, not sure how I missed that!

jwpeterson (approved):

PR #4251, opened 2025-09-16
Update autotools

roystgnr:
@loganharbour discovered some linker failures on newer conda environments that appear to trace to an issue with libtool; upgrading autotools seems to fix them for me.

I'm still doing more extensive testing, and even if it all works I'm probably going to need to also fix up our bootstrapped MetaPhysicL and TIMPI at least, but first I want to make sure that newer autotools doesn't cause us any problems on *older* conda environments.


loganharbour:
Thanks for digging into this. I guess it was probably time anyway...

roystgnr:
With the submodule autotools update commits added on, this is fixing new-conda builds for me.

I *hate* being forced into an upgrade like this but I don't think I've got time to figure out the underlying problem.  At least with our last upgrade, the new M1 architecture was a good reason to be pushed into it.  This time I'm not sure what the reason is other than "libtool broke".

roystgnr:
John's okay with this too, and the python failures are old and unrelated, so we're going for it.

We will be making a 1.9 release branch with the old autotools, just in case...

PR #4254, opened 2025-09-17
Cleanup unit test output

roystgnr:
This was all John needed to fix the broken distcheck/distclean in our release branches; odds are it's all we need here too.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -363,6 +363,7 @@ CLEANFILES = cube_mesh.xda \
              read_header_test.e \
              output.dat \
              periodic.e \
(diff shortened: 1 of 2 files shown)


roystgnr:
I've added both recipes with failing distcheck steps; no point in merging this unless both pass.

jwpeterson:
The "Distributed make check sweep (odd)" is still showing up as "Waiting for status to be reported" for me, but when I clicked on it, it was all green as of 17 hours ago, so I guess the hangup there is either on the GitHub or CIVET sides.

jwpeterson (approved):

PR #4256, opened 2025-09-24
Variational smoother: prism support

pbehne:
This PR adds prism support to the variational smoother. Additionally, APIs are added the the `VariationalMeshSmoother` to set solver tolerances and verbosity.

Ref. #4082.

### DIFF
--- a/include/systems/variational_smoother_constraint.h
+++ b/include/systems/variational_smoother_constraint.h
@@ -57,7 +57,7 @@ public:
    * @param point The point defining the constraint.
    * @param tol The tolerance to use for numerical comparisons.
    */
-  PointConstraint(const Point & point, const Real & tol = TOLERANCE);
+  PointConstraint(const Point &point, const Real &tol = TOLERANCE * TOLERANCE);
 
   /**
    * Comparison operator for ordering PointConstraint objects.
@@ -131,7 +131,8 @@ public:
    * @param direction the direction of the constraining line.
    * @param tol The tolerance to use for numerical comparisons.
    */
-  LineConstraint(const Point & point, const Point & direction, const Real & tol = TOLERANCE);
+  LineConstraint(const Point &point, const Point &direction,
+                 const Real &tol = TOLERANCE * TOLERANCE);
 
   /**
    * Comparison operator for ordering LineConstraint objects.
@@ -230,7 +231,8 @@ public:
    * @param normal the direction normal to the constraining plane.
    * @param tol The tolerance to use for numerical comparisons.
    */
-  PlaneConstraint(const Point & point, const Point & normal, const Real & tol = TOLERANCE);
+  PlaneConstraint(const Point &point, const Point &normal,
+                  const Real &tol = TOLERANCE * TOLERANCE);
 
   /**
    * Comparison operator for ordering PlaneConstraint objects.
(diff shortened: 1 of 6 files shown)


include/systems/variational_smoother_constraint.h
```diff
@@ -57,7 +57,7 @@ class PointConstraint
    * @param point The point defining the constraint.
    * @param tol The tolerance to use for numerical comparisons.
    */
-  PointConstraint(const Point &point, const Real &tol = TOLERANCE * TOLERANCE);
+  PointConstraint(const Point & point, const Real & tol = TOLERANCE);
```
roystgnr: This is a pretty drastic loosening.  Six orders of magnitude with double precision!  What happened?

src/systems/variational_smoother_constraint.C
```diff
@@ -195,7 +195,8 @@ bool PlaneConstraint::operator<(const PlaneConstraint & other) const
 
 bool PlaneConstraint::operator==(const PlaneConstraint & other) const
 {
-  if (!(_normal.absolute_fuzzy_equals(other.normal(), _tol)))
+  if (_normal.cross(other.normal()).norm() > _tol)
```
roystgnr: This is more expensive but I'm having trouble seeing how it's better.

tests/mesh/mesh_smoother_test.C
```diff
@@ -871,7 +871,8 @@ public:
   void testVariationalPrism18()
   {
     ReplicatedMesh mesh(*TestCommWorld);
-    VariationalMeshSmoother variational(mesh);
+    VariationalMeshSmoother variational(
+        mesh, 0.5, true, TOLERANCE * TOLERANCE, 100 * TOLERANCE * TOLERANCE);
```
roystgnr: We're having trouble just hitting 1e-12 residual reduction?  How far do we get before we stall out?

include/systems/variational_smoother_constraint.h
pbehne: IIRC, using 1e-12 was a bit too tight for comparisons of basically equivalent points, lines, and planes. For example, when `Plane::operator==` returns false on two basically equal planes, the intersection of the planes is sought as the constraint. Since the planes are parallel, there are issues here.

I don't think I needed to loosen the tolerance by 6 orders of magnitude, but doing so made everything work and didn't break any existing tests.

I wrote this code several months ago, so I'll play with it a bit and see if anything has changed. Same for your other comments.

tests/mesh/mesh_smoother_test.C
pbehne: The prism2 test gets to 2e-12, prism21 gets to 8e-12. I went ahead and changed that factor 100 to 10. prism18 converged to 1e-12, so I took the factor out.

include/systems/variational_smoother_constraint.h
pbehne: This doesn't seem to be an issue anymore, so I reverted it.

src/systems/variational_smoother_constraint.C
pbehne: This doesn't seem to be an issue anymore, so I reverted it.

tests/mesh/mesh_smoother_test.C
pbehne: >prism18 converged to 1e-12, so I took the factor out.

Does not converge on civet, though, so I put the factor back in. 1e-12 is right on the edge, so makes sense.

PR #4258, opened 2025-09-24
Make some constants as constant expression

NamjaeChoi:
`numeric_limits` seems to be constant expression now. With the current code, the invalid ID cannot be accessed on GPU because it is considered as a variable on CPU.

### DIFF
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -241,14 +241,9 @@ public:
    * \note We don't use the static_cast(-1) trick here since
    * \p subdomain_id_type is sometimes a *signed* integer for
    * compatibility reasons (see libmesh/id_types.h).
-   *
-   * \note Normally you can declare static const integral types
-   * directly in the header file (C++ standard, 9.4.2/4) but
-   * std::numeric_limits<T>::max() is not considered a "constant
-   * expression".  This one is therefore defined in elem.C.
-   * http://stackoverflow.com/questions/2738435/using-numeric-limitsmax-in-constant-expressions
    */
-  static const subdomain_id_type invalid_subdomain_id;
+  static constexpr subdomain_id_type invalid_subdomain_id
+    = std::numeric_limits<subdomain_id_type>::max();
 
   /**
    * \returns true iff this element type can vary in topology (e.g.
--- a/include/base/dof_object.h
+++ b/include/base/dof_object.h
@@ -479,18 +479,18 @@ public:
   /**
    * An invalid \p id to distinguish an uninitialized \p DofObject
    */
-  static const dof_id_type invalid_id = static_cast<dof_id_type>(-1);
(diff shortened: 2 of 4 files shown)


NamjaeChoi:
@roystgnr Concerns?

include/geom/elem.h
```diff
@@ -241,14 +241,9 @@ class Elem : public ReferenceCountedObject<Elem>,
...
-   *
-   * \note Normally you can declare static const integral types
-   * directly in the header file (C++ standard, 9.4.2/4) but
-   * std::numeric_limits<T>::max() is not considered a "constant
-   * expression".  This one is therefore defined in elem.C.
-   * http://stackoverflow.com/questions/2738435/using-numeric-limitsmax-in-constant-expressions
    */
-  static const subdomain_id_type invalid_subdomain_id;
+  static constexpr subdomain_id_type invalid_subdomain_id
+    = std::numeric_limits<subdomain_id_type>::max();
```
jwpeterson, line 246: I thought the whole thing about static variables was that they need to be defined outside of the class definition, typically in the .C file, but I guess in-class initialization has been allowed since C++11... Just curious though, did it fail to compile with `constexpr` unless you also moved the initialization to the header?

jwpeterson:
> the invalid ID cannot be accessed on GPU because it is considered as a variable on CPU.

Can you elaborate on this? I guess there was a compilation error with the nvidia compiler or something? Because I would like to know if the compiler is actually right here or if it's being too strict, etc.

include/geom/elem.h
NamjaeChoi: Haven't tried to compile my change before push 😅

NamjaeChoi:
So in CUDA,

```
18.5.14. [Const-qualified variables](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#const-qualified-variables)
Let ‘V’ denote a namespace scope variable or a class static member variable that has const qualified type and does not have execution space annotations (for example, __device__, __constant__, __shared__). V is considered to be a host code variable.

The value of V may be directly used in device code, if
- V has been initialized with a constant expression before the point of use,
- the type of V is not volatile-qualified, and
- it has one of the following types:
    - built-in floating point type except when the Microsoft compiler is used as the host compiler,
    - built-in integral type.
```
Apparently initializing in the source seems to violate the condition `V has been initialized with a constant expression before the point of use`, because the nvcc compiler can't see it.

NamjaeChoi:
So when I tried to use `libMesh::Elem::invalid_subdomain_id`, I got
`error: identifier "libMesh::Elem::invalid_subdomain_id" is undefined in device code`

jwpeterson (approved):
Change looks OK to me. I don't love writing special code to cater to specific compilers, but in this case the changes are reasonable from a general C++ point of view as well IMO. I'll give @roystgnr a chance to comment as well before merging.

PR #4260, opened 2025-09-29
Tell FESubdivision to calculate phi, dphi, d2phi

romanongit:
Avoid deprecation warning in FEGenericBase::determine_calculations() by telling FESubdivision to calculate phi, dphi, and d2phi. See https://github.com/libMesh/libmesh/discussions/4259

### DIFF
--- a/src/fe/fe_subdivision_2D.C
+++ b/src/fe/fe_subdivision_2D.C
@@ -686,6 +686,9 @@ void FESubdivision::reinit(const Elem * elem,
   libmesh_assert_equal_to(sd_elem->get_ordered_valence(2), 6);
 
   // We're calculating now!  Time to determine what.
+  this->calculate_phi = true;
+  this->calculate_dphi = true;
+  this->calculate_d2phi = true;
   this->determine_calculations();
 
   // no custom quadrature support


roystgnr:
I think @jwpeterson and I misinterpreted "Is each specific FE supposed to set calculate_phi etc" in #4259 - it's not that the `FE` itself is supposed to set those values to true regardless of what application code prerequests, it's that the application prerequests by making the `FE::get_foo()` calls, and that's where the `FE` sets the appropriate `calculate_foo` values to true.

If we were to move the declarations of `phi`/`dphi`/`d2phi` in miscellaneous_ex11.C to after the `reinit()` call, that should cause the `reinit()` to fail (in `--disable-deprecated` builds) for lack of prerequests, but the fix wouldn't be this PR, it would be to move the declarations back, to make the postrequests into prerequests again.

On the other hand, maybe I'm misinterpreting something *here* - is there something in particular about the `FESubdomain` that would require it to compute all of `phi`, `dphi`, and `d2phi` regardless of what the user requests?  (e.g. an `FE` transformation might request the `FEMap` dxi/dx even if it itself was only requested to calculate `d2phi`, because the map's first derivatives appear in the equations for the mapped shape functions' second derivatives)

rvett:
I see. I think it's my non-standard way of using `FESubdivision` that's the problem. For performance reasons in a transient simulation, I first cache the quadrature weights (which does not need `phi`/`dphi`/`d2phi`) and then the `phi`/`dphi`/`d2phi`. I indeed did not call `get_foo()` in between the `reinit()`s. With those added, the deprecation warning is gone. Thanks for clarifying, the PR can be closed.

PR #4261, opened 2025-10-01
Fix an off-by-one error in HDF5 matrix reads

roystgnr:
I accidentally missed an edit when patterning this code for 0-based matrix files off the 1-based matlab matrix reader.


roystgnr:
This fixes parallel sweeps with a regression test downstream; I'm still trying to figure out how to concoct test coverage for this sort of thing that isn't inadequate (like our test with `tests/matrices/geom_1_extraction_op.h5`) or just redundant with that.

roystgnr:
Merging a bit early; this is passing sweeps and we want it downstream ASAP; without it those reads are fine in serial but can mis-preallocate in parallel depending on matrix sparsity pattern.

PR #4262, opened 2025-10-02
Fixes for devel->master test failures

roystgnr:
I'm having trouble replicating every failure we're seeing in parallel sweeps post-#4248, but I'm hoping most of them are downstream of losing parallel sync after the first failure.  I'll add the failing recipes here and we'll see what happens.


jwpeterson (approved):

roystgnr:
We've passed with "No AMR" as well as with the sweeps on 9 and 13 processors, so we're at least doing better than `devel` is without these fixes.  The full dbg-mode sweeps are going to take forever, so I'm going to merge early and let the `devel`->`master` recipes get on it.

PR #4263, opened 2025-10-02
[WIP] Allow specifying element range in project_vector

hugary1995:
Using std::optional to avoid breaking API.
This is needed in moose element subdomain modifier where we need to project a function onto a subset of the mesh.

### DIFF
--- a/include/numerics/wrapped_functor.h
+++ b/include/numerics/wrapped_functor.h
@@ -123,14 +123,28 @@ protected:
 template <typename Output>
 void WrappedFunctor<Output>::init_context (const FEMContext & c)
 {
-  for (auto dim : c.elem_dimensions())
+  // If the context was constructed with an active var subset (e.g. projector
+  // projecting only selected variables), then only those FE objects are
+  // guaranteed to exist.
+  const std::vector<unsigned int> * active = c.active_vars();
+
+  auto check_and_init = [&c](unsigned int v, unsigned short dim)
     {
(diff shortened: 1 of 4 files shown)


hugary1995:
I will add some tests later. @roystgnr please let me know if this looks okay to you before I go too far.

roystgnr:
Sorry I missed the comment here!

Definitely we want to add this feature.

I'd have made the range argument a pointer with a `nullptr` default, but out of force of habit rather than because it's a good idea, and I'm happy with `std::optional` here.  Someone correct me if I'm wrong, but best practices now is to use `std::optional` for any object that's lightweight enough to not worry about copying, right?  (and probably also for heavyweight objects, eventually, when C++26 adds `optional<T&>` support?)  Our ranges are currently like 48 bytes, so the copies should be fine.

ChengHauYang:
Dear @roystgnr, 

I added one test case and modified it to support `project_solution` with elem_ranges as well.
Could you please review it and assist us in activating the CI?

Thanks!

ChengHauYang:
Dear @roystgnr,

We tried to support libMesh `project_vector` and `project_solution` with target variables as well. 
And, I modified the test to test this as well.
If you are available and have a chance, could you please review it and help us mark this PR as "ready for review"?

Thanks for your time!

ChengHauYang:
It appears that https://civet.inl.gov/repo/8/ does not recognize this PR, so as @roystgnr suggested, we plan to close this PR and open a new PR (#4368).

PR #4264, opened 2025-10-03
Fix for poly2tri headers not being installed with correct directory structure

jwpeterson:
The simplicity of the fix belies the difficulty I had in finding it... the automake documentation is not good, so I was relying on dozens of desperate Google searches before finally stumbling on the right fix.

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -169,7 +169,7 @@ autoreconf=$(which autoreconf 2>/dev/null)
 if test "$autoreconf" != ""; then
     if test -x $autoreconf; then
         $autoreconf -iv --force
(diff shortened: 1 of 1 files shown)


jwpeterson:
I'm not sure if the Test 32bit target is hung, I'll give it a few more minutes though. I see this in the downloaded output:
```
*****************************************************

CIVET: Output size exceeded limit (5242880 bytes), pausing live output!

*****************************************************
```
but other than that it seems to be running fine.

jwpeterson:
This should fix an issue that we ran into with our internal code while trying to include poly2tri headers, so I'm going to go ahead and merge now.

PR #4267, opened 2025-10-06
Variational smoother: mix element tests

pbehne:
This PR adds regression tests for 2D and 3D meshes composed of mixed element types.

2D test of quad4 and tri3 elements. Left is distorted, right is smoothed.
image

3D test of hex8 and prism6 elements. Left is distorted, right is smoothed.
image

Ref. #4082.


roystgnr:
I looked again at the code to make sure you were disabling renumbering, so a DistributedMesh wouldn't be confused when you check against a gold mesh by node number ... and it's not an issue, because we're hard-coding `ReplicatedMesh` on even our older tests.  Is that just because that's what we did for the `LaplaceMeshSmoother` tests?  That's necessary there, because `LaplaceMeshSmoother` sucks, but you're writing code that doesn't suck and we ought to be able to test accordingly!

PR #4275, opened 2025-10-15
Drop deprecated DofObject::set_unique_id() API

jwpeterson:
This has been deprecated since approximately February 2025. Since we started the 1.9.x release, we'll have one more release series with the old API present, and then it will be gone in 1.10.x.


roystgnr:
For future reference, I think this was another API that was `libmesh_deprecated()` at runtime in 2020 but that had compile-time deprecation added in Februrary.

PR #4277, opened 2025-10-16
Drop deprecated Elem::centroid() API

jwpeterson:
This has also been deprecated for a few releases now so it's time to remove it. Let's use the CI testing to see if any apps are still using the deprecated function.


jwpeterson:
The new API has been present since release 1.7.0 back in 2021 (6e80c5a2627488ce33d5c3d55b87e7ffd0562c1c) so it's definitely reasonable to remove this now.

PR #4278, opened 2025-10-16
Drop deprecated Elem::which_node_am_i() API

jwpeterson:
This has been officially deprecated since Feb 2025 (4154046815) when a bunch of APIs were deprecated. If nothing important is still using it, it's a good time to get rid of it.

### DIFF
--- a/include/geom/elem.h
+++ b/include/geom/elem.h
@@ -462,14 +462,6 @@ public:
   virtual unsigned int local_edge_node(unsigned int edge,
                                        unsigned int edge_node) const = 0;
 
-#ifdef LIBMESH_ENABLE_DEPRECATED
-  /**
-   * This function is deprecated, call local_side_node(side, side_node) instead.
-   */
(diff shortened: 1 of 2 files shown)


roystgnr (approved):
I was a little wary (Feb 2025? but I guess that predates the 1.9 branch...) but Feb was just when we added the compile-time ifdefs - there's been a `libmesh_deprecated()` call in there since 2020.  This should be long gone!

jwpeterson:
> but Feb was just when we added the compile-time ifdefs

Oops, yeah looks like the official deprecation commit for this one was 0acea1ef7932d7298dc042ea90793aeab4197fe3. I was `git blaming` the `LIBMESH_ENABLE_DEPRECATED` lines, but I should be looking at the `libmesh_deprecated()` lines.

jwpeterson:
The tests have been done for a while now, I think there's some issue with the coverage now.

PR #4279, opened 2025-10-16
Added verbosity system to variational smoother

pbehne:
Added verbosity system to variational smoother
Ref. #4082

### DIFF
--- a/include/mesh/mesh_smoother_vsmoother.h
+++ b/include/mesh/mesh_smoother_vsmoother.h
@@ -84,7 +84,19 @@ public:
    * modify sobdomain boundaries.
    * @param relative_residual_tolerance Solver setting for the relative residual tolerance.
    * @param absolute_residual_tolerance Solver setting for the absolute residual tolerance.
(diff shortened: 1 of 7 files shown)


include/mesh/mesh_smoother_vsmoother.h
```diff
@@ -84,16 +84,14 @@ class VariationalMeshSmoother : public MeshSmoother
    * modify sobdomain boundaries.
    * @param relative_residual_tolerance Solver setting for the relative residual tolerance.
    * @param absolute_residual_tolerance Solver setting for the absolute residual tolerance.
-   * @param solver_quiet Whether to make the solver quiet.
-   * @param solver_verbose Whether to make the solver verbose.
+   * @param verbosity Smoother verbosity option specifying how much information is output.
```
roystgnr: I was picturing a scale from 0 to 10, to make granularity available later, or maybe even 1 to 100 to have a natural "percent of maximum" interpretation, but it sounds like we've currently got a scale from 0 to 2?  However we do it we need to document what this means.  How do I know if I want 2 information-output or 20?

PR #4280, opened 2025-10-16
Variational Smoother: Target element bug fixes

pbehne:
This commit patches two bugs related to target elements:
1. To switch integrals from the reference element space to the target element space, we need to adjust weights by DIVIDING by the determinants of the target-to-reference jacobians. This is the same as multiplying by the determinants of the reference-to-target jacobians. Since we store the former and not the later, the erroneous multiplication operations have been changed to division.
2. When switching S from the reference to the target element (S *= target_jacobians), we had forgot to update dS/dR by target_jacobians accordingly. This error has been fixed.

Ref. #4082


roystgnr:
It's good to see that the test cases where we had exact answers are still passing, but a couple where we just had gold solutions are failing.  We can probably just regold those.

pbehne:
Re-golded solutions look good.

2D test (distorted on left, smoothed on right):
image

3D test (distorted on left, smoothed on right):
image

PR #4283, opened 2025-10-21
Enable disjoint neighbor detection across disconnected boundary pairs in libMesh

ChengHauYang:
## Motivation

Currently, libMesh assumes physical continuity (base on node ID) between neighboring elements, which limits its ability to handle cases where two regions share coincident boundaries but are geometrically disconnected -- for instance, across interfaces with discontinuous fields or temperature jumps.
This PR introduces a mechanism to explicitly register disconnected boundary pairs, enabling the framework to reconstruct logical neighbor relationships even when elements are not directly connected in the mesh topology (base on node ID).
This functionality is particularly relevant for methods such as the Cohesive Zone Method (CZM), where field continuity is enforced weakly across interfaces.

This work is motivated by long-standing MOOSE discussions about *“fake”* or *“pseudo”* neighbors -- element pairs that were once topological neighbors but become disconnected due to node replacement (e.g., in `BreakMeshByBlockGenerator`).  
See the following issues for historical context:
- [MOOSE #12033](https://github.com/idaholab/moose/issues/12033): *Support for non-geometric neighbor connections* -- initial proposal to allow interface coupling between geometrically separated regions.
- [MOOSE #21161](https://github.com/idaholab/moose/issues/21161): *BreakMeshByBlock connectivity* -- highlighted that after node replacement, elements that were formerly neighbors lose topological connectivity, motivating the need for a libMesh-level mechanism to handle such cases.

This PR effectively formalizes that capability at the libMesh level, turning what was previously a MOOSE-side workaround into a supported feature.

## Design

1. Disconnected Boundary Framework: Introduced an API in `MeshBase` (`add_disconnected_boundaries()` and `get_disconnected_boundaries()`) that allows users to register and access disconnected boundary pairs. These pairs are stored as `PeriodicBoundaries` objects and used during neighbor search to establish logical connections between geometrically separated regions.
2. Neighbor Reconstruction and Validation: Extended `UnstructuredMesh::find_neighbors()` to detect and link virtual (fake, or disconnected) neighbors across disconnected interfaces (boundaries). A new regression test (`disconnected_neighbor_test.C`) verifies correct neighbor mapping and demonstrates a temperature-jump interface problem that matches the analytical solution.

## Impact
Facilitates coupling between subdomains separated by non-matching meshes or duplicated boundaries.

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -1850,8 +1850,8 @@ public:
 protected:
 
 #ifdef LIBMESH_ENABLE_PERIODIC
-  /// @brief The disjoint boundary id pairs.
-  std::unique_ptr<PeriodicBoundaries> _disjoint_boundary_pairs;
+  /// @brief The disjoint neighbor boundary id pairs.
+  std::unique_ptr<PeriodicBoundaries> _disjoint_neighbor_boundary_pairs;
 #endif
 
   /**
--- a/src/mesh/mesh_base.C
+++ b/src/mesh/mesh_base.C
@@ -159,13 +159,13 @@ MeshBase::MeshBase (const MeshBase & other_mesh) :
 
 #ifdef LIBMESH_ENABLE_PERIODIC
   // Deep copy of all periodic boundaries
-  if (other_mesh._disjoint_boundary_pairs)
+  if (other_mesh._disjoint_neighbor_boundary_pairs)
     {
-      _disjoint_boundary_pairs = std::make_unique<PeriodicBoundaries>();
+      _disjoint_neighbor_boundary_pairs = std::make_unique<PeriodicBoundaries>();
 
-      for (const auto & [id, pb] : *other_mesh._disjoint_boundary_pairs)
+      for (const auto & [id, pb] : *other_mesh._disjoint_neighbor_boundary_pairs)
         if (pb)
-          (*_disjoint_boundary_pairs)[id] = pb->clone();
+          (*_disjoint_neighbor_boundary_pairs)[id] = pb->clone();
     }
 #endif
 
@@ -218,13 +218,13 @@ MeshBase& MeshBase::operator= (MeshBase && other_mesh)
   // Deep copy of all periodic boundaries:
   // We must clone each PeriodicBoundaryBase in the source map,
   // since unique_ptr cannot be copied and we need independent instances
-  if (other_mesh._disjoint_boundary_pairs)
+  if (other_mesh._disjoint_neighbor_boundary_pairs)
     {
-      _disjoint_boundary_pairs = std::make_unique<PeriodicBoundaries>();
+      _disjoint_neighbor_boundary_pairs = std::make_unique<PeriodicBoundaries>();
 
-      for (const auto & [id, pb] : *other_mesh._disjoint_boundary_pairs)
+      for (const auto & [id, pb] : *other_mesh._disjoint_neighbor_boundary_pairs)
         if (pb)
-          (*_disjoint_boundary_pairs)[id] = pb->clone();
+          (*_disjoint_neighbor_boundary_pairs)[id] = pb->clone();
     }
 #endif
 
@@ -336,11 +336,11 @@ bool MeshBase::locally_equals (const MeshBase & other_mesh) const
     return false;
 
   // First check whether the "existence" of the two pointers differs (one present, one absent)
-  if ((bool)_disjoint_boundary_pairs != (bool)other_mesh._disjoint_boundary_pairs)
+  if ((bool)_disjoint_neighbor_boundary_pairs != (bool)other_mesh._disjoint_neighbor_boundary_pairs)
     return false;
   // If both exist, compare the contents (Weak Test: just compare sizes like `_ghosting_functors`)
-  if (_disjoint_boundary_pairs &&
-      (_disjoint_boundary_pairs->size() != other_mesh._disjoint_boundary_pairs->size()))
+  if (_disjoint_neighbor_boundary_pairs &&
+      (_disjoint_neighbor_boundary_pairs->size() != other_mesh._disjoint_neighbor_boundary_pairs->size()))
     return false;
 
   const constraint_rows_type & other_rows =
@@ -1999,10 +1999,10 @@ void MeshBase::detect_interior_parents()
                                              const RealVectorValue & translation)
     {
       // Lazily allocate the container the first time it’s needed
-      if (!_disjoint_boundary_pairs)
-        _disjoint_boundary_pairs = std::make_unique<PeriodicBoundaries>();
+      if (!_disjoint_neighbor_boundary_pairs)
+        _disjoint_neighbor_boundary_pairs = std::make_unique<PeriodicBoundaries>();
 
-      PeriodicBoundaries & db = *_disjoint_boundary_pairs;
+      PeriodicBoundaries & db = *_disjoint_neighbor_boundary_pairs;
 
       // Create forward and inverse boundary mappings
       PeriodicBoundary forward(translation);
@@ -2020,22 +2020,22 @@ void MeshBase::detect_interior_parents()
 
   PeriodicBoundaries * MeshBase::get_disjoint_neighbor_boundary_pairs()
     {
-      return _disjoint_boundary_pairs.get();
+      return _disjoint_neighbor_boundary_pairs.get();
     }
 
   const PeriodicBoundaries * MeshBase::get_disjoint_neighbor_boundary_pairs() const
     {
-      return _disjoint_boundary_pairs.get();
+      return _disjoint_neighbor_boundary_pairs.get();
     }
 
     void MeshBase::remove_disjoint_boundary_pair(const boundary_id_type b1,
                                                    const boundary_id_type b2)
     {
       // Nothing to remove if not allocated or empty
-      if (!_disjoint_boundary_pairs || _disjoint_boundary_pairs->empty())
+      if (!_disjoint_neighbor_boundary_pairs || _disjoint_neighbor_boundary_pairs->empty())
         return;
 
-      auto & pairs = *_disjoint_boundary_pairs;
+      auto & pairs = *_disjoint_neighbor_boundary_pairs;
 
       // Helper to check and erase both directions
       auto erase_if_match = [](boundary_id_type key,
--- a/src/mesh/mesh_refinement.C
+++ b/src/mesh/mesh_refinement.C
@@ -686,7 +686,7 @@ bool MeshRefinement::refine_elements ()
 
   // Prevent refinement when the mesh has disjoint neighbor boundary pairs.
   //
-  // Disjoint boundary interfaces violate the topological continuity
+  // Disjoint neighbor boundary interfaces violate the topological continuity
   // assumed by the refinement algorithm. Refinement on such meshes can
   // produce invalid neighbor relationships (for example reverse indices
   // equal to invalid_uint), which may trigger assertions like
@@ -697,7 +697,7 @@ bool MeshRefinement::refine_elements ()
   if (_mesh.get_disjoint_neighbor_boundary_pairs())
     {
       libmesh_not_implemented_msg(
-        "Mesh refinement is not yet implemented for meshes with disjoint boundary interfaces.\n");
+        "Mesh refinement is not yet implemented for meshes with disjoint neighbor boundary interfaces.\n");
     }
 
   if (_face_level_mismatch_limit)
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -1879,14 +1879,14 @@ UnstructuredMesh::stitching_helper (const MeshBase * other_mesh,
       // Case 1: On "this" mesh, a or b exist but are not paired,
       //         while the other mesh pairs them.
       if (!in_this && (pb_this_a || pb_this_b) && in_other)
-        libmesh_error_msg("Disjoint boundary pairing mismatch: on 'this' mesh, "
+        libmesh_error_msg("Disjoint neighbor boundary pairing mismatch: on 'this' mesh, "
                           "boundary (" << a << " or " << b
                           << ") exists but is not paired; on 'other' mesh the pair is present.");
 
       // Case 2: On "other" mesh, a or b exist but are not paired,
       //         while this mesh pairs them.
       if (!in_other && (pb_other_a || pb_other_b) && in_this)
-        libmesh_error_msg("Disjoint boundary pairing mismatch: on 'other' mesh, "
+        libmesh_error_msg("Disjoint neighbor boundary pairing mismatch: on 'other' mesh, "
                           "boundary (" << a << " or " << b
                           << ") exists but is not paired; on 'this' mesh the pair is present.");
 
--- a/tests/systems/disjoint_neighbor_test.C
+++ b/tests/systems/disjoint_neighbor_test.C
@@ -624,7 +624,7 @@ private:
     }
     catch (const std::exception &e)
     {
-      CPPUNIT_ASSERT(std::string(e.what()).find("Disjoint boundary pairing mismatch") != std::string::npos);
+      CPPUNIT_ASSERT(std::string(e.what()).find("Disjoint neighbor boundary pairing mismatch") != std::string::npos);
     }
   }
 


include/base/periodic_boundaries.h
```diff
@@ -68,7 +68,8 @@ class PeriodicBoundaries : public std::map<boundary_id_type, std::unique_ptr<Per
                         const PointLocatorBase & point_locator,
                         const Elem * e,
                         unsigned int side,
-                        unsigned int * neigh_side = nullptr) const;
+                        unsigned int * neigh_side = nullptr,
+                        bool skip_found_check = false) const;
```
roystgnr: We need documentation here for the new parameter

include/mesh/mesh_base.h
```diff
@@ -36,6 +36,10 @@
 #include <string>
 #include <memory>
 
+// periodic boundary condition support
+#include "libmesh/periodic_boundaries.h"
+#include "libmesh/periodic_boundary.h"
```
roystgnr: Let's make the new functions non-inline and move them to mesh_base.C so we can get away with forward declarations here.

include/mesh/mesh_base.h
```diff
@@ -1820,8 +1824,50 @@ class MeshBase : public ParallelObject
...
+  void add_disconnected_boundaries(const boundary_id_type b1,
+                                  const boundary_id_type b2)
+    {
+      // Lazily allocate the container the first time it’s needed
+      if (!_disconnected_boundary_pairs)
+        _disconnected_boundary_pairs = std::make_unique<PeriodicBoundaries>();
+
+      // Create forward and inverse boundary mappings
+      PeriodicBoundary forward(RealVectorValue(0., 0., 0.));
+      PeriodicBoundary inverse(RealVectorValue(0., 0., 0.));
```
roystgnr: Is it safe to assume that we're always going to have coincident boundaries here, no gaps?  I guess if we do add gap use cases later it'll be easy enough to extend the API then.

src/base/periodic_boundaries.C
```diff
@@ -80,6 +81,10 @@ const Elem * PeriodicBoundaries::neighbor(boundary_id_type boundary_id,
   const MeshBase & mesh = point_locator.get_mesh();
   for(const Elem * elem_it : candidate_elements)
     {
+
+      if (elem_it == e) // skip self
+        continue;
```
roystgnr: I'd expect this to break the "single element across the domain because we haven't refined yet" case, and also to break angular periodicity at a "tip"/r=0 element at any refinement level.  I think we've got a test case for the latter somewhere (a regression test in MOOSE?); I'm going to be disappointed in our feature coverage if all our recipes pass.

tests/systems/disconnected_neighbor_test.C
```diff
@@ -0,0 +1,304 @@
...
+    sys.attach_assemble_function(assemble_temperature_jump);
+
+    // Ensure the DofMap creates sparsity entries between neighboring elements.
+    // Without this, PETSc may report "New nonzero at (a,b) caused a malloc."
+    sys.get_dof_map().set_implicit_neighbor_dofs(true);
+
+    es.init();
+    sys.solve();
+
+    ExodusII_IO(mesh).write_equation_systems("temperature_jump.e", es);
```
roystgnr: If we're not testing the output then let's not bother with it; you could leave this commented out if it's useful for debugging but as-is all it's going to do is break our `--disable-exodus` test builds.

src/mesh/unstructured_mesh.C
```diff
@@ -993,6 +993,41 @@ void UnstructuredMesh::find_neighbors (const bool reset_remote_elements,
...
+            {
+              // Skip if this side already has a valid neighbor (including remote neighbors)
+              if (element->neighbor_ptr(ms) != nullptr &&
+                  element->neighbor_ptr(ms) != remote_elem)
+                continue;
+
+                for (const auto & [id, boundary_ptr] : *db)
+                {
+                  unsigned int neigh_side;
+                  const Elem * neigh = db->neighbor(id, *point_locator, element, ms, &neigh_side, true /*skip_found_check*/);
```
roystgnr: What's the use case for `skip_found_check` here?  In general we prefer error messages over wrong answers, and it looks like here we expect to see a neighbor, and with `skip_found_check == false` any violation of that expectation would give us an error message, and with `skip_found_check == true` (plus any kernel depending on that neighbor pointer) we'd expect a wrong answer.

I'd figure the only hard case here would be the distributed-mesh "neighbor is remote" case, but that's supposed to be handled by the case where it's already `remote_elem` because we got those links right already (either a priori or in `delete_remote_elements`).

src/base/periodic_boundaries.C
roystgnr: And ... I got delayed in hitting the review button, and it does indeed look like we're failing MOOSE wedge and orthogonal and trapezoid tests.  Yay MOOSE!  Now I'm only disappointed in the *libMesh* feature coverage.

roystgnr (changes requested):
I really like this, in principle, and the MOOSE test failures should be trivial to fix.  Hopefully the distributed unit test sweep failures won't be much harder.

roystgnr:
Do we have discussion of the problem motivating this anywhere on Github that we can link to, for future reference?  I know there's been multiple discussions of the "MOOSE can abuse neighbor pointers" issue somewhere, but I can't recall whether any were here rather than on Slack or over video.

I'm not sure how *relevant* the discussions would be, because IIRC in most of them my opinion had been "we should come up with a different mechanism for MOOSE to use instead of `neighbor_ptr`" and I think you've managed to convert me to "we should make it so that misuse of `neighbor_ptr` is actually a supported use", but it'd still be good to be able to look up the history.

roystgnr:
Ooh.  One thing I missed in the review: this has a good chance of failing in interesting ways as soon as we use it in the presence of refined elements on either side of the pseudo-neighbor links, doesn't it?  And we don't have any coverage of that, do we?  We need another unit test or two; it'd probably be adequate to just refactor your existing test but refine both elements in the second test and refine just one element in the third.

include/base/periodic_boundaries.h
ChengHauYang: Thank you, @roystgnr ! Based on your comments, I’ll remove this parameter.

Previously, I added this parameter because my earlier implementation of finding disconnected neighbors involved looping over all element sides and then over the disconnected boundary pairs. In that version, I called
`neighbor(id, *point_locator, element, ms, &neigh_side, true /*skip_found_check*/);`
to test whether the neighbor was valid (i.e., not `remote_elem` and not `nullptr`), so I needed to skip the error check in that context.

Now, I plan to first call `this->get_boundary_info().has_boundary_id(element, ms, id)` to verify that the element actually has the disconnected boundary on this side (`ms` = my side). With this check in place, I can safely revert the `neighbor` function call back to the version that includes the standard libMesh error check.

src/mesh/unstructured_mesh.C
ChengHauYang: Thanks for your feedback. I think I should revert the function. The reason why I have this is because of what I mentioned earlier: https://github.com/libMesh/libmesh/pull/4283#discussion_r2449633753.

tests/systems/disconnected_neighbor_test.C
ChengHauYang: Sure. I will comment this out! Thanks for the feedback, @roystgnr !

src/base/periodic_boundaries.C
ChengHauYang: Thanks for your careful review and insightful feedback.
Previously, I was worried that the function might incorrectly return the original element itself, even when that wasn’t intended. However, I just realized that
`std::vector<unsigned int> neigh_sides = mesh.get_boundary_info().sides_with_boundary_id(elem_it, b->pairedboundary);`
is already sufficient to prevent this situation.

ChengHauYang:
Thank you, @roystgnr ! I tried to add  [MOOSE #12033](https://github.com/idaholab/moose/issues/12033) and [MOOSE #21161](https://github.com/idaholab/moose/issues/21161) to the Motivation section of the PR so that others can easily look up the historical context.

include/mesh/mesh_base.h
ChengHauYang: Sure. We will think about this more. My newest commit has addressed all your comments except for this one.

ChengHauYang: Right now, we can customize the gap in the function. Thanks!

src/base/periodic_boundaries.C
```diff
@@ -105,14 +106,13 @@ const Elem * PeriodicBoundaries::neighbor(boundary_id_type boundary_id,
   // either we're on a ghosted element with a remote periodic neighbor
   // or we're on a mesh with an inconsistent periodic boundary.
   libmesh_error_msg_if(mesh.is_serial() ||
-                       (e->processor_id() == mesh.processor_id()),
-                       "Periodic boundary neighbor not found");
+                      (e->processor_id() == mesh.processor_id()),
+                      "Periodic boundary neighbor not found");
```
roystgnr: Are there any substantive changes left in periodic_boundaries.C?  You can revert the whitespace regressions.

include/mesh/mesh_base.h
roystgnr: There are more complicated PBC types than a constant-translation gap, but we can always add a different overload if we ever want to use one of those; the underlying container is flexible enough.

roystgnr:
Gah, I forgot to re-activate CI for you.  Sorry!  I'm going to have you sent an Associates team invite so any future pushes can auto-activate CI.

roystgnr:
Jobs manually activated, hopefully the last time that's necessary; let me know if the invite doesn't get to you or gives you any problems.

roystgnr:
I'm also adding a few optional recipes that I'm not 100% sure will pass; I don't want to merge if this passes here only to find we have to fix something up for our devel->master merges.

src/mesh/unstructured_mesh.C
```diff
@@ -993,6 +997,42 @@ void UnstructuredMesh::find_neighbors (const bool reset_remote_elements,
...
+      for (const auto & element : this->element_ptr_range())
+        {
+          for (auto ms : element->side_index_range())
+            {
+              // Skip if this side already has a valid neighbor (including remote neighbors)
+              if (element->neighbor_ptr(ms) != nullptr &&
+                  element->neighbor_ptr(ms) != remote_elem)
+                continue;
+
+                for (const auto & [id, boundary_ptr] : *db)
```
roystgnr: Oh, wow - speaking of whitespace issues.  The first build failure I looked at (and probably most of the rest) was from a `-Werror=misleading-indentation` warning:

```
this 'if' clause does not guard... [-Werror=misleading-indentation]
               if (element->neighbor_ptr(ms) != nullptr &&
               ^~
../src/mesh/unstructured_mesh.C:1018:17: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
                 for (const auto & [id, boundary_ptr] : *db)
```

src/base/periodic_boundaries.C
ChengHauYang: Sure! I will revert it! Thanks, @roystgnr!

BTW, I have received the Associates team invite. Thanks!

src/mesh/unstructured_mesh.C
ChengHauYang: I tried to fix the indent here in the commit [64e4597](https://github.com/libMesh/libmesh/pull/4283/commits/64e4597796194763d45fe5a9b55fcac25cc3f534). Thanks!

roystgnr:
Well, the good news is that the invite went through and CI is autorunning now.  The bad news is that we're failing multiple recipes.  "Test mac" is a red herring, but the distributed-mesh test looks like a real failure and I can't even imagine *what* is going on in the --disable-amr and --disable-deprecated tests.

ChengHauYang:
Thank you, Roy, for your feedback. I think this issue is related to whether disconnected elements are being added as ghosted elements. I recently implemented a RelationshipManager in MOOSE, which allows me to properly handle disconnected neighbors on distributed meshes. However, this functionality is not yet available on the libMesh side. I plan to add a GhostingFunctor for disconnected neighbors in libMesh to see if that resolves the issue. Thanks again for taking the time to look into this.

ChengHauYang:
Dear @roystgnr ,

Thank you for your guidance. I’ve fixed the bugs related to distributed meshes, as well as those occurring when using `--disabled-amr` or `--disabled-deprecated`.

Here’s a brief summary of my recent changes:

1. Implemented a new GhostFunctor that utilizes `point_locator` and disconnected boundary pairs to add disconnected elements as ghost elements. This resolves the errors associated with distributed meshes.
2. Fixed a bug where internal boundaries could produce invalid side IDs when calling `side_with_boundary_id` under `--disabled-amr`. A corresponding unit test was added for verification.
3. Added three unit tests for disconnected neighbor handling and resolved test failures triggered by `--disabled-deprecated`. The previous failures occurred because `fe->reinit(...)` was called before `fe->get_phi()`.

All tests are now passing. Could you please provide your feedback and review when you have a chance?

Thanks again!

build-aux/ltmain.sh
```diff
@@ -7648,6 +7648,11 @@ func_mode_link ()
 	arg=$func_stripname_result
 	;;
 
+      -Wl,--as-needed)
+	deplibs="$deplibs $arg"
+	continue
+	;;
+
```
roystgnr: Any idea where the libtool changes here are coming from?

roystgnr:
I think that set of `OverlappingCouplingGhostingTest` failures in the Test No AMR recipe is a bug of mine; rebasing on `devel` again should fix it.  But this:

```
3) test: DisconnectedNeighborTest::testTempJumpLocalRefineFail (F) 
expected exception not thrown
```
is from this PR - you'll want to disable that test when `LIBMESH_ENABLE_AMR` isn't defined.

Review to follow.

src/ghosting/disconnected_neighbor_coupling.C
```diff
@@ -0,0 +1,52 @@
...
+            db->neighbor(id, *point_locator, elem, s, &neigh_side);
+
+        if (!neigh || neigh == remote_elem)
+          continue;
+
+        if (neigh->processor_id() != p)
+          coupled_elements.emplace(neigh, _dof_coupling);
+      }
+    }
+}
```
roystgnr: The new functor here looks reasonable, for what it does ... but wouldn't it be more reasonable to users to put what it does into `GhostPointNeighbors`?  We're already hiding the more-complicated PBC checks in an `if (check_periodic_bcs)` block there, and adding one more shouldn't hurt much.

src/mesh/boundary_info.C
```diff
@@ -2325,19 +2325,22 @@ unsigned int BoundaryInfo::side_with_boundary_id(const Elem * const elem,
...
           while (p != nullptr)
           {
             const Elem * parent = p->parent();
             if (parent && !parent->is_child_on_side(parent->which_child_am_i(p), side))
               break;
             p = parent;
           }
+#else
+          // do not forget to return the internal boundary when AMR is disabled
+          return side;
```
roystgnr, line 2343: Wow, good catch!  I take it your unit tests cover the previous failure?

src/mesh/mesh_refinement.C
```diff
@@ -684,6 +684,25 @@ bool MeshRefinement::refine_elements ()
...
+  // assumed by the refinement algorithm. Refinement on such meshes can
+  // produce invalid neighbor relationships (for example reverse indices
+  // equal to invalid_uint), which may trigger assertions like
+  // `rev < neigh->n_neighbors()` or corrupt neighbor data.
+  //
+  // For safety, refinement is disabled while disconnected boundaries are
+  // present.
+  if (_mesh.get_disconnected_boundaries())
+    {
+      libmesh_error_msg(
```
roystgnr: Let's make this a `libmesh_not_implemented_msg`?  It's not an error for someone to attempt this, it's just not something we support yet.

src/mesh/unstructured_mesh.C
```diff
@@ -993,6 +999,48 @@ void UnstructuredMesh::find_neighbors (const bool reset_remote_elements,
...
+                      auto neigh_changeable = this->elem_ptr(neigh->id());
+                      element->set_neighbor(ms, neigh_changeable);
+                      neigh_changeable->set_neighbor(neigh_side, element);
+                    }
+                }
+            }
+        }
+
+      // Ghost the disconnected elements
+      this->add_ghosting_functor(std::make_unique<DisconnectedNeighborCoupling>(*this));
```
roystgnr: No, this definitely isn't the way to do it.  It's better than requiring the user to handle adding a special ghosting functor, but if the user is modifying their mesh and relying on `find_neighbors` afterward, we don't want to be adding a new copy of a ghosting functor with each call.

src/mesh/boundary_info.C
roystgnr: Looks like - thanks again!

ChengHauYang:
Thank you, @roystgnr, for your carefully checking. I just fixed that test and rebased the code again.

src/mesh/unstructured_mesh.C
roystgnr: If we can just add the disconnected-neighbors check to the existing ghosting functor, then we're done.  If not ... I guess as a first pass we can do things this way, but at the very least `#ifdef LIBMESH_ENABLE_RTTI` we should be checking for an existing `DisconnectedNeighborCoupling` and not adding another one redundantly.

tests/systems/disjoint_neighbor_test.C
```diff
@@ -0,0 +1,541 @@
...
+            const Number exact = heat_exact(p,params,"","");
+            const Number approx = sys.point_value(0,p);
+            LIBMESH_ASSERT_NUMBERS_EQUAL(exact, approx, 1e-2);
+          }
+      }
+  }
+
+
+  void testTempJumpRefine()
+  {
```
roystgnr, line 340: Can you add PerfLog entries (`LOG_UNIT_TEST;`) to these?  Adding unit tests that require a solve makes me wary, and I'd at least like to see *how* wary I should have been.

include/mesh/mesh_base.h
```diff
@@ -1820,8 +1830,27 @@ class MeshBase : public ParallelObject
...
+
+  const PeriodicBoundaries * get_disconnected_boundaries() const;
+#endif
+
+
 protected:
 
+ #ifdef LIBMESH_ENABLE_PERIODIC
+  /// @brief The disconnected boundary id pairs.
+  std::unique_ptr<PeriodicBoundaries> _disconnected_boundary_pairs;
```
roystgnr: And now I finally see something I should have noticed right away, the fact that I *don't* see some things:

We're not assigning these in `operator=` or the copy constructor, and we're not testing for size equality in `operator==` via `locally_equals`.  We definitely need both those things to be done, as a matter of principle, and because MOOSE will happily `clone()` meshes left and right and we don't want to lose `_disconnected_boundary_pairs` or violate sanity checks when it does.

I'd say we should also be doing *something* in `UnstructuredMesh::stitching_helper`, given how heavily MOOSE users rely on that these days.  Probably a `if (have_disc_bdys) libmesh_deprecated_msg()` there is sufficient for now, and users can work around it by adding disconnected boundary pairs only at the end of their mesh generation process, but we can't just silently drop disconnected boundaries from one of two stitched mesh inputs.

build-aux/ltmain.sh
ChengHauYang: This text was generated automatically when I ran autoreconf on my local machine. I have reverted this and other Makefiles to their original state because I removed `DisconnectedNeighborCoupling.[C,h]`.

src/ghosting/disconnected_neighbor_coupling.C
ChengHauYang: Thanks, @roystgnr. I have removed this and put the logic in `GhostPointNeighbors ` instead.

src/mesh/unstructured_mesh.C
ChengHauYang: Sure. I removed the call here. This is currently being handled within `GhostPointNeighbors` (the existing ghosting functor). Thanks for the feedback.

src/mesh/mesh_refinement.C
ChengHauYang: Sure. I have modified it.

src/mesh/boundary_info.C
ChengHauYang: You are welcome! I added one unit test in `tests/mesh/boundary_info.C`

tests/systems/disjoint_neighbor_test.C
ChengHauYang: Sure! I added them.

include/mesh/mesh_base.h
ChengHauYang: Thanks, Roy. This is a good catch!
I have implemented:
1. `operator= ` includes `_disconnected_boundary_pairs`.
2. `locally_equals` includes checking for `_disconnected_boundary_pairs`.
3. `stitching_helper ` includes `if (have_disc_bdys) libmesh_deprecated()`.

Thank you for your feedback!

build-aux/ltmain.sh
roystgnr: I thought that might have been what happened.  It's a moot point now, so don't waste much time looking into it, but I'd still love to understand what happened.  Most of your autoreconf changes were straightforward output of your autotools input file changes, but not this file.  Has our version of libtool gone out of date, and you did an autoreconf with something newer?

include/mesh/mesh_base.h
roystgnr: I swear my conscious mind was thinking `libmesh_not_implemented_msg` here, and I have no idea what malfunction in my brain made my fingers type `libmesh_deprecated_msg` instead.  Could you correct it?  Proper stitching of meshes with disconnected boundary pairs is something we could implement if users turn out to need it, not something we used to support but are now deprecating.

build-aux/ltmain.sh
ChengHauYang: I think the difference may come from my local versions.
Locally I ran `autoreconf -fi`
with
1. automake 1.18.1
2. autoconf 2.72
3. libtool 2.5.4
while CIVET may be built uses the other toolchain.
So ltmain.sh and a few Makefile.in files look different even though the .am files didn’t change.

include/mesh/mesh_base.h
ChengHauYang: Sure, @roystgnr! 😃 
I have changed `libmesh_deprecated` to `libmesh_not_implemented_msg `.
Thanks!

build-aux/ltmain.sh
roystgnr: Good thought, but those are our exact versions too.  I don't suppose you've got a version of those with patches or non-default settings?  IIRC I had some hassle a decade ago when Red Hat configured libtool one way and Debian or Ubuntu another.

src/mesh/mesh_base.C
```diff
@@ -199,6 +202,10 @@ MeshBase& MeshBase::operator= (MeshBase && other_mesh)
   _node_integer_default_values = std::move(other_mesh._node_integer_default_values);
   _point_locator_close_to_point_tol = other_mesh.get_point_locator_close_to_point_tol();
 
+#ifdef LIBMESH_ENABLE_PERIODIC
+  _disconnected_boundary_pairs = std::move(other_mesh._disconnected_boundary_pairs);
```
roystgnr: This should probably work for the move constructor (it's possible for users to derive a PeriodicBoundary subclass that would break here because its `clone()` kept a reference to the original mesh, but they'd have to work at it), but we need a copy in the copy constructor above too.

roystgnr: The copy constructor copy is going to be a bit trickier - each entry in the new PeriodicBoundaries will need to be a `clone()` of the corresponding entry from the source.

src/mesh/mesh_base.C
```diff
@@ -305,6 +312,12 @@ bool MeshBase::locally_equals (const MeshBase & other_mesh) const
     return false;
   if (*boundary_info != *other_mesh.boundary_info)
     return false;
+  // First check whether the "existence" of the two pointers differs (one present, one absent)
+  if ((bool)_disconnected_boundary_pairs != (bool)other_mesh._disconnected_boundary_pairs)
+    return false;
+  // If both exist, compare the contents
+  if (_disconnected_boundary_pairs && (*_disconnected_boundary_pairs != *other_mesh._disconnected_boundary_pairs))
```
roystgnr: This definitely isn't right ... which also means we're lacking test coverage.

The test is easy enough to add: build a `mesh` with DBPs, clone it, and `CPPUNIT_ASSERT(mesh == cloned_mesh)`.

The fix here is trickier.  We don't have any sort of virtual operator== equivalent for `PeriodicBoundaryBase`, and since we clone PBCs  we definitely can't just compare two `unique_ptr` and demand that they're equal.  We'll probably have to do the same weak test we do with ghosting functors, where we just assert that the containers are the same size and don't go deeper into the contents.

roystgnr (changes requested):
Getting closer, at least!

With subsequent changes, could you append commits rather than force-pushing them?  It is literally asymptotically faster to re-review when I only need to look at new changes rather than all changes.

src/mesh/mesh_base.C
ChengHauYang: Thanks, @roystgnr! In both the move assignment operator and the copy constructor, I now perform a deep copy of each entry in `_disconnected_boundary_pairs` using `clone()`.

src/mesh/mesh_base.C
ChengHauYang: Sure. Thanks for the feedback! I’ve updated the comparison to only check the size of the disconnected boundary pairs, and I also added a unit test with a disconnected boundary that verifies the cloned mesh equals the original one.

build-aux/ltmain.sh
ChengHauYang: I’m using the Conda build of GNU Libtool 2.5.4, which may not be identical to the upstream GNU release. It’s possible that my local macOS (Apple M4) environment introduces some CLI differences compared to the standard Linux setup.

src/mesh/mesh_base.C
roystgnr: Hmm... that's probably not necessary in the move constructor, but it probably is safer; I like it.

roystgnr:
This was kind of a brutal review process; sorry!  Thanks for pushing through it with me!  Though, honestly, even after all that this is a much less massive change than I'd have expected it to become.  We are still at the most basic stages if we're not supporting stitching or refinement yet, but I wouldn't have expected even the most basic support to be possible with only ~400 lines of library code.  I've still got a "waiting for the other shoe to drop" feeling, and I won't be too surprised if trying to use this in MOOSE this brings up some unpleasant surprises, but with ~600 lines of tests hopefully we've preempted anything *too* nasty.

ChengHauYang:
Thank you, @roystgnr, for providing me with much constructive feedback during this process! Really appreciate your guidance!

hugary1995:
We should probably test idaholab/moose#31636 against this PR branch and make sure it's green.

build-aux/ltmain.sh
jwpeterson: Just to be clear, there are no longer any stray bootstrap changes on this branch, i.e. the changes in Makefile.in are intentional, correct? IMO, the best approach if you need to run bootstrap again is to run: 
```
./bootstrap --build-autotools
```
and then temporarily change your `$PATH` so that the autotools built by libmesh come first, e.g. 
```
export PATH=$HOME/projects/libmesh/contrib/autotools/bin:$PATH
```
that way you get consistent versions for all the miscellaneous utilities (`libtool` and `libtoolize` for example) that bootstrap might end up calling.

jwpeterson:
> I want to give @jwpeterson a chance to look at this too before merging,

It sounds like you've already reviewed it pretty thoroughly so :+1: from me.

hugary1995:
I think it'll be a pain to enable CI for https://github.com/idaholab/moose/pull/31636 for @ChengHauYang with a libmesh submodule update. Let's just go ahead and merge this first, and we can wait for the next libmesh update in moose.

@ChengHauYang can you confirm that your changes on the moose side are compatible with this libmesh PR branch? I would like to make sure tests related to `BreakMeshByBlockGenerator` are passing (or can be easily adapted).

ChengHauYang:
Hi @hugary1995 ,  

I’ve removed most instances of `mesh_mode = 'replicated'` in the test files and reran the following tests:

MOOSE core tests  
```
./run_tests --re=break_mesh_by_block_generator --distributed-mesh -j 8
./run_tests --re=break_mesh_by_block_generator -j 8

./run_tests --re=interface_material --distributed-mesh -j 8
./run_tests --re=interface_material -j 8

./run_tests --re=equal_value_boundary_constraint --distributed-mesh -j 8
./run_tests --re=equal_value_boundary_constraint -j 8

./run_tests --re=no_failed_point_inversions --distributed-mesh -j 8
./run_tests --re=no_failed_point_inversions -j 8
```

Solid Mechanics module  
```
./run_tests --re=cohesive_zone_model --distributed-mesh -j 8
./run_tests --re=cohesive_zone_model -j 8
```

Heat Transfer module  
```
./run_tests --re=thin_layer_heat_transfer --distributed-mesh -j 8
./run_tests --re=thin_layer_heat_transfer -j 8
```


Contact module  
```
./run_tests --re=cohesive_zone_model --distributed-mesh -j 8
./run_tests --re=cohesive_zone_model -j 8
```


Combined module  
```
./run_tests --re=break_mesh_interface_contact --distributed-mesh -j 8
./run_tests --re=break_mesh_interface_contact -j 8
```


All tests above now pass successfully in both replicated and distributed modes.  

**However, one test (`stitch_boundary_mesh_generator`) still fails, since stitching after `BreakMeshByBlock` doesn’t yet handle disconnected neighbors. I’ve started modifying `stitching_helper` to add support for this case. The changes help fix broken tests related to stitching in MOOSE.**

@roystgnr, could you please review this section of the code in my latest commit ([8fb68ab](https://github.com/libMesh/libmesh/pull/4283/commits/8fb68ab0b97f52699329c000f62d337e6ef2f34e))? I really appreciate your help!

src/mesh/mesh_base.C
```diff
@@ -2025,6 +2025,34 @@ void MeshBase::detect_interior_parents()
...
+                                                     const boundary_id_type b2)
+    {
+      // Nothing to remove if not allocated or empty
+      if (!_disconnected_boundary_pairs || _disconnected_boundary_pairs->empty())
+        return;
+
+      auto & pairs = *_disconnected_boundary_pairs;
+
+      // Helper to check and erase both directions
+      auto erase_if_match = [&](boundary_id_type key, boundary_id_type pair)
```
roystgnr: I'm not a big fan of the "just capture everything" symbol in general, but in a subroutine this small I'm okay with it.

include/mesh/unstructured_mesh.h
```diff
@@ -310,7 +310,8 @@ class UnstructuredMesh : public MeshBase
                                 bool skip_find_neighbors,
                                 bool merge_boundary_nodes_all_or_nothing,
                                 bool remap_subdomain_ids,
-                                bool prepare_after_stitching);
+                                bool prepare_after_stitching,
+                                bool stitching_intended_disconnected_neighbors = true);
```
roystgnr: With the other parameters we're not documenting them here, because they're just being passed in from the non-helper functions where they are documented, but ... this one doesn't seem to be user-controllable?  It's a parameter to a private method, and it defaults to `true`, and we never call that method with `false`, so it will never be `false`, right?  If we want to expose this to the user then we should probably add it to the `public:` stitching methods.  If we don't want to expose it to the user (which IMHO is reasonable; users who don't want to include the disconnected-neighbor boundaries when stitching can delete them manually) then we should remove the extra parameter from this method; it's already got a ridiculous number of them.

src/mesh/unstructured_mesh.C
```diff
@@ -1793,13 +1835,49 @@ UnstructuredMesh::stitching_helper (const MeshBase * other_mesh,
...
+  auto * other_db = (other_mesh ? other_mesh->get_disconnected_boundaries() : nullptr);
+
+  const bool have_disc_bdys = ((this_db && !this_db->empty()) ||
+                              (other_db && !other_db->empty()));
+
+  if (have_disc_bdys && stitching_intended_disconnected_neighbors)
+    {
+      account_for_disconnected_boundaries = true;
+
+      // In most cases, other_mesh is nullptr (i.e., stitching the mesh to itself).
```
roystgnr: I don't think this is true

src/mesh/unstructured_mesh.C
```diff
@@ -1793,13 +1835,49 @@ UnstructuredMesh::stitching_helper (const MeshBase * other_mesh,
...
-                                    bool prepare_after_stitching)
+                                    bool prepare_after_stitching,
+                                    bool stitching_intended_disconnected_neighbors)
 {
 #ifdef DEBUG
   // We rely on neighbor links here
   MeshTools::libmesh_assert_valid_neighbors(*this);
 #endif
 
+  // Determine whether to activate disconnected boundary stitching:
```
roystgnr: This is ambiguous.  Does "disconnected boundary stitching" mean stitching across disconnected boundaries within a mesh?  Preservation of disconnected boundaries from input meshes?  From context I'd have assumed the latter but the phrase sounds like the former.

include/mesh/unstructured_mesh.h
roystgnr: If we do keep this, too, is there a better name to use?  I know we can't squeeze a full doc paragraph into 4 words, but I honestly can't parse this.

src/mesh/unstructured_mesh.C
```diff
@@ -2555,6 +2639,11 @@ UnstructuredMesh::stitching_helper (const MeshBase * other_mesh,
         }
     }
 
+#ifdef LIBMESH_ENABLE_PERIODIC
+  if (account_for_disconnected_boundaries)
+    this->remove_disconnected_boundaries_pair(this_mesh_boundary_id, other_mesh_boundary_id);
```
roystgnr: Okay, so we're just handling the case where we have a disconnected_boundaries pair and we're trying to stitch across it.  That leaves me with questions:

When would that ever occur with one side of the pair in `this` and the other in `other_mesh`?  Shouldn't `this` already be pairing with itself and vice-versa?

What happens if `other_mesh` has disconnected_boundaries pairs?  We don't seem to be copying them - are we just throwing them away?  That's the sort of bug we've had to fix repeatedly, e.g. in #2850, and that's actually what I *thought* we needed to start supporting here; I foolishly hadn't considered the (in hindsight quite reasonable) case where we had a disconnected boundary pair and wanted to connect it.

roystgnr:
That "No MPI/No PETSc" failure looks like a false positive here, something brought on by https://github.com/libMesh/libmesh/pull/4300

include/mesh/unstructured_mesh.h
roystgnr: I'm even more confused after getting to the code.  If this setting is just turning on support for stitching a disconnected boundary pair together, why would anyone ever want to stitch with it false?  Wouldn't that just be not-stitching, but slower?

roystgnr (changes requested):

src/mesh/mesh_base.C
ChengHauYang: Thanks, Roy! I updated the lambda to avoid the “capture everything” pattern.
Now it takes pairs as a reference argument instead of capturing it by `[&]`:
```
      // Helper to check and erase both directions
      auto erase_if_match = [](boundary_id_type key,
                              boundary_id_type pair,
                              PeriodicBoundaries & pb_map)
        {
          auto it = pb_map.find(key);
          if (it != pb_map.end())
            {
              const auto & pb = *(it->second);
              // Check both directions
              if ((pb.myboundary == key && pb.pairedboundary == pair) ||
                  (pb.pairedboundary == key && pb.myboundary == pair))
                pb_map.erase(it);
            }
        };
```

include/mesh/unstructured_mesh.h
ChengHauYang: > we should remove the extra parameter from this method; it's already got a ridiculous number of them

Agree

The original reason for adding this parameter was to allow users to stitch across regular boundaries but skip stitching the disconnected (pseudo/fake) neighbor pairs.
However, whether such stitching happens already depends entirely on the input arguments `this_mesh_boundary_id` and `other_mesh_boundary_id`.

As long as the user doesn’t specify a pair of boundary IDs (`this_mesh_boundary_id` and `other_mesh_boundary_id`) that correspond to a disconnected boundary pair, there’s no need for an extra flag -- and if they do specify such a pair, setting this flag to false would simply mean “not stitching, but slower,” exactly as you mentioned.

So yes, I’ll remove this `stitching_intended_disconnected_neighbors` argument to simplify the interface.

src/mesh/unstructured_mesh.C
ChengHauYang: Thanks! That’s exactly right.

After rethinking this, the current implementation assumes that both sides of each disconnected boundary pair are always defined "within the same mesh". Therefore, the stitching logic only handles the self-stitching case, stitching elements across the registered disconnected boundary pairs "within a single mesh".

Cross-mesh disconnected stitching is not supported by design. This keeps the logic simpler and consistent with the concept of disconnected_boundaries, which represent internal connectivity information within one mesh rather than between meshes.

If `other_mesh` also needs to stitch its own disconnected boundary pairs, the user can simply call `other_mesh->stitching_helper()` separately. We would not take care of this here.

ChengHauYang:
Thanks, @roystgnr, for taking your time to review my changes! I’ve updated the implementation and clarified the comments. Could you give me some feedback if you are available? Thanks!
The function now explicitly handles only the disconnected boundary pairs registered within `this` mesh, while ignoring those defined in `other_mesh`. Cross-mesh disconnected-boundary stitching would not happen. (These details are now properly documented within the function comments.)
I revised the code and made the comments more understandable.
I hope this resolves the earlier confusion; thank you again for the helpful discussion!

src/mesh/unstructured_mesh.C
roystgnr: This is still wrong, though, if it's not copying disconnected boundary pairs from a distinct `other_mesh` to `this`.  We need to `libmesh_not_implemented_msg()` in that case in the very least, but I don't think adding the copy (and a test for it, for that matter) would be hard.  Ideally we'd want to sanity check as we go (if a and b also exist as boundary ids on `this`, we should be screaming and dying, not adding neighbor pointers between them) too.

Right now, it looks to me like, if we stitch `other_mesh` X with disconnected boundary pair a,b into `this` mesh Y!=X, then we just silently drop the PeriodicBoundary objects for a,b.  We can't drop mesh features when we're stitching them.

roystgnr: A unit test here would probably only take 4 elements in a row - elements 1 and 2 with a disconnected boundary in between them, elements 3 and 4 likewise, then stitch the other side of 2 to the other side of 3, and make sure that both the 1-2 and the 3-4 pairing still exist afterward.

roystgnr: Actually, on the sanity checking: if a and b already exist on `this` we should be screaming and dying *unless* they already have PeriodicBoundary objects connecting them too.  It could be useful to users to be able to extend a boundary pair via the mesh they're stitching in.

src/mesh/unstructured_mesh.C
ChengHauYang: I removed the misleading sentence about other_mesh usually being nullptr. stitching_helper is used both for "self-stitching" and for "stitching two meshes"; I shouldn’t imply a typical case here.

src/mesh/unstructured_mesh.C
ChengHauYang: We have already renamed the "disconnected" boundary to be the "disjoint neighbor" boundary. It is clearer now.
What I want to do here is to determine whether we want to stitch the "disjoint neighbor" boundary.

src/mesh/unstructured_mesh.C
ChengHauYang: > if it's not copying disconnected boundary pairs from a distinct other_mesh to this. We need to libmesh_not_implemented_msg() in that case in the very least, but I don't think adding the copy (and a test for it, for that matter) would be hard.

I have already copy disjoint neighbor boundary pairs from `other_mesh` to `this` in the newest implementation.

> Actually, on the sanity checking: if a and b already exist on this we should be screaming and dying unless they already have PeriodicBoundary objects connecting them too.

Yes. Right now I have that implementation in the newest commit: duplicates are skipped, mismatches raise errors.

> A unit test here would probably only take 4 elements in a row - elements 1 and 2 with a disconnected boundary in between them, elements 3 and 4 likewise, then stitch the other side of 2 to the other side of 3, and make sure that both the 1-2 and the 3-4 pairing still exist afterward.

I added two unit tests for this: 
(a) One is `this` mesh `X` has four disjoint elements (other mesh `Y` is nullptr). After stitching elem2 and elem3, check whether 1-2 and the 3-4 pairing still exist.
(b) `X` has two disjoint elements and `Y` has two disjoint elements. After stitching elem2 and elem3, check whether 1-2 and the 3-4 pairing still exist.

ChengHauYang:
Hi @roystgnr,

Thanks for the feedback!! The newest two commits try to address your comments:

(a) Rename `_disconnected_boundary_pairs` to be `_disjoint_neighbor_boundary_pairs` for clarity. All related identifiers using "disconnected" have been updated to "disjoint."

(b) `remove_disjoint_boundary_pair` only removes the pair when we successfully stitch two disjoint neighbor boundaries together, so that they should not be considered as "disjoint boundary pair" anymore. I added two unit tests for this: 
One is `this` mesh `X` has four disjoint elements (other mesh `Y` is nullptr). After stitching elem2 and elem3, we check whether the 1–2 and 3–4 pairings still exist.
The second one is that `X` has two disjoint elements and `Y` has two disjoint elements. After stitching elem2 and elem3, we check whether the 1–2 and 3–4 pairings still exist.

(c) Copy disjoint neighbor boundary pairs from `other_mesh` to `this` in `stitching_helper`. If `(a,b)` from `other_mesh` already exists on `this`, we would be screaming unless they already have PeriodicBoundary objects connecting them.

Could you help me review the newest commits if you are available? Thanks!!

roystgnr (approved):
This looks like it's in great shape now.

roystgnr:
MOOSE clang recipe failures are an unrelated downstream bug.

roystgnr:
Looks like one of the unit tests here breaks in our `Test No Exceptions` recipe.  I'll have a fix up in a sec.

PR #4285, opened 2025-10-21
Verify PointLocator validity

roystgnr:
This was useful in verifying for me that a potential downstream bug was just a red herring; it might be useful for others' debugging too.

Actually running this after every construction is expensive, but so is dbg in general; I don't think it's more much expensive than some of our existing tests ... or even much more expensive than the PointLocatorTree construction, come to think of it.


roystgnr:
Well, that swath of timeouts means so much for my "it'll be cheap enough to run in dbg mode" theory.

jwpeterson (approved):

PR #4286, opened 2025-10-22
DegenerateMapping exception for bad Jacobians and/or Newton convergence in FEMap

roystgnr:
In https://github.com/idaholab/moose/pull/31758 I took the position that we should avoid making it harder to fix the issue where we report problematic element mappings as LogicError exceptions even though they're not logic errors; after a little thought, this isn't a hard issue and I could probably spend more time arguing about hypothetically fixing it than it would take to actually fix it.

In the near term, I think a good migration path here is for user code to `catch (std::exception & e)` but then check whether `e.what()` matches `"Jacobian"` to determine whether it's caught a mapping issue rather than a real internal error or assertion failure.

### DIFF
--- a/src/fe/fe_map.C
+++ b/src/fe/fe_map.C
@@ -483,7 +483,7 @@ void FEMap::compute_single_point_map(const unsigned int dim,
                 libmesh_degenerate_mapping_msg
                   ("Jacobian " << det_J << " at or under tolerance " <<
                    jacobian_tolerance << " at point " << xyz[p] <<
-                   " in element: " << elem_info);
+                   " in element:\n" << elem_info);
               }
             else
               {
@@ -492,7 +492,7 @@ void FEMap::compute_single_point_map(const unsigned int dim,
                 libmesh_degenerate_mapping_msg
                   ("Jacobian " << det_J << " at or under tolerance " <<
                    jacobian_tolerance << " at point index " << p <<
-                   " in element: " << elem_info);
+                   " in element:\n" << elem_info);
               }
           }
         else


roystgnr:
Good MOOSE framework: "A zero mapping Jacobian isn't necessarily an error; it's an exception that we can often catch and recover from."

Evil MOOSE test suite: "If a zero mapping Jacobian doesn't output the exact phrase `ERROR: negative Jacobian 0` then you don't pass."

roystgnr:
Rebasing on devel, to see if MOOSE CI has gotten any happier with this after my compatibility improvements and fixes downstream.

roystgnr:
`MOOSE` `clang` failures are known and unrelated; `Test mac` failure seems new but it's definitely unrelated, and it looks intermittent so I'm just going to kick it.

roystgnr:
Adding a couple more recipes; John wisely suspects that this is going to fail `Test No Exceptions` without an ifdef or two.

src/reduced_basis/rb_data_deserialization.C
```diff
@@ -329,14 +327,12 @@ void load_parameter_ranges(RBParametrized & rb_evaluation,
         discrete_parameter_values[discrete_names[i]] = values;
       }
   }
-#ifdef LIBMESH_ENABLE_EXCEPTIONS
-  catch (std::exception & e)
+  libmesh_catch (std::exception & e)
   {
     libmesh_error_msg("Error loading parameter ranges from capnp reader.\n"
                       "This usually means that the training data either doesn't exist or is out of date.\n"
                       "Detailed information about the error is below:\n\n" << e.what() << "\n");
```
jwpeterson, line 337: I chatted with @roystgnr, and he's attaching or already attached a "Test No Exceptions" recipe. I think we can't refer to `e.what()` here when exceptions are disabled, this is the same error that @pbehne ran into.

roystgnr: But we're foiled in our testing by `!LIBMESH_HAVE_CAPNPROTO` - I'll fix this regardless.

roystgnr: Yeah, I should just drop this commit.  Looks like the existing code was correct and I shouldn't risk breaking it to save a couple preprocessor lines.

PR #4287, opened 2025-10-22
Variational smoother: tet support

pbehne:
This PR adds support and testing for using the `VariationalMeshSmoother` on tets.

Ref. #4082

### DIFF
--- a/tests/mesh/mesh_smoother_test.C
+++ b/tests/mesh/mesh_smoother_test.C
@@ -20,7 +20,6 @@
 
 #include "test_comm.h"
 #include "libmesh_cppunit.h"
-#include <iomanip>
 
 #include <random>
 
@@ -647,7 +646,6 @@ public:
       }
 
     smoother.smooth();
-    mesh.write("smoothed.e");
 
     if (type_is_tri)
       {


include/systems/variational_smoother_constraint.h
```diff
@@ -517,6 +504,19 @@ class VariationalSmootherConstraint : public System::Constraint
...
   virtual void constrain() override;
+
+  /**
+   * Given a mesh and a node in the mesh, the vector will be filled with
+   * every node directly attached to the given one. IF NO neighbors are found,
+   * all nodes on the containing side are treated as neighbors. This is useful
+   * when the node does not lie on an edge, such as the central face node in
+   * HEX27 elements.
+   */
+  static void find_nodal_or_face_neighbors(
```
roystgnr: I hate moving things to `public:` before we need to, lest we find that someone starts using them and we need to change the API and now we have to worry about a transition path for compatibility.  It doesn't look to me like we'll need to change this API, but what do you think?

src/systems/variational_smoother_system.C
```diff
@@ -1083,6 +1083,67 @@ VariationalSmootherSystem::get_target_elem(const ElemType & type)
...
+      // the base face and h is the height of the apex node. The area of an
+      // equilateral triangle with side length s is b = sqrt(3) s^2 / 4.
+      // For all faces to have side length s, the height of the apex node is
+      // h = sqrt(2/3) * s. Then the volume is v = sqrt(2) * s^3 / 12.
+      // Solving for s, the side length that will preserve the volume of the
+      // reference element is s = (12 * v / sqrt(2))^(1/3), where v is the volume
+      // of the non-optimal reference element (i.e., a right tet).
+
+      // Side length that preserves the volume of the reference element
+      const auto side_length = std::pow(12. * ref_vol / sqrt_2, 1. / 3.);
```
roystgnr: So ... `6 * sqrt_2 * ref_vol`?  I swear I was *just* teaching the middle schoolers about "simplest radical form" yesterday.

Also, could we make that `1/Real(3)`?  Otherwise we're truncating in higher-than-double precision.  I don't know when we're ever going to care about smoothing in higher-than-double precision, but it's a good idiom to go to by default.

tests/mesh/mesh_smoother_test.C
```diff
@@ -345,6 +351,54 @@ public:
...
+    unsigned int num_centered = 0;
+    unsigned int num_face = 0;
+    for (const auto d : make_range(3))
+      {
+        const auto num_half_lengths = point(d) / half_length;
+        if (std::abs(num_half_lengths - std::round(num_half_lengths)) > tol)
+          // This coordinante does not occur at a sub-cube face or center
+          continue;
+
+        if ((unsigned int)std::round(num_half_lengths) % 2)
```
roystgnr: Don't let John catch you writing a C-style cast!

Also, do we want to `std::round` and then cast?  Wouldn't `std::lround` be more straightforward?

tests/mesh/mesh_smoother_test.C
```diff
@@ -686,6 +746,207 @@ public:
...
+                                    const Real x = (type == TET10) ? 0.42895 : 0.41486;
+                                    CPPUNIT_ASSERT(node.relative_fuzzy_equals(
+                                        other + x * (cube_center - other), 1e-3));
+                                  }
+
+                                else if (pointIsCubeVertex(other, side_length))
+                                  {
+                                    const Real x = (type == TET10) ? 0.55389 : 0.58094;
+                                    CPPUNIT_ASSERT(node.relative_fuzzy_equals(
+                                        other + x * (cube_center - other), 1e-3));
```
roystgnr: Not sure if it never occurred to me to ask before or if I asked and just forgot the answer, but ... we can't get better than 1e-3 here?  I'd have thought we could at least pull off the default `TOLERANCE` in most cases, and 1e-4 in the high-order Tet+Pyramid cases.

include/systems/variational_smoother_constraint.h
pbehne: This method is an extension of the `MeshTools::find_nodal_neighbors` method. Since it has the same signature, I'd say we probably won't need to change the API. However, I could make it private and have the `VariationalSmootherConstraint` class be a friend of the `MeshSmootherTest` class. Would you prefer that?

roystgnr: That's a good point - this really doesn't depend on the variational smoothing, right?  It's a tool used by the variational smoother?  Since the API looks solid enough, would it make sense to move it into `MeshTools` too?

tests/mesh/mesh_smoother_test.C
pbehne: I didn't know about `lround`. Thanks for the tip!

tests/mesh/mesh_smoother_test.C
pbehne: On nice, I was able to tighten the tolerances now that the target element bugs have been patched. I'll play with tightening the tolerances for other element types in another PR!

include/systems/variational_smoother_constraint.h
pbehne: > would it make sense to move it into MeshTools too?
Works for me, will do!

src/systems/variational_smoother_system.C
jwpeterson: As long as we're picking nits, you could consider using [std::cbrt](https://en.cppreference.com/w/cpp/numeric/math/cbrt.html) and sidestep the issue with the precision of Real(3) vs. 3.

roystgnr (approved):
I'm happy with this as-is; I do like John's `cbrt` suggestion, though.  Looks like I missed the same opportunity in the Netgen interface code, and we've got the same problem in a couple example codes...

roystgnr:
Test mac failure is unrelated

PR #4289, opened 2025-10-23
Pass along flip_tris parameter

roystgnr:
I have no idea why neither g++ nor clang++ were giving me an unused parameter warning here, but thanks to nvc++ for catching it; we haven't been properly testing both orientations of manifolds in our tetrahedralization code!


roystgnr:
Test mac is failing due to PETSc upgrade deprecation warnings, unrelated.

PR #4290, opened 2025-10-24
Delete auto_ptr.h

jwpeterson:
As of idaholab/moose@6a8a56e82a, MOOSE no longer includes this header, but it hasn't been relevant (containing only deprecated code) for a much longer time than that, so I think this is a good time to get rid of it. To be clear, this header will still be in the 1.9.x series which isn't finalized yet, but will be gone in 1.10.x, whenever we get around to creating it.


jwpeterson:
We have another PR (#4291) that fixes the Test mac error (but triggers a different error in Min gcc). So I will merge this now since we are not letting the Test mac failure hold up otherwise unrelated PRs.

PR #4293, opened 2025-10-27
Updated LaplaceMeshSmoother signatures to look like the VariationalMeshSmoother

pbehne:
This involved moving the `n_iterations` parameter from the `smooth` method to the constructor and storing it as a class attribute. The old functionality is deprecated.

Ref #4082

### DIFF
--- a/src/mesh/mesh_smoother_laplace.C
+++ b/src/mesh/mesh_smoother_laplace.C
@@ -57,7 +57,7 @@ void LaplaceMeshSmoother::smooth() {
   // determined. We store the new positions here
   std::vector<Point> new_positions;
 
-  for (const auto n : make_range(_n_iterations))
+  for (unsigned int n=0; n<_n_iterations; n++)
     {
       new_positions.resize(_mesh.max_node_id());
 


src/mesh/mesh_smoother_laplace.C
```diff
@@ -63,62 +57,62 @@ void LaplaceMeshSmoother::smooth(unsigned int n_iterations)
...
-
-      // Now the nodes which are ghosts on this processor may have been moved on
-      // the processors which own them.  So we need to synchronize with our neighbors
-      // and get the most up-to-date positions for the ghosts.
-      SyncNodalPositions sync_object(_mesh);
-      Parallel::sync_dofobject_data_by_id
-        (_mesh.comm(), _mesh.nodes_begin(), _mesh.nodes_end(), sync_object);
-
-    } // end for n_iterations
+  for (unsigned int n = 0; n < _n_iterations; n++) {
```
roystgnr: Yeah, let's fix all this.  Whitespace changes should be in separate commits, and also they should change formatting *to* libMesh standards, not *away from* libMesh standards.  ;-)

roystgnr (changes requested):

roystgnr (approved):

roystgnr:
Failure unrelated

PR #4297, opened 2025-10-29
M4 typo

roystgnr:
I got an error from this during a manual `--disable-optional` build; not sure why that happened, or why it doesn't happen to our --disable-optional recipe, but the fix is straightforward.


roystgnr:
I believe we have this idiom elsewhere, but `git grep --recurse-submodules '^ *\$\w*='` doesn't show this type of bug elsewhere.

PR #4298, opened 2025-10-29
--disable-optional nvc++ warning fixes

roystgnr:
No serious bugs here, but the tests/ fixes should very slightly reduce runtime by removing a few degraded unit tests from configurations that don't support running those tests for real.


roystgnr:
We'll need to backport #4303 too to avoid the regression here.  :-P

PR #4299, opened 2025-10-29
Check for our current blessed autotools version

roystgnr:
Users of older-but-compatible autotools can run autoreconf manually if they really don't want to upgrade.

After 2023, this marks twice in a row I've inadvertently sabotaged Alex's autotools verification.  Fingers crossed for 2027!


roystgnr:
Failures are unrelated

PR #4303, opened 2025-10-30
Disable more tests with --disable-amr

roystgnr:
This should fix that devel->master recipe; I added an assert to catch bad tests, and disabled some, but was in a configuration that was already disabling other tests for other reasons.


jwpeterson:
Cherry-picked to 1.8.x and 1.9.x release branches.

PR #4306, opened 2025-11-03
Variational smoother: distributed mesh testing

pbehne:
In order to use the `VariationalMeshSmoother` on distributed meshes, this PR parallelizes some logic in the nodal constraints and smoother tester that was previously lacking. The test suite has been updated to perform testing on distributed rather than replicated meshes.

Ref #4082.

### DIFF
--- a/tests/mesh/mesh_smoother_test.C
+++ b/tests/mesh/mesh_smoother_test.C
@@ -492,14 +492,25 @@ public:
           // smallest distance
           auto get_closet_point_accross_all_procs =
               [&mesh, &proc_id](const std::map<dof_id_type, Real> &dist_map) {
-                // Iterator to the map entry with the smallest distance
-                auto min_it =
-                    std::min_element(dist_map.begin(), dist_map.end(),
-                                     [](const auto &a, const auto &b) {
-                                       return a.second < b.second;
-                                     });
 
-                const auto &d_min_local = min_it->second;
+                processor_id_type broadcasting_proc = 0;
+                Real d_min_local = std::numeric_limits<Real>::max();
+                dof_id_type node_id = DofObject::invalid_id;
+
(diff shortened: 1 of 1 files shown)


src/systems/variational_smoother_constraint.C
```diff
@@ -342,6 +342,7 @@ void VariationalSmootherConstraint::constrain()
 {
   const auto & mesh = _sys.get_mesh();
   const auto dim = mesh.mesh_dimension();
+  const auto &rank = mesh.comm().rank();
```
roystgnr:
```suggestion
  const auto &rank = mesh.processor_id();
```

Just to match our usual idioms

tests/mesh/mesh_smoother_test.C
```diff
@@ -457,6 +457,7 @@ public:
     MeshTools::Modification::redistribute(mesh, dh);
 
     const auto & boundary_info = mesh.get_boundary_info();
+    const auto &rank = mesh.comm().rank();
```
roystgnr:
```suggestion
    const auto &rank = mesh.processor_id();
```

tests/mesh/mesh_smoother_test.C
```diff
@@ -1023,199 +1069,199 @@ public:
 #ifdef LIBMESH_ENABLE_VSMOOTHER
   void testVariationalEdge2()
   {
-    ReplicatedMesh mesh(*TestCommWorld);
+    DistributedMesh mesh(*TestCommWorld);
```
roystgnr: Just make all these `Mesh`, not `DistributedMesh`; that way we still get `ReplicatedMesh` coverage in most of our tests but we get `DistributedMesh` coverage in the sweeps that test that.

roystgnr (changes requested):

PR #4308, opened 2025-11-04
Add ExodusII_IO option to set {node,elem} unique_ids based on {node,elem}_num_map

jwpeterson:
The basic idea here is that the Elems and Nodes will always be numbered sequentially starting from zero in libmesh while their "true" ids are stored as their unique_id (this idea is from @roystgnr). The idea behind this option is to better support non-contiguous numberings by simply using a contiguous numbering within libmesh and maintaining the {node,elem}_num_maps on the Elems and Nodes themselves until the information is written back to file. There will likely be a lot of follow-up work on this, but this PR adds the option (off by default) so that I can start experimenting further with it.

### DIFF
--- a/include/mesh/exodusII_io_helper.h
+++ b/include/mesh/exodusII_io_helper.h
@@ -316,14 +316,16 @@ public:
   dof_id_type get_libmesh_elem_id(int exodus_elem_id);
 
   /**
-   * Helper function that sets the unique_id of the passed-in
-   * Node/Elem if _set_unique_ids_from_maps == true, does nothing if
-   * that flag is false. The input index is assumed to be a zero-based
-   * index into the {node,elem}_num_map array.
+   * Helper function that conditionally sets the unique_id of the
+   * passed-in Node/Elem.  Calling this function does nothing if
+   * _set_unique_ids_from_maps == false, otherwise it sets the
+   * unique_id based on the entries of the {node,elem_num_map}.  The
+   * input index is assumed to be a zero-based index into the
+   * {node,elem}_num_map array.
    */
-  void set_node_unique_id(
+  void conditionally_set_node_unique_id(
     MeshBase & mesh, Node * node, int zero_based_node_num_map_index);
-  void set_elem_unique_id(
+  void conditionally_set_elem_unique_id(
     MeshBase & mesh, Elem * elem, int zero_based_elem_num_map_index);
 
 private:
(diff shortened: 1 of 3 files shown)


jwpeterson:
The "Test MOOSE GCC min" failures should be completely unrelated.
```
[300.0s]  TIMEOUT stochastic_tools/test:web_server_control.stochastic_control/batch_reset FAILED (TIMEOUT)
[1.159s]     FAIL phase_field/tutorials/spinodal_decomposition.s2_fasttest FAILED (CRASH)
```

include/mesh/exodusII_io.h
```diff
@@ -140,6 +140,23 @@ class ExodusII_IO : public MeshInput<MeshBase>,
...
+   *    elem_num_map and node_num_map ids directly on Nodes and Elems
+   *    read in from the Exodus file, set their unique_ids
+   *    instead. Normally libmesh chooses the unique_ids automatically
+   *    but in this case we override that choice.  As a consequence
+   *    the libMesh Elems/Nodes will be numbered based on their
+   *    implied ordering in the exo file (sequentially by block and
+   *    without any gaps in the numbering).
+   * .) When writing an Exodus file, populate the elem_num_map and
+   *    node_num_map with the unique_ids of the Elems/Nodes being
+   *    written.
```
roystgnr, line 156: Let's add a comment on why one might want to do this?  (ReplicatedMesh using `O(max_id)` storage)

include/mesh/exodusII_io_helper.h
```diff
@@ -305,6 +306,42 @@ class ExodusII_IO_Helper : public ParallelObject
...
+  dof_id_type get_libmesh_node_id(int exodus_node_id);
+  dof_id_type get_libmesh_elem_id(int exodus_elem_id);
+
+  /**
+   * Helper function that sets the unique_id of the passed-in
+   * Node/Elem if _set_unique_ids_from_maps == true, does nothing if
+   * that flag is false. The input index is assumed to be a zero-based
+   * index into the {node,elem}_num_map array.
+   */
+  void set_node_unique_id(
```
roystgnr: I do not like `set_foo_unique_id` for the name of a method that may set the unique id or may not do anything ... but I can't actually think of a better name, I'm afraid.  If you don't have any better ideas we can keep this.

jwpeterson: `possibly_set_node_unique_id()`
`conditionally_set_node_unique_id()`
?

roystgnr: Either of those is an improvement.  I like conditionally.

PR #4313, opened 2025-11-05
VariationalMeshSmoother improved error message

pbehne:
Added VariationalMeshSmoother-specific error message to any caught runtime_errors when the smoother solve fails.
This is useful for informing moose users what part of their input is crashing.

Ref. #4082.

### DIFF
--- a/src/mesh/mesh_smoother_vsmoother.C
+++ b/src/mesh/mesh_smoother_vsmoother.C
@@ -140,7 +140,15 @@ void VariationalMeshSmoother::smooth(unsigned int)
   if (!_setup_called)
     setup();
 
-  system()->solve();
+  try
+    {
+      system()->solve();
+    }
+
+  catch (std::runtime_error& e)
+    {
+      throw std::runtime_error(std::string("VariationalMeshSmoother: issue encountered during solve:\n") + e.what());
+    }
 
   // Update _mesh from _mesh_copy
   for (auto * node_copy : _mesh_copy->local_node_ptr_range())
--- a/src/mesh/mesh_smoother_vsmoother.C
+++ b/src/mesh/mesh_smoother_vsmoother.C
@@ -145,10 +145,10 @@ void VariationalMeshSmoother::smooth(unsigned int)
       system()->solve();
     }
 
-  libmesh_catch (std::runtime_error& e)
+  libmesh_catch (ConvergenceFailure& e)
     {
 #ifdef LIBMESH_ENABLE_EXCEPTIONS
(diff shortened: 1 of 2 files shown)


src/mesh/mesh_smoother_vsmoother.C
```diff
@@ -140,7 +140,15 @@ void VariationalMeshSmoother::smooth(unsigned int)
   if (!_setup_called)
     setup();
 
-  system()->solve();
+  try
```
roystgnr: Let's change these to libmesh_try and libmesh_catch for the `--disable-exceptions` builds?

roystgnr (changes requested):

roystgnr:
... how did that pass the Test No Exceptions recipe?

roystgnr:
`checking if mpicc supports -fno-rtti -fno-exceptions... no`

Well, there's something to look into, anyway ... though it looks like we're not actually *using* those flags in that recipe anyway, just flipping the `#define` that's supposed to make it safe to use `-fno-exceptions`.  We'll just merge this after prepending the try/catch, and hope nobody else has slipped something through that will break as soon as we start coaxing a compiler into breaking it.

roystgnr (approved):

src/mesh/mesh_smoother_vsmoother.C
```diff
@@ -145,7 +145,7 @@ void VariationalMeshSmoother::smooth(unsigned int)
       system()->solve();
     }
 
-  libmesh_catch (std::runtime_error& e)
+  libmesh_catch(std::runtime_error& e)
     {
       throw std::runtime_error(std::string("VariationalMeshSmoother: issue encountered during solve:\n") + e.what());
```
jwpeterson: I've seen the other discussion regarding the `-fno-exceptions` flag not actually working right now, but I wanted to add that it doesn't make sense to use `throw` either if exceptions are disabled... you'll want to change that to a `libmesh_error_msg()` I think.

roystgnr:
Oh, nasty.  That's not even going to compile if we change my `if (0)` trick to `if constexpr (0)`, is it?  We'll need actual `#ifdef` guards here, then.

At least the `Test No Exceptions` recipe turns out to be good for something, though.

src/mesh/mesh_smoother_vsmoother.C
roystgnr: If exceptions are disabled, our `libmesh_catch(foo)` lines turn into `if (0)` ... and I thought that was clever, but clearly that's not sufficient in every case.  We're going to need to `#ifdef LIBMESH_ENABLE_EXCEPTIONS` guard the whole catch block.

jwpeterson (approved):

PR #4314, opened 2025-11-06
Remove superfluous semicolon

roystgnr:
Not sure why we don't have more recipes screaming about this, but at least a few of the devel->master recipe compilers caught it.

### DIFF
--- a/src/numerics/sparse_matrix.C
+++ b/src/numerics/sparse_matrix.C
@@ -70,7 +70,7 @@ namespace
   {
     if (hdf5val < 0)
       libmesh_error_msg("HDF5 error from " + objname + " in " + filename);
-  };
+  }
 #endif
 }
 


roystgnr:
The only difference between the passing-on-devel "Min gcc" and the failing-on-devel "Min gcc debug" is `opt` vs `dbg` builds ... and there the difference is that `opt` doesn't set `-pedantic` (or `-W -Wall -Wextra -Wno-long-long -Wunused -Wpointer-arith -Wformat -Wparentheses -Woverloaded-virtual` ...)

Any objections to us starting to turn on all warnings on all build levels?  IMHO `opt` should entail skipping internal checks that might slow us down at runtime, but not warnings at compile time.  This obviously isn't urgent to fix, but I might just change it in ACSM and then we'll pick it up whenever we next update that for something more pressing.

jwpeterson:
Hmm... I guess I don't know why opt and dbg mode *warning* flags should be different, other than we have just done that historically. So I guess I'd be fine with making warning flags the same for all optimization levels.

roystgnr:
This passed at least one of the previously failing recipes; not going to wait on everything.

PR #4316, opened 2025-11-06
Variational smoother: increase code coverage

pbehne:
Set two unit test to maximum verbosity to increase code coverage. Ref #4082.

### DIFF
--- a/tests/mesh/mesh_smoother_test.C
+++ b/tests/mesh/mesh_smoother_test.C
@@ -1103,10 +1103,17 @@ public:
   void testVariationalQuad4()
   {
     Mesh mesh(*TestCommWorld);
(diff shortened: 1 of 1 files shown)


roystgnr (approved):
I generally like *reducing* verbosity in our unit tests, to reduce "chaff" when debugging other failures.  But it's hard to argue with "increase code coverage", so I'm good with this.

Crazy thought:  We're outputting to `libMesh::out`, right?  Would it be possible to temporarily (saving the original streambuf and restoring it after, as in libmesh.C) redirect `libMesh::out` to an `ostringstream`?  Then not only could we keep extra verbosity off the console, we could search the string afterward to make sure we're printing what we expect.

PR #4320, opened 2025-11-10
Drop deprecated MeshFunction::init() taking an arg

jwpeterson:
This has been deprecated since 149a8e85 (Feb 2021), which is in the 1.7.x (and all later) release series.

### DIFF
--- a/include/mesh/mesh_function.h
+++ b/include/mesh/mesh_function.h
@@ -109,18 +109,6 @@ public:
    */
   virtual void init () override;
 
-#ifdef LIBMESH_ENABLE_DEPRECATED
-  /**
-   * The actual initialization process.  Takes an optional argument which
-   * specifies the method to use when building a \p PointLocator
-   *
-   * \deprecated The input argument is not used (was it ever?) to
-   * control the PointLocator type which is built, so one should
-   * instead call the version of init() taking no args.
-   */
-  void init (const Trees::BuildType point_locator_build_type);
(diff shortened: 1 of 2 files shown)


jwpeterson:
FYI @roystgnr, the errors are unrelated, looks like there is an unused private field warning treated as an error in MOOSE:
```
/tmp/build/moose/framework/build/header_symlinks/FunctionInterface.h:150:23: error: private field '_fni_object' is not used [-Werror,-Wunused-private-field]
  const MooseObject & _fni_object;
                      ^
1 error generated.
make: *** [/tmp/build/moose/framework/build.mk:166: /tmp/build/moose/framework/build/unity_src/functions_Unity.x86_64-pc-linux-gnu.opt.lo] Error 1
make: *** Waiting for unfinished jobs....
[08:38:52] ERROR: Exiting with code 2
```

jwpeterson:
FYI @NamjaeChoi we have some failing builds (warning treated as an error) here which I think were caused by one of your recent changes in MOOSE.
```
/tmp/build/moose/framework/build/header_symlinks/FunctionInterface.h:150:23: error: private field '_fni_object' is not used [-Werror,-Wunused-private-field]
  150 |   const MooseObject & _fni_object;
      |                       ^
1 error generated.
```

NamjaeChoi:
Yes, it was noticed and a fix was pushed. Thanks for the notice

PR #4321, opened 2025-11-10
Fix compilation error in print_coreform_hdf5() when complex numbers are enabled

jwpeterson:
We do not currently support writing complex-valued SparseMatrices to HDF5 files, since current versions of HDF5 do not seem to have a native type corresponding to std::complex, although that seems like it will be changed in HDF5 soon, so we may revisit this then. We also now throw a "not implemented" error if this function is called with a non-complex `Number` type that is larger than `double`, since our implementation relies on copying the matrix values into `doubles`.


jwpeterson:
Test failures are the same (unrelated) -Werrors as before.

PR #4323, opened 2025-11-12
Drop deprecated FE "compute everything" behavior when nothing is pre-requested

jwpeterson:
This behavior (computing everything when the user hasn't explicitly requested anything) has been deprecated since 5f2cb9c405 (Mar 2020), which has been in every release series since 1.6.x.

What I have mainly observed from this deprecation period is that people continue to write code with missing FE pre-requests (since almost everyone except libmesh developers builds with LIBMESH_ENABLE_DEPRECATED). Thus, there is nothing pushing/reminding people to fix their missing pre-requests, and code continues to be written in the same way.

Therefore, I think the best thing to do going forward is to make this kind of behavior an error so that it can be fixed immediately rather than after being silently present (and computationally inefficient) for long periods of time. So, instead of a libmesh_assert(), this commit makes it into a libmesh_error_msg() which is active in optimized mode.

### DIFF
--- a/src/fe/inf_fe.C
+++ b/src/fe/inf_fe.C
@@ -330,56 +330,25 @@ void InfFE<Dim, T_radial, T_map>::determine_calculations()
 {
   this->calculations_started = true;
 
-  // If the user forgot to request anything, but we're enabling
-  // deprecated backwards compatibility, then we'll be safe and
-  // calculate everything.  If we haven't enable deprecated backwards
-  // compatibility then we'll scream and die.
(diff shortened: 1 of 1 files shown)


jwpeterson:
> I'd turn that error into an assert, but I'm okay with this as-is.

My thinking with using a libmesh_error_msg is that the extra cost of logical-ORing a bunch of booleans is worth it to get a straightforward error message rather than (I assume) a difficult to track down segfault...

jwpeterson:
Something seems to be wrong with the Coverage recipe here, so I'm just going to merge by hand.

PR #4335, opened 2025-11-20
Don't assume TypeVector T can use std math functions

lindsayad:
This is the case for MetaPhysicL types now

### DIFF
--- a/include/numerics/type_vector.h
+++ b/include/numerics/type_vector.h
@@ -961,9 +961,10 @@ template <typename T>
 auto
 TypeVector<T>::l1_norm() const
 {
-  decltype(std::abs(T())) ret{};
+  using std::abs;
+  decltype(abs(T())) ret{};
   for (const auto i : make_range(libmesh_dim))
-    ret += std::abs(_coords[i]);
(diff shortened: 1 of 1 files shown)


include/numerics/type_vector.h
```diff
@@ -1053,6 +1054,8 @@ T solid_angle(const TypeVector<T> & v01,
               const TypeVector<T> & v02,
               const TypeVector<T> & v03)
 {
+  using std::atan;
+
```
jwpeterson, line 1058: So what do the `using` statements do exactly? Just ensure that the `std::` version of the function takes part in the template argument deduction process?

lindsayad: I wouldn't exactly use that language, but yea they introduce the `std` implementations as viable candidate overloads

jwpeterson: Some of the `<cmath>` functions like [atan](https://en.cppreference.com/w/cpp/numeric/math/atan.html) are templated, so I think you should use exactly that language :rofl:

jwpeterson (approved):

PR #4337, opened 2025-11-21
Drop deprecated InfFE::n_dofs(), n_shape_functions() taking an ElemType

jwpeterson:
These functions have been marked libmesh_deprecated() since 1ec8fcd7 (Jun 2020), in release series 1.6.x and later.

This removal also triggers the removal of the FEInterface functions ifem_n_shape_functions() and ifem_n_dofs() since they called the version of InfFE::n_dofs() which is being removed. Technically, these FEInterface functions have not been marked libmesh_deprecated() for quite as long (a133c3d2, Feb 2025) but they have *effectively* been deprecated for just as long.

To avoid also dropping the deprecated FE APIs with the same names, we've simply placed libmesh_not_implemented() error messages in those deprecated functions for now to handle the infinite Elem case. This seems like a reasonable compromise to avoid dropping both the FE and InfFE deprecated functions at the same time.

### DIFF
--- a/include/fe/fe_interface.h
+++ b/include/fe/fe_interface.h
@@ -890,28 +890,8 @@ private:
    * the calls to \p FE and \p InfFE.
    */
 
-#ifdef LIBMESH_ENABLE_DEPRECATED
-  /**
-   * \deprecated Call the version of ifem_n_shape_functions() which
-   * takes a pointer-to-Elem instead.
-   */
-  static unsigned int ifem_n_shape_functions(const unsigned int dim,
-                                             const FEType & fe_t,
-                                             const ElemType t);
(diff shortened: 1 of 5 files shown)


jwpeterson:
I decided to split the removal of the deprecated `FE` and `InfFE` APIs into two parts, and I'm removing the deprecated `InfFE` functions first since I think they are being used much less widely (if at all). 

The deprecation dates here are somewhat inconsistent, for example sometimes we would deprecate a lower-level API, but forget to deprecate the higher-level APIs that called the lower-level ones until much later, even though they were "effectively" also deprecated the entire time. Everything removed in this branch was effectively deprecated in at least the 1.8.x series, and sometimes much earlier.

roystgnr (approved):

PR #4341, opened 2025-11-24
ExodusII_IO: if the mesh is renumbered then don't load solution vector data.

drwells:
Follow-up to #4340 - as a user, I'd rather set an additional flag then unknowingly load wrong nodal data.

### DIFF
--- a/tests/mesh/mesh_input.C
+++ b/tests/mesh/mesh_input.C
@@ -653,12 +653,12 @@ public:
 
     {
       MeshType mesh(*TestCommWorld);
+      mesh.allow_renumbering(false);
       IOType meshinput(mesh);
 
       // Avoid getting Nemesis solution values mixed up
       if (meshinput.is_parallel_format())
         {
-          mesh.allow_renumbering(false);
           mesh.skip_noncritical_partitioning(true);
         }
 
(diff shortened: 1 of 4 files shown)


jwpeterson (commented):
This seems like a good addition to me, and I think we could do something similar in `ExodusII_IO::copy_elemental_solution()`, which would have the same problem. I'm actually curious if any of our existing CI tests will hit this new error message.

As discussed in #4340, it would also be great to add a comment in the docs for this function, but I can tack that on later if you don't have the time.

drwells:
Sounds good - I'll handle it!

drwells:
Does this also apply to `copy_scalar_solution`?

jwpeterson:
> Does this also apply to `copy_scalar_solution`?

No, that one should not depend on the mesh reunumbering, it looks like we just match up the `SCALAR` variable names from libmesh to the ones in the Exodus file in that case.

jwpeterson (approved):

tests/mesh/mesh_input.C
```diff
@@ -653,12 +653,12 @@ public:
 
     {
       MeshType mesh(*TestCommWorld);
+      mesh.allow_renumbering(false);
```
drwells, line 656: I haven't worked on libMesh's test suite before - I think this will fix the failing CI.

jwpeterson: Yes, I think this case worked before because the exo file we read in was already numbered the way libmesh liked it.

drwells:
Thanks for the help with this issue!

PR #4346, opened 2025-12-05
Add TypeTensor check for Hermitian positive-definiteness

jwpeterson:
I ran into a situation where this was useful for error checking in the real-valued case. Since TypeTensor is also instantiated for `T=std::complex`, I made the check general enough to handle both the real- and complex-valued cases. I know that TypeTensor can also be used with MetaPhysicL types (?) but I made no attempt to support that.


jwpeterson:
> I'd prefer checking the smaller minors first, in the hopes that we can short circuit out before we get to the most expensive case, but I'm happy with this as-is.

Good suggestion, I will make that update.

PR #4404, opened 2026-02-19
Add some node-load-balancing test coverage

roystgnr:
I never got around to upgrading these tests to "rigorous testing of partitioner quality" rather than "testing to see if those code paths outright crash or corrupt partitioning", so this commit languished, but even a little test coverage here is better than nothing.


roystgnr:
Huh, maybe that's why I never opened up a PR for this - we need either a behavior change or more complicated scripting for the petscpartitioner option not to break `make check` with non-PETSc builds.

But on the other hand, that `--load-balanced-nodes-linear` failure on 4 and 5 processors looks real, so we really do need some test coverage of all these.

PR #4408, opened 2026-02-20
Add more integrity checks for tetrahedralization

roystgnr:
Hopefully this will catch whatever's going wrong on Guillaume's test case; I'm terrified of the possibility of having to diagnose a Netgen bug.

### DIFF
--- a/include/mesh/mesh_tet_interface.h
+++ b/include/mesh/mesh_tet_interface.h
@@ -131,6 +131,16 @@ protected:
   [[nodiscard]] std::set<SurfaceIntegrity> check_hull_integrity() const;
 
   /**
+   * This function checks the integrity of the current set of
+   * elements in the Mesh, and corrects what it can.
+   *
+   * \returns A set of SurfaceIntegrity codes from \p
+   * check_hull_integrity() if there are problems it can't fix, or an
+   * empty set otherwise.
+   */
+  [[nodiscard]] std::set<SurfaceIntegrity> improve_hull_integrity();
+
+  /**
    * This function prints an informative message and throws an
    * exception based on the output of the check_hull_integrity()
    * function.  It is a separate function so that you can check hull
--- a/include/mesh/mesh_tet_interface.h
+++ b/include/mesh/mesh_tet_interface.h
@@ -107,7 +107,7 @@ protected:
   /**
    * verbosity setting
    */
-  const unsigned int _verbosity;
+  unsigned int _verbosity;
 
   /**
    * Remove volume elements from the given mesh, after converting
@@ -199,6 +199,17 @@ protected:
    */
   std::unique_ptr<std::vector<std::unique_ptr<UnstructuredMesh>>> _holes;
 };
+
+
+// ------------------------------------------------------------
+// MeshTetInterface class member functions
+inline
+void
+MeshTetInterface::set_verbosity(unsigned int v)
+{
+  this->_verbosity = v;
+}
+
 
 } // namespace libMesh
 
--- a/include/mesh/mesh_tet_interface.h
+++ b/include/mesh/mesh_tet_interface.h
@@ -92,7 +92,22 @@ public:
    */
   virtual void triangulate () = 0;
 
+  /**
+   * Sets a verbosity level, defaulting to 0 (print nothing), to be
+   * set as high as 100 (print everything).
+   *
+   * For verbosity >= 50, print all detected surface mesh integrity
+   * issues as they're found.  Subclasses may add other output at
+   * other verbosity levels.
+   */
+  void set_verbosity(unsigned int v);
+
 protected:
+
+  /**
+   * verbosity setting
+   */
+  const unsigned int _verbosity;
 
   /**
    * Remove volume elements from the given mesh, after converting
--- a/include/mesh/mesh_tet_interface.h
+++ b/include/mesh/mesh_tet_interface.h
@@ -128,7 +128,9 @@ protected:
     MISSING_BACKLINK = 4,   // an element neighbor isn't linked back to it
     BAD_NEIGHBOR_NODES = 5, // an element neighbor isn't linked to expected nodes
     NON_ORIENTED = 6,       // an element neighbor has inconsistent orientation
-    BAD_NEIGHBOR_LINKS = 7  // an element neighbor has other inconsistent links
+    BAD_NEIGHBOR_LINKS = 7, // an element neighbor has other inconsistent links
+    DEGENERATE_ELEMENT = 8, // an element has zero area
+    DEGENERATE_MESH = 9     // the mesh clearly bounds zero volume
   };
 
   /**
--- a/include/mesh/mesh_tet_interface.h
+++ b/include/mesh/mesh_tet_interface.h
@@ -104,21 +104,31 @@ protected:
   static BoundingBox volume_to_surface_mesh (UnstructuredMesh & mesh);
 
   /**
+   * Enumeration of possible surface mesh integrity issues.
+   */
+  enum SurfaceIntegrity {
+    NON_TRI3 = 1,           // a non-TRI3 element is found
+    MISSING_NEIGHBOR = 2,   // an element with a nullptr-neighbor is found
+    EMPTY_MESH = 3,         // the mesh is empty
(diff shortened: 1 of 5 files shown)


roystgnr:
I'm putting this up to make sure it doesn't trigger any false positives in downstream CI and to give @GiudGiud a chance to play with it right away, but leaving it as "Draft" until I find time to add some unit tests for it.

roystgnr:
@GiudGiud reports that this finally was enough to catch the mixed orientation in his input.  I'm going to tack on some commits to *fix* those problems, not just to test the tests for them, but this should be ready to review/merge at any time; I could add later work to a different PR instead.

roystgnr:
Test coverage is good. That failure in the distributed sweeps wasn't a distributed-mesh failure, it was a general bug triggered by `DistributedMesh` post-repartitioning renumbering.

I think we're solid now, though.

src/mesh/mesh_tet_interface.C
```diff
@@ -335,65 +335,270 @@ BoundingBox MeshTetInterface::volume_to_surface_mesh(UnstructuredMesh & mesh)
...
+  Real best_abs_normal_0;
 
-  if (result != 0)
+  for (dof_id_type id : attached_elements)
     {
-      err_msg << "Error! Conforming Delaunay mesh tetrahedralization requires a convex hull." << std::endl;
+      Elem * elem = this->_mesh.elem_ptr(id);
+      const Point e01 = elem->point(1) - elem->point(0);
+      const Point e02 = elem->point(2) - elem->point(0);
+      const Point normal = e01.cross(e02).unit();
```
GiudGiud, line 523: could detect a sliver / 0 volume element there ?

src/mesh/mesh_tet_interface.C
```diff
@@ -335,65 +335,270 @@ BoundingBox MeshTetInterface::volume_to_surface_mesh(UnstructuredMesh & mesh)
...
+    }
+  if (result.count(NON_ORIENTED))
+    {
+      err_msg << "At least one triangle neighbor with an inconsistent orientation was found in the input boundary mesh.  ";
+      err_msg << "A constrained Delaunay tetrahedralization boundary must be an oriented Tri3 mesh." << std::endl;
+    }
+  if (result.count(BAD_NEIGHBOR_LINKS))
+    {
+      err_msg << "At least one triangle neighbor with inconsistent node and neighbor links was found in the input boundary mesh." << std::endl;
+    }
```
GiudGiud: This is actually a little hard to find once we get (only) that message. 
Can we output the issues as they are found?

include/mesh/mesh_tet_interface.h
```diff
@@ -103,30 +103,50 @@ class MeshTetInterface
...
+   * Enumeration of possible surface mesh integrity issues.
+   */
+  enum SurfaceIntegrity {
+    NON_TRI3 = 1,           // a non-TRI3 element is found
+    MISSING_NEIGHBOR = 2,   // an element with a nullptr-neighbor is found
+    EMPTY_MESH = 3,         // the mesh is empty
+    MISSING_BACKLINK = 4,   // an element neighbor isn't linked back to it
+    BAD_NEIGHBOR_NODES = 5, // an element neighbor isn't linked to expected nodes
+    NON_ORIENTED = 6,       // an element neighbor has inconsistent orientation
+    BAD_NEIGHBOR_LINKS = 7  // an element neighbor has other inconsistent links
```
GiudGiud: are these other intgrity issues covered by unit tests?

src/mesh/mesh_tet_interface.C
GiudGiud: actually better earlier in the first call to check_hull_integrity

src/mesh/mesh_tet_interface.C
```diff
@@ -335,65 +335,270 @@ BoundingBox MeshTetInterface::volume_to_surface_mesh(UnstructuredMesh & mesh)
...
+            }
+
+          if (!frontier_neigh && !finished_neigh)
+            frontier_elements.insert(neigh_id);
         }
 
-      if (result==3)
-        err_msg << "The input Mesh was empty!" << std::endl;
+      finished_elements.insert(elem_id);
+      frontier_elements.erase(elem_id);
```
GiudGiud, line 580: this only works if we have a single connected component. Do we have a check for that integrity issue?
(not asking for it now)

src/mesh/mesh_tet_interface.C
GiudGiud: even if libmesh is able to fix the issues, it's still good to fix the input mesh instead

src/mesh/mesh_tet_interface.C
roystgnr: > this only works if we have a single connected component.

Nah - unless the mesh is self-intersecting (and *that's* something I'd love to check for, but I don't know how to do it in O(N)), `best_elem` will have been part of the external component, so the external component will end up properly oriented.

> Do we have a check for that integrity issue?

No... which is partly by design.  If you supply an input mesh which defines a boundary with multiple connected components, we take the outermost as the actual boundary, so you don't have to worry about "filling in" any holes.

That's a less useful feature when the input is surface rather than volume elements, I admit, but we keep the same behavior there out of ~~laziness~~ consistency.

src/mesh/mesh_tet_interface.C
roystgnr: Adding it.

src/mesh/mesh_tet_interface.C
roystgnr: IMHO that one in particular we shouldn't; it's something that might be automatically fixed up, if the cause was just a poorly prepared input.

For the ones that are definitely unfixable input problems ... I don't want to output anything by default, honestly, but I'll add a `verbose` option.

include/mesh/mesh_tet_interface.h
roystgnr: `NON_ORIENTED` is covered; I figured that was the urgent one.

It's tricky to get the others to trigger realistically.  My first thought was to refine an element, but that just gets caught by the `libmesh_not_implemented_msg` in `MeshModification::all_tri()`.

src/mesh/mesh_tet_interface.C
GiudGiud: verbose sounds great!

include/mesh/mesh_tet_interface.h
```diff
@@ -92,8 +92,23 @@ class MeshTetInterface
...
 
+  /**
+   * Sets a verbosity level, defaulting to 0 (print nothing), to be
+   * set as high as 100 (print everything).
+   *
+   * For verbosity >= 50, print all detected surface mesh integrity
+   * issues as they're found.  Subclasses may add other output at
+   * other verbosity levels.
+   */
+  void set_verbosity(unsigned int v);
```
GiudGiud, line 103: missing definition?

src/mesh/mesh_tet_interface.C
```diff
@@ -335,65 +335,305 @@ BoundingBox MeshTetInterface::volume_to_surface_mesh(UnstructuredMesh & mesh)
...
+    {
+      err_msg << "At least one triangle neighbor without expected node links was found in the input boundary mesh.  ";
+      err_msg << "A constrained Delaunay tetrahedralization boundary must be a conforming and non-adaptively-refined mesh." << std::endl;
+    }
+  if (result.count(NON_ORIENTED))
+    {
+      err_msg << "At least one triangle neighbor with an inconsistent orientation was found in the input boundary mesh.  ";
+      err_msg << "A constrained Delaunay tetrahedralization boundary must be an oriented Tri3 mesh." << std::endl;
+    }
+  if (result.count(BAD_NEIGHBOR_LINKS))
```
GiudGiud, line 631: we are missing the messages for degenerate meshes and elements?

src/mesh/mesh_tet_interface.C
GiudGiud: works for me

GiudGiud (approved):

src/mesh/mesh_tet_interface.C
roystgnr: Added

PR #4411, opened 2026-02-25
Mesh preparation assertions

roystgnr:
Now that we're happy with the updated libMesh submodule in MOOSE and we can get access there to the new fine-grained mesh preparation APIs, it's probably time to start (at least in `--disable-deprecated` builds) testing for preparation status.

This is the last swath of libMesh-level work that came out of #3759

### DIFF
--- a/examples/miscellaneous/miscellaneous_ex15/miscellaneous_ex15.C
+++ b/examples/miscellaneous/miscellaneous_ex15/miscellaneous_ex15.C
@@ -221,6 +221,10 @@ int main (int argc, char** argv)
          sync_obj);
     }
 
+  // We'll need to update caches of subdomain ids too, to keep the
+  // mesh prepared.
+  mesh.cache_elem_data();
+
   // Now we set the sources of the field: prism-shaped objects that are
   // determined here by containing certain points:
   auto p_pt_lctr = mesh.sub_point_locator();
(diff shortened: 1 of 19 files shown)


roystgnr:
Everything worked on the first try, both in CI and on my workstation.  That makes me suspicious; I'm turning on more recipes.

roystgnr:
Well, there's at least a performance concern.  With the extra testing, `Debug No Deprecated` goes from a 90 minute recipe to a "300 minute and it's just now getting past adaptivity_ex2" recipe.  This would be really useful downstream but not if it's likely to start tests there timing out...

roystgnr:
No, not a performance concern after all.  After scanning a zillion perf log reports trying to figure out why the new assertions might take so much longer than expected, I just kicked the CI recipe, which finished in 45 minutes this time.  I suppose it's possible there's an intermittent hang that triggered in one example on that first run, but I've never seen anything like that in my own testing.  We should be good to merge.

roystgnr:
I just realized that those remaining MOOSE test failures were false positives.  This should be ready to merge ... probably right after the current MOOSE libMesh submodule update, to avoid throwing any wrenches into that.  In the meantime I'll see if there are more chunks of this that could be separate PRs.

roystgnr:
Rebasing to strip out the already-merged commits and to retry CI with the MOOSE multiapp fixes in.

PR #4412, opened 2026-02-26
Allow integrating hypre AMS/ADS in FSPs

nmnobre:
See also https://gitlab.com/petsc/petsc/-/merge_requests/9077.

### DIFF
--- a/src/numerics/petsc_preconditioner.C
+++ b/src/numerics/petsc_preconditioner.C
@@ -183,10 +183,6 @@ void PetscPreconditioner<T>::set_petsc_aux_data(PC & pc, System & sys, const uns
     // If not ams/ads, we quit with nothing to do
     if (std::string(hypre_type) == "ams")
     {
-      // If multiple variables, we error out as senseless
-      libmesh_error_msg_if(sys.n_vars() > 1,
-                           "Error applying hypre AMS to a system with multiple "
-                           "variables");
       // If not a 1st order Nédélec or a 2d 1st order Raviart-Thomas system, we
       // error out as we do not support anything else at the moment
       libmesh_error_msg_if(sys.variable(v).type() != FEType(1, NEDELEC_ONE) &&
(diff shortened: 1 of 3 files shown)


lindsayad:
Is this going to rely on the libMesh DM?

nmnobre:
I was using `petsc_auto_fieldsplit()` with `--solver-variable-names` to test this (is that what you meant?).
The current implementation in `set_petsc_aux_data()` only requires a variable/split index (so there must be a one-to-one correspondence there between no. of vars and splits), but I don't think there were any particular assumptions on who/what does the splitting, as long as PETSc is informed about it somehow.

In the discussion I linked above we agreed that the proper way to do this was probably to provide a callback that PETSc would call just-in-time whenever it needs info to build the preconditioner with sufficient information passed to the client application (i.e. libMesh) so it can disambiguate what variable/split needs preconditioning. That would then lift the requirement for only a single-level of field-split with one-to-one variables and splits (like I have here at present).

lindsayad:
I probably don't fully understand the situation here but in MOOSE we have examples of 
- nested field splits
- one split with multiple variables
- multiple splits of one variable (contact-interface dofs and non-contact-interface dofs for example)

I'll note that it's been a long-standing goal of mine to unify the libMesh and MOOSE DMs. I think each has features that the other does not. The MOOSE DM, for instance, doesn't have geometric multigrid support

nmnobre:
I know we do, and I know as it stands this implementation is limiting (thus the condition to return early if the assumption doesn't hold).

> I probably don't fully understand the situation here

I think the problem is _I_ don't fully understand the situation. That's (partially) why this is on hold. I don't really know how PETSc DM works.

> I'll note that it's been a long-standing goal of mine to unify the libMesh and MOOSE DMs

Yes(!), I never understood why some of the functionality was reimplemented, and to be frank I always had doubts as to whether they're actually playing perfectly nice with each other. I remember lamenting the situation with @roystgnr, but I just can't find that conversation anywhere, might've dreamt it.

PR #4417, opened 2026-03-06
Use cached data for n_subdomains()

roystgnr:
Thanks and apologies to @NamjaeChoi for pointing out that this was O(N) when he was expecting O(1).


NamjaeChoi:
Thank you for the fix!

roystgnr:
The issues in the MOOSE side here overlap somewhat with issues I'm working on in another MOOSE branch.  I'll try to fix these there too.

PR #4433, opened 2026-04-10
Avoid unused, uninitialized and truncation warnings

nmnobre:
Plus libMesh/TIMPI#160.

### DIFF
--- a/include/base/getpot.h
+++ b/include/base/getpot.h
@@ -1712,8 +1712,9 @@ GetPot::_internal_managed_copy(const std::string& Arg) const
     return it->get();
 
   // Otherwise, create a new one
-  const std::size_t bufsize = strlen(arg)+1;
-  auto newcopy = std::make_unique<char[]>(bufsize);
+  auto newcopy = std::make_unique<char[]>(strlen(arg)+1);
+  // Note: strcpy() is safe here because we allocated enough space and
+  // "arg" is guaranteed to be null-terminated.
   strcpy(newcopy.get(), arg);
   auto pr = _internal_string_container.insert(std::move(newcopy));
   return pr.first->get();
--- a/include/base/getpot.h
+++ b/include/base/getpot.h
@@ -1714,7 +1714,7 @@ GetPot::_internal_managed_copy(const std::string& Arg) const
   // Otherwise, create a new one
   const std::size_t bufsize = strlen(arg)+1;
   auto newcopy = std::make_unique<char[]>(bufsize);
-  strncpy(newcopy.get(), arg, bufsize);
+  strcpy(newcopy.get(), arg);
   auto pr = _internal_string_container.insert(std::move(newcopy));
   return pr.first->get();
 }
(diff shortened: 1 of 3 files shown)


include/base/getpot.h
```diff
@@ -1714,7 +1714,7 @@ GetPot::_internal_managed_copy(const std::string& Arg) const
   // Otherwise, create a new one
   const std::size_t bufsize = strlen(arg)+1;
   auto newcopy = std::make_unique<char[]>(bufsize);
-  strncpy(newcopy.get(), arg, bufsize);
+  strcpy(newcopy.get(), arg);
```
jwpeterson, line 1718: What was the warning here exactly? This isn't our code originally, so I'm not that familiar with it, but maybe the compiler is worried about the possibility that we won't end up with a null-terminated string in the case where there are no null characters in the first `bufsize` characters of `arg`? I'm wondering if explicitly setting
```
newcopy[bufsize-1] = '\0';
```
would therefore also satisfy the compiler...

roystgnr: Don't we already get `newcopy[bufsize-1] = '\0'` via `make_unique`?  `make_unique` for array types with unknown bound is supposed to value-initialize, which is supposed to value-initialize each array member, which is supposed to zero-initialize if the member type is `char`.

jwpeterson: OK, good point. I think it's safe to say that the warning, whatever it is, is probably overzealous and we should just silence it whatever way is easiest...

nmnobre: `c_str()` is guaranteed to return a null-terminated string. The problem is rather that `strncpy` is being used unnecessarily, and the compiler flags this:
```
In file included from ../src/reduced_basis/rb_construction.C:36:
./include/libmesh/getpot.h: In function 'const char* GetPot::_internal_managed_copy(const std::string&) const':
./include/libmesh/getpot.h:1717:10: warning: 'char* strncpy(char*, const char*, size_t)' specified bound depends on the length of the source argument [-Wstringop-truncation]
 1717 |   strncpy(newcopy.get(), arg, bufsize);
      |   ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./include/libmesh/getpot.h:1715:37: note: length computed here
 1715 |   const std::size_t bufsize = strlen(arg)+1;
      |                               ~~~~~~^~~~~
```

jwpeterson: > specified bound depends on the length of the source argument

OK, so the compiler thinks that we are using the length of the "source" string in the call to strncpy, a common bug with strncpy, so it is attempting to warn us about that. It seems to miss the fact that the destination string is also the same length...

I think I'm happy with your proposed change, but I'm going to push a comment explaining why the use of `strcpy()` is safe here (and terser than C++ solutions based on std::string::copy or std::copy which both require manual null-termination).

jwpeterson (approved):

PR #4441, opened 2026-04-22
Kokkos support for Lagrange and Monomial elements 

rochi00:
1. Shape functions — nativeShape, nativeGradShape for LAGRANGE (12 types) + MONOMIAL (orders 0-5)                                                                       
  2. Type dispatch — FEShapeKey, classFromTopology, nDofs, templated nativeMapShape                                                                                         
  3. Quadrature — GaussQuadrature device-callable for EDGE, QUAD, HEX, TRI, TET + fillQuadrature host helper                                                               
  4. Physical map — physicalPoint, jacobian, physicalPointAndJacobian, faceJacobian, faceJxW, faceNormal (templated + runtime versions)                                     
  5. Device-safe asserts — printf + Kokkos::abort() for debug builds on GPU


include/base/libmesh_device.h
```diff
@@ -0,0 +1,49 @@
...
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+#pragma once
```
roystgnr: In practice I think literally every compiler we care about supports this (though Wiki lists one I've never heard of that doesn't!), but since it never made it into a C++ standard we've always used include guards instead; let's do that here too.

include/base/libmesh_exceptions.h
```diff
@@ -212,7 +212,13 @@ class TerminationException
 #ifdef LIBMESH_ENABLE_EXCEPTIONS
 #define libmesh_noexcept noexcept
 
+#if defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__)
+// GPU device code does not support C++ exceptions — silently no-op.
```
roystgnr: Mostly when we throw an exception it's because we can't sanely go on; even if we don't support exceptions on device I don't think silently going on is the right way to handle that.  When we don't support exceptions on the host we print the error message and abort; should we not do that here too?

include/enums/enum_fe_elem_class.h
```diff
@@ -0,0 +1,50 @@
...
+
+
+
+#ifndef LIBMESH_ENUM_FE_ELEM_CLASS_H
+#define LIBMESH_ENUM_FE_ELEM_CLASS_H
+
+namespace libMesh {
+
+/**
+ * \enum libMesh::FEElemClass groups element types by topological class,
```
roystgnr, line 26: Concept is good, but I don't like the name.  FEFamily really is FE Method specific; an Elem is an Elem even if we're doing Finite Volume or pure geometry/meshing work (or in the future FDM stencils or whatever) instead.  Let's just call this `ElemClass`, unless someone has an idea for a better name?

include/gpu/kokkos_fe_base.h
```diff
@@ -0,0 +1,44 @@
...
+//   Hex:   (xi, eta, zeta)    in [-1,1]^3
+//   Tri:   (xi, eta)          in unit triangle, xi >= 0, eta >= 0, xi+eta <= 1
+//   Tet:   (xi, eta, zeta)    in unit tetrahedron
+//
+// Unused coordinate arguments (e.g. zeta on a 2D element) are accepted but
+// ignored, so call sites can always pass all three without special-casing.
+//
+// Order is only meaningful for MONOMIAL specializations.
+// Lagrange specializations always use Order = 0 (the default) because the
+// ElemType (e.g. QUAD9) already encodes the polynomial order.
```
roystgnr: No it doesn't.  p=1 Lagrange on a Quad9 makes perfect sense and is commonly used for the pressure variable in incompressible flow simulation.

I get that we're not supporting everything right out of the gate, but we need to make sure that we're not digging ourselves into any holes in our documentation or (more importantly, since it'll be much harder to change later) our APIs.  Is there something in the design that would already prevent us from returning evaluations of a p=1 Lagrange on a higher-order geometric element?

include/gpu/kokkos_fe_evaluator.h
```diff
@@ -0,0 +1,396 @@
...
+#include "libmesh/enum_elem_type.h"
+#include "libmesh/enum_fe_family.h"
+
+namespace libMesh::Kokkos
+{
+
+// ── On-device helpers: element class -> spatial dimension ─────────────────────
+
+LIBMESH_DEVICE_INLINE unsigned int
+dimFromClass(FEElemClass cls)
```
roystgnr: I'd like to use snake_case rather than camelCase, like we do everywhere else in libMesh code.

We should also try to get this to resemble existing code in less superficial ways if possible.  We've got `static const unsigned int Elem::type_to_dim_map[]` already, for instance; could we either add an `Elem::class_to_dim_map[]` too, or if elem.h would cause Kokkos fits, would it make sense for us to still use the same array-based design to make things easier to merge later?

include/gpu/kokkos_fe_evaluator.h
```diff
@@ -0,0 +1,396 @@
...
+      return 3;
+    default:
+      return 0;
+  }
+}
+
+// ── On-device helper: (class, order) -> canonical Lagrange topology ────────────
+
+LIBMESH_DEVICE_INLINE libMesh::ElemType
+lagrangeTopologyForClassAndOrder(FEElemClass cls, unsigned int order)
```
roystgnr: Same snake_case question, but also ... is this a "digging ourselves into hole" case?  If someone wants to use a Quad8 later, are they just going to be forever out of luck?

It seems like maybe we're starting with all the data we need (elem type, order), then throwing away some of that data to get to (elem class, order), then trying to reconstruct (elem type, order) again, but couldn't we just hang on to (elem type, order) from the start?

include/gpu/kokkos_fe_evaluator.h
```diff
@@ -0,0 +1,396 @@
+// Kokkos on-device FE shape function dispatch (fe_evaluator.h).
+//
+// Provides:
+//   nativeMapShape    — isoparametric Lagrange shape (topology-based)
```
roystgnr: We parameterize on FEFamily, right?  So the API is flexible enough, but let's still not hard-code "Lagrange" in the documentation.

include/gpu/kokkos_fe_types.h
```diff
@@ -0,0 +1,280 @@
...
+#endif
+
+namespace libMesh::Kokkos
+{
+
+// Bring FEElemClass into this namespace so existing unqualified uses compile.
+using libMesh::FEElemClass;
+
+// ── Shape function space key ──────────────────────────────────────────────────
+// Uniquely identifies a physics FE space independent of the geometric topology.
```
roystgnr: No, it doesn't.  LAGRANGE, QUAD, 2 could mean either QUAD8 or QUAD9, ad those are two different bases.

include/gpu/kokkos_fe_types.h
```diff
@@ -0,0 +1,280 @@
...
+
+/// Return the ElemType of any side of parent element type \p parent.
+/// For all standard element types the side topology is uniform across sides,
+/// so the side index is not needed.
+LIBMESH_DEVICE_INLINE libMesh::ElemType
+getSideTopology(libMesh::ElemType parent)
+{
+  switch (parent)
+  {
+    // 1D: sides are vertex nodes, represented as EDGE2 (degenerate)
```
roystgnr: They shouldn't be.  A node side is a NODEELEM.

include/gpu/kokkos_fe_types.h
```diff
@@ -0,0 +1,280 @@
...
+struct FEShapeKey
+{
+  libMesh::FEFamily   family;
+  libMesh::FEElemClass cls;
+  unsigned int         order;
+};
+
+// ── Device-callable conversion helpers ───────────────────────────────────────
+
+/// Return the ElemType of any side of parent element type \p parent.
```
roystgnr: No it doesn't.  You're returning TRI3 for PYRAMID6, for example, but that's not "any side", that's 2 out of 5 sides.

I'm not sure how getSideTopologyOrMaybeNot is useful, but if it is useful we need to make sure the name and the docs match the behavior.

include/gpu/kokkos_fe_types.h
```diff
@@ -0,0 +1,280 @@
...
+            libmesh_error_msg("nDofs(FEShapeKey): unsupported LAGRANGE order "
+                              << key.order << " for TRI");
+#endif
+            return 0;
+        }
+      case libMesh::FEElemClass::QUAD:
+        switch (key.order)
+        {
+          case 1: return 4;
+          case 2: return 9;
```
roystgnr: This is wrong for QUAD8

include/gpu/kokkos_fe_types.h
```diff
@@ -0,0 +1,280 @@
...
+            libmesh_error_msg("nDofs(FEShapeKey): unsupported LAGRANGE order "
+                              << key.order << " for TET");
+#endif
+            return 0;
+        }
+      case libMesh::FEElemClass::HEX:
+        switch (key.order)
+        {
+          case 1: return 8;
+          case 2: return 27;
```
roystgnr: This is wrong for HEX20.

include/gpu/kokkos_fe_types.h
```diff
@@ -0,0 +1,280 @@
...
+            libmesh_error_msg("nDofs(FEShapeKey): unsupported LAGRANGE order "
+                              << key.order << " for HEX");
+#endif
+            return 0;
+        }
+      case libMesh::FEElemClass::PRISM:
+        switch (key.order)
+        {
+          case 1: return 6;
+          case 2: return 21;
```
roystgnr: This is wrong in every case!  For PRISM18, PRISM20, and PRISM21, it should be 18; for PRISM15 15.

include/gpu/kokkos_fe_types.h
```diff
@@ -0,0 +1,280 @@
...
+            libmesh_error_msg("nDofs(FEShapeKey): unsupported LAGRANGE order "
+                              << key.order << " for PRISM");
+#endif
+            return 0;
+        }
+      case libMesh::FEElemClass::PYRAMID:
+        switch (key.order)
+        {
+          case 1: return 5;
+          case 2: return 14;
```
roystgnr: Should be 13 for PYRAMID13.

include/libmesh/rebuild_makefile.sh
```diff
@@ -2,7 +2,7 @@
 
 built_sources=""
 
-headers=`find .. -name "*.h" -a -not -name libmesh_config.h -type f | LC_COLLATE=POSIX sort`
+headers=`find .. -name "*.h" -a -not -name libmesh_config.h -a -not -path "../kokkos/*" -type f | LC_COLLATE=POSIX sort`
```
roystgnr: Why aren't we making symlinks for Kokkos headers?  Are we not installing them to a libmesh/ subdirectory?  If not, where are we installing them?  If so, don't we need to match them in the build directory?

include/gpu/kokkos_fe_evaluator.h
```diff
@@ -0,0 +1,396 @@
...
+        case libMesh::TET10:
+          return FEEvaluator<libMesh::LAGRANGE, libMesh::TET10>::shape(i, xi, eta, zeta);
+        case libMesh::HEX8:
+          return FEEvaluator<libMesh::LAGRANGE, libMesh::HEX8>::shape(i, xi, eta, zeta);
+        case libMesh::HEX20:
+          return FEEvaluator<libMesh::LAGRANGE, libMesh::HEX20>::shape(i, xi, eta, zeta);
+        case libMesh::HEX27:
+          return FEEvaluator<libMesh::LAGRANGE, libMesh::HEX27>::shape(i, xi, eta, zeta);
+        default:
+          return Real(0);
```
roystgnr: We really need to be printing an error message and aborting in every case like this, not silently returning incorrect results, at least in dbg/devel builds if for some reason it's too expensive to be safe on the GPU in opt builds.

include/gpu/kokkos_fe_evaluator.h
```diff
@@ -0,0 +1,396 @@
...
+nativeGradMapShape(unsigned int i, Real xi, Real eta, Real zeta)
+{
+  return FEEvaluator<family, topo>::grad_shape(i, xi, eta, zeta);
+}
+
+// ── Runtime topology versions (larger GPU stack usage) ───────────────────
+
+/// Evaluate the i-th geometric map shape function at (xi, eta, zeta).
+LIBMESH_DEVICE_INLINE Real
+nativeMapShape(libMesh::ElemMappingType mapping_type,
```
roystgnr: This sure feels like we're just rewriting a second copy of FEInterface.  I suppose porting that to GPU right off the bat would be too hard, but could we try to match the name conventions?  Maybe even some of the calling conventions - if we've got type_vector.h GPU-compatible now, why are we passing individual xi, eta, zeta components here rather than a single xi vector?

include/gpu/kokkos_fe_evaluator.h
roystgnr: Also the usual weDontUseCamelCase complaints, and:

What does "native" mean in this context?  Is there a foreignMapShape we'll be introducing later?

include/gpu/kokkos_fe_face_map.h
```diff
@@ -0,0 +1,70 @@
+//* This file is part of the libMesh library.
+//* https://www.libmesh.net
```
roystgnr: This URL doesn't exist.

include/gpu/kokkos_fe_face_map.h
```diff
@@ -0,0 +1,70 @@
...
+//*
+//* All rights reserved, see COPYRIGHT for full restrictions
```
roystgnr: This file doesn't exist.

The idea of a code review is to take the number of people who have read the code from 1 to at least 2; would it be accurate for me to conclude here that I'm actually taking that number from 0 to 1?

Let's get proper copyright headers on all the new files here, and say a little prayer that when AI gets control of the nukes too we're not all gonna die.

include/gpu/kokkos_scalar_types.h
```diff
@@ -0,0 +1,202 @@
...
+
+using Real = libMesh::Real;
+
+// Forward declaration so Real3::cartesian_product can return Real33.
+struct Real33;
+
+// ---------------------------------------------------------------------------
+// Real3 — GPU-safe 3-component real vector
+// ---------------------------------------------------------------------------
+struct Real3 : public libMesh::VectorValue<Real>
```
roystgnr: This has gotten to be a very thin shim.  Which is good, but can we go no farther?  Why not just use `libMesh::RealVectorValue`, call `a.cross(b)` instead of `a.cross_product(b)`, `a * b` instead of `a.dot_product(b)`, and `a.outer_product(b)` instead of `a.cartesian_product(b)`?

"Cartesian product" in particular is really weird terminology; I suppose the index set of the outer product is the cartesian product of the index sets of the vectors, but that doesn't mean the outer product is a cartesian product.

include/gpu/kokkos_scalar_types.h
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  // Outer product: returns a Real33 (defined below after Real33 is complete).
+  LIBMESH_DEVICE_INLINE Real33 cartesian_product(const Real3 & v) const;
+};
+
+// ---------------------------------------------------------------------------
+// Real33 — GPU-safe 3x3 real tensor
+// ---------------------------------------------------------------------------
+struct Real33 : public libMesh::TensorValue<Real>
```
roystgnr: Any reason you're inheriting from TensorValue here and TypeVector above, rather than TypeTensor and TypeVector or TensorValue and VectorValue?

include/gpu/kokkos_scalar_types.h
```diff
@@ -0,0 +1,202 @@
...
+        (*this)(i, j) = scalar;
+  }
+
+  // Construct from base TypeTensor.
+  LIBMESH_DEVICE_INLINE Real33(const libMesh::TypeTensor<Real> & t) : Base(t) {}
+
+  // ----------- Legacy API kept for backward compatibility ------------------
+
+  // Set to dim×dim identity (remaining entries zero).
+  LIBMESH_DEVICE_INLINE void identity(const unsigned int dim = 3)
```
roystgnr: This doesn't return an identity, so I'd call it `set_identity`, or make it a static member that does return an identity.

Either way, could we move it up into TypeTensor?  Seems it would be useful for others.

include/gpu/kokkos_scalar_types.h
```diff
@@ -0,0 +1,202 @@
...
+  {
+    this->zero();
+    for (unsigned int i = 0; i < dim; ++i)
+      (*this)(i, i) = Real(1);
+  }
+
+  // Determinant of the leading dim×dim sub-block.
+  LIBMESH_DEVICE_INLINE Real determinant(const unsigned int dim = 3) const
```
roystgnr: I like this optimization, but let's put it in TypeTensor

include/gpu/kokkos_scalar_types.h
```diff
@@ -0,0 +1,202 @@
...
+    if (dim == 0) return Real(1);
+    if (dim == 1) return (*this)(0, 0);
+    if (dim == 2)
+      return (*this)(0,0) * (*this)(1,1) - (*this)(0,1) * (*this)(1,0);
+    // dim == 3: delegate to TypeTensor::det()
+    return this->det();
+  }
+
+  // Inverse of the leading dim×dim sub-block.
+  LIBMESH_DEVICE_INLINE Real33 inverse(const unsigned int dim = 3) const
```
roystgnr: Likewise I like this; we could move it up too.

We should either give them both new names (to avoid having to add if/else to the existing case) or merge them both with the existing functions (for simplicity), though; giving one a new name and the other the same as the existing functions is just weird.

include/gpu/kokkos_scalar_types.h
```diff
@@ -0,0 +1,202 @@
...
+  Real33 tensor;
+  for (unsigned int i = 0; i < 3; ++i)
+    for (unsigned int j = 0; j < 3; ++j)
+      tensor(i, j) = (*this)(i) * v(j);
+  return tensor;
+}
+
+// ---------------------------------------------------------------------------
+// Free operators not covered by TypeVector/TypeTensor arithmetic
+// (scalar±vector forms that don't exist in the base classes)
```
roystgnr: They don't, because that's kind of horrifying for type safety.  Even if they did we'd probably decide that the proper projection of s from ℝ→ℝ³ was (s,0,0), not (s,s,s).  That's not even of magnitude s!

Where do we use these?  If the answer is "nowhere" let's delete them; if not let's give them new names so I can grep for those names and stare at the hits suspiciously.

include/libmesh/rebuild_makefile.sh
roystgnr: We're installing to libmesh/gpu/; I'm okay with the extra complication there, I guess, but then why won't we have trouble in the build directory when libmesh/gpu/ doesn't exist yet?

include/gpu/kokkos_fe_face_map.h
```diff
@@ -0,0 +1,70 @@
...
+  // 1-D elements: the "side" is a single vertex node.  There are no reference
+  // coordinates on a point element, so the parent reference coordinate is
+  // simply the node's own position in the parent reference element.
+  if (n == 1)
+  {
+    const auto & pt = side_in_parent.point(0);
+    parent_pt(0) = pt(0);
+    parent_pt(1) = pt(1);
+    parent_pt(2) = pt(2);
+    return parent_pt;
```
roystgnr: Wait, what?

There are reference coordinates on a point element; they're all just (0,0,0).

But (x,y,z) is what `point(0)` will give you, and that is *not* (xi,eta,zeta).  point(0) on a NodeElem will be a point in physical space.

include/gpu/kokkos_fe_face_map.h
```diff
@@ -0,0 +1,70 @@
...
+  {
+    const Real s = face_qpt(0);
+    const Real t = face_qpt(1);
+    const Real psi = nativeMapShape(mapping_type, side_topo, k, s, t, 0.0);
+
+    const auto & pt = side_in_parent.point(k);
+    parent_pt(0) += psi * pt(0);
+    parent_pt(1) += psi * pt(1);
+    parent_pt(2) += psi * pt(2);
+  }
```
roystgnr, line 126: Okay, it's midnight, and I'm getting tired of tearing my hair out, and I think I've already written enough to keep you going for a while, so I'm going to post this review half done, with one final issue:

Is this whole block not just wrong?  It gives the location in physical space of `face_qpt` in side-reference-space, but the docs here say we're returning a point in parent-reference space.  There has to be an `inverse_map()` afterward to get back to that.

Please tell me I'm wrong, and I just lose too many IQ points by midnight?  We clearly need *way* more test coverage if I'm somehow our first line of defense for a bug this big.  I know we don't have the environments set up to test in PR yet, but ideally you're running `make check` yourself and at some point in the new tests we're doing a side integral?

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "contrib/metaphysicl"]
 	path = contrib/metaphysicl
-	url = ../../libMesh/MetaPhysicL
+	url = ../../rochi00/MetaPhysicL
```
roystgnr: No.  Definitely not.

Makefile.am
```diff
@@ -257,6 +257,12 @@ SUBDIRS += contrib
 
 include src/libmesh_SOURCES
 
+# When Kokkos support is enabled, add the Kokkos FE type-conversion sources.
+# These are kept in src/kokkos/ and not in src/libmesh_SOURCES (which is
+# auto-generated) so that they don't affect non-Kokkos builds.
+if LIBMESH_ENABLE_KOKKOS
+endif
```
roystgnr: Actually, one more comment.

There's an interesting story about the rock band Van Halen:

When they were on tour, they had a fairly complicated and safety-critical set of contract stipulations that venues had to meet: heavy platforms and heavy equipment, serious electrical requirements, pyrotechnics, etc.  And they had a little contract clause that demanded a bowl of M&Ms for them backstage, with all the brown M&Ms removed.

If someone challenged that clause, they waived it.  They didn't care about not eating brown M&Ms.  They just cared that the big long critical pile of contract text had actually been respected, entirely read by the people with the primary responsibility of reading it and making sure everything was set up correctly.  If they hadn't been reading carefully enough to notice the brown-M&M-hating-divas bit, could they really be trusted to be getting the subtle, tricky mechanical and electrical and thermal stuff done and inspected safely?  No; the band roadies would essentially have to reinspect everything to see which parts of it they'd have to redo.

It was not a good idea to be a venue that made the hard rock band and their annoyed roadies unable to trust any of your work.

roystgnr (changes requested):

Makefile.am
rochi00: Sorry, I mostly paid attention to the tests that were written, and once they all passed assumed the internal code was correct. That was clearly not a good idea. I will review everything now. The tests clearly didn't have enough coverage

rochi00: The claude generated comments had so many issues, it is terrifying how much it overclaims

PR #4442, opened 2026-04-23
MeshTools::all_rbb() to generate IGA rational quadratic meshes

roystgnr:
This needs proper test coverage before merging but I've been letting it sit in a separate branch for too long and I want to at least make sure it's not interfering with any CI builds.

### DIFF
--- a/src/mesh/mesh_modification.C
+++ b/src/mesh/mesh_modification.C
@@ -1459,7 +1459,7 @@ void MeshTools::Modification::all_rbb (MeshBase & mesh)
   // mapped node
   const Real default_weight = 1.0;
 
-  weight_index = cast_int<unsigned char>
+  const auto weight_index =
     (mesh.add_node_datum<Real>("rational_weight", true,
                                &default_weight));
 
@@ -1467,7 +1467,12 @@ void MeshTools::Modification::all_rbb (MeshBase & mesh)
   mesh.set_default_mapping_data(weight_index);
 
   // Out of loop to reduce heap allocations
-  std::unique_ptr<Elem> edge, face;
+  std::unique_ptr<Elem> edge_ptr, face_ptr;
+
+  // Utility for checking extrusion directions later
+  auto almost_equal = [](Real a, Real b) {
+    return (std::abs(a-b) < TOLERANCE*TOLERANCE);
+  };
 
   for (auto & elem : mesh.element_ptr_range())
     {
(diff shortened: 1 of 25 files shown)


tests/mesh/all_rbb.C
```diff
@@ -0,0 +1,272 @@
...
+using namespace libMesh;
+
+class AllRBBTest : public CppUnit::TestCase
+{
+  /**
+   * The goal of this test is to verify proper operation of the Mesh Extruder
+   * with the optional object callback for setting custom subdomain IDs.
+   * We pass a custom object for generating subdomains based on the old element
+   * ID and the current layer and assert the proper values.
+   */
```
GiudGiud, line 26:
```suggestion
```
or update

tests/mesh/all_rbb.C
```diff
@@ -0,0 +1,272 @@
...
+      CPPUNIT_ASSERT_EQUAL(elem->mapping_type(), RATIONAL_BERNSTEIN_MAP);
+      CPPUNIT_ASSERT(elem->has_affine_map());
+    }
+
+    for (auto & node : mesh.node_ptr_range())
+    {
+      const Real w = node->get_extra_datum<Real>(weight_index);
+
+      CPPUNIT_ASSERT_EQUAL(Real(1), w);
+    }
```
GiudGiud, line 131: volume should be easy to check too
as well as hmin/hmax ? these change with the mapping right

tests/mesh/all_rbb.C
```diff
@@ -0,0 +1,272 @@
...
+
+    const Real radius = 1;
+    const Real circumference = 2 * pi * radius;
+    const Real tol = TOLERANCE*TOLERANCE;
+
+    // Build a filled circle
+    MeshTools::Generation::build_sphere (interior_mesh, radius,
+                                         n_refinements, QUAD9);
+
+    // Get just the outer EDGE9 circle mesh
```
GiudGiud:
```suggestion
    // Get just the outer EDGE3 circle mesh
```

src/mesh/mesh_modification.C
```diff
@@ -1452,6 +1452,336 @@ void MeshTools::Modification::all_tri (MeshBase & mesh)
...
+        if (face.type() == TRI6)
+          return;
+
+        if (face.type() != QUAD9)
+          libmesh_not_implemented_msg
+            ("all_rbb() currently only supports mid-face nodes on Quad9 faces");
+
+        // We only use [4,8) but matching indices is nice and stack is
+        // cheap.
+        // that we're only using what we set, though.
```
GiudGiud: ?

src/mesh/mesh_modification.C
```diff
@@ -1452,6 +1452,336 @@ void MeshTools::Modification::all_tri (MeshBase & mesh)
...
+                  almost_equal(w[12], w[13]) &&
+                  almost_equal(w[12], w[21]) &&
+                  almost_equal(w[14], w[15]) &&
+                  almost_equal(w[14], w[23]) &&
+                  almost_equal(w[17], w[19]) &&
+                  almost_equal(w[17], w[25]) &&
+                  almost_equal(w[22], w[24]))
+                extrude_dir = 0;
+
+              if (almost_equal(w[0], w[3]) &&
```
GiudGiud: else?

GiudGiud: elif

src/mesh/mesh_modification.C
```diff
@@ -1452,6 +1452,336 @@ void MeshTools::Modification::all_tri (MeshBase & mesh)
...
+                  almost_equal(w[12], w[15]) &&
+                  almost_equal(w[12], w[24]) &&
+                  almost_equal(w[13], w[14]) &&
+                  almost_equal(w[13], w[22]) &&
+                  almost_equal(w[16], w[18]) &&
+                  almost_equal(w[16], w[25]) &&
+                  almost_equal(w[21], w[23]))
+                extrude_dir = 1;
+
+              if (almost_equal(w[0], w[4]) &&
```
GiudGiud: else if

tests/mesh/all_rbb.C
roystgnr: None change in this trivial remapping, but all are good to check

src/mesh/mesh_modification.C
roystgnr: Somehow missed a line deletion when I was making a huge change; thanks!

src/mesh/mesh_modification.C
```diff
@@ -1452,6 +1452,335 @@ void MeshTools::Modification::all_tri (MeshBase & mesh)
...
+        const Point vertex_avg = (face.point(0) + face.point(1) +
+                                  face.point(2) + face.point(3))/4;
+
+        const Real w_xi  = (w[7] + w[5])/2;
+        const Real w_eta = (w[4] + w[6])/2;
+        const Real w_mid = w_xi * w_eta;
+
+        Node & midnode = face.node_ref(8);
+        midnode.set_extra_datum<Real>(weight_index, w_mid);
+        midnode = ((1+w_mid)/(w_xi+w_eta) * (w_xi*xi_avg + w_eta*eta_avg) - vertex_avg)/w_mid;
```
GiudGiud, line 1608:
```suggestion
        midnode = ((1.+w_mid)/(w_xi+w_eta) * (w_xi*xi_avg + w_eta*eta_avg) - vertex_avg)/w_mid;
```

src/mesh/mesh_modification.C
```diff
@@ -1452,6 +1452,335 @@ void MeshTools::Modification::all_tri (MeshBase & mesh)
...
+                       almost_equal(w[8], w[21]) &&
+                       almost_equal(w[9], w[17]) &&
+                       almost_equal(w[9], w[22]) &&
+                       almost_equal(w[10], w[18]) &&
+                       almost_equal(w[10], w[23]) &&
+                       almost_equal(w[11], w[19]) &&
+                       almost_equal(w[11], w[24]) &&
+                       almost_equal(w[20], w[25]))
+                extrude_dir = 2;
+
```
GiudGiud: should you assert the coordinates fit a 2.5D extrusion, not just the weights?

GiudGiud: if the elements are coming from a regular mesh the weights are all defaults at this point?

GiudGiud:
btw I can't resolve threads with this level of permissions either

src/mesh/mesh_modification.C
roystgnr: No need here - if w_mid is `Real` then `int` operations with it immediately promote to `Real`.  Where you have to be careful is with `(1/2*w)` where an `int` operates on another `int` before promotion.

This "we have to put a period after every int literal" heuristic is actually kind of a step back in `--enable-single-precision` configurations - instead of doing every operation as `float` we'd end up doing a bunch (5 here) as `double` unintentionally.

src/mesh/mesh_modification.C
roystgnr: Not sure what you mean by regular, but at this point we've already adjusted edges and faces, and the weights are all defaults if the hex originally had planar faces, in which case it's still isogeometric no matter what we do with the mid-volume node.

Though, in that case, it would still give us much better approximations on shapes like frustums to get something like the "minimum-energy" mapping in 3D as well as 2D, rather than just to use this hack.  Let me try that.

GiudGiud: if we are looking to check for a truly extruded extruded, I'm thinking the coordinates are easier to check? If extruded in x, we expect two of the faces to exhibit a +x and -x normal?

roystgnr: My general code is passing tests with extruded meshes, so I think I'm much happier just pretending the hack that *only* worked on (some) extruded meshes never existed.

src/mesh/mesh_modification.C
```diff
@@ -1452,6 +1452,281 @@ void MeshTools::Modification::all_tri (MeshBase & mesh)
...
+        const Real w_center = cos_phi*std::sqrt(w0*w1);
+
+        n_center.set_extra_datum<Real>(weight_index, w_center);
+
+        // Now let's get the control point location.  This comes from
+        // a lot of back-and-forth with Gemini, but fortunately I'm
+        // rewriting it after I've already added unit tests that
+        // should scream if it's badly wrong.
+        const Real w_mid = w0/4 + w1/4 + w_center/2;
```
GiudGiud, line 1549: do you have an equation number from a reference by any chance?

GiudGiud (approved):

src/mesh/mesh_modification.C
roystgnr: Not today, I'm afraid.  I've been working from The NURBS Book, which got me to my first draft of this PR but was too opaque about some of the more complicated cases.  Gemini claims that equation is in Cottrell, Hughes, and Bazilevs, which honestly I should have bought a copy of 5 or 15 years ago, but I won't be able to easily double-check that soon.

PR #4443, opened 2026-04-23
Add capability to request subset of libMesh::n_threads()

### LINKED ISSUE #4185: Thread initialization preferences
Context: I'm looking for some advice on how to handle initialization for threads with OpenMP in libMesh. I'm currently working on a library, XDG, that provides a DAGMC-like boundary to boundary ray tracing capability and tracklength tallies for libMesh meshes (and MFEM eventually). It is thread-safe and the downstream application (OpenMC) will access it from many threads using OpenMP threaded paralellism. 

Trouble: OpenMC sets the number of OpenMP threads when it starts up, but inside the libMeshInit constructor the number of OpenMP threads will be reset to the number of threads passed to the constructor (default of1). libMeshInit is called inside the intermediate library XDG and as a result it now needs to communicate the number of threads to libMesh. I can currently do this by either a) exposing a threads argument in XDG's config/constructor and passing it through or b) collecting the number of threads set for OpenMP with omc_get_num_threads and passing it on to the libMeshInit constructor.

a) is overhead that isn't needed used XDG and b) requires me to add OpenMP as a dependency solely to query the number of threads set. Right now I'm doing b) because, despite the extra dependency, it's easier to reverse. 

For clarity, XDG operations are the same regardless of the number of threads, and in this context libMesh doesn't either. This is entirely to ensure that number of threads isn't changed for OpenMC. And for some history, this previously wasn't an issue because OpenMC was initializing libMesh directly and could pass the number of threads OpenMC as needed.

An ideal situation for me would be to set the `OMP_NUM_THREADS` variable and have that be used, but I'm open to any suggestions for alternative approaches!

lindsayad:
Closes #4185

### DIFF
--- a/configure
+++ b/configure
@@ -57353,11 +57353,22 @@ then :
 then :
   withtbb=/usr
 fi
-                  as_ac_Header=`printf "%s\n" "ac_cv_header_$withtbb/include/tbb/task_scheduler_init.h" | sed "$as_sed_sh"`
-ac_fn_cxx_check_header_compile "$LINENO" "$withtbb/include/tbb/task_scheduler_init.h" "$as_ac_Header" "$ac_includes_default"
-if eval test \"x\$"$as_ac_Header"\" = x"yes"
+
+                                                      case $withtbb in
+                    "~/"*) withtbb=$HOME${withtbb#"~"} ;;
+                  esac
+
+                                                                                          tbb_is_onetbb=no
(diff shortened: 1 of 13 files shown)


roystgnr (approved):
Did you manually test the TBB build case?  I'm not sure we have any of those in CI.

lindsayad:
> Did you manually test the TBB build case? I'm not sure we have any of those in CI.

I haven't but I will try

PR #4444, opened 2026-04-23
Add dependencies to our Netgen .buildstamp

roystgnr:
My recent update to our Netgen fork caused build failures in pre-existing build dirs, because it added a new API but our build system didn't realize that this required a recompile of part of Netgen.  The workaround (`rm contrib/netgen/.buildstamp`) is simple to do but not simple for users to figure out, so let's try to make automatic compilation work correctly.


jwpeterson (approved):

roystgnr:
Test failure unrelated

PR #4445, opened 2026-04-29
Avoid AnalyticFunction segfaults in debug mode by asserting valid pointers

jwpeterson:
There are two ways of constructing an `AnalyticFunction` (either with a scalar-valued or a vector-valued function), but `System::reinit_constraints()` expects to be able to use the latter. Previously, it was a segfault even in debug mode when you attempted to use a scalar-valued `AnalyticFunction`. With this patch, at least in debug mode, we now get an assert with a more informative error message.

### DIFF
--- a/include/numerics/analytic_function.h
+++ b/include/numerics/analytic_function.h
@@ -111,7 +111,8 @@ inline
 Output AnalyticFunction<Output>::operator() (const Point & p,
                                              const Real time)
 {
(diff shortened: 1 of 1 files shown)


roystgnr (approved):

jwpeterson:
No idea what's holding up the MOOSE clang and MOOSE debug tests for over 6 hours, but I'm just going to go ahead and merge now...

PR #4446, opened 2026-05-04
Add capability of threading over IntRange ranges, use in l2_norm_diff()

roystgnr:
This request turned out to be harder than it should have been, but if it works then there's a ton of other opportunities that ought to be much easier to make threaded now that IntRange has been fixed up to do that more easily.

### DIFF
--- a/include/utils/int_range.h
+++ b/include/utils/int_range.h
@@ -61,7 +61,7 @@ public:
     using pointer = T;
     using reference = T&;
 
-    iterator (T i) : _i(i) {}
+    explicit iterator (T i) : _i(i) {}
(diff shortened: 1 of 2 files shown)


roystgnr:
@NamjaeChoi tested this and reports semidecent speedup (37.6s in serial, 8.0s on 8 threads).  This is in `NumericVector` so the awful `virtual operator()` overhead dominates everything and a *real* optimization would probably be a `PetscVector` specialization, but at least we're confirming that it works and there aren't any massive weird thread contention issues, even with the tricky `PetscVector::_get_array` locking.

PR #4447, opened 2026-05-05
Build apps, too, when building in oprof

roystgnr:
Usually when we're profiling we have an actual application to profile with, but I just caught a performance bug *with* one of our helper apps, and in hindsight there are lots of I/O etc. that's well exercised by things like the calculator app.

Building these when we're profiling is probably worth the extra bit of compiling+linking time.


jwpeterson (approved):
I don't love that we have to remember to manually update the top-level Makefile.am every time we add a new app (?) but fixing that is a problem for another PR.

PR #4448, opened 2026-05-05
Add runtime --enable-sigint option

roystgnr:
It's useful to be able to still get a stack trace and PerfLog output after Ctrl+C.

### DIFF
--- a/include/base/libmesh_exceptions.h
+++ b/include/base/libmesh_exceptions.h
@@ -54,6 +54,11 @@ void enableFPE(bool on);
 void enableSEGV(bool on);
 
 /**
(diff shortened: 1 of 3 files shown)


jwpeterson (approved):
Failures here look unrelated to the changes. And I guess there's no way to have a test for this, so we'll just have to take your word for it :)

roystgnr:
When we add tests for the apps, we could run something in the background that busy-loops, then kill it with SIGINT (to grep the output for PerfLog stuff), then sleep and kill it again with SIGKILL (in case there's a bug in our handler that would prevent it exiting).

I intend to add tests for the apps any decade now.

PR #4451, opened 2026-05-06
Kokkos build and numerics support 

rochi00:
This PR adds the first upstreamable slice of libMesh Kokkos support: build-system integration for .K translation units plus
  standalone Kokkos numerics oracle tests.

  The scope is intentionally limited to:

  - Kokkos-aware build support for libMesh tests
  - device-safe numerics support in core vector/tensor types
  - standalone Kokkos numerics oracle coverage
  - generated autotools updates as a separate final commit


include/base/libmesh_common.h
```diff
@@ -299,6 +309,18 @@ extern bool warned_about_auto_ptr;
 #define libmesh_assert_less_equal_msg(expr1,expr2, msg)  ((void) 0)
 #define libmesh_assert_greater_equal_msg(expr1,expr2, msg)  ((void) 0)
 
+#elif defined(LIBMESH_DEVICE_ASSERT)
```
roystgnr: I feel like it would be safer to test `LIBMESH_KOKKOS_COMPILATION` here (which should always be true or false if we're compiling with or without Kokkos respectively), just in case someone later decides it wouldn't hurt to have `LIBMESH_DEVICE_ASSERT` always defined.

include/base/libmesh_device.h
```diff
@@ -0,0 +1,74 @@
...
+       ::Kokkos::abort(msg);                                            \
+     } while (0)
+
+#  define LIBMESH_DEVICE_ERROR_MSG_IF(cond, msg)                        \
+     do { if (cond) { LIBMESH_DEVICE_ERROR_MSG(msg); } } while (0)
+
+#else
+#  define LIBMESH_DEVICE_INLINE inline
+#  define LIBMESH_IN_DEVICE_CODE 0
+#  define LIBMESH_DEVICE_ERROR_MSG(msg) libmesh_error_msg(msg)
```
roystgnr: We set up libmesh_assert to forward to LIBMESH_DEVICE_ASSERT rather than the other way around; shouldn't we do the same with libmesh_error?

include/base/libmesh_exceptions.h
```diff
@@ -217,7 +213,13 @@ class TerminationException
 #ifdef LIBMESH_ENABLE_EXCEPTIONS
 #define libmesh_noexcept noexcept
 
+#if LIBMESH_IN_DEVICE_CODE
+// Kokkos device code does not support C++ exceptions.
+#define LIBMESH_THROW(e) do { LIBMESH_DEVICE_ERROR_MSG((e).what()); } while (0)
```
roystgnr, line 223: Do we not need this below too, to handle the `!LIBMESH_ENABLE_EXCEPTIONS && LIBMESH_IN_DEVICE_CODE` case?

m4/libmesh_optional_packages.m4
```diff
@@ -861,6 +861,134 @@ AM_CONDITIONAL(LIBMESH_ENABLE_METAPHYSICL, test x$enablemetaphysicl = xyes)
...
+            esac
+          ],
+          [AC_MSG_RESULT([Using caller-provided KOKKOS_CXX=$KOKKOS_CXX])])
+
+        dnl Set defaults for any variables not provided by caller or auto-detect
+        KOKKOS_CPPFLAGS="${KOKKOS_CPPFLAGS:--DLIBMESH_KOKKOS_COMPILATION -I$KOKKOS_DIR/include}"
+        KOKKOS_LDFLAGS="${KOKKOS_LDFLAGS:--L$KOKKOS_DIR/lib}"
+        KOKKOS_LIBS="${KOKKOS_LIBS:--lkokkoscore}"
+
+        AC_DEFINE([HAVE_KOKKOS], [1],
```
roystgnr: `HOPEFULLY_HAVE_KOKKOS`?

That's a joke, but it does seem like we're taking a lot from the user and environment on faith here.  IMHO this is good enough for now, but if we can do a test compilation and report any errors here at configure time it'd be far more user friendly than claiming to configure with Kokkos support only to hit errors well into compile time.

include/numerics/type_tensor.h
```diff
@@ -728,14 +750,14 @@ const T & TypeTensor<T>::operator () (const unsigned int i,
...
-inline
+LIBMESH_DEVICE_INLINE
 T & TypeTensor<T>::operator () (const unsigned int i,
                                 const unsigned int j)
 {
 #if LIBMESH_DIM < 3
 
-  libmesh_error_msg_if(i >= LIBMESH_DIM || j >= LIBMESH_DIM,
-                       "ERROR:  You are assigning to a tensor component that is out of range for the compiled LIBMESH_DIM!");
+  LIBMESH_DEVICE_ERROR_MSG_IF(i >= LIBMESH_DIM || j >= LIBMESH_DIM,
```
roystgnr: Yeah, let's definitely forward `libmesh_error_msg_if` to device macros, not the other way around.

contrib/bin/test_installed_headers.sh
```diff
@@ -49,12 +49,16 @@ fi
...
+    PKG_CONFIG=pkg-config
+fi
+
 if test "$test_CXXFLAGS" = ""; then
 
     # testing_installed_tree="yes"
 
     if test "$PKG_CONFIG" != "no"; then
-        test_CXXFLAGS=$(pkg-config libmesh --cflags)
+        test_CXXFLAGS=$($PKG_CONFIG libmesh --cflags)
```
roystgnr, line 61: My god.  Does pkg-config *still* not let you distinguish between CFLAGS and CXXFLAGS?  Then I guess this is fine, by which I mean it's horrifying but our reaction to it is as good as we can do.

tests/Makefile.am
```diff
@@ -358,9 +379,27 @@ $(top_builddir)/libmesh_oprof.la: FORCE
 
 
 if LIBMESH_ENABLE_CPPUNIT
-TESTS = run_unit_tests.sh
+TESTS += run_unit_tests.sh
 endif
 
+# Compile .K translation units with the Kokkos device compiler.
+# $(MPI_INCLUDES) is needed because KOKKOS_CXX may be nvcc/hipcc
+# instead of the MPI compiler wrapper, so mpi.h won't be found implicitly.
```
roystgnr: This is definitely a problem, but I don't see how we've solved it.  If we're configuring using an MPI compiler wrapper to compile, we don't interrogate it with a lead pipe until it reveals its include paths to us and then add them redundantly, we just use the wrapper everywhere, leaving `$MPI_INCLUDES` empty.

We can probably still work with this for now, using `--with-mpi`* to force `$MPI_INCLUDES` to end up set to something useful, but we'll need to figure out a long-term user-friendlier solution too.

roystgnr: (as opposed to a user-friendlier solution for the past, which would involve a time machine, compiler wrapper authors, and the aforementioned lead pipe)

include/gpu/kokkos_tensor_ops.h
```diff
@@ -0,0 +1,1033 @@
...
+    for (unsigned int col = 0; col < LIBMESH_DIM; ++col)
+      if (tensor_get_component(T_in, row, col) != tensor_value_type_t<TensorLike>(0))
+        return false;
+
+  return true;
+}
+
+template <typename LeftTensor, typename RightTensor>
+LIBMESH_DEVICE_INLINE
+bool tensor_equal(const LeftTensor & left, const RightTensor & right)
```
roystgnr: This feels wrong.  Correct me if I'm misunderstanding:  We're defining `operator==`, but we can't *call* it `operator==` because with these template arguments that would end up adding false candidates for `anything==anything_else`, so we give it a weird cnamespace_style name instead?  Why not ...

But no, wait, I am misunderstanding.  We *do* define `operator==` later, and we cleverly use `enable_if_t` and SFINAE to make sure it's only applied to tensor-type arguments ... and then we make it a shim to this.

So why not just move the definition here into there, so we have one name for one function instead of two?

include/gpu/kokkos_tensor_ops.h
```diff
@@ -0,0 +1,1033 @@
...
+          typename std::enable_if<std::is_void<Dummy>::value, int>::type = 0>
+LIBMESH_DEVICE_INLINE
+tensor_semantic_type_t<TensorLike> tensor_divide(const TensorLike & T_in, const Scalar & alpha)
+{
+  return tensor_divide<tensor_semantic_type_t<TensorLike>>(T_in, alpha);
+}
+
+template <typename TensorLike>
+LIBMESH_DEVICE_INLINE
+auto tensor_determinant(const TensorLike & T_in, const unsigned int dim = LIBMESH_DIM)
```
roystgnr: And here we've got both `tensor_determinant` and `det` for the same function.

I'd also say it's weird to have as a global function something that's really begging to be a method, but I have to admit that there's something to be said for using generic programming to get a determinant working for anything for which we can specialize `tensor_get_component` and get an arithmetic-capable type out.

I don't think this is the right API in the long run - it looks like the kind of thing you see when a Fortran77 developer hastily learns C++ and for a while they just basically write Fortran77 designs and translate them - but I could possibly be talked out of that.  And in the short run this might be fine.  The translation to object-based idioms might be as simple as turning every `tensor_foo(const TensorLike & T_in, ...)` into `TensorShim<TensorLike>::foo(...) const`.

roystgnr (changes requested):
We're definitely getting there.  I'd say the only thing that really ought to be fixed *now* is the error macro forwarding direction and (unless there's a good reason not to that I'm missing) the function redundancies.

m4/libmesh_optional_packages.m4
```diff
@@ -1002,6 +1002,7 @@ AS_IF([test "x$KOKKOS_DIR" != "xno"],
         CXXFLAGS="$CXXFLAGS $KOKKOS_CXXFLAGS"
         LDFLAGS="$LDFLAGS $KOKKOS_LDFLAGS"
         LIBS="$LIBS $KOKKOS_LIBS"
+        AC_LANG_PUSH([C++])
```
roystgnr: Split all commits like this in two in the future - the stuff we actually change in the first, and the stuff that `bootstrap` (or `autoreconf`, however you do it) changes in the second.

roystgnr (changes requested):
Make system and error handling look good to go now!

We've still got more duplication than we want in the math operators, though; not all of the "an operator which just shims a custom-named function that doesn't need to exist separately" cases are gone, and there's some more subtle "N operators shim a custom-named function" cases than could be turned into "N-1 operators call the one of them that now has the implementation in it".

roystgnr:
Obviated by #4499

PR #4452, opened 2026-05-07
Cached elem ranges

roystgnr:
These make it much easier for us to do threading, by simplifying the developer code, and much easier for the CPU to do threading, by amortizing the range creation over more than just a single thread block.

For applications which use ElemRange and can switch to the cached ranges (FEMSystem, hopefully MOOSE soon) this ought to be free, but I could be persuaded that we shouldn't merge this until after eliminating the overhead for other applications in the single-thread-per-rank case.

I'm pretty happy with the simplicity of `parallel_for` code combined with lambda functions here, but `parallel_reduce` is still a bit more hassle than I'd like.

I've tested this with 16-threads each on 4 ranks, with pthreads, but I could be talked into not merging until I build and try it out on TBB too.

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -234,8 +234,7 @@ public:
    * consider ourself prepared.  This is a very coarse setting; it is
    * generally more efficient to mark finer-grained settings instead.
    */
-  void unset_is_prepared()
-  { _preparation = false; }
+  void unset_is_prepared();
 
   /**
    * Tells this we have done some operation creating unpartitioned
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -59,6 +59,8 @@ enum ElemMappingType : unsigned char;
 template <class MT>
 class MeshInput;
 
+template <typename iterator_type, typename object_type>
+class StoredRange;
 
 /**
  * This is the \p MeshBase class. This class provides all the data necessary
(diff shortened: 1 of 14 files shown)


roystgnr:
Warning fixes are added in the last commit, but I had to force-push to strip out the two commits I'd accidentally included from #4450

roystgnr:
> I could be persuaded that we shouldn't merge this until after eliminating the overhead for other applications in the single-thread-per-rank case.

John didn't try to talk me into it, and even in the single-thread-per-rank case I'm seeing `LIBMESH_BENCHMARK=1` timing of 13'49" on 16 ranks on my system, [same as master right now](https://github.com/libMesh/libmesh/pull/2935#issuecomment-4443553803), so I guess the cache building is trivial.

Adding `--n_threads=4` to my usual benchmark (so using all 64 cores) gets me ... only to 13'31" with this PR, vs 13'46" in master?  Okay, apparently the next thing I need to be doing is way more threaded benchmarking.  I know we don't pass `n_threads()` along to solvers, and we don't even always do threaded assembly outside of FEMSystem, but still!

John's concern on Slack was:

> OK, I'm pretty neutral on the change because we don't use threads at all... at least for now. I do remember how easy it is to write threaded bugs in perfectly innocent looking code, however.

Which is fair.  My only bug here was from accidentally switching to a different range to thread over, not an actual threading bug, so I think we're off to a good start, but maybe that's because I'm only threading the simplest of loops and skipping the tricky ones for now.  If benchmarking reveals that that needs to change then I'll get a helgrind recipe in CI or something as backstop before I try to thread anything hairy.

roystgnr:
>  I know we don't pass `n_threads()` along to solvers

Turns out we kinda do (via `omp_set_num_threads(N)`), but it doesn't matter in those benchmarks because default PETSc solvers aren't multithreaded.  Even adding `-blas_num_threads N` to the command line doesn't make a difference.  My actual users are carefully setting up more scalable solvers, so this might not be a practical problem, but it's definitely an issue with getting useful `LIBMESH_BENCHMARK` threaded scaling results, since *non*-carefully setting up more scalable solvers (`-pc_type hypre -pc_hypre_type boomeramg` and cross my fingers!) doesn't always even converge.

PR #4453, opened 2026-05-08
side_vertex_average_normal p-refinement compatibility fix

roystgnr:
Also a bunch of unit test refactoring and expanded unit test feature coverage for that method, but really the core here is the one-line fix in elem.C.

### DIFF
--- a/tests/geom/side_vertex_average_normal_test.C
+++ b/tests/geom/side_vertex_average_normal_test.C
@@ -31,12 +31,12 @@ public:
   LIBMESH_CPPUNIT_TEST_SUITE( SideVertexAverageNormalTest );
   CPPUNIT_TEST( testEdge2 );
   CPPUNIT_TEST( testEdge3 );
-  CPPUNIT_TEST( testTri3 );
-  CPPUNIT_TEST( testQuad4 );
-  CPPUNIT_TEST( testTet4 );
-  CPPUNIT_TEST( testPyramid5 );
-  CPPUNIT_TEST( testPrism6 );
-  CPPUNIT_TEST( testHex8 );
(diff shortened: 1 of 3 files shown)


GiudGiud (approved):

NamjaeChoi (approved):

tests/geom/side_vertex_average_normal_test.C
```diff
@@ -234,132 +220,108 @@ public:
...
-      const Point n5 = pyr5.side_vertex_average_normal(4);
-      LIBMESH_ASSERT_FP_EQUAL(0, n5(0), TOLERANCE*TOLERANCE);
-      LIBMESH_ASSERT_FP_EQUAL(0, n5(1), TOLERANCE*TOLERANCE);
-      LIBMESH_ASSERT_FP_EQUAL(-1, n5(2), TOLERANCE*TOLERANCE);
-    }
+    // Reference
+    const Elem * pyrs[] = {&ReferenceElem::get(PYRAMID5),
+                           &ReferenceElem::get(PYRAMID13),
+                           &ReferenceElem::get(PYRAMID14),
+                           &ReferenceElem::get(PYRAMID18)};
```
GiudGiud, line 259: this is getting us some new coverage on pyramids, nice

roystgnr: On everything!  Well, everything except infinite elements and shell elements, anyway; I was too lazy to add those in.

My first thought had been that maybe the p-refinement compatibility bug would only occur on HEX27 or something major and so I'd do some major setup changes to get a plain plus a p-refined test for every element type, but after I discovered it triggered on EDGE3 and I realized it would trigger on *any* element I figured that was overkill.

PR #4454, opened 2026-05-11
Assert that we don't PerfLog push/pop in threads

roystgnr:
Our stacks here aren't thread-safe, we don't want to add expensive locks to make them thread-safe, but we also don't want people crashing with corrupted stacks.

With libMesh Threads:: this was already solved by temporarily disabling perf logging in threads, but we might as well assert that that's working.

Users are writing custom OpenMP code, and we already have detection for whether OpenMP's enabled, so it's cheap enough in dbg/devel to see if we're in a non-libMesh OpenMP thread section.

### DIFF
--- a/include/parallel/threads_pthread.h
+++ b/include/parallel/threads_pthread.h
@@ -276,16 +276,16 @@ void parallel_for (const Range & range, const Body & body,
                        "global thread count (" << libMesh::n_threads() << ").");
   Threads::BoolAcquire b(Threads::in_threads);
 
+  unsigned int actual_threads = num_pthreads(range, n_threads);
+
   // If we're running in serial - just run!
-  if (n_threads == 1)
+  if (actual_threads == 1)
   {
     body(range);
     return;
   }
 
   DisablePerfLogInScope disable_perf;
-
-  unsigned int actual_threads = num_pthreads(range, n_threads);
 
   std::vector<std::unique_ptr<Range>> ranges(actual_threads);
   std::vector<RangeBody<const Range, const Body>> range_bodies(actual_threads);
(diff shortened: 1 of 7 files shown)


roystgnr:
Pinging @NamjaeChoi to try this out and make sure (in dbg/devel modes) it screams for him.

roystgnr:
It might be worthwhile to talk about future possibilities here too.

For non-`Threads::` users, IMHO we're done.  If we don't have an opportunity to do our own setup and teardown around a parallel loop then there's really not much we can do.

For `Threads::` users, including ourselves ... we already don't disable perf logging in the --n_threads=1 case, but it might be interesting to consider creating a per-thread PerfLog in `parallel_foo`, adding a level of pointer indirection or something else cheaper than a mutex to let us use those instead of the global log, then merging them back into the global log at the end of each parallel section.

include/parallel/threads_pthread.h
```diff
@@ -276,16 +276,18 @@ void parallel_for (const Range & range, const Body & body,
...
   // If we're running in serial - just run!
-  if (n_threads == 1)
+  if (actual_threads == 1)
   {
     body(range);
     return;
   }
 
-  DisablePerfLogInScope disable_perf;
+  Threads::RAIIAcquire<int> a(Threads::active_threads, actual_threads);
```
jwpeterson: What's the thinking behind using simply "a" for the `RAIIAcquire` variable name here/elsewhere? It looks like the rest of the code doesn't refer to `a` so I guess the intent was just to use a "throwaway" variable name? IMO it might be better to use something a little more self-documenting.

roystgnr: I got away with being lazy about `BoolAcquire b`, and I forgot the old trope about how thrill-seeking criminals always get caught because they return to the scene of the crime?

New commit added with name changes.

jwpeterson (approved):

PR #4455, opened 2026-05-12
Always print ReferenceCounter info for leaks

roystgnr:
This info is kind of trivial (and so tempting to suppress via the command-line option) when there's not a leak, but in the event there is a leak, we shouldn't be hiding any details.

See the discussion in idaholab/moose#32877

### DIFF
--- a/src/base/libmesh.C
+++ b/src/base/libmesh.C
@@ -772,11 +772,18 @@ LibMeshInit::~LibMeshInit()
   // Ask the \p ReferenceCounter to print its reference count
   // information, if printing is enabled.  This allows us to find
   // memory leaks.
-  ReferenceCounter::print_info ();
+  //
+  // If we don't have any leaks, we'll print to cout, only if we've
+  // been told to.  If we do, we'll print to cerr, to make sure the
+  // user doesn't miss a leak that's only on a processor other than 0.
+  if (ReferenceCounter::n_objects() == 0)
+    ReferenceCounter::print_info ();
 
   // Scream specifically if we detect a memory leak
   if (ReferenceCounter::n_objects() != 0)
     {
+      ReferenceCounter::print_info (libMesh::err);
+
       libMesh::err << "Memory leak detected!"
                    << std::endl;
 
--- a/src/base/libmesh.C
+++ b/src/base/libmesh.C
@@ -765,16 +765,16 @@ LibMeshInit::~LibMeshInit()
   // Clear the thread task manager we started
   task_scheduler.reset();
 
-  // Force the \p ReferenceCounter to print
-  // its reference count information.  This allows
-  // us to find memory leaks.  By default the
-  // \p ReferenceCounter only prints its information
-  // when the last created object has been destroyed.
-  // That does no good if we are leaking memory!
+  // If we have a memory leak, we want the details printed.
+  if (ReferenceCounter::n_objects() != 0)
+    ReferenceCounter::enable_print_counter_info();
+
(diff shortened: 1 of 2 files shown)


roystgnr:
If @jwpeterson agrees, I'd even be happy to go as far as defaulting to `disable_print_counter_info()` when there's no leak (with an optional `--enable-refcount-printing` argument to override, instead of the current `--disable-` argument).  The reference counter info is occasionally interesting for profiling and it's often interesting for debugging leaks but 99.9% of the time it's just clutter.

loganharbour:
Works for me. I think the quiet behavior being the default is a good idea

jwpeterson:
> If @jwpeterson agrees, I'd even be happy to go as far as defaulting to disable_print_counter_info() when there's no leak

I can't remember the details, but I think there was a situation (maybe in MOOSE?) where "Memory leak detected!" was being printed at the end of *every* simulation, but it either wasn't "real" and/or there was nothing we could actually do about it. So that was why we added the ability to disable the leak message entirely. 

I searched for the first instance of `disable_print_counter_info()` and found 0cc5cecd56510408b876a7ffefcecfd01e362a42 (from 2010) and it refers to "Paul", so maybe it was GRINS-related? Anyway, I'm fine with the update, but I guess we'll need to see if there's more false positives once it's turned on all the time.

jwpeterson (approved):

src/base/libmesh.C
```diff
@@ -765,16 +765,16 @@ LibMeshInit::~LibMeshInit()
...
-  // That does no good if we are leaking memory!
-  ReferenceCounter::print_info ();
+  // If we have a memory leak, we want the details printed.
+  if (ReferenceCounter::n_objects() != 0)
+    ReferenceCounter::enable_print_counter_info();
 
+  // Ask the \p ReferenceCounter to print its reference count
+  // information, if printing is enabled.  This allows us to find
+  // memory leaks.
+  ReferenceCounter::print_info ();
```
loganharbour: Should we print this to err instead now that we're only printing in the case of leaks? as the next message is to err too

roystgnr: That's a really good idea.  A big leak on a big parallel run will spew a ton of output now, but that's much better than just getting the error message without any hint of the cause if the leak doesn't happen to trigger on rank 0.

src/base/libmesh.C
```diff
@@ -765,18 +765,25 @@ LibMeshInit::~LibMeshInit()
...
+  // been told to.  If we do, we'll print to cerr, to make sure the
+  // user doesn't miss a leak that's only on a processor other than 0.
+  if (ReferenceCounter::n_objects() == 0)
+    ReferenceCounter::print_info ();
 
-  // Print an informative message if we detect a memory leak
+  // Scream specifically if we detect a memory leak
   if (ReferenceCounter::n_objects() != 0)
     {
+      ReferenceCounter::print_info (libMesh::err);
```
lindsayad, line 785: I feel like this would read a little cleaner. Comment would probably need updating
```suggestion
  ReferenceCounter::print_info(ReferenceCounter::n_objects() ? libMesh::err : libMesh::out);

  // Scream specifically if we detect a memory leak
  if (ReferenceCounter::n_objects() != 0)
  {
```
The read is a little jumbled to me even with the proposed cleanup as I feel like a caller who says `print_info` is saying they want to print info, but actually whether the info gets printed is predicated on the status of calls to `enable/disable_print_counter_info()`. But changing that would obviously induce changes in user observed behavior

PR #4456, opened 2026-05-14
Update qhull

lindsayad:
This adds qhull as a submodule. qhull now is cmake based so I had Claude model its m4 off of what's done in netgen's m4

I feel like we should likely just remove the 2012 version, but I haven't done that yet


jwpeterson:
```
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: clone of 'git@github.com:qhull/qhull.git' into submodule path '/tmp/build/libmesh/contrib/qhull/git-qhull' failed
Failed to clone 'contrib/qhull/git-qhull'. Retry scheduled
Cloning into '/tmp/build/libmesh/contrib/qhull/git-qhull'...
Host key verification failed.
```

I don't think we can we use the SSH protocol to clone repos in CI?  At least for Eigen and Nanoflann, we use an https URL in .gitmodules.

roystgnr (approved):
> I had Claude model its m4 off of what's done in netgen's m4

You had it work using my autotools->cmake integration as a model, and it didn't argue or refuse?  We definitely haven't reached AGI or solved Alignment.

PR #4460, opened 2026-05-18
Try to fix "incomplete type" in Test No Optional

roystgnr:
I can't actually seem to reproduce this, but my best guess is that our regular builds are getting <atomic> indirectly from other libMesh headers, my own builds are getting <atomic> indirectly from other system headers, but our Test No Optional recipe actually needs to directly include <atomic> if it wants to use std::atomic.


jwpeterson (approved):

roystgnr:
MOOSE modules failure unrelated

PR #4461, opened 2026-05-19
Add circumcenter calculation, unit tests

roystgnr:
I need this to expand all_rbb() compatibility in #4442, but it's pretty self-contained too.


jwpeterson (approved):
My only request (over IM) was to update the docs regarding collinear points, thanks!

roystgnr:
Feels silly to re-run CI for a comment change, but I don't need this as urgently as I thought (#4442 is ready without it and it'll be at least a couple weeks before the next phase of that needs it) so I'll just turn on auto-merge here and we'll see if there's been any downstream breakage during the last 19 hours.

PR #4463, opened 2026-05-19
Have qhull do PIC in libs

lindsayad:
Not sure why we haven't run into this anywhere else but when linking
on Ubuntu 26.04 I'm getting errors saying I need to rebuild the qhull
libs with PIC.

### DIFF
--- a/configure
+++ b/configure
@@ -59326,8 +59326,9 @@ then :
 
                                         mkdir -p contrib/qhull/install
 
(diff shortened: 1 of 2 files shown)


roystgnr (approved):
I don't understand why *I'm* not seeing this on Ubuntu 26.04, but this looks like the right fix for the problem and it shouldn't hurt anything.

PR #4464, opened 2026-05-21
All tri for polygonal and polyhedra elements

GiudGiud:
Have not reviewed it yet hence the draft
EDIT: ok I reviewed it locally and in the git API and it looks good for consideration

If there is such a label, should be marked as `AI assisted`


roystgnr (approved):
Modules test failure unrelated

The AI commits are actually disturbingly good here - how hard was Claude to prompt on this one?

GiudGiud:
It wasn't very complicated either but yes I am fairly pleased at them too. I did tell it what to do

roystgnr:
MOOSE modules test failure is unrelated

PR #4465, opened 2026-05-22
Make atomic matrix init methods public

lindsayad:
I have a matrix which is not square, hence not initializable with the DofMap, and I want to be able to initialize said matrix with hash table based matrix assembly. Making `init_without_preallocation` and `finish_initialization` public allows me to easily do this. An alternative would be to add another `init` overload but I think we have enough of those

### DIFF
--- a/include/numerics/petsc_matrix.h
+++ b/include/numerics/petsc_matrix.h
@@ -293,7 +293,6 @@ public:
 
   virtual void restore_original_nonzero_pattern() override;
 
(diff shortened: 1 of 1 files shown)


jwpeterson (approved):

roystgnr (approved):
I think these are clean enough to make public.

roystgnr:
MOOSE modules failure is unrelated

PR #4467, opened 2026-05-27
Disable AMR-dependent tests without --enable-amr

roystgnr:

PR #4468, opened 2026-05-27
Fix build_sphere with DistributedMesh

roystgnr:
This one was a pre-existing bug!  I'm not sure how we managed to avoid triggering it before the all_rbb() tests.


jwpeterson (approved):

roystgnr:
We've already passed one previously-failing test with each of these PRs; I'm going to merge this and #4467 immediately and see if that unblocks devel->master.

PR #4469, opened 2026-05-27
Check for higher C++ standards

roystgnr:
We're passing this on to downstream compilers for applications, and we're about to be doing so for users of requires-C++20 versions of Kokkos, so unless the user specifically requests --with-cxx-std-max= something lower, let's give them the highest standard that passes tests, not just C++17.

I want to make sure this passes CI, but also make sure this still gives us C++17 with our "Min gcc" (g++ 9) recipe, where the compiler supports "c++2a" (partial C++20 support) but we shouldn't accept that.


roystgnr:
Wow, even MOOSE isn't advanced-clang-warning clean.  Let me try just bumping us up to default C++20.

roystgnr:
On the bright side, the Min gcc tests do confirm that we're not accidentally getting a half-baked "C++2a" there.

roystgnr:
Even `-std=c++20 -Werror` is enough to break MOOSE with `clang++` 20.1.8, huh?  Okay.  I'll mothball this, and put the warning fixes in a separate PR.

PR #4470, opened 2026-05-28
Fix clang++ warnings in newer C++ standards

roystgnr:
I'm not entirely sure why this is a warning (these are like 16 bytes each, tops; can't the copy fit in a couple registers?) or why it's only a warning when we bump up our C++ standard version, but it's breaking our --enable-werror test recipes.

Stripped out of #4469

### DIFF
--- a/src/numerics/static_condensation_dof_map.C
+++ b/src/numerics/static_condensation_dof_map.C
@@ -93,7 +93,7 @@ void StaticCondensationDofMap::add_uncondensed_dof_plus_constraint_dofs(
                             uncondensed_local_dof_number,
                             constraint_dofs);
   if (is_constrained)
-    for (const auto [full_constraining_dof, weight] : it->second)
+    for (const auto & [full_constraining_dof, weight] : it->second)
       {
         libmesh_ignore(weight);
         // Our constraining dofs may themselves be constrained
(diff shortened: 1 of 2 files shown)


jwpeterson:
I'm curious what the warning was, exactly? I don't use clang at the moment, but I do think we have some structured binding code like this that doesn't use a reference.

jwpeterson (approved):

roystgnr:
```
../src/numerics/static_condensation_dof_map.C:96:21: error: loop variable '[full_constraining_dof, weight]' creates a copy from type 'std::pair<const unsigned long, double> const' [-Werror,-Wrange-loop-construct]
    for (const auto [full_constraining_dof, weight] : it->second)
                    ^
../src/numerics/static_condensation_dof_map.C:96:10: note: use reference type 'std::pair<const unsigned long, double> const &' to prevent copying
    for (const auto [full_constraining_dof, weight] : it->second)
```

I can't actually reproduce it locally with my MacBook clang++, but you can see it from https://civet.inl.gov/job/3859124/

Seriously, though: 12 bytes, 16 on other configurations, vs an 8-byte pointer.  Just make the stupid copy, LLVM.  It looks to me like this might even be a pessimization (although so trivial I wouldn't care) in the petsc_dm_wrapper.C cases, where the compiler would be forced to dereference that pointer more than once (those maps are passed by reference, so we can't be sure nobody else has their own pointer to their data, and we call functions in between the data usage, so we can't be sure nobody used their own pointer to change that data), whereas a copy we'd just keep in a register after a single dereference.

PR #4473, opened 2026-06-01
Patched variational smoother indexing bug

pbehne:
The `VariationalMeshSmoother` logic that determines nodal constraints had a pathway in which the code attempts to access an element of an empty vector, resulting in a crash. This PR closes that pathway. Ref. #4082


roystgnr (approved):

roystgnr:
I'd love a unit test that triggers (used to trigger) this bug, but I know how hard it can be to boil down a real mesh to a minimal replication.

PR #4475, opened 2026-06-03
NURBS sphere tests

roystgnr:
Plus a couple things thrown in when I caught a bug in a previous test and when I was looking at performance of some of the new tests.

### DIFF
--- a/src/mesh/mesh_smoother_laplace.C
+++ b/src/mesh/mesh_smoother_laplace.C
@@ -30,6 +30,7 @@
 #include "libmesh/parallel_algebra.h" // StandardType<Point>
 #include "libmesh/int_range.h"
 #include "libmesh/elem_side_builder.h"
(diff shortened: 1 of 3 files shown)


roystgnr:
Wait a minute, I'm an idiot.  Should have looked at the git diff *before* opening the PR.  My performance bug was caused by some heavyweight tests I hacked into place while looking into a possible correctness bug.  I'm not seeing stupidly slow (12s in opt mode!) results when tests get 448 elements out of `build_sphere()`, I'm seeing understandably slow results when tests get 229,376 elements out of it.  I can probably reenable the 448-element cases, once I tell them to actually only request 448 elements each...

roystgnr:
Yeah, with the overrefined runs removed we're down to a 30th of a second for all the AllRBB tests in opt mode, half a second in dbg.  Much better.

PR #4476, opened 2026-06-03
Make our Threads:: pthreads backend respect grainsize, make grainsize runtime-configurable

roystgnr:
This doesn't yet fix the MOOSE performance issues I was expecting it to, but I think it's either a step towards that or a worthwhile change on its own.

### DIFF
--- a/include/base/libmesh_base.h
+++ b/include/base/libmesh_base.h
@@ -41,6 +41,13 @@ processor_id_type global_processor_id();
 unsigned int n_threads();
 
 /**
+ * \returns The default grain size (work unit, chunk size), typically
+ * measured in nodes or elements or vector indices, below which work
+ * will not be split between additional threads.
+ */
+unsigned int default_grainsize();
+
+/**
  * Namespaces don't provide private data,
  * so let's take the data we would like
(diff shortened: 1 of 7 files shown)


roystgnr:
Oh, now I'm starting to understand.  The reason MOOSE doesn't speed up is that it's got a hardcoded `GRAIN_SIZE = 1` in its main ranges ... and also MOOSE has a whole `ThreadedGeneralUserObject` system that had the same "assume there's no grainsize" bug our recent unit tests did.  I was going to talk some smack about the Abaqus plugin there for making bad assumptions, but it turned out to be a framework bug that only the Abaqus interface triggered, with proper test coverage and a semi-helpful error message to boot.

roystgnr:
Even with the `GRAIN_SIZE` thing fixed I'm not seeing nearly as much speedup as I should, though.  Still looking into that.

roystgnr:
> Still looking into that.

I found another MOOSE bug that explains most or all of it, and that suggests other downstream issues which could explain the rest.

The changes here should be good, though I'll want to wait to re-run CI on a post- https://github.com/idaholab/moose/pull/33049 MOOSE before we merge.

PR #4479, opened 2026-06-09
MeshTools::Modification::interpolate_boundary

roystgnr:
After futzing with build_sphere() recently, talking to @GiudGiud about level set meshing has me thinking this code could be useful on its own too.

I probably should add more test coverage for various old and new options before we merge this, just wanted to get it up (and testing with CI configuration variants) before I forgot about it.

### DIFF
--- a/src/mesh/mesh_generation.C
+++ b/src/mesh/mesh_generation.C
@@ -2346,58 +2346,10 @@ void MeshTools::Generation::build_sphere (UnstructuredMesh & mesh,
 
       mesh_refinement.uniformly_refine(1);
 
-      for (const auto & elem : mesh.active_element_ptr_range())
-        for (auto s : elem->side_index_range())
-          if (elem->neighbor_ptr(s) == nullptr || (mesh.mesh_dimension() == 2 && !flat))
-            {
-              elem->build_side_ptr(side, s);
-
-              // Pop each point to the sphere boundary.  Keep track of
-              // any points we don't own, so we can push their "moved"
-              // status to their owners.
-              for (auto n : side->node_index_range())
-                {
-                  Node & side_node = side->node_ref(n);
-                  side_node =
-                    sphere.closest_point(side->point(n));
-
-                  if (!is_replicated &&
-                      side_node.processor_id() != mesh.processor_id())
-                    moved_ghost_nodes.insert(side_node.id());
-                }
-            }
-
-      if (!is_replicated)
-        {
-          std::map<processor_id_type, std::vector<dof_id_type>> moved_nodes_map;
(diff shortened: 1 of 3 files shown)


src/mesh/mesh_generation.C
```diff
@@ -2346,58 +2346,10 @@ void MeshTools::Generation::build_sphere (UnstructuredMesh & mesh,
...
-          Parallel::push_parallel_vector_data
-            (mesh.comm(), moved_nodes_map, action_functor);
-
-          // Then get node positions to anyone else with them ghosted
-          SyncNodalPositions sync_object(mesh);
-          Parallel::sync_dofobject_data_by_id
-            (mesh.comm(), mesh.nodes_begin(), mesh.nodes_end(),
-             sync_object);
-        }
+      const bool move_only_boundary_nodes =
```
GiudGiud, line 2349: add a comment?
how is 3D needing the same thing as a single-level flat refinement tree?
is that what flat means? this isnt super clear 

my confusion is why we need a comment hehe

src/mesh/mesh_modification.C
```diff
@@ -1963,6 +1965,105 @@ void MeshTools::Modification::smooth (MeshBase & mesh,
 
 
 
+void MeshTools::Modification::interpolate_boundary (MeshBase & mesh,
+                                                    const Surface & surface,
+                                                    std::set<std::size_t> ids,
+                                                    bool boundary_nodes)
```
GiudGiud: should this last one be "only_boundary_nodes" ?

src/mesh/mesh_generation.C
roystgnr: Does https://github.com/libMesh/libmesh/pull/4478/changes#diff-dcf2d4c21179efacd0872f7b3ceb49e27a665fd2c79c3985097cc34f0a98bc6bR127 help?  I noticed how woefully underdocumented build_sphere() was but figured that was independent of the refactoring.

src/mesh/mesh_modification.C
roystgnr: I feel like that's still confusing, just in different ways.  That flag also controls whether we interpret `ids` as boundary ids or subdomain ids.

How about `use_boundaries`?  And perhaps more importantly, change `interpolate_boundary()` to `interpolate_surface()` so that doesn't introduce as much confusion in the first place?

GiudGiud: sounds better

src/mesh/mesh_generation.C
GiudGiud: helps

jwpeterson (approved):

PR #4481, opened 2026-06-12
TRI7 center node adjustment and TRI6/7 degeneracy checks after midpoint relocation

### LINKED ISSUE #4480: Quadratic Triangulation Degeneracy After Midpoint Relocation
When `TriangulatorInterface` creates quadratic elements, it first generates linear TRI3 elements and then converts them to TRI6 or TRI7. If the input mesh is quadratic, the midside nodes of the resulting TRI6/TRI7 elements are subsequently moved to match the input geometry.

This approach introduces two potential issues:

- Center node displacement: For TRI7 elements, moving an edge midpoint can shift the center node outside the physical element boundaries.

- Element tangling: Shifting midside nodes can distort the element mapping, potentially leading to tangled or invalid geometries.

To resolve this, the center nodes of TRI7 elements should be re-adjusted following midside node relocation. Additionally, all TRI6/TRI7 elements must be verified for geometric validity, throwing an error if any degenerate elements are detected.

miaoyinb:
closes #4480

### DIFF
--- a/include/mesh/triangulator_interface.h
+++ b/include/mesh/triangulator_interface.h
@@ -387,6 +387,23 @@ protected:
   void increase_triangle_order();
 
   /**
+   * Helper called at the end of \p increase_triangle_order() for \p
+   * TRI7 to reposition the interior node to the curved-mapping image
+   * of the reference centroid after boundary midpoints have been
+   * snapped.  Safe to call unconditionally for TRI7.
+   */
+  void fixup_tri7_center_nodes();
+
+  /**
+   * Helper called at the end of \p increase_triangle_order() to verify
+   * that no boundary-midpoint snap has tangled a quadratic triangle,
+   * by sampling the signed element Jacobian at the reference nodes and
+   * centroid.  Errors out with a diagnostic naming the bad element and
+   * side if the Jacobian is non-positive.
+   */
+  void verify_quadratic_elements();
+
+  /**
    * Helper function to check holes for intersections if requested.
    */
   void verify_holes(const Hole & outer_bdy);
--- a/include/mesh/triangulator_interface.h
+++ b/include/mesh/triangulator_interface.h
@@ -264,6 +264,17 @@ public:
   bool get_verify_hole_boundaries() const {return _verify_hole_boundaries;}
 
   /**
+   * For \p TRI7 elements, after boundary mid-edge nodes have been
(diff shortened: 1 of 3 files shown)


miaoyinb:
As this PR moves TRI7 center node, MOOSE's `meshgenerators/xy_delaunay_generator.xydelaunay/quadratic_tri7` test needs to be regolded.

roystgnr:
As a general rule "midpoint goes at the centroid" isn't always right, since you can do some clever things with boundary layers by using slightly-off-center higher order nodes in the grading, but in *this* context I think this is exactly the right thing to do.

If you don't want to create a big "have to get the submodule updated to regold" vs "can't merge the PR without breaking a ton of recipes" chicken-and-egg problem, the general way to do that is to add new behavior as an option before we change behavior by default.  Default the new behavior to "off", then after MOOSE is updated turn it on there and regold, then change the default to "on".

miaoyinb:
Good point, I will make it optional.

miaoyinb:
Added the option and made the original approach default.

Are the two failed tests related? Thanks @roystgnr

roystgnr:
Sorry, I've been out on vacation for a couple weeks.

The failed tests are definitely related.  I can reproduce the failures with a `configure --enable-distmesh` build of libMesh, with `LIBMESH_RUN='mpirun -np 3' LIBMESH_OPTIONS='--re Poly2TriEdge' make -j 65 check -C tests/` (substitute your own `-j` argument there).  I haven't dug more deeply than identifying the failing test, but it looks like rank 0 thinks it's passing but one of the other two ranks is returning a non-0 exit code, probably failing a CppUnit assertion.

miaoyinb:
> Sorry, I've been out on vacation for a couple weeks.
> 
> The failed tests are definitely related. I can reproduce the failures with a `configure --enable-distmesh` build of libMesh, with `LIBMESH_RUN='mpirun -np 3' LIBMESH_OPTIONS='--re Poly2TriEdge' make -j 65 check -C tests/` (substitute your own `-j` argument there). I haven't dug more deeply than identifying the failing test, but it looks like rank 0 thinks it's passing but one of the other two ranks is returning a non-0 exit code, probably failing a CppUnit assertion.

Thanks, let me look into this.

roystgnr (approved):

miaoyinb:
@roystgnr just a friendly reminder that this still needs to be merged. Thanks.

PR #4483, opened 2026-06-15
Prevents circumscribing other nodes in the triangulation

ConnorOuellette:
Swapped ordering of code blocks to prevent early return that caused the issue. Resolves bug discovered during MOOSE dual_mesh development.

### DIFF
--- a/src/mesh/poly2tri_triangulator.C
+++ b/src/mesh/poly2tri_triangulator.C
@@ -15,7 +15,6 @@
 // License along with this library; if not, write to the Free Software
 // Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 
-
 #include "libmesh/libmesh_config.h"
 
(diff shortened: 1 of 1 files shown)


roystgnr (changes requested):
I believe the actual change buried in here is a necessary bugfix, but we can't bury it in a thousand lines of clang-format changes

ConnorOuellette:
Should now be much more coherent and reasonable

roystgnr (approved):
Looks perfect.  We should "Squash and Merge" rather than the usual, though, to keep the history clean.

Assuming CI passes, that is.  It's entirely possible that we golded something downstream that was affected by the same bug.

roystgnr:
>  It's entirely possible that we golded something downstream that was affected by the same bug.

Looks like we did?  We might need to temporarily disable that test downstream so we can safely get this in.

roystgnr:
Changes merged in #4535

PR #4484, opened 2026-06-16
Add the capability to specify subdomain IDs for boundary elements

lindsayad:
This will make it easier to make these APIs direct replacements for downstream lower-d block mesh generators

### DIFF
--- a/include/mesh/boundary_info.h
+++ b/include/mesh/boundary_info.h
@@ -182,10 +182,18 @@ public:
    * generates a mesh with elements of mixed dimension.
    *
    * Only boundary elements with the specified ids are created.
+   *
+   * If \p new_subdomain_ids is empty, newly created elements retain the
+   * default subdomain id.  If it contains a single entry, all new elements
+   * are assigned that subdomain id.  If its size equals
+   * \p requested_boundary_ids.size(), new elements are assigned subdomain
+   * ids in one-to-one correspondence with the boundary ids (iterated in
+   * sorted order).
    */
   void add_elements (const std::set<boundary_id_type> & requested_boundary_ids,
                      UnstructuredMesh & boundary_mesh,
-                     bool store_parent_side_ids = false);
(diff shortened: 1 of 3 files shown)


include/mesh/boundary_info.h
```diff
@@ -182,10 +182,18 @@ class BoundaryInfo : public ParallelObject
    * generates a mesh with elements of mixed dimension.
    *
    * Only boundary elements with the specified ids are created.
```
jwpeterson: It's not part of your original diff, but I'd change this comment line to something like "Only boundary element sides with boundary ids in the `requested_boundary_ids` set are considered for addition." 

Since now there are two different "sets" of ids that one can "specify" for the function, I think it would be good to be more specific.

src/mesh/boundary_info.C
```diff
@@ -791,6 +799,17 @@ void BoundaryInfo::add_elements(const std::set<boundary_id_type> & requested_bou
...
 
+      // Assign subdomain id if requested.
+      if (new_subdomain_ids.size() == 1)
+        new_elem->subdomain_id() = new_subdomain_ids[0];
+      else if (new_subdomain_ids.size() > 1)
+        {
+          auto it = requested_boundary_ids.find(triggering_bcid);
+          libmesh_assert(it != requested_boundary_ids.end());
+          new_elem->subdomain_id() =
+            new_subdomain_ids[std::distance(requested_boundary_ids.begin(), it)];
```
jwpeterson: Minor comment, although I can't imagine it ever showing up in profiling, computing the `std::distance` between `std::set` iterators is O(N) complexity, so a potential improvement would be to create a `requested_boundary_ids_vec` from the original set passed in by the user. You can still look up items quickly in it using `Utility::binary_find()` and the subsequent `std::distance` check will also be O(1) in the vector.

jwpeterson (approved):

PR #4485, opened 2026-06-16
Add preparation flag related to sub id to name map

lindsayad:
And make sure we do sync the map when we're completing preparation

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -341,6 +341,16 @@ public:
   { _preparation.has_boundary_id_sets = false; }
 
   /**
+   * Tells this we have done some operation which may have left the
+   * subdomain id to name map inconsistent across processors.
+   *
+   * User code which adds or changes subdomain names should call this
+   * method.
+   */
+  void unset_has_synched_subdomain_name_map()
+  { _preparation.has_synched_subdomain_name_map = false; }
+
+  /**
    * \returns \p true if all elements and nodes of the mesh
    * exist on the current processor, \p false otherwise
    */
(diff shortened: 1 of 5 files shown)


roystgnr:
At first glance it looks like we just need to fix up downstream use cases before we can rebase this and retry CI ...

But at second glance, it looks like at least some of those failures downstream are because we're e.g. asserting `_mesh->is_prepared()` where we should only be asserting `_mesh->_preparation.has_boundary_id_sets`?  We might be able to fix that here; I could push a commit to try, if my comment here isn't clear enough.

roystgnr:
Huh.  Now I'm seeing `is_prepared()` assertions failing in `DofMap`, and I don't actually want to weaken those.  We really shouldn't be getting to DoF numbering or renumbering without a `prepare_for_use()` (or ideally `complete_preparation()`) first, and I'm not sure how we are.

lindsayad:
Those failures will be fixed by https://github.com/idaholab/moose/pull/33276, so we can wait for that to make it to MOOSE master and then invalidate here

lindsayad:
Might not have needed to force push to revert a commit just added into devel, but I did. https://github.com/libMesh/libmesh/pull/4485/commits/facfa4bba9b6f9e48d4f4e254ee411d8edd2c8a7 is the only new commit since @roystgnr's last comment

lindsayad:
This will still fail until https://github.com/idaholab/moose/pull/33276 makes it into master

lindsayad:
@roystgnr things are finally green here

roystgnr (approved):

PR #4489, opened 2026-07-02
Print all failed assertion expressions

roystgnr:
We print all the specialized assertions' expressions, but in the general case we've apparently just been relying on file/line output, which is less friendly.


roystgnr:
And to boot, our file/line-number output is a bit obscured, because we print it *after* the stack trace, which only lists function names.  Maybe I should move that while I'm at it...

jwpeterson (approved):

PR #4490, opened 2026-07-02
Add Polygon/Polyhedron NSIDED/NFACED ExodusII output

ConnorOuellette:
## Summary

This adds ExodusII output support for libMesh’s arbitrary polygon and polyhedron elements:

- writes `C0POLYGON` elements as Exodus `NSIDED` blocks
- writes `C0POLYHEDRON` elements as Exodus `NFACED` blocks
- emits the associated face block, face-node connectivity, face counts, element-face connectivity, and element face counts needed for `NFACED` output

## Notes

- If using ParaView, the default IOSS Reader does not support polygonal/polyhedral input. Please use the Legacy Exodus Reader.
- Fixed an error with build_cube() where corner nodes were scaled twice

### DIFF
--- a/include/mesh/exodus_header_info.h
+++ b/include/mesh/exodus_header_info.h
@@ -57,7 +57,8 @@ public:
     // Pack individual integers into vector
     std::vector<int> buffer =
       {num_dim, num_elem, num_elem_blk, num_node_sets,
-       num_side_sets, num_elem_sets, num_edge_blk, num_edge};
+       num_side_sets, num_elem_sets, num_edge_blk, num_edge,
+       num_face_blk, num_face};
 
     // broadcast integers
     comm.broadcast(buffer);
@@ -72,6 +73,8 @@ public:
     num_elem_sets = buffer[ctr++];
     num_edge_blk  = buffer[ctr++];
     num_edge      = buffer[ctr++];
+    num_face_blk  = buffer[ctr++];
+    num_face      = buffer[ctr++];
   }
 
   std::vector<char> title;
@@ -84,6 +87,8 @@ public:
   int num_elem_sets;
   int num_edge_blk;
   int num_edge;
+  int num_face_blk;
+  int num_face;
 };
 
 } // namespace libMesh
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -1344,9 +1344,18 @@ void ExodusII_IO_Helper::read_elem_in_block(int block)
   EX_CHECK_ERR(ex_err, "Error getting block info.");
   message("Info retrieved successfully for block: ", block);
 
-  const auto & conv = get_conversion(std::string(elem_type.data()));
-  const bool is_c0polygon = (conv.libmesh_elem_type() == C0POLYGON);
-  const bool is_c0polyhedron = (conv.libmesh_elem_type() == C0POLYHEDRON);
+  // Nemesis uses "Empty" as the element type for blocks with no
+  // elements on the current processor.  There is no element conversion
+  // for that sentinel type, nor is one needed for an empty block.
+  const bool is_bezier = is_bezier_elem(elem_type.data());
+  const Conversion * conversion = nullptr;
+  if (num_elem_this_blk || is_bezier)
+    conversion = &get_conversion(std::string(elem_type.data()));
+
+  const bool is_c0polygon =
+    conversion && conversion->libmesh_elem_type() == C0POLYGON;
+  const bool is_c0polyhedron =
+    conversion && conversion->libmesh_elem_type() == C0POLYHEDRON;
 
   // Warn or error when we don't currently support reading blocks with extended info.
   // Note: the docs say -1 will be returned for this but I found that it was
@@ -1362,9 +1371,11 @@ void ExodusII_IO_Helper::read_elem_in_block(int block)
   // If we have a Bezier element here, then we've packed constraint
   // vector connectivity at the end of the nodal connectivity, and
   // num_nodes_per_elem reflected both.
-  const bool is_bezier = is_bezier_elem(elem_type.data());
   if (is_bezier)
-    num_nodes_per_elem = conv.n_nodes;
+    {
+      libmesh_assert(conversion);
+      num_nodes_per_elem = conversion->n_nodes;
+    }
   else if (is_c0polygon)
     {
       elem_node_counts.resize(num_elem_this_blk);
@@ -1504,9 +1515,8 @@ void ExodusII_IO_Helper::read_elem_in_block(int block)
               // for p>2 and aren't useful for p<2, and we don't
               // support anisotropic p...
 #ifndef NDEBUG
-              const auto & conv = get_conversion(std::string(elem_type.data()));
-
-              for (auto d : IntRange<int>(0, conv.dim))
+              libmesh_assert(conversion);
+              for (auto d : IntRange<int>(0, conversion->dim))
                 libmesh_assert_equal_to(bex_elem_degrees[d], 2);
 #endif
             }
--- a/src/mesh/exodusII_io_helper.C
+++ b/src/mesh/exodusII_io_helper.C
@@ -723,6 +723,9 @@ void ExodusII_IO_Helper::open(const char * filename, bool read_only)
   if (read_only)
   {
     opened_for_reading = true;
+    elem_node_counts.clear();
+    elem_face_counts.clear();
+    c0polyhedron_face_connect.clear();
 
     // ExodusII reads truncate to 32-char strings by default; we'd
     // like to support whatever's in the file, so as early as possible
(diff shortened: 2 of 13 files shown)


src/mesh/exodusII_io_helper.C
```diff
@@ -2361,6 +2380,25 @@ void ExodusII_IO_Helper::initialize(std::string str_title, const MeshBase & mesh
 
   num_elem = n_active_elem;
   num_nodes = 0;
+  num_face = 0;
+  num_face_blk = 0;
+
+  dof_id_type local_num_c0polyhedron_faces = 0;
+  int has_c0polyhedron = 0;
```
roystgnr: Let's make this a `bool` and use `false`/`true`, for clarity?  You can still `max()` it with TIMPI.

include/mesh/exodus_header_info.h
```diff
@@ -84,6 +87,8 @@ class ExodusHeaderInfo
   int num_elem_sets;
   int num_edge_blk;
   int num_edge;
+  int num_face_blk;
+  int num_face;
```
roystgnr, line 91: These additions look like The Right Way To Do It to me, but I'll want to give @jwpeterson time to chime in before merging; IIRC the `ExodusHeaderInfo` refactor was an addition of his for Akselos usage.

src/mesh/exodusII_io_helper.C
```diff
@@ -66,6 +67,8 @@ static constexpr int libmesh_max_str_length = MAX_LINE_LENGTH;
 
 using namespace libMesh;
 
+constexpr int c0polyhedron_face_block_id = 1;
```
roystgnr: Are we hard-coding a block id here?  That definitely needs to be fixed.  It'll break as soon as someone tries to output a mesh that already has that id as a subdomain id.

There's a `build_subdomain_map` that we use to handle this problem when we've been requested to output side elements for visualization, and it might be what we want to extend for this side output too, whether because we need it or just because we really want all this hackery to be in one place so that the different optional features don't stomp on each other's toes.

src/mesh/exodusII_io_helper.C
```diff
@@ -3039,30 +3429,92 @@ void ExodusII_IO_Helper::write_elements(const MeshBase & mesh, bool use_disconti
                                << Utility::enum_to_string(conv.libmesh_elem_type())
                                << " in the same block!");
 
-          for (unsigned int j=0; j<static_cast<unsigned int>(num_nodes_per_elem); ++j)
+          if (is_c0polygon_block)
```
roystgnr, line 3509: The swath of `if`/`else if`/`else` here seems like it's leading to a lot of duplicated lines; could we use a lambda function or something to cut them down?

src/mesh/mesh_generation.C
```diff
@@ -921,8 +924,8 @@ void MeshTools::Generation::build_cube(UnstructuredMesh & mesh,
             std::vector<Node *> node_list;
 
             // Start with a layer of triangles on the boundary
-            const auto dx_tri = (xmax - xmin) / (nx);
-            const auto dy_tri = (ymax - ymin) / (ny + 1);
+            const auto dx_tri = Real(1) / nx;
+            const auto dy_tri = Real(1) / (ny + 1);
```
roystgnr, line 928: I want to whine that the mesh generation fixes should have been in a separate commit, but the important thing is that the bug got fixed; thanks!

src/mesh/mesh_generation.C
```diff
@@ -1207,8 +1213,14 @@ void MeshTools::Generation::build_cube(UnstructuredMesh & mesh,
...
           case PRISM6:
+          case C0POLYHEDRON:
             {
-              mesh.reserve_nodes( (nx+1)*(ny+1)*(nz+1) );
+              const dof_id_type grid_nodes =
+                cast_int<dof_id_type>((nx+1)*(ny+1)*(nz+1));
+              mesh.reserve_nodes(grid_nodes +
+                                 (type == C0POLYHEDRON ?
+                                  cast_int<dof_id_type>(nx*ny*nz) :
+                                  0));
```
roystgnr: The extra entries here cover the mid-polyhedron nodes that we generate to keep internal tetrahedralization robust?  Let's put that in a separate `mid_polyhedron_nodes` variable and add a comment explaining it?

roystgnr (changes requested):
I'd have split this into like 6 commits instead of 2 (re-bootstrap should be separate, tests could be separate, additional build_cube option should be separate), but the contents look mostly solid.  That hardcoded block id (unless I'm just misunderstanding?) is the only thing that looks wrong.

roystgnr:
Looks like you turned on some auto-code-reformatting option in your editor?  That works within MOOSE, so long as you're pointing clang-format to the MOOSE config file and you're using the right clang-format version, but libMesh doesn't use the same (or any clang-format config-describable) formatting standards, and auto-reformatting utterly wrecked your https://github.com/libMesh/libmesh/pull/4490/changes/ff02c5caefc12c2565d382918bbda58c045a70ae commit.  I can't find the needle of the actual changes in the 4000-line haystack of the reformatting.

As a general safety rule, never `git add some/file` or `git commit -a` without `git diff` to look at the changes first to make sure you're not committing anything you didn't mean to.  Even in repos like MOOSE where CI will block formatting changes, it's still easy to accidentally commit diagnostic code, accidentally put multiple distinct changesets into a single commit, etc.

src/mesh/exodusII_io_helper.C
roystgnr: Looks like you hit "Resolve conversation" by mistake on this one?

src/mesh/exodusII_io_helper.C
roystgnr: So the answer was "no"?  ;-)  I'd say you're right that the lambdas at least helped readability a bit, though, so let's leave them in.

src/mesh/exodusII_io_helper.C
ConnorOuellette: Whoops. Got left out when I undid the formatting and redid changes. I promise I'm reading these!

roystgnr (approved):
The elem quality updates should have gone in a separate PR, but fortunately I'm not seeing any problems with them, so we might as well get them in via this one.

Civet's not letting me activate the jobs here, though; hopefully it's an easy fix, but we'll have to wait on CI before merging.

roystgnr:
-Wshadow errors ... but the second one actually does look like poor name choice overlap.

roystgnr:
Huh.  These remaining errors are real, and there's more than one variety of them.  I can reproduce *some* kind of error; I'll try to look into it soon.

src/parallel/parallel_elem.C
```diff
@@ -300,9 +336,28 @@ Packing<const Elem *>::pack (const Elem * const & elem,
   else
     *data_out++ =(DofObject::invalid_id);
 
+  const bool has_variable_topology =
+    elem->type() == C0POLYGON || elem->type() == C0POLYHEDRON;
```
roystgnr: Let's just use `elem->runtime_topology()` here?

src/parallel/parallel_elem.C
```diff
@@ -159,9 +181,23 @@ unsigned int
 Packing<const Elem *>::packable_size (const Elem * const & elem,
                                       const MeshBase * mesh)
 {
+  unsigned int variable_topology_size = 0;
+  if (elem->type() == C0POLYGON || elem->type() == C0POLYHEDRON)
```
roystgnr: Let's just use `elem->runtime_topology()` here?

src/parallel/parallel_elem.C
```diff
@@ -73,20 +77,38 @@ Packing<const Elem *>::packed_size (std::vector<largest_id_type>::const_iterator
...
+  unsigned int n_sides = Elem::type_to_n_sides_map[type];
+  unsigned int n_edges = Elem::type_to_n_edges_map[type];
+  unsigned int variable_topology_size = 0;
 
   if (n_nodes == invalid_uint)
-    libmesh_not_implemented_msg("Support for Polygons/Polyhedra not yet implemented");
+    {
+      libmesh_error_msg_if(type != C0POLYGON && type != C0POLYHEDRON,
+                           "Unsupported variable-topology element "
+                           << Utility::enum_to_string(type));
```
roystgnr: Can we find a way to be somewhat forwards compatible here?  `if (elem->has_specific_property())` is generally much more robust in the long term than if (elem->type() == FOO || elem->type() == BAR || elem->type() == BAZ ...)`

src/parallel/parallel_elem.C
```diff
@@ -73,20 +77,38 @@ Packing<const Elem *>::packed_size (std::vector<largest_id_type>::const_iterator
...
-    Elem::type_to_n_sides_map[type];
+      auto topology_in = in + header_size;
+      n_nodes = cast_int<unsigned int>(*topology_in++);
+      n_sides = cast_int<unsigned int>(*topology_in++);
+      n_edges = cast_int<unsigned int>(*topology_in++);
+      variable_topology_size = 3;
 
-  const unsigned int n_edges =
-    Elem::type_to_n_edges_map[type];
+      if (type == C0POLYHEDRON)
```
roystgnr: Let's just check type_to_dim_map[type] and see if it's 3?

src/parallel/parallel_elem.C
```diff
@@ -159,9 +181,23 @@ unsigned int
...
 {
+  unsigned int variable_topology_size = 0;
+  if (elem->type() == C0POLYGON || elem->type() == C0POLYHEDRON)
+    {
+      // Store the dynamic node, side, and edge counts.
+      variable_topology_size = 3;
+
+      // A polygon's node ordering fully specifies its topology.  A
+      // polyhedron additionally needs each side's local node indices.
+      if (elem->type() == C0POLYHEDRON)
```
roystgnr: Likewise we can just check `elem->dim() == 3` here; even when we have more nodes per side in 2D for higher-polynomial-order polygons, it'll be a fixed number per side.

src/parallel/parallel_elem.C
```diff
@@ -300,9 +336,28 @@ Packing<const Elem *>::pack (const Elem * const & elem,
...
+    {
+      *data_out++ = elem->n_nodes();
+      *data_out++ = elem->n_sides();
+      *data_out++ = elem->n_edges();
+    }
+
   for (const Node & node : elem->node_ref_range())
     *data_out++ = node.id();
 
+  if (elem->type() == C0POLYHEDRON)
```
roystgnr: Another dim()==3

src/parallel/parallel_elem.C
```diff
@@ -516,6 +572,73 @@ Packing<Elem *>::unpack (std::vector<largest_id_type>::const_iterator in,
...
+    {
+      libmesh_error_msg_if(type != C0POLYGON && type != C0POLYHEDRON,
+                           "Unsupported variable-topology element "
+                           << Utility::enum_to_string(type));
+
+      n_nodes = cast_int<unsigned int>(*in++);
+      n_sides = cast_int<unsigned int>(*in++);
+      n_edges = cast_int<unsigned int>(*in++);
+
+      if (type == C0POLYGON)
```
roystgnr: type_to_dim_map[type]==2

src/parallel/parallel_elem.C
```diff
@@ -516,6 +572,73 @@ Packing<Elem *>::unpack (std::vector<largest_id_type>::const_iterator in,
...
+        libmesh_error_msg_if
+          (n_nodes < 3 || n_sides != n_nodes || n_edges != n_nodes,
```
roystgnr: Good instincts, but we're definitely going to want to support higher order later.  Maybe `n_sides < 3 || n_nodes % n_sides || n_edges != n_sides` to prepare for that case?

src/parallel/parallel_elem.C
```diff
@@ -516,6 +572,73 @@ Packing<Elem *>::unpack (std::vector<largest_id_type>::const_iterator in,
...
+           "Invalid packed C0POLYGON topology with "
```
roystgnr: s/C0POLYGON/2D variable/

src/parallel/parallel_elem.C
```diff
@@ -516,6 +572,73 @@ Packing<Elem *>::unpack (std::vector<largest_id_type>::const_iterator in,
...
+           << n_nodes << " nodes, " << n_sides << " sides, and "
+           << n_edges << " edges");
+      else
+        libmesh_error_msg_if
+          (n_nodes < 4 || n_sides < 4,
+           "Invalid packed C0POLYHEDRON topology with "
```
roystgnr: s/C0POLYHEDRON/3D variable/

src/parallel/parallel_elem.C
```diff
@@ -516,6 +572,73 @@ Packing<Elem *>::unpack (std::vector<largest_id_type>::const_iterator in,
...
+           << n_nodes << " nodes and " << n_sides << " sides");
+    }
+
+  const auto node_ids_in = in;
+  in += n_nodes;
+
+  std::vector<std::vector<unsigned int>> polyhedron_side_nodes;
+  if (type == C0POLYHEDRON)
```
roystgnr: dim == 3

src/parallel/parallel_elem.C
```diff
@@ -516,6 +572,73 @@ Packing<Elem *>::unpack (std::vector<largest_id_type>::const_iterator in,
   // plus the real data header
   libmesh_assert_equal_to (in - original_in, header_size + 1);
 
+  if (n_nodes == invalid_uint)
+    {
+      libmesh_error_msg_if(type != C0POLYGON && type != C0POLYHEDRON,
```
roystgnr: Likewise we can hopefully drop this

src/parallel/parallel_elem.C
```diff
@@ -516,6 +572,73 @@ Packing<Elem *>::unpack (std::vector<largest_id_type>::const_iterator in,
...
+  if (type == C0POLYHEDRON)
+    {
+      polyhedron_side_nodes.resize(n_sides);
+      std::vector<bool> node_seen(n_nodes, false);
+      unsigned int next_new_node = 0;
+      for (auto & side_nodes : polyhedron_side_nodes)
+        {
+          const unsigned int n_side_nodes =
+            cast_int<unsigned int>(*in++);
+          libmesh_error_msg_if(n_side_nodes < 3,
```
roystgnr: I'd rather use `libmesh_assert_less` (and other asserts for other use cases) exclusively in this file - we're only ever unserializing things we serialized ourselves, hopefully correctly, so in `opt` mode we should be able to safely let the compiler shrink the binary code and remove the tests here.

src/parallel/parallel_elem.C
```diff
@@ -516,6 +572,73 @@ Packing<Elem *>::unpack (std::vector<largest_id_type>::const_iterator in,
...
+                     "Packed C0POLYHEDRON side connectivity first encounters "
+                     "local node " << node << " where local node "
+                     << next_new_node << " was expected");
+                  node_seen[node] = true;
+                  ++next_new_node;
+                }
+            }
+        }
+
+      libmesh_error_msg_if
```
roystgnr: This I wouldn't mind leaving with the test as-is (but with the macro converted to an assert), if we wrap it in a separate type==C0POLYHEDRON check.  The rest of this code we can make forward compatible with future types, but I like this check and yet it's definitely C0POLYHEDRON-specific.

src/parallel/parallel_elem.C
```diff
@@ -531,15 +654,42 @@ Packing<Elem *>::unpack (std::vector<largest_id_type>::const_iterator in,
       libmesh_assert_equal_to (elem->processor_id(), processor_id);
       libmesh_assert_equal_to (elem->subdomain_id(), subdomain_id);
       libmesh_assert_equal_to (elem->type(), type);
+
+      if (type == C0POLYGON || type == C0POLYHEDRON)
```
roystgnr: We've already overwritten the invalid_uint in n_nodes by this point, but could we test for it first to set a `bool runtime_topology` to test here?

src/parallel/parallel_elem.C
```diff
@@ -772,9 +963,28 @@ Packing<Elem *>::unpack (std::vector<largest_id_type>::const_iterator in,
       // Assign the connectivity
       libmesh_assert_equal_to (elem->n_nodes(), n_nodes);
 
-      for (unsigned int n=0; n != n_nodes; n++)
-        elem->set_node (n, mesh->node_ptr
-                        (cast_int<dof_id_type>(*in++)));
+      if (type == C0POLYHEDRON)
```
roystgnr: These new tests are applicable to any element, aren't they?  After we turn the `error` macros into `assert` macros we could probably just get rid of the `if (type ==` here.

tests/mesh/exodus_test.C
```diff
@@ -157,6 +157,7 @@ public:
       exii_input.read("write_exodus_C0POLYGON.e");
 
     MeshCommunication().broadcast(input_mesh);
+    MeshSerializer input_serializer(input_mesh);
```
roystgnr: This is going to hurt our feature coverage for this case a bit.  Can we instead just replace the `elem_ptr`  call below with `query_elem_ptr`, assert that at least one rank found elem 0, and skip the tests on ranks that didn't find it?

tests/mesh/exodus_test.C
```diff
@@ -317,14 +318,17 @@ public:
...
-    const Elem & output_elem = mesh.elem_ref(0);
-    expected_nodes_on_side.reserve(output_elem.n_sides());
-    for (auto s : output_elem.side_index_range())
-      {
-        expected_nodes_on_side.emplace_back();
-        for (const auto n : output_elem.nodes_on_side(s))
-          expected_nodes_on_side.back().push_back(output_elem.node_id(n));
-      }
+    {
+      MeshSerializer output_serializer(mesh);
```
roystgnr: Likewise here; if we switch `elem_ref` to `query_elem_ptr` and so forth we shouldn't need the serializer.

tests/mesh/exodus_test.C
```diff
@@ -337,6 +341,7 @@ public:
       exii_input.read("write_exodus_C0POLYHEDRON_HEXPRISM_READ.e");
 
     MeshCommunication().broadcast(input_mesh);
+    MeshSerializer input_serializer(input_mesh);
```
roystgnr: And here `query_elem_ptr` should work.

roystgnr (changes requested):
Yes, "Request changes" at this point is greedy of me, because IMO this is now ready to merge as-is, but I still think we can make it better.

src/parallel/parallel_elem.C
ConnorOuellette: I _think_ I understand what this means
... I went ahead and changed 'C0polygon' to '2D variable topology'?

roystgnr (approved):

roystgnr:
I added some more heavyweight CI recipes, both the "we might have missed an ifdef for --disable-exodus builds" cases and the "we might have some incredibly subtle problem that only gets caught by a distributed mesh in dbg mode" cases, but assuming that all passes too, we're done here!

PR #4491, opened 2026-07-07
Cleanup output system for writing truly constant monomial data

lindsayad:
There's a decent amount of `operator==` comparisons with `FEType`, and this would lead to rejection of the new `p_refinement == false` option for `CONSTANT MONOMIAL` finite element types when determining elemental data for output. This changeset consolidates some code into `EquationSystems` for use in both exodus and nemesis. We also consolidate our definition of elemental data to always include scalar field and vector field monomials. Consistent with previous library behavior we treat `CONSTANT MONOMIAL` with `p_refinement == true` as elemental data even though it's possible that if the mesh has been p-refined this conceptually may not be the case.

### DIFF
--- a/include/systems/equation_systems.h
+++ b/include/systems/equation_systems.h
@@ -36,6 +36,7 @@
 
 // C++ includes
 #include <cstddef>
+#include <functional>
 #include <map>
 #include <set>
 #include <string>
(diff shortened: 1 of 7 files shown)


roystgnr (approved):
That conda env failure doesn't look like it could *possibly* be related...

PR #4493, opened 2026-07-08
Include XYZ as elemental data

lindsayad:


roystgnr (approved):
Test failure unrelated

PR #4494, opened 2026-07-08
Add a side hierarchic nodal soln implementation

lindsayad:
Given the previous comment here (which I've temporarily kept), I anticipate some debate on this, although I think the implementation is defensible given what we already do for projecting discontinuous solutions onto nodes for other output cases

### DIFF
--- a/src/fe/fe_side_hierarchic.C
+++ b/src/fe/fe_side_hierarchic.C
@@ -23,6 +23,7 @@
 #include "libmesh/fe.h"
 #include "libmesh/fe_interface.h"
 #include "libmesh/fe_macro.h"
+#include "libmesh/int_range.h"
 
 namespace libMesh
 {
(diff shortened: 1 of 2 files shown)


lindsayad:
If we agree on this implementation, I'd probably go and temporarily skip the affected MOOSE tests

lindsayad:
This is how `side_hiearchic_out.e` in MOOSE looks like on QUAD9 with this PR visualized via paraview. My current choice to not encode a value for element interior nodes is obvious. What do people think? I'll note that previously this gold file just had all zeroes for the `aux_side_var` and didn't even have any representation of `side_var` due to some quirks in the MOOSE output system. I'm a little surprised that @roystgnr decided this `Exodiff` test fulfilled this stated requirement, but I've certainly done worse in my time I'm sure.

> The system shall initialize SIDE_HIERARCHIC variables as instructed

### With this PR: CONSTANT SIDE_HIERARCHIC side_var

Screenshot 2026-07-08 at 1 31 08 PM

### With this PR: FIRST SIDE_HIERARCHIC aux_side_var

Screenshot 2026-07-08 at 1 31 22 PM

roystgnr:
> Given the previous comment here (which I've temporarily kept), I anticipate some debate on this

There's definitely a lot of room for debate here, and a whole spectrum of possible positions to take from "Roy was being too lazy" on the one end through "Roy was just being thoughtless" on the other, but I'm not sure I want to actually find out which one of those positions wins, and I am sure I like your solution here better.

roystgnr (approved):
Let's skip the affected tests and run CI again to be paranoid before merging; I just finally fixed that stupid openmp bug and I'm already getting used to seeing everything green again.

PR #4496, opened 2026-07-09
Work around nvc++ 26.5 bugs

roystgnr:
I'm reporting the compilation errors that prompted these upstream, but in the meantime the changes here (a couple intermediate references) seem like a reasonable workaround to me, for the first nvc++ release (literally in forever? at least since I started testing in 2022; see pull #3409) that passes all our unit tests at all optimization levels without problems.

Marking this as "Draft" until I'm done playing with example apps, MOOSE, etc. and I'm confident we don't need more libMesh-level changes.

### DIFF
--- a/tests/fe/fe_test.h
+++ b/tests/fe/fe_test.h
@@ -503,6 +503,9 @@ public:
       {
         for (auto elem : _mesh->active_element_ptr_range())
           {
+            // Workaround for nvc++ bug
+            Elem & e = *elem;
+
             const unsigned int nv = elem->n_vertices();
             const unsigned int nn = elem->n_nodes();
             // We want interiors in lower dimensional elements treated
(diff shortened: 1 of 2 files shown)


roystgnr:
You'd think "Draft" would have indicated sufficient tempering of my excitement that the sum wouldn't count as hubris, but no: I'm getting a segfault in miscellaneous_ex16.

roystgnr:
On the other hand, that might be an Eigen bug?  Debugger shows it failing deep within Eigen, and my first instinct (see if the Eigen project's own test suite passes with nvc++) was thwarted by a compilation bug there that does look like it's the fault of Eigen, not of nvc++.

roystgnr:
With Eigen disabled and #4497 to make sure everything knows it's disabled, `make check` is passing (with hybrid MPI + threading) for me with nvc++.

tests/fe/fe_test.h
```diff
@@ -503,6 +503,9 @@ class FETestBase : public CppUnit::TestCase {
       {
         for (auto elem : _mesh->active_element_ptr_range())
           {
+            // Workaround for nvc++ bug
+            Elem & e = *elem;
```
jwpeterson, line 507: Don't we have this same loop idiom in like a million places? Why does nvc++ compiler only have an issue with two of them?

jwpeterson (approved):

tests/fe/fe_test.h
roystgnr: I wish I knew!  Boiling it down to a reproducer with no red herrings (no libMesh dependencies, no CppUnit dependencies, not even libstdc++ dependencies just in case there was something dubious in headers there) was a nightmare, because there were so many seemingly innocuous layers of complexity that had to stay in otherwise the bug went away.  I did manage to end up with a reproducer that was only a few hundred lines, but in addition to this relatively-clean workaround I found like 4 other workarounds that still were applicable to the reproducer but that were even more confusing than this one.

I'd still call this a huge win, if it wasn't for that Eigen segfault I still haven't figured out the blame for.  Compilation failures are an annoyance, but much less horrific than incorrect optimizer output.

roystgnr:
MOOSE doesn't support builds without Eigen, and I'm not sure when I'll have time to dig into the Eigen issue, but a GRINS build with Eigen and VTK disabled (weird linking failures) passes all the basic (I didn't bother getting chemistry dependencies built) tests for me, also MPI+threads hybrid.

roystgnr:
Oh, and for the record the Nvidia guys were able to reproduce the problem upstream and are working on a fix.

PR #4498, opened 2026-07-11
Add "cxxflags_extra" output to example CXXFLAGS

roystgnr:
It's ridiculous that pkgconfig doesn't have a better way to do this in 2026, but here we are.

I suspect we need that -fopenmp flag to fix some linker issues in one of our conda environments, and of course grabbing stuff like -O2 doesn't hurt either.


roystgnr:
Looks like this fixes our recent `Test MOOSE conda` failures, and doesn't mess with any other tests.

PR #4502, opened 2026-07-22
Fix polyhedron edge adjacency for sparce side-edge lookups

ConnorOuellette:
## Reason

`Polyhedron::_edge_lookup` stores each physical polyhedron edge once, indexed from the side and edge where it is first encountered.

Because shared edges are omitted from later sides, stored indices for a side are not necessarily consecutive. For example, a side may contribute local edges 1 and 3 while edge 2 was already encountered on another side.

`Polyhedron::edges_adjacent_to_node()` previously advanced through this lookup while iterating every local side edge. When the current side edge had no corresponding `_edge_lookup` entry, advancing the lookup cursor could move it to a later edge. The code did not recheck the local edge index after advancing, so it could test the nodes of one edge while returning the ID of another.

This was occurred during C0POLYHEDRON element-quality test like so:

    Assertion `second_edge_node_0 == n' failed.
    second_edge_node_0 = 6
    n = 0

## Changes

- Iterate over unique entries in `_edge_lookup`.
- Use each entry's side and local side-edge index to determine its endpoint nodes.
- Reorder the C0POLYHEDRON cube fixture so a side contributes nonconsecutive local edges.

### DIFF
--- a/src/geom/cell_polyhedron.C
+++ b/src/geom/cell_polyhedron.C
@@ -444,65 +444,29 @@ Polyhedron::edges_adjacent_to_node(const unsigned int n) const
   // overridden this.
   libmesh_assert_less(n, this->n_vertices());
 
-  const unsigned int ne = this->n_edges();
-  libmesh_assert_equal_to(ne, _edge_lookup.size());
+  libmesh_assert_equal_to(this->n_edges(), _edge_lookup.size());
 
   std::vector<unsigned int> adjacent_edges;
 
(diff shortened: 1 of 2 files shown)


roystgnr (approved):
I think I was trying to be efficient via cleverness here; I was definitely failing at "clever" and I'll bet I was failing at "efficient" too.  Thanks for catching this!

roystgnr:
Looks like those last Civet tests passed but the results didn't get reported back to Github.

PR #4503, opened 2026-07-23
Skip SCALAR variables during AMR projection

BozoVazic:
# Skip SCALAR variables during AMR projection

## Summary

Exclude global `SCALAR` variables from spatial AMR projection and copy their degrees of freedom directly between the old and new solution vectors.

## Problem

`System::project_vector()` classified variables using `FEInterface::field_type()`. Both spatial scalar-valued variables and global `SCALAR` variables have `TYPE_SCALAR`, so global `SCALAR` variables were incorrectly passed to `GenericProjector`.

During fine-to-coarse projection with element-interior degrees of freedom, this caused `GenericProjector::ProjectInteriors` to call `FE<2, SCALAR>::reinit()`, which failed because global `SCALAR` variables do not have a spatial finite-element representation.

## Change

Separate variables with the `SCALAR` FE family before classifying the remaining spatial variables.

The global `SCALAR` variables are then handled only by the existing direct degree-of-freedom mapping logic.

## Testing

Added a CppUnit regression that:

* creates a refined two-dimensional mesh;
* combines a `CONSTANT MONOMIAL` spatial variable with a global `SCALAR` variable;
* performs a fine-to-coarse `EquationSystems::reinit()`;
* verifies that the global scalar value is preserved.

The test fails without this change in `FE<2, SCALAR>::reinit()` and passes with the fix.

Also ran the complete `SystemsTest` and `EquationSystemsTest` selection: 47 tests passed.

## Related report

Addresses idaholab/moose#33366 also discussed in idaholab/moose#33008

## Future possible work

Other projection paths contain similar interactions between spatial projection and explicit handling of global SCALAR variables. While working on this I also came accros issue but did not reproduce the same failure there - this is becuase I'm not sure how to reproduce them as for this PR I actually had a MOOSE input file. Thus, this PR is only to System::project_vector() issue. Other projection paths may warrant separate follow-up review e.g., project_matrix.

### DIFF
--- a/src/systems/system_projection.C
+++ b/src/systems/system_projection.C
@@ -385,10 +385,12 @@ void System::project_vector (const NumericVector<Number> & old_v,
           std::iota(vars.begin(), vars.end(), 0);
         }
 
-      std::vector<unsigned int> regular_vars, vector_vars;
+      std::vector<unsigned int> regular_vars, vector_vars, scalar_vars;
       for (auto var : vars)
       {
-        if (FEInterface::field_type(this->variable_type(var)) == TYPE_SCALAR)
+        if (this->variable(var).type().family == SCALAR)
+          scalar_vars.push_back(var);
+        else if (FEInterface::field_type(this->variable_type(var)) == TYPE_SCALAR)
           regular_vars.push_back(var);
         else
           vector_vars.push_back(var);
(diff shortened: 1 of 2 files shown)


BozoVazic:
@roystgnr let me know if this is a solid fix or do I need to dig deeper?

roystgnr (approved):
We could probably test for exact equality in the new unit test, but it's not important; almost anything that would break such an assertion will exceed the error tolerance too.

Otherwise this looks perfect; thank you!

jwpeterson:
FYI @roystgnr the new unit test here fails the `No AMR` recipe in CI:
```
../../tests/systems/systems_test.C: In member function 'void SystemsTest::testProjectScalarCoarsening()':
../../tests/systems/systems_test.C:845:40: error: variable 'libMesh::MeshRefinement mesh_refinement' has initializer but incomplete type
     MeshRefinement mesh_refinement(mesh);
                                        ^
../../tests/systems/systems_test.C:847:13: error: 'class libMesh::Elem' has no member named 'set_refinement_flag'
       elem->set_refinement_flag(Elem::REFINE);
             ^~~~~~~~~~~~~~~~~~~
../../tests/systems/systems_test.C:847:39: error: 'REFINE' is not a member of 'libMesh::Elem'
       elem->set_refinement_flag(Elem::REFINE);
                                       ^~~~~~
../../tests/systems/systems_test.C:873:13: error: 'class libMesh::Elem' has no member named 'set_refinement_flag'
       elem->set_refinement_flag(Elem::COARSEN);
             ^~~~~~~~~~~~~~~~~~~
../../tests/systems/systems_test.C:873:39: error: 'COARSEN' is not a member of 'libMesh::Elem'
       elem->set_refinement_flag(Elem::COARSEN);
                                       ^~~~~~~
```

BozoVazic:
@jwpeterson I defined AMR guard for test registration but I forgot to add the same guard for test function so that's where this is comming from. I'll fix it today. Thanks for the heads up - my libmesh experience is noob like to say at least.

jwpeterson:
No worries, the "experienced" developers make the same omission all the time as well :grimacing:

PR #4504, opened 2026-07-27
C0polygon/Polyhedron Element Quality Metrics

ConnorOuellette:
Adds element quality metrics for c0polyhedron and c0polygon block types, enabling quality evaluation for polyhedral meshes.

### DIFF
--- a/tests/geom/elem_test.C
+++ b/tests/geom/elem_test.C
@@ -7,6 +7,7 @@
 #include <libmesh/mesh_refinement.h>
 #include <libmesh/parallel_implementation.h>
 #include <libmesh/enum_to_string.h>
+#include <libmesh/elem_quality.h>
 
 using namespace libMesh;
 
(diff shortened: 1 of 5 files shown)


tests/geom/elem_test.C
```diff
@@ -95,8 +150,10 @@ public:
         if (elem->dim() > 1)
           CPPUNIT_ASSERT_GREATEREQUAL((std::acos(Real(2)/std::sqrt(Real(6))) * 180 / libMesh::pi) - TOLERANCE, min_angle);
 
-        // MIN,MAX_DIHEDRAL_ANGLE are implemented for all 3D elements
-        if (elem->dim() > 2)
+        // MIN,MAX_DIHEDRAL_ANGLE are implemented for the fixed-topology
+        // 3D elements tested here.  The generic implementation is not
+        // suitable for arbitrary C0Polyhedron faces and angles.
+        if (elem->dim() > 2 && elem->type() != C0POLYHEDRON)
```
roystgnr: Why not?

I admit I didn't write it with polyhedra in mind, but all the methods it relies on (`edge_index_range`, `sides_on_edge`, `nodes_on_side`) have polyhedral implementations, and the big assumption in the math (planar faces) is one we actually should meet!

If I get rid of this change, we're still passing the 45<min and max<90 tests.

roystgnr (changes requested):

tests/geom/elem_test.C
ConnorOuellette: True. I have no qualms in that case

roystgnr (approved):

PR #4506, opened 2026-07-29
Enable Quadratic Elements in NetGenMeshInterface

### LINKED ISSUE #4505: Support Quadratic Elements in NetGenMeshInterface
NetGen-based meshing currently only support linear elements, while quadratic element capability is needed for some 3D meshes. It is beneficial to enable the quadratic element capability for `NetGenMeshInterface` following the similar approach as used for 2D Triangulation.

miaoyinb:
closes #4505

### DIFF
--- a/src/mesh/mesh_netgen_interface.C
+++ b/src/mesh/mesh_netgen_interface.C
@@ -168,9 +168,31 @@ void NetGenMeshInterface::triangulate ()
       if (_elem_type == TET4)
         return;
 
+      // NetGen (and volume_to_surface_mesh) can leave gaps in rank 0's
+      // element and node id spaces, while broadcast() compacts them on the
+      // other ranks.  Renumber here -- on every rank, in lockstep -- so all
+      // ranks agree on n_elem()/max_elem_id() before the collective order
+      // increase runs its cross-rank consistency checks.  Note that
+      // renumber_nodes_and_elements() itself performs a collective reduction
+      // for the unique-id counter, so it must be called on all ranks together,
+      // never on rank 0 alone (that would deadlock the others in broadcast()).
+      this->_mesh.renumber_nodes_and_elements();
+
       // find_neighbors() is needed before all_second_order() can place
       // shared edge midpoints correctly.
       this->_mesh.find_neighbors();
+
+      // Refresh the cached element dimensions.  We just replaced the 2D
+      // TRI3 surface elements with 3D TET4 volume elements, but the mesh's
+      // cached dimension is still that of the original surface (2).
+      // all_second_order()/all_complete_order() derive the per-element
+      // unique_id reservation width from mesh_dimension(): a stale value of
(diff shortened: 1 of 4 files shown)


roystgnr (changes requested):
Are we supporting `TET14` here?  There's clearly some partial support (which I'd say leave in even if we don't get to total support yet), but it looks to me like we're only handling mid-edge node movement, not mid-face nodes.

If the answer is "yes", could we add a unit test?

If "no", could we take it out of the `if()` before that `libmesh_not_implemented()`?

Otherwise this looks great.

miaoyinb:
Ah, yes. I missed the face centroid of TRI7 for TET14, will add it.

miaoyinb:
Added full TET14 support

src/mesh/mesh_netgen_interface.C
```diff
@@ -168,9 +168,31 @@ void NetGenMeshInterface::triangulate ()
       if (_elem_type == TET4)
         return;
 
+      // NetGen (and volume_to_surface_mesh) can leave gaps in rank 0's
+      // element and node id spaces, while broadcast() compacts them on the
+      // other ranks.  Renumber here -- on every rank, in lockstep -- so all
```
roystgnr, line 173: Come again?  I'm fine with merging this as-is, since you're just working around this problem, not creating it, but I'd love to know what's happening exactly and get a reproducer.  If there's some conditions under which broadcast() does not result in an identical mesh on all ranks, that's a bug, and I want to get that replicated and fixed, not just worked around!

roystgnr (approved):

src/mesh/mesh_netgen_interface.C
miaoyinb: Thanks, I think we can merge this for now. Here are more details about the issue (`testNetGenQuadratic` can be a good reproducer at `np=2`)

- On rank 0, `triangulate()` deletes the input surface elements (the octahedron's 8 boundary TRIs) and adds the 4 new TETs. `delete_elem()` on a ReplicatedMesh leaves null holes in the `_elements` storage vector and does not shrink it, so rank 0 ends up with 4 live elements occupying an id space of size 8 (`max_elem_id()=8`).
- `broadcast()` on the receivers does `mesh.clear()` and then unpacks only the live entities, preserving their ids. So rank 1 gets a freshly-rebuilt, compact mesh (`max_elem_id()=4`).
- `broadcast()`'s own postcondition asserts `verify(n_elem())` / `verify(n_nodes())`, which hold on both ranks. It does not assert `max_elem_id()`, which is why the id-space-size difference slips through.

So the assertion that trips is `all_increased_order_range`'s `comm().verify(mesh.max_elem_id())`

roystgnr: Yeah, this is working great as a reproducer; I can't believe it's even tripping on a ReplicatedMesh.

PR #4507, opened 2026-07-30
Guard SCALAR projection test when AMR is disabled

BozoVazic:
Follow-up to #4503.

Guards testProjectScalarCoarsening() with LIBMESH_ENABLE_AMR so no-AMR builds compile successfully.

AMR build: test passes
No-AMR build: test is excluded

### DIFF
--- a/tests/systems/systems_test.C
+++ b/tests/systems/systems_test.C
@@ -830,6 +830,7 @@ public:
 #endif
   }
 
+#ifdef LIBMESH_ENABLE_AMR
   void testProjectScalarCoarsening()
   {
     Mesh mesh(*TestCommWorld);
@@ -887,6 +888,7 @@ public:
                                  Number(7.25),
                                  TOLERANCE);
   }
+#endif
 
   void test2DProjectVectorFE(const ElemType elem_type)
   {


jwpeterson:
FYI I invited you to join a group that gets their PRs tested automatically. I'll start these ones by hand for now.

jwpeterson:
I also attached the "Test No AMR" testing which normally only runs on the devel->master merge.

jwpeterson (approved):

BozoVazic:
Great, invitation accepted. Hopefully my local no-AMR compilation/testing will match the merge no-amr test. I'll be honest most of the local no-amr was done by AI becuase if I tried to do this myself it would take me large chunk of month of August 😄

roystgnr (changes requested):
I think we need to be wrapping the test invocation too, not just the test definition, right?

roystgnr:
Or maybe I'm wrong somehow; "Test No AMR" already passed.  Wait - are we actually invoking that test at all?

roystgnr (approved):
We already had the test invocation wrapped.  This is ready to merge.

BozoVazic:
The test invocation is already guarded by LIBMESH_ENABLE_AMR in the CPPUNIT_TEST registration. This PR adds the matching guard around the test definition that I missed.

roystgnr:
I'm just going to turn on auto-merge here, but if John wants to get devel->master passing ASAP we can just merge early; the recipes already completed are more than sufficient.

jwpeterson:
Auto merge works for me.

PR #4508, opened 2026-07-30
Polygonal/Polyhedral support for CheckpointIO

ConnorOuellette:
Adds support for Checkpoint I/O for polygonal and polyhedral elements. Writes runtime node counts for C0polygon and node counts, faces, and node indices for c0polyhedron. 

Includes some tests.

### DIFF
--- a/tests/mesh/checkpoint.C
+++ b/tests/mesh/checkpoint.C
@@ -3,6 +3,7 @@
 #include "libmesh/distributed_mesh.h"
 #include "libmesh/face_c0polygon.h"
 #include "libmesh/int_range.h"
+#include "libmesh/mesh.h"
 #include "libmesh/mesh_generation.h"
 #include "libmesh/node.h"
 #include "libmesh/parallel.h"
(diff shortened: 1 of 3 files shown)


tests/mesh/checkpoint.C
```diff
@@ -127,6 +137,162 @@ public:
...
+      return;
+
+    Parallel::Communicator comm_self;
+    const std::string filename =
+      std::string("checkpoint_c0polygon.cp") + (binary ? "r" : "a");
+    const std::vector<Point> points =
+      {{0., 0.}, {1., 0.}, {1.5, 0.5}, {1., 1.}, {0., 1.}};
+
+    {
+      ReplicatedMesh mesh(comm_self, 2);
```
roystgnr: Is this just another case where we can switch elem_ref() to query_elem_ptr() and we'll be fine, or was there some other reason to force us to be replicated here and below?

roystgnr (changes requested):
The library code all looks good to me (though in the future I may try to refactor a bit between this and the new unpacking code); all I'm suspicious about is the use of `ReplicatedMesh` rather than `Mesh` in the new tests.

roystgnr (approved):
Yeah, no changes needed to library code, and only the usual query_elem_ptr() changes plus an explicit node id setting needed to make the new tests DistributedMesh-compatible too.

roystgnr:
Turning on auto-merge, and praying I don't need to look into that creepily consistent MOOSE ARM mac recipe failure.

PR #4513, opened 2026-08-04
Fix broadcast() bug

### LINKED ISSUE #4512: Mesh broadcast() can fail post-broadcast verification
https://github.com/libMesh/libmesh/pull/4506 has a workaround for the problem, but simply commenting out the renumber call there and running the new unit test on 2 ranks (even with ReplicatedMesh!) reproduces it.

roystgnr:
And fix the underlying ReplicatedMesh::update_parallel_id_counts() that triggered it, and remove the workaround from #4506 for it.

This fixes #4512 for me.

### DIFF
--- a/src/mesh/replicated_mesh.C
+++ b/src/mesh/replicated_mesh.C
@@ -662,6 +662,15 @@ void ReplicatedMesh::update_parallel_id_counts()
   _next_unique_id = this->parallel_max_unique_id();
 #endif
 
+  // Implicitly get max_elem_id/max_node_id by trimming the vectors
+  auto trim_vec = [](auto & dof_vec) {
+    auto last_non_null = std::find_if(dof_vec.rbegin(), dof_vec.rend(), [](DofObject * d) { return (d != nullptr); });
+    dof_vec.resize(last_non_null.base()-dof_vec.begin());
+  };
+
+  trim_vec(this->_nodes);
+  trim_vec(this->_elements);
+
   this->_preparation.has_synched_id_counts = true;
 }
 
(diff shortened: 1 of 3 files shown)


roystgnr:
Hmm... I swear I tested that unit test distributed.  Maybe I did so incorrectly, or maybe I regressed something else here.

src/mesh/replicated_mesh.C
```diff
@@ -662,6 +662,15 @@ void ReplicatedMesh::update_parallel_id_counts()
   _next_unique_id = this->parallel_max_unique_id();
 #endif
 
+  // Implicitly get max_elem_id/max_node_id by trimming the vectors
+  auto trim_vec = [](auto & dof_vec) {
+    auto last_non_null = std::find_if(dof_vec.rbegin(), dof_vec.rend(), [](DofObject * d) { return (d != nullptr); });
+    dof_vec.resize(last_non_null.base()-dof_vec.begin());
```
jwpeterson, line 668: This is finding the last non-nullptr entry in the vector and then erasing everything after it? Would it make more sense to use std::vector::erase() instead of resize()?
```
dof_vec.erase(last_non_null.base(), dof_vec.end());
```

roystgnr: Yeah, that seems more straightforward.

This PR needs to wait on #4517, which at least on my system fixes the test failures for it.  I'll make that change when that goes in and I rebase.

roystgnr:
Rebased; hopefully when CI is up again we'll confirm that further exposing the now-fixed bug in devel was the only problem here.

PR #4515, opened 2026-08-04
Enable custom config for a single solve.

### LINKED ISSUE #4514: Enable sharing KSP between different linear systems.
At the moment linearimplicitsystem does not support sharing the linearsolver. 
When solve() is called it takes the absolute tolerancefrom the config object but takes default relative tolerances.
Uses its own solver by default.
Having a functionality where we override all of these at solve level would be nice.

It would make the adaptation of some convenient linear system solve functions in MOOSE within the NAvier Stokes module.
Thinking about a signature like this: solve(options).

grmnptr:
Closes #4514

### DIFF
--- a/include/systems/linear_implicit_system.h
+++ b/include/systems/linear_implicit_system.h
@@ -25,6 +25,7 @@
 
 // C++ includes
 #include <cstddef>
+#include <optional>
 
 namespace libMesh
 {
@@ -33,7 +34,32 @@ namespace libMesh
 // Forward Declarations
 template <typename T> class LinearSolver;
 template <typename T> class ShellMatrix;
+class SolverConfiguration;
 
+/**
+ * Per-call options for LinearImplicitSystem::solve().
+ */
+struct LinearImplicitSystemSolveOptions
+{
+  /**
+   * Solver to use for this call.  A null pointer selects the solver
+   * owned by the system.  The caller retains ownership of a borrowed
+   * solver and must keep it alive for the duration of the call.
+   */
+  LinearSolver<Number> * solver = nullptr;
+
+  /**
+   * Configuration to attach for this call.  When non-null, the
+   * configuration supplies rel_tol, abs_tol, and max_its.
+   */
+  SolverConfiguration * solver_configuration = nullptr;
+
+  /**
+   * Per-call assembly policy.  An unset value uses the system's
+   * persistent assemble_before_solve setting.
+   */
+  std::optional<bool> assemble_before_solve;
+};
 
 /**
  * \brief Manages consistently variables, degrees of freedom, coefficient
@@ -126,6 +152,12 @@ public:
    * Assembles & solves the linear system A*x=b.
    */
   virtual void solve () override;
+
+  /**
+   * Optionally assembles and solves the linear system A*x=b using
+   * per-call solver options.
+   */
+  void solve(const LinearImplicitSystemSolveOptions & options);
 
   /**
    * \returns A pointer to a linear solver appropriate for use in
--- a/src/systems/linear_implicit_system.C
+++ b/src/systems/linear_implicit_system.C
@@ -23,6 +23,7 @@
 #include "libmesh/linear_implicit_system.h"
 #include "libmesh/linear_solver.h"
 #include "libmesh/equation_systems.h"
+#include "libmesh/libmesh_call_mpi.h"
 #include "libmesh/numeric_vector.h" // for parameter sensitivity calcs
 //#include "libmesh/parameter_vector.h"
 #include "libmesh/sparse_matrix.h" // for get_transpose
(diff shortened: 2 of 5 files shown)


include/systems/linear_implicit_system.h
```diff
@@ -33,7 +34,32 @@ namespace libMesh
...
+   * Configuration to attach for this call.  When non-null, the
+   * configuration supplies rel_tol, abs_tol, and max_its.
+   */
+  SolverConfiguration * solver_configuration = nullptr;
+
+  /**
+   * Per-call assembly policy.  An unset value uses the system's
+   * persistent assemble_before_solve setting.
+   */
+  std::optional<bool> assemble_before_solve;
```
roystgnr, line 61: Could we move this into SolverConfiguration?  It's applicable to many different System types.

include/systems/linear_implicit_system.h
```diff
@@ -33,7 +34,32 @@ namespace libMesh
...
+   * owned by the system.  The caller retains ownership of a borrowed
+   * solver and must keep it alive for the duration of the call.
+   */
+  LinearSolver<Number> * solver = nullptr;
+
+  /**
+   * Configuration to attach for this call.  When non-null, the
+   * configuration supplies rel_tol, abs_tol, and max_its.
+   */
+  SolverConfiguration * solver_configuration = nullptr;
```
roystgnr, line 55: Since we attach this to the LinearSolver, could we get away with just making *that* the argument to the overloaded solve-with-special-options?  Or would that be too inconvenient for the use case where we want to use the standard solver but a non-standard configuration?

src/systems/linear_implicit_system.C
```diff
@@ -117,39 +193,75 @@ void LinearImplicitSystem::restrict_solve_to (const SystemSubset * subset,
...
-    linear_solver->restrict_solve_to(nullptr);
+      if (_shell_matrix)
+        // 1.) Shell matrix with or without user-supplied preconditioner.
+        rval = solver->solve(*_shell_matrix, preconditioner, *solution, *rhs, tol, maxits);
+      else
+        // 2.) No shell matrix, with or without user-supplied preconditioner.
+        rval = solver->solve(*matrix, preconditioner, *solution, *rhs, tol, maxits);
+    }
+
+  subset_guard.clear();
```
roystgnr, line 264: Why's it not sufficient to just let the destructor do this?

src/systems/linear_implicit_system.C
```diff
@@ -117,39 +193,75 @@ void LinearImplicitSystem::restrict_solve_to (const SystemSubset * subset,
...
+void LinearImplicitSystem::solve(const LinearImplicitSystemSolveOptions & options)
+{
+  LinearSolver<Number> * const solver = options.solver ? options.solver : linear_solver.get();
+  libmesh_assert(solver);
+
+#ifndef NDEBUG
+#ifdef LIBMESH_HAVE_MPI
+  int communicator_comparison;
+  libmesh_call_mpi(
+      MPI_Comm_compare(solver->comm().get(), this->comm().get(), &communicator_comparison));
```
roystgnr, line 210: Since this isn't urgent, could we get it up into TIMPI?

src/systems/linear_implicit_system.C
```diff
@@ -117,39 +193,75 @@ void LinearImplicitSystem::restrict_solve_to (const SystemSubset * subset,
...
-  linear_solver->init_systems(*this);
+  solver->init_systems(*this);
 
-  // Get the user-specified linear solver tolerance
-  const auto [maxits, tol] = this->get_linear_solve_parameters();
+  SubsetRestrictionGuard subset_guard(*solver);
 
   if (_subset != nullptr)
-    linear_solver->restrict_solve_to(&_subset->dof_ids(),_subset_solve_mode);
+    subset_guard.restrict(&_subset->dof_ids(), _subset_solve_mode);
```
roystgnr, line 236: To get cleaner RAII, could we put `_subset` in the `SubsetRestrictionGuard` constructor arguments instead, something like how the `MeshSerializer` optional arguments get used?

src/systems/linear_implicit_system.C
```diff
@@ -117,39 +193,75 @@ void LinearImplicitSystem::restrict_solve_to (const SystemSubset * subset,
...
 
   // Solve the linear system.  Several cases:
   std::pair<unsigned int, Real> rval = std::make_pair(0,0.0);
-  if (_shell_matrix)
-    // 1.) Shell matrix with or without user-supplied preconditioner.
-    rval = linear_solver->solve(*_shell_matrix, this->request_matrix("Preconditioner"), *solution, *rhs, tol, maxits);
+  SparseMatrix<Number> * const preconditioner = this->request_matrix("Preconditioner");
+
+  if (options.solver_configuration)
```
roystgnr, line 242: I want to directly test `solver->solver_configuration()` here, for less fragility.

Instead of copy-and-pasting the inner if/else to add a case without tol/maxits arguments, can we just set tol and maxits from the solver_configuration if it exists?

roystgnr (changes requested):

include/systems/linear_implicit_system.h
lindsayad: Not to nonlinear ones though

roystgnr: IIRC we originally put `assemble_before_solve` in `System` in part with the intent to use it in nonlinear systems too; "update the residual on every Newton step but update the Jacobian less frequently" actually can be a good solver heuristic for some problems, especially if you've got an expensive preconditioner.

lindsayad: If `assemble` just means Jacobian for nonlinear, then I think the name should be changed or there needs to be clear doxygen for that. Because when I think of finite element assembly, that means RHS/residual and matrix

PR #4516, opened 2026-08-04
Deprecate bad setter

roystgnr:
We got a little overzealous with the unset call in the old "pseudo-setter" in #4485.  Every read-only access on a non-const object ends up resolving to the "setter" overload instead of the const "getter" overload, and thereby calls the setter code unnecessarily.

But, we can get the same benefits as long as we restrict that call and our up-to-date users to a "real" setter.

In hindsight we should just never have accepted "setter" methods that return references rather than taking values; they're not much better encapsulated than `public:` member variables.  They look nicer but they never seem to be robust or flexible enough in these cases where we actually *need* a setter.

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -1697,11 +1697,20 @@ public:
    */
   virtual void libmesh_assert_valid_parallel_ids() const {}
 
+#ifdef LIBMESH_ENABLE_DEPRECATED
   /**
-   * \returns A writable reference for getting/setting an optional
-   * name for a subdomain.
+   * \deprecated
+   * \returns A writable reference for setting an optional name for a
+   * subdomain.  This method is deprecated; use set_subdomain_name()
+   * instead.
    */
   std::string & subdomain_name(subdomain_id_type id);
+#endif // LIBMESH_ENABLE_DEPRECATED
+
+  /**
+   * \returns A reference for getting an optional name for a
+   * subdomain.
+   */
   const std::string & subdomain_name(subdomain_id_type id) const;
 
   /**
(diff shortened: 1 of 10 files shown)


jwpeterson:
Looks like we are actually using the deprecated setter in a few of the Mesh IO classes...
```
../src/mesh/abaqus_io.C: In member function 'void libMesh::AbaqusIO::assign_subdomain_ids()':
../src/mesh/abaqus_io.C:1041:52: error: no match for 'operator=' (operand types are 'const string' {aka 'const std::__cxx11::basic_string<char>'} and 'std::__cxx11::string' {aka 'std::__cxx11::basic_string<char>'})
             the_mesh.subdomain_name(computed_id) = computed_name;
                                                    ^~~~~~~~~~~~~
```

```
../src/mesh/exodusII_io.C: In member function 'virtual void libMesh::ExodusII_IO::read(const string&)':
../src/mesh/exodusII_io.C:474:45: error: no match for 'operator=' (operand types are 'const string' {aka 'const std::__cxx11::basic_string<char>'} and 'std::__cxx11::string' {aka 'std::__cxx11::basic_string<char>'})
         mesh.subdomain_name(subdomain_id) = subdomain_name;
                                             ^~~~~~~~~~~~~~
```

```
../src/mesh/gmsh_io.C:798:81: error: no match for 'operator=' (operand types are 'const string' {aka 'const std::__cxx11::basic_string<char>'} and 'const string' {aka 'const std::__cxx11::basic_string<char>'})
                 mesh.subdomain_name(restrict_int<subdomain_id_type>(phys_id)) = phys_name;
                                                                                 ^~~~~~~~~
```

jwpeterson (approved):

PR #4517, opened 2026-08-05
Fix order elevation on unpartitioned distributed meshes

roystgnr:
There's also an underlying fix here for `update_parallel_id_counts()` on meshes in which unpartitioned elements are added out-of-sync, which is something I'd originally thought should be forbidden but which is useful enough that we should try to tolerate it, especially in cases like this where tolerating it only costs us a handful of int ops.

I've got some more robustness improvements/fixes but I'll get them in a subsequent PR; this PR is necessary to fix a regression that devel->master CI is catching and getting stuck on.

### DIFF
--- a/src/mesh/distributed_mesh.C
+++ b/src/mesh/distributed_mesh.C
@@ -261,23 +261,19 @@ void DistributedMesh::update_parallel_id_counts()
   _max_node_id = this->parallel_max_node_id();
   _max_elem_id = this->parallel_max_elem_id();
 
-  if (_next_free_unpartitioned_elem_id < _max_elem_id)
-    _next_free_unpartitioned_elem_id =
-      ((_max_elem_id-1) / (this->n_processors() + 1) + 1) *
-      (this->n_processors() + 1) + this->n_processors();
-  if (_next_free_local_elem_id < _max_elem_id)
-    _next_free_local_elem_id =
-      ((_max_elem_id + this->n_processors() - 1) / (this->n_processors() + 1) + 1) *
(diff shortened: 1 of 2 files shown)


roystgnr:
Adding the expensive debug distributed sweeps too, since those are the failures this is supposed to fix.

roystgnr:
Darn, didn't make it through all four sections of the distributed debug sweep before the CI farm outage started.  I think we're good to merge based on the 1-12 results, though.  I'll run my own sweep (on just the breaking tests) to be safe.

roystgnr:
Yeah, the new tests are good out to 64 ranks (at least) now.

src/mesh/distributed_mesh.C
```diff
@@ -261,23 +261,19 @@ void DistributedMesh::update_parallel_id_counts()
...
-    _next_free_unpartitioned_node_id =
-      ((_max_node_id-1) / (this->n_processors() + 1) + 1) *
-      (this->n_processors() + 1) + this->n_processors();
-  if (_next_free_local_node_id < _max_node_id)
-    _next_free_local_node_id =
-      ((_max_node_id + this->n_processors() - 1) / (this->n_processors() + 1) + 1) *
-      (this->n_processors() + 1) + this->processor_id();
+  _next_free_unpartitioned_elem_id =
+    ((_max_elem_id-1) / (this->n_processors() + 1) + 1) *
+    (this->n_processors() + 1) + this->n_processors();
```
jwpeterson, line 266: How could we go wrong with such a simple formula?

jwpeterson (approved):

PR #4520, opened 2026-08-11
all_tri() pyramid support, refactoring, better unit testing

### LINKED ISSUE #4495: all_tri does not support pyramids
For cavity meshing in MOOSE I am building a transition layer with pyramids then tets around a component A, to avoid tetrahedralizing any element in A (A is full-hex and used for finite volume CFD) when using the XYZ delaunay tetrahedralizaiton on a volume enclosing A.

Then I use the transition layer mesh as a hole. But because it's inside is pyramids, the mesh generator wants to call all_tri on it. 

This fails because all_tri does not support pyramids it seems:
```

Still Setting UpError, encountered unimplemented element PYRAMID5 in MeshTools::Modification::all_tri()...

                                                                         [ 16.63 s] [ 2057 MB]


*** ERROR ***
The following occurred in the main Application of type MooseTestApp.

Error: feature not implemented!
```

Implementing all_tri for pyramids here is likely the nicest solution.
The second solution would be to only call conversion routines on the side of the hole that faces the Delaunay region (and ignore the other(s))

roystgnr:
I was just going to throw a `case PYRAMID5:` in there as a quick fix for @GiudGiud's particular case, but then I noticed some atavisms, and a refactoring opportunity, and that made adding `PYRAMID14` support less annoying, and after that I got worried about the refactoring and noticed a bunch of additional assertions we could add to the unit tests, and then I noticed that the tests themselves could use a bit of refactoring...

Now this essentially fixes #4495 entirely - the remaining unsupported `PYRAMID13` case is analogous to the other serendipity `PRISM15` and `HEX20` cases that I don't intend to support until someone specifically requests them.

### DIFF
--- a/src/mesh/mesh_modification.C
+++ b/src/mesh/mesh_modification.C
@@ -953,31 +953,12 @@ void MeshTools::Modification::all_tri (MeshBase & mesh)
               //
               // Split on 0-2 diagonal
               if (split_first_diagonal(elem, 0,2, 1,3))
-                {
-                  subelem[0]->set_node(0, elem->node_ptr(0));
-                  subelem[0]->set_node(1, elem->node_ptr(1));
-                  subelem[0]->set_node(2, elem->node_ptr(2));
-                  subelem[0]->set_node(3, elem->node_ptr(4));
-
-                  subelem[1]->set_node(0, elem->node_ptr(0));
-                  subelem[1]->set_node(1, elem->node_ptr(2));
-                  subelem[1]->set_node(2, elem->node_ptr(3));
-                  subelem[1]->set_node(3, elem->node_ptr(4));
-                }
+                set_nodes({{0,1,2,4},{0,2,3,4}});
               // Split on 1-3 diagonal
               else
                 {
                   libmesh_assert (split_first_diagonal(elem, 1,3, 0,2));
-
-                  subelem[0]->set_node(0, elem->node_ptr(0));
-                  subelem[0]->set_node(1, elem->node_ptr(1));
-                  subelem[0]->set_node(2, elem->node_ptr(3));
-                  subelem[0]->set_node(3, elem->node_ptr(4));
-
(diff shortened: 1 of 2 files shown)


roystgnr:
Huh.  Regular for loops got init-statement support in C++17, but range-for didn't get it until C++20?

src/mesh/mesh_modification.C
```diff
@@ -588,36 +581,14 @@ void MeshTools::Modification::all_tri (MeshBase & mesh)
               if ((elem->point(0) - elem->point(2)).norm() <
                   (elem->point(1) - elem->point(3)).norm())
                 {
-                  subelem[0]->set_node(0, elem->node_ptr(0));
-                  subelem[0]->set_node(1, elem->node_ptr(1));
-                  subelem[0]->set_node(2, elem->node_ptr(2));
-                  subelem[0]->set_node(3, elem->node_ptr(4));
-                  subelem[0]->set_node(4, elem->node_ptr(5));
+                  set_nodes({{0,1,2,4,5},{0,2,3,3,6,7}});
```
GiudGiud, line 584:
```suggestion
                  set_nodes({{0,1,2,4,5},{0,2,3,6,7}});
```
should you assert same size too?

GiudGiud: oh you just override that set_node right below

GiudGiud (approved):

src/mesh/mesh_modification.C
roystgnr: I thought about asserting same-size as a check, but it seemed like a false sense of security - there are cases like this where same-size isn't necessary (because adding the new nodes requires a "spacer" in one simplex but not the other), and there are cases (like your proposed fix when you thought of the spacer as a typo!) where same-size just gives a false sense of confidence.

I generally prefer assertions in the code over unit tests, but I think here the unit tests are better and sufficient.

PR #4521, opened 2026-08-11
libMesh::isfinite() subroutines

roystgnr:
From https://github.com/idaholab/moose/pull/33439#discussion_r3693684525:

> in the longer term maybe should be a class method on `TypeVector`. What do you think @roystgnr ?

I think it should be a global method so future generic code can use the same `isfinite(foo)` and get `std::isfinite` via a `using` or `libMesh::isfinite` via ADL, but otherwise I agree.

@lindsayad - we should probably be adding `isfinite` to MetaPhysicL too, and `isnan` here, don't you think?  More urgently the former?  I think long ago I got into the habit of using `isnan` in cases which in hindsight should have been `isnan` 20% of the time and `isfinite` 80% of the time, simply because I formed that habit pre-C++11.

### DIFF
--- a/include/base/libmesh_common.h
+++ b/include/base/libmesh_common.h
@@ -212,6 +212,15 @@ template <typename T>
 inline bool libmesh_isinf(std::complex<T> a)
 { return (std::isinf(std::real(a)) || std::isinf(std::imag(a))); }
 
+// std::isfinite() doesn't support complex, but we really want an
+// isfinite to support Number, at least in our own namespace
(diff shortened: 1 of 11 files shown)


lindsayad:
> @lindsayad - we should probably be adding `isfinite` to MetaPhysicL too, and `isnan` here, don't you think? More urgently the former?

Yea agree!

lindsayad (approved):
LGTM!

PR #4522, opened 2026-08-13
Isnan, isinf

roystgnr:
Yeah, I know, the MetaPhysicL cases are the higher priority, but this was easy enough to do in between working on other things.

### DIFF
--- a/include/numerics/dense_matrix.h
+++ b/include/numerics/dense_matrix.h
@@ -1232,6 +1232,41 @@ bool isfinite (const DenseMatrix<T> & var)
 }
 
 
+template <typename T>
+bool isinf (const DenseMatrix<T> & var)
+{
+  using std::isinf;
+  using libMesh::isinf; // for T==complex
+  using std::isnan;
+  using libMesh::isnan;
+  const auto m = var.m(), n = var.n();
+  bool has_inf = false;
+  for (auto i : make_range(m))
+    for (auto j : make_range(n))
+      {
+        // NaN anywhere makes us NaN, not inf
+        if (isnan(var(i,j)))
+          return false;
+        has_inf = has_inf || isinf(var(i,j));
+      }
+  return has_inf;
+}
+
+
+template <typename T>
+bool isnan (const DenseMatrix<T> & var)
+{
+  using std::isnan;
+  using libMesh::isnan; // for T==complex
+  const auto m = var.m(), n = var.n();
+  for (auto i : make_range(m))
+    for (auto j : make_range(n))
+      if (isnan(var(i,j)))
+        return true;
+  return false;
+}
+
+
 } // namespace libMesh
 
 #ifdef LIBMESH_HAVE_METAPHYSICL
--- a/include/numerics/dense_matrix.h
+++ b/include/numerics/dense_matrix.h
@@ -1218,6 +1218,7 @@ T DenseMatrix<T>::transpose (const unsigned int i,
 // }
 
 
+// A matrix is finite iff every component is
 template <typename T>
 bool isfinite (const DenseMatrix<T> & var)
 {
@@ -1232,6 +1233,13 @@ bool isfinite (const DenseMatrix<T> & var)
 }
 
 
+// A matrix is infinite iff some component is infinite but no
+// component is NaN.
+//
+// This is arguably inconsistent with our std::complex overload (and
+// the C99 Annex G recommendations for _Complex, and C++ std::complex
+// arithmetic), which treats mixed (inf,NaN) pairs as infinite, but
+// this is probably safer for users.
 template <typename T>
 bool isinf (const DenseMatrix<T> & var)
 {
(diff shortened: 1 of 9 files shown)


include/numerics/dense_matrix.h
```diff
@@ -1232,6 +1232,41 @@ bool isfinite (const DenseMatrix<T> & var)
...
+  using libMesh::isinf; // for T==complex
+  using std::isnan;
+  using libMesh::isnan;
+  const auto m = var.m(), n = var.n();
+  bool has_inf = false;
+  for (auto i : make_range(m))
+    for (auto j : make_range(n))
+      {
+        // NaN anywhere makes us NaN, not inf
+        if (isnan(var(i,j)))
```
GiudGiud, line 1256: The Number template does not do what that comment says though?

include/numerics/dense_submatrix.h
```diff
@@ -243,6 +243,42 @@ bool isfinite (const DenseSubMatrix<T> & var)
...
+  using std::isnan;
+  using libMesh::isnan;
+  const auto m = var.m(), n = var.n();
+  bool has_inf = false;
+  for (auto i : make_range(m))
+    for (auto j : make_range(n))
+      {
+        // NaN anywhere makes us NaN, not inf
+        if (isnan(var(i,j)))
+          return false;
```
GiudGiud, line 268: Same comment

GiudGiud (commented):
Same comment repeated a couple times.

include/numerics/dense_matrix.h
roystgnr: How'd I'd misphrase it?  If we have a `DenseMatrix<Number>`, then `isnan(var(i,j))` invokes `isnan(std::complex<Real>)`, which returns `true` if the imaginary or real parts (or both) are `NaN`.  If we see that `true` here for any `i,j`, then we immediately return false, because this is `isinf`, and `NaN` is not infinity.

More tersely: NaN anywhere (in any real or imaginary part of any matrix entry) makes us (the whole matrix, conceptually) NaN, not inf (since NaN isn't infinity) (and so since we're in `isinf`, we return `false`).

GiudGiud: Yes this logic is correct here


But in isinf (number) you never check isNan so you don't apply the same rule

I should have just put the comment there ?

roystgnr: Oh!  Yep, `isinf(std::complex)` is wrong in the mixed inf/NaN case.  Thanks!

roystgnr: Or ... possibly it's not wrong, because my specification was wrong, and my implementation of the specification was wrong, and two wrongs made a right here?

Way back in C99 ("Annex G") they decided "A complex or imaginary number is infinite if one of its parts is infinite, even if the other part is NaN." - https://en.cppreference.com/c/numeric/complex

The rationale here is kind of specific to the complex numbers: in general the theorems on ℂ apply on the Riemann Sphere, which is basically ℂ∪{∞}, a "one-point compactification" where essentially infinity in every direction is all the same point, and even limits which don't converge in the usual metric (so e.g. they've got NaN as one component) all converge to the same ∞ in this space.

And C/C++ seem to be taking this seriously!  If I have z=a+bi, and I set a=∞, b=NaN, and evaluate w=c+di=1/z, naively you'd expect c=∞/(∞²+NaN²) and d=(0·∞-NaN)/(∞²+NaN²), which obviously would go NaN like *mad* ... but instead you get w=0+0i, exactly what you should from complex arithmetic on the Riemann sphere.  I'm reading that other languages have been following suit here too.

So I'm definitely going to want to follow the standard interpretation for `isinf(complex)` - my original implementation is right, albeit only accidentally.  The complex standard itself is an imperfect compromise (on the Riemann sphere you always have ∞+∞=NaN, but in C/C++ you get NaN for some cases and ∞ for others, apparently because they thought that complex multiplication was already costly enough for them to check for NaNs and handle them but complex addition was too cheap to mess with) but it looks like it's what people who've thought about this hard will want, unless they're specifying `-fcx-limited-range` to disable all the custom NaN handling, which apparently everybody who really cares about speed and doesn't care about infinity does (turns out complex multiplication *wasn't* already costly enough to handle non-poisoning NaNs), but in that case we don't really care what `isinf()` does either.

That was a deep rabbit hole to go down...

For isinf(vector/matrix)?  In general it looks like languages and most libraries don't have that (they require you to create a temporary and then reduce that, cringe), so we can do what we want without worrying about consistency?  And for our purposes I *think* poisoning-NaN is generally what we want for safety's sake even in `isinf` ... but this topic turned out to be *way* more complicated than I expected, so I'm open to arguments there.

In the meantime, even if my code is correct, obviously this requires *way* more commentary than I'd added.  I think of myself as a bit of a Standards Nerd and I'd never looked at C99 Annex G before today.  I'll try to make this clearer in the comments above our declarations.

jwpeterson (approved):

GiudGiud (approved):

PR #4523, opened 2026-08-16
transform_reduce algorithms

roystgnr:
I haven't created benchmarks for these specific changes in libMesh, but in some test code the results were surprisingly awesome.

In theory it's possible for a seq (default) STL reduce to be slower than a hand-coded loop; in practice I've found one case (plain integer accumulation) where g++ slows down by 20% (though clang++ gets it right), and for pretty much everything else the STL is like 50% faster or more, even in sequential mode, if only because using "reduce()" tells the compiler that we don't care about element ordering and so it can do SIMD tricks that treat FP arithmetic as associative.

I'm hoping to eventually get good enough with these algorithms to do GPGPU tricks with nvc++ in new code paths, but even before then we can probably get some great threaded and some decent serial speedup in slightly-updated existing code paths.

Edit: I'm going to restrict this PR to the serial simplification+speedup, especially since it now requires configure tests just to be sure we can use std::transform_reduce at all.  Parallelizing C++17 algorithms requires TBB with g++/clang++, so I'll need some more configure tweaking before we turn that on.

Edit 2: My tests on -O2 builds of real work are ranging from 2%-50% faster, and now I see that trying -O3 and/or -march only makes them help a bit more, but with -O0 on tiny kernels I'm seeing ~60% slowdown over hand-coded loops.  Our dbg mode is already expected to be quite slow, but if we start using these everywhere it's not going to help.

### DIFF
--- a/include/numerics/dense_matrix.h
+++ b/include/numerics/dense_matrix.h
@@ -1142,8 +1142,9 @@ auto DenseMatrix<T>::min () const -> decltype(libmesh_real(T(0)))
 {
   libmesh_assert (this->_m);
   libmesh_assert (this->_n);
+  typedef decltype(libmesh_real(T(0))) realfromT;
   return std::transform_reduce
-    (_val.begin(), _val.end(), std::numeric_limits<T>::max(),
+    (_val.begin(), _val.end(), std::numeric_limits<realfromT>::max(),
      [](const auto & a, const auto & b){using std::min; return min(a,b);},
      [](const T & v){return libmesh_real(v);});
 }
@@ -1156,8 +1157,9 @@ auto DenseMatrix<T>::max () const -> decltype(libmesh_real(T(0)))
 {
   libmesh_assert (this->_m);
   libmesh_assert (this->_n);
+  typedef decltype(libmesh_real(T(0))) realfromT;
   return std::transform_reduce
-    (_val.begin(), _val.end(), std::numeric_limits<T>::lowest(),
+    (_val.begin(), _val.end(), std::numeric_limits<realfromT>::lowest(),
      [](const auto & a, const auto & b){using std::max; return max(a,b);},
      [](const T & v){return libmesh_real(v);});
 }
--- a/include/numerics/dense_matrix.h
+++ b/include/numerics/dense_matrix.h
@@ -24,6 +24,7 @@
 #include "libmesh/libmesh_common.h"
 #include "libmesh/dense_matrix_base.h"
 #include "libmesh/int_range.h"
+#include "libmesh/parallel_algorithms.h"
 
 // For the definition of PetscBLASInt.
 #if (LIBMESH_HAVE_PETSC)
@@ -44,7 +45,6 @@
 // C++ includes
 #include <algorithm>
 #include <initializer_list>
-#include <numeric>
 #include <vector>
 
 #ifdef LIBMESH_HAVE_METAPHYSICL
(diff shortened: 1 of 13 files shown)


include/numerics/dense_matrix.h
```diff
@@ -1084,17 +1142,10 @@ auto DenseMatrix<T>::min () const -> decltype(libmesh_real(T(0)))
...
-      for (unsigned int j=0; j!=this->_n; j++)
-        {
-          auto current = libmesh_real((*this)(i,j));
-          my_min = (my_min < current? my_min : current);
-        }
-    }
-  return my_min;
+  return std::transform_reduce
+    (_val.begin(), _val.end(), std::numeric_limits<T>::max(),
+     [](auto a, auto b){using std::min; return min(a,b);},
```
GiudGiud: should we do references? in case we end up doing DenseMatrix<ADreal >

include/numerics/dense_matrix.h
```diff
@@ -1105,17 +1156,10 @@ auto DenseMatrix<T>::max () const -> decltype(libmesh_real(T(0)))
...
-      for (unsigned int j=0; j!=this->_n; j++)
-        {
-          auto current = libmesh_real((*this)(i,j));
-          my_max = (my_max > current? my_max : current);
-        }
-    }
-  return my_max;
+  return std::transform_reduce
+    (_val.begin(), _val.end(), std::numeric_limits<T>::lowest(),
+     [](auto a, auto b){using std::max; return max(a,b);},
```
GiudGiud: same

include/numerics/dense_matrix.h
roystgnr: I had to "Copy Markdown" to see `DenseMatrix<ADReal>`, but yeah, you're quite right.

roystgnr:
Well, aside from the `T=std::complex` regression, the test failures are interesting.  Looks like g++ 8.5.0 (or more specifically the libstdc++ it's bundled with) doesn't support `std::transform_reduce`, which I guess is part of the [Parallel algorithms and execution policies](https://en.cppreference.com/cpp/compiler_support/17) that didn't make it in until GCC 9.  I thought we were requiring C++17 support already but it turns out we're just requiring *most of C++17* support.

I'll add a shim as a workaround for now, I think.

include/numerics/dense_vector.h
```diff
@@ -694,15 +719,10 @@ Real DenseVector<T>::linfty_norm () const
...
-  const int N = cast_int<int>(_val.size());
-  for (int i=1; i!=N; i++)
-    {
-      Real current = TensorTools::norm_sq((*this)(i));
-      my_norm = (my_norm > current? my_norm : current);
-    }
-  return sqrt(my_norm);
+  return libmesh_transform_reduce
+    (_val.begin(), _val.end(), Real(0),
+     [](auto a, auto b){using std::max; return max(a,b);},
```
GiudGiud: refenrece? for ADreal case

roystgnr: Gah!  Good catch again.

roystgnr:
Okay, `git grep '(auto [a-zA-Z][^:]*$'` is now only finding a couple for loops where the initializer constructs an iterator; I think we're good.

GiudGiud (approved):
cannot resolve comments but looks good

roystgnr:
Turning on auto-merge, but if we get a failure that's just another one of these MOOSE `test:interaction_integral.ii_3d` CSV diffs, I'm pretty sure that's downstream and I'll just merge by hand.

PR #4528, opened 2026-08-20
Guard use of max_new_nodes_per_elem

lindsayad:
In the parameter list `max_new_nodes_per_elem` is wrapped by our unique id ifdef guards, so we must do the same for use of the variable in the function body


jwpeterson (approved):
Test failures don't seem to be related.

lindsayad:
@loganharbour are these failures a function of the new containers?

roystgnr (approved):

roystgnr:
Looks like these aren't failures from the new containers, but from interactions with some of the new coverage.sh changes that just got merged too.

PR #4529, opened 2026-08-20
Hdiv physical gradients

### LINKED ISSUE #4524: Add mapping of gradients from reference to physical space for H(div) elements
Based off the discussion in https://github.com/idaholab/moose/discussions/33545

lindsayad:
Closes #4524 and refs https://github.com/idaholab/moose/discussions/33545

### DIFF
--- a/src/fe/hdiv_fe_transformation.C
+++ b/src/fe/hdiv_fe_transformation.C
@@ -164,8 +164,8 @@ void HDivFETransformation<OutputShape>::map_dphi(const unsigned int dim,
 
     case 2:
       {
-        const std::vector<RealGradient> & dxyz_dxi   = fe.get_fe_map().get_dxyzdxi();
-        const std::vector<RealGradient> & dxyz_deta  = fe.get_fe_map().get_dxyzdeta();
+        const std::vector<RealGradient> & dxyz_dxi  = fe.get_fe_map().get_dxyzdxi();
+        const std::vector<RealGradient> & dxyz_deta = fe.get_fe_map().get_dxyzdeta();
 
         const std::vector<RealGradient> & d2xyz_dxi2    = fe.get_fe_map().get_d2xyzdxi2();
         const std::vector<RealGradient> & d2xyz_deta2   = fe.get_fe_map().get_d2xyzdeta2();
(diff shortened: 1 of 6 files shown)


joe61vette:
@lindsayad Hi Alex.  Know you are busy.  Any estimate on when this might be available?

Thanks,
   Joe

lindsayad:
I'll invalidate the jobs here now that containers should be working better. Once testing looks relatively good I'll ping Roy for review. This is not too complicated so I don't anticipate too many rounds of review, but I can't make any guarantees. Nonetheless if I had to give you a number ... 1-2 weeks

src/fe/hdiv_fe_transformation.C
```diff
@@ -19,6 +19,44 @@
...
+// map_phi and map_dphi (which also needs the physical shape function value
+// for the term coming from the derivative of J^{-1}).
+template<typename OutputShape>
+OutputShape hdiv_piola_map(const RealGradient & dx_dxi,
+                           const RealGradient & dx_deta,
+                           Real J,
+                           const OutputShape & phi_ref)
+{
+  OutputShape phi_val;
+  phi_val(0) = (dx_dxi(0)*phi_ref(0) + dx_deta(0)*phi_ref(1))/J;
```
roystgnr: Can't we do this as a single vector equation rather than component-by-component?

src/fe/hdiv_fe_transformation.C
```diff
@@ -19,6 +19,44 @@
...
+
+template<typename OutputShape>
+OutputShape hdiv_piola_map(const RealGradient & dx_dxi,
+                           const RealGradient & dx_deta,
+                           const RealGradient & dx_dzeta,
+                           Real J,
+                           const OutputShape & phi_ref)
+{
+  OutputShape phi_val;
+  phi_val(0) = (dx_dxi(0)*phi_ref(0) + dx_deta(0)*phi_ref(1) + dx_dzeta(0)*phi_ref(2))/J;
```
roystgnr: Likewise here, although in this case it doesn't save us the `#if`

roystgnr (approved):
I'm happy with this as-is, but I'd be happier if we could shave off a few lines there.

src/fe/hdiv_fe_transformation.C
```diff
@@ -108,27 +128,229 @@ void HDivFETransformation<OutputShape>::map_phi(const unsigned int dim,
...
+                  //   C =  (1/J) * SUM_n (dxi_n/dx_l) * SUM_m F_{km} * dphihat_m/dxi_n
+
+                  // Term A: -phi_k(x) * SUM_n (dxi_n/dx_l)*(1/J)*(dJ/dxi_n)
+                  const Real phik = (k == 0) ? phi_val(0) : (k == 1) ? phi_val(1) : phi_val(2);
 
-              Real dy_dxi   = dxyz_dxi[p](1);
-              Real dy_deta  = dxyz_deta[p](1);
-              Real dy_dzeta = dxyz_dzeta[p](1);
+                  // Term B: (1/J) * SUM_n (dxi_n/dx_l) * SUM_m (d^2 x_k/dxi_m dxi_n) * phihat_m
+                  const Real d2x_k_dxi2    = (k == 0) ? d2xyz_dxi2[p](0)    : (k == 1) ? d2xyz_dxi2[p](1)    : d2xyz_dxi2[p](2);
```
nmnobre:
```suggestion
                  const Real d2x_k_dxi2    = d2xyz_dxi2[p](k);
```
And likewise for the other occurrences (I think 16 in total). It also feels like `d2xk_dxi2` would be more natural, given `dphik_{...}` and `Fk_{...}`.

src/fe/hdiv_fe_transformation.C
```diff
@@ -19,6 +19,34 @@
...
+namespace {
+
+using namespace libMesh;
+
+// The contravariant Piola map phi = J^{-1} * (dx/dxi) * phihat, shared by
+// map_phi and map_dphi (which also needs the physical shape function value
+// for the term coming from the derivative of J^{-1}).
+template<typename OutputShape>
+OutputShape hdiv_piola_map(const RealGradient & dx_dxi,
+                           const RealGradient & dx_deta,
```
nmnobre:
```suggestion
OutputShape hdiv_piola_map(const RealGradient & dxyz_dxi,
                           const RealGradient & dxyz_deta,
```
And likewise for the 3d case.

src/fe/hdiv_fe_transformation.C
```diff
@@ -34,6 +62,9 @@ template<typename OutputShape>
 void HDivFETransformation<OutputShape>::init_map_dphi(const FEGenericBase<OutputShape> & fe) const
 {
   fe.get_fe_map().get_dxidx();
+#ifdef LIBMESH_ENABLE_SECOND_DERIVATIVES
+  fe.get_fe_map().get_d2xyzdxi2();
```
nmnobre, line 78: This is actually probably my bad, but do you want to use `get_d2xidxyz2` instead for consistency?

src/fe/hdiv_fe_transformation.C
lindsayad: Lol that's some egg on my face for sure

src/fe/hdiv_fe_transformation.C
lindsayad: I elected for consistency but in the opposite direction

nmnobre (approved):
I can't resolve comment threads, so you'll have to do that. :P
I want to review the math with pen and paper, mostly for the joy of it, but I won't make you wait for that. :)

nmnobre:
Ah, the only other thought that jumped to mind was: should/can we limit this to `L2_RAVIART_THOMAS` and forbid usage for `RAVIART_THOMAS`?

lindsayad:
@joe61vette explicitly wants the continuous normal `RAVIART_THOMAS` element. Normal continuity plus the requirement that the divergence of the velocity space be a subset of the pressure space gives you the result that your velocity is exactly/strongly divergence free which means that you can do things like decouple a transported scalar's space from the pressure space and still have a compatible discretization (e.g. you can do things like maintain a constant field in the transported scalar space). If you break the space (`L2_RAVIART_THOMAS`), then the only way to recover the strongly divergence free property is via hybridization, which adds more dofs and is not what @joe61vette wants. If using an RT element for Navier-Stokes, @joe61vette has to use a DG scheme regardless of whether normal continuity is broken or not in order to avoid integrating the discontinuous tangential component over faces; granting use of `RAVIART_THOMAS` is no more dangerous than granting use of `L2_RAVIART_THOMAS` in that vein

lindsayad:
I'll wait to see whether you agree with my last thought before merging

nmnobre:
I do agree.

lindsayad:
Failures unrelated. Thanks for the reviews!

PR #4530, opened 2026-08-26
Remove (inoffensive) duplicate insertion

nmnobre:


roystgnr (approved):
Well seeing that is still offensive in a personal sense!  I wrote that `semiverify(p_vui)` bug (performance bug, anyway) in 2012, and yet it survived repeated edits (one clearly manual, by me!) without me noticing it for 14 years?

I don't feel quite as bad about the 2023 didn't-even-affect-performance bug, but I'm still glad for the fix!

roystgnr:
Most of the failing recipes here are due to Logan juggling CI container naming, but "Test 32 bit" and "Test debug" could actually conceivably be related to the PR somehow ... and HA!  They are!  That double `semiverify(p_vui)` wasn't a cut-and-paste error like the double `"EDGE2"` entry was; the first is to match up with the `semiverify(&n_vars)` and the second with the `semiverify(&n_comp)` above it.

It might be a good idea to add comments there, but we definitely don't want to remove the duplicate call.

roystgnr (changes requested):
The "EDGE2" fix should be fine; the `p_vui` commit should either just be a comment explaining the correct-but-misleading code or removed entirely.

nmnobre:
My bad, clearly too eager, didn't read code appropriately. I've added a comment for clarity.

roystgnr (approved):

roystgnr:
Okay, *now* it looks like the remaining failures are all CI-container-changes related.

PR #4531, opened 2026-08-26
Link without -Werror

roystgnr:
When compiling we can `#include` our `ignore_warnings.h`, but Link-Time Optimization doesn't give us any way to selectively disable warnings-as-errors for third-party code.  We link with the C++ compiler, and Autotools helpfully (and rightfully, in cases where we're optimizing or using -fopenmp etc etc) adds our $CXXFLAGS to that, so we should get -Werror out of $CXXFLAGS.
    
This fixes library compilation in one of my environments (g++ 15, and a PETSc build that brought in a BLAS implementation triggering unused variable warnings).


roystgnr:
Test failures are CI container issues

roystgnr:
Remaining test failure is downstream

lindsayad (approved):

PR #4533, opened 2026-08-28
Revamp the DistributedMesh copy constructor

lindsayad:
It took me a fair while to understand the DistributedMesh copy constructor. I've added some comments along the way which should hopefully help the next developer. One genuine bug that I fixed was that synched_id_counts would always come out true prior to my revisions regardless of whether the incoming mesh had synched_id_counts or not due to the unconditional call to `update_parallel_id_counts()`. There is some performance improvement here as well I think in that there are no longer any communications within the copy constructor. I think there is one real change in how the next unique ids are computed when copying from a replicated mesh to a distributed mesh, but it's to use the same "round-up" trick to make sure we're into `n_processors()+1` alignment as-is done in `update_parallel_id_counts()`. Indeed both the delegate copy constructor and `update_parallel_id_counts()` now route through the same private helper method.

### DIFF
--- a/src/mesh/distributed_mesh.C
+++ b/src/mesh/distributed_mesh.C
@@ -151,28 +151,7 @@ DistributedMesh::~DistributedMesh ()
 DistributedMesh::DistributedMesh (const DistributedMesh & other_mesh) :
   DistributedMesh(static_cast<const MeshBase &>(other_mesh))
 {
-  _is_serial = other_mesh._is_serial;
-  _is_serial_on_proc_0 = other_mesh._is_serial_on_proc_0;
   _deleted_coarse_elements = other_mesh._deleted_coarse_elements;
-
-  _n_nodes = other_mesh.n_nodes();
-  _n_elem  = other_mesh.n_elem();
-  _max_node_id = other_mesh.max_node_id();
-  _max_elem_id = other_mesh.max_elem_id();
-  _next_free_local_node_id =
-    other_mesh._next_free_local_node_id;
-  _next_free_local_elem_id =
-    other_mesh._next_free_local_elem_id;
-  _next_free_unpartitioned_node_id =
-    other_mesh._next_free_unpartitioned_node_id;
-  _next_free_unpartitioned_elem_id =
-    other_mesh._next_free_unpartitioned_elem_id;
-#ifdef LIBMESH_ENABLE_UNIQUE_ID
-  _next_unique_id =
-    other_mesh._next_unique_id;
-  _next_unpartitioned_unique_id =
-    other_mesh._next_unpartitioned_unique_id;
-#endif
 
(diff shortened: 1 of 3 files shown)


src/mesh/distributed_mesh.C
```diff
@@ -209,15 +188,50 @@ DistributedMesh::DistributedMesh (const MeshBase & other_mesh) :
...
+  // reflect an incomplete local view rather than other_mesh's totals,
+  // and need to be copied afresh.
+  _n_nodes = other_mesh.n_nodes();
+  _n_elem = other_mesh.n_elem();
+  _max_node_id = other_mesh.max_node_id();
+  _max_elem_id = other_mesh.max_elem_id();
 
+  // If other_mesh is actually a DistributedMesh, we can just copy its raw
+  // counters directly, mirroring other_mesh's own bookkeeping exactly
+  if (const DistributedMesh * other_dist_mesh = dynamic_cast<const DistributedMesh *>(&other_mesh))
```
roystgnr: Any way we can do this without RTTI?  Our `-fno-rtti` compatibility status is "probably not 100% working, because we only check every couple years", but "will never work with DistributedMesh again" would still be a major downgrade.

roystgnr (changes requested):

src/mesh/distributed_mesh.C
lindsayad: The first thing that came to my mind was adding an `is_distributed_type = false` defaulted parameter to this delegate copy constructor and using that to toggle whether to call the "recompute these data members functions." And then for the `is_distributed_type` case, leaving the setting of those data members to the actual `DistributedMesh` copy constructor initializer list

roystgnr (approved):

roystgnr:
Test failures are unrelated

src/mesh/distributed_mesh.C
roystgnr: Works for now.  I think at some point I'll want to replace it but at the moment I can't think of anything better.

roystgnr:
Well, the failures in the tests we *ran* were unrelated, anyway.  As of the merge commit here ( 8fa69b835407ac5f8ab6f2697617b5f895ad6bef ) we're failing devel→master testing, because those run the DistributedMesh-in-debug-mode recipes that I obviously should have activated here first, and some of the unit tests on those are failing.

Looks like `LIBMESH_RUN='mpiexec -n 2' LIBMESH_OPTIONS='--re testVariationalEdge2'` is enough to trigger one of the failures; I'm looking into it now.

lindsayad:
I'm happy to look into it as well if you want me to

roystgnr:
Oh, I forgot to link back here.  Much simpler bug than I was fearing; https://github.com/libMesh/libmesh/pull/4545 fixed it.

PR #4535, opened 2026-08-31
Prevents circumscribing other nodes in the triangulation

roystgnr:
Rebased #4483


roystgnr:
Test failures are unrelated.

Which is quite the surprise, but it looks like Logan fixed the actually-breaking downstream test when his `-march` changes triggered it too.

PR #4536, opened 2026-09-01
Repair the dual basis on QUAD8 and TRI6 faces

Andreas-Lepak:
`compute_dual_shape_coeffs` builds the dual basis from `D = diag(\int N_k)`, but that weight is
exactly 0 at a TRI6 vertex -- leaving those vertex dual shape functions identically zero -- and -1/3
at a QUAD8 corner. Neither is diagnosed.

This biorthogonalizes against the locally quadratic transformed basis of Popp et al. (2012, Sec.
4.4.1) on those two element types, which reduces to a sparse update of `D` with no extra solve. The
transform is the identity everywhere else and `alpha = 0` reproduces `diag(d)` exactly, so nothing
else changes, there is nothing to regold, and no API changes.

New QUAD8/TRI6 cases in `dual_shape_verification_test` fail without the fix. Full unit suite passes;
MOOSE first-order dual mortar output is bit-identical before and after.

Refs idaholab/moose#33558

### DIFF
--- a/include/fe/fe_base.h
+++ b/include/fe/fe_base.h
@@ -208,6 +208,27 @@ public:
   { libmesh_assert(!calculations_started || calculate_phi);
     calculate_phi = true; return phi; }
 
+  /**
+   * \returns The dual (biorthogonal) shape function values at the quadrature
+   * points on the element.
+   *
+   * The dual basis is constructed to satisfy
+   * \f$ \int \Phi_j N_k \, d\gamma = \delta_{jk} \int N_k \, d\gamma \f$
+   * against the primal basis \f$ N \f$. That biorthogonality is what makes a
+   * mortar coupling matrix diagonal, so a Lagrange multiplier can be condensed
+   * with a diagonal inverse.
+   *
+   * \note On QUAD8 and TRI6 these functions are biorthogonal to a locally
+   * transformed primal basis rather than to \f$ N \f$ itself: \f$ \int N_k \f$
+   * is not positive on those faces (exactly 0 at a TRI6 vertex, -1/3 at a QUAD8
+   * corner), so the usual construction degenerates. For those two element types
+   * \f$ \int \Phi_j N_k \f$ is therefore not diagonal; it factors as
+   * \f$ D = \tilde{D} T^{-1} \f$ with \f$ \tilde{D} \f$ diagonal, leaving
+   * \f$ D^{-1} = T \tilde{D}^{-1} \f$ sparse and cheap to apply. The functions
+   * returned here are still expressed in the primal basis, so the type and
+   * meaning of this return value are unchanged. See Popp, Wohlmuth, Gee and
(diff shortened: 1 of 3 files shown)


roystgnr:
I've started up CI.

Walk me through the math on this, and make sure I'm not misunderstanding?  The standard dual basis on these elements can't be made to satisfy both biorthogonality and the partition-of-unity property at once without going degenerate, so we do a slight change-of-basis on the primal bases and then we find a dual to that changed basis instead?  How does that work at the application level, when we no longer have a dual to the original basis but we aren't exposing the altered basis?

I think I'm okay with this, under the circumstances (though it pains me to see a non-serendipity element behaving as badly and requiring as much of a hack as a serendipity!), but we've got to document what's going on in the headers, not just the implementation.  Somehow I was an idiot and approved the original addition of `FEBase::get_dual_phi()` without a Doxygen comment, but let's fix that now.  "get_dual_phi()" at least used to do what it said on the tin, but if we want "get_dual_phi()" to actually mean "get_dual_phi_but_maybe_not_our_dual_just_close_enough()" then we're going to need a comment to warn users of that.

Andreas-Lepak:
Partition of unity isn't the obstruction. In §4.4.1 of [(Popp, Wohlmuth, Gee, Wall)](https://epubs.siam.org/doi/10.1137/110848190) the authors note the transformation is chosen to preserve it. What fails is their integral positivity condition, Eq. (4.2): ∫Φ_j dγ > 0. Eq. (4.3) gives < 0 at a quad8 corner and Eq. (4.4) gives exactly 0 at a tri6 corner, where they say biorthogonality "does not even make sense". So tri6 is the genuinely degenerate case; quad8 just has the wrong sign.
The application side is Eq. (4.12): `D = D̃T⁻¹,` so `D` itself isn't diagonal, but `D⁻¹ = TD̃⁻¹` stays trivial and sparse but you just can't assume diagonality. Nothing new is exposed: the transformed basis never leaves compute_dual_shape_coeffs.

Andreas-Lepak:
Done in bc22f267.

`get_dual_phi()` now carries the Doxygen block, with a `\note` that on quad8/tri6 it is biorthogonal
to a transformed primal basis rather than to `N`, so `∫Φ_j N_k` is not diagonal there, along with the
`D = D̃T⁻¹` factorization that keeps `D⁻¹` sparse. It turned out none of `get_dual_dphi()`,
`get_dual_d2phi()` or `get_dual_coeff()` had comments either, so those get short `\returns` lines
pointing at it. I left `request_dual_phi()`/`request_dual_dphi()` alone, since the primal
`request_phi()`/`request_dphi()` are undocumented too and I didn't want to introduce a new asymmetry.

Also dropped "serendipity" where I had wrongly applied it to tri6, and switched the implementation
comment to the paper's own "integral positivity" wording with Eqs. (4.2)-(4.4).

Full unit suite still passes locally.

*(generated by Claude)*

roystgnr:
Test failures are a known downstream bug

PR #4537, opened 2026-09-01
Nonlinear solver generalizations

lindsayad:
I want to use MATSHELL for the Jacobian and possibly MATAIJ for the preconditioning matrix so need to develop some additional support for that use case

### DIFF
--- a/include/numerics/petsc_mffd_matrix.h
+++ b/include/numerics/petsc_mffd_matrix.h
@@ -48,7 +48,15 @@ public:
 
   explicit PetscMFFDMatrix(const Parallel::Communicator & comm_in);
 
-  PetscMFFDMatrix & operator=(Mat m);
+  /**
+   * Adopt an existing, externally-owned Mat, without destroying it when this
+   * object goes out of scope. \p set_context controls whether we attach a
+   * context pointer to \p m allowing \p get_context() to recover this object
+   * from the Mat later; skip this when this wrapper is short-lived (e.g. a
+   * function-local variable) so we don't leave a dangling context on \p m
+   * after we're destroyed.
+   */
+  void assign(Mat m, bool set_context);
 
   virtual void init(const numeric_index_type,
                     const numeric_index_type,
@@ -101,11 +109,14 @@ PetscMFFDMatrix<T>::PetscMFFDMatrix(const Parallel::Communicator & comm_in)
 }
 
 template <typename T>
-PetscMFFDMatrix<T> &
-PetscMFFDMatrix<T>::operator=(Mat m)
+void
+PetscMFFDMatrix<T>::assign(Mat m, bool set_context)
 {
   this->_mat = m;
-  return *this;
+  this->_is_initialized = true;
+  this->_destroy_mat_on_exit = false;
+  if (set_context)
+    this->set_context();
 }
 
 template <typename T>
--- a/include/numerics/petsc_mffd_matrix.h
+++ b/include/numerics/petsc_mffd_matrix.h
@@ -48,6 +48,8 @@ public:
 
   explicit PetscMFFDMatrix(const Parallel::Communicator & comm_in);
 
+  PetscMFFDMatrix & operator=(Mat m) { this->assign(m, false); return *this; }
+
   /**
    * Adopt an existing, externally-owned Mat, without destroying it when this
    * object goes out of scope. \p set_context controls whether we attach a
--- a/include/systems/nonlinear_implicit_system.h
+++ b/include/systems/nonlinear_implicit_system.h
@@ -257,9 +257,19 @@ public:
   virtual void reinit () override;
 
   /**
-   * Assembles & solves the nonlinear system R(x) = 0.
+   * Assembles & solves the nonlinear system R(x) = 0. If a matrix has been registered via
+   * set_operator_matrix(), it is used as the actual Jacobian operator (Amat) while \p matrix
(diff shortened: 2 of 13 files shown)


lindsayad:
Still waiting on that MOOSE master merge to get those solid mechanics test fixes!

include/numerics/petsc_mffd_matrix.h
```diff
@@ -48,7 +48,15 @@ class PetscMFFDMatrix : public PetscMatrixBase<T>
 
   explicit PetscMFFDMatrix(const Parallel::Communicator & comm_in);
 
-  PetscMFFDMatrix & operator=(Mat m);
```
roystgnr, line 51: Let's keep this around for backwards compatibility?  Pretty specialized so you can deprecate it if you want, but it's been around for two years and was never marked `libmesh_experimental()`.

src/numerics/petsc_matrix_shell_matrix.C
```diff
@@ -54,6 +54,116 @@ PetscMatrixShellMatrix<T>::init(ParallelType libmesh_dbg_var(type))
   this->set_context();
 }
 
+template <typename T>
+void
+PetscMatrixShellMatrix<T>::zero()
+{
+  libmesh_error();
```
roystgnr: Could we `libmesh_error_msg("Called a method that is unimplemented for arbitrary shell matrices")` or something, here and elsewhere in the new overloads?

include/systems/nonlinear_implicit_system.h
```diff
@@ -335,6 +345,12 @@ class NonlinearImplicitSystem : public ImplicitSystem
    * The final residual for the nonlinear system R(x)
    */
   Real _final_nonlinear_residual;
+
+  /**
+   * An optional matrix to use as the actual Jacobian operator (Amat), distinct from \p matrix
+   * (used as the preconditioning matrix, Pmat). See set_operator_matrix().
```
roystgnr, line 351: This seems like the opposite of what I would have expected.  `implicit_system.h` refers to its matrix as the "system matrix" in solves of "Ax=b" at one point, and as the Jacobian matrix at another.  Shouldn't that stay "A", and "P" be the new matrix?

roystgnr (changes requested):

include/numerics/petsc_mffd_matrix.h
lindsayad: Should the backwards compatibility include its undesirable behavior like `_destroy_on_exit = true`? 😢

include/systems/nonlinear_implicit_system.h
lindsayad: For a nonlinear solver, it is very common to unseat the A matrix and replace it with matrix-free approximations of the Jacobian action via finite differencing of residual evaluations, whereas it is much less common to unseat the preconditioning matrix because you generally need an explicit matrix representation in order to form a preconditioner. Put another way, if we were to assert that the libmesh system matrix is the A matrix, that assertion would fail for perhaps more than 50% of MOOSE simulations because of use of PJFNK and `-snes_mf_operator` whereas if you were to assert that the libmesh system matrix is the P matrix then I'm not aware of any places where that assertion would fail. But I'm sure you could point me to a libmesh example that would fail 😄

lindsayad: I want to do something similar as PETSc matrix-free finite-differencing, but use `MATSHELL` instead of `MATMFFD` where the shell matrix is owned by libMesh/MOOSE, not created and held by PETSc as the latter is done. Our shell matrix will define operations like `MatGetDiagonal` and `MatMult`

src/numerics/petsc_matrix_shell_matrix.C
lindsayad: done

roystgnr (approved):
I think you've convinced me on this, but it's a big enough design decision that I'd like to give @jwpeterson some time to look too

include/numerics/petsc_mffd_matrix.h
jwpeterson: Did you want to deprecate this like @roystgnr suggested? Also, any reason that it became inlined when it wasn't before?

include/systems/nonlinear_implicit_system.h
jwpeterson: It might be good if some of this discussion/clarification made it into a comment in the code regarding the colloquial naming conventions surrounding "Jacobian", "system", "Pmat" and "Amat".

jwpeterson (approved):
If I understand correctly, this is about adding better MATSHELL support for the nonlinear solver, which sounds useful. I also like that you added several tests.

include/numerics/petsc_mffd_matrix.h
lindsayad: I'll remove the inlining which will decrease the PR diff. I don't think I'll deprecate at this point because I think the new `operator=` behavior (through `assign`) is what we want

include/systems/nonlinear_implicit_system.h
lindsayad: Ok I've added discussion in https://github.com/libMesh/libmesh/pull/4537/commits/01c924a942924103eea22f757bf3b07b79405a01 but I'm not 100% sure this is the right place nor whether it helps reduce confusion

roystgnr:
I think we're good here; I'll merge tomorrow unless @lindsayad has anything else to add or @jwpeterson has any suggestions on the discussion blurb.

lindsayad:
thanks for the reviews!

PR #4538, opened 2026-09-02
Refspace fixes

roystgnr:
In one of @rochi00's branches his new test coverage discovered some utterly broken master point coordinate definitions for Pyramid13, Pyramid14, and especially Pyramid18.  I fixed those, added some tests for the fix, and it turns out Tri7/Prism20/Prism21 also had issues.  All fixed and in test coverage now.


roystgnr:
I think we may have a threading race condition problem with `ReferenceElem::get()`, despite doing a bunch of work in `reference_elem.C` to avoid that...

src/geom/cell_pyramid.C
```diff
@@ -44,11 +44,15 @@ const Real Pyramid::_master_points[14][3] =
...
-    {0, 0, 0}
+    {-0.5, -0.5, 0.5},
+    {0.5, -0.5, 0.5},
+    {0.5, 0.5, 0.5},
+    {-0.5, 0.5, 0.5},
+    {0, 0, 0},
+    {0, -Real(2)/3, Real(1)/3},
+    { Real(2)/3, 0, Real(1)/3},
+    {0,  Real(2)/3, Real(1)/3},
+    {-Real(2)/3, 0, Real(1)/3}
```
jwpeterson, line 55: Hmm... I don't think I realized that the base class's `_master_points` array was meant to have entries for all possible subtypes that might exist? Maybe we should document that better somehow?

jwpeterson (approved):

src/geom/cell_pyramid.C
roystgnr: I'll add docs for that (to this PR, which will end up being a "just consolidate reference point definitions" after I rebase it post-#4539 and fix the race condition), yeah ... and in the comments in the .C I'll probably explicitly mention the types that are currently supported, because IIRC I tried to make sure I was finding all the code that needed changing by e.g. searching for "Pyramid13" and "PYRAMID13", and those searches totally missed these spots.

PR #4539, opened 2026-09-02
Just refspace fixes

roystgnr:
I thought my consolidation in https://github.com/libMesh/libmesh/pull/4538 was very clever, but multithreaded recipes disagree, so let's just see if we can at least merge the fixes and new unit tests without it.


roystgnr:
Shoot, did I get the restriction of those unit tests into the wrong commit?

rochi00 (commented):
The fixes for the master_points are correct. The values inside get_refspace_nodes() are also correct.

roystgnr:
Test failure is downstream and unrelated

PR #4540, opened 2026-09-08
Assert caches are prepared

roystgnr:
Pulling these out of #4411, partly because Alex was right that we really should have assertions like this and so I want to get them in *before* someone adds a new violation of them, partly because I'm hoping that if I keep stripping #4411 down smaller and smaller then I'll finally be able to get the whole thing merged by fixing the problems as they come.


roystgnr:
I might need to pull a third commit from #4411 before this will pass tests; fingers crossed.

GiudGiud (approved):

PR #4541, opened 2026-09-08
find_neighbors() check_non_remote=false option

roystgnr:
This fixes a few corner cases where a ghosted element should see another ghosted element but, depending on how repartitioning was done, might only see a remote_elem link.  The fix shouldn't affect simulation codes that only work via local-ghost interactions, but I need it to be able to allow for more hardcore preparation assertions.

Stripped out of #4411 both because it really is an independent fix and because it's the most likely culprit for the distributed-mesh-dbg-mode failures that I see in CI there but can't replicate.

### DIFF
--- a/include/mesh/mesh_base.h
+++ b/include/mesh/mesh_base.h
@@ -986,10 +986,17 @@ public:
    *
    * If \p assert_valid is left as true, then in dbg mode extensive
    * consistency checking is performed before returning.
+   *
+   * If \p check_non_remote is set to false, then only sides which
+   * currently have remote neighbors are checked for possible local
+   * neighbors.  This is intended to handle a corner case where
+   * ancestor neighbors are redistributed to a processor only by other
+   * processors who do not see that neighbor link.
    */
   virtual void find_neighbors (const bool reset_remote_elements = false,
                                const bool reset_current_list    = true,
-                               const bool assert_valid          = true) = 0;
+                               const bool assert_valid          = true,
+                               const bool check_non_remote      = true) = 0;
 
   /**
    * Removes any orphaned nodes, nodes not connected to any elements.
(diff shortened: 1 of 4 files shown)


roystgnr:
Test failure from unrelated downstream bug

include/mesh/mesh_base.h
```diff
@@ -986,10 +986,17 @@ class MeshBase : public ParallelObject
    *
    * If \p assert_valid is left as true, then in dbg mode extensive
    * consistency checking is performed before returning.
+   *
+   * If \p check_non_remote is set to false, then only sides which
+   * currently have remote neighbors are checked for possible local
+   * neighbors.  This is intended to handle a corner case where
```
GiudGiud, line 992: so it's an optimization in a way? you know you're only missing neighbors on the remote neighbor side?

by local neighbor do you mean ghosted?

by neighbor do you mean side-neighbor consistently or could some be point neighbors? (e.g. a side neighbor is remote but some of the point neighbors are ghosted)

roystgnr: Pure optimization IIRC; the extra call that uses this was an added cost but I figured I could get it down to O(N^((d-1)/d)) rather than O(N).

I definitely meant "ghosted"; I'll fix that.

I meant "side-neighbor" consistently, which I'm okay with since that's our standard "neighbor" meaning, but since I'm having to fix the comment anyway I'll clarify this too.

GiudGiud (approved):

include/mesh/mesh_base.h
GiudGiud: you prefer semilocal to ghosted?

roystgnr: I've been going back and forth.  In theory thinking about it now I only expect ghosted matches, and "ghosted" would be more precise; in practice it's been months since I hit the bug this fixes and I can't swear it wasn't a newly-local element that was left unmatched and I'd rather be imprecise than possibly-inaccurate.

PR #4542, opened 2026-09-08
Add dependencies of constrained ghost DOFs to send list

NamjaeChoi:
In Kokkos-MOOSE, we cannot use `VecSetValues` and `MatSetValues` which can dynamically cache remote entries. Instead, we allocate ghost entries in advance based on libMesh `send_list` and accumulate to those, and do the assembly at once. When constraints exist, we might have to add residual/Jacobian contributions to the dependencies of ghost DOFs. So additional ghosting is required.

### DIFF
--- a/src/base/dof_map_constraints.C
+++ b/src/base/dof_map_constraints.C
@@ -4495,7 +4495,7 @@ void DofMap::process_constraints (MeshBase & mesh)
 
   // Now that we have our root constraint dependencies sorted out, add
   // them to the send_list
-  this->add_constraints_to_send_list();
+  this->add_constraints_to_send_list(mesh);
 }
 
 
@@ -5308,7 +5308,7 @@ void DofMap::gather_constraints (MeshBase & /*mesh*/,
     }
 }
 
-void DofMap::add_constraints_to_send_list()
+void DofMap::add_constraints_to_send_list (const MeshBase & mesh)
 {
   // This function must be run on all processors at once
   parallel_object_only();
@@ -5344,7 +5344,7 @@ void DofMap::add_constraints_to_send_list()
 
   // If we only need constraint DoFs constraining DoFs which are
   // algebraically local, we're done.
-  if (!this->has_static_condensation())
+  if (!this->has_static_condensation() && !_need_ghost_constraints)
     return;
 
   // If we use StaticCondensation, though, we may need constraint DoFs
@@ -5352,10 +5352,16 @@ void DofMap::add_constraints_to_send_list()
   // node) but which are supported on local elements.  Let's get those
   // too if we have to.
   //
+  // Kokkos-MOOSE also requires that, because it cannot use VecSetValues
+  // and MatSetValues which can dynamically cache remote entries. Instead,
+  // it allocates ghost entries in advance based on libMesh send_list and
+  // accumulates to those, and does the assembly at once. When ghost DOFs
+  // are constrained, it has to add residual/Jacobian contributions to the
+  // dependencies of ghost DOFs locally as well.
+  //
   // We'll potentially be hitting the same constrained DoFs from
   // multiple directions.
-  std::unordered_set<dof_id_type> extra_dependencies;
-  for (auto & elem : this->get_static_condensation().mesh().active_local_element_ptr_range())
+  for (auto & elem : mesh.active_local_element_ptr_range())
     {
       std::vector<dof_id_type> di;
       this->dof_indices (elem, di);
--- a/include/base/dof_map.h
+++ b/include/base/dof_map.h
@@ -270,9 +270,16 @@ public:
   void set_constrained_sparsity_construction(bool use_constraints);
 
   /**
-   * Sets need_full_sparsity_pattern to true regardless of the requirements by matrices
+   * Sets _need_full_sparsity_pattern to true regardless of the
+   * requirements by matrices
    */
   void full_sparsity_pattern_needed();
+
+  /**
+   * Sets _need_ghost_constraints to true regardless of the requirements
+   * by static condensation
+   */
+  void ghost_constraints_needed();
 
   /**
    * Returns true iff the current policy when constructing sparsity
@@ -2066,7 +2073,7 @@ private:
    * which are dependencies for constraint equations on the current
    * processor.
    */
-  void add_constraints_to_send_list();
+  void add_constraints_to_send_list(const MeshBase & mesh);
 
   /**
    * Adds any spline constraints from the Mesh to our DoF constraints.
@@ -2242,7 +2249,13 @@ private:
    * Default false; set to true if any attached matrix requires a full
    * sparsity pattern.
    */
-  bool need_full_sparsity_pattern;
+  bool _need_full_sparsity_pattern;
+
+  /**
+   * Default false; set to true if the dependencies of constrained ghost
+   * DOFs supported by local elements should also be ghosted
+   */
+  bool _need_ghost_constraints;
 
   /**
    * The sparsity pattern of the global matrix.  If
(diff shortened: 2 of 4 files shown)


src/base/dof_map_constraints.C
```diff
@@ -5337,11 +5337,18 @@ void DofMap::add_constraints_to_send_list()
...
   };
 
+  auto send_list = _send_list;
+
   // We usually only need dependencies of our own constrained dofs
   for (const auto & [constrained_dof, constraint_row] : _dof_constraints)
     if (this->local_index(constrained_dof))
       add_row(constraint_row);
 
+  // But sometimes we need dependencies of our ghosted dofs
```
GiudGiud: let's explain more

NamjaeChoi: Okay to have a comment about its downstream in upstream?

src/base/dof_map_constraints.C
```diff
@@ -5337,11 +5337,24 @@ void DofMap::add_constraints_to_send_list()
...
+  // Kokkos-MOOSE cannot use VecSetValues and MatSetValues which can dynamically
+  // cache remote entries, because there is no device support for those APIs.
+  // Instead, they allocate ghost entries in advance based on libMesh send_list
+  // and accumulate to those, and do the assembly at once. When ghost DOFs are
+  // constrained, they might have to add residual/Jacobian contributions to the
+  // dependencies of ghost DOFs locally. So additional ghosting is required.
+  for (const auto ghost_dof : send_list)
+    if (_dof_constraints.count(ghost_dof))
+      add_row(_dof_constraints[ghost_dof]);
```
roystgnr: This code doesn't seem to do what the comment says.  It's adding dependencies of dependencies (but not dependencies of dependencies of dependencies! there are corner cases where those show up!) but not dependencies of all ghosted DoFs.

I'm also a little surprised it's doing anything.  I think a decade or so ago we started expanding our constraint rows in `process_constraints()` to try to be more efficient in subsequent constraint applications, and after that expansion we shouldn't really *have* dependencies of dependencies in `_dof_constraints` because we've already collapsed a→b,b→c into a→c.

Could you add a unit test where (at least in parallel runs) we can see what's going wrong and verify that the fix doesn't regress later?

roystgnr (changes requested):

src/base/dof_map_constraints.C
NamjaeChoi: Dependencies of only constrained ghost DOFs are what I intend. I don't have a case to demonstrate this yet.

NamjaeChoi: If a->b, b->c was collapsed to a->c, c is already in `send_list`, right?

NamjaeChoi: The ghost DOFs might not be dependencies of another

roystgnr: > Dependencies of only constrained ghost DOFs are what I intend.

This won't add all dependencies of constrained ghost DoFs, only of those which are also constraints of constrained local DoFs.

> I don't have a case to demonstrate this yet.

It'll only be a matter of time before something regresses it, then.  We could beat the rush by just not merging this.  ;-)

Can you cook up a unit test by manually adding constraint rows?

Do you have a MOOSE input that I could use to replicate the problem?

> If a->b, b->c was collapsed to a->c, c is already in send_list, right?

If `a` is a local DoF, then AFAIK yes, it should be.  If it's not then I want to figure out why.

NamjaeChoi: > This won't add all dependencies of constrained ghost DoFs, only of those which are also constraints of constrained local DoFs.

If I put the ghost DOF as the key of `_dof_constraints`, don't I get its constraints, if there is any? Not sure why I only get "constraints of constrained local DoFs"

NamjaeChoi: Let's say G is algebraic ghost and H is a remote DOF which is not algebraic ghost. H is the constraint of G. I want H to be in ghost as well. How do I do?

roystgnr: If you put `ghost_dof` as the key, you get **its** constraints.  But you're only looping over `ghost_dof : send_list`, so you're only getting ghost DoFs that are already in the send list ... and I'm being an idiot here.  I was thinking about the cases of ghosted elements+nodes that aren't algebraically ghosted, but you don't care about those, do you, you only care about ghosted DoFs.  This would be fine for that.

This is kind of a sweeping change, though.  In the presence of constraints it adds more bandwidth to basically every algebraic communication step on every ghosted vector, right?

I've still got questions:

Could we add a flag for this, only to be enabled by downstream code that needs it?

Do you really need every ghosted DoF, not just e.g. the ones that are supported on local elements (as in the static-condensation case below)?

If we were to merge this as-is, wouldn't it have made the static-condensation code path redundant?

Do you need a hand cooking up some test coverage for this?

NamjaeChoi: > but you don't care about those, do you, you only care about ghosted DoFs.

That is correct.

> Do you really need every ghosted DoF, not just e.g. the ones that are supported on local elements (as in the static-condensation case below)?

That is also correct (only need locally-supported ghosted DOFs)

> If we were to merge this as-is, wouldn't it have made the static-condensation code path redundant?

I was unsure about the following code
```
this->get_static_condensation().mesh().active_local_element_ptr_range()
```
It read like it has a separate mesh. If `mesh()` is the mesh we are operating on, I think the code is exactly what I want.

I will try to set up a flag and make a test.

NamjaeChoi: Any suggestions on the flag name?

lindsayad: Oh yes my beautiful static condensation code providing the answer 😄

NamjaeChoi:
I vibe-coded a test. I will try to verify it as a human, but I wonder if you could also have a look at the test.

src/base/dof_map_constraints.C
roystgnr: Hmm... `ghost_ghosted_dependencies`?  I'd be open to other suggestions too.

roystgnr: Yours was better.

roystgnr (approved):
This looks good to me.  I've got enough paranoia that I've turned on the dbg-mode DistributedMesh sweeps, but enough hubris that I'm expecting them to pass and turning on auto-merge.

roystgnr:
Wait, where's my auto-merge button?  My fault for waiting to add the extra recipes until the standard ones had all passed, probably.

NamjaeChoi:
Hmm the unit test only covers the case with two partitions (rank 0 and n - 1). Not sure if sweeps make difference. Should it be more?

roystgnr:
Oh, it's not the sweep that makes a difference, it's the `METHOD=dbg`.  Though if I'd been thinking I'd only have turned on the 1-4 sweep.

NamjaeChoi:
Looks good!

src/base/dof_map_constraints.C
```diff
@@ -5344,18 +5344,24 @@ void DofMap::add_constraints_to_send_list()
...
-  if (!this->has_static_condensation())
+  if (!this->has_static_condensation() && !_need_ghost_constraints)
     return;
 
   // If we use StaticCondensation, though, we may need constraint DoFs
   // constraining DoFs which are not local (they're on someone else's
   // node) but which are supported on local elements.  Let's get those
   // too if we have to.
   //
+  // Kokkos-MOOSE also requires that, because it cannot use VecSetValues
```
lindsayad, line 5355: For consistency with a review I left on @GiudGiud's PR earlier today I would've asked to remove the downstream context here but I was too slow

lindsayad: Additional context is of course fine but if it's applicable to one downstream package then it can likely be described in a general conceptual way applicable to other potential downstream users. Anyway Roy doesn't seem bothered by it and he's the top dog in this repo 😄

PR #4544, opened 2026-09-09
Variational smoother quadrature changes

GiudGiud:
This doesnt change the default yet
but it's evidence we can get something else from QNODAL that would be valuable


include/mesh/mesh_smoother_vsmoother.h
```diff
@@ -129,6 +129,19 @@ class VariationalMeshSmoother : public MeshSmoother
...
+   * VariationalSmootherSystem to integrate the distortion-dilation metric.
+   * The default (\p QGAUSS) samples the element interior only, which can miss
+   * degeneracies localized at element corners (e.g. an element collapsing
+   * toward one of its nodes). A vertex-sampling rule such as \p QTRAP,
+   * \p QSIMPSON, \p QNODAL, or \p QGAUSS_LOBATTO evaluates the metric at the
+   * element nodes so a folding corner is penalized (and flagged as tangled) as
+   * it should be. Must be called before smooth()/setup(). See
+   * VariationalSmootherSystem::set_quadrature_type() for details and caveats.
+   */
+  void set_quadrature_type(QuadratureType qt) { _quadrature_type = qt; }
```
GiudGiud, line 143: I think this all I need from MOOSE side

GiudGiud:
Test failure unrelated.
@roystgnr when you have time

tests/mesh/mesh_smoother_test.C
```diff
@@ -399,6 +401,37 @@ public:
...
+  // it looks like a triangle).
+  //
+  // For a valid quad, the signed areas of the four corner triangles---each
+  // formed by a vertex and its two edge-neighbors---all share the same sign. A
+  // sign flip means an interior angle has crossed 180 degrees: the element has
+  // folded at that corner. This corner-localized degeneracy is exactly what the
+  // default interior-point Gauss rule fails to "see" (the interior Jacobians can
+  // stay positive while a corner collapses), and what a vertex-sampling
+  // quadrature rule is meant to prevent.
+  static bool quadIsFolded(const Elem & elem)
```
roystgnr: Do we need a new function here?  Is `has_invertible_map` (or one of the quality metrics) not sufficient?

roystgnr (changes requested):
This is good enough as-is but I'm wondering if we can simplify the tests a bit

tests/mesh/mesh_smoother_test.C
GiudGiud: good call this should be doing it now

roystgnr (approved):

Issue #4252, opened 2025-09-16
Make Threads take n_threads arguments


lindsayad:
Currently we rely on the global `libMesh::n_threads()` but we want to be able to set thread counts differently for different MOOSE applications running in MultiApp contexts. Proposed by @GiudGiud

lindsayad:
I think we can fold this into https://github.com/libMesh/libmesh/issues/4185

PR #4547, opened 2026-09-10
Add additional thread id limiting clause

GiudGiud:
refs #4252

claude saw this while working on moose per-app threading

### DIFF
--- a/include/parallel/threads_pthread.h
+++ b/include/parallel/threads_pthread.h
@@ -322,7 +322,11 @@ void parallel_for (const Range & range, const Body & body,
   // encounters this line of code when -fopenmp is not passed to the
   // compiler.
 #ifdef LIBMESH_HAVE_OPENMP
-#pragma omp parallel for schedule (static)
+// Cap the OpenMP team to the actual number of threads this dispatch uses. Without this clause the
+// team defaults to the process-wide libMesh::n_threads(), so omp_get_thread_num() (consumed by
+// MOOSE's ParallelUniqueId) could exceed a reduced per-application thread count and index past
+// per-thread storage sized to that count.
+#pragma omp parallel for schedule (static) num_threads(actual_threads)
 #endif
   for (int i=0; i<static_cast<int>(actual_threads); i++)
     {
@@ -433,7 +437,11 @@ void parallel_reduce (const Range & range, Body & body,
   // warns about an "unknown pragma" if it encounters this line of
   // code when -fopenmp is not passed to the compiler.
 #ifdef LIBMESH_HAVE_OPENMP
-#pragma omp parallel for schedule (static)
+// Cap the OpenMP team to the actual number of threads this dispatch uses. Without this clause the
+// team defaults to the process-wide libMesh::n_threads(), so omp_get_thread_num() (consumed by
+// MOOSE's ParallelUniqueId) could exceed a reduced per-application thread count and index past
+// per-thread storage sized to that count.
+#pragma omp parallel for schedule (static) num_threads(actual_threads)
 #endif
   // The use of 'int' instead of unsigned for the iteration variable
   // is deliberate here.  This is an OpenMP loop, and some older


roystgnr (approved):
This looks right to me but IIRC @lindsayad just put this feature in and I'll wait to merge for a bit in case he wants a look at the fix too.

include/parallel/threads_pthread.h
```diff
@@ -322,7 +322,11 @@ void parallel_for (const Range & range, const Body & body,
   // encounters this line of code when -fopenmp is not passed to the
   // compiler.
 #ifdef LIBMESH_HAVE_OPENMP
-#pragma omp parallel for schedule (static)
+// Cap the OpenMP team to the actual number of threads this dispatch uses. Without this clause the
+// team defaults to the process-wide libMesh::n_threads(), so omp_get_thread_num() (consumed by
+// MOOSE's ParallelUniqueId) could exceed a reduced per-application thread count and index past
```
lindsayad:
```suggestion
// team defaults to the process-wide libMesh::n_threads(), so omp_get_thread_num()
// could exceed a reduced per-application thread count and index past
```
I don't think we should put downstream context upstream. If repository context is not currently sufficient, then we should add a test to make it so

include/parallel/threads_pthread.h
```diff
@@ -433,7 +437,11 @@ void parallel_reduce (const Range & range, Body & body,
   // warns about an "unknown pragma" if it encounters this line of
   // code when -fopenmp is not passed to the compiler.
 #ifdef LIBMESH_HAVE_OPENMP
-#pragma omp parallel for schedule (static)
+// Cap the OpenMP team to the actual number of threads this dispatch uses. Without this clause the
+// team defaults to the process-wide libMesh::n_threads(), so omp_get_thread_num() (consumed by
+// MOOSE's ParallelUniqueId) could exceed a reduced per-application thread count and index past
```
lindsayad, line 442:
```suggestion
// team defaults to the process-wide libMesh::n_threads(), so omp_get_thread_num()
// could exceed a reduced per-application thread count and index past
```

lindsayad (dismissed):
If we can reasonably add a unit test for this, we should

tests/parallel/parallel_test.C
```diff
@@ -10,6 +10,16 @@
 #include <set>
 #include <thread>
 
+// The OpenMP team cap these tests exercise (the num_threads(actual_threads)
+// clause on the parallel-for pragmas in threads_pthread.h) only exists in the
+// pthread threading backend's OpenMP code path. TBB provides its own
+// parallel_for/parallel_reduce and the no-threads backend never spawns a team,
+// so there is nothing to check in those configurations.
```
GiudGiud, line 17: @roystgnr for fact check on this

roystgnr: I'm not sure if this is necessarily true (Intel's worked with making TBB operate alongside OpenMP in the past, so I wouldn't be surprised if they did some of the "notify OpenMP about our own threading choices" stuff we do), but it's definitely not *guaranteed* to be false, which is what we'd need to risk reading OpenMP settings that we only ever might have written indirectly via TBB object arguments.  Plus I think with TBB we only set these things via proper local settings rather than the globals OpenMP uses.

This isn't how I'd have worded it but it's explanation enough.

roystgnr:
I'm still happy with the PR once Alex is ... and definitely want to wait for his re-review too.  I'm kind of kicking myself for not demanding a unit test, especially after (properly, that time) giving Namjae a hard time about coverage.

include/parallel/threads_pthread.h
lindsayad: I don't know why this was marked resolved

lindsayad (commented):
Other than the comment  this LGTM!

include/parallel/threads_pthread.h
GiudGiud: I ll pull it in the same commit and force push

GiudGiud: for the reason, I assumed it was git showing me the same thing twice and resolved

lindsayad (approved):

PR #4548, opened 2026-09-10
Add missing header

roystgnr:
This fixes --disable-second builds (which presumably lose an indirect inclusion) for me


roystgnr:
Merging right away; this is simple and it should be the last thing we need to fix devel→master CI

PR #4549, opened 2026-09-11
HDiv dphi partial support for --disable-second builds

roystgnr:
It would have been simpler to just disable all the tests to pass the `--disable-second` recipe, but supporting affine elements in that case was pretty straightforward.


roystgnr:
Added `Test No Second Derivatives`; I'll merge once that one (and at least one of the more normal cases) passes

PR #4550, opened 2026-09-11
Adapt neighbor finding logic to polys

GiudGiud:
Two issues:
- poly sides might have been created differently and they wont be ordered the same way
- a polygon side does not say true when == with a tri side of a tet for example

### DIFF
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -973,9 +973,9 @@ bool sides_are_the_same_face(const Elem & a, const Elem & b)
   if (a == b)
     return true;
 
-  const bool a_poly = dynamic_cast<const Polygon *>(&a);
-  const bool b_poly = dynamic_cast<const Polygon *>(&b);
-  if (a_poly == b_poly)
+  // The check above would have triggered already if they were both
+  // not polygons. If either is a polygon, we need to check the nodes
+  if (!a.runtime_topology() && !b.runtime_topology())
     return false;
 
   const unsigned int nv = a.n_vertices();
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -25,6 +25,7 @@
 #include "libmesh/libmesh_logging.h"
 #include "libmesh/elem.h"
 #include "libmesh/elem_range.h"
+#include "libmesh/face_polygon.h"
 #include "libmesh/mesh_tools.h" // For n_levels
 #include "libmesh/parallel.h"
 #include "libmesh/remote_elem.h"
@@ -952,6 +953,47 @@ UnstructuredMesh::~UnstructuredMesh ()
 
 
 
+namespace {
+/**
+ * \returns \p true if element sides \p a and \p b are the same face,
+ * for the purpose of linking them as neighbors in find_neighbors().
+ *
+ * This is normally just Elem::operator==, which compares (sorted) node
+ * ids.  But operator== first requires the two sides to have the same
+ * element type, so it never matches a Polyhedron's polygonal side
+ * against the TRI3/QUAD4 side of an adjacent standard element (tet,
+ * hex, ...) even when they are geometrically the same face.  At such a
+ * mixed interface -- exactly one side is a polygon -- we fall back to
+ * comparing the vertex node ids, which is what low_order_key() already
+ * keys on.  Purely standard/standard and polygon/polygon pairs are left
+ * entirely to operator==.
+ */
+bool sides_are_the_same_face(const Elem & a, const Elem & b)
+{
+  if (a == b)
+    return true;
+
+  const bool a_poly = dynamic_cast<const Polygon *>(&a);
+  const bool b_poly = dynamic_cast<const Polygon *>(&b);
+  if (a_poly == b_poly)
+    return false;
+
+  const unsigned int nv = a.n_vertices();
+  if (nv != b.n_vertices())
+    return false;
+
+  std::vector<dof_id_type> a_ids(nv), b_ids(nv);
+  for (unsigned int v = 0; v != nv; ++v)
+    {
+      a_ids[v] = a.node_id(v);
+      b_ids[v] = b.node_id(v);
+    }
+  std::sort(a_ids.begin(), a_ids.end());
+  std::sort(b_ids.begin(), b_ids.end());
+  return a_ids == b_ids;
+}
+}
+
 
 
 void UnstructuredMesh::find_neighbors (const bool reset_remote_elements,
@@ -1040,7 +1082,7 @@ void UnstructuredMesh::find_neighbors (const bool reset_remote_elements,
                         // for matching level() to avoid setting our
                         // neighbor pointer to any of our neighbor's
                         // descendants.
-                        if ((*my_side == *their_side) &&
+                        if (sides_are_the_same_face(*my_side, *their_side) &&
                             (element->level() == neighbor->level()))
                           {
                             // So share a side.  Is this a mixed pair
--- a/src/mesh/unstructured_mesh.C
+++ b/src/mesh/unstructured_mesh.C
@@ -955,7 +955,7 @@ UnstructuredMesh::~UnstructuredMesh ()
 
 namespace {
 /**
- * \returns \p true if element sides \p a and \p b are the same face,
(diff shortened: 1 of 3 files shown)


src/mesh/unstructured_mesh.C
```diff
@@ -952,6 +953,47 @@ UnstructuredMesh::~UnstructuredMesh ()
...
+ * entirely to operator==.
+ */
+bool sides_are_the_same_face(const Elem & a, const Elem & b)
+{
+  if (a == b)
+    return true;
+
+  const bool a_poly = dynamic_cast<const Polygon *>(&a);
+  const bool b_poly = dynamic_cast<const Polygon *>(&b);
+  if (a_poly == b_poly)
```
GiudGiud:
```suggestion
  if (!a_poly && a_poly == b_poly)
```
might be better

GiudGiud: ah not necessary, this only happens when a!=b already

src/mesh/unstructured_mesh.C
```diff
@@ -952,6 +953,47 @@ UnstructuredMesh::~UnstructuredMesh ()
...
+ * comparing the vertex node ids, which is what low_order_key() already
+ * keys on.  Purely standard/standard and polygon/polygon pairs are left
+ * entirely to operator==.
+ */
+bool sides_are_the_same_face(const Elem & a, const Elem & b)
+{
+  if (a == b)
+    return true;
+
+  const bool a_poly = dynamic_cast<const Polygon *>(&a);
```
roystgnr: Let's avoid the RTTI here.  Maybe use `runtime_topology()` for the test?

src/mesh/unstructured_mesh.C
```diff
@@ -952,6 +953,47 @@ UnstructuredMesh::~UnstructuredMesh ()
...
+  const bool b_poly = dynamic_cast<const Polygon *>(&b);
+  if (a_poly == b_poly)
+    return false;
```
roystgnr, line 979: If they're pointers to the same `Polygon` then they're the same face?  If that's correct then I'm going to need a comment explaining how.

roystgnr: Or if neither is a `Polygon` and they're both `null` for that matter?  Okay, I'm definitely confused at this point; I'll pause my review here.

roystgnr (changes requested):

src/mesh/unstructured_mesh.C
GiudGiud: if they are both null, then they were not polygons, then the a == b right above returned true already if they were the same

GiudGiud: > If they're pointers to the same Polygon then they're the same face?

why not? If they are pointers to the same tri then they are the same face, that's our critierion for tets

src/mesh/unstructured_mesh.C
GiudGiud: nice yes that's a good idea

GiudGiud: I was wondering the same but my thinking only got so far as can I check the type and that seemed brittle. runtime_topology is best logic here

src/mesh/unstructured_mesh.C
roystgnr: Oh, I'm an idiot - I missed the `a==b` above and I keep forgetting that we don't allow a different `Polygon` from each `Polyhedron`.  Yeah, this is good.

src/mesh/unstructured_mesh.C
```diff
@@ -952,6 +953,47 @@ UnstructuredMesh::~UnstructuredMesh ()
...
+ * keys on.  Purely standard/standard and polygon/polygon pairs are left
+ * entirely to operator==.
+ */
+bool sides_are_the_same_face(const Elem & a, const Elem & b)
+{
+  if (a == b)
+    return true;
+
+  // The check above would have triggered already if they were both
+  // not polygons. If either is a polygon, we need to check the nodes
```
roystgnr:
```suggestion
  // The check above would have been sufficient if they were both
  // polygons or both not. If just one is a polygon, we need to check the nodes
```

src/mesh/unstructured_mesh.C
```diff
@@ -952,6 +953,47 @@ UnstructuredMesh::~UnstructuredMesh ()
...
+ * ids.  But operator== first requires the two sides to have the same
+ * element type, so it never matches a Polyhedron's polygonal side
+ * against the TRI3/QUAD4 side of an adjacent standard element (tet,
+ * hex, ...) even when they are geometrically the same face.  At such a
+ * mixed interface -- exactly one side is a polygon -- we fall back to
+ * comparing the vertex node ids, which is what low_order_key() already
+ * keys on.  Purely standard/standard and polygon/polygon pairs are left
+ * entirely to operator==.
+ */
+bool sides_are_the_same_face(const Elem & a, const Elem & b)
```
roystgnr: Let's rename this `sides_are_the_same_elem`, and correct the comment accordingly?  (We are still hitting this method when the sides are edges or nodes rather than faces, right?)

tests/mesh/mesh_generation_test.C
```diff
@@ -383,6 +423,70 @@ public:
...
+  // standard element (here a TET4) across their shared triangular face.
+  // The polyhedron's side is a C0POLYGON while the tet's side is a TRI3,
+  // so Elem::operator== alone (which requires matching element types)
+  // would not match them; find_neighbors() falls back to a vertex-id
+  // comparison at such a mixed interface.
+  void polyhedronTetNeighbor ()
+  {
+    LOG_UNIT_TEST;
+
+    ReplicatedMesh mesh(*TestCommWorld, /*dim=*/3);
```
roystgnr: Let's just make this a `Mesh`, so we get `DistributedMesh` coverage too?  We can loop over elements and assert that each has one neighbor without having to hang on to specific `add_elem` results that might be remoted.

roystgnr (changes requested):

src/mesh/unstructured_mesh.C
GiudGiud:
```suggestion
bool sides_are_the_same_elem(const Elem & a, const Elem & b)
```

tests/mesh/mesh_generation_test.C
GiudGiud:
```suggestion
    Mesh mesh(*TestCommWorld, /*dim=*/3);
```

src/mesh/unstructured_mesh.C
```diff
@@ -1040,7 +1082,7 @@ void UnstructuredMesh::find_neighbors (const bool reset_remote_elements,
                         // for matching level() to avoid setting our
                         // neighbor pointer to any of our neighbor's
                         // descendants.
-                        if ((*my_side == *their_side) &&
+                        if (sides_are_the_same_face(*my_side, *their_side) &&
```
GiudGiud:
```suggestion
                        if (sides_are_the_same_elem(*my_side, *their_side) &&
```

src/mesh/unstructured_mesh.C
```diff
@@ -952,6 +953,47 @@ UnstructuredMesh::~UnstructuredMesh ()
 
 
 
+namespace {
+/**
+ * \returns \p true if element sides \p a and \p b are the same face,
```
GiudGiud:
```suggestion
 * \returns \p true if element sides \p a and \p b are the same elem,
```

src/mesh/unstructured_mesh.C
GiudGiud: done

GiudGiud: and yes this has replaced the ==

GiudGiud:
wow I can resolve comments I have rights now

roystgnr:
Oh, we can't *just* `s/\<Mesh/DistributedMesh`; we really do also need to replace those `add_elem` return values with a loop over elements.  Once you have more than 2 procs you end up with procs where there are 0 local and thus 0 ghosted elements, and those pointers become dangling.

GiudGiud:
smh was passing locally in parallel so it's a bigger rework of the test. Covers distributed now

roystgnr (approved):

GiudGiud:
Thanks for the review!

PR #4551, opened 2026-09-14
Add quality metrics for quads (and all elements for some)

GiudGiud:
~~not rebased yet, too early for review~~

- a different skew metric
- condition number
- size
- taper for quads

aside from different skew metric, mostly filling unimplemented options


include/enums/enum_elem_quality.h
```diff
@@ -52,7 +52,8 @@ enum ElemQuality : int {
                   EDGE_LENGTH_RATIO,
                   MAX_DIHEDRAL_ANGLE,
                   MIN_DIHEDRAL_ANGLE,
-                  SCALED_JACOBIAN};
+                  SCALED_JACOBIAN,
+                  SKEW_ANGLE};
```
GiudGiud, line 56: this because our current SKEW actually is "SHEAR" there:
https://coreform.com/cubit_help/mesh_generation/mesh_quality_assessment/quadrilateral_metrics.htm

src/geom/cell_hex.C
```diff
@@ -516,11 +553,12 @@ std::pair<Real, Real> Hex::qual_bounds (const ElemQuality q) const
       bounds.second = 4.;
       break;
 
-    case SKEW:
+    case SKEW_ANGLE:
       bounds.first  = 0.;
       bounds.second = 0.5;
```
GiudGiud, line 558: the ranges for our skew were wrong, 1 was best, 0 was bad

src/geom/elem.C
```diff
@@ -1996,6 +1996,164 @@ Real Elem::quality (const ElemQuality q) const
...
+      // edges meeting at a node), averaged over the corner nodes.
+      //
+      // Both undersized (J < 1) and oversized (J > 1) elements are
+      // penalized, and an element the size of the unit reference
+      // element scores the ideal value of 1. This differs from the
+      // Verdict/CUBIT relative size, which normalizes J by the
+      // mesh-average element size; that requires mesh-wide context not
+      // available to this per-element method, so we use the reference
+      // element instead. Unlike the standard Verdict metric, J is not
+      // squared here.
```
GiudGiud: this makes it less useful. maybe to the point i should drop the change?

src/geom/cell_hex.C
```diff
@@ -491,6 +491,43 @@ Real Hex::quality (const ElemQuality q) const
...
+          X1 = (x1 - x0) + (x2 - x3) + (x5 - x4) + (x6 - x7),
+          X2 = (x3 - x0) + (x2 - x1) + (x7 - x4) + (x6 - x5),
+          X3 = (x4 - x0) + (x5 - x1) + (x6 - x2) + (x7 - x3);
+
+        const Real n1 = X1.norm(), n2 = X2.norm(), n3 = X3.norm();
+
+        // Degenerate element: return 0 (the Verdict convention) if any
+        // principal axis has zero length.
+        if (n1 == 0. || n2 == 0. || n3 == 0.)
+          return 0.;
```
roystgnr, line 522: Not a fan of this.  I guess I'm okay matching their convention, but let's make sure to put it in the `describe()` string too.

src/geom/elem.C
```diff
@@ -1996,6 +1996,164 @@ Real Elem::quality (const ElemQuality q) const
...
+      // edge vectors as its columns; its (Frobenius-norm) condition
+      // number is kappa = |A|_F * |A^{-1}|_F / N. Following the other
+      // algebraic metrics (SHAPE, SKEW), the reference (weight) matrix
+      // is the identity, so kappa = 1 for an orthogonal, equal-length
+      // (ideal) corner and grows without bound as the corner is
+      // stretched or skewed. A degenerate corner (zero Jacobian
+      // determinant) has an infinite condition number, reported as 0
+      // following the convention used elsewhere (e.g. EDGE_LENGTH_RATIO)
+      // that 0 stands in for infinity.
+    case CONDITION:
```
roystgnr, line 2086: We thoughtfully added `qual_bounds(CONDITION)` recommendations for 4 different categories of elements, but we didn't previously have an implementation for `quality(CONDITION)`?

Oh, wow.  Looks like those bounds were added by Ben, in 2002 or 2003, in libMesh's 3rd non-trivial CVS commit, the 1st commit with files for some Elem subclasses, presumably as placeholders in the hope that somebody would add the implementation *someday*.

This PR is awesome.  "Lafayette, we are here!"

src/geom/elem_quality.C
```diff
@@ -162,6 +166,17 @@ std::string Quality::describe (const ElemQuality q)
...
+           << "based on the nodal Jacobian\n"
+           << "skew matrices. 1 is ideal,\n"
+           << "smaller values are worse.\n"
+           << '\n'
+           << "Suggested ranges:\n"
+           << "Hexes: (0.3 -> 1)\n"
+           << "Quads: (0.3 -> 1)";
+      break;
+
+    case SKEW_ANGLE:
```
roystgnr, line 179: Really good fix here, but pretty damning of our test coverage that we had bounds for one metric but were computing another.

In elem_test.C, should we add a new test that loops `q` over `Quality::valid(elem->type())`, evaluates `elem->quality(a)` and asserts (since we're on simple default meshes there) that the result is within `elem->qual_bounds(q)`?

roystgnr (changes requested):
I'd love another unit test too, but beggars can't be choosers, and the `describe()` clarification is the important thing IMHO.

src/geom/elem.C
roystgnr: Oh, I forgot to respond to this bit, sorry.

IMHO normalizing by the reference element size isn't very useful, but normalizing by the "ideal" Jacobian, the |J| for a perfectly-shaped element of the same `volume()`, would give something valuable.

Oh, no.  Oh no, no, no.

A "perfectly-shaped element" is *not*, in general, proportional to our master element.  Our edge/square/cube elements are fine as exemplars, but our prisms+pyramids are debateable and our tris+tets are outright wrong.  For this and for `CONDITION` we need to be getting the Jacobian for the transformation from an ideal element, not from a master element, just like we had to do in the variational mesh smoother.

include/geom/reference_elem.h
```diff
@@ -46,6 +52,21 @@ namespace ReferenceElem
...
+ * of the reference element. For element types that have no distinct
+ * ideal shape (e.g. quads and hexes, whose reference element is already
+ * regular), this returns a copy of the reference element.
+ *
+ * The returned Elem holds pointers into the returned Nodes, so the
+ * caller must keep the Node vector alive for at least as long as the
+ * Elem.
+ */
+std::pair<std::unique_ptr<Elem>, std::vector<std::unique_ptr<Node>>>
+ideal_target (const ElemType type);
```
GiudGiud, line 68: I put it here since reference not too far from ideal ideologically

could also just go in elem.h

src/geom/cell_hex.C
GiudGiud: added

src/geom/elem.C
GiudGiud: ok using the ideal element from the variable smoother now

src/geom/cell_hex.C
GiudGiud: done

src/geom/elem_quality.C
GiudGiud: added that test

GiudGiud:
tests need fixing, on it

include/geom/reference_elem.h
roystgnr: I really like it here.

tests/geom/elem_test.C
```diff
@@ -253,6 +254,86 @@ public:
       }
   }
 
+  void test_quality_bounds()
+  {
+    LOG_UNIT_TEST;
+
+    // For a well-shaped, unit-scale "ideal" element of this type, every
```
roystgnr, line 261: Do we need "ideal" here?  I'd rather test the elements we've already built, and get coverage for everything, instead of adding more code to get coverage of just four cases.  And if it turns out that our build_cube() is ever giving us poor-quality elements we'd want to know that too.

include/geom/reference_elem.h
GiudGiud: sounds good resolving

src/geom/elem.C
```diff
@@ -1996,23 +1997,17 @@ Real Elem::quality (const ElemQuality q) const
...
-      // element scores the ideal value of 1. This differs from the
-      // Verdict/CUBIT relative size, which normalizes J by the
-      // mesh-average element size; that requires mesh-wide context not
-      // available to this per-element method, so we use the reference
-      // element instead. Unlike the standard Verdict metric, J is not
-      // squared here.
+      // Relative size metric: min over the corner nodes of min(tau,
+      // 1/tau), where tau is the ratio of the corner's nodal Jacobian
+      // determinant to the determinant of the nodal Jacobian of an
+      // ideal (regular) element of the *same volume*. We take that
```
roystgnr, line 2003: But that ratio shouldn't be what we're looking for; you could skew the master element in other directions and end up with the same ratio of determinants.

If we have mapping Jacobians J (from master element to physical element, the usual) and K (from master element to an ideal element of the same volume()), we don't want |J|/|K|, we want |JK⁻¹|

roystgnr (changes requested):

src/geom/elem.C
roystgnr: I'm a little worried that we can't trust Claude with some of this stuff.  Maybe we should add a section to that unit test that runs a test element through all its permutations and asserts that each quality() is unchanged by permutation?

GiudGiud: > we don't want |J|/|K|, we want |JK⁻¹|

I think we should be fine here since det(AB) = det(A) det(B) ?

> maybe we should add a section to that unit test that runs a test element through all its permutations and asserts that each quality() is unchanged by permutation?

ok I ll look into that

roystgnr: Of course you're right.  I remember math, I swear I do.  Let's add the permutation test anyway, though?

PR #4554, opened 2026-09-16
Generate second-order node maps from the linear tables

rochi00:
Put the second-order element side/edge topology and reference-node
coordinates into one constexpr derivation in
fe_reference_element_traits.h: a second-order side row is the linear
corners, then the edge midpoints from the edge table, then the face
centre; higher-order reference nodes are subentity centroids.

The element classes' side_nodes_map/edge_nodes_map are now generated
from that derivation at compile time instead of being 26 hand-typed
literals.  Every table is bit-identical to the one it replaces and is
constant-initialized.  The maps keep their names, their [side][node]
indexing and their public visibility (MOOSE ray tracing reads them
directly); only their type changes, to a ReferenceElementTable that
indexes like the old arrays.


include/fe/fe_reference_element_traits.h
```diff
@@ -0,0 +1,965 @@
...
+#include "libmesh/enum_elem_type.h"
+#include "libmesh/libmesh.h"
+#include "libmesh/libmesh_device.h"
+#include "libmesh/point.h"
+
+namespace libMesh
+{
+
+template <unsigned int N>
+struct ReferenceElementVector
```
roystgnr: This looks like an atavism that never got used?

include/geom/cell_hex20.h
```diff
@@ -220,13 +221,13 @@ class Hex20 final : public Hex
    * This maps the \f$ j^{th} \f$ node of the \f$ i^{th} \f$ side to
    * element node numbers.
    */
-  static const unsigned int side_nodes_map[num_sides][nodes_per_side];
+  static const ReferenceElementTable<num_sides, nodes_per_side> side_nodes_map;
```
roystgnr: These are public members; is there any way we can avoid changing their type and thus breaking backwards compatibility?

include/fe/fe_reference_element_traits.h
```diff
@@ -0,0 +1,965 @@
...
+    {1, 4, 10},
+    {2, 5, 11},
+    {3, 4, 12},
+    {4, 5, 13},
+    {3, 5, 14}
+  }};
+}
+
+LIBMESH_DEVICE_INLINE constexpr ReferenceElementTable<8, 3>
+pyramid_edge_nodes()
```
roystgnr: Let's have a less-broad name for this?  pyramid_quadratic_edge_nodes() maybe?  It's valid for Pyramid13/14/18, but not Pyramid5 and not for many hypothetical future subclasses like a cubic Pyramid30.

roystgnr: Same issue with tet, hex, and prism cases; pyramid is just where I noticed it first.

include/fe/fe_reference_element_traits.h
```diff
@@ -0,0 +1,965 @@
...
+  LIBMESH_DEVICE_INLINE constexpr const Row & operator[](unsigned int i) const
+  { return values[i]; }
+
+  LIBMESH_DEVICE_INLINE constexpr unsigned int operator()(unsigned int i, unsigned int j) const
+  { return values[i][j]; }
+};
+
+
+LIBMESH_DEVICE_INLINE constexpr ReferenceElementTable<5, 4>
+prism6_side_nodes()
```
roystgnr: Can we move the element-specific cases to element-specific headers?  E.g. cell_prism6.h for this and cell_prism.h for the quadratic version?

include/fe/fe_reference_element_traits.h
```diff
@@ -0,0 +1,965 @@
...
+    {0, 3, 8},
+    {0, 4, 9},
+    {1, 4, 10},
+    {2, 4, 11},
+    {3, 4, 12}
+  }};
+}
+
+LIBMESH_DEVICE_INLINE bool
+requires_side_specific_topology(ElemType parent)
```
roystgnr: These should probably be static member functions (or static arrays like `type_to_n_sides_map` in the simpler cases) in `Elem::`.

include/fe/fe_reference_element_traits.h
```diff
@@ -0,0 +1,965 @@
...
+    case PYRAMID14:
+    case PYRAMID18:
+      return true;
+    default:
+      return false;
+  }
+}
+
+LIBMESH_DEVICE_INLINE ElemType
+side_topology_or_invalid(ElemType parent,
```
roystgnr: This is just `Elem::side_type` but using an `ElemType` instead of `this`, and we can probably just overload that name.  "topology" is a bit overwrought for something literally returning `ElemType`, and if we had to add "or_invalid" to any method that might return an error flag then our APIs would require a *lot* more typing.

include/fe/fe_reference_element_traits.h
roystgnr: And the name here ... this isn't a "requirement" so much as a fact.  Maybe `type_to_has_multiple_side_types`?

include/fe/fe_reference_element_traits.h
```diff
@@ -0,0 +1,965 @@
...
+      return pyramid_tri ? TRI7 : QUAD9;
+    default:
+      return INVALID_ELEM;
+  }
+}
+
+// Mid-side node counts are uniform per element type except for the
+// prisms and pyramids, whose triangular and quadrilateral faces differ.
+LIBMESH_DEVICE_INLINE constexpr unsigned int
+side_node_count_or_zero(ElemType parent,
```
roystgnr: `or_zero` is another bad naming, and `n_nodes` would match our existing naming better than `node_count`.

include/fe/fe_reference_element_traits.h
```diff
@@ -0,0 +1,965 @@
...
+  }
+}
+
+// Every element type with edge nodes has 3 nodes (two vertices plus a
+// midpoint) on each of its edges; only the edge count varies.  The edge
+// tables cover the second-order families only: the linear elements'
+// vertex-pair edges stay with their classes, which device code never
+// queries.
+LIBMESH_DEVICE_INLINE constexpr unsigned int
+edge_node_count_or_zero(ElemType parent,
```
roystgnr: Likewise `edge_n_nodes`

include/fe/fe_reference_element_traits.h
roystgnr: And the variable name!  `parent` has a long-standing meaning in libMesh!  This is not it, and that's too confusing.

include/fe/fe_reference_element_traits.h
roystgnr: No, wait, I'm also not even thinking far enough here.  If we have a function returning an edge `ElemType` based on an elem's type (which we should; it'll be simpler than the side case above), and we have a constexpr way to return a number of nodes for a side (we do; just turn `Elem::type_to_n_nodes_map` from `const` to `constexpr`), the correct way to compose functions is `Foo(Bar(x))`, not `Foo_of_Bar(x)`, because we like having N functions and not N².

include/fe/fe_reference_element_traits.h
```diff
@@ -0,0 +1,965 @@
...
+    case PYRAMID14:
+    case PYRAMID18:
+      return edge < 8 ? 3 : 0;
+    default:
+      return 0;
+  }
+}
+
+LIBMESH_DEVICE_INLINE constexpr bool
+try_local_edge_node(ElemType parent,
```
roystgnr: Documentation?

include/fe/fe_reference_element_traits.h
```diff
@@ -0,0 +1,965 @@
...
+    case PYRAMID18:
+      node = pyramid_edge_nodes()(edge, edge_node);
+      return true;
+    default:
+      return false;
+  }
+}
+
+LIBMESH_DEVICE_INLINE constexpr ElemType
+linear_sibling_or_invalid(ElemType type)
```
roystgnr: Like "parent", AMR/C called dibs on "sibling" first.

Is this just `Elem::first_order_equivalent_type()`, but rewritten with a different name?  We cannot keep letting AI try to slip that by us!  That does not get us from an API with N functions to one with N², it gets us from N to ∞.

include/fe/fe_reference_element_traits.h
```diff
@@ -0,0 +1,965 @@
...
+      return side == 4 ? 13 : invalid_uint;
+    case PYRAMID18:
+      return side == 4 ? 13 :
+             side < 4 ? 14 + side : invalid_uint;
+    default:
+      return invalid_uint;
+  }
+}
+
+// Corner k of a linear element's side, from the stored linear tables.
```
roystgnr: What does "corner" mean here?  vertex?  But we don't seem to be doing any tests on k here; this will happily return a non-vertex.

roystgnr: This also seems like something that could be a composition rather than a single function.

include/fe/fe_reference_element_traits.h
```diff
@@ -0,0 +1,965 @@
...
+    default:
+      return 0;
+  }
+}
+
+LIBMESH_DEVICE_INLINE constexpr bool
+try_local_edge_node(ElemType parent,
+                    unsigned int edge,
+                    unsigned int edge_node,
+                    unsigned int & node)
```
roystgnr: We have an `invalid_uint`; could we just return that for the `false` case rather than a separate `bool`?

include/fe/fe_reference_element_traits.h
roystgnr: We can drop the `try_` on all these names; this is the same thing as our `local_edge_node` but taking an input `ElemType` (which again shouldn't be called "parent") rather than `Elem * this`, right?  We can just overload `Elem::local_edge_node` with a static, right?  I'd prefer to throw an error (ideally there'd be some way to `libmesh_assert()` in runtime invocations while doing `static_assert` at compile-time, though I confess I'm not sure how) when we're asking for an invalid node, but if we really do need to call functions like these in cases where it might return `false`, we can return `invalid_uint` instead.

roystgnr (changes requested):
Didn't quite make it through the whole thing, but this should give you enough to chew on for a first pass.  ;-)

I love this in theory, but in practice?  Adding 1000 lines of code to replace under 300 is never a good sign for a refactor.  I assume some of this is prerequisite for Kokkos work, giving us static lookups for data we previously had to get via virtual functions?  Those cases count as genuinely new features and justify some new lines, but I'll bet we can factor this way down.

tests/geom/elem_test.C
```diff
@@ -75,6 +75,129 @@ public:
...
+  {
+    LOG_UNIT_TEST;
+
+    for (const auto & elem : this->_mesh->active_local_element_ptr_range())
+      {
+        // Polytopes have no compile-time topology to compare against
+        if (elem->runtime_topology())
+          continue;
+
+
```
roystgnr: Too much white space here!

tests/geom/elem_test.C
```diff
@@ -75,6 +75,129 @@ public:
...
+              {
+                CPPUNIT_ASSERT_EQUAL(elem->local_side_node(s, n),
+                                     Elem::local_side_node(type, s, n));
+                CPPUNIT_ASSERT_EQUAL(nodes[n],
+                                     Elem::local_side_node(type, s, n));
+              }
+          }
+
+        // 1D elements have no edges, and Face::local_edge_node() makes a
+        // 2D element's edges its sides, which the loop above covered
```
roystgnr: The loop above did cover this in spirit but didn't cover the same APIs; we should leave the 2D test coverage in here.

roystgnr (changes requested):
Well, that was a massive improvement.

This PR is such a fundamental change that I don't want to approve until the `Coverage` recipe finishes and I look for gaps there, but based on my own eyeballs alone we seem to have gone from "dozens of ugly flaws" to "I have one kinda-pedantic nitpick and one comically pedantic nitpick".

I'll also ping @jwpeterson at this point, but just on the same "fundamental change" grounds, not because we _need_ an extra review.

Issue #4555, opened 2026-09-17
Add poly reading in VTK reader


GiudGiud:
exodus is nice but it's a lot harder to get there with the paraview plugin not supporting polys

The IOSS plugin's writer produces ill-formed files with polys

and meshio also crashes on vtu -> exodus with polys

PR #4556, opened 2026-09-17
Read VTK_POLYHEDRON cells in the VTU reader

GiudGiud:
refs #4555

did not do polygons

VTKIO::read() now handles VTK_POLYHEDRON cells, which VTK describes by a face stream rather than an ordered node list.  Each such cell is built as a C0Polyhedron from C0Polygon faces (obtained via
vtkUnstructuredGrid::GetFaceStream), mirroring the Exodus NFACED reader; the interior mid-element node the C0Polyhedron construction may create is added to the mesh.

Adds a testVTKReadPolyhedra unit test that reads a hexagonal-prism polyhedron .vtu fixture and checks the element type, side and face connectivity.


roystgnr (changes requested):
This looked fine to me, but I figured I'd attach some applicable extra test recipes to be paranoid, and I guess it paid off.  I'm somewhat baffled by the linker failures, though.  We're not introducing any new std::filesystem usage here, are we?

Issue #4019, opened 2024-11-26
Make Exodus writer handle a mesh block with mixed types of elements


YaqiWang:
Exodus format requires each mesh block to have only one type of elements, which is not a requirement for libMesh data structure, nor possibly some other mesh formats. In cases that we have a mixed types of elements in a mesh and we use Exodus writer, we will get an error message. We possibly can improve the writer so that it can automatically split the mesh block based on the element types. If a mesh block only contains one element type, it will be a no-op, otherwise, it appends the block name with `_[ElemType]`, like `_TRI3`. In this way, the writer can still work properly without erroring out. Tag @roystgnr

PR #4557, opened 2026-09-17
Split element blocks by element type when writing exodus files

GiudGiud:
refs #4019

I need this because when reading from VTK and outputting to exodus straight away (to keep my pipeline with exodus in MOOSE, maybe I don't need to though), I dont have a step in libmesh to fix the element block assignments. I only have that in moose

build_subdomain_map() grouped elements by subdomain id alone, so a subdomain containing more than one element type (e.g. a mix of C0POLYHEDRON and HEX8 cells) became a single Exodus block whose type was taken from its first element.  That either errored ("all elements with a given subdomain ID [must] be the same type") or, when the first element was a standard type, crashed in ex_put_concat_all_blocks while writing the polyhedra as that type.

Group elements by (subdomain, element type): single-type subdomains keep their previous block id and behavior, while additional element types in a subdomain get synthesized block ids.  Exodus requires a single element type per block, so this is what lets a mixed polyhedron/hex mesh be written and read back.

Adds test_write_and_read_mixed_poly_hex, which round-trips a mesh with a C0POLYHEDRON and a HEX8 in the same subdomain.


src/mesh/exodusII_io_helper.C
```diff
@@ -206,34 +206,67 @@ const std::vector<int> prism_inverse_face_map = {4, 1, 2, 3, 5};
...
+    subdomain_id_type next_synth_block_id = have_elems ? max_subdomain_id + 1 : 0;
+
+    for (auto & [sbd_id, type_map] : elems_by_subdomain_type)
+      {
+        bool first_type = true;
+        for (auto & [elem_t, elem_ids] : type_map)
+          {
+            libmesh_ignore(elem_t);
+            const subdomain_id_type block_id =
+              first_type ? sbd_id : next_synth_block_id++;
```
GiudGiud, line 249: do we want a warning here? or just an info message

I m leaning info message, and adding the new entry into a subdomain name map

src/mesh/exodusII_io_helper.C
GiudGiud: this is no longer correct... 
claude not so smart

src/mesh/exodusII_io_helper.C
```diff
@@ -3021,7 +3058,43 @@ void ExodusII_IO_Helper::write_elements(const MeshBase & mesh, bool use_disconti
...
+
+          std::string block_name = mesh.subdomain_name(elem_subdomain_id);
+
+          if (is_synthesized_block)
+            {
+              // Prefix the (original) subdomain's name with the element
+              // type, so the synthesized block is identifiable and its
+              // name stays distinct from the block that kept the
+              // subdomain id.  This block becomes its own subdomain when
+              // the file is read back in.
```
GiudGiud, line 3080: that might be undesirable but also merging someone's random mesh same-name blocks (to try to recover the original subdomain assignments from subdomains names being the same and IDs being different) might also not work well

GiudGiud:
moose failures expected, patch here: 
https://github.com/idaholab/moose/pull/33792

PR #4558, opened 2026-09-21
Skip the spatial global indexing when temporarily partitioning a replicated mesh

lindsayad:
`Partitioner::partition()` assigns a temporary partitioning to any unpartitioned elements before handing the mesh to the real partitioner, so that every element carries a valid processor id. `partition_unpartitioned_elements()` built that assignment from a spatial global indexing, which costs a parallel sort over the range.

On a replicated mesh every processor traverses the elements in the same order, so an element's position in the range is already a globally consistent index into `[0, n_unpartitioned_elements)`, and the sort only buys a different, equally arbitrary linear split of the same range. Use the position directly there and keep the spatial indexing for distributed meshes, where traversal order differs between processors.

This matters most for a freshly generated mesh, where no element is partitioned yet and the temporary assignment therefore covers the whole mesh before being overwritten. On a 262144 element mesh partitioned onto 24 processors, `partition_unpartitioned_elements()` falls from 0.706 s to 0.018 s and the enclosing `partition()` from 1.86 s to 1.14 s. For context, METIS itself accounts for 0.13 s of that.

(body written by Claude)


roystgnr:
We don't do a Hilbert sort because we need a globally consistent ordering but didn't notice that we already had one; we started doing it it because we noticed that our "black box" preconditioners gave better-quality outputs when starting from better-quality inputs, and contiguous sections of a libHilbert SFC were pretty good at that.

Maybe that's out of date today, with decades of METIS+ParMETIS improvements?  But I'd like some evidence of that.  It wouldn't take much less efficiency on a quarter-element mesh for solver timing changes to undo a 0.7s generation savings.

lindsayad:
> We don't do a Hilbert sort because we need a globally consistent ordering but didn't notice that we already had one; we started doing it it because we noticed that our "black box" preconditioners gave better-quality outputs when starting from better-quality inputs, and contiguous sections of a libHilbert SFC were pretty good at that.
> 
> Maybe that's out of date today, with decades of METIS+ParMETIS improvements? But I'd like some evidence of that. It wouldn't take much less efficiency on a quarter-element mesh for solver timing changes to undo a 0.7s generation savings.

I think I just saw this phenomenon on an MFEM simulation just today

Issue #4560, opened 2026-09-22
CheckpointIO split-not-found error doesn't report which splits actually exist


kyriv-lab:
## Bug Description

When `CheckpointIO::select_split_config()` can't find a header file for
the requested split count (and the serial/single-split fallback also
fails), the error only names the two paths it looked for:

ERROR: Neither one of the following files can be located:
    '/path/to/mesh.cpr/10/header.cpr' nor
    '/path/to/mesh.cpr'
If you are running a parallel job, double check that you've created a
split for 10 ranks.

This doesn't tell the user what splits *do* exist under
`mesh.cpr/`, so they're left guessing whether they ran with the wrong
number of ranks, pointed at the wrong file entirely, or need
`--distributed-mesh`. Reported downstream as
idaholab/moose#23913, where the same confusion led to a separate,
also-unhelpful error in MOOSE's own restart/recovery file lookup
(idaholab/moose#23910).

## Steps to Reproduce

Point `CheckpointIO`/`Mesh/file=` at a `.cpr` checkpoint directory that
only contains a split for N ranks, then run with a different rank count.

## Design

When the header lookup fails for both the requested and the
single-split fallback, look at the numbered subdirectories that exist
under the checkpoint directory and report them (e.g. "Splits that do
exist: 1, 4, 16") instead of only reporting the paths that were tried.

## Impact

Confusing, unhelpful error message; lost user time debugging what should
be an easy fix (correct `-n`/rank count or split file).

PR #4561, opened 2026-09-22
Report existing splits when a checkpoint split cannot be found (#4560)

kyriv-lab:
This PR resolves issue #4560


roystgnr (changes requested):
Looks like a great UI improvement, but could we get a unit test for it too?

roystgnr (approved):

PR #4563, opened 2026-09-22
Allow elemental Exodus output of higher-order discontinuous variables

### LINKED ISSUE #4562: Support elemental Exodus output for higher-order discontinuous (non-CONSTANT) variables
Resolves idaholab/moose#31427. 

 MOOSE's Exodus output of higher-order (order > 0) discontinuous MONOMIAL variables currently falls back to nodal output, which averages values across neighboring elements and smears sharp features (e.g. at subdomain interfaces).

 The blocker is on the libMesh side:
 - EquationSystems::is_elemental_data_fe_type() only accepts CONSTANT order.
 - EquationSystems::build_parallel_elemental_solution_vector() copies a single DOF per component, which is only valid for CONSTANT order.
- ExodusII_IO::write_element_data() has no way to opt into elemental output for these variables.

 The fix should be opt-in only, since existing exodiff-based regression tests depend on the current default behavior.

kyriv-lab:
Fixes #4562


kyriv-lab:
@roystgnr please review when you get the chance.

